PR #6296, opened 2020-01-16
wallet2: reject zero keys in json input

moneromooo-monero:
Because the user might do this for reasons unknown.
Values beyond l-1 will be reduced, so are accepted.

Reported by who-biz.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -591,6 +591,8 @@ std::pair<std::unique_ptr<tools::wallet2>, tools::password_container> generate_f
       }
       viewkey = *reinterpret_cast<const crypto::secret_key*>(viewkey_data.data());
       crypto::public_key pkey;
+      if (viewkey == crypto::null_skey)
+        THROW_WALLET_EXCEPTION(tools::error::wallet_internal_error, tools::wallet2::tr("view secret key may not be all zeroes"));
       if (!crypto::secret_key_to_public_key(viewkey, pkey)) {
         THROW_WALLET_EXCEPTION(tools::error::wallet_internal_error, tools::wallet2::tr("failed to verify view key secret key"));
       }
@@ -607,6 +609,8 @@ std::pair<std::unique_ptr<tools::wallet2>, tools::password_container> generate_f
       }
       spendkey = *reinterpret_cast<const crypto::secret_key*>(spendkey_data.data());
       crypto::public_key pkey;
+      if (spendkey == crypto::null_skey)
+        THROW_WALLET_EXCEPTION(tools::error::wallet_internal_error, tools::wallet2::tr("spend secret key may not be all zeroes"));
       if (!crypto::secret_key_to_public_key(spendkey, pkey)) {
         THROW_WALLET_EXCEPTION(tools::error::wallet_internal_error, tools::wallet2::tr("failed to verify spend key secret key"));
       }


who-biz:
who-biz submitted a report to Monero.	Jan 9th (8 days ago)
Summary:

Due to the use of boost::value_initialized() many places in code for initialization of various values, the wallet and network interpret addresses created with nulled secret keys as valid. More troublesome is the fact that when a wallet is created using these values as recovery seeds, that the pubkeys resulting have a value equivalent to the rct::identity() element.

Releases Affected:

All.

Steps To Reproduce:

[ Create a json-formatted file null-init.jsonwith the following contents:
```
{
    "version": 1,
    "filename":  "null_init",
    "scan_from_height": 0,
    "viewkey": "0000000000000000000000000000000000000000000000000000000000000000",
    "spendkey": "0000000000000000000000000000000000000000000000000000000000000000",
    "password": ""
}
```
Start monero-rpc-wallet with the following options:

`./monero-wallet-rpc --generate-from-json /path/to/null-initjson --rpc-bind-port=1212
Open with monero-walet-cli, to verify address creation:

 ``` This is the command line monero wallet. It needs to connect to a monero daemon to work correctly. WARNING: Do not reuse your Monero keys on another fork, UNLESS this fork has key reuse mitigations built in. Doing so will harm your privacy.
Monero 'Carbon Chamaeleon' (v0.15.0.1-release)
Logging to ./monero-wallet-cli.log
Wallet password: 
Opened wallet: 41fJjQDhryD111111111111111111111111111111111113CUsUpv9u11111111111111111111111111111111117KTBwG
```
## Impact

The resulting wallet will be able to function normally on the network.   The fact that the created wallet has pubkeys with value `0000000000000000000000000000000000000000000000000000000000000000`  may mean that confidentiality of all transactions could be in question, as this value is normally used as a check against recv derivations in key expansions/checks.

who-biz posted a comment.	Updated Jan 9th (8 days ago)
Sorry, the impact section should read:

Impact

The resulting wallet will be able to function normally on the network. The fact that the created wallet has pubkeys with value 0100000000000000000000000000000000000000000000000000000000000000 may mean that confidentiality of all transactions could be in question, as this value is normally used as a check against recv derivations in key expansions/checks.

The rct::identity() scenario above, occurs when using the nulled values as recovery seeds, when generating a wallet. i.e. a non-deterministic restore from seed with viewkey seed = 0000000000000000000000000000000000000000000000000000000000000000 and spendkey seed of the same value.


who-biz posted a comment.	Jan 9th (8 days ago)
Please note that the address created above was never connected to the network in any way.


anonimal changed the status to Triaged.	Jan 10th (7 days ago)
Hi @who-biz, thank you for your report and it's good to see you again.

A couple quick notes to our readers for clarification:

boost::value_initialized(), or any initialization, is actually a good thing because C++ dynamic memory management can point to things we really don't want to point to. I'll defer to @moneromooo but, as @who-biz stated, I believe the problem is within the key-handling implementation and not necessarily any variable initialization.

Just a small thing: rct::identity() is referenced but since RingCT was introduced into 2016, I'm not sure if all monero releases are affected (but it's not like anyone is/should be using a release older than that). Without looking at the code, I can suspect that the issue does indeed apply to all releases because it doesn't seem to apply only to RingCT but I could be mistaken since I don't see any code snippets.

Thanks again, @moneromooo will respond soon.


who-biz posted a comment.	Jan 10th (7 days ago)
Thanks for the quick response. Good to see both of you again, as well. Glad my report could be more helpful this time. And yes, I agree that initializing variables is, in fact, a good practice. It’s simply the use of these values in the validation that presents an issue.

Also, you’re right on #2. Apologies - I am used to looking at code where RingCT is always enabled. Believe you’re probably right about it only affecting releases since 2016, in that case.

More than happy to help you guys work out a fix, as this affects BLUR as well, and I have not yet worked out a patch.

I think a good way to go about the fix might be using values for initialization that aren’t recognized as valid. (Not sure if 0xFF bytes would result in the same issue, but might be sufficient)


moneromooo posted a comment.	Jan 11th (7 days ago)
Can you point out where this causes a problem for the network (the check against 0 value you mention) ?

The "anyone can guess this key if you use it" is not a problem. If you start banning a key, why not ban the next one, etc ? I actually made a patch before to ban "small" keys for just this reason (ease of brute forcing - or so I thought), and one of the MRL researchers pointed out it makes the keys not be equiprobable, which has deleterious effects on the crypto (I don't know the details, but even a very small bias apparently has a disproportionate effect).


who-biz posted a comment.	Jan 11th (6 days ago)
My concern would be in checking against recv derivations, as I said. The short keys issue would indeed affect bias, which could be problematic. This issue is not the same, and has nothing to do with key bias, or ease of brute forcing.

Rather, this appears to affect confidentiality of transactions, since these keys aren’t actually valid curve points, but appear to check as valid... this could result in confidentiality loss, or maybe even the ability to spend funds that are not rightfully ours to spend, I think. This depends on whether or the resulting keys are used as spend or view keys.


moneromooo posted a comment.	Jan 11th (6 days ago)
Checking key validity could certainly be done, yes. For the purposes of H1 though, we're after exploits of some kind so I'll focus on that. Can you point to the exploits that are not "I purposefully select a known key, and someone can then guess it".


anonimal posted a comment.	Jan 11th (6 days ago)
The resulting wallet will be able to function normally on the network.

@who-biz, if I'm reading this report correctly, are you suggesting an attack similar to transaction flooding but, in our case, the network is flooded with these trivially discernible keys?

equiprobable

I imagine this attack would break the cover of equiprobability not only because the distribution of legitimate inputs is poisoned (equal distribution of bad data is as equally bad as unequal distribution of good data) but because the keys are so easily seen as compromised.

even a very small bias apparently has a disproportionate effect

Yes, reducing bias is good, but banning can also be good: if poisoned keys are banned, legitimate txs will have an ever shrinking pool of observable inputs to choose from. On the other hand, to knowingly mixin poisoned inputs is... well.. I'm not sure what the technical word for that is :)

I'm not condoning banning, just pointing out the trade-off/compromise.

Sidenote: I can see a tagging attack where the fear of mixing in obviously compromised keys could create a herding-effect among nodes. Network topologies could be mapped based on key droppage - but I believe this is not specific to null keys and is actually more specific to nodes across anonymous overlay networks.


who-biz posted a comment.	Jan 11th (6 days ago)
Sorry, no. Not what I’m referring to. Hold on, I’ve just gotten home and have some proof of concept code I can paste from my work on rewriting DPoW for ringCT.


who-biz posted a comment.	Updated Jan 11th (6 days ago)
std::vector<crypto::key_derivation> recv_derivations;
    std::vector<std::pair<crypto::public_key,size_t>> recv_outkeys;
    int i = -1;
    for (int j = 0; j < count; j++) {
      i = signers_index[count-1];
      crypto::secret_key viewkey = notary_viewkeys[i];
      cryptonote::transaction tx = pen_tx.tx;
      //crypto::public_key real_out_tx_key = get_tx_pub_key_from_extra(tx, 0);

      rct::rctSig &rv = tx.rct_signatures;
      if (rv.outPk.size() != tx.vout.size())
      {
        LOG_PRINT_L1("Failed to parse transaction from blob, bad outPk size in tx " << get_transaction_hash(tx));
        return false;
      }
      for (size_t n = 0; n < tx.rct_signatures.outPk.size(); ++n)
      {
        if (tx.vout[n].target.type() != typeid(cryptonote::txout_to_key))
        {
          LOG_PRINT_L1("Unsupported output type in tx " << get_transaction_hash(tx));
          return false;
        }
        rv.outPk[n].dest = rct::pk2rct(boost::get<cryptonote::txout_to_key>(tx.vout[n].target).key);
        std::pair<crypto::public_key,size_t> each = std::make_pair(reinterpret_cast<const crypto::public_key&>(rv.outPk[n].dest), n);
        recv_outkeys.push_back(each);
      }

    size_t pk_index = 0;
    crypto::public_key recv_tx_key = get_tx_pub_key_from_extra(tx, pk_index++);
    for (const auto& each : notary_viewkeys) {
      crypto::key_derivation recv_derivation;
      bool R = generate_key_derivation(recv_tx_key, each, recv_derivation);
      if (!R) {
        MERROR("Failed to generate recv_derivation at append_ntz_sig! recv_tx_key = " << recv_tx_key << ", notary_viewkey = " << epee::string_tools::pod_to_hex(each));
        return false;
      } else {
        MWARNING("Recv derivation = " << recv_derivation);
        if (epee::string_tools::pod_to_hex(recv_derivation) != "0100000000000000000000000000000000000000000000000000000000000000") {
          MERROR("Recv derivation does not equal rct identity element! Validation failed, trying next pk_index...!");
          recv_tx_key = get_tx_pub_key_from_extra(tx, pk_index);
          bool R_two = generate_key_derivation(recv_tx_key, each, recv_derivation);
          if (!R_two) {
            MERROR("Validation failed at first two pk_indexes! (zero and one)");
            if (!m_wallet->remove_ntzpool_tx(epee::string_tools::pod_to_hex(get_transaction_hash(tx)), ptx_to_string_hash(pen_tx).second)) {
              MERROR("Error when removing ntzpool tx:" << epee::string_tools::pod_to_hex(get_transaction_hash(tx)) << "!");
              return false;
            }
          }
        } else {
          recv_derivations.push_back(recv_derivation);
        }
The problematic line occurs in the generate_key_derivation(recv_tx_key, each, recv_derivation)

In the mechanism I've written, these pubkeys are taken from a table located here: https://github.com/blur-network/dpow-blur/blob/test-wallets/src/cryptonote_core/komodo_notaries.h

Obviously, these pubkeys are known... But the implications are somewhat broader for anyone who knows a public key of a destination, potentially.

Those values for the subscript in signers_index simply correspond to a given line in the notaries_elected table... i.e. signers_index[12] corresponds to the given pubkeys in line 12 of notaries_elected

count simply corresponds to the amount of signatures on a multisig transaction, thus far in the multisignature process.


moneromooo posted a comment.	Jan 12th (5 days ago)
This proof of concept proves what exactly ? Can you point to somewhere in monero which causes a consensus failure or other security bug which is not "someone can guess I used that particular key" ?


who-biz posted a comment.	Jan 12th (5 days ago)
I believe the equivalent line of code where I would imagine an issue to arise would be here: https://github.com/monero-project/monero/blob/master/src/cryptonote_basic/cryptonote_format_utils.cpp#L274


who-biz posted a comment.	Jan 12th (5 days ago)
Note also, that the procedure followed in that PoC code is the same that is employed in Monero’s msout cases of transaction creation/expansion.


moneromooo posted a comment.	Jan 12th (5 days ago)
Just in case, I double checked with a MRL member: someone using a zero key (or other invalid key) does not cause breakdown in the crypto (though that's a preliminary opinion since they only had half an hour to think about it).
So if you think there is, you need to explain what is the exploit. 
Again, if you use a zero key and end up with lost money and/or lost privacy, it's your own fault and doesn't qualify as an exploit (unless the monero software generates these due to a bug like what you hinted at in your first post with the init and no subsequent init).


who-biz posted a comment.	Jan 12th (5 days ago)
We aren’t talking about zero keys here. They’re keys equal to the rct::identity element.


who-biz posted a comment.	Updated Jan 12th (5 days ago)
The zero values are a recovery seed. That zero value recovery seed generates keys equal to rct::identity

Zero value keys might not result in a breakdown in the crypto... but judging by how they cause a breakdown in the key derivation... I’d say that might be an issue too. Regardless, the derived value (pubkey = rct::identity) is the issue... Not the zero value input.


moneromooo posted a comment.	Jan 13th (4 days ago)
Don't be that way. A zero secret key yields an identity public key. Your first post even has a zero key in the sample json. I'm waiting for you to tell us what the exploit is you've found.


who-biz posted a comment.	Jan 13th (4 days ago)
Is a zero value recovery seed yielding a pubkey equal to rct::identityan intended outcome?


who-biz posted a comment.	Jan 13th (4 days ago)
Sorry, I misspoke. The generated public key is indeed a zero value key. The private key generated is equal to rct::identity. Regardless, my question above stands about this being an intended outcome.


who-biz posted a comment.	Jan 13th (4 days ago)
Additionally, as per your VRP, you are responsible for explaining why this is “not a vulnerability”. I am disclosing a vulnerability, not engineering it into an exploit.


moneromooo posted a comment.	Jan 14th (3 days ago)
Zero is an invalid key. zero times G should be 0, you're right. I'm not playing word games. AFAIK there is no vuln. If you think there is one, say what you think it is, not just "if you have zero, I think there might be a vuln which I'm not pointing to". Given your history here, I do not give you the benefit of the doubt anymore, and I'm not wasting my time trying to guess what you're referring to. Be precise.


who-biz posted a comment.	Jan 14th (3 days ago)
There is no benefit of the doubt needing to be given. There is no guesswork needed, and I’m not playing semantics... but you want me to be precise, and talk about history? Okay, we can do both. Give me a bit.


who-biz posted a comment.	Jan 15th (3 days ago)
Just kidding. Let's stay on topic. Here's a patch:

---
 src/crypto/crypto.cpp | 6 ++++--
 1 file changed, 4 insertions(+), 2 deletions(-)

diff --git a/src/crypto/crypto.cpp b/src/crypto/crypto.cpp
index 0ec992de..6cee9c7f 100644
--- a/src/crypto/crypto.cpp
+++ b/src/crypto/crypto.cpp
@@ -69,8 +69,10 @@ namespace crypto {
 #include "random.h"
   }

-  const crypto::public_key null_pkey = crypto::public_key{};
-  const crypto::secret_key null_skey = crypto::secret_key{};
+  unsigned char const np[32] = { 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254 };
+  unsigned char const ns[32] = { 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254 };
+  crypto::public_key const null_pkey = *reinterpret_cast<const public_key*>(np);
+  crypto::secret_key const null_skey = *reinterpret_cast<const secret_key*>(ns);

   static inline unsigned char *operator &(ec_point &point) {
     return &reinterpret_cast<unsigned char &>(point);

moneromooo posted a comment.	Jan 15th (3 days ago)
What I want is understand what vulnerability you claim to have found.
So far all I have is "if the user selects a zero secret key, there might be an exploit to do with generate_key_derivation failing". There might or might not be a vulnerability there, so please explain what it is.


who-biz posted a comment.	Updated Jan 16th (2 days ago)
Oof, upon a second check, that patch doesn’t fix the key issue. Maybe you’re right. I’ll look into multisig, and file a second report if I find something. For now, feel free to treat this as exhausted. We can go with anonimal’s proposed scenario, or the key guessing for now. Whichever you prefer.

As I’ve said before, I’m not here for a bounty. Let me know how you guys want to drum up a solution, if at all.


moneromooo posted a comment.	Jan 16th (about 1 day ago)
OK, I'll add some checking to the json input anyway, not a vuln but still a bug technically so something good came off this :)


who-biz posted a comment.	Updated Jan 17th (3 hrs ago)
So, with this being triaged... please advise on how to proceed (i.e. if disclosure window of 90 days still holds, or if I am allowed to disclose as is)


anonimal posted a comment.	Jan 17th (3 hrs ago)
@who-biz what exactly is there to disclose? @moneromooo has made a patch for the bug, but what's the vulnerability now?


who-biz posted a comment.	Updated Jan 17th (3 hrs ago)
The bug. I plan on letting others know about it, where affected.

If you don’t feel as though there is any vulnerability, nor need to disclose... that’s fine. But I don’t want to step on the toes of any disclosure you guys may have planned.


who-biz posted a comment.	Jan 17th (3 hrs ago)
I feel as though this is important, seeing as last January 4th, the same date as the “additional keys” vulnerability disclosure was locked by you guys, someone apparently attempted to test the same exploit on RYO’s chain.


who-biz posted a comment.	Jan 17th (2 hrs ago)
Maybe the date is wrong, but it was the same week as that was locked, and the “ledger bug” took place.


anonimal closed the report and changed the status to Informative.	Updated Jan 17th (2 hrs ago)
If you don’t feel as though there is any vulnerability, nor need to disclose... that’s fine. But I don’t want to step on the toes of any disclosure you guys may have planned.

Disclosures are reserved for proven vulnerabilities only. Thank you for pointing out the lack of filtering but it currently doesn't count as a proven vulnerability. I believe a pull request with the patch has been opened so feel free to share your findings with any affected projects.

As for locked reports, if any reports are locked it's because a reporter is behaving out of line or spamming the report. If the report is never disclosed, it's because it wasn't a proven vulnerability.


who-biz posted a comment.	Jan 17th (about 1 hr ago)
Sure. I feel as though some may disagree that there was no vulnerability there... if due to no other reason than the fact that the ledger bug (discovered the same DAY as that report being locked) occurred due to the exact same discrepancy (lack of accounting for additional keys)... but tomato tomahto, right?

Regardless, my concern is the live testing for proof of an exploit on other chains in production :)


who-biz posted a comment.	Updated Jan 17th (about 1 hr ago)
I will avoid disclosure until action is taken to approve this for disclosure, seeing as this has been changed to informative rather than triaged

who-biz requested to disclose this report.	Jan 17th (about 1 hr ago)

anonimal posted a comment.	Jan 17th (41 mins ago)
What is it exactly that you wish to disclose? Writeup a summary for me of:

Provably existing vulnerability that was not solved by moneromooo's patch
Provably existing solution
And take a long look at the many successful reports we have disclosed. They all A) provide a provable vulnerability B) provide a provable solution.


who-biz posted a comment.	Updated Jan 17th (26 mins ago)
I wish to disclose the “bug”.... worth noting that the ledger bug was identified as a bug and not an exploit... but was still disclosed with a formal write up. Im not suggesting a formal write up is appropriate or necessary here... but disclosing this on h1 would save me a lot of leg room, and I think more projects check here regularly, than I’d be able to reach manually.

I agree, though, that @moneromooo posting his pull request here as a patch would be beneficially “informative”


anonimal posted a comment.	Updated Jan 17th (14 mins ago)
Ok, I understand now. My apologies friend, but HackerOne reports are reserved for vulnerabilities only. Disclosing here would certainly save us both time but it would also be a disservice to the other researchers who worked to ensure that they have only reported proven vulnerabilities.

Disclosing here would also set a precedent of "every bug should be sent to HackerOne". In the future, you can always email or IRC us to confirm if a bug is a vulnerability and then open a report if it is in fact a vulnerability.

What I would suggest you do instead is: copy/paste any of your writing from this report into moneromooo's pull request and then reference that pull request into any other announcements you make. That you way you get credit and can inform other communities as well. Your research here can serve as a basis for your further research and should be noted in that pull request too.

I certainly want more research to be done, especially with my proposed attack, but it's simply not appropriate to make any claims without the appropriate evidence (even for me). I think you can agree with that.

Does all of that make sense?


who-biz posted a comment.  Jan 17th (2 mins ago)
Sure. I can do that. Thanks for the guidance.

who-biz:
Posting here as per @anonimal’s guidance

moneromooo-monero:
> Is this really a problem that needs a solution?

Not really, it's mostly a pedantic crossing of ts and dotting of is. But OTOH it's not like it adds a lot of code to check.

fart-master:
That was my suspicion. Not a vulnerability. Just a check in case a stupid person decides to try something stupid. But as you say, the check is trivial. Thanks to who-biz for yet another declaration of a vulnerability that doesn't exist!

who-biz:
I did it just for you @fart-master 💗

fart-master:
Thanks. I enjoy pull requests to patch fake vulnerabilities. But you keep trying man, one day you'll be able to explain something you found

who-biz:
A cartoon bear can hope 🐻

vtnerd:
This patch seems fine, but my eyes crossed while reading the comment thread.

who-biz:
Is there any reason this shouldn't be disallowed in `hwdev` instead?  It seems almost all functions in `src/cryptonote_basic/cryptonote_format_utils.h` are accessed through that library, anyway.  Preventing this wallet-side only, just seems like it doesn't do much.

fart-master:
The patch is fine, I just question it's usefulness. As was already stated it's a simple check and a small PR, but why check for a zero key and not any other easy to guess key.

who-biz:
As I’ve said before... the issue is not the fact that it’s “easy to guess”... it’s the fact that it causes the key derivation function to put out a value equivalent to `rct::identity`. I’m not sure what the fixation on guessing keys, is all about.

who-biz:
See here: https://github.com/blur-network/blur/commit/f640346130851292ce1ace8dd36f29b6cf7c8334... For the fix I’m talking about. 

Note that the above fix seems that it may interfere with “watch-only” wallets, as written. You are free to use any of the above, in your code, if you see fit.

fart-master:
Yes, but has already been stated many times to you, there is no vulnerability in that which anyone could discern. So what is your actual point. 

> Just in case, I double checked with a MRL member: someone using a zero key (or other invalid key) does not cause breakdown in the crypto

So what exactly is the problem? Sure, check for it. But there doesn't actually seem to be a problem in the code if this PR is not merged. There is no vulnerability here, so why check for this specific invalid key and not any other?

who-biz:
Are zero-keys an intended input, and `rct::identity` keys an equivalently intended output?

If not, this would seem a bug. I'm not making a claim of vulnerability.  That was up to @moneromooo-monero and @anonimal to discern.  They have already classified this as a non-vuln... So i'm really not sure what you're on about.

PR #6298, opened 2020-01-17
[randomx] Add missing randomx_vm_set_cache()

cohcho:


hyc:
This is for the case of trying to mine with insufficient RAM for a dataset?

cohcho:
Yes.

PR #6110, opened 2019-11-08
tests: fix HF12 chaingen

ph4r05:
Currently, the `chaingen.cpp` does not support generating testing block chain with the HF12, which uses RandomX. The problem is that the miner function requires blockchain object to get certain block hashes.

I've created a simple workaround which extracts the block and hardforks from the `events` list. If caller needs to build HF12 blocks, he should set events and network type for the blockchain object. 

It is initial implementation of a working fix, it is not optimal as the blockchain object is constructed with each new block (if HF>=12), but this is robust as there is no need to maintain blockchain db consistency with the tests, it is just created from events when needed.

What do you think?

PR #6300, opened 2020-01-19
fix tests bug added in #6110

ph4r05:
- e.g., fixes gen_block_big_major_version test, error: generation failed: what=events not set, cannot compute valid RandomX PoW
- ask for events only if difficulty > 1 (when it really matters)
- throwing an exception changed to logging, so it is easy to spot a problem if tests start to fail. Throwing an exception could be too harsh for some tests, it may be a desired situation.

### DIFF
--- a/tests/core_tests/chaingen.cpp
+++ b/tests/core_tests/chaingen.cpp
@@ -393,11 +393,17 @@ void test_generator::fill_nonce(cryptonote::block& blk, const difficulty_type& d
   const cryptonote::Blockchain *blockchain = nullptr;
   std::unique_ptr<cryptonote::Blockchain> bc;
 
-  if (blk.major_version >= RX_BLOCK_VERSION)
+  if (blk.major_version >= RX_BLOCK_VERSION && diffic > 1)
   {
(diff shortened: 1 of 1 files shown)


ph4r05:
@selsta @moneromooo-monero

moneromooo-monero:
That looks like "ignore the problem", no ?

ph4r05:
depends, if `difficulty == 1` then randomx will work anyway. It seems all the core tests have difficulty 1 so it was passing before. 

I went for minimal changes to fix previously working tests.

moneromooo-monero:
So for those it reverts to what it was doing before. Which is OK by me, but defeats the point of your first patch.

ph4r05:
IMO the #6110 mainly enables a possibility to support `difficulty > 1` when needed in tests, as it is in cases like Trezor integration tests, where the difficulty goes up. So some tests may require #6110 to work.

To fix it globally (always have difficulty > 1) I would have to dig through all the core tests to figure out why difficulty is always 1 and PoW is actually not checked. Thats maybe for someone who knows those tests better. 

I could take a look on it, but it would take a bit more time than this PR which fixes at least current errors.

moneromooo-monero:
Alright.

PR #6301, opened 2020-01-21
simplewallet: warn about correlations if print-ring-members is not set

moneromooo-monero:
The warning about spending more than one output with similar creation
time was skipped if print-ring-members was not set, and it defaults to
false, which means most people probably aren't getting this warning if
they spend correlated outputs.

Reported by SeventhAlpaca.


Snipa22:
@moneromooo-monero Merge conflict, if you could rebase please. :)

moneromooo-monero:
done

PR #6302, opened 2020-01-22
wallet2_api: implement estimateTransactionFee

xiphon:


src/wallet/api/wallet.cpp
```diff
@@ -1671,6 +1671,26 @@ void WalletImpl::disposeTransaction(PendingTransaction *t)
     delete t;
 }
 
+uint64_t WalletImpl::estimateTransactionFee(const std::vector<std::pair<std::string, uint64_t>> &destinations,
+                                            PendingTransaction::Priority priority) const
+{
+    const size_t pubkey_size = 34;
+    const size_t encrypted_paymentid_size = 10;
+    const size_t extra_size = pubkey_size + encrypted_paymentid_size;
```
UkoeHB: I believe pubkey size is 33 (32 bytes key + 1 byte tag + 0 bytes length), and encrypted payment id size is 11 (8 bytes encrypted ID + 1 byte extra nonce tag + 1 byte extra nonce length + 1 byte payment ID extra nonce tag).

Gingeropolous:
> The wallet2 method could have just called the existing to avoid having to plump up the diff, but it's small enough.

wait, does that mean if that the method has to be changed in the future, it now has to be changed twice?

xiphon:
> it now has to be changed twice?

no

src/wallet/api/wallet.cpp
xiphon: Updated

PR #6307, opened 2020-01-27
Refresh language files

erciccione:
```
    Found 1148 source text(s) (46 new and 1102 already existing)
    Removed 9 obsolete entries
```


erciccione:
I'm closing this. Instead i will push some new strings from weblate and do the update at the same time.

PR #6097, opened 2019-11-04
wallet2: fix pool txes not being flushed when mined

moneromooo-monero:

PR #6310, opened 2020-01-28
Wallet: Fix unused variable compiler warning

tmoravec:
g++ 7.4.0 on Ubuntu 18.04

Alternatively, if we prefer to keep the variable around, we can silence it with something like 

```
   bool refreshed = false;
+  (void)refreshed;
```


moneromooo-monero:
I'd rather this stays in for now because I still have to investigate what value it should have.

tmoravec:
> I'd rather this stays in for now because I still have to investigate what value it should have.

Ok, the other option works as well. Updated the commit.

moneromooo-monero:
Since this also removes the warning, I don't see how it's better :)

selsta:
Removed here: https://github.com/monero-project/monero/pull/6097#issuecomment-549979567

selsta:
Superseded by #8359

PR #6314, opened 2020-01-31
Adding Dandelion++ support to public networks:

vtnerd:
- New flag in NOTIFY_NEW_TRANSACTION to indicate stem mode
  - Stem loops detected in tx_pool.cpp
  - Embargo timeout for a blackhole attack during stem phase

### DIFF
--- a/src/blockchain_db/blockchain_db.h
+++ b/src/blockchain_db/blockchain_db.h
@@ -160,7 +160,7 @@ struct txpool_tx_meta_t
   uint64_t max_used_block_height;
   uint64_t last_failed_height;
   uint64_t receive_time;
-  uint64_t last_relayed_time;
+  uint64_t last_relayed_time; //!< If Dandelion++ stem, randomized embargo timestamp. Otherwise, last relayed timestmap.
   // 112 bytes
   uint8_t kept_by_block;
   uint8_t relayed;
@@ -168,12 +168,16 @@ struct txpool_tx_meta_t
   uint8_t double_spend_seen: 1;
   uint8_t pruned: 1;
   uint8_t is_local: 1;
-  uint8_t bf_padding: 5;
+  uint8_t dandelionpp_stem : 1;
+  uint8_t bf_padding: 4;
 
   uint8_t padding[76]; // till 192 bytes
 
   void set_relay_method(relay_method method) noexcept;
   relay_method get_relay_method() const noexcept;
+
+  //! \return True if `get_relay_method()` now returns `method`.
+  bool upgrade_relay_method(relay_method method) noexcept;
 
   //! See `relay_category` description
   bool matches(const relay_category category) const noexcept
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1284,6 +1284,7 @@ namespace cryptonote
             break;
           case relay_method::block:
           case relay_method::fluff:
+          case relay_method::stem:
             public_req.txs.push_back(std::move(std::get<1>(tx)));
             break;
         }
@@ -1295,9 +1296,9 @@ namespace cryptonote
          re-relaying public and private _should_ be acceptable here. */
       const boost::uuids::uuid source = boost::uuids::nil_uuid();
       if (!public_req.txs.empty())
-        get_protocol()->relay_transactions(public_req, source, epee::net_utils::zone::public_);
+        get_protocol()->relay_transactions(public_req, source, epee::net_utils::zone::public_, relay_method::fluff);
       if (!private_req.txs.empty())
-        get_protocol()->relay_transactions(private_req, source, epee::net_utils::zone::invalid);
+        get_protocol()->relay_transactions(private_req, source, epee::net_utils::zone::invalid, relay_method::local);
     }
     return true;
   }
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -46,6 +46,7 @@
 #include "warnings.h"
 #include "common/perf_timer.h"
 #include "crypto/hash.h"
+#include "crypto/duration.h"
 
 #undef MONERO_DEFAULT_LOG_CATEGORY
 #define MONERO_DEFAULT_LOG_CATEGORY "txpool"
@@ -58,6 +59,29 @@ namespace cryptonote
 {
   namespace
   {
+    /*! The Dandelion++ has formula for calculating the average embargo timeout:
+                          (-k*(k-1)*hop)/(2*log(1-ep))
+        where k is the number of hops before this node and ep is the probability
+        that one of the k hops hits their embargo timer, and hop is the average
+        time taken between hops. So decreasing ep will make it more probable
+        that "this" node is the first to expire the embargo timer. Increasing k
+        will increase the number of nodes that will be "hidden" as a prior
+        recipient of the tx.
+
+        As example, k=5 and ep=0.1 means "this" embargo timer has a 90%
+        probability of being the first to expire amongst 5 nodes that saw the
+        tx before "this" one. These values are independent to the fluff
+        probability, but setting a low k with a low p (fluff probability) is
+        not ideal since a blackhole is more likely to reveal earlier nodes in
+        the chain.
+
+        This value was calculated with k=10, ep=0.10, and hop = 175 ms. A
+        testrun from a recent Intel laptop took ~80ms to
+        receive+parse+proces+send transaction. At least 50ms will be added to
+        the latency if crossing an ocean. So 175ms is the fudge factor for
+        a single hop with 173s being the embargo timer. */
+    constexpr const std::chrono::seconds dandelionpp_embargo_average{CRYPTONOTE_DANDELIONPP_EMBARGO_AVERAGE};
+
     //TODO: constants such as these should at least be in the header,
     //      but probably somewhere more accessible to the rest of the
     //      codebase.  As it stands, it is at best nontrivial to test
@@ -262,34 +286,51 @@ namespace cryptonote
       }
     }else
     {
-      //update transactions container
-      meta.weight = tx_weight;
-      meta.fee = fee;
-      meta.max_used_block_id = max_used_block_id;
-      meta.max_used_block_height = max_used_block_height;
-      meta.last_failed_height = 0;
-      meta.last_failed_id = null_hash;
-      meta.receive_time = receive_time;
-      meta.last_relayed_time = time(NULL);
-      meta.relayed = relayed;
-      meta.set_relay_method(tx_relay);
-      meta.double_spend_seen = false;
-      meta.pruned = tx.pruned;
-      meta.bf_padding = 0;
-      memset(meta.padding, 0, sizeof(meta.padding));
-
       try
       {
         if (kept_by_block)
           m_parsed_tx_cache.insert(std::make_pair(id, tx));
         CRITICAL_REGION_LOCAL1(m_blockchain);
         LockedTXN lock(m_blockchain.get_db());
-        m_blockchain.remove_txpool_tx(id);
-        if (!insert_key_images(tx, id, tx_relay))
-          return false;
 
-        m_blockchain.add_txpool_tx(id, blob, meta);
-        m_txs_by_fee_and_receive_time.emplace(std::pair<double, std::time_t>(fee / (double)(tx_weight ? tx_weight : 1), receive_time), id);
+        const bool existing_tx = m_blockchain.get_txpool_tx_meta(id, meta);
+        if (existing_tx)
+        {
+          /* If Dandelion++ loop. Do not use txes in the `local` state in the
+             loop detection - txes in that state should be outgoing over i2p/tor
+             then routed back via public dandelion++ stem. Pretend to be
+             another stem node in that situation, a loop over the public
+             network hasn't been hit yet. */
+          if (tx_relay == relay_method::stem && meta.dandelionpp_stem)
+            tx_relay = relay_method::fluff;
+        }
+        else
+          meta.set_relay_method(relay_method::none);
+
+        if (meta.upgrade_relay_method(tx_relay) || !existing_tx) // synchronize with embargo timer or stem/fluff out-of-order messages
+        {
+          //update transactions container
+          meta.last_relayed_time = std::numeric_limits<decltype(meta.last_relayed_time)>::max();
+          meta.receive_time = receive_time;
+          meta.weight = tx_weight;
+          meta.fee = fee;
+          meta.max_used_block_id = max_used_block_id;
+          meta.max_used_block_height = max_used_block_height;
+          meta.last_failed_height = 0;
+          meta.last_failed_id = null_hash;
+          meta.relayed = relayed;
+          meta.double_spend_seen = false;
+          meta.pruned = tx.pruned;
+          meta.bf_padding = 0;
+          memset(meta.padding, 0, sizeof(meta.padding));
+
+          if (!insert_key_images(tx, id, tx_relay))
+            return false;
+
+          m_blockchain.remove_txpool_tx(id);
+          m_blockchain.add_txpool_tx(id, blob, meta);
+          m_txs_by_fee_and_receive_time.emplace(std::pair<double, std::time_t>(fee / (double)(tx_weight ? tx_weight : 1), receive_time), id);
+        }
         lock.commit();
       }
       catch (const std::exception &e)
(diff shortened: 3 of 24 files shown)


src/blockchain_db/blockchain_db.h
```diff
@@ -160,21 +160,25 @@ struct txpool_tx_meta_t
   uint64_t max_used_block_height;
   uint64_t last_failed_height;
   uint64_t receive_time;
-  uint64_t last_relayed_time;
+  uint64_t last_relayed_time; //!< If Dandelion++ stem, randomized embargo timestamp. Otherwise, last relayed timestmap.
```
vtnerd: Reviewers may want to take notice: I hijacked `last_relayed_time` in mempool to mean "embargo timeout timestamp" when the tx is marked as Dandelion++ stem. There's some padding space that could be used instead, if preferred.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -905,29 +905,60 @@ namespace cryptonote
...
-      m_core.handle_incoming_tx({arg.txs[i], crypto::null_hash}, tvc, relay_method::fluff, true);
-      if(tvc.m_verifivation_failed)
+      tx_relay = relay_method::fluff;
+      stem_txs.reserve(arg.txs.size());
+    }
+    else
+    {
+      tx_relay = relay_method::stem;
+      fluff_txs.reserve(arg.txs.size());
+    }
```
moneromooo-monero: Are the reserved vectors swapped w.r.t arg.dandelionpp_fluff ?

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -698,9 +776,31 @@ namespace levin
...
     {
-      zone_->strand.dispatch(fluff_notify{zone_, std::move(txs), source});
+      switch (tx_relay)
+      {
+        default:
+        case relay_method::none:
+        case relay_method::block:
+          return false;
+        case relay_method::stem:
+	  tx_relay = relay_method::fluff; // don't set stempool embargo when skipping to fluff
```
moneromooo-monero: Why does this do so without checking the embargo timer ?

src/cryptonote_core/tx_pool.cpp
```diff
@@ -280,34 +304,51 @@ namespace cryptonote
...
+          /* If Dandelion++ loop. Do not use txes in the `local` state in the
+             loop detection - txes in that state should be outgoing over i2p/tor
+             then routed back via public dandelion++ stem. Pretend to be
+             another stem node in that situation, a loop over the public
+             network hasn't been hit yet. */
+          if (tx_relay == relay_method::stem && meta.dandelionpp_stem)
+            tx_relay = relay_method::fluff;
+        }
+	else
+	  meta.set_relay_method(relay_method::none);
```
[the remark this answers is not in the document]
vtnerd: Yikes, a tab probably slipped in here. Will fix.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
vtnerd: Nope, I botched this. Initially the field was called `dandelionpp_stem`, but I flipped the logic so that zero initialization of the `struct` would default to stem mode. And since this can't be easily observed through testing ... anyway will change.

src/cryptonote_protocol/levin_notify.cpp
vtnerd: Looks like another tab snuck in here too or something. Will fix that.

This might require a bigger comment to explain. This is purely defensive programming. If someone requests stem mode over i2p/tor, the request cannot be done correctly because `on_transactions_relayed` (line 799) will put the tx in the stempool, making it eligible for broadcasting over ipv4/6 after an embargo timeout. So this was a quick hack so that "if I2P/Tor zone without white noise":

|  Requested Mode  |  New TxPool State  |
|-------------------------|--------------------------|
|           local             |            local             |
|           stem            |            fluff             |
|            fluff              |             fluff             |

Perhaps a `return false;` is better in this situation, indicating it can't work?

Also, if ipv4/6 local and stem requests map to Dandelion++ mode.

moneromooo-monero: Why would it be a problem to place an incoming tx in stem mode from i2p/tor in the txpool and fluff it after the embargo if not seen again ?

src/cryptonote_core/tx_pool.cpp
```diff
@@ -280,34 +304,51 @@ namespace cryptonote
...
+        const bool existing_tx = m_blockchain.get_txpool_tx_meta(id, meta);
+        if (existing_tx)
+        {
+          /* If Dandelion++ loop. Do not use txes in the `local` state in the
+             loop detection - txes in that state should be outgoing over i2p/tor
+             then routed back via public dandelion++ stem. Pretend to be
+             another stem node in that situation, a loop over the public
+             network hasn't been hit yet. */
+          if (tx_relay == relay_method::stem && meta.dandelionpp_stem)
+            tx_relay = relay_method::fluff;
```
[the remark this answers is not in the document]
vtnerd: Yes, you understood this correctly. The Dandelion++ states as footnote when describing their analysis implementation: "Our implementation enters fluff mode if there is a loop in the stem". Grin behavior is identical. Unfortunately the paper isn't specific on what to do when this occurs (its only mentioned in the footnote), but when calculating the probabilities, one of the scenarios was "the stem loops back to v, which terminates the stem". So I _think_ their analysis already takes this into account.

 The detection mode here is the most simple: if we sent as a Dandelion++ stem once, and its being re-added as a Dandelion++ stem, then mark it as a loop.

Some thoughts I have:
  * Record incoming connection uuid, and only claim loop if the recorded uuid differs? This would allow for re-attempting a stem - since each node has an independent epoch, the path may not be identical on the second attempt and might work the second time.
  * This may not be the most appropriate place (from a purely architecture view) to detect and alter stem loop behavior, but it is the easiest.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1299,9 +1300,9 @@ namespace cryptonote
...
       const boost::uuids::uuid source = boost::uuids::nil_uuid();
       if (!public_req.txs.empty())
-        get_protocol()->relay_transactions(public_req, source, epee::net_utils::zone::public_);
+        get_protocol()->relay_transactions(public_req, source, epee::net_utils::zone::public_, relay_method::fluff);
       if (!private_req.txs.empty())
-        get_protocol()->relay_transactions(private_req, source, epee::net_utils::zone::invalid);
+        get_protocol()->relay_transactions(private_req, source, epee::net_utils::zone::invalid, relay_method::local);
     }
     return true;
   }
```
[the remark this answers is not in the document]
vtnerd: The protocol says to relay in fluff mode when the embargo timer expires. I don't see any difficulties with mixing other older fluff txes, but maybe I need to think about it more. I **do** need to add a "sort" before sending fluff txes to hide order received and stem embargo vs fluff. I want to do this in another PR so that it can be discussed separately.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -280,34 +304,51 @@ namespace cryptonote
...
+        }
+	else
+	  meta.set_relay_method(relay_method::none);
+
+        if (meta.upgrade_relay_method(tx_relay) || !existing_tx) // synchronize with embargo timer or stem/fluff out-of-order messages
+        {
+          //update transactions container
+          meta.last_relayed_time = std::numeric_limits<decltype(meta.last_relayed_time)>::max();
```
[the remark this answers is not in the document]
vtnerd: If sending back to the prior node in the stem, the stem loop detection will take precedence and immediately fluff. The `last_relayed_time` value is only overwritten if the relay_method change goes up in the chain of none->local->stem->fluff->block. The value will be reset in `on_transactions_relayed`, assuming the transaction makes it to the relay logic code. The only time this is prevented is when `relay_method::none`, which is indicating this value is invalid anyway.

 I _probably_ need to a filter in `cryptonote_protocol_handler.inl` that drops transactions in the stem phase on outgoing connections (only incoming connections should be sending stem transactions). This strengthens the protocol a bit (not sure of any good attacks this prevents though).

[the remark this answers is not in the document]
vtnerd: Also, as per your attack, honest nodes should behave identically to the originator. So nothing should be gained by sending back. If this is incorrect somewhere, then it definitely needs to be fixed.

src/cryptonote_core/cryptonote_core.cpp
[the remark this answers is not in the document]
moneromooo-monero: I do not see any test for that embargo timer though.

src/cryptonote_core/tx_pool.cpp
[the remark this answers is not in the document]
moneromooo-monero: If it's in the original paper then I guess it's fine. And thinking about it, it's a difference between "Eve can force Alice to switch to fluff early" and "Eve can either sit on the tx and delay the tx, or fluff herself", so it doesn't seem too annoying.

src/cryptonote_core/cryptonote_core.cpp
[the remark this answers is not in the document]
vtnerd: The embargo timer is leveraging existing logic in `get_relayable_transactions`. The filtering logic in that function now checks for the `dandelionpp_stem` flag in tx_meta, checks the embargo timer, etc. Previously it was filtering on a longer, not randomized timeout.

The name of the function is poor, because it does give the impression that it will be returning all transactions in the pool not marked as `do_not_relay`.

src/cryptonote_protocol/levin_notify.cpp
vtnerd: A transaction received over i2p/tor is forwarded over public ipv4/6 in stem mode. The issue is the public interface allows a request to send transaction(s) over i2p/tor in stem mode, which should arguable be rejected (`return false`). Dandelion++ over i2p/tor requires mempool tracking the "zone" a transaction was received for the embargo timeout. I'm not aware of anything else blocking Dandelion++ over i2p/tor, but it might probably worth thinking about this mode separately. A sybil actor should already have difficulties when tracing origin IP over i2p/tor. However, Dandelion++ over i2p/tor can get interesting with white noise since it makes advanced ISP spying more difficult too.

moneromooo-monero: I don't understand, and I don't even know which part I don't understand here, but I'm happy enough given this is on purpose and not a mistake :)

rbrunner7:
A question: How to best test that "from the outside"? I.e. as someone that runs a Dandelion++ enabled node and wants to see from log output whether some Dandelion++ enabled transactions arrive and are either passed on or are finally accepted into the mempool and re-broadcast (if I understand the protocoll correctly)?

I looked over the code but did not find something like a new Dandelion++ related log category. Maybe something like that could be very useful? I imagine only a few log lines at strategic points would already help a great deal watching a daemon "doing its thing".

vtnerd:
There is log of every outgoing p2p message (`net.p2p.traffic`) which includes the recipient. Command `2002` refers to transaction notifications. A stem should have one send, and a fluff one for every connection - 1 (or every connection on re-send). Logs can be added that partially duplicate those other messages - it might clarify whats happening when multiple transactions are being processed within a short time window (the command # / recipient is already fairly telling).

erciccione:
@vtnerd i'm gonna make a reddit post later today asking people to build this PR and run a testnet node with it, to test dandelion. I see this branch is 58 commits behind upstream master, could you rebase?

Shouldn't be an issue anyway, but since people are going to run a testnet node with this branch, better having them running current master + this PR, if possible.

rbrunner7:
> There is log of every outgoing p2p message (`net.p2p.traffic`) which includes the recipient. Command `2002` refers to transaction notifications. 

I could not find `net.p2p.traffic` as a log category, but `net.p2p.msg` which indeed outputs a message `NOTIFY_NEW_TRANSACTIONS` which has a number of 2002.

On mainnet that leads to quite some amount of messages where it may be quite difficult to watch out for the communication patterns that you describe. On testnet most of the time nothing happens regarding transactions, so there it's probably feasible.

vtnerd:
rebased and added log statements in the `net.p2p.tx` category.

erciccione:
Call for testers: https://www.reddit.com/r/Monero/comments/f9cksh/help_us_test_dandelion_on_testnet_instructions/

(At the moment the post is waiting for manual approval, as it was erroneously considered a request for support by automod. Will be unlocked soon)

rbrunner7:
I just tried with the new code, and somehow I could not get it to work: Everything worked transaction-wise, i.e. transactions were pooled, broadcasted and mined, but I never saw any additional log message using the new category `net.p2p.tx`.

But my daemon now gives me the following line four times, whether the new log category is set or not:

     2020-02-25 19:45:39.607 E Failed to get tx meta from txpool

Any idea what I might do wrong, or what might be the problem here?

selsta:
gh-actions functional tests failed.

vtnerd:
> 2020-02-25 19:45:39.607 E Failed to get tx meta from txpool

This is "noise" that can be ignored for now - it occurs when the tx comes back to the node.

The problem is that the tx is removed from the DB, then key_images are inserted into a local map which does a DB lookup (which is new), and then the tx is added to the DB. This was always the call sequence, so I avoided changing it. But this error needs to be removed before a release - investigating what to do about it.

>  but I never saw any additional log message using the new category net.p2p.tx.

That's because I pushed the rebase, but not the log additions. Crap. The latest push should have the log statements.

erciccione:
Now seems to be working. After sending a transaction i get:

```
Sent 1 transaction(s) to 201f939a-2526-4a8c-8cdb-1652bd3bed98 using Dandelion++ stem
```
The *Failed to get tx meta from txpool* message for times and then

```
Queueing 1 transaction(s) for Dandelion++ fluffing
```

vtnerd:
A [comment on reddit](https://www.reddit.com/r/Monero/comments/fcvaom/dandelion_stops_bad_actors_from_linking_your_ip/fjftif8?utm_source=share&utm_medium=web2x) pointed to a discussion on Github/Bitcoin about DoSing with Dandelion.

One thing I think should be added is a rejection if an existing key-image with a new tx_hash is seen. Currently this is not rejected, because its treated as seeing the same key-image/tx_hash pair a second time (which it pretends it never saw for privacy reasons).

This would hurt privacy in the rare situation that a wallet tries to use the same unspent output in two different transactions before the transaction ages out of the mempool. However, this is already disallowed.

@moneromooo-monero ? You were asking about these key-image checks, in a prior review (the relevant code isn't actually changing in this diff). This is _probably_ worthy of a new PR since its already in the codebase.

vtnerd:
Rebased (there was no conflicts, but users attempting this weekends stress test will testing latest HEAD). Also, fixed the error log issue, see my new inline comment.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -280,34 +304,51 @@ namespace cryptonote
...
+          meta.relayed = relayed;
+          meta.double_spend_seen = false;
+          meta.pruned = tx.pruned;
+          meta.bf_padding = 0;
+          memset(meta.padding, 0, sizeof(meta.padding));
+
+          if (!insert_key_images(tx, id, tx_relay))
+            return false;
+
+          m_blockchain.remove_txpool_tx(id);
```
vtnerd: In the force push, this was moved here below the call to `insert_key_images` now on line 345.

Explanation:
Initially the call sequence was `remove_txpool_tx` -> `add_txpool_tx` -> `insert_key_images`. In PR 5965 I flipped `insert_key_images` -> `add_txpool_tx` because the `insert_key_images` was modified to make a DB check to get the "state" of the transaction. This was necessary because the key image check should not fail if the tx state is going from stem->fluff.

This meant that if the tx was going from stem -> fluff the tx had its metadata removed before the key images check. This was/is hard to detect via the `core_tests` because `cryptonote_core::core` will filter double spends before it even gets to this line. Txes going from stem->fluff are not filtered, which is why the error was logged.

vtnerd:
> gh-actions functional tests failed.

This tests now pass, with the exception of an unrelated windows macro  issue.

dEBRUYNE-1:
@vtnerd - Could you rebase?

vtnerd:
Rebased.

vtnerd:
Some core tests are now fail. ahhh hold on.

vtnerd:
Ok, now that should be a proper rebase.

selsta:
Functional tests failed.

vtnerd:
Rebased to fix conflicts. PR 6445 should hopefully fix functional tests. Or at least be closer. Enough changes were made, I decided to spin that into its own PR.

vtnerd:
Rebased again.

PR #6321, opened 2020-02-07
wallet2: mention --daemon-ssl-allow-any-cert in --proxy error msg

xiphon:


moneromooo-monero:
The condition triggers if there's no strong verification. Allowing any cert seems pretty weak.

xiphon:
> The condition triggers if there's no strong verification. Allowing any cert seems pretty weak.

Any cert option is already allowed in the current codebase. The commit just notifies a user that such option exists.

moneromooo-monero:
OK, that can be addressed in another patch, if it gets addressed at all.

PR #6322, opened 2020-02-07
device_ledger: include status code name in error message

xiphon:


src/device/device_ledger.cpp
```diff
@@ -64,6 +69,81 @@ namespace hw {
       crypto::secret_key dbg_spendkey;
     #endif
 
+    #define LEDGER_STATUS(status) {status, #status}
+    static const std::map<unsigned int, const char *> status_codes = {
```
vtnerd: If these codes were sorted, this can be a `constexpr` table with a `std::lower_bound` search over it. Unfortunately, `constexpr` constructors for `std::pair` aren't available in C++11. Only a `struct` and lambda usage is necessary, so the coding overhead is still minimal.

This would saves a bit of memory and startup cost, and eliminates the possibility of static initialization order issues. I haven't attempt to track down the latter, so its possible this is guaranteed to work on all scenarios (right now).

xiphon: Good idea, let's avoid static initialization with `constexpr` table. Updated.

PR #3027, opened 2017-12-29
Reserve proof

stoffu:
I was inspired by a reddit post and came up with a new kind of cryptographic proof that proves your partial reserve to an auditor (https://www.reddit.com/r/Monero/comments/7mkoi8/random_question/druqe05/), called *reserve proof*. The basic idea is to disclose only a fixed set of your unspent outputs with signed key images (and to **NOT** hand over your view secret key). This subset can be either the entirety of your available outputs across all subaddress accounts, or a set of outputs within the currently selected account that can cover the requested amount with the smallest set size.

Here's the command synopsis:
```
Command usage: 
  get_reserve_proof (all|<amount>) [<message>]

Command description: 
  Generate a signature proving that you own at least this much, optionally with a challenge string <message>.
  If 'all' is specified, you prove the entire sum of all of your existing accounts' balances.
  Otherwise, you prove the reserve of the smallest possible amount above <amount> available in your current account.


Command usage: 
  check_reserve_proof <address> <signature_file> [<message>]

Command description: 
  Check a signature proving that the owner of <address> holds at least this much, optionally with a challenge string <message>.
```

And here are some usage examples:
```
          Account               Balance      Unlocked balance                 Label
 *       0 9xLMUj       36.644383933000       36.644383933000       Primary account
         1 BgeJGK        8.897020120000        8.897020120000             checking1
         2 BeU2Yg        0.200000000000        0.200000000000             checking2
         3 BchbAG        0.300000000000        0.300000000000             checking3
         4 BfCBvC        0.400000000000        0.400000000000              savings1
         5 Bb4YGv        0.500000000000        0.500000000000              savings2
         6 BexYvg        0.600000000000        0.600000000000              savings3
----------------------------------------------------------------------------------
          Total       47.541404053000       47.541404053000
Currently selected account: [0] Primary account
Tag: (No tag assigned)
Balance: 36.644383933000, unlocked balance: 36.644383933000
Background refresh thread started

[wallet 9xLMUj]: get_reserve_proof all banana
signature file saved to: monero_reserve_proof

[wallet 9xLMUj]: check_reserve_proof 9xLMUjRpNfEQLoYwogkmiBg9H6XP18mJ3UJphkvhWo9SB3vwbJFSryNfVz6vJFigYwLadki17xHQG7qsxQnnZPuPC1XvpHq monero_reserve_proof banana
Good signature -- total: 47.541404053000, spent: 0.000000000000, unspent: 47.541404053000

[...]

PR #6329, opened 2020-02-10
Updated transaction proofs and tests

SarangNoether:
Updates InProofV1, OutProofV1, and ReserveProofV1 to new V2 variants that include all public proof parameters in Schnorr challenges, along with hash function domain separators. Includes new randomized unit tests.

While the unit tests include checks for cross-version proof generation and verification, this has not been specifically checked in functional tests, since there is no exposed way to easily generate a V1 proof from a transaction anymore.


UkoeHB:
`ReserveProofV1` was originally added in PR #3027. It includes proofs of knowledge of all private spend keys for the proof's unspent owned outputs (e.g. outputs can be owned by different subaddresses), which mean the signer actually owns those outputs (has spend-rights over them). However, it does not include proofs that any subaddresses correspond with the main address. Is that worth including, or superfluous? @stoffu

PR #6332, opened 2020-02-11
Allow wallet2.h to run in WebAssembly.

woodser:
- Add contrib/epee/include/net/abstract_http_client.h which
http_client.h extends.
- Replace `simple_http_client` with `abstract_http_client` in wallet2,
message_store, message_transporter, and node_rpc_proxy.
- Skip code incompatible with webassembly using `#if defined
__EMSCRIPTEN__` directives.
- Add ability to import and export wallet file buffers in wallet2
instead of using file system.

Code and collaboration credits: moneromooo, endogenic, mymonero

Feedback is requested on removing the call to `miner::find_nonce_for_given_block` in cryptonote_tx_utils.cpp and changing the default http method from `GET` to `POST` in wallet2.h and http_abstract_invoke.h (otherwise the browser issues `GET` requests which fail).

### DIFF
--- a/contrib/epee/include/net/abstract_http_client.h
+++ b/contrib/epee/include/net/abstract_http_client.h
@@ -1,0 +1,87 @@
+// Copyright (c) 2006-2013, Andrey N. Sabelnikov, www.sabelnikov.net
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+// * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+// * Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+// * Neither the name of the Andrey N. Sabelnikov nor the
+// names of its contributors may be used to endorse or promote products
+// derived from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+// ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+// WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER  BE LIABLE FOR ANY
+// DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+// (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+// ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <string>
+#include <boost/optional/optional.hpp>
+#include "http_auth.h"
+#include "net/net_ssl.h"
+
+namespace epee
+{
+namespace net_utils
+{
+  inline const char* get_hex_vals()
+  {
+    static constexpr const char hexVals[16] = {'0','1','2','3','4','5','6','7','8','9','A','B','C','D','E','F'};
+    return hexVals;
+  }
+
+  inline const char* get_unsave_chars()
+  {
+    //static constexpr char unsave_chars[] = "\"<>%\\^[]`+$,@:;/!#?=&";
+    static constexpr const char unsave_chars[] = "\"<>%\\^[]`+$,@:;!#&";
+    return unsave_chars;
+  }
+
+  bool is_unsafe(unsigned char compare_char);
+  std::string dec_to_hex(char num, int radix);
+  int get_index(const char *s, char c);
+  std::string hex_to_dec_2bytes(const char *s);
+  std::string convert(char val);
+  std::string conver_to_url_format(const std::string& uri);
+  std::string convert_from_url_format(const std::string& uri);
+  std::string convert_to_url_format_force_all(const std::string& uri);
+
+namespace http
+{
+  class abstract_http_client
+  {
+  public:
+    abstract_http_client() {}
+    virtual ~abstract_http_client() {}
+    bool set_server(const std::string& address, boost::optional<login> user, ssl_options_t ssl_options = ssl_support_t::e_ssl_support_autodetect);
+    virtual void set_server(std::string host, std::string port, boost::optional<login> user, ssl_options_t ssl_options = ssl_support_t::e_ssl_support_autodetect) = 0;
+    virtual void set_auto_connect(bool auto_connect) = 0;
+    virtual bool connect(std::chrono::milliseconds timeout) = 0;
+    virtual bool disconnect() = 0;
+    virtual bool is_connected(bool *ssl = NULL) = 0;
+    virtual bool invoke(const boost::string_ref uri, const boost::string_ref method, const std::string& body, std::chrono::milliseconds timeout, const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) = 0;
+    virtual bool invoke_get(const boost::string_ref uri, std::chrono::milliseconds timeout, const std::string& body = std::string(), const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) = 0;
+    virtual bool invoke_post(const boost::string_ref uri, const std::string& body, std::chrono::milliseconds timeout, const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) = 0;
+    virtual uint64_t get_bytes_sent() const = 0;
+    virtual uint64_t get_bytes_received() const = 0;
+  };
+
+  class http_client_factory
+  {
+  public:
+    virtual ~http_client_factory() {}
+    virtual std::unique_ptr<abstract_http_client> create() = 0;
+  };
+}
+}
+}
--- a/contrib/epee/include/net/http_client.h
+++ b/contrib/epee/include/net/http_client.h
@@ -47,6 +47,7 @@
 
 #include "string_tools.h"
 #include "reg_exp_definer.h"
+#include "abstract_http_client.h"
 #include "http_base.h" 
 #include "http_auth.h"
 #include "to_nonconst_iterator.h"
@@ -105,140 +106,11 @@ namespace net_utils
 
 
 	//---------------------------------------------------------------------------
-	static inline const char* get_hex_vals()
-	{
-		static const char hexVals[16] = {'0','1','2','3','4','5','6','7','8','9','A','B','C','D','E','F'};
-		return hexVals;
-	}
-
-	static inline const char* get_unsave_chars()
-	{
-		//static char unsave_chars[] = "\"<>%\\^[]`+$,@:;/!#?=&";
-		static const char unsave_chars[] = "\"<>%\\^[]`+$,@:;!#&";
-		return unsave_chars;
-	}
-
-	static inline bool is_unsafe(unsigned char compare_char)
-	{
-		if(compare_char <= 32 || compare_char >= 123)
-			return true;
-
-		const char* punsave = get_unsave_chars();
-
-		for(int ichar_pos = 0; 0!=punsave[ichar_pos] ;ichar_pos++)
-			if(compare_char == punsave[ichar_pos])
-				return true;
-
-		return false;
-	}
-	
-	static inline 
-		std::string dec_to_hex(char num, int radix)
-	{
-		int temp=0;
-		std::string csTmp;
-		int num_char;
-
-		num_char = (int) num;
-		if (num_char < 0)
-			num_char = 256 + num_char;
-
-		while (num_char >= radix)
-		{
-			temp = num_char % radix;
-			num_char = (int)floor((float)num_char / (float)radix);
-			csTmp = get_hex_vals()[temp];
-		}
-
-		csTmp += get_hex_vals()[num_char];
-
-		if(csTmp.size() < 2)
-		{
-			csTmp += '0';
-		}
-		
-		std::reverse(csTmp.begin(), csTmp.end());
-		//_mbsrev((unsigned char*)csTmp.data());
-
-		return csTmp;
-	}
-	static inline int get_index(const char *s, char c) { const char *ptr = (const char*)memchr(s, c, 16); return ptr ? ptr-s : -1; }
-	static inline 
-		std::string hex_to_dec_2bytes(const char *s)
-	{
-		const char *hex = get_hex_vals();
-		int i0 = get_index(hex, toupper(s[0]));
-		int i1 = get_index(hex, toupper(s[1]));
-		if (i0 < 0 || i1 < 0)
-			return std::string("%") + std::string(1, s[0]) + std::string(1, s[1]);
-		return std::string(1, i0 * 16 | i1);
-	}
-
-	static inline std::string convert(char val)
-	{
-		std::string csRet;
-		csRet += "%";
-		csRet += dec_to_hex(val, 16);
-		return  csRet;
-	}
-	static inline std::string conver_to_url_format(const std::string& uri)
-	{
-
-		std::string result;
-
-		for(size_t i = 0; i!= uri.size(); i++)
-		{
-			if(is_unsafe(uri[i]))
-				result += convert(uri[i]);
-			else
-				result += uri[i];
-
-		}
-
-		return result;
-	}
-	static inline std::string convert_from_url_format(const std::string& uri)
-	{
-
-		std::string result;
-
-		for(size_t i = 0; i!= uri.size(); i++)
-		{
-			if(uri[i] == '%' && i + 2 < uri.size())
-			{
-				result += hex_to_dec_2bytes(uri.c_str() + i + 1);
-				i += 2;
-			}
-			else
-				result += uri[i];
-
-		}
-
-		return result;
-	}
-
-  static inline std::string convert_to_url_format_force_all(const std::string& uri)
-  {
-
-    std::string result;
-
-    for(size_t i = 0; i!= uri.size(); i++)
-    {
-        result += convert(uri[i]);
-    }
-
-    return result;
-  }
-
-
-
-
-
 	namespace http
 	{
 
 		template<typename net_client_type>
-		class http_simple_client_template: public i_target_handler
+    class http_simple_client_template : public i_target_handler, public abstract_http_client
 		{
 		private:
 			enum reciev_machine_state
@@ -279,7 +151,7 @@ namespace net_utils
 
 		public:
 			explicit http_simple_client_template()
-				: i_target_handler()
+				: i_target_handler(), abstract_http_client()
 				, m_net_client()
 				, m_host_buff()
 				, m_port()
@@ -299,26 +171,19 @@ namespace net_utils
 			const std::string &get_host() const { return m_host_buff; };
 			const std::string &get_port() const { return m_port; };
 
-			bool set_server(const std::string& address, boost::optional<login> user, ssl_options_t ssl_options = ssl_support_t::e_ssl_support_autodetect)
-			{
-				http::url_content parsed{};
-				const bool r = parse_url(address, parsed);
-				CHECK_AND_ASSERT_MES(r, false, "failed to parse url: " << address);
-				set_server(std::move(parsed.host), std::to_string(parsed.port), std::move(user), std::move(ssl_options));
-				return true;
-			}
+			using abstract_http_client::set_server;
 
-			void set_server(std::string host, std::string port, boost::optional<login> user, ssl_options_t ssl_options = ssl_support_t::e_ssl_support_autodetect)
+			void set_server(std::string host, std::string port, boost::optional<login> user, ssl_options_t ssl_options = ssl_support_t::e_ssl_support_autodetect) override
 			{
 				CRITICAL_REGION_LOCAL(m_lock);
 				disconnect();
 				m_host_buff = std::move(host);
 				m_port = std::move(port);
-                                m_auth = user ? http_client_auth{std::move(*user)} : http_client_auth{};
+				m_auth = user ? http_client_auth{std::move(*user)} : http_client_auth{};
 				m_net_client.set_ssl(std::move(ssl_options));
 			}
 
-			void set_auto_connect(bool auto_connect)
+			void set_auto_connect(bool auto_connect) override
 			{
 				m_auto_connect = auto_connect;
 			}
@@ -330,25 +195,25 @@ namespace net_utils
 				m_net_client.set_connector(std::move(connector));
 			}
 
-      bool connect(std::chrono::milliseconds timeout)
+      bool connect(std::chrono::milliseconds timeout) override
       {
         CRITICAL_REGION_LOCAL(m_lock);
         return m_net_client.connect(m_host_buff, m_port, timeout);
       }
 			//---------------------------------------------------------------------------
-			bool disconnect()
+			bool disconnect() override
 			{
 				CRITICAL_REGION_LOCAL(m_lock);
 				return m_net_client.disconnect();
 			}
 			//---------------------------------------------------------------------------
-			bool is_connected(bool *ssl = NULL)
+			bool is_connected(bool *ssl = NULL) override
 			{
 				CRITICAL_REGION_LOCAL(m_lock);
 				return m_net_client.is_connected(ssl);
 			}
 			//---------------------------------------------------------------------------
-			virtual bool handle_target_data(std::string& piece_of_transfer)
+			virtual bool handle_target_data(std::string& piece_of_transfer) override
 			{
 				CRITICAL_REGION_LOCAL(m_lock);
 				m_response_info.m_body += piece_of_transfer;
@@ -361,15 +226,14 @@ namespace net_utils
         return true;
       }
 			//---------------------------------------------------------------------------
-			inline 
-				bool invoke_get(const boost::string_ref uri, std::chrono::milliseconds timeout, const std::string& body = std::string(), const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list())
+			inline bool invoke_get(const boost::string_ref uri, std::chrono::milliseconds timeout, const std::string& body = std::string(), const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) override
 			{
 					CRITICAL_REGION_LOCAL(m_lock);
 					return invoke(uri, "GET", body, timeout, ppresponse_info, additional_params);
 			}
 
 			//---------------------------------------------------------------------------
-			inline bool invoke(const boost::string_ref uri, const boost::string_ref method, const std::string& body, std::chrono::milliseconds timeout, const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list())
+			inline bool invoke(const boost::string_ref uri, const boost::string_ref method, const std::string& body, std::chrono::milliseconds timeout, const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) override
 			{
 				CRITICAL_REGION_LOCAL(m_lock);
 				if(!is_connected())
@@ -442,7 +306,7 @@ namespace net_utils
 				return false;
 			}
 			//---------------------------------------------------------------------------
-			inline bool invoke_post(const boost::string_ref uri, const std::string& body, std::chrono::milliseconds timeout, const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list())
+			inline bool invoke_post(const boost::string_ref uri, const std::string& body, std::chrono::milliseconds timeout, const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) override
 			{
 				CRITICAL_REGION_LOCAL(m_lock);
 				return invoke(uri, "POST", body, timeout, ppresponse_info, additional_params);
@@ -456,12 +320,12 @@ namespace net_utils
 				return handle_reciev(timeout);
 			}
 			//---------------------------------------------------------------------------
-			uint64_t get_bytes_sent() const
+			uint64_t get_bytes_sent() const override
 			{
 				return m_net_client.get_bytes_sent();
 			}
 			//---------------------------------------------------------------------------
-			uint64_t get_bytes_received() const
+			uint64_t get_bytes_received() const override
 			{
 				return m_net_client.get_bytes_received();
 			}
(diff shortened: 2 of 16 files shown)


selsta:
@woodser Functional tests failed.

woodser:
Anyone know why `test-ubuntu` is not passing or how I can find more detail about the error?
```
Start  5: functional_tests_rpc
5/21 Test  #5: functional_tests_rpc .............***Failed  433.23 sec

**Comment by selsta** [comment]
make
    cd build/release/tests/functional_tests
    ctest --output-on-failure

File: src/wallet/wallet2.cpp [hunk 1]
```diff
@@ -1124,7 +1124,8 @@ void wallet_device_callback::on_progress(const hw::device_progress& event)
     wallet->on_device_progress(event);
 }
 
-wallet2::wallet2(network_type nettype, uint64_t kdf_rounds, bool unattended):
+wallet2::wallet2(network_type nettype, uint64_t kdf_rounds, bool unattended, std::shared_ptr<epee::net_utils::http::abstract_http_client> http_client):
+  m_http_client(http_client),
```
xiphon:
```suggestion
  m_http_client(std::move(http_client)),
```

src/wallet/wallet2.h
```diff
@@ -1495,7 +1511,7 @@ namespace tools
     std::string m_wallet_file;
     std::string m_keys_file;
     std::string m_mms_file;
-    epee::net_utils::http::http_simple_client m_http_client;
+    std::shared_ptr<epee::net_utils::http::abstract_http_client> m_http_client;
```
xiphon:
```suggestion
    const std::shared_ptr<epee::net_utils::http::abstract_http_client> m_http_client;
```

src/wallet/wallet2.cpp
```diff
@@ -3657,6 +3658,30 @@ void wallet2::clear_soft(bool keep_key_images)
...
+  if (e) {
+      boost::filesystem::remove(tmp_file_name);
+      LOG_ERROR("failed to update wallet keys file " << keys_file_name);
+      return false;
+  }
+
+  return true;
+}
+//----------------------------------------------------------------------------------------------------
+bool wallet2::get_keys_file_data(const epee::wipeable_string& password, bool watch_only, wallet2::keys_file_data &keys_file_data)
```
xiphon:
```suggestion
boost::optional<keys_file_data> wallet2::get_keys_file_data(const epee::wipeable_string& password, bool watch_only)
```

src/wallet/wallet2.cpp
```diff
@@ -3916,17 +3924,25 @@ void wallet2::change_password(const std::string &filename, const epee::wipeable_
  * \param keys_file_name Name of wallet file
  * \param password       Password of wallet file
  */
-bool wallet2::load_keys(const std::string& keys_file_name, const epee::wipeable_string& password)
+bool wallet2::load_keys(const std::string& keys_file_name, const epee::wipeable_string& password, const std::string& keys_buf)
 {
+  // determine if loading from file system or string buffer
+  bool use_fs = keys_file_name != ".keys";
```
xiphon: Would be better to implement `load_keys` overload accepting a buffer as an argument.

woodser:
With the changes in this PR, the tests are failing with:

```
AssertionError: {u'jsonrpc': u'2.0', u'id': u'0', u'error': {u'message': u'no connection to daemon', u'code': -38}}
```

contrib/epee/src/mlocker.cpp
```diff
@@ -160,18 +181,30 @@ namespace epee
 
   size_t mlocker::get_num_locked_pages()
   {
+#if defined __EMSCRIPTEN__
+    return 0;
+#endif
```
moneromooo-monero: Why add all of these when you can just not define HAVE_MLOCK ? Is this somehow not working ?

contrib/epee/include/net/http_client.h
```diff
@@ -299,26 +171,22 @@ namespace net_utils
...
 
-			void set_server(std::string host, std::string port, boost::optional<login> user, ssl_options_t ssl_options = ssl_support_t::e_ssl_support_autodetect)
+			virtual void set_server(std::string host, std::string port, boost::optional<login> user, ssl_options_t ssl_options = ssl_support_t::e_ssl_support_autodetect) override
 			{
 				CRITICAL_REGION_LOCAL(m_lock);
 				disconnect();
 				m_host_buff = std::move(host);
 				m_port = std::move(port);
-                                m_auth = user ? http_client_auth{std::move(*user)} : http_client_auth{};
+        m_auth = user ? http_client_auth{std::move(*user)} : http_client_auth{};
```
moneromooo-monero: Why these ?

contrib/epee/include/storages/http_abstract_invoke.h
```diff
@@ -125,14 +125,14 @@ namespace epee
...
+    bool invoke_http_json_rpc(const boost::string_ref uri, std::string method_name, const t_request& out_struct, t_response& result_struct, t_transport& transport, std::chrono::milliseconds timeout = std::chrono::seconds(15), const boost::string_ref http_method = "POST", const std::string& req_id = "0")
     {
       epee::json_rpc::error error_struct;
       return invoke_http_json_rpc(uri, method_name, out_struct, result_struct, error_struct, transport, timeout, http_method, req_id);
     }
 
     template<class t_command, class t_transport>
-    bool invoke_http_json_rpc(const boost::string_ref uri, typename t_command::request& out_struct, typename t_command::response& result_struct, t_transport& transport, std::chrono::milliseconds timeout = std::chrono::seconds(15), const boost::string_ref http_method = "GET", const std::string& req_id = "0")
+    bool invoke_http_json_rpc(const boost::string_ref uri, typename t_command::request& out_struct, typename t_command::response& result_struct, t_transport& transport, std::chrono::milliseconds timeout = std::chrono::seconds(15), const boost::string_ref http_method = "POST", const std::string& req_id = "0")
```
moneromooo-monero: Can you explain why the switch from GET to POST ? It is probably a good idea in theory since they can have side effects, but are there actual reasons beyond this ?

src/cryptonote_core/cryptonote_tx_utils.cpp
```diff
@@ -663,9 +663,13 @@ namespace cryptonote
...
-    miner::find_nonce_for_given_block([](const cryptonote::block &b, uint64_t height, unsigned int threads, crypto::hash &hash){
-      return cryptonote::get_block_longhash(NULL, b, hash, height, threads);
-    }, bl, 1, 0);
+
+    // TODO: miner cannot run in emscripten.  genesis nonce is set from config in wallet2::generate_genesis().  is this call necessary at all?
+    #if !defined __EMSCRIPTEN__
+      miner::find_nonce_for_given_block([](const cryptonote::block &b, uint64_t height, unsigned int threads, crypto::hash &hash){
+        return cryptonote::get_block_longhash(NULL, b, hash, height, threads);
+      }, bl, 1, 0);
+    #endif
```
moneromooo-monero: return false

src/wallet/wallet2.cpp
```diff
@@ -1313,7 +1314,7 @@ bool wallet2::set_daemon(std::string daemon_address, boost::optional<epee::net_u
   }
 
   MINFO("setting daemon to " << get_daemon_address());
-  return m_http_client.set_server(get_daemon_address(), get_daemon_login(), std::move(ssl_options));
+  return m_http_client->set_server(get_daemon_address(), get_daemon_login(), std::move(ssl_options));
 }
```
moneromooo-monero: All this massive spam all over the file should be avoided. You can probably do so easily by making m_http_client a reference, initialized in the ctor.

src/wallet/wallet2.cpp
```diff
@@ -3657,6 +3662,30 @@ void wallet2::clear_soft(bool keep_key_images)
...
+  std::error_code e = tools::replace_file(tmp_file_name, keys_file_name);
+  lock_keys_file();
+
+  if (e) {
+    boost::filesystem::remove(tmp_file_name);
+    LOG_ERROR("failed to update wallet keys file " << keys_file_name);
+    return false;
+  }
+
+  return true;
```
moneromooo-monero: I've not read this yet, but why is the keys file code changed in this PR ?

contrib/epee/src/mlocker.cpp
woodser: It is somehow not working (i.e. something is calling one of the mlocker class functions).  I will look into it further.

contrib/epee/include/net/http_client.h
woodser: I have updated the indentation to be the same as the lines above it.

contrib/epee/include/storages/http_abstract_invoke.h
woodser: The browser issues `GET` requests which fail unless these default to `POST`.  I was not able to track how `GET` works at all or where `GET` is overriden with `POST` in monero-project, as the RPC requests require `POST`.

src/cryptonote_core/cryptonote_tx_utils.cpp
woodser: Can you clarify?  Do you mean:

```
  #if defined __EMSCRIPTEN__
    return false;
  #endif

File: src/wallet/wallet2.cpp (hunk 9 above)
Comment (woodser) [reply src/wallet/wallet2.cpp]: Making `m_http_client` a reference will require it to be initialized in the constructor.  Since `abstract_http_client` is abstract, the default value could be an instance of `http_simple_client`, but that cannot be initialized in emscripten, so I do not have a suitable default value in that environment and execution fails.

I have updated the PR to use a reference which compiles but does not execute in emscripten because initializing `http_simple_client` in that environment fails.

Please advise.  Thanks.

File: src/wallet/wallet2.cpp (hunk 10 above)
Comment (woodser) [reply src/wallet/wallet2.cpp]: The key file code is changed in order to directly import and export wallet data to and from wallet2.  Otherwise there is no way to save and re-open a wallet in WebAssembly because the file system is not accessible.

File: src/cryptonote_core/cryptonote_tx_utils.cpp (hunk 8 above)
Comment (moneromooo-monero) [reply src/cryptonote_core/cryptonote_tx_utils.cpp]: Yes. Otherwise the function returrns success with an invalid block.

File: src/wallet/wallet2.cpp (hunk 9 above)
Comment (moneromooo-monero) [reply src/wallet/wallet2.cpp]: Off the top of my head, you can add a simple object, like:

struct redir {
redir(http_client *c): C(c) {}
operator http_client &() { return *c; }
http_client *C;
};

Then replace http_client *m_http_client with redir m_http_client in wallet2.h.
That should work AFAICT. Add default ctor if needed, and catch NULL if someone uses it before set.

Comment (vtnerd) [reply src/wallet/wallet2.cpp]: Two thoughts:

Use a `std::unique_ptr` instead of a raw pointer to clearly designate ownership. Unless `std::shared_ptr` behavior was necessary (it doesn't appear so at a glance).

Adding `*` and `->` stuff isn't really that hard to read, and we're now adding code just to reduce total line changes ... just why?

File: contrib/epee/include/net/abstract_http_client.h [hunk 11]
```diff
@@ -0,0 +1,173 @@
...
+
+#undef MONERO_DEFAULT_LOG_CATEGORY
+#define MONERO_DEFAULT_LOG_CATEGORY "net.http"
+
+namespace epee
+{
+  namespace net_utils
+  {
+    //---------------------------------------------------------------------------
+    static inline const char* get_hex_vals()
```
vtnerd: Drop `static` from function and add `constexpr` to char array.

contrib/epee/include/net/abstract_http_client.h
```diff
@@ -0,0 +1,173 @@
...
+    {
+      static const char hexVals[16] = {'0','1','2','3','4','5','6','7','8','9','A','B','C','D','E','F'};
+      return hexVals;
+    }
+
+    static inline const char* get_unsave_chars()
```
vtnerd: Same thing for this function (-`static` +`constexpr`).

contrib/epee/include/net/abstract_http_client.h
```diff
@@ -0,0 +1,173 @@
...
+    {
+      //static char unsave_chars[] = "\"<>%\\^[]`+$,@:;/!#?=&";
+      static const char unsave_chars[] = "\"<>%\\^[]`+$,@:;!#&";
+      return unsave_chars;
+    }
+
+    static inline bool is_unsafe(unsigned char compare_char)
```
vtnerd: Take the opportunity to put every function from here down into a cpp. Drop `static inline` from every function signature.

contrib/epee/include/net/abstract_http_client.h
```diff
@@ -0,0 +1,173 @@
+#pragma once
+
+#include <boost/optional/optional.hpp>
+#include "http_base.h"
```
vtnerd: I think `http_base.h` and `net_parse_helpers.h` can be moved to the cpp (see other changes). `#incude <string>` is also needed (its being included indirectly somehow now).

contrib/epee/include/net/abstract_http_client.h
```diff
@@ -0,0 +1,173 @@
...
+    {
+      /**
+       * Abstract HTTP client interface.
+       */
+      class abstract_http_client
+      {
+      public:
+        abstract_http_client() {}
+        virtual ~abstract_http_client() {}
+        virtual bool set_server(const std::string& address, boost::optional<login> user, ssl_options_t ssl_options = ssl_support_t::e_ssl_support_autodetect)
```
vtnerd: Drop `virtual` from this signature, unless there is a need to overload the behavior (doesn't seem so). Also, move to cpp.

contrib/epee/include/net/abstract_http_client.h
```diff
@@ -0,0 +1,173 @@
+#pragma once
+
+#include <boost/optional/optional.hpp>
+#include "http_base.h"
+#include "http_auth.h"
+#include "net/net_ssl.h"
+#include "net_parse_helpers.h"
+
+#undef MONERO_DEFAULT_LOG_CATEGORY
+#define MONERO_DEFAULT_LOG_CATEGORY "net.http"
```
vtnerd: Move to cpp ... I think? @moneromooo-monero

contrib/epee/include/net/http_client.h
```diff
@@ -105,140 +106,11 @@ namespace net_utils
...
-
-
-
-
 	namespace http
 	{
 
 		template<typename net_client_type>
-		class http_simple_client_template: public i_target_handler
+    class http_simple_client_template: public i_target_handler, public abstract_http_client
```
vtnerd: Mark it as `class http_simple_client_template final : ...`. This prevents anyone from overriding this class, and allows the compiler to de-virtualize all `this->` virtual method calls within this class.

src/wallet/wallet2.cpp
moneromooo-monero: To avoid all the diff spam changing lines that don't actually change.

contrib/epee/include/net/abstract_http_client.h
moneromooo-monero: I don't know offhand if any show up on the profile.

src/wallet/wallet2.cpp
vtnerd: Seems like poor choice of words on my part - I know why you suggested it. But creating a class to prevent simple line changes seems incorrect. Especially since it hides pointer semantics - the value can be `NULL` but this is hidden at every usage point with an "auto" dereference.

moneromooo-monero: Not sure if you checked as the current diff does not show them anymore, but it changed every single RPC call for example. It was all over the place in wallet2.cpp.

contrib/epee/include/net/abstract_http_client.h
```diff
@@ -0,0 +1,173 @@
+#pragma once
```
vtnerd: This file has no copyright notice. Unfortunately, I think it will need to copy the notice at the top of the original file since this code dates back to the cryptonote fork.

src/wallet/wallet2.cpp
woodser: I also think it should be based on what is most correct and natural over these line diffs.  I've updated the PR to work with `std::unique_ptr` until moneromooo-monero's suggestion can be implemented for comparison.

contrib/epee/include/net/http_client.h
woodser: `epee::net_utils::http::http_simple_client` is extended by `download_client` in `download.cpp` so adding final causes a compile error.  Please advise.  Thanks.

https://github.com/monero-project/monero/blob/39e18902763a5fd6f355294d8b135c8f5bc2ec99/src/common/download.cpp#L94

contrib/epee/include/net/abstract_http_client.h
woodser: No compile errors when moved to cpp.

contrib/epee/src/mlocker.cpp
woodser: `mlocker` functions are being called by using the `mlocked` struct like [here](https://github.com/monero-project/monero/blob/39e18902763a5fd6f355294d8b135c8f5bc2ec99/src/wallet/wallet2.cpp#L3923) in `wallet2::setup_keys()`, so currently, individual functions in `mlocker` need guarded.

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -0,0 +1,142 @@
...
+      return std::string(1, i0 * 16 | i1);
+    }
+    //----------------------------------------------------------------------------------------------------
+    std::string convert(char val)
+    {
+      std::string csRet;
+      csRet += "%";
+      csRet += dec_to_hex(val, 16);
+      return  csRet;
+    }
```
moneromooo-monero: Not related to your change, but this does two allocs per byte... Just noting it so it can be changed later if it turns out to be a bottleneck somewhere.

contrib/epee/src/mlocker.cpp
moneromooo-monero: I don't understand the implication. Can you explain ?

src/wallet/wallet2.cpp
```diff
@@ -5772,7 +5817,24 @@ void wallet2::store_to(const std::string &path, const epee::wipeable_string &pas
...
-  
+}
+//----------------------------------------------------------------------------------------------------
+bool wallet2::get_cache_file_data(const epee::wipeable_string &password, wallet2::cache_file_data &cache_file_data)
+{
+  trim_hashchain();
+
+  std::stringstream oss;
+  boost::archive::portable_binary_oarchive ar(oss);
+  ar << *this;
```
moneromooo-monero: That's throwy for a function that returns a bool.

woodser: I made get_cache_file_data() return boost::optional<wallet2::cache_file_data> to be consistent with get_keys_file_data(), and an exception is thrown if boost::none is returned.

contrib/epee/src/mlocker.cpp
woodser: Execution in WebAssembly will fail on `CRITICAL_REGION_LOCAL(mutex())` in mlocker.cpp even if HAVE_MLOCK is not defined because the `mlocked` struct is used externally without checking for this definition.  The current checks in question prevent that.  Do you prefer another way?

moneromooo-monero:
I've just noticed one of my comments was on the wrong context.

My comment "I don't understand the implication. Can you explain ?" was intended to point to:

> Execution in WebAssembly will fail on CRITICAL_REGION_LOCAL(mutex()) in mlocker.cpp even
> if HAVE_MLOCK is not defined because the mlocked struct is used externally without checking
> for this definition. The current checks in question prevent that. Do you prefer another way?

woodser:
I believe that comment was my response to your question.  The implication is execution of `CRITICAL_REGION_LOCAL(mutex())` needs to be prevented in order to run in WebAssembly.  It is currently done with the `#if defined` checks, but I'm open to better suggestions.  Maybe I don't understand your question?

moneromooo-monero:
Why does "the mlocked struct is used externally without checking
for this definition" imply "execution of CRITICAL_REGION_LOCAL(mutex()) needs to be prevented" ?

woodser:
The former does not imply the latter.  In response to the original question, "Why add all of these when you can just not define HAVE_MLOCK ? Is this somehow not working ?", that does not work because the mlocked struct is used externally without checking the HAVE_MLOCK definition.  Therefore, without the added #if defined checks, CRITICAL_REGION_LOCAL(mutex()) will execute and crash in WebAssembly, which needs to be prevented.

moneromooo-monero:
Why does CRITICAL_REGION_LOCAL(mutex()) crash ?

woodser:
It was a consequence of calling `boost::this_thread::sleep_for` in the CRITICAL_REGION_LOCAL definition when pthreads are not enabled in emscripten (such as for browsers that do not support them).  I made an update to define CRITICAL_REGION_LOCAL with an empty block in this case.

vtnerd:
> It was a consequence of calling boost::this_thread::sleep_for in the CRITICAL_REGION_LOCAL definition when pthreads are not enabled in emscripten (such as for browsers that do not support them). I made an update to define CRITICAL_REGION_LOCAL with an empty block in this case.

Is the target always single-thread in that case?

woodser:
Yes.

moneromooo-monero:
*Plenty* of monero code uses threads. Those don't crash ?

woodser:
No, I don't think the issue is use of multithread apis which work even without pthreads.  For some reason, calling sleep_for here causes loading the compiled webassembly module to hang at full utilization.  Tagging @endogenic because MyMonero also needed to comment this out in their webassembly module and in case they know any more about the issue.

woodser:
The tests are passing with the last commit.  Shall I squash all commits?

moneromooo-monero:
Yes please.

contrib/epee/include/syncobj.h
```diff
@@ -150,10 +150,15 @@ namespace epee
...
 
 
-#define  CRITICAL_REGION_LOCAL(x) {boost::this_thread::sleep_for(boost::chrono::milliseconds(epee::debug::g_test_dbg_lock_sleep()));}   epee::critical_region_t<decltype(x)>   critical_region_var(x)
-#define  CRITICAL_REGION_BEGIN(x) { boost::this_thread::sleep_for(boost::chrono::milliseconds(epee::debug::g_test_dbg_lock_sleep())); epee::critical_region_t<decltype(x)>   critical_region_var(x)
+
+#if defined(__EMSCRIPTEN__)
+#define  CRITICAL_REGION_LOCAL(x) {} epee::critical_region_t<decltype(x)>   critical_region_var(x) // TODO: calling sleep_for here causes loading wasm module to hang at full utilization
+#else
+#define  CRITICAL_REGION_LOCAL(x) {boost::this_thread::sleep_for(boost::chrono::milliseconds(epee::debug::g_test_dbg_lock_sleep()));} epee::critical_region_t<decltype(x)>   critical_region_var(x)
+#endif
```
moneromooo-monero: I think you can remove the sleep altogether, I intended to do this for a while.

contrib/epee/src/net_ssl.cpp
```diff
@@ -558,3 +567,36 @@ bool ssl_support_from_string(ssl_support_t &ssl, boost::string_ref s)
...
+    }
+
+    CertFreeCertificateContext(pContext);
+    CertCloseStore(hStore, 0);
+
+    // attach X509_STORE to boost ssl context
+    SSL_CTX_set_cert_store(ctx, store);
+}
+#endif
+
```
moneromooo-monero: That's existing code. Are you adding a copy for some reason ?

src/cryptonote_basic/miner.cpp
```diff
@@ -476,7 +476,7 @@ namespace cryptonote
     for(; bl.nonce != std::numeric_limits<uint32_t>::max(); bl.nonce++)
     {
       crypto::hash h;
-      gbh(bl, height, tools::get_max_concurrency(), h);
+      gbh(bl, height, diffic <= 100 ? 0 : tools::get_max_concurrency(), h);
```
moneromooo-monero: Also existing code. You probably messed up some rebase somewhere.

src/cryptonote_core/cryptonote_tx_utils.cpp
```diff
@@ -663,6 +663,11 @@ namespace cryptonote
     bl.minor_version = CURRENT_BLOCK_MINOR_VERSION;
     bl.timestamp = 0;
     bl.nonce = nonce;
+
+    #if defined __EMSCRIPTEN__  // miner not accessible in webassembly
+      return false;
+    #endif
```
moneromooo-monero: If you can't generate the genesis block, you can't even start the wallet2 hash chain. There is a C implementation of Cryptonight. Does it not work ?

woodser: The wallet2 instances created through webassembly generally work; they're able to get the balance, send, receive, etc.  Is there a problem I can expect to see?

Can you elaborate how to use the C implementation of Cryptonight here?

moneromooo-monero: They'll get a wrong genesis block hash. I guess it might not matter in practice for the wallet only...

To use the C version of Cryptonight, you ensure the C version is compiled in slow-hash.c. It should be the case if you define NO_AES, looking at the code.

woodser: That worked.  I've removed this change.

src/wallet/wallet2.cpp
```diff
@@ -1183,7 +1184,7 @@ wallet2::wallet2(network_type nettype, uint64_t kdf_rounds, bool unattended):
   m_light_wallet_balance(0),
   m_light_wallet_unlocked_balance(0),
   m_original_keys_available(false),
-  m_message_store(),
+  m_message_store(std::unique_ptr<epee::net_utils::http::abstract_http_client>(m_http_client->clone())),
```
xiphon, line 1187: Consider to use `std::shared_ptr` instead

woodser: A shared pointer would introduce a functional change from current master, which uses a separate http client instance for each class.  The functional tests fail when changed to a shared pointer.

xiphon: I don't like `virtual abstract_http_client* clone() const = 0;` idea.

woodser: The message store needs to be initialized with a http client of the same type as wallet2, which the clone() achieves.  Do you suggest any different or better approach?

woodser: By the way, a shared_ptr was my first choice as well.

moneromooo-monero: I don't understand why switching to a shared_ptr instead of unique_ptr would break (I don't care which one is used otherwise). Can you expand on this a bit ?

xiphon: Clone implies we are copying some existing instance state. But actually the message store just needs a new http client instead (not a clone of an existing one).

Off the top of my head, you could pass http client factory to wallet2 and message store.

In the simplest case it could be just a callback returning new `std::unique_ptr<abstract_http_client>` instance.

xiphon: Ah, i didn't see your guys comments before posting my reply.

woodser: Ok I added a factory and removed the clone().  I can make name and styling change if there are any suggestions.

contrib/epee/include/syncobj.h
```diff
@@ -150,10 +150,10 @@ namespace epee
...
-#define  CRITICAL_REGION_BEGIN(x) { boost::this_thread::sleep_for(boost::chrono::milliseconds(epee::debug::g_test_dbg_lock_sleep())); epee::critical_region_t<decltype(x)>   critical_region_var(x)
+#define  CRITICAL_REGION_LOCAL(x) {} epee::critical_region_t<decltype(x)>   critical_region_var(x)
+#define  CRITICAL_REGION_BEGIN(x) {boost::this_thread::sleep_for(boost::chrono::milliseconds(epee::debug::g_test_dbg_lock_sleep())); epee::critical_region_t<decltype(x)>   critical_region_var(x)
 #define  CRITICAL_REGION_LOCAL1(x) {boost::this_thread::sleep_for(boost::chrono::milliseconds(epee::debug::g_test_dbg_lock_sleep()));} epee::critical_region_t<decltype(x)>   critical_region_var1(x)
-#define  CRITICAL_REGION_BEGIN1(x) {  boost::this_thread::sleep_for(boost::chrono::milliseconds(epee::debug::g_test_dbg_lock_sleep())); epee::critical_region_t<decltype(x)>   critical_region_var1(x)
+#define  CRITICAL_REGION_BEGIN1(x) {boost::this_thread::sleep_for(boost::chrono::milliseconds(epee::debug::g_test_dbg_lock_sleep())); epee::critical_region_t<decltype(x)>   critical_region_var1(x)
```
xiphon, line 156: Looks like just an indentation change.

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -0,0 +1,142 @@
+#include "net/abstract_http_client.h"
+#include "net/http_base.h"
+#include "net/net_parse_helpers.h"
+
+#undef MONERO_DEFAULT_LOG_CATEGORY
+#define MONERO_DEFAULT_LOG_CATEGORY "net.http"
+
+namespace epee
+{
+  namespace net_utils
```
xiphon, line 10: Don't indent the namespace

src/wallet/message_store.h
```diff
@@ -202,7 +203,8 @@ namespace mms
   class message_store
   {
   public:
-    message_store();
+    message_store(std::unique_ptr<epee::net_utils::http::abstract_http_client> http_client = std::unique_ptr<epee::net_utils::http::http_simple_client>(new epee::net_utils::http::http_simple_client()));
```
xiphon, line 206: Why do we need a default `= std::unique_ptr<epee::net_utils::http::http_simple_client>(new epee::net_utils::http::http_simple_client())`?

src/wallet/wallet2.cpp
```diff
@@ -1178,7 +1179,7 @@ wallet2::wallet2(network_type nettype, uint64_t kdf_rounds, bool unattended):
   m_light_wallet_balance(0),
   m_light_wallet_unlocked_balance(0),
   m_original_keys_available(false),
-  m_message_store(),
+  m_message_store(std::move(http_client_factory->create())),
```
xiphon, line 1182:
```suggestion
  m_message_store(http_client_factory->create()),
```

src/wallet/wallet2.cpp
```diff
@@ -1328,7 +1329,12 @@ bool wallet2::init(std::string daemon_address, boost::optional<epee::net_utils::
   m_is_initialized = true;
   m_upper_transaction_weight_limit = upper_transaction_weight_limit;
   if (proxy != boost::asio::ip::tcp::endpoint{})
-    m_http_client.set_connector(net::socks::connector{std::move(proxy)});
+  {
+    epee::net_utils::http::abstract_http_client* abstract_http_client = m_http_client.get();
+    epee::net_utils::http::http_simple_client* http_simple_client = dynamic_cast<epee::net_utils::http::http_simple_client*>(abstract_http_client);
+    CHECK_AND_ASSERT_MES(http_simple_client != nullptr, false, "http_simple_client must be used to set proxy");
+    http_simple_client->set_connector(net::socks::connector{std::move(proxy)});
+  }
```
xiphon, line 1337: I would avoid `dynamic_cast`. On the other hand, if others are OK with it, i could fix this later in a separate PR.

src/wallet/message_transporter.h
```diff
@@ -83,7 +83,7 @@ typedef epee::misc_utils::struct_init<transport_message_t> transport_message;
 class message_transporter
 {
 public:
-  message_transporter();
+  message_transporter(std::unique_ptr<epee::net_utils::http::abstract_http_client> http_client = std::unique_ptr<epee::net_utils::http::http_simple_client>(new epee::net_utils::http::http_simple_client()));
```
xiphon, line 86: I would drop this default too

contrib/epee/include/net/abstract_http_client.h
```diff
@@ -0,0 +1,87 @@
...
+
+#include <string>
+#include <boost/optional/optional.hpp>
+#include "http_auth.h"
+#include "net/net_ssl.h"
+
+namespace epee
+{
+namespace net_utils
```
esarachan, line 35: NIT: nested namespaces should probably be indented.

selsta: See https://github.com/monero-project/monero/pull/6332#discussion_r407177508

I don’t think this should get changed again.

PR #6335, opened 2020-02-12
wipeable_string: split - treat CR, LF and Tabs as separators

xiphon:


contrib/epee/src/wipeable_string.cpp
```diff
@@ -188,13 +188,14 @@ void wipeable_string::split(std::vector<wipeable_string> &fields) const
   while (len--)
   {
     const char c = *ptr++;
-    if (c != ' ')
```
vtnerd: Whats the source for this field? Is it CLI? Should this use `std::isspace` so that its locale aware? The locale instance can be cached to speed up processing a bit.

xiphon: Updated

PR #6337, opened 2020-02-12
Support for supercop ASM in wallet, and benchmark for supercop

vtnerd:
Sorry for the do not merge tag again. But this requires [a PR in another repo](http://github.com/monero-project/supercop/pull/3).

This is a working example of the amd64 ASM from supercop to speedup wallet scanning. The speedup is less now, presumably due to slow downs in other areas of the wallet. Scanning from block 1 to 2031549 (mainnet) drops from 5.8 minutes to 4.14 minutes (~28.6%) on my test machine. The benchmark utility (see below) shows a crypto speedup of ~146% with two output transactions.

The test machine is a ryzen 3900x / 32 GiB PC3200 RAM / 970pro nvme / x570 desktop. The daemon and wallet were on the same box. The first run for each crypto implementation was tossed in an effort to "normalize" the file cache. The disk is encrypted, so LMDB hitting disk has an additional hit.

This uses namespace aliases, and therefore has ZERO overhead when disabled. Not even a hit to the executable size. When this is enabled, everything using the "device" library will increase in size due to the extra crypto code. This is unfortunate.

This the output from the benchmark utility included with this PR. It can "sideload" multiple crypto implementations to test the relative performance.

```
Running benchmark using 1000 iterations
Transaction Component Benchmarks
--------------------------------
+ derive_subaddress_public_key step 14248881 ns (+0%)
|-+ standard 14248881 ns (+0%)
| |----> amd64_64_24k => 14248881 ns (+0%)
| |----> amd64_51_30k => 15410920 ns (+8.1553%)
| |--------------> cn => 34357061 ns (+141.121%)
|
+ generate_key_derivation step 37039500 ns (+159.947%)
|-+ standard 37039500 ns (+0%)
| |----> amd64_64_24k => 37039500 ns (+0%)
| |----> amd64_51_30k => 40765969 ns (+10.0608%)
| |--------------> cn => 93774626 ns (+153.175%)
|
Transaction Benchmarks
----------------------
+ Transactions with 2 outputs 65504972 ns (+0%)
|-+ standard 65504972 ns (+0%)
| |----> amd64_64_24k =>  65504972 ns (+0%)
| |----> amd64_51_30k =>  71416899 ns (+9.02516%)
| |--------------> cn => 161385237 ns (+146.371%)
|
+ Transactions with 4 outputs 93948623 ns (+43.4221%)
|-+ standard 93948623 ns (+0%)
| |----> amd64_64_24k =>  93948623 ns (+0%)
| |----> amd64_51_30k => 102212480 ns (+8.79614%)
| |--------------> cn => 229769098 ns (+144.569%)
|
```

### DIFF
--- a/tests/benchmark.cpp
+++ b/tests/benchmark.cpp
@@ -1,0 +1,437 @@
+// Copyright (c) 2020, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "tests/benchmark.h"
+
+#include <boost/fusion/adapted/std_tuple.hpp>
+#include <boost/fusion/algorithm/iteration/fold.hpp>
+#include <boost/preprocessor/seq/enum.hpp>
+#include <boost/preprocessor/seq/for_each.hpp>
+#include <boost/preprocessor/seq/seq.hpp>
+#include <boost/preprocessor/seq.hpp>
+#include <boost/preprocessor/stringize.hpp>
+#include <boost/spirit/include/karma_char.hpp>
+#include <boost/spirit/include/karma_format.hpp>
+#include <boost/spirit/include/karma_repeat.hpp>
+#include <boost/spirit/include/karma_right_alignment.hpp>
+#include <boost/spirit/include/karma_sequence.hpp>
+#include <boost/spirit/include/karma_string.hpp>
+#include <boost/spirit/include/karma_uint.hpp>
+#include <boost/spirit/include/qi_char.hpp>
+#include <boost/spirit/include/qi_list.hpp>
+#include <boost/spirit/include/qi_parse.hpp>
+#include <boost/spirit/include/qi_uint.hpp>
+#include <chrono>
+#include <cstring>
+#include <functional>
+#include <iostream>
+#include <stdexcept>
+#include <string>
+#include <tuple>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+#include "crypto/crypto.h"
+#include "cryptonote_basic/cryptonote_basic.h"
+#include "monero/crypto/amd64-64-24k.h"
+#include "monero/crypto/amd64-51-30k.h"
+
+#define CHECK(...)                           \
+    if(!( __VA_ARGS__ ))                      \
+        throw std::runtime_error{             \
+            "TEST FAILED (line  "             \
+            BOOST_PP_STRINGIZE( __LINE__ )    \
+            "): "                             \
+            BOOST_PP_STRINGIZE( __VA_ARGS__ ) \
+        }
+
+//! Define function that forwards arguments to `crypto::func`.
+#define FORWARD_FUNCTION(func)                             \
+    template<typename... T>                                \
+    static bool func (T&&... args)                         \
+    {                                                      \
+        return ::crypto:: func (std::forward<T>(args)...); \
+    }
+
+#define CRYPTO_FUNCTION(library, func)                        \
+    BOOST_PP_CAT(BOOST_PP_CAT(monero_crypto_, library), func)
+
+#define CRYPTO_BENCHMARK(r, _, library)                                                                                                                                  \
+    struct library                                                                                                                                                       \
+    {                                                                                                                                                                    \
+        static constexpr const char* name() noexcept { return BOOST_PP_STRINGIZE(library); }                                                                             \
+        static bool generate_key_derivation(const ::crypto::public_key &tx_pub, const ::crypto::secret_key &view_sec, ::crypto::key_derivation &out)                     \
+        {                                                                                                                                                                \
+            return CRYPTO_FUNCTION(library, _generate_key_derivation) (out.data, tx_pub.data, view_sec.data) == 0;                                                       \
+        }                                                                                                                                                                \
+        static bool derive_subaddress_public_key(const ::crypto::public_key &spend_pub, const ::crypto::key_derivation &d, std::size_t index, ::crypto::public_key &out) \
+        {                                                                                                                                                                \
+            ::crypto::ec_scalar scalar;                                                                                                                                  \
+            ::crypto::derivation_to_scalar(d, index, scalar);                                                                                                            \
+            return CRYPTO_FUNCTION(library, _generate_subaddress_public_key) (out.data, spend_pub.data, scalar.data) == 0;                                               \
+        }                                                                                                                                                                \
+    };
+
+
+namespace
+{
+    //! Default number of iterations for benchmark timing.
+    constexpr const unsigned default_iterations = 1000;
+
+    //! \return Byte compare two objects of `T`.
+    template<typename T>
+    bool compare(const T& lhs, const T& rhs) noexcept
+    {
+        static_assert(!epee::has_padding<T>(), "type might have padding");
+        return std::memcmp(std::addressof(lhs), std::addressof(rhs), sizeof(T)) == 0;
+    }
+
+    //! Benchmark default monero crypto library - a re-arranged ref10 implementation.
+    struct cn
+    {
+        static constexpr const char* name() noexcept { return "cn"; }
+        FORWARD_FUNCTION( generate_key_derivation );
+        FORWARD_FUNCTION( derive_subaddress_public_key );
+    };
+
+    // Define functions for every library except for `cn` which is the head library.
+    BOOST_PP_SEQ_FOR_EACH(CRYPTO_BENCHMARK, _, BOOST_PP_SEQ_TAIL(BENCHMARK_LIBRARIES));
+
+    // All enabled benchmark libraries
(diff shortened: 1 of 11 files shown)


binaryFate:
Impressive numbers, great work!

What's the additional size to expect roughly?

vtnerd:
On this setup, static builds of `monerod` increases by 88.02 KiB (0.5%) and `monero-wallet-cli` increases by 84.02 KiB (0.5%). The daemon increase is exactly 4Kib (to the byte), so something involving padding was probably triggered.

`monerod` increases because it has a dependency on `device` via `cryptonote_basic`, `cryptonote_core`, and `ringct`. AFAIK, it should be dead code too, but it cannot be dropped because `default_device` is created in static memory, so linking the associated file brings in every function associated with that class. Fixing this is going to require a bit of additional work. It should likely be done regardless of this PR status, because the `device` design/implementation is dragging too much code into `monerod` already.

It is possible to build `monerod` without this ASM, then build `monero-walet-cli` with the ASM, but a cmake change is required between `make` calls. I'm not aware of an easier technique.

erciccione:
> [a PR in another repo](https://github.com/monero-project/monero/pull/%60github.com/monero-project/supercop/pull/3%60)

That link is broken, i guess you meant to link to https://github.com/monero-project/supercop/pull/3 :)

Nice improvement btw, looking forward to see this merged.

vtnerd:
I tested scanning on two machines with less awesome CPUs. The same ryzen 3900x / 970pro is being used to run the daemon, with the wallet scanning on:

Fanless laptop with i7-7Y75 (1.3 GHz 2/4 cores 3MiB cache): 46.07 minutes to 20.52 minutes (~55.4%).

2014 Mac Mini i5 (2.5 GHz 2/4 cores 4MiB cache): 38.29 minutes to 24.00 minutes (~37%).

I was incorrect earlier about changes to the code affecting performance - these numbers are similar to what was seen previously. I need a second identical system to see how running the daemon + wallet on a single machine affects performance. Or perhaps just look at the CPU utilization more closely.

Either way, the increase is fairly big for lower-core and/or lower clock speed x86-64 systems.

selsta:
2014 MacBook Pro (`Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz`)

```
selsta@mbpR ~/d/m/s/c/wallet (master)> ./monero-wallet-crypto-bench 
Running benchmark using 1000 iterations
Transaction Component Benchmarks
--------------------------------
+ derive_subaddress_public_key step 22453845 ns (+0%)
|-+ standard 22453845 ns (+0%)
| |----> amd64_51_30k => 22453845 ns (+0%)
| |----> amd64_64_24k => 23175596 ns (+3.21438%)
| |--------------> cn => 49038406 ns (+118.396%)
| 
+ generate_key_derivation step 55552415 ns (+147.407%)
|-+ standard 55552415 ns (+0%)
| |----> amd64_51_30k =>  55552415 ns (+0%)
| |----> amd64_64_24k =>  62587561 ns (+12.664%)
| |--------------> cn => 123768234 ns (+122.795%)
| 
Transaction Benchmarks
----------------------
+ Transactions with 2 outputs 106088524 ns (+0%)
|-+ standard 106088524 ns (+0%)
| |----> amd64_51_30k => 106088524 ns (+0%)
| |----> amd64_64_24k => 115446729 ns (+8.82113%)
| |--------------> cn => 224049592 ns (+111.191%)
| 
+ Transactions with 4 outputs 146760073 ns (+38.3374%)
|-+ standard 146760073 ns (+0%)
| |----> amd64_51_30k => 146760073 ns (+0%)
| |----> amd64_64_24k => 154913459 ns (+5.55559%)
| |--------------> cn => 316314887 ns (+115.532%)
| 
```

Wallet scan: ~17min -> ~12,5min, did not do it super scientific so could be off a bit.

selsta:
ping @vtnerd 

Now that the supercop PR is merged, how should we continue here? Is this still do not merge?

vtnerd:
Removed the "[Do not merge]" tag.

vtnerd:
Whoops, re-added the "[Do Not Merge]" tag. I need to update the git attributes so that a specific hash is referenced (for security purposes) to the external supercop repo.

vtnerd:
Force pushed as rebase with a pinned `external/supercop` to the recently merged commit.

selsta:
@vtnerd Does not seem to compile on Windows CI.

moneromooo-monero:
"Didn't find assembler"

vtnerd:
Force pushed a cmake change. I couldn't reproduce this locally, so its bit of a guess as to whether this clears the CI.

@selsta @moneromooo-monero I can drop the benchmark code, leaving cmake and a bit of C++ glue code, and then issue a separate PR for the benchmark to get things moving. The benchmark was useful for me to run sanity checking - it compares results against the mainline Monero and aborts if the cryptographic output differs. And does this for all of the crypto backends instead of just the one linked against in the wallet.

vtnerd:
Ok so fixed one cmake issue, but there's definitely a bug in the upstream supercop repo for autodetecting the targeted architecture. I tested it at some point, but not well enough. I think its a missing parenthesis, but I'll need to install some cross-compilers first. The windows build on travis properly filters out the libraries though. Hmm,

moneromooo-monero:
That line you removed was added for "cmake: ARCH_ID fixes for cross compilation". I'll ping TheCharlatan to comment.

moneromooo-monero:
I thought I had reviewed that before. I hadn't. Sorry. I'm happy with that patch, assuming TheCharlatan agrees with the cmake change. The benchmark should really be in tests (maybe performance_tests, or another) though.

vtnerd:
> That line you removed was added for "cmake: ARCH_ID fixes for cross compilation". I'll ping TheCharlatan to comment.

I need to update it anyway - it needs to be set to the target architecture. Depends refers to this as the host machine, but `SYSTEM_CMAKE_HOST_PROCESSOR` is the architecture doing the compiling.

The other issue I have right is now the `if (UNIX ...)` check in supercop. This returns true when compiling within a cygwin like environment (i.e. the github CI), but the calling convention used by the compiler is incorrect for the ASM provided (Windows and _not_ System V AMD64).

vtnerd:
> The benchmark should really be in tests (maybe performance_tests, or another) though.

The cmake I have here is a little funky, but I think this is possible. Will try later today to move it.

iDunk5400:
I managed to get
```
-- The ASM-ATT compiler identification is GNU
-- Found assembler: /usr/bin/x86_64-w64-mingw32-as
Monero crypto autodetect selected "amd64-64-24k" for target platform
```
in both MSYS2 and Ubuntu depends cross-compiler (cross-compilers cmake config shown above), but looks like `monero-wallet-crypto-bench.exe` does not get built in either.

Btw, the native build fix is simple:
```diff
diff --git a/cmake/32-bit-toolchain.cmake b/cmake/32-bit-toolchain.cmake
index 6bee56482..00d0ffb49 100644
--- a/cmake/32-bit-toolchain.cmake
+++ b/cmake/32-bit-toolchain.cmake
@@ -33,6 +33,7 @@ endif()
 set (GCC_PREFIX i686-w64-mingw32)
 set (CMAKE_C_COMPILER ${GCC_PREFIX}-gcc)
 set (CMAKE_CXX_COMPILER ${GCC_PREFIX}-g++)
+set (CMAKE_ASM-ATT_COMPILER as)
 set (CMAKE_AR ar CACHE FILEPATH "" FORCE)
 set (CMAKE_NM nm CACHE FILEPATH "" FORCE)
 set (CMAKE_LINKER ld CACHE FILEPATH "" FORCE)
diff --git a/cmake/64-bit-toolchain.cmake b/cmake/64-bit-toolchain.cmake
index 5eaa17596..82d1cd01e 100644
--- a/cmake/64-bit-toolchain.cmake
+++ b/cmake/64-bit-toolchain.cmake
@@ -33,6 +33,7 @@ endif()
 set (GCC_PREFIX x86_64-w64-mingw32)
 set (CMAKE_C_COMPILER ${GCC_PREFIX}-gcc)
 set (CMAKE_CXX_COMPILER ${GCC_PREFIX}-g++)
+set (CMAKE_ASM-ATT_COMPILER as)
 set (CMAKE_AR ar CACHE FILEPATH "" FORCE)
 set (CMAKE_NM nm CACHE FILEPATH "" FORCE)
 set (CMAKE_LINKER ld CACHE FILEPATH "" FORCE)
```

iDunk5400:
Hm ok, I see now `monero-wallet-crypto-bench` is not built by default. I built it in MSYS2, but it segfaults. I'll run it in gdb.

iDunk5400:
```
Running benchmark using 1000 iterations
Transaction Component Benchmarks
--------------------------------

Thread 1 received signal SIGSEGV, Segmentation fault.
0x000000000040b33d in crypto_sign_ed25519_amd64_64_fe25519_square ()
(gdb) bt
#0  0x000000000040b33d in crypto_sign_ed25519_amd64_64_fe25519_square ()
#1  0x0000000000000048 in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```
That's on a Ryzen 5 1600X on Windows 10 20.04.

moneromooo-monero:
Sounds like "but the calling convention used by the compiler is incorrect for the ASM provided".

vtnerd:
TL;DR - to fix the depends macOS/darwin build will require a change in the supercop repo. Native builds on macOS/darwin have always worked.

The cross-compile needs `CMAKE_ASM-ATT_COMPILER_TARGET`, but this appears to be broken for `ASM-ATT` with the Clang assembler. I can get the `CMAKE_ASM_COMPILER_TARGET` flag to work - a look at CMake code looks like it doesn't call a specific macro with `ASM-ATT` (only `ASM`) .

My testing in macOS and Linux indicates that only `enable_language(ASM)` is needed despite the files being specific AT&T syntax. It also generates lots of warnings in Linux (due to unused macro defined on command line). I'm not certain if there are larger issues. CMake has a function to check for an AT&T assembler that I will add to supercop as a mitigation.

Otherwise - the `if (UNIX...` check in the supercop repo appears to work for cross-builds or msys builds - Github and Travis are all **not** compiling the amd64 variant into the Windows build (as desired). All other x86-64 builds are compiling in the amd64-64-24k as expected, except the previously mentioned macOS/darwin cross-compile fails.

>> The benchmark should really be in tests (maybe performance_tests, or another) though.
>
> The cmake I have here is a little funky, but I think this is possible. Will try later today to move it.

I hacked at this for a while, and failed my first couple attempts. But a possible hack around this just occurred to me while typing the above, so I'll try yet again.

vtnerd:
@iDunk5400 confirmed that this latest commit, in addition to the [PR in the supercop repo](https://github.com/monero-project/supercop/pull/4) should fix the depends build (travis).

I also successfully moved the benchmark utility into `tests/` with minimal crappiness.

@TheCharlatan @moneromooo-monero the changes to the `depends/` folder are worth reviewing now.

cmake/32-bit-toolchain.cmake
```diff
@@ -33,6 +33,8 @@ endif()
 set (GCC_PREFIX i686-w64-mingw32)
 set (CMAKE_C_COMPILER ${GCC_PREFIX}-gcc)
 set (CMAKE_CXX_COMPILER ${GCC_PREFIX}-g++)
+set (CMAKE_AS_COMPILER as)
```
moneromooo-monero, line 36: Is the AS naming intended here (see below) ?

cmake/64-bit-toolchain.cmake
```diff
@@ -33,6 +33,8 @@ endif()
 set (GCC_PREFIX x86_64-w64-mingw32)
 set (CMAKE_C_COMPILER ${GCC_PREFIX}-gcc)
 set (CMAKE_CXX_COMPILER ${GCC_PREFIX}-g++)
+set (CMAKE_ASM_COMPILER as)
```
moneromooo-monero, line 36: Is the ASM naming intended here (see above) ?

tests/benchmark.cpp
```diff
@@ -0,0 +1,436 @@
...
+        std::cout << "Transaction Benchmarks" << std::endl;
+        std::cout << "----------------------" << std::endl;
+        for (const unsigned num : nums)
+            val.push_back(run_suite(tx_suite{{args, num}}));
+        std::sort(val.begin(), val.end());
+        print(val, std::cout, 0);
+    }
+    catch (const std::exception& e)
+    {
+        std::cerr << "Error: " << e.what() << std::endl;
```
moneromooo-monero, line 432: return 1 ?

vtnerd: Added.

vtnerd:
Updated supercop hash to latest commit for macOS/darwin build, which has now passed.

cmake/32-bit-toolchain.cmake
vtnerd: After some thought, I will drop both of these changes (and below). The Windows builds don't use the assembly, and the supercop check should still work with Windows.

vtnerd: Removed here and below.

PR #6338, opened 2020-02-13
Hash domain separation

SarangNoether:
Moves some hash function domain separators to a common configuration location, to reduce the likelihood of future collision.

### DIFF
--- a/src/cryptonote_config.h
+++ b/src/cryptonote_config.h
@@ -204,6 +204,17 @@ namespace config
   std::string const GENESIS_TX = "013c01ff0001ffffffffffff03029b2e4c0281c0b02e7c53291a94d1d0cbff8883f8024f5142ee494ffbbd08807121017767aafcde9be00dcfd098715ebcf7f410daebc582fda69d24a28e9d0bc890d1";
   uint32_t const GENESIS_NONCE = 10000;
 
+  // Hash domain separators
+  const char HASH_KEY_BULLETPROOF_EXPONENT[] = "bulletproof";
+  const char HASH_KEY_RINGDB[] = "ringdsb";
+  const char HASH_KEY_SUBADDRESS[] = "SubAddr";
+  const unsigned char HASH_KEY_ENCRYPTED_PAYMENT_ID = 0x8d;
+  const unsigned char HASH_KEY_WALLET = 0x8c;
+  const unsigned char HASH_KEY_WALLET_CACHE = 0x8d;
+  const unsigned char HASH_KEY_RPC_PAYMENT_NONCE = 0x58;
+  const unsigned char HASH_KEY_MEMORY = 'k';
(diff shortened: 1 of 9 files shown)


src/multisig/multisig.cpp
```diff
@@ -33,19 +33,21 @@
...
 
-static const rct::key multisig_salt = { {'M', 'u', 'l', 't' , 'i', 's', 'i', 'g', 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 } };
-
 namespace cryptonote
 {
   //-----------------------------------------------------------------
   crypto::secret_key get_multisig_blinded_secret_key(const crypto::secret_key &key)
   {
+    rct::key multisig_salt;
+    memcpy(multisig_salt.bytes, config::HASH_KEY_MULTISIG, sizeof(rct::key));
```
iamamyth: Although the config value happens to match `sizeof(rct::key)`, this seems a bit confusing for future readers. I presume the config value is declared an unsigned char array because the `rct::key` type isn't available, so I would suggest a static assertion here that `sizeof(rct::key) == sizeof(config::HASH_KEY_MULTISIG)` in order to clarify the intent and reduce the potential for future accidents (a stray delete in the config could end up copying from uninitialized memory, which may or may not produce a test failure).

SarangNoether: Good point. Will update.

xiphon: Would be better to use compile-time `static_assert` (that @iamamyth suggested) instead of runtime `CHECK_AND_ASSERT_THROW_MES` check.

SarangNoether: Done.

src/cryptonote_config.h
```diff
@@ -207,6 +207,16 @@ namespace config
   std::string const GENESIS_TX = "013c01ff0001ffffffffffff03029b2e4c0281c0b02e7c53291a94d1d0cbff8883f8024f5142ee494ffbbd08807121017767aafcde9be00dcfd098715ebcf7f410daebc582fda69d24a28e9d0bc890d1";
   uint32_t const GENESIS_NONCE = 10000;
 
+  // Hash domain separators
+  const std::string HASH_KEY_BULLETPROOF_EXPONENT("bulletproof");
```
vtnerd: Why is this a `std::string`? It requires initialization at runtime, and the order of this initialization is undefined. A string can be constructed when its needed, it looks like temporary string construction is used anyway.

SarangNoether: It was taken from the corresponding bulletproofs code.

SarangNoether: Updated.

UkoeHB:
I believe subaddress private keys are also domain separated. See [get_subaddress_secret_key()](https://github.com/monero-project/monero/blob/b8ab510f238e5a4b9907ea09ed0f364b34ab1ae5/src/device/device_default.cpp#L198).

SarangNoether:
> I believe subaddress private keys are also domain separated. See [get_subaddress_secret_key()](https://github.com/monero-project/monero/blob/b8ab510f238e5a4b9907ea09ed0f364b34ab1ae5/src/device/device_default.cpp#L198).

Updated.

PR #6342, opened 2020-02-17
rpc: Add check for too old timestamps in payment signatures.

glv2:
Or is there a reason to allow old timestamps?

### DIFF
--- a/src/rpc/rpc_payment_signature.cpp
+++ b/src/rpc/rpc_payment_signature.cpp
@@ -102,6 +102,11 @@ namespace cryptonote
       MDEBUG("Timestamp is in the future");
       return false;
     }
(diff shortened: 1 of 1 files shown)


moneromooo-monero:
No particular reason, it doesn't matter either way, unless you found a reason it does.

glv2:
I tried sending a few RPC requests to a node requiring payment, and the client was disappearing very fast from the list given by the ```rpc_payments``` command.

I then realized that the timestamps I had put in the requests were in seconds instead of microseconds, and I think this is why the node thought the client had been inactive for a very long time and dropped it.

Note: the client had 0 credit on the node; I have not tested yet if having some credits and using old timestamps has the same behaviour.

moneromooo-monero:
OK, that's as a good a reason as any :)

PR #6347, opened 2020-02-19
cryptonote_core: skip block notify on blockchain switching rollback

xiphon:


moneromooo-monero:
Presumably you already got notifications for the blocks in the alt chain that untimately failed, right ? If so, it seems to make more sense to not omit those notifications ?

xiphon:
Updated.

1. Failed to switch to alt chain - don't send any notification
2. Successfully switched to alt chain - send 1) reorg notification and 2) block notification for each alt chain block

src/cryptonote_core/blockchain.cpp
```diff
@@ -1044,6 +1044,11 @@ bool Blockchain::switch_to_alternative_blockchain(std::list<block_extended_info>
     reorg_notify->notify("%s", std::to_string(split_height).c_str(), "%h", std::to_string(m_db->height()).c_str(),
         "%n", std::to_string(m_db->height() - split_height).c_str(), "%d", std::to_string(discarded_blocks).c_str(), NULL);
 
+  std::shared_ptr<tools::Notify> block_notify = m_block_notify;
```
vtnerd: I don't see any benefit to copying the `std::shared_ptr` in the original code. The copy is not atomic/thread-safe (with respect to assignment of the same `std::shared_ptr` object) unless the appropriate functions are used. Was `std::atomic_store` and `std::atomic_load` intended to be used here?

vtnerd: Sorry ping @moneromooo-monero

moneromooo-monero: Yes, I thought the pointer copy was atomic.

vtnerd: I think the risks of issue here is low due to the init process. But something to remember for sure.

PR #6350, opened 2020-02-23
Use byte_slice for sending zmq messages - removes data copy within zmq

vtnerd:
Sending JSON response on the ZMQ RPC server has two allocations + copies - one into a std::string and a second internally within ZMQ. This removes that second copy (but ZMQ still allocates internally!), which is an obvious performance gain.

Its possible to remove the allocation+copy into `bye_slice` by getting `rapidjson` to write directly into a `byte_slice` compatible buffer. I am "saving" that for a future commit because its a bit more code.

I ran with this with clang ASAN mode, no memory leaks detected in the unit tests or in the ZMQ JSON-RPC server usage.

### DIFF
--- a/src/rpc/daemon_handler.cpp
+++ b/src/rpc/daemon_handler.cpp
@@ -33,6 +33,7 @@
 #include <stdexcept>
 
 #include <boost/uuid/nil_generator.hpp>
+#include <boost/utility/string_ref.hpp>
 // likely included by daemon_handler.h's includes,
 // but including here for clarity
 #include "cryptonote_core/cryptonote_core.h"
@@ -48,7 +49,7 @@ namespace rpc
 {
   namespace
   {
-    using handler_function = std::string(DaemonHandler& handler, const rapidjson::Value& id, const rapidjson::Value& msg);
+    using handler_function = epee::byte_slice(DaemonHandler& handler, const rapidjson::Value& id, const rapidjson::Value& msg);
     struct handler_map
     {
       const char* method_name;
@@ -66,7 +67,7 @@ namespace rpc
     }
 
     template<typename Message>
-    std::string handle_message(DaemonHandler& handler, const rapidjson::Value& id, const rapidjson::Value& parameters)
+    epee::byte_slice handle_message(DaemonHandler& handler, const rapidjson::Value& id, const rapidjson::Value& parameters)
     {
       typename Message::Request request{};
       request.fromJson(parameters);
@@ -903,7 +904,7 @@ namespace rpc
     return true;
   }
 
-  std::string DaemonHandler::handle(const std::string& request)
+  epee::byte_slice DaemonHandler::handle(const std::string& request)
   {
     MDEBUG("Handling RPC request: " << request);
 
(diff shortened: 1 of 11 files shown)


vtnerd:
Rebased latest master changes.

src/rpc/daemon_handler.h
```diff
@@ -132,7 +133,7 @@ class DaemonHandler : public RpcHandler
 
     void handle(const GetOutputDistribution::Request& req, GetOutputDistribution::Response& res);
 
-    std::string handle(const std::string& request);
+    epee::byte_slice handle(const std::string& request) override final;
```
xiphon, line 136:
```suggestion
    epee::byte_slice handle(const std::string& request) final;
```

src/rpc/daemon_handler.cpp
```diff
@@ -921,8 +922,11 @@ namespace rpc
       if (matched_handler == std::end(handlers) || matched_handler->method_name != request_type)
         return BAD_REQUEST(request_type, req_full.getID());
 
-      std::string response = matched_handler->call(*this, req_full.getID(), req_full.getMessage());
-      MDEBUG("Returning RPC response: " << response);
+      epee::byte_slice response = matched_handler->call(*this, req_full.getID(), req_full.getMessage());
+
+      const boost::string_ref response_view{reinterpret_cast<const char*>(response.data()), response.size()};
```
xiphon, line 927: Could be moved inside `MDEBUG(...)`, so it doesn't get evaluated of debug logs are disabled.

vtnerd: That's a huge line to save a pointer copy and length copy. Since all of those definitions are visible in the header, the compiler could elide/delay through analysis already. I'll see how I clean this up though.

vtnerd:
Mergers hold off, going to make the other change suggested ... and perhaps change the second one.

src/rpc/daemon_handler.h
vtnerd: So I just looked at this. I don't understand, do you think the `override` keyword should be dropped?

src/rpc/daemon_handler.cpp
vtnerd: I made this change with a multi-line debug statement. Maybe more eye-pleasing too.

src/rpc/daemon_handler.h
xiphon: Either way is fine to me

src/rpc/daemon_handler.cpp
vtnerd: Ok, now I remember why I did it this way. The comment for the `boost::string_ref` constructor breaks the `MDEBUG` macro which interprets it as an additional macro argument. It should be possible to "fix" the macro, but this temporary object is fairly light-weight to construct.

And it just occurred to me that the compiler is likely to always copy the pointer and length because of the insane rules to observable effects (in the function call to check current log level).

I'll leave as-is for now, its not a huge penalty, but it does requiring changing the macro.

vtnerd:
Ok, everything should be resolved.

vtnerd:
Ah crap, that change wasn't supposed to be pushed. watch the hash changes, the next should be moving it back to the old hash commit.

vtnerd:
And after sketchy git manipulating, we are back at the previous commit hash.

selsta:
Any idea why Windows cross compilation fails?
```
/home/travis/build/monero-project/monero/src/serialization/json_object.cpp: In function 'void cryptonote::json::fromJsonValue(const Value&, cryptonote::txin_v&)':
/home/travis/build/monero-project/monero/src/serialization/json_object.cpp:331:31: error: 'const Value {aka const class rapidjson::GenericValue<rapidjson::UTF8<> >}' has no member named 'GetObjectA'; did you mean 'GetObject'?
   for (auto const& elem : val.GetObject())
                               ^
/home/travis/build/monero-project/monero/src/serialization/json_object.cpp: In function 'void cryptonote::json::fromJsonValue(const Value&, cryptonote::tx_out&)':
/home/travis/build/monero-project/monero/src/serialization/json_object.cpp:555:31: error: 'const Value {aka const class rapidjson::GenericValue<rapidjson::UTF8<> >}' has no member named 'GetObjectA'; did you mean 'GetObject'?
   for (auto const& elem : val.GetObject())
                               ^
```

vtnerd:
The code was forked before the xiphon fix for the terrible Windows macro. I will rebase - the PR is relatively short for reviewers that need to scan again.

PR #6351, opened 2020-02-23
Improve hex->bin Conversion

vtnerd:
The initial reason for the change was hex->bin conversion in the ZMQ JSON-RPC server required a temporary `std::string` for each attempt. The individual commits should explain the other changes.

@hyc and @moneromooo-monero should have copyrights, allowing the move into the monero copyrighted file. This move isn't strictly necessary to improve performance, but does it make easier on the compiler (less code in the header AND declared in a header with less overhead). The latter is useful because the ZMQ JSON code can (eventually) just include `hex.h` instead of the more expensive `string_tools.h`.

### DIFF
--- a/contrib/epee/src/hex.cpp
+++ b/contrib/epee/src/hex.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2019, The Monero Project
+// Copyright (c) 2017-2020, The Monero Project
 //
 // All rights reserved.
 //
@@ -32,6 +32,8 @@
 #include <limits>
 #include <ostream>
 #include <stdexcept>
+
+#include "storages/parserse_base_utils.h"
 
 namespace epee
 {
@@ -84,7 +86,42 @@ namespace epee
     return write_hex(out, src);
   }
 
-  std::vector<uint8_t> from_hex::vector(boost::string_ref src)
+
+  bool from_hex::to_string(std::string& out, const boost::string_ref src)
+  {
+    out.resize(src.size() / 2);
+    return to_buffer_unchecked(reinterpret_cast<std::uint8_t*>(&out[0]), src);
+  }
+
+  bool from_hex::to_buffer(span<std::uint8_t> out, const boost::string_ref src) noexcept
+  {
+    if (src.size() / 2 != out.size())
+      return false;
+    return to_buffer_unchecked(out.data(), src);
+  }
+
+  bool from_hex::to_buffer_unchecked(std::uint8_t* dst, const boost::string_ref s) noexcept
+  {
+      if (s.size() % 2 != 0)
+        return false;
+
+      const unsigned char *src = (const unsigned char *)s.data();
+      for(size_t i = 0; i < s.size(); i += 2)
+      {
+        int tmp = *src++;
+        tmp = epee::misc_utils::parse::isx[tmp];
+        if (tmp == 0xff) return false;
+        int t2 = *src++;
+        t2 = epee::misc_utils::parse::isx[t2];
+        if (t2 == 0xff) return false;
+        *dst++ = (tmp << 4) | t2;
+      }
+
+      return true;
+  }
+
+
+  std::vector<uint8_t> from_hex_locale::to_vector(const boost::string_ref src)
   {
     // should we include a specific character
     auto include = [](char input) {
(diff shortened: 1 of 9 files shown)


hyc:
Anything that moves code out of header files sounds good to me.

contrib/epee/src/hex.cpp
```diff
@@ -84,7 +86,42 @@ namespace epee
...
+
+  bool from_hex::to_string(std::string& out, const boost::string_ref src)
+  {
+    out.resize(src.size() / 2);
+    return to_buffer_unchecked(reinterpret_cast<std::uint8_t*>(&out[0]), src);
+  }
+
+  bool from_hex::to_buffer(span<std::uint8_t> out, const boost::string_ref src) noexcept
+  {
+    if (src.size() / 2 != out.size())
```
omartijn: Wouldn't it be nicer to multiple out.size() by 2 to compare instead of doing integer division? If `src.size()` is odd, this would continue to the `to_buffer_unchecked` before it is rejected.

contrib/epee/src/hex.cpp
```diff
@@ -84,7 +86,42 @@ namespace epee
...
+        t2 = epee::misc_utils::parse::isx[t2];
+        if (t2 == 0xff) return false;
+        *dst++ = (tmp << 4) | t2;
+      }
+
+      return true;
+  }
+
+
+  std::vector<uint8_t> from_hex_locale::to_vector(const boost::string_ref src)
```
omartijn: I can't comment on the offending line, because it already exists in master apparently, but this code assumes that a `boost::string_ref` is always null-terminated, which it doesn't necessarily have to be.

The function contains the following incorrect comment:

> // the data to work with (std::string is always null-terminated)

I am unsure whether we should fix it in this PR or whether a separate fix-PR should be made instead. I tend to lean towards fixing it here, since fixing it in another PR would likely yield merge conflicts when this PR gets merged first.

contrib/epee/src/hex.cpp
vtnerd: Although unlikely, multiplication by 2 can overflow.

contrib/epee/src/hex.cpp
vtnerd: I noticed this as well but wanted to do it in a separate PR so that reviewers would focus just on that.

contrib/epee/src/hex.cpp
omartijn: That's true, but since they're both signed it's perfectly defined behaviour. And since both operands are of the same type (`std::size_t`) if the multiplication overflows the division wouldn't compare the same either.

I do get your point, though. Also, the current code does of course work correctly.

vtnerd: > but since they're both signed

I think you meant `unsigned`? I don't think multiplication works for all cases. One (again, extreme example), would be if `out.size() == (size_t(1) << 63) | size_t(1)` and if `src.size() == 2`. Then `out.size() * 2 == src.size()`, and `src` is still divisible by `2` so the check in `to_buffer_unchecked` succeeds. This still has defined behavior since `out` is larger than `src`, but its no longer rejecting some inexact `out.size()`s.

This is admittedly a contrived example, but the division by 2 is going to be a right shift by 1 in ASM. So the only penalty is sometimes an additional bitmask and comparison is performed if the division check passes.

contrib/epee/src/hex.cpp
vtnerd: See PR #6359.

vtnerd:
Rebased latest master changes.

vtnerd:
Force pushed a change which removed the inclusion of `string_tools.h` which is no longer needed by zmq json object.

src/serialization/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2016-2019, The Monero Project
+# Copyright (c) 2016-2020, The Monero Project
```
xiphon: Do you want to update just the copyright year in the file?

vtnerd: Reverted.

vtnerd:
Rebased. This now has a `#undef GetObject` in the cpp.

src/serialization/CMakeLists.txt
xiphon: Seems not

vtnerd: This time the change was definitely added to the commit list.

PR #6353, opened 2020-02-24
daemon: if no banned IPs print something

### LINKED ISSUE #6352: "bans" command shows an empty string if there are no banned peers
Maybe would better to add a message like "No banned peers".

sumogr:
Closes #6352

### DIFF
--- a/src/daemon/rpc_command_executor.cpp
+++ b/src/daemon/rpc_command_executor.cpp
@@ -1699,14 +1699,18 @@ bool t_rpc_command_executor::print_bans()
         }
     }
 
-    for (auto i = res.bans.begin(); i != res.bans.end(); ++i)
+    if (!res.bans.empty())
     {
-        tools::msg_writer() << i->host << " banned for " << i->seconds << " seconds";
+        for (auto i = res.bans.begin(); i != res.bans.end(); ++i)
+        {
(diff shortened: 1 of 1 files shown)


src/daemon/rpc_command_executor.cpp
```diff
@@ -1699,15 +1699,19 @@ bool t_rpc_command_executor::print_bans()
...
+            tools::msg_writer() << i->host << " banned for " << i->seconds << " seconds";
+        }
     }
+    else 
+        tools::msg_writer() << "No IPs are banned";
 
     return true;
 }
 
-
```
selsta: Please try to avoid unrelated whitespace changes.

sumogr: It was in excess. please check bool functions above and below

sumogr:
@selsta do you insist on restoring that whitespace that i think its in excess so as to approve or not this one as well ?

moneromooo-monero:
IMHO, extra restyling hunks are annoying. They spam the diff, making it harder to review, make git history harder to use, cause conflicts to whoever changes surrounding code. All for no good reason except "I like it better that way". I tend to let it go when it's a single line, especially when just removed, but in general, if you make a patch to do X, you make a patch to do X and not a patch to do X and reindent stuff.

This particular case here is pretty mild though, the above is just a general opinion on whitespace diffs :)

PR #6354, opened 2020-02-25
Add randomized delay when forwarding txes from i2p/tor -> ipv4/6

vtnerd:
The last bullet point on the CCS for transaction privacy is a randomized delay before a node takes an incoming transaction from i2p/tor and sends it over ipv4/6. My thought was an attempt to conceal the bandwidth/timing analysis of i2p/tor. If the the i2p/tor hidden service only has one incoming connection, this delay possibly/probably does nothing. But if the hidden service has 2+ incoming connections sending white noise, the transaction could've been received over any of the white noise connections.

This creates another mempool state.

@SarangNoether might be interested.

### DIFF
--- a/src/blockchain_db/blockchain_db.cpp
+++ b/src/blockchain_db/blockchain_db.cpp
@@ -63,6 +63,7 @@ bool matches_category(relay_method method, relay_category category) noexcept
   {
     default:
     case relay_method::local:
+    case relay_method::forward:
     case relay_method::stem:
       return false;
     case relay_method::block:
@@ -79,6 +80,7 @@ void txpool_tx_meta_t::set_relay_method(relay_method method) noexcept
   kept_by_block = 0;
   do_not_relay = 0;
   is_local = 0;
+  is_forwarding = 0;
   dandelionpp_stem = 0;
 
   switch (method)
@@ -89,8 +91,8 @@ void txpool_tx_meta_t::set_relay_method(relay_method method) noexcept
     case relay_method::local:
       is_local = 1;
       break;
-    default:
-    case relay_method::fluff:
+    case relay_method::forward:
+      is_forwarding = 1;
       break;
     case relay_method::stem:
       dandelionpp_stem = 1;
@@ -98,26 +100,45 @@ void txpool_tx_meta_t::set_relay_method(relay_method method) noexcept
     case relay_method::block:
       kept_by_block = 1;
       break;
+    default:
+    case relay_method::fluff:
+      break;
   }
 }
 
 relay_method txpool_tx_meta_t::get_relay_method() const noexcept
 {
-  if (kept_by_block)
-    return relay_method::block;
-  if (do_not_relay)
-    return relay_method::none;
-  if (is_local)
-    return relay_method::local;
-  if (dandelionpp_stem)
-    return relay_method::stem;
+  const uint8_t state =
+    uint8_t(kept_by_block) +
+    (uint8_t(do_not_relay) << 1) +
+    (uint8_t(is_local) << 2) +
+    (uint8_t(is_forwarding) << 3) +
+    (uint8_t(dandelionpp_stem) << 4);
+
+  switch (state)
+  {
+    default: // error case
+    case 0:
+      break;
+    case 1:
+      return relay_method::block;
+    case 2:
+      return relay_method::none;
+    case 4:
+      return relay_method::local;
+    case 8:
+      return relay_method::forward;
+    case 16:
+      return relay_method::stem;
+  };
   return relay_method::fluff;
 }
 
 bool txpool_tx_meta_t::upgrade_relay_method(relay_method method) noexcept
 {
   static_assert(relay_method::none < relay_method::local, "bad relay_method value");
-  static_assert(relay_method::local < relay_method::stem, "bad relay_method value");
+  static_assert(relay_method::local < relay_method::forward, "bad relay_method value");
+  static_assert(relay_method::forward < relay_method::stem, "bad relay_method value");
   static_assert(relay_method::stem < relay_method::fluff, "bad relay_method value");
   static_assert(relay_method::fluff < relay_method::block, "bad relay_method value");
 
--- a/src/blockchain_db/blockchain_db.h
+++ b/src/blockchain_db/blockchain_db.h
@@ -160,7 +160,7 @@ struct txpool_tx_meta_t
   uint64_t max_used_block_height;
   uint64_t last_failed_height;
   uint64_t receive_time;
-  uint64_t last_relayed_time; //!< If Dandelion++ stem, randomized embargo timestamp. Otherwise, last relayed timestmap.
+  uint64_t last_relayed_time; //!< If received over i2p/tor, randomized forward time. If Dandelion++stem, randomized embargo time. Otherwise, last relayed timestamp
   // 112 bytes
   uint8_t kept_by_block;
   uint8_t relayed;
@@ -169,7 +169,8 @@ struct txpool_tx_meta_t
   uint8_t pruned: 1;
   uint8_t is_local: 1;
   uint8_t dandelionpp_stem : 1;
-  uint8_t bf_padding: 4;
+  uint8_t is_forwarding: 1;
+  uint8_t bf_padding: 3;
 
   uint8_t padding[76]; // till 192 bytes
 
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1272,6 +1272,7 @@ namespace cryptonote
     {
       NOTIFY_NEW_TRANSACTIONS::request public_req{};
       NOTIFY_NEW_TRANSACTIONS::request private_req{};
+      NOTIFY_NEW_TRANSACTIONS::request stem_req{};
       for (auto& tx : txs)
       {
         switch (std::get<2>(tx))
@@ -1281,6 +1282,9 @@ namespace cryptonote
             break;
           case relay_method::local:
             private_req.txs.push_back(std::move(std::get<1>(tx)));
+            break;
+          case relay_method::forward:
+            stem_req.txs.push_back(std::move(std::get<1>(tx)));
             break;
           case relay_method::block:
           case relay_method::fluff:
@@ -1299,6 +1303,8 @@ namespace cryptonote
         get_protocol()->relay_transactions(public_req, source, epee::net_utils::zone::public_, relay_method::fluff);
       if (!private_req.txs.empty())
         get_protocol()->relay_transactions(private_req, source, epee::net_utils::zone::invalid, relay_method::local);
+      if (!stem_req.txs.empty())
+        get_protocol()->relay_transactions(stem_req, source, epee::net_utils::zone::public_, relay_method::stem);
     }
     return true;
   }
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -91,6 +91,8 @@ namespace cryptonote
     time_t const MAX_RELAY_TIME = (60 * 60 * 4); // at most that many seconds between resends
     float const ACCEPT_THRESHOLD = 1.0f;
 
+    constexpr const std::chrono::seconds forward_delay_average{CRYPTONOTE_FORWARD_DELAY_AVERAGE};
+
     // a kind of increasing backoff within min/max bounds
     uint64_t get_relay_delay(time_t now, time_t received)
     {
@@ -309,8 +311,14 @@ namespace cryptonote
 
         if (meta.upgrade_relay_method(tx_relay) || !existing_tx) // synchronize with embargo timer or stem/fluff out-of-order messages
         {
+          using clock = std::chrono::system_clock;
+          auto last_relayed_time = std::numeric_limits<decltype(meta.last_relayed_time)>::max();
+          if (tx_relay == relay_method::forward)
+            last_relayed_time = clock::to_time_t(clock::now() + crypto::random_poisson_seconds{forward_delay_average}());
+          // else the `set_relayed` function will adjust the time accordingly later
+
           //update transactions container
-          meta.last_relayed_time = std::numeric_limits<decltype(meta.last_relayed_time)>::max();
+          meta.last_relayed_time = last_relayed_time;
           meta.receive_time = receive_time;
           meta.weight = tx_weight;
           meta.fee = fee;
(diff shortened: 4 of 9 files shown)


src/blockchain_db/blockchain_db.h
```diff
@@ -160,15 +160,14 @@ struct txpool_tx_meta_t
...
   uint64_t receive_time;
-  uint64_t last_relayed_time;
+  uint64_t last_relayed_time; //!< If received over i2p/tor, randomized forward time. Otherwise, last relayed timestamp
   // 112 bytes
-  uint8_t kept_by_block;
+  uint8_t state;
   uint8_t relayed;
   uint8_t do_not_relay;
   uint8_t double_spend_seen: 1;
   uint8_t pruned: 1;
```
vtnerd: This drops a mempool meta bit flag that has never been in a release, but has been in `master`. I realized it was more efficient to use `kept_by_block` instead. Anyone running from `master` that has a `is_local` flag in their pool will see incorrect behavior when running this change. I think its better to clean this up before a release, but ... ?

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1472,7 +1500,8 @@ namespace cryptonote
     {
       const bool kept = pass == 1;
       bool r = m_blockchain.for_all_txpool_txes([this, &remove, kept](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata *bd) {
-        if (!!kept != !!meta.kept_by_block)
+        const bool kept_by_block = meta.get_relay_method() == relay_method::block;
+        if (!!kept != !!kept_by_block)
```
xiphon:
```suggestion
        if (kept != kept_by_block)
```

src/blockchain_db/blockchain_db.h
```diff
@@ -160,15 +160,14 @@ struct txpool_tx_meta_t
   uint64_t max_used_block_height;
   uint64_t last_failed_height;
   uint64_t receive_time;
-  uint64_t last_relayed_time;
+  uint64_t last_relayed_time; //!< If received over i2p/tor, randomized forward time. Otherwise, last relayed timestamp
   // 112 bytes
-  uint8_t kept_by_block;
```
xiphon: Seems reasonable to introduce 
```c++
bool kept_by_block() const noexcept;
```

moneromooo-monero: And grep well since anything that does "if (foo.kept_by_block)" will be turned into "if (true)".

src/blockchain_db/blockchain_db.h
moneromooo-monero: Are you saying kept_by_block is new ? It's not. Or are you talking about something else here ?

src/blockchain_db/blockchain_db.h
vtnerd: I will definitely make this change, provided the DB change to `master` is acceptable.

src/cryptonote_core/tx_pool.cpp
vtnerd: I decided to keep this as-is for whatever reason, but seems reasonable to change.

src/blockchain_db/blockchain_db.h
vtnerd: Didn't see any objections (or any comments at all), so this function was added and used where appropriate.

vtnerd:
Rebased, and made changes suggested by @xiphon .

src/blockchain_db/blockchain_db.cpp
```diff
@@ -77,35 +78,47 @@ bool matches_category(relay_method method, relay_category category) noexcept
...
+    default:
+    case 0:
+      break;
+    case 1:
+      return relay_method::block;
+    case uint8_t(relay_method::local) + 1:
+      return relay_method::local;
+    case uint8_t(relay_method::forward) + 1:
+      return relay_method::forward;
+   };
```
moneromooo-monero, line 118: tbh, this is really dodgy. Does this really need to be obfuscated like this ?

vtnerd: The code could continue using the inefficient bit mechanism, or could "break" the existing txpool between releases (this code breaks the txpool for users running master branch only).

moneromooo-monero: Why do we care about efficient bit twiddling when it's about I/O ?

vtnerd: Its less wasted space and less CPU cycles to just store/retrieve this value as the enum value.

moneromooo-monero: Well, whatever. If you prefer saving a few cycles hidden next to massive amounts of them for confusing code, fine.

vtnerd: No, my preference was to break the existing txpool (pain should be minimal) to do: `m_state = uint8_t(tx_relay);`.

I think in the end, I will drop this. This creates a hard dependency on the enum value and DB value. How about something like:

```c++
const uin8_t state =
  uint8_t(do_not_relay) +
  uint8_t(kept_by_block) << 1 +
  uint8_t(forward) << 2;

switch (state)
{
  default:
  case 0:
    return relay_method::fluff;
  case 1:
    return relay_method::none;
  case 2:
    return relay_method::block;
  case 4:
    return relay_method::forward:
}
```
which also handles the incorrect DB state, and maps it to fluff. Or the if-else ladder (which maps the incorrect DB state to some other value - but I doubt that matters anyway).

vtnerd:
Rebased and reverted to a bitfield. Reviewers should inspect new implementation of `get_relay_method()`.

src/blockchain_db/blockchain_db.cpp
moneromooo-monero: I much prefer this since I find it much clearer. It looks equivalent to the even simpler:

if (forward)
  return fluff;
if (kept_by_block)
  return block;
if (do_not_relay)
  return none;
return fluff;

moneromooo-monero: Amusingly, it looks close to what you had before. The thing I was finding objectionable was not the set of tests, it was the case on enum+1.

vtnerd: Yes, the original version was iffy because it required knowledge of the enum values to read it correctly. And a change in enum values could break the function. This version is similar looking, but everything is self-contained within the function.

FWIW, I disassembled this version and gcc 9 was able to optimize the function into a single comparison for the "error" case. Otherwise it used adds, shifts and a 64-bit constant to put the enum value in the return register. Roughly half of the instructions could be done in parallel by the CPU backend.

The if-ladder is less x64-64 instructions but 3-total comparison instructions. The compiler can still generate the same ASM with the switch version, but chose one with more math instructions and less comparisons.

moneromooo-monero: For clarity, I really don't care about a dozen cycles saved for this. What I care about is the complexity of the C code.

dEBRUYNE-1:
@vtnerd - Needs a rebase.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1282,6 +1283,9 @@ namespace cryptonote
           case relay_method::local:
             private_req.txs.push_back(std::move(std::get<1>(tx)));
             break;
+          case relay_method::forward:
```
vtnerd, line 1286: Another reasonable big change. Transactions in the forward state get relayed over Dandelion++ stems when the timer expires.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -730,20 +738,33 @@ namespace cryptonote
...
-        if (meta.dandelionpp_stem && meta.last_relayed_time < now) // for dandelion++ stem, this value is the embargo timeout
-          return true;
+        const relay_method tx_relay = meta.get_relay_method();
+        switch (tx_relay)
+        {
+          case relay_method::stem:
+          case relay_method::forward:
+            if (meta.last_relayed_time > now)
+              return true; // continue to next tx
+            break;
```
vtnerd, line 748: This code block here should block the PR from getting merged. My initial thought is to reset the timer for expired `forward` and `stem` txes transactions a bit further in the future to the standard (longer) re-relay timeout. I.e. when the initial forward/stem timer expires, the `txpool` needs to protect the txes from being relayed again too soon. Putting it here will guard against other changes to the codebase.

After Dandelion++, the `set_relayed` method is called asynchronously following a `get_relayable_transactions` call. The `get_relayable_transactions` method could be called again, and core could think it needs to relay `forward` transactions again 1 second later. This cannot happen with Dandelion++ stem, because the tx gets bumped to `fluff`, and is therefore marked as `fluff` synchronously (only Dandelion++ stem is delayed, as it asynchronously gets queued for inspecting local fluffing status).

src/cryptonote_core/tx_pool.cpp
```diff
@@ -309,8 +311,14 @@ namespace cryptonote
 
         if (meta.upgrade_relay_method(tx_relay) || !existing_tx) // synchronize with embargo timer or stem/fluff out-of-order messages
         {
+	  using clock = std::chrono::system_clock;
```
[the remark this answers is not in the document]
vtnerd: And @moneromooo-monero I ruined tabs again just for you. I switched to emacs recently :/ will fix.

vtnerd:
Tabs removed, and everything should go green. But see my own review above.

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -357,11 +357,15 @@ namespace levin
...
+        /* Always send with `fluff` flag, even over i2p/tor. The hidden service
+	   will disable the forwarding delay and immediately fluff. The i2p/tor
+	   network is therefore replacing the sybil protection of Dandelion++.
+	   Dandelion++ stem phase over i2p/tor is also worth investigating
+	   (with/without "noise"?). */
         for (auto& connection : connections)
         {
           std::sort(connection.first.begin(), connection.first.end()); // don't leak receive order
-          make_payload_send_txs(*zone_->p2p, std::move(connection.first), connection.second, zone_->pad_txs, zone_->is_public);
+          make_payload_send_txs(*zone_->p2p, std::move(connection.first), connection.second, zone_->pad_txs, true);
```
vtnerd, line 368: See the change here, and associated comment above.

tests/unit_tests/levin.cpp
```diff
@@ -1017,7 +1154,7 @@ TEST_F(levin_notify, private_fluff_without_padding)
             auto notification = receiver_.get_notification<cryptonote::NOTIFY_NEW_TRANSACTIONS>().second;
             EXPECT_EQ(txs, notification.txs);
             EXPECT_TRUE(notification._.empty());
-            EXPECT_FALSE(notification.dandelionpp_fluff);
+            EXPECT_TRUE(notification.dandelionpp_fluff);
```
vtnerd, line 1157: Changed within this PR (see other comment within this review in `levin_notify.cpp`). All similar changes were for same reason.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -722,28 +730,45 @@ namespace cryptonote
...
+        switch (tx_relay)
+        {
+          case relay_method::stem:
+          case relay_method::forward:
+            if (meta.last_relayed_time > now)
+              return true; // continue to next tx
+            change_timestamps.emplace_back(txid, meta);
+            break;
+          default:
+          case relay_method::none:
```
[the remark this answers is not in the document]
vtnerd: The `return true` after this chunk is within the callback given to `for_all_txpool_txes`, so it should continue to the next transaction. If you meant should the entire function unconditionally `return true`, there is a long standing `TODO` to investigate whether `true` should always be returned. I have yet to investigate this longstanding `TODO`.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -722,28 +730,45 @@ namespace cryptonote
   //TODO: investigate whether boolean return is appropriate
   bool tx_memory_pool::get_relayable_transactions(std::vector<std::tuple<crypto::hash, cryptonote::blobdata, relay_method>> &txs) const
   {
-    CRITICAL_REGION_LOCAL(m_transactions_lock);
-    CRITICAL_REGION_LOCAL1(m_blockchain);
+    std::vector<std::pair<crypto::hash, txpool_tx_meta_t>> change_timestamps;
     const uint64_t now = time(NULL);
     txs.reserve(m_blockchain.get_txpool_tx_count());
```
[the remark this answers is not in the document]
vtnerd: I don't see where the extra allocation would come from - the `std::vector` gets reserved either way. The only difference is that the allocation for `std::vector` occurs before acquiring the lock. This typically hits another resource atomic/lock thingy within the allocator anyway, so its probably not mattering much.

src/cryptonote_core/tx_pool.cpp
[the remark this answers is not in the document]
vtnerd: Sorry just got your comment, misread it previously. Yes, I will add an explicit `return` for the `none` label.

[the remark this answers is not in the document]
vtnerd: Although to clarify - `get_relay_method()` returns `relay_method::none` when `meta.do_not_relay == true` which is already filtered above, but another explicit check cannot hurt.

src/cryptonote_core/tx_pool.cpp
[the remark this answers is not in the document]
moneromooo-monero: Since the reserve is before the locks, another thread might add a tx to the txpool betewen the reserve and the locking.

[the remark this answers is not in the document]
moneromooo-monero: That said, it's trivial and of no consequence really. I'm ok with it as is too if there's even a slight reason for the move.

src/cryptonote_core/tx_pool.cpp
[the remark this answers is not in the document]
moneromooo-monero: If it's a case that cannot happen, then it's fine. I meant just to avoid going into the subsequent age test. I did not realize this case was impossible in practice.

src/cryptonote_core/tx_pool.cpp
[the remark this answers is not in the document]
vtnerd: This was a good catch. It was a call on `m_blockchain` directly, so I think its worth putting back inside the lock. Updated.

src/cryptonote_core/tx_pool.cpp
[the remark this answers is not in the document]
vtnerd: I added a `return true` for both `default` and `relay_method::none` cases. It seemed worth adding for the former, so why not.

PR #6357, opened 2020-02-26
daemon: auto public nodes - cache and prioritize most stable nodes

xiphon:
Improved public node selection algo used in `Simple mode`/`---bootstrap-daemon-address auto`.

1. In-memory public nodes cache storing up to N (N=1000 by default) entries
2. If the cache doesn't have at least one good public node (having fails count == 0):
    1. Get public nodes from P2P
    2. Add new white public nodes to cache with initial fails count = 0
    3. Add new gray public nodes to cache with initial fails count = 1
    4. Truncate the cache to N public nodes, keep the nodes having lower fails count
3. Select most stable public node (sorted by fails count), if there are multiple - choose a random one
4. Send an RPC request:
    1. On fail - increase the node's fails count by 2 and go to step 2
    2. On success - decrease the node's fails count by 2 and go to step 4

### DIFF
--- a/src/rpc/bootstrap_daemon.cpp
+++ b/src/rpc/bootstrap_daemon.cpp
@@ -2,6 +2,8 @@
 
 #include <stdexcept>
 
+#include <boost/thread/locks.hpp>
+
 #include "crypto/crypto.h"
 #include "cryptonote_core/cryptonote_core.h"
 #include "misc_log_ex.h"
@@ -12,15 +14,22 @@
 namespace cryptonote
 {
 
-  bootstrap_daemon::bootstrap_daemon(std::function<boost::optional<std::string>()> get_next_public_node)
-    : m_get_next_public_node(get_next_public_node)
+  bootstrap_daemon::bootstrap_daemon(
+    std::function<std::map<std::string, bool>()> get_public_nodes,
+    bool rpc_payment_enabled)
+    : m_selector(new bootstrap_node::selector_auto(std::move(get_public_nodes)))
+    , m_rpc_payment_enabled(rpc_payment_enabled)
   {
   }
 
-  bootstrap_daemon::bootstrap_daemon(const std::string &address, const boost::optional<epee::net_utils::http::login> &credentials)
-    : bootstrap_daemon(nullptr)
+  bootstrap_daemon::bootstrap_daemon(
+    const std::string &address,
+    boost::optional<epee::net_utils::http::login> credentials,
+    bool rpc_payment_enabled)
+    : m_selector(nullptr)
+    , m_rpc_payment_enabled(rpc_payment_enabled)
   {
-    if (!set_server(address, credentials))
+    if (!set_server(address, std::move(credentials)))
     {
       throw std::runtime_error("invalid bootstrap daemon address or credentials");
     }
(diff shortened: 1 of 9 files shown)


src/rpc/bootstrap_daemon.cpp
```diff
@@ -79,14 +93,17 @@ namespace cryptonote
...
 
-    const boost::optional<std::string> address = m_get_next_public_node();
-    if (address) {
-      return set_server(*address);
+    const auto node = [this]() {
+      const boost::unique_lock<boost::mutex> lock(m_selector_mutex);
+      return m_selector->next_node();
+    }();
+    if (node) {
+      return set_server(node->address, node->credentials);
```
moneromooo-monero: Is that a verbose version of:

const boost::unique_lock<boost::mutex> lock(m_selector_mutex);
const auto node = m_selector->next_node();
lock.unlock();
if (node) {
?

src/rpc/bootstrap_daemon.h
```diff
@@ -1,26 +1,34 @@
...
 namespace cryptonote
 {
 
   class bootstrap_daemon
   {
   public:
-    bootstrap_daemon(std::function<boost::optional<std::string>()> get_next_public_node);
-    bootstrap_daemon(const std::string &address, const boost::optional<epee::net_utils::http::login> &credentials);
+    bootstrap_daemon(
+      std::function<std::map<std::string, bool>()> get_public_nodes,
```
moneromooo-monero: could pass it by const ref, unless there's a good reason not to. I think these objects can get large if they've got capture data (not sure about the details though).

src/rpc/bootstrap_node_selector.cpp
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  void selector_auto::truncate()
+  {
+    auto &nodes_by_fails = m_nodes.get<by_fails>();
+    auto it = nodes_by_fails.end();
+    for (size_t total = m_nodes.size(); total > m_max_nodes; --total)
+    {
+      nodes_by_fails.erase(--it);
+    }
```
moneromooo-monero: That construct looks potentially fishy, as it might access an invalidated iterator. If you know it does not because you checked, then all is good. rbegin/++ might be what you're after ?

tests/unit_tests/bootstrap_node_selector.cpp
```diff
@@ -0,0 +1,172 @@
...
+      "gray_node_2:18089", false
+    }
+  };
+
+  std::map<std::string, bool> nodes;
+};
+
+TEST_F(bootstrap_node_selector, selector_auto_empty)
+{
+  cryptonote::bootstrap_node::selector_auto selector([this]() {
```
selsta:
```
/Users/selsta/dev/monero/tests/unit_tests/bootstrap_node_selector.cpp:64:55: warning: lambda capture 'this' is not used [-Wunused-lambda-capture]
  cryptonote::bootstrap_node::selector_auto selector([this]() {
                                                      ^~~~
```

xiphon: Fixed the warning

src/rpc/bootstrap_node_selector.cpp
xiphon: > If you know it does not because you checked, then all is good

Yep, it is covered with the unit tests added in this PR.

Anyway, updated the code with the cleaner version.

src/rpc/bootstrap_daemon.cpp
xiphon: Dropped the lambda

src/rpc/bootstrap_daemon.h
xiphon: Updated

xiphon: Decided to use move semantic instead.

PR #6358, opened 2020-02-26
blockchain_db: faster fetching of consecutive txes

moneromooo-monero:
Useful for wallet refresh or node sync

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -3107,6 +3107,104 @@ bool BlockchainLMDB::get_pruned_tx_blobs_from(const crypto::hash& h, size_t coun
   return true;
 }
 
+bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_count, size_t max_count, size_t max_size, std::vector<std::pair<std::pair<cryptonote::blobdata, crypto::hash>, std::vector<std::pair<crypto::hash, cryptonote::blobdata>>>>& blocks, bool pruned, bool skip_coinbase, bool get_miner_tx_hash) const
+{
+  LOG_PRINT_L3("BlockchainLMDB::" << __func__);
+  check_open();
+
+  TXN_PREFIX_RDONLY();
+  RCURSOR(blocks);
+  RCURSOR(tx_indices);
+  RCURSOR(txs_pruned);
+  if (!pruned)
+  {
+    RCURSOR(txs_prunable);
+  }
+
+  blocks.reserve(std::min<size_t>(max_count, 10000)); // guard against very large max count if only checking bytes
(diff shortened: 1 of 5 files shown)


src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -3107,6 +3107,103 @@ bool BlockchainLMDB::get_pruned_tx_blobs_from(const crypto::hash& h, size_t coun
...
+  for (uint64_t h = start_height; h < blockchain_height && blocks.size() < max_count && (size < max_size || blocks.size() < min_count); ++h)
+  {
+    MDB_cursor_op op = h == start_height ? MDB_SET : MDB_NEXT;
+    int result = mdb_cursor_get(m_cur_blocks, &key, &v, op);
+    if (result == MDB_NOTFOUND)
+      throw0(BLOCK_DNE(std::string("Attempt to get block from height ").append(boost::lexical_cast<std::string>(h)).append(" failed -- block not in db").c_str()));
+    else if (result)
+      throw0(DB_ERROR(lmdb_error("Error attempting to retrieve a block from the db", result).c_str()));
+
+    blocks.resize(blocks.size() + 1);
```
hyc: Incremental reallocs are slow and cause heap fragmentation. Since we have a maxsize parameter, should just size it to max before the loop, and trim down to actual size after end of loop if necessary.

selsta:
rescan_bc:

- master: 12:30
- 6358: 08:22

Does this match with your speed improvements @moneromooo-monero ?

selsta:
@vtnerd Would also be interesting to see what kind speed improvements you get with this PR + SuperCop ASM

moneromooo-monero:
I don't remember tbh. It'd be interesting to check with a HDD, these might see the biggest improvements if it avoids some seeks.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -3107,6 +3107,104 @@ bool BlockchainLMDB::get_pruned_tx_blobs_from(const crypto::hash& h, size_t coun
...
+      if (result)
+        throw0(DB_ERROR(lmdb_error("Error attempting to retrieve block coinbase transaction from the db: ", result).c_str()));
+
+      const txindex *tip = (const txindex *)v.mv_data;
+      tx_id = tip->data.tx_id;
+      val_tx_id.mv_data = &tx_id;
+      val_tx_id.mv_size = sizeof(tx_id);
+    }
+
+    if (skip_coinbase)
```
vtnerd, line 3166: How does this logic work if `!skip_coinbase` ? There doesn't appear to be call anywhere that does this, so I'm not sure if it was tested. The cursor should be unset on the call to `MDB_NEXT` below this. And the coinbase tx is never extracted.

PR #6364, opened 2020-02-29
device: Ledger - fix wide char hidapi error string conversion

xiphon:


src/device/device_io_hid.cpp
```diff
@@ -44,8 +44,15 @@ namespace hw {
...
-        const char* error_str = (const char*)hid_error(hwdev);
-        return  std::string(error_str == nullptr ? "Unknown error" : error_str);
+        const wchar_t* error_wstr = hid_error(hwdev);
+        if (error_wstr == nullptr)
+        {
+          return "Unknown error";
+        }
+        std::mbstate_t state{};
+        std::string error_str(std::wcsrtombs(nullptr, &error_wstr, 0, &state) + 1, 0);
+        std::wcsrtombs(&error_str[0], &error_wstr, error_str.size(), &state);
```
moneromooo-monero: If the call fails, this calls &error_str[0] on a string of size 0 (return -1, adds 1). I'm not sure this is valid.

xiphon: Added explicit error check

PR #6370, opened 2020-03-05
Fix compatibility with boost < 1.64

omartijn:
The reason for this is that we use `boost::asio::ssl::context::tls`,
which was introduced in version 1.64, the current master will not
compile with an older boost version.

Additionally, the check in CMake for boost version smaller than 1.62 was
removed, since this no longer applies (the configure would have failed
because the required minimum version was not found).

The instructions for building boost from source on debian Jessie have
been slightly changed, with the old instructions we would have tried to
remove all installed boost packages - which are also used by other
packages installed on the system. Instead we now only remove the header
files.

The downloaded version for boost has also been bumped to 1.72. If we are
building boost from source, there is no reason not to use the most
recent version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,13 +332,13 @@ If you are using the older Raspbian Jessie image, compiling Monero is a bit more
 
 * Then, install the dependencies for Monero except `libunwind` and `libboost-all-dev`
 
-* Install the latest version of boost (this may first require invoking `apt-get remove --purge libboost*` to remove a previous version if you're not using a clean install):
+* Install the latest version of boost (this may first require invoking `apt-get remove --purge libboost*-dev` to remove a previous version if you're not using a clean install):
 
     ```bash
     cd
-    wget https://sourceforge.net/projects/boost/files/boost/1.64.0/boost_1_64_0.tar.bz2
-    tar xvfo boost_1_64_0.tar.bz2
-    cd boost_1_64_0
+    wget https://sourceforge.net/projects/boost/files/boost/1.72.0/boost_1_72_0.tar.bz2
+    tar xvfo boost_1_72_0.tar.bz2
+    cd boost_1_72_0
     ./bootstrap.sh
     sudo ./b2
     ```
--- a/contrib/epee/src/net_ssl.cpp
+++ b/contrib/epee/src/net_ssl.cpp
@@ -289,7 +289,9 @@ ssl_options_t::ssl_options_t(std::vector<std::vector<std::uint8_t>> fingerprints
 
 boost::asio::ssl::context ssl_options_t::create_context() const
 {
-  boost::asio::ssl::context ssl_context{boost::asio::ssl::context::tls};
+  // note: this enables a lot of old and insecure protocols, which we
+  // promptly disable below - if the result is actually used
+  boost::asio::ssl::context ssl_context{boost::asio::ssl::context::sslv23};
   if (!bool(*this))
     return ssl_context;
 


xiphon:
> The reason for this is that we use `boost::asio::ssl::context::ssl`

Could you point to the corresponding line/lines?

omartijn:
I see that I made a typo in the description, we are using `boost::asio::ssl::context::tls`, I have edited this now. It's used in [contrib/epee/src/net_ssl.cpp:292](https://github.com/monero-project/monero/blob/master/contrib/epee/src/net_ssl.cpp#L292).

vtnerd:
Ubuntu 16.04 has not reached EOL yet and is on Boost 1.58. OTOH, those users may want to upgrade for bug fixes, etc., anyway.

omartijn:
@vtnerd If we cannot upgrade the required boost version we must do something about using `boost::asio::ssl::context::tls`.

The enum just maps to some OpenSSL option that is set on the SSL_CTX, so we could of course set that ourselves. That would work with older versions of boost as well.

Without either fixing that, or upgrading boost, the current master won't build on Ubuntu 16.04.

sumogr:
By searching I found a workaround that maybe so "stupid" it might work, set it from the beginning at sslv23 which is wrong and we dont want (but is accepted by old boost versions) which will be avoided by the configure options below. sslv2, sslv3 are disallowed as well as tlsv1 and tlsv1.1 so tlsv1.2 is what we upgrade it to again
```
  boost::asio::ssl::context ssl_context{boost::asio::ssl::context::sslv23};
  if (!bool(*this))
    return ssl_context;

  // only allow tls v1.2 and up
  ssl_context.set_options(boost::asio::ssl::context::default_workarounds);
  ssl_context.set_options(boost::asio::ssl::context::no_sslv2);
  ssl_context.set_options(boost::asio::ssl::context::no_sslv3);
  ssl_context.set_options(boost::asio::ssl::context::no_tlsv1);
  ssl_context.set_options(boost::asio::ssl::context::no_tlsv1_1);
```

omartijn:
@sumogr That's interesting. Looking through the asio code, we can see that the `method` given to the constructor of the `boost::asio::ssl::context` is used to determine which method to use for constructing the handle.

For method `tls` it will use `TLS_method()` while for `sslv23` it will use `SSLv23_method`. The latter is defined to be the same since OpenSSL 1.1.0, which contains this line:

`#define SSLv23_method           TLS_method`

Looking through older OpenSSL releases, SSLv23_method is still its own proper function, but it's documented as "choosing the highest supported version, allowing fallback to older versions". If we then explicitly disable all these insecure methods, it should yield the correct behaviour even on older OpenSSL versions (pre 1.1.0).

I can rewrite the PR to do this instead of bumping the required boost version. Is this preferred over bumping the boost version? The older Ubuntu version (16.04) is, as @vtnerd mentions not EOL yet and is supported by canonical until April 2021.

selsta:
What would the downside be of doing a simple

`#if boost_version > ...`

with fallback to previous behavior?

sumogr:
> 
> 
> @sumogr That's interesting. Looking through the asio code, we can see that the `method` given to the constructor of the `boost::asio::ssl::context` is used to determine which method to use for constructing the handle.
> 
> For method `tls` it will use `TLS_method()` while for `sslv23` it will use `SSLv23_method`. The latter is defined to be the same since OpenSSL 1.1.0, which contains this line:
> 
> `#define SSLv23_method TLS_method`
> 
> Looking through older OpenSSL releases, SSLv23_method is still its own proper function, but it's documented as "choosing the highest supported version, allowing fallback to older versions". If we then explicitly disable all these insecure methods, it should yield the correct behaviour even on older OpenSSL versions (pre 1.1.0).
> 
> I can rewrite the PR to do this instead of bumping the required boost version. Is this preferred over bumping the boost version? The older Ubuntu version (16.04) is, as @vtnerd mentions not EOL yet and is supported by canonical until April 2021.

IMHO just set the method to sslv23 we are already disallowing sslv2 and sslv3 as well tls v1 and 1.1. 1.2 and up is the only thing remaining.
I ve found a similar workaround as well by SkyllaDB (cloudious systems) so it should be working https://github.com/cloudius-systems/osv/commit/272325d1b556ba468db8895368b381a8e807dc3d

omartijn:
I've undone the required version bump and implemented the backwards-compatible workaround suggested by @sumogr.

PR #6376, opened 2020-03-07
Added logging for dropped local txes with no i2p/tor connections (0.15 branch)

vtnerd:


vtnerd:
Identical to PR 6375 but for 0.15 branch.

vtnerd:
I force pushed a change just after your comment. This adds a log statement for public networks and anonymity networks when noise is disabled. This differs from the patch to master, because this branch does not have "diffusion" - it is using the simpler "flood" algorithm for notification.

rbrunner7:
Well, after checking again and also comparing with the code changes for the `master` branch I have to admit maybe I was a little fast and not the right person to review this after all ...

PR #6388, opened 2020-03-14
workaround for GetObject windows.h macro and rapidjson conflict

xiphon:
Quick and dirty workaround. Could be fixed in other ways at cost of additional changes.

### DIFF
--- a/src/serialization/json_object.h
+++ b/src/serialization/json_object.h
@@ -30,11 +30,11 @@
 
 #include <boost/utility/string_ref.hpp>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


vtnerd:
This conflicts with PR 6351 which is removing `#include "string_tools.h"` entirely. I guess one of the other headers is still pulling in `windows.h`.

I love how this fixes the issue - every `GetObject` token in the rapidjson source becomes `GetObjectA`. Which is apparently how this worked the whole time.

selsta:
@vtnerd Are you okay with merging this first and you rebasing #6351 afterwards? I would like to get CI fixed quickly.

vtnerd:
No one has even approved that PR, so its inevitable that I will be forced to rebase.

Issue #6396, opened 2020-03-20
./monero-wallet-rpc: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by ./monero-wallet-rpc)


GongSuiLi:
When I start monero-wallet-rpc v0.15.0.5 use ./monero-wallet-rpc --config-file= ,connect monerod failed, return ./monero-wallet-rpc: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by ./monero-wallet-rpc).

OS: centos7
version: v0.15.0.5

dEBRUYNE-1:
Similar issue reported here:

https://www.reddit.com/r/monerosupport/comments/flq740/error_with_01505_cli_and_ubuntu_1604/

selsta:
Did v0.15.0.1 start correctly?

xiphon:
Reproducible binaries don't support anything older than Ubuntu bionic 18.04.

If you are using an old OS, the best way would be to build Monero from sources following the manual https://github.com/monero-project/monero#compiling-monero-from-source

xiphon:
Ignore my previous comment, while inspecting the binaries i misread `gnu_get_libc_version@@GLIBC_2.2.5` as `GLIBC_2.25`.

xiphon:
The reason is bumped libsodium `1.0.18` version. Libsodium uses `getrandom(...)` which has been added in glibc `2.25`.

sumogr:
> The reason is bumped libsodium `1.0.18` version. Libsodium uses `getrandom(...)` which has been added in glibc `2.25`.

It might be one of the reasons. make release-static fails as well on xenial
I think its a python3 issue

GongSuiLi:
> Did v0.15.0.1 start correctly?

v0.15.0.0 start correctly

BKdilse:
same issue for me.  15.0.1 worked fine, but 15.0.5 causing this issue.

selsta:
We could move back to libsodium 1.0.16, there is still a mirror on Github: https://github.com/jedisct1/libsodium/releases/download/1.0.16/libsodium-1.0.16.tar.gz

@TheCharlatan what is your opinion?

TheCharlatan:
I should have checked for version increases, sorry everybody! I'll work on tooling to catch this in the future. It's probably easiest to revert the version, yes. It's not just the rpc client effected, it's the wallet as well. Alternatively we could see if we can do some clever patching with it, but this will require some time.

[...]

PR #6397, opened 2020-03-20
depends: libsodium disable getrandom()/getentropy() (glibc < 2.25)

xiphon:
Resolves https://github.com/monero-project/monero/issues/6396

### DIFF
--- a/contrib/depends/packages/sodium.mk
+++ b/contrib/depends/packages/sodium.mk
@@ -3,7 +3,7 @@ $(package)_version=1.0.18
 $(package)_download_path=https://download.libsodium.org/libsodium/releases/
 $(package)_file_name=libsodium-$($(package)_version).tar.gz
 $(package)_sha256_hash=6f504490b342a4f8a4c4a02fc9b866cbef8622d5df4e5452b46be121e46636c1
-$(package)_patches=fix-whitespace.patch
+$(package)_patches=disable-glibc-getrandom-getentropy.patch fix-whitespace.patch
 
 define $(package)_set_vars
 $(package)_config_opts=--enable-static --disable-shared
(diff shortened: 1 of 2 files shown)


contrib/depends/patches/sodium/disable-glibc-getrandom-getentropy.patch
```diff
@@ -0,0 +1,25 @@
...
++
+ unsigned char buf;
+ (void) getrandom((void *) &buf, 1U, 0U);
+   ]])],
+@@ -825,6 +829,9 @@ unsigned char buf;
+ # include <sys/random.h>
+ #endif
+ ]], [[
++#ifdef __linux__
++# error getentropy() is currently disabled on Linux to support glibc < 2.25
```
TheCharlatan: Are we linking `getentropy` as well? I could only find calls to `getrandom` in the symbol table.

xiphon: We don't invoke any libsoduim code that requires `getentropy` right now.

But `getentropy` has been added in glibc 2.25 too. Disabled it as a precaution. 

Thoughts?

TheCharlatan: I see, then it looks good to me, seems like all the HAVE_... guards are in place in libsodium.

contrib/depends/packages/sodium.mk
```diff
@@ -15,6 +15,7 @@ endef
 define $(package)_config_cmds
   ./autogen.sh &&\
   patch -p1 < $($(package)_patch_dir)/fix-whitespace.patch &&\
+  patch -p1 < $($(package)_patch_dir)/disable-glibc-getrandom-getentropy.patch &&\
```
TheCharlatan: You need to include the patch in the `$(package)_patches` section above.

xiphon: Done

TheCharlatan: The patch needs to be before autogen, otherwise it does not reach the configure script. So move it one line up (but keep the other patch where it is).

xiphon: Yeah

PR #6398, opened 2020-03-20
depends: update qt 5.7.1 download link

sumogr:
5.7.1 was removed from official qt.io archive, this is one of the few mirrors remaining that keep a copy.
I think it would be better if @TheCharlatan updated to a newer version soon than merging this PR


selsta:
ping @TheCharlatan 

should we merge this PR or update qt?

PR #6401, opened 2020-03-24
BlockchainDB Rework transactions to improve thread safety

omartijn:
This is a large PR, which fixes a number of issues.

- the mdb_cursors struct was cast to an array, which is technically undefined behaviour
- the m_writer variable was not atomic, nor protected by a lock, resulting in possible race conditions
- some functions called mutliple functions without an overarching transaction (inconsistent view)

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -70,14 +70,14 @@ struct pre_rct_output_data_t
 #pragma pack(pop)
 
 template <typename T>
-inline void throw0(const T &e)
+[[noreturn]] inline void throw0(const T &e)
 {
   LOG_PRINT_L0(e.what());
   throw e;
 }
 
 template <typename T>
-inline void throw1(const T &e)
+[[noreturn]] inline void throw1(const T &e)
 {
   LOG_PRINT_L1(e.what());
   throw e;
@@ -95,41 +95,13 @@ struct MDB_val_copy: public MDB_val
   MDB_val_copy(const T &t) :
     t_copy(t)
   {
+    static_assert(std::is_integral<T>::value, "T must be an integral type");
+
     mv_size = sizeof (T);
     mv_data = &t_copy;
   }
 private:
   T t_copy;
-};
-
-template<>
-struct MDB_val_copy<cryptonote::blobdata>: public MDB_val
-{
-  MDB_val_copy(const cryptonote::blobdata &bd) :
-    data(new char[bd.size()])
-  {
-    memcpy(data.get(), bd.data(), bd.size());
-    mv_size = bd.size();
-    mv_data = data.get();
-  }
-private:
-  std::unique_ptr<char[]> data;
-};
-
-template<>
-struct MDB_val_copy<const char*>: public MDB_val
-{
-  MDB_val_copy(const char *s):
-    size(strlen(s)+1), // include the NUL, makes it easier for compares
-    data(new char[size])
-  {
-    mv_size = size;
-    mv_data = data.get();
-    memcpy(mv_data, s, size);
-  }
-private:
-  size_t size;
-  std::unique_ptr<char[]> data;
 };
 
 }
@@ -212,31 +184,6 @@ namespace
  *
  * The output_amounts table doesn't use a dummy key, but uses DUPSORT.
  */
-const char* const LMDB_BLOCKS = "blocks";
-const char* const LMDB_BLOCK_HEIGHTS = "block_heights";
-const char* const LMDB_BLOCK_INFO = "block_info";
-
-const char* const LMDB_TXS = "txs";
-const char* const LMDB_TXS_PRUNED = "txs_pruned";
-const char* const LMDB_TXS_PRUNABLE = "txs_prunable";
-const char* const LMDB_TXS_PRUNABLE_HASH = "txs_prunable_hash";
-const char* const LMDB_TXS_PRUNABLE_TIP = "txs_prunable_tip";
-const char* const LMDB_TX_INDICES = "tx_indices";
-const char* const LMDB_TX_OUTPUTS = "tx_outputs";
-
-const char* const LMDB_OUTPUT_TXS = "output_txs";
-const char* const LMDB_OUTPUT_AMOUNTS = "output_amounts";
-const char* const LMDB_SPENT_KEYS = "spent_keys";
-
-const char* const LMDB_TXPOOL_META = "txpool_meta";
-const char* const LMDB_TXPOOL_BLOB = "txpool_blob";
-
-const char* const LMDB_ALT_BLOCKS = "alt_blocks";
-
-const char* const LMDB_HF_STARTING_HEIGHTS = "hf_starting_heights";
-const char* const LMDB_HF_VERSIONS = "hf_versions";
-
-const char* const LMDB_PROPERTIES = "properties";
 
 const char zerokey[8] = {0};
 const MDB_val zerokval = { sizeof(zerokey), (void *)zerokey };
@@ -257,25 +204,10 @@ inline void lmdb_db_open(MDB_txn* txn, const char* name, int flags, MDB_dbi& dbi
 }  // anonymous namespace
 
 #define CURSOR(name) \
-	if (!m_cur_ ## name) { \
-	  int result = mdb_cursor_open(*m_write_txn, m_ ## name, &m_cur_ ## name); \
-	  if (result) \
-        throw0(DB_ERROR(lmdb_error("Failed to open cursor: ", result).c_str())); \
-	}
+    auto m_cur_ ## name = m_write_txn.cursor(source::name);
 
 #define RCURSOR(name) \
-	if (!m_cur_ ## name) { \
-	  int result = mdb_cursor_open(m_txn, m_ ## name, (MDB_cursor **)&m_cur_ ## name); \
-	  if (result) \
-        throw0(DB_ERROR(lmdb_error("Failed to open cursor: ", result).c_str())); \
-	  if (m_cursors != &m_wcursors) \
-	    m_tinfo->m_ti_rflags.m_rf_ ## name = true; \
-	} else if (m_cursors != &m_wcursors && !m_tinfo->m_ti_rflags.m_rf_ ## name) { \
-	  int result = mdb_cursor_renew(m_txn, m_cur_ ## name); \
-      if (result) \
-        throw0(DB_ERROR(lmdb_error("Failed to renew cursor: ", result).c_str())); \
-	  m_tinfo->m_ti_rflags.m_rf_ ## name = true; \
-	}
+    auto m_cur_ ## name = m_txn.cursor(source::name);
 
 namespace cryptonote
 {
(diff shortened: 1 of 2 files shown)


selsta:
Nice PR! Can you look into why it didn’t compile on WIndows CI?
```
In file included from D:/a/monero/monero/src/blockchain_db/blockchain_db.cpp:37:
D:/a/monero/monero/src/blockchain_db/lmdb/db_lmdb.h:193:12: error: 'atomic' in namespace 'boost' does not name a template type
  193 |     boost::atomic<boost::thread::id> m_writer_thread;
      |            ^~~~~~
D:/a/monero/monero/src/blockchain_db/lmdb/db_lmdb.h: In constructor 'cryptonote::mdb_writer_data::mdb_writer_data(MDB_env*, cryptonote::mdb_databases&)':
D:/a/monero/monero/src/blockchain_db/lmdb/db_lmdb.h:170:7: error: class 'cryptonote::mdb_writer_data' does not have any field named 'm_writer_thread'
  170 |       m_writer_thread{},
      |       ^~~~~~~~~~~~~~~
[ 62%] Building CXX object src/cryptonote_core/CMakeFiles/obj_cryptonote_core.dir/cryptonote_tx_utils.cpp.obj
852
```

omartijn:
@selsta Looks like the header dependencies for Linux/Windows are different somehow so that under Linux `boost::atomic` was available and under Windows it wasn't.

I've added an explicit include now which should fix it. If there are no other issues I'll squash the commits. Unfortunately I don't have a Windows system at hand so the only way to test it is through the CI.

selsta:
Yep looks good now.

omartijn:
Squashed the fix

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -1577,44 +1830,28 @@ void BlockchainLMDB::reset()
...
+  m_databases.drop(txn, source::block_heights);
+  m_databases.drop(txn, source::txs_pruned);
+  m_databases.drop(txn, source::txs_prunable);
+  m_databases.drop(txn, source::txs_prunable_hash);
+  m_databases.drop(txn, source::txs_prunable_tip);
+  m_databases.drop(txn, source::tx_indices);
+  m_databases.drop(txn, source::tx_outputs);
+  m_databases.drop(txn, source::output_txs);
+  m_databases.drop(txn, source::output_amounts);
+  m_databases.drop(txn, source::spent_keys);
```
moneromooo-monero: Are those actually necessary ? They cause a very large amount of possibly spurious diffs.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -460,49 +471,279 @@ void mdb_txn_safe::allow_new_txns()
...
+  m_writer_thread.store(boost::this_thread::get_id(), boost::memory_order_relaxed);
+  m_writer_depth = 1;
+
+  return m_txn;
+}
+
+void mdb_writer_data::release(bool error)
+{
+  if (current_thread() != boost::this_thread::get_id())
+    throw1(DB_ERROR{"transaction owned by other threayyd"});
```
selsta: typo?

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -460,49 +471,279 @@ void mdb_txn_safe::allow_new_txns()
...
+{
+  return *m_cursors;
+}
 
-  mdb_env_info(env, &mei);
-  uint64_t new_mapsize = mei.me_mapsize;
+MDB_cursor* mdb_writer_data::cursor(source database)
+{
+  if (current_thread() != boost::this_thread::get_id())
+    throw1(DB_ERROR{"transaction owned by other threayyd"});
```
selsta: also here?

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -351,98 +283,177 @@ typedef struct outtx {
     uint64_t local_index;
 } outtx;
 
+const std::string& mdb_databases::name(source database)
+{
+  static std::array<std::string, 18> names{
```
vtnerd: Do these need to be `std::string`? This could be `const char*` with the entire array marked as `constexpr`. The return value would be `const char*` too, I'm not seeing `std::string` required usage anywhere. Its looks like its being used either as a C-string to LMDB or an error temp `std::string`.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -351,98 +283,177 @@ typedef struct outtx {
...
+void lmdb_resized(MDB_env *env)
 {
-  MDB_cursor **cur = &m_ti_rcursors.m_txc_blocks;
-  unsigned i;
-  for (i=0; i<sizeof(mdb_txn_cursors)/sizeof(MDB_cursor *); i++)
-    if (cur[i])
-      mdb_cursor_close(cur[i]);
-  if (m_ti_rtxn)
-    mdb_txn_abort(m_ti_rtxn);
+  mdb_txn_safe::prevent_new_txns();
```
vtnerd: Since your cleaning this up already ... several lines can throw (including each log statement), making this call exception unsafe. There's `epee::on_scoped_exit` (which can be re-written to not allocate) that would do that, or a couple of other usual techniques.

src/blockchain_db/lmdb/db_lmdb.cpp
vtnerd: Yes, there appears to be lots of re-written code unrelated to things commented on above. The re-writes appear to be mostly good. I've tended to avoid churning/rewriting the DB code because of the risk, but it could use some love. The refactoring appears to be "on the edges" so to speak, and not modifying the bulk of the processing code.

src/blockchain_db/lmdb/db_lmdb.h
```diff
@@ -32,6 +32,7 @@
 #include "cryptonote_basic/blobdatatype.h" // for type blobdata
 #include "ringct/rctTypes.h"
 #include <boost/thread/tss.hpp>
+#include <boost/atomic/atomic.hpp>
```
vtnerd: Just use `std::atomic` for this.

src/blockchain_db/lmdb/db_lmdb.h
```diff
@@ -45,126 +46,215 @@ typedef struct txindex {
...
+      for (auto cursor: m_cursors) {
+        if (cursor != nullptr) {
+          mdb_cursor_close(cursor);
+        }
+      }
+    }
+
+    MDB_cursor*& get(source cursor)
+    {
+      return m_cursors.at(static_cast<uint8_t>(cursor));
```
vtnerd: Can save some typing in the future by doing `uint8_t(cursor)`.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -460,49 +471,279 @@ void mdb_txn_safe::allow_new_txns()
...
+  LOG_PRINT_L3("mdb_writer_data: acquire()");
+
+  if (auto txn = try_acquire())
+    return txn;
+
+  while (mdb_txn_safe::creation_gate.test_and_set());
+  mdb_txn_safe::num_active_txns++;
+  mdb_txn_safe::creation_gate.clear();
+
+  auto result = lmdb_txn_begin(m_env, nullptr, 0, &m_txn);
```
vtnerd: Should a temp value `m_txn` value be given to this function? It seems fairly iffy - a second thread is blocked inside of the (eventual) LMDB call but is using the variable that the active writer thread is using. It _probably_ works, assuming LMDB has a full-fence inside somewhere, but ...

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -460,49 +471,279 @@ void mdb_txn_safe::allow_new_txns()
...
 }
 
-inline int lmdb_txn_renew(MDB_txn *txn)
+mdb_write_handle::~mdb_write_handle()
 {
-  int res = mdb_txn_renew(txn);
-  if (res == MDB_MAP_RESIZED) {
-    lmdb_resized(mdb_txn_env(txn));
-    res = mdb_txn_renew(txn);
+  if (m_data == nullptr || m_txn == nullptr)
```
vtnerd: I notice a `std::current_exception()` which clearly indicates this is designed to grab errors as the stack is being unwound. But this function also potentially throws on the last two lines, so this will need to be an empty `try{} catch {}` to prevent an immediate process abortion.

src/blockchain_db/lmdb/db_lmdb.cpp
hyc: LMDB write txns are fully serialized, there can be no race here.

vtnerd: I know they are serialized, my comment was on re-using the same `&m_txn` in use by the other thread. If LMDB "touches" the memory for this value before the other thread releases the writer lock, then its racy. Maybe LMDB doesn't do that, but this seems to rely on a specific internal implementation.

hyc: You have just described a race condition. I have just told you there is no race condition; that is a contract of the LMDB API.

src/blockchain_db/lmdb/db_lmdb.cpp
omartijn: This one isn't strictly necessary, but given that the _cursors_ have been changed from a struct to an actual array (to fix the undefined behaviour caused from the cast that happened there), I believe it makes sense to change this one as well to keep some uniformity between working with databases and cursors.

Also: The old code had a lot of repeated "open the database, do error handling"-code. To me, at least, abstracting that logic out is an improvement in its own right. If that is not an opinion you and others share with me I can of course undo this change.

src/blockchain_db/lmdb/db_lmdb.cpp
omartijn: Indeed, fixed.

src/blockchain_db/lmdb/db_lmdb.cpp
omartijn: Indeed, fixed.

src/blockchain_db/lmdb/db_lmdb.cpp
omartijn: It's not really necessary indeed. I did it this way mainly because it's only initialized _once_ in any case (so there shouldn't be a measurable run-time overhead. I'll change the array type to a `const char*`.

src/blockchain_db/lmdb/db_lmdb.cpp
omartijn: The simplest (and safest) way is probably to use an `std::mutex` and `std::unique_lock` (or their boost equivalents). That way the destructor of `std::unique_lock` will ensure the lock is released.

Ignoring the possible exceptions thrown from logging, however, I wonder what the correct course of action would be here. If the environment is resized and `mdb_env_set_mapsize` fails then we cannot use lmdb at that point (because another process may be writing outside of our allotted map size). This would mean that even if we correctly release the lock here, we'd be left in an unusable state (the next call will probably fail again with MDB_MAP_RESIZED).

Does it make sense to retry a failing `mdb_env_set_mapsize`?

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -351,98 +283,177 @@ typedef struct outtx {
...
+  mdb_txn_safe::prevent_new_txns();
+
+  MGINFO("LMDB map resize detected.");
+
+  MDB_envinfo mei;
+
+  mdb_env_info(env, &mei);
+  uint64_t old = mei.me_mapsize;
+
+  mdb_txn_safe::wait_no_active_txns();
```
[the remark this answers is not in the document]
omartijn: I'm unsure why you're referencing this line specifically. The `wait_no_active_txns` doesn't _prevent_ anything, it merely waits until all transactions stop on their own accord. How can that be transaction unsafe? The `wait_no_active_txns` could easily be marked `noexcept` itself (just as all the remaining member functions on `mdb_txn_safe` for that matter).

src/blockchain_db/lmdb/db_lmdb.h
```diff
@@ -45,126 +46,215 @@ typedef struct txindex {
...
-    return &m_txn;
-  }
+  MDB_txn* acquire(MDB_env* env);
+  void release();
 
-  uint64_t num_active_tx() const;
+  MDB_txn *m_ti_rtxn;	// per-thread read txn
+  mdb_txn_cursors m_ti_rcursors;	// per-thread read cursors
+  bool m_rf_txn;
+  std::bitset<18> m_valid_cursor;
```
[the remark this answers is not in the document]
omartijn: No. There are basically three states when it comes to read cursors:

1. Not initialized (a nullptr)
2. Created using the current transaction (not a nullptr)
3. Created using an aborted transaction (not a nullptr)

State 3 is useful because it allows us to _re-use_ a cursor that was previously allocated for us by LMDB. However, since 2 and 3 both have a read cursor that is not a nullptr we need something to distinguish between these two states.

From [the LMDB documentation for mdb_cursor_renew](http://www.lmdb.tech/doc/group__mdb.html#gac8b57befb68793070c85ea813df481af):

> Renew a cursor handle.
> 
> A cursor is associated with a specific transaction and database. Cursors that are only used in read-only transactions may be re-used, to avoid unnecessary malloc/free overhead. The cursor may be associated with a new read-only transaction, and referencing the same database handle as it was created with. This may be done whether the previous transaction is live or dead.

This feature was already used by the `blockchain_db`, however, since I needed to change the `mdb_txn_cursors` struct to use a real array to fix the UB, I also changed the read flags to a real array (in this case an `std::bitset` because it's more-or-less an array of bools).

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -460,49 +471,279 @@ void mdb_txn_safe::allow_new_txns()
...
+}
+
+MDB_txn* mdb_writer_data::acquire()
+{
+  LOG_PRINT_L3("mdb_writer_data: acquire()");
+
+  if (auto txn = try_acquire())
+    return txn;
+
+  while (mdb_txn_safe::creation_gate.test_and_set());
```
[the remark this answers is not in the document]
omartijn: It used to be part of the `mdb_txn_safe` constructor, which is no longer used. I've abstracted it into its own function and added it to read transactions as well.

src/blockchain_db/lmdb/db_lmdb.h
omartijn: Sounds like a plan. This does yield the question of when and where we use boost-specific equivalents over the standard library. There is quite a lot of code using `boost::shared_ptr`, for example, which might just as well use `std::shared_ptr`. Is this only for legacy reasons?

I know that `boost::thread` is an exception, because it has some functionality that `std::thread` does not, which we actively use. Is that the case for other things as well?

omartijn: Note: I had to also change `boost::thread::id` to `std::thread::id` to be able to use it with `std::atomic`.

src/blockchain_db/lmdb/db_lmdb.h
omartijn: I was unaware of this syntax. It does sound useful indeed. I'll remember this for when I need it.

src/blockchain_db/lmdb/db_lmdb.cpp
omartijn: Interesting point. Isn't the fact that logging can throw a potentially really big problem in our code-base? We do seem to log from destructors in many other places which basically means these can potentially throw as well, which is rather iffy.

I fixed it here, also because the `release()` can throw if committing the transaction fails.

src/blockchain_db/lmdb/db_lmdb.h
hyc: The main reason we used boost::thread, originally, was because std::thread was broken on Windows, back in gcc 4.x days. I'm not sure that reason still applies.

src/blockchain_db/lmdb/db_lmdb.h
hyc: Was unfamiliar with that myself. Not sure it's worth using something that the rest of people don't recognize.

src/blockchain_db/lmdb/db_lmdb.cpp
moneromooo-monero: Where is that cast to array ?

src/blockchain_db/lmdb/db_lmdb.cpp
vtnerd: Moo has been adding empty catch blocks in some places. It kind of stinks because you can't really even log, unless its a double catch or a `noexcept` function. The more likely case is a throw from `release(...)` though.

vtnerd: I never answered your first question. Yes, a `noexcept` log function could be useful because its more intuitive even if logging in a destructor fails. Big uphill battle on that  though.

vtnerd: I take that back, it should be a matter of wrapping the call in a catch block. The issue is that it would return an error code that isn't checked anywhere. And there are cases where nothing could be done on error anyway (which is basically what this is about).

src/blockchain_db/lmdb/db_lmdb.cpp
vtnerd: The documentation doesn't indicate when the out pointer is touched. Sounds like its guaranteed to be after any other thread has completed its work.

src/blockchain_db/lmdb/db_lmdb.h
vtnerd: Might want to continue using `boost::atomic<boost::thread::id>` then.

src/blockchain_db/lmdb/db_lmdb.h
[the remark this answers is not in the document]
vtnerd: I know how the cursors work, but I hadn't read the relevant code that used this, so made the cardinal sin of commenting too early.

src/blockchain_db/lmdb/db_lmdb.cpp
vtnerd: The scheme is actually a reader_writer_lock, but it might take some refactoring to get there. This is also (basically) guaranteed to be a user-space spinlock, but everyone has differing opinions on those.

src/blockchain_db/lmdb/db_lmdb.cpp
hyc: Of course, because doing *anything* else would constitute a race condition.

selsta:
@omartijn CI tests didn’t pass, looks like they got canceled due to them getting stuck during core tests.

omartijn:
Unfortunately, due to a lack of time on my part I am no longer able to continue work on this PR.

I'll close it. If anybody feels up to it, feel free to continue.

Issue #6406, opened 2020-03-27
Epee/ZMQ Serialization Future


vtnerd:
I've analyzed the serialization code in epee (p2p, http), cryptonote (internal db), and zmq. The problems I've identified:

- ZMQ only supports JSON
  - Binary data has to be converted to/from hex which is slower and increases the payload size 
  - The hex conversion is a definite hot-spot when profilng (Monero uses _lots_ of binary data in its RPC for cryptographic data).
- ZMQ serialization requires specifying the field names twice - for both reading and writing (unlike epee stuff).
- Epee uses a DOM internally when reading/writing.
  - _At least_ as slow (not actually clocked) as rapidjson DOM code due to its more generic `std::map<std::string, variant<...>>` design.
  - Removing the rapidjson DOM from zmq-json output had a ~33% performance gain.
- Epee currently has everything in header (via templates), which increases compile times.

After (too?) many iterations - the solution below should support JSON, epee binary, or msgpack binary, and can be used by the ZMQ, P2P, and HTTP components. The ZMQ will be the easiest to retrofit, and the other components (P2P, HTTP) can be done optionally.  The code CAN support reading without a DOM, but the simpler (intermediate DOM) implementation was used here for reading (writing uses no DOM).

This uses `BOOST_PP` AND C++11 variadics. So its a double-shot of fun, but actually less total code than epee or cryptonote schemes (but far more dense).

I'm primarily sharing this to start a discussion on whether its worth doing the remaining integration (I had versions of this working with ZMQ actually, but needs some cleanup). Also, it might be worth preserving this effort for ideas.

The usage looks like:
```c++
// multiple types could be supported with different macro names.
// header
WIRE_JSON_DECLARE(cryptonote::transaction);
WIRE_JSON_DECLARE(cryptonote::tx_out);
// cpp
WIRE_JSON_DEFINE(
  cryptonote::transaction, (
    version,
    unlock_time,
    WIRE_KV_N(vin, "inputs"),
    WIRE_KV_N(vout, "outputs"),
    extra,
    signatures,
    WIRE_KV_N(rct_signatures, "ringct")
  )
);

namespace
{
  template<typename F, typename T>
  void map_txout(F& format, T& self)
  {
    wire::object(format,
      MONERO_FIELD(amount),
      wire::variant_field(self.target,
        wire::option<txout_to_key>{"to_key"},
        wire::option<txout_to_script>{"to_script"},
        wire::option<txout_to_scripthash>{"to_scripthash"}
[...]

PR #6409, opened 2020-03-31
Improve/zmq json zero copy

vtnerd:
Sorry for the churn. I intended to hoard this until the [code in my recent issue](https://github.com/monero-project/monero/issues/6406) was PR'ed, but its probably a better idea to push this concept out now inside of letting it rot. The first commit is from PR 6350, with the second commit being the new code.

This creates a class that rapidjson can use to write directly to a buffer that can be passed into ZMQ without any copy. This same class can be used by the epee binary output code (with nearly zero changes), but code expecting `std::string`s need to change. Epee JSON output can also be replaced, but that requires a few `operator<<` functions. See comments in `byte_stream.h` for why this is an improvement over `rapidjson::StringBuffer` and `std::ostringstream`.

### DIFF
--- a/contrib/epee/src/byte_slice.cpp
+++ b/contrib/epee/src/byte_slice.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2019, The Monero Project
+// Copyright (c) 2019-2020, The Monero Project
 //
 // All rights reserved.
 //
@@ -34,6 +34,7 @@
 #include <utility>
 
 #include "byte_slice.h"
+#include "byte_stream.h"
 
 namespace epee
 {
@@ -117,6 +118,12 @@ namespace epee
     }
   } // anonymous
 
+  void release_byte_buffer::operator()(std::uint8_t* buf) const noexcept
+  {
+    if (buf)
+      std::free(buf - sizeof(raw_byte_slice));
+  }
+
   byte_slice::byte_slice(byte_slice_data* storage, span<const std::uint8_t> portion) noexcept
     : storage_(storage), portion_(portion)
   {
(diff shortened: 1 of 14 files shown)


vtnerd:
I forgot to mention why epee might want this. The epee tcp server stores `std::deque<byte_slice>` internally. So the epee binary and json output can be passed without a copy if switched from `std::ostringstream`. Since the levin (p2p) header has a fixed size, its possible to reserve space for the header, serialize the entire binary epee payload, then write the levin header last. This is useful because relaying blocks would then require 1-2 allocations (typically) for the buffer and zero copies irrespective of the connection count.

contrib/epee/src/byte_slice.cpp
```diff
@@ -217,4 +232,29 @@ namespace epee
...
+  {
+    if (std::numeric_limits<std::size_t>::max() - sizeof(raw_byte_slice) < length)
+      return nullptr;
+
+    std::uint8_t* data = buf.get();
+    if (data != nullptr)
+      data -= sizeof(raw_byte_slice);
+
+    buf.release();
+    data = static_cast<std::uint8_t*>(std::realloc(data, sizeof(raw_byte_slice) + length));
```
xiphon, line 246: Memory leak.

vtnerd: yeah crap, the release must be moved after the `nullptr` check.

vtnerd:
Rebased and fixed bug reported by @xiphon .

luigi1111:
@vtnerd rebase one more time.

vtnerd:
Rebased.

PR #6411, opened 2020-03-31
workflows: retry on apt failure / increase timeout

selsta:


.github/workflows/build.yml
```diff
@@ -51,6 +56,11 @@ jobs:
         submodules: recursive
     - name: remove bundled boost
       run: sudo rm -rf /usr/local/share/boost
+    - name: set apt conf
+      run: |
+        echo "APT::Acquire::Retries \"3\";" | sudo tee -a /etc/apt/apt.conf.d/80-custom
```
xiphon, line 61:
```suggestion
        echo "Acquire::Retries \"3\";" | sudo tee -a /etc/apt/apt.conf.d/80-custom
```

PR #6414, opened 2020-04-01
depends: update openssl package link

sumogr:
Its obsolete since long. My proposal would be to store all depends on getmonero.org server and create a fallback address like https://getmonero.org/depends-sources
Also update the link for getting openssl 1.0.2r, its been moved

### DIFF
--- a/contrib/depends/packages/openssl.mk
+++ b/contrib/depends/packages/openssl.mk
@@ -1,6 +1,6 @@
 package=openssl
 $(package)_version=1.0.2r
-$(package)_download_path=https://www.openssl.org/source
+$(package)_download_path=https://ftp.openssl.org/source/old/1.0.2
 $(package)_file_name=$(package)-$($(package)_version).tar.gz
 $(package)_sha256_hash=ae51d08bba8a83958e894946f15303ff894d75c2b8bbd44a852b64e3fe11d0d6
 $(package)_patches=fix_arflags.patch


selsta:
Why did you remove the fallback? It is not obsolete, see here: https://bitcoincore.org/depends-sources/qtbase-opensource-src-5.7.1.tar.gz

sumogr:
> [https://bitcoincore.org/depends-sources/](https://bitcoincore.org/depends-sources/qtbase-opensource-src-5.7.1.tar.gz)

Hi @selsta 
Yes boost works as well  ( i saw that it was grabbing qt as well from that fallback from travis test)
**but**
https://bitcoincore.org/depends-sources/libusb-1.0.22.tar.bz2 (fails)
https://bitcoincore.org/depends-sources/libsodium-1.0.18.tar.gz (fails)
https://bitcoincore.org/depends-sources/openssl-1.0.2r.tar.gz (fails)
https://bitcoincore.org/depends-sources/protobuf-cpp-3.6.1.tar.gz (fails)
https://bitcoincore.org/depends-sources/zeromq-4.1.7.tar.gz (fails)
https://bitcoincore.org/depends-sources/unbound-1.6.8.tar.gz (fails)
https://bitcoincore.org/depends-sources/hidapi-0.9.0.tar.gz (fails)
https://bitcoincore.org/depends-sources/ncurses-6.1.tar.gz (fails)
and the list goes on.... (its not a fallback address if it only provides a couple of deps)
Your call if you wish to restore it I ll close this

selsta:
Yes, but before removing it we need a new fallback address. Please update this PR to only include the openssl change.

sumogr:
Done

PR #6418, opened 2020-04-02
Adding ZMQ/Pub support for txpool_add and chain_main events

vtnerd:
Supports txpool and mainchain events. Will be updated after PR 6357 is merged. Markdown file has description on how clients should use it, and code has decent comments on implementation quirks.

### DIFF
--- a/ZMQ.md
+++ b/ZMQ.md
@@ -1,0 +1,61 @@
+# The Current/Future Status of ZMQ in Monero
+
+## ZMQ Pub/Sub
+Client `ZMQ_SUB` sockets must "subscribe" to topics before it receives any data.
+This allows filtering on the server side, so network traffic is reduced. Monero
+allows for filtering on: (1) format, (2) context, and (3) event.
+
+ * **format** refers to the _wire_ format (i.e. JSON) used to send event
+   information.
+ * **context** allows for a reduction in fields for the event, so the
+   daemon doesn't waste cycles serializing fields that get ignored.
+ * **event** refers to status changes occurring within the daemon (i.e. new
+   block to main chain).
+
+ * Formats:
+   * `json`
+ * Contexts:
+   * `full` - the entire block or transaction is transmitted (the hash can be
+     computed remotely).
+   * `minimal` - the bare minimum for a remote client to react to an event is
+     sent.
+ * Events:
+   * `chain_main` - changes to the primary/main blockchain.
+   * `txpool_add` - new _publicly visible_ transactions in the mempool.
+     Includes previously unseen transactions in a block but _not_ the
+     `miner_tx`. Does not "re-publish" after a reorg. Includes `do_not_relay`
+     transactions.
+
+The subscription topics are formatted as `format-context-event`, with prefix
+matching supported by both Monero and ZMQ. The `format`, `context` and `event`
+will _never_ have hyphens or colons in their name. For example, subscribing to
+`json-minimal-chain_main` will send minimal information in JSON when changes
+to the main/primary blockchain occur. Whereas, subscribing to `json-minimal`
+will send minimal information in JSON on all available events supported by the
+daemon.
+
+The Monero daemon will ensure that events prefixed by `chain` will be sent in
+"chain-order" - the `prev_id` (hash) field will _always_ refer to a previous
+block. On rollbacks/reorgs, the event will reference an earlier block in the
+chain instead of the last block. The Monero daemon also ensures that
+`txpool_add` events are sent before `chain_*` events - the `chain_*` messages
+will only serialize miner transactions since the other transactions were
+previously published via `txpool_add`. This prevents transactions from being
+serialized twice, even when the transaction was first observed in a block.
+
+ZMQ Pub/Sub will drop messages if the network is congested, so the above rules
+for send order are used for detecting lost messages. A missing gap in `height`
+or `prev_id` for `chain_*` events indicates a lost pub message. Missing
+`txpool_add` messages can only be detected at the next `chain_` message.
+
+Since blockchain events can be dropped, clients will likely want to have a
+timeout against `chain_main` events. The `GetLastBlockHeader` RPC is useful
+for checking the current chain state. Dropped messages should be rare in most
+conditions.
+
+The Monero daemon will send a `txpool_add` pub exactly once for each
+transaction, even after a reorg or restarts. Clients should use the
+`GetTransactionPool` after a reorg to get all transactions that have been put
+back into the tx pool or been invalidated due to a double-spend.
+
+
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -1234,10 +1234,15 @@ bool Blockchain::switch_to_alternative_blockchain(std::list<block_extended_info>
     reorg_notify->notify("%s", std::to_string(split_height).c_str(), "%h", std::to_string(m_db->height()).c_str(),
         "%n", std::to_string(m_db->height() - split_height).c_str(), "%d", std::to_string(discarded_blocks).c_str(), NULL);
 
-  std::shared_ptr<tools::Notify> block_notify = m_block_notify;
-  if (block_notify)
-    for (const auto &bei: alt_chain)
-      block_notify->notify("%s", epee::string_tools::pod_to_hex(get_block_hash(bei.bl)).c_str(), NULL);
+  for (const auto& notifier : m_block_notifiers)
+  {
+    std::size_t notify_height = split_height;
+    for (const auto& bei: alt_chain)
+    {
+      notifier(notify_height, {std::addressof(bei.bl), 1});
+      ++notify_height;
+    }
+  }
 
   MGINFO_GREEN("REORGANIZE SUCCESS! on height: " << split_height << ", new blockchain size: " << m_db->height());
   return true;
@@ -4236,12 +4241,9 @@ leave:
   get_difficulty_for_next_block(); // just to cache it
   invalidate_block_template_cache();
 
-  if (notify)
-  {
-    std::shared_ptr<tools::Notify> block_notify = m_block_notify;
-    if (block_notify)
-      block_notify->notify("%s", epee::string_tools::pod_to_hex(id).c_str(), NULL);
-  }
+
+  for (const auto& notifier: m_block_notifiers)
+    notifier(new_height - 1, {std::addressof(bl), 1});
 
   return true;
 }
@@ -5130,6 +5132,15 @@ void Blockchain::set_user_options(uint64_t maxthreads, bool sync_on_blocks, uint
   m_db_sync_on_blocks = sync_on_blocks;
   m_db_sync_threshold = sync_threshold;
   m_max_prepare_blocks_threads = maxthreads;
+}
+
+void Blockchain::add_block_notify(boost::function<void(std::uint64_t, epee::span<const block>)>&& notify)
+{
+  if (notify)
+  {
+    CRITICAL_REGION_LOCAL(m_blockchain_lock);
+    m_block_notifiers.push_back(std::move(notify));
+  }
 }
 
 void Blockchain::safesyncmode(const bool onoff)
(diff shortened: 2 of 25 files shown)


src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1015,39 +1037,48 @@ namespace cryptonote
...
       const uint64_t weight = results[i].tx.pruned ? get_pruned_transaction_weight(results[i].tx) : get_transaction_weight(results[i].tx, it->blob.size());
-      ok &= add_new_tx(results[i].tx, results[i].hash, tx_blobs[i].blob, weight, tvc[i], tx_relay, relayed);
+      ok &= add_new_tx(results[i].tx, results[i].tx.hash, tx_blobs[i].blob, weight, tvc[i], tx_relay, relayed);
 
       if(tvc[i].m_verifivation_failed)
-      {MERROR_VER("Transaction verification failed: " << results[i].hash);}
+      {MERROR_VER("Transaction verification failed: " << results[i].tx.hash);}
       else if(tvc[i].m_verifivation_impossible)
-      {MERROR_VER("Transaction verification impossible: " << results[i].hash);}
+      {MERROR_VER("Transaction verification impossible: " << results[i].tx.hash);}
```
moneromooo-monero, line 1068: While these changes seem OK in practice, tx.hash is not really meant to be public. It's a cache that's only valid is hash_valid is set. Encouraging this use instead of get_transaction_hash will likely bite us in the rear later. So this adds a couple dozen mostly pointless lines and makes the code more brittle IMHO, even though correct.

ZMQ.md
```diff
@@ -0,0 +1,61 @@
...
+chain instead of the last block. The Monero daemon also ensures that
+`txpool_add` events are sent before `chain_*` events - the `chain_*` messages
+will only serialize miner transactions since the other transactions were
+previously published via `txpool_add`. This prevents transactions from being
+serialized twice, even when the transaction was first observed in a block.
+
+ZMQ Pub/Sub will drop messages if the network is congested, so the above rules
+for send order are used for detecting lost messages. A missing gap in `height`
+or `prev_id` for `chain_*` events indicates a lost pub message. Missing
+`txpool_add` messages can only be detected at the next `chain_` message.
```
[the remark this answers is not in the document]
vtnerd: This is the behavior of ZMQ pub/sub. There isn't a good method to fix or change it. If a downstream clients fails to read any pub messages, the server is then forced to block/wait or queue an infinite amount of messages. The number of messages should be fairly low for our purposes, so its not expected to happen frequently if at all.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -262,6 +264,13 @@ namespace cryptonote
   {
     m_blockchain_storage.set_enforce_dns_checkpoints(enforce_dns);
   }
+  //-----------------------------------------------------------------------------------
+  void core::set_txpool_listener(std::weak_ptr<listener::zmq_pub> zmq_pub)
+  {
+    CRITICAL_REGION_LOCAL(m_incoming_tx_lock);
+    m_zmq_pub = std::move(zmq_pub);
+  }
```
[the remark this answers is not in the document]
vtnerd: I can create a new library for `zmq_pub.cpp` with the core and blockchain dependent on that instead. A callback isn't going to work as desired, because the code still needs to be invoked. Those other libs need to know where to jump either way.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -968,44 +989,45 @@ namespace cryptonote
...
           {
-            results[i].res = handle_incoming_tx_post(*it, tvc[i], results[i].tx, results[i].hash);
+            results[i].valid = handle_incoming_tx_post(*it, tvc[i], results[i].tx, results[i].tx.hash);
           }
           catch (const std::exception &e)
           {
             MERROR_VER("Exception in handle_incoming_tx_post: " << e.what());
             tvc[i].m_verifivation_failed = true;
-            results[i].res = false;
+            results[i].valid = false;
```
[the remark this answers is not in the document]
vtnerd: I don't think its pointless, `res` is not descriptive for what the field represents. The change is pretty simple to revert though.

src/cryptonote_core/cryptonote_core.cpp
vtnerd: Then why is the field public? `get_transaction_hash` should be a public member function that checks the private internal variable and then forwards to the global function. Otherwise, this is clearly indicating accessibility to all. 

This would seem to only "bite us" if one of these functions modifies the hash after its already been calculated once. Seems like a weird update, but yeah sure I'll just duplicate it in memory like before.

vtnerd:
Rebased, and changes from resolved comments.

selsta:
@vtnerd Please rebase again.

vtnerd:
Rebased.

src/rpc/zmq_pub.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+    dest.StartObject();
+    INSERT_INTO_JSON_OBJECT(dest, id, id);
+    INSERT_INTO_JSON_OBJECT(dest, blob_size, blob_size);
+    dest.EndObject();
+  }
+
+  void json_full_chain(rapidjson::StringBuffer& buf, const std::uint64_t height, const cryptonote::block& bl)
+  {
+    json_pub(buf, std::cref(bl));
```
vtnerd, line 161: The JSON schema only allows one block at a time. When the daemon is synchronizing, this sends lots of individual ZMQ messages. Actually changing this will require a bit more work in the "core" code, but the schema doesn't allow any flexibility right now.

I'm thinking about making the schema:
```json
{"first_height": ..., "last_id": ..., "blocks: [ ... ]}
```
I initially decided against this for micro-optimization purposes on the receiver side - they need the equivalent of a `std::vector` which is less efficient for the more common single block case. However, this can be mitigated in several different ways (`boost::container::small_vector`, etc).

The other advantage is that the full block send doesn't contain the hash directly right now (it must be computed by duplicating what the daemon does), and getting the block height for full block requires poking into the `miner_tx`. So these fields are redundant for the full block, but still kind of helpful due to the complexity of the schema.

Any NACKs?

src/rpc/zmq_pub.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+  // boost::adaptors are in place "views" - no copy/move takes place
+  // moving transactions (via sort, etc.), is expensive!
+
+  void json_full_txpool(rapidjson::StringBuffer& buf, epee::span<const cryptonote::txpool_event> txes)
+  {
+    namespace adapt = boost::adaptors;
+    const auto to_full_tx = [](const cryptonote::txpool_event& event)
+    {
+      return event.tx;
```
vtnerd, line 177: It just occurred to me that the full txes don't have the hash included either. They must be computed remotely by duplicating the daemon code. Worth sending the hash here too?

vtnerd:
This has been updated to use the new `byte_slice` send feature with ZMQ that was recently merged. Should've done this during the rebase, but the existing code worked without it.

src/rpc/CMakeLists.txt
```diff
@@ -83,6 +87,8 @@ set(daemon_rpc_server_private_headers
 monero_private_headers(rpc
   ${rpc_private_headers})
 
+set(rpc_pub_private_headers)
```
xiphon, line 90: If it is empty, why we need it here?

src/cryptonote_core/blockchain.cpp
```diff
@@ -1044,10 +1044,13 @@ bool Blockchain::switch_to_alternative_blockchain(std::list<block_extended_info>
...
         "%n", std::to_string(m_db->height() - split_height).c_str(), "%d", std::to_string(discarded_blocks).c_str(), NULL);
 
-  std::shared_ptr<tools::Notify> block_notify = m_block_notify;
-  if (block_notify)
-    for (const auto &bei: alt_chain)
-      block_notify->notify("%s", epee::string_tools::pod_to_hex(get_block_hash(bei.bl)).c_str(), NULL);
+  std::size_t notify_height = split_height;
+  for (const auto& bei: alt_chain)
+  {
+    for (const auto& notifier: m_block_notifiers)
```
xiphon, line 1050: Consider to swap the loops

src/cryptonote_core/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ target_link_libraries(cryptonote_core
     ringct
     device
     hardforks
+    rpc_pub
```
xiphon, line 62: Don't like this. IMO we should avoid introducing this dependency.

src/cryptonote_core/blockchain.cpp
vtnerd: Any more information? I wrote it this way because the same block should remain in the CPU data cache for each call. However, this should thrash the CPU instruction cache. Are you saying you think the opposite is preferred? Some other reason?

xiphon: In case of empty `m_block_notifiers`, we could skip iteration over `alt_chain`.

src/cryptonote_core/CMakeLists.txt
moneromooo-monero: Agreed. If it's a pain to avoid I could live with it though.

src/rpc/CMakeLists.txt
vtnerd: This is the style in this file. The empty set is declared like this and passed to the `monero_add_library` function. I can revert to `""` is desired, I only did it this way to follow the existing style.

src/cryptonote_core/blockchain.cpp
vtnerd: Done.

src/cryptonote_core/CMakeLists.txt
vtnerd: Using `boost::function` (not `std::function`, see comments). The penalty is an indirect function call instead of a direct function call when `zmq_pub` is enabled. Its also less obvious now  that the call is doing a `shared_ptr` lock, but the code for blockchain notification has the same issue.

src/rpc/zmq_pub.cpp
vtnerd: I switched to arrays for `chain_main` events (future extensibility), but did not change the full block output. Minimal block events have differed slightly with array change.

vtnerd:
* Rebased
* `chain_main` now sends block information in arrays
* Changed the serialization routine (necessary for rebase) to allocate memory more efficiently
* Cryptonote core does not depend on ZMQ-PUB
* Fixed bug in `byte_slice` (another PR coming shortly).

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1016,21 +1038,23 @@ namespace cryptonote
...
     bool ok = true;
     it = tx_blobs.begin();
     for (size_t i = 0; i < tx_blobs.size(); i++, ++it) {
-      if (!results[i].res)
+      if (!results[i].valid)
       {
         ok = false;
         continue;
       }
+      results[i].valid = false;
```
moneromooo-monero, line 1057: Why ? Looks like you set valid to false for txes we alreay have in the txpool or the chain, and therefore are valid.

vtnerd: This is probably the result of several edits, and is now confusing. I will remove this line and 1074, then add an `else { results[i].valid = false }` section below the `if` block that starts on line 1071. The intent was to make sure that txes _not_ added to the pool are marked as `!invalid` so that they are not reported over ZMQ (the ZMQ notify code skips those txes).

vtnerd:
* Rebased
* `valid` -> `res` (revert name change)
* Updating `if`/`else` block to be more clear.

src/cryptonote_core/cryptonote_core.cpp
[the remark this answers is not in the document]
vtnerd: Reverted.

vtnerd:
Ok, pushed the changes forreal this time.

moneromooo-monero:
I'm going to assume these are the only changes, and not actually re-read due to the huge rebase diffs.

PR #6424, opened 2020-04-03
Add timelock verification on device

cslashm:
On device time lock validation is requested when using `locked_transfer` command

### DIFF
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -128,6 +128,20 @@ namespace cryptonote
 namespace cryptonote
 {
   //---------------------------------------------------------------
+  void get_transaction_prefix_hash(const transaction_prefix& tx, crypto::hash& h, hw::device &hwdev)
+  {
+    hwdev.get_transaction_prefix_hash(tx,h);    
+  }
+
+  //---------------------------------------------------------------  
+  crypto::hash get_transaction_prefix_hash(const transaction_prefix& tx, hw::device &hwdev)
+  {
+    crypto::hash h = null_hash;
+    get_transaction_prefix_hash(tx, h, hwdev);
+    return h;
+  }
+  
+  //---------------------------------------------------------------  
   void get_transaction_prefix_hash(const transaction_prefix& tx, crypto::hash& h)
   {
     std::ostringstream s;
--- a/src/device/device_ledger.cpp
+++ b/src/device/device_ledger.cpp
@@ -191,7 +191,7 @@ namespace hw {
 
     static int device_id = 0;
 
-    #define PROTOCOL_VERSION                    2
+    #define PROTOCOL_VERSION                    3
 
     #define INS_NONE                            0x00
     #define INS_RESET                           0x02
@@ -228,6 +228,7 @@ namespace hw {
     #define INS_BLIND                           0x78
     #define INS_UNBLIND                         0x7A
     #define INS_GEN_TXOUT_KEYS                  0x7B
+    #define INS_PREFIX_HASH                     0x7D
     #define INS_VALIDATE                        0x7C
     #define INS_MLSAG                           0x7E
     #define INS_CLOSE_TX                        0x80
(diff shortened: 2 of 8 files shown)


selsta:
Can PR this against master branch instead? I don’t think we will do another v0.15.0.X release anymore.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -128,6 +128,18 @@ namespace cryptonote
 namespace cryptonote
 {
   //---------------------------------------------------------------
+  void get_transaction_prefix_hash(const transaction_prefix& tx, crypto::hash& h, hw::device &hwdev) {
```
sumogr, line 131: mind the braces here and on L136 for file's consistency reasons.

TheCharlatan:
Do you have the device-side code already?

src/device/device_ledger.cpp
```diff
@@ -1346,6 +1347,75 @@ namespace hw {
...
+      #ifdef DEBUG_HWDEVICE
+      crypto::hash h_x;
+      crypto::hash h_xx;
+      this->controle_device->get_transaction_prefix_hash(tx,h_x);
+      MDEBUG("get_transaction_prefix_hash [[IN]] h_x/1 "<<h_x);
+      #endif
+    
+      std::ostringstream s_x;
+      binary_archive<true> a_x(s_x);
+      ::serialization::serialize(a_x, const_cast<cryptonote::transaction_prefix&>(tx));
```
moneromooo-monero, line 1364: Return code not checked

src/device/device_ledger.cpp
```diff
@@ -1346,6 +1347,75 @@ namespace hw {
...
+      pref_length = s_x.str().size();
+      char pref[pref_length];
```
moneromooo-monero, line 1366: Potential way for someone to pass a huge tx and smash the stack ?

src/device/device_ledger.cpp
```diff
@@ -1346,6 +1347,75 @@ namespace hw {
...
+
+      offset = set_command_header_noopt(INS_PREFIX_HASH,1);
+
+      //version as varint
+      pref_offset = 0;
+      do {
+        this->buffer_send[offset] = pref[pref_offset];
+        offset += 1;
+        pref_offset += 1;
+      } while (pref[pref_offset]&0x80);
```
moneromooo-monero, line 1377: That looks (ie, I did not check carefully) wrong, if it's monero varints. The high bit tells you whether there is another byte to follow. Here, you check the high bit of a byte you won't use. This can also buffer overrun with crafted data (both read and write).

src/device/device_ledger.cpp
cslashm: I juste copy/paste the orignal code. I will add check

src/device/device_ledger.cpp
cslashm: I dont know. Do you prefere a malloc/free based code?

src/device/device_ledger.cpp
cslashm: I take the whole varint and pass it to device, which will decode it. This code has been tested with device (app not yep pushed) and it works.

In fact I recompute the prefix hash on device and will this once during the signature. So I send the raw serialization prefix hash to the app.  

OK I will check buffer overrun, in a general way, I assert the the passed argument are well-formed.

cslashm:
@TheCharlatan , the app code will be pushed only in few day for internal reasons. Sorry for the delay.

cslashm:
@selsta @moneromooo-monero I will made a second PR on master. rebase is a pain....And jusr in case I prefer to keep this one.

cslashm:
Note: This PR has been sucessfully tested on stagenet

src/device/device_ledger.cpp
moneromooo-monero: If the range s_x.str() is not bounded to a known size that's suitable for a stack, yes.

src/device/device_ledger.cpp
moneromooo-monero: Until proven otherwise, I think it's working by chance. Imagine you've got a buffer with a varint with value 1, which encodes as just "1". Imagine the next data is a set of 0x80. The code will copy the first byte (1), then increase pref_offset to 1, then check that for 0x80. It matches, so the loop will continue. That is wrong. Is the above wrong ?

cslashm: Seems you are right yes! I rewrite my initial `while` loop to a `do while` one, a bit too quickly!

src/device/device_ledger.cpp
cslashm: ok, I will do

src/device/device_ledger.cpp
cslashm: the function return void

moneromooo-monero: That's odd. Can you point to it ? The ones I see return bool but I could be missing some overrides.

cslashm: I found this one https://github.com/monero-project/monero/blob/release-v0.15/src/cryptonote_basic/cryptonote_boost_serialization.h#L153

But I'm sure at all about if it is the right one or not

moneromooo-monero: These are for the boost serialization, used for files (ie, saving the wallet cache). Wire serialization is not that, since boost serialization is dependent on boost version and endianness. You want src/serialization/serialization.h

cslashm: thanks for the pointer. I will check

cslashm: I added a CHECK_AND_ASSERT_THROW_MES

moneromooo-monero:
+duplicate

Closing in favour of #6436

PR #6430, opened 2020-04-05
daemon: add a "disk" command to check current available disk space

sumogr:
Useful both while syncing and while being synced

### DIFF
--- a/src/daemon/command_server.cpp
+++ b/src/daemon/command_server.cpp
@@ -165,6 +165,11 @@ t_command_server::t_command_server(
     , "Show the current difficulty."
     );
   m_command_lookup.set_handler(
+      "disk"
+    , std::bind(&t_command_parser_executor::show_disk, &m_parser, p::_1)
+    , "Show remaining disk space."
+    );
+  m_command_lookup.set_handler(
       "status"
     , std::bind(&t_command_parser_executor::show_status, &m_parser, p::_1)
     , "Show the current status."
--- a/src/daemon/rpc_command_executor.cpp
+++ b/src/daemon/rpc_command_executor.cpp
@@ -438,6 +438,36 @@ static float get_sync_percentage(const cryptonote::COMMAND_RPC_GET_INFO::respons
   return get_sync_percentage(ires.height, ires.target_height);
 }
 
+bool t_rpc_command_executor::show_disk() {
+  cryptonote::COMMAND_RPC_GET_INFO::request ireq;
+  cryptonote::COMMAND_RPC_GET_INFO::response ires;
+  epee::json_rpc::error error_resp;
+
+  std::string fail_message = "Problem fetching info";
+
+  if (m_is_rpc)
+  {
+    if (!m_rpc_client->rpc_request(ireq, ires, "/getinfo", fail_message.c_str()))
+    {
+      return true;
+    }
+  }
+  else
+  {
+    if (!m_rpc_server->on_get_info(ireq, ires) || ires.status != CORE_RPC_STATUS_OK)
+    {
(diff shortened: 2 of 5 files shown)


src/daemon/command_server.cpp
```diff
@@ -164,6 +164,11 @@ t_command_server::t_command_server(
     , std::bind(&t_command_parser_executor::show_difficulty, &m_parser, p::_1)
     , "Show the current difficulty."
     );
+  m_command_lookup.set_handler(
+      "disk"
+    , std::bind(&t_command_parser_executor::show_disk, &m_parser, p::_1)
+    , "Show db size and remaining disk space."
```
moneromooo-monero, line 170: The code below just shows free space, not db size.

src/daemon/rpc_command_executor.cpp
```diff
@@ -438,6 +438,36 @@ static float get_sync_percentage(const cryptonote::COMMAND_RPC_GET_INFO::respons
...
+  else
+  {
+    if (!m_rpc_server->on_get_info(ireq, ires) || ires.status != CORE_RPC_STATUS_OK)
+    {
+      tools::fail_msg_writer() << make_error(fail_message, ires.status);
+      return true;
+    }
+  }
+
+  uint64_t freespace = (ires.free_space)/1000000;
```
moneromooo-monero, line 464: 1024*1024 ?

sumogr: to turn into MB

sumogr:
> I'm a bit doubtful we should bring OS commands in, but if others want it, it goes in.

I just saw that many people are struggling with insufficient disk space and wonder why they fail to sync and thought that it might be helpful. I agree and its not something too important anyway. I ll close if there is concern

sumogr:
Maybe just a global warning if availability drops below 2 Gb might do the job. Up to you

xiphon:
> I just saw that many people are struggling with insufficient disk space and wonder why they fail to sync and thought that it might be helpful

Monero already have a code to print `!! WARNING: Insufficient free space to extend database !!: X MB available, XX MB needed` error in this case.

sumogr:
> > I just saw that many people are struggling with insufficient disk space and wonder why they fail to sync and thought that it might be helpful
> 
> Monero already has a code to print `!! WARNING: Insufficient free space to extend database !!: X MB available, XX MB needed` error in this case.

perfect. closing

sumogr:
Actually @xiphon, i searched for it, it is an error phrased as a warning    
` if(si.available < add_size)
    {
      MERROR("!! WARNING: Insufficient free space to extend database !!: " <<
          (si.available >> 20L) << " MB available, " << (add_size >> 20L) << " MB needed");
      return;
    }`
that is catched when it is already too late.
I ll try and think a "smart" way of getting a warning before its too late

trasherdk:
Maybe start getting noisy when down to a few GB of free space.

xiphon:
> I just saw that many people are struggling with insufficient disk space and **wonder why they fail to sync**

They shouldn't wonder, the error message is pretty self-explanatory.

> I ll try and think a "smart" way of getting a warning before its too late

I'm sure it is primarily an OS responsibility.
Windows, MacOS, Ubuntu will notify a user of a lack of free disk space.

sumogr:
> Maybe just a global warning if availability drops below 2 Gb might do the job. Up to you

sorry i thought you were replying to this by posting that error message cause it came right after.

>I'm sure it is primarily an OS responsibility.
Windows, MacOS, Ubuntu will notify a user of a lack of free disk space.

True. Np

PR #6431, opened 2020-04-05
lmdb: print available disk space after each db resize

sumogr:
and warn us hard if it is below 4 Gb

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -30,6 +30,7 @@
 #include <boost/filesystem.hpp>
 #include <boost/format.hpp>
 #include <boost/circular_buffer.hpp>
+#include <boost/optional.hpp>
 #include <memory>  // std::unique_ptr
 #include <cstring>  // memcpy
 
@@ -573,8 +574,22 @@ void BlockchainLMDB::do_resize(uint64_t increase_size)
   int result = mdb_env_set_mapsize(m_env, new_mapsize);
   if (result)
     throw0(DB_ERROR(lmdb_error("Failed to set new mapsize: ", result).c_str()));
+  
+  boost::filesystem::path path(m_folder);
+  boost::filesystem::space_info si = boost::filesystem::space(path);
+  boost::optional<uint64_t> space_available = si.available;
 
-  MGINFO("LMDB Mapsize increased." << "  Old: " << mei.me_mapsize / (1024 * 1024) << "MiB" << ", New: " << new_mapsize / (1024 * 1024) << "MiB");
+  if (space_available)
+  {
+   if ((*space_available / (1024 * 1024)) < 4294)
+     MGINFO_RED("Warning you are running low on disk space!\n" << "LMDB Mapsize increased." << "  Old: " << mei.me_mapsize / (1024 * 1024) << "MiB" << ", New: " << new_mapsize / (1024 * 1024) << "MiB" << ", Available disk remaining: " << *space_available / (1024 * 1024) << "MiB" );
+   else if ((*space_available / (1024 * 1024)) >= 4294)
+     MGINFO("LMDB Mapsize increased." << "  Old: " << mei.me_mapsize / (1024 * 1024) << "MiB" << ", New: " << new_mapsize / (1024 * 1024) << "MiB" << ", Available disk remaining: " << *space_available / (1024 * 1024) << "MiB" );
+  }
+  else 
+  {
+    MGINFO("LMDB Mapsize increased." << "  Old: " << mei.me_mapsize / (1024 * 1024) << "MiB" << ", New: " << new_mapsize / (1024 * 1024) << "MiB");
+  }
 
   mdb_txn_safe::allow_new_txns();
 }


src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -518,10 +518,10 @@ void BlockchainLMDB::do_resize(uint64_t increase_size)
   const uint64_t add_size = 1LL << 30;
 
   // check disk capacity
+  boost::filesystem::path path(m_folder);
+  boost::filesystem::space_info si = boost::filesystem::space(path);
```
selsta, line 522: What about the try/catch block?

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -574,7 +574,10 @@ void BlockchainLMDB::do_resize(uint64_t increase_size)
   if (result)
     throw0(DB_ERROR(lmdb_error("Failed to set new mapsize: ", result).c_str()));
 
-  MGINFO("LMDB Mapsize increased." << "  Old: " << mei.me_mapsize / (1024 * 1024) << "MiB" << ", New: " << new_mapsize / (1024 * 1024) << "MiB");
+  if ((si.available / (1024 * 1024)) > 2000)
+  MGINFO("LMDB Mapsize increased." << "  Old: " << mei.me_mapsize / (1024 * 1024) << "MiB" << ", New: " << new_mapsize / (1024 * 1024) << "MiB" << ", Available disk remaining: " << si.available / (1024 * 1024) << "MiB" );
```
selsta, line 578: Indentation seems to be missing

sumogr: done

src/blockchain_db/lmdb/db_lmdb.cpp
sumogr: there `si` redeclared

selsta: boost::filesystem::space can still fail

sumogr: oh got it now

sumogr: How about this boost's magic potion @selsta ? (i think it can be used instead of that try/catch block as well)

selsta: This does not look correct to me. si.available is uintmax_t and not an optional type. This does not catch filesystem_error exceptions.

sumogr: ok you most likely misunderstand the use of boost::optional 's class template.
_DESCRIPTION:
The class template boost::optional manages an optional contained value (a value that or may not be present).A common use of boost::optional is in cases when the return value of a function may fail in some cases._

selsta: I know the use case of boost optional but this requires a boost::optional return value which is not the case here. See here as an example:

https://www.boost.org/doc/libs/1_66_0/libs/optional/doc/html/boost_optional/quick_start.html

But I will leave this to the more experienced devs :P

sumogr: Actually boost::option maybe used for any nullable object see this example https://www.boost.org/doc/libs/1_65_1/libs/optional/doc/html/index.html
I am closing this and leave more experienced devs :P to keep on suggesting code to your project

selsta: Your example also has an boost::optional return value: `boost::optional<int> getConfigParam`

> I am closing this and leave more experienced devs :P to keep on suggesting code to your project

Not sure why you closed this now. I only said that I will leave this review to devs who are more experienced than me.

sumogr: Pardon me for being jumpy.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -573,8 +574,22 @@ void BlockchainLMDB::do_resize(uint64_t increase_size)
...
     throw0(DB_ERROR(lmdb_error("Failed to set new mapsize: ", result).c_str()));
+  
+  boost::filesystem::path path(m_folder);
+  boost::filesystem::space_info si = boost::filesystem::space(path);
+  boost::optional<uint64_t> space_available = si.available;
 
-  MGINFO("LMDB Mapsize increased." << "  Old: " << mei.me_mapsize / (1024 * 1024) << "MiB" << ", New: " << new_mapsize / (1024 * 1024) << "MiB");
+  if (space_available)
+  {
+   if ((*space_available / (1024 * 1024)) < 4294)
```
moneromooo-monero, line 584: Why this particular odd threshold, out of curiosity ?

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -573,8 +574,22 @@ void BlockchainLMDB::do_resize(uint64_t increase_size)
...
+     MGINFO_RED("Warning you are running low on disk space!\n" << "LMDB Mapsize increased." << "  Old: " << mei.me_mapsize / (1024 * 1024) << "MiB" << ", New: " << new_mapsize / (1024 * 1024) << "MiB" << ", Available disk remaining: " << *space_available / (1024 * 1024) << "MiB" );
+   else if ((*space_available / (1024 * 1024)) >= 4294)
```
moneromooo-monero, line 586: FWIW, that test will always trigger, no ?

src/blockchain_db/lmdb/db_lmdb.cpp
moneromooo-monero: selsta is right. Assigning a return value will happen if no exception is thrown by the call. If an exception is thrown, that assignment will be skipped anyway. The optional is pointless in this case. Use try/catch and get rid of the optional. Using optional would only make sense for an API which was designed to use it, which this seems not to.

dEBRUYNE-1:
@sumogr - ping. There are a few open questions.

src/blockchain_db/lmdb/db_lmdb.cpp
sumogr: well i empirically presume that below 4 Gb there will most likely be no more room for the next resize, especially when synced or approaching the end of the chain while syncing. If a more reasonable threshold is proposed i ll change it

src/blockchain_db/lmdb/db_lmdb.cpp
sumogr: yes

PR #6441, opened 2020-04-08
use memwipe on secret k/alpha values

moneromooo-monero:
Reported by UkoeHB_

### DIFF
--- a/src/multisig/multisig.cpp
+++ b/src/multisig/multisig.cpp
@@ -82,6 +82,7 @@ namespace cryptonote
     {
       rct::key sk = rct::scalarmultKey(rct::pk2rct(k), rct::sk2rct(blinded_skey));
       crypto::secret_key msk = get_multisig_blinded_secret_key(rct::rct2sk(sk));
+      memwipe(&sk, sizeof(sk));
       multisig_keys.push_back(msk);
       sc_add(spend_skey.bytes, spend_skey.bytes, (const unsigned char*)msk.data);
     }
@@ -126,10 +127,10 @@ namespace cryptonote
   //-----------------------------------------------------------------
   crypto::secret_key generate_multisig_view_secret_key(const crypto::secret_key &skey, const std::vector<crypto::secret_key> &skeys)
   {
-    rct::key view_skey = rct::sk2rct(get_multisig_blinded_secret_key(skey));
+    crypto::secret_key view_skey = get_multisig_blinded_secret_key(skey);
     for (const auto &k: skeys)
-      sc_add(view_skey.bytes, view_skey.bytes, rct::sk2rct(k).bytes);
-    return rct::rct2sk(view_skey);
+      sc_add((unsigned char*)&view_skey, rct::sk2rct(view_skey).bytes, rct::sk2rct(k).bytes);
+    return view_skey;
   }
   //-----------------------------------------------------------------
   crypto::public_key generate_multisig_M_N_spend_public_key(const std::vector<crypto::public_key> &pkeys)
--- a/src/ringct/rctTypes.h
+++ b/src/ringct/rctTypes.h
@@ -48,6 +48,7 @@ extern "C" {
 
 #include "hex.h"
 #include "span.h"
+#include "memwipe.h"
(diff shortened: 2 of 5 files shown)


src/ringct/rctTypes.h
```diff
@@ -106,6 +107,8 @@ namespace rct {
         key L;
         key R;
         key ki;
+
+        ~multisig_kLRki() { memwipe(&k, sizeof(k)); }
```
UkoeHB, line 111: Does this mean when a kLRki struct is destroyed, it also memwipes the addresses corresponding to k? Seems quite useful

UkoeHB:
Would you mind going through src/multisig/multisig.cpp as well? There appear to be a number of functions with unwiped multisig key variables.

src/ringct/rctTypes.h
moneromooo-monero: Yes.

src/multisig/multisig.cpp
```diff
@@ -126,10 +125,10 @@ namespace cryptonote
   //-----------------------------------------------------------------
   crypto::secret_key generate_multisig_view_secret_key(const crypto::secret_key &skey, const std::vector<crypto::secret_key> &skeys)
   {
-    rct::key view_skey = rct::sk2rct(get_multisig_blinded_secret_key(skey));
+    crypto::secret_key view_skey = get_multisig_blinded_secret_key(skey);
     for (const auto &k: skeys)
-      sc_add(view_skey.bytes, view_skey.bytes, rct::sk2rct(k).bytes);
-    return rct::rct2sk(view_skey);
+      sc_add((unsigned char*)&view_skey, rct::sk2rct(view_skey).bytes, rct::sk2rct(k).bytes);
+    return view_skey;
```
UkoeHB, line 131: Are instances of crypto::secret_key wiped on destruction? I see "using secret_key = epee::mlocked<tools::scrubbed<ec_scalar>>;" in src/crypto/crypto.h (not that I understand what it does).

moneromooo-monero: Yes.

UkoeHB:
Today remembered all of the transaction proofs. Some more secret variables:
src/crypto/crypto.cpp generate_ring_signature() 'k'.
src/crypto/crypto.cpp crypto_ops::generate_tx_proof() 'k'.

Maybe getting out of scope of this pull request, but the 'secret index' which says which member of a ring signature is the true signer, should also be wiped. In generate_ring_signature() it is 'sec_index' and in MLSAG_gen() it is 'index'. Those are passed by value and not by reference, so I imagine they'd need to be tracked down and wiped in all cases.

src/multisig/multisig.cpp
```diff
@@ -80,8 +80,7 @@ namespace cryptonote
     crypto::secret_key blinded_skey = get_multisig_blinded_secret_key(keys.m_spend_secret_key);
     for (const auto &k: spend_keys)
     {
-      rct::key sk = rct::scalarmultKey(rct::pk2rct(k), rct::sk2rct(blinded_skey));
-      crypto::secret_key msk = get_multisig_blinded_secret_key(rct::rct2sk(sk));
+      crypto::secret_key msk = get_multisig_blinded_secret_key(rct::rct2sk(rct::scalarmultKey(rct::pk2rct(k), rct::sk2rct(blinded_skey))));
```
vtnerd, line 83: The compiler is still freely allowed to use temporaries ... ?

moneromooo-monero: I'm not writing a compiler just for this though.

vtnerd: My comment was about the change removing the temporary being just for show. The compiler isn't required to create a temporary address with the existing code, yet can still create one after this change. This is generally why these techniques are _nearly_ pointless, its hard to target every memory address in practice.

moneromooo-monero: OK. Then I guess it's best to make very simple statements, with explicit wipes. Agreed ?

moneromooo-monero: I've done that then.

PR #6442, opened 2020-04-10
Fix possible realloc memory leak in external/db_drivers/liblmdb

draane:
Fix possible realloc memory leak for variable `ids` and `idn` in `mdb_midl_shrink` and `mdb_midl_grow`.
If `realloc` fails the variables are nulled but not freed, leading to a possible memory leak.

### DIFF
--- a/external/db_drivers/liblmdb/midl.c
+++ b/external/db_drivers/liblmdb/midl.c
@@ -120,9 +120,11 @@ void mdb_midl_free(MDB_IDL ids)
 void mdb_midl_shrink( MDB_IDL *idp )
 {
 	MDB_IDL ids = *idp;
+	MDB_IDL new_ids;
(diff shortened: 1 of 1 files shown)


moneromooo-monero:
I don't see the leaks here. Can you be more specific ?
The one thing I see is a possible off by 1 in _shrink, regardless of the patch.

hyc:
There are no bugs in midl.c

moneromooo-monero:
If you can't point to an actual leak (ie, sequence of events leading to a leak), I'll close soon.

hyc:
There is no leak. This PR is invalid.

Come on, he's patched a void function - it doesn't return anything. It doesn't matter whether it NULLs out the variable or not, it all gets discarded.

PR #6443, opened 2020-04-10
p2p: startup speedup, init seed nodes on first 'connect_to_seed()'

xiphon:
Speed up the daemon startup, ~15-20sec gain.

1. On-demand seed nodes initialization (DNS seed nodes lookup with fallback to hardcoded list). Shouldn't do this on each daemon startup, but only when we have no white peers and want to connect to a seed node.
Run the task in background, i.e. in p2p connection thread on first `connect_to_seed()` call.
2. `m_seed_nodes` access should be thread-safe
3. `m_fallback_seed_nodes_added` flag should be atomic

### DIFF
--- a/src/p2p/net_node.h
+++ b/src/p2p/net_node.h
@@ -384,6 +384,7 @@ namespace nodetool
     bool is_addr_recently_failed(const epee::net_utils::network_address& addr);
     bool is_priority_node(const epee::net_utils::network_address& na);
     std::set<std::string> get_seed_nodes(cryptonote::network_type nettype) const;
+    std::set<std::string> get_seed_nodes();
     bool connect_to_seed();
 
     template <class Container>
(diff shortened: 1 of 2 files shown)


src/p2p/net_node.inl
```diff
@@ -633,6 +635,115 @@ namespace nodetool
...
+  {
+    if (m_nettype == cryptonote::TESTNET)
+    {
+      return get_seed_nodes(cryptonote::TESTNET);
+    }
+    if (m_nettype == cryptonote::STAGENET)
+    {
+      return get_seed_nodes(cryptonote::STAGENET);
+    }
+    if (!m_exclusive_peers.empty() || m_offline)
```
moneromooo-monero, line 648: Should this not be first ?

xiphon: Is the way it is done in the current code.

Yeah, moved it.

src/p2p/net_node.h
```diff
@@ -467,7 +468,9 @@ namespace nodetool
     std::list<epee::net_utils::network_address>   m_priority_peers;
     std::vector<epee::net_utils::network_address> m_exclusive_peers;
     std::vector<epee::net_utils::network_address> m_seed_nodes;
-    bool m_fallback_seed_nodes_added;
+    std::atomic_flag m_seed_nodes_initialized;
+    boost::shared_mutex m_seed_nodes_lock;
+    std::atomic_flag m_fallback_seed_nodes_added;
```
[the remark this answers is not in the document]
xiphon: Updated

PR #6446, opened 2020-04-12
trezor: adapt to new passphrase mechanism

ph4r05:
- choice where to enter the passphrase is now made on the host (Trezor changed this in new firmwares)
- use wipeable string in the Trezor communication stack
- wipe passphrase memory after use
- protocol optimizations, prepare for new firmware version which scales better with subaddresses
  - subaddresses are not pre-computed but each UTXO carries also subaddr index so all key are computed on the fly, constant memory w.r.t. number of subaddresses
  - same for key image refresh
  - sending only required info to the Trezor for each step, saves Trezor CPU & RAM, faster protocol
  - optimizations detailed in https://github.com/trezor/trezor-firmware/pull/786
- minor fixes and improvements
- tests: fix and update Trezor tests, HF12, disable long payment ID test
- wallet API update for passphrases, GUI related PR: https://github.com/monero-project/monero-gui/pull/2829
- Trezor related PR: https://github.com/trezor/trezor-firmware/pull/942
- Trezor CLSAG PR: https://github.com/trezor/trezor-firmware/pull/943 
- more on passphrase redesign: https://docs.trezor.io/trezor-firmware/common/communication/passphrase-redesign-migration.html

### DIFF
--- a/src/device/device.hpp
+++ b/src/device/device.hpp
@@ -78,7 +78,7 @@ namespace hw {
         virtual void on_button_request(uint64_t code=0) {}
         virtual void on_button_pressed() {}
         virtual boost::optional<epee::wipeable_string> on_pin_request() { return boost::none; }
-        virtual boost::optional<epee::wipeable_string> on_passphrase_request(bool on_device) { return boost::none; }
+        virtual boost::optional<epee::wipeable_string> on_passphrase_request(bool & on_device) { on_device = true; return boost::none; }
         virtual void on_progress(const device_progress& event) {}
         virtual ~i_device_callback() = default;
     };
--- a/src/device_trezor/device_trezor_base.cpp
+++ b/src/device_trezor/device_trezor_base.cpp
@@ -28,6 +28,7 @@
 //
 
 #include "device_trezor_base.hpp"
+#include "memwipe.h"
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/split.hpp>
 #include <boost/regex.hpp>
@@ -151,7 +152,7 @@ namespace trezor {
 
     bool device_trezor_base::disconnect() {
       TREZOR_AUTO_LOCK_DEVICE();
-      m_device_state.clear();
+      m_device_session_id.clear();
       m_features.reset();
 
       if (m_transport){
@@ -292,8 +293,8 @@ namespace trezor {
         case messages::MessageType_PassphraseRequest:
           on_passphrase_request(input, dynamic_cast<const messages::common::PassphraseRequest*>(input.m_msg.get()));
           return true;
-        case messages::MessageType_PassphraseStateRequest:
-          on_passphrase_state_request(input, dynamic_cast<const messages::common::PassphraseStateRequest*>(input.m_msg.get()));
+        case messages::MessageType_Deprecated_PassphraseStateRequest:
+          on_passphrase_state_request(input, dynamic_cast<const messages::common::Deprecated_PassphraseStateRequest*>(input.m_msg.get()));
           return true;
         case messages::MessageType_PinMatrixRequest:
           on_pin_request(input, dynamic_cast<const messages::common::PinMatrixRequest*>(input.m_msg.get()));
@@ -361,23 +362,34 @@ namespace trezor {
       return false;
     }
 
-    void device_trezor_base::device_state_reset_unsafe()
+    void device_trezor_base::device_state_initialize_unsafe()
     {
       require_connected();
+      std::string tmp_session_id;
       auto initMsg = std::make_shared<messages::management::Initialize>();
+      const auto data_cleaner = epee::misc_utils::create_scope_leave_handler([&]() {
+        memwipe(&tmp_session_id[0], tmp_session_id.size());
+      });
 
-      if(!m_device_state.empty()) {
-        initMsg->set_allocated_state(&m_device_state);
+      if(!m_device_session_id.empty()) {
+        tmp_session_id.assign(m_device_session_id.data(), m_device_session_id.size());
+        initMsg->set_allocated_session_id(&tmp_session_id);
       }
 
       m_features = this->client_exchange<messages::management::Features>(initMsg);
-      initMsg->release_state();
+      if (m_features->has_session_id()){
+        m_device_session_id = m_features->session_id();
+      } else {
+        m_device_session_id.clear();
+      }
+
+      initMsg->release_session_id();
     }
 
     void device_trezor_base::device_state_reset()
     {
       TREZOR_AUTO_LOCK_CMD();
-      device_state_reset_unsafe();
+      device_state_initialize_unsafe();
     }
 
 #ifdef WITH_TREZOR_DEBUGGING
@@ -441,48 +453,89 @@ namespace trezor {
         pin = m_pin;
       }
 
-      // TODO: remove PIN from memory
+      std::string pin_field;
       messages::common::PinMatrixAck m;
       if (pin) {
-        m.set_pin(pin.get().data(), pin.get().size());
+        pin_field.assign(pin->data(), pin->size());
+        m.set_allocated_pin(&pin_field);
       }
+
+      const auto data_cleaner = epee::misc_utils::create_scope_leave_handler([&]() {
+        m.release_pin();
+        if (!pin_field.empty()){
+          memwipe(&pin_field[0], pin_field.size());
+        }
+      });
+
       resp = call_raw(&m);
     }
 
     void device_trezor_base::on_passphrase_request(GenericMessage & resp, const messages::common::PassphraseRequest * msg)
     {
       CHECK_AND_ASSERT_THROW_MES(msg, "Empty message");
-      MDEBUG("on_passhprase_request, on device: " << msg->on_device());
+      MDEBUG("on_passhprase_request");
+
+      // Backward compatibility, migration clause.
+      if (msg->has__on_device() && msg->_on_device()){
+        messages::common::PassphraseAck m;
+        resp = call_raw(&m);
+        return;
+      }
+
+      bool on_device = true;
+      if (msg->has__on_device() && !msg->_on_device()){
+        on_device = false;  // do not enter on device, old devices.
+      }
+
+      if (on_device && m_features && m_features->capabilities_size() > 0){
+        on_device = false;
+        for (auto it = m_features->capabilities().begin(); it != m_features->capabilities().end(); it++) {
+          if (*it == messages::management::Features::Capability_PassphraseEntry){
+            on_device = true;
+          }
+        }
+      }
+
       boost::optional<epee::wipeable_string> passphrase;
-      TREZOR_CALLBACK_GET(passphrase, on_passphrase_request, msg->on_device());
+      TREZOR_CALLBACK_GET(passphrase, on_passphrase_request, on_device);
 
-      if (!passphrase && m_passphrase){
-        passphrase = m_passphrase;
-      }
-
-      m_passphrase = boost::none;
-
+      std::string passphrase_field;
       messages::common::PassphraseAck m;
-      if (!msg->on_device() && passphrase){
-        // TODO: remove passphrase from memory
-        m.set_passphrase(passphrase.get().data(), passphrase.get().size());
+      m.set_on_device(on_device);
+      if (!on_device) {
+        if (!passphrase && m_passphrase) {
+          passphrase = m_passphrase;
+        }
+
+        if (m_passphrase) {
+          m_passphrase = boost::none;
+        }
+
+        if (passphrase) {
+          passphrase_field.assign(passphrase->data(), passphrase->size());
+          m.set_allocated_passphrase(&passphrase_field);
+        }
       }
 
-      if (!m_device_state.empty()){
-        m.set_allocated_state(&m_device_state);
-      }
+      const auto data_cleaner = epee::misc_utils::create_scope_leave_handler([&]() {
+        m.release_passphrase();
+        if (!passphrase_field.empty()){
+          memwipe(&passphrase_field[0], passphrase_field.size());
+        }
+      });
 
       resp = call_raw(&m);
-      m.release_state();
     }
 
-    void device_trezor_base::on_passphrase_state_request(GenericMessage & resp, const messages::common::PassphraseStateRequest * msg)
+    void device_trezor_base::on_passphrase_state_request(GenericMessage & resp, const messages::common::Deprecated_PassphraseStateRequest * msg)
     {
       MDEBUG("on_passhprase_state_request");
       CHECK_AND_ASSERT_THROW_MES(msg, "Empty message");
 
-      m_device_state = msg->state();
-      messages::common::PassphraseStateAck m;
+      if (msg->has_state()) {
+        m_device_session_id = msg->state();
+      }
+      messages::common::Deprecated_PassphraseStateAck m;
       resp = call_raw(&m);
     }
 
@@ -510,7 +563,7 @@ namespace trezor {
       }
 
       auto msg = std::make_shared<messages::management::LoadDevice>();
-      msg->set_mnemonic(mnemonic);
+      msg->add_mnemonics(mnemonic);
       msg->set_pin(pin);
       msg->set_passphrase_protection(passphrase_protection);
       msg->set_label(label);
@@ -535,7 +588,8 @@ namespace trezor {
       return boost::none;
     }
 
-    boost::optional<epee::wipeable_string> trezor_debug_callback::on_passphrase_request(bool on_device) {
+    boost::optional<epee::wipeable_string> trezor_debug_callback::on_passphrase_request(bool & on_device) {
+      on_device = true;
       return boost::none;
     }
 
--- a/src/device_trezor/device_trezor_base.hpp
+++ b/src/device_trezor/device_trezor_base.hpp
@@ -70,7 +70,7 @@ namespace trezor {
 
       void on_button_request(uint64_t code=0) override;
       boost::optional<epee::wipeable_string> on_pin_request() override;
-      boost::optional<epee::wipeable_string> on_passphrase_request(bool on_device) override;
+      boost::optional<epee::wipeable_string> on_passphrase_request(bool & on_device) override;
       void on_passphrase_state_request(const std::string &state);
       void on_disconnect();
     protected:
@@ -94,7 +94,7 @@ namespace trezor {
 
       std::string m_full_name;
       std::vector<unsigned int> m_wallet_deriv_path;
-      std::string m_device_state;  // returned after passphrase entry, session
+      epee::wipeable_string m_device_session_id;  // returned after passphrase entry, session
       std::shared_ptr<messages::management::Features> m_features;  // features from the last device reset
       boost::optional<epee::wipeable_string> m_pin;
       boost::optional<epee::wipeable_string> m_passphrase;
@@ -117,7 +117,7 @@ namespace trezor {
       void require_initialized() const;
       void call_ping_unsafe();
       void test_ping();
-      virtual void device_state_reset_unsafe();
+      virtual void device_state_initialize_unsafe();
       void ensure_derivation_path() noexcept;
 
       // Communication methods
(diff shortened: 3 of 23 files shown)


selsta:
Any idea about CI failing with `error: invalid path 'defs/ethereum/aux.png'` on Windows?

ph4r05:
> Any idea about CI failing with `error: invalid path 'defs/ethereum/aux.png'` on Windows?

Interesting! That will be some recursive submodule stuff. Is Windows build doing a recursive submodule update? @tsusanka any chance you experienced it?

selsta:
> Is Windows build doing a recursive submodule update?

Yes, it usually works fine. Looks like updating the submodule in this PR broke it.

tsusanka:
> @tsusanka any chance you experienced it?

We did! Funny thing - it turns out you can't have filename called "aux" on windows. See https://github.com/trezor/trezor-firmware/issues/870 and this commit https://github.com/trezor/trezor-firmware/commit/ae7522165e33f96ad0d5664fd23a0ec1ecc54278 fixes it, so updating the submodule should fix the problem.

You are however using trezor-common repo right? It's a bit behind trezor-firmware's master so let me update it first.

ph4r05:
Aaah that will be it! Thanks a lot. Yes, we use trezor-common repo here.

tsusanka:
@ph4r05 trezor-common updated, try to bump the submodule and it should be fixed :tada:.

ph4r05:
Great! CI is now passing

ph4r05:
Jut an info: I will add some more functionality related to the CLSAG and new protocol features in the next 2 days.

SarangNoether:
> Jut for info: I will add some more functionality related to the CLSAG and new protocol features in the next 2 days.

The master branch does not have CLSAG code merged. Perhaps consider adding such functionality to [this branch](https://github.com/SarangNoether/monero/tree/clsag-device) instead.

ph4r05:
> > Jut for info: I will add some more functionality related to the CLSAG and new protocol features in the next 2 days.
> 
> The master branch does not have CLSAG code merged. Perhaps consider adding such functionality to [this branch](https://github.com/SarangNoether/monero/tree/clsag-device) instead.

Ah OK, sure. Once this PR is merged to monero master I can create a new PR for CLSAG as I will need changes from this branch.

I also could do both PRs and keep them in sync but it seems there is enough time to do it sequentially as it is easier.

ph4r05:
I am done. Ready for review :)

src/device_trezor/trezor/protocol.cpp
```diff
@@ -500,12 +551,19 @@ namespace tx {
...
     tsx_data.set_account(tx.subaddr_account);
     tsx_data.set_monero_version(std::string(MONERO_VERSION) + "|" + MONERO_VERSION_TAG);
     tsx_data.set_hard_fork(m_aux_data->hard_fork ? m_aux_data->hard_fork.get() : 0);
-    assign_to_repeatable(tsx_data.mutable_minor_indices(), tx.subaddr_indices.begin(), tx.subaddr_indices.end());
+
+    if (client_version() <= 1){
+      assign_to_repeatable(tsx_data.mutable_minor_indices(), tx.subaddr_indices.begin(), tx.subaddr_indices.end());
+    }
+
+    if (tsx_data.hard_fork() >= 13){
```
moneromooo-monero, line 564: Could use HF_VERSION_CLSAG later when that patch is merged. Also did you mean < ?

src/device_trezor/trezor/protocol.cpp
```diff
@@ -976,6 +991,41 @@ namespace tx {
...
+    // Opening the sealed signatures
+    if (client_version() >= 3){
+      if(!ack->has_opening_key()){
+        throw exc::ProtocolException("Client version 3+ requires sealed signatures");
+      }
+
+      for(size_t i = 0; i < m_ct.signatures.size(); ++i){
+        std::string nonce = compute_sealing_key(ack->opening_key(), i, true);
+        std::string key = compute_sealing_key(ack->opening_key(), i, false);
+        size_t plen = m_ct.signatures[i].size() - crypto::chacha::TAG_SIZE;
```
moneromooo-monero, line 1004: Is this checked somewhere so that can now wrap down to positive ?

src/device_trezor/trezor/protocol.cpp
```diff
@@ -976,6 +991,41 @@ namespace tx {
...
+        std::unique_ptr<uint8_t[]> plaintext(new uint8_t[plen]);
```
moneromooo-monero, line 1005: Though I suppose this would throw anyway...

src/device_trezor/trezor/protocol.hpp
```diff
@@ -221,16 +224,33 @@ namespace tx {
...
       CHECK_AND_ASSERT_THROW_MES(m_tx_idx < m_unsigned_tx->txes.size(), "Invalid transaction index");
       return m_unsigned_tx->txes[m_tx_idx];
     }
 
+    const tx_construction_data & cur_tx(){
+      return m_ct.tx_data;
+    }
+
+    const tools::wallet2::transfer_details & get_transfer(size_t idx){
+      CHECK_AND_ASSERT_THROW_MES(idx < m_unsigned_tx->transfers.second.size() + m_unsigned_tx->transfers.first, "Invalid transfer index");
```
moneromooo-monero, line 237: && idx >= m_unsigned_tx->transfers.first ? Also should be const.

src/device_trezor/trezor/protocol.hpp
```diff
@@ -221,16 +224,33 @@ namespace tx {
     unsigned m_client_version;
     bool m_multisig;
 
-    const tx_construction_data & cur_tx(){
+    const tx_construction_data & cur_src_tx(){
       CHECK_AND_ASSERT_THROW_MES(m_tx_idx < m_unsigned_tx->txes.size(), "Invalid transaction index");
       return m_unsigned_tx->txes[m_tx_idx];
     }
 
+    const tx_construction_data & cur_tx(){
```
moneromooo-monero, line 232: const ?

src/device_trezor/trezor/protocol.hpp
```diff
@@ -221,16 +224,33 @@ namespace tx {
...
+      return m_ct.tx_data;
+    }
+
+    const tools::wallet2::transfer_details & get_transfer(size_t idx){
+      CHECK_AND_ASSERT_THROW_MES(idx < m_unsigned_tx->transfers.second.size() + m_unsigned_tx->transfers.first, "Invalid transfer index");
+      return m_unsigned_tx->transfers.second[idx - m_unsigned_tx->transfers.first];
+    }
+
+    const tools::wallet2::transfer_details & get_source_transfer(size_t idx){
```
moneromooo-monero, line 241: const ?

src/device_trezor/trezor/protocol.hpp
```diff
@@ -221,16 +224,33 @@ namespace tx {
...
+      const auto & sel_transfers = cur_tx().selected_transfers;
+      CHECK_AND_ASSERT_THROW_MES(idx < sel_transfers.size(), "Invalid source index");
+      CHECK_AND_ASSERT_THROW_MES(idx < m_ct.source_permutation.size(), "Invalid source index - permutation");
+      return get_transfer(sel_transfers.at(m_ct.source_permutation[idx]));
```
moneromooo-monero, line 245: You might have meant:

CHECK_AND_ASSERT_THROW_MES(idx < m_ct.source_permutation.size(), "Invalid source index - permutation");
CHECK_AND_ASSERT_THROW_MES(m_ct.source_permutation[idx] < sel_transfers.size(), "Invalid source index");

src/device_trezor/trezor/transport.cpp
```diff
@@ -56,6 +56,11 @@ namespace trezor{
     return true;
   }
 
+  bool t_serialize(const epee::wipeable_string & in, std::string & out){
+    out.assign(in.data(), in.size());
+    return true;
+  }
```
moneromooo-monero, line 62: Does the serializtion buffer get wiped later ?

src/device_trezor/trezor/transport.hpp
```diff
@@ -84,6 +86,15 @@ namespace trezor {
     additional_params.push_back(std::make_pair("Content-Type","application/json; charset=utf-8"));
 
     const http::http_response_info* pri = nullptr;
+    const auto data_cleaner = epee::misc_utils::create_scope_leave_handler([&]() {
+      if (!req_param.empty()) {
+        memwipe(&req_param[0], req_param.size());
+      }
+      if (pri != nullptr && !pri->m_body.empty()){
+        memwipe(&(const_cast<http::http_response_info*>(pri)->m_body)[0], pri->m_body.size());
```
moneromooo-monero, line 94: That's really iffy. Looking at the code, maybe you could add a transport.wipe_response, which would wipe.

src/simplewallet/simplewallet.cpp
```diff
@@ -5576,14 +5576,19 @@ boost::optional<epee::wipeable_string> simple_wallet::on_device_pin_request()
...
-boost::optional<epee::wipeable_string> simple_wallet::on_device_passphrase_request(bool on_device)
+boost::optional<epee::wipeable_string> simple_wallet::on_device_passphrase_request(bool & on_device)
 {
-  if (on_device){
-    message_writer(console_color_white, true) << tr("Please enter the device passphrase on the device");
-    return boost::none;
+  if (on_device) {
+    std::string accepted = input_line(tr(
+        "Device asks for passphrase. Do you want to enter the passphrase on device (Y) (or on the host (N))?"));
+    if (std::cin.eof() || command_line::is_yes(accepted)) {
```
moneromooo-monero, line 5584: eof is really "no TTY, don't prompt". I think I'd cancel in this case.

tests/trezor/daemon.cpp
```diff
@@ -313,7 +313,7 @@ void mock_daemon::mine_blocks(size_t num_blocks, const std::string &miner_addres
 {
   bool blocks_mined = false;
   const uint64_t start_height = get_height();
-  const auto mining_timeout = std::chrono::seconds(30);
+  const auto mining_timeout = std::chrono::seconds(120);
```
moneromooo-monero, line 316: What is the trezor daemon ? I'm surprised to see mining stuff in there.

src/device/device.hpp
```diff
@@ -78,7 +78,7 @@ namespace hw {
         virtual void on_button_request(uint64_t code=0) {}
         virtual void on_button_pressed() {}
         virtual boost::optional<epee::wipeable_string> on_pin_request() { return boost::none; }
-        virtual boost::optional<epee::wipeable_string> on_passphrase_request(bool on_device) { return boost::none; }
+        virtual boost::optional<epee::wipeable_string> on_passphrase_request(bool & on_device) { return boost::none; }
```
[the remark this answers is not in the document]
ph4r05: Makes sense, it's done

src/device_trezor/device_trezor_base.hpp
```diff
@@ -94,7 +94,7 @@ namespace trezor {
 
       std::string m_full_name;
       std::vector<unsigned int> m_wallet_deriv_path;
-      std::string m_device_state;  // returned after passphrase entry, session
+      std::string m_device_session_id;  // returned after passphrase entry, session
```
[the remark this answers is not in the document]
ph4r05: It is not that sensitive as passphrase itself, but yes, it can be wipeable_string. It is now.

src/device_trezor/trezor/protocol.cpp
```diff
@@ -369,6 +359,31 @@ namespace tx {
...
+      throw std::invalid_argument("index is too big");
+    }
+
+    keccak_init(&ctx);
+    keccak_update(&ctx, (const uint8_t *) master_key.data(), master_key.size());
+    keccak_update(&ctx, (const uint8_t *) sep.data(), sep.size());
+    keccak_update(&ctx, hash, 12 - sep.size());
+    keccak_update(&ctx, (const uint8_t *) idx_data.data(), idx_data.size());
+    if (idx_data.size() < 4) {
+      keccak_update(&ctx, hash, 4 - idx_data.size());
```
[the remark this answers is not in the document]
ph4r05: I see. Yes, it is on purpose, we fix field lengths so we need to hash zeros to the full length. For that purpose we use the `hash` array which is initialized to zero. Near the function end the real hash is then computed to the `hash`.

src/device_trezor/trezor/protocol.cpp
ph4r05: Sure! I've added TODO note to adapt this piece of code and use `HF_VERSION_CLSAG` once it is available in the code. 

Here we have `>= 13` as it is normal `if`, even though assert would have `<`

src/device_trezor/trezor/protocol.cpp
ph4r05: Good point! I've added the check to be sure.

src/device_trezor/trezor/protocol.hpp
ph4r05: That is a good catch, thanks!

src/device_trezor/trezor/protocol.hpp
ph4r05: Fair point, thanks!

src/device_trezor/trezor/transport.cpp
ph4r05: Good question! Yes, the caller is responsible for wiping it after usage. 
It is used only here: https://github.com/monero-project/monero/blob/62ca79d96aa3e941f7faa53fc1c78b03bc1a19f7/src/device_trezor/trezor/transport.hpp#L83

The scope leave handler wipes it then.

src/device_trezor/trezor/transport.hpp
ph4r05: I've added `transport.wipe_response`, pls check if that is what you had in mind

src/simplewallet/simplewallet.cpp
ph4r05: Hmm if stdin is eof then we basically say: "Enter passphrase on the device" now, as `boost::none` is returned.

Do you think it would be better to throw an exception which would abort the operation that led here?

tests/trezor/daemon.cpp
ph4r05: This one was a bit tricky. This is libwallet test, testing the whole stack of the Wallet API which is used by the monero-gui for example.

The test constructs the transaction with Trezor. The I need to check whether the transaction is cosntructed properly. Other tests do that by manipulating blockchain and basically also "mining" a new block so the transaction constructed by Trezor is added to a new block. Then the new blockchain is checked and Trezor transaction is validated as it would be by a normal Monero node on mainnet. 

In Wallet API I don't have direct access to blockchain so I just call mining for a while until the Trezor transaction gets into a new block for validation and transaction checks. 
The test also verifies that the new Trezor transaction is correctly detected by Wallet API on the blockchain.

src/device_trezor/device_trezor_base.cpp
```diff
@@ -151,7 +152,8 @@ namespace trezor {
 
     bool device_trezor_base::disconnect() {
       TREZOR_AUTO_LOCK_DEVICE();
-      m_device_state.clear();
+      m_device_session_id.wipe();
```
xiphon, line 155:
```suggestion
```

src/device_trezor/device_trezor_base.cpp
```diff
@@ -361,23 +363,35 @@ namespace trezor {
...
+        tmp_session_id.assign(m_device_session_id.data(), m_device_session_id.size());
+        initMsg->set_allocated_session_id(&tmp_session_id);
       }
 
       m_features = this->client_exchange<messages::management::Features>(initMsg);
-      initMsg->release_state();
+      if (m_features->has_session_id()){
+        m_device_session_id = m_features->session_id();
+      } else {
+        m_device_session_id.wipe();
```
xiphon, line 384:
```suggestion
```

src/device_trezor/device_trezor_base.cpp
```diff
@@ -441,48 +455,90 @@ namespace trezor {
...
+      std::string passphrase_field;
+      messages::common::PassphraseAck m;
+      m.set_on_device(on_device);
+      if (!on_device) {
+        if (!passphrase && m_passphrase) {
+          passphrase = m_passphrase;
+        }
+
+        if (m_passphrase) {
+          m_passphrase->wipe();
```
xiphon, line 513:
```suggestion
```

src/device_trezor/device_trezor_base.cpp
ph4r05: You mean wipe is not needed? I need to invalidate the state, so maybe just `.clear()`?

xiphon: > 
> 
> You mean wipe is not needed? I need to invalidate the state, so maybe just `.clear()`?

Right. You are clearing it on the next line, so wipe is not needed.

ph4r05: Aah I've just double-checked and clear indeed wipes the buffer (as it should). I've just wrongly assumed it just resets the size. Thanks!

src/device_trezor/device_trezor_base.cpp
```diff
@@ -441,48 +453,89 @@ namespace trezor {
...
+          if (*it == messages::management::Features::Capability_PassphraseEntry){
+            on_device = true;
+          }
+        }
       }
 
-      if (!m_device_state.empty()){
-        m.set_allocated_state(&m_device_state);
+      boost::optional<epee::wipeable_string> passphrase;
+      TREZOR_CALLBACK_GET(passphrase, on_passphrase_request, on_device);
```
xiphon, line 500: What happens if the device doesn't support PassphraseEntry (i.e. invoking this with `on_device == false`), but the callback returns `on_device = true`?

ph4r05: Currently, Monero is supported only by Trezor versions that support passphrase entry, i.e., model T. If callback returns `on_device = true` for device not supporting passprase entry, it would lead to error.

It is the same as we would throw exception in the code.

src/device_trezor/trezor/protocol.cpp
moneromooo-monero: But CLSAG are implemented from 13. >= 13, they're valid. < 13, they're not. This is erroring out if using CLSAG at 13 or later.

src/device_trezor/trezor/transport.hpp
moneromooo-monero: No. This change just moves the code, but does the same. What I had in mind is adding a "wipe" function in http_client (or whatever class t_transport is here), then call wipe  instead of calling memwipe. This avoids the const cast.

src/device_trezor/trezor/protocol.cpp
ph4r05: Hmmm by this condition I mean "if hard fork version is >= 13" throw an error as CLSAGs are not yet supported by the Trezor and Monero client code. 

Once https://github.com/trezor/trezor-firmware/pull/942 and CLSAG PRs is merged to the Monero I will remove this condition completely. For now it is just simple check to abort transaction which uses too new hard fork which is not yet suported.

src/device_trezor/trezor/transport.hpp
ph4r05: Aah so direct modification of the simple_http_client, OK I can do that

ph4r05: @moneromooo-monero pls check if this is what you had in mind `transport.wipe_response();`. I've called it `wipe_response` as the client has another fields that could be cleaned in some scenarios, such as http credentials, but we are not using it right now.

- https://github.com/monero-project/monero/pull/6446/files#diff-f53fddf8049bb20838aa3e7698a69509
- https://github.com/monero-project/monero/pull/6446/files#diff-dff12913bf1880e90814097d1eedbcdd
- https://github.com/monero-project/monero/pull/6446/files#diff-fbd330073222eed440fa85924896bcd3

src/device_trezor/trezor/protocol.cpp
moneromooo-monero: Oh, I misunderstood the purpose then, sorry.

src/device_trezor/trezor/transport.hpp
moneromooo-monero: Yes, that's exactly what I intended. Thanks.

ph4r05:
Thanks guys for nice feedback and fast reviews!

Issue #6448, opened 2020-04-13
Regarding simplewallet's new "address one-off <major> <minor>" command


sumogr:
https://github.com/monero-project/monero/pull/6394 seems buggy.
steps to replicate
Create a new wallet, create an one-off subaddress by setting a non existing major index (say 1) and any minor, it still creates a sub address, send some coins back to this address, poof! no key to find them **(do that on testnet of course not with actual xmr :P)**

moneromooo-monero:
Yes, it's annoying. Maybe it should create an account to go with it. Doing so optionally means you can get money and still keep it "hidden" until such time you want to spend it.  Thoughts ?

sumogr:
> Yes, it's annoying. Maybe it should create an account to go with it. Doing so optionally means you can get money and still keep it "hidden" until such time you want to spend it. Thoughts ?

I can think of no other way than creating a new major index (when the chosen account does not exist) along with the one-off. Just think of a way to remind people that they keep hidden money in that account cause there is going to be confusion :D

moneromooo-monero:
That'd be counter productive, wouldn't it. It would not be hidden by definition :P

sumogr:
> That'd be counter productive, wouldn't it. It would not be hidden by definition :P

I guess,  if possible, just a check then that the major index selected for the one-off already exists

moneromooo-monero:
Maybe a "sweep_account" command would be better, since you really don't want to create an account with a very high index, it'd take a lot of time (and space).

sumogr:
true the command can easily be abused. well the advantages of it start to seem less than the disadvantages. I dont know i cant think of anything clever

moneromooo-monero:
https://github.com/monero-project/monero/pull/6449

PR #6449, opened 2020-04-13
simplewallet: add sweep_account command

moneromooo-monero:
Expects an account number, then the usual sweep_all options

Useful to move monero that was accidentally sent to a subaddress
with a very large account index.

PR #6450, opened 2020-04-13
simplewallet: restrict one-off subaddress to current accounts and indices

sumogr:
Along with @moneromooo-monero 's https://github.com/monero-project/monero/pull/6449 resolves https://github.com/monero-project/monero/issues/6448

Edit: Although if you create new addresses on the account containing the one-off up to its index you selected, you have the address back. IMHO a proper name should be hidden subaddress or something catchy in that area

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -9309,6 +9309,16 @@ bool simple_wallet::print_address(const std::vector<std::string> &args/* = std::
       fail_msg_writer() << tr("failed to parse index: ") << local_args[0] << " " << local_args[1];
       return true;
     }
+    if (epee::string_tools::get_xtype_from_string(major, local_args[0]) && epee::string_tools::get_xtype_from_string(minor, local_args[1]) && major > m_wallet->get_num_subaddress_accounts() - 1)
+    {
+      fail_msg_writer() << tr(" Account number selected (major index: ") << local_args[0] << (") does not exist");
+      return true;
+    }
+    if (epee::string_tools::get_xtype_from_string(major, local_args[0]) && epee::string_tools::get_xtype_from_string(minor, local_args[1]) && minor <= m_wallet->get_num_subaddresses(major) - 1)
+    {
+      fail_msg_writer() << tr(" Subaddress index selected (minor index: ") << local_args[1] << (") already exists");
+      return true;
+    }
     m_wallet->create_one_off_subaddress({major, minor});
     success_msg_writer() << boost::format(tr("Address at %u %u: %s")) % major % minor % m_wallet->get_subaddress_as_str({major, minor});
   }


selsta:
> You cannot create a one-off subaddress on your main address' account

Why not? I don’t see any reason to not allow this. Account zero is like any other account.

> Subaddress index selected already exists

I think that’s fine, it should still print out the address. Why disallow it?

> Account number selected does not exist

Also here, why disallow this? With sweep_account this should not be a problem.

sumogr:
>Why not? I don’t see any reason to not allow this. Account zero is like any other account.
clone current master and try creating one

>I think that’s fine, it should still print out the address. Why disallow it?
Cause it beats the purpose of having an one-off (hidden) subaddress

>Account number selected does not exist
Read the issue again. I think you didnt get it. Create an one-off address at an account number your wallet doesnt have and send coins back to this address you wont be able to see them you wont be able to retrieve them and you wont be able to sweep a non existing account

selsta:
> Cause it beats the purpose of having an one-off (hidden) subaddress.

See the help text:

> > If "one-off" is specified, the address for the specified index is generated and displayed, and remembered by the wallet

Nothing here suggests that you can’t use this to display the address of existing subaddresses. This would even be one of my intended use cases. But I’m sure mooo can clarify how he intended this feature.

> You can only retrieve them by creating accounts with account new up to that point so that the one-off pops up

The whole point of `sweep_account` is that you can sweep a whole account without having to create it with `account new`.

sumogr:
>Nothing here suggests that you can’t use this to display the address of existing subaddresses at specific indices. This would even be one of my intended use cases. But I’m sure mooo can clarify how he intended this feature.

Well then that leaves the one-off with no purpose. cause if it can be displayed (although its not with `address all` unless you select an already occupied index or reach up to that index with `address new`) and re-used it is then just another subaddress

>The whole point of sweep_account is that you can sweep a whole account at a specified index without having to create it with account new.

`sweep_account` needs a major index and when you create an one-off at a major index not already existing it wont work cause it wont create one

What i did actually is patch a command with a fuzzy usage whith an obvious bug. I think indeed mooo should clarify it a bit cause i might be of course

moneromooo-monero:
The point of this is to create an address away from what currently exists. If you forbid creating most of them, then the command has not much point left. Originally, the intent was to allow people who mistakenly sent to a random address to recover their funds, though it can also be used on purpose to "hide" monero. For that second usage, your patch would be OK I guess, though it makes brute forcing possible.

sumogr:
>The point of this is to create an address away from what currently exists.

I dont get how you ll be able to recover mistakenly send funds to a random address with an one-off address, I guess I am missing something.
Regarding the hiding part yes I see, but you're gonna lose the funds if the major you pick is not there already (only way to gain access to them is to reach that index by keep on `account new`-ing) and `sweep_account` wouldnt work cause the actual major index number is not created with one-off if not existing. 
Well som one screwing with the core code can of course brute force vulnerabilities but i dont see a reason why would someone want to change that part.
Dont know. Your call :)

moneromooo-monero:
Works for me. Did you try it and it did not work ?

sumogr:
>Originally, the intent was to allow people who mistakenly sent to a random address to recover their funds,
> Works for me. Did you try it and it did not work ?

Send monero from any of my wallet's accounts to a random monero address and then recover them through an one-off address i created earlier? I still cannot comprehend it how is that possible. I might be way off. Sorry for that

moneromooo-monero:
The original problem was:
- Alice creates a subaddress A
- Alice sends Bob A
- Bob sends A some monero
- Alice realizes A's index is some random large numbers (some bug somewhere IIRC)
- Alice can't create that address normally because it would require creating all accounts with index < A's account, major and minor
- Alice can now create a one off address with the right major/minor and sweep the money to wherever (presumably a low account address of hers)

sumogr:
oh the purpose actually is "**create a subadress with any major and minor you select**" feature then. Got it now. Ok got the concept, thank you mooo, patch still applies though cause any major beyond the ones already created would be inaccessible and same goes for minor cause any minor selected that is already there will just give you a subaddress you already "own"

moneromooo-monero:
AFAICT, your patch prevents creation for new accounts, and is a NOP for subaddresses that already exist. Am I wrong ? sweep_account lets you get the money that was sent to whatever index it was sent (unless you found a case where it doesn't). So I think your patch should not be applied, but maybe I'm missing your point here.

sumogr:
> AFAICT, your patch prevents creation for new accounts, and is a NOP for subaddresses that already exist. Am I wrong ? sweep_account lets you get the money that was sent to whatever index it was sent (unless you found a case where it doesn't). So I think your patch should not be applied, but maybe I'm missing your point here.

Correct! 
-Lets say you have a wallet with 3 accounts (major indices 0,1,2)
-Create an one-off address with major 3 and minor whatever
-Send funds at the one-off spawned
i. You wont be able to see what you received
ii. You wont be able to switch to account 3 with `account switch 3` to use those funds
iii.  You cannot use `sweep_account` cause it needs a major index to go along with it and the wallet doesnt even know there is an account 3. so you wont be able to sweep.
The one of code expands the indices to those requested but it doesnt actually "creates" the account so only way to access those funds and sweep them with `sweep_account` or whatever is `account new` till you reach that index and the account is actually created

That's the reasoning behind it. If it still fails to make any sense i m closing it

selsta:
> iii. You cannot use sweep_account cause it needs a major index to go along with it and the wallet doesnt even know there is an account 3. so you wont be able to sweep.

You simply type `sweep_account 3 <address>`

sumogr:
> > iii. You cannot use sweep_account cause it needs a major index to go along with it and the wallet doesnt even know there is an account 3. so you wont be able to sweep.
> 
> You simply type `sweep_account 3 <address>`

if you tested iii and it works and i and ii is of no importance then ok

moneromooo-monero:
I tried iii, and it worked. That was on my work branch before I merged to master though. i is only half correct. You get the notification when you receive the monero, though you don't see it with balance, since ii is correct. ii is of no importance to me however. If it's of importance to people generally, it'd require a overhaul of how addresses are generated wrt lookahead.

sumogr:
Closing

PR #6451, opened 2020-04-15
Bulletproofs: verification speedup

SarangNoether:
Speeds up Bulletproof verification by eliminating unnecessary point compression and decompression when performing multiply-by-8 operations on group elements to ensure proper subgroup membership.

Updates performance and unit tests.

Example speedups on test machine:
- single 1-output proof: 5.2% speedup
- single 2-output proof: 2.9% speedup
- 4-batch of 2-output proofs: 10.6% speedup
- 64-batch of 2-output proofs: 25.7% speedup


SarangNoether:
> Yup, a bunch of necessary field computations gone away.

Hopefully you mean "unnecessary" field computations!

luigi1111:
This is extremely sensitive code: I would like explicit approval from @vtnerd first, and then hopefully get many to run it along-side the current release, before including in a new release (even if it all seems very straightforward).

PR #6452, opened 2020-04-15
FindCcache automatically

mj-xmr:
If ccache is found in the PATH, it is automatically picked up and used by CMake.

### DIFF
--- a/cmake/FindCcache.cmake
+++ b/cmake/FindCcache.cmake
@@ -1,0 +1,57 @@
+# Copyright (c) 2014-2020, The Monero Project
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are
+# permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of
+#    conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this list
+#    of conditions and the following disclaimer in the documentation and/or other
+#    materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may be
+#    used to endorse or promote products derived from this software without specific
+#    prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# - Try to find readline include dirs and libraries 
+#
+# Automatically finds ccache build accelerator, if it's found in PATH.
+#
+# Usage of this module as follows:
+#
+#     project(monero)
+#     include(FindCcache) # Include AFTER the project() macro to be able to reach the CMAKE_CXX_COMPILER variable
+#
+# Properties modified by this module:
+#
+#  GLOBAL PROPERTY RULE_LAUNCH_COMPILE     set to ccache, when ccache found
+#  GLOBAL PROPERTY RULE_LAUNCH_LINK        set to ccache, when ccache found
+
+find_program(CCACHE_FOUND ccache)
+if (CCACHE_FOUND)
+	set(TEMP_CPP_FILE "${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp/test-program.cpp")
+	file(WRITE "${TEMP_CPP_FILE}" "int main() { return 0; }")
+	execute_process(COMMAND "${CCACHE_FOUND}" "${CMAKE_CXX_COMPILER}" "${TEMP_CPP_FILE}"  RESULT_VARIABLE RET)
+	if (${RET} EQUAL 0)
+		message("found usable ccache: ${CCACHE_FOUND}")
+		set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_FOUND}")
+		set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK    "${CCACHE_FOUND}")
+	else()
+		message("found ccache ${CCACHE_FOUND}, but is UNUSABLE! Return code: ${RET}")
+	endif()	    	
+else()
+	message("ccache NOT found!")
+endif()
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,8 @@ message(STATUS "CMake version ${CMAKE_VERSION}")
 
 project(monero)
 
+include(FindCcache) # Has to be included after the project() macro, to be able to read the CXX variable.
+
 enable_language(C ASM)
 
 function (die msg)


sumogr:
i'm asking out of curiosity why dont you include the *.cmake file code directly into cmakelists?

cmake/FindCcache.cmake
```diff
@@ -0,0 +1,50 @@
...
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# - Try to find readline include dirs and libraries 
+#
+# Automatically finds ccache build accelerator, if it's found in PATH.
+# TODO: On Windows, try to find clcache
```
xiphon, line 31: We don't use MSVC

cmake/FindCcache.cmake
```diff
@@ -0,0 +1,50 @@
...
+#
+# Properties modified by this module:
+#
+#  GLOBAL PROPERTY RULE_LAUNCH_COMPILE     set to ccache, when ccache found
+#  GLOBAL PROPERTY RULE_LAUNCH_LINK        set to ccache, when ccache found
+
+find_program(CCACHE_FOUND ccache)
+if (CCACHE_FOUND)
+    message("ccache found")
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
```
xiphon, line 45: Consider to use `${CCACHE_FOUND}` instead of `ccache` here. It's probably worth it to rename `CCACHE_FOUND` then.

mj-xmr: I don't understand. This command is meant to point to the ccache binary in PATH. Replacing `ccache` with the `${CCACHE_FOUND}` will effectively change the binary string type to a boolean type.

Multiple other sources do exactly the same thing:
https://github.com/gammacombo/gammacombo/blob/master/CMakeLists.txt
https://github.com/IENT/vvcDecoderLib/blob/master/CMakeLists.txt
https://stackoverflow.com/questions/1815688/how-to-use-ccache-with-cmake

cmake/FindCcache.cmake
mj-xmr: Will remove the comment.

cmake/FindCcache.cmake
sumogr: He means `set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_FOUND}")` which is more or less the same but more proper

mj-xmr:
> i'm asking out of curiosity why dont you include the *.cmake file code directly into cmakelists?

To prevent the main CMakeLists.txt from becoming polluted with more and more distracting code, or in other words, to keep it modular.

sumogr:
> > i'm asking out of curiosity why dont you include the *.cmake file code directly into cmakelists?
> 
> To prevent the main CMakeLists.txt from becoming polluted with more and more distracting code, or in other words, to keep it modular.

Ok. I would prefer getting cmakelists more "polluted" with a couple of extra lines of code than adding a brand new file for them but this is just me ;)

cmake/FindCcache.cmake
mj-xmr: Indeed. I have tested it, and it works, like you described.

mj-xmr:
> 
> Ok. I would prefer getting cmakelists more "polluted" with a couple of extra lines of code than adding a brand new file for them but this is just me ;)

I agree, that the ccache module is a very bad example of having to write a new file, but doing so enforces structure on other developers, who will more likely follow it. The made effort will yield whenever the other developers' modules are larger.
Thanks for not enforcing your opinion.

cmake/FindCcache.cmake
```diff
@@ -42,9 +41,9 @@
 find_program(CCACHE_FOUND ccache)
 if (CCACHE_FOUND)
     message("ccache found")
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK    ccache)
-else(CCACHE_FOUND)
```
mj-xmr: I removed the (CCACHE_FOUND), from else() since it's kind of oldschool.

mj-xmr:
The CI on clang is failing. I will try to compile a simple hello world program in CMake, before using ccache.

sumogr:
> The CI on clang is failing. I will try to compile a simple hello world program in CMake, before using ccache.

clang is working see the macos ci. Its just travis on FreeBSD cross-compiling failing either create a workaround to fix travis (which i wouldnt want to touch it) or for the time being exclude ccache from FreeBSD (although it works on building natively on FreeBSD). Its getting out of hand too many files for just a simple thing. I dont think it will get approved this way :)

moneromooo-monero:
Is this actually needed ? When ccache is installed, it has symlinks from cc, gcc, clang, etc, so should be picked up automatically, no ?

sumogr:
> Is this actually needed ? When ccache is installed, it has symlinks from cc, gcc, clang, etc, so should be picked up automatically, no ?

I am not familiar with how travis is configured and how travis caches previous builds in general. for some reason gcc and clang are happy in all crossbuilds travis performs besides the one for FreeBSD.  IMHO just disallow your feature for freebsd in cmakelists till someone who can do it configures travis (ok the rare monero developer on freebsd might be sad :) for the time being ). Also squash your commits and since this is a proffesional project test thoroughly on your own repo before you push :) (just a very friendly advice with no bad intentions at all)

mj-xmr:
> Is this actually needed ? When ccache is installed, it has symlinks from cc, gcc, clang, etc, so should be picked up automatically, no ?

It will only work if you prepend the path with the ccache's one. Then the result should be the same as we see here.\ 
As to whether this is needed, please compare the build times between for instance these two branches:
https://travis-ci.org/github/monero-project/monero/builds/675722204 \
https://travis-ci.org/github/monero-project/monero/builds/675745032

sumogr:
> > Is this actually needed ? When ccache is installed, it has symlinks from cc, gcc, clang, etc, so should be picked up automatically, no ?
> 
> It will only work if you prepend the path with the ccache's one. Then the result should be the same as we see here.\
> As to whether this is needed, please compare the build times between for instance these two branches:
> https://travis-ci.org/github/monero-project/monero/builds/675722204
> https://travis-ci.org/github/monero-project/monero/builds/675745032

On that i agree its also mentioned on the bionic man pages for ccache http://manpages.ubuntu.com/manpages/bionic/man1/ccache.1.html

By the way squash some commits

mj-xmr:
> By the way squash some commits

Does it make sense to squash commits after a PR has been created? If yes, then I can do it of course.

mj-xmr:
> > Is this actually needed ? When ccache is installed, it has symlinks from cc, gcc, clang, etc, so should be picked up automatically, no ?
> 
> I am not familiar with how travis is configured and how travis caches previous builds in general. for some reason gcc and clang are happy in all crossbuilds travis performs besides the one for FreeBSD. IMHO just disallow your feature for freebsd in cmakelists till someone who can do it configures travis (ok the rare monero developer on freebsd might be sad :) for the time being ). Also squash your commits and since this is a proffesional project test thoroughly on your own repo before you push :) (just a very friendly advice with no bad intentions at all)

Sure thing.

sumogr:
> > By the way squash some commits
> 
> Does it make sense to squash commits after a PR has been created? If yes, then I can do it of course.

of course it does

cmake/FindCcache.cmake
```diff
@@ -0,0 +1,55 @@
...
+#
+# Properties modified by this module:
+#
+#  GLOBAL PROPERTY RULE_LAUNCH_COMPILE     set to ccache, when ccache found
+#  GLOBAL PROPERTY RULE_LAUNCH_LINK        set to ccache, when ccache found
+
+find_program(CCACHE_FOUND ccache)
+if (CCACHE_FOUND)
+    # Test if works
+    execute_process(COMMAND "${CCACHE_FOUND}" cc "${CMAKE_SOURCE_DIR}/cmake/test-program.cpp" RESULT_VARIABLE RET)
```
TheCharlatan, line 44: I'd prefer the test program inline. Check the CmakeTestCCompiler for an example, then you can also write the file to the temporary cmake directory and properly consume its output. So something along the lines of:
```
 file(WRITE ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp/test-program.cpp  
    "int main() {
    "    return 0;"
    "}")
```
and then running either `try_compile` (for better output) or execute_process.

moneromooo-monero:
Can't see (needs js) but if it helps at least some platforms that don't have it setup transparently, sounds good.

moneromooo-monero:
FWIW, this is what I get from a clean build currently:

-- The C compiler identification is GNU 9.2.1
-- The CXX compiler identification is GNU 9.2.1
-- Check for working C compiler: /usr/lib64/ccache/cc
-- Check for working C compiler: /usr/lib64/ccache/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/lib64/ccache/c++
-- Check for working CXX compiler: /usr/lib64/ccache/c++ -- works

cmake/FindCcache.cmake
mj-xmr: Since this was not yielding a proper result - it didn't detect any failures - and crashed later on anyway, I have decided to scratch this idea. Thanks for the suggestion anyway. \
I will simply detect FreeBSD for now.

sumogr:
Dont struggle with the FreeBSD exclude it this way
`if(NOT CMAKE_SYSTEM_NAME MATCHES "kFreeBSD.*|FreeBSD")`

mj-xmr:
> Dont struggle with the FreeBSD exclude it this way
> `if(NOT CMAKE_SYSTEM_NAME MATCHES "kFreeBSD.*|FreeBSD")`

Thank you. I tried this, but under docker it's all Linux :(
Last try, using the CXX variable.

mj-xmr:
> > Dont struggle with the FreeBSD exclude it this way
> > `if(NOT CMAKE_SYSTEM_NAME MATCHES "kFreeBSD.*|FreeBSD")`
> 
> Thank you. I tried this, but under docker it's all "Linux" :(
> Last try, using the CXX variable.

And it worked :)
Just some final polish for the non-travis CI.

mj-xmr:
> Can't see (needs js) but if it helps at least some platforms that don't have it setup transparently, sounds good.

image

image

cmake/FindCcache.cmake
```diff
@@ -0,0 +1,53 @@
...
+#
+#  GLOBAL PROPERTY RULE_LAUNCH_COMPILE     set to ccache, when ccache found
+#  GLOBAL PROPERTY RULE_LAUNCH_LINK        set to ccache, when ccache found
+
+if("${CMAKE_CXX_COMPILER}" MATCHES "x86_64-unknown-freebsd")
+	message("Temporarily ignoring ccache on FreeBSD, since it breaks cross compilation.")
+else()
+	find_program(CCACHE_FOUND ccache)
+	if (CCACHE_FOUND)
+	    message("ccache found.")
```
xiphon, line 46:
```suggestion
	    message("ccache found ${CCACHE_FOUND}")
```

mj-xmr:
> Tested and decreases Github CI time from ~30mins to ~5mins on Linux / macOS: [selsta@3af7196](https://github.com/selsta/monero/commit/3af719682e65983eeb400cabf2a21b6db0e6e821)
> 
> Will PR the CI changes at a later point once I get Windows working.

Cool.
Since my latest build takes somewhat more than 2.5 minutes, unlike before, I suspect, that the default ccache size is rather too small.

cmake/FindCcache.cmake
mj-xmr: Done.

mj-xmr:
Could somebody tell me if the failing tests are a normal thing?

selsta:
Until #6445 is merged, yes.

mj-xmr:
> I don't think your Workflows CI changes have any effect.
> I guess you can:
> 
>     * cherry-pick @selsta's commit to this PR with appropriate Workflows CI changes
> 
>     * either drop CI changes

I dropped the changes.

mj-xmr:
> FWIW, this is what I get from a clean build currently:
> 
> -- The C compiler identification is GNU 9.2.1
> -- The CXX compiler identification is GNU 9.2.1
> -- Check for working C compiler: /usr/lib64/ccache/cc
> -- Check for working C compiler: /usr/lib64/ccache/cc -- works
> -- Detecting C compiler ABI info
> -- Detecting C compiler ABI info - done
> -- Detecting C compile features
> -- Detecting C compile features - done
> -- Check for working CXX compiler: /usr/lib64/ccache/c++
> -- Check for working CXX compiler: /usr/lib64/ccache/c++ -- works

Looking at the version of the GCC, I assume you use a bleeding edge distribution, which implies bleeding edge CMake, able to do the ccache trick automatically.

moneromooo-monero:
Fair enough, though I've been using ccache like this since pretty much the start, so 4 years. I'm on Fedora 30 atm (and I should upgrade), I might have been on 23 when I started.

cmake/FindCcache.cmake
mj-xmr: I'm thinking now, that perhaps the idea wasn't bad at all, but there were two problems:
1) I called "${CCACHE_FOUND}" **cc** , while the CXX variable should have been used
2) The CXX variable wasn't even available before I moved the FindCCache() macro BELOW the project(monero) macro. 
I'll give it a try now.

mj-xmr: Workee!

selsta:
Can you remove the `p2p: add another seed node` commit?

mj-xmr:
> Can you remove the `p2p: add another seed node` commit?

Can you tell me how I should do it exactly before I screw up even more?

selsta:
Force push might be enough.

mj-xmr:
> Force push might be enough.

I think it should be fine now.

mj-xmr:
> clang is working see the macos ci. Its just travis on FreeBSD cross-compiling failing either create a workaround to fix travis (which i wouldnt want to touch it) or for the time being exclude ccache from FreeBSD (although it works on building natively on FreeBSD). Its getting out of hand too many files for just a simple thing. I dont think it will get approved this way :)

Problem fixed. See: https://github.com/monero-project/monero/pull/6452#discussion_r410944099

moneromooo-monero:
The commit message seems to be a jumble of random text.

mj-xmr:
> The commit message seems to be a jumble of random text.

Fixed.

mj-xmr:
I'm done. No more ideas from my side on this one.

vtnerd:
So I just noticed there is no opt-in or opt-out strategy for this. Seems to force the usage of `ccache` if installed ... ? Any reason why, or an oversight?

selsta:
I would say oversight.

mj-xmr:
Thanks for handling this PR.
Regarding forcing of ccache - this is fixable. I'll start another PR, where we can discuss this.

mj-xmr:
> So I just noticed there is no opt-in or opt-out strategy for this. Seems to force the usage of `ccache` if installed ... ? Any reason why, or an oversight?

As requested:
https://github.com/monero-project/monero/pull/6501

PR #6455, opened 2020-04-16
Wrap boost serialization

mj-xmr:
A POC of wrapping boost::serialization and boost::archive. \
\
The point here is to group together parts, which are meant to be compiled together: the boost serialization headers and the serialization of the classes itself. This ensures, that the serialization, and most importantly, the boost headers, aren't compiled at every incremental change of the serialized classes. This reduces the compilation time of these parts by about 20 seconds for each large file. Naturally the 20 seconds have to be spent in the wrappers, but these parts will be recompiled rarely. \
\
An additional side effect is grouping and wrapping the serialization in a central place, allowing to easily switch the underlying library at a later point. \
\
Also changed the header inclusion order in some files. Reference:
https://stackoverflow.com/questions/2762568/c-c-include-header-file-order

### DIFF
--- a/src/ringct/bulletproofs.cc
+++ b/src/ringct/bulletproofs.cc
@@ -905,7 +905,7 @@ bool bulletproof_VERIFY(const std::vector<const Bulletproof*> &proofs)
   rct::key m_y0 = rct::zero(), y1 = rct::zero();
   int proof_data_index = 0;
   rct::keyV w_cache;
-  rct::keyV proof8_V, proof8_L, proof8_R;
+  std::vector<ge_p3> proof8_V, proof8_L, proof8_R;
   for (const Bulletproof *p: proofs)
   {
     const Bulletproof &proof = *p;
@@ -918,13 +918,17 @@ bool bulletproof_VERIFY(const std::vector<const Bulletproof*> &proofs)
     const rct::key weight_z = rct::skGen();
 
     // pre-multiply some points by 8
-    proof8_V.resize(proof.V.size()); for (size_t i = 0; i < proof.V.size(); ++i) proof8_V[i] = rct::scalarmult8(proof.V[i]);
-    proof8_L.resize(proof.L.size()); for (size_t i = 0; i < proof.L.size(); ++i) proof8_L[i] = rct::scalarmult8(proof.L[i]);
-    proof8_R.resize(proof.R.size()); for (size_t i = 0; i < proof.R.size(); ++i) proof8_R[i] = rct::scalarmult8(proof.R[i]);
-    rct::key proof8_T1 = rct::scalarmult8(proof.T1);
-    rct::key proof8_T2 = rct::scalarmult8(proof.T2);
-    rct::key proof8_S = rct::scalarmult8(proof.S);
-    rct::key proof8_A = rct::scalarmult8(proof.A);
+    proof8_V.resize(proof.V.size()); for (size_t i = 0; i < proof.V.size(); ++i) rct::scalarmult8(proof8_V[i], proof.V[i]);
+    proof8_L.resize(proof.L.size()); for (size_t i = 0; i < proof.L.size(); ++i) rct::scalarmult8(proof8_L[i], proof.L[i]);
+    proof8_R.resize(proof.R.size()); for (size_t i = 0; i < proof.R.size(); ++i) rct::scalarmult8(proof8_R[i], proof.R[i]);
+    ge_p3 proof8_T1;
+    ge_p3 proof8_T2;
+    ge_p3 proof8_S;
+    ge_p3 proof8_A;
+    rct::scalarmult8(proof8_T1, proof.T1);
+    rct::scalarmult8(proof8_T2, proof.T2);
+    rct::scalarmult8(proof8_S, proof.S);
+    rct::scalarmult8(proof8_A, proof.A);
 
     PERF_TIMER_START_BP(VERIFY_line_61);
     sc_mulsub(m_y0.bytes, proof.taux.bytes, weight_y.bytes, m_y0.bytes);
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ include(CheckLinkerFlag)
 include(CheckLibraryExists)
 include(CheckFunctionExists)
 include(FindPythonInterp)
+# include(FindCcache) temporarily disabled
```
mj-xmr, line 41: Will fix it in the other PR. I already have a plan.

moneromooo-monero:
Well, that's exactly the huge diffs that suck.
20 seconds out of how many are gained ?

mj-xmr:
> Well, that's exactly the huge diffs that suck.
> 20 seconds out of how many are gained ?

Out of half an hour, but you can imagine, that it's the 20 seconds, that you won't have to wait, if you change something, that doesn't have to do anything with serialization. \ 
\ 
Getting a large project in shape is a long process. In my own project a single file compiles in less than 1.5 second, or it's an incorrectly designed file.

vtnerd:
I'm not sure if `BoostArchiveWrapper` actually works (the archives carry state so you'd have to use statics to "recover" the archive) or whether its stubbed, but there is already [`boost::poymorphic_oarchive`](https://www.boost.org/doc/libs/1_39_0/libs/serialization/doc/archive_reference.html#polymorphic), and an `iarchive` instance too. That's the built-in way of splitting the header/cpp since Boost 1.39+. The only difference from the demo is that you can "forward" both of the cpp functions to a single templated function in the cpp to prevent duplication when listing the fields. Macros can assist with the drudgery of listing the stub functions in header and cpp (or just type the small amount of C++).

vtnerd:
And as per the 20 seconds, how did you determine that? And is the code functional the same? Because 20 seconds improvement on partially working code is a vague improvement ...

mj-xmr:
> And as per the 20 seconds, how did you determine that? And is the code functional the same? Because 20 seconds improvement on partially working code is a vague improvement ...

By comparing how long it compiles before the change and after the change :)
The code should function the same. No changes there.

mj-xmr:
> 20 seconds out of how many are gained ?

It was 40 for the single file before, if that's what you're asking.

mj-xmr:
> As a proof of concept it definitely has some validity. What really scares me is "partially" reorganising and restructuring such a complex project with years upon years of commits. If it was a new project built from scratch it would be a different case. What I would love to see though would be monero switched to C++17 and replacing boost constructs wherever possible with standard library

I totally get your concerns. I must also admit, that if I were to fixup the whole project, I'd have to spend good 2 months.
I'm fine if you guys decide, that you don't want this. Please tell me if I should continue.

sumogr:
LOL no i am not saying that you should do it. I am nobody in this. mooo and vtnerd will give a final review and approve or not

moneromooo-monero:
> switched to C++17 and replacing boost constructs

Like what ?

sumogr:
> > switched to C++17 and replacing boost constructs
> 
> Like what ?

regex, optional, filesystem, variant to name a few

vtnerd:
>  I'm fine if you guys decide, that you don't want this. Please tell me if I should continue.

Any response to the boost polymorphic archive stuff ?

mj-xmr:
> > I'm fine if you guys decide, that you don't want this. Please tell me if I should continue.
> 
> Any response to the boost polymorphic archive stuff ?

I will have a look for sure. For now I wanted to get the CCache into a working state, since it was blocking my workflow.

mj-xmr:
> Any response to the boost polymorphic archive stuff ?

1) The concrete archive types were failing to become converted to the polymorphic types for some reason
2) I still prefer some small boilerplate C++ code over macros, as any MISRA check would label macros as invalid code, for the reason, that it's quite impossible to debug it.
3) I have decided to back off from changing the p2p code, for the reasons, that it was too large a change in one PR and because it was failing a unit test, even though I haven't changed, the functionality. So I hoped at least. My initial intention was to provide more examples of the wrapper's usage.

moneromooo-monero:
>  I'm fine if you guys decide, that you don't want this. Please tell me if I should continue.

I don't know. It depends on what is the end result in terms of invasive changes and speedup. So you can continue if you think you'll get major speedups without too massive diffs.

vtnerd:
> The concrete archive types were failing to become converted to the polymorphic types for some reason

We have a custom boost archive format. Update its inheritance to `boost::polymorphic_iarchive` and `boost::polymorphic_oarchive`, and mark the functions as `virtual` + `override final`. Fill in missing functions.

If any of the archive types carried state, this code wouldn't work. Luckily, I think none of them carry state (outside of the stream object), but I don't see this being specified anywhere in the documentation. Just use the polymorphic archives, unless I'm missing something ?

> I still prefer some small boilerplate C++ code over macros, as any MISRA check would label macros as invalid code, for the reason, that it's quite impossible to debug it.

Is this referring to some boost macro? I have no idea why people find macros so abhorrent. I can write impenetrable and/or impossible to use templates or macros. I don't see a reason to demonize either based on some guide attempting to bludgeon programmers into writing magically correct code. And if there is a boost macro that is bothersome, then just type out the code manually instead.

moneromooo-monero:
FWIW you seem to have pushed a bad tree. I see merge commits and commits by vtnerd.

mj-xmr:
> FWIW you seem to have pushed a bad tree. I see merge commits and commits by vtnerd.

Yes. That was my mistake.
I don't understand what github has done now, however. After I realized my mistake, I flagged the branch as Draft, then rebased it locally nicely, then force pushed, and suddenly I see, that the branch was closed by me, which is not true.
Anyway, I'll start again the same thing but in smaller steps, taking the CI more seriously. Please don't delete the branch yet, to keep the valuable comments.

PR #6466, opened 2020-04-19
simplewallet: print locked blocks upon locked tx receipt

sumogr:
Well since i misunderstood https://github.com/monero-project/monero/issues/6463 and the branch started on the wrong foot at least make some use of it than deleting it :)


sumogr:
:(

PR #6471, opened 2020-04-22
Seed nodes

selsta:
So that we don’t have to rebase constantly. If you want to add your seed node comment.


Gingeropolous:
I can confirm that these 2 are managed by me:

   full_addrs.insert("104.238.221.81:18080");
      full_addrs.insert("66.85.74.134:18080");

PR #6472, opened 2020-04-22
cryptonote: fix reuse of non default tx data when relaying

moneromooo-monero:
An automatic tx variable is initialized properly on the first
run through the loop, but not the second. Moving the variable
inside the loop ensures the ctor is called again to init it.

### DIFF
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -146,6 +146,8 @@ namespace cryptonote
     if (tx.version >= 2 && !is_coinbase(tx))
     {
       rct::rctSig &rv = tx.rct_signatures;
+      if (rv.type == rct::RCTTypeNull)
+        return true;
       if (rv.outPk.size() != tx.vout.size())
       {
         LOG_PRINT_L1("Failed to parse transaction from blob, bad outPk size in tx " << get_transaction_hash(tx));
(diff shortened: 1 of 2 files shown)


src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -146,6 +146,8 @@ namespace cryptonote
     if (tx.version >= 2 && !is_coinbase(tx))
     {
       rct::rctSig &rv = tx.rct_signatures;
+      if (rv.type == rct::RCTTypeNull)
```
vtnerd, line 149: Is this related?

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1308,9 +1308,9 @@ namespace cryptonote
     std::vector<crypto::hash> tx_hashes{};
     tx_hashes.resize(tx_blobs.size());
 
-    cryptonote::transaction tx{};
     for (std::size_t i = 0; i < tx_blobs.size(); ++i)
     {
+      cryptonote::transaction tx{};
```
vtnerd, line 1313: Crap. The internal memory gets freed every cycle, but there is no easier fix.

src/cryptonote_basic/cryptonote_format_utils.cpp
moneromooo-monero: This is what made me find that bug in the first place. I guess it can be omitted since all null rct types are coinbase so will be caught by the previous check. But pedantically it's best to check the rct data exists before we get it below.

dEBRUYNE-1:
@vtnerd - Can this be merged in the current state or do you have any further comments?

PR #6474, opened 2020-04-24
Update in-memory key encryption and unit test

SarangNoether:
Simplifies and updates in-memory key encryption. Updates unit test to include multisignature keys.


SarangNoether:
Currently fails this assertion: https://github.com/SarangNoether/monero/blob/key-chacha/tests/unit_tests/serialization.cpp#L633

SarangNoether:
Needs additional updates for compatibility with existing key files.

PR #6480, opened 2020-04-26
wallet2: check_connection return false on get_version status != OK

xiphon:


moneromooo-monero:
Why, and in what case does this happen ?

xiphon:
> Why, and in what case does this happen ?

Happens due to https://github.com/monero-project/monero/pull/6481 bug. Fixing just the latter will be enough.

OTOH, should `check_connection(uint32_t *version, ...)` return `true` and `version = 0` if it successfully connected to *something*, but `get_version` request failed?

moneromooo-monero:
That's what I was wondering. I guess it made sense earlier because get_version was new but after thinking some more I think your patch is better.

PR #6487, opened 2020-04-29
txpool.cpp: rename var to fix for old g++ version (xenial default)

sumogr:
makes silly g++ 5.4.0 happy by removing the confusion for `const auto relay_method` var so as it compiles on xenial


sumogr:
> Sorry for bikeshedding a bit, but how about something a little less typoey, like tx_relay_method or the like ? :)

sure :)

sumogr:
> gcc++ 5.5.0 is also unhappy.

what do you mean? that it was not compiling for 5.5.0 as well? well probably, i dont know when they fixed that but it should compile in all versions now. I just happened to notice it on my xenial which was "wearing" 5.4.0 by default. (i tried it in 5.4.0 and 7.5.0 and its fine)

trasherdk:
Without this patch, building with 5.5.0 failed.

PR #6488, opened 2020-04-29
simplewallet: add show_qr_code command

selsta:
- Untested on Windows

Screenshot 2020-04-29 at 18 59 47

### DIFF
--- a/external/qrcodegen/CMakeLists.txt
+++ b/external/qrcodegen/CMakeLists.txt
@@ -1,0 +1,7 @@
+project(libqrcodegen)
+
+add_library(qrcodegen STATIC QrCode.cpp)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
+
+target_include_directories(qrcodegen PUBLIC 
+    ${CMAKE_CURRENT_SOURCE_DIR})
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -69,10 +69,12 @@
 #include "version.h"
 #include <stdexcept>
 #include "wallet/message_store.h"
+#include "QrCode.hpp"
 
 #ifdef WIN32
 #include <boost/locale.hpp>
 #include <boost/filesystem.hpp>
+#include <fcntl.h>
 #endif
 
 #ifdef HAVE_READLINE
@@ -268,6 +270,7 @@ namespace
   const char* USAGE_RPC_PAYMENT_INFO("rpc_payment_info");
   const char* USAGE_START_MINING_FOR_RPC("start_mining_for_rpc");
   const char* USAGE_STOP_MINING_FOR_RPC("stop_mining_for_rpc");
+  const char* USAGE_SHOW_QR_CODE("show_qr_code [<subaddress_index>]");
   const char* USAGE_VERSION("version");
   const char* USAGE_HELP_ADVANCED("help_advanced [<command>]");
   const char* USAGE_HELP("help");
@@ -2387,6 +2390,62 @@ bool simple_wallet::stop_mining_for_rpc(const std::vector<std::string> &args)
   return true;
 }
 
+bool simple_wallet::show_qr_code(const std::vector<std::string> &args)
+{
+  uint32_t subaddress_index = 0;
+  if (args.size() >= 1)
+  {
+    if (!string_tools::get_xtype_from_string(subaddress_index, args[0]))
+    {
+      fail_msg_writer() << tr("invalid index: must be an unsigned integer");
+      return true;
+    }
+    if (subaddress_index >= m_wallet->get_num_subaddresses(m_current_subaddress_account))
+    {
+      fail_msg_writer() << tr("<subaddress_index> is out of bounds");
+      return true;
+    }
+  }
+
+#ifdef _WIN32
+#define PRINT_UTF8(pre, x) std::wcout << pre ## x
+#define WTEXTON() _setmode(_fileno(stdout), _O_WTEXT)
+#define WTEXTOFF() _setmode(_fileno(stdout), _O_TEXT)
+#else
+#define PRINT_UTF8(pre, x) std::cout << x
+#define WTEXTON()
+#define WTEXTOFF()
+#endif
+
+  WTEXTON();
+  try
+  {
+    const std::string address = "monero:" + m_wallet->get_subaddress_as_str({m_current_subaddress_account, subaddress_index});
+    const qrcodegen::QrCode qr = qrcodegen::QrCode::encodeText(address.c_str(), qrcodegen::QrCode::Ecc::LOW);
+    for (int y = -2; y < qr.getSize() + 2; y+=2)
+    {
+      for (int x = -2; x < qr.getSize() + 2; x++)
+      {
+        if (qr.getModule(x, y) && qr.getModule(x, y + 1))
+          PRINT_UTF8(L, "\u2588");
+        else if (qr.getModule(x, y) && !qr.getModule(x, y + 1))
+          PRINT_UTF8(L, "\u2580");
+        else if (!qr.getModule(x, y) && qr.getModule(x, y + 1))
+          PRINT_UTF8(L, "\u2584");
+        else
+          PRINT_UTF8(L, " ");
+      }
+      PRINT_UTF8(, std::endl);
+    }
+  }
+  catch (const std::length_error&)
+  {
+    fail_msg_writer() << tr("Failed to generate QR code, input too large");
+  }
+  WTEXTOFF();
+  return true;
+}
+
 bool simple_wallet::set_always_confirm_transfers(const std::vector<std::string> &args/* = std::vector<std::string>()*/)
 {
   const auto pwd_container = get_and_verify_password();
@@ -3610,6 +3669,10 @@ simple_wallet::simple_wallet()
                            boost::bind(&simple_wallet::stop_mining_for_rpc, this, _1),
                            tr(USAGE_STOP_MINING_FOR_RPC),
                            tr("Stop mining to pay for RPC access"));
+  m_cmd_binder.set_handler("show_qr_code",
+                           boost::bind(&simple_wallet::on_command, this, &simple_wallet::show_qr_code, _1),
+                           tr(USAGE_SHOW_QR_CODE),
+                           tr("Show address as QR code"));
   m_cmd_binder.set_handler("help_advanced",
                            boost::bind(&simple_wallet::on_command, this, &simple_wallet::help_advanced, _1),
                            tr(USAGE_HELP_ADVANCED),
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -80,4 +80,5 @@ endif()
 
 add_subdirectory(db_drivers)
 add_subdirectory(easylogging++)
+add_subdirectory(qrcodegen)
 add_subdirectory(randomx EXCLUDE_FROM_ALL)
--- a/external/qrcodegen/QrCode.cpp
+++ b/external/qrcodegen/QrCode.cpp
@@ -1,0 +1,862 @@
+/* 
+ * QR Code generator library (C++)
+ * 
+ * Copyright (c) Project Nayuki. (MIT License)
+ * https://www.nayuki.io/page/qr-code-generator-library
+ * 
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ * - The above copyright notice and this permission notice shall be included in
+ *   all copies or substantial portions of the Software.
+ * - The Software is provided "as is", without warranty of any kind, express or
+ *   implied, including but not limited to the warranties of merchantability,
+ *   fitness for a particular purpose and noninfringement. In no event shall the
+ *   authors or copyright holders be liable for any claim, damages or other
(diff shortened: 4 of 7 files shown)


src/simplewallet/simplewallet.cpp
```diff
@@ -2386,6 +2388,43 @@ bool simple_wallet::stop_mining_for_rpc(const std::vector<std::string> &args)
...
+    }
+    if (subaddress_index >= m_wallet->get_num_subaddresses(m_current_subaddress_account))
+    {
+      fail_msg_writer() << tr("<subaddress_index> is already out of bound");
+      return true;
+    }
+  }
+
+  const std::string address = "monero:" + m_wallet->get_subaddress_as_str({m_current_subaddress_account, subaddress_index});
+  const qrcodegen::QrCode qr = qrcodegen::QrCode::encodeText(address.c_str(), qrcodegen::QrCode::Ecc::LOW);
```
selsta, line 2409: This can throw when the input string is too long. It is unlikely that this will happen here. Should I still add a try / catch?

iDunk5400:
I tried this on Windows and...

monero-wallet-cli_qrcode_nok

This ugly diff fixed it for me
```diff
diff --git a/src/simplewallet/simplewallet.cpp b/src/simplewallet/simplewallet.cpp
index c3ac8e895..441613229 100644
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -74,6 +74,7 @@
 #ifdef WIN32
 #include <boost/locale.hpp>
 #include <boost/filesystem.hpp>
+#include <fcntl.h>
 #endif

 #ifdef HAVE_READLINE
@@ -2390,6 +2391,9 @@ bool simple_wallet::stop_mining_for_rpc(const std::vector<std::string> &args)

 bool simple_wallet::show_qr_code(const std::vector<std::string> &args)
 {
+#ifdef WIN32
+  _setmode(_fileno(stdout), _O_WTEXT);
+#endif
   uint32_t subaddress_index = 0;
   if (args.size() >= 1)
   {
@@ -2412,16 +2416,39 @@ bool simple_wallet::show_qr_code(const std::vector<std::string> &args)
     for (int x = -2; x < qr.getSize() + 2; x++)
     {
       if (qr.getModule(x, y) && qr.getModule(x, y + 1))
+#ifdef WIN32
+        std::wcout << L"\u2588";
+#else
         std::cout << "\u2588";
+#endif
       else if (qr.getModule(x, y) && !qr.getModule(x, y + 1))
+#ifdef WIN32
+        std::wcout << L"\u2580";
+#else
         std::cout << "\u2580";
+#endif
       else if (!qr.getModule(x, y) && qr.getModule(x, y + 1))
+#ifdef WIN32
+        std::wcout << L"\u2584";
+#else
         std::cout << "\u2584";
+#endif
       else
+#ifdef WIN32
+        std::wcout << L" ";
+#else
         std::cout << " ";
+#endif
     }
+#ifdef WIN32
+    std::wcout << std::endl;
+#else
     std::cout << std::endl;
+#endif
   }
+#ifdef WIN32
+  _setmode(_fileno(stdout), _O_TEXT);
+#endif
   return true;
 }

```

monero-wallet-cli_qrcode_ok

And my phone recognizes the QR code.

external/qrcodegen/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+project(libqrcodegen)
+
+add_library(qrcodegen STATIC QrCode.cpp)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
```
moneromooo-monero, line 4: Does -fPIC not get passed already ? If it doesn't, do the security flags get not passed either ?

src/simplewallet/simplewallet.cpp
```diff
@@ -2386,6 +2388,43 @@ bool simple_wallet::stop_mining_for_rpc(const std::vector<std::string> &args)
...
+  if (args.size() >= 1)
+  {
+    if (!string_tools::get_xtype_from_string(subaddress_index, args[0]))
+    {
+      fail_msg_writer() << tr("invalid count: must be an unsigned integer");
+      return true;
+    }
+    if (subaddress_index >= m_wallet->get_num_subaddresses(m_current_subaddress_account))
+    {
+      fail_msg_writer() << tr("<subaddress_index> is already out of bound");
```
moneromooo-monero, line 2403: already ? and bounds.

src/simplewallet/simplewallet.cpp
```diff
@@ -2386,6 +2388,43 @@ bool simple_wallet::stop_mining_for_rpc(const std::vector<std::string> &args)
...
 }
 
+bool simple_wallet::show_qr_code(const std::vector<std::string> &args)
+{
+  uint32_t subaddress_index = 0;
+  if (args.size() >= 1)
+  {
+    if (!string_tools::get_xtype_from_string(subaddress_index, args[0]))
+    {
+      fail_msg_writer() << tr("invalid count: must be an unsigned integer");
```
moneromooo-monero, line 2398: s/count/index/ ?

src/simplewallet/simplewallet.cpp
```diff
@@ -2386,6 +2388,43 @@ bool simple_wallet::stop_mining_for_rpc(const std::vector<std::string> &args)
...
+    if (subaddress_index >= m_wallet->get_num_subaddresses(m_current_subaddress_account))
+    {
+      fail_msg_writer() << tr("<subaddress_index> is already out of bound");
+      return true;
+    }
+  }
+
+  const std::string address = "monero:" + m_wallet->get_subaddress_as_str({m_current_subaddress_account, subaddress_index});
+  const qrcodegen::QrCode qr = qrcodegen::QrCode::encodeText(address.c_str(), qrcodegen::QrCode::Ecc::LOW);
+  for (int y = -2; y < qr.getSize() + 2; y+=2)
```
moneromooo-monero, line 2410: Are those odd bounds part of the API you use ? It sounds pretty dodgy...

src/simplewallet/simplewallet.cpp
```diff
@@ -2386,6 +2388,43 @@ bool simple_wallet::stop_mining_for_rpc(const std::vector<std::string> &args)
...
+  {
+    for (int x = -2; x < qr.getSize() + 2; x++)
+    {
+      if (qr.getModule(x, y) && qr.getModule(x, y + 1))
+        std::cout << "\u2588";
+      else if (qr.getModule(x, y) && !qr.getModule(x, y + 1))
+        std::cout << "\u2580";
+      else if (!qr.getModule(x, y) && qr.getModule(x, y + 1))
```
moneromooo-monero, line 2418: Calls to getModule could be cached if they're expensive. Don't bother it it's mostly just a lookup.

moneromooo-monero:
The QrCode files seem substantially different from the ones in the GUI repo. Is there a reason for this ? I did not review the QrCode code, I was hoping I could just check it's the same as in the GUI :)

selsta:
I didn’t even realize that the GUI uses the same library. I took the code from here: https://github.com/nayuki/QR-Code-generator/tree/master/cpp

external/qrcodegen/CMakeLists.txt
selsta: I had to add `-fPIC` to get Travis CI cross compilation working.

The security flags probably don’t get passed either.

src/simplewallet/simplewallet.cpp
selsta: The odd bounds are used to add a border around the QR code. The API demo code does the same and the API seems to check for it: https://github.com/nayuki/QR-Code-generator/blob/master/cpp/QrCode.cpp#L384

src/simplewallet/simplewallet.cpp
selsta: Displaying a QR code is basically instant. I don’t think there would a noticeable difference. Would you still prefer to cache it?

src/simplewallet/simplewallet.cpp
selsta: thanks, fixed. looks like I copied the error message without reading them...

src/simplewallet/simplewallet.cpp
moneromooo-monero: If it's ~instant for the user, no need.

external/qrcodegen/CMakeLists.txt
moneromooo-monero: I don't either, I just hack at it till it works :D It can stay that way, I or someone will fix it later.

src/simplewallet/simplewallet.cpp
moneromooo-monero: Might as well, easy to add just in case.

src/simplewallet/simplewallet.cpp
```diff
@@ -3610,6 +3657,10 @@ simple_wallet::simple_wallet()
                            boost::bind(&simple_wallet::stop_mining_for_rpc, this, _1),
                            tr(USAGE_STOP_MINING_FOR_RPC),
                            tr("Stop mining to pay for RPC access"));
+  m_cmd_binder.set_handler("show_qr_code",
+                           boost::bind(&simple_wallet::show_qr_code, this, _1),
```
moneromooo-monero, line 3661: on_command

src/simplewallet/simplewallet.cpp
```diff
@@ -2387,6 +2389,51 @@ bool simple_wallet::stop_mining_for_rpc(const std::vector<std::string> &args)
...
+    }
+  }
+
+  try
+  {
+    const std::string address = "monero:" + m_wallet->get_subaddress_as_str({m_current_subaddress_account, subaddress_index});
+    const qrcodegen::QrCode qr = qrcodegen::QrCode::encodeText(address.c_str(), qrcodegen::QrCode::Ecc::LOW);
+    for (int y = -2; y < qr.getSize() + 2; y+=2)
+    {
+      for (int x = -2; x < qr.getSize() + 2; x++)
```
moneromooo-monero, line 2415: The sample code from the nayuki library does not go out of bounds like this:

    for (int y = 0; y < qr1.getSize(); y++) {
        for (int x = 0; x < qr1.getSize(); x++) {
            (... paint qr1.getModule(x, y) ...)

The code is safe (and returns false) for those. Is this to ensure a border to have less chance of confusing scanners ? If so, that'd be a good reason I guess.

selsta: That’s where I got the demo code from with the border: https://github.com/nayuki/QR-Code-generator/blob/master/cpp/QrCodeGeneratorDemo.cpp#L190

> Is this to ensure a border to have less chance of confusing scanners ?

Yep, I don’t think it is 100% necessary though.

selsta:
```diff
diff --git a/src/simplewallet/simplewallet.cpp b/src/simplewallet/simplewallet.cpp
index 0725226e1..8546edb32 100644
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -74,6 +74,7 @@
 #ifdef WIN32
 #include <boost/locale.hpp>
 #include <boost/filesystem.hpp>
+#include <fcntl.h>
 #endif
 
 #ifdef HAVE_READLINE
@@ -2406,6 +2407,16 @@ bool simple_wallet::show_qr_code(const std::vector<std::string> &args)
     }
   }
 
+#ifdef WIN32
+#define PRINT_UTF8(x) std::wcout << Lx
+#else
+#define PRINT_UTF8(x) std::cout << x
+#endif
+
+#ifdef WIN32
+  _setmode(_fileno(stdout), _O_WTEXT);
+#endif
+
   try
   {
     const std::string address = "monero:" + m_wallet->get_subaddress_as_str({m_current_subaddress_account, subaddress_index});
@@ -2415,15 +2426,19 @@ bool simple_wallet::show_qr_code(const std::vector<std::string> &args)
       for (int x = -2; x < qr.getSize() + 2; x++)
       {
         if (qr.getModule(x, y) && qr.getModule(x, y + 1))
-          std::cout << "\u2588";
+          PRINT_UTF8("\u2588");
         else if (qr.getModule(x, y) && !qr.getModule(x, y + 1))
-          std::cout << "\u2580";
+          PRINT_UTF8("\u2580");
         else if (!qr.getModule(x, y) && qr.getModule(x, y + 1))
-          std::cout << "\u2584";
+          PRINT_UTF8("\u2584");
         else
-          std::cout << " ";
+          PRINT_UTF8(" ");
       }
+#ifdef WIN32
+      std::wcout << std::endl;
+#else
       std::cout << std::endl;
+#endif
     }
   }
   catch (const std::length_error&)
@@ -2431,6 +2446,10 @@ bool simple_wallet::show_qr_code(const std::vector<std::string> &args)
     fail_msg_writer() << tr("Failed to generate QR code, input too large");
   }
 
+#ifdef WIN32
+  _setmode(_fileno(stdout), _O_TEXT);
+#endif
+
   return true;
 }
 
```

That’s the best I could do with @iDunk5400 patch. Still ugly.

moneromooo-monero:
Good enough. Please use _WIN32 though, IIRC it's more robust. I assume you don't need the fcntl.h header since the new code is all windows ?

iDunk5400:
IIRC, _O_WTEXT and _O_TEXT are defined in fcntl.h. If fcntl.h is omitted, numerical values would have to be used instead. IIRC, WIN32 works once windows.h is included.
It would be good if this was tested with various locales, because I don't know if _O_TEXT is the default everywhere.

Anyway, https://github.com/monero-project/monero/commit/c24095db3711b918517521254ad60c35fd1b8b31 would work best, I think, if it didn't break current implementation of paths/filenames with wide characters, as it would also fix the display of mnemonic seeds in different languages with wide characters on Windows.

moneromooo-monero:
OK. I didn't realize windows would have something very unixy like fcntl.h ^_^

selsta:
@iDunk5400 force pushed, optimized the macro a bit and switched WIN32 to _WIN32. Guess the only thing left now is test with different locales?

PR #6493, opened 2020-05-01
Update in-memory key encryption and unit tests

SarangNoether:


tests/unit_tests/serialization.cpp
```diff
@@ -616,6 +616,25 @@ TEST(Serialization, serializes_ringct_types)
...
+  const cryptonote::network_type nettype = cryptonote::TESTNET;
+  tools::wallet2 w(nettype);
+  const boost::filesystem::path wallet_file = unit_test::data_dir / "wallet_9svHk1";
+  string password = "test";
+  bool r = false;
+  try
+  {
+    w.load(wallet_file.string(), password); // old key method
+    w.rewrite(wallet_file.string(), password);
+    w.load(wallet_file.string(), password); // new key method
```
iamamyth, line 630: It might be good to assert that you get the same keys post-rewrite and load as you did the first time around.

src/cryptonote_basic/account.cpp
```diff
@@ -70,62 +71,81 @@ DISABLE_VS_WARNINGS(4244 4345)
...
+    const char *ptr = stream.data();
     for (size_t i = 0; i < sizeof(crypto::secret_key); ++i)
       m_spend_secret_key.data[i] ^= *ptr++;
     for (size_t i = 0; i < sizeof(crypto::secret_key); ++i)
       m_view_secret_key.data[i] ^= *ptr++;
     for (crypto::secret_key &k: m_multisig_keys)
     {
       for (size_t i = 0; i < sizeof(crypto::secret_key); ++i)
-        k.data[i] ^= *ptr++;
+        k.data[i] ^= *ptr;;
```
iamamyth, line 113: Looks as though this may be an error, which would also suggest a missing test.

tests/unit_tests/serialization.cpp
SarangNoether: I suppose this could be specifically tested. However, the load process does check that the wallet address corresponds correctly to the decrypted keys, so the odds that this would pass with invalid keys is negligible.

A better test would be to ensure that the new method is actually being used, rather than falling back to the legacy method in the event that the keys weren't properly rewritten; however, that would require passing information back from the loader, and it's not clear whether that would cause other problems elsewhere.

src/cryptonote_basic/account.cpp
SarangNoether: Good catch!

SarangNoether: It looks like the unit test wallet file doesn't currently include any multisig addresses.

tests/unit_tests/serialization.cpp
iamamyth: Checking that the new method is used would be reasonable, although you could also just confirm that the serialized representation has changed, which seems good enough.

As for checking the key post-load, I wouldn't want to rely on the implementation detail that the wallet asserts the sanity of the data, unless there's a separate test which explicitly confirms that behavior, which I don't see in this suite. The prior deserialization test, `portability_wallet`, goes to great lengths to validate all the loaded key material, so it seems appropriate to at least confirm that loading an upgraded wallet produces, e.g., the proper public address, as, at the moment, all the prior test suite sanity checks don't happen for the new (and now default) wallet format. In other words, as it stands, for the common case of "create new format wallet, load new format wallet", there's nothing in this test suite confirming data sanity, as was done for the earlier format.

SarangNoether: I'm already making changes to explicitly check the decryption method used, so I can further add a check that the keys reconstruct the proper address.

src/wallet/wallet2.cpp
```diff
@@ -4329,7 +4329,17 @@ bool wallet2::load_keys_buf(const std::string& keys_buf, const epee::wipeable_st
   {
     if (encrypted_secret_keys)
     {
-      m_account.decrypt_keys(key);
+      m_account.decrypt_legacy(key); // try legacy decryption
```
iamamyth, line 4332: I haven't looked closely enough at the new format to be certain if this would work, but, if possible, you may want to switch the ordering here to assume the new format and fall back to the old one, that way everyone upgrades once and then stops wasting cycles on the old format, rather than doing wasted work in perpetuity.

src/wallet/wallet2.cpp
```diff
@@ -4329,7 +4329,17 @@ bool wallet2::load_keys_buf(const std::string& keys_buf, const epee::wipeable_st
...
+      load_info.is_legacy_key_encryption = true;
```
iamamyth, line 4333: This might be a little clearer as an `else` for the below `if` statement.

tests/unit_tests/serialization.cpp
```diff
@@ -616,6 +616,50 @@ TEST(Serialization, serializes_ringct_types)
...
+    const crypto::secret_key view_secret_key = w.get_account().get_keys().m_view_secret_key;
+
+    w.rewrite(wallet_file.string(), password); // transition to new key format
+
+    w.load(wallet_file.string(), password); // new key method
+    ASSERT_FALSE(w.get_load_info().is_legacy_key_encryption);
+    ASSERT_EQ(w.get_account().get_keys().m_view_secret_key,view_secret_key);
+
+    // Restore the old files
+    MINFO("Restoring wallet file...");
```
iamamyth, line 647: Why not just use `rename`?

tests/unit_tests/serialization.cpp
```diff
@@ -616,6 +616,50 @@ TEST(Serialization, serializes_ringct_types)
...
+  const boost::filesystem::path wallet_file = unit_test::data_dir / "wallet_9svHk1";
+  const boost::filesystem::path key_file = unit_test::data_dir / "wallet_9svHk1.keys";
+  const boost::filesystem::path new_wallet_file = unit_test::data_dir / "wallet_9svHk1_new";
+  const boost::filesystem::path new_key_file = unit_test::data_dir / "wallet_9svHk1_new.keys";
+  string password = "test";
+  bool r = false;
+  try
+  {
+    // Copy the old files before we overwrite them for this test
+    boost::filesystem::copy(wallet_file,new_wallet_file);
```
iamamyth, line 632: The naming here confused me: As best as I can tell `_new` is a clone of the original data, so it's not really new so much as a backup.

src/wallet/wallet2.cpp
SarangNoether: This order is swapped in a pending commit.

tests/unit_tests/serialization.cpp
SarangNoether: For some reason I thought there was could be issues using `rename` that depended on source and destination file sizes if the destination already existed. But apparently this isn't the case.

tests/unit_tests/serialization.cpp
SarangNoether: Fair enough, and easy to change.

moneromooo-monero:
Could you please add a sensible commit message ? "Update X" is not very useful. Just "use chacha the canonical way for in-memory encryption" or whatever else is appropriate.

SarangNoether:
> Could you please add a sensible commit message ? "Update X" is not very useful. Just "use chacha the canonical way for in-memory encryption" or whatever else is appropriate.

Done.

src/cryptonote_basic/account.cpp
```diff
@@ -99,33 +113,39 @@ DISABLE_VS_WARNINGS(4244 4345)
...
+    {
+      // Spend key
+      crypto::secret_key temp_key;
+      chacha20((char *) &m_spend_secret_key, sizeof(crypto::secret_key), derived_key, m_encryption_iv, (char *) &temp_key);
+      memcpy(&m_spend_secret_key, &temp_key, sizeof(crypto::secret_key));
+      memwipe(&temp_key, sizeof(crypto::secret_key));
+
+      // Multisig keys
+      std::vector<crypto::secret_key> temp_keys;
+      temp_keys.reserve(m_multisig_keys.size());
```
moneromooo-monero, line 135: That one seems pointless

tests/unit_tests/serialization.cpp
```diff
@@ -616,6 +616,48 @@ TEST(Serialization, serializes_ringct_types)
...
+  {}
+
+  // Restore the original files
+  MINFO("Restoring wallet file...");
+  boost::filesystem::rename(backup_wallet_file,wallet_file);
+  MINFO("Restoring key file...");
+  boost::filesystem::rename(backup_key_file,key_file);
+
+  ASSERT_TRUE(r);
+}
```
moneromooo-monero, line 659: It would seem better to do the test on the scratch file, rather than the revision controlled one (ie, if the process SIGSEGVs for instance, you end up with a git change).

SarangNoether: Updated.

src/cryptonote_basic/account.cpp
SarangNoether: Updated.

moneromooo-monero:
Looks good, please squash now.

selsta:
This breaks in wallet password entry (refresh / lock) for me on all wallets.

SarangNoether:
Recommend reverting this, as the tests are likely missing functionality.

SarangNoether:
After discussion, a safer alternative would be to apply this method only to new wallets. The existing method should not be dangerous in practice, and applying the updated method only to new wallets minimizes the risk of data loss.

PR #6495, opened 2020-05-02
workflows: add ccache support

selsta:
~I’m still working on Windows.~

Done. PR is ready now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,14 +5,24 @@ on: [push, pull_request]
 jobs:
   build-macos:
     runs-on: macOS-latest
+    env:
+      CCACHE_COMPRESS: 1
+      CCACHE_TEMPDIR: /tmp/.ccache-temp
     steps:
     - uses: actions/checkout@v1
       with:
         submodules: recursive
+    - uses: actions/cache@v2
+      with:
+        path: ~/.ccache
+        key: ccache-macos-build-${{ github.sha }}
+        restore-keys: ccache-macos-build-
     - name: install dependencies
-      run: HOMEBREW_NO_AUTO_UPDATE=1 brew install boost hidapi zmq libpgm miniupnpc ldns expat libunwind-headers protobuf
+      run: HOMEBREW_NO_AUTO_UPDATE=1 brew install boost hidapi zmq libpgm miniupnpc ldns expat libunwind-headers protobuf ccache
     - name: build
-      run: make -j3
+      run: |
+        ccache --max-size=150M
+        make -j3
 
   build-windows:
     runs-on: windows-latest
@@ -32,10 +42,18 @@ jobs:
 
   build-ubuntu:
     runs-on: ubuntu-latest
+    env:
+      CCACHE_COMPRESS: 1
+      CCACHE_TEMPDIR: /tmp/.ccache-temp
     steps:
     - uses: actions/checkout@v1
       with:
         submodules: recursive
+    - uses: actions/cache@v2
+      with:
+        path: ~/.ccache
+        key: ccache-ubuntu-build-${{ github.sha }}
+        restore-keys: ccache-ubuntu-build-
     - name: remove bundled boost
       run: sudo rm -rf /usr/local/share/boost
     - name: set apt conf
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -5,14 +5,25 @@ on: [push, pull_request]
...
+        path: ~/.ccache
+        key: ccache-macos-build-${{ github.sha }}
+        restore-keys: ccache-macos-build-
     - name: update brew and install dependencies
-      run: brew update && brew install boost hidapi zmq libpgm miniupnpc ldns expat libunwind-headers protobuf
+      run: brew update && brew install boost hidapi zmq libpgm miniupnpc ldns expat libunwind-headers protobuf ccache
     - name: build
-      run: make -j3
+      run: |
+        ccache --max-size=100M
```
mj-xmr, line 25: My humble opinion is, that 100 megabytes for Monero is at least 20 times too little. The cached files will get replaced sooner than they can ever be reused.
My estimated amount comes from the size of the build directory - about 2GB.

selsta: Travis also uses 100MB cache and it seems to work quite well. I can increase it a bit but Github Actions has a 5GB/repo limit.

Would you share one cache between operating systems? I didn’t do this but would this be possible?

mj-xmr: Actually I've just realized, that my logic was flawed. The files, which concern ccache are object files ONLY. The large size of the build directory is the result of having so many targets linked statically, which is out of scope for ccache. To measure the real size of files getting into the cache, you may run the command:

`find . -type f -name '*.o' -exec du -ch {} + | grep total`

which yields:
`87M	total`

So 20 times - definitely not, but say 3 times would still be well below Github limits. Are there limits for processing power as well? If so, then we have to find some middle ground anyway.

As to sharing the cache between OSes - I wouldn't do it, since we're getting into a risky territory.

selsta:
Updated. Had to remove ccache support for `test-ubuntu` CI because tests would fail with `Exception: Illegal`.

Not sure if the added uncertainty is worth the time gains or if I should close this PR.

.github/workflows/build.yml
```diff
@@ -5,35 +5,67 @@ on: [push, pull_request]
...
+      CCACHE_COMPRESS: 1
+      CCACHE_TEMPDIR: /tmp/.ccache-temp
     steps:
     - uses: actions/checkout@v1
       with:
         submodules: recursive
+    - name: ccache
+      uses: actions/cache@v1
+      with:
+        path: ~/.ccache
```
mj-xmr, line 18: Please ELI5:
does the same path for all the *nix instances mean, that the exact files are shared? Or are these actually different files, whose path is just named the same way? I'm trying to settle whether you share the cached files across OSes, which shouldn't be done. Perhaps if this is the case, changing the paths to use individual ones would also be the cure?

mj-xmr:
> Updated. Had to remove ccache support for `test-ubuntu` CI because tests would fail with `Exception: Illegal`.
> 
> Not sure if the added uncertainty is worth the time gains or if I should close this PR.

Being the ccache "pusher", I'm not an independent voice, but I think, that any effort of reducing the turnaround time is worth it.

.github/workflows/build.yml
selsta: > does the same path for all the *nix instances mean, that the exact files are shared?

These are all separate VMs. No files are being shared between different OS or different jobs. Each of these jobs (build-macos, build-windows, build-ubuntu, libwallet-ubuntu) share one cache.

See: `restore-keys: ccache-windows-build-` for the Windows build job.

selsta:
I would be okay with it, if we run into issues we can remove it later anyway.

mj-xmr:
I think, that even if something is not working, it now clearly shows what the bottleneck of the CI is: testing and building for testing.
It would be wise to figure out in the next iteration what the exact cause is, just like I've done with the FreeBSD setup on Travis. It could uncover a hidden bug.

Once this is merged (or just by branching off this branch), an interesting experiment would be to make an intentional compilation error and push the changes, to see if ccache also crashes on CI. I'd like to eliminate the risk, that we're using differenent cached binaries, than we think we are.

Snipa22:
The branch has conflicts and is not mergeable.  @selsta can we get an update?

selsta:
Unsure if I want this to go in :/ I like ccache locally but not sure if we should add it to CI.

moneromooo-monero:
Does the CI bot keep the cache from one run to the next ?

selsta:
It would keep the cache, yes. My worry is that the CI would have false result due to some bug in ccache but not sure if that can happen.

xmrdsc:
If state persists between different CI builds... it might have security implications.

selsta:
Can you elaborate?

jonathancross:
Friendly ping @xmrdsc

mj-xmr:
I might help here out here soon, once I resolve my hi priority tasks. At least by trying to trigger and identify the root cause of the problem.

selsta:
Rebased. Should be merge ready now.

xiphon (approved):

xiphon (approved):

.github/workflows/build.yml
```diff
@@ -5,14 +5,24 @@ on: [push, pull_request]
...
+      CCACHE_COMPRESS: 1
+      CCACHE_TEMPDIR: /tmp/.ccache-temp
     steps:
     - uses: actions/checkout@v1
       with:
         submodules: recursive
+    - uses: actions/cache@v2
+      with:
+        path: ~/.ccache
+        key: ccache-macos-build-${{ github.sha }}
```
mj-xmr, line 18: Was the ${{ github.sha }} suffix missing previously?

selsta: I don't think so.

mj-xmr (approved):

mj-xmr:
Could you please run:

```
git commit -m "retrigger checks" --allow-empty
git push
```
in order to see if the build time decreases after reusing the cache and the tests still succeed?

selsta:
@mj-xmr we will see it anyway if there are issues :)

mj-xmr:
> @mj-xmr we will see it anyway if there are issues :)

Hardcore :)

PR #6500, opened 2020-05-03
Update expat.mk

ArqTras:
Sorry :)

### DIFF
--- a/contrib/depends/packages/expat.mk
+++ b/contrib/depends/packages/expat.mk
@@ -1,6 +1,6 @@
 package=expat
 $(package)_version=2.2.4
-$(package)_download_path=https://downloads.sourceforge.net/project/expat/expat/$($(package)_version)
+$(package)_download_path=https://github.com/libexpat/libexpat/releases/download/R_2_2_4
 $(package)_file_name=$(package)-$($(package)_version).tar.bz2
 $(package)_sha256_hash=03ad85db965f8ab2d27328abcf0bc5571af6ec0a414874b2066ee3fdd372019e
 


ArqTras:
Need to swich back to linux :D sorry for mes with commits. Now should be ok.
@selsta

contrib/depends/packages/expat.mk
```diff
@@ -1,6 +1,6 @@
 package=expat
 $(package)_version=2.2.4
-$(package)_download_path=https://downloads.sourceforge.net/project/expat/expat/$($(package)_version)
+$(package)_download_path=https://github.com/libexpat/libexpat/releases/download/R_2_2_4
```
hyc, line 3: This is really kind of ugly, since it requires manually updating the version twice. (line 2 and line 3). You can't
find a way to specify the version just once, that can be used in both places?

sumogr: you may use this link to comply with @hyc 's suggestion 
https://toolchains.bootlin.com/downloads/releases/sources/expat-2.2.4/expat-2.2.4.tar.bz2
`...`
`$(package)_download_path=https://toolchains.bootlin.com/downloads/releases/sources/expat-$($(package)_version)/`
`$(package)_file_name=$(package)-$($(package)_version).tar.bz2`
`...`

selsta: In this case I would not change the link at all, the sourceforge one seems to still work.

ArqTras: There was issue with access to sourceforge, that's why I suggested change it. Anyway it not happen very often so it can stay as is

iamamyth: It should be trivial to keep the canonical repo (which was previously sourceforge, but is now or will soon be github, per note on the github repo) while also having a single version definition, e.g.:
```
$(package)_download_path=https://github.com/libexpat/libexpat/releases/download/R_$(subst .,_,$($(package)_version))
```

I think the change of URL is warranted because sourceforge appears to no longer be the canonical code location, so it may eventually become impossible to obtain updated versions (or any version) from sourceforge.

PR #6501, opened 2020-05-04
Made ccache optional (opt out) and tidied up the FindCcache.cmake

mj-xmr:
Wrapped the FindCcache with an opt out option as requested.
Also tidied up the FindCcache itself with some comments.


mj-xmr:
@vtnerd: Isn't this needed anymore?

selsta:
We just didn’t do merges yet.

Issue #6496, opened 2020-05-03
cli wallet strange little bug


sumogr:
using latest release (15.0.5): on entering the wrong password upon opening an existing wallet  
`Error: failed to load wallet: invalid password` good

using compiled bins from latest master: on entering the wrong password upon opening an existing wallet 
` Error: failed to load wallet: std::exception`
it doesnt grab the e.what(). 
why?

sumogr:
nevermind its due to the wallet2.h for wasm restructuring i found what caused it

selsta:
Are you going to open a PR to fix it?

sumogr:
> Are you going to open a PR to fix it?

ok

PR #6502, opened 2020-05-04
wallet2: fix catching reason on fail to verify password on wallet opening

sumogr:
on wrong password typed on opening an existing wallet
`Error: failed to load wallet: std::exception` was printed on exit
instead of
`Error: failed to load wallet: invalid password`
that's a small bug introduced by vtnerd's load keys restructuring part to make wallet2.h work with web assembly (the original exception with the invalid password reason moved from `wallet2::load_keys` block to `wallet2::load_keys_buf`) .
This is just a patch that fixes it, i bet there are more elegant ways of fixing this but i dont want to mess with the structure,

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -3975,7 +3975,12 @@ bool wallet2::load_keys(const std::string& keys_file_name, const epee::wipeable_
   std::string keys_file_buf;
   bool r = load_from_file(keys_file_name, keys_file_buf);
   THROW_WALLET_EXCEPTION_IF(!r, error::file_read_error, keys_file_name);
-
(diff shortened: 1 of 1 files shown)


moneromooo-monero:
This seems to read the keys file twice, no ?

sumogr:
the two bool blocks? i think so yes but i didnt know if there is a purpose behind it thats why i didnt mess with it further

moneromooo-monero:
I have no idea what you're saying here.
But clearly you changed it. And it's now loading twice. Why.

selsta:
He tried to fix #6496

sumogr:
I dont know what you mean either. what did i change? the password verification? anyhow. ok

PR #6503, opened 2020-05-04
cryptonote_protocol: do not request pruned borromean sig txes

moneromooo-monero:
We don't have a function to calculate their weight from a pruned
version (yet).


src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1907,6 +1907,10 @@ skip:
     const uint32_t local_stripe = tools::get_pruning_stripe(m_core.get_blockchain_pruning_seed());
     if (local_stripe == 0)
       return false;
+    // don't request pre-rct pruned blocks, we can't reconstruct their weight (yet)
```
luigi1111, line 1910: Should this say pre-bp?

moneromooo-monero: fixed

PR #6507, opened 2020-05-05
cryptonote_core.cpp : update hashrate warning

sumogr:
Ok it might be "bad luck" when next block delays (bad luck for the miners i guess...) but for whom is it bad luck when more than the statistically expected blocks pop up during a preset period of time?
I think "coincidence" is the right word

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1908,7 +1908,7 @@ namespace cryptonote
       MDEBUG("blocks in the last " << seconds[n] / 60 << " minutes: " << b << " (probability " << p << ")");
       if (p < threshold)
       {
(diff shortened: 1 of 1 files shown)


moneromooo-monero:
It's bad luck because it's trying to hit an average, and it's clearly failing at this moment in time.

sumogr:
its seems to be failing at that moment but will compensate for sure afterwards to hit the difficulty window (unless indeed something is off). IMHO I dont think it can be described as a failure cause it is a legit possibility

sumogr:
> It's bad luck because it's trying to hit an average, and it's clearly failing at this moment in time.

i understand why "bad luck" feels ok to a developer and why it was used but in IMHO there is no such thing as bad luck in algorithms, there are just edge case (but statistically probable) outcomes that may occur from time to time for no reason whatsoever. Bad luck is a reason :) It just feels that "coincidence" is the proper word for it and bad luck is a colloquialism. Too much conversation for sth trivial :) i ll close this if you want to leave it unchanged

PR #6509, opened 2020-05-06
systemd: Make sure required folders exist

Forage:
The default monero.conf file depends on the existence of the folders `/var/log/monero/` and `/var/lib/monero/`.
This change makes sure systemd will create them, together with the proper permissions, if they don't exist.
`StateDirectory` can be considered an extra safety in case the user `monero` happens to have been created with specifying `/var/lib/monero/` as a home folder but without actually creating it.

### DIFF
--- a/utils/systemd/monerod.service
+++ b/utils/systemd/monerod.service
@@ -7,6 +7,8 @@ User=monero
 Group=monero
 WorkingDirectory=~
 RuntimeDirectory=monero
+StateDirectory=monero
+LogsDirectory=monero
 
 # Clearnet config
 #


Forage:
Sorry, I messed up the first PR.

To clarify:
- If `/var/lib/monero/` doesn't exist the service will fail to launch
- If `/var/log/monero/` does not exist nothing will be logged

Neither of them will be automatically created by monerod itself because the monero user won't have sufficient permissions to do so.

It also appears to me that `WorkingDirectory=~` is redundant, monerod will launch and work fine without.
The only time it will actually cause problems is when the user folder does not exist. E.g. with `/var/lib/monero/` without the proposed change or the default `/home/monero/` when creating a new user.
To reduce risks it could be removed if you ask me, but there might be a good reason I don't know about to keep it.

Snipa22:
This just ensures that the Monero directory is used for storing logs and state information, good for general management/cleanup.

PR #6516, opened 2020-05-11
remove double includes

sumogr:
there are quite a few as well that are implicitly included via another header while included directly as well but in some other pr

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.h
+++ b/contrib/epee/include/net/abstract_tcp_server2.h
@@ -48,12 +48,9 @@
 #include <boost/asio.hpp>
 #include <boost/asio/ssl.hpp>
 #include <boost/array.hpp>
-#include <boost/noncopyable.hpp>
-#include <boost/shared_ptr.hpp> //! \TODO Convert to std::shared_ptr
 #include <boost/enable_shared_from_this.hpp>
 #include <boost/interprocess/detail/atomic.hpp>
 #include <boost/thread/thread.hpp>
-#include <memory>
 #include "byte_slice.h"
 #include "net_utils_base.h"
 #include "syncobj.h"
--- a/contrib/epee/include/net/network_throttle.hpp
+++ b/contrib/epee/include/net/network_throttle.hpp
@@ -44,10 +44,7 @@
 #include <boost/shared_ptr.hpp>
 #include <atomic>
 
-#include <boost/asio.hpp>
 #include <boost/array.hpp>
-#include <boost/noncopyable.hpp>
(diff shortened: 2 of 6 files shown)


hyc:
Aside from general tidying, is there any actual benefit to this patch? Those header files are already guarded against double-inclusion so they don't get processed multiple times.

moneromooo-monero:
Besides, including what you use explicitely is more robust against your include being lost implicitely if an intervening header changes. I don't see the point in this unless you have an argument for it I'm not thinking about (admittedly my point is minor).

sumogr:
1. it's tidying indeed no harm is coming from double explicit includes but this is a pro project afaik
2. indeed mooo its more robust if you do it on purpose for all of your includes. these explicit-implict double just happened

iamamyth:
I don't understand what you mean by explicit-implicit doubles, but the first change for example removed `boost::noncopyable` include which is directly referenced by the including file.

sumogr:
> I found your description a little confusing, but it looks like some of these includes are repeated directly in the including file. Still I find some of the choices here strange, like picking:
> 
> ```
> #include <boost/noncopyable.hpp>
> #include <boost/shared_ptr.hpp>
> #include <atomic>
> ```
> 
> in abstract_tcp_server2.h versus keeping the later include grouped with the other boost includes.

Yes, the ones i deleted were written directly twice.
Indeed the proper way is stl includes first, boost later, local header files in the end however to be safe I deleted the second copy from each double occurence cause sometimes some headers depend on other headers and i was too lazy checking them.

mj-xmr:
> Indeed the proper way is stl includes first, boost later, local header files in the end

I disagree. Convention is one thing, but local first, then boost/3rd party, then stdlib gives the advantage of protecting your local classes from being distributed without naming their dependencies:
https://stackoverflow.com/questions/2762568/c-c-include-header-file-order

sumogr:
> 
> 
> > Indeed the proper way is stl includes first, boost later, local header files in the end
> 
> I disagree. Convention is one thing, but local first, then boost/3rd party, then stdlib gives the advantage of protecting your local classes from being distributed without naming their dependencies:
> https://stackoverflow.com/questions/2762568/c-c-include-header-file-order
> specifically:
> https://stackoverflow.com/a/14243688

Ok noted. includes order comes down to a matter of opinion indeed some prefer my suggestion cause stl cannot be broken by something and this is why it comes first and local headers might require something from stl or boost which is why they should come last and so on. as long as it compiles it is  "you like tomato and I like tomahto". However that was just a comment it doesnt have anything to do with the purpose of this pr which just deletes double typed includes on the same file

PR #6514, opened 2020-05-09
[master] build: prepare v0.16.0.0 release

selsta:
will close whichever PR we don’t need

PR #6518, opened 2020-05-12
[release-v0.16]: Update gitian yml files

iDunk5400:


moneromooo-monero:
> I also updated #6514 to include these changes.

Would you mind not doing so ? It is really annoying when what purports to be the same patch on two different branches are actually not the same ones, and makes it hard to spot stuff that's been forgot on one branch.

selsta:
Would you prefer if I revert it or for the future?

moneromooo-monero:
If it's not too much hassle, I'd rather have the same patches on both sides.

selsta:
Updated but wasn’t able to get the same commit hash. Not sure if that is possible.

moneromooo-monero:
That's expected, the commit hash also depends on the commit's ancestry.

PR #6525, opened 2020-05-13
simplewallet: don't complain about incoming payment ids on change

moneromooo-monero:


src/simplewallet/simplewallet.h
```diff
@@ -342,7 +342,7 @@ namespace cryptonote
 
     //----------------- i_wallet2_callback ---------------------
     virtual void on_new_block(uint64_t height, const cryptonote::block& block);
-    virtual void on_money_received(uint64_t height, const crypto::hash &txid, const cryptonote::transaction& tx, uint64_t amount, const cryptonote::subaddress_index& subaddr_index, uint64_t unlock_time);
+    virtual void on_money_received(uint64_t height, const crypto::hash &txid, const cryptonote::transaction& tx, uint64_t amount, const cryptonote::subaddress_index& subaddr_index, bool change, uint64_t unlock_time);
```
selsta, line 345: Minor comment:

```suggestion
    virtual void on_money_received(uint64_t height, const crypto::hash &txid, const cryptonote::transaction& tx, uint64_t amount, const cryptonote::subaddress_index& subaddr_index, bool is_change, uint64_t unlock_time);
```

The same in the other header file.

moneromooo-monero: Or maybe "bool is_change_but_false_might_still_be_change_if_out_of_date_multisig", I wonder.

selsta: You called it `bool is_change` in the source file implementation.

moneromooo-monero: Ah. OK, I thought it was just pointless bikeshedding :P

PR #6531, opened 2020-05-14
[master] build: fix boost 1.73 compatibility

selsta:


jtgrassie:
This breaks compilation on Ubuntu 16 LTS with boost 1.58

jtgrassie:
@vtnerd's suggestion is preferable as it works with both versions of boost.

luigi1111:
I welcome a PR that fixes boost 1.58, but merging this for now.

PR #6534, opened 2020-05-15
daemon: guard against rare 'difficulty drift' bug with checkpoints and recalculation

stoffu:
On startup, it checks against the difficulty checkpoints, and if any mismatch is found, recalculates all the blocks with wrong difficulties. Additionally, once a week it recalculates difficulties of blocks after the last difficulty checkpoint.


stoffu:
To give a context, this 'difficulty drift' bug has been known to exist for quite a long time:

- http://45.154.255.46:18089/getinfo
- http://165.227.4.64:18089/getinfo
- http://165.227.96.96:18089/getinfo
- http://162.210.173.15:18089/getinfo
- http://173.255.205.142:18081/getinfo

Notice how the difficulties are calculated wrong compared to the correct one at http://nodes.hashvault.pro:18081/getinfo

@moneromooo-monero mentioned that this may have been caused by a database hiccup. The bug is practically not reproducible and it seems very hard to get rid of it, so I propose to circumvent it by relying on hardcoded checkpoints and occasional recalculation (which shouldn't take long).

src/checkpoints/checkpoints.cpp
```diff
@@ -172,45 +194,45 @@ namespace cryptonote
...
+    ADD_CHECKPOINT2(1200000, "fa7d13a90850882060479d100141ff84286599ae39c3277c8ea784393f882d1f", "0x4c73503fc4aa3");
+    ADD_CHECKPOINT2(1300000, "31b34272343a44a9f4ac7de7a8fcf3b7d8a3124d7d6870affd510d2f37e74cd0", "0x723f49bc249d5");
+    ADD_CHECKPOINT2(1390000, "a8f5649dd4ded60eedab475f2bec8c934681c07e3cf640e9be0617554f13ff6c", "0xb4bba65e2841b");
+    ADD_CHECKPOINT2(1450000, "ac94e8860093bc7c83e4e91215cba1d663421ecf4067a0ae609c3a8b52bcfac2", "0x11a4aabdca9511");
+    ADD_CHECKPOINT2(1530000, "01759bce497ec38e63c78b1038892169203bb78f87e488172f6b854fcd63ba7e", "0x2819ce9f9e91e5");
+    ADD_CHECKPOINT2(1579000, "7d0d7a2346373afd41ed1e744a939fc5d474a7dbaa257be5c6fff4009e789241", "0x357a590e7dda83");
+    ADD_CHECKPOINT2(1668900, "ac2dcaf3d2f58ffcf8391639f0f1ebafcb8eac43c49479c7c37f611868d07568", "0x474226e475cc3b");
+    ADD_CHECKPOINT2(1775600, "1c6e01c661dc22cab939e79ec6a5272190624ce8356d2f7b958e4f9a57fdb05e", "0x5e3b9d206a27c6");
+    ADD_CHECKPOINT2(1856000, "9b57f17f29c71a3acd8a7904b93c41fa6eb8d2b7c73936ce4f1702d14880ba29", "0x6bde5e1caccee1");
+    ADD_CHECKPOINT2(1958000, "98a5d6e51afdf3146e0eefb10a66e8648d8d4d5c2742be8835e976ba217c9bb2", "0x79dd46d2a0971a");
```
selsta, line 235: Where would I find the difficulty when updating the checkpoints in the future?

stoffu: Run

    curl -X POST http://127.0.0.1:18081/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"get_block_header_by_height","params":{"height":1958000}}' -H 'Content-Type: application/json'

and check the `wide_cumulative_difficulty` field.

selsta:
Also is this in your opinion important enough to include in upcoming v0.16 release?

stoffu:
@selsta 
Pushed some fixes to make the tests pass.

I think the bug is serious and annoying, but on the other hand the network has been fine for a long time without fixing it, mostly because only a few nodes have been affected, and the differences in drifted difficulties are generally small.

So perhaps this patch can wait for a while in the master branch before going to release so that any possible flaw can be found by people running the master.

src/checkpoints/checkpoints.cpp
moneromooo-monero: print_block HEIGHT

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -2750,6 +2750,34 @@ difficulty_type BlockchainLMDB::get_block_difficulty(const uint64_t& height) con
...
 }
 
+void BlockchainLMDB::overwrite_block_cumulative_difficulty(const uint64_t& height, const difficulty_type& new_cumulative_difficulty)
+{
+  LOG_PRINT_L3("BlockchainLMDB::" << __func__);
+  check_open();
+  mdb_txn_cursors *m_cursors = &m_wcursors;
+
+  int result = 0;
+  block_wtxn_start();
```
moneromooo-monero, line 2760: Does the existing TXN_BLOCK_PREFIX macro not work here ?

src/cryptonote_core/blockchain.cpp
```diff
@@ -888,6 +897,94 @@ difficulty_type Blockchain::get_difficulty_for_next_block()
   return diff;
 }
 //------------------------------------------------------------------
+std::pair<bool, uint64_t> Blockchain::check_difficulty_checkpoints()
```
moneromooo-monero, line 900: const ?

fuwa0529:
What issue does this fix exactly?

sumogr:
> What issue does this fix exactly?

there's a slight deviation in difficulty and of course consequently in cum_diff between nodes. 

From here on is my humble opinion and only that: Monero's diff is quite big, its hashrate vastly decentralised and the deviation is small which is why i guess it didnt/wont create "unexpected" issues in the future. Most likely its a db thing but my feeling is it's a precision thing and the best way to reset is to hardcode difficulty on next hardfork block height and bring all nodes from the same starting point cause this commit might cause sync issues if the recalculation gives every time same different result among nodes which dont match the checkpoints'. I might be waaaay off though here :)

selsta:
> What issue does this fix exactly?

Loki had a chain split due to difficulty drift bug: https://github.com/loki-project/loki-core/releases/tag/v7.1.9

sumogr:
> > What issue does this fix exactly?
> 
> Loki had a chain split due to difficulty drift bug: https://github.com/loki-project/loki-core/releases/tag/v7.1.9

Loki's hashrate, as all other monero forks, have a more centralised hashrate and if a big pool has a big percentage of the network's hashrate then with a slight difference in difficulty it can drag along the nodes and fork the network.
 @selsta check here there, is a pool (herominers) that owns 70% of loki's hashrate https://miningpoolstats.stream/loki
If monero had such a centralised hashrate this bug would have caused the same issue but monero doesnt

selsta:
> If monero had such a centralised hashrate this bug would have caused the same issue but monero doesnt

It’s still a good idea to address this.

sumogr:
> > If monero had such a centralised hashrate this bug would have caused the same issue but monero doesnt
> 
> It’s still a good idea to address this.

Absolutely

fuwa0529:
Thanks @sumogr and @selsta

fuwa0529:
Wait, but this doesn't fix that diff drift bug, right?

So it's still not clear which issue this PR is fixing.

selsta:
> Wait, but this doesn't fix that diff drift bug, right?

Like it says in the title it guards against it which might be the only option if the bug is rarely reproducible.

iamamyth:
But if you still aren't sure what causes the bug then how can one be certain this change guards against it? What if someone introduces a checkpoint with a buggy computed difficulty? How do you know the checkpoint difficulties are the right ones? It seems that unless you can be certain the underlying computation works correctly, and the problem lies elsewhere, it would be premature to consider such a change.

iamamyth:
Also, to be clear, I'm not necessarily objecting to this change, or some variant thereof, I just think the hypothesis and expectations need to be clearly defined. For example, if the hypothesis is, "I think there was an earlier bug which has subsequently been fixed, and this change will undo local node drift and alert if it recurs," then that's a decent justification. But, if the hypothesis is, "I think there was an earlier bug, I have no idea if it was fixed, I want to devise an experiment to better understand the situation," then it may be more reasonable do something very different, like check for drifting data and *not* overwrite it so the state can first be exported and more clearly examined. And there are of course hybrids of those two strategies, like export then update. It's hard to determine the best approach without a clear understanding of the underlying assumptions and goals.

UkoeHB:
Has any more detailed information about this bug been collected? For example, how often does it crop up? Can the specific blocks where divergence appears be identified (maybe they all have cousins who got orphaned, so the problem is related to chain forks e.g. maybe difficulties aren't properly updated when orphaning a chain)? Does divergence occur when computing difficulties for a fresh chain, or just when actively collecting new blocks?

moneromooo-monero:
It's been here for quite a while. It's very rare. There are no obvious reasons (ie, reorgs). I had a patch that logged everything of interest that got run by some people who had the bug happen more than once, but it never saw anything. It's never happened to me AFAIK. Basically, since we can't find the bug, stoffu's patch cleans up after the fact, which is better than what we have now.

UkoeHB:
Ah too bad. Anyway, the recommended patch seems reasonable. A bit of redundancy here and there is nice.

hyc:
Since you currently have identified a set of nodes that have disagreeing difficulty, can't we export their blockchains and compare them to each other to see where the drift first occurred?

moneromooo-monero:
If you know who these people are, you could. When I got such data though, it did not help me find the reason at the time.

stoffu:
@hyc @moneromooo-monero 
It's easy to get all the block headers from these nodes suffering from the diff drift bug and identify the heights when the drift started using RPC `get_block_headers_range` without contacting the node operators. The problem is that it seems impossible to find a way to reproduce the wrong difficulty values from the same data (arrays of previous blocks' timestamps and cumulative difficulties). Assuming the bug was caused by a database hiccup, some values in the arrays may have been altered in some ways, but there are infinitely many ways to do so, and I don't see any hope of identifying the cause of the bug in this manner.

hyc:
Any chance all of these drifting nodes are on Windows? or have some other peculiarity in common?

stoffu:
@iamamyth 
> But if you still aren't sure what causes the bug then how can one be certain this change guards against it? What if someone introduces a checkpoint with a buggy computed difficulty? How do you know the checkpoint difficulties are the right ones? It seems that unless you can be certain the underlying computation works correctly, and the problem lies elsewhere, it would be premature to consider such a change.

We can be certain that the underlying computation works correctly. After all, that's what is being confirmed in the difficulty tests. The difficulty values are solely determined by the sequence of timestamps, and each block stores its timestamp, so the difficulty at any height should be the same for everyone as long as the block hash is the same. So we can be sure that the checkpointed difficulties are correct because that's what the majority healthy nodes report and what you get when you recalculate them.

iamamyth:
I would say your definition of certainty and mine are quite different. As best as I can tell, the difficulty tests don't ever even exceed 64 bits in cumulative difficulty (though, admittedly, the blockchain doesn't, either). Still, the tests could stand improvement, and who knows if the underlying boost implementation works as it should across platforms. "The database is mangling things" seems a bit problematic as an explanation: If it's true, why just difficulty values? Why not anything else? I wouldn't be so quick to write off a problem in the underlying difficulty calculation, even if we can agree the values in the checkpoints are correct.

moneromooo-monero:
> the difficulty tests don't ever even exceed 64 bits in cumulative difficulty

I thought I made them do so -_- Patches welcome :)

FWIW, I think the most likely is some race condition where we're reading stale data from the db somehow.

stoffu:
@iamamyth 
> As best as I can tell, the difficulty tests don't ever even exceed 64 bits in cumulative difficulty

Of course wide difficulties are tested, look more carefully at tests/difficulty.

> "The database is mangling things" seems a bit problematic as an explanation: If it's true, why just difficulty values? Why not anything else? I wouldn't be so quick to write off a problem in the underlying difficulty calculation, even if we can agree the values in the checkpoints are correct.

Of course it'd be best if we could figure out the bug exactly and solve it beautifully. That being extremely hard, the point of this patch is to apply a simple bandaid for now to the problem which could potentially cause a network split. Having this patch doesn't prevent anyone from figuring out the bug and fixing it. Actually this patch can help debugging since it checks and reports any possible difficulty drift when it occurs; if there are enough people reporting the error, perhaps we might get a clue.

src/blockchain_db/lmdb/db_lmdb.cpp
stoffu: I'm not very familiar with LMDB operations, but it didn't work when it replaced it with `TXN_BLOCK_PREFIX(0)` or `TXN_PREFIX(0)`.

src/cryptonote_core/blockchain.cpp
stoffu: ok

iamamyth:
@moneromoo-monero My bad, the wide generator does exercise > 64-bit values, I just misread the harness invocation.

@stoffu I wasn't really objecting to the patch, you would inevitably need something similar to repair any drift, and, as I said, I think the values in the checkpoints are correct. I simply worry that papering over the problem by auto-correcting it at startup risks carrying forward the bug indefinitely, and, as no one actually knows the cause, there could still be a netsplit down the line when the network is potentially bigger and harder to reconcile.

iamamyth:
@moneromooo-monero One last question: Did this happen prior to the introduction of wide difficulty, or only after?

stoffu:
@iamamyth 
> I simply worry that papering over the problem by auto-correcting it at startup risks carrying forward the bug indefinitely,

It's not papering over since it reports a red error message when the drift occurs (as I said above). I could make the error message even more sensational to urge people to report it.

> and, as no one actually knows the cause, there could still be a netsplit down the line when the network is potentially bigger and harder to reconcile.

It's indeed scary that the database can work in an unexpected manner in very rare cases; in that sense our network can be considered as fragile rather than rock solid. However, this patch does guard the network against this type of difficulty drift that's been observed for a long time, so I don't see any negative side to it. I don't understand what you're criticizing.

shopglobal:
Fwiw I believe this is an issue at the pool level causing difficulty to
become adjusted incorrectly, why the node accepts blocks from the incorrect
difficulty is beyond me at this point in time. Trying to fix it at the
difficulty level will just eliminate cheating pools, which will cause them
to run the correct pool soft.. Of that's the case then this PR ought to be
on track because if we somehow allow incorrect difficulty under
certain conditions then it's the cores duty to harden

shopglobal:
If*

iamamyth:
> I don't understand what you're criticizing.

As I've said twice in this thread, I do not object to this change. As for my original motivation, I thought it was spelled out in my earlier comment: "I'm not necessarily objecting to this change, or some variant thereof, I just think the hypothesis and expectations need to be clearly defined." Now, it seems, there is a problem statement on offer (incorrect difficulties are somehow being stored as of the latest release, but the precise mechanism is unknown and pre-dates wide difficulties), and the expectation is that perhaps the error logging will get people to submit more bug reports. If someone stumbles upon this problem at a later date, now they have a much more comprehensive understanding of what's happening, what's been attempted as a solution, and where to go from here, which seems helpful in an open source endeavor.

stoffu:
@shopglobal 
> Fwiw I believe this is an issue at the pool level causing difficulty to become adjusted incorrectly

This has nothing to do with pools. Pools are fed data by daemons they're connected to, and it's the daemon that is suffering from the bug.

> why the node accepts blocks from the incorrect difficulty is beyond me at this point in time

Pools with incorrect difficulties have been fine accepting the same blocks because the magnitude of the drift has been fairly small so far. If the difference grows significantly, and the wrongly calculated difficulty gets significantly smaller than the correct one, then the affected pool could produce a block that it thinks is valid but gets rejected by the majority network.

moneromooo-monero:
I've refreshed my old logging patch, and made it so it only logs once something wrong is detected. This way it can go in the coming release and only output if there's something interesting to report. We might get interesting data when it's run at scale. If you see something else that'd be nice to log, feel free to mention it.

https://github.com/monero-project/monero/pull/6544

stoffu:
Pushed a bugfix on handling of early blocks.

stoffu:
Pushed some progress reporting code as the rewriting process can take quite some time.

stoffu:
Turns out that calling `block_wtxn_start/stop` repetitively is very slow, so pushed some fix to make it a bulk process.

stoffu:
Looks like it needs a bit more testing.

stoffu:
Seems to be working.

stoffu:
Fixed yet another bug related to caching

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -2750,6 +2750,41 @@ difficulty_type BlockchainLMDB::get_block_difficulty(const uint64_t& height) con
   return diff1 - diff2;
 }
 
+void BlockchainLMDB::correct_block_cumulative_difficulties(const uint64_t& start_height, const std::vector<difficulty_type>& new_cumulative_difficulties)
+{
+  LOG_PRINT_L3("BlockchainLMDB::" << __func__);
+  check_open();
+  mdb_txn_cursors *m_cursors = &m_wcursors;
+
+  const uint64_t bc_height = height();
```
moneromooo-monero, line 2759: That probsbly wants to be after the txn start call below.

shopglobal:
Agree with you, definitely have a daemon issue here as it could be spotted
and corrected at the daemon. Although, I have witnessed a difficulty drift
occur at a peers pool where only a simple reboot of the pool would
re-engage the proper difficulty feed from his daemon. Must reboot pool
before mining any of the drifted blocks. He does that with a Bash bot.

 In his case somehow the daemon difficulty hadn't drifted before the pool
difficulty drifted... Revealing that it could be related to his pool soft.
Or some customizations he made to the pool soft, in an attempt to increase
his pools probability. Many pool ops have tricks up their sleeves. They
think they can change or remove a few pieces of code, and if the block is
still seen valid they don't care what they break if anything at all. They
are dumping every coin they can fill in their bags regardless. With enough
hash or enough time hashing those blocks won't get orphaned because the
rest of the network continues to recalculate next_difficulty based on their
local blocks... The drifted block that the rest of the network sees as
non-drifted is an issue.

How come next_difficulty doesn't recalculate properly for the rest of the
chain if they did in fact see the drifted block as valid? The drifted block
should had been seen invalid by the daemon and the network from the moment
it was mined. Stopping the drift in it's tracks and marking it to be
orphaned..

The drifted block hashes match that mined block's height hash on the rest
of the network. I guess the rest of the chain doesn't recognize the drift
occurred until the next difficulty window... That's when in certain cases
where the drift is too far, a simple block reorg might fail, and then sync
from 0 issues arise. If hash was unevenly distributed, at which time two or
more alt chains begin running in parallel, and if there is enough hash on
those alt chains they will continue to find new blocks. If majority of hash
is directed at one chain fork over another that particular fork could run
off with the chain... Beginning to drag the network along... If an exchange
follows one alt chain dragging it along over another then many transfers /
mined blocks submitted to other alt chains potentially could go to waste if
it's not corrected properly. As we know the longest chain tends to win...

Hope this helps...

stoffu:
@shopglobal 
> Although, I have witnessed a difficulty drift occur at a peers pool where only a simple reboot of the pool would re-engage the proper difficulty feed from his daemon. Must reboot pool
before mining any of the drifted blocks. He does that with a Bash bot.

Interesting, and thanks for sharing this information. Perhaps this bug may be caused by a particular way the daemon and some pool software communicate with each other. In any case, it's the daemon's bug since it must compute the difficulty correctly no matter how other clients (e.g. wallet, pool, block explorer) request data via RPC. If some pool operators can reproduce this bug and report it to us, hopefully we'll be able to fix the bug once and for all.

> In his case somehow the daemon difficulty hadn't drifted before the pool difficulty drifted...

There's no such thing as the pool difficulty, the difficulty at a given height is unique to everyone and is computed deterministically based on the sequence of blocks timestamps.

> Many pool ops have tricks up their sleeves. They think they can change or remove a few pieces of code, and if the block is still seen valid they don't care what they break if anything at all.

It doesn't matter how pool ops change their daemon or pool code, a valid block is valid no matter what, as long as the block contains the right nonce that produces a PoW hash that meets the formula: pow_hash * difficulty <= 2<sup>256</sup>. Again, the difficulty is computed deterministically as above, so there's no cheating possible.

> The drifted block that the rest of the network sees as non-drifted is an issue.

Likewise, there's no such think as a drifted block. The bug here is that the daemon very rarely computes a drifted difficulty from the exact same set of blocks.

stoffu:
Gah, fixed another bug on starting recalculation from height 1...

src/cryptonote_core/blockchain.cpp
```diff
@@ -888,6 +898,111 @@ difficulty_type Blockchain::get_difficulty_for_next_block()
...
+      CHECK_AND_ASSERT_THROW_MES(timestamps.size() == DIFFICULTY_BLOCKS_COUNT + 1, "Wrong timestamps size: " << timestamps.size());
+      timestamps.erase(timestamps.begin());
+      difficulties.erase(difficulties.begin());
+    }
+    last_cum_diff = recalculated_cum_diff;
+  }
+
+  if (drift_start_height > 0)
+  {
+    LOG_ERROR("Writing to the DB...");
```
moneromooo-monero, line 988: Leftover debug helper

stoffu: Actually I intentionally did this because it can take some time when the number of blocks to correct is large. Is this a problem?

moneromooo-monero: Ah, OK. Make it a MGINFO instead then ? Or I guess it kinda makes sense as an error. It just feels odd to have a "writing to the DB" error :)

stoffu: I'd leave it this way since I consider the whole process of correcting cumulative difficulties as an error state, and I'm using LOG_ERROR for reporting progress (XX %) and completion ("Corrected difficulties for XXX blocks").

shopglobal:
Most, if not all pools have a nodeJS utility to translate the raw fetched
from daemon. At the node level there could be misconduct, and
misconfiguration leading to, but not limited to, a 'difficulty drift' at
the pool level. Where as the daemon has not yet 'drifted'... When a valid
block is found the 'drift' is not recognized at the daemon, and the pool
basically gets away with a 'drifted' block. Usually the difference in
difficulty is minimal. He doesn't see why that would be profitable for an
attack scenario. Basically dealing with a case of precision, and not
vulnerability...

shopglobal:
There's reason to believe greedy pools mine blocks set to 20 second target

stoffu:
@shopglobal 

You fundamentally misunderstand how things work and keep saying nonsense. What I commented above is all I have to say. I won’t reply any further.

shopglobal:
Hey if you don't take the feedback for what it is, that's up to you. I
don't appreciate your snarky remarks I certainly understand the entire
blockchain, and I manage several forks. Perhaps there is an inconsistency
across the forks with different daa, or another variable.

Don't insult my intelligence, and no reply is necessary your attitude is
not welcome.

Best 🙏

UkoeHB:
@shopglobal As I understand it, this bug is encountered at the daemon level. There should be no way for the daemon to receive an invalid block (i.e. receive an invalid timestamp) since the block hash must be legitimate. This means, within the daemon, sometime after a block has been validated and by the the time the cumulative difficulty is updated, the difficulty computation goes awry. Even if the nodeJS utility is making mistakes, it shouldn't be able to affect the daemon.

selsta:
#6650

stoffu:
What's stopping this PR from getting merged?

selsta:
We did not do merges yet apart from bringing master branch to the same level as release branch.

PR #6127, opened 2019-11-13
simplewallet: noob-friendly help menu

wowario:

PR #6541, opened 2020-05-16
[release-v0.16] revert noob-friendly help menu

selsta:
Not ready for this release yet, breaks existing behaviour (`help command`). We can improve it in `master` branch and include it in the next major release.

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1649,9 +1649,8 @@ namespace cryptonote
         << "You can set the level of process detailization through \"set_log <level|categories>\" command," << ENDL
         << "where <level> is between 0 (no details) and 4 (very verbose), or custom category based levels (eg, *:WARNING)." << ENDL
         << ENDL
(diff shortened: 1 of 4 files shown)


jtgrassie:
I suggest applying this PR to master as well as it seems largely agreed the patch this reverts won't remain in its existing form. Good alternative suggestions (like gdb help / sections / using an external pager) are preferable approaches to address the *perceived problem.

*_perceived_ as I've not really seen any complaints with the way help has always functioned.

selsta:
> *perceived as I've not really seen any complaints with the way help has always functioned.

`help` prints 120 lines of commands, this is not user friendly.

jtgrassie:
> help prints 120 lines of commands, this is not user friendly.

To whom though? I would say most CLI users are _very_ used to paging/scrolling through large help pages and 120 lines is tiny. 

In any case, from the discussions, even if we do want to keep something _like_ #6127, it would very likely be changed (either external pager or sectioned help commands), hence it seems to make sense to PR this revert to master, which will ease a more thought through solution (if its agreed there is even a problem to fix).

PR #6544, opened 2020-05-16
blockchain: detect and log bad difficulty calculations

moneromooo-monero:


src/cryptonote_core/blockchain.cpp
```diff
@@ -812,12 +812,21 @@ bool Blockchain::get_block_by_hash(const crypto::hash &h, block &blk, bool *orph
...
+  LOG_PRINT_L3("Blockchain::" << __func__);
+
+  std::stringstream ss;
+  bool print = false;
+
+  int done = 0;
+  ss << "get_difficulty_for_next_block: height " << m_db->height() << std::endl;
   if (m_fixed_difficulty)
   {
+    ss << "  -> fixed: " << (m_db->height() ? m_fixed_difficulty : 1) << std::endl;
```
stoffu, line 824: Is this really needed? If so, it should be accompanied by some actual logging, otherwise it's NO-OP.

moneromooo-monero: Doesn't seem to be, I'll remove it.

PR #6550, opened 2020-05-17
easylogging++: sanitize log payload

moneromooo-monero:
Some of it might be coming from untrusted sources

### DIFF
--- a/external/easylogging++/easylogging++.cc
+++ b/external/easylogging++/easylogging++.cc
@@ -2475,6 +2475,100 @@ void DefaultLogDispatchCallback::handle(const LogDispatchData* data) {
   }
 }
 
+
+template<typename Transform>
+static inline std::string utf8canonical(const std::string &s, Transform t = [](wint_t c)->wint_t { return c; })
+{
+    std::string sc = "";
+    size_t avail = s.size();
+    const char *ptr = s.data();
+    wint_t cp = 0;
+    int bytes = 1;
+    char wbuf[8], *wptr;
+    while (avail--)
+    {
+      if ((*ptr & 0x80) == 0)
+      {
+        cp = *ptr++;
+        bytes = 1;
+      }
+      else if ((*ptr & 0xe0) == 0xc0)
+      {
+        if (avail < 1)
+          throw std::runtime_error("Invalid UTF-8");
+        cp = (*ptr++ & 0x1f) << 6;
+        cp |= *ptr++ & 0x3f;
+        --avail;
+        bytes = 2;
+      }
+      else if ((*ptr & 0xf0) == 0xe0)
+      {
+        if (avail < 2)
+          throw std::runtime_error("Invalid UTF-8");
+        cp = (*ptr++ & 0xf) << 12;
+        cp |= (*ptr++ & 0x3f) << 6;
+        cp |= *ptr++ & 0x3f;
+        avail -= 2;
+        bytes = 3;
+      }
+      else if ((*ptr & 0xf8) == 0xf0)
+      {
+        if (avail < 3)
+          throw std::runtime_error("Invalid UTF-8");
+        cp = (*ptr++ & 0x7) << 18;
+        cp |= (*ptr++ & 0x3f) << 12;
+        cp |= (*ptr++ & 0x3f) << 6;
+        cp |= *ptr++ & 0x3f;
+        avail -= 3;
+        bytes = 4;
+      }
+      else
+        throw std::runtime_error("Invalid UTF-8");
+
+      cp = t(cp);
+      if (cp <= 0x7f)
+        bytes = 1;
+      else if (cp <= 0x7ff)
+        bytes = 2;
+      else if (cp <= 0xffff)
+        bytes = 3;
+      else if (cp <= 0x10ffff)
+        bytes = 4;
+      else
+        throw std::runtime_error("Invalid code point UTF-8 transformation");
+
+      wptr = wbuf;
+      switch (bytes)
+      {
+        case 1: *wptr++ = cp; break;
+        case 2: *wptr++ = 0xc0 | (cp >> 6); *wptr++ = 0x80 | (cp & 0x3f); break;
+        case 3: *wptr++ = 0xe0 | (cp >> 12); *wptr++ = 0x80 | ((cp >> 6) & 0x3f); *wptr++ = 0x80 | (cp & 0x3f); break;
+        case 4: *wptr++ = 0xf0 | (cp >> 18); *wptr++ = 0x80 | ((cp >> 12) & 0x3f); *wptr++ = 0x80 | ((cp >> 6) & 0x3f); *wptr++ = 0x80 | (cp & 0x3f); break;
+        default: throw std::runtime_error("Invalid UTF-8");
+      }
+      *wptr = 0;
+      sc.append(wbuf, bytes);
+      cp = 0;
+      bytes = 1;
+    }
+    return sc;
+}
+
+void sanitize(std::string &s)
+{
+  s = utf8canonical(s, [](wint_t c)->wint_t {
+    if (c == 9 || c == 10 || c == 13)
+      return c;
+    if (c < 0x20)
+      return '?';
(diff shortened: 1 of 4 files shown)


src/common/utf8.h
```diff
@@ -0,0 +1,121 @@
...
+      *wptr = 0;
+      sc += T(wbuf, bytes);
+      cp = 0;
+      bytes = 1;
+    }
+    return sc;
+  }
+
+  template<typename T>
+  inline T utf8canonical(const T &s)
```
vtnerd, line 117: This overload is unnecessary due to the default argument.

src/common/utf8.h
```diff
@@ -0,0 +1,121 @@
...
+
+#include <cctype>
+#include <cwchar>
+#include <stdexcept>
+
+namespace tools
+{
+
+  template<typename T, typename Transform>
+  inline T utf8canonical(const T &s, Transform t = [](wint_t c)->wint_t { return c; })
```
vtnerd, line 39: Why is this entire function repeated in this header, and the other cpp? Was it supposed to be removed from the other cpp?

external/easylogging++/easylogging++.cc
```diff
@@ -2475,6 +2475,100 @@ void DefaultLogDispatchCallback::handle(const LogDispatchData* data) {
...
+      switch (bytes)
+      {
+        case 1: *wptr++ = cp; break;
+        case 2: *wptr++ = 0xc0 | (cp >> 6); *wptr++ = 0x80 | (cp & 0x3f); break;
+        case 3: *wptr++ = 0xe0 | (cp >> 12); *wptr++ = 0x80 | ((cp >> 6) & 0x3f); *wptr++ = 0x80 | (cp & 0x3f); break;
+        case 4: *wptr++ = 0xf0 | (cp >> 18); *wptr++ = 0x80 | ((cp >> 12) & 0x3f); *wptr++ = 0x80 | ((cp >> 6) & 0x3f); *wptr++ = 0x80 | (cp & 0x3f); break;
+        default: throw std::runtime_error("Invalid UTF-8");
+      }
+      *wptr = 0;
+      sc += T(wbuf, bytes);
```
vtnerd, line 2550: Just use `.append()` here. The code already assumes a particular constructor and `operator+=` interface; `append` is usually going to be more efficient.

external/easylogging++/easylogging++.cc
```diff
@@ -2475,6 +2475,100 @@ void DefaultLogDispatchCallback::handle(const LogDispatchData* data) {
...
+      cp = 0;
+      bytes = 1;
+    }
+    return sc;
+}
+
+void sanitize(std::string &s)
+{
+  s = utf8canonical(s, [](wint_t c)->wint_t {
+    if (c == 9 || c == 10 || c == 13)
```
vtnerd, line 2560: What are you trying to do here? `std::isprint` ? There's also [`boost::spirit::char_encoding::unicode::isprint`](https://www.boost.org/doc/libs/1_64_0/boost/spirit/home/support/char_encoding/unicode.hpp) which works specifically with unicode points regardless of locale settings. There isn't a quick/guaranteed `isprint` for unicode in C++11.

vtnerd: Ok, it looks like this is filtering for printable ascii and only space characters? This check might be off for upper characters I'd have to look again.

src/common/utf8.h
moneromooo-monero: Because easylogging++ does not depend on common.

external/easylogging++/easylogging++.cc
moneromooo-monero: Keeping stuff we want to get in logs, but not control characters.

moneromooo-monero:
It avoids going through all the code to replace LOG(s) with LOG(sanitize_for_log(s)). Which is annoying, might miss some, and does not ensure future ones don't get added.

moneromooo-monero:
I guess later utf8canonical could be changed to be two pass, first pass checks, second pass happens only if the first pass detected the string should change and makes the change.

src/common/utf8.h
vtnerd: Some independent thoughts:
* Why have template type `T`? Each "copy" of the function uses a single type.
* Using/instantiating templated code doesn't create a library dependency.
* The mnemonics code already depends on epee and easylogging.

moneromooo-monero: Because the original used string and wipeable_string. I'll remove the template on that one.

Are you telling me "you could #include "mnemonics/something" or even "common/something" in easylogging ? That'd be gross.

src/mnemonics/language_base.h
```diff
@@ -73,78 +74,11 @@ namespace Language
...
-    }
-    return sc;
-  }
-
   struct WordHash
   {
     std::size_t operator()(const epee::wipeable_string &s) const
     {
-      const epee::wipeable_string sc = utf8canonical(s);
+      const epee::wipeable_string sc = tools::utf8canonical(s, [](wint_t c) -> wint_t { return std::tolower(c); });
```
vtnerd, line 81: The calls to `std::tolower` are (and have been previously) invalid since simple wallet changes to the user environment locale which may not be unicode compatible. Presumably, everyone has been using utf8 encodings in their locale, or this would've failed. A comment or Github issue? If someone hits this edge case, lots should go wrong in the code I think.

There's `boost::spirit::char_encoding::unicode::tolower` (from the same file mentioned previously), but that drags in unicode tables from Boost. I don't see any other easy method, with Boost or C++, as you have to specify some kind of correct locale which is funky and may not be present. Boost locale can convert the entire string, but it only returns `std::string`.

external/easylogging++/easylogging++.cc
vtnerd: Comment on whats being allowed?

src/common/utf8.h
moneromooo-monero: Doesn't really belong in a logging API though. If you feel strongly about it, I guess I can do it that way.

moneromooo-monero:
What kind of comment are you after ? Those just make sense as they're already logged.

moneromooo-monero:
I'm just going to ignore the bikeshedding. If you want to change, the code to use boost or add micro optimisations, fine, but I'm not going to do it here.

src/mnemonics/language_base.h
moneromooo-monero: towlower might work ? I just saw it in the manpges, takes a wint_t, says "in the current locale", no mention of UTF-8.

vtnerd: `towlower` is better but still has the same user environment locale/encoding issue. If the user environment uses `shift_jis`, then all of this is incorrect. Although, the wallet is probably unusable in that environment (i.e. scope larger than just this snippet), and no one has filed any bug reports so (nearly) everyone uses utf8 presumable.

I commented on it because the CLI output is probably going to be messed up in that situation, and it may not be immediately obvious why.

moneromooo-monero: I'll use towlower then. Whatever improvements can be done later if you find a good way (or boost).

vtnerd:
> I'm just going to ignore the bikeshedding. If you want to change, the code to use boost or add micro optimisations, fine, but I'm not going to do it here.

This isn't bikeshedding or micro-optimizing. The codebase routinely has large functions with many variables instead of smaller re-usable functions. Ideally we move in the other direction. And this PR duplicates entire functions within the same codebase. This PR is to the master branch, not the 0.16 variant - I'm not sure why there's a need to accept this code with some performance impact when an alternative implementation too difficult.

This patch copies+allocates a new string for every log message, while a lock is being held. Log messages are rare, but the same lock must be acquired (basically) for all error/warning/info log statements to determine whether a message is to be displayed (its worse if the user bumps up logging in one category).

moneromooo-monero:
This is not new code, in case you did not notice. It was just moved (the duplication is unfortunate, though I agree). Stuff like changing the ctor to take out a strlen on a "" is time wasting though.

Issue #6547, opened 2020-05-17
import_key_images not working


SomaticFanatic:
To recreate issue:

1. Clone master. Create a fresh testnet wallet. Move over some tXMR. Send a single tx so your wallet has a mixture of sent and received funds. 

2. Create an associated view wallet. In the view wallet type `refresh`. View wallet will say `Balance: 123.456790123456, unlocked balance: 123.456790123456 (Some owned outputs have missing key images - import_key_images needed)`

3. Open original wallet. Type `export_key_images keyimagefile`. Wallet will say `Signed key images exported to keyimagefile`.

4. Open view wallet back up. Type `import_key_images keyimagefile`. Wallet will say `Signed key images imported to height [height], 0.000000000000 spent, 123.456790123456 unspent`

   0.000 spent? For some reason it isn't showing the spent funds in imported key images.

5. Type `refresh` in view wallet. Wallet will still say `Balance: 123.456790123456, unlocked balance: 123.456790123456 (Some owned outputs have missing key images - import_key_images needed)`

moneromooo-monero:
Did you export/import outputs in the other direction first ?

SomaticFanatic:
No. I’ve exported key images before and the above process worked fine. Why is it important to export outputs from a view wallet?

moneromooo-monero:
Because the cold wallet calculates key images on the outputs it gets from the hot wallet.

SomaticFanatic:
Interesting. It works! Perhaps the message can be edited to say “export_outputs then import_key_images needed”

moneromooo-monero:
FWIW, this is done automatically when you make a tx (outputs and key images piggy back on the tx signing files).

PR #6552, opened 2020-05-17
simplewallet.cpp: Add export_outputs help

SomaticFanatic:
#6547


moneromooo-monero:
If you're going to add this, might as well add the whole set of commands. Either the actual step needed, or the series, not just a couple ones in the series.

SomaticFanatic:
You’re saying I need to outline the entire offline signing process?

moneromooo-monero:
Maybe. I'm not sure what you mean by that. What I mean is that mentioning two commands out of the four is worse than mentioning either the four ones or just the one that is actually needed.

SomaticFanatic:
Tied up at the moment but I will try to get around to this next month.

SomaticFanatic:
How does this sound?

moneromooo-monero:
Looks OK. Please squash.

SomaticFanatic:
Done

SomaticFanatic:
Wew screwed up this PR so I'll close it and open a new one

PR #6556, opened 2020-05-18
serialization: fix bad rapidjson api usage

moneromooo-monero:


moneromooo-monero:
That makes me wonder whether it'd be exploitable if a user can send small strings. Guess I'll PR for the  branch too then just in case.

PR #6559, opened 2020-05-19
ByteSlice: Fix persisting ptr to std::moved SSO buffer

Doy-lee:
The Bug:
1. Construct `byte_slice.portion_` with `epee::span(buffer)` which copies a pointer to the SSO buffer to `byte_slice.portion_`
2. It constructs `byte_slice.storage_` with `std::move(buffer)` (normally this swap pointers, but SSO means a memcpy and clear on the original SSO buffer)
3. `slice.data()` returns a pointer from `slice.portion_` that points to the original SSO cleared buffer, `slice.storage_` has the actual string

### DIFF
--- a/contrib/epee/src/byte_slice.cpp
+++ b/contrib/epee/src/byte_slice.cpp
@@ -133,10 +133,13 @@ namespace epee
 
   template<typename T>
   byte_slice::byte_slice(const adapt_buffer, T&& buffer)
-    : storage_(nullptr), portion_(to_byte_span(to_span(buffer)))
+    : storage_(nullptr), portion_(nullptr)
   {
     if (!buffer.empty())
+    {
       storage_ = allocate_slice<adapted_byte_slice<T>>(0, std::move(buffer));
+      portion_ = to_byte_span(to_span(static_cast<adapted_byte_slice<T> *>(storage_.get())->buffer));
+    }
   }
 
   byte_slice::byte_slice(std::initializer_list<span<const std::uint8_t>> sources)
--- a/tests/unit_tests/epee_utils.cpp
+++ b/tests/unit_tests/epee_utils.cpp
@@ -387,6 +387,29 @@ TEST(ByteSlice, Construction)
   EXPECT_FALSE(std::is_copy_assignable<epee::byte_slice>());
 }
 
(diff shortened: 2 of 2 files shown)


moneromooo-monero:
Does it happen in practice ? If so, can you please also PR against release-0.16 ?

contrib/epee/src/byte_slice.cpp
```diff
@@ -133,10 +133,13 @@ namespace epee
 
   template<typename T>
   byte_slice::byte_slice(const adapt_buffer, T&& buffer)
-    : storage_(nullptr), portion_(to_byte_span(to_span(buffer)))
+    : storage_(nullptr)
   {
     if (!buffer.empty())
+    {
       storage_ = allocate_slice<adapted_byte_slice<T>>(0, std::move(buffer));
+      portion_ = to_byte_span(to_span(reinterpret_cast<adapted_byte_slice<T> *>(storage_.get())->buffer));
```
vtnerd, line 141: `static_cast`

contrib/epee/src/byte_slice.cpp
```diff
@@ -133,10 +133,13 @@ namespace epee
 
   template<typename T>
   byte_slice::byte_slice(const adapt_buffer, T&& buffer)
-    : storage_(nullptr), portion_(to_byte_span(to_span(buffer)))
+    : storage_(nullptr)
```
vtnerd, line 136: Explicit initialize `portion_(nullptr)`. Per C++ rules, its going to be initialized anyway to the same value, but it doesn't "look" possibly uninitialized that way.

vtnerd:
> Does it happen in practice ? If so, can you please also PR against release-0.16 ?

`byte_slice` objects are only used for outgoing data currently, so triggering this is somewhat difficult (empty p2p messages?). It should likely be applied to 0.16 anyway.

contrib/epee/src/byte_slice.cpp
Doy-lee: Added

contrib/epee/src/byte_slice.cpp
Doy-lee: Added

PR #6570, opened 2020-05-19
remove std::move() on temporary http_client object to resolve warning

woodser:
This PR fixes the following compiler warning:

monero-core/src/wallet/wallet2.cpp:1128:17: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
  m_http_client(std::move(http_client_factory->create())),
                ^
monero-core/src/wallet/wallet2.cpp:1128:17: note: remove std::move call here
  m_http_client(std::move(http_client_factory->create())),


woodser:
#6332 has this change without the subsequent merge.

PR #6571, opened 2020-05-19
net_node: remove dead seed nodes

selsta:
ping @fluffypony, you added most of them years ago, do you still have access to these machines?

I would say we should keep this PR open for a while and if they are still offline then we can merge this.


fluffypony:
Sorry not sure how I missed the ping. I asked for volunteers in #monero-dev at the time and added some that we ran ourselves. I can confirm that 5.9.100.248 and 163.172.182.165 (that I operated) are dead and can be removed. I'd imagine that any others that are unreachable should be removed, so approved from my side.

PR #6575, opened 2020-05-20
rpc: lock access to the rpc payment object

moneromooo-monero:


rbrunner7:
Just a thought that crossed my mind, without looking at the complete source: Do any of the now guarded methods call each other under some circumstances, or are they all pure "top-level" calls? If any one calls any other, that would deadlock I think.

moneromooo-monero:
Not to my knowledge.

PR #6578, opened 2020-05-21
daemon: complain if data dir resides on FAT32 volume (Windows)

xiphon:


src/daemon/main.cpp
```diff
@@ -233,6 +247,13 @@ int main(int argc, char const * argv[])
     boost::filesystem::path data_dir = boost::filesystem::absolute(
         command_line::get_arg(vm, cryptonote::arg_data_dir));
 
+#ifdef WIN32
+    if (isFat32(data_dir.root_name().c_str()))
+    {
+      MERROR("Data directory resides on FAT32 volume that has 4GiB file size limit, blockchain might get corrupted.");
+    }
```
moneromooo-monero, line 254: + return error ?

src/daemon/main.cpp
```diff
@@ -107,6 +107,20 @@ uint16_t parse_public_rpc_port(const po::variables_map &vm)
...
+bool isFat32(const wchar_t* root_path)
+{
+  std::vector<wchar_t> fs(MAX_PATH + 1);
+  if (!::GetVolumeInformationW(root_path, nullptr, 0, nullptr, 0, nullptr, &fs[0], MAX_PATH))
+  {
+    MERROR("Failed to get '" << root_path << "' filesystem name. Error code: " << ::GetLastError());
+    return false;
+  }
+
+  return wcscmp(L"FAT32", &fs[0]) == 0;
```
moneromooo-monero, line 120: Would this also catch FAT16 (and any other earlier variants that may exists) which have presumably lower or equal limits ?

xiphon: No reason to do that, FAT16 max volume size is 2GiB, even less for all the other variants.

src/daemon/main.cpp
xiphon: Think we shouldn't. We didn't hit a error yet (using `MERROR` to print the red warning). Will work pretty fine till the blockchain grows to 4GiB.
Who knows what the actual blockchain size is. Also

Also will prevent running the daemon in `--no-sync` mode.

PR #6582, opened 2020-05-23
workflows: fix windows build

selsta:


SarangNoether:
Matches the corresponding [GUI PR](https://github.com/monero-project/monero-gui/pull/2921).

PR #6587, opened 2020-05-24
fix typo in pick_preferred_rct_inputs

Parean:
change check of m_key_image_partial from td to td2


moneromooo-monero:
Could you please also PR this against the release-0.16 branch ? Please keep this one open too, we just need the two of them.

Parean:
did it #6665

moneromooo-monero:
Thanks

PR #6588, opened 2020-05-24
daemon: remove time based "update needed" status string

moneromooo-monero:


src/daemon/rpc_command_executor.cpp
```diff
@@ -521,7 +521,6 @@ bool t_rpc_command_executor::show_status() {
     % get_mining_speed(cryptonote::difficulty_type(ires.wide_difficulty) / ires.target)
     % (unsigned)hfres.version
     % get_fork_extra_info(hfres.earliest_height, net_height, ires.target)
-    % (hfres.state == cryptonote::HardFork::Ready ? "up to date" : hfres.state == cryptonote::HardFork::UpdateNeeded ? "update needed" : "out of date, likely forked")
```
hyc, line 524: Do these enums serve any other purpose any more? Or should they go away too?

trasherdk: My build of release-v0.16 claims it needs updating :)

moneromooo-monero: I don't think they're used for anything. They could go if we're reasonably sure we won't need them again.

PR #6589, opened 2020-05-25
change language around Electrum-style seeds

fluffypony:
I think "Electrum-style", or any reference to Electrum, is confusing to newcomers. Whilst we can and should keep Electrum references in the underlying code, since we use their seed derivation mechanism and even some of their word lists, I think it's ok to change the language around error and interface messages.

I've additionally renamed ```--electrum-seed``` to ```--mnemonic-seed```, and then created an aliased CLI flag for ```--electrum-seed``` to retain backwards compatibility.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -158,10 +158,11 @@ namespace
   const command_line::arg_descriptor<std::string> arg_generate_from_multisig_keys = {"generate-from-multisig-keys", sw::tr("Generate a master wallet from multisig wallet keys"), ""};
   const auto arg_generate_from_json = wallet_args::arg_generate_from_json();
   const command_line::arg_descriptor<std::string> arg_mnemonic_language = {"mnemonic-language", sw::tr("Language for mnemonic"), ""};
-  const command_line::arg_descriptor<std::string> arg_electrum_seed = {"electrum-seed", sw::tr("Specify Electrum seed for wallet recovery/creation"), ""};
-  const command_line::arg_descriptor<bool> arg_restore_deterministic_wallet = {"restore-deterministic-wallet", sw::tr("Recover wallet using Electrum-style mnemonic seed"), false};
-  const command_line::arg_descriptor<bool> arg_restore_from_seed = {"restore-from-seed", sw::tr("alias for --restore-deterministic-wallet"), false};
-  const command_line::arg_descriptor<bool> arg_restore_multisig_wallet = {"restore-multisig-wallet", sw::tr("Recover multisig wallet using Electrum-style mnemonic seed"), false};
+  const command_line::arg_descriptor<std::string> arg_electrum_seed = {"electrum-seed", sw::tr("Alias for --mnemonic-seed"), ""};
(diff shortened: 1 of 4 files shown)


erciccione:
I'm not sure we should change the language used. As you said we would keep all references to electrum in the code and the alias is nice to have, but if this gets merged we will have to adapt all documentation and also change the references in the GUI. This will also result in additional work for translators. Furthermore, people visiting stack exchange or external docs (like monerodocs.org, to mention one) will still see the references to `--electrum-seed` but they will not find them in the UI, whcih could lead to a lot of questions like "i read on stackexchange to use `--electrum-seed`, but i don't find it in the docs". The alias would make possible to still use `--electrum-seed`, but i think there would be confusion anyway.

I agree `--mnemonic-seed` is better, but i don't know if it worth the change. We could add a more specific description in the help message and/or the docs.

fluffypony:
@erciccione Alias is broken anyway - I did it in a rush, so let's keep the convo going and if we want this in then I'll fix it

selsta:
I have been confused by `--electrum-seed` in the past so I like this change.

iamamyth:
> I'm not sure we should change the language used. As you said we would keep all references to electrum in the code and the alias is nice to have, but if this gets merged we will have to adapt all documentation and also change the references in the GUI. This will also result in additional work for translators.

I somewhat agree, but with the opposite conclusion: It's not an electrum seed, so the language is wrong. Keeping the same language indefinitely means that: a) the seed format itself is confusing, as it's not an electrum seed; and b) potential changes to the seed, like encoding restore height, further muddy the waters. 

At the moment, it's the worst of all worlds: You have a non-electrum seed which is called an electrum seed, so translation accuracy is counter-productive. Anyone who knows the meaning of the term thinks they have something which *isn't* an electrum seed, so they look elsewhere; anyone following a random guide on the internet would stumble into the alias, or, eventually, the updated argument name. 

While it's ill-advised to confuse existing users for no reason, there's good reason here, and the total current user population is a tiny fraction of its potential size; delaying such a change only magnifies the disaster for everyone, and, if the software fails to add users, will anyone care either way?

SomaticFanatic:
Am a lurker to this repo, but IMHO using ‘—word-seed’ is better than ‘—mnemonic-seed’ since people don’t typically know what the word mnemonic means and the silent m confuses newcomers (good terms must be easy to remember, which studies prove correspond with being easy to pronounce).

fluffypony:
@SomaticFanatic I agree that an appropriate, accessible term should be chosen for the GUI, but the CLI is a power-user tool, and is filled with technical terms like "PEM format private key"  and "ZMQ RPC server", so I think we can get away with "mnemonic" in it;)

trasherdk:
A far as I remember, it's being called a mnemonic seed everywhere, when talked about.

Also `wallet-cli` has parameters and help text mentioning `mnemonic`
`--mnemonic-language` - `Language for mnemonic`
(It should probably say `Language for mnemonic seed`)

So, making `--electrum-seed` an alias of `--mnemonic-seed` and bitch about deprecation for a release or two, has my vote.

fluffypony:
Closing for now, will redo this later when we have some agreement

PR #6600, opened 2020-05-28
Update wallet message signing

SarangNoether:
Updates wallet message signing to support:
- hash function domain separation
- signing with either view or spend keys
- binding signatures to full addresses and signing mode


src/cryptonote_config.h
```diff
@@ -219,6 +219,7 @@ namespace config
   const unsigned char HASH_KEY_RPC_PAYMENT_NONCE = 0x58;
   const unsigned char HASH_KEY_MEMORY = 'k';
   const unsigned char HASH_KEY_MULTISIG[] = {'M', 'u', 'l', 't' , 'i', 's', 'i', 'g', 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
+  const char HASH_KEY_MESSAGE_SIGNING[] = "MessageSignature";
```
TheCharlatan, line 222: Since this key will often be used with external systems, I would propose to change this to `MoneroMessageSignature`.

Snipa22:
@SarangNoether Could you rebase this at your convenience please?  Thanks!

src/cryptonote_config.h
SarangNoether: Good idea. Will fix.

moneromooo-monero (approved):

selsta (approved):

moneromooo-monero:
That last change weirdly embeds some of my changes, Might go away once rebased though. Gotta be careful they're not undone before merge.

PR #6601, opened 2020-05-30
Improve ZMQ GetBlocksFast (Pruned Txes)

vtnerd:
Reading a chunk of transactions (via ZMQ `GetBlocksFast`) in larger blocks takes 0.5 seconds (yes seconds!), which makes for a bad time when re-scanning older viewkeys with light wallet server.

Fixing pruned txes with `GetBlocksFast` has a ~73% speedup on a ryzen 3 testbox. This is specifically for large borromean blocks, but has good speedups overall considering the newer transaction volume. The other two patches have a combined ~31% speedup for pruned blocks, and ~36% speed for unpruned blocks.

Overall reading `GetBlocksFast` went from ~0.55 seconds to ~0.11 seconds (~82% faster), which is still somewhat poor (but better). I have other optimizations to push (a schema breaking change in particular), but even the DOMless I-know-where-hex-data-is mode still tops out at 75 milliseconds; text protocols have their limitation for this sort of content.

### DIFF
--- a/src/serialization/json_object.cpp
+++ b/src/serialization/json_object.cpp
@@ -33,6 +33,8 @@
 #include <limits>
 #include <type_traits>
 
+#include "cryptonote_basic/cryptonote_basic_impl.h"
+
 // drop macro from windows.h
 #ifdef GetObject
   #undef GetObject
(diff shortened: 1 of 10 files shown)


vtnerd:
Forced push a change to the move constructor - it must be deleted since the rapidjson dom has pointers to the internal contents of a `std::string`. There are no move operations on this object in this codebase, so no other changes were necessary.

src/serialization/json_object.cpp
```diff
@@ -265,8 +268,18 @@ void fromJsonValue(const rapidjson::Value& val, cryptonote::transaction& tx)
...
   GET_FROM_JSON_OBJECT(val, tx.rct_signatures, ringct);
+
+  const auto& sigs = val.FindMember("signatures");
+  if (sigs != val.MemberEnd())
+  {
+    fromJsonValue(sigs->value, tx.signatures);
+  }
+
+  const auto& rsig = tx.rct_signatures;
+  const bool is_coinbase = !tx.vin.empty() && boost::get<txin_gen>(std::addressof(tx.vin[0]));
```
moneromooo-monero, line 280: FWIW there's a in_coinbase in cryptonote:: and it checks for vin.size() == 1, not just !empty. I'm not sure it's really that important though.

moneromooo-monero: Probably fine, since invalid txes will have got rejected before that.

selsta:
@vtnerd Please rebase

src/serialization/json_object.cpp
vtnerd: Now using `cryptonote::is_coinbase`.

vtnerd:
* Rebased (fixing some conflicts along the way)
* Switched to `cryptonote::is_coinbase`

PR #6603, opened 2020-05-31
Change ZMQ-JSON txextra to hex and remove unnecessary base fields

vtnerd:
This changes the `txextra` field from an array of integers to a hex string.  This results in a ~12% speedup on pruned txes and a ~2% speed up non-pruned txes. The mythical (but real) DOMless mode has a ~17% speedup with this change (the mode primarily accelerates hex processing).

Since this is a breaking change, I changed the major version number of the server. I decided to simultaneously drop two fields which are superfluous - the same values are always written. If the status is not `OK`, then the JSON-RPC error field is used instead.

### DIFF
--- a/src/serialization/json_object.cpp
+++ b/src/serialization/json_object.cpp
@@ -146,6 +146,26 @@ void fromJsonValue(const rapidjson::Value& val, std::string& str)
   str = val.GetString();
 }
 
+void toJsonValue(rapidjson::Writer<epee::byte_stream>& dest, const std::vector<std::uint8_t>& src)
+{
+  const std::string hex = epee::to_hex::string(epee::to_span(src));
+  dest.String(hex.data(), hex.size());
+}
+
+void fromJsonValue(const rapidjson::Value& val, std::vector<std::uint8_t>& dest)
+{
+  if (!val.IsString())
+  {
+    throw WRONG_TYPE("string");
+  }
+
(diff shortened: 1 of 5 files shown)


src/serialization/json_object.cpp
```diff
@@ -146,6 +146,26 @@ void fromJsonValue(const rapidjson::Value& val, std::string& str)
...
+  {
+    throw WRONG_TYPE("string");
+  }
+
+  dest.resize(val.GetStringLength() / 2);
+  if (!epee::from_hex::to_buffer(epee::to_mut_span(dest), {val.GetString(), val.GetStringLength()}))
+  {
+    throw BAD_INPUT();
+  }
+}
```
moneromooo-monero, line 167: If you send "f" as val, this resizes dest to 0, and to_buffer seems to do the same thing, it looks like you'll get an empty result.

vtnerd: `from_hex::to_buffer` checks the destination buffer size before writing, and returns false on size mismatches. I'll add another check here for readability + maintenance churn proofing.

moneromooo-monero: Ah you're right, it's only to_string which doesn't check for odd.

vtnerd:
Now with a `(val.GetStringLength() % 2) != 0` check (explicitly in the code referenced above) and unit test to verify.

PR #6612, opened 2020-06-01
Fix boost <1.60 compilation and fix boost 1.73+ warnings

vtnerd:
This fixes the build with Boost 1.58 and fixes all warnings with Boost 1.73+ (by including `boost/bind/bind.hpp`). Only `src/simple/simplewallet.cpp` still uses the global `_1` variables.

The [differences between `std::bind` and `boost::bind`](https://stackoverflow.com/questions/10555566/difference-between-c11-stdbind-and-boostbind).

### DIFF
--- a/contrib/epee/include/net/http_server_impl_base.h
+++ b/contrib/epee/include/net/http_server_impl_base.h
@@ -31,7 +31,7 @@
 
 
 #include <boost/thread.hpp>
-#include <boost/bind.hpp> 
+#include <boost/bind/bind.hpp>
 
 #include "net/abstract_tcp_server2.h"
 #include "http_protocol_handler.h"
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -31,7 +31,7 @@
 // IP blocking adapted from Boolberry
 
 #include <algorithm>
-#include <boost/bind.hpp>
+#include <boost/bind/bind.hpp>
 #include <boost/date_time/posix_time/posix_time.hpp>
 #include <boost/filesystem/operations.hpp>
 #include <boost/optional/optional.hpp>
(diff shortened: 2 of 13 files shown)


contrib/epee/include/net/http_server_impl_base.h
```diff
@@ -31,7 +31,7 @@
 
 
 #include <boost/thread.hpp>
-#include <boost/bind.hpp> 
+#include <boost/bind/bind.hpp>
```
moneromooo-monero, line 34: boost bind still ?

src/p2p/net_node.inl
```diff
@@ -31,7 +31,7 @@
 // IP blocking adapted from Boolberry
 
 #include <algorithm>
-#include <boost/bind.hpp>
+#include <boost/bind/bind.hpp>
```
moneromooo-monero, line 34: same ?

contrib/epee/include/net/http_server_impl_base.h
vtnerd: There are 88 occurrences of `boost::bind`, outside of the usage in `src/simplewallet/simplewallet.cpp`. There are 120 occurrences in just that file. I decided to only update the places that used `_1`, etc, to keep this smaller and readable. Updating the remaining seemed like it could be done in a subsequent patch, but I can do a `sed` call right now and re-test the compilations ... ? With or without updating `simplewallet.cpp`.

vtnerd: Update - only 39 of 88 are from in `src/` but outside of `src/simplewallet/simplewallet.cpp`. Meaning, some of the stuff in `contrib` is dead unused/code I think, so I may not need to update all 88 occurrences. Its difficult to know how many without trial and error though.

moneromooo-monero: Makes sense. I'm in favour of leaving it as is if it doesn't need fixing :)

PR #6613, opened 2020-06-01
Avoid some temporary strings when reading off the database

moneromooo-monero:


TheCharlatan:
I found some more candidates `blobdata` -> `blobdata_ref` candidates, is there a selection criteria that I am missing?

moneromooo-monero:
I don't think so. This was an old patch, so maybe those appeared later. Or I just did not know come across them. I did this back when I was trying to speed sync up, so anything not in that path would not be changed I think. Feel free to list any that are speed sensitive.

selsta:
Please rebase.

TheCharlatan:
I only saw the follow up PR's providing more context later, so while there are more potential candidates these do indeed seem to be the performance critical ones.

TheCharlatan (approved):

PR #6615, opened 2020-06-01
blockchain: store spend height with key images

moneromooo-monero:


selsta:
Please rebase.

jeffro256:
Is this PR still desired? I have some opinions about it that I would opine on, but if the feature isn't desired still by @moneromooo-monero, then we could just simply close it.

PR #6616, opened 2020-06-01
wallet: historical reserve proofs, and fix spent height on rescan

moneromooo-monero:


selsta:
Also needs a rebase.

moneromooo-monero:
Now fills blockchain_height in check_reserve_proof RPC response.

PR #6618, opened 2020-06-02
Mention correct libusb dependency

MaxXor:
In the readme the wrong libusb dependency on Debian/Ubuntu and Fedora is mentioned and outputs this error while building:
```
-- Found HIDAPI: /usr/lib/x86_64-linux-gnu/libhidapi-libusb.a  
CMake Warning at cmake/FindHIDAPI.cmake:53 (message):
  libusb-1.0 library not found, binaries may fail to link.
  CMakeLists.txt:446 (find_package)
```


MaxXor:
I found out that there is also a [missing dependency to libudev](https://github.com/monero-project/monero/blob/9da6d2969bc50e028e59c4e9cd637ebe408294b7/cmake/FindHIDAPI.cmake#L44). I'm not sure if these should be also added as udev is disabled here again:
https://github.com/monero-project/monero/blob/ac0e6f7ecde492099b4d1ccd7e2cc81d4f8f3d6b/contrib/depends/packages/libusb.mk#L11-L16

However `release-static` builds are failing without libudev installed, so I'll add this too.

moneromooo-monero:
Can you avoid changing the whole table please ?

MaxXor:
Okay, done.

PR #6629, opened 2020-06-05
blockchain: fix total_height in getblocks.bin response

moneromooo-monero:


ksemaev:
@selsta this fix didn't get into 0.16.0.1?

moneromooo-monero:
Wrong.

vbonini:
Do you have an ETA as to when this fix might get implemented?

moneromooo-monero:
It is fixed (unless you tested it and it's not).

vbonini:
Thank you for the clarification. We will test and confirm.

vbonini:
[I found your commit](https://github.com/monero-project/monero/commit/b7812d1674b9a2c5df4bd4c98f18de340098a3c4), so your previous comment now makes sense. 

We tested and confirmed fixed in 0.16.0.1. cc @ksemaev @gutenye

moneromooo-monero:
Cool, thanks.

PR #6656, opened 2020-06-14
Leak fixes in fuzz tests, and a new tx extra fuzz test

moneromooo-monero:


moneromooo-monero:
They are loaded and parsed. They don't need to be either valid or invalid, they're just there to check the parser does not crash on them.

moneromooo-monero:
We're going to merge this before the usual 10 days to try and fix oss-fuzz failing.

PR #6685, opened 2020-06-24
Update Readme Copyright; 2019 -> 2020

dginovker:
Not sure if it was left intentionally that way, if so just close.


selsta:
There is already a PR open for this :)

PR #6690, opened 2020-06-25
replace most boost serialization with existing monero serialization

moneromooo-monero:
This reduces the attack surface for data that can come from
malicious sources (exported output and key images, multisig
transactions...) since the monero serialization is already
exposed to the outside, and the boost lib we were using had
a few known crashers.

For interoperability, a new load-deprecated-formats wallet
setting is added (off by default). This allows loading boost
format data if there is no alternative. It will likely go
at some point, along with the ability to load those.

Notably, the peer lists file still uses the boost serialization
code, as the data it stores is define in epee, while the new
serialization code is in monero, and migrating it was fairly
hairy. Since this file is local and not obtained from anyone
else, the marginal risk is minimal, but it could be migrated
later if needed.

Some tests and tools also do, this will stay as is for now.


selsta:
Does not compile on macOS:


```
2020-06-25T13:30:42.0402140Z /usr/local/include/boost/serialization/serialization.hpp:126:5: error: call to 'serialize' is ambiguous
2020-06-25T13:30:42.0403680Z     serialize(ar, t, v);
2020-06-25T13:30:42.0403840Z     ^~~~~~~~~
```

moneromooo-monero:
More context ?

selsta:
See https://github.com/monero-project/monero/pull/6690/checks?check_run_id=807444020 (click on … an View raw logs)

or https://travis-ci.org/github/monero-project/monero/builds/702024225 (FreeBSD / macOS failing)

rbrunner7:
I tested the loading of an "old" MMS file, written with Boost serialization. I could not get that to work.

Steps to reproduce:
* Create a wallet `test.wallet` with master code; testnet; password empty
* Activate MMS: `mms init 2/2 alice test`. Exit wallet. You have now `test.wallet.mms` in "old format"
* Remove `test.wallet` as new code (i.e. this branch) won't load it
* Load `test.wallet` with new code, set `load-deprecated-formats` to 1
* Exit and reload wallet
* The internal attempt to load the "old" .mms file results in an error, see this log file with level 1: https://paste.debian.net/1153944/

rbrunner7:
I tested creating and then loading again a "new" MMS file, written with Epee serialization. I could not get it to work.

Steps to reproduce, all done with code from this branch, debug build:
* Create a new wallet `test.wallet`, testnet, password empty
* Activate MMS (no running pyBitmessage necessary for that): `mms init 2/2 alice test`
* Exit
* Reload
* The internal attempt to load the "new" .mms file results in an error, see this log file with level 1: https://paste.debian.net/1153946/

rbrunner7:
A comment about informing the user: In both failed tests above I never got any interactive "end-user" message on the console that there is a `.mms` present and MMS therefore should be available for the wallet, but isn't because of errors while loading. As far as I can see the errors go only into the log. IMHO this is a little unfortunate.

moneromooo-monero:
I didn't realize you could test this without bitmessage. Thanks, I'll fix.

moneromooo-monero:
done, both new and old load fine now

moneromooo-monero:
Checking whether that fixes the mac build.

rbrunner7:
I just finished some reasonably extensive MMS test for the "new" serialization that this branch implements, using 2/2 multisig and PyBitmessage. Worked.

moneromooo-monero:
Thanks. The latest tweaks I made are to fix the mac build and don't change anythinf functional.

selsta:
test-ubuntu CI: `functional_tests_rpc` time out after 360m (appears to be endless loop), it should usually take 90 minutes.

src/wallet/message_store.cpp
```diff
@@ -182,8 +181,8 @@ bool message_store::signer_labels_complete() const
 void message_store::get_signer_config(std::string &signer_config)
 {
   std::stringstream oss;
-  boost::archive::portable_binary_oarchive ar(oss);
-  ar << m_signers;
+  binary_archive<true> ar(oss);
```
vtnerd, line 184: `boost::serialization` has an implicit string-header in every archive as a quick sanity check. I recommend the same thing here - it will quickly/correctly reject `boost::serialization` archives, identify incorrectly encoded `binary_archive`'s (this detection cannot be done currently), and will make future format changes easier in the future.

The primary problem is that this format is binary and compact. A future format could overlap significantly with the encoding scheme, resulting in undetected decoding bugs.

The quickest method is likely inheriting from the existing type but with a read/write call in the constructor.

src/wallet/wallet2.cpp
```diff
@@ -6614,11 +6638,16 @@ bool wallet2::parse_unsigned_tx_from_str(const std::string &unsigned_tx_st, unsi
...
+    {
+      LOG_PRINT_L0("Not loading deprecated format");
+      return false;
+    }
     try
     {
       std::istringstream iss(s);
-      boost::archive::portable_binary_iarchive ar(iss);
-      ar >> exported_txs;
+      binary_archive<false> ar(iss);
```
vtnerd, line 6649: Is this this change accidental? This is the older version, which I think implies the file is using the previous format.

src/wallet/wallet2.cpp
```diff
@@ -6846,10 +6896,10 @@ std::string wallet2::sign_tx_dump_to_str(unsigned_tx_set &exported_txs, std::vec
 
   // save as binary
   std::ostringstream oss;
-  boost::archive::portable_binary_oarchive ar(oss);
+  binary_archive<true> ar(oss);
   try
   {
-    ar << signed_txes;
+    ::serialization::serialize(ar, signed_txes);
```
vtnerd, line 6902: This uses a return code for primary error reporting, so the serialization-output code needs updating in several places.

src/serialization/containers.h
```diff
@@ -0,0 +1,128 @@
...
+#include "serialization.h"
+
+template <template <bool> class Archive, class T> bool do_serialize(Archive<false> &ar, std::vector<T> &v);
+template <template <bool> class Archive, class T> bool do_serialize(Archive<true> &ar, std::vector<T> &v);
+
+template <template <bool> class Archive, class T> bool do_serialize(Archive<false> &ar, std::deque<T> &v);
+template <template <bool> class Archive, class T> bool do_serialize(Archive<true> &ar, std::deque<T> &v);
+
+template<typename K, typename V>
+class serializable_unordered_map: public std::unordered_map<K, V>
```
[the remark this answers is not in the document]
moneromooo-monero: Yes, I want the calls to automatically find unordered_map when it can, but unordered_map is not serializable as is because its contents typedef is const even for a non const map IIRC. So I want to override just as little as I can.

src/serialization/container.h
```diff
@@ -103,7 +99,7 @@ bool do_serialize_container(Archive<true> &ar, C &v)
       return false;
     if (i != v.begin())
       ar.delimit_array();
-    if(!::serialization::detail::serialize_container_element(ar, const_cast<typename C::value_type&>(*i)))
+    if(!::serialization::detail::serialize_container_element(ar, (typename C::value_type&)*i))
```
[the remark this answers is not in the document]
moneromooo-monero: Doesn't compile, can't recall why right now.

src/wallet/wallet2.cpp
moneromooo-monero: Looks like it was, I'll revert.

src/serialization/container.h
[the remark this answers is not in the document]
vtnerd: Ok, I can take a really good guess as to whats going on here (with help from your other response).

If the type `C` is `std::map<std::string, int>` then `C::value_type` is _always_ `std::pair<const std::string, int>` regardless of the constness of `C` itself. The STL `map` and `unordered_map` containers do not allow in-place modification of the key, only the value.

The expression `*i` is going to return `std::pair<const std::string, int>` but the `serialized_map::value_type` type is `std::pair<std::string, int>` which doesn't have this same inner-constness, thus the compilation break here with the `const_cast`. However, without the `serialized_map::value_type` "hack", then [compilation will break here](https://github.com/monero-project/monero/blob/5041de8a3b48f4db5b1a7b8f454d14bd2e3759c9/src/serialization/pair.h#L64) when trying to read from the binary archive because the `.first` element is `const`.

So this PR violates const rules, but its probably not going to invoke any bugs (right now). However, this does make the template more brittle since it removes useful compile-time checks in the C++ cast (in fact, this may have introduced a bug I haven't thought of at the moment).

Arguably that section in `pair.h` should be:

```c++
...
  typename std::remove_cv<F>::type first{}; //new
  if (!::serialization::detail::serialize_pair_element(ar, first)) // modified
    return false;
  if (!ar.stream().good())
    return false;
  ar.delimit_array();
  typename std::remove_cv<S>::type second{}; // new
  if (!::serialization::detail::serialize_pair_element(ar, second)) //modified
    return false;
  if (!ar.stream().good())
    return false;
  ar.end_array();
  p = {std::move(first), std::move(second)}; // new
}
```
then:
```c++
template<typename K, typename V>
void do_add(std::map<K, V>& c, std::pair<const K, V>&& e) { c.insert(std::move(e)); }
```
OR
```c++
template<typename K, typename V, typename T>
void do_add(std::map<K, V>& c, T&& e) { c.insert(std::forward<T>(e)); }
```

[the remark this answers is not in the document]
moneromooo-monero: The keys are not modified in the map though, I did check first.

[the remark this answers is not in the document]
moneromooo-monero: I've tried this code, but it doesn't fix, the compiler still errors out on the const_cast:

> src/serialization/container.h:103:66: error: invalid const_cast from type ‘std::pair<const crypto::public_key, cryptonote::rpc_payment::client_info>*’ to type ‘serializable_unordered_map<crypto::public_key, cryptonote::rpc_payment::client_info>::value_type*’ {aka ‘std::pair<crypto::public_key, cryptonote::rpc_payment::client_info>*’}

Could be another brick needing changing.

src/serialization/containers.h
```diff
@@ -0,0 +1,128 @@
...
+  namespace detail
+  {
+    template <typename T> void do_reserve(std::vector<T> &c, size_t N) { c.reserve(N); }
+    template <typename T> void do_add(std::vector<T> &c, T &&e) { c.emplace_back(std::move(e)); }
+
+    template <typename T> void do_add(std::deque<T> &c, T &&e) { c.emplace_back(std::move(e)); }
+
+    template <typename K, typename V> void do_add(serializable_unordered_map<K, V> &c, std::pair<K, V> &&e) { c.insert(std::move(e)); }
+
+    template <typename K, typename V> void do_add(serializable_map<K, V> &c, std::pair<K, V> &&e) { c.insert(e); }
```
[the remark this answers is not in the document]
moneromooo-monero: I used std::forward as per your comment above. It originally did not compile with std::move IIRC. std::forward is fine.

src/serialization/container.h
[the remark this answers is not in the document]
vtnerd: > The keys are not modified in the map though, I did check first.

Yes, this should all be legal because the map itself isn't being modified. Dropping to a C-cast means the template code has less checks now and in the future though.

> I've tried this code, but it doesn't fix, the compiler still errors out on the const_cast:

I forgot to mention the part where `serialize_unordered_map` type should removed entirely for this to work - the `value_type` specified is incorrect for the `const_cast`. The quickest way to test would be to comment out the `value_type` `typedef` in `serialize_unordered_map`. If that line can be commented out, the specialized map types are not needed for serialization.

As a recap, `value_type` specified by map types have a `const` key, which caused a compilation error in the `std::pair` serialization code. The new map (`serialized_unordered_map` ...) types specified a `value_type` without a `const` key to fix that issue, but then broke the `const_cast` here. Changing the few lines in the `std::pair` serialization should mean that both of those other changes (C-cast and custom map types with a different `value_type`) should no longer be needed.

src/serialization/containers.h
```diff
@@ -0,0 +1,128 @@
...
+  namespace detail
+  {
+    template <typename T> void do_reserve(std::vector<T> &c, size_t N) { c.reserve(N); }
+    template <typename T> void do_add(std::vector<T> &c, T &&e) { c.emplace_back(std::forward<T>(e)); }
+
+    template <typename T> void do_add(std::deque<T> &c, T &&e) { c.emplace_back(std::forward<T>(e)); }
+
+    template <typename K, typename V> void do_add(serializable_unordered_map<K, V> &c, std::pair<K, V> &&e) { c.insert(std::forward<std::pair<K, V>>(e)); }
+
+    template <typename K, typename V> void do_add(serializable_map<K, V> &c, std::pair<K, V> &&e) { c.insert(std::forward<std::pair<K, V>>(e)); }
```
vtnerd, line 97: I tried (unsuccessfully) to describe this earlier - the usage of `std::forward` here should work, but isn't the intended usage. The usage here should be the equivalent to a `std::move`. Only the `T&&` cases should use `std::forward`. The difference is that `template<typename T> foo(T&& arg)` undergoes ["reference collapsing"](http://thbecker.net/articles/rvalue_references/section_08.html) when `foo` is called, as part of the algorithm to compute `T`. The `const`ness and reference for the argument isn't computed with `K` and `V` here (in this PR).

Example:
```c++
template<typename T> foo(T&&);

int main()
{
  foo(std::string{}); // T -> std::string

  std::string one;
  foo(one); // T -> std::string&

  const std::string two;
  foo(two); // T -> const std::string&
}
```
whereas `std::pair<K, V>&&` is fixed to move/temporaries of type `std::pair` only.

src/serialization/containers.h
```diff
@@ -0,0 +1,128 @@
...
+
+    template <typename K, typename V> void do_add(serializable_unordered_multimap<K, V> &c, std::pair<K, V> &&e) { c.insert(std::forward<std::pair<K, V>>(e)); }
+
+    template <typename T> void do_add(std::unordered_set<T> &c, T &&e) { c.insert(std::forward<T>(e)); }
```
vtnerd, line 101: It just occurred to me that isn't perfect-forwarding because `T` has to match on two template arguments. Typically the second argument would be a `U&&` because it gives the caller more flexibility (an actual universal reference). I guess this can be left as-is, but the code+compiler might infuriate someone if they aren't looking closely when making changes to the serialization stuff.

moneromooo-monero: How about you just patch it instead of hoping I type the expected code by chance since I know nothing of this ? :)

src/serialization/containers.h
UkoeHB: Learned about [perfect forwarding](https://eli.thegreenplace.net/2014/perfect-forwarding-and-universal-references-in-c) today. What you say makes sense vtnerd. @moneromooo-monero it seems like you could replace `std::forward->std::move` since `std::forward` will always return an r-value reference (i.e. it will always be passed an r-value reference). An example why an l-value can't be passed into `e`:
```
void print(int &x)
{
	std::cout << x << " l-value print\n";
}

void print(int &&x)
{
	std::cout << x << " r-value print\n";
}

template <typename T>
void example(T &x, T &&y)
{
	std::cout << x << '\n';		//for compiler warning
	print(std::forward<T>(y));
}

template <typename T>
void example2(T &&x)
{
	std::cout << x << '\n';		//for compiler warning
	print(std::forward<T>(x));
}

int main()
{
	int x{3};

	example(x,4);	  //r-value print
	//example(x,x);  //error, no matching function; special type-deduction rule doesn't apply
	example2(x);	  //l-value print, special type-deduction rule applies
	example2(4);	  //r-value print

	return 0;
}
```

UkoeHB: vtnerd might it be better to keep `std::forward` if it reflects the intended behavior of `do_add`? e.g. if there were an overloaded version that accepted l-values, would they be forwarded or moved?

selsta:
Please rebase.

selsta (approved):
Tested wallet loading, import/export outputs and key images on macOS.

PR #6691, opened 2020-06-25
cmake: allow custom openssl path on macOS

selsta:
It is now possible to specify a custom openssl dir by using `-DOPENSSL_ROOT_DIR` on macOS.

Tested with gitian, works as expected now. The old `OpenSSL_DIR` was probably a typo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -426,7 +426,7 @@ endif ()
 
 if (APPLE AND NOT IOS)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=x86-64 -fvisibility=default -std=c++11")
-  if (NOT OpenSSL_DIR)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -426,7 +426,7 @@ endif ()
 
 if (APPLE AND NOT IOS)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=x86-64 -fvisibility=default -std=c++11")
-  if (NOT OpenSSL_DIR)
```
vtnerd, line 429: Does this fetch and install OpenSSL automatically? That seems odd to have in the Monero build process.

selsta: If `OPENSSL_ROOT_DIR` is not set manually it will check if `openssl` installed with brew and get its path. Nothing gets fetched or downloaded.

PR #6692, opened 2020-06-25
[oaes-lib] Fix 'ftime deprecated' warning, add ISAAC for use in oaes

who-biz:
- ftime has been deprecated, and gcc 9.3+ issues a warning as a result
- this commit ports in code from the public domain, as well as the Android Open Source Project
- should prevent things from breaking in the future

### DIFF
--- a/src/crypto/CMakeLists.txt
+++ b/src/crypto/CMakeLists.txt
@@ -33,6 +33,7 @@ set(crypto_sources
   crypto-ops-data.c
   crypto-ops.c
   crypto.cpp
+  ftime.c
   groestl.c
   hash-extra-blake.c
   hash-extra-groestl.c
@@ -43,6 +44,7 @@ set(crypto_sources
   jh.c
   keccak.c
   oaes_lib.c
+  rand.c
   random.c
   skein.c
   slow-hash.c
@@ -63,6 +65,7 @@ set(crypto_private_headers
   chacha.h
   crypto-ops.h
   crypto.h
+  ftime.h
   generic-ops.h
   groestl.h
   groestl_tables.h
@@ -74,9 +77,11 @@ set(crypto_private_headers
   keccak.h
   oaes_config.h
   oaes_lib.h
+  rand.h
   random.h
   skein.h
   skein_port.h
+  standard.h
   CryptonightR_JIT.h
   CryptonightR_template.h)
 
--- a/src/crypto/ftime.c
+++ b/src/crypto/ftime.c
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) 2013 The Android Open Source Project
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *  * Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  * Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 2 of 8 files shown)


who-biz:
Windows build appears to be failing with error(s), from CI:
```
      |                             ^~~~~
D:/a/monero/monero/src/crypto/ftime.c:31:5: error: conflicting types for 'ftime'
   31 | int ftime(struct timeb* tb) {
      |     ^~~~~
In file included from C:/msys64/mingw64/x86_64-w64-mingw32/include/time.h:300,
                 from C:/msys64/mingw64/x86_64-w64-mingw32/include/sys/time.h:10,
                 from D:/a/monero/monero/src/crypto/ftime.c:28:
C:/msys64/mingw64/x86_64-w64-mingw32/include/sys/timeb.h:108:29: note: previous definition of 'ftime' was here
  108 |   __CRT_INLINE void __cdecl ftime(struct timeb *_Tmb) {
      |                             ^~~~~
```

I just force-pushed in an attempt to fix the error on Windows. Will amend if CI fails again

who-biz:
Still failing on Windows. Reverted previous force-push, until I drum up a workaround

sumogr:
lose ftime.c and ftime.h and add 
```
#include <sys/time.h>
#include <sys/timeb.h>
```
and use gettimeofday everywhere instead of ftime

who-biz:
@sumogr 

We could do that.  But, previously, ISAAC wasn't being used for random number generation, despite OpenAES making use of it.  This commit includes files to make that possible, too.

If it's preferred, I can switch things over to `gettimeofday`, but retain the ISAAC additions

moneromooo-monero:
Please, no. If you must fix an error with recent gcc, just remove those ftime calls. We don't use this code (and should not). Feel free to also add an assert that it's not called.

who-biz:
> Please, no. If you must fix an error with recent gcc, just remove those ftime calls. We don't use this code (and should not). Feel free to also add an assert that it's not called.

That what isn’t called?

moneromooo-monero:
The "random" number generator of course, the thing you were modifying.

who-biz:
You’re mistaken. It is called in place of `srand/rand` anywhere that is wrapped with `OAES_HAVE_ISAAC`, provided that this line, and wrapping conditional, are uncommented (https://github.com/monero-project/monero/blob/93257997bd69e25c76de8ebf3f46bfeb44f11589/src/crypto/oaes_config.h#L38) in`oaes_config.h`. I assume those lines were commented out due to previously lacking support for ISAAC.

who-biz:
Is there a reason to prefer `srand`?

moneromooo-monero:
What is even ISAAC and why might we want it ?

who-biz:
For one: it’s fast. Here’s some <a href=“ https://www.burtleburtle.net/bob/rand/isaacafa.html”>info on it</a>. There’s also some (limited) info on Wikipedia.

moneromooo-monero:
Interesting. Still, we have no use for it. As I said, we don't use the RNG stuff in oaes. The fact some of it is run at struct init is irrelevant. We only use keys from outside oaes. So either fix the ftime warning if you want, or close this, because we don't want to import an extra PRNG for no good reason.

who-biz:
> As I said, we don't use the RNG stuff in oaes.

That's not true.

`oaes_get_seed()` is called in `oaes_alloc()`... This occurs in the same spot (in `oaes_lib.c`) that the ISAAC's `rand` function is used as well as `srand`, if ISAAC isn't present.  `oaes_alloc()` is called as frequently as `oaes_key_import_data()` in `slow-hash.c`.

Why is this relevant?  Well, look at the <a href="https://eprint.iacr.org/2017/298.pdf">performance difference between glibc's random() and ISAAC</a>:

who-biz:
That said, if you still can't be persuaded, I can remove it and we can just use `gettimeofday`.

But using functionality that OpenAES can support makes sense when there are visible benefits

moneromooo-monero:
As I said. We do not use these keys. The fact that the ISAAC-or-libc PRNG is initialized is *irrelevant*. We still do not use those keys. Do not add gettimeofday. Either just removing ftime, or remove the whole thing so it can't be misused. Last reply here, I'll change it myself whenever I get problem locally if you don't post a patch that just removes.

who-biz:
Eh, wound up being buggy anyway after further testing. All good

PR #6693, opened 2020-06-25
[depends] Upgrade boost to 1.68

who-biz:
- Higher versions apparently append '-x64' to the library names, which breaks things
- As a result, upgrading to 1.69+ will require futher modifications.

Please also note: I had tried updating OpenSSL to a version above 1.0, as everything below 1.1 has reached EOL.  

However, openssl has overhauled their build process such that `Makefile.org` is no longer included in the repo.  They have transitioned to a 'Configure-first' build.  As such, build scripts appear to require some heavy modifications prior to upgrading to 1.1+.

### DIFF
--- a/contrib/depends/packages/boost.mk
+++ b/contrib/depends/packages/boost.mk
@@ -1,8 +1,8 @@
-package=boost                                                                                                                                                                                                                      
-$(package)_version=1_64_0
-$(package)_download_path=https://dl.bintray.com/boostorg/release/1.64.0/source/
+package=boost
+$(package)_version=1_68_0
+$(package)_download_path=https://dl.bintray.com/boostorg/release/1.68.0/source/
 $(package)_file_name=$(package)_$($(package)_version).tar.bz2
-$(package)_sha256_hash=7bcc5caace97baa948931d712ea5f37038dbb1c5d89b43ad4def4ed7cb683332
+$(package)_sha256_hash=7f6130bc3cf65f56a618888ce9d5ea704fa10b462be126ad053e80e553d6d8b7
 $(package)_dependencies=libiconv
 $(package)_patches=fix_aroptions.patch
 


jtgrassie:
What's the specific reason for this upgrade of boost?

who-biz:
@jtgrassie there are many updates that occurred in `boost` between versions 1.64 and 1.68. Most of which that affect Monero, are within `boost::asio` and `boost::thread`.  However, other components of boost that Monero use are also affected. 

<a href=“ https://www.boost.org/users/history/version_1_67_0.html”>Boost 1.67’s changelog</a> is the largest, of the 4 versions included in this update.

jtgrassie:
@who-biz I'd like to know the *specific* reason why an upgrade is required here. Not a changelog between the library versions. I.e. is there some bug, security vulnerability or feature that means we must update to 1.68 specifically or is this just an upgrade for upgrade sake?

who-biz:
OpenSSL 1.1 broke compatibility with boost 1.62 and below.  Since OpenSSL 1.0 is no longer being supported, it seems like a good practice to keep boost as updated as possible (without requiring additional changes).  Monero already supports higher versions than 1.70 for boost. So, this seems a worthwhile update to me

jtgrassie:
> OpenSSL 1.1 broke compatibility with boost 1.62 and below.

Right but we are using 1.64, not 1.62.

> Monero already supports higher versions than 1.70 for boost

Supporting higher versions is not the same as the what the minimum supported version is. IMO we should be using the lowest supported version in depends, not some arbitrarily chosen higher version, unless there's a specific reason. which it seems is not the case here.

who-biz:
Whatever you want, Jethro.

jtgrassie:
It's not whatever I want. I just asked for the specific reason as to why the need to upgrade our current 1.64 to 1.68. If there is no reason, the merge is not needed.

moneromooo-monero:
Really, it's down to TheCharlatan to decide, since this is what gets used by the depends system, so it's down to whatever was found to work.

who-biz:
As I said above, this is the highest version of boost that would work without modifications. Seems reasonable to leave it up to @TheCharlatan, either way.

selsta:
ping @TheCharlatan

who-biz:
Closing this. Recommend updating to boost 1.75.

1.74 is broken, header files are excluded from the shipped boost for library version serializations and require manual modification of boost files.  1.75 works fine, and build process does not need modified.  Based on the response I got above, won't be PRing it.

PR #6695, opened 2020-06-27
fix warning caused by std::move()

woodser:
Sorry, I accidentally deleted the source branch for this PR again.  It will stay put this time. :)


woodser:
Nevermind, this change is open in #6570

PR #6700, opened 2020-07-07
db_lmdb: set MDB_NOLOCK in read only mode

moneromooo-monero:
Not setting it would still write to the lock file, which is
likely read only if we're opening in read only mode


hyc:
No. If there are other processes using the environment, using NOLOCK here would allow this process to get corrupted reads.

moneromooo-monero:
How can you open a readonly db ?

hyc:
Even if you only want read access, you still need write access to the lock file, if you're sharing with other processes.
The only way to get readonly access without any other write access is if the DB is on a read-only filesystem.

moneromooo-monero:
Thanks. In my case I was trying to access a db from another user account with only read access, and monerod was writing.

PR #6703, opened 2020-07-11
daemon: don't print "(pruned)" for coinbase txes

moneromooo-monero:
Pruned coinbase txes are the same as unpruned ones, so the
prunable data is empty


UkoeHB:
Might be nice to add a comment indicating under what conditions this is true:

`res.txs.front().prunable_hash != empty_hash;`

e.g. "//the prunable_hash will equal empty_hash when the transaction is coinbase; coinbase can't be pruned, but without this check are mistakenly labeled 'pruned'"

PR #6718, opened 2020-07-21
rolling_median_t: fix wraparound case

UkoeHB:
Originally when the rolling median has an even number of elements stored, it returns the average of the middle two elements like this:
`return (a+b)/2`
However, if `(a + b)` is greater than the size of the type of `a` and `b`, the result might wrap around (e.g. if it's an integral type, etc.).

While it is unlikely wraparound will ever occur based on how Monero uses the rolling median, it's better to have safe code in case this template gets reused elsewhere. The new equation is
`(a+b)/2 = (a/2) + (b/2) + ((a - 2*(a/2)) + (b - 2*(b/2)))/2`

This is intended to play nice with integral division, however floating point types may have more rounding errors due to the additional operations.

### DIFF
--- a/contrib/epee/include/misc_language.h
+++ b/contrib/epee/include/misc_language.h
@@ -106,6 +106,14 @@ namespace misc_utils
 		return true;
 	}
 
+  template <typename T>
+  T get_mid(const T &a, const T &b)
+  {
+    //returns the average of two numbers; overflow safe and works with at least all integral and floating point types
+    //(a+b)/2 = (a/2) + (b/2) + ((a - 2*(a/2)) + (b - 2*(b/2)))/2
+    return (a/2) + (b/2) + ((a - 2*(a/2)) + (b - 2*(b/2)))/2;
+  }
+
   template<class type_vec_type>
   type_vec_type median(std::vector<type_vec_type> &v)
   {
@@ -122,7 +130,7 @@ namespace misc_utils
       return v[n];
     }else 
     {//2, 4, 6...
-      return (v[n-1] + v[n])/2;
+      return get_mid<type_vec_type>(v[n-1],v[n]);
     }
 
   }
--- a/contrib/epee/include/rolling_median.h
+++ b/contrib/epee/include/rolling_median.h
@@ -34,6 +34,8 @@
 
 #pragma once
 
+#include "misc_language.h"
+
 #include <stdlib.h>
 #include <stdint.h>
 
@@ -226,7 +228,7 @@ public:
     Item v = data[heap[0]];
     if (minCt < maxCt)
     {
-      v = (v + data[heap[-1]]) / 2;
+      v = get_mid<Item>(v, data[heap[-1]]);
     }
     return v;
   }
--- a/tests/unit_tests/rolling_median.cpp
+++ b/tests/unit_tests/rolling_median.cpp
@@ -170,6 +170,17 @@ TEST(rolling_median, history_blind)
   }
 }
 
+TEST(rolling_median, overflow)
+{
+  epee::misc_utils::rolling_median_t<uint64_t> m(2);
+
+  uint64_t over_half = static_cast<uint64_t>(3) << static_cast<uint64_t>(62);
+  m.insert(over_half);
+  m.insert(over_half);
+  ASSERT_EQ((over_half + over_half) < over_half, true);
+  ASSERT_EQ(over_half, m.median());
+}
+
 TEST(rolling_median, size)
 {
   epee::misc_utils::rolling_median_t<uint64_t> m(10);


moneromooo-monero:
That looks awfully complicated. Why does something like this not work ?
```
  w = data[heap[-1]];
  v = v/2 + w/2 + (v&w&1);
```
Also get rid of the gratuitous rename please.

UkoeHB:
`w = data[heap[-1]];
  v = v/2 + w/2 + (v&w&1);`
Doesn't work for non-integral types. Also note that `b + (a - b)/2` is not wraparound safe for signed types.

"Also get rid of the gratuitous rename please."
It's a readibility improvement that is directly related to the update, trivial, and safely isolated.

moneromooo-monero:
The element is a uint64_t, wihch is integral and unsigned.
A readability improvement from renaming v to a ? Did you upload the wrong patch ?

moneromooo-monero:
Oh whatever, the renaming is small enough, I can't be arsed arguing further, have it in if you want. The formula's way more complicated than it needs to though, since you appear to have made it work for types we don't need.

UkoeHB:
I agree the equation isn't pretty, however this template has no controls on its type. What about adding a type trait check?
```
include <type_traits>

template<typename Item, typename std::enable_if_t<std::is_unsigned<Item>::value>>
```

`std::is_unsigned` apparently only returns true for unsigned integral types.

moneromooo-monero:
Sure, that would be fine.

moneromooo-monero:
I am not sure whether the heap values will be always sorted. Might be worth adding a test to avoid underflowing too unless you're sure the values will always be sorted.

UkoeHB:
Good point, this algorithm is incredibly opaque. Apparently maxHeap uses negative indices, contrary to what I thought all along. To be on the safe side I used your recommended `(a/b) + (b/2) + (a&b&1)` expression.

UkoeHB:
A test failed. The rolling median is tested against a static median that also appears vulnerable to overthrow. [This line](https://github.com/monero-project/monero/blob/36344e8e046261e93116c71b6363cb79b850492e/contrib/epee/include/misc_language.h#L125) related to [this test](https://github.com/monero-project/monero/blob/5d850dde99ca282d2c6f9a1b1eb00a3812cb15be/tests/unit_tests/rolling_median.cpp#L65).

It seems like the median in misc_language.h should also be updated. However, epee::misc_utils::median() isn't just used for unsigned integers. [For example...](https://github.com/monero-project/monero/blob/a0d179e528220137885fee0817dc792127aec18c/tests/unit_tests/output_selection.cpp#L130)

Therefore unfortunately I believe we need to insert that ungainly expression into epee::misc_utils::median().

moneromooo-monero:
std::min(a,b) + (std::max(a,b) - std::min(a,b))/2 ?

moneromooo-monero:
Ah, nvm, off by one for integrals. Guess it could be made with a template get_mid function.

moneromooo-monero:
Actually it might not be off by one.

UkoeHB:
The problem is signed integrals can overflow.
```
char a = 100;
char b = -100;
char median = b + (a - b)/2;
assert(median == 0);  //false, due to overflow it's -100 + (200)/2 = ???
```

UkoeHB:
Updated with a new 'get_mid' template, and a unit test for demonstrating the rolling median is safe against overflow. Let me know if there is a better file for 'get_mid', or anything like that. While the expression looks absurd, it at least works generically.

UkoeHB:
I rolled back the type trait check, since 'get_mid' works generically.

moneromooo-monero:
The equation symbolically balances. I have no idea whether it's overflow/underflow safe though (apart from trying a few examples). Is it mentioned in literature ?

UkoeHB:
I'm not aware of any official-sounding exposition of the expression, although [this website](https://www.geeksforgeeks.org/compute-average-two-numbers-without-overflow/) has something similar. Here is a short argument, let me know what you think.
```
Variables a, b
(a+b)/2 = (a/2) + (b/2) + ((a-2*(a/2)) + (b-2*(a/2)))/2
(a-2*(a/2)) -> information lost by integer division

TYPE_MAX >= a >= b
TYPE_MAX/2 >= a/2 + lost_info_a/2
TYPE_MAX/2 >= b/2 + lost_info_b/2
TYPE_MAX/2 + TYPE_MAX/2 >= a/2 + b/2 + lost_info_a/2 + lost_info_b/2
TYPE_MAX >= a/2 + b/2 + (lost_info_a + lost_info_b)/2
```

moneromooo-monero:
Not quite the same as in the patch, but reading the one in the patch carefully, I understand it and it looks equivalent to "add one if both a and b are odd", which is right. Thanks.

PR #6727, opened 2020-07-30
A few error handling fixes

moneromooo-monero:


external/easylogging++/easylogging++.cc
```diff
@@ -2968,6 +2968,16 @@ void Writer::initializeLogger(Logger *logger, bool needLock) {
 }
 
 void Writer::processDispatch() {
+  static std::atomic_flag in_dispatch;
```
vtnerd, line 2971: Is this to manage the case where an exception is thrown from the logger, and therefore its state needs to be reset? This is ... a hack. If something _must_ be cleaned up, like releasing a lock, that's the purpose for RAII in C++. Is the upstream easylogging++ implemented like this (improper lock cleanup)?

moneromooo-monero:
No. That'd basically be the whole code, since RPC can reach a huge amount of stuff. Can you point to an example of an exception where killing the process is wanted ?

external/easylogging++/easylogging++.cc
moneromooo-monero: Yes. In this case, an exception in sanitizing the string to be logged. I'd say yes to your last question, though it's not impossible they made sure no exception could arise. Sounds doubtful though.

src/cryptonote_core/blockchain.cpp
```diff
@@ -4396,6 +4399,14 @@ bool Blockchain::add_new_block(const block& bl, block_verification_context& bvc)
 
   rtxn_guard.stop();
   return handle_block_to_main_chain(bl, id, bvc);
+
+  }
+  catch (const std::exception &e)
+  {
+    LOG_ERROR("Exception at [add_new_block], what=" << e.what());
```
xiphon, line 4406: Unfortunately `LOG_ERROR` might throw as well. If we are ok with it, we could add exceptions logging to `epee: guard against exceptions in RPC handlers` try catch blocks.

moneromooo-monero: True, but that's like I/O error and OOM, local things. If you remove logging, you still have things like assigning a string literal to a std::string, which can throw too (OOM), and the packing of the structure to json, and the I/O to send the error response back to the client.

xiphon: Sounds like it might be worth it to add `e.what()` logging to https://github.com/monero-project/monero/pull/6727/files#diff-f6bde7308082ff9a0f25bf8800332d06R113 and similar places.

vtnerd:
> No. That'd basically be the whole code, since RPC can reach a huge amount of stuff. Can you point to an example of an exception where killing the process is wanted ?

I can think of several places, just not in the HTTP-RPC path. I'd have to look. My primary question is still why, when the changes were ultimately a red-herring to the original issue reported (caused by buggy log code + stack trace injection logic, AFAIK).

moneromooo-monero:
I don't understand your point. EA breaks when logging from the dispatch function, and this patch ensures this does not happen. Are you saying "it would be better to make this reentrant instead of bailing" ? If so, I agree (but I have no idea how to do that atm). Are you saying omething else ?

external/easylogging++/easylogging++.cc
vtnerd: This chunk of code is done while a lock is held, so the `atomic_flag` can be a `bool` which will reduce the atomic instructions emitted. I don't think its possible for this chunk to execute without a lock - the entire function would likely have to be re-written.

Additionally, this likely fixes a particular symptom but not the cause. Calling the log function from the custom `__cxa_throw` is likely still problematic since the logger has multiple places where it can potentially throw. Luckily, these cases are rare. However, it seems prudent to disable stack printing on release builds entirely. Or at least make that the default. Our custom `__cxa_throw` calls several functions which can throw, and all of this occurs before the library starts it logic to track the number of active exceptions. Again, probably rare, but if easylogging++ throws on file write failures for instance, the result is likely a stackoverflow of attempted exception throws.

vtnerd:
> I don't understand your point. EA breaks when logging from the dispatch function, and this patch ensures this does not happen. Are you saying "it would be better to make this reentrant instead of bailing" ? If so, I agree (but I have no idea how to do that atm). Are you saying omething else ?

The `try { } catch {}` blocks in the HTTP server likely don't do anything to correct the original issue reported. The check in the logger was the fix. So this change is likely altering control-flow to fix a bug that it cannot fix.

external/easylogging++/easylogging++.cc
moneromooo-monero: This would defeat the purpose though, which is to get people to check whether interesting exceptions were thrown when something breaks.

moneromooo-monero: AFAICT only tools::stack_trace can throw, the rest is C. I suppose I could put an atomic flag in there to forbid reentrancy. That seems more encompassing...

vtnerd: As a reminder - this `std::atomic_flag` can be a `bool`. This code is inside a lock.

> This would defeat the purpose though, which is to get people to check whether interesting exceptions were thrown when something breaks.

The information obtained doesn't worth the risk in hijacking the library routine. The daemon logs plenty already. A macro can achieve the same thing in critical areas without covertly hijacking control flow at _every_ `throw`.

> AFAICT only tools::stack_trace can throw, the rest is C. I suppose I could put an atomic flag in there to forbid reentrancy. That seems more encompassing...

Adding another atomic flag in the stack trace function is masking the symptom, just like this flag here. Easylogging++ was not written with the programmer thinking: "any potentially throwing function call can immediately call back into the top-level log routine" - the expectation is that easylogging++ functions will not do any logging. Hopefully the location for the flag is sufficient and not hastily selected.

And the "only one function can throw" is a misrepresentation - the easylogging++ log function or `std::cout` calls many C++ functions which can potentially throw. Most (probably nearly all) are `std::bad_alloc`, while others are from iostream. Either could result in infinite exceptions with the library being hijacked, but it should be uncommon. I did a quick scan (less than a few minutes) to find other likely to throw functions but nothing stood out (thankfully).

moneromooo-monero:
The epee try/catch does fix the problem.

PR #6733, opened 2020-08-01
Don't forget size of prunable txn part

### LINKED ISSUE #6732: monero-blockchain-stats underreports size of prunable txs
Patch coming shortly

hyc:
Fixes #6732

### DIFF
--- a/src/blockchain_utilities/blockchain_stats.cpp
+++ b/src/blockchain_utilities/blockchain_stats.cpp
@@ -70,6 +70,7 @@ int main(int argc, char* argv[])
   const command_line::arg_descriptor<bool> arg_hours  = {"with-hours", "with txns per hour", false};
   const command_line::arg_descriptor<bool> arg_emission  = {"with-emission", "with coin emission", false};
   const command_line::arg_descriptor<bool> arg_fees  = {"with-fees", "with txn fees", false};
+  const command_line::arg_descriptor<bool> arg_diff  = {"with-diff", "with difficulty", false};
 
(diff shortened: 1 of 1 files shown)


src/blockchain_utilities/blockchain_stats.cpp
```diff
@@ -203,16 +210,20 @@ plot 'stats.csv' index "DATA" using (timecolumn(1,"%Y-%m-%d")):4 with lines, ''
...
   uint64_t totins = 0, totouts = 0, totrings = 0;
   boost::multiprecision::uint128_t prevemission = 0, prevfees = 0;
   boost::multiprecision::uint128_t emission = 0, fees = 0;
-  uint32_t minins = 10, maxins = 0;
-  uint32_t minouts = 10, maxouts = 0;
-  uint32_t minrings = 50, maxrings = 0;
+  boost::multiprecision::uint128_t totdiff = 0, mindiff = 0, maxdiff = 0;
+  uint32_t minins = MAX_INOUT, maxins = 0;
+  uint32_t minouts = MAX_INOUT, maxouts = 0;
+  uint32_t minrings = MAX_RINGS, maxrings = 0;
```
moneromooo-monero, line 226: Why not just 0xffffffff or std::numeric_limits<uint32_t>::max() ?

hyc: Didn't seem necessary at the time, and the latter is just too long to type every time. But sure, I can use one of those in the #define instead.

Issue #6726, opened 2020-07-30
Transaction sanity check failed with a larger amount TX


Lafudoci:
When I was trying to move all funds from my old wallet to a new one, it showed `internal error: Transaction sanity check failed`. I've tried to delete wallet cache, recover a new wallet from 25 words and change different remote nodes. None of them worked.

But when I sent a smaller amount likes 1 XMR instead, the TX went through without any issue.

Here is the log, any idea?
https://pastebin.ubuntu.com/p/xPWvhG3vgn/

selsta:
Can you try to do a transaction and if it fails do the following:

`unset_ring <txid>`

and then try again?

Lafudoci:
> Can you try to do a transaction and if it fails do the following:
> 
> `unset_ring <txid>`
> 
> and then try again?

Could you please explain more? The failed tx didn't provide txid.

moneromooo-monero:
From the message, it's failing to even create a tx, it's the wallet failing before actually making a tx, not the daemon. xiphon added retry code, but it looks like it's failing this, which seems very very unlikely, so there is probably a bug in this code.

Lafudoci:
> From the message, it's failing to even create a tx, it's the wallet failing before actually making a tx, not the daemon. xiphon added retry code, but it looks like it's failing this, which seems very very unlikely, so there is probably a bug in this code.

I see. Is there a temporary workaround that I can move all my fund? Will splitting into small amount txs do the trick?

selsta:
Please see #6815, might be related.

Lafudoci:
> Please see #6815, might be related.

Yes, I finally got my XMR successfully sent. The issue was resolved by PR 6815. Thank you.

PR #6736, opened 2020-08-02
epee: further defending against exceptions in command handlers

moneromooo-monero:


vtnerd:
My thoughts for this patch mirror those in [PR #6726](https://github.com/monero-project/monero/pull/6727). Doing a `catch (const std::exception&)` isn't getting to the source of any reported issues, its merely preventing shutdown. There are definitely exceptions in the DB code (such as LMDB map resize fail, invalid metadata state, etc) or `std::logic_error`s that are intended to unrecoverable shutdown conditions. Most of the original cryptonote code used return-codes for recoverable conditions, with the major exception being wallet code (but otherwise this seemed to be the generally control flow).

I'm not aware of any immediate bugs this introduces, but hard-to-debug issues like "stops accepting connections but keeps running" are more likely to occur. This risk is probably worth taking in the release branch, but the master branch is where the actual issue(s) are ideally fixed.

PR #6745, opened 2020-08-05
Use recent block times for unlock time

moneromooo-monero:
From an original patch by TheCharlatan.


src/cryptonote_core/blockchain.h
```diff
@@ -1427,16 +1440,6 @@ namespace cryptonote
...
-    /**
-     * @brief get the "adjusted time"
-     *
-     * Currently this simply returns the current time according to the
-     * user's machine.
-     *
-     * @return the current time
-     */
-    uint64_t get_adjusted_time() const;
-
```
UkoeHB, line 1439: Why was this declaration moved up 400 lines? New spot seems less appropriate based on other declarations nearby.

src/simplewallet/simplewallet.cpp
```diff
@@ -8456,8 +8456,8 @@ bool simple_wallet::get_transfers(std::vector<std::string>& local_args, std::vec
         }
         else
         {
-          uint64_t current_time = static_cast<uint64_t>(time(NULL));
-          uint64_t threshold = current_time + (m_wallet->use_fork_rules(2, 0) ? CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V2 : CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V1);
+          const uint64_t adjusted_time = m_wallet->get_daemon_adjusted_time();
+          uint64_t threshold = adjusted_time + (m_wallet->use_fork_rules(2, 0) ? CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V2 : CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V1);
```
UkoeHB, line 8460: Looks like terminology confusion again, since 'adjusted_time' has been changed to mean 'median time of last 60 blocks'. Which makes the behavior here change.

src/cryptonote_core/blockchain.cpp
```diff
@@ -3798,17 +3818,14 @@ bool Blockchain::check_block_timestamp(const block& b, uint64_t& median_ts) cons
...
-  for(;offset < h; ++offset)
+  median_ts = get_adjusted_time(h);
+  if(b.timestamp < median_ts)
   {
-    timestamps.push_back(m_db->get_block_timestamp(offset));
+    MERROR_VER("Timestamp of block with id: " << get_block_hash(b) << ", " << b.timestamp << ", less than median of last " << BLOCKCHAIN_TIMESTAMP_CHECK_WINDOW << " blocks, " << median_ts);
+    return false;
   }
 
-  return check_block_timestamp(timestamps, b, median_ts);
```
UkoeHB, line 3811: Is this 3-parameter overload of `check_block_timestamp()` being deprecated? If so, might be nice to remove it completely.

src/simplewallet/simplewallet.cpp
```diff
@@ -10156,8 +10156,8 @@ bool simple_wallet::show_transfer(const std::vector<std::string> &args)
       }
       else
       {
-        uint64_t current_time = static_cast<uint64_t>(time(NULL));
-        uint64_t threshold = current_time + (m_wallet->use_fork_rules(2, 0) ? CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V2 : CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V1);
+        const uint64_t adjusted_time = m_wallet->get_daemon_adjusted_time();
+        uint64_t threshold = adjusted_time + (m_wallet->use_fork_rules(2, 0) ? CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V2 : CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V1);
```
UkoeHB, line 10160: Behavior change

src/wallet/wallet2.cpp
```diff
@@ -5932,7 +5932,7 @@ uint64_t wallet2::balance(uint32_t index_major, bool strict) const
   return amount;
 }
 //----------------------------------------------------------------------------------------------------
-uint64_t wallet2::unlocked_balance(uint32_t index_major, bool strict, uint64_t *blocks_to_unlock, uint64_t *time_to_unlock) const
+uint64_t wallet2::unlocked_balance(uint32_t index_major, bool strict, uint64_t *blocks_to_unlock, uint64_t *time_to_unlock)
```
UkoeHB, line 5935: Why was const removed from all of these functions?

src/wallet/wallet2.cpp
```diff
@@ -6225,12 +6225,14 @@ bool wallet2::is_tx_spendtime_unlocked(uint64_t unlock_time, uint64_t block_heig
...
-    uint64_t current_time = static_cast<uint64_t>(time(NULL));
+    uint64_t adjusted_time;
+    try { adjusted_time = get_daemon_adjusted_time(); }
+    catch(...) { adjusted_time = time(NULL); } // use local time if no daemon to report blockchain time
     // XXX: this needs to be fast, so we'd need to get the starting heights
     // from the daemon to be correct once voting kicks in
     uint64_t v2height = m_nettype == TESTNET ? 624634 : m_nettype == STAGENET ? 32000  : 1009827;
     uint64_t leeway = block_height < v2height ? CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V1 : CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V2;
-    if(current_time + leeway >= unlock_time)
+    if(adjusted_time + leeway >= unlock_time)
```
UkoeHB, line 6235: Behavior change

selsta:
Is the unbound submodule update intended?

src/cryptonote_core/blockchain.cpp
```diff
@@ -3682,7 +3684,7 @@ bool Blockchain::is_tx_spendtime_unlocked(uint64_t unlock_time) const
   else
   {
     //interpret as time
-    uint64_t current_time = static_cast<uint64_t>(time(NULL));
+    const uint64_t current_time = hf_version >= 13 ? get_adjusted_time(m_db->height()) : time(NULL);
```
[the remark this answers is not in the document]
moneromooo-monero: Can't recall. It seems pointless anyway.
The whole point is to start using blockchain timestamps instead of local time. This is intended. A 30 minute delay is fine, since this is typically used to lock things long term, not for a few minutes I think ? Define broken. Adding an offset would need changing some tests IIRC, and means you might unlck too early instead of too late. Why do you think this may be needed ?

src/cryptonote_core/blockchain.h
moneromooo-monero: To make it public.

moneromooo-monero: I guess I could add public: and private: and keep it where it was though. Seemed a bit meh but it minimizes the diff...

src/simplewallet/simplewallet.cpp
moneromooo-monero: Looks like it was already broken. I'm not sure why it's adding that define. The threshold really should be adjusted_time. If you can't think of why the define gets added either, I'll remove.

src/cryptonote_core/blockchain.cpp
moneromooo-monero: It looks like it's still used.

src/simplewallet/simplewallet.cpp
moneromooo-monero: Looks like the same as the one above. I think I might be missing your point though, since the point of the patch is to change the behaviour. So can you expand on whether you think something else (than the addition) is wrong ?

src/wallet/wallet2.cpp
moneromooo-monero: Because they now call RPC.

moneromooo-monero:
No. Not sure how it ended up here, thanks.

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
UkoeHB: "A 30 minute delay is fine, since this is typically used to lock things long term, not for a few minutes I think ?" Presumably if someone is locking a transaction with UTC time, they are trying to be more precise than a block height lock could provide. As the PR stands, users will have to wait an extra 60mins on average for their outputs to be unlocked (relative to the unlock time).

It sounds like you are concerned about future times being used for timelock checks. What about correcting the offset by doing a minimum of the current block's block time and the median block's time + 60mins? If blocks are accelerating then the current block's time will be more accurate than the median + 60, and if the current block has a future time then the median + 60 should be closer to the real time.

Note that even without the offset someone who can manipulate the median could push it into the future up to +1hr (since block times are limited by `local_time + 2hr` according to `check_block_timestamp()`. Adding the corrected offset allows timelocked tx to be spent up to 2hrs early in the worst case, while being much closer to +/- 10mins in the typical case.

src/simplewallet/simplewallet.cpp
UkoeHB: Not sure what you are referring to. Define?

src/simplewallet/simplewallet.cpp
UkoeHB: Sorry, just pointing out that every use of 'adjusted_time' is really '60 mins ago on average'. Since they are all replacing 'current time' the behavior changes.

src/cryptonote_core/blockchain.h
UkoeHB: Ah if it's being made public that makes sense.

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
TheCharlatan: The concern goes in both directions, faster and slower blocks, meaning both block timestamps into the future and the past. I'd argue a window of +-60 minutes is the time precission that we can provide and that it is good enough. If you lock a transaction's outputs with unix time exactly a year into the future, you get a guaranteed precision of +-60 minutes, while if you use blocks, the precision is unknown and could vary days.

Having the ability to push the median significantly would probably be at the limit of our security boundaries, since it is miners that are pushing it. I have not done any statistics on this though.

src/simplewallet/simplewallet.cpp
TheCharlatan: I think we should rename it to `locktime_median_time_past` or something along those lines to avoid confusion.

moneromooo-monero:
Ah, I found the behaviour change, a place is changing this without checking for v13. Will fix.

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
UkoeHB: Sorry, rather than 'current block' for corrected offset it should be 'previous block + 2mins' since the current block may be under construction and we don't want to use local time.

Well, +/- 60mins might be ok, but I'm wondering why not push for even more precision if it's possible?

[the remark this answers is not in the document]
moneromooo-monero: I don't really mind either way. Let's see if other people have an opinion and I'll go with what seems the majority opinion.

src/simplewallet/simplewallet.cpp
moneromooo-monero: By define, I meant adding CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_SECONDS_V2 etc. I don't see what purpose this has.

UkoeHB: Is this a Chesterton's fence? It does feel like an idea that didn't get fully fleshed out.

moneromooo-monero: On second thought, it doesn't hurt anything either. So as you imply, safer to keep it in at least for now.

moneromooo-monero: The name seems clear to me. It's getting the adjusted time from the daemon, not the current time.

src/cryptonote_core/blockchain.cpp
```diff
@@ -3752,11 +3755,28 @@ bool Blockchain::check_tx_input(size_t tx_version, const txin_to_key& txin, cons
 }
 //------------------------------------------------------------------
 //TODO: Is this intended to do something else?  Need to look into the todo there.
```
TheCharlatan, line 3757: Can this TODO be removed now?

dEBRUYNE-1:
Fyi, has conflicts with ` src/cryptonote_config.h` as far as I can see.

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
xiphon: Agree with @UkoeHB, IMO it would be okay to return `the median time + 60mins` (i.e. `+ BLOCKCHAIN_TIMESTAMP_CHECK_WINDOW / 2 * DIFFICULTY_TARGET_V2` seconds).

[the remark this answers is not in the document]
TheCharlatan: I'm fine with either; leaving it up to the tx creator to add an extra 60 minutes, or adding it straight into the calculation. Support seems to be in favour of adding the extra 60 minutes, so lets add it in then. It does seem more intuitive for the tx creator. 
@xiphon : What about koe's proposal to take the minimum between the last block timestamp and the median time? I have thought about it a bit and did not come up with anything that is obviously wrong or broken yet.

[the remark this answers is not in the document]
xiphon: @TheCharlatan

> doing a minimum of the current block's block time and the median block's time + 60mins
`The current block's block time` can't be less than `the median block's time`. Taking the minimum won't have any effect.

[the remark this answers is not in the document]
TheCharlatan: It does have an effect if you compare it to the median block time + 60 minutes (which is what koe is proposing).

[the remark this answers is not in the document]
xiphon: Yep, i misread this. Anyway `the current block's block time` can be manipulated as well, so IMO taking minimum doesn't fix the issue.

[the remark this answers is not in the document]
TheCharlatan: I feel similarly, so would suggest to go with just adding the extra 60 minutes to the `get_adjusted_time` calculation.

moneromooo-monero:
Now with the expected extra time. I'll rebase later since that'll pollute the diff massively.

moneromooo-monero:
I do not have the space to sync another chain here, it would be nice if someone could try a full sync to make sure this does not break historical blocks :)

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
UkoeHB: Taking the minimum is meant to provide better accuracy when blocks are accelerated (appearing faster than 2mins on average). If we don't take the minimum then even if all miners are honest the 'adjusted time' will be 'in the future' and locked transactions can be unlocked early (by probably only a few minutes, but still it will be a common misalignment).

Of course, the converse will be just as common, where blocks are slowing down and the adjusted time is 'in the past'. However, it is safe for locked transaction to be unlocked late. We can think about the unlock time as 'locked until _at least_ the lock time', which is harmonious with the reality of blockchains where tx authors have to wait for miners to add txs to a block (lock time of '0' means 'as soon as possible').

What about manipulation? There are a few combinations ('adjusted time' is 'median + 60', 'current time' is block timestamp, 'real time' is e.g. UTC):
1. Current time honest
a. Median low: adjusted time < current time = real time; tx unlocked up to *chain age* late or until median improves
b. Median high: adjusted time > current time = real time; [minimum: tx unlocked on time] [no minimum: tx unlocked up to 1hr early since current time >= median time is a protocol rule]
c. Median accurate: no problem, median + 60 ~= current time, ignoring local time variances
2. Current time high (capped at ~local time + 2hr by consensus)
a. Median low: adjusted time < real time < current time; tx unlocked up to *chain age* late or until median improves; current time inaccuracy has no effect
b. Median high: adjusted time & current time > real time; tx unlocked early [minimum: tx unlocked closest to real time possible] [no minimum: tx unlocked based on adjusted time, which may be earlier than current time]
c. Median accurate: adjusted time ~= real time < current time; no problem, tx unlocked on time
3. Current time low
a. Median low: adjusted time & current time < real time; [minimum: tx unlocked as late as possible] [no minimum: tx unlocked late]
b. Median high: current time < real time < adjusted time; [minimum: tx unlocked up to 1hr late] [no minimum: tx unlocked up to 1hr early]
c. Median accurate: current time < adjusted time ~= real time; [minimum: tx unlocked up to 1hr late] [no minimum: tx unlocked on time]

Using a minimum is a safety improvement for (1b), (2b), (3b), and an inconvenience loss for (3a), (3c). In net, an improvement to the adjusted time's safety profile.

moneromooo-monero:
I'm wondering whether this change can lock an ouput after it was unlocked, if the adjusted time goes back in time (ie, a new block has an earlier ts than the one 60 block ago). That could be annoying.

TheCharlatan:
If the last block is re-orged then sure, this can happen. However this is even true if you take just the median block times without additional modification. As long as the time used for the validation is relative to whatever the state is when the transaction is first included in a block, a change of the time into the past in a later block does not matter.

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
TheCharlatan: I spent a bit of time going through some of the potential problems with koe's proposal and think it is safe enough to go with it. It does introduce some extra complexity though and detecting a few of the corner cases has been challenging. The proposal does allow a miner to block inclusion of a time locked transaction into the next block by setting a low timestamp of his own mined block. In the current protocol, there are no incentives or rewards gained by doing this. Given this added complexity I would like to get a bit more review here.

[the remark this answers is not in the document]
UkoeHB: To get the median I think it should be like this:
```
// project the median to match approximately when the block being validated will appear
// the median is calculated from a chunk of past blocks, so we use +1 to offset onto the current block
median_ts += (BLOCKCHAIN_TIMESTAMP_CHECK_WINDOW + 1) * DIFFICULTY_TARGET_V2 / 2;

// project the current block's time based on the previous block's time
// we don't use the current block's time directly to mitigate timestamp manipulation
uint64_t adjusted_current_block_ts = m_db->get_block_timestamp(height - 1) + DIFFICULTY_TARGET_V2;

// return minimum of ~current block time and adjusted median time
// we do this since it's better to report a time in the past than a time in the future
return (previous_block_ts < median_ts ? previous_block_ts : median_ts);
```
Note: on the odd looking +1 in `median_ts`... imagine:
1. window = 2, blocks are indexed b0, b1, b2; block b2 is being validated; median of window is at timestamp (b0 + b1)/2 = 0.5b; adding offset of window/2 means adjusted timestamp would be 1.5b so we need to add another 0.5b -> 0.5b + (window + 1)/2 = 2b
2. window = 3, blocks are indexed b0, b1, b2, b3; block b3 is being validated; median of window is at timestamp 1b; adding offset of window/2 means adjusted timestamp would be 2.5b so we need to add another 0.5b -> 1b + (window + 1)/2 = 3b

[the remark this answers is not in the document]
moneromooo-monero: I assume adjusted_current_block_ts and previous_block_ts are the same variable here, right ?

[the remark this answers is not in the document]
UkoeHB: Yeah fixed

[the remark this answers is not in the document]
moneromooo-monero: Pushed with that change.

src/cryptonote_core/blockchain.h
```diff
@@ -1038,6 +1038,17 @@ namespace cryptonote
...
+    /**
+     * @brief get the "adjusted time"
+     *
+     * Returns the median timestamp of the past 60 blocks, if enough blocks are
+     * available. Returns the current time on the machine, if less than 60
+     * blocks are available.
+     *
+     * @return median time
+     */
+    uint64_t get_adjusted_time(uint64_t height) const;
```
UkoeHB, line 1050: This comment is out of date, otherwise everything looks good.
```
     /**
     * @brief get the "adjusted time"
     *
     * Computes the median timestamp of the previous 60 blocks, projects it
     * onto the current block to get an 'adjusted median time' which approximates
     * what the current block's timestamp should be. Also projects the previous
     * block's timestamp to estimate the current block's timestamp.
     * 
     * Returns the minimum of the two projections, or the current time on the
     * machine if less than 60 blocks are available.
     *
     * @return current time approximated from chain data
     */
```

UkoeHB (approved):

selsta (approved):

src/cryptonote_core/blockchain.cpp
```diff
@@ -3840,12 +3843,38 @@ bool Blockchain::check_tx_input(size_t tx_version, const txin_to_key& txin, cons
...
+  }
+  uint64_t median_ts = epee::misc_utils::median(timestamps);
+
+  // project the median to match approximately when the block being validated will appear
+  // the median is calculated from a chunk of past blocks, so we use +1 to offset onto the current block
+  median_ts += (BLOCKCHAIN_TIMESTAMP_CHECK_WINDOW + 1) * DIFFICULTY_TARGET_V2 / 2;
+
+  // project the current block's time based on the previous block's time
+  // we don't use the current block's time directly to mitigate timestamp manipulation
+  uint64_t adjusted_current_block_ts = m_db->get_block_timestamp(height - 1) + DIFFICULTY_TARGET_V2;
```
xiphon, line 3873:
```suggestion
  uint64_t adjusted_current_block_ts = timestamps.back() + DIFFICULTY_TARGET_V2;
```

selsta (approved):

UkoeHB (approved):

PR #6747, opened 2020-08-05
cmake: Use job pool feature to limit concurrent jobs

00-matt:
Add two new options, MONERO_PARALLEL_COMPILE_JOBS and MONERO_PARALLEL_LINK_JOBS to try and prevent running out of memory when building everything.

Requires >= cmake 3.0.0, and the use of the Ninja generator.

If the values are unset, or if Ninja is not used, it behaves as before.

##### Useful links

* https://cmake.org/cmake/help/latest/prop_gbl/JOB_POOLS.html
* https://reviews.llvm.org/D6304

##### Example

```bash
cmake -Bbuild -GNinja -DCMAKE_BUILD_TYPE=Release -DMONERO_PARALLEL_LINK_JOBS=1
cd build
ninja
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,20 @@ else()
 	message(STATUS "ccache deselected")
 endif()
 
+if (${CMAKE_VERSION} VERSION_GREATER "3.0.0" AND CMAKE_MAKE_PROGRAM MATCHES "ninja")
+  set(MONERO_PARALLEL_COMPILE_JOBS "" CACHE STRING "The maximum number of concurrent compilation jobs.")
+  if (MONERO_PARALLEL_COMPILE_JOBS)
+    set_property(GLOBAL APPEND PROPERTY JOB_POOLS compile_job_pool=${MONERO_PARALLEL_COMPILE_JOBS})
+    set(CMAKE_JOB_POOL_COMPILE compile_job_pool)
+  endif ()
+
+  set(MONERO_PARALLEL_LINK_JOBS "" CACHE STRING "The maximum number of concurrent link jobs.")
+  if (MONERO_PARALLEL_LINK_JOBS)
+    set_property(GLOBAL APPEND PROPERTY JOB_POOLS link_job_pool=${MONERO_PARALLEL_LINK_JOBS})
+    set(CMAKE_JOB_POOL_LINK link_job_pool)
+  endif ()
+endif()
+
 enable_language(C ASM)
 
 function (die msg)


selsta:
Sounds quite niche, I would guess almost no one uses ninja to build monero?

00-matt:
It's pretty good, noticeably faster during development.

Also maybe Ninja be used in Gitian builds? It would (hopefully) fix the out of memory issues people get when building with high core count CPUs.

hyc:
No, the fix for running out of memory in gitian builds is just to use a lower number of build threads, period.
You generally need 2GB RAM per instance, and some files require 3GB or more.

I would reject this patch.

moneromooo-monero:
It seems small and self contained (I've not read it yet though).

jagerman:
Why is this better than `ninja -j8` (or whatever value you want)?

00-matt:
@jagerman `ninja -jN` sets the limit on all jobs.

You could set `MONERO_PARALLEL_LINK_JOBS` to 1 and leave everything else default so that you have multiple low memory (except in the case of some files like wallet2) compile jobs making the most of your CPU, but only one link job to prevent running out of memory.

hyc:
Sounds like what you're really looking for is a feature to prevent make from spawning new jobs if free RAM is below a certain threshold. While there's no feature built in to make to handle this, you could easily fake it by writing a wrapper program that checks memory stats before passing its arguments to a real shell. Then set SHELL=<memory wrapper> in the Makefiles, and forget about it.

jagerman:
It looks like Ninja has at least one PR to do more or less what @hyc suggested - https://github.com/ninja-build/ninja/pull/1354

CMakeLists.txt
```diff
@@ -55,6 +55,21 @@ else()
 	message(STATUS "ccache deselected")
 endif()
 
+# Job pool feature requires Ninja.
+if (${CMAKE_VERSION} VERSION_GREATER "3.0.0")
```
selsta, line 59:
```suggestion
if (${CMAKE_VERSION} VERSION_GREATER "3.0.0" AND CMAKE_MAKE_PROGRAM MATCHES "ninja")
```

00-matt:
Not sure if there is much point to this, it already does nothing if you
are not using Ninja.

selsta:
@00-matt But without it cmake will incorrectly display MONERO_PARALLEL_COMPILE_JOBS / MONERO_PARALLEL_LINK_JOBS when doing `cmake -LAH ..`, correct behaviour would be to only display it when doing `cmake -LAH -G Ninja ..`

00-matt:
Ah didn't think of that, I guess it would show up in the ccmake GUI too. Updated.

selsta (approved):

PR #6757, opened 2020-08-12
threadpool bulletproofing

moneromooo-monero:


hyc:
The 2nd commit (cea3ef4) is undone by the 3rd commit (ad2efb4) - are you just going to squash all 3 into 1?

Looks OK aside from that.

moneromooo-monero:
It wasn't intended, I'll fix. I could squash I guess.

moneromooo-monero:
Oh no, that's because it was pointless with the last change. I'll squash them all.

src/cryptonote_core/blockchain.cpp
```diff
@@ -4845,7 +4844,7 @@ bool Blockchain::prepare_handle_incoming_blocks(const std::vector<block_complete
     {
       m_blocks_longhash_table.clear();
       uint64_t thread_height = height;
-      tools::threadpool::waiter waiter;
+      tools::threadpool::waiter waiter(tpool);
```
vtnerd, line 4847: There doesn't seem to be an effective method to see if this section failed via exception.

src/wallet/wallet2.cpp
```diff
@@ -2009,7 +2009,7 @@ void wallet2::process_new_transaction(const crypto::hash &txid, const cryptonote
           tpool.submit(&waiter, boost::bind(&wallet2::check_acc_out_precomp_once, this, std::cref(tx.vout[i]), std::cref(derivation), std::cref(additional_derivations), i,
             std::cref(is_out_data_ptr), std::ref(tx_scan_info[i]), std::ref(output_found[i])), true);
         }
-        waiter.wait(&tpool);
```
vtnerd, line 2012: The section above seems to lack indication of exception failure.

src/wallet/wallet2.cpp
```diff
@@ -2036,7 +2036,7 @@ void wallet2::process_new_transaction(const crypto::hash &txid, const cryptonote
         tpool.submit(&waiter, boost::bind(&wallet2::check_acc_out_precomp_once, this, std::cref(tx.vout[i]), std::cref(derivation), std::cref(additional_derivations), i,
             std::cref(is_out_data_ptr), std::ref(tx_scan_info[i]), std::ref(output_found[i])), true);
       }
-      waiter.wait(&tpool);
```
vtnerd, line 2039: Here too.

src/wallet/wallet2.cpp
```diff
@@ -2701,7 +2701,7 @@ void wallet2::process_parsed_blocks(uint64_t start_height, const std::vector<cry
     }
   }
   THROW_WALLET_EXCEPTION_IF(txidx != num_txes, error::wallet_internal_error, "txidx does not match tx_cache_data size");
-  waiter.wait(&tpool);
```
vtnerd, line 2704: Not seeing any flags to indicate exception failure here either.

src/wallet/wallet2.cpp
```diff
@@ -2730,7 +2730,7 @@ void wallet2::process_parsed_blocks(uint64_t start_height, const std::vector<cry
         gender(iod);
     }, true);
   }
-  waiter.wait(&tpool);
```
vtnerd, line 2733: Same here.

src/wallet/wallet2.cpp
```diff
@@ -2850,14 +2850,14 @@ void wallet2::pull_and_parse_next_blocks(uint64_t start_height, uint64_t &blocks
...
     tools::threadpool& tpool = tools::threadpool::getInstance();
-    tools::threadpool::waiter waiter;
+    tools::threadpool::waiter waiter(tpool);
     parsed_blocks.resize(blocks.size());
     for (size_t i = 0; i < blocks.size(); ++i)
     {
       tpool.submit(&waiter, boost::bind(&wallet2::parse_block_round, this, std::cref(blocks[i].block),
         std::ref(parsed_blocks[i].block), std::ref(parsed_blocks[i].hash), std::ref(parsed_blocks[i].error)), true);
     }
-    waiter.wait(&tpool);
```
vtnerd, line 2860: This section is tricky - if an exception is thrown `parsed_blocks[i].error` may remain false, which would be incorrect.

src/wallet/wallet2.cpp
```diff
@@ -2883,7 +2883,7 @@ void wallet2::pull_and_parse_next_blocks(uint64_t start_height, uint64_t &blocks
         }, true);
       }
     }
-    waiter.wait(&tpool);
+    waiter.wait();
```
vtnerd, line 2886: Don't see immediately how exceptions are reported here.

src/wallet/wallet2.cpp
```diff
@@ -2778,7 +2778,7 @@ void wallet2::process_parsed_blocks(uint64_t start_height, const std::vector<cry
     }
   }
   THROW_WALLET_EXCEPTION_IF(txidx != tx_cache_data.size(), error::wallet_internal_error, "txidx did not reach expected value");
-  waiter.wait(&tpool);
+  waiter.wait();
```
vtnerd, line 2781: `geniod` throws on some errors, which now get memory holed.

src/common/threadpool.h
```diff
@@ -55,12 +55,13 @@ class threadpool
...
     boost::condition_variable cv;
+    threadpool &pool;
     int num;
     public:
     void inc();
     void dec();
-    void wait(threadpool *tpool);  //! Wait for a set of tasks to finish.
-    waiter() : num(0){}
+    void wait();  //! Wait for a set of tasks to finish.
+    waiter(threadpool &pool) : pool(pool), num(0){}
```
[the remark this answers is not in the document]
moneromooo-monero: Because it allows the wait on dtor to use the correct pool (using NULL as before would hang).

moneromooo-monero:
If the code running some functions in parallel wants to do anything on failure, then it should do so in the functions themselves. The waiter has no idea about the semantics, it just runs code it's given. OK ?

moneromooo-monero:
https://github.com/moneromooo-monero/bitmonero/commit/8de024522ce163a3d0720890c2a991d4fdfe5cba

Would that address your point ?

hyc:
Seems kludgy to me, also doesn't tell you anything about how many or which threads hit an error.

IMO catching errors is only meaningful if you have enough info to recover from the error. If there are no recovery steps, then tough luck, forget about it. What meaningful steps can a caller do when a bunch of its threaded tasks fail?

moneromooo-monero:
Most of these are from verification. If anything fails when verifying, it's typically safe to just reject the block/tx or whatever you were verifying. Sure beats accessing uninitialized memory or hanging the process.

moneromooo-monero:
That sounds useful. Reminds me I used these exception forfwarding in the wallet refresh, I had forgot about it.
Anyway, are you OK with the last path I linked to ? It's not in this PR yet, I'll add it if you agree with it.

vtnerd:
I somehow commented on that patch 2 days ago. And the last two sentences above were in reference to that patch (I thought it was included in the PR).

moneromooo-monero:
Thanks, I'd have missed the comment on the patch if you hadn't mentioned it :)
I'll push this squashed and with no except added.

selsta (approved):

PR #6758, opened 2020-08-13
Create apparmor profiles for monerod and monero-wallet-cli

### LINKED ISSUE #6705: Create and maintain apparmor profiles for Monero binaries
Create and maintain apparmor profiles for the Monero binaries.  
  
Start with the most popular binaries: monerod, monero-wallet-cli and monero-wallet-gui.  
  
Suggest to start with a basic blacklist approach. Block access to locations like ~/.ssh/ and other obvious ones. Take a look at apparmor.d/abstractions/private-files and private-files-strict. Experiment a while to get a good idea about which locations and permissions to use and gradually increase the restrictions.  
  
Motivation: defense in depth.

00-matt:
Closes #6705

Draft. Looking for feedback, and testing for use with hardware wallets like the Ledger and Trezor.

To install and use:

```bash
# as root
cp -r contrib/apparmor/* /etc/apparmor.d/
aa-complain monerod monero-wallet-cli # or aa-enforce
tail -f /var/log/syslog # and start daemon/wallet in another window and watch for audit events
```

### DIFF
--- a/contrib/apparmor/usr.bin.monerod
+++ b/contrib/apparmor/usr.bin.monerod
@@ -1,0 +1,76 @@
+#----------------------------------------------------------------------------------------
+# Copyright (c) 2020, The Monero Project
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are
+# permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of
+#    conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this list
+#    of conditions and the following disclaimer in the documentation and/or other
+#    materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may be
+#    used to endorse or promote products derived from this software without specific
+#    prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#----------------------------------------------------------------------------------------
+# vim: ft=apparmor
+
+abi <abi/3.0>,
+
+include <tunables/global>
+include <tunables/monero>
+
+@{MONEROD_DATA_DIRS}=@{MONEROD_DATA_DIR}/{,stagenet/,testnet/,fake/}
+
+# Add `flags=(attach_disconnected)` before the opening curly bracket if you
+# have audit log messages like "Failed name lookup - disconnected path". This
+# reduces the effectiveness of AppArmor.
+#
+# See: https://lists.ubuntu.com/archives/apparmor/2018-July/011718.html
+# See: https://github.com/monero-project/monero/pull/6758#issuecomment-699275306
+profile monerod /usr/{,local/}bin/monerod {
+  include <abstractions/base>
+  include <abstractions/nameservice>
+  include <abstractions/openssl>
+
+  capability sys_nice,
+
+  /etc/inputrc r,
+  /etc/terminfo/** r,
+
+  /sys/devices/**/rotational r,
+
+  /etc/tor/torsocks.conf r,
+
+  # Allow another apparmored program to start monerod.
+  /usr/{,local/}bin/monerod mr,
+
+  /etc/monerod.conf r,
+
+  owner @{MONEROD_DATA_DIRS}/ rw,
+  owner @{MONEROD_DATA_DIRS}/lmdb/ w,
+  owner @{MONEROD_DATA_DIRS}/lmdb/data.mdb rwk,
+  owner @{MONEROD_DATA_DIRS}/lmdb/lock.mdb rwk,
+  owner @{MONEROD_DATA_DIRS}/ssl/* r,
+  owner @{MONEROD_DATA_DIRS}/bitmonero.conf r,
+  owner @{MONEROD_DATA_DIRS}/bitmonero.log* w,
+  owner @{MONEROD_DATA_DIRS}/p2pstate.bin rw,
+  owner @{MONEROD_DATA_DIRS}/monerod.pid w,
+
+  # Site-specific additions and overrides. See local/README for details.
+  #include if exists <local/usr.bin.monerod>
+}
--- a/contrib/apparmor/tunables/monero
+++ b/contrib/apparmor/tunables/monero
@@ -1,0 +1,54 @@
+#----------------------------------------------------------------------------------------
+# Copyright (c) 2020, The Monero Project
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are
+# permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of
+#    conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this list
+#    of conditions and the following disclaimer in the documentation and/or other
+#    materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may be
+#    used to endorse or promote products derived from this software without specific
+#    prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#----------------------------------------------------------------------------------------
+# vim:ft=apparmor
+
+# @{MONEROD_DATA_DIR} is a list of directories where the Monero daemon is allowed to
+# store its blockchain data, configuration files, and logs.
+#
+# Change this value if you launch monerod with a parameter like --data-dir.
+
+@{MONEROD_DATA_DIR}=@{HOME}/.bitmonero/ /var/lib/monero/
+
+# @{MONERO_WALLET_DIR} is a list of directories where the Monero wallet is allowed to
+# store its private keys, wallet cache, and log files.
+#
+# You should start monero-wallet-cli from this directory, and store your wallet files
+# here.
+#
+# Change this value if you don't like it.
+
+@{MONERO_WALLET_DIR}=@{HOME}/{,Documents/,Desktop/}{m,M}onero/
+
+# @{MONERO_WALLET_RINGDB_DIR} is a list of directories where the Monero wallet is
+# allowed to store its shared ring database.
+#
+# Change this value if you launch monero-wallet-cli with --shared-ringdb-dir.
+
+@{MONERO_WALLET_RINGDB_DIR}=@{HOME}/.shared-ringdb/
--- a/contrib/apparmor/usr.bin.monero-wallet-cli
+++ b/contrib/apparmor/usr.bin.monero-wallet-cli
@@ -1,0 +1,60 @@
+#----------------------------------------------------------------------------------------
+# Copyright (c) 2020, The Monero Project
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are
+# permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of
+#    conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this list
+#    of conditions and the following disclaimer in the documentation and/or other
+#    materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may be
+#    used to endorse or promote products derived from this software without specific
+#    prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#----------------------------------------------------------------------------------------
+# vim: ft=apparmor
+
+abi <abi/3.0>,
+
+include <tunables/global>
+include <tunables/monero>
+
+@{MONERO_WALLET_RINGDB_DIRS}=@{MONERO_WALLET_RINGDB_DIR}/{,stagenet/,testnet/}
+
+profile monero-wallet-cli /usr/{,local/}bin/monero-wallet-cli {
+  include <abstractions/base>
+  include <abstractions/nameservice>
+  include <abstractions/openssl>
+
+  /etc/inputrc r,
+  /etc/terminfo/** r,
+
+  /etc/tor/torsocks.conf r,
+
+  owner @{MONERO_WALLET_RINGDB_DIRS}/ w,
+  owner @{MONERO_WALLET_RINGDB_DIRS}/data.mdb rwk,
+  owner @{MONERO_WALLET_RINGDB_DIRS}/lock.mdb rwk,
+
+  owner @{MONERO_WALLET_DIR}/** rwk,
+
+  # Allow config files to be read, but not modified.
+  audit deny ${MONERO_WALLET_DIR}/**.conf wklx,
+
+  # Site-specific additions and overrides. See local/README for details.
+  #include if exists <local/usr.bin.monero-wallet-cli>
+}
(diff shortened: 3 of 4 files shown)


contrib/apparmor/usr.bin.monerod
```diff
@@ -0,0 +1,61 @@
...
+
+  /sys/devices/**/rotational r,
+
+  @{MONEROD_DATA_DIRS}/ w,
+  @{MONEROD_DATA_DIRS}/lmdb/ w,
+  @{MONEROD_DATA_DIRS}/lmdb/data.mdb rwk,
+  @{MONEROD_DATA_DIRS}/lmdb/lock.mdb rwk,
+  @{MONEROD_DATA_DIRS}/p2pstate.bin rw,
+  @{MONEROD_DATA_DIRS}/bitmonero.conf r,
+  @{MONEROD_DATA_DIRS}/bitmonero.log w,
```
moneromooo-monero, line 57: There will also be bitmonero.log-*, but they're never written to, just renamed from bitmonero.log. That might be handled by the above already ?

contrib/apparmor/usr.bin.monerod
```diff
@@ -0,0 +1,61 @@
...
+
+  /etc/inputrc r,
+  /etc/terminfo/** r,
+
+  /sys/devices/**/rotational r,
+
+  @{MONEROD_DATA_DIRS}/ w,
+  @{MONEROD_DATA_DIRS}/lmdb/ w,
+  @{MONEROD_DATA_DIRS}/lmdb/data.mdb rwk,
+  @{MONEROD_DATA_DIRS}/lmdb/lock.mdb rwk,
```
moneromooo-monero, line 54: For my curiosity, what is k ?

contrib/apparmor/usr.bin.monerod
00-matt: Ah I forgot about the log rotation, didn't run into that during testing. Will amend

contrib/apparmor/usr.bin.monerod
00-matt: It is for locking. Named k because l is already used for links

contrib/apparmor/usr.bin.monerod
00-matt: Never mind, it's already covered. It has permission to rename/move since w is enabled for the directory.

garlicgambit:
@00-matt thanks for the PR.  
  
A couple suggestions for monerod and monero-wallet-cli profiles:  
  
- Add 'owner' to @{MONEROD_DATA_DIRS}, @{MONERO_WALLET_RINGDB_DIRS} and @{MONERO_WALLET_DIR} entries  
- Add read access to /etc/tor/torsocks.conf to make the binaries work with torsocks  
  
For monerod only:  
  
- Add read/write access by 'owner' to /var/run/monerod.pid and /run/monerod.pid for pid file option.  
  
Will do more testing and report back in a couple of days.

garlicgambit:
Question: What is the motivation to include 'capability sys_nice' for monerod? Is it for mining?  
  
Couple more suggestions:  
  
General best practice would be to put as much monerod files in the ~/.bitmonero/ directory. This would simplify the apparmor profile. This change would replace /var/run/monerod.pid and /run/monerod.pid rules with @{MONEROD_DATA_DIRS}/monerod.pid.  
  
Monerod bitmonero.log log rotate requires the following changes:  
  
- Add 'flags=(attach_disconnected)' to monerod profile  
- Add read access to '@{MONEROD_DATA_DIRS}/ w,' Example: 'owner @{MONEROD_DATA_DIRS}/ rw,'  
- Add read access to '@{MONEROD_DATA_DIRS}/bitmonero.log w,' and append a '*' to the end. Example: 'owner @{MONEROD_DATA_DIRS}/bitmonero.log* rw,'  
  
Monerod RPC ssl files suggestions:  
  
- Use a special directory to store the ssl files  
- Suggest to call this directory: ssl  
- Give read access to owner  
- Example: 'owner @{MONEROD_DATA_DIRS}/ssl/* r,'  
  
Monerod regtest option:  
  
- It needs access to @{MONEROD_DATA_DIR}/fake/  
- Append it to '@{MONEROD_DATA_DIRS}=@{MONEROD_DATA_DIR}/ ...@{MONEROD_DATA_DIR}/testnet/'  
  
Will do more testing and report back soon.

00-matt:
garlicgambit writes:
> Question: What is the motivation to include 'capability sys_nice' for
> monerod? Is it for mining?

Hmm, not sure where that came from. After a grep I couldn't find
anywhere in the Monero codebase that calls nice. Will remove.

> General best practice would be to put as much monerod files in the
> ~/.bitmonero/ directory. This would simplify the apparmor
> profile. This change would replace /var/run/monerod.pid and
> /run/monerod.pid rules with @{MONEROD_DATA_DIRS}/monerod.pid.

Sure I think that's a good idea too. Anyone that wants to use /var/run/
can just add it to their local file or maybe change a variable.

> Monerod bitmonero.log log rotate requires the following changes:
>   
> - Add 'flags=(attach_disconnected)' to monerod profile

Can you elaborate on what this does? I read the manual a couple of times
and still do not understand :(

> - Add read access to '@{MONEROD_DATA_DIRS}/ w,' Example: 'owner @{MONEROD_DATA_DIRS}/ rw,'  
> - Add read access to '@{MONEROD_DATA_DIRS}/bitmonero.log w,' and append a '*' to the end. Example: 'owner @{MONEROD_DATA_DIRS}/bitmonero.log* rw,'  

Thanks, I didn't manage to test the log rotate myself yet. Will add.

00-matt:
Updated.

garlicgambit:
Also not 100% sure what the `attach_disconnected` flag is doing. It seems necessary for loading log files. But it could be specific to our setup. It might work without it. Can you and/or other people check if the `attach_disconnected` flag is required for log rotation? We'll do more testing ourselves as well.  
  
Suggestions for monero-wallet-cli:  
  
- Make the wallet directories case insensitive. Allow both 'Monero' and 'monero'. Don't want to bother users with whether something is written with a capital or not. This will save troubleshooting requests.  
- Consider adding `@{HOME}/Desktop/monero` and `@{HOME}/Desktop/Monero` to the list of wallet directories. This is a flexibility/convenience change.  
- Allow the creation of subdirectories. This is useful if you have a lot of wallets. This is a flexibility/convenience change. Example code: `owner @{MONERO_WALLET_DIR}/** rwk,`  
  
monero-wallet-cli will need the same changes as monerod for log rotation.  
  
Will do more testing and report back soon.

garlicgambit:
Suggestions for monero-wallet-cli:  
  
- Make access to the monero config file read only. You can use a config file with: `--config-file`. Example code: `audit deny @{MONERO_WALLET_DIR}/monero.conf wklx,`. This will override the allow permissions. Users will need to use `monero.conf` as the config file to benefit from this. An alternative would be to make all `.conf` files read only. Example code: `audit deny @{MONERO_WALLET_DIR}/**.conf wklx,`. We prefer the latter, unless there is a good reason against it.  
  
Also tested the monero-wallet-cli profile with monero-wallet-rpc. Initial tests have revealed no issues. The profile seems to work fine with monero-wallet-rpc. You only need to replace `monero-wallet-cli` with `monero-wallet-rpc` on the `profile` line.

iamamyth:
> Also not 100% sure what the attach_disconnected flag is doing

The documentation on the topic isn't terribly well-written, but, here's a decent explanation: https://lists.ubuntu.com/archives/apparmor/2018-July/011718.html.

garlicgambit:
@iamamyth, thanks for the link.  
  
Suggestion for the monerod profile:  
  
- Add read and mmap to the monerod binary. This is required when another apparmored program wants to start monerod. One example of this would be the monero-wallet-gui. Example code: `/usr/{,local/}bin/monerod mr,`.

garlicgambit:
Apparmor will generate `operation="file_inherit"` messages when monerod is launched from the monero-wallet-gui. It will ask for the following permissions:  
  
- monerod requests append access to: `~/.bitmonero/monero-wallet-gui.log`  
- monerod requests read/write access to: `~/.shared-ringdb/data.mdb`   
  
Does monerod need these permissions? Or can they be denied? We prefer to restrict access for monerod as much as possible.  
  
We've noticed this behavior on Tails. Have not reproduced it on other distros yet.

moneromooo-monero:
The ringdb is a wallet only thing.

garlicgambit:
@moneromooo-monero, thanks.  
  
It is likely that monerod won't need access to monero-wallet-gui.log either. Recommend to keep both blocked.  
  
Will do more testing and report back soon.

00-matt:
garlicgambit writes:
> Also tested the monero-wallet-cli profile with
> monero-wallet-rpc. Initial tests have revealed no issues. The profile
> seems to work fine with monero-wallet-rpc. You only need to replace
> monero-wallet-cli with monero-wallet-rpc on the profile line.

Do you think it should be added into the same profile file? e.g. the
first line contains /usr/{,local/}bin/monero-wallet-{cli,rpc}? Or is it
worth duplicating the file for now.

garlicgambit:
@00-matt, we see the following issues with this: what if you want to set different permissions for one of the binaries? And what if you want to disable/debug apparmor for one of the binaries?  
  
We prefer separate apparmor files, unless there is a solution for this.

00-matt:
Yes that makes sense. I'll copy the cli wallet profile in the next push.

garlicgambit:
The repo ships with the following templates: `utils/systemd/monerod.service` and `utils/conf/monero.conf`. They are mentioned in the README.md.  
  
Add the directories and files listed in those templates to the monerod apparmor profile? Any reasons against it?  
  
We favor to add them.

00-matt:
garlicgambit writes:
> The repo ships with the following templates:
> `utils/systemd/monerod.service` and `utils/conf/monero.conf`. They are
> mentioned in the README.md.
>   
> Add the directories and files listed in those templates to the monerod
> apparmor profile? Any reasons against it?
>   
> We favor to add them.

I don't see any reason why the systemd unit should be added, and
monero.conf is already included in the monerod profile. By default the
daemon expects the config file to be named bitmonero.conf, but the one
in the repository is named monero.conf.

garlicgambit:
@00-matt log rotation will work without the `flags=(attach_disconnected)` option. There may be some exceptions where it is needed. We've found that live distros like Tails need the option. Recommend to not use `flags=(attach_disconnected)` in the (template) profiles. Users/distros can add the option if they need it.  
  
About the systemd monerod.service and monerod.conf (not monero.conf, typo by us) files: the user will not be able to use these files without modifications. For example: the monerod.service file requires the config file to be at: `/etc/monerod.conf` and the monerod.conf file requires the data dir to be at: `/var/lib/monero`. Our preference would be to add the required permissions for both files to the monerod apparmor profile. It will make it easier to configure systemd in combination with apparmor.

00-matt:
garlicgambit writes:
> @00-matt log rotation will work without the
> `flags=(attach_disconnected)` option. There may be some exceptions
> where it is needed. We've found that live distros like Tails need the
> option. Recommend to not use `flags=(attach_disconnected)` in the
> (template) profiles. Users/distros can add the option if they need it.

I needed the option on Gentoo Linux with kernel 5.8.9 and AppArmor
2.13.4 but not on Debian Stable. Haven't been able to figure out what
exactly requires it but it is not logrotate. Maybe best to leave it in?

> About the systemd monerod.service and monerod.conf (not monero.conf,
> typo by us) files: the user will not be able to use these files
> without modifications. For example: the monerod.service file requires
> the config file to be at: `/etc/monerod.conf` and the monerod.conf
> file requires the data dir to be at: `/var/lib/monero`. Our preference
> would be to add the required permissions for both files to the monerod
> apparmor profile. It will make it easier to configure systemd in
> combination with apparmor.

Ah sorry, I misread your previous message. Yes it seems reasonable to
add those.

iamamyth:
Per the earlier thread I linked, `attach_disconnected` has nothing to do with log rotation; whether it's needed depends on the process context the daemon inherits, so it will generally be needed if the daemon inherits stdout/stderr (or any other used resource) from another process via `exec`.

iamamyth:
As to the broader question of whether to add `attach_disconnected` or not, to a degree, it completely undermines the purpose of apparmor, as it allows a process to access an inherited resource outside of its own sandbox. Of course, such inheritance is incredibly common, especially in container environments, so one could argue the apparmor model has serious deficits. In short, you're trading making the profile actually work in many environments for possibly escaping the security sandbox, a tradeoff I've seen many projects resolve by adding `attach_disconnected`, which, more than anything, points to a flaw in apparmor.

iamamyth:
I think https://github.com/monero-project/monero/pull/6808 may eliminate the need for `attach_disconnected` when using systemd. @00-matt When you wrote, "I needed the option on Gentoo Linux with kernel 5.8.9 and AppArmor 2.13.4", what was the exact scenario under test? Systemd or just starting monerod via the cli, and with what arguments?

00-matt:
> what was the exact scenario under test? Systemd or just starting monerod via the cli, and with what arguments?

It happened with OpenRC calling start-stop-daemon to start monerod with `--non-interactive`. Although even with the profile set to enforcing and no `attach_disconnected` flag, the daemon still seemed to sync and work okay while the audit log got filled up with errors, so maybe it's best for the flag to be removed.

iamamyth:
I think the sane default would be to omit `attach_disconnected`, as doing so eliminates the risk of accessing resources not vetted by the security policy. As you note, most of the time, `attach_disconnected` simply eliminates log spam, as it typically impacts writing to stdout or stderr, which the daemon and wallet programs don't require in order to operate as usable services. If someone wants to pass external resources to the daemon, thereby creating a risk, it's incumbent upon them to address that risk. But, I would suggest perhaps adding a comment to the daemon profile noting that some situations (esp. container deployments) may require `attach_disconnected` and linking to the relevant mailing list discussion \(https://lists.ubuntu.com/archives/apparmor/2018-July/011718.html\).

00-matt:
Thank you for your explanations. I have added a comment to the monerod profile.

garlicgambit:
Agreed with motivations by iamamyth to not enable `attach_disconnected` by default.  
  
The apparmor devs mention that they are planning to create a secure alternative for attach_disconnected on the mailing list discussion. Has there been any progress with this? Couldn't find any updates. This may be relevant for setups that require `attach_disconnected`.  
  
Is there any progress with adding hardware wallet support to the apparmor profiles? Is there someone with access to Trezor and/or Ledger devices?

00-matt:
Sorry for the delayed reply.

> Is there any progress with adding hardware wallet support to the apparmor profiles? Is there someone with access to Trezor and/or Ledger devices?

I have a Ledger Nano S that I plan on using for testing. Would be nice if someone else had a Trezor :)

garlicgambit:
@00-matt any progress with the Ledger?  
  
We've [opened an issue](https://gitlab.com/apparmor/apparmor/-/issues/125) on the apparmor repo about the state of the secure alternatives for attach_disconnected. In short: it will probably take a while for them to land. There is however a patch that may mitigate some of the problems with disconnected paths.

00-matt:
> any progress with the Ledger?

Yes, here are all of the extra files `monero-wallet-cli` tries to open when I use `--restore-from-device`:

```
"/dev/bus/usb"
"/dev/bus/usb/001/006"
"/etc/udev/udev.conf"
"/run/udev/data/c189:0"
"/run/udev/data/c189:128"
"/run/udev/data/c189:256"
"/run/udev/data/c189:257"
"/run/udev/data/c189:384"
"/run/udev/data/c189:5"
"/run/udev/data/c189:512"
"/run/udev/data/c189:514"
"/run/udev/data/c189:515"
"/run/udev/data/c189:640"
"/run/udev/data/+usb:1-0:1.0"
"/run/udev/data/+usb:1-5:1.0"
"/run/udev/data/+usb:1-5:1.1"
"/run/udev/data/+usb:2-0:1.0"
"/run/udev/data/+usb:3-0:1.0"
"/run/udev/data/+usb:3-4:1.0"
"/run/udev/data/+usb:3-4:1.2"
"/run/udev/data/+usb:4-0:1.0"
"/run/udev/data/+usb:5-0:1.0"
"/run/udev/data/+usb:5-1:1.0"
"/run/udev/data/+usb:5-2:1.0"
"/run/udev/data/+usb:5-2:1.1"
"/run/udev/data/+usb:5-2:1.2"
"/run/udev/data/+usb:6-0:1.0"
"/sys/bus"
"/sys/bus/usb/devices"
"/sys/bus/usb/devices/1-5/bConfigurationValue"
"/sys/bus/usb/devices/1-5/busnum"
"/sys/bus/usb/devices/1-5/descriptors"
"/sys/bus/usb/devices/1-5/devnum"
"/sys/bus/usb/devices/1-5/speed"
"/sys/bus/usb/devices/3-4/bConfigurationValue"
"/sys/bus/usb/devices/3-4/busnum"
"/sys/bus/usb/devices/3-4/descriptors"
"/sys/bus/usb/devices/3-4/devnum"
"/sys/bus/usb/devices/3-4/speed"
"/sys/bus/usb/devices/5-1/bConfigurationValue"
"/sys/bus/usb/devices/5-1/busnum"
"/sys/bus/usb/devices/5-1/descriptors"
"/sys/bus/usb/devices/5-1/devnum"
"/sys/bus/usb/devices/5-1/speed"
"/sys/bus/usb/devices/5-2/bConfigurationValue"
"/sys/bus/usb/devices/5-2/busnum"
"/sys/bus/usb/devices/5-2/descriptors"
"/sys/bus/usb/devices/5-2/devnum"
"/sys/bus/usb/devices/5-2/speed"
"/sys/bus/usb/devices/usb1/bConfigurationValue"
"/sys/bus/usb/devices/usb1/busnum"
"/sys/bus/usb/devices/usb1/descriptors"
"/sys/bus/usb/devices/usb1/devnum"
"/sys/bus/usb/devices/usb1/speed"
"/sys/bus/usb/devices/usb2/bConfigurationValue"
"/sys/bus/usb/devices/usb2/busnum"
"/sys/bus/usb/devices/usb2/descriptors"
"/sys/bus/usb/devices/usb2/devnum"
"/sys/bus/usb/devices/usb2/speed"
"/sys/bus/usb/devices/usb3/bConfigurationValue"
"/sys/bus/usb/devices/usb3/busnum"
"/sys/bus/usb/devices/usb3/descriptors"
"/sys/bus/usb/devices/usb3/devnum"
"/sys/bus/usb/devices/usb3/speed"
"/sys/bus/usb/devices/usb4/bConfigurationValue"
"/sys/bus/usb/devices/usb4/busnum"
"/sys/bus/usb/devices/usb4/descriptors"
"/sys/bus/usb/devices/usb4/devnum"
"/sys/bus/usb/devices/usb4/speed"
"/sys/bus/usb/devices/usb5/bConfigurationValue"
"/sys/bus/usb/devices/usb5/busnum"
"/sys/bus/usb/devices/usb5/descriptors"
"/sys/bus/usb/devices/usb5/devnum"
"/sys/bus/usb/devices/usb5/speed"
"/sys/bus/usb/devices/usb6/bConfigurationValue"
"/sys/bus/usb/devices/usb6/busnum"
"/sys/bus/usb/devices/usb6/descriptors"
"/sys/bus/usb/devices/usb6/devnum"
"/sys/bus/usb/devices/usb6/speed"
"/sys/class"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.1/usb1/1-0:1.0/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.1/usb1/1-5/1-5:1.0/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.1/usb1/1-5/1-5:1.1/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.1/usb1/1-5/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.1/usb1/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.1/usb2/2-0:1.0/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.1/usb2/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.3/usb3/3-0:1.0/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.3/usb3/3-4/3-4:1.0/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.3/usb3/3-4/3-4:1.2/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.3/usb3/3-4/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.3/usb3/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.3/usb4/4-0:1.0/uevent"
"/sys/devices/pci0000:00/0000:00:01.2/0000:02:00.0/0000:03:08.0/0000:05:00.3/usb4/uevent"
"/sys/devices/pci0000:00/0000:00:08.1/0000:0a:00.3/usb5/5-0:1.0/uevent"
"/sys/devices/pci0000:00/0000:00:08.1/0000:0a:00.3/usb5/5-1/5-1:1.0/uevent"
"/sys/devices/pci0000:00/0000:00:08.1/0000:0a:00.3/usb5/5-1/uevent"
"/sys/devices/pci0000:00/0000:00:08.1/0000:0a:00.3/usb5/5-2/5-2:1.0/uevent"
"/sys/devices/pci0000:00/0000:00:08.1/0000:0a:00.3/usb5/5-2/5-2:1.1/uevent"
"/sys/devices/pci0000:00/0000:00:08.1/0000:0a:00.3/usb5/5-2/5-2:1.2/uevent"
"/sys/devices/pci0000:00/0000:00:08.1/0000:0a:00.3/usb5/5-2/uevent"
"/sys/devices/pci0000:00/0000:00:08.1/0000:0a:00.3/usb5/uevent"
"/sys/devices/pci0000:00/0000:00:08.1/0000:0a:00.3/usb6/6-0:1.0/uevent"
"/sys/devices/pci0000:00/0000:00:08.1/0000:0a:00.3/usb6/uevent"
```

I am not sure how system specific this is though.

garlicgambit:
@00-matt suggest to put the Ledger and Trezor related rules in separate `abstractions` files. These rules can be shared with other apparmored applications, like: web browsers (FIDO/U2F), SSH, GPG and other crypto currency wallets. It will also make it easier to enable and disable them.  
  
Ledger, Trezor and other hardware wallet manufacturers can become the upstream maintainers of the apparmor profiles for their products. They provide security products to clients that want to secure their data. Apparmor provides an extra layer of security for their customers. We'll add support for their products to Monero in the mean time. This work can help persuade them to create and/or maintain apparmor profiles for their own products.  
  
Changes for `usr.bin.monerod`:  
Remove `capability sys_nice,`  
  
Change `...bitmonero.log* w,` to `...bitmonero.log* rw,`  
  
Recent versions of monerod require a new read/write permission for a file in the `lmdb` directory. This may or may not be the best apparmor compatible regex for the file:  
`@{MONEROD_DATA_DIRS}/lmdb/[0-9a-f]*-[0-9a-f]*-[0-9a-f]*-[0-9a-f]* rw,`  
  
The monerod.conf and monerod.service (systemd) template files need the following permissions:  
`/var/log/monero/ rw,`  
`/var/log/monero/monero.log* rw,`  
`/run/monero/monerod.pid w,`

iamamyth:
`/run/monero/monerod.pid w,`

Not required since https://github.com/monero-project/monero/commit/d3e7c5a8ba1dc7acfbc106f758267840d02f3684#diff-a7e03b1e89d83971fe578522eec42c521a23e74463cef1205920b21e9c0c1cb8.

garlicgambit:
@iamamyth correct, used an older version of monerod.service.

garlicgambit:
@00-matt, please reopen this PR.

PR #6760, opened 2020-08-14
enforce claiming maximum coinbase amount

moneromooo-monero:
Claiming a slightly lesser amount does not yield the size gains
that were seen pre rct, so this closes a fingerprinting vector


Mitchellpkt:
Thanks! Always a good time to take legacy slack out of the protocol. ^_^

I think that theoretically 
` if (version < 2 || version >= HF_VERSION_EXACT_COINBASE)`
and
` if (version >= HF_VERSION_EXACT_COINBASE)`
would produce the same practical result upon sync, since the strict V1 criteria (`=`) satisfies the the looser (`<=`) criteria prior to HF_VERSION_EXACT_COINBASE

Not sure which is preferable in terms of code hygiene & clarity.

moneromooo-monero:
It would if it was <=.

src/cryptonote_core/blockchain.cpp
```diff
@@ -1393,7 +1393,7 @@ bool Blockchain::validate_miner_transaction(const block& b, size_t cumulative_bl
     return false;
   }
   // From hard fork 2, we allow a miner to claim less block reward than is allowed, in case a miner wants less dust
```
UkoeHB, line 1395: The comment is out of date. Could say
```
// From hard fork 2, we allow a miner to claim less block reward than is allowed, in case a miner wants less dust; after hard fork 6 dust became irrelevant due to RingCT, and in hard fork 13 we returned to requiring exact block rewards.
```

Mitchellpkt:
> It would if it was <=.

Sorry for ambiguity, I meant 

"... would produce the same practical result upon sync, since the strict V1 criterion (miner_claimed=max_allowed) satisfies the the looser (miner_claimed<=max_allowed) criterion prior to HF_VERSION_EXACT_COINBASE"

UkoeHB:
Is there any precedent of removing an old protocol rule that has no functional value? I like how the code documents the full history of protocol changes. On Thu, Aug 20, 2020 at 6:34 PM, Mitchell P. Krawiec-Thayer, Ph.D. <notifications@github.com> wrote:  

It would if it was <=.

Sorry for ambiguity, I meant
"... would produce the same practical result upon sync, since the strict V1 criterion (miner_claimed=max_allowed) satisfies the the looser (miner_claimed<=max_allowed) criterion prior to HF_VERSION_EXACT_COINBASE"

PR #6763, opened 2020-08-15
rpc: always send raw txes through P2P (don't use bootstrap daemon)

xiphon:


xiphon:
> Might be worth a line in the commit message saying why these are bypassed.

Done.

PR #6766, opened 2020-08-16
qrcodegen: fix compilation with old gcc, use modern cmake

selsta:


moneroexamples:
Thanks. Works on ubuntu 16.04.

PR #6767, opened 2020-08-16
Fix build with Boost 1.74

moneromooo-monero:
Thanks iDunk for testing


selsta:
Seems like this needs an ifdef or so? Older boost don’t have this header.

PR #6786, opened 2020-08-28
Fix protocol with Ledger's HW when using CLSAG

grydz:
- Update protocol version to allow MLSAG and CLSAG in Ledger Monero application
- Update minimal version required for Ledger Monero application
- Fix method to send scalar `z` in `clsag_sign()` which can't be encrypted in the Monero client because it hasn't been generated on the HW.


moneromooo-monero (approved):
For the record, I'm told in private message that the bounds checking removal is safe due to other code (I've not checked).

SarangNoether:
@grydz: Will any additional Monero codebase PRs be required for Ledger support?

PR #6798, opened 2020-09-03
Trezor: Enable CLSAG support for Trezor client

ph4r05:
- Enables CLSAG transaction support for Trezor
- Fixes Trezor tests, testing HF \in {12, 13}
- Requires https://github.com/trezor/trezor-firmware/pull/1246
- Testing: https://github.com/ph4r05/monero/releases/tag/v16.0.0.0-tests-u18.04-01
- @SarangNoether for notif

### DIFF
--- a/src/device_trezor/trezor/protocol.cpp
+++ b/src/device_trezor/trezor/protocol.cpp
@@ -561,11 +561,6 @@ namespace tx {
       assign_to_repeatable(tsx_data.mutable_minor_indices(), tx.subaddr_indices.begin(), tx.subaddr_indices.end());
     }
 
-    // TODO: use HF_VERSION_CLSAG after CLSAG is merged
-    if (tsx_data.hard_fork() >= 13){
-      throw exc::ProtocolException("CLSAG is not yet implemented");
-    }
-
     // Rsig decision
     auto rsig_data = tsx_data.mutable_rsig_data();
     m_ct.rsig_type = get_rsig_type(tx.rct_config, tx.splitted_dsts.size());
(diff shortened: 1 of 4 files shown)


src/device_trezor/trezor/protocol.cpp
```diff
@@ -1017,14 +1012,23 @@ namespace tx {
       }
     }
 
-    // CLSAG support comes here once it is merged to the Monero
-    m_ct.rv->p.MGs.reserve(m_ct.signatures.size());
-    for(size_t i = 0; i < m_ct.signatures.size(); ++i) {
-      rct::mgSig mg;
-      if (!cn_deserialize(m_ct.signatures[i], mg)) {
-        throw exc::ProtocolException("Cannot deserialize mg[i]");
+    m_ct.rv->p.CLSAGs.reserve(m_ct.signatures.size());
```
moneromooo-monero, line 1015: This is technically correct, though it looks like it should be inside the if below, and the reserve for MGs restored in the else.

tests/trezor/trezor_tests.cpp
```diff
@@ -1245,10 +1263,14 @@ void gen_trezor_base::set_hard_fork(uint8_t hf)
   m_top_hard_fork = hf;
   if (hf < 9){
     throw std::runtime_error("Minimal supported Hardfork is 9");
-  } else if (hf == 9){
+  } else if (hf <= 11){
     rct_config({rct::RangeProofPaddedBulletproof, 1});
-  } else {
+  } else if (hf == 12){
     rct_config({rct::RangeProofPaddedBulletproof, 2});
+  } else if (hf == 13){
```
moneromooo-monero, line 1270: There are HF_VERSION_* defines for those if needed.

moneromooo-monero (approved):
Just a couple minor things which don't prevent this from being merged.

tests/trezor/trezor_tests.cpp
ph4r05: fixed

src/device_trezor/trezor/protocol.cpp
ph4r05: Aah correct. I've fixed it. Thanks!

moneromooo-monero (approved):

PR #6808, opened 2020-09-08
systemd: Don't use Type=forking

ndorf:
Per the systemd.service(5) man page, "It is generally recommended to use Type=simple for long-running services whenever possible."

It's better than Type=forking because the service (monerod) will be a direct child of systemd, which can then directly monitor it with the wait() family of syscalls.

AFAIK, the only reason to fork is to disassociate the controlling terminal and login session from the process. This isn't needed under systemd which already handles all of that.

Without the --detach option, monerod will send log messages to stdout/err, in addition to its own log file. By default, these duplicate messages would end up in the journal (and/or syslog). The two `Standard*=null` directives fix that.

### DIFF
--- a/utils/systemd/monerod.service
+++ b/utils/systemd/monerod.service
@@ -6,16 +6,15 @@ After=network.target
 User=monero
 Group=monero
 WorkingDirectory=~
-RuntimeDirectory=monero
 StateDirectory=monero
 LogsDirectory=monero
 
 # Clearnet config
 #
-Type=forking
-PIDFile=/run/monero/monerod.pid
-ExecStart=/usr/bin/monerod --config-file /etc/monerod.conf \
-    --detach --pidfile /run/monero/monerod.pid
+Type=simple
+ExecStart=/usr/bin/monerod --config-file /etc/monerod.conf --non-interactive
+StandardOutput=null
+StandardError=null
 
 # Tor config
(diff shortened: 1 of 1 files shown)


00-matt:
Do we still need `RuntimeDirectory=monero`, since the pid file isn't being stored in there any more? Or do we put anything else in that directory?

ndorf:
It's not used in this configuration. The `StateDirectory` and `LogsDirectory` aren't (and weren't) used either, so it might make sense to get rid of all three.

The `RuntimeDirectory` could be useful for additional options not present in this configuration, such as `--zmq-pub`. Of course, the operator can always add it back in themselves, if they want to use it.

00-matt:
I think that `StateDirectory` and `LogsDirectory` are used (at least they are mentioned in `utils/conf/monero/monerod.conf`).

Maybe while we're here messing with the unit file, we could enable one of the newer options like `DynamicUser=yes`? That way the admin doesn't bother having to create a monero user.

ndorf:
Oops, you're right. Fixed.

Regarding `DynamicUser`, that seems intriguing but I think it should be done as a separate PR, as it would require more testing.

00-matt (approved):

xmrdsc (approved):

PR #6809, opened 2020-09-08
Fix "wihch" typo for command line argument description

reinaldoacosta:


selsta:
We only change the text inside src/wallet/wallet2.cpp, the other files get updated once we update the translations, not necessary inside this PR.

PR #6810, opened 2020-09-08
New add_aux_pow RPC to support merge mining

moneromooo-monero:
Note about possible conflict of interest: I made this patch for the benefit of another coin. However, I believe making it easier to merge mine is beneficial to monero due to the chance to get additional hash rate from (ideally) many merge mined chains.

### DIFF
--- a/src/crypto/hash-ops.h
+++ b/src/crypto/hash-ops.h
@@ -87,6 +87,10 @@ void hash_extra_jh(const void *data, size_t length, char *hash);
 void hash_extra_skein(const void *data, size_t length, char *hash);
 
 void tree_hash(const char (*hashes)[HASH_SIZE], size_t count, char *root_hash);
+bool tree_path(size_t count, size_t idx, uint32_t *path);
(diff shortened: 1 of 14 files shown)


sethsimmons:
Is this being added to Monero to simplify Tari merge-mining specifically?

Curious, as I haven't heard any other coin mention Monero MMing outside of them.

moneromooo-monero:
After the IRC chat, I want to make clear that this patch is not strictly needed for a coin to merge mine with monero: this patch just makes it (much) easier. If it is not merged, every coin wanting to merge mine with monero will still be able to do so, but will have to include some some software that manually alters the extra field in the coinbase tx to include the merkle root of the aux blocks (which is indeed what Tari originally did).

moneromooo-monero:
It has to be a bit more complicated to prevent a hash reuse attack. A nonce needs adding to generate fixed positions for each chain in the aux pow merkle tree. I'll change the patch soon to do this.

moneromooo-monero:
Bug fix, rebase, and get rid of a rider patch that got rejected earlier.

selsta:
CORE_RPC_VERSION_MINOR bump missing :)

selsta (approved):

PR #6820, opened 2020-09-14
rpc: allow an unused second parameter in submitblock

moneromooo-monero:
This allows an upstream merge mining proxy to use this parameter
and passing around to us (because we might be a proxy too, it
doesn't know). The proxies can use this parameter to store the
difficulty of a given PoW hash, so proxies can route the block
to whichever daemon will accept that difficulty without having
to calculate the hash, saving some work.

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -1793,7 +1793,7 @@ namespace cryptonote
       }
     }
     CHECK_CORE_READY();
-    if(req.size()!=1)
+    if(req.size()!=1 && req.size()!=2)
     {
       error_resp.code = CORE_RPC_ERROR_CODE_WRONG_PARAM;
       error_resp.message = "Wrong param";


iamamyth:
Can't the proxy just use headers to convey the difficulty? Polluting the RPC protocol for proxy compatibility seems very much backwards.

moneromooo-monero:
I suppose it could. I'd understand the comment if it was a not a tiny change, but here it's reaching :) 
Using headers would mean every proxy would have to add code for this, rather than use an existing mechanism. Seems... wrong.

moneromooo-monero:
Actually, each proxy woud not have to do it, it'd be optional since it's just for an optimization (though one every pool would want in practice).

moneromooo-monero:
Another doubting vote on IRC, closing then.

moneromooo-monero:
FWIW, I added the header based version, and it's much less code than I expected, so it seems like a good solution as any other coin can easily do it too AFAICT.

PR #6826, opened 2020-09-18
Update error messages in daemon

tmoravec:
When given a wrong argument, some daemon commands failed with "unknown
command" error, instead of a meaningful error message. This patch
brings consistency into the error messages.

In several places, this patch removes relatively useful messages,
and replaces them with more generic ones. E.g.,

    -    std::cout << "use: print_pl [white] [gray] [<limit>] [pruned] [publicrpc]" << std::endl;
    +    std::cout << "Invalid syntax: Too many parameters. For more details, use the help command." << std::endl;

There are two reasons for this:

1. Consistency.
2. Removing duplicates.

The detailed information about the parameters is present in
the help messages already. Having it in two places increases
the risk that the messages would get out of date.

### DIFF
--- a/src/daemon/command_parser_executor.cpp
+++ b/src/daemon/command_parser_executor.cpp
@@ -50,7 +50,7 @@ bool t_command_parser_executor::print_peer_list(const std::vector<std::string>& 
 {
   if (args.size() > 3)
   {
-    std::cout << "use: print_pl [white] [gray] [<limit>] [pruned] [publicrpc]" << std::endl;
+    std::cout << "Invalid syntax: Too many parameters. For more details, use the help command." << std::endl;
     return true;
   }
 
@@ -79,7 +79,7 @@ bool t_command_parser_executor::print_peer_list(const std::vector<std::string>& 
     }
     else if (!epee::string_tools::get_xtype_from_string(limit, args[i]))
     {
-      std::cout << "unexpected argument: " << args[i] << std::endl;
+      std::cout << "Invalid syntax: Unexpected parameter: " << args[i] << ". For more details, use the help command." << std::endl;
       return true;
     }
   }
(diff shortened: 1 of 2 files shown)


src/daemon/command_parser_executor.cpp
```diff
@@ -50,7 +50,7 @@ bool t_command_parser_executor::print_peer_list(const std::vector<std::string>&
 {
   if (args.size() > 3)
   {
-    std::cout << "use: print_pl [white] [gray] [<limit>] [pruned] [publicrpc]" << std::endl;
+    std::cout << "Invalid syntax: Too many parameters. For more details, use the help command." << std::endl;
```
moneromooo-monero, line 53: Actually displaying the output of "help print_pl" here seems better, what do you think? Or at least point to "help print_pl".

moneromooo-monero (commented):
I really like when you get the synopsis when you run the command without parameters, it's kinda canonical. Other than that, good patch, thanks.

src/daemon/command_parser_executor.cpp
tmoravec: As a user, I like that better, too. But I also like consistency, and error messages that actually reflect reality. The trouble here is that the output of "help print_pl" is not defined here - the string I'm removing was copy/pasted from `t_command_server` constructor. If print_pl changes in the future, there's some chance that one of these strings would not be updated.

So the right way to get the output here is to get the help strings from `t_command_server` constructor (or the other way around but that would be fighting against epee). Which is tricky, because the command executor gets constructed in the same constructor. Not impossible for sure, but it would require refactoring of the relationship between `t_command_server`, `t_command_parser_executor`, and `epee::console_handlers_binder m_command_lookup`.

Even pointing to "help print_pl" requires copypasta as the command names don't always match the method names in the executor. Pointing to the help command in general was a poor man's attempt in the same direction, without copy pasting strings between files.

So there are three options:

1. Leave it as it is.
2. Copy/paste the command names (low effort).
3. Get the real help strings from `t_command_server` (major refactoring).

Which one would be preferable in your opinion?

moneromooo-monero: If it's not easily accessible, then I guess we can kick the can down the road.

moneromooo-monero (approved):

PR #6836, opened 2020-09-22
functional_tests: rewrite mining code, clearer timeout info

selsta:
It only happens sometimes, not consistently to reproduce. I don’t know if setting it to 90 fixes it.

- https://github.com/monero-project/monero/pull/6835/checks?check_run_id=1146997271
- https://github.com/monero-project/monero/pull/6758/checks?check_run_id=1082850323

### DIFF
--- a/tests/functional_tests/mining.py
+++ b/tests/functional_tests/mining.py
@@ -95,20 +95,19 @@ class MiningTest():
         assert res_status.block_reward >= 600000000000
 
         # wait till we mined a few of them
+        target_height = prev_height + 5
+        height = prev_height
         timeout = 60 # randomx is slow to init
-        timeout_height = prev_height
-        while True:
-            time.sleep(1)
-            res_info = daemon.get_info()
-            height = res_info.height
-            if height >= prev_height + 5:
-                break
-            if height > timeout_height:
-              timeout = 5
-              timeout_height = height
+        while height < target_height:
+            seen_height = height
+            for _ in range(timeout):
+                time.sleep(1)
+                height = daemon.get_info().height
+                if height > seen_height:
+                    break
             else:
-              timeout -= 1
-            assert timeout >= 0
+                assert False, 'Failed to mine successor to block %d (initial block = %d)' % (seen_height, prev_height)
+            timeout = 5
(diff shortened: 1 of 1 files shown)


tests/functional_tests/mining.py
```diff
@@ -95,7 +95,7 @@ def mine(self, via_daemon):
         assert res_status.block_reward >= 600000000000
 
         # wait till we mined a few of them
-        timeout = 60 # randomx is slow to init
+        timeout = 90 # randomx is slow to init
```
iamamyth, line 98: The way this test is written makes it difficult to distinguish timeouts waiting for the initial block versus subsequent ones. It's possible (and I would say, likely) that the timeouts you see are actually for the subsequent blocks, and you should be changing `timeout = 5`, below, rather than this startup timeout. I also think the code would be much clearer (and easier to diagnose failures) if the init and subsequent block cases were separated; for example:
```
for _ in range(60):
    time.sleep(1)
    height = daemon.get_info().height
    if height > prev_height:
        break
else:
    assert False, 'Failed to mine initial block'

target_height = prev_height + 5
while height < target_height:
    for _ in range(5):
        time.sleep(1)
        height = daemon.get_info().height
        if height >= target_height:
            break
    else:
        assert False, 'Failed to mine successor to block %d' % (height,)
```

iamamyth: Alternately, you could keep the single loop and just refactor it a bit, which may be superior:
```
while height < target_height:
    timeout = 60 if height == prev_height else 5
    cur_height = height
    for _ in range(timeout):
        time.sleep(1)
        height = daemon.get_info().height
        if height > cur_height:
            break
    else:
        assert False, 'Failed to mine successor to block %d' % (height,)
```

selsta: Thank you, I committed your second suggestion.

moneromooo-monero (approved):

selsta:
@iamamyth Seems like it starts failing again: https://github.com/monero-project/monero/pull/6867/checks?check_run_id=1222708036

Any idea which timeout I should increase?

PR #6842, opened 2020-09-24
fix a couple bugs found by OSS-fuzz

moneromooo-monero:
- index out of bounds when importing outputs
- accessing invalid CLSAG data

### DIFF
--- a/src/ringct/rctTypes.h
+++ b/src/ringct/rctTypes.h
@@ -368,6 +368,12 @@ namespace rct {
         template<bool W, template <bool> class Archive>
         bool serialize_rctsig_prunable(Archive<W> &ar, uint8_t type, size_t inputs, size_t outputs, size_t mixin)
         {
+          if (inputs >= 0xffffffff)
+            return false;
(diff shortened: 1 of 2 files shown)


src/ringct/rctTypes.h
```diff
@@ -368,6 +368,12 @@ namespace rct {
         template<bool W, template <bool> class Archive>
         bool serialize_rctsig_prunable(Archive<W> &ar, uint8_t type, size_t inputs, size_t outputs, size_t mixin)
         {
+          if (inputs >= 0xffffffff)
```
xiphon, line 371: Could you elaborate? Is it an integer overflow bug? Didn't find the exact place where the bug gets triggered at a quick glance.

moneromooo-monero: It is. I can elaborate later.

xiphon (approved):

vtnerd:
`std::numeric_limits<uint32_t>::max()` or `UINT32_MAX` here ?

PR #6846, opened 2020-09-26
Add informational comments for recently added HF heights (mainnet)

dEBRUYNE-1:


src/hardforks/hardforks.cpp
```diff
@@ -68,7 +68,10 @@ const hardfork_t mainnet_hard_forks[] = {
   // version 12 starts from block 1978433, which is on or around the 30th of November, 2019. Fork time finalised on 2019-10-18.
   { 12, 1978433, 0, 1571419280 },
 
+  // version 13 starts from block 2210000, which is on or around the 17th of October, 2020. Fork time finalised on 2019-09-02.
```
TheCharlatan, line 71: The finalization date is wrong.

dEBRUYNE-1: I used the date of the pull request opening. Should I use the merge date?

TheCharlatan: The previous entries just explain the date of the unix timestamp, e.g. `1571419280` is `10/18/2019 @ 5:21pm` and `1550225678` is `02/15/2019 @ 10:14am`. So `1598180817` translates to `08-23-2020`, which is the date of the irc meeting when the fork time was finalized. Notice the 2019 as well, I'm sure this was not opened last year.

dEBRUYNE-1: Thanks for the clarification. The `2019` entry was my error. Will make sure the proper date is used.

moneromooo-monero: Those comments are just pointless anyway, it's not like they explain anything, so their only use is possibly making things more ambiguous (not that I'm objecting to the change either, I'm fine either way)

selsta:
Closing this as it contains the wrong date and lack of update.

PR #6849, opened 2020-09-27
rpc_private_headers declaration fix

malbit:
rpc_daemon was not declared whatsoever.

These includes should point to rpc instead.


moneromooo-monero (approved):

TheCharlatan:
The `rpc_private_headers` are not set either. I removed them and it still compiled fine. Are these used for another build target that I may be missing?

malbit:
> The `rpc_private_headers` are not set either. I removed them and it still compiled fine. Are these used for another build target that I may be missing?

Thats the point. It will compile within or without but {rpc_private_headers} are declared on linking.

PR #6862, opened 2020-10-06
Fix lgamma/signgam dependency

hyc:
This is a new indirect dependency due to the use of poisson_distribution,
introduced in PR#6354 commit 67ade8005. It causes a dependency on
glibc 2.23, while we previously only required 2.17.


TheCharlatan (approved):
Great stuff! I can confirm that after a gitian build no glibc symbols versioned 2.2x remain.

PR #6874, opened 2020-10-09
make Monero understand Tor seeds, and add one

moneromooo-monero:


src/p2p/net_node.inl
```diff
@@ -567,8 +569,16 @@ namespace nodetool
     {
       net::get_network_address_host_and_port(addr, host, port);
     }
-    MINFO("Resolving node address: host=" << host << ", port=" << port);
 
+    if (boost::ends_with(host, ".onion"))
+    {
+      epee::net_utils::network_address na{MONERO_UNWRAP(net::tor_address::make(host, (uint16_t)atoi(port.c_str())))}; // gets us 0 if not an integer, works for me
```
xiphon, line 575: `atoi` doesn't handle well "out of int range" case, the result is undefined.

src/p2p/net_node.inl
```diff
@@ -567,8 +569,16 @@ namespace nodetool
     {
       net::get_network_address_host_and_port(addr, host, port);
     }
-    MINFO("Resolving node address: host=" << host << ", port=" << port);
 
+    if (boost::ends_with(host, ".onion"))
```
vtnerd, line 573: This section is now `net::get_network_address` but without i2p support. Calling that function would require re-splitting host/port (that function does it internally too) if `addr` is a hostname, but this section should be called infrequently anyway.

This section should be easier if `get_seed_nodes()` returns a `std::vector<network_address>`, I'll look into this.

vtnerd (commented):
This should be inconsistent for Tor seed node connections. `connect_to_seed()` does not take a `zone` argument, so it only selects Tor seeds by "chance" when `connections_maker()` determines the IP address connection count is too low. I believe both of those functions need updating.

I would not close this PR so that the Tor address is linked to a commit+PR from you. I will update those functions in a separate PR (assuming you didn't want to) - I'm obligated via the CCS proposal anyway.

moneromooo-monero:
Thanks, I guess I'll readd the seed once you've done those changes.

selsta (approved):

moneromooo-monero:
Obsoleted by https://github.com/monero-project/monero/pull/7068

PR #6875, opened 2020-10-10
Dandelion++: skip desynced peers in stem phase

xiphon:
Monero node ignores incoming txes while syncing. Thus should avoid sending a tx to desynced peer in Dandelion++ stem phase..

Resolves https://github.com/monero-project/monero-gui/issues/3140, https://github.com/monero-project/monero/issues/6885

### DIFF
--- a/src/cryptonote_core/cryptonote_core.h
+++ b/src/cryptonote_core/cryptonote_core.h
@@ -663,7 +663,7 @@ namespace cryptonote
       *
       * @param target_blockchain_height the target height
       */
-     uint64_t get_target_blockchain_height() const;
+     virtual uint64_t get_target_blockchain_height() const override;
 
      /**
       * @brief returns the newest hardfork version known to the blockchain
--- a/src/cryptonote_core/i_core_events.h
+++ b/src/cryptonote_core/i_core_events.h
@@ -39,6 +39,7 @@ namespace cryptonote
     virtual ~i_core_events() noexcept
     {}
 
+    virtual uint64_t get_target_blockchain_height() const = 0;
     virtual void on_transactions_relayed(epee::span<const cryptonote::blobdata> tx_blobs, relay_method tx_relay) = 0;
   };
 }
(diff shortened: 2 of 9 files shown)


Gingeropolous:
Is this still gonna be susceptible to nefarious peers feeding bad information? E.g., they just tell you they're at the right height, so you send them the tx, and then they just sinkhole it? 

Nice find though. This could explain some txs just never making it out of the stem phase.

xiphon:
> Is this still gonna be susceptible to nefarious peers feeding bad information? E.g., they just tell you they're at the right height, so you send them the tx, and then they just sinkhole it?

Yep, but mitigating this is out of the current PR's scope.

Gingeropolous:
> Yep, but mitigating this is out of the current PR's scope.

Agreed. I'm just curious if you've had any thoughts on how to mitigate this.

vtnerd:
> Agreed. I'm just curious if you've had any thoughts on how to mitigate this.

You probably can't. And The situation is identical to a node that alters the codebase locally to drop all stem transactions - a known issue mentioned in the original Dandelion++ paper.

vtnerd (commented):
The stems are typically selected before the send - this change only triggers when the selected stem goes bad or if there weren't enough outgoing links in the initial attempt. The epoch timer has to clear out previously selected stems already, so I had the timer callback immediately select the next two stems when things are (usually) idle. This also worked well with the "white noise" code for i2p/tor, which shares the dandelion++ stem epoch code.

The two ways to fix that: (1) make the stem selection lazy, or (2) pass `cryptonote::core` to the constructor of the notify class. (2) helps with `dandelion_notify` which already needs that object anyway, and (1) requires changes to `send_noise` (white noise) and `start_epoch`.

xiphon:
> pass `cryptonote::core` to the constructor of the notify class

Updated

moneromooo-monero (approved):

vtnerd (approved):
@moneromooo-monero can you confirm whether `m_state` for the "context" used by `cryptonote_protocol.h` is just the local node's state wrt to the remote node? In other words, the `synchronizing` state means the local node is behind the remote node. It is unfortunate this isn't tracking the state of the remote node, but the inverse makes sense for the purposes of downloading new blocks to "catch up".

I _think_ the core object can safely be put in `detail::notify`, which reduces the copying/moving necessary for some of the callbacks. Minor though, something that can be tweaked later.

moneromooo-monero:
synchronizing means the local node is in the synchronizing state, which should mean the peer's height is higher, at least most of the time.  You don't know whether the peer itself is syncing, and you probably should not know that.

PR #6877, opened 2020-10-11
p2p: add --max-connections-per-ip daemon option

moneromooo-monero:
Helps daemons behind a proxy get more than one connection.
Defaults to 1 (no change)

### DIFF
--- a/src/p2p/net_node.cpp
+++ b/src/p2p/net_node.cpp
@@ -166,6 +166,7 @@ namespace nodetool
     const command_line::arg_descriptor<bool> arg_pad_transactions = {
       "pad-transactions", "Pad relayed transactions to help defend against traffic volume analysis", false
     };
+    const command_line::arg_descriptor<uint32_t> arg_max_connections_per_ip = {"max-connections-per-ip", "Maximum number of connections allowed from the same IP address", 1};
 
     boost::optional<std::vector<proxy>> get_proxies(boost::program_options::variables_map const& vm)
     {
--- a/src/p2p/net_node.h
+++ b/src/p2p/net_node.h
@@ -258,7 +258,8 @@ namespace nodetool
         m_igd(no_igd),
         m_offline(false),
         is_closing(false),
-        m_network_id()
+        m_network_id(),
+        max_connections(1)
     {}
     virtual ~node_server();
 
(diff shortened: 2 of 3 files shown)


stmax82:
I just thought about this parameter vs "--in-peers". Wouldn't a boolean flag "--allow-multiple-connections-per-ip" be a better fit? That way "--in-peers" can keep handling the total number of connections as it does now.

moneromooo-monero:
That would allow someone to easily DoS a node using this option, I think.

stmax82:
If I understand this right, `--max-connections-per-ip 100` would have the same effect as `--in-peers 100 --allow-multiple-connections-per-ip` when behind a proxy?

The second case with the boolean flag seems more consistent since it doesn't allow for conflicting combinations like `--in-peers 10 --max-connections-per-ip 100`.

moneromooo-monero:
--max-connections-per-ip 100 would not override --in-peers. If you're behind a proxy that makes all incoming IPs look the same, it'll be de facto doing this, though, yes. It doesn't gain anything changing it to boolean though, and makes it much worse when not being a proxy. So no. Unless I'm misunderstanding your point.

stmax82:
I see what you mean... the boolean flag `--allow-multiple...` without restricting `--in-peers` (which is infinite by default) would allow someone to DoS the node by opening many connections until it runs out of memory (or sockets / handles / whatever).

I guess that could be prevented by allowing `--allow-multiple...` only in conjuction with `--in-peers != -1`... or by only allowing multiple connections from private ip ranges. It's not worth the trouble though... `--max-conns-per-ip <int>` is fine 👍🏼

stmax82:
@moneromooo-monero any update on this? My node's still stuck with 1 incoming connection in docker... :/

stmax82:
It's actually been getting worse with one of the last updates (I guess due to the sybil attack) - I'm now getting lots of "Host 172.17.0.1 blocked." messages, preventing even a single incoming connection. I guess I'll just give up on incoming connections and keep the node running with outgoing connections only... feel free to dispose this PR if it causes any other problems.

moneromooo-monero:
Do you mean after setting max conections per ip to more than one, you still get only one incoming connection ?

stmax82:
> Do you mean after setting max conections per ip to more than one, you still get only one incoming connection ?

No - I'm running release 0.17.1.5 - I was wondering if this could get merged because my node's waiting for it :)

selsta (approved):

selsta (approved):

PR #6890, opened 2020-10-12
wallet2: skip reorgs exceeding max-reorg-depth wallet setting

xiphon:


src/wallet/wallet2.cpp
```diff
@@ -3465,6 +3466,18 @@ void wallet2::refresh(bool trusted_daemon, uint64_t start_height, uint64_t & blo
...
 
+      if (!next_blocks.empty() && !m_blockchain.is_in_bounds(next_blocks_start_height))
+      {
+        const uint64_t expected_start_height = std::max(blocks_start_height, uint64_t(1)) - 1;
+        const uint64_t reorg_depth = expected_start_height - std::min(expected_start_height, next_blocks_start_height);
+        if (reorg_depth > m_max_reorg_depth)
+        {
+          MERROR("Rejected " << reorg_depth << " blocks reorg (max allowed " << m_max_reorg_depth << ")"
+            << ", set max-reorg-depth " << reorg_depth << " if you want to allow it.");
+          throw std::runtime_error("reorg depth exceeds maximum allowed");
```
moneromooo-monero, line 3477: The option suggestion should probably be in the exception, walet users won't see logs.

src/wallet/wallet2.cpp
```diff
@@ -3465,6 +3466,18 @@ void wallet2::refresh(bool trusted_daemon, uint64_t start_height, uint64_t & blo
           throw std::runtime_error("proxy exception in refresh thread");
       }
 
+      if (!next_blocks.empty() && !m_blockchain.is_in_bounds(next_blocks_start_height))
```
moneromooo-monero, line 3469: A reorg of more than the threshold could still have the height be in the hash chain bounds. Since this can be used to defend against daemons trolling wallets, that check should probably be removed.

xiphon: Fixed

src/wallet/wallet2.cpp
xiphon: Updated, introduced new exception class for that

moneromooo-monero (approved):

xiphon:
Updated, fixed the build.

selsta (approved):

PR #6891, opened 2020-10-12
core_rpc_server: on_get_blocks - forward bootstrap daemon error

xiphon:


selsta (approved):

selsta (approved):

src/rpc/core_rpc_server.cpp
```diff
@@ -521,9 +521,15 @@ namespace cryptonote
   bool core_rpc_server::on_get_blocks(const COMMAND_RPC_GET_BLOCKS_FAST::request& req, COMMAND_RPC_GET_BLOCKS_FAST::response& res, const connection_context *ctx)
   {
     RPC_TRACKER(get_blocks);
-    bool r;
-    if (use_bootstrap_daemon_if_necessary<COMMAND_RPC_GET_BLOCKS_FAST>(invoke_http_mode::BIN, "/getblocks.bin", req, res, r))
-      return r;
+
+    {
+      boost::shared_lock<boost::shared_mutex> lock(m_bootstrap_daemon_mutex);
```
moneromooo-monero, line 526: That doesn't look like a recursive mutex (though some platforms might make it recursive) and use_bootstrap_daemon_if_necessary takes it too. That might lockup.

xiphon: Yep, fixed. Introduced a race condition instead, but i consider this one is acceptable.

selsta (approved):
tested

moneromooo-monero (approved):
Yes, the race  makes it do what it does now, so I agree it's acceptable

PR #6896, opened 2020-10-13
crypto: add a rw lock to rx_slow_hash for robustness

moneromooo-monero:
I have seen a case where the current locking is not enough to prevent
corruption when calculating hashes from both the current height seed
hash and other seed hashes. This happened in townforge, when syncing
historical merge mined blocks (hence the mix of current and arbitrary
seed hashes) but it seems to also apply to monero since one can ask
for pow calculations on arbitrary blocks while syncing, though I have
not actually tried to repro this on monero.

Thanks to hyc for discussion and help debugging.


moneromooo-monero:
It would be nice if someone with a tuned system for many threads could double check this doesn't substantially decrease H/s, just in case.

hyc (approved):

selsta:
Fails to build on Windows.

```
2020-10-13T16:11:17.9289149Z In file included from D:/a/monero/monero/src/crypto/rx-slow-hash.c:40:
2020-10-13T16:11:17.9293300Z D:/a/monero/monero/src/crypto/c_threads.h:51:26: error: 'PTHREAD_MUTEX_INITIALIZER' undeclared here (not in a function)
2020-10-13T16:11:18.0306805Z    51 | #define CTHR_RWLOCK_INIT PTHREAD_MUTEX_INITIALIZER
2020-10-13T16:11:18.0656178Z       |                          ^~~~~~~~~~~~~~~~~~~~~~~~~
2020-10-13T16:11:18.0668610Z D:/a/monero/monero/src/crypto/rx-slow-hash.c:52:34: note: in expansion of macro 'CTHR_RWLOCK_INIT'
2020-10-13T16:11:18.0672485Z    52 | static CTHR_RWLOCK_TYPE rwlock = CTHR_RWLOCK_INIT;
2020-10-13T16:11:18.0697913Z       |                                  ^~~~~~~~~~~~~~~~
```

hyc:
Looks like Windows should be using these: https://docs.microsoft.com/en-us/windows/win32/sync/slim-reader-writer--srw--locks

selsta:
Resolved in https://github.com/monero-project/monero/commit/dab7d01dc0357a926993e5205d02a063b6bdcda0

PR #6897, opened 2020-10-13
Add support for i2p and tor seed nodes

vtnerd:
This should allow for i2p and tor seed nodes internally within the codebase.

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -435,9 +435,9 @@ namespace nodetool
 
     if (command_line::has_arg(vm, arg_p2p_seed_node))
     {
-      boost::unique_lock<boost::shared_mutex> lock(m_seed_nodes_lock);
+      boost::unique_lock<boost::shared_mutex> lock(public_zone.m_seed_nodes_lock);
 
-      if (!parse_peers_and_add_to_container(vm, arg_p2p_seed_node, m_seed_nodes))
+      if (!parse_peers_and_add_to_container(vm, arg_p2p_seed_node, public_zone.m_seed_nodes))
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/p2p/net_node.inl
```diff
@@ -1620,18 +1640,16 @@ namespace nodetool
...
-    size_t start_conn_count = get_public_outgoing_connections_count();
-    if(!get_public_white_peers_count() && !connect_to_seed())
+    for(auto& zone : m_network_zones)
     {
-      return false;
-    }
+      size_t start_conn_count = get_outgoing_connections_count(zone.second);
+      if(!zone.second.m_peerlist.get_white_peers_count() && !connect_to_seed(zone.first))
+      {
+        return false;
```
moneromooo-monero, line 1648: Should that be continue instead ? (and maybe return false if all fail) ?

vtnerd: Yup, thought about this shortly after I posted. And I think a bit more tracking needs to be done to keep the status code accurate (if all seeds fail for all networks, `return false`).

moneromooo-monero (approved):

selsta:
I started with this PR and fresh p2pstate bin like this:

`./monerod --tx-proxy tor,127.0.0.1:9050 --tx-proxy i2p,127.0.0.1:8060`

And the daemon only connected Tor and I2P seed nodes, not normal clearnet ones. Restarting helped. Does this PR make sure that clearnet seed nodes are always used?

selsta:
Two out of two times with freshly deleted p2pstate it did not find any peers: https://www.irccloud.com/pastebin/uh6ML9OD/

selsta (approved):

PR #6905, opened 2020-10-18
blockchain: fix sync at v14 boundary

moneromooo-monero:
Miners with MLSAG txes which they'd already verified included
a couple in that block, but the consensus rules had changed
in the meantime, so that block is technically invalid and any
node which did not already have those two txes in their txpool
could not sync. Grandfather them in, since it has no effect in
practice.


TheCharlatan (approved):
Code looks good, have not tested it against my node.

rbrunner7 (approved):
I can confirm that with this applied syncing to top from below the fork height is again possible.

PR #6912, opened 2020-10-18
Allow setting start block on export [release-v0.17]

hyc:
And make import honor the starting block# recorded in a bootstrap file

### DIFF
--- a/src/blockchain_utilities/bootstrap_file.cpp
+++ b/src/blockchain_utilities/bootstrap_file.cpp
@@ -52,7 +52,7 @@ namespace
 
 
 
-bool BootstrapFile::open_writer(const boost::filesystem::path& file_path)
+bool BootstrapFile::open_writer(const boost::filesystem::path& file_path, uint64_t start_block, uint64_t stop_block)
 {
   const boost::filesystem::path dir_path = file_path.parent_path();
   if (!dir_path.empty())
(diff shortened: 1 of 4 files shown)


src/blockchain_utilities/bootstrap_file.cpp
```diff
@@ -368,11 +371,35 @@ uint64_t BootstrapFile::seek_to_first_chunk(std::ifstream& import_file, uint8_t
   MINFO("bootstrap magic size: " << sizeof(file_magic));
   MINFO("bootstrap header size: " << bfi.header_size);
 
+  uint32_t buflen_blocks_info;
+
+  import_file.read(buf1, sizeof(buflen_blocks_info));
+  str1.assign(buf1, sizeof(buflen_blocks_info));
+  if (! import_file)
+    throw std::runtime_error("Error reading expected number of bytes");
```
moneromooo-monero, line 379: That looks endian dependent. Is the file format already endian dependent ?

hyc: Yes, this is literally just a copy-paste of the existing code from line 351-on https://github.com/monero-project/monero/blob/6e594e53bb25f02d3b2572a33f353e15b7c05395/src/blockchain_utilities/bootstrap_file.cpp#L351

moneromooo-monero (approved):

src/blockchain_utilities/bootstrap_file.cpp
hyc: Shouldn't be endian dependent. Those bytes are read and then passed to ::serialization::parse_binary().

fluffypony (approved):
Reviewed

PR #6913, opened 2020-10-18
wallet2: wait for propagation timeout before marking a tx as failed

xiphon:


selsta (approved):

vtnerd (commented):
If the RPC is **not** restricted, then all transactions should be returned. Only in the restricted case, should there be issues with invisible Dandelion++ stem transactions.

xiphon:
> If the RPC is **not** restricted, then all transactions should be returned

Introducing this timeout is the correct way to handle pending/failed state

if a user will flush local daemon's tx pool right after sending a tx, the daemon will return an empty list, is is already broadcasted to the network.

xiphon:
The main case the fix addresses: in bootstrap mode daemon doesn't forward `send_raw_transaction` to bootstrap daemon now, but sends a tx through P2P instead (i made the change a while back to prevent linking txes to user's ip on bootstrap daemon side). As a side effect there is some time frame when the tx has already been broadcasted to the network through P2P (Dandelion++), but hasn't yet arrived at the bootstrap daemon's memory pool.

selsta:
It also applies to just using a restricted remote node. Currently when sending it instantly switches to failed due to Dandelion++

tobtoht (approved):

vtnerd:
> Introducing this timeout is the correct way to handle pending/failed state.

Well ... a check for restricted mode or synced daemon before doing the timeout still seems a little more accurate. Although if both of those conditions are met, the RPC should return a failure if pushing to local mempool failed.

> (i made the change a while back to prevent linking txes to user's ip on bootstrap daemon side)

I don't see how this helps. Nodes currently syncing do not relay transactions. Therefore, if a peer sees a syncing node broadcasting a transaction, then that node must be the origin. The only viable technique (before syncing) is likely over I2P/Tor, since the IP is hidden over that connection.

There's possibly some edge cases where the node doesn't report its synced status immediately to a peer, cause the logic to fail if the node is almost synced. There's also the case of nodes lying, etc. But in the typical case this is still a leak afaik.

selsta:
> The only viable technique (before syncing) is likely over I2P/Tor, since the IP is hidden over that connection.

FWIW I tested bootstrap daemon with Tor tx-proxy and it worked well without any code change necessary. If we bundle Tor with e.g. GUI this is something we can do to improve IP linkage

PR #6914, opened 2020-10-18
[release-v0.17] wallet2: wait for propagation timeout before marking tx as failed

xiphon:


src/wallet/wallet2.cpp
```diff
@@ -2988,9 +2990,11 @@ void wallet2::update_pool_state(std::vector<std::tuple<cryptonote::transaction,
         LOG_PRINT_L1("Pending txid " << txid << " not in pool, marking as not in pool");
         pit->second.m_state = wallet2::unconfirmed_transfer_details::pending_not_in_pool;
       }
-      else if (pit->second.m_state == wallet2::unconfirmed_transfer_details::pending_not_in_pool && refreshed)
+      else if (pit->second.m_state == wallet2::unconfirmed_transfer_details::pending_not_in_pool && refreshed &&
+        now > std::chrono::system_clock::from_time_t(pit->second.m_timestamp) + tx_propagation_timeout)
```
moneromooo-monero: m_sent_time is probably better here, though I dont' remember all the details... Both are time(NULL) at init time though.

moneromooo-monero (approved):

src/wallet/wallet2.cpp
xiphon: updated

selsta (approved):
Tested

fluffypony (approved):
Reviewed

PR #6922, opened 2020-10-20
Make Blockchain::get_fee_quantization_mask() compile time

SChernykh:
This also removes potential thread safety bug in that function.

### DIFF
--- a/src/common/powerof.h
+++ b/src/common/powerof.h
@@ -1,0 +1,26 @@
+#pragma once
+
+#include <stdint.h>
+
+namespace tools
+{
+  template<uint64_t a, uint64_t b>
+  struct PowerOf
+  {
+    enum Data : uint64_t
+    {
+      // a^b = a * a^(b-1)
+      Value = a * PowerOf<a, b - 1>::Value,
+    };
+  };
+
+  template<uint64_t a>
+  struct PowerOf<a, 0>
+  {
+    enum Data : uint64_t
+    {
+      // a^0 = 1
+      Value = 1,
+    };
+  };
+}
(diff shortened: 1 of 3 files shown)


TheCharlatan (approved):
Nice find

src/common/powerof.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <stdint.h>
+
+namespace tools
```
vtnerd, line 5: With C++11 `constexpr` this is just:
```c++
constexpr std::uint64_t powerof(std::uint64_t a, std::uint64_t b)
{
    return b == 0 ? 1 : a * powerof(a, b - 1);
}
```
which is computed at compile time with `-O1`, `-02` or `-Ofast` (basically any optimizations enabled). A `constexpr uint64_t`  could be placed in the cpp file which forces compile time calculation even with optimizations disabled. No static memory is used if the address is ignored. This works with at least Gcc 4.9.4 (maybe further back).

moneromooo-monero (approved):

src/common/powerof.h
SChernykh: I know about that, but I wanted to have a header-only code, plus it guarantees that no code/static storage will be generated in the binary regardless of compiler version and flags. For such simple function (a^b) it's a matter of taste, but yes - I'd use constexpr function for something more tricky.

vtnerd: `constexpr` is header-only ? And I doubt there is a difference in practice after compiler optimizations. The primary difference is the use of an `enum` in the inner class (presumably to prevent storage for an inner constant), which is still valid with `constexpr`:

```c++
constexpr std::uint64_t powerof(std::uint64_t a, std::uint64_t b)
{
  return b == 0 ? 1 : a * powerof(a, b - 1);
}

enum constants : uint64_t
{
  quantization_mask = powerof(10, CRYPTONOTE_DISPLAY_DECIMAL_POINT - PER_KB_FEE_QUANTIZATION_DECIMALS)
};
```

PR #6931, opened 2020-10-23
Add "address mnew" command for monero-wallet-cli

codesoap:
When I create a cold wallet, I want to generate a bunch of subaddresses in advance and store them within easy reach. This way I always have "fresh" subaddresses at hand, even when I don't have quick access to my secret keys.

It annoyed me, that I could only generate one subaddress at a time with `monero-wallet-cli`, so I thought a `address mnew <n>` command, which generates n addresses at once, would be useful. I chose the `mnew` name in the style of FTP's `put`/`mput` commands. I'm still very new to Monero, so please let me know if I reinvented a wheel here.

A downside to my implementation is, that the addresses cannot be labeled. I'm not sure if this is a big deal; I don't care about the labels, since I don't store the wallet anyway - I only store my mnemonic.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -143,6 +143,7 @@ typedef cryptonote::simple_wallet sw;
 #define CREDITS_TARGET 50000
 #define MAX_PAYMENT_DIFF 10000
 #define MIN_PAYMENT_RATE 0.01f // per hash
+#define MAX_MNEW_ADDRESSES 1000
 
 enum TransferType {
   Transfer,
@@ -203,7 +204,7 @@ namespace
                             "  account tag <tag_name> <account_index_1> [<account_index_2> ...]\n"
                             "  account untag <account_index_1> [<account_index_2> ...]\n"
                             "  account tag_description <tag_name> <description>");
-  const char* USAGE_ADDRESS("address [ new <label text with white spaces allowed> | all | <index_min> [<index_max>] | label <index> <label text with white spaces allowed> | device [<index>] | one-off <account> <subaddress>]");
+  const char* USAGE_ADDRESS("address [ new <label text with white spaces allowed> | mnew <amount of new addresses> | all | <index_min> [<index_max>] | label <index> <label text with white spaces allowed> | device [<index>] | one-off <account> <subaddress>]");
   const char* USAGE_INTEGRATED_ADDRESS("integrated_address [device] [<payment_id> | <address>]");
   const char* USAGE_ADDRESS_BOOK("address_book [(add (<address>|<integrated address>) [<description possibly with whitespaces>])|(delete <index>)]");
   const char* USAGE_SET_VARIABLE("set <option> [<value>]");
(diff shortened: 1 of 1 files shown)


moneromooo-monero (approved):

codesoap:
I have created my [own "cold wallet generator"](https://github.com/codesoap/iglu) by now, so I'm not too eager on getting this pull request merged anymore. I'll still keep it open in case this feature is considered useful by the maintainers.

selsta:
Please squash your commits.

codesoap:
> Please squash your commits.

I have removed the upstream-merge commit, that accidentally got into this PR. Hope this is enough.

selsta (approved):

src/simplewallet/simplewallet.cpp
```diff
@@ -9508,6 +9509,32 @@ bool simple_wallet::print_address(const std::vector<std::string> &args/* = std::
...
+    if (n > MAX_MNEW_ADDRESSES)
+    {
+      fail_msg_writer() << tr("the amount of new addresses must be lower or equal to ") << MAX_MNEW_ADDRESSES;
+      return true;
+    }
+    for (uint32_t i = 0; i < n; ++i)
+    {
+        m_wallet->add_subaddress(m_current_subaddress_account, tr("(Untitled address)"));
+        print_address_sub(m_wallet->get_num_subaddresses(m_current_subaddress_account) - 1);
+        m_wallet->device_show_address(m_current_subaddress_account, m_wallet->get_num_subaddresses(m_current_subaddress_account) - 1, boost::none);
```
selsta, line 9535: Do you know if this is correct? Isn’t that showing a subaddress on hardware wallet? Guess that would not work for 1000 subaddresses.

codesoap: I'm not sure if I understood the question fully. This is the same [code used for the `address new` command](https://github.com/codesoap/monero/blob/4c0ea2df00ccb29f8368c892bda0094dcf1a1547/src/simplewallet/simplewallet.cpp#L9510), so I would guess, that it has no more problems, than typing in `address new` a 1000 times.

codesoap: Oh and in case there is a misunderstanding: 1000 is just the **limit** for how many addresses can be generated in a single `address mnew` call. I would not expect people to generate this many subaddresses often.

selsta: If you generate an address they get displayed on the hardware wallet. This makes sense for one address, but quickly looping through up to 1000 addresses does not make sense. I would remove that line.

codesoap: I was not aware of this interaction of `monero-wallet-cli` and hardware wallets. Thanks for educating me. I will remove the line.

codesoap: The `address mnew` command still works as expected when I remove the line. I have force pushed the change.

selsta: ok thanks for testing

selsta (approved):

PR #6933, opened 2020-10-25
p2p: use /16 filtering on IPv4-within-IPv6 addresses

moneromooo-monero:
IPv6 addresses include a range that can map IPv4 addresses,
which allowed those mapped addresses to bypass filtering.

This filter should be replaced by AS filtering at some point.

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -1432,6 +1432,20 @@ namespace nodetool
             const uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
             classB.insert(actual_ip & 0x0000ffff);
           }
+#if BOOST_VERSION > 106600
+          else if (cntxt.m_remote_address.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
+          {
+            const epee::net_utils::network_address na = cntxt.m_remote_address;
+            const boost::asio::ip::address_v6 &actual_ip = na.as<const epee::net_utils::ipv6_network_address>().ip();
+            if (actual_ip.is_v4_mapped())
+            {
+              boost::asio::ip::address_v4 v4ip = make_address_v4(boost::asio::ip::v4_mapped, actual_ip);
+              uint32_t actual_ipv4;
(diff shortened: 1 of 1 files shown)


src/p2p/net_node.inl
```diff
@@ -1452,6 +1465,19 @@ namespace nodetool
...
           }
+#if BOOST_VERSION > 106600
+          else if (skip_duplicate_class_B && pe.adr.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
+          {
+            const epee::net_utils::network_address na = pe.adr;
+            boost::asio::ip::address_v6 actual_ip = na.as<const epee::net_utils::ipv6_network_address>().ip();
+            if (actual_ip.is_v4_mapped())
+            {
+              boost::asio::ip::address_v4 v4ip = make_address_v4(boost::asio::ip::v4_mapped, actual_ip);
+              uint32_t actual_ip4 = (((unsigned)v4ip.to_bytes()[1]) << 8) | (unsigned)v4ip.to_bytes()[0];
```
selsta, line 1476: Any reason you are casting `(unsigned)` here and not on line 1443? Does not matter I guess?

moneromooo-monero: In case it'd shift a uint8_t, the bits would then get dropped into the bit bucket.

vtnerd: @selsta was commenting on why the cast isn't being done above where the code is identical.

src/p2p/net_node.inl
```diff
@@ -1432,6 +1432,19 @@ namespace nodetool
             const uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
             classB.insert(actual_ip & 0x0000ffff);
           }
+#if BOOST_VERSION > 106600
+          else if (cntxt.m_remote_address.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
+          {
+            const epee::net_utils::network_address na = cntxt.m_remote_address;
+            boost::asio::ip::address_v6 actual_ip = na.as<const epee::net_utils::ipv6_network_address>().ip();
```
vtnerd, line 1439: `const &` ?

selsta (approved):

PR #6934, opened 2020-10-25
Compil time: Remove boost/optional from headers

mj-xmr:
Removed almost all occurrences of boost/optional from headers (where it was possible without introducing a new interface), thus limiting the header only to those files, which need it.

Notice, that for the class wipeable_string I had to create a new wrapper method: parse_hexstr_optional(), in order to encapsulate the optional.hpp, as the template hex_to_pod was the only place, where the optional was queried, so I thought it might be worth it.

According to this graph, where I temporarily replaced the inclusion of the original options header with my wrapper for better tracing, boost optional is now included only where it's needed, because all of its direct 1st level dependees use the optional as one of their class members:
image
while the below graph shows how many unnecessary inclusions of the original header there were, by showing the dependees of just the forward declaration header:
image

Speed improvements:
| Parsing (frontend):         1611.3 s |     Parsing (frontend):         1514.9 s   |  -5.98% | 
| Codegen & opts (backend):   1663.8 s | Codegen & opts (backend):  1571.9  | -5.52% |

Reduction rate of 6% overall. The affected files were not listed in the below report, because they weren't the largest bottlenecks.

[cba-result.txt](https://github.com/monero-project/monero/files/6369404/cba-result.txt)

Work time:
5.5h Active
0.61h Testing
0.4h verification (in parallel with testing time)
2.1h additional 2 verifications for the review

### DIFF
--- a/contrib/epee/include/wipeable_string.h
+++ b/contrib/epee/include/wipeable_string.h
@@ -28,7 +28,7 @@
 
 #pragma once
 
-#include <boost/optional/optional.hpp>
+#include "fwd/boost_monero_optional_fwd.h"
 #include <stddef.h>
 #include <vector>
 #include <string>
@@ -37,6 +37,7 @@
 
 namespace epee
 {
+  struct ws_optional;
   class wipeable_string
   {
   public:
@@ -78,6 +79,7 @@ namespace epee
 
   private:
     void grow(size_t sz, size_t reserved = 0);
+    bool parse_hexstr_optional(wipeable_string & str_to_init) const;
 
   private:
     std::vector<char> buffer;
@@ -88,12 +90,12 @@ namespace epee
     static_assert(std::is_pod<T>::value, "expected pod type");
     if (size() != sizeof(T) * 2)
       return false;
-    boost::optional<epee::wipeable_string> blob = parse_hexstr();
-    if (!blob)
+    epee::wipeable_string str;
+    if (! parse_hexstr_optional(str))
       return false;
-    if (blob->size() != sizeof(T))
+    if (str.size() != sizeof(T))
       return false;
-    pod = *(const T*)blob->data();
+    pod = *(const T*)str.data();
     return true;
   }
 }
--- a/src/3rd/boost_monero_optional.h
+++ b/src/3rd/boost_monero_optional.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2020, The Monero Project
+// Copyright (c) 2014-2020, The Monero Project
 // 
 // All rights reserved.
 // 
@@ -25,43 +25,14 @@
 // INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+// 
+// Parts of this file are originally copyright (c) 2012-2013 The Cryptonote developers
 
 #pragma once
 
-#include "fwd/boost_monero_optional_fwd.h"
-#include <cstdint>
-#include <string>
-#include <vector>
-#include "byte_slice.h"
-#include "crypto/hash.h"
+/*
+Only boost::optional wrapper for easier searches in Doxygen
+*/
 
-namespace cryptonote
-{
-class core;
+#include <boost/optional/optional.hpp>
 
-namespace rpc
-{
-
-struct output_distribution_data
-{
-  std::vector<std::uint64_t> distribution;
-  std::uint64_t start_height;
-  std::uint64_t base;
-};
-
-class RpcHandler
-{
-  public:
-    RpcHandler() { }
-    virtual ~RpcHandler() { }
-
-    virtual epee::byte_slice handle(std::string&& request) = 0;
-
-    static boost::optional<output_distribution_data>
-      get_output_distribution(const std::function<bool(uint64_t, uint64_t, uint64_t, uint64_t&, std::vector<uint64_t>&, uint64_t&)> &f, uint64_t amount, uint64_t from_height, uint64_t to_height, const std::function<crypto::hash(uint64_t)> &get_hash, bool cumulative, uint64_t blockchain_height);
-};
-
-
-}  // rpc
-
-}  // cryptonote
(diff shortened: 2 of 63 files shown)


src/device/device.hpp
```diff
@@ -34,6 +34,9 @@
 #include "ringct/rctTypes.h"
 #include "cryptonote_config.h"
 
+//#include "wipeable_string.h"
```
iamamyth, line 37: This appears to be a debugging remnant.

mj-xmr: Corrected. Thanks.

moneromooo-monero (approved):

contrib/epee/include/wipeable_string.h
```diff
@@ -88,12 +90,12 @@ namespace epee
     static_assert(std::is_pod<T>::value, "expected pod type");
     if (size() != sizeof(T) * 2)
       return false;
-    boost::optional<epee::wipeable_string> blob = parse_hexstr();
-    if (!blob)
+    epee::wipeable_string str;
+    if (! parse_hexstr_optional(str))
```
moneromooo-monero, line 94: The _optional suffix doesn't really belong here tbh, but that's really nitpicking, ok either way

mj-xmr: I prefer normally adding suffixes, rather than overloading methods, in order to prevent a possible stack overflow, in case somebody confuses the parameters and calls the same method by mistake from within that method. With a different name it's harder to confuse.
Unless you mean, that you'd change the suffix?

moneromooo-monero: No, I'd just have removed it, since to the caller the fact there's an optional is irrelevant, it just confused things. But I'm not asking that to change if you prefer it that way, it's really minor.

vtnerd (commented):
There's a couple of cases where `boost::optional` is used as a class member with a switch to the forward declare header. I don't think this is valid C++ (arguably it could be in certain situations), and the header is likely pulled in indirectly in these cases to allow compilation to complete. Its probably best to revert those headers.

It's also not obvious with the time reports what's due to changes in the files, and effects of OS file caching. Either way, seems like a bump in some cpp files should be plausible.

mj-xmr:
> There's a couple of cases where `boost::optional` is used as a class member with a switch to the forward declare header. I don't think this is valid C++ (arguably it could be in certain situations), and the header is likely pulled in indirectly in these cases to allow compilation to complete. Its probably best to revert those headers.

Correct, thanks for catching it. It also makes me think, that these optionals are then dragged from somewhere else. And yes, forward declaring something, that was already fully declared is a mistake, which shouldn't even compile.

> 
> It's also not obvious with the time reports what's due to changes in the files, and effects of OS file caching. Either way, seems like a bump in some cpp files should be plausible.

I write all the output files into RAM disk. I can repeat the test overnight, running the branch's version first and then the original, to exclude positive effects on caching.

mj-xmr:
> I write all the output files into RAM disk. I can repeat the test overnight, running the branch's version first and then the original, to exclude positive effects of caching. The test will be even more relevant now, since you helped identifying those fwd-s which were incorrect.
I ran a test overnight, and honestly there was no measurable difference. There must be just one large header, which propagates these large headers. I think, that the CLangBuildAnalyser will be able to help me to find the bottlenecks dynamically.
It's just that I need to keep my PRs small and transactional, even if they bring no gain initially.

moneromooo-monero:
-H makes cpp output the header tree for a particular TU.

mj-xmr:
I'm putting this on hold for now, as I apparently need to introduce specialized tooling, in order to do things right and in a more targeted manner.
I'm strongly considering:
https://github.com/Leedehai/C-include-2-dot

mj-xmr:
After expanding the Doxygen generation to outside the src dir, so including the contrib directory, it became clear what pushes the header lower: it's the epee http_auth:
image
Which is then user in other places:
image
All this makes me think, that I should now indeed focus more on proper tooling, rather than shooting in the dark. It was somewhat easier for boost::options, but here I see that I have to give up on this one for the time being. In the end I will be doing more abstractions

src/3rd/boost_monero_optional.h
```diff
@@ -0,0 +1,38 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+// 
+// Parts of this file are originally copyright (c) 2012-2013 The Cryptonote developers
+
+#pragma once
+
+/*
+Only boost::optional wrapper for easier searches in Doxygen
```
mj-xmr, line 34: I don't know what you think about this idea. It's easy to revert, but it helps in searching for how the dependencies are spread across.

CMakeLists.txt
```diff
@@ -54,6 +54,13 @@ if (USE_CCACHE)
 else()
 	message(STATUS "ccache deselected")
 endif()
+option (USE_COMPILATION_TIME_PROFILER "Use compilation time profiler (for CLang >= 9 only)" OFF)
```
mj-xmr, line 57: This has been cherry picked from another PR. I can revert it upon request.

mj-xmr:
Resolved merge conflicts.

mj-xmr:
Related to https://github.com/monero-project/monero/pull/7735
I need to setup a proper framework for testing compatibility between various boost versions. 

Putting to a draft more until then.

mj-xmr:
Blocked by https://github.com/monero-project/monero/pull/7770

mj-xmr:
Resuming, as #7770 has been superseded by the Mac OSX test using the latest Boost version now, same as my local setup.

selsta:
Seeing that this was brought up on IRC I redid the test, on different hardware this time.

This time with following CMake settings, taken from your `clang-build-time-analyzer-run.sh` script and with fresh restart between tests:

```
ccache --clear
cmake .. -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DUSE_CCACHE=OFF -DUSE_COMPILATION_TIME_PROFILER=ON -DBUILD_SHARED_LIBS=ON -DBUILD_TESTS=ON
time make -j1
```

master
```
Executed in   16.43 mins    fish           external
   usr time   15.90 mins    0.07 millis   15.90 mins
   sys time    0.48 mins    1.03 millis    0.48 mins
```

patch
```
Executed in   16.40 mins    fish           external
   usr time   15.87 mins    0.23 millis   15.87 mins
   sys time    0.49 mins    1.81 millis    0.49 mins
```

Both seem within a margin of error equivalent.

mj-xmr:
Closed due to too small a change, compared to number of files touched.

PR #6935, opened 2020-10-25
keep track of peers' peer lists

moneromooo-monero:


selsta:
This has mostly diagnostics value. Should the be merged?

selsta:
No reply, if this becomes relevant again then the code can be taken from this PR.

PR #6937, opened 2020-10-26
Add RELINK_TARGETS and monero_add_target_no_relink 

mj-xmr:
Fortunately Monero was quite prepared for the change, that I wanted to make, so it went smoothly. I have reused the functions:
- monero_add_executable
- monero_add_library_with_deps

to inject a new function: monero_add_target_no_relink,  and switched many (about 20) targets to use the above two functions.
Also added monero_add_minimal_executable, mostly for tests, since they don't need monero_add_executable and haven't even got its definition.

From now on, for shared library builds, by default the relinkage won't be needed for targets, whose dependent dynamic libraries have been relinked, without having their interface changed. This is perfect for TDD, where you write the unit test first, and only modify the production code, recompiling it quite often.

The behavior can be disabled by setting the RELINK_TARGETS variable to ON with:
```
cmake -DRELINK_TARGETS=ON -DBUILD_SHARED_LIBS=ON (...)
```

**Trigger with:**
```
cmake -DBUILD_SHARED_LIBS=ON ../monero-src
make && touch ../monero-src/src/checkpoints/checkpoints.cpp && touch ../monero-src/src/common/notify.cpp && make
```
and notice how only the checkpoints.so and common.so is generated, but nothing else.

**Verification**
For the ease of verification, I remove the bin/* contents and measure the time of regenerating (linking) the binaries on a magnetic drive. An SSD would be faster of course, but would lead to wear of the drive.
From each measurement a dry run of make has been subtracted, with the value of:
time make > /dev/null
 real	0m4,365s
 
1) Typical use case: Debug shared relink(old) vs Debug shared no-relink(new and default)
cmake ../monero-src -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DUSE_CCACHE=OFF -DBUILD_TESTS=OFF
make && rm bin/monero*; sudo sh -c "echo 3 > /proc/sys/vm/drop_caches" && time make
real	0m22,860s-4.3 =18.5s -> time saved for generating the bin directory
2) New reasonable use case: Release shared relink(old) vs Release shared no-relink(new and default, only when -DBUILD_SHARED_LIBS=ON is set specifically)
cmake ../monero-src -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DUSE_CCACHE=OFF -DBUILD_TESTS=OFF
real	0m8,907s-4.3 = 4.6s

Work time:
1.8h
Verification time:
0.3h

### DIFF
--- a/src/blocks/CMakeLists.txt
+++ b/src/blocks/CMakeLists.txt
@@ -45,4 +45,4 @@ foreach(BLOB_NAME checkpoints testnet_blocks stagenet_blocks)
     )
 endforeach()
 
-add_library(blocks STATIC blocks.cpp ${GENERATED_SOURCES})
+monero_add_library(blocks blocks.cpp ${GENERATED_SOURCES})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,24 @@ function (add_definition_if_library_exists library function header var)
   endif()
 endfunction()
 
+option(RELINK_TARGETS "Relink targets, when just a dependant .so changed, but not its header?" OFF)
+function (monero_set_target_no_relink target)
+  if (RELINK_TARGETS MATCHES OFF)
+    # Will not relink the target, when just its dependant .so has changed, but not it's interface
+    set_target_properties("${target}" PROPERTIES LINK_DEPENDS_NO_SHARED true)
+  endif()
+endfunction()
+
+function (monero_add_minimal_executable name)
+  source_group("${name}"
+    FILES
+      ${ARGN})
+
+  add_executable("${name}"
+    ${ARGN})
+    monero_set_target_no_relink( ${name} )
+endfunction()
+
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type" FORCE)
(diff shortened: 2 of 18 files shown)


mj-xmr:
Fortunately I'm able to reproduce this locally. My environment setup was not showing this fault initially.

src/blocks/CMakeLists.txt
```diff
@@ -45,4 +45,4 @@ foreach(BLOB_NAME checkpoints testnet_blocks stagenet_blocks)
     )
 endforeach()
 
-add_library(blocks STATIC blocks.cpp ${GENERATED_SOURCES})
```
mj-xmr, line 48: Was there a reason why blocks had to be STATIC?

moneromooo-monero: If the symbols can be read just fine, I don't think it needs to be static.

mj-xmr:
I will squash the commits and I'm ready.
During the night I will perform some performance tests.

mj-xmr:
> I will squash the commits and I'm ready.
> During the night I will perform some performance tests.

Done.

moneromooo-monero (approved):

PR #6939, opened 2020-10-27
p2p: give all hosts the same chance of being picked for connecting

moneromooo-monero:
This has two stowaway patches necessary for the patch to apply. They were PRed separately.

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -1502,13 +1502,28 @@ namespace nodetool
         });
       }
 
+      auto get_host_string = [](const epee::net_utils::network_address &address) {
+#if BOOST_VERSION > 106600
+        if (address.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
+        {
+          boost::asio::ip::address_v6 actual_ip = address.as<const epee::net_utils::ipv6_network_address>().ip();
+          if (actual_ip.is_v4_mapped())
+          {
+            boost::asio::ip::address_v4 v4ip = make_address_v4(boost::asio::ip::v4_mapped, actual_ip);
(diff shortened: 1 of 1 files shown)


src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -72,6 +72,8 @@
 #define PASSIVE_PEER_KICK_TIME (60 * 1000000) // microseconds
 #define DROP_ON_SYNC_WEDGE_THRESHOLD (30 * 1000000000ull) // nanoseconds
 #define LAST_ACTIVITY_STALL_THRESHOLD (2.0f) // seconds
+#define WAIT_FOR_BLOCK_TIME (20) // seconds
+#define DROP_PEERS_ON_SCORE -2
```
Gingeropolous, line 76: should these lines be user modifiable? (75 and 76), so they are command line options? I can imagine network conditions that would cause block prop delays requiring longer WAIT_FOR_BLOCK_TIME ... i guess, in other words, are 20 and -2 magic numbers or is there a foundation? 

also, is there a way for a peer to regain the confidence of another peer? So the peer gets dropped to ... the gray list? or are they banned for good? 

also, is there a time decay for this score? an AHP could build up a good score (100+), and then stop providing blocks. It would take 102 blocks until this mechanism determined they are AHP

moneromooo-monero: Maybe. Not worth the bother for me though. They're arbitrary. Peers are blocked if they fail twice, a failure being defined as not sending two blocks twice in a row. Score can't get > 5 for this reason.

xiphon (approved):

selsta (approved):

selsta:
Please rebase.

selsta (approved):

PR #6940, opened 2020-10-27
protocol: don't stem transactions to peers who don't relay blocks

moneromooo-monero:


moneromooo-monero:
Note I'm not certain this does not have side effects, I'll ping vtnerd about it.

selsta:
```
2020-10-27T14:55:51.7294630Z /Users/runner/work/monero/monero/src/cryptonote_protocol/levin_notify.cpp:119:108: error: no member named 'm_score' in 'nodetool::p2p_connection_context_t<cryptonote::cryptonote_connection_context>'
2020-10-27T14:55:51.7296980Z         if (!context.m_is_income && context.m_remote_blockchain_height >= min_blockchain_height && context.m_score >= min_score)
2020-10-27T14:55:51.7299470Z                                                                                                    ~~~~~~~ ^
```

selsta:
What happens if no peer has a score of >= 0? Would it go to fluff phase?

PALANTIR-IP:
No

moneromooo-monero:
It should behave as if there are no peers. Looking at the code, I'm not sure whether that means resend after ~3 minutes or at the 4 hour global resend.

vtnerd:
If there are no Dandelion++ peers the tx gets resent at Dandelion++ embargo timeout. I have tested this. This should probably be tweaked to immediate fluffing, but I'm not sure whether this should always be done or whether only in the case of the tx originating at that node.

Gingeropolous:
should this be revisited, perhaps merged?

PR #6942, opened 2020-10-27
p2p: rewrite boost's make_address_v4 to cater for < 1.66

moneromooo-monero:


selsta (approved):

src/p2p/net_node.inl
```diff
@@ -71,6 +71,22 @@
 
 #define MIN_WANTED_SEED_NODES 12
 
+namespace
+{
+
+boost::asio::ip::address_v4 make_address_v4_from_v6(boost::asio::ip::v4_mapped_t m, const boost::asio::ip::address_v6& a)
+{
+  const auto &bytes = a.to_bytes();
```
vtnerd, line 79: The boost implementation has an additional check:
```c++
if (!v6_addr.is_v4_mapped())
  {
    bad_address_cast ex;
    boost::asio::detail::throw_exception(ex);
  }
```
Might be worth doing, and if so I would move to a cpp file.

moneromooo-monero: It's only called after that check. Not really worth it I think.

moneromooo-monero:
Already in in another PR

PR #6947, opened 2020-10-28
add a convenience script to start monero with inbound tor

moneromooo-monero:


hyc:
Would be nice to also have an option to turn on public node support, and add all the appropriate flags for that too.
Would need to add `HiddenServicePort 18089 127.0.0.1:18089` to the torrc, and `--rpc-restricted-bind-ip=0.0.0.0 --rpc-restricted-bind-port=18089 --public-node` to the monerod invocation. The latter would require PR#6948.

moneromooo-monero:
I'm not really interested in making that easier at first glance. We want more nodes.

selsta (approved):

PR #6954, opened 2020-10-31
Fix tx flush callback queueing

vtnerd:
Found this via audit. If an exception is thrown in the flush txes callback (txes are now randomly delayed per-connection based on Dandelion++ recommendations), the state was _not_ reset possibly preventing another callback from being queued. This is related to catching all-exceptions and continuing. There was also a small possibility of multiple queued callbacks, thus the reference count (another thread queues the callback for dispatching while prepping another callback). Async is fun.

Unfortunately I do not believe this to be source of the reported issues, because the exceptions in the callback should be uncommon and the descriptions don't quite match.

### DIFF
--- a/src/cryptonote_protocol/levin_notify.cpp
+++ b/src/cryptonote_protocol/levin_notify.cpp
@@ -283,9 +283,10 @@ namespace levin
           strand(io_service),
           map(),
           channels(),
-          flush_time(std::chrono::steady_clock::time_point::max()),
           connection_count(0),
+          flush_callbacks(0),
           nzone(zone),
+          is_public(is_public),
           pad_txs(pad_txs),
           fluffing(false)
       {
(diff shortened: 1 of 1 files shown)


src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -305,36 +305,28 @@ namespace levin
...
+        ++this_zone.flush_callbacks;
         this_zone.flush_txs.expires_at(flush_time);
-        this_zone.flush_txs.async_wait(this_zone.strand.wrap(fluff_flush{std::move(zone), flush_time}));
+        this_zone.flush_txs.async_wait(this_zone.strand.wrap(fluff_flush{std::move(zone)}));
       }
 
       void operator()(const boost::system::error_code error)
       {
-        if (!zone_ || !zone_->p2p)
+        if (!zone_ || --zone_->flush_callbacks || !zone_->p2p)
```
moneromooo-monero, line 322: That can underflow on error (ie, flush_callbacks is 0 upon entry), which will presumably have bad effects, so an error message would be good.

moneromooo-monero (approved):
I don't think I really understand it all fully, but it looks ok as far as I can tell.

src/cryptonote_protocol/levin_notify.cpp
vtnerd: Underflow can only occur if ASIO invokes a queued callback multiple times. This should never happen, but all of the callbacks already have defensive code. So I added a zero check before the decrement.

selsta (approved):

selsta:
Please rebase.

vtnerd:
Rebased.

selsta (approved):

PR #6961, opened 2020-11-02
[release-v0.17] net_node: add --ban-list option

selsta:


sumogr:
IMO add a permanent ban option something like `ban -1 <IP>` which will ban an ip at max time_t limit and only those IPs will be exported to that file. Also make that file to be re-read upon each daemon restart to automatically ban these "asshole" ips.
I did that somewhere else i can port it if you wish

selsta:
You can save a ban list at any place and add ban-list=/path/to/list.txt and it will permanently get read on startup.

sumogr:
> You can save a ban list at any place and add ban-list=/path/to/list.txt and it will permanently get read on startup.
true it works as well
however i insist on my opinion of banning "asshole" nodes beyond the 2 days or so ban limit

selsta:
Not sure what you mean. This bans them for basically unlimited time.

sumogr:
> Not sure what you mean. This bans them for basically unlimited time.

`block_host(*parsed_addr, std::numeric_limits<time_t>::max());`
my bad i've just seen it ;)

moneromooo-monero (approved):

PR #6962, opened 2020-11-02
p2p: use /16 filtering on IPv4-within-IPv6 addresses

moneromooo-monero:
IPv6 addresses include a range that can map IPv4 addresses,
which allowed those mapped addresses to bypass filtering.

This filter should be replaced by AS filtering at some point.


selsta (approved):

src/p2p/net_node.inl
```diff
@@ -1436,6 +1436,20 @@ namespace nodetool
             const uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
             classB.insert(actual_ip & 0x0000ffff);
```
xiphon, line 1437: Could also apply `ntohl` here while we are at it.

PR #6966, opened 2020-11-02
Fix CLI and unrestricted RPC relay_tx with stempool [0.17]

vtnerd:
Same as PR #6949 but on 0.17 branch.


selsta (approved):
Equivalent to reviewed master PR

PR #6967, opened 2020-11-02
Fix tx flush callback queueing [0.17]

vtnerd:
Same as PR #6954 but on 0.17 branch.


selsta (approved):
Equivalent to reviewed master PR

PR #6975, opened 2020-11-03
Run Valgrind checks for any binary in a list mj-xmr

mj-xmr:
The new tool (utils/health/valgrind-tests.sh) is able to run Valgrind checks (cachegrind, callgrind and memcheck) on a list of executables (and their arguments) provided line-by-line via a text file, being the parameter of the script.
The *.out files can be analyzed with the kcachegrind tool.

I have minimized all of the repetitions. This was somewhat harder than expected, because the memcheck tool has a bit different interface than the other tools.

Below are the reports generated for the unit_tests executable:
[Valgrind reports for unit tests](https://github.com/monero-project/monero/files/5483682/valgrind-output.zip)
Summary:
- there were no memory leaks (!) (however, this doesn't mean, that an ever extending vector can't consume all the memory, before the daemon is closed)
- there's an apparent bottleneck within the epee's rolling median's insert method 

monero-rolling-median-bottleneck

Active work time 4.2h
Report generation and trials: 1.5 day (think in terms of power costs, rather than work)

### DIFF
--- a/utils/health/valgrind-tests.sh
+++ b/utils/health/valgrind-tests.sh
@@ -1,0 +1,161 @@
+#!/bin/bash -e
+
+# Copyright (c) 2014-2020, The Monero Project
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are
+# permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of
+#    conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this list
+#    of conditions and the following disclaimer in the documentation and/or other
+#    materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may be
+#    used to endorse or promote products derived from this software without specific
+#    prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# This script is able to run valgrind's callgrind, cachegrind and memcheck for a given set of executables.
+# It expects ONE PARAMETER, which points to a file with paths to executables and their arguments, written line by line.
+
+if [ "$#" -ne 1 ]; then
+    echo "Please provide an argument, which points to a file with paths to executables and their arguments, written line by line. For example:"
+    echo ""
+    echo "ls -l -h"
+    echo "build/tests/unit_tests/unit_tests"
+    exit 1
+fi
+
+FILE_IN="$1"
+DIR_OUT="build/valgrind-output" # Using build as the base output directory, as it's ignored in .gitignore
+
+function is_file_or_exit {
+	FILE="${1}"
+	if [ -f $FILE ]; then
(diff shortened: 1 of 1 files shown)


vtnerd (commented):
I don't see how this is useful. Running the various valgrind tools then loading them isn't particularly difficult, is it? That it requires a bash script? Particularly since it just runs everything run one after the other, which is a strange use case typically (at least how and most people I know have used the tools).

mj-xmr:
> I don't see how this is useful. Running the various valgrind tools then loading them isn't particularly difficult, is it? That it requires a bash script? Particularly since it just runs everything run one after the other, which is a strange use case typically (at least how and most people I know have used the tools).

Everything, that you want to automate needs a script. A fixed script guarantees you uniformity of measurements across a varying data sample, so that the samples become objectively comparable, thanks to the fixed lense.

Why you need to automate:
- to be able to run these extremely long tasks during the night (or rather many nigths)
- to be able to loop the script(s) across various commits in order to generate and publish a summary report with all the sub reports accessible via click without having to re-run the checks on every dev's machine. Such a report helps finding regressions
- to focus on the real problems and not pretenting to be busy by wasting time on repetitive tasks

moneromooo-monero (approved):
As a side note, I've noticed you add your nick on every patch's commit message. git has an author field where this is supposed to go.

mj-xmr:
> As a side note, I've noticed you add your nick on every patch's commit message. git has an author field where this is supposed to go.

I think I saw others doing it the same way or I misinterpreted it. Will correct this.

vtnerd:
> Everything, that you want to automate needs a script. A fixed script with fixed options guarantees you uniformity of measurements across varying data samples, so that the samples become objectively comparable, thanks to the fixed lense.

Callgrind/cachegrind are consistent and repeatable across machines, but they don't capture wait/blocking time for threaded or I/O operations nor instruction-level parallelism. And simulating real-production usage+behavior can be tricky at times too. So the "samples become objectively comparable, thanks to the fixed lense" is definitely too strong; you typically have to dig into the details to verify whether the numbers are expected to improve production CPU time or throughput.

In fact, I cannot recall allowing a single developer say "callgrind 1 was X instructions, callgrind 2 was Y instructions, therefore X is preferred" without seeing a developer explain why the results changed. I've seen incorrect "instruction number go down" analysis far too much. Often the biggest changes in callgrind results are requests for memory and heap _state_. The latter, in many cases, is not useful for determining anything objective between two implementations.

I guess if this gets people using the tools somehow its a plus.

> to be able to run these extremely long tasks during the night (or rather: many nigths)

Queueing them up would be kind of interesting. But I frequently want to run so many callgrinds in a row when doing analysis ... but sure I could see me using it in this context.

> to be able to loop the script(s) across various commits in order to generate and publish a summary report with all the sub reports accessible via click without having to re-run the checks on every dev's machine. Otherwise, the devs will refrain from running these checks in the first place. Such an automated report helps finding regressions btw.

This is a reasonable argument. We might also want to enable the ASAN option for the Ubuntu unit test CI anyway, because it does several useful checks faster than valgrind. Running this script against all PR changes might be rough on the Github CPU time (not sure how that is allotted). Maybe this can be run on master and release branch changes only?

> to focus on the real problems (like analysing the results and applying countermeasures), as opposed to pretenting to be busy by wasting time on repetitive and distractive tasks (like trying to recall a given tool's syntax and waiting for results, while you could already have them before breakfast)

This argument is dubious. If someone is "pretending to be busy" while waiting for results of analysis tools or taking a good chunk of time to repeatedly lookup usage of the tools, providing a script isn't somehow going to make that person change their behavior.

vtnerd:
Perhaps I'm being too hard on this. Callgrind is more useful in our CI setup since its not bare metal. Its just frustrating because its an indirect measurement of what we actually want to know (how does X instructions translate to Y clock-time). Coming with more/better performance tests is probably the key though, these scripts would just be a first pass for something that developer can reproduce with wall-clock time on a bare-metal machine.

mj-xmr:
> Callgrind/cachegrind are consistent and repeatable across machines, but they don't capture wait/blocking time for threaded or I/O operations nor instruction-level parallelism. And simulating real-production usage+behavior can be tricky at times too. So the "samples become objectively comparable, thanks to the fixed lense" is definitely too strong; 

The repeatability only speaks positively of Valgrind. But you can run the tool with different options, which won't deliver consistent results. Fixing the options in a script resolves this problem. And since the additional options are costly, you want them to be ran overnight.

>you typically have to dig into the details to verify whether the numbers are expected to improve production CPU time or throughput.

Oh yes, I personally use Valgrind just to tell me where to look and then I rely only on myself. Typically it's just a silly mistake, or an inefficient algorithm - linear search, where logarithmic search could be applied, or an abuse of dynamic memory.

> Queueing them up would be kind of interesting. But I frequently want to run so many callgrinds in a row when doing analysis ... but sure I could see me using it in this context.

Sure. Me to, but for a reasonably smaller sample. This leaves a gap of having no time for running it on a larger sample.

> This is a reasonable argument. We might also want to enable the ASAN option for the Ubuntu unit test CI anyway, because it does several useful checks faster than valgrind. Running this script against all PR changes might be rough on the Github CPU time (not sure how that is allotted). Maybe this can be run on master and release branch changes only?

Yes, this was my intention, plus being able to compare my local branch with a couple of reference master builds.
And yeah, I'd be careful about bloating the CI. If it's faster than Valgrind, then it should be OK though.

Also, referring to your next comment - having some optional entry level check, not requiring much investment and not taking away CI's time, doesn't mutually exclude searching and implementing better solutions in the future. As you could see in this PR's description, it was able to immediately catch a dominating bottleneck in the epee, which was otherwise continuously overlooked.
(I haven't looked at the code myself yet, but when I see a bottleneck in "insert", I intuitively think of a linear search for the insert index or moving a lot of memory around, while a simple std::map could have been used instead)

selsta (approved):

PR #6978, opened 2020-11-04
Assholes 3 0.17

moneromooo-monero:


moneromooo-monero:
Clicked PR before it managed to resolve the 0.17 switch

PR #6986, opened 2020-11-05
Balance includes unconfirmed payments

woodser:


woodser:
This PR adds unconfirmed payments to the wallet's balance when `strict=false`.

src/wallet/wallet2.cpp
```diff
@@ -6017,6 +6017,18 @@ std::map<uint32_t, uint64_t> wallet2::balance_per_subaddress(uint32_t index_majo
...
+   for (const auto& utx: m_unconfirmed_payments)
+   {
+    if (utx.second.m_pd.m_subaddr_index.major == index_major)
+    {
+      auto found = amount_per_subaddr.find(utx.second.m_pd.m_subaddr_index.minor);
+      if (found == amount_per_subaddr.end())
+        amount_per_subaddr[utx.second.m_pd.m_subaddr_index.minor] = utx.second.m_pd.m_amount;
+      else
+        found->second += utx.second.m_pd.m_amount;
+    }
```
moneromooo-monero, line 6030: These 5 lines can be just:
amount_per_subaddr[utx.second.m_pd.m_subaddr_index.minor] += utx.second.m_pd.m_amount;
since the new element is zero initialized.

moneromooo-monero:
Why do you think this is a good idea ?

dEBRUYNE-1:
I personally don't think this should be enabled by default, as it may confuse users.

woodser:
Today when a Monero wallet receives funds, the balance remains unchanged until the first confirmation despite showing the incoming transfer.  As a result:

- The balance does not reflect incoming transfers - outgoing transfers which is incorrect accounting.
- The user must wait up to several minutes before seeing their balance credited with the transfer.
- Monero's wallet experience is inconsistent with other wallet software and user expectation.

This change aims to improve the user experience by immediately crediting unconfirmed deposits to the balance, thereby:

- Treating unconfirmed deposits and withdraws exactly the same so the accounting is always correct.
- Providing immediate feedback to the user to increase comfort with the system.

This will work as intended the vast majority of cases, which I believe the user experience should be streamlined for.  If the tx is invalidated due to reorg or double spend:

- Unconfirmed withdraws will be affected in exactly the same way, requiring the wallet to update its state.
- The UI can/should be responsible to indicate if such an event occurs, rather than compromising the wallet's accounting.

woodser:
@dEBRUYNE-1 I am wondering how not crediting the balance when an incoming transfer is received would be less confusing to users than crediting the balance, which would be more in accordance with user expectation.

moneromooo-monero:
>    The balance does not reflect incoming transfers - outgoing transfers which is incorrect accounting.

Does it reflect the ones that are actually in the blockchain ?

>    The user must wait up to several minutes before seeing their balance credited with the transfer.

The transaction is not in the blockchain yet, is it ?

moneromooo-monero:
Maybe we want three balances then. Actual balance, unlocked balance, and your new balance-that-will-be-if-this-gets-mined.

woodser:
> > The balance does not reflect incoming transfers - outgoing transfers which is incorrect accounting.
> 
> Does it reflect the ones that are actually in the blockchain ?

It reflects the ones that are actually in the blockchain + unconfirmed outgoing transfers, but not unconfirmed incoming transfers.

> 
> > The user must wait up to several minutes before seeing their balance credited with the transfer.
> 
> The transaction is not in the blockchain yet, is it ?

It is not in the blockchain, but it is very probable to be added.

woodser:
> Maybe we want three balances then. Actual balance, unlocked balance, and your new balance-that-will-be-if-this-gets-mined.

Currently the "actual balance" includes transfers in the blockchain + new balance-that-will-be-if-this-gets-mined for outgoing transfers but not incoming transfers.  Why include one and not the other?

xiphon:
> It reflects the ones that are actually in the blockchain + unconfirmed outgoing transfers, but not unconfirmed incoming transfers.

Agree. Should be consistent.

moneromooo-monero:
That's true. Those outgoing ones ought not be there either, but I agree it would be confusing to users.

woodser:
> That's true. Those outgoing ones ought not be there either, but I agree it would be confusing to users.

This describes the behavior with `strict=true`, whereas `strict=false` considers non-blockchain data (unconfirmed spent outputs are deducted from the balance, unconfirmed received outputs are proposed in this PR).

moneromooo-monero (approved):

PR #6977, opened 2020-11-04
Clang-tidy integrated into CMakeLists.txt

mj-xmr:
The tool should be called from the source directory with:
utils/health/clang-tidy-run.sh

Apart from the above script, the changes include integrating the tool into the main CMakeLists.txt and writing a CMake module.

The report generation will take a good deal of time (it took 10 hours). But once it's done, I'm sure you'll find it horribly annoying :-)
[Here's a snippet](https://github.com/monero-project/monero/files/5490147/clang-tidy-result.txt)
[Here's a C++ only report (unpack with tar -xvf). Weights 528 MB after unpacking!](https://github.com/monero-project/monero/files/5496427/clang-tidy-result.txt.txz.gz)
[Here's a C only report (unpack with tar -xvf)](https://github.com/monero-project/monero/files/5535122/clang-tidy-result-C.txt.txz.gz)

```
[100%] Building CXX object external/easylogging++/CMakeFiles/easylogging.dir/easylogging++.cc.o
/tmp/monero/monero/external/easylogging++/ea_config.h:1:1: warning: header is missing header guard [llvm-header-guard]
#pragma once
^
/tmp/monero/monero/external/easylogging++/ea_config.h:3:10: warning: inclusion of deprecated C++ header 'limits.h'; consider using 'climits' instead [hicpp-deprecated-headers]
#include <limits.h>
         ^
note: this fix will not be applied because it overlaps with another fix
/tmp/monero/monero/external/easylogging++/ea_config.h:3:10: warning: inclusion of deprecated C++ header 'limits.h'; consider using 'climits' instead [modernize-deprecated-headers]
note: this fix will not be applied because it overlaps with another fix
/tmp/monero/monero/external/easylogging++/ea_config.h:6:9: warning: macro 'ELPP_DEFAULT_LOG_FILE' used to declare a constant; consider using a 'constexpr' constant [cppcoreguidelines-macro-usage]
#define ELPP_DEFAULT_LOG_FILE ""
        ^
/tmp/monero/monero/external/easylogging++/easylogging++.cc:23:1: warning: initialization of 'elCrashHandler' with static storage duration may throw an exception that cannot be caught [cert-err58-cpp]
INITIALIZE_EASYLOGGINGPP
```

Active work time: 4.7h

PR #6989, opened 2020-11-06
Clang IWYU header checker script

mj-xmr:
Run with utils/health/clang-include-what-you-use-run.sh
include-what-you-use analyses the header file hierarchy and delivers hints on how to reduce their complexity.

Within this branch, I figured out how to ignore errors of the linting tools, and used this trick in the clang-tidy branch, which shares the same problem. It's the -k option for make.

This concludes the set of tools, that I find crucial to begin a targeted compilation time reduction. I'm now able to perform the changes in the code. Once all my tools (scripts) are merged, in parallel I will be creating the promised HTML summary report, with all the sub-reports accessible for every developer.

The report generation crashed at 59%, even with the -k option. I think, that the info it provided it already useful.
[iwyu-result.txt](https://github.com/monero-project/monero/files/5504445/iwyu-result.txt)

Active work time: 2.3h

### DIFF
--- a/utils/health/README.md
+++ b/utils/health/README.md
@@ -15,6 +15,10 @@ On the first run, the script will complain about the missing ClangBuildAnalyzer 
 `utils/health/clang-tidy-run.sh`
 Performs Lint checks on the source code and stores the result in the build directory. More information on the [home page](https://clang.llvm.org/extra/clang-tidy/).
 
+##include-what-you-use
+`utils/health/clang-include-what-you-use-run.sh`
+Analyses the header file hierarchy and delivers hints on how to reduce their complexity. More information on the [home page](https://include-what-you-use.org/).
+
 
 ##Valgrind checks
 `utils/health/valgrind-tests.sh`
--- a/utils/health/clang-include-what-you-use-run.sh
+++ b/utils/health/clang-include-what-you-use-run.sh
@@ -1,0 +1,75 @@
+#!/bin/bash -e
+
+# Copyright (c) 2014-2020, The Monero Project
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are
+# permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of
+#    conditions and the following disclaimer.
+#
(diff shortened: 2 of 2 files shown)


mj-xmr:
Same as with [Clang-Tidy](https://github.com/monero-project/monero/pull/6977#issuecomment-723436312):

> I have ran a test with Clang Tidy on C files with the make -k option, and it also crashed at about 59%, so -k is a good idea, but compiling C files should still be a separate use case. I think will enable it separately. Putting this branch on hold for now.

mj-xmr:
> I have ran a test with Clang Tidy on C files with the make -k option, and it also crashed at about 59%, so -k is a good idea, but compiling C files should still be a separate use case. I think will enable it separately. Putting this branch on hold for now.

After disabling C support in iwyu, there was no improvement. Still crashed at 59%. It has to stay as it is.

moneromooo-monero (approved):

selsta (approved):

selsta:
Please rebase.

mj-xmr:
> Please rebase.

Done.

selsta:
Please squash the commits so that we can merge this one.

mj-xmr:
> Please squash the commits so that we can merge this one.

Not that I necessarily want to stay in opposition, but the second commit is an empty one. As such, it should automatically be squashed. Would you like to try it out or you don't want to risk and prefer, that I squash?

selsta:
We don’t use any Github merge features so is no automatic squashing as far as I know.

mj-xmr:
After reading up, it seems I'm wrong:
https://stackoverflow.com/a/12470056

"To arbitrarily create a new commit, such as for re-triggering build tooling (via mattLummus) or for the sake of personal logging or metrics (via DynamiteReed). However, think twice: depending on your branch/merge structure, commits may live for a very long time, so a "just commit nothing" strategy may be inadvertently pollute your team's repository with temporary workflow artifacts and make it hard to separate code revisions from ephemeral cruft."

mj-xmr:
> Please squash the commits so that we can merge this one.

Done.

mj-xmr:
Rebased to get the latest CI fix.

selsta (approved):

PR #6990, opened 2020-11-06
Add support for i2p and tor seed nodes [0.17]

vtnerd:
PR #6897 on 0.17 branch.


selsta (approved):
Tested, also branch equivalent to reviewed PR.

PR #6993, opened 2020-11-07
p2p: fix accessing non existent element of map

moneromooo-monero:


selsta (approved):

vtnerd (commented):
Where was the invalid access coming from? That is the more relevant question. And Tor/I2P seeds are still in the incorrect location. Accesses to i2p/tor nodes were likely from the p2p state file file/command line, and not the seed list.

PR #6999, opened 2020-11-09
epee: tighten the URL parsing regexp a little

moneromooo-monero:
fixes massive amounts of time spent on pathological inputs

Found by OSS-Fuzz

### DIFF
--- a/contrib/epee/include/net/net_parse_helpers.h
+++ b/contrib/epee/include/net/net_parse_helpers.h
@@ -131,8 +131,8 @@ namespace net_utils
   inline
     bool parse_url_ipv6(const std::string url_str, http::url_content& content)
   {
-    STATIC_REGEXP_EXPR_1(rexp_match_uri, "^((.*?)://)?(\\[(.*)\\](:(\\d+))?)(.*)?", boost::regex::icase | boost::regex::normal);
-    //                                     12         3   4      5 6        7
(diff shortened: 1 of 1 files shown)


xiphon (approved):

contrib/epee/include/net/net_parse_helpers.h
```diff
@@ -131,8 +131,8 @@ namespace net_utils
   inline
     bool parse_url_ipv6(const std::string url_str, http::url_content& content)
   {
-    STATIC_REGEXP_EXPR_1(rexp_match_uri, "^((.*?)://)?(\\[(.*)\\](:(\\d+))?)(.*)?", boost::regex::icase | boost::regex::normal);
-    //                                     12         3   4      5 6        7
+    STATIC_REGEXP_EXPR_1(rexp_match_uri, "^(([^:]*?)://)?(\\[(.*)\\](:(\\d+))?)(.*)?", boost::regex::icase | boost::regex::normal);
```
iamamyth, line 134: This could be restricted a bit further:
```
^(([^:]+)://)?...
```

The `*?` makes no sense in this context both semantically (url schemes must contain at least one character, hence the change to `+`) and logically (there's a negated one-character class followed by the character being negated, so a greedy match is the same as a non-greedy one; in both cases, the match must end at the `:`, meaning `*?` can be replaced with `*`).

moneromooo-monero: I kept this after seeing: https://www.loggly.com/blog/five-invaluable-techniques-to-improve-regex-performance/ (Lazy Quantifiers). I did not try to time with and without though, but the patch makes the test case pretty much instant compared to giving up after a few seconds so...

selsta (approved):

PR #6956, opened 2020-11-01
Clang build time analyser mj xmr

mj-xmr:
Integrated [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer) into utils. You can call it without any parameters from the Monero source directory with:

`utils/health/clang-build-time-analyzer-run.sh`

It will try to find the ClangBuildAnalyzer binary, and if it's not found (the initial state), it will ask the user to call a script, that clones and builds the CBA:

`utils/health/build-scripts/clang-build-time-analyzer-clone-build.sh `

I will now proceed according to the time measurements, but beware: some code will have to be moved around.

Below are some snippets from the generated report:

```
**** Files that took longest to codegen (compiler backend):
147200 ms: tests/core_tests/CMakeFiles/core_tests.dir/chaingen_main.cpp.o
114191 ms: src/wallet/CMakeFiles/obj_wallet.dir/wallet2.cpp.o
 94273 ms: src/rpc/CMakeFiles/obj_rpc.dir/core_rpc_server.cpp.o
 78506 ms: src/wallet/CMakeFiles/wallet_rpc_server.dir/wallet_rpc_server.cpp.o

*** Expensive headers:
168138 ms: /tmp/monero/monero/contrib/epee/include/storages/portable_storage.h (included 82 times, avg 2050 ms), included via:
  load_from_binary.cpp.o portable_storage_template_helper.h  (3539 ms)
  load_from_json.cpp.o portable_storage_template_helper.h  (3526 ms)
  ...

155623 ms: /tmp/monero/monero/contrib/epee/include/storages/portable_storage_template_helper.h (included 79 times, avg 1969 ms), included via:
  load_from_binary.cpp.o  (3571 ms)
  load_from_json.cpp.o  (3561 ms)
  ...

147344 ms: /tmp/monero/monero/src/cryptonote_core/cryptonote_core.h (included 48 times, avg 3069 ms), included via:
  node_server.cpp.o  (6755 ms)
  rpc_handler.cpp.o  (6363 ms)
  ...

127540 ms: /tmp/monero/monero/src/cryptonote_basic/cryptonote_basic.h (included 121 times, avg 1054 ms), included via:
  hardfork.cpp.o  (2484 ms)
  tx_sanity_check.cpp.o  (2460 ms)
  account.cpp.o account.h  (2241 ms)
  ...
```

[And the full report is here](https://github.com/monero-project/monero/files/5471196/cba-result.txt)

PR #7000, opened 2020-11-09
Reduce compilation time of epee/portable_storage.h

mj-xmr:
I have touched here the header, that cumulatively was taking the longest time, as reported by ClangBuildAnalyser.
The change involves mainly moving inline functions into .cpp files, so that their dependencies can be hidden within.

portable_storage.h now
portable_storage.h now

portable_storage.h now (outdated - see review):
portable_storage.h now

portable_storage.h before:
portable_storage.h before

Speed improvements:
| Prevous   |      Current      |
|----------|-------------|
|Compilation (798 times): |  Compilation (802 times): |
| Parsing (frontend):         1728.2 s |     Parsing (frontend):         1488.6 s   | 
| Codegen & opts (backend):   1739.0 s | Codegen & opts (backend):   1587.8  |
| 168138 ms: portable_storage.h (included 82 times, avg 2050 ms) | 136062 ms: portable_storage.h (included 83 times, avg 1639 ms) |

Reduction by 14 or by 19%, depending on how you calculate it.

[Full report](https://github.com/monero-project/monero/files/5517820/cba-result.txt)
[Reference report available here](https://github.com/monero-project/monero/pull/6956)
 
  

Active work time so far: 3.5 h

### DIFF
--- a/contrib/epee/include/misc_language.h
+++ b/contrib/epee/include/misc_language.h
@@ -28,9 +28,11 @@
 
 #pragma once
 
-#include <limits>
-#include <boost/thread.hpp>
 #include <boost/utility/value_init.hpp>
+#include <boost/shared_ptr.hpp>
+#include <limits>
+#include <functional>
+#include <vector>
 namespace epee
 {
 #define STD_TRY_BEGIN() try {
@@ -95,16 +97,7 @@ namespace misc_utils
       return memcmp(&_Left, &_Right, sizeof(_Left)) < 0;
   }
 	
-
-	inline
-	bool sleep_no_w(long ms )
-	{
-		boost::this_thread::sleep( 
-			boost::get_system_time() + 
-			boost::posix_time::milliseconds( std::max<long>(ms,0) ) );
-		
-		return true;
-	}
+	bool sleep_no_w(long ms );
 
   template <typename T>
   T get_mid(const T &a, const T &b)
--- a/contrib/epee/include/storages/portable_storage.h
+++ b/contrib/epee/include/storages/portable_storage.h
@@ -24,20 +24,12 @@
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-
-
 #pragma once 
 
-#include <type_traits>
-
-#include "misc_language.h"
 #include "portable_storage_base.h"
-#include "portable_storage_from_bin.h"
-#include "portable_storage_to_json.h"
-#include "portable_storage_from_json.h"
 #include "portable_storage_val_converters.h"
+#include "misc_log_ex.h"
 #include "span.h"
-#include "int-util.h"
 
 namespace epee
 {
@@ -92,7 +84,7 @@ namespace epee
       //-------------------------------------------------------------------------------
       bool		store_to_binary(byte_slice& target, std::size_t initial_buffer_size = 8192);
       bool		load_from_binary(const epee::span<const uint8_t> target, const limits_t *limits = NULL);
-      bool		load_from_binary(const std::string& target, const limits_t *limits = NULL) { return load_from_binary(epee::strspan<uint8_t>(target), limits); }
+      bool		load_from_binary(const std::string& target, const limits_t *limits = NULL);
       template<class trace_policy>
       bool		  dump_as_xml(std::string& targetObj, const std::string& root_name = "");
       bool		  dump_as_json(std::string& targetObj, size_t indent = 0, bool insert_newlines = true);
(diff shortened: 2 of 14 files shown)


contrib/epee/include/storages/portable_storage_template_helper.h
```diff
@@ -84,7 +85,7 @@ namespace epee
     }
     //-----------------------------------------------------------------------------------------------------------
     template<class t_struct>
-    bool load_t_from_binary(t_struct& out, const epee::span<const uint8_t> binary_buff)
+    bool load_t_from_binary(t_struct& out, const epee::span<const uint8_t> & binary_buff)
```
vtnerd, line 88: Drop the `&`, this was not a mistake or oversight. If this was done elsewhere in the patch too, which is unrelated to the PR title, please revert them. There are numerous internet discussions about using `string_view`/`string_ref`/`span` objects by value instead of reference. The type is has 1 pointer and 1 `size_t` value (small), and is trivial to copy and destruct.

A quick summary of the discussions - this type by value requires setting two registers on all modern ABIs, and even on legacy x86 passing the value is fairly simple. In the majority of cases, this object is copying values from a `std::string` or `std::vector` so a `&` forces the compiler to push space onto the stack, copy a pointer+length, then set a register to that stack location. Then the backend optimizer is forced to consider aliasing with other `span` objects of this type (which should rarely be an issue, but why bother). In contrast, by value requires setting two registers and the backend knows the `span<>` type itself cannot be aliased.

contrib/epee/include/misc_language.h
```diff
@@ -95,16 +97,7 @@ namespace misc_utils
       return memcmp(&_Left, &_Right, sizeof(_Left)) < 0;
   }
 	
-
-	inline
```
vtnerd, line 99: This seems unrelated to the other changes, but it is small.

contrib/epee/include/storages/parserse_base_utils.h
```diff
@@ -28,10 +28,7 @@
 
 #pragma once 
 
-#include <algorithm>
```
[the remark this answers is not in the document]
mj-xmr: Thanks for the hint and the coordination effort.
Merge conflicts is the last thing, that I want.

contrib/epee/include/storages/portable_storage_from_bin.h
```diff
@@ -32,6 +32,8 @@
 #include "portable_storage_base.h"
 #include "portable_storage_bin_utils.h"
```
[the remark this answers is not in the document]
mj-xmr: Now, since I moved it from away from portable_storage.h, yes, it doesn't have that much impact anymore.

contrib/epee/include/misc_language.h
mj-xmr: It prevents propagation of boost::thread. It appears small, but can make a large difference cumulative.

contrib/epee/include/storages/portable_storage_template_helper.h
mj-xmr: It's not the size of the construct, that matters to me in this regard, but the impossibility of forward declaring a class or a template class, which is being passed by a value. I can prove to which extent a forward declaration of a moderately heavy template improves, but whenever discussing such topic with people whose priority is runtime optimization, they rarely bring up any measurements.

In this case, I reverted the change, as I was forced to include the span.h anyway, because it's constructor is being used elsewhere. Template segmentation will be the answer to this problem, but I don't want to make too large changes in one go.

contrib/epee/include/storages/portable_storage_base.h
```diff
@@ -29,10 +29,11 @@
 #pragma once 
 
 #include <boost/variant.hpp>
-#include <boost/any.hpp>
+//#include <boost/any.hpp>
```
mj-xmr: To remove

contrib/epee/include/storages/portable_storage.h
```diff
@@ -24,24 +24,15 @@
...
-#include "portable_storage_base.h"
-#include "portable_storage_to_bin.h"
-#include "portable_storage_from_bin.h"
-#include "portable_storage_to_json.h"
-#include "portable_storage_from_json.h"
 #include "portable_storage_val_converters.h"
-#include "span.h"
-#include "int-util.h"
+#include "portable_storage_base.h"
+#include "misc_log_ex.h"
```
xiphon, line 31: I would better sort it alphabetically

contrib/epee/src/portable_storage.cpp
```diff
@@ -0,0 +1,243 @@
...
+        return nullptr;
+      array_entry_t<section>& sec_array = boost::get<array_entry_t<section>>(ar_entry);
+      section* psec = sec_array.get_first_val();
+      if(!psec)
+        return nullptr;
+      h_child_section = psec;
+      return &ar_entry;
+      CATCH_ENTRY("portable_storage::get_first_section", nullptr);
+    }
+    //---------------------------------------------------------------------------------------------------------------
```
xiphon, line 188: -

contrib/epee/src/portable_storage.cpp
```diff
@@ -0,0 +1,243 @@
...
+    hsection portable_storage::insert_new_section(const std::string& pentry_name, hsection psection)
+    {
+      TRY_ENTRY();
+      storage_entry* pse = insert_new_entry_get_storage_entry(pentry_name, psection, section());
+      if(!pse) return nullptr;
+      return &boost::get<section>(*pse);
+      CATCH_ENTRY("portable_storage::insert_new_section", nullptr);
+    }
+    
+
```
xiphon, line 166: -

xiphon (approved):

contrib/epee/include/storages/portable_storage.h
mj-xmr: Header order has impact on the ease of usage of the classes / libraries, that you define. It's nicely described in this short blog post:
https://blog.knatten.org/2010/07/01/the-order-of-include-directives-matter/

contrib/epee/src/portable_storage.cpp
mj-xmr: Removed.

contrib/epee/include/storages/portable_storage.h
mj-xmr: Ordered alphabetically.

contrib/epee/src/portable_storage.cpp
mj-xmr: Removed.

contrib/epee/src/portable_storage.cpp
```diff
@@ -1,10 +1,44 @@
...
 #include "byte_slice.h"
 #include "byte_stream.h"
 #include "misc_log_ex.h"
 #include "span.h"
 #include "storages/portable_storage.h"
+#include "storages/portable_storage_to_json.h"
+#include "storages/portable_storage_from_json.h"
 #include "storages/portable_storage_to_bin.h"
+#include "storages/portable_storage_from_bin.h"
+#include "span.h"
```
selsta, line 36: span.h is not included twice, intended?

contrib/epee/include/storages/portable_storage.h
```diff
@@ -91,8 +83,8 @@ namespace epee
 
       //-------------------------------------------------------------------------------
       bool		store_to_binary(byte_slice& target, std::size_t initial_buffer_size = 8192);
-      bool		load_from_binary(const epee::span<const uint8_t> target, const limits_t *limits = NULL);
-      bool		load_from_binary(const std::string& target, const limits_t *limits = NULL) { return load_from_binary(epee::strspan<uint8_t>(target), limits); }
+      bool		load_from_binary(const epee::span<const uint8_t> & target, const limits_t *limits = NULL);
```
selsta, line 86:
```suggestion
      bool		load_from_binary(const epee::span<const uint8_t>  target, const limits_t *limits = NULL);
```

I would feel better if we could revert the pass by reference, this PR should not include extra changes like this.

selsta:
Left two minor comments, no other code changes detected.

contrib/epee/src/portable_storage.cpp
mj-xmr: No, an overlook.

contrib/epee/include/storages/portable_storage.h
mj-xmr: Two voices are enough to convince me.
Soon the measurements will show if span.h is an outstanding bottleneck anyway.

contrib/epee/include/storages/portable_storage_template_helper.h
mj-xmr: After gathering a 2nd opinion, I switched the span reference back to by-value.

contrib/epee/include/storages/portable_storage.h
selsta: I don't have an opinion on reference vs copy here but this PR should not introduce changes like this so that we avoid introducing unintended changes.

selsta (approved):

PR #7005, opened 2020-11-10
Allow byte_stream->byte_slice conversion to shrink unused buffer space

vtnerd:
This allows for the optional shrinkage of the `byte_stream` buffer when converting to an immutable `byte_slice`. This uses `realloc` so in the bulk of calls this can be done without a copy. Larger buffers are likely to be "overcommited" by OSes anyway, so it frequently could be a NOP. Its still beneficial given that Monero supports 32-bit builds, where the address space reclamation can be beneficial.

### DIFF
--- a/contrib/epee/src/byte_slice.cpp
+++ b/contrib/epee/src/byte_slice.cpp
@@ -36,6 +36,11 @@
 #include "byte_slice.h"
 #include "byte_stream.h"
 
+namespace
+{
+  const std::size_t page_size = 4096;
+}
+
 namespace epee
 {
   struct byte_slice_data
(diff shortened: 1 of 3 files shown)


contrib/epee/src/byte_slice.cpp
```diff
@@ -173,16 +178,27 @@ namespace epee
     : byte_slice(adapt_buffer{}, std::move(buffer))
   {}
 
-  byte_slice::byte_slice(byte_stream&& stream) noexcept
+  byte_slice::byte_slice(byte_stream&& stream, const bool shrink)
     : storage_(nullptr), portion_(stream.data(), stream.size())
   {
-    if (stream.size())
```
perfect-daemon, line 179:
```diff
From cecca7a0fe43044e8f163a748347868f38b1b5e6 Mon Sep 17 00:00:00 2001
From: anon <anon [at] nowhere>
Date: Sat, 20 Feb 2021 19:56:02 +0000
Subject: [PATCH] byte_slice: simplify

---
 contrib/epee/src/byte_slice.cpp | 22 +++++++++-------------
 1 file changed, 9 insertions(+), 13 deletions(-)

diff --git a/contrib/epee/src/byte_slice.cpp b/contrib/epee/src/byte_slice.cpp
index 453b63a4c..2e8a2f820 100644
--- a/contrib/epee/src/byte_slice.cpp
+++ b/contrib/epee/src/byte_slice.cpp
@@ -181,24 +181,20 @@ namespace epee
   byte_slice::byte_slice(byte_stream&& stream, const bool shrink)
     : storage_(nullptr), portion_(stream.data(), stream.size())
   {
-    if (portion_.size())
+    if (stream.size())
     {
-      byte_buffer buf;
       if (shrink && page_size <= stream.available())
       {
-          buf = byte_buffer_resize(stream.take_buffer(), portion_.size());
-          if (!buf)
-            throw std::bad_alloc{};
-          portion_ = {buf.get(), portion_.size()};
+        auto buf = byte_buffer_resize(stream.take_buffer(), portion_.size());
+        if (!buf)
+          throw std::bad_alloc{};
+        portion_ = {buf.get(), portion_.size()};
+        storage_.reset(new (buf.release() - sizeof(raw_byte_slice)) raw_byte_slice{});
       }
-      else // no need to shrink buffer
-        buf = stream.take_buffer();
-
-      std::uint8_t* const data = buf.release() - sizeof(raw_byte_slice);
-      new (data) raw_byte_slice{};
-      storage_.reset(reinterpret_cast<raw_byte_slice*>(data));
+      else
+        storage_.reset(new (stream.take_buffer().release() - sizeof(raw_byte_slice)) raw_byte_slice{});
     }
-    else // empty stream
+    else
       portion_ = nullptr;
   }
 
-- 
2.30.1
```

perfect-daemon (approved):
correct

contrib/epee/src/byte_slice.cpp
vtnerd: How is this an improvement? It ends up duplicating the placement `new` call with no obvious benefit.

perfect-daemon:
```
The above patch can be ignored if there is no obvious benefit.
```

PR #7009, opened 2020-11-11
Change epee binary output from std::stringstream to byte_stream

vtnerd:
The primary motivation for this patch is to remove the internal copy in `std::stringstream::str()` and switch to a `realloc` call for buffer increases. The former is not possible until C++20, and the latter is not possible (yet) with the allocator design of `std::stringstream`. Construction of the object is also faster, and `write` calls have less overhead (no polymorphism or `sentry` objects). In some cases there's still a copy, but these can hopefully be cleaned up in future edits.

Taking glibc as an example - allocation requests over 128KiB are directly `mmap`ed via the kernel. Increasing this memory chunk with `realloc` can be done via `mremap` (on Linux and some BSDs), which either extends the virtual memory space or shuffles around the page table without copying the memory. This obviously helps with manipulating big chunks of memory, which typically happens when returning blocks via p2p or blocks/transactions via RPC. Even with smaller chunks, heap implementations can extend the buffer size without copy if there is free space behind the buffer. 

This results in ~30-70%+ drop in "DOM->binary" wall clock time on a ryzen 3900x. Although this is in some part due to the new aggressive initial buffer sizes, there's still less copying of the data and less overhead on writes. DOM->binary for block responses via p2p should take around ~1ms (at least if you have a beefy CPU).  I expect the performance impact on slower and smaller cache CPUs to a bit more.

There might be some jumps in CPU time due to the syscalls+aggressive buffer requests. These were less than 1 microsecond in total serialization time, but its something that may need tweaking.

Unfortunately, the bulk of the p2p response time is spent on I/O and some other stuff, but this does show in profiles here and there (depending on whether you are pushing blocks, and which blocks being pushed).

### DIFF
--- a/contrib/epee/include/net/abstract_http_client.h
+++ b/contrib/epee/include/net/abstract_http_client.h
@@ -70,7 +70,7 @@ namespace http
     virtual bool connect(std::chrono::milliseconds timeout) = 0;
     virtual bool disconnect() = 0;
     virtual bool is_connected(bool *ssl = NULL) = 0;
-    virtual bool invoke(const boost::string_ref uri, const boost::string_ref method, const std::string& body, std::chrono::milliseconds timeout, const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) = 0;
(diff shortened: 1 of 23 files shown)


moneromooo-monero (approved):

selsta:
@vtnerd please rebase

vtnerd:
Rebased

selsta (approved):

Issue #7004, opened 2020-11-10
Inbound and outbound i2p connections have the same peer ID


MoneroArbo:
With the most recent version of monerod (0.17.1.3) I have been occasionally connected to INC and OUT i2p peers that have the same peer ID. As discussed in #6380, this is going to result in the ability to correlate transactions relayed over i2p with a particular i2p address.

[Output of print_cn](https://i.imgur.com/IMsoDr9.png)

[Output of sync_info](https://i.imgur.com/iTJRIDA.png)

@vtnerd I was suggested to ping you on this one

vtnerd:
I closed the prior issue, and I will work on a fix for this shortly.

moneromooo-monero:
Patch merged.

PR #7017, opened 2020-11-12
Do not use peer_id tracking method over i2p/tor

vtnerd:
This is for [issue 7004](https://github.com/monero-project/monero/issues/7004). The changes the `peer_id` system to be used for IP p2p connections only, and not for I2P/Tor. The fallout:

  * Misconfigurations with `--anonymous-inbound` can result in I2P/Tor self-connections.
  * I2P/Tor could make an outbound connection when an inbound connection to the same node already exists.

The second bullet-point must be an accepted downside, because you cannot have privacy and tracking at the same time.

I decided against a per-connection `peer_id` because it still be used to identify users if an aggressive spy appeared. Take the `peer_id` received from a peer, and try to make some connections to `.i2p` and `.onion` addresses to see which one rejected the connection. This attack cannot be completely removed if the `peer_id` system was still be used for I2P/Tor. Dropping the use of `peer_id` was mitigated by checking for self address connections, leaving only misconfiguration (bullet point 1) as the potential issue.

Testers for this would be welcome too.

### DIFF
--- a/src/p2p/net_node.h
+++ b/src/p2p/net_node.h
@@ -149,7 +149,7 @@ namespace nodetool
     {
       config_t()
         : m_net_config(),
-          m_peer_id(crypto::rand<uint64_t>()),
+          m_peer_id(1),
           m_support_flags(0)
       {}
 
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -127,7 +127,9 @@ namespace nodetool
     if (storage)
       m_peerlist_storage = std::move(*storage);
 
-    m_network_zones[epee::net_utils::zone::public_].m_config.m_support_flags = P2P_SUPPORT_FLAGS;
+    network_zone& public_zone = m_network_zones[epee::net_utils::zone::public_];
+    public_zone.m_config.m_support_flags = P2P_SUPPORT_FLAGS;
+    public_zone.m_config.m_peer_id = crypto::rand<uint64_t>();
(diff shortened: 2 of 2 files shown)


moneromooo-monero:
> The second bullet-point must be an accepted downside, because you cannot have privacy and tracking at the same time.

You can (though not in this PR) by making the peer id a random value per connection, and checking the peer id of incoming connections with the outgoing connections in the process of being made.

vtnerd:
> You can (though not in this PR) by making the peer id a random value per connection, and checking the peer id of incoming connections with the outgoing connections in the process of being made.

I decided against this style approach because tracking should still be possible:

  1. Honest node generates new `peer_id` for outgoing connection to a hidden service (which happens to be a spy node).
  2. Spy node has up to `P2P_DEFAULT_CONNECTION_TIMEOUT` (5secs) to probe hidden services with identical `COMMAND_HANDSHAKE` and same `peer_id`.
  3. Honest node sees same `peer_id`, and terminates connection, notifying spy node that its inbound connection is likely from that node.

It might be possible to get around this issue by having a randomized response to an incoming handshake. My comment about tracking vs privacy was noting that the `peer_id` technique is trying to use a unique value for tracking purposes, making the technique anti-privacy in a sense.

moneromooo-monero (approved):

selsta:
How to test this specifically?

selsta (approved):

PR #7018, opened 2020-11-12
cryptonote_core: dandelion - use local height or median height if syncing

xiphon:


src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -105,8 +105,31 @@ namespace levin
...
+          return 0;
+        }
+
+        const size_t n = remote_heights.size() / 2;
+        std::sort(remote_heights.begin(), remote_heights.end());
+        if (remote_heights.size() % 2 != 0)
+        {
+          return remote_heights[n];
+        }
+        return remote_heights[n-1];
```
moneromooo-monero, line 128: There is a median function in epee.

vtnerd:
What is the rationale for having unverified remote node reporting influencing the behavior at all? A node that is still syncing? And once a node is caught up it seems like you'd never want to use this method for safety purposes.

src/cryptonote_protocol/levin_notify.cpp
xiphon: Don't want to get a mean in `remote_heights.size() % 2 == 0` case, but using `remote_heights[n-1]` instead.

xiphon:
Updated

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -104,7 +104,7 @@ namespace cryptonote
     bool get_payload_sync_data(CORE_SYNC_DATA& hshd);
     bool on_callback(cryptonote_connection_context& context);
     t_core& get_core(){return m_core;}
-    bool is_synchronized(){return m_synchronized;}
+    virtual bool is_synchronized() const override {return !no_sync() && m_synchronized;}
```
vtnerd, line 107: I didn't comment on this in the 0.17 review - it can be updated later (or just left as-is). I would mark the function as `final` too. The calls within the class can be de-virtualized+inlined easliy with this change.

vtnerd (approved):

src/cryptonote_protocol/cryptonote_protocol_handler.h
vtnerd: Actually, I think the other `virtual`+`override` functions in `cryptonote::core` can be marked as a `final` too.

xiphon: Updated both PRs

selsta (approved):

PR #7021, opened 2020-11-15
Fix timeout checks for forwarded and Dandelion++ stem txes

vtnerd:
Currently, the txpool DB is only checked every 2 minutes for relayable transactions. This causes longer than expected delays for Dandelion++ embargo timeouts, and forwarded transactions (from I2P/Tor). This changes the code so that the `cryptonore_core::core::on_idle()` function always checks the txpool (currently once a second). The txpool now has an atomic value for skipping all locking and DB reads until a specific amount of time has past.

This currently does **not** change the DB check time if a transaction transitions from stem to fluff (i.e. not blackholed), so there is some slight inefficiency on the locking and DB reads. Changing this would be a little more complex, and should be less important since a node is only doing a subset of total transactions via stem/i2p/tor.

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1725,7 +1725,7 @@ namespace cryptonote
       m_starter_message_showed = true;
     }
 
-    m_txpool_auto_relayer.do_call(boost::bind(&core::relay_txpool_transactions, this));
+    relay_txpool_transactions(); // txpool handles periodic DB checking
     m_check_updates_interval.do_call(boost::bind(&core::check_updates, this));
     m_check_disk_space_interval.do_call(boost::bind(&core::check_disk_space, this));
     m_block_rate_interval.do_call(boost::bind(&core::check_block_rate, this));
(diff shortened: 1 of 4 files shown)


hyc:
Fixed wakeup intervals like this will be extremely wasteful, particularly on mobile devices. Would be better to just use a timestamped queue, and only wakeup when the next entry in the queue is due.

vtnerd:
> Fixed wakeup intervals like this will be extremely wasteful, particularly on mobile devices. Would be better to just use a timestamped queue, and only wakeup when the next entry in the queue is due.

I don't disagree, but the code is already waking up on fixed intervals for a series of checks. I sought to re-arrange this a bit for the txpool, but gave-up because the `on_idle()` function was already executing every second anyway. So I tried to keep the work done every second to a minimum.

Switching to an on-demand setup seems possible (or so it appears!), but is likely worth its own PR - a number of calls have to be updated to provide their next expected wakeup time and passed up a chain. A bunch of independent syscalls to get the current time can be removed too with some minimal effort. Given the nature of P2P, I'm not sure how much battery this would save, but its certainly a move in a better direction.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1725,7 +1725,7 @@ namespace cryptonote
       m_starter_message_showed = true;
     }
 
-    m_txpool_auto_relayer.do_call(boost::bind(&core::relay_txpool_transactions, this));
+    relay_txpool_transactions(); // txpool handles periodic DB checking
```
vtnerd, line 1728: There's a `m_mempool.on_idle()` call below I didn't hjijack, but perhaps should've Pointing this out for reviewers.

moneromooo-monero (approved):

PR #7038, opened 2020-11-23
build: prepare v0.17.1.5

selsta:


dEBRUYNE-1:
Perhaps it is worthwhile to skip to v0.17.1.5 to ensure the CLI and GUI are back in sync again (with respect to version number)?

selsta:
Good point, was not intentional

selsta:
updated

moneromooo-monero (approved):

PR #7039, opened 2020-11-23
[release-v0.17] cryptonote_core: dandelion - use local height or median height if syncing

xiphon:


vtnerd:
My comment had no response in PR #7018 so I'll retry here - @hyc -

This is taking the median of all reported heights from incoming+outgoing nodes. This still allows an attacker (at least temporarily) to manipulate the Dandelion++ routing selection of a particular IP by filling up the incoming peers table.

Why the median of incoming+outgoing nodes, even after the local node has synced?

xiphon:
> This is taking the median of all reported heights from incoming+outgoing peers. This still allows an attacker (at least temporarily) to manipulate the Dandelion++ routing selection of a particular IP by filling up the incoming peers table.

Will update to use only outgoing peers.

> Why the median of incoming+outgoing nodes, even after the local node has synced?

Will have a look.

xiphon:
Updated

vtnerd (approved):

selsta (approved):

PR #7051, opened 2020-12-01
[Design] Abstract target device_abstr, extracted from device target

mj-xmr:
Solves [issue 7053](https://github.com/monero-project/monero/issues/7053).

**TL;DR: Cleaner device architecture will allow the hardware wallet vendors to integrate with Monero, without needing to merge their implementations into MoneroCore.**

Apologies for the PR being somewhat overweight.

## Problem
The the `libdevice` depends on `libcryptonote_basic` through `cryptonote::get_transaction_prefix_hash(cryptonote::transaction_prefix const&, crypto::hash&)`, used in `device_default.cpp`, while and `libcryptonote_basic` depends on `libdevice`, via the file `crypronote_basic.h`, where the `device.hpp` is included and used.

## Solution
Extracted `device_abstr` and using it whenever possible within the libraries. The implementation target `device` is only fed to the client applications.

### Current design
Abstract device:
image

Concrete implementation is used only where necessary:
image

## Extra work
I took the opportunity to remove the inclusion of the `device.hpp` from the `cryptonote_basic.hpp`, limiting the dependencies a bit. Compilation time reduction: about 3-4%. Compare with [master here](http://enjo.hopto.org/pub/monero/).

```
**** Time summary:
Compilation (814 times):
  Parsing (frontend):         1718.3 s
  Codegen & opts (backend):   1721.3 s
```

## Future work
This is out of scope for this PR, but Ideally, after only the client apps have the possibility to instantiate concrete devices, they should inject them into the library code via Abstract Factory. In our case, we wouldn't see a direct call to device creation anymore, like in `cryptonote_basic::account.cpp` through `hw::get_device`, which is a singleton, calling a concrete factory, but the function would receive the `AbstractDeviceFactory` via parameter from the client code, which instantiates a concrete factory, derived from `AbstractDeviceFactory`. This will allow for further abstraction of the client devices, not limiting Monero to concrete, predefined manufacturers in our own code.

```
src/cryptonote_basic/account.cpp:

void account_base::create_from_device(const std::string &device_name)
  {
    hw::device &hwdev =  hw::get_device(device_name);
    hwdev.set_name(device_name);
    create_from_device(hwdev);
  }
```

Active work time: 5h


vtnerd:
There appears to be a mis-match between description and behavior. You can already compile non-static releases by setting `-DBUILD_SHARED_LIBS=ON` during the cmake configuration stage. Instead, this PR changes the default behavior to shared libraries for release builds. I don't see a compelling reason for this because a significant number of function calls become indirect lookups - each directory is its own shared library. The current behavior seems like what most expect currently, unless things are re-arranged for a `libmonero`, etc.

mj-xmr:
> There appears to be a mis-match between description and behavior. (...)

Yes, this is true. It's a change of the default behavior, and this can be achieved with the flag BUILD_SHARED_LIBS you mentioned. The problem is, that there are no Makefile targets in the root directory, that would trigger switch this flag on for release targets.
Other than that, maybe it would be better if I just document somewhere how to change this flag via cmake.

mj-xmr:
In case you wonder, I'm trying to reproduce the cyclical dependency problem here.

src/cryptonote_basic/account.h
```diff
@@ -33,17 +33,20 @@
...
   struct account_keys
   {
+      account_keys();
+      
     account_public_address m_account_address;
     crypto::secret_key   m_spend_secret_key;
     crypto::secret_key   m_view_secret_key;
     std::vector<crypto::secret_key> m_multisig_keys;
-    hw::device *m_device = &hw::get_device("default");
+    hw::device *m_device = nullptr;
```
mj-xmr, line 49: Moved to account_keys constructor.

mj-xmr:
The core tests are failing. I'm expecting the problem to be the singleton, holding the device, being doubled in the library code, as well as in the client applications. At this time there's no way back, but introduce the AbstractFactory.

```
2021-02-14 16:18:34.413	E REPORT:
2021-02-14 16:18:34.413	E   Test run: 148
2021-02-14 16:18:34.413	E   Failures: 1
2021-02-14 16:18:34.413	E FAILED TESTS:
2021-02-14 16:18:34.413	E   one_block
```

PR #7054, opened 2020-12-02
A couple P2P network layer improvements

moneromooo-monero:


src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -2564,6 +2603,14 @@ skip:
   }
   //------------------------------------------------------------------------------------------------------------------------
   template<class t_core>
+  void t_cryptonote_protocol_handler<t_core>::hit_score(cryptonote_connection_context &context, unsigned int score)
+  {
+    context.m_score -= score;
```
iamamyth, line 2608: This does not appear to be safe against overflow, although that possibility seems rather unlikely with current rules. Also, it seems a bit odd to specify widths for the p2p score field, but not for the increments.

moneromooo-monero: What do you mean by widths ?

selsta:
Tested all protocol commits and they seem to be working as specified in the title.

src/cryptonote_basic/connection_context.h
```diff
@@ -66,7 +67,8 @@ namespace cryptonote
     uint16_t m_rpc_port;
     uint32_t m_rpc_credits_per_hash;
     bool m_anchor;
-    //size_t m_score;  TODO: add score calculations
+    int32_t m_score;
+    int m_expect_response;
```
xiphon, line 71: Does this limits the protocol now to one request at a time? I would use a set instead.

xiphon: Or even a map/vector so you can send multiple requests of the same type.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
iamamyth: By widths, I mean `int32` for the field vs. using a native unsigned int for the increment. If the goal is to clarify the mathematics with explicit integer bit-widths, best to use those everywhere; if the goal is to avoid false dependencies by using native integer types, that works, too; either way, it seems superior to be consistent.

src/cryptonote_basic/connection_context.h
moneromooo-monero: I don't think we pipeline requests anywhere.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1666,24 +1693,46 @@ skip:
...
+
+    for (const auto &e: idle_peers)
+    {
+      const auto &uuid = e.first;
+      if (!m_p2p->for_connection(uuid, [&](cryptonote_connection_context& ctx, nodetool::peerid_type peer_id, uint32_t f)->bool{
+        MINFO(ctx << "dropping idle peer with negative score");
+        drop_connection_with_score(ctx, e.second, false);
+        return true;
+      }))
+        MDEBUG("Failed to find peer we wanted to drop");
```
vtnerd, line 1733: Is this useful? I mean, wouldn't this imply that the peer was already dropped somehow?

src/cryptonote_basic/connection_context.h
vtnerd: Ok, I've thought about this some more. Legit responders should always do in-order responses. This is a side-effect of how the TCP server is designed more than anything. The requesting side does not enforce any order, and can probably issue multiple pipelined requests before a prior response. This is the difference between an `invoke` call and a `notification` call. The chain request and block requests are actually notifications, and the requester doesn't explicitly wait for a response which would block the thread (in the epee design).

moneromooo-monero: This system only applies to get objects and get chain, not to things like new txes and new blocks. The former aren't pipelined, but a new tx or block in the middle will not break the check (unless you found a case where it does).

contrib/epee/src/net_utils_base.cpp
```diff
@@ -83,6 +94,27 @@ namespace epee { namespace net_utils
...
+		else if (this_id == ipv6_network_address::get_type_id() && other.get_type_id() == ipv4_network_address::get_type_id())
+		{
+			const boost::asio::ip::address_v6 &actual_ip = this->as<const epee::net_utils::ipv6_network_address>().ip();
+			if (actual_ip.is_v4_mapped())
+			{
+				const uint32_t v4ip = make_address_v4_from_v6(actual_ip);
+				return other.is_same_host(ipv4_network_address(v4ip, 0));
+			}
+		}
 		if (typeid(*self_) != typeid(*other_self)) return false;
```
SChernykh, line 118: This check is redundant now, I think you can just return false here

SChernykh (changes requested):

selsta (approved):

src/cryptonote_basic/connection_context.h
vtnerd: With regards to pipelining chain or block requests - a request for block or chain is typically in response to an unsolicited notification message. There's no explicit check to prevent another chain or block request while one is outstanding, but instead appears to be enforced through a combination of checks in `core` and PoW. Not exactly ideal because this potentially makes things more brittle. Provided this is the intent, I don't have any other recommendations at the moment.

iamamyth: It might be advisable to at some point explicitly state the design intent here via code comment (which seems to be, perhaps, preventing multiple in-flight requests). From my perspective, disallowing multiple in-flight requests tends to be the correct strategy unless there's a very compelling case for pipelining, as pipelining introduces a lot of subtle complexity relating to resource usage (how many pipelined requests are permitted simultaneously? what state do you need to maintain per-connection? etc). Intuitively, as the core data structure is a list of blocks, a pipeline's utility seems dubious (e.g. requesting successor blocks before validating the current tip could end up wasting tons of cycles trying to compute PoW for bogus successors).

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1029,6 +1042,7 @@ namespace cryptonote
       drop_connection(context, false, false);
       return 1;
     }
+    context.m_last_request_time = boost::posix_time::microsec_clock::universal_time();
```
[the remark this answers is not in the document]
moneromooo-monero: Why set it here ? It does not expect a reply here AFAICT.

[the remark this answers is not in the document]
moneromooo-monero: Good point about post_notify, I'll do that later though.

PR #7057, opened 2020-12-03
Provide last error from http client to UI

vtnerd:
This stores the last `std::error_code` while doing HTTP client I/O to make it available to UIs. With Boost 1.65+ this provides more useful messages with the CLI wallet, and hopefully GUI wallet too. The primary purpose is for an (shortly) upcoming DANE/TLSA patch that uses DNSSEC for the "autodetect" SSL mode with wallets. I broke this into its own patch because its useful even if the DANE/TLSA patch is rejected or reverted.

### DIFF
--- a/contrib/epee/include/net/abstract_http_client.h
+++ b/contrib/epee/include/net/abstract_http_client.h
@@ -26,6 +26,7 @@
 #pragma once
 
 #include <string>
+#include <system_error>
 #include <boost/optional/optional.hpp>
 #include "http_auth.h"
 #include "net/net_ssl.h"
@@ -70,6 +71,7 @@ namespace http
     virtual bool connect(std::chrono::milliseconds timeout) = 0;
     virtual bool disconnect() = 0;
     virtual bool is_connected(bool *ssl = NULL) = 0;
+    virtual std::error_code last_error();
     virtual bool invoke(const boost::string_ref uri, const boost::string_ref method, const std::string& body, std::chrono::milliseconds timeout, const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) = 0;
     virtual bool invoke_get(const boost::string_ref uri, std::chrono::milliseconds timeout, const std::string& body = std::string(), const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) = 0;
     virtual bool invoke_post(const boost::string_ref uri, const std::string& body, std::chrono::milliseconds timeout, const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) = 0;
--- a/contrib/epee/include/net/connection_basic.hpp
+++ b/contrib/epee/include/net/connection_basic.hpp
@@ -135,7 +135,7 @@ class connection_basic { // not-templated base class for rapid developmet of som
 		bool handshake(boost::asio::ssl::stream_base::handshake_type type)
 		{
 			//m_state != nullptr verified in constructor
-			return m_state->ssl_options().handshake(socket_, type);
+			return !bool(m_state->ssl_options().handshake(socket_, type));
 		}
 
 		template<typename MutableBufferSequence, typename ReadHandler>
--- a/contrib/epee/include/net/net_helper.h
+++ b/contrib/epee/include/net/net_helper.h
@@ -45,6 +45,9 @@
 #include <boost/interprocess/detail/atomic.hpp>
 #include <boost/system/error_code.hpp>
 #include <functional>
+#include <system_error>
+#include <utility>
+#include <vector>
 #include "net/net_utils_base.h"
 #include "net/net_ssl.h"
 #include "misc_language.h"
@@ -59,6 +62,8 @@
 
 namespace epee
 {
+	std::error_code convert_error_code(boost::system::error_code error);
+
 namespace net_utils
 {
 	struct direct_connect
@@ -70,15 +75,6 @@ namespace net_utils
 
   class blocked_mode_client
 	{
-		enum try_connect_result_t
-		{
-			CONNECT_SUCCESS,
-			CONNECT_FAILURE,
-			CONNECT_NO_SSL,
-		};
-
-		
-		
 				struct handler_obj
 				{
 					handler_obj(boost::system::error_code& error,	size_t& bytes_transferred):ref_error(error), ref_bytes_transferred(bytes_transferred)
@@ -98,6 +94,15 @@ namespace net_utils
 					}
 				};
 
+		static std::error_code guarantee_error(const boost::system::error_code error)
+		{
+			if (error)
+				return convert_error_code(error);
+			return make_error_code(std::errc::not_connected);
+		}
+
+		boost::system::error_code try_connect(const std::string& addr, const std::string& port, std::chrono::milliseconds timeout);
+
 	public:
 		inline
 			blocked_mode_client() :
@@ -106,8 +111,7 @@ namespace net_utils
 				m_ssl_socket(new boost::asio::ssl::stream<boost::asio::ip::tcp::socket>(m_io_service, m_ctx)),
 				m_connector(direct_connect{}),
 				m_ssl_options(epee::net_utils::ssl_support_t::e_ssl_support_autodetect),
-				m_initialized(true),
-				m_connected(false),
+				m_stream_error(make_error_code(std::errc::not_connected)),
 				m_deadline(m_io_service, std::chrono::steady_clock::time_point::max()),
 				m_shutdowned(0),
 				m_bytes_sent(0),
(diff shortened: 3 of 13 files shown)


contrib/epee/include/net/abstract_http_client.h
```diff
@@ -70,6 +71,7 @@ namespace http
     virtual bool connect(std::chrono::milliseconds timeout) = 0;
     virtual bool disconnect() = 0;
     virtual bool is_connected(bool *ssl = NULL) = 0;
+    virtual std::error_code last_error();
```
vtnerd, line 74: @woodser a new method was added to the abtract http interface. The default behavior is to return a generic `not_connected` code when not connected, but can be extended to return more useful messages.

selsta:
Our release binaries use boost 1.64. Would it be useful to bump it? If yes, which version would you suggest?

src/wallet/api/wallet.cpp
```diff
@@ -2062,20 +2062,20 @@ bool WalletImpl::verifyMessageWithPublicKey(const std::string &message, const st
 
 bool WalletImpl::connectToDaemon()
 {
-    bool result = m_wallet->check_connection(NULL, NULL, DEFAULT_CONNECTION_TIMEOUT_MILLIS);
+    const expect<void> result = m_wallet->check_connection(NULL, NULL, DEFAULT_CONNECTION_TIMEOUT_MILLIS);
```
vtnerd, line 2065: @xiphon @selsta 

This patch stores additional information in the wallet API error message. Sometimes the message from OpenSSL in particular is a little funky, but the `error_code` typically provides more information as to why the `check_connection` attempt failed. Is there a better place for this?

vtnerd:
> Our release binaries use boost 1.64. Would it be useful to bump it? If yes, which version would you suggest?

Crap. I should've checked before doing this work. Its useful to get this fixed because the official release builds will not have useful messages displayed in the UI.

It should be possible to re-write the patch around `boost::system::error_code` so that upgrading is unnecessary. In hindsight, it may have been a mistake to use `std::error_code` previously given our reliance on Boost (and ASIO in particular). And Boost 1.69+ can have `constexpr` `error_category` objects. This may incur the wraith of some reviewers because it'll be result in yet another PR by me to convert things.

OTOH, Boost provides a `boost::system::error_code` -> `std::error_code` conversion (that's kind of clunky IMO) in Boost 1.65+, and Boost provides no support for older releases.

@selsta do you know how Boost 1.64 was selected previously? Just because it was the latest release?

selsta:
AFAIK there is no reason for Boost 1.64. We can upgrade to Boost 1.68 without any changes (#6693), if we want a even newer version some changes to the depends build system are necessary.

contrib/epee/include/net/abstract_http_client.h
woodser: What would be returned when connected and there is no error?

Could the error not be thrown from the other methods for standard exception handling?

contrib/epee/include/net/connection_basic.hpp
```diff
@@ -135,7 +135,7 @@ class connection_basic { // not-templated base class for rapid developmet of som
 		bool handshake(boost::asio::ssl::stream_base::handshake_type type)
 		{
 			//m_state != nullptr verified in constructor
-			return m_state->ssl_options().handshake(socket_, type);
+			return !bool(m_state->ssl_options().handshake(socket_, type));
```
moneromooo-monero, line 138: Why ! ?

contrib/epee/include/net/net_helper.h
```diff
@@ -155,97 +159,10 @@ namespace net_utils
...
-						}
-					}
-					return CONNECT_SUCCESS;
-				}else
-				{
-					MWARNING("Some problems at connect, expected open socket");
-					return CONNECT_FAILURE;
-				}
 
-		}
```
moneromooo-monero, line 206: That seems to do a lot more than exposing last error.

contrib/epee/include/net/abstract_http_client.h
vtnerd: > What would be returned when connected and there is no error?

A default constructed empty `error_code` indicating no failure.

> Could the error not be thrown from the other methods for standard exception handling?

From where exactly? The `connect` function? The helper `invoke` functions?

The `connect` method uses `bool` for status and catches all exceptions, the control flow for every call site would have to be changed and inspected. `invoke` uses bool too, but doesn't try to catch exceptions.

contrib/epee/include/net/connection_basic.hpp
vtnerd: `handshake` returns a `boost::system::error_code` which has inverted `bool` behavior unfortunately. Wrapping it in an `expect<void>` gives expected `bool` conversion, but that class would have to be moved to the `epee` playground. I could do `bool`+out parameter for `error_code` here ?

contrib/epee/include/net/net_helper.h
vtnerd: Originally this changed a bunch of lines for DANE/TLSA. This can be rolled back somewhat for this more limited PR, but there are still a number of changes for `boost::system::error_code` return from `ssl_options_t::handshake`.

There was some redundant checks that got removed, and it seemed like an opportunity to just move a chunk of code to the cpp. The amount moved wasn't huge in the end.

contrib/epee/include/net/connection_basic.hpp
moneromooo-monero: No, as long as it's intended it's fine.

contrib/epee/include/net/net_helper.h
moneromooo-monero: Maybe it hasn't changed, but it makes reviewing a lot more annoying since manually reading two bits of source to see they're indeed the same is a pita. But hey, they're not *that* large I guess,

contrib/epee/src/net_helper.cpp
```diff
@@ -78,6 +118,85 @@ namespace net_utils
...
+			m_deadline.expires_at(std::chrono::steady_clock::time_point::max());
+			// SSL Options
+			if (m_ssl_options.support == ssl_support_t::e_ssl_support_enabled || m_ssl_options.support == ssl_support_t::e_ssl_support_autodetect)
+			{
+				if ((error = m_ssl_options.handshake(*m_ssl_socket, boost::asio::ssl::stream_base::client, addr, timeout)))
+				{
+					MWARNING("Failed to establish SSL connection");
+					boost::system::error_code ignored_ec;
+					m_ssl_socket->next_layer().shutdown(boost::asio::ip::tcp::socket::shutdown_both, ignored_ec);
+					m_ssl_socket->next_layer().close();
```
moneromooo-monero, line 149: Do you need to clear m_stream_error here, if not used only by connect ?

src/simplewallet/simplewallet.cpp
```diff
@@ -4760,10 +4760,11 @@ bool simple_wallet::try_connect_to_daemon(bool silent, uint32_t* version)
   uint32_t version_ = 0;
   if (!version)
     version = &version_;
-  if (!m_wallet->check_connection(version))
+  expect<void> status{};
+  if (!(status = m_wallet->check_connection(version)))
```
moneromooo-monero, line 4764: That is *really* confusing. Above, you inverted the return value check when swithing from bool to expect. Here and below, not.

contrib/epee/include/net/net_helper.h
vtnerd: I reverted the move to the cpp, to make it easier to read.

src/simplewallet/simplewallet.cpp
vtnerd: No longer using `expect<void>`, everything is `boost::system::error_code` with the expected returns.

contrib/epee/src/net_helper.cpp
vtnerd: I do not believe `m_stream_error` needs to be reset here. `try_connect` is called from `connect` which always resets the value based on the results from `try_connect`.

vtnerd:
@0xFFFC0000 @jeffro256 @j-berman This has been around for years, and will help with a potential DANE/TSLA or trust-on-first patch. It simply stores the last `error_code` in the http client, so that it can be displayed to the user UI.

vtnerd:
Forgot to add, I made the changes moo suggested years ago, and rebased.

src/wallet/api/wallet.cpp
```diff
@@ -2095,21 +2095,21 @@ bool WalletImpl::verifyMessageWithPublicKey(const std::string &message, const st
...
-    return result;
+    return !err;
 }
 
 Wallet::ConnectionStatus WalletImpl::connected() const
 {
     uint32_t version = 0;
     bool wallet_is_outdated = false, daemon_is_outdated = false;
-    m_is_connected = m_wallet->check_connection(&version, NULL, DEFAULT_CONNECTION_TIMEOUT_MILLIS, &wallet_is_outdated, &daemon_is_outdated);
+    m_is_connected = bool(m_wallet->check_connection(&version, NULL, DEFAULT_CONNECTION_TIMEOUT_MILLIS, &wallet_is_outdated, &daemon_is_outdated));
```
jeffro256, line 2112: Shouldn't this be `!bool()` since error code 0 means success?

jeffro256 (commented):
Looks like good changes to me, just one note:

0xFFFC0000:
LGTM too.

src/wallet/api/wallet.cpp
vtnerd: Good catch. Initially this returned `expect<void>`, but decided to keep things consistent with `error_code`.

vtnerd:
Force pushed a fix to `api/wallet.cpp`.

jeffro256 (approved):

0xFFFC0000 (approved):
Finally today I had some time to run some testing on this. Thanks @vtnerd for this PR.

vtnerd:
@selsta can this get in the merge queue?

vtnerd:
@jeffro256 @0xFFFC0000 @selsta I made a mistake in **not** updating all of the `check_connection` calls. I instead decided against inverting the call behavior, and instead added a `boost::system::error_code*` function parameter. Let's see if this goes green (it should).

So please look/inspect again!

PR #7058, opened 2020-12-03
wallet2: set propagation timeout to current max timeout

selsta:
Due to outdated nodes still running the old Dandelion++ parameter (and the bug fixed in #7021) it can take up to

```
(173 * 3 / 2 + 120 + 120) = 499.5 seconds
```

before a transaction ends up in the public mempool. I added half a second for network latency.

After the v15 hardfork every node will run the same Dandelion++ parameters so we can set it back to the current value.


moneromooo-monero:
This is probabilistic, so the half second is just noise :)

moneromooo-monero (approved):

PR #7059, opened 2020-12-03
[release-v0.17] wallet2: set propagation timeout to current max timeout

selsta:
Release branch equivalent of #7058


Snipa22 (approved):
Identical change to the approved version in #7058 7058 just aimed at branch.

PR #7062, opened 2020-12-03
crypto: fix non zero scalar being 0 after reducing

moneromooo-monero:


src/crypto/crypto.cpp
```diff
@@ -128,8 +128,8 @@ namespace crypto {
     do
     {
       generate_random_bytes_thread_safe(32, bytes);
-    } while (!sc_isnonzero(bytes) && !less32(bytes, limit)); // should be good about 15/16 of the time
-    sc_reduce32(bytes);
+      sc_reduce32(bytes);
+    } while (!sc_isnonzero(bytes) || !less32(bytes, limit)); // should be good about 15/16 of the time
```
jagerman, line 132: This doesn't fix the bias: `less32(bytes, limit)` will always be true because you have already done a `mod L` on the value.

jagerman: (Fixing the bias will also break the values in tests/crypto/tests.txt because roughly 1/16 of rvs should get rejected, but this change didn't affect them at all).

jagerman: I think it needs to look something like this (though it's getting a bit gross):
```C++
    do
    {
      do
      {
        generate_random_bytes_thread_safe(32, bytes);
      } while (!less32(bytes, limit)); // should be good about 15/16 of the time
      sc_reduce32(bytes);
    } while (!sc_isnonzero(bytes));
```

moneromooo-monero: Ugh. Stupid. Thanks.

moneromooo-monero: That looks simpler, the reduce is not needed since it's mooted by the test:
```
    do
    {
      generate_random_bytes_thread_safe(32, bytes);
    } while (!sc_isnonzero(bytes) || !less32(bytes, limit)); // should be good about 15/16 of the time
```
Looks OK to you ?

moneromooo-monero: Oh wait, it's got the problem you reported :)

moneromooo-monero: And less32 is not a mod check, nvm

moneromooo-monero: er, it is I think. If less32 returns true, mod will do nothing, I think those two values aew the same.

moneromooo-monero: It's not, I found L, it's not that. That must be a multiple of L. Oh well.

moneromooo-monero: How about this then:
```
    while(1)
    {
      generate_random_bytes_thread_safe(32, bytes);
      if (!less32(bytes, limit))
        continue;
      sc_reduce32(bytes);
      if (sc_isnonzero(bytes))
        break;
    }
  }
```

jagerman: > It's not, I found L, it's not that. That must be a multiple of L. Oh well.

Yeah, limit = 15L.  (The comment there could be improved a bit to make that more obvious, I struggled with that first time reading it).

> How about this then: ...

Yeah, that should do it.

moneromooo-monero: Fair point, I'll make it clearer. Thanks.

jagerman:
Note that this changes the RNG sequence (since the code is now properly discarding ~1/16 of draws, on average) which breaks `tests/crypto/cncrypto-test`.

SChernykh (approved):

PR #7066, opened 2020-12-03
wallet_api: consider daemon synchronized if it is within negligible distance of target height

tobtoht:
This ensures the wallet will remain refreshed and able to send and receive transactions, even if the daemon is slightly below the reported target height.

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -1075,7 +1075,10 @@ bool WalletImpl::daemonSynced() const
     if(connected() == Wallet::ConnectionStatus_Disconnected)
         return false;
     uint64_t blockChainHeight = daemonBlockChainHeight();
-    return (blockChainHeight >= daemonBlockChainTargetHeight() && blockChainHeight > 1);
+    // Consider daemon synchronized if blockChainHeight is within negligible distance of target height.
+    // Mitigates an issue where bad peers can trick the daemon into reporting a slightly overestimated
+    // target height which would cause wallets to stop refreshing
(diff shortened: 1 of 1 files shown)


xiphon:
I wouldn't do this on the wallet side. Also the issue should be fixed with the recent daemon patches.

selsta:
CLI already does this, so it would make sense to add this to GUI too. It obviously is not a fix but it will help users who connect to outdated nodes.

xiphon:
> CLI already does this, so it would make sense to add this to GUI too. It obviously is not a fix but it will help users who connect to outdated nodes.

I wouldn't touch the wallet api. It sounds okay to add it to the GUI.

tobtoht:
No way to add it to the GUI as far as I can tell, the refresh check happens [here](https://github.com/monero-project/monero/blob/1639d0ca7f054e1cf6d0445732b99ac61fc320a2/src/wallet/api/wallet.cpp#L2174).

xiphon:
> No way to add it to the GUI as far as I can tell, the refresh check happens here.

I see. Still wouldn't touch the wallet api.

From my point of view it's the daemon issue, monerod should report correct data instead. Fixing the daemon resolves current (*nonexistent* IMO) wallet issue.

Anyway, others may have different opinion on this.

selsta:
If we merge this I would recommend reverting it in a future release again.

tobtoht:
>If we merge this I would recommend reverting it in a future release again.

I think this is the best approach. This issue is causing a lot of support burden at the moment.

PR #7070, opened 2020-12-04
p2p: fix race condition accessing a deleted context

moneromooo-monero:


selsta (approved):

xiphon (approved):

src/p2p/net_node.inl
```diff
@@ -237,7 +237,8 @@ namespace nodetool
       limit = std::numeric_limits<time_t>::max();
     else
       limit = now + seconds;
-    m_blocked_hosts[addr.host_str()] = limit;
+    const std::string host_str = addr.host_str();
```
vtnerd, line 240: `addr` can be taken by value to fix this. Consider doing that in addition to the other changes since it guards against future issues. The penalty is an atomic increment and 16-byte `memcpy`.

vtnerd (approved):

selsta (approved):

xiphon (approved):

PR #7072, opened 2020-12-04
p2p: drop peers that spam peer lists

moneromooo-monero:
There's always some people who just want to abuse things

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -2051,6 +2051,11 @@ namespace nodetool
   template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::handle_remote_peerlist(const std::vector<peerlist_entry>& peerlist, const epee::net_utils::connection_context_base& context)
   {
+    if (peerlist.size() > P2P_MAX_PEERS_IN_HANDSHAKE)
+    {
+      MWARNING(context << "peer sent " << peerlist.size() << " peers, considered spamming");
+      return false;
+    }
     std::vector<peerlist_entry> peerlist_ = peerlist;
     if(!sanitize_peerlist(peerlist_))
       return false;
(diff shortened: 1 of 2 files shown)


selsta (approved):

src/p2p/net_node.inl
```diff
@@ -2051,6 +2051,11 @@ namespace nodetool
   template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::handle_remote_peerlist(const std::vector<peerlist_entry>& peerlist, const epee::net_utils::connection_context_base& context)
   {
+    if (peerlist.size() > P2P_DEFAULT_PEERS_IN_HANDSHAKE)
```
xiphon, line 2054: Breaks the current protocol. It doesn't have a limit currently.
Should handle first `P2P_DEFAULT_PEERS_IN_HANDSHAKE` items instead of dropping a connection.

Either i would introduce ``P2P_MAX_PEERS_IN_HANDSHAKE` limit and do a protocol breaking change.

selsta (approved):

src/p2p/net_node.inl
vtnerd: This would only apply to someone that changed default configuration and compiled locally ? Although compatibility would be simply doing a resize to truncate.

xiphon (approved):

src/p2p/net_node.inl
xiphon: > Although compatibility would be simply doing a resize to truncate.

Yeah, would prefer doing this in order to handle only first `P2P_DEFAULT_PEERS_IN_HANDSHAKE` items as it's trivial and backward-compatible change.

moneromooo-monero: It would not incentivize people from not spamming though. They could still send massive amounts of data. I don't like it.

iamamyth: With this change, the variable name is misleading, it's no longer a default, but a requirement. I really don't care whether the name changes or not, but the definition should at least have a comment attached to clarify what it actually means.

PR #7075, opened 2020-12-04
A couple minor tweaks to prevent future screwups

moneromooo-monero:


src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -286,7 +286,6 @@ namespace levin
           connection_count(0),
           flush_callbacks(0),
           nzone(zone),
-          is_public(is_public),
```
hyc, line 289: Looks like this got added in a merge error due to dff1d8067c38ecebce9b5d7a7a3d5edc54669b33, probably wasn't rebased correctly.

selsta (approved):

src/p2p/net_node.inl
```diff
@@ -799,14 +799,16 @@ namespace nodetool
           "4pixvbejrvihnkxmduo2agsnmc3rrulrqc7s3cbwwrep6h6hrzsibeqd.onion:18083"
         };
       }
+      break;
```
vtnerd, line 802: These should be `return {};` to indicate a valid zone value but no available seed nodes.

selsta (approved):

vtnerd (approved):

PR #7084, opened 2020-12-05
crypto: fix constant randomx exceptions in large page allocator

moneromooo-monero:
If allocating large pages fails, we don't try again.
This has the obvious drawback of not being able to use large pages
if they fail once.


ZRFLDF:
this large pages needs to be enabled manually on linux? 
https://github.com/xmrig/xmrig/blob/dev/scripts/enable_1gb_pages.sh   Is that OK?

moneromooo-monero:
Depends on your OS/arch. This looks plausible. hugepages-1048576kB may not exist on your system.

SChernykh (approved):

selsta (approved):

Issue #7078, opened 2020-12-05
P2P Encryption+Authentication Discussion


vtnerd:
My apologies to @moneromooo-monero  for previously pushing back against TLS of p2p links. As outlined below, encryption of p2p connections is funky ...

I've got a massive writeup on thoughts about encryption of the p2p links, including quasi-specifications, but I will try to keep thing short so that I get some feedback first.

Adding encryption and authentication to p2p increases the attack surface, and increases the value of a zero-day in the Monero p2p code. A zero-day on a p2p network allows an attacker to move through multiple machines with a difficult to audit trail. Eclipse attacks cannot be prevented either - an attacker can hijack TCP sessions, and the local node doesn't know whether a MitM is occurring or whether the remote node has changed pubkeys (dynamic IP or settings change). An attacker can also drop all packets until an IP is dropped from the peerlist (assume peer is permanently offline). Authentication also provides a mechanism to track a specific node over time via the pubkey. There is also more overhead for initiating an encrypted link, so auditing the incoming connections path is needed too.

However, encryption stops passive surveillance of p2p communications. An ISP cannot just sit idle and monitor transaction flow to determine source. The TCP stream must be hijacked to inject different pubkeys into the channel. Once the connection is established, nothing can be observed unless the attacker drops packets to force a new connection. The primary benefit of authentication is that zero round-trip encryption can be done with the noise protocol (the initial handshake message can be encrypted without waiting for server response).

At a high-level the options are:
 * Change nothing
 * Use TLS 1.3 with/without authentication
 * Use Noise Protocol framework with/without authentication

TLS 1.3 requires 1 round-trip to initiate connection, whereas the Noise protocol can be done in 0 round-trips in certain modes. TLS 1.3 requires ~21 bytes overhead per 2^14 bytes in a p2p message, whereas Noise protocol can be done in 16 bytes overhead for p2p messages of any size. The Noise protocol should have a smaller overall attack surface, with the downside that the Monero project has to maintain/test a bit more code to get it working (risk of new code).

My suggestions for Noise Protocol are `Noise_NK` if using authentication, and `Noise_NN` if not doing authentication. The rationale is a big longer, and the focus should be selecting a high-level option above.

moneromooo-monero:
If we use TLS, can we essentially look like any other encrypted P2P connection (disregarding traffic timing/volums analysis for now) ?

xiphon:
[...]

PR #7088, opened 2020-12-07
Add TLSA support to DNSSEC fetching

vtnerd:
This adds a function to retrieve IPv4 or IPv6 records, along with TLSA records if the domain has DNSSEC enabled. These functions are unused in this patch. They will potentially be used by p2p encryption (#7078 ) to DNS seed nodes (wth TLS or noise encryption), or by `wallet2` SSL autodetect mode. If both features are rejected by the community, this PR will be closed or reverted.

This has been tested by SSL autodetect mode with the wallet. The associated OpenSSL code is not in this patch.

EDIT: From the OpenSSL description on DANE/TLSA, which helps explain:

> It is expected that the majority of clients employing DANE TLS will be doing "opportunistic DANE TLS" in the sense of RFC7672 and RFC7435. That is, they will use DANE authentication when DNSSEC-validated TLSA records are published for a given peer, and otherwise will use unauthenticated TLS or even cleartext.

The p2p mode would be for publishing pubkeys for DNS seeds, but only if they cannot be included directly in the codebase. Given the context of not hard-coding IP addresses for these seeds, it seems possible that this will be needed.

### DIFF
--- a/src/net/resolve.cpp
+++ b/src/net/resolve.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, The Monero Project
+// Copyright (c) 2020, The Monero Project
 //
 // All rights reserved.
 //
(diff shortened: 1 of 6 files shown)


src/net/resolve.cpp
```diff
@@ -0,0 +1,71 @@
...
+        return {net::error::bogus_dnssec};
+      if (ip_records.empty())
+        return {net::error::dns_query_failure};
+    }
+
+    std::vector<std::string> tlsa{};
+    if (dnssec_available && !tlsa_port.empty())
+    {
+      tlsa = resolver.get_tlsa_tcp_record(addr, tlsa_port, dnssec_available, dnssec_valid);
+      if (!dnssec_valid)
```
moneromooo-monero, line 65: It seems possible that dnssec_available might return false on that last call.

vtnerd: This is intended but perhaps incorrect. If the IP records could not be fetched with DNSSEC the assumption is that TLSA records for the domain could not be fetched either. This also guarantees that the service port number is never leaked via DNS when a domain has DNSSEC disabled.

moneromooo-monero: I see. The case I see is if a record lapses between the two calls. But looking at the implementation of get_record, it seems fine anyway, so disregard my comment.

perfect-daemon (approved):
correct

PR #7094, opened 2020-12-08
protocol: batch tx verification on new blocks and tx messages

moneromooo-monero:


src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -455,18 +455,31 @@ namespace cryptonote
       m_core.resume_mine();
       return 1;
     }
-    for(auto tx_blob_it = arg.b.txs.begin(); tx_blob_it!=arg.b.txs.end();tx_blob_it++)
+
+    std::vector<tx_blob_entry> tx_blobs;
+    tx_blobs.reserve(arg.b.txs.size());
```
iamamyth, line 460: In terms of code clarity and performance, it seems helpful, both here and below, to declare a local `const` representing the vector size and use that value everywhere: As the size of the vector does not depend on any kind of conditional, the reader can more easily tell what's happening (these vectors are all the same size, namely, `arg.b.txs.size()`), and the compiler can, too.

PR #7096, opened 2020-12-08
new --min-relay-fee-multiplier monerod option

moneromooo-monero:


moneromooo-monero:
It seemed like a good idea at the time.

PR #7102, opened 2020-12-09
cmake: fix "release" version string on tagged releases

moneromooo-monero:
It would not say "release" if we're on a previous tag

### DIFF
--- a/cmake/GitVersion.cmake
+++ b/cmake/GitVersion.cmake
@@ -48,29 +48,21 @@ function (get_version_tag_from_git GIT)
         message(STATUS "You are currently on commit ${COMMIT}")
 
         # Get all the tags
-        execute_process(COMMAND "${GIT}" rev-list --tags --max-count=1 --abbrev-commit 
+        execute_process(COMMAND "${GIT}" tag -l --points-at HEAD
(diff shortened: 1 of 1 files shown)


cmake/GitVersion.cmake
```diff
@@ -48,29 +48,22 @@ function (get_version_tag_from_git GIT)
...
+        execute_process(COMMAND "${GIT}" tag -l --points-at HEAD
                         WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
                         RESULT_VARIABLE RET
-                        OUTPUT_VARIABLE TAGGEDCOMMIT
+                        OUTPUT_VARIABLE TAG
                         OUTPUT_STRIP_TRAILING_WHITESPACE)
 
-        if(NOT TAGGEDCOMMIT)
-            message(WARNING "Cannot determine most recent tag. Make sure that you are building either from a Git working tree or from a source archive.")
+        message(STATUS "*******************: ${TAGGEDCOMMIT}")
```
hyc, line 57: This message() doesn't make much sense - TAGGEDCOMMIT isn't defined now. ?

moneromooo-monero: Oops, debug stuff I left in ^_^

ndorf (approved):

PR #7106, opened 2020-12-10
protocol: more restrictive checks on chain entry response

moneromooo-monero:


selsta (approved):

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -2532,14 +2532,21 @@ skip:
...
     for (size_t i = 0; i < arg.m_block_ids.size(); ++i)
     {
+      blocks_found.insert(arg.m_block_ids[i]);
       const uint64_t block_weight = arg.m_block_weights.empty() ? 0 : arg.m_block_weights[i];
       context.m_needed_objects.push_back(std::make_pair(arg.m_block_ids[i], block_weight));
       if (++added == n_use_blocks)
         break;
     }
-    context.m_last_response_height -= arg.m_block_ids.size() - n_use_blocks;
+    if (blocks_found.size() != arg.m_block_ids.size())
```
SChernykh, line 2544: This assumes that the previous loop always iterates without early exit. Are you sure `n_use_blocks` is always the same as `arg.m_block_ids.size()`?

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -2486,7 +2486,7 @@ skip:
       drop_connection(context, true, false);
       return 1;
     }
-    if (arg.total_height < arg.m_block_ids.size() || arg.start_height > arg.total_height - arg.m_block_ids.size())
+    if (arg.total_height < arg.m_block_ids.size() || arg.start_height > arg.total_height - arg.m_block_ids.size() || arg.start_height >= m_core.get_current_blockchain_height())
```
SChernykh, line 2489: Just to clarify. Is it a response to our request, so starting height must always be below current blockchain height?

src/cryptonote_protocol/cryptonote_protocol_handler.inl
SChernykh: I think it's safer to check the return value of `blocks_found.insert(arg.m_block_ids[i])` to see if it actually inserted new element or not. From C++ docs:
>  the function returns a pair object whose first element is an iterator pointing either to the newly inserted element in the container or to the element whose key is equivalent, and a bool value indicating whether the element was successfully inserted or not.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
moneromooo-monero: It is.

SChernykh (approved):

Issue #7013, opened 2020-11-11
Wallet started with --offline option showing "Error: wallet failed to connect to daemon"


jonathancross:
When starting the v0.17.1.1 wallet on an offline system, I am seeing this confusing error message despite using the `--offline` option:
```
Error: wallet failed to connect to daemon: http://localhost:18081. Daemon either is not started or wrong port was passed. Please make sure daemon is running or change the daemon address using the 'set_daemon' command.
Background refresh thread started
```

moneromooo-monero:
https://github.com/monero-project/monero/pull/7113

PR #7113, opened 2020-12-10
simplewallet: don't complain about connecting to the daemon when offline

moneromooo-monero:


src/simplewallet/simplewallet.cpp
```diff
@@ -4763,9 +4763,14 @@ bool simple_wallet::try_connect_to_daemon(bool silent, uint32_t* version)
   if (!m_wallet->check_connection(version))
   {
     if (!silent)
-      fail_msg_writer() << tr("wallet failed to connect to daemon: ") << m_wallet->get_daemon_address() << ". " <<
-        tr("Daemon either is not started or wrong port was passed. "
-        "Please make sure daemon is running or change the daemon address using the 'set_daemon' command.");
+    {
+      if (m_wallet->is_offline())
+        fail_msg_writer() << tr("wallet failed to connect to daemon, because it is set to offline mode");
```
jonathancross, line 4768: Wouldn't this still show the error message, even when there is no error because the user has explicitly requested `--offline`?

moneromooo-monero: This is inside "if (!m_wallet->check_connection(version))" (ie, the "falied" branch).

jonathancross: Sorry, I am not following...
Are you saying _"Error: wallet failed to connect to daemon, because it is set to offline mode"_ will not be displayed if user specifies ` --offline`?

moneromooo-monero: No, the opposite :)

jonathancross: Then... it seems it does the _opposite_ of fixing #7013 ;-)

Can we change the message to "Wallet is running in offline mode." and make it purely informational (rather than showing up as an error)?

moneromooo-monero: Well, it doesn't have a daemon so it can't do whatever you asked it to do, right ? Am I just missing your point ?

vtnerd (approved):

src/simplewallet/simplewallet.cpp
jonathancross: I was seeing this "error" immediately upon opening the wallet using the `--offline` option.  In my case at least, it was not an error, but deliberate.  Maybe that situation is already fixed by preventing `m_wallet->auto_refresh()` below?

moneromooo-monero: It doesn't try to auto refresh now when --offline is set. It *should* display that error only when you try to use the daemon with things like.... manual refresh  I guess. Or start_mining, etc.

jonathancross: Okay, let's see how this goes and I can open another issue if the "error" is still showing up in the wrong circumstances.
Thank you.

jonathancross (approved):

PR #7129, opened 2020-12-11
bump the number of out connections in sync mode

moneromooo-monero:
Defaults to 4 times the default number, can be overridden
with --out-peers-sync-boost


moneromooo-monero:
It would be nice to get feedback from various people whether this helps or not. I have a feeling it makes things worse if your connection's already maxed out at 12 peers, in which case it'd have to do some download speed checks to know when to stop adding more peers.

PR #7130, opened 2020-12-11
protocol: include first new block in chain entry response

moneromooo-monero:


src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1813,6 +1813,16 @@ skip:
       LOG_ERROR_CCONTEXT("Failed to handle NOTIFY_REQUEST_CHAIN.");
       return 1;
     }
+    if (r.m_block_ids.size() >= 1)
+    {
+      cryptonote::block b;
+      if (!m_core.get_block_by_hash(r.m_block_ids[1], b))
```
xiphon, line 1819: Should be `r.m_block_ids[0]`

moneromooo-monero: No, [0] is the overlapping block, 1 is the first new block. I want this one.

xiphon: Got it. Should be ` if (r.m_block_ids.size() >= 2)` then

moneromooo-monero: Thanks. For reference, 7135 is what uses it (might not compile, it was full of conflicts and it's not ready so I did not bother building)

xiphon (approved):

PR #7135, opened 2020-12-12
WIP - protocol: check first new block pow before accepting a new set of hashes

moneromooo-monero:
This means an extra PoW call per peer per ~9k blocks in sync mode.
It also means we can relax the timeout for non responsive peers, since
this should make it a lot harder to lie about new blocks.

NOT READY yet, just here to show how #7130 is meant to be used.


selsta:
This idea should be revisited in the future, but this PR was never finished and it's been a while so I'm closing this.

PR #7136, opened 2020-12-12
Remove payload copy in all outgoing p2p messages

vtnerd:
This removes a copy of the payload in all outgoing p2p messages. The biggest impact is on requests for recent blocks - the entire p2p response will typically exceed the cpu cache which has an impact on the copy performance. Combined with other changes, modern GNU/glibc can go from DOM->bytes with zero copies (other than initial write), even if the buffer needs to be resized (kernel performs page table shuffling in worse case).

This also has an additional optimization for block notifications. The entire notification is generated once, and then ref-counted into the outgoing buffer queue for each connection.

If an encryption scheme is (eventually) used for p2p, another copy would be required (encrypted payload). However, the encryption will likely occur after the entire message is generated, with 64KiB chunks being encrypted at a time.

### DIFF
--- a/contrib/epee/include/net/levin_protocol_handler_async.h
+++ b/contrib/epee/include/net/levin_protocol_handler_async.h
@@ -51,6 +51,21 @@
 #define MIN_BYTES_WANTED	512
 #endif
 
+template<typename context_t>
+void on_levin_traffic(const context_t &context, bool initiator, bool sent, bool error, size_t bytes, const char* category)
+{
+  MCINFO("net.p2p.traffic", context << bytes << " bytes " << (sent ? "sent" : "received") << (error ? "/corrupt" : "")
+      << " for category " << category << " initiated by " << (initiator ? "us" : "peer"));
+}
+
+template<typename context_t>
+void on_levin_traffic(const context_t &context, bool initiator, bool sent, bool error, size_t bytes, int command)
+{
+  char buf[32];
+  snprintf(buf, sizeof(buf),  "command-%u", command);
+  on_levin_traffic(context, initiator, sent, error, bytes, buf);
+}
+
 namespace epee
 {
 namespace levin
@@ -88,11 +103,10 @@ public:
   uint64_t m_max_packet_size;
   uint64_t m_invoke_timeout;
 
-  int invoke(int command, const epee::span<const uint8_t> in_buff, std::string& buff_out, boost::uuids::uuid connection_id);
+  int invoke(int command, message_writer in_msg, std::string& buff_out, boost::uuids::uuid connection_id);
   template<class callback_t>
-  int invoke_async(int command, const epee::span<const uint8_t> in_buff, boost::uuids::uuid connection_id, const callback_t &cb, size_t timeout = LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED);
+  int invoke_async(int command, message_writer in_msg, boost::uuids::uuid connection_id, const callback_t &cb, size_t timeout = LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED);
 
-  int notify(int command, const epee::span<const uint8_t> in_buff, boost::uuids::uuid connection_id);
   int send(epee::byte_slice message, const boost::uuids::uuid& connection_id);
   bool close(boost::uuids::uuid connection_id);
   bool update_connection_context(const t_connection_context& contxt);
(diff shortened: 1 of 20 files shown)


contrib/epee/include/byte_stream.h
```diff
@@ -175,7 +175,7 @@ namespace epee
     void put_n(const std::uint8_t ch, const std::size_t count)
     {
       check(count);
-      std::memset(tellp(), count, ch);
+      std::memset(tellp(), ch, count);
```
vtnerd, line 178: There is going to be a separate PR for this change.

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -87,11 +102,10 @@ class async_protocol_handler_config
   uint64_t m_max_packet_size; 
   uint64_t m_invoke_timeout;
 
-  int invoke(int command, const epee::span<const uint8_t> in_buff, std::string& buff_out, boost::uuids::uuid connection_id);
+  int invoke(int command, message_writer in_msg, std::string& buff_out, boost::uuids::uuid connection_id);
   template<class callback_t>
-  int invoke_async(int command, const epee::span<const uint8_t> in_buff, boost::uuids::uuid connection_id, const callback_t &cb, size_t timeout = LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED);
+  int invoke_async(int command, message_writer in_msg, boost::uuids::uuid connection_id, const callback_t &cb, size_t timeout = LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED);
```
moneromooo-monero, line 107: Why are these passed by value ? The object seems not too simple, and part of it is pointers so would point to the same data, while some of the data will not be propagated back to the caller.

vtnerd: The custom move function for `byte_stream` will set the pointers in the _moved from_ object to `NULL`, so two objects are not sharing the same data. Constructing the "new" object by move is just copying the three pointers.


The biggest penalty is a mandatory function call for the move, unless LTO is enabled and catches it. I did some inspection, I will put the move constructor in the header instead. The differences in code generation is (practically) trivial, and by-value seemed more natural here.

Extended for posterity: The only (obvious) difference in code generation are optimizations associated with setting the pointers to `NULL`; there's typically more optimization opportunities by-value but less total `mov` instructions if 2+ moved-from objects cannot be optimized away.

moneromooo-monero: As usual, I find your code hard to understand and bound to bite us in the ass later on because of all the implicit stuff that we'll never remember when modifying (unless it's only you ever modifying it I guess).

vtnerd: You're going to need to be more specific.

The `byte_stream` / `message_writer` behavior is identical to `std::string` except copy operations are disabled. If a programmer can understand `std::string` semantics, they can modify anything in this patch competently. Taking `std::string` by-value is a newer design pattern with C++11, but this is discussed in length elsewhere. I probably went too overboard discussing the intricacies of code generation, and should've stated it like that. 

Some possible specifics I can think of:

There's a "brittle" design with how the p2p responses are generated. Its not obvious that the provided `byte_stream` must have space at the beginning for a levin header, and switching to `message_writer` to convey that concept through code results in a performance decrease for notification message handling. I could probably find a solution to that latter problem if this was a concern.

The `send` message already exists, but could be used incorrectly if the documentation isn't read.

------------------

However, I don't see any of this code being hard to understand ... there's no new templates or any obscure design patterns.

moneromooo-monero: I mean the fact these are passed by value, but the origin gets modified, which is how I understood what you said earlier. And your tendency to use operator() (ie, functions without a name). Subjective I guess.

vtnerd: The source `byte_stream` object is modified during a move operation, which follows STL and Boost behavior, along with [common C++11 advice](https://docs.microsoft.com/en-us/cpp/cpp/move-constructors-and-move-assignment-operators-cpp?view=msvc-160). The move operation is transferring ownership of memory, so the pointers in the source object typically have to be altered so the object remains in a "valid state". Otherwise, two objects could be `free`ing or altering the same block of memory in the destructor. Typically this goes unnoticed because the source object is never used again, and the compiler removes the `src.ptr = NULL` lines during dead-code elimination.

Rust got to try again from scratch with new language rules. Self-assignment and re-use of "moved-from" objects are prevented at compile time. This simplifies typical move operation implementations.

moneromooo-monero: Wait, does this only happen if you pass the parameter with std::move (in which case it'd look sane again) ?

vtnerd: Yes. Although after all this discussion, I might make this a `&&` after further disassembling different scenarios. Fairly small difference though.

moneromooo-monero: Thinking about this again, what would have happened if I were to call this function without std::move ?

vtnerd: Compiler error, unless directly from function call return (implicit move).

vtnerd:
Rebased.

moneromooo-monero (approved):

selsta:
@vtnerd Please reopen this PR with compilation fixed.

PR #7137, opened 2020-12-13
readline_buffer: Avoid consecutive duplicates in the history

codesoap:
I often find myself executing the same command multiple times in a row, when using monero-wallet-cli. For example, I often call `refresh` until I can see that my balance is confirmed. When I then try to get to a previous command I entered, by pressing the "up" key, I found that I had to press it many times, before I surpassed the `refresh`s.

With this change, the history will not contain consecutive, identical entries anymore. This works both when the same command is re-entered manually and when it's entered by pressing "up" and "return". This is the same behaviour, that I know from [zsh](https://www.zsh.org/).

### DIFF
--- a/contrib/epee/src/readline_buffer.cpp
+++ b/contrib/epee/src/readline_buffer.cpp
@@ -6,6 +6,7 @@
 #include <boost/thread/lock_guard.hpp>
 #include <boost/algorithm/string.hpp>
 
+static bool same_as_last_line(const std::string&);
 static void install_line_handler();
 static void remove_line_handler();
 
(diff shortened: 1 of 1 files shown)


contrib/epee/src/readline_buffer.cpp
```diff
@@ -192,6 +196,16 @@ static void handle_line(char* line)
   return;
 }
 
+// same_as_last_line returns true, if the last line in the history is
+// equal to test_line.
+static bool same_as_last_line(std::string test_line)
```
xiphon, line 201:
```suggestion
static bool same_as_last_line(const std::string& test_line)
```

codesoap: Thanks for the feedback! I have also adapted the function declaration and force pushed the change.

moneromooo-monero (approved):

xiphon (approved):

PR #7138, opened 2020-12-13
Optional DNS based blocklist

moneromooo-monero:
If enabled, pulls IPs to block on blocklist.moneropulse.*, and
blocks then for 8 days (so IPs dropping from the list will
eventually get unblocked, and DNS failures don't result in
instant clearing of the blocklist).

Enable with --enable-dns-blocklist

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -37,6 +37,7 @@
 #include <boost/optional/optional.hpp>
 #include <boost/thread/thread.hpp>
 #include <boost/uuid/uuid_io.hpp>
+#include <boost/algorithm/string.hpp>
 #include <atomic>
 #include <functional>
 #include <limits>
@@ -120,6 +121,7 @@ namespace nodetool
     command_line::add_arg(desc, arg_ban_list);
     command_line::add_arg(desc, arg_p2p_hide_my_port);
     command_line::add_arg(desc, arg_no_sync);
+    command_line::add_arg(desc, arg_enable_dns_blocklist);
     command_line::add_arg(desc, arg_no_igd);
     command_line::add_arg(desc, arg_igd);
     command_line::add_arg(desc, arg_out_peers);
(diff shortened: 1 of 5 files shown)


src/p2p/net_node.inl
```diff
@@ -497,6 +502,8 @@ namespace nodetool
     if (command_line::has_arg(vm, arg_no_sync))
       m_payload_handler.set_no_sync(true);
 
+    m_enable_dns_blocklist = command_line::has_arg(vm, arg_enable_dns_blocklist);
```
vtnerd, line 505: The user could explicitly specify `false`, which gets forced to `true` here.

src/p2p/net_node.inl
```diff
@@ -1963,6 +1970,52 @@ namespace nodetool
...
+  //-----------------------------------------------------------------------------------
+  template<class t_payload_net_handler>
+  bool node_server<t_payload_net_handler>::update_dns_blocklist()
+  {
+    if (!m_enable_dns_blocklist)
+      return true;
+    if (m_nettype != cryptonote::MAINNET)
+      return true;
+
+    static const std::vector<std::string> dns_urls = {
```
vtnerd, line 1985: `static constexpr const std::initializer_list<const char*>` or `static constexpr const char*[]`. And below would just be `(records, {dns_urls.begin(), dns_urls.end()})` or `(records, {std::begin(dns_urls), std::end(dns_urls)})`.

vtnerd (commented):
Is this a centralization risk?

There's also a small chance of someone manipulating the records to block hosts, to eclipse an user. Unlikely I think, and only a few entities in the DNS chain could manage it, but worth mentioning.

I guess its fundamentally no different than seeding here, and is disabled by default.

src/p2p/net_node.inl
moneromooo-monero: Why ? AFAICT this would create strings/vector every time it runs rather than just once, which runs counter to your usual suggetions, so I might be missing the point.

selsta (approved):

PR #7142, opened 2020-12-13
daemon: the ban command can now load IPs from a file (ban @filename)

moneromooo-monero:


src/daemon/command_parser_executor.cpp
```diff
@@ -602,7 +602,43 @@ bool t_command_parser_executor::ban(const std::vector<std::string>& args)
       return false;
     }
   }
-  return m_executor.ban(ip, seconds);
+  if (boost::starts_with(args[0], "@"))
+  {
+    const std::string ban_list = args[0].substr(1);
+      const boost::filesystem::path ban_list_path(ban_list);
```
vtnerd, line 608: I hate myself for commenting on this, but the spacing switches from 2 to 4 here.

src/daemon/command_parser_executor.cpp
```diff
@@ -602,7 +602,43 @@ bool t_command_parser_executor::ban(const std::vector<std::string>& args)
...
+      boost::system::error_code ec;
+      if (!boost::filesystem::exists(ban_list_path, ec))
+      {
+        std::cout << "Can't find ban list file " + ban_list + " - " + ec.message();
+        return true;
+      }
+
+      std::string banned_ips;
+      if (!epee::file_io_utils::load_file_to_string(ban_list_path.string(), banned_ips))
```
vtnerd, line 617: I commented on this before with @selsta - I just re-read `load_file_to_string` (again), and I'm not seeing a compelling reason to use it. Last time I commented about win32 specific code, but that too seems unnecessary. This is just a single `std::ifstream` instead of the intermediate step.

selsta (approved):

PR #7144, opened 2020-12-13
p2p: ignore incoming peer list entries when we have them blocked

moneromooo-monero:


xiphon (approved):

vtnerd (commented):
This seems like unnecessary churn - doesn't `is_remote_host_allowed` prevent this before an inbound connection is even allowed? I see in two cases `handle_remote_peerlist` can be called from async-responses, so is there is an issue where a connection is blocked, not completely removed from the read-data processing queue, then calls this, then finally closed (as per blocking policy)?

moneromooo-monero:
This is about adding entries from a peer list sent by Alice. Alice's peer list may include arbitrary IPs (including Bob's), which we currently add to the grey list, even if those hosts are blocked. Now, outgoing connections to Bob will not happen if Bob's blocked, but it'd still pollute the grey list.

PR #7145, opened 2020-12-13
p2p: remove peers from grey and anchors lists when blocked

moneromooo-monero:


moneromooo-monero:
It occurs to me that this will remove the peer, but not any other peer at the same host. However, blocks are per host, so if one daemon on a multi-peer host gets blocked, other peers on this host will get blocked, but not removed from the peer list. However, if we change this to remove all peers on this host, it allows someone on a multiuser system to cause others to be banned. I think it's better to remove all peers on that host, but I'll wait for any comment first.

selsta (approved):

PR #7146, opened 2020-12-13
p2p: remove peers from grey and anchors lists when blocked

moneromooo-monero:


vtnerd (approved):
When reviewing how a peer gets put into the anchorlist, I [found something](https://github.com/monero-project/monero/blob/e5decd0cdebb344fa9b483a5610a072c6e8051f1/src/p2p/net_node.inl#L1388) which doesn't appear correct. Where are the other places that an entry gets put into the anchorlist?

moneromooo-monero:
AFAICT this means "record this address in the set of peers we'll try as anchor". Then it picks by "first seen". I'm a bit doubtful it really works as it's meant to but this particular line doens't strike me as wrong. Can you expand on this ?

vtnerd:
I didn't expect the address to be unconditionally added to the anchor list at the end. There is at least one `return true` above though, so I guess it makes sense.

selsta (approved):

PR #7149, opened 2020-12-14
fuzz_tests: add a test for utf8canonical

moneromooo-monero:


selsta:
CI fails with

```
Unknown CMake command "monero_add_test_executable".
```

selsta (approved):

Issue #7126, opened 2020-12-11
Add wallet name/path to the inactivity lock message.


normoes:
I very much like and appreciate the locking of the wallet.

Since I tend to work with several wallets in all the Monero networks it is quite confusing to see the same lock message everywhere.

It would be nice to have the wallet name (file name) or complete file path in the lock message (maybe even along with the network mode):
It could look like this:
```
 ____________________________________________   
/ I locked the '<my_wallet_file_name_here>' Monero 'mainnet' wallet to protect you \
| while you were away                        |
\ see "help set" to configure/disable        /
 --------------------------------------------
```
instead of just:
```
 ____________________________________________   
/ I locked your Monero wallet to protect you \
| while you were away                        |
\ see "help set" to configure/disable        /
 --------------------------------------------
```

normoes:
Is `m_wallet_file` in https://github.com/monero-project/monero/blob/master/src/simplewallet/simplewallet.cpp the right variable to use?

How can I easily  getthe network mode? I bet there is a variable as well.

moneromooo-monero:
Yes. See wallet_info, do the same. Put the (user selected) name in the text below, not the cow speech.

normoes:
Understood. Thanks.

normoes:
So, the result looks like this.
* Nothing changed in the cow speech.
* See the lnked PR https://github.com/monero-project/monero/pull/7153.
```
 ____________________________________________   
/ I locked your Monero wallet to protect you \
| while you were away                        |
\ see "help set" to configure/disable        /
 --------------------------------------------
        \   (__)
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||


Locked due to inactivity. The wallet password is required to unlock the console.
Filename: <my_wallet_file_name_here>
Network type: Stagenet
Wallet password: 
```

bill-mcgonigle:
Can we make this 'set'-able and a non-default behavior?  I have use cases where I would 'set' it, but if we assume that locking exists for the scenario where somebody walks away from their screen and we're protecting that user from would-be troublemakers, having the wallet name on the lock screen may be too big of an information disclosure in some situations.

moneromooo-monero:
That's a fair point.

normoes:
@moneromooo-monero 
The suggestion above is implemented:
```
set show-wallet-name-when-locked 1|0
```

Could you please check?

selsta:
This seems resolved.

PR #7153, opened 2020-12-15
Add wallet name, network mode to inactivity lock.

normoes:
See issue: https://github.com/monero-project/monero/issues/7126

This PR adds the wallet name and the network mode to the inactivity lock message.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -2988,6 +2988,19 @@ bool simple_wallet::set_track_uses(const std::vector<std::string> &args/* = std:
   return true;
 }
 
+bool simple_wallet::set_show_wallet_name_when_locked(const std::vector<std::string> &args/* = std::vector<std::string>()*/)
+{
+  const auto pwd_container = get_and_verify_password();
+  if (pwd_container)
+  {
+    parse_bool_and_use(args[1], [&](bool r) {
+      m_wallet->show_wallet_name_when_locked(r);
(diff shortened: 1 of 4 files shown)


moneromooo-monero (approved):

selsta:
@normoes do you have any opinions on https://github.com/monero-project/monero/issues/7126#issuecomment-752300609 ?

normoes:
@selsta 
Yes, good point.

I can give it a try.
However, I need to resurrect the PR state and merge from `master`.

Could you give me a hint on where to find an example of setting/getting (checking) such a configuration option?

Thanks a lot.

normoes:
@selsta 

OK, I think I found the right places.
I built it locally and tried it with a mainnet wallet.

Seems to work pretty well so far. 
However I am not sure I know all the ways to interact with this, other than:
* `set show-wallet-name-when-locked 1|0`


The message (in case `set show-wallet-name-when-locked 1`) looks like this:
```
 ____________________________________________   
/ I locked your Monero wallet to protect you \
| while you were away                        |
\ see "help set" to configure/disable        /
 --------------------------------------------
        \   (__)
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||


Locked due to inactivity. The wallet password is required to unlock the console.
Filename: my_wallet_file_name
Network type: Mainnet
Wallet password: 
```

selsta (approved):
Looks good now! Thanks.

PR #7154, opened 2020-12-15
Reject existing claimed blocks in sync mode

moneromooo-monero:


src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -2570,6 +2550,13 @@ skip:
         drop_connection(context, true, false);
         return 1;
       }
+      if (!first && m_core.have_block(h))
```
SChernykh, line 2553: Can you clarify why the check is skipped for the first block?

moneromooo-monero: It's the block we have in common, where the new claimed chain attaches.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -2570,6 +2558,17 @@ skip:
         drop_connection(context, true, false);
         return 1;
       }
+      if (height < blockchain_height)
+      {
+        const crypto::hash block_in_chain = m_core.get_block_id_by_height(height);
+        if ((height < context.m_expect_height - 1 && block_in_chain == h) || (height == context.m_expect_height - 1 && block_in_chain != h))
```
SChernykh, line 2564: It's really hard to understand this condition which is the key part of this PR. Why is it `block_in_chain != h` in the end?

SChernykh (approved):

src/cryptonote_protocol/cryptonote_protocol_handler.inl
moneromooo-monero: We expect a set of new hashes from context.m_expect_height. We also expect an overlapping hash right before it, which is the point where the new hashes attach. Anything we get below that (I'm not sure it's valid to get anything below it right now, but let's assume so for the argument) cannot match the hash we have, or the new chain would have grafted higher. For the hash at the attach point, it must match what we have.
This is technically able to get false positives if there was a reorg between the query and the reply.

PR #7157, opened 2020-12-15
rpc: add a busy_syncing field to get_info

moneromooo-monero:
true if and pretty much only if new blocks are being added


xiphon (approved):

selsta (approved):

selsta (approved):

hyc:
Why do we need both this and #7161?

xiphon:
> Why do we need both this and #7161?

`busy_syncing` doesn't report synchronization status while `synchronized` does. I.e. `busy_syncing == false` doesn't mean a node is synchronized.

PR #7159, opened 2020-12-16
Try -O2 optimisation under Linux

mj-xmr:
TBD


vtnerd:
What would be the point? Directly from gcc manpage:

> Disregard strict standards compliance.  -Ofast enables all -O3 optimizations.  It also enables optimizations that are not valid for all standard-compliant programs.  It turns on -ffast-math and the Fortran-specific -fstack-arrays, unless -fmax-stack-var-size is specified, and -fno-protect-parens.

This implies that `-O2` optimizations are enabled already, and therefore switching to it is not experimental. Is there evidence that the optimization level needs to be reduced (the manpage sounds scary, etc.)?

mj-xmr:
> What would be the point?  (the manpage sounds scary, etc.)?

Exactly this!

PR #7160, opened 2020-12-16
Multihost parallel compilation via USE_ICECC=ON or release-all-icecc …

mj-xmr:
…or debug-all-icecc

Build time without icecream: real	27m10s
Build time with icecream     : real	8m35s
Almost 70% decrease of build time on my network.

YMMV depending on how many hosts you have in your network.
Distributing the compilation across many nodes gives an additional benefit of distributing the heat evenly. Your CPU will live longer.

To use icecream please install on the host machine the icecc and icecc-monitor (for fun), and on all nodes in your LAN, only the icecc package.
One of the nodes has to have a scheduler running. Please put into your `/etc/rc.local` the line : `icecc-scheduler -d` and `sudo chmod +x /etc/rc.local` All the nodes in your LAN should be able to automatically find the scheduler.
The command which starts the icecream monitor (as on the picture below) is:
```
icemon
```

icecc

The official docu is here:
https://github.com/icecc/icecream
This documentation is also a prominent one:
https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Using_Icecream

In short, this is a LAN-distributed (dynamically) compilation scheduler. It plays nicely with ccache.
I will write more documentation of my new stuff this Christmas, taking into account Monero's specifics.

This PR doesn't affect the CI in its current setup, as the only 2 (new) targets, which preset the icecream are: release-all-icecc and debug-all-icecc. They are not called by the CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,10 @@ debug-all:
 	mkdir -p $(builddir)/debug
 	cd $(builddir)/debug && cmake -D BUILD_TESTS=ON -D BUILD_SHARED_LIBS=OFF -D CMAKE_BUILD_TYPE=Debug $(topdir) && $(MAKE)
 
+debug-all-icecc:
+	mkdir -p $(builddir)/debug
+	cd $(builddir)/debug && cmake -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=debug -D USE_ICECC=ON -D BUILD_SHARED_LIBS=ON $(topdir) && $(MAKE) -j25
+
 debug-static-all:
 	mkdir -p $(builddir)/debug
 	cd $(builddir)/debug && cmake -D BUILD_TESTS=ON -D STATIC=ON -D CMAKE_BUILD_TYPE=Debug $(topdir) && $(MAKE)
@@ -101,6 +105,11 @@ release-test:
 release-all:
 	mkdir -p $(builddir)/release
 	cd $(builddir)/release && cmake -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=release $(topdir) && $(MAKE)
+
+release-all-icecc:
+	mkdir -p $(builddir)/release
+	cd $(builddir)/release && cmake -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=release -D USE_ICECC=ON -D BUILD_SHARED_LIBS=ON $(topdir) && $(MAKE) -j25
+
 
 release-static:
 	mkdir -p $(builddir)/release
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,14 @@ endif()
 cmake_minimum_required(VERSION 3.5)
 message(STATUS "CMake version ${CMAKE_VERSION}")
 
+option (USE_ICECC "Use icecc/icecream if a usable instance is found" OFF) # Off by default, since it influences the CPU architecture
+if (USE_ICECC)
+	include(FindIcecream) # Has to be called before project() macro to preset the C and CXX vars.
+	set(ARCH "default") # The top condition for icecream to work is to compile a generic binary.
+else()
+	message(STATUS "icecc deselected")
+endif()
+
 project(monero)
 
 option (USE_CCACHE "Use ccache if a usable instance is found" ON)
@@ -54,6 +62,7 @@ if (USE_CCACHE)
 else()
 	message(STATUS "ccache deselected")
 endif()
+
 option (USE_COMPILATION_TIME_PROFILER "Use compilation time profiler (for CLang >= 9 only)" OFF)
 if (USE_COMPILATION_TIME_PROFILER)
 	if (NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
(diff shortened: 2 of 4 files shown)


rbrunner7:
Can you please give a short explanation for us non-gurus that this "icecream" thingy is, why it is so much faster, what kind of drawbacks the use of it could possibly have? Wonders happen in IT damned rarely, and this looks like one, that's why I am a bit suspicious.

If you explained this already to great lengths in some other place that I just don't know of, apologies, and do you have a link?

mj-xmr:
> Can you please give a short explanation for us non-gurus that this "icecream" thingy is, why it is so much faster, what kind of drawbacks the use of it could possibly have? Wonders happen in IT damned rarely, and this looks like one, that's why I am a bit suspicious.
> 
> If you explained this already to great lengths in some other place that I just don't know of, apologies, and do you have a link?

The official docu is here:
https://github.com/icecc/icecream
This documentation is also a prominent one:
https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Using_Icecream

In short, this is a LAN-distributed (dynamically) compilation scheduler. It plays nicely with ccache.
I will write more documentation of my new stuff this Christmas, taking into account Monero's specifics.

This is not a wonder at all. The "work" of compilation is being done in the same way on physical CPUs, but it's being spread across many machines, which can compile different parts at the same time.

Drawbacks:
1) The compiled binary has to be generic (in our case ARCH="default"), so that the nodes speak the same language. Otherwise, the icecc setting is simply ignored and the given target is compiled locally. After running a couple of tests, I saw that the generic binary is actually 1-10% faster, for bigger and smaller machines accordingly, so no loss here, just a slight change.
2) It takes some good deal of RAM, typically more than a local compilation. This can be controlled by the number of jobs a given host is allowed to take (for example a laptop with 4 GB RAM should take just 1 job). In case a job exceeds the RAM limits, or the network connection is broken, the job returns an exception, and the local icecream daemon will insist on performing the compilation of the failed file locally, costing 2x that compilation time, but that's it.

00-matt:
Does distcc/ccache/icecc/whatever really belong in the CMake files? Anyone that wants to use it can just set it in their environment before they call CMake.

mj-xmr:
> Does distcc/ccache/icecc/whatever really belong in the CMake files? Anyone that wants to use it can just set it in their environment before they call CMake.

It's a similar reasoning as with ccache. Putting this into CMake files adds the following benefits:
1) Doing extra checks, making sure that what you wanted is what you got
2) It provides a self documented demo of what can be achieved by using customized cmake command.
3) Specifically it sets the ARCH to "default", which is not an obvious reason, why it works. I have been presetting icecc for a long time, like you suggest, and was not able to find out why it only worked for some initial C dependencies of Monero. 

It all could be simply documented, but why not have a working preset?

mj-xmr:
> Does distcc/ccache/icecc/whatever really belong in the CMake files? Anyone that wants to use it can just set it in their environment before they call CMake.

And regarding ccache, we have already had a lot of trouble with ccache without the extra checks on exotic setups (FreeBSD). What you see there was able to both speed up compilation for a normal case, and gently continue without ccache, when it was proven not to work, instead of just crashing. The CI was relieved.

Makefile
```diff
@@ -75,6 +75,10 @@ debug-all:
 	mkdir -p $(builddir)/debug
 	cd $(builddir)/debug && cmake -D BUILD_TESTS=ON -D BUILD_SHARED_LIBS=OFF -D CMAKE_BUILD_TYPE=Debug $(topdir) && $(MAKE)
 
+debug-all-icecc:
+	mkdir -p $(builddir)/debug-icecc
```
mj-xmr, line 79: Should it be debug-icecc or just debug as with all other targets? Note - icecc compiles a generic (not native) binary.

iamamyth: I understand your rationale, but I don't think it holds: "native" itself depends on cmake settings (machine architecture), so it's highly variable, just like debug (in fact, moreso, because for icecc debug, you end up building a universal binary). If there were a per-config cache, then what you're saying might make sense, but that doesn't exist, and would be a rather atypical make setup (delete all products and rebuild, or set a custom output dir, tends to be the appropriate strategy with make).

mj-xmr: Ok, thanks.
I'm just being extra careful with changing the default behavior/assumptions.

mj-xmr: Now building in standard locations.

mj-xmr:
> Does distcc/ccache/icecc/whatever really belong in the CMake files? Anyone that wants to use it can just set it in their environment before they call CMake.

Some proof why it's useful to have a customized and centralized ccache management:

[Recent FreeBSD build](
https://github.com/monero-project/monero/pull/7209/checks?check_run_id=1614593328):

-- Check for working CXX compiler: /home/runner/work/monero/monero/contrib/depends/x86_64-unknown-freebsd/native/bin/clang++-8 - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
**ccache: error: execv of /home/runner/work/monero/monero/contrib/depends/x86_64-unknown-freebsd/native/bin/clang++-8 failed: Exec format error
-- Found ccache /usr/bin/ccache, but is UNUSABLE! Return code: 1**
-- The ASM compiler identification is Clang

mj-xmr:
I have added some broad and Monero specific documentation regarding icecc (and ccache along the way).

selsta:
https://github.com/monero-project/monero/pull/10586 also supports multihost compilation, so I'm closing this.

PR #7162, opened 2020-12-17
wallet_api: permit refreshing when daemon is not busy syncing

xiphon:


src/wallet/node_rpc_proxy.cpp
```diff
@@ -85,6 +86,15 @@ void NodeRPCProxy::invalidate()
...
 }
 
+boost::optional<std::string> NodeRPCProxy::get_busy_syncing(bool &busy_syncing)
+{
+  auto res = get_info();
+  if (res)
+    return res;
+  busy_syncing = m_busy_syncing;
+  return boost::optional<std::string>();
+}
```
moneromooo-monero, line 96: This will actually check every 30 seconds minimum, which may not be quite what you want here. Is this busy flag going to be checked very often ?

src/wallet/wallet2.cpp
```diff
@@ -12137,6 +12137,18 @@ uint64_t wallet2::get_daemon_blockchain_target_height(string &err)
   return target_height;
 }
 
+bool wallet2::is_daemon_busy_syncing()
+{
+  if (m_offline || !tools::is_local_address(get_daemon_address()))
```
moneromooo-monero, line 12142: Why the local check ?

xiphon: Moved the check into `doRefresh(..)`. We care about local daemons only.

src/wallet/node_rpc_proxy.cpp
xiphon: > Is this busy flag going to be checked very often ?

On every refresh

xiphon:
OTOH i would just nuke `daemonSynced()` check instead (and the PR) and leave it up to API client code to decide.

https://github.com/monero-project/monero/blob/eee0452aa80db9a6f3b319c0fdaab36c3ff7be8a/src/wallet/api/wallet.cpp#L2171

xiphon:
https://github.com/monero-project/monero/pull/7163

xiphon:
Closed in favor of https://github.com/monero-project/monero/pull/7163

PR #7163, opened 2020-12-17
wallet_api: don't skip refresh if the daemon is syncing

xiphon:
Taking care of possible high resource utilization due to simultaneous local daemon syncing and wallet refresh is out of the Wallet API scope.
Should be handled on the client side instead, e.g. checking `busy_syncing` daemon flag and invoking`pauseRefresh()`/`startRefresh()` Wallet APIs.


moneromooo-monero (approved):

selsta:
Continuing in #8930

PR #7169, opened 2020-12-18
tx_pool: full tx revalidation on fork boundaries

moneromooo-monero:
avoids mining txes after a fork that are invalid by this fork's
rules, but were valid by the previous fork rules at the time
they were verified and added to the txpool.

### DIFF
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -1568,61 +1568,59 @@ namespace cryptonote
   {
     CRITICAL_REGION_LOCAL(m_transactions_lock);
     CRITICAL_REGION_LOCAL1(m_blockchain);
-    size_t tx_weight_limit = get_transaction_weight_limit(version);
-    std::unordered_set<crypto::hash> remove;
 
-    m_txpool_weight = 0;
-    m_blockchain.for_all_txpool_txes([this, &remove, tx_weight_limit](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref*) {
-      m_txpool_weight += meta.weight;
-      if (meta.weight > tx_weight_limit) {
-        LOG_PRINT_L1("Transaction " << txid << " is too big (" << meta.weight << " bytes), removing it from pool");
-        remove.insert(txid);
-      }
-      else if (m_blockchain.have_tx(txid)) {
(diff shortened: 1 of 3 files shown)


j-berman (approved):
LGTM

One question: it looks like all tx's from the pool are loaded into memory -- would this be a concern? Thinking of a scenario where someone malicious tries to cause issues at hard fork time by spamming. The default cap on the pool is [~618 MB](https://github.com/monero-project/monero/blob/b328fbecc24c8897f32284c88d37c5e90b4f5d91/src/cryptonote_config.h#L190) and it can be adjusted lower via a param for lower end hardware, so probably not I would think (which is why I approved)

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1568,61 +1568,59 @@ namespace cryptonote
...
+      crypto::hash txid;
+      cryptonote::transaction tx;
+      cryptonote::blobdata blob;
+      txpool_tx_meta_t meta;
+    };
+
+    // get all txids
+    std::vector<tx_entry_t> txes;
+    m_blockchain.for_all_txpool_txes([this, &txes](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref*) {
+      if (!meta.pruned) // skip pruned txes
```
j-berman, line 1587: Also, my reading of this is that at the fork, pruned nodes won't have any validation logic a number of their tx's meet the fork's new rules (even on tx weight). So at fork time, pruned nodes would be relying on non-pruned nodes to avoid any issues, which seems alright? Am I reading that right?

moneromooo-monero: In normal "cruise" usage, transactions get pruned after 5500 blocks.
The only way a pruned tx can get into the txpool is when it is an old tx that was requested by a syncing node. In that case, this only gets requested if the node has a known block hash to compare it with, and this test will not matter.

moneromooo-monero:
That is a good point, they don't seem to need to be loaded at once. I'll fix that.

moneromooo-monero:
Well, loaded fully parsed in memory at least.

moneromooo-monero:
Done, though tbh I don't remember the original test case. Simple enough though.

j-berman (approved):
Update looks good

I also used the view tag stuff I've been working on to test behavior. I repro'd the scenario without this code, then added this code to make sure invalid tx's that were valid prior to fork height would get ejected from the pool at fork height. It worked as expected

PR #7194, opened 2020-12-26
restrict public node checks a little

moneromooo-monero:
do not include blocked hosts in peer lists or public node lists by default,
warn about no https on clearnet and about untrusted peers likely being spies


moneromooo-monero:
Already PRed...

PR #7206, opened 2020-12-27
protocol: fix false positives dropping peers

moneromooo-monero:
it'd trigger on reorgs


Snipa22:
Approving as 7205 was approved previously.

PR #7209, opened 2020-12-27
make the wallet request fewer blocks at once on refresh failure

moneromooo-monero:
it can hit the new portable storage limits


vtnerd (commented):
Similar thoughts on this as in #7213 . Just drop the global restrictions entirely or loosen them.  Or special behavior for RPC client side. The RPC server still has use resources to generate the original response, and the client has all the resources prepared only to stop suddenly while processing. The scan speed is likely impacted too due to impacted due to the smaller chunks being returned in one request.

moneromooo-monero:
The array ones could be dropped I guess, but the object ones should stay. It's all a subjective tradeoff anyway.

moneromooo-monero:
Couple people don't want this.

PR #7210, opened 2020-12-27
make the wallet request fewer blocks at once on refresh failure

moneromooo-monero:
it can hit the new portable storage limits


moneromooo-monero:
Couple people don't want this.

PR #7212, opened 2020-12-27
rpc: pass output indices as a blob in get_blocks.bin

moneromooo-monero:


selsta:
requires #7209, fixes wallet sync

moneromooo-monero:
Someone doesn't like this

PR #7213, opened 2020-12-27
rpc: pass output indices as a blob in get_blocks.bin

moneromooo-monero:


vtnerd (commented):
I recommend undoing the arbitrary array element limitations instead of adding more code to maintain here.

moneromooo-monero:
Someone doesn't like this

PR #7216, opened 2020-12-28
Use precompiled headers via monero_use_pch_private

mj-xmr:
Example in `unit_tests` & `core_tests`. It reduced the compilation time there by 11%. The PCH solution is no silver bullet, but rather an intermediate step, that helps a little, until the headers are cleaned up properly. Also when handled incorrectly (too broadly), the PCHs can actually increase the compilation time. Imagine having to include a large precompiled header file in a small .cpp file, which has no large dependencies.

Usage: set `-D USE_PCH=ON`

In this PR I provide this as a tool to use. The actual usage will depend on other factors, like dynamic measurements, using the other tools, that I prepared.

Speed improvements:
| Parsing (frontend):         1434.1 s |     Parsing (frontend):         1269.1 s   | -11.51% | 
| Codegen & opts (backend):   1523.3 s | Codegen & opts (backend):   1544.3  | 1.38% |

Total work time: 6.5h

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,28 @@ if (USE_COMPILATION_TIME_PROFILER)
 	add_compile_options("-ftime-trace")
 endif()
 
+option(USE_PCH "Use precompiled headers" OFF)
+if (USE_PCH)
+	set(PCH_MIN_CMAKE_VERSION "3.16.6")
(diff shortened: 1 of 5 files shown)


selsta:
Closing as this appears unlikely to get approved, adds quite a bit of code for just example usage.

PR #7226, opened 2020-12-29
[CI] Fix MacOSX ccache by caching the right directory

mj-xmr:
For MacOSX the right directory of ccache for storing and restoring is:
`/Users/runner/Library/Caches/ccache`

In the current setup, the directory ~/.ccache does Not default to the above directory.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
         submodules: recursive
     - uses: actions/cache@v2
       with:
-        path: ~/.ccache
+        path: /Users/runner/Library/Caches/ccache
         key: ccache-macos-build-${{ github.sha }}
         restore-keys: ccache-macos-build-
     - name: install dependencies


selsta (approved):

mj-xmr:
I will want to push one more (empty) change after the successful build, to see the whole system reacts when rereading the ccache. The change will be squashed of course.

mj-xmr:
@selsta : Mac works by finding a "usable ccache" although I don't see the ccache being used, nor stored at all. So perhaps the test was totally correct and the proper library path (for libc++) has to be added to CMake for MacOs explicitly, and this will actually be the fix. This would mean, that the change, that I made, has to be reverted.

mj-xmr:
I give up on this one for now. The test is able to compile the test program with ccache correctly, but no cache is actually created.

selsta:
> I give up on this one for now.

CI does not seem to have this issue. Might be a local issue on my system, not sure what exactly it is causing it.

mj-xmr:
> > I give up on this one for now.
> 
> CI does not seem to have this issue. Might be a local issue on my system, not sure what exactly it is causing it.

I guess, I was fixing a different issue. Because here the ccache doesn't fail, but it isn't used either. You see in the ccache logs, that at least the restored cache size is ~0 bytes.

selsta (approved):

PR #7228, opened 2020-12-30
Add aggressive restrictions to pre-handshake p2p buffer limit

vtnerd:
I have tested this on my local node, but it could use some more testing before merge (this PR is to v0.17 branch). The buffer limit for p2p payloads is reduced to 256 KiB until a handshake is requested (outgoing) or successfully processed (incoming).

This places a tighter binding between the context and the levin code. Previously, the context was only required to have `boost::uuids::uuid m_connection_id` as a member.


src/p2p/net_node.h
```diff
@@ -118,6 +118,8 @@ namespace nodetool
         m_in_timedsync(false)
     {}
 
+    static constexpr int handshake_command() noexcept { return 1001; }
```
moneromooo-monero, line 121: Can that be COMMAND_HANDSHAKE::ID ?

moneromooo-monero (commented):
Should the reset to full packet size be in the notify code path ? If it is, it might be a way for a peer to make it look like a handshake was done, but not actually call the handshake handler (not sure whether it gets called in that case). Similarly, maybe it should be done only if the handler does not error out.

vtnerd:
> Should the reset to full packet size be in the notify code path ? If it is, it might be a way for a peer to make it look like a handshake was done, but not actually call the handshake handler (not sure whether it gets called in that case). Similarly, maybe it should be done only if the handler does not error out.

The full packet size only gets sets in `async_invoke` and `invoke` when the local side is initiating the handshake, and after the `invoke` handler has processed a handshake and changed from `state_before_handshake`. I didn't use the return code because there are [positive return codes for errors](https://github.com/monero-project/monero/blob/master/src/p2p/net_node.inl#L2488) and for success. I didn't consider changing the return codes.

src/p2p/net_node.h
vtnerd: `COMMAND_HANDSHAKE` is a `typedef` for `COMMAND_HANDSHAKE_T<t_payload_net_handler::payload_type>` inside of `node_server`, and the type `t_payload_net_handler` is a template argument to `node_server`. This is outside of the `node_server` class entirely, so I decided to just hard-code and add a `static_assert`.

Instead, I will do `{ return COMMAND_HANDSHAKE_T<int>::ID; }` with a comment, and the `static_assert` to ensure no specializations.

vtnerd: Or wait - is it just easier to leave as `{ return 1001; }` with the `static_assert`? Its unfortunate, but catches mistakes at build time.

vtnerd:
> The full packet size only gets sets in async_invoke and invoke when the local side is initiating the handshake, and after the invoke handler has processed a handshake and changed from state_before_handshake. I didn't use the return code because there are positive return codes for errors and for success. I didn't consider changing the return codes.

And to clarify further - incoming and outgoing notify messages go to different code paths.

src/p2p/net_node.h
moneromooo-monero: Fine by me to leave as is if's not that simple as just replacing.

moneromooo-monero (approved):

binaryFate:
Reviewed from security perspective.

PR #7229, opened 2020-12-30
Add CLSAG fuzz tests

unseddd:
Add basic fuzz drivers for CLSAG signature algorithm


tests/fuzz/clsag_corpus.cpp
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 2017-2020, The Monero Project
```
unseddd, line 1: This file/binary is used to generate basic corpus files. Can remove if wanted.

moneromooo-monero:
I have doubts as to the usefulness of anything but the parsing. The most extreme example (message) executes exactly the same code every time. Am I missing something ?

unseddd:
> The most extreme example (message) executes exactly the same code every time.

All the tests besides the one in `tests/fuzz/clsag.cpp` explored many (hundreds of) different paths. `tests/fuzz/clsag.cpp` does seem to only explore 3 paths, which is weird. I can definitely remove it, or replace it with the test in `tests/fuzz/clsag_serialize.cpp` (since they essentially do the same thing).

Each of the `clsag_*` tests are meant to fuzz the different inputs to the function `verRctCLSAGSimple`, targeting all attacker controllable inputs.

unseddd:
> The most extreme example (message) executes exactly the same code every time.

All the tests besides the one in `tests/fuzz/clsag.cpp` explored many (hundreds of) different paths. `tests/fuzz/clsag.cpp` does seem to only explore 3 paths, which is weird. I can definitely remove it, change it to only read fuzzer data into public keys, or just replace it with the test in `tests/fuzz/clsag_serialize.cpp` (since they essentially do the same thing).

Each of the `clsag_*` tests are meant to fuzz the different inputs to the function `verRctCLSAGSimple`, targeting all attacker controllable inputs.

unseddd:
Replaced `clsag_fuzz_tests` with what was `clsag_deserialize_fuzz_tests`. Tried a `clsag_pubs` fuzzer, but it bails after two paths. May be able to get deeper coverage with a dictionary for public keys, but didn't try.

Added the change as a fixup for easier review, will squash if you approve of these tests.

Re-ran all the fuzzers, `clsag_fuzz_tests` (deserialize) gets the most paths explored (hundreds), and the most new edges (~30). The other two tests explore similar amounts of paths, but less new edges.

If you still think the `clsag_cout` and `clsag_message` tests are unnecessary, I can remove them. IMHO they explore different paths than just deserializing `clsag` data structure, though.

moneromooo-monero:
If the instrumentation reports many path, let's go with it then. I just don't want to have the fuzzing servers busy doing nothing rather than doing useful fuzzing :)

unseddd:
Squashed the fixup commit, and rebase on latest master.

jonathancross:
Ping @unseddd there are conflicts.

unseddd:
> Ping @unseddd there are conflicts.

Fixed

unseddd:
Tests failed due to changes unrelated to this PR: https://github.com/monero-project/monero/pull/7229/checks?check_run_id=1786660301#step:8:1822

Looks like `functional_tests_rpc` failed for reasons...

selsta:
Yea you can ignore the false positives in functional tests.

selsta:
I have opened an updated version of this PR and added you as co-author. Superseded by #10823.

PR #7233, opened 2020-12-30
Fix conditional __AFL_LOOP function definition

### LINKED ISSUE #7230: __AFL_LOOP macro used as a function name
The `__AFL_LOOP` macro is used as a function name in [tests/fuzz/fuzzer.cpp](https://github.com/monero-project/monero/blob/master/tests/fuzz/fuzzer.cpp#L39)

This causes fuzz tests to not compile when not using `OSS_FUZZ` builds, since __AFL_LOOP is defined in [aflplusplus](https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-cc.c#L834) and [afl](https://github.com/google/AFL/blob/fab1ca5ed7e3552833a18fc2116d33a9241699bc/llvm_mode/afl-clang-fast.c#L251).

I can add another macro in `tests/fuzz/fuzzer.h` if persistent mode would be useful for others. Something like:

```
#define BEGIN_SIMPLE_PERSISTENT_FUZZER(X) \
  BEGIN_SIMPLE_FUZZER() \
  while ( __AFL_LOOP((X)) ) {

#define END_SIMPLE_PERSISTENT_FUZZER \
  } \
  END_SIMPLE_FUZZER()
```

unseddd:
`__AFL_LOOP` function name conflicts with macro definition in `AFL`/`AFLPlusPlus`.

Resolves #7230

### DIFF
--- a/tests/fuzz/fuzzer.cpp
+++ b/tests/fuzz/fuzzer.cpp
@@ -35,7 +35,7 @@
 
 #ifndef OSSFUZZ
 
-#if (!defined(__clang__) || (__clang__ < 5))
+#ifndef __AFL_LOOP
 static int __AFL_LOOP(int)
 {
   static int once = 0;


moneromooo-monero:
That'd break cases where it's not defined. It'd be better to fix the #if to catch the correct cases.

unseddd:
> That'd break cases where it's not defined. It'd be better to fix the #if to catch the correct cases.

Ah, yeah, I understand your point. Fixed to directly test whether `__AFL_LOOP` is defined.

moneromooo-monero:
Assuming it's always a define, looks good.

selsta:
This solution is a bit fragile, I have opened https://github.com/monero-project/monero/pull/10822 instead.

PR #7236, opened 2020-12-30
protocol: fix wrong command in logs

moneromooo-monero:
this is not a levin packet, this is just its payload


vtnerd (approved):
This is also fixed in #7136 but that will take a bit longer to get merged.

PR #7241, opened 2020-12-31
Avoid -march=native when reproducible builds are wanted

bmwiedemann:
Distribution builds that intend to have verifiable binaries, set the `SOURCE_DATE_EPOCH` variable.
Those builds should not default to `-march=native` so that verification builds can produce bit-identical results on other CPUs of the same architecture.

See https://reproducible-builds.org/ for why this is good.

This patch was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

Note: this is tested a bit, but might have room for improvement.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,6 +178,10 @@ string(TOLOWER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_LOWER)
 # when ARCH is not set to an explicit identifier, cmake's builtin is used
 # to identify the target architecture, to direct logic in this cmake script.
 # Since ARCH is a cached variable, it will not be set on first cmake invocation.
+
+if (NOT ARCH AND DEFINED ENV{SOURCE_DATE_EPOCH})
+  set(ARCH "default")
+endif()
 if (NOT ARCH_ID)
 if (NOT ARCH OR ARCH STREQUAL "" OR ARCH STREQUAL "native" OR ARCH STREQUAL "default")
   if(CMAKE_SYSTEM_PROCESSOR STREQUAL "")


hyc:
This patch shouldn't be needed, since we already took care of this in fcfa6665a94f670e1eb73cb89b20022602abcc99

bmwiedemann:
We are at 0.17.1.7 and that patch was in there, but the build still had `-march=native` by default.

The patch commit message you linked also does not tell anything in that direction.

moneromooo-monero (approved):
It'd be nice to find something less ad hoc than SOURCE_DATE_EPOCH if possible, but if not, looks ok

bmwiedemann:
A nice alternative would be to always be reproducible by default and only use ARCH=native when requested by the one building it.

hyc:
Making a reproducible build is about more than just setting a consistent timestamp in object files. That's why our reproducible builds use the depends system, driven by the gitian build scripts. IMO setting this in the master CMake file is deceptive because it implies that the build is "reproducible" when it actually isn't, without all of the other dependencies being addressed.

moneromooo-monero:
Oh, this wasn't actually using gitian ?

hyc:
This situation can only arise if not using gitian/depends, since contrib/depends/toolchain.cmake.in explicitly sets ARCH=default.

bmwiedemann:
This is how we build:
https://github.com/bmwiedemann/openSUSE/blob/master/packages/m/monero/monero.spec#L177
And it builds reproducibly with `-DARCH=default` added. IMHO this should be the default.

iamamyth:
Defaulting to native architecture tends to be the default for most cmake builds I've seen, as it better matches typical usage patterns (building on the hardware which will run the binary and wanting a well-optimized optimized release binary). If you want to use cmake to generate a generic build, why not just explicitly set march=default in your particular run configuration?

selsta:
I agree with the points made above and since there was not much support for this change I'm closing this. Package maintainers can set it manually.

PR #7244, opened 2020-12-31
portable_storage: check object limit where appropriate

moneromooo-monero:
also fix pedantic off by one in check


contrib/epee/include/storages/portable_storage_from_bin.h
```diff
@@ -163,6 +163,10 @@ namespace epee
       array_entry_t<type_name> sa;
       size_t size = read_varint();
       CHECK_AND_ASSERT_THROW_MES(size <= m_count / ps_min_bytes<type_name>::strict, "Size sanity check failed");
+      if (typeid(type_name) == typeid(section))
```
vtnerd, line 166: The compiler can probably optimize this completely at compile-time anyway, but this is also:
```c++
if (std::is_same<type_name, section>())
```
which is definitely computed at compile-time.

vtnerd (approved):

selsta (approved):

selsta (approved):

vtnerd (approved):

PR #7245, opened 2020-12-31
Split fuzz tests during fuzz build

### LINKED ISSUE #7232: Fuzz test memory exhaustion
During fuzz testing, I ran into a problem of all available RAM and swap space being exhausted by the build.

The problem seems to come from the [chaingen_main](https://github.com/monero-project/monero/blob/master/tests/core_tests/chaingen_main.cpp) core tests.

Maybe related to #7186 and #7225

The kind of hacky workaround I used was to disable all builds of non-fuzz tests in the `tests/CMakeLists.txt` build file.

Something like:

```
if (FUZZ || OSS_FUZZ)
  add_subdirectory(fuzz)
else()
  add_subdirectory(core_tests)
  add_subdirectory(crypto)
  add_subdirectory(functional_tests)
  add_subdirectory(performance_tests)
  add_subdirectory(core_proxy)
  add_subdirectory(unit_tests)
  add_subdirectory(difficulty)
  add_subdirectory(block_weight)
  add_subdirectory(hash)
  add_subdirectory(net_load_tests)
endif()
```

Only including fuzz tests in a fuzz build is likely a good idea anyway, since it will decrease overall build time, and none of the other tests need to be instrumented by `AFL`/`libFuzzer` anyways.

unseddd:
Only build fuzz tests in a fuzz build, and don't build other tests.

Keeps fuzz compilers from instrumenting other tests, which are not fuzzed.

Resolves #7232


moneromooo-monero:
I'd rather keep the fuzz tests built when not in a fuzz build, since it avoids breaking them and I might run them just once on the samples as a sanity check in the "normal" test suite at some point.

moneromooo-monero (approved):

PR #7248, opened 2021-01-01
add a max levin packet size by command type

moneromooo-monero:


selsta (approved):

src/cryptonote_basic/connection_context.h
```diff
@@ -73,6 +75,7 @@ namespace cryptonote
     int m_expect_response;
     uint64_t m_expect_height;
     size_t m_num_requested;
+    std::map<int, size_t> m_max_bytes;
```
vtnerd, line 78: A sorted `std::vector` or a `boost::container::flat_map` would get better cache locality and require little changes. The values are currently constant, so they could be in a [fixed `static constexpr` array with a `std::lower_bound` lookup](L80).

If there isn't going to be unique values per-connection (eventually), the map should also be moved to some reusable location (or the constant version used) so the memory isn't duplicated. The config in levin handler actually make some sense - there's already a max packet value there, so the `net_node.inl` code can initialize this when it sets the "handler".

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -585,10 +585,11 @@ class async_protocol_handler
           m_cache_in_buffer.erase(sizeof(bucket_head2));
           m_state = stream_state_body;
           m_oponent_protocol_ver = m_current_head.m_protocol_version;
-          if(m_current_head.m_cb > max_packet_size)
+          const size_t max_bytes = m_connection_context.get_max_bytes(phead.m_command);
+          if(m_current_head.m_cb > std::min<size_t>(max_packet_size, max_bytes))
```
vtnerd, line 589: A check needs to be added around 472 - the fragmented packet case where tor/i2p transactions are sent in fixed sized messages at randomized intervals.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -2833,6 +2833,25 @@ skip:
...
+  {
+    context.set_max_bytes(nodetool::COMMAND_HANDSHAKE_T<cryptonote::CORE_SYNC_DATA>::ID, 65536);
+    context.set_max_bytes(nodetool::COMMAND_TIMED_SYNC_T<cryptonote::CORE_SYNC_DATA>::ID, 65536);
+    context.set_max_bytes(nodetool::COMMAND_PING::ID, 1024);
+    context.set_max_bytes(nodetool::COMMAND_REQUEST_SUPPORT_FLAGS::ID, 256);
+
+    context.set_max_bytes(cryptonote::NOTIFY_NEW_BLOCK::ID, 1024 * 1024 * 128); // 128 MB (max packet is a bit less than 100 MB though)
+    context.set_max_bytes(cryptonote::NOTIFY_NEW_TRANSACTIONS::ID, 1024 * 1024 * 128); // 128 MB (max packet is a bit less than 100 MB though)
+    context.set_max_bytes(cryptonote::NOTIFY_REQUEST_GET_OBJECTS::ID, 1024 * 1024 * 2); // 2 MB
+    context.set_max_bytes(cryptonote::NOTIFY_RESPONSE_GET_OBJECTS::ID, 1024 * 1024 * 128); // 128 MB (max packet is a bit less than 100 MB though)
```
vtnerd, line 2846: I'm not sure how effective this patch will be - an attacker can just choose these message types for an attack. Presumably more work could be done "on top of this patch, but this "hole" is seems difficult to fix

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -2833,6 +2833,25 @@ skip:
   }
   //------------------------------------------------------------------------------------------------------------------------
   template<class t_core>
+  void t_cryptonote_protocol_handler<t_core>::on_connection_new(cryptonote_connection_context &context)
+  {
+    context.set_max_bytes(nodetool::COMMAND_HANDSHAKE_T<cryptonote::CORE_SYNC_DATA>::ID, 65536);
```
vtnerd, line 2838: This group of values - the ping and request flags in particular - seem pretty aggressive and might make expansion complicated. Although it seems unlikely that they get changed at all.

vtnerd (changes requested):

contrib/epee/include/net/levin_protocol_handler_async.h
moneromooo-monero: I'm not sure what check you want, I di not understand that cusrom code.

vtnerd: The `m_current_header` has the levin header just like it does on this line. So the check is identical.

selsta (approved):

src/cryptonote_basic/connection_context.h
moneromooo-monero: It's very small so I kept to a linear search, simpler (for me).

vtnerd: Are these values going to be per-connection modified? This can still be per-zone (at least) or even constant array.

And `std::lower_bound` provides the insert location of a sorted array (possible `end()`). It works with `std::pair` directly in this case, but is comparing the second element unnecessarily. Otherwise it requires a custom lambda which is a pita.

moneromooo-monero: Probably not per connection (though maybe for Tor/I2P since those don't get blocks so those can be set to 0), but possibly over time, to match expected block size. I guess I'll make it a binary search since you insist.

moneromooo-monero:
I'm wondering what's a good limit for unknown commands. Going up to 100 MB seems not a good idea, but having some leeway allows adding more commands when needed without having to sync on a fork, 2 MB ?

moneromooo-monero:
Well, I've done that on top. It means we don't drop connection on unknown command, except if it's > 2 MB. I'll add that to the other branch if people seem ok with it.

vtnerd:
> Well, I've done that on top. It means we don't drop connection on unknown command, except if it's > 2 MB. I'll add that to the other branch if people seem ok with it.

I thought I mentioned this previously, but looks like I haven't. I don't understand the point exactly - it seems to be just limiting ourselves but not an attacker. The attacker can send a `NOTIFY_RESPONSE_GET_OBJECTS` ?

moneromooo-monero:
Yes, but are you using "there is a hole here" as a reason not to plug a second hole ?

selsta (approved):

vtnerd:
Restricting the size of the command has the benefit of keeping the eventual C++ data structures to a minimum. This PR reduces the likelihood of some memory usage blow-up in the DOM->internal_struct phase.

However, restricting unknown commands has little benefit (right now) because unknown commands are never parsed. If the attacker wants to use the full 100 MiB allotment for the `std::vector<uin8_t>` buffer, they can just send a known command with unrecognized fields or the "slack space" trick in levin. So its limiting ourselves more than any attacker.

moneromooo-monero:
I did not realize they were not parsed. Alright then, I'll leave it as maximum.

selsta (approved):

PR #7255, opened 2021-01-01
p2p: don't consider connecting to peers that failed before on first pass

moneromooo-monero:
this helps weed out bad nodes a lot faster


selsta (approved):

selsta:
Seems to cause deadlock during sync.

PR #7258, opened 2021-01-01
p2p: fix cubic selection in filtered peer list

moneromooo-monero:
Integer quantization biased the picks a lot (leading some indices
to never be selected)

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -1243,8 +1243,8 @@ namespace nodetool
     if(!max_index)
       return 0;
 
-    size_t x = crypto::rand<size_t>()%(max_index+1);
-    size_t res = (x*x*x)/(max_index*max_index); //parabola \/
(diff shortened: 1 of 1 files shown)


src/p2p/net_node.inl
```diff
@@ -1243,8 +1243,8 @@ namespace nodetool
     if(!max_index)
       return 0;
 
-    size_t x = crypto::rand<size_t>()%(max_index+1);
-    size_t res = (x*x*x)/(max_index*max_index); //parabola \/
+    size_t x = crypto::rand<size_t>()%(16*max_index+1);
```
mj-xmr, line 1246: I'd:
1) use const for the the 2 variables.
2) use a const or constexpr for the 16 and reuse it.

mj-xmr:
Was the intention to make the parabola growing slower with x?
Does 16 have a special meaning here, that a variable would describe?

moneromooo-monero:
No idea.

selsta (approved):

PR #7261, opened 2021-01-02
portable_storage: better sanity checking

moneromooo-monero:


moneromooo-monero:
Wallet sync shows ~50% more fields than objects currently. About 40k objects max, 60k fields max.

selsta (approved):

PR #7270, opened 2021-01-02
cmake: Use CMAKE_COMMAND.

dewyatt:
The cmake executable name can vary across platforms.


moneromooo-monero (approved):

CMakeLists.txt
```diff
@@ -522,7 +522,7 @@ ExternalProject_Add(generate_translations_header
   BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/translations"
   STAMP_DIR ${LRELEASE_PATH}
   CMAKE_ARGS -DLRELEASE_PATH=${LRELEASE_PATH}
-  INSTALL_COMMAND cmake -E echo "")
```
mj-xmr, line 525: Was it the only instance of cmake across the project and dependencies?

PR #7271, opened 2021-01-03
fix accessing an network address in a deleted context

moneromooo-monero:
Both drop_connection and add_host_fail can drop the connection,
which invalidates the context, and thus the address it contains.

Thanks to wfaressuissia[m] for lots of help and prodding when
debugging this

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -1217,8 +1217,9 @@ namespace nodetool
       if(!handle_remote_peerlist(rsp.local_peerlist_new, context))
       {
         LOG_WARNING_CC(context, "COMMAND_TIMED_SYNC: failed to handle_remote_peerlist(...), closing connection.");
+        const auto remote_address = context.m_remote_address;
         m_network_zones.at(context.m_remote_address.get_zone()).m_net_server.get_config_object().close(context.m_connection_id );
-        add_host_fail(context.m_remote_address);
+        add_host_fail(remote_address);
       }
       if(!context.m_is_income)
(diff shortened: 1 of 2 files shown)


selsta (approved):

src/p2p/net_node.inl
```diff
@@ -1381,8 +1382,8 @@ namespace nodetool
 
     if(just_take_peerlist)
     {
-      zone.m_net_server.get_config_object().close(con->m_connection_id);
       LOG_DEBUG_CC(*con, "CONNECTION HANDSHAKED OK AND CLOSED.");
+      zone.m_net_server.get_config_object().close(con->m_connection_id);
```
hyc, line 1386: this change seems gratuitous, and also makes the log message a lie,
since the message is logged before the actual close(). Likewise for the
change below at line 1448.

yezz123 (approved):

src/p2p/net_node.inl
moneromooo-monero: The log will dereference the context to prints its address, and the context would be technically deleted. But you're right that it's better to copy the address first, keep the order, and print the copied address directoy in the log. I'll do that.

selsta (approved):

selsta (approved):

hyc (approved):

vtnerd (commented):
Can the function calls be reordered? That way the host block is stored before notifying the remote connection of a termination.

moneromooo-monero:
add_host_fail might drop the connection, so you can't use the context after that.

PR #7272, opened 2021-01-03
fix accessing an network address in a deleted context

moneromooo-monero:
Both drop_connection and add_host_fail can drop the connection,
which invalidates the context, and thus the address it contains.

Thanks to wfaressuissia[m] for lots of help and prodding when
debugging this


selsta (approved):

hyc (approved):

moneromooo-monero:
Mooted by 7308

PR #7275, opened 2021-01-03
p2p: make REQUEST_SUPPORT_FLAGS optional, pass flags in node data

moneromooo-monero:
removes a back and forth on connect, and the RPC can be removed
once all peers have updated


src/p2p/net_node.inl
```diff
@@ -1188,6 +1189,7 @@ namespace nodetool
     }
     else if (!just_take_peerlist)
     {
+      if (context_.support_flags == 0)
```
vtnerd, line 1192: The refusal to make whitespace changes seems a bit much here (and below). The behavior of these three lines is changing after all.

vtnerd (approved):

PR #7283, opened 2021-01-05
Run tests in parallel, whenever possible

mj-xmr:
using `tests/run-tests-parallel.sh` script.

With the current simplistic approach I was able to achieve a 20% speedup. This new, small test execution framework will allow for further improvements in the future, by separating conflicting portions of code (or simply by selecting non-conflicting ports of monerod).

The speedup was calculated by summing up the time taken by tests, which were being ran while others, longer tests were running (dt), and was divided by the sum of (total run time + dt).

The test run script has a fuse box. If the tests fail for any reason, they are being restarted using a single core. Note, that it's the tests that fail the most often (rpc tests failing randomly) and quickest (unit tests), which are being ran at the beginning, giving you a hint early enough.

Work time: ~6.7h plus a day of passive tests while doing something else.

### DIFF
--- a/tests/run-tests-parallel.py
+++ b/tests/run-tests-parallel.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+# Copyright (c) 2014-2021, The Monero Project
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are
+# permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of
+#    conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this list
+#    of conditions and the following disclaimer in the documentation and/or other
+#    materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may be
+#    used to endorse or promote products derived from this software without specific
+#    prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# The parallelism celing can be controlled with MONERO_PARALLEL_TEST_JOBS env variable,
+# for example:
+#
+# MONERO_PARALLEL_TEST_JOBS=1     tests/run-tests-parallel.py
+# MONERO_PARALLEL_TEST_JOBS=nproc tests/run-tests-parallel.py
+#
+# The minimum between the ceiling and a currently selected reasonable number of threads is used in the end.
+# The reasonable number is selected as a number, that delivers the solution in the shortest time.
+"""
+
+import os
+import sys
+import subprocess
+import multiprocessing
+from timeit import default_timer as timer
+
+NUM_PROC_MAX_REASONABLE = 2 # This might be increased, once the core_tests are divided into many more independent pieces or simply sped up
+TESTS_EXCLUDED_REGEX = "unit_tests|functional_tests_rpc" # These tests collide with core_tests
+ERR_CODE = 1
+
+def get_forced_job_ceiling():
+    try:
+        ceiling_str = os.environ['MONERO_PARALLEL_TEST_JOBS']
+    except KeyError:
+        ceiling = multiprocessing.cpu_count()
+        print("No parallelism ceiling selected. Defaulting to nproc, so", ceiling)
+    else:
+        if ceiling_str == "nproc":
+            ceiling = multiprocessing.cpu_count()
+        else:
+            ceiling = int(ceiling_str)
+
+        print("Parallelism ceiling selected. Using", ceiling, "jobs.")
+
+    return ceiling
+
+
+def run(num_proc_final, fail_fast=False):
+    """
+    Run the excluded tests first, then everything but excluded.
+    In the current situation, the excluded tests are the most probable and quickest to fail,
+    giving an early feedback when fail_fast is set to true.
+    """
+    cmds = []
+    cmds.append(get_ctest_command(num_proc_final, '-R'))
+    cmds.append(get_ctest_command(num_proc_final, '-E'))
+    
+    error = False
+    for cmd in cmds:
+        status = run_cmd(cmd)
+        if status != 0:
+            error = True
+            if fail_fast:
+                return status
+
+    return ERR_CODE if error else 0
+
+def get_ctest_command(num_proc_final, option):
+    cmd = ['ctest', '-j{}'.format(num_proc_final), option, TESTS_EXCLUDED_REGEX]
+    return cmd
+
+def run_cmd(cmd):
+    result = subprocess.run(cmd, stderr=sys.stderr, stdout=sys.stdout)
+    return result.returncode
+
+
+job_ceiling = get_forced_job_ceiling()
+num_proc_final = min(NUM_PROC_MAX_REASONABLE, job_ceiling)
+print("Job number ceiling is" ,job_ceiling, "and the reasonable maximum is currently", NUM_PROC_MAX_REASONABLE, ".")
+print("Using the minimum of the two, so", num_proc_final, "jobs.")
+
+start = timer()
+ret = run(num_proc_final, fail_fast=False)
+tdiff = timer() - start
+print("Testing took:", round(tdiff / 60), "minutes.")
+print("Status:", ret)
+
+exit(ret)
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -153,7 +153,7 @@ jobs:
         CTEST_OUTPUT_ON_FAILURE: ON
       run: |
         ${{env.BUILD_DEFAULT_LINUX}}
-        cmake --build build --target test
+        cd build && MONERO_PARALLEL_TEST_JOBS=2 ../tests/run-tests-parallel.py
 
 # ARCH="default" (not "native") ensures, that a different execution host can execute binaries compiled elsewhere.
 # BUILD_SHARED_LIBS=ON speeds up the linkage part a bit, reduces size, and is the only place where the dynamic linkage is tested.


tests/run-tests-parallel.sh
```diff
@@ -0,0 +1,41 @@
...
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+TESTS_EXCLUDED_REGEX="unit_tests|functional_tests_rpc" # These tests collide with core_tests
+
+# Run the excluded tests first, then everything but excluded, and if any of these fail, 
+# then fallback to the original single processing version
+time ( \
+ctest -j2 -R "${TESTS_EXCLUDED_REGEX}" && \
```
iamamyth, line 37: A fixed number of jobs here seems somewhat undesirable; it would be nice if the script picked a suitable default (based on `nproc`) and allowed a user to override it.

tests/run-tests-parallel.sh
```diff
@@ -0,0 +1,41 @@
...
+ctest -j2 -E "${TESTS_EXCLUDED_REGEX}" \
+) \
+|| ctest
```
iamamyth, line 40: As best as I can tell, ctest returns a non-zero exit status if a test fails, meaning this `||` ends up double-running the test suite on failure. Even if it worked right, I'm not sure it's a good idea, as it seems prudent to expose test runner issues to users so they can be fixed (e.g. if users can specify a parallelism override, then a failure can be manually overridden as needed and hopefully addressed at some juncture, rather than papered over by this wrapper).

mj-xmr: You understand it correctly.

I'm ok with removing the "|| ctest", but I'd like to hear a 3rd opinion. My rationale is, that it would prevent failures of the parallelism itself. Imagine, that somebody adds a new test, which like the 2 excluded ones is incompatible with the core tests. To fix the CI, you need to call the maintainer. Otherwise you could get on with your high priority task and wait until somebody fixes it, independently of your task.

tests/run-tests-parallel.sh
mj-xmr: I was thinking about it. For the first 2 tests more than 2 cores obviously doesn't make sense.

For the second part adding more cores will actually only prolong the test time. It can't currently get any lower, since core_tests take much more time, than all the other tests combined, so already with 2 cores, you experience thread starvation, because of the too long time of core tests. Adding more cores at this time (say nproc as an extreme), will only make the longest test core_tests run longer, since the CPU cache is used up quicker at the beginning, and then when the other test finish sooner, the cache will not be used better by the core_tests.

So adding more cores will only help, if the core tests are split into more l, independent pieces, which I will be doing later. Only after I do it, it will be reasonable to increase the number of cores.

A general remark to everybody: please don't treat my changes as the final state, but just an iteration, in order to keep the PRs small. 20% gain for such a small change is in my opinion worth it.

tests/run-tests-parallel.sh
mj-xmr: And one more remark.
To specify a parallelism override, it's enough for the user just to call `ctest` in the build directory.

mj-xmr: So maybe it would make sense to do the following two moves at the same time:

1) remove the || ctest from the bash script and make it parallel only, with adjustable number of cores in future, once it makes sense.
2) move the || ctest into the Makefile, so that the CI doesn't suffer sudden breaks of the runner script

iamamyth: > move the || ctest into the Makefile, after the runner script is now being called, so that the CI doesn't suffer from sudden breaks of the script.

I had thought to suggest this in my earlier comment, so yes, it seems like a fine solution. But, I see some other options:

1. Add a Makefile target specifically for CI.
2. Have CI run `|| ctest`, rather than including this in the Makefile, as it's really something an ordinary user of the Makefile doesn't typically want or need.

That said, I don't agree with the premise of CI parallelism blocking progress. If a commit causes CI to start failing, that commit should fix the problem (e.g. by not running the new tests in parallel). Otherwise, you'll end up simply never running anything in parallel, because someone will break build parallelism, hit the `||` and not notice the breakage, and master will continue with serial builds. Either the parallelism should be useful and minimally invasive to contributors, such that it's worth their time to fix parallelism-induced issues, or, it shouldn't exist at all.

tests/run-tests-parallel.sh
iamamyth: If certain tests should never run in parallel, and others have a maximum parallelism factor, then you can (and should) limit the maximum number of runners based on the tests (i.e. `nproc` is an upper bound). The point is that using two runners on a single-core setup actively makes things worse, which means some kind of variability is important. Once you accept that premise, it's little extra work (a single branch) to permit users to decide the upper limit, rather than defaulting to `nproc` (and, this feature seems important, otherwise you're potentially making things worse for people who have multiple cores but don't want to dedicate more than one to the build).

> A general remark to everybody: please don't treat my changes as the final state, but just an iteration, in order to keep the PRs small (but still mergable)

No PR is ever a "final state", the question is whether the change makes sense enough on its own. Here, for example, everyone gets stuck with -j2, even if they only have a single core to run the tests. One could argue that is a real step backwards for many users.

moneromooo-monero:
If would be nice if this were not default. It's really annoying when people screw with parallel settings inside makefiles.

selsta:
Seems like we could enable this for CI but yes, probably best if it isn't ok by default.

tests/run-tests-parallel.sh
mj-xmr: To this I fully agree. I should take the minimum between 2 and nproc, or even of 2 and (nproc-1), since a single core setup with a virtual core won't make it any faster either, except for the situation when core_tests are blocked by IO or network.

So would this logic be OK?

```
num cores = min(2, nproc - 1)
if (num_cores < 1) 
{ // in case we're dealing with a single core CPU without a virtual core
   num_cores = 1
}
```

mj-xmr:
> Seems like we could enable this for CI but yes, probably best if it isn't on by default.

So I'll make a separate target for CI and "make" it from the Workflows.

Should the new target go to a separate Makefile as well? I'd rather keep it in the original Makefile, as this will spare a lot of duplication.

tests/run-tests-parallel.sh
mj-xmr: Alright.

iamamyth:
Per my earlier comments, you could also just use an env var to enable parallel builds, which is how cmake itself does things (default to off, have CI set the var and pick it up in the test build script).

tests/run-tests-parallel.sh
iamamyth: Seems OK.

Makefile
```diff
@@ -98,6 +98,10 @@ release-test:
 	mkdir -p $(builddir)/release
 	cd $(builddir)/release && cmake -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=release $(topdir) && $(MAKE) && $(MAKE) test
 
+release-test-ci:
+	mkdir -p $(builddir)/release
+	cd $(builddir)/release && cmake -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=release -D BUILD_SHARED_LIBS=ON $(topdir) && $(MAKE) && MONERO_PARALLEL_TEST_JOBS=nproc $(topdir)/tests/run-tests-parallel.sh
```
mj-xmr, line 103: Notice, that I smuggled -D BUILD_SHARED_LIBS=ON, in order to leverage faster linkage for the release version.

mj-xmr: The target has a default of nproc, and the runner script also has a default of nproc, if the env var isn't set. Then it tries to find the minimum between nproc (or the forced setting) and the **currently** reasonable value of 2 jobs.

mj-xmr:
The RPC tests are repeatedly failing on the mining part, as usual, but it proves the usefulness of this separation: The CI delivers a failure in the first 40 minutes, as opposed to having to wait 1.5 hour for the negative result.

tests/run-tests-parallel.sh
mj-xmr: In the end, I've decided not to pretend, that a single core machine has 2 logical cores by default (re: nproc -1). If it has, then as said before, the core tests have some IO/Network blocking moments, when other tests may run in the virtual thread. If this is not wanted for a single physical core with 1 virtual core, then I let the user preset this with the environment variable.

mj-xmr:
> Per my earlier comments, you could also just use an env var to enable parallel builds, which is how cmake itself does things (default to off, have CI set the var and pick it up in the test build script).
> 
> Edit: I should add that you don't even need to introduce an extra boolean env var, you could just make use of one which enables controlling the parallelism ceiling (e.g. the CI setup can do `MONERO_TEST_PARALLELISM=nproc run-tests-parallel.sh`).

Thanks. I've made something in-between.

tests/run-tests-parallel.sh
```diff
@@ -0,0 +1,68 @@
...
+# The reasonable number is selected as a number, that delivers the solution in the shortest time.
+
+NUM_PROC_MAX_REASONABLE=2 # This might be increased, once the core_tests are divided into many more independent pieces or simply sped up
+TESTS_EXCLUDED_REGEX="unit_tests|functional_tests_rpc" # These tests collide with core_tests
+
+if [ -z $MONERO_PARALLEL_TEST_JOBS ]; then
+	echo "No parallelism ceiling selected. Using nproc."
+	MONERO_PARALLEL_TEST_JOBS=$(nproc)
+else
+	if [ "$MONERO_PARALLEL_TEST_JOBS" == "nproc" ]; then
```
iamamyth, line 47: Not really a big deal, but do you need this feature? Seems the CI runner can just not set anything and get the default, makes everything a bit simpler.

iamamyth:
Thanks for the changes, overall this looks pretty good to me.

tests/run-tests-parallel.sh
```diff
@@ -0,0 +1,68 @@
...
+# MONERO_PARALLEL_TEST_JOBS=1     tests/run-tests-parallel.sh
+# MONERO_PARALLEL_TEST_JOBS=nproc tests/run-tests-parallel.sh
+#
+# The minimum between the ceiling and a currently selected reasonable number of threads is used in the end.
+# The reasonable number is selected as a number, that delivers the solution in the shortest time.
+
+NUM_PROC_MAX_REASONABLE=2 # This might be increased, once the core_tests are divided into many more independent pieces or simply sped up
+TESTS_EXCLUDED_REGEX="unit_tests|functional_tests_rpc" # These tests collide with core_tests
+
+if [ -z $MONERO_PARALLEL_TEST_JOBS ]; then
```
iamamyth, line 43: If you're using bash here, might as well switch to `[[`, that way you skip an invocation of the `test` program.

tests/run-tests-parallel.sh
mj-xmr: I cover here two distinct use cases at the same time:

1) The new target in the Makefile sets the variable to nproc, so the minimum information is satisfied, but can be also changed in the future.

2) Manually `cd` into the build directory by hand (not by the `make release-test-ci` ) and calling the runner with ../../tests/run-tests-parallel.sh, without having to bother about some env vars. One could then theoretically call `ctest -j$(nproc)`, but we know that it's currently not enough and it wouldn't work.

tests/run-tests-parallel.sh
mj-xmr: I didn't know this. I have a formal training in objective languages, but Bash was always just a hobby.

mj-xmr:
> Thanks for the changes, overall this looks pretty good to me.

Thanks for reasonable comments to all 3 of you.

selsta:
I did not follow the whole discussion but currently it only seems to run

```
    Start 6: functional_tests_rpc
    Start 8: unit_tests
```

Is this intentional? Ideally you would want to run all tests to see which ones fail.

mj-xmr:
> I did not follow the whole discussion but currently it only seems to run
> 
> ```
>     Start 6: functional_tests_rpc
>     Start 8: unit_tests
> ```
> 
> Is this intentional? Ideally you would want to run all tests to see which ones fail.

This is intentional, but can be changed to run all, even though the first typically fail. My philosophy is, that if the tests most probable to fail do fail, you'll have to rerun the whole thing again anyway, just like you'd fail a compilation on the first error, not to have to wait the remaining time to get the feedback. In this case, you'd have to wait additional hour to learn, that your build has failed.
If the other, high level tests do fail afterwards, the proper way to handle this is to write a unit test, that simulates exactly this failure, without having to allocate all the resources. As a unit test it will exposing the weakness in a much shorter time, than the smoke test does.

If this doesn't convince you, I can run them all. Any other opinions on this one? @moneromooo-monero  @iamamyth ?

selsta:
The problem is with these false positives we have (mining, p2p functional tests). As long as we didn't fix these it does not make sense to fail early.

mj-xmr:
> The problem is with these false positives we have (mining, p2p functional tests). As long as we don't fix these it does not make sense to fail early.

If this is how you see it, then I'll run them all, alright!

tests/run-tests-parallel.sh
hyc: I don't think it's a good idea to assume that bash is being used. Should always target vanilla sh.

mj-xmr: The logic got much more complex, than in the initial state, so if I have to switch to sh, I will prefer to do it in Python3 after all. The majority of time was consumed by gathering requirements anyway.

mj-xmr:
Don't be fooled by the green build. The error code propagation from the 1st set of tests didn't work.

mj-xmr:
Reimplemented in Python, without much problems, gaining better readability.

tests/run-tests-parallel.py
```diff
@@ -0,0 +1,120 @@
...
+        return ret # Fail fast here
+    ret2 = run_cmd(ctest2)
+    if ret2 != 0:
+        return ret
+    
+    return 0
+
+def run_all():
+    # Run all of the tests, no matter if the first part fails
+    ctest1 = get_ctest_command(num_proc_final, '-R')
```
iamamyth, line 85: These function looks very much like run_all, they could easily be one function which accepts a `fail_fast` param; e.g.:
```
def run(nproc, fail_fast=False):
    cmds = [get_test_command(...), ...]
    error = False
    for cmd in cmds:
        status = run_cmd(cmd)
        if status != 0:
            if fail_fast:
                return status
            error = True
    return ERR_CODE if error else 0
```

tests/run-tests-parallel.py
```diff
@@ -0,0 +1,120 @@
...
+# The reasonable number is selected as a number, that delivers the solution in the shortest time.
+"""
+
+import os
+import sys
+import subprocess
+import multiprocessing
+from timeit import default_timer as timer
+
+NUM_PROC_MAX_REASONABLE=2 # This might be increased, once the core_tests are divided into many more independent pieces or simply sped up
```
iamamyth, line 48: The spacing style in this file is a bit inconsistent: Sometimes it uses the variant required by shells (`var=val`, no spaces), and sometimes the variant strongly encouraged, though not required, in python (`var = val`, single space on each side of `=`). I would opt for the latter across the board (it's a little jarring as a reader to see different programming styles in a single context).

tests/run-tests-parallel.py
```diff
@@ -0,0 +1,120 @@
...
+TESTS_EXCLUDED_REGEX="unit_tests|functional_tests_rpc" # These tests collide with core_tests
+ERR_CODE=1
+
+def get_forced_ceiling():
+    key = 'MONERO_PARALLEL_TEST_JOBS'
+    if key not in os.environ:
```
iamamyth, line 54: A bit easier on the eyes, and saves a dict lookup, to just fetch the key and catch the exception:
```
try:
    ceiling = os.environ['MONERO_PARALLEL_TEST_JOBS']
except KeyError:
    ceiling = multiprocessing.cpu_count()
else:
    if ceiling == 'nproc':
        ceiling = multiprocessing.cpu_count()
    else:
        ceiling = int(ceiling)
```

tests/run-tests-parallel.py
mj-xmr: Yeah, it was a bit of copy paste here.

tests/run-tests-parallel.py
mj-xmr: Makes sense. I always have (this)[https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why] mind.

tests/run-tests-parallel.py
mj-xmr: Done, but [this principle](https://www.informit.com/articles/article.aspx?p=1392524) kept me from doing this initially. In python at least you can name the parameter.

tests/run-tests-parallel.py
iamamyth: Two things:
1. I find it best not to require the reviewer to read forum links and similar (I'm not much inclined to visit stackexchange), and instead briefly state whatever may be contained therein that you find relevant.
2. As I can glean the contents of the argument from the URL, I will simply state that most arguments of the form "x pattern is bad" tend to be irrelevant on their own: In general, the pattern exists because it can be useful; like all tools, when improperly used, it can be harmful. So, the question is, whether the pattern proves reasonable in a particular context. Here, the exception-based version is quite readable, as it's a standard python idiom, and the overhead isn't a concern, but, if you don't like exceptions, then you could instead use `.get`.

tests/run-tests-parallel.py
iamamyth: Same thoughts as in the below case, but I will merely add, the function here does one thing: Run the test commands and return a status code; it just may not run *all* the commands, depending on the value of `fail_fast`. Anyways, this is a great example of principle-vs-principle being unhelpful: "DRY" vs. "boolean arguments are bad" gets you nowhere. The better questions to ask might be: "does this function make sense as a concept?"; "can I easily describe, e.g. in form of a comment, what this function does?"; "is the internal complexity of the function incredibly high, such that it needs an alternate composition?"; etc.

tests/run-tests-parallel.py
mj-xmr: The discussion under the link suggests, that: 
1) In general it's not a good approach to use exceptions for control the logic
2) In Python it's actually expected to use them this way. Most probably, because the runtime overhead of exceptions in case of Python is much lower than in other languages

So my thinking is, that we shouldn't do this for the core implementation, but I will leave it like it is for Python. Providing a link brings more context and more independent voices. I hope that my 2 above points would describe the content under the link well enough.

Anyway, for me the PR is ready, and if you have no more comments, I'd ask you to approve.
Cheers.

tests/run-tests-parallel.py
mj-xmr: The point of Uncle Bob here (as well in his book), is that whenever there is a boolean parameter, it suggests, that there should be 2 functions. 
But I'm not here to argue about principles, really. If the reviewers prefer it this way, I'm OK with this. Both approaches have pros and cons, but again, in Python at least you can name the boolean parameter when calling the function.

mj-xmr:
Please merge https://github.com/monero-project/monero/pull/7371 first.

The reason is, that since the mining in functional tests requires some more processing power during operation, it could easily overshoot the timeout given for the mining.

iamamyth (approved):

mj-xmr:
**Important!**
Please merge https://github.com/monero-project/monero/pull/7387 first.

The reason is, that since the mining test in functional tests requires some more processing power during operation, it could easily overshoot the timeout given for the mining.

mj-xmr:
This is not yet production ready due to conflicting ports.

mj-xmr:
I have ran the tests several times, [see here from test number 470](https://github.com/mj-xmr/monero/actions/workflows/build.yml?query=event%3Aschedule). There were two failures on `node_server.bind_same_p2p_port`. Initially during a parallel execution of Unit Tests with Functional Tests. Even after executing both in single threaded mode, I got the same error on the same test. The total probability of failure is 10%.

tests/run-tests-parallel.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+job_ceiling = get_forced_job_ceiling()
+num_proc_final = min(NUM_PROC_MAX_REASONABLE, job_ceiling)
+print("Job number ceiling is" ,job_ceiling, "and the reasonable maximum is currently", NUM_PROC_MAX_REASONABLE, ".")
+print("Using the minimum of the two, so", num_proc_final, "jobs.")
+
+start = timer()
+#TODO: set fail_fast=True after mining / RPC random test failures are resolved
+ret = run(num_proc_final, fail_fast=False)
```
mj-xmr, line 105: @selsta : Since the mining test ("Functional") is now stable, do you think I could set the fail fast to True?

mj-xmr: The above test will be fixed with: https://github.com/monero-project/monero/pull/7646

selsta: I don't like the concept of fail fast personally. I want to see everything that fails.

mj-xmr: Then I will remove that TODO from line 104.

mj-xmr:
Rebased.

selsta:
Not necessary after https://github.com/monero-project/monero/pull/10686.

PR #7284, opened 2021-01-06
docs: Remove outdated bits about anonymity network seed nodes

ndorf:
According to IRC, seed nodes now exist for both Tor and I2P, but the doc still says otherwise.

### DIFF
--- a/ANONYMITY_NETWORKS.md
+++ b/ANONYMITY_NETWORKS.md
@@ -36,10 +36,6 @@ with additional exclusive IPv4 address(es).
 
 ## Usage
 
-Anonymity networks have no seed nodes (the feature is still considered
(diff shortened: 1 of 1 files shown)


selsta (approved):

selsta:
@ndorf This has to be PRed against master, else no one will see it.

erciccione (approved):
I'd say it's only necessary to PR to master. PRing a doc change on a branch wouldn't be of much utility.

ndorf:
Sorry for the delay, the PR to master is #7339 

@erciccione not sure I follow the reasoning; people who run from the release branch should have an up to date & correct doc, no?

erciccione:
I was making wrong assumptions. Makes sense to have the double PR.

PR #7290, opened 2021-01-07
blockchain: return empty chain hashes if the start is pruned

moneromooo-monero:
This will still cause the peer to drop us since we're probably
no use to it, but this will happen without the peer having to wait
for a timeout first


selsta:
Closing as `release-v0.17` is not used anymore.

PR #7298, opened 2021-01-09
p2p: only try to connect to IPv6 addresses if using --p2p-use-ipv6

moneromooo-monero:


selsta:
```
In file included from /Users/runner/work/monero/monero/src/rpc/instanciations.cpp:32:
/Users/runner/work/monero/monero/src/p2p/net_node.inl:1602:25: error: 'this' cannot be implicitly captured in this context
          if (!skip && !m_use_ipv6 && pe.adr.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
                        ^
/Users/runner/work/monero/monero/src/p2p/net_node.inl:1602:25: error: 'this' cannot be implicitly captured in this context
/Users/runner/work/monero/monero/src/rpc/instanciations.cpp:36:37: note: in instantiation of member function 'nodetool::node_server<cryptonote::t_cryptonote_protocol_handler<cryptonote::core> >::make_new_connection_from_peerlist' requested here
namespace nodetool { template class node_server<cryptonote::t_cryptonote_protocol_handler<cryptonote::core>>; }
                                    ^
```

moneromooo-monero:
That appears to be in "p2p: don't consider connecting to peers that failed before on first pass", which is then a prerequisite,

selsta:
Superseded by #10799, which includes the same checks and also rejects IPv4-mapped IPv6 peers.

PR #7300, opened 2021-01-09
core_tests: Caching generation, take 1: get_block_longhash()

mj-xmr:
Using a static std::map in order to store the hashes, generated by get_block_longhash()

The core_tests run about 42% faster (from 3619 down to 2100 seconds).

Faster core_tests, means better leveraging of parallel test runs, resulting from [this branch](https://github.com/monero-project/monero/pull/7283), since they're the longest ones.

I'd like to know if you'd like me to continue in this fashion, or should I try to find more sophisticated solutions, which would require a lot of redesign in my view for a similar gain.
If I should continue this way, then please tell me it the further improvements of core_tests can be put into this branch.

Work time: 2h

### DIFF
--- a/src/cryptonote_core/cryptonote_tx_utils.cpp
+++ b/src/cryptonote_core/cryptonote_tx_utils.cpp
@@ -704,7 +704,30 @@ namespace cryptonote
       rx_slow_hash(main_height, seed_height, hash.data, bd.data(), bd.size(), res.data, seed_hash ? 0 : miners, !!seed_hash);
     } else {
       const int pow_variant = b.major_version >= 7 ? b.major_version - 6 : 0;
-      crypto::cn_slow_hash(bd.data(), bd.size(), res, pow_variant, height);
+      
+      // Caching pattern declaration
+      using key_type = std::tuple<blobdata, int, uint64_t>;
+      const key_type key = std::make_tuple(bd, pow_variant, height);
+      static std::map<key_type, crypto::hash> cache_blobdata_2_hash;
+      // Caching pattern declaration ends
+      
+      const auto itcache = cache_blobdata_2_hash.find(key);
+      if (itcache == cache_blobdata_2_hash.end())
+      {
+          // Cache was not hit. Need to generate the hash.
+          crypto::cn_slow_hash(bd.data(), bd.size(), res, pow_variant, height);
+          cache_blobdata_2_hash.emplace(key, res); // Store the resulting hash in the cache for later reuse.
+      }
+      else
+      {
+          // Cache was hit.
+          static uint64_t times_hit = 0;
+          if (++times_hit % 10000000 == 0) // Limit verbosity
+          {
+              MCINFO("cryptonote::get_block_longhash", "Cache hit : " << times_hit <<  " times already" << ENDL);
+          }
+          res = itcache->second; // Assign the found hash to be returned.
+      }
     }
     return true;
   }


src/cryptonote_core/cryptonote_tx_utils.cpp
```diff
@@ -704,7 +704,30 @@ namespace cryptonote
       rx_slow_hash(main_height, seed_height, hash.data, bd.data(), bd.size(), res.data, seed_hash ? 0 : miners, !!seed_hash);
     } else {
       const int pow_variant = b.major_version >= 7 ? b.major_version - 6 : 0;
-      crypto::cn_slow_hash(bd.data(), bd.size(), res, pow_variant, height);
+      
+      // Caching pattern declaration
+      using key_type = std::tuple<blobdata, int, uint64_t>;
+      const key_type key = std::make_tuple(bd, pow_variant, height);
+      static std::map<key_type, crypto::hash> cache_blobdata_2_hash;
+      // Caching pattern declaration ends
```
mj-xmr, line 712: I envision, that the pattern could be used as a template of a map with a key_type and value_type as <K, V>, and accepting a lambda, used to generate the value. The question is, whether it's acceptable to use the static map. I fear of thread safety, when the map is used simultaneously, but against this, the template could arrange a scoped lock.

In case you don't know, if your compiler supports it (it should in 2021), the C++11 standard

mj-xmr: Perhaps you have got other thoughts to share, before I go all-in with this methodology?

selsta: When using monerod normally, are there also cache hits? That is something to consider. If there are always cache misses (outside core tests) then this might not be a good idea.

mj-xmr: I'm currently mostly into reducing the testing time. Helping the normal operation wouldn't hurt, so I setup an experiment for this, but I don't see this code being called at all (whether cache be hit or missed). At least not while synchronizing.

moneromooo-monero:
This is for cn_slow_hash. It'd only be called before the randomx fork. And FWIW when I sped this up vtnerd NAKed it because changes to this function are dangerous (this one in particular is not thread safe as you suspected).

mj-xmr:
> This is for cn_slow_hash. It'd only be called before the randomx fork. And FWIW when I sped this up vtnerd NAKed it because changes to this function are dangerous (this one in particular is not thread safe as you suspected).

Thread safety is something that I can do in 20 minutes for this.

But the first part worries me a bit. If a bottleneck is found in a critical part, why would we avoid improving it? A critical part should mean simply more unit tests around it. Secondly, if this was needed before the randomx, why do we keep calling it? Would it then make more sense to remove the calls to it (here: the tests calling it)?

selsta:
> If a bottleneck is found in a critical part, why would we avoid improving it?

The speedup was only tests related.

mj-xmr:
> The speedup was only tests related.

ph4r05:
I haven't studied the improvements here, just thinking out loud. Won't removing cn_slow_hash and redirecting it to the RX break other blockchain tests that use fork versions before RX was applied? 

When changing this, pls try building also trezor_tests `BUILD_TESTS=ON -D TREZOR_DEBUG=ON`. Or ping me and I will try to run them against the device.

I have no objections to removing old code. I just need to have Trezor-related code tested.

iamamyth:
This change is definitely no good: When syncing from scratch, there are many blocks which would end up being cached, creating a rather large cache. Additionally, while a static cache of "known" blobdata doesn't suffer from thread-safety issues, a shared, static cache, populated dynamically, most certainly does. I don't know which tests are causing a slowdown, but this doesn't seem like a sane solution. If tests take a very long time to run, that's arguably a problem with the design of the tests, but it's most certainly not a problem with how the hash function itself works (if you don't like the slowness of the real hash function, use it sparingly in the tests; don't mess with the implementation, unless you're adding a generic performance optimization which has independent merit).

mj-xmr:
I'm closing this, as per feedback. The finding is important, but the approach has to be different.

PR #7306, opened 2021-01-11
Add test for string emptiness before wiping in http_base

ph4r05:
As discussed in issue https://github.com/monero-project/monero-gui/issues/3296 it is safer to check for string emptiness before wiping it.

### DIFF
--- a/contrib/epee/include/net/http_base.h
+++ b/contrib/epee/include/net/http_base.h
@@ -204,7 +204,8 @@ namespace net_utils
 
 			void wipe()
 			{
-				memwipe(&m_body[0], m_body.size());
+				if (!m_body.empty())
+					memwipe(&m_body[0], m_body.size());
 			}
 		};
 	}


xiphon (approved):

moneromooo-monero:
And I was wrong there, after testing.

ph4r05:
@moneromooo-monero do you propose to close this without merging?

moneromooo-monero:
I don't really mind either code, but it's pointless so...

vtnerd:
The `m_body[0]` expression is only valid if `!m_body.empty()`. It probably works most of the time, but technically isn't allowed. C++17 adds a mutable `char* std::string::data()` function, which is the better (unavailable) alternative.

vtnerd (approved):

moneromooo-monero (approved):

ph4r05:
@vtnerd, @selsta pointed out that emptiness testing is not needed in this case, https://github.com/monero-project/monero/pull/7335#issuecomment-763970830

Do you agree? If yes, we can close this PR.

selsta:
@ph4r05 7335 has been merged, I think this can be closed.

PR #7308, opened 2021-01-12
boosted_tcp_server: fix connection lifetime

selsta:


contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -891,12 +891,22 @@ template<class t_connection_context> template<class callback_t>
...
 {
   CRITICAL_REGION_LOCAL(m_connects_lock);
-  for(auto& c: m_connects)
-  {
-    async_protocol_handler<t_connection_context>* aph = c.second;
-    if(!cb(aph->get_context_ref()))
+  std::vector<typename connections_map::mapped_type> conn;
+  conn.reserve(m_connects.size());
+
+  auto scope_exit_handler = misc_utils::create_scope_leave_handler([&conn]{
```
vtnerd, line 897: Good grief.

I'm being lazy (at the moment) by not running this locally, but `start_outer_call()` and `finish_outer_call()` shouldn't be needed in the change. The issue is that you cannot modify `m_connects` while iterating over it. The lock protects against other threads, but it is a recursive lock so the callback can interfere with the container.

```c++
std::vector<typename connections_map::mapped_type> conn{m_connects};
for (auto &aph : conn)
```
should be sufficient and incur less CPU penalty.

The `start_outer_call()` isn't needed because `m_connects_lock` protects other threads from removing the connection, and if the callback tries to destroy the connection the destructor waits for 60s before blowing up. Which means the callback cannot destroy the connection in this design.

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -907,6 +917,10 @@ bool async_protocol_handler_config<t_connection_context>::for_connection(const b
   async_protocol_handler<t_connection_context>* aph = find_connection(connection_id);
   if (!aph)
     return false;
+  if (!aph->start_outer_call())
```
vtnerd, line 920: This should also be unnecessary. The `m_connects_lock` provides protects other threads from destroying the object, and the callback cannot destroy the connection object safely.

vtnerd (changes requested):

contrib/epee/include/net/levin_protocol_handler_async.h
vtnerd: Or wait, copying the `.second` requires more than a constructor call but I think the idea is clear.

vtnerd (approved):
I ran this locally, and it helped highlight that I forgot about the `add_ref()` call in `start_outer_call()`.

The fundamental problem is this is a `weak_ptr` via some sketchy raw pointer code. I actually looked into making this a `weak_ptr` once and dropped the patch because it impacted these foreach calls. I'll probably look into reviving it, because the external count cleans up a lot of the code. This test case will help verifying some sanity checking too.

contrib/epee/include/net/levin_protocol_handler_async.h
vtnerd: Disregard.

contrib/epee/include/net/levin_protocol_handler_async.h
vtnerd: Disregard.

selsta:
Segfaults in some rare cases, more work is needed.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -269,8 +269,6 @@ PRAGMA_WARNING_DISABLE_VS(4355)
     //_dbg3("[sock " << socket().native_handle() << "] add_ref, m_peer_number=" << mI->m_peer_number);
     CRITICAL_REGION_LOCAL(self->m_self_refs_lock);
     //_dbg3("[sock " << socket().native_handle() << "] add_ref 2, m_peer_number=" << mI->m_peer_number);
-    if(m_was_shutdown)
-      return false;
```
moneromooo-monero, line 273: AFAICT, this would mean operations get started on shutdown connections. Why is this needed, rather then let those operations fail in the first place ?

moneromooo-monero (approved):

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: Operations could be requested after shutdown before or after this change. The lock was immediately released a few lines below, with requests for sending data occurring after that. This might cause some more calls into the send function to fail, but it doesn't appear incorrect.

There's still a "race" to the `add_invoke_response_handler` call but that should only delay cleanup of the object until the async invoke timeout is triggered.

vtnerd (approved):

PR #7311, opened 2021-01-12
miniupnp: set submodule to upstream, disable install

selsta:
No one maintains our fork, switch to upstream.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,8 +4,7 @@
 	branch = monero
 [submodule "external/miniupnp"]
 	path = external/miniupnp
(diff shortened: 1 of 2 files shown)


xiphon (approved):

moneromooo-monero (approved):

luigi1111:
Merging through github.com UI (so unsigned), because https://stackoverflow.com/questions/56813479/cloning-a-repo-with-nested-submodules-does-not-work I think.

Gingeropolous:
note that you have to do this to make it build now if you had an old repo

git submodule sync
git submodule update

Issue #7291, opened 2021-01-07
[PROPOSAL] Add advanced command to monero-wallet-cli: scan_block


hMihaiDavid:
I propose to add a command to monero-wallet-cli that allows an advanced user to specify a list of block heights that the client will scan for outputs.

## Motivation

If an advanced user doesn't want to wait for scanning the whole blockchain for outputs they own (when, for example, restoring the wallet from a mnemonic seed), currently they can specify the block height to start the scanning from (by using the '--restore-height' or the interactive prompt during initialization). However, this can still be a lengthy process since the user would have to specify, ideally, the block height of the first transaction directed to them. This can still take a long time if restoring a paper wallet first used several years ago (all blocks from that height to the latest one will be scanned).

If an advanced user can omit the initial scanning, they can specify the list of blocks with all the known outputs by hand using the proposed command. These could be stored along the seed in a paper wallet. I am not aware for a current mechanism that implements this feature (either for block heights or tx hashes).

The user could omit the initial scanning by just specifying as restore height the current block or current date, which would take very little time to initialize.

## Discussion

I am aware that starting a full node for the first time will still need to linearly traverse the whole blockchain. However, this command would help the restore time for paper wallets using a remote node already synced, either a trusted or untrusted node. There may be privacy concerns by leaking block ids of user's outputs to untrusted nodes which may be interesting to discuss.

Please let me know if this is already implemented or if the devs consider this as a very unnecessary feature that would not be merged.

If it is deemed to be somewhat worthwhile I would like to start developing this feature to alleviate dev time.

Have a nice day.

moneromooo-monero:
I think it might be useful, and should be a fairly simple shell over the mid-level wallet2 functions.

moneromooo-monero:
In fact, an ordered list of txids is probably best. A user will know these, but not necessarily the block heights they got mined, and this'll be even faster.

hMihaiDavid:
Hi @moneromooo-monero ,
Please correct me if I'm wrong, I'm completely new to the monero codebase.

I couldn't find any wallet2.h exported API to use from simplewallet.c that would allow to retrieve any tx in the blockchain by hash, or any block by hash or height. wallet2 exposes the high-level functionality of a wallet, so implementing this would require to get into wallet2.

[...]

PR #7312, opened 2021-01-12
monero-wallet-cli: Added comman scan_tx

hMihaiDavid:
Please see gh-7291

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -282,6 +282,7 @@ namespace
   const char* USAGE_VERSION("version");
   const char* USAGE_HELP("help [<command> | all]");
   const char* USAGE_APROPOS("apropos <keyword> [<keyword> ...]");
+  const char* USAGE_SCAN_TX("scan_tx <txid> [<txid> ...]");
 
   std::string input_line(const std::string& prompt, bool yesno = false)
   {
@@ -3214,6 +3215,45 @@ bool simple_wallet::apropos(const std::vector<std::string> &args)
   return true;
 }
 
+bool simple_wallet::scan_tx(const std::vector<std::string> &args)
+{
+  if (args.empty())
+  {
+    PRINT_USAGE(USAGE_SCAN_TX);
+    return true;
+  }
+
+  // Parse and dedup args
+  std::unordered_set<crypto::hash> txids;
+  for (const auto &s : args) {
+    crypto::hash txid;
+    if (!epee::string_tools::hex_to_pod(s, txid)) {
+      fail_msg_writer() << tr("Invalid txid specified: ") << s;
+      return true;
+    }
+    txids.insert(txid);
+  }
+  std::vector<crypto::hash> txids_v(txids.begin(), txids.end());
+
+  if (!m_wallet->is_trusted_daemon()) {
+    message_writer(console_color_red, true) << tr("WARNING: this operation may reveal the txids to the remote node and affect your privacy");
+    if (!command_line::is_yes(input_line("Do you want to continue?", true))) {
+      message_writer() << tr("You have canceled the operation");
+      return true;
+    }
+  }
+
+  LOCK_IDLE_SCOPE();
+  m_in_manual_refresh.store(true);
+  try {
+    m_wallet->scan_tx(txids_v);
+  } catch (const std::exception &e) {
+    fail_msg_writer() << e.what();
+  }
+  m_in_manual_refresh.store(false);
+  return true;
+}
+
 simple_wallet::simple_wallet()
   : m_allow_mismatched_daemon_version(false)
   , m_refresh_progress_reporter(*this)
@@ -3763,6 +3803,10 @@ simple_wallet::simple_wallet()
                            boost::bind(&simple_wallet::on_command, this, &simple_wallet::apropos, _1),
                            tr(USAGE_APROPOS),
                            tr("Search all command descriptions for keyword(s)"));
+ m_cmd_binder.set_handler("scan_tx",
+                           boost::bind(&simple_wallet::on_command, this, &simple_wallet::scan_tx, _1),
+                           tr(USAGE_SCAN_TX),
+                           tr("Scan the transactions given by <txid>(s), processing them and looking for outputs"));
   m_cmd_binder.set_unknown_command_handler(boost::bind(&simple_wallet::on_command, this, &simple_wallet::on_unknown_command, _1));
   m_cmd_binder.set_empty_command_handler(boost::bind(&simple_wallet::on_empty_command, this));
   m_cmd_binder.set_cancel_handler(boost::bind(&simple_wallet::on_cancelled_command, this));
@@ -11648,4 +11692,3 @@ bool simple_wallet::mms(const std::vector<std::string> &args)
   return true;
 }
 // End MMS ------------------------------------------------------------------------------------------------
-
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -30,6 +30,7 @@
 
 #include <numeric>
 #include <tuple>
+#include <queue>
 #include <boost/format.hpp>
 #include <boost/optional/optional.hpp>
 #include <boost/algorithm/string/classification.hpp>
@@ -1600,6 +1601,47 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
     return "";
   }
   return m_subaddress_labels[index.major][index.minor];
+}
+//----------------------------------------------------------------------------------------------------
+void wallet2::scan_tx(const std::vector<crypto::hash> &txids)
+{
+  // Get the transactions from daemon in batches and add them to a priority queue ordered in chronological order
+  auto cmp_tx_entry = [](const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry& l, const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry& r)
+  { return l.block_height > r.block_height; };
+
+  std::priority_queue<cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry, std::vector<COMMAND_RPC_GET_TRANSACTIONS::entry>, decltype(cmp_tx_entry)> txq(cmp_tx_entry);
+  const size_t SLICE_SIZE =  100; // RESTRICTED_TRANSACTIONS_COUNT as defined in rpc/core_rpc_server.cpp, hardcoded in daemon code
+  for(size_t slice = 0; slice < txids.size(); slice += SLICE_SIZE) {
+    cryptonote::COMMAND_RPC_GET_TRANSACTIONS::request req = AUTO_VAL_INIT(req);
+    cryptonote::COMMAND_RPC_GET_TRANSACTIONS::response res = AUTO_VAL_INIT(res);
+    req.decode_as_json = false;
+    req.prune = true;
+
+    size_t ntxes = slice + SLICE_SIZE > txids.size() ? txids.size() - slice : SLICE_SIZE;
+    for (size_t i = slice; i < slice + ntxes; ++i)
+     req.txs_hashes.push_back(epee::string_tools::pod_to_hex(txids[i]));
+
+    {
+      const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};
+      req.client = get_client_signature();
+      bool r = epee::net_utils::invoke_http_json("/gettransactions", req, res, *m_http_client, rpc_timeout);
+      THROW_WALLET_EXCEPTION_IF(!r, error::wallet_internal_error, "Failed to get transaction from daemon");
+      THROW_WALLET_EXCEPTION_IF(res.txs.size() != req.txs_hashes.size(), error::wallet_internal_error, "Failed to get transaction from daemon");
+    }
+
+    for (auto& tx_info : res.txs)
+      txq.push(tx_info);
+  }
+
+  // Process the transactions in chronologically ascending order
+  while(!txq.empty()) {
+    auto& tx_info = txq.top();
+    cryptonote::transaction tx;
+    crypto::hash tx_hash;
+    THROW_WALLET_EXCEPTION_IF(!get_pruned_tx(tx_info, tx, tx_hash), error::wallet_internal_error, "Failed to get transaction from daemon (2)");
+    process_new_transaction(tx_hash, tx, tx_info.output_indices, tx_info.block_height, 0, tx_info.block_timestamp, false, tx_info.in_pool, tx_info.double_spend_seen, {}, {});
+    txq.pop();
+  }
 }
 //----------------------------------------------------------------------------------------------------
 void wallet2::set_subaddress_label(const cryptonote::subaddress_index& index, const std::string &label)
--- a/src/simplewallet/simplewallet.h
+++ b/src/simplewallet/simplewallet.h
@@ -158,6 +158,7 @@ namespace cryptonote
     bool set_credits_target(const std::vector<std::string> &args = std::vector<std::string>());
     bool help(const std::vector<std::string> &args = std::vector<std::string>());
     bool apropos(const std::vector<std::string> &args);
+    bool scan_tx(const std::vector<std::string> &args);
     bool start_mining(const std::vector<std::string> &args);
     bool stop_mining(const std::vector<std::string> &args);
     bool set_daemon(const std::vector<std::string> &args);
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -1264,6 +1264,8 @@ private:
     std::string get_spend_proof(const crypto::hash &txid, const std::string &message);
     bool check_spend_proof(const crypto::hash &txid, const std::string &message, const std::string &sig_str);
 
+    void scan_tx(const std::vector<crypto::hash> &txids);
+
     /*!
      * \brief  Generates a proof that proves the reserve of unspent funds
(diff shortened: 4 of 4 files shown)


src/simplewallet/simplewallet.cpp
```diff
@@ -3214,6 +3215,41 @@ bool simple_wallet::apropos(const std::vector<std::string> &args)
...
+    txids.insert(s);
+  }
+
+  message_writer() << tr("This operation may reveal the txids to the remote node and affect your privacy in case of using an untrusted node");
+  if (!command_line::is_yes(input_line("Do you want to continue?", true))) {
+    message_writer() << tr("You have canceled the operation");
+    return true;
+  }
+
+  auto saved_m_in_manual_refresh = m_in_manual_refresh.load();
```
xiphon, line 3242: Use `exchange` instead

xiphon: Btw, i don't think you should restore the previous `m_in_manual_refresh` value.

hMihaiDavid: you mean I should do just a `store(true)` ?

xiphon: And set it to `false` once you finished.

xiphon: https://github.com/monero-project/monero/blob/01fdea56adb2ef967b93a3c6b15d291670650f01/src/simplewallet/simplewallet.cpp#L1359-L1360

src/simplewallet/simplewallet.cpp
```diff
@@ -3214,6 +3215,40 @@ bool simple_wallet::apropos(const std::vector<std::string> &args)
...
+{
+  if (args.empty())
+  {
+    PRINT_USAGE(USAGE_SCAN_TX);
+    return true;
+  }
+
+  // Dedup args and make sure they are hashes
+  std::set<string> txids;
+  for (auto s : args) {
```
xiphon, line 3228:
```suggestion
  for (const auto &s : args) {
```

src/simplewallet/simplewallet.cpp
```diff
@@ -3214,6 +3215,40 @@ bool simple_wallet::apropos(const std::vector<std::string> &args)
...
+
+  message_writer() << tr("This operation may reveal the txids to the remote node and affect your privacy in case of using an untrusted node");
+  if (!command_line::is_yes(input_line("Do you want to continue?", true))) {
+    message_writer() << tr("You have canceled the operation");
+    return true;
+  }
+
+  m_in_manual_refresh.store(true);
+  try {
+    m_wallet->scan_tx(std::vector<string>(txids.begin(), txids.end()));
```
xiphon, line 3244: `scan_tx` should accept `std::set` instead.

src/wallet/wallet2.cpp
```diff
@@ -1602,6 +1602,46 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
...
+      THROW_WALLET_EXCEPTION_IF(!r, error::wallet_internal_error, "Failed to get transaction from daemon");
+      THROW_WALLET_EXCEPTION_IF(res.txs.size() != 1, error::wallet_internal_error, "Failed to get transaction from daemon");
+    }
+
+    cryptonote::transaction tx;
+    crypto::hash tx_hash;
+    THROW_WALLET_EXCEPTION_IF(!get_pruned_tx(res.txs.front(), tx, tx_hash), error::wallet_internal_error, "Failed to get transaction from daemon");
+
+    process_new_transaction(tx_hash, tx, res.txs.front().output_indices, res.txs.front().block_height, /*block_version=*/0, res.txs.front().block_timestamp, /*miner_tx=*/false, res.txs.front().in_pool, res.txs.front().double_spend_seen, {}, {});
+  }
```
xiphon, line 1642: Why not request all transactions at once?

hMihaiDavid: I was unsure about the max transactions allowed per request and decided to do this as first approach. Should I change it and request them all at once?

src/wallet/wallet2.cpp
```diff
@@ -1602,6 +1602,46 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
   return m_subaddress_labels[index.major][index.minor];
 }
 //----------------------------------------------------------------------------------------------------
+void wallet2::scan_tx(std::set<std::string> txids)
```
xiphon, line 1605:
```suggestion
void wallet2::scan_tx(const std::set<std::string> &txids)
```

src/wallet/wallet2.cpp
hMihaiDavid: And if it fails... batch them in smaller batches?

xiphon: Request `RESTRICTED_TRANSACTIONS_COUNT` transactions per request. Split into multiple requests if needed.

hMihaiDavid: I'll post a commit with this tomorrow. I have to go now. Thanks for the help.

hMihaiDavid:
Had to convert txids from std::set to std::vector to be able to use indexing. Tell me if there's a more efficient way and I'll refactor it.

hMihaiDavid:
Had to convert txids from std::set to std::vector to be able to use indexing. Tell me if there's a more efficient way and I'll refactor it.

src/simplewallet/simplewallet.cpp
```diff
@@ -3214,6 +3215,40 @@ bool simple_wallet::apropos(const std::vector<std::string> &args)
...
+
+  // Dedup args and make sure they are hashes
+  std::set<string> txids;
+  for (const auto &s : args) {
+    if (!string_tools::validate_hex(64, s)) {
+      fail_msg_writer() << tr("Invalid txid specified");
+      return true;
+    }
+    txids.insert(s);
+  }
```
moneromooo-monero, line 3234: This will still accept duplicate txids.

src/simplewallet/simplewallet.cpp
```diff
@@ -3214,6 +3215,40 @@ bool simple_wallet::apropos(const std::vector<std::string> &args)
...
+  {
+    PRINT_USAGE(USAGE_SCAN_TX);
+    return true;
+  }
+
+  // Dedup args and make sure they are hashes
+  std::set<string> txids;
+  for (const auto &s : args) {
+    if (!string_tools::validate_hex(64, s)) {
+      fail_msg_writer() << tr("Invalid txid specified");
```
moneromooo-monero, line 3230: Could print the bad one, to ease fixing it

src/simplewallet/simplewallet.cpp
```diff
@@ -3214,6 +3215,40 @@ bool simple_wallet::apropos(const std::vector<std::string> &args)
...
+      return true;
+    }
+    txids.insert(s);
+  }
+
+  message_writer() << tr("This operation may reveal the txids to the remote node and affect your privacy in case of using an untrusted node");
```
moneromooo-monero, line 3236: You could use m_wallet->is_trusted_daemon() here.

src/simplewallet/simplewallet.cpp
```diff
@@ -3214,6 +3215,40 @@ bool simple_wallet::apropos(const std::vector<std::string> &args)
...
+  if (!command_line::is_yes(input_line("Do you want to continue?", true))) {
+    message_writer() << tr("You have canceled the operation");
+    return true;
+  }
+
+  m_in_manual_refresh.store(true);
```
moneromooo-monero, line 3242: You want a LOCK_IDLE_SCOPE here

src/simplewallet/simplewallet.cpp
```diff
@@ -3214,6 +3215,40 @@ bool simple_wallet::apropos(const std::vector<std::string> &args)
...
+  try {
+    m_wallet->scan_tx(txids);
```
moneromooo-monero, line 3244: txids is a set, so sorted by txid. This will mess up spend detection. You want to order them chronologically (either by letting hte user give them in order, or sorting them in wallet2, but the latter is not done).

src/wallet/wallet2.cpp
```diff
@@ -1602,6 +1602,53 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
   return m_subaddress_labels[index.major][index.minor];
 }
 //----------------------------------------------------------------------------------------------------
+void wallet2::scan_tx(const std::set<std::string> txids)
+{
+  // If keys are encrypted, ask for password now to avoid timing leak when finding first output.
+  if (m_ask_password == AskPasswordToDecrypt && !m_unattended && !m_watch_only )//&& !m_multisig_rescan_k)
```
moneromooo-monero, line 1608: Leftover debug thing ?

src/wallet/wallet2.cpp
```diff
@@ -1602,6 +1602,53 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
...
+    CRITICAL_REGION_LOCAL(password_lock);
+    if (!m_encrypt_keys_after_refresh)
+    {
+        boost::optional<epee::wipeable_string> pwd = m_callback->on_get_password("scan_tx");
+        THROW_WALLET_EXCEPTION_IF(!pwd, error::password_needed, tr("Password is needed to scan transactions"));
+        THROW_WALLET_EXCEPTION_IF(!verify_password(*pwd), error::password_needed, tr("Invalid password: password is needed to scan transactions"));
+        decrypt_keys(*pwd);
+        m_encrypt_keys_after_refresh = *pwd;
+    }
+  }
```
moneromooo-monero, line 1620: This should not be needed. Was it in your tests ?

src/simplewallet/simplewallet.cpp
xiphon: Oh, i forgot about chronological ordering.

@hMihaiDavid sorry, using the vector was/is okay.

src/wallet/wallet2.cpp
hMihaiDavid: Yes, sry, I should remove it and scan_output would ask for password when a first output was encountered.

src/simplewallet/simplewallet.cpp
hMihaiDavid: My bad, I was not aware of this restriction. I propose to follow the second option by @moneromooo-monero. Ask the daemon for txs with txids in the order returned by the deduped set (ascending lexicographical ordering), order them in chronological order by the timestamp returned by daemon and process them in that order. If you think this is necessary I will let the function receive a vector instead of a set and warn the user to give the txids in chron order.

hMihaiDavid: Correction: unnecessary

Gingeropolous:
this is gonna revolutionize mobile wallet experience. 

so, does this allow a wallet to ask a daemon for a specific set of blocks?

or is it asking for a specific tx_id? 

I was really excited about your previous comment from the proposal thread:
> After resolving this command, I'd like to add a similar command for block heights instead of txids. My initial idea was to write on paper some information for fast recovery of a paper wallet, and tx hashes seem very long.

src/wallet/wallet2.cpp
```diff
@@ -1602,6 +1603,48 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
   return m_subaddress_labels[index.major][index.minor];
 }
 //----------------------------------------------------------------------------------------------------
+void wallet2::scan_tx(const std::set<std::string> txids)
```
moneromooo-monero, line 1606: The copy seems pointless here. Also, wallet2 usually uses crypto::hash in its functions, simplewallet or the RPC server uses strings and converts.

src/wallet/wallet2.cpp
```diff
@@ -1602,6 +1603,48 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
...
+{
+  // Get the transactions from daemon in batches and add them to a priority queue ordered in chronological order
+  auto cmp_tx_entry = [](cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry& l, cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry& r)
```
moneromooo-monero, line 1609: The arguments could be const.

src/wallet/wallet2.cpp
```diff
@@ -1602,6 +1603,48 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
...
+  { return l.block_timestamp > r.block_timestamp; };
```
moneromooo-monero, line 1610: Use block height (timestamps are not monotonic). There does not seem to be a foolproof way to sort within a block as a tie breaker but I don't think this should create issues in practice.

hMihaiDavid:
Replying to @Gingeropolous

> this is gonna revolutionize mobile wallet experience.

Please elaborate more, I'm interested (I thought most mobile wallets were light wallets).

> so, does this allow a wallet to ask a daemon for a specific set of blocks?
> or is it asking for a specific tx_id?
> I was really excited about your previous comment from the proposal thread:
> > After resolving this command, I'd like to add a similar command for block heights instead of txids. My initial idea was to write on paper some information for fast recovery of a paper wallet, and tx hashes seem very long.

The feature so far (scan_tx), allows for asking only for a specific set of txids, not blocks.
When this is done I'd love to continuing adding the same feature but for block heights. It should not be that difficult after doing the scan_tx.

Would you prefer getting blocks by block hashes instead of heights? Is this more useful for the mobile wallets?. I think I could add both apis, for block heights and block hashes, since the rpc allows to get blocks by both identifiers.

Gingeropolous:
> Please elaborate more, I'm interested (I thought most mobile wallets were light wallets).

Mobile wallets are light wallets. The wallet connects to a remote daemon over daemon RPC. So currently, when you create a wallet on your phone using this architecture, the phone has to request all of the blocks from 0 that it wants to scan. Even if you create a new wallet, it retrieves something and still takes a while even though there's no need to scan those blocks. Try it for yourself using cakewallet on iphone or android.

Additionally, after your wallet is synchronized to the daemon, you generally just close the wallet and don't use it for a week or months. Now, when you start up your wallet again, the wallet has to request the intervening 2 weeks of blocks from the daemon in order for the wallet to create transactions again. This is ridiculous, because you, the user, knows that you didn't use your wallet in those 2 weeks, so there's no reason to scan those blocks. 

Thus, when opening a dormant wallet and the user just wants to create a transaction immediately, the user could just hit a "fast forward" button, so that the wallet just requests the last block or the last 5 blocks (or no blocks), and then the user can start crafting transactions immediately instead of having to wait for the wallet to refresh. This provides better UX and also reduces the load on the remote nodes. 

or, if the user did know the received a transaction on january 20th, and they just opened their wallet today, they could tell the wallet to just request blocks from that date. This would cut down significantly on the server load for remote nodes, and provide better user experience. 

> Would you prefer getting blocks by block hashes instead of heights? Is this more useful for the mobile wallets?. I think I could add both apis, for block heights and block hashes, since the rpc allows to get blocks by both identifiers.

I think heights is the easiest bet for the use case i'm imagining, but block hash would probably be useful as well.

hMihaiDavid:
@Gingeropolous

Nice. Thanks for the explanation. I used to think light wallet ran all the blockchain scanning on the serverside with the client just signing txs, I didn't know they connect over rpc to a remote monerod. It's more privacy-preserving this way.

I checked and it seems that cakewallet doesn't use this official wallet-cli code for the wallet, so this feature (scan_tx, scan_block) would have to be implemented there too. Seems like a good idea, it would serve as the basis for the fast-forward mechanism.

hMihaiDavid:
Also btw, I don't do the "check_rpc_cost" stuff after the node rpc call. Should I look into that? Thx.

Gingeropolous:
> Also btw, I don't do the "check_rpc_cost" stuff after the node rpc call. Should I look into that? Thx.

Yes.... maybe? Are you familiar with rpc-pay? I assume check_rpc_cost is associated with the rpc-pay function. Briefly, it allows a remote node user (someone connecting with a wallet) to provide mining hashes towards a block the remote node operator (running the daemon) is mining on. So, its a way for a user to pay for services. No one uses it right now because its not really implemented into any wallets besides the CLI. I think once its in the GUI though it'll be more of a thing.

src/simplewallet/simplewallet.cpp
```diff
@@ -3214,6 +3215,54 @@ bool simple_wallet::apropos(const std::vector<std::string> &args)
...
+  // Convert txids from string to crypto::hash
+  std::vector<crypto::hash> txids;
+  for (const auto &s : txids_s) {
+    cryptonote::blobdata txid_data;
+    if (!epee::string_tools::parse_hexstr_to_binbuff(s, txid_data) || txid_data.size() != sizeof(crypto::hash)) {
+      fail_msg_writer() << tr("Invalid txid specified: ") << s;
+      return true;
+    }
+    txids.push_back(*reinterpret_cast<const crypto::hash*>(txid_data.data()));
+  }
```
moneromooo-monero, line 3245: That's OK, but there's simpler:

   std::unordered_set<crypto::hash> txids;
   for (const auto &s : args)
  {
    crypto::hash txid;
    if (!epee::string_tools::hex_to_pod(s, txid))
    {
        fail_msg_writer() << tr("Invalid txid specified: ") << s;
        return true;
    }
    txids.insert(txid);  // could also check return value second to warn on dedup
  }

moneromooo-monero (approved):

moneromooo-monero:
Please squash.

hMihaiDavid:
should I open a new ph for scan_block or use this one?
also sorry for the delay and messing it up with the commits. I had to force push or the local squash would be undone by a merge. Is this ok?

PR #7313, opened 2021-01-12
build: omit '-pie' linker flag in native non static WIN32 GCC builds 

xiphon:
Even tested GCC 10.2, Windows binaries still die on startup with PIE enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -714,9 +714,10 @@ else()
   endif()
 
   # linker
-  if (NOT SANITIZE AND NOT OSSFUZZ AND NOT (WIN32 AND (CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_LESS 9.1)))
+  if (NOT SANITIZE AND NOT OSSFUZZ AND NOT (WIN32 AND (CMAKE_C_COMPILER_ID STREQUAL "GNU" AND (CMAKE_C_COMPILER_VERSION VERSION_LESS 9.1 OR NOT STATIC))))
     # PIE executables randomly crash at startup with ASAN
     # Windows binaries die on startup with PIE when compiled with GCC <9.x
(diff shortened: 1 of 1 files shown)


iDunk5400:
Native Windows build of `release-static-win64` target works fine here with -pie. Tested with monerod only and a fully up-to-date Windows 10 and MSYS2 (`mingw-w64-x86_64-gcc 10.2.0-6`).

xiphon:
> Native Windows build of `release-static-win64` target works fine here with -pie. Tested with monerod only and a fully up-to-date Windows 10 and MSYS2 (`mingw-w64-x86_64-gcc 10.2.0-6`).

Yeah, seems the issue doesn't affect static builds, breaks non static builds though.
You can test building with `-D STATIC=OFF` to reproduce.  

Updated.

CMakeLists.txt
```diff
@@ -714,7 +714,7 @@ else()
   endif()
 
   # linker
-  if (NOT SANITIZE AND NOT OSSFUZZ AND NOT (WIN32 AND (CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_LESS 9.1)))
+  if (NOT SANITIZE AND NOT OSSFUZZ AND NOT (WIN32 AND (CMAKE_C_COMPILER_ID STREQUAL "GNU" AND (CMAKE_C_COMPILER_VERSION VERSION_LESS 9.1 OR NOT STATIC))))
```
iamamyth, line 717: It would be helpful to update the comments here to match the motivation you've described in this PR (Windows dynamically-linked builds fail with PIE regardless of gcc version).

moneromooo-monero (approved):

CMakeLists.txt
xiphon: Updated. Appended the comment.

selsta (approved):

PR #7315, opened 2021-01-12
Convert expect<T> to boost::system::error_code

vtnerd:
This PR was in response to #7057 . This change is **not** strictly needed, so reject if the "churn" is too high. Its mostly comments and tests. #7057 could remove `expect<void>` usage, but the syntax is more natural.

This switches `expect<T>` to `boost::system::error_code`. This makes it easier to interoperate with `boost::asio` and `boost::filesystem` before Boost 1.65 (which provides conversions to `std::error_code`). Nothing in `std::` uses `std::error_code` until C++17 really, so `boost::system::error_code` makes more sense.

This also aliases things into the `monero::` namespace, so that it could be switched back to `std::error_code` if necessary. Definitely need some feedback/criticism on this approach.

I've taken a look at the disassembly, and with all of these changes (see some comments) the differences in runtime should be negligible.

### DIFF
--- a/src/common/error.cpp
+++ b/src/common/error.cpp
@@ -30,8 +30,12 @@
 
 namespace
 {
-    struct category final : std::error_category
+    struct category final : monero::error_category
     {
+        category() noexcept
+          : monero::error_category()
+        {}
+
         virtual const char* name() const noexcept override final
         {
             return "common_category()";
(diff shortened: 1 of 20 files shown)


vtnerd:
Rebased

selsta:
Can this be closed? We use C++17 and boost 1.69 now.

vtnerd:
Yes, I think this can be closed.

We are better off using the boost -> std conversion when needed, and otherwise use `std::error_code` as much as possible.

PR #7324, opened 2021-01-18
do not display non tor/i2p public nodes by default

moneromooo-monero:
most of them will likely be spies


moneromooo-monero:
Annoyingly, if a node isn't itself set up for tor, it will not receive peer lists for tor, so the wallet won't see any of those with public RPC even if the wallet could use it (ie, using torsocks).

moneromooo-monero:
After thinking about it, not a good idea yet.

PR #7326, opened 2021-01-19
bootstrap_daemon: proxy configuration support

xiphon:
Daemon command
`set_bootstrap_daemon (auto | none | host[:port] [username] [password]) [proxy_ip:proxy_port]"`

CLI switch
`--bootstrap-daemon-proxy` - `<ip>:<port> socks proxy to use for bootstrap daemon connections`

RPC
`/set_bootstrap_daemon` - added `proxy` field

### DIFF
--- a/src/daemon/command_parser_executor.cpp
+++ b/src/daemon/command_parser_executor.cpp
@@ -986,17 +986,67 @@ bool t_command_parser_executor::check_blockchain_pruning(const std::vector<std::
 
 bool t_command_parser_executor::set_bootstrap_daemon(const std::vector<std::string>& args)
 {
-  const size_t args_count = args.size();
-  if (args_count < 1 || args_count > 3)
+  struct parsed_t
+  {
+    std::string address;
+    std::string user;
+    std::string password;
+    std::string proxy;
+  };
+
+  boost::optional<parsed_t> parsed = [&args]() -> boost::optional<parsed_t> {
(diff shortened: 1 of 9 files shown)


src/daemon/command_parser_executor.cpp
```diff
@@ -986,17 +986,30 @@ bool t_command_parser_executor::check_blockchain_pruning(const std::vector<std::
...
+  {
+    const auto last_arg = args.rbegin();
+    constexpr const char proxy_prefix[] = "proxy=";
+    if (last_arg->find(proxy_prefix) == 0)
+    {
+      constexpr size_t proxy_prefix_length = sizeof(proxy_prefix) - sizeof(proxy_prefix[0]);
+      proxy = last_arg->substr(proxy_prefix_length);
+      --args_count;
+    }
+  }
```
moneromooo-monero, line 1006: I think this will allow something like:
set_bootstrap_daemon a.b.c.d:2000   user pass foo

Also, I did not realize password was here, it should probably prompt, since you cna't otherwise have spaces etc (as a separate patch though, and not necessarily by you).

moneromooo-monero (approved):

src/daemon/command_parser_executor.cpp
xiphon: Updated, reworked arguments parsing.

selsta (approved):

PR #7329, opened 2021-01-20
systemd: Revamp and harden systemd configuration to make full use of systemd sandboxing capabilities

aquilacooper:


moneromooo-monero:
Why should we merge this ? You did not say what it's supposed to change.

iamamyth:
A couple of thoughts:

1. There's no sense in renaming monerod.service to monero.service, especially as the unit file in fact controls the daemon service, so the prior name is more accurate.

2. Commit comments should specify something not discoverable from the git object graph; `update <filename> ` fails this test. Everyone can see the commit updates the service file, the question, as noted by moneromoo, is, "why the update?" What is the goal of the changes?

aquilacooper:
> Why should we merge this ? You did not say what it's supposed to change.

The old unit file was inadequate for running `monerod` as a hardened, long-running systemd service. It included no security, sandboxing, and system call filtering options that systemd offers except for `StateDirectory` and `LogsDirectory`. It also relied on `torsocks` for enabling Tor functionality, which has been obsoleted by `--tx-proxy` and `--anonymous-inbound`.

The updates that I'm proposing aim to take as much advantage of systemd's security, sandboxing, and system call filtering options as possible without affecting `monerod`'s functionality and stability. They also take into account `monerod`'s use of `--tx-proxy` and `--anonymous-inbound` for Tor functionality if the user opts for it. I've also included comments in the unit file that I hope would help explain some of the choices made and elaborate on what the specified options do.

My apologies for not expounding on the changes that I made in the initial pull request comment. I presumed that both the changes and the necessity of those changes would have been apparent from the comparison between the two unit files and from the comments written in the updated unit file. The changes are many, and I was therefore loathe to write a detailed essay expounding on all of the changes when the comments in the updated unit file and the systemd man pages would have done a far better job at making clear what those changes did and why they were made.

>  1. There's no sense in renaming monerod.service to monero.service, especially as the unit file in fact controls the daemon service, so the prior name is more accurate.

I took inspiration from how various systemd services like the OpenSSH service are named `ssh.service` and not `sshd.service`. Nevertheless, I leave it up to the maintainers to decide on the best naming scheme for the updated unit file, if they decide to merge it.

>  2. Commit comments should specify something not discoverable from the git object graph; `update <filename> ` fails this test. Everyone can see the commit updates the service file, the question, as noted by moneromoo, is, "why the update?" What is the goal of the changes?

The proposed changes are many, and so I did not know how else to condense all of that into a single line. I had hoped that the changes themselves and the comments written in the updated unit file would have made themselves and their goals clear, but apparently I'm mistaken. My apologies.

iamamyth:
> The proposed changes are many, and so I did not know how else to condense all of that into a single line.

There's no need for commit comments to be a single line. Ideally you have a single-sentence summary and then a more detailed description (if needed), which can be as long as needed. The summary seems apparent to me, something to the effect of, "Harden systemd configuration security".

> I had hoped that the changes themselves and the comments written in the updated unit file would have made themselves and their goals clear, but apparently I'm mistaken

Yes, that would be a mistaken understanding of how almost every project, public and private, operates. Think of the burden implied by such a strategy: Anyone trying to interrogate the history via git will have to read every line of every single commit to understand what it does; that's simply impractical for a large software project.

ndorf:
Why the reversion back to `Type=forking`? The systemd.service(5) manual page specifically recommends that `Type=simple` be used "whenever possible."

aquilacooper:
> Why the reversion back to `Type=forking`? The systemd.service(5) manual page specifically recommends that `Type=simple` be used "whenever possible."

These two snippets from the `Type=` entry in the systemd.service(5) man page largely explain why I opted for `forking` instead of `simple`:

>  Note that this means `systemctl start` command lines for simple services will report success even if the service's binary cannot be invoked successfully (for example because the selected `User=` doesn't exist, or the service binary is missing).

> However, as this service type [i.e., `simple`] won't propagate service start-up failures and doesn't allow ordering of other units against completion of initialization of the service (which for example is useful if clients need to connect to the service through some form of IPC, and the IPC channel is only established by the service itself — in contrast to doing this ahead of time through socket or bus activation or similar), it might not be sufficient for many cases.

To be sure, I do not know whether monerod behaves in the ways that the two snippets warn about. Therefore, for the sake of reliability, I thought it best to err on the safer side by opting for `forking`, which is what the man page recommends if `dbus` or `notify` are not supported by the service.

As far as I'm aware, monerod does not provide a D-Bus interface, nor does it appear to invoke `sd_notify()` or an equivalent API after starting up. This means that monerod cannot be set up as a `dbus` or a `notify` type service.

Besides, monerod itself provides provisions for running as a `forking` type service with the `--detach` and `--pidfile` options. I also wasn't sure if the user would be able to run commands against the `monerod` binary itself when it's launched as a `simple` service with `--non-interactive` enabled. If the user isn't able to, then that would mean that I would not be able to specify `ExecStop=/usr/local/bin/monerod exit` in the unit file, which is my preferred way of asking monerod to stop instead of relying on systemd's default process-killing behavior. The `TimeoutStopSec=300` option then gives monerod up to five minutes to gracefully stop before systemd sends it a `SIGKILL`.

Other than a possible short delay in the boot process, there does not appear to be any other significant drawback to setting up monerod as a `forking` type service. The rationale for setting it up as a `simple` type service in the old unit file—i.e., that `forking` messes with torsocks—is no longer relevant, as the use of torsocks for monerod's Tor functionality has been made obsolete by the `--tx-proxy` and `--anonymous-inbound` options.

iamamyth:
> To be sure, I do not know whether monerod behaves in the ways that the two snippets warn about. Therefore, for the sake of reliability, I thought it best to err on the safer side by opting for forking, which is what the man page recommends if dbus or notify are not supported by the service.

When changing a configuration which itself has recently been changed, it's a good idea to understand the motivation for the change. More broadly, effecting a change when you "don't know" whether it's needed should be avoided. 

I believe you've misinterpreted the man page as well as the operational context for this service:

1. Per the man page, "It is generally recommended to use Type=simple for long-running services whenever possible, as it is the simplest and fastest option. However, as this service type won't propagate service start-up failures and doesn't allow ordering of other units against completion of initialization of the service... it might not be sufficient for many cases." Nothing depends on monerod, so simple works just fine. Therefore, the man page advises its use. If a dependency were to exist, it would most likely be a wallet or some other service depending on the RPC, and forking doesn't really help there: Even if forking were to only report the service as started after the RPC server initializes, that wouldn't actually be useful to consumers, because the daemon could still be busy initializing and return useless responses. Any consumer needs to check the RPC response, at which point, it can also easily check "unable to connect to RPC".

2. As for the behavior of systemctl start, the issue described, that the exit code may be zero even if the service didn't start, applies to literally every simple job. In other words, it's very rarely, if ever, a reason not to use type simple, else that type wouldn't exist; it's merely a usage note.

ndorf:
> > Why the reversion back to `Type=forking`? The systemd.service(5) manual page specifically recommends that `Type=simple` be used "whenever possible."
> 
> These two snippets from the `Type=` entry in the systemd.service(5) man page largely explain why I opted for `forking` instead of `simple`:
> 
> > Note that this means `systemctl start` command lines for simple services will report success even if the service's binary cannot be invoked successfully (for example because the selected `User=` doesn't exist, or the service binary is missing).
> 
> > However, as this service type [i.e., `simple`] won't propagate service start-up failures and doesn't allow ordering of other units against completion of initialization of the service (which for example is useful if clients need to connect to the service through some form of IPC, and the IPC channel is only established by the service itself — in contrast to doing this ahead of time through socket or bus activation or similar), it might not be sufficient for many cases.
> 
> To be sure, I do not know whether monerod behaves in the ways that the two snippets warn about. Therefore, for the sake of reliability, I thought it best to err on the safer side by opting for `forking`, which is what the man page recommends if `dbus` or `notify` are not supported by the service.

This could be useful, but only if both a) other systemd units exist which depend on `monerod`, which is not a standard configuration _and_ b) `monerod` only forked after successful initialization, which is not the case. 

> Besides, monerod itself provides provisions for running as a `forking` type service with the `--detach` and `--pidfile` options. I also wasn't sure if the user would be able to run commands against the `monerod` binary itself when it's launched as a `simple` service with `--non-interactive` enabled.

I'm not sure why you'd just assume that the default configuration would be so broken as to disallow RPC commands to the daemon. These use a TCP connection to localhost:18081 and do not depend in any way on whether the daemon forks or not.

> If the user isn't able to, then that would mean that I would not be able to specify `ExecStop=/usr/local/bin/monerod exit` in the unit file, which is my preferred way of asking monerod to stop instead of relying on systemd's default process-killing behavior. The `TimeoutStopSec=300` option then gives monerod up to five minutes to gracefully stop before systemd sends it a `SIGKILL`.

No, it gives `monerod` up to five minutes before systemd sends it a `SIGTERM` and waits _another_ five minutes before sending it a `SIGKILL`, for a total maximum wait of ten minutes. Given that the daemon handles `SIGTERM` and the `exit` command identically, this would seem to be redundant and pointless. Nonetheless, if you prefer this configuration for whatever reason, nothing about `Type=simple` prevents you from doing it.

jtgrassie:
Renaming monerod.service to monero.service makes tracking the changes here difficult and only seems to serve personal preference in file naming convention.

utils/systemd/monero.service
```diff
@@ -0,0 +1,88 @@
...
+
+Type=simple
+
+### The following "ExecStart=" line assumes that:
+### (a) the monerod binary has been copied to the /usr/local/bin directory and is owned by the "monero" system user and group,
+### (b) the /etc/monero directory exists and is owned by the "monero" system user and group, and
+### (c) the monerod.conf configuration file exists, is located in /etc/monero, and is owned by the "monero" system user and group.
+
+ExecStart=/usr/local/bin/monerod --config-file /etc/monero/monerod.conf --non-interactive
+ExecStop=/usr/local/bin/monerod exit
```
jtgrassie, line 41: This shouldn't be needed.

aquilacooper: If you mean that the `ExecStop=/usr/local/bin/monerod exit` directive isn't needed, I've removed that directive in commit `cb5b8f5`.

aquilacooper:
> Renaming monerod.service to monero.service makes tracking the changes here difficult and only seems to serve personal preference in file naming convention.

Understood. I've reverted the unit file name to `monerod.service` in commit [`cb5b8f5`](https://github.com/monero-project/monero/pull/7329/commits/cb5b8f570776d57df6190548f43cc30df9029343).

utils/systemd/monerod.service
```diff
@@ -1,37 +1,87 @@
...
+# Wants=tor.service
+
+### The following [Install] section allows monero.service to be enabled via "systemctl enable",
+### which in turn causes monero.service to be automatically started at system startup.
+### The "Alias=monerod.service" directive allows monero.service to be invoked as monerod.service in various systemctl commands
+### if monero.service is enabled via "systemctl enable".
+
+[Install]
+
+Alias=monerod.service
```
jtgrassie, line 25: This is not needed as already the service name.

aquilacooper: Oh, right. I forgot about that.

utils/systemd/monerod.service
```diff
@@ -1,37 +1,87 @@
...
+Type=simple
+
+### The following "ExecStart=" line assumes that:
+### (a) the monerod binary has been copied to the /usr/local/bin directory and is owned by the "monero" system user and group,
+### (b) the /etc/monero directory exists and is owned by the "monero" system user and group, and
+### (c) the monerod.conf configuration file exists, is located in /etc/monero, and is owned by the "monero" system user and group.
+
+ExecStart=/usr/local/bin/monerod --config-file /etc/monero/monerod.conf --non-interactive
+RestartSec=60
+TimeoutStopSec=300
```
jtgrassie, line 42: This seems wrong/unduly long. The default (which is 90 sec) should be more than adequate.

aquilacooper: The decision to set `TimeoutStopSec` to five minutes was inspired by Bitcoin's decision to set theirs to 10 minutes. See their rationale in commit [`99cebc9`](https://github.com/bitcoin/bitcoin/commit/99cebc922cba1590ea663d6ef3109c0cf2960b9d#diff-6ca46037cb8f71de1d76344d1892f28d00ccd5b9e09725842809d288293abbbf).

I found 10 minutes to be overkill, so I halved it to five minutes.

The `systemd.service(5)` man page does not specify a recommended value for the `TimeoutStopSec=` directive. Similarly, the `systemd-system.conf(5)` man page does not specify a recommended value for the `DefaultTimeoutStopSec=` directive.

jtgrassie: Simply copying Bitcoin's choice is not a good enough rational. These are different daemons with different shutdown actions. Normally the monerod daemon shutdown is is quick, no more than a few seconds. I'd expect anyone editing this file actually tests what an appropriate timeout should be rather than finger in the air guesswork.

utils/systemd/monerod.service
```diff
@@ -1,37 +1,84 @@
...
+### log-file=/var/log/monero/monerod.log
+
+RuntimeDirectory=monero
 StateDirectory=monero
 LogsDirectory=monero
+ConfigurationDirectory=monero
+RuntimeDirectoryMode=0700
+StateDirectoryMode=0700
+LogsDirectoryMode=0700
+ConfigurationDirectoryMode=0700
```
jtgrassie, line 71: I'm not clear why you're adding these more restrictive permissions . Bitcoin's example service adds things like this presumably because the Bitcoin daemon is also a _wallet_. Monero's daemon is _not_ a wallet and thus I'm not clear why there's such a need to restrict permissions beyond the default 755?

utils/systemd/monerod.service
aquilacooper: Bitcoind's ordinary shutdown duration is also quick, and yet they saw fit to set `TimeoutStopSec=` to 10 minutes. But I digress.

The timeout will not be reached under ordinary circumstances. But in the event of an extraordinary situation where monerod takes much longer to gracefully stop, it is better to have a longer timeout than a shorter one. If the user does not want to wait the whole five minutes, then they can either manually send a `SIGKILL` or edit the `monerod.service` file to shorten the timeout duration for all future runs. After all, this file is a sample file. The user is free to adapt the file to their own needs.

utils/systemd/monerod.service
aquilacooper: Why not? It is good security practice to restrict as much as possible without affecting functionality and stability, regardless of whether monerod is a wallet or not.

This proposed monerod.service file configures monerod to be run by the `monero` system user. There is therefore no need for other users both within and without the `monero` system group to have access to those directories. In fact, for even more robust sandboxing, `ConfigurationDirectoryMode=` can (and perhaps should) be set to `0400`, because monerod does not need to write to `/etc/monero`.

jtgrassie: Fair enough.

utils/systemd/monerod.service
jtgrassie: Fair enough. How about `RestartSec`?

utils/systemd/monerod.service
iamamyth: The config directory mode should most definitely be 0400 (letting the process alter its own service config can simplify certain attacks), but logs should probably be 0755 (letting non-privileged users view logs with mundane information can be helpful and shouldn't pose a risk).

ndorf: Setting the mode to 0400 will not prevent the process from altering its own config. Since it's running as the user that owns the file, it can simply call chmod() and then modify the file. The correct way to do it would be to make it owned by user root, group monero, mode 0640 (or 0440).

ndorf: Also, I would agree that logs should be either world- or at least group-readable.

jtgrassie: Indeed, and why I raised the question. 0700 prevents a normal user viewing daemon logs.

jtgrassie: Thus I agree that "logs should be either world- or at least group-readable".

utils/systemd/monerod.service
aquilacooper: It gives systemd time to restore the system to a pre-`monerod.service` state before the next run, after it receives a non-zero exit signal from `monerod` (because `Restart=` is set to `on-failure`).

The default `RestartSec=` value of 100 milliseconds is too short, especially on slower systems. A repeat of the memory exhaustion attack would also mean that the system may not have enough time to flush its swap memory before systemd relaunches monerod. One minute is long enough for systemd to do what it needs to do on slower systems, but not too long that it would be a significant annoyance.

Again, this is merely a sample unit file. The user is free to set a shorter `RestartSec=` value if they see fit. One minute between runs in `monerod.service`'s case is a good default for most systems.

utils/systemd/monerod.service
aquilacooper: > Also, I would agree that logs should be either world- or at least group-readable.

There's `sudo [command]`, or `sudo --user=monero [command]` if the user does not want to view the logs as root, for that.

> Indeed, and why I raised the question. 0700 prevents a normal user viewing daemon logs.

Tor uses this approach (of setting `0700` for all of its directories), and it's emphatically not a wallet.

ndorf: Why should one have to sudo just to view a logfile, which emphatically doesn't contain any sensitive information?

ndorf: Further, many systems don't use sudo. That leaves you having to `su` to view the logfile. For what benefit, exactly?

aquilacooper:
It appears that certain contributors love nothing more than to strain out a gnat and swallow a camel.

The whole point of this pull request is to revamp the old `monerod.service` unit file into one that makes the fullest use of systemd's security, sandboxing, and system call filtering capabilities without negatively affecting monerod's ability to operate. In other words, the goal of the proposed changes is to restrict as much as possible while giving enough leeway for monerod to function properly.

It therefore baffles me that certain contributors to a project that depends on robust security would complain that the changes are _too_ restrictive, while being up in arms over certain functionally inconsequential proposals like running monerod as a `Type=forking` service and renaming the sample unit file to reflect more canonical naming conventions. It seems to me that either they have a dangerously cavalier mindset concerning what constitutes good security principles, or they find great fun out of picking the most miniscule of nits out of others' work, particularly first-timers'. If this is indicative of the culture and mindset within the wider Monero developer community—I certainly hope not!—then there is great cause for concern about the future direction and robustness of the project.

I have more to say concerning the rather hostile welcome given to a first-time contributor to the project, but I suppose that this is not the place for it. There would very likely be even more hostility hurled my way on top of that which no doubt is already brewing as you read this and begin formulating your defense. Regardless, my patience with you has run out, and my time (and no doubt yours too) is better spent elsewhere. If you had the energy to scrutinize a relatively inconsequential sample unit file to the degree of over 30 comments so far, then surely you have the exertional wherewithal to craft your own.

Farewell.

erciccione:
Hi @aquilacooper seems like you are not used to open collaborative projects. You have to understand that working in a community is complex and slower than working by yourself or in a structured company. People have their own opinions and they are discussing them. You should be happy that so many devs are interested in your PR and are spending their time writing suggestions and discussing how to improve your changes :).

As far as i can see the discussion is civil and people are just trying to improve your PR, sharing their experience and opinions. This is how collective development works, but i understand it's not for everybody. If you just want to have your changes merged without discussion and reading other's opinions and share your thoughts with them is an effort you don't want to go through, then indeed this is probably not the right place for you and closing this PR is the way to go. Thank you for your time anyway, i hope you'll be back with more patience and ready to receive criticism/feedback. I assure you will be welcome :)

PR #7331, opened 2021-01-20
Updated copyright date in README.md

joe-sonrichard:
Changed latest year from 2020 -> 2021. A minor change I know, but what an easy opportunity to get a change into something as cool as Monero!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Monero
 
-Copyright (c) 2014-2020 The Monero Project.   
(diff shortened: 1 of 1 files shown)


joe-sonrichard:
Looks like this and probably all other PRs are failing because https://download.qt.io/official_releases/qt/5.15/5.15.1/submodules is taking too long to respond/not available.

selsta (approved):

Issue #7337, opened 2021-01-21
Fee is not carved out of partial output with split transactions


lxop:
When a transfer is split into multiple transactions and one of the addresses is only partially paid in the first transaction, the fee for the first transaction should be carved out of that partial payment (according to the comments on `wallet2::create_transactions_2`). The logic for this behaviour exists, but it is never hit because of the test
```c++
if (inputs < outputs)
{
  LOG_PRINT_L2("We don't have enough for the basic fee, switching to adding_fee");
  adding_fee = true;
  goto skip_tx;
}
```
which is before that logic gets a chance to run and will always be hit if that carving logic would have been used.

The consequence of this bug is that a whole extra input is added to the transaction to pay the fee and the rest of that input value is locked up as change.

NB: I have also observed this behaviour in reality while experimenting with Monero transactions, it is not just my theory about the code :-)

The code that is causing this problem was added in #2989.

Perhaps the solution to this is to perform the "carve it out of the partial payment" logic *before* calling `transfer_selected[_rct]` (and possibly also afterwards, where it currently is)?

PR #348, opened 2015-07-21
wallet: new transaction construction algorithm

moneromooo-monero:
It should avoid a lot of the issues sending more than half the
wallet's contents due to change.

Actual output selection is still random. Changing this would
improve the matching of transaction amounts to output sizes,
but may have non obvious effects on blockchain analysis.

Mapped to the new transfer_new command in simplewallet, and
transfer uses the existing algorithm.

To use in RPC, add "new_algorithm: true" in the transfer_split
JSON command. It is not used in the transfer command.

PR #7342, opened 2021-01-22
Attempt to carve the fee from a partial payment early

lxop:
This re-establishes the behaviour added by #348 whereby the wallet attempts to fund the tx fee by reducing the amount of a partial payment if one exists in the tx.

Fixes: #7337


hyc:
A user manataw on IRC complaining about a fee problem has confirmed that this patch solved his problem.

lxop:
The tests are all passing, but instabilities & other external problems have prevented both CI checks from passing at the same time. As such I'm having to make meaningless changes to the branch to trigger the tests to rerun (sigh).

selsta:
> I'm having to make meaningless changes to the branch to trigger the tests to rerun (sigh).

This is not necessary, we will ignore failed CI if it fails for unrelated reasons :)

selsta (approved):

PR #7345, opened 2021-01-23
Change p2p connection map from raw pointers to weak_ptrs

vtnerd:
This converts the raw pointers p2p connection map to `std::weak_ptr` usage. This helps with code clarity (clear ownership behavior), and reduces the risk of an omitted `finish_outer_call` which would cause a memory leak. There are a slight performance increases in several cases, except for a few lesser used functions. There's a decent amount of changes, particularly in the tests, to re-arrange code so that the connection map can have a proper `weak_ptr` without additional allocations. 

After #7308 , `foreach` and `for_connection` are no longer faster with raw pointers - this PR is actually faster (one less allocation, two less lock acquires, and one less atomic CAS per connection). `invoke`, `invoke_async`, and `notify` all have two less lock acquires and one less atomic CAS, and the last two have one less allocation. `del_functions`, `cancel` and `reset_timer` are slower due to additional atomic operations. Only `reset_timer` is called with regularity.

This was tested with clang thread-sanitizer and address-sanitizer with a daemon running on testnet.

In the future, block/tx relaying performance can be bumped marginally by allowing `std::weak_ptr` to be used outside of the p2p class instead of `boost::uuids::uuid`. This removes the second lock and lookup - the benefit primarily being less thread-stalls.

Also, the `foreach_connection` and `for_connection` calls with this patch or #7308 , can invoke the callback after releasing the connections lock. The calls into this function will have to be inspected for possible race conditions, because current usage may be relying on lock behavior (in fact, they definitely are). The benefit is less thread blocking since the callback performance won't impact wait-time for another thread that needs to inspect the connections list.

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -370,9 +370,9 @@ namespace net_utils
         {
           m_state.stat.in.throttle.handle_trafic_exact(bytes_transferred);
           const auto speed = m_state.stat.in.throttle.get_current_speed();
-          m_conn_context.m_current_speed_down = speed;
-          m_conn_context.m_max_speed_down = std::max(
-            m_conn_context.m_max_speed_down,
+          get_context().m_current_speed_down = speed;
+          get_context().m_max_speed_down = std::max(
+            get_context().m_max_speed_down,
             speed
           );
           if (speed_limit_is_enabled()) {
@@ -383,8 +383,8 @@ namespace net_utils
             ).handle_trafic_exact(bytes_transferred);
           }
           connection_basic::logger_handle_net_read(bytes_transferred);
-          m_conn_context.m_last_recv = time(NULL);
-          m_conn_context.m_recv_cnt += bytes_transferred;
+          get_context().m_last_recv = time(NULL);
+          get_context().m_recv_cnt += bytes_transferred;
           start_timer(get_timeout_from_bytes_read(bytes_transferred), true);
         }
         handle_read(bytes_transferred);
@@ -428,7 +428,7 @@ namespace net_utils
       [this, self, bytes_transferred]{
         bool success = false;
         TRY_ENTRY();
-        success = m_handler.handle_recv(
+        success = get_protocol_handler().handle_recv(
           reinterpret_cast<char *>(m_state.data.read.buffer.data()),
           bytes_transferred
         );
@@ -546,9 +546,9 @@ namespace net_utils
         {
           m_state.stat.out.throttle.handle_trafic_exact(bytes_transferred);
           const auto speed = m_state.stat.out.throttle.get_current_speed();
-          m_conn_context.m_current_speed_up = speed;
-          m_conn_context.m_max_speed_down = std::max(
-            m_conn_context.m_max_speed_down,
+          get_context().m_current_speed_up = speed;
+          get_context().m_max_speed_down = std::max(
+            get_context().m_max_speed_down,
             speed
           );
           if (speed_limit_is_enabled()) {
@@ -559,8 +559,8 @@ namespace net_utils
             ).handle_trafic_exact(bytes_transferred);
           }
           connection_basic::logger_handle_net_write(bytes_transferred);
-          m_conn_context.m_last_send = time(NULL);
-          m_conn_context.m_send_cnt += bytes_transferred;
+          get_context().m_last_send = time(NULL);
+          get_context().m_send_cnt += bytes_transferred;
 
           start_timer(get_default_timeout(), true);
         }
@@ -678,7 +678,7 @@ namespace net_utils
       return;
     m_state.protocol.wait_release = true;
     m_state.lock.unlock();
-    m_handler.release_protocol();
+    get_protocol_handler().release_protocol();
     m_state.lock.lock();
     m_state.protocol.wait_release = false;
     m_state.protocol.released = true;
@@ -854,7 +854,7 @@ namespace net_utils
         return true;
 
       if (m_connection_type == e_connection_type_P2P) {
-        MWARNING("Connection " << m_conn_context.m_connection_id << " tripped write limit, terminating");
+        MWARNING("Connection " << get_context().m_connection_id << " tripped write limit, terminating");
         terminate_async();
         return false;
       }
@@ -997,7 +997,7 @@ namespace net_utils
     if (ec.value())
       return false;
     connection_basic::m_is_multithreaded = is_multithreaded;
-    m_conn_context.set_details(
+    get_context().set_details(
       boost::uuids::random_generator()(),
       *real_remote,
       is_income,
@@ -1021,7 +1021,7 @@ namespace net_utils
     );
     m_state.protocol.wait_init = true;
     guard.unlock();
-    m_handler.after_init_connection();
+    static_cast<shared_state&>(connection_basic::get_state()).after_init_connection(connection<T>::shared_from_this());
     guard.lock();
     m_state.protocol.wait_init = false;
     m_state.protocol.initialized = true;
@@ -1065,13 +1065,13 @@ namespace net_utils
     t_connection_context&& initial
   ):
     connection_basic(io_context, std::move(socket), shared_state, ssl_support),
-    m_handler(this, *shared_state, m_conn_context),
+    service_endpoint<T>(check_and_get(shared_state)),
     m_connection_type(connection_type),
     m_io_context{io_context},
-    m_conn_context(std::move(initial)),
     m_strand{m_io_context},
     m_timers{m_io_context}
   {
+    get_context() = std::move(initial);
   }
 
   template<typename T>
@@ -1126,7 +1126,7 @@ namespace net_utils
       " connection type " << std::to_string(m_connection_type) <<
       " " << connection_basic::socket().local_endpoint().address().to_string() <<
       ":" << connection_basic::socket().local_endpoint().port() <<
-      " <--> " << m_conn_context.m_remote_address.str() <<
+      " <--> " << get_context().m_remote_address.str() <<
       " (via " << address << ":" << port << ")"
     );
   }
@@ -1175,7 +1175,7 @@ namespace net_utils
     // stopping the server, we don't want the io_context to stop before the shutdown sequence completes, since we
     // execute terminate inside m_strand. So we wait for the connection's shutdown sequence to complete before stopping
     // the io_context.
-    MDEBUG("Waiting for connection " << m_conn_context.m_connection_id << " to shutdown, current state: " << m_state.status);
+    MDEBUG("Waiting for connection " << get_context().m_connection_id << " to shutdown, current state: " << m_state.status);
     const bool shutdown = m_state.condition.wait_for(
       m_state.lock,
       std::chrono::seconds(5),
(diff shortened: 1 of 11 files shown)


selsta:
One suggestion from IRC:

```patch
From 5068277fe6dc549046068101273667641b93d35f Mon Sep 17 00:00:00 2001
From: anon <anon [at] nowhere>
Date: Sat, 23 Jan 2021 07:05:29 +0000
Subject: [PATCH] async_protocol_handler: replace incorrect assert

---
 contrib/epee/include/net/levin_protocol_handler_async.h | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/contrib/epee/include/net/levin_protocol_handler_async.h b/contrib/epee/include/net/levin_protocol_handler_async.h
index 63a2bfdb9..26e2452a8 100644
--- a/contrib/epee/include/net/levin_protocol_handler_async.h
+++ b/contrib/epee/include/net/levin_protocol_handler_async.h
@@ -577,7 +577,7 @@ public:
   template<class callback_t>
   bool async_invoke(std::shared_ptr<net_utils::service_endpoint<async_protocol_handler>> self, int command, const epee::span<const uint8_t> in_buff, const callback_t &cb, size_t timeout = LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED)
   {
-    assert(this == self.get());
+    assert(this == &(self.get()->m_protocol_handler));
 
     if(timeout == LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED)
       timeout = m_config.m_invoke_timeout;
-- 
2.30.0
```

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -177,29 +176,35 @@ PRAGMA_WARNING_DISABLE_VS(4355)
     // create a random uuid, we don't need crypto strength here
     const boost::uuids::uuid random_uuid = boost::uuids::random_generator()();
 
-    context = t_connection_context{};
+    get_context() = t_connection_context{};
```
iamamyth, line 179: Seems you could reduce delta noise in this function and also help the compiler with:
```
auto context = get_context() = t_connection_context{};
```

vtnerd: This hides a member variable from a derived class, but it does limit the changes. Done.

vtnerd:
Rebased to master changes.

vtnerd:
Rebased again.

vtnerd:
Rebased

vtnerd:
Rebased to include changes on TCP server. Re-tested with asan and tsan on `monerod` and `unit_tests`. Includes `unit_test` memory leak fix from #8538 .

vtnerd:
Force pushed a rebase update on the tests.

vtnerd:
Updated with another rebase.

Gingeropolous:
so my daemon with this on xmrchain has been OOM killed. I am switching back to master to see if it also gets OOM killed, but I don't remember that happening without this PR.

vtnerd:
That's not good. Sounds like a possible memory leak, likely some `shared_ptr` cyclic reference.

How long was the server up roughly?

Gingeropolous:
hours. from what I can gather from systemd logs, 4 hours, and then 8 hours. 

though I just remembered / noticed that my systemd file has some memory management.

```
MemoryAccounting=yes
MemoryHigh=40G
MemoryMax=40G
```

whereas my home server doesn't have that setup. I may fiddle with that to see if it stays up longer on the xmrchain server... but again, I never noticed my monerod getting oom killed before I pulled in this PR.

vtnerd:
According to the initial comment, I ran address sanitizer against this, so it's not obvious what is going on here. Maybe the data structures aren't being pruned properly until shutdown.

Gingeropolous:
well, the daemon with this patch has now been up for 31 hours, and I've disabled the systemd memory management. I think it just uses more memory than the normal version, and that amount hit the memory cap I had placed on the process in systemd.  I'm re-enabling the memory management and running again.

Gingeropolous:
So it hit the memory limit of 40G and was restarted. I'm going to increase it to 50GB. 

Is this expected behavior on a 64GB system? I know monero can run on potato hardware. I guess the behavior I'm seeing is not related to this patch.

selsta:
@Gingeropolous definitely not expected behavior. It's unclear from your messages if you also were able to reproduce this issue with just master or the latest release version?

I'm not aware of any OOM bug reports in the recent months so this isn't anything widespread and the majority of nodes have way less than 40GB RAM.

vtnerd:
I agree with @selsta , there's likely something wrong with the patch, but it's going to be difficult to find out where.

jeffro256:
Could you please mark this PR as a draft while merge conflicts and the OOM issue is sorted out?

vtnerd:
I don't see a way to do this in the UI. I think it would require closing this PR and re-opening a draft PR, which I don't want to do.

selsta:
I added a tag for it.

vtnerd:
I rebased the changes just before all of the madness started, and began running this locally for various reasons. After 8 days the system is only using 13 GiB of memory and has not him OOM killer. This is also through all of the reorg stuff.

I'm just not seeing any memory issue, but I'll do another sweep with asan at some point.

FWIW this handles cleanup of connections in a much better way; there is an outstanding bug on new boost+gcc+rdrand bug on ryzen 3 (yes really), that is handled poorly with the existing approach. It doesn't matter too much realistically as the bug is so bad that you can't maintain any connections.

vtnerd:
Got a little ahead of myself - finally saw the OOM killer on day 9. This will be difficult to find, as stated.

vtnerd:
Rebased + add fix that hopefully handles OOM issue (thanks @selsta !), and cleaned up the async handler a bit more.

vtnerd:
Another rebase after some changes on `master`.

vtnerd:
Another rebase.

This patch has been running for 11 days without issue, and stable memory usage. Looks like the memory leak has been solved.

tests/unit_tests/levin.cpp
```diff
@@ -816,9 +841,15 @@ TEST_F(levin_notify, local_without_padding)
     bool has_fluffed = false;
     while (!has_stemmed || !has_fluffed)
     {
+/*<<<<<<< HEAD
```
jpk68: This isn't supposed to be here, is it?

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -733,60 +708,55 @@ void async_protocol_handler_config<t_connection_context>::del_in_connections(siz
...
   CRITICAL_REGION_END();
-  m_pcommands_handler->on_connection_new(pconn->m_connection_context);
-}
-//------------------------------------------------------------------------------------------
-template<class t_connection_context>
-async_protocol_handler<t_connection_context>* async_protocol_handler_config<t_connection_context>::find_connection(boost::uuids::uuid connection_id) const
-{
-  auto it = m_connects.find(connection_id);
-  return it == m_connects.end() ? 0 : it->second;
+  m_pcommands_handler->on_connection_new(pconn->context);
```
jpk68: Should this have a null check (like is done on line 670)?

tests/unit_tests/levin.cpp
```diff
@@ -164,43 +153,78 @@ namespace
...
+
         //\return Number of messages processed
         std::size_t process_send_queue(const bool valid = true)
         {
             std::size_t count = 0;
-            for ( ; !endpoint_.send_queue_.empty(); ++count, endpoint_.send_queue_.pop_front())
+            for ( ; !send_queue_.empty(); ++count, send_queue_.pop_front())
             {
-                EXPECT_EQ(valid, handler_.handle_recv(endpoint_.send_queue_.front().data(), endpoint_.send_queue_.front().size()));
+                // invalid messages shoudn't be possible in this test;
```
jpk68: Nit: typo

contrib/epee/include/net/http_protocol_handler.h
```diff
@@ -64,6 +64,14 @@ namespace net_utils
 			std::size_t m_max_private_ip_connections{25};
 			std::size_t m_max_connections{100};
 			critical_section m_lock;
+
+			template<typename T>
+			static bool after_init_connection(const std::shared_ptr<T>& self)
+			{
+        if (!self)
```
jpk68: Nit: indentation

contrib/epee/include/net/http_protocol_handler.h
```diff
@@ -96,7 +104,9 @@ namespace net_utils
 			{
 				return true;
 			}
-			bool after_init_connection();
+
+      bool after_init_connection();
```
jpk68: Same as above.

jpk68:
Other than the few comments I left, these changes look fine to me. Apologies for not leaving all comments in one "review", I somehow forgot that existed :)

tests/unit_tests/levin.cpp
vtnerd: Removed.

contrib/epee/include/net/levin_protocol_handler_async.h
vtnerd: Can't hurt.

tests/unit_tests/levin.cpp
vtnerd: Fixed.

contrib/epee/include/net/http_protocol_handler.h
vtnerd: Fixed.

contrib/epee/include/net/http_protocol_handler.h
vtnerd: Fixed.

tests/unit_tests/levin.cpp
```diff
@@ -816,8 +841,8 @@ TEST_F(levin_notify, local_without_padding)
     bool has_fluffed = false;
     while (!has_stemmed || !has_fluffed)
     {
-        // run their "their" txes first
-        auto context = contexts_.begin();
+	// run their "their" txes first
```
jpk68: Nit: indentation

tests/unit_tests/levin.cpp
```diff
@@ -1231,9 +1256,15 @@ TEST_F(levin_notify, local_with_padding)
...
     {
+/* <<<<<<< HEAD
       // run their "their" txes first
         auto context = contexts_.begin();
         EXPECT_TRUE(notifier.send_txs(their_txs, context->get_id(), cryptonote::relay_method::stem));
+======= */
+      // run their "their" txes first
+        auto context = boost::make_indirect_iterator(contexts_.begin());
+        EXPECT_TRUE(notifier.send_txs(their_txs, context->get_id(), cryptonote::relay_method::stem));
+//>>>>>>> 4538bc0ed (Change p2p connection map from raw pointers to weak_ptrs)
```
jpk68: Same as from earlier, not sure if this is supposed to be here

jpk68 (approved):

selsta (approved):

vtnerd:
@selta needs back port to ... 0.19 ?

vtnerd:
@selsta correct person this time - see last comment, this should go on 0.19?

selsta:
Yes, so that we don't have diverging branches directly from the start.

PR #7346, opened 2021-01-23
Remove unused variables

kevinbarbour:
There are quite a few variables in the code that are no longer (or perhaps never were) in use. These were discovered by enabling compiler warnings for unused variables and cleaning them up.

In most cases where the unused variables were the result of a function call the call was left but the variable assignment removed, unless it was obvious that it was a simple getter with no side effects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -629,7 +629,7 @@ else()
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ARCH_FLAG}")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ARCH_FLAG}")
 
-  set(WARNINGS "-Wall -Wextra -Wpointer-arith -Wundef -Wvla -Wwrite-strings -Wno-error=extra -Wno-error=deprecated-declarations -Wno-unused-parameter -Wno-unused-variable -Wno-error=unused-variable -Wno-error=undef -Wno-error=uninitialized")
+  set(WARNINGS "-Wall -Wextra -Wpointer-arith -Wundef -Wvla -Wwrite-strings -Wno-error=extra -Wno-error=deprecated-declarations -Wno-unused-parameter -Wno-error=unused-variable -Wno-error=undef -Wno-error=uninitialized")
   if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
     if(ARM)
       set(WARNINGS "${WARNINGS} -Wno-error=inline-asm")
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -897,7 +897,6 @@ uint64_t BlockchainLMDB::add_transaction_data(const crypto::hash& blk_hash, cons
     throw0(DB_ERROR(lmdb_error("Failed to add tx data to db transaction: ", result).c_str()));
 
   const cryptonote::blobdata_ref &blob = txp.second;
-  MDB_val_sized(blobval, blob);
 
   unsigned int unprunable_size = tx.unprunable_size;
   if (unprunable_size == 0)
@@ -3190,9 +3189,8 @@ bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_block_cou
   uint64_t size = 0;
   size_t num_txes = 0;
   MDB_val_copy<uint64_t> key(start_height);
-  MDB_val k, v, val_tx_id;
+  MDB_val v, val_tx_id;
   uint64_t tx_id = ~0;
-  MDB_cursor_op op = MDB_SET;
   for (uint64_t h = start_height; h < blockchain_height && blocks.size() < max_block_count && (size < max_size || blocks.size() < min_block_count); ++h)
   {
     MDB_cursor_op op = h == start_height ? MDB_SET : MDB_NEXT;
@@ -3314,7 +3312,7 @@ bool BlockchainLMDB::get_prunable_tx_hash(const crypto::hash& tx_hash, crypto::h
   RCURSOR(txs_prunable_hash);
 
   MDB_val_set(v, tx_hash);
-  MDB_val result, val_tx_prunable_hash;
+  MDB_val result;
   auto get_result = mdb_cursor_get(m_cur_tx_indices, (MDB_val *)&zerokval, &v, MDB_GET_BOTH);
   if (get_result == 0)
   {
@@ -4311,7 +4309,6 @@ bool BlockchainLMDB::get_output_distribution(uint64_t amount, uint64_t from_heig
     return false;
   distribution.resize(db_height - from_height, 0);
 
-  bool fret = true;
   MDB_val_set(k, amount);
   MDB_val v;
   MDB_cursor_op op = MDB_SET;
@@ -5106,11 +5103,10 @@ void BlockchainLMDB::migrate_0_1()
 void BlockchainLMDB::migrate_1_2()
 {
   LOG_PRINT_L3("BlockchainLMDB::" << __func__);
-  uint64_t i, z;
+  uint64_t i;
   int result;
   mdb_txn_safe txn(false);
-  MDB_val k, v;
-  char *ptr;
+  MDB_val v;
 
   MGINFO_YELLOW("Migrating blockchain from DB version 1 to 2 - this may take a while:");
   MINFO("updating txs_pruned and txs_prunable tables...");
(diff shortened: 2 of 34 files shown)


src/gen_multisig/gen_multisig.cpp
```diff
@@ -74,8 +74,6 @@ namespace
   const command_line::arg_descriptor<bool, false> arg_testnet = {"testnet", genms::tr("Create testnet multisig wallets"), false};
   const command_line::arg_descriptor<bool, false> arg_stagenet = {"stagenet", genms::tr("Create stagenet multisig wallets"), false};
   const command_line::arg_descriptor<bool, false> arg_create_address_file = {"create-address-file", genms::tr("Create an address file for new wallets"), false};
-
-  const command_line::arg_descriptor< std::vector<std::string> > arg_command = {"command", ""};
```
iamamyth, line 78: Not sure why you'd want to remove the command argument here, seems like a mistake.

kevinbarbour: As far as I can tell that argument was not in use at all. If you look below, it is never added as an argument like the others. 

https://github.com/monero-project/monero/blob/56bbfd2921cc9f2a3a619de1481c01f28de01bc3/src/gen_multisig/gen_multisig.cpp#L166-L173

iamamyth: My bad, misread the surrounding context.

src/wallet/wallet2.cpp
```diff
@@ -10303,7 +10300,7 @@ bool wallet2::sanity_check(const std::vector<wallet2::pending_tx> &ptx_vector, s
 {
   MDEBUG("sanity_check: " << ptx_vector.size() << " txes, " << dsts.size() << " destinations");
 
-  hw::device &hwdev = m_account.get_device();
+  m_account.get_device();
```
xiphon, line 10303: Drop the line entirely

src/wallet/wallet2.cpp
```diff
@@ -7984,7 +7981,7 @@ void wallet2::light_wallet_get_outs(std::vector<std::vector<tools::wallet2::get_
   
   // Check if we got enough outputs for each amount
   for(auto& out: ores.amount_outs) {
-    const uint64_t out_amount = boost::lexical_cast<uint64_t>(out.amount);
+    boost::lexical_cast<uint64_t>(out.amount);
```
xiphon, line 7984: ?

src/wallet/wallet2.cpp
```diff
@@ -367,7 +367,7 @@ std::unique_ptr<tools::wallet2> make_basic(const boost::program_options::variabl
 
   // user specified CA file or fingeprints implies enabled SSL by default
   epee::net_utils::ssl_options_t ssl_options = epee::net_utils::ssl_support_t::e_ssl_support_enabled;
-  if (command_line::get_arg(vm, opts.daemon_ssl_allow_any_cert))
+  if (daemon_ssl_allow_any_cert)
```
xiphon, line 370: Why not just drop unused `daemon_ssl_allow_any_cert` variable declaration instead?

src/wallet/message_store.cpp
```diff
@@ -1340,7 +1340,7 @@ bool message_store::check_for_messages(const multisig_wallet_state &state, std::
     }
   }
   std::vector<transport_message> transport_messages;
-  bool r = m_transporter.receive_messages(destinations, transport_messages);
+  m_transporter.receive_messages(destinations, transport_messages);
```
xiphon, line 1343: Return `false` if `receive_messages` failed.

src/simplewallet/simplewallet.cpp
```diff
@@ -1357,7 +1357,7 @@ bool simple_wallet::import_multisig_main(const std::vector<std::string> &args, b
   {
     m_in_manual_refresh.store(true, std::memory_order_relaxed);
     epee::misc_utils::auto_scope_leave_caller scope_exit_handler = epee::misc_utils::create_scope_leave_handler([&](){m_in_manual_refresh.store(false, std::memory_order_relaxed);});
-    size_t n_outputs = m_wallet->import_multisig(info);
+    m_wallet->import_multisig(info);
     // Clear line "Height xxx of xxx"
     std::cout << "\r                                                                \r";
     success_msg_writer() << tr("Multisig info imported");
```
xiphon, line 1363: Add the number of outputs updated to the log message

src/ringct/multiexp.cc
```diff
@@ -450,11 +449,9 @@ rct::key straus(const std::vector<MultiexpData> &data, const std::shared_ptr<str
   STEP = STEP ? STEP : 192;
 
   MULTIEXP_PERF(PERF_TIMER_START_UNIT(setup, 1000000));
-  static constexpr unsigned int mask = (1<<STRAUS_C)-1;
```
xiphon, line 453: It is actually used inside `#elif 1` condition below.

src/wallet/wallet2.cpp
kevinbarbour: Just did it this way for the sake of consistency. All the other surrounding arguments were assigned to variables before use.

src/ringct/multiexp.cc
kevinbarbour: Good catch, overlooked that one.

src/wallet/wallet2.cpp
kevinbarbour: Should have removed entirely, fixed.

src/wallet/message_store.cpp
```diff
@@ -1340,7 +1340,9 @@ bool message_store::check_for_messages(const multisig_wallet_state &state, std::
     }
   }
   std::vector<transport_message> transport_messages;
-  bool r = m_transporter.receive_messages(destinations, transport_messages);
+  if (!m_transporter.receive_messages(destinations, transport_messages)) {
+    return false;
+  }
```
xiphon, line 1345: Styling remark. Insert a line break before the opening curly bracket for consistency.

CMakeLists.txt
```diff
@@ -629,7 +629,7 @@ else()
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ARCH_FLAG}")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ARCH_FLAG}")
 
-  set(WARNINGS "-Wall -Wextra -Wpointer-arith -Wundef -Wvla -Wwrite-strings -Wno-error=extra -Wno-error=deprecated-declarations -Wno-unused-parameter -Wno-unused-variable -Wno-error=unused-variable -Wno-error=undef -Wno-error=uninitialized")
+  set(WARNINGS "-Wall -Wextra -Wpointer-arith -Wundef -Wvla -Wwrite-strings -Wno-error=extra -Wno-error=deprecated-declarations -Wno-unused-parameter -Wno-error-unused-variable -Wno-error=undef -Wno-error=uninitialized")
```
xiphon, line 632: Should be `-Wno-error=unused-variable`

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -5311,7 +5307,6 @@ void BlockchainLMDB::migrate_2_3()
         if (result)
           throw0(DB_ERROR(lmdb_error("Failed to open a cursor for block_info: ", result).c_str()));
         if (!i) {
-          MDB_stat db_stat;
```
moneromooo-monero, line 5314: That one was not unused. Removing it will instead cause a write to another in a previous scope. Which happens to be fine in practice in this case though.

src/cryptonote_core/blockchain.cpp
```diff
@@ -3438,7 +3438,7 @@ bool Blockchain::check_tx_inputs(transaction& tx, tx_verification_context &tvc,
       bool failed = false;
       for (size_t i = 0; i < tx.vin.size(); i++)
       {
-        const txin_to_key& in_to_key = boost::get<txin_to_key>(tx.vin[i]);
+        boost::get<txin_to_key>(tx.vin[i]);
```
moneromooo-monero, line 3441: This seems pointless to keep.

moneromooo-monero (approved):

src/blockchain_db/lmdb/db_lmdb.cpp
kevinbarbour: I believe this one was actually unused, it appears to be using `db_stats` everywhere, never `db_stat`

moneromooo-monero: Indeed, I did ready too quickly :)

selsta:
Please squash your commits.

iamamyth (approved):

PR #7349, opened 2021-01-25
7344 Fix describe_transfer in the case of multiple transactions in the set

### LINKED ISSUE #7344: describe_transfer has incorrect output when multiple transactions are present
When `transfer_split` is used on a view-only wallet, it produces an unsigned transaction set. As implied by the term "set", this can comprise multiple transactions; this can occur when the required inputs mean that a single transaction would be too large. (Once #7322 is merged, it will also occur if more than 15 outputs are specified).

If this transaction set is inspected with `describe_transfer`, the response will contain multiple lists of recipients (one list per transaction) but the lists are incorrect. Specifically, the first list will correctly show the recipients from the first transaction, but the second list will contain the recipients from both the first and the second transactions, the third list contains recipients from the first, second, and third transactions, and so forth: each list contains the aggregated outputs from all transactions so far instead of just its own.

lxop:
- Reset the recipients map for each transaction so that the output only shows the recipients for each individual transaction in the set.
- Add a field to the output of `describe_transfer` that includes the total amounts and the aggregated set of recipients.

Fixes #7344

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -1197,13 +1197,16 @@ namespace tools
     try
     {
       // gather info to ask the user
-      std::unordered_map<cryptonote::account_public_address, std::pair<std::string, uint64_t>> dests;
+      std::unordered_map<cryptonote::account_public_address, std::pair<std::string, uint64_t>> tx_dests;
+      std::unordered_map<cryptonote::account_public_address, std::pair<std::string, uint64_t>> all_dests;
       int first_known_non_zero_change_index = -1;
       for (size_t n = 0; n < tx_constructions.size(); ++n)
       {
         const tools::wallet2::tx_construction_data &cd = tx_constructions[n];
         res.desc.push_back({0, 0, std::numeric_limits<uint32_t>::max(), 0, {}, "", 0, "", 0, 0, ""});
         wallet_rpc::COMMAND_RPC_DESCRIBE_TRANSFER::transfer_description &desc = res.desc.back();
+        // Clear the recipients collection ready for this loop iteration
+        tx_dests.clear();
 
         std::vector<cryptonote::tx_extra_field> tx_extra_fields;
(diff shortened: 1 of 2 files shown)


src/wallet/wallet_rpc_server.cpp
```diff
@@ -1279,30 +1282,46 @@ namespace tools
...
           else
             ++desc.dummy_outputs;
-          ++i;
         }
 
         if (desc.change_amount > 0)
         {
           const tools::wallet2::tx_construction_data &cd0 = tx_constructions[0];
           desc.change_address = get_account_address_as_str(m_wallet->nettype(), cd0.subaddr_account > 0, cd0.change_dts.addr);
+          res.summary.change_address = desc.change_address;
```
moneromooo-monero, line 1308: That address, if any, could be different for all txes.

moneromooo-monero (approved):
The dests renaming is a bit spammy but small enough I guess.

src/wallet/wallet_rpc_server.cpp
lxop: If that were the case an error would be triggered on line 1278, so it is safe to assume all change goes to one address here.

lxop:
> The dests renaming is a bit spammy but small enough I guess.

@moneromooo-monero I renamed `dests` -> `tx_dests` because there are now two "destinations" maps and I wanted more clarity about the purpose of each of them and which one is being used on any given line.

selsta:
Can you bump WALLET_RPC_VERSION_MINOR ?

selsta:
Thanks, please squash your commits and then we can add this to the merge list.

moneromooo-monero (approved):

selsta:
@lxop could you please rebase? src/wallet/wallet_rpc_server.cpp conflicts due to a recent merge

selsta:
ping @lxop, I would appreciate it if you could rebase so that we can merge this.

PR #7350, opened 2021-01-25
fix serialization being different on mac

moneromooo-monero:
On Mac, size_t is a distinct type from uint64_t, and some
types (in wallet cache as well as cold/hot wallet transfer
data) use pairs/containers with size_t as fields. Mac would
save those as full size, while other platforms would save
them as varints. Might apply to other platforms where the
types are distinct.

There's a nasty hack for backward compatibility, which can
go after a couple forks.

### DIFF
--- a/src/cryptonote_core/cryptonote_tx_utils.h
+++ b/src/cryptonote_core/cryptonote_tx_utils.h
@@ -44,10 +44,10 @@ namespace cryptonote
     typedef std::pair<uint64_t, rct::ctkey> output_entry;
 
     std::vector<output_entry> outputs;  //index + key + optional ringct commitment
-    size_t real_output;                 //index in outputs vector of real output_entry
+    uint64_t real_output;               //index in outputs vector of real output_entry
     crypto::public_key real_out_tx_key; //incoming real tx public key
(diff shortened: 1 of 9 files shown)


xmrdog:
Hi again :) The following happens on macOS (using the same wallet as before):

1. Delete wallet cache (leaving only .keys).
2. Open on v0.17.1.9 and full refresh.
3. Exit.
4. Open with your new commit. **It fails with std::bad_alloc.**

I'm sorry but this is now like the 5th time we do these back&forths :)

Can we start a funding round to get @moneromooo-monero a Mac? I'd be happy to chip in.

These back&forths make no sense to me anymore, slow things down, and are not the most solid way of developing cross-platform software IMHO.

moneromooo-monero:
Fair enough. I'm not buying anything on the internet. Requires private info and/or risk of pwned hardware. It'll have to wait for someone with a mac to finish it then.

moneromooo-monero:
selsta reminded me thre's a mac build machine I can ssh in. I'll try and fix it there some time later.

moneromooo-monero:
Simple typo in the change from review. Works on the mac build machine.

selsta (approved):

Issue #7179, opened 2020-12-20
Mining for rpc within monero wallet cli: set number of threads


t-900-a:
Low priority request, until rpc pay is adopted for other things (i.e. games, multimedia, website access) this feature is not needed

When calling start_mining_for_rpc
allow the user to specify the same argument as they do start_mining  [<number_of_threads>] [bg_mining] [ignore_battery]

Stackexchange thread:
https://monero.stackexchange.com/questions/12194/mining-for-rpc-set-number-of-threads

PR #7358, opened 2021-01-31
wallet_rpc_payments: implement multithreading

gdmojo:
Opening PR for early review. Will solve issue #7179

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -276,7 +276,7 @@ namespace
   const char* USAGE_PUBLIC_NODES("public_nodes");
   const char* USAGE_WELCOME("welcome");
   const char* USAGE_RPC_PAYMENT_INFO("rpc_payment_info");
-  const char* USAGE_START_MINING_FOR_RPC("start_mining_for_rpc");
+  const char* USAGE_START_MINING_FOR_RPC("start_mining_for_rpc [<number_of_threads>]");
   const char* USAGE_STOP_MINING_FOR_RPC("stop_mining_for_rpc");
   const char* USAGE_SHOW_QR_CODE("show_qr_code [<subaddress_index>]");
   const char* USAGE_VERSION("version");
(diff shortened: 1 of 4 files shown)


moneromooo-monero:
I hate the reindent diff, but I guess it has to be. Seems OK, except hardcoding threads to 4 is not good, but easy to change.

gdmojo:
Implemented user specified threads defaulting to 1.

src/simplewallet/simplewallet.cpp
```diff
@@ -2360,6 +2360,25 @@ bool simple_wallet::start_mining_for_rpc(const std::vector<std::string> &args)
...
+  else
+  {
+    m_rpc_payment_threads = 1;
+  }
+
+  if (!ok)
+  {
+    PRINT_USAGE(USAGE_START_MINING_FOR_RPC);
+    return true;
+  }
```
moneromooo-monero, line 2380: This looks like if you run:
> start_mining_for_rpc 1
> start_mining_for_rpc 0

Then second one will fail as ok will be false, but m_rpc_payment_threads will still be set and will find its way into the call to search_for_rpc_payment later. Admittedly it's a "don't do that then" situation :)

SamsungGalaxyPlayer:
Can the default be the max # of threads on the system? Is there an easy way to detect that?

moneromooo-monero:
Yes, boost::thread::hardware_concurrency()

src/simplewallet/simplewallet.cpp
gdmojo: Well spotted.

gdmojo:
Changed `m_rpc_payment_threads` so that value of `0` defaults to `boost::thread::hardware_concurrency` threads.

mj-xmr:
> Changed `m_rpc_payment_threads` so that value of `0` defaults to `boost::thread::hardware_concurrency` threads.

Not super important, but in case you want to know, since C++11, there's also:

std::thread::hardware_concurrency

https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency

moneromooo-monero (approved):

PR #7361, opened 2021-02-01
ITS#9007 don't free loose writemap pages

hyc:
Broken in ITS#8756

PR #7363, opened 2021-02-01
lmdb: add segfault demo

selsta:
Segfault fixed in #7361. Unit test to avoid regression in the future.


perfect-daemon (approved):

selsta:
Closing as this appears unlikely to get reviewed and the test is more lmdb specific than monero.

Issue #7372, opened 2021-02-11
Prevent randomx from polluting monerod log.


crocket:
Unless huge pages are enabled by a command like `sysctl vm.nr_hugepages=1280`, randomx large page allocator keeps printing exceptions in monerod log.

It occupies almost 100% of monerod log. I can't analyze monerod log at all because of the large page allocator exceptions.

This problem should be documented or fixed properly.

selsta:
Fixed in https://github.com/tevador/RandomX/pull/241

PR #7373, opened 2021-02-11
Honor LARGE_PAGES flag in randomx umask

hyc:
Flag in MONERO_RANDOMX_UMASK was being ignored, leaving no way to disable use of large pages.


moneromooo-monero (approved):

crocket:
I don't think it really fixes #7372 until people don't have to fiddle with anything to avoid exceptions or until the workaround is documented somewhere.

PR #7374, opened 2021-02-11
Fix txs_prunable_hash sort functions [DO NOT MERGE]

hyc:
The key is a uint64, the value is a hash

Since the current code is only comparing the first 8 bytes of the 32 byte hash, it is possible
that there have been collisions in txs_prunable_hash table, and possibly dropped records as
a result. As such it seems that the only safe way to fix this is re-import the entire blockchain
so that all the table entries can be recreated.


hyc:
No bug here.

PR #7376, opened 2021-02-11
[DON'T MERGE] Ci test regular

mj-xmr:
This PR will serve me only as a regular CI's health probe.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,8 +4,7 @@
 	branch = monero
 [submodule "external/miniupnp"]
 	path = external/miniupnp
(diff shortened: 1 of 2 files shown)


selsta:
Might make more sense to use scheduled event in your forked repo: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events

mj-xmr:
> Might make more sense to use scheduled event in your forked repo: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events

Thanks. I will explore this.

PR #7380, opened 2021-02-13
easylogging++: fix potential memory corruption

moneromooo-monero:
The m_typedConfigurations pointer is copied from one object to the next,
but deleted in the dtor, leading to potential double free. It is also
deleted first thing in the copy ctor, deleting uninitialized memory.

This does not seem to actually happen in practice (those functions do
not get called), but seems safer that way.

Coverity 1446562

### DIFF
--- a/external/easylogging++/easylogging++.cc
+++ b/external/easylogging++/easylogging++.cc
@@ -713,9 +713,8 @@ Logger::Logger(const std::string& id, const Configurations& configurations,
 }
 
 Logger::Logger(const Logger& logger) {
-  base::utils::safeDelete(m_typedConfigurations);
   m_id = logger.m_id;
-  m_typedConfigurations = logger.m_typedConfigurations;
+  m_typedConfigurations = logger.m_typedConfigurations ? new base::TypedConfigurations(*logger.m_typedConfigurations) : nullptr;
   m_parentApplicationName = logger.m_parentApplicationName;
   m_isConfigured = logger.m_isConfigured;
   m_configurations = logger.m_configurations;
@@ -727,7 +726,7 @@ Logger& Logger::operator=(const Logger& logger) {
   if (&logger != this) {
     base::utils::safeDelete(m_typedConfigurations);
     m_id = logger.m_id;
-    m_typedConfigurations = logger.m_typedConfigurations;
+    m_typedConfigurations = logger.m_typedConfigurations ? new base::TypedConfigurations(*logger.m_typedConfigurations) : nullptr;
     m_parentApplicationName = logger.m_parentApplicationName;
     m_isConfigured = logger.m_isConfigured;
     m_configurations = logger.m_configurations;


external/easylogging++/easylogging++.h
```diff
@@ -1954,6 +1954,12 @@ class TypedConfigurations : public base::threading::ThreadSafe {
   const Configurations* configurations(void) const {
     return m_configurations;
   }
+  Configurations* configurations(void) {
```
iamamyth, line 1957: Why not just make a constructor which accepts a non-const Configurations instance? Exposing these non-const pointers seems to unnecessarily complicate reasoning about object behavior ("any other object might have a ref" is quite different from "this class may steal refs from other instances").

moneromooo-monero: I'm not sure what you want. The ctor has a non const Configuration as parameter. Post the code you want to see.

external/easylogging++/easylogging++.cc
```diff
@@ -713,9 +713,8 @@ Logger::Logger(const std::string& id, const Configurations& configurations,
 }
 
 Logger::Logger(const Logger& logger) {
-  base::utils::safeDelete(m_typedConfigurations);
   m_id = logger.m_id;
-  m_typedConfigurations = logger.m_typedConfigurations;
+  m_typedConfigurations = logger.m_typedConfigurations ? new base::TypedConfigurations(logger.m_typedConfigurations->configurations(), logger.m_typedConfigurations->logStreamsReference()) : NULL;
```
iamamyth, line 717: This would be the proposed change:
```
m_typedConfigurations = logger.m_typedConfigurations ? new base::TypedConfigurations(logger.m_typedConfigurations) : NULL;
```

external/easylogging++/easylogging++.cc
```diff
@@ -713,9 +713,8 @@ Logger::Logger(const std::string& id, const Configurations& configurations,
 }
 
 Logger::Logger(const Logger& logger) {
-  base::utils::safeDelete(m_typedConfigurations);
   m_id = logger.m_id;
-  m_typedConfigurations = logger.m_typedConfigurations;
+  m_typedConfigurations = logger.m_typedConfigurations ? new base::TypedConfigurations(*logger.m_typedConfigurations) : NULL;
```
ndorf, line 717: I think this should be `nullptr`, not `NULL`.

moneromooo-monero: Explain the bug ?

ndorf: No "bug" in this instance, it's just deprecated since C++11 and shouldn't be used in new code.

mj-xmr:
The below patch will protect your changes against regressions. It segfaults on master, but goes fine on your branch:

[ut-logging-segfault.patch.txt](https://github.com/monero-project/monero/files/6064057/ut-logging-segfault.patch.txt)

selsta (approved):

PR #7387, opened 2021-02-19
[CI] Adaptive mining timeout, depending on available CPU power

mj-xmr:
### Intro
A more serious approach to determining the appropriate timeout for the mining test in `functional_tests_rpc`. The timeouts are calculated separately for RX init and for the mining of subsequent blocks themselves. Printing also available RAM, in case the RX init went wrong, which needs above 3 GB of RAM. Added comprehensive description.

### More details:
As per the discussion with @iamamyth in https://github.com/monero-project/monero/pull/7371 , the adjustment is done by testing the time of generation of N next digits of Pi - once for a single core (mining use case) and once for as many cores as there are available (RX init). This test helps orientating in the currently available CPU processing power, even in the event of:
- other jobs taking up the virtual resources at the same time
- `ctest` being ran in parallel (with `-j$(nproc)`)

The obtained measurements serve as multipliers for the assumed timeout constants. For the case of RX init, the number of cores serves additionally as a divisor of the timeout constant, since availability of more cores decreases the init time (ideally) linearly.

### Example runs 
(see the bottom of the `tests` log):
[-1-](https://github.com/mj-xmr/monero/runs/1934048564?check_suite_focus=true) [-2-](https://github.com/mj-xmr/monero/runs/1934758910?check_suite_focus=true) [-3-](https://github.com/mj-xmr/monero/runs/1935221231?check_suite_focus=true) [-4-](https://github.com/mj-xmr/monero/runs/1934786457?check_suite_focus=true)

### Example log:
6: [TEST STARTED] mining
6: Resetting blockchain
6: Creating wallet

6: Test mining via daemon
6: Available RAM = 17.0 GB
6: Measuring the currently available CPU power...
6: Time taken to calculate 10000 Pi digits on **4 core(s) was 2.03 s**.
6: Measuring the currently available CPU power...
6: Time taken to calculate 10000 Pi digits on _1 core(s) was 1.31 s_.
6: Timeout for init,   adjusted for the currently available **CPU power, is 31 s**
6: Timeout for mining, adjusted for the currently available _CPU power, is 27 s_
6: Time taken for RX init + mining 1st block = 12 s.
6: Time taken for total mining = 12 s.

6: Test mining via wallet
6: Available RAM = 16.8 GB
6: Measuring the currently available CPU power...
6: Time taken to calculate 10000 Pi digits on **4 core(s) was 2.13 s.**
6: Measuring the currently available CPU power...
6: Time taken to calculate 10000 Pi digits on _1 core(s) was 1.3 s._
6: Timeout for init,   adjusted for the currently available **CPU power, is 33 s**
6: Timeout for mining, adjusted for the currently available _CPU power, is 27 s_
6: Time taken for RX init + mining 1st block = 11 s.
6: Time taken for total mining = 23 s.

6: Test submitblock
6: Resetting blockchain
6: Test RandomX
6: Recreating the chain
6: Mining from genesis block again
6: Adding one to reorg
6: [TEST PASSED] mining

Work time: ~15 h

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -130,8 +130,8 @@ jobs:
       run: sudo apt update
     - name: install monero dependencies
       run: sudo apt -y install build-essential cmake libboost-all-dev miniupnpc libunbound-dev graphviz doxygen libunwind8-dev pkg-config libssl-dev libzmq3-dev libsodium-dev libhidapi-dev libnorm-dev libusb-1.0-0-dev libpgm-dev libprotobuf-dev protobuf-compiler
-    - name: install requests
-      run: pip install requests
+    - name: install Python dependencies
+      run: pip install requests psutil monotonic
     - name: tests
       env:
         CTEST_OUTPUT_ON_FAILURE: ON
--- a/tests/functional_tests/CMakeLists.txt
+++ b/tests/functional_tests/CMakeLists.txt
@@ -64,6 +64,7 @@ target_link_libraries(make_test_signature
     ${CMAKE_THREAD_LIBS_INIT}
     ${EXTRA_LIBRARIES})
 
+monero_add_minimal_executable(cpu_power_test cpu_power_test.cpp)
 find_program(PYTHON3_FOUND python3 REQUIRED)
 
 execute_process(COMMAND ${PYTHON3_FOUND} "-c" "import requests; import psutil; import monotonic; print('OK')" OUTPUT_VARIABLE REQUESTS_OUTPUT OUTPUT_STRIP_TRAILING_WHITESPACE)
--- a/tests/functional_tests/mining.py
+++ b/tests/functional_tests/mining.py
@@ -31,6 +31,10 @@
 from __future__ import print_function
 import time
 import os
+import math
+import monotonic
+import util_resources
+import multiprocessing
 
 """Test daemon mining RPC calls
 
@@ -71,12 +75,20 @@ class MiningTest():
     def mine(self, via_daemon):
         print("Test mining via " + ("daemon" if via_daemon else "wallet"))
 
+        cores_init = multiprocessing.cpu_count() # RX init uses all cores
+        cores_mine = 1                           # Mining uses a parametric number of cores
+        time_pi_single_cpu = self.measure_cpu_power_get_time(cores_mine)
+        time_pi_all_cores = self.measure_cpu_power_get_time(cores_init)
+        # This is the last measurement, since it takes very little time and can be placed timewise-closer to the mining itself.
+        available_ram = self.get_available_ram() # So far no ideas how to use this var, other than printing it
+
+        start = monotonic.monotonic()
         daemon = Daemon()
         wallet = Wallet()
 
         # check info/height/balance before generating blocks
         res_info = daemon.get_info()
-        prev_height = res_info.height
+        initial_height = res_info.height
         res_getbalance = wallet.get_balance()
         prev_balance = res_getbalance.balance
 
@@ -85,29 +97,71 @@ class MiningTest():
         if via_daemon:
             res = daemon.start_mining('42ey1afDFnn4886T7196doS9GPMzexD9gXpsZJDwVjeRVdFCSoHnv7KPbBeGpzJBzHRCAs9UxqeoyFQMYbqSWYTfJJQAWDm', threads_count = 1)
         else:
-            res = wallet.start_mining(threads_count = 1)
+            res = wallet.start_mining(threads_count = cores_mine)
 
         res_status = daemon.mining_status()
         assert res_status.active == True
-        assert res_status.threads_count == 1
+        assert res_status.threads_count == cores_mine
         assert res_status.address == '42ey1afDFnn4886T7196doS9GPMzexD9gXpsZJDwVjeRVdFCSoHnv7KPbBeGpzJBzHRCAs9UxqeoyFQMYbqSWYTfJJQAWDm'
         assert res_status.is_background_mining_enabled == False
         assert res_status.block_reward >= 600000000000
 
         # wait till we mined a few of them
-        target_height = prev_height + 5
-        height = prev_height
-        timeout = 240 # randomx is slow to init
+        target_height = initial_height + 5
+        height = initial_height
+        
+        """
+        Randomx init has high variance on CI machines due to noisy neighbors, 
+        taking up resources in parallel (including by our own jobs).
+        
+        Mining is organized in the following scheme:
+            1) first loop's pass: RandomX init and mining
+            2) every next pass:   only mining
+        Pass 1) takes much more time than pass 2)
+        Pass 1) uses all cores, pass 2) just one (currently)
+        For the above reasons both passes need separate timeouts and adjustments.
+        After the first pass, the timeout is being reset to a lower value.
+        """
+        
+        def calc_timeout(seconds_constant, time_pi, cores):
+            """
+            The time it took to calculate pi under certain conditions 
+            is proportional to the time it will take to calculate the real job.
+            
+            The number of cores used decreases the time almost linearly.
+            """
+            timeout = float(seconds_constant) * time_pi / float(cores)
+            return timeout
+
+        timeout_base_init = 60 # RX init needs more time
+        timeout_base_mine = 20
+        timeout_init = calc_timeout(timeout_base_init, time_pi_all_cores,  cores_init)
+        timeout_mine = calc_timeout(timeout_base_mine, time_pi_single_cpu, cores_mine)
+        
+        msg = "Timeout for {} adjusted for the currently available CPU power, is {:.1f} s"
+        print(msg.format("init,  ", timeout_init))
+        print(msg.format("mining,", timeout_mine))
+        
+        timeout = timeout_init
+        rx_inited = False  # Gets initialized in the first pass of the below loop
         while height < target_height:
             seen_height = height
-            for _ in range(timeout):
+            for _ in range(int(math.ceil(timeout))):
                 time.sleep(1)
+                seconds_passed = monotonic.monotonic() - start
                 height = daemon.get_info().height
                 if height > seen_height:
                     break
             else:
-                assert False, 'Failed to mine successor to block %d (initial block = %d)' % (seen_height, prev_height)
-            timeout = 10
+                assert False, 'Failed to mine successor to block %d (initial block = %d) after %d s. RX initialized = %r' % (seen_height, initial_height, round(seconds_passed), rx_inited)
+            if not rx_inited:
+                rx_inited = True
+                timeout = timeout_mine # Resetting the timeout after first mined block and RX init
+                self.print_time_taken(start, "RX init + mining 1st block")
+            else:
+                self.print_time_taken(start, "mining iteration")
+
+        self.print_time_taken(start, "mining total")
 
         if via_daemon:
             res = daemon.stop_mining()
@@ -123,7 +177,7 @@ class MiningTest():
         wallet.refresh()
         res_getbalance = wallet.get_balance()
         balance = res_getbalance.balance
-        assert balance >= prev_balance + (new_height - prev_height) * 600000000000
+        assert balance >= prev_balance + (new_height - initial_height) * 600000000000
 
         if via_daemon:
             res = daemon.start_mining('42ey1afDFnn4886T7196doS9GPMzexD9gXpsZJDwVjeRVdFCSoHnv7KPbBeGpzJBzHRCAs9UxqeoyFQMYbqSWYTfJJQAWDm', threads_count = 1, do_background_mining = True)
@@ -143,6 +197,22 @@ class MiningTest():
             res = wallet.stop_mining()
         res_status = daemon.mining_status()
         assert res_status.active == False
+        
+    def measure_cpu_power_get_time(self, cores):
+        print("Measuring the currently available CPU power...")
+        time_pi = util_resources.get_time_pi_seconds(cores)
+        print("Time taken to calculate Pi on {} core(s) was {:.2f} s.".format(cores, time_pi))
+        return time_pi
+    
+    def get_available_ram(self):
+        available_ram = util_resources.available_ram_gb()
+        threshold_ram = 3
+        print("Available RAM =", round(available_ram, 1), "GB")
+        if available_ram < threshold_ram:
+            print("Warning! Available RAM =", round(available_ram, 1), 
+                  "GB is less than the reasonable threshold =", threshold_ram,
+                  ". The RX init might exceed the calculated timeout.")
+        return available_ram
 
     def submitblock(self):
         print("Test submitblock")
@@ -170,6 +240,10 @@ class MiningTest():
             res = daemon.get_height()
             assert res.height == height + i + 1
             assert res.hash == block_hash
+            
+    def print_time_taken(self, start, msg_context):
+        seconds_passed = monotonic.monotonic() - start
+        print("Time taken for", msg_context, "=", round(seconds_passed, 1), "s.")
 
     def test_randomx(self):
         print("Test RandomX")
--- a/tests/functional_tests/util_resources.py
+++ b/tests/functional_tests/util_resources.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2021 The Monero Project
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are
+# permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of
+#    conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this list
+#    of conditions and the following disclaimer in the documentation and/or other
+#    materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may be
+#    used to endorse or promote products derived from this software without specific
+#    prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+    Help determine how much CPU power is available at the given time
+    by running numerical calculations
+"""
+
+from __future__ import print_function
+import subprocess
+import psutil
+
+def available_ram_gb():
+    ram_bytes = psutil.virtual_memory().available
+    kilo = 1024.0
+    ram_gb = ram_bytes / kilo**3
+    return ram_gb
+
+def get_time_pi_seconds(cores):
+    app_path = './cpu_power_test'
+    time_calc = subprocess.check_output([app_path, str(cores)])
+    decoded = time_calc.decode('utf-8')
+    miliseconds = int(decoded)
+
+    return miliseconds / 1000.0
--- a/tests/README.md
+++ b/tests/README.md
@@ -52,6 +52,11 @@ To run the same tests on a release build, replace `debug` with `release`.
 [TODO]
 Functional tests are located under the `tests/functional` directory.
 
+Building all the tests requires installing the following dependencies:
+```bash
+pip install requests psutil monotonic
+```
+
 First, run a regtest daemon in the offline mode and with a fixed difficulty:
 ```bash
 monerod --regtest --offline --fixed-difficulty 1
--- a/tests/functional_tests/cpu_power_test.cpp
+++ b/tests/functional_tests/cpu_power_test.cpp
@@ -1,0 +1,112 @@
+// Copyright (c) 2014-2021, The Monero Project
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+// 
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+// 
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+// 
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+const char * descr = 
+"This program prints the time (ms) needed to calculate a mathematical challenge. It's purpose \n\
+is to be able to use the result to compare the CPU power available on the machine it's being executed \n\
+and under the given circumstances, which can differ for the same machine. For example: \n\
+The printed value will be different when using 2 of 2 cores, when there an another process \n\
+running on one of the cores core, and when no other intense process running. \n\
+ \n\
+The program expects a one argument: a numerical value of the threads to start the calculations on. \n\
+ \n\
+Prints: \n\
+Time to calculate a mathematical challenge in MILLISECONDS. \n\
+ \n\
+Returns: \n\
+0 on success, \n\
+1 on a missing argument, \n\
+2 on an incorrect format of the argument.\n";
+
+#include <iostream>
+#include <future>
+#include <vector>
+#include <sstream>
+#include <chrono>
+
+using namespace std;
+
+/**
+Uses Leibniz Formula for Pi.
+https://en.wikipedia.org/wiki/Leibniz_formula_for_%CF%80
+*/
+static double calcPi(const size_t max_iter)
+{
+    const double n = 4;
+    double pi = 0;
+    double d = 1;
+    for (size_t i = 1; i < max_iter; ++i)
+    {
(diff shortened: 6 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -131,7 +131,7 @@ jobs:
     - name: install monero dependencies
       run: sudo apt -y install build-essential cmake libboost-all-dev miniupnpc libunbound-dev graphviz doxygen libunwind8-dev pkg-config libssl-dev libzmq3-dev libsodium-dev libhidapi-dev libnorm-dev libusb-1.0-0-dev libpgm-dev libprotobuf-dev protobuf-compiler
     - name: install requests
-      run: pip install requests
+      run: pip install requests psutil
```
mj-xmr, line 134: for RAM readouts

tests/functional_tests/util_resources.py
```diff
@@ -0,0 +1,117 @@
...
+
+def pi_digits_lib(digits):
+    """
+    Returns the number of digits of Pi
+    """
+    return float('{:.{}f}'.format(pi, digits))
+
+def pi_digits_numeric_yield(digits):
+    """Generate x digits of Pi.
+    https://stackoverflow.com/a/54967370
```
iamamyth, line 56: Using code from stackoverflow is a serious licensing issue. Also, as a practical matter, you can perhaps do something much simpler here, such as a calculation with the decimal or math libraries.

tests/functional_tests/mining.py
```diff
@@ -97,17 +105,55 @@ def mine(self, via_daemon):
...
+        print(str(msg).format("init,  "), timeout_init, "s")
+        print(str(msg).format("mining,"), timeout_mine, "s")
+        
+        timeout = timeout_init
+        init_rx = False
         while height < target_height:
             seen_height = height
             for _ in range(timeout):
                 time.sleep(1)
+                delta = datetime.datetime.now() - start
```
iamamyth, line 144: This is not a correct methodology (depends on system clock and timezone), use time.monotonic.

tests/functional_tests/mining.py
```diff
@@ -97,17 +105,55 @@ def mine(self, via_daemon):
...
+        def get_timeout_int_float(time_pi, seconds_base, cores):
+            timeout_float = seconds_base * time_pi / cores
+            timeout_int = int(round(timeout_float + 1)) # +1 could be useful for smaller measurments
+            return timeout_int, timeout_float
+
+        timeout_base_init = 60 # RX init needs more time
+        timeout_base_mine = 20
+        cores_init = multiprocessing.cpu_count() # RX init uses all cores
+        cores_mine = res_status.threads_count    # Mining uses parametric number of cores
+        timeout_init, timeout_float_init = get_timeout_int_float(time_pi_all_cores,  timeout_base_init, cores_init)
```
iamamyth, line 131: Why would anyone need the value as an int and a float? All the relevant python timing functions use floats.

mj-xmr: I need `int` for printing it out to the user and for it to be a valid argument for the `for` loop a few lines below.
Float is obviously used for calculations, in order not to loose on the precision.

tests/functional_tests/util_resources.py
mj-xmr: The proper copyrights actually belong to these gents:

This is implementing a spigot algorithm by A. Sale, D. Saada, S. Rabinowitz, mentioned on http://mail.python.org/pipermail/edu-sig/2012-December/010721.html

https://gist.github.com/deeplook/4947835

tests/functional_tests/mining.py
iamamyth: The loop issue can be trivially solved, just use `math.ceil` (i.e. `range(math.ceil(timeout))`), or switch to a `while time.monotonic() < timeout_time` format. As for printing, I don't know what you mean, floats can be printed and formatted arbitrarily, and the precise value seems more useful than a rounded one.

tests/functional_tests/util_resources.py
iamamyth: Of course algorithms to compute pi are plentiful in the public domain. The issue here is you've now attributed work to stackoverflow and taking that attribution back would be doubly bad. Plus, using an optimized C-based mathematical function, as found in decimal or math, seems maybe a better strategy.

tests/functional_tests/util_resources.py
```diff
@@ -0,0 +1,117 @@
...
+    This helps in estimating how well the problem scales on multiple cores.
+    """
+    # Inputs are the number of digits to calculate for each CPU
+    inp = [num_digits] * cores
+    # Start measuring the time
+    start = datetime.datetime.now()
+    # Perform the calculations on multiple cores in parallel to stress the CPU
+    #with multiprocessing.Pool(cores) as pool: # Python 3.3+
+    #    pool.map(pi_digits_numeric, inp)
+    pool = multiprocessing.Pool(cores)  # Python < 3.3
```
iamamyth, line 102: Should probably use a try/finally here (terminate in the finally).

tests/functional_tests/util_resources.py
```diff
@@ -0,0 +1,117 @@
...
+        
+def get_time_pi(num_digits, cores=multiprocessing.cpu_count()):
+    """   
+    Get the time of generation of pi, using the chosen number of cores.
+    This helps in estimating how well the problem scales on multiple cores.
+    """
+    # Inputs are the number of digits to calculate for each CPU
+    inp = [num_digits] * cores
+    # Start measuring the time
+    start = datetime.datetime.now()
```
iamamyth, line 98: The methodology here ends up counting process startup time, which could lead to wildly inaccurate results. A simple workaround would be to have each task time itself and push results onto a Queue.

tests/functional_tests/util_resources.py
mj-xmr: I think that as long as the PR isn't merged, I don't give nor take anything. That's what a PR is for. Otherwise you create a destructive mental pressure on contributors, making them refrain from doing any contributions.
Besides, a link is not a copyright clause.

tests/functional_tests/mining.py
```diff
@@ -170,6 +232,11 @@ def submitblock(self):
             res = daemon.get_height()
             assert res.height == height + i + 1
             assert res.hash == block_hash
+            
+    def printTimeTaken(self, start, msgContext):
```
iamamyth, line 236: Why is this using camelcase when everything else follows the standard python convention of lowercase, underscore separated identifiers?

tests/functional_tests/util_resources.py
```diff
@@ -0,0 +1,117 @@
...
+    val = float(valStr)
+    rounded = round(val, digits)
+    
+    # Compare if the calcualted Pi is the same as the library one
+    pi_library = pi_digits_lib(digits)
+    if pi_library != rounded:
+        raise RuntimeError("Calculated Pi = ", rounded, "Library Pi = ", pi_library)
+    return rounded
+        
+def get_time_pi(num_digits, cores=multiprocessing.cpu_count()):
```
iamamyth, line 90: Generally calling a function which inspects the CPU count at the module level is poor form, as it requires a fairly complex operation as a side-effect of module import; plus, the default seems unnecessary, only `__main__` uses it, so why not just call `multiprocessing.cpu_count` in main and make the argument non-defaulted?

tests/functional_tests/mining.py
```diff
@@ -97,17 +105,55 @@ def mine(self, via_daemon):
...
+            return timeout_int, timeout_float
+
+        timeout_base_init = 60 # RX init needs more time
+        timeout_base_mine = 20
+        cores_init = multiprocessing.cpu_count() # RX init uses all cores
+        cores_mine = res_status.threads_count    # Mining uses parametric number of cores
+        timeout_init, timeout_float_init = get_timeout_int_float(time_pi_all_cores,  timeout_base_init, cores_init)
+        timeout_mine, timeout_float_mine = get_timeout_int_float(time_pi_single_cpu, timeout_base_mine, cores_mine)
+        
+        msg = "Timeout for {} adjusted for the currently available CPU power, is"
```
iamamyth, line 134: Why not use a format operation to inject both the operation name and the numeric value here? Using format to add a fixed value seems quite odd and unnecessary.

tests/functional_tests/mining.py
```diff
@@ -97,17 +105,55 @@ def mine(self, via_daemon):
...
+        print(str(msg).format("init,  "), timeout_init, "s")
```
iamamyth, line 135: Calling `str` here is pointless, of course the value is a string, it's literally declared as such one line above.

tests/functional_tests/util_resources.py
iamamyth: I'm not saying that you've broken the project by introducing a PR with an SO link, I'm saying it could be problematic to merge said code, given its history.

tests/functional_tests/mining.py
mj-xmr: Overlook

tests/functional_tests/util_resources.py
mj-xmr: Makes sense.

tests/functional_tests/mining.py
mj-xmr: It creates a temporary object, so that I can reuse the msg one line below in an unchanged form.

I'm happy to learn a less odd alternative.

iamamyth: Python strings are immutable.

tests/functional_tests/util_resources.py
mj-xmr: Then I'm very happy, that I misunderstood you.

The Decimal solution work equally fine.

tests/functional_tests/util_resources.py
mj-xmr: I'm not sure about this. If I catch the exception, then I should handle it with a backup solution. For this test an exception here would have to mean aborting the test.

tests/functional_tests/util_resources.py
mj-xmr: Yes, it does count the process start up time, but if it increases by a large margin, so will an equivalent functionality of the RX init just in the next 2 seconds after this calculation.

tests/functional_tests/mining.py
mj-xmr: Done.

tests/functional_tests/mining.py
mj-xmr: Done.

tests/functional_tests/util_resources.py
mj-xmr: Done.

tests/functional_tests/mining.py
mj-xmr: Done.

tests/functional_tests/mining.py
mj-xmr: Done.

tests/functional_tests/mining.py
mj-xmr: Done.

Makefile
```diff
@@ -96,7 +96,8 @@ release: cmake-release
 
 release-test:
 	mkdir -p $(builddir)/release
-	cd $(builddir)/release && cmake -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=release $(topdir) && $(MAKE) && $(MAKE) test
+	#cd $(builddir)/release && cmake -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=release $(topdir) && $(MAKE) && $(MAKE) test
+	cd $(builddir)/release && cmake -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=release $(topdir) && $(MAKE) && ctest -V -R functional_tests_rpc # Temporary
```
mj-xmr, line 100: Will be reverted once the review is done.

tests/functional_tests/CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ execute_process(COMMAND ${PYTHON_EXECUTABLE} "-c" "import requests; print('OK')"
 if (REQUESTS_OUTPUT STREQUAL "OK")
   add_test(
     NAME    functional_tests_rpc
-    COMMAND ${PYTHON_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/functional_tests_rpc.py" "${PYTHON_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_BINARY_DIR}" all)
+    COMMAND ${PYTHON_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/functional_tests_rpc.py" "${PYTHON_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_BINARY_DIR}" mining)
```
mj-xmr, line 71: Will be reverted once the review is done.

tests/functional_tests/util_resources.py
mj-xmr: The decimal solution isn't right. The same code runs for 2s on my machine and 2000s on the CI machine. I'll try to avoid Decimal.

mj-xmr: Trying the Leibnitz' formula now.

tests/functional_tests/util_resources.py
iamamyth: try/finally doesn't swallow any exceptions.

tests/functional_tests/util_resources.py
iamamyth: Well, first off, it's not accurate to assume process startup time will be relatively constant over a short timeframe, but second, even if one makes such an assumption, I fail to see how it resolves my concern.

.github/workflows/build.yml
```diff
@@ -131,7 +131,7 @@ jobs:
     - name: install monero dependencies
       run: sudo apt -y install build-essential cmake libboost-all-dev miniupnpc libunbound-dev graphviz doxygen libunwind8-dev pkg-config libssl-dev libzmq3-dev libsodium-dev libhidapi-dev libnorm-dev libusb-1.0-0-dev libpgm-dev libprotobuf-dev protobuf-compiler
     - name: install requests
-      run: pip install requests
+      run: pip install requests psutil monotonic
```
iamamyth, line 134: Why do you need this library? time.monotonic is a python built-in.

tests/functional_tests/util_resources.py
```diff
@@ -0,0 +1,120 @@
...
+        Calculate PI using Leibnitz' formula
+        
+        https://en.wikipedia.org/wiki/Leibniz_formula_for_%CF%80
+        https://proofwiki.org/wiki/Leibniz's_Formula_for_Pi
+    """
+
+    pi = 0.0
+    n = 4.0
+    d = 1.0
+    pi_lib = pi_digits_lib(digits)
```
iamamyth, line 66: Wouldn't it be better to just run a fixed number of iterations? Who even cares whether this converges on pi or not?

tests/functional_tests/util_resources.py
mj-xmr: OK then.

.github/workflows/build.yml
mj-xmr: I also wanted to use time.monotinic(), but the CI crashed. It turned out, that we were using Python2 all the time, hence:
https://github.com/monero-project/monero/pull/7396

Either we leave it like it is (Python2 code), or we quickly merge the https://github.com/monero-project/monero/pull/7396

tests/functional_tests/util_resources.py
mj-xmr: Yes, I was considering this.

A greater concern to me is however, that I get much faster calculation of the same algo under Python3 locally, than in the Python2, so you will see a huge overshoot in the CI's logs. So if we don't update to Python3, then even this algo doesn't seem right. I'm only worried, that other Python scripts won't work after the update. I'll let it process for the night and get back here.

tests/functional_tests/util_resources.py
mj-xmr: You are right from the logical PoV. I would just like to cover the startup up with a calculation long enough, that the process init becomes negligible. I have a small problem with your approach, that it makes a simple thing slightly more complex. I wanted to solve this with averaging the returned processing time of the threads. I only fear, that the time calculation within the threads will be not accurate. Do you think the monotonic timer will help?

.github/workflows/build.yml
iamamyth: I haven't looked at the runner in detail, but these files have a shebang line which claims python3 is to be expected, so it's problematic to be using python2. I also note that the files import `print_function` from `__future__`, which would imply python2 is in use. Overall, I would try to clean things up to use python3, as python2 is EOL and dying rather quickly (the only reason it's still available, and sometimes the default in certain distros, is because core system components have assumed it, and it takes a while to make all the requisite changes, though I think most distros are quite close and it'll be gone in a year).

tests/functional_tests/util_resources.py
iamamyth: I think you're inevitably going to see fluctuations with your current approach, as, if you look at, e.g. microbenchmark data from different python3 point releases, it's all over the map (some get faster, some slower). You'd be much better off running a benchmark closer to the machine (i.e. written in C). I suggest decimal because it's a thin wrapper around the C mpdec library, but, unfortunately, it could also fall back to pure-python if mpdec isn't installed, and, in any case, it's not as good as some vanilla C program.

iamamyth:
I should probably contextualize my comments up to this point, as I think there's a danger they could be misconstrued: Normally system benchmarks would be written in C, so as to minimize overhead and maximize accuracy. However, creating an accurate measure of "system performance" is itself a fraught and difficult exercise, and one that I don't think proves necessary here. Instead, I imagine this PR as introducing a "smoke test" for something having gone very wrong with the mining code (say, a 5x or 10x regression). A more focused, narrowly bounded mining benchmark might be a good idea, but you can't reasonably achieve that result in a noisy neighbor environment. To that end, your bounds in this PR can be quite loose (1.5 or even 2x baseline), so you shouldn't need to worry about python version (if you're seeing > 50% delta based on python version, probably you should use a different test). I still think proper benchmark hygiene (using a sane timing methodology) should be observed, to the extent that it's not terribly difficult and in some ways makes the code both clearer and more useful (a poorly constructed benchmark begs the question, "Was this an oversight, or just intentionally lax in its methodology?" While code comments can answer that question, it's often simpler to avoid asking it in the first place).

tests/functional_tests/util_resources.py
iamamyth: I think running a benchmark with a broken methodology makes everything more complex, as it's very difficult to reason about a calculation with many different noise sources (are they acceptable? did the code intentionally leave them in play, and if so, why? etc). Using a Queue maybe adds 5-10 LoC and much clarity. As for timing process startup, I proposed just not even bothering and assuming some fixed value, so that's not really a concern, I think (you could easily time it, too, but I don't have a strong opinion on the matter).

tests/functional_tests/util_resources.py
mj-xmr: I can write a C++ app. I prefer not to touch  threads in C

tests/functional_tests/util_resources.py
mj-xmr: Not needed anymore. Launching an external C++ app.

tests/functional_tests/util_resources.py
mj-xmr: Done. The C++ app delivers very good results.

Still to do for the app: Copyright and documentation.

mj-xmr:
Still to do for the new C++ timing app: Copyright and documentation.

tests/functional_tests/util_resources.py
```diff
@@ -0,0 +1,60 @@
...
+    decoded = time_calc.decode('utf-8')
+    try:
+        miliseconds = int(decoded)
+    except ValueError as ex:
+        print("Couldn't convert " + decoded)
+        raise ex
+    
+    return miliseconds / 1000.0
+     
+if __name__ == '__main__':
```
iamamyth, line 58: This seems pointless.

tests/functional_tests/util_resources.py
```diff
@@ -0,0 +1,60 @@
...
+    ram_gb = ram_bytes / kilo / kilo / kilo
+    return ram_gb
+
+def get_time_pi_seconds(cores):
+    app_path = './cpu_power_test'
+    time_calc = subprocess.check_output([app_path, str(cores)])
+    decoded = time_calc.decode('utf-8')
+    try:
+        miliseconds = int(decoded)
+    except ValueError as ex:
```
iamamyth, line 52: The try/exception seems unproductive, if conversion fails it'll throw a ValueError which includes the value and rationale for the failure. Plus, utility functions should generally avoid `print`, unless their primary purpose is producing output.

tests/functional_tests/util_resources.py
```diff
@@ -0,0 +1,60 @@
...
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+    Help determine how much CPU power is available at the given time
+    by running numerical calculations
+"""
+
+from __future__ import print_function
```
iamamyth, line 36: Makes no sense in python3.

tests/functional_tests/util_resources.py
```diff
@@ -0,0 +1,60 @@
...
+import subprocess
+import psutil
+
+def available_ram_gb():
+    ram_bytes = psutil.virtual_memory().available
+    kilo = 1024.0
+    ram_gb = ram_bytes / kilo / kilo / kilo
```
iamamyth, line 43: Not super important, but this seems rather verbose an unnecessarily slow (three divisions), when you could just do:
```
return psutil.virtual_memory().available / (2**30)
```

tests/functional_tests/mining.py
```diff
@@ -97,17 +107,59 @@ def mine(self, via_daemon):
...
+        msg = "Timeout for {} adjusted for the currently available CPU power, is {} s"
+        print(msg.format("init,  ", round(timeout_init, 1)))
+        print(msg.format("mining,", round(timeout_mine, 1)))
+        
+        timeout = timeout_init
+        init_rx = False
         while height < target_height:
             seen_height = height
-            for _ in range(timeout):
+            for _ in range(int(math.ceil(timeout))):
```
iamamyth, line 149: `int` is superfluous, `ceil` returns an int.

tests/functional_tests/mining.py
```diff
@@ -143,6 +195,22 @@ def mine(self, via_daemon):
             res = wallet.stop_mining()
         res_status = daemon.mining_status()
         assert res_status.active == False
+        
+    def measure_cpu_power_get_time(self, cores):
+        print("Measuring the currently available CPU power...")
+        time_pi = util_resources.get_time_pi_seconds(cores)
+        print("Time taken to calculate Pi on {} core(s) was {} s.".format(cores, round(time_pi, 2)))
```
iamamyth, line 202: If you're using format here, you may as well use it to print the value to two digits of precision, rather than rounding.

tests/functional_tests/mining.py
```diff
@@ -97,17 +107,59 @@ def mine(self, via_daemon):
...
+            return timeout
+
+        timeout_base_init = 60 # RX init needs more time
+        timeout_base_mine = 20
+        cores_init = multiprocessing.cpu_count() # RX init uses all cores
+        cores_mine = res_status.threads_count    # Mining uses parametric number of cores
+        timeout_init = calc_timeout(timeout_base_init, time_pi_all_cores,  cores_init)
+        timeout_mine = calc_timeout(timeout_base_mine, time_pi_single_cpu, cores_mine)
+        
+        msg = "Timeout for {} adjusted for the currently available CPU power, is {} s"
```
iamamyth, line 141: Similar thought here regarding using number formatting rather than rounding the injected value.

tests/functional_tests/mining.py
```diff
@@ -31,6 +31,10 @@
 from __future__ import print_function
```
iamamyth, line 31: Makes no sense in python3.

tests/functional_tests/mining.py
```diff
@@ -71,6 +75,12 @@ def create(self):
     def mine(self, via_daemon):
         print("Test mining via " + ("daemon" if via_daemon else "wallet"))
 
+        time_pi_single_cpu = self.measure_cpu_power_get_time(1)
+        time_pi_all_cores = self.measure_cpu_power_get_time(multiprocessing.cpu_count())
```
iamamyth, line 79: Would suggest putting the core count in a variable, as you use it repeatedly in calculations.

tests/functional_tests/mining.py
```diff
@@ -97,17 +107,59 @@ def mine(self, via_daemon):
...
                 time.sleep(1)
+                seconds_passed = monotonic.monotonic() - start
                 height = daemon.get_info().height
                 if height > seen_height:
                     break
             else:
-                assert False, 'Failed to mine successor to block %d (initial block = %d)' % (seen_height, prev_height)
-            timeout = 10
+                assert False, 'Failed to mine successor to block %d (initial block = %d) after %d s. RX initialized = %r' % (seen_height, prev_height, round(seconds_passed), init_rx)
+            if not init_rx:
```
iamamyth, line 157: You don't really need this boolean, can just use `seen_height == prev_height`.

tests/functional_tests/mining.py
mj-xmr: Only since Python3

tests/functional_tests/util_resources.py
mj-xmr: I used your previous suggestion of kilo ** 3 for better readability.

tests/functional_tests/util_resources.py
mj-xmr: Until https://github.com/monero-project/monero/pull/7396 is merged, we're stuck with Python2

tests/functional_tests/util_resources.py
mj-xmr: Removed.

tests/functional_tests/util_resources.py
mj-xmr: Removed.

tests/functional_tests/mining.py
mj-xmr: Done.

tests/functional_tests/mining.py
mj-xmr: Done.

tests/functional_tests/mining.py
mj-xmr: I need it for readability. Same as with documentation and self-documenting code: the goal is not only to write the most minimal code, but code that can be undrestood by others.

tests/functional_tests/mining.py
mj-xmr: Done. 

Regarding nproc, this variable name would suggest the source and quantity of the variable, but not the intention how it's meant to be used.

tests/functional_tests/mining.py
iamamyth: I don't find the variable terribly offensive or anything, but it seems you're perhaps papering over the real readability issue, which is, I think, that `prev_height` is a rather poor name, and should be something along the lines or `initial_height`. With that change, I think the conditional proves quite clear without an extra variable, but I won't belabor the point.

tests/functional_tests/util_resources.py
iamamyth: I see your point, but I think, perhaps, you've ended up making your life more difficult by splitting commits in this way: Though each commit and PR should achieve a single logical goal, there are cases where a clear dependency exists, and ignoring the dependency therefore introduces extra work and complexity, the very reasons to ordinarily avoid PRs with multiple aims. For example, this PR adds multiple python2-specific pieces of code, which will eventually have to be removed. On top of that, the files all still mark themselves as targeting python3 via the shebang line, which makes life very confusing for the reader. You could have, instead, either branched this PR from the "use python3" PR, or introduced a more limited change which just runs the mining tests with python3, as a single commit which forms part of this multi-commit PR. I don't think it's an issue worthy of further discussion, but certainly something to consider in future work.

tests/functional_tests/mining.py
iamamyth: All I will say here is that if you find `nproc` insufficient to document intent, you can always alias it while still avoiding repeatedly querying the machine's core count.

mj-xmr: The suggestion was correct, and I query the cpu count only once now (well twice, since the method is being called twice)

tests/functional_tests/mining.py
mj-xmr: The `initial_height` is indeed a much better name for the variable, especially since it's a constant.

Still the `rx_inited` var is used twice - once in the print function and secondly below the `for` loop. Using the comparison twice is error-prone IMHO.

BTW. I added an additional printout, which will be executed if one `while` loop's iteration is not enough.

mj-xmr:
It sounds to me, like we're nearing the end of the review. I'm rebasing.

iamamyth (approved):
Please make sure to open a PR which removes the python2 compat code (montonic module, `__future__` imports, etc) once the python3 test runner PR gets merged.

mj-xmr:
Thanks for the review.

> Please make sure to open a PR which removes the python2 compat code (montonic module, `__future__` imports, etc) once the python3 test runner PR gets merged.

For the record, we're talking about https://github.com/monero-project/monero/pull/7396 , which needs to be approved and merged first.

mj-xmr:
A successful ad-hoc run [under MacOSX](https://github.com/mj-xmr/monero/runs/2107703942?check_suite_focus=true).

mj-xmr:
Just added the Python dependency list to the documentation.

mj-xmr:
As requested by @selsta , I [ran the mining test multiple times](https://github.com/mj-xmr/monero/actions?query=event%3Aschedule) (19) and each of the run was successful. This should serve as the ultimate proof, that the concept works fine.
Within the logs of the tests, you're looking for the string:
`Test mining via daemon`

mj-xmr:
Rebased and integrated latest test fixes.

selsta (approved):

luigi1111:
@mj-xmr I went ahead and merged #7396

mj-xmr:
> @mj-xmr I went ahead and merged #7396

Got it. Will resolve very soon.

PR #7390, opened 2021-02-19
async_protocol_handler_config: remove connection correctly

perfect-daemon:


moneromooo-monero (commented):
Why is this change necessary ? del_connection does the same, with an extra lock (but the lock is recursive AFAIK) and notifies on_connection_close.

selsta:
My daemon deadlocked, this patch was a result of the backtrace.

selsta:
but @perfect-daemon has to explain why this change fixes it

moneromooo-monero (approved):

PR #7396, opened 2021-02-22
[CI] Use Python3 for funcional tests, instead of Python2

mj-xmr:
It turned out, that the test part of CI was using Python2 all the time. Upgrading to Python3 brings compatibility, as well as runtime speed benefits.

### DIFF
--- a/tests/functional_tests/CMakeLists.txt
+++ b/tests/functional_tests/CMakeLists.txt
@@ -64,15 +64,17 @@ target_link_libraries(make_test_signature
     ${CMAKE_THREAD_LIBS_INIT}
     ${EXTRA_LIBRARIES})
 
-execute_process(COMMAND ${PYTHON_EXECUTABLE} "-c" "import requests; print('OK')" OUTPUT_VARIABLE REQUESTS_OUTPUT OUTPUT_STRIP_TRAILING_WHITESPACE)
+find_program(PYTHON3_FOUND python3 REQUIRED)
+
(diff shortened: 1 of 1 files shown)


iamamyth:
Don't most of the test files have a shebang line which explicitly specifies `python3`? Why not just be explicit in all the invocations, rather than updating alternatives, which could have some unfortunate and perhaps unwanted side effects?

mj-xmr:
Normally I'm open for alternatives, but this won't work here. In case of mining, python2 is used, even though pyyhon3 is explicitly selected. I think it is because the "caller", the 'functional_tests_rpc', is being called with the default python, which CMake chooses to be Python2. Changing this call would make all these tests run Python3 anyway. What it would change, is that nothing else than 'functional_tests_rpc' would change. But this is why I ran all the tests, and they haven't reported any regressions.

iamamyth:
I'm saying you should change the runner to use python3. How can that approach possibly not work?

mj-xmr:
> I'm saying you should change the runner to use python3. How can that approach possibly not work?

This will work.

selsta:
After this patch functional tests will only run with python3? Currently it supports 2 and 3? Did I understand that correctly?

mj-xmr:
> After this patch functional tests will only run with python3? Currently it supports 2 and 3? Did I understand that correctly?

It will still work under 2 and 3. But the goal is to modify the tests, after applying this patch, to use only Python3 syntax and libraries. 

It's a seperate decision, but what we have to keep in mind, thay Python2 is not supported anymore.
https://www.python.org/doc/sunset-python-2/

Sunsetting Python 2

We are volunteers who make and take care of the Python programming language. We have decided that January 1, 2020, was the day that we sunset Python 2. That means that we will not improve it anymore after that day, even if someone finds a security problem in it. You should upgrade to Python 3 as soon as you can.

moneromooo-monero (approved):

PR #7402, opened 2021-02-25
Epee: add headers to project files (via glob)

mj-xmr:
This change allows the epee headers to be searched for easily via IDEs. Additionally, the IDEs will automatically save the changed headers, when compile command is issued.

### DIFF
--- a/contrib/epee/src/CMakeLists.txt
+++ b/contrib/epee/src/CMakeLists.txt
@@ -26,11 +26,20 @@
 # STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 # THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+set(EPEE_INCLUDE_DIR_BASE "${CMAKE_CURRENT_SOURCE_DIR}/../include")
+
+# Adding headers to the file list, to be able to search for them in IDEs.
+file(GLOB EPEE_HEADERS_PUBLIC 
+	"${EPEE_INCLUDE_DIR_BASE}/*.h*"    # h* will include hpps as well.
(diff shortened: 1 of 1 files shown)


contrib/epee/src/CMakeLists.txt
```diff
@@ -26,11 +26,19 @@
 # STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 # THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+set(EPEE_INCLUDE_DIR_BASE "${CMAKE_CURRENT_SOURCE_DIR}/../include")
+
+file(GLOB EPEE_HEADERS_PUBLIC 
+	"${EPEE_INCLUDE_DIR_BASE}/*.h*"
+	"${EPEE_INCLUDE_DIR_BASE}/**/*.h*"
```
mj-xmr, line 33: h* to be able to find the hpp files as well, if such are added.

moneromooo-monero, line 33: Is ** a special case for "any number of subdirectories" ?

mj-xmr: Yes. [Here's](https://stackoverflow.com/a/62985520) some discussion.

Shall I simply explain this in a comment?

moneromooo-monero: Up to you, I think I've seen this kind of syntax elsewhere before, it seems fairly self explanatory there even if you don't know what it does, you can guess.

moneromooo-monero (approved):

moneromooo-monero (approved):

contrib/epee/src/CMakeLists.txt
ndorf: Why not `{h,hpp}` instead of `h*`?

mj-xmr: I've just checked it. It doesn't work.

PR #7404, opened 2021-02-26
updates: verify DNS hash against GItian reproducible build hashes

xiphon:
Implemented additional new version hash verification against published GItian reproducible build hashes (https://repo.getmonero.org/monero-project/gitian.sigs Gitlab repo).

By default it requires at least 3 build hashes published by Gitian contributors (builders). If there are more available, it will check all of them.  Both parameters are configurable.

Errors out if any Gitian assert doesn't contain expected new version hash for specific platform.

CLI option `--no-verify-update-hash` disables the logic.

Applies to:
1. Periodic update check (`--check-updates [disabled|notify|download|update]` CLI option)
2.  `/update` daemon RPC
3. `update [check|download]` daemon console command 

PS: Doesn't verify Gitian OpenPGP signatures yet. Will implement this in a separate PR.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -43,6 +43,7 @@ set(common_sources
   pruning.cpp
   spawn.cpp
   threadpool.cpp
+  update_hash_checker.cpp
   updates.cpp
   aligned.c
   timings.cc
@@ -83,6 +84,7 @@ set(common_private_headers
   spawn.h
   stack_trace.h
   threadpool.h
+  update_hash_checker.h
   updates.h
   aligned.h
   timings.h
--- a/src/common/update_hash_checker.cpp
+++ b/src/common/update_hash_checker.cpp
@@ -1,0 +1,384 @@
+// Copyright (c) 2021, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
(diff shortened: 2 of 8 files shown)


moneromooo-monero:
Connecting to github from monerod feels... icky.

selsta:
Would it be better to mirror it on getmonero?

moneromooo-monero:
IMHO yes. Whether it's better than now, not sure though. monerod connecting to the mothership every 12 hours feels wrong too. Though less than connecting to MS.

moneromooo-monero:
Actually, if it does so just when it sees a new version, it'd download anyway (if enabled) so it seems OK.

xiphon:
No problem. Will implement Gitlab API support to fetch GItian hashes from https://repo.getmonero.org/monero-project/gitian.sigs. Though we have to mirror/move the repo first (it's empty at the moment).

xiphon:
> Will implement Gitlab API support to fetch GItian hashes from https://repo.getmonero.org/monero-project/gitian.sigs.

Done.

xiphon:
Implemented unit tests, 100% test coverage.

selsta:
```
D:/a/monero/monero/src/common/update_hash_checker.cpp:298:35: error: 'get_xtype_from_string' is not a member of 'epee::string_tools'; did you mean 'get_ip_int32_from_string'?
1023
```

xiphon:
> ```
> D:/a/monero/monero/src/common/update_hash_checker.cpp:298:35: error: 'get_xtype_from_string' is not a member of 'epee::string_tools'; did you mean 'get_ip_int32_from_string'?
> ```

Yeah, it was moved recently. Updated.

selsta:
This appears obsolete because we don't use gitian anymore. A new PR based on this version might make sense in the future.

Issue #7408, opened 2021-02-28
[feature] Add gRPC wallet interface


vdero133:
Opening this issue to discuss and track a potential addition of a gRPC wallet interface.

Some discussion has already happened on [on reddit](https://old.reddit.com/r/Monero/comments/l6fxkl/gauging_interest_grpc_api/) and IRC/matrix.

The basis for this would be to unify the functionality of unary RPC calls and pub-sub ZMQ messages into a single interface. The interface proposed is minimal, as the scope of it would be to ease integration but not replace the JSON-RPC or ZMQ APIs.

I believe gRPC would achieve this by: 

- Provide a single dependency and framework to work with for synchronous and asynchronous calls.
- Enable more granular payment notifications.
- Possibility to combine documentation to the interface definition.
- Extended wallet client library support to all languages supported by gRPC.
- Autogenerated documentation.

In terms of risks:

- New dependency on the gRPC plugin for the protoc compiler (protobuf already being a dependency for trezor support).
- Reliance on Google maintaining the library/plugin.

Potential questions that may arise:

- Authentication can be done via [TLS](https://grpc.io/docs/guides/auth/)

Submitted  #7409 for initial feedback before going further.

vtnerd:
This is another serialization format _and_ rpc layer. I think the negatives outweigh the benefits.

* The scheme is less supported in other systems than HTTP-JSON-RPC
* `protobuf` is one of the most irritating formats - its not self-describing and the output of objects/structs (basically) has to be buffered to write the variable-length size field before the object. Whereas the ZeroMQ code already "streams" outputs to JSON, and the HTTP (epee) code could do the same with some (slightly painful) refactoring. This has a big impact on performance because many binary values have to be sent for wallet processing.
* `protobuf` injects its auto-generated structures into codebases, so now there's a third copy (ZeroMQ and HTTP already could use a good merging). I'm closer to providing a "slice" concept to the epee serialization code so that even when copies are necessary its just a ref count change. The `protobuf` code enforces `std::string` in both directions in the auto-generated code. FWIW ZeroMQ code, which uses rapidjson for reading in, probably has this limitation with its DOM (assuming it can't be templated to something else).


> Provide a single dependency and framework to work with for synchronous and asynchronous calls.

You'd have to expand on this. ZeroMQ can do all of the stuff gRPC is doing at the protocol layer (afaik - see ROUTER/DEAER sockets), but gRPC might have built-in mechanisms for async callbacks in the language of choice.


[...]

PR #7409, opened 2021-02-28
add proto and rpc definitions

vdero133:


selsta:
https://github.com/monero-project/monero/issues/7408#issuecomment-814344865

PR #7413, opened 2021-03-01
cmake: fix ccache detection

perfect-daemon:


cmake/FindCcache.cmake
```diff
@@ -42,12 +42,19 @@
...
+	set(TEST_PROJECT "${CMAKE_BINARY_DIR}/${CMAKE_FILES_DIRECTORY}/CMakeTmp")
+	file(WRITE "${TEST_PROJECT}/CMakeLists.txt" [=[
+cmake_minimum_required(VERSION 3.1)
+project(test)
+option (CCACHE "")
+file(WRITE "${CMAKE_SOURCE_DIR}/test.cpp" "int main() { return 0; }")
+set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE}")
+set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK    "${CCACHE}")
+add_executable(test test.cpp)
+]=])
```
mj-xmr, line 54: Please tabulate this block twice.

mj-xmr (changes requested):

mj-xmr:
I have checked the logs. It appears to work properly. Everything works like before (of course FreeBSD's ccache is still failing, like before)

cmake/FindCcache.cmake
perfect-daemon:
```
Bracket argument can't be formatted due to limitations[0] of CMake language without affecting result.
[0] https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#bracket-argument
```

selsta (approved):
Fixes ccache detection on my Mac.

cmake/FindCcache.cmake
mj-xmr: OK then.

mj-xmr (approved):

selsta (approved):

PR #7416, opened 2021-03-02
Epee: dynamic linkage; move monero_add_library to root CMakeLists.txt

mj-xmr:
For a dynamic library build the change saves 9.5% of disk space and some linkage time, especially if the epee's implementation is to be changed. More savings may come, as more epee's implementation moves from headers to cpps. Similarly when you try to do Test Driven Development, you'll have to waste cumulatively a very long time on linkage.

In order to achieve this change, I had to move the `monero_add_library` from `src/CMakeLists.txt` to the root `CMakeLists.txt`

image

### DIFF
--- a/contrib/epee/src/CMakeLists.txt
+++ b/contrib/epee/src/CMakeLists.txt
@@ -34,7 +34,7 @@ file(GLOB EPEE_HEADERS_PUBLIC
 	"${EPEE_INCLUDE_DIR_BASE}/**/*.h*" # Any number of subdirs will be included.
 )
 
-add_library(epee STATIC byte_slice.cpp byte_stream.cpp hex.cpp abstract_http_client.cpp http_auth.cpp mlog.cpp net_helper.cpp net_utils_base.cpp string_tools.cpp
+monero_add_library(epee byte_slice.cpp byte_stream.cpp hex.cpp abstract_http_client.cpp http_auth.cpp mlog.cpp net_helper.cpp net_utils_base.cpp string_tools.cpp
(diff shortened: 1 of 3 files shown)


mj-xmr:
Resolved conflicts.

contrib/epee/src/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ file(GLOB EPEE_HEADERS_PUBLIC
 	"${EPEE_INCLUDE_DIR_BASE}/**/*.h*" # Any number of subdirs will be included.
 )
 
-add_library(epee STATIC byte_slice.cpp byte_stream.cpp hex.cpp abstract_http_client.cpp http_auth.cpp mlog.cpp net_helper.cpp net_utils_base.cpp string_tools.cpp
+monero_add_library(epee byte_slice.cpp byte_stream.cpp hex.cpp abstract_http_client.cpp http_auth.cpp mlog.cpp net_helper.cpp net_utils_base.cpp string_tools.cpp
```
vtnerd, line 37: The major change was the removal of the `static` keyword ... ? With that change, a cmake configuration request for dynamic libraries will build `epee` as a dynamic library regardless. So the moving of the macro is more for consistency?

vtnerd (approved):

contrib/epee/src/CMakeLists.txt
mj-xmr: Moving the macro to a broader scope allowed to use it here, in the Epee.

The macro doesn't expect the linkage type, as this is being configured by the user. It may as well be static, if the user chooses so. The common macro allows for other things as well.

PR #7417, opened 2021-03-02
documentation: Make pruning section more consistent in README.md

Luxter77:
use syntax highlighting on commands and command line arguments

### DIFF
--- a/README.md
+++ b/README.md
@@ -712,6 +712,8 @@ to prune an existing blockchain using the `monero-blockchain-prune` tool or usin
 with an existing chain. If an existing chain exists, pruning will temporarily require disk space to store both the full
 and pruned blockchains.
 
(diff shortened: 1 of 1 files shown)


selsta (approved):

erciccione (commented):
Since you are already touching this section, could you add a link to the 'Pruning' entry in the Moneropedia? There are more detailed info: https://www.getmonero.org/resources/moneropedia/pruning.html.

Also, could you update `the full Monero blockchain file is about 80 GB. One can store a pruned blockchain, which is about 28 GB.` to `the full Monero blockchain file is about 100 GB. One can store a pruned blockchain, which is about 30 GB.`?

Luxter77:
Done

selsta (approved):

erciccione (approved):

PR #7422, opened 2021-03-03
wallet rpc: return output keys with incoming transfers

benevanoff:
Mentioned in IRC

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -1888,6 +1888,7 @@ namespace tools
         rpc_transfers.tx_hash      = epee::string_tools::pod_to_hex(td.m_txid);
         rpc_transfers.subaddr_index = {td.m_subaddr_index.major, td.m_subaddr_index.minor};
         rpc_transfers.key_image    = td.m_key_image_known ? epee::string_tools::pod_to_hex(td.m_key_image) : "";
+        rpc_transfers.pubkey       = epee::string_tools::pod_to_hex(td.get_public_key());
         rpc_transfers.block_height = td.m_block_height;
         rpc_transfers.frozen       = td.m_frozen;
         rpc_transfers.unlocked     = m_wallet->is_transfer_unlocked(td);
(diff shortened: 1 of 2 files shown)


ndorf (approved):

src/wallet/wallet_rpc_server.cpp
```diff
@@ -1888,6 +1888,13 @@ namespace tools
         rpc_transfers.tx_hash      = epee::string_tools::pod_to_hex(td.m_txid);
         rpc_transfers.subaddr_index = {td.m_subaddr_index.major, td.m_subaddr_index.minor};
         rpc_transfers.key_image    = td.m_key_image_known ? epee::string_tools::pod_to_hex(td.m_key_image) : "";
+        std::vector<std::string> keys;
+        for (auto o : td.m_tx.vout)
```
mj-xmr, line 1892: Can it be const auto & o ?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -1888,6 +1888,13 @@ namespace tools
         rpc_transfers.tx_hash      = epee::string_tools::pod_to_hex(td.m_txid);
         rpc_transfers.subaddr_index = {td.m_subaddr_index.major, td.m_subaddr_index.minor};
         rpc_transfers.key_image    = td.m_key_image_known ? epee::string_tools::pod_to_hex(td.m_key_image) : "";
+        std::vector<std::string> keys;
+        for (auto o : td.m_tx.vout) 
+        {
+          std::string key = epee::string_tools::pod_to_hex(boost::get<cryptonote::txout_to_key>(o.target).key);
```
mj-xmr, line 1894: This can be const for sure.

mrtestyboy781, line 1894: In the similar wallet-cli command it actually only lists one output per key image, as I read, this is listing all the vouts. I believe we can filter this to only be the single vout matching the key image

jorgeelmundoso: I guess what should be taken care of is to make sure if the get_transfers call is filtered by account.major/minor only the correlating vout/vouts should be returned.

benevanoff:
@mrtestyboy781 I've changed it to only return the owned public key associated with that key image :)

mrtestyboy781:
Tested and lgtm, thank you @benevanoff

moneromooo-monero (approved):

PR #7430, opened 2021-03-05
wallet_rpc_server: set seed language in generate_from_keys

moneromooo-monero:
Also sanity check language name

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -3118,17 +3118,7 @@ namespace tools
     }
     std::string wallet_file = req.filename.empty() ? "" : (m_wallet_dir + "/" + req.filename);
     {
-      std::vector<std::string> languages;
-      crypto::ElectrumWords::get_language_list(languages, false);
-      std::vector<std::string>::iterator it;
-
-      it = std::find(languages.begin(), languages.end(), req.language);
-      if (it == languages.end())
-      {
-        crypto::ElectrumWords::get_language_list(languages, true);
-        it = std::find(languages.begin(), languages.end(), req.language);
-      }
-      if (it == languages.end())
+      if (!crypto::ElectrumWords::is_valid_language(req.language))
       {
         er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
         er.message = "Unknown language: " + req.language;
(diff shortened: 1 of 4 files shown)


src/wallet/wallet_rpc_server.cpp
```diff
@@ -3155,7 +3155,21 @@ namespace tools
       er.message = "Failed to create wallet";
       return false;
     }
-    wal->set_seed_language(req.language);
+    if (!req.language.empty())
```
hyc, line 3158: You're duplicating code from just a few lines up, why? (lines 3122-... here)

hyc: Since language validation appears to be done in at least 3 places, maybe this chunk should be moved into its own function.

moneromooo-monero: Because I missed it as it was done earlier than I looked.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3118,17 +3127,7 @@ namespace tools
...
-      if (it == languages.end())
-      {
-        crypto::ElectrumWords::get_language_list(languages, true);
-        it = std::find(languages.begin(), languages.end(), req.language);
-      }
-      if (it == languages.end())
+      if (!is_valid_language(req.language))
       {
         er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
         er.message = "Unknown language: " + req.language;
```
hyc, line 3133: Just wondering if we should make these error messages more uniform between 2 of these cases. I guess these strings aren't part of the translation project are they?

moneromooo-monero: They're not, they'd have tr().

src/wallet/wallet_rpc_server.cpp
```diff
@@ -104,6 +104,16 @@ namespace
         entry.suggested_confirmations_threshold = std::max(entry.suggested_confirmations_threshold, (unlock_time - now + DIFFICULTY_TARGET_V2 - 1) / DIFFICULTY_TARGET_V2);
     }
   }
+  //------------------------------------------------------------------------------------------------------------------------------
+  bool is_valid_language(const std::string &language)
+  {
+    std::vector<std::string> language_list;
```
iamamyth, line 110: Might make more sense for this to be an ElectrumWords method, that way you can avoid copying the language list to a temp vector.

iamamyth (approved):

Issue #7392, opened 2021-02-21
Building For M1 Macs Natively


jbrot:
HI! I just compiled the repo on the new Mac Mini with M1 processor, and I'd like to report on my experience.

## Compiling

The root `CMakeLists.txt` is incompatible with the ARM Macs, because of [lines 486 to 494 of CMakeLists.txt](https://github.com/monero-project/monero/blob/master/CMakeLists.txt#L486):
```cmake
if (APPLE AND NOT IOS)
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=x86-64 -fvisibility=default -std=c++11")
  if (NOT OPENSSL_ROOT_DIR)
      EXECUTE_PROCESS(COMMAND brew --prefix openssl
        OUTPUT_VARIABLE OPENSSL_ROOT_DIR
        OUTPUT_STRIP_TRAILING_WHITESPACE)
    message(STATUS "Using OpenSSL found at ${OPENSSL_ROOT_DIR}")
  endif()
endif()
``` 
The problem here is that this adds `-march=x86-64` to `CMAKE_CXX_FLAGS`, while the M1 is an ARM processor. I commented out this block and the repo compiled successfully. I suspect just commenting out [line 487](https://github.com/monero-project/monero/blob/master/CMakeLists.txt#L487) would be sufficient, but I don't know if this line is needed under other circumstances. Hopefully someone more familiar with the project will be able to weigh in.

Furthermore, `README.md` suggests ZeroMQ 3.0 should be installed, but this resulted in a cryptic templating error. The error was ultimately caused because the function `zmq_send_const` was not present in my header files. I then installed ZeroMQ 4.3.4 and the repo compiled successfully. This suggests that the information in `README.md` is outdated.

## Testing

I started running the unit tests, ~~but skipped `core_tests` initially in the interest of time~~. I saw two failures:
1. [check_missing_rpc_methods](https://github.com/monero-project/monero/blob/master/tests/functional_tests/CMakeLists.txt#L78) failed. Further investigation revealed this was because I did not have the python package `requests` installed system-wide. I noticed that `functional_tests_rpc` above was only enabled when `requests` was found. My experience suggests that `check_missing_rpc_methods` should probably be gated under this condition, as well. I think `requests` should also be listed as an optional dependency in `README.md`.
[...]

PR #3430, opened 2018-03-18
Add Depends to monero

TheCharlatan:
Depends is a localized cross compilation environment originally developed by Cory Fields. It allows its users to create cross compiled binaries for 32 and 64 bit linux and windows, arm hf, aarch64 and macos.

This pull request places depends into contrib/depends. The required monero dependencies are generated for each platform triplet (for example `x86_64-w64-mingw32`) and saved in their own directory in depends. Recipes for each package can be found in `contrib/depends/packages`, `packages.mk` defines the packages used for each platform triplet. The allowed ${triplet}'s are: `i686-w64-mingw32`, `x86_64-w64-mingw32`, `x86_64-apple-darwin11`, `x86_64-linux-gnu`, `i686-linux-gnu`, `arm-linux-gnueabihf` and `aarch64-linux-gnu`.

Libraries and tools are generated in depends by running `make HOST=${triplet}`. This step first downloads tar archives of the required dependency source files and places them in sources/. These archives are then checked for their sha256sum hash and compared with a hardcoded value in the package recipe.  They are then extracted and placed into work/build during a build process. After building the package binaries/libraries are installed in work/staging . Once completed their staging directory is compressed and placed into built/ , after which their content in work/ is deleted. When every package is in built/ their archives are decompressed into ${triple}/ . This is where the monero build system will pick them up from. 

A cmake toolchain file is uniquely generated for ${triple} and placed in `contrib/depends/${triplet}/share/toolchain.cmake`. They can then be accessed by the monero build system by executing for example: ``cmake -DCMAKE_TOOLCHAIN_FILE=`pwd`/contrib/depends/${triplet}/share/toolchain.cmake`` . The toolchain file also sets the DEPENDS variable to identify a depends build to the monero build system. This is used to exclude some additions from external as well as correcting some make flags.

General usage is also briefly reflected in the Readme.md under "Building portable statically linked binaries".

### Example of a depends build for 64 bit windows: 

Install build essentials, nsis and g++-mingw-w64-x86-64 on your host system. Then:
```
    cd contrib/depends
    make HOST=x86_64-w64-mingw32
```
Once it is finished building go back to the root directory and execute: 
```
    cmake -DCMAKE_TOOLCHAIN_FILE=`pwd`/contrib/depends/x86_64-w64-mingw32/share/toolchain.cmake
```

### Advantages of using depends:
- Serves as a basis for reproducible builds on all plattforms
- Allows cross compilation of windows and macos monero binaries
[...]

PR #7435, opened 2021-03-05
Mac: Enable on ARM, by not enforcing x86_64

mj-xmr:
Fixes https://github.com/monero-project/monero/issues/7392

The use case for this change is ARM Mac (i.e. the "M1" SoC)
Both x86_64 Mac builds (real and `depends`) compile fine.

@selsta : Would you be so kind and run a short test on your Mac with this change?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -484,7 +484,7 @@ if (CMAKE_SYSTEM_NAME MATCHES "(SunOS|Solaris)")
 endif ()
 
 if (APPLE AND NOT IOS)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=x86-64 -fvisibility=default -std=c++11")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=default -std=c++11")
   if (NOT OPENSSL_ROOT_DIR)
       EXECUTE_PROCESS(COMMAND brew --prefix openssl
         OUTPUT_VARIABLE OPENSSL_ROOT_DIR


moneromooo-monero:
Do you know why this got added, and why removing it is fine ?

mj-xmr:
> Do you know why this got added, and why removing it is fine ?

I don't know why it was added. It appears to me, that it's an overfitting assumption. I need to gather tests from owners of the machines. I will also check the resulting architecture of the produced artifacts on the CI. If it's x86_64 on x86_64 machines, then I see no problem in removing this enforcement. Or do you suggest more fine grained approach, like allowing the user to set the architecture specifically?

mj-xmr:
I got this result for Mac:
 build/Darwin/_HEAD_detached_at_pull_7435_merge_/release/bin/monero-wallet-cli: Mach-O 64-bit executable x86_64

and this for Cross-Mac:
/home/runner/work/monero/monero/build/x86_64-apple-darwin11/release/bin/monero-wallet-cli: Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|WEAK_DEFINES|BINDS_TO_WEAK|PIE|HAS_TLV_DESCRIPTORS>

mj-xmr:
[Good news](https://github.com/monero-project/monero/issues/7392#issuecomment-792050564) from the issue starter, @jbrot:

"When I was validating @mj-xmr 's branch above, I noticed that this test no longer crashed. Some further investigation revealed that adding the -fvisibility=default flag is what caused the issue to appear. (...)"

iamamyth:
The hard-coded MacOS arch appears to have been introduced in this [commit](https://github.com/monero-project/monero/commit/56b6e41ea7b7c17b513c24797d234e5cf9df8fa2) which in turn was part of this [PR](https://github.com/monero-project/monero/pull/3430). As best as I can tell from the PR comments and the commit itself, no reason was given for forcing the arch to x86-64, and it may very well have been unwarranted. The only reason I can imagine for specifying a fixed arch would be to ensure the depends-based build system picked the right architecture for cross-compilation, but that doesn't make a whole lot of sense, as the depends build system infers the target architecture from the HOST build triplet.

mj-xmr:
I reverted the workflows changes, so that the branch can be merged.

iamamyth (approved):

PR #7439, opened 2021-03-06
Reduced executable size; reduced call sequence to "allowed" log function

vtnerd:
This small patch reduces `monerod` by ~3.1%, `monero-wallet-rpc` by ~2.8%, and `monero-wallet-cli` by ~2.4%. This numbers are from a static build with gcc 9.3.

This is a bit of micro-optimization, but this will help the CPU instruction cache slightly. Every call to log a message : (1) calls a function to get a global pointer, (2) constructs a `std::string`, (3) dereferences a pointer, (4) sets 3 register values for a function call. This moves steps (1), (2), and (3) to a single instance in another cpp file and reduces the number of register values needed to 2.

This also moves the "quick check" variable to static memory and delays `std::string` construction, which shortens the number of instructions executed to do the quick check. Since an overwhelming majority of calls "fails" this quick check, this improves things a bit further.

### DIFF
--- a/contrib/epee/include/misc_log_ex.h
+++ b/contrib/epee/include/misc_log_ex.h
@@ -41,7 +41,7 @@
 #define MAX_LOG_FILES 50
 
 #define MCLOG_TYPE(level, cat, color, type, x) do { \
-    if (ELPP->vRegistry()->allowed(level, cat)) { \
+    if (el::Loggers::allowed(level, cat)) { \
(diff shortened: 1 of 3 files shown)


moneromooo-monero (approved):

vtnerd:
Clang complained about an ambiguity between two `allowed` overloads. I renamed the new function `priority_allowed` to differentiate the function calls.

selsta (approved):

PR #7460, opened 2021-03-07
async_protocol_handler_config: fix deadlock

perfect-daemon:


moneromooo-monero (approved):

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -860,18 +856,19 @@ int async_protocol_handler_config<t_connection_context>::invoke_async(int comman
...
     for (auto &aph: conn)
       aph->finish_outer_call();
   });
 
+  CRITICAL_REGION_BEGIN(m_connects_lock);
+  conn.reserve(m_connects.size());
   for (auto &e: m_connects)
     if (e.second->start_outer_call())
       conn.push_back(e.second);
+  CRITICAL_REGION_END()
```
vtnerd, line 871: I have a comment about this in my conversion to `shared_ptr`/`weak_ptr` patch - I would revert this line right here. Several of the callbacks appear to assume that they have some exclusive access with respect to other calls to the `foreach_connection`. So unless you've stepped through the logic, its safer to revert this. Which might be an issue for deadock, which means recursive mutex.

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -769,36 +769,32 @@ void async_protocol_handler_config<t_connection_context>::del_connection(async_p
...
+    for (auto &aph: connections)
+      aph->finish_outer_call();
+  });
+
   CRITICAL_REGION_BEGIN(m_connects_lock);
   for (auto& c: m_connects)
   {
     if (c.second->m_connection_context.m_is_income == incoming)
-      connections.push_back(c.first);
+      if (c.second->start_outer_call())
```
vtnerd, line 783: Do the test cases below trigger something? The object has to wait in the destructor for `m_connects_lock`, which is somewhat gross, but it cannot be fully cleaned up. It looks safer to add this, but I'm not sure that its strictly necessary (probably a red herring for problems you've observed).

The `shared_ptr`/`weak_ptr` patch also does this, but its mandatory to get the real pointer from the `weak_ptr`.

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -883,11 +880,8 @@ bool async_protocol_handler_config<t_connection_context>::foreach_connection(con
 template<class t_connection_context> template<class callback_t>
 bool async_protocol_handler_config<t_connection_context>::for_connection(const boost::uuids::uuid &connection_id, const callback_t &cb)
 {
-  CRITICAL_REGION_LOCAL(m_connects_lock);
-  async_protocol_handler<t_connection_context>* aph = find_connection(connection_id);
-  if (!aph)
-    return false;
-  if (!aph->start_outer_call())
+  async_protocol_handler<t_connection_context>* aph = nullptr;
```
vtnerd, line 883: Again, I'm not sure about the thread-safety expectations of the callbacks here.

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -953,12 +947,14 @@ int async_protocol_handler_config<t_connection_context>::send(byte_slice message
...
-  if (!aph)
+  async_protocol_handler<t_connection_context>* aph = nullptr;
+  if (find_and_lock_connection(connection_id, aph) != LEVIN_OK)
     return false;
+  auto scope_exit_handler = misc_utils::create_scope_leave_handler(
+    boost::bind(&async_protocol_handler<t_connection_context>::finish_outer_call, aph));
   if (!aph->close())
     return false;
+  CRITICAL_REGION_LOCAL(m_connects_lock);
   m_connects.erase(connection_id);
```
vtnerd, line 958: This `erase` call cannot be outside of the lock, and therefore I don't see the benefits to the changes in this function.

vtnerd:
> 1. Provided unit test demonstrates few ways to encounter deadlock.

This is response to which comment? Please respond to the actual comments inline, it makes the conversation more clear. You might be responding purely via for privacy purposes, so at least paste the comment with a `>` so I can keep track.

> 2. It's expected the patch is correct and fixes the above deadlock.
>3. If there is any edge case that isn't handled properly then provide any verifiable proof in order to crack this solution.

Thread sanitizer already complains about the context object numerous times. Which is primarily why I wasn't willing to push it further without looking at the existing usage.

Most of the calls are "read" calls but I'm not sure that all are. How closely did you look? Did you run thread-santizer at all?

contrib/epee/include/net/levin_protocol_handler_async.h
vtnerd: Sorry read through this code too quickly earlier, this is moving the lock call downward instead of doing a `CRITICAL_REGION_END` call.

This is now acquiring the lock twice. I'm not immediately seeing where deadlock would occur, after running through the call sequence of `aph->close()`. If I revert these changes, the unit test will deadlock here? This call is used in many places throughout the code, so I'm a little skeptical is necessary.

perfect-daemon:
```
if there are so many places where you are not sure then try to find at least one edge case that would prove incorrectness of this patch.
But It's expected that there are no such cases.
```

contrib/epee/include/net/levin_protocol_handler_async.h
vtnerd: This is long, but please read the first paragraph to at least understand my point -

There's two different potential deadlock cases in this function, and these tests only cover one of them (presumably because @selsta stacktraces only contained this case). If you move this lock release to the bottom, the tests still pass. The "primary" deadlock case is the connection releases occurring before the lock release (the order of which was changed above). This occurs when `foreach_connection` has the last reference to a connection.

The second deadlock case is dependent on callback behavior; the callback tries to acquire `m_connects_lock` somehow. Unfortunately, there are "naked" reads/writes occurring in these callbacks (see below), so its not obvious that this is an improvement (it might be, but I think this is subjective). My recommendation was/is to not release the lock here, and to keep the change above where the scoped release order changes, since (afaik) there are no known deadlock cases but known race conditions. Also, there's no test for this right now.

-----
"Proof":

There's already naked read/write calls in `foreach_connection` - here's an [instance of `m_state` being written](https://github.com/monero-project/monero/blob/master/src/cryptonote_protocol/cryptonote_protocol_handler.inl#L1690), with many [read calls inside of `for_each_connection`](https://github.com/monero-project/monero/blob/master/src/cryptonote_protocol/cryptonote_protocol_handler.inl#L1771) (which forwards to `foreach_connection`) within this file. There are more read calls that just this one instance.

This particular variable is already an issue elsewhere, but removing the lock hold is exacerbating the problem. Unless the proof you want is showing that this is worse than deadlock - which is difficult because this type of UB can be harder to identify than deadlock. Thread-saniziter already complains constantly about the context object - the code was mosty written with the assumption that the context was exclusively being accessed.

perfect-daemon:
```
The above thread unsafe access to connection context must be fixed separately
  since it existed before this patch and will continue to exist after.
Provided unit test has been updated to catch both ways to encounter deadlock
  within async_protocol_handler_config<T>::foreach_connection().
```

vtnerd: > The above thread unsafe access to connection context must be fixed separately
  since it existed before this patch and will continue to exist after.

I agree this is where the code needs to eventually be, but as stated above, whether this better now is subjective since it adds more race conditions but fixes no known deadlocks caused by real callbacks. Deadlocks stop progress and can be backtraced; race conditions are harder to diagnose.

If others seem fine with this, then this is difficult to object. My recommendation would be to merge on `master` first for some testnet stressing before merging onto 0.17.

> Provided unit test has been updated to catch both ways to encounter deadlock
  within async_protocol_handler_config<T>::foreach_connection().

Might as well add one to `for_connection`, the situation is identical.

perfect-daemon:
```
Provided unit test has been updated to catch both ways to encounter deadlock
  within async_protocol_handler_config<T>::for_connection().
```

selsta (approved):

PR #7469, opened 2021-03-07
core_tests: cache data for each test separately, then replay

### LINKED ISSUE #7442: Tests: core_tests caching generation from the test perspective
I still want to speed up the core tests, by addressing the generation bottleneck [mentioned here](https://github.com/monero-project/monero/pull/7300), but without affecting the implementation itself. 

For a recap, I discovered there, that 30% of the core_tests time is taken by the generation step, so I wanted to cache it by diving deep into the implementation and putting there a cache. The solution was not accepted for valid reasons. I would like to propose a different solution: give the core_tests a new parameter: `--generate-cache-and-play`, which would perform serialization on the first run, and deserialization on the second run. For safety, the cache would have an ID, based on the md5 sum of the core_tests binary.

Would such a solution be accepted? Any objections?

mj-xmr:
Fixes https://github.com/monero-project/monero/issues/7442

### Results
After the initial generation, the next run of `core_tests` takes about 8 minutes, which compared to the usual run of an hour, is an **8 fold time reduction**, or **52 minutes**.
The whole test cache weights only 6 MB (compressed).

### Description
`core_tests` caches data for each test separately, then replays them. On the next run of the workflow, the generated files are reused with the exception of those data, which deal with time. These have to be regenerated each time.
This has been achieved by chaining the `core_tests` execution, using arguments: `--generate_test_data` and `--play_test_data` in the `core_tests` CMakeLists, as suggested in https://github.com/monero-project/monero/issues/7442.

The implementation of `core_tests` has been changed in such way, that the previous macros have been extended. The macros `GENERATE` and `PLAY` were not being used previously in a realistic use case. Now they are and they reuse some of the macro code.

### User's perspective
The user can control, whether the data is to be reused (yes, by default), or regenerated each time the `core_tests` is executed, via a new option `TESTS_REUSE_CACHE`. The option is used to steer the `add_test()` command in `core_tests`'s `CMakeFile.txt`. If the option is set to OFF, then the previous `core_tests` argument is used (`--generate_and_play_test_data`), which generates the data in place, without storing them, like it was done before.


mj-xmr:
Demo commit completed. Please see the logs of `test_ubuntu` job.

.github/workflows/build.yml
```diff
@@ -3,55 +3,7 @@ name: ci/gh-actions/cli
 on: [push, pull_request]
 
 jobs:
-  build-macos:
```
mj-xmr, line 6: This will be reverted, except the `test_cache` section

Makefile
```diff
@@ -96,7 +96,8 @@ release: cmake-release
 
 release-test:
 	mkdir -p $(builddir)/release
-	cd $(builddir)/release && cmake -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=release $(topdir) && $(MAKE) && $(MAKE) test
```
mj-xmr, line 99: This will be reverted.

iamamyth:
My initial reaction would be that this could be a lot simpler using a collection of bit flags for the "mode" (e.g. 0 == none, 1 == generate, 2 == play), then you can simply set bits based on the command-line args (generate or play), rather than using compound args, and, similarly, act on each individual bit in the execution statements, rather than checking for the compound state specifically.

mj-xmr:
@iamamyth : What the mode Generate and its macro do is actually generating and serializing the generated sequence. The mode Play deserializes the sequence and plays them. So they should really be called GenerateStore and RestorePlay.
The mode GenerateAndPlay is different in this regard, that it generates sequences in memory but doesn't store or restore them (as per the original design). So the GenerateAndPlay is actually quite a distinct mode. 

With that in mind, do you still think, that a bit mask would help here?

iamamyth:
I see your point, so no, a bit mask probably isn't the right approach. The existing code seems to use the term "replay" to mean making use of the in-memory cache, "play" to mean reading from disk into cache and then replaying, and "generate" to mean constructing and optionally saving the in-memory cache. So, I would advise maintaining those definitions, in which case, `GENERATE_AND_PLAY` is a misnomer, it should be GENERATE_AND_REPLAY. It would have been a lot clearer if "populate" meant filling the cache and generate meant filling it and storing the data, then you'd have POPULATE_AND_PLAY, but I'd avoid mucking about with the prior naming convention.

mj-xmr:
I will keep this in mind. This part of code needs quite some refactoring anyway on various dimensions, but currently it's hard to work with something, that takes 1 hour to execute.

iamamyth:
Right, state bits could make sense in a future refactor, as logically you have 4 possible operations: generate, save, restore, and execute. Of course, some bits depend on others, so you really have a state consisting of two components: generate/restore (the two options should be exclusive of each other) and execute+save (not mutually exclusive). But, as I said, I'd keep the current naming convention so long as most of the code establishing said convention stays in place.

mj-xmr:
Since there were no more comments, I reverted the simplifications and prepared the branch for merging.

moneromooo-monero:
How does this tell when the cache should be regenerated ?

tests/core_tests/chaingen.h
```diff
@@ -981,25 +990,41 @@ inline bool do_replay_file(const std::string& filename)
...
-        if (!tools::serialize_obj_to_file(events, filename)) \
-        { \
-            MERROR("Failed to serialize data to file: " << filename); \
-            throw std::runtime_error("Failed to serialize data to file"); \
-        } \
+#define GENERATE(filename, genclass)                                                                       \
+  if (epee::file_io_utils::is_file_exist(filename))                                                        \
+  {                                                                                                        \
+    MGINFO_GREEN("Already generated file: " << filename);                                                  \
+  }                                                                                                        \
```
moneromooo-monero, line 997: I meant to comment on that part.

mj-xmr: > How does this tell when the cache should be regenerated ?

Unfortunately this is the missing part. Handling this in a smart way would overbloat the PR, which is already large.
What I can offer right now, is setting the CMake variable `TESTS_REUSE_CACHE` to `ON` in the Makefile target, so that it's clearer to the caller, that he can set it to `OFF`, if his PR is about working with the regeneration of the data. 
Would that be enough for this PR?

moneromooo-monero: So if you're running as a one off, it won't speed up anything, and if you're running many times, presumably because you're changing the code, you have to reconfigure every time you know the generated data changed, or did not change when it had changed the time before ? When does this actually speed things up ?

mj-xmr: The cache is being created on the first run.
Only when you know, that your changes alter the generated part, you could alter the Makefile to switch the cache to off. 

There is a risk, that somebody generates the original cache, alters the generation with a failty code, reuses the cache and the tests pass.

Therefore I'd propose:
1) leaving the original generation on the fly as the default option (so no caching),
2) allowing the contributor to temporarily change it (cache the data) until the PR is in shape (or just doing this locally),
3) expecting the contributor to change it back to the original form,
4) confirming, that the tests pass with the original in-place generation.

moneromooo-monero: It seems much simpler to just use keccak instead of randomx in core_tests. It speeds up massively and avoids issues of bad cache (that's what I use). Then again, if the default is to not cache, I'm ok with this patch too.

mj-xmr: > It seems much simpler to just use keccak instead of randomx in core_tests. It speeds up massively and avoids issues of bad cache (that's what I use). Then again, if the default is to not cache, I'm ok with this patch too.

An interesting idea to explore.

I set the caching to OFF by default.

mj-xmr:
Removed the CI changes, as requested by @selsta . The change will now only be able to affect local runs.

selsta:
Closing as there was not much support for merging this and core tests can be sped up in other ways.

PR #7480, opened 2021-03-08
Epee: CppCheck fixes

mj-xmr:
Fixes some potential memory access bugs and styling, as well as adds braces around one-liner loop iterations.
Current scope: Epee (to keep the branches easier to review)

### DIFF
--- a/contrib/epee/src/abstract_http_client.cpp
+++ b/contrib/epee/src/abstract_http_client.cpp
@@ -27,17 +27,14 @@ namespace net_utils
   std::string dec_to_hex(char num, int radix)
   {
     std::string csTmp;
-    int num_char;
-
-    num_char = (int) num;
+    int num_char = (int) num;
     if (num_char < 0)
       num_char = 256 + num_char;
 
     while (num_char >= radix)
     {
-      const int temp = num_char % radix;
+      csTmp = get_hex_vals()[num_char % radix];
       num_char = (int)floor((float)num_char / (float)radix);
-      csTmp = get_hex_vals()[temp];
     }
 
     csTmp += get_hex_vals()[num_char];
--- a/contrib/epee/src/abstract_http_client.cpp
+++ b/contrib/epee/src/abstract_http_client.cpp
@@ -26,7 +26,6 @@ namespace net_utils
   //----------------------------------------------------------------------------------------------------
   std::string dec_to_hex(char num, int radix)
   {
-    int temp=0;
     std::string csTmp;
     int num_char;
 
@@ -36,7 +35,7 @@ namespace net_utils
 
     while (num_char >= radix)
     {
-      temp = num_char % radix;
+      const int temp = num_char % radix;
       num_char = (int)floor((float)num_char / (float)radix);
       csTmp = get_hex_vals()[temp];
     }
--- a/contrib/epee/src/byte_slice.cpp
+++ b/contrib/epee/src/byte_slice.cpp
@@ -146,8 +146,10 @@ namespace epee
     : byte_slice()
   {
     std::size_t space_needed = 0;
-    for (const auto source : sources)
+    for (const auto & source : sources)
+    {
       space_needed += source.size();
+    }
 
     if (space_needed)
     {
@@ -155,7 +157,7 @@ namespace epee
       span<std::uint8_t> out{reinterpret_cast<std::uint8_t*>(storage.get() + 1), space_needed};
       portion_ = {out.data(), out.size()};
 
-      for (const auto source : sources)
+      for (const auto & source : sources)
       {
         std::memcpy(out.data(), source.data(), source.size());
         if (out.remove_prefix(source.size()) < source.size())
--- a/contrib/epee/src/mlog.cpp
+++ b/contrib/epee/src/mlog.cpp
@@ -204,20 +204,21 @@ void mlog_configure(const std::string &filename_base, bool console, const std::s
           static_assert(std::is_integral<time_t>(), "bad time_t");
           return ta < tb;
         });
-        for (size_t i = 0; i <= found_files.size() - max_log_files; ++i)
+        for (size_t i = 0; i < found_files.size() - max_log_files; ++i)
         {
           try
           {
+            const boost::filesystem::path & found_file = found_files.at(i);
             boost::system::error_code ec;
-            boost::filesystem::remove(found_files[i], ec);
+            boost::filesystem::remove(found_file, ec);
             if (ec)
             {
-              MERROR("Failed to remove " << found_files[i] << ": " << ec);
+              MERROR("Failed to remove " << found_file << ": " << ec);
             }
           }
           catch (const std::exception &e)
           {
-            MERROR("Failed to remove " << found_files[i] << ": " << e.what());
+            MERROR("Failed to remove file at index " << i << ": " << e.what());
           }
         }
       }
--- a/contrib/epee/src/mlog.cpp
+++ b/contrib/epee/src/mlog.cpp
@@ -204,7 +204,8 @@ void mlog_configure(const std::string &filename_base, bool console, const std::s
           static_assert(std::is_integral<time_t>(), "bad time_t");
           return ta < tb;
         });
-        for (size_t i = 0; i < found_files.size() - max_log_files; ++i)
+        const int max_index_found_files = static_cast<int>(found_files.size()) - static_cast<int>(max_log_files); // Prevents unsigned var overflowing
+        for (int i = 0; i < max_index_found_files; ++i)
         {
           try
           {
--- a/contrib/epee/src/network_throttle-detail.cpp
+++ b/contrib/epee/src/network_throttle-detail.cpp
@@ -81,7 +81,7 @@ namespace net_utils
 
 class connection_basic_pimpl {
 	public:
-		connection_basic_pimpl(const std::string &name);
+		explicit connection_basic_pimpl(const std::string &name);
 
 		static int m_default_tos;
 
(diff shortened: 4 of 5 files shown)


contrib/epee/src/mlog.cpp
```diff
@@ -204,20 +204,21 @@ void mlog_configure(const std::string &filename_base, bool console, const std::s
...
             if (ec)
             {
-              MERROR("Failed to remove " << found_files[i] << ": " << ec);
+              MERROR("Failed to remove " << found_file << ": " << ec);
             }
           }
           catch (const std::exception &e)
           {
-            MERROR("Failed to remove " << found_files[i] << ": " << e.what());
+            MERROR("Failed to remove file at index " << i << ": " << e.what());
```
mj-xmr, line 221: Here I decided, that we cannot assume, that the file was accessible from the vector at all.

contrib/epee/src/network_throttle-detail.cpp
```diff
@@ -279,7 +282,7 @@ void network_throttle::calculate_times(size_t packet_size, calculate_times_struc
 	//				update_overheat();
 	cts.average = Epast/cts.window; // current avg. speed (for info)
 
-	if (Epast <= 0) {
+	if (Epast == 0) {
```
mj-xmr, line 285: An unsigned value cannot be negative.

contrib/epee/src/mlog.cpp
```diff
@@ -204,20 +204,21 @@ void mlog_configure(const std::string &filename_base, bool console, const std::s
           static_assert(std::is_integral<time_t>(), "bad time_t");
           return ta < tb;
         });
-        for (size_t i = 0; i <= found_files.size() - max_log_files; ++i)
+        for (size_t i = 0; i < found_files.size() - max_log_files; ++i)
```
mj-xmr, line 207: If `max_log_files` happens to be 0 for some reason, then accessing the files under the condition of <= will crash.

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -36,7 +35,7 @@ namespace net_utils
 
     while (num_char >= radix)
     {
-      temp = num_char % radix;
+      const int temp = num_char % radix;
```
iamamyth, line 38: Would suggest just inlining this in the array indexing operation, rather than having a single-use variable named `temp`.

iamamyth: Note: The above comment requires moving the `csTmp` assign above the `num_char` assign.

contrib/epee/src/mlog.cpp
```diff
@@ -204,7 +204,8 @@ void mlog_configure(const std::string &filename_base, bool console, const std::s
           static_assert(std::is_integral<time_t>(), "bad time_t");
           return ta < tb;
         });
-        for (size_t i = 0; i < found_files.size() - max_log_files; ++i)
+        const int max_index_found_files = static_cast<int>(found_files.size()) - static_cast<int>(max_log_files); // Prevents unsigned var overflowing
+        for (int i = 0; i < max_index_found_files; ++i)
```
mj-xmr, line 208: What do you think of this change? This will prevent the index to overflow if the `max_log_files` is greater than `found_files.size()`
I also need an advice. Would simply writing:

`static_cast<int>(found_files.size() - max_log_files);`

be enough?

iamamyth: I think you're missing the context here, all of this code is inside `if (found_files.size() >= max_log_files)`.

contrib/epee/src/byte_slice.cpp
```diff
@@ -146,16 +146,18 @@ namespace epee
     : byte_slice()
   {
     std::size_t space_needed = 0;
-    for (const auto source : sources)
+    for (const auto & source : sources)
```
vtnerd, line 149: Did CppCheck blindly recommend to add an `&` here? There's little benefits since its a 16-byte value with a trivial copy-constructor - the compiler is free to optimize as necessary anyway.

contrib/epee/src/mlog.cpp
xiphon: > all of this code is inside if (found_files.size() >= max_log_files).

Ah, yeah. The `max_log_files > found_files.size()` case is already handled.

moneromooo-monero:
A number of those seem to be gratuitous changes. I guess it's a lot less diff space than some of the other patches though :)
Which one is the bad memory access ?

mj-xmr:
I have to get a little bit philosophical, which I normally dislike. 

I am fixing “potential” mistakes here. You can always argue, that in currently assumed outer conditions a given crash will never occur. You can’t guarantee to me however, that somebody in future doesn’t:
- change the conditions in a way, that your assumption doesn’t apply anymore and the reviewers overlook it,
- copy-paste the inner code to a different place, omitting the outer conditions which serve as restrictions, preventing from crashing.

Analogous to unit testing, you don’t only unit test your code for the happy path, but also (or rather especially) for corner cases – situations which typically don’t happen, but could happen when some unusual combination of input parameters is chosen.

contrib/epee/src/mlog.cpp
```diff
@@ -204,20 +204,22 @@ void mlog_configure(const std::string &filename_base, bool console, const std::s
...
         });
-        for (size_t i = 0; i <= found_files.size() - max_log_files; ++i)
+        const int max_index_found_files = static_cast<int>(found_files.size()) - static_cast<int>(max_log_files); // Prevents unsigned var overflowing
+        for (int i = 0; i < max_index_found_files; ++i)
         {
           try
           {
+            const boost::filesystem::path & found_file = found_files.at(i);
             boost::system::error_code ec;
-            boost::filesystem::remove(found_files[i], ec);
```
mj-xmr, line 212: @moneromooo-monero : Here's the potential UB. Replaced [] access with .at().

contrib/epee/src/byte_slice.cpp
mj-xmr: No. This was my own initiative while touching the package.

The compiler optimizes this only in the release mode. I see little reason not to use a const reference, since it doesn't hurt. Especially if `auto` is used, one can't assume, that in future the `sources` remains the same 16-byte value, for which the const reference currently brings no benefit, I admit.

I can still revert it, if it's not convincing for you.

mj-xmr:
> I guess it's a lot less diff space than some of the other patches though :)

_You can't make an omelette without breaking a few eggs_
With that said, some of these other patches pass all tests, are already reviewed and approved ;)

contrib/epee/src/mlog.cpp
mj-xmr: What happens if the `found_files` is empty and `max_log_files` is zero?

Then the only line, preventing the code from crashing is [at line 174](https://github.com/monero-project/monero/pull/7480/files#diff-6676c0bac6e2626df8a81502a138684333631486fd1493787b383d37750e9890R174), located whole 30 lines above. So far away, that it seems out of context.

contrib/epee/src/mlog.cpp
mj-xmr: Because `.at(i)` throws an exception when `i` is out of the container's bounds. In such case, another call to `.at(i)` would kill the purpose of catching the exception in the first place.

contrib/epee/src/mlog.cpp
moneromooo-monero: I don't see it and I'm getting really suspicious it's cargo cult given the accompanying changes. Can you give details of how you end up with an out of bounds here ?

iamamyth:
> I am fixing potential mistakes here. You can always argue, that under the currently assumed outer conditions a given crash will never occur.

This logic falls prey to infinite regress: No code ever works if one arbitrarily detaches it from the surrounding context. In cases where the context proves especially critical and volatile, there are comments and assertions. But, making code which can never be broken via edit is a literal impossibility (even an assert one line prior could be accidentally deleted, edited, or not copied). Moreover, proceeding based on such an assumption of fragility makes the code harder to read and write and therefore increases the likelihood of the very mistakes you aim to avoid (e.g. some of these loop-related edits are horribly unreadable, as they imply the loop could overflow the bounds in certain scenarios, when, in fact, care has been taken to prevent that problem; therefore, the reader assumes the cognitive burden of double-checking bounds everywhere, as you've arbitrarily decided the standard technique of "prevent overflow via conditional" isn't good enough, which really begs the question of how one avoids an overflow).

iamamyth:
One additional thought: I don't mind grouping the occasional change to avoid one-liner PR spam, but grouping unrelated changes into a "fix lint" PR really should not be done, as it makes review very difficult. Linter-flagged problems suggest a particular review methodology, whereas other work requires a totally different standard of scrutiny. So, applying typical linter-related PR rules, these changes don't pass muster. What you've done here is similar to changing an actual line of code in an "update copyright" PR, please avoid such things in the future (and no, I don't mean to imply any untoward motivation).

contrib/epee/src/mlog.cpp
SChernykh: We should settle at not touching this place at all, given the previous check `if (found_files.size() >= max_log_files)`. Don't just do changes because cppcheck said so, it's a dumb static analyzer. At most, you can replace `i <= found_files.size() - max_log_files` with `i  + max_log_files <= found_files.size()` if you worry about it.

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -26,19 +26,15 @@ namespace net_utils
   //----------------------------------------------------------------------------------------------------
   std::string dec_to_hex(char num, int radix)
```
SChernykh, line 27: This whole function is an abomination of the common sense (not related to your changes though). Why call it `dec_to_hex` and then supply `radix` argument? If it's hex only, remove the argument and rewrite it with just a few lines of code to make 2 hex characters out of a single-byte `char` value. Using `std::string` and `std::reverse` here is a cherry on the top, why? To make it 10 times slower than needed with some memory allocation? Example: http://cpp.sh/9xt2hk

SChernykh: Just checked: it's used only in 1 place and only with radix = 16, so it makes sense to rewrite it completely

contrib/epee/src/mlog.cpp
SChernykh: There's no way `found_files[i]` can be out of range, given the for-loop check above. We lose valuable information (file name) in the log with this change. Not recommended.

SChernykh (changes requested):

PR #7481, opened 2021-03-08
Warnings: treat warnings as errors: 'switch' & 'return-type'

mj-xmr:
Treating such warnings as errors, that when ignored are so severe, that they can segfault or even [UB](https://en.cppreference.com/w/cpp/language/ub) any application.

- `return-type`: aborts compilation when you forget to return a value from a function, when the caller expects it (or worse: there's a unnoticed branch of a complex function, which ends without returning the expected value)
- `switch`: aborts compilation when a new case isn't handled for a newly added enum entry. With this error in place, you can safely delete all `default:` cases made for runtime safety checks and enjoy static checks instead (so less need for time consuming  tests execution, as well as writing).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -744,6 +744,13 @@ else()
     add_linker_flag_if_supported(-Wl,--high-entropy-va LD_SECURITY_FLAGS)
   endif()
 
+  # Warnings, that when ignored are so severe, that they can segfault or even UB any application.
+  # Treat them as errors.
+  add_c_flag_if_supported(  -Werror=switch      C_SECURITY_FLAGS)
+  add_cxx_flag_if_supported(-Werror=switch      CXX_SECURITY_FLAGS)
+  add_c_flag_if_supported(  -Werror=return-type C_SECURITY_FLAGS)
+  add_cxx_flag_if_supported(-Werror=return-type CXX_SECURITY_FLAGS)
+
   message(STATUS "Using C security hardening flags: ${C_SECURITY_FLAGS}")
   message(STATUS "Using C++ security hardening flags: ${CXX_SECURITY_FLAGS}")
   message(STATUS "Using linker security hardening flags: ${LD_SECURITY_FLAGS}")


CMakeLists.txt
```diff
@@ -756,6 +756,11 @@ else()
   # is fixed in the code (Issue #847), force compiler to be conservative.
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-strict-aliasing")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-strict-aliasing")
+  
+  # Warnings, that when ignored are so severe, that they can segfault or even UB any application.
+  set(WARNINGS_AS_ERRORS "-Werror=switch -Werror=return-type")
```
vtnerd, line 761: From the [Gcc docs](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html):

> Warn whenever a switch statement has an index of enumerated type and lacks a case for one or more of the named codes of that enumeration. (The presence of a default label prevents this warning.) case labels outside the enumeration range also provoke warnings when this option is used (even if there is a default label). This warning is enabled by -Wall. 

The project should already have `-Wall` enabled, so this warning should already be enabled. And a `default` case disables the warning; the compiler doesn't track existing labels for new ones.

CMakeLists.txt
```diff
@@ -756,6 +756,11 @@ else()
...
+  set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS}   ${WARNINGS_AS_ERRORS}")
```
vtnerd, line 762: `add_c_flag_if_supported` and `add_cxx_flag_if_supported` might be useful here.

CMakeLists.txt
mj-xmr: Yes, `-Wall` is enabled. Once the `switch` warning is treated as error, I could then go through all the switch/cases and remove the `default:` labels, triggering a compilation warning treated as error, if a newly case was forgotten to be handled.

Maybe not all `default:` cases will be removable, but any default case, which throws an exception due to unhandled case is a target, as an example.

vtnerd (approved):

PR #7489, opened 2021-03-09
CI: ccache for ubuntu-test

mj-xmr:
Result:
Typical run of the ubuntu-test lasts 2 hours. With ccache, the time is reduced down to 1:25 h, so by about 25%.

The reason why ccache was not enabled for test-ubuntu job, was that the compiled binaries for the (by default) native architecture, could be making use of registers (like AXV2) not available at a different host, where the test execution happens on the next run of the Action (as per recent IRC discussion with @selsta ).

The solution for this problem is using the generic build, achievable in Monero by setting the CMake variable ARCH to "default". In order not to interfere with the already existing Makefile target, designed for testing, the `release-test`, I have created an additional, new target, made specifically for CI, called `release-test-ci`, as suggested by @vtnerd some time ago. While creating the new target, I thought, that it would be also a good opportunity to test the dynamic linkage, as this dimension is not tested in neither of the other Workflow jobs. It gives a small linkage speed boost as well, but this is not the most important thing for this job.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -115,10 +115,19 @@ jobs:
   test-ubuntu:
     needs: build-ubuntu
     runs-on: ubuntu-latest
+    env:
+      CCACHE_COMPRESS: 1
+      CCACHE_TEMPDIR: /tmp/.ccache-temp
     steps:
     - uses: actions/checkout@v1
       with:
         submodules: recursive
+    - name: ccache
+      uses: actions/cache@v2
+      with:
+        path: ~/.ccache
+        key: test-ubuntu-ccache-${{ github.sha }}
+        restore-keys: test-ubuntu-ccache-
     - name: remove bundled boost
       run: sudo rm -rf /usr/local/share/boost
     - name: set apt conf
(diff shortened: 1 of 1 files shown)


moneromooo-monero (approved):
It's a bit disquieting if this speeds things up since it implies there is state leak between runs, but hey, it looks fine.

mj-xmr:
> It's a bit disquieting if this speeds things up since it implies there is state leak between runs, but hey, it looks fine.

The state is shared for each job, which uses any cache. This one is an exception where you actually want to execute the compiled binaries.

I'd say, that it's a good test, proving that the `ARCH="default"` delivers the expected result. Somebody could break its handling in CMakeLists.txt and we wouldn't notice.

mj-xmr:
Since run No. 443 of my Actions, I'm running a [scheduled runs for this branch](https://github.com/mj-xmr/monero/actions/workflows/build.yml?query=event%3Aschedule), to make sure, that the default architecture doesn't break the tests.

mj-xmr:
@selsta 
Test concluded with 16/16 [successful runs.](https://github.com/mj-xmr/monero/actions/workflows/build.yml?query=event%3Aschedule)

selsta:
I don't think adding this to the makefile is necessary as it is only used for CI.

Just add the mkdir build, cmake .. -D part to the GitHub actions file.

mj-xmr:
> I don't think adding this to the makefile is necessary as it is only used for CI.
> 
> Just add the mkdir build, cmake .. -D... part to the GitHub actions file.

Done. Are we rebasing and merging now?

selsta (approved):

PR #7491, opened 2021-03-09
CI: fail-fast set to false for build.yml, to better leverage ccache

mj-xmr:
Copying part of the description from https://github.com/monero-project/monero/pull/7489 :

Since the compilation might fail prematurely, even on the first run, the default `fail-fast: true` strategy would not allow to store the ccache's output as Workflow's cache, leaving us with no possibility to use ccache until the first entirely successful compilation. The fail-fast strategy is therefore set to false for all jobs (except test-ubuntu, where there's currently no ccache enabled before https://github.com/monero-project/monero/pull/7489 is merged).

P.S. The depends.yml already disables the `fail-fast` strategy

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,8 @@ on: [push, pull_request]
 jobs:
   build-macos:
     runs-on: macOS-latest
+    strategy:
+      fail-fast: false # Allows to store the cache, even if there's an error. (i.e. during the compilation)
     env:
       CCACHE_COMPRESS: 1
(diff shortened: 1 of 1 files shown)


selsta:
> When set to true, GitHub cancels all in-progress jobs if any matrix job fails. Default: true

We don't use matrix so not sure if this is correct. Did you test this?

mj-xmr:
> > When set to true, GitHub cancels all in-progress jobs if any matrix job fails. Default: true
> 
> We don't use matrix so not sure if this is correct. Did you test this?

You're right. Either I misread this, or it was just my wishful thinking. Closing the PR.

PR #7534, opened 2021-03-12
What will happen to Monero when Riccardo Spagni will get charged with cryptocurrency fraud like McAfee?

finelyscuff:
https://news.sky.com/story/anti-virus-software-creator-john-mcafee-charged-with-fraud-in-cryptocurrency-scam-12237417

https://www.reddit.com/r/Monero/comments/6d5yt5/what_fluffypony_just_did_is_not_ok/

----

Spam or truth? There is only one way to find out. Look into the sources yourself.


moneromooo-monero:
spam

Issue #7539, opened 2021-03-12
bitmonero.log is full of "Failed to read line."


M0ns1gn0r:
I have 90 log files with only this message:

```
2021-03-12 19:48:12.919	12540	ERROR	console_handler	contrib/epee/include/console_handler.h:397	Failed to read line.
2021-03-12 19:48:12.919	12540	ERROR	console_handler	contrib/epee/include/console_handler.h:397	Failed to read line.
2021-03-12 19:48:12.919	12540	ERROR	console_handler	contrib/epee/include/console_handler.h:397	Failed to read line.
2021-03-12 19:48:12.919	12540	ERROR	console_handler	contrib/epee/include/console_handler.h:397	Failed to read line.
2021-03-12 19:48:12.919	12540	ERROR	console_handler	contrib/epee/include/console_handler.h:397	Failed to read line.
2021-03-12 19:48:12.919	12540	ERROR	console_handler	contrib/epee/include/console_handler.h:397	Failed to read line.
2021-03-12 19:48:12.919	12540	ERROR	console_handler	contrib/epee/include/console_handler.h:397	Failed to read line.
2021-03-12 19:48:12.920	12540	ERROR	console_handler	contrib/epee/include/console_handler.h:397	Failed to read line.
2021-03-12 19:48:12.920	12540	ERROR	console_handler	contrib/epee/include/console_handler.h:397	Failed to read line.
2021-03-12 19:48:12.920	12540	ERROR	console_handler	contrib/epee/include/console_handler.h:397	Failed to read line.
```

A new 100 MB file is produced every minute.

Monero GUI 0.17.1.9-release on Win 7 x64

moneromooo-monero:
Are you piping anything into whatever program generates this ? Or executing it from a non shell program, that kind of thing ?

Morbidlystrum671:
Take a look #7540

flushRelay34:
View #7543

M0ns1gn0r:
@moneromooo-monero 

I open `monero-wallet-gui.exe` from `Total Commander`, then the GUI automatically starts the `monerod.exe` (I use the simple mode). 

But the problem persists even if I start `monero-wallet-gui.exe` directly from the explorer.

Smartlycount707:
View #7547

Lastsew987:
Check out #7558

Wiselyconserve14:
View #7564

M0ns1gn0r:
[...]

PR #7610, opened 2021-03-17
Epee logging: Compression of rotated logs; Add zlib to Boost

mj-xmr:
It's now possible to use file compression using `epee::file_compression` namespace, which is a wrapper around `boost::iostreams`. The current use case is compression of rotated logs in `epee::mlog.cpp`, but it could be also used to compress for instance the generated test data from https://github.com/monero-project/monero/pull/7469.

The `epee::file_compression` also offers a decompression routine, which is able to decompress the data right into an abstract `std::ostream`, which could be either a file (`std::ofstream`), memory (`std::ostringstream`) or some network stream, derived from `std::ostream` (see [example here](https://stackoverflow.com/a/19933011)).

The bulk of the time for this PR was taken by managing the `depends` package and multiple recompilations of the toolchain.

### DIFF
--- a/contrib/epee/src/file_compression.cpp
+++ b/contrib/epee/src/file_compression.cpp
@@ -1,0 +1,82 @@
+// Copyright (c) 2014-2021, The Monero Project
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+// 
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+// 
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+// 
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "file_compression.h"
+
+#ifdef WIN32
+    #include <boost/iostreams/filter/zlib.hpp>
+    #define COMPRESSOR zlib_compressor()
+    #define DECOMPRESSOR zlib_decompressor()
+    #define EXTENSION ".zip"
+#else
+    #include <boost/iostreams/filter/gzip.hpp>
+    #define COMPRESSOR gzip_compressor()
+    #define DECOMPRESSOR gzip_decompressor()
+    #define EXTENSION ".gz"
+#endif // WIN32
+// Other possibilities to consider with newer Boost:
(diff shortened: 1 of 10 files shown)


iamamyth:
Is this really necessary, when a tool like logrotate can accomplish the same task? In-daemon log rotation can be helpful to avoid having to watch a filehandle for changes, but once the daemon isn't using the log file, it's quite easy to compress or archive or otherwise process the logs.

selsta:
I guess it can be useful on e.g. Windows where you don't have logrotate

mj-xmr:
> I guess it can be useful on e.g. Windows where you don't have logrotate

Exactly. This use case was in fact the inspiration to start this PR. https://github.com/monero-project/monero/issues/7539

hyc:
Sounds like it would make more sense to just port logrotate to Windows. Separation of concerns, attack surface and all that.

mj-xmr:
> Sounds like it would make more sense to just port logrotate to Windows. Separation of concerns, attack surface and all that.

Wouldn't it then make sense to merge this, changing 2 lines of production code to achieve the same, and then, when time permits, just port Logrotate to Windows and set it up for all platforms?

selsta:
This PR adds ~200 LOC (including copyright header and unit test) so IMO it should be good to add, even if other programs could do this too.

It it would have been significantly larger I would have a different opinion.

iamamyth:
> This PR adds ~200 LOC (including copyright header and unit test) so IMO it should be good to add, even if other programs could do this too.

It adds two dependencies (boost iostreams and zlib), as well as the daemon code. All of which, in the best case, provide a lower-bound approximation to any sane log rotation tool. As best as I can tell, there's a possibly working port of [logrotate for windows](https://github.com/plecos/logrotatewin), and, if that doesn't work, I think the desired rotation can be achieved with a two-line batch script.

mj-xmr:
> It adds two dependencies (boost iostreams and zlib), 

Log rotation aside, is adding a multi purpose mature tested-to-death libraries for all supported hosts, and wrapping them in a lightweight API a bad thing?

> as well as the daemon code. 

2 lines for the daemon code.

> All of which, in the best case, provide a lower-bound approximation to any sane log rotation tool. 

That's all what we currently have. I agree with the lower-bound approximation argument, but this is not the only thing, that's lower-bound here.

> As best as I can tell, there's a possibly working port of [logrotate for windows](https://github.com/plecos/logrotatewin)

Speaking of adding dependencies:

> Requirements: .NET Framework v2.0 or better

This opens a can of worms in a whole new dimension.

> I think the desired rotation can be achieved with a two-line batch script.

Which again needs special handling under Windows or forcing the Windows user to install Python for production code. They won't be happy.

mj-xmr:
From IRC session today: yet another empirical reason for integrating `boost::iostreams` and `zlib` into Monero:

> tobtoht: Constructing a transactions takes between 2 and 8 rpc calls, and nearly a megabyte of data.
> tobtoht: The main culprit is /get_output_distribution.bin which uses 750KB.
> tobtoht: Even though this this call returns supposedly compressed data the majority of the response is filled with 0x01 bytes.
> tobtoht: Zipping the binary output yields a 90kb file, so there is definitely room for improvement here.

xiphon:
> From IRC session today: yet another empirical reason for integrating `boost::iostreams` and `zlib` into Monero:

Introducing `zlib` dependency for something that doesn't exist yet and not guaranteed to get implemented in the future, right?
Implementing GZIP support on the Epee HTTP server side (and introducing `zlib` dependency for that) sounds good.

Looks like the discussion is about logs compression (and adding `zlib` dependency for it).

Why do we need to compress monerod logs?

If monerod is spamming the logs, just fix it.

If we are talking about disk space consumption, log rotation should handle this instead and limit the total logs size.

mj-xmr:
> Introducing `zlib` dependency for something that doesn't exist yet and not guaranteed to get implemented in the future, right?

Yes. Obviously I can't implement all possible use cases of a library in a PR, which introduces the library itself. I could possibly get a hint, that the PR is too large to be merged.

> Why do we need to compress monerod logs?
> If monerod is spamming the logs, just fix it.

Sounds easy enough.

> If we are talking about disk space consumption, log rotation should handle this instead and limit the total logs size.

There's no log rotation available under Windows. See https://github.com/monero-project/monero/pull/7610#issuecomment-801640568

contrib/epee/src/file_compression.cpp
```diff
@@ -0,0 +1,82 @@
...
+    #include <boost/iostreams/filter/zlib.hpp>
+    #define COMPRESSOR zlib_compressor()
+    #define DECOMPRESSOR zlib_decompressor()
+    #define EXTENSION ".zip"
+#else 
+    #include <boost/iostreams/filter/gzip.hpp>
+    #define COMPRESSOR gzip_compressor()
+    #define DECOMPRESSOR gzip_decompressor()
+    #define EXTENSION ".gz"
+#endif // WIN32
```
moneromooo-monero, line 41: So this uses a different compression format for windows and others ?

mj-xmr: Yes. For Windows it uses just ZIP, so it's easier for the users to unpack yhe archives with the OS supported functionality (Windows Explorer). 

This concerns files only. For eventual inter OS comms, I will of course have to use GZ everywhere.

vtnerd:
My criticism of the existing log rotation was that it was duplicating work that another process should be doing. I recall @jtgrassie agreeing, but ultimately it was decided to merge. My feelings remain the same here, but I suppose neither is going to kill the project.

jtgrassie:
@vtnerd yes, I agree

mj-xmr:
Resolved CMakeLists conflicts.

selsta:
Closing as there was no support for this change.

PR #7611, opened 2021-03-18
blockchain_import: fix wrong reported block hash on error

moneromooo-monero:


selsta (approved):

src/blockchain_utilities/blockchain_import.cpp
```diff
@@ -192,8 +192,11 @@ int check_flush(cryptonote::core &core, std::vector<block_complete_entry> &block
 
     if(bvc.m_verifivation_failed)
     {
-      MERROR("Block verification failed, id = "
-          << epee::string_tools::pod_to_hex(get_blob_hash(block_entry.block)));
+      cryptonote::block block;
+      if (cryptonote::parse_and_validate_block_from_blob(block_entry.block, block))
+        MERROR("Block verification failed, id = " << cryptonote::get_block_hash(block));
+      else
+        MERROR("Block verification failed, block is unparsable");
```
xiphon, line 199: There is a similar check a few lines above. Am i missing something?

https://github.com/monero-project/monero/blob/6a5b792c976e6d39a17d4d9e386bd53f8e26865e/src/blockchain_utilities/blockchain_import.cpp#L143-L155

moneromooo-monero: Just one for txes too. I think that's the lot.

xiphon: I mean won't it early exit at line 151 if the block is unparsable?

moneromooo-monero: It should. You're asking to remove the error check ?

xiphon: Trying to figure out how to reproduce the issue, i.e. to get `Block verification failed, block is unparsable` error printed.

moneromooo-monero: You probably can't. The issue was printing the wrong hash on a block that was presumably parsable, but bad.

selsta (approved):

PR #7612, opened 2021-03-18
core: speed up print_coinbase_tx_sum

moneromooo-monero:
It only needs to parse the tx headers, not the full tx data

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -2608,7 +2608,7 @@ bool Blockchain::get_split_transactions_blobs(const t_ids_container& txs_ids, t_
 }
 //------------------------------------------------------------------
 template<class t_ids_container, class t_tx_container, class t_missed_container>
-bool Blockchain::get_transactions(const t_ids_container& txs_ids, t_tx_container& txs, t_missed_container& missed_txs) const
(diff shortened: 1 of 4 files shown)


src/cryptonote_core/blockchain.cpp
```diff
@@ -2619,10 +2619,12 @@ bool Blockchain::get_transactions(const t_ids_container& txs_ids, t_tx_container
     try
     {
       cryptonote::blobdata tx;
-      if (m_db->get_tx_blob(tx_hash, tx))
+      bool res = pruned ? m_db->get_pruned_tx_blob(tx_hash, tx) : m_db->get_tx_blob(tx_hash, tx);
+      if (res)
       {
         txs.push_back(transaction());
-        if (!parse_and_validate_tx_from_blob(tx, txs.back()))
+        res = pruned ? parse_and_validate_tx_base_from_blob(tx, txs.back()) : parse_and_validate_tx_from_blob(tx, txs.back());
```
mj-xmr, line 2626: Maybe I need better glasses, but what is the difference between these two `parses_*` ?

mj-xmr, line 2626: Minor, but in my code I'd probably use a separate result boolean for a different call, so that I can declare each of them as const values and can check their statuses independently via either logging or a debugger.

mj-xmr (approved):

selsta (approved):

PR #7615, opened 2021-03-18
Fixed issues found by static analysis

SChernykh:
Copy of #7098 for the release branch:

- rolling_median: tried to free uninitialized pointer in a constructor
- net_node.inl: erase-remove idiom was used incorrectly. remove_if doesn't actually remove elements, see http://cpp.sh/6fcjv
- bulletproofs.cc: call to sizeof() instead of vector.size(), luckily it only impacts performance and not code logic there

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -2919,8 +2919,8 @@ namespace nodetool
     const uint32_t index = stripe - 1;
     CRITICAL_REGION_LOCAL(m_used_stripe_peers_mutex);
     MINFO("adding stripe " << stripe << " peer: " << context.m_remote_address.str());
-    std::remove_if(m_used_stripe_peers[index].begin(), m_used_stripe_peers[index].end(),
-        [&context](const epee::net_utils::network_address &na){ return context.m_remote_address == na; });
+    m_used_stripe_peers[index].erase(std::remove_if(m_used_stripe_peers[index].begin(), m_used_stripe_peers[index].end(),
+        [&context](const epee::net_utils::network_address &na){ return context.m_remote_address == na; }), m_used_stripe_peers[index].end());
     m_used_stripe_peers[index].push_back(context.m_remote_address);
   }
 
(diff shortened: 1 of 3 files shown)


selsta (approved):

src/p2p/net_node.inl
```diff
@@ -2919,8 +2919,8 @@ namespace nodetool
     const uint32_t index = stripe - 1;
     CRITICAL_REGION_LOCAL(m_used_stripe_peers_mutex);
     MINFO("adding stripe " << stripe << " peer: " << context.m_remote_address.str());
-    std::remove_if(m_used_stripe_peers[index].begin(), m_used_stripe_peers[index].end(),
-        [&context](const epee::net_utils::network_address &na){ return context.m_remote_address == na; });
+    m_used_stripe_peers[index].erase(std::remove_if(m_used_stripe_peers[index].begin(), m_used_stripe_peers[index].end(),
+        [&context](const epee::net_utils::network_address &na){ return context.m_remote_address == na; }), m_used_stripe_peers[index].end());
```
mj-xmr, line 2923: The code looks like a valid applucation of the Erase-Remove idiom, but I would prefer, that you stored the result of `m_used_stripe_peers[index]` as a non-const referece `peer` and reuse it, for much better readability.

mj-xmr (changes requested):

src/p2p/net_node.inl
SChernykh: This would be an unrelated change to this bugfix. Besides, indexing an `std::array` gives almost 0 overhead after compiler optimizations.

selsta: Seeing that this is already merged to master no point in changing anymore.

mj-xmr: > This would be an unrelated change to this bugfix

I see it a bit differently. It is a chance to improve readability since you are already touching a given line. Perhaps others would disagree.

> Besides, indexing an std::array gives almost 0 overhead after compiler optimizations.

As I said, it's not about speed, but readability. Even without optimisations the gain w.r.t. the remaining code would be negligible.

Plus, this is not a request to change anymore. Just my hint for future.

Issue #6708, opened 2020-07-15
torsocks: obtaining initial peerlist often takes anywhere from 5 to 20 minutes


tobtoht:
If monerod is started with torsocks and it tries to obtain an initial peerlist by connecting to a seednode that happens to be offline it will incur a two minute timeout (see logs below). After failure, monerod attempts to connect to a different seed node. Quite a few seed nodes appear to be offline and it can often take anywhere from 5 to 20 minutes before monerod is able to reach a seed node that is online. This is an issue for simple/bootstrap mode in the GUI on Tails as the user expects to connect to a remote node right away, instead having to wait an arbitrary length of time without feedback.

```
2020-07-15 00:59:00.981	D Trying to connect to 161.67.132.39:18080, bind_ip = 0.0.0.0
2020-07-15 00:59:23.537	T Checking for idle peers...
2020-07-15 00:59:30.539	T BlockchainLMDB::for_all_txpool_txes
2020-07-15 00:59:30.539	T BlockchainLMDB::block_rtxn_start
2020-07-15 00:59:30.539	T mdb_txn_safe: destructor
2020-07-15 00:59:51.544	T Blockchain::get_current_blockchain_height
2020-07-15 00:59:51.544	T BlockchainLMDB::height
2020-07-15 00:59:51.544	T BlockchainLMDB::block_rtxn_start
2020-07-15 00:59:51.544	T mdb_txn_safe: destructor
2020-07-15 00:59:51.544	D Not checking block rate, offline or syncing
2020-07-15 00:59:53.545	T Checking for idle peers...
2020-07-15 01:00:00.546	T BlockchainLMDB::for_all_txpool_txes
2020-07-15 01:00:00.546	T BlockchainLMDB::block_rtxn_start
2020-07-15 01:00:00.546	T mdb_txn_safe: destructor
2020-07-15 01:00:15.549	T Blockchain::get_current_blockchain_height
2020-07-15 01:00:15.549	T BlockchainLMDB::height
2020-07-15 01:00:15.549	T BlockchainLMDB::block_rtxn_start
2020-07-15 01:00:15.549	T mdb_txn_safe: destructor
2020-07-15 01:00:15.549	T Checking for outgoing syncing peers...
2020-07-15 01:00:15.549	T 0 syncing, 0 synced
2020-07-15 01:00:23.552	T Checking for idle peers...
2020-07-15 01:00:30.553	T BlockchainLMDB::for_all_txpool_txes
[...]

PR #7616, opened 2021-03-18
daemon: add command line arg to set socks proxy

tobtoht:
Patch originally provided by wfaressuissia[m] in #monero-dev, with minor changes.

Tails does not transparently torify traffic, as such torsocks must be used in absence of application-level configuration. This introduces various issues, such as long timeouts (See: https://github.com/monero-project/monero/issues/6708#issuecomment-658899258) and problems closing monerod because torsocks likes to eat sigints. This patch eliminates the need for torsocks by introducing a command line argument for setting the socks proxy. This will also allow the GUI to spawn monerod without using torsocks to fix simple and bootstrap mode on Tails.

### DIFF
--- a/src/daemon/core.h
+++ b/src/daemon/core.h
@@ -32,6 +32,7 @@
 #include "cryptonote_core/cryptonote_core.h"
 #include "cryptonote_protocol/cryptonote_protocol_handler.h"
 #include "misc_log_ex.h"
+#include "daemon/command_line_args.h"
 
 #undef MONERO_DEFAULT_LOG_CATEGORY
 #define MONERO_DEFAULT_LOG_CATEGORY "daemon"
@@ -66,7 +67,14 @@ public:
 #else
     const cryptonote::GetCheckpointsCallback& get_checkpoints = nullptr;
 #endif
-    if (!m_core.init(m_vm_HACK, nullptr, get_checkpoints))
+
+    if (command_line::is_arg_defaulted(vm, daemon_args::arg_proxy) && command_line::get_arg(vm, daemon_args::arg_proxy_allow_dns_leaks)) {
+      MLOG_RED(el::Level::Warning, "--" << daemon_args::arg_proxy_allow_dns_leaks.name << " is enabled, but --"
+        << daemon_args::arg_proxy.name << " is not specified.");
+    }
+
+    const bool allow_dns = command_line::is_arg_defaulted(vm, daemon_args::arg_proxy) || command_line::get_arg(vm, daemon_args::arg_proxy_allow_dns_leaks);
+    if (!m_core.init(m_vm_HACK, nullptr, get_checkpoints, allow_dns))
     {
       throw std::runtime_error("Failed to initialize core");
     }
--- a/src/p2p/net_node.h
+++ b/src/p2p/net_node.h
@@ -259,6 +259,7 @@ namespace nodetool
         m_offline(false),
         is_closing(false),
         m_network_id(),
+        m_enable_dns_seed_nodes(true),
         max_connections(1)
     {}
     virtual ~node_server();
@@ -267,7 +268,7 @@ namespace nodetool
 
     bool run();
     network_zone& add_zone(epee::net_utils::zone zone);
-    bool init(const boost::program_options::variables_map& vm);
+    bool init(const boost::program_options::variables_map& vm, const std::string& proxy = {}, bool proxy_dns_leaks_allowed = {});
     bool deinit();
     bool send_stop_signal();
     uint32_t get_this_peer_port(){return m_listening_port;}
@@ -516,6 +517,7 @@ namespace nodetool
 
     epee::net_utils::ssl_support_t m_ssl_support;
 
+    bool m_enable_dns_seed_nodes;
     bool m_enable_dns_blocklist;
 
     uint32_t max_connections;
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -723,6 +723,12 @@ namespace nodetool
     {
       return get_ip_seed_nodes();
     }
+    if (!m_enable_dns_seed_nodes)
+    {
+      // TODO: a domain can be set through socks, so that the remote side does the lookup for the DNS seed nodes.
+      m_fallback_seed_nodes_added.test_and_set();
+      return get_ip_seed_nodes();
+    }
 
     std::set<std::string> full_addrs;
 
@@ -861,10 +867,21 @@ namespace nodetool
   }
   //-----------------------------------------------------------------------------------
   template<class t_payload_net_handler>
-  bool node_server<t_payload_net_handler>::init(const boost::program_options::variables_map& vm)
+  bool node_server<t_payload_net_handler>::init(const boost::program_options::variables_map& vm, const std::string& proxy, bool proxy_dns_leaks_allowed)
   {
     bool res = handle_command_line(vm);
     CHECK_AND_ASSERT_MES(res, false, "Failed to handle command line");
+    if (proxy.size())
+    {
+      const auto endpoint = net::get_tcp_endpoint(proxy);
+      CHECK_AND_ASSERT_MES(endpoint, false, "Failed to parse proxy: " << proxy << " - " << endpoint.error());
+      network_zone& public_zone = m_network_zones[epee::net_utils::zone::public_];
+      public_zone.m_connect = &socks_connect;
+      public_zone.m_proxy_address = *endpoint;
+      public_zone.m_can_pingback = false;
+      m_enable_dns_seed_nodes &= proxy_dns_leaks_allowed;
+      m_enable_dns_blocklist &= proxy_dns_leaks_allowed;
+    }
 
     if (m_nettype == cryptonote::TESTNET)
     {
--- a/src/rpc/core_rpc_server.h
+++ b/src/rpc/core_rpc_server.h
@@ -91,7 +91,8 @@ namespace cryptonote
         const boost::program_options::variables_map& vm,
         const bool restricted,
         const std::string& port,
-        bool allow_rpc_payment
+        bool allow_rpc_payment,
+        const std::string& proxy = {}
       );
     network_type nettype() const { return m_core.get_nettype(); }
 
@@ -289,6 +290,7 @@ private:
     nodetool::node_server<cryptonote::t_cryptonote_protocol_handler<cryptonote::core> >& m_p2p;
     boost::shared_mutex m_bootstrap_daemon_mutex;
     std::unique_ptr<bootstrap_daemon> m_bootstrap_daemon;
+    std::string m_bootstrap_daemon_proxy;
     bool m_should_use_bootstrap_daemon;
     std::chrono::system_clock::time_point m_bootstrap_height_check_time;
     bool m_was_bootstrap_ever_used;
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -388,6 +388,7 @@ namespace cryptonote
     m_fluffy_blocks_enabled = !get_arg(vm, arg_no_fluffy_blocks);
     m_offline = get_arg(vm, arg_offline);
     m_disable_dns_checkpoints = get_arg(vm, arg_disable_dns_checkpoints);
+
     if (!command_line::is_arg_defaulted(vm, arg_fluffy_blocks))
       MWARNING(arg_fluffy_blocks.name << " is obsolete, it is now default");
 
@@ -460,7 +461,7 @@ namespace cryptonote
     return m_blockchain_storage.get_alternative_blocks_count();
   }
   //-----------------------------------------------------------------------------------------------
-  bool core::init(const boost::program_options::variables_map& vm, const cryptonote::test_options *test_options, const GetCheckpointsCallback& get_checkpoints/* = nullptr */)
+  bool core::init(const boost::program_options::variables_map& vm, const cryptonote::test_options *test_options, const GetCheckpointsCallback& get_checkpoints/* = nullptr */, bool allow_dns)
   {
     start_time = std::time(nullptr);
 
@@ -471,6 +472,7 @@ namespace cryptonote
     }
     bool r = handle_command_line(vm);
     CHECK_AND_ASSERT_MES(r, false, "Failed to handle command line");
+    m_disable_dns_checkpoints |= not allow_dns;
 
     std::string db_sync_mode = command_line::get_arg(vm, cryptonote::arg_db_sync_mode);
     bool db_salvage = command_line::get_arg(vm, cryptonote::arg_db_salvage) != 0;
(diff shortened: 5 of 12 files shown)


xiphon:
i think the option should prevent monerod from making any clearnet connections.

For example, DNS queries will still go over clearnet, leaking the original IP. Same for bootstrap daemon connections.

> This will also allow the GUI to spawn monerod without using torsocks to fix simple and bootstrap mode on Tails.

Not yet. The current implementation won't use specified proxy for bootstrap daemon connections.

selsta:
@xiphon this is updated now

selsta (approved):

selsta:
@vtnerd Any objections here? Otherwise I will add it to the merge queue.

cirocosta:
nice!

for anyone wanting to try the changes, an easy way of doing so on linux is to leverage iptable's logging matching on traffic generated by `monerod`

e.g., one can do that by making `monerod` run under a group that's only going to run `monerod` and nothing else that touches the networking stack, and then using the [`owner matching iptables mod`](https://ipset.netfilter.org/iptables-extensions.man.html), log the traffic  


1. create `monero` group and add your user to it

```
# addgroup --system monero
# adduser `whoami` monero 
```

2. add the iptables rules that will log any outgoing traffic from processes in that group

```
# iptables   -A OUTPUT -m owner --gid-owner monero -j LOG --log-prefix 'monero: '
```

3. clear the kernel messages ring buffer (iptables being a kernel module, will output its logs there) 

```
# dmesg --clear
```

4. switch the current group to `monero`, then run `monerod` under that group with the flags you want to test

```
# newgrp monero
$ monerod \
        --data-dir=/home/ubuntu/monerod-data \
        --enforce-dns-checkpointing \
        --no-igd \
        --non-interactive \
        --p2p-bind-port=58080 \
        --rpc-bind-port=58081 \
        --rpc-restricted-bind-port=58089 \
        --tx-proxy tor,127.0.0.1:9050,10 \
        --proxy 127.0.0.1:9050
```

5. look for the logs under the ring buffer

```
dmesg
``` 

without `--proxy-allow-dns-leaks` we can see that there are only connections to the proxy (127.0.0.1:9050), and with it set, then we see connections to `127.0.0.1:53` (local dns resolver) and subsequently connections to clearnet ips 👍 

---

before we forget, should [docs/ANNONIMITY_NETWORKS.md](https://github.com/monero-project/monero/blob/ed506006d2563c474e52db5a91ba2e20af86f30f/docs/ANONYMITY_NETWORKS.md#outbound-connections) be updated to mention the new flags? `Outbound Connections` section, more specifically.

src/p2p/net_node.h
```diff
@@ -516,6 +516,7 @@ namespace nodetool
 
     epee::net_utils::ssl_support_t m_ssl_support;
 
+    bool m_enable_dns_seed_nodes{true};
```
vtnerd, line 519: For consistency, I'd put this in the initializer list.

src/p2p/net_node.inl
```diff
@@ -723,6 +723,11 @@ namespace nodetool
     {
       return get_ip_seed_nodes();
     }
+    if (not m_enable_dns_seed_nodes)
+    {
+      m_fallback_seed_nodes_added.test_and_set();
+      return get_ip_seed_nodes();
```
vtnerd, line 729: You can set a domain through socks, so that the remote side does the lookup for the DNS seed nodes. This would require some more refactoring of the code though.

Could you add a comment about this here?

src/rpc/core_rpc_server.h
```diff
@@ -289,6 +290,7 @@ namespace cryptonote
     nodetool::node_server<cryptonote::t_cryptonote_protocol_handler<cryptonote::core> >& m_p2p;
     boost::shared_mutex m_bootstrap_daemon_mutex;
     std::unique_ptr<bootstrap_daemon> m_bootstrap_daemon;
+    std::string m_bootstrap_daemon_proxy{};
```
vtnerd, line 293: Same thing here with the inconsistency.

src/daemon/core.h
```diff
@@ -66,7 +67,8 @@ class t_core final
 #else
     const cryptonote::GetCheckpointsCallback& get_checkpoints = nullptr;
 #endif
-    if (!m_core.init(m_vm_HACK, nullptr, get_checkpoints))
+    const bool allow_dns = command_line::is_arg_defaulted(vm, daemon_args::arg_proxy) || command_line::get_arg(vm, daemon_args::arg_proxy_allow_dns_leaks);
```
vtnerd, line 70: Should it be an error if someone sets the `proxy_allow_dns_leaks = false` flag without setting the other flag? It works as-is, but the error reporting helps the user without pointless configurations.

vtnerd (approved):
Can you think of anywhere to put a comment about this mode? There's no way to prevent future mistakes with socket connections in the daemon (afaik), but having this written somewhere at the top of `node_node.inl` perhaps? I guess the reviewers will just have to remain vigilante really.

I will do the dull work of updating `docs/ANONYMITY_NETWORKS.md` unless you want to.

xanoni:
Great PR, I'm sure many people will appreciate it. Assume this could also be used to more easily hide `monerod` behind a VPN or SSH tunnel, for the less paranoid who don't want to sacrifice performance.

**Q1)**
Could someone briefly summarize what the trade-offs are, if one has the choice between using `monerod` over Tor with this new switch (Option A) vs. using `monerod` without switch & without torsocks, but with `--tx-proxy=tor,xxx` (Option B)?

E.g.:
- (A) ensures that the ISP and other people along the network can't detect the presence of a `monerod`
- (B) is more performant and faster, and also puts less stress on the Tor network (less of a problem if a VPN is used)
- (A) increases the likelihood of Sybil attacks on nodes
- ...

**Q2)**
Just to confirm, would this send ALL traffic over Tor, including block downloads, DNS, etc.?

src/daemon/core.h
tobtoht: Added a warning message above (I assume only showing a message is sufficient, since the behavior is unaffected). Let me know if there is a better place to put this, though.

vtnerd:
>  Just to confirm, would this send ALL traffic over Tor, including high volume block downloads?

Everything is sent over the socks proxy, so yes.

>  Could someone briefly summarize what the trade-offs are, if one has the choice between using monerod over Tor with this new switch (Option A) vs. using monerod without switch & without torsocks, but with --tx-proxy=tor,xxx (Option B)?

The option in this PR uses exit nodes, whereas the previous option uses hidden services. Your examples are spot-on for the tradeoffs, and I'm not aware of any other major ones.

The other interesting part of this mode is using a VPN/VPS proxy. You can conceal your home IP address but still have the blockchain verified on a machine you control. This pushes an eclipse attack to the VPN/VPS network from your local ISP, but whether this is a better or worse will depend heavily on the local ISP.

>  How come the opinion on this changed? When I read the docs not too long ago, it seemed that sending everything over Tor was not desirable.

The opinion hasn't changed, defending against eclipse attacks are difficult. Attackers can get exit nodes banned, forcing connections through a handful of nodes. Or with hidden services, many sybils can be created cheaply, and there is nothing analogous to IPv4 subnet zones to "spread" connections.

Unfortunately, how much this matter is hard to quantify. Some people find the tradeoffs worth it. Its unlikely to be the default setup for the GUI, but maybe this will change (i.e. perhaps the sybil case with hidden services is much harder in practice).

xanoni:
> The option in this PR uses exit nodes, whereas the previous option uses hidden services.

Let's make sure I understand this.

**Scenario A:**
Send all traffic over Tor, don't use hidden service. 

```
$ monerod \
[...]
    --proxy=127.0.0.1:9050 \
    --tx-proxy tor,127.0.0.1:9050,10 \
    --tx-proxy i2p,127.0.0.1:4447,10 \
    --p2p-ignore-ipv4=0 \
    --no-igd \
    --p2p-bind-ip=0.0.0.0 \
    --p2p-bind-port=58080 \
    --rpc-bind-ip=0.0.0.0 \
    --rpc-bind-port=58081 \
    --rpc-restricted-bind-port=58089 \
[...]
```

_Questions:_
1. Is it correct that there is no way to achieve inbound connections here, given that without an `.onion` other nodes/wallets have no way of figuring out how/where to reach us?
2. ... or would the `--(p2p|rpc)-bind-ip=0.0.0.0` arguments create leaks by ignoring the proxy?
3. What effect does `--tx-proxy=tor` have here? I would assume you could leave it out? Then where would you specify the `.onion` connection count (",10")?
4. Similarly, I assume the I2P tx-proxy statement would fully bypass `--proxy`?


**Scenario B:**
Send all traffic over Tor, use hidden service. 

```
## only showing the (changed|new) parameters vs. (A)
##########################################################

$ monerod \
[...]
    --anonymous-inbound=<OnionAddr>.onion:313373,127.0.0.1:313373,20
    --anonymous-inbound=<I2PAddr>.b32i2p:313374,127.0.0.1:313374,20
[...]
```

_Question_:
Would inbound connections via `.i2p` / `.onion` now be possible, and they would would fully bypass the `--proxy` setting?


**Scenario C:**
As in B, but also disable external IPv4 / IPv6 listening.

```
## only showing the (changed|new) parameters vs. (A)
##########################################################

$ monerod \
[...]
    --anonymous-inbound=<OnionAddr>.onion:313373,127.0.0.1:313373,20
    --anonymous-inbound=<I2PAddr>.b32i2p:313374,127.0.0.1:313374,20
    --p2p-bind-ip=127.0.0.1 \
    --rpc-bind-ip=127.0.0.1 \
    --p2p-ignore-ipv4=1
    --p2p-use-ipv6=0 \
    --rpc-ignore-ipv4=1
    --rpc-use-ipv6=0 \
[...]
```

_Question_:
In practice, is this the same as (B), given we can't have inbound clearnet connections anyway?


Thank you!

xanoni:
> Or with hidden services, many sybils can be created cheaply, and there is nothing analogous to IPv4 subnet zones to "spread" connections.

**Defense 1:**
Can significant Sybil resistance be gained by listening on both I2P and on Tor? Or is there a realistic way for an attacker to link the I2P and Tor addresses? E.g., does `monerod` broadcast/exhibit some type of unique fingerprint / identifier / key that would allow an attacker to map the addresses?

**Defense 2:**
Besides, would it be possible to configure multiple `.onion` and `.b32.i2p` addresses for the same node, to then listen on all of them in parallel (I know that Tor and I2P are not the constraint, but maybe `monerod` would get confused)? If yes, would there be a way for an attacker to find all your `.onion`'s or all your `.b32.i2p`'s to then Sybil each of them?

vtnerd:
Scenario (A) will send transactions over hidden services but all other traffic (IPv4+6) over VPN/exit node. The bind-ip arguments could be considered a leak if someone could port-scan that external IP address remotely. Anonymous inbound is helpful for routing transactions from other nodes. Inbound IPv4+6 connections only work if the VPN can listen+forward to the external IP, which isn't going to be possible with Tor. In most situations, this will have to be a small VPS running SSH twice - one for outbound socks connections and one for forwarded inbound connections.

Scenario (B) does nothing but allow incoming transactions over those networks. You aren't allowing inbound IPv4+6 connections. Blocks can only be fetched/synced over IPv4+6.

Scenario (C) _may_ be better for privacy if there is no firewall/nat between the host and internet. Otherwise port scanning will reveal an open port.

----

> Can significant Sybil resistance be gained by listening on both I2P and on Tor? Or is there a realistic way for an attacker to link the I2P and Tor addresses? E.g., does monerod broadcast/exhibit some type of unique fingerprint / identifier / key that would allow an attacker to map the addresses?

A node is not supposed to be linkable across "zones", so this should not be the case. So syncing top-of-chain over Tor+I2P might mitigate the sybil/eclipse attack. In the worse case an attacker just floods the network with hidden services though, as aliases are cheaper to create than IPv4 (although IPv6 has some massive subnets). The worse such an attacker can do at the moment is prevent transactions over Tor/I2P from being broadcast. In the block-syncing scenario the attacker controls the perceived "top block".

---

> Besides, would it be possible to configure multiple .onion and .b32.i2p addresses for the same node, to then listen on all of them in parallel (I know that Tor and I2P are not the constraint, but maybe monerod would get confused)? If yes, would there be a way for an attacker to find all your .onion's or all your .b32.i2p's to then Sybil each of them?

It's not possible to listen in "parallel" without significant changes. Only one instance of Tor or I2P is allowed internally, and I don't know how "parallel" instances would work with command-line specified nodes. Currently everything with .i2p gets routed to the I2P instance, etc.

I'm also not certain on the amount of work to change this (ignoring the command line funkiness). The mode would be helpful in cases where spamming the entire network with aliases is unfeasible, unworkable, etc.

xanoni:
> The bind-ip arguments could be considered a leak if someone could port-scan that external IP address remotely.

Thank you for explaining the scenarios. The interaction of these arguments can get very confusing. Maybe worth adding some warning messages if people do seemingly dumb things.

In scenario A, is `--tx-proxy` required to communicate with Tor hidden services, or would the `--proxy` already handle that?

PS: This is out of scope and I think I mentioned it in other tickets already, but I believe several options should be renamed to make clear that they only refer to a certain network (e.g., `in-peers` and `out-peers` only affect clearnet ... for others I still have no idea if they also affect Tor/I2P, such as for `limit-rate-(up|down)` and `hide-my-port`).

vtnerd:
> In scenario A, is --tx-proxy required to communicate with Tor hidden services, or would the --proxy already handle that?

`--tx-proxy` is needed to communicate with hidden services, otherwise transactions are sent through Tor exit nodes.

> PS: This is out of scope and I think I mentioned it in other tickets already, but I believe several options should be renamed to make clear that they only refer to a certain network (e.g., in-peers and out-peers only affect clearnet ... for others I still have no idea if they also affect Tor/I2P, such as for limit-rate-(up|down) and hide-my-port).

`--in-peers` and `--out-peers` apply to all zones independently. The poster from the other issue seemed to think that this was a global limit, but that seemed like a bad idea (as it could prevent connections in zones). It is not possible to set per-zone limits.

`--limit-rate-(up|down)` should be universally applied to all TCP connections (including HTTP server, and p2p connections from every zone). The limit is per-connection.

xanoni:
> `--in-peers` and `--out-peers` apply to all zones independently. The poster from the other issue seemed to think that this was a global limit, but that seemed like a bad idea (as it could prevent connections in zones). It is not possible to set per-zone limits.

Not sure if that's what you're saying, but I found this setting to only affect my clearnet connections. E.g., if I set `out_peers 0` in the RPC console then all my IP connections disappear, however Tor and I2P connections are unaffected. This is the way it's supposed to be, yes? So it should really be called `ip_out_peers` and `ip_in_peers`?


> `--limit-rate-(up|down)` should be universally applied to all TCP connections (including HTTP server, and p2p connections from every zone). The limit is per-connection.

When you say per connection, do you mean per peer? I thought it was a global default across all connections (e.g., peer connections), I was just not sure if it also affected Tor/I2P connections. I didn't digest all of the code, but I thought I had seen that there was a global cap across all connections (or at least IP connections, not sure where Tor/I2P come in).

https://github.com/monero-project/monero/blob/dcba757dd283a3396120f0df90fe746e3ec02292/contrib/epee/src/connection_basic.cpp#L237

https://github.com/monero-project/monero/blob/dcba757dd283a3396120f0df90fe746e3ec02292/contrib/epee/src/network_throttle-detail.cpp#L136

https://github.com/monero-project/monero/blob/dcba757dd283a3396120f0df90fe746e3ec02292/contrib/epee/src/network_throttle-detail.cpp#L274

selsta (approved):

PR #7617, opened 2021-03-20
Hotfix: temporarily disabled SSL certificate storing from PR 7366

mj-xmr:


moneromooo-monero:
"Hotfix" ?

moneromooo-monero:
Is it hard to fix it rather than just paper over it ?

mj-xmr:
> Is it hard to fix it rather than just paper over it ?

We have a failing master, and I think the prio 1 now is to bring it back to life, because it's blocking other PRs. I don't want to mess with other author's sandbox. At least not in such a small time frame.

mj-xmr:
Or to put it in different words, since I'm not the author, my "real fix" could bring even more new risks.

moneromooo-monero:
I wasn't asking you to fix it :) Just that since it's just a test breaking, disabling this outright seems like an overreaction. THen again, I don't really care that much either way.

mj-xmr:
> I wasn't asking you to fix it :) Just that since it's just a test breaking, disabling this outright seems like an overreaction. THen again, I don't really care that much either way.

By breaking the test, you are now unable to test 20 other features, which this test was supposed to test.

selsta:
Should be solved by #7619

mj-xmr:
> Should be solved by #7619

Correct. It solved it.

mj-xmr:
Fixed by https://github.com/monero-project/monero/pull/7619

PR #7618, opened 2021-03-20
python-rpc: add function for new scan_tx RPC

moneromooo-monero:


mj-xmr:
Built and tested locally with a green result for check_missing_rpc_methods.

selsta (approved):

PR #7619, opened 2021-03-21
Fix monerod --rpc-ssl disabled and --non-interactive combo

vtnerd:
When both of those options are set, OpenSSL call fails stopping the daemon from launching.


moneromooo-monero:
I can't see how --non-interactive could interact with this... Can you explain a bit more please ?

vtnerd:
I updated the commit comment. I thought this may have been UB at first, but I tracked down OpenSSL code and its not. The `--rpc-ssl disabled` test case is easy to botch because if the associated files are present it works. The combination of `--testnet` and folder have to match, etc.

selsta (approved):

PR #7620, opened 2021-03-22
cryptonote_protocol_handler: fix race condition

perfect-daemon:


mj-xmr:
Maybe I'm not the right reviewer for this, but what I can already complain is, that there's no description and not a single line of comment for all the new code.

selsta:
@vtnerd would ideally review this, it's the race condition he previously pointed out in the deadlock patch PR.

vtnerd (commented):
Is the primary purpose to move the two writes on `context.m_state` inside the strand? There are still naked (non-atomic) reads, but this is certainly an improvement.

mj-xmr:
Please rebase, in order to grab the latest fixes of the tests. Otherwise you might introduce regressions, shadowed by the test failures.

selsta (approved):

PR #7631, opened 2021-03-24
cmake: Don't default to -march=native on OpenBSD

ndorf:
Under OpenBSD/amd64 6.8, with its boost package, the `-march=native` option causes a compile failure in any file that includes `boost/thread.hpp`. The same is true for `-march=<any other x86-64 platform>`.

Under OpenBSD/arm64 6.8, the `-march=native` flag isn't even supported. This fact is detected, and the flag is already not used by default, so this change has no effect there.

The compilation error can be reproduced as follows:

```
$ echo '#include <boost/thread.hpp>' | c++ -I/usr/local/include -x c++ - -march=westmere
In file included from <stdin>:1:
In file included from /usr/local/include/boost/thread.hpp:17:
In file included from /usr/local/include/boost/thread/once.hpp:26:
In file included from /usr/local/include/boost/thread/pthread/once_atomic.hpp:20:
In file included from /usr/local/include/boost/atomic.hpp:12:
In file included from /usr/local/include/boost/atomic/atomic.hpp:20:
In file included from /usr/local/include/boost/atomic/fences.hpp:21:
In file included from /usr/local/include/boost/atomic/detail/operations.hpp:17:
In file included from /usr/local/include/boost/atomic/detail/operations_lockfree.hpp:21:
In file included from /usr/local/include/boost/atomic/detail/ops_gcc_atomic.hpp:24:
/usr/local/include/boost/atomic/detail/ops_gcc_x86_dcas.hpp:460:20: error: address argument to atomic builtin must be a
      pointer to integer or pointer ('volatile boost::atomics::detail::gcc_dcas_x86_64::storage_type *' (aka 'volatile
      boost::atomics::detail::storage128_t *') invalid)
        expected = __sync_val_compare_and_swap(&storage, old_expected, desired);
                   ^                           ~~~~~~~~
1 error generated.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -501,6 +501,17 @@ endif()
 # Trezor support check
 include(CheckTrezor)
 
+# As of OpenBSD 6.8, -march=<anything> breaks the build
+function(set_default_arch)
+  if (OPENBSD)
+    set(ARCH default)
+  else()
+    set(ARCH native)
+  endif()
+
+  set(ARCH ${ARCH} CACHE STRING "CPU to build for: -march value or 'default' to not pass -march at all")
+endfunction()
+
 if(MSVC)
   add_definitions("/bigobj /MP /W3 /GS- /D_CRT_SECURE_NO_WARNINGS /wd4996 /wd4345 /D_WIN32_WINNT=0x0600 /DWIN32_LEAN_AND_MEAN /DGTEST_HAS_TR1_TUPLE=0 /FIinline_c.h /D__SSE4_1__")
   # set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Dinline=__inline")
@@ -514,7 +525,7 @@ if(MSVC)
 else()
   include(TestCXXAcceptsFlag)
   if (NOT ARCH)
-    set(ARCH native CACHE STRING "CPU to build for: -march value or 'default' to not pass -march at all")
+    set_default_arch()
   endif()
   message(STATUS "Building on ${CMAKE_SYSTEM_PROCESSOR} for ${ARCH}")
   if(ARCH STREQUAL "default")


CMakeLists.txt
```diff
@@ -501,6 +501,17 @@ endif()
...
 
+# As of OpenBSD 6.8, -march=<anything> breaks the build
+function(set_default_arch)
+  if (OPENBSD)
+    set(ARCH default)
+  else()
+    set(ARCH native)
+  endif()
+
+  set(ARCH ${ARCH} CACHE STRING "CPU to build for: -march value or 'default' to not pass -march at all")
```
moneromooo-monero, line 512: Are you intending to set ARCH twice, with different config here ? Or is the first one meant to be a local ?

ndorf: The first one is local to the function, it's why I added the function in the first place.

moneromooo-monero: How do you tell ?

ndorf: It's how cmake works, variables are scoped to the enclosing function or, if not in a function, the current directory. IOW `function` and `add_subdirectory` are the only way to create a new scope. https://cmake.org/cmake/help/latest/command/set.html

selsta: Why is the first set local to the function and the second is not? cmake always confuses me.

ndorf: The second one uses the `CACHE` option to the `set` command and so sets a cache variable, which is a totally different thing altogether. (Yes, cmake is, um, special.) These are meant to be user-settable (e.g. with the `cmake -Dx=y` option), persist across runs, and have project-wide scope. See also https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-variables

selsta (approved):

CMakeLists.txt
moneromooo-monero: Thanks

moneromooo-monero (approved):

PR #7649, opened 2021-04-05
Cleanup TCP shutdown sequence in epee server

vtnerd:
Some cleanup/refactoring to the shutdown sequence in the TCP server. Affects P2P and RPC. The "flushing" behavior of the send queue should be more obvious. Shutdown calls on the socket and timer are also moved "inside" the strand too, to synchronize with read/write callbacks.

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -36,6 +36,7 @@
 #include <boost/uuid/random_generator.hpp>
 #include <boost/chrono.hpp>
 #include <boost/utility/value_init.hpp>
+#include <boost/asio/coroutine.hpp>
 #include <boost/asio/deadline_timer.hpp>
 #include <boost/date_time/posix_time/posix_time.hpp> // TODO
 #include <boost/thread/condition_variable.hpp> // TODO
@@ -105,8 +106,7 @@ PRAGMA_WARNING_DISABLE_VS(4355)
 		m_throttle_speed_in("speed_in", "throttle_speed_in"),
 		m_throttle_speed_out("speed_out", "throttle_speed_out"),
 		m_timer(GET_IO_SERVICE(socket_)),
-		m_local(false),
-		m_ready_to_close(false)
+		m_local(false)
   {
     MDEBUG("test, connection constructor set m_connection_type="<<m_connection_type);
   }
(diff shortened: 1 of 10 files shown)


selsta:
```
2021-04-05T20:14:49.1242790Z In file included from /Users/runner/work/monero/monero/contrib/epee/include/net/abstract_tcp_server2.h:419:
2021-04-05T20:14:49.1246300Z /Users/runner/work/monero/monero/contrib/epee/include/net/abstract_tcp_server2.inl:811:98: error: no member named 'shutdown_timeout' in 'epee::net_utils::connection<(anonymous namespace)::test_protocol_handler>::shared_state'
2021-04-05T20:14:49.1248480Z               self->m_timer.expires_from_now(static_cast<const shared_state&>(self->get_state()).shutdown_timeout());
2021-04-05T20:14:49.1249440Z                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
```

vtnerd:
Yup, forgot a file when committing, updating shortly.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -790,66 +771,108 @@ PRAGMA_WARNING_DISABLE_VS(4355)
...
+      shutdown_routine(shutdown_routine&&) = default;
+      ~shutdown_routine() = default;
+
+      void operator()(const boost::system::error_code = {})
+      {
+        // A "forced" (unclean) shutdown can be called while waiting
+        if (!self || status::cleanup <= self->m_shutdown_status)
+          return;
+
+        const boost::shared_ptr<connection>& self_copy = self; // for lambda
```
moneromooo-monero, line 801: Why not a copy then ?

moneromooo-monero (approved):

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: The compiler is required to do another atomic inc/dec if this line is by copy - it can't be optimized out. Overkill as usual, but achieves the same goal easily.

mj-xmr: It is after all just a const reference (an alias). If the pointer type was to be changed, then the copy could fail even.

vtnerd:
This is no longer relevant due to the changes in the TCP server.

PR #7651, opened 2021-04-06
monero-wallet-cli: improve error message when tx amount is zero

wirrickelliot:


src/simplewallet/simplewallet.cpp
```diff
@@ -607,9 +607,14 @@ void simple_wallet::handle_transfer_exception(const std::exception_ptr &e, bool
...
+    {
+      fail_msg_writer() << sw::tr("destination amount is zero");
+      warn_of_possible_attack = false;
+    }
     catch (const tools::error::zero_destination&)
     {
-      fail_msg_writer() << sw::tr("one of destinations is zero");
+      fail_msg_writer() << sw::tr("transaction has no destination");
       warn_of_possible_attack = false;
     }
```
moneromooo-monero, line 619: You'll want the same in wallet_rpc_server.cpp too.

wirrickelliot: Do you mean that I should just update the error message (i.e. change it to `Transaction has no destination`)? Or do I need to also add a conditional to `validate_transfer()` in the event that the amount is zero?

moneromooo-monero: Add the new exception to handle_exception, the same way you did here.

moneromooo-monero (approved):

PR #7654, opened 2021-04-08
unit_tests: fix boost 1.58 compatibility

perfect-daemon:


selsta (approved):

moneromooo-monero (approved):
1.69 has max_connections marked as deprecated, so it's going to go soon probably. Adding a boost version ifdef now is probably better. 1.69 has both so seems like a good threshold.

PR #7656, opened 2021-04-09
ITS#9496 fix mdb_env_open bug from #8704

hyc:
Broken in 9c6eb75c656363176a55c63c81803feb3cfa614d


hyc:
In this bug, if a process tries to open an env while another process is committing a write txn, it's possible
for the committed txn to get lost/overwritten by the new opener.

PR #7661, opened 2021-04-13
Improve cryptonote (block and tx) binary read performance

vtnerd:
This improves block and transaction blob/binary to C/C++ structure performance by ~60% (in some cases, see below). The changes are pretty basic, but touches 27 files in some of the most critical areas of of code. The read performance is already fairly quick, so this might be overzealous. These functions will show up in profiles at times because of the frequency of their use (basically, everytime a block or transaction needs to be "read"). As an example, responding to a p2p download request improves by a tiny ~1.6% on a Ryzen 3 CPU. 

The bulk of the gains come from dropping a copy of the payload before reading. The remainder comes from eliminating virtual methods calls and some `std::istream` stuff. There is also a difficult to measure benefit with CPU cache usage reduction.

Unpacking 100 blocks (2253800-2253799)
      CPU/Build             |  Microseconds  | Difference
------------------------------|----------------------|---------------
Ryzen 3900x / PR       |            86          |    +0%
i7-7Y75 / PR                |            91          |    +5%
Ryzen 3900x / master |            205         |   +135%
i7-7Y75 / master          |            231         |   +168% (+153% relative)

Unpacking 4,224 transactions from above blocks
      CPU/Build             |  Microseconds  | Difference
------------------------------|----------------------|---------------
Ryzen 3900x / PR       |          7,820       |   +0%
i7-7Y75 / PR                |        10,343       |   +32%
Ryzen 3900x / master |        15,623       |   +100%
i7-7Y75 / master          |         25,481      |   +226% (+146% relative)

One thing to note is that the fanless laptop CPU is now +32% from a Ryzen 3900x, instead of the current +63%. My expectation is that the improvements will have more benefits to smaller cache CPUs (amongst other less sophisticated internals).

### DIFF
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -211,9 +211,7 @@ namespace cryptonote
   //---------------------------------------------------------------
   bool parse_and_validate_tx_from_blob(const blobdata_ref& tx_blob, transaction& tx)
   {
-    std::stringstream ss;
-    ss << tx_blob;
-    binary_archive<false> ba(ss);
+    binary_archive<false> ba{epee::strspan<std::uint8_t>(tx_blob)};
     bool r = ::serialization::serialize(ba, tx);
     CHECK_AND_ASSERT_MES(r, false, "Failed to parse transaction from blob");
     CHECK_AND_ASSERT_MES(expand_transaction_1(tx, false), false, "Failed to expand transaction data");
@@ -224,9 +222,7 @@ namespace cryptonote
   //---------------------------------------------------------------
   bool parse_and_validate_tx_base_from_blob(const blobdata_ref& tx_blob, transaction& tx)
   {
-    std::stringstream ss;
-    ss << tx_blob;
-    binary_archive<false> ba(ss);
+    binary_archive<false> ba{epee::strspan<std::uint8_t>(tx_blob)};
     bool r = tx.serialize_base(ba);
     CHECK_AND_ASSERT_MES(r, false, "Failed to parse transaction from blob");
     CHECK_AND_ASSERT_MES(expand_transaction_1(tx, true), false, "Failed to expand transaction data");
@@ -236,9 +232,7 @@ namespace cryptonote
   //---------------------------------------------------------------
   bool parse_and_validate_tx_prefix_from_blob(const blobdata_ref& tx_blob, transaction_prefix& tx)
   {
-    std::stringstream ss;
-    ss << tx_blob;
-    binary_archive<false> ba(ss);
+    binary_archive<false> ba{epee::strspan<std::uint8_t>(tx_blob)};
     bool r = ::serialization::serialize_noeof(ba, tx);
     CHECK_AND_ASSERT_MES(r, false, "Failed to parse transaction prefix from blob");
     return true;
@@ -246,9 +240,7 @@ namespace cryptonote
   //---------------------------------------------------------------
   bool parse_and_validate_tx_from_blob(const blobdata_ref& tx_blob, transaction& tx, crypto::hash& tx_hash)
   {
-    std::stringstream ss;
-    ss << tx_blob;
-    binary_archive<false> ba(ss);
+    binary_archive<false> ba{epee::strspan<std::uint8_t>(tx_blob)};
     bool r = ::serialization::serialize(ba, tx);
     CHECK_AND_ASSERT_MES(r, false, "Failed to parse transaction from blob");
     CHECK_AND_ASSERT_MES(expand_transaction_1(tx, false), false, "Failed to expand transaction data");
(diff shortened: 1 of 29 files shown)


vtnerd:
Crap, didn't rebase this correctly so the tests are going to bomb. Will push out the correction tomorrow. The above numbers were from a properly rebased codebase, but with unacceptable p2p changes for testing.

vtnerd:
Ok, maybe all green now.

src/serialization/binary_archive.h
```diff
@@ -145,9 +141,11 @@ struct binary_archive<false> : public binary_archive_base<std::istream, false>
...
   void serialize_uvarint(T &v)
   {
-    typedef std::istreambuf_iterator<char> it;
-    if (tools::read_varint(it(stream_), it(), v) < 0)
-      stream_.setstate(std::ios_base::failbit);
+    auto current = bytes_.begin();
+    auto end = bytes_.end();
+    good_ &= (0 <= tools::read_varint(current, end, v));
+    current = std::min(current, bytes_.end());
+    bytes_ = {current, std::size_t(bytes_.end() - current)};
```
xiphon, line 148: Might lead to integer overflow if we assume that `bytes_.end()` might be less than `current` (previous line 147).

xiphon: Nevermind, Missed the `current =` assignment.

xiphon (approved):

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -531,9 +523,7 @@ namespace cryptonote
     if(tx_extra.empty())
       return true;
 
-    std::string extra_str(reinterpret_cast<const char*>(tx_extra.data()), tx_extra.size());
-    std::istringstream iss(extra_str);
-    binary_archive<false> ar(iss);
+    binary_archive<false> ar{epee::to_span(tx_extra)};
```
mj-xmr, line 526: +1 for getting rid or reinterpret_cast.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -543,9 +533,8 @@ namespace cryptonote
       CHECK_AND_NO_ASSERT_MES_L1(r, false, "failed to deserialize extra field. extra = " << string_tools::buff_to_hex_nodelimer(std::string(reinterpret_cast<const char*>(tx_extra.data()), tx_extra.size())));
       tx_extra_fields.push_back(field);
 
-      std::ios_base::iostate state = iss.rdstate();
-      eof = (EOF == iss.peek());
-      iss.clear(state);
+      eof = ar.eof();
+      ar.clear_fail();
```
mj-xmr, line 537: The old version looked unnecessarily complex. I hope the new version works equivalently. Do you know which test verifies this?

src/serialization/binary_archive.h
```diff
@@ -145,9 +141,11 @@ struct binary_archive<false> : public binary_archive_base<std::istream, false>
   template <class T>
   void serialize_uvarint(T &v)
   {
-    typedef std::istreambuf_iterator<char> it;
-    if (tools::read_varint(it(stream_), it(), v) < 0)
-      stream_.setstate(std::ios_base::failbit);
+    auto current = bytes_.begin();
+    auto end = bytes_.end();
```
mj-xmr, line 145: Minor, but could it be:

`const auto end = bytes_.cend();`

and then reused at lines 147 and 148?

src/serialization/binary_archive.h
```diff
@@ -166,26 +164,27 @@ struct binary_archive<false> : public binary_archive_base<std::istream, false>
...
+  bool good_;
   bool varint_bug_backward_compatibility_;
 };
 
 template <>
-struct binary_archive<true> : public binary_archive_base<std::ostream, true>
+struct binary_archive<true> : public binary_archive_base<true>
 {
-  explicit binary_archive(stream_type &s) : base_type(s) { }
+  typedef std::ostream stream_type;
```
mj-xmr, line 180: I see, that the assumption now is, that every stream has to be compatible with `std::ostream` . Is this the case for other types of streams?

src/serialization/serialization.h
```diff
@@ -213,7 +213,7 @@ inline bool do_serialize(Archive &ar, bool &v)
  * \brief self-explanatory
  */
 #define END_SERIALIZE()				\
-  return ar.stream().good();			\
+  return ar.good();				\
```
mj-xmr, line 216: I personally couldn't care less, but since everybody puts remarks on this, there are trailing spaces at the end of the line here and the other instances of  `if (!r || !ar.good()) return false;`

mj-xmr:
It makes perfect sense, that the time taken reduced by 50%, since the same work was basically being done twice.

src/cryptonote_basic/cryptonote_format_utils.cpp
moneromooo-monero: The proof functional tests might test this, at least one type of proof runs two serializations on the same buffer, concatenated.

src/serialization/binary_archive.h
vtnerd: Changed to `cbegin()`/`cend()`.

I would prefer to make line 145 `const auto end = bytes_.cend()`, but the types have to match exactly. And I don't want `end` to be modified by the function call (it doesn't, but another call to `cend()` guarantees the value is not modified).

src/serialization/serialization.h
vtnerd: The `\` character is at the end of the line, so this isn't trailing whitespace.

This was formatted by emacs, and the `\` characters lineup in emacs. But not in vim. Running `git blame` + searching for tabs shows that this portion of the file has been formatted like this since 2014. So this maintains the existing style, however frustrating that may be.

src/serialization/binary_archive.h
vtnerd: I'm not sure what your asking - writing always required a type derived from `std::ostream`. This `typedef` was originally in `binary_archive_base` but has been moved to here.

src/cryptonote_basic/cryptonote_format_utils.cpp
vtnerd: I'm going to add a unit test for this. The code here changes the intent slightly, although it doesn't matter because the only error type is EOF now.

After a refactor these two lines will be just `ar.eof()`. If this triggers an error in the stream type (which it can't for this type), then the reset/clear behavior needs to go in `ar.eof()` instead.

vtnerd: And there'll definitely be a comment in header for `eof()` explaining this, in case an overload for `std::istream` is re-added for some reason (read from unencrypted file would be only use case).

src/serialization/serialization.h
mj-xmr: ok

src/serialization/binary_archive.h
mj-xmr: Understood.

vtnerd:
Rebased and updated after review.

mj-xmr (approved):

selsta (approved):

PR #7670, opened 2021-04-19
workflows: upload source archive artifact

selsta:
Copy pasted from GUI repo

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,3 +154,20 @@ jobs:
 # ARCH="default" (not "native") ensures, that a different execution host can execute binaries compiled elsewhere.
 # BUILD_SHARED_LIBS=ON speeds up the linkage part a bit, reduces size, and is the only place where the dynamic linkage is tested.
 
+  source-archive:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: recursive
+    - name: archive
+      run: |
+        pip install git-archive-all
+        export VERSION="monero-$(git describe)"
+        export OUTPUT="$VERSION.tar"
+        echo "OUTPUT=$OUTPUT" >> $GITHUB_ENV
+        /home/runner/.local/bin/git-archive-all --prefix "$VERSION/" --force-submodules "$OUTPUT"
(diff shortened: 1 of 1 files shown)


moneromooo-monero (approved):

.github/workflows/build.yml
```diff
@@ -154,3 +154,20 @@ jobs:
...
+    - uses: actions/checkout@v1
+      with:
+        submodules: recursive
+    - name: archive
+      run: |
+        pip install git-archive-all
+        export VERSION="monero-$(git describe)"
+        export OUTPUT="$VERSION.tar"
+        echo "OUTPUT=$OUTPUT" >> $GITHUB_ENV
+        /home/runner/.local/bin/git-archive-all --prefix "$VERSION/" --force-submodules "$OUTPUT"
```
mj-xmr, line 169: The archive is a .tar file, stored uncompressed. While downloading it, GitHub zips it on the fly to a .zip, which reduces the size 4 times. Since we run the job on Ubuntu anyway, wouldn't it make sense to store it compressed as a tar.gz file in the first place?

If you agree, then the suggested change would be:

```
export OUTPUT_TAR="$VERSION.tar"
export OUTPUT="$OUTPUT_TAR.gz"
echo "OUTPUT=$OUTPUT" >> $GITHUB_ENV
/home/runner/.local/bin/git-archive-all --prefix "$VERSION/" --force-submodules "$OUTPUT_TAR"
tar -czvf "$OUTPUT" "$OUTPUT_TAR"
```

mj-xmr (changes requested):

.github/workflows/build.yml
selsta: Github always compresses the download so no point in double compressing.

mj-xmr: It will take less space to store it, if anybody cares.

selsta: It does compress it:

https://github.com/actions/upload-artifact/issues/109

We don't pay for storage anyway so I would leave it as is.

mj-xmr: 👌

mj-xmr (approved):

PR #7676, opened 2021-04-21
Compil time: move epee storages/parserse_base_utils.h to .cpp

mj-xmr:
This micro change made it possible to reduce the visibility of the logging class and the `<algorithm>` header down to the implementation file.

The PR is intentionally small, not to abuse the reviewers' time. But as such, there will most probably be no noticeable gains from this PR alone.

I kept the trivial calls as inline, while all longer calls are moved to the corresponding .cpp file.

### DIFF
--- a/contrib/epee/include/storages/parserse_base_utils.h
+++ b/contrib/epee/include/storages/parserse_base_utils.h
@@ -28,13 +28,8 @@
 
 #pragma once 
 
-#include <algorithm>
-#include <boost/utility/string_ref.hpp>
-
-#include "misc_log_ex.h"
-
-#undef MONERO_DEFAULT_LOG_CATEGORY
-#define MONERO_DEFAULT_LOG_CATEGORY "serialization"
+#include <boost/utility/string_ref_fwd.hpp>
+#include <string>
 
 namespace epee 
 {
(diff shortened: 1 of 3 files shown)


contrib/epee/include/storages/parserse_base_utils.h
```diff
@@ -304,25 +133,7 @@ namespace misc_utils
...
-              star_end_string = --it;
-              return;
-            }else 
-              ASSERT_MES_AND_THROW("failed to match word number in json entry: " << std::string(star_end_string, buf_end));
-          }
-        }
-        ASSERT_MES_AND_THROW("failed to match word number in json entry: " << std::string(star_end_string, buf_end));
-      }
+      void match_word2(std::string::const_iterator& star_end_string, std::string::const_iterator buf_end, boost::string_ref& val);
       inline bool match_word(std::string::const_iterator& star_end_string, std::string::const_iterator buf_end, boost::string_ref& val)
```
xiphon, line 137: Why don't move all the remaining code out of the header as well?

mj-xmr: Because:
1) it doesn't hurt, as they impose no additional dependencies in the header list (my PoV),
2) the runtime performance is slightly improved when all that such a trivial call does is entering and redirecting the parameters further (both my and @vtnerd 's PoV).

contrib/epee/include/storages/parserse_base_utils.h
```diff
@@ -28,14 +28,8 @@
 
 #pragma once 
 
-#include <algorithm>
 #include <boost/utility/string_ref.hpp>
```
vtnerd, line 31: `#include <string>` - one of the functions signatures references `std::string::const_iterator` which should require the entire definition for `std::string`.

Also looks like `#include <boost/utility/string_ref_fwd.hpp>` is now possible in this header.

contrib/epee/include/storages/parserse_base_utils.h
vtnerd: This `try`/`catch` block is little different since it has to generate a "landing pad" for the catch, but seems fine to leave this `inline`d. The compiler can still elect to _not_ inline the code.

vtnerd (commented):
Minor suggestions, once a repush is done I'll make sure the lines were not altered during the move.

contrib/epee/include/storages/parserse_base_utils.h
mj-xmr: Both done.

mj-xmr:
> Minor suggestions, once a repush is done I'll make sure the lines were not altered during the move.

Thanks. Ready for a recheck.

mj-xmr:
Rebased.

vtnerd (approved):

PR #7678, opened 2021-04-22
[release-v0.17] backport wallet api changes from master

selsta:
Includes not merged yet #7677

We have multiple open GUI PRs that depend on this and it is not clear when we will branch from master again so I backported them.

### DIFF
--- a/src/wallet/api/transaction_history.cpp
+++ b/src/wallet/api/transaction_history.cpp
@@ -92,6 +92,17 @@ std::vector<TransactionInfo *> TransactionHistoryImpl::getAll() const
     return m_history;
 }
 
+void TransactionHistoryImpl::setTxNote(const std::string &txid, const std::string &note)
+{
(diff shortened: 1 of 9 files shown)


src/wallet/api/transaction_history.cpp
```diff
@@ -92,6 +92,17 @@ std::vector<TransactionInfo *> TransactionHistoryImpl::getAll() const
     return m_history;
 }
 
+void TransactionHistoryImpl::setTxNote(const std::string &txid, const std::string &note)
+{
+    cryptonote::blobdata txid_data;
+    if(!epee::string_tools::parse_hexstr_to_binbuff(txid, txid_data) || txid_data.size() != sizeof(crypto::hash))
+        return;
```
moneromooo-monero, line 99: Should this set m_errorCode (can be in another patch later since this is just backporting though) ?

moneromooo-monero (approved):

PR #7679, opened 2021-04-22
wallet2_base abstraction POC. See uri.cpp UTest.

mj-xmr:
# Results
The example use case, the unit test `uri.cpp` **compiles 40% faster** after subtracting the I/O overhead, and still 35% faster with the I/O overhead included. It could be reduced even further after cleaning up the headers which still had to be included by the new interface.
Lately I have added other use cases as well.

Global improvements:
|Compilation (904 times): |  Compilation (907 times): |  0.33% |
| Parsing (frontend):         1434.1 s |     Parsing (frontend):         1398.8 s   |  -2.46% | 
| Codegen & opts (backend):   1523.3 s | Codegen & opts (backend):   1499.1  | -1.59% |
| 110471 ms: wallet2.h (included 19 times, avg 5814 ms): | 77017 ms: wallet2.h (included 14 times, avg 5501 ms) | -30.28%  | 

# Explanation
The `wallet2` class now inherits from the `wallet2_base` interface, which is an abstract class (with at least one pure virtual method). The interface only includes the absolute minimum, while the rest remains forward-declared. In the end state, most of the methods will be pure virtual, while the static ones may remain in the `wallet2_base.cpp`. The important thing to note here, is that the final API isn't affected.
Currently the `wallet2_base` creates the `wallet2` instance with a static method. This will be moved to a factory, or ideally an abstract factory.

# Review considerations
The changes to the original code are intentionally tiny to avoid confusion. OTOH the provided use case (the UT `uri.cpp`) clearly shows where this is going. It shows, that especially for classes as large as `wallet2`, not all clients will have to include all the class' dependencies in order to use the class itself and that the class (or interface) can be very minimalistic as far as its dependencies are concerned.

# Further work
A lot more can be achieved, once the `wallet2`'s internal structs are extracted from the the class itself, and will thus be forward-declareable, but this is planned for another PR. Also, related to this, I have discovered [this redefinition of such a struct](https://github.com/monero-project/monero/blob/master/src/wallet/message_store.h#L240). Extracting the structs will make the mentioned repetition obsolete and it will make it possible to safely remove the repetition.

# Alternatives
A [PIMPL](https://en.cppreference.com/w/cpp/language/pimpl) pattern could be used as well, although this would make the PR much larger than the inheritance approach, so I though that I'd start with inheritance. I find it easier to grasp. Later on, based on the discussion, we can decide in which direction to go by reusing the classes for both the implementation (`wallet2`) and the interface (`wallet2_base`).

time taken: 8h

### DIFF
--- a/src/gen_multisig/gen_multisig.cpp
+++ b/src/gen_multisig/gen_multisig.cpp
@@ -45,7 +45,10 @@
 #include "common/util.h"
 #include "common/scoped_message_writer.h"
 #include "wallet/wallet_args.h"
-#include "wallet/wallet2.h"
+#include "wallet/wallet2_base.h"
+#include "common/password.h"
+#include "ringct/rctOps.h"
+#include "cryptonote_basic/account.h"
 
 using namespace std;
 using namespace epee;
@@ -85,11 +88,11 @@ static bool generate_multisig(uint32_t threshold, uint32_t total, const std::str
   try
   {
     // create M wallets first
-    std::vector<boost::shared_ptr<tools::wallet2>> wallets(total);
+    std::vector<std::unique_ptr<tools::wallet2_base>> wallets(total);
     for (size_t n = 0; n < total; ++n)
     {
       std::string name = basename + "-" + std::to_string(n + 1);
-      wallets[n].reset(new tools::wallet2(nettype, 1, false));
+      wallets[n] = tools::wallet2_base::create(nettype, 1, false);
       wallets[n]->init("");
       wallets[n]->generate(name, pwd_container->password(), rct::rct2sk(rct::skGen()), false, false, create_address_file);
     }
@@ -100,7 +103,7 @@ static bool generate_multisig(uint32_t threshold, uint32_t total, const std::str
     for (size_t n = 0; n < total; ++n)
     {
       wallets[n]->decrypt_keys(pwd_container->password());
-      if (!tools::wallet2::verify_multisig_info(wallets[n]->get_multisig_info(), sk[n], pk[n]))
+      if (!tools::wallet2_base::verify_multisig_info(wallets[n]->get_multisig_info(), sk[n], pk[n]))
       {
         tools::fail_msg_writer() << genms::tr("Failed to verify multisig info");
         return false;
@@ -135,7 +138,7 @@ static bool generate_multisig(uint32_t threshold, uint32_t total, const std::str
       std::vector<crypto::public_key> signers(total);
       for (size_t n = 0; n < total; ++n)
       {
-        if (!tools::wallet2::verify_extra_multisig_info(extra_info[n], pkeys, signers[n]))
+        if (!tools::wallet2_base::verify_extra_multisig_info(extra_info[n], pkeys, signers[n]))
         {
           tools::fail_msg_writer() << genms::tr("Error verifying multisig extra info");
           return false;
--- a/src/wallet/CMakeLists.txt
+++ b/src/wallet/CMakeLists.txt
@@ -32,6 +32,7 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
 set(wallet_sources
   wallet2.cpp
+  wallet2_base.cpp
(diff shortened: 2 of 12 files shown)


mj-xmr:
Rebased,

vtnerd:
I don't understand how this differs from the effort in `wallet/api/wallet2_api.h`? I guess that provides a high-level interface and this a low-level interface?

I'm against this because it isn't really addressing the problems in wallet2, and might be making things worse for the sake of speeding the compilation of a few files. Much of the serialization code should be movable into a cpp to reduce the headers brought in by `wallet2.h`. That will bring down the compilation time of `wallet2.h` based files considerable without having to maintain this extra redirection file.

Then wallet2 ideally would be broken into subcomponents such that file operations and wallet operations are separated (which should help downstream projects integrate even more than the existing wallet_api). Splitting the two would also result in two smaller more manageable cpp/hpp files.

paulshapiro:
I agree that wallet2 needs to be broken into subcomponents.
More specifically there are implementations in wallet2 that other consumers would need to use and which could be said to not semantically belong in wallet2 alone, and which would benefit from central / shared implementation for multiple consumers, etc..
However I would say that part of the complexity of accomplishing such a factoring for certain parts of wallet2's code, like the transaction creation preamble, does involve the extraction or partial generalization of various types that are currently declared in wallet2 domain.

So I can actually see why this PR might be a step in the right direction. But it does strike me as being in danger of putting the cart before the horse in terms of problem solving. It is possible, though, that we're suffering from wallet2_api PTSD in apprehension towards going in the other direction of creating abstract or base classes that establish the interfaces we need to fix wallet2. I just want to echo the concern that we should be careful to avoid spending such skilled energy in non-critical directions for the moment, or continuously ask ourselves what problem we truly think we're solving in each specific code refactor. For example wallet2/api is IMO able to be seen as ill-advised because, well, wallet2's method signatures, properties, types, etc. already do establish an 'API'. 

Factoring out implementations to better qualified or specific domains would probably also enable additional build time improvements, among other things.

paulshapiro:
And just to add some more background and history - [past work](https://github.com/mymonero/mymonero-core-cpp/blob/master/src/monero_transfer_utils.cpp#L228) by mymonero and my continued currently private refinements and enhancements have already done a lot to factor shared implementations into other domains with reasonable interfaces - and this stuff could just be PR'd straight to Monero (as we will do soon) with wallet2's required modifications only being proxy/passthrough calls to those new functions, similarly to how this PR by mj-xmr is calling into the base class impls. .... What my recent work has been doing is, instead of e.g. `monero_address_utils`, going towards `monero::addresses`.

UkoeHB:
> Then wallet2 ideally would be broken into subcomponents such that file operations and wallet operations are separated (which should help downstream projects integrate even more than the existing wallet_api). Splitting the two would also result in two smaller more manageable cpp/hpp files.

FWIW I believe this is definitely doable. My current refactor of multisig key gen removes 8 function signatures from wallet2  (and adds only 1 for a new feature).

mj-xmr:
> I don't understand how this differs from the effort in `wallet/api/wallet2_api.h`? I guess that provides a high-level interface and this a low-level interface?

Yes. One that can be used as a replacement for the bare `wallet2` class without sacrificing the compilation time.

> 
> I'm against this because it isn't really addressing the problems in wallet2, and might be making things worse for the sake of speeding the compilation of a few files. Much of the serialization code should be movable into a cpp to reduce the headers brought in by `wallet2.h`. That will bring down the compilation time of `wallet2.h` based files considerable without having to maintain this extra redirection file.
> 
> Then wallet2 ideally would be broken into subcomponents such that file operations and wallet operations are separated (which should help downstream projects integrate even more than the existing wallet_api). Splitting the two would also result in two smaller more manageable cpp/hpp files.

Both paragraphs sound like a very good long term plan. My PR helps keeping the compilation time sane, until the above happens. Afterwards, it will be OK to thin down the `wallet2_base`, as much as the `wallet2` will have been thinned, or even completely abandon the `wallet2_base`, if it doesn't make sense to keep it anymore.

selsta:
Closing as this is a POC that is unlikely to be implemented as is.

Issue #7673, opened 2021-04-19
View-only wallet doesn't show outgoing sweep_single transfers


ndorf:
Given an offline (cold) wallet and an associated view-only (hot) wallet, with all outputs and key_images exported & imported, the `show_transfers` command correctly shows outgoing transactions that were made using the `transfer` command, but not those that were made using the `sweep_single` command.

The swept output still shows up as spent in `incoming_transfers`, and the output of `balance` is still correct (doesn't include the amount of the spent/swept output).

ndorf:
Correction: they do show up as expected in `show_transfers`, after the relevant `submit_transfer` was performed in the same wallet. They go missing if/when when the wallet is regenerated from the view key, or the `rescan_bc` command is used.

Here is a stagenet wallet that demonstrates the issue:

Address: 54e7mbbPdb5SbxSkDpzS163msDHKbTtqsB6krkWTUBui3ZV78VocotrCnATu2AJpiaFxdWNJ7M1rBF6H4WCWEK2cP97otsN
View key: c7b0484dd03dbf5ebcbc4af835d5ade6353a0f648ed90597604e04adfba7da0f
Restore height: 817248

The `show_transfers` output is correct in the original, view-only wallet which created and submitted the two outgoing transactions:

```
  817248     in unlocked       2021-04-19 18:43:32       0.990000000000 d4acb7d9d5f32bd3d8a0f1d6afac55f4e14fd132bfb67ec3a428bbbc3593a0e5 0000000000000000 0.000000000000 54e7mb:0.990000000000 0 - 
  817250     in unlocked       2021-04-19 18:45:33       0.810000000000 634ec03d9841813403d46fe8a85a29bcf037aa8e893851b5ad7b0d08a0572d04 0000000000000000 0.000000000000 54e7mb:0.810000000000 0 - 
  817252     in unlocked       2021-04-19 18:56:08       0.630000000000 f01ef29a50c793e1646903acf3539b7cd799a4de27cc3cc68a74e4020e937e75 0000000000000000 0.000000000000 72nyqL:0.630000000000 1 - 
[...]

PR #7680, opened 2021-04-22
[release-0.17] wallet2: Fix rescan_bc keep_ki option

ndorf:
Currently, the `rescan_bc keep_ki` command silently fails. The assertion in `hash_m_transfers()` is incorrectly inverted, and always fails on the first call, when literal `-1` is passed. This exception isn't handled properly, so no message is shown.

This function also incorrectly treated some integer members of `struct transfer_details` as pointers, causing memory errors.

With these fixed, this command seems to provide at least a good workaround for  #7673.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -5800,7 +5800,7 @@ bool simple_wallet::refresh_main(uint64_t start_height, enum ResetType reset, bo
   if (reset != ResetNone)
   {
     if (reset == ResetSoftKeepKI)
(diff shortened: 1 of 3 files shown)


ndorf:
Updated with a further improvement: if one or more of our transfers disappears after a reorg, handle it cleanly, in the same way as if the hashes didn't match. Otherwise the same exception would have been thrown and the wallet might be left in an inconsistent state.

selsta:
Please also PR against master.

moneromooo-monero (approved):

PR #7683, opened 2021-04-25
CMake: strip targets (optional) with -D STRIP_TARGETS=ON

mj-xmr:
`-D STRIP_TARGETS=ON` is also possible for testing the debug build when you're low on ram-disk space.

Results:

| Case | Previous   |      New      | Change |
|----------|----------|-------------|-------------|
| Release-static | 381 MB |  353 MB | -7.35% |
| Release-shared | 225 MB | 213 MB | -5.33% | 
| Debug-shared | 6213 MB | 4582 MB | -26.25% | 

strip

strip-debug

Whether to make stripping the default option for the `release` target, is a separate decision.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,15 @@ function (monero_set_target_no_relink target)
   endif()
 endfunction()
 
+option(STRIP_TARGETS "Strip symbols from targets?" OFF)
+function (monero_set_target_strip target)
+  if (STRIP_TARGETS)
+    set_target_properties("${target}" PROPERTIES LINK_FLAGS_RELEASE -s)
+    set_target_properties("${target}" PROPERTIES LINK_FLAGS_DEBUG -s)
+    # Stripping from Debug might make sense if you're low on disk space, but want to test if debug version builds properly.
+  endif()
+endfunction()
+
 function (monero_add_minimal_executable name)
   source_group("${name}"
     FILES
@@ -161,7 +170,8 @@ function (monero_add_minimal_executable name)
 
   add_executable("${name}"
     ${ARGN})
-    monero_set_target_no_relink( ${name} )
+    monero_set_target_no_relink("${name}")
+    monero_set_target_strip    ("${name}")
 endfunction()
 
 # Finds all headers in a directory and its subdirs, to be able to search for them and autosave in IDEs.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -154,14 +154,27 @@ function (monero_set_target_no_relink target)
   endif()
 endfunction()
 
+option(STRIP_TARGETS_RELEASE "Strip symbols from targets in Release mode?" OFF)
+option(STRIP_TARGETS_DEBUG   "Strip symbols from targets in Debug mode?"   OFF)
+function (monero_set_target_strip target)
+  if (STRIP_TARGETS_RELEASE MATCHES ON)
```
moneromooo-monero, line 160: Does that trigger on ON or also 1, etc ?

mj-xmr: Good catch. It wouldn't.

moneromooo-monero (approved):
I'd have avoided having two different settings but hey, good enough.

mj-xmr:
> I'd have avoided having two different settings but hey, good enough.

I thought the same, but many people could (rightfully) say, that stripping symbols from the Debug build is so pointless, that it should be a separate case.

moneromooo-monero:
Well, then you don't set it. It's optional. And tbh I wanted to say it but tried hard to refrain :D

selsta:
You either build release or debug. Why have two options for the same thing?

mj-xmr:
OK. Done as one variable.

selsta (approved):

mj-xmr:
> You either build release or debug. Why have two options for the same thing? Making it one simplifies the code and you also have to only remember one flag :)

My point was, that with 2 vars, you could have set the release version as strip by default and debug as not strip by default. In order to achieve the same behavior now, you'll have to do many modifications in the main Makefile.

selsta (approved):

hyc:
I vote against this option. It would make stack traces in released binaries useless.

mj-xmr:
> I vote against this option. It would make stack traces in released binaries useless.

Agree, and this is why it's OFF by default. There's even no possibility to switch it on via Makefile, only cia cmake command, so as a purely developer's tool.

selsta (approved):

PR #7684, opened 2021-04-26
blockchain: speed up repeated pops (mostly in tests)

moneromooo-monero:
The heavy rolling median reset only has to be performed after
all blocks are popped


mj-xmr (approved):
It looks ok and the tests pass, but maybe another reviewer could also have a look.

The rolling median is indeed the bottleneck of UTs, as far as my weekly report shows via the Valgrind artifact.

PR #7686, opened 2021-04-26
Tests: mining cpu power app's path fix; Disabling measurements on demand

mj-xmr:
4 changes:
- the path of the app doing cpu power measurements was adapted to enable local tests.
- measurements may be disabled with `MINING_NO_MEASUREMENT` env var. Then timeouts large enough are chosen. The use case are low end machines, where you know, that you're not modifying the mining code at a given time, but want to run other mining tests anyway.
- the mining logging may be disabled with `MINING_SILENT` env var.
- unified the build directory in `functional_tests_rpc.py` based on the existing pattern.


moneromooo-monero (approved):

mj-xmr:
Rebased.

selsta (approved):

PR #7691, opened 2021-04-27
cmake: set 3.5 as minimum version

selsta:
- Current 2.8.7 is ancient and predates "modern" cmake. I'm doubtful monero even compiles with it.
- Setting it to 3.5 allows us to use some newer features and also silences some warnings.
- Ubuntu 16.04 (which seems to be the lowest we are targeting) ships with 3.5
- Clang is now AppleClang on Mac so I changed `STREQUAL` -> `MATCHES `


moneromooo-monero (approved):

mj-xmr (approved):
I verified with:
`egrep -inr --include={CMakeLists.txt,*.cmake} cmake_minimum_required .`
that all the occurrences were addressed,

PR #7692, opened 2021-04-27
cmake: forbid and fix undefined symbols / multiple definitions

### LINKED ISSUE #6666: Device and cryptonote_basic cyclic dependency, debug build fails
Debug compilation (with shared lib compilation) fails due to the: 

```
[ 95%] Linking CXX shared library libdevice.dylib
Undefined symbols for architecture x86_64:
  "cryptonote::get_transaction_prefix_hash(cryptonote::transaction_prefix const&, crypto::hash&)", referenced from:
      hw::core::device_default::get_transaction_prefix_hash(cryptonote::transaction_prefix const&, crypto::hash&) in device_default.cpp.o
ld: symbol(s) not found for architecture x86_64
```

The problem is that the `device` module contains `device_default` and `device_ledger` while those are using `cryptonote_basic` functions. `cryptonote_basic` depends on the `device` module (probably due to the interfaces defined in the module). So `device` cannot depend on the `cryptonote_basic` to fix the compilation error as the cyclic dependency would be created.

A similar problem happened to me several times. It would be maybe better to leave only basic interfaces in the `device` and separate implementations to a new module which could depend on the `cryptonote_basic` and other modules.

What do you think @moneromooo-monero @selsta ?
(I am aware that statical linking is a valid workaround)

### LINKED ISSUE #7053: [Design] Cyclical dependency between libdevice and libcryptonote_basic
In this [issue](https://github.com/monero-project/monero/pull/7051) (where you'll find info how to trigger the problem), I have noticed the following cyclical dependency:

The `libdevice` depends on `libcryptonote_basic through cryptonote::get_transaction_prefix_hash(cryptonote::transaction_prefix const&, crypto::hash&)`, used in `device_default.cpp`, while `libcryptonote_basic` depends on `libdevice`, via the file `crypronote_basic.h`, where the `device.hpp` is included and used.

The remedy would be to feed the `libcryptonote_basic` with the class (or rather interface) `device` from `libdevice`, but extracting any implementations (i.e. `device_default`) outside of `libdevice` into, say, `libdeviceImpl`.

This is not high on my prio list currently, but I would be capable of resolving this one once I have more capacity.

selsta:

This fixes build with shared libraries. Was tested on Linux, Mac and mingw64.

Referencing https://github.com/monero-project/monero/pull/1094

Resolves #6666
Resolves #7053


src/wallet/api/CMakeLists.txt
```diff
@@ -71,6 +71,8 @@ target_link_libraries(wallet_api
     mnemonics
     ${LMDB_LIBRARY}
     ${Boost_CHRONO_LIBRARY}
+    ${Boost_LOCALE_LIBRARY}
```
selsta, line 74: These were missing from wallet_api.

https://github.com/monero-project/monero-gui/pull/3448

Once this is merged these can be removed from the GUI side.

CMakeLists.txt
```diff
@@ -164,6 +164,53 @@ function (monero_add_minimal_executable name)
     monero_set_target_no_relink( ${name} )
 endfunction()
 
+function(forbid_undefined_symbols)
+  add_linker_flag_if_supported(-Wl,--no-undefined TMP)
+  add_linker_flag_if_supported(-Wl,-udefined,error TMP)
```
moneromooo-monero, line 169: maybe a typo ? Not sure what linker this is for.

CMakeLists.txt
```diff
@@ -164,6 +164,53 @@ function (monero_add_minimal_executable name)
     monero_set_target_no_relink( ${name} )
 endfunction()
 
+function(forbid_undefined_symbols)
```
mj-xmr, line 167: I would love to see about 2 lines of documentation what this function does and how, as well as some detailed but shorter explanation of what goes on within the function.

contrib/epee/src/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ file(GLOB EPEE_HEADERS_PUBLIC
 	"${EPEE_INCLUDE_DIR_BASE}/**/*.h*" # Any number of subdirs will be included.
 )
 
-add_library(epee STATIC byte_slice.cpp byte_stream.cpp hex.cpp abstract_http_client.cpp http_auth.cpp mlog.cpp net_helper.cpp net_utils_base.cpp string_tools.cpp
+add_library(epee byte_slice.cpp byte_stream.cpp hex.cpp abstract_http_client.cpp http_auth.cpp mlog.cpp net_helper.cpp net_utils_base.cpp string_tools.cpp
     wipeable_string.cpp levin_base.cpp memwipe.c connection_basic.cpp network_throttle.cpp network_throttle-detail.cpp mlocker.cpp buffer.cpp net_ssl.cpp
```
mj-xmr, line 38: Already addressed here in a smaller, more managable PR : https://github.com/monero-project/monero/pull/7416

CMakeLists.txt
selsta: Thanks, updated. It wasn't caught because `-undefined error` is default on Mac anyway.

CMakeLists.txt
selsta: Added a comment.

mj-xmr (approved):

contrib/epee/src/CMakeLists.txt
mj-xmr: Actually it won't hurt to leave it. Let's just push it through. My PR also changes one thing more.

selsta:
Force pushed to set parent_scope correctly.

mj-xmr (approved):

CMakeLists.txt
```diff
@@ -164,6 +164,52 @@ function (monero_add_minimal_executable name)
...
+    )
+    try_compile(SUCCESS "${TEST_PROJECT}/build" "${TEST_PROJECT}" test
+      CMAKE_FLAGS
+      "-DCMAKE_SHARED_LINKER_FLAGS=${CMAKE_SHARED_LINKER_FLAGS}"
+      "-DCMAKE_MODULE_LINKER_FLAGS=${CMAKE_MODULE_LINKER_FLAGS}"
+      "-DEXPECT_SUCCESS=${EXPECT}"
+    )
+    if (NOT ${SUCCESS} STREQUAL ${EXPECT})
+      message(FATAL_ERROR "Undefined symbols test failure: expect(${EXPECT}), success(${SUCCESS})")
+    endif()
```
xiphon, line 207: The test is unrelated to monero codebase, it is testing particular compiler's code, i wouldn't commit it.

CMakeLists.txt
```diff
@@ -210,6 +210,18 @@ add_library(l3 OBJECT incorrect_source.cpp)
...
+if (MINGW)
+  function(export_all_symbols)
+    unset(TMP)
+    add_linker_flag_if_supported(-Wl,--export-all-symbols TMP)
+    string(APPEND CMAKE_SHARED_LINKER_FLAGS ${TMP})
+    string(APPEND CMAKE_MODULE_LINKER_FLAGS ${TMP})
+    set(CMAKE_SHARED_LINKER_FLAGS ${CMAKE_SHARED_LINKER_FLAGS} PARENT_SCOPE)
+    set(CMAKE_MODULE_LINKER_FLAGS ${CMAKE_MODULE_LINKER_FLAGS} PARENT_SCOPE)
+  endfunction()
+  export_all_symbols()
```
xiphon, line 222: Why do we need a function?

Any benefits comparing to

```cmake
add_linker_flag_if_supported(-Wl,--export-all-symbols EXPORT_ALL_SYMBOLS_FLAG)
string(APPEND CMAKE_SHARED_LINKER_FLAGS ${EXPORT_ALL_SYMBOLS_FLAG})
string(APPEND CMAKE_MODULE_LINKER_FLAGS ${EXPORT_ALL_SYMBOLS_FLAG})
```

or even (yep, will execute the test twice, though, do we really care)?
```cmake
add_linker_flag_if_supported(-Wl,--export-all-symbols CMAKE_SHARED_LINKER_FLAGS)
add_linker_flag_if_supported(-Wl,--export-all-symbols CMAKE_MODULE_LINKER_FLAGS)
```

selsta: @perfect-daemon did it this way due to

```
09:23 <wfaressuissia[m]> I want to keep that code wrapped into function since it will keep related calls together and will not pollute parent scope with temporary variables
09:25 <wfaressuissia[m]> and I prefer to wrap related code into function with verbose name export_all_symbols(...) instead of unwrapped code and single line comment
10:38 <wfaressuissia[m]> also with separate function there is no need to check parent scope for usage of the same temporary variables
```

CMakeLists.txt
selsta: The test was useful during development and also helped us find bugs like #7715. It might also help to find regressions in the future.

I don't particularly care if we keep it in, though I don't think it would hurt.

FWIW we have similar tests in the codebase, e.g. https://github.com/monero-project/monero/blob/master/cmake/FindCcache.cmake#L55 or something like `add_linker_flag_if_supported` also tests the compiler.

selsta:
Resolved conflict.

CMakeLists.txt
mj-xmr: It's positive, that you can copy-paste this working solution outside of Monero and I'm happy for the contribution.

src/cryptonote_basic/CMakeLists.txt
```diff
@@ -36,6 +36,14 @@ if(APPLE)
   endif()
 endif()
 
+add_library(cryptonote_format_utils_basic
```
mj-xmr, line 39: Please use `monero_add_library`

mj-xmr (changes requested):

src/cryptonote_basic/CMakeLists.txt
selsta: done

mj-xmr (approved):

PR #7693, opened 2021-04-27
[release-v0.17] cmake: forbid and fix undefined symbols / multiple definitions

selsta:
See #7692


mj-xmr (approved):

mj-xmr (approved):

mj-xmr (changes requested):
Here I wouldn't be very restrictive, but 2 things bother me:
- there are 3 commits, not 1
- one check didn't pass and a total of only 5 have even been ran.

selsta:
> there are 3 commits, not 1

It's intentional. These patches do separate things.

> one check didn't pass and a total of only 5 have even been ran.

This is the release branch. We don't have more checks here.

mj-xmr:
> > there are 3 commits, not 1
> 
> It's intentional. These patches do separate things. These are not "fixup" patches that should be squashed.
> 
> > one check didn't pass and a total of only 5 have even been ran.
> 
> This is the release branch. We don't have more checks here.

Ok, I didn't know. Then let's wait for the checks to finish.

mj-xmr (approved):

PR #7696, opened 2021-04-28
cmake: set xcode file type

selsta:


mj-xmr:
Were there any symptoms/drawbacks on Mac before? Did the source need to get recompiled every time or sth?

selsta:
It fails to compile unless you manually set the source type to assembler. This is for Xcode (IDE) only, not if compiling via command line

mj-xmr:
> It fails to compile unless you manually set the source type to assembler. This is for Xcode (IDE) only, not if compiling via command line

OK, so it works now.

mj-xmr (approved):

selsta:
> OK, so it works now.

It does not work yet but we are one step closer with this PR.

PR #7698, opened 2021-04-28
cmake: fix non portable code

selsta:

Required for e.g. Xcode or ninja msys2

### DIFF
--- a/src/blocks/CMakeLists.txt
+++ b/src/blocks/CMakeLists.txt
@@ -28,20 +28,32 @@
 
 set(GENERATED_SOURCES "")
 
+set(GENERATOR "${CMAKE_CURRENT_BINARY_DIR}/blocks_generator.cmake")
+file(GENERATE OUTPUT ${GENERATOR} CONTENT [=[
+file(READ "${INPUT_DAT_FILE}" DATA HEX)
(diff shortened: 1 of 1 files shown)


src/blocks/CMakeLists.txt
```diff
@@ -28,20 +28,32 @@
...
+string(REGEX REPLACE "[0-9a-fA-F][0-9a-fA-F]" "0x\\0," DATA "${DATA}")
+file(WRITE "${OUTPUT_C_SOURCE}" "
+#include <stddef.h>
+const unsigned char ${BLOB_NAME}[]={
+  ${DATA}
+};
+const size_t ${BLOB_NAME}_len = sizeof(${BLOB_NAME});
+"
+)
+]=])
```
xiphon, line 43: Why do you need to dynamically generate this `.cmake` file?

selsta: No need for it but one less file in source tree. It does not matter really.

mj-xmr: I would personally vote for a file in source tree. We are mixing syntax of two languages here.

selsta: We will mix two languages anyway.

mj-xmr (approved):
Actually when I look at it again, it appears much cleaner than the previous version.

PR #7704, opened 2021-04-30
CMake: Install and publish stripped binaries for Depends

mj-xmr:
Using CMake's special `install/strip` target, which strips the final binaries while installing them into the installation directory. The directory is pointed by the `CMAKE_INSTALL_PREFIX` variable.

The below changes were measured for the unpacked artifacts, so nearest to the end-user's space consumption.

| Target | Previous   |      Current      | Change |
|--------|----------|-------------|-------------|
|Win64 | 63.7 MB  |  32.9 MB |  -48.35% |
| x86_64 Linux | 55.8 MB | 44.4 MB  |  -20.43%  | 
| Cross-Mac| 	47.3 MB  |  41.6 MB  | -12.05% |

### DIFF
--- a/.github/workflows/depends.yml
+++ b/.github/workflows/depends.yml
@@ -81,5 +81,5 @@ jobs:
       with:
         name: ${{ matrix.toolchain.name }}
         path: |
-          /home/runner/work/monero/monero/build/${{ matrix.toolchain.host }}/release/bin/monero-wallet-cli*
-          /home/runner/work/monero/monero/build/${{ matrix.toolchain.host }}/release/bin/monerod*
+          /home/runner/work/monero/monero/build/${{ matrix.toolchain.host }}/release/install/bin/monero-wallet-cli*
+          /home/runner/work/monero/monero/build/${{ matrix.toolchain.host }}/release/install/bin/monerod*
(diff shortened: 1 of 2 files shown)


selsta:
Hmm, here we will also have the same issue as we discussed with release binaries :/ These CI bins are mostly used if someone has a bug so stripping them will also be suboptimal in case they have to use gdb.

Seeing that Github storage is free anyway I would leave them like it is.

With #7683 it will be easy to strip them anyway if someone wants to do that locally.

mj-xmr:
For the record: it's not about Github storage here, but about showing how much one can spare with the binaries, delivered to the end user.

@selsta also pointed out, that even the released binaries are sometimes used for debugging.
`asymptotical` otoh said, that one could deliver stripped binaries to the end user, but on demand (for debugging) the user could be supplied with the symbols themselves, as a separate package. I confirm from my experience, that such a technique is being practised in the industry.

I'll leave this PR open in draft mode for now, so that everybody can learn, and after

hyc:
IMO, moving debug symbols into a separate package is just added hassle. An extra hoop for users to jump through before we can get meaningful diagnostics from a crash.

Issue #5979, opened 2019-10-11
Option to set external advertised IP


b3rsrk:
### Problem
I would like to run my Monero nodes (at home) behind my edge server (cloud). The nodes should use my home internet connection for outgoing connections and accept connections on the edge server.

My current setup:
![monerod](https://user-images.githubusercontent.com/29061652/66636116-d2190e00-ec10-11e9-8947-175187ba81a3.png)

My current solution is to route the upstream traffic through the VPN as well. But the Edge Server got multiple public IPs, so there is still an issue.

### Proposal
I'd propose an option similar to `p2p-external-port` to set the advertised external IP: `p2p-external-ip`.

### Reference
bitcoind already got an option like this: https://github.com/bitcoin/bitcoin/issues/15
Additional reference: https://monero.stackexchange.com/questions/6600/monerod-advertise-different-external-ip

jbg:
This feature is necessary for running a Monero node in any environment where the IP that outbound traffic will be SNAT'ed to differs from the IP that inbound traffic should connect to.

An example is Kubernetes, where the Service abstraction for incoming connections provides an IP dedicated to the service, whereas all outbound connections from the cluster node (or in the case of private clusters, the whole cluster) are usually NAT'ed to the same IP or one of a pool of IPs. At the moment there doesn't appear to be any way to run a Monero node reliably in this kind of environment, as only outbound connections work.

jbg:
Saving this from #monero here:

````
18:27 < moneromooo> Should be fairly easy to add. If you're looking for a guide for doing it yourself:
18:27 < moneromooo> - Add the same option as --external-p2p-port (if the name's right).
18:28 < moneromooo> - Add a field for an IP in the handshake structure, in one of src/{p2p,protocol}/*defs.h
18:28 < moneromooo> - Write it there when creating your own handshake packet
18:29 < moneromooo> - when you receive a handshake, check whether an IP is set in that packet, and use that IP instead of the incoming IP if so
18:29 < moneromooo> - use that IP to add to the peer list instead of always the incoming IP
18:29 < moneromooo> That should be it.
```

I'm hoping to work on this and send a PR next week.

qertoip:
I would also welcome this option.

tortxoFFoxtrot:
Did this ever happen?

selsta:
@tortxoFFoxtrot doesn't appear implemented yet, otherwise this issue would be closed

PR #7707, opened 2021-05-06
Option to set external advertised IP #5979 

Gingeropolous:
from <KSaunderson> on IRC , who was unable to make the PR due to the spam restrictions.


src/p2p/net_node.inl
```diff
@@ -2289,7 +2304,10 @@ namespace nodetool
     CHECK_AND_ASSERT_MES(address_ok, false,
         "Only IPv4 or IPv6 addresses are supported here");
 
-    const epee::net_utils::network_address na = context.m_remote_address;
+    epee::net_utils::network_address na = context.m_remote_address;
+    if (node_data.my_advertise_network_address.get_type_id() != epee::net_utils::address_type::invalid)  {
+      na = node_data.my_advertise_network_address;
+    }
```
mj-xmr, line 2310: If you write a small function like this (pseudocode):

```
get_preferred_address(context, node_data) {
    epee::net_utils::network_address na = context.m_remote_address;
    if (node_data.my_advertise_network_address.get_type_id() != epee::net_utils::address_type::invalid)  {
      na = node_data.my_advertise_network_address;
    }
   return na;
}
```
then in the line `2292`, the `na` can remain const:
`const epee::net_utils::network_address na = get_preferred_address(context, node_data)`

Also lines `2476` - `2481`  would be reduced to:

`const auto preferred_na = get_preferred_address(context, node_data)`

src/p2p/net_node.inl
```diff
@@ -2523,14 +2547,14 @@ namespace nodetool
...
+      try_ping(arg.node_data, context, [peer_id_l, port_l, context, preferred_na, this]()
       {
-        CHECK_AND_ASSERT_MES((context.m_remote_address.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id() || context.m_remote_address.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id()), void(),
+        CHECK_AND_ASSERT_MES((preferred_na.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id() || preferred_na.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id()), void(),
             "Only IPv4 or IPv6 addresses are supported here");
         //called only(!) if success pinged, update local peerlist
         peerlist_entry pe;
-        const epee::net_utils::network_address na = context.m_remote_address;
-        if (context.m_remote_address.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id())
+        const epee::net_utils::network_address na = preferred_na;
```
mj-xmr, line 2556: Why is the copy needed here?

src/p2p/p2p_protocol_defs.h
```diff
@@ -194,11 +196,13 @@ namespace nodetool
       KV_SERIALIZE_VAL_POD_AS_BLOB(network_id)
       KV_SERIALIZE(peer_id)
       KV_SERIALIZE(my_port)
+      KV_SERIALIZE(my_advertise_network_address)
       KV_SERIALIZE_OPT(rpc_port, (uint16_t)(0))
       KV_SERIALIZE_OPT(rpc_credits_per_hash, (uint32_t)0)
       KV_SERIALIZE_OPT(support_flags, (uint32_t)0)
     END_KV_SERIALIZE_MAP()
   };
+  typedef basic_node_data_base<epee::net_utils::network_address> basic_node_data;
```
mj-xmr, line 205: How about C++11's `using`?
`using basic_node_data = basic_node_data_base<epee::net_utils::network_address>;`

mj-xmr (changes requested):

src/p2p/net_node.inl
```diff
@@ -2297,6 +2297,15 @@ namespace nodetool
     return true;
   }
   //-----------------------------------------------------------------------------------
+  epee::net_utils::network_address get_preferred_address(const epee::net_utils::connection_context_base& context, basic_node_data& node_data)
```
mj-xmr, line 2300: One last note: 
Could the last parameter be made a `const` reference? So:
`epee::net_utils::network_address get_preferred_address(const epee::net_utils::connection_context_base& context, const basic_node_data& node_data)`
Because without it, it appears that this is an output parameter, while it's not. I'm assuming that the `get_type_id()` is a const method.

mj-xmr (changes requested):

src/p2p/net_node.inl
mj-xmr: Once this is done, then it's OK from my side.

mj-xmr:
Ok.
Now please squash and force push, so that you end up with just one commit.
`git rebase -i master`
Then replace all `pick` with `s`, except the first one. Next
`git push -f`

@luigi1111 could you please allow CI to be ran on this one?

mj-xmr:
oops. It appears, that you merged, rather than rebased.

src/p2p/net_node.inl
```diff
@@ -2311,7 +2335,7 @@ namespace nodetool
     }
     network_zone& zone = m_network_zones.at(na.get_zone());
 
-    if(!zone.m_peerlist.is_host_allowed(context.m_remote_address))
+    if(!zone.m_peerlist.is_host_allowed(na))
```
moneromooo-monero, line 2338: I think this might be abusable by making lots of connections to a node, embedding a return IP of a target host, and you get to spam the host, shielding the real origin from the target. Not sure whether that's a viable attack though, since it doesn't amplify anything.

src/p2p/net_node.inl
```diff
@@ -2457,20 +2481,23 @@ namespace nodetool
   template<class t_payload_net_handler>
   int node_server<t_payload_net_handler>::handle_handshake(int command, typename COMMAND_HANDSHAKE::request& arg, typename COMMAND_HANDSHAKE::response& rsp, p2p_connection_context& context)
   {
-    if(arg.node_data.network_id != m_network_id)
+
+    const epee::net_utils::network_address preferred_na = get_preferred_address(context, arg.node_data);  //If client is sending a value...
+
+    if(arg.node_data.network_id != m_network_id)  //TODO: figure out if arg.node_data.network_id should be set from the advertised na instead?
```
moneromooo-monero, line 2487: I do not understand that TODO. network id is independent of this anyway.

translations/CMakeLists.txt
```diff
@@ -73,7 +73,7 @@ string(REPLACE ".ts" ".qm" qm_files "${ts_files}")
 
 add_custom_command(TARGET generate_translations_header
   POST_BUILD
-  COMMAND ./generate_translations_header ${qm_files}
+  COMMAND $<TARGET_FILE:generate_translations_header> ${qm_files}
```
moneromooo-monero, line 76: Unrelated

src/p2p/net_node.inl
```diff
@@ -2157,6 +2159,19 @@ namespace nodetool
       node_data.my_port = m_external_port ? m_external_port : m_listening_port;
     else
       node_data.my_port = 0;
+    if (m_external_na.length() > 0)
```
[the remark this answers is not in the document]
mj-xmr: Since C++11, yes.

src/p2p/net_node.inl
```diff
@@ -2507,7 +2532,7 @@ namespace nodetool
       return 1;
     }
 
-    if(has_too_many_connections(context.m_remote_address))
+    if(has_too_many_connections(preferred_na))
```
vtnerd, line 2535: The check must be from the external address. Otherwise a node could connect multiple times, using this new field as way to bypass this check entirely.

src/p2p/net_node.inl
```diff
@@ -2292,7 +2314,7 @@ namespace nodetool
     CHECK_AND_ASSERT_MES(address_ok, false,
         "Only IPv4 or IPv6 addresses are supported here");
 
-    const epee::net_utils::network_address na = context.m_remote_address;
+    const epee::net_utils::network_address na = get_preferred_address(context, node_data);
```
vtnerd, line 2317: The code below is now busted, the cast to ipv6 can fail if a non IPv4 or IPV6 address is given. See other comment about ensuring the addresses are from the same zone.

src/p2p/net_node.inl
```diff
@@ -2282,6 +2295,15 @@ namespace nodetool
     return true;
   }
   //-----------------------------------------------------------------------------------
+  epee::net_utils::network_address get_preferred_address(const epee::net_utils::connection_context_base& context, const basic_node_data& node_data)
+  {
+    epee::net_utils::network_address na = context.m_remote_address;
+    if (node_data.my_advertise_network_address.get_type_id() != epee::net_utils::address_type::invalid)  {
+      na = node_data.my_advertise_network_address;
```
vtnerd, line 2302: I think the check should be `if (context.m_remote_address.get_type_id() == node_data.my_advertise_network_address.get_type_id())` - otherwise someone could specify an external address from a different zone which should never be allowed.

vtnerd (commented):
What's the use case?

And I think some changes definitely need to be made before this can be merged.

mootinator:
> What's the use case?

I've seen 3 reasons people want this:

1. Node is behind a load balancer.
2. Node running in a docker container.
3. Node running behind an ExpressVPN router (Port forwarding is allowed but your incoming public IP isn't the same as outgoing).

PR #7710, opened 2021-05-08
CI: depends downloads relax retries

mj-xmr:
The download part of `depends` was occasionally failing, so I relaxed the conditions for re-download by increasing:
- the timeout
- the number of retries

~~I also noticed, that the fallback URL was no more, so I found and provided a newer one, plus an alternative one (commented out).~~


selsta:
https://bitcoincore.org/depends-sources does not have to change as it still exists, see e.g.

http://bitcoincore.org/depends-sources/boost_1_64_0.tar.bz2
https://bitcoincore.org/depends-sources/sdks/MacOSX10.11.sdk.tar.gz

FWIW I will work on setting up our own depends mirror.

mj-xmr:
> The other two variables can be changed.

Done.

I understand now why I thought that the original URL was no more: because the error was for the latest `openssl 1.1.1k`, which the `bitcoincore` still doesn't host.

selsta (approved):

PR #7711, opened 2021-05-08
cmake: use CMAKE_CURRENT_LIST_DIR in FindLibUSB

selsta:
If we add monero as a subdirectory in the GUI it will look for `test-libusb-version.c` inside the GUI repo, which we don't want.

### DIFF
--- a/cmake/FindLibUSB.cmake
+++ b/cmake/FindLibUSB.cmake
@@ -134,7 +134,7 @@ if ( LibUSB_FOUND )
 
         try_compile(LibUSB_COMPILE_TEST_PASSED
                 ${CMAKE_BINARY_DIR}
-                "${CMAKE_SOURCE_DIR}/cmake/test-libusb-version.c"
+                "${CMAKE_CURRENT_LIST_DIR}/test-libusb-version.c"
                 CMAKE_FLAGS
                     "-DINCLUDE_DIRECTORIES=${LibUSB_INCLUDE_DIRS}"
                     "-DLINK_DIRECTORIES=${LibUSB_LIBRARIES}"


cmake/FindLibUSB.cmake
```diff
@@ -134,7 +134,7 @@ if ( LibUSB_FOUND )
 
         try_compile(LibUSB_COMPILE_TEST_PASSED
                 ${CMAKE_BINARY_DIR}
-                "${CMAKE_SOURCE_DIR}/cmake/test-libusb-version.c"
+                "${PROJECT_SOURCE_DIR}/cmake/test-libusb-version.c"
```
xiphon, line 137:
```suggestion
                "${CMAKE_CURRENT_SOURCE_DIR}/test-libusb-version.c"
```

selsta:
```
CMake Error at /Users/selsta/dev/monero-gui/build/CMakeFiles/CMakeTmp/CMakeLists.txt:14 (add_executable):
  Cannot find source file:

    /Users/selsta/dev/monero-gui/monero/test-libusb-version.c
```

 "${CMAKE_CURRENT_SOURCE_DIR}/cmake/test-libusb-version.c"` works.

Which one do you prefer? Current PR or ^?

mj-xmr: I'd vote for `CMAKE_CURRENT_SOURCE_DIR/cmake`, since the `PROJECT_SOURCE_DIR` is subject to change.

selsta: subject to change in what way?

mj-xmr: When the "project" becomes something different, then the variable will point to a different directory.
Or do you want to say, that it would work nevertheless?

selsta: ${PROJECT_SOURCE_DIR} is the top level directory of the current project. I don't see how this can become something else.

I think it does not matter which one is used.

xiphon: > Which one do you prefer? Current PR or ^?

Ah, sorry, i actually meant `${CMAKE_CURRENT_LIST_DIR}`.

```suggestion
                "${CMAKE_CURRENT_LIST_DIR}/test-libusb-version.c"
```

selsta: Updated.

xiphon (approved):

xiphon (approved):

mj-xmr (approved):

PR #7712, opened 2021-05-09
epee: include public openssl header in cmake

selsta:
Allows us to remove the openssl dependency from GUI.


mj-xmr (approved):
Once it becomes clearer what is needed by what, it will make sense to create a CMake's `interface` target, pointing to all 3rd party dependencies.

PR #7714, opened 2021-05-11
mnemonics: drop EnglishOld language, has been deprecated for ~7 years

xiphon:


moneromooo-monero:
Yes, please leave it in unless there's a good reason for removing (in which case, tell us).

PR #7716, opened 2021-05-12
CI: minimize boost libs for Ubuntu builds

mj-xmr:
After having all Ubuntu libraries under one variable, now it is possible to minimize the boost libraries, like described in the `README.md`, adding the benefit of having the idea tested.


selsta:
This kinda makes it more complicated.

mj-xmr:
> This kinda makes it more complicated for no apparent benefit.

It's definitely less robust for changes, so a big minus here for sure. The positives of such change may be tested locally as well. There's no urgent need to do it on the CI.

mj-xmr:
No upvotes for the feature, so closing the PR.

PR #7717, opened 2021-05-12
core_tests: speed up generation by reusing derivation and commiting RCT rarer

mj-xmr:
Achieved about 10% speed up for core_tests, by reusing the `crypto::key_derivation` variable, which was previously needlessly regenerated for exactly the same parameters. In order to achieve this, I had to refactor `cryptonote_format_utils.cpp` a bit, so that the dependencies, namely `crypto::key_derivation` and `crypto::public_key`, may be generated independently. The refactoring also allowed to get rid of two duplications.

[core_tests-callgrind-after.out.tar.gz](https://github.com/monero-project/monero/files/6468652/core_tests-callgrind-after.out.tar.gz)
[core_tests-callgrind-before.out.tar.gz](https://github.com/monero-project/monero/files/6468653/core_tests-callgrind-before.out.tar.gz)

### After:
core-tests-after

### Before:
core-tests-before


src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -910,28 +910,80 @@ namespace cryptonote
     return res;
   }
   //---------------------------------------------------------------
-  bool is_out_to_acc(const account_keys& acc, const txout_to_key& out_key, const crypto::public_key& tx_pub_key, const std::vector<crypto::public_key>& additional_tx_pub_keys, size_t output_index)
+  bool is_pk_out_key(const crypto::key_derivation & derivation, const crypto::public_key & pk, const txout_to_key& out_key)
```
vtnerd, line 913: The functions not listed in the header should be put in anonymous namespace or declared as `static`.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -910,28 +910,80 @@ namespace cryptonote
...
+    }
+    // try additional tx pubkeys if available
+    return is_out_to_acc_additional(acc, out_key, additional_tx_pub_keys, output_index);
+  }
+
+  bool generate_key_derivation(const account_keys& acc, const crypto::public_key& tx_pub_key, crypto::key_derivation & derivation)
+  {
+    const bool r = acc.get_device().generate_key_derivation(tx_pub_key, acc.m_view_secret_key, derivation);
+    CHECK_AND_ASSERT_MES(r, false, "Failed to generate key derivation");
+    return r;
```
vtnerd, line 978: This should always `return true` given the above check.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -910,28 +910,80 @@ namespace cryptonote
...
+  }
+
+  bool derive_public_key(const account_keys& acc, const crypto::key_derivation & derivation, size_t output_index, crypto::public_key & pk)
+  {
+    const bool r = acc.get_device().derive_public_key(derivation, output_index, acc.m_account_address.m_spend_public_key, pk);
+    CHECK_AND_ASSERT_MES(r, false, "Failed to derive public key");
+    return r;
```
vtnerd, line 985: Same here.

tests/core_tests/chaingen.cpp
```diff
@@ -448,22 +447,40 @@ bool init_output_indices(map_output_idx_t& outs, std::map<uint64_t, std::vector<
...
+            crypto::key_derivation derivation;
 
             for (size_t j = 0; j < tx.vout.size(); ++j) {
                 const tx_out &out = tx.vout[j];
-
-                output_index oi(out.target, out.amount, boost::get<txin_gen>(*blk.miner_tx.vin.begin()).height, i, j, &blk, vtx[i]);
-                oi.set_rct(tx.version == 2);
-                oi.unlock_time = tx.unlock_time;
-                oi.is_coin_base = i == 0;
+                const txout_to_key & out_key = boost::get<txout_to_key>(out.target);
```
vtnerd, line 459: This can now throw; line 461 was checking to make sure the variant had a `txout_to_key` object.

tests/core_tests/chaingen.cpp
```diff
@@ -448,22 +447,40 @@ bool init_output_indices(map_output_idx_t& outs, std::map<uint64_t, std::vector<
         // TODO: add all other txes
         for (size_t i = 0; i < vtx.size(); i++) {
             const transaction &tx = *vtx[i];
+            const crypto::public_key& tx_pub_key = get_tx_pub_key_from_extra(tx);
+            const std::vector<crypto::public_key>& additional_tx_pub_keys = get_additional_tx_pub_keys_from_extra(tx);
+
+            bool is_key_derived = false;
+            bool is_key_derived_success = false;
+            crypto::key_derivation derivation;
```
vtnerd, line 455: Brace initialize just in case.

vtnerd (changes requested):

src/cryptonote_basic/cryptonote_format_utils.cpp
mj-xmr: Provided that the macro works fine, but can be changed to true, no prob.

tests/core_tests/chaingen.cpp
mj-xmr: Good catch. I'll move it below former 461, as there's not other reason for it to be up here other than some purism.

src/cryptonote_basic/cryptonote_format_utils.cpp
mj-xmr: I prefer static.

src/cryptonote_basic/cryptonote_format_utils.cpp
mj-xmr: Done.

tests/core_tests/chaingen.cpp
mj-xmr: Done.

PR #7706, opened 2021-05-05
cmake: don't add fstack-clash-protection with clang

selsta:
```
clang: warning: argument unused during compilation: '-fstack-clash-protection' [-Wunused-command-line-argument]
```

PR #7718, opened 2021-05-12
cmake: treat warnings as error in compiler flag tests

selsta:
More generic fix for #7706.


CMakeLists.txt
```diff
@@ -106,16 +106,18 @@ function (die msg)
 endfunction ()
 
 function (add_c_flag_if_supported flag var)
-  string(REPLACE "-" "_" supported ${flag}_c)
-  check_c_compiler_flag(${flag} ${supported})
+  set(TMP "-Werror ${flag}")
+  string(REGEX REPLACE "[- ]" "_" supported ${TMP}_c)
```
mj-xmr, line 110: It would be great to have a comment, showing what this is needed for (a use case), since the previous version was much easier to imagine.

selsta: > (if I understand it correctly)

This just adds `-Werror` to the check_c_compiler_flag test. Replacing `-` with `_` is already in the current code.

An example would be here: https://github.com/monero-project/monero/blob/master/CMakeLists.txt#L773-L774

Clang produces the following warning but no error:

```
clang: warning: argument unused during compilation: '-fstack-clash-protection' [-Wunused-command-line-argument]
```

We don't want to add any flags that produce warnings so we add `-Werror`. With this patch the flag does not get added and no warning shows up during compilation, as expected.

mj-xmr: Once you explained it to me it makes sense, but for somebody who just reads the `CMakeLists.txt` file it will still be hard.

I'll then just create an issue that will remind us to document it.

mj-xmr (approved):
Follow up issue: https://github.com/monero-project/monero/issues/7785

PR #7722, opened 2021-05-14
set simple counter to exit after too many loops

### LINKED ISSUE #6192: Cannot Ctrl-C close the wallet when it's "Locked due to inactivity."
As the title says, it does not react to signal without first entering the password.
Is that on purpose? If so what is the reasoning?

I find it mildly frustrating when I have several tabs open with several wallets, don't remember which one it is and hence the password, and just want to close it and be done with it. So I'm curious if there is a good reason to be blocking not just wallet but also access back to the shell?

selsta:
The inactivity lock screen will endlessly prompt for password and ctrl+c does not work to exit the session. I typically resort to ctrl+z and killing the PID or starting another SSH session and killing a container (depending on how I'm running `monero-wallet-cli`).

This should (hopefully) improve UX a tiny bit.

(Opened for @ lalanza808)

Closes #6192

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -6518,10 +6518,18 @@ void simple_wallet::check_for_inactivity_lock(bool user)
           "                ||     ||" << std::endl <<
           "" << std::endl;
     }
+    int loop_counter = 0;
+    const int MAX_UI_LOOPS = 3;
     while (1)
     {
+      if (loop_counter++ >= MAX_UI_LOOPS)
+      {
+          tools::fail_msg_writer() << tr("Error: Too many attempts. Exiting.");
+          m_wallet->store();
+          exit(5);
+      }
       const char *inactivity_msg = user ? "" : tr("Locked due to inactivity.");
-      tools::msg_writer() << inactivity_msg << (inactivity_msg[0] ? " " : "") << tr("The wallet password is required to unlock the console.");
+      tools::msg_writer() << inactivity_msg << (inactivity_msg[0] ? " " : "") << tr("The wallet password is required to unlock the console. ") << (MAX_UI_LOOPS - loop_counter + 1) << " attempts remaining.";
       try
       {
         if (get_and_verify_password())


src/simplewallet/simplewallet.cpp
```diff
@@ -6520,6 +6521,10 @@ void simple_wallet::check_for_inactivity_lock(bool user)
     }
     while (1)
     {
+      if (loop_counter > 3)
```
mj-xmr, line 6524: Is there a reason why it can't be realized via:
`while (loop_counter++ <= 3)` ?

If there is, then it's safer to write:
`if (loop_counter++ > 3)`
and remove line 6536

Ideally, the `3` should be declared as a constant, so finally:

```C++
const int MAX_UI_LOOPS = 3;
while (loop_counter++ <= MAX_UI_LOOPS)
{
 /*... */
}
```

mj-xmr, line 6524: Is there a reason why it can't be realized via:
`while (loop_counter++ <= 3)` ?

If there is, then it's safer to write:
`if (loop_counter++ > 3)`
and remove line 6536

Ideally, the `3` should be declared as a constant, so finally:

```C++
const int MAX_UI_LOOPS = 3;
while (loop_counter++ <= MAX_UI_LOOPS)
{
 /*... */
}
```

mj-xmr (changes requested):

src/simplewallet/simplewallet.cpp
```diff
@@ -6498,6 +6498,7 @@ void simple_wallet::check_for_inactivity_lock(bool user)
 #endif
     tools::clear_screen();
     m_in_command = true;
+    int loop_counter = 0;
```
mj-xmr, line 6501: A small nitpick, but unlike in C, where you're forced to declare the variables at the beginning of the scope, in higher level languages it's possible and expected to declare variables as close as their usage as possible (if it doesn't lead to multiple excess computations that is).

So:

```C++
int loop_counter = 0;
while (/*...*/
```

mj-xmr:
<mj-xmr_> But I'm sure that any type of exception would be better than just exit(3)
<selsta> The wallet should save before closing.
<mj-xmr_> selsta, won't an exception auto-close it?
<mj-xmr_> *auto-save
<selsta> I don't know.
<mj-xmr_> OK. Something to check for sure.

lalanza808:
Working on this in on my fork's PR: https://github.com/lalanza808/monero/pull/1/files

Got almost all the feedback accounted for, though still trying to figure out how to exit cleanly.

moneromooo-monero:
That feels wrong somehow. Not hugely though I guess.

selsta:
#8754

Issue #7720, opened 2021-05-13
Add ability to freeze and thaw outputs using monero-wallet-rpc


woodser:
This issue requests the ability freeze and thaw outputs using monero-wallet-rpc.

The following RPC methods should be supported with `key_image` or `output_index` as inputs:

- `freeze`
- `thaw`
- `frozen`

PR #7729, opened 2021-05-18
support freeze, thaw, and frozen in wallet rpc

woodser:


woodser:
Fixes issue #7720

selsta (changes requested):
bump WALLET_RPC_VERSION_MINOR

woodser:
Done.

moneromooo-monero (approved):

vtnerd (commented):
I missed this unfortunately - @woodser next time use `crypto::key_image` directly in the request object. The `struct_init` pattern ensures the object is always zero initialized, and unpacking into the struct is automatically done by `epee`. Hopefully in the near future will be more efficient too - I hope to change the epee DOM to use slices internally such that there is exactly two copies from kernel to C struct.

PR #7732, opened 2021-05-22
Doc: reference monero-patches project in Compiling doc

mj-xmr:
I reference my [monero-patches](https://github.com/mj-xmr/monero-patches) project in the `docs/COMPILING_DEBUGGING_TESTING.md`, as a temporary solution for integrating the (yet unmerged) improvements mentioned a bit later in the document.

### DIFF
--- a/docs/COMPILING_DEBUGGING_TESTING.md
+++ b/docs/COMPILING_DEBUGGING_TESTING.md
@@ -72,6 +72,8 @@ If everything looks fine, then after setting some breakpoints of your choice, th
 `Debug -> Start/Continue`
 
 ## To be done (and merged):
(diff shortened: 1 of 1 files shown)


selsta:
I would be fine with this but some might be unhappy with linking to external patches that aren't reviewed and can be edited at all times.

mj-xmr:
> I would be fine with this but some might be unhappy with linking to external patches that aren't reviewed and can be edited at all times.

I agree. The external project has a description that addresses it, but still this could change under the feet. Shall I duplicate in a shorter form that text? Like:

> The patches are not yet fully approved by the Monero Maintainers and must not enter the Monero's master branch. They can only serve as temporary solutions.

Or would you like to see other text?

trasherdk:
Maybe a note at the top, in **BOLD** saying something like:
# Only for test or staging networks.

mj-xmr:
> Maybe a note at the top, in **BOLD** saying something like:
> 
> # Only for test or staging networks.

Bold text would be good, but the patches shouldn't enter any official branch. They are like tools, you'd put away to the toolbox after you've done your work 5 times faster, tgan without them.

PR #7735, opened 2021-05-24
Fix boost 1.76.0 compatibility

selsta:
Untested.

Add missing header boost/mpl/contains.hpp
monero-project/monero/issues/7728

patch by @loqs

### DIFF
--- a/contrib/epee/include/storages/portable_storage.h
+++ b/contrib/epee/include/storages/portable_storage.h
@@ -31,6 +31,8 @@
 #include "misc_log_ex.h"
 #include "span.h"
 
+#include <boost/mpl/contains.hpp>
+
 namespace epee
 {
   class byte_slice;


contrib/epee/include/storages/portable_storage.h
```diff
@@ -31,6 +31,8 @@
 #include "misc_log_ex.h"
 #include "span.h"
 
+#include <boost/mpl/contains.hpp>
```
mj-xmr, line 34: It's OK for a fix, but will increase the compilation time, i.e. something that I'll have to handle later on.

mj-xmr (approved):

perfect-daemon (approved):
correct

contrib/epee/include/storages/portable_storage.h
moneromooo-monero: Suggest the better one now.

moneromooo-monero: (unless it's not an easy header swap ?)

mj-xmr: I would like to be able to tell this, but I can't do it with 100% certainty without this framework.

moneromooo-monero (approved):

contrib/epee/include/storages/portable_storage.h
vtnerd: There's `mp11` which is (compile-time) faster but requires a newer minimum boost version.

mj-xmr: Yes. It's better to sacrifice some compilation time for the time being, while allowing a wider range of boost versions.

PR #7739, opened 2021-06-04
wallet_api: move adjust_mixin call within try block

tobtoht:
adjust_mixin() can throw.

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -58,7 +58,6 @@ namespace Monero {
 
 namespace {
     // copy-pasted from simplewallet
-    static const size_t DEFAULT_MIXIN = 6;
     static const int    DEFAULT_REFRESH_INTERVAL_MILLIS = 1000 * 10;
     // limit maximum refresh interval as one minute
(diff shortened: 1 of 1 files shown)


vtnerd (commented):
This changes the order of operations big time. Are there hidden side effects to `m_wallet` or static variables here? A quick glance suggests no, but I think its useful to know why this is being moved downward.

tobtoht:
`adjust_mixin()` can throw if the daemon height cache is expired and the subsequent get_info call fails, hence it should be caught to avoid unexpected behavior.

Specifically, libwalletqt (used by GUI, Feather Wallet) will not emit a signal to indicate that transaction creation has failed if `wallet::createTransaction` throws an exception. Without confirmation that transaction construction has failed, either wallet will wait indefinitely. While this can be worked around downstream, the proposed solution is more elegant.

I spot no side effects. Moreover, `adjust_mixin()` will always return 10 under current consensus rules, regardless of the mixin set in `m_wallet->default_mixin()`, and the `mixin` value passed to the function. Setting either to an allowed value under current consensus rules will only suppress a warning message in `adjust_mixin()` that the requested ring size it too high or too low.

vtnerd (approved):

PR #7748, opened 2021-06-04
wallet_api: getPassword

tobtoht:
This way there is no need to cache it on the GUI side, which can cause nasty bugs if the cache isn't updated.

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -837,6 +837,11 @@ bool WalletImpl::setPassword(const std::string &password)
     return status() == Status_Ok;
 }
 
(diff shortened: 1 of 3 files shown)


selsta (approved):

src/wallet/api/wallet.cpp
```diff
@@ -837,6 +837,11 @@ bool WalletImpl::setPassword(const std::string &password)
     return status() == Status_Ok;
 }
 
+std::string WalletImpl::getPassword() const
```
vtnerd, line 840: By reference?

selsta (approved):

PR #7751, opened 2021-06-11
python-rpc: add missing python api for new freeze/thaw/frozen RPC

moneromooo-monero:


mj-xmr:
Should fix master after https://github.com/monero-project/monero/pull/7729

mj-xmr (approved):

PR #7760, opened 2021-06-22
connection: fix implementation

perfect-daemon:


vtnerd:
Please provide more information, is this for performance increase, code clarity, or? This is going to be a big review for everyone else.

perfect-daemon:
In the worst case It's just an approximation of a fix for all added tests and some other known problems.

vtnerd:
Can you at least tell me (even privately) what bugs this is fixing? Otherwise this is possibly a re-write for clarity sake, bordering on a re-write to meet your favored style of callbacks. 

Granted, I prefer asio coroutines or this style over the current style, but I'm not sure its worth a giant re-write. The review and testing for this is going to take a while because it impacts the p2p code too.

perfect-daemon:
"Can you at least tell me (even privately) what bugs this is fixing?" at least those 3 that are triggered by added tests

perfect-daemon:
Did you try to run added tests from the first commit ?

vtnerd:
> Did you try to run added tests from the first commit ?

Yes, I can read the tests, but was a re-write the best way to fix these issues? Can the code changes be done in piecemeal so that more reviewers can reasonably comment on this code over time? You've provided little auxiliary justification for these changes. The PR is basically asserting that your rewrite of the TCP server code, _all at once_, is worthwhile.

I'm going to assume by your sparse comments that you insist that an at once re-write is necessary - which is likely to delay merging due to less reviewers. My TCP shutdown code is still waiting to be merged, and its much smaller scope.

perfect-daemon:
It fixes everything that is possible without changing public interface of connection.
There are many other problems that will require even bigger changes.
It doesn't even fix all problems within connection since the rest of the code relies no some internal ugliness of connection.

perfect-daemon:
I'm touching more at once since there are interconnected problems that require such big scope.

perfect-daemon:
> I'm going to assume by your sparse comments that you insist that an at once re-write is necessary - which is likely to delay merging due to less reviewers. My TCP shutdown code is still waiting to be merged, and its much smaller scope.
That ssl shutdown patch doesn't fix segfault within ssl handshake and other problems that are located within connection.

contrib/epee/include/net/abstract_tcp_server2.h
```diff
@@ -86,6 +90,180 @@ namespace net_utils
...
+
+      struct socket_t {
+        bool connected;
+
+        bool wait_handshake;
+        bool cancel_handshake;
+
+        bool wait_read;
+        bool handle_read;
+        bool cancel_read;
```
vtnerd, line 182: All of these `cancel_` booleans are unnecessary; the standard technique is checking `ec == boost::asio::error::operation_aborted` (nearly every example uses this technique). This method is preferable because it catches all cancellations, even ones added in the future (which are currently only caught if the flag is set to true).

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1019 @@ namespace net_utils
...
+      std::seed_seq seed(rand, rand + engine::state_size);
+      engine rng(seed);
+      return std::chrono::milliseconds(
+        std::uniform_int_distribution<>(5000, 6000)(rng)
+      );
+    };
+    if (state.data.write.wait_consume ||
+      state.data.write.queue.size() > ABSTRACT_SERVER_SEND_QUE_MAX_COUNT
+    )
+      state.condition.wait_for(state.lock, random_delay(),
```
vtnerd, line 762: This `wait_for` is going to need some justification, in the comments, and with some performance testing. Or just any explanation at all. Because this is slowing down all outgoing writes by an arbitrary amount of time.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1019 @@ namespace net_utils
...
+    ) {
+      state.data.write.queue.emplace_front(std::move(message));
+      start_write();
+      return true;
     }
-    
+    state.data.write.wait_consume = true;
+    while (!message.empty()) {
+      if (state.data.write.queue.size() > ABSTRACT_SERVER_SEND_QUE_MAX_COUNT) {
+        state.condition.wait_for(state.lock, random_delay(),
```
vtnerd, line 787: This is still a DoS case - the thread is still blocked.This should be changed to a drop (this is one of the reasons why I am pushing for a piecemeal approach ...).

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1019 @@ namespace net_utils
...
+      if (duration > duration_t{}) {
+        ec_t ec;
+        timers.throttle.out.expires_from_now(duration, ec);
+        state.timers.throttle.out.wait_expire = true;
+        timers.throttle.out.async_wait([this, self](const ec_t &ec){
+          lock_guard_t guard(state.lock);
+          state.timers.throttle.out.wait_expire = false;
+          if (state.timers.throttle.out.cancel_expire) {
+            state.timers.throttle.out.cancel_expire = false;
+            if (state.status == status_t::RUNNING)
```
vtnerd, line 462: `switch`

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1019 @@ namespace net_utils
...
-
-    reset_timer(boost::posix_time::milliseconds(m_local ? NEW_CONNECTION_TIMEOUT_LOCAL : NEW_CONNECTION_TIMEOUT_REMOTE), false);
+    state.socket.wait_write = true;
+    auto on_write = [this, self](const ec_t &ec, size_t bytes_transferred){
+      lock_guard_t guard(state.lock);
+      state.socket.wait_write = false;
+      if (state.socket.cancel_write) {
+        state.socket.cancel_write = false;
+        state.data.write.queue.clear();
+        if (state.status == status_t::RUNNING)
```
vtnerd, line 484: `switch`

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1019 @@ namespace net_utils
...
+    if (state.socket.wait_shutdown)
+      return;
+    auto self = connection<T>::shared_from_this();
+    state.socket.wait_shutdown = true;
+    auto on_shutdown = [this, self](const ec_t &ec){
+      lock_guard_t guard(state.lock);
+      state.socket.wait_shutdown = false;
+      if (state.socket.cancel_shutdown) {
+        state.socket.cancel_shutdown = false;
+        if (state.status == status_t::RUNNING)
```
vtnerd, line 559: `switch`

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1019 @@ namespace net_utils
...
+  }
 
-    if (remote_ep.address().is_v4())
-    {
-      const unsigned long ip_ = boost::asio::detail::socket_ops::host_to_network_long(remote_ep.address().to_v4().to_ulong());
-      return start(is_income, is_multithreaded, ipv4_network_address{uint32_t(ip_), remote_ep.port()});
+  template<typename T>
+  void connection<T>::start_read()
+  {
+    if (state.timers.throttle.in.wait_expire || state.socket.wait_read ||
```
vtnerd, line 301: `wait_read` is only necessary because the callbacks invoke `start_read` ... again. This is unnecessary if the callback were a local struct that contains only the inner looping read calls. I think this is preferable because it separates the start and looping code, whereas this is mixing the two. It also allows for less state machines.

vtnerd (commented):
> That ssl shutdown patch doesn't fix segfault within ssl handshake and other problems that are located within connection.

It was supposed to address one issue, not claim to fix an unspecified amount of things in a rewrite. FWIW, I think this is going to take multiple reviews to hammer out because its likely just swapping one set of issues for another.

As an example, every outgoing write now appears to have a randomized `wait_for` call which should have a measurable performance hit. And the following `wait_for` call is still a DoS as the thread is blocked (but not busy spin looping at least). So its worse performance for ... ?

This was a light skim through, not really a review.

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: Wait this was inside an `if` statement? Ugh, sorry, can we fix the style of this at least?

vtnerd: Ok I see why this was duplicated, but there has to be a simpler more concise technique. And seriously `{}` this!

contrib/epee/include/net/abstract_tcp_server2.inl
perfect-daemon: Both connection and sender should be asynchronous to remove this problem.
It's partially solved now, as much as possible without changing the rest of the code.

perfect-daemon: This queue will be removed from connection later, once everything else is asynchronous.

contrib/epee/include/net/abstract_tcp_server2.inl
perfect-daemon: It will be needed later when everything is asynchronous since reader may not consume every received chunk immediately and it's possible that there is no active callbacks at all.

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: You can't remove a queue, or at least you shouldn't. Because `post`ing to a strand is still a queue.

AFAIK there are 3 choices: (1) close connection (drop data), (2) block a thread to wait, or (3) unbounded queue. I think this needs to transition to (1) if an entire re-write is being done.

perfect-daemon: I can, will and it's needed. But it isn't easy.

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: The flag is only needed because `start_read` is both starting and continuing the reads. It is possible to separate the two, removing the need for the flag. After every async read (and subsequent processing), the continue function is called.

contrib/epee/include/net/abstract_tcp_server2.inl
perfect-daemon: `post` and any other forward will be replaced with asynchronous generator

vtnerd: Asynchronous generator? You mean asynchronously generate bytes piecemeal? That's just a more complicated outgoing buffer. This has to be a drop or block.

perfect-daemon: What are the sources of data for outgoing queue ?
1. Synchronous forward of notifications with txs/blocks which can be made asynchronous with fetching from local blockchain storage.
2. Requests/responses required by explicit synchronization which can be generated asynchronously as soon as outgoing/incoming fixed size buffer is free.

In general, daemon sends and receives only verified data that is stored/loaded to/from blockchain db.
This is the only available unbounded buffer to store data.

contrib/epee/include/net/abstract_tcp_server2.inl
perfect-daemon: I didn't try to optimize number of states / transitions.
It can be done later unless it's needed for significant performance increase or it's related to correctness of the code.

vtnerd: This is just my personal opinion obviously, but -

if you want to do a re-write of the entire connection code, then it should be "optimized" with the fewest number of states / transitions possible. Otherwise, what is the purpose of the re-write? I'm being critical because if this were done piecemeal, the code could've looked similar in the end to this PR, but the unnecessary code/states could be removed along the way easier during review. Removing the unnecessary bits makes it easier to maintain and debug.

As an example, the `cancel_` state flags have to be removed, unless there is some justification why the standard way of checking for cancellation is inappropriate (none was provided above in the other comment thread). And this `wait_read`  (that started this thread) doesn't appear to have any justification (yet), given how ASIO callbacks are typically done.

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: I already know exactly what you are proposing, and I am pointing out that some p2p messages require a response, and a queue of asynchronous events is still a queue. The commentary by the ZeroMQ developers on Pub/Sub seems appropriate here - if the read side is behind eventually you have to drop data and/or the connection.

All of these changes don't really change the behavior of the existing problem. Multiple threads can get stuck busy waiting for a queue to clear. The previous method had the threads waiting indefinitely at a lock, whereas here they wait indefinitely in conditional variables.

perfect-daemon: > This is still a DoS case - the thread is still blocked.This should be changed to a drop (this is one of the reasons why I am pushing for a piecemeal approach ...).

I didn't understand that you were talking about any action after failed conditional sleep.
I've added missing `terminate(); return false`.

contrib/epee/include/net/abstract_tcp_server2.inl
perfect-daemon: Updated

contrib/epee/include/net/abstract_tcp_server2.inl
perfect-daemon: > Otherwise, what is the purpose of the re-write?
>
These explicit flags are needed to not depend on internal behaviour of boost asio which doesn't allow to cancel already dispatched callback.

perfect-daemon: > ... Removing the unnecessary bits makes it easier to maintain and debug.
>
It depends.

perfect-daemon: >if you want to do a re-write of the entire connection code, then it should be "optimized" with the fewest number of states / transitions possible. Otherwise, what is the purpose of the re-write? 
>
The main goal is to fix known problems and prepare code for design changes, which can't be done in case of greedy early optimizations of every component.

vtnerd: > These explicit flags are needed to not depend on internal behaviour of boost asio which doesn't allow to cancel already dispatched callback.

I've thought about this case before, and typically its not worth stressing about. In the read/write callbacks either reset the timer immediately or call `cancel` immediately. The timer will be in the strand queue with the appropriate `error_code`.

In the off chance that it queued the timer just after the other event but before the cancel could occur  ... the timer was so close to executing its not worth the extra state variable and logic. Let ASIO handle all of the state tracking because its better "defensive coding" - the `error_code` cancel check handles every future update to how the timer is used.

> It depends.

Just chop out these unnecessary states. There's a giant `if` ladder than get be chopped down for instance.

> The main goal is to fix known problems and prepare code for design changes, which can't be done in case of greedy early optimizations of every component.

I dislike your usage of the word optimization in this context because it makes it sound like I am asking for a performance comparison. This is about making the code more concise+legible - particularly the `start_read` function.

perfect-daemon: "concise - short and clear, expressing what needs to be said without unnecessary words"
"optimization - the process of making something as good or effective as possible"
Non-programming language is ambiguous enough to allow misinterpretation of anything so If current code is correct then any further changes can be treated as optimization.

contrib/epee/include/net/abstract_tcp_server2.h
```diff
@@ -86,6 +90,181 @@ namespace net_utils
     public i_service_endpoint,
     public connection_basic
   {
+  private:
+    using string_t = std::string;
```
vtnerd, line 94: This "alias everything" style breaks from existing coding conventions, and it would preferable to revert to something closer to the existing style.

contrib/epee/include/net/abstract_tcp_server2.h
vtnerd: Repeating this again since it got lost in the back-and-forth, but the `cancel_` booleans are not necessary. Requesting their removal is not "an optimization request" - if you are going to re-write the entirety of the connection code then don't leave new things for others to remove and deal with.

contrib/epee/src/net_ssl.cpp
```diff
@@ -522,30 +522,101 @@ bool ssl_options_t::handshake(
...
+}
+
+bool ssl_options_t::handshake(
+  boost::asio::ssl::stream<boost::asio::ip::tcp::socket> &socket,
+  boost::asio::ssl::stream_base::handshake_type type,
+  boost::asio::const_buffer buffer,
+  const std::string& host,
+  std::chrono::milliseconds timeout) const
+{
+  configure(socket, type, host);
```
vtnerd, line 534: There's no reason to include these SSL changes with the connection re-write. Include these changes in a new PR with an explanation as to why its being changed.

And if the issue being fixed is the racy `ec = ...`, this can be fixed with less changes. All that was needed is an `std::atomic_flag` to signal the change (a safe "happens before/after synchronization), these other changes are overkill.

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: You're trying to avoid making changes, which is frustrating because this PR contains more code than necessary, leaving more work for reviewers. I will try again in a new review to point out some things.

mrx23dot:
Please do a release so we can do in depth testing.

j-berman (commented):
Working through tests 2 and 3 still. Question on test 1...

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
+      else if (ec.value())
+        terminate();
+      else {
+        cancel_timer();
+        on_interrupted();
+      }
+    };
+    strand.post(
+      [this, self, on_shutdown]{
+        connection_basic::socket_.async_shutdown(
```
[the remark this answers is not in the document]
j-berman: The PR author's reasoning to not invoke `socket_.async_shutdown` when terminating a connection seems to be two-fold:

1) [the risk of a truncation attack is not practically significant](https://libera.monerologs.net/monero-research-lab/20220527#c99767-c99768)
2) [using `async_shutdown` on connection close would require a timeout in the event of an unresponsive client, which arguably opens a wider & more significant surface for abuse than a truncation attack](https://libera.monerologs.net/monero-research-lab/20220529#c100874-c100875)

Seems reasonable to me.

contrib/epee/include/net/abstract_tcp_server2.h
SChernykh: +1 on this. Things like `std::string` or `std::mutex` shouldn't be renamed at all to avoid confusion. Long `epee` and `boost` names can probably be renamed.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
-		t_connection_type connection_type,
-		ssl_support_t ssl_support
-	)
-	: connection(boost::asio::ip::tcp::socket{io_service}, std::move(state), connection_type, ssl_support)
+  template<typename T>
+  unsigned int connection<T>::host_count(int delta)
   {
+    static lock_t hosts_mutex;
+    lock_guard_t guard(hosts_mutex);
+    static std::map<string_t, unsigned int> hosts;
```
SChernykh, line 89: hosts_mutex and hosts should be moved to `connection` class as static members to avoid generating initialization code inside the function. Also, this function doesn't remove an entry from `hosts` when `val` is 0, so `hosts` can grow indefinitely and consume all memory if there is an attacker exploiting this.

SChernykh: Also, if `delta` is 0, it should have code like this
```
if (!delta) {
    auto it = hosts.find(host);
    return (it != hosts.end()) ? it->second : 0;
}
```
to avoid using `operator[]` which modifies the map and is slower.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
-  //---------------------------------------------------------------------------------
-  template<class t_protocol_handler>
-  boost::asio::io_service& connection<t_protocol_handler>::get_io_service()
+  template<typename T>
+  void connection<T>::async_wait_timer()
   {
-    return GET_IO_SERVICE(socket());
+    if (state.timers.general.wait_expire)
+      return;
+    state.timers.general.wait_expire = true;
```
SChernykh, line 161: This sequence is not thread-safe. Should it be protected by `state.lock`? Maybe making `wait_expire` atomic and
```
    if (state.timers.general.wait_expire.exchange(true))
        return;
```
will be enough?

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
-  template<class t_protocol_handler>
-  void connection<t_protocol_handler>::call_back_starter()
+  template<typename T>
+  void connection<T>::cancel_timer()
   {
-    TRY_ENTRY();
-    _dbg2("[" << print_connection_context_short(context) << "] fired_callback");
-    m_protocol_handler.handle_qued_callback();
-    CATCH_ENTRY_L0("connection<t_protocol_handler>::call_back_starter()", void());
+    if (not state.timers.general.wait_expire)
```
SChernykh, line 187: This is very rarely used in C++ and it's definitely not used in other parts of the Monero codebase. Change it to `if (!state.timers.general.wait_expire)` please

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
-    if (m_was_shutdown)
-        return;
+    if (state.socket.wait_handshake)
+      return;
+    static_assert(
+      epee::net_utils::get_ssl_magic_size() <= sizeof(state.data.read.buffer),
+      ""
+    );
+    auto self = connection<T>::shared_from_this();
+    if (not state.ssl.forced and not state.ssl.detected) {
```
SChernykh, line 205: Same here, use normal C++ operator syntax `!` `&&`

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
-      }else
-      {
-        reset_timer(get_timeout_from_bytes_read(bytes_transferred), false);
-        async_read_some(boost::asio::buffer(buffer_),
-          strand_.wrap(
-            boost::bind(&connection<t_protocol_handler>::handle_read, connection<t_protocol_handler>::shared_from_this(),
-              boost::asio::placeholders::error,
-              boost::asio::placeholders::bytes_transferred)));
-        //_info("[sock " << socket().native_handle() << "]Async read requested.");
+    state.socket.wait_handshake = true;
```
SChernykh, line 270: This is separated from the `if (state.socket.wait_handshake)` check and again not thread safe. It should be done as an atomic operation and in one place, either here or in the beginning of the function.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
-    // disappear and the object will be destroyed automatically after this
-    // handler returns. The connection class's destructor closes the socket.
-    CATCH_ENTRY_L0("connection<t_protocol_handler>::handle_receive", void());
+  template<typename T>
+  void connection<T>::start_shutdown()
+  {
+    if (state.socket.wait_shutdown)
+      return;
+    auto self = connection<T>::shared_from_this();
+    state.socket.wait_shutdown = true;
```
SChernykh, line 573: Again the same non-thread safe pattern.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
-        _dbg1("sleep for queue: " << ms);
-	if (m_was_shutdown)
-		return false;
-
-        if (retry > retry_limit) {
-            MWARNING("send que size is more than ABSTRACT_SERVER_SEND_QUE_MAX_COUNT(" << ABSTRACT_SERVER_SEND_QUE_MAX_COUNT << "), shutting down connection");
-            shutdown();
-            return false;
+        engine rng(seed);
+        return std::chrono::milliseconds(
```
SChernykh, line 774: This is just plain stupid nonsense. You initialize RNG once and then use it, not seed it from std::random_device every time (~2500 bytes!!!). At this stage, you could've just used std::random_device directly. This implementation will deplete system's entropy very quickly.

SChernykh: Yes, seeding the full state (2500 bytes) from std::random_device is totally ok and even good if you do it once. But this code must be rewritten.

mj-xmr:
Please fix the merge conflicts first and allow the new code to pass the **unchanged** tests.

contrib/epee/include/net/abstract_tcp_server2.inl
mj-xmr: > This implementation will deplete system's entropy very quickly.

That's indeed bad for a crypto project, like Monero.

SChernykh: It turns out this code has been there for a long time. Still, since this PR is almost a full rewrite, it's better to fix it like in #8365

mj-xmr: I agree. Large rewrites aren't tolerated here from anybody, and for valid reasons.

Plus we can't use double standards.

mj-xmr:
> Can you at least tell me (even privately) what bugs this is fixing? Otherwise this is possibly a re-write for clarity sake, bordering on a re-write to meet your favored style of callbacks.
>
> Granted, I prefer asio coroutines or this style over the current style, but I'm not sure its worth a giant re-write. The review and testing for this is going to take a while because it impacts the p2p code too.
> 
>  -- <cite>vtnerd</cite>

> "Can you at least tell me (even privately) what bugs this is fixing?" at least those 3 that are triggered by added tests
>
> -- <cite>([perfect-daemon ](https://github.com/monero-project/monero/pull/7760#issuecomment-866326552))</cite> 

@perfect-daemon 
I'd like to learn more about Monero's vulnerability process.
- Regarding the Zero-Day vulnerabilities, that you helped uncover for bad actors especially through your unit tests, and kept them open for the public for already a year: were they addressed through the [Monero's Vulnerability Response Process](https://github.com/monero-project/monero#vulnerability-response)?  
- Can you explain why the branch is not merged yet after almost a whole year? Have you done everything you could, in order to satisfy your Reviewers' requests?
- If you haven't done enough, do you need help, maybe?

selsta:
@mj-xmr You claimed you are involved with reviewing this patch but so far I only see you trolling. Not helpful.

mj-xmr:
@selsta That's because I'm stuffed with so much other work. But if you'd like me to help, I'd gladly create my own version, based on the unit tests written by @perfect-daemon . This would mean that I have to switch my priorities and focus on the discovered vulnerabilities. No problem.

selsta:
@mj-xmr No, I'd prefer if you don't pollute this PR (and my email) with troll comments.

mj-xmr:
I'll try anyway.

jeffro256 (changes requested):
review round one (mostly net_ssl)

contrib/epee/src/net_ssl.cpp
```diff
@@ -522,30 +522,101 @@ bool ssl_options_t::handshake(
...
+      }
+    };
+
+    deadline.async_wait(on_timer);
+    strand.post(
+      [&]{
+        socket.async_handshake(
+          type,
+          boost::asio::buffer(buffer),
+          strand.wrap(on_handshake)
```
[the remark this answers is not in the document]
selsta: I have run this code with boost versions >= 1.66, any idea?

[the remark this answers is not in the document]
jeffro256: Here is the relevant line in `strand.hpp` version 1.58 which provides the method wrap: https://github.com/boostorg/asio/blob/36f5fda32fd518714c00ae6fe23a352ed64a7f66/include/boost/asio/strand.hpp#L223

There is no such method in the commit tagged for Boost 1.6: https://github.com/boostorg/asio/blob/310f245a35a20af59a4cd0030626a4de23c7309c/include/boost/asio/strand.hpp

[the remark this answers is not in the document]
jeffro256: I don't know @selsta, you might have been running a compatibility library? Or the versioning system is off?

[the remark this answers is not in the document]
selsta: Could it be that you looked in the wrong file?

https://github.com/boostorg/asio/blob/develop/include/boost/asio/io_context_strand.hpp has wrap

contrib/epee/src/net_ssl.cpp
```diff
@@ -522,30 +522,101 @@ bool ssl_options_t::handshake(
...
+    using socket_t = boost::asio::ip::tcp::socket;
+
+    auto &io_context = GET_IO_SERVICE(socket);
+    if (io_context.stopped())
+      io_context.reset();
+    strand_t strand(io_context);
+    timer_t deadline(io_context, timeout);
+
+    struct state_t {
+      lock_t lock;
```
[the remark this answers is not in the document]
selsta: Did you check the `strand_deadlock` test case? I'm quite sure some of these design decisions were intentional to solve this issue, together with setting `-DBOOST_ASIO_ENABLE_SEQUENTIAL_STRAND_ALLOCATION`

But I also don't want to give false info so maybe ignore if this doesn't make sense.

contrib/epee/src/net_ssl.cpp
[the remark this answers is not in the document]
jeffro256: Figured it out: [it's deprecated](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/io_context__strand/wrap.html)

contrib/epee/src/net_ssl.cpp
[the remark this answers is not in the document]
jeffro256: Yes I'm sure the author put the locks in there for a reason, due to the current design of `handshake`, but since this is a complete rewrite,  it doesn't make sense to have both code running asynchronously from a single executor AND have status booleans AND have it be manually, externally locked. Even if it functions correctly, this will likely gum up the `io_context` of `socket` waiting for locks and it makes it a ton harder to track what's going on.

contrib/epee/src/net_ssl.cpp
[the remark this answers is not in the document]
selsta: We still support 1.59 and bind_executor isn't supported in that verison so we would have to add `#if`

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
-    TRY_ENTRY();
-    _dbg2("[" << print_connection_context_short(context) << "] fired_callback");
-    m_protocol_handler.handle_qued_callback();
-    CATCH_ENTRY_L0("connection<t_protocol_handler>::call_back_starter()", void());
+    if (not state.timers.general.wait_expire)
+      return;
+    state.timers.general.cancel_expire = true;
+    state.timers.general.reset_expire = false;
+    ec_t ec;
+    timers.general.cancel(ec);
```
jeffro256, line 192: `cancel` overload with error code is deprecated. `cancel` overload without error code out parameter throws `system_error`

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
+              else if (state.status == status_t::INTERRUPTED)
+                on_interrupted();
+              else if (state.status == status_t::TERMINATING)
+                on_terminating();
+            }
+            else if (ec.value()) {
+              terminate();
+            }
+            else if (
+              not epee::net_utils::is_ssl(
```
jeffro256, line 231: like other have suggested, use `!` instead of `not`

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
+        );
+      }
+    };
+    if (not state.ssl.enabled)
+      connection_basic::socket_.next_layer().async_read_some(
+        boost::asio::buffer(
+          state.data.read.buffer.data(),
+          state.data.read.buffer.size()
+        ),
+        strand.wrap(on_read)
```
jeffro256, line 430: deprecated

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -81,860 +81,1049 @@ namespace net_utils
...
+      );
+    else
+      strand.post(
+        [this, self, on_read]{
+          connection_basic::socket_.async_read_some(
+            boost::asio::buffer(
+              state.data.read.buffer.data(),
+              state.data.read.buffer.size()
+            ),
+            strand.wrap(on_read)
```
jeffro256, line 440: deprecated

jeffro256 (changes requested):
Round 2

tests/unit_tests/epee_boosted_tcp_server.cpp
```diff
@@ -464,3 +476,241 @@ TEST(test_epee_connection, test_lifetime)
...
+  ssl_context_t ssl_context{boost::asio::ssl::context::sslv23};
+  io_context_t io_context;
+  ssl_socket_t socket(io_context, ssl_context);
+  ec_t ec;
+  socket.next_layer().connect(endpoint, ec);
+  EXPECT_EQ(ec.value(), 0);
+  socket.handshake(boost::asio::ssl::stream_base::client, ec);
+  EXPECT_EQ(ec.value(), 0);
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  while (server.get_config_shared()->get_connections_count() < 1);
```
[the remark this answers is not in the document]
kayabaNerve: This test expects this condition to be met, yet should instead error if this condition isn't, as this condition not being met is a failure of the code being run. I'd also appreciate a sleep statement being added to reduce CPU utilization while this condition is checked for. Same for the other infinite loop noted below this (and likely any others).

j-berman (commented):
Generally, I agree the style is non-conforming and it would be nicer if it conformed to existing conventions better. But I don't think that is a good blocker for a PR that solves reliability issues, especially one that (justifiably, imo) rewrites large sections to do so. The author has fair justification for the decisions made in the code (some reasoning explained via DM). As is, this PR is a clear improvement over the current code.

Further, in discussions with the author, and [in some of the answers given here](https://github.com/monero-project/monero/pull/7760/files#r656730637), this PR is a stepping stone to even more improvements. To that end, I do think this is in line with the desired piecemeal approach to solving issues in the daemon. This PR does indeed seem to be a smaller piece of a puzzle toward even larger improvements, even if it is fairly large in its own right.

I'll approve it once merge conflicts are resolved satisfactorily.

_____

Separate note...

I strongly suggest anyone who wants to fully understand this code run the old code with the provided tests and observe where the failures occur, then review the 3 commits of [this branch](https://github.com/monero-project/monero/compare/master...j-berman:7760-streamlined-logic) to understand the code in this PR that is fixing the issues, then read and step through this PR's code to understand how the code is working from a bigger picture perspective. Happy to answer questions/help make it clear what the issues are and what is being fixed. The PR author was also helpful to me in understanding this PR over DM, mainly after I made it clear I did step through the code myself and demonstrated I understood various problems it's solving.

contrib/epee/include/net/abstract_tcp_server2.inl
j-berman: > All of these changes don't really change the behavior of the existing problem. Multiple threads can get stuck busy waiting for a queue to clear. The previous method had the threads waiting indefinitely at a lock, whereas here they wait indefinitely in conditional variables.

AFAICT this approach is a marked improvement over the current code:

1. Reads/writes are no longer waiting on `handle_recv` or callbacks to complete (which execute in the newly introduced 2nd `strand_`), therefore they can process quicker and clear from the strand quicker.
2. It uses a conditional wait that can unblock as soon as the queue is less than the max.

Without a demonstrable DoS on this new code as is which is evidently harder to DoS, I think the code is merge-able as is, and worth continuing to improve, which the PR author intends to do building from the structure of this PR's code

Additionally there is a risk in dropping the connection here, which is that it can feasibly harm honest usage of the daemon. It's going to take more work to dig deeper to make this entire section air tight to hit the optimal sweet spot imo. For now, this code is an improvement moving in a better direction.

Gingeropolous:
because of #8426, can this PR be closed?

selsta:
Merged in https://github.com/monero-project/monero/pull/8426

PR #7762, opened 2021-06-24
trezor: add #if for ByteSizeLong

selsta:
Turns out Ubuntu 18.04 ships with an old protobuf version.

### DIFF
--- a/src/device_trezor/trezor/transport.cpp
+++ b/src/device_trezor/trezor/transport.cpp
@@ -157,7 +157,11 @@ namespace trezor{
 #define PROTO_HEADER_SIZE 6
 
   static size_t message_size(const google::protobuf::Message &req){
+#if GOOGLE_PROTOBUF_VERSION < 3006001
+    return size_t(req.ByteSize());
+#else
     return req.ByteSizeLong();
+#endif
   }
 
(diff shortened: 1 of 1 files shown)


src/device_trezor/trezor/transport.cpp
```diff
@@ -157,7 +157,11 @@ namespace trezor{
 #define PROTO_HEADER_SIZE 6
 
   static size_t message_size(const google::protobuf::Message &req){
+#if GOOGLE_PROTOBUF_VERSION < 3006001
+    return static_cast<size_t>(req.ByteSize());
```
vtnerd, line 161: Trivial, but `size_t(req.ByteSize())` (no `static_cast` needed here) is also valid in C++ (constructor syntax). The behavior is also identical.

vtnerd (approved):

src/device_trezor/trezor/transport.cpp
selsta: updated

mj-xmr (approved):

mj-xmr:
https://github.com/monero-project/monero/pull/7769 confirms that the fix works across Ubuntu 20.04 and 18.04

PR #7765, opened 2021-06-30
depends: add getmonero package mirror

selsta:
~~Package archives are not uploaded yet to https://downloads.getmonero.org/depends-sources~~ done

I attempted to edit `fetch_file` so that `$(package)_file_name` is always used for the fallback download.

### DIFF
--- a/contrib/depends/funcs.mk
+++ b/contrib/depends/funcs.mk
@@ -32,7 +32,7 @@ endef
 define fetch_file
     ( test -f $$($(1)_source_dir)/$(4) || \
     ( $(call fetch_file_inner,$(1),$(2),$(3),$(4),$(5)) || \
-      $(call fetch_file_inner,$(1),$(FALLBACK_DOWNLOAD_PATH),$(3),$(4),$(5))))
+      $(call fetch_file_inner,$(1),$(FALLBACK_DOWNLOAD_PATH),$(4),$(4),$(5))))
 endef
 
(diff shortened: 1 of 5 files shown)


contrib/depends/funcs.mk
```diff
@@ -32,7 +32,7 @@ endef
 define fetch_file
     ( test -f $$($(1)_source_dir)/$(4) || \
     ( $(call fetch_file_inner,$(1),$(2),$(3),$(4),$(5)) || \
-      $(call fetch_file_inner,$(1),$(FALLBACK_DOWNLOAD_PATH),$(3),$(4),$(5))))
+      $(call fetch_file_inner,$(1),$(FALLBACK_DOWNLOAD_PATH),$(4),$(4),$(5))))
```
mj-xmr, line 35: Could you explain why this change and the changes in the packages were needed?

selsta: Some download file names only contain the version, e.g. `contrib/depends/packages/eudev.mk` download file name is `v3.2.6.tar.gz`, which can clash with other packages and also isn't self describing, so it gets renamed to `eudev-v3.2.6.tar.gz` after download.

The fallback mirror has `eudev-v3.2.6.tar.gz` uploaded so the call for `fetch_file_inner` has to be updated in a way that the download file name is never used.

mj-xmr (approved):

PR #7766, opened 2021-07-01
wallet2: output labeling

tobtoht:
An often requested feature relevant to coin control is the ability to individually label outputs. This PR adds a description field to `transfer_details` and adds a function to set it.


selsta (approved):

tobtoht:
Cache no longer uses boost serialization (7175dcb10) and Monero serialization has missing `VERSION_FIELD` on `transfer_details` struct, so this _does_ break old caches. Seems excessive for such a minor change.

Mooo pointed out that `set_tx_note` can be used for the purpose of output labeling, will use that instead.

PR #7767, opened 2021-07-01
Fix describe_transfer for multiple txes in a txset

### LINKED ISSUE #7344: describe_transfer has incorrect output when multiple transactions are present

selsta:
Rebase of #7349

This ensures each list of recipients is only the recipients
for one transaction. It also adds a new field "summary"
that describes the txset as a whole.

Fixes #7344

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -1292,13 +1292,20 @@ namespace tools
     try
     {
       // gather info to ask the user
-      std::unordered_map<cryptonote::account_public_address, std::pair<std::string, uint64_t>> dests;
+      std::unordered_map<cryptonote::account_public_address, std::pair<std::string, uint64_t>> tx_dests;
+      std::unordered_map<cryptonote::account_public_address, std::pair<std::string, uint64_t>> all_dests;
       int first_known_non_zero_change_index = -1;
+      res.summary.amount_in = 0;
+      res.summary.amount_out = 0;
+      res.summary.change_amount = 0;
+      res.summary.fee = 0;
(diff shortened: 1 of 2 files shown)


src/wallet/wallet_rpc_server.cpp
```diff
@@ -1374,29 +1377,45 @@ namespace tools
...
 
         desc.fee = desc.amount_in - desc.amount_out;
         desc.unlock_time = cd.unlock_time;
         desc.extra = epee::to_hex::string({cd.extra.data(), cd.extra.size()});
+
+        // Update summary items
+        res.summary.amount_in += desc.amount_in;
+        res.summary.amount_out += desc.amount_out;
+        res.summary.change_amount += desc.change_amount;
+        res.summary.fee += desc.fee;
```
moneromooo-monero, line 1413: Those aren't initialized at the top. It might be they always are in practice, but it seems best to init them at the start of the function.

moneromooo-monero (approved):

PR #7768, opened 2021-07-03
CI: Boost ver. matrix for build-ubuntu

mj-xmr:
Followed by the issue https://github.com/monero-project/monero/issues/7728 and its corresponding fix: https://github.com/monero-project/monero/pull/7735 , I have decided that it would be reasonable to start checking backward compatibility of Monero. The new jobs test the build process for:
- Ubuntu 18.04
- Ubuntu 20.04

as well as Boost versions available for both distro releases:
- 1.62 (18.04)
- 1.65 (18.04)
- 1.67 (20.04)
- 1.71 (20.04)

Temp. info: I discovered that the `Ubuntu 18.04` build fails because of this fix for `20.04` https://github.com/monero-project/monero/pull/7719 . Fixing this is being realized by https://github.com/monero-project/monero/pull/7762. The current PR can only be merged after merging https://github.com/monero-project/monero/pull/7762 , hence flagging the current one as a draft, even though it's (almost) complete.

PR #7769, opened 2021-07-04
CI: ubuntu 18.04 backward compat. test

mj-xmr:
A follow up of https://github.com/monero-project/monero/pull/7768 , where it was decided, that Ubuntu 18.04 shall be tested, but is a separate use case than Boost matrix use case. The Ubuntu 18.04 use case thus doesn't need the Boost matrix, and can be built with the default options.

Temp. info: This draft PR contains a commit to be reverted, taken from https://github.com/monero-project/monero/pull/7762 , which has to be merged first. Afterwards this PR will be published and rebased. The specific to-be-reverted commit helps to check if the both the fix and the verification method work.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,11 +60,17 @@ jobs:
         ccache --max-size=150M
         make release-static-win64 -j2
 
+# See the OS labels and monitor deprecations here:
+# https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
+
   build-ubuntu:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     env:
       CCACHE_COMPRESS: 1
       CCACHE_TEMPDIR: /tmp/.ccache-temp
+    strategy:
+      matrix:
+        os: [ubuntu-latest, ubuntu-18.04]
     steps:
     - uses: actions/checkout@v1
       with:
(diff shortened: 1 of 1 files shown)


selsta:
This should be closed with reference to #7770

mj-xmr:
> This should be closed with reference to #7770

For the record: these 2 PRs have become 2 distinctive things. This one tests for a generic Ubuntu 18 compatibility with all the default packages (read: no mentioning of Boost), while #7770 is about Boost specifically.

selsta:
Ok, it's fine keeping them separate. Please remove the Trezor test and the draft status.

mj-xmr:
> Ok, it's fine keeping them separate. Please remove the Trezor test and the draft status.

Will do soon, but stay adviced, that this will make the checks fail. As long as the PRs are merged in order, this will not be a problem, so please make an exception here.
@luigi1111 : FYI

selsta:
This will get merged later anyway.

mj-xmr:
Done.

selsta (changes requested):
This should be a matrix with both 18.04 and 20.04 added to the existing `build-ubuntu` test.

mj-xmr:
> This should be a matrix with both 18.04 and 20.04 added to the existing `build-ubuntu` test.

OK will do.

.github/workflows/build.yml
```diff
@@ -171,3 +179,4 @@ jobs:
       with:
         name: ${{ env.OUTPUT }}
         path: /home/runner/work/monero/monero/${{ env.OUTPUT }}
+
```
selsta, line 182: unintended newline

selsta (approved):

mj-xmr:
@selsta : Shall I rebase and get the checks right, since #7762 has been merged?

selsta:
However you want.

mj-xmr:
So I rebased.

selsta (approved):

PR #7770, opened 2021-07-05
Ci boost min max test

mj-xmr:
As a follow up of https://github.com/monero-project/monero/pull/7768 , I wanted to separate Boost compatibility test from Ubuntu 18.04 backward compatibility test.

## Findings:
- Ubuntu 20.04 (a.k.a. "latest" at the time of being) is able to compile the latest Boost version (1.76), as well as the minimum officially supported version (1.58), so building the older Boost versions under Ubuntu 18.04 isn't necessary.
- Ubuntu 20.04's OpenSSL 1.1.1k refuses to link against Boost of version less than 1.62 (1).
- Ubuntu 18.04 is being employed, to test if < 1.62 can be used, ideally the minimum supported version 1.58, but it fails, giving a similar error (2), this time for 18.04's OpenSSL 1.1.1.

## Conclusions:
- A properly patched, and still supported Ubuntu version imposes a minimum Boost version of 1.62. This has to be reflected in near future.
- Ubuntu 18.04 doesn't need to be tested for Boost compatibility, as 20.04 does the job already.

## Temp info
The branch contains some commits to be reverted after the reviewers take a proper look. They are marked as such. There should be 2 CI checks - one with the minimum version of 1.62, that succeeds and will be scheduled for merging, and one on top of this - simulating a usage of a version just one tick older than 1.62 -> 1.61, where I want to prove the OpenSSL's reaction.


mj-xmr:
ping @selsta - this is mostly relevant to you.

selsta (commented):
There is no point in testing both Ubuntu 18.04 and 20.04 with multiple boost version. We have to consider that we don't have unlimited parallel jobs.

I think it would be fine:

- Ubuntu 18.04 with apt boost
- Ubuntu 20.04 with latest boost manually compiled.

And nothing else.

mj-xmr:
> There is no point in testing both Ubuntu 18.04 and 20.04 with multiple boost version. We have to consider that we don't have unlimited parallel jobs.

I know. 18.04 is there just to show that even 18.04 isn't able to compile with boost 1.58, so the minimum has to be raised to 1.62. (see "Conclusions")

> I think it would be fine:
> 
> * Ubuntu 18.04 with apt boost

This is done in https://github.com/monero-project/monero/pull/7769, which in itself has nothing to do with Boost. But I can move this here, if you prefer.

> * Ubuntu 20.04 with latest boost manually compiled.
> 
> And nothing else.

And what about being able to test what the minimum currently supported version is? Without the test of 1.62 under 20.04 we wouldn't be able to tell, that the minimum is not 1.58 anymore.

I will prepare the PR in the way how I would see it:
- moving https://github.com/monero-project/monero/pull/7769 here, if you insist
- keeping Boost 1.62 (minimum) and the latest with 20.04

mj-xmr:
TODOs from IRC:

- compile a fitting version of OpenSSL from source, that would match Boost 1.58
- try to achieve this under Ubuntu 20.04, so that Ubuntu 18.04 doesn't need to be used

mj-xmr:
Cleared the branch of foreign commit.

.github/workflows/build.yml
```diff
@@ -5,6 +5,9 @@ on: [push, pull_request]
 # The below variables reduce repetitions across similar targets
 env:
   REMOVE_BUNDLED_BOOST : rm -rf /usr/local/share/boost
+  APT_PURGE_DEFAULT_BOOST : |
```
selsta, line 8: This isn't reused so no need to put this up here.

selsta (commented):
Thinking about this again, it doesn't really make sense. We already test the latest boost version on macOS, there is no point to also test it under Ubuntu.

.github/workflows/build.yml
mj-xmr: Removed.

mj-xmr:
> Thinking about this again, it doesn't really make sense. We already test the latest boost version on macOS, there is no point to also test it under Ubuntu.

It didn't catch the error, which inspired this PR.

selsta:
> It was only visible under Linux.

This is incorrect, see how the GUI Mac CI still fails due to this exact error: https://github.com/monero-project/monero-gui/runs/3042790411

This PR also would not have caught it because it isn't auto updating.

mj-xmr:
Ok. The reason why Mac didn't catch this error was, that at that time the Mac's "boost latest" wasn't actually the latest yet.
I will focus more on local tests then and publish them as a report. Locally I already use Boost latest under Ubuntu.

PR #7776, opened 2021-07-08
workflows: fix libwallet ci test

selsta:


mj-xmr (approved):

.github/workflows/build.yml
```diff
@@ -112,8 +112,8 @@ jobs:
     - name: build
       run: |
         ccache --max-size=150M
-        cmake -DBUILD_GUI_DEPS=ON
-        make -j3
+        cmake .
+        make wallet_api -j3
```
mj-xmr, line 116: Could you try instead:
```
ccache --max-size=150M
mkdir -p build && cd build
cmake .. -DBUILD_GUI_DEPS=ON
make -j3 libwallet_api_tests
ctest
```
? 

This should enter the `libwallet_api_tests` directory as per [this line](https://github.com/monero-project/monero/blob/master/tests/CMakeLists.txt#L100), while the `ctest` should be able to execute the only built test, the `libwallet_api_tests`.

selsta: This PR here is just for checking if libwallet compiles, as it was previously.

Libwallet tests can be looked into separately.

PR #7779, opened 2021-07-09
Depends: mingw32-update-gcc-posix.sh script

mj-xmr:
Extracted the MinGW POSIX setting (1) into a script, which may be reused without much effort, ideal for local tests. Also used the script in the `depends.yaml` GitHub Action.

From Monero's main directory:

```bash
sudo contrib/depends/util/mingw32-update-gcc-posix.sh x86_64-w64-mingw32
sudo contrib/depends/util/mingw32-update-gcc-posix.sh i686-w64-mingw32
```

instead of formerly:

```bash
sudo update-alternatives --set x86_64-w64-mingw32-g++ $(which x86_64-w64-mingw32-g++-posix)
sudo update-alternatives --set x86_64-w64-mingw32-gcc $(which x86_64-w64-mingw32-gcc-posix)

sudo update-alternatives --set i686-w64-mingw32-g++ $(which i686-w64-mingw32-g++-posix)
sudo update-alternatives --set i686-w64-mingw32-gcc $(which i686-w64-mingw32-gcc-posix)
```

(1) The setting is required for the MinGW cross-compiler to build Monero.


selsta:
There is not much benefit to this change.

PR #7780, opened 2021-07-10
Workflows: depends cache (static) separated from ccache (volatile)

mj-xmr:
I separated `depends`' caches, which were treated as one, into 3 categories:
- most volatile: ccache
- much less volatile: depends packages, that are the base of the toolchains
- static: OSX SDK

In order to achieve this, each cache type uses a different key generation mechanism:
- ccache: the traditional git's commit hash
- depends: a hash of the package list, that build the binaries in the `built` directory
- OSX SDK: simply the version number

This strategy should help reducing the size cache by not mixing the volatile caches with the much more static ones, since there will be fewer writes of the static ones. The newly created volatile caches will not include the static data. This will also help us stay within our limits, ultimately preventing GitHub from deleting our caches due to their cumulative size restrictions. See below:

> **Usage limits and eviction policy**
GitHub will remove any cache entries that have not been accessed in over 7 days. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 5 GB. If you exceed this limit, GitHub will save your cache but will begin evicting caches until the total size is less than 5 GB.

[Source: GitHub docs](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy)

### DIFF
--- a/.github/workflows/depends.yml
+++ b/.github/workflows/depends.yml
@@ -45,15 +45,29 @@ jobs:
     - uses: actions/checkout@v1
       with:
         submodules: recursive
+# Most volatile cache
     - name: ccache
       uses: actions/cache@v2
       with:
-        path: |
-            ~/.ccache
-            contrib/depends/built
-            contrib/depends/sdk-sources
+        path: ~/.ccache
         key: ccache-${{ matrix.toolchain.host }}-${{ github.sha }}
         restore-keys: ccache-${{ matrix.toolchain.host }}-
+# Less volatile cache
+    - name: depends cache
+      uses: actions/cache@v2
+      with:
+        path: contrib/depends/built
+        key: depends-${{ matrix.toolchain.host }}-${{ hashFiles('contrib/depends/packages/*') }}
+        restore-keys: |
+          depends-${{ matrix.toolchain.host }}-${{ hashFiles('contrib/depends/packages/*') }}
+          depends-${{ matrix.toolchain.host }}-
+# Static cache
+    - name: OSX SDK cache
+      uses: actions/cache@v2
+      with:
+        path: contrib/depends/sdk-sources
+        key: sdk-${{ matrix.toolchain.host }}-${{ matrix.toolchain.osx_sdk }}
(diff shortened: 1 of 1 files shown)


selsta:
We are nowhere near the 5GB limit so why exactly this change? The linked documentation does not explain it to me.

mj-xmr:
> We are nowhere near the 5GB limit so why exactly this change? The linked documentation does not explain it to me.

For each push we create 9 depends jobs with ~200 MB each + 4 `builds.yml` with ~120 MB each, which gives almost 2.3 GB in total. I'd say, that we prune our old caches every 3rd time we push to the repo.

selsta:
I still don't understand how splitting them up helps in any way.

mj-xmr:
> I still don't understand how splitting them up helps in any way.

Because the other 2 (depends and SDK) will not be created anew, since their keys are not subject to change at every new commit. The key for `depends` is the hash of the build scripts, while in the extreme case of the SDK, it's just the SDK's version.

selsta (approved):

mj-xmr:
There's just a download timeout problem with one `depends` build, but since you've just approved, I won't retrigger the checks since this would mean another push.

mj-xmr:
> > I still don't understand how splitting them up helps in any way.
> 
> Because the other 2 (depends and SDK) will not be created anew, since their keys are not subject to change at every new commit. The key for `depends` is the hash of the build scripts, while in the extreme case of the SDK, it's just the SDK's version.

Here's a proof:
https://github.com/mj-xmr/monero/runs/3036341529?check_suite_focus=true

selsta:
> There's just a download timeout problem with one depends build,

Link?

mj-xmr:
> > There's just a download timeout problem with one depends build,
> 
> Link?

https://github.com/monero-project/monero/pull/7780/checks?check_run_id=3036470355

It appears to be a temporary problem. I might as well re-trigger the checks.

selsta:
Should stop being a problem once the depends mirror is merged.

mj-xmr:
@luigi1111 The failed check is only due to connection problems. Otherwise all ok and is approved by @selsta.

PR #7788, opened 2021-07-13
wallet2: Don't auto lock device on process parsed blocks

tobtoht:
Let device implementation handle locking, if necessary. With ledger, if the viewkey was exported no locking is required because the software implementation is used. With Trezor, the software implementation is always used. This patch speeds up wallet refreshes by 6-7x.

Requires #7745.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -2773,9 +2773,8 @@ void wallet2::process_parsed_blocks(uint64_t start_height, const std::vector<cry
   {
     if (tx_cache_data[i].empty())
       continue;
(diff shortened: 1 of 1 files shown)


selsta (changes requested):
```
/Users/selsta/dev/monero/src/wallet/wallet2.cpp:2777:29: warning: lambda capture 'hwdev' is not used [-Wunused-lambda-capture]
    tpool.submit(&waiter, [&hwdev, &gender, &tx_cache_data, i]() {
                           ~^~~~~~
```

selsta (approved):

Issue #7783, opened 2021-07-11
Trezor: rephrase "Device wallet does not match wallet adress" error message


ph4r05:
Recently, few users reported problems with opening their Trezor-based Monero wallets:

- https://www.reddit.com/r/monerosupport/comments/ohfegp/help_cant_access_my_gui_wallet_trezor_model_t/h4q4z2b/
- more: https://github.com/trezor/trezor-firmware/issues/1686

Often happens that the culprit is usage of a different Trezor passphrase (or typo in the passphrase, capitalization, ...). Some users are not quite aware that different passphrase yields a different Monero address, which causes confusion (and support tickets). 

So it might be beneficial to rephrase error from [wallet2.cpp](https://github.com/monero-project/monero/blob/de3456e1275836725291ba71036b7ef0e2cda91f/src/wallet/wallet2.cpp) so it indicates there might be problem with a passphrase. https://github.com/monero-project/monero/blob/de3456e1275836725291ba71036b7ef0e2cda91f/src/wallet/wallet2.cpp#L4425-L4427

I am not sure what a rephrased error message should be and if it is the right place to address the passphrase problem, thus I am starting this issue to discuss. 

What do you think about:
```cpp
THROW_WALLET_EXCEPTION_IF(device_account_public_address != m_account.get_keys().m_account_address, error::wallet_internal_error, 
  "Device wallet address does not match wallet address. " 
  "Device address: " + cryptonote::get_account_address_as_str(m_nettype, false, device_account_public_address) + 
  ", wallet address: " + m_account.get_public_address_str(m_nettype) + 
  ". If device uses passphrase, please check it is correctly entered (case-sensitive, it might be misspelled, different passphrase generates different address)"); 
```

Feel free to suggest something better. 

What do you think @dEBRUYNE-1, @selsta ? Thanks!

dEBRUYNE-1:
Agree with the proposal in principle. With respect to the text, perhaps we can rename as follows:

> If the device uses the passphrase feature, please check whether the passphrase was entered correctly (it may have been misspelled - different passphrases generate different wallets)

PR #7790, opened 2021-07-14
wallet: rephrase error message on invalid device address

ph4r05:
A proposal to address issues like #7783, https://github.com/trezor/trezor-firmware/issues/1686
Added @dEBRUYNE-1 's improvement.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -4422,7 +4422,7 @@ bool wallet2::load_keys_buf(const std::string& keys_buf, const epee::wipeable_st
 
     account_public_address device_account_public_address;
     THROW_WALLET_EXCEPTION_IF(!hwdev.get_public_address(device_account_public_address), error::wallet_internal_error, "Cannot get a device address");
-    THROW_WALLET_EXCEPTION_IF(device_account_public_address != m_account.get_keys().m_account_address, error::wallet_internal_error, "Device wallet does not match wallet address. "
+    THROW_WALLET_EXCEPTION_IF(device_account_public_address != m_account.get_keys().m_account_address, error::wallet_internal_error, "Device wallet does not match wallet address. If the device uses the passphrase feature, please check whether the passphrase was entered correctly (it may have been misspelled - different passphrases generate different wallets, passphrase is case-sensitive). "
                                                                                                                                      "Device address: " + cryptonote::get_account_address_as_str(m_nettype, false, device_account_public_address) +
                                                                                                                                      ", wallet address: " + m_account.get_public_address_str(m_nettype));
     LOG_PRINT_L0("Device inited...");


selsta (approved):

tobtoht (approved):

ph4r05:
@dEBRUYNE-1 should this be opened also against release branch?

selsta:
@ph4r05 yes please

rating89us:
Possible causes for the "device wallet doesn't match wallet address" error are:
- User entered wrong passphrase.
- User opened a Monero GUI wallet file with non-blank PF but Trezor device has PF protection disabled.
- Trezor device has been initialized/restored with a seed that is different from the one used to create the Monero GUI wallet file.

The error message in this PR is only contemplating the first cause, but it should also contemplate causes #2 and #3.

ph4r05:
#1 is the most common issue, hence we exicitly mention it in the error message as users are usually stil unaware how passphrase works.

#2 is a subset of #1 and I don't think we need to address this explicitly. If user disabled passphrase somehow, he likely knows how to enable it and Trezor Suite enables it by default for all. 

#3 is "you have wrong device". I think the error message indicates this as well.

I don't think we need to mention all possible causes in the error messages. The longer it gets the more confusing it will be. So I believe we are doing some tradeoffs w.r.t. UX.

I am happy with current version of the message in this PR. However, feel free to come up with other version of the message, it's up to debate. Without anything better, this stays :)

rating89us:
Current message is:
`"Device wallet does not match wallet address. If the device uses the passphrase feature, please check whether the passphrase was entered correctly (it may have been misspelled - different passphrases generate different wallets, passphrase is case-sensitive)."`

The "If the device uses the passphrase feature" can be removed, because after https://github.com/monero-project/monero/pull/7793 this message will only appear when passphrase feature is enabled on the device or when the device has a wrong seed (user has two devices and plugged the wrong one/user restored a different seed on the same device).

I would still include cause #2 because you said it will not be solved soon on Monero GUI, and this was the problem that originated this whole discussion about passphrases.

So I suggest the following error message, it's shorter (240 characters vs. 257 of current message), it contemplates all 3 causes, and additionally it gives hints on how to solve causes #1 and #2.:
`"Device wallet does not match wallet address. This can be caused by a wrong passphrase (passphrase is case-sensitive), a device with passphrase encryption disabled (can be enabled on Trezor Suite), or a device with a different recovery seed".`

ph4r05:
> The "If the device uses the passphrase feature" can be removed, because after #7793 this message will only appear when passphrase feature is enabled on the device or when the device has a wrong seed

#7793 will be merged after this one, for sure. I am not even sure #7793 makes it to the release-v0.17. 

Also, this message also displays on PF-disabled Trezors and other hardware wallets (Ledges does not have passphrase AFAIK). 

After is this merged, we may address error message once again. I would prefer to close this PR soon and deal with changes later (#7793 can include this one).

Also, "passphrase encryption" is not correct term. Encryption should not be mentioned at all in this context.

rating89us:
>  I would prefer to close this PR soon and deal with changes later (#7793 can include this one).

Ok. No problem.

> Also, "passphrase encryption" is not correct term. Encryption should not be mentioned at all in this context.

This is the term displayed on Trezor screen when you disable/enabled it on Trezor Suite.

ph4r05:
@rating89us thanks for discussion on Passphrase-related error messages. We may get back to this topic once this one is merged and there are reviews on #7793.

> This is the term displayed on Trezor screen when you disable/enable passphrase feature on Trezor Suite.

This is interesting, I indeed found it in https://github.com/trezor/trezor-suite/blob/4be72d9dfe93910297f94df77c19232d015b0bbc/packages/suite/src/support/messages.ts#L1628 , I am now discussing this wording with Trezor devs.

PR #7793, opened 2021-07-14
trezor: try empty passphrase first [WIP]

ph4r05:
@matejcik in https://github.com/trezor/trezor-firmware/issues/1686#issuecomment-879789424 proposed interesting idea for opening wallets with passphrases:

> An improvement that can be done in Monero GUI immediately and even without the standard/hidden thing:
when you're opening the wallet, query the empty-passphrase address. If it matches, do not even show the passphrase dialog and proceed straight to wallet.

This PR implements the proposal. The flow is:
- If device is Trezor, try loading address of the wallet with empty passphrase (no user prompt)
- If address matches wallet address, open without passphrase prompt
- If address does not match, try once again to load address, without passphrase prompt suppression. If there is a passphrase used, user can enter it. 

Benefits:
- Easier access for Trezors with passphrase enabled (soon, all devices will have this by default, as Trezor Suite enables passphrase by default)
- Simplifies wallet opening for empty passphrases

Cons:
- Opening standard wallet with non-empty passphrase takes a bit longer, as empty passphrase is tried first. If needed, we could address this by storing non-empty-passphrase to wallet file and skipping this step.

This is just a PoC for discussion. Thanks for ideas.


src/wallet/wallet2.cpp
```diff
@@ -4421,7 +4421,25 @@ bool wallet2::load_keys_buf(const std::string& keys_buf, const epee::wipeable_st
...
+
+    ::hw::device_cold* dev_cold = nullptr;
+    if (m_key_device_type == hw::device::device_type::TREZOR && (dev_cold = dynamic_cast<::hw::device_cold*>(&hwdev)) != nullptr) {
+      THROW_WALLET_EXCEPTION_IF(!dev_cold->get_public_address_with_no_passphrase(device_account_public_address), error::wallet_internal_error, "Cannot get a device address");
+      if (device_account_public_address == m_account.get_keys().m_account_address) {
+        LOG_PRINT_L0("Wallet opened with an empty passphrase");
+        fetch_device_address = false;
+        dev_cold->set_use_empty_passphrase(true);
+      } else {
+        fetch_device_address = true;
```
ph4r05, line 4434: I wanted to use `dev_cold-> seen_passphrase_entry_prompt()` and to try again with passphrase only if device obviously asked for any. But to avoid potential problems we can just ask anyway. The worst case is that address is loaded twice.

ph4r05:
@selsta what is pls your opinion about this?

selsta:
will take a look

src/device_trezor/device_trezor_base.cpp
```diff
@@ -491,19 +498,21 @@ namespace trezor {
       }
 
       boost::optional<epee::wipeable_string> passphrase;
-      TREZOR_CALLBACK_GET(passphrase, on_passphrase_request, on_device);
+      if (m_reply_with_empty_passphrase || m_always_use_empty_passphrase) {
+        MDEBUG("Answering passphrase prompt with an empty passphrase, always use empty: " << m_always_use_empty_passphrase);
+        on_device = false;
+        passphrase = epee::wipeable_string("");
+      } else if (m_passphrase){
+        MWARNING("Answering passphrase prompt with a stored passphrase (do not use; passphrase can be seen by a potential malware / attacker)");
```
selsta, line 506: Can you explain "stored passphrase" ?

selsta:
I tested it and it works well, I think this is something we can add.

src/device_trezor/device_trezor_base.cpp
ph4r05: Sure, it was part of the old API, you could set `m_passphrase` to the device object, the passphrase will be then used instead of prompting user. I've added this warning as in general, we don't want users entering their passhprases on computers, but this may have useful use-cases when you know what are you doing. E.g., in tests, automation, etc. It was part of the API so I don't feel like removing it now.

rating89us:
I tested this PR with my six Monero GUI wallet files that I created when testing (see https://github.com/trezor/trezor-firmware/issues/1686#issuecomment-879766277).

First I tried opening the six Monero GUI wallet files with PF-disabled Trezor. This PR doesn't fix the "ERROR: couldn't open wallet. Device wallet (46...) does not match wallet address (48...)" observed when I try to open Monero GUI wallet files with PF (wallets C & D). It seems that Trezor is still returning a blank PF to Monero GUI.

Then I tried opening the six Monero GUI wallet files with PF-enabled Trezor. This PR fixes the UX experience problem observed when I try to open Monero GUI wallet files with no PF (Wallets E & F), since it doesn't requires anymore the user to enter a blank PF. You just have to type the password and the wallet opens normally, without asking a PF.

selsta:
> First I tried opening the six Monero GUI wallet files with PF-disabled Trezor. This PR doesn't fix the "ERROR: couldn't open wallet. Device wallet (46...) does not match wallet address (48...)" observed when I try to open Monero GUI wallet files with PF (wallets C & D). It seems that Trezor is still returning a blank PF to Monero GUI.

Trezor will in the future always enable passphrase so this shouldn't be an issue anymore.

rating89us:
If Monero GUI could detected whether the Trezor device has PF protection enabled or not, we could display this message (instead of displaying address doesn't match error): "_You are opening a wallet file with PF, but you are using a PF-disabled Trezor device. Please open Trezor Suite and enable PF protection for your device in order to open this wallet file_."

Or maybe it's possible to offer directly on Monero GUI the option to enable PF protection for the Trezor device.

ph4r05:
Interesting points. Thanks for feedback!

This PR indeed does not address disabled passphrase Trezor. We could detect if passphrase is disabled on Trezor (and eventually show this in error message), but currently we cannot say if wallet was created with or without passphrase. I also don't see way to detect this. If user is entering passphrase on Trezor, we have no information about it. We could ask if user wants to use a "hidden wallet" when creating it and mark it to the wallet file. But other than that we have only very buggy heuristic. Same for existing wallets. So for sake of clarity and consistency I would maybe refrain for asking users whether to use passphrase or not during wallet creation.

What are your thoughts?

rating89us:
Two ideas for 'address doesn't match error' in PF disabled Trezor:

Idea 1: detect if Trezor device has PF protection disabled and when address doesn't match, display the following error message: "ERROR: couldn't open wallet. Device wallet does not match wallet address. If this Monero GUI wallet is protected by a passphrase, first you need to enable passphrase protection for your Trezor device on Trezor Suite."

Idea 2: temporarily enable passphrase protection on Monero GUI. User clicks to open the wallet -> address doesn't match -> Monero GUI temporarily enables PF protection on Trezor device -> PF dialog opens in Monero GUI -> when PF dialog is closed, Monero GUI disables PF protection on Trezor device.

ph4r05:
#1 I am not confident this is the way to go. Other contributors would have to overrule me :)

#2 this won't work. I am strictly against manipulating passphrase settings in Trezor. It's up to user to do this, we won't change Trezor settings. And nothig is atomic here, if app crashes in the middle of the process, passphrase stays enabled, this would be very confusing and can lead to fund loss and even more support tickets.

#3 this is quite a lot of work and I don't feel like implementing it now. Gains are too low to justify the API overhaul.

ph4r05:
@moneromooo-monero pls also tis one, if you find some spare minutes 🙏 , thanks!

ph4r05:
ping :)

selsta (approved):

PR #7794, opened 2021-07-15
cmake: test is a reserved keyword

selsta:
```
CMake Error at /home/ubuntu/build/wownero/build/CMakeFiles/CMakeTmp/CMakeLists.txt:7 (add_executable):
  The target name "test" is reserved or not valid for certain CMake features,
  such as generator expressions, and may result in undefined behavior.


CMake Error at cmake/FindCcache.cmake:55 (try_compile):
  Failed to configure test project build system.
  CMakeLists.txt:53 (include)
```


mj-xmr (approved):
Interesting. How did it work all the time with this error then?

selsta:
It's only reserved in older cmake versions.

PR #7797, opened 2021-07-16
CI: CCache in FreeBSD works

mj-xmr:
I've discovered why ccache wasn't working under FreeBSD: because the cross compiler command was not an executable file, but rather a script, pointing to the exe file with all the options, that it needs. CCache expects an executable file. The solution was to parse and segment the script. I have described it in a comment of the changed file.

Unfortunately I couldn't get the automated test to work, so I might need the author's ( @perfect-daemon ) help here, please. See line 80.

### DIFF
--- a/cmake/FindCcache.cmake
+++ b/cmake/FindCcache.cmake
@@ -41,6 +41,29 @@
 
 find_program(CCACHE_FOUND ccache)
 if (CCACHE_FOUND)
+	string(REGEX MATCH ".*depends/x86_64-unknown-freebsd" IS_FREEBSD_XCOMPILING "${CMAKE_CXX_COMPILER}")
+	if (IS_FREEBSD_XCOMPILING)
+		# During FreeBSD cross compilation the ccache fails, due to the "CMAKE_CXX_COMPILER" pointing actually to a script,
+		# rather than to an executable, which is not accepted by ccache.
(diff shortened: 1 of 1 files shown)


cmake/FindCcache.cmake
```diff
@@ -50,17 +70,21 @@ option (CCACHE "")
...
 	unset(TEST_PROJECT)
-	if (${RET})
+	if (${RET} OR IS_FREEBSD_XCOMPILING)
 		# Success
 		message(STATUS "Found usable ccache: ${CCACHE_FOUND}")
 		set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_FOUND}")
 		set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK    "${CCACHE_FOUND}")
 	else()
 		message(STATUS "Found ccache ${CCACHE_FOUND}, but is UNUSABLE! Return code: ${RET}")
+		message(FATAL_ERROR "Found ccache ${CCACHE_FOUND}, but is UNUSABLE! Return code: ${RET}")
```
mj-xmr, line 87: An error. To be removed.

mj-xmr:
Converted to draft for now, as the presented imperfect approach has inspired developer `wfaressuissa` to find a better solution.

mj-xmr:
Superseded by https://github.com/monero-project/monero/pull/7797

PR #7809, opened 2021-07-27
p2p: remove blocked addresses/hosts from peerlist

moneromooo-monero:


perfect-daemon (approved):
correct

selsta (approved):
Tested

src/p2p/net_node.inl
```diff
@@ -83,6 +83,19 @@ static inline boost::asio::ip::address_v4 make_address_v4_from_v6(const boost::a
   return boost::asio::ip::address_v4(v4);
 }
 
+static inline uint16_t get_port(const epee::net_utils::network_address &na)
```
vtnerd, line 86: Looks like this should be a function of `network_address`.

* Add `virtual std::uint16_t port() const = 0;` [here](https://github.com/monero-project/monero/blob/master/contrib/epee/include/net/net_utils_base.h#L232) to declare the interface.
* Add `virtual std::uint16_t port() const override { return value.port(); }` [here](https://github.com/monero-project/monero/blob/master/contrib/epee/include/net/net_utils_base.h#L262) to bind every instance of `network_address` to the interface.
* Add `std::uint16_t port() const { return self ? self->port() : 0 }` [here](https://github.com/monero-project/monero/blob/master/contrib/epee/include/net/net_utils_base.h#L308) to handle the wrapper (which handles the case where the `network_address` was never set).

Then just call `na.port()` everywhere.

src/p2p/net_node.inl
```diff
@@ -332,6 +353,13 @@ namespace nodetool
       for (const auto &c: conns)
         zone.second.m_net_server.get_config_object().close(c);
 
+      for (int i = 0; i < 2; ++i)
+        zone.second.m_peerlist.filter(i == 0, [&subnet](const peerlist_entry &pe){
```
vtnerd, line 357: How big do these lists get? Can we do a `2*log(n)` range search then erase? This should be possible because the iteration key is sorted. The one below is going to be more complicated though.

vtnerd:
This also fails the functional_tests, but I don't know if it was caused by this patch or just unrelated noise.

selsta:
Seems unrelated.

src/p2p/net_node.inl
cirocosta: https://github.com/monero-project/monero/blob/ed506006d2563c474e52db5a91ba2e20af86f30f/src/cryptonote_config.h#L130-L131

moneromooo-monero: Subnets are only banned at startup atm anyway. I doubt it matters much in practice.

selsta (approved):

PR #7813, opened 2021-07-27
blocks: fix cmake syntax

selsta:
Old cmake versions dislike having no space.


perfect-daemon (approved):
correct

ndorf (approved):
This fixes the build with cmake 3.17.2 (OpenBSD 6.8).

PR #7814, opened 2021-07-27
cmake: don't forbid undefined symbols on openbsd

selsta:
Fails to compile with undefined symbols, possibly related: https://github.com/mesonbuild/meson/issues/3570#issuecomment-405459430 (linked comment and the following)


perfect-daemon (approved):
correct

ndorf (approved):
This fixes the libminiupnpc link errors on OpenBSD 6.8.

PR #7819, opened 2021-07-29
Fee changes from ArticMine

moneromooo-monero:


moneromooo-monero:
For the record, it conflicts with both the BP+ patch and the triptych patch.

erciccione:
Attaching the PDF linked in the commit to my comment, so that it will be available in this PR in case ArticMine's repository is deleted/moved/whatever. It's good to have references directly on the PR.
[MoneroScaling2021-02.pdf](https://github.com/monero-project/monero/files/6906846/MoneroScaling2021-02.pdf)

SamsungGalaxyPlayer:
Allowing the maximum block size to continuously grow at ~32x per year seems overly generous to me. I would rather lower the value from 2 to 1.7 max. This will still allow substantial maximum growth of 14x per year.

In cases that we exceed such a huge growth rate, it will be dealt with a higher fee market, incentives to use slightly more efficient transactions, and incentives to use some off-chain transfer method (such as a wrapped asset). We will still have an enormously generous growth rate which will provide substantial pressures to keep transactions costs down as available block size space grows.

I don't see temporary situations of full blocks as a failure case. This is a natural situation in other blockchains like Bitcoin and Ethereum, and providing a generous growth rate as a form of escape valve is more than sufficient imo.

Reference: https://github.com/monero-project/research-lab/issues/70

ArticMine:
I commented on this in https://github.com/monero-project/research-lab/issues/70. 

> With respect to the cap I considered 1.6 / 0.625, 2 / 0.5 and 2.5 / 0.4. I considered transaction rate changes not only in Monero but also Bitcoin (before the limit was reached), Litecoin, Dogecoin, Dash, and Bitcoin Cash. Here is a chart for Monero, Bitcoin and Litecoin. https://bitinfocharts.com/comparison/transactions-btc-ltc-xmr.html#log. Increases of 8x over a six month period and even 80x over a 1 year period have occurred in the past and can occur in the future. This is especially the case if there is a disruption in a competing coin as occurred in 2017 with Bitcoin reaching a hard limit triggering an 80x increase in Litecoin transactions. Setting this cap below 2 does run the real risk of triggering this issue if one allows the short term median to remain above the long term median for many months. On the other hand increasing the limit to say 2.5 / 0.4 would lead to a much larger spread in fees. There is also little evidence for uncontrolled growth over multiple years even in chains with no penalty to increase the blocksize below the hard cap such as Litecoin. One can take a look at Litecoin transactions in 2018 or Monero transactions in early 2017 for example. Also it is best to use figures where the reciprocal can be expressed as an exact decimal. 2 / 0.5 vs say 1.7 / 0.58823529... .

Since I made the above comment there have been developments in Dogecoin that illustrate how a sharp increase in fees can have a very significant negative impact in adoption https://bitinfocharts.com/comparison/transactionfees-transactions-doge.html#1y Note how the fall in transactions per day followed the sharp increase in fees to a level  well below that before the fee increase. Furthermore just recently the daily transactions per day of Dogecoin have fallen significantly below that of Monero. This has never happened before. https://bitinfocharts.com/comparison/transactions-doge-xmr.html#log. I will note that is Dogecoin full blocks was not the issue, rather it was a very high minimum fee cause by an increase in price of ~200x over a period of a few months. This is a real risk in Monero if we lower the growth of the long term median below 2. 

I do see temporary sharp increases in fees as a very significant failure. In fact this is what research-lab issue #70 is about. I also consider that such sharp increases in fees can have long lasting adverse impacts that can be very difficult to foresee. Bitcoin has chosen high fees and constrained block sizes because of the eventual need to replace the block rewards with transaction fees. The result has been the failure of Bitcoin as a "A Peer-to-Peer Electronic Cash System". The fact that the Bitcoin community has very successfully re-positioned Bitcoin as a store of value does negate Bitcoin failure as a transactional currency. The critical difference here is that Monero has a tail emission so competitive fees do not mean a loss in POW security.

SamsungGalaxyPlayer:
I think we simply disagree on this point. I see DOGE's increase as mostly temporary, which Monero would also be somewhat susceptible to. This is a cap on the long-term growth cap, which seems aggressively permissive. DOGE's network also became extremely unreliable for other reasons, including antiquated code and weak node infrastructure. The increase in transactions and their accompanying sizes certainly didn't help matters.

I would like to hear other comments on this point. Should we allow Monero's block size to increase 32 times per year with this proposal. This seems reckless to me.

ArticMine:
@SamsungGalaxyPlayer

I guess we disagree. The recent events with DOGE is just one of many examples. Furthermore focusing on the annual rate is also very arbitrary, since the rapid raises can occur over much serious time frames typically a 2 - 6 month period. The current limit of 1.4 for the long term median means a limit of less than 2 over 4.5 months. This has been exceeded on a regular basis not only occurred in Monero in the past in the but also is many other coins. 

I suggest taking a close look at the historical data not only in XMR

https://bitinfocharts.com/comparison/monero-transactions.html#log

Note: the sustained rate of ~2.4x a year going back to May 2016, and the significant volatility in transaction rates. The current rate (1.4) provides ~5.4x over a year. A rate of 1.7 provides 14.2x a year. On the surface this sounds like it might work until one factors in the current ~2.4x long term rate. This leaves under 6x a year. The simply does not work to cover historical fluctuations let alone an external event such as COVID-19 or a serious problem in a competing coin. With 2 we would have a net rate of  ~13.3 which would at least cover most historical fluctuations and at least mitigate external events. I would consider 2 to be the absolute minimum. There is actually a case for 2.5 even with the much higher fee spreads.

Here are additional charts that show how anything under 2 could under many scenarios cause serious problems. 

BTC 

https://bitinfocharts.com/comparison/transactions-btc-xmr.html#log

For BTC the perioud before 2013 is very relevant since there was no constraint on the blocksize

LTC

https://bitinfocharts.com/comparison/transactions-ltc-xmr.html#log

BCH

https://bitinfocharts.com/comparison/transactions-bch-xmr.html#log

I will note that artificially inflating fees by allowing the short term median to remain well above the short term median for an extended period of time opens the door to among other things the miners accepting transactions directly on a website providing a lower price in exchange for personal information. 

We must keep in mind that the long term median was introduced to deter spam attacks such as Flood XMR by forcing to sustain the short term median for over 2 months while the long term median catches up. It was never intended to put an arbitrary low cap over a period of 3 months to a year.

src/cryptonote_config.h
```diff
@@ -182,8 +182,10 @@
 #define HF_VERSION_EXACT_COINBASE               13
 #define HF_VERSION_CLSAG                        13
 #define HF_VERSION_DETERMINISTIC_UNLOCK_TIME    13
+#define HF_VERSION_2021_SCALING                 17
 
 #define PER_KB_FEE_QUANTIZATION_DECIMALS        8
+#define CRYPTONOTE_SCALING_2021_FEE_ROUNDING_PLACES 3
```
UkoeHB, line 188: Originally I recommended 2 significant digits. Has there been any discussion around this, reasoning for 3 vs 2?

src/cryptonote_core/blockchain.cpp
```diff
@@ -3651,8 +3651,19 @@ uint64_t Blockchain::get_dynamic_base_fee(uint64_t block_reward, size_t median_b
...
+      lo -= lo / 20;
+      lo = cryptonote::round_money(lo, CRYPTONOTE_SCALING_2021_FEE_ROUNDING_PLACES);
+      return lo;
+    }
+    else
+    {
+      lo /= 5;
+      return lo;
+    }
   }
```
UkoeHB, line 3667: Technically correct for the existing return value of `get_min_block_weight()`, but will become incorrect if that return value changes.

Fix:

```
    lo = mul128(block_reward, DYNAMIC_FEE_REFERENCE_TRANSACTION_WEIGHT, &hi);
    div128_64(hi, lo, median_block_weight, &hi, &lo, NULL, NULL);

    if (version >= HF_VERSION_2021_SCALING)
    {
      // min_fee_per_byte = 0.95 * block_reward * ref_weight / (fee_median^2)
      // note: fee_median computation changed at hardfork HF_VERSION_2021_SCALING
      div128_64(hi, lo, median_block_weight, &hi, &lo, NULL, NULL);
      assert(hi == 0);
      lo -= lo / 20;
      lo = cryptonote::round_money(lo, CRYPTONOTE_SCALING_2021_FEE_ROUNDING_PLACES);
      return lo;
    }
    else
    {
      // min_fee_per_byte = 0.2 * block_reward * ref_weight / (min_penalty_free_zone * fee_median)
      div128_64(hi, lo, min_block_weight, &hi, &lo, NULL, NULL);
      assert(hi == 0);
      lo /= 5;
      return lo;
    }
  }
```

I can't extend the Github selection zone to higher lines... annoying.

UkoeHB: The fee update research did not recommend rounding the minimum fee from the node's point of view - only the transaction author. However, I think it is fine to round the node's minimum fee, since it will enforce rounding on nodes to some extent.

UkoeHB (changes requested):
Review part 1

src/cryptonote_core/blockchain.cpp
```diff
@@ -4471,7 +4571,11 @@ bool Blockchain::update_next_cumulative_weight_limit(uint64_t *long_term_effecti
     get_last_n_blocks_weights(weights, CRYPTONOTE_REWARD_BLOCKS_WINDOW);
 
     uint64_t short_term_median = epee::misc_utils::median(weights);
-    uint64_t effective_median_block_weight = std::min<uint64_t>(std::max<uint64_t>(CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_V5, short_term_median), CRYPTONOTE_SHORT_TERM_BLOCK_WEIGHT_SURGE_FACTOR * m_long_term_effective_median_block_weight);
+    uint64_t effective_median_block_weight;
+    if (hf_version >= HF_VERSION_2021_SCALING)
+      effective_median_block_weight = std::min<uint64_t>(std::max<uint64_t>(long_term_median, short_term_median), CRYPTONOTE_SHORT_TERM_BLOCK_WEIGHT_SURGE_FACTOR * m_long_term_effective_median_block_weight);
+    else
+      effective_median_block_weight = std::min<uint64_t>(std::max<uint64_t>(CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_V5, short_term_median), CRYPTONOTE_SHORT_TERM_BLOCK_WEIGHT_SURGE_FACTOR * m_long_term_effective_median_block_weight);
```
UkoeHB, line 4578:
```suggestion
    if (hf_version >= HF_VERSION_2021_SCALING)
      // effective median = short_term_median bounded to range [long_term_median, 50*long_term_median], but it can't be smaller than the
      // minimum penalty free zone (a.k.a. 'full reward zone')
      effective_median_block_weight = std::min<uint64_t>(std::max<uint64_t>(long_term_median, short_term_median), CRYPTONOTE_SHORT_TERM_BLOCK_WEIGHT_SURGE_FACTOR * m_long_term_effective_median_block_weight);
    else
      // effective median = short_term_median bounded to range [0, 50*long_term_median], but it can't be smaller than the
      // minimum penalty free zone (a.k.a. 'full reward zone')
      effective_median_block_weight = std::min<uint64_t>(std::max<uint64_t>(CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_V5, short_term_median), CRYPTONOTE_SHORT_TERM_BLOCK_WEIGHT_SURGE_FACTOR * m_long_term_effective_median_block_weight);
```

Adds comments.

UkoeHB: @ArticMine

Is the computation of `M_s` recursive over `M_l`, or is it supposed to be `max(M_l, median(most recent 100 block weights)`? The latter is how it was implemented here, which seems fine to me.

src/cryptonote_core/blockchain.cpp
```diff
@@ -4450,7 +4547,10 @@ bool Blockchain::update_next_cumulative_weight_limit(uint64_t *long_term_effecti
       long_term_median = get_long_term_block_weight_median(db_height - nblocks - 1, nblocks);
     }
 
+    const uint64_t previous_long_term_effective_median_block_weight = m_long_term_effective_median_block_weight;
     m_long_term_effective_median_block_weight = std::max<uint64_t>(CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_V5, long_term_median);
+    if (hf_version >= HF_VERSION_2021_SCALING)
+      m_long_term_effective_median_block_weight = std::max<uint64_t>(m_long_term_effective_median_block_weight, previous_long_term_effective_median_block_weight / 2);
```
UkoeHB, line 4554: I don't think these changes belong here. As noted in another comment, it should be in `get_next_long_term_block_weight()`.

UkoeHB (changes requested):
Review part 2

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -1016,6 +1016,54 @@ namespace cryptonote
...
+        --ptr;
+        if (*ptr == '9')
+          *ptr = '0';
+        else
+        {
+          ++*ptr;
+          bump = false;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
moneromooo-monero: What are you asking here ? The function rounds amounts, whatever their value. Your case should round to 1000 I think. I'll check it does.

src/cryptonote_core/blockchain.cpp
moneromooo-monero: At first glance, a get_next_... function should not modify a member var. Is it actually incorrect if it's here ?

src/cryptonote_core/blockchain.cpp
```diff
@@ -3724,6 +3735,80 @@ bool Blockchain::check_fee(size_t tx_weight, uint64_t fee) const
...
+  // we want Mlw = median of max((min(Mbw, 2Ml), Zm), Ml/2)
+  // Mbw: block weight for the last 99990 blocks, 0 for the next 10
+  // Ml: penalty free zone (dynamic), aka long_term_median, aka median of max((min(Mb, 2Ml), Zm), Ml/2)
+  // Zm: 300000 (minimum penalty free zone)
+  //
+  // So we copy the current rolling median state, add 10 (grace_blocks) zeroes to it, and get back Mlw
+
+  epee::misc_utils::rolling_median_t<uint64_t> rm = m_long_term_block_weights_cache_rolling_median;
+  for (size_t i = 0; i < grace_blocks; ++i)
+    rm.insert(0);
```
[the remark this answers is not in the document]
moneromooo-monero: The intent here is to have a fee level that's known to be good for grace_blocks blocks, and the worst case is 0 block size. So I think this is right.

src/cryptonote_core/blockchain.cpp
moneromooo-monero: Removing "lo /= median_block_weight;" is a typo, right ?

moneromooo-monero: And the other change seems wrong, since it causes the tests to fail (by much more than just the rounding change from 3 to 2 places) - unless the values from the PDF test cases were wrong in the first place.

src/cryptonote_basic/cryptonote_format_utils.cpp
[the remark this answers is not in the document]
moneromooo-monero: The rounding was borked when the carry reached the top, fixed now. I assume that's what you meant (though I don't understand the significance of the endianness comment).

src/cryptonote_core/blockchain.cpp
```diff
@@ -3724,6 +3735,80 @@ bool Blockchain::check_fee(size_t tx_weight, uint64_t fee) const
...
+  const uint64_t Fn = 4 * base_reward * /*Brlw*/ DYNAMIC_FEE_REFERENCE_TRANSACTION_WEIGHT / (Mfw * Mfw);
+
+  // const uint64_t Fm = 16 * base_reward * Br / Mfw; fold Br from above
+  const uint64_t Fm = 16 * base_reward * DYNAMIC_FEE_REFERENCE_TRANSACTION_WEIGHT / (CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_V5 * Mfw);
+
+  const uint64_t Fp = 2 * base_reward / Mnw;
+
+  // fold Br from above, move 4Fm in the max to decrease quantization effect
+  //const uint64_t Fh = 4 * Fm * std::max<uint64_t>(1, Mfw / (32 * DYNAMIC_FEE_REFERENCE_TRANSACTION_WEIGHT * Mnw / CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_V5));
+  const uint64_t Fh = std::max<uint64_t>(4 * Fm, 4 * Fm * Mfw / (32 * DYNAMIC_FEE_REFERENCE_TRANSACTION_WEIGHT * Mnw / CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_V5));
```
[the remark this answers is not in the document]
moneromooo-monero: That one is too close to integer overflow for comfort. It also doesn't simplify really, it just reorders.

src/wallet/wallet2.cpp
```diff
@@ -321,15 +321,15 @@ void do_prepare_file_names(const std::string& file_path, std::string& keys_file,
   mms_file = file_path + ".mms";
 }
 
-uint64_t calculate_fee(uint64_t fee_per_kb, size_t bytes, uint64_t fee_multiplier)
+uint64_t calculate_fee(uint64_t fee_per_kb, size_t bytes)
```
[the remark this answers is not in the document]
moneromooo-monero: It is, it's in an anonymous namespace.

src/cryptonote_core/blockchain.cpp
UkoeHB: In your code,

```
div128_64(hi, lo, min_block_weight, &hi, &lo, NULL, NULL);
```

Is canceled out by 

```
lo *= CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_V5;
```

Why cancel it at all? My suggestion is to skip these calculations because they are zero sum. I don't know why the tests would fail.

> Removing "lo /= median_block_weight;" is a typo, right ?

It is replaced with this, which should be equivalent.

```
div128_64(hi, lo, median_block_weight, &hi, &lo, NULL, NULL);
```

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
UkoeHB: > the worst case is 0 block size

```
max(min(0, 2Ml), Zm, Ml/2) = max(Zm, Ml/2)
```

The `max()` calculation is in the next line (3790). A fee level that is good for grace_blocks blocks therefore should use recursive over `Ml/2` like in my suggestion.

src/cryptonote_core/blockchain.cpp
UkoeHB: I did not recommend modifying a member var in `get_next_long_term_block_weight()`. I recommended moving this functionality into `get_next_long_term_block_weight()`, which is used to set `long_term_median` in line 4450 referenced by this comment, which is used to set `m_long_term_effective_median_block_weight` in line 4453 referenced by this comment.

src/cryptonote_core/blockchain.cpp
moneromooo-monero: nvm, I had missed another change in the block. Anyway, I've done this except the removal of these two "canceling" lines since it's not obvious they cancel and it's not like an extra div here really slows it down.

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
moneromooo-monero: I do not understand that, and since it works, you can change it later if you think it should be.

src/cryptonote_core/blockchain.cpp
moneromooo-monero: If you mean moving the entire calc there, IIRC it's called a lot and would noticeably slow down, so I'm leaving it here.

src/cryptonote_core/blockchain.cpp
UkoeHB: > I've done this except the removal of these two "canceling" lines since it's not obvious they cancel and it's not like an extra div here really slows it down.

I am confused by this reply. Where in the fee doc does it specify making these computations?

src/cryptonote_core/blockchain.cpp
UkoeHB: Sorry, what? A call to `std::max<uint64_t>()` will "noticeably slow [it] down"?

UkoeHB:
```suggestion

    m_long_term_effective_median_block_weight = std::max<uint64_t>(CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_V5, long_term_median);

```

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
UkoeHB: The long term median is recursive over `median of max((min(Mb, 2Ml), Zm), Ml/2)`. If your block weight is `Mb = 0`, then it simplifies to recursive over `max(Zm, Ml/2)`. It is incorrect to push back 0s.

It is not a matter of what I think it 'should' be.

src/cryptonote_core/blockchain.cpp
UkoeHB: After your update to `get_next_long_term_block_weight()` these lines don't do anything any more.

src/cryptonote_core/blockchain.cpp
```diff
@@ -3648,11 +3648,25 @@ uint64_t Blockchain::get_dynamic_base_fee(uint64_t block_reward, size_t median_b
...
+    {
+      // min_fee_per_byte = round_up( 0.95 * block_reward * ref_weight / (fee_median^2) )
+      // note: since hardfork HF_VERSION_2021_SCALING, fee_median (a.k.a. median_block_weight) equals effective long term median
+      div128_64(hi, lo, median_block_weight, &hi, &lo, NULL, NULL);
+      assert(hi == 0);
+      lo *= CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_V5;
+      lo /= median_block_weight;
+      lo -= lo / 20;
+      return lo;
+    }
```
UkoeHB, line 3662:
```suggestion
    if (version >= HF_VERSION_2021_SCALING)
    {
      // min_fee_per_byte = round_up( 0.95 * block_reward * ref_weight / (fee_median^2) )
      // note: since hardfork HF_VERSION_2021_SCALING, fee_median (a.k.a. median_block_weight) equals effective long term median
      div128_64(hi, lo, median_block_weight, &hi, &lo, NULL, NULL);
      assert(hi == 0);
      lo -= lo / 20;
      return lo;
    }
```

I am confused why you want these two lines in this function. Am I missing something in the fee doc?

moneromooo-monero: Are they not in the doc ? I assume they are if I ended up with that code. I'll double check later and remove if I don't find them and it still works with them out.

moneromooo-monero: I looked and I cannot see that. I have no idea why I put those in. Removed. I'll push once tests have run.

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
moneromooo-monero: I'm unconvinced and it looks like playing on words in that last comment. I'll keep it as is.

src/cryptonote_core/blockchain.cpp
moneromooo-monero: The fee doc does not specify those.

src/cryptonote_core/blockchain.cpp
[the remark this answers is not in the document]
UkoeHB: @ArticMine @vtnerd any thoughts on this? It is the last point before I can accept this PR.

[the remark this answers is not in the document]
ArticMine: I believe that the grace blocks should 0 as @moneromooo-monero is proposing. This is for clarity, robustness and simplicity of the code. The principle here is that the grace blocks are a worst case scenario. One then proceeds to calculate the medians.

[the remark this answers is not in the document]
UkoeHB: @ArticMine This function copies the ledger's rolling long term median variable, then inserts 0 at the end for `num_graceblocks`. It doesn't affect consensus, if that was your concern. I may be misunderstanding your proposal though.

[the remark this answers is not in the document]
ArticMine: > This function copies the ledger's rolling long term median variable, then inserts 0 at the end for num_graceblocks. It doesn't affect consensus, if that was your concern. I may be misunderstanding your proposal though.

That is my understanding also. What I am missing is why we have recursion after inserting the 0 at the end for num_graceblocks. For the next block we repeat the process starting again with the rolling long term median variable.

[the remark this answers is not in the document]
ArticMine: Yes to clarify. There is no recursion in the PR code for the wallet fees using 0, but if I understand this correctly we use max(Zm, Ml/2) we would use recursion for 10 blocks each time to calculate the wallet fee median from the consensus long term median.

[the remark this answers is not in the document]
UkoeHB: Ok since @ArticMine thinks this is fine, I will accept it.

tests/unit_tests/scaling_2021.cpp
```diff
@@ -0,0 +1,183 @@
...
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#define IN_UNIT_TESTS
+
+#include "gtest/gtest.h"
```
UkoeHB, line 31:
```suggestion
// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.

////
// References:
// - https://github.com/ArticMine/Monero-Documents/blob/master/MoneroScaling2021.pdf
// - https://github.com/monero-project/research-lab/issues/70
///

#define IN_UNIT_TESTS

#include "gtest/gtest.h"
```

UkoeHB (changes requested):

src/cryptonote_config.h
```diff
@@ -182,8 +182,10 @@
 #define HF_VERSION_EXACT_COINBASE               13
 #define HF_VERSION_CLSAG                        13
 #define HF_VERSION_DETERMINISTIC_UNLOCK_TIME    13
+#define HF_VERSION_2021_SCALING                 17
 
 #define PER_KB_FEE_QUANTIZATION_DECIMALS        8
+#define CRYPTONOTE_SCALING_2021_FEE_ROUNDING_PLACES 2
```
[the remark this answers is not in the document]
moneromooo-monero: Someone clearly thought so :) AFAICT this is an attempt at having new constants be constants (as opposed to macros) in namespaces.  I prefer adding things in the groups where they belong. Others may add elsewhere.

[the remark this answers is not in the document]
UkoeHB: I see

UkoeHB (approved):
Thank you

ArticMine:
I am proposing changing the rate of growth of ML from 2 to 1.7 as per https://github.com/monero-project/research-lab/issues/70#issuecomment-1025334284

moneromooo-monero:
I think that's all that's needed to switch the patch to 1.7. I'll squash before it gets merged.

src/cryptonote_core/blockchain.cpp
```diff
@@ -3762,6 +3774,83 @@ bool Blockchain::check_fee(size_t tx_weight, uint64_t fee) const
...
+  fees[2] = cryptonote::round_money_up(Fm, CRYPTONOTE_SCALING_2021_FEE_ROUNDING_PLACES);
+  fees[3] = cryptonote::round_money_up(Fh, CRYPTONOTE_SCALING_2021_FEE_ROUNDING_PLACES);
+}
+
+void Blockchain::get_dynamic_base_fee_estimate_2021_scaling(uint64_t grace_blocks, std::vector<uint64_t> &fees) const
+{
+  const uint8_t version = get_current_hard_fork_version();
+  const uint64_t db_height = m_db->height();
+  const uint64_t nblocks = std::min<uint64_t>(m_long_term_block_weights_window, db_height);
+  const uint64_t long_term_median = get_long_term_block_weight_median(db_height - nblocks, nblocks);
```
j-berman, line 3818: Unused, and doesn't seem like it matters but I can't tell for certain

src/cryptonote_core/blockchain.cpp
```diff
@@ -3686,11 +3686,23 @@ uint64_t Blockchain::get_dynamic_base_fee(uint64_t block_reward, size_t median_b
...
+      div128_64(hi, lo, median_block_weight, &hi, &lo, NULL, NULL);
+      assert(hi == 0);
+      lo -= lo / 20;
+      return lo;
+    }
+    else
+    {
+      // min_fee_per_byte = 0.2 * block_reward * ref_weight / (min_penalty_free_zone * fee_median)
+      div128_64(hi, lo, min_block_weight, &hi, &lo, NULL, NULL);
+      lo /= 5;
```
j-berman, line 3703: missing the `assert(hi == 0)` from above?

src/cryptonote_config.h
```diff
@@ -182,8 +182,10 @@
 #define HF_VERSION_EXACT_COINBASE               13
 #define HF_VERSION_CLSAG                        13
 #define HF_VERSION_DETERMINISTIC_UNLOCK_TIME    13
+#define HF_VERSION_2021_SCALING                 17
```
j-berman, line 185: 15?

src/cryptonote_core/blockchain.cpp
```diff
@@ -3762,6 +3774,83 @@ bool Blockchain::check_fee(size_t tx_weight, uint64_t fee) const
...
+
+  uint64_t already_generated_coins = db_height ? m_db->get_block_already_generated_coins(db_height - 1) : 0;
+  uint64_t base_reward;
+  if (!get_block_reward(m_current_block_cumul_weight_limit / 2, 1, already_generated_coins, base_reward, version))
+  {
+    MERROR("Failed to determine block reward, using placeholder " << print_money(BLOCK_REWARD_OVERESTIMATE) << " as a high bound");
+    base_reward = BLOCK_REWARD_OVERESTIMATE;
+  }
+
+  return get_dynamic_base_fee_estimate_2021_scaling(grace_blocks, base_reward, Mnw, Mlw_penalty_free_zone_for_wallet, fees);
```
j-berman, line 3851: also surprised the compiler doesn't catch returning a value on a function declared void?

moneromooo-monero: That call is also void.

src/cryptonote_core/blockchain.cpp
```diff
@@ -4448,6 +4544,7 @@ bool Blockchain::check_blockchain_pruning()
   return m_db->check_pruning();
 }
 //------------------------------------------------------------------
+// returns min(Mb, 2Ml) as per https://github.com/ArticMine/Monero-Documents/blob/master/MoneroScaling2021-02.pdf from HF_VERSION_LONG_TERM_BLOCK_WEIGHT
```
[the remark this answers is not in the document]
moneromooo-monero: AFAICT it's returning min(Mb, 1.7 x) with x being max(long term median, 300000). and seems to be assuming that median(max(x,y)) == max(median(x),y), which I think is true (a test program I've just made seems to corroborate). So I'm not sure why you think this functoin should return something else. Do you disagree with how it's used maybe ?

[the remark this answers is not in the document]
j-berman: No I think it's ok given how it's used and that this is ok. Was just a little confused by it since it seemed a little different and wanted to double check

j-berman (approved):

src/cryptonote_core/blockchain.cpp
```diff
@@ -3762,6 +3775,82 @@ bool Blockchain::check_fee(size_t tx_weight, uint64_t fee) const
...
+  fees[1] = cryptonote::round_money_up(Fn, CRYPTONOTE_SCALING_2021_FEE_ROUNDING_PLACES);
+  fees[2] = cryptonote::round_money_up(Fm, CRYPTONOTE_SCALING_2021_FEE_ROUNDING_PLACES);
+  fees[3] = cryptonote::round_money_up(Fh, CRYPTONOTE_SCALING_2021_FEE_ROUNDING_PLACES);
+}
+
+void Blockchain::get_dynamic_base_fee_estimate_2021_scaling(uint64_t grace_blocks, std::vector<uint64_t> &fees) const
+{
+  const uint8_t version = get_current_hard_fork_version();
+  const uint64_t db_height = m_db->height();
+  const uint64_t nblocks = std::min<uint64_t>(m_long_term_block_weights_window, db_height);
```
j-berman, line 3818: sorry I hate to do this but this is also unused. was just going to make a new PR for this, but figure makes sense to double check this is safe by you to remove now. Everything looks good to me

src/wallet/wallet2.cpp
```diff
@@ -7579,6 +7579,40 @@ uint64_t wallet2::get_base_fee()
...
 }
 //----------------------------------------------------------------------------------------------------
+uint64_t wallet2::get_base_fee(uint32_t priority)
+{
+  // clamp and map to 0..3 indices, mapping 0 (default, but should not end up here) to 0, and 1..4 to 0..3
+  if (priority == 0)
+    priority = 1;
+  else if (priority > 4)
+    priority = 4;
+  --priority;
```
j-berman, line 7589: Think this general logic (starting from the comment) belongs inside `if (use_2021_scaling)`, otherwise this causes a discrepancy with current v17.3.0 clients until the hard fork (`get_fee_multiplier(priority)` gets called with a value of `priority` 1 less than it should be called with down in the else)

src/wallet/node_rpc_proxy.cpp
```diff
@@ -238,13 +239,24 @@ boost::optional<std::string> NodeRPCProxy::get_dynamic_base_fee_estimate(uint64_
     m_dynamic_base_fee_estimate = resp_t.fee;
     m_dynamic_base_fee_estimate_cached_height = height;
     m_dynamic_base_fee_estimate_grace_blocks = grace_blocks;
+    m_dynamic_base_fee_estimate_vector = std::move(resp_t.fees);
```
j-berman, line 242: Suggestion:

```
m_dynamic_base_fee_estimate_vector = !resp_t.fees.empty() ? std::move(resp_t.fees) : std::vector<uint64_t>{m_dynamic_base_fee_estimate};
```

Fixes a seg fault when wallet tries to construct a transaction before the fork (tries to access `fees[0]` on an empty vector in `get_dynamic_base_fee_estimate` below)

tests/functional_tests/blockchain.py
```diff
@@ -80,10 +80,10 @@ def _test_generateblocks(self, blocks):
         assert ok
 
         res = daemon.get_fee_estimate()
-        assert res.fee == 234562
+        assert res.fee == 1200000
```
j-berman, line 83: I can't see why yet but this is causing a test to fail. Working through it

j-berman (commented):
Ran into a few more things testing

Also can get rid of this: https://github.com/moneromooo-monero/bitmonero/blob/a3316d785c8cff7816af2435378165c58f207815/src/wallet/api/wallet.cpp#L1764

tests/functional_tests/blockchain.py
moneromooo-monero: Probably because it's not on v15 yet. You need to add that fork to the table.

j-berman: Yep, adding to the end of the mainnet table does the trick. Added it in the wrong way and was getting a weird result

j-berman (approved):

UkoeHB (approved):

UkoeHB (approved):

Issue #7807, opened 2021-07-27
wallet2 decoy selection algorithm ignores very recent outputs


j-berman:
UPDATE: it was discovered that MyMonero has been using a [separate implementation](https://github.com/vtnerd/monero-lws/blob/faa51780f3f8e6c5c0c4235499b95c246e074f29/src/util/gamma_picker.cpp) of the decoy selection algorithm that *did not have the bug*. Therefore, MyMonero users could have feasibly constructed transactions that selected newly spendable outputs as decoys, which means newly spent outputs observed on chain are not guaranteed identifiable as real outputs.

## Overview

The decoy selection algorithm in wallet2 has next to 0 chance of selecting extremely recent outputs as decoys. Today, if a user spends an output right in the block that it unlocks, and the output was originally created in a block that has fewer than 100 outputs total in it, their real output would be *more clearly identifiable in the ring (edit: outputs are not guaranteed identifiable primarily because of MyMonero's existing popular alternative algorithm in use). For reference, the trailing yearly average is around 63 outputs per block. ~Therefore, outputs that are spent immediately when they unlock are likely identifiable in rings today.~

To be clear, I have run this by core devs before sharing publicly.

## Technical explanation

It helps to explain the entire intuition behind the decoy selection algorithm, and provide some charts to explain the bug. Apologies if some of this is rudimentary, and also apologies if I got some of this wrong.

The decoy selection algorithm selects 10 decoys to include in a ring, and hides the real output among them using a criteria that best hides the real output. It does this by selecting decoys based on the expected time of when outputs are spent.

In the paper by [Miller et al](https://arxiv.org/pdf/1704.04299/) (section 6.1), they recommended using a particular distribution known as a gamma distribution to select decoys that correctly mirrors real world spending patterns observed on Monero back when Monero outputs could be linked on-chain. Here is what the gamma distribution recommended in the paper looks like:

![exp(x)](https://user-images.githubusercontent.com/26468430/127079746-4060a09f-2a7f-4cfc-9675-39f8e0913e0a.png)

[...]

PR #7821, opened 2021-07-30
wallet2: apply gamma distribution from chain tip when selecting decoys (#7807)

j-berman:
## Overview

When the wallet selects decoys using the gamma distribution, the expected distribution is supposed to be fit to the chain tip, but instead is being fit prior to the unlock time. The fix in this PR simply shifts the result of the gamma forward the expected duration of the unlock time. If the gamma spits out an output spent less than the unlock time, then it places that output in a random block within the first 50 spendable blocks (while still factoring in block density). I assumed that outputs younger than the unlock time that the gamma suggests should still be factored into the distribution because one would expect outputs selected by the gamma in this range to be spent soon after unlock. "Someone who spent an output after 1 block back when it was allowed likely would spend that output soon after it unlocks today."

The material negative impact of this general issue still appears to be contained to the very earliest spents, however deeper analysis would be needed to arrive at harder figures.

## Reasoning behind gamma being fit to the start of the chain

The decoy selection algo is applying the gamma distribution starting [10 blocks prior to the chain tip](https://github.com/monero-project/monero/blob/de3456e1275836725291ba71036b7ef0e2cda91f/src/wallet/wallet2.cpp#L1018-L1034) (bc of unlock time), but it appears the gamma distribution should be applied starting at the chain tip.

As mentioned in "The fix" of #7807 , this was my first thought as to the issue, but I figured it may have been incorrect because the gamma suggests that there is a non-negligible chance of spending an output between 1 and 9 blocks old, which I assumed had no chance of being plausible because of the unlock time, and so there must be some other explanation for the issue. But it seems the gamma distribution in the paper was taken from a time when the unlock time [wasn't enforced by consensus](https://monero.stackexchange.com/a/4588). And as a [talk done by isthmus demonstrated](https://www.youtube.com/watch?v=XIrqyxU3k5Q&t=1019s), some didn't follow this convention at the time. Therefore, there were *some* outputs on chain spent before 10 blocks that the gamma distribution factors in.

I reached out to one of the authors of the paper to sanity check, and [the above does seem to be the case]( https://github.com/maltemoeser/moneropaper/blob/master/Monero%20Analysis.ipynb). See the final section where it fits the chain data to the gamma, specifically:

```
def spendTimes():
    query = """
      MATCH (i:Input)<-[:TX_INPUT]-(tx:Transaction)-[:IN_BLOCK]->(b2:Block)
      WHERE i.mixin > 0 AND (i)-[:SPENDS]->()
      WITH i, b2.timestamp as ts
      MATCH (b1:Block)<-[:IN_BLOCK]-(:Transaction)-[:TX_OUTPUT]->(:Output)<-[:SPENDS]-(i)
      RETURN ID(i) as iId, ts - b1.timestamp as timeDiff
    """
    df = to_data_frame(graph.run(query))
    return df
 ```

## Analysis of the fix

See [this comment](https://github.com/monero-project/monero/pull/7821#issuecomment-898258800) for an explanation of how I arrived at using a window of 50 to place outputs spit out by the gamma more recent than the unlock time. It also provides analysis on the fix's impact.

# The analysis below is outdated, but keeping it here for posterity so the flow of discussion in comments below makes more sense

## Results of the fix

I used [this code](https://github.com/j-berman/monero/pull/2) to simulate `get_outs` with the fix in this PR, and plotted against the current:

10 block unlock fix vs  current  1

10 block unlock fix vs  current  2

You can see in the above charts that the very earliest outputs are selected in higher frequency with the fix. As you move further right, the outputs are selected in marginally lower frequency. Then it hits a steady state of selecting roughly equivalent outputs.

## The numbers explaining the above observation

The gamma's expected probability of a spent output between 0 to 10 blocks old is ~2.1%.

Between 1 to 10 blocks, it's still ~2.1% (since 0 to 1 block is negligible).

Between 10 to 11 blocks, it's ~0.3%.

Between 11 to 20 blocks, it's ~2.3%.

Between 20 to 30 blocks, it's ~2.2%.

With the fix, we would expect outputs less than < 10 blocks old produced by the gamma to be spent in the first available block; thus, we should expect 2.1% + 0.3% of outputs to be spent between 10 to 11 blocks. However, the current algorithm suggests that between 10 to 11 blocks, close to 0% of outputs would be selected as decoys (before factoring in density). Thus, between 10 to 11 blocks, it would seem the current algorithm under-selects decoys by about 2.4%.  This means that the outputs observed in this age range of 10 to 11 blocks are more likely to be real spents. In practice, however, thanks to the decoy selection algorithm factoring in density and therefore likely selecting *some* decoys in this age range, and considering we've only observed ~0.5% of outputs in this age range, it appears as though only a very small percentage of real spents would have been identifiable.

Between 11 to 20 blocks, the current algorithm is expected to match the gamma's 1 to 10 block range of 2.1%. And as noted, the gamma's expected probability of a spent output between 11 and 20 blocks is 2.3%. Thus, the current algorithm only slightly under-selects outputs in this age range, which means it is likely that nothing definitive can be gained from observing outputs in rings in this age range (ignoring the impact of block density).

Between 20 to 30 blocks, the current algorithm is expected to match the gamma's 10 to 20 block range of 0.3% + 2.3%. And as noted, for comparison, the gamma's expected is 2.2%. Therefore the current algorithm over-selects decoys in this interval, and thus, outputs spent in this range are overly protected, aka are even *less* likely to be deducible as real. The same goes for the range between 30 to 250 blocks.

Beyond 250 blocks, both current algorithm and fixed algorithm are roughly equivalent.

## Conclusion

As first reported, the impact appears *mostly* contained to the very earliest spents that are spent right when they unlock, and were created in relatively smaller blocks than average. The initial maximum estimate of 1% of transactions affected seems corroborated by the above findings. A deeper analysis factoring in block density would be necessary to arrive at harder figures.

## Appendix: impact on tx uniformity

As a result of the fix, approximately 1 in 5 rings `((1 - [1 - (2.1% + 0.3%)]^10) = 21% = 1 in 5)` are now expected to include at least 1 very early decoy. With this information, someone would likely be able to guess that a transaction is coming from a fixed wallet with a higher degree of certainty, however, I do not see a practical vulnerability that can stem from this knowledge.

Edit 1: [small change](https://github.com/monero-project/monero/pull/7821#issuecomment-889664618) to select unspendable output from first spendable block randomly
Edit 2: corrected charts for small change
Edit 3: updated for using a window of 50

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -144,6 +144,9 @@ using namespace cryptonote;
 
 #define IGNORE_LONG_PAYMENT_ID_FROM_BLOCK_VERSION 12
 
+#define DEFAULT_UNLOCK_TIME (CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE * DIFFICULTY_TARGET_V2)
+#define RECENT_SPEND_WINDOW (50 * DIFFICULTY_TARGET_V2)
+
 static const std::string MULTISIG_SIGNATURE_MAGIC = "SigMultisigPkV1";
 static const std::string MULTISIG_EXTRA_INFO_MAGIC = "MultisigxV1";
 
@@ -1028,6 +1031,34 @@ uint64_t gamma_picker::pick()
 {
   double x = gamma(engine);
   x = exp(x);
+
+  if (x > DEFAULT_UNLOCK_TIME) 
+  {
+    // We are trying to select an output from the chain that appeared 'x' seconds before the
+    // current chain tip, where 'x' is selected from the gamma distribution recommended in Miller et al.
+    // (https://arxiv.org/pdf/1704.04299/).
+    // Our method is to get the average time delta between outputs in the recent past, estimate the number of
+    // outputs 'n' that would have appeared between 'chain_tip - x' and 'chain_tip', select the real output at
+    // 'current_num_outputs - n', then randomly select an output from the block where that output appears.
+    // Source code to paper: https://github.com/maltemoeser/moneropaper
+    //
+    // Due to the 'default spendable age' mechanic in Monero, 'current_num_outputs' only contains
+    // currently *unlocked* outputs, which means the earliest output that can be selected is not at the chain tip!
+    // Therefore, we must offset 'x' so it matches up with the timing of the outputs being considered. We do
+    // this by saying if 'x` equals the expected age of the first unlocked output (compared to the current
+    // chain tip - i.e. DEFAULT_UNLOCK_TIME), then select the first unlocked output.
+    x -= DEFAULT_UNLOCK_TIME;
+  }
+  else 
+  {
+    // If the spent time suggested by the gamma is less than the unlock time, that means the gamma is suggesting an output
+    // that is no longer feasible to be spent (possible since the gamma was constructed when consensus rules did not enforce the
+    // lock time). The assumption made in this code is that an output expected spent quicker than the unlock time would likely
+    // be spent within RECENT_SPEND_WINDOW after allowed. So it returns an output that falls between 0 and the RECENT_SPEND_WINDOW.
+    // The RECENT_SPEND_WINDOW was determined with empirical analysis of observed data.
+    x = crypto::rand_idx(static_cast<uint64_t>(RECENT_SPEND_WINDOW));
+  }
+
   uint64_t output_index = x / average_output_time;
   if (output_index >= num_rct_outputs)
     return std::numeric_limits<uint64_t>::max(); // bad pick

### CODE REVIEW

**Review by UkoeHB — APPROVED** [review]

**Comment by j-berman** [comment]
@UkoeHB Made a small change for when the gamma spits out an output more recent than unlock time. In that case, it now selects a random value between 0 and 120, rather than scales down the gamma to between 0 and 120 because the scaling down approach biases toward 60 and 120. Don't see why it should have a bias like that, and it seems all outputs should actually be equally likely to be picked from that initial 0 to 120 should it fall into that else.

**Review by UkoeHB — APPROVED** [review]

**Review by moneromooo-monero — APPROVED** [review]

**Review by SChernykh — APPROVED** [review]

**Comment by j-berman** [comment]
While working through the impact of tx uniformity in #7798, I recognized a potential area of concern as a result of this PR's impact on tx uniformity. I don't think it's a reason to hold this PR back, but figured it's worth sharing to better understand this risk.

If you assume that 99% of clients update, and 1% do not, then the 1% non-updated clients may stick out.

Assume that a user constructs a transaction with 20 rings and not a single one has a very early output in it. The chances of an updated client doing that are very low (`[1 - 21%]^20 = 0.9%`), so it is safe to assume that if a tx is constructed in this way, it is from a non-updated client. Assume a user then spends outputs created in that tx in similar fashion (many rings, non-updated client). That would then be very strong evidence that the outputs in the ring coming from the original non-updated client tx are real.

Such a circumstance seems extraordinarily rare. I can, however, see it negatively harming a user who is very late to update their wallet. But I cannot see it impacting *other* users who do not update. And considering that current users are being harmed today by the implications of this issue, it needs to be patched for those users immediately.

**Comment by j-berman** [comment]
### Latest development

In monero-dev IRC, @luigi1111  suggested smoothing out the simulation of the decoy selection algorithm and plotting that, in order to get a better idea of how the patch would alter the algorithm. After graphing a smooth simulation of the decoy selection algorithm, I observed that my initial proposed fix alters the shape of the distribution such that it seems to perform marginally worse for outputs 15-100 blocks old:

patch  #7821 alone + window of 1

I went back to the drawing board and applied many different combinations of fixes to arrive at what I believe is the safest, do-no-harm approach, that patches the issue at hand, while simultaneously providing sufficient protection for the earliest spents.

### The current proposed patch

1. Apply the gamma from the chain tip, same as initially proposed.

2. When the gamma picks an output younger than the unlock time, replace that output with a randomly selected output from the 50 most recent spendable blocks.

The justification for continuing to factor in outputs younger than the unlock time picked by the gamma is still that people who spent outputs very quickly back when the gamma was observed, likely would still be spending outputs relatively quickly today. The justification for using a window of 50 to slot them in is that empirically it seems to perform well:

patch  #7821 alone + window of 50


### Math backing up the decision


As mentioned by a fellow by the username of Rucknium to me in a 1-on-1 IRC, as well as in Miller et al, the [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) is a test to quantify the distance between an observed distribution, and expected. As such, it can be used to quantify how well the proposed patch would perform compared to the current decoy selection algorithm. The lower the distance given by the Kolmogorov-Smirnov statistic, the better the algorithm is at matching the observed distribution.

If my math is correct (see below), the K-S statistic for the current decoy selection algorithm is 0.0167, while the K-S statistic of the current proposed fix is 0.0071. Thus, the patch is a material improvement over the current.

Finally, the K-S statistic of the patch I initially proposed (dumping outputs in the first spendable block) is 0.0190. Thus, the fix I'm proposing now is a material improvement over the patch I initially proposed.


### Math to recreate K-S Statistic

Download the [following CSV](https://github.com/monero-project/monero/files/6968268/output_age_data.zip), then run the following python script:

```
import csv

n_observed = 0
n_current_algo = 0
n_initial_proposed_fix = 0
n_current_proposed_fix = 0

cumulative_sum_observed = []
cumulative_sum_current_algo = []
cumulative_sum_initial_proposed_fix = []
cumulative_sum_current_proposed_fix = []

with open('output_age_data.csv','r') as csvfile:
    rows = csv.reader(csvfile, delimiter = ',')
    
    # get n outputs for each distribution type & cumulative sums
    idx = 0
    for row in rows:
        if idx > 0:
            observed         = int(row[1])
            current_algo     = int(row[2])
            initial_proposed = int(row[3])
            current_proposed = int(row[4])

            n_observed              += observed
            n_current_algo          += current_algo
            n_initial_proposed_fix  += initial_proposed
            n_current_proposed_fix  += current_proposed
            
            if idx == 1:
                cumulative_sum_observed.append(observed)
                cumulative_sum_current_algo.append(current_algo)
                cumulative_sum_initial_proposed_fix.append(initial_proposed)
                cumulative_sum_current_proposed_fix.append(current_proposed)
            else:
                cumulative_sum_observed.append(observed + cumulative_sum_observed[len(cumulative_sum_observed) - 1])
                cumulative_sum_current_algo.append(current_algo + cumulative_sum_current_algo[len(cumulative_sum_current_algo) - 1])
                cumulative_sum_initial_proposed_fix.append(initial_proposed + cumulative_sum_initial_proposed_fix[len(cumulative_sum_initial_proposed_fix) - 1])
                cumulative_sum_current_proposed_fix.append(current_proposed + cumulative_sum_current_proposed_fix[len(cumulative_sum_current_proposed_fix) - 1])
               
        idx += 1

cdf_observed = []
cdf_current_algo = []
cdf_initial_proposed_fix = []
cdf_current_proposed_fix = []

# iterate over cumulative sums and divide by sum to get F(x) for each block
idx = 0
for cumulative_observed in cumulative_sum_observed:
    cdf_observed.append(cumulative_sum_observed[idx] / n_observed)
    cdf_current_algo.append(cumulative_sum_current_algo[idx] / n_current_algo)
    cdf_initial_proposed_fix.append(cumulative_sum_initial_proposed_fix[idx] / n_initial_proposed_fix)
    cdf_current_proposed_fix.append(cumulative_sum_current_proposed_fix[idx] / n_current_proposed_fix)

    idx += 1

# iterate over F(x)'s and get abs(F(observed) - F(x))
set_of_diffs_current_algo = []
set_of_diffs_initial_proposed_fix = []
set_of_diffs_current_proposed_fix = []

idx = 0
for f_of_observed in cdf_observed:
    set_of_diffs_current_algo.append(abs(f_of_observed - cdf_current_algo[idx]))
    set_of_diffs_initial_proposed_fix.append(abs(f_of_observed - cdf_initial_proposed_fix[idx]))
    set_of_diffs_current_proposed_fix.append(abs(f_of_observed - cdf_current_proposed_fix[idx]))

    idx += 1

# get the max of each
print("K-S statistic for current algorithm:    ", max(set_of_diffs_current_algo))
print("K-S statistic for initial proposed fix: ", max(set_of_diffs_initial_proposed_fix))
print("K-S statistic for current proposed fix: ", max(set_of_diffs_current_proposed_fix))
```

**Comment by j-berman** [comment]
## Latest Development

In monero-dev IRC, @luigi1111 requested to see what using a window of 20 would look like, and @Gingeropolous  requested to see simulations run over different epochs, as well as comparisons to the gamma distribution (ignoring block density).

- I re-ran the simulations using windows of 1, 3, 10, 20, 40, 50, 75, and 100 (I had previously looked at 1, 3, 40, 50, and 100 locally before settling on 50).

- I re-ran the analysis over different hard fork intervals using @neptuneresearch 's [suggestion as a guide](https://github.com/neptuneresearch/monero-ringmember-age-distribution#v11-ringct-ringmember-age-distribution-1788720---1978433).

- I included the gamma on each plot to allow the reader to form a basis of comparison.

After doing the above, I still lean towards the current window of 50 as best achieving the sanest, least-potential-for-harm approach that I described in IRC as follows:

> the idea of this fix is to get something out the door that does a good enough job to cover the earliest spent issue, and doesn't cause any potential harm. I can't imagine that getting [the decoy selection algorithm to select outputs that are] marginally closer to the [observed output age distribution on-chain] causes harm (and may help), but moving further away could hurt.

As discussed in IRC, beyond this PR, we could continue with deeper research toward an even stronger solution that takes another crack at the assumptions laid out in Miller et al, and factors in observed chain data since then. Said research will take a fair amount of time to complete, and this PR's do-no-harm approach offers a solid patch until that research is finished. Also shoutout to @Rucknium who has some excellent ideas and is an applied statistician by trade who has offered to contribute in this area :)

## Results simulating different windows, over different epochs

### v14 (2210720 - 2413735) 

different_windows_v14

##### First general takeaways from this chart

The current decoy selection algorithm (orange) appears to be under-selecting outputs relative to the observed output ages in rings (blue), as is apparent by the large triangular gap above the orange line to the blue. Basically, there are many more outputs observed on-chain than the current decoy selection algorithm would produce over that age range, which is visualized as the triangular gap.

Here are 3 potential causes for this triangular gap:

1. A popular wallet is not implementing the same decoy selection algorithm, and is suggesting too recent of outputs.

2. A nefarious actor with many outputs is modifying their client to select more recent outputs.

3. Recent real outputs unaccounted for by the decoy selection algorithm are appearing in that range.

After doing a fair amount of investigating, I lean strongly toward 3 for a host of reasons, and I can dive further into my reasoning for why. But for the sake of staying focused, will hold off unless asked. And will continue with the assumption that 3 seems most likely (that the decoy selection algorithm as is likely is under-selecting recent real outputs).

Continuing with that assumption, it would make sense to try and arrive at a solution that would bridge the gap between the current decoy selection algorithm and observed.

Additionally, it would be apparent that the gamma distribution (green line) is not perfectly applicable on its own as fitting the distribution to, and serving as "the source of truth". Meaning that if we were to try and match the decoy selection algorithm identically to the gamma (which is what the wallet did before block density was factored in), then the algorithm would likely perform even worse by under-selecting even fewer outputs in the range in the chart (since green is a fair amount below orange, which is already below blue).


##### Comparing the different windows

The window of 20 seems to perform well between ages 11 and 30 (by "well", I mean it bridges the gap from current to observed on-chain data, which is defined as "well" because of the assumption above that the gap from current to observed is caused by the algorithm missing real outputs). However, around age 30, it shifts below the current decoy selection algorithm, which is also below observed on-chain. Thus, it starts to marginally under-select outputs, and is thus performing marginally worse than the current. This, to me, felt like potential for harm (it is why I initially chose 50 over 40, because you can see that 40 starts to move *slightly* below orange line later on as well).

The potential for harm does, however, seem very small. Here are the Kolmogorov-Smirnov statistics for each distribution when comparing to observed on-chain data (recall, smaller means the distance to observed is lower, and therefore it performs "better" based on the above assumption that getting closer to observed is the desirable outcome):


```
Distribution type              |   K-S Stat
-------------------------------------------------
 Current decoy selection algo  |   0.02105
 Normal gamma                  |   0.03318
 Window 1                      |   0.01943
 Window 3                      |   0.01596
 Window 10                     |   0.00892
 Window 20                     |   0.00891
 Window 40                     |   0.00895
 Window 50                     |   0.00897
 Window 75                     |   0.01289
 Window 100                    |   0.01539
```

Given the above K-S stats, it seems you can't really go wrong with a window between 10-50. Thus, I figure visual analysis combined with the K-S stat seems a prudent route to arrive at a window of 50.


### v12 (1978433 - 2210000)

different_windows_v12

```

Distribution type              |   K-S Stat
-------------------------------------------------
 Current decoy selection algo  |   0.02528
 Normal gamma                  |   0.04111
 Window 1                      |   0.01744
 Window 3                      |   0.01300
 Window 10                     |   0.01150
 Window 20                     |   0.01155
 Window 40                     |   0.01153
 Window 50                     |   0.01285
 Window 75                     |   0.01714
 Window 100                    |   0.01981

```

Seems to exhibit very similar properties to v14, and the same analysis above applies.


### v11 (1788720 - 1978433)

different_windows_v11

```
Distribution type              |   K-S Stat
-------------------------------------------------
 Current decoy selection algo  |   0.04014
 Normal gamma                  |   0.01847
 Window 1                      |   0.04279
 Window 3                      |   0.04277
 Window 10                     |   0.04269
 Window 20                     |   0.04271
 Window 40                     |   0.04272
 Window 50                     |   0.04272
 Window 75                     |   0.04287
 Window 100                    |   0.04273
```


In this interval, it appears clear that the gamma does apply best. There isn't really much to go off of here in way of deciding between the different windows. I believe this interval looks like this because the code to factor in block density was not released until [July 17, 2019](https://github.com/monero-project/monero/releases/tag/v0.14.1.0), or circa block 1881000. Further, this code wasn't released as part of a hard fork. So I don't think it makes much sense to use it as part of analysis to decide how to modify the current approach.



### average_output_time shift from 2 to 1 (2383730 - 2413735)

different_windows_average_output_time_shift

```
Distribution type              |   K-S Stat
-------------------------------------------------
 Current decoy selection algo  |   0.03079
 Normal gamma                  |   0.04920
 Window 1                      |   0.02568
 Window 3                      |   0.02553
 Window 10                     |   0.02557
 Window 20                     |   0.02563
 Window 40                     |   0.02548
 Window 50                     |   0.02557
 Window 75                     |   0.02550
 Window 100                    |   0.02545
```

As highlighted in #7798, the decoy selection algorithm is currently missing recent outputs at a step in the selection calculation by a factor of nearly 2x because of an integer truncation issue. Since block 2383730, `average_output_time` is being truncated down to 1, when it should really be around 1.9. This epoch demonstrates how the current decoy selection algorithm is now under-selecting recent outputs by a wider margin as a result (the integer truncation issue has the largest factor impact in this epoch).

Not much extra to add to this analysis in way of choosing between the windows, however.



## Conclusion

I still believe a window of 50 offers the least-potential-for-harm approach, as it won't marginally start to under-select outputs between ages 30 - 100 (like a smaller window appears to do), and it will also select a decent-sized share of younger outputs, as appears to be desired.


## My data + code to plot diagrams and calculate K-S statistics

### Output age data for each epoch

Unzip the following csv's:

[v14 (2210720 - 2413735).zip](https://github.com/monero-project/monero/files/6980710/v14.2210720.-.2413735.zip)
[v12 (1978433 - 2210000).zip](https://github.com/monero-project/monero/files/6980712/v12.1978433.-.2210000.zip)
[v11 (1788720 - 1978433).zip](https://github.com/monero-project/monero/files/6980713/v11.1788720.-.1978433.zip)
[average_output_time shift from 2 to 1 (2383730 - 2413735).zip](https://github.com/monero-project/monero/files/6980715/average_output_time.shift.from.2.to.1.2383730.-.2413735.zip)

My code to produce the csv's is a bit messy, but happy to clean it up and share if desired. I heavily modified `blockchain_usage.cpp` to write the files (the code doesn't really resemble `blockchain_usage.cpp` at all).


### Python to plot diagrams

```
import matplotlib.pyplot as plt
import csv
import matplotlib.ticker as mticker

plots = [
    'v14 (2210720 - 2413735)',
    'v12 (1978433 - 2210000)', 
    'v11 (1788720 - 1978433)', 
    'average_output_time shift from 2 to 1 (2383730 - 2413735)'
]

for plot in plots:
    headers = []
    values = []

    with open(plot + '.csv', 'r') as csvfile:

        rows = csv.reader(csvfile, delimiter = ',')
    
        idx = 0
        for row in rows:
            if idx == 0:
                for h in row:
                    headers.append(h)

                i = 0
                while (i < len(row)):
                    values.append([int(0)])
                    i += 1

            elif idx > 0:
                j = 0
                for v in row:
                    values[j].append(int(v))
                    j += 1

            idx += 1
  
    ax = plt.axes()

    idx = 1
    while (idx < len(headers)):
        blocks = values[0]

        line_style = '-'
        if idx > 3:
            line_style = '--'

        ax.plot(blocks, values[idx], line_style, label=headers[idx])

        idx += 1

    plt.legend()
    plt.xscale('log')

    ax.xaxis.set_major_formatter(mticker.StrMethodFormatter('{x:.0f}'))
    ax.xaxis.set_minor_formatter(mticker.NullFormatter())

    ax.yaxis.set_major_formatter(mticker.StrMethodFormatter('{x:.0f}'))
    ax.yaxis.set_minor_formatter(mticker.NullFormatter())

    plt.xlim(left=10)

    plt.title('Output age [' + plot + ']')

    plt.ylabel('Frequency')
    plt.xlabel('Output Age (blocks)')
    plt.show()
```


### Python to calculate K-S statistics

```
import csv

plots = [
    # 'output_age_data',
    'v14 (2210720 - 2413735)',
    'v12 (1978433 - 2210000)', 
    'v11 (1788720 - 1978433)', 
    'average_output_time shift from 2 to 1 (2383730 - 2413735)'
]

for plot in plots:
    headers = []
    cumulative_sums = []

    with open(plot + '.csv', 'r') as csvfile:

        rows = csv.reader(csvfile, delimiter = ',')
    
        # get cumulative number of outputs at each block, for each distribution
        idx = 0
        for row in rows:
            if idx == 0:
                for h in row:
                    headers.append(h)

                i = 1 # skip "Blocks" column, counting outputs in other columns
                while (i < len(row)):
                    cumulative_sums.append([int(0)])
                    i += 1
            else:
                i = 1 # skip "Blocks" column
                while (i < len(row)):
                    prior_cumulative_sum_index = len(cumulative_sums[i - 1]) - 1
                    prior_cumulative_sum = cumulative_sums[i - 1][prior_cumulative_sum_index]

                    current_cumulative_sum = prior_cumulative_sum + int(row[i])

                    cumulative_sums[i - 1].append(current_cumulative_sum)
                    i += 1

            idx += 1

    # get cdf's for each distribution
    cdfs = []

    i = 0
    for cumulative_sum_distr in cumulative_sums:
        cdfs.append([])

        for cumulative_sum in cumulative_sum_distr:
            num_outputs_observed = cumulative_sum_distr[len(cumulative_sum_distr) - 1]
            cdfs[i].append(cumulative_sum / num_outputs_observed)

        i += 1

    # iterate over F(x)'s and get abs(F(observed) - F(x))
    set_of_diffs = []
    cdf_observed = cdfs[0]

    i = 0
    for cdf in cdfs:
        if i > 0:
            set_of_diffs.append([])

            j = 0
            for cumulative_percentage in cdf:
                set_of_diffs[i - 1].append(abs(cdf_observed[j] - cumulative_percentage))
                j += 1

        i += 1

    # get the max of each to get K-S statistic
    i = 2
    print("BLOCK RANGE: ", plot, "\n")
    print("%-30s |   %-6s" % ("Distribution type", "K-S Stat"))
    print("-------------------------------------------------")
    for diff_set in set_of_diffs:
        ks_stat = max(diff_set)
        print("%-30s |   %1.5f" % (headers[i], ks_stat))
        i += 1

    print("\n*****************\n")
```

**Comment by luigi1111** [comment]
I agree with the analysis on the surface, and the conclusion that something in the 10-50 range is the "least wrong" of the current set. I have no particular objection to 50 as the choice. In any case, will give this some additional time for further percolation and input, if any.

**Comment by Rucknium** [comment]
From a statistical perspective, I support the latest version. What is accomplished here is "thickening" the probability density function of the selection algorithm in the section closest to zero. This more closely mimics the observed distribution of mixins + real spends. However, in the near future it is crucial that we consider moving away from the current selection algorithm that is based on Moser et al. 2018. I have some ideas about how to accomplish this.

File: src/wallet/wallet2.cpp [hunk 1]
```diff
@@ -1028,6 +1031,26 @@ uint64_t gamma_picker::pick()
...
   double x = gamma(engine);
   x = exp(x);
+
+  if (x > DEFAULT_UNLOCK_TIME) 
+  {
+    // The gamma recommended in Miller et al (https://arxiv.org/pdf/1704.04299/) starts from the chain tip.
+    // For example, if exp(x) = 5000, this suggests an output was spent 5000 seconds prior to the current height.
+    // Since we are subtracting the gamma value from num_rct_outputs prior to the unlock time, we need to pull 
+    // the gamma forward to the expected chain tip to apply the gamma distribution correctly.
+    // Source code to paper: https://github.com/maltemoeser/moneropaper
```
UkoeHB, line 1041:
```suggestion
    // We are trying to select an output from the chain that appeared 'x' seconds before the
    // current chain tip, based on the gamma recommended in Miller et al (https://arxiv.org/pdf/1704.04299/).
    // Our method is to get the average time delta between outputs in the recent past, estimate the number of
    // outputs 'n' that would have appeared between 'chain_tip - x' and 'chain_tip', then select the real output at
    // 'current_num_outputs - n'.
    // Source code to paper: https://github.com/maltemoeser/moneropaper
    //
    // However, due to the 'default spendable age' mechanic in Monero, 'current_num_outputs' only contains
    // currently *unlocked* outputs, which means the earliest output that can be selected is not at the chain tip!
    // Therefore, we must offset 'x' so it matches up with the timing of the outputs being considered. We do
    // this by saying if 'x == DEFAULT_UNLOCK_TIME' (the expected age of the first unlocked output at any time,
    // compared to the current chain tip), then select the first unlocked output.
```

UkoeHB (changes requested):

src/wallet/wallet2.cpp
j-berman: Much clearer, thank you :)

src/wallet/wallet2.cpp
```diff
@@ -1028,6 +1031,33 @@ uint64_t gamma_picker::pick()
...
+    // Our method is to get the average time delta between outputs in the recent past, estimate the number of
+    // outputs 'n' that would have appeared between 'chain_tip - x' and 'chain_tip', then select the real output at
+    // 'current_num_outputs - n'.
+    // Source code to paper: https://github.com/maltemoeser/moneropaper
+    //
+    // However, due to the 'default spendable age' mechanic in Monero, 'current_num_outputs' only contains
+    // currently *unlocked* outputs, which means the earliest output that can be selected is not at the chain tip!
+    // Therefore, we must offset 'x' so it matches up with the timing of the outputs being considered. We do
+    // this by saying if 'x == DEFAULT_UNLOCK_TIME' (the expected age of the first unlocked output at any time,
+    // compared to the current chain tip), then select the first unlocked output.
```
UkoeHB, line 1048:
```suggestion
    // We are trying to select an output from the chain that appeared 'x' seconds before the
    // current chain tip, where 'x' is selected from the gamma distribution recommended in Miller et al.
    // (https://arxiv.org/pdf/1704.04299/).
    // Our method is to get the average time delta between outputs in the recent past, estimate the number of
    // outputs 'n' that would have appeared between 'chain_tip - x' and 'chain_tip', select the real output at
    // 'current_num_outputs - n', then randomly select an output from the block where that output appears.
    // Source code to paper: https://github.com/maltemoeser/moneropaper
    //
    // Due to the 'default spendable age' mechanic in Monero, 'current_num_outputs' only contains
    // currently *unlocked* outputs, which means the earliest output that can be selected is not at the chain tip!
    // Therefore, we must offset 'x' so it matches up with the timing of the outputs being considered. We do
    // this by saying if 'x` equals the expected age of the first unlocked output (compared to the current
    // chain tip - i.e. DEFAULT_UNLOCK_TIME), then select the first unlocked output.
```

New day, fresh eyes. Sorry!

UkoeHB (changes requested):

UkoeHB (approved):
Thank you!

selsta (approved):

PR #7825, opened 2021-08-01
release-v0.17 backport

selsta:


mj-xmr (approved):
I know almost all of the changes and they look OK here as well.

The one thing that I can't say much about are the changes in the `wallet2.cpp`

mj-xmr (approved):
Compared wallet2 changes against Master and it looks OK.

PR #7827, opened 2021-08-03
Ci make single threaded after failure

mj-xmr:
If the parallel version of `make` fails, running `make` in single threaded mode with the same parameters as before. This brings no side effects if the parallel version succeeds, but if it fails, then it makes it much more clearer what the error was. This is because for the single threaded version of make the first error that occurred, will be visible at the very bottom of the log, instead of being cluttered with potentially many unrelated warnings below the actual error.


selsta:
Closing as there was no support for merging this.

PR #7828, opened 2021-08-03
EasyLogging++: new anti-UB test and propagating exception at misconfiguration

mj-xmr:
There was a branch of `EasyLogging++` code, that lead to UB during incomplete configuration (see the added UTest). Now the branch will throw an exception, expecting to be handled at a higher level.


external/easylogging++/easylogging++.h
```diff
@@ -2022,10 +2022,11 @@ class TypedConfigurations : public base::threading::ThreadSafe {
     if (it == confMap->end()) {
       try {
         return confMap->at(Level::Global);
-      } catch (...) {
+      } catch (std::exception & ex){
```
vtnerd, line 2025: Revert `catch (...)` and do a `throw;` instead, which rethrows the current exception.

mj-xmr: Done.

selsta (approved):

PR #7833, opened 2021-08-05
misc_utils: add factorial and n_choose_k functions

UkoeHB:
Adds an `n_choose_k<T>()` function for a future PR (multisig key generation). Also adds `factorial<T>()` to support it.


UkoeHB:
Does the CI system run the tests in `contrib/epee/tests/src/tests.cpp`?

contrib/epee/include/misc_language.h
```diff
@@ -128,6 +128,59 @@ namespace misc_utils
 
   }
 
+  template<typename T>
```
vtnerd, line 131: `epee` is typically networking related code and stuff to support the networking code. Based on your description, it will not be used by networking code, so move to `src/` somewhere.

contrib/epee/include/misc_language.h
```diff
@@ -128,6 +128,59 @@ namespace misc_utils
...
+  T factorial(T n, T s = 0)
```
vtnerd, line 132: Why make this templated? The difficult part is that since its not used anywhere (right now), I have no idea whether the template is useful or not. It _seems like_ only one type will eventually be necessary.

Each type ends up being its own assembly, so its best to converge where possible.

vtnerd (commented):
The `boost::math` library has both of these functions already, and the only difference is the `s` parameter for `factorial`. However, you get the same thing through subtraction before the function call. I'm not sure its worth re-writing these given our current reliance on boost (its unlikely to be removed in the near future).

UkoeHB:
Boost's `binomial_coefficient` [does not support integral types](https://github.com/boostorg/math/blob/b2538faaf9802af8e856c603b9001e33db826676/include/boost/math/special_functions/binomial.hpp#L24).

> However, you get the same thing through subtraction before the function call.

Not true: `n! / s! != (n - s)!`. My goal here is to support integral types in `n_choose_k` while minimizing integer overflow. However, the current approach isn't perfect so I am thinking about a better algorithm (factorize numerator and denominator into lists, then search the lists to remove common factors).

contrib/epee/include/misc_language.h
UkoeHB: I am ok with moving it, but feel uncertain about `is typically networking related code and stuff to support the networking code`.

I chose this file because it already has `median` and `get_mid`.
- `median` is only used in `src` stuff (and tests) as far as I can tell.
- `get_mid` is used in `median` and `rolling_median`
- `rolling_median`, while defined in `epee`, is not related to networking.

contrib/epee/include/misc_language.h
UkoeHB: It feels unpleasant to tailor an otherwise generic math function to one use-case.

vtnerd:
Or we could avoid re-inventing the wheel and use the functions from `boost::math`.

vtnerd:
Sorry, fired that off too soon. That is unfortunate, I guess I get to actually review the code then.

contrib/epee/include/misc_language.h
vtnerd: Why is it unpleasant? We probably only need to use unsigned integers of a certain width anyway.

contrib/epee/include/misc_language.h
vtnerd: Possibly, if they aren't used in epee elsewhere.

contrib/epee/include/misc_language.h
```diff
@@ -128,6 +128,59 @@ namespace misc_utils
...
+  template<typename T>
+  T n_choose_k(T n, T k)
+  {
+    static_assert(std::is_integral<T>::value, "Integral type required.");
+
+    // n choose k = n! / (k! * (n - k)!)
+    if (n < 0 || k < 0)
+      return 0;
+    else if (n < k)
+      return 1;
```
vtnerd, line 163: Why `return 1;` here? This is basically another error case as choose k elements from n only makes sense if this statement is `true`, but why differentiate this instead of using `0` like all the other error cases?

contrib/epee/include/misc_language.h
```diff
@@ -128,6 +128,59 @@ namespace misc_utils
 
   }
 
+  template<typename T>
+  T factorial(T n, T s = 0)
+  {
+    static_assert(std::is_integral<T>::value, "Integral type required.");
+
+    if (s > n || n < 0 || s < 0)
```
vtnerd, line 136: `static_assert(!std::is_signed<T>::value, ...);` ?

contrib/epee/include/misc_language.h
```diff
@@ -128,6 +128,59 @@ namespace misc_utils
...
+      n--;
+    }
+
+    return r;
+  }
+
+  template<typename T>
+  T n_choose_k(T n, T k)
+  {
+    static_assert(std::is_integral<T>::value, "Integral type required.");
```
vtnerd, line 157: Same thing here with an unsigned static assert.

contrib/epee/include/misc_language.h
```diff
@@ -128,6 +128,59 @@ namespace misc_utils
...
+    else if (n < k)
+      return 1;
+    else if (k >= n - k)
+    {
+      // (n! / k!) / (n - k)!
+      T numerator{factorial<T>(n - k)};
+      if (numerator == 0)
+        return 0;
+      else
+        return factorial<T>(n, k) / numerator;
```
vtnerd, line 171: So if this overflow its `return 0;` ... hmm. I suppose that makes sense as an error case value, a comment br both functions may help with that (check for `0` as its the error case).

vtnerd:
> My goal here is to support integral types in n_choose_k while minimizing integer overflow. However, the current approach isn't perfect so I am thinking about a better algorithm (factorize numerator and denominator into lists, then search the lists to remove common factors).

Why? Is this going to be on the hotpath? These functions are unlikely ever to hit in the thousands of cycles (even considering stalls), so why bother with the complexity and less readability?

UkoeHB:
> why bother with the complexity and less readability?

The complexity is to mitigate integer overflow. The function should **only** fail if the return value exceeds numerical limits. The current version does not achieve that, so I am redoing it with a new algorithm.

vtnerd:
> The complexity is to mitigate integer overflow. The function should only fail if the return value exceeds numerical limits. The current version does not achieve that, so I am redoing it with a new algorithm.

I made a stealth edit - could you provide a resource on how this is supposed to be used? I assumed the inputs to these functions would be small integers, and that isn't the case?

UkoeHB:
> could you point to a resource on how this is being used by multisig? I'm not aware of the changes necessary or what this is even for .

I am using this to detect how many shared keys there should be in each round of multisig key exchange, in order to properly validate inputs.

> I assumed the inputs to these functions would be small integers, and that isn't the case?

It probably will be in the typical case, but I would like the function to have a robust interface.

vtnerd:
> It probably will be in the typical case, but I would like the function to have a robust interface.

With `uint64_t` the interface starts to possibly overflow around `n == 21`, and with `boost::multiprecision::uint256_t` it begins to possibly overflow around `n == 57`.

The multiprecision code will definitely be a lot slower, but for the use case it seems doesn't seem like it will matter. And `n==57` is fairly big, assuming N is participants (wasn't quite clear on that with your edit).

UkoeHB:
Hmm in practice multisig probably can't exceed 20 participants. The problem is you need a LOT of shared keys if there are a lot of participants. At N = 20, each person will have up to 184k shared secrets if it is a 10-of-20 group.

I am not sure the best approach (that is personally satisfying).
A) multisig can't realistically exceed N = 20
B) a robust function interface requires complexity and inefficiency

UkoeHB:
I threw together a C++ project to study different approaches: https://github.com/UkoeHB/bin-coeff-integral

Numerical limit results:
```
binomial_coefficient_integral_fast<std::uint16_t>(n,k) -> max n = 11
binomial_coefficient_integral_fast<std::uint32_t>(n,k) -> max n = 17
binomial_coefficient_integral_fast<std::uint64_t>(n,k) -> max n = 29
binomial_coefficient_integral_slow -> max n = 34
```

- `binomial_coefficient_integral_fast<T>()` is the function currently in this PR.
- `binomial_coefficient_integral_slow()` is a slower version with ideal interface (only fails on numerical limits if `ret_val > T::max()`)

contrib/epee/include/misc_language.h
UkoeHB: I changed my mind, `n < k` is also an error case (will update later).

contrib/epee/include/misc_language.h
UkoeHB: My idea was to support all integral types, and just error out if negative numbers are passed in. I suppose in professional grade code you expect very strict use of types, so maybe a stricter interface is appropriate here.

mj-xmr:
> Does the CI system run the tests in `contrib/epee/tests/src/tests.cpp`?

Unfortunately no, and it's a flaw. Please use the attached patch (`git apply epee_tests.patch.txt`) to learn what the CMakeFiles.txt changes have to be, in order to get the epee tests under the automated CI.
I had to comment out almost all of the old `tests.cpp` code, because if failed to compile. 
[epee_tests.patch.txt](https://github.com/monero-project/monero/files/6968460/epee_tests.patch.txt)

mj-xmr:
> The complexity is to mitigate integer overflow. The function should **only** fail if the return value exceeds numerical limits. The current version does not achieve that, so I am redoing it with a new algorithm.

Exactly for this reason, I'd love to see this under CI (see patch above), as well have those corner cases of the overflowing library exposed in the unit tests (with an failure expectation), while at the same time, I'd love to see the new implementation being called with exactly the same parameters and have the expectation set to true. Pseudocode:

```
static fun_cornercases(int x, int y, int expected)
{
 EXPECT_FALSE(OldLib::Calc(x, y)) // Or EXPECT_ANY_THROW or whatever makes sense
 EXPECT_EQ(NewLib::Calc(x, y), expected);
}

TEST_CORNERCASE_1()
{
   fun_cornercases(-100, 200, 10);
}
```

UkoeHB:
I believe this PR would end up over 600 lines with my updated algorithm (plus tests, etc.), just to get `n_choose_k` for one use-case.

How about just wrapping boost's `binomial_coefficient` with a conversion from floating point to uint?

```
#include <boost/math/special_functions/binomial.hpp>
#include <cmath>

auto n_choose_k_lam{
        [](const std::uint32_t n, const std::uint32_t k) -> std::int32_t
        {
            double fp_result = boost::math::binomial_coefficient<double>(n, k);

            if (fp_result < 0)
                return 0;

            if (fp_result > std::numeric_limits<std::int32_t>::max())
                return 0;

            return std::round(fp_result);
        }
    };

```

mj-xmr:
As you wish.
Does your wrapper have to be a lambda?

UkoeHB:
No, I will write a proper function with some tests so this PR can make progress.

mj-xmr:
I can do the CI part in a separate PR. Since the patch is already written, I can drive it.

It would be enough if you just put the tests into one of the files of the `unit_tests` target.

UkoeHB:
I am going to open a new PR, since this one has changed scope and intent too much.

vtnerd:
> Hmm in practice multisig probably can't exceed 20 participants. The problem is you need a LOT of shared keys if there are a lot of participants. At N = 20, each person will have up to 184k shared secrets if it is a 10-of-20 group.

This is why your initial simple approach was likely sufficient. The likelihood of another `n_choose_k` call in the code is unlikely, and the code can be updated at that stage (this isn't a deep architectural decision impacting the code).

If you wanted the boost version, I'm not sure if it loses precision whilst calculating, or ? As in, the result fits within 2^53 bits, but the intermediate value didn't and lost some precision. Unfortunately, I don't have any experience with this library.

UkoeHB:
> This is why your initial simple approach was likely sufficient. The likelihood of another `n_choose_k` call in the code is unlikely, and the code can be updated at that stage (this isn't a deep architectural decision impacting the code).

I am nervous about committing anything less than ideal (or at least 'robust') to the code base. In that path lies accumulating technical debt - death by a thousand suboptimalities.

> If you wanted the boost version, I'm not sure if it loses precision whilst calculating, or ? As in, the result fits within 2^53 bits, but the intermediate value didn't and lost some precision. Unfortunately, I don't have any experience with this library.

I'm not sure about this... The docs say:

> **[Binomial coefficient](https://www.boost.org/doc/libs/1_56_0/libs/math/doc/html/math_toolkit/factorials/sf_binomial.html)**: The accuracy will be the same as for the factorials for small arguments (i.e. no more than one or two epsilon), and the beta function for larger arguments.
>  **[Beta function](https://www.boost.org/doc/libs/1_56_0/libs/math/doc/html/math_toolkit/sf_beta/beta_function.html)**: The following table shows peak errors for various domains of input arguments... Note that only results for the widest floating point type on the system are given as narrower types have effectively zero error...
> Note that the worst errors occur when a or b are large, and that when this is the case the result is very close to zero, so absolute errors will be very small.

vtnerd:
> I am nervous about committing anything less than ideal (or at least 'robust') to the code base. In that path lies accumulating technical debt - death by a thousand suboptimalities.

There is zero technical debt in something so small. Its used in one spot. The function interface wouldn't need to change if a more robust implementation was needed. Technical debt is something like `wallet2`.

> I'm not sure about this... The docs say:

Which is possibly worse than the limited overflow implementation. Although, I wouldn't anticipate issues for the single use case.

PR #7845, opened 2021-08-11
wallet2: protect client from divide by 0 caused by integer truncation

j-berman:
This PR is meant to replace #7798  for next release -- I recommend keeping 7798 on hold for now.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1019,7 +1019,13 @@ gamma_picker::gamma_picker(const std::vector<uint64_t> &rct_offsets, double shap
   end = rct_offsets.data() + rct_offsets.size() - CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE;
   num_rct_outputs = *(end - 1);
   THROW_WALLET_EXCEPTION_IF(num_rct_outputs == 0, error::wallet_internal_error, "No rct outputs");
+  THROW_WALLET_EXCEPTION_IF(outputs_to_consider == 0, error::wallet_internal_error, "No rct outputs to consider");
   average_output_time = DIFFICULTY_TARGET_V2 * blocks_to_consider / outputs_to_consider; // this assumes constant target over the whole rct range
+  if (average_output_time == 0) {
+     // TODO: apply this to all cases; do so alongside a hard fork, where all clients will update at the same time, preventing anonymity puddle formation
+    average_output_time = DIFFICULTY_TARGET_V2 * blocks_to_consider / static_cast<double>(outputs_to_consider);
+  }
+  THROW_WALLET_EXCEPTION_IF(average_output_time == 0, error::wallet_internal_error, "Average seconds per output cannot be 0.");
 };
 
 gamma_picker::gamma_picker(const std::vector<uint64_t> &rct_offsets): gamma_picker(rct_offsets, GAMMA_SHAPE, GAMMA_SCALE) {}


mj-xmr:
It looks OK from technical point if view (UB / div by 0).

No comment on the functionality side from me, unfortunately.

mj-xmr:
Not directly related, but there will be some truncation done [in line 1035](https://github.com/monero-project/monero/pull/7845/files#diff-04cf14f64d2023c7f9cd7bd8e51dcb32ed400443c6a67535cb0105cfa2b62c3cR1035):

`uint64_t output_index = x / average_output_time;`

In this case the output variable should be rounded first, then cast:

`uint64_t output_index = static_cast<uint64_t>(round(x / average_output_time));`

A candidate for a future PR / issue?

src/wallet/wallet2.cpp
```diff
@@ -1018,8 +1018,12 @@ gamma_picker::gamma_picker(const std::vector<uint64_t> &rct_offsets, double shap
   begin = rct_offsets.data();
   end = rct_offsets.data() + rct_offsets.size() - CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE;
   num_rct_outputs = *(end - 1);
-  THROW_WALLET_EXCEPTION_IF(num_rct_outputs == 0, error::wallet_internal_error, "No rct outputs");
+  THROW_WALLET_EXCEPTION_IF(num_rct_outputs == 0 || outputs_to_consider == 0, error::wallet_internal_error, "No rct outputs");
```
mj-xmr, line 1021: One nitpick here would be that my prefference is:

```
THROW_WALLET_EXCEPTION_IF(num_rct_outputs == 0,      error::wallet_internal_error, "No rct outputs");
THROW_WALLET_EXCEPTION_IF(outputs_to_consider == 0, error::wallet_internal_error, "No rct outputs to consider");
```

mj-xmr (changes requested):

j-berman:
> Not directly related, but there will be some truncation done [in line 1035](https://github.com/monero-project/monero/pull/7845/files#diff-04cf14f64d2023c7f9cd7bd8e51dcb32ed400443c6a67535cb0105cfa2b62c3cR1035):
> 
> `uint64_t output_index = x / average_output_time;`
> 
> In this case the output variable should be rounded first, and next cast (optionally, just to avoid warnings):
> 
> `uint64_t output_index = static_cast<uint64_t>(round(x / average_output_time));`
> 
> If rounding sounds like a bad idea, because it adds +1 to the index when the underlying value is >= 0.5, then `floor` should be used instead of `round` to make it more explicit what is expected from this line.
> 
> A candidate for a future PR / issue?

Agree with this! I'll submit a separate PR that uses `floor` for that exact circumstance :)

mj-xmr (approved):

src/wallet/wallet2.cpp
```diff
@@ -1019,7 +1019,12 @@ gamma_picker::gamma_picker(const std::vector<uint64_t> &rct_offsets, double shap
   end = rct_offsets.data() + rct_offsets.size() - CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE;
   num_rct_outputs = *(end - 1);
   THROW_WALLET_EXCEPTION_IF(num_rct_outputs == 0, error::wallet_internal_error, "No rct outputs");
+  THROW_WALLET_EXCEPTION_IF(outputs_to_consider == 0, error::wallet_internal_error, "No rct outputs to consider");
   average_output_time = DIFFICULTY_TARGET_V2 * blocks_to_consider / outputs_to_consider; // this assumes constant target over the whole rct range
+  if (average_output_time == 0) {
+    average_output_time = DIFFICULTY_TARGET_V2 * blocks_to_consider / static_cast<double>(outputs_to_consider);
```
UkoeHB, line 1025:
```suggestion
     // TODO: apply this to all cases; do so alongside a hard fork, where all clients will update at the same time, preventing anonymity puddle formation
    average_output_time = DIFFICULTY_TARGET_V2 * blocks_to_consider / static_cast<double>(outputs_to_consider);
```

UkoeHB (changes requested):

UkoeHB (approved):

mj-xmr:
Once the code appears to be in a final stage, please squash to just 1 commit and rebase.

selsta:
There doesn't appear the need to rebase, squashing and force pushing should be fine.

selsta (approved):

mj-xmr (approved):

PR #7851, opened 2021-08-11
Simplewallet: aborting get_mnemonic_language() in case a negative index is used

mj-xmr:
Inside `get_mnemonic_language()` of `Simplewallet`:
In case the error case is indeed reached, the appropriate error logs are stored, but the function carries on with accessing a vector, using a negative index. Using the `vector`'s `at()` method protects against this, by aborting the branch and throwing an exception. Leaving this as it was causes UB.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -4895,7 +4895,7 @@ std::string simple_wallet::get_mnemonic_language()
       fail_msg_writer() << tr("invalid language choice entered. Please try again.\n");
     }
   }
-  return language_list_self[language_number];
+  return language_list_self.at(language_number);
 }
 //----------------------------------------------------------------------------------------------------
 boost::optional<tools::password_container> simple_wallet::get_and_verify_password() const


selsta (commented):
Can you explain how this would work?

As far as I can see, if `language_number` is negative it will stay inside the while loop, so it isn't possible for `language_list_self[language_number];` to be called unless it is zero or positive.

mj-xmr:
> Can you explain how this would work?
> 
> As far as I can see, if `language_number` is negative it will stay inside the while loop, so it isn't possible for `language_list_self[language_number];` to be called unless it is zero or positive.

I think you're right. The line [4887](https://github.com/monero-project/monero/pull/7851/files#diff-25b9648b3da507060201d4ce30539c281676672c1af9a8e4d0e16ec0585a987bR4887) handles negative as well as overflowing indexes correctly in the end.

```C++
if (!((language_number >= 0) && (static_cast<unsigned int>(language_number) < language_list.size())))
{
   language_number = -1;
   fail_msg_writer() << tr("invalid language choice entered. Please try again.\n");
}
```

Should I abandon this or do you think that some extra exception safety won't hurt? (in case somebody screws the logic inside the `while()` loop under his feet?

PR #7852, opened 2021-08-11
[seraphis] wallet2: replace ambiguous 'multisig()' function with 'multisig_status' struct

UkoeHB:
## Summary

This is a PR in my 'upstreaming seraphis_lib project'. Refactoring with this PR will make it easier to transition to seraphis in the future, since the seraphis multisig wallet will use `multisig::multisig_account_status`.

- Replaces `wallet2::multisig()` with `wallet2::get_multisig_status()`, which returns a `multisig::multisig_account_status` struct containing info about the wallet's transformation into a 'multisig wallet'.

## Explanation

- The function `wallet2::multisig()` is ambiguous (returns `true` if the wallet `is_multisig`), and badly defined (uses return by pointer reference to set variables related to the wallet's multisig status).
- It is also awkward to extend, requiring an updated function signature. I want to add `kex_rounds_remaining` to the status, which is much cleaner to do with an added member in the return struct `multisig_status`.

This PR is intended to improve code quality. Note that `wallet2_api` uses a similar pattern already (a `MultisigState` struct).

## Future PRs
- Add `kex_rounds_remaining` to status, to show how many more rounds of communication are required before a multisig address is complete.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -786,7 +786,6 @@ bool simple_wallet::print_seed(bool encrypted)
 {
   bool success =  false;
   epee::wipeable_string seed;
-  bool ready, multisig;
 
   if (m_wallet->key_on_device())
   {
@@ -799,10 +798,10 @@ bool simple_wallet::print_seed(bool encrypted)
     return true;
   }
 
-  multisig = m_wallet->multisig(&ready);
-  if (multisig)
+  const multisig::multisig_account_status ms_status{m_wallet->get_multisig_status()};
+  if (ms_status.multisig_is_active)
   {
-    if (!ready)
+    if (!ms_status.is_ready)
     {
       fail_msg_writer() << tr("wallet is multisig but not yet finalized");
       return true;
@@ -811,7 +810,7 @@ bool simple_wallet::print_seed(bool encrypted)
 
   SCOPED_WALLET_UNLOCK();
 
-  if (!multisig && !m_wallet->is_deterministic())
+  if (!ms_status.multisig_is_active && !m_wallet->is_deterministic())
   {
     fail_msg_writer() << tr("wallet is non-deterministic and has no seed");
     return true;
@@ -826,7 +825,7 @@ bool simple_wallet::print_seed(bool encrypted)
     seed_pass = pwd_container->password();
   }
 
-  if (multisig)
+  if (ms_status.multisig_is_active)
     success = m_wallet->get_multisig_seed(seed, seed_pass);
   else if (m_wallet->is_deterministic())
     success = m_wallet->get_seed(seed, seed_pass);
@@ -865,7 +864,7 @@ bool simple_wallet::seed_set_language(const std::vector<std::string> &args/* = s
     fail_msg_writer() << tr("command not supported by HW wallet");
     return true;
   }
-  if (m_wallet->multisig())
+  if (m_wallet->get_multisig_status().multisig_is_active)
   {
     fail_msg_writer() << tr("wallet is multisig and has no seed");
     return true;
@@ -1011,7 +1010,7 @@ bool simple_wallet::prepare_multisig_main(const std::vector<std::string> &args, 
     fail_msg_writer() << tr("command not supported by HW wallet");
     return false;
   }
-  if (m_wallet->multisig())
+  if (m_wallet->get_multisig_status().multisig_is_active)
   {
     fail_msg_writer() << tr("This wallet is already multisig");
     return false;
@@ -1058,7 +1057,7 @@ bool simple_wallet::make_multisig_main(const std::vector<std::string> &args, boo
     fail_msg_writer() << tr("command not supported by HW wallet");
     return false;
   }
-  if (m_wallet->multisig())
+  if (m_wallet->get_multisig_status().multisig_is_active)
   {
     fail_msg_writer() << tr("This wallet is already multisig");
     return false;
@@ -1103,9 +1102,7 @@ bool simple_wallet::make_multisig_main(const std::vector<std::string> &args, boo
     auto local_args = args;
     local_args.erase(local_args.begin());
     std::string multisig_extra_info = m_wallet->make_multisig(orig_pwd_container->password(), local_args, threshold);
-    bool ready;
-    m_wallet->multisig(&ready);
-    if (!ready)
+    if (!m_wallet->get_multisig_status().is_ready)
     {
       success_msg_writer() << tr("Another step is needed");
       success_msg_writer() << multisig_extra_info;
@@ -1123,13 +1120,13 @@ bool simple_wallet::make_multisig_main(const std::vector<std::string> &args, boo
     return false;
   }
 
-  uint32_t total;
-  if (!m_wallet->multisig(NULL, &threshold, &total))
+  const multisig::multisig_account_status ms_status{m_wallet->get_multisig_status()};
+  if (!ms_status.multisig_is_active)
   {
     fail_msg_writer() << tr("Error creating multisig: new wallet is not multisig");
     return false;
   }
-  success_msg_writer() << std::to_string(threshold) << "/" << total << tr(" multisig address: ")
+  success_msg_writer() << std::to_string(ms_status.threshold) << "/" << ms_status.total << tr(" multisig address: ")
       << m_wallet->get_account().get_public_address_str(m_wallet->nettype());
 
   return true;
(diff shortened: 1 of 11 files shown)


src/wallet/wallet_rpc_server.cpp
```diff
@@ -4151,23 +4158,23 @@ namespace tools
...
-    if (ready)
+    if (status.is_ready)
     {
       er.code = WALLET_RPC_ERROR_CODE_ALREADY_MULTISIG;
       er.message = "This wallet is multisig, and already finalized";
       return false;
     }
 
-    if (req.multisig_info.size() < 1 || req.multisig_info.size() > total)
+    if (req.multisig_info.size() + 1 < status.total)
```
UkoeHB, line 4177: Bug fix. There should be at least one message per other participant here.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -4027,22 +4034,22 @@ namespace tools
...
-    if (!ready)
+    if (!status.is_ready)
     {
       er.code = WALLET_RPC_ERROR_CODE_NOT_MULTISIG;
       er.message = "This wallet is multisig, but not yet finalized";
       return false;
     }
 
-    if (req.info.size() < threshold - 1)
+    if (req.info.size() + 1 < status.threshold)
```
UkoeHB, line 4052: Bug fix, `threshold` is an unsigned int so arbitrary subtractions risk integer overflow.

src/wallet/wallet2.cpp
```diff
@@ -5433,20 +5433,33 @@ bool wallet2::verify_extra_multisig_info(const std::string &data, std::unordered
...
+  {
+    ret.is_multisig = false;
+    ret.threshold = 0;
+    ret.total = 0;
+    ret.is_ready = false;
+    ret.rounds_remaining = 0;
+  }
+
+  return ret;
+}
```
UkoeHB, line 5461: The PR's core change.

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -2399,12 +2399,14 @@ namespace wallet_rpc
...
       uint32_t threshold;
       uint32_t total;
+      uint32_t rounds_remaining;
 
       BEGIN_KV_SERIALIZE_MAP()
         KV_SERIALIZE(multisig)
         KV_SERIALIZE(ready)
         KV_SERIALIZE(threshold)
         KV_SERIALIZE(total)
+        KV_SERIALIZE(rounds_remaining)
```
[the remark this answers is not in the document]
rbrunner7: Yes, I think info about how many key exchange rounds are remaining can certainly be helpful.

src/simplewallet/simplewallet.cpp
```diff
@@ -792,10 +791,10 @@ bool simple_wallet::print_seed(bool encrypted)
     return true;
   }
 
-  multisig = m_wallet->multisig(&ready);
-  if (multisig)
+  tools::wallet2::multisig_status status{m_wallet->get_multisig_status()};
```
vtnerd, line 794: Trivial, but this can be `const` (and several more times below).

src/wallet/api/wallet.cpp
```diff
@@ -1324,7 +1326,12 @@ void WalletImpl::setSubaddressLabel(uint32_t accountIndex, uint32_t addressIndex
 
 MultisigState WalletImpl::multisig() const {
     MultisigState state;
-    state.isMultisig = m_wallet->multisig(&state.isReady, &state.threshold, &state.total);
+    tools::wallet2::multisig_status status{m_wallet->get_multisig_status()};
```
vtnerd, line 1329: These structs are duplicates, and the naming schemes are different :/

Is there somewhere `monero::MultisigState` can be placed to so that `wallet2` can re-use it? The problem seems to be that `src/wallet` and `src/wallet/api` aren't supposed to include header files from each other _in the header_ (only in the source of `src/wallet/api`).

vtnerd (commented):
I agree that if written again it should be done in this style, but at times style only changes have been rejected as being spurious churn. FWIW, I went through all of the changes and everything looked fine, but some other reviewers may want to peek at this.

src/wallet/api/wallet.cpp
UkoeHB: I am pretty confused about the purpose/use of `src/wallet/api`.

The correct architecture should be like this:
- multisig should be in `namespace cryptonote { namespace multisig { } }`
- wallets should own a `cryptonote::multisig::multisig_account` class object that handles all multisig account operations
- `cryptonote::multisig::multisig_account` should have a member function `get_status()` that returns a `cryptonote::multisig::multisig_account_status` struct. Presumably, `WalletImpl` would also be able to use that struct.

Right now, `wallet2` haphazardly owns a lot of multisig account details. That's why I added `multisig_status` to `wallet2`.

I think I'll move `src/wallet/wallet2 multisig_account -> src/multisig/multisig multisig_account_status` so we are a little closer to the ideal architecture.

rbrunner7 (approved):
For once a PR that I can fully review and approve because it's simple enough for me to understand the code.

IMHO the new call returning a structure is cleaner and better than the replaced one, and we should merge this to include in the upcoming hardfork.

UkoeHB:
I added `kex_is_done` to the multisig status and updated the RPC and MMS to include it.

UkoeHB:
Update: I plan to reimagine this PR as a step toward seraphis migration, with the expectation that new seraphis multisig wallet code will use the status struct.

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -2402,12 +2402,14 @@ namespace wallet_rpc
...
       bool multisig;
+      bool kex_is_done;
       bool ready;
       uint32_t threshold;
       uint32_t total;
 
       BEGIN_KV_SERIALIZE_MAP()
         KV_SERIALIZE(multisig)
+        KV_SERIALIZE(kex_is_done)
         KV_SERIALIZE(ready)
```
jeffro256, line 2413: I would suggest this for best backwards compatibility 

```suggestion
        KV_SERIALIZE(ready)
        KV_SERIALIZE_OPT(kex_is_done, ready)
```

src/wallet/wallet2.cpp
```diff
@@ -5263,20 +5263,30 @@ std::string wallet2::get_multisig_first_kex_msg() const
...
+
+  if (m_multisig)
   {
-    *ready = !(get_account().get_keys().m_account_address.m_spend_public_key == rct::rct2pk(rct::identity())) &&
+    ret.multisig_is_active = true;
+    ret.threshold = m_multisig_threshold;
+    ret.total = m_multisig_signers.size();
+    ret.kex_is_done = !(get_account().get_keys().m_account_address.m_spend_public_key == rct::rct2pk(rct::identity())) &&
+      (m_multisig_rounds_passed >= multisig::multisig_kex_rounds_required(m_multisig_signers.size(), m_multisig_threshold));
+    ret.is_ready = ret.kex_is_done &&
```
jeffro256, line 5277: The value of `is_ready` is different from the old `ready` under this PR, no? There's an extra stipulation for it to be true: `m_multisig_rounds_passed >= multisig::multisig_kex_rounds_required(m_multisig_signers.size(), m_multisig_threshold)`. Could this cause downstream code to act differently?

jeffro256 (changes requested):
In general, I approve of this PR on the condition that the new `multisig_account_status` is bound to have fields added. I have a couple comments about backwards compatibility.

src/wallet/wallet_rpc_server_commands_defs.h
jeffro256: Not for this codebase, but it would be it would be a problem for downstream code if they used `wallet_rpc_server_commands_defs.h`.

jeffro256:
You also need to update `WALLET_RPC_VERSION_MINOR`

src/wallet/wallet2.cpp
jeffro256: Are these setup rounds after the key exchange rounds?

src/wallet/wallet_rpc_server_commands_defs.h
selsta: I think we usually don't care about downstream code that use `src/wallet/wallet_rpc_server_commands_defs.h`, at least I'm not aware of any such projects that make use of it.

jeffro256: Okay fine then I'll drop it

PR #7853, opened 2021-08-11
Update README.md - added simple shields from shields.io to make README look more professional

limiteci:


BraINstinct0:
Caveat: should link to relevant URL when clicked.
use the badge itself as the 'text', and add the urls.

Plus I generally do love shields.io but relying on third-party service is sometimes avoided.

selsta:
I prefer no unnecessary embeds to third party services in the README.

PR #7854, opened 2021-08-12
CMake epee: add minimalistic test setup +CI

mj-xmr:
Integrating patch from https://github.com/monero-project/monero/pull/7833#issuecomment-896796218.

The discussion there was about how to test epee code, and whether it makes sense to test it before it is being distributed to the dependees. My opinion is, that whenever possible, a given package should be responsible to test its own basic functionality in isolation from other components as the first line of defense.

The second point in the mentioned PR was, that such tests, in case of epee as well as others, aren't part of the CI. This PR adds the boilerplate code, necessary to include the test in our CI. The PR also (for now) removes the epee's `tests.cpp` file from the CMakeLists.txt file, since it doesn't compile anymore. In the future, we can re-include the working tests from `tests.cpp` step by step. For now, I only extracted the math tests as a starting example.

### Result:
```
Running tests...
Test project /home/runner/work/monero/monero/build
      Start  1: tests_epee
 1/23 Test  #1: tests_epee .......................   Passed    0.00 sec
      Start  2: hash-target
 2/23 Test  #2: hash-target ......................   Passed    0.35 sec
      Start  3: wallet-crypto-bench
```

### DIFF
--- a/contrib/epee/CMakeLists.txt
+++ b/contrib/epee/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2014-2020, The Monero Project
+# Copyright (c) 2014-2021, The Monero Project
 # 
 # All rights reserved.
 # 
@@ -27,4 +27,8 @@
 # THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 add_subdirectory(src)
+
+if (BUILD_TESTS)
+	add_subdirectory(tests/src)
+endif()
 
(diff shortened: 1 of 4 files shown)


selsta:
I don't know yet what we should do with this. The epee tests have never been touched since project launch, all maintained tests are inside `monero/tests`.

IMO we should move the existing epee tests to `monero/tests` and remove the whole tests subfolder inside `monero/contrib/epee`.

jtgrassie:
`contrib/epee/tests` have a slightly different license (original author vs Monero project), hence an argument to leave where they are?

selsta:
epee tests were removed in #8211

PR #7856, opened 2021-08-12
refactor: move math functionality from epee to src

UkoeHB:
## Summary

- Moves math functions `median()` and `rolling_median_t` from `epee` to `src`. These two things are not being used outside of `src`.
- Adds test coverage of `median()` and `get_mid<T>()` to unit tests. Previously, as noted in #7854, the test for `median()` that lived in `epee` was not being compiled or executed by CI.
- Minor bug fix for the `get_median()` function in `epee`.
- **Note**: The `get_mid<T>()` function remains in `epee` because it used there.

This PR was inspired by @vtnerd's [comment](https://github.com/monero-project/monero/pull/7833#discussion_r683556281) that `epee` is focused on network-related things, so math stuff should go elsewhere.

This PR conflicts with #7854, so I am happy to hear advice on the best way forward.

## Future work

- Add an `n_choose_k` math function to support a different future PR (for multisig key gen).


contrib/epee/include/stats.inl
```diff
@@ -86,7 +87,7 @@ Tpod Stats<T, Tpod>::get_median() const
     }
     else
     {
-      median = (sorted[(sorted.size() - 1) / 2] + sorted[sorted.size() / 2]) / 2;
+      median = epee::misc_utils::get_mid(sorted[(sorted.size() - 1) / 2], sorted[sorted.size() / 2]);
```
UkoeHB, line 90: Bug fix: potential integer overflow.

UkoeHB: Update: moved to a new PR.

src/math/math_utils.h
```diff
@@ -0,0 +1,62 @@
...
+
+#include <algorithm>
+#include <vector>
+
+#include <boost/utility/value_init.hpp>
+
+#include "misc_language.h"
+
+
+namespace cryptonote
```
mj-xmr, line 39: I'd say that a namespace math would be better. Or even better cryptonote::math.

tests/core_tests/block_reward.cpp
```diff
@@ -77,7 +78,7 @@ namespace
     std::vector<size_t> block_weights;
     generator.get_last_n_block_weights(block_weights, get_block_hash(blk_prev), median_block_count);
 
-    size_t median = misc_utils::median(block_weights);
+    size_t median = cryptonote::median(block_weights);
```
mj-xmr, line 81: Since there's `using namespace cryptonote` already there, you can ommit it here.

UkoeHB: IIRC this gave me a compile error. A naming conflict with the variable `size_t median`.

mj-xmr: Right.
Then I insist on the math namespace.

mj-xmr:
_This PR conflicts with #7854, so I am happy to hear advice on the best way forward._

I don't mind. I use math there only as an example. 
But honestly I question a bit the need of moving the math outside epee.

UkoeHB:
I think it would be helpful to have a discussion about the future of epee. What _should_ it look like, and how can this PR be one step in the right direction?

mj-xmr:
> I think it would be helpful to have a discussion about the future of epee. What _should_ epee look like, and how can this PR be one step in the right direction?

I will just review the techicals through (soon) and leave the decision to others.

src/math/math_utils.h
```diff
@@ -0,0 +1,65 @@
...
+{
+namespace math
+{
+
+  template<class type_vec_type>
+  type_vec_type median(std::vector<type_vec_type> &v)
```
mj-xmr, line 45: It't not related to this PR, but this parameter should really be a copy, not a reference. I don't expect a function, that returns a median, to sort my vector. It will even not work at all, if the passed vector is `const`, forcing the caller to be non-const-correct.

src/math/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+set(math_sources
+  math.cpp)
+
+set(math_headers
+  rolling_median.h)
+
+set(math_private_headers
```
mj-xmr, line 35: Please use here and instead of 32 just:

`monero_find_all_headers(math_headers, "${CMAKE_CURRENT_SOURCE_DIR}")`

src/math/math.cpp
```diff
@@ -0,0 +1,40 @@
...
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "math_utils.h"
+#include "rolling_median.h"
```
mj-xmr, line 30: In my opinion this is correct, since this helps in syntax checks of templates' declarations before the parses leaves the component.

tests/unit_tests/math.cpp
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2021, The Monero Project
```
mj-xmr, line 1: With this file we have 2 problems:
- it's a copy-paste of [contrib/epee/tests/src/misc/test_math.h](https://github.com/monero-project/monero/blob/master/contrib/epee/tests/src/misc/test_math.h), leading to duplication of the test code
- the copyright has been changed from:
`Copyright (c) 2006-2013, Andrey N. Sabelnikov, www.sabelnikov.net` 
to:
` Copyright (c) 2021, The Monero Project`.

The second point concerns the new file `src/math/math_utils.h` as well.

src/math/math_utils.h
```diff
@@ -0,0 +1,65 @@
...
+    std::sort(v.begin(), v.end());
+    //nth_element(v.begin(), v.begin()+n-1, v.end());
+    if(v.size()%2)
+    {//1, 3, 5...
+      return v[n];
+    }else 
+    {//2, 4, 6...
+      return epee::misc_utils::get_mid<type_vec_type>(v[n-1],v[n]);
+    }
+  }
```
mj-xmr, line 62: No unintended changes spotted while moving the code.

mj-xmr (commented):
Except for small changes needed for `CMakeLists.txt`, the technicals are alright.

Copyright change is an issue IMO.

tests/unit_tests/math.cpp
UkoeHB: > it's (almost) a copy-paste of contrib/epee/tests/src/misc/test_math.h, leading to duplication of the test code

The test in `test_math.h` was deleted, so there is no duplication.

> the copyright has been changed from:

TBH I don't know the proper way to define copyrights... let me know what this file should say and I will fix it.

mj-xmr: I see now.

Well the copyrights have to stay as they were, even if this file is supposed to be moved.

UkoeHB: Do you need to add anything to the copyright if tests are added?

mj-xmr: > Do you need to add anything to the copyright if tests are added?

This is a good question, that I can't answer 100%.
The best way to go around this is to add a new file with Monero's copyright, that would contain any new tests, beside the ones written by Andrey.

UkoeHB: What about adding a line like this:
> // Parts of this file are originally copyright (c) 2006-2013, Andrey N. Sabelnikov, www.sabelnikov.net

That pattern is used in cryptonote files.

@selsta what do you think?

mj-xmr: This could do, since the licences are compatible.

selsta: Yes, add 
```
    // Parts of this file are originally copyright (c) 2006-2013, Andrey N. Sabelnikov, www.sabelnikov.net
```

to the file to be on the safe side.

tests/unit_tests/math.cpp
```diff
@@ -25,6 +25,8 @@
 // INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Parts of this file are originally copyright (c) 2006-2013, Andrey N. Sabelnikov, www.sabelnikov.net
```
mj-xmr, line 29: You need the same for the templated implementation.

src/math/math_utils.h
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2019-2021, The Monero Project
```
mj-xmr, line 1: ... here

mj-xmr (changes requested):
Copyright changed

src/math/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+set(math_sources
+  math.cpp)
+
+monero_find_all_headers(math_headers, "${CMAKE_CURRENT_SOURCE_DIR}")
+
+monero_private_headers(math
```
mj-xmr, line 34: Something to remove.

mj-xmr (changes requested):
Just that small nitpick in CMakeLists.txt an it's ready for me.

mj-xmr (approved):
You will be asked for squashing the commits before the branch can be merged. I'm not sure myself when the best time for this is.

mj-xmr (approved):
Re-reviewed the squashed commit.
All green.

UkoeHB:
Rebased (maybe not necessary in hindsight)

UkoeHB:
Follow up: it might be fine to move the math stuff to the `tools` namespace in `src/common`.

PR #7859, opened 2021-08-13
LMDB: protection against UB, by not calling virtual methods from destructor

mj-xmr:
The class `BlockchainLMDB` was calling pure virtual methods from its destructor, which leads to UB.

The PR retains the functionality by using private, non virtual counterparts, being able to be called from the destructor, without causing UB, while at the same time the public interface can be used, since the public overridden methods also call the private methods.

From 10.4 Abstract Classes parag. 6 in the Standard :

_"Member functions can be called from a constructor (or destructor) of an abstract class; the effect of making a virtual call to a pure virtual function directly or indirectly for the object being created (or destroyed) from such a constructor (or destructor) is undefined."_


src/blockchain_db/lmdb/db_lmdb.h
```diff
@@ -413,7 +420,7 @@ class BlockchainLMDB : public BlockchainDB
 
   bool prune_worker(int mode, uint32_t pruning_seed);
 
-  virtual bool is_read_only() const;
+  bool is_read_only() const;
```
mj-xmr, line 423: I'm not 100% sure if this change was required, but since the method is private anyway, I don't see a reason for it to be virtual anyway.

vtnerd (approved):

mj-xmr:
My bad. A small new change.

The virtual `is_read_only()` was actually a good clue. It **is** and overridden virtual method after all. My mind fooled me into thinking it's not, since it was private. 

I gave the method the same `_private` counterpart as with all the other touched methods.

Ping @vtnerd

PR #7861, opened 2021-08-14
LMDB: protection against UB, by not calling virtual methods from destructor - v0.17

mj-xmr:
For v0.17

For description, see: https://github.com/monero-project/monero/pull/7859


selsta (approved):

mj-xmr:
See small changes from https://github.com/monero-project/monero/pull/7859

PR #7866, opened 2021-08-16
Warnings: LMDB's virtual methods decorated with 'override'

mj-xmr:
In C++11, an overridden virtual method may be decorated with the `override` keyword, which delivers a compile-time check whether the code works as intended or not - that the method, that is expected to work polymorphic, will indeed work so. If, after such decoration, the code fails to compile, then either the base class' method wasn't marked as virtual, or there was a typo in either the base class or the overridden class. The `override` keyword binds these methods for good.

Ignoring this C++11 feature causes compilation warnings under CLang (although not in our CI for some reason).

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.h
+++ b/src/blockchain_db/lmdb/db_lmdb.h
@@ -185,131 +185,131 @@ public:
   BlockchainLMDB(bool batch_transactions=true);
   ~BlockchainLMDB();
 
-  virtual void open(const std::string& filename, const int mdb_flags=0);
+  virtual void open(const std::string& filename, const int mdb_flags=0) override;
 
-  virtual void close();
+  virtual void close() override;
 
-  virtual void sync();
+  virtual void sync() override;
 
-  virtual void safesyncmode(const bool onoff);
+  virtual void safesyncmode(const bool onoff) override;
 
-  virtual void reset();
+  virtual void reset() override;
 
-  virtual std::vector<std::string> get_filenames() const;
+  virtual std::vector<std::string> get_filenames() const override;
 
-  virtual bool remove_data_file(const std::string& folder) const;
+  virtual bool remove_data_file(const std::string& folder) const override;
 
-  virtual std::string get_db_name() const;
+  virtual std::string get_db_name() const override;
 
-  virtual bool lock();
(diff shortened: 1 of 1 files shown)


UkoeHB (approved):

UkoeHB:
IIRC [best practice](https://stackoverflow.com/a/39932616/14080627) is to use _either_ `virtual` _or_ `override`, but not both.

selsta:
> Ignoring this C++11 feature causes compilation warnings under CLang (although not in our CI for some reason).

FWIW I also don't get any warning about this.

vtnerd:
Marking the functions as `final` or the entire class as `final` (I think the latter is possible), will help optimizations in a few cases (de-virtualized function calls). Seems like a good time to investigate either.

> IIRC best practice is to use either virtual or override, but not both.

I don't see why this matters, except for the extra columns needed for the text.

UkoeHB:
> I don't see why this matters, except for the extra columns needed for the text.

If you're going to do something, might as well do it the best way possible.

vtnerd:
> If you're going to do something, might as well do it the best way possible (within reason).

The best technique is subjective here since its redundant to the compiler but (possibly) helpful to the programmer (most C++ programmers at this point should understand that `override` implies `virtual`). I don't think its worth stressing over either way, I doubt many programmers are hacking on a 80-character wide terminal.

mj-xmr:
> IIRC [best practice](https://stackoverflow.com/a/39932616/14080627) is to use _either_ `virtual` _or_ `override`, but not both.

If I'm to choose between them, I'd definitely stick to `override`, since it improves the safety, while `virtual` is just a loose pre-C++11 contract. Before C++11, it was even one of MISRA rules (from 2008) to add the `virtual` to the overridden methods, as this was the only way for the reader of the derived class' interface to know that the method was polymorphic, and for instance couldn't be used in constructors/destructors.

@vtnerd 
I'm not a big fan `final` as it prohibits further extension (obviously), but as long as it can always be removed in the future, if needed, I can set it here.

So Gents, I'm:
- adding final to the class
- removing `virtual` when there's `override`

mj-xmr:
Please let me know when it's good to squash.

selsta:
Superseded by #9925.

PR #7867, opened 2021-08-16
UB: virtual method in destructor of WalletImpl

mj-xmr:
Done for the same reasons as https://github.com/monero-project/monero/pull/7859 and using exactly the same mitigation technique.


selsta (commented):
miniupnp submodule changes seem unrelated

mj-xmr:
> miniupnp submodule changes seem unrelated

Thanks. Overlooked due to switching branches between master and v.0.17

mj-xmr:
The branch appears to fix the occasional (10% of cases on GH and my machine) `functional_tests_rpc` crash with the error message: "_Couldn't stop mining due to unknown error_".

[Here's a](https://github.com/mj-xmr/monero-mj/actions) set of 20 hourly runs of this test, showing no crashes:

image

selsta:
> Here's a set of 20 hourly runs of this test, showing no crashes:

These runs didn't run any tests at all.

mj-xmr:
> These runs didn't run any tests at all

Thanks for checking. Was in a hurry.

mj-xmr:
Another series of runs, this time done right:

image

selsta:
Something doesn't add up here. `api/wallet.cpp` isn't used at all in `functional_tests_rpc`.

mj-xmr:
> Something doesn't add up here. `api/wallet.cpp` isn't used at all in `functional_tests_rpc`.

One thing is, that I admit not having a great way of reproducing the error, which prevents the mining from being stopped, like we see through my above comments. The truth is, that so far nobody has.

However independently, this PR alone can fix things, that might be related to the above mining error or not, but other errors as well, that we are not even aware of.

PR #7873, opened 2021-08-18
node_server: fix race condition

perfect-daemon:


src/p2p/net_node.h
```diff
@@ -117,6 +117,18 @@ namespace nodetool
...
+      base_type(context),
+      fluff_txs{},
+      flush_time{},
+      peer_id(context.peer_id),
+      support_flags(context.support_flags),
+      m_in_timedsync(context.m_in_timedsync),
+      sent_addresses{}
+    {}
+    p2p_connection_context_t &operator=(const p2p_connection_context_t &context) {
+      return *new(this) p2p_connection_context_t(context);
```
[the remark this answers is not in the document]
perfect-daemon: Is it still memory leak with added `this->~p2p_connection_context_t` ?
Why is it undefined behaviour ?
What's about this place ?
https://github.com/monero-project/monero/blob/82149bfe441e56d5f9d72a13ffa25ce1df6d8391/contrib/epee/include/net/net_utils_base.h#L431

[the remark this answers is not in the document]
jagerman: It's undefined behaviour because it doesn't destruct any members in the current object, it just rams new initialization of top of the existing members.  So the vector and the set, for instance, are never destructed and their allocated memory is never freed.

An explicit constructor would indeed explicitly destruct, and would therefore solve the UB.  But you still *shouldn't* do either because it is still a horrible abuse of the purpose of placement new (and it is horrible where it is being done in epee, too).

jagerman:
What race condition is this fixing, and how/why does the explicit copy ctor & assignment operator fix it?

It's also worth pointing out that currently this is removing the implicit move ctor & move assignment (by adding explicit copy ctor/assignment), which seems like a bad thing because anywhere you try to move you'll now end up copying, *and* copying now clears fluff_txs, flush_time, and sent_addresses, which is highly unintuitive.

perfect-daemon:
> What race condition is this fixing, and how/why does the explicit copy ctor & assignment operator fix it?
> 

It's temporary workaround, efficiency loss due to lack of move constructor/assignment isn't a important since it isn't used in practice now.

> It's also worth pointing out that currently this is removing the implicit move ctor & move assignment (by adding explicit copy ctor/assignment), which seems like a bad thing because anywhere you try to move you'll now end up copying, _and_ copying now clears fluff_txs, flush_time, and sent_addresses, which is highly unintuitive.
>
Try to run the first commit under debugger.

src/p2p/net_node.h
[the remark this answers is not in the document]
vtnerd: Yeah if the constructor throws on line 131, then the object is in UB zombie state.

I'm also not a fan of using the copy-constructor and copy-assignment like this. Its not actually copying the source object. Just create a new function and call it on the single line (IIRC) that needs this functionality.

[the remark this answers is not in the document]
vtnerd: The other reason for a new function is that the base class might have the same problem at some point. So some properly named function would be useful.

vtnerd:
> It's temporary workaround, efficiency loss due to lack of move constructor/assignment isn't a important since it isn't used in practice now.

I made a similar comment while this is back-and-forth was going on. Just create a single new function for whatever this is doing. Because its not copy-construction or assignment.

perfect-daemon:
```cpp
#include <vector>
#include <thread>
int main() {
  struct obj_t {
    bool field1;
    std::vector<bool> field2;
  };
  obj_t obj;
  std::thread t1([&]{
    size_t counter = 0;
    while(true) {
      counter += obj_t(obj).field1;
    }
  });

  std::thread t2([&]{
    while(true) {
      if (obj.field2.size() > 1024 * 1024)
        obj.field2.clear();
      else
        obj.field2.emplace_back();
    }
  });
  t1.join();
  t2.join();
  return 0;
}
```
This problem was created with addition of field2 into obj_t.
Since It's difficult to remove field2 easily then workaround is to remove field2 from implicit methods of obj_t.
It isn't right to replace implicit methods with separate explicit methods, since the source of problem is addition of field2, but not usage of implicit methods of obj_t.

jagerman:
> ```c++
> #include <vector>
> #include <thread>
> // ...
> ```
> This problem was created with addition of field2 into obj_t.
> Since It's difficult to remove field2 easily then workaround is to remove field2 from implicit methods of obj_t.
> It isn't right to replace implicit methods with separate explicit methods, since the source of problem is addition of field2, but not usage of implicit methods of obj_t.

This is incorrect: the *source* of the problem is that this code is performing unsynchronized reading and writing of the same data from different threads.  Avoiding copying of the complex data structure (vector, set) may avoid a segfault (mainly because those involve memory allocations), but you *still* have a non-atomic copies happening with the other fields that aren't going to give you consistent results.  It won't segfault because the memory access is all well defined, but it could give you pretty much any values in the copy if some other thread is writing to the data around the same time you are trying to read it.

So the proper fix here is that the code needs to be redesigned so the same instance doesn't get shared across threads, or to use a mutex to guard instances that are shared.

perfect-daemon:
> So the proper fix here is that the code needs to be redesigned so the same instance doesn't get shared across threads, or to use a mutex to guard instances that are shared.
>
I know.
It's temporary fix for critical problem, that happens in practice.
Once design is clear, I'll know which general solution is to apply.

vtnerd:
>  Once design is clear, I'll know which general solution is to apply.

Then also, don't hijack the copy functions. The copy assignment race issue should be in only one spot IIRC.

perfect-daemon:
> Then also, don't hijack the copy functions. The copy assignment race issue should be in only one spot IIRC.
It isn't important currently.
Those fields shouldn't be added in the first place.
It isn't safe in anyway to copy any connection context now, since it may be used by another thread currently.

vtnerd:
>  Those fields shouldn't be added in the first place.

The Dandelion++ protocol requires a randomized flushing per connection - so why wouldn't the fields be placed in the object meant for tracking connection information? Same thing with the `sent_addresses` field. A separate data structure that needs to be synchronized (cleaned up) with the connection list? Arbitrary inheritance to limit what is copied?

>  It isn't safe in anyway to copy any connection context now, since it may be used by another thread currently.

It was never safe to copy that object at the point in question, that was the point by @jagerman . The context class never really had a strategy for synchronization. TSAN goes crazy on that class (I brought this up in the `for_each_connection` review), regardless of whether the fields with heap memory are present.

The Dandelion++ protocol uses a single strand to keep everything straight, but I missed the copy-assignment (I incorrectly thought the `std::vector` was only touched after construction in the Dandelion++ code).

perfect-daemon:
May you suggest fix that would be accept for you ?

perfect-daemon:
>Since It's difficult to remove field2 easily then workaround is to remove field2 from implicit methods of obj_t.
>
Turns I knew enough to implement it quickly. Thanks for useful suggestions.

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -764,6 +770,34 @@ namespace levin
     );
   }
 
+  void notify::on_connection_new(const boost::uuids::uuid &id, bool is_income)
+  {
+    if (!zone_)
+      return;
+
+    const auto zone = zone_;
```
vtnerd, line 778: We're now aliasing the purpose of the `connection_context` concept in another file, and technically the raciness of the original code was never fixed since the context copies everything including UUID objects, etc.

Just mark the copy+move constructors+operators as deleted, and add a named function that copies the useful fields for the async callback. Ideally the fields copied would be the bare minimum needed (the UUID probably doesn't actually need to be touched, etc).

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -764,6 +770,34 @@ namespace levin
...
+    zone_->strand.dispatch([zone, id, is_income]{
+      if (zone->contexts.find(id) != zone->contexts.end())
+        return;
+      zone->contexts[id] = {
```
vtnerd, line 782: A second lookup is unnecessary.

src/p2p/net_node.h
```diff
@@ -111,15 +111,11 @@ namespace nodetool
   struct p2p_connection_context_t: base_type //t_payload_net_handler::connection_context //public net_utils::connection_context_base
   {
     p2p_connection_context_t()
-      : fluff_txs(),
-        flush_time(std::chrono::steady_clock::time_point::max()),
-        peer_id(0),
+      : peer_id(0),
```
vtnerd, line 114: `sent_addresses` only works because it is only modified in receive functions that are in a strand, and the single `operator=` call also happens to only occur in that same strand. Add a comment about this, as its far easier to botch than the Dandelion++ fields that were removed.

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -764,6 +770,34 @@ namespace levin
     );
   }
 
+  void notify::on_connection_new(const boost::uuids::uuid &id, bool is_income)
+  {
+    if (!zone_)
+      return;
+
+    const auto zone = zone_;
+    zone_->strand.dispatch([zone, id, is_income]{
```
vtnerd, line 779: This copies `zone` twice, how about an explicit struct to prevent that ref-count destroying change? Its not really much code.

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -764,6 +770,34 @@ namespace levin
...
+    });
+  }
+
+  void notify::on_connection_close(const boost::uuids::uuid &id)
+  {
+    if (!zone_)
+      return;
+
+    const auto zone = zone_;
+    zone_->strand.dispatch([zone, id]{
```
vtnerd, line 796: Same here.

src/cryptonote_protocol/levin_notify.cpp
vtnerd: Or let me re-state my argument -

The problematic line is the single (implicit) copy-assignment call. The copy+move constructor+assignment should (almost certainly in hindsight) be deleted from the context object since it is read+write by multiple threads. Or a thread-safe explicit version should be created. Its too easy to add a field and be unaware that there's an unsynchronized copy that occurs elsewhere due to the implicit function.

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -764,6 +770,34 @@ namespace levin
     );
   }
 
+  void notify::on_connection_new(const boost::uuids::uuid &id, bool is_income)
```
vtnerd, line 773: Rename `on_handshake_complete` as that is equally important to the connection status, and refactor could easily lose this pertinent information.

src/cryptonote_protocol/levin_notify.cpp
vtnerd: I went back through the code again, and this suggestion is a pain but doable. The duplicate `context` class and tracking logic is unfortunate, but might be the lesser of evils at the moment. Ideally those values are in the "real" context, with copying/moving disabled on that class.

I'll investigate if that call sequence with the copies can be re-thought. The context object is huge to be copied, when only like 5 trivial fields max are needed.

elibroftw:
In what case would the threads created in the this new test not join? My branch is failing only this test, even though all I did was refactor the variables used in two header files. https://github.com/elibroftw/monero/runs/4281540443?check_suite_focus=true

PR #7874, opened 2021-08-18
node_server: fix race condition

perfect-daemon:


jagerman:
(discussion in #7873)

selsta (approved):

PR #7875, opened 2021-08-18
Device: remove unused and incorrect non-default constructor

mj-xmr:
See https://github.com/monero-project/monero/pull/7875#issuecomment-902196927

Previous description:

The variable `mode` in `device` was being uninitialized when the non-empty constructor was being used. By delegating the constructor to the default constructor (C++11 feature), where the variable is initialized, I get the variable initialized and prevent code duplication.

### DIFF
--- a/src/device/device.hpp
+++ b/src/device/device.hpp
@@ -91,7 +91,6 @@ namespace hw {
     public:
 
         device(): mode(NONE)  {}
-        device(const device &hwdev) {}
         virtual ~device()   {}
 
         explicit virtual operator bool() const = 0;


UkoeHB (approved):

selsta (commented):
Only the default constructor is used in the codebase. Would it be better to just delete the other one?

mj-xmr:
> Only the default constructor is used in the codebase. Would it be better to just delete the other one?

I'd say delete. The code compiles fine without it.
I can't find any online references for this, but in my code, I rarely set expectations on the construction structure of the derived classes via enforcement of special (non-default) constructors in the base class (interface), like it's done here. If I do, then I'd expect the passed reference to at least be accessible in the derived via a protected method. Since it's not being done here, the non-default constructor violates the rule of least astonishment: You'd expect it to do something, while it does something else or nothing (as in this case).

Unless there's a strong opinion to keep the non-default constructor soon, I will delete it.

selsta:
Would say delete.

mj-xmr:
> Would say delete.

Done.

selsta (approved):

PR #7876, opened 2021-08-18
Ledger: member 'mode' was shadowing that of the base class

mj-xmr:
The base class `device` already declares the variable `mode`. By redeclaring it in `device_ledger` class the base class' (protected) `mode` is being shadowed and it's not certain if the ledger wants to modify (or use) the own instance or the bases. For sure it doesn't use the base' one.

For the reason that I can't test this, I'm setting the branch as a draft.

@ph4r05 Would you be able to test it coincidentally?


selsta:
@ph4r05 is Trezor only

mj-xmr:
I took a closer look at where and how the subclass' variable `mode` is used, and this is what I found in `device_ledger.cpp`:

```C++
    bool  device_ledger::set_mode(device_mode mode) {
        AUTO_LOCK_CMD();

        int offset;

        switch(mode) {
        case TRANSACTION_CREATE_REAL:
        case TRANSACTION_CREATE_FAKE:
          offset = set_command_header_noopt(INS_SET_SIGNATURE_MODE, 1);
          //account
          this->buffer_send[offset] = mode;
          offset += 1;

          this->buffer_send[4] = offset-5;
          this->length_send = offset;
          this->exchange();

          this->mode = mode;
          break;

        case TRANSACTION_PARSE: 
        case NONE:
          this->mode = mode;
          break;
        default:
           CHECK_AND_ASSERT_THROW_MES(false, " device_ledger::set_mode(unsigned int mode): invalid mode: "<<mode);
        }
        MDEBUG("Switch to mode: " <<mode);
        return device::set_mode(mode);
    }
```
I have the following observations:
- for each case, the mode is being set, except for the default one, where an exception is thrown
- at the same time the parent's variable `mode` is also set at the end of the method, which is consistent with the behavior above.
- the subclass' variable itself is never used to increment/decrement anything, only to set things

I conclude, that there shouldn't be any side effects of this cleanup.

PR #7884, opened 2021-08-19
support cors wildcard

### LINKED ISSUE #5172: Add wildcard support to CORS Access-Control-Allow-Origin header in Daemon RPC
This issue requests wildcard support in the CORS Access-Control-Allow-Origin header in Daemon RPC.  Doing so would allow web apps to query a Monero daemon to enable fully client-side web wallets.  Since the daemon is a neutral provider of information about the Monero blockchain, I see no reason for it to discriminate against clients by not allowing web clients.

woodser:
Resolves #5172.

### DIFF
--- a/contrib/epee/include/net/http_protocol_handler.inl
+++ b/contrib/epee/include/net/http_protocol_handler.inl
@@ -668,7 +668,7 @@ namespace net_utils
 		// Cross-origin resource sharing
 		if(m_query_info.m_header_info.m_origin.size())
 		{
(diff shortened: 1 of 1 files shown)


vtnerd:
Should this be `std::binary_search(...begin(), ...end(), "*")`? This won't work in some situations where multiple arguments are provided on the CLI, however, I don't see a point if the wildcard is provided.

woodser:
> Should this be `std::binary_search(...begin(), ...end(), "*")`? This won't work in some situations where multiple arguments are provided on the CLI, however, I don't see a point if the wildcard is provided.

Sure, I changed it to search for wildcard.

vtnerd (approved):

PR #7886, opened 2021-08-20
Mac: fix unit test bind_same_p2p_port

mj-xmr:
After fixing the random failure under Ubuntu of the test `bind_same_p2p_port` belonging to the suite `node_server`, the same test under Mac OSX ceased to work completely. @TimSycamore noted, that under OSX, the IP has to stay 127.0.0.1, and not 127.0.0.2.

There are some temporary commits in this PR to prove the point - that the UTs under OSX were failing without the IP fix. Also the unit tests for Mac are temporarily enabled, but maybe it wouldn't be a bad idea to run them regularly either, since they don't cost that much time (< 2 minutes)? Of course enabling them for good shall be done in a separate PR.

### DIFF
--- a/tests/unit_tests/node_server.cpp
+++ b/tests/unit_tests/node_server.cpp
@@ -303,6 +303,9 @@ TEST(node_server, bind_same_p2p_port)
 
     Relevant part about REUSEADDR from man:
     https://www.man7.org/linux/man-pages/man7/ip.7.html
+
+    For Mac OSX, set the following alias, before running the test, or else it will fail:
+    sudo ifconfig lo0 alias 127.0.0.2
     */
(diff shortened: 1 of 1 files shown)


selsta:
I would undo the changes to CI – it's supposed to test compilation and not if tests pass. Merging the comment is fine until a different solution is found.

mj-xmr:
> I would undo the changes to CI – it's supposed to test compilation and not if tests pass. Merging the comment is fine until a different solution is found.

Done.
It tests the compilation now only, but of course not compilation of the tests, which is default. If the tests should be compiled as well, we'd need to have this line back:
`cmake -S . -B build -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=release && cmake --build build -j3`

selsta (approved):

PR #7891, opened 2021-08-26
RPC and ZeroMQ APIs to support p2pool

SChernykh:
Adds the following:

- "get_miner_data" to RPC API
- "json-full-miner_data" to ZeroMQ subscriber contexts

Both provide the necessary data to create a custom block template. They are used by p2pool.

Data provided:

- major fork version
- current height
- previous block id
- RandomX seed hash
- network difficulty
- median block weight
- coins mined by the network so far
- mineable mempool transactions

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -1238,6 +1238,12 @@ bool Blockchain::switch_to_alternative_blockchain(std::list<block_extended_info>
     reorg_notify->notify("%s", std::to_string(split_height).c_str(), "%h", std::to_string(m_db->height()).c_str(),
         "%n", std::to_string(m_db->height() - split_height).c_str(), "%d", std::to_string(discarded_blocks).c_str(), NULL);
 
+  crypto::hash prev_id;
+  if (!get_block_hash(alt_chain.back().bl, prev_id))
+    MERROR("Failed to get block hash of an alternative chain's tip");
+  else
+    send_miner_notifications(prev_id, alt_chain.back().already_generated_coins);
+
   for (const auto& notifier : m_block_notifiers)
   {
     std::size_t notify_height = split_height;
@@ -1782,6 +1788,30 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
 bool Blockchain::create_block_template(block& b, const account_public_address& miner_address, difficulty_type& diffic, uint64_t& height, uint64_t& expected_reward, const blobdata& ex_nonce, uint64_t &seed_height, crypto::hash &seed_hash)
 {
   return create_block_template(b, NULL, miner_address, diffic, height, expected_reward, ex_nonce, seed_height, seed_hash);
+}
+//------------------------------------------------------------------
+bool Blockchain::get_miner_data(uint8_t& major_version, uint64_t& height, crypto::hash& prev_id, crypto::hash& seed_hash, difficulty_type& difficulty, uint64_t& median_weight, uint64_t& already_generated_coins, std::vector<tx_block_template_backlog_entry>& tx_backlog)
+{
+  prev_id = m_db->top_block_hash(&height);
+  ++height;
+
+  major_version = m_hardfork->get_ideal_version(height);
+
+  seed_hash = crypto::null_hash;
+  if (m_hardfork->get_current_version() >= RX_BLOCK_VERSION)
+  {
+    uint64_t seed_height, next_height;
+    crypto::rx_seedheights(height, &seed_height, &next_height);
+    seed_hash = get_block_id_by_height(seed_height);
+  }
+
+  difficulty = get_difficulty_for_next_block();
+  median_weight = m_current_block_cumul_weight_median;
+  already_generated_coins = m_db->get_block_already_generated_coins(height - 1);
+
+  m_tx_pool.get_block_template_backlog(tx_backlog);
+
+  return true;
 }
 //------------------------------------------------------------------
 // for an alternate chain, get the timestamps from the main chain to complete
@@ -4362,6 +4392,7 @@ leave:
   get_difficulty_for_next_block(); // just to cache it
   invalidate_block_template_cache();
 
+  send_miner_notifications(id, already_generated_coins);
 
   for (const auto& notifier: m_block_notifiers)
     notifier(new_height - 1, {std::addressof(bl), 1});
@@ -5270,12 +5301,21 @@ void Blockchain::set_user_options(uint64_t maxthreads, bool sync_on_blocks, uint
   m_max_prepare_blocks_threads = maxthreads;
 }
 
-void Blockchain::add_block_notify(boost::function<void(std::uint64_t, epee::span<const block>)>&& notify)
+void Blockchain::add_block_notify(BlockNotifyCallback&& notify)
 {
   if (notify)
   {
     CRITICAL_REGION_LOCAL(m_blockchain_lock);
     m_block_notifiers.push_back(std::move(notify));
+  }
+}
+
+void Blockchain::add_miner_notify(MinerNotifyCallback&& notify)
+{
+  if (notify)
+  {
+    CRITICAL_REGION_LOCAL(m_blockchain_lock);
+    m_miner_notifiers.push_back(std::move(notify));
   }
 }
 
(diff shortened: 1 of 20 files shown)


src/cryptonote_core/blockchain.cpp
```diff
@@ -5531,6 +5571,33 @@ void Blockchain::cache_block_template(const block &b, const cryptonote::account_
   m_btc_valid = true;
 }
 
+void Blockchain::send_miner_notifications(const crypto::hash &prev_id, uint64_t already_generated_coins)
+{
+  if (m_miner_notifiers.empty())
+    return;
+
+  const uint64_t height = m_db->height();
```
vtnerd, line 5579: Would be nice if some of these values could be retrieved lazily only when there is a sub to the corresponding endpoint. But this notification is infrequent, so I guess not worth butchering the interface -

Could you add a comment here about lazily retrieving though? Just to document somewhere that the data (usually) gets dropped on the floor. The notifier signature could be updated to `boost::optional<miner_data>(const crypto::hash&, std::uint64_t /* already_generated_coins */, blockchain&, const miner_data*)`, which is a bit more efficient but admittedly funky.

Basically the first notifier that has a subscriber retrieves the data and returns it for the next one, etc.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -913,6 +913,34 @@ namespace cryptonote
...
+  void tx_memory_pool::get_block_template_backlog(std::vector<tx_block_template_backlog_entry>& backlog, bool include_sensitive) const
+  {
+    CRITICAL_REGION_LOCAL(m_transactions_lock);
+    CRITICAL_REGION_LOCAL1(m_blockchain);
+    const relay_category category = include_sensitive ? relay_category::all : relay_category::broadcasted;
+    backlog.reserve(m_blockchain.get_txpool_tx_count(include_sensitive));
+    cryptonote::blobdata tmp_blob;
+    txpool_tx_meta_t tmp_meta;
+    m_blockchain.for_all_txpool_txes([this, &backlog, &tmp_blob, &tmp_meta](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref *bd){
+      tmp_blob = m_blockchain.get_txpool_tx_blob(txid, relay_category::all);
```
vtnerd, line 925: If the `false` on line 941 is changed to `true` then this call is unnecessary because `bd` is a pointer to the transaction blob. This should be a lot more efficient.

src/rpc/core_rpc_server.cpp
```diff
@@ -1861,6 +1861,43 @@ namespace cryptonote
...
+      LOG_ERROR("Failed to get miner data");
+      return false;
+    }
+
+    res.tx_backlog.clear();
+    res.tx_backlog.reserve(tx_backlog.size());
+
+    for (const auto& entry : tx_backlog)
+    {
+      res.tx_backlog.emplace_back(COMMAND_RPC_GETMINERDATA::response::tx_backlog_entry{string_tools::pod_to_hex(entry.id), entry.weight, entry.fee});
```
vtnerd, line 1890: Is the only difference between the types a `std::string` to be compliant with the the epee JSON output engine? That is a glaring mistake in the internal epee DOM (not differentiating between strings and bytes).

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -940,6 +940,56 @@ namespace cryptonote
     typedef epee::misc_utils::struct_init<response_t> response;
   };
 
+  struct COMMAND_RPC_GETMINERDATA
+  {
+    struct request_t: public rpc_request_base
```
vtnerd, line 945: I don't see a corresponding change to ZMQ-RPC. Add in a later patch, or ? Because I think this requires pool implementations to use both interfaces (at least for the initialization).

vtnerd (commented):
This looks basically ready - comment in `tx_pool` is most "critical" (yet can still be cleanup later). Changing a few lines should make retrieving the tx backlog a bit more efficient.

src/rpc/zmq_pub.cpp
```diff
@@ -116,13 +119,30 @@ namespace
...
   {
     const cryptonote::transaction& tx;
+    crypto::hash hash;
+    uint64_t blob_size;
+    uint64_t weight;
+    uint64_t fee;
   };
 
-  void toJsonValue(rapidjson::Writer<epee::byte_stream>& dest, const minimal_chain self)
+  void toJsonValue(rapidjson::Writer<epee::byte_stream>& dest, const minimal_chain& self)
```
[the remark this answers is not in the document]
SChernykh: The `&` was intentionally added back as passing by reference is still cheaper than copying 24 bytes, also to make all `toJsonValue` functions in zmq_pub.cpp follow the same standard.

src/rpc/zmq_pub.cpp
```diff
@@ -116,13 +119,30 @@ namespace
     const epee::span<const cryptonote::block> blocks;
   };
 
+  //! Object for miner data serialization
+  struct miner_data
```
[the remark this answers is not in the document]
SChernykh: I don't see why when other structs are right next to it. Removing `&` will make a copy of the vector which is bad for performance. It is currently created in one place and never copied, only the reference is passed around. This vector can contain hundreds of transactions.

src/rpc/core_rpc_server.cpp
SChernykh: JSON outputs binary data here for entry.id if I don't convert to string explicitly.

src/rpc/core_rpc_server_commands_defs.h
SChernykh: ZMQ-RPC doesn't have half of needed APIs - not even get_block_template/submit_block, so pool implementations are forced to use regular RPC.

src/cryptonote_core/blockchain.cpp
SChernykh: `m_miner_notifiers` is not empty only when there is `--zmq-pub` in monerod command line, so it early outs in 99% cases. The remaining monerod setups that do use `--zmq-pub` most likely will use it for p2pool, so they will use this data. Refactoring it makes sense later when there are more uses for ZMQ.

vtnerd: Mining is not the only use case for pub/sub, so there will be cases where all of this information is collected only to be dropped. The other endpoints are forwarding data that is already available.

src/rpc/zmq_pub.cpp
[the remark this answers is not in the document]
vtnerd: The `&` would not be a copy because the entire `struct` would be passed by reference. And it would cleanup the long parameter list.

src/rpc/core_rpc_server_commands_defs.h
vtnerd: Seems like something to add in another PR then.

src/cryptonote_core/blockchain.cpp
SChernykh: Is there an easy way to detect if anyone is connected to zmq-pub port and subscribed to `json-full-miner_data`? I tried to do something like this initially, but I couldn't find a way to detect connected subscribers reliably. IIRC ZMQ somehow cached it internally and `add_subscriptions` was only called once even when I connected two subscribers, or something like that.

src/rpc/zmq_pub.cpp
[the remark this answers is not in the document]
vtnerd: I re-checked this - the ASM generated should be the same currently, so might as well make the change.

If the `struct` were just 2 fields, then each field could be passed in a register. The only caller doesn't have a `minimal_chain` object in working memory, so its copying the data anyway, its just a matter of how and where. However, this is caller specific, and there's a clause I missed at the end of the ABI document that restricts when this type of optimization can occur (from 4 fields to just 2 fields).

src/cryptonote_core/blockchain.cpp
vtnerd: The `miner_subs_` array should be an accurate count. ZMQ may coalesce TCP connections on the client side, so you need to create two `zmq_context` objects or use a separate process entirely. If this still doesn't work, then I'd have to re-read the documentation because I misunderstood how `XPUB` works.

src/cryptonote_core/blockchain.cpp
```diff
@@ -4362,6 +4392,7 @@ bool Blockchain::handle_block_to_main_chain(const block& bl, const crypto::hash&
   get_difficulty_for_next_block(); // just to cache it
   invalidate_block_template_cache();
 
+  send_miner_notifications(id, m_db->get_block_already_generated_coins(m_db->height() - 1));
```
j-berman, line 4395: Is it not safe to use the `already_generated_coins` already in memory here?

SChernykh: I actually didn't see it. I'll test it and if it has the same value I'll change the code.

src/rpc/zmq_pub.cpp
```diff
@@ -143,19 +163,27 @@ namespace
...
-    if (!get_transaction_hash(self.tx, id, blob_size))
-    {
-      MERROR("ZMQ/Pub failure: get_transaction_hash");
-      return;
-    }
+    dest.StartObject();
+    INSERT_INTO_JSON_OBJECT(dest, major_version, self.major_version);
+    INSERT_INTO_JSON_OBJECT(dest, height, self.height);
+    INSERT_INTO_JSON_OBJECT(dest, prev_id, self.prev_id);
+    INSERT_INTO_JSON_OBJECT(dest, seed_hash, self.seed_hash);
```
j-berman, line 172: Thoughts on treating `seed_hash` as optional if `major_version >= RX_BLOCK_VERSION` (here & in other parts of the code too) just in case the endpoint is being used for earlier blocks too?

SChernykh: It's always used for the current blockchain height, not for any earlier blocks. Even if it's used for major_version before that, seed_hash will just be all 0, but it only matters for testnet experiments (mining older blocks for example).

j-berman: Sounds fine to me :)

PR #7899, opened 2021-08-27
wallet: fix unused lambda capture warning

selsta:
```
/Users/selsta/dev/monero-gui/monero/src/wallet/wallet2.cpp:2821:33: warning: lambda capture 'i' is not used [-Wunused-lambda-capture]
      tpool.submit(&waiter, [&, i, n_vouts, txidx](){ geniod(tx, n_vouts, txidx); }, true);
                              ~~^
1 warning generated.
```

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -2860,7 +2860,7 @@ void wallet2::process_parsed_blocks(uint64_t start_height, const std::vector<cry
       THROW_WALLET_EXCEPTION_IF(txidx >= tx_cache_data.size(), error::wallet_internal_error, "txidx out of range");
       const cryptonote::transaction& tx = parsed_blocks[i].block.miner_tx;
       const size_t n_vouts = (m_refresh_type == RefreshType::RefreshOptimizeCoinbase && tx.version < 2) ? 1 : tx.vout.size();
-      tpool.submit(&waiter, [&, i, n_vouts, txidx](){ geniod(tx, n_vouts, txidx); }, true);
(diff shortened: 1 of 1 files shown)


src/wallet/wallet2.cpp
```diff
@@ -2860,7 +2860,7 @@ void wallet2::process_parsed_blocks(uint64_t start_height, const std::vector<cry
       THROW_WALLET_EXCEPTION_IF(txidx >= tx_cache_data.size(), error::wallet_internal_error, "txidx out of range");
       const cryptonote::transaction& tx = parsed_blocks[i].block.miner_tx;
       const size_t n_vouts = (m_refresh_type == RefreshType::RefreshOptimizeCoinbase && tx.version < 2) ? 1 : tx.vout.size();
-      tpool.submit(&waiter, [&, i, n_vouts, txidx](){ geniod(tx, n_vouts, txidx); }, true);
+      tpool.submit(&waiter, [&, tx, n_vouts, txidx](){ geniod(tx, n_vouts, txidx); }, true);
```
UkoeHB, line 2863: Is this making a copy of the tx, or only a copy of the reference?

SChernykh: Good catch, it is actually copied: https://stackoverflow.com/questions/19676473/in-a-lambda-how-reference-is-being-captured-by-value

SChernykh: `[&, n_vouts, txidx]` should be better.

selsta: Thanks, updated.

UkoeHB (approved):

mj-xmr (approved):

PR #7906, opened 2021-08-29
[release-v0.17] miniupnp: set submodule to upstream, disable install 

selsta:
Should fix the issue with having to run `git submodule sync`.


mj-xmr (approved):

mj-xmr:
Checked that it still compiles.

PR #7912, opened 2021-08-31
gitian: remove ccache

hyc:
It's just wasted overhead since the build VMs are always deleted
and recreated fresh for each run.

### DIFF
--- a/contrib/gitian/gitian-android.yml
+++ b/contrib/gitian/gitian-android.yml
@@ -24,7 +24,6 @@ packages:
 - "ca-certificates"
 - "python"
 - "cmake"
-- "ccache"
 - "protobuf-compiler"
 - "libdbus-1-dev"
 - "libharfbuzz-dev"
--- a/contrib/gitian/gitian-freebsd.yml
+++ b/contrib/gitian/gitian-freebsd.yml
@@ -25,7 +25,6 @@ packages:
 - "ca-certificates"
 - "python"
 - "cmake"
-- "ccache"
 - "protobuf-compiler"
 - "libdbus-1-dev"
 - "libharfbuzz-dev"
--- a/contrib/gitian/gitian-linux.yml
+++ b/contrib/gitian/gitian-linux.yml
@@ -36,7 +36,6 @@ packages:
 - "ca-certificates"
 - "python"
 - "cmake"
-- "ccache"
 - "protobuf-compiler"
 - "libdbus-1-dev"
 - "libharfbuzz-dev"


mj-xmr (approved):
Agreed.

No more `ccache` instances found in the gitian config files.

mj-xmr:
Why was the recent change necessary? Was there some issue with Gitian without it, like a warning of missing package? Or was it an error? Is it all working now from a clean environment?

hyc:
Looks like just leftover cruft that was never needed. Yes, it all builds fine from a clean environment.

contrib/depends/packages/packages.mk
```diff
@@ -1,15 +1,18 @@
...
 hardware_packages := hidapi protobuf libusb
 hardware_native_packages := native_protobuf
 
 android_native_packages = android_ndk
 android_packages = ncurses readline sodium
 
-darwin_native_packages = native_biplist native_ds_store native_mac_alias $(hardware_native_packages)
-darwin_packages = sodium ncurses readline $(hardware_packages)
+darwin_native_packages = $(hardware_native_packages)
+darwin_packages = ncurses readline sodium $(hardware_packages)
```
mj-xmr, line 15: Just out of curiosity: does the difference in ordering change something here?

mj-xmr (approved):

contrib/depends/packages/packages.mk
hyc: No difference, I was just lining this up to be uniform with the other platforms.

mj-xmr:
@selsta any chance for having this merged before the next release, please?

contrib/gitian/gitian-osx.yml
```diff
@@ -93,7 +93,7 @@ script: |
   # Build dependencies for each host
   export TAR_OPTIONS=--mtime=2000-01-01T12:00:00
   for i in $HOSTS; do
-    make ${MAKEOPTS} -C ${BASEPREFIX} HOST="${i}"
+    make ${MAKEOPTS} -C ${BASEPREFIX} HOST="${i}" GITIAN=1
```
selsta, line 96: Ideally it should only be `GITIAN=1` if it gets launched from `gitian-build.py`.

There is a (work in progress) dockerfile that takes `gitian-*.yml` files as input and that also uses ccache, if `GITIAN=1` is hardcoded it would lose ccache functionality.

hyc: Is that also using the gbuild script from https://github.com/devrandom/gitian-builder.git or is it using something else?

hyc: OK, if it's not using gbuild then I can switch on one of gbuild's env vars instead.

selsta (approved):

PR #7926, opened 2021-09-08
Fix export_transfers amount for send-to-self outputs

jagerman:
When you send a transaction to yourself then `export_transfers` outputs the wrong amount (and thus balance) in the exported file.

This happens because the unconfirmed transfer only has the actual change amount constructed in the transaction, but proper balance tracking requires identifying *all* outputs returning to the same wallet as change for the purposes of export details.

This fixes the issue by updating the change amount in the transaction to be the sum of all received outputs, instead of just the constructed change value.

(Note that the same issue does *not* occur if we receive the transaction in another copy of the wallet -- in that case all the outputs to self are already counted as change).

I've attached export csv files demonstrating the problem and fix, using a transfer of 6 outputs: 3 to my own wallet of values 0.01, 0.02, 0.03; and 3 to another wallet with the same amounts:
[0-premined.csv](https://github.com/monero-project/monero/files/7131970/0-premined.csv) - the export before the transfer is mined, showing an amount of 0.14 out.  (This is sort of wrong, but happens because unconfirmed transfer details do not look for transfers-to-self.  This is not affected by this PR.).
[1-mined.csv](https://github.com/monero-project/monero/files/7131971/1-mined.csv) - a mined output, showing the wrong amount (shows 0.14 out instead of 0.07 out).
[2-rescan_bc.csv](https://github.com/monero-project/monero/files/7131973/2-rescan_bc.csv) - showing what you get if you `rescan_bc` (or export from another wallet), showing the correct 0.07 out.
[3-fixed.csv](https://github.com/monero-project/monero/files/7131974/3-fixed.csv)- another transfer with this patch applied, using the same recipients and amounts as before, this time showing the correct 0.07 out (post-mining).

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1979,8 +1979,9 @@ void wallet2::process_new_transaction(const crypto::hash &txid, const cryptonote
   PERF_TIMER(process_new_transaction);
   // In this function, tx (probably) only contains the base information
   // (that is, the prunable stuff may or may not be included)
+  confirmed_transfer_details* just_confirmed = nullptr;
   if (!miner_tx && !pool)
-    process_unconfirmed(txid, tx, height);
+    just_confirmed = process_unconfirmed(txid, tx, height);
 
   // per receiving subaddress index
   std::unordered_map<cryptonote::subaddress_index, uint64_t> tx_money_got_in_outs;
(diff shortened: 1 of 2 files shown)


selsta:
This breaks `functional_tests_rpc`

```
[TEST STARTED] transfer
Resetting blockchain
Creating wallets
Mining some blocks
Creating transfer to self
Checking short payment IDs cannot be used when not in an integrated address
Checking long payment IDs are rejected
Checking empty destination is rejected
Creating transfer to another, manual relay
Creating multi out transfer
  File "/home/runner/work/monero/monero/tests/functional_tests/transfer.py", line 837, in <module>
    TransferTest().run_test()
  File "/home/runner/work/monero/monero/tests/functional_tests/transfer.py", line 45, in run_test
    self.transfer()
  File "/home/runner/work/monero/monero/tests/functional_tests/transfer.py", line 374, in transfer
    assert e.amount == amount
AssertionError
[TEST FAILED] transfer
```

PR #7928, opened 2021-09-09
CI: ccache common settings variable and apt tidying

mj-xmr:
There was a mixture of `ccache` settings set via env vars and command line. This PR unifies these settings and reuses them as a variable. Also made a tiny cleanup of the apt settings.

# OUTDATED (see comments to learn why): 

`hash_dir=false` helps in reusing the cache, when your project is named differently than the original one.

# References:
https://ccache.dev/manual/latest.html

## hash_dir (CCACHE_HASHDIR or CCACHE_NOHASHDIR, see Boolean values above)

If true (which is the default), ccache will include the current working directory (CWD) in the hash that is used to distinguish two compilations when generating debug info (compiler option -g with variations). Exception: The CWD will not be included in the hash if base_dir is set (and matches the CWD) and the compiler option -fdebug-prefix-map is used. See also the discussion under Compiling in different directories.

https://ccache.dev/manual/latest.html#_compiling_in_different_directories

## Compiling in different directories

Some information included in the hash that identifies a unique compilation can contain absolute paths:

-    The preprocessed source code may contain absolute paths to include files if the compiler option -g is used or if absolute paths are given to -I and similar compiler options.

-    Paths specified by compiler options (such as -I, -MF, etc) on the command line may be absolute.

-    The source code file path may be absolute, and that path may substituted for __FILE__ macros in the source code or included in warnings emitted to standard error by the preprocessor.

This means that if you compile the same code in different locations, you can’t share compilation results between the different build directories since you get cache misses because of the absolute build directory paths that are part of the hash.

Here’s what can be done to enable cache hits between different build directories:

-    If you build with -g (or similar) to add debug information to the object file, you must either:

-   use the compiler option -fdebug-prefix-map=<old>=<new> for relocating debug info to a common prefix (e.g. -fdebug-prefix-map=$PWD=.); or

-    set hash_dir = false.

-    If you use absolute paths anywhere on the command line (e.g. the source code file path or an argument to compiler options like -I and -MF), you must set base_dir to an absolute path to a “base directory”. Ccache will then rewrite absolute paths under that directory to relative before computing the hash.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,19 +6,20 @@ on: [push, pull_request]
 env:
   REMOVE_BUNDLED_BOOST : rm -rf /usr/local/share/boost
   BUILD_DEFAULT_LINUX: |
-        ccache --max-size=150M
         cmake -S . -B build -D ARCH="default" -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=release && cmake --build build -j3
   APT_INSTALL_LINUX: 'sudo apt -y install build-essential cmake libboost-all-dev miniupnpc libunbound-dev graphviz doxygen libunwind8-dev pkg-config libssl-dev libzmq3-dev libsodium-dev libhidapi-dev libnorm-dev libusb-1.0-0-dev libpgm-dev libprotobuf-dev protobuf-compiler ccache'
   APT_SET_CONF: |
-        echo "Acquire::Retries \"3\";" | sudo tee -a /etc/apt/apt.conf.d/80-custom
+        echo "Acquire::Retries \"3\";"         | sudo tee -a /etc/apt/apt.conf.d/80-custom
         echo "Acquire::http::Timeout \"120\";" | sudo tee -a /etc/apt/apt.conf.d/80-custom
-        echo "Acquire::ftp::Timeout \"120\";" | sudo tee -a /etc/apt/apt.conf.d/80-custom
+        echo "Acquire::ftp::Timeout \"120\";"  | sudo tee -a /etc/apt/apt.conf.d/80-custom
+  CCACHE_SETTINGS: |
+        ccache --max-size=150M
+        ccache --set-config=compression=true
 
 jobs:
   build-macos:
     runs-on: macOS-latest
     env:
-      CCACHE_COMPRESS: 1
       CCACHE_TEMPDIR: /tmp/.ccache-temp
     steps:
     - uses: actions/checkout@v1
@@ -33,13 +34,12 @@ jobs:
       run: HOMEBREW_NO_AUTO_UPDATE=1 brew install boost hidapi zmq libpgm miniupnpc ldns expat libunwind-headers protobuf ccache
     - name: build
       run: |
-        ccache --max-size=150M
+        ${{env.CCACHE_SETTINGS}}
         make -j3
 
   build-windows:
     runs-on: windows-latest
     env:
-      CCACHE_COMPRESS: 1
       CCACHE_TEMPDIR: C:\Users\runneradmin\.ccache-temp
       CCACHE_DIR: C:\Users\runneradmin\.ccache
     defaults:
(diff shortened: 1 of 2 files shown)


cirocosta:
neat, i wasn't aware of `hash_dir`, thanks for the very detailed description! 

how much difference are we talking about here though? just curious

lgtm!

mj-xmr:
> how much difference are we talking about here though? just curious

It either works or it doesn't :)
If `hash_dir` is kept `true` (as per default), then if ccache sees a different directory where the source is placed, like in my case: monero **-mj**, it will have to rebuild the from scratch cache.

cirocosta (approved):

selsta:
Is there a case where `hash_dir` makes a difference for CI?

mj-xmr:
> Is there a case where `hash_dir` makes a difference for CI?

It's a good question. It is possible to reuse caches of the upstream project. If yours has a different name, it can't reuse them without `hash_dir`.

selsta:
AFAIK our caches aren't shared in the first place.

mj-xmr:
I'll prepare an experiment for this later.

mj-xmr:
> I'll prepare an experiment for this later.

Currently I'm running another experiment on my fork, to prove another point. This creates a bunch of own caches, fresh enough to ignore the base project's caches, which is a shame because the base project has just refreshed its caches. @cirocosta @selsta : If you haven't started any builds of your fork's master lately, now it would be a good chance to do this, and see the commit hashes of the restored caches. A draft PR would do. I can't help it currently.

> AFAIK our caches aren't shared in the first place.

I'm not 100% sure if this piece is convincing enough.

https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

## Restrictions for accessing a cache

A workflow can access and restore a cache created in the current branch, **the base branch (including base branches of forked repositories), or the default branch (usually main). For example, a cache created on the default branch would be accessible from any pull request.** Also, if the branch feature-b has the base branch feature-a, a workflow triggered on feature-b would have access to caches created in the default branch (main), feature-a, and feature-b.

Access restrictions provide cache isolation and security by creating a logical boundary between different workflows and branches. For example, a cache created for the branch feature-a (with the base main) would not be accessible to a pull request for the branch feature-b (with the base main).

mj-xmr:
> I believe it's used indeed over reruns only atm:
> 
> https://github.com/monero-project/monero/blob/f862d7b4ddb57e483d67b8b94f3c433435571c7b/.github/workflows/build.yml#L135-L140
> 
> *reruns because the key has the commit sha: key: `ccache-ubuntu-build-ubuntu-latest-${{ github.sha }}`

This is how you do it. The write key has to be unique enough, while the restore keys not. See here:


## [Example using the cache action](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action)

This example creates a new cache when the packages in package-lock.json file change, or when the runner's operating system changes. The cache key uses contexts and expressions to generate a key that includes the runner's operating system and a SHA-256 hash of the package-lock.json file.

mj-xmr:
@selsta Here is a proof, that we reuse the caches from the master of the upstream project. Notice the repeated hash of the stored and then restored cache - `a39b1d56c8835798e2f5e3cc43c33b5f2d8e13da`

[This is the last `depends` build](https://github.com/monero-project/monero/runs/3560265492?check_suite_focus=true) of the latest set of merges on 2021-09-09.
image

```
Post job cleanup.
/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /home/runner/work/monero/monero --files-from manifest.txt
Cache Size: ~101 MB (105653470 B)
Cache saved successfully
Cache saved with key: ccache-riscv64-linux-gnu-a39b1d56c8835798e2f5e3cc43c33b5f2d8e13da
```

[While this is the first build of a PR, pushed just 2 days later](https://github.com/monero-project/monero/runs/3575081554?check_suite_focus=true):
image

```
Run actions/cache@v2
Received 0 of 105653470 (0.0%), 0.0 MBs/sec
Received 88080384 of 105653470 (83.4%), 42.0 MBs/sec
Received 105653470 of 105653470 (100.0%), 42.7 MBs/sec
Cache Size: ~101 MB (105653470 B)
/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/807e767e-92af-4b9e-87d8-7d9b2f9ea796/cache.tzst -P -C /home/runner/work/monero/monero
Cache restored successfully
Cache restored from key: ccache-riscv64-linux-gnu-a39b1d56c8835798e2f5e3cc43c33b5f2d8e13da
```

selsta:
Both of the linked CI builds aren't run in a forked repository.

selsta:
This is the CI from the forked repository and it did not access any upstream caches: https://github.com/SChernykh/monero/runs/3575081407

mj-xmr:
You're right. I've just been checking it myself.
Similar examples would be your:
https://github.com/selsta/monero/runs/3576103049?check_suite_focus=true
and mm's:
https://github.com/moneromooo-monero/bitmonero/runs/3562240401?check_suite_focus=true#logs

Moreover, mm's run on the upstream doesn't contain the special `bitmonero` project name anymore, unlike the fork. So for these 2 reasons combined the `hashdir` setting doesn't make sense.

selsta (approved):

PR #7910, opened 2021-08-31
Add simple script for running gitian-build in docker

hyc:
This shell script creates a docker image for running the gitian-build.py
script, to avoid version incompatibilities.

Only docker.io and apt-cacher-ng are required to be installed on the
host. The MacOSX SDK should also be present in this directory if Mac
builds are desired. A container named "gitrun" is used to launch
the builds, and all output will reside in this container as well.

Run as e.g.
`  OPT="-j 8" ./docker.sh v0.17.2.3`

PR #7934, opened 2021-09-11
Dockerfile: use depends

### LINKED ISSUE #7938: Docker Container uses a end of life version of Ubuntu
Hi,

just a small heads up: the Monero container uses Ubuntu 16.04 which is EOL for non enterprise customer since **April 2021**.

https://wiki.ubuntu.com/Releases

https://github.com/monero-project/monero/blob/a39b1d56c8835798e2f5e3cc43c33b5f2d8e13da/Dockerfile#L182

----

I don't know how the release workflow works exactly but I would recommend to not compile monero again within the Dockerfile since its already built by the CI for ubuntu. This would simplify the Dockerfile itself, reduces duplicated environment configuration (Gitlab CI & Dockerfile) and hashes for the ubuntu compiled version would match with the hash inside the container.

https://github.com/monero-project/monero/blob/a39b1d56c8835798e2f5e3cc43c33b5f2d8e13da/.github/workflows/build.yml#L69-L95

----

Additionally it would be awesome if the Container could be published within the GitHub releases via the [Github Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry).

selsta:
Resolves #7938

Should be easier to maintain in regards to dependencies. Untested.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,192 +1,45 @@
 # Multistage docker build, requires docker 17.05
 
 # builder stage
-FROM ubuntu:16.04 as builder
+FROM ubuntu:20.04 as builder
 
 RUN set -ex && \
     apt-get update && \
-    apt-get --no-install-recommends --yes install \
-        ca-certificates \
-        cmake \
-        g++ \
-        make \
-        pkg-config \
-        graphviz \
-        doxygen \
-        git \
-        curl \
-        libtool-bin \
-        autoconf \
+    DEBIAN_FRONTEND=noninteractive apt-get --no-install-recommends --yes install \
         automake \
-        bzip2 \
-        xsltproc \
-        gperf \
-        unzip
-
-WORKDIR /usr/local
-
-ENV CFLAGS='-fPIC'
-ENV CXXFLAGS='-fPIC'
-
-#Cmake
-ARG CMAKE_VERSION=3.14.6
-ARG CMAKE_VERSION_DOT=v3.14
-ARG CMAKE_HASH=4e8ea11cabe459308671b476469eace1622e770317a15951d7b55a82ccaaccb9
-RUN set -ex \
-    && curl -s -O https://cmake.org/files/${CMAKE_VERSION_DOT}/cmake-${CMAKE_VERSION}.tar.gz \
-    && echo "${CMAKE_HASH}  cmake-${CMAKE_VERSION}.tar.gz" | sha256sum -c \
-    && tar -xzf cmake-${CMAKE_VERSION}.tar.gz \
-    && cd cmake-${CMAKE_VERSION} \
-    && ./configure \
-    && make \
-    && make install
-
-## Boost
-ARG BOOST_VERSION=1_70_0
-ARG BOOST_VERSION_DOT=1.70.0
-ARG BOOST_HASH=430ae8354789de4fd19ee52f3b1f739e1fba576f0aded0897c3c2bc00fb38778
-RUN set -ex \
-    && curl -s -L -o  boost_${BOOST_VERSION}.tar.bz2 https://downloads.sourceforge.net/project/boost/boost/${BOOST_VERSION_DOT}/boost_${BOOST_VERSION}.tar.bz2 \
-    && echo "${BOOST_HASH}  boost_${BOOST_VERSION}.tar.bz2" | sha256sum -c \
-    && tar -xvf boost_${BOOST_VERSION}.tar.bz2 \
-    && cd boost_${BOOST_VERSION} \
-    && ./bootstrap.sh \
-    && ./b2 --build-type=minimal link=static runtime-link=static --with-chrono --with-date_time --with-filesystem --with-program_options --with-regex --with-serialization --with-system --with-thread --with-locale threading=multi threadapi=pthread cflags="$CFLAGS" cxxflags="$CXXFLAGS" stage
-ENV BOOST_ROOT /usr/local/boost_${BOOST_VERSION}
-
-# OpenSSL
-ARG OPENSSL_VERSION=1.1.1i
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Multistage docker build, requires docker 17.05
 
 # builder stage
-FROM ubuntu:16.04 as builder
+FROM ubuntu:20.04 as builder
 
 RUN set -ex && \
     apt-get update && \
```
cirocosta, line 7: Screen Shot 2021-09-11 at 6 25 35 PM

might be worth setting `DEBIAN_FRONTEND=noninteractive` so that it doesn't block and expect user interaction

selsta: updated

selsta: Or should I set it for apt update?

cirocosta: lgtm, `install` is the one to go 👍

Dockerfile
```diff
@@ -23,170 +23,27 @@ RUN set -ex && \
         gperf \
         unzip
```
cirocosta, line 24: looks like `patch` is missing (either from `patch` package or `build-essential`) 

I think we could reproduce go with the same set of packages as https://github.com/monero-project/monero/blob/a39b1d56c8835798e2f5e3cc43c33b5f2d8e13da/.github/workflows/depends.yml#L77 - wdyt?

cirocosta: (although `ccache` would not really be useful unless we use https://github.com/moby/buildkit/blob/e07f3883e343ab311ca96be8ca9be3619ed27db7/frontend/dockerfile/docs/syntax.md#run---mounttypecache)

Dockerfile
```diff
@@ -23,170 +23,27 @@ RUN set -ex && \
...
 
-ENV USE_SINGLE_BUILDDIR=1
 ARG NPROC
 RUN set -ex && \
     git submodule init && git submodule update && \
     rm -rf build && \
     if [ -z "$NPROC" ] ; \
-    then make -j$(nproc) release-static ; \
-    else make -j$NPROC release-static ; \
+    then make -j$(nproc) depends target=x86_64-linux-gnu ; \
```
cirocosta, line 34: I _think_ we can actually do multiplatform builds given how easy depends makes it so 🤔 

I've done some work on that before for another project (https://github.com/cirocosta/concourse-arm/blob/4081f81c4f51b7d477a9ddbc66ab9f7a543d9904/Dockerfile#L1-L4), I think we might (if we want) be able to do something similar (might be questionable whether that's desired or not though - could reserve that discussion to a different pr/issue)

Dockerfile
selsta: done

Dockerfile
selsta: Would keep this for a different PR.

Dockerfile
```diff
@@ -1,192 +1,43 @@
...
-        doxygen \
-        git \
+        bsdmainutils \
         curl \
-        libtool-bin \
-        autoconf \
-        automake \
-        bzip2 \
-        xsltproc \
-        gperf \
```
cirocosta, line 23: whoops, `gperf` seems to be required actually (not sure why it works fine in github actions though)

selsta: Seems to be installed here: https://github.com/monero-project/monero/blob/master/.github/workflows/depends.yml#L35

Don't know if the others are also required.

cirocosta: oh true

sounds like no - I just finished a build with the addition of just `gperf` 👍

cirocosta (approved):

Dockerfile
selsta: No build-essentials / patch?

Malinero (approved):

hyc:
Would've saved yourself a lot of trouble to just use the setup I had in here https://github.com/monero-project/monero/pull/7910

selsta:
> Would've saved yourself a lot of trouble to just use the setup I had in here #7910

This here isn't for reproducible builds. I'm fine with also deleting the docker file here.

mj-xmr:
> Should be easier to maintain in regards to dependencies. Untested.

I will get it tested until this Sunday.

mj-xmr:
> This here isn't for reproducible builds. AFAIK it got added as an easy way to build and run monero with Docker. I'm also fine with deleting this whole file.

I'm against deleting it. It gives more options in a reproducible and isolated way.

mj-xmr:
It builds and runs fine under `Ubuntu 20.04.3 LTS`

```
Removing intermediate container baec40e8ff7a
 ---> 21075e8db9b3
Step 16/16 : ENTRYPOINT ["monerod", "--p2p-bind-ip=0.0.0.0", "--p2p-bind-port=18080", "--rpc-bind-ip=0.0.0.0", "--rpc-bind-port=18081", "--non-interactive", "--confirm-external-bind"]
 ---> Running in 7b7e8ed2fd64
Removing intermediate container 7b7e8ed2fd64
 ---> 9571f294849c
Successfully built 9571f294849c
Successfully tagged monero-selsta:latest
mjxmr@zot:~/store/devel/github/monero/monero-selsta$ sudo docker run monero-selsta
[sudo] password for mjxmr: 
2021-09-17 15:10:49.361	I Monero 'Oxygen Orion' (v0.17.0.0-a95e6e0d4)
2021-09-17 15:10:49.361	I Initializing cryptonote protocol...
2021-09-17 15:10:49.361	I Cryptonote protocol initialized OK
2021-09-17 15:10:49.364	I Initializing core...
2021-09-17 15:10:49.365	I Loading blockchain from folder /home/monero/.bitmonero/lmdb ...
2021-09-17 15:10:49.459	I Loading checkpoints
2021-09-17 15:10:49.459	I Core initialized OK
(...)
```

mj-xmr (approved):

PR #7939, opened 2021-09-14
Add function flush_txes_from_pool() to enable thread safe method of flushing the entire pool

j-berman:
### Overview

Add a thread safe way to flush all transactions from the pool.

### Details

Prior to this PR, the only function available to flush the pool was `blockchain.cpp:flush_txes_from_poolI(txids)`. The caller would need to provide `txids` of all hashes to remove. If the caller wanted to flush the entire pool, they would first need to read the db for all pool txes, and then pass the `txids` of those txes into the `flush_txes_from_pool`. With that approach, txes could enter/exit the pool *after* the caller retrieves the list of all `txids`, but *before* flushing.

**Solution**

- Added a new function `flush_txes_from_pool()` with no params that flushes the entire pool
- The function grabs the tx pool lock and reads all txes from the pool, then calls the updated `flush_txes_from_pool(txids)` with the pool's `txids`. Since it's inside the critical region, no tx's should enter/exit the pool when the caller flushes the entire pool

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -3995,6 +3995,17 @@ bool Blockchain::flush_txes_from_pool(const std::vector<crypto::hash> &txids)
   return res;
 }
 //------------------------------------------------------------------
+bool Blockchain::flush_txes_from_pool()
+{
+  CRITICAL_REGION_LOCAL(m_tx_pool);
+
+  std::vector<crypto::hash> txids;
+  bool include_sensitive = true;
+  m_tx_pool.get_transaction_hashes(txids, include_sensitive);
+
+  return flush_txes_from_pool(txids);
+}
+//------------------------------------------------------------------
 //      Needs to validate the block and acquire each transaction from the
 //      transaction mem_pool, then pass the block and transactions to
 //      m_db->add_block()
--- a/src/cryptonote_core/blockchain.h
+++ b/src/cryptonote_core/blockchain.h
@@ -877,6 +877,13 @@ namespace cryptonote
     bool flush_txes_from_pool(const std::vector<crypto::hash> &txids);
 
     /**
+     * @brief remove all transactions from the transaction pool (if present)
+     *
+     * @return false if any removals fail, otherwise true
+     */
+    bool flush_txes_from_pool();
+
+    /**
      * @brief return a histogram of outputs on the blockchain
      *
      * @param amounts optional set of amounts to lookup
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -2493,30 +2493,16 @@ namespace cryptonote
   {
     RPC_TRACKER(flush_txpool);
 
-    bool failed = false;
+    bool failed_to_parse = false;
     std::vector<crypto::hash> txids;
-    if (req.txids.empty())
-    {
-      std::vector<transaction> pool_txs;
-      bool r = m_core.get_pool_transactions(pool_txs, true);
-      if (!r)
-      {
-        res.status = "Failed to get txpool contents";
-        return true;
-      }
-      for (const auto &tx: pool_txs)
-      {
-        txids.push_back(cryptonote::get_transaction_hash(tx));
-      }
-    }
-    else
+    if (!req.txids.empty())
     {
       for (const auto &str: req.txids)
       {
         crypto::hash txid;
         if(!epee::string_tools::hex_to_pod(str, txid))
         {
-          failed = true;
+          failed_to_parse = true;
         }
         else
         {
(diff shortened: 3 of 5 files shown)


src/cryptonote_core/blockchain.cpp
```diff
@@ -3973,20 +3973,46 @@ void Blockchain::return_tx_to_pool(std::vector<std::pair<transaction, blobdata>>
   }
 }
 //------------------------------------------------------------------
-bool Blockchain::flush_txes_from_pool(const std::vector<crypto::hash> &txids)
+bool Blockchain::flush_txes_from_pool(const std::vector<crypto::hash> &txids, bool flush_entire_pool)
```
iamamyth, line 3976: I would opt for acquiring this lock in flush_txes_from_pool (the mutex should be re-entrant), then passing the relevant data here and keeping the function as-is, otherwise:

1. You substantially increase the complexity of this function (suitable test cases would need to re-exercise the per-tx flushing behavior for the flush entire pool case, vs merely asserting the right txes have been passed to this function).

2. You end up with a bizarre calling convention, whereby flushing the entire pool requires creating an unneeded empty vector.

j-berman: Agree with this. Didn't (and should have) realize(d) the mutex is reentrant - much cleaner/simpler this way.

See latest - thoughts on the extra boolean so no need to reread the db for every tx in the flush entire pool case?

iamamyth: Regarding the boolean, the `have_tx` check didn't make a whole lot of sense to me, so I checked the code, and it seems the reason for that test is that `take_tx` can fail if the tx data isn't parseable. So, assuming parseable transactions, you can gain the same efficiency by simply swapping the order of the conditional (i.e. `take_tx && !have_tx`). Also, I note there's a `get_transaction_hashes` method, which you can use in place of the `get_transaction_hash` loop. Lastly, I'll note that in both the flush all and flush selected cases, it seems odd for any parsing to happen at all. I suspect that, in a future PR, it may be possible to introduce a `drop_tx` method to the pool, and gain some efficiency by using that method, rather than `take_tx`.

j-berman: > Also, I note there's a get_transaction_hashes method, which you can use in place of the get_transaction_hash loop.

True, will include this.

Appreciate this discussion, it's a helpful exercise as I ramp up contributing and working across the code base :)

> So, assuming parseable transactions, you can gain the same efficiency by simply swapping the order of the conditional (i.e. take_tx && !have_tx)

If this were safe, I figure it would be safe to simply remove `have_tx` entirely, since `have_tx` doesn't look like it can be false if `take_tx` is true. Both functions read the same record from the db.

By placing `have_tx` in front and `!take_tx` behind, it seems to separate the 2 functions such that any unexpected failure in `take_tx` is picked up on and indicated to the user via a failed response (and not just a parsing error by the looks of it, but also a thrown exception inside `remove_transaction_keyimages` for example, which also seems like it should never happen). While placing `have_tx` in front allows the function to ignore any non-existent tx's from the pool with no indication to the user.

It seems `have_tx` is an additional unnecessary read itself that could be done away with, but at this point it feels I'm moving too strongly into the territory of over-optimizing.

> Lastly, I'll note that in both the flush all and flush selected cases, it seems odd for any parsing to happen at all

The parsing appears to be necessary to update values held in memory (`m_txpool_weight` and `m_spent_key_images` in `remove_transaction_keyimages`)

j-berman: Placing `!take_tx` in front of `have_tx` introduces the red `MERROR` message `E Failed to find tx_meta in txpool` in the daemon when the tx doesn't exist, which also seems undesirable

j-berman: I'm just going to get rid of the extra boolean and isolate this PR to enable thread safety when flushing the entire pool. Can handle perf gains in a future PR.

iamamyth: > Placing !take_tx in front of have_tx introduces the red MERROR message E Failed to find tx_meta in txpool in the daemon when the tx doesn't exist, which also seems undesirable.

The logging in `take_tx` tends to be of the opinion that the tx to take must exist, which is a little unfortunate, considering that the rest of the method looks quite similar to a potential `drop_tx` implementation. A future PR might want to make the "log non-existing" optional via a boolean. 

> The parsing appears to be necessary to update values held in memory (m_txpool_weight and m_spent_key_images in remove_transaction_keyimages)

Forgot to scroll down. A `drop_tx` function would have some potential benefits (don't log missing txes, don't return throw-away values), but probably not enough to overcome the duplication.
 
> I'm just going to get rid of the extra boolean and isolate this PR to enable thread safety when flushing the entire pool. Can handle perf gains in a future PR.

Makes sense to me.

PR #7943, opened 2021-09-15
rpc: don't set error code as status string

selsta:
Seen on Loki repo.

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -2849,7 +2849,7 @@ namespace cryptonote
     {
       if (req.limit_down != -1)
       {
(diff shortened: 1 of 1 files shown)


src/rpc/core_rpc_server.cpp
```diff
@@ -2849,7 +2849,7 @@ namespace cryptonote
     {
       if (req.limit_down != -1)
       {
-        res.status = CORE_RPC_ERROR_CODE_WRONG_PARAM;
+        res.status = "Invalid parameter";
```
mj-xmr, line 2852: The `status` variable is a string, so makes sense.

mj-xmr (approved):

PR #7945, opened 2021-09-15
cmake: set required C/C++ standard to 11

selsta:
Now that we have CMake 3.5 as minimum we can properly set the required C/C++ standard in a cross platform way.

Seen on Loki repo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,14 @@ endif()
 
 enable_language(C ASM)
 
+# Require C11/C++11 and disable extensions for all targets
+set(CMAKE_C_STANDARD 11)
+set(CMAKE_C_STANDARD_REQUIRED ON)
+set(CMAKE_C_EXTENSIONS OFF)
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
 function (die msg)
   if (NOT WIN32)
     string(ASCII 27 Esc)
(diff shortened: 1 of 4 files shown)


mj-xmr (commented):
Have you deliberately omitted the occurrences in 796-803?

https://github.com/monero-project/monero/pull/7945/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL796-R809

```
  try_compile(STATIC_ASSERT_RES "${CMAKE_CURRENT_BINARY_DIR}/static-assert" "${CMAKE_CURRENT_SOURCE_DIR}/cmake/test-static-assert.c" COMPILE_DEFINITIONS "-std=c11")
  if(STATIC_ASSERT_RES)
    set(STATIC_ASSERT_FLAG "")
  else()
    set(STATIC_ASSERT_FLAG "-Dstatic_assert=_Static_assert")
  endif()
  try_compile(STATIC_ASSERT_CPP_RES "${CMAKE_CURRENT_BINARY_DIR}/static-assert" "${CMAKE_CURRENT_SOURCE_DIR}/cmake/test-static-assert.cpp" COMPILE_DEFINITIONS "-std=c++11")
```

mj-xmr (commented):
There is also a mention in `epee/demo`, but I guess we don't care about it:

https://github.com/monero-project/monero/blob/master/contrib/epee/demo/CMakeLists.txt#L13

`SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wno-reorder -D_GNU_SOURCE")`

selsta:
> Have you deliberately omitted the occurrences in 796-803?

Initially yes because the feature was part of CMake 3.8 (https://cmake.org/cmake/help/latest/command/try_compile.html#id3) but I found a different way to solve it. Updated.

> There is also a mention in epee/demo, but I guess we don't care about it:

Yes, don't really care about it. IMO it should be deleted.

mj-xmr (approved):

PR #7958, opened 2021-09-19
LMDB: fix deadlock in resized detection

### LINKED ISSUE #7957: LMDB possible deadlock with low probability in the wrapper in lmdb_txn_begin()
Hi guys:

I think I caught a possible deadlock.

In [`lmdb_txn_begin()`](https://github.com/monero-project/monero/blob/ad9956d9873ad8281fd88daf8b9202c74db3f796/src/blockchain_db/lmdb/db_lmdb.cpp#L495) function, the `mdb_txn_begin()` is called, and then in rare cases, the result can be `MDB_MAP_RESIZED`, which will require to call a resize function, `lmdb_resized()`. However, notice that `lmdb_resized()` calls `mdb_txn_safe::wait_no_active_txns();`. This function waits for `num_active_txns` to be zero before it returns... but that will never happen! Because, looking at an example of when `lmdb_txn_begin()` is called, you construct `mdb_txn_safe()` right before it, like you do [here](https://github.com/monero-project/monero/blob/ad9956d9873ad8281fd88daf8b9202c74db3f796/src/blockchain_db/lmdb/db_lmdb.cpp#L3976), which will increment the count of `num_active_txns`.

So, in summary, [here](https://github.com/monero-project/monero/blob/ad9956d9873ad8281fd88daf8b9202c74db3f796/src/blockchain_db/lmdb/db_lmdb.cpp#L3976), you increase the count of `num_active_txns` by one, then in the next line (if `MDB_MAP_RESIZED`) you wait for it to go back to zero *forever* [here](https://github.com/monero-project/monero/blob/ad9956d9873ad8281fd88daf8b9202c74db3f796/src/blockchain_db/lmdb/db_lmdb.cpp#L499), because of [this](https://github.com/monero-project/monero/blob/ad9956d9873ad8281fd88daf8b9202c74db3f796/src/blockchain_db/lmdb/db_lmdb.cpp#L481).

Maybe I'm wrong... I don't know. It's 10:30 PM here, and I'm just too tired.

Regards.

hyc:
Fix #7957 - account for the current thread's active txn, if any.


src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -490,13 +497,15 @@ void lmdb_resized(MDB_env *env)
   MGINFO("LMDB Mapsize increased." << "  Old: " << old / (1024 * 1024) << "MiB" << ", New: " << new_mapsize / (1024 * 1024) << "MiB");
 
   mdb_txn_safe::allow_new_txns();
+  if (isactive)
+    mdb_txn_safe::increment_txns(1);
```
TheQuantumPhysicist, line 501: Perhaps it's a better idea to put this back right after `wait_no_active_txns()`. Because, if that exception is thrown, and it's caught, and then the `mdb_txn_safe` object is destroyed, it'll lead to `num_active_txns` of `-1`.

Alternatively, something like `BOOST_SCOPE_EXIT` can be used to ensure that RAII will make the value go back up before the function ends.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -478,6 +483,8 @@ void lmdb_resized(MDB_env *env)
   mdb_env_info(env, &mei);
   uint64_t old = mei.me_mapsize;
 
+  if (isactive)
```
TheQuantumPhysicist, line 486: Just for the sake of nitpicking, instead of an ugly if, something like this can be done:

`mdb_txn_safe::increment_txns(-isactive);`
`mdb_txn_safe::increment_txns(isactive);`

Or to ensure that no bad value is passed, since bools and ints play nice:

`mdb_txn_safe::increment_txns(-!!isactive);`
`mdb_txn_safe::increment_txns(!!isactive);`

If you don't like any of this, a ternary operator still looks better since the expression is very small:

`mdb_txn_safe::increment_txns(isactive ? -1 : 0);`
`mdb_txn_safe::increment_txns(isactive ? 1 : 0);`

src/blockchain_db/lmdb/db_lmdb.h
```diff
@@ -154,6 +154,7 @@ struct mdb_txn_safe
   static void prevent_new_txns();
   static void wait_no_active_txns();
   static void allow_new_txns();
+  static void increment_txns(const int);
```
TheQuantumPhysicist, line 157: The const in superfluous

src/blockchain_db/lmdb/db_lmdb.cpp
hyc: The point of the `if` is to avoid the function call at all, when not needed.

src/blockchain_db/lmdb/db_lmdb.h
hyc: ok

selsta (approved):

TheQuantumPhysicist (approved):

PR #7969, opened 2021-09-23
serialization: fix unportable code

selsta:

### DIFF
--- a/src/cryptonote_protocol/cryptonote_protocol_defs.h
+++ b/src/cryptonote_protocol/cryptonote_protocol_defs.h
@@ -309,6 +309,10 @@ namespace cryptonote
         else
           KV_SERIALIZE_OPT(cumulative_difficulty_top64, (uint64_t)0)
         KV_SERIALIZE_CONTAINER_POD_AS_BLOB(m_block_ids)
+        // backward incompatible portable encoding
+        KV_SERIALIZE_N(m_block_weights, "m_block_weights_")
+        // inefficient and simple backward compatibility
+        // TODO: can be removed after v15+ hf
         KV_SERIALIZE_CONTAINER_POD_AS_BLOB(m_block_weights)
         KV_SERIALIZE(first_block)
       END_KV_SERIALIZE_MAP()
@@ -352,6 +356,10 @@ namespace cryptonote
       BEGIN_KV_SERIALIZE_MAP()
         KV_SERIALIZE_VAL_POD_AS_BLOB(block_hash)
         KV_SERIALIZE(current_blockchain_height)
+        // backward incompatible portable encoding
+        KV_SERIALIZE_N(missing_tx_indices, "missing_tx_indices_")
+        // inefficient and simple backward compatibility
+        // TODO: can be removed after v15+ hf
         KV_SERIALIZE_CONTAINER_POD_AS_BLOB(missing_tx_indices)
       END_KV_SERIALIZE_MAP()
     };
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -3171,7 +3171,7 @@ namespace cryptonote
           return false;
         }
 
-        res.distributions.push_back({std::move(*data), amount, "", req.binary, req.compress});
+        res.distributions.push_back({std::move(*data), amount, req.binary, true});
       }
     }
     catch (const std::exception &e)
@@ -3218,7 +3218,7 @@ namespace cryptonote
           return true;
         }
 
-        res.distributions.push_back({std::move(*data), amount, "", req.binary, req.compress});
+        res.distributions.push_back({std::move(*data), amount, req.binary, true});
       }
     }
     catch (const std::exception &e)
--- a/src/rpc/core_rpc_server_commands_defs.h
+++ b/src/rpc/core_rpc_server_commands_defs.h
@@ -2370,7 +2370,6 @@ namespace cryptonote
     {
       rpc::output_distribution_data data;
       uint64_t amount;
-      std::string compressed_data;
       bool binary;
       bool compress;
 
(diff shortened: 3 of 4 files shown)


src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -2379,28 +2379,15 @@ namespace cryptonote
         KV_SERIALIZE_N(data.start_height, "start_height")
         KV_SERIALIZE(binary)
         KV_SERIALIZE(compress)
-        if (this_ref.binary)
+        if (this_ref.binary && is_store && this_ref.compress)
```
iamamyth, line 2382: Are the following two macro expansions equivalent in effect? If not, this change appears to break backwards compatibility:

> KV_SERIALIZE_CONTAINER_POD_AS_BLOB_N(data.distribution, "distribution")

> KV_SERIALIZE_N(data.distribution, "distribution")


Also, if the proposed simplification is, in fact, backwards compatible, it can be expressed as follows:

```
if (this_ref.binary && this_ref.compress)
{
  if (is_store) 
  {
    const_cast<std::string&>(this_ref.compressed_data) = compress_integer_array(this_ref.data.distribution);
    KV_SERIALIZE(compressed_data)  
  }
  else
  {
    KV_SERIALIZE(compressed_data)
    const_cast<std::vector<uint64_t>&>(this_ref.data.distribution) = decompress_integer_array<uint64_t>(this_ref.compressed_data);
  }
}
else
{
  KV_SERIALIZE_N(data.distribution, "distribution")
}
```

The proposed change would have the effect of clarifying the branching strategy: binary + compressed follow one path (with load and store being distinct), everything else follows a different path.

selsta: So `KV_SERIALIZE_CONTAINER_POD_AS_BLOB_N(data.distribution, "distribution")` isn't portable between little / big endian systems. We work around it by forcing all client requests to set "compress" to true so the portable method gets called instead.

iamamyth: You probably shouldn't consider "endianness of response depends on endianness of host" a behavior to preserve, as the "remote" in RPC implies the behavior should be uniform regardless of the host. If you just forced everything to little endian, very few clients would be affected and there'd be a clear specification for future callers.

selsta: > You probably shouldn't consider "endianness of response depends on endianness of host" a behavior to preserve, as the "remote" in RPC implies the behavior should be uniform regardless of the host.

`portable_storage` encodes everything as little endian for the network transfer and then encodes it with / without bytes swap depending on receiver machine.

> If you just forced everything to little endian, very few clients would be affected and there'd be a clear specification for future callers.

This code should support all cases now. Could you rephrase what exactly you are suggesting and why?

iamamyth: I meant that, based on the comments and current code, it seems you intend to preserve the old key and behavior until some future hardfork, while introducing a new key with portable semantics, whereas the path of least resistance would perhaps involve fixing the prior serialization.

src/cryptonote_protocol/cryptonote_protocol_defs.h
```diff
@@ -309,6 +309,10 @@ namespace cryptonote
         else
           KV_SERIALIZE_OPT(cumulative_difficulty_top64, (uint64_t)0)
         KV_SERIALIZE_CONTAINER_POD_AS_BLOB(m_block_ids)
+        // backward incompatible portable encoding
+        KV_SERIALIZE_N(m_block_weights, "m_block_weights_")
```
vtnerd, line 313: Should this line be placed after line 316? The de-serialization code should try each field in-order listed here, so the value of `m_block_weights` will take precedence over (replace) `m_block_weights_`.

src/cryptonote_protocol/cryptonote_protocol_defs.h
```diff
@@ -352,6 +356,10 @@ namespace cryptonote
       BEGIN_KV_SERIALIZE_MAP()
         KV_SERIALIZE_VAL_POD_AS_BLOB(block_hash)
         KV_SERIALIZE(current_blockchain_height)
+        // backward incompatible portable encoding
+        KV_SERIALIZE_N(missing_tx_indices, "missing_tx_indices_")
```
vtnerd, line 360: Same here.

src/rpc/core_rpc_server.cpp
```diff
@@ -3171,7 +3171,7 @@ namespace cryptonote
           return false;
         }
 
-        res.distributions.push_back({std::move(*data), amount, "", req.binary, req.compress});
+        res.distributions.push_back({std::move(*data), amount, req.binary, true});
```
vtnerd, line 3174: I think a check for `req.binary && !req.compress` should be done, and an error returned in that case. This would only trigger on a (extremely unlikely) custom client.

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -2379,31 +2378,24 @@ namespace cryptonote
         KV_SERIALIZE_N(data.start_height, "start_height")
         KV_SERIALIZE(binary)
         KV_SERIALIZE(compress)
-        if (this_ref.binary)
+        if (this_ref.binary && this_ref.compress)
```
vtnerd, line 2381: Should the `compress` variable just be removed?

src/cryptonote_protocol/cryptonote_protocol_defs.h
selsta: It's intentional in this order, otherwise this code would be required due to how portable storage works internally:

```
         {
          if (is_store)
            KV_SERIALIZE_CONTAINER_POD_AS_BLOB(m_block_weights)
          else {
            std::vector<std::uint64_t> m_block_weights;
            epee::serialization::selector<is_store>::serialize_stl_container_pod_val_as_blob(
              m_block_weights, stg, hparent_section, "m_block_weights");
            if (this_ref.m_block_weights.empty() and not m_block_weights.empty())
              this_ref.m_block_weights = std::move(m_block_weights);
          }
        }
        KV_SERIALIZE_N(m_block_weights, "m_block_weights_")
```

src/rpc/core_rpc_server_commands_defs.h
selsta: Just from the `if`? I could do that.

src/rpc/core_rpc_server.cpp
selsta: Wouldn't wallet clients before https://github.com/monero-project/monero/pull/7969/commits/267dc5a08ba4e362cb22f4b0aabde92eea92339a#diff-04cf14f64d2023c7f9cd7bd8e51dcb32ed400443c6a67535cb0105cfa2b62c3cR8283 trigger this?

selsta:
This should be addressed by #8867 in the future.

PR #7970, opened 2021-09-24
[DO NOT MERGE] Some LOG_ERROR messages disappear when headers are included

jeffro256:
Runing the command `monerod --rpc-access-control-origins '*'` on master shows an error message (this is fine):

    E rpc-access-control-origins requires RPC server password --rpc-login cannot be empty

However, running the same command with these changes does not show that error. This behavior is consistent no matter what other arguments are supplied, network type, platform, etc. I will note that the message is logged in bitmonero.log, just not on the terminal. It is not clear to me why this is happening, but it appears to be a bug. Finding out why this is happening might benefit future development.

### DIFF
--- a/src/daemon/command_line_args.h
+++ b/src/daemon/command_line_args.h
@@ -31,6 +31,7 @@
 
 #include "common/command_line.h"
 #include "cryptonote_config.h"
+#include "cryptonote_core/cryptonote_core.h"
 #include "daemonizer/daemonizer.h"
 
 namespace daemon_args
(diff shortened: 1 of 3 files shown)


selsta:
It still gets printed with `--log-level 1`, though not sure why changing the includes causes this.

jeffro256:
Do you think this has anything to do with the recent easylogging++ PR?

selsta:
I'll close this as this doesn't really fit into pull requests.

PR #7971, opened 2021-09-25
cmake: set cpp14 standard

selsta:
It's 2021, we can easily switch to C++14 without any troubles.

I also looked into C++17 but it would require us to drop macOS 10.10 - 10.13 support, which wouldn't be ideal, so C++14 for now.


mj-xmr (approved):
It's OK with me, especially when we can drop some of Boost. Just don't be surprised, that it will increase the compilation time by a little bit, since every next standard is heavier on the compiler.

trasherdk:
:sweat_smile: Okay, I'm still good on `Slackware 14.2`

```sh
# gcc -v --help 2> /dev/null | sed -n '/^ *-std=\([^<][^ ]\+\).*/ {s//\1/p}' | grep -P 'c\+\+'
c++03
c++0x
c++11
c++14
c++17
c++1y
c++1z
c++98
```

mj-xmr:
There are still so.e instances of 2011 standard. Are they being left by accident?

selsta:
Easylogging++ is a separate project and I don't care about epee tests / demo. Anything else?

vtnerd:
I'm mostly excited about this change, but I thought it worthwhile pointing out some things -

Ubuntu 18.04 is still supported, and the default compiler is GCC 4.7. There is a package for GCC 7 however. This might make the depends script unusable on that older Ubuntu - however running cmake with the fully qualified path of the compiler should still work.

Also, the first version that came with C++14 on by default was GCC 6. There were some stackoverflow that libstdc++ may not have been compliant until that version. Unfortunately, there is no page that indicates what release features were added so its difficult to know what to avoid.  GCC compiler features (as opposed to library) are all marked as supported by GCC 5.

selsta:
According to https://askubuntu.com/a/1045071 Ubuntu 18.04 ships with `GCC 7.4.0` by default. This should not be a problem at all.

vtnerd:
Crap, I read that as `4.7.4` not `4:7.4`. After some searching, that `4:` refers to the "epoch" (inherited from Debian). If omitted it defaults to 0. This allows Ubuntu to "downgrade" as a default if desired.

I looked into the GCC 5 vs 6 thing, and in an [obscure case there is a difference](https://stackoverflow.com/questions/47282939/since-which-version-of-gcc-is-c14-supported). I'm not sure its worth bumping the min for, considering we haven't hit that edge case.

mj-xmr:
> Easylogging++ is a separate project so I prefer to keep it as is and I don't care about epee tests / demo. Anything else that I overlooked?

https://github.com/monero-project/monero/pull/7971/files#R798
```CMake
try_compile(STATIC_ASSERT_CPP_RES "${CMAKE_CURRENT_BINARY_DIR}/static-assert" "${CMAKE_CURRENT_SOURCE_DIR}/cmake/test-static-assert.cpp" CMAKE_FLAGS -DCMAKE_CXX_STANDARD=11)
```

https://github.com/monero-project/monero/pull/7971/files#R791
```CMake
try_compile(STATIC_ASSERT_RES "${CMAKE_CURRENT_BINARY_DIR}/static-assert" "${CMAKE_CURRENT_SOURCE_DIR}/cmake/test-static-assert.c" CMAKE_FLAGS -DCMAKE_CXX_STANDARD=11)
```

The line 791 is particularily disturbing, since it sets C++11 standard, while a C file is being tested.

The rest is, like you said, in the external projects.

selsta:
I intentionally kept them at 11, no reason to set this file to C++14: https://github.com/monero-project/monero/blob/master/cmake/test-static-assert.cpp

> The line 791 is particularily disturbing, since it sets C++11 standard, while a C file is being tested.

#7972

mj-xmr:
OK then. Completely forgot about #7972

PR #7986, opened 2021-10-01
wallet2: keep around transaction prefix for confirmed transfers

tobtoht:
Would allow showing output pubkeys alongside destination addresses. Requested downstream [here](https://git.featherwallet.org/feather/feather/issues/379).


UkoeHB (approved):
Odd that this info wasn't already included in `confirmed_transfer_details`.

SamsungGalaxyPlayer:
Ty @UkoeHB and @tobtoht !

PR #7989, opened 2021-10-04
cmake: don't optimize debug build

selsta:


mj-xmr:
Why do you want to disable this feature?

From [GCC Manual](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html):

-Og
Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0.

Like -O0, -Og completely disables a number of optimization passes so that individual options controlling them have no effect. Otherwise -Og enables all -O1 optimization flags except for those that may interfere with debugging:

-fbranch-count-reg  -fdelayed-branch 
-fdse  -fif-conversion  -fif-conversion2  
-finline-functions-called-once 
-fmove-loop-invariants  -fmove-loop-stores  -fssa-phiopt 
-ftree-bit-ccp  -ftree-dse  -ftree-pta  -ftree-sra

selsta:
Why optimize a debug build instead of using the CMake default? It just results in less information in gdb which makes debugging harder.

If someone wants optimizations then they can use release build or manually change it in CMake.

See for example https://stackoverflow.com/questions/31435771/variables-optimized-out-with-g-and-the-og-option

PR #7495, opened 2021-03-10
REF: copyright years changed to 2021.

JaydenElliott:
Copyright years changed to 2021. This includes copyright text of the following format:
- Copyright (c) 2018-2020 
- Copyright (c) 2018

Process: Scanned through all file types searching for this pattern and replaced with appropriate year.

PR #7992, opened 2021-10-04
Update copyright year to 2021

getgle:
it is the current year


selsta:
Thanks. Please see #7495

PR #7999, opened 2021-10-07
p2p: drop-in replacement for serialization 

perfect-daemon:


vtnerd:
@cirocosta 

This removes the binary->DOM portion of serialization and instead does direct binary->C++. This (likely) improves the performance and eliminates _some_ of the de-serialization attacks seen 10 months ago. Unpacking integers still needs aggressive checks (and this patch may have removed that, haven't gone through it entirely). This also (kind of) makes use of cpp files a bit better, but still has lots of stuff in headers (might be able to clean some of that up).

The primary downsides: (1) this isn't actually a drop-in for serialization (no JSON support), and (2) the templating to make this work is fairly aggressive. And although this is admittedly subjective, this looks like one of the worse interfaces for a C++ serialization utility I've seen. The no-DOM thing can be achieved with a bit more simplicity.

As a point of reference - I wrote an alternative [serialization](https://github.com/vtnerd/monero-lws/blob/develop/src/wire/read.h) and [deserialization](https://github.com/vtnerd/monero-lws/blob/develop/src/wire/write.h) utility with the same characteristics _plus_ multi-format support, and but haven't had serious plans on integrating that into epee (not anytime soon). I was primarily concerned with the template usage, which is far more minimal than whats contained in this PR. This attempt also is more straightforward in [usage](https://github.com/vtnerd/monero-lws/blob/develop/src/wire/json.h#L45) because its closer to the style of the [existing code](https://github.com/vtnerd/monero-lws/blob/develop/src/rpc/rates.cpp#L64).

selsta:
Bootstrap daemon mode seems broken currently with this patch.

```
./monerod --bootstrap-daemon-address selsta2.featherwallet.net:18081 --no-sync
status
```

Prints the following error


```
2022-09-08 20:45:14.680	I Failing RPC /getinfo due to peer return status 
2022-09-08 20:45:14.710	I Failing RPC hard_fork_info due to peer return status 
```

PR #8000, opened 2021-10-08
add --sync-notify option

moneromooo-monero:
Triggered when starting/stopping syncing

%s: 1 if starting syncing, 0 if stopping
%h: current chain height
%t: target chain height


jeffro256:
Looking through older PRs, I think this is a great idea, the only way I think it could be made better is if the program was also passed an argument which is a start/stop flag

moneromooo-monero:
What would it be that's not %s ?

jeffro256:
I don't know what I meant by that comment lol. I guess I didn't see the `%s`? I'm gonna review this more in-depth over the next couple days

PR #8015, opened 2021-10-20
make_unique (and shared) in wallet2 instead of new

mj-xmr:
Thanks to the recently merged @selsta's https://github.com/monero-project/monero/pull/7971, which introduced C++14 features, it's now possible to use `std::make_unique` for a bit of more safety.

From [modern-cpp-features](https://github.com/AnthonyCalandra/modern-cpp-features#stdmake_unique):

`std::make_unique` is the recommended way to create instances of `std::unique_ptr`s due to the following reasons:
* Avoid having to use the `new` operator.
* Prevents code repetition when specifying the underlying type the pointer shall hold.
* Most importantly, it provides exception-safety. Suppose we were calling a function `foo` like so:
```c++
foo(std::unique_ptr<T>{new T{}}, function_that_throws(), std::unique_ptr<T>{new T{}});
```
The compiler is free to call `new T{}`, then `function_that_throws()`, and so on... Since we have allocated data on the heap in the first construction of a `T`, we have introduced a leak here. With `std::make_unique`, we are given exception-safety:
```c++
foo(std::make_unique<T>(), function_that_throws(), std::make_unique<T>());
```

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -492,7 +492,7 @@ std::unique_ptr<tools::wallet2> make_basic(const boost::program_options::variabl
     catch (const std::exception &e) { }
   }
 
-  std::unique_ptr<tools::wallet2> wallet(new tools::wallet2(nettype, kdf_rounds, unattended));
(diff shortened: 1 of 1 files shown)


selsta:
I agree with mooo here.

PR #8016, opened 2021-10-20
epee: avoid ADL selecting C++14 std::quoted

selsta:


vtnerd (approved):
You can also do `(quoted)(...)` at every function call to disable ADL lookups (totally not a joke, although maybe C++ is). This is the more obvious/natural way though.

PR #8024, opened 2021-10-27
unit tests: Harden flaky decoy output selection tests

j-berman:
### Overview

Since gamma tests are not deterministic, there is a non-0 chance they can fail. The recent set of changes to the decoy selection algo (#7821) seem to cause a marginally higher failure rate in the tests. This PR makes small changes to the tests to get them to pass at a much higher frequency, so that they don't disrupt CI, while still maintaining a solid bar to pass.

It would be nice to eventually have deterministic tests using a random seed(s) so that results are reproducible.

### Summary of changes

**select_outputs.gamma**

This tests the median age of selected outputs is the expected median. I decreased the expected median to account for recent changes to the algorithm & added logic to re-attempt the test with a 10x larger sample size if the test fails on first try.

**select_outputs.density**

This tests that outputs from blocks of various sizes are picked in sufficient proportion by the algorithm. I allowed a wider deviation from chain data to selected data for larger blocks, and a smaller deviation for smaller blocks (the allowed deviation is proportional to size now) + tested some other sensible heuristics.

**select_outputs.same_distribution**

This tests that the distribution of outputs picked matches the distribution of blocks which they are part of are picked. I allowed a slightly wider deviation from chain data to selected data.

### Results

I ran all gamma tests 130k times *before* making the final changes included in this PR (special thank you to @Gingeropolous  for letting me use his monster machines), here were my results:

**9 failures total**
`select_outputs.gamma:` 0 failures
`select_outputs.density`: 2 failures
`select_outputs.same_distribution`: 7 failures

#### select_outputs.density

Using a `MAX_DEVIATION` of 2.0, I got the following 2 errors:

```
Expected: (MAX_DEVIATION * chain_ratio) > (selected_ratio), actual: 0.00381708 vs 0.003865
Expected: (MAX_DEVIATION * chain_ratio) > (selected_ratio), actual: 0.00743647 vs 0.007587
```

Going with a `MAX_DEVIATION` > 2.113 would have avoided both. So I increased the `MAX_DEVIATION` to 2.25 in this final PR just to add a decent buffer within reason.

#### select_outputs.same_distribution

Sticking with the original allowed `avg_dev` of 0.02, I got the following 7 errors:

```
Expected: (avg_dev) < (0.02), actual: 0.020848 vs 0.02
Expected: (avg_dev) < (0.02), actual: 0.0204963 vs 0.02
Expected: (avg_dev) < (0.02), actual: 0.0201313 vs 0.02
Expected: (avg_dev) < (0.02), actual: 0.0200123 vs 0.02
Expected: (avg_dev) < (0.02), actual: 0.0214195 vs 0.02
Expected: (avg_dev) < (0.02), actual: 0.0218481 vs 0.02
Expected: (avg_dev) < (0.02), actual: 0.020335 vs 0.02
```

Bumping to 0.0219 would have avoided all the above errors, so I went with 0.025 following the same logic.

### DIFF
--- a/tests/unit_tests/output_selection.cpp
+++ b/tests/unit_tests/output_selection.cpp
@@ -109,28 +109,48 @@ TEST(select_outputs, order)
     offset = n_outs += (n); \
   }
 
+void pick_outputs(size_t npicks, std::vector<uint64_t> offsets, size_t n_outs, tools::gamma_picker picker, double &median)
+{
+  std::vector<double> ages(npicks);
+  double age_scale = 120. * (offsets.size() / (double)n_outs);
+  for (size_t i = 0; i < ages.size(); )
+  {
+    uint64_t o = picker.pick();
+    bool acceptable_bad_pick = o == std::numeric_limits<uint64_t>::max();
+    if (acceptable_bad_pick)
+      continue;
+    ages[i] = (n_outs - 1 - o) * age_scale;
+    ASSERT_GE(ages[i], CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE * 120);
+    ASSERT_LE(ages[i], offsets.size() * 120);
+    ++i;
+  }
+  median = epee::misc_utils::median(ages);
+}
+
+// TODO: deterministic tests would be nice
 TEST(select_outputs, gamma)
 {
   std::vector<uint64_t> offsets;
 
   MKOFFSETS(300000, 1);
   tools::gamma_picker picker(offsets);
-  std::vector<double> ages(100000);
-  double age_scale = 120. * (offsets.size() / (double)n_outs);
-  for (size_t i = 0; i < ages.size(); )
-  {
-    uint64_t o = picker.pick();
-    if (o >= n_outs)
-      continue;
-    ages[i] = (n_outs - 1 - o) * age_scale;
-    ASSERT_GE(ages[i], 0);
-    ASSERT_LE(ages[i], offsets.size() * 120);
-    ++i;
-  }
-  double median = epee::misc_utils::median(ages);
-  MDEBUG("median age: " << median / 86400. << " days");
-  ASSERT_GE(median, 1.3 * 86400);
-  ASSERT_LE(median, 1.4 * 86400);
+
+  size_t NPICKS = 100000;
+  double median = 0;
+  pick_outputs(NPICKS, offsets, n_outs, picker, median);
+
+  // expected median is ~115,100s, or 1.33 * 86400. Calculated by running the algorithm
+  // under the same conditions as the paper (outputs <10 blocks old can be selected, and
+  // are selected from chain tip)
+  double MEDIAN_LOWER_BOUND = 1.28 * 86400;
+  double MEDIAN_UPPER_BOUND = 1.38 * 86400;
+
+  // should be rare, but if needed, 10x NPICKS and try again to get a more accurate median
+  if (median < MEDIAN_LOWER_BOUND || median > MEDIAN_UPPER_BOUND)
+    pick_outputs(NPICKS * 10, offsets, n_outs, picker, median);
+
+  ASSERT_GE(median, MEDIAN_LOWER_BOUND);
+  ASSERT_LE(median, MEDIAN_UPPER_BOUND);
 }
 
 TEST(select_outputs, density)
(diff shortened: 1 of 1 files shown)


tests/unit_tests/output_selection.cpp
```diff
@@ -109,28 +109,48 @@ TEST(select_outputs, order)
...
-  MKOFFSETS(300000, 1);
-  tools::gamma_picker picker(offsets);
-  std::vector<double> ages(100000);
+  std::vector<double> ages(npicks);
   double age_scale = 120. * (offsets.size() / (double)n_outs);
   for (size_t i = 0; i < ages.size(); )
   {
     uint64_t o = picker.pick();
-    if (o >= n_outs)
+    bool acceptable_bad_pick = o == std::numeric_limits<uint64_t>::max();
```
iamamyth, line 119: Minor style issue, but I think using variable assignments in place of comments only further confuses things:
1. A variable assignment implies the need for repeated references, but this is a throw-away expression which results in a continue.
2. The exact syntax in this situation (with many equal signs near each other) proves difficult to read, at least for me.

If you find max as a placeholder for "no real pick" confusing, then write a clarifying comment. The variable name alone doesn't work, because it says, in effect, the same thing as the code would without the variable: "Allow this max pick"; what it doesn't say, which may or may not be self-evident, is why such a choice should be allowed.

tests/unit_tests/output_selection.cpp
```diff
@@ -109,28 +109,48 @@ TEST(select_outputs, order)
...
     ++i;
   }
-  double median = epee::misc_utils::median(ages);
-  MDEBUG("median age: " << median / 86400. << " days");
-  ASSERT_GE(median, 1.3 * 86400);
-  ASSERT_LE(median, 1.4 * 86400);
+  median = epee::misc_utils::median(ages);
+}
+
+// TODO: deterministic tests would be nice
```
iamamyth, line 130: I'd say they're not just "nice", but necessary, though I have noted other tests in this suite failing due to non-determinism. A flaky test suite, particularly when the underlying topic proves somewhat complex, only makes everything worse:

1. It generates a lot of noise in ordinary CI runs, which breaks simple default rules like "don't merge stuff that broke the tests", and makes work for contributors who need to check what failed, and whether the new commit introduced the failure.
2. It makes it hard to tell whether a new commit, even if it does interact with the test suite, actually made things worse. Was the commit wrong, or the test? What if the commit still passes the test sometimes?

I appreciate the motivation: You want to test the statistical properties of the function, and that means you need a lot of samples. To address that challenge, I suggest either:
1. Control the environment such that you can guarantee an outcome (force seed the generator, etc);
2. Limit when this test suite runs and increase the expense of the test so it has a very low probability of generating false positives.

tests/unit_tests/output_selection.cpp
```diff
@@ -109,28 +109,48 @@ TEST(select_outputs, order)
...
+  double median = 0;
+  pick_outputs(NPICKS, offsets, n_outs, picker, median);
+
+  // expected median is ~115,100s, or 1.33 * 86400. Calculated by running the algorithm
+  // under the same conditions as the paper (outputs <10 blocks old can be selected, and
+  // are selected from chain tip)
+  double MEDIAN_LOWER_BOUND = 1.28 * 86400;
+  double MEDIAN_UPPER_BOUND = 1.38 * 86400;
+
+  // should be rare, but if needed, 10x NPICKS and try again to get a more accurate median
```
iamamyth, line 148: What if you iteratively added new picks and instead made an assertion regarding convergence to the expected median? So, you'd keep the old picks, add new ones, re-compute the combined median, and expect that the median gets closer to target. You can add a tolerance to the convergence expectation such that, for example, a single non-converging iteration tries again, but two or three or whatever would fail. This way, you can abort a lot quicker on what are probable failures, but also not fail on what are really OK implementations.

tobtoht:
I would like to go forward with this PR despite room for improvement.

I reran CI for this PR, but two tests are now failing:

```
 /__w/monero/monero/tests/unit_tests/output_selection.cpp:172: Failure
Expected: (idx) < (offsets.size() - 10), actual: 299990 vs 299990

[  FAILED  ] select_outputs.density (115 ms)
[ RUN      ] select_outputs.same_distribution
/__w/monero/monero/tests/unit_tests/output_selection.cpp:225: Failure
Expected: (idx) < (offsets.size() - 10), actual: 299990 vs 299990

[  FAILED  ] select_outputs.same_distribution (119 ms)
```

PR #8027, opened 2021-10-27
cmake: remove including non-existent directory

selsta:
There is no `/usr/include/malloc` on macOS and it builds fine without.

Also CMP0042 always exists with out minimum version, no need to check for it explicitly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -451,10 +451,7 @@ endif()
 include_directories(external/rapidjson/include external/easylogging++ src contrib/epee/include external external/supercop/include)
 
 if(APPLE)
-  include_directories(SYSTEM /usr/include/malloc)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -451,10 +451,7 @@ endif()
 include_directories(external/rapidjson/include external/easylogging++ src contrib/epee/include external external/supercop/include)
 
 if(APPLE)
-  include_directories(SYSTEM /usr/include/malloc)
```
mj-xmr, line 454: If it was declared as SYSTEM include, then it means, that the intention was to suppress some compiler warnings coming from there. But since the directory doesn't exist, it doesn't matter anymore.

mj-xmr (approved):

PR #8033, opened 2021-10-31
Avoid nullptr dereference when constructing `Blockchain` and `tx_memory_pool`

DarkWingMcQuack:
Until now `Blockchain` and `tx_memory_pool` is constructed like the following:
```cpp
std::unique_ptr<Blockchain> core_storage;
tx_memory_pool m_mempool(*core_storage);
core_storage.reset(new Blockchain(m_mempool));
```
This contains a nullptr dereference.

This PR fixes this by introducing a new struct `BlockchainAndPool`
```cpp
struct BlockchainAndPool {
  cryptonote::Blockchain blockchain;
  cryptonote::tx_memory_pool tx_pool;

  BlockchainAndPool() : blockchain(tx_pool), tx_pool(blockchain) {}
};
```
which is used to construct `Blockchain` and `tx_memory_pool` to avoid the nullptr dereferencation.

### DIFF
--- a/src/blockchain_utilities/blockchain_and_pool.h
+++ b/src/blockchain_utilities/blockchain_and_pool.h
@@ -1,0 +1,44 @@
+// Copyright (c) 2014-2020, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+// this list
+//    of conditions and the following disclaimer in the documentation and/or
+//    other materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors
+// may be
+//    used to endorse or promote products derived from this software without
+//    specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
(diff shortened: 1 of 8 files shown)


src/blockchain_utilities/blockchain_and_pool.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "cryptonote_core/blockchain.h"
+#include "cryptonote_core/tx_pool.h"
+
+struct BlockchainAndPool {
+  cryptonote::Blockchain blockchain;
+  cryptonote::tx_memory_pool tx_pool;
+
+  BlockchainAndPool() : blockchain(tx_pool), tx_pool(blockchain) {}
+};
```
UkoeHB, line 44: It seems strange to initialize an object with an uninitialized other object. Would this pattern be more intuitive?
```
struct BlockchainAndPool {
  cryptonote::Blockchain blockchain;
  cryptonote::tx_memory_pool tx_pool;

  BlockchainAndPool() : blockchain(), tx_pool(blockchain) {
    blockchain.set_pool(tx_pool);
  }
};
```

DarkWingMcQuack: I was following how `cryptonode::core` does it here:
https://github.com/monero-project/monero/blob/528cd729687fcc73e2c082225af203d5f67b3eb6/src/cryptonote_core/cryptonote_core.cpp#L228

@moneromooo-monero has proposed this pattern on IRC/Matrix

Maybe a problem with using a setter would be that if the setter is not called the object would be in an invalid state.
Another reason is that `Blockchain` and `tx_memory_pool` hold a reference to each other and those references need to be initialized in the constructor.

UkoeHB: I see. The original design is dubious but I guess yours is the best solution without more intrusive changes.

selsta (commented):
Please undo the unrelated indentation changes :)

src/blockchain_utilities/blockchain_prune_known_spent_data.cpp
```diff
@@ -160,9 +161,10 @@ int main(int argc, char* argv[])
   const std::string input = command_line::get_arg(vm, arg_input);
 
   LOG_PRINT_L0("Initializing source blockchain (BlockchainDB)");
-  std::unique_ptr<Blockchain> core_storage;
-  tx_memory_pool m_mempool(*core_storage);
-  core_storage.reset(new Blockchain(m_mempool));
+  std::unique_ptr<BlockchainAndPool> core_storage = std::make_unique<BlockchainAndPool>();
+  // tx_memory_pool m_mempool(*core_storage);
+  // core_storage.reset(new Blockchain(m_mempool));
```
selsta, line 166:
```suggestion
```

Unless you kept this intentionally?

selsta:
Please squash.

DarkWingMcQuack:
done

src/blockchain_utilities/blockchain_prune_known_spent_data.cpp
moneromooo-monero, line 166: Forgot those ?

moneromooo-monero (approved):

src/blockchain_utilities/blockchain_prune_known_spent_data.cpp
DarkWingMcQuack: damn it...

UkoeHB:
Why does this PR pass CI? In `core_tests/chaingen.cpp` the `init_blockchain()` function makes a `Blockchain`, but this PR does not touch that file. (I am getting a segfault [at this line](https://github.com/monero-project/monero/blob/cc9ca953a937979c3f4cb0a332db9ac44372191f/tests/core_tests/chaingen.cpp#L163)).

mj-xmr (approved):
Thanks for handling this. It was bothersome.

mj-xmr:
As @UkoeHB noted, in case the PR is supposed to be merged as it is now, let's not forget to address later:

- tests/core_tests/chaingen.cpp init_blockchain()
- tests/block_weight/block_weight.cpp PREFIX_WINDOW()
- tests/unit_tests/output_distribution.cpp get_output_distribution()

selsta:
Continuing in #8924

PR #8052, opened 2021-11-09
crypto: fix ge_p3_is_point_at_infinity()

UkoeHB:
### Summary
The function `ge_p3_is_point_at_infinity()` is naively evaluating field elements that haven't been reduced by the field order.

### Problem severity/impact
Low, this function is not currently used for anything important (only performance testing of some `multiexp` functions). In general, whether a crypto element is reduced mod its group/field order or not, is only important in byte-aware contexts such as byte-wise comparisons (as in this case), or hash inputs (`ge_p3` point representations are never hashed in Monero).

### Fix
Convert all field elements in the input `ge_p3` point to canonical representation (32-byte numbers reduced `mod q`) before evaluating them.

### Visibility
I added test cases with experimentally-found failures of the old function (these occurred in ~1 in 600k random `ge_p3` point representations of the identity element).

### Impact
Unfortunately, doing these conversions somewhat reduces the utility of `ge_p3_is_point_at_infinity()`, which was originally much more efficient than fully converting a `ge_p3` point representation to compressed-point representation with `ge_tobytes()` and doing a byte-wise comparison with the identity element. It is still more efficient, just not _as_ efficient.

### References
- [Paper that introduced extended Twisted Edwards coordinates (i.e. `ge_p3`)](https://eprint.iacr.org/2008/522)

### DIFF
--- a/src/crypto/crypto-ops.c
+++ b/src/crypto/crypto-ops.c
@@ -3830,15 +3830,51 @@ int sc_isnonzero(const unsigned char *s) {
     s[27] | s[28] | s[29] | s[30] | s[31]) - 1) >> 8) + 1;
 }
 
-int ge_p3_is_point_at_infinity(const ge_p3 *p) {
-  // X = 0 and Y == Z
-  int n;
-  for (n = 0; n < 10; ++n)
+int ge_p3_is_point_at_infinity_vartime(const ge_p3 *p) {
+  // https://eprint.iacr.org/2008/522
+  // X == T == 0 and Y/Z == 1
+  // note: convert all pieces to canonical bytes in case rounding is required (i.e. an element is > q)
+  // note2: even though T = XY/Z is true for valid point representations (implying it isn't necessary to
+  //        test T == 0), the input to this function might NOT be valid, so we must test T == 0
+  char result_X_bytes[32];
+  fe_tobytes((unsigned char*)&result_X_bytes, p->X);
+
+  // X != 0
+  for (int i = 0; i < 32; ++i)
   {
-    if (p->X[n] | p->T[n])
-      return 0;
-    if (p->Y[n] != p->Z[n])
+    if (result_X_bytes[i])
       return 0;
   }
-  return 1;
+
+  char result_T_bytes[32];
+  fe_tobytes((unsigned char*)&result_T_bytes, p->T);
+
+  // T != 0
+  for (int i = 0; i < 32; ++i)
+  {
+    if (result_T_bytes[i])
+      return 0;
+  }
+
+  char result_Y_bytes[32];
+  char result_Z_bytes[32];
+  fe_tobytes((unsigned char*)&result_Y_bytes, p->Y);
+  fe_tobytes((unsigned char*)&result_Z_bytes, p->Z);
+
+  // Y != Z
+  for (int i = 0; i < 32; ++i)
+  {
+    if (result_Y_bytes[i] != result_Z_bytes[i])
+      return 0;
+  }
+
+  // is Y nonzero? then Y/Z == 1
+  for (int i = 0; i < 32; ++i)
+  {
+    if (result_Y_bytes[i] != 0)
+      return 1;
+  }
+
+  // Y/Z = 0/0
+  return 0;
 }
--- a/tests/crypto/main.cpp
+++ b/tests/crypto/main.cpp
@@ -259,6 +259,16 @@ int main(int argc, char *argv[]) {
       if (expected != actual) {
         goto error;
       }
+    } else if (cmd == "check_ge_p3_identity") {
+      cerr << "Testing: " << cmd << endl;
+      public_key point;
+      bool expected_bad, expected_good, result_badfunc, result_goodfunc;
+      get(input, point, expected_bad, expected_good);
+      result_badfunc = check_ge_p3_identity_failure(point);
+      result_goodfunc = check_ge_p3_identity_success(point);
+      if (expected_bad != result_badfunc || expected_good != result_goodfunc) {
+        goto error;
+      }
     } else {
       throw ios_base::failure("Unknown function: " + cmd);
     }
--- a/src/crypto/crypto-ops.h
+++ b/src/crypto/crypto-ops.h
@@ -162,4 +162,4 @@ void fe_add(fe h, const fe f, const fe g);
 void fe_tobytes(unsigned char *, const fe);
 void fe_invert(fe out, const fe z);
 
-int ge_p3_is_point_at_infinity(const ge_p3 *p);
+int ge_p3_is_point_at_infinity_vartime(const ge_p3 *p);
--- a/src/ringct/multiexp.cc
+++ b/src/ringct/multiexp.cc
@@ -235,7 +235,7 @@ rct::key bos_coster_heap_conv_robust(std::vector<MultiexpData> data)
   heap.reserve(points);
   for (size_t n = 0; n < points; ++n)
   {
-    if (!(data[n].scalar == rct::zero()) && !ge_p3_is_point_at_infinity(&data[n].point))
+    if (!(data[n].scalar == rct::zero()) && !ge_p3_is_point_at_infinity_vartime(&data[n].point))
       heap.push_back(n);
   }
   points = heap.size();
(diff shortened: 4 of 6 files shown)


src/crypto/crypto-ops.c
```diff
@@ -3831,14 +3831,39 @@ int sc_isnonzero(const unsigned char *s) {
...
+  fe_tobytes((unsigned char*)&result_X_bytes, p->X);
+  fe_tobytes((unsigned char*)&result_Y_bytes, p->Y);
+  fe_tobytes((unsigned char*)&result_Z_bytes, p->Z);
+  fe_tobytes((unsigned char*)&result_T_bytes, p->T);
+
+  // X != 0 || T != 0
+  for (int i = 0; i < 32; ++i)
   {
-    if (p->X[n] | p->T[n])
+    if (result_X_bytes[i] | result_T_bytes[i])
```
vtnerd, line 3849: Does `T` need to be evaluated at all? I think it can be skipped entirely in this function. Dropping the `fe_tobytes` for `T` should be a performance win.

The point at infinity should be `(x,y) = (0, 1)`, and `T` is defined as `(X*Y)/Z` (or `t=x*y`). So if `X` is `0`, then `T` must also be zero ... ?

src/crypto/crypto-ops.c
```diff
@@ -3831,14 +3831,39 @@ int sc_isnonzero(const unsigned char *s) {
 }
 
 int ge_p3_is_point_at_infinity(const ge_p3 *p) {
```
vtnerd, line 3833: Is this a good opportunity to relabel this with `_vartime` at the end? This would follow the typical naming scheme for crypto functions.

tests/crypto/main.cpp
```diff
@@ -259,6 +259,16 @@ int main(int argc, char *argv[]) {
       if (expected != actual) {
         goto error;
       }
+    } else if (cmd == "check_ge_p3_identity") {
```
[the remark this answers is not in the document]
UkoeHB: `tests/crypto` seems to have the most similarities with the test I wanted to add. `tests/unit_tests/crypto.cpp` appears to be testing higher-level functions.

src/crypto/crypto-ops.c
UkoeHB: What you're saying sounds right, but I don't have the expertise to confidently agree. Hopefully we can get more opinions.

src/crypto/crypto-ops.c
UkoeHB: Sounds good to me!

src/crypto/crypto-ops.c
```diff
@@ -3831,14 +3831,39 @@ int sc_isnonzero(const unsigned char *s) {
...
+  // https://eprint.iacr.org/2008/522
+  // X == T == 0 and Y/Z == 1
+  // note: convert all pieces to canonical bytes in case rounding is required (i.e. an element is > q)
+  char result_X_bytes[32];
+  char result_Y_bytes[32];
+  char result_Z_bytes[32];
+  char result_T_bytes[32];
+  fe_tobytes((unsigned char*)&result_X_bytes, p->X);
+  fe_tobytes((unsigned char*)&result_Y_bytes, p->Y);
+  fe_tobytes((unsigned char*)&result_Z_bytes, p->Z);
```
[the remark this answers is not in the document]
UkoeHB: I split them all up.

src/crypto/crypto-ops.c
UkoeHB: One comment/question I got was whether skipping this test would be safe if given an input not produced by a 'correct' method.

dkabel-meister: My understanding of the coordinates used in the reference [https://eprint.iacr.org/2008/522](url) is that X=0 is a sufficient condition for T=0 since Z != 0 by definition and therefore Y is bounded from above.  If the ``point at infinity'' is defined as (X:Y:Z:T)=(0:1:0:1) within the equivalence class of $\lambda$-rescalings, then the definition T=XY/Z, Z\neq 0, implies that checking X=0 and Y/Z=1 is enough. In other words, X=0 and Y/Z=1 is a sufficient condition for the point to a conformal representation of the identity in these coordinates.

UkoeHB: > The point at infinity should be (x,y) = (0, 1), and T is defined as (X*Y)/Z (or t=x*y). So if X is 0, then T must also be zero ... ? @vtnerd 

`T = XY/Z` is only true for valid point representations. If an invalid point representation is passed in (where `Y/Z == 1`, `X == 0`, `T != 0`), then this function should _not_ label it as the point-at-infinity. Here is some algebra (using equation 5 from [the paper](https://eprint.iacr.org/2008/522.pdf)) showing that adding an invalid point `P2` to a normal point `P1` will not produce `P1`: `P1 + P2 != P1`.

```
P1: {X1, Y1, T1, Z1}
P2 (invalid identity): {X2 = 0, Y2 != 0, T2 != 0, Z2 = Y2}

1. apply point-addition: P1 + P2 = P3
X3 = X1Y2*(Z1Y2 - dT1T2) = X1*(Z1Y2^2 - dT1T2Y2)
Y3 = Y1Y2*(Z1Y2 + dT1T2) = Y1*(Z1Y2^2 + dT1T2Y2)
T3 = Y1Y2*X1Y2           = X1Y1Y2^2
Z3 = (Z1Y2 - dT1T2)*(Z1Y2 + dT1T2) = (Z1Y2)^2 - (dT1T2)^2

2. multiply by 1/Y2
X3 = X1Y2*(Z1Y2 - dT1T2) = X1*(Z1Y2 - dT1T2)
Y3 = Y1Y2*(Z1Y2 + dT1T2) = Y1*(Z1Y2 + dT1T2)
T3 = Y1Y2*X1Y2           = X1Y1Y2
Z3 = (Z1Y2 - dT1T2)*(Z1Y2 + dT1T2) = (1/Y2) [ (Z1Y2)^2 - (dT1T2)^2 ]

3. check T-property
X3Y3 = X1Y1 [ (Z1Y2)^2 - (dT1T2)^2 ]
X3Y3/Z3 = X1Y1Y2 = T3

4. multiply by 1/Z1Y2
X3 = X1Y2*(Z1Y2 - dT1T2) = X1 - X1*dT1T2/Z1Y2
Y3 = Y1Y2*(Z1Y2 + dT1T2) = Y1 + Y1*dT1T2/Z1Y2
T3 = Y1Y2*X1Y2           = X1Y1/Z1
Z3 = (Z1Y2 - dT1T2)*(Z1Y2 + dT1T2) = Z1 - (dT1T2)^2/Z1Y2^2

If T2 == 0, then all the junk disappears and P3 == P1 as expected. Can you make the junk go away when T2 != 0?
```

src/crypto/crypto-ops.c
```diff
@@ -3830,15 +3830,51 @@ int sc_isnonzero(const unsigned char *s) {
...
 
-int ge_p3_is_point_at_infinity(const ge_p3 *p) {
-  // X = 0 and Y == Z
-  int n;
-  for (n = 0; n < 10; ++n)
+int ge_p3_is_point_at_infinity_vartime(const ge_p3 *p) {
+  // https://eprint.iacr.org/2008/522
+  // X == T == 0 and Y/Z == 1
+  // note: convert all pieces to canonical bytes in case rounding is required (i.e. an element is > q)
+  // note2: even though T = XY/Z is true for valid point representations (implying it isn't necessary to
```
vtnerd, line 3837: This (as mentioned in your reply to me) seems rather pessimistic, but it doesn't hurt things much.

src/crypto/crypto-ops.c
```diff
@@ -3830,15 +3830,51 @@ int sc_isnonzero(const unsigned char *s) {
...
+
+  // T != 0
+  for (int i = 0; i < 32; ++i)
+  {
+    if (result_T_bytes[i])
+      return 0;
+  }
+
+  char result_Y_bytes[32];
+  char result_Z_bytes[32];
```
vtnerd, line 3860: Only two `char[32]` buffers are needed in this function, but I don't see any reason to block the patch for that -

vtnerd (approved):

dkabel-meister (approved):

UkoeHB:
Squashed

selsta (approved):

PR #8054, opened 2021-11-09
cmake: fix incorrect else syntax

selsta:


selsta:
According to https://cmake.org/cmake/help/latest/command/if.html this is apparently valid legacy syntax.

> Per legacy, the else() and endif() commands admit an optional <condition> argument. If used, it must be a verbatim repeat of the argument of the opening if command.

I guess we can still merge this as it isn't used elsewhere and it's a bit confusing at first sight.

ndorf (approved):

PR #8067, opened 2021-11-17
make_uri only allows short payment id

woodser:


selsta (commented):
We don't want to support separate payment ids, only integrated addresses. So maybe it would be better to completely remove support for the `payment_id` arg and always write an error if someone uses it?

woodser:
Yeah that would make sense to me.

woodser:
Ok an error is thrown if a standalone payment id is provided. Let me know if there are suggestions on the error message.

selsta (approved):

PR #8079, opened 2021-11-23
Remove duplicate word at line 78

git-sgmoore:
Remove duplicate word at line 78


jtgrassie:
The mistake here is not that the word integer is duplicated but that the first "integer" should read "endian".

```diff
- A signed 32-bit little integer integer representing the response from the peer
+ A signed 32-bit little endian integer representing the response from the peer
```

selsta:
Fixed in #8041

PR #8080, opened 2021-11-23
Grammar fix at line 4

git-sgmoore:
Grammar fix at line 4


Uckermark:
I think the original text is alright, because the „is“ is about the usage, but not „these networks“.

selsta:
It seems "is" is correct, I also asked on IRC. If someone disagrees please comment and I can reopen.

PR #8085, opened 2021-11-26
blocks: update checkpoints.dat [release-v0.17]

selsta:


rbrunner7 (approved):
Looks good. Does was it says it does. Got the same hash (sha256 over the "checkpoints.dat" file).

Issue #6639, opened 2020-06-10
[PROPOSAL] Encode restore height as 26th word of the mnemonic seed


dEBRUYNE-1:
The restore height is currently a value that has to be entered manually for wallets that are restored from either the keys or the mnemonic seed. The wallet will essentially ignore blocks (only pulling block hashes) before the restore height and start scanning (looking for transactions that belong to the wallet) from the restore height block. 

User experience is degraded if the user accidentally sets a restore height that is too 'high' (i.e. after the first transaction to the wallet), as the wallet will 'miss' certain or all transactions, thereby causing an improper balance (as well as transaction history) to be displayed. 

In order to improve user experience, we could encode an approximate restore height as additional word of the mnemonic seed. The restore height would then be set automatically upon restoring the wallet, thereby ensuring users will not inadvertently set an erroneous restore height. 

I personally do not see many drawbacks of this proposal. Guides will have to be updated to reflect the new format and users need to be informed. Users further, initially, may be slightly confused due to two different seed formats being present. However, I think ultimately the proposal is net beneficial to user experience.

SChernykh:
This is too obscure, just use `Jun/2020` or something like this instead of a 26th word in the seed.

rbrunner7:
As discussed on IRC, summarizing it here for broader publicity and discussion:

I am in full favor of adding one seed word to encode restore height.

But if we touch the seed system and add a "new" kind of seed encoding the restore height, I vote for taking the chance and add two more worthwhile changes at the same time. (Changing anything with seeds will be a larger endeavor,  and IMHO it would be a strategic mistake to come back to this with "new new" seeds a year later or so).

The "checksum" as implemented with the checksum word being simply a copy of one of the other words is very weak i.e. it does not catch a lot of errors. This can stay a single checksum word, but it should be calculated using a much more robust algorithm going over **all** words of the seed.

Furthermore, one more word should get added as the first word of the seed, encoding a seed version. The words used for the version should be **different** from all other seed words so you can reliably detect whether the first word given is such a version word or not.

This will enable a very robust UX. You can for example generate useful error messages if somebody enters only the first 25 words of a "new" seed for whatever reason, be it conviction that "more than 25 words are wrong", or input forms just not allowing for more words because not yet reworked / upgraded for "new" seeds.

Seed versions would also allow for adjustments in the word list, for whatever crazy reasons that may pop up, like some words becoming "politically incorrect", or more or less banned outright e.g. for Chinese seeds.

[...]

PR #8088, opened 2021-11-27
Daemon RPC: add a new method to get height from a timestamp

tevador:
This is for wallets that need to convert a user-provided date to a blockchain height.

Similar function is currently implemented in wallet2::get_blockchain_height_by_date, but it's done quite inefficiently by pulling whole block blobs from the daemon. The code is ported from there with some optimizations. The precision is increased from 2 days to 1 hour.

This RPC method will be especially useful if #6639 is implemented and the mnemonic seed encodes the wallet birthday timestamp.


hyc:
Is it worth replacing the wallet2 code with this RPC call then?

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -88,7 +88,7 @@ namespace cryptonote
 // advance which version they will stop working with
 // Don't go over 32767 for any of these
 #define CORE_RPC_VERSION_MAJOR 3
-#define CORE_RPC_VERSION_MINOR 8
+#define CORE_RPC_VERSION_MINOR 9
```
hyc, line 91: This will conflict, we're already at 9 in master.

tevador: OK, rebased and bumped the version to 10.

tevador:
> Is it worth replacing the wallet2 code with this RPC call then?

Yes and I'm assuming the GUI and 3rd party wallets will use it too. Also services like MyMonero that scan on behalf of users.

selsta:
> Yes and I'm assuming the GUI and 3rd party wallets will use it too.

One thing to consider is that at least the GUI doesn't have a daemon connection yet during wallet restore. We use this code currently which is a port from wallet2: https://github.com/monero-project/monero-gui/blob/master/js/Wizard.js#L134

tevador:
At some point the GUI has to connect to a daemon to start scanning outputs, so the request could be deferred until then.

I know it is possible to make a reasonably good estimate without a daemon connection (by hardcoding some timestamps and heights), but it's a much less maintainable solution. I tested `get_height_by_time` on stagenet and it works very well even with large gaps between blocks (for example 9 days between 268999 and 269000 etc.).

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -2718,4 +2718,31 @@ namespace cryptonote
...
+      uint64_t height;
+      uint64_t timestamp;
+
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE_PARENT(rpc_access_response_base)
+        KV_SERIALIZE(height)
+        KV_SERIALIZE(timestamp)
+      END_KV_SERIALIZE_MAP()
+    };
+    typedef epee::misc_utils::struct_init<response_t> response;
```
Agorist-Action, line 2745: Prefer `using` to `typedef`. For example: `using response = epee::misc_utils::struct_init<response_t>;`

src/rpc/core_rpc_server.cpp
```diff
@@ -3570,6 +3570,93 @@ namespace cryptonote
...
+    if((uint64_t)time(NULL) <= req.timestamp)
+    {
+      error_resp.code = CORE_RPC_ERROR_CODE_TOO_BIG_TIMESTAMP;
+      error_resp.message = "Requested timestamp is in the future";
+      return false;
+    }
+    CHECK_PAYMENT(req, res, COST_PER_HEIGHT_BY_TIME);
+
+    constexpr uint64_t buffer_time = 1 * 60 * 60; //1 hour
+    constexpr uint64_t buffer_height = 1 * 30;    //1 hour
```
Agorist-Action, line 3589: Is this 30 from 60 minutes in an hour divided by a block every two minutes? In other words, what is `buffer_height`?

src/rpc/core_rpc_server.cpp
```diff
@@ -3570,6 +3570,93 @@ namespace cryptonote
     return true;
   }
   //------------------------------------------------------------------------------------------------------------------------------
+  bool core_rpc_server::on_get_height_by_time(const COMMAND_RPC_GET_HEIGHT_BY_TIME::request& req, COMMAND_RPC_GET_HEIGHT_BY_TIME::response& res, epee::json_rpc::error& error_resp, const connection_context *ctx)
```
Agorist-Action, line 3573: Might want to also follow up with a request to add this method to the online developer documentation if it gets merged. See [this](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html).

Agorist-Action (commented):
Suggestions for this MR.

src/rpc/core_rpc_server.cpp
tevador: Refer to lines 3638-3642. The binary search ends when the boundary block heights are closer than `buffer_height`. The search uses both a time-based and a height-based condition to ensure a good behavior even with irregular block times.

src/rpc/core_rpc_server_commands_defs.h
tevador: That is not my decision. I'm simply following the established coding practice for this header file.

src/rpc/core_rpc_server.cpp
tevador: Yes, this should be done if this PR ever gets merged.

src/rpc/core_rpc_server.cpp
```diff
@@ -3570,6 +3570,93 @@ namespace cryptonote
...
+
+    uint64_t height_min, height_mid, height_max;
+    uint64_t timestamp_min = 0, timestamp_mid, timestamp_max = 0;
+
+    height_min = 0;
+    height_max = m_core.get_current_blockchain_height() - 1;
+
+    auto& db = m_core.get_blockchain_storage().get_db();
+
+    for (;;)
```
serhack, line 3599: I don't like to be fussy or pedantic, but I'm not too sure about these endless cycles. Would be possible to implement the same functionality without the for(;;) and break? It's very tricky to manage all the hypothetical branches imho.

woodser:
Any chance this PR can be revived and move forward?

Without this, there is currently no way for RPC clients to restore a wallet from a date, since there is no way to translate from the date to the restore height.

selsta:
ping @SNeedlewoods, this might be interesting for you if you add support for date to blockchain height approximation

PR #8090, opened 2021-11-28
CI: Mac OS-11 transitioning before 13 Dec

mj-xmr:
GH Actions will replace the `macOS-10.15` (aliased as `macOS-latest`)  with `Mac OS-11` on 13 Dec.

The change uses an OS Matrix, to make the change smoother, as well as future changes and backward compatibility efforts.

[Reference](https://github.com/actions/virtual-environments/issues/4060)

> Breaking changes
> macOS-11 is ready to be the default version for the "macos-latest" label in GitHub Actions and Azure DevOps.

> Target date
> This change will be rolled out over a period of several weeks beginning on September, 15. We plan to complete the migration > by November, 3 December, 13.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,9 +18,15 @@ env:
 
 jobs:
   build-macos:
-    runs-on: macOS-latest
+    runs-on: ${{ matrix.os }}
     env:
       CCACHE_TEMPDIR: /tmp/.ccache-temp
+    strategy:
+      matrix:
+        os: [macOS-latest, macOS-11]
+# TODO: After 13 Dec. 2021 (*) the above line should become either:  (* https://github.com/actions/virtual-environments/issues/4060 )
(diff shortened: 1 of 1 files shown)


selsta (commented):
Hmm, I don't see much point in this. macOS 10.15 or macOS 11 is irrelevant for us, and Mac CI jobs are quite limited, so I would prefer to not run 2 per commit.

Basically we can keep the current code as is without any changes, it won't break.

mj-xmr:
Thanks to this branch, we can day it for sure.
So shall I change the line to

`os: [macOS-latest]`

for easier transition in the future,  or simply drop this PR?

selsta:
> simply drop this PR

This PR has shown that it compiles with macOS 11 so I would drop it.

mj-xmr:
Checks successful before the deadline of 13 Dec.

PR #8101, opened 2021-12-01
Bump versions in DOCKRUN.md and add log commands

sethforprivacy:


contrib/gitian/DOCKRUN.md
```diff
@@ -74,12 +74,18 @@ You can set other options for that script by setting the OPT variable when runni
...
-OPT="-j 8" ./dockrun.sh v0.17.2.3
+OPT="-j 8" ./dockrun.sh v0.17.3.0
 ```
 
 You can also examine the build and install logs by running a shell in the container, e.g.
 
 ```bash
+# Tail running logs
+docker exec -it gitrun /bin/bash
+tail -f builder/var/*.log
```
hyc, line 85: I would use `tail -F builder/var/build.log`

contrib/gitian/DOCKRUN.md
```diff
@@ -56,7 +56,7 @@ The dockrun.sh script will do everything to build the binaries. Just specify the
 version to build as its only argument, e.g.
 
 ```bash
-./dockrun.sh v0.17.2.3
+./dockrun.sh v0.17.3.0
```
hyc, line 59: Since this is just an example, does it really need to be updated for every release?

contrib/gitian/DOCKRUN.md
```diff
@@ -74,12 +74,18 @@ You can set other options for that script by setting the OPT variable when runni
...
 
 ```bash
+# Tail running logs
+docker exec -it gitrun /bin/bash
+tail -f builder/var/*.log
+
+# Inspect logs
 docker exec -it gitrun /bin/bash
 more builder/var/install-linux.log
+more builder/var/build.log
```
hyc, line 90: `build.log` is the name that gitian writes to while the build is running. When the build is finished it gets moved to build-<os>.log, so that's what you would want to check on a successful run. Of course, if it fails during the build then it would still be build.log.

contrib/gitian/DOCKRUN.md
sethforprivacy: I just wanted to bump it while I was in there anyways, it's obviously not required :P

sethforprivacy:
Made the requested changes, thanks @hyc.

contrib/gitian/DOCKRUN.md
```diff
@@ -74,12 +74,19 @@ You can set other options for that script by setting the OPT variable when runni
...
+# Run build processes with 8 threads
+OPT="-j 8" ./dockrun.sh v0.17.3.0
 ```
 
 You can also examine the build and install logs by running a shell in the container, e.g.
 
 ```bash
+# Tail running logs
+docker exec -it gitrun /bin/bash
+tail -F builder/var/*.log
```
hyc, line 86: Do you really want *.log here? That will open all existing log files (including finished ones from previous builds).

sethforprivacy: Should it instead be install.log and build.log (without the OS specified)?

I'm just trying to make it simple for people to see progress, I don't think it really matters that it will print a few lines of old logs.

hyc: OK, fair enough.

hyc: OK, fair enough.

hyc (approved):

hyc:
Oh hey, since we're here and not merged yet - how about adding a few lines on where to find the output archives.

sethforprivacy:
> Oh hey, since we're here and not merged yet - how about adding a few lines on where to find the output archives.

I actually have no idea where to find them, so not sure I can help there, sadly.

sethforprivacy:
@hyc I did add a note just now on SSH keys in Github, as the script currently uses SSH instead of HTTPS for git.

hyc:
Ah yeah, it has to use SSH since it needs push access to your fork of gitian sigs.

The outputs are on the gitrun docker in out/&lt;version>/&lt;platform>

sethforprivacy:
@hyc the last remaining piece is Mac build verification -- even when downloading the SDK the dockrun.sh script doesn't seem to build/verify Mac builds.

Is there a special flag needed for that in some step of the process?

sethforprivacy:
i.e.:

```bash
tiny in tiny in monero/contrib/gitian on  HEAD (ab18fea) [?] ❯ ./dockrun.sh v0.17.3.0

44f63bab270db179bbc0b3560435757baff305e8a9379df169f88a6c53afcad1
Cannot build for MacOS, SDK does not exist. Will build for other OSes
v0.17.3.0
Note: checking out 'v0.17.3.0'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at ab18fea35 Merge pull request #8096

Checking Depends Freshness

make: Entering directory '/home/ubuntu/builder/inputs/monero/contrib/depends'
make[1]: Entering directory '/home/ubuntu/builder/inputs/monero/contrib/depends'
Fetching boost_1_64_0.tar.bz2 from https://downloads.sourceforge.net/project/boost/boost/1.64.0/
^C
tiny in tiny in monero/contrib/gitian on  HEAD (ab18fea) [?] took 3s ❯ ls -la
total 46256
drwxrwxr-x   4 tiny tiny     4096 Dec  2 14:28 .
drwxrwxr-x  10 tiny tiny     4096 Nov 29 19:49 ..
drwxrwxr-x   2 tiny tiny     4096 Dec  1 13:50 docker
-rw-rw-r--   1 tiny tiny     3231 Nov 29 19:49 DOCKRUN.md
-rwxrwxr-x   1 tiny tiny     3480 Nov 29 19:49 dockrun.sh
-rw-rw-r--   1 tiny tiny     3753 Nov 29 19:49 gitian-android.yml
-rwxrwxr-x   1 tiny tiny     9543 Nov 29 19:49 gitian-build.py
-rw-rw-r--   1 tiny tiny     3814 Nov 29 19:49 gitian-freebsd.yml
-rw-rw-r--   1 tiny tiny     5710 Nov 29 19:49 gitian-linux.yml
-rw-rw-r--   1 tiny tiny     3663 Nov 29 19:49 gitian-osx.yml
-rw-rw-r--   1 tiny tiny     4213 Nov 29 19:49 gitian-win.yml
-rw-rw-r--   1 tiny tiny 47285925 Dec  2 14:28 MacOSX10.11.sdk.tar.gz
-rw-rw-r--   1 tiny tiny    10056 Nov 29 19:49 README.md
drwxrwxr-x 107 tiny tiny     4096 Dec  1 16:55 sigs
tiny in tiny in monero/contrib/gitian on  HEAD (ab18fea) [?] ❯
```

hyc:
Yeah that's strange. It should see it and copy it into the container. WorksForMe anyway. No, no special flag needed, it should just grab it when the SDK is present in your contrib/gitian directory.

sethforprivacy:
Pushed new git clone and cd commands to clarify where users pwd should be when they start running commands post-Docker install.

hyc:
regarding the outputs, something like this would be useful:
```bash
mkdir out
docker cp gitrun:/home/ubuntu/out/&lt;version> out
```
Will copy all of the generated archives for the specified version from the container to the host.

sethforprivacy:
> regarding the outputs, something like this would be useful:
> 
> ```shell
> mkdir out
> docker cp gitrun:/home/ubuntu/out/<version> out
> ```
> 
> Will copy all of the generated archives for the specified version from the container to the host.

Added that set of commands as well. Should be all set now AFAICT.

hyc (approved):

HoverHalver:
There are many `git clone ...` commands here and there, manual and in the script(s).
afaik, when you do that you are leaking your IP to github.com.
It is certainly fine for a majority of people, but maybe the doc could mention that point ?
And also that for people willing to avoid such leakage, doing all the needed `git clone ...` with torsocks before launching the python script is a solution (as mentioned by Rucknium in the monero-dev room).
Just suggestions.

sethforprivacy:
> There are many `git ...` (eg `git clone ...`) commands here and there, manual and in the script(s). afaik, when you do that you are leaking your IP to github.com. It is certainly fine for a majority of people, but maybe the doc could mention that point ? And also that for people willing to avoid such leakage, doing all the needed `git clone ...` with `torsocks` before launching the python script is a solution (as mentioned by Rucknium in the monero-dev room).
> 
> Of course, this IP leakage happens with all other git commands. I guess it is probably not feasible to protect all the git commands with torsocks or via another way.

While that is a legitimate use-case, it's probably well beyond the scope of this document.

Would probably be good to have a general "Using git privately" doc or guide we could link out to, however.

hyc:
> While that is a legitimate use-case, it's probably well beyond the scope of this document.

Agreed. Aside from that, there's multiple HTTP requests to download all of the other dependencies, as well as all of the package manager requests to obtain packages from the Ubuntu repos, etc... If you're worried about leaking your IP address, you need to do more than just wrap `git` with torsocks. If you're worried about that you should already be using something like Tails and protecting all traffic on your OS.

HoverHalver:
One more suggestion for the readme.md doc.
Disk requirements
The uploading of all the libraries needed for the compilations necessitates :
- ~500 MB in /var/cache/apt-cacher-ng.
- ~4.0 GB in /home/gitianuser/.
Verify you have this place available before beginning, and do some cleaning if not.
If a `partition full` error happens, the process will stop, and you'll have to resume (and clean nevertheless).
In case this is not your first gitian Monero build, it is similarly wise to clean the concerned repos from previous stuff before beginning.
Just suggestions of course.

PR #8102, opened 2021-12-02
Update Dockerfile

MinhasMA:


selsta:
Any reason for this change?

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Multistage docker build, requires docker 17.05
 
 # builder stage
-FROM ubuntu:20.04 as builder
+FROM ubuntu:latest as builder
```
mj-xmr, line 4: It sounds like a good idea for the version to be self-adjusting across time, but I'd prefer having the version fixed in stone, until there's somebody, who updates it to the next latest version, and collects the medal for fixing all the resulting incompatibilities.

Your proposed change would strike you when you're least expecting it.

Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN set -ex && \
         pkg-config \
         gperf
 
-WORKDIR /src
+WORKDIR /src/1
```
mj-xmr, line 22: Could you explain this?
Did you mean something like `/src/temp` ?

Dockerfile
ndorf: I agree, the base should be a known-good version.

selsta:
Closing as the author did not reply.

PR #8103, opened 2021-12-04
functional_tests: update uri test with external pid

selsta:
#8067

### DIFF
--- a/tests/functional_tests/uri.py
+++ b/tests/functional_tests/uri.py
@@ -142,15 +142,11 @@ class URITest():
         assert res.uri.recipient_name == utf8string[0]
         assert not 'unknown_parameters' in res or len(res.unknown_parameters) == 0
 
-        res = wallet.make_uri(address = address, recipient_name = utf8string[0], tx_description = utf8string[1], amount = 1000000000000, payment_id = '1' * 64)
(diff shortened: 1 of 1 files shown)


tests/functional_tests/uri.py
```diff
@@ -142,16 +142,6 @@ def test_monero_uri(self):
         assert res.uri.recipient_name == utf8string[0]
         assert not 'unknown_parameters' in res or len(res.unknown_parameters) == 0
 
-        res = wallet.make_uri(address = address, recipient_name = utf8string[0], tx_description = utf8string[1], amount = 1000000000000, payment_id = '1' * 64)
```
mj-xmr, line 145: How about the following, to be able to test if the error is being detected:
```Python
try:
    res = wallet.make_uri(address = address, recipient_name = utf8string[0], tx_description = utf8string[1], amount = 1000000000000, payment_id = '1' * 64)
    except Exception:
       # 8067
       # make_uri only allows short payment id
       assert True
    else:
       # Exception not raised
      assert False

    # spaces must be encoded as %20
   (...)
```

mj-xmr (changes requested):

tests/functional_tests/uri.py
selsta: updated

mj-xmr (approved):

PR #8105, opened 2021-12-06
gitian: optional APT cacher for docker builds

vdo:
The requirement of `apt-cacher-ng` becomes optional and enabled only if the binary is present. This opens the possibility to run deterministic builds in a wider range of distributions. I used this branch to run mine on Manjaro:

https://github.com/monero-project/gitian.sigs/commit/8d64b1542c9b135a4b128389423fe4249ac7fd87

### DIFF
--- a/contrib/gitian/dockrun.sh
+++ b/contrib/gitian/dockrun.sh
@@ -7,12 +7,19 @@ fi
 VERSION=$1
 
 DOCKER=`command -v docker`
-CACHER=`command -v apt-cacher-ng`
+CACHER_BIN="/usr/sbin/apt-cacher-ng"
 
-if [ -z "$DOCKER" -o -z "$CACHER" ]; then
-	echo "$0: you must first install docker.io and apt-cacher-ng"
-	echo "  e.g. sudo apt-get install docker.io apt-cacher-ng"
+if [ -z "$DOCKER" ]; then
+	echo "$0: you must first install docker.io"
+	echo "  e.g. sudo apt-get install docker.io"
 	exit 1
+fi
+
+# only use APT cacher if package is present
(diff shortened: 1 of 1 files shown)


hyc:
I'd recommend, instead, adapting dockrun.sh to use apt-cacher-ng inside the gitrun container and pointing the gitian-target container at it. Without apt-cacher all of the other container builds go a lot more slowly, and also the builds can't be done at all without live internet access.

vdo:
> I'd recommend, instead, adapting dockrun.sh to use apt-cacher-ng inside the gitrun container and pointing the gitian-target container at it. Without apt-cacher all of the other container builds go a lot more slowly, and also the builds can't be done at all without live internet access.

So your suggestion would be to link both containers, and then run apt-cacher-ng instead of an infinite sleep in the gitrun, right? This will also make the link more deterministic, rather than using the host's IP (some docker configurations may use a different subnet)

vdo:
> I'd recommend, instead, adapting dockrun.sh to use apt-cacher-ng inside the gitrun container and pointing the gitian-target container at it. Without apt-cacher all of the other container builds go a lot more slowly, and also the builds can't be done at all without live internet access.

I'm trying to figure out a way to link both containers, but looks like gitian does not allow extra args to the docker builder script (like `EXTRA_ARGS=--link gitrun:gitrun`), so it's not trivial to point the gitian target to the gitrun

See:
https://github.com/devrandom/gitian-builder/blob/46093250577a646d71b20bde4bc007b3c5dc0d21/libexec/start-target#L44

selsta:
`gitian` support has been removed from master and replaced with `guix`.

nahuhh:
@selsta do you mind if this is backported to release?

as is, i have to modify gitian (similar to above) on debian 12

PR #8110, opened 2021-12-12
HTTP wallet for monero-wallet-rpc

tevador:
### How to use

1. Run `./monero-wallet-rpc --testnet --rpc-bind-port=12345 --wallet-dir=/path/to/dir --rpc-login=test:test --http-wallet`
2. Open http://localhost:12345/wallet in your browser.
3. Enter the RPC username and password

The wallet offers only the most basic functionality at the moment: opening/creating a wallet, balance, sync progress, creating accounts/subaddresses, sending to a single destination and limited transaction history.

### Implementation details

Changes to existing code are minimal. One small change was needed in wallet RPC method `get_height`, which now also returns the daemon height.

The HTTP wallet itself consists of 5 static files:
* index.html
* styles.css
* wallet.js
* icon.png
* logo.png

These files are served under the virtual directory `/wallet`. The wallet must be explicitly enabled with the `--http-wallet` command line option and requires `--wallet-dir` to be set.

### Security

The wallet is sent to the browser with the `Content-Security-Policy: default-src 'self';` header. This means that the wallet can only comunicate with `monero-wallet-rpc` and should be otherwise completely isolated from the internet.


ACK-J:
To get this UI to work using a remote node use the following command
`./monero-wallet-rpc --testnet --rpc-bind-port=12345 --wallet-dir=/home/User/Desktop/Wallets --rpc-login=test:test --http-wallet --daemon-address testnet.xmr-tw.org:28081 --trusted-daemon`

And then open `http://127.0.0.1:12345/wallet` in a browser

ACK-J:
# A quick security assessment of the application
Some nit-picky small things
- Possible framable response (clickjacking) 
  - The server should return `X-Frame-Options: DENY` as a header

Slightly less nitpicky
- `Account labels` and `Sub address` labels are not HTML entity encoded 
- The content security policy stops execution. Im not aware of any way to bypass it at the moment but its best practice to encode user supplied input.
image
- HTML injection is possible and will execute
image

Overall, love this idea and great application! Thanks for taking the time to develop it.

PR #8114, opened 2021-12-14
multisig: fixes

perfect-daemon:


UkoeHB (approved):
This is a solid patch for the known vulnerabilities (to be disclosed publicly after this patch is merged). Thank you for all the hard work.

p.s. yes, I checked the diff against my local branch

src/common/varint.h
```diff
@@ -78,6 +78,22 @@ namespace tools {
...
   }
 
+  /*! \brief writes a varint to a fixed-size uchar array of size N.
+   */
+  template<typename T, std::size_t N>
+  void encode_varint(T t, unsigned char (&out)[N]) {
+    static_assert(std::is_integral<T>::value, "");
+    static_assert(std::is_unsigned<T>::value, "");
+    static_assert((sizeof(T) * 8 + 6) / 7 <= N, "");  //output array must be large enough to store any varint encoding of 't'
+    for (std::size_t i = 0; i < N && t; ++i) {
```
moneromooo-monero, line 88: This will write nothing if the value passed is 0. I don't think this is the intent. A unit test checking it writes the same as the existing varint writer (or that the reader gets back the same values) would find this.

src/multisig/signing_protocol.cpp
```diff
@@ -0,0 +1,967 @@
...
+    rct::keyV& s = rv.p.CLSAGs[i].s;
+    const rct::key& C_offset = rv.p.pseudoOuts[i];
+    rct::keyV P(ring_size);
+    rct::keyV C_nonzero(ring_size);
+
+    if (not reconstruction) {
+      s.resize(ring_size);
+      for (std::size_t j = 0; j < ring_size; ++j) {
+        if (j != l)
+          s[j] = rct::skGen();  //make fake responses
```
vtnerd, line 702: At a glance, this doesn't seem correct. In the source paper (Musig2), the randomly point (`R`) is chosen cooperatively so that secret shares never leak to the other participants. It seems like something similar would have to be done here, but I haven't gone through the math details to verify any potential issue.

Has the adaption of musig2 been reviewed by anyone else (besides me)?

src/multisig/signing_protocol.cpp
```diff
@@ -0,0 +1,967 @@
...
+  // compute initial challenge from real spend components
+  c = rct::hash_to_scalar(c_params);
+
+  // 1) c_0: find the CLSAG's challenge for index '0', which will be stored in the proof
+  // 2) c:   find the final challenge for the multisig signers to respond to
+  for (std::size_t i = (l + 1) % n; i != l; i = (i + 1) % n) {
+    if (i == 0)
+      c_0 = c;
+    rct::addKeys3(c_params[c_params_L_offset], s[i], G_precomp.k, c, W_precomp[i].k);
+    rct::addKeys3(c_params[c_params_R_offset], s[i], H_precomp[i].k, c, wH_l_precomp.k);
```
vtnerd, line 220: This is a follow-up to my other comment about the random `s` values being selected by a single participant. The participants providing randomized public keys affect the outcome of these iterative `c` hash values, but that output value was never controllable (that's the point of this hash function usage from the verifiers perspective). That should also mean that the randomized public keys are trivial in terms of the purpose with respect to the musig algorithm - it doesn't seem like they serve any useful purpose unless they affect the selection of the `s` values.

src/multisig/signing_protocol.cpp
```diff
@@ -0,0 +1,967 @@
...
+    a[num_sources - 1] = rct::zero();
+    for (std::size_t i = 0; i < num_destinations; ++i) {
+      sc_add(
+        a[num_sources - 1].bytes,
+        a[num_sources - 1].bytes,
+        output_amount_masks[i].bytes
+      );
+    }
+    for (std::size_t i = 0; i < num_sources - 1; ++i) {
+      rct::skGen(a[i]);
```
vtnerd, line 646: FWIW, I don't think the randomized values from all participants needs to influence this value, just the `s` values.

vtnerd (commented):
My review isn't quite finished, but I have a question that I felt needed to go out sooner rather than later. It may require someone else from the MRL to respond (either here or directly me to IRC). The adaption of the relevant paper doesn't seem accurate, but the adaption is necessarily a bit more complex due to CLSAG (when compared to schnorr), so its possible that I misunderstood some key detail.

src/multisig/signing_protocol.cpp
```diff
@@ -0,0 +1,967 @@
...
+  c_params_R_offset = c_params.size();
+  b_params_R_offset = b_params.size();
+  c_params.resize(c_params.size() + 1);  //this is where R will be inserted later
+  b_params.resize(b_params.size() + multisig::signing::kAlphaComponents);  //multisig aggregate public nonces for R will be inserted here later
+  b_params.emplace_back(I);
+  b_params.emplace_back(D);
+  b_params.insert(b_params.end(), s.begin(), s.begin() + l);    //fake responses before 'l'
+  b_params.insert(b_params.end(), s.begin() + l + 1, s.end());  //fake responses after 'l'
+  b_params.emplace_back();
+  tools::encode_varint(l, b_params.back().bytes);  //real signing index 'l'
```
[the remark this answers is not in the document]
UkoeHB: `b_params` is hashed by `key cn_fast_hash(const keyV &keys)`, so all elements need to be `rct::keys` (not ints).

src/multisig/signing_protocol.cpp
UkoeHB: This conditional is generating the fake CLSAG responses for all decoy ring members.

src/multisig/signing_protocol.cpp
UkoeHB: This code block is a bit hard to read due to the indexing. It's only using the decoy ring members' fake responses (not the responses from the multisig participants).

You can think of a ring signature as a fancy Schnorr signature, where the real signature challenge is a complicated hashing sequence of decoy members and fake responses. This code block is just computing the real signature challenge for multisig participants to respond to.

The line `c = rct::hash_to_scalar(c_params);` above this block is where the multisig participants' nonces are hashed (the real responses `s[l]` aren't defined in this function, they are defined way down in `finalize_tx()`).

src/multisig/signing_protocol.cpp
UkoeHB: This code block is only executed `if (not reconstruction)`, which is only true for the initial tx proposer.

src/multisig/signing_protocol.cpp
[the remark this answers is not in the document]
vtnerd: It doesn't matter - the bit encoding just needs to be same. Doing an endianess swap and `memcpy` into a `rct::key`.

src/multisig/signing_protocol.cpp
vtnerd: Yes, I know.

vtnerd: Sorry - this comment was (somewhat) pointless, I was stating that I think all participants need to alter `s` somehow, but I didn't think these randomly generated `a` values needed that.

src/multisig/signing_protocol.cpp
vtnerd: I had a reasonably good idea of what this was doing (as in, you just explained bits I already knew).

My question (below, this was mostly noise here as I was thinking outloud) was whether `s` was a set of nonce values, and since you did a review already, presumably your answer is that they are not nonce values.

The CLSAG paper doesn't list any values as nonce, only that `alpha` and a series of `s` values are selected randomly. At a glance this musig algorithm appears to be violating the paper, because randomness isn't provided/enforced for the `s` values (the initiator gets to select all of the values).

The "fake"/random `s` values are definitely just for symmetry to conceal the real `P` in use - more obvious after some fresh eyes (was staring at this too long earlier). You can't recover anything from a re-use in those values because the results are computed independently without any secret keys.

src/multisig/signing_protocol.cpp
vtnerd: I will probably finish up my review tomorrow, the hardest part is done (going through the relevant papers to track how it mapped to code, and changes necessary for the custom CLSAG scheme). So far it appears to be conforming to the two relevant papers; my review is "downhill" from here.

src/multisig/signing_protocol.cpp
UkoeHB: 'Nonce' is the same as `alpha`. IIRC the newer papers all use nonce, which imo is a better semantic than `alpha`.

UkoeHB: > The "fake"/random s values are definitely just for symmetry (logical "OR") to conceal the real P in use - more obvious after some fresh eyes (was staring at this too long earlier). You can't recover anything from a re-use in those values because they are used in equations isolated from any secret keys

Right, the real `s` is the real response, which is computed from a sum of multisig partial responses. The fake `s` are just random numbers/decoy responses. Do you still think these fake responses need to be computed from material provided by all multisig participants? Afaict, the only advantage this offers is if the tx proposer has weak RNG, then adding RNG from other participants will improve it (i.e. prevent observers from figuring out the real spend's index). If the tx proposer is malicious (or any of the signers), then adding more entropy from other participants is meaningless (the malicious individual can just publish the real index).

vtnerd: > Do you still think these fake responses need to be computed from material provided by all multisig participants?

I do not think the `s` values need to be included in the shared random computation.

> Afaict, the only advantage this offers is if the tx proposer has weak RNG, then adding RNG from other participants will improve it (i.e. prevent observers from figuring out the real spend's index). If the tx proposer is malicious (or any of the signers), then adding more entropy from other participants is meaningless (the malicious individual can just publish the real index).

The "penalty" for making the scheme easier appears to be a standard mechanism for publicly leaking the real spend. Since any of the signers can do this anyway, as you stated, this doesn't appear to be an issue worth fixing.

src/multisig/signing_protocol.cpp
vtnerd: I don't have anything else to add, but there are some comments that the OP should comment on or update.

UkoeHB: > but there are some comments that the OP should comment on or update.

Care to elaborate?

vtnerd: * An aliasing issue that can be trivial removed
* The new varint function isn't needed (chop it for maintenance purposes)
* A `scalarmult` can be easily removed
* A `std::move` can be added easily
* A simple `const` can be added

They are all simple changes, so it seemed easiest to make them now. Only the first two are useful before merge. The varint thing in particular would need to be chopped before a release, otherwise it becomes part of the protocol and must be maintained. The simple swap+`memcpy` is preferable to the maintenance of that specific function.

UkoeHB: Oh, I thought you meant code comments.

rbrunner7:
I did some functional tests on this, i.e. I took the code, compiled it and made multisig transactions in the normal way, to check whether something might be broken. I did not look at the code, and I did not try to check whether the security problems this code tries to solve are in fact solved. Still, I thought just checking whether anything still runs normally should be worthwhile.

Running the CLI wallet plus MMS on the branch I built new testnet multisig wallets for 2/2, 2/3 and 3/5 multisig, funded those and then made outgoing transactions. Especially the 3/5 case gave me the opportunity to do a row of transactions with varying signers.

I connected to a daemon running current *master* code.

Everything worked without any problem.

I could only test a single error condition: Does the wallet detect that a signer tries to sign a transaction a second time? Result: It does.

selsta:
@jonathancross this is fixed in master

PR #8117, opened 2021-12-22
cmake: enable build on riscv{64, 32}

Xeonacid:
tested on Arch Linux RISC-V.


selsta (approved):

CMakeLists.txt
```diff
@@ -325,6 +325,16 @@ if(ARCH_ID STREQUAL "s390x")
...
 endif()
 
+if(ARCH_ID STREQUAL "riscv64")
+set(RISCV   1)
+set(RISCV64 1)
+endif()
+
+if(ARCH_ID STREQUAL "riscv32")
+set(RISCV   1)
+set(RISCV32 1)
```
mj-xmr, line 335: Is this variable and its 64 variant used somewhere or did you add it for possible future use?

Xeonacid: @mj-xmr For possible future use. If you think they are now redundant, please tell me and I will remove them.

selsta: IMO it's fine to keep it

mj-xmr: I was asking mainly if I missed something in other part of the file.

It looks reasonable to have this.

mj-xmr (approved):

PR #8126, opened 2021-12-31
Use CMD in Dockerfile

KevinThomas0:
According to Dockerfile best practices
(https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint)
the best use for ENTRYPOINT is to set the image’s main command and then
use CMD as the default flags. This makes it easier to change the flags
passed to monerod when running the Docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,5 +61,6 @@ EXPOSE 18081
 # switch to user monero
 USER monero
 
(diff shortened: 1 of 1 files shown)


mj-xmr (approved):
Looks good. And thanks for the explanation.

PS. The parameters are intact (unchanged).

PR #8145, opened 2022-01-17
support authentication in monero-wallet-rpc set_daemon

### LINKED ISSUE #8143: Support setting daemon login credentials in monero-wallet-rpc `set_daemon`
This issue requests support to set daemon login credentials in monero-wallet-rpc's [`set_daemon`](https://github.com/monero-project/monero/blob/4f08d52dfacb62469b44c1ab68a37852011c8e3b/src/wallet/wallet_rpc_server.cpp#L4357) rpc call.

woodser:
Closes #8143

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -4426,7 +4426,11 @@ namespace tools
       return false;
     }
 
-    if (!m_wallet->set_daemon(req.address, boost::none, req.trusted, std::move(ssl_options)))
+    boost::optional<epee::net_utils::http::login> daemon_login{};
+    if (!req.username.empty() || !req.password.empty())
+      daemon_login.emplace(req.username, req.password);
+
+    if (!m_wallet->set_daemon(req.address, daemon_login, req.trusted, std::move(ssl_options)))
     {
       er.code = WALLET_RPC_ERROR_CODE_NO_DAEMON_CONNECTION;
       er.message = std::string("Unable to set daemon");
--- a/src/wallet/wallet_rpc_server_commands_defs.h
+++ b/src/wallet/wallet_rpc_server_commands_defs.h
@@ -47,7 +47,7 @@
 // advance which version they will stop working with
 // Don't go over 32767 for any of these
 #define WALLET_RPC_VERSION_MAJOR 1
-#define WALLET_RPC_VERSION_MINOR 23
+#define WALLET_RPC_VERSION_MINOR 24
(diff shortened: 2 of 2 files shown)


woodser:
Fixes https://github.com/monero-project/monero/issues/8143

selsta (commented):
Bump minor RPC version is missing.

also can you add the "Closes ..." message to the original PR message  so that it closes automatically?

woodser:
> Bump minor RPC version is missing.

Added it.

> also can you add the "Closes ..." message to the original PR message so that it closes automatically once it's merged?

Sure no problem.

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -47,7 +47,7 @@
 // advance which version they will stop working with
 // Don't go over 32767 for any of these
 #define WALLET_RPC_VERSION_MAJOR 1
-#define WALLET_RPC_VERSION_MINOR 23
+#define WALLET_RPC_VERSION_MINOR 24
```
vtnerd, line 50: Does this need to be bumped? If the client doesn't send user/pass, I think this just silently provides `boost::none` ?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -4426,7 +4426,11 @@ namespace tools
       return false;
     }
 
-    if (!m_wallet->set_daemon(req.address, boost::none, req.trusted, std::move(ssl_options)))
+    boost::optional<epee::net_utils::http::login> daemon_login{};
+    if (!req.username.empty())
```
vtnerd, line 4430: `if (!req.username.empty() || !req.password.empty())` - I think the empty string for username is valid in the HTTP login spec.

src/wallet/wallet_rpc_server_commands_defs.h
selsta: It might not be necessary technically but we usually always do it when editing that file.

src/wallet/wallet_rpc_server.cpp
woodser: Added `|| !req.password.empty()` and left the minor version at 24. Please let me know if should be changed back to 23.

selsta (approved):

src/wallet/wallet_rpc_server.cpp
vtnerd: Nope

vtnerd (approved):

PR #8061, opened 2021-11-15
Add view tags to outputs to reduce wallet scanning time

j-berman:
## Overview

Implements view tags as proposed by @UkoeHB in in MRL issue https://github.com/monero-project/research-lab/issues/73

At tx construction, the sender adds a 1-byte view tag to each output. The view tag is derived from the sender-receiver shared secret (view tag = first byte of hash(shared secret) . When scanning for outputs, the receiver can check the view tag for a match in order to reduce scanning time. When the view tag matches (expected match 1/256 outputs), the wallet proceeds to derive the output public key to check if the output belongs to the receiver. When the view tag does not match, the wallet avoids the more expensive EC operations when deriving the output public key.<sup>1</sup> In tests on my machine, I saw reductions in scanning time upwards of 40% (and a minimum around 30%).

Update: switched to using `cn_fast_hash` to hash the shared secret (edit: see discussion [below](https://github.com/monero-project/monero/pull/8061#issuecomment-970549760) for more on this choice)
~~As suggested by @UkoeHB, the shared secret is hashed using [SipHash](https://github.com/veorq/SipHash) 2-4 (the default), a keyed hash function that is specifically designed to protect a secret key and yield collisions, and is speedier than `cn_fast_hash`.~~

In order to maintain transaction uniformity, view tags are not supported until the fork height. At the fork height, all outputs *must* have a view tag.

<sup>1</sup> It's expected the skippable EC operations would be skipped for approximately 99.6% of outputs  (expected false positive rate = 1/2<sup>8</sup> = 1/256 = 0.4%, therefore expected negative rate = 100% - 0.4% = 99.6%). This is why 1 byte is the optimal sweet spot for max performance gains without needing more bytes. Skipping just 0.4% more outputs does not yield noticeable gains.

## Choices I made worth pointing out

For starters, I thought through various approaches deeply, and settled on what I feel is the sanest one. However, I'm absolutely prepared to refactor if necessary. I understand the various sections of the code that touch this fairly well enough at this point, would be happy to refactor if there is a stronger approach.

### Added a new `tx_out_to_tagged_key` boost variant type, to replace `tx_out_to_key` at the next fork height

This was likely the most significant choice I made. It seemed to be the most sensible way to fit view tags into the code without introducing overly cumbersome structural changes across the code. I think the biggest issue with this approach is in JSON parsing. Example: on the chain today (with no view tags), the `tx.vout` target serializes to JSON as follows:

```JSON
{ 
  "target": {
    "key": "ea3f..."
[...]

PR #8149, opened 2022-01-18
Multisig: signature fixes

UkoeHB:
Follow up to PR #8114. I lost contact with the original author two weeks ago.

This PR has the same commits as #8114, with [EDIT: additional commits to address reviewers, fix rebases, and general code quality].


vtnerd (commented):
Eh sorry, found one last thing to comment about with the commitment portion of musig. I don't think it matters, but I was curious about your thoughts (and I will probably think about it some more in the interim).

src/multisig/signing_protocol.cpp
```diff
@@ -0,0 +1,1039 @@
...
+    b_params_L_offset = b_params.size();
+    c_params.resize(c_params.size() + 1);  //this is where L will be inserted later
+    b_params.resize(b_params.size() + multisig::signing::kAlphaComponents);  //multisig aggregate public nonces for L will be inserted here later
+    c_params_R_offset = c_params.size();
+    b_params_R_offset = b_params.size();
+    c_params.resize(c_params.size() + 1);  //this is where R will be inserted later
+    b_params.resize(b_params.size() + multisig::signing::kAlphaComponents);  //multisig aggregate public nonces for R will be inserted here later
+    b_params.emplace_back(I);
+    b_params.emplace_back(D);
+    b_params.insert(b_params.end(), s.begin(), s.begin() + l);    //fake responses before 'l'
```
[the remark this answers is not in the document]
vtnerd: Or no wait, that value is garbage isn't it?

vtnerd (approved):
The remaining comments are trivial.

src/multisig/signing_protocol.cpp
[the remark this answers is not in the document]
vtnerd: Yup, definitely is. Brainfart since first review. I don't see how it could be recovered due to the nonces from each signer.

src/multisig/signing_protocol.cpp
```diff
@@ -0,0 +1,1039 @@
...
+  std::memcpy(tmp.bytes, str, N);
+  return tmp;
+}
+
+static void encode_int_to_key_le(const unsigned int &i, rct::key &k_out)
+{
+  std::uint64_t temp_i{SWAP64LE(i)};
+  static_assert(sizeof(std::uint64_t) <= sizeof(rct::key), "");
+  std::memcpy(k_out.bytes, &temp_i, sizeof(temp_i));
+  temp_i = 0;
```
UkoeHB, line 68: Btw @vtnerd won't the compiler just delete this line? I think I passed by reference so `i` could be memwipable.

vtnerd (approved):

src/multisig/signing_protocol.cpp
vtnerd: Yup. Its passed around in many functions, and its gotta be difficult already without many more changes.

Domol:
Just wanted to note here that I tested these changes:
- Generating new 2/3 multisig wallet,
- Creating transaction,
- Restoring from seed,
- Creating transaction with restored bit.

Probably not all possible cases, but still.

Thanks for fixing!

arnuschky:
I care about this getting merged. Is there anything I can do to help?

UkoeHB:
Thank you @Domol!

@arnuschky I'm not sure... there are some non-public review processes going on that I need to respect. It's mostly a waiting game I guess.

tmoravec:
> Note that this PR is not super useful without #7877.

Would you mind elaborating a bit, please? #7877 is a major change that updates the flows, changes the APIs, and so on. It requires heavy integration work. This one is closer to a bugfix that can more or less be merged. 

Is it possible to merge this one to get the security improvement, and leave #7877 for later, possibly several months later for example?

Thank you!

UkoeHB:
@tmoravec Without #7877, you cannot be confident that an M-of-N multisig wallet is actually an M-of-N multisig wallet. With key cancellation or an address hostage attack during the setup ceremony, it could be 1-of-N or (1+(M-1))-of-N, where the attacker always must participate in signing.

I'm pretty sure the only API/flow change with #7877 is changing `finalize_multisig()` to a no-op. All the other changes are opaque (unless you have some custom code to deserialize/examine multisig kex messages).

> Is it possible to merge this one to get the security improvement, and leave https://github.com/monero-project/monero/pull/7877 for later, possibly several months later for example?

Unfortunately it's pretty much out of my hands. Review doesn't happen instantaneously.

tmoravec:
@UkoeHB Fantastic, thanks for the explanation!

> Unfortunately it's pretty much out of my hands. Review doesn't happen instantaneously.

I meant merge to downstream libraries and products, not to monero core master. I would never dare to push anyone :) .

UkoeHB:
@tmoravec Yes, you are free to use any PR as you wish (assuming it isn't marked WIP or has an active discussion), with the caveat that an unmerged PR is probably a not-fully-reviewed PR.

tmoravec:
If it helps, I've backported the changes to `v0.17.3.0` and performed a basic smoke test:

* Create a wallet.
* Turn three wallets into 2/3 multisig.
* Receive and send transactions.

It works :) . Thanks for the fixes!

UkoeHB:
Note (todo): It will take a bit of work to rebase this PR onto #8061.

jonathancross:
Seems this needs a rebase?

UkoeHB:
rebased

UkoeHB:
rebase (onto BP+)

src/wallet/wallet2.cpp
```diff
@@ -13297,6 +13391,14 @@ cryptonote::blobdata wallet2::export_multisig()
     // if we have 2/4 wallet with signers: A, B, C, D and A is a transaction creator it will need to pick up 1 signer from 3 wallets left.
     // That means counting combinations for excluding 2-of-3 wallets (k = total signers count - threshold, n = total signers count - 1).
     size_t nlr = tools::combinations_count(m_multisig_signers.size() - m_multisig_threshold, m_multisig_signers.size() - 1);
+
+    // 'td.m_multisig_k' is an expansion of [{alpha_0, alpha_1, ...}, {alpha_0, alpha_1, ...}, {alpha_0, alpha_1, ...}],
+    //    - each '{alpha_0, alpha_1, ...}' tuple can be used for one tx attempt using this 'transfer' (i.e. output)
```
kayabaNerve, line 13396: This comment doesn't seem to have been properly communicated. Nonces shouldn't be reused at all, not just for this output, as despite having a unique private key per output the factors needed to undo the unique key's derivation are known by the other members of the signing group.

kayabaNerve (commented):
Read through the code, except the tests due to the other people saying this does work according to external expectations. The cryptography seems solid and well implemented to me, yet Ukoe definitely is the more informed here. I had one comment arguably more about ringct and then I think one comment is miswritten with what would be a critical security flaw, by my reading of it (yet not one that seems to be present in the code). Once that's clarified, I'd be happy to put my approval behind this :)

I can read/write C++, but I haven't contributed to Monero before so it's not my strongest suit. My only other concern would be if the TX handling code had some level of duplication which could be reduced (input sorting, yet it's a pretty trivial few lines...).

src/multisig/signing_protocol.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    const rct::key b = rct::hash_to_scalar(b_params);
+
+    // 1) store combined public nonces in the 'L' and 'R' slots for computing the initial challenge
+    //    - L = sum_i(b^i total_alpha_G[i])
+    //    - R = sum_i(b^i total_alpha_H[i])
+    // 2) compute the local signer's combined private nonce
+    //    - alpha_combined = sum_i(b^i * alpha[i])
+    rct::key& L_l = c_params[c_params_L_offset];
+    rct::key& R_l = c_params[c_params_R_offset];
+    rct::key b_i = rct::identity();
```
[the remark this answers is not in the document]
UkoeHB: I agree it's kinda janky, but imo isn't a problem to solve in this PR.

src/wallet/wallet2.cpp
UkoeHB: Ok updated the comments, let me know if it's more clear.

UkoeHB:
rebased

src/wallet/wallet2.cpp
kayabaNerve: LGTM

UkoeHB:
New commit: derive multisig tx secret keys from an entropy source (provided by the multisig tx proposer) plus the tx inputs' key images. This prevents the tx proposer from re-using tx secret keys that were used in some other transaction (that is in the ledger), which could allow them to burn funds (i.e. make duplicate onetime addresses appear on-chain).

Shout out to @kayabaNerve for identifying this issue.

src/multisig/signing_protocol.cpp
```diff
@@ -514,8 +580,35 @@ static bool compute_keys_for_destinations(
...
+  std::vector<crypto::secret_key> all_tx_secret_keys;
+  make_tx_secret_keys(tx_secret_key_seed, num_tx_keys, all_tx_secret_keys);
+
+  // split up tx secret keys
+  crypto::secret_key tx_secret_key_temp;
+  std::vector<crypto::secret_key> tx_aux_secret_keys_temp;
+  if (not collect_tx_secret_keys(all_tx_secret_keys, tx_secret_key_temp, tx_aux_secret_keys_temp))
+    return false;
+
+  if (reconstruction)
```
kayabaNerve, line 600: This if case shouldn't need to exist. We can not communicate the keys and simply always generate them here. That may not be a change worth it, as it's a minor optimization which changes data availability as this code is flowed through (and removes a sanity check while doing so), yet something to consider.

kayabaNerve (approved):
LGTM, barring one possible optimization which may not be worth it.

src/multisig/signing_protocol.cpp
UkoeHB: The keys have to be transmitted in the `pending_tx` structure. My first attempt at this just sent the entropy, but then I had test failures because there is a `wallet2::sanity_check()` that needs the tx secret keys.

kayabaNerve: ACK

src/wallet/wallet2.h
```diff
@@ -635,6 +648,9 @@ namespace tools
         FIELD(dests)
         FIELD(construction_data)
         FIELD(multisig_sigs)
+        if (version < 1)
+          return true;
+        FIELD(multisig_tx_key_entropy)
```
moneromooo-monero, line 653: This will not be initialized after loading a v0 record. Might be OK in practice, or setting to crypto::null_hash in the early out would be needed.

src/wallet/wallet2.cpp
```diff
@@ -7148,77 +7146,114 @@ bool wallet2::sign_multisig_tx(multisig_tx_set &exported_txs, std::vector<crypto
...
+        rct::keyM local_nonces_k(sd.selected_transfers.size(), rct::keyV(multisig::signing::kAlphaComponents));
         rct::key skey = rct::zero();
-        auto wiper = epee::misc_utils::create_scope_leave_handler([&](){ memwipe(k.data(), k.size() * sizeof(k[0])); memwipe(&skey, sizeof(skey)); });
-
-        for (size_t idx: sd.selected_transfers)
-          k.push_back(get_multisig_k(idx, sig.used_L));
+        auto wiper = epee::misc_utils::create_scope_leave_handler([&]{
+          for (auto& e: local_nonces_k)
+            memwipe(static_cast<rct::key *>(e.data()), e.size() * sizeof(rct::key));
+          memwipe(static_cast<rct::key *>(&skey), sizeof(rct::key));
```
moneromooo-monero, line 7203: Why is this cast necessary ?

src/wallet/wallet2.h
UkoeHB: Hmm ok, setting to `crypto::null_hash` then asserting it is not null in the multisig code would be good. By early out do you mean like this?
```
if (version < 1)
{
  multisig_tx_key_entropy = crypto::null_hash;
  return true;
}
FIELD(multisig_tx_key_entropy)
```

src/wallet/wallet2.cpp
UkoeHB: Don't think it is, will clean up.

src/wallet/wallet2.h
moneromooo-monero: Yes (might or might not be better to do the assignment if loading only, I don't really mind either way)

UkoeHB:
Squashed

UkoeHB:
Undid squash (some miscommunication sorry).

UkoeHB:
Another rebase

UkoeHB:
@tmoravec is BP+ active on your branch (i.e. is your wallet2 trying to make BP+ txs?)? This PR only works with BP+.

rbrunner7:
FWIW - multisig is a moving target for quite a while already:

Over the last few days I tested this as current master plus all commits in this PR, on testnet. I built 2/2, 2/3 and 3/5 wallets, funded, synced and then built out-transactions to spend some funds.

Everything worked flawlessly.

rbrunner7:
I tested the latest update today, building transactions for stagenet in 2/2 and 4/4 wallets.

Again no problems detected.

I also did a 2/2 testnet transaction with this code, making sure those did not break. Worked as well.

tmoravec:
Thanks @UkoeHB for the pre-BP+ patch! I've tested it and it works with my 2/3 wallet tests. Creating and funding wallets, sending transactions, etc. on stagenet.

src/multisig/multisig_clsag_context.cpp
```diff
@@ -0,0 +1,252 @@
...
+  initialized = true;
+  return true;
+}
+//----------------------------------------------------------------------------------------------------------------------
+bool CLSAG_context_t::combine_alpha_and_compute_challenge(
+  const rct::keyV& total_alpha_G,
+  const rct::keyV& total_alpha_H,
+  const rct::keyV& alpha,
+  rct::key& alpha_combined,
+  rct::key& c_0,
```
kayabaNerve, line 179: The CLSAG code calls this c1. While I honestly kind of prefer c0, and it turns out that's the term from the paper, the discrepancy isn't preferred. We also can't change it within the CLSAG code anymore, which is unfortunate.

src/multisig/multisig_tx_builder_ringct.cpp
```diff
@@ -0,0 +1,936 @@
...
+
+namespace multisig {
+
+namespace signing {
+//----------------------------------------------------------------------------------------------------------------------
+//----------------------------------------------------------------------------------------------------------------------
+bool view_tag_required(const int bp_version)
+{
+  // view tags were introduced at the same time as BP+, so they are needed after BP+ (v4 and later)
+  if (bp_version <= 3)
```
kayabaNerve, line 62: I'd hope we could utilize a proper enum here.

src/multisig/multisig_tx_builder_ringct.cpp
```diff
@@ -0,0 +1,936 @@
...
+    rct::keyV& s = rv.p.CLSAGs[i].s;
+    const rct::key& C_offset = rv.p.pseudoOuts[i];
+    rct::keyV P(ring_size);
+    rct::keyV C_nonzero(ring_size);
+
+    if (not reconstruction) {
+      s.resize(ring_size);
+      for (std::size_t j = 0; j < ring_size; ++j) {
+        if (j != l)
+          s[j] = rct::skGen();  //make fake responses
```
kayabaNerve, line 641: This can be converted into a RNG in a manner similar to tx_secret_key to reduce bandwidth and consolidate code between creator/reconstructors.

src/multisig/multisig_tx_builder_ringct.cpp
```diff
@@ -0,0 +1,936 @@
...
+//----------------------------------------------------------------------------------------------------------------------
+tx_builder_ringct_t::~tx_builder_ringct_t()
+{
+  memwipe(static_cast<rct::key *>(cached_w.data()), cached_w.size() * sizeof(rct::key));
+}
+//----------------------------------------------------------------------------------------------------------------------
+bool tx_builder_ringct_t::init(
+  const cryptonote::account_keys& account_keys,
+  const std::vector<std::uint8_t>& extra,
+  const std::uint64_t unlock_time,
```
kayabaNerve, line 723: While I understand why this is technically valid, the unlock time has minimal benefit and is largely considered extremely detrimental. Including it here in multisig does not seem beneficial.

src/wallet/wallet2.cpp
```diff
@@ -8949,10 +8985,8 @@ void wallet2::transfer_selected_rct(std::vector<cryptonote::tx_destination_entry
     src.real_output_in_tx_index = td.m_internal_output_index;
     src.mask = td.m_mask;
     if (m_multisig)
-    {
-      auto ignore_set = ignore_sets.empty() ? std::unordered_set<crypto::public_key>() : ignore_sets.front();
-      src.multisig_kLRki = get_multisig_composite_kLRki(idx, ignore_set, used_L, used_L);
-    }
+      //TODO: multisig_kLRki as used in tx_source_entry is just a key image shuttle into the multisig tx builder, need to simplify
```
kayabaNerve, line 8988: Left over TODO?

src/wallet/wallet2.cpp
```diff
@@ -9012,42 +9075,77 @@ void wallet2::transfer_selected_rct(std::vector<cryptonote::tx_destination_entry
...
+      // note: record the ignore set so when other signers go to add their signatures (sign_multisig_tx()), they
+      //       can skip this tx attempt if they aren't supposed to sign it; this only works for signers who provided
+      //       multisig_infos to the last 'import_multisig()' call by the local signer, all 'other signers' will encounter
+      //       a 'need to export multisig_info' wallet error if they try to sign this partial tx, which means if they want to sign a tx
+      //       they need to export_multisig() -> send to the local signer -> local signer calls import_multisig() with fresh
+      //       multisig_infos from all signers -> local signer makes completely new tx attempts (or a different signer makes tx attempts)
+      sig.ignore = ignore_sets[i];
+      sig.signing_keys = signing_keys;  //the local signer signed with ALL of their multisig key shares, record their pubkeys for reference by other signers
+    }
+    if (m_multisig_threshold <= 1) {
```
kayabaNerve, line 9142: Do I dare ask what 0 is?

kayabaNerve (commented):
None of these are required, yet there's a few questions/nice to haves.

I'm not the most familiar with wallet2 as a whole, leading to C++ omissions in my review such as not checking for proper usage of memwipe everywhere, yet the cryptography in this change set should be secure and implement a secure multisig signing process for Monero as a whole. User tests confirm it meets the expected UX, and nonces are explicitly removed from memory on first usage (as distinct from a previously noted work with Monero multisig).

src/multisig/multisig_clsag_context.cpp
```diff
@@ -0,0 +1,252 @@
...
+bool CLSAG_context_t::init(
+  const rct::keyV& P,
+  const rct::keyV& C_nonzero,
+  const rct::key& C_offset,
+  const rct::key& message,
+  const rct::key& I,
+  const rct::key& D,
+  const unsigned int l,
+  const rct::keyV& s,
+  const std::size_t num_alpha_components
```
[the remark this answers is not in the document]
UkoeHB: I agree it's a little overkill, although from a design standpoint it's nice to have a centralized config parameter instead of a hard-coded value.

src/multisig/multisig_clsag_context.cpp
```diff
@@ -0,0 +1,252 @@
...
+  c_params.insert(c_params.end(), C_nonzero.begin(), C_nonzero.end());
+  b_params.insert(b_params.end(), C_nonzero.begin(), C_nonzero.end());
+  c_params.emplace_back(C_offset);
+  b_params.emplace_back(C_offset);
+  c_params.emplace_back(message);
+  b_params.emplace_back(message);
+  c_params_L_offset = c_params.size();
+  b_params_L_offset = b_params.size();
+  c_params.resize(c_params.size() + 1);  //this is where L will be inserted later
+  b_params.resize(b_params.size() + num_alpha_components);  //multisig aggregate public nonces for L will be inserted here later
```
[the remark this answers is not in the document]
UkoeHB: Musig2 is simpler and there is no concrete reason to change the design at this point.

src/multisig/multisig_clsag_context.cpp
```diff
@@ -0,0 +1,252 @@
...
+  //    - alpha_combined = sum_i(b^i * alpha[i])
+  rct::key& L_l = c_params[c_params_L_offset];
+  rct::key& R_l = c_params[c_params_R_offset];
+  rct::key b_i = rct::identity();
+  L_l = rct::identity();
+  R_l = rct::identity();
+  alpha_combined = rct::zero();
+  for (std::size_t i = 0; i < num_alpha_components; ++i) {
+    rct::addKeys(L_l, L_l, rct::scalarmultKey(total_alpha_G[i], b_i));
+    rct::addKeys(R_l, R_l, rct::scalarmultKey(total_alpha_H[i], b_i));
```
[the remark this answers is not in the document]
UkoeHB: Technically more efficient yes, but negligible in practice. If this were verification code or if it were egregious then I'd change it. I generally prefer legibility over marginal efficiency gains in cases like this.

src/wallet/wallet2.h
```diff
@@ -594,13 +594,24 @@ namespace tools
...
+
       BEGIN_SERIALIZE_OBJECT()
-        VERSION_FIELD(0)
+        VERSION_FIELD(1)
         FIELD(sigs)
         FIELD(ignore)
         FIELD(used_L)
         FIELD(signing_keys)
         FIELD(msout)
+        if (version < 1)
```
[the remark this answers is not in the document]
UkoeHB: This is the standard pattern for updating serialization versions in this codebase (when the old version is assumed to be deprecated).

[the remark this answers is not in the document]
UkoeHB: Although it does make me wonder if this should return false instead, since old stuff won't work anyway.

src/multisig/multisig_clsag_context.cpp
UkoeHB: I agree this isn't ideal... however, `c_0` is set here:
```
for (std::size_t i = (l + 1) % n; i != l; i = (i + 1) % n) {
    if (i == 0)
      c_0 = c;
   ...
}
```
So I think `c_0` makes more sense than doubling down on `c1`.

I added a comment about the notation discrepancy.

src/multisig/multisig_tx_builder_ringct.cpp
UkoeHB: I agree you _can_ do this, and it does reduce bandwidth. I'm not sure it is useful in practice, because the payload size would only shrink by a few hundred bytes at most (in the typical case). The goal here is to get something that works correctly, rather than engineer it into the ideal state.

src/multisig/multisig_tx_builder_ringct.cpp
UkoeHB: The calling code (wallet2 tx builder) expects timelocks to work as intended by the user (who has to manually specify the unlock time). Deprecating timelocks is a bigger project that's out of scope here.

src/multisig/multisig_clsag_context.cpp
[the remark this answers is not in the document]
kayabaNerve: This has been resolved with the most recent edits :)

src/wallet/wallet2.h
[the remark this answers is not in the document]
kayabaNerve: Resolved

kayabaNerve (approved):
While I have nitpicks, my noted flaws (except the burning bug which was moved out of the PR) appear fixed, and Inference AG's appears competently responded to as well.

UkoeHB:
Commits squashed

kayabaNerve (approved):

PR #8158, opened 2022-01-25
Balance includes unconfirmed transfers to self

### LINKED ISSUE #8073: Add change output to wallet balance immediately
When a user spends XMR, their wallet spends outputs and creates a change output.

Currently, the wallet does not add the change output to the wallet's balance until confirmed.

As a result, the balance decreases by more than the amount spent, which is alarming to users.

This issue requests that the wallet adds the change output into its balance immediately so only the spend amount is deducted from the balance.

woodser:
Closes #8073

The wallet balance should correctly reflect unconfirmed transfers and change when used with #6986 and when strict mode is false.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -6126,6 +6126,19 @@ std::map<uint32_t, uint64_t> wallet2::balance_per_subaddress(uint32_t index_majo
         amount_per_subaddr[0] = utx.second.m_change;
       else
         found->second += utx.second.m_change;
+
+      // add transfers to same wallet
+      for (const auto &dest: utx.second.m_dests) {
+        auto index = get_subaddress_index(dest.addr);
+        if (index && (*index).major == index_major)
+        {
+          auto found = amount_per_subaddr.find((*index).minor);
+          if (found == amount_per_subaddr.end())
+            amount_per_subaddr[(*index).minor] = dest.amount;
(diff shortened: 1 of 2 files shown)


src/wallet/wallet2.cpp
```diff
@@ -6126,6 +6126,19 @@ std::map<uint32_t, uint64_t> wallet2::balance_per_subaddress(uint32_t index_majo
         amount_per_subaddr[0] = utx.second.m_change;
       else
         found->second += utx.second.m_change;
+
+      // add transfers to same wallet
+      for (const auto &dest: utx.second.m_dests) {
```
woodser, line 6131: Does anyone know a better way than using destinations?

It may not be ideal or as efficient to reverse lookup from a destination address. (?)

Also, destination addresses are only present in the wallet cache, but this change only affects unconfirmed funds so the balance becomes correct when the funds are confirmed.

moneromooo-monero:
Looks ok assuming the change output doesn't end up in m_dests - I assume you checked.

woodser:
I don't see any instances of the change amount being added to the destinations.

selsta (commented):
transfer tests fail

woodser:
> transfer tests fail

Passing now. Thanks.

j-berman (approved):
Tested in cli wallet

PR #8163, opened 2022-01-30
wallet inits cache if file and blob missing

woodser:


woodser:
Same as #8123 merged to master

selsta (approved):

PR #8165, opened 2022-01-31
[Feature Request] Give monerod node operators the ability to add a custom string to the get_info rpc call

bltshzzr:
Hi, a project I'm working on would benefit greatly from being able to verify that a pseudonymous entity is running a specific full node.

One way a node operator could verify ownership is by returning a secret in the public rpc call "get_info" using an environment variable (MONERO_NODE_INFO?) or an empty JSON object if that's not set. This would populate a key named "node_info" and could be used for all sorts of other things too.

I wasn't sure where the "framework" import comes from in the file `tests/functional_tests/daemon_info.py` so I couldn't verify that this patch passes that test; however, I did run curl against the rpc endpoint and confirmed that it is functional. Maybe someone can chime in with what to pip install to run that test.

Any and all feedback is much appreciated. Thanks!

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -534,6 +534,7 @@ namespace cryptonote
     res.version = restricted ? "" : MONERO_VERSION_FULL;
     res.synchronized = check_core_ready();
     res.busy_syncing = m_p2p.get_payload_object().is_busy_syncing();
+    res.node_info = getenv("MONERO_NODE_INFO") ? getenv("MONERO_NODE_INFO") : "{}";
 
     res.status = CORE_RPC_STATUS_OK;
     return true;
--- a/src/rpc/core_rpc_server_commands_defs.h
+++ b/src/rpc/core_rpc_server_commands_defs.h
@@ -687,6 +687,7 @@ namespace cryptonote
       uint64_t database_size;
       bool update_available;
       bool busy_syncing;
+      std::string node_info;
       std::string version;
       bool synchronized;
 
@@ -728,6 +729,7 @@ namespace cryptonote
         KV_SERIALIZE(database_size)
         KV_SERIALIZE(update_available)
         KV_SERIALIZE(busy_syncing)
+        KV_SERIALIZE(node_info)
         KV_SERIALIZE(version)
         KV_SERIALIZE(synchronized)
       END_KV_SERIALIZE_MAP()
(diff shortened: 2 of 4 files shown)


selsta:
In general having nodes return custom strings is something where we have to be careful with. I remember that Electrum used that to display fake update messages that linked to malicious binaries.

How would you use this feature?

bltshzzr:
I'm starting a project to encourage more groups to engage in the p2p nature of Monero and take responsibility
for their own privacy. My use case is verifying that a specific individual owns / operates a given public Monero node,
similar to using DNS records to prove authority to administer a domain (i.e. google search console, etc).

If we can use an existing feature or implement another way of having restricted rpc nodes "identify" themselves through some kind of public key system, I'd like to discuss it. Otherwise, I think that it's a very extensible mechanism that won't harm privacy by default and will enable valuable new uses of the daemon.

Due to the opt in nature of the change for users I'd say this proposal carries minimal risks to end user security. If wallets
want to make use of this new field I trust the community will ensure that we learn from the past's mistakes.

I am NOT a C++ developer though and am a little worried about potential security implications (buffer overflows?) of
assigning a string variable to the result of "std::getenv" with no preprocessing. Any thoughts there?

Thanks!

bltshzzr:
Update: a more direct answer to how I would use this feature.

```bash
$ MONERO_NODE_INFO="{\"CUSTOM_ORG_KEY\": \"0xDEADBEEF\"}" ./bin/monerod
```

```bash
curl http://127.0.0.1:18081/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"get_info"}' -H 'Content-Type: application/json'
```
```json
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    ...
    "node_info": "{\"CUSTOM_ORG_KEY\": \"0xDEADBEEF\"}",
  }
}
```

trasherdk:
You can use the existing `ssl`options, and verify a node's certificate.
```
  --rpc-ssl arg (=autodetect)           Enable SSL on RPC connections: 
  --rpc-ssl-private-key arg             Path to a PEM format private key
  --rpc-ssl-certificate arg             Path to a PEM format certificate
  --rpc-ssl-ca-certificates arg         Path to file containing concatenated 
  --rpc-ssl-allowed-fingerprints arg    List of certificate fingerprints to 
  --rpc-ssl-allow-chained               Allow user (via --rpc-ssl-certificates)
  --rpc-ssl-allow-any-cert              Allow any peer certificate
```

bltshzzr:
Sweet, thanks for the tip. I'll go ahead and close this PR since that handles my use case perfectly.

PR #8169, opened 2022-02-07
rpc: add explicit restricted flag to /get_info

tobtoht:


selsta (approved):

jeffro256:
This is a lot cleaner than asking for /update and returning false when it fails

jeffro256 (approved):
Tested with stagenet daemon

PR #8172, opened 2022-02-09
Remove unused SMTP code

jeffro256:
## Description:
I removed three files: contrib/epee/include/net/smtp*

## Motivation:
 It is unused throughout the entire project and hasn't been touched since 2014.

## Testing:
I have compiled on Ubuntu 20 without error

PR #8175, opened 2022-02-10
Epee Cleanup

jeffro256:
This will be my working draft of cleaning up the EPEE module. So far I have just removed unused files. This first commit is the combination #8172, #8173, #8174. Future commits will go into this PR.


tobtoht (approved):

jeffro256:
Okay, I'm not finished, but this PR is getting to be very large already.

paulshapiro:
aww how am i going to exfiltrate keys without our epee smtp support?

majestrate:
> aww how am i going to exfiltrate keys without our epee smtp support?

do not worry, there are plenty of places where a `std::string` is copy constructed from a `epee::wipeable_string` containing private key material, just use those instead. :>

tobtoht (approved):

jeffro256:
Comment related to last commit:

The last point is likely where something would go wrong if something did go wrong. Looking at the 2006 copyright in EPEE, I assume get_ns_count() was thrown together before std::chrono was a thing. I looked at each individual time implementation and they all seemed to their respective system's CPU timer. One of the big features of get_ns_count() that the author seems to be going for is monotonicty, which chrono::steady_clock provides. Also looking at get_ns_count()'s usage within the codebase, it is always used for timing quick things. So I gutted it and replaced it with steady_clock with a duration measured in std::nano. I tested std::chrono::steady_clock::period for Ubuntu 20 and Msys2-64bit and the period for both is in nanoseconds, which is a good thing. IIRC, that should also be the case in MacOS.

This change should improve timing precision, compile times, code readability, and reduce reliance on system-specific headers.

jeffro256:
The revised, easier to follow PR is here: #8211

moneromooo-monero:
> do not worry, there are plenty of places where a std::string is copy constructed from a epee::wipeable_string containing private key material, just use those instead. :>

If you see any that's not in RPC (or similar that we don't care about because they're going to network), list them so they can be fixed.

majestrate:
> I'll reopen that for a bit since majestrate might not be seeing this while it's closed. Again, point to those cases of wipeable_string that get copeid to non wipeable that aren't RPC or similar please.

here's one https://github.com/monero-project/monero/blob/master/src/wallet/wallet2.cpp#L4506

moneromooo-monero:
Thanks, please list them all if you have them.

moneromooo-monero:
Could you be more specific about this please ? It seems ok to me. To be clear, I'm not trying to wipe everything from the keys file, but two things: the spend secret key and the wallet password. Any settings etc in the keys file are uninteresting. AFAICT, this function does not leak either.

majestrate:
> Could you be more specific about this please ? It seems ok to me. To be clear, I'm not trying to wipe everything from the keys file, but two things: the spend secret key and the wallet password. Any settings etc in the keys file are uninteresting. AFAICT, this function does not leak either.

I really don't see how my tongue a cheek jab needs to hold up the removal of 10K lines of dead code.

moneromooo-monero:
So... trolling ? Were you just wasting my time here ?

moneromooo-monero:
Well, closing again then. If there are actual bugs to do with this or something else, file a new bug.

majestrate:
> So... trolling ? Were you just wasting my time here ?

oh no not at all, i was just excited to potentially see monero catch up in the dead code removal department. should not have gotten my hopes up either way.

selsta:
@majestrate https://github.com/monero-project/monero/pull/8211 is the cleanup PR..

moneromooo-monero:
Ooooh I get it. My comments were in no way preventing this particular PR from being merged. I just asked about the leaks here because that's where you commented. They are unrelated issues :)

PR #8177, opened 2022-02-11
simplewallet: rename duplicate amount headers for clarity

### LINKED ISSUE #8171: "Export_Transfers All" contains a duplicate attribute
After running `export_transfers all` within the `monero-wallet-cli`, a csv is written that contains the following header:

block, direction, unlocked, timestamp, **amount**, running balance, hash, payment ID, fee, destination, **amount**, index, note

Notice that the attribute **amount** is accidentally used twice ( 5th and 11th), I have confirmed that the two columns contain redundant data and is simply a duplicate

reemuru:
Closes #8171 

There are duplicate headers for 'amount' when performing export_transfers
all. Column five represents the total amount of the transaction while
column eleven represents the amount transferred per destination (e.g.
multi-destination transactions). The columns have been renamed to
'transaction amount' and 'destination amount` respectively.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -8818,7 +8818,7 @@ bool simple_wallet::export_transfers(const std::vector<std::string>& args_)
   // header
   file <<
       boost::format("%8.8s,%9.9s,%8.8s,%25.25s,%20.20s,%20.20s,%64.64s,%16.16s,%14.14s,%106.106s,%20.20s,%s,%s") %
-      tr("block") % tr("direction") % tr("unlocked") % tr("timestamp") % tr("amount") % tr("running balance") % tr("hash") % tr("payment ID") % tr("fee") % tr("destination") % tr("amount") % tr("index") % tr("note")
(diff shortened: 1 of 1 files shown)


tobtoht:
The first amount column is the total amount transferred in the transaction. The second amount column is the amount transferred to a destination address. The values will differ for multi-destination transactions. Perhaps this distinction could be named more clearly.

reemuru:
@tobtoht thanks for the clarification! I see the diff now on multi-destination transfers. How about `transaction amount` and `destination amount` respectively? Squash in on [30ab49d](https://github.com/monero-project/monero/pull/8177/commits/30ab49d5cb5f7d5b089d20919cf4f02b063f2e6b)

tobtoht (approved):
Looks good to me.

ACK-J:
Other issues I've been having with the exporter include:
- There is an abundance of excess spaces between each comma-separated value, for example here is the heading once exported.
```
   block,direction,unlocked,                timestamp,              amount,     running balance,                                                            hash,      payment ID,           fee,                                                                                               destination,              amount,index,note

```
- The column named `hash` is not very descriptive and should probably be changed to `transaction hash`
- There is no `output public key` column, this is a very relevant piece of information that I am trying to use in my project

I appreciate all the hard work you guys put into the repo!

PR #8178, opened 2022-02-12
Bump ring size to 16 for v15

j-berman:
As per:
- https://github.com/monero-project/research-lab/issues/79
- https://github.com/monero-project/meta/issues/655

-----

This PR follows the "double fork" pattern to allow ring sizes of either 11 or 16 during a grace period, and then requires 16. This way users of updated wallets who submit tx's right before the fork height that make it into the pool will still have their tx's processed.

Also removed the obsolete command to allow users to set a default ring size in the client. Updated wallets will start to use ring size 16 at the fork height.

------

I'm currently syncing the chain from scratch with this code as a sanity check. Currently on block ~2.3mn, which is after the latest fork. Will update here when finished. Update: synced from scratch successfully.

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -3309,7 +3309,7 @@ bool Blockchain::check_tx_inputs(transaction& tx, tx_verification_context &tvc, 
     size_t n_unmixable = 0, n_mixable = 0;
     size_t min_actual_mixin = std::numeric_limits<size_t>::max();
     size_t max_actual_mixin = 0;
-    const size_t min_mixin = hf_version >= HF_VERSION_MIN_MIXIN_10 ? 10 : hf_version >= HF_VERSION_MIN_MIXIN_6 ? 6 : hf_version >= HF_VERSION_MIN_MIXIN_4 ? 4 : 2;
+    const size_t min_mixin = hf_version >= HF_VERSION_MIN_MIXIN_15 ? 15 : hf_version >= HF_VERSION_MIN_MIXIN_10 ? 10 : hf_version >= HF_VERSION_MIN_MIXIN_6 ? 6 : hf_version >= HF_VERSION_MIN_MIXIN_4 ? 4 : 2;
     for (const auto& txin : tx.vin)
     {
       // non txin_to_key inputs will be rejected below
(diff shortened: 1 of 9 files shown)


rbrunner7:
I tried to review but bumped against a partial lack of knowledge:

Is there an explanation somewhere about the meaning of *unmixable* and how a tx looks if I build it today that has such unmixable outputs? If not I would be glad for some short explanations; I tried to get the "mechanics" of unmixable / pre-RingCT / post-RingCT txs from the code alone but could not break through.

My current understanding that has still some holes in it sees errors both in the "old" now-active code as well as in your new code, but that does not seem very probable ...

j-berman:
> Is there an explanation somewhere about the meaning of unmixable and how a tx looks if I build it today that has such unmixable outputs?

Before RingCT, transparent amount outputs could only mix with other outputs of the same amount. Let's say back in the pre-RingCT days, I sent you an output worth 0.0004 Monero, and no other 0.0004 amount outputs ever came into existence in the chain (hypothetical). This 0.0004 Monero output is thus "unmixable", since there aren't 10 other decoy outputs of the same amount with which you can mix this output with today. In this case, consensus allows you to construct a pre-RingCT transaction with a ring size of 1, where the 0.0004 amount output is used as an input, and pre-RingCT outputs come out on the other end. 

To simulate this, you can set up a test network where you mine a bunch of pre-RingCT outputs, then advance the hard fork to current, then call `sweep_unmixable` in `monero-wallet-cli` and check the tx

tobtoht (approved):
Reviewed.

rbrunner7 (approved):

src/cryptonote_core/blockchain.cpp
```diff
@@ -3321,14 +3321,12 @@ bool Blockchain::check_tx_inputs(transaction& tx, tx_verification_context &tvc,
...
-      MERROR_VER("Tx " << get_transaction_hash(tx) << " has invalid ring size (" << (min_actual_mixin + 1) << "), it should be 11");
-      tvc.m_low_mixin = true;
-      return false;
-    }
-
-    if (min_actual_mixin < min_mixin)
+    // The only circumstance where ring sizes less than expected are
+    // allowed is when spending unmixable non-RCT outputs in the chain.
+    // Caveat: at HF_VERSION_MIN_MIXIN_15, temporarly allow ring sizes
+    // of 10 are to allow a grace period in the transition to larger
```
moneromooo-monero, line 3327: 11

moneromooo-monero (commented):
The simplewallet default ring change has nothing to do with the ring size consensus change, and should not be in the same patch at all.
The other patch is pretty much pointless, but doesn't hurt I guess.
The rest looks good.

j-berman:
@moneromooo-monero  depending on the value of `DEFAULT_MIX`, simplewallet wouldn't be able to construct tx's (either before or after the fork) if it's not updated to use `m_wallet->min_ring_size()` instead of `DEFAULT_MIX` (it would fail in [this if statement](https://github.com/monero-project/monero/blob/70ceab6c104d5792a7dbbe96efbcc9efcb51c822/src/simplewallet/simplewallet.cpp#L6533-L6542)). Given that, does the change to use `min_ring_size()` still belong in a different PR or did I misread that suggestion?

Also, instead of getting rid of the command to set the default ring size in this PR, what if I just change the logic of `set_default_ring_size` to stop using the constants `DEFAULT_MIX` and `MIN_RING_SIZE` and always call `m_wallet->min_ring_size()`?

AFAICT, this ring size bump is the first one happening *after* the fixed ring size is already required, which is why I can't update the constants to the higher ring size in simplewallet (or leave them as is) otherwise it behaves incorrectly. That's why usage of the constants in simplewallet seemed worth updating in this PR (and I figured might as well just remove the command to set the default ring size instead of messing with that logic)

moneromooo-monero:
I meant that two very different things should be in different patches. I'm fine with the change itself. But I see your point now, I guess it's ok since there is a good reason, which I did not realize before.

rbrunner7 (approved):
Checked / reviewed again after rebase

Issue #8120, opened 2021-12-23
load of value, which is not a valid value for type 'bool'  discovered by LibFuzzer


swirsz:
2 similar bugs detected via Libfuzzer & UndefinedBehaviorSanitizer.

./levin_fuzz_tests crash-1a5

[crash-1a5.txt](https://github.com/monero-project/monero/files/7766778/crash-1a5.txt)

/src/monero/monero/contrib/epee/include/net/levin_protocol_handler_async.h:543:31: runtime error: load of value 247, which is not a valid value for type 'bool'
SUMMARY: UndefinedBehaviorSanitizer: 

./load-from-binary_fuzz_tests crash-f31

[crash-f31.txt](https://github.com/monero-project/monero/files/7766781/crash-f31.txt)

/src/monero/monero/contrib/epee/include/storages/portable_storage_from_bin.h:157:17: runtime error: load of value 249, which is not a valid value for type 'bool'
SUMMARY: UndefinedBehaviorSanitizer

moneromooo-monero:
Those seem easy to fix with !!. Thanks.

selsta:
Fixed in #8186

PR #8186, opened 2022-02-19
Undefined behaviour fixes

SChernykh:
Fixes issues reported in #8120


moneromooo-monero (approved):

jeffro256:
The longer you look into bools in C++, the weirder they get.

Issue #8168, opened 2022-02-05
Don't require --rpc-login with --rpc-access-control-origins


woodser:
This issue requests removing the requirement to use RPC authentication when starting monerod with `--rpc-access-control-origins`.

That way, public nodes may serve browser clients.

jeffro256:
This would be great! I've run into this issue myself

jeffro256:
Tangentially related issue: #7951

jeffro256:
@woodser check out the above PR. There may be some security implications with allowing --rpc-access-control-origins with public nodes, but I think that would mostly be a concern for the browser applications, not the daemon operators. In that case, the incumbent security concerns of using an untrusted node overshadow what issues --rpc-access-control-origins bring into the equation,

woodser:
Great. The PR should be opened on release-0.17 to be incorporated into the next release, too.

jeffro256:
@woodser done: #8227

PR #8187, opened 2022-02-19
Fixes: Don't require --rpc-login with --rpc-access-control-origins

jeffro256:
This PR removes the requirement for --rpc-login to be specified if --rpc-access-control-origins is.
This will allow public nodes to serve cross-origin requests. You can still use --rpc-login with
--rpc-access-control-origins, but it is no longer mandatory.

Original Issue: #8168


reemuru (commented):
Tested.

selsta (approved):

PR #8195, opened 2022-02-28
wallet2: update stagenet rollback blocks

selsta:


rbrunner7 (approved):
The extension from testnet correction to testnet + stagenet correction looks correct to me. I trust your estimate of 30,000 stagenet blocks, did not actually check that myself ...

PR #8200, opened 2022-03-02
Compil. time: http_client.h -> abstract_http_client.h

mj-xmr:
## Result
Compilation time reduced by -10% .

## Description

Below is the complexity of the `http_client` class:

image

It's a class derived from `epee::net_utils::http::abstract_http_client` and could be pointed at using this base class' pointer.

This is how the `http_client`'s dependees tree looks like on master:
image

and this is how I was able to reduce it:
image

According to my [compilation time report](http://cryptog.hopto.org/monero/health/data/9aab19f34/9aab19f34-cba-result.txt), on master, the header `net/http_client.h` takes over a minute to compile:

```
78479 ms: /tmp/monero/monero-clean/external/easylogging++/easylogging++.h (included 226 times, avg 347 ms), included via:
(...)
68317 ms: /tmp/monero/monero-clean/contrib/epee/include/net/http_client.h (included 38 times, avg 1797 ms), included via:
(...)
64580 ms: /tmp/monero/monero-clean/src/cryptonote_basic/difficulty.h (included 89 times, avg 725 ms), included via:
(...)
```

A side effect was, that I had to add the following 3 pure virtual methods to `epee::net_utils::http::abstract_http_client`:
```c++
    virtual const std::string &get_host() const = 0;
    virtual const std::string &get_port() const = 0;
    virtual void wipe_response() = 0;
```

## Measurements
using `time make` on master and on the branch:
| real	49m0,890s |     real	44m16,309s  |  -10.06% | 

Of course, individual verification is welcome.

### DIFF
--- a/src/device_trezor/CMakeLists.txt
+++ b/src/device_trezor/CMakeLists.txt
@@ -88,6 +88,7 @@ if(DEVICE_TREZOR_READY)
             ringct_basic
             cryptonote_core
             common
+            net
             ${SODIUM_LIBRARY}
             ${Boost_CHRONO_LIBRARY}
             ${Protobuf_LIBRARY}
--- a/src/net/http.h
+++ b/src/net/http.h
@@ -29,18 +29,12 @@
 
 #pragma once
 
-#include "net/http_client.h"
+#include "net/abstract_http_client.h"
 
 namespace net
 {
 namespace http
 {
-
-class client : public epee::net_utils::http::http_simple_client
-{
-public:
-  bool set_proxy(const std::string &address) override;
-};
 
 class client_factory : public epee::net_utils::http::http_client_factory
 {
--- a/.github/workflows/build-time-test.yml
+++ b/.github/workflows/build-time-test.yml
@@ -36,7 +36,6 @@ jobs:
       run: |
         ${{env.CCACHE_SETTINGS}}
         ${{env.CLONE_PATCHES}}
-        ${{env.TEST_BUILD_TIME}}
 
 # See the OS labels and monitor deprecations here:
 # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
--- a/.github/workflows/build-time-test.yml
+++ b/.github/workflows/build-time-test.yml
@@ -19,7 +19,7 @@ env:
         ccache --max-size=150M
         ccache --set-config=compression=true
   CLONE_PATCHES: git clone --recursive https://github.com/mj-xmr/monero-patches.git
-  TEST_BUILD_TIME: time ./monero-patches/src/monero-test-build-time.sh compil-epee-http-client wallet/api/wallet_api
+  TEST_BUILD_TIME: time ./monero-patches/src/monero-test-build-time.sh compil-epee-http-client src/wallet/api
 
 jobs:
   time-macos:
--- a/.github/workflows/build-time-test.yml
+++ b/.github/workflows/build-time-test.yml
@@ -1,0 +1,88 @@
+name: ci/gh-actions/build-time-test
+
+on:
+  push:
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '**/README.md'
+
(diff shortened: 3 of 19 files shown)


src/device_trezor/CMakeLists.txt
```diff
@@ -88,6 +88,7 @@ if(DEVICE_TREZOR_READY)
             ringct_basic
             cryptonote_core
             common
+            net
```
mj-xmr, line 91: device_trezor needs the client factory from `net`

src/net/http.h
```diff
@@ -28,19 +28,13 @@
...
 {
 namespace http
 {
 
-class client : public epee::net_utils::http::http_simple_client
-{
-public:
-  bool set_proxy(const std::string &address) override;
-};
-
```
mj-xmr, line 43: Encapsulated in `http.cpp`

reemuru (commented):
@mj-xmr compilation tested. Are there many more of these time optimizations to be found, like a to-do list? 

`time make -j$(nproc)`
|                   | master | [6fc9dba](https://github.com/monero-project/monero/pull/8200/commits/6fc9dbaba77c8a65e8e8fd5587970c3462e3391b) | change |
| -------------- | -------------------------------- | ----------------------- | ---------------  |
| real            |  31m16.651s                  | 28m52.107s         | ~8%             |

mj-xmr:
Thanks for testing, @reemuru .
Yes, there's a lot of potential here, which you can see in the [CBTA report](https://github.com/monero-project/monero/pull/7028) . The Doxygen report to the right helps in devising an appropriate strategy.

mj-xmr:
As a side note, single threaded tests are preferred, due to multithreaded tests suffering from thread starvation.

mj-xmr:
@reemuru The correct link to the report [is here](http://cryptog.hopto.org/monero/health/).

reemuru:
> @reemuru The correct link to the report [is here](http://cryptog.hopto.org/monero/health/).

@mj-xmr thanks for the info and report. I will keep a mental note for single-threaded tests.

selsta:
master
```
Executed in   17.34 mins    fish           external
   usr time   16.40 mins    0.14 millis   16.40 mins
   sys time    0.86 mins    1.41 millis    0.86 mins
```

patch
```
Executed in   17.00 mins    fish           external
   usr time   16.10 mins    0.27 millis   16.10 mins
   sys time    0.83 mins    3.83 millis    0.83 mins
```

Seeing as this is a small diff and other have higher gains this looks good to merge once reviewed.

mj-xmr:
@selsta I'd hate to argue all the time about this, but it seems, that you might be hitting some I/O bottlenecks, since the differences on your machine are always smaller.

jeffro256:
Here are my results:

    master
    real	63m45.448s
    user	60m5.132s
    sys	3m28.104s

    compil-epee-http-client
    real	61m43.451s
    user	58m18.784s
    sys	3m22.170s

For each run, I ran:
```bash
rm -rf build/
ccache --clear
time make
```
So I had a (3825s - 3703s) / (3825s) * 100% = 3.190% spedup

mj-xmr:
@jeffro256 Thanks for the test. I should have said it beforehand, but you didn't need to sacrifice your ccache for me though :) 

Here's a softer and more objective way to test the compilation time alone:
```
mkdir -p build/time && cd build/time
git checkout master
cmake ../.. -DUSE_CCACHE=OFF -DBUILD_TESTS=ON -DBUILD_SHARED_LIBS=ON && make clean && time make > /dev/null
git checkout TESTED_BRANCH_NAME
cmake ../.. -DUSE_CCACHE=OFF -DBUILD_TESTS=ON -DBUILD_SHARED_LIBS=ON && make clean && time make > /dev/null
```
On top of that, to reduce the effects of I/O bottlenecks, my build directory, as well as source are both placed in a RAMDisk.

mj-xmr:
I'll be doing a couple small tests here, publish the results, and then you decide what to do with this PR.

PR #8201, opened 2022-03-02
Add Ninja to list of Mingw dependencies in README.md

jeffro256:
Upon trying to build monero from scratch on Msys2, I ran into the following error:

```
CMake Error: CMake was unable to find a build program corresponding to "Ninja". CMAKE_MAKE_PROGRAM is not set. You probably need to select a different build tool
```

Installing the packages in question resolved that issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,13 +403,13 @@ application.
     To build for 64-bit Windows:
 
     ```bash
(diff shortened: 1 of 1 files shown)


jeffro256:
This fixed my issue but there may be something deeper here. On matrix, @mj-xmr said "This is weird and means, that CMake resorts to default to Ninja for some reason, which is wrong."

selsta:
What are the exact commands you used to build? CMake defaults to make.

jeffro256:
@selsta I did `make -j4 release-static-win64`

mj-xmr:
Since I know you're under msys: how about simply: `make`?

jeffro256:
@mj-xmr  I couldn't reproduce it unfortunately. If I remove the make package then type `cmake .` then I get the same error.

So if someone else has a similar error, I would try reinstalling make and the mingw toolchain

PR #8203, opened 2022-03-03
multisig: add key exchange round booster

UkoeHB:
This PR adds a small optimization to multisig account creation that is especially useful for 2-of-3 groups (disclaimer: there is a bounty for it [here](https://github.com/haveno-dex/haveno/issues/86)).

Multisig account creation is a series of rounds. Each round consists of sending messages to all the other group members. A member cannot complete a round until they have received a message for that round from each other member.

A standard 2-of-3 setup ceremony looks like this:
**Round 1**: Members A, B, C send initial messages to all other members.
**Round 2**: Members A, B, C use the previous round's messages to make their personal shares of the final key. They send a second message to all other members.
**Finalization**: Once a member has a full set of round 2 messages, they can finish the account.

That setup ceremony requires 2 communication rounds. For 2-of-3 escrowed markets, the buyer party should instead be able to complete the ceremony in one step so they can fund the multisig wallet right away.

With boosting, we can instead have this setup ceremony:
1. Members A, B send initial messages to each other and to member C.
2. Members A, B create 'round 2 booster' messages and send them to member C.
3. Member C completes round 1 immediately. He creates a round 1 and a round 2 message and sends them to members A, B. Member C then uses the booster messages to complete round 2, finalizing his local multisig account in the same step.
4. Members A, B complete round 1 then make round 2 messages and send them to each other.
5. Members A, B complete round 2 and finalize the multisig account.

So, by adding rounds to members A and B, member C can complete a 2-of-3 multisig account in one step.

**DISCLAIMER**: Kex boosting is expected to be used in conjunction with account force-updating. Extreme care must be taken when designing a system that uses these features to avoid attack vectors that lead to funds getting stuck or possibly stolen (I'd hesitate to use it at all until I or someone else does a write-up on how to do it safely).

#### TODO

- Still need to update `simplewallet` (and the `mms` presumably @rbrunner7 ).
- This PR is missing functional tests for the expanded wallet2 api (`tests/functional_tests/multisig.py`). It would be cool if someone who likes Python could help me with this.

#### Future Work

- Try to refactor `wallet2` so creating a multisig wallet doesn't involve converting an existing normal wallet's keys. Once a multisig wallet has been created, there should be no way to 'recreate' it using a different M/N configuration. Without that refactor, a user could boost a multisig account using one M/N configuration with a normal account (before converting their normal account to multisig), then later on when they actually convert to a multisig account use a different M/N configuration. If either of those M/N configs was an (N-1)-of-N, then the user will leak their multisig private key shares for that setup when setting up the _other_ configuration (via the key exchange ceremony).

### DIFF
--- a/src/multisig/multisig_account_kex_impl.cpp
+++ b/src/multisig/multisig_account_kex_impl.cpp
@@ -567,7 +567,7 @@ namespace multisig
 
     // note: do NOT remove the local signer from the pubkey origins map, since the post-kex round can be force-updated with
     //       just the local signer's post-kex message (if the local signer were removed, then the post-kex message's pubkeys
-    //       would be completely lost)
+    //       would be completely deleted)
 
     // evaluate pubkeys collected
 
@@ -608,7 +608,7 @@ namespace multisig
   * INTERNAL
   *
   * brief: multisig_kex_process_round_msgs - Process kex messages for the active kex round.
-  *    - A wrapper around evaluate_multisig_kex_round_msgs() -> multisig_kex_make_next_msg().
+  *    - A wrapper around evaluate_multisig_kex_round_msgs() -> multisig_kex_make_round_keys().
   *      - In other words, evaluate the input messages and try to make a message for the next round.
   *    - Note: Must be called on the final round's msgs to evaluate the final key components
   *            recommended by other participants.
@@ -623,7 +623,7 @@ namespace multisig
   * outparam: keys_to_origins_map_out - map between round keys and identity keys
   *    - If in the final round, these are key shares recommended by other signers for the final aggregate key.
   *    - Otherwise, these are the local account's DH derivations for the next round.
-  *      - See multisig_kex_make_next_msg() for an explanation.
+  *      - See multisig_kex_make_round_keys() for an explanation.
   * return: multisig kex message for next round, or empty message if 'current_round' is the final round
   */
   //----------------------------------------------------------------------------------------------------------------------
@@ -684,59 +684,67 @@ namespace multisig
   //----------------------------------------------------------------------------------------------------------------------
   // multisig_account: INTERNAL
   //----------------------------------------------------------------------------------------------------------------------
-  void multisig_account::initialize_kex_update(const std::vector<multisig_kex_msg> &expanded_msgs,
-    const std::uint32_t kex_rounds_required,
-    std::vector<crypto::public_key> &exclude_pubkeys_out)
+  std::vector<crypto::public_key> multisig_account::get_kex_exclude_pubkeys() const
   {
+    // exclude all keys the local account recommends
+    std::vector<crypto::public_key> exclude_pubkeys;
+
     if (m_kex_rounds_complete == 0)
     {
-      // the first round of kex msgs will contain each participant's base pubkeys and ancillary privkeys
-
-      // collect participants' base common privkey shares
-      // note: duplicate privkeys are acceptable, and duplicates due to duplicate signers
-      //       will be blocked by duplicate-signer errors after this function is called
-      std::vector<crypto::secret_key> participant_base_common_privkeys;
-      participant_base_common_privkeys.reserve(expanded_msgs.size() + 1);
-
-      // add local ancillary base privkey
-      participant_base_common_privkeys.emplace_back(m_base_common_privkey);
-
-      // add other signers' base common privkeys
-      for (const auto &expanded_msg : expanded_msgs)
-      {
-        if (expanded_msg.get_signing_pubkey() != m_base_pubkey)
-        {
-          participant_base_common_privkeys.emplace_back(expanded_msg.get_msg_privkey());
-        }
-      }
-
-      // make common privkey
-      make_multisig_common_privkey(std::move(participant_base_common_privkeys), m_common_privkey);
-
-      // set common pubkey
-      CHECK_AND_ASSERT_THROW_MES(crypto::secret_key_to_public_key(m_common_privkey, m_common_pubkey),
-        "Failed to derive public key");
-
-      // if N-of-N, then the base privkey will be used directly to make the account's share of the final key
-      if (kex_rounds_required == 1)
-      {
-        m_multisig_privkeys.clear();
-        m_multisig_privkeys.emplace_back(m_base_privkey);
-      }
-
-      // exclude all keys the local account recommends
-      // - in the first round, only the local pubkey is recommended by the local signer
-      exclude_pubkeys_out.emplace_back(m_base_pubkey);
+      // in the first round, only the local pubkey is recommended by the local signer
+      exclude_pubkeys.emplace_back(m_base_pubkey);
     }
     else
     {
-      // in other rounds, kex msgs will contain participants' shared keys
-
-      // ignore shared keys the account helped create for this round
+      // in other rounds, kex msgs will contain participants' shared keys, so ignore shared keys the account helped
+      //     create for this round
       for (const auto &shared_key_with_origins : m_kex_keys_to_origins_map)
-      {
-        exclude_pubkeys_out.emplace_back(shared_key_with_origins.first);
-      }
+        exclude_pubkeys.emplace_back(shared_key_with_origins.first);
+    }
+
+    return exclude_pubkeys;
+  }
+  //----------------------------------------------------------------------------------------------------------------------
+  // multisig_account: INTERNAL
+  //----------------------------------------------------------------------------------------------------------------------
+  void multisig_account::initialize_kex_update(const std::vector<multisig_kex_msg> &expanded_msgs,
+    const std::uint32_t kex_rounds_required)
+  {
+    // initialization is only needed during the first round
+    if (m_kex_rounds_complete > 0)
+      return;
+
+    // the first round of kex msgs will contain each participant's base pubkeys and ancillary privkeys, so we prepare
+    //    them here
+
+    // collect participants' base common privkey shares
+    // note: duplicate privkeys are acceptable, and duplicates due to duplicate signers
+    //       will be blocked by duplicate-signer errors after this function is called
+    std::vector<crypto::secret_key> participant_base_common_privkeys;
+    participant_base_common_privkeys.reserve(expanded_msgs.size() + 1);
+
+    // add local ancillary base privkey
+    participant_base_common_privkeys.emplace_back(m_base_common_privkey);
+
+    // add other signers' base common privkeys
+    for (const multisig_kex_msg &expanded_msg : expanded_msgs)
+    {
+      if (expanded_msg.get_signing_pubkey() != m_base_pubkey)
+        participant_base_common_privkeys.emplace_back(expanded_msg.get_msg_privkey());
+    }
+
+    // make common privkey
+    make_multisig_common_privkey(std::move(participant_base_common_privkeys), m_common_privkey);
+
+    // set common pubkey
+    CHECK_AND_ASSERT_THROW_MES(crypto::secret_key_to_public_key(m_common_privkey, m_common_pubkey),
+      "Failed to derive public key");
+
+    // if N-of-N, then the base privkey will be used directly to make the account's share of the final key
+    if (kex_rounds_required == 1)
+    {
+      m_multisig_privkeys.clear();
+      m_multisig_privkeys.emplace_back(m_base_privkey);
     }
   }
   //----------------------------------------------------------------------------------------------------------------------
@@ -771,9 +779,7 @@ namespace multisig
       result_keys.reserve(result_keys_to_origins_map.size());
 
       for (const auto &result_key_and_origins : result_keys_to_origins_map)
-      {
         result_keys.emplace_back(result_key_and_origins.first);
-      }
 
       // compute final aggregate key, update local multisig privkeys with aggregation coefficients applied
       m_multisig_pubkey = generate_multisig_aggregate_key(std::move(result_keys), m_multisig_privkeys);
@@ -811,7 +817,8 @@ namespace multisig
         // derived pubkey = multisig_key * G
         crypto::public_key_memsafe derived_pubkey;
         m_multisig_privkeys.push_back(
-          calculate_multisig_keypair_from_derivation(derivation_and_origins.first, derived_pubkey));
+            calculate_multisig_keypair_from_derivation(derivation_and_origins.first, derived_pubkey)
+          );
 
         // save the account's kex key mappings for this round [derived pubkey : other signers who will have the same key]
         m_kex_keys_to_origins_map[derived_pubkey] = std::move(derivation_and_origins.second);
@@ -863,8 +870,7 @@ namespace multisig
       "Multisig kex has already completed all required rounds (including post-kex verification).");
 
     // initialize account update
-    std::vector<crypto::public_key> exclude_pubkeys;
-    initialize_kex_update(expanded_msgs, kex_rounds_required, exclude_pubkeys);
+    this->initialize_kex_update(expanded_msgs, kex_rounds_required);
 
     // process messages into a [pubkey : {origins}] map
     multisig_keyset_map_memsafe_t result_keys_to_origins_map;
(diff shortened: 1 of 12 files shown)


UkoeHB:
This PR should not be merged until after #8220.

gs522084:
I'm not a programmer, I don't quite understand the logic of this solution, can you explain its benefits in detail? thanks。。。

UkoeHB:
@gs522084 the PR description is written for non-programmers. Not sure what else I can add to it.

UkoeHB:
This PR requires [PR #8329](https://github.com/monero-project/monero/pull/8329).

src/multisig/multisig_account_kex_impl.cpp
```diff
@@ -716,11 +724,44 @@ namespace multisig
...
+      pubkey_origins_map);
+    CHECK_AND_ASSERT_THROW_MES(round == expected_msgs_round, "Kex messages were not for expected round.");
+
+    // dummy variable to satisfy function signature
+    // - do not save the DH derivations produced
+    // - the booster is not used directly for building a multisig account
+    std::unordered_map<crypto::public_key_memsafe, std::unordered_set<crypto::public_key>> pubkey_mapping_dummy;
+
+    // produce a kex message for the next round with given input pubkeys
+    return multisig_kex_make_next_msg(m_base_privkey, round + 1, threshold, num_signers,
```
vtnerd, line 763: I can't find the implementation of this function. Where is it? The builds supposedly pass, but it doesn't appear in the current codebase either. What am I missing?

src/wallet/wallet2.cpp
```diff
@@ -5169,16 +5181,65 @@ std::string wallet2::exchange_multisig_keys(const epee::wipeable_string &passwor
...
+    CHECK_AND_ASSERT_THROW_MES(num_signers == m_multisig_signers.size(),
+      "Expected number of signers does not match multisig wallet setting.");
+
+    // reconstruct multisig account
+    get_reconstructed_multisig_account(multisig_account);
+  }
+  else
+  {
+    // make_multisig() has not been called
+    // DANGER: If 'num_signers - threshold > 1', but this wallet's future multisig settings
```
vtnerd, line 5224: Shouldn't this be a hard error (exception)? Why allow it?

vtnerd (changes requested):
I don't see one of the functions, so I'm wondering whether there is a missing file or something.

src/wallet/wallet2.cpp
```diff
@@ -4944,32 +4944,75 @@ void wallet2::restore(const std::string& wallet_, const epee::wipeable_string& p
...
-  const std::vector<std::string> &initial_kex_msgs,
-  const std::uint32_t threshold)
+epee::misc_utils::auto_scope_leave_caller wallet2::decrypt_account_for_multisig(const epee::wipeable_string &password)
 {
   // decrypt account keys
+  // note: this conditional's clauses are old and undocumented
   epee::misc_utils::auto_scope_leave_caller keys_reencryptor;
   if (m_ask_password == AskPasswordToDecrypt && !m_unattended && !m_watch_only)
   {
     crypto::chacha_key chacha_key;
```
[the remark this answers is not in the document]
UkoeHB: Looks like [it already is](https://github.com/monero-project/monero/blob/fc907a957078cae2dc68348886a33363848dd089/src/crypto/chacha.h#L54)

src/wallet/wallet2.cpp
```diff
@@ -4944,32 +4944,75 @@ void wallet2::restore(const std::string& wallet_, const epee::wipeable_string& p
...
     crypto::generate_chacha_key(password.data(), password.size(), chacha_key, m_kdf_rounds);
-    m_account.encrypt_viewkey(chacha_key);
-    m_account.decrypt_keys(chacha_key);
+    decrypt_keys(password);
```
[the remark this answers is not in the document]
UkoeHB: Just trying to remove some code duplication while I'm here

src/wallet/wallet2.cpp
UkoeHB: In this case, the user passed in `num_signers` and `threshold` manually for a multisig account that doesn't exist yet (e.g. because you are missing one participant's initial kex message - this will happen when boosting round 2 of a 2-of-3 setup where you get the round 1 initial kex message of one participant and then boost round 2 for the other who isn't available). If the manually defined `num_signers` and `threshold` are different from the `num_signers` and `threshold` that the actual future multisig account gets, then there can be a problem (as described in the code comment).

UkoeHB:
Rebased, should be good to go now

UkoeHB:
Force-pushed again to update the unit test for boosting a multisig wallet so that the boosted wallet uses force-update to finish the post-kex round.

src/wallet/wallet2.cpp
```diff
@@ -5063,17 +5106,19 @@ std::string wallet2::make_multisig(const epee::wipeable_string &password,
...
+    // add signer
+    signers.push_back(expanded_msgs.back().get_signing_pubkey());
   }
 
-  // add self to signers
-  signers.push_back(multisig_account.get_base_pubkey());
+  // expect that self is in the input list (this guarantees that the input list size always equals the number of intended
+  //   signers for the account [when combined with duplicate checking])
+  CHECK_AND_ASSERT_THROW_MES(std::find(signers.begin(), signers.end(), multisig_account.get_base_pubkey()) != signers.end(),
+    "The local account's signer key was not found in initial multisig kex messages when converting a wallet to multisig.");
```
UkoeHB, line 5121: This is a breaking change. In existing code, adding your own initial message to the `make_multisig()` command is optional.

src/wallet/wallet2.cpp
```diff
@@ -5021,40 +5021,83 @@ void wallet2::restore(const std::string& wallet_, const epee::wipeable_string& p
...
   }
 
-  // create multisig account
-  multisig::multisig_account multisig_account{
-      multisig::get_multisig_blinded_secret_key(get_account().get_keys().m_spend_secret_key),
-      multisig::get_multisig_blinded_secret_key(get_account().get_keys().m_view_secret_key)
+  return keys_reencryptor;
+}
+//----------------------------------------------------------------------------------------------------
+void wallet2::get_uninitialized_multisig_account(multisig::multisig_account &account_out) const
```
vtnerd, line 5045: I'm not sure why an out parameter was chosen, but there's little benefit to it. The code generated isn't going to be more efficient; all modern ABIs have the caller pass a pointer via register that the callee uses to `std::move` or directly initialize into. So the primary difference is (typically) a worse interface for the caller.

Same thing for the function below since it appears to be overwriting `out` entirely too.

src/wallet/wallet2.cpp
vtnerd: Yikes. Should the wallet track the values given in the booster request to ensure that they don't change in the future? Downstream projects can enforce this policy, but it seems to do it here if possible.

vtnerd (changes requested):
The only thing holding me from approving this is whether the error tracking on the uninitialized account can be improved (without too much hassle).

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -2558,6 +2558,35 @@ namespace wallet_rpc
     typedef epee::misc_utils::struct_init<response_t> response;
   };
 
+  struct COMMAND_RPC_GET_MULTISIG_KEY_EXCHANGE_BOOSTER
+  {
+    struct request_t
+    {
+      std::string password;
+      std::vector<std::string> multisig_info;
```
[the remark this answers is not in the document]
UkoeHB: The `multisig_info` here is a vector of `b58` strings, because we want the multisig stuff to be copy-pastable.

src/wallet/wallet2.cpp
UkoeHB: Do you mean 'instead of returning an account'? Outparams are just a convention I decided to stick with for uniform interfaces (although I do break that occasionally, especially for trivial return types).

src/wallet/wallet2.cpp
UkoeHB: The wallet _should_ not couple multisig and normal wallet code, so that the first step of making a multisig wallet is defining M and N and those values are baked into the multisig core account file.

In general, this boosting mechanism is very easy to use incorrectly since it only makes sense to use in combination with force updating the post-kex verification round (which is unsafe if used without nailing down the trust matrix of your use-case), so it should only be used by people who know what they are doing. Maybe we need some kind of gate for this 'enterprise-level' feature...

src/wallet/wallet_rpc_server_commands_defs.h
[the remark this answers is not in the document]
vtnerd: Ah crap, I forgot about that. I think I even mentioned this to you?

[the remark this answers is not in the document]
UkoeHB: I don't remember, but I do remember stuff breaking when I didn't use `b58` (maybe the deserializer can't handle a vector of strings that don't contain characters? I don't remember exactly).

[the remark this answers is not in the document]
vtnerd: Oh yeah, I remember the discussion more clearly now - we definitely selected base58 because it also helped with passing messages via email, etc. Damn, too focused on serialization or something :/

src/wallet/wallet2.cpp
vtnerd: Crap.

I've been pushing things towards `optional<T>` or `expect<T>` for various reasons, and away from the existing style of `bool` or `void` with out parameter (and towards a style where exceptions are _typically_ only for shutdown related events, which I kind-of violated in `wire::` serialization code).

Returning the type directly, especially when there is no `optional<T>` or `expect<T>`, is often easier on the caller code. Especially when the result is going directly into another function call, etc.

I thought perhaps you did this for performance reasons (typically that is what programmers cite), but this rationale is a bit different. If your happy with it, I'm not going to waste my time pushing back. Although it would be nice if seraphis used return-by-value where easily possible; (because as mentioned in last paragraph) return-by-value is often more composable.

UkoeHB: I'm actually skeptical about the value of ubiquitous function composition. Sometimes what you need for visual clarity is less code density, and sometimes what you need in order to realize 'I am writing spaghetti and need to decompose into more functions' is an excessive amount of lines of code.

UkoeHB: Also re: efficiency, sometimes an outparam is used to initialize an object in the caller and sometimes it is assigned to an object in the caller. For the assignment case I want to avoid excessive copies (although you seem to be implying this won't happen, so idk).

UkoeHB: > The code generated is typically close in both cases, but it also helps with bugs where an out parameter is aliasing another in parameter (waiting for Rust programmers to pile on here).

I suppose we are getting a bit off-topic, but this 'dangerous aliasing' is used for efficiency gains all over the crypto library.

```
crypto::secret_key a, b, c;
sc_add(a, a, b);  //a += b
sc_add(a, b, c);  //a = b + c
```

Where `sc_add()` is writing directly into the 32-byte outparam (the first parameter).

vtnerd: > Also re: efficiency, sometimes an outparam is used to initialize an object in the caller and sometimes it is assigned to an object in the caller. For the assignment case I want to avoid excessive copies (although you seem to be implying this won't happen, so idk).

I'm going overkill on this explanation, but it is interesting.

If the out parameter is only being partially modified, it is clearly almost always more efficient than return-by-value. Otherwise, return-by-value is typically better because RVO/NRVO produces ASM (basically) identical to an out parameter that is uninitialized (by caller) and marked with the `restrict` keyword. Since usage of the `restrict` keyword is rare, and types are often initialized to some value immediately in C++, the "conventional wisdom" is to now prefer return-by-value in most situations.

When RVO/NRVO fails, moving is mandatory if possible, otherwise the fallback is copying. The move case is typically similar (in performance but not ASM) to out parameter or slightly slower. Might be slightly faster sometimes since `restrict` is rarely used on out parameters.

The fallback copying case basically always stinks, unless its a simple bit-copy (where move and copy are identical).

----
This confused me at first, so to explain the ASM - the caller just provides an uninitialized chunk of memory on the stack via register to the callee, which then does one single direct construction on the chunk of memory when RVO/NRVO is possible. This memory address can then be passed via register to another function - so `void handle(std::string); std::string a_string(); void run() {handle(a_string()); }` can produce ASM with 0 copies or moves and 1 initialization ... all the way back in C++03 days.

Compare to out parameter - `std::string` must be initialized by caller, and then mutated or overwritten by callee, etc.

UkoeHB: > When RVO/NRVO fails, moving is mandatory if possible, otherwise the fallback is copying. The move case is typically similar (in performance but not ASM) to out parameter or slightly slower. Might be slightly faster sometimes since restrict is rarely used on out parameters.

I may be underestimating compilers, but from what I understand RVO/NRVO doesn't help when you have composite types that are basically big blobs, which is what we are dealing with in the seraphis library.

For example I have this (64 bytes):

```
struct SpEnote final
{
    rct::key m_onetime_address;
    rct::key m_amount_commitment;
};
```

And this function:

```
void make_seraphis_enote_core(const rct::key &onetime_address,
    const crypto::secret_key &amount_blinding_factor,
    const rct::xmr_amount amount,
    SpEnote &enote_core_out)
{
    enote_core_out.m_onetime_address = onetime_address;
    enote_core_out.m_amount_commitment = rct::commit(amount, rct::sk2rct(amount_blinding_factor));
}
```

Both of the following cases require you to 1) instantiate an `SpEnote`, 2) copy-assign the members of the enote.

```
// 1) return value (internally make `SpEnote temp;` then return it to RVO initialize the `enote`)
const SpEnote enote{make_seraphis_enote_core(Ko, x, a)};

// 2) outparam
SpEnote enote;
make_seraphis_enote_core(Ko, x, a, enote);
```

With the outparam, I can reuse an existing `SpEnote` instance to make and consume multiple enotes in sequence (hypothetically) without copy-assigning it in the loop. [Here's](https://github.com/UkoeHB/monero/blob/c35686fcbbd17838cf23130c53495a8dd0e50502/src/seraphis/serialization_demo_utils.cpp#L157) an example of something like that (I did a search on my code and discovered I don't actually use this pattern very often).

I can also do partial-construction of objects without copy-assigning twice (found a bunch of examples in enote record recovery [for example](https://github.com/UkoeHB/monero/blob/c35686fcbbd17838cf23130c53495a8dd0e50502/src/seraphis/tx_enote_record_utils.cpp#L387)).

UkoeHB: > >    I can also do partial-construction of objects
> Meaning some of the memory is left uninitialized? This is a bad idea unless value/brace initialization is used everywhere for the structs or a default constructor initializes everything.

Sorry, I meant step-based member-wise initialization of an outparameter that is reused between function calls. With a return value you need to copy assign all the members then copy assign the return value to the temporary shuttle object.

> The standard recommendation is to always initialize variables on declaration in C++, which tips things towards return-by-value a bit.

I definitely agree that non-data types with some kind of 'behavior' must have a strict C++ interface with constructors and member functions. Doing that for data types where all your operations are data member writes and reads creates a kind of big mess of getters and setters. Maybe there is a way to do it that's clean, but I couldn't figure it out.

vtnerd: I'm hihacking this PR (sorry), but for posterity:

> Sorry, I meant step-based member-wise initialization of an outparameter that is reused between function calls. With a return value you need to copy assign all the members then copy assign the return value to the temporary shuttle object.

I might be misunderstanding this statement, but if you meant something like:

```c++
struct blob { char value[32]; };
struct two_blobs { blob values[2]; };
blob return_blob();
two_blobs return_blobs() { return {return_blob(), return_blob()}; }
```

take that code snippet to gcc godbolt, you get:

```
return_blobs():
        push    rbx
        mov     rbx, rdi
        call    return_blob()
        lea     rdi, [rbx+32]
        call    return_blob()
        mov     rax, rbx
        pop     rbx
        ret
```

You should get similar results if `return_blobs()` used a named return variable. The callee just forwards the ("hidden") out address it received to the next function for initialization. It's identical to forwarding the address of the out parameter field/member. This of course, doesn't always work out exactly (so I'm _not_ proposing a change away from out variables).

UkoeHB: I mean something like this:

```
struct two_blobs { blob values[2]; };

bool try_get_blobs(const std::size_t i, two_blobs &blobs_out)
{
    if (!try_get_blob1(i, blobs_out[0])
        return false;

    get_blob2(blobs_out[1]);

    return true;
}

int main()
{
    two_blobs blobs_temp;
    for (std::size_t i{0}; i < n; ++i)
    {
        if (!try_get_blobs(i, blobs_temp)
            continue;

        do_something(blobs_temp);
    }
}
```

src/wallet/wallet2.cpp
UkoeHB: Updating wallet2 to have safer multisig-wallet-creation semantics is out of scope for this PR. I will mark it as a 'todo potential future PR'.

vtnerd (approved):

selsta:
@UkoeHB needs a rebase

selsta (approved):

PR #8207, opened 2022-03-05
CMake: Add missing headers via monero_find_all_headers macro

mj-xmr:
Using the `monero_find_all_headers()` macro to add missing headers, where appropriate.

Whenever it wasn't a good idea, like in `wallet_api`, where the Author actually distinguishes between API and private headers, the macro wasn't used.

## Special cases
- `serialization` was actually missing all of its headers, except for one.
- `daemon` was referencing external headers (`cryptonote_protocol` and `p2p`), which are imported anyway through `target_link_libraries()`, so I took the liberty of removing these references

### DIFF
--- a/src/blockchain_db/CMakeLists.txt
+++ b/src/blockchain_db/CMakeLists.txt
@@ -33,10 +33,7 @@ set(blockchain_db_sources
 
 set(blockchain_db_headers)
 
-set(blockchain_db_private_headers
-  blockchain_db.h
-  lmdb/db_lmdb.h
-  )
+monero_find_all_headers(blockchain_db_private_headers "${CMAKE_CURRENT_SOURCE_DIR}")
 
 monero_private_headers(blockchain_db
   ${crypto_private_headers})
--- a/src/daemon/CMakeLists.txt
+++ b/src/daemon/CMakeLists.txt
@@ -37,30 +37,7 @@ set(daemon_sources
 
 set(daemon_headers)
 
-set(daemon_private_headers
-  command_parser_executor.h
-  command_server.h
-  core.h
-  daemon.h
-  executor.h
-  p2p.h
-  protocol.h
-  rpc.h
-  rpc_command_executor.h
-
-  # cryptonote_protocol
-  ../cryptonote_protocol/cryptonote_protocol_defs.h
-  ../cryptonote_protocol/cryptonote_protocol_handler.h
-  ../cryptonote_protocol/cryptonote_protocol_handler.inl
-  ../cryptonote_protocol/cryptonote_protocol_handler_common.h
-
-  # p2p
-  ../p2p/net_node.h
-  ../p2p/net_node_common.h
-  ../p2p/net_peerlist.h
-  ../p2p/net_peerlist_boost_serialization.h
-  ../p2p/p2p_protocol_defs.h
-  ../p2p/stdafx.h)
+monero_find_all_headers(daemon_private_headers "${CMAKE_CURRENT_SOURCE_DIR}")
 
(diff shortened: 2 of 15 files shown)


src/blockchain_db/CMakeLists.txt
```diff
@@ -33,10 +33,7 @@ set(blockchain_db_sources
 
 set(blockchain_db_headers)
 
-set(blockchain_db_private_headers
-  blockchain_db.h
-  lmdb/db_lmdb.h
-  )
+monero_find_all_headers(blockchain_db_private_headers "{CMAKE_CURRENT_SOURCE_DIR}")
```
selsta, line 36:
```suggestion
monero_find_all_headers(blockchain_db_private_headers "${CMAKE_CURRENT_SOURCE_DIR}")
```

mj-xmr: Good catch. Resolved.

src/daemon/CMakeLists.txt
```diff
@@ -37,30 +37,7 @@ set(daemon_sources
...
-  ../cryptonote_protocol/cryptonote_protocol_handler_common.h
-
-  # p2p
-  ../p2p/net_node.h
-  ../p2p/net_node_common.h
-  ../p2p/net_peerlist.h
-  ../p2p/net_peerlist_boost_serialization.h
-  ../p2p/p2p_protocol_defs.h
-  ../p2p/stdafx.h)
+monero_find_all_headers(daemon_private_headers "${CMAKE_CURRENT_SOURCE_DIR}")
```
jeffro256, line 40: Is it okay to use monero_find_all_headers here since it won't capture the ../cryptonote_protocol and ../p2p headers?

mj-xmr: They are (and should be) collected by their individual CMakeLists.txt and only exported to the dependent libraries. In our case, the headers are being exported because the `src` directory is the "header directory". Therefore we access these headers via:
`#include "p2p/p2p_protocol_defs.h"`
and not just:
`#include "p2p_protocol_defs.h"`

mj-xmr: And if only the `GNU make` discovers that any of the headers changes, it will recompile the source code, whether they are explicitly mentioned as before or not. 

The difference is in the IDE itself. If a header is not captured, it won't be visible in the IDE anymore, and even if you open the header in the IDE, change it and expect the dependent sources to compile, they won't since the IDE wouldn't autosave the header's changes, thus giving no signal to `GNU make`. I checked both the `p2p` and `cryptonote_protocol` and their files are all globed.

jeffro256: > They are (and should be) collected by their individual CMakeLists.txt

That's what I thought! What then was the original point then of putting these headers in here explicitly? Maybe it was done automatically ? Anyways, I'm sure that you thought about that, it was just confusing me.  

> I checked both the p2p and cryptonote_protocol and their files are all globed

Nice

mj-xmr: Without spending any time on researching it, I could imagine, that somebody first added these references here, because he was working on the daemon. Then somebody else noticed that the these individual packets don't work properly and globed them inside these packets. I've seen many things already :) Including "borrowing" entire source files, thus compiling them again, just because these people weren't feeling responsible for keeping the build files in order, or simply because they had more important things to do and "The Sprint" was reaching its end ...

selsta (approved):

selsta:
@mj-xmr please rebase

mj-xmr:
@selsta : rebased.

selsta (approved):

PR #8211, opened 2022-03-08
EPEE Cleanup Reorganized

jeffro256:
This PR is identical to #8175 but organized in a way that I believe is easier to verify for others. I will not adding any more of my own commits to this PR, but I'm ready to squash or add changes if/when this PR gets approved. If you want more granular comments on changes, I would refer to the old PR, but I'm happy to answer any questions. Thanks for your time!


mj-xmr:
I would review this if the commits were separated into individual PRs. The reasons behind this are, that this reduces the risk of merge conflicts and that it's a lot of work to do in one go. For these 2 reasons you will typically not get a 2nd opinion, so in effect the PR will not be merged, which would lead to a lot of time waste from your and my side.

The 3rd reason for the split, that I personally have, is that some code can possibly still be useful in the future, even if it's unused for now. I'd like to have a possibility to express my opinion about it. The larger the PR, the harder it will be to reach a consensus on this, since the decision is basically binary (merge or not merge).

selsta:
@mj-xmr they are separate here: https://github.com/monero-project/monero/pull/8175

but I think the problem was that 40 commits became too large

selsta:
@mj-xmr I initially said to keep it in one PR and I think it's fine if we discuss everything here, otherwise we will have 10 open PRs about epee cleanup. Most of these changes should be quite uncontroversial anyway.

mj-xmr:
See? We have the 1st disagreement already :)

jeffro256:
> I would review this if the commits were separated into individual PRs. 

@mj-xmr What @selsta said about the old PR #8175 is correct. The old PR shows more of my thought process, has much finer changes, and more comments about each change. The problem is that it is lengthy and hard to digest. This PR I focused on ease of review. 

> The 3rd reason for the split, that I personally have, is that some code can possibly still be useful in the future, even if it's unused for now. 

I considered this for sure, but I tend to fall under the camp of YAGNI. Many of the deletions are files/functions that were inherited from some botnet that predates the project itself and were never used. EPEE is very daunting for new developers and it doesn't help that ~25% of it is red herring code. Also I figured that the less cruft that you all have to deal with the better.

> I'd like to have a possibility to express my opinion about it. The larger the PR, the harder it will be to reach a consensus on this, since the decision is basically binary (to merge or not to merge).

If you have an issue with a specific part of this PR I'd be happy to hear it, and I think I have it fairly well organized under the old PR, so I can rebase it if needed. Thanks for your time btw!

mj-xmr:
@jeffro256 I will conform to how it was expected from you from @selsta , since this was the 1st suggestion in the row, while mine was the 2nd. Otherwise we'd be going in circles.

I'll have a look then. I just wish that the PR isn't left hanging **after** being reviewed.

jeffro256:
@selsta I think the vast majority of this PR should be uncontroversial but I'll try to point out a couple of things that might not be:

* Refactoring `get_ns_count()` in `time_helper.h` (previously in `misc_os_dependent.h` to use `std::chrono::steady_clock` instead of the medley of implementation-specific hacks it was using before
* When factoring out `pragma_comp_defs.h`, I removed some macros which were suppressing warnings. I think I covered my bases, but there's a chance the change might generate a lot of spam on compilers that aren't GCC.
* Moving copyable_atomic to be a subclass of cryptonote_connection_context (I commented on that change in the code)
* Removing epee/tests since they weren't be used and would take a lot of refactoring to get to where they could compile

jeffro256:
@mj-xmr @selsta Another thing that I forgot to mention is that either PR should build with no issues no matter at which commit in the branch you pick. I hope that makes it a little easier :)

mj-xmr:
> why would there merge conflicts in files untouched for 5+ years?

I touched some of them myself not so long ago. Somebody else might be right now modifying some of them and starting a PR. But even if not, it's kind of a bad habit and it might bite back if continued this way. And after all, if in meantime somebody were to merge a change to just one of the files and create a conflict, this whole PR has to be updated, rather than a given smaller PR from the whole group. But anyway, I'll just get my hands dirty and start reviewing.

mj-xmr:
@jeffro256 

> I considered this for sure, but I tend to fall under the camp of YAGNI. Many of the deletions are files/functions that were inherited from some botnet that predates the project itself and were never used. EPEE is very daunting for new developers and it doesn't help that ~25% of it is red herring code. Also I figured that the less cruft that you all have to deal with the better.

I agree. Especially since I saw, that with removal of some code from the templates, you took out some heavy boost headers along. I'm just saying that there still might be some nice code from the deletions that you did (like 5%?). Let me have a look and I'll just share my opinion based on my experience, which can be also safely ignored.

contrib/epee/src/file_io_utils.cpp
```diff
@@ -102,29 +102,6 @@ namespace file_io_utils
 	}
 
 
-	bool get_file_time(const std::string& path_to_file, time_t& ft)
```
mj-xmr, line 105: I think it could be useful and nicely wraps boost::filesystem's heavy headers.

contrib/epee/src/file_io_utils.cpp
```diff
@@ -102,29 +102,6 @@ namespace file_io_utils
...
-		boost::system::error_code ec;
-		ft = boost::filesystem::last_write_time(boost::filesystem::path(path_to_file), ec);
-		if(!ec)
-			return true;
-		else
-			return false;
-	}
-
-
-	bool set_file_time(const std::string& path_to_file, const time_t& ft)
```
mj-xmr, line 116: I think it could be useful and nicely wraps boost::filesystem's heavy headers.

contrib/epee/include/global_stream_operators.h
```diff
@@ -1,35 +0,0 @@
...
-// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
-// ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
-// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
-// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-// 
-
-
-#pragma once
-
-std::stringstream& operator<<(std::stringstream& out, const std::wstring& ws)
```
mj-xmr, line 30: This looks useful.

contrib/epee/include/ado_db_helper.h
```diff
@@ -1,1095 +0,0 @@
...
-// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-// 
-
-
-#ifndef _DB_ADO_HELPER_H_
-#define _DB_ADO_HELPER_H_
-
-#include <vector>
-#include <comutil.h>
-#include "string_coding.h"
```
mj-xmr, line 33: This header can probably also be deleted.

contrib/epee/include/misc_language.h
```diff
@@ -30,74 +30,14 @@
...
-    {
-      while(count--)
-        it--;
-      return it;
-    }
-
-
-	// TEMPLATE STRUCT less
-	template<class _Ty>
-	struct less_as_pod
```
mj-xmr, line 85: Appears weird and unreliable.

contrib/epee/include/misc_language.h
```diff
@@ -30,74 +30,14 @@
...
-	template<typename t_iterator>
-		t_iterator move_it_forward(t_iterator it, size_t count)
-		{
-			while(count--)
-				it++;
-			return it;
-		}
-
-    template<typename t_iterator>
-    t_iterator move_it_backward(t_iterator it, size_t count)
```
mj-xmr, line 75: I'd leave it.

contrib/epee/include/misc_language.h
```diff
@@ -30,74 +30,14 @@
...
 {
-	template<typename t_type>
-		t_type get_max_t_val(t_type t)
-		{
-			return (std::numeric_limits<t_type>::max)();
-		}
-
-		
-	template<typename t_iterator>
-		t_iterator move_it_forward(t_iterator it, size_t count)
```
mj-xmr, line 67: I'd leave it.

contrib/epee/include/misc_language.h
```diff
@@ -30,74 +30,14 @@
...
-	}
-
-
-
 #define AUTO_VAL_INIT(v)   boost::value_initialized<decltype(v)>()
 
 namespace misc_utils
 {
-	template<typename t_type>
-		t_type get_max_t_val(t_type t)
```
mj-xmr, line 60: I'd leave it.

contrib/epee/include/net/abstract_tcp_server.h
```diff
@@ -1,318 +0,0 @@
...
-		closesocket(pthread_context->m_socket);
-
-		abstract_tcp_server* powner = pthread_context->powner;
-		LOG_PRINT("Handler thread with socket=" << pthread_context->m_socket << " STOPPED",  LOG_LEVEL_2);
-		powner->m_connections_lock.lock();
-		::CloseHandle(pthread_context->m_htread);
-		pthread_context->powner->m_connections.erase(pthread_context->m_self_it);
-		powner->m_connections_lock.unlock();
-		CoUninitialize();
-		::InterlockedDecrement(&pthis->m_threads_count);
```
mj-xmr, line 157: Too much Windoze here anyway.

selsta:
My personal opinion, I would remove everything unless it can be argued that a function is likely to be used in the future. Something just generally looking useful without any particular use in monero should be removed.

Ideally we would shrink epee over time with all new code getting added to `src/`.

contrib/epee/include/storages/portable_storage_to_bin.h
```diff
@@ -70,13 +64,13 @@ namespace epee
         return pack_varint_t<uint32_t>(strm, PORTABLE_RAW_SIZE_MARK_DWORD, val);
       }else
       {
-        CHECK_AND_ASSERT_THROW_MES(val <= 4611686018427387903, "failed to pack varint - too big amount = " << val);
+        // Same as checking val <= 4611686018427387903 except that it's portable for 32-bit size_t
+        CHECK_AND_ASSERT_THROW_MES(!(val >> 31 >> 31), "failed to pack varint - too big amount = " << val);
```
mj-xmr, line 68: I tested locally that both versions work the same way.

contrib/epee/include/storages/gzipped_inmemstorage.h
```diff
@@ -1,68 +0,0 @@
...
-#define _GZIPPED_INMEMSTORAGE_H_
-
-#include "zlib_helper.h"
-namespace epee
-{
-namespace StorageNamed
-{
-
-	template<class t_base_storage>
-	class gziped_storage: public t_base_storage
```
mj-xmr, line 38: Open for debate, but I find such a "demo" on how to use zipping important.

contrib/epee/include/time_helper.h
```diff
@@ -28,132 +28,60 @@
...
 	}
 
 	inline std::string get_time_interval_string(const time_t& time_)
 	{
-		std::string res;
 		time_t tail = time_;
-PRAGMA_WARNING_PUSH
-PRAGMA_WARNING_DISABLE_VS(4244)
-		int days = tail/(60*60*24);
+		const int days = (int) (tail/(60*60*24));
```
mj-xmr, line 61: In C++ you'd use `static_cast<int>(tail/(60*60*24);` for better clarity and to avoid warnings.

jeffro256: Good point, see the most recent commit. Thanks for reviewing!

contrib/epee/src/file_io_utils.cpp
jeffro256: On the other hand, `boost::filesystem` functions are already mentioned 245 times in 9 files throughout the project, and I believe boost already has functions for `get_file_time` and `set_file_time`: [https://www.boost.org/doc/libs/1_58_0/libs/filesystem/doc/reference.html#last_write_time](https://www.boost.org/doc/libs/1_58_0/libs/filesystem/doc/reference.html#last_write_time)

contrib/epee/include/misc_language.h
jeffro256: Why use `get_max_t_val` instead of `std::numeric_limits<t_type>::max()` directly? `std::numeric_limits` is already used 148 times throughout the project and IMO its confusing to pass in a raw value and receive a static limit unrelated to the value passed to the function.

contrib/epee/include/time_helper.h
```diff
@@ -28,132 +28,60 @@
...
 namespace epee
 {
 namespace misc_utils
 {
-
-#ifdef __ATLTIME_H__
-
-	inline
-	bool get_time_t_from_ole_date(DATE src, time_t& res)
+	inline bool get_gmt_time(time_t t, struct tm &tm)
```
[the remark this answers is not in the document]
jeffro256: `time_helper.h` definitely changed the most out of any file in this PR

contrib/epee/include/storages/gzipped_inmemstorage.h
jeffro256: Should we create a doc then? Maybe we could it in the EPEE Readme.

contrib/epee/src/file_io_utils.cpp
mj-xmr: You have to start from somewhere and lead by a good example, even if at the beginning it appears to make no sense, because it's already much worse somewhere else. Bit by bit the better solutions can purge the worse :)

contrib/epee/include/misc_language.h
mj-xmr: As far as I see it evaluates to the same type and value.

It's not a state of the art function though anyway.

contrib/epee/include/time_helper.h
mj-xmr: N.P. The latest commit looks good in this regard. 

I'm only 25% done for now. I shall continue tomorrow.

contrib/epee/include/misc_language.h
jeffro256: Yeah it should evaluate to to the same value, but it feels clunkier to use IMO because you pass a raw value of the type that you want to get the limit of, instead of just asking for it based on type.

contrib/epee/include/storages/gzipped_inmemstorage.h
mj-xmr: Sounds like a compromise for sure.

contrib/epee/include/zlib_helper.h
```diff
@@ -1,139 +0,0 @@
...
-
-#pragma once
-extern "C" { 
-#include "zlib/zlib.h"
-}
-#pragma comment(lib, "zlibstat.lib")
-
-namespace epee 
-{
-namespace zlib_helper
```
mj-xmr, line 36: Another one, that I'd find worth keeping, since we're having problems with spamming logs sometimes. 

But if I have to choose between this and https://github.com/monero-project/monero/pull/7610 , I prefer having the latter.

contrib/epee/tests/src/tests.cpp
```diff
@@ -1,59 +0,0 @@
...
-    log_space::log_singletone::get_default_log_file().c_str(), 
-    log_space::log_singletone::get_default_log_folder().c_str());
-
-
-  string_tools::command_line_params_a start_params;
-  string_tools::parse_commandline(start_params, argc, argv);
-  std::string tests_data_path;
-  string_tools::get_xparam_from_command_line(start_params, std::string("/tests_folder"), tests_data_path);
-
-  if(string_tools::have_in_command_line(start_params, std::string("/run_net_tests")))
```
mj-xmr, line 28: Hmmm. Are these tests already duplicated in Monero's `tests` dir?

src/cryptonote_basic/connection_context.h
```diff
@@ -55,6 +55,37 @@ namespace cryptonote
...
     };
 
+    /*
+      This class was originally from the EPEE module. It is identical in function to std::atomic<uint32_t> except
+      that it has copy-construction and copy-assignment defined, which means that earliers devs didn't have to write
+      custom copy-contructors and copy-assingment operators for the outer class, cryptonote_connection_context.
+      cryptonote_connection_context should probably be refactored because it is both trying to be POD-like while
+      also (very loosely) controlling access to its atomic members.
+    */
+    class copyable_atomic: public std::atomic<uint32_t>
```
mj-xmr, line 65: That's cool that you took it over, but I'd ask moving it to the `common` dir and giving it its own source file. At least a header file, if header file + a .cpp file is too much of a change for now.

src/p2p/net_peerlist.h
```diff
@@ -214,6 +215,16 @@ namespace nodetool
     }
   }
   //--------------------------------------------------------------------------------------------------
+  inline
+  peerlist_entry peerlist_manager::get_nth_latest_peer(peers_indexed& peerlist, const size_t n)
```
mj-xmr, line 219: I personally think we need more such refactorings. Also thanks for adding the comment below.

src/p2p/net_peerlist.h
```diff
@@ -214,6 +215,16 @@ namespace nodetool
...
+  {
+    // Is not thread-safe nor does it check bounds. Do this before calling. Indexing starts at 0.
+    peers_indexed::index<by_time>::type& by_time_index = peerlist.get<by_time>();
+    auto by_time_it = --by_time_index.end();
+    std::advance(by_time_it, -((long long) n));
```
mj-xmr, line 224:
```C++
std::advance(by_time_it, -(static_cast<long long>(n)));
```
How did it behave without the cast, BTW?

jeffro256: I didn't try without a cast because n is type `size_t` and `size_t` is unsigned, so I'm pretty sure `-n` would be undefined. Even if it isn't undefined, I'm sure some compiler would throw out a warning regarding negating an unsigned type. I will change it to a `static_cast<>` though!

contrib/epee/include/time_helper.h
```diff
@@ -28,132 +28,60 @@
...
+		const int minutes = static_cast<int>(tail/60);
 		tail = tail%(60);
-		int seconds = tail;
-PRAGMA_WARNING_POP
-		res = std::string() + "d" + boost::lexical_cast<std::string>(days) + ".h" + boost::lexical_cast<std::string>(hours) + ".m" + boost::lexical_cast<std::string>(minutes) + ".s" + boost::lexical_cast<std::string>(seconds);
-		return res;
+		const int seconds = static_cast<int>(tail);
+
+		char tmpbuf[64] = {0};
+		snprintf(tmpbuf, sizeof(tmpbuf) - 1, "d%d.h%d.m%d.s%d", days, hours, minutes, seconds);
```
mj-xmr, line 70: This is definitely faster than the boost version (by about 10 times, based on my experiments with similar data), but was it a bottleneck? 

I'm only asking because introducing C functions like this, even though they are faster, introduces some risks. I see that you handled it well though.

contrib/epee/include/time_helper.h
```diff
@@ -28,132 +28,60 @@
...
-		if(TRUE != ::SystemTimeToVariantTime(&st, &oledb_date))
-			return false;
-		return true;
+		typedef std::chrono::duration<uint64_t, std::nano> ns_duration;
+		const ns_duration ns_since_epoch = std::chrono::steady_clock::now().time_since_epoch();
+		return ns_since_epoch.count();
 	}
 
-#endif
+	inline uint64_t get_tick_count()
```
mj-xmr, line 82: Is this used anywherere?

src/cryptonote_basic/connection_context.h
jeffro256: What I was going to do originally was rewrite the copy construction / copy assignment for `cryptonote_connection_context`, because that's the "correct" way to use atomic members. There's a reason that `std::atomic` has its copy construction and assignment deleted. The issue is that that creates a new function that you have to update when you add or remove members from `cryptonote_connection_context`.

Ideally, cryptonote_connection_context would be refactored to not have any atomic members, and instead be accessed through a mutex (for example).  That would take a decent amount of refactoring of pretty critical code that I am not as familiar with. My idea was that this specific change would not permanent if it got merged, and would get removed completely somewhere down the line

contrib/epee/tests/src/tests.cpp
jeffro256: Good point, we might be able to increase coverage with these tests. Either way, they don't compile, but maybe I can see if I can add these tests to the main tests/ directory.

mj-xmr:
I have no more comments. 

Other than the tests, that I don't like having removed, before they're reflected fully in our own source code, it looks OK.

src/p2p/net_peerlist.h
mj-xmr: I understand. OK this makes sense.

contrib/epee/include/misc_language.h
jeffro256: `std::advance` does the same thing, and it also does a compile-time check to see if the iterator is random-access to turn an O(count) operation into a O(1) operation.

contrib/epee/include/misc_language.h
jeffro256: `std::advance` does the same thing, and it also does a compile-time check to see if the iterator is random-access to turn an O(count) operation into a O(1) operation.

contrib/epee/include/zlib_helper.h
jeffro256: I don't think zlib_helper is very usable for anything bigger than some small strings. It's just not written very well, and we could probably write something better for the more specific task of streaming out compressed logs. For example, zlib_helper naively allocates 48x the size of the packed string when unpacking. I think if we were to write custom log compression code, that we should toss this and do it correctly.

contrib/epee/include/time_helper.h
jeffro256: It definitely wasn't a bottleneck, but I think its a bunch clearer and prettier than a long chain of `boost::lexical_cast<std::string>`. That along with the fact that that line is now half the length was the main reason. Ffaster runtime and compile times are a nice bonus though. I appreciate you noticing that though :)

contrib/epee/tests/src/tests.cpp
jeffro256: @mj-xmr Here's my more detailed analysis of epee/tests:

epee/tests/src/misc/test_math.h is still valid, will incorporate those these
epee/tests/src/storages/storages_tests.h tests StorageNamed, which doesn't exist anymore
epee/tests/src/storages/portable_storages_test.h uses an outdated version of EPEE serialization
contrib/epee/tests/src/net/test_net.h tests levin_protocol_handler_async.h for which there is already a unit_test
tests.cpp ties them all together

jeffro256:
@mj-xmr Here's my more detailed analysis of epee/tests:

epee/tests/src/misc/test_math.h is still valid, will incorporate those these
epee/tests/src/storages/storages_tests.h tests StorageNamed, which doesn't exist anymore
epee/tests/src/storages/portable_storages_test.h uses an outdated version of EPEE serialization
contrib/epee/tests/src/net/test_net.h tests levin_protocol_handler_async.h for which there is already a unit_test
tests.cpp ties them all togethe

mj-xmr:
> epee/tests/src/misc/test_math.h is still valid, will incorporate those these

I think I remember somebody has done this already. Just hinting before you start working.

src/cryptonote_basic/connection_context.h
mj-xmr: I understand.

jeffro256:
That last force push was to remove the commit where I added the test. This is because there are already PRs in progress (#7856 and #7819) which refactor `epee::misc_utils::median`.

mj-xmr (approved):
The math tests that would otherwise be gone to Nirvana are actually already handled in https://github.com/monero-project/monero/pull/7856

Other than this, there's nothing lost, that was very valuable.

tobtoht (approved):
Looks good to me.

jeffro256:
This force push is to fix merge conflicts with #8186. The author fixed undefined behavior in three files which are unused and which I deleted in this PR: `levin_client.inl`, `levin_client_async.h`, and `levin_helper.h`.How I rebased:
```bash
git checkout epee_cleanup_rev
git rebase master
cd contrib/epee/include/net/
git rm levin_client.inl levin_client_async.h levin_helper.h
git rebase --continue
git push --force
```

mj-xmr (approved):

mj-xmr:
I went through and still approve.

@majestrate :

> why would there merge conflicts in files untouched for 5+ years?

I'd hate to be a smartass, but what happened is exactly what I had predicted and in the very first set of merges.

majestrate:
> I went through and still approve.
> 
> @majestrate :
> 
> > why would there merge conflicts in files untouched for 5+ years?
> 
> I'd hate to be a smartass, but what happened is exactly what I had predicted and in the very first set of merges.

@jeffro256 already separated the commits into multiple PRs at first then combined them into one big PR (this one) and the first feedback was to "make a bunch of smaller PRs". All for removing 10K lines of dead code. I hate to be smartass too but this is bike shedding not code review.

majestrate:
the existence of merge conflicts on files used in dead code paths is an indicator of a deeper problem in the development and code review process. @jeffro256 don't stop, you're doing great.

selsta:
@jeffro256 please rebase

mj-xmr:
How could there be any merge conflicts? ^_^

>> I would review this if the commits were separated into individual PRs. The reasons behind this are, that this reduces the risk of merge conflicts and that it's a lot of work to do in one go. For these 2 reasons you will typically not get a 2nd opinion, so in effect the PR will not be merged, which would lead to a lot of time waste from your and my side.

>> The 3rd reason for the split, that I personally have, is that some code can possibly still be useful in the future, even if it's unused for now. I'd like to have a possibility to express my opinion about it. The larger the PR, the harder it will be to reach a consensus on this, since the decision is basically binary (to merge or not to merge).

> why would there merge conflicts in files untouched for 5+ years?

>> I touched some of them myself not so long ago. Somebody else might be right now modifying some of them and starting a PR. But even if not, it's kind of a bad habit and it might bite back if continued this way. And after all, if in meantime somebody were to merge a change to just one of the files and create a conflict, this whole PR has to be updated, rather than a given smaller PR from the whole group. But anyway, I'll just get my hands dirty and start reviewing.

majestrate:
>>> The 3rd reason for the split, that I personally have, is that some code can possibly still be useful in the future, even if it's unused for now. 

the dead code will still be in the git history if you need. just remove it so the development can speed up.

>> How could there be any merge conflicts? ^_^

they can happen easily when contributors want to keep dead code around. this code existing in the current development branches really holds monero back, if you really want to see it you can always access it via an older tag. we removed these hunks and more in oxen ages ago and it allowed us to do magical things without being encumbered by tech debt. i can understand why no one has in monero, it's a painful tedious thankless task as no one is being paid full time to make the code tolerable and accessible. 

> the existence of merge conflicts on files used in dead code paths is an indicator of a deeper problem in the development and code review process.

you guys should invest in a project manager or a full time salaried senor lead developer.

jeffro256:
@selsta Branch has been rebased. Rebasing process (after fetching and merging `master` on GH):

    git checkout master
    git pull
    git checkout epee_cleanup_rev
    git rebase master
    git rm contrib/epee/include/net/net_fwd.h src/p2p/stdafx.h
    git rebase --continue
    git push --force

jeffro256:
@selsta Do you want me to squash these commits now?

selsta:
@jeffro256 They are separate patches so I would not squash. If we want to revert anything specific in the future it's easier this way.

selsta (approved):

mj-xmr (approved):
Rechecked.

mj-xmr:
> the dead code will still be in the git history if you need. just remove it so the development can speed up.

I agree.
 
> > How could there be any merge conflicts? ^_^
>
> they can happen easily when contributors want to keep dead code around. this code existing in the current development branches really holds monero back, if you really want to see it you can always access it via an older tag. we removed these hunks and more in oxen ages ago and it allowed us to do magical things without being encumbered by tech debt. i can understand why no one has in monero, it's a painful tedious thankless task as no one is being paid full time to make the code tolerable and accessible.
> 
> the existence of merge conflicts on files used in dead code paths is an indicator of a deeper problem in the development and code review process.

Knowing ahead of time of the existence of the problem (plenty of dead code + one large PR, even though the commits are separate) and its accompanying risks (merge conflicts, preventing the merge of all of the commits, rather than only the affected ones), making a prediction, which got confirmed twice in both tries (merge bursts), yet still being confronted with criticism and disbelief(?) is a funny feeling.

BTW, I didn't **insist** on keeping the dead code around, just shared an opinion about a few specific functions for others to consider. 

Anyway, we're on the same page when it comes to highlighting the need of maintenance... in many forms.

> 
> you guys should invest in a project manager or a full time salaried senor lead developer.

The Monero's CCS is open for anybody. However I think, that Managers would not find this place very attractive for them. We need people who can do the actual work. Ability to manage is just a perk and is not reserved just to Managers.

What I think we should do is to simply learn from our and each others' mistakes and carry on doing what we already do.

tobtoht (approved):

mj-xmr (approved):
Rechecked the recently synchronized file.

jeffro256:
@luigi1111 looks good!

mj-xmr (approved):
Nothing changed from the reviewer's perspective.

Gingeropolous:
Huzzah! Take that entropy!

PR #8215, opened 2022-03-12
wallet_rpc_server: support regex for tags in get_accounts

### LINKED ISSUE #8214: Regexp support for listing accounts
From NakedKing on IRC, who would like the ability to use get\_accounts with a regular expression filter.

reemuru:
Closes #8214 
This commit adds a 'regexp' boolean field to the get_accounts
request. The flag is set to false by default and maintains backwards
compatibility. When set to true the user can search tags by regular
expression filters. An additional error message was added for failed
regular expression searches. Bump minor version to 25.

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -622,7 +622,7 @@ namespace tools
       res.total_unlocked_balance = 0;
       cryptonote::subaddress_index subaddr_index = {0,0};
       const std::pair<std::map<std::string, std::string>, std::vector<std::string>> account_tags = m_wallet->get_account_tags();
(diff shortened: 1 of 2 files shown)


moneromooo-monero:
It's probably a good idea to add a "regexp" field to the request, false by default, for backward compat. Also bump the minor RPC version.

reemuru:
@moneromooo-monero thank you for your review! I have squashed in the required changes and tested on a few account tags.

moneromooo-monero (approved):
The no_match line has a LOT of negatives, it'd gain from simplifying, and a new error code for "no error, just no matches" would be nice, but that's nothing that can't be done later so need not stop this patch.

PR #8220, opened 2022-03-17
multisig: add post-kex verification round

UkoeHB:
**Problem**: When creating a group multisig account, each step of key exchange relies on messages from all other participants. If a malicious player only sends their final message to one honest player, but not others, then the other honest players will be stuck without completed accounts. If that one honest player receives funds to the address he created, then those funds will either be unspendable or only spendable with cooperation from malicious group members.

**Solution**: Add a round to multisig account creation to verify that all other participants have completed their accounts.

This PR also has some semantic and control-flow cleanup of the multisig kex implementation (I am not satisfied with the readability of my earlier work - cleanup will continue in PR #8203 and alongside the Seraphis library which adds support for aggregation-style multisig signing).

**TODO**: In PR #8203 (or maybe a separate PR), I will add an `unsafe_update_use_with_care` flag to the `multisig_account::initialize_kex()` and `multisig_account::kex_update()` methods. This flag will allow users to fast-forward through the extra round added by this PR, and under some conditions even do so safely (without needing to trust other participants).

### DIFF
--- a/src/multisig/multisig_account.cpp
+++ b/src/multisig/multisig_account.cpp
@@ -73,7 +73,7 @@ namespace multisig
     const crypto::public_key &multisig_pubkey,
     const crypto::public_key &common_pubkey,
     const std::uint32_t kex_rounds_complete,
-    kex_origins_map_t kex_origins_map,
+    multisig_keyset_map_memsafe_t kex_origins_map,
     std::string next_round_kex_message) :
       m_base_privkey{base_privkey},
       m_base_common_privkey{base_common_privkey},
@@ -89,6 +89,20 @@ namespace multisig
     CHECK_AND_ASSERT_THROW_MES(crypto::secret_key_to_public_key(m_base_privkey, m_base_pubkey),
       "Failed to derive public key");
     set_multisig_config(threshold, std::move(signers));
+
+    // kex rounds should not exceed post-kex verification round
+    const std::uint32_t kex_rounds_required{multisig_kex_rounds_required(m_signers.size(), m_threshold)};
+    CHECK_AND_ASSERT_THROW_MES(m_kex_rounds_complete <= kex_rounds_required + 1,
+      "multisig account: tried to reconstruct account, but kex rounds complete counter is invalid.");
+
+    // once an account is done with kex, the 'next kex msg' is always the post-kex verification message
+    //   i.e. the multisig account pubkey signed by the signer's privkey AND the common pubkey
+    if (main_kex_rounds_done())
+    {
+      m_next_round_kex_message = multisig_kex_msg{kex_rounds_required + 1,
+        m_base_privkey,
+        std::vector<crypto::public_key>{m_multisig_pubkey, m_common_pubkey}}.get_msg();
+    }
   }
   //----------------------------------------------------------------------------------------------------------------------
   // multisig_account: EXTERNAL
@@ -100,14 +114,24 @@ namespace multisig
   //----------------------------------------------------------------------------------------------------------------------
   // multisig_account: EXTERNAL
   //----------------------------------------------------------------------------------------------------------------------
-  bool multisig_account::multisig_is_ready() const
+  bool multisig_account::main_kex_rounds_done() const
   {
     if (account_is_active())
-      return multisig_kex_rounds_required(m_signers.size(), m_threshold) == m_kex_rounds_complete;
+      return m_kex_rounds_complete >= multisig_kex_rounds_required(m_signers.size(), m_threshold);
     else
       return false;
   }
-    //----------------------------------------------------------------------------------------------------------------------
+  //----------------------------------------------------------------------------------------------------------------------
+  // multisig_account: EXTERNAL
+  //----------------------------------------------------------------------------------------------------------------------
+  bool multisig_account::multisig_is_ready() const
+  {
+    if (main_kex_rounds_done())
+      return m_kex_rounds_complete >= multisig_kex_rounds_required(m_signers.size(), m_threshold) + 1;
+    else
+      return false;
+  }
+  //----------------------------------------------------------------------------------------------------------------------
   // multisig_account: INTERNAL
   //----------------------------------------------------------------------------------------------------------------------
   void multisig_account::set_multisig_config(const std::size_t threshold, std::vector<crypto::public_key> signers)
@@ -119,10 +143,6 @@ namespace multisig
 
     for (auto signer_it = signers.begin(); signer_it != signers.end(); ++signer_it)
     {
-      // signers should all be unique
-      CHECK_AND_ASSERT_THROW_MES(std::find(signers.begin(), signer_it, *signer_it) == signer_it,
-        "multisig account: tried to set signers, but found a duplicate signer unexpectedly.");
-
       // signer pubkeys must be in main subgroup, and not identity
       CHECK_AND_ASSERT_THROW_MES(rct::isInMainSubgroup(rct::pk2rct(*signer_it)) && !(*signer_it == rct::rct2pk(rct::identity())),
         "multisig account: tried to set signers, but a signer pubkey is invalid.");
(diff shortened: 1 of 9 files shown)


src/multisig/multisig_account_kex_impl.cpp
```diff
@@ -429,6 +434,7 @@ namespace multisig
     }
 
     // sanitize input messages
+    multisig_keyset_map_memsafe_t pubkey_origins_map;
```
moneromooo-monero, line 437: This seems unneeded.

moneromooo-monero (commented):
I have no idea if it's OK algorithm wise, I just reviewed C code, and it's good.

src/multisig/multisig_account_kex_impl.cpp
UkoeHB: The definition of `pubkey_origins_map` in this function was removed, so this line adds it back in.

src/multisig/multisig_account.cpp
```diff
@@ -143,12 +153,11 @@ namespace multisig
...
-      [](const crypto::public_key &key1, const crypto::public_key &key2) -> bool
-        {
-          return memcmp(&key1, &key2, sizeof(crypto::public_key)) < 0;
-        }
-      );
+    std::sort(signers.begin(), signers.end());
+
+    // signers should all be unique
+    CHECK_AND_ASSERT_THROW_MES(std::adjacent_find(signers.begin(), signers.end()) == signers.end(),
+      "multisig account: tried to set signers, but there are duplicate signers unexpectedly.");
```
UkoeHB, line 160: Slight cleanup

rbrunner7:
As you predicted this needed some more small changes in `simplewallet.cpp` in order to work correctly. It was not only a question about the MMS; also the normal "manual" multisig workflow did not work correctly anymore. You can see the necessary changes here to copy them over into your code and update it: https://github.com/rbrunner7/monero/commit/fbb6055a60be6f74d25aeb9bcfe7e90f8a13397a

I ran my usual tests with your code plus my changes, using the MMS, for 2/2, 2/3 and 3/5 multisig. Everything - building wallets, receiving a tx, sending a tx - worked flawlessly. Of course with an almost fully automatic system like the MMS you hardly notice the additional data exchange round.

rbrunner7:
> To get around that blocking vector, we need a solution in the messaging layer. There are a few options.

I confess I don't fully understand everything you explain and propose here as options. But I brainstormed a little nevertheless how supporting such options could look in the MMS, and how much effort would be needed to implement.

I came up with an estimate of a few working days. It's not too hard, but not trivial either. It will probably need a handful of new commands, new logic within the `mms next` command and new message types.

With the low usage that the MMS sees and the corresponding low importance for the Monero ecosystem as a whole I don't see me implementing anything here in the near future, frankly.

rbrunner7 (approved):
For this PR it's the same as with the previous multisig-related one: I can't review the code, so this is not a proper code approval, but I did carefully test the code with 2/2, 2/3 and 3/5 multisig and can confirm that everything fully worked.

I think together with @moneromooo-monero 's look at the code from a technical point of view this should give something like a full review ...

UkoeHB:
@rbrunner7 thank you!

UkoeHB:
@rbrunner7 for MMS documentation, you could recommend that users manually verify with each other that their accounts are complete before sending any money to the multisig address. It's not automated like the other solutions, but would still work.

UkoeHB:
Update: the post-kex verification round must also check that all players have the same common pubkey (which is used as the view key by wallets).

src/multisig/multisig_account_kex_impl.cpp
```diff
@@ -536,43 +570,66 @@ namespace multisig
...
+    }
+    else if (current_round == kex_rounds_required)
+    {
+      // last kex round: collect the key shares recommended by other signers for the final aggregate key
+      keys_to_origins_map_out = std::move(evaluated_pubkeys);
+    }
+    else //(current_round == kex_rounds_required + 1)
+    {
+      // post-kex verification round: save the keys found in input messages
+      keys_to_origins_map_out = std::move(evaluated_pubkeys);
```
vtnerd, line 625: This is the same as the above branch, why not just combine them?

src/multisig/multisig_account_kex_impl.cpp
```diff
@@ -629,13 +686,29 @@ namespace multisig
...
+    multisig_keyset_map_memsafe_t result_keys_to_origins_map)
   {
+    std::vector<crypto::public_key> next_msg_keys;
+
     // prepare for next round (or complete the multisig account fully)
-    if (rounds_required == m_kex_rounds_complete + 1)
+    if (m_kex_rounds_complete == kex_rounds_required)
+    {
+      // post-kex verification round: check that the multisig pubkey and common pubkey were recommended by other signers
+      CHECK_AND_ASSERT_THROW_MES(result_keys_to_origins_map.find(m_multisig_pubkey) != result_keys_to_origins_map.end(),
```
vtnerd, line 698: I don't think its worth pushing a new update, but there is `result_keys_to_origins_map.count(m_multisig_pubkey)` which is slightly fewer characters.

src/multisig/multisig_account_kex_impl.cpp
```diff
@@ -679,48 +759,66 @@ namespace multisig
...
+
+      for (const auto &derivation_and_origins : result_keys_to_origins_map)
+        next_msg_keys.push_back(derivation_and_origins.first);
+
+      // save the account's kex keys for this round [DH derivation : other signers who should have the same derivation]
       m_kex_keys_to_origins_map = std::move(result_keys_to_origins_map);
     }
 
     // a full set of msgs has been collected and processed, so the 'round is complete'
     ++m_kex_rounds_complete;
```
vtnerd, line 785: Should there be a clamping on this? At a quick glance it looks like this will get incremented everytime `key_update_impl` is called, but this should be stopped after reaching +1 expected rounds.

vtnerd (approved):
My comments are minor.

src/multisig/multisig_account_kex_impl.cpp
vtnerd: Oh wait, this one might be worth visiting before merging ...

UkoeHB: This is checked upstream in `multisig_account::kex_update()` by checking if `!multisig_is_ready()`. However, it wouldn't hurt to add a redundant check in the internal calls.

src/multisig/multisig_account_kex_impl.cpp
UkoeHB: I thought it was easier to follow the control like this, but you're probably right.

src/multisig/multisig_account_kex_impl.cpp
UkoeHB: I don't mind it for this use-case.

vtnerd (approved):
A couple of nagging things, sorry :/ Only the one is interesting (the unused parameter in a function).

src/multisig/multisig_account.cpp
```diff
@@ -89,6 +89,20 @@ namespace multisig
...
+    // kex rounds should not exceed post-kex verification round
+    const std::uint32_t kex_rounds_required{multisig_kex_rounds_required(m_signers.size(), m_threshold)};
+    CHECK_AND_ASSERT_THROW_MES(m_kex_rounds_complete <= kex_rounds_required + 1,
+      "multisig account: tried to reconstruct account, but kex rounds complete counter is invalid.");
+
+    // once an account is done with kex, the 'next kex msg' is always the post-kex verification message
+    //   i.e. the multisig account pubkey signed by the signer's privkey AND the common pubkey
+    if (main_kex_rounds_done())
+    {
+      m_next_round_kex_message = multisig_kex_msg{kex_rounds_required + 1,
```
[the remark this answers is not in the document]
UkoeHB: Imo this would add a burden to the caller to take care that the field is empty when reconstructing an account (which is counter-intuitive). Overriding the field just ensures the correct message is available, no matter what the caller does.

src/multisig/multisig_account_kex_impl.cpp
```diff
@@ -281,27 +278,28 @@ namespace multisig
...
+  /**
+  * INTERNAL
+  *
+  * brief: check_messages_round - Check that a set of messages have an expected round number.
+  * param: expanded_msgs - set of multisig kex messages to process
+  * param: expected_round - round number the kex messages should have
+  */
+  //----------------------------------------------------------------------------------------------------------------------
+  static void check_messages_round(const std::vector<multisig_kex_msg> &expanded_msgs,
+    const std::uint32_t expected_round)
```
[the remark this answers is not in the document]
UkoeHB: Nice catch, thanks :) (missed this in the refactor)

vtnerd (approved):

UkoeHB:
Squashed

selsta (approved):

PR #8223, opened 2022-03-19
Eliminate dependence on boost::interprocess

jeffro256:
In this repo, `boost::interprocess` was being used soley to make `uint32_t` operations atomic. So I replaced each instance of
`boost::interprocess::ipcdetail::atomic(...)32` with `std::atomic` methods. I replaced member declarations as applicable. For example, when I needed to change a `volatile uint32_t` into a `std::atomic<uint32_t>`. Sometimes, a member was being used a boolean flag, so I replaced it with `std::atomic<bool>`.

You may notice that I didn't touch `levin_client_async.h`. That is because this file is entirely unused and will be deleted in PR #8211.

When I get more time, I will benchmark this to see if this changed the compile time in any manner.

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -384,7 +384,7 @@ PRAGMA_WARNING_DISABLE_VS(4355)
       {  
         //_info("[sock " << socket().native_handle() << "] protocol_want_close");
         //some error in protocol, protocol handler ask to close connection
-        boost::interprocess::ipcdetail::atomic_write32(&m_want_close_connection, 1);
+        m_want_close_connection = true;
         bool do_shutdown = false;
         CRITICAL_REGION_BEGIN(m_send_que_lock);
         if(!m_send_que.size())
@@ -478,7 +478,7 @@ PRAGMA_WARNING_DISABLE_VS(4355)
       if (!handshake(boost::asio::ssl::stream_base::server, boost::asio::const_buffer(buffer_.data(), buffer_ssl_init_fill)))
       {
         MERROR("SSL handshake failed");
-        boost::interprocess::ipcdetail::atomic_write32(&m_want_close_connection, 1);
+        m_want_close_connection = true;
         m_ready_to_close = true;
         bool do_shutdown = false;
         CRITICAL_REGION_BEGIN(m_send_que_lock);
@@ -834,7 +834,7 @@ PRAGMA_WARNING_DISABLE_VS(4355)
     CRITICAL_REGION_BEGIN(m_send_que_lock);
     send_que_size = m_send_que.size();
     CRITICAL_REGION_END();
-    boost::interprocess::ipcdetail::atomic_write32(&m_want_close_connection, 1);
+    m_want_close_connection = true;
     if(!send_que_size)
     {
       shutdown();
@@ -889,7 +889,7 @@ PRAGMA_WARNING_DISABLE_VS(4355)
     m_send_que.pop_front();
     if(m_send_que.empty())
     {
-      if(boost::interprocess::ipcdetail::atomic_read32(&m_want_close_connection))
+      if(m_want_close_connection)
       {
         do_shutdown = true;
       }
@@ -1119,7 +1119,7 @@ POP_WARNINGS
   bool boosted_tcp_server<t_protocol_handler>::worker_thread()
   {
     TRY_ENTRY();
-    uint32_t local_thr_index = boost::interprocess::ipcdetail::atomic_inc32(&m_thread_index); 
+    const uint32_t local_thr_index = m_thread_index++; // atomically increment, getting value before increment
     std::string thread_name = std::string("[") + m_thread_name_prefix;
     thread_name += boost::to_string(local_thr_index) + "]";
     MLOG_SET_THREAD_NAME(thread_name);
--- a/contrib/epee/include/net/levin_protocol_handler_async.h
+++ b/contrib/epee/include/net/levin_protocol_handler_async.h
@@ -28,7 +28,6 @@
 #include <boost/asio/deadline_timer.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/unordered_map.hpp>
-#include <boost/interprocess/detail/atomic.hpp>
 #include <boost/smart_ptr/make_shared.hpp>
 
 #include <atomic>
@@ -166,7 +165,7 @@ public:
   };
 
   std::atomic<bool> m_protocol_released;
-  volatile uint32_t m_invoke_buf_ready;
+  std::atomic<bool> m_invoke_buf_ready;
 
   volatile int m_invoke_result_code;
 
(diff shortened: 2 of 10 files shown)


contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -352,13 +351,13 @@ class async_protocol_handler
       MERROR(m_connection_context << "[levin_protocol] -->> start_outer_call failed");
       return false;
     }
-    boost::interprocess::ipcdetail::atomic_inc32(&m_wait_count);
+    m_wait_count++;
```
UkoeHB, line 354: Not sure if it needs to be changed but... I had to look up whether post-incrementing an atomic int had any performance implications (like it does for iterators). (it doesn't, but someone else might have the same question needlessly)

src/cryptonote_basic/miner.cpp
```diff
@@ -524,7 +523,7 @@ namespace cryptonote
   //-----------------------------------------------------------------------------------------------------
   bool miner::worker_thread()
   {
-    uint32_t th_local_index = boost::interprocess::ipcdetail::atomic_inc32(&m_thread_index);
+    uint32_t th_local_index = m_thread_index++;
```
UkoeHB, line 526:
```suggestion
    const uint32_t th_local_index = m_thread_index;
    ++m_thread_index;
```

UkoeHB (changes requested):

src/cryptonote_basic/miner.cpp
jeffro256: I thought about making the value `const` but it didn't match the rest of the code. Should I go ahead and do it anyways?

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1119,7 +1119,7 @@ POP_WARNINGS
   bool boosted_tcp_server<t_protocol_handler>::worker_thread()
   {
     TRY_ENTRY();
-    uint32_t local_thr_index = boost::interprocess::ipcdetail::atomic_inc32(&m_thread_index); 
+    uint32_t local_thr_index = m_thread_index++;
```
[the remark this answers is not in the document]
jeffro256: Good point, I'll put an inline comment

contrib/epee/include/net/levin_protocol_handler_async.h
jeffro256: Does postfix for iterators take a hit mainly because it has to copy construct itself? If so, it would make sense that `std::atomic<uint32_t>` wouldn't have that issue (except for single 4 byte `mov` instruction`, which is probably optimized away anyways)

UkoeHB: The main problem here is cognitive load with no functional benefit. Best to do the simplest thing possible, which is a pre-increment.

contrib/epee/include/net/abstract_tcp_server2.inl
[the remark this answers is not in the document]
jeffro256: @UkoeHB check out the new commit and lmk what you think. Thanks!

src/cryptonote_basic/miner.cpp
UkoeHB: If you're touching the line, might as well go all the way.

contrib/epee/include/net/abstract_tcp_server2.inl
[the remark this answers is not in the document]
UkoeHB: Yes that conveys your intent, thanks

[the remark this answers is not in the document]
UkoeHB: I just realized this operation has to be atomic, so post-increment is correct. My bad!

src/cryptonote_basic/miner.cpp
SChernykh: Same. It must be done in a single operation, not two lines of code!

moneromooo-monero:
Does this actually fix anything ?

moneromooo-monero:
I'm not going to block this since I'm not contributing much at all anymore, but I'm generally against pointless code churn. It possibly adds bugs, can add gratuitous conflicts, adds review load, complicates debugging/bisecting. There's no upside beyond a feeling that the author likes the new code better.

selsta:
I'm usually against code churn but I'm also happy with everything that gets rid of unnecessary boost dependencies, so personally I'd be fine with it.

jeffro256:
@moneromooo-monero That's a really good point, but I think that there is some value to this PR beyond it "feeling better". It completely eliminates all references to `boost::interprocess`. In every file I changed (besides `regexp_definer.h`), `std::atomic` was already being used, which increases code compile times, complexity, and could make readers confused as to why the different libraries were used in each instance, when in fact they were being used in a similar manner.

moneromooo-monero:
Changing volatile to atomic is a bug fix if it's meant to be atomic.

jeffro256:
@moneromooo-monero Here's a list of variables whose declaration I changed:

    abstract_tcp_server2::m_thread_index
    connection_basic::m_want_close_connection
    async_protocol_handler::m_invoke_buf_ready
    async_protocol_handler::m_wait_count
    async_protocol_handler::m_close_called
    blocked_mode_client::m_shutdowned
    regexp_initialized_1, 2, and 3 in regexpdefiner.h
    miner::m_thread_index
    
In all of these cases, these variables were being treated as atomic, only being accessed thru `boost::interprocess::ipcdetail::atomic[...]` functions. There is even a comment on line 396 of `abstract_tcp_server2.h` that says `// TODO change to std::atomic`. It seems that the previous authors had meant to make some of these atomic, but hadn't gotten around to it. Making these variables explicitly atomic will clarify to future readers that the variable should be atomic while actually enforcing it.

jeffro256:
Yes, it is atomic if used correctly and it is accessed solely through these functions. I don't see your point.

contrib/epee/include/net/levin_protocol_handler_async.h
jeffro256: Is a postfix increment harder to understand than a prefix increment?

UkoeHB: Doing more things = more things to think about, that's all.

moneromooo-monero:
Ah, did you mean something like "std::atomic ensures no bugs if someone forgets to use boost functions on a raw integer" ? If so, I guess it's a valid argument for the change.

jeffro256:
Yeah, this change isn't a bug fix but more of a "make code clearer and prevent future mistakes where one line is now not atomic and it messes everything else up" fix.

UkoeHB (approved):
I think pre-incrementing is cleaner, but that's ultimately a personal style choice. The PR is good.

jeffro256:
@UkoeHB does the latest commit look good to you?

UkoeHB (approved):
Ah, thanks for humoring me :)

jeffro256:
Last push squashes the previous 3 commits. How to reproduce my squash:

    git reset --soft master
    git commit

jeffro256:
@selsta this PR has been squashed

selsta (approved):

PR #8226, opened 2022-03-25
Fix typo: SERIALIZE_TYPE_DUOBLE

jeffro256:


reemuru (commented):
it's esperanto though 😂

mj-xmr (approved):
Reviewing this trivial PR was duoble, even for me.

tobtoht (approved):

selsta (commented):
I don't think this is written down anywhere but usually we don't merge changes like this to avoid:

- polluting `git blame` with unnecessary changes
- avoiding merge conflicts

Fixing code typos like this is fine when the code touched gets changed for other reasons anyway. At least that's how we have done it in the past.

jeffro256:
@selsta Yeah I understand if you don't want to merge it, but I figured it was pretty low hanging fruit and it isn't actually used yet so it shouldn't cause (too many) merge conflicts. Maybe it would be good to be proactive about it before someone wants to serialize doubles. Your call though, I'll close it if you want me to.

jeffro256:
Edit: I forgot to mention, the only reason I'm considering this is because its a well-exposed variable in a relatively prominent library file. And this chunk of code is very likely to be viewed by other developers since its related to serialization.

PR #8228, opened 2022-03-26
Remove dead code from parserse_base_utils and fix unit tests

jeffro256:
* Remove `match_string()`, `match_number()`, and `match_word()`
* Remove `match_word_with_extrasymb()` and `match_word_til_equal_mark()`
* Adapt unit test for `match_number()` to `match_number2()`
* Adapt unit test for `match_string()` to `match_string2()`

Note: the unit tests were testing for the old version of the functions, and
the interfaces for these functions changed slightly, so I had to also edit
the tests.

As of writing, this PR has no merge conflicts with #8211

### DIFF
--- a/tests/unit_tests/epee_utils.cpp
+++ b/tests/unit_tests/epee_utils.cpp
@@ -1748,68 +1748,44 @@ TEST(parsing, isdigit)
 
 TEST(parsing, number)
 {
-  boost::string_ref val;
-  std::string s;
-  std::string::const_iterator i;
+  struct match_number_test_data {
+    std::string in_str;
+    std::string expect_out_str;
+    bool expect_is_float;
+    bool expect_is_signed;
+  };
+
+  // Add test cases as needed
+  struct match_number_test_data test_data[] = {
+    {         "0 ",         "0", false, false },
+    {       "000 ",       "000", false, false },
+    {        "10x",        "10", false, false },
+    {     "10.09/",     "10.09",  true, false },
+    {       "-1.r",       "-1.",  true,  true },
+    {      "-49.;",      "-49.",  true,  true },
+    {      "0.78/",      "0.78",  true, false },
+    {      "33E9$",      "33E9",  true, false },
+    {     ".34e2=",     ".34e2",  true, false },
+    {  "-9.34e-2=",  "-9.34e-2",  true,  true },
+    { "+9.34e+03=", "+9.34e+03",  true, false }
+  };
 
   // the parser expects another character to end the number, and accepts things
   // that aren't numbers, as it's meant as a pre-filter for strto* functions,
   // so we just check that numbers get accepted, but don't test non numbers
+  // We set is_float/signed_val to the opposite of what we expect the result to
+  // make sure that match_number2 is changing the bools as expected
 
-  s = "0 ";
-  i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "0");
-
-  s = "000 ";
-  i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "000");
-
-  s = "10x";
-  i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "10");
-
-  s = "10.09/";
-  i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "10.09");
-
-  s = "-1.r";
-  i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "-1.");
-
-  s = "-49.;";
-  i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "-49.");
-
-  s = "0.78/";
-  i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "0.78");
-
-  s = "33E9$";
-  i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "33E9");
-
-  s = ".34e2=";
-  i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, ".34e2");
-
-  s = "-9.34e-2=";
-  i = s.begin();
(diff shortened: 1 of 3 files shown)


tests/unit_tests/epee_utils.cpp
```diff
@@ -1818,13 +1842,44 @@ TEST(parsing, unicode)
   std::string s;
   std::string::const_iterator si;
```
UkoeHB, line 1843: This `si` variable seems unnecessary (just one more odd thing from this ancient code). You could replace the `si` parameter passing with `s.begin()`. The same goes for `i` in the other unit test.

UkoeHB (changes requested):

tests/unit_tests/epee_utils.cpp
UkoeHB: Ok maybe not... this function parameter is a non-const l-value reference, which can't be initialized from an r-value... sigh. One example of why code quality is important: always make non-modifiable non-sink function parameters const.

UkoeHB (approved):

tests/unit_tests/epee_utils.cpp
```diff
@@ -1751,65 +1751,89 @@ TEST(parsing, number)
   boost::string_ref val;
   std::string s;
   std::string::const_iterator i;
+  bool is_float_val;
```
mj-xmr, line 1754: it's a bad idea to leave these 2 unassigned at declaration. It will behave deterministically in the debug mode (will be assigned to `false`), but nondeterministically in the release mode.

Since the expected result is that these become `false`, I suggest that they be assigned `true` at their declaration..

tests/unit_tests/epee_utils.cpp
```diff
@@ -1751,65 +1751,89 @@ TEST(parsing, number)
...
+  epee::misc_utils::parse::match_number2(i, s.end(), val, is_float_val, is_signed_val);
+  EXPECT_EQ(val, "0");
+  EXPECT_FALSE(is_float_val);
+  EXPECT_FALSE(is_signed_val);
 
   s = "000 ";
   i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "000");
+  epee::misc_utils::parse::match_number2(i, s.end(), val, is_float_val, is_signed_val);
```
mj-xmr, line 1770: ... and reassigned to `true` before this line:

`is_float_val = is_float_val = true`

tests/unit_tests/epee_utils.cpp
```diff
@@ -1751,65 +1751,89 @@ TEST(parsing, number)
...
+  EXPECT_EQ(val, "000");
+  EXPECT_FALSE(is_float_val);
+  EXPECT_FALSE(is_signed_val);
 
   s = "10x";
   i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "10");
+  epee::misc_utils::parse::match_number2(i, s.end(), val, is_float_val, is_signed_val);
```
mj-xmr, line 1777: same here

tests/unit_tests/epee_utils.cpp
```diff
@@ -1751,65 +1751,89 @@ TEST(parsing, number)
...
+  EXPECT_EQ(val, "10");
+  EXPECT_FALSE(is_float_val);
+  EXPECT_FALSE(is_signed_val);
 
   s = "10.09/";
   i = s.begin();
-  epee::misc_utils::parse::match_number(i, s.end(), val);
-  ASSERT_EQ(val, "10.09");
+  epee::misc_utils::parse::match_number2(i, s.end(), val, is_float_val, is_signed_val);
```
mj-xmr, line 1784: and here obviously:

is_float_val = false; is_signed_val = true`

and so on...

jeffro256:
@mj-xmr @UkoeHB  Thanks for reviewing! @mj-xmr Would you like to check out the last commit?

mj-xmr:
@jeffro256 It looks OK how you set the booleans, but I'd still ask to to initialize them at the declaration to `false`.

selsta:
Seems a bit unnecessary to set something to false only to set it to true in the next line? :)

mj-xmr:
> Seems a bit unnecessary to set something to false only to set it to true in the next line? :)
> 
> not that it matters anyway

- It's a bad habit
- It's new code
- it generates warnings

UkoeHB:
I was taught to default initialize when the value won't be used:
`bool is_float_val{};`

mj-xmr:
> I was taught to default initialize when the value won't be used: `bool is_float_val{};`

Right. We're on cpp11 as the min. standard already

selsta (approved):

tests/unit_tests/epee_utils.cpp
```diff
@@ -1757,79 +1757,103 @@ TEST(parsing, number)
   // the parser expects another character to end the number, and accepts things
   // that aren't numbers, as it's meant as a pre-filter for strto* functions,
   // so we just check that numbers get accepted, but don't test non numbers
+  // We set is_float/signed_val to the opposite of what we expect the result to
+  // make sure that match_number2 is changing the bools as expected
 
   s = "0 ";
   i = s.begin();
+  is_float_val = true;
```
iamamyth, line 1765: I think it would be more readable and less error-prone to split match_number and match_number2 into separate tests and structure the `number2` cases as a loop over a sequence of "test scenario" structs consisting of the input and the expected out values. This way, you clarify how to exercise the code under test (preconditions and postconditions), have clear rules for state initialization, and it's trivial to add tests as warranted. The test assertions can include the input value as the message in order to make the failure clear, rather than using line numbers.

I realize that perhaps you ended up with copy/paste because that was the original design, but at this point the refactor creates *less* noise because the original code was much briefer and somewhat incorrect, so applying the corrections in a single location makes the review simpler, as opposed to copy/pasting many lines into many places.

jeffro256: @iamamyth I'm deleting `match_number()`, as it is unused, so this should be the only unit test for `parsing.number`. Looping through a list of structs is good idea though, I think I'll do that

jeffro256:
@mj-xmr @UkoeHB @iamamyth I should have addressed all these concerns with the latest commit

mj-xmr (approved):
The struct is a fine concept.

jeffro256:
Last force-push was to squash previous 3 commits into 1. Commands used to squash:

    git reset --soft master
    git commit
    git push --force

jeffro256:
@selsta this PR is also now squashed

selsta (approved):

mj-xmr (approved):

iamamyth (approved):

PR #8232, opened 2022-03-31
Added support for Ledger Nano S Plus

hbs:
The [Ledger nano S plus](https://shop.ledger.com/pages/ledger-nano-s-plus?r=00d147b67b6d) has a new Product ID, this PR adds it to the list of known IDs so the Nano S Plus can be used.

### DIFF
--- a/src/device/device_ledger.cpp
+++ b/src/device/device_ledger.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2020, The Monero Project
+// Copyright (c) 2017-2022, The Monero Project
 // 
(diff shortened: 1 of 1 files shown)


mj-xmr:
Could you point to the exact documentation, where this product ID is mentioned?

hbs:
The product ID is mentioned here (Ledger Test was the S+):

https://github.com/LedgerHQ/udev-rules/blob/master/20-hw1.rules#L14

selsta (approved):

mj-xmr (approved):
Thanks

PR #8235, opened 2022-04-02
CMake: Add missing BUILD_SHARED_LIBS option

mj-xmr:
The option `BUILD_SHARED_LIBS` was inaccessible to GUI CMake's wrappers. One such GUI wrapper is used by VSCodium, and is the only way to change the variables from that IDE. VSCodium was the most popular IDE, as [a result of my Reddit poll](https://www.reddit.com/r/Monero/comments/t9iy3b/poll_what_ide_would_you_like_to_have_documented/). The option is required to minimize linking time when practicing test driven development, [as described here](https://github.com/monero-project/monero/blob/master/docs/COMPILING_DEBUGGING_TESTING.md#test-driven-development-tdd---shared-libraries-for-release-builds).

In order not to break the traditional flow, the option is only advertised after the line 476:
```CMake
if (NOT STATIC AND CMAKE_BUILD_TYPE_LOWER STREQUAL "debug")
```
and only given distinct default values for each of the 2 cases.

The practical impact of this change is, that whenever the debug version is selected via CMake, the linking will still produce shared libraries by default, but the variable's value still be changed via CMake's GUIs, that work in-source (VSCode / VSCodium). Without it, building the release version under these in-source IDEs forces the build to be statically linked. Out-of-source IDEs, like Code::Blocks, are able to get around it without this PR.

## Measurements

| Directory |  Static    |      Dynamic      | Change |
|-------------|----------|-------------|-------------|
| total  |   700M  | 434M  |  -38.04% |
| bin   |   121M  | 15M  | -88.24%  |

As you can see, quite a lot was achieved on the total directory, but even more in the `bin` directory, which will serve as a regularizer for future, should more executables be created.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -472,9 +472,11 @@ endif()
 option(STATIC "Link libraries statically" ${DEFAULT_STATIC})
 
 # This is a CMake built-in switch that concerns internal libraries
-if (NOT DEFINED BUILD_SHARED_LIBS AND NOT STATIC AND CMAKE_BUILD_TYPE_LOWER STREQUAL "debug")
-    set(BUILD_SHARED_LIBS ON)
+set(BUILD_SHARED_LIBS_DEFAULT OFF)
+if (NOT STATIC AND CMAKE_BUILD_TYPE_LOWER STREQUAL "debug")
+	set(BUILD_SHARED_LIBS_DEFAULT ON)
 endif()
+option(BUILD_SHARED_LIBS "Build internal libraries as shared" ${BUILD_SHARED_LIBS_DEFAULT})
 
 if (BUILD_SHARED_LIBS)
   message(STATUS "Building internal libraries with position independent code")


CMakeLists.txt
```diff
@@ -472,8 +472,11 @@ endif()
 option(STATIC "Link libraries statically" ${DEFAULT_STATIC})
 
 # This is a CMake built-in switch that concerns internal libraries
-if (NOT DEFINED BUILD_SHARED_LIBS AND NOT STATIC AND CMAKE_BUILD_TYPE_LOWER STREQUAL "debug")
-    set(BUILD_SHARED_LIBS ON)
+set(BUILD_SHARED_LIBS_HELP_TEXT "Build internal libraries as shared")
+if (NOT STATIC AND CMAKE_BUILD_TYPE_LOWER STREQUAL "debug")
+    option(BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS_HELP_TEXT} ON)  # ON for debug, otherwise the build consumes way too much disk space.
```
selsta, line 477: Correct me if I'm wrong but it's not on for static debug, I would edit the comment otherwise the code and the comment don't match.

selsta, line 477: Correct me if I'm wrong but it's not on for static debug, I would edit the comment otherwise the code and the comment don't match.

mj-xmr: You are right, as long as the external libraries are concerned.

Something like this then?

`# ON for debug & shared external libraries, otherwise the build consumes way too much disk space.`

selsta: Did you measure the space reduction for debug build with `BUILD_SHARED_LIBS` on and off?

selsta: Why not just remove all the comments apart from `# This is closest to the final product` ? Unless you tested the storage difference then you can add `# Otherwise the build consumes way too much disk space`

Don't write anything about `ON for debug & shared external libraries` and `OFF by default`, everyone can read the if anyway...

selsta: In the end you could just add a single line but that's just my preference.

```
option(BUILD_SHARED_LIBS "Build internal libraries as shared" NOT STATIC AND CMAKE_BUILD_TYPE_LOWER STREQUAL "debug"} 
```

jeffro256:
@mj-xmr Is there a way you recommend others to test this functionality on different platforms?

selsta:
The functionality existed already, this is just for registering as a proper option and adding a help text.

mj-xmr:
> The functionality existed already, this is just for registering it as a proper option and adding a help text.

I believe @jeffro256 is asking about how to enable it to see the result. This option is explained exactly [in this paragraph](https://github.com/monero-project/monero/blob/master/docs/COMPILING_DEBUGGING_TESTING.md#test-driven-development-tdd---shared-libraries-for-release-builds), where the option is being *forced* by the `cmake` command.

The current PR enables `ccmake` and `cmake-gui` to actually *see* the option and show it to the developer, finally allowing to switch it interactively. I hope this explains this.

CMakeLists.txt
mj-xmr: > In the end you could just add a single line but that's just my preference.
> 
> ```
> option(BUILD_SHARED_LIBS "Build internal libraries as shared" NOT STATIC AND CMAKE_BUILD_TYPE_LOWER STREQUAL "debug"} 
> ```

No prob. Will do it so.

mj-xmr: > Did you measure the space reduction for debug build with `BUILD_SHARED_LIBS` on and off?

I did some time ago, but will do it again to see fresher results.

mj-xmr:
@jeffro256 A little bit more context is due:

Since I'm expanding the documentation to include VSCodium, this change is necessary, since the IDE doesn't allow you to change the parameters, like you can do with CLI `cmake`, only the `option`s like the one I introduce here.

jeffro256:
@mj-xmr Ah, that makes more sense

CMakeLists.txt
mj-xmr: > In the end you could just add a single line but that's just my preference.
> 
> ```
> option(BUILD_SHARED_LIBS "Build internal libraries as shared" NOT STATIC AND CMAKE_BUILD_TYPE_LOWER STREQUAL "debug"} 
> ```

According to [this post from 2020](https://stackoverflow.com/questions/62487808/can-i-set-a-cmake-variable-to-the-result-of-a-boolean-expression), one-liners aren't possible in CMake.

mj-xmr: > Did you measure the space reduction for debug build with `BUILD_SHARED_LIBS` on and off?

Measurements are now placed in the description. almost 40% reduction totally.

selsta:
The second commit looks nicer and seems to be as small as CMake allows it apparently.

mj-xmr:
> The second commit looks nicer and seems to be as small as CMake allows it apparently. Could you remove the first?

Done.

selsta (approved):

Issue #8234, opened 2022-04-02
Compilation failure of v0.17.3.0 with OpenBSD 7 on RiscV-64


antanst:
System is an Unmatched HiFive RISCV-64 running OpenBSD 7.1-beta.

```shell
$ uname -a
OpenBSD hi5.lan 7.1 GENERIC.MP#66 riscv64
$ clang -v
OpenBSD clang version 13.0.0
Target: riscv64-unknown-openbsd7.1
```

Monero is built according to README instructions and fails with this error:

```
$ env DEVELOPER_LOCAL_TOOLS=1 BOOST_ROOT=/usr/local gmake release-static 
mkdir -p build/"OpenBSD/_HEAD_detached_at_v0.17.3.0_"/release
cd build/"OpenBSD/_HEAD_detached_at_v0.17.3.0_"/release && cmake -D STATIC=ON -D ARCH="x86-64" -D BUILD_64=ON -D CMAKE_BUILD_TYPE=release ../../../.. && gmake
....
<snip>
...
gmake[3]: Leaving directory '/home/user/work/monero/build/OpenBSD/_HEAD_detached_at_v0.17.3.0_/release'
[ 21%] Building C object external/unbound/CMakeFiles/unbound.dir/sldns/keyraw.c.o
gmake[3]: Entering directory '/home/user/work/monero/build/OpenBSD/_HEAD_detached_at_v0.17.3.0_/release'
[ 21%] Built target genversion                                                                                                                                                                 
[ 21%] Building C object src/crypto/CMakeFiles/obj_cncrypto.dir/aesb.c.o
[ 21%] Built target qrcodegen
[ 21%] Building C object src/crypto/CMakeFiles/obj_cncrypto.dir/blake256.c.o        
[ 21%] Building C object external/unbound/CMakeFiles/unbound.dir/sldns/parseutil.c.o
cc: error: invalid arch name 'x86-64', string must begin with rv32{i,e,g} or rv64{i,g}
gmake[3]: *** [src/crypto/CMakeFiles/obj_cncrypto.dir/build.make:76: src/crypto/CMakeFiles/obj_cncrypto.dir/aesb.c.o] Error 1
gmake[3]: *** Waiting for unfinished jobs....
cc: error: invalid arch name 'x86-64', string must begin with rv32{i,e,g} or rv64{i,g}
[...]

PR #8240, opened 2022-04-06
randomx: update submodule

selsta:
#8234


mj-xmr (approved):
I checked the failing functional test, suspecting it could have sth. to do with RX (the mining test), but it's the integrated address test that failed, so unrelated.

PR #8248, opened 2022-04-10
Remove serialization/enableable

jeffro256:
Currently working on an EPEE [ser/de]ialization library for Rust and at first glance, EPEE seemed to have support for optional wrappers. However, after looking into it, this feature appears to be half-baked and unused. Furthermore, adding support for optional values would be better suited to implement at the storage level, in my opinion. That would make parsing DOMs cleaner and less error-prone. If anyone is currently using this code, please comment. Thanks!

At the time of writing, this PR has no merge conflicts with #8211


vtnerd (approved):
If done again from scratch, it would probably involve `boost::optional` or similar somehow. I don't see a reason to keep this.

PR #8250, opened 2022-04-10
Remove src/platform

jeffro256:
All the files in `src/platform` are currently unused and/or unnecessary:

* `mingw/alloca.h`: unused throughout project
* `msc/sys/param.h`: *Justifications by file below*
    1. `fix_darwin.patch`: `sys/param.h` is well-defined to be included only in OpenBSD environment
    2. `int-util.h` already handles when `sys/param.h` is not present and injects its own values
    3. `db_drivers/liblmdb/mdb.c` is a similar situation: already explicitly handles different platforms
    4. `src/crypto/chacha.h` uses `int-util.h` for endianness context
* `msc/alloca.h`: unused
* `msc/inline_c.h`: unused and not needed. the `inline` keyword is supported for C99 I believe, and certainly for C11, C14+
* `msc/stdbool.h`: `stdbool.h` is standard since C99, and MSVC has supported it for a long time (~10 years)


jeffro256:
> Moreover, there are other files in external/miniupnp/miniupnp/ that include sys/param.h that should be looked at

Do you have any specific examples of things that may cause problems?

> Also, after those removals, are there any unused #includes that could be removed?

I don't think so, most of it was unused anyways.

jeffro256:
@tobtoht

tobtoht (approved):

PR #8258, opened 2022-04-13
simplewallet: transfer description fix

hinto-janaiyo:
The `help transfer` text is reverse from the actual `transfer` command syntax

Before:
```
Command description: 
  Transfer <amount> to <address> ...
```
After:
```
Command description: 
Transfer to <address> this <amount> ...
```
`Transfer <amount> to <address>` makes more intuitive sense to me but I'm not sure which lines in `wallet2` or `simplewallet` to change to actually swap the argument positions


hinto-janaiyo:
Hmm... you're right, I completely overlooked the command usage section.

Closing PR, don't think it's actually a "fix"

PR #8261, opened 2022-04-16
Fix languages list being defined twice and add language validation on wallet creation

eversinc33:
Hi. First contribution here, so please bear with me, since I have a question regarding my first commit.

In the first commit I changed `electrum-words.cpp`, since the list of languages was defined in two separate places, which has a potential for introducint future bugs, if only one list is updated. Here, the `get_language_list`-function should be used. This list was also missing the "OldEnglish" variant, which is deprecated, but should IMO be kept here for backwards compatibility. However, there is one thing where I am not sure on how to handle and would appreciate a comment: In `simple_wallet::get_mnemonic_language`, `get_language_list` is called, which would now show the deprecated "OldEnglish" as an option. Should I add code here to remove the "OldEnglish" language from the list when being polled here? 

In the second commit I removed two TODOs by adding validation of the language when creating a wallet.


eversinc33:
Force-pushed to squash my commits into one.

src/mnemonics/electrum-words.cpp
```diff
@@ -91,29 +91,17 @@ namespace
   bool find_seed_language(const std::vector<epee::wipeable_string> &seed,
     bool has_checksum, std::vector<uint32_t> &matched_indices, Language::Base **language)
   {
-    // If there's a new language added, add an instance of it here.
-    std::vector<Language::Base*> language_instances({
-      Language::Singleton<Language::Chinese_Simplified>::instance(),
-      Language::Singleton<Language::English>::instance(),
-      Language::Singleton<Language::Dutch>::instance(),
-      Language::Singleton<Language::French>::instance(),
-      Language::Singleton<Language::Spanish>::instance(),
```
iamamyth, line 100: The order here differs slightly from the one in `get_language_list`, which could theoretically change the behavior of this function in case of languages with shared words.

eversinc33: That's right, although I do believe that this should not be an issue, as long as OldEnglish comes after English?

iamamyth: I think it's improbable, but possible, that it could break behavior for someone. What happens if you make the canonical ordering the one used by the restore list, so as not to change the restoration behavior?

eversinc33: That would be possible, but then this *could* change the behaviour of all the other function that use `get_language_list`.

iamamyth: I appreciate that, absent further analysis, modifying get_language_list *could* break its callers, but that seems like a question that can be determined empirically, and possibly corrected, as well, whereas the callers of "restore wallet" are end-users who already have created wallets. That said, if the language selection in restore wallet were already ambiguous (apart from the case of "old english"), that would present a problem irrespective of this change. When choosing a particular path, it helps to clarify the rationale for the choice, and the potential problems, so that reviewers can assess risk and match user reports to issues considered during review.

eversinc33: I also checked each use of `get_language_list` and don't see any use where the order could be important.

PR #8264, opened 2022-04-18
Tests: libwallet_api_tests fixing scripts

mj-xmr:
As a first step towards fixing the `libwallet_api_tests`, I updated the relevant scripts. I based my work on @MartyHav's [commits](https://github.com/haven-protocol-org/haven-main/commits/master/tests/libwallet_api_tests), but extended them further. Also updated documentation.

### DIFF
--- a/tests/libwallet_api_tests/scripts/README.md
+++ b/tests/libwallet_api_tests/scripts/README.md
@@ -5,6 +5,11 @@
   By default, tests expect daemon running at ```localhost:38081```,
   can be overridden with environment variable ```TESTNET_DAEMON_ADDRESS=<your_daemon_address>```
   [Manual](https://github.com/moneroexamples/private-testnet) explaining how to run private testnet.
+  It is benefitial to run the node with the `--disable-rpc-ban` option, because the test will be abusing the node.
+
+* Running monero node, linked to mainnet.
+  By default, tests expect daemon running at ```localhost:18081```,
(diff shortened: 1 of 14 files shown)


jeffro256:
I'm guessing that there's some historical reason for the mismatch, but shouldn't the default testnet port be 28081, not 38081?. This could possible be confusing for someone running these tests

jeffro256:
Also, what do you mean by "fixing" in this instance? Besides that, I really like how much cleaner the scripts are.

mj-xmr:
> Also, what do you mean by "fixing" in this instance? Besides that, I really like how much cleaner the scripts are.

The way they were written, using them lead to segfault of the tests.

mj-xmr:
> I'm guessing that there's some historical reason for the mismatch, but shouldn't the default testnet port be 28081, not 38081?. This could possible be confusing for someone running these tests

I'll have a closer look and answer later.

mj-xmr:
> I'm guessing that there's some historical reason for the mismatch, but shouldn't the default testnet port be 28081, not 38081?. This could possible be confusing for someone running these tests

Here are the test's defaults:

https://github.com/monero-project/monero/blob/master/tests/libwallet_api_tests/main.cpp#L90-L91

```C++
std::string TESTNET_DAEMON_ADDRESS = "localhost:38081";
std::string MAINNET_DAEMON_ADDRESS = "localhost:18081";
```

If the documentation / scripts are meant to describe the current state, then I think it's fine. Should there be changes to these ports, then they'd have to be widely communicated first and documented in other HOWTOs across the space.

PR #8272, opened 2022-04-20
Replace in-house MD5 with OpenSSL

jeffro256:
### Pros:
* ~4x speed up for MD5 hashes
* ~800 less lines of code
* Marginally smaller build sizes

### Cons:
* Less control over implementation
* According to @hyc, there is a chance that some OpenSSL implementations will not support MD5

### Specs:

    Ubuntu 20.04
    Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    31 GB RAM

### Hashing 1M strings 1000 characters long:

    Hashing using OpenSSL Implementation:  2651850801 ns
    Hashing using EPEE Implementation:    11596632481 ns
    Speedup: 4.37x

### Hashing 1M strings 100 characters long:

    Hashing using OpenSSL Implementation:  550373610 ns
    Hashing using EPEE Implementation:    2146028452 ns
    Speedup: 3.90x

## Conclusions:
This could help boost wallet sync times where many RPC requests are made and authenticated over a short period of time.


jeffro256:
[Here](https://pastebin.com/KRVcyMh8) is my benchmark code. To compile you need the following files:
* `memwipe.h` and `memwipe.c`
* `md5_l.h` and `md5_l.inl`
* `md5global.h`
* `hmac-md5.h`
* And obviously the OpenSSL libraries

hyc:
Introducing this kind of dependency is a bit dicy, since MD5 support is optional in OpenSSL, and since MD5 has been deprecated in favor of SHA for a few years already, it's not uncommon for sites to omit it from their OpenSSL builds.

Authentication traffic shouldn't be a large percentage of a wallet RPC session, so how much does this really save in practice?

jeffro256:
> it's not uncommon for sites to omit it from their OpenSSL builds

Not trying to be contrarian, but do you have any examples of builds of OpenSSL for platforms we support which omit MD5?

>  since MD5 has been deprecated in favor of SHA for a few years already, it's not uncommon for sites to omit it from their OpenSSL builds.

For this reason, shouldn't we be moving away from MD5 digest authentication anyways? There are a ton of more secure authentication protocols with nearly as much, if not more, support than MD5 Digest Authentication. I'm hoping that this is not permanent, it would make happier the quicker we ditch MD5 authentication, or at least require digest authentication to happen over SSL.

> Authentication traffic shouldn't be a large percentage of a wallet RPC session, so how much does this really save in practice?

I haven't bench-marked that yet, but it's probably not *too* much. What do you think would be the most realistic way to stress test this?

hyc:
> Not trying to be contrarian, but do you have any examples of builds of OpenSSL for platforms we support which omit MD5?

No, because as I said it's about sites, not platforms.

> I haven't bench-marked that yet, but it's probably not too much. What do you think would be the most realistic way to stress test this?

Run a private node, sync'd but disconnected from mainnet, and aim a wallet at it. If you use monero-wallet-rpc you ought to be able to automate the entire sequence to eliminate run variations.

jeffro256:
Closing this because MD5 is deprecated since OpenSSL 3.0, and like @hyc said, some people don't build their sites with MD5.

PR #8275, opened 2022-04-21
add a sanity check to RPC input data size

moneromooo-monero:
reported by m31007


selsta (approved):

src/rpc/core_rpc_server.cpp
```diff
@@ -365,6 +365,8 @@ namespace cryptonote
       std::move(rpc_config->access_control_origins), std::move(http_login), std::move(rpc_config->ssl_options)
     );
 
+    m_net_server.get_config_object().m_max_content_length = MAX_RPC_CONTENT_LENGTH;
```
web-sharp, line 368: should this test the config before going straight to main net?

PR #8277, opened 2022-04-21
Preserve commitment format inside transactions

kayabaNerve:
A change to the format commitment / 8 was included with BP+, which is a minor optimization which increased complexity, spread out where crypto code is handled, and created a disparate state between transactions and disk (as they were saved in their full form thanks to a processing step inside the DB code).

This PR reverts this planned change. The blockchain_db.cpp edits may look to be further edited than just this reversion, yet that was its code pre-BP+.

### DIFF
--- a/src/blockchain_db/blockchain_db.cpp
+++ b/src/blockchain_db/blockchain_db.cpp
@@ -241,15 +241,8 @@ void BlockchainDB::add_transaction(const crypto::hash& blk_hash, const std::pair
     }
     else
     {
-      rct::key commitment;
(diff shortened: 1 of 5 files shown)


UkoeHB (approved):
LGTM. Every change here lines up 1:1 with conditionals introduced by #7170 (BP+).

moneromooo-monero (approved):
Technically one hunk's missing from the revert (const bool plus), but it doesn't seem needed for now.

It'd be nice to see the verification speed change this makes though. I don't remember what it was.

Issue #8167, opened 2022-02-02
Outgoing transfer counting change as incoming payment 


mrtestyboy781:
# Issue
On two rare occasions (less than .01% of all payments) we have seen the wallet create a `transfer` where the change is counted as an incoming payment and is visible in `get_bulk_payments`, `get_transfers`, and `get_transfer_by_txid` as `type="in"`

The change itself is using a payment id we haven't issued before, so it looks like an erroneous deposit to our wallet, when instead it's just change, and would normally not show up as an incoming payment.  

The payment itself constructed in a view-only / offline setup, where `transfer` is used to create the unsigned tx, its send to offline wallet to `sign_transfer` and broadcast later with `submit_transfer` 

Monerod Version : v0.17.1.9

# Details

### Expected output
In a normal outgoing transfer, you would see `get_transfer_by_txid` return
```
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "transfer": {
      "address": "OUTGOING_ADDRESS",
      "amount": 1171780000000,
      "confirmations": 11,
      "destinations": [{
        "address": "OUTGOING_ADDRESS",
        "amount": 1171780000000
      }],
      "double_spend_seen": false,
      "fee": 6930000,
      "height": 2550596,
      "locked": false,
      "note": "",
      "payment_id": "XXXXXXXXXXX",
      "subaddr_index": {
        "major": 0,
        "minor": 0
      },
      "subaddr_indices": [{
        "major": 0,
        "minor": 0
      }],
      "suggested_confirmations_threshold": 2,
      "timestamp": 1643794568,
      "txid": "XXXXXXXXXXXX",
      "type": "out",
      "unlock_time": 0
    },
    "transfers": [{
      "address": "OUTGOING_ADDRESS",
      "amount": 1171780000000,
      "confirmations": 11,
      "destinations": [{
        "address": "OUTGOING_ADDRESS",
        "amount": 1171780000000
      }],
      "double_spend_seen": false,
      "fee": 6930000,
      "height": 2550596,
      "locked": false,
      "note": "",
      "payment_id": "XXXXXX",
      "subaddr_index": {
        "major": 0,
        "minor": 0
      },
      "subaddr_indices": [{
        "major": 0,
        "minor": 0
      }],
      "suggested_confirmations_threshold": 2,
      "timestamp": 1643794568,
      "txid": "XXXXXXX",
      "type": "out",
      "unlock_time": 0
    }]
  }
}
``` 
[...]

PR #8285, opened 2022-04-23
wallet2: Erase payments for self-spends or change when importing key images

j-berman:
When self-spending to a subaddress, a view-only wallet that had not already imported key images prior to scanning the tx will misinterpret the change in the tx as a "payment". This PR ensures that change and self-spends to the spending account are removed from payments upon importing key images.

Related to (and hopefully fixes) #8167

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -12519,6 +12519,37 @@ std::pair<uint64_t, std::vector<std::pair<crypto::key_image, crypto::signature>>
   return std::make_pair(offset, ski);
 }
 
+/**
+ *
+ * Erase corresponding incoming payments to the same subaddress account that is
+ * spending in a tx. "Payments" are supposed to be incoming payments from other
+ * users, or from the same user but different account. However, when scanning
+ * with just a view key and without either a spend key or knowing key images for
+ * received outputs (e.g. in a watch-only wallet that has not yet imported key
+ * images), when the scanner picks up a received output in a tx, it won't know
+ * if this received output is actually change or a self-spend, so it will look
+ * like an incoming "payment", and will incorrectly place it in the payments
+ * container. This function erases payments that are actually change, or are
+ * self-spends to the same account.
+ *
+ */
+void erase_incoming_payments_for_spent_tx(wallet2::payment_container &payments, const crypto::hash &spent_txid, uint32_t spending_subaddr_account)
+{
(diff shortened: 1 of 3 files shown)


tests/functional_tests/cold_signing.py
```diff
@@ -160,6 +166,25 @@ def transfer(self):
...
 
+        self.hot_wallet.rescan_blockchain(hard = True)
+
+        res = self.hot_wallet.export_outputs(all_ = True)
+        self.cold_wallet.import_outputs(res.outputs_data_hex)
+        res = self.cold_wallet.export_key_images(all_ = True)
+        self.hot_wallet.import_key_images(res.signed_key_images, offset = res.offset)
+
+        res = self.hot_wallet.get_transfers()
+        assert len([x for x in (res['in'] if 'in' in res else []) if x.txid == txid]) == 0
```
j-berman, line 177: Of note, this is the only assertion that would fail without the change to wallet2

src/wallet/wallet2.cpp
```diff
@@ -12519,6 +12519,37 @@ std::pair<uint64_t, std::vector<std::pair<crypto::key_image, crypto::signature>>
...
+ * like an incoming "payment", and will incorrectly place it in the payments
+ * container. This function erases payments that are actually change, or are
+ * self-spends to the same account.
+ *
+ */
+void erase_incoming_payments_for_spent_tx(wallet2::payment_container &payments, const crypto::hash &spent_txid, uint32_t spending_subaddr_account)
+{
+  for (auto j = payments.begin(); j != payments.end(); )
+  {
+    if (j->second.m_tx_hash == spent_txid && j->second.m_subaddr_index.major == spending_subaddr_account)
```
jeffro256, line 12540: We're now requiring an extra condition to be filled for the payment to be erased: `j->second.m_subaddr_index.major == spending_subaddr_account`. So is it correct to say that now it won't be erased if sent to a different subaddress of the same account?

j-berman: Correct -- this is to match the behavior of the default scanner:

https://github.com/monero-project/monero/blob/9750e1fa103539b3e533455500610aae76e253a5/src/wallet/wallet2.cpp#L2399-L2404

j-berman: Sorry, slight correction: the payment won't be erased if sent to a **different account** than the spending account, from the same wallet

And this is to match behavior of the default scanner :)

PR #8293, opened 2022-04-26
BP+ optimization save one inversion in prover, use sc_muladd

ph4r05:
Just a minor thing - I noticed one field inversion per round can be saved in the BP+ prover while I was implementing it for Trezor
@SarangNoether @moneromooo-monero 

Not sure if this is already fixed in any of your branches or other PRs, I didn't find it among PRs. If it is a duplicate, feel free to close :)


jeffro256:
Nice find!

PR #8295, opened 2022-04-26
EPEE: Assimilate math_helper.h and document time_helper.h

jeffro256:
* Move `once_a_time` (and related specializations) to `time_helper.h`
* Delete `math_helper.h`
* Change `once_a_time::get_time()` to use steady time instead of wall time
* Make methods of `once_a_time` `inline`
* Add member `m_force_do` which signals that functor should be executed on `do_call()` regardless of timing mechanism
* Remove `once_a_time_seconds_range` because it was a pointless, confusing pass-though struct for `once_a_time`
* Change includes of `math_helper.h` appropriately
* Change some names from `once_a_time_seconds_range` in `simplewallet.cpp`
* Add a lot of additional comments/docs inside `time_helper.h`

One of the more controversial (but hopefully beneficial) things I did was switch the interval clock from wall time
to steady/CPU time for `once_a_time`. After looking at the use cases in the project, I figured that this would be a
good decision b/c it/'s used a lot as a way to do some kind of network refreshing or other polling. These tasks
shouldn't completely block because the system time changed backwards.


jeffro256:
Closing as this is mostly a preference rewrite

PR #8296, opened 2022-04-27
Gitian: refresh the stale Monero dir via --setup switch

mj-xmr:
Prevents symptoms such as:

`Submodule 'external/miniupnp' is not up-to-date.`

during performing "Reproducible Builds".

A much less obtrusive alternative to https://github.com/monero-project/monero/pull/8292

Here I'm simply deleting the `monero` directory via the `--setup` switch and allow it to be cloned from scratch.
Thanks to @hyc for pointing out the easy way.

### DIFF
--- a/contrib/gitian/README.md
+++ b/contrib/gitian/README.md
@@ -133,10 +133,11 @@ Common setup part:
 su - gitianuser
 
 GH_USER=YOUR_GITHUB_USER_NAME
-VERSION=v0.17.2.0
+VERSION=v0.17.3.2
 ```
 
 Where `GH_USER` is your GitHub user name and `VERSION` is the version tag you want to build. 
+The `gitian-build.py`'s `--setup` switch will also refresh the environment of any stale files and submodules.
 
 Setup for LXC:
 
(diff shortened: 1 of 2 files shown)


jeffro256:
Ah, the good ole "turn it off and turn it back on"

mj-xmr:
> Where are the submodules populated in this script, and could you simplify the script by adding `--recursive` when cloning `monero` repo?

Unfortunately the problem is, that they're (re-)populated in the `gitian-build` project, effectively overwriting what you'd do here, like you can see in https://github.com/monero-project/monero/pull/8292 . That project is out of our control and patching it like I did, requires a leap of faith form the Reviewers/Users.

Even though it sounds ok for the `gitian-build` to download the subs, the script doesn't properly synchronizes them, after we update our references to new versions.

hyc (approved):

jeffro256:
Ah, which was why you changed the gitian repo URL in the other PR, makes sense. It seems like Bitcoin Core plans on sunsetting `gitian-build` in favor of GUIX, so maybe we will have to maintain our own `gitian` anyways sometime in the future. At any rate, I like this solution much better.

jeffro256 (approved):

hyc (approved):

jeffro256 (approved):

PR #8297, opened 2022-04-27
Update year to 2022

AkritW:
Update year from 2021 to 2022

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2014-2021, The Monero Project
+Copyright (c) 2014-2022, The Monero Project
 
 All rights reserved.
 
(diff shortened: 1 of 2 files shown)


tobtoht (approved):
Looks like this slipped past @mj-xmr's recent PR. (Pinging you in case you want to update the script)

mj-xmr:
> Looks like this slipped past @mj-xmr's recent PR. (Pinging you in case you want to update the script)

Thank you. Indeed. The reason is that this file has no extension. Will update.

mj-xmr (approved):

mj-xmr (changes requested):
Please update the `Makefile` as well, and it's good to go.

AkritW:
Updated the Makefile.

tobtoht (approved):

mj-xmr (approved):
You'd typically be asked to squash the commits to just one and then force push, but for such a small change it probably doesn't matter.

selsta:
We will continue in #8302. @AkritW you will get added as co-author.

PR #8304, opened 2022-04-29
More tests for view tags

SChernykh:
P2Pool can create transactions with more than 128 outputs, which make output_index's varint larger than 1 byte. Added this test case.


SChernykh:
@j-berman please review

j-berman (approved):
LGTM, also sanity checked using [this alternative implementation](https://github.com/monero-ecosystem/monero-python/pull/116/files#diff-b1c9cdf99f6eee72551fc2a8de7a2dec8b82c022f7f8f41572701f4c7e34a9c9R152-R158) and got the same results.

Also a good related test on varints here: https://github.com/monero-project/monero/blob/67e5ca9ad6f1c861ad315476a88f9d36c38a0abb/tests/unit_tests/varint.cpp#L48-L64

I was thinking there may be an overflow risk when writing the varint, which could lead to fingerprinting 2 outputs to the same address. That test provides assurance that isn't a risk since it correctly reads the varint back as the same idx (I also tried it with `size_t` instead of `uint64_t`, same as `derive_view_tag`)

jeffro256 (approved):

Issue #8314, opened 2022-05-05
Support checking if wallet rpc is connected to daemon


woodser:
This issue requests the ability to check if monero-wallet-rpc is connected to a daemon using an rpc call.

One workaround is to call `check_reserve_proof` and check for the error message "Failed to connect to daemon", but this is roundabout and [causes an error](https://github.com/monero-ecosystem/monero-javascript/issues/56) in monero-wallet-rpc's console unless a real message signature is used.

jeffro256:
This is a good idea

PR #8317, opened 2022-05-06
wallet_rpc_server: add daemon connection check

### LINKED ISSUE #8316: [bug] - Set daemon allows any address
Related to #8314 

Current behavior:
The `set_daemon` wallet rpc command allows setting the the daemon to an invalid address. This in turn results in the wallet rpc staying alive but continuously throwing the "Exception while refreshing, what=no connection to daemon" error message. 

Expected behavior:
Throw an error message when there is no daemon running at the address passed instead of returning HTTP status 200.

Steps to replicate
1. Start monerod
2. Connect monero-wallet-rpc and open a wallet
3. Call `set_daemon` with an invalid daemon address

Reference:
https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#set_daemon

reemuru:
Resolves #8316

Add check_connection boolean flag to the set_daemon wallet rpc
command in order to validate that the daemon is actually connected.
At the bare minimum a valid daemon address should be passed along
with this flag. If no valid address is passed or the current address
has a daemon that is no longer running error code -38 is returned.

### DIFF
--- a/src/wallet/wallet_rpc_server_commands_defs.h
+++ b/src/wallet/wallet_rpc_server_commands_defs.h
@@ -47,7 +47,7 @@
 // advance which version they will stop working with
 // Don't go over 32767 for any of these
 #define WALLET_RPC_VERSION_MAJOR 1
-#define WALLET_RPC_VERSION_MINOR 24
+#define WALLET_RPC_VERSION_MINOR 25
 #define MAKE_WALLET_RPC_VERSION(major,minor) (((major)<<16)|(minor))
 #define WALLET_RPC_VERSION MAKE_WALLET_RPC_VERSION(WALLET_RPC_VERSION_MAJOR, WALLET_RPC_VERSION_MINOR)
 namespace tools
@@ -2666,6 +2666,7 @@ namespace wallet_rpc
       std::string ssl_ca_file;
       std::vector<std::string> ssl_allowed_fingerprints;
       bool ssl_allow_any_cert;
+      bool check_connection;
 
       BEGIN_KV_SERIALIZE_MAP()
         KV_SERIALIZE(address)
@@ -2678,6 +2679,7 @@ namespace wallet_rpc
         KV_SERIALIZE(ssl_ca_file)
         KV_SERIALIZE(ssl_allowed_fingerprints)
         KV_SERIALIZE_OPT(ssl_allow_any_cert, false)
+        KV_SERIALIZE_OPT(check_connection, false)
       END_KV_SERIALIZE_MAP()
     };
     typedef epee::misc_utils::struct_init<request_t> request;
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -330,6 +330,13 @@ namespace tools
       return false;
   }
   //------------------------------------------------------------------------------------------------------------------------------
+  bool wallet_rpc_server::not_connected(epee::json_rpc::error& er)
+  {
+      er.code = WALLET_RPC_ERROR_CODE_NO_DAEMON_CONNECTION;
+      er.message = "No connection to daemon";
+      return false;
+  }
+  //------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -2666,6 +2666,7 @@ namespace wallet_rpc
       std::string ssl_ca_file;
       std::vector<std::string> ssl_allowed_fingerprints;
       bool ssl_allow_any_cert;
+      bool is_connected;
```
jeffro256, line 2669: I think that this field would better be named something else like `check_connection` because it's a flag/request to the wallet to do something rather than the response itself.

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -2678,6 +2679,7 @@ namespace wallet_rpc
         KV_SERIALIZE(ssl_ca_file)
         KV_SERIALIZE(ssl_allowed_fingerprints)
         KV_SERIALIZE_OPT(ssl_allow_any_cert, false)
+        KV_SERIALIZE_OPT(is_connected, false)
```
jeffro256, line 2682: This  is up for debate, as it *may* break people's expectations of this RPC call, but I think that the connection checking should default to `true`. This depends on how most people use this function, but I think it makes the most sense to fail if the daemon you try setting can't be connected to, even if updating just the information itself succeeded. Discerning between the 2 scenarios can be left up to error codes.

jeffro256 (changes requested):

src/wallet/wallet_rpc_server_commands_defs.h
reemuru: Since the wallet rpc will fail fast if a user attempts to start with no daemon running this extension of the rpc would most likely be useful after the fact. Enter in the last known good address with this flag to ensure it is still online. My mindset was that the bug is missing functionality / error code so I left it as false in order to maintain backwards compatibility. Let's see what others think as well. @woodser do you think this will suffice for #8314? I know it was a request for a separate addition to wallet rpc but I thought this might be an opportunity to kill to birds with one one commit.

src/wallet/wallet_rpc_server_commands_defs.h
reemuru: changed to `check_connection` on [e71d7f6](https://github.com/monero-project/monero/pull/8317/commits/e71d7f64509ba39d911b40aa64b71725fcec652f)

woodser:
I think this is a fine addition to `set_daemon`, but it serves a slightly different purpose than https://github.com/monero-project/monero/issues/8314. The user may want to check the current daemon connection without setting it.

If we do add this, I'd favor `check_connection` defaulting to false for backwards compatibility and because setting the connection does not necessarily mean immediately using it.

jeffro256:
Yeah I don't think it resolves #8314, but I think this is a great addition by itself, and I think it wold be cleanest to open a separate PR for that standalone "check daemon connection" RPC endpoint that @woodser is talking about.

jeffro256:
Okay so I was doing functional tests, and it appears that the call to `m_wallet->check_connection` causes some kind of deadlock. When `check_connection` is set to `true`, the RPC call hangs indefinitely until I stop the `monero-wallet-rpc` server.

reemuru:
> Okay so I was doing functional tests, and it appears that the call to `m_wallet->check_connection` causes some kind of deadlock. When `check_connection` is set to `true`, the RPC call hangs indefinitely until I stop the `monero-wallet-rpc` server.
> 
> Can anyone else reproduce this?

Thanks for testing @jeffro256. I will take a deeper look into this later.

jeffro256:
Did you ever reproduce the deadlock? Since some connection overhauls were merged into master, it could be worth a shot ti test again

reemuru:
> Did you ever reproduce the deadlock? Since some connection overhauls were merged into master, it could be worth a shot ti test again

@jeffro256 oops, forgot about this. Will re-test in a bit and report back.

reemuru:
@jeffro256 yeah I was able to reproduce. It only happens when try to set the daemon on same port as rpc. I guess could parse the address and check the port if that is the only port that causes deadlock?

PR #8322, opened 2022-05-10
Update sha256sum code to comply with OpenSSL 3.0

jeffro256:
As of OpenSSL 3.0, `SHA256_Init`, `SHA256_Update`, and `SHA256_Final` are deprectaed in favor of the higher-level `EVP_*` class of functions. This causes compiler warnings, and sooner or later, will cause build errors as these functions are excluded from distro builds.

This PR does a few things:
* Updates the `sha256sum()` code to comply with OpenSSL 3.0
* Moves `sha256sum` out of `common/util` and into its own aptly named files
* Adds a bunch of error logging
* Uses the system-defined buffer size instead of a hardcoded 4096 value
* Won't throw on file error and functions are now explicitly `noexcept`
* Documentation & extra unit tests


jeffro256:
To make this PR easier to review, you can use this as a reference example: <https://wiki.openssl.org/index.php/EVP_Message_Digests>

src/crypto/sha256sum.cpp
```diff
@@ -0,0 +1,117 @@
...
+namespace {
+	constexpr size_t SHA256_MD_SIZE = 32; // 256/8
+}
+
+namespace tools
+{
+	bool sha256sum(const uint8_t *data, size_t len, crypto::hash &hash)
+	{
+		// Check hash size
+		if (sizeof(hash.data) < SHA256_MD_SIZE) {
```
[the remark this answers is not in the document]
jeffro256: Yes, true, it is a compile-time constant size `HASH_SIZE`. Would it be best to change it to a `static_assert`?

src/crypto/sha256sum.cpp
```diff
@@ -0,0 +1,117 @@
...
+			MERROR("crypto::hash instance passed to sha256sum() is too small: only " << sizeof(hash.data) << " bytes");
+			return false;
+		}
+
+		unsigned int final_digest_size;
+		if (!EVP_Digest(data, len, (unsigned char*) hash.data, &final_digest_size, EVP_sha256(), NULL)) {
+			MERROR("Failed to digest with OpenSSL SHA-256 EVP");
+			return false;
+		} else if (final_digest_size > sizeof(hash.data)) { // EVP_Digest overflows buffer. shouldn't ever happen, but you never know
```
hyc, line 32: These size checks also seem a bit too paranoid to me. This only goes wrong if you're not actually using a SHA256 algorithm, no? Which would imply a pretty broken OpenSSL build.

jeffro256: Yeah, they're pretty paranoid I'd say. In practice, these would only trigger if the OpenSSL implementation was broken. IIUC though, officially, up to `EVP_MAX_MD_SIZE` (36 bytes I believe)  can be written from a call to `EVP_DigestFinal_ex` which is larger than `HASH_SIZE`.

src/crypto/sha256sum.cpp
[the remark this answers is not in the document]
hyc: Yeah that'd be better.

src/crypto/sha256sum.cpp
hyc: If you asked for SHA256 and got 36 bytes back you didn't get what you asked for. I think we should be allowed to expect that our library dependencies are actually doing what we expect them to.

jeffro256: That's a good point, I guess we've got other things to worry about if OpenSSL is broken. I pushed a new commit with your suggestions

src/crypto/sha256sum.cpp
```diff
@@ -0,0 +1,90 @@
...
+		EVP_MD_CTX* ctx;
+		if (NULL == (ctx = EVP_MD_CTX_new()))
+		{
+			MERROR("Failed to create new OpenSSL EVP_MD_CTX");
+			return false;
+		}
+		else if (!EVP_DigestInit_ex(ctx, EVP_sha256(), NULL))
+		{
+			MERROR("Failed to initialize OpenSSL SHA-256 EVP context");
+			return false;
```
hyc, line 55: This and all subsequent error returns will leak the ctx.

hyc:
I don't see the need for reorganizing sha256 to its own source file. Seems like gratuitous churn to me. Anyone else have any opinion there?

UkoeHB:
Somehow the ZtM2 pdf ended up in this PR.

selsta:
I guess the idea was to use a file with known hash, but it might be better to use something smaller so that we don't add a full megabyte to the repo.

UkoeHB:
FWIW, I support gratuitous changes that makes code easier to navigate. Big files that 'do everything' (e.g. `utils.cpp`) make it hard to find things and build a mental model of the code-base. It's easier to look through a list of well-organized directories and file names than it is to hunt through a giant list of forward declarations or function bodies.

src/crypto/evp.h
```diff
@@ -0,0 +1,83 @@
...
+
+		/**
+		 * \brief Since recreating EVP_MD_CTXs is non-trival, copy construction is not allowed
+		 */
+		DigestContext(const DigestContext& other) = delete;
+
+		/**
+		 *\brief Since recreating EVP_MD_CTXs is non-trival, copy assignment is not allowed
+		 */
+		DigestContext& operator=(const DigestContext& other) = delete;
```
UkoeHB, line 56:
```suggestion
		/**
		 * \brief Since recreating EVP_MD_CTXs is non-trival, copies and moves are not allowed.
		 */
		DigestContext& operator=(DigestContext&&) = delete;
```

The so-called 'desdemova' trick to disable both copies and moves.

hyc:
Adding a new class to wrap a structure that is never used outside of the functions that create it seems pretty bloaty to me.

jeffro256:
@hyc Yeah I agree. Currently working on finding the least ugly way to express this.

jeffro256:
Okay I made it a `unique_ptr` with a deleter. Its really small and since this is the only point where EVP *would* be used in the main codebase, AFAIC, I think this is a good solution.

src/crypto/sha256sum.cpp
SChernykh: No, it won't leak. `ctx` is an `std::unique_ptr` instance which will call `evp_md_ctx_freer` in destructor.

selsta:
This alternative patch was suggested on IRC: https://paste.debian.net/hidden/74158896/

jeffro256:
@selsta My PR differs from this diff in a couple of ways:

1. Logs what causes the errors at each step to help debug
2. Doesn't throw on file error, and can be used in `noexcept` contexts
3. Interface is documented
4. Logic is moved to aptly named file
5. Changes how the file is buffered

selsta:
Sure, I meant it's an alternative patch to do "Update sha256sum code to comply with OpenSSL 3.0 ". Your code obviously does more.

jeffro256:
This PR seems to have consumed a lot of dev time, so I want to hold a vote to see if I should close it. Thumbs up to keep open, thumbs down to close.

selsta:
I mean we will need a patch to `comply with OpenSSL 3.0`, whether it's this PR or a different one.

The easiest way we can get it merged is by keeping the diff as minimal as possible, adding tests and documentation is also good in my opinion, it doesn't make things more difficult to review.

All the other changes are personal preference which will result in discussion and we had an unwritten rule about not rewriting code for personal preference.

jeffro256:
@hyc Now the only existing modified files in `src/` are `util.cpp`, `util.h`, and `crypto/CMakeLists.txt` and those diffs are really easy to review. `sha256sum.cpp` and `sha256sum.h` remain the same.

selsta:
This would be a mix between both patches, I added you as the author: https://github.com/selsta/monero/commit/32ec76faab82baccede4bac85ead0f9d9ae1d99c

jeffro256:
Continue at #8340

Issue #8251, opened 2022-04-10
Feature: resubmit a failed tx from the wallet


j-berman:
### Overview

My wallet is connected to my node via tor, my node uses the tx proxy over tor. This happened the other day:

1. I created a tx in a mobile wallet and hit send. It submit successfully and entered the "pending" state in my wallet, then after several minutes the tx entered the "failed" state. The other person didn't see the tx in their wallet, nor did I see the tx on a global block explorer (i.e. my node apparently didn't broadcast it to the rest of the network).
2. I tried to create another tx and got a "rejected because of double spend" error.
3. I created another tx from a separate account (same wallet same node) that broadcast to the rest of the network and confirmed on chain without issue. I knew this 2nd tx was risky but I wanted to pay the guy and I wanted to use my node.
4. Some number of minutes later, the first failed tx from step 1 ended up on chain without action on my part.

So I ended up paying someone twice and never saw them again unfortunately (the one that got away). I figure there could probably be some fail-safe to protect against this scenario better. I think the best fail-safe is enabling the user to resubmit a failed tx to a node from their wallet in a cleaner way. I don't see an RPC endpoint for this particular scenario.

### My investigation so far

First off, this issue happens to me sporadically and I've yet to capture it in useful logs, but I'm hoping I will soon.

This is what I believe is happening: upon initial successful tx submission, my tx first enters the pool with the [`relay_method::local` state](https://github.com/monero-project/monero/blob/9a124f681119855949f6406ecd69c2ad91da9770/src/rpc/core_rpc_server.cpp#L1291), and then *sometimes* it takes quite a bit of time before the tx reaches the `relay_method::fluff` state (which AFAIU occurs after another node tells mine it knows about the tx and [Dandellion++ has completed](https://github.com/monero-project/monero/blob/f49fc9b4876382d5c6f08fd7a6125b554c49e260/src/cryptonote_protocol/cryptonote_protocol_handler.inl#L1006)). The tx then hits the failed state from [the 500 second propagation timeout here](https://github.com/monero-project/monero/blob/f49fc9b4876382d5c6f08fd7a6125b554c49e260/src/wallet/wallet2.cpp#L3035-L3040), and my wallet indicates this to me in the UI.

Note [this TODO inside `core_rpc_server::on_send_raw_tx`](https://github.com/monero-project/monero/blob/9a124f681119855949f6406ecd69c2ad91da9770/src/rpc/core_rpc_server.cpp#L1292):

[...]

Issue #6938, opened 2020-10-26
tx-proxy I2P / Tor, no outbound connections (only inbound)


selsta:
I host I2P / Tor monero nodes with --tx-proxy and --anonymous-inbound, that works quite well but on both nodes I don’t have any outbound peers.

It looks like this on both nodes.

```
INC :                         I2P     no    xxx    0                   160847(9)/6408(81)            normal                   1898                0           0             0         0            
INC :                         I2P     no    xxx    0                   21417(14)/15039(14)           normal                   2510                0           0             0         0            
INC :                         I2P     no    xxx    0                   414479(69)/15491(69)          normal                   2778                0           0             0         0            
INC :                         Tor     no    xxx    0                   23893(9)/1916(17)             normal                   92                  0           0             0         0            
INC :                         Tor     no    xxx    0                   9428(20)/6631(20)             normal                   1014                0           0             0         0            
INC :                         Tor     no    xxx    0                   57781(6)/7451(39)             normal                   1036                0           0             0         0            
INC :                         Tor     no    xxx    0                   10015(50)/7246(50)            normal                   1032                0           0             0         0            
INC :                         Tor     no    xxx    0                   11440(24)/7451(24)            normal                   1038                0           0             0         0            
INC :                         Tor     no    xxx    0                   86729(69)/6426(69)            normal                   956                 0           0             0         0            
INC :                         Tor     no    xxx    0                   89972(16)/7451(16)            normal                   1038                0           0             0         0            
INC :                         Tor     no    xxx    0                   18702(81)/10342(81)           normal                   3423                0           0             0         0            
INC :                         Tor     no    xxx    0                   448998(0)/16082(2)            normal                   3451                0           0             0         0            
INC :                         Tor     no    xxx    0                   261783(40)/20408(40)          normal                   3531                0           0             0         0            
```

The missing outbound peers results in constant messages like this:

```
Lost all outbound connections to anonymity network - currently unable to send transaction(s)
```

Why do I have plenty inbound peers but not outbound ones? Should this get better once we have more I2P / Tor peers?

Sometimes I get outbound ones after restarting, but they drop pretty quickly and I stay at only inbound ones.

selsta:
Most likely related https://github.com/monero-project/monero/issues/6631

MoneroArbo:
[...]

Issue #6631, opened 2020-06-06
monerod only maintains an i2p outbound connection if --add-priority-node is used


MoneroArbo:
Basically what it says on the tin.

If I use the --add-peer flag, or no flag, I'm basically never able to maintain an outbound i2p connection. Sometimes it will successfully establish one, but it always drops after about a minute, give or take. It will try to establish new connections, but so many i2p addresses in my peer list are offline that it fails like 98% of the time. This is an even bigger issue than it would be because of #6590 where, when a tx is sent and there are no available i2p connections, it can get stuck in the local mempool even after an i2p connection is established.

The workaround I have found is to add a known working node as a priority node. The outbound connections still gets dropped just as often, but this way it immediately tries to re-establish a connection to a known working peer, as opposed to cycling through a list of offline peers.

I am including my [bitmonero.log](https://github.com/monero-project/monero/files/4739886/bitmonero.log.txt), which is split into two main sections. At the beginning, the log without a priority node. Then later, logs WITH a priority node. Sections of the log are labeled.

Not sure what the "real" issue is here -- maybe that offline nodes stay in the peer list for so long. Maybe that i2p connections only last ~1 minute. Maybe both. Maybe neither. But for me, --tx-proxy with i2p is essentially non-functional without taking this step of adding a priority node.

MoneroArbo:
My [daemon log](https://github.com/monero-project/monero/files/4747047/bitmonero.log.txt) (with log level 0) over the past ~2 days, to give an idea of just how often outbound i2p connections fail. This is WITH --add-priority-node option using the dsc* address at https://github.com/i2p-zero/i2p-zero/blob/master/mipseed.md.

Without the option I actually see the error less often, I suppose because connections are rarely established to begin with.

knaccc:
I'm having similar issues. @selsta and niocbrrrrrr kindly set up mipseeds in the last few hours, and I've added their new seed nodes to the list in this doc: https://github.com/i2p-zero/i2p-zero/blob/master/mipseed.md (Note that the `core5....` address is not currently active).

I'm using this command to monitor I2P connections: `watch 'curl -H "Content-Type: application/json" --data "{\"public_only\":false}" http://127.0.0.1:18081/get_peer_list 2>/dev/null | grep -A 8 "i2p" | grep "i2p\|last_seen"'`

[...]

PR #8324, opened 2022-05-12
p2p: Don't send peer list larger than max allowed

j-berman:
Tor/i2p nodes will sometimes send peer lists of size > the expected max to outgoing connections, which causes tor/i2p connections to drop. This PR ensures the peer list remains <= the max size.

I have a decent degree of confidence this is a significant cause of tor/i2p dropped connection issues in #6938 and #6631 (and sporadic failed tx submissions on tor nodes like in #8251), but it's not the only cause of these issues. Still in the process of investigating and patching what look like clearly apparent bugs, separate from this one.

In testing on my machine, I observed that my tor node will push peer lists larger than the max size on average every 1-3 minutes or so (sometimes the gaps are larger, sometimes smaller),  which is followed by the existing outgoing tor connection dropping. Here's what I believe is happening under the hood:

1. OUTGOING tor node runs [peer_sync_idle_maker](https://github.com/monero-project/monero/blob/9f814edbd78c70c70b814ca934c1ddef58768262/src/p2p/net_node.inl#L2004) in the idle loop
2. In 1-3 minute intervals, the OUTGOING node calls [do_peer_timed_sync](https://github.com/monero-project/monero/blob/9f814edbd78c70c70b814ca934c1ddef58768262/src/p2p/net_node.inl#L2117), which makes the [COMMAND_TIMED_SYNC](https://github.com/monero-project/monero/blob/9f814edbd78c70c70b814ca934c1ddef58768262/src/p2p/net_node.inl#L1231) request to MY tor node
3. MY tor node responds with peer list of size > expected max in [handle_timed_sync](https://github.com/monero-project/monero/blob/9f814edbd78c70c70b814ca934c1ddef58768262/src/p2p/net_node.inl#L2486-L2487), **because it adds my own .onion to the peer list even though the list is already at max** (the bug)
4. OUTGOING node [processes the response](https://github.com/monero-project/monero/blob/9f814edbd78c70c70b814ca934c1ddef58768262/src/p2p/net_node.inl#L1161) and sees that the peer list is too large in [handle_remote_peerlist](https://github.com/monero-project/monero/blob/9f814edbd78c70c70b814ca934c1ddef58768262/src/p2p/net_node.inl#L2161-L2165), then closes the connection

This PR's fix makes sure MY node won't response with a peer list larger than the max when adding my own .onion to the peer list in step 3.

Credit to @selsta for helping debug too

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -2462,8 +2462,12 @@ namespace nodetool
     const epee::net_utils::zone zone_type = context.m_remote_address.get_zone();
     network_zone& zone = m_network_zones.at(zone_type);
 
+    //will add self to peerlist if in same zone as outgoing later in this function
+    const bool outgoing_to_same_zone = !context.m_is_income && zone.m_our_address.get_zone() == zone_type;
+    const uint32_t max_peerlist_size = P2P_DEFAULT_PEERS_IN_HANDSHAKE - (outgoing_to_same_zone ? 1 : 0);
+
     std::vector<peerlist_entry> local_peerlist_new;
-    zone.m_peerlist.get_peerlist_head(local_peerlist_new, true, P2P_DEFAULT_PEERS_IN_HANDSHAKE);
+    zone.m_peerlist.get_peerlist_head(local_peerlist_new, true, max_peerlist_size);
 
     //only include out peers we did not already send
     rsp.local_peerlist_new.reserve(local_peerlist_new.size());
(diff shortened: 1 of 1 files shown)


moneromooo-monero:
The local peer will be added at the end of the list, it should probably be shuffled in another patch (not sure if this can be used when the local address is not known to the peer already, but shuffling makes sure this can never be an issue).

PR #8328, opened 2022-05-14
disable multisig by default

moneromooo-monero:
There are vulnerabilities in multisig protocol if the parties do not
trust each other, and while there is a patch for it, it has not been
throroughly reviewed yet, so it is felt safer to disable multisig by
default for now.
If all parties in a multisig setup trust each other, then it is safe
to enable multisig.


kayabaNerve:
Can the flag itself communicate this is unsafe? enable-unsafe-multisig? enable-multisig-experimental? etc

I think the former is more accurate if this will be applied to the existing multisig protocol as well as the new one, whereas the latter makes sense for just the new one.

src/simplewallet/simplewallet.cpp
```diff
@@ -986,12 +997,14 @@ bool simple_wallet::print_fee_info(const std::vector<std::string> &args/* = std:
 
 bool simple_wallet::prepare_multisig(const std::vector<std::string> &args)
 {
+  CHECK_MULTISIG_ENABLED();
   prepare_multisig_main(args, false);
```
rbrunner7, line 1001: Well, if there is a `main` method in addition to the direct method for interactive use that one will already check for multisig enabled, so strictly speaking a number of such checks are superfluous, but maybe it's a good idea to have them nevertheless so it's crystal clear there is a check and everybody reading the code sees this as early as possible.

rbrunner7:
I reviewed, and it looks good, but I wonder why there seems to be no info or warning to the user **why** multisig is disabled by default anywhare at any time. The new macro `CHECK_MULTISIG_ENABLED` in `simplewallet.cpp` simply tells now:

    Multisig is disabled, you can enable it with:
       set enable-multisig 1

Do you wait for a good message formulated by @UkoeHB to add in order to complete the PR with it?

moneromooo-monero:
I can change the message if I get one, yes.
The extra checks won't get you two messages, due to the early out. I don't particularly want to verify whether every one can be called only via the top level one.

src/simplewallet/simplewallet.cpp
```diff
@@ -3806,6 +3851,7 @@ bool simple_wallet::set_variable(const std::vector<std::string> &args)
     success_msg_writer() << "auto-mine-for-rpc-payment-threshold = " << m_wallet->auto_mine_for_rpc_payment_threshold();
     success_msg_writer() << "credits-target = " << m_wallet->credits_target();
     success_msg_writer() << "load-deprecated-formats = " << m_wallet->load_deprecated_formats();
+    success_msg_writer() << "enable-multisig = " << m_wallet->enable_multisig();
```
UkoeHB, line 3854:
```suggestion
    success_msg_writer() << "enable-multisig-experimental = " << m_wallet->enable_multisig();
```

src/simplewallet/simplewallet.cpp
```diff
@@ -3872,6 +3918,7 @@ bool simple_wallet::set_variable(const std::vector<std::string> &args)
     CHECK_SIMPLE_VARIABLE("persistent-rpc-client-id", set_persistent_rpc_client_id, tr("0 or 1"));
     CHECK_SIMPLE_VARIABLE("auto-mine-for-rpc-payment-threshold", set_auto_mine_for_rpc_payment_threshold, tr("floating point >= 0"));
     CHECK_SIMPLE_VARIABLE("credits-target", set_credits_target, tr("unsigned integer"));
+    CHECK_SIMPLE_VARIABLE("enable-multisig", set_enable_multisig, tr("0 or 1"));
```
UkoeHB, line 3921:
```suggestion
    CHECK_SIMPLE_VARIABLE("enable-multisig-experimental", set_enable_multisig, tr("0 or 1"));
```

src/wallet/wallet2.h
```diff
@@ -1297,6 +1297,8 @@ namespace tools
     void set_rpc_client_secret_key(const crypto::secret_key &key) { m_rpc_client_secret_key = key; m_node_rpc_proxy.set_client_secret_key(key); }
     uint64_t credits_target() const { return m_credits_target; }
     void credits_target(uint64_t threshold) { m_credits_target = threshold; }
+    bool enable_multisig() const { return m_enable_multisig; }
```
UkoeHB, line 1300:
```suggestion
    bool multisig_is_enabled() const { return m_enable_multisig; }
```
Can you disambiguate the two functions here? One `enables` and the other asks `is it enabled?`.

src/wallet/wallet2.h
```diff
@@ -1811,6 +1813,7 @@ namespace tools
     crypto::secret_key m_rpc_client_secret_key;
     rpc_payment_state_t m_rpc_payment_state;
     uint64_t m_credits_target;
+    bool m_enable_multisig;
```
UkoeHB, line 1816:
```suggestion
    bool m_multisig_enabled;
```
`enable` is a verb, `enabled` is an adjective

src/wallet/wallet_rpc_server.cpp
```diff
@@ -61,6 +61,17 @@ using namespace epee;
 
 #define DEFAULT_AUTO_REFRESH_PERIOD 20 // seconds
 
+#define CHECK_MULTISIG_ENABLED() \
+  do \
+  { \
+    if (m_wallet->multisig() && !m_wallet->enable_multisig()) \
+    { \
+      er.code = WALLET_RPC_ERROR_CODE_DISABLED; \
+      er.message = "This wallet is multisig, and multisig is disabled. You can enable it by running this once in monero-wallet-cli: set enable-multisig 1"; \
```
UkoeHB, line 70:
```suggestion
      er.message = "This wallet is multisig, and multisig is disabled. You can enable it by running this once in monero-wallet-cli: set enable-multisig-experimental 1"; \
```

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -2416,7 +2416,10 @@ namespace wallet_rpc
   {
     struct request_t
     {
+      bool enable_multisig;
```
UkoeHB, line 2419:
```suggestion
      bool enable_multisig_experimental;
```
For any user-facing API, add the experimental qualifier.

src/wallet/wallet_rpc_server_error_codes.h
```diff
@@ -78,3 +78,4 @@
 #define WALLET_RPC_ERROR_CODE_ATTRIBUTE_NOT_FOUND    -45
 #define WALLET_RPC_ERROR_CODE_ZERO_AMOUNT            -46
 #define WALLET_RPC_ERROR_CODE_INVALID_SIGNATURE_TYPE -47
+#define WALLET_RPC_ERROR_CODE_DISABLED               -48
```
UkoeHB, line 81:
```suggestion
#define WALLET_RPC_ERROR_CODE_MULTISIG_DISABLED      -48
```

utils/python-rpc/framework/wallet.py
```diff
@@ -490,10 +490,11 @@ def is_multisig(self):
         }
         return self.rpc.send_json_rpc_request(is_multisig)
 
-    def prepare_multisig(self):
+    def prepare_multisig(self, enable_multisig = False):
         prepare_multisig = {
             'method': 'prepare_multisig',
             'params' : {
+                'enable_multisig': enable_multisig,
```
UkoeHB, line 497:
```suggestion
                'enable_multisig_experimental': enable_multisig_experimental,
```

UkoeHB (changes requested):
I agree with @kayabaNerve that `experimental` cleanly communicates why multisig needs to be enabled. Sorry that my semantic recommendations might cause a big diff compared to the existing commit (I didn't suggest changes for all cases, only the main ones).

UkoeHB:
Here is a blurb about multisig (not sure exactly where/how you want to use it).

```
Multisig is an experimental feature and may have bugs. Things that could go wrong include: funds sent to a multisig wallet can't be spent at all, funds can only be spent with the participation of a malicious group member, funds can be stolen by a malicious group member.
```

rbrunner7 (approved):
Personally I find the name of the option and the warning about the possible bugs a bit drastic, but hey, better err on the side of caution.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -61,6 +61,17 @@ using namespace epee;
 
 #define DEFAULT_AUTO_REFRESH_PERIOD 20 // seconds
 
+#define CHECK_MULTISIG_ENABLED() \
+  do \
+  { \
+    if (m_wallet->multisig() && !m_wallet->is_multisig_enabled()) \
+    { \
+      er.code = WALLET_RPC_ERROR_CODE_DISABLED; \
+      er.message = "This wallet is multisig, and multisig is disabled. Multisig is an experimental feature and may have bugs. Things that could go wrong include: funds sent to a multisig wallet can't be spent at all, can only be spent with the participation of a malicious group member, or can be stolen by a malicious group member. You can enable it by running this once in monero-wallet-cli: set enable-multisig 1"; \
```
UkoeHB, line 70:
```suggestion
      er.message = "This wallet is multisig, and multisig is disabled. Multisig is an experimental feature and may have bugs. Things that could go wrong include: funds sent to a multisig wallet can't be spent at all, can only be spent with the participation of a malicious group member, or can be stolen by a malicious group member. You can enable it by running this once in monero-wallet-cli: set enable-multisig-experimental 1"; \
```
typo

src/wallet/wallet_rpc_server_error_codes.h
UkoeHB: For `WALLET_RPC_ERROR_CODE_DISABLED` are you trying to add a generic 'disabled' code, or should it be `WALLET_RPC_ERROR_CODE_MULTISIG_DISABLED`?

UkoeHB (changes requested):
minor additional comments

moneromooo-monero:
Yes, there's no particularly good reason to restrict it to multisig really.

UkoeHB (approved):
Thanks

PR #8330, opened 2022-05-15
p2p: don't exclusively drop tor/i2p outgoing cxns in idle loop

j-berman:
### Context

`update_sync_search` runs in the idle loop, checking to see if we have outgoing connections >= the max allowed. If so, a sync'd connection gets dropped so that the daemon can try to connect to nodes that need to sync.

### Before this PR

Currently it's [comparing the tally of **clearnet + tor + i2p** outgoing connections to the max of outgoing **clearnet** peers](https://github.com/monero-project/monero/blob/8349cfe4a63cfc63d50ce3818886b67a05e240a4/src/cryptonote_protocol/cryptonote_protocol_handler.inl#L1798); but, the tally of clearnet + tor + i2p outgoing peers is usually always going to be greater than the max allowed outgoing clearnet peers. Because of this, it's usually just going to drop a tor or i2p connection [right here](https://github.com/monero-project/monero/blob/8349cfe4a63cfc63d50ce3818886b67a05e240a4/src/cryptonote_protocol/cryptonote_protocol_handler.inl#L1802) (because they are last when iterating over all connections), even if the daemon has just 1 outgoing tor or i2p connection.

This seems to be another source of tor/i2p connection reliability issues mentioned in #6938, #6631, #8251.

### This PR

This PR makes sure to tally outgoing connections by zone, then compare to the max by zone. If the max peer count set for tor is 10 for example, and there are <10 outgoing tor connections, an outgoing tor connection won't get dropped.

### Future PR

I left a couple TODO's to investigate updating the pruning logic correctly. Didn't seem as trivial to figure out, so I left the logic how it currently is (just comparing the tally to the clearnet max).

### DIFF
--- a/src/cryptonote_protocol/cryptonote_protocol_handler.h
+++ b/src/cryptonote_protocol/cryptonote_protocol_handler.h
@@ -113,12 +113,23 @@ namespace cryptonote
     const block_queue &get_block_queue() const { return m_block_queue; }
     void stop();
     void on_connection_close(cryptonote_connection_context &context);
-    void set_max_out_peers(unsigned int max) { m_max_out_peers = max; }
+    void set_max_out_peers(epee::net_utils::zone zone, unsigned int max) { CRITICAL_REGION_LOCAL(m_max_out_peers_lock); m_max_out_peers[zone] = max; }
+    unsigned int get_max_out_peers(epee::net_utils::zone zone) const
+    {
+      CRITICAL_REGION_LOCAL(m_max_out_peers_lock);
+      const auto it = m_max_out_peers.find(zone);
+      if (it == m_max_out_peers.end())
+      {
+        MWARNING(epee::net_utils::zone_to_string(zone) << " max out peers not set, using default");
+        return P2P_DEFAULT_CONNECTIONS_COUNT;
+      }
+      return it->second;
+    }
     bool no_sync() const { return m_no_sync; }
     void set_no_sync(bool value) { m_no_sync = value; }
     std::string get_peers_overview() const;
     std::pair<uint32_t, uint32_t> get_next_needed_pruning_stripe() const;
-    bool needs_new_sync_connections() const;
+    bool needs_new_sync_connections(epee::net_utils::zone zone) const;
     bool is_busy_syncing();
 
   private:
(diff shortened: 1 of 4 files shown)


src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -113,12 +113,22 @@ namespace cryptonote
...
     void stop();
     void on_connection_close(cryptonote_connection_context &context);
-    void set_max_out_peers(unsigned int max) { m_max_out_peers = max; }
+    void set_max_out_peers(epee::net_utils::zone zone, unsigned int max) { m_max_out_peers[zone] = max; }
+    unsigned int get_max_out_peers(epee::net_utils::zone zone) const
+    {
+      const auto it = m_max_out_peers.find(zone);
+      if (it == m_max_out_peers.end())
+      {
+        MWARNING(epee::net_utils::zone_to_string(zone) << " max out peers not set, using default");
```
moneromooo-monero, line 122: Do you really want to spam people who set no limit ? I assume this'd do. It maybe a rare case, but it's legitimate. Or is a max always set even if the user doesn't set one ?

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -171,7 +181,7 @@ namespace cryptonote
     epee::math_helper::once_a_time_milliseconds<100> m_standby_checker;
     epee::math_helper::once_a_time_seconds<101> m_sync_search_checker;
     epee::math_helper::once_a_time_seconds<43> m_bad_peer_checker;
-    std::atomic<unsigned int> m_max_out_peers;
+    std::unordered_map<epee::net_utils::zone, std::atomic<unsigned int>> m_max_out_peers;
```
moneromooo-monero, line 184: The map should be atomic to match the previous semantics.

src/cryptonote_protocol/cryptonote_protocol_handler.h
j-berman: Maybe I'm missing something but I can't see how it wouldn't get set by a regular user; it should always get set in here even if a user sets a limit of 0 or doesn't include the `--out-peers` flag: https://github.com/j-berman/monero/blob/299fbd007687534a3b5abb87c2b3f2f596ecdacb/src/p2p/net_node.inl#L388

The only way I believe it wouldn't get set is if a dev is doing something custom where they skip setting it before calling a function that reads it, like in the unit test I updated in this PR (which was hanging before I made this change). I.e. it should always be set explicitly before it's read in normal use.

Maybe there's a better way to handle this? Default to 0 in case it's not set, and maybe only warn once?

moneromooo-monero: I did not look. If it's impossible with "normal" usage, it's fine by me.

moneromooo-monero: I did not look. If it's impossible with "normal" usage, it's fine by me.

moneromooo-monero (approved):

selsta:
@j-berman please squash

PR #8331, opened 2022-05-15
wallet2: fix spurious reorg detection with untrusted nodes

moneromooo-monero:
When forced to deal with an untrusted node, a wallet will quantize
its current height to disguise the real height to the adversary, to
try and minimize the daemon's ability to distinguish returning
wallets.
Daemons will thus return more blocks than the wallet needs, starting
from earlier in the chain. These extra blocks will be disregarded
by the wallet, which had already scanned them.
However, for the purposes of reorg size detection, the wallet assumes
all blocks the daemon sends are different, which is only correct if
the wallet hasn't been coy, which is only the case for trusted
daemons (which you should use). This causes an issue when the size
of this "fake reorg" is above the sanity check threshold at which
the wallet refuses a reorg.
To fix this, the reorg size check is moved later on, when the reorg
is about to actually happen, after the wallet has checked which
blocks are actually different from the ones it expects.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -2882,6 +2882,11 @@ void wallet2::process_parsed_blocks(uint64_t start_height, const std::vector<cry
         " (height " + std::to_string(start_height) + "), local block id at this height: " +
         string_tools::pod_to_hex(m_blockchain[current_index]));
 
+      const uint64_t reorg_depth = m_blockchain.size() - current_index;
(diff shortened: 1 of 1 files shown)


rbrunner7 (approved):
Looks ok to me, and I tested the code which correctly solved the problem that I had using a daemon un-trustedly.

SChernykh (approved):

serhack (approved):

jeffro256 (approved):

PR #8333, opened 2022-05-16
Focus Doxygen documentation

jeffro256:
Right now Doxygen is documenting everything in the repo including submodules, everything in contrib, util, tests, etc. This bogs down the documentation to the point where it is very hard to navigate. I think it would be a good move to focus on documenting only the main C++ code which is specific to this repo.

Right now this means documenting `src/` (without SUPERCOP) and `contrib/epee/`. After this commit, Doxygen went from running >6000 graphs to about 2200 graphs.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -754,7 +754,7 @@ WARN_LOGFILE           =
 # spaces.
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = .
+INPUT                  = contrib/epee external/easylogging++ src 
 
 # This tag can be used to specify the character encoding of the source files
 # that doxygen parses. Internally doxygen uses the UTF-8 encoding. Doxygen uses
@@ -805,7 +805,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
-EXCLUDE_PATTERNS       = */build/* */contrib/depends/*
+EXCLUDE_PATTERNS       = */src/crypto/crypto_ops_builder/ref10*
 
 # The EXCLUDE_SYMBOLS tag can be used to specify one or more symbol names
 # (namespaces, classes, functions, etc.) that should be excluded from the


jeffro256:
I'm also planning on adding a main page in a later commit

Doxyfile
```diff
@@ -754,7 +754,7 @@ WARN_LOGFILE           =
 # spaces.
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = .
+INPUT                  = contrib/epee src
```
mj-xmr, line 757: I'd add easylogging++,

mj-xmr (changes requested):

Doxyfile
jeffro256: This is obviously up for debate, but I decided to not include easylogging++, qrcode, and db_drivers in the documentation since they all have their own separate documentation elsewhere and we would just be redoing their work.

mj-xmr: easylogging++ is definitely under our maintenance now.

LeoNero:
I agree with improving Doxygen, given that the first time I opened it, I found that there were to many information not related to the Monero codebase, which does not help beginners that much :)

Doxyfile
LeoNero: Agree with mj. Also, adding links to the docs of the other projects would definitely help.

Doxyfile
```diff
@@ -805,7 +805,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
-EXCLUDE_PATTERNS       = */build/* */contrib/depends/*
+EXCLUDE_PATTERNS       = */src/crypto/crypto_ops_builder/ref10*
```
LeoNero, line 808: just curiosity, why exclude it?

jeffro256: Because the SUPERCOP ref10 is already documented outside of Monero and it causes a huge disproportionate amount of documentation to be created

mj-xmr (approved):

PR #8335, opened 2022-05-16
net_ssl: Use EVP API to create EVP_PKEYs when available

jeffro256:
The following functions are deprecated in OpenSSL 3.0 and used in net_ssl.cpp: `RSA_free`, `EC_KEY_free`, `RSA_new`, `RSA_generate_key_ex`, `EC_KEY_set_group`, `EC_KEY_generate_key`.

Since `EVP_RSA_gen` was added in OpenSSL 3.0 and since the aforementioned functions were deprecated in OpenSSL 3.0, I check if the OpenSSL version >= 3.0, and use `EVP_RSA_gen` instead of everything else up to assigning the private key to the certificate in `create_rsa_ssl_certificate`.

I also deleted `create_ec_ssl_certificate` since it wasn't used.

### DIFF
--- a/contrib/epee/include/net/net_ssl.h
+++ b/contrib/epee/include/net/net_ssl.h
@@ -38,8 +38,10 @@
 #include <boost/asio/ssl.hpp>
 #include <boost/filesystem/path.hpp>
 #include <boost/system/error_code.hpp>
+#include <openssl/evp.h>
 
 #define SSL_FINGERPRINT_SIZE 32
+#define SSL_RSA_BITS 4096
 
 namespace epee
 {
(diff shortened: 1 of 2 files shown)


jeffro256:
The git diff is freaking out but the change is pretty straightforward:

* Remove `create_ec_ssl_certificate`, `openssl_bio`, `openssl_rsa`, `openssl_bignum`, `openssl_ec_key`, and `openssl_group`. *NOT* `openssl_pkey` though.
* Add a function `gen_rsa_pkey_compat()` that is equivalent to `EVP_RSA_gen(SSL_RSA_BITS)` if OpenSSL >= 3.0, otherwise use the old logic for generating a RSA private and assigning it to a `EVP_PKEY`. 
* In `create_rsa_ssl_certificate`, say `pkey = gen_rsa_pkey_compat()` instead of `pkey = EVP_PKEY_new()`.

hyc:
> I also deleted create_ec_ssl_certificate since it wasn't used.

We should think about whether that's a good thing or not, since EC certs are smaller and faster to use than RSA.

jeffro256:
True. Do you know of any code of the top of your head in the project that would break if we replaced RSA certs with EC certs?

jeffro256:
Will open up a PR for elliptic curve certificates soon

jeffro256:
Continued at #8354

PR #8340, opened 2022-05-17
common: update sha256sum to use OpenSSL 3.0 API, add test cases

selsta:
Patch by @jeffro256 and from IRC, continued from #8322. This unfortunately ended in a lot of discussion so I wanted to open this so that we can just move on.


src/common/util.h
```diff
@@ -231,7 +231,25 @@ namespace tools
...
+   */
   bool sha256sum(const uint8_t *data, size_t len, crypto::hash &hash);
+
+  /**
+   * \brief Creates a SHA-256 digest of a file's contents, equivalent to the sha256sum command in Linux
+   *
+   * \param[in] filename path to target file
+   * \param[out] hash where message digest will be written to
+   *
+   * \returns true if successful, false otherwise (OpenSSL backend failure, file I/O exceptions, etc)
```
jeffro256, line 251: Since the line `f.exceptions(std::ifstream::failbit | std::ifstream::badbit);"` from `sha256sum` was kept in this PR, the function will throw on file I/O exceptions. Either that line must be removed, or this documentation should be updated.

jeffro256 (changes requested):
I appreciate that you're trying to find common ground and that you made me co-author ;)

src/common/util.cpp
```diff
@@ -961,8 +961,8 @@ std::string get_nix_version_display_string()
     if (!f)
       return false;
     std::ifstream::pos_type file_size = f.tellg();
-    SHA256_CTX ctx;
-    if (!SHA256_Init(&ctx))
+    std::unique_ptr<EVP_MD_CTX, decltype(&EVP_MD_CTX_free)> ctx(EVP_MD_CTX_new(), &EVP_MD_CTX_free);
+    if (!EVP_DigestInit(ctx.get(), EVP_sha256()))
```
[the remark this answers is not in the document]
jeffro256: `EVP_DigestInit_ex` is faster when you use the context only once, as we are doing in this case.

src/common/util.cpp
```diff
@@ -973,12 +973,12 @@ std::string get_nix_version_display_string()
...
       if (!f || !f.good())
         return false;
-      if (!SHA256_Update(&ctx, buf, read_size))
+      if (!EVP_DigestUpdate(ctx.get(), buf, read_size))
         return false;
       size_left -= read_size;
     }
     f.close();
-    if (!SHA256_Final((unsigned char*)hash.data, &ctx))
+    if (!EVP_DigestFinal(ctx.get(), (unsigned char*)hash.data, nullptr))
```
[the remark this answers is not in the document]
jeffro256: Same as above

src/common/util.h
```diff
@@ -231,7 +231,25 @@ namespace tools
...
+   */
   bool sha256sum(const uint8_t *data, size_t len, crypto::hash &hash);
+
+  /**
+   * \brief Creates a SHA-256 digest of a file's contents, equivalent to the sha256sum command in Linux
+   *
+   * \param[in] filename path to target file
+   * \param[out] hash where message digest will be written to
+   *
+   * \returns true if successful, false otherwise (OpenSSL backend failure, etc)
```
jeffro256, line 251:
```suggestion
   * \returns true if successful, false if the file can not be opened or there is an OpenSSL failure
   *
   * \throws ios_base::failure if after the file is successfully opened, an error occurs during reading
```

jeffro256: Okay this should be the last suggestion!

jeffro256 (approved):

src/common/util.cpp
```diff
@@ -941,12 +941,7 @@ std::string get_nix_version_display_string()
 
   bool sha256sum(const uint8_t *data, size_t len, crypto::hash &hash)
   {
-    SHA256_CTX ctx;
-    if (!SHA256_Init(&ctx))
-      return false;
-    if (!SHA256_Update(&ctx, data, len))
-      return false;
-    if (!SHA256_Final((unsigned char*)hash.data, &ctx))
+    if (!EVP_Digest(data, len, (unsigned char*) hash.data, NULL, EVP_sha256(), NULL))
```
hyc, line 944: for a oneline if/else like this, you should just do:
```return EVP_Digest(...) != 0;
```

selsta: force pushed it wrong, fixing it currently

hyc (approved):

PR #8344, opened 2022-05-20
core_rpc_server: new file: rpc_ssl.fingerprint

jeffro256:
## Motivation
This PR adds a new file to the data directory: `rpc_ssl.fingerprint`. This is a convenience file that stores the SHA-256 X509 digest of the PEM certificate at `rpc_ssl.crt`. This is ease-of-use improvement for those who want to manually add their node's fingerprint to the CLI wallet / other application (and hopefully GUI eventually, working on that currently). Now there aren't any commands to memorize; no searching for which types of fingerprints `net_ssl` uses. I hope this encourages more users to utilize secure connections over RPC.

## PR structure

There are 4 four part of this PR:
1. Move `get_fingerprint` function from `gen_multisig` to `net_ssl` (to be used below)
2. In `store_ssl_keys` (when storing the `.crt` and `.key` files for the first time), we now also store the `.fingerprint` file
3. Add `get_hr_ssl_fingerprint_from_file` which is basically just the command `openssl x509 -in <filename> -fingerprint -sha256 -noout` as a C++ function
4. Finally, for those existing nodes, in `core_rpc_server`, we check if we already have those `.key` and `.crt` files but *not* the `.fingerprint` file. If so, create it.

### DIFF
--- a/contrib/epee/src/net_ssl.cpp
+++ b/contrib/epee/src/net_ssl.cpp
@@ -641,6 +641,56 @@ bool ssl_options_t::handshake(
   return true;
 }
 
+std::string get_hr_ssl_fingerprint(const X509 *cert, const EVP_MD *fdig)
+{
+  unsigned int j;
+  unsigned int n;
+  unsigned char md[EVP_MAX_MD_SIZE];
+  std::string fingerprint;
+
+  CHECK_AND_ASSERT_THROW_MES(cert && fdig, "Pointer args to get_hr_ssl_fingerprint cannot be null");
+
+  if (!X509_digest(cert, fdig, md, &n))
+  {
+    const unsigned long ssl_err_val = static_cast<int>(ERR_get_error());
+    const boost::system::error_code ssl_err_code = boost::asio::error::ssl_errors(static_cast<int>(ssl_err_val));
+    MERROR("Failed to create SSL fingerprint: " << ERR_reason_error_string(ssl_err_val));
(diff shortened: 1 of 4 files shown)


jeffro256:
All the commits *should* compile at each step (only tested on my machine), so I recommend to step thru the commits one at a time to help reviewing if that is easier for you. Obviously I can squash them when it's time to merge

contrib/epee/src/net_ssl.cpp
```diff
@@ -641,6 +641,54 @@ bool ssl_options_t::handshake(
...
 }
 
+std::string get_hr_ssl_fingerprint(const X509 *cert, const EVP_MD *fdig)
+{
+  unsigned int j;
+  unsigned int n;
+  unsigned char md[EVP_MAX_MD_SIZE];
+  std::string fingerprint;
+
+  if (!X509_digest(cert, fdig, md, &n))
```
vtnerd, line 651: Check for `cert` and `fdig` against `nullptr`? Or does openssl do this check already (this is already in the codebase like this, so)..

vtnerd (approved):

contrib/epee/src/net_ssl.cpp
jeffro256: Yeah that can't hurt...

jeffro256: Done!

selsta (approved):

PR #8348, opened 2022-05-23
epee: http_server_handlers_map2 macros to use fully qualified names

jeffro256:
Quick patch which fixes the issue where if you use some macros from `http_server_handlers_map2.h` you have to be in the `epee` namespace or it doesn't compile. Now can remove `using namespace epee;` from header file `core_rpc_server.h`, which caused a couple of name qualifying mistakes down the line.


jeffro256:
Windows CI build error is unrelated and fixed by this PR: https://github.com/monero-project/monero/pull/8349

jeffro256:
rebased against master

selsta (approved):

PR #8354, opened 2022-05-24
net_ssl: Switch to elliptic curve SSL certificates

jeffro256:
I opened an eariler to PR to make our `net_ssl` RSA certificate generation code to comply with OpenSSL 3.0, but @hyc suggested that we switch to elliptic curve SSL certificates since they are smaller and faster. These PR also complies with OpenSSL 3.0.

### DIFF
--- a/contrib/epee/include/net/net_ssl.h
+++ b/contrib/epee/include/net/net_ssl.h
@@ -148,7 +148,17 @@ namespace net_utils
 	bool is_ssl(const unsigned char *data, size_t len);
 	bool ssl_support_from_string(ssl_support_t &ssl, boost::string_ref s);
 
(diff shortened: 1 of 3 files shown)


hyc:
I suggested we think about it. Haven't seen much other discussion yet.

selsta:
Both patches exist now, guess we can talk about it in the next dev meeting.

jeffro256:
@hyc Yes, sorry should've clarified "suggested". I definitely welcome discussion about why this is a bad idea or not.

selsta:
@vtnerd probably has an opinion on this too

selsta:
Can you force push here to rerun CI?

jeffro256:
rebased against master

jeffro256:
I couldn't find a reliable cross-platform solution that didn't use non-rigid cryptography

PR #8356, opened 2022-05-26
wallet_api: add scanTransactions function

selsta:
Introduced in #7312

Tested on the GUI side and working as expected.

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -1280,6 +1280,42 @@ bool WalletImpl::importOutputs(const string &filename)
     return true;
 }
 
+bool WalletImpl::scanTransactions(const std::vector<std::string> &txids)
+{
+    if (txids.empty())
+    {
+        setStatusError(string(tr("Failed to scan transactions: no transaction ids provided.")));
+        return false;
+    }
+
+    // Parse and dedup args
+    std::unordered_set<crypto::hash> txids_u;
+    for (const auto &s : txids)
+    {
+        crypto::hash txid;
+        if (!epee::string_tools::hex_to_pod(s, txid))
+        {
+            setStatusError(string(tr("Invalid txid specified: ")) + s);
+            return false;
+        }
+        txids_u.insert(txid);
+    }
+    std::vector<crypto::hash> txids_v(txids_u.begin(), txids_u.end());
+
+    try
+    {
+        m_wallet->scan_tx(txids_v);
+    }
+    catch (const std::exception &e)
+    {
+        LOG_ERROR("Failed to scan transaction: " << e.what());
+        setStatusError(string(tr("Failed to scan transaction: ")) + e.what());
+        return false;
+    }
+
+    return true;
+}
+
 void WalletImpl::addSubaddressAccount(const std::string& label)
 {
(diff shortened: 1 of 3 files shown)


tobtoht (approved):

src/wallet/api/wallet.cpp
```diff
@@ -1280,6 +1280,42 @@ bool WalletImpl::importOutputs(const string &filename)
...
+    {
+        crypto::hash txid;
+        if (!epee::string_tools::hex_to_pod(s, txid))
+        {
+            setStatusError(string(tr("Invalid txid specified: ")) + s);
+            return false;
+        }
+        txids_u.insert(txid);
+    }
+    std::vector<crypto::hash> txids_v(txids_u.begin(), txids_u.end());
```
jeffro256, line 1303: You can skip creating two containers (less memory, less copying) for the hashes if you use a `vector` for `txids_u` and call `.reserve()` before the `for` loop.

selsta: I just copied the simplewallet code. I think unordered set is used to avoid duplicates.

jeffro256: Ah, I see the `dedup` part now, is it bad if a TX gets scanned twice?

selsta: Scanning twice is definitely significantly slower than copying from a set to a vector. I think the current code is fine.

mj-xmr: That's the simplest way to dedup.

src/wallet/api/wallet.cpp
```diff
@@ -1280,6 +1280,42 @@ bool WalletImpl::importOutputs(const string &filename)
...
+    }
+
+    // Parse and dedup args
+    std::unordered_set<crypto::hash> txids_u;
+    for (const auto &s : txids)
+    {
+        crypto::hash txid;
+        if (!epee::string_tools::hex_to_pod(s, txid))
+        {
+            setStatusError(string(tr("Invalid txid specified: ")) + s);
```
mj-xmr, line 1298: Ignore if you wish but:

`std::string(tr` ?

src/wallet/api/wallet2_api.h
```diff
@@ -927,6 +927,13 @@ struct Wallet
      */
     virtual bool importOutputs(const std::string &filename) = 0;
 
+    /*!
+     * \brief scanTransactions - scan a list of transaction ids, this operation may reveal the txids to the remote node and affect your privacy
+     * \param txids            - list of transaction ids
+     * \return                 - true on success
```
mj-xmr, line 933: "True on success, meaning, that all transactions were valid"?

jeffro256 (approved):

src/wallet/api/wallet.cpp
selsta: if I would write new code yes, but every other occurrence in this file uses just `string` so I kept it consistent

src/wallet/api/wallet2_api.h
selsta: Success meaning all the transactions got successfully scanned.

PR #8338, opened 2022-05-17
GCC: Suppress -Wstringop-overflow= warnings in external/boost

jeffro256:
Resolves #8320

PR #8358, opened 2022-05-28
clang: warning fix for #8338

jeffro256:
Unlike with some other warnings, clang does not have a `stringop-overflow` group so it doesn't recognize the `#pragma GCC ...` directive in #8338

### DIFF
--- a/external/boost/archive/portable_binary_archive.hpp
+++ b/external/boost/archive/portable_binary_archive.hpp
@@ -44,12 +44,16 @@ reverse_bytes(signed char size, char *address){
     char * first = address;
     char * last = first + size - 1;
     for(;first < last;++first, --last){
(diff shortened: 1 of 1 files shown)


jeffro256:
@selsta Can you please double-check to make sure that this doesn't give a warning message with `clang`?

selsta:
see https://stackoverflow.com/questions/38499462/how-to-tell-clang-to-stop-pretending-to-be-other-compilers

you might need `__clang__`

jeffro256:
`clang` wants to be `gcc` so bad 😆 . Can you try it now @selsta? Thanks for slogging through this, I know this isn't what you want to be doing right now

selsta (approved):
Looks good now, no warnings with clang

Lissy93 (approved):

PR #8359, opened 2022-05-28
GCC: unused warnings

jeffro256:
This PR fixes 4 instances of `-Wunused...`

### DIFF
--- a/src/crypto/hash-extra-jh.c
+++ b/src/crypto/hash-extra-jh.c
@@ -36,7 +36,9 @@
 #include "jh.h"
 #include "hash-ops.h"
 
+#define JH_HASH_BITLEN HASH_SIZE * 8
+
 void hash_extra_jh(const void *data, size_t length, char *hash) {
-  int r = jh_hash(HASH_SIZE * 8, data, 8 * length, (uint8_t*)hash);
-  assert(SUCCESS == r);
+  // No need to check for failure b/c jh_hash only fails for invalid hash size
+  jh_hash(JH_HASH_BITLEN, data, 8 * length, (uint8_t*)hash);
 }
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -3418,6 +3418,10 @@ void wallet2::refresh(bool trusted_daemon, uint64_t start_height, uint64_t & blo
   uint64_t blocks_start_height;
   std::vector<cryptonote::block_complete_entry> blocks;
   std::vector<parsed_block> parsed_blocks;
+  // TODO moneromooo-monero says this about the "refreshed" variable:
+  // "I had to reorder some code to fix... a timing info leak IIRC. In turn, this undid something I had fixed before, ... a subtle race condition with the txpool.
+  // It was pretty subtle IIRC, and so I needed time to think about how to refix it after the move, and I never got to it."
+  // https://github.com/monero-project/monero/pull/6097
   bool refreshed = false;
   std::shared_ptr<std::map<std::pair<uint64_t, uint64_t>, size_t>> output_tracker_cache;
   hw::device &hwdev = m_account.get_device();
--- a/src/crypto/hash-extra-skein.c
+++ b/src/crypto/hash-extra-skein.c
@@ -34,7 +34,9 @@
 #include "hash-ops.h"
 #include "skein.h"
 
+#define SKEIN_HASH_BITLEN HASH_SIZE * 8
+
 void hash_extra_skein(const void *data, size_t length, char *hash) {
-  int r = skein_hash(8 * HASH_SIZE, data, 8 * length, (uint8_t*)hash);
-  assert(SKEIN_SUCCESS == r);
+  // No need to check for failure b/c skein_hash only fails for invalid hash size
+  skein_hash(SKEIN_HASH_BITLEN, data, 8 * length, (uint8_t*)hash);
 }
--- a/src/device/device_ledger.cpp
+++ b/src/device/device_ledger.cpp
@@ -43,6 +43,10 @@ namespace hw {
 
   #ifdef WITH_DEVICE_LEDGER
 
+    namespace {
+        bool apdu_verbose =true;
+    }
+
     #undef MONERO_DEFAULT_LOG_CATEGORY
     #define MONERO_DEFAULT_LOG_CATEGORY "device.ledger"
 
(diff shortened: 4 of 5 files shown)


src/wallet/wallet2.cpp
```diff
@@ -5127,7 +5124,6 @@ std::string wallet2::exchange_multisig_keys(const epee::wipeable_string &passwor
     expanded_msgs.emplace_back(msg);
 
   // reconstruct multisig account
-  crypto::public_key dummy;
```
selsta, line 5130: This one is already fixed in one of the multisig PRs. Can you undo this change?

jeffro256: yup

src/wallet/wallet2.cpp
```diff
@@ -3476,7 +3475,6 @@ void wallet2::refresh(bool trusted_daemon, uint64_t start_height, uint64_t & blo
       if (!first && blocks.empty())
       {
         m_node_rpc_proxy.set_height(m_blockchain.size());
-        refreshed = true;
```
mj-xmr, line 3479: I think this was set by the author to facilitate easier debugging. 

I personally don't need it, but just saying that's the reason I haven't removed it myself.

mj-xmr (approved):
It looks good, but please gather 2nd or even 3rd opinion about the wallet2.

src/crypto/hash-extra-jh.c
```diff
@@ -38,5 +38,6 @@
 
 void hash_extra_jh(const void *data, size_t length, char *hash) {
   int r = jh_hash(HASH_SIZE * 8, data, 8 * length, (uint8_t*)hash);
+  (void) r; // silence -Wunused-variable
```
mj-xmr, line 41: Before anybody mentions inconsistency in the casting style:

This is C, not C++, so `static_cast<>` can't be used.

hyc: So these are only unused because assert() is being no-op'd. These hash functions don't allocate memory so have no reason to fail. Maybe we can just get rid of `r` and the asserts too?

src/wallet/wallet2.cpp
hyc: That was my thought as well. I don't see any need to remove it.

src/crypto/hash-extra-jh.c
jeffro256: Yeah since `hashbitlen` is a compile-time constant, there should be no reason for `jh_hash` to not return `SUCCESS`

src/wallet/wallet2.cpp
jeffro256: @hyc pushed changes to keep the flag in and remove the warning

src/crypto/hash-extra-jh.c
jeffro256: Okay changes are pushed for that

src/wallet/wallet2.cpp
selsta: @jeffro256 "Easier debugging" was speculation. As it says in mooo's comment, he wanted to clean it up but probably forgot about it. We can ask him again to be sure but just silencing doesn't make much sense here.

jeffro256: Pinging @moneromooo-monero

src/crypto/hash-extra-jh.c
hyc: Hi, sorry for the wishy-washiness but I'm gonna take back what I said and go with your initial patch. Safer to leave the asserts in place so that they're present in Debug builds, and will catch any accidental misuses of the code in the future.

jeffro256: Should I make the `assert` statements conditional on whether `NDEBUG` is defined?

hyc: OK, sounds good.

src/wallet/wallet2.cpp
selsta:
```
12:34 <moneromooo> I had to reorder some code to fix... a timing info leak IIRC. In turn, this undid something I had fixed before, which I believe was a subtle race condition with the txpool.
12:35 <moneromooo> It was pretty subtle IIRC, and so I needed time to think about how to refix it after the move, and I never got to it. I do not remember the details of the race.
12:35 <moneromooo> I don't think I'm likely to ever get to it now.
```

So either we remove the unused variable or we keep it. Silencing doesn't make sense in my opinion.

mj-xmr: +1 for commenting out rather than silencing.

There will surely be a lot of debugging to do there in future, but then it can be uncommented on demand, still not causing warnings to everybody else and not having to resort to compiler hacks, which might not be cross-platform.

hyc: Probably should add some of mooo's commentary as a comment in the source too. Saves digging thru github issues to find out what it's all about.

jeffro256: done

src/wallet/wallet2.cpp
```diff
@@ -3418,7 +3418,10 @@ void wallet2::refresh(bool trusted_daemon, uint64_t start_height, uint64_t & blo
   uint64_t blocks_start_height;
   std::vector<cryptonote::block_complete_entry> blocks;
   std::vector<parsed_block> parsed_blocks;
-  bool refreshed = false;
+  // moneromooo-monero says this about the "refreshed" variable:
+  // "I had to reorder some code to fix... a timing info leak IIRC. In turn, this undid something I had fixed before, ... a subtle race condition with the txpool.
+  // It was pretty subtle IIRC, and so I needed time to think about how to refix it after the move, and I never got to it."
```
selsta, line 3423:
```suggestion
  // TODO: moneromooo-monero says this about the "refreshed" variable:
  // "I had to reorder some code to fix... a timing info leak IIRC. In turn, this undid something I had fixed before, ... a subtle race condition with the txpool.
  // It was pretty subtle IIRC, and so I needed time to think about how to refix it after the move, and I never got to it."
  // https://github.com/monero-project/monero/pull/6097
```

`TODO` so that it can be quickly found via `git grep`. Also adding a URL for context.

jeffro256: done!

moneromooo-monero:
At this point I'm unlikely to spend a lot of time thinking about it and fixing it well. So if people want to kill the warning, be my guest. I'd be better if people wanted to fix it properly instead of just killing the warning, but that's up to them.

selsta:
@jeffro256 Can you keep the `refreshed` variable untouched and only add the comment? We can keep the warning as a reminder for someone to look into, if it stays unfixed for another year then we can decide what to do next with it.

Sorry for this huge discussion :)

jeffro256:
@selsta done!

mj-xmr (approved):

PR #8361, opened 2022-05-29
oaes_lib: gut everything that is unused

jeffro256:
Remove all code conditional on the `OAES_DEBUG` and `OAES_HAS_ISAAC` macro definitions, as well as all code not relevant to `oaes_alloc`, `oaes_free`, `oaes_key_import`, and `oaes_key_import_data`. These functions are used in `slow-hash.c` and nowhere else.

Some notes:
* I was able to remove some fields from `oaes_ctx`
* `oaes_alloc` and `oaes_free` now no longer seed the `srand`/`rand` API, but this is fine because it is no longer used in the implementation.
* This also fixes a deprecatation warning for `ftime`, which is what started this PR in the first place


hyc:
As a general rule, we should never unilaterally modify externally derived libraries unless there's a critical fix needed, etc. The only reason to modify these should be to incorporate changes from upstream. Making our own modifications will impede merging from upstream. (Upstream here is https://github.com/jhjin/OpenAES)

PR #8371, opened 2022-06-01
Improve consistency between on_money_received and on_money_received_unconfirmed

kayabaNerve:
unconfirmed solely uses a - b, and received now accepts b so it can provide more detailed logs on what occurred (printing a - b, yet also a and b themselves).

The updated print statement ("a yet b") was discussed with mooo. This PR has not been reviewed by them however.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -5673,14 +5673,18 @@ void simple_wallet::on_new_block(uint64_t height, const cryptonote::block& block
     m_refresh_progress_reporter.update(height, false);
 }
 //----------------------------------------------------------------------------------------------------
-void simple_wallet::on_money_received(uint64_t height, const crypto::hash &txid, const cryptonote::transaction& tx, uint64_t amount, const cryptonote::subaddress_index& subaddr_index, bool is_change, uint64_t unlock_time)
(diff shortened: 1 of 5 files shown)


moneromooo-monero (approved):
Looks OK. While looking, I saw that there's no existing on_money_received in the case where the output received is smaller in amount than the existing output with the same pubkey. It should probably call with amount==burnt in that case.

kayabaNerve:
That actually is the current behavior. While we could extend with that functionality, it's somewhat pointless (it's dead data that at best helps wallet devs debug, at worst is misinterpreted) and would cause underflows unless that's also handled. If we call with amount == burnt, that'd also be incorrect when the actual output amount is dust. While I'm not against adding handling for this case if beneficial, I question if it's beneficial, and I'd hope to have it labelled out of scope for this PR specifically.

moneromooo-monero:
It's pointless but technically more correct.
If burnt==received as I suggested, it cannot cause underflows I think. At worst confusing.
I do not understand "with amount == burnt, that'd also be incorrect when the actual output amount is dust".
I'm not asking you to do it since I OK'd the patch.

kayabaNerve:
Oh. Sorry.

> It should probably call with amount==burnt in that case.

I took this as calling it with amount = burnt, as in, the past output's amount, not burnt = received like you just said. Your statement now makes sense and shouldn't be an issue for a followup :) Sorry for misunderstanding.

j-berman (approved):

PR #8381, opened 2022-06-08
Fixed get_block_template_backlog performance

SChernykh:
Before the fix, it processed all transactions in the mempool which could be very slow when mempool grows to several MBs in size. I observed `get_block_template_backlog` taking up to 15 seconds of CPU time under high mempool load.

After the fix, only transactions that can potentially be mined in the next block will be processed (a bit more than the current block median weight).

### DIFF
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -917,26 +917,61 @@ namespace cryptonote
   {
     CRITICAL_REGION_LOCAL(m_transactions_lock);
     CRITICAL_REGION_LOCAL1(m_blockchain);
-    const relay_category category = include_sensitive ? relay_category::all : relay_category::broadcasted;
-    backlog.reserve(m_blockchain.get_txpool_tx_count(include_sensitive));
-    txpool_tx_meta_t tmp_meta;
-    m_blockchain.for_all_txpool_txes([this, &backlog, &tmp_meta](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref *bd){
-      transaction tx;
-      if (!(meta.pruned ? parse_and_validate_tx_base_from_blob(*bd, tx) : parse_and_validate_tx_from_blob(*bd, tx)))
-      {
-        MERROR("Failed to parse tx from txpool");
-        // continue
-        return true;
-      }
-      tx.set_hash(txid);
 
-      tmp_meta = meta;
+    std::vector<tx_block_template_backlog_entry> tmp;
+    uint64_t total_weight = 0;
 
-      if (is_transaction_ready_to_go(tmp_meta, txid, *bd, tx))
-        backlog.push_back({txid, meta.weight, meta.fee});
-
+    // First get everything from the mempool, filter it later
+    m_blockchain.for_all_txpool_txes([&tmp, &total_weight](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref*){
+      tmp.emplace_back(tx_block_template_backlog_entry{txid, meta.weight, meta.fee});
+      total_weight += meta.weight;
       return true;
-    }, true, category);
+    }, false, include_sensitive ? relay_category::all : relay_category::broadcasted);
+
+    // Limit backlog to 112.5% of current median weight. This is enough to mine a full block with the optimal block reward
+    const uint64_t median_weight = m_blockchain.get_current_cumulative_block_weight_median();
+    const uint64_t max_backlog_weight = median_weight + (median_weight / 8);
+
+    // If the total weight is too high, choose the best paying transactions
+    if (total_weight > max_backlog_weight)
+      std::sort(tmp.begin(), tmp.end(), [](const auto& a, const auto& b){ return a.fee * b.weight > b.fee * a.weight; });
+
+    backlog.clear();
+    uint64_t w = 0;
+
+    std::unordered_set<crypto::key_image> k_images;
+
+    for (const tx_block_template_backlog_entry& e : tmp)
+    {
+      try
+      {
+        txpool_tx_meta_t meta;
+        if (!m_blockchain.get_txpool_tx_meta(e.id, meta))
+          continue;
+
+        cryptonote::blobdata txblob;
+        if (!m_blockchain.get_txpool_tx_blob(e.id, txblob, relay_category::all))
+          continue;
+
+        cryptonote::transaction tx;
+        if (is_transaction_ready_to_go(meta, e.id, txblob, tx))
+        {
+          if (have_key_images(k_images, tx))
+            continue;
+          append_key_images(k_images, tx);
+
+          backlog.push_back(e);
+          w += e.weight;
+          if (w > max_backlog_weight)
+            break;
+        }
+      }
+      catch (const std::exception &e)
+      {
+        MERROR("Failed to check transaction readiness: " << e.what());
+        // continue, not fatal
+      }
+    }
   }
   //------------------------------------------------------------------
   void tx_memory_pool::get_transaction_stats(struct txpool_stats& stats, bool include_sensitive) const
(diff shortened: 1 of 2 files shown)


src/cryptonote_core/tx_pool.cpp
```diff
@@ -917,26 +917,53 @@ namespace cryptonote
...
 
+    // First get everything from the mempool, filter it later
+    m_blockchain.for_all_txpool_txes([&tmp, &total_weight](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref*){
+      tmp.emplace_back(tx_block_template_backlog_entry{txid, meta.weight, meta.fee});
+      total_weight += meta.weight;
       return true;
-    }, true, category);
+    }, true, include_sensitive ? relay_category::all : relay_category::broadcasted);
+
+    // Limit backlog to 112.5% of current median weight. This is enough to mine a full block with the optimal block reward
```
UkoeHB, line 931: The max block weight is 2x the median weight, which can occur if fee pressure is high enough. I'd recommend 2.25x the median weight.

UkoeHB (changes requested):

src/cryptonote_core/tx_pool.cpp
UkoeHB: Or you could use the block penalty algo to choose txs.

SChernykh: I need to think about it more. The problem is that `is_transaction_ready_to_go` is a very slow function (it checks all RCT signatures), which is why this PR was created in the first place. So setting as low limit as reasonably possible is the goal here.

UkoeHB: Maybe this is a deeper architecture problem... txs in the mempool should ALWAYS be ready to go. If there are any conditions where a tx isn't ready to go, that implies a problem with how the mempool is being managed internally.

SChernykh: Transactions can get invalid after a reorg, or even after a new block is received in case this block has conflicting transactions

UkoeHB: At minimum, a tx in the mempool should have valid proofs. The semantics could be invalid due to a hardfork (for example, changing the max output count), and the references could be invalid due to reorgs (ring member references and key images). Sadly, `check_tx_inputs()` is a do-everything function so you have to call it to check _anything_.

UkoeHB: The 'ideal' is out of scope for this PR though. Maybe an API endpoint for 'request more txs not in this tx set' would help you.
1. use the notifier to get an initial batch of txs
2. start mining
3. request small additional batches on a separate thread until you can't increase the size of your tx blob any more without incurring a penalty

src/cryptonote_core/tx_pool.cpp
```diff
@@ -917,26 +917,53 @@ namespace cryptonote
...
+    const uint64_t median_weight = m_blockchain.get_current_cumulative_block_weight_median();
+    const uint64_t max_backlog_weight = median_weight + (median_weight / 8);
+
+    // If the total weight is too high, choose the best paying transactions
+    if (total_weight > max_backlog_weight)
+      std::sort(tmp.begin(), tmp.end(), [](const auto& a, const auto& b){ return a.fee * b.weight > b.fee * a.weight; });
```
moneromooo-monero, line 937: There's already a list of txes sorted by fee/byte, avoids enumerating all txes from the db above.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -917,26 +917,53 @@ namespace cryptonote
...
+    m_blockchain.for_all_txpool_txes([&tmp, &total_weight](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref*){
+      tmp.emplace_back(tx_block_template_backlog_entry{txid, meta.weight, meta.fee});
+      total_weight += meta.weight;
       return true;
-    }, true, category);
+    }, true, include_sensitive ? relay_category::all : relay_category::broadcasted);
+
+    // Limit backlog to 112.5% of current median weight. This is enough to mine a full block with the optimal block reward
+    const uint64_t median_weight = m_blockchain.get_current_cumulative_block_weight_median();
+    const uint64_t max_backlog_weight = median_weight + (median_weight / 8);
```
moneromooo-monero, line 933: Technically, if all txes in the txpool pay over the odds, you can go to 2x. Not sure you care here.

moneromooo-monero (approved):
Just a couple minor things, still good to go.

src/cryptonote_core/tx_pool.cpp
SChernykh: Are you talking about `m_txs_by_fee_and_receive_time`? Yes, it will be better to use it but I'm not sure it's up to date at the time this code is executed (right after a new block has been added to the blockchain, in the end of `handle_block_to_main_chain`). Can you clarify?

SChernykh: On the other hand, even though it will be running faster, using `m_txs_by_fee_and_receive_time` will require copy-pasting a lot of code from `fill_block_template` and I prefer to have cleaner code here. 90% of the run time will be spent inside `is_transaction_ready_to_go` anyway.

SChernykh: I've added a commit which should make `for_all_txpool_txes` call much faster, so it shouldn't be an issue.

src/cryptonote_core/tx_pool.cpp
SChernykh: I care more about sending data to miner faster, so 1.125x is a compromise value. As I wrote above, miner will get new transaction notifications anyway and will grow its transaction pool above 1.125x.

moneromooo-monero (approved):

src/cryptonote_core/tx_pool.cpp
mj-xmr: Is median_weight always divisible by 8?

If not, then please do:

`const uint64_t max_backlog_weight = median_weight + static_cast<uint64_t>(std::roundl(median_weight / static_cast<long double>(8));`

src/cryptonote_core/tx_pool.cpp
```diff
@@ -917,26 +917,59 @@ namespace cryptonote
...
-        return true;
-      }
-      tx.set_hash(txid);
+        txpool_tx_meta_t meta;
+        if (!m_blockchain.get_txpool_tx_meta(e.id, meta))
+          continue;
 
-      tmp_meta = meta;
+        cryptonote::transaction tx;
+        cryptonote::blobdata txblob = m_blockchain.get_txpool_tx_blob(e.id, relay_category::all);
```
mj-xmr, line 953: Is `cryptonote::blobdata & txblob` possible?

src/cryptonote_core/tx_pool.cpp
```diff
@@ -917,26 +917,59 @@ namespace cryptonote
...
+        if (is_transaction_ready_to_go(meta, e.id, txblob, tx))
+        {
+          if (have_key_images(k_images, tx))
+            continue;
+          append_key_images(k_images, tx);
 
-      return true;
-    }, true, category);
+          backlog.push_back(e);
+          w += e.weight;
```
mj-xmr, line 962: Perhaps risky but might be worth it to flip these 2 lines and emplace back the `e`, if this is a bottleneck.

src/cryptonote_core/tx_pool.cpp
SChernykh: This rewrite will just change rounding mode from truncation (integer division) to nearest (FP division and roundl). Why would it matter here?

mj-xmr: If it doesn't matter, then it's OK.

moneromooo-monero (approved):

PR #8382, opened 2022-06-09
renamed and added comments to get_destination_view_key_pub(..)

paulshapiro:
This function was somewhat ambiguously named and its function is a bit difficult to interpret at first glance. It's only used in two locations for the purpose of selecting to which of the destinations a payment ID should be encrypted so I thought I'd submit an improvement to its naming and provide some comments about its implementation.
It would probably be better yet to split this function into two - one which selects the pid encr dest, and another to extract the view key from it - though there is a conflation in the functionality here related to the return and interpretation of a null key.


src/cryptonote_core/cryptonote_tx_utils.cpp
```diff
@@ -181,26 +181,27 @@ namespace cryptonote
...
     for (const auto &i : destinations)
     {
-      if (i.amount == 0)
+      if (i.amount == 0) // if this is a destination for fabricating a dummy output (0 amount), then the dummy dest would never be the recipient of the payment id
         continue;
-      if (change_addr && i.addr == *change_addr)
+      if (change_addr && i.addr == *change_addr) // ignore destinations with the change addr - and only if no other encryption destinations have been discovered and a change addr exists, then use it, at the end of this function
         continue;
-      if (i.addr == addr)
+      if (i.addr == addr) // if i.addr is null,null OR if i.addr is a duplicate of the one which has already been selected on a prior iteration - then at least this has the function of not incrementing count; as an aside, afaict, no destinations should be allowed to have a null .addr, so this if's condition probably is only intended to be true when i.addr has been found on a prior iteration
```
mj-xmr, line 195: This line might enjoy being broken into two.

mj-xmr (approved):
I can't say with 100% certainty if all the comments are factually correct, but I can't see any regressions/typos.

Such changes are greatly needed IMO.

src/cryptonote_core/cryptonote_tx_utils.h
```diff
@@ -117,7 +117,7 @@ namespace cryptonote
   };
 
   //---------------------------------------------------------------
-  crypto::public_key get_destination_view_key_pub(const std::vector<tx_destination_entry> &destinations, const boost::optional<cryptonote::account_public_address>& change_addr);
+  crypto::public_key get_payment_id_encr_dest_view_key_pub(const std::vector<tx_destination_entry> &destinations, const boost::optional<cryptonote::account_public_address>& change_addr);
```
jeffro256, line 120: Could we please get a comment here describing the overall purpose of the function since we're renaming?

paulshapiro: That's basically the point of proper function naming, where possible, i.e. in this case I think we can get away without commenting the function, but I would be open to adding one?

mj-xmr: The function is named much better, and Robert C. Martin promotes such kind of naming in his books (*), but a bit of more human readable text wouldn't hurt I think.

(*) A digression: he'd also definitely say, that such a function has to be broken into many smaller functions, that *do one thing* each

paulshapiro: Since we're talking about it, I'd be happy to take a crack at splitting this function into its two (IMO) factors. Actually should be fairly easy. But it will create a higher overhead in review on a highly critical codepath. But.. also, hopefully is fairly trivial and hopefully has enough qualified eyes on reviewing it.

mj-xmr: Exactly for that reason, I think a separate PR would increase the chance of success :)

paulshapiro: Ok, cool. I will leave it for later then.

PR #8383, opened 2022-06-12
Move DB lock guard to later stage

SChernykh:
It's important to take the DB lock after m_tx_pool lock when get_block_template RPC is run in parallel with a new block being added, otherwise it will return the old data.


SChernykh:
@j-berman please review.

selsta:
I think this can be closed? We will instead revert the other patch for now.

PR #8387, opened 2022-06-15
Label multisig as unsafe, not experimental, as #8149 isn't merged

kayabaNerve:
These aren't maybes. These are facts which need to be communicated. "experimental" was a term chosen expecting #8149 to be merged. Since it hasn't been yet, it's being misapplied to the current multisig code while simultaneously implying #8149 (the experimental code) was merged. This clarifies expectations and ensures a lack of confusion/misunderstandings.

The other option given the current state of multisig would be to disable it at the compilation level, which would provoke distribution of modified binaries which is its own mess. This is the path of least resistance to ensure proper communication in the next release despite #8149's current non-presence.

Update to #8328. Prompted by the most recent MRL meeting.


UkoeHB (approved):
Until @8149 is merged, this would communicate the state of multisig more accurately.

jeffro256 (approved):

TheCharlatan (approved):

PR #8388, opened 2022-06-15
Chunk /gettransactions to avoid hitting restricted RPC limit

tobtoht:
Fixes the following issue when mempool contains more than 100 new transactions when connected to a restricted node: 

`Error calling gettransactions daemon RPC: r 1, status Too many transactions requested in restricted mode`

@selsta

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -3159,14 +3159,18 @@ void wallet2::update_pool_state(std::vector<std::tuple<cryptonote::transaction, 
     }
   }
 
-  // get those txes
-  if (!txids.empty())
+  // get_transaction_pool_hashes.bin may return more transactions than we're allowed to request in restricted mode
+  const size_t SLICE_SIZE = 100; // RESTRICTED_TRANSACTIONS_COUNT as defined in rpc/core_rpc_server.cpp
+  for (size_t offset = 0; offset < txids.size(); offset += SLICE_SIZE)
   {
     cryptonote::COMMAND_RPC_GET_TRANSACTIONS::request req;
     cryptonote::COMMAND_RPC_GET_TRANSACTIONS::response res;
-    for (const auto &p: txids)
-      req.txs_hashes.push_back(epee::string_tools::pod_to_hex(p.first));
-    MDEBUG("asking for " << txids.size() << " transactions");
+
+    const size_t n_txids = std::min<size_t>(SLICE_SIZE, txids.size() - offset);
+    for (size_t n = offset; n < (offset + n_txids); ++n) {
+      req.txs_hashes.push_back(epee::string_tools::pod_to_hex(txids.at(n).first));
+    }
+    MDEBUG("asking for " << req.txs_hashes.size() << " transactions");
     req.decode_as_json = false;
     req.prune = true;
 
@@ -3183,7 +3187,7 @@ void wallet2::update_pool_state(std::vector<std::tuple<cryptonote::transaction, 
     MDEBUG("Got " << r << " and " << res.status);
     if (r && res.status == CORE_RPC_STATUS_OK)
     {
-      if (res.txs.size() == txids.size())
+      if (res.txs.size() == req.txs_hashes.size())
       {
         for (const auto &tx_entry: res.txs)
         {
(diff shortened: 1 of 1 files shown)


src/wallet/wallet2.cpp
```diff
@@ -3159,14 +3159,18 @@ void wallet2::update_pool_state(std::vector<std::tuple<cryptonote::transaction,
     }
   }
 
-  // get those txes
-  if (!txids.empty())
+  // get_transaction_pool_hashes.bin may return more transactions than we're allowed to request in restricted mode
+  const size_t SLICE_SIZE = 100; // RESTRICTED_TRANSACTIONS_COUNT as defined in rpc/core_rpc_server.cpp
```
jeffro256, line 3163: Could we detect if the daemon is in restricted mode and chunk accordingly, as opposed to defaulting to the (rather low) value of 100?

plowsof: i think this is what happens already here https://github.com/monero-project/monero/blob/9750e1fa103539b3e533455500610aae76e253a5/src/rpc/core_rpc_server.cpp#L888

jeffro256: Right, but I mean but if the daemon is in unrestricted mode, then we can ask for more at once, which would definitely reduce overhead if/when the mempool gets to be the size of (for example) Bitcoin's with >1000 txs per block

selsta:
```
LOG_PRINT_L0("Expected " << txids.size() << " tx(es), got " << res.txs.size());
```

should this log message be updated? Maybe to something like

```
LOG_PRINT_L0("Expected " << n_txids << " out of " << txids.size() << " tx(es), got " << res.txs.size());
```

Because otherwise the log is a bit confusing.

```
2022-06-20 13:47:25.050	     0x1051b8580	DEBUG	wallet.wallet2	src/wallet/wallet2.cpp:3173	asking for 5 transactions
2022-06-20 13:47:25.098	     0x1051b8580	TRACE	net	contrib/epee/include/net/net_helper.h:460	READ ENDS: Success. bytes_tr: 5134
2022-06-20 13:47:25.098	     0x1051b8580	TRACE	net.http	contrib/epee/include/net/http_client.h:648	http_stream_filter::parse_cached_header(*)
2022-06-20 13:47:25.098	     0x1051b8580	DEBUG	wallet.wallet2	src/wallet/wallet2.cpp:3187	Got 1 and OK
2022-06-20 13:47:25.098	     0x1051b8580	WARNING	wallet.wallet2	src/wallet/wallet2.cpp:3226	Expected 59 tx(es), got 5
```

jeffro256:
> ```
> LOG_PRINT_L0("Expected " << txids.size() << " tx(es), got " << res.txs.size());
> ```
> 
> should this log message be updated? Maybe to something like
> 
> ```
> LOG_PRINT_L0("Expected " << n_txids << " out of " << txids.size() << " tx(es), got " << res.txs.size());
> ```
> 
> Because otherwise the log is a bit confusing.
> 
> ```
> 2022-06-20 13:47:25.050	     0x1051b8580	DEBUG	wallet.wallet2	src/wallet/wallet2.cpp:3173	asking for 5 transactions
> 2022-06-20 13:47:25.098	     0x1051b8580	TRACE	net	contrib/epee/include/net/net_helper.h:460	READ ENDS: Success. bytes_tr: 5134
> 2022-06-20 13:47:25.098	     0x1051b8580	TRACE	net.http	contrib/epee/include/net/http_client.h:648	http_stream_filter::parse_cached_header(*)
> 2022-06-20 13:47:25.098	     0x1051b8580	DEBUG	wallet.wallet2	src/wallet/wallet2.cpp:3187	Got 1 and OK
> 2022-06-20 13:47:25.098	     0x1051b8580	WARNING	wallet.wallet2	src/wallet/wallet2.cpp:3226	Expected 59 tx(es), got 5
> ```

To piggyback off of this, should this log statement be put inside the loop now since there are now multiple requests which could each individually go wrong?

selsta:
@jeffro256 it is already in the loop, or am I misunderstanding your suggestion?

selsta (approved):
Code reviewed and tested.

PR #8390, opened 2022-06-17
Add fish shell completions for monerod and monero-wallet-cli

LeoNero:
This PR adds fish shell completions for `monerod` and `monero-wallet-cli`. I am starting to use those binary files, and using `--help` all the time gets boring and slow. Maybe others that use Fish will find this useful too.

Basically, this is the output of `--help` adapted for Fish completions. I did a simple script for it and then manually imploved some things, such as adding `-a` flags, etc.


selsta (commented):
I wonder if `utils/` makes more sense, the `systemd` example script is in there.

Also can you squash this to a single commit and name it something like: `utils: add fish shell completion`, that would be more in line with how commits are named in this project.

LeoNero:
@selsta I'll do that when I get home. Thanks!

LeoNero:
@selsta done :)

selsta (approved):
I use fish so that's useful for me, thanks :)

LeoNero:
@selsta I just added completions for monero-wallet-rpc too.

LeoNero:
lmao just added it. My bad

LeoNero:
actually wait a second

utils/fish/monero-wallet-cli.fish
```diff
@@ -0,0 +1,61 @@
+complete -c monero-wallet-cli -f
```
selsta, line 1:
```suggestion
complete -c monero-wallet-cli -f
```

`monero-wallet-rpc` and `monerod` don't have this line, can you explain it?

LeoNero: this means that file completions for the flags are disabled by default and, if a flag accepts a file path as an argument, we then need to use `-F` in its definition. I will put this line in the other files for consistency

LeoNero: Actually, it is also important for the file completionst that find files with a specific suffix. SO it is necessary, will add it

LeoNero:
Done :)

selsta:
Screenshot 2022-07-01 at 01 04 08

Should this only suggest .json files? If yes, it does not seem to work yet.

LeoNero:
Yea, I saw that. Per documentation of `__fish_complete_suffix`, it should just show the `.json` files first, and the other files after.

I think it can be improved, but I did not want to spend much time on that. I'll see if there is a fix to it.

selsta (approved):

PR #8391, opened 2022-06-18
simplewallet: print usage when given no args

hinto-janaiyo:
Print usage for these functions when given 0 arguments. All functions that expect arguments do this, except for these 4.

```
transfer
locked_transfer
locked_sweep_all
sweep_below
```

BEFORE:
```
[wallet 48hint]: transfer
Error: wrong number of arguments
```
AFTER:
```
[wallet 48hint]: transfer
Error: usage: transfer [index=<N1>[,<N2>,...]] [<priority>] [<ring_size>] (<URI> | <address> <amount>) [<payment_id>]
```


src/simplewallet/simplewallet.cpp
```diff
@@ -7670,11 +7685,13 @@ bool simple_wallet::sweep_below(const std::vector<std::string> &args_)
   if (args_.size() < 1)
   {
     fail_msg_writer() << tr("missing threshold amount");
+    PRINT_USAGE(USAGE_SWEEP_BELOW);
     return true;
   }
   if (!cryptonote::parse_amount(below, args_[0]))
   {
     fail_msg_writer() << tr("invalid amount threshold");
+    PRINT_USAGE(USAGE_SWEEP_BELOW);
```
selsta, line 7694: I would remove this as it doesn't fit the "print usage when given no args" commit title and also isn't consistent when entering an invalid amount in other places.

selsta (approved):
Correct code and tested.

tobtoht (approved):

LeoNero (approved):
LGTM

PR #8392, opened 2022-06-18
cryptonote_basic: do not print name of a possibly invalid type

moneromooo-monero:


moneromooo-monero:
Not sufficient. There will be an update.

moneromooo-monero:
That wasn't the bug, I got the repro data now. After double checking serialization, the type has to be valid.

PR #8395, opened 2022-06-20
wallet2: fix invalid memory access importing invalid outputs

moneromooo-monero:


j-berman:
From DM's with moo so this doesn't get lost, I proposed the following change instead:

```
index e051946ad..fba0241f1 100644
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -349,6 +349,8 @@ private:
       uint64_t amount() const { return m_amount; }
       const crypto::public_key get_public_key() const {
         crypto::public_key output_public_key;
+        THROW_WALLET_EXCEPTION_IF(m_tx.vout.size() <= m_internal_output_index,
+          error::wallet_internal_error, "Too few outputs, outputs may be corrupted");
         THROW_WALLET_EXCEPTION_IF(!get_output_public_key(m_tx.vout[m_internal_output_index], output_public_key),
           error::wallet_internal_error, "Unable to get output public key from output");
```

AFAICT the only way to trigger this PR's invalid memory access in `import_outputs` is if a buggy or malicious output exporter has already gotten the user to import a tx with fewer outputs than the tx really has (or to import an `m_internal_output_index` that is too high). It will take more investigation to be certain a user can safely recover their already imported outputs after doing this, which seems out of scope for this PR.

My proposed exception above prevents the wallet from crashing in this circumstance, provides an error description, and protects other callers of `get_public_key()`. I probably should have included the above exception when I added this function in the first place.

PR #8404, opened 2022-06-24
wallet2: remove obsolete rpc version check

tobtoht:
It's been a couple hardforks since this check was added, all daemons support `/get_output_distribution.bin`.


selsta (approved):
Correct, introduced in https://github.com/monero-project/monero/commit/41f727ce42d0fce9e40e4da7bf4abda92eef1016, we can safely remove it now.

jeffro256 (approved):

PR #3419, opened 2018-03-16
wallet: make the segregation height settable

moneromooo-monero:
via user setting first, then DNS TXT record, hardcoded fallback

PR #8408, opened 2022-06-27
wallet2: don't use DNS to obtain segregation heights

### LINKED ISSUE #8407: Stop using MoneroPulse checkpointing and making unnecessary DNS queries unless explicitly enabled in wallet2
In finding a major (but rare) UX issue (transaction generation taking 20s+) was related to DNS, I also was pointed to the core problem being that wallet2 makes many DNS queries before it processes a transaction. These queries seem to go to many `moneropulse.*` domains (related to old checkpointing logic, more [here](https://monerodocs.org/infrastructure/monero-pulse/) and to a set of hard-coded DNS servers (found [here](https://github.com/monero-project/monero/blob/9a124f681119855949f6406ecd69c2ad91da9770/src/common/dns_utils.cpp#L46-L53)).

These queries happen even if using IP address for the selected node, and even if not sending to an OpenAlias address. Not only are these causing major UX issues for people who have strict capture/handling of DNS traffic (like DoH users, VPN users, and others), but it also allows trivial fingerprinting of Monero users along with pin-point accuracy to determine when a specific IP address generates (and likely sends) a transaction by a given DNS server along with all MoneroPulse operators (whoever they are).

This behavior should be moved to a manual flag (for those who still want MoneroPulse checkpointing for some reason) and disabled by default except for two specific use-cases:

1. A user is using a DNS address for their selected node
  - This should be handled by the devices DNS anyways, not by wallet2
2. A user enters an OpenAlias address in the address box of a given wallet
  - This also should be handled by the device itself and not wallet2

Here are some of the places these DNS addresses and servers are in the code:

- https://github.com/monero-project/monero/blob/master/src/wallet/wallet2.cpp
- https://github.com/monero-project/monero/blob/master/src/checkpoints/checkpoints.cpp
- https://github.com/monero-project/monero/blob/master/src/common/dns_utils.cpp

tobtoht:
Related: #8407

There are no other places where DNS is used in wallet2, so the --no-dns flag can be removed.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -14022,43 +14022,6 @@ uint64_t wallet2::get_segregation_fork_height() const
   if (m_segregation_height > 0)
     return m_segregation_height;
 
-  if (m_use_dns && !m_offline)
-  {
-    // All four MoneroPulse domains have DNSSEC on and valid
-    static const std::vector<std::string> dns_urls = {
-        "segheights.moneropulse.org",
-        "segheights.moneropulse.net",
-        "segheights.moneropulse.co",
-        "segheights.moneropulse.se"
-    };
-
-    const uint64_t current_height = get_blockchain_current_height();
-    uint64_t best_diff = std::numeric_limits<uint64_t>::max(), best_height = 0;
-    std::vector<std::string> records;
-    if (tools::dns_utils::load_txt_records_from_dns(records, dns_urls))
-    {
-      for (const auto& record : records)
-      {
-        std::vector<std::string> fields;
-        boost::split(fields, record, boost::is_any_of(":"));
-        if (fields.size() != 2)
-          continue;
-        uint64_t height;
-        if (!string_tools::get_xtype_from_string(height, fields[1]))
-          continue;
-
-        MINFO("Found segregation height via DNS: " << fields[0] << " fork height at " << height);
-        uint64_t diff = height > current_height ? height - current_height : current_height - height;
-        if (diff < best_diff)
-        {
-          best_diff = diff;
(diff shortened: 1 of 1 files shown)


sethforprivacy:
Just confirming that the hard-coded DNS addresses here match what I'm seeing in packet captures when attempting to generate transactions with Cake Wallet (which uses wallet2). Thanks for the quick PR!

selsta:
+1 for removing the segregation heights code. Maybe it would make sense to keep the `--no-dns` option for now. It might be needed in the future and also functionality can be added to disable OpenAlias when `--no-dns` is specified. What to you think @tobtoht?

tobtoht:
@selsta 

Suggestion adopted in the interest of getting this patch merged before the v0.18 tag. We can discuss how wallets should handle OpenAlias in a separate issue.

selsta (approved):
Code reviewed and tested.

jeffro256 (approved):
Beautiful. The fewer DNS requests made, the better everyone's privacy is

jeffro256:
This feature was added in #3419 by @moneromooo-monero. @moneromooo-monero what are your thoughts about this PR?

selsta:
@jeffro256 There are no TXT records added to e.g. `segheights.moneropulse.org` so I don't think this was ever used. Back in 2018 there was this MoneroV fork so that's why this got added, see for example:

https://monero.stackexchange.com/questions/7826/how-can-individuals-safeguard-themselves-and-the-community-against-a-key-reusing

If this is ever needed in the future and also if this is the right approach we can trivially revert it and bring it back in a point release.

j-berman (approved):

PR #8412, opened 2022-06-29
Rename build-macos CI Job

agowa338:
Rename the "build-macos" job within "depends", as it is not building on macOS as the similarly named job within build.yml does. Also, both names overlap which is confusing when looking for step-by-step build instruction examples or when looking at the logfile.


selsta (approved):
Must have been a copy paste error, not sure how I never noticed it :D

jeffro256 (approved):

tobtoht (approved):

PR #8414, opened 2022-06-30
hardforks: set mainnet and stagenet v15/16 fork height

selsta:
I set the stagenet fork to be around 1 week before mainnet, I'm not sure if this is enough.


tobtoht:
Does not match blockheight in blog post: 2,688,888

selsta:
@tobtoht updated

tobtoht (approved):

j-berman (approved):

Issue #2516, opened 2017-09-23
Add monero 'man' (manual) pages


jonathancross:
Per #985 - We need to do a better job documenting how monero command line tools work and providing basic examples.  `man` pages are the most appropriate place for this.

Maxithi:
Fixed (partially) with #2832 ?

jonathancross:
It's a great start, but still feels like we need more detailed configuration info (there are many subtle configuration options and incompatible combinations / options).  Examples with explanation would be good too.

adrelanos:
I suggest to write man pages in markdown and then convert them to `.roff` using `ronn`, which is a markdown to roff converter. That's imo a lot easier to create and maintain than hand written `.roff` files.

* https://packages.debian.org/buster/ronn
* https://github.com/apjanke/ronn-ng

PR #8416, opened 2022-07-01
Utils: add manpages

MexicanTakeout:
Implemented per #2516


hyc:
fwiw, usually daemons (like monerod) go in chapter 8, not chapter 1.

hyc:
Out of curiosity; the files say they were automatically generated by pandoc. What is the original info that was fed into pandoc to generate these?

MexicanTakeout:
> Out of curiosity; the files say they were automatically generated by pandoc. What is the original info that was fed into pandoc to generate these?

https://github.com/MexicanTakeout/monero-man-pages-wip
The markdown docs are originally from https://github.com/monerodocs/md/tree/master/docs/interacting
I edited them to follow the man-pages convention.

jtgrassie:
Much as I like the intent, I'm not sure it's a wise idea to have links in official (which is what they'd become if merged to this repository), man pages link to unofficial, partially maintained, external documentation. At a glance, I noticed a few things in this offsite unofficial documentation that are not exactly correct or outdated.

Ideally we'd want to automate generation of man pages from the embedded documentation.

MexicanTakeout:
We could use getmonero.org as a source directly, it's just a matter of converting to man page format using pandoc, but then those won't follow manpage format convention like missing Description, Options, References sections.

The reason I use unofficial sources at monerodocs because they have more information and example usages compared to getmonero.org. Also I see some contributors over there are monero project people so it would be a safe bet to refernce.

PR #8423, opened 2022-07-06
Doc anon networks

inoutgain:
see COMMIT_EDITMSG

### DIFF
--- a/docs/ANONYMITY_NETWORKS.md
+++ b/docs/ANONYMITY_NETWORKS.md
@@ -1,149 +1,175 @@
 # Anonymity Networks with Monero
 
-Currently only Tor and I2P have been integrated into Monero. The usage of
-these networks is still considered experimental - there are a few pessimistic
-cases where privacy is leaked. The design is intended to maximize privacy of
-the source of a transaction by broadcasting it over an anonymity network, while
-relying on IPv4 for the remainder of messages to make surrounding node attacks
-(via sybil) more difficult.
+
+Currently Tor/I2P is available. Anonymity networks (AN) wrap traffic and proxy it over several hops to the final destination, which can include some exit back into clearnet or be a hidden service inside the AN itself. The support is experimental and while anonymity networks conceal node information there will be corner-cases that could leak information.
 
 
-## Behavior
+## Table of Contents
 
-If _any_ anonymity network is enabled, transactions being broadcast that lack
-a valid "context" (i.e. the transaction did not come from a p2p connection),
-will only be sent to peers on anonymity networks. If an anonymity network is
-enabled but no peers over an anonymity network are available, an error is
-logged and the transaction is kept for future broadcasting over an anonymity
-network. The transaction will not be broadcast unless an anonymity connection
-is made or until `monerod` is shutdown and restarted with only public
-connections enabled.
+  - [Outbound connections](##Outbound-connections)
+     - [Proxy node traffic](###Proxy-node-traffic)
+    - [Connect to hidden services](###Connect-to-hidden-service)
+    - [Peer handling](###Peer-handling)
+  - [Inbound connections](##Inbound-connections)
+    - [Allow inbound RPC](###Allow-inbound-RPC-(Server))
+    - [Connect to remote daemon](###Connect-to-remote-daemon-(Client))
+  - [Default configurations examples](##Default-configurations)
+  - [Privacy Limitations](##Privacy-Limitations)
+  - [Using Tor on tails (whonix)](##Using-Tor-on-tails-(whonix))
+  - [Legacy options](##Legacy-options)
+    - [Wrapping-monerod-with-torsocks](###-Wrapping-monerod-with-torsocks-(now:-use---proxy))
+  
+Three major options configure `monerod` and differ in connection, traffic type and route:
 
-Anonymity networks can also be used with `monero-wallet-cli` and
-`monero-wallet-rpc` - the wallets will connect to a daemon through a proxy. The
-daemon must provide a hidden service for the RPC itself, which is separate from
-the hidden service for P2P connections.
+| Connection | Type   | Clearnet | Exit-Relay/I2PTunnel | .onions/.i2p        |
+|------------|--------|----------|----------------------|---------------------|
+| incoming   | tx     | default  |           -          | --anonymous-inbound |
+|            | blocks | default  |           -          | N/A                 |
+| outgoing   | tx     | default  | --proxy              | --tx-proxy          |
+|            | blocks | default  | "                    | N/A                 | 
 
+Clearnet (Ip4) is the default. Connections are hopened both ways and peers sync blocks and tx (handshakes, peer timed syncs and transactions). Setting up an AN with `--proxy`, relays blocks and peer transactions (p2p) over relays. Introducing hidden services for transactions increases anonymity. You can provide a service yourself `--anonymous-inbound` or connect to the hidden service of a peer.
 
-## P2P Commands
+If any anonymity network is enabled, transactions being broadcast that lack a valid "context" (i.e. the transaction did not come from a p2p connection), will only be sent to peers on anonymity networks. 
 
-Only handshakes, peer timed syncs and transaction broadcast messages are
-supported over anonymity networks. If one `--add-exclusive-node` p2p address
-is specified, then no syncing will take place and only transaction broadcasting
-can occur. It is therefore recommended that `--add-exclusive-node` be combined
-with additional exclusive IPv4 address(es).
+Block synchronisation via hidden services (.onions, .i2p) is currently not supported, while handshakes, peer timed syncs and local transactions are, see [--tx-proxy](##Conceal-local-rpc-transactions) for outgoing tx or option [--anonymous-inbound](##inbound-connections) for incoming connections to your own service. 
 
-
-## Usage
-
-### Outbound Connections
-
-Connecting to an anonymous address requires the command line option
-`--tx-proxy` which tells `monerod` the ip/port of a socks proxy provided by a
-separate process. On most systems the configuration will look like:
-
+To connect your local node to the AN, you need a running instance for a socket to enter.
 ```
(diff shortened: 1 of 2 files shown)


nahuhh:
@inoutgain ?

docs/ANONYMITY_NETWORKS.md
```diff
@@ -1,149 +1,176 @@
...
 ```
+which tells `monerod` that `.onion` p2p addresses can be forwarded to a socks
+proxy at IP 127.0.0.1 port 9050 with a max of 16 outgoing connections and
+".b32.i2p" p2p addresses can be forwarded to a socks proxy at IP 127.0.0.1 port
+9000 with the default max outgoing connections. The local tx is now exclusively sent over onions and withholds until an onion peer is found.
 
-Either option can be listed multiple times, and can specify any mix of Tor,
-I2P, and IPv4 addresses. Using `--add-exclusive-node` will prevent the usage of
-seed nodes on ALL networks, which will typically be undesirable.
+Adding `disable_noise` will tell the peer to relay the tx over tor instead of onions. It disables Dandelion++ and makes network propagation faster, but degrades anonymity, because the tx will propagate outside the AN.
```
vtnerd, line 79: The information about `disable_noise` is incorrect. By default a dummy message is sent every 10-15 seconds (randomized) to make timing correlation attacks over tor/i2p more difficult. This uses more bandwidth and causes its own "noise" (signature) over the network - adversaries will know that you are running monero but it should be harder to trace transactions. The situation is:

 * If an outgoing I2P connection is available, this is always preferred for transactions
 * If an outgoing Tor connection is available with no I2P outgoing connections, then Tor will be used for transactions
 * The daemon will refuse to send transactions if any AN is enabled but has no outgoing connections
 * Dandelion++ is not used over the AN, but the remote service receiving the tx over an AN is expected to use it after a randomized delay (no way to detect or enforce this policy however)
 * `disable_noise` only affects whether dummy messages are sent, but does not change any of the above rules.

docs/ANONYMITY_NETWORKS.md
```diff
@@ -1,149 +1,176 @@
...
 
+Incoming connections require a hidden service. Regards Tor, add
 ```
+# file /etc/tor/torrc
 HiddenServiceDir /var/lib/tor/data/monero
-HiddenServicePort 18081 127.0.0.1:18081
+# HiddenServicePort 18081 127.0.0.1:18081 # rpc port
+HiddenServicePort 18083 127.0.0.1:18083   # tx port
 ```
+and restart `sudo systemctl restart tor.service`. This will generate your `.onion` address in `/var/lib/tor/data/monero/hostname`.
```
vtnerd, line 118: Mention that the onion address from this file must be listed in the `monerod` configuration to enable inbound connections. Otherwise `monerod` doesn't know the address to advertise to peers.

docs/ANONYMITY_NETWORKS.md
```diff
@@ -1,149 +1,176 @@
...
 ```
---proxy 127.0.0.1:9050
---daemon-address rveahdfho7wo4b2m.onion
+HiddenServicePort 18081 127.0.0.1:18081 # rpc port
 ```
+Now not only localhost rpcs, but remote rpcs incoming from AN via port 18081 can use your service. Use with caution.
 
-The proxy must match the address type - a Tor proxy will not work properly with
-i2p addresses, etc.
+If you wish to run a public rpc service, further options for `monerod` are availble that restrict available commands, add rpc-ssl or require authentication. You can also require e.g. [Client Authoriztion](https://community.torproject.org/onion-services/advanced/) on the AN side.
```
vtnerd, line 131: "authorization" is misspelled. 

I would word this section differently - if the user does not specify some client authorization, then the RPC service is basically public (the server/port is listed in the available hidden services list). So the wording should steer users to adding some authentication.

docs/ANONYMITY_NETWORKS.md
```diff
@@ -1,149 +1,176 @@
...
+You can use a client to do rpc calls against a remote `monerod` peer. The wallet will be configured to use a onion/i2p address:
+```
+--proxy 127.0.0.1:9050             # tor or i2p entry port
+--daemon-address rveahdfho.onion   # tor or
+--daemon-address cmeua5767.b32.i2p # i2p
+```
+The onion and i2p addresses provide the information necessary to authenticate and encrypt the connection end-to-end, which is inherent to the address design. A _remote_ daemon is untrusted by default, unless `--trusted-daemon` is set. A _local_ daemon is trusted unless option `--untrusted-daemon` is set.
 
-#### Tor & I2P
+If desired, SSL can also be applied to the connection with `--daemon-address https://rveahdfho.onion` which requires a server certificate that is signed by a "root" certificate on the machine running the wallet. Alternatively, `--daemon-cert-file` can be used to specify a certificate to authenticate the server.
```
vtnerd, line 144: Mention that there is no benefit if `tord`/`i2pd` and `monerod` are running on the same machine. The AN provides encryption to the AN daemon; SSL is only useful if `monerod` is on a different box than the AN service.

vtnerd (changes requested):
The overall comment I have is about changing the style from a ~80 character break to an unterminated one. Does anyone else care? The markdown formatter doesn't care, the only advantage is in text editors that don't wrap by default.

docs/ANONYMITY_NETWORKS.md
vtnerd: @selsta reminded me that one of the points is incorrect. A delay from i2p/tor->clearnet is only done when noise is enabled, there is no benefit to the delay when `disable_noise` is used.

vtnerd: Wrong again, thanks @selsta .This section still needs to be updated because the language isn't quite correct. I updated my bullet-points from the first comment to be more accurate.

The most glaring issue is `tx will propagate outside the AN` is not accurate. When any AN is enabled, a local transaction is either sent over any AN or not sent at all. The `disable_noise` policy disables:
 * Periodic sending of dummy messages
 * Randomized delay on the hidden service side
 * Dandelion++ sending on the hidden service side (immediate flood/fluff)

docs/ANONYMITY_NETWORKS.md
```diff
@@ -1,149 +1,175 @@
...
 
+| Connection | Type   | Clearnet | Exit-Relay/I2PTunnel | .onions/.i2p        |
+|------------|--------|----------|----------------------|---------------------|
+| incoming   | tx     | default  |           -          | --anonymous-inbound |
+|            | blocks | default  |           -          | N/A                 |
+| outgoing   | tx     | default  | --proxy              | --tx-proxy          |
+|            | blocks | default  | "                    | N/A                 | 
 
-## P2P Commands
+Clearnet (Ip4) is the default. Connections are hopened both ways and peers sync blocks and tx (handshakes, peer timed syncs and transactions). Setting up an AN with `--proxy`, relays blocks and peer transactions (p2p) over relays. Introducing hidden services for transactions increases anonymity. You can provide a service yourself `--anonymous-inbound` or connect to the hidden service of a peer.
```
ridauhebauye, line 32: My understanding was that Dandelion++ nodes only send transactions on outgoing connections. If --proxy is set (and clearnet is bound to localhost, which I think is what most people would intend) then the node will have no incoming transactions and so only non-Dandelion++ nodes would ever relay transactions to it. Stating that --proxy impacts transaction relaying is a bit misleading if this is the case.

ridauhebauye:
https://github.com/monero-project/monero/blob/dca23ebfd2ec8705960784b69430627e7435195b/docs/ANONYMITY_NETWORKS.md#i2ptor-stream-used-twice

This needs clarification. 

The Tor 10 minute circuit dropping statement isn't quite correct. Firstly, connections aren't dropped after 10 minutes, the circuit is just not reused for new connections after 10 minutes. Secondly, hidden service circuits reset the 10 minutes whenever data is sent out over the circuit (unlike clearnet circuits). 

What exactly is the metadata that is leaked to the remote hidden service? I believe the way it is currently implemented, if a transaction is received on an anonymous inbound then it is guaranteed to have originated from the person who it was received from. In other words, there is no mixing or relaying involved. Obviously the main leakage is that the hidden service operator can link transactions from the same entity. 

Has any work been done on periodically disconnecting peers to prevent circuit reuse? If not, is there a good reason why? 

Could an argument to --tx-proxy preventing circuit reuse be feasible? 

Is it currently, or would it be possible to allow users to specify which hidden service to use when broadcasting the transaction? Then somebody that needs to transact frequently can use their own hidden service. Unless I'm missing something, this would be the best method to facilitate frequent broadcasts that I'm aware of.

ridauhebauye:
https://github.com/monero-project/monero/blob/dca23ebfd2ec8705960784b69430627e7435195b/docs/ANONYMITY_NETWORKS.md#i2ptor-stream-used-twice

This needs clarification.

The Tor 10 minute circuit dropping statement isn't quite correct. Firstly, connections aren't dropped after 10 minutes, the circuit is just not reused for new connections after 10 minutes. Secondly, hidden service circuits reset the 10 minutes whenever data is sent out over the circuit (unlike clearnet circuits). Obviously this is not a relaible method of dropping connections. Has any work been done on periodically disconnecting peers to prevent circuit reuse? If not, is there a good reason why?

Assuming my comment above about Dandelion++ and --proxy is correct, and assuming that nothing is currently done to prevent circuit reuse then this is a bit of a problem. 

One solution could be to host your own monerod hidden service and add this as the exclusive hidden service to the monerod that you are using to broadcast transactions. The hidden service will then send the transactions out via Dandelion++ and nobody can determine which transactions came from the same source. This protects leakage to both ISP and remote hidden serivce operators.

It would be ideal if there was an option to add an exclusive node only for transaction broadcasting, so that blocks could then by synced normally. I think currently you would have to also add an exclusive IPV4 node in order to get blocks synced. 

If I've not misunderstood something, then I think it would be good to mention my above approach as a mitigation to the I2P/Tor Stream Used Twice issue.

PR #8424, opened 2022-07-06
wallet2: try to avoid sending back change to the same recipient

moneromooo-monero:


moneromooo-monero:
A future improvement could be to look up another level (ie, two txes deep) and assign a relatedness between 0 and 1, still fairly high.

src/wallet/wallet2.cpp
```diff
@@ -6221,15 +6221,51 @@ float wallet2::get_output_relatedness(const transfer_details &td0, const transfe
...
-size_t wallet2::pop_best_value_from(const transfer_container &transfers, std::vector<size_t> &unused_indices, const std::vector<size_t>& selected_transfers, bool smallest) const
+size_t wallet2::pop_best_value_from(const transfer_container &transfers, std::vector<size_t> &unused_indices, const std::vector<size_t>& selected_transfers, bool smallest, const std::vector<cryptonote::tx_destination_entry> &dsts) const
 {
   std::vector<size_t> candidates;
   float best_relatedness = 1.0f;
+
+  bool to_self = true;
+  for (const auto &dst: dsts)
+  {
+    if (m_subaddresses.find(dst.addr.m_spend_public_key) == m_subaddresses.end())
```
jeffro256, line 6232: Doesn't this only check self spends to subaddresses and doesn't cover the case where you send to your own primary addresses?

src/wallet/wallet2.cpp
```diff
@@ -6221,15 +6221,51 @@ float wallet2::get_output_relatedness(const transfer_details &td0, const transfe
...
+          {
+            if (std::find_if(dsts.begin(), dsts.end(), [&dst](const cryptonote::tx_destination_entry &e){ return e.addr== dst.addr; }) != dsts.end())
+            {
+              MDEBUG("Output is from from a tx sent earlier to one of the destinations, we won't use it unless really necessary");
+              relatedness = 1.0f;
+              break;
+            }
+          }
+        }
+      }
```
jeffro256, line 6265:
```suggestion
      // if this output is change from a tx that was sent to any address in dsts,
      // we crank relatedness to the max to avoid a "returning customer" inference
      bool returning_customer = false;
      for (const auto &ct: m_confirmed_txs)
      {
        const crypto::hash &txid = ct.first;
        if (txid != candidate.m_txid) continue;
  
        // we found a confirmed transfer corresponding to the output under consideration:
        // we check whether this transfer was to one of the addresses we're sending to
        for (const auto &dst: ct.second.m_dests)
        {
          if (std::find_if(dsts.begin(), dsts.end(), [&dst](const cryptonote::tx_destination_entry &e){ return e.addr== dst.addr; }) != dsts.end())
          {
            MDEBUG("Output is from from a tx sent earlier to one of the destinations, we won't use it unless really necessary");
            relatedness = 1.0f;
            returning_customer = true;
            break;
          }
        }
        if (returning_customer) break;
      }
```

If we add a flag called `returning_customer` and modify the control flow a little bit, we can exit early from the outer loop as well when we sense a returning customer.

moneromooo-monero: That looks like a large change, but it might be mostly whitespace. In any case, breaking on relatedness == 1.0f is bother simpler and more generic. Agreed ?

jeffro256: Yeah I like that better actually

PR #8427, opened 2022-07-10
zmq: publish submitted txs via zmq

j-berman:
### Problem

When a daemon is in the "fluff" epoch and a user submits a tx to the daemon, the tx doesn't get published via zmq's `txpool_add`. Reported by @TrasherDK; see logs [1](https://libera.monerologs.net/monero-dev/20220707#c117558-c117568), [2](https://libera.monerologs.net/monero-dev/20220707#c117619-c117634), and [3](https://libera.monerologs.net/monero-dev/20220708#c117725-c117728).

### Solution

Publish "just broadcasted" txs via zmq.

Of note, the test I added hangs waiting to receive txs that never come over zmq with the code from before this PR.

### Deeper context

- upon tx submission, [`handle_incoming_tx` is called with `relay_method:local`](https://github.com/monero-project/monero/blob/8f48f464957c875af3183cd9d35769592e1aef48/src/rpc/core_rpc_server.cpp#L1258) with no knowledge of the daemon's fluff/stem epoch status or tx propagation method.
- zmq doesn't publish the tx [since the tx has not broadcasted yet](https://github.com/monero-project/monero/blob/9f814edbd78c70c70b814ca934c1ddef58768262/src/cryptonote_core/cryptonote_core.cpp#L1107-L1108) (and therefore is not visible in the pool).
- the daemon then ["fluffs" the tx](https://github.com/monero-project/monero/blob/8f48f464957c875af3183cd9d35769592e1aef48/src/rpc/core_rpc_server.cpp#L1302) and updates its relay status in the db via [`on_transactions_relayed`](https://github.com/monero-project/monero/blob/9a124f681119855949f6406ecd69c2ad91da9770/src/cryptonote_protocol/levin_notify.cpp#L573) making the tx visible to anyone who queries the daemon.
- it's possible no other daemons reflect the tx back to the daemon, which means `handle_incoming_txs` doesn't get called a second time after the initial submission (which is particularly unique to the "fluff" flow).
- even if daemons *do* reflect the tx back to the daemon, since the tx has already updated its relay status to a broadcasted state, it won't publish to zmq since [this `already_have` will be true](https://github.com/monero-project/monero/blob/8f48f464957c875af3183cd9d35769592e1aef48/src/cryptonote_core/cryptonote_core.cpp#L1086).

With the above context in mind, I think `on_transactions_relayed` is the best place to pub txs in this circumstance because:

1. It's the point we know txs are just set to be broadcasted, and become visible/query-able in the pool.
2. It's not guaranteed `handle_incoming_txs` will get called again after the node broadcasts the tx.
3. Even if `handle_incoming_txs` does get called once the node broadcasts the tx, we don't know if the tx was "just broadcasted" or if we're handling a tx we already knew about that we already pub'd via zmq.
4. It seems quite challenging to pass something into `handle_incoming_tx` on the initial submission to make it aware of the relay mechanism that will subsequently be used when the tx is ultimately relayed. Although it seems like it would be safe for "fluff" txs to be immediately visible in the db on initial submission during the "fluff" epoch (and to push them over zmq immediately), the way the relay code is structured, it would take a larger overhaul to safely implement that approach to avoid doing so for "stem" txs.

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1406,21 +1406,66 @@ namespace cryptonote
     return true;
   }
   //-----------------------------------------------------------------------------------------------
+  bool core::notify_txpool_event(const epee::span<const cryptonote::blobdata> tx_blobs, epee::span<const crypto::hash> tx_hashes, epee::span<const cryptonote::transaction> txs, const std::vector<bool> &just_broadcasted) const
+  {
+    if (!m_zmq_pub)
+      return true;
+
+    if (tx_blobs.size() != tx_hashes.size() || tx_blobs.size() != txs.size() || tx_blobs.size() != just_broadcasted.size())
+      return false;
+
+    /* Publish txs via ZMQ that are "just broadcasted" by the daemon. This is
+       done here in addition to `handle_incoming_txs` in order to guarantee txs
+       are pub'd via ZMQ when we know the daemon has/will broadcast to other
+       nodes & *after* the tx is visible in the pool. This should get called
+       when the user submits a tx to a daemon in the "fluff" epoch relaying txs
+       via a public network. */
+    if (std::count(just_broadcasted.begin(), just_broadcasted.end(), true) == 0)
+      return true;
+
+    std::vector<txpool_event> results{};
+    results.resize(tx_blobs.size());
+    for (std::size_t i = 0; i < results.size(); ++i)
+    {
+      results[i].tx = std::move(txs[i]);
+      results[i].hash = std::move(tx_hashes[i]);
+      results[i].blob_size = tx_blobs[i].size();
+      results[i].weight = results[i].tx.pruned ? get_pruned_transaction_weight(results[i].tx) : get_transaction_weight(results[i].tx, results[i].blob_size);
+      results[i].res = just_broadcasted[i];
+    }
+
+    m_zmq_pub(std::move(results));
+
+    return true;
+  }
+  //-----------------------------------------------------------------------------------------------
   void core::on_transactions_relayed(const epee::span<const cryptonote::blobdata> tx_blobs, const relay_method tx_relay)
   {
+    // lock ensures duplicate txs aren't pub'd via zmq
+    CRITICAL_REGION_LOCAL(m_incoming_tx_lock);
+
     std::vector<crypto::hash> tx_hashes{};
     tx_hashes.resize(tx_blobs.size());
 
+    std::vector<cryptonote::transaction> txs{};
+    txs.resize(tx_blobs.size());
+
     for (std::size_t i = 0; i < tx_blobs.size(); ++i)
     {
-      cryptonote::transaction tx{};
-      if (!parse_and_validate_tx_from_blob(tx_blobs[i], tx, tx_hashes[i]))
+      if (!parse_and_validate_tx_from_blob(tx_blobs[i], txs[i], tx_hashes[i]))
       {
         LOG_ERROR("Failed to parse relayed transaction");
         return;
       }
     }
-    m_mempool.set_relayed(epee::to_span(tx_hashes), tx_relay);
+
+    std::vector<bool> just_broadcasted{};
+    just_broadcasted.reserve(tx_hashes.size());
+
+    m_mempool.set_relayed(epee::to_span(tx_hashes), tx_relay, just_broadcasted);
+
+    if (m_zmq_pub && matches_category(tx_relay, relay_category::legacy))
+      notify_txpool_event(tx_blobs, epee::to_span(tx_hashes), epee::to_span(txs), just_broadcasted);
   }
   //-----------------------------------------------------------------------------------------------
   bool core::get_block_template(block& b, const account_public_address& adr, difficulty_type& diffic, uint64_t& height, uint64_t& expected_reward, const blobdata& ex_nonce, uint64_t &seed_height, crypto::hash &seed_hash)
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -820,8 +820,10 @@ namespace cryptonote
     return true;
   }
   //---------------------------------------------------------------------------------
-  void tx_memory_pool::set_relayed(const epee::span<const crypto::hash> hashes, const relay_method method)
+  void tx_memory_pool::set_relayed(const epee::span<const crypto::hash> hashes, const relay_method method, std::vector<bool> &just_broadcasted)
   {
+    just_broadcasted.clear();
+
     crypto::random_poisson_seconds embargo_duration{dandelionpp_embargo_average};
     const auto now = std::chrono::system_clock::now();
     uint64_t next_relay = uint64_t{std::numeric_limits<time_t>::max()};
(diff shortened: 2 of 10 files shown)


src/cryptonote_core/tx_pool.cpp
```diff
@@ -820,7 +820,7 @@ namespace cryptonote
     return true;
   }
   //---------------------------------------------------------------------------------
-  void tx_memory_pool::set_relayed(const epee::span<const crypto::hash> hashes, const relay_method method)
+  void tx_memory_pool::set_relayed(const epee::span<const crypto::hash> hashes, const relay_method method, std::vector<bool> &just_broadcasted)
```
SChernykh, line 823: It doesn't check that `just_broadcasted` is empty, so it can `emplace_back` in the wrong places in the loop.
```
if (!just_broadcasted.empty())
    just_broadcasted.clear();
```

SChernykh (changes requested):

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1406,21 +1406,69 @@ namespace cryptonote
     return true;
   }
   //-----------------------------------------------------------------------------------------------
+  bool core::notify_txpool_event(const epee::span<const cryptonote::blobdata> tx_blobs, const epee::span<const crypto::hash> tx_hashes, const epee::span<const cryptonote::transaction> txs, const std::vector<bool> &just_broadcasted) const
+  {
```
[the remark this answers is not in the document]
j-berman: Good defense, done

src/cryptonote_core/tx_pool.cpp
j-berman: Done

SChernykh:
I ran it for 24 hours on my server, both p2pool nodes didn't complain about any duplicate transactions. I also tested it by sending a transaction through the same monerod - the tx was broadcasted to p2pool nodes via ZMQ 42.7 seconds later, as expected.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1406,21 +1406,75 @@ namespace cryptonote
...
+    /* Publish txs via ZMQ that are "just broadcasted" by the daemon. This is
+       done here in addition to `handle_incoming_txs` in order to guarantee txs
+       are pub'd via ZMQ when we know the daemon has/will broadcast to other
+       nodes & *after* the tx is visible in the pool. This should get called
+       when the user submits a tx to a daemon in the "fluff" epoch relaying txs
+       via a public network. */
+    bool valid_events = false;
+    for (std::size_t i = 0; i < results.size(); ++i)
+    {
+      results[i].tx = std::move(txs[i]);
```
vtnerd, line 1429: You shouldn't be able to actually move these objects, its a `const` view of the transactions. I think this just results in the copy-constructor being called (which is annoying as the compiler doesn't actively complain).

Same thing on next line.

vtnerd:
You commented on this - but to highlight to everyone -

The primary issue is that this leaks the origin (or re-broadcaster) to all connected ZMQ clients, but does not leak "normal" stem phase transactions. It looks like your counter argument was the difficulty of re-structuring the code to detect whether the local client was in the `fluff` epoch.

If this fix gets merged, I think an issue needs to be opened up about this because it is inconsistent (there's more privacy for non-origin txes).

src/cryptonote_core/tx_pool.cpp
```diff
@@ -820,8 +820,11 @@ namespace cryptonote
...
   {
+    if (!just_broadcasted.empty())
```
vtnerd, line 825: This `if` is unnecessary.

vtnerd (commented):
Other than the blurb above about starting an issue, this patch looks fine.

src/cryptonote_core/cryptonote_core.cpp
vtnerd: And indeed, [this just results in a copy](https://stackoverflow.com/questions/28595117/why-can-we-use-stdmove-on-a-const-object). I have such a love/hate for C++.

vtnerd (approved):
Terrible brain fart. Ignore the above about creating a github issue. I was doing all of this from memory, `on_transactions_relayed` is the exact function you want as its only called in the levin notify code.

Sometimes its difficult for me to accept I wrote a bug :)

vtnerd (approved):

src/cryptonote_core/tx_pool.cpp
SChernykh: What @vtnerd probably meant to say that `just_broadcasted.clear();` can just be called without the `if` with the same end result.

src/cryptonote_core/cryptonote_core.cpp
j-berman: Ya this was odd of me to leave as is. Will fix.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1406,21 +1406,75 @@ namespace cryptonote
...
   }
   //-----------------------------------------------------------------------------------------------
+  bool core::notify_txpool_event(const epee::span<const cryptonote::blobdata> tx_blobs, const epee::span<const crypto::hash> tx_hashes, const epee::span<const cryptonote::transaction> txs, const std::vector<bool> &just_broadcasted) const
+  {
+    if (!m_zmq_pub)
+      return true;
+
+    if (tx_blobs.size() != tx_hashes.size() || tx_blobs.size() != txs.size() || tx_blobs.size() != just_broadcasted.size())
+      return false;
+
```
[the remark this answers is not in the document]
j-berman: Perhaps there is still a misunderstanding, but if not, I think you meant a truthy 3rd param :) If the count of "just broadcasted" txs is 0, then we have nothing to publish. But if the count of txs that were **not** "just broadcasted" is 0, then there may still be some "just broadcasted" that we need to publish

I think the test should make it clear it should be truthy too; it hangs with a falsey 3rd param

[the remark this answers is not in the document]
SChernykh:
```
if (std::count(just_broadcasted.begin(), just_broadcasted.end(), true) == 0)
  return true;
```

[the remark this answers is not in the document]
j-berman: Right, I think you meant that^ @vtnerd 

See the latest

src/cryptonote_core/tx_pool.cpp
vtnerd: Correct, the STL code has to do this check basically anyway internally.

src/cryptonote_core/cryptonote_core.cpp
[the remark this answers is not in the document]
vtnerd: Yes, I was doing this from memory quickly and forgot the API. Mine didn't make any sense.

src/cryptonote_core/cryptonote_core.cpp
vtnerd: This never changed, but I don't think its a big deal (other than being slightly deceptive). It probably makes sense just to leave it, because if the `const`ness if the function ever changes this function call changes (although I think `transasction` does not have a proper `move` function).

vtnerd (approved):

selsta (approved):

Lissy93 (approved):

PR #8428, opened 2022-07-11
keccak: error out if passed mdlen 100

moneromooo-monero:
If we were to call it with 100, it would cause rsiz to be 0,
leading to an infinite loop.
This is really a pedantic patch, but since there's already a
range test, might as well make it better.

### DIFF
--- a/src/crypto/keccak.c
+++ b/src/crypto/keccak.c
@@ -123,7 +123,7 @@ void keccak(const uint8_t *in, size_t inlen, uint8_t *md, int mdlen)
     size_t i, rsiz, rsizw;
 
     static_assert(HASH_DATA_AREA <= sizeof(temp), "Bad keccak preconditions");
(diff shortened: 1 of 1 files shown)


jeffro256:
There's another check against `mdlen` a few lines below (line 162 I believe) that I think you should bring back up where this edit is: `(size_t)mdlen % sizeof(uint64_t)) != 0`. It also aborts when that condition is met, but for some reason it's only checked after `memcpy` and `memset` and a bunch of expensive operations

moneromooo-monero:
True, but "expensive" here is lightning fast compared to process termination. I'd rather keep the diff minimal, unless you find that moving that test forward fixes some possible bug.

SChernykh (approved):

metafraction (approved):

jeffro256:
> True, but "expensive" here is lightning fast compared to process termination. I'd rather keep the diff minimal, unless you find that moving that test forward fixes some possible bug.

Fair enough

jeffro256 (approved):

PR #8433, opened 2022-07-13
wallet2: force using output distribution for ringct outs

tobtoht:


j-berman (approved):

j-berman:
Sanity checked spending normal/coinbase pre-ringct + normal/coinbase ringct outputs used the expected distributions on a local testnet

PR #8441, opened 2022-07-17
[CRITICAL] Fixes x86-64 compatibility problem

Thor-x86:
I'm sorry for shouting this as critical, but please **don't use** `-Ofast` flag for sake of compatibility as it causes compiler to produce opcode which older CPU cannot comprehend. As a result, that older generation CPU (as example Intel Atom D510) throws [IOT illegal instruction](https://www.intel.com/content/www/us/en/docs/programmable/683620/current/illegal-instruction.html) error while running monero utilities. This makes entire monero utilities useless on all older x86-64 machines. For the worst, all release binaries of monero at package managers are compiled with`-Ofast` flag. Enthusiasts with potato server are suffering while trying to contribute their server as a full node.

As a reference, this is [an explanation from manual](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html):

> **-Ofast:** Disregard strict standards compliance. -Ofast enables all -O3 optimizations. It also enables optimizations that are not valid for all standard-compliant programs. It turns on -ffast-math, -fallow-store-data-races and the Fortran-specific -fstack-arrays, unless -fmax-stack-var-size is specified, and -fno-protect-parens. It turns off -fsemantic-interposition.

See? This type of optimization breaks well-known standards! If a superuser want to optimize stuffs, let him/her compile it himself/herself. Please please do not bother to let others down. Thank you for your support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -335,12 +335,6 @@ set(RISCV   1)
 set(RISCV32 1)
 endif()
 
-if(WIN32 OR ARM OR PPC64LE OR PPC64 OR PPC)
-  set(OPT_FLAGS_RELEASE "-O2")
-else()
-  set(OPT_FLAGS_RELEASE "-Ofast")
-endif()
-
 # BUILD_TAG is used to select the build type to check for a new version
 if(BUILD_TAG)
   message(STATUS "Building build tag ${BUILD_TAG}")
@@ -372,6 +366,7 @@ if(NOT MANUAL_SUBMODULES)
   endif()
 endif()
 
+set(OPT_FLAGS_RELEASE "-O2")
 set(CMAKE_C_FLAGS_RELEASE "-DNDEBUG ${OPT_FLAGS_RELEASE}")
 set(CMAKE_CXX_FLAGS_RELEASE "-DNDEBUG ${OPT_FLAGS_RELEASE}")
 


jeffro256:
On newer systems, `-Ofast` can be generally be significantly faster than `-O2`. I think instead of dropping `-Ofast` completely, we should detect the older CPU (using `/proc/cpuinfo` or otherwise) and then set optimization flags accordingly. Would you be able to provide the exact illegal x86 instruction generated? Also are you running some flavor of Linux on this processor?

iamamyth:
Does the problem persist with -O3? I ask because, based on the issue filed, `-fallow-store-data-races` could be relevant.

CMakeLists.txt
```diff
@@ -335,12 +335,6 @@ set(RISCV   1)
 set(RISCV32 1)
 endif()
 
-if(WIN32 OR ARM OR PPC64LE OR PPC64 OR PPC)
```
iamamyth, line 338: I think it would help simplify review and reduce merge confusion with other PRs to keep the OPT_FLAG_RELEASE setting in this location and remove the conditionals.

Thor-x86:
> Does the problem persist with -O3? I ask because, based on the issue filed, `-fallow-store-data-races` could be relevant.

**It works.** Unfortunately my home server (Intel Atom machine) is under heavy load trying to sync the blockchain for now. So I only tested `-O3` with `monero-blockchain-stats` command.

> we should detect the older CPU (using /proc/cpuinfo or otherwise)

How to do that on CMakeLists.txt?

> Would you be able to provide the exact illegal x86 instruction generated?

Which tool you preferred to get the illegal opcodes?

> Also are you running some flavor of Linux on this processor?

Just a simple Arch Linux, no desktop environment, no exotic modification. I use `linux-zen` kernel by the way, I'm not sure if zen optimization modifies some instructions behavior.

hyc:
Just compiling with `-march=native` is supposed to tailor the generated code to the current machine's CPU. There should not be any other action needed. If `-march=native -Ofast` generates an unusable binary for the machine it was compiled on, that is a compiler bug and should be reported to the compiler project. Not our problem.

CMakeLists.txt
Thor-x86: Sorry for that, I thought it would be better to regroup somewhere relevant. Amending the commit seems too dangerous for now, any suggestion what should I do?

Thor-x86:
> Would you be able to provide the exact illegal x86 instruction generated?

Here's what GDB found:

```
Starting program: /usr/bin/monero-blockchain-stats
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
2022-07-19 16:39:52.980 W Starting...
2022-07-19 16:39:52.980 W Initializing source blockchain (BlockchainDB)
/usr/include/c++/12.1.0/bits/unique_ptr.h:445: typename std::add_lvalue_reference<_Tp>::type std::unique_ptr<_Tp, _Dp>::operator*() const [with _Tp = cryptonote::Blockchain; _Dp = std::default_delete<cryptonote::Blockchain>; typename std::add_lvalue_reference<_Tp>::type = cryptonote::Blockchain&]: Assertion 'get() != pointer()' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff773836c in ?? () from /usr/lib/libc.so.6
```

It was debug build but the symbol still `??`. Seems like something off at libc. If this trully compiler's fault, I'm going to report this to GCC team. Also I'm going to re-compile this with Clang on my next spare time in case of GCC-specific bug.

CMakeLists.txt
iamamyth: I'd say just leave it for now until there's a consensus on the best resolution, as, setting aside jeffro256's comment, the proposed solutions appears to be 0-1 liners.

iamamyth:
>If -march=native -Ofast generates an unusable binary for the machine it was compiled on, that is a compiler bug and should be reported to the compiler project. Not our problem.

If I'm reading the linked issue correctly, it says the binary was compiled on a different architecture (AMD Ryzen 7-3700U) from the one on which it was executed (Intel Atom D510). So, the precondition doesn't hold, and the result is expected.

jeffro256:
> Also I'm going to re-compile this with Clang on my next spare time in case of GCC-specific bug.

You are compiling on said Intel Atom processor, correct?

garth-xmr:
Considering community member @hyc response perhaps you can edit the title to remove the critical tag

moneromooo-monero:
SIGABRT is not an illegal insn, that'd be SIGILL.
The output you showed looks more like a bug in the application code. SIGABRT points to an assertion failure.

moneromooo-monero:
One thing you can do is (since it breaks right at startup) run it with --log-level 4, which will allow pinpoint a bit better where it dies. Somehow I'm thinking the NULL dereference trick we use, which the compiler might, er, trip on.

selsta:
Obsolete, we don't use Ofast anymore and the reported assertion was fixed in https://github.com/monero-project/monero/pull/8924

PR #8444, opened 2022-07-18
gpg_keys: add jeffro256 key

jeffro256:


selsta (approved):

jeffro256:
Same GPG here: https://github.com/monero-project/gitian.sigs/blob/master/gitian-pubkeys/jeffro256.asc

PR #8459, opened 2022-07-24
Update build.yml

akaramoozian:
better indentation at line 12


selsta:
Thank you for your contribution but unfortunately we don't accept pure whitespace changes like this.

PR #8465, opened 2022-07-27
ledger support for hf 15 (BP+, view tags) [master]

j-berman:
Pairs with: https://github.com/LedgerHQ/app-monero/pull/110

I tested sending/receiving mainnet txs + tested sending/receiving testnet txs (post-fork), both exporting view key and not.

### DIFF
--- a/src/device/device_ledger.cpp
+++ b/src/device/device_ledger.cpp
@@ -266,6 +266,7 @@ namespace hw {
     #define INS_DERIVE_PUBLIC_KEY               0x36
     #define INS_DERIVE_SECRET_KEY               0x38
     #define INS_GEN_KEY_IMAGE                   0x3A
+    #define INS_DERIVE_VIEW_TAG                 0x3B
     #define INS_SECRET_KEY_ADD                  0x3C
     #define INS_SECRET_KEY_SUB                  0x3E
     #define INS_GENERATE_KEYPAIR                0x40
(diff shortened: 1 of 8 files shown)


selsta:
```
In file included from /Users/selsta/dev/monero/tests/unit_tests/device.cpp:32:
/Users/selsta/dev/monero/src/device/device_default.hpp:104:19: warning: 'derive_view_tag' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
            bool  derive_view_tag(const crypto::key_derivation &derivation, const std::size_t output_index, crypto::view_tag &view_tag);
                  ^
/Users/selsta/dev/monero/src/device/device.hpp:180:23: note: overridden virtual function is here
        virtual bool  derive_view_tag(const crypto::key_derivation &derivation, const std::size_t output_index, crypto::view_tag &view_tag) = 0;
                      ^
1 warning generated.
```

src/device/device_ledger.cpp
```diff
@@ -1663,6 +1719,13 @@ namespace hw {
         recv_len -= 32;
       }
 
+      if (use_view_tags)
+      {
+        ASSERT_X(recv_len>=1, "Not enough data from device");
+        memmove(&view_tag.data, &this->buffer_recv[offset], 1);
+        recv_len -= 1;
```
moneromooo-monero, line 1726: I'd add ++ after offset here, in case stuff gets added later after this and goes out of sync

selsta (approved):

PR #8467, opened 2022-07-27
txpool: continue pool pruning even if a tx can't be found [release-v0.18]

j-berman:
https://github.com/monero-project/monero/pull/8415


selsta:
This PR was approved here: https://github.com/monero-project/monero/pull/8415

PR #8471, opened 2022-07-29
simplewallet: fix line-wrap spacing

hinto-janai:
Before
```
This is the command line monero wallet. It needs to connect to a monero
daemon to work correctly.
WARNING: Do not reuse your Monero keys on another fork, UNLESS this fork has key reuse mitigations built in. Doing so will harm your privacy.
```
```
NOTE: the following 25 words can be used to recover access to your wallet. Write them down and store them somewhere safe and secure. Please do not store them in your email or on file storage services outside of your immediate control.
```
After
```
This is the command line Monero wallet. It needs to connect to a Monero daemon to work correctly.
WARNING: Do not reuse your Monero keys on another fork, UNLESS this fork has key reuse mitigations built in. Doing so will harm your privacy.
```
```
NOTE: the following 25 words can be used to recover access to your wallet.
Write them down and store them somewhere safe and secure. Please do not store
them in your email or on file storage services outside of your immediate control.
```

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -4048,9 +4048,9 @@ bool simple_wallet::ask_wallet_create_if_needed()
  */
 void simple_wallet::print_seed(const epee::wipeable_string &seed)
 {
-  success_msg_writer(true) << "\n" << boost::format(tr("NOTE: the following %s can be used to recover access to your wallet. "
-    "Write them down and store them somewhere safe and secure. Please do not store them in "
-    "your email or on file storage services outside of your immediate control.\n")) % (m_wallet->multisig() ? tr("string") : tr("25 words"));
+  success_msg_writer(true) << "\n" << boost::format(tr("NOTE: the following %s can be used to recover access to your wallet.\n"
+    "Write them down and store them somewhere safe and secure. Please do not store\n"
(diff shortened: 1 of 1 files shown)


src/simplewallet/simplewallet.cpp
```diff
@@ -4048,9 +4048,9 @@ bool simple_wallet::ask_wallet_create_if_needed()
  */
 void simple_wallet::print_seed(const epee::wipeable_string &seed)
 {
-  success_msg_writer(true) << "\n" << boost::format(tr("NOTE: the following %s can be used to recover access to your wallet. "
-    "Write them down and store them somewhere safe and secure. Please do not store them in "
-    "your email or on file storage services outside of your immediate control.\n")) % (m_wallet->multisig() ? tr("string") : tr("25 words"));
+  success_msg_writer(true) << "\n" << boost::format(tr("NOTE: the following %s can be used to recover access to your wallet.\n"
+    "Write them down and store them somewhere safe and secure. Please do not store\n"
```
Lissy93, line 4052: Why does this need a line break in the middle of the sentence, should wrap automatically depending on size?

Just thinking, I've seen a bunch of other instances where line length is long, but without breaks, and maybe there should be a consistent standard throughout. Example:

https://github.com/monero-project/monero/blob/master/src/simplewallet/simplewallet.cpp#L5819-L5822

hinto-janaiyo: It does not wrap automatically.

wrap

PR #8299, opened 2022-04-27
feat(trezor): add HF15 support, BP+

ph4r05:
- HP15 changes implemented to the Trezor client in Monero wallet
- BP+ support added for Trezor
- old Trezor firmware version support removed, code cleanup

Trezor firmware changes: https://github.com/trezor/trezor-firmware/pull/2232

PR #8483, opened 2022-08-05
feat(trezor): add HF15 support, BP+, release-v0.18

ph4r05:
- BP+ support added for Trezor
- old Trezor firmware version support removed, code cleanup

Same as https://github.com/monero-project/monero/pull/8299 due to https://github.com/monero-project/monero/pull/8299#issuecomment-1204393941


selsta (approved):
Tested.

PR #8488, opened 2022-08-08
tests: update block_weight for 2021 scaling

moneromooo-monero:
the test was still performed with consensus rules from before
that change


Boog900 (approved):
LGTM

PR #8489, opened 2022-08-08
simplewallet: error added upon incorrect index placing

### LINKED ISSUE #8480: sweep_* command: Please add an "error" when args (index and address) are incorrectly placed
Hi,

This is my first post on Monero. I am reporting a strange behavior that I recently found, which I likely find a bug. Let me detail as below:

Assume a wallet with 6 unspent outputs (under account Index: 0, 5 sub-addresses and 1 primary address, with balances as below:
 
```
1. 0.000000001000
2. 0.000000001000
3. 0.000000001000
4. 0.000000001000
5. 0.000000001000
6. 0.100000000000
```
Notice, 5 of the above addresses have negligible amount. Negligible, with respect to "network fee" i.e. (~0.000005 XMR).

As funds sit on transactions' unspent outputs and in principle, a single transaction should be able to aggregate and spend outputs from multiple addresses (and by extension from multiple accounts). However, in the above scenario when the balance at a "sub-address" is less than the "network fee", that output is not being spent.
 
I also validated this by splitting an output into multiple small chunks and sweeping it off my wallet. I realized later this was the exact problem as posted below, the reason the wallet is not being emptied (off dust) in spite of adding sufficient amount to the wallet to clear it up. A balance of 000000005000 XMR is not possible to be spent.
 
https://www.reddit.com/r/Monero/comments/wcaykc/sadly_my_monero_practically_lost_forever/
 
This has three potential problems going forward:
1) People donating very small amounts to an address, would forever remain unspent at that address
2) Reduces Monero supply with growing adoption
3) This can become an issue when the network fee increases in future, thus blocking out greater amounts

This is just like saying cent, nickel and dime have no value today and cannot be spent. This should not be the case for digital money.
 
Looking forward to any views.

Cheers,
aFungible

Kyrilasa:
Closes  #8480

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -7131,22 +7131,36 @@ bool simple_wallet::sweep_main(uint32_t account, uint64_t below, bool locked, co
     return true;
 
   std::vector<std::string> local_args = args_;
-
+  auto index_iterator = std::find_if(local_args.begin(), local_args.end(), [](std::string &arg){return arg.find("index=") != std::string::npos;});
   std::set<uint32_t> subaddr_indices;
-  if (local_args.size() > 0 && local_args[0].substr(0, 6) == "index=")
+
+  if (index_iterator != local_args.end())
   {
-    if (local_args[0] == "index=all")
+    if (index_iterator != local_args.begin())
     {
-      for (uint32_t i = 0; i < m_wallet->get_num_subaddresses(m_current_subaddress_account); ++i)
-        subaddr_indices.insert(i);
-    }
-    else if (!parse_subaddress_indices(local_args[0], subaddr_indices))
-    {
+      fail_msg_writer() << tr("indexes should be the first arguments if given, transaction canceled");
       print_usage();
-      return true;
+      return false;
     }
-    local_args.erase(local_args.begin());
+    else
+    {
+      if (local_args[0].substr(0, 6) == "index=")
+      {
+        if (local_args[0] == "index=all")
+        {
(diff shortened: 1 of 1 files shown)


Kyrilasa:
> It should be discussed whether this should be an error or whether the command line argument logic should be modified to understand the command no matter where the `index=` argument is placed. I think `index=` is self describing and obviously can't be understood to be an address so the logic should just be modified to work.

Yeah @jeffro256, I was thinking about the same, to rework the argument parsing logic, but @moneromooo-monero told me on IRC that it would be the best to maintain the logic and not force my "way of thinking" to it. I will think about it, maybe I can come up with an idea.

moneromooo-monero:
Mind posting the line ? I do not remember saying that.

That said, I think it's better to error out, unless other uses of index= in other commands also change the same way, but I don't particularly care too much either way.

Kyrilasa:
> Mind posting the line ? I do not remember saying that.
> 
> That said, I think it's better to error out, unless other uses of index= in other commands also change the same way, but I don't particularly care too much either way.

Do not get me wrong, I did not intend to be hostile in any way, you just wrote me on IRC previously when I asked how I can contribute, that it is better not to change the logic right away, better to consult about it beforehand. Mea culpa I phrased it wrong.

moneromooo-monero:
I usually tell people not to rewrite code to a style they like better. That should not be construed to not improving stuff. Though of course "improvement" can be subjective...
I'm OK with either solution in this case.

Kyrilasa:
Do not get me wrong, I did not intend to be hostile in any way, you just
wrote me on IRC previously, that it is better not to change the logic right
away, better to consult about it beforehand. Mea culpa I phrased it wrong.

PR #8490, opened 2022-08-09
version: bump master version number

selsta:
Otherwise Ledger rejects the wallet. Alternatively I could set it to something like `0.18.99.99`, but I don't know if that's better.


j-berman (approved):

tobtoht (approved):

jeffro256 (approved):
Can confirm ledger version issue

PR #8495, opened 2022-08-10
repo: remove ldns leftovers

selsta:
`ldns` was removed in https://github.com/monero-project/monero/pull/160, from what I can see there is no reason to keep it around. This also means one less library to build for depends.


TheCharlatan (approved):

jeffro256 (approved):
Wow that's been in there doing nothing for quite some time

tobtoht (approved):

PR #8501, opened 2022-08-14
lmdb: update to latest master

selsta:
Needs extensive sync testing on all OS. Should include performance improvements on Windows: https://github.com/LMDB/lmdb/commit/da0527ac75b811419b7007202799f96b2edb5aef


hyc:
Note that currently our tree is based on mdb.master but also has the RAWPART patches from mdb.master3 and this PR is missing those.

selsta:
Continue in #8502

PR #8503, opened 2022-08-15
blockchain_prune: faster

moneromooo-monero:
on my anecdotal SSD, goes from about 9 hours to 1h20.

### DIFF
--- a/src/blockchain_utilities/blockchain_prune.cpp
+++ b/src/blockchain_utilities/blockchain_prune.cpp
@@ -52,8 +52,10 @@ using namespace cryptonote;
 static std::string db_path;
 
 // default to fast:1
-static uint64_t records_per_sync = 128;
+static uint64_t records_per_sync = 16 * 65536;
 static const size_t slack = 512 * 1024 * 1024;
+
+static std::vector<bool> is_v1;
 
 static std::error_code replace_file(const boost::filesystem::path& replacement_name, const boost::filesystem::path& replaced_name)
 {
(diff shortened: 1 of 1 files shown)


src/blockchain_utilities/blockchain_prune.cpp
```diff
@@ -601,26 +646,27 @@ int main(int argc, char* argv[])
...
   open(env0, paths[0], db_flags, true);
   open(env1, paths[1], db_flags, false);
-  copy_table(env0, env1, "blocks", MDB_INTEGERKEY, MDB_APPEND);
-  copy_table(env0, env1, "block_info", MDB_INTEGERKEY | MDB_DUPSORT| MDB_DUPFIXED, MDB_APPENDDUP, BlockchainLMDB::compare_uint64);
+  copy_table(env0, env1, "blocks", MDB_INTEGERKEY, 0);
+  copy_table(env0, env1, "block_info", MDB_INTEGERKEY | MDB_DUPSORT| MDB_DUPFIXED, 0, BlockchainLMDB::compare_uint64);
   copy_table(env0, env1, "block_heights", MDB_INTEGERKEY | MDB_DUPSORT| MDB_DUPFIXED, 0, BlockchainLMDB::compare_hash32);
   //copy_table(env0, env1, "txs", MDB_INTEGERKEY);
-  copy_table(env0, env1, "txs_pruned", MDB_INTEGERKEY, MDB_APPEND);
-  copy_table(env0, env1, "txs_prunable_hash", MDB_INTEGERKEY | MDB_DUPSORT | MDB_DUPFIXED, MDB_APPEND);
```
hyc, line 609: If you're copying something in DB order, in this case you could use both MDB_APPEND and MDB_APPENDDUP. Their effects are independent, MDB_APPEND applies to the primary key, MDB_APPENDDUP applies only to the dupsort values.

moneromooo-monero: OK. The doc implies otherwise: 
> MDB_APPENDDUP - as above, but for sorted dup data
might need rewording then.

hyc (approved):

src/blockchain_utilities/blockchain_prune.cpp
jtgrassie: The implementation seems to handle the case smartly enough though. I.e. if you mdb_cursor_put using MDB_APPENDDUP (on an MDB_DUPSORT db of course), and the key _doesn't_ already exist, it appends a new key and data. If there's an existing key, it appends the new data.

jtgrassie (approved):

plowsof:
testing on stagenet (~10GB unpruned)
time to prune without this PR:
real	43m18.272s
real	47m50.559s

with:
real	4m49.870s
real	4m49.730s

PR #8507, opened 2022-08-15
wallet2: chunk calls to is_key_image_spent in import_key_images

moneromooo-monero:
this will help avoid timeouts when importing large amounts of those


src/wallet/wallet2.cpp
```diff
@@ -12739,7 +12739,8 @@ uint64_t wallet2::import_key_images(const std::vector<std::pair<crypto::key_imag
     return 0;
   }
 
-  req.key_images.reserve(signed_key_images.size());
+  static const size_t chunk_size = 250;
+  req.key_images.reserve(chunk_size);
```
hyc, line 12743: should this be min(250, signed_key_images.size()) ? or does it not make much difference?

moneromooo-monero: Really doesn't make mych difference, but why not.

PR #3668, opened 2018-04-19
[NEXT POINT RELEASE] Build: use miniupnp submodule

anonimal:
Resolves https://hackerone.com/reports/340012 (public disclosure very soon).

@fluffypony @danrmiller We need to remove miniupnp from the **release** build system completely because:
1. We shouldn't trust CMake to be reliable (it's CMake)
2. Patches are needed to ensure system headers aren't used and, if they aren't reliable then see point 1.

To remove, on Ubuntu for example, you would need to remove both libminiupnpc-dev and miniupnpc packages.

Once @fluffypony sets up a monero-project miniupnp (like we did with unbound), I can push [branch monero](https://github.com/anonimal/miniupnp/tree/monero) from my fork which includes all the necessary patches and then update this PR.

Important note: I would even advise to completely remove system miniupnp from all buildbot machines after we adjust our CMake to build our submodule/in-tree exclusively. There's no legitimate reason (that I can think of) to continue to depend on a system's miniupnp.

PR #8509, opened 2022-08-16
cmake: only fallback to vendored miniupnpc

selsta:
The current code seems broken. It runs `find_package(Miniupnpc REQUIRED)` but the script can't even find anything due to incorrect capitalization.

```
  The package name passed to `find_package_handle_standard_args` (MiniUPnPc)
  does not match the name of the calling package (Miniupnpc).  This can lead
  to problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  cmake/FindMiniupnpc.cmake:39 (find_package_handle_standard_args)
  external/CMakeLists.txt:38 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

We added the vendored miniupnpc initially because of an vulnerability but that was 4 years ago: https://github.com/monero-project/monero/pull/3668

Fix it by first searching for system miniupnpc and then only fallback to the vendored version if it isn't found. Needs some testing to make sure it doesn't break static builds.

### DIFF
--- a/cmake/FindMiniupnpc.cmake
+++ b/cmake/FindMiniupnpc.cmake
@@ -37,7 +37,7 @@ set(MINIUPNP_STATIC_LIBRARIES ${MINIUPNP_STATIC_LIBRARY})
 
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(
-  MiniUPnPc DEFAULT_MSG
+  Miniupnpc DEFAULT_MSG
   MINIUPNP_INCLUDE_DIR
   MINIUPNP_LIBRARY
 )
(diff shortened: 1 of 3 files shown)


selsta:
Seems arch install miniupnpc in `miniupnpc/miniupnpc.h`, while Ubuntu, macOS and Windows use `miniupnp/miniupnpc/miniupnpc.h`

Don't know yet how to solve this.

hyc:
Should just use `#include <miniupnpc/miniupnpc.h>` and then add a "-I/usr/include/miniupnp" to CPPFLAGS if necessary

selsta:
Closing as this will be solved by different PRs.

Issue #8484, opened 2022-08-05
Segfault and corrupt wallet files caused by exporting and importing outputs with monero-wallet-rpc


woodser:
I'm able to segfault monero-wallet-rpc and corrupt the wallet files by calling export_outputs then import_outputs with the result. The calls succeed, indicating 0 outputs were imported. But then these errors start appearing when the next transaction is sent. My tests send funds back to the original wallet so it might be something specific to that. Otherwise, my build and tests are working on v0.18.0.0.

```
2022-08-04 23:38:41.438	W Spent money: 0.000000000000, with tx: <48219a8a2c1534c13b2d79448e79598192829327686f45fdd2fcc04649976c64>
2022-08-04 23:38:41.438	E Invalid index
2022-08-04 23:38:41.438	E Error parsing blocks: Invalid index
2022-08-04 23:38:46.206	E Public key 2788e1c6b262a767ee5bcaeea8da6a560c8e68c8d5f8c46aaf97f8b54e9fd5a2 from received 0.000000600000 output already exists with unspent 0.000000600000 in tx <48219a8a2c1534c13b2d79448e79598192829327686f45fdd2fcc04649976c64>, received output ignored
2022-08-04 23:38:46.208	E Public key c185f909de3c0ef5db4255ed423060c00ccc09ad379f08bc669dfdd820ee0fdd from received 0.000000600000 output already exists with unspent 0.000000600000 in tx <48219a8a2c1534c13b2d79448e79598192829327686f45fdd2fcc04649976c64>, received output ignored
```

I confirmed the issue is introduced with #8179 and resolved by reverting this PR.

moneromooo-monero:
I could make two txes in a row with cold signing, exporting/importing outputs/key images both times. Can you describe your steps in more detail ?

In any case, reading the code, I see something is missing, below. My test wasn't run with that patch. Untested yet.

```
diff --git a/src/wallet/wallet2.cpp b/src/wallet/wallet2.cpp
index 195763949..d87aa0e33 100644
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -13241,7 +13241,7 @@ size_t wallet2::import_outputs(const std::pair<uint64_t, std::vector<tools::wall
     exported_transfer_details etd = outputs.second[i];
     transfer_details &td = m_transfers[i + offset];
 
-    // setup td with "cheao" loaded data
+    // setup td with "cheap" loaded data
     td.m_block_height = 0;
[...]

Issue #8476, opened 2022-08-01
Offline sign transactions requires "Try using export_outputs all" periodically.


nape-max:
Periodically, on call 'sign_transfer', I get this message:
"code": -42,
"message": "Failed to sign unsigned tx: Imported outputs omit more outputs that we know of. Try using export_outputs all."

Context: Wallet, have 2 addresses: Primary and one subaddress. Send coins to subaddress and then try to send coins from account. All transfers correct and completed. But after 1-3 transfers, I received this message.

Offline wallet display only 'primary' address in response from call 'get_address'. I tried to call 'create_address' on view-only and offline wallet both. Got same result and same message after 1-3 transfers.

Cannot understand reasons of that.
When I'm using 'export_outputs' and 'import_outputs', I got 'num_imported': 0. But after 1-3 transactions, receive message specified above. But after 'export_outputs' with 'all': true and 'import_outputs' I got 'num_imported' with non zero value.

I think is is not correct. Can you exlplain it?

moneromooo-monero:
Most likely ae0a840fdaa15054a7c0529869fb20df0e1605ea being broken. It can be reverted safely.

mrtestyboy781:
bumping this issue, seeing it as well

PR #8510, opened 2022-08-16
wallet2: export_outputs + import_outputs + cold signing patches [release]

j-berman:
PR of the code discussed in #8484 (the base patches in this PR are written by @moneromooo-monero). This PR should also patch #8476:

- Only offline wallets can import outputs.
- Empty exports won't corrupt wallet files when signing.
- Subaddress indexes are exported & imported.

I added explicit tests for the buggy failure cases described in #8476. But I wasn't able to reproduce @woodser's issue described in [this comment](https://github.com/monero-project/monero/issues/8484#issuecomment-1213632430):

> it occasionally hangs on import_outputs (when used correctly with view-only and offline wallets) for some unknown reason

Requesting @woodser check if he can repro the hang with this code; I suspect the hang might have something to do with the slightly off empty exports, which should be good with this PR.

### DIFF
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -406,6 +406,8 @@ private:
 
       BEGIN_SERIALIZE_OBJECT()
         VERSION_FIELD(1)
+        if (version < 1)
+          return false;
         FIELD(m_pubkey)
         VARINT_FIELD(m_internal_output_index)
         VARINT_FIELD(m_global_output_index)
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -401,9 +401,11 @@ private:
       } m_flags;
       uint64_t m_amount;
       std::vector<crypto::public_key> m_additional_tx_keys;
+      uint32_t m_subaddr_index_major;
+      uint32_t m_subaddr_index_minor;
 
       BEGIN_SERIALIZE_OBJECT()
-        VERSION_FIELD(0)
+        VERSION_FIELD(1)
         FIELD(m_pubkey)
         VARINT_FIELD(m_internal_output_index)
         VARINT_FIELD(m_global_output_index)
@@ -411,6 +413,8 @@ private:
         FIELD(m_flags.flags)
         VARINT_FIELD(m_amount)
         FIELD(m_additional_tx_keys)
+        VARINT_FIELD(m_subaddr_index_major)
+        VARINT_FIELD(m_subaddr_index_minor)
       END_SERIALIZE()
     };
 
--- a/tests/functional_tests/functional_tests_rpc.py
+++ b/tests/functional_tests/functional_tests_rpc.py
@@ -40,8 +40,9 @@ except:
 N_MONERODS = 4
 
 # 4 wallets connected to the main offline monerod
-# a wallet connected to the first local online monerod
-N_WALLETS = 5
+# 1 wallet connected to the first local online monerod
+# 1 offline wallet
+N_WALLETS = 6
 
 WALLET_DIRECTORY = builddir + "/functional-tests-directory"
 FUNCTIONAL_TESTS_DIRECTORY = builddir + "/tests/functional_tests"
@@ -61,6 +62,7 @@ wallet_extra = [
   ["--daemon-port", "18180"],
   ["--daemon-port", "18180"],
   ["--daemon-port", "18182"],
+  ["--offline"],
 ]
 
 command_lines = []
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -13162,6 +13162,8 @@ size_t wallet2::import_outputs(const std::pair<uint64_t, std::vector<tools::wall
 {
   PERF_TIMER(import_outputs);
 
+  THROW_WALLET_EXCEPTION_IF(watch_only(), error::wallet_internal_error, "Hot wallets cannot import outputs");
+
   THROW_WALLET_EXCEPTION_IF(outputs.first > m_transfers.size(), error::wallet_internal_error,
       "Imported outputs omit more outputs that we know of");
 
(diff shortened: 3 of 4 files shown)


src/wallet/wallet2.h
```diff
@@ -401,16 +401,20 @@ namespace tools
...
-        VERSION_FIELD(0)
+        VERSION_FIELD(1)
         FIELD(m_pubkey)
         VARINT_FIELD(m_internal_output_index)
         VARINT_FIELD(m_global_output_index)
         FIELD(m_tx_pubkey)
         FIELD(m_flags.flags)
         VARINT_FIELD(m_amount)
         FIELD(m_additional_tx_keys)
+        VARINT_FIELD(m_subaddr_index_major)
```
moneromooo-monero, line 416: In theory I should have added a if (version < 1) return false; (since you can't really "default" to a sane value).

tests/functional_tests/functional_tests_rpc.py
```diff
@@ -82,6 +82,12 @@
   outputs.append(open(FUNCTIONAL_TESTS_DIRECTORY + '/wallet' + str(i) + '.log', 'a+'))
   ports.append(18090+i)
 
+# 1 offline wallet
+command_lines.append([str(18090+N_WALLETS) if x == "wallet_port" else x for x in wallet_base])
+command_lines[-1] += ["--offline"]
+outputs.append(open(FUNCTIONAL_TESTS_DIRECTORY + '/wallet' + str(N_WALLETS) + '.log', 'a+'))
+ports.append(18090+N_WALLETS)
+
```
moneromooo-monero, line 90: Any chance this cam be moved back to the loop since it's very similar ? It's just a if i == N_WALLETS-1:  command_lines[i] += "--offline"; AFAICT.

moneromooo-monero (commented):
Thanks for doing that.

moneromooo-monero (approved):

woodser:
With this PR I'm getting `Hot wallets cannot import outputs`, even though the wallet is offline and initialized with an empty "" server. I guess `m_offline` is false when it should be true without the server?

Side note, the error message is `Failed to import outputsHot wallets cannot import outputs` because there's no `: ` on [this line](https://github.com/monero-project/monero/blob/b6a029f222abada36c7bc6c65899a4ac969d7dee/src/wallet/wallet2.cpp#L13398).

j-berman:
Thinking more on this, it seems like I'm missing more ways that a wallet can technically be offline even if `--offline` isn't passed at startup or if `set_offline` isn't explicitly called (e.g. wi-fi turned off). An empty "" server seems like one (imo somewhat ambiguous) way of multiple to start a cold wallet.

I'm thinking the best option on the table is to require cold wallet implementations to explicitly set the offline flag (either when starting the wallet, or by calling `set_offline`/`setOffline` via the wallet API). And I'm thinking the error message could be more detailed for implementers:

```
Hot wallets cannot import outputs (you must pass the --offline flag when starting the wallet or explicitly set the wallet to the offline mode)
```

Maybe @moneromooo-monero has a different opinion on this.

(I also cleaned up the spacing; good spot, thank you)

j-berman:
Actually, perhaps there should even be a new required flag `--cold-wallet` which permanently ensures `m_offline` is false after the wallet is created. Otherwise it would be possible to trigger [this error](https://github.com/moneromooo-monero/bitmonero/commit/79e09fbf8c5f04a7f308762a5bc7a0927a4cd909#r80699311) (and other unexpected behavior) like this:

1. Create a full wallet
2. Set `m_offline` to true
3. Import outputs
4. Set `m_offline` to false
5. Transfer

moneromooo-monero:
Do you actually want to import outputs into a hot wallet ? If so, why ? If not, can you rephrase ?

woodser:
> Do you actually want to import outputs into a hot wallet ? If so, why ? If not, can you rephrase ?

I don't have any use case for that, but the error I'm seeing `Hot wallets cannot import outputs` isn't correct since the wallet is offline, not connected to any server.

moneromooo-monero:
OK, then I guess it can be made... "Only cold wallets can import outputs". Since it's what tests for IIRC.

woodser:
> OK, then I guess it can be made... "Only cold wallets can import outputs". Since it's what tests for IIRC.

So with this change, a user would need to manually call `set_offline` in addition to passing an empty server, to denote the wallet as permanently cold?

moneromooo-monero:
Looking at the code, I see it's using m_offline, migt have been because there is no easy way to tell cold wallet then... I thouht it was testing for cold, but there doesn't seem to be a way. In that case, the best might be to add a "make_cold" command, which would set a flag in wallet2, which could hten be tested. And the wallet could defualt to offline when this flag is set.
Do you see a drawback here ?

j-berman:
Agree, I don't see a better option than an explicit cold wallet flag

The primary drawback I can see is that cold wallet setups and implementations will break (they will need to add a flag to start a cold wallet). I would think wallets that implement cold signing should probably be notified and we'll need to e.g. update documentation [like this](https://monerodocs.org/cold-storage/offline-transaction-signing/) too

moneromooo-monero:
We could keep track of "have I ever refreshed blocks from a daemon". If so, reject attempt to import outputs.

woodser:
> We could keep track of "have I ever refreshed blocks from a daemon". If so, reject attempt to import outputs.

I like something like this in the spirit of simplicity and backwards compatibility. If a client initializes a wallet without the server uri, it's offline, and I would think it's eligible to import outputs. Otherwise, it'll be an extra step and breaking change for anyone currently using offline wallets.

j-berman:
+1^ sounds like a better solution

moneromooo-monero:
I added this check in #8513

woodser:
> I added this check in #8513

I'm getting an error "Nothing requested" when exporting outputs from a newly synced view-only wallet.

j-berman:
Closing this in favor of #8513 to consolidate discussion

PR #8515, opened 2022-08-18
Expose encrypt/decrypt to wallet rpc

jbakosi:
This enables (optionally signed) encryption and (optionally verified) decryption of arbitrary text by a wallet.


contrib/epee/src/string_tools.cpp
```diff
@@ -208,6 +208,14 @@ namespace string_tools
...
     return s;
   }
+
+  //----------------------------------------------------------------------------
+  std::string pad_to_div_by(size_t d, std::string s)
+  {
+    auto r = s.size() % d;
+    if (!r) return s;
+    return pad_string(s, s.size() + d - r);
+  }
```
moneromooo-monero, line 218: This will add d pointlessly if already a good size. Is this intended ?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -2209,6 +2210,71 @@ namespace tools
...
+
+    try
+    {
+      const auto& secret_key = m_wallet->get_account().get_keys().m_spend_secret_key;
+      auto padded_plaintext = string_tools::pad_to_div_by(4, req.plaintext);
+      res.pad = padded_plaintext.size() - req.plaintext.size();
+      auto ciphertext = m_wallet->encrypt(padded_plaintext, secret_key, req.authenticated);
+      const std::vector<uint8_t> binary(begin(ciphertext), end(ciphertext));
+      std::vector<char> z85(binary.size()*5/4 + 1);
+      if (not zmq_z85_encode(z85.data(), binary.data(), binary.size())) {
```
moneromooo-monero, line 2231: Why is this here and what does it do ?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -2209,6 +2210,71 @@ namespace tools
...
+        throw std::logic_error("z85 encode unsuccessful");
```
moneromooo-monero, line 2232: Does this cause an actual sane RPC error packet back (instead of returning false with an error string) ?

tests/functional_tests/CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ if (REQUESTS_OUTPUT STREQUAL "OK")
   add_test(
     NAME    functional_tests_rpc
     COMMAND ${PYTHON3_FOUND} "${CMAKE_CURRENT_SOURCE_DIR}/functional_tests_rpc.py" "${PYTHON3_FOUND}" "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_BINARY_DIR}" all)
+    message("ft: ${PYTHON3_FOUND} ${CMAKE_CURRENT_SOURCE_DIR}/functional_tests_rpc.py ${PYTHON3_FOUND} ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_BINARY_DIR} all")
```
moneromooo-monero, line 75: Forgotten debug trace

tests/functional_tests/encrypt_decrypt_message.py
```diff
@@ -0,0 +1,78 @@
...
+    - encrypt
+    - decrypt
+"""
+
+from framework.wallet import Wallet
+
+class StringEncryptionTest():
+    def run_test(self):
+      self.create()
+      for text in ['Restricted', 'Restricted0', 'Restricted01', 'Restricted012']:
```
moneromooo-monero, line 45: Please add empty string, and all byte lengths up to the cipher block size ?

tests/functional_tests/encrypt_decrypt_message.py
```diff
@@ -0,0 +1,78 @@
...
+            res = self.wallet[i].restore_deterministic_wallet(seed = seeds[i])
+            assert res.address == self.address[i]
+            assert res.seed == seeds[i]
+
+    def check_encrypt_decrypt(self, plaintext, authenticated):
+        print('Encrypting/decrypting strings with spend key with' + ('' if authenticated else 'out') + ' authentication')
+        encrypt_result = self.wallet[0].encrypt(plaintext, authenticated)
+        #print(encrypt_result)
+        decrypt_result = self.wallet[0].decrypt(encrypt_result.ciphertext_z85, encrypt_result.pad, authenticated)
+        #print(decrypt_result)
```
moneromooo-monero, line 74: Leftover traces

moneromooo-monero (commented):
If the size is not preserved, it should really be fixed in the encrypt/decrypt rather than being left to the user to keep an extra pad parameter.

src/wallet/wallet_rpc_server.cpp
jbakosi: If I change it to
```cpp
     if (zmq_z85_encode(z85.data(), binary.data(), binary.size())) {
        throw std::logic_error("z85 encode unsuccessful");
      }
```
it produces
```sh
Starting servers...
[TEST STARTED] encrypt_decrypt_message
Creating wallets
Encrypting/decrypting strings with spend key without authentication
  File "<monero>/tests/functional_tests/encrypt_decrypt_message.py", line 78, in <module>
    StringEncryptionTest().run_test()
  File "<monero>/tests/functional_tests/encrypt_decrypt_message.py", line 47, in run_test
    self.check_encrypt_decrypt(text, authenticated)
  File "<monero>/tests/functional_tests/encrypt_decrypt_message.py", line 71, in check_encrypt_decrypt
    encrypt_result = self.wallet[0].encrypt(plaintext, authenticated)
  File "<monero>/utils/python-rpc/framework/wallet.py", line 746, in encrypt
    return self.rpc.send_json_rpc_request(encrypt)
  File "<monero>/utils/python-rpc/framework/rpc.py", line 81, in send_json_rpc_request
    return self.send_request("/json_rpc", inputs, 'result')
  File "<monero>/utils/python-rpc/framework/rpc.py", line 73, in send_request
    assert 'error' not in res, res
AssertionError: {'error': {'code': -49, 'message': 'z85 encode unsuccessful'}, 'id': '0', 'jsonrpc': '2.0'}
[TEST FAILED] encrypt_decrypt_message
Stopping servers...
Done, 1/1 tests failed: encrypt_decrypt_message
```
Is this good or there is a better way?

tests/functional_tests/encrypt_decrypt_message.py
jbakosi: Removed.

jbakosi:
> If the size is not preserved, it should really be fixed in the encrypt/decrypt rather than being left to the user to keep an extra pad parameter.

I have thought about this too and I also don't like returning the `pad` arg. The problem is that `wallet2::encrypt()` needs a plaintext whose length must be divisible by 4. If it is not, the amount of padding needs to be trimmed after `wallet2::decrypt()`.

I see two options besides the current solution:
1. We keep `pad` in `wallet_rpc_server`'s state, but that introduces a constraint/ordering on client code having to call `wallet_rpc_server::decrypt()` after `wallet_rpc_server::encrypt()` on the same data. I don't like that at all.
2. We put the burden of dealing with `pad` on client code, which is also not great because it pushes the possibility of an error to a higher level, i.e., to client code.

The current one at least does not expose `pad` beyond wallet-rpc and can be used from a wallet/lib as a black box without having to worry about padding or ordering of multiple calls to `{en,de}crypt`. Maybe there is a third option, but I am not familiar with the code enough to see it.

tests/functional_tests/encrypt_decrypt_message.py
jbakosi: I added the empty-string text. What's the cipher block size? 64? Should I add tests for all string sizes from 0-64?

woodser:
@jbakosi any updates?

jbakosi:
Other than the discussion above, no.

jtgrassie:
@jbakosi 
>I see two options besides the current solution

I see a third option, don't use z85. 

If you use base64 (or even base16), you won't have to deal with passing a padding variable around. If someone wants to reduce bandwidth over the wire, compression can be used. base64 and base16 are widely supported also.

PR #8519, opened 2022-08-19
rpc: skip bootstrap nodes that are lower than last checkpoint

selsta:
~~Also outdated nodes now get 50 fail points instead of 2.~~ Not necessary anymore with #8544.

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -2290,6 +2290,12 @@ namespace cryptonote
         return m_bootstrap_daemon->handle_result(false, {});
       }
 
+      if (bootstrap_daemon_height < m_core.get_checkpoints().get_max_height())
+      {
+        MINFO("Bootstrap daemon height is lower than the latest checkpoint");
+        return m_bootstrap_daemon->handle_result(false, {});
+      }
+
       if (!m_p2p.get_payload_object().no_sync())
       {
         uint64_t top_height = m_core.get_current_blockchain_height();
--- a/src/cryptonote_core/blockchain.h
+++ b/src/cryptonote_core/blockchain.h
@@ -159,6 +159,13 @@ namespace cryptonote
     bool deinit();
 
     /**
+     * @brief get a set of blockchain checkpoint hashes
+     *
+     * @return set of blockchain checkpoint hashes
+     */
+    const checkpoints& get_checkpoints() const { return m_checkpoints; }
+
+    /**
      * @brief assign a set of blockchain checkpoint hashes
      *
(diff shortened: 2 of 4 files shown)


src/rpc/bootstrap_node_selector.h
```diff
@@ -83,7 +83,7 @@ namespace bootstrap_node
       std::string address;
       size_t fails;
 
-      void handle_result(bool success);
+      void handle_result(bool success, size_t score = 2);
```
serhack, line 86: Just a curiosity: why 2?

selsta: That's the existing number that got added in https://github.com/monero-project/monero/pull/6357

If a node has a lower height than our last checkpoint it should get more fail points than just 2, because it is useless to us.

serhack: Oh, right! Thanks for the explanation.

SChernykh:
> Also outdated nodes now get 50 fail points instead of 2

Hmm, I don't see "50" anywhere in this code.

selsta:
hmm, messed something up in git, will update

src/rpc/core_rpc_server.cpp
```diff
@@ -273,6 +273,8 @@ namespace cryptonote
     m_net_server.set_threads_prefix("RPC");
     m_net_server.set_connection_filter(&m_p2p);
 
+    m_checkpoints.init_default_checkpoints(nettype());
+
```
moneromooo-monero, line 277: Can you not use the existing one in cryptonote_core (maybe a bit later if not initialized yet) ?

jeffro256: Are you asking since it's a relatively expensive operation?

selsta: Updated.

selsta:
Verified that it initializes correctly, and tested mainnet, testnet and stagenet.

jeffro256 (changes requested):
reference to checkpoints object should be `const` since there's no good reason to modify your checkpoints st runtime

selsta:
@jeffro256 thanks, totally forgot to make it const

src/rpc/core_rpc_server.cpp
moneromooo-monero: I don't think it's that expensive. It just seems unnecessarily complex.

selsta:
Updated, with #8544 it isn't necessary anymore to set a custom fail points. This also makes the diff smaller.

j-berman (approved):

PR #8526, opened 2022-08-21
simplewallet: check status of parsing payment id

serhack:


moneromooo-monero (approved):

jeffro256:
I personally would just omit the `r` variable and do `if (tools::wallet2::...) { ... } else { ... }` for clarity and so someone doesn't recreate the same issue of misusing that status variable again down the line.

serhack:
Thanks for the feedback, I do agree with you. I'll commit that version.

PR #8527, opened 2022-08-22
daemon: remove redundant check

serhack:


jeffro256:
Why is this check redundant? If `start_block_index > 0`, then `(uint64_t) -start_block_index` is undefined.

vtnerd (approved):

vtnerd:
@jeffro256 expand the code section, this `if` statement is inside an `if` with a similar check. It doesn't hurt anything though; the compiler might be able to omit this check during optimzation phases.

jeffro256 (approved):
Ah, okay I see it now @vtnerd

serhack:
Exactly @vtnerd . By the way, I did not know if I should open an PR for that trivial issue. Please close if anyone thinks it's more clear to have the redundant check.

PR #8528, opened 2022-08-22
serialization: forbid reading 0 varints from empty data

moneromooo-monero:
this could plausibly be useful as a tail compression scheme,
but every other varint loading code treats 0 bytes read as an
error, and it goes with the principle of least surprise.

### DIFF
--- a/src/serialization/binary_archive.h
+++ b/src/serialization/binary_archive.h
@@ -145,7 +145,7 @@ struct binary_archive<false> : public binary_archive_base<false>
   {
     auto current = bytes_.cbegin();
     auto end = bytes_.cend();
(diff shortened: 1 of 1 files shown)


vtnerd:
The serialization code always behaved this way - so I'm a little hesitant to give an approval. However, the output side always writes at least one byte.

Is it it worth doing a full blockchain re-scan (skip the hashes) ... which would take a while. The only potential issue I can see is some historical block / transaction being encoded by a 3rd part that decodes properly. Seems like a small chance though.

moneromooo-monero:
That's a fair point. I guess someone can always insert such a tx after the patch goes in even if there's no historical example.

selsta:
Superseded by #10639.

PR #8534, opened 2022-08-23
build: prepare v0.18.1.1

selsta:
I would like to put out a point release somewhat soon with output export fixes, unicode path fixes and other smaller bug fixes.


BeholdersEye:
🍑
🔥

BeholdersEye:
> 10 days ago 

...

> somewhat soon

Clarification requested. Still awaiting a valid archive before updating from v0.17.

> All checks have passed

> This branch has no conflicts with the base branch

What is the impediment?

selsta:
Code is all ready. We need final review, merges and then a release.

BeholdersEye:
> You can also use v0.18.0.0 in the meantime

I was hoping to use the auto-updater and avoid having to manually verify the archive since I understood that was the purpose of having an auto-updater. But thanks for the suggestion.

j-berman (approved):

selsta:
@BeholdersEye everything is approved now, only thing missing is merges -> tag -> release. Don't have an exact date since it involves multiple people but likely in the next days, plus another day to activate auto updater.

BeholdersEye:
I am optimistic that Ubuntu's default archive manager will not crash when attempting to open the archive containing the release. Recommend checking the time stamps before making a release public just in case.

PR #8535, opened 2022-08-24
rpc: avoid superfluous buffer copy

jeffro256:
In the macro `MAP_URI_AUTO_BIN2`, the response body was written by streaming to a `epee::byte_stream` which is taken by a `epee::byte_slice` (by move) which is then copied into a `std::string` (the variable being `response_info.m_body`). This PR creates a new `store_t_to_binary` template helper which streams the formatted bytes directly into a `std::string`. This should help speed up requests to binary RPC endpoints.

This PR appears to have appropriate correctness (see unit test), but further testing is needed to compare performance.

### DIFF
--- a/contrib/epee/src/portable_storage.cpp
+++ b/contrib/epee/src/portable_storage.cpp
@@ -39,6 +39,30 @@
 #include <string>
 #include <sstream>
 
+namespace
+{
+  /**
+   * @brief Takes a std::string by reference and wraps it with ostream-like access
+   */
+  struct string_ref_stream
+  {
+    string_ref_stream(std::string& str_ref): m_str_ref(str_ref) {}
+
+    void write(const char* buf, const std::size_t len)
+    {
+      m_str_ref.insert(m_str_ref.size(), buf, len);
+    }
+
+    void write(const epee::span<const uint8_t>& buf_span)
+    {
+      write(reinterpret_cast<const char*>(buf_span.begin()), buf_span.size());
+    }
+
+    std::string& m_str_ref;
+  };
+
+} // anonymous namespace
+
 namespace epee
 {
 namespace serialization
@@ -57,6 +81,21 @@ namespace serialization
   bool portable_storage::store_to_binary(byte_stream& ss)
   {
     TRY_ENTRY();
+    storage_block_header sbh{};
+    sbh.m_signature_a = SWAP32LE(PORTABLE_STORAGE_SIGNATUREA);
+    sbh.m_signature_b = SWAP32LE(PORTABLE_STORAGE_SIGNATUREB);
+    sbh.m_ver = PORTABLE_STORAGE_FORMAT_VER;
+    ss.write(epee::as_byte_span(sbh));
+    pack_entry_to_buff(ss, m_root);
+    return true;
+    CATCH_ENTRY("portable_storage::store_to_binary", false);
+  }
+
+  bool portable_storage::store_to_binary(std::string& target, const std::size_t initial_buffer_size)
+  {
+    TRY_ENTRY();
+    target.reserve(initial_buffer_size);
+    string_ref_stream ss(target);
     storage_block_header sbh{};
     sbh.m_signature_a = SWAP32LE(PORTABLE_STORAGE_SIGNATUREA);
     sbh.m_signature_b = SWAP32LE(PORTABLE_STORAGE_SIGNATUREB);
--- a/contrib/epee/src/portable_storage.cpp
+++ b/contrib/epee/src/portable_storage.cpp
@@ -42,15 +42,15 @@
 namespace
 {
   /**
-   * @brief Takes a std::string by reference and wraps it with ostream-like access
+   * @brief Wraps a std::string by mutable reference, giving it epee::byte_stream-like access
    */
   struct string_ref_stream
   {
-    string_ref_stream(std::string& str_ref): m_str_ref(str_ref) {}
+    constexpr string_ref_stream(std::string& str_ref): m_str_ref(str_ref) {}
 
     void write(const char* buf, const std::size_t len)
     {
-      m_str_ref.insert(m_str_ref.size(), buf, len);
+      m_str_ref.append(buf, len);
     }
 
     void write(const epee::span<const uint8_t>& buf_span)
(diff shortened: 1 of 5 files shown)


contrib/epee/src/portable_storage.cpp
```diff
@@ -39,6 +39,30 @@
...
+  /**
+   * @brief Takes a std::string by reference and wraps it with ostream-like access
+   */
+  struct string_ref_stream
+  {
+    string_ref_stream(std::string& str_ref): m_str_ref(str_ref) {}
+
+    void write(const char* buf, const std::size_t len)
+    {
+      m_str_ref.insert(m_str_ref.size(), buf, len);
```
vtnerd, line 53: Trivial but this is : `m_str_ref.append(buf, len);`.

vtnerd (commented):
This should be a bit faster output on the HTTP servers with the downside that the entire epee binary serialization is re-generated for this additional type (there's basically two ASM versions - `byte_slice` and `std::string`).

contrib/epee/src/portable_storage.cpp
jeffro256: >  with the downside that the entire epee binary serialization is re-generated for this additional type.


Yes I was worried about this too, but my monerod binary size only increased from 15512512 bytes to 15525032 bytes (+12.5 KB).

jeffro256:
Also remove dead and unused `dump_as_xml`

jeffro256:
I made a repo which allows you to test deserialization speed: https://github.com/jeffro256/monero_8535_speedtest

I got quite impressive results:
```
old buffer simulation start
elapsed milliseconds: 27289
...
new string simulation start
elapsed milliseconds: 9254
...
```

This means that the process of converting an arbitrary type into binary was (27289 - 9254) / 9254 * 100% = 194% faster with the changes. This doesn't mean that the entire RPC call will be 3x as fast, but over 100K requests, 18 seconds of CPU time was saved.

selsta:
Not sure if I did something wrong.

```
selsta@mbp ~/d/monero_8535_speedtest (master)> ./xmr_8535_speedtest
old buffer simulation start
elapsed milliseconds: 4121
optimization prevent result: 18446744073709551607
new string simulation start
elapsed milliseconds: 3939
optimization prevent result: 18446744073709551587

selsta@mbp ~/d/monero_8535_speedtest (master)> ./xmr_8535_speedtest
old buffer simulation start
elapsed milliseconds: 4103
optimization prevent result: 131
new string simulation start
elapsed milliseconds: 3934
optimization prevent result: 133
```

jeffro256:
@selsta which CPU do you have? The first thing that comes to mind is that my CPU is suffering some real cache grief. I'll try the test again with different size payloads

selsta:
10 core M1 Pro

vtnerd:
`ubuntu_build.sh` script doesn't have an optimization lists. Both tests should be run with `-Ofast` to simulate the official build.

jeffro256:
@vtnerd good point. The speedtest repo was updated. Here is the new results for me:

```
old buffer simulation start
elapsed milliseconds: 22341
...
new string simulation start
elapsed milliseconds: 5936
...
```

I don't know why the new changes make it that much faster. Would you be willing to run it @vtnerd ? My CPU is a `Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz` btw

vtnerd:
I think the bulk of the time is due to the behavior of `byte_stream` / `byte_slice` rather than the copy itself. As proof (on a `i7-1185G7 @ 3.00GHz`) the current code:

```
old buffer simulation start
elapsed milliseconds: 5475
...
new string simulation start
elapsed milliseconds: 1777
```
but if the [second constructor here](https://github.com/monero-project/monero/blob/5256fdd7a155f7543bd710e991d5dd407981b35a/contrib/epee/src/portable_storage.cpp#L52) has a `false` added as a second parameter, then the timings become:
```
old buffer simulation start
elapsed milliseconds: 2034
...
new string simulation start
elapsed milliseconds: 1793
```
which is a more modest 11.8% improvement. If the copy is removed, then `byte_slice` wins out slightly (at least on this box), so its likely that the 11.8% represents copy time.

---------

We might need to add the boolean flag to the constructor. The flag toggles whether the buffer is truncated (`byte_stream` / `byte_slice` beats `std::string` typically because it aggressively allocates initially).

jeffro256:
@vtnerd That's very good. We could change this commit into a one-line change and get almost as good results for our systems. @selsta's system seems to have extremely poor `string.append` performance (probably even more conservative allocation as you mentioned) compared to `byte_stream` so I think we should go with that.

selsta:
With `-Ofast`

```
old buffer simulation start
elapsed milliseconds: 2270
...
new string simulation start
elapsed milliseconds: 2188
```

With `-Ofast` + `false` added like suggested

```
old buffer simulation start
elapsed milliseconds: 2265
...
new string simulation start
elapsed milliseconds: 2194
```

vtnerd:
@selsta There's almost no difference on macOS m1? My best guess is the heap implementation isn't bothering to free the memory - if the chunk is big enough the heap does a `mmap` which should be unbacked by physical memory until a soft page fault. So the heap could just check the "arena" the chunk is in, and do a NOP when `mmap`ed.

jeffro256:
@vtnerd why would `mmap` be used here (besides the initial file reading)? What file descriptor is being mapped?

vtnerd:
@jeffro256 I was thinking in terms of Linux syscalls, macOS may be slightly different but should have similar syscalls.

On Linux processes can request memory from the OS via `sbrk` or an "anonymous" `mmap` (no "real" file backing the memory). I looked into glibc details when creating `byte_stream` - glibc will use an anonymous `mmap` when the `alloc` request is over a certain size because they can be individually `free`'d (`munmap`) back to the OS. And with `mmap`, `madvise` can be used to "punch" holes in the physical/resident allocation (an anonymous `mmap` provides a chunk of virtual memory with no physical pages backing the chunk).

jeffro256:
@vtnerd Is pretty much the process you're describing? (On https://linux.die.net/man/3/malloc): "Normally, malloc() allocates memory from the heap, and adjusts the size of the heap as required, using [sbrk](https://linux.die.net/man/2/sbrk)(2). When allocating blocks of memory larger than MMAP_THRESHOLD bytes, the glibc malloc() implementation allocates the memory as a private anonymous mapping using [mmap](https://linux.die.net/man/2/mmap)(2). MMAP_THRESHOLD is 128 kB by default, but is adjustable using [mallopt](https://linux.die.net/man/3/mallopt)(3)"

That 128 kB limit seems applicable for my speedtest since the payload is 216 kB. So are you guessing that glibc will deallocate a 128kB chunk while shrinking the `byte_slice`  whereas macOS doesn't? Thanks for the information btw

jeffro256:
Closed in favor of #8543

PR #8540, opened 2022-08-30
Update levin async callbacks for C++14

vtnerd:
C++14 allows for `std::move`s into lambda captures, so this updates for that feature in the async framework for levin (p2p). This also updates the internal callback code within levin to reduce the copying as data is being received (timeout callback is moved/copied once, and thereafter only a `shared_ptr` is copied). With PR #7345 the callbacks become "safer" with `shared_ptr`/`weak_ptr` concepts.

Some testing with asan+tsan was also performed on the unit tests and `monerod`.

### DIFF
--- a/contrib/epee/include/net/levin_protocol_handler_async.h
+++ b/contrib/epee/include/net/levin_protocol_handler_async.h
@@ -109,7 +109,7 @@ public:
   std::chrono::milliseconds m_invoke_timeout;
 
   template<class callback_t>
-  int invoke_async(int command, message_writer in_msg, boost::uuids::uuid connection_id, const callback_t &cb, std::chrono::milliseconds timeout = LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED);
+  int invoke_async(int command, message_writer in_msg, boost::uuids::uuid connection_id, callback_t &&cb, std::chrono::milliseconds timeout = LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED);
 
   bool send(epee::byte_slice message, const boost::uuids::uuid& connection_id);
   bool close(boost::uuids::uuid connection_id, const bool wait_for_shutdown);
@@ -223,12 +223,13 @@ public:
     }
 
   public:
-    anvoke_handler(const callback_t& cb, const std::chrono::milliseconds timeout, std::shared_ptr<net_utils::service_endpoint<derived_handler>> con, int command)
+    template<typename F>
+    anvoke_handler(F&& cb, const std::chrono::milliseconds timeout, std::shared_ptr<net_utils::service_endpoint<derived_handler>> con, int command)
       : invoke_response_handler_base(),
         std::enable_shared_from_this<anvoke_handler<callback_t>>(),
         m_con(con),
         m_timer(con->get_io_context()),
-        m_cb(std::move(cb)),
+        m_cb(std::forward<F>(cb)),
         m_timeout(timeout),
         m_command(command),
         m_cancel_timer_called(false),
@@ -318,7 +319,7 @@ public:
   }
   
   template<class callback_t>
-  bool add_invoke_response_handler(const callback_t &cb, const std::chrono::milliseconds timeout,  std::shared_ptr<net_utils::service_endpoint<derived_handler>> con, int command)
+  bool add_invoke_response_handler(callback_t &&cb, const std::chrono::milliseconds timeout,  std::shared_ptr<net_utils::service_endpoint<derived_handler>> con, int command)
   {
     CRITICAL_REGION_LOCAL(m_invoke_response_handlers_lock);
     if (m_protocol_released)
@@ -326,7 +327,8 @@ public:
       MERROR("Adding response handler to a released object");
       return false;
     }
-    std::shared_ptr<invoke_response_handler_base> handler(std::make_shared<anvoke_handler<callback_t>>(cb, timeout, std::move(con), command));
+    std::shared_ptr<invoke_response_handler_base> handler =
+      std::make_shared<anvoke_handler<std::decay_t<callback_t>>>(std::forward<callback_t>(cb), timeout, std::move(con), command);
     if (handler->reset_timer(true))
     {
       m_invoke_response_handlers.push_back(std::move(handler));
@@ -631,7 +633,7 @@ public:
   }
 
   template<class callback_t>
-  bool async_invoke(std::shared_ptr<net_utils::service_endpoint<derived_handler>> self, int command, message_writer in_msg, const callback_t &cb, std::chrono::milliseconds timeout = LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED)
+  bool async_invoke(std::shared_ptr<net_utils::service_endpoint<derived_handler>> self, int command, message_writer in_msg, callback_t &&cb, std::chrono::milliseconds timeout = LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED)
   {
     assert(self && this == std::addressof(self->m_protocol_handler));
 
@@ -653,7 +655,7 @@ public:
         break;
       }
 
-      if(!add_invoke_response_handler(cb, timeout, std::move(self), command))
+      if(!add_invoke_response_handler(std::forward<callback_t>(cb), timeout, std::move(self), command))
       {
         err_code = LEVIN_ERROR_CONNECTION_DESTROYED;
         break;
@@ -774,13 +776,13 @@ std::shared_ptr<net_utils::service_endpoint<typename get_handler<t_connection_co
 }
 //------------------------------------------------------------------------------------------
 template<class t_connection_context> template<class callback_t>
-int async_protocol_handler_config<t_connection_context>::invoke_async(int command, message_writer in_msg, boost::uuids::uuid connection_id, const callback_t &cb, const std::chrono::milliseconds timeout)
+int async_protocol_handler_config<t_connection_context>::invoke_async(int command, message_writer in_msg, boost::uuids::uuid connection_id, callback_t &&cb, const std::chrono::milliseconds timeout)
 {
   std::shared_ptr<levin_endpoint> con = find_and_lock_connection(connection_id);
   if (!con)
     return LEVIN_ERROR_CONNECTION_NOT_FOUND;
   levin_endpoint& ref = *con;
-  return ref.m_protocol_handler.async_invoke(std::move(con), command, std::move(in_msg), cb, timeout);
+  return ref.m_protocol_handler.async_invoke(std::move(con), command, std::move(in_msg), std::forward<callback_t>(cb), timeout);
 }
 //------------------------------------------------------------------------------------------
 template<class t_connection_context> template<class callback_t>
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -2527,7 +2527,7 @@ namespace nodetool
   }
   //-----------------------------------------------------------------------------------
   template<class t_payload_net_handler> template<class t_callback>
-  bool node_server<t_payload_net_handler>::try_ping(basic_node_data& node_data, p2p_connection_context& context, const t_callback &cb)
+  bool node_server<t_payload_net_handler>::try_ping(basic_node_data& node_data, p2p_connection_context& context, t_callback &&cb)
   {
     if(!node_data.my_port)
       return false;
@@ -2570,9 +2570,9 @@ namespace nodetool
       address = epee::net_utils::network_address{epee::net_utils::ipv6_network_address(ipv6_addr, node_data.my_port)};
     }
     peerid_type pr = node_data.peer_id;
-    bool r = zone.m_net_server.connect_async(ip, port, zone.m_config.m_net_config.ping_connection_timeout, [cb, /*context,*/ address, pr, this](
+    bool r = zone.m_net_server.connect_async(ip, port, zone.m_config.m_net_config.ping_connection_timeout, [cb = std::forward<t_callback>(cb), /*context,*/ address, pr, this] (
       const typename net_server::t_connection_context& ping_context,
-      const boost::system::error_code& ec)->bool
+      const boost::system::error_code& ec) mutable ->bool
     {
       if(ec)
       {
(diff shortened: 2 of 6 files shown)


src/p2p/net_node.inl
```diff
@@ -2364,7 +2364,7 @@ namespace nodetool
       address = epee::net_utils::network_address{epee::net_utils::ipv6_network_address(ipv6_addr, node_data.my_port)};
     }
     peerid_type pr = node_data.peer_id;
-    bool r = zone.m_net_server.connect_async(ip, port, zone.m_config.m_net_config.ping_connection_timeout, [cb, /*context,*/ address, pr, this](
+    bool r = zone.m_net_server.connect_async(ip, port, zone.m_config.m_net_config.ping_connection_timeout, [cb = std::forward<t_callback>(cb), /*context,*/ address, pr, this](
```
UkoeHB, line 2367: If `cb` is an r-value reference, won't this move `cb` into the lambda, making the move down in `[this, pr, cb = std::move(cb), ping_context = std::move(ping_context), address = std::move(address)]` invalid?

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -190,115 +190,102 @@ class async_protocol_handler
...
     {
-      if(!m_cancel_timer_called)
+      if (!self || (!initial && !self->cancel_timer()))
+        return false;
+      if (initial && !self->m_con.start_outer_call())
+        return false;
+
+      auto& self_ = *self;
+      self_.m_timer.expires_from_now(boost::posix_time::milliseconds(self_.m_timeout));
+      self_.m_timer.async_wait([self = std::move(self)](const boost::system::error_code& ec)
```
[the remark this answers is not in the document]
UkoeHB: For example, if `async_wait()` were to move the lambda into another function, then after that function call tried to access a data member of `self_.m_timer`, the data member won't exist because the lambda would be destroyed in the function call, so there will be a segfault.

src/p2p/net_node.inl
vtnerd: No, but I think I need to add the `mutable` keyword here because `cb` should be implicit `const`, forcing a silent copy at the `std::move`.

UkoeHB: Oh yeah I forgot that lambda captures are `const` by default. Thanks for the reminder

contrib/epee/include/net/levin_protocol_handler_async.h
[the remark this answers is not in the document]
vtnerd: Crap, I can't believe I missed this. The safety of this is dependent on how `async_wait` is implemented internally (technically the only callee ensures safety here, but this is easy to miss in a refactor).

I did this elsewhere, and it definitely needs to be corrected in a couple of places (incoming PR!). Some of the existing cases uses an `asio::strand` though, and so you can "get away with it" if the `std::move` is done inside the strand (because the next op has to wait in the queue until the current instance returns).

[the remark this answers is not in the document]
vtnerd: After further thought, I think this is valid, but I'm still likely to change it.

There are plenty of examples of a single async callback starting another instance of itself in a "loop". The documentation states that no synchronization is needed because it is an implicit single strand of calls. If this line here is an issue, then those types of setups would also be an issue.

I looked at some of the implementations for the async functions, and they all seem to never touch the internals after queuing an operation. It's probably for the "implicit" strand case; I could find no examples of this exact setup.

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -190,115 +190,102 @@ class async_protocol_handler
...
+    invoke_response_handler_base(uint64_t timeout, async_protocol_handler& con, int command)
+      : m_timer(con.m_pservice_endpoint->get_io_service()),
+        m_con(con),
+        m_timeout(timeout),
+        m_command(command)
+    {}
+
+    virtual ~invoke_response_handler_base() {}
+
+    //! Run user handler. Only call if `cancel_timer()` returns true.
```
jeffro256, line 211: Thanks for the comments :)

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -190,115 +190,102 @@ class async_protocol_handler
...
-    async_protocol_handler& m_con;
+  protected:
     boost::asio::deadline_timer m_timer;
-    bool m_timer_started;
-    bool m_cancel_timer_called;
-    bool m_timer_cancelled;
-    uint64_t m_timeout;
-    int m_command;
-    virtual bool handle(int res, const epee::span<const uint8_t> buff, typename async_protocol_handler::connection_context& context)
+    async_protocol_handler& m_con;
```
jeffro256, line 195: What is the rationale for moving `anvoke_handler` logic into `invoke_response_handler_base` but not combining the two?

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -190,115 +190,102 @@ class async_protocol_handler
...
+      : m_timer(con.m_pservice_endpoint->get_io_service()),
+        m_con(con),
+        m_timeout(timeout),
+        m_command(command)
+    {}
+
+    virtual ~invoke_response_handler_base() {}
+
+    //! Run user handler. Only call if `cancel_timer()` returns true.
+    void handle(int res, const epee::span<const uint8_t> buff)
```
jeffro256, line 212: It would save an extra 8-byte copy to make `buff` a `const` reference instead of pass-by-value

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -190,115 +190,102 @@ class async_protocol_handler
...
-    bool m_cancel_timer_called;
-    bool m_timer_cancelled;
-    uint64_t m_timeout;
-    int m_command;
-    virtual bool handle(int res, const epee::span<const uint8_t> buff, typename async_protocol_handler::connection_context& context)
+    async_protocol_handler& m_con;
+    const uint64_t m_timeout;
+    const int m_command;
+
+    virtual void do_handle(int res, const epee::span<const uint8_t> buff)=0;
```
jeffro256, line 199: Same here: It would save an extra 8-byte copy to make buff a const reference instead of pass-by-value

contrib/epee/include/net/levin_protocol_handler_async.h
vtnerd: You mean, why not drop `anvoke_handler` completely in favor of `std::function` type erasure? The only reason is that this saves one allocation (most of the callbacks exceed the small size optimization of `std::` or `boost::` `function`.

If you meant something else, then you'll need to clarify what you meant here.

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -372,7 +359,7 @@ class async_protocol_handler
 
     // Never call callback inside critical section, that can cause deadlock. Callback can be called when
     // invoke_response_handler_base is cancelled
-    std::for_each(local_invoke_response_handlers.begin(), local_invoke_response_handlers.end(), [](const boost::shared_ptr<invoke_response_handler_base>& pinv_resp_hndlr) {
+    std::for_each(local_invoke_response_handlers.begin(), local_invoke_response_handlers.end(), [](const std::shared_ptr<invoke_response_handler_base>& pinv_resp_hndlr) {
       pinv_resp_hndlr->cancel();
     });
```
jeffro256, line 364:
```suggestion
    for (const auto& pinv_resp_hndlr : local_invoke_response_handlers) {
      pinv_resp_hndlr->cancel();
    }
```

This is clearer IMO and removes the possibility of unnecessary overhead from `std::for_each` and the closure.

vtnerd: Your right, since I'm touching this anyway, might as well update to it to C++11+ foreach loops.

contrib/epee/include/net/levin_protocol_handler_async.h
vtnerd: Hmm. Typically this is bad practice because the caller is required to provide a memory address for the span. On most (modern) ABIs this often results in less efficient code (because at most call-sites there is no existing span on the stack). In this instance, the function is invoked from one spot, and that chunk of code probably/maybe has a span on the stack already. 

I'm tempted to leave it because its dependent on the setup of the existing code. Could be persuaded differently, I guess.

vtnerd: I was just reminded by this [blog post(https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/) that by value is likely preferred here.

The callee doesn't have to generate indirect memory dereferences (the values are directly in the register), and it doesn't have to worry about the span being aliased (better optimizations in some cases).

I doubt either is going to matter much in this situation, but its a series of tradeoffs to save one move instruction in the optimal case.

contrib/epee/include/net/levin_protocol_handler_async.h
jeffro256: I mean that `invoke_response_handler_base` use to be a pure virtual base class, but now it contains a partial implementation. I'm not saying it's necessarily worse, but now there isn't that clean interface / implementation distinction. I was just wondering the reason for that decision. Since it's not used as a base class anywhere else, why not just combine `invoke_response_handler_base` and `anvoke_handler`?

contrib/epee/include/net/levin_protocol_handler_async.h
jeffro256: Yeah fair enough. The only reason I mentioned it was because it was just passing it through to `do_handle`. But either way, the profiling isn't worth effort. Nevermind this comment.

jeffro256:
> Some testing with asan+tsan was also performed on the unit tests and monerod

Curious, what results did that produce?

contrib/epee/include/net/levin_protocol_handler_async.h
vtnerd: I wanted to remove the `virtual` call for resetting the timer, thus the split of internal members. But otherwise, that wasn't strictly necessary.

However, some indirect call (`std::function`, function pointer, `virtual`, etc.) was still needed because of the templated callback. The `std::vector` has to store multiple different lambdas from `src/p2p/p2p.inl`. So that template argument isn't "fixed" to the levin handler, but rather generic per async invoke.

vtnerd:
>> Some testing with asan+tsan was also performed on the unit tests and monerod
> Curious, what results did that produce?

There was the single memory leak, fixed and recently merged. There are still several outstanding tsan errors, that do not appear related to this code. I couldn't immediately spot the source, so I left that for a bit later :/

contrib/epee/include/net/levin_protocol_handler_async.h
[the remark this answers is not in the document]
jeffro256: You could pass `self` as `std::weak_ptr`: https://floating.io/2017/07/lambda-shared_ptr-memory-leak/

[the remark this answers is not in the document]
vtnerd: It might be worth changing (despite the small bump in CPU cycles), because its a better guarantee that no memory leak occurs. The callback is invoked so infrequently, that I will just go ahead and make that change too.

I don't see how `release_protocol()` can be removed though, otherwise I would've make this a `std::weak_ptr` already. The current behavior is to invoke `handle(...)` when destroying the connection, and doing this automatically in destructor cleanup is iffy because `m_connection_context` could already be destroyed.

vtnerd:
Updated based on comments, and rebased to master.

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -190,115 +190,103 @@ class async_protocol_handler
...
     {
-      if(!m_cancel_timer_called)
+      if (!self || (!initial && !self->cancel_timer()))
+        return false;
+      if (initial && !self->m_con.start_outer_call())
+        return false;
+
+      std::weak_ptr<invoke_response_handler_base> self_ = self;
+      self->m_timer.expires_from_now(boost::posix_time::milliseconds(self->m_timeout));
+      self->m_timer.async_wait([self_ = std::move(self_)](const boost::system::error_code& ec)
```
UkoeHB, line 244: I don't follow how `weak_ptr` is better here than the reference you had before. It doesn't impact the lifetime of the wrapped pointer.

vtnerd: I'm not sure what you mean. The `weak_ptr` here adds little, which is why I didn't do it in the first attempt. It does guarantee cleanup of the callback before the `connection` destructor completes (instead of relying on `release_protocol()` to make that guarantee).

I could move the `weak_ptr` to the `deque` but that doesn't do much but add CPU cycles when invoking the callback.

Perhaps you were thinking of a change like #7345 where the callback has a `weak_ptr` to the `connection` object, instead of the reference. It might be possible to make that change with less changes, but I didn't see how.

vtnerd: Oh and @jeffro256 recommended switching to a `weak_ptr` here, to be extra cautious against cycles (and see first paragraph from prior response).

vtnerd: Sorry for the spam. After some thought - I shouldn't have made this change to `weak_ptr` here. I'll revert back to `shared_ptr` here. The `weak_ptr` here is always going to be valid because `start_outer_call()` forces a `shared_ptr` instance until `finish_outer_call()` is invoked.

The actual problem with lifetimes is fixed in #7345. This may have been what @jeffro256 was noticing, its somewhat messed up now because its not "correct" usage of `shared_ptr` / `weak_ptr` on the outer class.

This should be a `shared_ptr` and so should the one in the `deque<...>`. The internal reference in the callback here needs to be a `weak_ptr` to break the cycle and properly synchronize with destruction in all cases (it works now afaik, but is iffy and requires going through much more code to verify).

vtnerd:
Reverted `weak_ptr` usage.

contrib/epee/include/net/levin_protocol_handler_async.h
jeffro256: @UkoeHB `weak_ptr` in this context wouldn't change the lifetime of the pointer, but it would allow detection if the lifetime ended too early to prevent segfaults. Also like @vtnerd mentioned, it would prevent `shared_ptr` cycles.

vtnerd:
Rebased and updated to fix conflicts.

jpk68 (approved):

src/p2p/net_node.inl
```diff
@@ -2581,19 +2581,11 @@ namespace nodetool
...
-      auto cb_ = cb;*/
-
-      // GCC 5.1.0 gives error with second use of uint64_t (peerid_type) variable.
-      peerid_type pr_ = pr;
 
       network_zone& zone = m_network_zones.at(address.get_zone());
 
       bool inv_call_res = epee::net_utils::async_invoke_remote_command2<COMMAND_PING::response>(ping_context, COMMAND_PING::ID, req, zone.m_net_server.get_config_object(),
-        [=](int code, const COMMAND_PING::response& rsp, p2p_connection_context& context)
+        [this, pr, cb = std::move(cb), ping_context = std::move(ping_context), address = std::move(address)](int code, const COMMAND_PING::response& rsp, p2p_connection_context& context)
```
jpk68: I think the `std::move` should be dropped on `address`, as it seems to be read after being moved into the inner lambda.

Disclaimer: this was found by an LLM

vtnerd: This is a good catch, it looked like all references are in the inner lambda, but one is in the outer one.

jpk68 (approved):

jeffro256 (approved):

jeffro256 (approved):

selsta (approved):

PR #8543, opened 2022-08-31
epee: dont shrink slice when storing to binary

jeffro256:
competing with #8535 

Shrinking a moved `byte_stream` in `byte_slice::byte_slice(epee::byte_stream&&, bool)` is very expensive (in glibc) so avoiding shrinking can speed up storing types to binary in RPC calls. This PR is 1 line instead of 100.

Credit: @vtnerd


vtnerd (approved):
This might only have an effect on glibc+Linux, but thats the bulk of `monerod` deployment, so the change is likely worth it.

PR #8544, opened 2022-08-31
wallet2: check wallet compatibility with daemon's hard fork version [release]

j-berman:
### Overview

Since the fork, there have been a number of support requests caused by pre-fork clients pointing to post-fork daemons and vice versa. This PR ensures hard fork compatibility between client and daemon, and paves the way for a cleaner UX when client and daemon are incompatible. Credit to [reddit user u/MoneroArbo](https://www.reddit.com/r/monerosupport/comments/ws94u4/where_are_your_coins_update_your_wallet_do_not/ikz744n/) for the catalyst to this PR:

> post 0.18 wallets should at least warn you when a node reports a pre-HF version :(

### Implementation

1. If an **updated wallet** tries to connect to a **daemon that has not updated** for the latest hard fork AND the chain height has passed the fork height, the wallet will fail to connect to that daemon and error out.

2. If a **wallet has not updated** for the latest hard fork and tries to connect to an **updated daemon** AND the chain height has passed the fork height, the wallet will fail to connect to that daemon and error out.

3. If an **updated wallet** connects to a **daemon that has not updated** for the latest fork **when the fork height has not yet passed**, and then the fork height passes while the wallet is still connected, the wallet will error out when scanning blocks and give a clean error indicating the daemon has not updated.

4. If a **wallet has not updated** for the latest fork and connects to an **updated daemon** when the fork height **has not yet passed**, and then the fork height passes while the wallet is still connected, the wallet will error out when scanning blocks and give a clean error indicating the wallet has not updated.

### Reasoning

I think this versioning approach makes sense over using the RPC versioning system as-is because this approach allows wallets to function smoothly while connecting to incompatible daemons **until the hard fork height**. It also sets up a nice way for wallets to warn that a hard fork is coming in X blocks/days (though this PR does not give that warning), and to tell users to update either their daemon or wallet as needed, rather than immediately break wallets connecting to incompatible daemons upon release.


src/cryptonote_basic/hardfork.h
```diff
@@ -231,6 +231,11 @@ namespace cryptonote
      */
     uint64_t get_window_size() const { return window_size; }
 
+    /**
+     * @brief returns info for all known hard forks
+     */
+    std::vector<hardfork_t> get_hardforks() const { return heights; }
```
SChernykh, line 237: It's more efficient to return `const std::vector<hardfork_t>&` here.

src/cryptonote_core/blockchain.h
```diff
@@ -892,6 +892,13 @@ namespace cryptonote
      */
     uint64_t get_earliest_ideal_height_for_version(uint8_t version) const { return m_hardfork->get_earliest_ideal_height_for_version(version); }
 
+    /**
+     * @brief returns info for all known hard forks
+     *
+     * @return the hardforks
+     */
+    std::vector<hardfork_t> get_hardforks() const { return m_hardfork->get_hardforks(); }
```
SChernykh, line 900: Same, it's more efficient to return `const std::vector<hardfork_t>&` here.

SChernykh (approved):

SChernykh:
@j-berman looks good. Did you test it on a real v0.17 node?

j-berman:
@SChernykh yep :) here's one: http://support.ablative.hosting:18081

I manually tested every code path on real nodes where possible, and my own modified node where not

j-berman:
Rebased

SChernykh (approved):

src/wallet/node_rpc_proxy.cpp
```diff
@@ -82,28 +82,49 @@ void NodeRPCProxy::invalidate()
...
     cryptonote::COMMAND_RPC_GET_VERSION::response resp_t = AUTO_VAL_INIT(resp_t);
     {
       const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};
       bool r = net_utils::invoke_http_json_rpc("/json_rpc", "get_version", req_t, resp_t, m_http_client, rpc_timeout);
       RETURN_ON_RPC_RESPONSE_ERROR(r, epee::json_rpc::error{}, resp_t, "get_version");
     }
+
     m_rpc_version = resp_t.version;
+    for (const auto &hf : resp_t.hard_forks)
+      m_daemon_hard_forks.push_back(std::make_pair(hf.hf_version, hf.height));
```
jeffro256, line 110: If this method is called multiple times on the same object, then it's `m_daemon_hard_forks` field will keep expanding indefinitely.

j-berman: Added `m_daemon_hard_forks.clear()` above the for loop for better defense

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -2123,15 +2123,32 @@ namespace cryptonote
...
+      uint64_t target_height;
+      std::vector<entry> hard_forks;
 
       BEGIN_KV_SERIALIZE_MAP()
         KV_SERIALIZE_PARENT(rpc_response_base)
         KV_SERIALIZE(version)
         KV_SERIALIZE(release)
+        KV_SERIALIZE(current_height)
+        KV_SERIALIZE(target_height)
+        KV_SERIALIZE(hard_forks)
```
jeffro256, line 2151: `COMMAND_RPC_GET_VERSION::response_t` packets serialized by pre-PR daemons will not be able to be deserialized by post-PR wallets (`node_rpc_proxy` specifically). I would supply some optional values through `KV_SERIALIZE_OPT` to maintain backwards compatibility.

j-berman: > I would supply some optional values through KV_SERIALIZE_OPT

I agree this seems safer, see the latest.

FWIW before this change, I had tested a wallet running the PR's code pointing to pre-PR daemons and vice versa; it worked fine deserializing with 0 values thanks to [this](https://github.com/monero-project/monero/blob/6402dbee69867c76530b45e8a21b692850fea03f/src/wallet/node_rpc_proxy.cpp#L98). But I prefer handling it more explicitly as you suggested.

jeffro256 (approved):

src/simplewallet/simplewallet.cpp
```diff
@@ -4786,20 +4783,28 @@ bool simple_wallet::try_connect_to_daemon(bool silent, uint32_t* version)
...
     version = &version_;
-  if (!m_wallet->check_connection(version))
+  bool wallet_is_outdated, daemon_is_outdated = false;
+  if (!m_wallet->check_connection(version, NULL, 200000, &wallet_is_outdated, &daemon_is_outdated))
   {
     if (!silent)
     {
       if (m_wallet->is_offline())
         fail_msg_writer() << tr("wallet failed to connect to daemon, because it is set to offline mode");
+      else if (wallet_is_outdated)
```
hinto-janaiyo, line 4793: It seems all errors always take this branch, the elseif/else below won't activate. I tried using a v0.18 wallet on a couple v0.17 nodes as well (including the one linked: http://support.ablative.hosting:18081), same thing.

These all return the old wallet error:
- `monerod` online, but it's v0.17.x.x
- `monerod` online, but connected to wrong port
- `monerod` offline, either local or remote
- random fake ip: `1.2.3.4:18081`
- bad ip: `1:18081` (returns wallet outdated error after hanging for a bit)

PR #8545, opened 2022-09-01
multisig: fix seed restore (suggestions by @UkoeHB) [release]

### LINKED ISSUE #8537: Restoring multisig wallet fails
Steps to reproduce:

1. Generate 2/3 multisig wallet (the M and N are presumably irrelevant).
2. In the CLI wallet, run `seed` or `encrypted_seed` with no passphrase. Save the seed.
3. Run the CLI wallet with `--restore-multisig-wallet`.
4. Enter the seed from step 2.
5. Observe an error: `Error: failed to generate new wallet: invalid multisig seed`

If we use `encrypted_seed` with some passphrase, the CLI wallet segfaults on step 5 instead of showing an error.

Happens with both v0.17.3.2 and v0.18.1.0.

j-berman:
Solves the `invalid multisig seed` error when restoring a multisig wallet, reported by @tmoravec in #8537. Fixes suggested by @UkoeHB.

### [Fix 1](https://github.com/monero-project/monero/issues/8537#issuecomment-1233717683)

> multisig accounts no longer store a sum of multisig keyshares in the spend_secret_key slot. Instead it is the 'base multisig privkey', which is the private key used to sign key exchange messages post-update (and for the first step of key exchange both pre and post update). This explains why new multisig wallets are throwing an error (can be fixed by just deleting lines 4750-4754; there is no verification step like this for post-update spend secret keys).

### [Fix 2](https://github.com/monero-project/monero/issues/8537#issuecomment-1234611444)

> If restoring from 'multisig keys', presumably the wallet is fully set-up, so it's fine to set the rounds passed [in wallet2::generate()].

### Misc

Tested restoring both a v0.17.2.0 and v0.18.1.0 multisig wallet from plaintext seed using this PR applied to v0.18.1.0. Working smoothly on my end.

I have not yet gone deeper into the segfault reported with an encrypted seed in #8537. [UPDATE: solved in this PR]

### DIFF
--- a/src/multisig/multisig_account.cpp
+++ b/src/multisig/multisig_account.cpp
@@ -127,7 +127,7 @@ namespace multisig
   bool multisig_account::multisig_is_ready() const
   {
     if (main_kex_rounds_done())
-      return m_kex_rounds_complete >= multisig_kex_rounds_required(m_signers.size(), m_threshold) + 1;
+      return m_kex_rounds_complete >= multisig_setup_rounds_required(m_signers.size(), m_threshold);
     else
       return false;
   }
@@ -198,6 +198,13 @@ namespace multisig
     CHECK_AND_ASSERT_THROW_MES(num_signers >= threshold, "num_signers must be >= threshold");
     CHECK_AND_ASSERT_THROW_MES(threshold >= 1, "threshold must be >= 1");
     return num_signers - threshold + 1;
+  }
+  //----------------------------------------------------------------------------------------------------------------------
+  // EXTERNAL
+  //----------------------------------------------------------------------------------------------------------------------
+  std::uint32_t multisig_setup_rounds_required(const std::uint32_t num_signers, const std::uint32_t threshold)
+  {
+    return multisig_kex_rounds_required(num_signers, threshold) + 1;
   }
   //----------------------------------------------------------------------------------------------------------------------
 } //namespace multisig
--- a/src/multisig/multisig_account.h
+++ b/src/multisig/multisig_account.h
@@ -245,4 +245,13 @@ namespace multisig
   * return: number of kex rounds required
   */
   std::uint32_t multisig_kex_rounds_required(const std::uint32_t num_signers, const std::uint32_t threshold);
+
+  /**
+  * brief: multisig_setup_rounds_required - The number of setup rounds required to produce an M-of-N shared key.
+  *    - A participant must complete all kex rounds and 1 initialization round.
+  * param: num_signers - number of participants in multisig (N)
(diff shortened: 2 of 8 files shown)


UkoeHB (approved):
Thanks for spearheading this

tmoravec:
I've just tested this branch. Restoring 2/3 wallets, receiving funds into them, and sending outgoing transactions, all on stagenet. Wallets created with:

* v0.17.2.0
* v0.17.3.2
* v0.18.1.0

* * *

One semi-related suggestion: set enable-multisig-experimental to 1 when restoring a multisig wallet.

UkoeHB:
> One semi-related suggestion: set enable-multisig-experimental to 1 when restoring a multisig wallet.

I think any use of multisig should be gated by the enable-multisig-experimental, in case any user hasn't encountered it before.

j-berman:
I figure an explicit confirmation on restore couldn't hurt (see latest commit). But true that any use of multisig *after* restoring is still gated

UkoeHB (approved):

tmoravec:
@j-berman I meant the exact opposite. The user has _already_ explicitly said they want a multisig wallet with `--restore-multisig-wallet`. Hence we can set `enable-multisig-experimental` to 1. And any warnings are pointless when restoring a wallet anyway because the wallet already exists. 

So now we ask the user to say it THREE times (first on the CLI, then on restore, then on any use). Which is going beyond superfluous.

selsta:
> So now we ask the user to say it THREE times (first on the CLI, then on restore, then on any use). Which is going beyond superfluous.

Every user should know that multisig is in its current form experimental and unsafe for production (if untrusted participants are involved) until we have security proofs. Someone creating a multisig wallet before we marked it as experimental should also explicitly approve it.

j-berman:
> Someone creating a multisig wallet before we marked it as experimental should also explicitly approve it.

This is why I figured it makes sense to add on restore.

In the latest commit, I kept the explicit confirmation I added for restore, and then the user's confirmation enables multisig into perpetuity for the restored wallet. In sum:

- In order to create a multisig wallet, a user must `set experimental-multisig-enabled 1` once.
- In order to restore a multisig wallet from seed, a user must explicitly acknowledge multisig is experimental once via confirmation, and then the user does not need to `set experimental-multisig-enabled 1` again in order to perform any multisig actions after.

UkoeHB (approved):

j-berman:
PR onto this branch that solves the segfault when restoring an encrypted seed: https://github.com/j-berman/monero/pull/5

Figure it would be easier to test/review this way, and it won't hold up the release.

UkoeHB (approved):

PR #8546, opened 2022-09-02
Revert "Reduce comp. time of portable_storage_template_helper.h"

selsta:
This reverts commit 673c6d2d11a4ae639c43bbf40e31e808932fd625.

Causes issues with unicode characters in path: https://github.com/monero-project/monero-gui/issues/4002

Found by @tobtoht

If someone wants to look into this deeper we can avoid reverting it but for now this is a major issue so reverting is the easiest solution.


tobtoht (approved):
Might as well replace those tabs for spaces.

selsta:
#8547

PR #8557, opened 2022-09-08
serde: replace JSON and epee binary serialization

jeffro256:
The Monero repository currently uses a DOM model for p2p serialization (portable_storage in epee). This is sub-optimal for two reasons: inefficiency and underspecification. During normal operation of a node, even with no bad actors, a Monero node wastes valuable processor time and memory to build an in-memory model of serialized/deserialized data before copying that data into/out of C++ types. This is the "inefficient" component. However, in the worst case, as seen during the 2020 Christmas attack, this method of deserialization can have disatrous consequences for the network. Since, under a DOM model, an incoming packet can not be determined to contain nonsensical data until it is fully deserialized (as long the packet is not malformed according to the data format), this leaves a lot of degrees of freedom for crafters of DoS packets to exploit.

The aim of this PR is to create an easy to use, mostly backwards compatible interface for fully specifying the process of serialization (mostly deserialization)  for each C++ type without a DOM to mitigate the above issues. Currently, this PR only replaces JSON and epee serialization, but the interface/model should be flexible enough to integrate Boost portable archives in the future (maybe others), but I'll leave that to a later PR since this is already getting massive. This PR needs a lot of refinement, documentation, unit tests, and performance tests but the general concepts and code layout are finished.

The name of this PR and the design decisions I made are ripped off from the [Rust Serde Framework](https://serde.rs/). I wanted to mirror  an existing framework so I didn't have to make up and explain terminology ;). There are quite a few differences because the languages are so different. For example, the `Serialize` and `Deserialize` traits in Rust have been replaced by ADL functions `serialize_default` and `deserialize_default`, respectively. There is also `serialize_as_blob` and `deserialize_as_blob` which serve a similar purpose but transform types to/from a stream of calls differently. Also, `Deserializer`s doesn't specialize each method based on a `Visitor` passed, but rather are type-agnostic and let `Visitors` store the values, or references to values, themselves. They need to be collected later by `(de/ser)rialize_(default/as_blob)`.

Unfortunately, I lied a little before. Not all JSON deserialization is DOM-free in this PR. The older discrete JSON REST APIs are DOM-free. However the `/json_rpc` endpoints are not. This is because the JSON data itself determines which C++ type will be the target of the deserialization by way of the `"method"` parameter. If the `"params"` field is encountered before `"method"`, then the JSON deserializer can not know the specification to deserialize to any particular type until after the `"params"` field is already parsed. Instead of writing a custom DOM object, I created `ValueDeserializer` in `contrib/serde/include/serde/json/value.h` which allowed deserializing to/from the underlying DOM type: `rapidjson::Value`.

Potential alternatives to this PR:
* #7999 
* https://github.com/vtnerd/monero/tree/replace_p2p_serialization


vtnerd:
Ignore first comment, I went through the code a bit more, it was only fair :/

Found the code out-of-order, looks like it adapts everything using `tuple_cat` and `make_tuple`. I avoided tuple because I was attempting a pull-parser which required a c-array of field names/indexes (possible with `tuple` but clunky). The runtime performance is likely close to my variant for processing unknown/out-of-order fields.

I'm still not certain why this patch was made, its so similar to what I've already done ... ? The primary difference is that this is a "push" parser with the visitor design instead of a "pull" parser. What does the visitor pattern/push parser add here? It creates an extra layer of indirection, but I'm not certain of the benefit (at the moment). It might make it easier to add specification output support.

How much runtime considerations (in comparison to the other two variants) have you done? It looks like there will be a difficult to avoid vtable call per field due to the visitor design, whereas my variant is selectable based on derived/based being given to `wire::object(...)` and perfect-daemon always doing direct calls (with no option for indirect/code size optimization). And if you tried to compress executable size (using the deserializer base for code expansion through the cpp), you end up with two vtable calls per field (`deserialize_any` and then one in the call to the visitor) ... ?

jeffro256:
@vtnerd Thanks for taking the time to write that all out engage with this! I'll try to answer all of it

> Does this support out-of-order fields? What about unknown fields?

Yes. In this PR, the `BEGIN_KV_SERIALIZE_MAP` pattern creates a method which generates a tuple of `StructFields` on the stack which contain references to each field to be serialized. The `deserialize_default` operator for structs uses linear search over this tuple, and no particular order is required. However, the struct fields are always serialized in the order in which they appear in the `KV_SERIALIZE` pattern, which could be leveraged to increase performance in best case scenarios, as yours and perfect-daemon's PRs do. By handling unknown fields, do you mean that the unknown fields are ignored or have the possibility of being ignored? If so, then yes. By default, an error message is thrown if an unknown field is encountered, but the way that deserialization is handled, this could be changed in a couple lines.

I will admit, field performance is definitely the area in my PR which needs the most work. Right now it is very flexible, but could use some optimizing. I was much limited to how I built the fields using macros since this PR retains the old `KV_SERIALIZE` interface, unlike yours. So some speed is sacrificed for the sake of backwards code compatibility. Searching and indexing take O(N) time and can't/don't utilize any particular order fields are deserialized in. If the reviewers deem it worth the effort to scrap the old `KV_SERIALIZE` interface for the sake of speeding up the fields, I would gladly do so. I just wanted to start off more conservative if applicable. 

> We can probably omit out-of-order processing for p2p traffic, but probably not RPC. Definitely not JSON-RPC code - there are already multiple engines in use, etc. Even EPEE-RPC could be dicey at this point.

Ideally, in the future, we should switch to a serialization format which uses no keys and which there is no possible reordering of fields (think Boost portable archive but less language dependent) for p2p traffic. This would simplify deserialization significantly and reduce packet size. 

> Also, why create a 3rd version?

Correct me if I'm wrong, but since the `wire` module of yours was initially created for JSON, the `reader`s and `writers` don't make any distinction between all the different sizes of integers needed to support the epee binary format (i.e. `int16_t` vs `int32_t`), correct? For certain struct fields, this makes it impossible to create backwards compatible epee binary serialization. Also, a big design difference in my PR is the `Visitor` set of classes. This gives greater control over the deserialized data a field can/can't accept, especially for data whose structure is unknown. In addition, like I mentioned earlier, I retained the `KV_SERIALIZE` pattern in the repository and added adapters for the `epee::serialization::(load/store)...` set of functions. In my eyes, this will decrease the likelihood of a logic error slipping through.

vtnerd:
> Yes. In this PR, the BEGIN_KV_SERIALIZE_MAP pattern creates a method which generates a tuple of StructFields on the stack which contain references to each field to be serialized. The deserialize_default operator for structs uses linear search over this tuple, and no particular order is required. However, the struct fields are always serialized in the order in which they appear in the KV_SERIALIZE pattern, which could be leveraged to increase performance in best case scenarios, as yours and perfect-daemon's PRs do. By handling unknown fields, do you mean that the unknown fields are ignored or have the possibility of being ignored? If so, then yes. By default, an error message is thrown if an unknown field is encountered, but the way that deserialization is handled, this could be changed in a couple lines.

Yup, found it just before you posted. And yes, ignoring unknown fields is currently mandatory to upgrade and handle multiple active p2p client versions. A version field per `object` would be needed as a replacement in this PR; a version field per archive is likely risky due to unintentional errors.

> I will admit, field performance is definitely the area in my PR which needs the most work. Right now it is very flexible, but could use some optimizing. I was much limited to how I built the fields using macros since this PR retains the old KV_SERIALIZE interface, unlike yours. So some speed is sacrificed for the sake of backwards code compatibility. Searching and indexing take O(N) time and can't/don't utilize any particular order fields are deserialized in. If the reviewers deem it worth the effort to scrap the old KV_SERIALIZE interface for the sake of speeding up the fields, I would gladly do so. I just wanted to start off more conservative if applicable.

This is my frustration with the patch really. I spent time investigating (via disassembly) on how to optimize field unpacking performance, and then this patch just resets me entirely. I _think_ the compiler will avoid de-virtualizing the visitor calls due to the double-dispatch design. Basically, the compiler would have to emit unique "push" ASM foreach visitor type in the main processing loop. This seems unlikely.

If the compiler refuses to de-virtualize, this also hits the CPU harder because it will be unable to properly speculate on the jump address since (nearly) each field will swap out to a unique visitor type. A [recently blog entry](https://johnysswlab.com/the-true-price-of-virtual-functions-in-c/) suggests a 2.5x slowdown in addition to the ~18% slowdown of tight-loop vtable calls (when comparing a quick function call). This will probably only matter when we have many C++ objects in an array. Otherwise, performance will probably be close-ish ?

> Ideally, in the future, we should switch to a serialization format which uses no keys and which there is no possible reordering of fields (think Boost portable archive but less language dependent) for p2p traffic. This would simplify deserialization significantly and reduce packet size.

This sounds like hell to me, as it would have to be done across two hard-forks or something. Between the forks two versions of each command would have to be supported (we cannot guarantee support of new format until all versions are kicked from network). Its doable, but really messy for the handshake (as initiator must use old version and then "upgrade" based on server/responder state/response). This info could also be passed in peerlist info, but it seems like such a waste to track it.

> Correct me if I'm wrong, but since the wire module of yours was initially created for JSON, the readers and writers don't make any distinction between all the different sizes of integers needed to support the epee binary format (i.e. int16_t vs int32_t), correct? For certain struct fields, this makes it impossible to create backwards compatible epee binary serialization. 

I kinda of wish you would've emailed me or something.

My variant is backwards compatible - the current engine already accepts "oversized" values on the wire. If alternate implementations have problems (only an issue with RPC really), the `writer` class can have a function overload per size just like this PR. It almost never matters for the `reader`, unless you want to move a few runtime checks to a cpp instead of having them done inline. Lastly, if the derived `epee_writer` class is provided to `wire::object(...)`, the exact type is written on the wire anyway (oversized value is only written if using `writer` base class). So there is zero change for p2p output (because it uses derived type as there is no reason to use vtable). I also tested this against the live testnet, so it definitely has some compatibility with integers in both directions.

> Also, a big design difference in my PR is the Visitor set of classes. This gives greater control over the deserialized data a field can/can't accept, especially for data whose structure is unknown.

I don't see any difference - this is just using a "push" design so a visitor is needed.

C++ has function overloading and ADL. If a field needs a unique output separate from the defaults, just wrap it with `struct custom_output { std::string& self; };`, and then provide free functions for `custom_output`. This is similar to the visitor pattern you have here, except the free functions are "pulling" directly from the parser instead passing a visitor to the parser.

In fact, I'm not certain of anything that can be done with this design over the pull method in my branch. Every unique type has a free function that injects a visitor in the callpath for "push"ing values. Again, maybe schema output is easier? Its not obvious how at the moment.

> In addition, like I mentioned earlier, I retained the KV_SERIALIZE pattern in the repository and added adapters for the epee::serialization::(load/store)... set of functions. In my eyes, this will decrease the likelihood of a logic error slipping through.

The `KV_SERIALIZE` could be maintained, but commas would have to be added between fields. See `contrib/epee/include/serialization/wire.h` for the "begin" and "end" macros. Otherwise, I would need a clever way to terminate the list, probably using a tuple to c-array conversion something or other.

Basically, no difference here because adding commas into the existing macro system is a bit noisy, but low on potential bugs. I decided to just move everything instead for compile performance reasons.

PR #8563, opened 2022-09-12
Update CMakeLists.txt to fix Ubuntu 22.04 static builds

LocalMonero:
Ubuntu 22.04 build target `release-static` requires the addition of the libbsd and libmd libraries for libzmq, otherwise the build fails.

It also requires you remove the libzmq3-dev installed from the main ubuntu repository and instead compile it yourself:

```
git clone https://github.com/zeromq/libzmq.git
cd libzmq
./configure --enable-static --disable-shared --disable-libunwind --disable-perf --with-pic --without-docs --with-libsodium
make
sudo make install
cd ../
```


BeanTosser:
This fails due to the "configure" file not being present in the cloned libzmq directory.  There is a "configure.ac" file, but no proper "configure".

LocalMonero:
@BeanTosser ignore this PR. Use #8564 instead.

Issue #8439, opened 2022-07-16
ubuntu static-release target fails with undefined symbols


bladedoyle:
Trying to build v0.18.0.0

Getting undefined symbols linking on ubuntu 20.04 using "static-release" target.

Additional libraries are needed for static linking executables - required by libunbound

/usr/lib/x86_64-linux-gnu/libhogweed.a 
/usr/lib/x86_64-linux-gnu/libgmp.a 
/usr/lib/x86_64-linux-gnu/libnettle.a  
/usr/lib/x86_64-linux-gnu/libevent.a

Reference:
https://packages.debian.org/sid/libs/libunbound8

```
[ 96%] Linking CXX executable ../../bin/monero-gen-ssl-cert
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libunbound.a(ub_event_pluggable.o): in function `my_signal_new':
(.text+0xb9): undefined reference to `event_set'
...
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libunbound.a(val_secalgo.o): in function `_verify_nettle_rsa':
...
/usr/bin/ld: (.text+0x695): undefined reference to `__gmpz_clear'
...
(.text+0xb9): undefined reference to `event_set'
...
```

selsta:
Why not use `make depends target=x86_64-linux-gnu` instead? It will compile all the dependencies for you.

If you don't want to use `make depends` then try compile static unbound yourself: https://github.com/monero-project/monero/blob/master/contrib/depends/packages/unbound.mk#L11

bladedoyle:
Yes, building the depends rather than installing packages does work.
Though, it doesnt guess at the target architecture, and it also doesnt build/link static binaries.
Im able to call the arch guessing script as part of the command:
`make -j$(nproc) depends target=$(contrib/depends/config.guess)`

Should I close this issue, or leave it open to track the issue with static build/linking?

selsta:
What do you mean it doesn't build / link static binaries? We use depends to create release binaries which are static linked.

> Should I close this issue, or leave it open to track the issue with static build/linking?

I'm not aware of any issues, if you want to compile static binaries without depends you have to correctly compile unbound library. I've linked in my previous comment how you have to compile it.

bladedoyle:
The issue would be:  Previous to this release it was possible to build and link using package manager installed dependency libraries using the "static-release" target, and now that fails due to missing libraries in the cmake generated link.txt file.

[...]

PR #8564, opened 2022-09-12
Fix static builds for Ubuntu 22.04

LocalMonero:
Link libbsd, libmd, libprotokit in CMakeLists.txt for ZMQ to fix static builds for Ubuntu 22.04


selsta (approved):

jeffro256:
How were you linking before so that it didn't work? I'm using Ubuntu 22.04 and it works fine for me on `master`.

LocalMonero:
@jeffro256 You were running `make release-static` and it worked on your clean Ubuntu 22.04 install?

jeffro256:
Hi @LocalMonero, sorry it took so long to get back to you. I was trying out a bunch of Ubuntu installations trying to reproduce the linking errors that you were getting but I was not able to. From a clean installation of Ubuntu Server 22.04 on a VM, I 1) did `sudo apt update && sudo apt upgrade && sudo apt install <all monero dependencies>` 2) cloned the MiniUPNP repo and built and installed from source 3) cloned the monero repo and ran `make release-static`. I was able to build `release-static` without any errors. Is there anything else you can think of that you might have done to your install of Ubuntu 22 that would cause this error?

LocalMonero:
@jeffro256 we didn't clone and build MiniUPnP, only libzmq3.

jeffro256:
Sorry, that's what I meant to say: I built and installed libunbound from source. Do you have the package `libunbound-dev` installed? If you do, what happens when you `sudo apt remove` that package, then do a `make clean` and rebuild monero?

selsta:
I needed this patch to compile release-static on a fresh Ubuntu 22.04 machine.

jeffro256:
Does this work without removing the `libunbound-dev` package?

selsta:
I assume it depends where you install the custom compiled unbound and which path you tell CMake.

jeffro256 (approved):
This fixes when `libunbound`'s dependencies are linked statically through the `--enable-static --disable-shared` flags when `./configure`ing.

jtgrassie:
Shouldn't all these checks be wrapped in `if(STATIC)` so they only apply to static builds?

LocalMonero:
@jeffro256 building with `sudo apt install libunbound-dev` doesn't work, static or not. Only self-compiled libunbound works,

jeffro256:
@LocalMonero Right. I was saying that when the `libunbound-dev` package is installed, our CMake rules for finding `libunbound` might have linked against the wrong library (I don't know if they did or not). For those who are linking Monero statically, it is probably best to remove that package, as long as you don't have other packages which depend on it.

LocalMonero:
@jeffro256 libunbound has been an undocumented issue (the README.md still recommends using the package manager `libunbound-dev`) issue on Ubuntu builds ever since it stopped being vendored (#8439). On Ubuntu 22.04, assuming self-compiled libunbound is installed, this patch fixes the static build. Without this patch the static build fails. With the package manager libunbound the static build also fails, with or without this patch.

jeffro256:
> Shouldn't all these checks be wrapped in `if(STATIC)` so they only apply to static builds?

Probably? It *probably* shouldn't negatively affect non-static builds, but it would be likely be the better option, at least so the code is self-documenting in that this is a patch for static builds.

busyboredom:
This fixes builds for me too (also ubuntu 22.04). Would be nice if we could get this merged so master would build out-of-the-box again.

selsta:
@busyboredom it's in the merge queue already

jeffro256:
@selsta what do you think of @jtgrassie 's comment?

jtgrassie:
@jeffro256 
> Probably? It probably shouldn't negatively affect non-static builds, but it would be likely be the better option, at least so the code is self-documenting in that this is a patch for static builds.

My general stance is this: If it's not needed (based on make options), don't link it.

selsta:
We already link to libraries that are only needed for static, so it's fine to change it but I would do it in a separate commit later.

Issue #8531, opened 2022-08-23
`scan_tx` causes extra transfers with fee=0


woodser:
Calling `scan_txs` with already scanned tx ids causes `get_transfers` to start returning erroneous transfers with fee=0, which causes downstream errors reconciling the information reported from the wallet for libraries / applications.

It's accompanied with this output in monero-wallet-rpc, but this is probably expected:

2022-08-21 13:21:40.823	E Public key fdfa07f837a16a90d733fceb55581c0c0f8c6ba3574ef928738512083c65622a from received 35.173300860152 output already exists with spent 35.173300860152 in tx <ea6c9edae116224a5494e78bbeafc5508cc11734bc8e6e36a9784c6b7dcf160f>, received output ignored

j-berman:
I'm not seeing how to repro extra transfers with fee=0, but I can repro extra transfers with amount=0 and see what's causing that.

Are you certain the issue is extra transfers with fee=0 and not amount=0?

[This code](https://github.com/j-berman/monero/commit/7db2aa914e252c6a1bd83e80c3accb0d36e40e2a) should prevent the amount=0 issue, though I don't think it's a complete solution as is.

selsta:
Should be fixed with #8566

PR #8566, opened 2022-09-14
wallet2: fix rescanning tx via scan_tx [release]

j-berman:
## Issues the PR solves with `scan_tx`

- Calling `scan_tx` on already processed received txs causes duplicates. (#8531)
- Scanning txs out of order can corrupt the wallet and render it unable to recover from a reorg correctly.
- Coinbase txs don't get scanned as miner txs via `scan_tx`.
- If a user provides a tx with height > the wallet's known scan height, `get_transfers` will return inaccurate num confirmation data for that tx until the wallet refreshes.
- If a wallet has already processed an **incoming** tx in the pool, the tx is then mined, and then the user calls `scan_tx` with that tx before the wallet refreshes, the tx will be in a state of both "unconfirmed" and "confirmed" until the wallet refreshes.

## UX changes the PR brings

- `scan_tx` now requires a trusted daemon edit: if the user attempts to scan a tx that is not their most recent tx. I chose to add this requirement because if a user passes a txid into `scan_tx` that has a height lower than any txs the wallet already processed, then the wallet will re-request *all* of the wallet's already processed txs from the daemon. This reveals more of a user's txs to the daemon than the user might otherwise expect. Given the way the code is structured, it would seem the simplest solution to this is to require the feature be used with a trusted daemon.
- If a user provides a tx with height > the wallet's known scan height, the wallet will continue scanning from that height. If a user misses scanning a tx because of this, they either have to scan the missed tx manually with `scan_tx`, or use `rescan_blockchain`.

## Future considerations

- Assume tx2 spends an output received in tx1. tx2 is a sweep to some other wallet's address. The user calls `scan_tx(tx2)` first before `scan_tx(tx1)`. As is, the wallet is unable to figure out that tx2 spends from tx1 in this case (the user has to re-scan tx2 after tx1). In order to solve this, the wallet would need to query the daemon to see if tx1's associated key image(s) have been spent and in what tx(s).
- In `get_transfers`, if a daemon is connected, query the daemon to retrieve the latest height and block reward rather than use the wallet's latest state.
- When the user provides a tx to `scan_tx` with height > the wallet's known scan height, don't call `refresh` inside `scan_tx`. In this case, `scan_tx` waits until the wallet is sync'd or the user stops the wallet before responding. Use `fast_refresh` instead, or just don't do this at all once `get_transfers` uses the daemon's latest height and block reward.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1170,6 +1170,7 @@ wallet2::wallet2(network_type nettype, uint64_t kdf_rounds, bool unattended, std
   m_first_refresh_done(false),
   m_refresh_from_block_height(0),
   m_explicit_refresh_from_block_height(true),
+  m_skip_to_height(0),
   m_confirm_non_default_ring_size(true),
   m_ask_password(AskPasswordToDecrypt),
   m_max_reorg_depth(ORPHANED_BLOCKS_MAX_COUNT),
@@ -1624,14 +1625,13 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
   return m_subaddress_labels[index.major][index.minor];
 }
 //----------------------------------------------------------------------------------------------------
-void wallet2::scan_tx(const std::vector<crypto::hash> &txids)
+wallet2::tx_entry_data wallet2::get_tx_entries(const std::unordered_set<crypto::hash> &txids)
 {
-  // Get the transactions from daemon in batches and add them to a priority queue ordered in chronological order
-  auto cmp_tx_entry = [](const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry& l, const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry& r)
-  { return l.block_height > r.block_height; };
+  tx_entry_data tx_entries;
+  tx_entries.tx_entries.reserve(txids.size());
 
-  std::priority_queue<cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry, std::vector<COMMAND_RPC_GET_TRANSACTIONS::entry>, decltype(cmp_tx_entry)> txq(cmp_tx_entry);
   const size_t SLICE_SIZE =  100; // RESTRICTED_TRANSACTIONS_COUNT as defined in rpc/core_rpc_server.cpp, hardcoded in daemon code
+  std::unordered_set<crypto::hash>::const_iterator it = txids.begin();
   for(size_t slice = 0; slice < txids.size(); slice += SLICE_SIZE) {
     cryptonote::COMMAND_RPC_GET_TRANSACTIONS::request req = AUTO_VAL_INIT(req);
     cryptonote::COMMAND_RPC_GET_TRANSACTIONS::response res = AUTO_VAL_INIT(res);
@@ -1640,7 +1640,10 @@ void wallet2::scan_tx(const std::vector<crypto::hash> &txids)
 
     size_t ntxes = slice + SLICE_SIZE > txids.size() ? txids.size() - slice : SLICE_SIZE;
     for (size_t i = slice; i < slice + ntxes; ++i)
-     req.txs_hashes.push_back(epee::string_tools::pod_to_hex(txids[i]));
+    {
+      req.txs_hashes.push_back(epee::string_tools::pod_to_hex(*it));
+      ++it;
+    }
 
     {
       const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};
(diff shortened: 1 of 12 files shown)


plowsof:
I built the GUI with this PR. I confirmed the issue (duplicate entries in transactions with "Unknown" amount displayed after scanning a tx multiple times). This PR fixes the problem in the GUI, Thanks!

afungible (changes requested):
Suggested a few cosmetic & improvement comments.

moneromooo-monero:
tbh I'd skip the entire thing, partly processing seems like a recipe for trouble. Especially if the new scan results do not agree with the previous ones: I see you test amount, but it's not the only thing that could be different. And if it is different, it's a bug somewhere which isn't helped by the testing anyway.

And return some bool to the higher level caller for "already known" so the user knows.

woodser:
I'm no longer seeing extra transfers with this PR, so that's good.

I do see a few differences between scanned txs (left) versus regularly synced txs (right):

image

image

Looks like the only things missing are the number of confirmations (which should be knowable since the wallet has a connection to the daemon), and detecting if it's a miner tx.

j-berman:
> tbh I'd skip the entire thing, partly processing seems like a recipe for trouble

@moneromooo-monero this was my first thought too. When I looked into it, I think that will be fairly involved to do correctly unless I'm missing something. AFAICT will need to check `m_payments`, `m_unconfirmed_payments`, `m_confirmed_txs`, `m_unconfirmed_txs` to make sure the tx in the wallet's state matches what the daemon says the tx's state is, and if not, re-process. I also think this means the pool needs to be re-processed the same as it's done when scanning, but without re-querying for everything in the pool (I think some factored functions from #8076 would be re-used for this).

This PR seemed like the best quick fix that gets us to a better spot, although the concern for partly processing causing trouble is warranted I think. It seems the best move is to just do it all correctly now.

_________


@woodser seems these are existing issues. Going to take me a bit to get to it but what I have so far in case someone wants to pick it up:

1. The height used to set confirmations is the wallet's highest scanned block height:

https://github.com/monero-project/monero/blob/c9cfa251833856d76d5e206a4546e7fe3f71e43e/src/wallet/wallet_rpc_server.cpp#L362

https://github.com/monero-project/monero/blob/c9cfa251833856d76d5e206a4546e7fe3f71e43e/src/wallet/wallet2.h#L1106

https://github.com/monero-project/monero/blob/c9cfa251833856d76d5e206a4546e7fe3f71e43e/src/wallet/wallet2.cpp#L2621 

And then confirmations are set to 0 if the tx's height >= `blockchain_height`:

https://github.com/monero-project/monero/blob/c9cfa251833856d76d5e206a4546e7fe3f71e43e/src/wallet/wallet_rpc_server.cpp#L97-L98

As suggested, probably makes sense to use the daemon's height instead to set confirmations.

2. To pick up miner txs, `process_new_transaction` inside `scan_tx` can pass `cryptonote::is_coinbase(tx)` as the param for `bool miner_tx` instead of false (evidently miner txs weren't implemented for this feature). Might not be sufficient to solve miner txs, but that's a start.

woodser:
> seems these are existing issues. Going to take me a bit to get to it but what I have so far in case someone wants to pick it up:

Hoping this can be worked out, because it's quite a useful feature.

But either way, this PR makes the existing `scan_tx` stable as far as I can tell.

j-berman:
Approach from @moneromooo-monero to avoid trouble from calling `process_new_transaction` with an already scanned tx:

1: If a user tries to call `scan_tx` on a tx with a height lower than the wallet's highest scanned height, they get an error.
2: The highest manually scanned tx's height becomes the wallet's highest scanned height. (this also fixes the 0 confirmation issue @woodser pointed out [above](https://github.com/monero-project/monero/pull/8566#issuecomment-1255299200))

The UX downside is that a user may end up skipping their txs, and would need to do a full wallet rescan to fix it:
- assume a user's wallet has scanned up to block n.
- the user has received 2 txs: tx1 at block n+1, and tx2 at block n+2.
- the user calls `scan_tx` on tx2.
- the issue: the wallet will have skipped tx1. User must do a full wallet rescan to scan tx1.

A future PR could add the ability for a user to call `scan_tx` on a tx at a height lower than the last scanned height, so that they don't necessarily have to do a full wallet rescan if they know tx1's hash. Pseudo-implementation: inside `scan_tx`, pop already processed txs with height > the tx to be scanned, then re-process all txs in order.

selsta:
> 1: If a user tries to call scan_tx on a tx with a height lower than the wallet's highest scanned height, they get an error.

Doesn't this make `scan_tx` basically useless?

j-berman:
I see what you mean. Sounds like this would have to be the way to go as part of the approach then:

> inside `scan_tx`, pop already processed txs with height > the tx to be scanned, then re-process all txs in order.

j-berman:
Clarifying the approach:

1: If a user calls `scan_tx` on a tx with a height lower than any txs already scanned, pop those already scanned txs and then re-process all in order.
2: The wallet's highest scanned height becomes the max of `[highest manually scanned tx height, the wallet's existing highest scanned height]`

And to clarify the trouble the approach is seeking to avoid: scanning txs out of order messes up the `m_transfers` container, which is supposed to be ordered by a tx's height. Example: reorg handling expects transfers to be ordered by height:

https://github.com/monero-project/monero/blob/fc907a957078cae2dc68348886a33363848dd089/src/wallet/wallet2.cpp#L3761

This approach both prevents incorrect duplicate insertions when re-scanning, and ensures a correctly ordered transfers container.

afungible:
> trouble the approach is seeking to avoid: scanning txs out of order messes up the `m_transfers` container, which is supposed to be ordered by a tx's height.

I haven't deeply looked at this code. But from a quick look, can't the `m_transfers` container,  `std::vector<transfer_details>` be sorted forcefully by `tx_height` at the end of each "scan_tx" irrespective. Could this also be an approach? Only downside I can think of is std::sort(..) slowing down with too many tx in container (I doubt though).

j-berman:
To be honest, `process_new_transaction` and its complete impact on wallet2's state is such a headache to assess that I can't tell with absolute certainty if it's safe to send in either duplicates or out-of-order txs and then try to fix doing that after. I understand @moneromooo-monero's hesitation sending anything but new (and correctly ordered transactions) into the function. The "scorched earth" strategy of popping first and then re-processing seems easier to sleep at night with.

j-berman:
A better answer: assume tx2 spends an output from a prior tx1. When we call `process_new_transaction` on tx2, it's supposed to set tx1's output to spent via `set_spent`:

https://github.com/monero-project/monero/blob/fc907a957078cae2dc68348886a33363848dd089/src/wallet/wallet2.cpp#L2315-L2355

Therefore `process_new_transaction` *has* to process tx1 *before* processing tx2, in order to know that tx1's output has been spent in tx2 when processing tx2.

tests/functional_tests/transfer.py
```diff
@@ -62,18 +72,13 @@ def reset(self):
...
-        self.wallet = [None] * len(seeds)
-        for i in range(len(seeds)):
+        self.wallet = [None] * len(SEEDS)
+        for i in range(len(SEEDS)):
             self.wallet[i] = Wallet(idx = i)
             # close the wallet if any, will throw if none is loaded
             try: self.wallet[i].close_wallet()
             except: pass
-            res = self.wallet[i].restore_deterministic_wallet(seed = seeds[i])
+            res = self.wallet[i].restore_deterministic_wallet(seed = SEEDS[i])
```
moneromooo-monero, line 81: Seriously ? :D

moneromooo-monero (commented):
Allowing pool txes seem odd, otherwise looks good.

src/wallet/wallet2.cpp
```diff
@@ -1616,23 +1617,32 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
...
     cryptonote::COMMAND_RPC_GET_TRANSACTIONS::request req = AUTO_VAL_INIT(req);
     cryptonote::COMMAND_RPC_GET_TRANSACTIONS::response res = AUTO_VAL_INIT(res);
     req.decode_as_json = false;
     req.prune = true;
 
-    size_t ntxes = slice + SLICE_SIZE > txids.size() ? txids.size() - slice : SLICE_SIZE;
+    size_t ntxes = slice + SLICE_SIZE > txids_no_dups.size() ? txids_no_dups.size() - slice : SLICE_SIZE;
     for (size_t i = slice; i < slice + ntxes; ++i)
-     req.txs_hashes.push_back(epee::string_tools::pod_to_hex(txids[i]));
+      req.txs_hashes.push_back(epee::string_tools::pod_to_hex(txids_no_dups[i]));
```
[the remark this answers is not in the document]
j-berman: Was trying to de-dup in here since I noticed the RPC server doesn't de-dup, updated to just use an `unordered_set` for txids passed to `scan_tx` instead

j-berman:
Fixed an issue in my code where sweeps wouldn't re-process when calling `scan_tx` with a single txid that took place at a height below the sweep. Will rebase in a few days.

j-berman:
Squashed and rebased. Also noting #8617 is dependent on this code.

src/wallet/wallet2.cpp
```diff
@@ -1651,17 +1654,142 @@ void wallet2::scan_tx(const std::vector<crypto::hash> &txids)
...
+  auto cmp_tx_entry = [](const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry& l, const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry& r)
+    { return l.block_height < r.block_height && !l.in_pool; };
+  std::sort(sorted_tx_entries.tx_entries.begin(), sorted_tx_entries.tx_entries.end(), cmp_tx_entry);
+
+  return sorted_tx_entries;
+}
+//----------------------------------------------------------------------------------------------------
+std::vector<cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry> merge_sorted_tx_entries(const std::vector<cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry> &l, const std::vector<cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry> &r)
+{
+  std::vector<cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry> merged_txs;
```
jeffro256, line 1677: `merged_txs.reserve(l.size() + r.size());`

src/wallet/wallet2.cpp
```diff
@@ -5781,6 +5927,28 @@ void wallet2::trim_hashchain()
...
+    r = net_utils::invoke_http_json_rpc("/json_rpc", "getblockheaderbyheight", req, res, *m_http_client, rpc_timeout);
+    if (r && res.status == CORE_RPC_STATUS_OK)
+      check_rpc_cost("getblockheaderbyheight", res.credits, pre_call_credits, COST_PER_BLOCK_HEADER);
+  }
+
+  THROW_WALLET_EXCEPTION_IF(!r, tools::error::no_connection_to_daemon, "getblockheaderbyheight");
+  THROW_WALLET_EXCEPTION_IF(res.status == CORE_RPC_STATUS_BUSY, tools::error::daemon_busy, "getblockheaderbyheight");
+  THROW_WALLET_EXCEPTION_IF(res.status != CORE_RPC_STATUS_OK, tools::error::wallet_generic_rpc_error, "getblockheaderbyheight", m_trusted_daemon ? res.status : "daemon error");
+  return res.block_header;
+}
```
jeffro256, line 5950: Would you mind putting this as a method in `NodeRPCProxy` instead of here please? See #8605 if you want an example of what I mean. I'm trying to centralize all the raw RPC invocation code into one place.

hyc:
I'm confused, why is this PR, which appears to still be actively being developed, on the release branch? Development should be on master, and only finalized code should be PR'd to release branches.

selsta:
It was considered finished and we planned to include it in the last release but moneromooo had concerns with the approach used so it got redone.

j-berman:
> Development should be on master, and only finalized code should be PR'd to release branches.

Will stick to this rule on all my PR's going forward

plowsof:
When testing with restore deterministic wallet / height == 'now': passing the wallets entire in/out history (~30 txids) in a list to scan_tx i sometimes encounter this error: 
```
Error: '{'error': {'code': -1, 'message': 'deque::_M_new_elements_at_back'}, 'id': 8, 'jsonrpc': '2.0'}'
```

jeffro256:
Looks like some `std::deque` ran out of space. The error is thrown [in this file](https://gcc.gnu.org/onlinedocs/gcc-4.6.2/libstdc++/api/a00858_source.html`) at line 832.

jeffro256:
@plowsof This didn't happen to be a multisig wallet, did it?

plowsof:
@jeffro256 just a normal wallet which has 1 unspent output and 33 txids (in and out payments) example here https://paste.debian.net/1258259/ load existing wallet -> get list of txids -> attempt to scan them all. sometimes line 69 (first scan) will not complain - sometimes it will - if the 2nd scan throws the error (even though the first scan completed) it makes my balance 0 at the end. hopefully someone can reproduce, ive tried messing with ordering of txids but doesn't seem to make a difference.

j-berman:
The issue was: when re-scanning a tx at a height lower than the wallet's latest saved checkpoint `m_blockchain.offset()`, it would underflow when calling `m_blockchain.crop` -> `m_blockchain.resize(height - m_offset)`, which then throws the deque error because the underflowed value to resize to is larger than the max size of the underlying container (line 831 in that file you linked @jeffro256).

Fixed by making sure to only call `m_blockchain.resize(height - m_offset)` when `height >= m_offset`.

___

I also noticed a small off-by-1 bug where if the height of the tx to re-process was the same as the wallet's latest saved checkpoint, after re-processing, it wouldn't re-attach the lowest detached block back to the locally saved `m_blockchain`.

___

Good bugs to catch @plowsof

j-berman:
In discussion with @selsta and @plowsof, one of the common use cases for `scan_tx` today is a GUI simple mode user whose wallet missed a **recent** tx when scanning (for a wallet to miss a tx when scanning, either there is an unknown bug somewhere in the scanner, or the user was connected to a malicious node withholding txs). The user can quickly ingest the tx via `scan_tx` without needing to restore from their wallet's creation height and re-do the entire sync process. As such, `scan_tx` has been a successful support tool for users whose wallets miss **recent** txs.

Since I added the requirement that the feature be used with trusted daemons in this PR, GUI simple mode users would lose a valuable support feature.

### Proposed solution for untrusted daemons

- If the wallet is connected to an untrusted daemon:
	- has 0 txs **after** the one the user requests to scan, then `scan_tx` will continue to completion.
	- has 1+ txs **after** the one the user requests to scan, `scan_tx` will fail and the GUI will suggest the user restore their wallet instead.

This maintains `scan_tx`'s value as a support tool as it seems to be used today, and maintains the same privacy risk as the user understands it today.
______

### Additional detail

The reason I added the requirement of trusted nodes in the first place in this PR is because I decided to re-request **all** of a user's txs from the daemon **after** the one being scanned in order to re-process all txs in order. The reason I needed to request those txs from the daemon in order to re-process them is because the wallet throws away transaction data after processing a tx and saving the tx to the wallet (specifically RCT data no longer needed like the encrypted amount). Thus, in order to re-process these transactions via `process_new_transaction` to ensure all txs are re-processed in order when calling `scan_tx` on an earlier tx, I re-requested the txs from the daemon to get that data back.

Longer term, a larger re-write could seek to avoid needing to do all the re-processing inside `process_new_transaction` (obviously it's sub-optimal to need to e.g. decrypt amounts again and re-check outputs for ownership etc.). But the cost/benefit of doing such a re-write as part of this fix to `scan_tx` does not seem worth it to me.

jeffro256:
> has 0 txs after the one the user requests to scan, then scan_tx will continue to completion.

Doesn't this still give away to the daemon which transaction you're asking to scan, with the implication that you're probably involved with it (likely receiver)?

j-berman:
Yep, here's how the feature is exposed to a GUI user today:

GUI warning

The implications of exposing txs beyond just the single tx the user requests to scan are more severe and less intuitive. A user could in theory reveal their entire transaction history to a daemon just by requesting to re-scan the first tx in their wallet for example.

j-berman:
Updated to allow scanning the wallet's most recent tx when connected to an untrusted daemon.

plowsof:
Tested the new changes. Confirmed:
- Latest transaction can be scanned from an untrusted daemon
- Older transaction from untrusted displays a warning / works with trusted.    

Thank you! :clap: [here is my test script](https://gist.github.com/plowsof/c129e744e304d4e9879410d7d17238ba)

Gingeropolous:
couldn't the privacy concerns be assuaged by using a block height instead of a specific transaction hash? First, there's the basket of potential outputs in the block itself, plus you could double-down and request 10 other random blocks to scan. Or 15. So its like a ring-signature of blocks. But not really, because its not. But you know what i mean. 

Sure, it's gonna cost more in processing and bandwidth than 1 specific tx. But its a whole lot less than "hey lets just scan the whole blockchain!"

j-berman:
I think using heights instead of hashes in that way would improve the privacy profile of the feature when using an untrusted daemon. Plus it's arguably a UX improvement: as @hMihaiDavid mentioned in #7291, it's easier to write block heights on a paper backup together with a seed, versus writing hashes (though on the downside, it may be a bit harder to cleanly communicate the difference between the "restore height" and a feature to "rescan tx at just this height" in the UI).

As it relates to this PR, I would prefer to keep this PR as is since:
1) It fixes bugs with `scan_tx` via hashes today.
2) It makes sense not to remove `scan_tx` via hashes even if we were to introduce scanning via heights. Scanning via hashes is already supported in the RPC wallet.
3) I don't think this PR exposes users to *additional* privacy risk than what they would already expect in the feature today. In the [rare case](https://github.com/monero-project/monero/pull/8566#issuecomment-1294255769) where this PR *would* introduce additional privacy risk, `scan_tx` requires a trusted daemon.

I think it's worth re-considering adding the ability to scan txs via height(s) in the future, and moving away from using hashes in the GUI for future PR(s).

woodser:
Just checking on this. Anything preventing progress to merged?

src/wallet/wallet2.cpp
```diff
@@ -1651,17 +1654,187 @@ void wallet2::scan_tx(const std::vector<crypto::hash> &txids)
...
+        sorted_tx_entries.lowest_height = std::min(tx_info.block_height, sorted_tx_entries.lowest_height);
+        sorted_tx_entries.highest_height = std::max(tx_info.block_height, sorted_tx_entries.highest_height);
+      }
+    }
   }
 
+  // sort tx_entries by height from lowest height to highest; pool txs to the back
+  auto cmp_tx_entry = [](const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry& l, const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry& r)
+    { return l.block_height < r.block_height && !l.in_pool; };
+  std::sort(sorted_tx_entries.tx_entries.begin(), sorted_tx_entries.tx_entries.end(), cmp_tx_entry);
```
moneromooo-monero, line 1670: This will only sort partially, if two of the txes are in the same block (or are both in the txpool). Do you care ?

src/wallet/wallet2.cpp
```diff
@@ -1651,17 +1654,187 @@ void wallet2::scan_tx(const std::vector<crypto::hash> &txids)
...
+
+  THROW_WALLET_EXCEPTION_IF(blockchain.size() != dbd.original_chain_size,
+    error::wallet_internal_error, "Unexpected blockchain size after re-attaching");
+}
+//----------------------------------------------------------------------------------------------------
+bool has_unexpected_tx_above_height(uint64_t height, const std::unordered_set<crypto::hash> &expected_txids, const wallet2::transfer_container &transfers,
+    const wallet2::payment_container &payments, const serializable_unordered_map<crypto::hash, wallet2::confirmed_transfer_details> &confirmed_txs)
+{
+  for (const auto &td : transfers)
+    if (td.m_block_height >= height && expected_txids.find(td.m_txid) == expected_txids.end())
```
moneromooo-monero, line 1754: This has >= for a function named "above", intended ?

src/wallet/wallet2.cpp
```diff
@@ -1651,17 +1654,187 @@ void wallet2::scan_tx(const std::vector<crypto::hash> &txids)
...
+  }
+  // re-request txs from daemon to re-process with all tx data needed
+  tx_entry_data txs_to_reprocess = get_sorted_tx_entries(tx_hashes_to_reprocess);
+
+  process_scan_txs(txs_to_scan, txs_to_reprocess, tx_hashes_to_reprocess, dbd);
+  reattach_blockchain(m_blockchain, dbd);
+
+  // If the highest scan_tx height > the wallet's known scan height, then the
+  // wallet scanner should continue scanning from that tx's height and "pretend"
+  // it scanned all txs up to that height. This gurantees txs are processed
```
moneromooo-monero, line 1816: "guarantees"

src/wallet/wallet2.cpp
```diff
@@ -1651,17 +1654,187 @@ void wallet2::scan_tx(const std::vector<crypto::hash> &txids)
...
+    if (txids.find(tx_hash) == txids.end())
+      tx_hashes_to_reprocess.insert(tx_hash);
+  }
+  // re-request txs from daemon to re-process with all tx data needed
+  tx_entry_data txs_to_reprocess = get_sorted_tx_entries(tx_hashes_to_reprocess);
+
+  process_scan_txs(txs_to_scan, txs_to_reprocess, tx_hashes_to_reprocess, dbd);
+  reattach_blockchain(m_blockchain, dbd);
+
+  // If the highest scan_tx height > the wallet's known scan height, then the
```
iamamyth, line 1814: I think this comment mostly, but not entirely, conveys the rationale for skipping forward in the blockchain. I suggest:
```
If the highest scan_tx height exceeds the wallet's known scan height, then the wallet should skip ahead to the scan_tx's height in order to service the request in a timely manner. Skipping unrequested transactions avoids generating sequences of calls to process_new_transaction which process transactions out-of-order, relative to their order in the blockchain, as the process_new_transaction implementation requires transactions to be processed in blockchain order. 
```

tests/functional_tests/transfer.py
```diff
@@ -829,6 +834,218 @@ def check_is_key_image_spent(self):
...
+        assert res.amount == amount
+        assert res.fee > 0
+        fee = res.fee
+
+        expected_sender_balance = sender_starting_balance - (amount + fee)
+        expected_receiver_balance = receiver_starting_balance + amount
+
+        test = 'Checking scan_tx on outgoing pool tx'
+        for attempt in range(2): # test re-scanning
+            print(test + ' (attempt ' + str(attempt+1) + ')')
```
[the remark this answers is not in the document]
j-berman: Clarified this print statement in the latest. I put a print statement here on purpose because the first attempt is testing that `scan_tx` works on first attempt, then the second attempt is testing that re-scanning works, so the print statement helps clarify it's testing different things

tests/functional_tests/transfer.py
```diff
@@ -829,6 +834,218 @@ def check_is_key_image_spent(self):
...
+            assert tx.amount == amount
+            assert tx.fee == fee
+            assert len(tx.destinations) == 1
+            assert tx.destinations[0].amount == amount
+            assert tx.destinations[0].address == dst['address']
+            assert sender_wallet.get_balance().balance == expected_sender_balance
+
+        test = 'Checking scan_tx on incoming pool tx'
+        for attempt in range(2): # test re-scanning
+            print(test + ' (attempt ' + str(attempt+1) + ')')
```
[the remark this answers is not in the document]
j-berman: Same as above

src/wallet/wallet2.cpp
j-berman: I don't see any code that relies on processing txs in the order they appear in a block (or in the pool), just by "height order." So this should suffice

src/wallet/wallet2.cpp
j-berman: Renamed to `has_unexpected_tx_at_height_or_above`

src/wallet/wallet2.cpp
j-berman: Updated as per [this suggestion](https://github.com/monero-project/monero/pull/8566#discussion_r1118161316)

src/wallet/wallet2.cpp
j-berman: Done

src/wallet/wallet2.cpp
moneromooo-monero: It could screw up cold signing if the m_transfers vector is ordered differently between hot and cold wallets (inputs are refered to by index IIRC). Possibly same for multisig.
All theoretical though. But having m_transfers shuffle itself a bit seems like the kind of bad idea that comes back to bite us later. I suppose it's nothing that can't be fixed later if it does come back though so not a show stopper I suppose.

j-berman: You're right, I missed that. Re-ordering txs like this can cause re-importing outputs from cold wallet to hot wallet to fail (I haven't checked pool/multisig behavior).

This is unfortunate. I don't see a nice way to handle this here without the daemon returning additional data that enables the client to sort txs in the same block (or client re-fetches entire blocks). Checking `output_indices` seems it would only work for RCT txs, not pre-RCT. Do you see a nice way to handle this?

j-berman: Multisig has the same issue, you were correct there too (`get_multisig_composite_key_image` relies on `m_transfers` to be in order).

For a 2/2 multisig, assume wallet A has scanned up to chain tip normally, and wallet B restores at a height > restore height and then uses `scan_tx([tx2, tx1])` to scan txs. If tx2 and tx1 are in the same block and wallet B calls `scan_tx` with the txs out of order, then wallet B will be borked.

Only way to prevent wallet B from getting borked here is for both wallets to order the txs in a consistent way, which goes back to my point above:

> I don't see a nice way to handle this here without the daemon returning additional data that enables the client to sort txs in the same block (or client re-fetches entire blocks)

Curious if you've got ideas here

j-berman: > so getting the blocks they're in does not leak any more information, and it should not be too much extra data.

Done in the latest

moneromooo-monero (approved):
Lots of renaming and otherwise spammy changes in that new patch, but the core seems to do what it says, so approved given I was fine with the previous one except the sorting.

plowsof:
sanity checked the behaviour of adding missing "latest" / "oldest" / scanned multiple tx's - and 'it works' after squashing still :)

selsta (approved):

PR #8571, opened 2022-09-16
Keys: Update for expired 'rbrunner7' key, valid 2 years

rbrunner7:
Commit should be signed / verifyable with this very key.


jeffro256 (approved):
I can verify that the commit is signed by the @rbrunner7's old key with fingerprint `9F8C0A7E9CD85B2C522B9F306A251D571F7F7BC`. @rbrunner7 also confirmed this PR over Matrix.

PR #8576, opened 2022-09-20
Second thread pool for I/O

SChernykh:
monerod currently uses a single thread pool for everything, including DNS requests. If it runs on a low-end CPU (less than 8 threads), DNS requests can block incoming transactions handling entirely. This change fixes it by never mixing compute and I/O tasks. I/O thread pool has a fixed size of 8 threads because monerod never does more than 7 parallel DNS requests as far as I can see in the code. Also, 8 I/O threads will process all DNS requests in parallel on low-end CPUs whereas they were processed consecutively before.


moneromooo-monero (approved):

jeffro256 (approved):
Ideally, parallel I/O tasks would be handled using an asynchronous I/O executor, but this separation is definitely a step in the correct direction.

PR #8580, opened 2022-09-21
Add Gitian builds to README

duggavo:


moneromooo-monero (approved):
Would be even better if it mentioned these are the reproducible builds.

jeffro256 (approved):

Issue #8583, opened 2022-09-21
wrong version


budmaster13:
Daemon wont connect to GUI says using wrong version but i have latest version GUI version: 0.18.1.1-unknown (Qt 5.15.6)
Embedded Monero version: 0.18.1.1-unknown
Wallet path: /home//Monero/wallets
Wallet restore height: 2686804
Wallet log path: /home/.bitmonero/monero-wallet-gui.log
Wallet mode: Simple mode (bootstrap)
Graphics mode: OpenGL
I have used this wallet several times after fork with no issues. Now its saying im using wrong version.

selsta:
Can you go to Settings -> Log and post the output of "status" ?

selsta:
Simple mode connects to a random remote node. If you connect to a remote node that is outdated (using v0.17) it will now say "wrong version". If you wait a bit it _should_ connect to a new node.

budmaster13:
it wont let me drag and drop it here

selsta:
You should be able to copy paste it.

budmaster13:
yea tried that as well. Said something went wrong and it wouldnt let me.

selsta:
Can you make a screenshot?

budmaster13:
Sent from Proton Mail mobile

https://www.reddit.com/r/Monero/comments/xk9ayu/monero_gui_problems/ipd4zr4?utm_medium=android_app&utm_source=share&context=3 I wrote down the status output in this reddit thread
-------- Original Message --------

selsta:
It seems your daemon fails to start for some reason. Can you just set a custom remote node instead? I can give you instructions for it.

budmaster13:
Its a bootstrap simple wallet. The blockchain is downloaded to my hard drive. Even when i just try to connect to remote node same problem i have with local node. Wrong version

selsta:
Which remote node did you connect to? And did you switch to advanced mode for that?

budmaster13:
No i hit the icon in bottom left corner where it says switch to another public node when you hover your curser over it.

budmaster13:
Here is a picture of wallet

Sent from Proton Mail mobile

-------- Original Message --------

budmaster13:
I sent you a picture with status on my cell phone to your email

selsta:
The easiest way to get things working is to follow the steps in this comment to connect to a remote node: https://github.com/monero-project/monero-gui/issues/3989#issuecomment-1214412781

[...]

PR #8585, opened 2022-09-22
wallet2: fail to establish daemon cxn == "Disconnected" cxn status [release]

j-berman:
When the wallet fails to establish a connection to a daemon either because the daemon is non-existent or offline, sometimes the wallet will incorrectly think the failure is caused by incompatible versions. Reported by @hinto-janaiyo [here](https://github.com/monero-project/monero/pull/8544#discussion_r974825557) (also the cause of the "Wrong version" status in #8583).

The primary cause of the bug is undefined behavior introduced in #8544 when initializing `wallet_is_outdated`.

Separate from the above bug: in this PR, I also reduced `wallet.cpp`'s default connection timeout from 30s to 20s, which matches wallet2 and helps avoid an infinite "Connecting" loop in the GUI.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -4787,7 +4787,7 @@ bool simple_wallet::try_connect_to_daemon(bool silent, uint32_t* version)
   uint32_t version_ = 0;
   if (!version)
     version = &version_;
-  bool wallet_is_outdated, daemon_is_outdated = false;
+  bool wallet_is_outdated = false, daemon_is_outdated = false;
   if (!m_wallet->check_connection(version, NULL, 200000, &wallet_is_outdated, &daemon_is_outdated))
   {
     if (!silent)
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -63,8 +63,8 @@ namespace {
     static const int    MAX_REFRESH_INTERVAL_MILLIS = 1000 * 60 * 1;
     // Default refresh interval when connected to remote node
     static const int    DEFAULT_REMOTE_NODE_REFRESH_INTERVAL_MILLIS = 1000 * 10;
-    // Connection timeout 30 sec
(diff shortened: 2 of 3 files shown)


SChernykh (approved):

hinto-janaiyo (approved):
Tested, these show correct messages now:
- v0.18 wallet -> v0.17 daemon
- v0.18 wallet -> offline daemon

Didn't know it was just missing `= false`, I thought that was just valid syntax :D

This is probably a different issue but the actual interval values seem to be multiplied by 1.5x:

- Connection timeout takes 30s not 20s
- Refresh interval takes 15s not 10s

The hangs after each message funnily enough also lead to the wallet locking itself for inactivity anywhere from 30 seconds to immediately after the initial user prompt

selsta:
@hinto-janaiyo did you test the CLI wallet or the GUI?

hinto-janaiyo:
@selsta Above was CLI, just tested GUI with https://github.com/monero-project/monero-gui/pull/4036 + this PR, it show the correct error messages and the time interval is also increased. The wallet lock doesn't really apply I guess since the GUI default is 10min vs 1:30min in CLI, and you can still interact in the GUI while it's connecting

jeffro256:
For those who want to reproduce the bad behavior, where in this PR @j-berman added ` = false`, change it to ` = true`. C++ initialization is terrible :(. I wonder if there is something suppressing the uninitialized warning?

j-berman:
Thank you for helping test @hinto-janaiyo, much appreciated.

@jeffro256 I get the warning if I try to read the variable in the line after it's declared, so seems it's something to do with how the variable is used in `connect` -- generally has the feel of a compiler bug to me. Also unfortunate that my system seems to default initialize the value to false (in order to repro [this comment](https://github.com/monero-project/monero/pull/8544#discussion_r974825557), I either had to explicitly initialize as true like you suggested, or call `set_daemon` multiple times; I specifically tested all those cases when I worked on it).

PR #8593, opened 2022-09-24
workflows: ubuntu 18.04 is deprecated

selsta:
And will soon be removed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,7 @@ jobs:
       CCACHE_TEMPDIR: /tmp/.ccache-temp
     strategy:
       matrix:
-        os: [ubuntu-latest, ubuntu-18.04]
+        os: [ubuntu-22.04, ubuntu-20.04]
     steps:
     - uses: actions/checkout@v1
       with:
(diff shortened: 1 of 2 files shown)


hyc:
Every reference to bionic in contrib/gitian will need to be updated as well. And possibly some cross-compiler versions will need to be bumped up. This affects pretty much tne entire contents of that directory: *.yml, *.py, *.sh.

selsta:
@hyc sorry, I meant the Ubuntu 18.04 VM image on Github Actions is deprecated and will be removed, so the workflows file is the only thing we have to change at the moment.

jeffro256:
> Workflows using the `ubuntu-18.04` YAML workflow label should be updated to `ubuntu-20.04`, `ubuntu-22.04`, or `ubuntu-latest`

Shouldn't keeping the `ubuntu-latest` actions stay valid?

selsta:
Yes, but ubuntu-latest is currently 20.04 and that would mean we test the same OS twice.

plowsof:
Starting October 3rd - ubuntu-18.04 runners will randomly fail , approved

plowsof (approved):

SChernykh (approved):

PR #8569, opened 2022-09-15
levin_abstract_invoke: remove dead code

jeffro256:

PR #8598, opened 2022-09-26
Remove invoke function in levin handler

vtnerd:
The invoke function in the levin handler is currently unused. It is probably best to remove it because:
  * using it and `async_invoke` at the same time is going to be problematic on the ordering of responses
  * It blocks a thread waiting for a response, and requires another thread to process the response

#8569 removes related code that is also unused.

### DIFF
--- a/contrib/epee/include/net/levin_protocol_handler_async.h
+++ b/contrib/epee/include/net/levin_protocol_handler_async.h
@@ -102,7 +102,6 @@ public:
   uint64_t m_max_packet_size;
   uint64_t m_invoke_timeout;
 
-  int invoke(int command, message_writer in_msg, std::string& buff_out, boost::uuids::uuid connection_id);
   template<class callback_t>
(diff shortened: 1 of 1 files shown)


jeffro256 (approved):

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -521,21 +510,8 @@ class async_protocol_handler
...
-              {
-                CRITICAL_REGION_BEGIN(m_local_inv_buff_lock);
-                m_local_inv_buff = std::string((const char*)buff_to_invoke.data(), buff_to_invoke.size());
-                buff_to_invoke = epee::span<const uint8_t>((const uint8_t*)NULL, 0);
-                m_invoke_result_code = m_current_head.m_return_code;
-                CRITICAL_REGION_END();
-                m_invoke_buf_ready = true;
-              }
+              MERROR("Received levin response but have no invoke handlers");
+              return false;
```
SChernykh, line 514: Does this ever trigger when running monerod on mainnet or testnet?

vtnerd: I've never seen this message, this would indicate some error elsewhere that needs correcting.

PR #8602, opened 2022-10-01
workflows: build gitian on tag

selsta:
- only runs on tag
- each OS uses a separate runner
- hashes get printed on summary page
- the resulting build archives get uploaded as artifacts

### DIFF
--- a/.github/workflows/gitian.yml
+++ b/.github/workflows/gitian.yml
@@ -1,0 +1,49 @@
+name: ci/gh-actions/gitian
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  build-gitian:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        operating-system:
+          - name: "Linux"
+            option: "l"
+          - name: "Windows"
+            option: "w"
+          - name: "Android"
+            option: "a"
+          - name: "FreeBSD"
+            option: "f"
+          - name: "macOS"
+            option: "m"
+    name: ${{ matrix.operating-system.name }}
+    steps:
+    - name: prepare
+      run: |
+        sudo apt update
+        curl -O https://raw.githubusercontent.com/monero-project/monero/${{ github.ref_name }}/contrib/gitian/gitian-build.py
+        chmod +x gitian-build.py
+    - name: setup
+      run: |
+        ./gitian-build.py --setup --docker github-actions ${{ github.ref_name }}
+    - name: build
+      run: |
+        ./gitian-build.py --docker --detach-sign --no-commit --build -j 3 -o ${{ matrix.operating-system.option }} github-actions ${{ github.ref_name }}
+    - name: post build
+      run: |
+        cd out/${{ github.ref_name }}
+        shasum -a256 *
+        echo \`\`\` >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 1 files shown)


.github/workflows/gitian.yml
```diff
@@ -0,0 +1,46 @@
...
+    name: ${{ matrix.operating-system.name }}
+    steps:
+    - name: prepare gitian build host 
+      run: |
+        sudo usermod -aG docker $USER
+    - name: build
+      run: |
+        curl -O https://raw.githubusercontent.com/monero-project/monero/${{ github.ref_name }}/contrib/gitian/gitian-build.py
+        chmod +x gitian-build.py
+        sudo ./gitian-build.py --setup --docker github-actions ${{ github.ref_name }}
```
hyc, line 34: Why does this need a sudo if you've added $USER to the docker group?

selsta: Not sure, I got a permission error without it.

.github/workflows/gitian.yml
```diff
@@ -0,0 +1,46 @@
...
+            option: "a"
+          - name: "FreeBSD"
+            option: "f"
+          - name: "macOS"
+            option: "m"
+    name: ${{ matrix.operating-system.name }}
+    steps:
+    - name: prepare gitian build host 
+      run: |
+        sudo usermod -aG docker $USER
```
Malinero, line 29: this seems unnecessary as $USER is already a member of group docker

.github/workflows/gitian.yml
selsta: Updated and removed sudo.

.github/workflows/gitian.yml
selsta: Removed.

Malinero (approved):

plowsof (approved):

.github/workflows/gitian.yml
```diff
@@ -0,0 +1,44 @@
+name: ci/gh-actions/gitian
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  build-gitian:
+    runs-on: ubuntu-20.04
```
SChernykh, line 10: https://github.com/monero-project/monero/blob/master/contrib/gitian/README.md says
> Gitian host OS should be Ubuntu 18.04 "Bionic Beaver". If you are on a mac or windows for example, you can run it in a VM but will be slower.

I know that Github is going to deprecate 18.04 soon, but changing OS to 20.04 will probably create different binaries.

selsta: I did confirm that the hashes match.

SChernykh: The we can probably change the README as well

hyc: Gitian creates its own container, built on Ubuntu 18.04 regardless. Thus the binaries will always be the same. The difference in host OS can affect the format of the .assert files, which are generated by host tools, but that won't matter if nobody is expecting to sign these builds.

SChernykh (approved):

selsta:
Updated to add `sudo apt update` to avoid a potential file not found error message during `apt install` later during the gitian process. Also split the code up in more steps.

plowsof (approved):

SChernykh (approved):

PR #8605, opened 2022-10-05
wallet2: move /gettransactions invocations to node_rpc_proxy

jeffro256:
In the future, I plan to wrap all RPC invocations into node_rpc_proxy, those PRs can come later, but this will definitely be one of the bigger ones. This is not just refactoring for the sake of refactoring, however. I plan to implement a "community node" feature for wallet2, which would allow direct connections to different secured daemons without a dependency on starting a local p2p process.

In order to do this, we need one streamlined RPC interface for wallet2. Even if this "community node" feature does not come to fruition, abstracting away the RPC invocation details into a different class will allow us to easily do multi-threaded RPC requests, more responsive public node requests, etc, while the interface remains extremely simple. Also, if (when) the big wallet3 redesign comes around, this will make transitioning easier.

### DIFF
--- a/src/wallet/node_rpc_proxy.cpp
+++ b/src/wallet/node_rpc_proxy.cpp
@@ -56,10 +56,8 @@
  * @return std::string when cond is false, otherwise does not return
  */
 #define RETURN_ERROR_IF_FALSE(cond, msg)    \
-  do                                        \
-  {                                         \
-    if (!(cond))                            \
-    {                                       \
+  do {                                      \
+    if (!(cond)) {                          \
       std::stringstream ss;                 \
       ss << msg;                            \
       const std::string err_msg = ss.str(); \
@@ -69,7 +67,7 @@
   } while (0);                              \
 
 /**
- * @brief Acquire RPC mutex, invoke /json_rpc endpoint, and return on error
+ * @brief Acquire RPC mutex and invoke /json_rpc endpoint
  *
  * Only really useful within NodeRPCProxy because it assumes the existence of variables m_daemon_rpc_mutex,
  * m_http_client, and rpc_timeout. The lock_guard is scoped so that it lives as short as possible.
@@ -80,20 +78,23 @@
  *
  * @return std::string on RPC error, otherwise does not return
  */
-#define INVOKE_JSON_RPC_LOCK_AND_RET_ERR(method, req, res)                                                  \
-  do {                                                                                                      \
-    bool r = false;                                                                                         \
-    {                                                                                                       \
-      const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};                             \
-      r = epee::net_utils::invoke_http_json_rpc("/json_rpc", method, req, res, m_http_client, rpc_timeout); \
-    }                                                                                                       \
-    RETURN_ON_RPC_RESPONSE_ERROR(r, epee::json_rpc::error{}, resp_t, method);                               \
-  } while (0);                                                                                              \
+#define NRP_INVOKE_JSON_RPC(method, req, res)                                   \
+  do {                                                                          \
+    bool r = false;                                                             \
+    {                                                                           \
+      const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex}; \
+      r = epee::net_utils::invoke_http_json_rpc("/json_rpc", method, req, res,  \
+        m_http_client, rpc_timeout);                                            \
+    }                                                                           \
+    RETURN_ON_RPC_RESPONSE_ERROR(r, epee::json_rpc::error{}, resp_t, method);   \
+  } while (0);                                                                  \
 
 /**
- * @brief Acquire RPC mutex, invoke /json_rpc endpoint, check RPC cost, and return on error
+ * @brief Set client signature, acquire RPC mutex, invoke /json_rpc endpoint, and check RPC cost
  *
- * Same as INVOKE_JSON_RPC_LOCK_AND_RET_ERR but checks that the credits changed correctly.
+ * Only really useful within NodeRPCProxy because it assumes the existence of variables
+ * m_daemon_rpc_mutex, m_http_client, m_client_id_secret_key, and rpc_timeout. The lock_guard is
+ * scoped so that it lives as short as possible.
  *
  * @param method JSON RPC method name as a string
  * @param req cryptonote::COMMAND_RPC_x::request form which will be used with invoke_http_json_rpc
@@ -103,34 +104,35 @@
  * @return std::string on RPC error, otherwise does not return
  */
 
-#define INVOKE_JSON_RPC_LOCK_AND_RET_ERR_CHECK_COST(method, req, res, rpc_price)                                 \
-  do {                                                                                                           \
-    req.client = cryptonote::make_rpc_payment_signature(m_client_id_secret_key);                                 \
-    {                                                                                                            \
-      const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};                                  \
-      const uint64_t pre_call_credits = m_rpc_payment_state.credits;                                             \
-      bool r = epee::net_utils::invoke_http_json_rpc("/json_rpc", method, req, res, m_http_client, rpc_timeout); \
-      RETURN_ON_RPC_RESPONSE_ERROR(r, epee::json_rpc::error{}, resp_t, method);                                  \
-      check_rpc_cost(m_rpc_payment_state, method, res.credits, pre_call_credits, rpc_price);                     \
-    }                                                                                                            \
-  } while (0);                                                                                                   \
+#define NRP_INVOKE_JSON_RPC_WITH_PRICE(method, req, res, rpc_price)                          \
+  do {                                                                                       \
+    req.client = cryptonote::make_rpc_payment_signature(m_client_id_secret_key);             \
+    {                                                                                        \
+      const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};              \
+      const uint64_t pre_call_credits = m_rpc_payment_state.credits;                         \
+      const bool r = epee::net_utils::invoke_http_json_rpc("/json_rpc", method, req, res,    \
+        m_http_client, rpc_timeout);                                                         \
+      RETURN_ON_RPC_RESPONSE_ERROR(r, epee::json_rpc::error{}, resp_t, method);              \
+      check_rpc_cost(m_rpc_payment_state, method, res.credits, pre_call_credits, rpc_price); \
+    }                                                                                        \
+  } while (0);                                                                               \
 
 /**
- * @brief Acquire RPC mutex, invoke json endpoint, check RPC cost, and return on error
+ * @brief Set client signature, acquire RPC mutex, invoke json endpoint, and check RPC cost
  *
- * Only really useful within NodeRPCProxy because it assumes the existence of variables m_daemon_rpc_mutex,
- * m_http_client, and rpc_timeout. The lock_guard is scoped so that it lives as short as possible.
- * After RPC interaction, we check that the credits changed correctly.
+ * Only really useful within NodeRPCProxy because it assumes the existence of variables
+ * m_daemon_rpc_mutex, m_http_client, m_client_id_secret_key, and rpc_timeout. The lock_guard is
+ * scoped so that it lives as short as possible.
  *
  * @param uri URI of json endpoint
  * @param req cryptonote::COMMAND_RPC_x::request form which will be used with invoke_http_json_rpc
  * @param res cryptonote::COMMAND_RPC_x::response form which will be used with invoke_http_json_rpc
  * @param rpc_price RPC credit cost for given method, usually specified in macros named COST_PER_x
  *
- * @return on RPC error, otherwise does not return
+ * @return std::string on RPC error, otherwise does not return
  */
 
-#define INVOKE_JSON_LOCK_AND_RET_ERR_CHECK_COST(uri, req, res, rpc_price)                    \
+#define NRP_INVOKE_JSON_WITH_PRICE(uri, req, res, rpc_price)                                 \
   do {                                                                                       \
     req.client = cryptonote::make_rpc_payment_signature(m_client_id_secret_key);             \
     {                                                                                        \
@@ -247,7 +249,7 @@ boost::optional<std::string> NodeRPCProxy::get_rpc_version(uint32_t &rpc_version
     const time_t now = time(NULL);
     cryptonote::COMMAND_RPC_GET_VERSION::request req_t = AUTO_VAL_INIT(req_t);
     cryptonote::COMMAND_RPC_GET_VERSION::response resp_t = AUTO_VAL_INIT(resp_t);
-    INVOKE_JSON_RPC_LOCK_AND_RET_ERR("get_version", req_t, resp_t);
+    NRP_INVOKE_JSON_RPC("get_version", req_t, resp_t);
 
     m_rpc_version = resp_t.version;
     m_daemon_hard_forks.clear();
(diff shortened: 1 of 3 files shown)


vtnerd:
> In the future, I plan to wrap all RPC invocations into node_rpc_proxy

This does "break-up" the code into smaller cpp files, but I don't see much of a benefit outside of that. Given the size of `wallet2.cpp`, the refactor might be useful.

Is your goal to move the references to the http client and locks into `node_proxy`, and then modify the interface to be exportable to other projects or something? It seems funky given that the RPC is supposed to be the interface to the outside world, but this could be a generic set of functions that are commonly used for tx scanning (I guess).

> I plan to implement a "community node" feature for wallet2, which would allow direct connections to different secured daemons without a dependency on starting a local p2p process.

This mostly already exists, and I fail to see how this refactor is related to that goal.

> In order to do this, we need one streamlined RPC interface for wallet2. Even if this "community node" feature does not come to fruition, abstracting away the RPC invocation details into a different class will allow us to easily do multi-threaded RPC requests, more responsive public node requests, etc, while the interface remains extremely simple. Also, if (when) the big wallet3 redesign comes around, this will make transitioning easier.

I'm not certain what can be gained from multi-threaded requests in `wallet2` - where I'm assuming you mean multiple active requests to `monerod` at a time? The bulk of work in the past has been on interleaving requests with processing - the current code blocks a thread during the request, and is now "dispatched" to a background thread while block processing is queued up in a thread pool.

Having multi-threaded requests adds some more load to `monerod`, but it is not obvious that it will provide better throughput to wallet processing. I would recommend running some numbers with a rough mock-up before modifying this node_proxy that much.

jeffro256:
> Is your goal to move the references to the http client and locks into node_proxy, and then modify the interface to be exportable to other projects or something?

Yeah, basically. `wallet2` only really uses a small subset of all available RPC endpoints, and as it is, `m_node_rpc_proxy`, `m_http_client`, and `m_daemon_rpc_mutex` are a leaky abstraction which very much limits the flexibility on how `wallet2` can connect to daemon(s) (w/o a large overhaul and tricky concurrency).

> This mostly already exists, and I fail to see how this refactor is related to that goal.

Eh I disagree, with the main point here being "without a dependency on starting a local p2p process" part of that sentence. Currently, since `wallet2` only has the notion of connecting to 1 daemon, the simple mode in GUI is achieved by spawning a local `monerod` process in no-sync, bootstrap mode. This sucks from a UX point of view since starting/stopping `monerod` can take anywhere from 5-10 seconds in some cases, and `monerod` often gets deleted by antivirus because of the miner. This refactor will begin the process of centralizing code in `NodeRPCProxy` so that the wallet itself can manage its own connections without a bootstrap daemon. 

> I'm not certain what can be gained from multi-threaded requests in wallet2 - where I'm assuming you mean multiple active requests to monerod at a time?

That could certainly be an option, but I mostly meant multi-threaded requests across *different* daemons which is not remotely possible with the current way that `wallet2` uses RPC. To be clear, this PR does nothing new, it is just a refactoring PR, but it is refactoring towards making a certain goal possible. If we split the RPC request details from the "wallet-y" stuff into 2 seperate layers, it opens up a lot of flexibility for determining how the data is fetched underneath. Also it makes the code cleaner.

> I would recommend running some numbers with a rough mock-up before modifying this node_proxy that much.

The multi-threaded aspect is not the primary goal of this PR, but just an example of a secondary opportunity. The main point will be to eventually have `NodeRPCProxy` have the ability to manage multiple connections (maybe multi-threaded, probably just connection switching) to community nodes while keeping `wallet2` code simple and ignorant of the RPC details.

vtnerd:
> Yeah, basically. wallet2 only really uses a small subset of all available RPC endpoints, and as it is, m_node_rpc_proxy, m_http_client, and m_daemon_rpc_mutex are a leaky abstraction which very much limits the flexibility on how wallet2 can connect to daemon(s) (w/o a large overhaul and tricky concurrency).

It's really just more of cleaning up the copy/pasted code imo. And the HTTP client could just have been made thread-safe default, which would've dropped some of this.

> Eh I disagree, with the main point here being "without a dependency on starting a local p2p process" part of that sentence. Currently, since wallet2 only has the notion of connecting to 1 daemon, the simple mode in GUI is achieved by spawning a local monerod process in no-sync, bootstrap mode. This sucks from a UX point of view since starting/stopping monerod can take anywhere from 5-10 seconds in some cases, and monerod often gets deleted by antivirus because of the miner. This refactor will begin the process of centralizing code in NodeRPCProxy so that the wallet itself can manage its own connections without a bootstrap daemon.

The GUI does this to find and locate community run RPC nodes via the Monero p2p network. How is this patch going to improve that? `Wallet2` can already take a custom argument for a specific node, and the GUI team should be capable of easily making that an option. The problem of finding available nodes - in a decentralized manner - is not going to be improved with these refactors afaik. These changes could result in `wallet2` speaking directly to monero p2p, but that seems painful. A fixed list of community nodes could already be provided and used with `wallet2` and downstream projects.

> That could certainly be an option, but I mostly meant multi-threaded requests across different daemons which is not remotely possible with the current way that wallet2 uses RPC. To be clear, this PR does nothing new, it is just a refactoring PR, but it is refactoring towards making a certain goal possible. If we split the RPC request details from the "wallet-y" stuff into 2 seperate layers, it opens up a lot of flexibility for determining how the data is fetched underneath. Also it makes the code cleaner.

Yes this might improve things a bit. I'm just wary because the last attempts have resulted in unnecessary API layers to `wallet2`.

> The multi-threaded aspect is not the primary goal of this PR, but just an example of a secondary opportunity. The main point will be to eventually have NodeRPCProxy have the ability to manage multiple connections (maybe multi-threaded, probably just connection switching) to community nodes while keeping wallet2 code simple and ignorant of the RPC details.

Why have connections to multiple nodes? This just increases the load on the public RPC network, and technically makes the code a little more complex because the 2+ nodes could have temporary differences in chain tip. I'm not sure if I've seen any crypto project attempt this sort of thing, although it's definitely possible. Or is one like a backup node or something? There are many situations/goals where this doesn't seem worth the complexity.

jeffro256:
Before I say anything else, I'll define what I meany by "community nodes". This is something that has been discussed a certain amount on Matrix/IRC, and is part of my goals for my funded CCS proposal, but it's not necessarily super well known knowledge yet. Currently, simple mode in GUI is susceptible to attacks where public RPC nodes provide bad information, with the most pervasive cases being extremely high (think >20 XMR) fee estimation responses. Certain sanity checks can be put into place to help mitigate against some cases of RPC abuse, but guarding against this is a fundamentally hard problem because wallets in simple mode specifically choose to not sync block data. In addition, having users' wallets in simple mode run requests to random public nodes opens up certain privacy implications.

Because of this, it has become a popular idea to not use the public p2p RPC list to choose remote nodes from in simple GUI mode, but to instead use a hardcoded set of vetted nodes ran by more-or-less reputable members of the monero community, secured by explicit, hardcoded SSL certificates, so we can skip crawling the p2p network for nodes and we don't need `monerod`. This is what I mean by "community" nodes. 

> It's really just more of cleaning up the copy/pasted code imo. And the HTTP client could just have been made thread-safe default, which would've dropped some of this.

Yeah that's a lot of what this PR is: replacing hundreds of copy-pasted boilerplate RPC lines in `wallet2` with just  a handful of calls to `NodeRPCProxy`. 

> The GUI does this to find and locate community run RPC nodes via the Monero p2p network. How is this patch going to improve that?

If you have a fixed set of known nodes (or at least fixed set of "seed"-like or DNS-like nodes) before connecting (like in the proposed "community" node model), in the future we would just connect to and manage those connections directly from `NodeRPCProxy`, without a `monerod` process. We can also do slightly more complex connection schemes, like ranking daemons and switching daemon's if our current one isn't working. 

> Wallet2 can already take a custom argument for a specific node, and the GUI team should be capable of easily making that an option.

Yes, but that's the thing: it can take an argument *for a specific node*. This is due do a limitation of `wallet2` born out of the complexity and non-modular nature of how it handles RPC requests. If we logically split the request-y stuff and the wallet-y stuff, we can have more powerful connection managements, or like you mentioned, a "backup node" system could be implemented. 

> The problem of finding available nodes - in a decentralized manner - is not going to be improved with these refactors afaik.

True

> Yes this might improve things a bit. I'm just wary because the last attempts have resulted in unnecessary API layers to wallet2

Yes, that's a very good point. This is why I pushed things in this PR to an already existing class: `NodeRPCProxy`. `NodeRPCProxy` already acts as a layer between `wallet2` and RPC requests for some RPC methods, but not all. This PR would just be a continuation of fleshing out that layer. 

> Why have connections to multiple nodes? This just increases the load on the public RPC network, and technically makes the code a little more complex because the 2+ nodes could have temporary differences in chain tip.

1) Splitting RPC requests between nodes could theoretically improves loads or allow for more active load balancing  2) Splitting RPC requests across nodes while scanning could improve anonymity by making it harder to track which wallets are at which point in the sync process. The difference of chain tips could be a problem and I guess that would just have to an issue we tackle at that point in time, but in the case of the "community" node system, since we're at least making the assumption that we're not getting utterly malicious data, we could adopt a policy of "longest chain is the right one" and that could be "close enough". 

Thanks for engaging btw!

vtnerd:
> If you have a fixed set of known nodes (or at least fixed set of "seed"-like or DNS-like nodes) before connecting (like in the proposed "community" node model), in the future we would just connect to and manage those connections directly from NodeRPCProxy, without a monerod process. We can also do slightly more complex connection schemes, like ranking daemons and switching daemon's if our current one isn't working.

My questions were more focused on how `NodeRPCProxy` helps with some of the goals you mentioned. I'm almost knee-jerk reaction against when someone suggests API something or other in `wallet2` because it hasn't worked out for the better in the past (`wallet_api` and the hardware wallet stuff being the best examples). I think you have indirectly answered my questions in this last response.

> Yes, but that's the thing: it can take an argument for a specific node. This is due do a limitation of wallet2 born out of the complexity and non-modular nature of how it handles RPC requests. If we logically split the request-y stuff and the wallet-y stuff, we can have more powerful connection managements, or like you mentioned, a "backup node" system could be implemented.

The HTTP client already supports a changing the remote server, and `wallet2` can acquire the lock and call the associated function to make a server change. After reading these posts, I think you meant that `NodeRPCProxy` will have the code necessary to handle changes in the blockchain state when switching nodes? Because changing RPC servers isn't like a fail-over to a fully-synced MySQL backup node - so if that type of logic can be moved into one reusable class with a sane interface, then its certainly helpful.

> Splitting RPC requests between nodes could theoretically 1) improve loads or allow for more active load balancing 2) Splitting RPC requests across nodes while scanning could improve anonymity by making it harder to track which wallets are at which point in the sync process. The difference of chain tips could be a problem and I guess that would just have to an issue we tackle at that point in time, but in the case of the "community" node system, since we're at least making the assumption that we're not getting utterly malicious data, we could adopt a policy of "longest chain is the right one" and that could be "close enough".

Two nodes can be 100% legit and compliant, have the same block height, and still have a temporary difference in the top block hash (or two). Network propagation isn't instantaneous. I've never seen a load balance attempt like this before in a cryptocurrency wallet - although there are tons of wallets everywhere so I'm sure its been done. I don't think 2+ simultaneous connections are worth the effort - at least not until its been shown to be helpful somehow - and instead it should be more useful to: (1) write a good algorithm to detect a node under load and switch, (2) make sure a switch can be done "safely" given that node states could be different and, (3) provide a good random selection algorithm for initial attempts. Both should be needed even if 2+ connections are allowed anyway, so might as well focus on that approach first.

jeffro256:
> The HTTP client already supports a changing the remote server, and `wallet2` can acquire the lock and call the associated function to make a server change. After reading these posts, I think you meant that `NodeRPCProxy` will have the code necessary to handle changes in the blockchain state when switching nodes? Because changing RPC servers isn't like a fail-over to a fully-synced MySQL backup node - so if that type of logic can be moved into one reusable class with a sane interface, then its certainly helpful.

Okay, I think this is what we're really going back and forth about: I am going to (pending review) put the node-(switching/connecting) logic into `NodeRPCProxy` so that `wallet2` uses the same interface whether its connecting to a local daemon or several possible remote demons. Right now `wallet2` does that by having an extremely heavy dependency on `monerod`. Yes, I could implement this feature directly into `wallet2`, but I think it become unruly very quickly. I guess it comes down to a design choice, but packaging connection boilerplate into nicely into `NodeRPCProxy` would make this feature easily to implement IMO and less error-pone (specifically lock holding).

jeffro256:
Also for those who want to review, I would recommend taking this path: 1) review node_rpc_proxy.h and see if you like the interface 2) review node_rpc_proxy.cpp to see if you like the way it's implemented 3) look into wallet2.cpp and check that `get_transaction(s)` is called correctly to match the old behavior.

vtnerd:
> Okay, I think this is what we're really going back and forth about: I am going to (pending review) put the node-(switching/connecting) logic into NodeRPCProxy so that wallet2 uses the same interface whether its connecting to a local daemon or several possible remote demons. Right now wallet2 in simple mode does that by having an extremely heavy dependency on monerod. Yes, I could implement this feature directly into wallet2, but I think it become unruly very quickly. I guess it comes down to a design choice, but packaging connection boilerplate into nicely into NodeRPCProxy would make this feature easily to implement IMO and less error-pone (specifically lock holding).

No, I am currently fine with moving stuff into `NodeRPCProxy`, but I do not think connecting to multiple remote daemons simultaneously is a good idea (at least as an initial goal). It should be a bit more complex to implement correctly than it seems at first, and it will put a higher load on community run RPC servers.

Refactoring `wallet2` in this PR did wake me up initially because it hasn't been done well historically.

jeffro256:
Oh sorry, I just gave simultaneous connections as an example of something that would be incredibly difficult to do correctly in `wallet2` now without a lot of really tricky refactors and boilerplate since `wallet2` manages its own `m_daemon_rpc_mutex`

 What I actually plan on doing is similar to what `bootstrap_node` does now: ranking a list of potential connections, punishing connections if they are slow or respond with bad data, and switching the single connection accordingly. This is still messy to do without `NodeRPCProxy` as `m_daemon_rpc_mutex` needs to be acquired, then some central connecting logic needs to be called, then `m_http_client` may be switched out, and then the lock is released. To add this boilerplate to every RPC call would make the `wallet2` code over-verbose and error-prone. If we centralize this logic in `NodeRPCProxy`, the results would be much better IMO. 

> Refactoring wallet2 in this PR did wake me up initially because it hasn't been done well historically.

Do you think that the general idea behind this refactor is sound?

src/wallet/node_rpc_proxy.cpp
```diff
@@ -46,7 +47,154 @@
...
+ * This function can be used on full AND pruned transaction entries, but is mostly used for pruned transaction entries.
+ * Takes tx version into account when parsing and validating. Also checks that actual hash values match with expected hash values.
+ * However, with v1 transactions, the hash check is skipped for pruned transactions since there is not a way to validate it statically.
+ *
+ * @param[in] entry Transaction entry response from RPC server with pruned=true, decode_as_json=false
+ * @param[out] tx the result of parsing the transaction entry
+ * @param[out] tx_hash the result of hashing the pruned transaction
+ * @return true on success, false otherwise
+ */
+bool get_pruned_tx(const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry &entry, cryptonote::transaction &tx, crypto::hash &tx_hash)
```
jeffro256, line 161: Moved line-for-line from `wallet2.cpp`

src/wallet/wallet2.cpp
```diff
@@ -1626,42 +1589,33 @@ std::string wallet2::get_subaddress_label(const cryptonote::subaddress_index& in
...
-    req.decode_as_json = false;
-    req.prune = true;
-
-    size_t ntxes = slice + SLICE_SIZE > txids.size() ? txids.size() - slice : SLICE_SIZE;
-    for (size_t i = slice; i < slice + ntxes; ++i)
-     req.txs_hashes.push_back(epee::string_tools::pod_to_hex(txids[i]));
+  // Setup a vector containing tuples of corresponding txs, tx entries, and tx hashes
+  using txq_value_type = std::tuple<cryptonote::transaction, cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry, crypto::hash>;
+  std::vector<txq_value_type> txq;
+  txq.reserve(txids.size());
```
jeffro256, line 1595: For the changes in `scan_tx`, as well as updating to a `get_transactions()` call in place of the old RPC code, I changed the transaction priority queue container into a vector and then sorted at the end. This provides the same behavior since processing isn't intermingled with collecting, but should speed up `scan_tx` since the whole queue is only sorted once.

selsta:
I think this will conflict with https://github.com/monero-project/monero/pull/8566

jeffro256:
I am going to close this refactor in favor of something better.

PR #8607, opened 2022-10-09
Fix light wallet login

Malinero:
According to https://github.com/monero-project/meta/blob/master/api/lightwallet_rest.md#login

Tested with monero-lws, requires https://github.com/vtnerd/monero-lws/pull/53

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -9468,14 +9468,13 @@ bool wallet2::light_wallet_login(bool &new_address)
   request.view_key = string_tools::pod_to_hex(get_account().get_keys().m_view_secret_key);
   // Always create account if it doesn't exist.
   request.create_account = true;
+  request.generated_locally = true;
   m_daemon_rpc_mutex.lock();
-  bool connected = invoke_http_json("/login", request, response, rpc_timeout, "POST");
+  m_light_wallet_connected = invoke_http_json("/login", request, response, rpc_timeout, "POST");
   m_daemon_rpc_mutex.unlock();
   // MyMonero doesn't send any status message. OpenMonero does. 
-  m_light_wallet_connected  = connected && (response.status.empty() || response.status == "success");
+  //m_light_wallet_connected  = connected && (response.status.empty() || response.status == "success");
   new_address = response.new_address;
-  MDEBUG("Status: " << response.status);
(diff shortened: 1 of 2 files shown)


selsta:
Does the wallet2 light wallet code even work? Quite sure it hasn't been touched in years.

Malinero:
> Does the wallet2 light wallet code even work after login?

Nope

selsta:
@Malinero do you plan on getting it working?

Malinero:
> @Malinero do you plan on getting it working?

Yes, and its integration into the GUI

selsta:
perfect :)

src/wallet/wallet2.cpp
```diff
@@ -9468,14 +9468,13 @@ bool wallet2::light_wallet_login(bool &new_address)
...
   // Always create account if it doesn't exist.
   request.create_account = true;
+  request.generated_locally = true;
   m_daemon_rpc_mutex.lock();
-  bool connected = invoke_http_json("/login", request, response, rpc_timeout, "POST");
+  m_light_wallet_connected = invoke_http_json("/login", request, response, rpc_timeout, "POST");
   m_daemon_rpc_mutex.unlock();
   // MyMonero doesn't send any status message. OpenMonero does. 
-  m_light_wallet_connected  = connected && (response.status.empty() || response.status == "success");
+  //m_light_wallet_connected  = connected && (response.status.empty() || response.status == "success");
```
selsta, line 9476: Should this be removed completely?

SChernykh: This removes the check for `success` response status though, is it what was intended? As it's written right now, it'll consider the wallet to be "connected" even if http requests returns 404.

Malinero: The `success` field has been removed from the specs https://github.com/monero-project/meta/blob/master/api/lightwallet_rest.md#login

I left the comment for now because I don't know if newer versions of OpenMonero doest support it, and even if it does, would the community want to support a feature specific to a single implementation and is not documented in the standard spec ?

Malinero: > it'll consider the wallet to be "connected" even if http request returns 404

True, should add a check on response code

selsta:
Closing as we removed the light wallet code.

Issue #8614, opened 2022-10-18
creating a view only wallet on 18.1.2 always returns an error


anycolo:
E offset >= num_outputs. THROW EXCEPTION: error::wallet_internal_error
	E Error creating view only wallet: Offset is larger than total outputs

j-berman:
I see one way to trigger this error -- does the wallet that you're trying to create a view only wallet have 0 txs (is it a totally new wallet)?

j-berman:
Here are the errors I'm seeing in the GUI. When creating a view-only wallet from:

- A wallet with 0 transfers triggers: `Offset is larger than total outputs`
- A wallet with >1 transfer(s) triggers: `Imported outputs omit more outputs that we know of. Try using export_outputs all.`

I see the causes of the above two errors. Will PR fixes shortly but want to confirm this is what you're seeing as well.

anycolo:
Yes!!!

anycolo:
I've been having other issues as well:
https://github.com/monero-project/monero/issues/8613
Even if it returns an error, the gui still creates a wallet, which seems usable. I have used that wallet. Was it safe to use, or will i lose funds?

j-berman:
I don't see how you could lose funds. The issue of the view-only wallet displaying too high of a balance in #8613 seems entirely to do with not importing key images into the view-only wallet as mentioned in the comments there.

PR #8616, opened 2022-10-18
wallet: fix create view-only wallet from existing wallet [master]

j-berman:
A user can create a view-only wallet from a normal wallet by calling `createWatchOnly()`. The GUI supports this feature. If the user calls this function from:

### 1) A normal wallet with 0 transfers

The result of `export_outputs()` will have a 0 offset and 0 `num_outputs`, which triggers this error in `import_outputs` (reported in #8614):

https://github.com/monero-project/monero/blob/fc907a957078cae2dc68348886a33363848dd089/src/wallet/wallet2.cpp#L13415-L13416

**Fix:** I updated the error logic of the throw.

### 2) a normal wallet with >1 transfer(s)

Since all key images are known in a normal wallet, `export_outputs()` won't export any of the user's transfers into the fresh view-only wallet; the offset will advance past all of the user's transfers when exporting:

https://github.com/monero-project/monero/blob/fc907a957078cae2dc68348886a33363848dd089/src/wallet/wallet2.cpp#L13264-L13267

Then when importing, the view-only wallet won't know of any existing outputs prior to the offset, which triggers this error in `import_outputs`:

https://github.com/monero-project/monero/blob/fc907a957078cae2dc68348886a33363848dd089/src/wallet/wallet2.cpp#L13412-L13413

**Fix:** Call `export_outputs(all=true)` so the normal wallet will export **all** outputs into the fresh view-only wallet. I also added `export_key_images(all=true)` to make sure to get all key images imported as well.

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -535,7 +535,7 @@ bool WalletImpl::createWatchOnly(const std::string &path, const std::string &pas
         view_wallet->generate(path, password, address, viewkey);
 
         // Export/Import outputs
-        auto outputs = m_wallet->export_outputs();
+        auto outputs = m_wallet->export_outputs(true/*all*/);
         view_wallet->import_outputs(outputs);
 
         // Copy scanned blockchain
@@ -553,7 +553,7 @@ bool WalletImpl::createWatchOnly(const std::string &path, const std::string &pas
 
         // Export/Import key images
         // We already know the spent status from the outputs we exported, thus no need to check them again
-        auto key_images = m_wallet->export_key_images();
+        auto key_images = m_wallet->export_key_images(true/*all*/);
         uint64_t spent = 0;
         uint64_t unspent = 0;
         view_wallet->import_key_images(key_images.second, key_images.first, spent, unspent, false);
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -13330,9 +13330,7 @@ size_t wallet2::import_outputs(const std::tuple<uint64_t, uint64_t, std::vector<
   THROW_WALLET_EXCEPTION_IF(offset > m_transfers.size(), error::wallet_internal_error,
       "Imported outputs omit more outputs that we know of");
 
-  THROW_WALLET_EXCEPTION_IF(offset >= num_outputs, error::wallet_internal_error,
-      "Offset is larger than total outputs");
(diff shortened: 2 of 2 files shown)


plowsof:
sanity check: exact same errors seen with 0 txns wallet / >1 txns wallet

With this PR , view-only wallets are created (in both scenarios) without error, Thank you!

anycolo:
When will this be released?

selsta:
It will be included in the next release, but we don't have a date yet.

anycolo:
Can i use an older monero gui release? Does console monero cli have this issue as well?

selsta:
@anycolo You can manually restore a view only wallet:

Main menu -> Restore wallet from keys -> Select restore wallet from keys -> Enter public key and view key, leave spend key blank.

This will work without issues using the current GUI version.

anycolo:
I need to plug the view wallet intro a btcpayserver, and i have some issues with it.

selsta:
What issues do you have? The steps I explained in https://github.com/monero-project/monero/pull/8616#issuecomment-1295152048 will create a fully functioning view only wallet.

plowsof:
BTCpay provides no support/instructions for Monero integration, however, they do link to a guide from one of our community members at [sethforprivacy.com](https://sethforprivacy.com/guides/accepting-monero-via-btcpay-server/) , if you have any issues following this guide please contact the author.

anycolo:
I used the btcpay integration.
I generated a wallet and a view wallet. The view wallet generated an error after generation, but i used it anyway. I used the monero rpc connection of the btcpayserver to spend money, and it showed a larger amount than the desktop wallet, even though this wasn't happening before, probably because btcpayserver was using the wallets differently or because of the view wallet errors when it was generated. I don't understand why it needs a normal and a view only wallet anyway. I'm trying now to do what selsta instructed. I hope i haven't lost any funds.
Now my desktop monero wallet gui isn't syncing using a remote node that i own, and it takes 10 minutes to close it.. it's very tiresome.

anycolo:
> @anycolo You can manually restore a view only wallet:
> 
> Main menu -> Restore wallet from keys -> Select restore wallet from keys -> Enter public key and view key, leave spend key blank.
> 
> This will work without issues using the current GUI version.
In that menu, i have account address (public), view key (private) and spend key(private), but not public key as you said.

anycolo:
I have created a new view wallet as instructed, but it still shows a larger balance than the regular wallet.

selsta (approved):

PR #8617, opened 2022-10-19
wallet: background sync with just the view key [release]

j-berman:
See #8619


src/wallet/api/wallet.cpp
```diff
@@ -54,6 +54,40 @@ using namespace cryptonote;
...
+    { \
+        setStatusError(tr("HW wallet cannot use background sync")); \
+        return false; \
+    } \
+    if (m_wallet->watch_only()) \
+    { \
+        setStatusError(tr("View only wallet cannot use background sync")); \
+        return false; \
+    } \
+    if (m_wallet->multisig()) \
```
rbrunner7, line 84: The *master* PR uses `get_multisig_status().multisig_is_active` instead here which is ok

src/wallet/wallet2.cpp
```diff
@@ -6243,60 +6623,76 @@ void wallet2::load(const std::string& wallet_, const epee::wipeable_string& pass
...
+{
+  if (m_wallet_file.empty())
+    return;
+  if (m_background_syncing || m_is_background_wallet)
+    return;
+  if (m_background_sync_type == BackgroundSyncOff)
+    return;
 
-  if (!m_persistent_rpc_client_id)
-    set_rpc_client_secret_key(rct::rct2sk(rct::skGen()));
```
rbrunner7, line 6248: These 2 lines are not in *master*, that's why the master PR diff does not show them as deleted, which is ok

src/wallet/wallet2.h
```diff
@@ -792,6 +811,54 @@ namespace tools
...
+        VARINT_FIELD(block_timestamp)
+        FIELD(double_spend_seen)
+      END_SERIALIZE()
+    };
+
+    struct background_sync_data_t
+    {
+      bool first_refresh_done = false;
+      uint64_t start_height = 0;
+      serializable_unordered_map<crypto::hash, background_synced_tx_t> txs;
```
rbrunner7, line 842: *master* PR can use `std::unordered_map` here thanks to improvements in the serializer

rbrunner7 (approved):
I systematically compared the changes in this PR with the changes in the corresponding [PR to the master branch](https://github.com/monero-project/monero/pull/8619). I commented on the very few differences that all have "natural" explanations because of differences between *master* and *release* unrelated to this PR, and the differences do not lead to *any* change in functionality.

I can therefore approve this PR.

MrCyjaneK:
will this get released as a part of v0.18.3.4?

selsta:
@MrCyjaneK It is currently only merged to master so that it can get tested for a bit.

PR #8620, opened 2022-10-21
functional_tests: Give monero-wallet-rpc instances log files

jeffro256:
Makes it a lot easier to debug functional test failures on the wallet side after the fact


j-berman (approved):

moneromooo-monero:
They already have log files... I use them all the time.

jeffro256:
@moneromooo-monero Really? Where are they stored at?

jeffro256:
Nevermind, I found them :)

PR #8627, opened 2022-10-25
Fix dandelion++ fluff/stem bug with local txes

vtnerd:
An IRC user reported a Dandelion++ bug - the `monerod` Dandelion++ implementation does not match the paper exactly. When a user submits their transaction to their `monerod` for broadcasting, if their local daemon is in "fluff" mode their transaction must still be broadcast in "stem" mode. Our current implementation "fluffs" in this situation; now the implementation will "stem" in this situation.

The fallout from this is a potential targeted attack on a user - an attacker tries to fill up a users incoming table, thereby increasing the chances of knowing whether a particular transaction is received over "stem" (or not). If an attacker fills the incoming table, a fluffed transaction either originated at the victim node or 1+ outgoing links from the victim are in fluff mode. The attacker would then have to do _additional_ trace analysis on the victim's outgoing connections to determine who initiated the fluffing. The randomized delays in fluffing should make this difficult (compared to pre-dandelion++ implementation).

This IRC user also pointed out that following the paper has a similar issue - if an attacker can determine whether a node is in fluff vs stem, and get the victim to make an outgoing connection to them, there's a chance they can identify with high probability that the victim node is the origin for the transaction. However, this should more difficult to perform, assuming it is difficult to spam the peerlist aggressively (which is a whole other topic).

Users should upgrade immediately at next point release, but also not panic because mass surveillance should still be difficult.

### DIFF
--- a/src/cryptonote_protocol/levin_notify.cpp
+++ b/src/cryptonote_protocol/levin_notify.cpp
@@ -542,6 +542,7 @@ namespace levin
       i_core_events* core_;
       std::vector<blobdata> txs_;
       boost::uuids::uuid source_;
+      relay_method tx_relay;
 
       //! \pre Called in `zone_->strand`
       void operator()()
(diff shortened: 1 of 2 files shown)


j-berman:
For reference, a relevant snippet from the paper:

> (2) Transaction Forwarding (own): Every time a node generates a transaction of its own, it forwards
the transaction, in stem phase, along the same outbound edge in the anonymity graph.

Source (pg. 29:5): https://arxiv.org/pdf/1805.11060.pdf

Good spot by the IRC user.

j-berman (approved):

jeffro256 (approved):

PR #8637, opened 2022-11-14
http_client: Add support for virtual servers

jeffro256:
With this patch, we can make HTTP requests where the value of the `Host:` field is different from the real IP address. I plan on using this feature for community root node load balancing, where the root node doesn't have to be a real `monerod` process but can be a lightweight shared server which serves a JSON response,  directing wallets to other real `monerod` servers.


selsta:
@jeffro256 any way to test this?

Issue #8520, opened 2022-08-20
monerod: incoming peers unreasonably low


computerfreak94:
Running monero mainnet, stagenet, testnet and wownero mainnet on the same hardware and IP address.

Since months i have a very low count of `in` peers (5 at most and sometimes 0 over many days).
It started on testnet and stagenet long before they forked and since the recent hardfork mainnet is affected too.

Sometimes the `in` peers suddenly go up to 50-150 (mainnet) and are back close to zero a few hours later.

Changing IP address doesn't help and switching `monerod` off for a couple of days doesn't help either.

`wownerod` seems to be unaffected so i don't think it's related to my hardware or internet and `Monero CLI` was downloaded and hash verified from the official website.

I watched the peerlist size behavior on `wownerod` and it looks healthy how the greylist decreases and the whitelist increases until a new chunk is added to the greylist but with `monerod` the amount of peers in both lists seems to stay almost the same which could indicate a problem with greylist housekeeping in `monerod` but i don't know how to investigate further.

selsta:
Since which version do you have this issue?

I run multiple nodes and so far did not notice this behavior.

computerfreak94:
> Since which version do you have this issue?

I don't know anymore which version.

I wrote [this post ](https://www.reddit.com/r/Monero/comments/r9ib3j/open_your_ports/) 9 months ago and i already had the problem at least 4 months earlier.
It could be even earlier because i was not checking peerlist sizes back then.

I got inspired to finally write this bug report because [it seems i might not be alone anymore](https://www.reddit.com/r/Monero/comments/wsd6tc/monerod_connection_problem/).

computerfreak94:
Wait, what i wrote in the post 9 months ago is not the same problem...

computerfreak94:
9 months ago apparently i had the problem to find outgoing peers, now i have a problem to find incoming peers.

selsta:
Which OS do you have?

computerfreak94:
> Which OS do you have? 4 of my Linux nodes are okay, my macOS node can reproduce the issue.

Ubuntu 22.04

selsta:
When did this issue started showing up for you?

selsta:
Can you run your daemon with log level 2 for a couple minutes and then upload logs? Just trying to confirm if you have the same error.

computerfreak94:
> When did this issue started showing up for you?

[...]

PR #8640, opened 2022-11-16
p2p: fix back ping to discover healthy peers to connect to

j-berman:
### Overview

When Node A receives an incoming connection from Node B, Node A will attempt to "back ping" Node B to see if Node B would be a healthy peer to connect to in the future. #8426 introduced a bug that causes Node A to fail to back ping Node B: Node A attempts to read Node B's response over an SSL stream, which errors out (looks like because Node A does not establish an SSL connection with Node B ... [see more on this here](https://github.com/monero-project/monero/issues/8520#issuecomment-1312308716)).

The back ping seems to be [a critical component of the p2p protocol for getting incoming connections](https://github.com/monero-project/monero/issues/8520#issuecomment-1310975634) (also see [this comment](https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/src/p2p/p2p_protocol_defs.h#L277-L283)). As such, I think fixing the back ping in this PR should help resolve or alleviate the lack of incoming peers issue reported in #8520 once enough nodes on the network start running it. To be clear, if you run this PR with your node, this PR wouldn't help *you* get incoming connections directly -- the nodes your node connects to must *also* run this update in order to help *you* get incoming connections.

### The fix

The fix matches `connect_async`'s logic to [this synchronous call to connect()](https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/src/p2p/net_node.inl#L3140-L3142), ensuring that a node attempting to back ping its incoming connection will use the expected SSL setting (as the code is structured today, this specific `m_ssl_support` [appears to always be disabled](https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/src/p2p/net_node.inl#L952) since SSL isn't supported for p2p connections).

### How to manually test

Run 2 nodes, pointing one to the other. Start Node A normally with `--log-level 2` and ensure that Node A can receive incoming connections. Start Node B with the flag `--add-priority-node <Node A>` and also make sure that Node B can receive incoming connections.

Once Node B initiates the connection to Node A, make sure that Node A successfully back pings Node B. In order to make sure this happens, grep Node A's logs for "PING SUCCESS" and you should see Node B's <IP:port number>.

If you try repeating the above test without this PR, there should be no "PING SUCCESS" in the logs. A lack of "PING SUCCESS" indicates that Node A does not add Node B to its white list, which means that neither Node A nor the rest of the network will know that Node B is a healthy peer to connect to right off the bat.

Note: I plan on writing an explicit test for this behavior to ensure we don't end up breaking back ping behavior again in the future.

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -2413,7 +2413,7 @@ namespace nodetool
         return false;
       }
       return true;
-    });
+    }, "0.0.0.0", m_ssl_support);
     if(!r)
     {
       LOG_WARNING_CC(context, "Failed to call connect_async, network error.");


j-berman:
@selsta and I just ran the following test:

1. I connected to a seed node @selsta has access to on testnet using the `--add-priority-node` flag. Neither of us were running this patch on our nodes. I confirmed that my node connected to @selsta's node and to other nodes fine, but after 10 minutes, I had 0 incoming connections.

2. @selsta updated the seed node using this patch.

3. I re-connected to @selsta's seed node. The back ping from the seed node to my node succeeded and I immediately started getting incoming connections.

moneromooo-monero:
This implies that autodetecting SSL does not work, right ? If so, this seems to be the real problem, and disabling SSL just bodges it.

j-berman:
I don't think it would make sense to call `connect_async` with SSL support set to autodetect here anyway because we don't use SSL for p2p. I figure disabling SSL support in this call to `connect_async` is a justified contained fix that prevents this particular back ping bug.

While I think a fix for autodetect in `connect_async` is probably also warranted, I don't think it being broken has any practical ramifications. I don't see where we should expect to make an outgoing connection with SSL support using the TCP server connection, but maybe I'm missing something there. If I run `monerod --rpc-ssl=autodetect` and `monero-wallet-rpc --daemon-ssl=autodetect --rpc-ssl=autodetect`, I have no issues even when connecting to the respective SSL-enabled backends, autodetect seems to work as expected.

____

To fix autodetect in `connect_async`, it seems we'd need to implement this same handshake sequence the synchronous `connect()` implements before calling `start` on the connection:

Synchronous `connect()`:

https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/contrib/epee/include/net/abstract_tcp_server2.inl#L1752-L1769

`try_connect():`

https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/contrib/epee/include/net/abstract_tcp_server2.inl#L1636-L1657

I believe the fact that `connect_async` doesn't do the above handshake sequence implies SSL is broken via `connect_async`. It doesn't look like mishandling autodetect was the primary issue here; it looks like it was just lucky that the back ping worked before the connection patch.

____

Circling back to my original point: I think this PR is still justified as is toward preventing the back ping because SSL should be disabled for p2p connections. I think it's worth a separate PR to make sure `connect_async` handles SSL correctly.

j-berman:
Clarifying: when a wallet initiates an outbound connection to a daemon, SSL autodetect is handled correctly in this totally separate area of the code:

https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/contrib/epee/include/net/net_helper.h#L158-L248

I don't see how a client would ever initiate an outbound SSL connection **using the `boosted_tcp_server` connection** anywhere. It seems the reason there is what looks like unused SSL code for **a server** initiating the outbound `connect` (and `connect_async`) is that the initial PR to add SSL support originally supported SSL for p2p, which was then removed: https://github.com/monero-project/monero/pull/4054#issuecomment-453301372. It looks like code that was left in there by accident. Which would mean fixing `connect_async` to correctly handle autodetect (and SSL) would just be fixing unused code.

And circling back again: regardless, I'm not seeing why `connect_async` should be called here in this p2p handshake with SSL autodetect as is. I still think this PR is a correct solution in its own right

vtnerd (approved):

vtnerd:
And FWIW I read the backlog of discussion here, and had my cup of coffee.

The `connect_async` function still needs to be fixed for autodetect. This small patch is better for now because its simpler, and we want this behavior even with a fixed `connect_async`. Defaulting to autodetect is "noisy", has wasted RTT, etc.

vtnerd (approved):
Ugh. Wrong PR.

j-berman:
@vtnerd agree with the above :) Will submit a new maintenance PR with SSL fixed in `connect_async`+ parameter positions flipped

moneromooo-monero:
Sounds good.

jeffro256:
Does a node ever attempt to back-ping on a loopback address?

j-berman:
@jeffro256

https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/src/p2p/net_node.inl#L2352-L2353

https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/src/p2p/net_peerlist.h#L265-L269

jeffro256:
@j-berman That's exactly what I was looking for, thank you! I was looking for the reason why the back-ping test wasn't working locally.

jeffro256 (approved):
I was able to confirm this patch fixed the node so that back pings work 👍🏻

PR #8641, opened 2022-11-16
p2p: fix back ping to discover healthy peers to connect to [release]

j-berman:
See #8640


vtnerd (approved):
Don't worry about the parameter position flipping suggesting from the other branch, not as important in the release branch.

jeffro256 (approved):

PR #8642, opened 2022-11-16
Fix static builds for Ubuntu 22.04

woodser:
Link libbsd, libmd, libprotokit for ZMQ to fix static builds for Ubuntu 22.04

From @LocalMonero


selsta (approved):

jeffro256 (approved):
Diff matches #8564

PR #8648, opened 2022-11-21
DNSResolver: fix not handling hostnames without dot characters

jeffro256:
Unrelated, but similar code-wise to #8643. There is a check in `DNSResolver` which automatically fails to resolve hostnames which do not contain the `.` character. This PR removes that check.


selsta:
Do we know why this check was added in the first place?

selsta (approved):

PR #8650, opened 2022-11-22
blockchain_export: add checks for args to `blocks.dat` format

jeffro256:
Add 2 checks which help give feedback to user on bad arguments instead of silently doing the wrong thing.
1. The `blocks.dat` format must start from height 0. If the user also gives a `--block-start` option, throw an error.
2. The `blocks.dat` format chunks the data every `HASH_OF_HASHES_STEP` (512) blocks. If the user does not align `--block-stop` with this paramter, align the block stop and give them a warning message.

Before both of these arguments were silently ignored if incorrect, e.g. running with `--blocksdat` and `--block-stop 510` exported no data. Also, running with `--blocksdat`, `--block-start 509`, `--block-stop 511` was the same as if you changed the option to `--block-start 0`.

### DIFF
--- a/src/blockchain_utilities/blockchain_export.cpp
+++ b/src/blockchain_utilities/blockchain_export.cpp
@@ -113,6 +113,7 @@ int main(int argc, char* argv[])
     return 1;
   }
   bool opt_blocks_dat = command_line::get_arg(vm, arg_blocks_dat);
+  CHECK_AND_ASSERT_MES(!(opt_blocks_dat && block_start), 1, "Can't specify both --block-start and --blocksdat");
 
   std::string m_config_folder;
 
--- a/src/blockchain_utilities/blocksdat_file.cpp
+++ b/src/blockchain_utilities/blocksdat_file.cpp
@@ -137,10 +137,14 @@ bool BlocksdatFile::store_blockchain_raw(Blockchain* _blockchain_storage, tx_mem
   uint64_t progress_interval = 100;
   block b;
 
-  uint64_t block_start = 0;
+  // Check requested_block_stop for validity
+  const bool has_explicit_block_stop = (requested_block_stop > 0) && (requested_block_stop < m_blockchain_storage->get_current_blockchain_height());
+  const bool is_hash_hash_aligned = (requested_block_stop + 1) % HASH_OF_HASHES_STEP == 0; // + 1 since inclusive range
+  CHECK_AND_ASSERT_MES(is_hash_hash_aligned || !has_explicit_block_stop, false, "Number of blocks (block-stop incl.) must be a multiple of " << HASH_OF_HASHES_STEP);
+
   uint64_t block_stop = 0;
   MINFO("source blockchain height: " <<  m_blockchain_storage->get_current_blockchain_height()-1);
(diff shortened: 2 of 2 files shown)


jeffro256:
I'd like some feedback on this. Is this the correct thing to do long-term or should we make it so it includes those mod 512 block hashes in the final result?

hyc:
re: (1) The blocks.dat format must start from height 0. If the user also gives a --block-start option, throw an error.

No, wrong.

Please read the commit history of the code you're modifying before modifying anything.

hyc:
for (2) prob better to just print a warning and proceed with the rounded off values.

jeffro256:
> re: (1) The blocks.dat format must start from height 0. If the user also gives a --block-start option, throw an error.
> 
> No, wrong.
> 
> Please read the commit history of the code you're modifying before modifying anything.

Please, explain then. If you're talking about [this commit](b7dd8349f470633788f33e776480c66d19ecc8ba), that doesn't affect the `blocks.dat` format, just the bootstrap format. The `block_start` variable, derived from the `--block-start` CLI option, is not even included in the call to `store_blockchain_raw` as you can see [here](https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/src/blockchain_utilities/blockchain_export.cpp#L180) and [here](https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/src/blockchain_utilities/blocksdat_file.cpp#L133). In the [core `for` loop](https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/src/blockchain_utilities/blocksdat_file.cpp#L159), `m_cur_height` is initialized to the local variable `block_start` (not to be confused with a variable of the same name in blockchain_export.cpp), which is always `0`.

hyc:
my comment on (1) was a mistake and I already deleted it.

jeffro256:
@hyc Now it gives a warning message instead of throwing an error

selsta:
I usually generate the `blocks.dat` file for releases and I never did any alignment?

hyc:
Probably because for releases, we just use the entire blockchain, so don't need start/stop arguments. And that means the last chunk has always been silently rounded off before.

jeffro256:
> I usually generate the `blocks.dat` file for releases and I never did any alignment? What consequences has this?

@selsta Whatever you set the block stop at for the release checkpoints, there are some blocks which are not included in the hash. For example, if you set the `--block-stop` to be 1000, then you only get the hashes for heights 0-511 (inclusive). If you set the `--block-stop` to be 1001, you still only get the hashes for heights 0-511. 

> Probably because for releases, we just use the entire blockchain, so don't need start/stop arguments.

@hyc I'm pretty sure the entire blockchain is not used; there are very specific stop points which are agreed upon beforehand.

hyc:
Still, irrelevant. Even if the actual stop is different from what was specified, the difference will always be consistent.

jeffro256:
Though it will be consistent with the same `--block-stop`, it shouldn't be consistent from one `--block-stop` to another, IMO, or the user should get an error.

PR #8653, opened 2022-11-28
workflows: set a public DNS when doing tests

selsta:
To avoid any weird Github Actions related DNS issues.


plowsof:
solves koe's issue

plowsof (approved):

UkoeHB (approved):
Thanks (this fixes an issue with the `AddressFromURL` unit test).

PR #8658, opened 2022-11-29
fix a tiny typo (I think)

Kuplynx:
It's more common for "be" to be after "just"


selsta:
"could be just" shows more results on Google than "could just be".

See for example https://tatoeba.org/en/sentences/show/2953038

hyc:
Fwiw the wording in this patch is what I'm most used to, but at the same time I don't think the distinction is important enough to warrant a patch. I personally would ignore all this.

Kuplynx:
Yeah, it's what I'm used to and when I saw the message in the daemon I immediately said "oh, that looks wrong"

hyc:
Except it isn't clearly wrong. What's clear is that a majority of people speak/write with that pattern. Should close this ticket.

Kuplynx:
Alright

PR #8659, opened 2022-12-01
performance_tests: framework updates

UkoeHB:
This is a PR in my 'upstreaming `seraphis_lib` project', the changes here are not used anywhere yet.

- Allow custom parameter packs for tests. This enables changing parameters programmatically and reduces the number of test templates you need to compile.
- Improve error reporting for failed tests.

### DIFF
--- a/tests/performance_tests/main.cpp
+++ b/tests/performance_tests/main.cpp
@@ -101,12 +101,14 @@ int main(int argc, char** argv)
 
   const std::string filter = tools::glob_to_regex(command_line::get_arg(vm, arg_filter));
   const std::string timings_database = command_line::get_arg(vm, arg_timings_database);
-  Params p;
+  Params core_params;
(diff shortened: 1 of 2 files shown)


moneromooo-monero:
I do not see the point of this patch. Could you expand ?
At a push, the return value change might help in reproducing a failure quicker (if there was a per loop random seed, which I do not think there is atm), but the rest seems gratuitous. What am I missing ?

UkoeHB:
@moneromooo-monero I needed these changes to support the big seraphis perf tests:
https://github.com/UkoeHB/monero/blob/seraphis_lib/tests/performance_tests/seraphis_tx.h
https://github.com/UkoeHB/monero/blob/31a518f7db7dca3150acf4e3427b535e1d1943f8/tests/performance_tests/main.cpp#L247

j-berman (approved):

PR #8660, opened 2022-12-01
common: add variant class

UkoeHB:

- Add a `tools::variant` class with a useful interface.


jeffro256:
I've gotta ask: what's the point of having a possible `boost::blank` underlying type for the variant class? That just seems like it would add a lot of edge cases where they aren't needed.

jeffro256:
@UkoeHB Thanks for the fast response! I immediately realized right after I posted that that it was kind of a dumb question since I could easily think of some uses for it and deleted it lol, but you were too fast

src/common/variant.h
```diff
@@ -0,0 +1,135 @@
...
+        return type_index_of_T;
+    }
+
+    /// check if two variants have the same type
+    static bool same_type(const variant<Types...> &v1, const variant<Types...> &v2)
+    { return v1.type_index() == v2.type_index(); }
+
+    /// apply a visitor to the variant
+    template <typename VisitorT>
+    typename VisitorT::result_type visit(const VisitorT &visitor) const
```
vtnerd, line 124: I would `&& visitor` with `std::forward<VisitorT>(visitor)` this too since `boost::apply_visitor` is doing this anyway.

vtnerd (changes requested):

src/common/variant.h
```diff
@@ -0,0 +1,135 @@
...
+    /// get the type index of a requested type (compile error for invalid types)
+    template <typename T>
+    static int type_index_of()
+    {
+        static const int type_index_of_T{VType{T{}}.which()};  //todo: this is slower than it needs to be
+        return type_index_of_T;
+    }
+
+    /// check if two variants have the same type
+    static bool same_type(const variant<Types...> &v1, const variant<Types...> &v2)
```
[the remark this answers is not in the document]
UkoeHB: Static matches the semantics - it's a function called on two variants, not a behavior of one variant.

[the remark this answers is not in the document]
vtnerd: I forgot about ADL briefly, are you invoking this like:
```c++
same_type(val1, val2);
```
? This should work, provided no one else adds a `same_type` function to the namespace. This could be a standalone function - I don't see anything wrong with that either.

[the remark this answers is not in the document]
vtnerd: Nevermind, I would just leave as-is unless it becomes an issue. Might be confusing for the person to change though.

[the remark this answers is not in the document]
UkoeHB: [This](https://github.com/UkoeHB/monero/blob/6e7b72ccc3c576f2deff936f1dd4f085a9064798/src/multisig/multisig_signing_helper_utils.cpp#L789) is the only place it's used right now.

src/common/variant.h
```diff
@@ -0,0 +1,135 @@
...
+
+public:
+//constructors
+    /// default constructor
+    variant() = default;
+    variant(boost::none_t) : variant{} {}  //act like boost::optional
+
+    /// construct from variant type
+    template <typename T>
+    variant(const T &value) : m_value{value} {}
```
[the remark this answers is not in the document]
UkoeHB: My goal is a variant with assignable value semantics, so I think this shouldn't be `explicit`.

src/common/variant.h
```diff
@@ -0,0 +1,135 @@
...
+        throw std::runtime_error("variant: tried to visit an empty variant.");
+        static const ResultT dummy{};
+        return dummy;
+    }
+};
+
+template <typename... Types>
+class variant final
+{
+    using VType = boost::variant<boost::blank, Types...>;
```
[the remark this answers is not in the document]
UkoeHB: `boost::blank` is designed specifically for this use-case afaik. Plus, a variant isn't a pointer class so I think `nullptr` is a little inconsistent.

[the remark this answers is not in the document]
vtnerd: I would continue to use `boost::blank`. An overload with `nullptr_t` (not needed now), could definitely be useful in generic code like in my serialization library. There are situations where the concepts between pointer, optional, and variant are nearly identical - assign this thing a "dead" value - but with three different techniques for assignment, requires 3 independent functions until C++17 `constexpr if` (which also confuses typical C++ devs).

Anyway, this is why I always thought `std::nullptr_t`/`nullptr` has always seemed just as appropriate though. `std::optional` has `std::nullopt_t` and `nullopt`, and the only difference is slightly more accurate type-checking but little else. Yes, `std::optional` is not _literally_ a pointer, but the intended objectives are identical and why get so fancy with unique types (because it makes writing template/generic functions _more_ complicated).

jeffro256:
Question: Do you ever plan on supporting recursive variants?

UkoeHB:
A variant of variants? Not sure where or why you'd want to use that.

jeffro256:
They’re common for representing deserialized unstructured data in memory. For example, see epee::serialization::storage_entry in contrib/epee/include/storages/portable_storage_base.h, or Value in rapidjson.

UkoeHB:
Is there some particular change required to support that?

UkoeHB:
Changes: added `noexcept` to `.index()` and `::type_index_of()`.

UkoeHB:
Changes: updated `is_type()` to use a cheap index check instead of `boost::strict_get()`.

jeffro256:
> Is there some particular change required to support that?

Actually, probably not but the user would have to use `boost::recursive_variant_wrapper` which is sort of a leaky abstraction.

src/common/variant.h
```diff
@@ -0,0 +1,149 @@
...
+        if (!value_ptr) variant_unwrap_err();
+        return *value_ptr;
+    }
+
+    /// get the type index of the currently stored type
+    int index() const noexcept { return m_value.which(); }
+
+    /// get the type index of a requested type (compile error for invalid types) (boost::mp11 is boost 1.66.0)
+    template <typename T>
+    static int type_index_of() noexcept
```
jeffro256, line 124:
```suggestion
    static constexpr int type_index_of() noexcept
```

jeffro256 (changes requested):

src/common/variant.h
jeffro256: This works as a `constexpr`, just tested it

src/common/variant.h
```diff
@@ -0,0 +1,135 @@
...
+    ResultT operator()(const boost::blank) const
+    {
+        throw std::runtime_error("variant: tried to visit an empty variant.");
+        static const ResultT dummy{};
+        return dummy;
+    }
+};
+
+template <typename... Types>
+class variant final
```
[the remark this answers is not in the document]
UkoeHB: `final` is part of a class's interface. I mark everything `final` that is not explicitly intended to be inherited from. In general, I feel 'implementation inheritance' should be avoided unless there is a very strong justification. Note that making it non-final would mean a virtual destructor is required.

UkoeHB:
Changes: add `try_unwrap()`, mark `type_index_of()` as `constexpr`.

src/common/variant.h
```diff
@@ -0,0 +1,156 @@
...
+//third party headers
+#include <boost/blank.hpp>
+#include <boost/mpl/begin_end.hpp>
+#include <boost/mpl/distance.hpp>
+#include <boost/mpl/find.hpp>
+#include <boost/mpl/vector.hpp>
+#include <boost/none_t.hpp>
+#include <boost/variant/apply_visitor.hpp>
+#include <boost/variant/get.hpp>
+#include <boost/variant/static_visitor.hpp>
```
jeffro256, line 44:
```suggestion
#include <boost/variant/recursive_wrapper.hpp>
#include <boost/variant/static_visitor.hpp>
```

src/common/variant.h
```diff
@@ -0,0 +1,156 @@
...
+// - the variant is 'optional' - an empty variant will evaluate to 'false' and an initialized variant will be 'true'
+///
+template <typename ResultT>
+struct variant_static_visitor : public boost::static_visitor<ResultT>
+{
+    /// provide visitation for empty variants
+    /// - add this to your visitor with: using variant_static_visitor::operator();
+    [[noreturn]] ResultT operator()(const boost::blank) const { variant_static_visitor_blank_err(); }
+};
+
```
jeffro256, line 75:
```suggestion

template <typename T>
using variant_recursive_wrapper = boost::recursive_wrapper<T>;

typedef boost::recursive_variant_ variant_placeholder;

template <typename... Ts>
using make_recursive_variant = boost::make_recursive_variant<Ts...>;
```

jeffro256 (commented):
<s>This will give the variant class the ability to make recursive variants just like Boost has: https://www.boost.org/doc/libs/1_58_0/doc/html/variant/tutorial.html#variant.tutorial.recursive</s>

Don't worry about this actually. It doesn't need this to work technically, and this might be an instance of YAGNI

jeffro256:
Made a PR against this PR here: https://github.com/UkoeHB/monero/pull/6. Just includes some unit tests

UkoeHB:
Changes: added some nice unit tests thanks to @jeffro256, added a commit on top of that with some formatting/legibility changes (will squash once @jeffro256 can take a look at the diff).

tests/unit_tests/variant.cpp
```diff
@@ -26,41 +26,45 @@
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-#include <sstream>
-#include <type_traits>
-#include <vector>
+#include "common/variant.h"
```
UkoeHB, line 29: Including this first means the header doesn't depend on transitive includes.

jeffro256:
You cant use `this->` in static methods :)

tests/unit_tests/variant.cpp
```diff
@@ -26,41 +26,45 @@
...
-#include "common/variant.h"
+#include <sstream>
+#include <type_traits>
+#include <vector>
 
 using tools::variant;
 using tools::variant_static_visitor;
 
 namespace
 {
```
UkoeHB, line 45: I still think the anonymous namespace is redundant, but leaving it in here since it's your file.

tests/unit_tests/variant.cpp
```diff
@@ -207,7 +219,7 @@ struct test_stringify_visitor: public variant_static_visitor<std::string>
     template <class Variant, typename T>
     static void test_visitation(const Variant& v, const T& t)
     {
-        EXPECT_EQ(stringify(t), v.visit(test_stringify_visitor()));
+        EXPECT_EQ(this->stringify(t), v.visit(test_stringify_visitor()));
```
UkoeHB, line 222: Explicit `this` helps disambiguate between free and member function calls.

tests/unit_tests/variant.cpp
```diff
@@ -398,4 +411,5 @@ TEST(variant, ad_hoc_recursion)
 
     EXPECT_EQ(5, left_count);
     EXPECT_EQ(4, right_count);
-}
\ No newline at end of file
+}
+//-------------------------------------------------------------------------------------------------------------------
```
UkoeHB: Github likes having a newline at the bottom.

tests/unit_tests/variant.cpp
UkoeHB: @jeffro256 oh you're right, whoops

tests/unit_tests/variant.cpp
jeffro256: I particularly like using anonymous namespaces because `static` doesn't prevent ODR violations. I don't know, I don't care too much either way.

src/common/variant.h
```diff
@@ -0,0 +1,149 @@
...
+    variant(boost::none_t) : variant{} {}  //act like boost::optional
+
+    /// construct from variant type (use enable_if to avoid issues with copy/move constructor)
+    template <typename T,
+        typename std::enable_if<
+                !std::is_same<
+                        std::remove_cv_t<std::remove_reference_t<T>>,
+                        variant<Types...>
+                    >::value,
+                bool
```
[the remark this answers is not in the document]
vtnerd: I meant to delete the above comment, just whatever.

PR #8661, opened 2022-12-01
[seraphis] common: add container helpers

UkoeHB:

- Adds templates to assist with handling containers.

### DIFF
--- a/src/common/container_helpers.h
+++ b/src/common/container_helpers.h
@@ -47,21 +47,20 @@
 namespace tools
 {
 
-/// use operator< to get operator==
-/// WARNING: equality is not always implied by operator<, depending on implementation
-struct equals_from_less final
+/// test if a container is sorted and unique according to a comparison criteria (defaults to operator<)
+template <typename T, typename ComparisonOpT = std::less<typename T::value_type>>
+bool is_sorted_and_unique(const T &container, const ComparisonOpT &ComparisonOp = ComparisonOpT{})
 {
-    template <typename T>
-    bool operator()(const T &a, const T &b) { return !(a < b) && !(b < a); }
-};
-/// note: test for sorted and uniqueness using the same criteria (operator<)
-template <typename T>
-bool is_sorted_and_unique(const T& container)
-{
-    if (!std::is_sorted(container.begin(), container.end()))
+    if (!std::is_sorted(container.begin(), container.end(), ComparisonOp))
         return false;
 
-    if (std::adjacent_find(container.begin(), container.end(), equals_from_less{}) != container.end())
+    if (std::adjacent_find(container.begin(),
+                container.end(),
+                [&ComparisonOp](const typename T::value_type &a, const typename T::value_type &b) -> bool
+                {
+                    return !ComparisonOp(a, b) && !ComparisonOp(b, a);
+                })
+            != container.end())
         return false;
 
     return true;
--- a/src/common/container_helpers.h
+++ b/src/common/container_helpers.h
@@ -47,7 +47,14 @@
 namespace tools
 {
 
+/// convert a binary comparison function to a functor
+template <typename T, typename ComparisonOpT = bool(const T &a, const T &b)>
+inline auto compare_func(const ComparisonOpT &comparison_function)
+{
+    return [&comparison_function](const T &a, const T &b) -> bool { return comparison_function(a, b); };
+}
 /// test if a container is sorted and unique according to a comparison criteria (defaults to operator<)
+/// NOTE: ComparisonOpT must establish 'strict weak ordering' https://en.cppreference.com/w/cpp/named_req/Compare
 template <typename T, typename ComparisonOpT = std::less<typename T::value_type>>
 bool is_sorted_and_unique(const T &container, const ComparisonOpT &ComparisonOp = ComparisonOpT{})
 {
@@ -64,6 +71,13 @@ bool is_sorted_and_unique(const T &container, const ComparisonOpT &ComparisonOp 
         return false;
 
     return true;
+}
+/// specialization for raw function pointers
+template <typename T>
+bool is_sorted_and_unique(const T &container,
+    bool (*const ComparisonOpFunc)(const typename T::value_type &a, const typename T::value_type &b))
+{
+    return is_sorted_and_unique(container, compare_func<typename T::value_type>(ComparisonOpFunc));
 }
 /// convenience wrapper for checking if a mapped object is mapped to a key embedded in that object
 /// example: std::unorderd_map<rct::key, std::pair<rct::key, rct::xmr_amount>> where the map key is supposed to
--- a/src/common/container_helpers.h
+++ b/src/common/container_helpers.h
@@ -36,10 +36,8 @@
 
 //standard headers
 #include <algorithm>
-#include <functional>
 #include <unordered_map>
 #include <utility>
-#include <vector>
 
 //forward declarations
 
@@ -49,23 +47,50 @@ namespace tools
 
 /// convert a binary comparison function to a functor
 template <typename T, typename ComparisonOpT = bool(const T &a, const T &b)>
-inline auto compare_func(const ComparisonOpT &comparison_function)
+inline auto compare_func(ComparisonOpT comparison_op_func)
 {
-    return [&comparison_function](const T &a, const T &b) -> bool { return comparison_function(a, b); };
+    static_assert(
+            std::is_same<
+                    bool,
+                    decltype(
+                        comparison_op_func(
+                            std::declval<std::remove_cv_t<T>>(),
+                            std::declval<std::remove_cv_t<T>>()
+                        )
+                    )
+                >::value,
+            "invalid callable - expected callable in form bool(T, T)"
+        );
+
+    return [func = std::move(comparison_op_func)] (const T &a, const T &b) -> bool { return func(a, b); };
 }
 /// test if a container is sorted and unique according to a comparison criteria (defaults to operator<)
 /// NOTE: ComparisonOpT must establish 'strict weak ordering' https://en.cppreference.com/w/cpp/named_req/Compare
 template <typename T, typename ComparisonOpT = std::less<typename T::value_type>>
-bool is_sorted_and_unique(const T &container, const ComparisonOpT &ComparisonOp = ComparisonOpT{})
+bool is_sorted_and_unique(const T &container, ComparisonOpT comparison_op = ComparisonOpT{})
 {
-    if (!std::is_sorted(container.begin(), container.end(), ComparisonOp))
+    using ValueT = typename T::value_type;
+    static_assert(
+            std::is_same<
+                    bool,
+                    decltype(
+                        comparison_op(
+                            std::declval<std::remove_cv_t<ValueT>>(),
+                            std::declval<std::remove_cv_t<ValueT>>()
+                        )
+                    )
+                >::value,
+            "invalid callable - expected callable in form bool(ValueT, ValueT)"
+        );
+
+    if (!std::is_sorted(container.begin(), container.end(), comparison_op))
         return false;
 
     if (std::adjacent_find(container.begin(),
                 container.end(),
-                [&ComparisonOp](const typename T::value_type &a, const typename T::value_type &b) -> bool
+                [comparison_op](const ValueT &a, const ValueT &b) -> bool
                 {
-                    return !ComparisonOp(a, b) && !ComparisonOp(b, a);
+                    return !comparison_op(a, b) && !comparison_op(b, a);
                 })
             != container.end())
         return false;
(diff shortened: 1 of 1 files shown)


src/common/container_helpers.h
```diff
@@ -0,0 +1,106 @@
...
+bool is_sorted_and_unique(const T& container)
+{
+    if (!std::is_sorted(container.begin(), container.end()))
+        return false;
+
+    if (std::adjacent_find(container.begin(), container.end(), equals_from_less{}) != container.end())
+        return false;
+
+    return true;
+}
```
moneromooo-monero, line 68: While this seems correct, silently defaulting to two < instead of a == will hinder performance (I suspect operator== is typically faster (or at least not slower) than even just one operator<). So this seems an odd choice. Is this because operator< is more likely to be available ?

src/common/container_helpers.h
```diff
@@ -0,0 +1,106 @@
...
+/// convenience wrapper for checking if a mapped object is mapped to a key embedded in that object
+template <typename KeyT, typename ValueT>
+bool keys_match_internal_values(const std::unordered_map<KeyT, ValueT> &map,
```
moneromooo-monero, line 71: You probably want an example in a comment, this one is confusing

src/common/container_helpers.h
UkoeHB: Checking if a container is sorted uses `operator<`. I want to use the same criteria for testing uniqueness. For example, I use this to check if tx output enotes are sorted by onetime address and also that all onetime addresses are unique.

UkoeHB: Tbh I am not convinced this is the best solution. Even though it is clean at the call-site, it is not clear what's happening. I may update this PR in the next couple days with an alternative.

UkoeHB: Ok I updated this to allow a user-defined comparison operator so you can add call-site clarity for non-trivial types by defining a comparison function.

Note that testing uniqueness with `adjacent_find` only works if you use the same comparison criteria as `is_sorted`.

vtnerd (changes requested):

src/common/container_helpers.h
```diff
@@ -0,0 +1,121 @@
...
+
+
+namespace tools
+{
+
+/// convert a binary comparison function to a functor
+template <typename T, typename ComparisonOpT = bool(const T &a, const T &b)>
+inline auto compare_func(const ComparisonOpT &comparison_function)
+{
+    return [&comparison_function](const T &a, const T &b) -> bool { return comparison_function(a, b); };
```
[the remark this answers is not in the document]
UkoeHB: I need this adaptor to disambiguate function overloads at the callsite. [Example of adaptor](https://github.com/UkoeHB/monero/blob/3517c428dd1232cdf14dda9b4d74fe0eb6b1bbe2/src/seraphis/tx_builders_mixed.cpp#L857), [example of is_sorted_and_unique()](https://github.com/UkoeHB/monero/blob/3517c428dd1232cdf14dda9b4d74fe0eb6b1bbe2/src/seraphis/tx_validators.cpp#L356).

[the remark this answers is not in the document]
UkoeHB: However, it's probably fine to update the adaptor to return a `std::cref()`.

src/common/container_helpers.h
```diff
@@ -0,0 +1,121 @@
...
+}
+/// test if a container is sorted and unique according to a comparison criteria (defaults to operator<)
+/// NOTE: ComparisonOpT must establish 'strict weak ordering' https://en.cppreference.com/w/cpp/named_req/Compare
+template <typename T, typename ComparisonOpT = std::less<typename T::value_type>>
+bool is_sorted_and_unique(const T &container, const ComparisonOpT &ComparisonOp = ComparisonOpT{})
+{
+    if (!std::is_sorted(container.begin(), container.end(), ComparisonOp))
```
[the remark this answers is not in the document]
UkoeHB: Looks like I call it 20 times in `seraphis_lib`. Hand-rolling an is_sorted check seems excessive.

src/common/container_helpers.h
```diff
@@ -0,0 +1,121 @@
...
+}
+/// convenience wrapper for checking if a mapped object is mapped to a key embedded in that object
+/// example: std::unorderd_map<rct::key, std::pair<rct::key, rct::xmr_amount>> where the map key is supposed to
+///   reproduce the pair's rct::key; use the predicate to get the pair's rct::key element
+template <typename KeyT, typename ValueT>
+bool keys_match_internal_values(const std::unordered_map<KeyT, ValueT> &map,
+    const std::function<
+            const typename std::unordered_map<KeyT, ValueT>::key_type&
+            (const typename std::unordered_map<KeyT, ValueT>::mapped_type&)
+        > &get_internal_key_func)
```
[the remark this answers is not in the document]
UkoeHB: I wrote something that works, that's all. Do you have an alternative?

src/common/container_helpers.h
```diff
@@ -0,0 +1,121 @@
...
+}
+/// specialization for raw function pointers
+template <typename T>
+bool is_sorted_and_unique(const T &container,
+    bool (*const ComparisonOpFunc)(const typename T::value_type &a, const typename T::value_type &b))
+{
+    return is_sorted_and_unique(container, compare_func<typename T::value_type>(ComparisonOpFunc));
+}
+/// convenience wrapper for checking if a mapped object is mapped to a key embedded in that object
+/// example: std::unorderd_map<rct::key, std::pair<rct::key, rct::xmr_amount>> where the map key is supposed to
```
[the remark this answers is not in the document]
UkoeHB: I use this a few times in the multisig utilities, [example](https://github.com/UkoeHB/monero/blob/3517c428dd1232cdf14dda9b4d74fe0eb6b1bbe2/src/multisig/multisig_signing_helper_utils.cpp#L619).

src/common/container_helpers.h
```diff
@@ -0,0 +1,121 @@
...
+template <typename ContainerT>
+typename ContainerT::value_type& add_element(ContainerT &container)
+{
+    container.emplace_back();
+    return container.back();
+}
+/// convenience erasor for unordered maps: std::erase_if(std::unordered_map) is C++20
+template <typename KeyT, typename ValueT>
+void for_all_in_map_erase_if(std::unordered_map<KeyT, ValueT> &map_inout,
+    const std::function<bool(const typename std::unordered_map<KeyT, ValueT>::value_type&)> &predicate)
```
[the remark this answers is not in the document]
UkoeHB: Alternative?

src/common/container_helpers.h
```diff
@@ -0,0 +1,121 @@
...
+/// convenience wrapper for getting the last element after emplacing back
+template <typename ContainerT>
+typename ContainerT::value_type& add_element(ContainerT &container)
+{
+    container.emplace_back();
+    return container.back();
+}
+/// convenience erasor for unordered maps: std::erase_if(std::unordered_map) is C++20
+template <typename KeyT, typename ValueT>
+void for_all_in_map_erase_if(std::unordered_map<KeyT, ValueT> &map_inout,
```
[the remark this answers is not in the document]
UkoeHB: `erase_if` takes a pair of iterators, this function iterates the entire map.

src/common/container_helpers.h
```diff
@@ -0,0 +1,121 @@
...
+                    return !ComparisonOp(a, b) && !ComparisonOp(b, a);
+                })
+            != container.end())
+        return false;
+
+    return true;
+}
+/// specialization for raw function pointers
+template <typename T>
+bool is_sorted_and_unique(const T &container,
```
vtnerd, line 77: I missed this earlier, no overload is neeeded for raw functions if the argument is taken by-value.

src/common/container_helpers.h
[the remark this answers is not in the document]
vtnerd:
```c++
template<typename K, typename V, typename F>
bool keys_match_internal_values(const std::unordered_map<K, V>& map, F f)
```
also works with all callables, and `std::cref`/`std::ref` can be used by caller as needed (or similar).

src/common/container_helpers.h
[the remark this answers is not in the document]
vtnerd: > disambiguate function overloads at the callsite.

Yikes. There has to be a built-in way to handle this. It's easy with templated overloads, but not certain (off-top-of-head) how with non-templated functions.

Anyway, yes, take the argument by-value and then do the lambda capture by-value. Prevents any unexpected nastiness if someone uses the function in some funky way too.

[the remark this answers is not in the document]
vtnerd: And it could be a C++14 move-capture-by-value too.

src/common/container_helpers.h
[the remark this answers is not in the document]
vtnerd: The `boost::range` library and Niebler's rangev3 library use the same naming syntax despite taking ranges. This is basically an `erase_if`, that could take `typename T` and probably work with all containers (afaik).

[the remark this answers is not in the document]
UkoeHB: A range is similar to a pair of iterators...

src/common/container_helpers.h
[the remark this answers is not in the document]
vtnerd: Just remembered the typical C++ way to handle this -
```c++
struct compare_KI_
{
    bool operator()(const key_image1&, const key_image2&) const;
    bool operator()(const key_image2&, const key_image2&) const;
};
constexpr const compare_KI_ compare_KI{};
```
And then the overload selection is chosen inside `std::sort` instead of at the call-site to `std::sort`.  But this technique you've come up with has a bit less boilerplate and doesn't require the declarations to be in the same header.

src/common/container_helpers.h
UkoeHB: Except for the problem of overload resolution.

vtnerd: My frustration with these solutions for compare function overloads - plenty of examples of people using generic predicates don't run into this issue by: (1) using unique function names, (2) using the `struct` solution given in another comment, or (3) using `std::bind` like so: `std::bind<bool(int, int)>(my_compare, _1, _2)`.

I'm not completely against these custom solutions (although I'm wondering suspiciously why solution (1) above isn't being used), but ALSO why not use the `compare_func` above instead of a custom `is_sorted_and_unique` which only works with `const&` arguments for the compare function?

--------------

I also wanted to document (somewhere) this more generic solution I came up with:

```c++
template <typename F>
inline auto select_func(F f)
{
    return [f = std::move(f)](auto... args) {
        return f(std::forward<decltype(args)>(args)...); 
    };
}

template<typename T, typename F = bool(const T&, const T&)>
inline auto compare_func(F f)
{
    return select_func(std::move(f));
}
```
which can be used like: 
```c++
bool my_compare(int, int);
bool my_compare(const public_key&, const public_key&);
std::sort(vec1.begin(), vec1.end(), select_func<bool(int, int)>(my_compare));
std::sort(vec2.begin(), vec2.end(), compare_func<public_key>(my_compare));
```

UkoeHB: I wrote a tool that works for my use case. You're kind of saying I shouldn't have this tool because it doesn't work for code I didn't write... (comparison functions that aren't using `const&` parameters).

In practice, user-defined comparison free-functions should almost always be using `const&` parameters. Trivial types don't need a user-defined function - just use the built-in `operator<`. If anything, enforcing `const&` prevents accidentally writing expensive copies.

For completeness I added your `select_func` thing (renamed to `as_functor`).

src/common/container_helpers.h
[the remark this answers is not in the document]
UkoeHB: I updated `keys_match_internal_values()` so the predicate is in charge of checking the key mapping, which should be a little more generic and understandable at the call-site.

src/common/container_helpers.h
UkoeHB: > return [f = std::move(f)](auto... args) {
>        return f(std::forward<decltype(args)>(args)...); 
>    };

Btw, don't the `args` here become value parameters (copies), not references?

vtnerd: Yup, I botched that particular portion, it needs to be `auto&&...`.

src/common/container_helpers.h
```diff
@@ -0,0 +1,170 @@
...
+                        )
+                    )
+                >::value,
+            "invalid callable - expected callable in form bool(ValueT, ValueT)"
+        );
+
+    if (!std::is_sorted(container.begin(), container.end(), comparison_op))
+        return false;
+
+    if (std::adjacent_find(container.begin(),
```
vtnerd, line 96: If the container is sorted, `std::lower_bound` instead? If the list is large, `lower_bound` will make a difference since its a logarithmic search.

src/common/container_helpers.h
```diff
@@ -0,0 +1,170 @@
...
+    using MapValueT = typename std::unordered_map<KeyT, ValueT>::value_type;
+    static_assert(
+            std::is_same<
+                    bool,
+                    decltype(predicate(std::declval<std::remove_cv_t<MapValueT>>()))
+                >::value,
+            "invalid callable - expected callable in form bool(Value)"
+        );
+
+    for (auto map_it = map_inout.begin(); map_it != map_inout.end();)
```
vtnerd, line 161: So the difference here is that the key,value are given to the predicate, as opposed to just the key with the standard search functions.

vtnerd (approved):

src/common/container_helpers.h
vtnerd: Could've been solved by me reading `adjacent_find` documentation.

src/common/container_helpers.h
vtnerd: Yup, could've spotted that with some better inspection.

vtnerd (approved):

PR #8662, opened 2022-12-01
[seraphis] crypto: add generators file

UkoeHB:

- Provide direct access to different representations of canonical fixed generators.
- Includes automatic validation of generator constants in debug mode.

**Future work**
- Add seraphis generators `X` and `U`.

### DIFF
--- a/src/crypto/generators.cpp
+++ b/src/crypto/generators.cpp
@@ -1,0 +1,186 @@
+// Copyright (c) 2022, The Monero Project
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+// 
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+// 
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+// 
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "generators.h"
+
+#include "crypto.h"
+extern "C"
+{
+#include "crypto-ops.h"
+}
+#include "hash.h"
+
+#include <cassert>
+#include <cstddef>
+#include <cstdint>
+#include <mutex>
+
+namespace crypto
+{
+
+/// constexpr assert for old gcc bug: https://stackoverflow.com/questions/34280729/throw-in-constexpr-function
+/// - this function won't compile in a constexpr context if b == false
+constexpr void constexpr_assert(const bool b) { b ? 0 : throw std::runtime_error("constexpr assert failed"); };
+
+/// constexpr paste bytes into an array-of-bytes type
+template<typename T>
+constexpr T bytes_to(const std::initializer_list<unsigned char> bytes)
+{
+    T out{}; // zero-initialize trailing bytes
+
+    auto current = std::begin(out.data);
+    constexpr_assert(static_cast<long>(bytes.size()) <= std::end(out.data) - current);
+
+    for (const unsigned char byte : bytes)
+        *current++ = byte;
+    return out;
+}
+
+// generators
+//standard ed25519 generator G: {x, 4/5} (positive x when decompressing y = 4/5)
+constexpr public_key G = bytes_to<public_key>({ 0x58, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66,
+    0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66 });
+//pedersen commitment generator H: toPoint(cn_fast_hash(G))
+constexpr public_key H = bytes_to<public_key>({ 0x8b, 0x65, 0x59, 0x70, 0x15, 0x37, 0x99, 0xaf, 0x2a, 0xea, 0xdc, 0x9f, 0xf1,
+    0xad, 0xd0, 0xea, 0x6c, 0x72, 0x51, 0xd5, 0x41, 0x54, 0xcf, 0xa9, 0x2c, 0x17, 0x3a, 0x0d, 0xd3, 0x9c, 0x1f, 0x94 });
+static ge_p3 G_p3;
+static ge_p3 H_p3;
+static ge_cached G_cached;
+static ge_cached H_cached;
+
+// misc
+static std::once_flag init_gens_once_flag;
+
+//-------------------------------------------------------------------------------------------------------------------
+//-------------------------------------------------------------------------------------------------------------------
+static public_key reproduce_generator_G()
+{
+    // G = {x, 4/5 mod q}
+    fe four, five, inv_five, y;
+    fe_0(four);
+    fe_0(five);
+    four[0] = 4;
+    five[0] = 5;
+    fe_invert(inv_five, five);
+    fe_mul(y, four, inv_five);
+
+    public_key reproduced_G;
+    fe_tobytes(to_bytes(reproduced_G), y);
+
+    return reproduced_G;
+}
+//-------------------------------------------------------------------------------------------------------------------
+//-------------------------------------------------------------------------------------------------------------------
+static public_key reproduce_generator_H()
+{
+    // H = 8*to_point(keccak(G))
+    // note: this does not use the point_from_bytes() function found in H_p(), instead directly interpreting the
+    //       input bytes as a compressed point (this can fail, so should not be used generically)
+    // note2: to_point(keccak(G)) is known to succeed for the canonical value of G (it will fail 7/8ths of the time
+    //        normally)
+    ge_p3 temp_p3;
+    ge_p2 temp_p2;
+    ge_p1p1 temp_p1p1;
+
+    hash H_temp_hash{cn_fast_hash(to_bytes(G), sizeof(ec_point))};
+    (void)H_temp_hash;  //suppress unused warning
+    assert(ge_frombytes_vartime(&temp_p3, reinterpret_cast<const unsigned char*>(&H_temp_hash)) == 0);
+    ge_p3_to_p2(&temp_p2, &temp_p3);
+    ge_mul8(&temp_p1p1, &temp_p2);
+    ge_p1p1_to_p3(&temp_p3, &temp_p1p1);
+
+    public_key reproduced_H;
+    ge_p3_tobytes(to_bytes(reproduced_H), &temp_p3);
+
+    return reproduced_H;
+}
+//-------------------------------------------------------------------------------------------------------------------
+// Make generators, but only once
+//-------------------------------------------------------------------------------------------------------------------
+static void init_gens()
+{
+    std::call_once(init_gens_once_flag,
+        [&](){
+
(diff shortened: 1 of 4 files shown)


UkoeHB:
Oh whoops this one depends on another PR. Will rebase when the dependency is resolved.

src/crypto/generators.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    public_key reproduced_H;
+    ge_p3_tobytes(to_bytes(reproduced_H), &temp_p3);
+
+    return reproduced_H;
+}
+//-------------------------------------------------------------------------------------------------------------------
+// Make generators, but only once
+//-------------------------------------------------------------------------------------------------------------------
+static void init_gens()
```
jeffro256, line 129: I think it's really cool how you're showing your work through code; I genuinely think its refreshing. But since all the calls `get_G, getH, etc` call `init_gens` internally, this introduces extra release binary size and slows down runtime. Maybe this would be better as a unit test? Also, that way, it would be checked at compile-time on the many platforms we run CI on every single PR.

UkoeHB: The `init_gens()` is necessary for converting the `unsigned char[]` representation to `char[]`, which is what `crypto::public_key` has internally.

jeffro256: You can have `G` and `H` represented directly as a `constexpr` if you put in the correct `char` values, so you don't need any copying at the beginning of the program, no initialization functions calls, or `std::once_flag`s. It's a little uglier to read because of the negative signs, but they're either the right values or they're not.

jeffro256: Example:
```
constexpr const public_key G = { {88, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102, 102} };
constexpr const public_key H = { {-117, 101, 89, 112, 21, 55, -103, -81, 42, -22, -36, -97, -15, -83, -48, -22, 108, 114, 81, -43, 65, 84, -49, -87, 44, 23, 58, 13, -45, -100, 31, -108} };
```

UkoeHB: Well you'd still need the once_flag to generate the other representations.

jeffro256: Unless there's something I'm not seeing, you wouldn't need the whole `std::call_once` block since `G`, `H`, `G_p3`, `H_p3`, `G_cached`, and `H_cached` all have POD types and the process for deriving these constants are deterministic. It's always going to be the same in the end so you can just cut out that work right? If you still want to check the constants at runtime on every run for debug builds, we can still do this if we shuffle everything into a function `reproduce_and_check_all_generators` or something. 

We can also have the raw byte representations in a comment right above it as well as the mathematical way to derive the point.

jeffro256: That way there's 0 overhead but we still get the same safety guarantees.

jeffro256: Okay I did better performance tests and dang... the compiler was better at optimizing that stuff away than I was expecting. Once that dependency is resolved, I think I'll be good.

src/crypto/generators.cpp
```diff
@@ -0,0 +1,192 @@
...
+static ge_cached G_cached;
+static ge_cached H_cached;
+
+// misc
+static std::once_flag init_gens_once_flag;
+
+//-------------------------------------------------------------------------------------------------------------------
+// hash-to-point: H_p(x) = 8*point_from_bytes(keccak(x))
+//-------------------------------------------------------------------------------------------------------------------
+static void hash_to_point(const hash &x, crypto::ec_point &point_out)
```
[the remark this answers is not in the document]
UkoeHB: Oh this will be used when I add the seraphis generators `X` and `U`.

src/crypto/generators.cpp
UkoeHB: Which dependency?

src/crypto/generators.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+// generators
+//standard ed25519 generator G: {x, 4/5} (positive x when decompressing y = 4/5)
+static const uint_point G_uint = { {0x58, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66} };
+static public_key G;
+//pedersen commitment generator H: toPoint(cn_fast_hash(G))
+static const uint_point H_uint = { {0x8b, 0x65, 0x59, 0x70, 0x15, 0x37, 0x99, 0xaf, 0x2a, 0xea, 0xdc, 0x9f, 0xf1, 0xad, 0xd0, 0xea, 0x6c, 0x72, 0x51, 0xd5, 0x41, 0x54, 0xcf, 0xa9, 0x2c, 0x17, 0x3a, 0x0d, 0xd3, 0x9c, 0x1f, 0x94} };
+static public_key H;
+static ge_p3 G_p3;
+static ge_p3 H_p3;
```
moneromooo-monero, line 60: That one's already in src/crypto (assuming it's H in ge_p3 form)

src/crypto/generators.cpp
```diff
@@ -0,0 +1,192 @@
...
+static ge_cached G_cached;
+static ge_cached H_cached;
+
```
moneromooo-monero, line 63: Why are those not public in crypto-ops-data with the other similar data ?

src/crypto/generators.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include <mutex>
+#include <string>
+
+namespace crypto
+{
+
+struct uint_point
+{
+    unsigned char bytes[32];
+};
```
moneromooo-monero, line 50: Why yet another type for this ? There's already rct::key and ec_point, which is already annoying. Is this for different semantics ?

src/crypto/generators.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+// generators
+//standard ed25519 generator G: {x, 4/5} (positive x when decompressing y = 4/5)
+static const uint_point G_uint = { {0x58, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66} };
+static public_key G;
+//pedersen commitment generator H: toPoint(cn_fast_hash(G))
+static const uint_point H_uint = { {0x8b, 0x65, 0x59, 0x70, 0x15, 0x37, 0x99, 0xaf, 0x2a, 0xea, 0xdc, 0x9f, 0xf1, 0xad, 0xd0, 0xea, 0x6c, 0x72, 0x51, 0xd5, 0x41, 0x54, 0xcf, 0xa9, 0x2c, 0x17, 0x3a, 0x0d, 0xd3, 0x9c, 0x1f, 0x94} };
```
moneromooo-monero, line 57: Already there in src/ringct, along with G. Why duplicate ?

src/crypto/generators.cpp
```diff
@@ -0,0 +1,192 @@
...
+    ge_p3 temp_p3;
+    ge_p2 temp_p2;
+    ge_p1p1 temp_p1p1;
+
+    cn_fast_hash(reinterpret_cast<const unsigned char*>(&x), sizeof(hash), h);
+    ge_fromfe_frombytes_vartime(&temp_p2, reinterpret_cast<const unsigned char*>(&h));
+    ge_mul8(&temp_p1p1, &temp_p2);
+    ge_p1p1_to_p3(&temp_p3, &temp_p1p1);
+    ge_p3_tobytes(to_bytes(point_out), &temp_p3);
+}
```
moneromooo-monero, line 82: This is mostly hash_to_p3 (plus the ge_p3_tobytes at the end).

src/crypto/generators.cpp
```diff
@@ -0,0 +1,192 @@
...
+        // copy uint generator representations to standard 'ec_point' representations (char arrays)
+        memcpy(G.data, G_uint.bytes, 32);
+        memcpy(H.data, H_uint.bytes, 32);
+
+        // build ge_p3 representations of generators
+        const int G_deserialize = ge_frombytes_vartime(&G_p3, to_bytes(G));
+        const int H_deserialize = ge_frombytes_vartime(&H_p3, to_bytes(H));
+
+        (void)G_deserialize; assert(G_deserialize == 0);
+        (void)H_deserialize; assert(H_deserialize == 0);
```
moneromooo-monero, line 143: This is OK I guess since it's totally constant but this type of asserts means they only ever get caught if you build without -DNDEBUG (which I pretty much never do, but I'm not sure it's common). So it looks tested, but it might not in practice. So OK in this particular case, but something more dynamic would probably not be.

moneromooo-monero (commented):
Looks like a fair bit of redundancy here.

src/crypto/generators.cpp
UkoeHB: Yes looks like `ge_p3_H` is in `crypto-ops-data.c`. That is not a file I consider part of the src/crypto interface (it's not even a header).

src/crypto/generators.cpp
UkoeHB: `crypto-ops-data.c` is not something I want to be including everywhere.

src/crypto/generators.cpp
UkoeHB: I want a unified interface for getting generators, with both `ge_p3` and `ge_cached` representations, and functions to auto-validate that the recorded points are valid. Picking things up from all over the code base is too haphazard.

This file will contain seraphis generators in the future.

src/crypto/generators.cpp
UkoeHB: Yep, but `src/ringct` is not linked to `src/crypto` and doing so would create a circular dependency.

src/crypto/generators.cpp
vtnerd: Adding more to the discussion - the compiler is good at optimizing, but every `init_gens()` call produces a read-acquire statement that requires synchronization across cores. As an example, use gcc godbolt to generate code for a function-local static - the x86-64 ASM looks like a simple double-check lock but the arm64 ASM has a `dmb` instruction (memory barrier). This isn't a huge impact on performance, but does add stalls/wait cycles.

Constexpr was expanded in C++14, so compile-time conversion and 2's complement checking should now be possible across platforms:
```c++
template<typename T>
constexpr T bytes_to(const std::initializer_list<unsigned char> in)
{
    T out{}; // zero-initialize trailing bytes

    auto current = std::begin(out.data);
    if (std::end(out.data) - current < in.size())
        throw std::runtime_error{"invalid initializer size"};

    for (const auto byte : in)
        *current++ = byte;
    return out;
}

struct char_test { char data[5]; };
constexpr const auto test_case = bytes_to<char_test>({0x00, 0x7f, 0x80, 0xff});
static_assert(test_case.data[0] == 0, "");
static_assert(test_case.data[1] == 127, "");
static_assert(test_case.data[2] == char(-128), "");
static_assert(test_case.data[3] == char(-1), "");
static_assert(test_case.data[4] == 0, "");

constexpr const auto G = bytes_to<public_key>(
    {0x58, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 
    0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66}
);
```
Tested against x86-64, arm64, and risc-v on gcc godbolt.

vtnerd: And I was suggesting this for `G` and `H`, but leaving the `init_gens()` for the other values. The test harness will still call `init_gens()` by calling values _other than_ `G` and `H`.

vtnerd (changes requested):

src/crypto/generators.h
```diff
@@ -0,0 +1,47 @@
...
+extern "C"
+{
+#include "crypto-ops.h"
+}
+#include "crypto.h"
+
+namespace crypto
+{
+
+public_key get_G();
```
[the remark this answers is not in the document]
UkoeHB: I am somewhat paranoid about the possibility of modifying these constants via `const_cast`. Doing so is undefined behavior - which means the safety attributes are compiler-dependent.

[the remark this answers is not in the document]
vtnerd: I meant `get_G()` could return `const public_key&` instead of a copy, etc. I'm not aware of any issues, but might've overlooked something.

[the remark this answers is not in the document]
vtnerd: Oh sorry, you were concerned about some reckless programmer? Yikes, that's fairly pessimistic, but I guess a 32-byte copy is fairly quick since it will always be vectorized-aligned by definition.

UkoeHB:
Rebased and did slight cleanup (improved comments, removed unused function).

src/crypto/generators.cpp
```diff
@@ -0,0 +1,186 @@
...
+constexpr void constexpr_assert(const bool b) { b ? 0 : throw std::runtime_error("constexpr assert failed"); };
+
+/// constexpr paste bytes into an array-of-bytes type
+template<typename T>
+constexpr T bytes_to(const std::initializer_list<unsigned char> bytes)
+{
+    T out{}; // zero-initialize trailing bytes
+
+    auto current = std::begin(out.data);
+    constexpr_assert(static_cast<long>(bytes.size()) <= std::end(out.data) - current);
```
jeffro256, line 57:
```suggestion
    constexpr_assert(static_cast<long>(bytes.size()) == std::end(out.data) - current);
```

I would make it `==` just so some bytes aren't left uninitialized on accident.

UkoeHB: @vtnerd this was your suggestion, any comment?

vtnerd: Afaik, initialized data is only possible if a custom default constructor leaves data uninitialized, which is against standard practices.

vtnerd (approved):

src/crypto/generators.cpp
vtnerd: @jeffro256 had a comment about uninitialized data. With brace initialization, this is only possible if a default constructor is provided for the type, and that default constructor leaves some bytes uninitialized.

PR #8663, opened 2022-12-02
[seraphis] crypto: miscellaneous updates

UkoeHB:

- Make some `crypto-ops.c` static methods public. These will be used in PR #8662 for recomputing the main ed25519 generator.
- Allow sorting of key images.
- Add `to_bytes()` methods for converting `ec_point` and `ec_scalar` to unsigned char pointers. These methods are SUPER helpful when doing anything with the `sc_*` methods. Note that there is a variant of this in `device_default.cpp`.
- Rename `rct::ecdhHash()` to `rct::genAmountEncodingFactor()` and make it public.

### DIFF
--- a/src/crypto/crypto-ops.h
+++ b/src/crypto/crypto-ops.h
@@ -30,6 +30,8 @@
 
 #pragma once
 
+#include <stdint.h>
+
 /* From fe.h */
 
 typedef int32_t fe[10];
@@ -161,5 +163,7 @@ void ge_sub(ge_p1p1 *r, const ge_p3 *p, const ge_cached *q);
 void fe_add(fe h, const fe f, const fe g);
 void fe_tobytes(unsigned char *, const fe);
 void fe_invert(fe out, const fe z);
+void fe_mul(fe out, const fe, const fe);
+void fe_0(fe h);
 
 int ge_p3_is_point_at_infinity_vartime(const ge_p3 *p);
--- a/src/crypto/crypto.h
+++ b/src/crypto/crypto.h
@@ -335,7 +335,15 @@ namespace crypto {
 
   inline bool operator<(const public_key &p1, const public_key &p2) { return memcmp(&p1, &p2, sizeof(public_key)) < 0; }
   inline bool operator>(const public_key &p1, const public_key &p2) { return p2 < p1; }
+  inline bool operator<(const key_image &p1, const key_image &p2) { return memcmp(&p1, &p2, sizeof(key_image)) < 0; }
+  inline bool operator>(const key_image &p1, const key_image &p2) { return p2 < p1; }
 }
+
+// type conversions for easier calls to sc_add(), sc_sub(), hash functions
+inline unsigned char* to_bytes(crypto::ec_scalar &scalar) { return &reinterpret_cast<unsigned char&>(scalar); }
+inline const unsigned char* to_bytes(const crypto::ec_scalar &scalar) { return &reinterpret_cast<const unsigned char&>(scalar); }
+inline unsigned char* to_bytes(crypto::ec_point &point) { return &reinterpret_cast<unsigned char&>(point); }
+inline const unsigned char* to_bytes(const crypto::ec_point &point) { return &reinterpret_cast<const unsigned char&>(point); }
 
 CRYPTO_MAKE_HASHABLE(public_key)
 CRYPTO_MAKE_HASHABLE_CONSTANT_TIME(secret_key)
--- a/src/crypto/crypto-ops.c
+++ b/src/crypto/crypto-ops.c
@@ -38,7 +38,6 @@ DISABLE_VS_WARNINGS(4146 4244)
 
 /* Predeclarations */
 
-static void fe_mul(fe, const fe, const fe);
 static void fe_sq(fe, const fe);
 static void ge_madd(ge_p1p1 *, const ge_p3 *, const ge_precomp *);
 static void ge_msub(ge_p1p1 *, const ge_p3 *, const ge_precomp *);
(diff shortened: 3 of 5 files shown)


src/crypto/crypto-ops.h
```diff
@@ -30,6 +30,8 @@
 
 #pragma once
 
+#include <stdint.h>
```
SChernykh, line 33: Why was it added here? `crypto-ops.c` already includes it before `crypto.h`

SChernykh: It makes sense then to remove `#include <stdint.h>` from `crypto-ops.c`

UkoeHB: This cleans up static analyzer errors.

If you [google transitive includes](https://www.google.com/search?client=firefox-b-1-d&q=transitive+includes) you don't find anyone recommending they be used. Here is the somewhat authoritative [cpp core guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-implicit).

src/crypto/crypto.h
```diff
@@ -335,11 +335,19 @@ namespace crypto {
 
   inline bool operator<(const public_key &p1, const public_key &p2) { return memcmp(&p1, &p2, sizeof(public_key)) < 0; }
   inline bool operator>(const public_key &p1, const public_key &p2) { return p2 < p1; }
+  inline bool operator<(const key_image &p1, const key_image &p2) { return memcmp(&p1, &p2, sizeof(key_image)) < 0; }
+  inline bool operator>(const key_image &p1, const key_image &p2) { return p2 < p1; }
```
jeffro256, line 339: Looks good all around, but just personal preference:
```suggestion
  inline bool operator>(const key_image &p1, const key_image &p2) { return memcmp(&p1, &p2, sizeof(key_image)) > 0; }
```
There's a 90% chance this will get optimized out by the compiler, but since this is gonna likely take place many, many times a second , it might be better to write it out explicitly.

src/crypto/crypto-ops.h
```diff
@@ -161,5 +163,12 @@ void ge_sub(ge_p1p1 *r, const ge_p3 *p, const ge_cached *q);
 void fe_add(fe h, const fe f, const fe g);
 void fe_tobytes(unsigned char *, const fe);
 void fe_invert(fe out, const fe z);
+void fe_mul(fe out, const fe, const fe);
+void fe_0(fe h);
 
 int ge_p3_is_point_at_infinity_vartime(const ge_p3 *p);
+
+// miscellaneous
```
jeffro256, line 171: Since this is in an important file, I'd leave a more descriptive comment here, more or less what you put in the post, explaining why this needs to be exposed in this header file.

src/crypto/crypto.h
UkoeHB: The other operator is `inline`. If this is not inlined then idk what to say - compilers must be trash.

src/crypto/crypto-ops.h
UkoeHB: Oh looks like these are leftovers I'm not actually using. Will remove

jeffro256 (commented):
Looks good to me

jeffro256 (approved):

src/crypto/crypto.h
```diff
@@ -335,8 +335,16 @@ namespace crypto {
...
 }
 
+// type conversions for easier calls to sc_add(), sc_sub(), hash functions
+inline unsigned char* to_bytes(crypto::ec_scalar &scalar) { return &reinterpret_cast<unsigned char&>(scalar); }
```
vtnerd, line 343: These should be dropped in favor of `epee::to_byte_span()` usage. Or just redirect to that function and return a `span` here, although I don't see a purpose of re-creating those existing functions.

UkoeHB: If I use the span stuff directly, then:
```
sc_add(to_bytes(a), to_bytes(b), to_bytes(c));  //old
sc_add(epee::as_mut_byte_span(a).data(), epee::as_byte_span(b).data(), epee::as_byte_span(c).data());  //new
```

Using a `span` internally means you have to construct a `span` every time `to_bytes` is called, which is not ideal for one of the main use cases - high performance crypto algorithms.

src/crypto/crypto.h
vtnerd: This comment is old, but also the `inline` keyword is needed here or linker errors could occur.

src/crypto/crypto.h
vtnerd: The compiler should optimize away all references to the span if only the pointer is used. This is pretty basic stuff for the compiler actually.

I wasn't aware of the context though - these existing functions take a pointer with no other context anyway. My frustration with adding these functions is the the temptation to use these in non-crypto related functions. I guess reviewers will have to catch this.

UkoeHB: It's hard for me to imagine a case where `to_bytes()` would be used for anything other than exposing a raw pointer to C-style crypto functions.

jeffro256: I think I fall on the side of @UkoeHB here. Most of the underlying C crypto functions already use "unsafe" pointers insofar as the size of the area passed is implied by the contract of the function, not enforced at compile-time or runtime. Unless we want to rewrite these functions, using `epee::span` would just add an unused wrapper on top of a pointer which may or may not get optimized out by the compiler.

vtnerd (approved):

UkoeHB:
@vtnerd I squashed commits and added one thing. I need the method `rct::ecdhHash()`, so I renamed to `genAmountEncodingFactor()` and made it public. Better to shove that in this PR instead of making a whole new PR.

jeffro256 (approved):

vtnerd (approved):

PR #8673, opened 2022-12-11
removed definition of `refreshed` bool

cjac:
/usr/src/git/salsa/cryptocoin-team/monero/src/wallet/wallet2.cpp:3427:8: warning: variable ‘refreshed’ set but not used [-Wunused-but-set-variable]
 3427 |   bool refreshed = false;

Fails to build from source.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -3474,7 +3474,7 @@ void wallet2::refresh(bool trusted_daemon, uint64_t start_height, uint64_t & blo
   // "I had to reorder some code to fix... a timing info leak IIRC. In turn, this undid something I had fixed before, ... a subtle race condition with the txpool.
   // It was pretty subtle IIRC, and so I needed time to think about how to refix it after the move, and I never got to it."
   // https://github.com/monero-project/monero/pull/6097
-  bool refreshed = false;
+  bool refreshed;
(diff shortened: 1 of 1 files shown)


cjac:
18:29 < cjac> selsta: the FTBFS came from particularly strict gcc flags
18:29 < cjac> this is the debian policy, so it's keeping debian from building.  I 
              could commit this patch to debian, but I wanted to check with you 
              folks first.
18:30 < cjac> https://paste.debian.net/1263691/

cjac:
patch squashed so I can offer it to debian package maintainers, should this PR not be merged

UkoeHB:
This issue has appeared before, it is an intentional Chesterton's fence that @moneromooo-monero left in there.

jeffro256:
See conversation started here: https://github.com/monero-project/monero/pull/8359#discussion_r885652735

cjac:
patch offered to debian maintainers:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1025965

plowsof:
An update from the Debian maintainers: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1025927
```
There is no such compiler option added (-Werror?) and the package builds fine.
Please fix your build environment.
```

selsta:
Closing since we don't add this security flag anywhere and it seems debian also doesn't.

PR #8677, opened 2022-12-12
Refactored rx-slow-hash.c to make fast PoW verification possible

SChernykh:
Current `rx_slow_hash` code is too complex, and I tried to make it possible to use RandomX dataset for PoW verification a few times, but each time some bugs popped up. This is a rewrite that solves these issues and makes PoW verification faster in case `MONERO_RANDOMX_FULL_MEM` environment variable is set.

Full list of changes:

- Straight-forward call interface: `void rx_slow_hash(const char *seedhash, const void *data, size_t length, char *result_hash)`
- Consensus chain seed hash is now updated by calling `rx_set_main_seedhash` whenever a block is added/removed or a reorg happens
- `rx_slow_hash` will compute correct hash no matter if `rx_set_main_seedhash` was called or not (the only difference is performance), it was written to be robust (but `rx_set_main_seedhash` is called in all relevant places in this PR)
- New environment variable `MONERO_RANDOMX_FULL_MEM` to force use the full dataset for PoW verification (faster block verification)
- When dataset is used for PoW verification, dataset updates don't stall other threads (verification is done in light mode then)
- When mining is running, PoW checks now also use dataset for faster verification

@hyc @tevador


src/cryptonote_core/blockchain.cpp
```diff
@@ -4552,11 +4565,13 @@ bool Blockchain::handle_block_to_main_chain(const block& bl, const crypto::hash&
...
   }
 
-  send_miner_notifications(id, already_generated_coins);
+  const crypto::hash seedhash = get_block_id_by_height(crypto::rx_seedheight(new_height));
+  send_miner_notifications(new_height, seedhash, id, already_generated_coins);
 
   for (const auto& notifier: m_block_notifiers)
     notifier(new_height - 1, {std::addressof(bl), 1});
 
+  rx_set_main_seedhash(seedhash.data, std::thread::hardware_concurrency());
```
tevador, line 4574: Does this need a version check? It seems to unnecessarily initialize RandomX caches before the RandomX fork height.

tevador (commented):
I verified that the locking logic is correct.

src/cryptonote_basic/miner.cpp
```diff
@@ -30,6 +30,7 @@
 
 #include <sstream>
 #include <numeric>
+#include <thread>
```
hyc, line 33: probably should be using tools::get_max_concurrency() instead of using std::thread directly.

src/cryptonote_basic/miner.cpp
```diff
@@ -524,6 +524,8 @@ namespace cryptonote
   bool miner::worker_thread()
   {
     const uint32_t th_local_index = m_thread_index++; // atomically increment, getting value before increment
+    crypto::rx_set_miner_thread(th_local_index, std::thread::hardware_concurrency());
```
hyc, line 527: and again here, use tools::get_max_concurrency().

src/cryptonote_core/blockchain.cpp
```diff
@@ -30,6 +30,7 @@
 
 #include <algorithm>
 #include <cstdio>
+#include <thread>
```
hyc, line 33: Same comment, use tools::get_max_concurrency()

src/crypto/rx-slow-hash.c
```diff
@@ -191,187 +283,230 @@ typedef struct seedinfo {
...
+
+  char buf[HASH_SIZE * 2 + 1];
+
+  // Slow path (seedhash != main_seedhash, but seedhash == secondary_seedhash)
+  // Multiple threads can run in parallel in light mode, 10-15 ms per hash per thread
+  if (!secondary_cache) {
+    CTHR_RWLOCK_LOCK_WRITE(secondary_cache_lock);
+    if (!secondary_cache) {
+      hash2hex(main_seedhash, buf);
+      minfo(RX_LOGCAT, "RandomX new secondary seed hash is %s", buf);
```
jtgrassie, line 438: Only for printing, but shouldn't this be `hash2hex(seedhash, buf);`

src/crypto/rx-slow-hash.c
```diff
@@ -191,187 +283,230 @@ typedef struct seedinfo {
...
+  if (success) {
+    return;
+  }
+
+  // Slowest path (seedhash != main_seedhash, seedhash != secondary_seedhash)
+  // Only one thread runs at a time and updates secondary_seedhash if needed, up to 200-500 ms per hash
+  CTHR_RWLOCK_LOCK_WRITE(secondary_cache_lock);
+  if (!is_secondary(seedhash)) {
+    hash2hex(main_seedhash, buf);
+    minfo(RX_LOGCAT, "RandomX new secondary seed hash is %s", buf);
```
jtgrassie, line 466: Again, only for printing, but shouldn't this be `hash2hex(seedhash, buf);`

selsta:
@jtgrassie @hyc could you do another round of reviews or approve it?

CMakeLists.txt
```diff
@@ -791,7 +791,7 @@ else()
     set(USE_LTO_DEFAULT false)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--stack,10485760")
     if(NOT BUILD_64)
-      add_definitions(-DWINVER=0x0501 -D_WIN32_WINNT=0x0501)
+      add_definitions(-DWINVER=0x0600 -D_WIN32_WINNT=0x0600)
```
hyc, line 794: I'd prefer to see a separate commit and commit msg for this one

hyc:
Aside from one nit about CMakeLists.txt change, this all looks good to me. Builds cleanly, tests pass.

hyc (approved):

tevador (approved):

jtgrassie (approved):

PR #4219, opened 2018-08-03
Bp multi aggregation pippenger

moneromooo-monero:

PR #8679, opened 2022-12-14
[seraphis] ringct: adjust multiexp

UkoeHB:

- Refactors `pippenger_cached_data` to use a vector of `ge_cached` directly. This makes `pippenger_cached_data` reusable (I use it in a multiexp utility in the `seraphis_lib`).
- Exposes `ge_p3` versions of the multiexp functions for efficiency when desired.
- Replaces tabs with spaces in the multiexp notes, so they are legible in text editors that don't render tabs consistently (mine shows all the whitespace in the multiexp notes as single spaces).

### DIFF
--- a/src/ringct/multiexp.cc
+++ b/src/ringct/multiexp.cc
@@ -57,28 +57,28 @@ extern "C"
 //   1     1     52.8     70.4      70.2
 
 // Pippenger:
-// 	1	2	3	4	5	6	7	8	9	bestN
-// 2	555	598	621	804	1038	1733	2486	5020	8304	1
-// 4	783	747	800	1006	1428	2132	3285	5185	9806	2
-// 8	1174	1071	1095	1286	1640	2398	3869	6378	12080	2
-// 16	2279	1874	1745	1739	2144	2831	4209	6964	12007	4
-// 32	3910	3706	2588	2477	2782	3467	4856	7489	12618	4
-// 64	7184	5429	4710	4368	4010	4672	6027	8559	13684	5
-// 128	14097	10574	8452	7297	6841	6718	8615	10580	15641	6
-// 256	27715	20800	16000	13550	11875	11400	11505	14090	18460	6
-// 512	55100	41250	31740	26570	22030	19830	20760	21380	25215	6
-// 1024	111520	79000	61080	49720	43080	38320	37600	35040	36750	8
-// 2048	219480	162680	122120	102080	83760	70360	66600	63920	66160	8
-// 4096	453320	323080	247240	210200	180040	150240	132440	114920	110560	9
+//      1       2       3       4       5       6       7       8       9       bestN
+// 2    555     598     621     804     1038    1733    2486    5020    8304    1
+// 4    783     747     800     1006    1428    2132    3285    5185    9806    2
+// 8    1174    1071    1095    1286    1640    2398    3869    6378    12080   2
+// 16   2279    1874    1745    1739    2144    2831    4209    6964    12007   4
+// 32   3910    3706    2588    2477    2782    3467    4856    7489    12618   4
+// 64   7184    5429    4710    4368    4010    4672    6027    8559    13684   5
+// 128  14097   10574   8452    7297    6841    6718    8615    10580   15641   6
+// 256  27715   20800   16000   13550   11875   11400   11505   14090   18460   6
+// 512  55100   41250   31740   26570   22030   19830   20760   21380   25215   6
+// 1024 111520  79000   61080   49720   43080   38320   37600   35040   36750   8
+// 2048 219480  162680  122120  102080  83760   70360   66600   63920   66160   8
+// 4096 453320  323080  247240  210200  180040  150240  132440  114920  110560  9
 
-// 			2	4	8	16	32	64	128	256	512	1024	2048	4096
-// Bos Coster		858	994	1316	1949	3183	5512	9865	17830	33485	63160	124280	246320
-// Straus		226	341	548	980	1870	3538	7039	14490	29020	57200	118640	233640
-// Straus/cached	226	315	485	785	1514	2858	5753	11065	22970	45120	98880	194840
-// Pippenger		555	747	1071	1739	2477	4010	6718	11400	19830	35040	63920	110560
+//                2   4   8     16    32    64    128   256   512   1024  2048    4096
+// Bos Coster     858 994 1316  1949  3183  5512  9865  17830 33485 63160 124280  246320
+// Straus         226 341 548   980   1870  3538  7039  14490 29020 57200 118640  233640
+// Straus/cached  226 315 485   785   1514  2858  5753  11065 22970 45120 98880   194840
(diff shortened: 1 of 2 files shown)


src/ringct/multiexp.cc
```diff
@@ -700,8 +696,14 @@ rct::key pippenger(const std::vector<MultiexpData> &data, const std::shared_ptr<
...
   }
 
+  return result;
+}
+
+rct::key pippenger(const std::vector<MultiexpData> &data, const std::shared_ptr<pippenger_cached_data> &cache, const size_t cache_size, const size_t c)
+{
   rct::key res;
-  ge_p3_tobytes(res.bytes, &result);
+  const ge_p3 result_p3 = pippenger_p3(std::move(data), cache, cache_size, c);
```
jeffro256, line 705: Even though this compiles, the `std::vector<MultiexpData>` is copied here, not moved. I don't know if this was the intended consequence. This happens because `data` is a `const&`.

jeffro256: Since `pippenger_p3` takes ` const&`, you can just pass it as is

src/ringct/multiexp.cc
```diff
@@ -571,43 +577,33 @@ size_t get_pippenger_c(size_t N)
...
 }
 
 size_t pippenger_get_cache_size(const std::shared_ptr<pippenger_cached_data> &cache)
 {
-  return cache->size * sizeof(*cache->cached);
+  return cache->size() * sizeof(ge_cached);
 }
 
-rct::key pippenger(const std::vector<MultiexpData> &data, const std::shared_ptr<pippenger_cached_data> &cache, size_t cache_size, size_t c)
+ge_p3 pippenger_p3(const std::vector<MultiexpData> &data, const std::shared_ptr<pippenger_cached_data> &cache, size_t cache_size, size_t c)
```
jeffro256, line 602: This definition doesn't exactly match what's in the header because of the lack of `const` qualifiers, so it might produce warnings on some compilers.

src/ringct/multiexp.h
```diff
@@ -55,17 +61,19 @@ struct MultiexpData {
 };
 
 struct straus_cached_data;
-struct pippenger_cached_data;
+using pippenger_cached_data = std::vector<ge_cached, boost::alignment::aligned_allocator<ge_cached, 4096>>;
```
jeffro256, line 64: The old PR did aligned allocs as well. Do you know why?

jeffro256 (changes requested):
Looks like both `straus` and `pippenger` retain the exact same functionality . Would be nice to have some performance tests comparing the `_p3` versions to their counterparts to see the efficiency difference. Besides that, just one minor warning thing.

src/ringct/multiexp.h
UkoeHB: No clue, I just figured it's best to keep that in, in case it's somehow important.

src/ringct/multiexp.cc
UkoeHB: Good catch, thanks

UkoeHB:
You can check this perf test: https://github.com/monero-project/monero/blob/master/tests/performance_tests/ge_tobytes.h

It's not significant, but still nice to shave off easy wins like this (iirc sarang was a big fan of avoiding serialization/deserialization ops).

src/ringct/multiexp.h
jeffro256: I think @moneromooo-monero added this feature IIRC, so maybe we ask him

UkoeHB: Looks like it showed up in a commit called 'some speedups': https://github.com/monero-project/monero/pull/4219

SChernykh: 4096 is the virtual page size on x86, maybe this is why. It will only matter in low memory conditions when swapping starts - then the cache will fit in smaller number of pages. This alignment shouldn't affect performance otherwise.

src/ringct/multiexp.cc
```diff
@@ -571,43 +577,33 @@ size_t get_pippenger_c(size_t N)
...
   if (N == 0)
     N = data.size() - start_offset;
   CHECK_AND_ASSERT_THROW_MES(N <= data.size() - start_offset, "Bad cache base data");
-  std::shared_ptr<pippenger_cached_data> cache(new pippenger_cached_data());
+  std::shared_ptr<pippenger_cached_data> cache = std::make_shared<pippenger_cached_data>();
 
-  cache->size = N;
-  cache->cached = (ge_cached*)aligned_realloc(cache->cached, N * sizeof(ge_cached), 4096);
-  CHECK_AND_ASSERT_THROW_MES(cache->cached, "Out of memory");
+  cache->resize(N);
```
SChernykh, line 589: The old code had `CHECK_AND_ASSERT_THROW_MES(cache->cached, "Out of memory");` in case of allocation failure, this new code doesn't handle this case. `std::vector::resize` will throw `std::bad_alloc` on allocation failure, it should be handled here.

SChernykh: On the other hand, the only difference will be what type of assertion will be thrown here: `CHECK_AND_ASSERT_THROW_MES` throws `std::runtime_error` instead of `std::bad_alloc`. I couldn't find places in code that catch explicitly `std::runtime_error`, so it shouldn't matter.

SChernykh: Actually, I just found `catch (const std::runtime_error& e)` in `zmq_server.cpp` and `simplewallet.cpp`, so it makes sense to catch `std::bad_alloc` here and use the standard `CHECK_AND_ASSERT_THROW_MES`

UkoeHB: I'm 90% sure this is a `CHECK_AND_ASSERT_THROW_MES()` because A) `aligned_realloc` does not throw an exception and needs to be checked, B) @moneromooo-monero only uses `CHECK_AND_ASSERT_THROW_MES()` and would never write `throw std::bad_alloc;`. I have seen nowhere in the code base where bad allocs are rethrown as runtime errors.

SChernykh: `std::bad_alloc` is thrown in RandomX code. The other question is, should it be handled in the code at all, or just let everything crash. Probably the latter because out of memory situations are hard to recover from. I think it's fine to not catch it here.

moneromooo-monero: UkoeHB is right, I seldom send separate exceptions. Sometimes std::rnutime_error, that's it. I'm pretty sure I did not write code that relies on std::bad_alloc here (though of course who knows what anyone else might have done later on).

j-berman (approved):
Pointing out for other reviewers there's a nice unit test helping make sure `pippenger_cached_data` retains behavior here: https://github.com/monero-project/monero/blob/50aa0e8b7f11680be3954c176f2daa9ccf77b7dd/tests/unit_tests/multiexp.cpp#L235-L255

jeffro256 (approved):

Issue #8311, opened 2022-05-04
get_transactions returns as_hex: "" for miner transactions


kayabaNerve:
I'm currently running a local regtest network offering different responses for get_transactions based on whether or not the transaction in question is a miner transaction or not.

`curl http://127.0.0.1:18081/get_transactions -d '{"txs_hashes":["2d1fa93b2a76f3bfebf244df1f97368f5035350e8d8189560aa4308c0e10b3a9"]}' -H 'Content-Type: application/json'`

```
{
  "credits": 0,
  "status": "OK",
  "top_hash": "",
  "txs": [{
    "as_hex": "",
    "as_json": "",
    "block_height": 12519,
    "block_timestamp": 1651576837,
    "double_spend_seen": false,
    "in_pool": false,
    "output_indices": [12744],
    "prunable_as_hex": "",
    "prunable_hash": "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470",
    "pruned_as_hex": "02a36201ffe7610193c4a4b8ebe707022ae29fd80edc12030481348866ef8c3f70ec19376fe68b3495d903d51ac61d6a2401856a45ff8e753978324a17d6a3b9e1b6e016fe0fae7de9961b276e03dfcff9b102010000",
    "tx_hash": "2d1fa93b2a76f3bfebf244df1f97368f5035350e8d8189560aa4308c0e10b3a9"
  }],
  "txs_as_hex": [""],
  "untrusted": false
}
```

`curl http://127.0.0.1:18081/get_transactions -d '{"txs_hashes":["07c57cac19102310049264e2bd6056a801681839bca4d207f6f84573d4bcd073"]}' -H 'Content-Type: application/json'`

```
{
  "credits": 0,
  "status": "OK",
  "top_hash": "",
  "txs": [{
[...]

PR #8691, opened 2023-01-01
rpc: do not misidentify coinbase txes as pruned

moneromooo-monero:
There is potential for a double parsing of the tx base, which could be cached if it's found to be a performance issue later.


kayabaNerve:
Closes https://github.com/monero-project/monero/issues/8311.

jeffro256 (approved):
code looks good and I did functional tests to confirm functionality

PR #8692, opened 2023-01-01
epee: when loading a JSON RPC payload, ensure params is an object

moneromooo-monero:
this makes it easier to spot those mistakes by the caller

### DIFF
--- a/contrib/epee/include/net/http_server_handlers_map2.h
+++ b/contrib/epee/include/net/http_server_handlers_map2.h
@@ -152,6 +152,16 @@
        epee::serialization::store_t_to_json(static_cast<epee::json_rpc::error_response&>(rsp), response_info.m_body); \
        return true; \
     } \
+    epee::serialization::storage_entry se; \
+    if(ps.get_value("params", se, nullptr) && se.type() != typeid(epee::serialization::section)) \
+    { \
+       boost::value_initialized<epee::json_rpc::error_response> rsp; \
+       static_cast<epee::json_rpc::error_response&>(rsp).jsonrpc = "2.0"; \
+       static_cast<epee::json_rpc::error_response&>(rsp).error.code = -32701; \
+       static_cast<epee::json_rpc::error_response&>(rsp).error.message = "Params is not an object"; \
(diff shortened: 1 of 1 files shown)


plowsof:
```
curl http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"set_daemon","params": [{}]' -H 'Content-Type: application/json''
{
  "error": {
    "code": -32701,
    "message": "Params is not an object"
  },
  "id": 0,
  "jsonrpc": "2.0"
}
```

an empty list []

```
curl http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"set_daemon","params": []' -H 'Content-Type: application/json'
{
  "error": {
    "code": -38,
    "message": "Invalid ssl support mode"
  },
  "id": "0",
  "jsonrpc": "2.0"
}
```

moneromooo-monero:
The current parser makes no difference between an empty list and a missing field. It'll stay that way.

contrib/epee/include/net/http_server_handlers_map2.h
```diff
@@ -152,6 +152,16 @@
        epee::serialization::store_t_to_json(static_cast<epee::json_rpc::error_response&>(rsp), response_info.m_body); \
        return true; \
     } \
+    epee::serialization::storage_entry se; \
+    if(ps.get_value("params", se, nullptr) && se.type() != typeid(epee::serialization::section)) \
+    { \
+       boost::value_initialized<epee::json_rpc::error_response> rsp; \
```
vtnerd, line 158: `boost::value_initialized` isn't really needed anymore with C++11 brace initialization `{}`. Although it doesn't hurt, it's just wasted include times at this point. And changing to that style means all of the `static_cast`s below can be removed too.

vtnerd (approved):

contrib/epee/include/net/http_server_handlers_map2.h
moneromooo-monero: I'll leave it as is, that's what the surrounding code does and it is not buggy.

PR #8694, opened 2023-01-01
util: make GMT timestamps explicit for clarity

moneromooo-monero:
For privacy reasons, time functions use GMT, to avoid logs leaking timezones. It'd make more sense to use localtime for wallet output (which are not logged by default), but that adds inconsistencies which can also be confusing. So add a Z suffix for now to make it clear these are not local time.

### DIFF
--- a/src/common/util.cpp
+++ b/src/common/util.cpp
@@ -1067,7 +1067,7 @@ std::string get_nix_version_display_string()
     time_t tt = ts;
     struct tm tm;
     misc_utils::get_gmt_time(tt, tm);
-    strftime(buffer, sizeof(buffer), "%Y-%m-%d %H:%M:%S", &tm);
+    strftime(buffer, sizeof(buffer), "%Y-%m-%d %H:%M:%SZ", &tm);
     return std::string(buffer);
   }
 


selsta (approved):

jeffro256:
Since these timestamps are supposed to be be human readable, I think it would be more appropriate to end the timestamp with `UTC` or `GMT` spelled out versus `Z` b/c IMO it is more readable.

moneromooo-monero:
It's longer though. Opinions ? I'll change it to whatever consensus wants.

moneromooo-monero:
(within sane options)

selsta:
From what I have seen adding a Z suffix seems to be common, see for example: https://en.wikipedia.org/wiki/ISO_8601

jeffro256:
Yeah the Z suffix is standard, I just personally like the spelled out timezone better, but I'm fine with either.

charlestonxorg:
Agree

selsta:
I'd like to merge this with the `Z`, if people find it confusing we can still change it. Just adding `UTC` or `GMT` to all log messages feels a bit redundant.

jeffro256:
I'm fine with that

jeffro256 (approved):

Issue #8560, opened 2022-09-09
monerod fails to start after install or on windows startup when installed as windows service


zubair72:
Steps to reproduce error

 1.  Run windows elevated command
 2.  Switch to path where monerod.exe is located
 3.  Type monerod --install-service --data-dir DataLocation --other-command-line-options
 4.  After successful install service message appears try to start the service its fails to start
 5. If you mark the service to start at windows startup it does not start and and log error in windows system event

Note: Make sure before performing 5 steps monero-wallet-gui was not run and daemon was not started from monero-wallet-gui. If done so then issue does not appear.

Note 2: If monerod was run from elevated command prompt prior performing 5 steps issue does not appear again.

Summary: If monerod is run from monero-wallet-gui or from elevated command prompt at least once issue does not appear, as soon as you mark the service to start automatic, and restart windows you shall notice the service failed to start.

zubair72:
Anyone looking into this?

selsta:
Not yet, I'm not a Windows user myself, but I will ask someone who might look into it.

rbrunner7:
I tried to reproduce any of the issues and failed completely. Or, in other words, for me everything worked flawlessly throughout.

After installing I had no problem to start the service. Setting the service to *Automatic* and restarting Windows worked, the Monero service started without any problems. GUI wallet could connect to the daemon running as a service. All the service-controlling options like `--start-service` or `--stop-service` worked, from an elevated command prompt.

I use Windows 10 Pro, 21H2. I tested with the 64bit Windows installer version of the GUI wallet, with the daemon at version v0.18.1.0-release. I have no AV software running beyond Windows standard defender.

I don't have a good idea how to proceed from here, only the almost trivial question: Do you have AV software running that could interfere? And if yes, did you already try to take it out of the picture i.e. pause it or at least define exceptions for the Monero software?

0xFFFC0000:
The picture you posted just says the executable file is missing. Nothing more.

selsta:
Should be resolved by #8697 and #10335

PR #8697, opened 2023-01-02
win-service: enable auto startup

plowsof:
current behaviour:
1. from a terminal with admin rights `monerod.exe --data-dir G:\\ --install-service`
2. monerod is added to services but not running
3. `monerod.exe --start-service` - runs, and works perfectly
4. restart system - monerod service is "stopped"/not running after boot.

with [SERVICE_AUTO_START:](https://learn.microsoft.com/en-us/windows-hardware/drivers/install/inf-addservice-directive): 
4. restart system - monerod service is running as expected.

notes: (fixed by #8700 )
when service monerod is running, if you `monerod exit` the service gets uninstalled (?). so if e.g. the GUI implements this - clicking 'stop daemon' would uninstall the service, (run time flags will reveal that its in service mode)

### DIFF
--- a/src/daemonizer/windows_service.cpp
+++ b/src/daemonizer/windows_service.cpp
@@ -196,7 +196,7 @@ bool install_service(
       , 0
       //, GENERIC_EXECUTE | GENERIC_READ
       , SERVICE_WIN32_OWN_PROCESS
-      , SERVICE_DEMAND_START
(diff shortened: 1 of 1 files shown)


jeffro256 (approved):

selsta:
Closes #8560

PR #8698, opened 2023-01-02
Copyright: Update to 2023

mj-xmr:
Disclaimer for selsta:

No hidden dev fee was added to this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2014-2022, The Monero Project
+# Copyright (c) 2014-2023, The Monero Project
 #
 # All rights reserved.
 #
--- a/CMakeLists_IOS.txt
+++ b/CMakeLists_IOS.txt
@@ -1,4 +1,4 @@
-# Portions Copyright (c) 2017-2022, The Monero Project
+# Portions Copyright (c) 2017-2023, The Monero Project
 # This file is based off of the https://code.google.com/archive/p/ios-cmake/
 # It has been altered for Monero iOS development
 #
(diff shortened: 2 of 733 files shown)


plowsof:
during review i found 20 files that could be added to this PR. (ive made a PR to your branch here https://github.com/mj-xmr/monero-mj/pull/2) #8701 found them for me . otherwise lgtm

mj-xmr:
Good job!

iamamyth:
The following is not, and should not be construed as, legal advice, but I really do not understand the rationale for "change all copyright headers" PRs. At best, such bulk changes seem counterproductive, as the final year in the range of every header now reflects simply the current year (a fact totally divorced from the contents of the file), rather than the last year of significant modification (a fact related to the file itself, which, though possibly available via source code control, also reflects some author's judgment as to the materiality of a change). I also note that well-known open source projects, such as the Linux kernel (see, e.g., https://github.com/torvalds/linux/blob/master/kernel/cgroup/cgroup.c) do not follow this practice, and large companies, such as Google, also do not recommend this practice in their open source guidelines (https://opensource.google/documentation/reference/copyright). Surely polluting the repo history and spending time reviewing a change or tool for presumably legal reasons requires some affirmative legal justification, so, I ask: Has anyone received actual legal advice that such changes are productive? Or has this practice been proposed simply because others do it?

Notwithstanding the foregoing concerns, I suspect a "change all headers" tool could probably be repurposed to a "check for copyright year edits when editing a file, and warn if absent" tool, which seems maybe helpful to reviewers and contributors.

mj-xmr:
@iamamyth I would tend to agree with you.

I do it because of this:

> (...) Or has this practice been proposed simply because others do it?

and because I saw various contributors doing the same thing at the same time, each having their own sets of updated files, undeniably increasing the review effort, so I thought to have at least just one set, while they can focus on other stuff.

selsta (approved):

Issue #8647, opened 2022-11-19
No IPv6 addresses with print_cn


glv2:
The ```print_cn``` command on a monerod server lists empty fields instead of the IP addresses and ports for IPv6 connections.
Is it on purpose because these addresses can be long, or it is a bug?

moneromooo-monero:
Almost certainly a bug.

PR #8703, opened 2023-01-07
Show IPv6 addresses in connection list

glv2:
For issue #8647

### DIFF
--- a/src/daemon/rpc_command_executor.cpp
+++ b/src/daemon/rpc_command_executor.cpp
@@ -644,7 +644,14 @@ bool t_rpc_command_executor::print_connections() {
     }
   }
 
-  tools::msg_writer() << std::setw(30) << std::left << "Remote Host"
+  auto longest_host = *std::max_element(res.connections.begin(), res.connections.end(),
+                                        [](const auto &info1, const auto &info2)
+                                        {
+                                          return info1.address.length() < info2.address.length();
+                                        });
+  int host_field_width = std::max(15, 8 + (int) longest_host.address.length());
+
+  tools::msg_writer() << std::setw(host_field_width) << std::left << "Remote Host"
       << std::setw(8) << "Type"
       << std::setw(6) << "SSL"
       << std::setw(20) << "Peer id"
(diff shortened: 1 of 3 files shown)


jeffro256:
This PR makes RPC backwards-incompaible for code which assumes "non-empty ip field == IPv4 address". This may or may not be an issue which actually exists, but either way, at least the RPC minor version should be incremented to reflect the change in API.

src/daemon/rpc_command_executor.cpp
```diff
@@ -661,11 +661,18 @@ bool t_rpc_command_executor::print_connections() {
...
     std::string address = info.incoming ? "INC " : "OUT ";
-    address += info.ip + ":" + info.port;
+    if (((epee::net_utils::address_type) info.address_type) == epee::net_utils::ipv6_network_address::get_type_id())
+    {
+      address += "[" + info.ip + "]:" + info.port;
+    }
+    else
+    {
+      address += info.ip + ":" + info.port;
+    }
```
jeffro256, line 671:
```suggestion
    address += info.address;
```

Also handles Tor/I2P addresses

glv2:
Patch updated.
I hadn't seen this ```info.address``` variable that already has what we want and makes things simpler.
Thanks.

jeffro256 (approved):
Looks good!

src/daemon/rpc_command_executor.cpp
```diff
@@ -644,7 +644,7 @@ bool t_rpc_command_executor::print_connections() {
     }
   }
 
-  tools::msg_writer() << std::setw(30) << std::left << "Remote Host"
+  tools::msg_writer() << std::setw(74) << std::left << "Remote Host"
```
selsta, line 647: Would it be possible to keep it at 30 unless an IPv6 peer is connected? It makes the `print_cn` list look weird now and 99% of users won't have any IPv6 connections.

glv2: In fact, it's for i2p and onion addresses. They can be much longer than IPv6 addresses.
Is something relying on this field being fixed length? If not, we could make it variable length depending on the current peers' addresses.

selsta: >Is something relying on this field being fixed length

Not sure, I did some testing with the following code and it seemed to work well.

```cpp
int width = std::any_of(res.connections.begin(), res.connections.end(), [](const auto &info) {
  return (epee::net_utils::address_type)info.address_type == epee::net_utils::address_type::tor ||
         (epee::net_utils::address_type)info.address_type == epee::net_utils::address_type::i2p ||
         (epee::net_utils::address_type)info.address_type == epee::net_utils::address_type::ipv6;
}) ? 74 : 30;
```

glv2: I updated the patch to adapt the width of the host field to keep it as small as possible.  What do you think?

selsta (approved):

PR #8704, opened 2023-01-09
Output selection speedup for large amounts of outputs

moneromooo-monero:


src/wallet/wallet2.cpp
```diff
@@ -9336,17 +9336,22 @@ std::vector<size_t> wallet2::pick_preferred_rct_inputs(uint64_t needed_money, ui
...
+    if (!is_spent(td, false) && !td.m_frozen && td.is_rct() && is_transfer_unlocked(td) && td.m_subaddr_index.major == subaddr_account && subaddr_indices.count(td.m_subaddr_index.minor) == 1 && !td.m_key_image_partial)
+      if (td.amount() <= m_ignore_outputs_above && td.amount() >= m_ignore_outputs_below)
+        spendable[i] = true;
+  }
+
   // try to find a rct input of enough size
   for (size_t i = 0; i < m_transfers.size(); ++i)
   {
     const transfer_details& td = m_transfers[i];
-    if (!is_spent(td, false) && !td.m_frozen && td.is_rct() && td.amount() >= needed_money && is_transfer_unlocked(td) && td.m_subaddr_index.major == subaddr_account && subaddr_indices.count(td.m_subaddr_index.minor) == 1)
```
jeffro256, line 9343: The old and new conditions for this `if` branch are not the same. The old branch considered the transfer even with a partial key image, but this new branch does not (`... && !td.m_key_image_partial`). Is this intentional?

plowsof:
was messing around with this today, made around 4000 outputs from `transfer_split` and then performed a `sweep_all` without issue.

unrelated notes:
first transaction(s) from the `sweep_all` where included in this block https://community.rino.io/explorer/stagenet/block/1264338  and the final (30+ minutes later) included here https://community.rino.io/explorer/stagenet/block/1264349. 
I noticed that only 3 of these "146 in" transactions would be included in 1 block at a time, 

sometimes `transfer_split` complains e.g. you try and send 7000 tiny transactions to yourself - it works for several minutes and then errors "Transaction not possible. Available only 4.999659770000, transaction amount 4.953453912000 = 4.949663202000 + 0.003790710000 (fee)"

src/wallet/wallet2.cpp
moneromooo-monero: Yes.

moneromooo-monero:
Is this with the patch only, or did it sometimes also happen without the patch (given enough attempts to disregard chance) ?

plowsof:
sorry, unrelated to this patch / same behaviour. i cant reproduce the 'work for several minutes then error' - just getting an instant error about amounts now (i have less outputs to work with now). i think its just a limitation of  transfer_split

jeffro256:
Opened a PR against this branch here: https://github.com/moneromooo-monero/bitmonero/pull/7

moneromooo-monero:
< jeffro256[m]> moneromoooo How do you want to do the pick_preferred_rct_inputs PR? What if I add both commits you pushed, plus the patch I added on top for a total of 3 commits and you close the other PR? I just don't want 2 PRs open for the sake of the reviewers.

PR #8705, opened 2023-01-11
RandomX: print VM allocation warnings only once

SChernykh:
To reduce warning spam when large pages are unavailable.

### DIFF
--- a/src/crypto/rx-slow-hash.c
+++ b/src/crypto/rx-slow-hash.c
@@ -246,7 +246,11 @@ static void rx_init_full_vm(randomx_flags flags, randomx_vm** vm)
 
   *vm = randomx_create_vm((flags | RANDOMX_FLAG_LARGE_PAGES | RANDOMX_FLAG_FULL_MEM) & ~disabled_flags(), NULL, main_dataset);
   if (!*vm) {
-    mwarning(RX_LOGCAT, "Couldn't allocate RandomX full VM using large pages");
+    static int shown = 0;
+    if (!shown) {
+        shown = 1;
+        mwarning(RX_LOGCAT, "Couldn't allocate RandomX full VM using large pages (will print only once)");
+    }
     *vm = randomx_create_vm((flags | RANDOMX_FLAG_FULL_MEM) & ~disabled_flags(), NULL, main_dataset);
     if (!*vm) {
       merror(RX_LOGCAT, "Couldn't allocate RandomX full VM");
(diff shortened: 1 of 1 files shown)


selsta:
I'd prefer if this would get moved to info level log.

On macOS there is no way to remove this warning, we already use `VM_FLAGS_SUPERPAGE_SIZE_2MB` in RandomX which is the macOS equivalent for huge pages.

SChernykh:
> On macOS there is no way to remove this warning

I googled a bit and it seems you just need to to run with `sudo` or whatever it's called on macOS. Not a solution for everyone though.

src/crypto/rx-slow-hash.c
```diff
@@ -246,7 +246,11 @@ static void rx_init_full_vm(randomx_flags flags, randomx_vm** vm)
 
   *vm = randomx_create_vm((flags | RANDOMX_FLAG_LARGE_PAGES | RANDOMX_FLAG_FULL_MEM) & ~disabled_flags(), NULL, main_dataset);
   if (!*vm) {
-    mwarning(RX_LOGCAT, "Couldn't allocate RandomX full VM using large pages");
+    static int shown = 0;
+    if (!shown) {
+        shown = 1;
+        mwarning(RX_LOGCAT, "Couldn't allocate RandomX full VM using large pages (will print only once)");
+    }
```
jeffro256, line 253:
```suggestion
#ifndef __APPLE__
    static int shown = 0;
    if (!shown) {
        shown = 1;
        mwarning(RX_LOGCAT, "Couldn't allocate RandomX full VM using large pages (will print only once)");
    }
#endif
```

src/crypto/rx-slow-hash.c
```diff
@@ -269,7 +273,11 @@ static void rx_init_light_vm(randomx_flags flags, randomx_vm** vm, randomx_cache
 
   *vm = randomx_create_vm((flags | RANDOMX_FLAG_LARGE_PAGES) & ~disabled_flags(), cache, NULL);
   if (!*vm) {
-    mwarning(RX_LOGCAT, "Couldn't allocate RandomX light VM using large pages");
+    static int shown = 0;
```
jeffro256, line 276: same here

jeffro256 (changes requested):

SChernykh:
@selsta @jeffro256 I combined both your proposals in the second commit. This should work well for everyone.

hyc:
Are you sure MacOS hugepages aren't usable? They should still be supported on x86, even though there's no support on aarch64.

SChernykh:
We do use `VM_FLAGS_SUPERPAGE_SIZE_2MB` in virtual_memory.c for RandomX, so it should work on MacOS when run with sudo, but @selsta says it doesn't work. I don't have anything from Apple, so I can't check.

selsta:
I tested on aarch64, will check again with x86.

hyc:
Yeah, there are docker images to boot MacOS on a regular PC, but they only work with actual Intel CPUs, compatible to the models Apple used. I only have AMD CPUs here.

plowsof:
can we have this at log level 1 please, this caused me emotional distress because i thought my monerod was secretly mining and when e.g. running blockchain-stats i see the warning in the output:
```
# DATA
Date	Blocks/day	Blocks	Txs/Day	Txs	Bytes/Day	Bytes
2023-01-17 10:45:35.798	W Couldn't allocate RandomX cache using large pages
2023-01-12	423	2798323	10478	10478	26770465	26770465
2023-01-13	688	2799011	15464	25942	38953621	65724086
2023-01-14	774	2799785	15034	40976	38456559	104180645
2023-01-15	684	2800468	14076	55052	35789881	139970526
```

SChernykh:
I've changed large page allocation messages back to `mdebug` (the same as it was before the refactoring in #8677)

selsta (approved):

PR #8707, opened 2023-01-14
verRctNonSemanticsSimpleCached: fix fragility

jeffro256:
edit: this is now also a fix PR

Asserts serialization precondition of `verRctNonSemanticsSimpleCached` as mentioned [here](https://github.com/monero-project/monero/pull/8675#discussion_r1045782676), [here](https://github.com/monero-project/monero/pull/8675#discussion_r1053636419), and [here](https://github.com/monero-project/monero/pull/8675#issuecomment-1379738083).

edit: above assumption was found to be false. This PR fixes the issue with semantics differing from non-cached version where it was thought to be the same.

@vtnerd @SChernykh @kayabaNerve


tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+#define CACHED_SERIALIZATION_ASSUMPTION_SUBTEST(fieldmodifyclause)             \
+    do {                                                                       \
+        rct::rctSig modded_sig = get_original_sig();                           \
+        modded_sig.fieldmodifyclause;                                          \
+        const std::string mod_sig_hex = rct_sig_to_bin_hex_string(modded_sig); \
+        EXPECT_NE(original_sig_hex(), mod_sig_hex);                            \
+    } while (0);                                                               \
+
+namespace
```
UkoeHB, line 51: Instead of an anonymous `namespace` I'd encourage the `static` keyword to promote local reasoning (namespaces are less local than keywords). Not to mention all that added whitespace from the indent... (indenting namespaces is a huge pet peeve for me).

UkoeHB, line 51: Instead of an anonymous `namespace` I'd encourage the `static` keyword to promote local reasoning (namespaces are less local than keywords). Not to mention all that non-functional whitespace from the indent... (indenting namespaces is a huge pet peeve for me).

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,220 @@
...
+                "decoding mask from hex did not work"
+            );
+
+            ring0.push_back({dest, mask});
+        }
+
+        return {ring0};
+    }
+
+    const cryptonote::transaction& get_original_transaction()
```
UkoeHB, line 116: Does this need to be a pseudo-singleton? Note that this design locks you into a specific filename. Would be simpler and more future-proof to write a function that takes the file name for the transaction and returns the transaction. And separate functions that take a transaction and return the signatures and signatures hex. At most the additional cost is doing all this once_flag stuff in each `TEST()` case instead of once globally.

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,220 @@
...
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include <sstream>
+#include <mutex>
+
+#define IN_UNIT_TESTS // to access Blockchain::expand_transaction_2
```
UkoeHB, line 32: All I can say is... incredible that this exists for this situation.

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    // For all values of S and S' where S and S' are of type rct::rctSig and S =/= S',
+    // ::do_serialize(S) =/= ::do_serialize(S') must hold.
+    // In other words, ::do_serialize() must create a unique representation of rct::rctSig
+
+    // These are the subtests most likely to fail. Fields 'message' and 'mixRing' are not serialized
+    // when sent over the wire, since they can be reconstructed from transaction data.
+    // How signatures are serialized for the blockchain can be found in the methods
+    // rct::rctSigBase::serialize_rctsig_base and rct::rctSigPrunable::serialize_rctsig_prunable.
+    CACHED_SERIALIZATION_ASSUMPTION_SUBTEST(message.bytes[31]++)
```
UkoeHB, line 208: Not too excited about using a macro, at least what you are doing is simple enough. The main alternative that springs to mind is using a lambda passed to a helper function that replaces this macro.

kayabaNerve:
Thanks for adding this :)

As someone who doesn't focus on C++, my one note is key image edits are only indirectly tested via the message test. Ideally, that'd also be explicitly tested.

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: One big benefit of using a macro here is that when/if the unit test fails, we get the line number of that specific instance of the test, instead of the same line number of the `EXPECT_NE` inside the helper function.

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: `expand_transaction_from_bin_file_and_pubkeys` should be the change you're looking for. No more singleton, what do you think?

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: Removed the nonfunction whitespace and put `static` before the functions

jeffro256:
@kayabaNerve you bring up an interesting point. I found a test case which fails so I'll push it.

jeffro256:
So turns out, since the field `clsag::I` is not serialied, `::do_serialize(S) =/= ::do_serialize(S') when S =/= S'` is NOT guaranteed to hold. The behavior of `verRctNonSemanticsSimpleCached` is NOT equal to `verRctNonSemanticsSimple` is all cases.

kayabaNerve:
I don't believe master is broken because we hash the key image into the message, yet if we didn't, this would've enabled a double spend on mainnet.

Do we want to revert verRctNonSemanticsSimpleCached, explicitly have Cached additionally hash the key images, modify do_serialize (if we can), or accept this fragile indirection?

SChernykh:
@jeffro256 This is why it has this code:
```
      if (rv.type != RCTTypeBulletproofPlus)
        return verRctNonSemanticsSimple(rv);
```
and this comment:
```
      // RCT cache assumes that this function will serialize and hash all rv's fields used for RingCT verification
      // If you're about to add a new RCTType here, first you must check that binary_archive serialization writes all rv's fields to the binary blob
      // If it's not the case, rewrite this function to serialize everything, even some "temporary" fields which are not serialized normally
```
All because I noticed this thing with CLSAG too when I prepared that PR.

jeffro256:
> This is why it has this code:

This does not mitigate the issue mentioned. It will prevent future signature types from passing the test unintentionally, but it does not prevent an otherwise correct signature with modified key images from passing the test.

> `... you must check that binary_archive serialization writes all rv's fields to the binary blob`

This condition does not hold, so the semantics now differ between functions.

SChernykh:
> correct signature with modified key images from passing the test

But `RCTTypeBulletproofPlus` serializes key images, how it can happen? If you say that semantics now differs, prove it with a unit test that makes **current** code fail:
- calls `verRctNonSemanticsSimple`
- changes data
- calls `verRctNonSemanticsSimple` and gets a different result

kayabaNerve:
It doesn't, at least not via `do_serialize`. That's what this failing test case proves.

jeffro256:
> If you say that semantics now differs, prove it with a unit test that makes current code fail

Check latest commit

SChernykh:
So BulletproofPlus tx **does** have non-empty CLSAGs vector. Ok. What I suggest is add the serialization of all missing fields to the cache version. I can prepare a patch and you can add it to this PR - it makes sense to have a single PR for tests and fixes.

kayabaNerve:
https://github.com/monero-project/monero/pull/8675#issuecomment-1382701559

@jeffro256 I'd personally appreciate you make this edit in this PR/a linked followup.

tests/unit_tests/verRctNonSemanticsSimple.cpp
UkoeHB: It looks better but what is the point of keeping the anonymous namespace in there?

tests/unit_tests/verRctNonSemanticsSimple.cpp
UkoeHB: What you can do is return `bool` on your test helper function, then `EXPECT_TRUE()` in your `TEST()`.

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,248 @@
...
+    // Uncomment line below to sanity check CACHED_SERIALIZATION_ASSUMPTION_SUBTEST
+    // CACHED_SERIALIZATION_ASSUMPTION_SUBTEST(message) // should fail
+}
+
+TEST(verRctNonSemanticsSimple, cached_is_noncached)
+{
+    const cryptonote::transaction& tx = expand_transaction_from_bin_file_and_pubkeys
+        (tx1_file_name, tx1_input_pubkeys);
+    const rct::rctSig& rs = tx.rct_signatures;
+    const std::string rs_hex = rct_sig_to_hex_hash(rs);
```
UkoeHB, line 237: unused variable

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,248 @@
...
+/**
+ * @brief Mimic how rctSig's are serialized and hashed in verRctNonSemanticsSimpleCached
+ *
+ * We convert the hash into a hex string though for ease of use.
+ * Check the body of verRctNonSemanticsSimpleCached to make sure this matches.
+ *
+ * @param rs ring signature
+ * @return std::string hash of serialization of the rctSig
+ */
+static std::string rct_sig_to_hex_hash(const rct::rctSig& rs)
```
UkoeHB, line 86: Afaict this function can just return a `crypto::hash` (`get_rct_sig_hash()`).

UkoeHB (commented):
This PR should not be merged until `verRctNonSemanticsSimpleCached` is updated to include the transaction hash, which is a much more robust way of tracking validation results than bespoke hashing of the signature pieces.

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: Yeah it can, but I wanted to it be easier for humans to read. The hex representation is 1-to-1 so it shouldn't affect any outcomes.

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,248 @@
...
+TEST(verRctNonSemanticsSimple, tx_preconditions_1)
+{
+    // If this unit test fails, something changed about transaction deserialization / expansion or
+    // something changed about RingCT signature verification.
+
+    const cryptonote::transaction& tx = expand_transaction_from_bin_file_and_pubkeys
+        (tx1_file_name, tx1_input_pubkeys);
+    const rct::rctSig& rs = tx.rct_signatures;
+
+    ASSERT_EQ(1, tx.vin.size());
```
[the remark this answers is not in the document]
jeffro256: done

jeffro256:
@UkoeHB would you like to double check the changes to `verRctNonSemanticsSimpleCached`?

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: Avoiding linking conflicts and not polluting global namespace for tools like Doxygen

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: Especially useful in failure cases.

tests/unit_tests/verRctNonSemanticsSimple.cpp
UkoeHB: This is better thanks

tests/unit_tests/verRctNonSemanticsSimple.cpp
UkoeHB: In what circumstance is a human reading this and learning anything by looking at the bytes? Either the hashes match or they don't.

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: This is true, but it gets more complicated with `EXPECT_EQ`, `EXPECT_NE`, etc...

tests/unit_tests/verRctNonSemanticsSimple.cpp
UkoeHB: 1) linking conflicts mean there is a code smell (like introducing non-static functions into the global namespace, or badly named functions)
2) why would Doxygen be scraping `.cpp` files?

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: Good point

src/ringct/rctSigs.cpp
```diff
@@ -1600,17 +1603,24 @@ namespace rct {
 
       ::do_serialize(ar, const_cast<rctSig&>(rv));
 
-      crypto::hash h;
-      cryptonote::get_blob_hash(ss.str(), h);
+      const std::string ser_sig = ss.str();
+      const crypto::hash tx_hash = cryptonote::get_transaction_hash(tx);
+
+      crypto::hash sig_and_tx_hash;
+      KECCAK_CTX kctx; // not using cn_fast_hash here because we need 2 updates
```
SChernykh, line 1610: You could just do
```
ser_sig.append(reinterpret_cast<const char*>(&tx_hash), sizeof(tx_hash));
```
and then call `get_blob_hash` as usual. Keeping all keccak code in one place is better.

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,287 @@
...
+ *
+ * The purpose of this macro is to assert that changes in key images, while not necessarily changing
+ * serialization of rctSig's, will still cause verRctNonSemanticsSimpleCached to fail.
+ *
+ * @param modifyf Takes a cryptonote::transaction& and modifies it
+ */
+#define CACHED_IS_NON_CACHED_SUBTEST(og_tx, input_pubkeys, modifyf)                          \
+    do {                                                                                     \
+        EXPECT_TRUE(rct::verRctNonSemanticsSimple(og_tx.rct_signatures));                    \
+        EXPECT_TRUE(rct::verRctNonSemanticsSimpleCached(og_tx));                             \
```
SChernykh, line 70: `verRctNonSemanticsSimpleCached` goes through different code paths the first and subsequent times it encounters a tx, so it should be called two times in a row everywhere to be sure.

src/ringct/rctSigs.cpp
jeffro256: Yes, the only problem is that this might waste an extra allocation. `keccak_init`, `keccak_update`, and `keccak_finish` are used other places in the codebase besides here, `wallet2.cpp` for instance. I also added a unit test to confirm that `cn_fast_hash` uses keccak.

SChernykh: Yes, this code if good enough as is. Also, we shouldn't really care that `cn_fast_hash` is keccak because keccak here is not exposed to the calling code and can basically use any cryptographically secure hash here.

jeffro256: Yeah I just added it in case we migrate away from keccak at any point, a unit test will remind us that we can also upgrade this hashing code instead of us having to remember.

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: done

src/ringct/rctSigs.cpp
UkoeHB: I'd like to discourage non-default use of crypto library functions unless there is a good reason. In this case, just reserve capacity in the string before appending anything, then call to `cn_fast_hash()` like normal.

src/ringct/rctSigs.cpp
```diff
@@ -1600,17 +1602,21 @@ namespace rct {
 
       ::do_serialize(ar, const_cast<rctSig&>(rv));
 
-      crypto::hash h;
-      cryptonote::get_blob_hash(ss.str(), h);
+      std::string ser_sig_and_txhash = ss.str();
+      const crypto::hash tx_hash = cryptonote::get_transaction_hash(tx);
+      ser_sig_and_txhash.append(tx_hash.data, sizeof(crypto::hash));
```
UkoeHB, line 1607: Wow ok so this does require two allocations without some fancy stringstream ops to copy the underlying stringstream buffer, that's too bad.

You need to include a domain separator:

```
cryptonote_config.h

const constexpr char HASH_KEY_CONTEXTUAL_TX_VALIDATION_ID[] = "contextual_tx_validation_id";
```

```
in this file:

std::string ser_sig_and_txhash;
ser_sig_and_txhash.reserve(sizeof(config::HASH_KEY_CONTEXTUAL_TX_VALIDATION_ID) + ss.tellp() + sizeof(crypto::hash));
ser_sig_and_txhash.append(config::HASH_KEY_CONTEXTUAL_TX_VALIDATION_ID);
ser_sig_and_txhash.append(ss.str());
ser_sig_and_txhash.append(tx_hash.data, sizeof(crypto::hash));
```

tests/unit_tests/verRctNonSemanticsSimple.cpp
UkoeHB: In those cases you just use an assert with clear message, then `EXPECT_NO_THROW()` in the test body.

src/ringct/rctSigs.cpp
SChernykh: What's the point of a domain separator in a data structure that's not mixed with any other hashes anywhere? At this point it's "do everything by the book" just for the sake of doing by the book. Domain separator is unnecessary here.

src/ringct/rctSigs.cpp
```diff
@@ -1579,8 +1580,10 @@ namespace rct {
       }
     }
 
-    bool verRctNonSemanticsSimpleCached(const rctSig & rv)
+    bool verRctNonSemanticsSimpleCached(const cryptonote::transaction & tx)
```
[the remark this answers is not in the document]
SChernykh: This is a bugfix PR, not a refactoring PR. I agree it should be rewritten, just not in this PR.

src/ringct/rctSigs.cpp
UkoeHB: Doesn't matter if it's unnecessary, do it anyway.

SChernykh: Explain why? The code gets fatter and more complicated. I'm against this change.

UkoeHB: It's a policy for higher-quality cryptographic code.

SChernykh: I know what it is and why. Anyway, after the latest commit it's much cleaner and just 1 line of code, so good enough.

UkoeHB: The transcript utility I wrote for seraphis requires a domain separator in the constructor, to make it extra clear when **you're doing it wrong**.

jeffro256: Yes, check the latest commit. Should be a good compromise. All allocations happen in `std::stringstream`, and we use best practices of domain separation.

src/ringct/rctSigs.cpp
```diff
@@ -1579,13 +1579,31 @@ namespace rct {
...
+     * tx.rct_signatures.bulletproofs_plus[x].V should match values tx.rct_signatures_outPk
+     * (multiplied by group inverse).
+     *
+     * @param tx transaction as seen externally which contains signature (with BC data) to verify
+     * @return true if test passes
+     * @return false if test does not pass
+     */
+    bool verRctNonSemanticsSimpleCached(const cryptonote::transaction & tx)
     {
+      const rctSig & rv = tx.rct_signatures;
```
UkoeHB, line 1597: It looks like this function expects that `rv` was constructed by a call to `expand_transaction_2()` or equivalent. As a sanity check you could add an assert that `!(rv.message == rct::zero())`. It doesn't guarantee anything, but would make tracking down errors easier if this function is misused.

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+/**
+ * @brief Used in "cached_is_noncached" to test for cache robustness against key image modifications
+ *
+ * The purpose of this macro is to assert that changes in key images, while not necessarily changing
+ * serialization of rctSig's, will still cause verRctNonSemanticsSimpleCached to fail.
+ *
+ * @param modifyf Takes a cryptonote::transaction& and modifies it
+ */
+#define CACHED_IS_NON_CACHED_SUBTEST(og_tx, input_pubkeys, modifyf)                          \
```
UkoeHB, line 67: `modifyf` looks like a typo, maybe `modify_func` would work better (that's the pattern I use).

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,265 @@
...
+    // when sent over the wire, since they can be reconstructed from transaction data. Many images
+    // in prunable sig data e.g. 'p.CLSAGs[x].I' are not serialized as well for the same reason. 
+    // How signatures are serialized for the blockchain can be found in the methods
+    // rct::rctSigBase::serialize_rctsig_base and rct::rctSigPrunable::serialize_rctsig_prunable.
+    CACHED_SERIALIZATION_ASSUMPTION_SUBTEST(message.bytes[31]++)
+    CACHED_SERIALIZATION_ASSUMPTION_SUBTEST(mixRing.push_back({}))
+    CACHED_SERIALIZATION_ASSUMPTION_SUBTEST(mixRing[0].push_back({}))
+    CACHED_SERIALIZATION_ASSUMPTION_SUBTEST(mixRing[0][8].dest[10]--)
+    CACHED_SERIALIZATION_ASSUMPTION_SUBTEST(mixRing[0][15].mask[3]--)
+    //CACHED_SERIALIZATION_ASSUMPTION_SUBTEST(p.CLSAGs[0].I[2] += 101)
```
UkoeHB, line 242: Uncomment this now that it's fixed?

jeffro256: It's not "fixed" really since right now it's still just testing `do_serialize` uniqueness. I think I'll have both a `do_serialize` uniqueness test and a full implementation mimic uniqueness test

kayabaNerve:
The new impl looks good to me. I think koe addressed a lot of commentary.

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: Okay I separated tests for `do_serialize()` uniqueness from the whole sig + tx uniqueness.

src/ringct/rctSigs.cpp
```diff
@@ -1579,13 +1579,31 @@ namespace rct {
...
+
       // Hello future Monero dev! If you got this assert, read the following carefully:
       //
-      // RCT cache assumes that this function will serialize and hash all rv's fields used for RingCT verification
-      // If you're about to add a new RCTType here, first you must check that binary_archive serialization writes all rv's fields to the binary blob
-      // If it's not the case, rewrite this function to serialize everything, even some "temporary" fields which are not serialized normally
+      // RCT cache assumes that the the transaction passed is the one in which the ring signature
```
iamamyth, line 1601: Some grammar and clarity issues in this comment. Would suggest:
```
// ::do_serialize(rct::rctSig) omits certain transaction-linked rct::rctSig
// fields which affect the semantics of the underlying signature. Therefore,
// the transaction hash must be included in the cache key so that caching
// a rct::rctSig, modifying its transaction-linked fields, and then performing
// a cache lookup on the modified signature, does not yield a false cache hit. 
```

src/cryptonote_core/blockchain.h
```diff
@@ -1581,7 +1581,7 @@ namespace cryptonote
      * can be reconstituted by the receiver. This function expands
      * that implicit data.
      */
-    bool expand_transaction_2(transaction &tx, const crypto::hash &tx_prefix_hash, const std::vector<std::vector<rct::ctkey>> &pubkeys) const;
+    static bool expand_transaction_2(transaction &tx, const crypto::hash &tx_prefix_hash, const std::vector<std::vector<rct::ctkey>> &pubkeys);
```
iamamyth, line 1584: Seems strange to have a static function in a header file

jeffro256: Should be a static method of `class Blockchain`, no?

src/ringct/rctSigs.cpp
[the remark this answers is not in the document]
jeffro256: Actually since we confirmed that the behavior did not change, why don't we refactor it right now?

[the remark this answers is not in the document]
UkoeHB: This PR is already getting quite long, best to keep it scoped to 'unit test + bug fix' and follow up with a refactor (refactors should generally not change existing behavior, only interfaces and organization of code).

jeffro256:
Okay, I'm sorry to everyone who's reviewed thus far, but I've made the (hopefully) last change to the implementation. I think I'm finally happy with the blend of robustness and performance. This change makes a couple of abstractions bleed, but I'll release a refactoring PR later for that. Here's how the new implementation works:

* The function signature was changed to: `bool verRctNonSemanticsSimpleCached(cryptonote::transaction & tx, const rct::ctkeyM & mix_ring, const crypto::hash & tx_prefix_hash);`
* cache hash now = `H("txhash_and_mixring" || tx hash || mix ring blob)`. This is theoretically faster than original RCT cache PR because the tx hash is cached and the mix ring blob is much smaller than the blob of the whole signatures structure.
* `expand_transaction_2` is called inside `verRctNonSemanticsSimpleCached`, not before. So calling with an expanded transaction is not a precondition. One less footgun
* `expand_transaction_2` is only called when cache miss occurs. This increases performance because we can avoid a lot of copying and allocation.

One big note: code was removed in `blockchain.cpp` to check expansion postconditions after call to `expand_transaction_2` and before call to `verRctNonSemanticsSimpleCached`. This might make some people uneasy, as there are now fewer runtime checks for `expand_transaction_2`. As the current implementation stands, these checks cannot exist because the transaction is sometimes not expanded, so the checks will fail superfluously. I plan to add unit tests for `expand_transaction_2`.

There are two main options for those who are worried about the above change. 1) Always expand and always do the checks or 2) have `verRctNonSemanticsSimpleCached` return whether or not it did full verification or got a cache hit and do post-expansion checks in "blockchain.cpp` conditionally or 3) option #2 but do it inside `verRctNonSemanticsSimpleCached`.

src/ringct/rctSigs.cpp
jeffro256: This issue is resolved with the latest change to implementation

jeffro256:
(More unit tests to come though)

iamamyth:
It looks to me as though there's a middle ground between "create an RCT cache data structure which resides in the Blockchain instance (proposed by UkoeHB on the initial verRctNonSemanticsSimpleCached PR and deferred for the time being)" and "pretend verRctNonSemanticsSimpleCached belongs in the rct source tree, rather than the Blockchain implementation". In other words, I think you can fix a lot of visibility and safety issues in this PR by moving verRctNonSemanticsSimpleCached to Blockchain, as it almost certainly belongs there (as revealed by the data access issues and unfortunate data/call flow problems), which should also simplify the introduction of a dedicated, Blockchain instance-scoped RCT cache in a future PR.

src/cryptonote_core/blockchain.h
iamamyth: See https://github.com/monero-project/monero/pull/8707#issuecomment-1385838246, the various declaration changes suggest an ownership/scoping problem which should be fixed in this PR.

jeffro256:
Which "safety" issues are you referring to? I agree the function should be moved, but this PR is meant for the bug fix + unit tests. We can refactor in a later PR. However, if the reviewers are willing to do review the refactored version, I can move it in this PR. I was planning to make `verRctNonSemanticsSimpleCached` a private static method of `Blockchain`.

iamamyth:
I suggest thinking in terms of logical changesets. The PR which introduced `verRctNonSemanticsSimpleCached` relied on the faulty assumption that a suitable RCT cache key could be constructed independent of the transaction context, and, therefore, the function was placed in the rct namespace. This PR/changeset aims to correct that logical error, which introduces a transaction dependency, implying the function should move. Trying to split the changeset across a non-logical boundary introduces logical difficulties (private methods becoming public, unclear check_tx_inputs verification policy -- the "safety" issue I mentioned, etc). But, the cache can still remain static, as changing the storage layout has no relation to the purpose of this changeset.

tests/unit_tests/verRctNonSemanticsSimple.cpp
jagerman: > Avoiding linking conflicts and not polluting global namespace for tools like Doxygen

You won't get linking conflicts: anonymous namespace and `static` do exactly the same thing (for free functions).

jeffro256:
Okay well since `verRctNonSemanticsSimpleCached` made it into release, and since I changed most of the implementation anyways, I figured that I also combine the refactoring into this PR as well (which is just moving it to the `Blockchain` class). To review the changes made here:
* Function is moved to `Blockchain` class.
* The functions takes a mutable transaction, transaction prefix hash, and mix ring as arguments instead of just the RCT signature.
* The cache hash is changed to a domain separated hash as a function of the transaction hash (full hash not just prefix) and a hash of the mixring blob. This is faster and more complete  than a rct blob hash since the transaction hash is cached and the transaction hash contains all information that is provided by a user. The mixring blob captures all blokchain data referenced indirectly by the transaction.
* Unit tests

src/cryptonote_core/blockchain.cpp
```diff
@@ -3298,6 +3299,71 @@ bool Blockchain::expand_transaction_2(transaction &tx, const crypto::hash &tx_pr
...
+  crypto::hash tx_and_mixring_hash;
+  cryptonote::get_blob_hash(ss.str(), tx_and_mixring_hash);
+
+  static tools::data_cache<crypto::hash, 8192> cache;
+
+  if (cache.has(tx_and_mixring_hash))
+    return true;
+
+  // We had a cache miss, so now we must expand the mix ring and do full verification
+  return expand_tx_and_ver_rct_non_sem(tx, tx_prefix_hash, mix_ring);
```
SChernykh, line 3364: Wait a minute, I don't see where you add successful results to the cache? Where is `cache.add`?

SChernykh (changes requested):

src/cryptonote_core/blockchain.cpp
jeffro256: That's a great point. Removed it for debugging and forgot to add it back in haha. Check latest commits

jeffro256:
I also added the post-expansion checks back in for all cases

src/cryptonote_core/blockchain.cpp
```diff
@@ -3298,6 +3299,138 @@ bool Blockchain::expand_transaction_2(transaction &tx, const crypto::hash &tx_pr
...
+  // the transaction. Two, we take a hash of all *previously validated* blockchain data
+  // referenced by this transaction which is required to validate the ring signature. In our
+  // case, this is the mixring. Future versions of the protocol may differ in this regard, but
+  // if this assumptions holds true in the future, enable the RCT cache by modifying the first
+  // two asserts, and the `rv.type != RCTTypeBulletproofPlus` "if" branch below.
+  CHECK_AND_ASSERT_MES(tx.version <= 2, false, "Unknown TX type. Make sure RCT cache works correctly with this type and then enable it in the code here.");
+  CHECK_AND_ASSERT_MES(rv.type <= rct::RCTTypeBulletproofPlus, false, "Unknown RCT type. Make sure RCT cache works correctly with this type and then enable it in the code here.");
+  CHECK_AND_ASSERT_MES(!tx.pruned, false, "Pruned transaction will not pass verRctNonSemanticsSimple");
+
+  // Do RCT expansion, then do pos-expansion sanity checks, then do full non-semantics verification.
```
iamamyth, line 3321: Typo: `pos-expansion` => `post-expansion`

src/cryptonote_core/blockchain.cpp
```diff
@@ -3298,6 +3299,138 @@ bool Blockchain::expand_transaction_2(transaction &tx, const crypto::hash &tx_pr
...
+  binary_archive<true> ar(ss);
+  ::do_serialize(ar, const_cast<rct::ctkeyM&>(mix_ring));
+
+  // Calculate hash of TX hash and mix ring blob
+  crypto::hash tx_and_mixring_hash;
+  cryptonote::get_blob_hash(ss.str(), tx_and_mixring_hash);
+
+  static tools::data_cache<crypto::hash, 8192> cache;
+
+  // Search cache for successful verifcation of same TX + mix ring combination
```
iamamyth, line 3420: Typo: `verifcation` => `verification`

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,434 @@
...
+    T modded_obj = og_obj;
+    obj_modifier_func(modded_obj);
+    const std::string og_blob = stringify_with_do_serialize(og_obj);
+    const std::string modded_blob = stringify_with_do_serialize(modded_obj);
+    const bool did_change = modded_blob != og_blob;
+    if (did_change != expected_change)
+    {
+        const std::string og_hex = epee::to_hex::string(epee::strspan<uint8_t>(og_blob));
+        const std::string modded_hex = epee::to_hex::string(epee::strspan<uint8_t>(modded_blob));
+        MERROR("unexpected: modded_blob '" << modded_hex << "' vs og_blob ' << " << og_hex << "'");
```
iamamyth, line 218: Could make use of `AssertionResult`, `AssertionSuccess`, and `AssertionFailure` here, e.g.:
`
static testing::AssertionResult modification_changes_do_serialize
(
... og_obj,
... obj_modifier_func
// note removal of expected_change
)
{
  ..
  const std::string og_hex = epee::to_hex::string(epee::strspan<uint8_t>(og_blob));
  const std::string modded_hex = epee::to_hex::string(epee::strspan<uint8_t>(modded_blob));
  if (og_blob == modded_blob)
    return testing::AssertionFailure() <<  "serialized blobs match: modded_blob '" << modded_hex << "' og_blob '" << og_hex << "'";
  else
    return testing::AssertionSuccess() <<  "serialized blobs differ: modded_blob '" << modded_hex << "' og_blob '" << og_hex << "'"; 
```

Then, if you expect a change:
```
EXPECT_TRUE(modification_changes_do_serialize(og_obj, obj_modifier_func)
```
And if you expect no change:
EXPECT_FALSE(modification_changes_do_serialize(og_obj, obj_modifier_func)

File: tests/unit_tests/verRctNonSemanticsSimple.cpp [hunk 21]
```diff
@@ -0,0 +1,434 @@
...
+#define UNSERIALIZABLE_SIG_CHANGES_SUBTEST(fieldmodifyclause)                                    \
+    do {                                                                                         \
+        using sig_mod_func_t = std::function<void(rct::rctSig&)>;                                \
+        const sig_mod_func_t sig_modifier_func = [](rct::rctSig& rs) { rs.fieldmodifyclause; };  \
+        EXPECT_FALSE(modification_changes_do_serialize(original_sig, sig_modifier_func, false)); \
+    } while (0);                                                                                 \
+
+TEST(verRctNonSemanticsSimple, unserializable_sig_changes)
+{
+    // Hello, future visitors. If this unit test fails, than congrats! ::do_serialize(rctSig) became
```
iamamyth, line 360: Typo: `than` => `then`

tests/unit_tests/verRctNonSemanticsSimple.cpp
```diff
@@ -0,0 +1,434 @@
...
+    do {                                                                                          \
+        using mr_mod_func_t = std::function<void(rct::ctkeyM&)>;                                  \
+        const mr_mod_func_t mr_modifier_func = [&](rct::ctkeyM& mr) { mr fieldmodifyclause; };    \
+        EXPECT_TRUE(modification_changes_do_serialize(original_mixring, mr_modifier_func, true)); \
+    } while (0);                                                                                  \
+
+TEST(verRctNonSemanticsSimple, serializable_mixring_changes)
+{
+    // Hello, future Monero devs! If this unit test fails, a huge concensus-related assumption has
+    // been broken and verRctNonSemanticsSimpleCached needs to be reevalulated for validty. If is is
```
iamamyth, line 384: Typo: `validty` => `validity`

iamamyth:
Some minor cosmetic comments but the direction and general result look reasonable.

tests/unit_tests/verRctNonSemanticsSimple.cpp
jeffro256: But sometimes I want the modification to *not* result in a serializable change. In that case, using `AssertionFailure` doesn't seem quite right in that case, since matching blobs is the success case.

iamamyth: `AssertionFailure` and `Success` act like bools with a message attached (this exact use is very clearly documented in the gtest.h file), hence the call to `EXPECT_[False|True]`. I guess you can quibble with google's naming choices but that makes using the gtest library much more difficult.

iamamyth: I suppose I should note that I think your premise misses the mark a bit:
> But sometimes I want the modification to not result in a serializable change. In that case, using AssertionFailure doesn't seem quite right in that case,

The assertion failing in this case is that "modification changes do serialize". So, the result from the function makes sense. Then the caller decides whether the scenario should or should not change do serialize, i.e. whether the function's result constitutes a test failure or success. Whereas, the current version tries to combine the two separable concerns, causing the boolean value to repeat in the function call and the assertion (i.e. `EXPECT_TRUE(..., true)` and `EXPECT_FALSE(..., false)`).

jeffro256:
@UkoeHB done!

src/cryptonote_core/blockchain.h
```diff
@@ -1222,6 +1232,9 @@ namespace cryptonote
     uint64_t m_prepare_nblocks;
     std::vector<block> *m_prepare_blocks;
 
+    // cache for verifying transaction RCT non semantics
+    mutable rct_ver_cache_t m_rct_ver_cache;
```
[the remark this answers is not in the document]
jeffro256: We could alternatively make `Blockchain::check_tx_inputs` non-`const`.

src/cryptonote_core/tx_verification_utils.cpp
```diff
@@ -0,0 +1,166 @@
...
+    VER_ASSERT(!tx.pruned, "Pruned transaction will not pass verRctNonSemanticsSimple");
+
+    // Calculate prefix hash
+    const crypto::hash tx_prefix_hash = get_transaction_prefix_hash(tx);
+
+    // Expand mixring, tx inputs, tx key images, prefix hash message, etc into the RCT sig
+    const bool exp_res = Blockchain::expand_transaction_2(tx, tx_prefix_hash, mix_ring);
+    VER_ASSERT(exp_res, "Failed to expand rct signatures!");
+
+    const rct::rctSig& rv = tx.rct_signatures;
```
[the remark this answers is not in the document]
jeffro256: I don't know if that's necessarily true in this case, however. IMO this function has a fairly well defined scope. This reference is just here to make the function more readable and less unwieldy.

src/cryptonote_core/tx_verification_utils.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+        bool error;
+        if (rct::is_rct_clsag(rv.type))
+        {
+            error = memcmp(&nth_vin_image, &rv.p.CLSAGs[n].I, 32);
+        }
+        else
+        {
+            error = rv.p.MGs[n].II.empty() || memcmp(&nth_vin_image, &rv.p.MGs[n].II[0], 32);
+        }
```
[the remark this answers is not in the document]
jeffro256: How do like the latest rewrite of this part? It also gives more descriptive error messages while keeping the control flow simple.

tests/unit_tests/ver_rct_non_semantics_simple_cached.cpp
```diff
@@ -244,6 +236,8 @@ static const rct::ctkeyM tx1_input_pubkeys =
     make_ctkey("2d2ffade0f85ddd83a036469e49542e93cad94f9bea535f0ea2eb2f56304517e", "bcc48d00bd06dc5439200e749d0caf8a062b072d0c0eb1f78f6a4d8f2373e5f4"),
     make_ctkey("4ee857d0ce17f66eca9c81eb326e404ceb50c8198248f2f827c440ee7aa0c0d7", "a8a9d61d4abbfb123630ffd214c834cc45113eaa51dd2f904cc6ae0c3c5d70e3")
 }};
+
+static cryptonote::rct_ver_cache_t rct_ver_cache;
```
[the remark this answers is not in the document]
jeffro256: Agreed

selsta (approved):

PR #8712, opened 2023-01-17
blockchain-stats: print any remaining data at end of loop

### LINKED ISSUE #8711: monero-blockchain-stats omits data from final day
The shell command
`./monero-blockchain-stats --block-start=2797900 --block-stop=2800468`
prints
```
Date	Blocks/day	Blocks	Txs/Day	Txs	Bytes/Day	Bytes
2023-01-12	423	2798323	10478	10478	26770465	26770465
2023-01-13	688	2799011	15464	25942	38953621	65724086
2023-01-14	774	2799785	15034	40976	38456559	104180645
```

The start block, 2797900, has a timestamp of 2023-01-12 10:55:23 UTC. `monero-blockchain-stats` prints partial data from 2023-01-12, which is fine. The stop block, 2800468, is the final block on the day of 2023-01-15, but no data from 2023-01-15 is printed. (If `block-stop` is set to 2800469, i.e. one block into the next day (2023-01-16), still nothing from 2023-01-15 is printed. `block-stop` must be set to 2800470, i.e. two blocks into the next day, to print data from 2023-01-15.) Setting `block-stop` to a block with timestamp partway through 2023-01-15, e.g. 2800100, also prints nothing from 2023-01-15. This seems to be general behavior of the utility and not something specific to these particular days.

This isn't expected behavior. My suggestion is to print partial data for the final day, similar to the behavior for printing partial data for the first day. Another option is to not print any partial data (but then why would the flags allow the user to specify block heights and not dates directly?). In any case, if the user specifies a block height interval that completely covers a certain day, that day's data should be printed. That's not happening at the moment.

Thank you for any improvements that can be made.

hyc:
Fix #8711

### DIFF
--- a/src/blockchain_utilities/blockchain_stats.cpp
+++ b/src/blockchain_utilities/blockchain_stats.cpp
@@ -46,6 +46,77 @@ using namespace cryptonote;
 
 static bool stop_requested = false;
 
+static bool do_inputs, do_outputs, do_ringsize, do_hours, do_emission, do_fees, do_diff;
+
+static struct tm prevtm, currtm;
+static uint64_t prevsz, currsz;
+static uint64_t prevtxs, currtxs;
+static uint64_t currblks;
+static uint64_t h;
+static uint64_t totins, totouts, totrings;
+static boost::multiprecision::uint128_t prevemission, prevfees;
+static boost::multiprecision::uint128_t emission, fees;
+static boost::multiprecision::uint128_t totdiff, mindiff, maxdiff;
+
+#define MAX_INOUT	0xffffffff
+#define MAX_RINGS	0xffffffff
+
+static uint32_t minins = MAX_INOUT, maxins;
+static uint32_t minouts = MAX_INOUT, maxouts;
+static uint32_t minrings = MAX_RINGS, maxrings;
+static uint32_t io, tottxs;
+static uint32_t txhr[24];
+
+static void doprint()
+{
+  char timebuf[64];
+
+  strftime(timebuf, sizeof(timebuf), "%Y-%m-%d", &prevtm);
+  prevtm = currtm;
+  std::cout << timebuf << "\t" << currblks << "\t" << h << "\t" << currtxs << "\t" << prevtxs + currtxs << "\t" << currsz << "\t" << prevsz + currsz;
+  prevsz += currsz;
+  currsz = 0;
+  prevtxs += currtxs;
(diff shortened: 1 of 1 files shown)


jeffro256:
`./monero-blockchain-stats --block-start=2797900 --block-stop=2800468` before:
```
# DATA
Date	Blocks/day	Blocks	Txs/Day	Txs	Bytes/Day	Bytes
2023-01-12	423	2798323	10478	10478	26770465	26770465
2023-01-13	688	2799011	15464	25942	38953621	65724086
2023-01-14	774	2799785	15034	40976	38456559	104180645
```
`./monero-blockchain-stats --block-start=2797900 --block-stop=2800468` after:
```
# DATA
Date	Blocks/day	Blocks	Txs/Day	Txs	Bytes/Day	Bytes
2023-01-12	423	2798323	10478	10478	26770465	26770465
2023-01-13	688	2799011	15464	25942	38953621	65724086
2023-01-14	774	2799785	15034	40976	38456559	104180645
2023-01-15	683	2800468	14050	55026	35735318	139915963
```

src/blockchain_utilities/blockchain_stats.cpp
```diff
@@ -46,6 +46,77 @@ using namespace cryptonote;
...
+#define MAX_INOUT	0xffffffff
+#define MAX_RINGS	0xffffffff
+
+static uint32_t minins = MAX_INOUT, maxins;
+static uint32_t minouts = MAX_INOUT, maxouts;
+static uint32_t minrings = MAX_RINGS, maxrings;
+static uint32_t io, tottxs;
+static uint32_t txhr[24];
+
+static void doprint()
```
jeffro256, line 70: I don't like how `doprint()` is now a singleton, but the logic is correct. I verified the code body matches with `diff -b`. Zero-initialization is not needed for `static` data, so that is correct AFAIK. Good patch!

jeffro256 (approved):

src/blockchain_utilities/blockchain_stats.cpp
hyc: Yeah, unlike what OOP proponents would have you believe, not everything needs to be a class. Abstraction for simplifying complex systems makes sense. Abstracting something that's already braindead simple is just a pessimization.

jeffro256:
Can we get this on the release branch also?

PR #8715, opened 2023-01-19
wallet2: speed up pick_preferred_rct_inputs and get_output_relatednes

jeffro256:
Inside of `pick_preferred_rct_inputs`, when we are looking for 1 output which is spendable and fulfills the amount requirement, we create a local `vector` which stores the indices of only the "spendable" transfers (regardless of amounts). Then, when we look for a 2 output combination, we iterate through said list instead of the whole transfers list. This 1) reduces branching and 2) reduces the complexity of the algorithm from `O(T^2)` to `O(max(T, S^2))` where `T = # of total transfers` and `S = # of spendable transfers`. In the case where a large portion of the outputs are already spent, this can be speed up the output picking by quite a bit.

Inside `get_output_relatedness`, the expensive txid check is postponed until after the block height tests.

Co-authored-by: moneromooo-monero <moneromooo-monero@users.noreply.github.com>

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -9336,17 +9336,22 @@ std::vector<size_t> wallet2::pick_preferred_rct_inputs(uint64_t needed_money, ui
 
   LOG_PRINT_L2("pick_preferred_rct_inputs: needed_money " << print_money(needed_money));
 
+  // pre-cache a "is this output to be considered" array to speed up tests later
+  std::vector<bool> spendable(m_transfers.size(), false);
+  for (size_t i = 0; i < m_transfers.size(); ++i)
+  {
+    const transfer_details& td = m_transfers[i];
+    if (!is_spent(td, false) && !td.m_frozen && td.is_rct() && is_transfer_unlocked(td) && td.m_subaddr_index.major == subaddr_account && subaddr_indices.count(td.m_subaddr_index.minor) == 1 && !td.m_key_image_partial)
+      if (td.amount() <= m_ignore_outputs_above && td.amount() >= m_ignore_outputs_below)
+        spendable[i] = true;
+  }
+
   // try to find a rct input of enough size
   for (size_t i = 0; i < m_transfers.size(); ++i)
   {
     const transfer_details& td = m_transfers[i];
-    if (!is_spent(td, false) && !td.m_frozen && td.is_rct() && td.amount() >= needed_money && is_transfer_unlocked(td) && td.m_subaddr_index.major == subaddr_account && subaddr_indices.count(td.m_subaddr_index.minor) == 1)
+    if (spendable[i] && td.amount() >= needed_money)
     {
-      if (td.amount() > m_ignore_outputs_above || td.amount() < m_ignore_outputs_below)
-      {
-        MDEBUG("Ignoring output " << i << " of amount " << print_money(td.amount()) << " which is outside prescribed range [" << print_money(m_ignore_outputs_below) << ", " << print_money(m_ignore_outputs_above) << "]");
-        continue;
-      }
       LOG_PRINT_L2("We can use " << i << " alone: " << print_money(td.amount()));
       picks.push_back(i);
       return picks;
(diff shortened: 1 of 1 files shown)


jeffro256:
This PR is an extension to https://github.com/monero-project/monero/pull/8704.

selsta:
@jeffro256 can you specify extension? should we merge moneromooo's PR first and then yours or does your PR replace it?

jeffro256:
It replaces it, sorry for not being specific. It is a spiritual extension, but the implementation replaces what is in #8704. Specifically, #8704 uses a spendable vector as well, but it is a cache of `true`/`false` for each transfer, so the time complexity is the same as `master`.

moneromooo-monero:
I find this surprising, I was expecting your PR to follow up on mine, instead of merging my code in. Maybe I was confusing when we spoke (sorry). Then again, I don't really mind that much. It's just... weird and kinda not what one expects in a project where different contributors contribute changes. If the issue is that you can't actually PR the patch yet because github complains about conflicts, what I usually do in these cases is PR the whole set of commits, then rebase once the PR(s) I need are merged, leaving only my clean patch in the PR.

jeffro256:
Sorry, yeah that was why I originally PRed against your PR. This isn't really a patch to yours so much as it replaces the `pick_preferred_rct_inputs` commit. I figured since you seemed to want to avoid the whole "merging PRs against PRs" that I just open the new one, so reviewers don't have to do double the work, and so the commit history is a little cleaner. I could cherry-pick the `get_output_relatedness` commit and rebase against that one?

moneromooo-monero:
I guess if you want to replace my patch I can drop it from my PR if it makes more sense to you.

jeffro256:
You have closed your PR since then, do you still want to move forward with this one?

selsta:
Should we move forward with this one?

src/wallet/wallet2.cpp
```diff
@@ -39,7 +39,9 @@
 #include <boost/algorithm/string/split.hpp>
 #include <boost/algorithm/string/join.hpp>
 #include <boost/algorithm/string/predicate.hpp>
+#include <boost/archive/binary_iarchive.hpp>
 #include <boost/asio/ip/address.hpp>
+#include <boost/filesystem/operations.hpp>
```
Tzadiko, line 44: Consider std::filesystem added in C++17.

src/wallet/wallet2.cpp
```diff
@@ -6338,29 +6339,25 @@ namespace
 // their ordering, but it could become more murky if we add scores later.
 float wallet2::get_output_relatedness(const transfer_details &td0, const transfer_details &td1) const
 {
-  int dh;
+  const int64_t dh = td0.m_block_height > td1.m_block_height ? td0.m_block_height - td1.m_block_height : td1.m_block_height - td0.m_block_height;
```
Tzadiko, line 6342: Can you give this a better name please, while you're already editing this line?

src/wallet/wallet2.cpp
```diff
@@ -6338,29 +6339,25 @@ namespace
...
 
-  // expensive test, and same tx will fall onto the same block height below
-  if (td0.m_txid == td1.m_txid)
-    return 1.0f;
+  // don't think these are particularly related
+  if (dh >= 10)
```
Tzadiko, line 6345: Would be nice to give this number a name.

src/wallet/wallet2.cpp
```diff
@@ -9336,47 +9333,55 @@ std::vector<size_t> wallet2::pick_preferred_rct_inputs(uint64_t needed_money, ui
...
-    const transfer_details& td = m_transfers[i];
-    if (!is_spent(td, false) && !td.m_frozen && !td.m_key_image_partial && td.is_rct() && is_transfer_unlocked(td) && td.m_subaddr_index.major == subaddr_account && subaddr_indices.count(td.m_subaddr_index.minor) == 1)
-    {
-      if (td.amount() > m_ignore_outputs_above || td.amount() < m_ignore_outputs_below)
-      {
-        MDEBUG("Ignoring output " << i << " of amount " << print_money(td.amount()) << " which is outside prescribed range [" << print_money(m_ignore_outputs_below) << ", " << print_money(m_ignore_outputs_above) << "]");
-        continue;
-      }
+      const size_t i = spendable_transfer_indices[si];
+      const transfer_details& td = m_transfers[i];
```
Tzadiko, line 9378: NIT: I always like to use .at() instead of []. It's just a habit I've built because it will notify you via an exception if you accidentally out-of-bounds it. I see a case where someone accidentally does:

```
const size_t i = spendable_transfer_indices[si];
const transfer_details& td = m_transfers[si]; // Forgot to use i
```

And it doesn't get picked up in a code-review.

PR #8730, opened 2023-02-03
wallet_rpc_server: allow creating more than 64 addresses at once

moneromooo-monero:
it's too low a limit (at least one person mentioned having to call create_address in a loop due to it)

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -583,9 +583,9 @@ namespace tools
     if (!m_wallet) return not_open(er);
     try
     {
-      if (req.count < 1 || req.count > 64) {
+      if (req.count < 1 || req.count > 65536) {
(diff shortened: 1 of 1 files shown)


SamsungGalaxyPlayer:
65536 is a lot of addresses :)

moneromooo-monero:
Just to be clear, it's the max you ask for (ie, sanity check), you can always ask for fewer.

selsta (approved):

jeffro256 (approved):
The only potential issue I see here is that if someone creates many subaddresses (greater than 200 `SUBADDRESS_LOOKAHEAD_MINOR`) and they lose their wallet cache, this could cause rescanning problems that wouldn't happen with the limit at 64, but if they're making this many subaddresses, they probably know about this quirk, or could at least look it up.

tobtoht (approved):

plowsof (approved):

Issue #8550, opened 2022-09-06
monerod docker container 180% CPU usage (After full sync, public node, full node)


NeonMinnen:
Hello!
I am running a public node on a VPS but seem to be having an issue with Monerod utilizing too much CPU. For the first 10 hours it runs, it's completely fine and doesn't use more than 4-8% of CPU. (This is after sync, 100%, and running a full node).

After the 10 hours, I get a sharp uptake to pretty much a constant 180% CPU usage (4 core, 8GB RAM VPS). I'm not quite sure what's going on. It will stay like this until I restart monerod. Once I restart, it will be okay again for another 10-12 hours, and then it goes into the same pattern. Any help would be appreciated. I can post whatever logs you may require.

I am using this docker container https://hub.docker.com/r/sethsimmons/simple-monerod

Below is my docker-compose.

```
   monerod:
    image: sethsimmons/simple-monerod:latest
    restart: unless-stopped
    container_name: monerod
    volumes:
      - /mnt/volume-1/bitmonero:/home/monero/.bitmonero
    ports:
      - 18080:18080
      - 18089:18089
    command: >-
      --rpc-restricted-bind-ip=0.0.0.0
      --rpc-restricted-bind-port=18089
      --public-node
      --no-igd
      --no-zmq
      --enable-dns-blocklist
      --limit-rate-up 100000
      --limit-rate-down 100000
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.monero.rule=Host(`${MONERO_HOSTNAME}`)"
      - "traefik.http.routers.monero.entrypoints=websecure"
      - "traefik.http.routers.monero.tls.certresolver=selfhostedservices"
      - "traefik.http.services.monero.loadbalancer.server.port=18089"
```

Cheers

selsta:
As a first step try to run a node without docker and without config. If there are no issues re-add the config.

NeonMinnen:
Thought I would add to this for more info.

Here is also an image of the normal CPU usage to the left, versus when it goes into full CPU hog mode. Taken directly from the monerod CPU usage via Netdata.
![Screenshot-from-2022-09-06-22-22-13](https://user-images.githubusercontent.com/98937938/188795780-f557af9a-4494-4d57-8a94-71e80582d990.png)

Here are some logs that look concerning.
https://paste.decentralize.nl/?d0a8738ef4b524d5#5oChadBbzzsLR1iYqs83sqLFPKGBYMmwD47jELK6cZxY


```

[...]

PR #8731, opened 2023-02-03
epee net tcp server: properly terminate interrupted TCP connection. fixes #8685

### LINKED ISSUE #8685: monerod: High CPU usage after 2-3 days
I've been running a monero node "Monero 'Fluorine Fermi' (v0.18.1.2-release)" on Archlinux for a few months now.
Every few days I have to restart monerod, as it will constantly use around 160-180% CPU.
Regular CPU usage is around 2-10%

I think that this hasn't been an issue from the start, but is now constantly plaguing me.
I can still connect and use the daemon as a remote wallet when this happens.

Here's my config:
```
# Configuration for monerod
# Syntax: any command line option may be specified as 'clioptionname=value'.
#         Boolean options such as 'no-igd' are specified as 'no-igd=1'.
# See 'monerod --help' for all available options.

# Data directory (blockchain db and indices)
data-dir=/var/lib/monero

# Log file
log-file=/var/log/monero/monerod.log
log-level=0
max-log-file-size=0            # Prevent monerod from managing the log files; we want logrotate to take care of that

# P2P full node
p2p-bind-ip=0.0.0.0            # Bind to all interfaces (the default)
p2p-bind-port=18080            # Bind to default port

# RPC open node
rpc-bind-ip=0.0.0.0            # Bind to all interfaces
rpc-bind-port=18081            # Bind on default port
confirm-external-bind=1        # Open node (confirm)
restricted-rpc=1               # Prevent unsafe RPC calls
rpc-ssl=enabled
rpc-ssl-certificate=/etc/ssl/private/monerod.pem
rpc-ssl-private-key=/etc/ssl/private/monerod.key
rpc-login=****
no-igd=1                       # Disable UPnP port mapping
no-zmq=1                       # Disable ZMQ RPC server to decrease attack surface (it's not used)


# Emergency checkpoints set by MoneroPulse operators will be enforced to workaround potential consensus bugs
# Check https://monerodocs.org/infrastructure/monero-pulse/ for explanation and trade-offs
enforce-dns-checkpointing=1
enable-dns-blocklist=1

out-peers=16              # This will enable much faster sync and tx awareness; the default 8 is suboptimal nowadays
in-peers=64               # The default is unlimited; we prefer to put a cap on this

limit-rate-up=30000     # 1048576 kB/s == 1GB/s; a raise from default 2048 kB/s; contribute more to p2p network
limit-rate-down=100000   # 1048576 kB/s == 1GB/s; a raise from default 8192 kB/s; allow for faster initial sync
```

I've attached the monerod log file here, from start to finish:
[monerod.log.txt](https://github.com/monero-project/monero/files/10280925/monerod.log.txt)
Though I suppose I should've raised the log-level to 1 before reporting this, in any case I have now so just need to wait another two to three days.

CPU usage started rising around 2022-12-21 11:32:00
image

I've also made a core dump before restarting the monero daemon:
https://cloud.botox.bz/s/HrgRyR8d9Nk3gxb

Edit: I should've run perf to find out where exactly monerod was wasting cpu cycles, ooops. will do soon.

BotoX:
See issue https://github.com/monero-project/monero/issues/8685
Specifically the comment/patch by @j-berman https://github.com/monero-project/monero/issues/8685#issuecomment-1396659121

monerod would start using 160-180% CPU after a few days of uptime.
I was consistently running into this issue for multiple months.
With this patch monerod has been running for two weeks already, without high cpu usage.

Related/Same issue https://github.com/monero-project/monero/issues/8550

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -583,11 +583,8 @@ namespace net_utils
             break;
         }
       }
-      else if (ec.value())
-        terminate();
       else {
-        cancel_timer();
-        on_interrupted();
+        terminate();
       }
     };
     m_strand.post(


plowsof:
fix commit author

BotoX:
> fix commit author (or ask them to PR the fix)

done.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -586,8 +586,7 @@ namespace net_utils
       else if (ec.value())
```
iamamyth, line 586: Given the change below, this `else if` can be folded into the `else`.

j-berman (approved):
Repeating [my logic](https://github.com/monero-project/monero/issues/8685#issuecomment-1396659121) for this patch:

> I'm thinking until someone actually writes the code to reuse an "interrupted" connection, it makes sense to simply complete the connection termination sequence (and shut down the TCP connection) after the server shuts the SSL stream down. This way the server isn't leaving a connection around in this interrupted state that can't be "un-interrupted."

Adding some more logic to support this reasoning... once a connection has entered the interrupted state, it will stop reading data received over the socket thanks to this section:

https://github.com/monero-project/monero/blob/c5d10a4ac43941fe7f234d487f6dd54996a9aa33/contrib/epee/include/net/abstract_tcp_server2.inl#L412-L420

When the connection is in the interrupted state, it falls into the first if, preventing `start_read` from getting called in the else. `start_read` is what calls `async_read_some` to read the next chunk of data received over the socket. Thus, the interrupted state prevents the server from reading more data received over the socket.

As written, I don't see why the server should keep an interrupted connection around.

vtnerd:
> As written, I don't see why the server should keep an interrupted connection around.

What is the purpose of the state? At a glance I don't see where the state/status can become `INTERRUPTED`.

j-berman:
From what I recall in discussions with the author, their intent behind the connection's interrupted status was to allow the server to reuse a connection that hasn't been terminated yet. But the author didn't actually implement the code to reuse an interrupted connection (I don't see a way for a connection's status to go from `INTERRUPTED` to anything but `TERMINATING` or `WASTED`). I believe they intended to implement the ability to reuse the underlying TCP connection *in a future PR* building off this interrupted status.

More behind my thinking on this patch is here: https://github.com/monero-project/monero/issues/8685#issuecomment-1396659121

vtnerd:
@j-berman its tempting to slowly remove the interrupted state completely - it seems to complicated for our use cases.

vtnerd (approved):

PR #8733, opened 2023-02-05
Add a size limit for tx_extra in tx pool

tevador:
Related to #6668

This patch limits the maximum size of tx_extra that's allowed in the transaction pool. Mined transactions bypass this check, which means this is a non-consensus change.

I set the limit to 1060 bytes. A typical transaction with 16 outputs needs up to 547 bytes for the transaction public keys, which still leaves 32 bytes of space for some custom data to send to each recipient.

Note that Bitcoin also has a [limit](https://github.com/bitcoin/bitcoin/blob/e325e0fccba4981d28053b79473ddaa44355e6e8/src/kernel/mempool_options.h#L47-L54) for arbitrary data payloads that defaults to a much lower value of just 83 bytes per transaction.

The purpose of this change is to make it harder to "upload" large amounts of unprunable data onto the blockchain. [Here](https://xmrchain.net/tx/21d0218ba8141a07048986d9e74ce8a011d9d5fddc3283dce58339da74c4bacd) is an example of a transaction with a tx_extra size of 126 KB containing a PDF file. The limit of ~1 KB should be sufficient for useful content such as the metadata of decentralized exchanges.

Ping @kayabaNerve

### DIFF
--- a/src/cryptonote_basic/verification_context.h
+++ b/src/cryptonote_basic/verification_context.h
@@ -53,6 +53,7 @@ namespace cryptonote
     bool m_overspend;
     bool m_fee_too_low;
     bool m_too_few_outputs;
+    bool m_tx_extra_too_big;
   };
 
   struct block_verification_context
--- a/src/cryptonote_config.h
+++ b/src/cryptonote_config.h
@@ -206,6 +206,11 @@
 
 #define DNS_BLOCKLIST_LIFETIME (86400 * 8)
 
+//The limit is enough for the mandatory transaction content with 16 outputs (547 bytes),
+//a custom tag (1 byte) and up to 32 bytes of custom data for each recipient.
+// (1+32) + (1+1+16*32) + (1+16*32) = 1060
+#define MAX_TX_EXTRA_SIZE                       1060
+
 // New constants are intended to go here
 namespace config
 {
--- a/src/cryptonote_core/cryptonote_tx_utils.cpp
+++ b/src/cryptonote_core/cryptonote_tx_utils.cpp
@@ -437,6 +437,8 @@ namespace cryptonote
     if (!sort_tx_extra(tx.extra, tx.extra))
       return false;
 
+    CHECK_AND_ASSERT_MES(tx.extra.size() <= MAX_TX_EXTRA_SIZE, false, "TX extra size (" << tx.extra.size() << ") is greater than max allowed (" << MAX_TX_EXTRA_SIZE << ")");
+
     //check money
     if(summary_outs_money > summary_inputs_money )
     {
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -218,6 +218,15 @@ namespace cryptonote
       return false;
     }
 
+    size_t tx_extra_size = tx.extra.size();
+    if (!kept_by_block && tx_extra_size > MAX_TX_EXTRA_SIZE)
+    {
+      LOG_PRINT_L1("transaction tx-extra is too big: " << tx_extra_size << " bytes, the limit is: " << MAX_TX_EXTRA_SIZE);
+      tvc.m_verifivation_failed = true;
+      tvc.m_tx_extra_too_big = true;
+      return false;
+    }
+
     // if the transaction came from a block popped from the chain,
     // don't check if we have its key images as spent.
     // TODO: Investigate why not?
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -1275,6 +1275,8 @@ namespace cryptonote
           add_reason(reason, "fee too low");
         if ((res.too_few_outputs = tvc.m_too_few_outputs))
           add_reason(reason, "too few outputs");
+        if ((res.tx_extra_too_big = tvc.m_tx_extra_too_big))
+          add_reason(reason, "tx-extra too big");
         const std::string punctuation = reason.empty() ? "" : ": ";
         if (tvc.m_verifivation_failed)
         {
--- a/src/rpc/core_rpc_server_commands_defs.h
+++ b/src/rpc/core_rpc_server_commands_defs.h
@@ -88,7 +88,7 @@ namespace cryptonote
 // advance which version they will stop working with
 // Don't go over 32767 for any of these
 #define CORE_RPC_VERSION_MAJOR 3
-#define CORE_RPC_VERSION_MINOR 11
+#define CORE_RPC_VERSION_MINOR 12
 #define MAKE_CORE_RPC_VERSION(major,minor) (((major)<<16)|(minor))
 #define CORE_RPC_VERSION MAKE_CORE_RPC_VERSION(CORE_RPC_VERSION_MAJOR, CORE_RPC_VERSION_MINOR)
 
@@ -592,6 +592,7 @@ namespace cryptonote
       bool fee_too_low;
       bool too_few_outputs;
       bool sanity_check_failed;
+      bool tx_extra_too_big;
 
       BEGIN_KV_SERIALIZE_MAP()
         KV_SERIALIZE_PARENT(rpc_access_response_base)
@@ -606,6 +607,7 @@ namespace cryptonote
         KV_SERIALIZE(fee_too_low)
         KV_SERIALIZE(too_few_outputs)
         KV_SERIALIZE(sanity_check_failed)
+        KV_SERIALIZE(tx_extra_too_big)
       END_KV_SERIALIZE_MAP()
     };
     typedef epee::misc_utils::struct_init<response_t> response;


jeffro256:
I am generally against adding relay rules to tackle the `tx_extra` issue, because it will cause hard-to-debug issues for people trying to broadcast transactions that are otherwise compliant, especially if they are not using the reference code.

However, if a soft limit were to be implemented, this value should hard coded and NOT easily adjustable by the user (unless they want to compile their own node). Their should also be an error/warning for trying to construct a transaction with `tx_extra` being too large, so a user can know if a transaction won't be relayed before constructing and broadcasting the transaction.

tevador:
> it will cause hard-to-debug issues for people trying to broadcast transactions that are otherwise compliant

The daemon RPC call will return an error message `"tx-extra too big"`, which should make it obvious.

> this value should hard coded and NOT easily adjustable by the user

The adjustability is mostly for miners if they wanted to mine their own non-standard tx. But I see your point, since we also don't have an option to override the default minimum fee.

jtgrassie:
> this value should hard coded and NOT easily adjustable by the user

I tend to agree with this, as it's not obvious to me why even a miner should/would need to override. 

> 16 messages with ~30 characters per recipient ("Jim Jones, payroll 1/1/2023", "Happy birthday!")

The decision was already made wrt payment IDs to do away with custom non-encrypted messages in favor of 8 byte encrypted payment IDs, so I don't think we need revisit this use case.

> 1024 to be even more concise yet still tolerable to most protocols. At least do this as proposed.

1024 should suffice, but if others want to err on the side of caution, 2150 as proposed should be more than adequate.

kayabaNerve:
> The decision was already made wrt payment IDs to do away with custom non-encrypted messages in favor of 8 byte encrypted payment IDs, so I don't think we need revisit this use case.

Somewhat agree, largely disagree with this specific point. Not only are payment IDs only usable for a known context, which jumps past the issue of context establishment, there's only one encrypted payment ID per TX. While no, I don't think people should be placing "payroll" nor "happy birthday" on chain, encrypted payment IDs aren't a complete solution.

jtgrassie:
> While no, I don't think people should be placing "payroll" nor "happy birthday" on chain, encrypted payment IDs aren't a complete solution.

Yet you _are_ advocating for "~30 characters per recipient" (for a solution to an undefined use case). 

The encrypted payment IDs _are_ a solution to a specific use case, and one that removed the ability to easily embed custom data. I'd prefer we limit to known use cases and adjust in the future (as and when compelling use cases are presented).

jeffro256:
> The daemon RPC call will return an error message `"tx-extra too big"`, which should make it obvious.

One problem:This PR doesn't return how big `tx_extra` is allowed to be, so what should the wallet do in this scenario? Start binary searching how big it can make it? However, that could be fixed by making the RPC interface more expressive.

tevador:
Thanks for the feedback. Changes:

* Removed the command line option. The limit is now a compile-time constant.
* Reduced the limit from 2150 bytes to 1060 bytes.

> For a 16-out TX, I calculate 557 bytes off-hand (due to the mandatory payment ID)

The dummy payment ID is added only for 2-output transactions.

https://github.com/monero-project/monero/blob/50aa0e8b7f11680be3954c176f2daa9ccf77b7dd/src/cryptonote_core/cryptonote_tx_utils.cpp#L269-L271

The current limit of 1060 bytes equals `547 + 1 + 512`, so exactly enough space for 16 outputs, a 1-byte tag and 32 bytes for each recipient.

SamsungGalaxyPlayer:
@jtgrassie at the minimum, it's somewhat reasonable to allow for a refund address to be encoded in a transaction. Ideally this can be communicated out of band ofc.

I agree with the proposed limit. It won't prevent everything, but it'll prevent most forms of drive-by, low-effort malice.

kayabaNerve:
> The current limit of 1060 bytes equals 547 + 1 + 512, so exactly enough space for 16 outputs, a 1-byte tag and 32 bytes for each recipient.

Slight caveat. Will Monero fail to scan such a TX? https://github.com/monero-project/monero/blob/ab63fbc549b2534ce63726ba6ed2fdee29ca9a14/src/cryptonote_basic/cryptonote_format_utils.cpp#L553-L572

I'm pretty sure it requires intelligibility of extra. That'd mean it'd need *3* NONCE fields, each with <= 255 bytes. That's 8 bytes of TL data (tag, 224, tag, 224, tag, 32), not 1. Sorry if I'm misreading this.

jtgrassie:
@kayabaNerve 
> Slight caveat. Will Monero fail to scan such a TX? ... I'm pretty sure it requires intelligibility of extra. That'd mean it'd need 3 NONCE fields, each with <= 255 bytes. That's 8 bytes of TL data (tag, 224, tag, 224, tag, 32), not 1. Sorry if I'm misreading this.

There is no enforced structure to the data in tx extra.

@SamsungGalaxyPlayer 
> at the minimum, it's somewhat reasonable to allow for a refund address to be encoded in a transaction

1060 (or 1024) bytes still allows plenty of room for such a feature.

kayabaNerve:
@jtgrassie I said fail to scan. I understand monerod will accept and mine the TX. I'm unsure the 16 recipients will realize they received funds though under the current PR and one proposed usage.

tevador:
@kayabaNerve You may be right, but fixing the tx_extra parser is out of scope of this PR. I don't see why the whole tx would need to be skipped if some unknown data are found after the tx public keys.

jtgrassie:
> I said fail to scan. I understand monerod will accept and mine the TX. I'm unsure the 16 recipients will realize they received funds though under the current PR and one proposed usage.

This patch doesn't prevent scanning. It prevents txs being placed in the txpool if they have to big extra.

kayabaNerve:
@tevador My comment was if you use 8 bytes of tag data, Monero won't reject it. You calculated a number using only 1 byte of tag data. I was saying you may want to 'correct' it to +7. Else, it's actually just 505 bytes of usable data.

kayabaNerve:
:+1: to being 505.

I accused the extra parser, in Monero specifically, as it's a low priority topic which is always contentious. There's also no good way to handle unrecognized tags (immediately returning, I guess?). But a debate/PR for another day. Sorry if this sidetracked this PR too much.

jeffro256:
Sorry I don't know how to do PR code changes suggestions in GH for files outside the PR https://github.com/tevador/bitmonero/pull/1

tevador:
@jeffro256 Thanks, added.

jtgrassie:
This patch is really needed now that [someone](https://github.com/mooonero/mordinals) has released something to (ab)use the unconstrained size of the field.

jtgrassie (approved):

SamsungGalaxyPlayer (approved):
This is a sensible limitation. It doesn't completely prevent people from using tx_extra for any purpose, but it limits one's ability to include these transactions without direct miner cooperation. This should be the start of the response to tx_extra, not the end. We should set a consensus limit of 1060 bytes later as well.

Let's not avoid putting up any defenses while are still discussing the extent of the defenses we all want. 1060 bytes is an uncontroversial number, let's enforce it today and enforce other sensible limits later after those discussions have reached consensus.

plowsof:
With this PR. running the monero-wallet-cli from https://github.com/mooonero/mordinals with `--do-not-relay` to save the raw transaction file and then trying to `sendrawtransaction` i get this response:
```
{
  "credits": 0,
  "double_spend": false,
  "fee_too_low": false,
  "invalid_input": false,
  "invalid_output": false,
  "low_mixin": false,
  "not_relayed": false,
  "overspend": false,
  "reason": "",
  "sanity_check_failed": false,
  "status": "Failed",
  "too_big": false,
  "too_few_outputs": false,
  "top_hash": "",
  "tx_extra_too_big": true,
  "untrusted": false
}
```
the morbinal has been denied

jeffro256 (approved):

SChernykh (approved):

Gonbatfire:
I understand that no one wants bloat on a digital cash protocol, but can @tevador, @SamsungGalaxyPlayer or someone else explain to me how does using 50 transactions to inscribe 50KB image does LESS(?) harm to the privacy of the protocol than just 1 transaction?

Are you guys still prioritizing transaction privacy? This is worrisome.

adapt-L:
@Gonbatfire Your logic doesn't necessarily hold. You might as well say that a limit of 50KB is harming privacy because someone who wants to upload a 50MB image would have to use a thousand transactions, so we should actually *raise* the limit. What will probably happen is that they will be uploading 1KB files instead of 50KB files. Most of these images are less than a kilobyte anyways...

I agree that it doesn't completely prevent an adversarial use of tx_extra spam, but it does discourage the behavior and limits them to a kilobyte unless they want to redesign their protocol to handle files split across multiple transactions. Instead of uploading a massive file, now they will just post a Bittorrent or IPFS link or file hash, and distribute the whole file out-of-bound. 

I am not an expert, but my intuition is that it makes privacy slightly worse, even if these "fake NFT transactions" are identified and not used in decoy selection, because they compete with transactions that could be used as decoys. But I would suspect that the net effect is that less space will be used by these things because users will generally be uploading of 1kb pixel art instead of 50kb photographs.

Gonbatfire:
> Instead of uploading a massive file, now they will just post a Bittorrent
@adapt-L Completely disagree, it's trivial to update their protocol to just use multiple transactions, storing the data off-chain misses the point, the "value" is doing it all on-chain

At most what this change did is slightly raise the fee cost to store arbitrary data, which I don't oppose, but it also comes with the tradeoff that each "upload" will harm the network's privacy even more.

What I want you guys to understand is that this is fundamentally different from any other flood attack, because there is now **financial incentive** playing a role, that changes everything. (tagging @kayabaNerve)

For example 51% attacks are not a preoccupation because it generally goes against the financial incentive of miners to execute one, but if miners were to be financially incentivized to do 51% attacks (game theory is broken) then we would have a really bad time. Right now there is a group users who are financially incentivized to flood the chain, not good.

Recently a new ATH of $257,460 (9.28 BTC) was spent in fees (single day) to inscribe ordinals in Bitcoin, this is something that may as well catch on Monero eventually, we should strive to minimize their privacy impact as much as possible for when that happens.

adapt-L:
@Gonbatfire 
>the "value" is doing it all on-chain

Is it? I think this is a common misconception. I would argue that all the value is supposed to come from the issuer's signature, and the assumption that they (or others) will respect (care about) the metadata associated with the enote in the future. See projects like bitcoin's colored coins project, which existed way before projects like etherium's NFTs or bitcoin ordinals which standardized uploading tons of unnecessary metadata to the ledger. For example, if a vendor attaches a coupon to an enote (as described in this video: https://www.youtube.com/watch?v=889JSfIaPzs), the value of the colored coin rests on the assumption that they will accept the coupon at a later date. Or if some game developer like CS:GO uses colored coins to represent skins or something, the value of the skin is dependent on the authority issuing the colored coin not changing their game's client code to disrespect the metadata in the going forwards. It's the requirement of an issuer's signature which makes them scarce (if not valuable, like a autograph), because anyone else can easily create an output with similar metadata. 

Whether or not others will respect the metadata is the chief issue of NFTs, (if not cryptocurrency in general). You can goad other nodes into synchronizing the metadata by including it in OP_RETURNs or tx_extra's, but you can not force them to respect or use the metadata whatsoever. In monero, a good portion if its value comes from their fungibility, their lack of metadata (a.k.a. de-anonymization heuristics).

A similar system of colored coins could be easily instituted in monero, where the "minter" simply signs a certificate file that points to one of their outputs, and you use a log of tx proofs to indicate ownership. 
In fact, I would even be willing to help implement such a system if it would make these "mordinals" people stop spamming the chain with their image files! This would not only save on-chain data, but it would also improve privacy to the extent that these certs and tx_proofs are shared privately. Perhaps you can have a sidechain that shares valid tx_proofs and treats them like blackball transactions. However, it would not allow for rarity on the basis of the "order of inscriptions" (e.g. you could just make a certificate that points to an old enote). It may even be possible to create new type of cryptographic proof that replaces the tx_proofs in this situation with something that improves fungibility.

Additionally, you could just make it so that wallets only mix outputs with the same tx_extra data. That's a way of implementing colored coins that takes advantage of monero's existing privacy features!  For example, some issuer could create 30 colored coins that all have identical metadata, and they all use the same pool of outputs for decoy selection, so they are fungible amongst each other insofar as the other colored coins owners don't disclose their ownership.

>At most what this change did is slightly raise the fee cost to store arbitrary data, which I don't oppose, but it also comes with the tradeoff that each "upload" will harm the network's privacy even more.

>What I would like to make clear is that this is fundamentally different from any other flood attack, because there is now a financial incentive playing a role, that changes everything. (tagging @kayabaNerve)

It doesn't hurt privacy by output spamming, because the fact that they are not real transactions is transparent/public. Take whatever public stenographic mechanism they use to plant files on monero, and you have a mechanism to detect the fake transactions and add them to a blackball list. This is what the "mordinals" creator encourages and suggests. The problem is that these files will displace real transactions that would otherwise increase the anonset of monero. The solution is to encourage them not to waste so much space! I think the best way to do that is a combination of increasing the tx_extra fee and capping tx_extra at some reasonable limit to encourage them to use smaller files (perhaps use some sort of membership proofs to discourage output spamming? there was some debate about this on IRC)

>For example 51% attacks are not a preoccupation because it generally goes against the financial incentive of miners to execute one, but if miners were to be financially incentivized to do 51% attacks (game theory is broken) then we would have a really bad time. Right now there is a group users who are financially incentivized to flood the chain, not good.

That's predicated on the notion that their "inscriptions" will be valuable in the future, which is debatable. A more accurate analogy would be if it were uncertain if a 51% attack would be profitable or not.

>Recently a new ATH of $257,460 (9.28 BTC) was spent in fees (single day) to inscribe ordinals in Bitcoin, this is something that may as well catch on in Monero eventually, we should strive to minimize their privacy impact as much as possible for when that happens.

agreed. But if it helps monero gain some publicity it may actually improve the anonset of monero in the long run by attracting new users. It's not the greatest publicity though, as it encourages use amongst NFT grifters. It's also important to notice that monero does not have the same blocksize restrictions as bitcoin, so these types of comparisons are apples-and-oranges in a sense.

adapt-L:
Sorry for the long post, all.

kayabaNerve:
If people want to spam Monero with garbage, that's their perogative. Monero's job is to do its best to be a currency. Not to cater to garbage. This PR *disincentives* garbage, yet does move people from one type of garbage to a type of garbage arguably worse for privacy. I'm not mad at Monero for disincentivizing large file uploads in a TX by not accepting such TXs. I'm mad at idiots putting large files onto Monero. Use Bitcoin. Use Filecoin. Use torrents. Use IPFS. There's no advantage to the Monero protocol for this, at all.

I'd also point out, as prior pointed out, that by gonbatfire's interpretation the opposite of this patch would *increase* privacy, so we *should* adopt a 1 GB TX size limit so anyone uploading 1 GB files doesn't damage privacy by splitting across 10,000 TXs with a minimum of 20k outputs. I'd say don't put 1 GB files on Monero. Apparently some people may say differently.

Gonbatfire:
@kayabaNerve  I actually do think that it serves no point to limit tx_extra in any way (minus the blocksize limit of course), if we don't want people putting 1GB files on Monero, then fees should reflect that (and thankfully they do, it's expensive to put 1gb on-chain)

I would prioritize privacy over scalability, this patch has done the opposite, I just wanted to point it out.

I suppose it isn't useful to continue this until an actual dedicated attacker comes, so we will deal with it when time comes I suppose. Maybe by the time arbitrary data storage catches on we will already have full membership proofs, hopefully.

plowsof:
Following up to my earlier test. an image of 985 bytes was broadcast / mined on a node running this patch. my understanding of this patch is wrong, i assumed it would limit sending an image of 32 bytes, or, 255 bytes in 1 tx if you split the image over all outputs. Although small / low quality - a 985 byte image is still 'something to work with'.

PR #8724, opened 2023-01-31
wallet, simplewallet: Drop support for mine-to-use RPC system

jeffro256:
The main thing to test here is new wallet UX with old RPC mine-to-use node. Using post-PR wallet with pre-PR node should generate error message specific to this PR and not generic "command failed" mesages.

Issue #8624, opened 2022-10-25
Discussion: Community Nodes and Request For Volunteers


jeffro256:
## Who this affects

People who use Monero GUI wallet in Simple Mode and Simple Bootstrap Mode.

## How Monero GUI Simple Mode works Now

![Monero Community Nodes](https://user-images.githubusercontent.com/10839482/197677792-6f155ef3-efb1-41cf-a70f-0c334a7817d7.jpg)

On the wallet mode selection screen, if the user chooses Simple Mode, Simple Bootstrap Mode, or Advanced mode, the value of `persistentSettings.walletMode` [is](https://github.com/monero-project/monero-gui/blob/aef4a982dc644b3e971a89c8b65dd57d21d5e085/wizard/WizardModeSelection.qml#L99) [set](https://github.com/monero-project/monero-gui/blob/aef4a982dc644b3e971a89c8b65dd57d21d5e085/wizard/WizardModeSelection.qml#L53) [to](https://github.com/monero-project/monero-gui/blob/aef4a982dc644b3e971a89c8b65dd57d21d5e085/main.qml#L2271) 0, 1, or 2, respectively. This value is used to [set the flags of the `monerod` process](https://github.com/monero-project/monero-gui/blob/aef4a982dc644b3e971a89c8b65dd57d21d5e085/main.qml#L706-L708) that Monero GUI spins up in the background. For Simple Mode and Simple Bootstrap Mode (0 and 1), the `--bootstrap-daemon-address` option is set to `auto`. In Simple Mode (0), the `--no-sync` flag is also set.

The `--bootstrap-daemon-address` option is a feature of the core RPC server. When a core RPC server with a bootstrap daemon receives an RPC request, in which block data it does not have is requested for, it will pass that request though to the bootstrap daemon. In this manner, you can seamlessly interact with an unsynced `monerod` process as if it was fully synced with the network. When a `monerod` process is in `--no-sync` mode, this means that virtually all data requests will be passed off its bootstrap node, since it has no block data synced.

When the daemon has the bootstrap daemon address set to `auto`, as it does in Simple Mode and Simple Bootstrap Mode (0 and 1), `monerod` will search the p2p network list, filtering out remote nodes with an open RPC port. These remote nodes are ranked by the local `monerod` process based on how responsive they are plus a handful of sanity checks (e.g. whether the remote node's block head is higher than ours). Then a random node is picked to be the bootstrap node, unless it gets punished in the rankings.

[...]

PR #8734, opened 2023-02-06
wallet: move connection control into NodeRPCProxy

jeffro256:
This PR should not introduce any behavioral differences to the wallet code whatsoever. However, this PR simplifies connection handling and invocation in wallet2, and will allow for `NodeRPCProxy` to seamlessly switch out connections underneath `wallet2` without interrupting the wallet-y stuff. This is useful when the wallet wants to point to a "group" of potential daemons (think Electrum public nodes or the community node proposal #8624). 

Currently `wallet2` maintains a unique pointer to an abstract http client and `m_daemon_rpc_mutex` locks invokations to one-at-a-time. `wallet2` also maintains the "offline" status, daemon addresses, trusted/untrusted, login info, etc, as member fields. This PR moves that responsibility to `NodeRPCProxy`. Invokations are now also done through `NodeRPCProxy`.

The basic idea for future PRs will be to add a certain `tools::connection_settings` value that, when passed to `set_daemon`, will cause `NodeRPCProxy::get_transport_ref()` to reference a transport which manages connections to multiple different daemons and their penalty scores, not unlike `cryptonote::bootstrap_daemon`. 

Depends on #8724


jeffro256:
Rebased

PR #8735, opened 2023-02-07
serialization: remove unused function dump_json

jeffro256:


selsta (approved):

selsta:
@jeffro256  Can you double check if the mining RPC test fail is unrelated?

```
AssertionError: {'error': {'code': -32701, 'message': 'Params is not an object'}, 'id': 0, 'jsonrpc': '2.0'}
```

jeffro256:
That error should be caused by: https://github.com/monero-project/monero/commit/b67e931843cd0c8191ec49f80448c010cba39901. I'll rebase for good measure

selsta (approved):
You are right, I remember again.

selsta (approved):

PR #8750, opened 2023-02-26
build: fix BUILD_DEBUG_UTILITIES

jeffro256:
As you can see in [this post](https://stackoverflow.com/questions/44719963/cmake-default-value-for-options-doesnt-work), CMake options are cached and the values are not updated with the `option` command if they already exist in the cache. The effect of this for debug builds is that when a debug build is run after a release build, the debug utilities are not built. This PR makes it so that the debug utilities are build always during and only during debug builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -412,7 +412,7 @@ if (CMAKE_BUILD_TYPE STREQUAL "Debug")
 else()
   set(DEFAULT_BUILD_DEBUG_UTILITIES OFF)
 endif()
-option(BUILD_DEBUG_UTILITIES "Build debug utilities." DEFAULT_BUILD_DEBUG_UTILITIES)
+option(BUILD_DEBUG_UTILITIES "Build debug utilities." ${DEFAULT_BUILD_DEBUG_UTILITIES})
 
 if(OSSFUZZ)
   message(STATUS "Using OSS-Fuzz fuzzing system")


CMakeLists.txt
```diff
@@ -408,11 +408,10 @@ enable_testing()
 option(BUILD_DOCUMENTATION "Build the Doxygen documentation." ON)
 option(BUILD_TESTS "Build tests." OFF)
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")
-  set(DEFAULT_BUILD_DEBUG_UTILITIES ON)
+  set(BUILD_DEBUG_UTILITIES ON)
 else()
-  set(DEFAULT_BUILD_DEBUG_UTILITIES OFF)
+  set(BUILD_DEBUG_UTILITIES OFF)
 endif()
-option(BUILD_DEBUG_UTILITIES "Build debug utilities." DEFAULT_BUILD_DEBUG_UTILITIES)
```
selsta, line 415:
```cmake
set(BUILD_DEBUG_UTILITIES DEFAULT_BUILD_DEBUG_UTILITIES CACHE BOOL "Build debug utilities." FORCE)
```

This should overwrite the cache due to the `FORCE` keyword.

jeffro256: Is this for those who already have a CMake cache? If so, I could `unset(CACHE)` the variable before calling `set()`, so that the option doesn't appear in their cache despite it having no effect.

selsta:
I'll need some time to test this locally, I'm confused now :D

jeffro256:
Sorry haha I might not have made it clear: the point of this PR is to not make `BUILD_DEBUG_UTILITIES` an option and have it depend solely on whether or not we're doing a debug build. This fixes the issue of running a debug build after a release build, but now `BUILD_DEBUG_UTILITIES` isn't configurable.

jeffro256:
To test the bad behavior:
```
rm build/Linux/master/release/CMakeCache.txt
rm build/Linux/master/debug/CMakeCache.txt
make -j$(nproc) release
make -j$(nproc) debug
ll build/Linux/master/debug/bin/ | grep monero-utils
```

As you can see, the debug utilities are not built

selsta:
Right, that's how I understood it that's why I suggested this as the whole patch:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8c088f26d..acee2214e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -412,7 +412,7 @@ if (CMAKE_BUILD_TYPE STREQUAL "Debug")
 else()
   set(DEFAULT_BUILD_DEBUG_UTILITIES OFF)
 endif()
-option(BUILD_DEBUG_UTILITIES "Build debug utilities." DEFAULT_BUILD_DEBUG_UTILITIES)
+set(BUILD_DEBUG_UTILITIES ${DEFAULT_BUILD_DEBUG_UTILITIES} CACHE BOOL "Build debug utilities." FORCE)
 
 if(OSSFUZZ)
   message(STATUS "Using OSS-Fuzz fuzzing system")
```

This still allows people to build a release build with debug utilities and solves the issue you are describing in your previous comment. I confirmed it with the steps you posted.

jeffro256:
We could add a global cache option `ALWAYS_BUILD_DEBUG_UTILITIES` which defaults to `OFF` and set `BUILD_DEBUG_UTILITIES` equal to `CMAKE_BUILD_TYPE STREQUAL "Debug" OR ALWAYS_BUILD_DEBUG_UTILITIES`. So the debug utilities are always built in debug mode, but the user can also set them to build in release mode. I think that would be the best/clearest solution.

hyc:
Since release and debug have their own separate trees, and separate CMakeCache.txt files, I'm not seeing the linkage - how does the contents of the release cmakecache affect the debug cmakecache?

jeffro256:
That's exactly what I was thinking. Were you also able to reproduce the behavior? But also, just in principle, the debug builds should always include the debug utilities IMO.

jtgrassie:
@hyc is correct. The bug isn't actually `option` caching, it's the CMakeList not using a variable reference.

I found that by simply changing:

```diff
-option(BUILD_DEBUG_UTILITIES "Build debug utilities." DEFAULT_BUILD_DEBUG_UTILITIES)
+option(BUILD_DEBUG_UTILITIES "Build debug utilities." ${DEFAULT_BUILD_DEBUG_UTILITIES})
```

"fixed" the issue (make release / debug in any order picked up the intended default build of utilities).

jeffro256:
Wow... So I'm guessing that the `DEFAULT_BUILD_DEBUG_UTILITIES` value in the command `option(BUILD_DEBUG_UTILITIES "Build debug utilities." DEFAULT_BUILD_DEBUG_UTILITIES)` meant nothing to CMake and so `option()` defaulted to making a boolean option with the default value set to `NO`. Did CMake just silently ignore the value passed even though its a syntax error?

Can't believe I didn't notice the missing `${}` haha. I guess since this will fix the issue of defaulting to building debug utilities when in a debug build until explicitly modified by a user, it works for me.

selsta (approved):

jtgrassie (approved):

jtgrassie:
@jeffro256 
> Did CMake just silently ignore the value passed even though its a syntax error?

It's not a syntax error, it's just not a variable _reference_ without the `${}`, so the _option_ was being set to false/off (because the _string_ "DEFAULT_BUILD_DEBUG_UTILITIES" is not `ON|TRUE|1|YES|Y`).

Without the [fix above](https://github.com/monero-project/monero/pull/8750#issuecomment-1445429804), the debug utilities aren't built without an explicit set of `BUILD_DEBUG_UTILITIES` (e.g. `cmake -DBUILD_DEBUG_UTILITIES=ON ...`).

tobtoht (approved):

jeffro256:
Interesting... "A quoted string always evaluates to false unless: The string's value is one of the true constants,  ..." https://cmake.org/cmake/help/latest/command/if.html#basic-expressions. I did not know that.

PR #8751, opened 2023-02-28
replace deprecated boost copy_option

eversinc33:
As of boost 1.74.0, `boost::filesystem::copy_option` is deprecated and replaced by `boost::filesystem::copy_options`. See https://www.boost.org/doc/libs/1_75_0/libs/filesystem/doc/release_history.html

I replaced all occurrences with the new `copy_options` to avoid problems in the future when `copy_option` will be removed.

### DIFF
--- a/src/p2p/net_peerlist.cpp
+++ b/src/p2p/net_peerlist.cpp
@@ -200,7 +200,7 @@ namespace nodetool
     if (!out)
     {
       // if failed, try reading in unportable mode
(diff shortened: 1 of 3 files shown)


jtgrassie:
Our current minimum boost version is 1.58, which this PR doesn't support.

eversinc33:
> Our current minimum boost version is 1.58, which this PR doesn't support.

Did not think of this. Fair point!

PR #8752, opened 2023-02-28
trezor: support v2.5.2+, more tests, fix chaingen, static libusub fix

ph4r05:
- passphrase logic: remove backward compatibility for 2.4.3, code cleanup
- fix LibUSB cmake for static builds on OSX, FreeBSD
- fix Trezor support for FreeBSD, Android. Failt build if Trezor support is enabled but dependencies are missing
- tests: all Trezor tests now work with passphrase logic enabled by default. Passphrase test added with different passphrase. no_passphrase test added, Trezor pin test added. Testing wallet opening with correct and incorrect passphrase. Trezor test chain revamp, cleanup. Smaller chain, chain file versioning added (most of the changes in the PR).
- tests: Trezor tests support TEST_MINING_ENABLED, TEST_MINING_TIMEOUT env vars to change mining-related tests behaviour.
- monero tests: fix tx create from sources, input locking. Originally, creating a synthetic transactions with chaingen could create a transaction with outputs that are still locked in the current block, thus failing chain validation by the daemon. Simple unlock check was added. Some buggy tests were fixed as well as new unlock-checking version of tx creation rejected those, fixes are simple - mostly using correct block after a rewind to construct a transaction

As there are occasionally reported Trezor integration bugs (https://github.com/monero-project/monero-gui/issues/4105), this PR removes some old code, forces users to use the newest Trezor firmware and adds more tests. 

Tests are refactored so the passphrase flow is used for all transactions, besides one special test so we cover all code paths. 

------------------------
## Protobuf

Protobuf v21 is the latest compatible protobuf version, as Protobuf v23 requires C++17 (through its dependency Abseil library). 

To disable Protobuf version check and later compilation test, define `-DUSE_DEVICE_TREZOR_PROTOBUF_TEST=OFF`. But the compilation may fail later with less descriptive error.

To build with unlinked protobuf@21 on OSX:
```shell
CMAKE_PREFIX_PATH=$(find /opt/homebrew/Cellar/protobuf@21 -maxdepth 1 -type d -name "21.*" -print -quit) \
make debug-test-trezor -j8
```

Similarly on other systems, point CMAKE_PREFIX_PATH to Protobuf v21 installation.

OSX install:
```shell
brew install protobuf@21 
brew link protobuf@21
```

MSYS32:
```shell
curl -O https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-protobuf-c-1.4.1-1-any.pkg.tar.zst
curl -O https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-protobuf-21.9-1-any.pkg.tar.zst
pacman --noconfirm -U mingw-w64-x86_64-protobuf-c-1.4.1-1-any.pkg.tar.zst mingw-w64-x86_64-protobuf-21.9-1-any.pkg.tar.zst
```


tests/trezor/trezor_tests.cpp
```diff
@@ -360,6 +381,7 @@ static void setup_chain(cryptonote::core * core, gen_trezor_base & trezor_base,
     {
       trezor_base.clear();
       generated = trezor_base.generate(events);
+      trezor_base.fix_hf(events);
```
ph4r05, line 384: optimization: generate blockchain file up to the minimal HF version being tested. Previously was this step performed at runtime after loading the chain file.

tests/trezor/trezor_tests.cpp
```diff
@@ -958,7 +1018,11 @@ void gen_trezor_base::fix_hf(std::vector<test_event_entry>& events)
     ADD_HARDFORK(m_hard_forks, hf_to_add, hardfork_height);
     add_top_hfork(events, m_hard_forks);
     MDEBUG("Hardfork added at height: " << hardfork_height << ", from " << (int)current_hf << " to " << (int)hf_to_add);
-    rewind_blocks(events, 10, hf_to_add);
+    rewind_blocks(events, hf_to_add == m_top_hard_fork ? 10 : 2, hf_to_add);
```
ph4r05, line 1021: Wallet requires +10 blocks to activate last HF, thus 10 block rewind to get the latest HF active.

tests/trezor/trezor_tests.cpp
```diff
@@ -1931,14 +2078,20 @@ bool wallet_api_tests::generate(std::vector<test_event_entry>& events)
...
+  std::unique_ptr<Monero::Wallet> w{wmgr->createWalletFromDevice(wallet_path, "", api_net_type, m_trezor_path, 1, "", 1, this)};
   CHECK_AND_ASSERT_THROW_MES(w->init(daemon()->rpc_addr(), 0), "Wallet init fail");
 
   auto walletImpl = dynamic_cast<Monero::WalletImpl *>(w.get());
   CHECK_AND_ASSERT_THROW_MES(walletImpl, "Dynamic wallet cast failed");
   WalletApiAccessorTest::allow_mismatched_daemon_version(walletImpl, true);
+  walletImpl->setTrustedDaemon(true);
 
   CHECK_AND_ASSERT_THROW_MES(w->refresh(), "Refresh fail");
+
```
ph4r05, line 2090: This test worked previously fine by coincidence. By changing chain structure wallet API created double-spend transactions. We have to scan the wallet. 

The code now demonstrates the correct flow and tests also key image sync via API.

tests/core_tests/chaingen.h
```diff
@@ -426,13 +426,16 @@ bool construct_miner_tx_manually(size_t height, uint64_t already_generated_coins
                                  uint64_t fee, uint8_t hf_version = 1,
                                  cryptonote::keypair* p_txkey = nullptr);
 
+extern bool tx_construction_check_unlock_time;
```
ph4r05, line 429: This is not ideal. It is a global default value for `bool check_unlock_time` parameter to `construct_tx_to_key` functions below. By default, it is set to false so backward compatibility with all core_tests is not affected. (Setting it to true causes few tests to fail).

But for Trezor tests I need transaction to pass full daemon validation, i.e., there cannot be fake outputs which are still locked. This global variable enables Trezor tests to globally enable unlock time checking. Function caller can still provide own `check_unlock_time` value.

Another approach would be to modify all TX creation macros, create extended versions which would enable the unlock time checking. It would increase code size significantly.

tests/trezor/trezor_tests.cpp
```diff
@@ -1931,38 +2195,54 @@ bool wallet_api_tests::generate(std::vector<test_event_entry>& events)
...
+  uint64_t spent, unspent;
+  walletImpl->coldKeyImageSync(spent, unspent);
+  CHECK_AND_ASSERT_THROW_MES(walletImpl->rescanSpent(), "Rescan spent fail");
```
ph4r05, line 2210: This test worked previously fine by coincidence. By changing chain structure wallet API created double-spend transactions. We have to scan the wallet. 

The code now demonstrates the correct flow and tests also key image sync via API.

tests/core_tests/chaingen.h
ph4r05: Change: I fixed all related tests instead, resolving

ph4r05:
@selsta pinging to note this is ready. Diff sounds big, but most of if is just my trezor test code.

ph4r05:
I have to look into build test fails - probably newly enforced libusb requirement

ph4r05:
hey @hyc @TheCharlatan, pls, I noticed you configured most of the build for cross compile to OSX. 

I am currently struggling with adding LibUSB so it is correctly picked up by the Monero build (I've added LibUSB as mandatory dependency for Trezor).

LibUsb1.0.26 compilation test fails due to undefined references due to `___isOSVersionAtLeast`, so I tried to change OSX SDK from 11.1 to 12.3, but I am entering dependency hell with compilation problems with Boost, Clang and OSX SDK. I was wondering if you could perhaps have an idea if `___isOSVersionAtLeast` is caused by old SDK and what would be the best way forward. 

Thanks for any suggestions!

selsta:
Updating the macOS SDK causes all kinds of backwards compatibility issues :/

My the others have an idea how to solve this without updating the SDK.

ph4r05:
> Updating the macOS SDK causes all kinds of backwards compatibility issues :/
> 
> My the others have an idea how to solve this without updating the SDK.

Ah I see, maybe lowering libusb version will help.

ph4r05:
I had to downgrade LibUSB from 1.0.26 to 1.0.23 as higher versions won't cross-compile under OSX SDK 11.1, clang-9. 

We have to check if there are any known vulnerabilities in 1.0.23. But it would be great to migrate to newer SDK and cland versions.

ph4r05:
I will try patching libusb-1.0.26 so it builds in the current setup.

ph4r05:
Done, libusb-1.0.26 is patched so it cross-compiles also on OSX in the current setup. Ready for review when you have a minute guys 🙏  @selsta

ph4r05:
@moneromooo-monero what do you think pls? Shall I split this PR? Would you pls find some time to review it? Thanks! 🙏

ph4r05:
@selsta pls do you have any idea what could I do for this PR to get it reviewed? 😅 🙏

selsta:
@ph4r05 I'll try to get it reviewed.

ph4r05:
👀

selsta:
@ph4r05 I did not forget about this, it's just a large change and as far as I can see it's nothing we have to prioritise currently.

plowsof:
john_r365 has opened up a bounty to review this issue. It is currently at 1xmr. details can be found [here](https://bounties.monero.social/posts/85/1-000m-bounty-for-reviewing-trezors-latest-pull-request)

SChernykh:
@selsta @plowsof I'm going to start reviewing it today, and plan to finish the review by Sunday.

plowsof:
@SChernykh ACK - marked bounty as started by sech1 , thank you

tests/core_tests/block_reward.cpp
```diff
@@ -172,21 +172,21 @@ bool gen_block_reward::generate(std::vector<test_event_entry>& events) const
     return false;
 
   // Test: fee increases block reward
-  transaction tx_0(construct_tx_with_fee(events, blk_5, miner_account, bob_account, MK_COINS(1), 3 * TESTS_DEFAULT_FEE));
+  transaction tx_0(construct_tx_with_fee(events, blk_5r, miner_account, bob_account, MK_COINS(1), 3 * TESTS_DEFAULT_FEE));
```
SChernykh, line 175: Can you give a rationale of changing `blk_5` to `blk_5r` here, and for similar other changes in the tests? This code has never been changed before, so for sure the current Trezor changes shouldn't require it. If it's some bug in tests, I'd like to know where it was and why it required such changes.

SChernykh: > monero tests: fix tx create from sources, input locking. Originally, creating a synthetic transactions with chaingen could create a transaction with outputs that are still locked in the current block, thus failing chain validation by the daemon. Simple unlock check was added. Some buggy tests were fixed as well as new unlock-checking version of tx creation rejected those, fixes are simple - mostly using correct block after a rewind to construct a transaction

I've compiled this PR and I'm playing with tests now, to look how this all works because I didn't touch this code before.

src/device_trezor/device_trezor_base.hpp
```diff
@@ -227,9 +227,9 @@ namespace trezor {
         }
 
         if (network_type){
-          msg->set_network_type(static_cast<uint32_t>(network_type.get()));
+          msg->set_network_type(static_cast<messages::monero::MoneroNetworkType>(network_type.get()));
         } else {
-          msg->set_network_type(static_cast<uint32_t>(this->network_type));
+          msg->set_network_type(static_cast<messages::monero::MoneroNetworkType>(this->network_type));
```
SChernykh, line 232: Having `network_type` as a function parameter and as a class member is confusing, which is why you had to write `this->network_type` here. It's better to rename the class member to `m_network_type` to avoid further confusion.

tests/core_tests/chaingen.cpp
```diff
@@ -720,15 +730,17 @@ void block_tracker::get_fake_outs(size_t num_outs, uint64_t amount, uint64_t glo
...
     if (used.find(oi_idx) != used.end())
       continue;
+    if (fnc_accept && !(fnc_accept.get())({.oi=oi, .cur_height=cur_height}))
+      continue;
 
     rct::key comm = oi.commitment();
     auto out = boost::get<txout_to_key>(oi.out);
     auto item = std::make_tuple(oi.idx, out.key, comm);
-    outs.push_back(item);
+    outs.emplace_back(item);
```
SChernykh, line 743: The above 2 lines can be changed to `outs.emplace_back(oi.idx, out.key, comm);`

tests/core_tests/wallet_tools.cpp
```diff
@@ -119,18 +132,12 @@ bool wallet_tools::fill_tx_sources(tools::wallet2 * wallet, std::vector<cryptono
...
 
       // Acceptor function
-      if (fnc_accept){
-        tx_source_info_crate_t c_info{.td=&td, .src=&src, .selected_idx=&selected_idx, .selected_kis=&selected_kis,
-            .ntrans=ntrans, .iters=iters, .sum=sum, .cur_utxo=cur_utxo};
-
-        bool take_it = (fnc_accept.get())(c_info, abort);
-        if (!take_it){
+      if (fnc_accept && !(fnc_accept.get())({.td=&td, .src=&src, .selected_idx=&selected_idx, .selected_kis=&selected_kis,
+                                              .ntrans=ntrans, .iters=iters, .sum=sum, .cur_utxo=cur_utxo}, abort))
```
SChernykh, line 139: `abort` is a standard function defined in `cstdlib`, it's better to rename this variable.

tests/core_tests/wallet_tools.cpp
```diff
@@ -67,11 +79,12 @@ void wallet_tools::process_transactions(tools::wallet2 * wallet, const std::vect
     wallet_tools::gen_block_data(bt, bl, mtx, v_bche.back(), v_parsed_block.back(), idx == 1 ? start_height : height);
```
SChernykh, line 79: Not a bug that was introduced in this PR, but if this loop iterates more than once, it will use `height` which is never initialized.

tests/core_tests/wallet_tools.h
```diff
@@ -68,12 +68,14 @@ class WalletApiAccessorTest
...
+  static void gen_tx_src(size_t mixin, uint64_t cur_height, const tools::wallet2::transfer_details & td, cryptonote::tx_source_entry & src, block_tracker &bt, const boost::optional<fnc_accept_output_t>& fnc_accept = boost::none);
   static void gen_block_data(block_tracker &bt, const cryptonote::block *bl, const map_hash2tx_t & mtx, cryptonote::block_complete_entry &bche, tools::wallet2::parsed_block &parsed_block, uint64_t &height);
   static void compute_subaddresses(std::unordered_map<crypto::public_key, cryptonote::subaddress_index> &subaddresses, cryptonote::account_base & creds, size_t account, size_t minors);
   static void process_transactions(tools::wallet2 * wallet, const std::vector<test_event_entry>& events, const cryptonote::block& blk_head, block_tracker &bt, const boost::optional<crypto::hash>& blk_tail=boost::none);
+  static void process_transactions(std::vector<tools::wallet2*> wallets, const std::vector<test_event_entry>& events, const cryptonote::block& blk_head, block_tracker &bt, const boost::optional<crypto::hash>& blk_tail=boost::none);
```
SChernykh, line 75: `wallets` should be const reference: `const std::vector<tools::wallet2*>& wallets`

tests/core_tests/wallet_tools.h
```diff
@@ -68,12 +68,14 @@ class WalletApiAccessorTest
...
   static void process_transactions(tools::wallet2 * wallet, const std::vector<const cryptonote::block*>& blockchain, const map_hash2tx_t & mtx, block_tracker &bt);
-  static bool fill_tx_sources(tools::wallet2 * wallet, std::vector<cryptonote::tx_source_entry>& sources, size_t mixin, const boost::optional<size_t>& num_utxo, const boost::optional<uint64_t>& min_amount, block_tracker &bt, std::vector<size_t> &selected, uint64_t cur_height, ssize_t offset=0, int step=1, const boost::optional<fnc_accept_tx_source_t>& fnc_accept=boost::none);
+  static void process_transactions(std::vector<tools::wallet2*> wallets, const std::vector<const cryptonote::block*>& blockchain, const map_hash2tx_t & mtx, block_tracker &bt);
```
SChernykh, line 77: Same, `wallets` should be const reference: `const std::vector<tools::wallet2*>& wallets`

tests/core_tests/block_reward.cpp
ph4r05: Thanks for starting the review! ❤️ 
 Regarding the blk_5r question - I had to rewind so tests pass after fix of tx lock time.

ph4r05:
@SChernykh thanks for looking into this! I will address the comments soon

SChernykh:
I just did a quick glance over the code today, I'll do a deeper look over the weekend.

tests/trezor/trezor_tests.cpp
```diff
@@ -385,6 +450,10 @@ static void setup_chain(cryptonote::core * core, gen_trezor_base & trezor_base,
   }
 }
 
+static fnc_accept_output_t fnc_trezor_default_acceptor_tx_in = [] (const fnc_accept_output_crate_t &info) -> bool {
+  return info.oi.is_coin_base || info.oi.blk_height + CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE <= info.cur_height + 1;
```
SChernykh, line 454: I'm confused about this `+1` here and in other places. In wallet2.cpp, there is
```
uint64_t unlock_height = td.m_block_height + std::max<uint64_t>(CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE, CRYPTONOTE_LOCKED_TX_ALLOWED_DELTA_BLOCKS);
```
which means that if a transaction was mined in block `N`, its outputs can be spent in block `N+CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE`, but here you add additional 1. So, with current settings, a transaction mined in block 2914616 can be spent in block 2914626.
Or is `cur_height` a different height?

tests/trezor/trezor_tests.cpp
```diff
@@ -393,6 +462,20 @@ static device_trezor_test *ensure_trezor_test_device(){
...
 }
 
+static size_t min_mixin_for_hf(uint8_t hf)
+{
+  if (hf >= HF_VERSION_MIN_MIXIN_15)
+    return 15;
+  if (hf >= 8)
+    return 10;
+  return 0;
+}
```
SChernykh, line 472: From what I found in `blockchain.cpp`: https://github.com/monero-project/monero/blob/master/src/cryptonote_core/blockchain.cpp#L3337
It should be:
```
  if (hf >= 2)
    return hf >= HF_VERSION_MIN_MIXIN_15 ? 15 : hf >= HF_VERSION_MIN_MIXIN_10 ? 10 : hf >= HF_VERSION_MIN_MIXIN_6 ? 6 : hf >= HF_VERSION_MIN_MIXIN_4 ? 4 : 2;

  return 0;

File: tests/trezor/trezor_tests.cpp [hunk 15]
```diff
@@ -426,6 +509,52 @@ static crypto::public_key get_tx_pub_key_from_received_outs(const tools::wallet2
   }
 }
 
+static void generate_tx_block(std::vector<test_event_entry>& events, std::list<cryptonote::transaction> &txs_lst, cryptonote::block & head,
+                              const cryptonote::account_base& from, const cryptonote::account_base& to, uint64_t amount, size_t repeat,
+                              size_t mixin, bool rct, rct::RangeProofType range_proof_type, int bp_version)
+{
+  const size_t per_tx_limit = 14;
+  const size_t iters = std::ceil(repeat / (double)per_tx_limit);
```
SChernykh, line 517: If `repeat` is 0, `iters` will also be 0. I would do `repeat ? std::ceil(repeat / (double)per_tx_limit) : 1` to guarantee at least 1 iteration of the loop.

tests/trezor/trezor_tests.cpp
```diff
@@ -625,24 +782,31 @@ gen_trezor_base::gen_trezor_base(const gen_trezor_base &other):
     m_bob_account(other.m_bob_account), m_alice_account(other.m_alice_account), m_eve_account(other.m_eve_account),
     m_hard_forks(other.m_hard_forks), m_trezor(other.m_trezor), m_rct_config(other.m_rct_config), m_top_hard_fork(other.m_top_hard_fork),
     m_heavy_tests(other.m_heavy_tests), m_test_get_tx_key(other.m_test_get_tx_key), m_live_refresh_enabled(other.m_live_refresh_enabled),
-    m_network_type(other.m_network_type), m_daemon(other.m_daemon)
+    m_network_type(other.m_network_type), m_daemon(other.m_daemon), m_alice2_account(other.m_alice2_account),
+    m_trezor_path(other.m_trezor_path), m_trezor_pin(other.m_trezor_pin), m_trezor_passphrase(other.m_trezor_passphrase),
+    m_trezor_use_passphrase(other.m_trezor_use_passphrase), m_trezor_use_alice2(other.m_trezor_use_alice2),
+    m_gen_heavy_test_set(other.m_gen_heavy_test_set)
```
SChernykh, line 788: `m_head`, `m_events`, `m_mining_enabled`, `m_mining_timeout` and perhaps some other fields are missing in this copy constructor.

tests/trezor/trezor_tests.cpp
```diff
@@ -658,13 +822,22 @@ void gen_trezor_base::fork(gen_trezor_base & other)
...
+  other.m_mining_enabled = m_mining_enabled;
+  other.m_mining_timeout = m_mining_timeout;
+
   other.m_miner_account = m_miner_account;
   other.m_bob_account = m_bob_account;
   other.m_alice_account = m_alice_account;
+  other.m_alice2_account = m_alice2_account;
   other.m_eve_account = m_eve_account;
   other.m_trezor = m_trezor;
   other.m_generator.set_events(&other.m_events);
```
SChernykh, line 843: `m_no_change_in_tested_tx` is not copied in this function.

tests/trezor/trezor_tests.cpp
```diff
@@ -745,6 +947,7 @@ bool gen_trezor_base::generate(std::vector<test_event_entry>& events)
   // Another block with predefined timestamp.
   // Carefully set reward and already generated coins so it passes miner_tx check.
   cryptonote::block blk_0;
+  cryptonote::block & lask_block = blk_0;
```
SChernykh, line 950: A typo? `last_block`?

src/device_trezor/device_trezor_base.cpp
```diff
@@ -475,20 +472,9 @@ namespace trezor {
...
-      }
-
       m_seen_passphrase_entry_message = true;
       bool on_device = true;
-      if (msg->has__on_device() && !msg->_on_device()){
-        on_device = false;  // do not enter on device, old devices.
-      }
-
-      if (on_device && m_features && m_features->capabilities_size() > 0){
+      if (m_features && m_features->capabilities_size() > 0){
```
SChernykh, line 477: If `capabilities_size()` is 0 (no capabilities at all), this code will leave `on_device` as true. But if there are some capabilities, but none of them are `Capability_PassphraseEntry`, it will set `on_device` to false. I don't know the specifications, but it doesn't seem logical to me.

SChernykh: I think it's safe to remove `m_features->capabilities_size() > 0` check because if it's empty, the loop inside will not run any iterations, and `on_device` will be set to false.

tests/core_tests/chaingen.cpp
```diff
@@ -955,10 +969,14 @@ cryptonote::tx_destination_entry build_dst(const var_addr_t& to, bool is_subaddr
...
 
-std::vector<cryptonote::tx_destination_entry> build_dsts(const var_addr_t& to1, bool sub1, uint64_t am1)
+std::vector<cryptonote::tx_destination_entry> build_dsts(const var_addr_t& to1, bool sub1, uint64_t am1, size_t repeat)
 {
   std::vector<cryptonote::tx_destination_entry> res;
-  res.push_back(build_dst(to1, sub1, am1));
+  res.reserve(repeat);
+  for(size_t i = 0; i < repeat; ++i)
+  {
+    res.push_back(build_dst(to1, sub1, am1));
```
SChernykh, line 978: `emplace_back` will be more efficient here.

tests/core_tests/wallet_tools.cpp
ph4r05: in `gen_block_data` the height parameter is `uint64_t &height` and it set immediately (out param), it is ok then, no?

SChernykh: Ah, yes, I didn't notice it. Then it's ok.

tests/trezor/trezor_tests.cpp
ph4r05: Our Trezor integration does not support HF below 8 (i.e., `HF_VERSION_MIN_MIXIN_10`) so we are not using it also in our test code. Is that ok?

tests/core_tests/wallet_tools.h
ph4r05: I've got this warning:
> Clang-Tidy: Parameter 'wallets' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions

SChernykh: You probably didn't change it in the function definition. It should be the same in both places.

src/device_trezor/device_trezor_base.cpp
ph4r05: This has it's own history 😅 
https://docs.trezor.io/trezor-firmware/common/communication/passphrase-redesign-migration.html

tldr for quite a while there were no capabilities. Our firmware supports only Model TT, which has passphrase entry. Thus the default value=true. 

Then caps were added in new firmwares. When caps are there, we reset `on_device` to false and set back to true only if given cap is present. Thus, it made sense as old firmware -> no caps -> passphrase entry by default.

But now we dont support those old firmares anymore, so I will remove it.

tests/trezor/trezor_tests.cpp
ph4r05: `cur_height` in `fill_tx_sources` in 

```
if (fnc_accept && !(fnc_accept.get())({.oi=oi, .cur_height=head_height}))
```

is current height of the head, but we are looking for a transaction spendable in the next block, thus `cur_height + 1`

tests/trezor/trezor_tests.cpp
```diff
@@ -290,8 +326,14 @@ static bool serialize_chain_to_file(std::vector<test_event_entry>& events, gen_t
...
+      a << CUR_CHAIN_VERSION;
+      a << opts.heavy_tests;
+      a << CUR_CHAIN_MAGIC;
+
       a << events;
       a << test_base;
+      a << CUR_CHAIN_MAGIC;
       return !data_file.fail();
     }
     catch(...)
```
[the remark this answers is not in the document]
ph4r05: fixed

tests/core_tests/wallet_tools.h
ph4r05: maybe stale IDE analysis. added now

tests/core_tests/wallet_tools.h
ph4r05: maybe stale IDE analysis. added now

ph4r05:
heh now macosx build fails, suddenly

ph4r05:
hmm protobuf got updated and requires c++14 and higher, any plans for bumping this? :)
meanwhile, I pinned it to v21 with brew in GitHub actions builders

SChernykh:
> hmm protobuf got updated and requires c++14 and higher, any plans for bumping this? :) meanwhile, I pinned it to v21 with brew in GitHub actions builders

We do use c++14 to compile Monero binaries, but I can't find `set(CMAKE_CXX_STANDARD 14)` in CMakeLists.txt for the tests. Maybe you can try to add it and see if it compiles.

SChernykh:
The MacOS link errors are discussed here (with possible solution in the end): https://github.com/protocolbuffers/protobuf/issues/12292

tests/trezor/trezor_tests.cpp
```diff
@@ -797,85 +1000,106 @@ bool gen_trezor_base::generate(std::vector<test_event_entry>& events)
...
 
   // Simple transaction check
   bool resx = rct::verRctSemanticsSimple(txs_blk_5.begin()->rct_signatures);
   bool resy = rct::verRctNonSemanticsSimple(txs_blk_5.begin()->rct_signatures);
   CHECK_AND_ASSERT_THROW_MES(resx, "Tsx5[0] semantics failed");
   CHECK_AND_ASSERT_THROW_MES(resy, "Tsx5[0] non-semantics failed");
 
-  REWIND_BLOCKS_HF(events, blk_5r, blk_5, m_miner_account, CUR_HF);  // rewind to unlock
+  // Rewind blocks so there are more spendable coins
+  last_block = rewind_blocks_with_decoys(events, last_block, CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE + 2, CUR_HF, TEST_DECOY_MINING_COUNTS);
```
SChernykh, line 1055: `+ 2` was added in the recent commits, is it to pop more transactions into mempool? The other `rewind_blocks_with_decoys` further down doesn't have this `+ 2`.

ph4r05: correct, I had to increase number of transactions to pick from as the last test was failing, so there is enough buffer for that

SChernykh:
> Pls, when you have a minute, can you go over the comments and resolve those which are OK for you?

I don't see "resolved" button, I think you need to click it on the comments which you fixed already.

SChernykh:
I will check and approve today, probably late in the evening. So far, I don't have more comments.

.github/workflows/build.yml
```diff
@@ -39,7 +39,9 @@ jobs:
         key: ccache-${{ runner.os }}-build-${{ github.sha }}
         restore-keys: ccache-${{ runner.os }}-build-
     - name: install dependencies
-      run: HOMEBREW_NO_AUTO_UPDATE=1 brew install boost hidapi openssl zmq libpgm miniupnpc expat libunwind-headers protobuf ccache
+      run: |
+        HOMEBREW_NO_AUTO_UPDATE=1 brew install boost hidapi openssl zmq libpgm miniupnpc expat libunwind-headers protobuf@21 abseil ccache
```
selsta, line 43: Does this mean every macOS user has to now specifically install and link protobuf@21? This wasn't necessary previously and will break compilation for every Mac user who has the regular protobuf installed.

cmake/CheckTrezor.cmake
```diff
@@ -102,13 +108,14 @@ if(Protobuf_FOUND AND USE_DEVICE_TREZOR AND TREZOR_PYTHON)
         "${CMAKE_BINARY_DIR}/test-protobuf.pb.cc"
         "${CMAKE_CURRENT_LIST_DIR}/test-protobuf.cpp"
         CMAKE_FLAGS
+        CMAKE_EXE_LINKER_FLAGS ${CMAKE_TRY_COMPILE_LINKER_FLAGS}
         "-DINCLUDE_DIRECTORIES=${Protobuf_INCLUDE_DIR};${CMAKE_BINARY_DIR}"
-        "-DCMAKE_CXX_STANDARD=11"
-        LINK_LIBRARIES ${Protobuf_LIBRARY}
+        "-DCMAKE_CXX_STANDARD=17"
```
selsta, line 113: Is it possible to set this to 14? We don't support C++17 yet.

ph4r05: let's see :)

.github/workflows/build.yml
SChernykh: I'm fine with it, as long as it's in `contrib/brew/Brewfile` so MacOS users can build it while following instructions from readme.md

ph4r05: > Does this mean every macOS user has to now specifically install and link protobuf@21? This wasn't necessary previously and will break compilation for every Mac user who has the regular protobuf installed.

unfortunately, yes, new protobuf v23 requires newer cmake AFAIK (+some other deps). And we still dont have c++17. Will check what can be done in Brewfile

selsta: How does it work right now? I use 23 currently to compile monero.

selsta: I just want to make sure that there is no other way before breaking compilation for all Mac users.

selsta: macOS has the latest version of CMake installed using brew.

ph4r05: @selsta I will try and let you know. When I let CI build osx with Protobuf v23 it failed. There may be another locked dependencies

ph4r05: @selsta I am having the same problems as runner in GHA, when upgrading with brew, `make clean ; make debug-test-trezor -j8` ends with an error.

```
  /opt/homebrew/include/absl/utility/utility.h:164:12: error: no member named
  'in_place_t' in namespace 'std'

  using std::in_place_t;
```

and this is c++17 feature https://en.cppreference.com/w/cpp/utility/in_place

So unless monero supports c++17 (why not now?) it has to use old protobuf lib.

SChernykh: Maybe there's some macro definition you can supply to disable c++17 features in that code? I can't check right now.

SChernykh: @ph4r05 You can add `-DABSL_OPTION_USE_STD_OPTIONAL=0` to cmake command line (or add `add_definitions("-DABSL_OPTION_USE_STD_OPTIONAL=0")` to CMakeLists.txt) to disable C++17 in Abseil.

SChernykh: After checking some more, I can see that it should be done by Abseil automatically if you run the compiler with c++14 configured through the command line - `set(CMAKE_CXX_STANDARD 14)` in CMakeLists.txt should work.

ph4r05:
btw I've added

```
concurrency:
  group: ${{ github.repository }}-${{ github.ref }}-${{ github.workflow }}
  cancel-in-progress: true
```

to GHA, I believe this is the best practice. I was quite annoying that previous test runs kept runners busy even though new commits were pushed to the PR or it was force-updated. Especially when we have like 1.5hour test build. 

Is it ok?

tests/trezor/trezor_tests.cpp
ph4r05: @SChernykh can I pls resolve this one?

tests/core_tests/block_reward.cpp
ph4r05: @SChernykh can I pls resolve this one?

SChernykh:
> `cancel-in-progress: true`

I don't think it's good for the main repository. It can be good for quick testing/iterating in private repos, but when we submit things to the main repository, we want to have Github actions checked for every commit, even if they come in rapid succession.

tests/trezor/trezor_tests.cpp
ph4r05: @SChernykh can I pls resolve this one?

src/device_trezor/device_trezor_base.cpp
ph4r05: @SChernykh can I pls resolve this one?

.github/workflows/build.yml
ph4r05: @SChernykh `try_compile` fails with `"-DCMAKE_CXX_STANDARD=14"`, does not work. any suggestions?

ph4r05:
@SChernykh ok I've updated it according to the https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value, i.e., it cancels run only in PRs, other branches do build every commit

.github/workflows/build.yml
ph4r05: I mean, this is no 5 minute fix 😅  so maybe another PR to make it build with the latest protobuf? It is few hours of work, experimenting. Maybe not worth right now

ph4r05:
@SChernykh ok I've updated it according to the https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value, i.e., it cancels run only in PRs, other branches do build every commit
```
concurrency:
  group: ${{ github.repository }}-${{ github.head_ref || github.run_id }}-${{ github.workflow }}
  cancel-in-progress: true
```

tests/trezor/trezor_tests.cpp
SChernykh: Yes, and other unresolved comments too.

.github/workflows/build.yml
SChernykh: This whole protobuf/c++17 thing is not directly related to this PR, it compiled fine when you created this PR. So it can be fixed properly in another PR, I don't mind the current solution here.

ph4r05: @SChernykh it does not work with C++14, 
```
/opt/homebrew/include/absl/utility/utility.h:164:12: error: no member named 'in_place_t' in namespace 'std'
using std::in_place_t;
```

absl installation from homebrew has it turned on by default. Also `/opt/homebrew/include/absl/base/options.h`

```
#define ABSL_OPTION_USE_STD_OPTIONAL 1
```

meaning it needs C++17

SChernykh:
@ph4r05 I'm finished with my review, you can mark all my conversations as "resolved". As for the protobuf issue,
> I'm fine with it, as long as it's in contrib/brew/Brewfile so MacOS users can build it while following instructions from readme.md

and it seems to be fixed in the latest commits.

.github/workflows/build.yml
```diff
@@ -24,6 +24,10 @@ env:
         ccache --max-size=150M
         ccache --set-config=compression=true
 
+concurrency:
```
selsta, line 27: We usually do it this way: Make a private branch inside your fork and only push working commits to the (separate) PRed branch. Then you can always cancel builds that you don't need anymore here: https://github.com/ph4r05/monero/actions

cmake/CheckTrezor.cmake
```diff
@@ -102,13 +113,14 @@ if(Protobuf_FOUND AND USE_DEVICE_TREZOR AND TREZOR_PYTHON)
         "${CMAKE_BINARY_DIR}/test-protobuf.pb.cc"
         "${CMAKE_CURRENT_LIST_DIR}/test-protobuf.cpp"
         CMAKE_FLAGS
+        CMAKE_EXE_LINKER_FLAGS ${CMAKE_TRY_COMPILE_LINKER_FLAGS}
         "-DINCLUDE_DIRECTORIES=${Protobuf_INCLUDE_DIR};${CMAKE_BINARY_DIR}"
-        "-DCMAKE_CXX_STANDARD=11"
-        LINK_LIBRARIES ${Protobuf_LIBRARY}
+        "-DCMAKE_CXX_STANDARD=14"
```
selsta, line 118: What's even the point of the compile test? Seems it has been failing since forever, see here in latest master: https://github.com/monero-project/monero/actions/runs/5392471097/jobs/9790860150#step:5:145

If you remove this test we can use the latest version of protobuf again without issues.

ph4r05: nope, it is not failing neither for me locally nor for GHA runners. If I remove it and keep latest protobuf, it fails for me locally. Try upgrading protobuf to 23 locally and remove protobuf@21. maybe you have linked still v21?

.github/workflows/build.yml
ph4r05: sure, but is something wrong with the proposed change? why extra step of keeping 2 separate branches for one PR?

cmake/CheckTrezor.cmake
ph4r05: if you fetch this PR branch and try compiling brew-installed protobuf v23 with c++14 it fails regardless of try_compile result. Does GHA runner give another result? I tried it in this PR, it failed.

ph4r05: making it work with latest protobuf is effor worth of hours, IMO for another PR. I dont have time for this rn

ph4r05: also ad failing compile test - this PR fixes it, if compile test fails, trezor support is not included, thats one of primary reasons for this PR

selsta: I will try if I can fix it myself, I said this before but the problem is that this will break compilation and the brew file doesn't fix it, it will just cause conflicts with already installed version of protobuf -> support requests that I will have to deal with :/ I'm just unhappy with it currently but I can understand if you don't want to spend more time on it currently.

ph4r05: OK so in the next PR? :) 
Brewfile should relink protobuf@21 to be the main lib, so when installed, compilation should pass.

ph4r05: Also pls note that this PR makes Trezor support mandatory, if some check fails and Trezor support could not be enabled, build fails. This was not the case before. So if protobufv23 failed on try compile, Trezor support was disabled and compilation passed. We dont want this to happen anymore as official builds sometimes lacked Trezor support (it is still possible to explicitly disable trezor support via cmake option, but by default, it is enabled.)

ph4r05: another approach IMO: a) c++17, then lets use new protobuf, b) allow using protobufv21 in not linked version, but then user should override protobuf version that is going to be used - env vars probably

selsta: If I remember correctly we used C++14 because of using an old XCode version for macOS binaries. We updated the XCode version a year ago so it might be possible for us to switch to C++17. I'll look into it.

ph4r05: This also works

```
CMAKE_PREFIX_PATH=/opt/homebrew/Cellar/protobuf@21/21.12 make debug-test-trezor -j8
```

ph4r05:
@SChernykh thanks for the review! If it is ok by now, can you pls approve? Thanks! :)

SChernykh:
> @SChernykh thanks for the review! If it is ok by now, can you pls approve? Thanks! :)

Yes, but you should squash the commits first. Because if I approve now, I'll have to re-approve again later.

ph4r05:
ah, ok sure, I will squash it now. I thought we approve and then we squash&merge (should be only allowed stragety)

ph4r05:
squashed

SChernykh:
> I thought we approve and then we squash&merge (should be only allowed stragety)

I don't know about it, but for me it's simpler to compare the final single commit and the last one that I reviewed - if Github shows no differences, it's fine to approve.

SChernykh (approved):

ph4r05:
thanks for the review! Can you pls re-run tests? There seems to be nondeterministic test failure in `select_outputs.exact_unlock_block`

SChernykh:
I can't. Someone with write access to this repository needs to click "Re-run failed jobs" in https://github.com/monero-project/monero/actions/runs/5420602874/jobs/9855459005?pr=8752

selsta:
Opening / closing the pull requests re-runs CI.

ph4r05:
hmm this test is not in the code from this PR, is it post merge with master test?

```
/home/runner/work/monero/monero/tests/unit_tests/output_selection.cpp:259: Failure
Value of: picked_exact_unlock_block
  Actual: false
Expected: true
[  FAILED  ] select_outputs.exact_unlock_block (2840 ms)
```

cmake/CheckTrezor.cmake
```diff
@@ -102,13 +113,14 @@ if(Protobuf_FOUND AND USE_DEVICE_TREZOR AND TREZOR_PYTHON)
...
+        LINK_LIBRARIES ${Protobuf_LIBRARY} ${CMAKE_TRY_COMPILE_LINK_LIBRARIES}
         OUTPUT_VARIABLE OUTPUT
     )
     if(NOT Protobuf_COMPILE_TEST_PASSED)
-        message(STATUS "Protobuf Compilation test failed: ${OUTPUT}.")
+        message(FATAL_ERROR "Trezor: Protobuf Compilation test failed: ${OUTPUT}.")
```
tobtoht, line 123: This breaks development builds on Arch Linux, Brew, MSYS2, which all ship Abseil built with C++17.

https://github.com/abseil/abseil-cpp/issues/696#issuecomment-636949929

ph4r05: Do you pls have any CI link for this? All tests are passing here. Trezor support is mandatory now, unless it is explicitly turned off by compile switch. 

Also, when compile test fails, is the whole build passing with Trezor?

tobtoht: Rerun CI. MSYS2 updated to Protobuf 23.4 (from 21.23) 5 days ago.

ph4r05: Hmm, what do you propose? Can we pin protobuf to v 21? 
@selsta pls any update on c++17? Thanks!

selsta: I opened a PR for it but it's unclear if others agree with it: https://github.com/monero-project/monero/pull/8922

ph4r05: >     Disable Trezor support by default if Protobuf > 21.

This would result in official builds without Trezor support. Weird solution, rather pin protobuf to v21 or install from sources

ph4r05: AFAIK only win build is broken here. And according to the https://www.msys2.org/docs/package-management/ and https://github.com/msys2/setup-msys2 we can install older protobuf version explicitly. This handles CI case.
 (=official releases).

If user wants to do it on their own, they can a) disable Trezor support if they dont need it b) wait for c++17 support c) follow CI recipe and install compatible protobuf.

tobtoht: Official CLI builds are built with `depends` and it [doesn't use](https://github.com/monero-project/monero/blob/master/contrib/depends/packages/native_protobuf.mk#L2) Protobuf > 21. Monero GUI builds are pinned at an older version too.

FYI, the monero-gui Arch Linux [package](https://gitlab.archlinux.org/archlinux/packaging/packages/monero-gui/-/blob/main/PKGBUILD) (which is built against Protobuf 23.4) does not have Trezor support because of the broken test in CheckTrezor. So this should be fixed regardless.

tobtoht: MSYS2 does not support downgrades. It would be easier to configure CI to 
```
sed -e 's|CMAKE_CXX_STANDARD 14|CMAKE_CXX_STANDARD 17|' -i CMakeLists.txt
``` 
and fix the test.

tobtoht: |                             | Current state           | With this PR                                                         | 
|-----------------------------|-------------------------|----------------------------------------------------------------------|
| Official builds  (Pb <= 21) | ✅ Trezor support (not enforced) | ✅ Trezor support (enforced by default)                                             |
| Dev builds (Pb <= 21)       | ✅ Trezor support (not enforced) | ✅ Trezor support (enforced by default)                                               |
| Dev builds (Pb > 21) [0]       | ❌ Trezor support, [1]| ❌ **Builds fail**, [2]                   |
| Monero packages (Pb > 21)   | ❌ Trezor support, [1] | ❌ **Builds fail**, [1] |

[0] Currently: Arch Linux (and derivatives), MSYS2, and Homebrew (unless Pb is downgraded).

[1] To enable Trezor support user/maintainers need to set `CMAKE_CXX_STANDARD` to `17` and remove the test in `CheckTrezor.cmake`. (https://paste.debian.net/plainh/6361dd42)

[2] To disable Trezor support users need to manually pass `-DUSE_DEVICE_TREZOR=OFF`. 

A less drastic measure to require Trezor support for official **CLI** builds would be to check if we're building a release build with `if(DEPENDS)` and force it only then. Unfortunately, this doesn't work for the GUI as it doesn't use `depends` for all builds. Alternatively, you can check if `CMAKE_BUILD_TYPE` is `Release`, but this will catch some users building with `make release`, so this isn't ideal.

My point is that I would rather see graceful failure for Protobuf > 21, because it will otherwise result in an influx of support queries.

ph4r05: hmm @tobtoht I see, but what do you suggest? :) I could add a switch to disable compilation test. But is there a way to detect protobuf is > 21?

selsta: https://cmake.org/cmake/help/latest/module/FindProtobuf.html has a Protobuf_VERSION

SChernykh (approved):

ph4r05:
@selsta @tobtoht I've added a new switch `USE_DEVICE_TREZOR_PROTOBUF_TEST`, enabled by default. It checks if protobuf > 21, then it fails gracefully. Otherwise continues with the compilation test. When set to OFF, no Protobuf version check and compile test is done. Is this ok?

@SChernykh pls reapprove, this should be the last change :) Many thanks!

SChernykh (approved):

cmake/CheckTrezor.cmake
ph4r05: @selsta @tobtoht I've added a new switch USE_DEVICE_TREZOR_PROTOBUF_TEST, enabled by default. It checks if protobuf > 21, then it fails gracefully. Otherwise continues with the compilation test. When set to OFF, no Protobuf version check and compile test is done. Is this ok?

ph4r05:
Hello! I was wondering, how is it with a merge of this one? :)
e.g., @selsta

selsta:
@ph4r05 I was hoping the C++17 change was accepted first but that got kinda stuck :/

ph4r05:
@selsta ah, this can take quite some time. Can we pls merge this before? Over time, it can break again with other changes, new review will be needed, etc. Thanks for consideration!

ph4r05:
Yes pls, it is open for like 6 months. We can do other changes later

selsta:
I tried to explain it already but the problem is that merging this is going to break compilation on both macOS and Windows and you have to manually mess around with dependencies to fix it. With C++17 it would be possible to avoid this.

I'm the person who has to deal with all the resulting opened issues about compilation breaking so I don't want to merge this as is...

ph4r05:
I can think about it, but CI passes on all platforms -> if user downloads dependencies according to the readme, everuthing will work just fine. Just one install step to install older protobuf.

Newer protobuf (default now on brew/win) -> you need C++17.

But honestly, this is quite bizare now. This PR is open for half a year and we are still discussing this. What are users going to think about Trezor support for Monero? And now we are waiting another 6 months for C++17, which may or may not be merged? 

If build fails - either install correct deps or disable trezor support. Everything doable with one line. And people compiling it on our own - what is the percentage of those? Cannot they read proper docs when doing it already? I dont really see an issue here.

I would like to hear another opinions how to get this working. Otherwise it seems unsustainable to have decent Trezor support in monero.

ph4r05:
I can update readme in this PR if you approve it. Thats all I can do with this PR.

ph4r05:
Having compilation failed silently lead to official Monero releases without Trezor support. 

What do you propose? Make Trezor optional again? 

Waiting for C++17 is ridiculous

SChernykh:
> I tried to explain it already but the problem is that merging this is going to break compilation on both macOS and Windows

@selsta I see "All checks have passed" in this PR, including macOS and Windows checks. What compilation are you talking about? If it's some script somewhere else in the repo, then it should be just fixed and then this PR can be merged.

ph4r05:
I will go over this over the weekend, will see what can be done about it

ph4r05:
@selsta @SChernykh I've made Trezor support optional, controlled by `USE_DEVICE_TREZOR_MANDATORY` cmake setting and env var (default to OFF, but ON on CI).

If protobuf v21 is not installed, the following warning is shown and the trezor support is disabled, compilation continues:
```
CMake Warning at cmake/CheckTrezor.cmake:78 (message):
  [WARNING] Trezor support cannot be compiled!

  Skipping Trezor support.  Unsupported Protobuf version 24.3.0.  Please, use
  Protobuf v21.For more information, please check src/device_trezor/README.md
```

If you set `USE_DEVICE_TREZOR_MANDATORY=ON` and there is any fail, the following error is shown

```
CMake Error at cmake/CheckTrezor.cmake:53 (message):
  Trezor: Unsupported Protobuf version 24.3.0.  Please, use Protobuf v21.


  ==========================================================================

  To compile without Trezor support, set USE_DEVICE_TREZOR=OFF.  It is
  possible both via cmake variable and environment variable, e.g.,
  `USE_DEVICE_TREZOR=OFF make release`

  For more information, please check src/device_trezor/README.md
```

I may missed some spot where to set env var `USE_DEVICE_TREZOR_MANDATORY=ON` for CI, if you can, pls check it.

ph4r05:
@SChernykh pls when you have a minute to review new changes, thanks! :)

src/device_trezor/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+A custom high-level transaction signing protocol uses Trezor in a similar way a cold wallet is used. 
+Transaction is build incrementally on the device. 
+
+Trezor implements the signing protocol in [trezor-firmware] repository, in the [monero](https://github.com/trezor/trezor-firmware/tree/master/core/src/apps/monero) application.
+Please, refer to [monero readme](https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/monero/README.md) for more information.
+
+## Dependencies
+
+Trezor uses [Protobuf](https://protobuf.dev/) library. As Monero is compiled with C++14, the newest Protobuf library is cannot be compiled.
```
SChernykh, line 18: `is cannot be compiled` a typo? You can also add "because it requires C++17".

SChernykh: Nevermind, you mention C++17 further down.

SChernykh (approved):

SChernykh:
Looks good. I only found one typo in readme, but it doesn't really matter.

ph4r05:
@SChernykh thanks for the review! I fixed the typo, if it is ok now, pls approve 🙏  Thanks!

SChernykh (approved):

ph4r05:
@selsta I've made Trezor support optional, controlled by `USE_DEVICE_TREZOR_MANDATORY` cmake setting and env var (default to OFF, but ON on CI).

If protobuf v21 is not installed, the following warning is shown and the trezor support is disabled, compilation continues:
```
CMake Warning at cmake/CheckTrezor.cmake:78 (message):
  [WARNING] Trezor support cannot be compiled!

  Skipping Trezor support.  Unsupported Protobuf version 24.3.0.  Please, use
  Protobuf v21. For more information, please check src/device_trezor/README.md
```

If you set `USE_DEVICE_TREZOR_MANDATORY=ON` and there is any fail, the following error is shown

```
CMake Error at cmake/CheckTrezor.cmake:53 (message):
  Trezor: Unsupported Protobuf version 24.3.0.  Please, use Protobuf v21.


  ==========================================================================

  To compile without Trezor support, set USE_DEVICE_TREZOR=OFF.  It is
  possible both via cmake variable and environment variable, e.g.,
  `USE_DEVICE_TREZOR=OFF make release`

  For more information, please check src/device_trezor/README.md
```

I may missed some spot where to set env var `USE_DEVICE_TREZOR_MANDATORY=ON` for CI, if you can, pls check it.

Also, I've added few lines of documentation in the last commit.

So, after implementing what you suggested, is this a megreable state? Could we pls merge soon? Thanks! :)

selsta:
Looks good now, could you squash your commits? Also I would prefer if the CI concurrency changes are in a separate PR, since they are totally unrelated to the Trezor changes.

ph4r05:
ok @selsta, I've removed concurrency, will add it as separate PR. 
btw do you need me to squash the commits by force push? Can it be merged with "Squash and merge"? 

Screenshot 2023-09-29 at 15 26 36

selsta:
Yes, please squash and then force-push.

Regarding the "Squash and Merge" feature, as far as I know it means the commit signatures of the author get removed, and the maintainer would sign it with their own key. This isn't something we want.

ph4r05:
@selsta done

selsta:
Sorry for having another change request, the Trezor related changes seem good to me, but one suggestion from moneromooo is that the Trezor unrelated chaingen / core test bug fixes should be a in a separate commit so that it can be bisected and reverted if issues are found. This is really critical code so we have to be careful here with not mixing up too many changes in a large commit.

```
16:30 <moneromoooo> I try to make one commit per fix when I can. It's *usually* not hard to commit a fix separately when you code a larger thing and you uncover some bug while doing so. Not *always* easy, but usually. And since this patch isn't really needed since it's just for trezor, it applies even more here.
16:31 <moneromoooo> Quite often I'll have a string of random fixes or small improvements, then the main patch that adds a feature. It makes it easier to review, bisect, revert piecemeal if needed.
```

ph4r05:
So the same PR but different commit? 2 commits then? I doubt it makes sense to separate chaingen to more commits.

Also, do we all understand that by sepparating trezor changes and chaingen causes trezor commit alone to fail tests? Bisect would have to be done on the prticular test.

selsta:
If the chaingen bug is unrelated to the Trezor changes then it would ideally be first commit fixes chaingen bug, second commit the Trezor changes. If the bugs are intertwined then I guess it isn't possible.

ph4r05:
I get the point @moneromooo-monero. But it is not "just for Trezor", the bug is present there in general and Trezor tests are testing scenarios jot covered by unit tests, thus hiting edge cases when it has to be matching precisely on block boundaries. 

I will get to this in following days. Trezor commit (failing tests on its own), then chaingen fixes.

ph4r05:
Will see what can I do @selsta

selsta:
Thank you and sorry this is dragging out, just with core test changes we are usually extra careful.

ph4r05:
I get it, no worries. Glad we will merge it soon

ph4r05:
I am splitting the commits, I will commit the first one, wait for tests, then if it is all green another one (Trezor). 
btw regarding approval - we dont need to code review it again, right? 😅

jeffro256:
Sorry to pester you more, but for reference, do you have a minimal reproducible example code that triggers the core_tests to break without the changes in this commit?

ph4r05:
I am sorry but I am not sure I have time and/or energy for this

ph4r05:
@jeffro256 take for example this change, https://github.com/monero-project/monero/pull/8752/commits/056c9967036175e78b7ac1758d9a2943a9283bfc#r1238127696

Original code rewound the chain after some operations, i.e., added `CRYPTONOTE_MINED_MONEY_UNLOCK_WINDOW` blocks and moved it from `blk_5` to `blk_5r`.

Then the buggy code tried to build a transaction on top of `blk_5` (old block before rewinding), this does not make any sense, fund selection was broken, rewinding is precisely the reason so the transaction construction can use the funds according to the chain rules (unlock window). Creating transaction on top of `blk_5` uses funds that are not yet ready. 

When I then pass this code to the full fledged daemon validator (thats what Trezor tests do), it fails, as it should on the real network. Also, code already rewound the chain so I believe this is a systematic error. Author IMO intended to construct a transaction after unlock has passed. IMO It makes sense to have tests that correspond to the real chain rules. Otherwise it creates confusion, passing test chain would fail real daemon validation. And the rewinding was already there. Its not I am rewriting the whole tests so they pass the real daemon validation. I am just using correct block to build TX on. Which was I believe original intention of the author.

@selsta I split the commits. Do you think we can proceed with the merging? :)

ph4r05:
@SChernykh if you can, reapprove? 😅 If you want to diff it, this is branch you approved lately https://github.com/ph4r05/monero/tree/pr/trezor-tests-orig

this one is rebased on master, split to 2 commits. Thanks! :)

SChernykh:
Diff is too big now (700+ files) because of the rebase. I'll try to figure out how to get the minimal diff for review later today.

ph4r05:
if you rebase original branch on your own on the master, then do the diff?

SChernykh:
I did back-to-back comparison of https://github.com/monero-project/monero/compare/master...ph4r05:monero:pr/trezor-tests-orig and https://github.com/monero-project/monero/compare/master...ph4r05:monero:pr/trezor-tests and they're identical.

selsta:
@ph4r05 thank you, I think from your side everything is good now, I'm still trying to understand the chaingen changes and afterwards it can be approved

SChernykh (approved):

ph4r05:
Thanks @selsta!
I think tldr for chaingen is: create tx using funds that will be unlocked when tx is added to the chain so full validation passes on the chain.

selsta:
@ph4r05 jeffro256 is still reviewing the chain-gen changes, but it seems so far everything is good and he will approve it in the next days. Then we can add it to the merge queue.

jeffro256 (approved):
I didn't go in-depth for the trezor-specific tests, but I skimmed them and can confirm that at least they look reasonable and don't affect other parts of the code. As for the core tests, the changes made are correct and fix existing bugs, or add functionality for the trezor tests in a backwards-compatible way.

ph4r05:
Thank you all for your time!

PR #8754, opened 2023-03-01
Make it possible to exit cli wallet when locked

dan1338:
### This is a significant UX hindrance and should have been fixed long ago.

I attempted to find the root cause for segfault that happens if we just `w.stop()` in the handler but it appears to happen in the libstdc++ at_exit handlers resulted from corrupted state and further investigation with valgrind shows multiple data races which I don't see a way of solving.

It might not be an "ideal" solution but it's definitely due.

From what I gathered the architecture around simplewallet cli is poor and mostly patchworked with threading primitives.
The terminal is being read asynchronously from many places, such as "contrib/epee/include/console_handler.h" and "src/common/password.cpp", maybe if it was designed around a main synchronous thread reading stdin it would've been easier to fix but without a rewrite this will have to suffice.

I believe this solution to be the easiest, most painless way to exit the wallet after it inevitably locks itself.
Retires #7722 as well

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -10702,7 +10702,9 @@ int main(int argc, char* argv[])
     tools::signal_handler::install([&w](int type) {
       if (tools::password_container::is_prompting.load())
       {
-        // must be prompting for password so return and let the signal stop prompt
+        w.stop();
+        w.deinit();
+        _exit(0);
         return;
       }
 #ifdef WIN32
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -10702,6 +10702,8 @@ int main(int argc, char* argv[])
     tools::signal_handler::install([&w](int type) {
       if (tools::password_container::is_prompting.load())
       {
+        // Runs the full exit routine after receiving Ctrl-C
+        // In current state we're incapable of gracefully exiting from here without major redesign, hence the blunt approach
         w.stop();
         w.deinit();
         _exit(0);


src/simplewallet/simplewallet.cpp
```diff
@@ -10702,7 +10702,9 @@ int main(int argc, char* argv[])
     tools::signal_handler::install([&w](int type) {
       if (tools::password_container::is_prompting.load())
       {
-        // must be prompting for password so return and let the signal stop prompt
+        w.stop();
```
iamamyth, line 10705: Can you add a comment explaining the rationale for calling `deinit` and `_exit` here in addition to stop (in particular that it wasn't a UX design choice so much as a choice of necessity)? It doesn't have to be nearly as detailed as the PR description, but would probably help anyone trying to refactor in the input handling to allow a more graceful exit.

iamamyth:
I guess I should note that I see an alternative here: The password reading function in common/password.cpp returns a success/failure bool, which get_and_verify_password hides by coalescing to fail/success. You could add an optional bool ptr out argument to get_and_verify_password and, if it's not null, pass along the read status. Then the unlock checker could run the shutdown code if password reading fails, which should only happen on interrupt.

dan1338:
@iamamyth 
I considered something to that extent. But that would require changing all instances of it's usage and changing it's external API. And frankly it would get way more complex trying to implement it, linux can easily mask SIGINT in tcsetattr,. Can you do the same on windows? In essence it would be another patchwork solution, adding complexity for the sake of fixing bugs arisen from lack of architecture. At least my solution is direct and doesn't introduce unnecessary logic, it simply does what should be done.

iamamyth (approved):
I think a few specifics of my proposal may have been poorly communicated, but it's moot: This PR at least would net an end-user improvement and doesn't introduce further faulty assumptions.

plowsof:
confirmed behaviour noted in #7722 (ctrl+c endless password prompt) 

with this PR - ctrl+c kill the wallet (confirmed not appearing in process list after) - 1 tiny thing to be aware of- after exiting with ctrl+c on Ubuntu , i cant see text im typing into the terminal anymore, have to close it,

iamamyth:
@plowsof Yes in theory the approach I envision would fix that, but requires additional effort and could run into some snags. You should be able to type `reset` to fix the terminal (though of course you won't be able to see the chars), or a hotkey equivalent (depends on your OS/terminal setup).

selsta:
On macOS there are no issues with the Terminal after exiting with ctrl+c.

> You should be able to type reset to fix the terminal (though of course you won't be able to see the chars), or a hotkey equivalent (depends on your OS/terminal setup).

I feel like we would get issue reports about that :/

iamamyth:
> On macOS there are no issues with the Terminal after exiting with ctrl+c.
Observed behavior will depend heavily on the OS and terminal program.

> I feel like we would get issue reports about that :/
Maybe, but there are already reports regarding the current behavior. Seems like a two steps forward one step backward situation.

dan1338:
I know about the terminal but wasn't sure what's the best solutions, when testing I just typed `reset`. It happens because the terminal was left in raw mode by https://github.com/monero-project/monero/blob/c5d10a4ac43941fe7f234d487f6dd54996a9aa33/src/common/password.cpp#L127-L131
AFAIK contrib/epee uses readline so it would've restored the terminal mode to normal were we to exit() hitting the at_exit handlers.

But that is precisely what were trying to avoid, hence the _exit call.
Doing a tcsetattr before exiting would be in the spirit of pr but please feel free to propose alternative solutions

iamamyth:
> Doing a tcsetattr before exiting would be in the spirit of PR but please feel free to contribute alternative solutions

My preference would be to avoid this addition, though I would still approve the change: The signal handler code already violates the re-entrancy requirement (prior to this PR), so the current changeset does not alter the prior assumptions. However, tcsetattr certainly isn't re-entrant and, unlike the other functions currently invoked by the handler, may have effects even after the program exits. So, the very reason you want the change (terminal may be broken) is, in my mind, a reason to avoid it: What happens to the terminal if you re-enter tcsetattr (e.g. the user hits ctrl+c multiple times in short succession, a fairly common occurrence)? It could very well be in a worse state where not even `reset` works properly, or maybe it acquires a lock and now the program deadlocks, etc.

dan1338:
This gave me an idea.
How about we call `execve("/bin/reset")`? It exits without calling atexit handlers and resets the terminal to a working state.

I wouldn't mind that at all but the obvious problem is compatibility. Personally I'm only concerned with linux but AFAICS there is windows support in place so we need to support both?

It might just be that this isn't an issue on windows so we could simply #ifdef the `execve` to linux only, and `_exit` on windows. So far I'm liking this solution best.

iamamyth:
It looks like `tcsetattr` is actually signal safe, and so is `execve`. I tend to agree that `execve` produces a better result on linux, and tends to future-proof any password prompt attribute changes, but I wouldn't necessarily assume a program exists on the user's system.

selsta:
Superseded by https://github.com/monero-project/monero/pull/11136

Issue #8756, opened 2023-03-03
discussion: remove output blackballing


tobtoht:
#### Context

From https://monero-blackball.github.io/monero-blackball-site/

> This tool can be used to avoid including outputs that are known to be spent in other transactions. If you exclude these known-spent outputs, other higher-quality outputs can be included in your ring signatures, thus improving your privacy.

Once imported, blackballed outputs are stored in the shared ringdb.

#### Why remove blackballing?

Output blackballing has been "unnecessary" for over 4 years: https://old.reddit.com/r/Monero/comments/9rml9j/generating_and_importing_an_output_spent/

> This is a quick PSA to say that generating and importing a spent output (blackball) list is unnecessary with Monero's new, larger ringsize. It was also unnecessary with Monero's ringsize 7.

Removing output blackballing would eliminate ~2.2k lines of code (https://github.com/tobtoht/monero/commit/906f9c840d7f968a91dd296133962c225ee02a86), remove `monero-blockhain-mark-spent-outputs` (9.6 MB) from release archives, and slightly simplify decoy selection code. It is also one less thing to worry about during the seraphis transition.

SamsungGalaxyPlayer:
This was a really interesting project at the time, but I have to agree. There's no expected future for this, and no further support is needed.

Gingeropolous:
this might be useful if the chain gets filled with nfts

Gingeropolous:
so should this issue be closed?

selsta:
It has been removed from the GUI but not yet from the CLI. I would keep it open until there is consensus on if we should remove it or keep it.

Rucknium:
Vijayakumaran (2023) notes bugs when using the blackball tool with forked blockchains:

> The Monero reference implementation includes a tool for identifying spent outputs using the techniques described above [6]. It implements the cascade attack, finds transactions which cause the ring attack characterized by Wijaya et al. [22], attempts to identify closed sets, and performs the cross-chain analysis proposed by Wijaya et al. [24] and Hinteregger et al. [13]. It is included in every Monero release as an executable with the name monero- blockchain-mark-spent-outputs. It is informally called the “blackball tool” in the Monero community as the set of spent outputs represent a blacklist which should be avoided when sampling mixins. To perform cross-chain analysis, the tool takes the LMDB database files of all chains as input.

[...]

PR #8758, opened 2023-03-03
wallet: remove output blackballing

tobtoht:
#8756


selsta (approved):

UkoeHB (approved):
I did not check if everything related to blackballing was removed, if anything was missed it can be cleaned up in a future PR.

jeffro256 (commented):
I'm not approving the changes in principle *yet*, but this PR does successfully remove only blackball features and should not affect the behavior of anything else.

tobtoht:
Rebased.

jeffro256:
I believe we should keep the output blackballing feature as long as there still exists the possibility of poisoned outputs (e.g. Mordinals). If/when full-chain membership proofs get merged into mainnet, then it would be safe to remove blackballing entirely.

tobtoht:
I disagree. The current blackballing implementation is not helpful at mitigating effective ringsize reduction from large amounts of non-uniform outputs (such as Mordinals - which was [mitigated](https://github.com/monero-project/monero/pull/8733)), as it would require manually generating and importing up-to-date blacklists before sending a transaction. (About half of selected decoys are < 1 day old, so at least daily). Transactions that do not include these outputs in their rings are finger-printable on chain, reducing transaction uniformity and increasing traceability for the small subset of users that would go through the effort. 

A better solution for output blackballing would be to implement it on the daemon side and have it mark undesirable outputs in the [get_outs](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#get_outsbin) call. This would not require storing blackballed outputs in RingDB. In this way, all users would benefit and transaction uniformity is not harmed.

iamamyth (approved):

jeffro256 (approved):

PR #8760, opened 2023-03-03
wallet: remove defunct light wallet code

tobtoht:
This code is unmaintained and has been broken since forever. With plans to to get rid of `wallet2` with seraphis there is no point in trying to fix it now. Removing this code should disincentivize further development effort in this area (e.g. #8607).


selsta:
ping also @Malinero since you worked a bit on the lightwallet code

jeffro256:
I agree, this code should be scrapped since it only adds complexity while being useless in its current state. The new wallet code should ideally be less monolithic and so light wallet code won't have to shoved in, but can be built separately on top of wallet components.

jeffro256 (approved):

selsta:
@tobtoht merge conflict

jeffro256 (approved):

serhack (approved):

PR #8769, opened 2023-03-09
Allow option 'non-interactive' in monerod config file [release-v0.18]

### LINKED ISSUE #8753: monerod: unrecognised option 'non-interactive'
Hi there! When I try to add `non-interactive=1` option to the monerod.conf file and start with `--config-file monerod.conf`, monerod crashes with the following error:
```
Exception in main! unrecognised option 'non-interactive'
```

However, when I run monerod with the `--non-interactive` command line flag directly, it starts successfully.

almalh:
On the basis that it's cross-platform and not really a hidden option anymore; with some care not to impact wallet-rpc.

Closes #8753


jtgrassie:
First, the pull request should be against master, not on release until tested/approved.

Second, this is a process flag and in my opinion has no merit in the config file.

Lastly, seems a lot of code change just to move an option from one set to another.

PR #8774, opened 2023-03-14
Prevent tx broadcasts from using Tor circuit twice

53a45b2637452980:
https://github.com/monero-project/monero/blob/master/docs/ANONYMITY_NETWORKS.md
Addresses "Tor Stream Used Twice" by preventing same connection from broadcasting more than one local tx.
Drops onion connections with random delay and makes new connections with random delay. Parameters/strategy for dropping/reconnecting are arbitrary. Need review. 
I could not find a timestamp in peer timed sync, has it been removed?
Need to implement socks5 to improve solution.
Exclusive peer handling not final.
Small changes and it will work for I2P, I left it because I don't know how I2P works.
No tests written, good results in debug output "on my system".
This is a "rough draft"/"conversation starter". I have not done much C++.

### DIFF
--- a/src/cryptonote_protocol/levin_notify.cpp
+++ b/src/cryptonote_protocol/levin_notify.cpp
@@ -98,6 +98,13 @@ namespace levin
       return std::chrono::steady_clock::duration{crypto::rand_range(rep(0), range.count())};
     }
 
+    bool set_connection_can_broadcast_txs_to_peer(std::shared_ptr<connections> p2p, boost::uuids::uuid conn, bool val) {
+        return p2p->for_connection(conn, [&](detail::p2p_context& context){
+           context.can_broadcast_txs_to_peer = val;
+           return true;
+         });
+      }
+
     uint64_t get_median_remote_height(connections& p2p)
     {
         std::vector<uint64_t> remote_heights;
@@ -643,12 +650,20 @@ namespace levin
         const auto start = std::chrono::steady_clock::now();
         noise_channel& channel = zone_->channels.at(channel_);
 
+        bool can_broadcast = true;
+        zone_->p2p->for_connection(channel.connection, [&](detail::p2p_context& context){
+            can_broadcast = context.can_broadcast_txs_to_peer;
+            return true;
+        });
+
         if (!channel.connection.is_nil())
         {
           epee::byte_slice message = nullptr;
-          if (!channel.active.empty())
+          if (can_broadcast && !channel.active.empty())
+          {
             message = channel.active.take_slice(zone_->noise.size());
-          else if (!channel.queue.empty())
+          }
+          else if (can_broadcast && !channel.queue.empty())
           {
             channel.active = channel.queue.front().clone();
             message = channel.active.take_slice(zone_->noise.size());
@@ -658,8 +673,12 @@ namespace levin
 
           if (zone_->p2p->send(std::move(message), channel.connection))
           {
-            if (!channel.queue.empty() && channel.active.empty())
+            if (can_broadcast && !channel.queue.empty() && channel.active.empty())
+            {
               channel.queue.pop_front();
+              if (zone_->nzone == epee::net_utils::zone::tor)
+                set_connection_can_broadcast_txs_to_peer(zone_->p2p, channel.connection, false);
+            }
           }
           else
           {
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -1350,6 +1350,8 @@ namespace nodetool
   template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::try_to_connect_and_handshake_with_new_peer(const epee::net_utils::network_address& na, bool just_take_peerlist, uint64_t last_seen_stamp, PeerType peer_type, uint64_t first_seen_stamp)
   {
+    bool one_conn_below_max = false;
+
     network_zone& zone = m_network_zones.at(na.get_zone());
     if (zone.m_connect == nullptr) // outgoing connections in zone not possible
       return false;
@@ -1361,6 +1363,10 @@ namespace nodetool
     {
       return false;
     }
+    else if (zone.m_current_number_of_out_peers == zone.m_config.m_net_config.max_out_connection_count - 1)
+    {
+      one_conn_below_max = true;
+    }
     else if (zone.m_current_number_of_out_peers > zone.m_config.m_net_config.max_out_connection_count)
     {
       zone.m_net_server.get_config_object().del_out_connections(1);
@@ -1368,6 +1374,9 @@ namespace nodetool
       return false;
     }
 
+    if (na.get_zone() == epee::net_utils::zone::tor && time(nullptr) < m_make_next_tor_connection_after) {
+      return false;
+    }
 
     MDEBUG("Connecting to " << na.str() << "(peer_type=" << peer_type << ", last_seen: "
         << (last_seen_stamp ? epee::misc_utils::get_time_interval_string(time(NULL) - last_seen_stamp):"never")
@@ -1424,6 +1433,17 @@ namespace nodetool
     zone.m_peerlist.append_with_peer_anchor(ape);
     zone.m_notifier.on_handshake_complete(con->m_connection_id, con->m_is_income);
     zone.m_notifier.new_out_connection();
+
+    if (na.get_zone() == epee::net_utils::zone::tor) {
+      if (one_conn_below_max)
+      {
+        extend_make_next_connection_after(na.get_zone(), false);
+      }
+      else
+      {
+        extend_make_next_connection_after(na.get_zone(), true);
+      }
+    }
 
     LOG_DEBUG_CC(*con, "CONNECTION HANDSHAKED OK.");
     return true;
@@ -1788,6 +1808,35 @@ namespace nodetool
   }
   //-----------------------------------------------------------------------------------
   template<class t_payload_net_handler>
+  void node_server<t_payload_net_handler>::extend_make_next_connection_after(epee::net_utils::zone zone,
+                                                                             bool acquire_peers) {
+    if (zone == epee::net_utils::zone::public_) {
+      return;
+    }
+
+    auto scaling_factor = 1;
+    if (acquire_peers)
+    {
+      scaling_factor = ::config::ACQUIRE_PEERS_RECONNECT_SCALING_FACTOR;
+    }
+    auto offset = crypto::rand_range<time_t>(::config::RANDOM_CONNECTION_DROP_LOWER / scaling_factor,
+                                             ::config::RANDOM_CONNECTION_DROP_UPPER / scaling_factor) / 1000000;
+
+    if (zone == epee::net_utils::zone::tor)
+    {
+      if (m_make_next_tor_connection_after > time(nullptr))
+      {
+        return;
+      }
+      m_make_next_tor_connection_after = time(nullptr) + offset;
+    }
+    else if (zone == epee::net_utils::zone::i2p)
+    {
+      ; // not implemented
+    }
+  }
+  //-----------------------------------------------------------------------------------
+  template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::connections_maker()
   {
     using zone_type = epee::net_utils::zone;
@@ -1800,6 +1849,10 @@ namespace nodetool
     bool one_succeeded = false;
     for(auto& zone : m_network_zones)
     {
+      if (zone.first == zone_type::tor && m_make_next_tor_connection_after > time(nullptr)) {
+        continue;
+      }
+
       size_t start_conn_count = get_outgoing_connections_count(zone.second);
       if(!zone.second.m_peerlist.get_white_peers_count() && !connect_to_seed(zone.first))
       {
@@ -1811,7 +1864,6 @@ namespace nodetool
       size_t base_expected_white_connections = (zone.second.m_config.m_net_config.max_out_connection_count*P2P_DEFAULT_WHITELIST_CONNECTIONS_PERCENT)/100;
 
       // carefully avoid `continue` in nested loop
-      
       size_t conn_count = get_outgoing_connections_count(zone.second);
       while(conn_count < zone.second.m_config.m_net_config.max_out_connection_count)
       {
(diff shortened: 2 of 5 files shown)


selsta:
@vtnerd

jeffro256:
Thanks for the PR! Do you have a recommended way to test that this solution works? Have you tested it yourself yet?

53a45b2637452980:
For testing onion peer drop/reconnect I recommend adding debug output in function drop_random_outgoing_tor_connection that lists the outgoing peers in the tor zone and when they are dropped. You can add debug output for the value of p2p_context member can_broadcast_txs_to_peer which will change after a tx has been sent. I did a few hours of testing and the results were good. There are no onions on stagenet, it is better to test onion connections on mainnet.

53a45b2637452980:
I will update the documentation once the code has been reviewed as I expect that changes will be required. As I said, this PR is a "conversation starter", there's lots to discuss.

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -643,12 +654,20 @@ namespace levin
...
         if (!channel.connection.is_nil())
         {
           epee::byte_slice message = nullptr;
-          if (!channel.active.empty())
+          if (can_broadcast && !channel.active.empty())
+          {
             message = channel.active.take_slice(zone_->noise.size());
-          else if (!channel.queue.empty())
+          }
+          else if (can_broadcast && !channel.queue.empty())
```
vtnerd, line 670: This check should be reverted - a message already started on the channel, so it should continue.

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -98,6 +98,16 @@ namespace levin
       return std::chrono::steady_clock::duration{crypto::rand_range(rep(0), range.count())};
     }
 
+    bool set_connection_can_broadcast_txs_to_peer(std::shared_ptr<connections> p2p, boost::uuids::uuid conn, bool val, bool exclusive_peers) {
+      if (exclusive_peers && !val)
```
vtnerd, line 102: Why even have `val`? It looks to always be `false` ? Basically this function is designed to disable sending, and never re-enable it.

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -643,12 +654,20 @@ namespace levin
...
+        zone_->p2p->for_connection(channel.connection, [&](detail::p2p_context& context){
+            can_broadcast = context.can_broadcast_txs_to_peer;
+            return true;
+        });
+
         if (!channel.connection.is_nil())
         {
           epee::byte_slice message = nullptr;
-          if (!channel.active.empty())
+          if (can_broadcast && !channel.active.empty())
```
vtnerd, line 666: This check is insufficient for what you want - you need to actually remove the connection from the "channel". More specifically, the `update_channels` function needs to be filtering when it selects which connection will linked an active "channel.

src/p2p/net_node.inl
```diff
@@ -1350,6 +1350,8 @@ namespace nodetool
   template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::try_to_connect_and_handshake_with_new_peer(const epee::net_utils::network_address& na, bool just_take_peerlist, uint64_t last_seen_stamp, PeerType peer_type, uint64_t first_seen_stamp)
   {
+    bool one_conn_below_max = false;
```
vtnerd, line 1353: Why are you messing with the logic in this function?

src/p2p/net_node.inl
```diff
@@ -1361,13 +1363,20 @@ namespace nodetool
     {
       return false;
     }
+    else if (zone.m_current_number_of_out_peers == zone.m_config.m_net_config.max_out_connection_count - 1)
+    {
+      one_conn_below_max = true;
```
vtnerd, line 1368: Shouldn't this be a `<` check (without the `-1`)? Seems like if the connections dropped below the threshold, you want to run the additional logic. I'm just not sure why "one connection below max" is a special case given what you are trying to do.

I know it changes the scaling of when re-connects occur, but why is one less than max a special case (then further discussion about how its implemented)?

src/p2p/net_node.inl
```diff
@@ -1361,13 +1363,20 @@ namespace nodetool
...
+    }
     else if (zone.m_current_number_of_out_peers > zone.m_config.m_net_config.max_out_connection_count)
     {
       zone.m_net_server.get_config_object().del_out_connections(1);
       --(zone.m_current_number_of_out_peers); // atomic variable, update time = 1s
       return false;
     }
 
+    if (na.get_zone() == epee::net_utils::zone::tor && time(nullptr) < m_make_next_tor_connection_after) {
```
vtnerd, line 1377: This should be in the new `else if` ?

src/p2p/net_node.inl
```diff
@@ -1788,6 +1808,35 @@ namespace nodetool
...
   //-----------------------------------------------------------------------------------
   template<class t_payload_net_handler>
+  bool node_server<t_payload_net_handler>::extend_make_next_connection_after(epee::net_utils::zone zone,
+                                                                             bool acquire_peers) {
+    if (zone == epee::net_utils::zone::public_ || !m_exclusive_peers.empty())
+      return false;
+
+    auto scaling_factor = 1;
+    if (acquire_peers)
+      scaling_factor = ::config::ACQUIRE_PEERS_RECONNECT_SCALING_FACTOR;
```
vtnerd, line 1818: I would recommend using [`std::chrono::system_clock`](https://en.cppreference.com/w/cpp/chrono/system_clock) with `std::chrono::milliseconds`, etc. throughout this section.

src/p2p/net_node.inl
```diff
@@ -1425,6 +1434,17 @@ namespace nodetool
     zone.m_notifier.on_handshake_complete(con->m_connection_id, con->m_is_income);
     zone.m_notifier.new_out_connection();
 
+    if (na.get_zone() == epee::net_utils::zone::tor) {
```
vtnerd, line 1437: Trivial, but bracing doesn't follow the style in this file or this patch.

src/p2p/net_node.inl
```diff
@@ -1425,6 +1434,17 @@ namespace nodetool
...
+      if (one_conn_below_max)
+      {
+        extend_make_next_connection_after(na.get_zone(), false);
```
vtnerd, line 1440: Also trivial, but you don't need the `if` branching here. Just `extend_make_next_connection_after(na.get_zone(), !one_con_below_max)`.

vtnerd (changes requested):

vtnerd:
One other comment - I think you should be able to unit test some of the functionality in `levin_notify.cpp` ... ?

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -98,6 +98,16 @@ namespace levin
       return std::chrono::steady_clock::duration{crypto::rand_range(rep(0), range.count())};
     }
 
+    bool set_connection_can_broadcast_txs_to_peer(std::shared_ptr<connections> p2p, boost::uuids::uuid conn, bool val, bool exclusive_peers) {
```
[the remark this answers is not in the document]
53a45b2637452980: Will change.

src/p2p/net_node.inl
```diff
@@ -2001,6 +2066,43 @@ namespace nodetool
...
+            }
+            return true;
+        });
+    }
+
+    for (auto &cntx : outgoing_tor_connections) {
+      if (!cntx.can_broadcast_txs_to_peer) {
+        drop_connection(cntx);
+        // should use random auth for each connection but no socks5
+        // possible to use same circuit twice if drop initiated by remote
```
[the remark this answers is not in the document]
53a45b2637452980: Cheap/easy way to prevent connecting back until socks5 (mentioned in the comment).

src/p2p/net_node.inl
```diff
@@ -2001,6 +2066,43 @@ namespace nodetool
...
+        block_host(cntx.m_remote_address, (::config::DEFAULT_ONION_CIRCUIT_EXPIRY + 1) * 60, true);
+        return true;
+      }
+    }
+
+    if (auto len = outgoing_tor_connections.size()) {
+        auto index_to_drop = crypto::rand_idx(len);
+        drop_connection(outgoing_tor_connections[index_to_drop]);
+        block_host(outgoing_tor_connections[index_to_drop].m_remote_address,
```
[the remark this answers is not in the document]
53a45b2637452980: Cheap/easy way to prevent connecting back until socks5 (mentioned in the comment).

src/p2p/net_node.inl
```diff
@@ -2001,6 +2066,43 @@ namespace nodetool
...
+        drop_connection(cntx);
+        // should use random auth for each connection but no socks5
+        // possible to use same circuit twice if drop initiated by remote
+        block_host(cntx.m_remote_address, (::config::DEFAULT_ONION_CIRCUIT_EXPIRY + 1) * 60, true);
+        return true;
+      }
+    }
+
+    if (auto len = outgoing_tor_connections.size()) {
+        auto index_to_drop = crypto::rand_idx(len);
```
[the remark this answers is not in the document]
53a45b2637452980: The above loop has a return?

src/p2p/net_node.inl
```diff
@@ -2001,6 +2066,43 @@ namespace nodetool
...
+        tor_zone->second.m_net_server.get_config_object().foreach_connection([&](p2p_connection_context &cntx) {
+            if (!cntx.m_is_income) {
+                outgoing_tor_connections.push_back(cntx);
+            }
+            return true;
+        });
+    }
+
+    for (auto &cntx : outgoing_tor_connections) {
+      if (!cntx.can_broadcast_txs_to_peer) {
```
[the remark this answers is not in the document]
53a45b2637452980: Depending on how drop/reconnect parameters are set there might be lots of connections that can't broadcast and don't always want to drop the first. I forgot to add this extra logic in.

src/p2p/net_node.inl
```diff
@@ -2001,6 +2066,43 @@ namespace nodetool
...
     m_dns_blocklist_interval.do_call(boost::bind(&node_server<t_payload_net_handler>::update_dns_blocklist, this));
+    if (m_exclusive_peers.empty())
+      m_drop_random_outgoing_tor_connection_interval.do_call(boost::bind(&node_server<t_payload_net_handler>::drop_random_outgoing_tor_connection, this));
+    return true;
+  }
+  //-----------------------------------------------------------------------------------
+  template<class t_payload_net_handler>
+  bool node_server<t_payload_net_handler>::drop_random_outgoing_tor_connection()
+  {
+    std::vector<p2p_connection_context> outgoing_tor_connections;
```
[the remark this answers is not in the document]
53a45b2637452980: Will change.

src/cryptonote_protocol/levin_notify.cpp
53a45b2637452980: This was meant to be temporary, sorry you had to see it.

src/p2p/net_node.inl
53a45b2637452980: Is this not the perfect place for this logic? try_to_connect_and_handshake_with_new_peer has other logic related to connection counts and conditions for which connections should not be made, which is similar to my additions.

src/p2p/net_node.inl
53a45b2637452980: If it is one connection below max before a connection is made then can assume it is at max connections after the connection is made (later in this function). zone.m_current_number_of_out_peers is not guaranteed to be updated 
directly after the new connection so this is a cheap way to check.

src/p2p/net_node.inl
53a45b2637452980: I can't see why?

src/p2p/net_node.inl
53a45b2637452980: Will change.

src/p2p/net_node.inl
53a45b2637452980: Will change.

src/p2p/net_node.inl
53a45b2637452980: Will change.

src/cryptonote_protocol/levin_notify.cpp
53a45b2637452980: can_broadcast won't be set to false until the message is sent.

src/cryptonote_protocol/levin_notify.cpp
53a45b2637452980: Can you point out why this is insufficient? Once a connection is marked as unusable for broadcast it shouldn't ever broadcast again.Would you prefer immediately calling update_channels, or waiting for the new epoch? I'm concerned that messing with the connection pool for channels could help the ISP determine when a node broadcast based on noise patterns. If a node suddenly stops sending noise but the ISP didn't detect that a Tor stream was shutdown then the
probability that you just sent a transaction was high.

53a45b2637452980: Can you point out why this is insufficient? Once a connection is marked as unusable for broadcast it shouldn't ever broadcast again even though it can still be selected as the connection for a noise channel in the future. I can understand removing the connection from the pool but I'm concerned that it could result in noise patterns that the ISP could use to determine when a transaction was broadcast. If I implemented it by immediately removing connections from the pool and updating channels after broadcast then a node could quickly exhaust all connections and have nothing to send noise on. The ISP could detect that noise stopped but no Tor stream shutdown occurred and deduce that the node just broadcast. Eliminating a connection from the pool at the next epoch would probably be better but I don't know if it solves the issue entirely especially for setups with a low tx proxy count. Would you prefer immediately calling update_channels, or waiting for the new epoch?

src/p2p/net_node.inl
53a45b2637452980: On this topic, this patch works better if the tx proxy connection count is not set to a huge value otherwise the reconnect rate is never scaled down. Would you prefer that I implement logic to detect this based on connection count increase over time or just set a hard limit? Lots can go wrong with the first.

src/p2p/net_node.inl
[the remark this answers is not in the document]
vtnerd: But with so few tor peers already, this is going to reduce total connections on the network (because the peer is also running the same logic and may _need_ you as its outbound connection).

src/p2p/net_node.inl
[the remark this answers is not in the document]
vtnerd: See above.

src/p2p/net_node.inl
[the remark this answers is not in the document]
vtnerd: Yes, skimmed over this too quickly.

src/cryptonote_protocol/levin_notify.cpp
vtnerd: Then why have the check here? There's probably a way to re-organize this logic since its only set at the end.

src/cryptonote_protocol/levin_notify.cpp
vtnerd: Ok, I see your point, if noise is enabled, you cannot change anything or you alert observers that something changed (a tx was sent). I was thinking about the noise _disabled_ case unfortunately.

I noticed you made no changes for noise disabled, and I think that makes sense given that it just does a flood.

vtnerd: The user will be unable to send txes until the next epoch (and new outgoing connections are "selected"). Is that intended? Maybe, given the requirements of this patch.

The bigger issue is that when a new epoch starts, this current iteration may select a connection where sending is already disabled.

vtnerd: Sorry for not commenting this on my first comment. The primary issue is how to select the outgoing peers every epoch - it may want to check status before selecting (or not).

PR #8778, opened 2023-03-17
[WIP] bump OSX SDK, clang

ph4r05:
WIP


jeffro256:
What's the rational for bumping the OSX SDK version?

ph4r05:
I bumped SDK to 13.1, clang 12.0.0

I managed to fix OSX build issues by adding compat library for OSX detection https://github.com/monero-project/monero/blob/18ecddf4eb1169662f670021f48a3cd6e23dbb9c/contrib/depends/libs/osx-cross-compile/compat.cpp

Thing is that cross-compilation for OSX misses these functions, as mentioned in https://github.com/curl/curl/issues/4314

PR #8786, opened 2023-03-19
http_client: reduce number of packets sent for small bodies

jeffro256:
Right now, HTTP headers and the HTTP body are sent separately. This PR change sends the bodies and headers combined when appropriate. For small bodies, this can reduce the number of packets sent by 1. For a series of HTTP requests made with small bodies, the number of round trips can be reduced by up to 50%. 

You can test this by analyzing HTTP traffic with a packet sniffer like Wireshark.

### DIFF
--- a/contrib/epee/include/net/http_client.h
+++ b/contrib/epee/include/net/http_client.h
@@ -199,8 +199,18 @@ namespace net_utils
 					}
 				}
 
+				// This magic var determines the maximum length for when copying the body message in
+				// memory is faster/more preferable than the round-trip time for one packet
+				constexpr size_t BODY_NO_COPY_CUTOFF = 128 * 1024; // ~262 KB or ~175 packets
+
+				// Maximum expected total headers bytes
+				constexpr size_t HEADER_RESERVE_SIZE = 2048;
+
+				const bool do_copy_body = body.size() <= BODY_NO_COPY_CUTOFF;
+				const size_t req_buff_cap = HEADER_RESERVE_SIZE + (do_copy_body ? body.size() : 0);
+
 				std::string req_buff{};
-				req_buff.reserve(2048);
+				req_buff.reserve(req_buff_cap);
 				req_buff.append(method.data(), method.size()).append(" ").append(uri.data(), uri.size()).append(" HTTP/1.1\r\n");
 				add_field(req_buff, "Host", m_host_buff);
 				add_field(req_buff, "Content-Length", std::to_string(body.size()));
@@ -209,9 +219,7 @@ namespace net_utils
 				for(const auto& field : additional_params)
 					add_field(req_buff, field);
 
-				for (unsigned sends = 0; sends < 2; ++sends)
 				{
-					const std::size_t initial_size = req_buff.size();
 					const auto auth = m_auth.get_auth_field(method, uri);
 					if (auth)
 						add_field(req_buff, *auth);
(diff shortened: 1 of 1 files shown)


jeffro256:
To test:

1. Tack the following code to the end of the file `tests/unit_tests/http.cpp` for the the master branch:
```
TEST(HTTP_Client, Simple_Post)
{
  http::http_simple_client client;
  client.set_server("52.86.68.46", "80", boost::none, epee::net_utils::ssl_support_t::e_ssl_support_disabled);
  
  const http::http_response_info* presponse_info = NULL;
  EXPECT_TRUE(client.invoke("/post", "POST", "meep mop", std::chrono::seconds(10), &presponse_info));
}

```
2. Compile branch
3. Start capturing packets with Wireshark
4. Run `./build/Linux/<BRANCH NAME>/release/tests/unit_tests/unit_tests --gtest_filter=HTTP_Client.*`
5. Stop capturing with Wireshark
5. Count how many packets it takes to send request headers and body
6. Repeat steps 1-5 for this PR branch

plowsof:
i have an issue compiling:
```
monero/tests/unit_tests/http.cpp: In member function ‘virtual void HTTP_Client_Simple_Post_Test::TestBody()’:
/home/human/dev/monero-gui/monero/tests/unit_tests/http.cpp:749:9: error: ‘http_simple_client’ is not a member of ‘{anonymous}::http’
  749 |   http::http_simple_client client;
      |         ^~~~~~~~~~~~~~~~~~
monero/tests/unit_tests/http.cpp:750:3: error: ‘client’ was not declared in this scope
  750 |   client.set_server("52.86.68.46", "80", boost::none, epee::net_utils::ssl_support_t::e_ssl_support_disabled);
      |   ^~~~~~
monero/tests/unit_tests/http.cpp:750:72: error: ‘epee::net_utils::ssl_support_t’ has not been declared
  750 |   client.set_server("52.86.68.46", "80", boost::none, epee::net_utils::ssl_support_t::e_ssl_support_disabled);
      |                 
```

contrib/epee/include/net/http_client.h
```diff
@@ -209,21 +219,29 @@ namespace net_utils
...
 					req_buff += "\r\n";
 					//--
 
-					bool res = m_net_client.send(req_buff, timeout);
-					CHECK_AND_ASSERT_MES(res, false, "HTTP_CLIENT: Failed to SEND");
-					if(body.size())
+					if (do_copy_body) // small body
+					{
+						// Copy headers + body together and potentially send one fewer packet
+						req_buff.append(body.data(), body.size());
```
iamamyth, line 233: Why not change the client to offer a writev-style method? Then you can skip the copies.

contrib/epee/include/net/net_helper.h
```diff
@@ -487,79 +415,6 @@ namespace net_utils
...
-				LOG_ERROR("Some problems at read, message: " << er.what());
-        m_connected = false;
-				return false;
-			}
-			catch(...)
-			{
-				LOG_ERROR("Some fatal problems at read.");
-				return false;
-			}
-
```
iamamyth, line 557: Might be easier to split the dead code piece into its own PR.

jeffro256:
@plowsof Sorry I forgot that you also have to `#include "net/http_client.h"`

contrib/epee/include/net/http_client.h
jeffro256: `writev` isn't portable, would require much more code changes, and probably wouldn't be worth the effort since we only have a maximum of two distinct buffers to send. Since we're only copying small buffers in this PR, I'm making the performance bet that the time spent copying this small buffer is dwarfed by the network time it takes for TCP to make a round trip.

contrib/epee/include/net/net_helper.h
jeffro256: You're probably right.

jeffro256: Done

plowsof:
with this PR:
```
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from HTTP_Client
[ RUN      ] HTTP_Client.Simple_Post
[       OK ] HTTP_Client.Simple_Post (208 ms)
[----------] 1 test from HTTP_Client (208 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (208 ms total)
[  PASSED  ] 1 test.
```
Screenshot from 2023-03-19 23-48-55


without:
```
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from HTTP_Client
[ RUN      ] HTTP_Client.Simple_Post
[       OK ] HTTP_Client.Simple_Post (301 ms)
[----------] 1 test from HTTP_Client (301 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (301 ms total)
[  PASSED  ] 1 test.
```
Screenshot from 2023-03-20 00-14-34

Is this correct? i have `ip.addr == 52.86.68.46` as a filter in wireshark

jeffro256:
Yup! That looks just like how mine looked. Wireshark doesn't recognize that some of the TCP packets are being used for HTTP as well, but it you look through the contents of packets without the PR, you can see a packet with just the headers and a packet with the string "meep mop".

jeffro256:
So we went from sending/receiving 14 packets for this interaction down to 10.

0xFFFC0000:
I read the changes. To me this looks a great little optimization. Thanks @jeffro256 

I am running the tests in a few minutes and will update this comment.

0xFFFC0000 (approved):
LGTM. Passed the tests.

contrib/epee/include/net/http_client.h
vtnerd: I think @iamamyth was suggesting a change to the `m_net_client` interface to accept N buffers. I haven't looked too deeply on this, the easiest is probably `span<const boost::string_ref>` where a temporary array gets created here on the stack.

PR #8794, opened 2023-03-21
wallet2: fix infinite loop in fake out selection

moneromooo-monero:
The gamma picker and the caller code did not quite agree on the number of rct outputs available for use - by one block - which caused an infinite loop if the picker could never pick outputs from that block but already had picked all other outputs from previous blocks

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1006,7 +1006,7 @@ gamma_picker::gamma_picker(const std::vector<uint64_t> &rct_offsets, double shap
   const size_t blocks_to_consider = std::min<size_t>(rct_offsets.size(), blocks_in_a_year);
   const size_t outputs_to_consider = rct_offsets.back() - (blocks_to_consider < rct_offsets.size() ? rct_offsets[rct_offsets.size() - blocks_to_consider - 1] : 0);
   begin = rct_offsets.data();
-  end = rct_offsets.data() + rct_offsets.size() - CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE;
+  end = rct_offsets.data() + rct_offsets.size() - (std::max(1, CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE) - 1);
   num_rct_outputs = *(end - 1);
   THROW_WALLET_EXCEPTION_IF(num_rct_outputs == 0, error::wallet_internal_error, "No rct outputs");
   average_output_time = DIFFICULTY_TARGET_V2 * blocks_to_consider / static_cast<double>(outputs_to_consider); // this assumes constant target over the whole rct range
@@ -8615,7 +8615,7 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
       else
       {
         // the base offset of the first rct output in the first unlocked block (or the one to be if there's none)
-        num_outs = rct_offsets[rct_offsets.size() - CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE];
+        num_outs = gamma->get_num_rct_outs();
         LOG_PRINT_L1("" << num_outs << " unlocked rct outputs");
         THROW_WALLET_EXCEPTION_IF(num_outs == 0, error::wallet_internal_error,
             "histogram reports no unlocked rct outputs, not even ours");
@@ -8899,7 +8899,7 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
       }
       bool use_histogram = amount != 0;
       if (!use_histogram)
-        num_outs = rct_offsets[rct_offsets.size() - CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE];
+        num_outs = gamma->get_num_rct_outs();
 
       // make sure the real outputs we asked for are really included, along
       // with the correct key and mask: this guards against an active attack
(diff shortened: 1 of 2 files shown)


moneromooo-monero:
That's used for pre-rct outs only (line just above). And the gamma picker is used for rct.

moneromooo-monero:
I added this code instead, and just to be clear, it gives results off by one compared to the current gamma picker.

jeffro256:
Does this mean the gamma picker should be updated as well? Wasn't a similar off-by-one issue fixed by @j-berman recently?

jeffro256:
The gamma picker should be correct as-is since the consensus rules in `cryptonote::Blockchain` compares the maximum input height to the *current top block* and not the *block the transaction will end up in*. https://github.com/monero-project/monero/blob/f307621678c24ee7ca158fde38080d787d678733/src/cryptonote_core/blockchain.cpp#L3534

j-berman:
It looks like @UkoeHB is correct.

It also looks like `end` needs to be set like this:

```cpp
end = rct_offsets.data() + rct_offsets.size() - (std::max(1, CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE) - 1);
```

in order to use the correct range in the picker.

I sanity tested as follows:

1. Receive an output in a wallet, mine 10 blocks, note the output's `output_index`, then stop mining.
2. Have the `gamma_picker::pick()` loop 10000 times.
3. Make sure for some of those picks, it picks that received output as a decoy (replace the lines `// bad pick` with `continue`).
4. Confirm that you can spend that output from that wallet.

Still sitting with this to be certain.

moneromooo-monero:
Well, that code leads to failed unit_tests, the it == end assert triggers. Since rct_outputs[index] is dereferenced later, the assert seems correct.

moneromooo-monero:
m_db->height() returns the height of the chain, which also happens to be the height of the next block to be added.

moneromooo-monero:
Well, the tests are fixed by jberman's change above. Nice.

jeffro256:
```
auto top_spendable_it =
    std::next(
            rct_offsets.rbegin(),
            std::min(
                    rct_offsets.size(),
                    std::max(1, CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE) - 1
                )
        );
if (top_spendable_it == rct_offsets.rend())
    num_outs = 0;
else
    num_outs = *top_spendable_it;
```

This chunk of code can cause a seg fault since `rbegin` points to the last element, so going backwards `rct_offsets.size()` will point to the element before the first. This also (I think) is too conservative of picking by 1 since we're advancing from the last element, and not the next the index after the last.

jeffro256:
Nevermind, the `THROW_WALLET_EXCEPTION_IF(rct_offsets.size() <= CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE, error::wallet_internal_error, "Bad offset calculation");` clause protects against this, but also makes the `std::min` part unnecessary.

src/wallet/wallet2.cpp
```diff
@@ -1005,9 +1005,23 @@ gamma_picker::gamma_picker(const std::vector<uint64_t> &rct_offsets, double shap
...
+            rct_offsets.rbegin(),
+            std::min<size_t>(
+                    rct_offsets.size(),
+                    std::max(1, CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE) - 1
+                )
+        );
+  if (top_spendable_it == rct_offsets.rend())
+    num_rct_outputs = 0;
+  else
+    num_rct_outputs = *top_spendable_it;
```
jeffro256, line 1023:
```suggestion
num_rct_outputs = *(end - 1);
```

Now that we've changed the end iterator calculation, this is now correct again

jeffro256:
> > This chunk of code can cause a seg fault since rbegin points to the last element, so going backwards rct_offsets.size() will point to the element before the first.
> 
> rbegin + size = rend. Reverse iterators store the next element's forward iterator, so this is equivalent to end - size = begin

Okay, you're right about that, my bad. We can still skip the `top_spendable_it` calculation though, right?

j-berman:
Also looks like @UkoeHB was right calling out that other spot where `num_outs` is set here:

https://github.com/monero-project/monero/blob/f307621678c24ee7ca158fde38080d787d678733/src/wallet/wallet2.cpp#L8900-L8902

That `num_outs` is used here when either pre-rct or rct outputs from rings saved in ringdb are re-selected:

https://github.com/monero-project/monero/blob/f307621678c24ee7ca158fde38080d787d678733/src/wallet/wallet2.cpp#L8925-L8955

So it seems `num_outs` should probably be set in [that spot](https://github.com/monero-project/monero/pull/8794#issuecomment-1478064060) using `gamma->get_num_rct_outs()` as well to make sure decoys in that range can be re-added to a ring too.

j-berman (approved):

moneromooo-monero:
A much smaller patch is better.

j-berman (approved):

jeffro256 (approved):

jeffro256:
Please double check me on this, but can't we loosen the requirements on the `gamma_picker` constructor from `THROW_WALLET_EXCEPTION_IF(rct_offsets.size() <= CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE, error::wallet_internal_error, "Bad offset calculation")` to `THROW_WALLET_EXCEPTION_IF(rct_offsets.size() < CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE, error::wallet_internal_error, "Bad offset calculation")` (changed `<=` -> `<`) because of the new `end` calculation?

selsta (approved):

jeffro256 (approved):

Issue #8793, opened 2023-03-21
relay a confirmed tx to wallet produced dirty data


wangsai-silence:
- First, I create a tx with wallet `transfer` RPC 
```
curl http://localhost:18082/json_rpc \
-u admin:123 --digest \
-H 'Content-Type: application/json' \
-d '{"jsonrpc":"2.0","id":"0","method":"transfer","params":{"destinations":[{"amount":60,"address":"7BWQGPqwy3TAbykGqDoD9fXq6GS84121XRRKJvcb2y6Md2Bed42mvSwF5LqeVAqtrQVo1T9D3WsqP3THyLW4cZWTFjY8p2g"}],"account_index":0,"get_tx_key":false,"get_tx_hex":false, "do_not_relay": true, "get_tx_metadata":true}}' 
```

- Then, I relay it to chain
```
curl http://localhost:18082/json_rpc  \
-u admin:123 --digest \
-H 'Content-Type: application/json' \
-d '{"jsonrpc":"2.0","id":"0","method":"relay_tx","params":{"hex":"metadata..."}}'  \

{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "tx_hash": "3850e52507f5a9b7038417d758bb1c12e93bb57187d562c614be81844735024e"
  }
}

```

- After tx is confirmed. I check the wallet balance

```
curl http://127.0.0.1:18082/json_rpc \
-u admin:123 --digest \
-H 'Content-Type: application/json' \
-d '{"jsonrpc":"2.0","id":"0","method":"get_balance","params":{"account_index":0,"address_indices":[]}}' \

{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "balance": 61085212629330,
    "blocks_to_unlock": 5,
    "multisig_import_needed": false,
    "per_subaddress": [...],
    "time_to_unlock": 0,
    "unlocked_balance": 60906659842570
  }
}
```

- It seems all good, until I relay it again. The wallet accepts it and returns a tx_id as step 2
```
curl http://localhost:18082/json_rpc  \
-u admin:123 --digest \
-H 'Content-Type: application/json' \
-d '{"jsonrpc":"2.0","id":"0","method":"relay_tx","params":{"hex":"metadata..."}}'  \

{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "tx_hash": "3850e52507f5a9b7038417d758bb1c12e93bb57187d562c614be81844735024e"
  }
}

```

[...]

PR #8796, opened 2023-03-21
wallet2: do not commit transactions more than once

moneromooo-monero:


hyc:
Shouldn't you have "fix #8793" in this commit msg?

jeffro256:
Just to clarify: since the new checks in this PR occur after the RPC request to /sendrawtransaction, this will always attempt to relay the transaction regardless, but just won't commit the balance recovery information more than once, correct?

moneromooo-monero:
Yes.

plowsof:
repro'd #8793 , this PR fixes the issue (i will need to rtfm and get these python tests running)

jeffro256 (approved):

PR #8798, opened 2023-03-22
wallet2: make m_process_pool_txs local to refresh method

jeffro256:
Refines #8076 by reducing side effects of new pool state methods. This will make integrating `m_pool_info_query_time` into `NodeRPCProxy` easier.


jeffro256:
@rbrunner7 @j-berman

rbrunner7 (approved):
Looks good to me.

PR #8799, opened 2023-03-24
Blockchain: speed up get_output_distribution for RCT by about ~15%

jeffro256:
Note: this does make the `BlockchainDB::get_block_cumulative_rct_outputs()` method less flexible for one invocation, but I can't think of a good use case where it would need random arbitrary access to different heights.

### DIFF
--- a/src/blockchain_db/blockchain_db.h
+++ b/src/blockchain_db/blockchain_db.h
@@ -971,11 +971,13 @@ public:
    *
    * If the block does not exist, the subclass should throw BLOCK_DNE
    *
-   * @param height the height requested
+   * @param start_height the first height to start accumulating (inclusive)
+   * @param end_height the last height to accumulatate (inclusive)
+   * @param[out] base the cumulative number of rct outputs for all blocks with height < start_height
    *
    * @return the cumulative number of rct outputs
    */
-  virtual std::vector<uint64_t> get_block_cumulative_rct_outputs(const std::vector<uint64_t> &heights) const = 0;
+  virtual std::vector<uint64_t> get_block_cumulative_rct_outputs(uint64_t start_height, uint64_t end_height, uint64_t& base) const = 0;
 
   /**
    * @brief fetch the top block's timestamp
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -2490,16 +2490,20 @@ uint64_t BlockchainLMDB::get_block_timestamp(const uint64_t& height) const
   return ret;
 }
 
-std::vector<uint64_t> BlockchainLMDB::get_block_cumulative_rct_outputs(const std::vector<uint64_t> &heights) const
+std::vector<uint64_t> BlockchainLMDB::get_block_cumulative_rct_outputs(uint64_t start_height, uint64_t end_height, uint64_t& base) const
 {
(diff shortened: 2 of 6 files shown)


jeffro256:
To test performance, apply diff to both master and this PR:
```
diff --git a/src/cryptonote_core/cryptonote_core.cpp b/src/cryptonote_core/cryptonote_core.cpp
index 70431d23a..7b1fde46d 100644
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1504,7 +1504,17 @@ namespace cryptonote
   bool core::get_output_distribution(uint64_t amount, uint64_t from_height, uint64_t to_height, uint64_t &start_height, std::vector<uint64_t> &distribution, uint64_t &base) const
   {
-    return m_blockchain_storage.get_output_distribution(amount, from_height, to_height, start_height, distribution, base);
+    const uint64_t start = epee::misc_utils::get_tick_count();
+    const auto res = m_blockchain_storage.get_output_distribution(amount, from_height, to_height, start_height, distribution, base);
+    const uint64_t end = epee::misc_utils::get_tick_count();
+
+    if (amount == 0)
+    {
+      const uint64_t elapsed = end - start;
+      LOG_PRINT_L0("get_output_distribution for RCT took " << elapsed << " milliseconds to retreive data");
+    }
+
+    return res;
   }
   bool core::get_tx_outputs_gindexs(const crypto::hash& tx_id, std::vector<uint64_t>& indexs) const
```

Spin up `monerod` and execute `curl http://127.0.0.1:18081/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"get_output_distribution","params":{"amounts":[0],"from_height":0,"to_height":"2838000","cumulative":false,"binary":true,"compress":true}}' -H 'Content-Type: application/json'`

Then check the `monerod` log for the string "get_output_distribution for RCT took", and write down measurement, no need to change log levels.

Notice that this will only print once since the RCT distribution is cached by `cryptonote::rpc::RpcHandler`, so for multiple points of data, you will have to close and reopen `monerod`.

jeffro256:
@hyc

hyc:
I haven't compiled it yet but overall the patch looks sane.

plowsof:
confirming it compiles/works on stagenet: (will re-test on mainnet later then edit)
with: 231 milliseconds to retreive data
without: 255 milliseconds to retreive data

jeffro256:
> confirming it compiles/works on stagenet: (will re-test on mainnet later then edit) with: 231 milliseconds to retreive data without: 255 milliseconds to retreive data
> 
> im getting some inconsistent results on mainnet (sometimes slower/faster with the patch - maybe user error on my part) and restarting monerod seems to not remove the cache for retesting - i have to recompile

Yeah depending on how finicky your OS cache is, this PR can be very tricky to consistently benchmark. Since LMDB virtually maps the file directly into memory, if your OS doesn't keep those pages loaded in memory after you restart `monerod`, then the next time `monerod` tries to access those pages it causes a soft seg fault and has to retrieve it from storage again. This operation dwarfs the amount of time saved in this PR and will cause the results to look wonky, like some invocations taking 10x longer than others. My machine has 32GB of memory and happens to pretty much always have those pages cached after they are accessed once, so I always get the benefit of this PR. For some machines, this PR might make no difference performance-wise.

jeffro256:
Not worth review effort after FCMP++

PR #8806, opened 2023-03-27
tx_extra limitation fix [release-v0.18]

SChernykh:
- Fixed connection drop when a big tx_extra is received from another node
- Double check for MAX_TX_EXTRA_SIZE before sending out ZMQ txpool_event notifications to properly mark them as `invalid`

The second fix (ZMQ notifications) is to stop the notification spam when a new block with big tx_extra transactions is received - it brings unneccessary load to p2pool miners.

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1099,7 +1099,7 @@ namespace cryptonote
       else if(tvc[i].m_verifivation_impossible)
       {MERROR_VER("Transaction verification impossible: " << results[i].hash);}
 
-      if(tvc[i].m_added_to_pool)
+      if(tvc[i].m_added_to_pool && (results[i].tx.extra.size() <= MAX_TX_EXTRA_SIZE))
       {
         MDEBUG("tx added: " << results[i].hash);
         valid_events = true;
--- a/src/cryptonote_protocol/cryptonote_protocol_handler.inl
+++ b/src/cryptonote_protocol/cryptonote_protocol_handler.inl
@@ -1022,8 +1022,18 @@ namespace cryptonote
       tx_verification_context tvc{};
       if (!m_core.handle_incoming_tx({tx, crypto::null_hash}, tvc, tx_relay, true))
       {
-        LOG_PRINT_CCONTEXT_L1("Tx verification failed, dropping connection");
-        drop_connection(context, false, false);
+        // Don't drop connections which send transactions with too big tx_extra
+        // It will only cause updated nodes to isolate themselves from the rest of the network
+        // This check can be removed as soon as updated nodes are the majority (both by hashrate and by total count)
+        if (tvc.m_tx_extra_too_big)
+        {
+          LOG_PRINT_CCONTEXT_L1("Tx verification failed (too big tx_extra)");
+        }
+        else
+        {
+          LOG_PRINT_CCONTEXT_L1("Tx verification failed, dropping connection");
+          drop_connection(context, false, false);
+        }
         return 1;
       }
 


src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1092,14 +1092,19 @@ namespace cryptonote
 
       results[i].blob_size = it->blob.size();
       results[i].weight = results[i].tx.pruned ? get_pruned_transaction_weight(results[i].tx) : get_transaction_weight(results[i].tx, it->blob.size());
-      ok &= add_new_tx(results[i].tx, results[i].hash, tx_blobs[i].blob, results[i].weight, tvc[i], tx_relay, relayed);
+      if (!add_new_tx(results[i].tx, results[i].hash, tx_blobs[i].blob, results[i].weight, tvc[i], tx_relay, relayed))
+      {
+        // Transactions with big tx_extra are still ok to be mined, so just ignore them and don't return false
+        if (!tvc[i].m_tx_extra_too_big)
+          ok = false;
```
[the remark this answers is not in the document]
SChernykh: In short: the worst case, some connections that should've been dropped, will not be dropped sometimes, but otherwise invalid transactions will never be relayed or added to mempool.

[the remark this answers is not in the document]
SChernykh: Please constructive notes only. This PR was created not to refactor the code.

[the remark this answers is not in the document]
SChernykh: @UkoeHB I can change `tx_memory_pool::add_tx` to perform tx_extra check last, so second case will never happen. Then "tx extra is too big and something else" case will result in exactly the same node behavior as before this PR. Will it be good enough for you?

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1092,14 +1092,19 @@ namespace cryptonote
...
+      }
 
       if(tvc[i].m_verifivation_failed)
       {MERROR_VER("Transaction verification failed: " << results[i].hash);}
       else if(tvc[i].m_verifivation_impossible)
       {MERROR_VER("Transaction verification impossible: " << results[i].hash);}
 
-      if(tvc[i].m_added_to_pool)
+      if(tvc[i].m_added_to_pool && (results[i].tx.extra.size() <= MAX_TX_EXTRA_SIZE))
```
jeffro256, line 1107: Why is this check necessary? `m_added_to_pool` is only ever set *after* the `tx.extra` checks in `add_new_tx`. In other words, it seems redundant because `m_added_to_pool` will be false if `tx.extra` is too big anyways.

src/cryptonote_core/cryptonote_core.cpp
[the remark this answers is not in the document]
SChernykh: I'm telling you it's not. tvc will record "too big tx extra", and it will still record `tvc.m_verifivation_failed = true;`, so all code handling tvc will treat this transaction as invalid. Converstaion over, I will not change this logic.

src/cryptonote_core/cryptonote_core.cpp
SChernykh: It can happen if `kept_by_block` is true, and then I get tx spam in p2pool logs - all big tx_extra transactions get added to a new block and leak into ZMQ notifications, but they shouldn't because ZMQ notification are supposed to notify abou mempool only.

src/cryptonote_core/cryptonote_core.cpp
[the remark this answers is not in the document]
SChernykh: Ok, this is first valid point I got in this conversation. It can potentially break `core_rpc_server::on_send_raw_tx` and `DaemonHandler::handleTxBlob`. I'll probably move the connection drop fix to `cryptonote_protocol_handler.inl` where it really belongs.

src/cryptonote_core/cryptonote_core.cpp
SChernykh: @jeffro256 I just did a quick test by adding some logging: https://paste.debian.net/hidden/15de343a/
I can see both places filtering transactions from being sent to ZMQ subscribers, so it's correct. `tvc[i].m_added_to_pool` can be true but we still end up in the `else` branch, and the debug logging shows it.

SChernykh: Correction: https://paste.debian.net/hidden/d345d788/ - now I see only the first place filtering transactions, I'll keep running it to see if the second one triggers or not.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1022,8 +1022,18 @@ namespace cryptonote
       tx_verification_context tvc{};
       if (!m_core.handle_incoming_tx({tx, crypto::null_hash}, tvc, tx_relay, true))
       {
-        LOG_PRINT_CCONTEXT_L1("Tx verification failed, dropping connection");
-        drop_connection(context, false, false);
+        // Don't drop connections which send transactions with too big tx_extra
+        // It will only cause updated nodes to isolate themselves from the rest of the network
+        // This check can be removed as soon as updated nodes are the majority (both by hashrate and by total count)
```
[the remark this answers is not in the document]
SChernykh: "can be removed" != "must be removed", not even "should be removed"

src/cryptonote_core/cryptonote_core.cpp
SChernykh: When a new block arrives, monerod downloads missing transactions from this block, adds them to mempool for a split second and then removes them from mempool and adds them to the blockchain. All this juggling triggers ZMQ notifications which are useless to subscribers.

jeffro256: Since downloaded transactions are "forced into" the mempool  by `kept_by_block`, can we not generalize this condition to `tvc[i].m_added_to_pool && tx_relay != relay_method::block`? This would exclude any internal blockchain shufflings inside the pool, not just for `tx_extra`.

SChernykh: I'll run with debug logging to confirm that it's only `relay_method::block` there, and if it is, I'll change to `&& tx_relay != relay_method::block`

SChernykh: @jeffro256 I just realized that `kept_by_block` can make transactions stay in mempool for real, and be actually valid to be mined again, for example after a reorg. So it's a bad idea to change to `tx_relay != relay_method::block`, it's better to filter only big tx_extra.

SChernykh:
Replaced with #8813

PR #8807, opened 2023-03-27
tx_extra limitation fix

SChernykh:
- Fixed connection drop when a big tx_extra is received from another node
- Double check for MAX_TX_EXTRA_SIZE before sending out ZMQ txpool_event notifications to properly mark them as `invalid`

The second fix (ZMQ notifications) is to stop the notification spam when a new block with big tx_extra transactions is received - it brings unneccessary load to p2pool miners.


SChernykh:
Replaced with #8808

PR #8808, opened 2023-03-27
cryptonote core/protocol: don't drop peers for soft offenses

jeffro256:
Alternative to #8806, and addresses issue in https://github.com/monero-project/monero/pull/8806#issuecomment-1485863660. So far it is untested.

### DIFF
--- a/src/cryptonote_protocol/cryptonote_protocol_handler.inl
+++ b/src/cryptonote_protocol/cryptonote_protocol_handler.inl
@@ -1020,7 +1020,7 @@ namespace cryptonote
     for (auto& tx : arg.txs)
     {
       tx_verification_context tvc{};
-      if (!m_core.handle_incoming_tx({tx, crypto::null_hash}, tvc, tx_relay, true))
+      if (!m_core.handle_incoming_tx({tx, crypto::null_hash}, tvc, tx_relay, true) && !tvc.m_no_drop_offense)
       {
         LOG_PRINT_CCONTEXT_L1("Tx verification failed, dropping connection");
         drop_connection(context, false, false);
--- a/src/cryptonote_basic/verification_context.h
+++ b/src/cryptonote_basic/verification_context.h
@@ -42,7 +42,12 @@ namespace cryptonote
     static_assert(unsigned(relay_method::none) == 0, "default m_relay initialization is not to relay_method::none");
 
     relay_method m_relay; // gives indication on how tx should be relayed (if at all)
-    bool m_verifivation_failed; //bad tx, should drop connection
+    bool m_verifivation_failed; //bad tx, tx should not enter mempool and connection should be dropped unless m_no_drop_offense
+    // Do not add to mempool, do not relay, but also do not punish the peer for sending or drop
+    // connections to them. Used for low fees, tx_extra too big, "relay-only rules". Not to be
+    // confused with breaking soft fork rules, because tx could be later added to the chain if mined
+    // because it does not violate consensus rules.
+    bool m_no_drop_offense;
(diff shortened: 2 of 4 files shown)


src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -664,7 +664,7 @@ namespace cryptonote
           {
             MDEBUG("Incoming tx " << tx_hash << " not in pool, adding");
             cryptonote::tx_verification_context tvc = AUTO_VAL_INIT(tvc);                        
-            if(!m_core.handle_incoming_tx(tx_blob, tvc, relay_method::block, true) || tvc.m_verifivation_failed)
+            if((!m_core.handle_incoming_tx(tx_blob, tvc, relay_method::block, true) || tvc.m_verifivation_failed) && !tvc.m_no_drop_offense)
```
jeffro256, line 667: Only line I'm really unsure of is here

SChernykh:
`handle_notify_new_transactions` is where you should check `m_no_drop_offense` too.

jeffro256:
This can easily be done by deleting one line in `tx_pool.cpp` for the broken rule that you want it to drop from now on.

jeffro256:
Like it if you wanted to drop connections for large `tx_extra` starting today, you would just delete the corresponding `tvc.m_no_drop_offense = true;` line.

jeffro256:
This is definitely the final version, no more edits.

SChernykh:
If this replaces #8806 (actually #8807), it must also have the second commit with the ZMQ fix.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1020,7 +1020,7 @@ namespace cryptonote
     for (auto& tx : arg.txs)
     {
       tx_verification_context tvc{};
-      if (!m_core.handle_incoming_tx({tx, crypto::null_hash}, tvc, tx_relay, true))
+      if (!m_core.handle_incoming_tx({tx, crypto::null_hash}, tvc, tx_relay, true) && !tvc.m_no_drop_offense)
```
j-berman, line 1023: It looks like a node would relay a tx that has `m_no_drop_offense` set to true because it continues executing past the if statement.

Looks like you'd want something more like:

```cpp
if (!m_core.handle_incoming_tx({tx, crypto::null_hash}, tvc, tx_relay, true))
{
        if (tvc.m_no_drop_offense)
        {          
            LOG_PRINT_CCONTEXT_L1("Tx verification failed with a no drop offense");
            continue;
        }
        LOG_PRINT_CCONTEXT_L1("Tx verification failed, dropping connection");
        drop_connection(context, false, false);
        return 1;
}
```

j-berman: Nevermind, `tvc.m_relay` doesn't get set so should remain set to defaulted `relay_method::none` so it should fall into the final case of the switch. Sorry

j-berman (approved):

plowsof:
I ran the earlier patch from sech on my node for 24~ hours without connections dropping. I'm now running this with no problems. I ran a quick test using an old node connected to one using this patch as a priority node. began broadcasting ordinals. they did not appear in the mempool of the patched node (nor where they mined by it). ran a sweep_all too , 77 ish transactions with 146in16outs to sanity check a large normal tx and they where broadcast fine

src/cryptonote_protocol/cryptonote_protocol_handler.inl
SChernykh: I've added a debug log to check this on my node, and it's never anything else but `relay_method::none`

PR #8810, opened 2023-03-29
Fixed deadlock and crash when syncing with full dataset on Windows

SChernykh:
It's not allowed to use WaitForSingleObject with _beginthread, because the thread closes its own handle before exiting.

So the wait function will either wait on an invalid handle, or on a different handle used by something else.

Or, if it starts waiting before the thread exits, the behavior is undefined according to MS: "If this handle is closed while the wait is still pending, the function's behavior is undefined."

In my test sync I observed threads getting stuck infinitely on WaitForSingleObject, and then rx_set_main_seedhash spamming new threads when RandomX seed changes again. Eventually the system ran out of resources, and monerod aborted with "Couldn't start RandomX seed thread" message. See the attached screenshot of how it looked like before it crashed (note number of threads and handles).
image

This PR fixes it by using `_beginthreadex` instead and explicitly closing the handle when it's safe.

### DIFF
--- a/src/crypto/c_threads.h
+++ b/src/crypto/c_threads.h
@@ -42,10 +42,11 @@
 #define CTHR_RWLOCK_TRYLOCK_READ(x)	TryAcquireSRWLockShared(&x)
 
 #define CTHR_THREAD_TYPE	HANDLE
-#define CTHR_THREAD_RTYPE	void
-#define CTHR_THREAD_RETURN	return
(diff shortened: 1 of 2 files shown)


SChernykh:
@hyc you should review this.

hyc:
Looks good.

hyc (approved):

krypt0x (approved):
It looks alright!

PR #8815, opened 2023-03-30
decoy selection: spend like coinbase-ish outputs

jeffro256:
Implements solution recommended here (https://github.com/monero-project/research-lab/issues/109) in an efficient manner.

To retrieve the ENTIRE post-RCT coinbase output distribution from a running daemon, execute: `curl http://127.0.0.1:18081/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"get_output_distribution","params":{"amounts":[],"from_height":0,"to_height":0, "binary": false,"get_rct_coinbase":true}}' -H 'Content-Type: application/json'`.


jeffro256:
This PR is now ready to review. To test, create a bunch of transactions and make sure that each input ring contains only either coinbase enotes or non-coinbase enotes.

Here's a general overview of the design of the PR. To start off, the biggest assumption made for this PR is that for each RCT block (a block with major version >= 4), inside each block, all coinbase enotes have their amount index equal to 0 for retrieval, and all coinbase enote indexes are less than the non-coinbase enote indexes. This means that if a wallet knows the coinbase offset distribution in addition to the total RCT offset distribution, the wallet can determine whether or not any global enote index is a coinbase enote just from its index without any extra information.

On the daemon side, I added in a class called `CoinbaseOutputDistributionCache`, which caches the "coinbase enote distribution" (how many coinbase notes there are per block). I also made a custom compression algorithm that is able to crush down this data very efficiently. In `src/blocks`, you can see that I added compressed blobs of data, which are compressed coinbase enote distributions. For reference, the mainnet blob is 4044 bytes long and stores the distribution for over 1.2 million blocks.

As for RPC, the `get_output_distribution{.bin}` endpoint response was modified to add an extra field called `coinbase_distribution`. This field can be enabled by setting the `coinbase` field in the request to `true`. If `binary`, and `compress` are also set to `true`, the `coinbase_distribution` field is compressed by using the aforementioned compression algorithm to get a lot of savings in bandwidth.

Finally, on the wallet side, the wallet does decoy selection by retrieving the total enote offsets `rct_offsets` and the coinbase enote offsets `rct_coinbase_offsets`. We can then calculate the non-coinbase enote offsets `rct_non_coinbase_offsets` by performing `rct_offsets - rct_coinbase_offsets` element-wise. Then, for each incoming transfers, we determine if the transfer is a coinbase enote or not, and point the gamma picker to `rct_coinbase_offsets` or `rct_non_coinbase_offsets`, respectively. Then gamma selection is done as normal, and we take the index returned by the gamma picker and readjust it back to a global enote index based on which distribution the gamma picker was using.

jeffro256:
I have rewrote the PR to use `BlockchainLMDB` acceleration in #8896. This speeds up fetching significantly

jeffro256:
FCMP++ makes this obsolete

Issue #8722, opened 2023-01-28
Discussion: deprecation/removal of pay-to-use RPC system from core repo


jeffro256:
## What It Is
The pay-to-use RPC system was added in PR #5357. Node operators can specify an XMR address for public RPC which clients must mine to in order to access their RPC. The idea was that node operators could be compensated some of the cost of operating a node, and that they would compete for the lowest RPC cost by manually adjusting the amount of "credits" received for a hash of certain difficulty. "Credits" are rewarded by clients submitting a valid block (except for network difficulty) which is mined to the node operator receive address, similar to how miner shares are rewarded in a traditional centralized mining pool.

## Is this feature used in the wild?
I created a python network crawler which counts how many public nodes on mainnet have this feature enabled. This script requires that a local unrestricted daemon is running with a well established p2p list.

```python3
import random
import requests

my_daemon_address = 'localhost:18081'
timeout_secs = 10

def invoke_json(addr, uri, json_data):
	resp = requests.post("http://" + addr + '/' + uri, json=json_data, timeout=timeout_secs)
	#assert(resp.status_code == 200)
	resp_json = resp.json()
	if 'json_rpc' not in uri:
		assert(resp_json['status'] == 'OK')
	return resp_json

def invoke_json_rpc(addr, method, params):
	json_data = {'method': method, 'params': params}
	resp_json = invoke_json(addr, 'json_rpc', json_data)
	result = resp_json.get('result')
	error = resp_json.get('error')
	return result, error

def is_peer_info_public_ipv4(peer_info):
	return ':' not in peer_info['host'] and 'rpc_port' in peer_info

def make_ipv4_addr(peer_info):
	return peer_info['host'] + ':' + str(peer_info['rpc_port'])

def filter_nodes(peer_list):
	return [make_ipv4_addr(peer) for peer in peer_list if is_peer_info_public_ipv4(peer)]

# Build valid public IPv4 node list
peers_res = invoke_json(my_daemon_address, 'get_peer_list', {})
white_peer_addrs = filter_nodes(peers_res['white_list'])
gray_peer_addrs = filter_nodes(peers_res['gray_list'])
random.shuffle(white_peer_addrs)
random.shuffle(gray_peer_addrs)
ipv4_public_addrs = white_peer_addrs + gray_peer_addrs # try white nodes before gray nodes
[...]

PR #8843, opened 2023-05-08
rpc: remove mine-to-use RPC on node side

jeffro256:
This is a follow-up PR to #8724. I was able to remove >2000 lines of code and shrink the `monerod` binary by >6%. You can look at the original PR to check up on usage stats for this feature, but last time I checked it was less than 3%.


moneromooo-monero:
> shrink the monerod binary by >6%.

6% fewer bytes vs the ability to implement a fully decentralized microtransaction payment system.

The world never stopped sucking I guess. Time to find a tiny violin somewhere...

Gingeropolous:
> 6% fewer bytes vs the ability to implement a fully decentralized microtransaction payment system.

with no chain bloat and no on-boarding!

it was a beautiful dream.

though I thought the idea was to remove it from the wallet (because aNtIvIrUs) and let it live in the daemon in the hopes that some future 3rd party wallet would do something or other.

jeffro256:
> > 6% fewer bytes vs the ability to implement a fully decentralized microtransaction payment system.
> 
> with no chain bloat and no on-boarding!
> 
> it was a beautiful dream.
> 
> though I thought the idea was to remove it from the wallet (because aNtIvIrUs) and let it live in the daemon in the hopes that some future 3rd party wallet would do something or other.

Yes I'm not pushing incredibly hard for removal on the node side btw, but this PR is available in case that we decide otherwise.

jeffro256:
> > shrink the monerod binary by >6%.
> 
> 6% fewer bytes vs the ability to implement a fully decentralized microtransaction payment system.
> 
> The world never stopped sucking I guess. Time to find a tiny violin somewhere...

Even if the code was removed from the core repo, this could still work as a reverse proxy right?

jeffro256:
Closing for now

hinto-janai:
@jeffro256 Do you have any thoughts on this now?

IMO the points in https://github.com/monero-project/monero/issues/8722 make this worth removing on the node side as well. FWIW if this stays in `monerod`, Cuprate is planning to implement the related fields and routes (`credits`, `rpc_access_info`, etc) on the surface-level but disable it internally.

PR #8849, opened 2023-05-11
wallet: fix DNS leak on wallet open

tobtoht:
Starting `monero-wallet-cli` with arguments 

```
--daemon-address any_address.onion:18081 --proxy 127.0.0.1:9050
``` 

produces a DNS leak when a wallet is opened (credit to r4v3r23 / pokkst for finding this). A lookup for the .onion daemon address using the system's default DNS resolver is done in `tools::is_local_address`.

The first commit in this PR fixes a bug in `is_local_address` where the full daemon address (including port) is passed to `is_privacy_preserving_network` instead of just the host, which it expects.

In the second commit, changes are made so that `is_local_address` respects the `--no-dns` flag.

Checking if an address is local using DNS does not make sense when a proxy is set, since a daemon behind a proxy can't be considered local in the first place. (We probably shouldn't be using DNS for this at all. Are there any compelling reasons why someone would configure their nameserver to point some arbitrary domain to a loopback address? A simple solution would be to only parse IP addresses for this check and ask that users enter that instead. Maybe parse the hosts file if we're feeling generous). Anyway, the third commit disables DNS for checking if an address is local when a proxy is set. This prevents a leak when a proxy is used in combination with a clearnet daemon address.

### DIFF
--- a/src/common/util.cpp
+++ b/src/common/util.cpp
@@ -880,7 +880,7 @@ std::string get_nix_version_display_string()
     return false;
   }
 
-  bool is_local_address(const std::string &address)
+  bool is_local_address(const std::string &address, bool use_dns)
   {
     // extract host
     epee::net_utils::http::url_content u_c;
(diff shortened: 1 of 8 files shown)


tobtoht:
Closing in favor of #8854

PR #8854, opened 2023-05-12
common: do not use DNS to determine if address is local

tobtoht:
Alternative to #8849. 

I'm not convinced we need to support DNS lookup at all here. This PR considers a node "local" if the host is "localhost" or a IPv{4,6} address in the loopback range. The `is_privacy_preserving_network` check is no longer needed.

### DIFF
--- a/src/common/util.cpp
+++ b/src/common/util.cpp
@@ -610,13 +610,6 @@ namespace tools
 
   bool is_local_address(const std::string &address)
   {
-    // always assume Tor/I2P addresses to be untrusted by default
-    if (is_privacy_preserving_network(address))
-    {
-      MDEBUG("Address '" << address << "' is Tor/I2P, non local");
-      return false;
-    }
-
     // extract host
     epee::net_utils::http::url_content u_c;
     if (!epee::net_utils::parse_url(address, u_c))
@@ -630,20 +623,22 @@ namespace tools
       return false;
     }
 
-    // resolve to IP
-    boost::asio::io_service io_service;
-    boost::asio::ip::tcp::resolver resolver(io_service);
-    boost::asio::ip::tcp::resolver::query query(u_c.host, "");
-    boost::asio::ip::tcp::resolver::iterator i = resolver.resolve(query);
-    while (i != boost::asio::ip::tcp::resolver::iterator())
+    if (u_c.host == "localhost" || boost::ends_with(u_c.host, ".localhost")) { // RFC 6761 (6.3)
+      MDEBUG("Address '" << address << "' is local");
+      return true;
+    }
+
+    boost::system::error_code ec;
+    const auto parsed_ip = boost::asio::ip::address::from_string(u_c.host, ec);
+    if (ec) {
+      MDEBUG("Failed to parse '" << address << "' as IP address: " << ec.message() << ". Considering it not local");
+      return false;
+    }
+
+    if (parsed_ip.is_loopback())
     {
-      const boost::asio::ip::tcp::endpoint &ep = *i;
-      if (ep.address().is_loopback())
-      {
-        MDEBUG("Address '" << address << "' is local");
-        return true;
-      }
-      ++i;
+      MDEBUG("Address '" << address << "' is local");
+      return true;
     }
 
     MDEBUG("Address '" << address << "' is not local");
--- a/tests/unit_tests/CMakeLists.txt
+++ b/tests/unit_tests/CMakeLists.txt
@@ -91,6 +91,7 @@ set(unit_tests_sources
   unbound.cpp
   uri.cpp
   variant.cpp
+  util.cpp
   varint.cpp
   ver_rct_non_semantics_simple_cached.cpp
   ringct.cpp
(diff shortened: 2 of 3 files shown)


plowsof:
May i confirm my method of seeing the bug (after looking at 8849) - simply using wireshark to show that dns A records at the onion daemon address are being looked up at wallet open time?

tobtoht:
@plowsof, yes, A or AAAA records. It should no longer happen with 8849 or this PR.

pokkst:
@plowsof You can test it in just about every wallet. The only wallet I've tested so far that didn't seem to experience the issue is Feather. For other wallets, it happens for onion addresses, i2p addresses, and clearnet addresses (clearnet is not really an issue if proxy is disabled, of course).

plowsof:
with wireshark open and a display filter of `dns.qry.name contains ".onion"` 
with release and this PR i still see the same number of dns calls when i e.g. unlock the wallet. Am i testing with the wrong parameters? (no --proxy or torsocks, so the inputs wont work)
 `./monero-wallet-cli --daemon-address plowsof3t5hogddwabaeiyrno25efmzfxyro2vligremt7sxpsclfaid.onion:18089`

tobtoht:
@plowsof `--proxy` is needed for testing. A DNS leak happens elsewhere if `--proxy` is omitted, because it assumes it's connecting a clearnet host. It might be a good idea to have simplewallet warn* if a user tries to connect to a .onion/i2p daemon address without proxy, but this is a separate issue.

* Or refuse to connect, though this would frustrate users with transparent Tor proxying (e.g. Whonix).

src/common/util.cpp
```diff
@@ -902,20 +895,26 @@ std::string get_nix_version_display_string()
...
+    try
     {
-      const boost::asio::ip::tcp::endpoint &ep = *i;
-      if (ep.address().is_loopback())
-      {
-        MDEBUG("Address '" << address << "' is local");
-        return true;
-      }
-      ++i;
+      parsed_ip = boost::asio::ip::address::from_string(u_c.host);
```
vtnerd, line 906: Use the overload that takes a `boost::system::error_code`, `parsed_ip` can be made `const`, etc.

vtnerd (changes requested):

src/common/util.cpp
vtnerd: The error message might as well contain `error.message()` in the `MWARNING` output too.

tobtoht: Thanks

plowsof:
confirming that this PR fixes the issue (no dns leak with an .onion daemon and --proxy where release would otherwise do so)

j-berman (approved):
Small comment on the warning log, otherwise looks good to me.

I wrote some unit tests:

```
#include "common/util.h"

TEST(LocalAddress, localhost) { ASSERT_TRUE(tools::is_local_address("localhost")); }
TEST(LocalAddress, localhost_port) { ASSERT_TRUE(tools::is_local_address("localhost:18081")); }
TEST(LocalAddress, localhost_suffix) { ASSERT_TRUE(tools::is_local_address("test.localhost")); }
TEST(LocalAddress, loopback) { ASSERT_TRUE(tools::is_local_address("127.0.0.1")); }
TEST(LocalAddress, loopback_port) { ASSERT_TRUE(tools::is_local_address("127.0.0.1:18081")); }
TEST(LocalAddress, loopback_protocol) { ASSERT_TRUE(tools::is_local_address("http://127.0.0.1")); }
TEST(LocalAddress, loopback_hi) { ASSERT_TRUE(tools::is_local_address("127.255.255.255")); }
TEST(LocalAddress, loopback_lo) { ASSERT_TRUE(tools::is_local_address("127.0.0.0")); }
TEST(LocalAddress, loopback_ipv6) { ASSERT_TRUE(tools::is_local_address("[0:0:0:0:0:0:0:1]")); }

TEST(LocalAddress, onion) { ASSERT_FALSE(tools::is_local_address("vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd.onion")); }
TEST(LocalAddress, i2p) { ASSERT_FALSE(tools::is_local_address("xmrto2bturnore26xmrto2bturnore26xmrto2bturnore26xmr2.b32.i2p")); }
TEST(LocalAddress, valid_ip) { ASSERT_FALSE(tools::is_local_address("1.2.3.4")); }
TEST(LocalAddress, valid_ipv6) { ASSERT_FALSE(tools::is_local_address("[0:0:0:0:0:0:0:2]")); }
TEST(LocalAddress, valid_domain) { ASSERT_FALSE(tools::is_local_address("getmonero.org")); }
TEST(LocalAddress, local_prefix) { ASSERT_FALSE(tools::is_local_address("localhost.com")); }
TEST(LocalAddress, invalid) { ASSERT_FALSE(tools::is_local_address("test")); }
TEST(LocalAddress, empty) { ASSERT_FALSE(tools::is_local_address("")); }
```

tobtoht:
Added your unit tests and suggested change, rebased so Windows build passes. Thanks for the review. :)

j-berman (approved):

selsta (approved):

jeffro256 (approved):

PR #8859, opened 2023-05-14
wallet: remove support for deprecated formats

tobtoht:
It's been a couple hardforks since we deprecated these. Opening old boost serialized wallet files remains possible.


src/wallet/wallet2.cpp
```diff
@@ -7080,52 +7034,10 @@ bool wallet2::parse_tx_from_str(const std::string &signed_tx_st, std::vector<too
...
-        LOG_PRINT_L0("Failed to parse decrypted data from signed transaction");
-        return false;
-      }
-    }
-    catch (const std::exception &e)
-    {
-      LOG_PRINT_L0("Failed to decrypt signed transaction: " << e.what());
-      return false;
-    }
+    LOG_PRINT_L0("Unsupported version in signed transaction");
```
iamamyth, line 7039: It might help to note the last version to support the format in the log message (here and elsewhere in this file).

vtnerd:
This forces people who have old wallets to recover from seed if they haven't kept up. I'm not sure if we've ever done this before. A tool that upgrades would be nice, but probably a maintenance nightmare.

tobtoht:
Old wallet files can still be opened. Only boost serialized formats that were covered under `set load-deprecated-formats 1` are affected. (MMS, signed/unsigned txs, output exports, reserve proofs).

PR #8860, opened 2023-05-15
depends: update expat to 2.5.0

tobtoht:
https://github.com/libexpat/libexpat/blob/fe8ff0345f889b322215739a3fa7b6ca8a8a6bbe/expat/Changes


jeffro256:
Correct me if I'm wrong, but I don't believe that we use expat anymore so that dependency can be removed completely.

tobtoht:
It's a non-optional dependency for unbound.

See:
- https://github.com/monero-project/monero/blob/master/contrib/depends/packages/unbound.mk#L6
- https://github.com/NLnetLabs/unbound/blob/master/README.md?plain=1#L19

selsta (approved):

PR #8862, opened 2023-05-17
protocol: drop peers sending duplicate txes

moneromooo-monero:


vtnerd (approved):

jeffro256:
Would it be better to check to "seen" txids versus blobs since it's trivial to construct a transaction that it semantically equivalent with an equivalent TXID but which has a different blob?

moneromooo-monero:
Explain ?
The reason I used blobs is that it does not require parsing.

jeffro256:
Sorry I was getting the varint serialization formats mixed up, ignore my last comment.

jeffro256:
Could you force-push to run the tests again please?

selsta:
@jeffro256 can you approve?

selsta (approved):

jeffro256:
It looks okay at first glance, but I haven’t tested it yet and I’m not super familiar with this part of the codebase.

PR #8865, opened 2023-05-20
cryptonote_basic: remove unused to_script[_hash] vin/vout

jeffro256:
Structs were added 9 years ago and have yet to be used

### DIFF
--- a/src/cryptonote_basic/cryptonote_basic.h
+++ b/src/cryptonote_basic/cryptonote_basic.h
@@ -53,26 +53,7 @@
 
 namespace cryptonote
 {
-  typedef std::vector<crypto::signature> ring_signature;
-
-
   /* outputs */
-
-  struct txout_to_script
-  {
-    std::vector<crypto::public_key> keys;
-    std::vector<uint8_t> script;
-
-    BEGIN_SERIALIZE_OBJECT()
-      FIELD(keys)
-      FIELD(script)
-    END_SERIALIZE()
-  };
-
-  struct txout_to_scripthash
-  {
-    crypto::hash hash;
-  };
 
   // outputs <= HF_VERSION_VIEW_TAGS
   struct txout_to_key
@@ -107,34 +88,6 @@ namespace cryptonote
     END_SERIALIZE()
   };
 
-  struct txin_to_script
-  {
-    crypto::hash prev;
-    size_t prevout;
-    std::vector<uint8_t> sigset;
-
-    BEGIN_SERIALIZE_OBJECT()
-      FIELD(prev)
-      VARINT_FIELD(prevout)
-      FIELD(sigset)
-    END_SERIALIZE()
-  };
-
-  struct txin_to_scripthash
-  {
-    crypto::hash prev;
-    size_t prevout;
-    txout_to_script script;
-    std::vector<uint8_t> sigset;
-
-    BEGIN_SERIALIZE_OBJECT()
-      FIELD(prev)
-      VARINT_FIELD(prevout)
-      FIELD(script)
-      FIELD(sigset)
-    END_SERIALIZE()
-  };
-
   struct txin_to_key
   {
     uint64_t amount;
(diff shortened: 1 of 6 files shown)


vtnerd (commented):
Have you run this patch and validated the entire blockchain? Were these types allowed at any point? Could "stuck" funds be sent to them?

jeffro256:
> Have you run this patch and validated the entire blockchain? 

No, not yet, but that's a good idea.

> Were these types allowed at any point?

If you look at the `blockchain.cpp` code, you can see that when visiting these types of `vin/vout` variants, validation always fails, with no exceptions.

> Could "stuck" funds be sent to them?

I believe the tx will fail validation if they are ever attempted to be used.

jeffro256:
You can run this python script on a machine running `monerod` locally to convince yourself that the script/scripthash vin/vout types never appear on-chain:

```Python
import bisect
import json
import random
import requests

my_daemon_address = 'localhost:18081'
timeout_secs = 10

def invoke_json(addr, uri, json_data):
	resp = requests.post("http://" + addr + '/' + uri, json=json_data, timeout=timeout_secs)
	assert(resp.status_code == 200)
	resp_json = resp.json()
	if 'json_rpc' not in uri:
		assert(resp_json['status'] == 'OK')
	return resp_json

def invoke_json_rpc(addr, method, params):
	json_data = {'method': method, 'params': params}
	resp_json = invoke_json(addr, 'json_rpc', json_data)
	result = resp_json.get('result')
	error = resp_json.get('error')
	if error:
		print(error)
		raise ValueError("Error response in JSONRPC")
	return result

def get_height(addr):
	return invoke_json(addr, 'get_height', {})['height']

def get_block_txs_by_height(addr, height):
	result = invoke_json_rpc(addr, 'get_block', { 'height': height })
	block_json = json.loads(result['json'])
	return block_json['miner_tx'], block_json['tx_hashes']

def get_transactions(addr, tx_hashes):
	if not tx_hashes:
		return []
	result = invoke_json(addr, 'get_transactions', { 'txs_hashes': tx_hashes, 'decode_as_json': True, 'prune': True })
	assert(len(result['txs']) == len(tx_hashes))
	return [json.loads(tx['as_json']) for tx in result['txs']]

def extend_vinvout_map(txin_vs, txout_target_vs, tx):
	for txin_v in tx['vin']:
		assert(len(txin_v) == 1)
		vintype = next(iter(txin_v))
		if vintype not in txin_vs:
			txin_vs[vintype] = 0
		txin_vs[vintype] += 1
	for tx_out in tx['vout']:
		txout_target_v = tx_out['target']
		assert(len(txout_target_v) == 1)
		vouttype = next(iter(txout_target_v))
		if vouttype not in txout_target_vs:
			txout_target_vs[vouttype] = 0
		txout_target_vs[vouttype] += 1		

def main():
	top_height = get_height(my_daemon_address)
	txin_vs = {}
	txout_target_vs = {}
	print("Blockchain height:", top_height)
	
	for current_height in range(top_height):
		if current_height % 1000 == 0:
			print(current_height, '/', top_height)
		if current_height % 10000 == 0:
			print("tx in types: ", txin_vs)
			print("tx out types: ", txout_target_vs)
		miner_tx, block_tx_hashes = get_block_txs_by_height(my_daemon_address, current_height)
		block_txs = get_transactions(my_daemon_address, block_tx_hashes)
		extend_vinvout_map(txin_vs, txout_target_vs, miner_tx)
		for block_tx in block_txs:
			extend_vinvout_map(txin_vs, txout_target_vs, block_tx)
	print("Done scanning!")
	print("tx in types: ", txin_vs)
	print("tx out types: ", txout_target_vs)

if __name__ == '__main__':
	main()
```

This scans all txs on the entire chain, including miner transactions, and creates a histogram of all all/output types, outputting the results every 10K blocks.

src/cryptonote_basic/cryptonote_basic.h
```diff
@@ -148,12 +101,19 @@ namespace cryptonote
...
+  struct reserved
+  {
+    BEGIN_SERIALIZE()
+    END_SERIALIZE()
+  };
 
-  typedef boost::variant<txin_gen, txin_to_script, txin_to_scripthash, txin_to_key> txin_v;
+  typedef boost::variant<txin_gen, reserved<0>, reserved<1>, txin_to_key> txin_v;
 
-  typedef boost::variant<txout_to_script, txout_to_scripthash, txout_to_key, txout_to_tagged_key> txout_target_v;
```
jeffro256, line 154: We need to keep deserialzation code of `txout_to_script` around for `wallet2` hot wallet deserialization, because `tx_out` is value-initialized and stored here: https://github.com/monero-project/monero/blob/fc46361376a6f3a604df7ae84ddda848d43dd6cc/src/wallet/wallet2.cpp#L15057 :( Found by @j-berman

PR #8867, opened 2023-05-20
Replace serialization functions

vtnerd:
This replaces all serialization functions (epee binary, epee json, and zmq json), except the custom binary formats for transaction and block. The design is a DOMless vtable interface for reading and writing (2 separate interfaces). The base vtable class can be used to support multiple formats with minimal executable size OR the derived class can be used directly for performance. The design re-uses existing macros, although the newer macros should be lighter on compiler resources.

Memory consumption issues are handled by having all arrays provide a constraint - either a max element count or a min size per element (i.e. `std::vector<std::string>` ideally has a min size of `sizeof(std::string)` per element to prevent memory issues on unpack).

Most importantly, reading epee binary is ~81% faster and writing epee binary is ~56% faster (I re-used the benchmark by @jeffro256 on a ryzen 3 chip). This is primarily due to the removal of the DOM entirely - everything is done on the objects directly.

Unfortunately this diff is huge, and probably has no chance of ever being merged as-is. So I will try to break things into smaller chunks, with the caveat that if nobody wants the serialization to be replaced, its best to vote within this discussion thread here.


vtnerd:
Oops, those performance numbers are from epee binary, the read for json should be less.

vtnerd:
I forgot to document a benefit of this new scheme - anything marked as a `blob`, `epee::byte_slice`, `epee::span<const uint8_t>` or `std::vector<uint8_t>` is marked as binary and automatically converted to hex by the JSON output system. This means many of the manual `std::string` conversions that are performed no longer need to be. Additionally, this allows for epee binary, msgpack, and json to be efficiently by the same code.

selsta:
One thing I noticed during testing is that checking /get_info doesn't work anymore

```
curl 127.0.0.1:18081/get_info
```

```
2023-05-22 09:12:02.656	I HTTP [127.0.0.1] GET /get_info
2023-05-22 09:12:02.656	D Schema expected object
2023-05-22 09:12:02.656	E Failed to parse json: 
```

Is this intentional?

vtnerd:
It is expected behavior given the code, but is probably not desireable. This endpoint should work with an empty JSON object in the body of the http message. I didn't think about these endpoints; JSON-RPC now requires a `params: {}` for empty parameters, but that didn't seem so bad given that JSON-RPC requires additional fields.

I'll work out some sort of fix for this.

vtnerd:
@selsta another update - I don't think this one is easily fixed, unless pay-for-RPC is removed. This endpoint has an optional field (`client`) that is mandatory when pay-for-RPC is enabled. This is currently supported by `start_object()`, then checking for `client` (which can be omitted), and then calling `end_object()`.

It's a shame that such a simple thing could hold up this PR, but this one is difficult to fix without a DOM.

Maybe we re-enable DOM reading for requests but not for responses? It would be a shame to lose an ~81% improvement on these binary reads, but at least the requests are typically small anyway.

selsta:
We only removed it from the wallet side so far.

vtnerd:
Basically the "root" read-function has to do either: `std::is_empty<T>::value && buffer.empty()` OR `is_optional_root<T>::value && buffer.empty()`.

The latter is necessary currently because the `login` field yields `std::is_empty<T>::value == false`. For now, I will do:
```c++
template<typename T>
struct is_optional_root
  : std::is_empty<T>
{};
```
with overloads for a couple of these RPC functions.

vtnerd:
@selsta force pushed the fix I mentioned earlier. I ran your test, and it appears to work. Not all endpoints can be empty, only ones where `std::is_empty<T>::value == true` or where `wire::is_optional_root<T>::value == true`. I could allow _all_ endpoints, which would mean that every endpoint can be called with an empty "body" where all of the values are default-initialized (i.e. basically 0, etc.).

@selsta @UkoeHB thoughts on always allowing an empty root?

jeffro256:
> @selsta another update - assuming requiring a `{}` in the http body as not desirable, can we instead make it a requirement that the body be empty? The problem is DOMless mode has to assume/enforce a particular format (i.e. it cannot assume emptiness and `{}` at same time).

This could be fixed at the server mappers level: for all JSON requests, if the HTTP body is empty, just replace it with the string `{}`. That way we can have a good expected default behavior across all Monero RPC endpoints, but keep the underlying serialization code more strict.

vtnerd:
@jeffro256 not a bad idea, but I prefer the current approach I just force pushed (but I might be persuaded otherwise). Making the string `{}` is JSON specific. Other formats, such as epee_binary, require a binary header, while formats like msgpack require a single binary tag. That approach would also make _any_ type compatible with an empty buffer, whereas this only accepts empty `struct`s and opt-in types.

jeffro256:
> Memory consumption issues are handled by having all arrays provide a constraint - either a max element count or a min size per element (i.e. std::vector<std::string> ideally has a min size of sizeof(std::string) per element to prevent memory issues on unpack).

I might be misunderstanding here, but why would one want to put a minimum element size when deserializing a `std::vector<std::string>`? My system has `sizeof(std::string) == 32`, which would make for quite a large minimum bound. I can see it making sense for arrays of fixed-size POD-types, but is this bound applied for all regular strings?

jeffro256:
If one was expecting to deserialize an array of large blobs (like tx blobs) from a untrusted counterparty I could definitely see the utility in this.

vtnerd:
Rebased after recent changes caused merge conflicts.

vtnerd:
Just did a rebase.

vtnerd:
Another rebase, and updates based on @jeffro256 review (in the other PR #8970 ).

vtnerd:
Decided against `std::size`, and instead added a compile-time check to prevent input iterators to be used with `boost::size`.

contrib/epee/include/serialization/wire/field.h
```diff
@@ -51,7 +51,11 @@ namespace wire
     `value_type` is `T` with optional `std::reference_wrapper` removed.
     `value_type` needs a `read_bytes` function when parsing with a
     `wire::reader` - see `read.h` for more info. `value_type` needs a
+<<<<<<< HEAD
```
iamamyth, line 54: Botched rebase?

vtnerd: Pushed an update to fix this merge issue.

contrib/epee/src/wire/json/read.cpp
```diff
@@ -0,0 +1,452 @@
...
+      case error::schema::number:
+        value.number = i;
+	actual_ = error::schema::number;
+        break;
+      }
+      return true;
+    }
+    bool Uint64(std::uint64_t i) noexcept
+    {
+      switch(expected_)
```
jeffro256, line 138: Is `negative` supposed to be set here?

jeffro256: I guess it doesn't matter if you only use `rapidjson_sax` exactly once

vtnerd: No, because the actual value is not negative.

vtnerd:
Just did a rebase, and changed around some of the array limits. See other review which was just updated too.

vtnerd:
Another rebase with a switch to `std::optional` since we are on C++17 now. Also updated how `params` in `json_rpc` is handled, to match recent bug fixes.

vtnerd:
Another rebase with fixes along the way. Arguably should've been a merge from master, but its too late now.

vtnerd:
This merges changes from `master` into this branch, and includes comments by @thomasbuilds in #8970 . That review will be updated to include the latest changes (probably tomorrow).

contrib/epee/src/wire/json/read.cpp
```diff
@@ -0,0 +1,452 @@
...
+    bool EndArray(std::size_t) const noexcept { return expected_ == error::schema::none; }
+    bool StartObject() const noexcept { return expected_ == error::schema::none; }
+    bool EndObject(std::size_t) const noexcept { return expected_ == error::schema::none; }
+  };
+
+  void json_reader::read_next_value(rapidjson_sax& handler)
+  {
+    rapidjson::MemoryStream stream{reinterpret_cast<const char*>(remaining_.data()), remaining_.size()};
+    rapidjson::EncodedInputStream<rapidjson::UTF8<>, rapidjson::MemoryStream> istream{stream};
+    if (!reader_.Parse<rapidjson::kParseStopWhenDoneFlag>(istream, handler))
```
thomasbuilds: This parses recursively with no depth limit, and `skip_value()` sends unknown fields through it. One anonymous `POST /get_info` with a 350 KB body of the form `{"x":[[[...]]]}` (depth 175000) segfaults the daemon; master rejects the same body cleanly. The nesting has to sit in a field value, a bare `[[[...]]]` body is rejected by `do_start_object` first.

```suggestion
    if (!reader_.Parse<rapidjson::kParseStopWhenDoneFlag | rapidjson::kParseIterativeFlag>(istream, handler))
```

Against the vendored rapidjson, depth 200000 segfaults as written and 2000000 parses cleanly with the flag, successive `Parse` calls still stopping after each value. The iterative parser accepts any depth rather than rejecting it, so `max_read_depth()` on the skip path is still worth adding.

src/rpc/zmq_server.cpp
```diff
@@ -175,11 +175,16 @@ void ZmqServer::serve()
...
+          expect<epee::byte_slice> response = handler.handle(std::move(*message));
+
+          if (response)
+          {
+            const boost::string_ref response_view{reinterpret_cast<const char*>(response->data()), response->size()};
+            MDEBUG("Sending RPC reply: \"" << response_view << "\"");
+            MONERO_UNWRAP(net::zmq::send(std::move(*response), rep.get()));
+          }
+          else
+            MERROR("Failed to handle RPC message: " << response.error().message());
```
thomasbuilds: This path sends no reply on a REP socket, so the next `receive` fails with EFSM, `MONERO_THROW` propagates past the loop, and `serve()` exits, leaving ZMQ-RPC dead for all clients until restart. `{"jsonrpc":"2.0","method":"get_height"}` reaches it: `handle_message` returns a bare error code because `params` is absent, so nothing is sent. (Dropping `jsonrpc` instead fails earlier in `request_generic` and does get a `BAD_JSON` reply, so the request has to be well formed to hit this.)

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -260,26 +244,20 @@ inline const std::string get_rpc_status(const bool trusted_daemon, const std::st
...
-        {
-          KV_SERIALIZE(added_pool_txs)
-          KV_SERIALIZE_CONTAINER_POD_AS_BLOB(remaining_added_pool_txids)
-        }
-        if (pool_info_extent == POOL_INFO_EXTENT::INCREMENTAL)
-        {
-          KV_SERIALIZE_CONTAINER_POD_AS_BLOB(removed_pool_txids)
-        }
+        KV_SERIALIZE_ARRAY(added_pool_txs, min_pool_tx_info) // optional if empty
+        KV_SERIALIZE_CONTAINER_POD_AS_BLOB(removed_pool_txids) // optional if empty
```
thomasbuilds, line 252: `remaining_added_pool_txids` is missing from this map. It is still declared at :242, still filled in `core_rpc_server.cpp:688`, and still read by wallet2 at `wallet2.cpp:3144-3161`, so those mempool transactions never reach the wallet.

```suggestion
        KV_SERIALIZE_ARRAY(added_pool_txs, min_pool_tx_info) // optional if empty
        KV_SERIALIZE_CONTAINER_POD_AS_BLOB(remaining_added_pool_txids) // optional if empty
        KV_SERIALIZE_CONTAINER_POD_AS_BLOB(removed_pool_txids) // optional if empty
```

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -432,18 +395,22 @@ inline const std::string get_rpc_status(const bool trusted_daemon, const std::st
...
     };
     typedef epee::misc_utils::struct_init<response_t> response;
   };
+  WIRE_JSON_DECLARE_OBJECT(COMMAND_RPC_GET_TRANSACTIONS::entry);
+  WIRE_JSON_DECLARE_COMMAND(COMMAND_RPC_GET_TRANSACTIONS);
 
   //-----------------------------------------------
   struct COMMAND_RPC_IS_KEY_IMAGE_SPENT
   {
+    using max_request_size = wire::max_element_count<4096>;
```
thomasbuilds: `wallet2::import_key_images` (:13199) sends every signed key image in one request (:13261) and asserts the response size matches (:13263), so a wallet with more than 4096 outputs fails to import: 4096 is accepted, 4097 and 5000 return 400, master accepts 5000. `rescan_spent` chunks at 1000.

src/rpc/core_rpc_server_commands_defs.cpp
```diff
@@ -0,0 +1,314 @@
...
+    else if (binary && bool(binary->second) == dest.binary && !dest.compress)
+      dest.data.distribution = std::move(binary->first);
+    else
+      dest.data.distribution.clear();
+  }
+  template<typename W>
+  static void write_bytes(W& dest, const COMMAND_RPC_GET_OUTPUT_DISTRIBUTION::distribution& source)
+  {
+    if (!std::is_same<W, wire::epee_writer>() && (source.binary || source.compress))
+      WIRE_DLOG_THROW(wire::error::schema::binary, "Compressed and binary data only supported with epee format");
```
thomasbuilds, line 258: `binary` defaults to true, so `get_output_distribution` over JSON-RPC reaches this throw mid-write and returns 200 with an empty body. `binary:false` works.

src/cryptonote_basic/cryptonote_basic.h
```diff
@@ -53,8 +55,12 @@
 
 namespace cryptonote
 {
-  typedef std::vector<crypto::signature> ring_signature;
+  using max_inputs_per_tx = wire::max_element_count<WIRE_MAX_INPUTS_PER_TX>;
+  using max_outputs_per_tx = wire::max_element_count<WIRE_MAX_OUTPUTS_PER_TX>;
 
+  typedef std::vector<crypto::signature> ring_signature;
+  using block_blob_min = wire::min_element_size<73>;
+  using tx_blob_min = wire::min_element_size<41>;
```
thomasbuilds, line 62: `do_start_array` charges `count * min_element_size` against the remaining buffer while each blob costs `1 + length` bytes on the wire, so `<41>` enforces a 40-byte payload minimum. `too_small_txs_blobs` and `too_small_compact_txs_blobs` build 40-byte blobs and expect a decode error, so both fail on all three test jobs, while the `smallest_*` pair at 41 passes. Raising the constant to 42 would restore the intended boundary, but those two sit close to the limit, the compact one also paying the outer array's charge against `array_space_`, so they want re-running with any new value.

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -2590,10 +2585,88 @@ inline const std::string get_rpc_status(const bool trusted_daemon, const std::st
...
   };
 
-}
+} // cryptonote
+
+namespace wire
+{
+  // these endpoints are permitted to have an empty read buffer
+
+  WIRE_DECLARE_OPTIONAL_ROOT(cryptonote::COMMAND_RPC_GET_HEIGHT::request);
```
thomasbuilds: Incomplete list: empty bodies now 400 on 7 of 16 direct endpoints master accepted (`/get_transactions`, `/is_key_image_spent`, `/send_raw_transaction`, `/get_public_nodes`, `/get_transaction_pool`, `/get_outs` and aliases). Related, if intended: fields that were optional by omission under `KV_SERIALIZE` are now required `WIRE_FIELD`s throwing `MISSING_KEY`, so `hard_fork_info` without `version` returns -32602 where master succeeded, which the PR's python-rpc edits work around.

thomasbuilds (changes requested):
needs rebase
following points are AI assisted by probably_wrong (matrix)

contrib/epee/src/wire/json/read.cpp
vtnerd: Yes this was a major oversight.

src/rpc/core_rpc_server_commands_defs.h
vtnerd: Not sure how this got dropped, but added.

src/cryptonote_basic/cryptonote_basic.h
vtnerd: This is tricky because it's technically an artifact of epee encoding. The `min_element_size` value is intended to be the smallest possible size for any format. I guess this could include the length of the blob too, but I need to think about it a little more.

src/rpc/zmq_server.cpp
vtnerd: Yes, I think it was a mistake to change the return value to `expect<byte_slice>` and instead will revert back to `byte_slice`.

vtnerd: Changed my mind. This still returns `expect<byte_slice>` except now it only occurs when serializing error message to JSON failed. This should be extremely rare.

The caller now does a `MONERO_UNWRAP()` as it shouldn't be possible to trigger this remotely now.

src/rpc/core_rpc_server_commands_defs.cpp
vtnerd: This was intentional, and I'm going to leave unchanged unless there is a decent argument to change it. If I change the default behavior, this breaks tons of existing code that assumes this flag defaults to `true`.

The `binary` field already defaults to `true`, and when sent via JSON request it returns raw binary to caller without escaping or encoding the value at all. I doubt any existing JSON clients call this function without including `binary: false` because it cannot be interpreted properly.

Instead of invalid JSON, this should now return an error to the caller. The negative is that the user probably needs to read the `monerod` logs to figure out why the request failed.

src/rpc/core_rpc_server_commands_defs.h
vtnerd: Good catch, I should have this fixed in the next update.

If `sizeof(int) > 4` the code will throw an annoying `static_assert` to prevent memory explosion, but this is unlikely as I'm not aware of any platform where this is true.

src/cryptonote_basic/cryptonote_basic.h
vtnerd: I'm not changing this - at least for now. The problem is that the `size<...>` is supposed to be format agnostic, but practically all formats will require at least a one-byte size value (as you mention). Anyway, leaving this unresolved, there's still much work to be done on the "read interface" review.

contrib/epee/include/serialization/keyvalue_serialization.h
```diff
@@ -26,103 +26,82 @@
...
-  epee::serialization::selector<is_store>::serialize_t_val_as_blob(this_ref.varialble, stg, hparent_section, val_name); 
+  std::make_tuple(::wire::optional_field(val_name, ::wire::defaulted(std::ref(self.variable), default_value))),
 
 #define KV_SERIALIZE_VAL_POD_AS_BLOB_N(variable, val_name) \
-  static_assert(std::is_trivially_copyable_v<decltype(this_ref.variable)>, "t_type must be a trivially copyable type."); \
-  static_assert(std::is_standard_layout_v<decltype(this_ref.variable)>, "t_type must be a standard layout type."); \
-  KV_SERIALIZE_VAL_POD_AS_BLOB_FORCE_N(variable, val_name)
+  std::make_tuple(::wire::field(val_name, ::wire::blob(std::ref(self.variable)))),
+
+#define KV_SERIALIZE_CONTAINER_POD_AS_BLOB_N(variable, val_name) \
```
selsta, line 87: This now produces hex, previously it produced raw bytes inside an escaped JSON string. This affects `get_transaction_pool_hashes.bin` and `get_txpool_backlog`

contrib/epee/src/wire/epee/read.cpp
```diff
@@ -0,0 +1,375 @@
...
+    case SERIALIZE_TYPE_UINT16:
+    case SERIALIZE_TYPE_UINT8:
+      return {unsigned_integer()};
+    default:
+      break;
+    };
+    WIRE_DLOG_THROW(error::schema::number, "expected a boolean, integer, float or string");
+  }
+
+  bool epee_reader::boolean()
```
selsta, line 280: The previous reader explicitly read a `uint8_t`, rejected values above 1, and then converted to bool. This one accepts invalid booleans representations.

contrib/epee/include/serialization/keyvalue_serialization.h
```diff
@@ -26,103 +26,82 @@
...
+                                                                        \
+  template<class W>                                                     \
+  void write_bytes(W& dest) const                                     \
+  { ::epee::unpack_object(std::false_type{}, dest, get_field_list(dest, *this)); } \
+                                                                        \
+  template<class F, class T>                                            \
+  static auto get_field_list(F& format, T& self)                        \
+  { return std::tuple_cat(
+
+#define KV_SERIALIZE_N(variable, val_name) \
```
selsta, line 75: This now rejects supported alternative RPC selectors. For example `get_block` with params `{"height":0}` now fails because hash is absent.

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -388,33 +373,12 @@ inline const std::string get_rpc_status(const bool trusted_daemon, const std::st
...
-        {
-          KV_SERIALIZE(relayed)
-          KV_SERIALIZE(received_timestamp)
-        }
-      END_KV_SERIALIZE_MAP()
     };
 
     struct response_t: public rpc_access_response_base
     {
+      using max_count = wire::max_element_count<1024>;
```
selsta, line 380: This can now break wallet operations, like for example `/gettransactions`.

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -1351,6 +1346,8 @@ inline const std::string get_rpc_status(const bool trusted_daemon, const std::st
 
   struct COMMAND_RPC_GET_PEER_LIST
   {
+    using max_peer_list = wire::max_element_count<4096>;
```
selsta, line 1349: Existing gray list limit is 5000, so this can break things.

src/rpc/message_data_structs.cpp
```diff
@@ -0,0 +1,253 @@
...
+    template<typename F, typename T>
+    void tx_in_pool_map(F& format, T& self)
+    {
+      wire::object(format,
+        WIRE_FIELD(tx),
+        WIRE_FIELD(tx_hash),
+        WIRE_FIELD(blob_size),
+        WIRE_FIELD(weight),
+        WIRE_FIELD(fee),
+        WIRE_FIELD(max_used_block_hash),
```
selsta, line 99: `max_used_block_height` is dropped here from ZMQ pool responses. Previous JSON reader required it.

contrib/epee/include/serialization/keyvalue_serialization.h
vtnerd: This was intentional on my part, although this may frustrate downstream users.

AFAIK, this outputs invalid UTF-8 sequences, which is against the JSON spec. I figured this breaking change would be worth it, as currently some parsers likely reject the responses from those two endpoints.

Not sure how to proceed.

contrib/epee/src/wire/epee/read.cpp
vtnerd: This is more permissive, but it doesn't seem to matter much. Values above one now get merged to `true`, which isn't terrible. But I'll retain the logic.

contrib/epee/include/serialization/keyvalue_serialization.h
vtnerd: Yes, the existing macros/design is a terror in sloppiness of not specifying which fields were required and which ones were optional. You basically had to look at the handler code afaik (as compared to looking at the definition of the conversion function).

I can go through each endpoint to fix this properly, or do some hack to make it "work" as it did previously. I'll probably do the latter as its going to take too long to inspect each endpoint properly.

src/rpc/core_rpc_server_commands_defs.h
vtnerd: I'm not sure why I did this when I could use the min wire size trick. I'll just fix this (although both may be disappearing at the next hard fork anyway).

contrib/epee/include/serialization/wire/adapted/static_vector.h
```diff
@@ -0,0 +1,91 @@
...
+  template<typename R, typename T, std::size_t N>
+  inline void read_bytes(R& source, boost::container::static_vector<T, N>& dest)
+  {
+    wire_read::array(source, dest, min_element_size<0>{}, max_element_count<N>{});
+  }
+
+  /* `static_vector` never allocates, so it is useful for reading small variable
+     sized strings or binary data with a known fixed max. `char` and
+     `std::uint8_t` are not valid types for arrays in this design anyway
+     (because its clearly less efficient in every encoding scheme).  */
```
jpk68, line 64: Nit: typo (should be "it's")

contrib/epee/include/serialization/wire/json/base.h
```diff
@@ -79,14 +78,14 @@ namespace wire
...
 
-    template<typename T>
-    static std::error_code to_bytes(epee::byte_stream& dest, const T& source)
-    {
-      std::string buf{};
-      const std::error_code error = to_bytes(buf, source);
-      if (!error)
-        dest.write(epee::to_span(buf));
-      return error;
+    template<typename... T>
```
jpk68, line 81: Nit: is there a reason to add all of these spaces at the end of the line?

PR #8868, opened 2023-05-21
Write interface for new serialization system

vtnerd:
This PR contains the write serialization interface, and some glue code from #8867 . This does **not** replace any existing serialization; the goal is to start a discussion on the design of the write system.

The write system - for epee binary - is roughly 56% faster than the existing design. This is primarily due to the removal of the intermediate DOM building stage - everything is written directly from objects to the "sink". I will work on JSON output numbers later.

The code for epee binary writing and json writing is not included, that can be reviewed in a future PR.

### DIFF
--- a/contrib/epee/include/serialization/wire/field.h
+++ b/contrib/epee/include/serialization/wire/field.h
@@ -1,0 +1,141 @@
+// Copyright (c) 2021-2023, The Monero Project
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <functional>
+#include <utility>
+
+#include "serialization/wire/traits.h"
+
+//! A required field has the same key name and C/C++ name
+#define WIRE_FIELD(name)                                \
+  ::wire::field( #name , std::ref( self . name ))
+
+//! A required field has the same key name and C/C++ name AND is cheap to copy (faster output).
+#define WIRE_FIELD_COPY(name)                   \
+  ::wire::field( #name , self . name )
+
+//! The optional field has the same key name and C/C++ name
+#define WIRE_OPTIONAL_FIELD(name)                               \
+  ::wire::optional_field( #name , std::ref( self . name ))
+
+namespace wire
+{
+  /*! Links `name` to a `value` for object serialization.
+
+    `value_type` is `T` with optional `std::reference_wrapper` removed.
+    `value_type` needs a `read_bytes` function when parsing with a
+    `wire::reader` - see `read.h` for more info. `value_type` needs a
+    `write_bytes` function when writing with a `wire::writer` - see `write.h`
+    for more info.
+
+    Any `value_type` where `is_optional_on_empty<value_type> == true`, will
+    automatically be converted to an optional field iff `value_type` has an
+    `empty()` method that returns `true`. The old output engine omitted fields
+    when an array was empty, and the standard input macro would ignore the
+    `false` return for the missing field. For compability reasons, the
+    input/output engine here matches that behavior. See `wrapper/array.h` to
+    enforce a required field even when the array is empty or specialize the
+    `is_optional_on_empty` trait. Only new fields should use this behavior.
+
+    Additional concept requirements for `value_type` when `Required == false`:
+      * must have an `operator*()` function.
+      * must have a conversion to bool function that returns true when
+        `operator*()` is safe to call (and implicitly when the associated field
+        should be written as opposed to skipped/omitted).
+    Additional concept requirements for `value_type` when `Required == false`
+    when reading:
+      * must have an `emplace()` method that ensures `operator*()` is safe to call.
+      * must have a `reset()` method to indicate a field was skipped/omitted.
+
+    If a standard type needs custom serialization, one "trick":
+    ```
+    struct custom_tag{};
+    void read_bytes(wire::reader&, boost::fusion::pair<custom_tag, std::string&>)
+    { ... }
+    void write_bytes(wire::writer&, boost::fusion::pair<custom_tag, const std::string&>)
+    { ... }
+
+    template<typename F, typename T>
+    void object_map(F& format, T& self)
+    {
+      wire::object(format,
+        wire::field("foo", boost::fusion::make_pair<custom_tag>(std::ref(self.foo)))
+      );
+    }
+    ```
+
+    Basically each input/output format needs a unique type so that the compiler
+    knows how to "dispatch" the read/write calls. */
+  template<typename T, bool Required>
+  struct field_
+  {
+    using value_type = unwrap_reference_t<T>;
+
+    //! \return True if field is forced optional when `get_value().empty()`.
+    static constexpr bool optional_on_empty() noexcept
+    { return is_optional_on_empty<value_type>::value; }
+
+    static constexpr bool is_required() noexcept { return Required && !optional_on_empty(); }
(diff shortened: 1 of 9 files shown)


contrib/epee/include/serialization/wire/write.h
```diff
@@ -0,0 +1,290 @@
...
+    //! By default, insist on retrieving array size before writing array
+    static constexpr std::true_type need_array_size() noexcept { return{}; }
+
+    virtual void boolean(bool) = 0;
+
+    virtual void integer(int) = 0;
+    virtual void integer(std::intmax_t) = 0;
+
+    virtual void unsigned_integer(unsigned) = 0;
+    virtual void unsigned_integer(std::uintmax_t) = 0;
```
jeffro256, line 74: Is there any guarantee that `int` and `std::intmax_t` are different types on every system? Also, I haven't run benchmarks, but would it be a worthy tradeoff to crunch each `integer` and `unsigned_integer` virtual methods into one each? In other words, is the expanded vtables / extra binary size / worse caching worth not having to cast the integers?

vtnerd: The only advantage is 32-bit systems. Theoretically some platform could use 128-bit integers on a 64-bit cpu (making the overload useful again), but that doesn't seem likely. It might be worth dropping to simplify things a bit more.

And I don't think its possible for `int` and `std::intmax_t` to have an issue, because both `long` and `long long` are required by the standard, and `std::intmax_t` must be the largest possible type on the platform. Even if they are all 32-bit, they should be unique types (afaik).

Of course, if I drop the overload entirely then it doesn't matter.

jeffro256: I think from a semantics point of view, since there is no difference between these two methods anyways that the overloads should be taken away IMO; it simplifies the interface. If there was a difference (see `serde`), then it would make sense to keep them in

contrib/epee/include/serialization/wire/field.h
```diff
@@ -0,0 +1,141 @@
...
+    for more info.
+
+    Any `value_type` where `is_optional_on_empty<value_type> == true`, will
+    automatically be converted to an optional field iff `value_type` has an
+    `empty()` method that returns `true`. The old output engine omitted fields
+    when an array was empty, and the standard input macro would ignore the
+    `false` return for the missing field. For compability reasons, the
+    input/output engine here matches that behavior. See `wrapper/array.h` to
+    enforce a required field even when the array is empty or specialize the
+    `is_optional_on_empty` trait. Only new fields should use this behavior.
```
jeffro256, line 64: Why would we ever want to enforce a required field when the array is empty in the current codebase?

vtnerd: I'm not sure what you mean - when would want to make an empty array required? The only time would be if some external project required it for some reason. There's basically no code needed to support this feature, in fact supporting the empty-array-is-optional case is more code.

contrib/epee/include/serialization/wire/write.h
vtnerd: I removed the `int` and `unsigned` overloads, only the `intmax_t` and `uintmax_t` remain.

vtnerd: The unit tests just reminded me of one issue with removing the `unsigned` and `int` - when writing "through the base class" epee will _always_ write 64-bit integers. This can be prevented by checking the value and compressing the output (like my message pack implementation); the split previously allowed for a quick way to compress the types.

I think I'm going to make a change where the epee class always compresses the integer to the smallest possible (like message pack does now).

vtnerd: Nevermind, I remember why I cannot. MessagePack has a tag per array item, allowing you to compress individual integers, but epee does not. So if the base-class is used for writing epee data, then it will _always_ use 64-bit integers for outputs.

jeffro256: Also some binary formats (like bincode) have fixed size integer types. We might want to, in the future, add specific methods for this. Think like Rust's serde's `Serializer` interface with `serialize_u8`, `serialize_i16`, `serialize_u32`, `serialize_u64`, etc. By default, for most formats, these methods could feed directly into `integer()` and `unsigned_integer()`. But I guess we shouldn't over-engineer from the outset.

vtnerd: I'm not sure what you mean here - but MessagePack and Epee both have fixed integer types. Neither has variable size actually (whereas Protobuf does). That's not the issue. The problem is _every integer comes with its own tag_ for size with MessagePack, whereas with Epee there is one-fixed tag for every integer in an array.

I already wrote the Epee writer to serialize the exact fixed size when using the derived class (like serde), but it now fallsback to uint64 when using the base class. You can overload the write function multiple times if its a problem, so I don't see a need to engineer this further.

The MessagePack implement just compresses the value per integer, which is what most of the MessagePack serializers do.

jeffro256: What I mean is that in both of those cases, Epee and MessagePack, those formats are *self-describing*. You can determine the size of integers at runtime. Because of that property, you can be a little more willy-nilly casting integers to and forth into the format, and it isn't actually that important that we provide methods for every fixed size of integer. By contrast, in bincode, for example, this serialization information is assumed to be known beforehand, which saves space and computation, but if you don't know what size to deserialize an integer as, then reading from that point on might just give you garbage.

In MessagePack, since this tag is per-element (I'm assuming what you've told me I don't actually know anything about the Messagepack format), then you don't have to actually call a special `integer` method because the `writer` in this case could just assume you always want to compress if the value is small enough. And, likewise, since the `reader` can see the tags before deserializing the integers, we only need 2 methods for integers: signed and unsigned, and just cast everything accordingly.

I think it's fine like it is now with just `integer()/unsigned_integer()`, but if we wanted to support formats which use multiple sizes of integers but don't encode that integer size information in the data, then we would need more expressive interface with methods for each fixed integer size.

vtnerd: Yes, this is why I replaced everything but the internal crypto note serialization format.

contrib/epee/include/serialization/wire/write.h
```diff
@@ -0,0 +1,287 @@
...
+namespace wire
+{
+  template<typename W, typename T>
+  inline std::enable_if_t<is_array<T>::value> write_bytes(W& dest, const T& source)
+  {
+    wire_write::array(dest, source);
+  }
+
+  template<typename W, typename... T>
+  inline std::enable_if_t<std::is_base_of<writer, W>::value> object(W& dest, T... fields)
```
jeffro256, line 277: Why is the `enable_if_t` needed here?

jeffro256 (commented):
I'm getting very close to approval, just a couple of things left

contrib/epee/include/serialization/wire/write.h
```diff
@@ -0,0 +1,287 @@
...
+  inline void write_bytes(W& dest, const epee::byte_slice& source)
+  { write_bytes(dest, epee::to_span(source)); }
+
+  //! Use `write_bytes(...)` method if available for `T`.
+  template<typename W, typename T>
+  inline auto write_bytes(W& dest, const T& source) -> decltype(source.write_bytes(dest))
+  { return source.write_bytes(dest); }
+}
+
+namespace wire_write
```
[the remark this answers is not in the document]
vtnerd: If a `write_bytes` is added to the `::wire::` namespace then this prevents ADL lookup. There are many such overloads, because any type that is not "owned" by our repo has to go into the `::wire::` namespace.

So you need a "clean" namespace (including global) of any `write_bytes` calls.

contrib/epee/include/serialization/wire/field.h
```diff
@@ -0,0 +1,141 @@
...
+    void object_map(F& format, T& self)
+    {
+      wire::object(format,
+        wire::field("foo", boost::fusion::make_pair<custom_tag>(std::ref(self.foo)))
+      );
+    }
+    ```
+
+    Basically each input/output format needs a unique type so that the compiler
+    knows how to "dispatch" the read/write calls. */
```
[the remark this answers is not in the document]
vtnerd: I'm not sure what you mean here. You think specializations should be even easier than creating a wrapper type? I've done this in numerous places, and it didn't seem that bad in practice. Typically, I use a custom `struct` instead of this `boost::fusion::pair` object.

The other alternative is skipping usage of the `wire::object(...)` and instead call `start_object()` and `end_object()`, etc., directly. Its just that typically you'll need to handle out-of-order fields, and `wire::object(...)` does this for you.

contrib/epee/include/serialization/wire/write.h
```diff
@@ -0,0 +1,287 @@
...
+  { dest.unsigned_integer(std::uintmax_t(source));}
+
+  template<typename W>
+  inline void write_arithmetic(W& dest, const double source)
+  { dest.real(source); }
+
+  // Template both arguments to allow derived writer specializations
+  template<typename W, typename T>
+  inline std::enable_if_t<std::is_arithmetic<T>::value> write_bytes(W& dest, const T source)
+  { write_arithmetic(dest, source); }
```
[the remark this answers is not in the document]
vtnerd: The problem is `bool` gets accepted for one of those types. There's probably a SFINAE method to fix that too, but I didn't bother coming up with it.

contrib/epee/include/serialization/wire/write.h
vtnerd: There's an overload in `read.h` for the `wire::reader` case. I could omit this until the reader gets merged, but I think it's worth keeping in.

The overload allows for easy cpp usage - the `write_bytes` and `read_bytes` can forward to a single templated "map" function that calls `wire::object(...)` once for both the read and write case. i.e.

```c++
struct foo { int x; int y; }
void write_bytes(wire::writer&, const foo&);
void read_bytes(wire::reader&, foo&);

// .. cpp
template<typename F, typename T>
static void foo_map(F& format, T& self)
{ wire::object(format, WIRE_FIELD(x), WIRE_FIELD(y)); }

void write_bytes(wire::writer& dest, const foo& source) { foo_map(dest, source); }
void read_bytes(wire::reader& source, foo& dest) { foo_map(source, dest); } 
```

The boilerplate in the header and cpp can also further by macro'ed to make this a bit easier. This retains the "specify" once of the current method, while allowing for cpp files to reduce header bloat.

vtnerd: And ! I forgot to mention why not `wire::object(reader& .. )` - templating allows forwarding of the derived type for performance reasons. Many of the calls within our codebase only actually use one type (p2p only uses epee, etc.), so specifying the derived makes more sense.

jeffro256 (approved):

contrib/epee/include/serialization/wire/write.h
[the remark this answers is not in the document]
vtnerd: I misspoke, I did this in the read interface, not sure why I didn't do it here. So this will be updated, along with the namespace recommendation and some squashing.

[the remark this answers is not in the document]
vtnerd: Nope, still getting the ambiguity errors, despite it being nearly identical to the read code. I'm not wasting further time on this, maybe in a future PR this can get cleaned up a bit.

vtnerd:
I just force pushed a squashing commit, hopefully this won't be too hard to re-review.

selsta (approved):

PR #8870, opened 2023-05-21
Add to_hex::buffer

vtnerd:
Adds `to_hex::buffer` which is used by #8867 . This prevents allocations in some use cases.

### DIFF
--- a/contrib/epee/include/hex.h
+++ b/contrib/epee/include/hex.h
@@ -67,6 +67,9 @@ namespace epee
       return out;
     }
 
+    //! Write `src` as hex to `out`. `out` must be exactly 2x in size.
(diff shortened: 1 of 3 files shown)


contrib/epee/include/hex.h
```diff
@@ -67,6 +67,9 @@ namespace epee
       return out;
     }
 
+    //! Write `src` as has to `out`. `out` must be exactly 2x in size.
```
iamamyth, line 70: Appears to be a typo in the comment (`has` should be `hex`).

vtnerd: Yup, force pushed a correction.

jeffro256 (approved):

iamamyth (approved):

PR #8876, opened 2023-05-25
Update blockchain_stats.cpp [Potential Memory Leak]

RandomGuy199:
Potential Memory Leak: The db object is dynamically allocated using the new_db() function, but there is no corresponding delete or delete[] statement to release the memory. This can result in a memory leak.


selsta:
This patch is incorrect, there's no point in doing delete after a return statement. Also there's no memory leak here because the OS frees the memory anyway on program exit.

PR #8884, opened 2023-05-31
fix wallet2 dropping some errors on refresh

moneromooo-monero:


jeffro256:
Why not just `throw` to rethrow the error? It'd be clearer if using a debugger

moneromooo-monero:
It just did not occur to me to do it differently here. It could be done.

jeffro256 (approved):
I'm fine with it tho

PR #8890, opened 2023-06-03
core_rpc_server: return ID of submitted block

jeffro256:
Closes https://github.com/monero-project/monero/issues/8874

PR #8891, opened 2023-06-03
core_rpc_server: return ID of submitted block [RELEASE]

jeffro256:
#8890


selsta (approved):
This isn't really relevant for release branch but since the next release will be v0.18.3.0 I guess we can merge it. It will also contain a couple other new features.

PR #8892, opened 2023-06-03
wallet_rpc_server: dedup transfer RPC responses [RELEASE]

jeffro256:
#8888


vtnerd:
Does this need to be done against release? Its relatively innocuous, but its also not a bug fix.

jeffro256:
I plan to open the equivalent of #8861 against release which would have this PR as a dependency.

selsta (approved):

PR #8893, opened 2023-06-05
cryptonote_basic: speedup find_tx_extra_field_by_type

jeffro256:
The savings are pretty small per call, but it gets called multiple times per output per transaction per block per refresh, so it's probably worth it.


moneromooo-monero:
But the code doesn't do at all what the original code does.
I *think* the new code returns the Nth field if it is of the given type. The current code returns the Nth field of the given type.

jeffro256:
Ah damn, good point, I didn't notice the shortcut.

PR #8894, opened 2023-06-08
Speed up perf_timer init on x86

SChernykh:
All Monero binaries have 1 second startup delay because of this code. This is especially noticeable and affects UX in Monero GUI wallet with local node where it often starts another monerod instance to run commands and query node status.

### DIFF
--- a/src/common/perf_timer.cpp
+++ b/src/common/perf_timer.cpp
@@ -62,7 +62,7 @@ namespace tools
     while (1)
     {
       t1 = epee::misc_utils::get_ns_count();
-      if (t1 - t0 > 1*1000000000) break; // work one second
+      if (t1 - t0 > 1*100000000) break; // work 0.1 seconds
     }
 
     uint64_t r1 = get_tick_count();


plowsof:
confirming that the GUI benefits from the speed up. i added a `Qt.exit()` to the initial "add a desktop entry?" page and parsed the logs for time stamps. there is a clear reduction from 3~seconds to 2~ seconds.

tobtoht:
TSC is not a reliable way to measure time. https://stackoverflow.com/a/33473175

I think we can completely remove the x86_64 specific code here and fall back on `epee::misc_utils::get_ns_count()`.

SChernykh:
It wasn't reliable on old systems, but on 99% modern systems it's a fixed speed clock (usually running at base CPU freq), and it's synchronized between cores. Anyway, I'm not removing it because I didn't add it. This PR is the least invasive fix for the issue.

hinto-janai (approved):
```
time ./monerod --help
real	0m1.004s
user	0m1.004s
sys	0m0.000s

time speedup_tsc_init/release/bin/monerod --help
real	0m0.110s
user	0m0.107s
sys	0m0.000s
```

selsta (approved):

plowsof (approved):

PR #8896, opened 2023-06-09
blockchain_db/core_rpc_server: Get coinbase RCT output distribution

jeffro256:
This changes edits `BlockchainLMDB` to add an extra field to the block info structs, and updates the `get_output_distribution` RPC API to allow wallets to obtain the distribution of RCT coinbase outputs per block. This change requires a database migration and will add ~2MB to the database every year.

How to request the information:
1. In the `get_output_distribution[.bin]` RPC command, set a boolean field `get_rct_coinbase` to `true`.
2. The RCT coinbase distribution will be returned in the response field `rct_coinbase_distribution`.

This change will enable extremely fast fetching of the RCT coinbase distribution, which will let wallets efficiently perform decoy selection across only coinbase outputs or only non-coinbase outputs. This PR by itself does **not** affect decoy selection.

### DIFF
--- a/src/blockchain_db/blockchain_db.cpp
+++ b/src/blockchain_db/blockchain_db.cpp
@@ -274,6 +274,7 @@ uint64_t BlockchainDB::add_block( const std::pair<block, blobdata>& blck
   time1 = epee::misc_utils::get_tick_count();
 
   uint64_t num_rct_outs = 0;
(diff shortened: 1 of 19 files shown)


jeffro256:
Prereq for #8815

jeffro256:
FCMP++ makes this obsolete

PR #8899, opened 2023-06-09
Set SSL SNI even when server verification is disabled

vtnerd:
This was found while running LWS, but I don't see a reason to _not_ fix this upstream.

Someone was trying a connection to a cloudflare site (webhooks) - which requires SSL SNI usage to work correctly - and combined it with a call to **not** verify the cloudflare server. It took me a while to track down the issue: we only set SNI when verifying the server certificate. There's no reason for this, it can be used with or without verifying the certificates.

### DIFF
--- a/contrib/epee/src/net_ssl.cpp
+++ b/contrib/epee/src/net_ssl.cpp
@@ -497,6 +497,13 @@ void ssl_options_t::configure(
   const std::string& host) const
 {
   socket.next_layer().set_option(boost::asio::ip::tcp::no_delay(true));
+  {
+    // in case server is doing "virtual" domains, set hostname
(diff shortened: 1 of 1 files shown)


contrib/epee/src/net_ssl.cpp
```diff
@@ -497,6 +497,13 @@ void ssl_options_t::configure(
   const std::string& host) const
 {
   socket.next_layer().set_option(boost::asio::ip::tcp::no_delay(true));
+  {
```
jeffro256, line 500: I don't like the superfluous braces, but besides that, looks good to me!

jeffro256 (approved):

PR #8913, opened 2023-06-21
blockchain_db: add clarification to get_block_already_generated_coins

jeffro256:


vtnerd (approved):

src/blockchain_db/blockchain_db.h
```diff
@@ -1053,7 +1053,7 @@ class BlockchainDB
    * @brief fetch a block's already generated coins
    *
    * The subclass should return the total coins generated as of the block
-   * with the given height.
+   * with the given height, capped to a maximum value of MONEY_SUPPLY.
```
hyc, line 1056: if that's so, what's the point of this function now, since we're in tail emission, well beyond the original money supply?

jeffro256: That's a good point... I looked into it, and it happens that  `get_block_already_generated_coins()` is only ever fetched for the very top block. And since we are past tail emission, we don't need those values anyways and can hardcode a value of coins which gives us the same result for block reward calculations.

jeffro256: (maybe) I'll do some testing for that, since I wanted to do a migration anyways to add a field to `mdb_block_info` anyways, this could make it 0-cost ;)

moneromooo-monero: Removing it should preclude syncs without fast block sync.

PR #8914, opened 2023-06-22
wallet-rpc: restore from multisig seed (and fix generation)

jeffro256:
If the field `enable_multisig_experimental` in the command `restore_deterministic_wallet` is set to `true`, `monero-wallet-rpc` will attempt to generate the wallet from a multisig seed instead of a normal seed, allowing the multisig wallet user to not have to resync initialization data. This multisig seed is the same data provided from `query_key_info`.

Edit: Also restoring from multisig seed now works when M <= N-2 and with any number of private multisig keys.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -986,6 +986,21 @@ bool get_pruned_tx(const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry &entry,
   return false;
 }
 
+// Given M (threshold) and N (total), calculate the number of private multisig keys each
+// signer should have. This value is equal to (N - 1) choose (N - M)
+// Prereq: M >= 1 && N >= M && N <= 16
+uint64_t num_priv_multisig_keys_post_setup(uint64_t threshold, uint64_t total)
+{
+  THROW_WALLET_EXCEPTION_IF(threshold < 1 || total < threshold || threshold > 16,
+    tools::error::wallet_internal_error, "Invalid arguments to num_priv_multisig_keys_post_setup");
+
+  uint64_t n_multisig_keys = 1;
+  for (uint64_t i = 2; i <= total - 1; ++i) n_multisig_keys *= i; // multiply by (N - 1)!
+  for (uint64_t i = 2; i <= total - threshold; ++i) n_multisig_keys /= i; // divide by (N - M)!
+  for (uint64_t i = 2; i <= threshold - 1; ++i) n_multisig_keys /= i; // divide by ((N - 1) - (N - M))!
+  return n_multisig_keys;
+}
+
   //-----------------------------------------------------------------
 } //namespace
 
@@ -1394,7 +1409,7 @@ bool wallet2::get_seed(epee::wipeable_string& electrum_words, const epee::wipeab
   return true;
 }
 //----------------------------------------------------------------------------------------------------
-bool wallet2::get_multisig_seed(epee::wipeable_string& seed, const epee::wipeable_string &passphrase, bool raw) const
+bool wallet2::get_multisig_seed(epee::wipeable_string& seed, const epee::wipeable_string &passphrase) const
 {
   bool ready;
   uint32_t threshold, total;
@@ -1408,15 +1423,14 @@ bool wallet2::get_multisig_seed(epee::wipeable_string& seed, const epee::wipeabl
     std::cout << "This multisig wallet is not yet finalized" << std::endl;
     return false;
   }
-  if (!raw && seed_language.empty())
-  {
-    std::cout << "seed_language not set" << std::endl;
-    return false;
-  }
+
+  const uint64_t num_expected_ms_keys = num_priv_multisig_keys_post_setup(threshold, total);
 
   crypto::secret_key skey;
   crypto::public_key pkey;
   const account_keys &keys = get_account().get_keys();
+  THROW_WALLET_EXCEPTION_IF(num_expected_ms_keys != keys.m_multisig_keys.size(),
+    error::wallet_internal_error, "Unexpected number of private multisig keys")
   epee::wipeable_string data;
   data.append((const char*)&threshold, sizeof(uint32_t));
   data.append((const char*)&total, sizeof(uint32_t));
@@ -1441,18 +1455,7 @@ bool wallet2::get_multisig_seed(epee::wipeable_string& seed, const epee::wipeabl
     data = encrypt(data, key, true);
   }
 
-  if (raw)
-  {
-    seed = epee::to_hex::wipeable_string({(const unsigned char*)data.data(), data.size()});
-  }
-  else
-  {
-    if (!crypto::ElectrumWords::bytes_to_words(data.data(), data.size(), seed, seed_language))
-    {
-      std::cout << "Failed to encode seed";
-      return false;
-    }
-  }
+  seed = epee::to_hex::wipeable_string({(const unsigned char*)data.data(), data.size()});
 
   return true;
 }
@@ -4905,9 +4908,11 @@ void wallet2::generate(const std::string& wallet_, const epee::wipeable_string& 
   offset += sizeof(uint32_t);
   uint32_t total = *(uint32_t*)(multisig_data.data() + offset);
   offset += sizeof(uint32_t);
-  THROW_WALLET_EXCEPTION_IF(threshold < 2, error::invalid_multisig_seed);
-  THROW_WALLET_EXCEPTION_IF(total != threshold && total != threshold + 1, error::invalid_multisig_seed);
-  const size_t n_multisig_keys =  total == threshold ? 1 : threshold;
+
+  THROW_WALLET_EXCEPTION_IF(threshold < 1, error::invalid_multisig_seed);
+  THROW_WALLET_EXCEPTION_IF(total < threshold, error::invalid_multisig_seed);
+  THROW_WALLET_EXCEPTION_IF(threshold > 16, error::invalid_multisig_seed); // doing N choose (N - M + 1) might overflow
+  const uint64_t n_multisig_keys = num_priv_multisig_keys_post_setup(threshold, total);
   THROW_WALLET_EXCEPTION_IF(multisig_data.size() != 8 + 32 * (4 + n_multisig_keys + total), error::invalid_multisig_seed);
 
   std::vector<crypto::secret_key> multisig_keys;
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -3807,7 +3807,7 @@ namespace tools
     std::string old_language;
 
     // check the given seed
-    {
+    if (!req.enable_multisig_experimental) {
       if (!crypto::ElectrumWords::words_to_bytes(req.seed, recovery_key, old_language))
       {
         er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
@@ -3830,6 +3830,13 @@ namespace tools
 
     // process seed_offset if given
     {
+      if (req.enable_multisig_experimental && !req.seed_offset.empty())
+      {
+        er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+        er.message = "Multisig seeds are not compatible with seed offsets";
+        return false;
+      }
+
       if (!req.seed_offset.empty())
       {
         recovery_key = cryptonote::decrypt_key(recovery_key, req.seed_offset);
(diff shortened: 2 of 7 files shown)


src/wallet/wallet_rpc_server.cpp
```diff
@@ -3893,7 +3893,27 @@ namespace tools
     crypto::secret_key recovery_val;
     try
     {
-      recovery_val = wal->generate(wallet_file, std::move(rc.second).password(), recovery_key, true, false, false);
+      if (req.enable_multisig_experimental)
+      {
+        // Parse multisig seed into raw multisig data
+        epee::wipeable_string multisig_data;
+        multisig_data.resize(req.seed.size() / 2);
+        epee::span<std::uint8_t> ms_span{reinterpret_cast<std::uint8_t*>(multisig_data.data()), multisig_data.size()};
```
vtnerd, line 3901: This likely works, but I think unwrapping is a bit safer `epee::as_mut_span(unwrap(unwrap(multig_data)))`. This does some basic checks to make sure that casting to a `uint8_t*` is safe.

vtnerd (commented):
This multisig seed - how is it presented to the user (you can tell I've never used it before)? Possibly some confusion with the standard seed.

jeffro256:
The multisig seed can be extracted from the “seed” command in the CLI or the “query_key_info” RPC command.

jeffro256:
Yeah ”seed” could potentially be confusing since it’s not just a seed but also threshold information and signer participant  configuration data.

src/wallet/wallet_rpc_server.cpp
jeffro256: `epee::to_mut_span` does not work here because the `value_type` of `wipeable_string` is `char`, not `std::uint8_t`.

jeffro256: I could write a `to_mut_byte_span` function

vtnerd: Sorry, I meant `as_mut_byte_span(unwrap(unwrap(foo)))`, which already exists.

jeffro256: I think that function is only for POD types and not STL-like containers, though, right?

vtnerd: My apologies, I was thinking of the other class which works with `unwrap(...)`.

vtnerd (approved):

woodser:
@jeffro256 would you be able to PR this against the release branch too, so I can incorporate this change into my next release?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3893,7 +3893,26 @@ namespace tools
...
+        multisig_data.resize(req.seed.size() / 2);
+        if (!epee::from_hex::to_buffer(epee::to_mut_byte_span(multisig_data), req.seed))
+        {
+          er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+          er.message = "Multisig seed not represented as hexadecimal string";
+          return false;
+        }
+
+        // Generate multisig wallet
+        wal->generate(wallet_file, std::move(rc.second).password(), multisig_data, false);
```
woodser, line 3909: I think we also need to set `wal->enable_multisig(true)` as the parameter name suggests, or the wallet's multisig will not be enabled for creating txs and signing.

jeffro256: Good point, you could enable it later thru RPC IIRC, but I don't know why you'd want to restore a multisig seed and not enable that flag.

woodser:
Should the restoration be based on https://github.com/monero-project/monero/blob/00fd416a99686f0956361d1cd0337fe56e58d4a7/src/simplewallet/simplewallet.cpp#L4867, including restoring with a seed passphrase or offset?

jeffro256:
> I'm getting errors testing restoring multisig wallets from seed.
> 
> First, successfully restored 2/2 or 2/3 multisig wallets fail on `transfer` or `transfer_split` with error "No transaction created", whereas the transaction succeeds if the wallet is not first restored from seed.
> 
> Second, 2/4 multisig wallets fail to restore at all with error: "invalid multisig seed".

The multisig seed includes configuration data so you can't change the number of participants and get a valid multisig wallet. You will need to generate a new multisig wallet for each M/N combination.

woodser:
@jeffro256 Yes, these are tests for separate wallets. The M/N is never changed per wallet.

jeffro256:
Can I see the logs for the "invalid multisig seed" error? The line numbers for that error might help determine what is going on

woodser:
It's thrown from `THROW_WALLET_EXCEPTION_IF(total != threshold && total != threshold + 1, error::invalid_multisig_seed);` here: https://github.com/monero-project/monero/blob/65035231aac56ece05302d039a718a03e58f1060/src/wallet/wallet2.cpp#L5082

woodser:
I'm exchanging keys N - M + 1 times here in my [test](https://github.com/monero-ecosystem/monero-java/blob/e6ed9522ba9aa876afbcb98f2391cad338faf24f/src/test/java/test/TestMoneroWalletCommon.java#L3772), assuming that is right?

jeffro256:
> It's thrown from here when the wallet is generated:
> 
> https://github.com/monero-project/monero/blob/65035231aac56ece05302d039a718a03e58f1060/src/wallet/wallet2.cpp#L5082
> 
> Let me know if you want more log context from the key image exchange.

That specific assertion is only necessary for old multisig when thresholds were limited to N/N or (N-1)/N

jeffro256:
I don't know why it's still there

jeffro256:
Okay I pushed changes to do three things: 1) In `wallet2::generate()` (the multisig overload), I removed the restriction for only N/N and N-1/N multisig, 2) In that same method, I changed the calculation for `n_multisig_keys` to factor in the size of the multisig seed data since the size of the private multisig key vector can vary based on the rounds of finalization, and 3) I removed the option to turn the multisig seed into Electrum words since the smallest possible seed (2/2) needs at least 172 words to encode it (without a checksum) and that feature was unused in the codebase.

jeffro256:
I also updated the functional tests to actually try a transfer of funds after restoring from the multisig seed

woodser:
Can you also push the changes to the #8942 so I can test it from the release branch? Thanks.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3893,7 +3893,27 @@ namespace tools
     crypto::secret_key recovery_val;
     try
     {
-      recovery_val = wal->generate(wallet_file, std::move(rc.second).password(), recovery_key, true, false, false);
+      if (req.enable_multisig_experimental)
+      {
+        // Parse multisig seed into raw multisig data
+        epee::wipeable_string multisig_data;
+        multisig_data.resize(req.seed.size() / 2);
```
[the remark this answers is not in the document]
jeffro256: Yup!

src/wallet/wallet2.cpp
```diff
@@ -4905,10 +4889,12 @@ void wallet2::generate(const std::string& wallet_, const epee::wipeable_string&
...
-  THROW_WALLET_EXCEPTION_IF(threshold < 2, error::invalid_multisig_seed);
-  THROW_WALLET_EXCEPTION_IF(total != threshold && total != threshold + 1, error::invalid_multisig_seed);
-  const size_t n_multisig_keys =  total == threshold ? 1 : threshold;
-  THROW_WALLET_EXCEPTION_IF(multisig_data.size() != 8 + 32 * (4 + n_multisig_keys + total), error::invalid_multisig_seed);
+
+  THROW_WALLET_EXCEPTION_IF(threshold < 2 || threshold > total, error::invalid_multisig_seed);
+  THROW_WALLET_EXCEPTION_IF(multisig_data.size() < 8 + 32 * (4 + 1 + total), error::invalid_multisig_seed);
+  const size_t num_bytes_for_ms_keys = multisig_data.size() - 8 - 32 * (4 + total);
+  THROW_WALLET_EXCEPTION_IF(num_bytes_for_ms_keys % 32 != 0, error::invalid_multisig_seed);
+  const size_t n_multisig_keys = num_bytes_for_ms_keys / 32;
```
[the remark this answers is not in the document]
jeffro256: So it would be zero for N/N?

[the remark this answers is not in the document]
jeffro256: This doesn't appear to be the case, in the functional tests, the first 2/2 wallet only has 1 multisig private key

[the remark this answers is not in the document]
jeffro256: Then what is the N choose (N - M + 1)?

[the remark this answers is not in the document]
jeffro256: `(N - 1) choose (N - 1 - M + 1)`?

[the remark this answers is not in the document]
jeffro256: b/c `(N -1) choose (N-1) - M + 1` == `-M`

src/wallet/wallet2.cpp
```diff
@@ -986,6 +986,21 @@ bool get_pruned_tx(const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::entry &entry,
...
+{
+  THROW_WALLET_EXCEPTION_IF(threshold < 1 || total < threshold || threshold > 16,
+    tools::error::wallet_internal_error, "Invalid arguments to num_priv_multisig_keys_post_setup");
+
+  uint64_t n_multisig_keys = 1;
+  for (uint64_t i = 2; i <= total - 1; ++i) n_multisig_keys *= i; // multiply by (N - 1)!
+  for (uint64_t i = 2; i <= total - threshold; ++i) n_multisig_keys /= i; // divide by (N - M)!
+  for (uint64_t i = 2; i <= threshold - 1; ++i) n_multisig_keys /= i; // divide by ((N - 1) - (N - M))!
+  return n_multisig_keys;
+}
```
jeffro256, line 1002: @UkoeHB This formula appears to work correctly to calculate the number of private multisig keys per signer, thanks!

tests/functional_tests/multisig.py
```diff
@@ -156,6 +200,42 @@ def create_multisig_wallets(self, M_threshold, N_total, expected_address):
         assert res.threshold == M_threshold
         assert res.total == N_total
 
+    def remake_some_multisig_wallets_by_multsig_seed(self, threshold):
+      N = len(self.wallet)
+      signers_to_remake = set()
+      num_signers_to_remake = random.randint(1, N) # Do at least one
+      while len(signers_to_remake) < num_signers_to_remake:
+        signers_to_remake.add(random.randint(0, N - 1))
```
jeffro256, line 208: I changed the function to restore *some* not *all* multisig wallets in case bugs would arise when a seed-restored interacted w/ a normal wallet.

tests/functional_tests/multisig.py
```diff
@@ -54,30 +56,72 @@ def run_test(self):
...
+        self.import_multisig_info([0, 1, 2, 3], 6) # six outputs, same as before
+        txid = self.transfer([2, 3])
+        self.import_multisig_info([0, 1, 2, 3], 7) # seven outputs b/c we're dest plus change
+        self.check_transaction(txid)
+
+        self.create_multisig_wallets(1, 2, '4A8RnBQixry4VXkqeWhmg8L7vWJVDJj4FN9PV4E7Mgad5ZZ6LKQdn8dYJP2RePX6HMaSkbvTbrWUFhDNcNcHgtNmQ4S8RSB')
+        self.import_multisig_info([0, 1], 5)
+        txid = self.transfer([0])
+        self.import_multisig_info([0, 1], 6)
+        self.check_transaction(txid)
```
jeffro256, line 117: I added a test case for a 1/2 multisig wallet since none of the cases yet have 1 as the threshold

woodser:
With the latest push to the release branch, 2/4 multisig wallets are restoring successfully.

However, calling `transfer` or `transfer_split` after restoring any multisig wallet still gives the error "No transaction created". Hopefully this is reproducible in the unit tests.

It works as expected without restoring the multisig wallet from seed.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3807,7 +3807,7 @@ namespace tools
     std::string old_language;
 
     // check the given seed
-    {
+    if (!req.enable_multisig_experimental) {
```
woodser, line 3810: As is, `recovery_key` will not be initialized when restoring from multisig seed, which would cause an error if `seed_offset` is also provided.

Should `seed_offset` be supported when restoring from multisig seed?

If not we should assert that `seed_offset` isn't provided when restoring from multisig seed. Otherwise `recovery_key` needs initialized too.

jeffro256: Yes `seed_offset` is incompatible with multisig wallets so that condition should be guarded against.

selsta:
@woodser are all your issues resolved?

woodser:
@selsta No this isn't resolved yet; still waiting for a fix to the "No transactions created" error after restoring from seed.

jeffro256:
So I took the time to setup the monero-java JUnit tests using the PR branch that @woodser was using using a patched (with this PR) wallet rpc server. What I found was that the "no transactions created" error for me was 1) temporary and 2) affected both restoring with multisig seed and not restoring. I think that there might be a race condition in tests somewhere that is exacerbated by restoring by seed since the cache is flushed, but I have not been able to reliably reproduce this error; the test cases usually passes for me when restoring from multisig seed.

woodser:
After some digging based on @jeffro256 's comment, I found a bug in the Java tests when restoring from seed, where the keys were not being exchanged correctly (fixed in https://github.com/monero-ecosystem/monero-java/pull/51/commits/13004ccd3bd277dc47a166e8d63af37f1834a02d).

It's working now with that fix. :)

selsta (approved):

PR #8919, opened 2023-06-27
Fixup error message.

ComputeryPony:
Looks like the extra MWARNING was supposed to be guarded by the if statement.


selsta (approved):

jeffro256 (approved):

tobtoht (approved):

src/common/dns_utils.cpp
OrvilleRed, line 1: Sorry, I was the one messed that up the formatting here. Thank you for cleaning it, I will try to be mindful next time

PR #8925, opened 2023-07-01
I noticed that "which" was typo'd as "wihch" 46 times, so I fixed that

OrvilleRed:
There it is, the most trivial PR you'll see all day. I did this once long ago (PR 8803) but at that time I screwed up the commit. This time hopefully at least it's clean.

EDIT: It's up to 47 times, I copied and pasted the old PR's headline to ensure I'd get the exact misspelling. Sorry.

### DIFF
--- a/translations/monero.ts
+++ b/translations/monero.ts
@@ -5860,7 +5860,7 @@ WARNING: Do not reuse your Monero keys on another fork, UNLESS this fork has key
     </message>
     <message>
         <location filename="../src/wallet/wallet2.cpp" line="289"/>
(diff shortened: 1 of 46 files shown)


selsta:
The typo is already fixed inside the source code: https://github.com/monero-project/monero/blob/master/src/wallet/wallet2.cpp#L276

The .ts files are automatically generated, we usually don't edit them manually. @erciccione can you confirm?

erciccione:
> The .ts files are automatically generated, we usually don't edit them manually. @erciccione can you confirm?

That's correct. If the typo is fixed in the source code, we only need to update the language files using lupdate. The last time this was done was long ago, so makes sense that the old typo is still there

selsta:
Closing as per the previous comment.

PR #8928, opened 2023-07-01
Run miner threads at priority `SCHED_IDLE` (see `sched(7)`)

ppetr:
This way miners can fully utilize all spare CPU power without affecting performance of the rest of the system.


jtgrassie:
Why the change? Miners likely want mining to be at full-throttle, not at a lowered priority. There's also already the feature (built-in) to mine when idle (that's cross-platform).

ppetr:
I'm aware of the feature, but I believe it's sub-optimal: The background mining feature sort of mimics running a thread at an idle priority, but by nature it's always a crude approximation.

- When the load average is above `--bg-mining-idle-threshold`, the system can have spare CPU resources, but the background miner thread remains stopped.
- It reacts to the system's state only slowly based on `--bg-mining-min-idle-interval`. Which can be bad both ways - it can under-utilize available CPU power, or negatively affect other processes' performance (like when running on a desktop machine).

So running a thread at an idle priority seems to me to be better in multiple aspects: It lets the system to utilize full CPU power available, lets the system react immediately to the CPU demand, and it's simpler to implement.

If you prefer, I can add a new feature flag for this that'd complement the existing background mining feature.

jtgrassie:
> but I believe it's sub-optimal

I don't disagree the current solution is sub-optimal, but it is 1) cross-platform and 2) surely impacts your changes anyway. 

> If you prefer, I can add a new feature flag for this that'd complement the existing background mining feature.

I think that's probably a better approach. Though a user can already simply ignore the `--bg-mining-*` options (i.e. not ask for bg mining in the `start_mining` command), and launch `monerod` with `nice`. But if you can make your changes work across all platforms, and it's an option instead of (or complimenting) the existing approach, I suppose that's more desirable (as it only affects the mining threads).

src/cryptonote_basic/thread.h
```diff
@@ -0,0 +1,88 @@
...
+
+// Creates and returns a thread running at an idle priority (if possible).
+template<typename Callable>
+boost::thread create_background_thread(boost::thread::attributes& attrs, Callable func);
+
+#ifdef _WIN32
+
+// TODO: Add a Windows implementation.
+// Most likely by calling `SetThreadPriority` with
+// `THREAD_MODE_BACKGROUND_BEGIN` within the created thread.
```
jtgrassie, line 51: Probably needs to be something like:
```
auto thread = boost::thread(attrs, std::forward<Callable>(func));
auto handle = thread.native_handle();
SetThreadPriority(handle, THREAD_PRIORITY_LOWEST); 
```

(THREAD_MODE_BACKGROUND_BEGIN only works for the *current* thread according to the docs).

ppetr:
> a user can already simply ignore the --bg-mining-* options (i.e. not ask for bg mining in the start_mining command), and launch monerod with nice. 

I thought of this option too, but apparently that's problematic because it also prevents the main daemon from progressing as it should (https://monero.stackexchange.com/a/13983/16159).

OK, I'll try to adapt it to all platforms and add an appropriate flag.

ppetr:
> THREAD_MODE_BACKGROUND_BEGIN only works for the current thread according to the docs

Yes, I noticed that, and this was why I designed the function so that creates a thread, so that it can execute code within.

[PTAL](https://en.wiktionary.org/wiki/PTAL)?

src/cryptonote_basic/miner.cpp
```diff
@@ -401,7 +410,13 @@ namespace cryptonote
...
     for(size_t i = 0; i != m_threads_total; i++)
     {
-      m_threads.push_back(boost::thread(m_attrs, boost::bind(&miner::worker_thread, this)));
+      if (threads_idle) {
+        m_threads.push_back(create_background_thread(m_attrs, boost::bind(
+            &miner::worker_thread, this)));
+      } else {
+        m_threads.push_back(boost::thread(m_attrs, boost::bind(
+            &miner::worker_thread, this)));
+      }
```
jtgrassie, line 419: Minor gripe, but formatting should follow the rest of the file (here, braces on newlines).

src/cryptonote_basic/miner.cpp
```diff
@@ -411,6 +426,9 @@ namespace cryptonote
 
     if( get_is_background_mining_enabled() )
     {
+      if (threads_idle) {
+        MERROR("Feature --mining-threads-idle is an alternative to background mining and doesn't make sense to be used together.");
+      }
```
jtgrassie, line 431: Same, follow rest of file style

ppetr:
Thank you for the review. Hmm, one test is failing, IDK why, I'll need to figure this out.

ppetr:
> I can't speak to whether how you've now done it for Windows works or not (I presume you've tested), but just using THREAD_PRIORITY_LOWEST should suffice, and then launching in the same way as the non-windows version. I suppose if the way you've implemented works fine though, it may be better...

Yes, tested, this approach seems to works quite nicely. According to the [docs](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreadpriority) using `THREAD_MODE_BACKGROUND_BEGIN` should ensure that 

> _the system lowers the resource scheduling priorities of the thread so that it can perform background work without significantly affecting activity in the foreground._

I'll fix the formatting and also I still need to figure out why one of the tests is broken.

PR #8930, opened 2023-07-03
wallet_api: don't skip refresh if the daemon is syncing

jeffro256:
This PR revives #7163, which would allow the Monero GUI wallet (and anyone else who use `libwallet`) to refresh concurrently while the daemon is syncing. @xiphon appears to be inactive for a couple years so I don't think those merge conflicts would be resolved anytime soon ;)

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -2206,25 +2206,19 @@ void WalletImpl::doRefresh()
     boost::lock_guard<boost::mutex> guarg(m_refreshMutex2);
     do try {
         LOG_PRINT_L3(__FUNCTION__ << ": doRefresh, rescan = "<<rescan);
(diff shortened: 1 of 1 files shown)


tobtoht (approved):

selsta:
Did you test this with the GUI?

hinto-janai (approved):
Tested on Linux v0.18.3.1 GUI/CLI

plowsof:
some graphical glitches during sync with the GUI e.g. 33 seconds

[Screencast from 07-11-23 23:15:14.webm](https://github.com/monero-project/monero/assets/77655812/e1b12426-6faa-4337-992e-fd2751c2e3b6)

Tzadiko:
A bit confused by the change here. The comment states that synching the daemon and refreshing at the same time is expensive, but you just got rid of the `if` that checks for whether we are synced without an accompanying change to address doing both at the same time being expensive.

PR #8929, opened 2023-07-02
Bootstrappable Builds

tobtoht:
This PR proposes to replace Gitian with Guix to achieve bootstrappable builds for release binaries.

### What?

https://youtube.com/watch?v=I2iShmUTEl8

If you have 15 minutes, please consider watching this presentation by Carl Dong linked above as it covers the migration (Gitian → Guix) this PR proposes to make. It explains why reproducibility alone is not enough.

Other resources:

- https://github.com/Sjors/nado-book/blob/master/attacks/guix.md
- https://bootstrappable.org

### Why?

- **Greater build system security**: Bootstrappability allows us to _audit and reproduce_ our toolchain instead of blindly _trusting_ binary downloads. Our build environment can be built from source, [all the way down](https://guix.gnu.org/en/blog/2023/the-full-source-bootstrap-building-from-source-all-the-way-down/). It allows us to reduce our supply chain attack surface by only including the packages that we need, and nothing else.

- **Easier to set up**: Guix runs on any Linux distribution and on most architectures (x86_64, aarch64, riscv64). To produce reproducible release binaries, you only need to install Guix and run the build script. This hopefully leads to more participants in the pre-release verified reproduction process.

- **More flexibility**: Unlike Gitian, we are not limited to the package set of a particular Ubuntu version. Guix allows us to pick and choose our toolchains. This allows us to use the latest compilers while targeting older versions of glibc and would, for example, make it easier to add a modern Rust compiler to our build environment. Packages that are not available in Guix can easily be defined in the manifest or upstreamed.

- **Better developer UX**: Guix allows us to modify any detail about our build environment with ease. Debugging build issues takes less time because we have shell access to the build environment. The monero source code is bind mounted into the container, so edits to package definitions can be tested incrementally.

- **Future proof**: Guix is actively developed, Gitian is in [maintenance mode](https://github.com/devrandom/gitian-builder) (serious bug fixes only).

### How to test?

Requirements:

- any Linux distribution
- 50 GB of free disk space
- 4 or more cores recommended
- 2 GB RAM per thread

```bash
# Install guix
$ apt install guix  # ubuntu 22.04, debian 11, or later 
or 
# use the install script: https://guix.gnu.org/manual/en/html_node/Binary-Installation.html

# Clone the repo
$ git clone https://github.com/tobtoht/monero
$ cd monero
$ git checkout guix

# Run the builds
$ ./contrib/guix/guix-build  # this will take several hours, uses all available cores
or
[...]

Issue #8504, opened 2022-08-15
Couldn't allocate RandomX cache


slurfius:
macOS High Sierra v 10.13.6 
Monero 'Fluorine Fermi' (v0.18.1.0-release)
started to sync but after 5 minutes
Couldn't allocate RandomX cache
error then crashes

slurfius:
./monerod
2022-08-15 14:28:09.138	I Monero 'Fluorine Fermi' (v0.18.1.0-release)
2022-08-15 14:28:09.139	I Initializing cryptonote protocol...
2022-08-15 14:28:09.139	I Cryptonote protocol initialized OK
2022-08-15 14:28:09.139	I Initializing core...
2022-08-15 14:28:09.140	I Loading blockchain from folder /Users/Jonathan/.bitmonero/lmdb ...
2022-08-15 14:28:11.382	I Loading checkpoints
2022-08-15 14:28:11.383	I Core initialized OK
2022-08-15 14:28:11.383	I Initializing p2p server...
2022-08-15 14:28:11.391	I p2p server initialized OK
2022-08-15 14:28:11.391	I Initializing core RPC server...
2022-08-15 14:28:11.391	I Binding on 127.0.0.1 (IPv4):18081
2022-08-15 14:28:11.414	I core RPC server initialized OK on port: 18081
2022-08-15 14:28:11.415	I Starting core RPC server...
2022-08-15 14:28:11.415	I core RPC server started ok
2022-08-15 14:28:11.416	I Starting p2p net loop...
2022-08-15 14:28:12.420	I 
2022-08-15 14:28:12.420	I **********************************************************************
2022-08-15 14:28:12.420	I The daemon will start synchronizing with the network. This may take a long time to complete.
2022-08-15 14:28:12.420	I 
2022-08-15 14:28:12.420	I You can set the level of process detailization through "set_log <level|categories>" command,
2022-08-15 14:28:12.420	I where <level> is between 0 (no details) and 4 (very verbose), or custom category based levels (eg, *:WARNING).
2022-08-15 14:28:12.420	I 
2022-08-15 14:28:12.420	I Use the "help" command to see the list of available commands.
2022-08-15 14:28:12.420	I Use "help <command>" to see a command's documentation.
2022-08-15 14:28:12.420	I **********************************************************************
[...]

PR #8933, opened 2023-07-06
depends: use macOS 11 SDK with libc++ headers

tobtoht:
Needed for #8929.

Our current SDK is missing libc++ headers and instead uses whichever headers are included in the binary clang package downloaded in depends, which we can't use with bootstrappable builds.

Considering that macOS 10.14 has already been EOL for almost two years and according to selsta: "monero doesn't run on 10.8 anyway i think. iirc 10.13 or 10.14 is the last working.", it should not be an issue to bump the minimum OSX version.

Edit: lowered minimum OSX version to 10.13.

Builds are passing, but not yet tested.

#### How to reproduce this SDK

Download XCode 12.2 from [Apple](https://download.developer.apple.com/Developer_Tools/Xcode_12.2/Xcode_12.2.xip) (account required) or [archive.org](https://archive.org/details/xcode-12.2).

Follow the instructions [here](https://github.com/bitcoin/bitcoin/tree/master/contrib/macdeploy#sdk-extraction) to create an SDK tarball that includes libc++ headers.

The sha256sum of the SDK should match: `df75d30ecafc429e905134333aeae56ac65fac67cb4182622398fd717df77619`.


selsta:
For context this is the issue that shows RandomX related issues on 10.12 and 10.13: https://github.com/monero-project/monero/issues/8504

vtnerd (approved):

contrib/depends/packages/darwin_sdk.mk
```diff
@@ -1,8 +1,8 @@
 package=darwin_sdk
-$(package)_version=11.1
-$(package)_download_path=https://github.com/phracker/MacOSX-SDKs/releases/download/11.3/
-$(package)_file_name=MacOSX$($(package)_version).sdk.tar.xz
-$(package)_sha256_hash=68797baaacb52f56f713400de306a58a7ca00b05c3dc6d58f0a8283bcac721f8
+$(package)_version=12.2
+$(package)_download_path=https://featherwallet.org/files/SDK/
```
selsta, line 3: I would prefer if we could set this to

https://bitcoincore.org/depends-sources/sdks

instead, I get like 10x the speeds there.

selsta (approved):

selsta (approved):

PR #8940, opened 2023-07-07
epee binary: container can read from empty seq of any type

jeffro256:
Good news: this property holds for current epee binary code in core repo. For real world example of issue brought on if this property is not upheld: https://github.com/monero-rs/monero-epee-bin-serde/issues/49

### DIFF
--- a/tests/unit_tests/epee_serialization.cpp
+++ b/tests/unit_tests/epee_serialization.cpp
@@ -29,8 +29,11 @@
 
 #include <cstdint>
 #include <gtest/gtest.h>
+#include <vector>
 
+#include "serialization/keyvalue_serialization.h"
 #include "storages/portable_storage.h"
+#include "storages/portable_storage_template_helper.h"
 #include "span.h"
 
 TEST(epee_binary, two_keys)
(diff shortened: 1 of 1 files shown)


jeffro256:
@vtnerd does your new wire reading code safety accept any type marker for empty sequences without error?

vtnerd:
These tests should still work with the new serialization engine. The inner type is only checked when processing that first element.

tests/unit_tests/epee_serialization.cpp
```diff
@@ -54,3 +57,87 @@ TEST(epee_binary, duplicate_key)
...
+  EXPECT_TRUE(epee::serialization::load_t_from_binary(o, epee::span<const std::uint8_t>(data_empty_string)));
+  EXPECT_EQ(0, o.x.size());
+
+  EXPECT_TRUE(epee::serialization::load_t_from_binary(o, epee::span<const std::uint8_t>(data_empty_int64)));
+  EXPECT_EQ(0, o.x.size());
+
+  EXPECT_TRUE(epee::serialization::load_t_from_binary(o, epee::span<const std::uint8_t>(data_empty_object)));
+  EXPECT_EQ(0, o.x.size());
+
+  ObjOfObjs i;
```
rbrunner7, line 127: Shouldn't that be `ObjOfInts`?

rbrunner7 (commented):
See the single comment

vtnerd (approved):

tests/unit_tests/epee_serialization.cpp
jeffro256: Yes, good catch ;). Thanks for finding that

rbrunner7 (approved):

PR #8941, opened 2023-07-08
Wallet refresh improvements

moneromooo-monero:


selsta (approved):
Equivalent to https://github.com/monero-project/monero/pull/8355

plowsof:
+1 , the `--no-initial-sync` is a small but imo a huge ux++ for anyone working with monero-wallet-rpc. I can now simply [ask the wallet its current sync height](https://github.com/plowsof/scrape_ccs_fr/blob/5df3e2464bc7f52ebcc267810b418afb56788763/get_overfunded.py#L41) and receive an answer instead of hanging until its fully synced. I have tested this PR and confirm it works, (also run a github action that syncs a wallet once a day which has been benefiting from it) Thanks!

plowsof (approved):

src/wallet/wallet_rpc_server.cpp
```diff
@@ -60,6 +60,7 @@ using namespace epee;
 #define MONERO_DEFAULT_LOG_CATEGORY "wallet.rpc"
 
 #define DEFAULT_AUTO_REFRESH_PERIOD 20 // seconds
+#define REFRESH_INFICATIVE_BLOCK_CHUNK_SIZE 256    // just to split refresh in separate calls to play nicer with other threads
```
iamamyth, line 63: Was inficative (not a word) meant to be indicative? Even indicative seems like an odd name choice, it's more like a default.

moneromooo-monero: Yes, it was. IIRC that value is only respected by monerod if the blocks aren't too big, so it's only indicative. Maybe "max" would have been better.
I won't bother changing it here since it's a copy of the master patch, which is already merged.

selsta (approved):

binarybaron:
I'm using the `monero-wallet-rpc` in my application and I'm attempting to:
1. Refresh wallet by calling `refresh`
2. While `refresh` is pending I want to periodically check how far we've synced by calling `get_height`

Problem is that the rpc does not respond to `get_height` while it's refreshing? Is this expected behaviour. I thought the change made here to sync blocks in batches should have fixed that issue.
Anything I'm missing here?

PR #8948, opened 2023-07-13
Rough draft of a p2p noise protocol implementation

vtnerd:
My CCS is ending, so I wanted to post a draft of the noise protocol because that has been one of the primary efforts. I also wanted some final feedback if the Noise protocol (vs SSL) is worth doing, because I really have to hack the TCP server a bit for this to work.

The code is almost finalized - some stuff is needed on the detection end. I mainly wanted to highlight the painful hacks to squeeze this in.

### DIFF
--- a/contrib/epee/include/net/connection_basic.hpp
+++ b/contrib/epee/include/net/connection_basic.hpp
@@ -58,7 +58,6 @@ namespace epee
 {
 namespace net_utils
 {
-
 	class connection_basic_shared_state
 	{
 		ssl_options_t ssl_options_;
--- a/contrib/epee/include/net/levin_base.h
+++ b/contrib/epee/include/net/levin_base.h
@@ -71,7 +71,6 @@ namespace levin
   };
 #pragma pack(pop)
 
-
 #define LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED 0
 #define LEVIN_INITIAL_MAX_PACKET_SIZE  256*1024      // 256 KiB before handshake
 #define LEVIN_DEFAULT_MAX_PACKET_SIZE 100000000      //100MB by default after handshake
--- a/contrib/epee/include/net/net_ssl.h
+++ b/contrib/epee/include/net/net_ssl.h
@@ -45,11 +45,12 @@ namespace epee
 {
 namespace net_utils
 {
+  // ssl_support_t
 	enum class ssl_support_t: uint8_t {
 		e_ssl_support_disabled,
 		e_ssl_support_enabled,
-		e_ssl_support_autodetect,
-	};
+		e_ssl_support_autodetect
+    };
 
   enum class ssl_verification_t : uint8_t
   {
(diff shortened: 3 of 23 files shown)


jeffro256:
Does the noise code currently save any keys persistently like the SSL files?

jeffro256:
There may be a benefit to integrating an existing library (e.g. https://github.com/rweather/noise-c) versus writing the crypto code from scratch in that it will be less likely that our usage of noise is fingerprinted for some reason due to an oversight. Don't get me wrong, it would still be a lot of work hacking it into the TCP server, but are there other reasons you haven't used an existing library?

contrib/epee/include/net/connection_basic.hpp
```diff
@@ -58,7 +58,6 @@ namespace epee
 {
 namespace net_utils
 {
-
```
jeffro256, line 61: unnecessary whitespace changes

contrib/epee/include/net/levin_base.h
```diff
@@ -71,7 +71,6 @@ namespace levin
   };
 #pragma pack(pop)
 
-
```
jeffro256, line 74: another one

contrib/epee/include/net/net_ssl.h
```diff
@@ -45,11 +45,12 @@ namespace epee
...
 namespace net_utils
 {
+  // ssl_support_t
 	enum class ssl_support_t: uint8_t {
 		e_ssl_support_disabled,
 		e_ssl_support_enabled,
-		e_ssl_support_autodetect,
-	};
+		e_ssl_support_autodetect
+    };
```
jeffro256, line 53: more whitespace changes plus comment that prob isn't needed

jeffro256 (commented):
You could knock 3 of these files off the file modifications list by undoing whitespace changes

vtnerd:
> Does the noise code currently save any keys persistently like the SSL files?

No, all of the keys are generated per-stream and never re-used. Initially my writeup had static keys, but I was concerned about fingerprinting via those keys. In a future version upgrade, static keys can be added opt-in; I think automagically doing it for monero p2p is the wrong approach.

> There may be a benefit to integrating an existing library (e.g. https://github.com/rweather/noise-c) versus writing the crypto code from scratch in that it will be less likely that our usage of noise is fingerprinted for some reason due to an oversight. Don't get me wrong, it would still be a lot of work hacking it into the TCP server, but are there other reasons you haven't used an existing library?

I wanted more control over the protocol which may have been overkill - the length fields are also encrypted+tagged in my proposal. I might be able to get this working with noise-c but it's convoluted even with raw libsodium functions - I have to decrypt the header without auth first before doing the proper decrypt+authenticate call. This is probably overkill because you can get a rough idea of the message size based on timing analysis (one of the reasons why I plan to add a padding feature to all of this).

The noise protocol is somewhat simple to implement, and this was weighed against adding yet-another-dependency that implements all of the Noise variants (Monero should only need two of the variants). And I'm still only about 75% certain that using Noise is worth the effort over SSL/TLS.

vtnerd:
Closing, since the P2P SSL branch was done instead.

PR #8950, opened 2023-07-16
wallet2: when checking frozen multisig tx set, don't assume order

### LINKED ISSUE #8949: wallet2 `frozen` incorrectly asserts key image ordering and causes error on multisig tx signing
A [recent commit](https://github.com/monero-project/monero/commit/dc24312bc3c8c9d865e4576631832c81d6dd212d) to wallet2's `frozen` function incorrectly [asserts that key images are ordered](https://github.com/monero-project/monero/blob/ab826008d614a3722398642942c291be9650be6d/src/wallet/wallet2.cpp#L2008), resulting in the error `Mismatched key image b/t vin and construction data` when multisig participants sync key images such that they're ordered differently.

This issue requests fixing the assertion to not rely on key image ordering.

jeffro256:
Resolves #8949

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1983,14 +1983,14 @@ bool wallet2::frozen(const multisig_tx_set& txs) const
     CHECK_AND_ASSERT_THROW_MES(cd.sources.size() == ptx.tx.vin.size(), "mismatched multisg tx set source sizes");
     for (size_t src_idx = 0; src_idx < cd.sources.size(); ++src_idx)
     {
-      // Check that the key images are consistent between tx vin and construction data
+      // Extract keys images from tx vin and construction data
       const crypto::key_image multisig_ki = rct::rct2ki(cd.sources[src_idx].multisig_kLRki.ki);
       CHECK_AND_ASSERT_THROW_MES(ptx.tx.vin[src_idx].type() == typeid(cryptonote::txin_to_key), "multisig tx cannot be miner");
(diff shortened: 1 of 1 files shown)


woodser:
@jeffro256 Can you also open a PR against the release branch? Thanks.

jeffro256:
> Thanks jeffro256.
> 
> This PR solves this issue mentioned. I have a suggestion. Wouldn't it be better to check the vector?
> 
> For example, pseudocode would be:
> 
> ```
> exist i in 0 < max(vin) 
>     sources[src_idx].multisig_kLRki.ki == vin[i].k_image
> otherwise
>     assert false;
> ```
> 
> This way, we keep the assert, and do not fail when there are legitimate order mismatches.

This method is only checking if the  key images specified in a `multisig_tx_set` are frozen. I added in the assertion initially because I thought the key images were ordered, and I figured "why not" keep the assertions conservative so I didn't have to `insert` twice. However, since `insert` is idempotent in this case, I cover the case that the lists are mismatched and a counterparty tries to bypass the restriction. At any rate, if mismatched key image key lists cause an actually issue with signing, that's an entire other issue that needs to be addressed in another PR.

jeffro256:
> @jeffro256 Can you also open a PR against the release branch? Thanks.

To confirm, does this PR fix the issue your tests were having? It should, but I want to know

woodser:
> To confirm, does this PR fix the issue your tests were having? It should, but I want to know

I'm getting merge conflicts rebasing on the release branch to test.

woodser:
I applied the changes manually and they fix the problem as far as I can tell.

vtnerd (approved):

PR #8953, opened 2023-07-18
wallet2: when checking frozen multisig tx set, don't assume order [RELEASE]

jeffro256:
#8950


selsta (approved):
CI error is unrelated.

PR #8956, opened 2023-07-19
wallet2: fix `scan_tx` custom compare function incorrectly throwing on some envs

### LINKED ISSUE #8951: `scan_tx` failing with error "Expected block not returned by daemon" in release branch
`scan_tx` is failing in the release branch if the wallet is already synced, with the error message: "Expected block not returned by daemon".

The error message was introduced in the release branch with https://github.com/monero-project/monero/pull/8566.

j-berman:
Fixes #8951

On some envs, `scan_tx` throws when a synced wallet scans a tx older than the wallet's most recent tx and the wallet has lots of txs between those 2 txs.

### The bug

Essentially it's possible for this to throw:

```cpp
std::vector<int> v;
for (int i = 0; i < 32; ++i) v.push_back(i);
std::sort(v.begin(), v.end(), [](const int l, const int r) {
    if (l == r) throw std::runtime_error("Compared elem with itself");
    return l < r;
});
```

I incorrectly assumed this was not possible ([reasoning](https://stackoverflow.com/questions/38966516/should-sorting-algorithm-pass-same-element-in-the-comparison-function) for why an `std::sort` implementation chooses to compare an element against itself).

### The fix

Return false when comparing equal elems in the custom comparator; the comparator must meet the following specification:

> returns ​true if the first argument is less than (i.e. is ordered before) the second [and false otherwise]

> For all a, comp(a, a) == false

https://en.cppreference.com/w/cpp/algorithm/sort
https://en.cppreference.com/w/cpp/named_req/Compare

EDIT: clearer reasoning for the bug.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1715,21 +1715,26 @@ void wallet2::sort_scan_tx_entries(std::vector<process_tx_entry_t> &unsorted_tx_
     else // l.tx_entry.block_height == r.tx_entry.block_height
     {
       // coinbase tx is the first tx in a block
+      if (cryptonote::is_coinbase(r.tx))
+        return false;
       if (cryptonote::is_coinbase(l.tx))
         return true;
-      if (cryptonote::is_coinbase(r.tx))
+
+      // in case std::sort is comparing elem to itself
+      if (l.tx_hash == r.tx_hash)
         return false;
 
       // see which tx hash comes first in the block
       THROW_WALLET_EXCEPTION_IF(parsed_blocks.find(l.tx_entry.block_height) == parsed_blocks.end(),
-          error::wallet_internal_error, "Expected block not returned by daemon");
+          error::wallet_internal_error, std::string("Expected block not returned by daemon, ") +
+          "left tx: " + string_tools::pod_to_hex(l.tx_hash) + ", right tx: " + string_tools::pod_to_hex(r.tx_hash));
       const auto &blk = parsed_blocks[l.tx_entry.block_height];
       for (const auto &tx_hash : blk.tx_hashes)
       {
-        if (tx_hash == l.tx_hash)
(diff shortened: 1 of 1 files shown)


jeffro256 (approved):

jeffro256:
This makes the `cmp_tx_entry` lambda compliant with the C++ standard, but why would this cause the aforementioned assertion statement to not trigger?

j-berman:
A bit long-winded...

First thing to keep in mind: `unsorted_tx_entries` in this function has no duplicates (can trace up the call stack and see why).

Second thing to keep in mind: the wallet needs to process txs in the order they appear in the chain. If there are 2+ non-coinbase txs in a block that the wallet needs to process, then the wallet needs to fetch that block from the daemon in order to determine relative order of those 2+ txs. If there is just 1 non-coinbase tx in a block the wallet needs to process, the wallet doesn't need to fetch that tx's block from the daemon since its order in the chain is known relative to other txs it needs to process.

In @woodser 's env, the sort function calls `cmp_tx_entry` with the same tx for `l` and `r` in his test case (i.e. it compares the same tx against itself). Further, it's calling it with a non-pool and non-coinbase tx, for which the wallet doesn't request that tx's block from the daemon, since the wallet doesn't need to process any more non-coinbase txs from that tx's block. In that case, the tx falls through all if statements in `cmp_tx_entry` until it hits the error @woodser was seeing.

If the sort function never compares an element against itself, it doesn't hit that error because the only way to hit that error is if `l` and `r` have the same block height, and if they have the same block height and are distinct txs, then this section should have fetched the txs' block from the daemon:

https://github.com/monero-project/monero/blob/00fd416a99686f0956361d1cd0337fe56e58d4a7/src/wallet/wallet2.cpp#L1659-L1685

selsta:
Please also open against release.

PR #8957, opened 2023-07-19
wallet2: fix `scan_tx` custom compare function incorrectly throwing on some envs [release]

j-berman:
See #8956


woodser (approved):
These changes fix the problem I was having with scanning. Thanks!

PR #8958, opened 2023-07-19
blockchain_db: add optional k-anonymity to txid fetching

jeffro256:
Read more about k-anonymity [here](https://en.wikipedia.org/wiki/K-anonymity). We implement this feature in the monero daemon for transactions by providing a "Txid Template", which is simply a txid with all but `num_matching_bits` bits zeroed out, and the number `num_matching_bits`. We add an operation to `BlockchainLMDB` called `get_txids_loose` which takes a txid template and returns all txids in the database (chain and mempool) that satisfy that template. Thus, a client can ask about a specific transaction from a daemon without revealing the exact transaction they are inquiring about. The client can control the statistical chance that other TXIDs (besides the one in question) match the txid template sent to the daemon up to a power of 2. For example, if a client sets their `num_matching_bits` to 5, then statistically any txid has a 1/(2^5) chance to match. With `num_matching_bits`=10, there is a 1/(2^10) chance, so on and so forth.

### DIFF
--- a/src/cryptonote_basic/cryptonote_basic_impl.cpp
+++ b/src/cryptonote_basic/cryptonote_basic_impl.cpp
@@ -310,6 +310,53 @@ namespace cryptonote {
   bool operator ==(const cryptonote::block& a, const cryptonote::block& b) {
     return cryptonote::get_block_hash(a) == cryptonote::get_block_hash(b);
   }
+  //--------------------------------------------------------------------------------
+  int compare_hash32_reversed_nbits(const crypto::hash& ha, const crypto::hash& hb, unsigned int nbits)
+  {
+    static_assert(sizeof(uint64_t) * 4 == sizeof(crypto::hash), "hash is wrong size");
+
+    // We have to copy these buffers b/c of the strict aliasing rule
+    uint64_t va[4];
+    memcpy(va, &ha, sizeof(crypto::hash));
+    uint64_t vb[4];
+    memcpy(vb, &hb, sizeof(crypto::hash));
+
+    for (int n = 3; n >= 0 && nbits; --n)
+    {
+      const unsigned int msb_nbits = std::min<unsigned int>(64, nbits);
+      const uint64_t lsb_nbits_dropped = static_cast<uint64_t>(64 - msb_nbits);
+      const uint64_t van = SWAP64LE(va[n]) >> lsb_nbits_dropped;
+      const uint64_t vbn = SWAP64LE(vb[n]) >> lsb_nbits_dropped;
+      nbits -= msb_nbits;
+
+      if (van < vbn) return -1; else if (van > vbn) return 1;
+    }
+
+    return 0;
+  }
+
+  crypto::hash make_hash32_loose_template(unsigned int nbits, const crypto::hash& h)
+  {
+    static_assert(sizeof(uint64_t) * 4 == sizeof(crypto::hash), "hash is wrong size");
+
+    // We have to copy this buffer b/c of the strict aliasing rule
+    uint64_t vh[4];
+    memcpy(vh, &h, sizeof(crypto::hash));
+
+    for (int n = 3; n >= 0; --n)
+    {
+      const unsigned int msb_nbits = std::min<unsigned int>(64, nbits);
+      const uint64_t mask = msb_nbits ? (~((std::uint64_t(1) << (64 - msb_nbits)) - 1)) : 0;
+      nbits -= msb_nbits;
+
+      vh[n] &= SWAP64LE(mask);
+    }
+
+    crypto::hash res;
+    memcpy(&res, vh, sizeof(crypto::hash));
+    return res;
+  }
+  //--------------------------------------------------------------------------------
 }
 
 //--------------------------------------------------------------------------------
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -3533,6 +3533,82 @@ namespace cryptonote
     return true;
   }
   //------------------------------------------------------------------------------------------------------------------------------
+  bool core_rpc_server::on_get_txids_loose(const COMMAND_RPC_GET_TXIDS_LOOSE::request& req, COMMAND_RPC_GET_TXIDS_LOOSE::response& res, epee::json_rpc::error& error_resp, const connection_context *ctx)
+  {
+    RPC_TRACKER(get_txids_loose);
+
+    // Maybe don't use bootstrap since this endpoint is meant to retreive TXIDs w/ k-anonymity,
+    // so shunting this request to a random node seems counterproductive.
+
+#if BYTE_ORDER == LITTLE_ENDIAN
+    const uint64_t max_num_txids = RESTRICTED_SPENT_KEY_IMAGES_COUNT * (m_restricted ? 1 : 10);
+
+    // Sanity check parameters
+    if (req.num_matching_bits > 256)
+    {
+      error_resp.code = CORE_RPC_ERROR_CODE_WRONG_PARAM;
+      error_resp.message = "There are only 256 bits in a hash, you gave too many";
+      return false;
+    }
+
+    // Attempt to guess when bit count is too low before fetching, within a certain margin of error
+    const uint64_t num_txs_ever = m_core.get_blockchain_storage().get_db().get_tx_count();
+    const uint64_t num_expected_fetch = (num_txs_ever >> std::min((int) req.num_matching_bits, 63));
+    const uint64_t max_num_txids_with_margin = 2 * max_num_txids;
+    if (num_expected_fetch > max_num_txids_with_margin)
+    {
+      error_resp.code = CORE_RPC_ERROR_CODE_WRONG_PARAM;
+      error_resp.message = "Trying to search with too few matching bits, detected before fetching";
+      return false;
+    }
+
+    // Convert txid template to a crypto::hash
+    crypto::hash search_hash;
+    if (!epee::string_tools::hex_to_pod(req.txid_template, search_hash))
+    {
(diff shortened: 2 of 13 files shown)


jeffro256:
A real-world application of this feature would be to implement a privacy-preserving block explorer like @ACK-J is planning to work on

ACK-J:
The application of k-anonymity into the onion-monero block explorer can be tracked [here](https://github.com/moneroexamples/onion-monero-blockchain-explorer/issues/284)
I hope to have the txid part integrated soon.

src/cryptonote_basic/cryptonote_basic_impl.cpp
```diff
@@ -310,6 +310,42 @@ namespace cryptonote {
   bool operator ==(const cryptonote::block& a, const cryptonote::block& b) {
     return cryptonote::get_block_hash(a) == cryptonote::get_block_hash(b);
   }
+  //--------------------------------------------------------------------------------
+  int compare_hash32_reversed_nbits(const crypto::hash& ha, const crypto::hash& hb, unsigned int nbits)
+  {
+    const uint64_t* const va = reinterpret_cast<const uint64_t*>(ha.data);
+    const uint64_t* const vb = reinterpret_cast<const uint64_t*>(hb.data);
+
+    for (int n = 3; n >= 0 && nbits; --n)
```
vtnerd, line 319: `static_assert` that `sizeof(uint64_t) * 4 == sizeof(crypto::hash)`. I doubt we'll ever change this, but its a nice catch regardless.

src/cryptonote_basic/cryptonote_basic_impl.cpp
```diff
@@ -310,6 +310,42 @@ namespace cryptonote {
...
+
+    return 0;
+  }
+
+  crypto::hash make_hash32_loose_template(unsigned int nbits, const crypto::hash& h)
+  {
+    crypto::hash res = h;
+    uint64_t* vh = reinterpret_cast<uint64_t*>(res.data);
+
+    for (int n = 3; n >= 0; --n)
```
vtnerd, line 338: `static_assert` here too.

src/cryptonote_basic/cryptonote_basic_impl.cpp
```diff
@@ -310,6 +310,42 @@ namespace cryptonote {
...
+    {
+      const unsigned int msb_nbits = std::min((unsigned int) 64, nbits);
+      const uint64_t mask = msb_nbits ? (~((static_cast<uint64_t>(1) << (64 - msb_nbits)) - 1)) : 0;
```
vtnerd, line 341: For future reference, `std::uint64_t(1)` does the same thing as the `static_cast` but with fewer characters.

src/cryptonote_basic/cryptonote_basic_impl.cpp
```diff
@@ -310,6 +310,42 @@ namespace cryptonote {
...
+      if (van < vbn) return -1; else if (vbn > van) return 1;
+    }
+
+    return 0;
+  }
+
+  crypto::hash make_hash32_loose_template(unsigned int nbits, const crypto::hash& h)
+  {
+    crypto::hash res = h;
+    uint64_t* vh = reinterpret_cast<uint64_t*>(res.data);
```
vtnerd, line 336: This is technically an aliasing violation. It likely works nearly all the time, but optimizations could theoretical blow this up. I know we do this in a few places, but its suboptimal.

Instead, use a `std::array<std::uint64_t, 4>` with a memcpy from `h` into this array, then a `memcpy` back out for the return address.

tests/unit_tests/lmdb.cpp
```diff
@@ -53,6 +55,24 @@ namespace
         MDB_val right_val = lmdb::to_val(right);
         return (*cmp)(&left_val, &right_val);
     }
+
+    crypto::hash postfix_hex_to_hash(const std::string& hex)
```
vtnerd, line 59: I don't see any direct tests for the compare function.

src/rpc/core_rpc_server.cpp
```diff
@@ -3533,6 +3533,82 @@ namespace cryptonote
...
+    catch (const std::exception& e)
+    {
+      error_resp.code = CORE_RPC_ERROR_CODE_INTERNAL_ERROR;
+      error_resp.message = std::string("Error during get_txids_loose: ") + e.what();
+      return false;
+    }
+
+    res.status = CORE_RPC_STATUS_OK;
+    return true;
+#else // BYTE_ORDER == BIG_ENDIAN
```
vtnerd, line 3604: Which part is not big endian safe? I saw several swaps that should work .. no? I didn't inspect this part too closely so maybe I missed something.

src/cryptonote_basic/cryptonote_basic_impl.cpp
jeffro256: I actually don't think this is an aliasing violation because `crypto::hash::data` is a `char` array, and `char` pointers are an exception to the strict aliasing rule, but I could be wrong.

src/rpc/core_rpc_server.cpp
jeffro256: The code that I wrote is big-endian safe, but the way that LMDB orders the keys is not: https://github.com/monero-project/monero/blob/00fd416a99686f0956361d1cd0337fe56e58d4a7/src/blockchain_db/lmdb/db_lmdb.cpp#L148-L160 In other words, the behavior of `cryptonote::make_hash32_template` is different from `BlockchainLMDB::compare_hash32` on big-endian systems so the templates will just return garbage (likely nothing).

src/cryptonote_basic/cryptonote_basic_impl.cpp
vtnerd: Casting **to** char is always allowed, but casting **from** char has restrictions.

jeffro256: How does the new commit look in terms of aliasing?

jeffro256:
There was a bug in certain cases in `compare_hash32_reversed_nbits`, and unit tests were added to assert that it stays correct. Thanks @vtnerd

src/cryptonote_basic/cryptonote_basic_impl.cpp
```diff
@@ -310,6 +310,49 @@ namespace cryptonote {
...
+
+    return 0;
+  }
+
+  crypto::hash make_hash32_loose_template(unsigned int nbits, const crypto::hash& h)
+  {
+    static_assert(sizeof(uint64_t) * 4 == sizeof(crypto::hash), "hash is wrong size");
+
+    uint64_t resarr[4];
+    const uint64_t* vh = reinterpret_cast<const uint64_t*>(h.data);
```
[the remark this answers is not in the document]
jeffro256: Yeah you're right... It just feels so wrong to copy these buffers even though it's the exact same thing in memory.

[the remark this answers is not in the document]
jeffro256: Hopefully the compiler can optimize it away

vtnerd (approved):

Lissy93 (approved):

SChernykh:
@jeffro256 Do I understand it correctly that you scan the entire list of tx hashes to find matching ones? So it will be very slow API call...

vtnerd:
@SChernykh it performs a LMDB range search to limit the number of transactions scanned. The actual number scanned should depend on how many leading bits are used for the search. Perhaps a runtime argument that specifies a minimum number of bits is needed to keep public nodes from being spammed and bogged down.

ACK-J:
Performance:
```
# Find transactions that end with the characters "aea473c43708aa50b2c9eaf0e441aa209afc9b43458fb09e" --> 1 results in less than 0.1 second
time curl http://127.0.0.1:18081/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"get_txids_loose","params":{"txid_template":"0000000000000000aea473c43708aa50b2c9eaf0e441aa209afc9b43458fb09e","num_matching_bits":192}}' -H Content-Type:application/json
real	   0m0.009s
user   0m0.009s
sys	   0m0.000s

# Find transactions that end with the characters "09e" --> 8,455 results in 0.1 seconds
time curl http://127.0.0.1:18081/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"get_txids_loose","params":{"txid_template":"000000000000000000000000000000000000000000000000000000000000009e","num_matching_bits":12}}' -H Content-Type:application/json
real	   0m0.134s
user   0m0.012s
sys	   0m0.017s
```

hyc:
Yeah, range searches are pretty easy for LMDB, nothing to worry about. Would be pretty ugly in many other DB engines.

SChernykh:
Good. But minimum number of bits should be restricted because another concern is too much data to return (traffic). The number of bits should be limited to return no more than 1000 transactions, I think.

ACK-J:
A minimum of 10 bits is enforced and a user can specify the maximum number of transactions returned using the max_num_txs parameter.

selsta:
> Perhaps a runtime argument that specifies a minimum number of bits is needed to keep public nodes from being spammed and bogged down.

Agree, I would also like a more aggressive limit for restricted RPC nodes, or even a command line flag for it so that this can't be abused by attackers to DoS a public node.

jeffro256:
I used the default `MAX_RESTRICTED_GLOBAL_FAKE_OUTS_COUNT=5000` as a maximum value for the TXID range search while in restricted mode. This value is the maximum number of entries returned by `gets_outs{.bin}`. Requesting output information uses more bandwidth per entry and the entries aren't necessarily located next to each other in the database, so I thought that this seemed like a reasonable limit.

tobtoht (approved):

PR #8965, opened 2023-08-01
core_rpc_server: silence unused  warnings

jeffro256:


vtnerd (approved):

src/rpc/core_rpc_server.cpp
```diff
@@ -2356,6 +2356,7 @@ namespace cryptonote
   bool core_rpc_server::use_bootstrap_daemon_if_necessary(const invoke_http_mode &mode, const std::string &command_name, const typename COMMAND_TYPE::request& req, typename COMMAND_TYPE::response& res, bool &r)
   {
     res.untrusted = false;
+    r = false;
```
selsta, line 2359: Can you write it like this? Or maybe add a comment that this is silencing an unused warning to avoid confusion.

```suggestion
    (void)r;
```

We use

jeffro256: I wrote it like this since it's good code practice to initialize output parameters anyways, but I could be convinced otherwise

selsta (approved):

hinto-janai (approved):

PR #8970, opened 2023-08-11
Read interface for new serialization system

vtnerd:
This is the read interface for my proposed serialization system, along with the remainder of the (major) adapters/wrappers that were not in the write interface review. After this review are the implementations (epee, json, msgpack), and then the changes needed to the existing code.

### DIFF
--- a/contrib/epee/include/serialization/wire/adapted/static_vector.h
+++ b/contrib/epee/include/serialization/wire/adapted/static_vector.h
@@ -1,0 +1,91 @@
+// Copyright (c) 2022, The Monero Project
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <cstdint>
+#include <boost/container/static_vector.hpp>
+#include <boost/utility/string_ref.hpp>
+
+#include "serialization/wire/read.h"
+
+namespace wire
+{
+  // enable writing of static vector arrays
+  template<typename T, std::size_t N>
+  struct is_array<boost::container::static_vector<T, N>>
+    : std::true_type
+  {};
+  
+  // `static_vector`s of `char` and `uint8_t` are not arrays
+  template<std::size_t N>
+  struct is_array<boost::container::static_vector<char, N>>
+    : std::false_type
+  {};
+  template<std::size_t N>
+  struct is_array<boost::container::static_vector<std::uint8_t, N>>
+    : std::false_type
+  {};
+
+  // `static_vector` can be used without specialized macro for every type, it provides max element count
+  template<typename R, typename T, std::size_t N>
+  inline void read_bytes(R& source, boost::container::static_vector<T, N>& dest)
+  {
+    wire_read::array(source, dest, min_element_size<0>{}, max_element_count<N>{});
+  }
+
+  /* `static_vector` never allocates, so it is useful for reading small variable
+     sized strings or binary data with a known fixed max. `char` and
+     `std::uint8_t` are not valid types for arrays in this design anyway
+     (because its clearly less efficient in every encoding scheme).  */
+
+  template<typename R, std::size_t N>
+  inline void read_bytes(R& source, boost::container::static_vector<char, N>& dest)
+  {
+    dest.resize(N);
+    dest.resize(source.string(epee::to_mut_span(dest), /* exact= */ false));
+  }
+
+  template<typename W, std::size_t N>
+  inline void write_bytes(W& dest, const boost::container::static_vector<char, N>& source)
+  {
+    dest.string(boost::string_ref{source.data(), source.size()});
+  }
+
+  template<typename R, std::size_t N>
+  inline void read_bytes(R& source, boost::container::static_vector<std::uint8_t, N>& dest)
+  {
+    dest.resize(N);
+    dest.resize(source.binary(epee::to_mut_span(dest), /* exact= */ false));
+  }
+
+  template<typename W, std::size_t N>
+  inline void write_bytes(W& dest, const boost::container::static_vector<std::uint8_t, N>& source)
+  {
+    dest.binary(epee::to_span(source));
+  }
+}
--- a/contrib/epee/include/serialization/wire/read.h
+++ b/contrib/epee/include/serialization/wire/read.h
@@ -1,0 +1,547 @@
+// Copyright (c) 2023, The Monero Project
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <cstdint>
+#include <limits>
+#include <string>
+#include <type_traits>
+
+#include "byte_slice.h"
+#include "serialization/wire/epee/fwd.h"
+#include "serialization/wire/error.h"
+#include "serialization/wire/field.h"
+#include "serialization/wire/fwd.h"
+#include "serialization/wire/traits.h"
+#include "span.h"
+
+/*
+  Custom types (e.g. `type` in namespace `ns`) can define an input function by:
+    * `namespace wire { template<> struct is_blob<ns::type> : std::true_type {}; }`
+    * `namespace wire { void read_bytes(reader&, ns::type&); }`
+    * `namespace ns { void read_bytes(wire::reader&, type&); }`
+
+  See `traits.h` for `is_blob` requirements. `read_bytes` function can also
+  specify derived type for faster output (i.e.
+  `namespace ns { void read_bytes(wire::epee_reader&, type&); }`). Using the
+  derived type allows the compiler to de-virtualize and allows for custom
+  functions not defined by base interface. Using base interface allows for
+  multiple formats with minimal instruction count. */
+
+namespace wire
+{
+  //! Interface for converting "wire" (byte) formats to C/C++ objects without a DOM.
+  class reader
+  {
+    std::size_t depth_; //!< Tracks number of recursive objects and arrays
+
+    //! \throw wire::exception if max depth (which is exclusive) is reached
+    void increment_depth();
+    //! \throw std::logic_error if already `depth() == 0`.
+    void decrement_depth();
+
+    /*! \param min_element_size of each array element in any format - if known.
+          Derived types with explicit element count should verify available
+          space, and throw a `wire::exception` on issues.
+        \throw wire::exception if next value not array
+        \throw wire::exception if not enough bytes for all array elements
+          (with epee/msgpack which has specified number of elements).
+        \return Number of values to read before calling `is_array_end()`. */
+    virtual std::size_t do_start_array(std::size_t min_element_size) = 0;
+
+    //! \throw wire::exception if not object begin. \return State to be given to `key(...)` function.
+    virtual std::size_t do_start_object() = 0;
+
+  protected:
+    epee::span<const std::uint8_t> remaining_; //!< Derived class tracks unprocessed bytes here
+
+    reader(const epee::span<const std::uint8_t> remaining) noexcept
+      : depth_(0), remaining_(remaining)
+    {}
+
+    reader(const reader&) = delete;
+    reader(reader&&) = delete;
+    reader& operator=(const reader&) = delete;
+    reader& operator=(reader&&) = delete;
+ 
+  public:
+    struct key_map
+    {
+      const char* name;
+      unsigned id; //<! For integer key formats;
+    };
+
+    //! \return Maximum read depth (exclusive) for both objects and arrays before erroring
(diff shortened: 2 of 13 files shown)


jeffro256:
What will `basic_value` be used for?

contrib/epee/include/serialization/wire/read.h
```diff
@@ -0,0 +1,532 @@
...
+  class reader
+  {
+    std::size_t depth_; //!< Tracks number of recursive objects and arrays
+
+  protected:
+    epee::span<const std::uint8_t> remaining_; //!< Derived class tracks unprocessed bytes here
+
+    //! \throw wire::exception if max depth is reached
+    void increment_depth();
+    void decrement_depth() noexcept { --depth_; }
```
jeffro256, line 68: Protected method may cause private field to underflow

contrib/epee/include/serialization/wire/wrapper/array.h
```diff
@@ -0,0 +1,161 @@
...
+    const array_& operator*() const noexcept { return *this; }
+
+    void reset() { get_container().clear(); }
+  };
+
+  //! Nested array case
+  template<typename T, typename C, typename D>
+  struct array_<array_<T, C>, D>
+  {
+    // compute `container_type` and `value_type` recursively
```
jeffro256, line 109: So are we choosing to support nested arrays now ?

vtnerd:
`basic_value` is used by JSON-RPC `id` field, and no one where else currently.

contrib/epee/include/serialization/wire/read.h
vtnerd: Iff there's a mismatch in `increment_depth()` calls. Are you advocating for a runtime check here, to guard against that too?

contrib/epee/include/serialization/wire/wrapper/array.h
vtnerd: I thought we always supported nested arrays? This occurs in the ZeroMQ code, and perhaps elsewhere. You can "get-around" this by putting a class in-between like:
```c++
struct nested { std::vector<uint32_t> inner; }
std::vector<nested> outer;
```
instead of
```c++
std::vector<std::vector<uint32_t>>
```
The complexity of the nested case is a little higher than the rest of the code, so I can understand if you think it should be removed.

contrib/epee/include/serialization/wire/read.h
jeffro256: I think the best design would be to make `increment_depth` and `decrement_depth` private, add a runtime check to `decrement_depth`, and then have `start_object` and `start_array` both call `increment_depth`, just like how `end_object` and `end_array` call `decrement_depth`.

This is because a `reader` implementation or any buggy serialization code for any type could cause this value to underflow. Also, this value merely acts (from my PoV) as a DDoS prevention measure. I can't think of when the actual value of `depth_` by itself would be useful to a format implementation. For this reason, I think it would be cleaner and more stable to hide the depth value, and have subclasses call the parent `start/end_object/array` methods as appropriate.

jeffro256: Also, if there happens to be different structural limits we want to impose (e.g. only on objects) and we can do it in this base class, if we force the subclasses to go thru `start/end_object/array` methods, it will be easier to make those adjustments correctly.

contrib/epee/include/serialization/wire/read.h
```diff
@@ -0,0 +1,532 @@
...
+    virtual std::size_t binary(epee::span<std::uint8_t> dest, const bool exact) = 0;
+
+    /*! \param min_element_size of each array element in any format - if known.
+          Derived types with explicit element count should verify available
+          space, and throw a `wire::exception` on issues.
+        \throw wire::exception if next value not array
+        \throw wire::exception if not enough bytes for all array elements
+          (with epee/msgpack which has specified number of elements).
+        \return Number of values to read before calling `is_array_end()`. */
+    virtual std::size_t start_array(std::size_t min_element_size) = 0;
```
jeffro256, line 145: What is this supposed to return for JSON, CBOR, or any other format with indefinite array lengths?

contrib/epee/include/serialization/wire/read.h
```diff
@@ -0,0 +1,532 @@
...
+
+    //! \return True if there is another element to read.
+    virtual bool is_array_end(std::size_t count) = 0;
+
+    void end_array() noexcept { decrement_depth(); }
```
jeffro256, line 150: I would mark this as `virtual` in case it is ever needed to be overridden by a subclass. Since we usually use `reader` in a templated context, the compiler should usually be able to de-virtualize this method most of the time anyways. Same with `end_object`.

contrib/epee/include/serialization/wire/read.h
vtnerd: They `return 0`, and `is_array_end(size_t)` is called for each entry until the end is found. This is why `is_array_end(size_t)` is `virtual` and `end_array()` is not - the derived classes don't need to override the default.

Msgpack and epee return the number of items, and `is_array_end(size_t)` is only invoked once.

contrib/epee/include/serialization/wire/read.h
vtnerd: I haven't had a need to `override` for msgpack, epee, protobuf, or JSON (and unlikely for CBOR either). `is_array_end(size_t)` and `key(...)` are both `virtual`, and are used to determine the end whereas these functions are simply used by the base class to track placement in the tree.

vtnerd: Also, this is trivial to change this if a future format needs to override it.

contrib/epee/include/serialization/wire/read.h
vtnerd: I'm leaning towards making this change, but I'll first describe how this is handled in the current design. Its not well documented, a short-coming that needs fixing (or else dropped).

The `from_bytes` template method has the following:
```c++
       R in{std::forward<T>(source)};
       bytes(in, dest);
       in.check_complete();
```
this requires that every reader type have a method called `check_complete()`. This method checks to make sure that the tree is "complete":
```c++
   void json_reader::check_complete() const
   {
     if (depth())
       WIRE_DLOG_THROW(error::rapidjson_e(rapidjson::kParseErrorUnspecificSyntaxError), "Unexpected end");
   }
```
So the potential underflow is intentional - either the next call to `increment_depth()` detects an overflow case or the `check_complete()` detects a non-zero depth.

AFAIK, this achieves the same thing, but with one-less branch per call to `decrement_depth()`. It probably looks better to have the check though, and the extra check is trivial in the design.
---------
I left `increment_depth()` and `decrement_depth()` as `protected` so that "skipping" a field could call them. I _think_ this has never been necessary, as `start_object()` and `end_object()`, etc, have been used instead (so at a minimum I would make the functions `private`.

vtnerd: After further thought, I will just add the check. It's possible to call `end_object()` enough times to rollback to `0`, which is unintentional, so a runtime-check is needed. I'll also move them to `private`, as I can't recall using them in the derived classes.

vtnerd: I added the checks in the other review, but _might_ revert it. The fallout is that `do_start_object()` and `do_start_array(size_t)` can no longer be de-virtualized.

vtnerd:
@jeffro256 Earlier I mentioned protobuf, and I forgot completely that this interface cannot support reading of protobuf arrays (without some modifications to the interface). Write support should be possible (although I haven't tried to implement it yet). I've otherwise implemented reading/writing of everything else protobuf.

I think its better to introduce the necessary changes when/if protobuf is added as potential format. A couple of people have asked for it years ago, but its pointless unless we can auto-generate schemas from this interface (another thing I failed to figure out - at least yet).

Otherwise, epee_binary, json, msgpack are all implemented completely, and CBOR should be possible as well.

-------

### Reading Protobuf Arrays
Reading protobuf arrays is difficult because a field within an object appears multiple times, and does not have to be contiguous. The binary format looks something like:
```
  {"1": 100, "2": "string", "1": 300}
```
which specifies an integer array of length 2: `[100, 300]`.

vtnerd:
@jeffro256 I finally updated this PR. I thought it was done previously, but I didn't merge in the requested changes for the read interface.

Changes:
  * Rebased
  * Changes requested by @jeffro256 
  * Add API support for reading integer/msgpack keys

vtnerd:
Rebased (although technically not needed), and merged changes from other review (due to rebasing changes).

contrib/epee/include/serialization/wire/read.h
```diff
@@ -0,0 +1,532 @@
...
+    bool more = count;
+    const std::size_t start_bytes = source.remaining().size();
+    while (more || !source.is_array_end(count))
+    {
+      // check for json/cbor formats
+      if (source.delimited_arrays() && max_element_count <= dest.size())
+        throw_exception(wire::error::schema::array_max_element, "", nullptr);
+
+      wire_read::array_insert(source, dest);
+      --count;
```
selsta, line 364: Can this underflow?

contrib/epee/include/serialization/wire/read.h
```diff
@@ -0,0 +1,532 @@
...
+      more &= bool(count);
+
+      if (((start_bytes - source.remaining().size()) / dest.size()) < min_element_size)
```
selsta, line 367: Could `array_unchecked` track an `elements_read` counter instead of using `dest.size()`, or is "size grows once per decoded element" intended to be part of the array container contract? The `emplace_hint` insertion path would not necessarily guarantee that for set/map-like containers.

contrib/epee/src/wire/read.cpp
```diff
@@ -0,0 +1,69 @@
...
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "serialization/wire/read.h"
+
+#include <stdexcept>
+
+void wire::reader::increment_depth()
+{
+  if (++depth_ == max_read_depth())
```
selsta, line 34:
```suggestion
  if (max_read_depth() < ++depth_)
```

or the `max_read_depth()` comment should say it's exclusive

contrib/epee/include/serialization/wire/adapted/static_vector.h
```diff
@@ -0,0 +1,91 @@
...
+  /* `static_vector` never allocates, so it is useful for reading small variable
+     sized strings or binary data with a known fixed max. `char` and
+     `std::uint8_t` are not valid types for arrays in this design anyway
+     (because its clearly less efficient in every encoding scheme).  */
+
+  template<typename R, std::size_t N>
+  inline void read_bytes(R& source, boost::container::static_vector<char, N>& dest)
+  {
+    dest.resize(N);
+    dest.resize(source.string(epee::to_mut_span(dest), /* exact= */ false));
```
selsta, line 70: Is it okay if this throws?

contrib/epee/include/serialization/wire/read.h
```diff
@@ -0,0 +1,532 @@
...
+#include "serialization/wire/field.h"
+#include "serialization/wire/fwd.h"
+#include "serialization/wire/traits.h"
+#include "span.h"
+
+/*
+  Custom types (e.g. `type` in namespace `ns`) can define an input function by:
+    * `namespace wire { template<> struct is_blob<ns::type> : std::true_type {}; }`
+    * `namespace wire { void read_bytes(writer&, ns::type&); }`
+    * `namespace ns { void read_bytes(wire::writer&, type&); }`
```
selsta:
```suggestion
    * `namespace wire { void read_bytes(reader&, ns::type&); }`
    * `namespace ns { void read_bytes(wire::reader&, type&); }`
```

contrib/epee/include/serialization/wire/read.h
vtnerd: Yes this can underflow, but this is intentionally allowed. JSON returns `0` for every `start_array` call, and uses `is_array_end` to signal the end.

An extra guard is added so that once `more` is set to `false`, it remains false forever even after `count` underflows.

contrib/epee/include/serialization/wire/read.h
vtnerd: > or is "size grows once per decoded element" intended to be part of the array container contract

This. If the map insertion fails due to a duplicate key, no additional memory was needed, and therefore the ratio between packed and unpacked size remains below `min_element_size`.

contrib/epee/src/wire/read.cpp
vtnerd: It was written this way to protect against `max_read_depth()` being `size_t::max` (which is arguably unrealistic).

I will tweak the documentation.

contrib/epee/include/serialization/wire/read.h
vtnerd: Fixed.

contrib/epee/include/serialization/wire/adapted/static_vector.h
vtnerd: This should be acceptable, and only possible if `reader` violates the specification of the interface.e

selsta (approved):

contrib/epee/include/serialization/wire/read.h
```diff
@@ -0,0 +1,547 @@
...
+    bytes(source, dest.get_value());
+  }
+
+  template<typename R, typename T>
+  inline void unpack_field(R& source, wire::field_<T, false>& dest)
+  {
+    if (!bool(dest.get_value()))
+      dest.get_value().emplace();
+    bytes(source, *dest.get_value());
+  }
```
thomasbuilds, line 412: `unpack_field` is pinned to `I=0` (`field_<T, true>`/`field_<T, false>`), but `reset_field` above (L386–398) takes `unsigned I`. Since `set_mapping` propagates `field_.id()` and `try_read` calls `unpack_field` at L455, a `WIRE_FIELD_ID(<non-zero>, …)` field produces `field_<…, true, N>` and matches no overload.

```suggestion
  template<typename R, typename T, unsigned I>
  inline void unpack_field(R& source, wire::field_<T, true, I>& dest)
  {
    bytes(source, dest.get_value());
  }

  template<typename R, typename T, unsigned I>
  inline void unpack_field(R& source, wire::field_<T, false, I>& dest)
  {
    if (!bool(dest.get_value()))
      dest.get_value().emplace();
    bytes(source, *dest.get_value());
  }
```

The same `I=0` pinning affects the write path: `wire_write::field()` in `write.h` and `available()` in `field.h`.

contrib/epee/include/serialization/wire/read.h
```diff
@@ -0,0 +1,547 @@
...
+
+#pragma once
+
+#include <cstdint>
+#include <limits>
+#include <string>
+#include <type_traits>
+
+#include "byte_slice.h"
+#include "serialization/wire/epee/fwd.h"
```
thomasbuilds, line 36: This includes `serialization/wire/epee/fwd.h`, which doesn't exist in any commit of this repo, so read.h can't be `#include`d as-is. It also appears unused

Also the three new sources `wire/asio.cpp`, `wire/basic_value.cpp`, `wire/read.cpp` aren't referenced by any CMakeLists ?

contrib/epee/include/serialization/wire/read.h
```diff
@@ -0,0 +1,547 @@
...
+    //! \throw wire::exception if parsing is incomplete.
+    virtual void check_complete() const = 0;
+
+    //! \throw wire::exception if array, object, or end of stream.
+    virtual basic_value basic() = 0;
+
+    //! \throw wire::exception if next value not a boolean.
+    virtual bool boolean() = 0;
+
+    //! \throw wire::expception if next value not an integer.
```
thomasbuilds, line 123:
```suggestion
    //! \throw wire::exception if next value not an integer.
```

contrib/epee/include/serialization/wire/wrapper/array.h
```diff
@@ -0,0 +1,161 @@
...
+    if write-only.
+
+    This wrapper meets the requirements for an optional field; `wire::field`
+    and `wire::optional_field` determine whether an empty array must be
+    encoded on the wire. Historically, empty arrays were always omitted on
+    the wire (a defacto optional field).
+
+    The `is_array` trait can also be used, but is default treated as an optional
+    field. The trait `is_optional_on_empty` traits can be specialized to disable
+    the optional on empty behavior. See `wire/traits.h` for more ifnormation
```
thomasbuilds, line 55:
```suggestion
    the optional on empty behavior. See `wire/traits.h` for more information
```

thomasbuilds (commented):
we don't want me creating more fix typos PRs ;)

contrib/epee/src/wire/basic_value.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <stdexcept>
+#include "serialization/wire/read.h"
+#include "serialization/wire/write.h"
+
+namespace wire
+{
+  static void write_bytes(writer& dest, const std::nullptr_t&)
+  {
+    throw std::logic_error{"nullptr output not yet defined for wire::writer"};
+  }
```
thomasbuilds, line 40: Writing a `basic_value` holding `nullptr` throws `std::logic_error`. That isn't a `wire::exception`, so it escapes `wire_write::to_bytes`'s error-code path: it's caught only by `catch(...)`, which rethrows instead of returning an error code.

contrib/epee/include/serialization/wire/read.h
vtnerd: Yes, if something actually used the `I` parameter, compilation would break here.

I'm currently moving away from the proposal to add integer keys, so the result will be an updated diff dropping the `I` entirely. It does provide some reasonable space savings, but the pack/unpack difference isn't as great as I thought it would be. "Finding" the correct value for a given input seems to dominate the CPU cycles. And worst of all use of the integer keys will likely frustrate downstream projects trying to determine their meaning from our source.

This does mean that this interface will not support Protobuf, unless the integer keys were re-introduced. Its not difficult to re-add them, provided you're clever in C++ templates.

contrib/epee/include/serialization/wire/read.h
vtnerd: Yes, must've been an unintentional exclusion when creating this diff.

contrib/epee/src/wire/basic_value.cpp
vtnerd: Yes, this was intentionally designed this way, but perhaps undesirable. Thinking whether I should move to a wire exception here.

thomasbuilds: In `replace_serialization` the response id is a `WIRE_OPTIONAL_FIELD`, so a `nullptr` `basic_value` omits the member and this throw is never reached. But omitting the id violates JSON-RPC 2.0: the response id is REQUIRED, and "If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null". Master currently emits `"id": null` on that path: `FullMessage::getResponse` always writes the id, and `BAD_JSON` passes a default-constructed (null) `rapidjson::Value`, so the port regresses this. A compliant response needs a required field holding null, which is exactly the case that throws here, and `json_writer` has no null output (`json_reader::basic()` already maps `null` to `basic_value{nullptr}`). So rather than changing the exception type, `wire::writer` probably needs null output (at least for JSON), keeping the throw as a `wire::exception` for formats without a null representation like epee binary.

contrib/epee/include/serialization/wire/wrapper/array_blob.h
```diff
@@ -0,0 +1,95 @@
...
+    Additional concept requirements for `container_type` when writing:
+      * must work with foreach loop (`std::begin` and `std::end`).
+      * must have `size()` method that returns number of elements. */
+  template<typename T>
+  struct array_as_blob_
+  {
+    using container_type = unwrap_reference_t<T>;
+    using value_type = typename container_type::value_type;
+    static constexpr std::size_t value_size() noexcept { return sizeof(value_type); }
+    static_assert(std::is_pod<value_type>::value, "container value must be POD");
```
selsta, line 67: This should also require no padding.

contrib/epee/include/serialization/wire/adapted/static_vector.h
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  template<typename W, std::size_t N>
+  inline void write_bytes(W& dest, const boost::container::static_vector<char, N>& source)
+  {
+    dest.string(boost::string_ref{source.data(), source.size()});
```
selsta, line 76: needs to be `std::string_view`

PR #8974, opened 2023-08-17
wallet2: fix version check at hf version 1

j-berman:
Fixes:
- create a new wallet
- set restore height to 0
- wallet fails to sync with `Error: refresh failed: refresh error: Unexpected hard fork version v1 at height 0. Make sure the node you are connected to is running the latest version. Blocks received: 0`

Synced a wallet restored from 0 with this change.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -3299,7 +3299,7 @@ void check_block_hard_fork_version(cryptonote::network_type nettype, uint8_t hf_
     if (wallet_hard_forks[fork_index].version == hf_version)
       break;
   THROW_WALLET_EXCEPTION_IF(fork_index == wallet_num_hard_forks, error::wallet_internal_error, "Fork not found in table");
-  uint64_t start_height = wallet_hard_forks[fork_index].height;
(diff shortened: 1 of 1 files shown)


selsta (approved):
Please also open against release.

j-berman:
Release doesn't have the issue; it was a regression in this commit on master c61d33e24cd0e79a6f538e02ebfdf11d1ab6bfbe

j-berman:
I figure it's reasonable to consider the genesis block a special case in that table; plus it seems risky to change a table that's been there so long/used in consensus etc. I see your view though, I'd change it if you push me on it (and will make sure it's safe to do)

moneromooo-monero:
Townforge has it at 0. Looking at the git log, it was part of the original commit that forked off Monero in 2019. Of course, I don't remember why I changed it, so it's plausible there were other changes that should go with it. I don't particularly mind which patch you go with if you think the bodge is safer.

j-berman:
Looked through other parts that touch the table and I would rather keep the bodge as is

jeffro256 (approved):

plowsof (approved):

PR #8978, opened 2023-08-21
sync_info format peer list

maogo:


vtnerd:
How does this formatting work with smaller screens (admitting I haven't looked at code too closely)?

maogo:
yeah, that depends upon the screen resolution and visible area of terminal . When the horizontal width is not enough that table won't display correctly.       

If  i2p/tor address joined:
   with resolution 1920 * 1080 works fine , not sure lower resolution, like 1280 * 1024 maybe  No.

vtnerd:
It's temping to not allow the patch on that basis, interested to know what others think.

Gingeropolous:
could make it an option for the command
```
sync_info pretty
```

or if commands can't get options, just a new command (one thats not included in the default output of help because well that thing is already cluttered).

```
sync_info_pretty
```

src/daemon/rpc_command_executor.cpp
```diff
@@ -2274,31 +2274,58 @@ bool t_rpc_command_executor::sync_info()
...
+        }
+    }
+    longest_host_width += 8;
     tools::success_msg_writer() << "Downloading at " << current_download << " kB/s";
     if (res.next_needed_pruning_seed)
       tools::success_msg_writer() << "Next needed pruning seed: " << res.next_needed_pruning_seed;
 
     tools::success_msg_writer() << std::to_string(res.peers.size()) << " peers";
-    tools::success_msg_writer() << "Remote Host                        Peer_ID   State   Prune_Seed          Height  DL kB/s, Queued Blocks / MB";
+    std::string table_line = "+" + std::string(longest_host_width - 1,'-')+ "+" + std::string(20 - 1,'-') + "+" + std::string(20 - 1,'-') + "+" + std::string(15 - 1,'-') + "+" + std::string(15 - 1,'-') + "+" + std::string(15 - 1,'-') + "+" + std::string(15 - 1,'-') + "+" + std::string(15 - 1,'-') + "+";
```
iamamyth, line 2292: I find the style here quite brittle: To change a column width, you have to edit three separate lines, and they don't even use the same constant (first two are X, last one is X - 1). It'd be way easier to read, and less error-prone, to use a structured array/vector here (e.g. heading + width, where width defaults to size, and can be adjusted for columns with wide content, like `remote host`); such an approach would also make it much easier to wrap the table across lines to fit the terminal, if someone wanted.

maogo:
Hard-code here is bad ,  A well-encapsulated class for reuse is a better way, I am not c++ guy , If someone wants to make a new PR that's better .

Issue #8980, opened 2023-08-31
simplewallet: set subaddress-lookahead does not work after wallet creation


benevanoff:
# Description

While attempting to resolve #8954, I first tried to understand fully how it is implemented in the cli since woodser linked to the code snippet where it shows that there are class member variables `m_subaddress_lookahead_major` and `m_subaddress_lookahead_minor` that gets updated and saved to the wallet cache, but this made me realize that the `m_subaddresses` table of `m_wallet` is never actually updated and prompted me to test it, finding that it does not actually make the wallet "look ahead" any further.

# User Impact

Users that have used many subaddresses may not be able to find their e-notes after restoring from their seed when adjusting the lookahead with `set subaddress-lookahead`

# How to Reproduce

1. create a new wallet: `wallet_1`
2. run `set` - this will show the lookahead set to the default, 50:200
3. `set subaddress-lookahead 50:300`
4. run `set` again - this will now show 50:300
5. (optional) close and reopen the wallet then run `set` again to confirm the change has persisted
6. export the wallet seed and create a new wallet by restoring from the seed to create `wallet_2`
7. in `wallet_2`, run `address new` 300 times
8. run `address all`
9. copy one of the subaddresses from the 210-290 range
10. send some moneroj from a different wallet to that address from step 9.
11. wait until the transaction is confirmed
12. open and refresh `wallet_1` and `wallet_2`
13. you will see a new transaction in `wallet_2` but not `wallet_1`
14. run `set` in each wallet one more time to confirm that the subaddress-lookahead should be the same (they will both display 50:300) but they are in fact different, evidenced by the balance difference indicating that `wallet_1`'s subaddress pubkey hashtable has not been properly updated to "lookahead" past the 200th subaddress to the 300th subaddress.

# How to Fix

Modify `wallet2::set_subaddress_lookahead` to actually expand/shrink the `m_subaddresses` class member variable based on the input.

Issue #8954, opened 2023-07-18
Support subaddress lookahead in monero-wallet-rpc


woodser:
This issue requests setting the wallet's subaddress lookahead when creating wallets with monero-wallet-rpc.

A new parameter, `subaddress_lookahead`, can be added to these methods in wallet_rpc_server.cpp:

- `create_wallet`
- `restore_deterministic_wallet`
- `generate_from_keys`

By calling `m_wallet->set_subaddress_lookahead()`.

jeffro256:
IMO, it be more preferable to create a new RPC endpoint whose sole purpose is to set the lookahead values to minimize code replication in the wallet rpc server.

woodser:
> IMO, it would be more preferable to create a new RPC endpoint whose sole purpose is to set the lookahead values to minimize code replication in the wallet rpc server.

That sounds good to me, to create a new RPC call, e.g. `set_subaddress_lookahead`, as long as it works after wallet creation. I recall there may have been some issue related to that, but might be fixed now. I updated the issue description.

SamsungGalaxyPlayer:
It would be very useful to additionally support setting a subaddress lookahead as a parameter when restoring or creating a wallet with monero-wallet-rpc. This applies to the following commands:


* restore_deterministic_wallet
* create_wallet
* generate_from_keys

PR #8981, opened 2023-09-03
wallet2: ensure subaddress keys table is at least size of requested lookahead

benevanoff:
A fix for #8980

Basically the current behavior is that if you make a wallet with the default settings in the cli and then do `set subaddress-lookahead 50:300`, then the subaddress pubkey table hasnt actually expanded and you aren't looking any further when you scan. Instead, the table will get expanded to this look ahead only after either a new enote is found with the old lookahead or if the user requests a new subaddress with `address new`.

This patch makes sure that the wallet adjusts the pubkey table immediately after sending the set command if an increase in the lookahead is requested.


src/wallet/wallet2.cpp
```diff
@@ -1909,6 +1909,23 @@ void wallet2::set_subaddress_lookahead(size_t major, size_t minor)
...
   THROW_WALLET_EXCEPTION_IF(minor > 0xffffffff, error::wallet_internal_error, "Subaddress minor lookahead is too large");
+
+  if (major > m_subaddress_lookahead_major || minor > m_subaddress_lookahead_minor) // if increasing the lookahead
+  {
+    // then generate new subaddress pubkeys and add them to m_subaddresses table
+    std::unordered_map<cryptonote::subaddress_index, crypto::public_key> subaddresses_inverted;
+    for (const auto& pair : m_subaddresses) {
+      subaddresses_inverted[pair.second] = pair.first; // inverting the keys and values makes it easier to lookup indices
+    }
+    for (uint32_t i = 0; i < m_subaddress_labels.size()+major-1; i++) { // m_subaddress_labels tells us how many accounts the user has actually requested without lookahead
```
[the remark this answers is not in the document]
benevanoff: Just slightly outside of the diff, the first thing this function does is assert that major is greater than zero and `.size` returns an unsigned int :) I still changed it

benevanoff:
You're right, thank you. I think the expressions I used this time only make as many calls to `create_one_off_subaddress` as necessary, but yes afaik the map structure doesn't store duplicate keys and operator[] will write over any old value if the key was already there so there.

benevanoff:
OK, I made a mess of the changelog but it should be ready for review now :)

tests/unit_tests/subaddress.cpp
```diff
@@ -101,3 +101,29 @@ TEST_F(WalletSubaddress, OutOfBoundsIndexes)
...
+  subaddr = w1.get_subaddress(test_idx);
+  EXPECT_NE(boost::none, w1.get_subaddress_index(subaddr));
+  // next test expanding the minor lookahead
+  w1.set_subaddress_lookahead(100, 300);
+  EXPECT_EQ(100, w1.get_subaddress_lookahead().first);
+  EXPECT_EQ(300, w1.get_subaddress_lookahead().second);
+  test_idx = {100, 299};
+  subaddr = w1.get_subaddress(test_idx);
+  EXPECT_NE(boost::none, w1.get_subaddress_index(subaddr));
+}
```
selsta, line 129: Minor but please add a newline here.

benevanoff: Got it 👍

benevanoff:
I've updated the PR to address #8954 as well. I thought about opening a separate PR but figured I would just lump it in here since the RPC endpoint depends on these changes to wallet2. I hope that's alright, but I can still break them apart if y'all would prefer.

selsta (commented):
‎I think WALLET_RPC_VERSION_MINOR needs to be bumped

woodser:
Will be nice to have this implemented. Thanks.

woodser:
>  the change does not persist to disk when the wallet closes and changing the wallet-rpc program to save those cached values looks like kinda a pain but I hope that the requestor finds that something is better than nothing for now.

In that case, `/set_subaddress_lookahead` would need to be called each time the wallet is opened in order to scan all subaddresses in excess of 250 (if I recall the default correctly)?

And just confirming, there's no other way to set the wallet's subaddress lookahead through monero-wallet-rpc to avoid that, right?

Something is better than nothing in that case since it can be manually set when opened, but of course not ideal.

SamsungGalaxyPlayer:
Related to https://github.com/monero-project/monero/issues/8954, which is a request to support setting the subaddress lookahead when creating or restoring wallets.

selsta:
Continuing in #9953

PR #8987, opened 2023-09-10
unit_tests: set data dir relative to exe & add log-level arg

jeffro256:
Setting the data directory this way is nice when the unit tests write to files while running (see new `store_to()` PR). Otherwise, files in the data directory will show up on git and you won't be able to have copies for different branches when running tests from root source directory. I also have wanted `--log-level` in the unit tests for a while now.


0xFFFC0000 (approved):
Thank you. IMHO this is much cleaner than what we have.

selsta (approved):

PR #8989, opened 2023-09-10
[seraphis] cryptonote_base: add account generators

UkoeHB:
## Summary

This is a PR in my 'upstreaming seraphis_lib project', it is not used anywhere yet.

- Adds `account_generators.h` to `cryptonote_basic`.

## Explanation

Multisig account generation is the same for ringct and seraphis, so it's best to reuse all that code if possible. The only difference is the base generators used. Refactoring multisig accounts to depend on `account_generator_era` will allow seraphis to be easily supported.

## Future PRs
- Refactor multisig accounts to use `account_generator_era`.
- Introduce seraphis generators.

### DIFF
--- a/src/cryptonote_basic/account_generators.h
+++ b/src/cryptonote_basic/account_generators.h
@@ -1,0 +1,71 @@
+// Copyright (c) 2021, The Monero Project
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+// 
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+// 
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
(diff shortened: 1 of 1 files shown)


vtnerd (approved):

src/cryptonote_basic/account_generators.h
```diff
@@ -0,0 +1,71 @@
...
+{
+  if (era == account_generator_era::cryptonote)
+    return crypto::get_G();
+  else
+    return crypto::null_pkey;  //error
+}
+
+inline crypto::public_key get_secondary_generator(const account_generator_era era)
+{
+  if (era == account_generator_era::cryptonote)
```
jeffro256, line 60: Are they both meant to return G or is one supposed to return H?

jeffro256: Sweet, just wanted to double check since Seraphis uses two different generators (I'm still woefully ignorant of the Crptonote multisig math).

jeffro256 (approved):
Matches code in `seraphis_lib`, and looks to exhibit expected behavior while missing Seraphis generator options.

PR #8996, opened 2023-09-19
Add SSL support to P2P

vtnerd:
First, my apologies to @moneromooo-monero for NACKing his/her SSL proposal years ago, only for me to bring it back.

--------------------------------------

This does exactly as the title suggests, with options to disable P2P encryption entirely, and an option to re-use a SSL certificate between "runs". The default generates a new SSL certificate each `monerod` run, so that the node cannot be tracked across IP address changes.

Nodes _do not_ trust encryption information from peers. Instead, every peer is assumed to be in autodetect mode, unless overridden on the CLI or via handshake/ping messages directly from the peer.

**Possibly bad**:
If a node chooses to re-use SSL certificates, a change in certificates will cause connection failures until the node is removed from the white+gray lists OR the node makes a direct connection and provides the new SSL certificate.

-------------------------------

EDIT:
I will also attempt some unit test changes; there is enough to review here that I expect this diff to be up a while.

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -984,7 +984,7 @@ namespace net_utils
       boost::uuids::random_generator()(),
       *real_remote,
       is_income,
-      connection_basic::m_ssl_support == ssl_support_t::e_ssl_support_enabled
+      connection_basic::m_ssl_support
     );
     m_host = real_remote->host_str();
     try { host_count(1); } catch(...) { /* ignore */ }
@@ -1742,7 +1742,7 @@ namespace net_utils
   }
   //---------------------------------------------------------------------------------
   template<class t_protocol_handler>
-  typename boosted_tcp_server<t_protocol_handler>::try_connect_result_t boosted_tcp_server<t_protocol_handler>::try_connect(connection_ptr new_connection_l, const std::string& adr, const std::string& port, boost::asio::ip::tcp::socket &sock_, const boost::asio::ip::tcp::endpoint &remote_endpoint, const std::string &bind_ip, uint32_t conn_timeout, epee::net_utils::ssl_support_t ssl_support)
+  typename boosted_tcp_server<t_protocol_handler>::try_connect_result_t boosted_tcp_server<t_protocol_handler>::try_connect(connection_ptr new_connection_l, const std::string& adr, const std::string& port, boost::asio::ip::tcp::socket &sock_, const boost::asio::ip::tcp::endpoint &remote_endpoint, const std::string &bind_ip, uint32_t conn_timeout, epee::net_utils::ssl_options_t& ssl_options)
   {
     TRY_ENTRY();
 
@@ -1818,7 +1818,7 @@ namespace net_utils
     {
       // Handshake
       MDEBUG("Handshaking SSL...");
-      if (!new_connection_l->handshake(boost::asio::ssl::stream_base::client))
+      if (!new_connection_l->client_handshake(ssl_options))
       {
         if (ssl_support == epee::net_utils::ssl_support_t::e_ssl_support_autodetect)
         {
@@ -1832,6 +1832,7 @@ namespace net_utils
           sock_.close();
         return CONNECT_FAILURE;
       }
+      new_connection_l->set_ssl_enabled();
     }
 
     return CONNECT_SUCCESS;
@@ -1840,11 +1841,11 @@ namespace net_utils
   }
   //---------------------------------------------------------------------------------
   template<class t_protocol_handler>
-  bool boosted_tcp_server<t_protocol_handler>::connect(const std::string& adr, const std::string& port, uint32_t conn_timeout, t_connection_context& conn_context, const std::string& bind_ip, epee::net_utils::ssl_support_t ssl_support)
+  bool boosted_tcp_server<t_protocol_handler>::connect(const std::string& adr, const std::string& port, uint32_t conn_timeout, t_connection_context& conn_context, const std::string& bind_ip, epee::net_utils::ssl_options_t ssl_options)
   {
     TRY_ENTRY();
 
-    connection_ptr new_connection_l(new connection<t_protocol_handler>(io_context_, m_state, m_connection_type, ssl_support) );
+    connection_ptr new_connection_l(new connection<t_protocol_handler>(io_context_, m_state, m_connection_type, ssl_options.support) );
     connections_mutex.lock();
     connections_.insert(new_connection_l);
     MDEBUG("connections_ size now " << connections_.size());
@@ -1932,24 +1933,22 @@ namespace net_utils
     //boost::asio::ip::tcp::endpoint remote_endpoint(boost::asio::ip::address::from_string(addr.c_str()), port);
     boost::asio::ip::tcp::endpoint remote_endpoint(*iterator);
 
-    auto try_connect_result = try_connect(new_connection_l, adr, port, sock_, remote_endpoint, bind_ip_to_use, conn_timeout, ssl_support);
+    auto try_connect_result = try_connect(new_connection_l, adr, port, sock_, remote_endpoint, bind_ip_to_use, conn_timeout, ssl_options);
     if (try_connect_result == CONNECT_FAILURE)
       return false;
-    if (ssl_support == epee::net_utils::ssl_support_t::e_ssl_support_autodetect && try_connect_result == CONNECT_NO_SSL)
+    if (ssl_options.support == epee::net_utils::ssl_support_t::e_ssl_support_autodetect && try_connect_result == CONNECT_NO_SSL)
     {
       // we connected, but could not connect with SSL, try without
       MERROR("SSL handshake failed on an autodetect connection, reconnecting without SSL");
       new_connection_l->disable_ssl();
-      try_connect_result = try_connect(new_connection_l, adr, port, sock_, remote_endpoint, bind_ip_to_use, conn_timeout, epee::net_utils::ssl_support_t::e_ssl_support_disabled);
+      ssl_options = epee::net_utils::ssl_support_t::e_ssl_support_disabled;
+      try_connect_result = try_connect(new_connection_l, adr, port, sock_, remote_endpoint, bind_ip_to_use, conn_timeout, ssl_options);
       if (try_connect_result != CONNECT_SUCCESS)
         return false;
     }
 
     // start adds the connection to the config object's list, so we don't need to have it locally anymore
-    connections_mutex.lock();
-    connections_.erase(new_connection_l);
-    connections_mutex.unlock();
-    bool r = new_connection_l->start(false, 1 < m_threads_count);
+    bool r = remove_connection(new_connection_l) && new_connection_l->start(false, 1 < m_threads_count);
     if (r)
     {
       new_connection_l->get_context(conn_context);
@@ -1968,10 +1967,10 @@ namespace net_utils
   }
   //---------------------------------------------------------------------------------
   template<class t_protocol_handler> template<class t_callback>
-  bool boosted_tcp_server<t_protocol_handler>::connect_async(const std::string& adr, const std::string& port, const std::chrono::milliseconds conn_timeout, const t_callback &cb, const std::string& bind_ip, epee::net_utils::ssl_support_t ssl_support, t_connection_context&& initial)
+  bool boosted_tcp_server<t_protocol_handler>::connect_async(const std::string& adr, const std::string& port, const std::chrono::milliseconds conn_timeout, const t_callback &cb, const std::string& bind_ip, epee::net_utils::ssl_options_t ssl_options, t_connection_context&& initial)
   {
     TRY_ENTRY();    
-    connection_ptr new_connection_l(new connection<t_protocol_handler>(io_context_, m_state, m_connection_type, ssl_support, std::move(initial)) );
+    connection_ptr new_connection_l(new connection<t_protocol_handler>(io_context_, m_state, m_connection_type, ssl_options.support, std::move(initial)) );
     connections_mutex.lock();
     connections_.insert(new_connection_l);
     MDEBUG("connections_ size now " << connections_.size());
@@ -2048,60 +2047,118 @@ namespace net_utils
         return false;
       }
     }
-    
-    std::shared_ptr<boost::asio::steady_timer> sh_deadline(std::make_shared<boost::asio::steady_timer>(io_context_));
-    //start deadline
-    sh_deadline->expires_after(conn_timeout);
-    sh_deadline->async_wait([=](const boost::system::error_code& error)
-      {
-          if(error != boost::asio::error::operation_aborted) 
-          {
-            _dbg3("Failed to connect to " << adr << ':' << port << ", because of timeout (" << conn_timeout.count() << ")");
-            new_connection_l->socket().close();
-          }
-      });
-    //start async connect
-    sock_.async_connect(remote_endpoint, [=](const boost::system::error_code& ec_)
-      {
-        t_connection_context conn_context = AUTO_VAL_INIT(conn_context);
-        boost::system::error_code ignored_ec;
-        boost::asio::ip::tcp::socket::endpoint_type lep = new_connection_l->socket().local_endpoint(ignored_ec);
-        if(!ec_)
-        {//success
-          if(!sh_deadline->cancel())
-          {
-            cb(conn_context, boost::asio::error::operation_aborted);//this mean that deadline timer already queued callback with cancel operation, rare situation
-          }else
-          {
-            _dbg3("[sock " << new_connection_l->socket().native_handle() << "] Connected success to " << adr << ':' << port <<
-              " from " << lep.address().to_string() << ':' << lep.port());
 
-            // start adds the connection to the config object's list, so we don't need to have it locally anymore
-            connections_mutex.lock();
-            connections_.erase(new_connection_l);
-            connections_mutex.unlock();
-            bool r = new_connection_l->start(false, 1 < m_threads_count);
-            if (r)
-            {
-              new_connection_l->get_context(conn_context);
-              cb(conn_context, ec_);
-            }
-            else
-            {
-              _dbg3("[sock " << new_connection_l->socket().native_handle() << "] Failed to start connection to " << adr << ':' << port);
-              cb(conn_context, boost::asio::error::fault);
-            }
-          }
-        }else
-        {
-          _dbg3("[sock " << new_connection_l->socket().native_handle() << "] Failed to connect to " << adr << ':' << port <<
-            " from " << lep.address().to_string() << ':' << lep.port() << ": " << ec_.message() << ':' << ec_.value());
-          cb(conn_context, ec_);
-        }
-      });
-    return true;
+    ssl_options.configure(new_connection_l->socket_, boost::asio::ssl::stream_base::client);
+    return connect_async_internal(new_connection_l, remote_endpoint, conn_timeout, cb);
     CATCH_ENTRY_L0("boosted_tcp_server<t_protocol_handler>::connect_async", false);
   }
-  
+
+  template<class t_protocol_handler> template<class t_callback>
+  bool boosted_tcp_server<t_protocol_handler>::connect_async_internal(const connection_ptr& new_connection_l, const boost::asio::ip::tcp::endpoint& remote_endpoint, const std::chrono::milliseconds conn_timeout, const t_callback &cb)
+  {
+    if (!new_connection_l)
+      return false;
+
+    TRY_ENTRY();
+
+    const auto on_timer = [=](boost::system::error_code error)
+      {
+        if (error != boost::asio::error::operation_aborted)
+        {
+          _dbg3("Failed to connect to " << remote_endpoint << ", because of timeout (" << conn_timeout.count() << " ms)");
+          new_connection_l->socket().close(error); // ignore errors
+        }
+      };
+
+    auto sh_deadline = std::make_shared<boost::asio::steady_timer>(io_context_);
+    sh_deadline->expires_after(conn_timeout);
+    sh_deadline->async_wait(new_connection_l->wrap(on_timer));
+
+    new_connection_l->socket().async_connect(remote_endpoint, new_connection_l->wrap([=](const boost::system::error_code& ec_)
+      {
+        const auto on_cancel = [=](const boost::system::error_code& error)
+        {
+          boost::system::error_code ignored_ec{};
+          const auto lep = new_connection_l->socket().local_endpoint(ignored_ec);
+          _dbg3("[sock " << new_connection_l->socket().native_handle() << "] to " << remote_endpoint << " from " << lep << " failed: " << error.message());
+          if (remove_connection(new_connection_l))
+            cb(t_connection_context{}, error);
+        };
+
+        if(!ec_)
+        {//success
+          const auto on_ready = [=] ()
+            {
+              if (sh_deadline->cancel())
+              {
+                boost::system::error_code ignored_ec{};
+                const auto lep = new_connection_l->socket().local_endpoint(ignored_ec);
+                _dbg3("[sock " << new_connection_l->socket().native_handle() << "] Connected successfully to " << remote_endpoint <<
+                  " from " << lep.address().to_string() << ':' << lep.port());
+
+                if (remove_connection(new_connection_l) && new_connection_l->start(false, 1 < m_threads_count))
+                {
+                  t_connection_context conn_context{};
+                  new_connection_l->get_context(conn_context);
+                  cb(conn_context, ec_);
+                }
+                else
+                  on_cancel(boost::asio::error::fault);
+              }
+              else // if timer already expired
+                on_cancel(boost::asio::error::operation_aborted);
+            };
+
+          if (new_connection_l->get_ssl_support() != ssl_support_t::e_ssl_support_disabled)
+          {
+            // set new timer for handshake
+            if (sh_deadline->expires_after(conn_timeout))
+            {
+              sh_deadline->async_wait(new_connection_l->wrap(on_timer));
+              new_connection_l->socket_.async_handshake(boost::asio::ssl::stream_base::client, new_connection_l->wrap([=] (boost::system::error_code ec)
+                {
+                  if (ec)
+                  {
+                    sh_deadline->cancel();
+                    if (new_connection_l->get_ssl_support() == ssl_support_t::e_ssl_support_autodetect)
+                    {
+                      _dbg3("[sock " << new_connection_l->socket().native_handle() << "] SSL connection to " <<
+                        remote_endpoint << " failed: " << ec.message() << ". Trying without SSL");
+                      new_connection_l->disable_ssl();
+                      new_connection_l->socket().close(ec); // ignore errors
+                      connect_async_internal(new_connection_l, remote_endpoint, conn_timeout, cb);
+                    }
+                    else // ssl mandatory and failed
+                      on_cancel(ec);
+                  }
+                  else // ssl handshake complete
+                  {
+                    new_connection_l->set_ssl_enabled();
+                    on_ready();
+                  }
+                }));
+            }
+            else // if timer already expired
+              on_cancel(boost::asio::error::operation_aborted);
+          }
+          else // ssl disabled
+            on_ready();
+        }
+        else // ec_ has error
+          on_cancel(ec_);
+      }));
+    return true;
+    CATCH_ENTRY_L0("boosted_tcp_server<t_protocol_handler>::connect_async_internal", false);
+  }
+
+  template<class t_protocol_handler>
+  bool boosted_tcp_server<t_protocol_handler>::remove_connection(const connection_ptr& new_connection)
+  {
+    if (!new_connection)
+      return false;
+    const boost::lock_guard<boost::mutex> sync{connections_mutex};
+    connections_.erase(new_connection);
+    return true;
+  }
 } // namespace
 } // namespace
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -1004,7 +1004,7 @@ namespace net_utils
       boost::uuids::random_generator()(),
       *real_remote,
       is_income,
-      connection_basic::m_ssl_support == ssl_support_t::e_ssl_support_enabled
+      connection_basic::m_ssl_support
     );
     m_host = real_remote->host_str();
     try { host_count(1); } catch(...) { /* ignore */ }
@@ -1737,7 +1737,7 @@ namespace net_utils
   }
   //---------------------------------------------------------------------------------
   template<class t_protocol_handler>
-  typename boosted_tcp_server<t_protocol_handler>::try_connect_result_t boosted_tcp_server<t_protocol_handler>::try_connect(connection_ptr new_connection_l, const std::string& adr, const std::string& port, boost::asio::ip::tcp::socket &sock_, const boost::asio::ip::tcp::endpoint &remote_endpoint, const std::string &bind_ip, uint32_t conn_timeout, epee::net_utils::ssl_support_t ssl_support)
+  typename boosted_tcp_server<t_protocol_handler>::try_connect_result_t boosted_tcp_server<t_protocol_handler>::try_connect(connection_ptr new_connection_l, const std::string& adr, const std::string& port, boost::asio::ip::tcp::socket &sock_, const boost::asio::ip::tcp::endpoint &remote_endpoint, const std::string &bind_ip, uint32_t conn_timeout, epee::net_utils::ssl_options_t& ssl_options)
   {
     TRY_ENTRY();
 
@@ -1813,7 +1813,7 @@ namespace net_utils
     {
       // Handshake
       MDEBUG("Handshaking SSL...");
-      if (!new_connection_l->handshake(boost::asio::ssl::stream_base::client))
+      if (!new_connection_l->client_handshake(ssl_options))
       {
         if (ssl_support == epee::net_utils::ssl_support_t::e_ssl_support_autodetect)
         {
@@ -1827,6 +1827,7 @@ namespace net_utils
           sock_.close();
         return CONNECT_FAILURE;
       }
+      new_connection_l->set_ssl_enabled();
     }
 
     return CONNECT_SUCCESS;
@@ -1835,11 +1836,11 @@ namespace net_utils
   }
   //---------------------------------------------------------------------------------
   template<class t_protocol_handler>
-  bool boosted_tcp_server<t_protocol_handler>::connect(const std::string& adr, const std::string& port, uint32_t conn_timeout, t_connection_context& conn_context, const std::string& bind_ip, epee::net_utils::ssl_support_t ssl_support)
+  bool boosted_tcp_server<t_protocol_handler>::connect(const std::string& adr, const std::string& port, uint32_t conn_timeout, t_connection_context& conn_context, const std::string& bind_ip, epee::net_utils::ssl_options_t ssl_options)
   {
     TRY_ENTRY();
 
-    connection_ptr new_connection_l(new connection<t_protocol_handler>(io_context_, m_state, m_connection_type, ssl_support) );
+    connection_ptr new_connection_l(new connection<t_protocol_handler>(io_context_, m_state, m_connection_type, ssl_options.support) );
     connections_mutex.lock();
     connections_.insert(new_connection_l);
     MDEBUG("connections_ size now " << connections_.size());
@@ -1927,24 +1928,22 @@ namespace net_utils
     //boost::asio::ip::tcp::endpoint remote_endpoint(boost::asio::ip::address::from_string(addr.c_str()), port);
     boost::asio::ip::tcp::endpoint remote_endpoint(*iterator);
 
-    auto try_connect_result = try_connect(new_connection_l, adr, port, sock_, remote_endpoint, bind_ip_to_use, conn_timeout, ssl_support);
+    auto try_connect_result = try_connect(new_connection_l, adr, port, sock_, remote_endpoint, bind_ip_to_use, conn_timeout, ssl_options);
     if (try_connect_result == CONNECT_FAILURE)
       return false;
-    if (ssl_support == epee::net_utils::ssl_support_t::e_ssl_support_autodetect && try_connect_result == CONNECT_NO_SSL)
+    if (ssl_options.support == epee::net_utils::ssl_support_t::e_ssl_support_autodetect && try_connect_result == CONNECT_NO_SSL)
     {
       // we connected, but could not connect with SSL, try without
       MERROR("SSL handshake failed on an autodetect connection, reconnecting without SSL");
       new_connection_l->disable_ssl();
-      try_connect_result = try_connect(new_connection_l, adr, port, sock_, remote_endpoint, bind_ip_to_use, conn_timeout, epee::net_utils::ssl_support_t::e_ssl_support_disabled);
+      ssl_options = epee::net_utils::ssl_support_t::e_ssl_support_disabled;
+      try_connect_result = try_connect(new_connection_l, adr, port, sock_, remote_endpoint, bind_ip_to_use, conn_timeout, ssl_options);
       if (try_connect_result != CONNECT_SUCCESS)
         return false;
     }
 
     // start adds the connection to the config object's list, so we don't need to have it locally anymore
-    connections_mutex.lock();
-    connections_.erase(new_connection_l);
-    connections_mutex.unlock();
-    bool r = new_connection_l->start(false, 1 < m_threads_count);
+    bool r = remove_connection(new_connection_l) && new_connection_l->start(false, 1 < m_threads_count);
     if (r)
     {
       new_connection_l->get_context(conn_context);
@@ -1963,10 +1962,10 @@ namespace net_utils
   }
   //---------------------------------------------------------------------------------
   template<class t_protocol_handler> template<class t_callback>
-  bool boosted_tcp_server<t_protocol_handler>::connect_async(const std::string& adr, const std::string& port, const std::chrono::milliseconds conn_timeout, const t_callback &cb, const std::string& bind_ip, epee::net_utils::ssl_support_t ssl_support, t_connection_context&& initial)
+  bool boosted_tcp_server<t_protocol_handler>::connect_async(const std::string& adr, const std::string& port, const std::chrono::milliseconds conn_timeout, const t_callback &cb, const std::string& bind_ip, epee::net_utils::ssl_options_t ssl_options, t_connection_context&& initial)
   {
     TRY_ENTRY();    
-    connection_ptr new_connection_l(new connection<t_protocol_handler>(io_context_, m_state, m_connection_type, ssl_support, std::move(initial)) );
+    connection_ptr new_connection_l(new connection<t_protocol_handler>(io_context_, m_state, m_connection_type, ssl_options.support, std::move(initial)) );
     connections_mutex.lock();
     connections_.insert(new_connection_l);
     MDEBUG("connections_ size now " << connections_.size());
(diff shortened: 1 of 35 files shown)


vtnerd:
Added sodium to the dependency list, hopefully builds pass now.

src/cryptonote_protocol/cryptonote_protocol_defs.h
```diff
@@ -87,9 +88,11 @@ namespace cryptonote
     uint8_t address_type;
 
     BEGIN_KV_SERIALIZE_MAP()
+      uint8_t ssl_i = uint8_t(ssl);
```
vtnerd, line 91: This is a bug, as `ssl_support_t` doesn't strictly define enum values.

vtnerd:
Did another force-push with a rebase, to get rid of the `external` changes that were not intended.

vtnerd:
I added a crude functional test for SSL. It just tests that autodetection and disabling SSL works.

docs/LEVIN_PROTOCOL.md
```diff
@@ -135,6 +135,50 @@ the `Expect Response` field zeroed. When a message of this type is received, the
...
+
+When a peerlist is shared, the encryption information is ignored (not
+trusted), and is put into `autodetect` mode by default. The node data shared
+during handshakes OR pings can change the node state to `no encryption` or
+`authenticated encryption`, based on direction information from the peer.
+
+> The server/responder must always be in autodetect SSL mode, so that it can
+handle any possible state that the client/initiator is in.
+
+> If a node changes certificates, "old" peers will be unable to connect until
```
j-berman, line 152: It looks as though if monerod generates a new temporary cert every run, that the node won't share the fingerprint with its peers. Therefore this should only affect nodes that have a persistent cert and then change it, ya?

src/p2p/net_node.cpp
```diff
@@ -142,17 +142,23 @@ namespace nodetool
...
-    const command_line::arg_descriptor<std::vector<std::string> > arg_p2p_add_peer   = {"add-peer", "Manually add peer to local peerlist"};
-    const command_line::arg_descriptor<std::vector<std::string> > arg_p2p_add_priority_node   = {"add-priority-node", "Specify list of peers to connect to and attempt to keep the connection open"};
-    const command_line::arg_descriptor<std::vector<std::string> > arg_p2p_add_exclusive_node   = {"add-exclusive-node", "Specify list of peers to connect to only."
-                                                                                                  " If this option is given the options add-priority-node and seed-node are ignored"};
-    const command_line::arg_descriptor<std::vector<std::string> > arg_p2p_seed_node   = {"seed-node", "Connect to a node to retrieve peer addresses, and disconnect"};
+    const command_line::arg_descriptor<std::vector<std::string> > arg_p2p_add_peer   = {"add-peer", "Manually add peer to local peerlist. Append \",sha-256 fingerprint\" to make authenticated connection."
```
j-berman: Nits:

```suggestion
    const command_line::arg_descriptor<std::vector<std::string> > arg_p2p_add_peer   = {"add-peer", "Manually add peer to local peerlist. Append \",<sha-256 fingerprint>\" to make authenticated connection. " 
```

Same for other references to the fingerprint

contrib/epee/include/net/net_utils_base.h
```diff
@@ -400,7 +402,7 @@ namespace net_utils
                                m_remote_address(),
                                m_is_income(false),
                                m_started(time(NULL)),
-                               m_ssl(false),
+			       m_ssl(ssl_support_t::e_ssl_support_disabled),
```
j-berman: nit: the whitespace has tabs here and doesn't line up in my editor

```suggestion
                               m_ssl(ssl_support_t::e_ssl_support_disabled),
```

contrib/epee/include/net/connection_basic.hpp
```diff
@@ -132,37 +132,15 @@ class connection_basic { // not-templated base class for rapid developmet of som
...
-			if (m_ssl_support == epee::net_utils::ssl_support_t::e_ssl_support_enabled)
-				socket_.async_read_some(buffers, std::forward<ReadHandler>(handler));
-			else
-				socket().async_read_some(buffers, std::forward<ReadHandler>(handler));
+			return ssl.handshake(socket_, boost::asio::ssl::stream_base::client);
 		}
 
-		template<typename ConstBufferSequence, typename WriteHandler>
-		void async_write_some(const ConstBufferSequence &buffers, WriteHandler &&handler)
+		bool server_handshake(boost::asio::const_buffer buffer)
```
j-berman: This is unused

j-berman:
Started a review and have an initial comment. I didn't get around to fixing `connect_async` as we discussed over [here](https://github.com/monero-project/monero/pull/8640#issuecomment-1318883080), and that needs to be done for this PR now that `connect_async` can get called with an SSL setting.

Running this code on a node that can get incoming connections, you can see that the back ping to an incoming node always fails (log level 2 you'll see `back ping invoke failed` or `back ping connect failed`, and no `PING SUCCESS`) after an incoming node connects to you.

I'm happy to do this since I said I would over in that PR, my bad for not getting to it sooner.

vtnerd:
@j-berman Fixed SSL connections with `connection_async`.

vtnerd:
Forgot to include newest code in last push. Trying again!

vtnerd:
Rebased, and trying to fix a linking issue that I cannot reproduce locally.

SyntheticBird45:
Please excuse my ignorance, I'm not familiar with the boost library and I'm therefore wondering what cipher are actually available ? afaiu, SSL namespace include TLS up to TLSv1.3: https://www.boost.org/doc/libs/1_74_0/boost/asio/ssl/context_base.hpp

I'm probably blind but i don't find in the commit changes where are cipher options. That would be good to enforce at least TLSv1.2

vtnerd:
The code already[ forces TLS 1.2+](https://github.com/monero-project/monero/blob/master/contrib/epee/src/net_ssl.cpp#L305). It also limits the cipher list a few lines below that.

vtnerd:
Bad rebase, will fix.

vtnerd:
The test that just failed should be spurious.

vtnerd:
I rebased against latest master changes, which included a test merge and a seed node merge.

vtnerd:
Force pushed a rebase. Nothing should have changed. Looking at the diff again.

vtnerd:
Force pushed another small update removing some unintended changes.

vtnerd:
The failed test is not related. May push a dummy update so that all unit tests get a chance to run.

vtnerd:
Force pushed a rebase merge for p2p functional tests.

docs/LEVIN_PROTOCOL.md
```diff
@@ -135,6 +135,50 @@ the `Expect Response` field zeroed. When a message of this type is received, the
...
+requesting one of the other states, or (2) the peer sends a handshake or ping
+message indicating it should be in another state.
+
+When connecting to a peer of this type, the client/initiator should attempt
+SSL with no server/responder verification. If the connection fails, it should
+immediately attempt a non-SSL connection.
+
+### Authenticated Encryption
+If a peer chooses to have a persistent SSL certificate, it will send
+`encryption_ver = 2` AND `ssl_finger = <binary sha256 of cert>` in the
```
j-berman:
```suggestion
`encryption_mode = 2` AND `ssl_finger = <binary sha256 of cert>` in the
```

contrib/epee/src/net_ssl.cpp
```diff
@@ -652,6 +654,44 @@ bool ssl_options_t::handshake(
...
+  md.resize(n);
+  return md;
+}
+
+std::string get_ssl_fingerprint(boost::asio::ssl::context &ssl_context, const EVP_MD *fdig)
+{
+  SSL_CTX *ctx = ssl_context.native_handle();
+  CHECK_AND_ASSERT_THROW_MES(ctx, "Failed to get SSL context");
+  
+  X509* ssl_cert = nullptr;
```
j-berman, line 571: Free with `X509_free` when done?

contrib/epee/include/storages/levin_abstract_invoke2.h
```diff
@@ -111,7 +111,8 @@ namespace epee
       levin::message_writer to_send;
       stg.store_to_binary(to_send.buffer);
 
-      int res = transport.send(to_send.finalize_notify(command), conn_id);
+      const bool ssl = (context.m_ssl != ssl_support_t::e_ssl_support_disabled);
+      int res = transport.send(to_send.finalize_notify(command, ssl), conn_id);
```
j-berman: What if we used autodetect and found no encryption?

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1934,60 +1933,113 @@ namespace net_utils
...
+                else
+                  on_cancel(boost::asio::error::fault);
+              }
+              else // if timer already expired
+                on_cancel(boost::asio::error::operation_aborted);
+            };
+
+          if (new_connection_l->get_ssl_support() != ssl_support_t::e_ssl_support_disabled)
+          {
+            // set new timer for handshake
```
j-berman, line 2109: Thoughts on using `new_connection_l->client_handshake` here? That handshake function handles waiting, cleaning up on failure, synchronization. Reusing that code seems like one less thing to need to worry about here

OTOH, that code also shuts down and closes the socket on failure, which could complicate things a little here.

Your synchronization looks good because you wrap the callbacks in the strand. I don't have a strong preference so long as this works.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1934,60 +1933,113 @@ namespace net_utils
...
+              new_connection_l->socket_.async_handshake(boost::asio::ssl::stream_base::client, new_connection_l->wrap([=] (const boost::system::error_code& ec)
+                {
+                  if (ec)
+                  {
+                    if (new_connection_l->get_ssl_support() == ssl_support_t::e_ssl_support_autodetect)
+                    {
+                      _dbg3("[sock " << new_connection_l->socket().native_handle() << "] SSL connection to " <<
+                        remote_endpoint << " failed: " << ec.message() << ". Trying without SSL");
+                      new_connection_l->disable_ssl();
+                      connect_async_internal(new_connection_l, remote_endpoint, conn_timeout, cb);
```
j-berman, line 2124: Looks like you want to make sure the locally scoped timer isn't still running before you enter this function

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -486,7 +486,7 @@ class async_protocol_handler
 
           bool is_response = (m_oponent_protocol_ver == LEVIN_PROTOCOL_VER_1 && m_current_head.m_flags&LEVIN_PACKET_RESPONSE);
 
-          MDEBUG(m_connection_context << "LEVIN_PACKET_RECEIVED. [len=" << m_current_head.m_cb
+	  MDEBUG(m_connection_context << "LEVIN_PACKET_RECEIVED. [len=" << m_current_head.m_cb
```
j-berman: Nit: whitespace

tests/unit_tests/levin.cpp
```diff
@@ -1068,7 +1068,7 @@ TEST_F(levin_notify, fluff_with_padding)
         {
             auto notification = receiver_.get_notification<cryptonote::NOTIFY_NEW_TRANSACTIONS>().second;
             EXPECT_EQ(txs, notification.txs);
-            EXPECT_FALSE(notification._.empty());
+            EXPECT_TRUE(notification._.empty());
```
j-berman, line 1100: Checking my understanding: because the padding doesn't get serialized into `_` as part of the NOTIFY_NEW_TRANSACTIONS request anymore, and instead gets stuffed into the buffer, the `_` field is now empty even though this request should have been padded. And the reason you don't pad into a `_` field anymore is because you want padding on every message, rather than strictly NOTIFY_NEW_TRANSACTIONS

It would be nice if there was an explicit way to see/test the padding at this level, but not worth changing the internals for it

A comment explaining how these `_with_padding` tests are just testing smooth notifications even with `pad` true would be solid, since they're no longer explicitly testing the padding

contrib/epee/src/levin_base.cpp
```diff
@@ -41,15 +43,25 @@ namespace levin
...
+      pad_bytes = randombytes_uniform(P2P_MAX_LEVIN_PAD_BYTES);
+      if (std::numeric_limits<std::size_t>::max() - payload_size() < pad_bytes)
+        throw std::runtime_error{"levin_write::finalize failed to pad bytes due to overflow"};
+    }
+
+    header head = make_header(command, payload_size() + pad_bytes, flags, expect_response);
     head.m_return_code = SWAP32LE(return_code);
 
     std::memcpy(buffer.tellp() - buffer.size(), std::addressof(head), sizeof(head));
+    buffer.put_n(0, pad_bytes);
```
j-berman, line 67: Just to check my understanding: when de-serializing a message, the de-serializer is unaffected by these padded bytes because these are just dangling bytes that have no associated key, ya? The de-serializer can safely just ignore them

src/p2p/net_node.inl
```diff
@@ -1004,8 +1027,39 @@ namespace nodetool
...
+            {
+              p2p_ssl.auth.private_key_path = key_files + ".key";
+              p2p_ssl.auth.certificate_path = key_files + ".crt";
+            }
+            else
+              store_cert = true;
+          }
+        }
+        else
+          MINFO("Generating new temporary SSL certificate for p2p");
```
j-berman: Looks like this was meant to be an else for `if (command_line::get_arg(vm, arg_p2p_persistent_cert))`

j-berman (commented):
This generally looks ok to me, though I'm not particularly well versed with SSL. Mostly nits and some questions

contrib/epee/src/levin_base.cpp
vtnerd: Yes, the epee parser always ignored extra data after the root object. The "noise" mode in Tor uses this "feature" already - each message is padded to a specific length with possible trailing. So we have a working example at least.

Any future parser already has to support this feature due to the noise mode.

contrib/epee/src/net_ssl.cpp
vtnerd: The documentation for `SSL_get_certificate` states:

> Returned objects are owned by the SSL object and should not be freed by users of these functions.

So the `dflt_SSL` `unique_ptr` should cleanup the memory for this object.

docs/LEVIN_PROTOCOL.md
vtnerd: Correct, if a fingerprint is sent in the peerlist exchange, it is removed by the receiving node. The fingerprint is also exchanged in the handshake, iff the user has a fixed certificate. In this situation, the peer caches it and is equivalent to a unique ip/port pair in terms of treatment and recovery.

src/p2p/net_node.cpp
vtnerd: Fixed.

contrib/epee/include/net/net_utils_base.h
vtnerd: I think I fixed this. Always annoying in editors.

contrib/epee/include/net/connection_basic.hpp
vtnerd: Removed.

tests/unit_tests/levin.cpp
vtnerd: Commented the best I could in the test file.

src/p2p/net_node.inl
vtnerd: I changed this to `else if (zone.first == epee::net_utils::zone::public_)` with an updating message saying "Not using encryption for p2p".

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: Leaving this open for now, as I need to investigate further.

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: Good catch, added `sd_deadline->cancel()` in the `if (ec)` section.

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: Looked into it. The `client_handshake` function isn't asynchronous, so I just rolled this here. For better worse. Unlikely to change, but not marking as resolved either.

vtnerd:
Force pushed a rebase and @j-berman changes.

j-berman (approved):

jonathancross:
@vtnerd Looks like there are conflicts to resolve.

vtnerd:
Force pushed a rebase to solve merge conflicts.

tests/functional_tests/functional_tests_rpc.py
```diff
@@ -56,9 +56,9 @@
 monerod_extra = [
   ["--offline"],
   ["--rpc-payment-address", "44SKxxLQw929wRF6BA9paQ1EWFshNnKhXM3qz6Mo3JGDE2YG3xyzVutMStEicxbQGRfrYvAAYxH6Fe8rnD56EaNwUiqhcwR", "--rpc-payment-difficulty", str(DIFFICULTY), "--rpc-payment-credits", "5000", "--offline"],
-  ["--add-exclusive-node", "127.0.0.1:18283"],
-  ["--add-exclusive-node", "127.0.0.1:18282"],
-  ["--rpc-login", "md5_lover:Z1ON0101", "--offline"],
+  ["--add-exclusive-node", "127.0.0.1:18283", "--add-exclusive-node", "127.0.0.1:18284", "--p2p-disable-encryption"],
+  ["--add-exclusive-node", "127.0.0.1:18282", "--add-exclusive-node", "127.0.0.1:18284"],
+  ["--rpc-login", "md5_lover:Z1ON0101", "--add-exclusive-node", "127.0,0,1:18282,no_encryption", "--add-exclusive-node", "127.0.0.1:18283"],
```
selsta:
```suggestion
  ["--rpc-login", "md5_lover:Z1ON0101", "--add-exclusive-node", "127.0.0,1:18282,no_encryption", "--add-exclusive-node", "127.0.0.1:18283"],
```

vtnerd:
Force pushed another rebase to handle change conflicts to master.

tests/functional_tests/functional_tests_rpc.py
vtnerd: Fixed.

vtnerd:
Sorry another force push - fixed the change @selsta recommended.

vtnerd:
This latest force push should fix the build with newer boost versions.

vtnerd:
The latest push should fix the functional tests.

vtnerd:
Force pushed a rebase to fix recent conflicts.

SyntheticBird45:
> Force pushed a rebase to fix recent conflicts.

3 years and still doing god work

vtnerd:
Our old friend (`race_condition`) showed up again. So this latest rebase looks good.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -2017,60 +2016,114 @@ namespace net_utils
...
+                {
+                  if (ec)
+                  {
+                    sh_deadline->cancel();
+                    if (new_connection_l->get_ssl_support() == ssl_support_t::e_ssl_support_autodetect)
+                    {
+                      _dbg3("[sock " << new_connection_l->socket().native_handle() << "] SSL connection to " <<
+                        remote_endpoint << " failed: " << ec.message() << ". Trying without SSL");
+                      new_connection_l->disable_ssl();
+                      connect_async_internal(new_connection_l, remote_endpoint, conn_timeout, cb);
```
selsta, line 2124: Should this perform a fresh TCP connect, like the synchronous path does?

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -2017,60 +2016,114 @@ namespace net_utils
...
+                    }
+                    else // ssl mandatory and failed
+                      on_cancel(ec);
+                  }
+                  else // ssl handshake complete
+                    on_ready();
```
selsta, line 2132: Should `set_ssl_enabled()` be added here?

src/p2p/net_node.inl
```diff
@@ -979,8 +1002,39 @@ namespace nodetool
...
+        bool share_fingerprint = false;
+        net::ssl_options_t p2p_ssl{net::ssl_support_t::e_ssl_support_disabled};
+        const auto key_files = m_config_folder + "/p2p_public";
+        if (zone.first == epee::net_utils::zone::public_ && !command_line::get_arg(vm, arg_p2p_disable_encryption))
+        {
+          p2p_ssl = net::ssl_options_t{net::ssl_support_t::e_ssl_support_autodetect};
+          if (command_line::get_arg(vm, arg_p2p_persistent_cert))
+          {
+            share_fingerprint = true;
+            p2p_ssl = net::ssl_options_t{net::ssl_support_t::e_ssl_support_enabled};
```
selsta: Does using `e_ssl_support_enabled` here make the inbound P2P listener require TLS? The new Levin docs say the responder must always be in autodetect SSL mode so it can handle both encrypted and unencrypted initiators.

contrib/epee/src/levin_base.cpp
```diff
@@ -41,15 +43,25 @@ namespace levin
...
   {
     if (buffer.size() < sizeof(header))
       throw std::runtime_error{"levin_writer::finalize already called"};
 
-    header head = make_header(command, payload_size(), flags, expect_response);
+    std::uint32_t pad_bytes = 0;
+    if (pad)
+    {
+      static_assert(P2P_MAX_LEVIN_PAD_BYTES <= std::numeric_limits<std::uint32_t>::max(), "invalid max ssl pad bytes");
+      pad_bytes = randombytes_uniform(P2P_MAX_LEVIN_PAD_BYTES);
```
selsta:
```suggestion
      pad_bytes = randombytes_uniform(P2P_MAX_LEVIN_PAD_BYTES + 1);
```

or update the `pad == true` comment to `0..8191` instead of `0..8192`.

contrib/epee/include/storages/levin_abstract_invoke2.h
selsta:
```suggestion
      int res = transport.send(to_send.finalize_notify(command, context.should_pad()), conn_id);
```

also here?

src/p2p/net_node.inl
```diff
@@ -77,6 +78,19 @@ static inline boost::asio::ip::address_v4 make_address_v4_from_v6(const boost::a
...
+    namespace net = epee::net_utils;
+    const auto pos = adr.find(',');
+    net::ssl_options_t e2e_mode{net::ssl_support_t::e_ssl_support_autodetect};
+    if (boost::string_ref{adr}.ends_with("no_encryption"))
+      e2e_mode = net::ssl_options_t{net::ssl_support_t::e_ssl_support_disabled};
+    else if (pos != std::string::npos)
+      e2e_mode = net::ssl_options_t{{epee::from_hex_locale::to_vector(boost::string_ref{adr}.substr(pos + 1))}, {}};
+    adr.erase(std::min(adr.size(), adr.find(',')));
+    return {std::move(adr) , std::move(e2e_mode)};
+  }
```
selsta:
```suggestion
static bool parse_address(string_address& out, std::string adr)
{
  namespace net = epee::net_utils;

  net::ssl_options_t e2e_mode{net::ssl_support_t::e_ssl_support_autodetect};

  const auto comma = adr.find(',');
  if (comma == std::string::npos)
  {
    out = {std::move(adr), std::move(e2e_mode)};
    return true;
  }

  const std::string option = adr.substr(comma + 1);
  adr.erase(comma);

  if (adr.empty() || option.empty())
    return false;

  if (option == "no_encryption")
  {
    out = {
      std::move(adr),
      net::ssl_options_t{net::ssl_support_t::e_ssl_support_disabled}
    };
    return true;
  }

  std::vector<std::uint8_t> fingerprint;
  try
  {
    fingerprint = epee::from_hex_locale::to_vector(boost::string_ref{option});
  }
  catch (const std::exception&)
  {
    return false;
  }

  if (fingerprint.size() != 32)
    return false;

  e2e_mode = net::ssl_options_t{{std::move(fingerprint)}, {}};
  out = {std::move(adr), std::move(e2e_mode)};
  return true;
}
```

more strict parsing

tests/functional_tests/p2p.py
```diff
@@ -454,5 +472,29 @@ def bench_p2p_heavy_block_propagation_speed(self):
...
+        connections = daemon2.get_connections().connections
+        for connection in connections:
+            assert connection.ssl == 0
+
+        connections = daemon3.get_connections().connections
+        for connection in connections:
+            if connection.port == "18282":
+                assert connection.ssl == 0
+            elif connection.port == "18284":
+                assert connection.ssl == 2
```
selsta, line 490: Should this assert here check for `autodetect` or `enabled`?

tests/functional_tests/p2p.py
```diff
@@ -148,20 +157,26 @@ def test_p2p_reorg(self):
...
         time.sleep(5)
         res = daemon3.get_info()
         assert res.height == height + 6
         assert res.top_block_hash == daemon2_top_block_hash
 
         # disconnect and mine a lot on daemon3
         daemon2.out_peers(0)
+        daemon2.in_peers(0)
         daemon3.out_peers(0)
+        daemon2.in_peers(0)
```
selsta:
```suggestion
        daemon3.in_peers(0)
```

src/p2p/net_node.inl
vtnerd: This was a bug, good catch.

Looks like I mixed this up with what the client should be doing when a fingerprint is provided.

contrib/epee/src/levin_base.cpp
vtnerd: Changed the comment.

contrib/epee/include/storages/levin_abstract_invoke2.h
vtnerd: Good catch, I think the suggestion is better.

tests/functional_tests/p2p.py
vtnerd: Fixed.

tests/functional_tests/p2p.py
vtnerd: The test is currently checking for `autodetect`. The code never moves the state to `enabled`, that state should be exclusive to when a user specifies a fingerprint of when the connection (via handshake) signals it is using a persistent cert.

This can be seen in the [status printer](https://github.com/monero-project/monero/pull/8996/changes#diff-eb68bc70cb6aeb0af3d4d6ebb9cfc2512547dc8590ae4b78f427dfe255b6e1c0R665) where autodetect and enabled have different labels printed.

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: Yes, this does appear to be useful/needed.

src/p2p/net_node.inl
vtnerd: I'm probably going to do this, but it won't be included in the SSL+beta-stressnet PR.

vtnerd:
Force pushed changes recommended by @selsta

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: It shouldn't matter afaik. The `async_connect` routine automatically closes existing sockets.

vtnerd:
Force pushed another rebase to fix conflicts.

selsta:
Node server race condition test failed in a new way, which appears related to this PR.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -2048,60 +2047,117 @@ namespace net_utils
...
+  }
+
+  template<class t_protocol_handler>
+  bool boosted_tcp_server<t_protocol_handler>::remove_connection(const connection_ptr& new_connection)
+  {
+    if (!new_connection)
+      return false;
+    const boost::lock_guard<boost::mutex> sync{connections_mutex};
+    connections_.erase(new_connection);
     return true;
```
selsta, line 2156:
```suggestion
    return connections_.erase(new_connection) != 0;
```

to avoid later calling `cb` later from a stale async path

contrib/epee/include/net/abstract_tcp_server2.inl
selsta: I checked Boost 1.90 source for `basic_socket::async_connect`, it only opens the socket if `!is_open()` and does not close/reset an already-open socket. The synchronous autodetect path explicitly does `shutdown()` + `close()` before retrying without SSL.

vtnerd:
GH is bugged at the moment, as I can't reply to this directly:

>  checked Boost 1.90 source for basic_socket::async_connect, it only opens the socket if !is_open() and does not close/reset an already-open socket. The synchronous autodetect path explicitly does shutdown() + close() before retrying without SSL.

Yes, I just noticed that `boost::asio::async_connect`, a standalone function, has different behavior than the `socket::async_connect` (the former closes before connecting). Will fix.

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: Isn't this an existing issue (that should never happen)?

selsta: If it matches existing behavior (and there isn't more risk with the new code) then this can be ignored.

vtnerd: The prior pattern was:

```c++
    connections_mutex.lock();
    connections_.erase(new_connection_l);
    connections_mutex.unlock();
```

where the result of the erase was ignored. I moved this into a separate function, and the result still doesn't matter, but the `nullptr` check is _somewhat_ useful.

I should be able to remove the `connections_` map code from the async connect entirely. I kept it because the existing async_connect code had it.

vtnerd:
> Node server race condition test failed in a new way, which appears related to this PR.

Finally got this fixed. Wasted a lot of time on an issue with random `boost::uuid`s on my Ryzen chip.

src/p2p/net_node.inl
```diff
@@ -976,8 +1034,38 @@ namespace nodetool
...
+        else if (zone.first == epee::net_utils::zone::public_)
+          MINFO("Not using encryption for p2p");
+
+        zone.second.m_ssl_support = p2p_ssl.support;
+        res = zone.second.m_net_server.init_server(zone.second.m_port, zone.second.m_bind_ip, ipv6_port, ipv6_addr, m_use_ipv6, m_require_ipv4, std::move(p2p_ssl));
         CHECK_AND_ASSERT_MES(res, false, "Failed to bind server");
+        if (store_cert)
+          net::store_ssl_keys(zone.second.m_net_server.get_ssl_context(), key_files);
+        if (share_fingerprint)
+          zone.second.m_ssl_finger = net::get_ssl_fingerprint(zone.second.m_net_server.get_ssl_context());
```
selsta, line 1198: This needs to be conditional on `store_ssl_keys()` succeeding, otherwise a non-persistent fingerprint is shared.

src/p2p/net_node.inl
```diff
@@ -2292,6 +2385,8 @@ namespace nodetool
     node_data.rpc_credits_per_hash = zone.m_can_pingback ? m_rpc_credits_per_hash : 0;
     node_data.network_id = m_network_id;
     node_data.support_flags = zone.m_config.m_support_flags;
+    node_data.encryption_mode = get_encryption_mode(zone.m_ssl_support);
```
selsta:
```suggestion
    node_data.encryption_mode = zone.m_ssl_finger.empty() ? get_encryption_mode(zone.m_ssl_support) : emode_ssl_enabled;
```

Otherwise `get_p2p_encryption()` returns early and ignores the fingerprint.

src/p2p/net_peerlist.h
selsta: `encryption_mode` and `cert_finger` are not populated here. `encryption_mode` is indeterminate and `cert_finger` is empty. `COMMAND_TIMED_SYNC` replaces the outgoing peer's stored encryption state with corrupt data.

selsta (changes requested):

src/p2p/net_node.inl
vtnerd: Good catch, fixed.

contrib/epee/src/levin_base.cpp
selsta: `PING` and `REQUEST_SUPPORT_FLAGS` allow only 4096 bytes, so `P2P_MAX_LEVIN_PAD_BYTES` can randomly exceeds command limits.

src/p2p/net_node.inl
```diff
@@ -2507,7 +2610,7 @@ namespace nodetool
         return false;
       }
       return true;
-    }, "0.0.0.0", m_ssl_support, p2p_connection_context{true /* is_ping */});
+    }, "0.0.0.0", get_p2p_encryption(node_data), p2p_connection_context{true /* is_ping */});
```
selsta: `--p2p-disable-encryption` is ignored here.

src/p2p/net_node.inl
vtnerd: Yes, I stepped through all of the logic again, and this is defintely needed as `m_ssl_support` is typically `autodetect` so that the server is in autodetect mode (otherwise SSL is forcibly required by the node).

src/p2p/net_peerlist.h
vtnerd: Actually this should be fine AFAIK, but it's not obvious how.

Nearly every location uses brace/zero initialization instead of default initialization. The one place that default initializes correctly sets `encryption_mode` and `cert_finger`. So this patch is fine as-is.

Despite this, I am switching to brace initialization everywhere, and adding a default constructor.

contrib/epee/src/levin_base.cpp
vtnerd: I added a fix where the randomized padding is capped at the message max size. Most messages are capped at the 8 KiB global padding max.

src/p2p/net_node.inl
vtnerd: Great catch. This was fixed too.

vtnerd:
Force pushed a rebase and fixes to @selsta comments.

vtnerd:
Force pushed a compilation fix for fuzz tests.

src/p2p/p2p_protocol_defs.h
```diff
@@ -48,22 +50,91 @@ namespace nodetool
...
+    // Fingerprint specified, require SSL
+    std::vector<std::vector<std::uint8_t>> fingers{net::convert_fingerprint(peer.cert_finger)};
+    return {std::move(fingers), {}};
+  }
+
   
   template<typename AddressType>
   struct peerlist_entry_base
   {
+    peerlist_entry_base()
```
selsta, line 111: `pruning_seed` is missing here.

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -207,7 +179,7 @@ namespace levin
 
     bool make_payload_send_txs(connections& p2p, std::vector<blobdata>&& txs, const boost::uuids::uuid& destination, const bool pad, const bool fluff)
     {
-      epee::byte_slice blob = make_tx_message(std::move(txs), pad, fluff).finalize_notify(NOTIFY_NEW_TRANSACTIONS::ID);
+      epee::byte_slice blob = make_tx_message(std::move(txs), fluff).finalize_notify(NOTIFY_NEW_TRANSACTIONS::ID, pad);
```
selsta, line 182: `finalize_notify` now expects a maximum padded size, not a boolean.

src/p2p/net_node.cpp
```diff
@@ -135,17 +135,23 @@ namespace nodetool
...
     const command_line::arg_descriptor<std::vector<std::string> > arg_tx_proxy = {"tx-proxy", "Send local txes through proxy: <network-type>,[socks5://[user:pass@]]<socks-ip:port>[,max_connections][,disable_noise] i.e. \"tor,127.0.0.1:9050,100,disable_noise\""};
     const command_line::arg_descriptor<std::vector<std::string> > arg_anonymous_inbound = {"anonymous-inbound", "<hidden-service-address>,<[bind-ip:]port>[,max_connections] i.e. \"x.onion,127.0.0.1:18083,100\""};
     const command_line::arg_descriptor<std::string> arg_ban_list = {"ban-list", "Specify ban list file, one IP address per line"};
     const command_line::arg_descriptor<bool> arg_p2p_hide_my_port   =    {"hide-my-port", "Do not announce yourself as peerlist candidate", false, true};
     const command_line::arg_descriptor<bool> arg_no_sync = {"no-sync", "Don't synchronize the blockchain with other peers", false};
     const command_line::arg_descriptor<bool> arg_enable_dns_blocklist = {"enable-dns-blocklist", "Apply realtime blocklist from DNS", false};
+    const command_line::arg_descriptor<bool> arg_p2p_disable_encryption = {"p2p-disable-encryption", "Disable all P2P encryption", false};
```
selsta, line 153: Not a bug, but should this be renamed to `--p2p-ssl` with the following options, `autodetect`, `disabled`, with `autodetect` being the default? This would allow us to add `enabled` in the future.

contrib/epee/src/net_ssl.cpp
```diff
@@ -412,18 +413,19 @@ void ssl_options_t::configure(
...
+    socket.set_verify_callback([host, this_copy = std::move(this_copy)](const bool preverified, boost::asio::ssl::verify_context &ctx)
     {
       // preverified means it passed system or user CA check. System CA is never loaded
       // when fingerprints are whitelisted.
       const bool verified = preverified &&
-        (verification != ssl_verification_t::system_ca || host.empty() || MONERO_HOSTNAME_VERIFY(host)(preverified, ctx));
+        (this_copy.verification != ssl_verification_t::system_ca || host.empty() || MONERO_HOSTNAME_VERIFY(host)(preverified, ctx));
 
-      if (!verified && !has_fingerprint(ctx))
+      if (!verified && !this_copy.has_fingerprint(ctx))
```
selsta, line 425: A system-trusted certificate bypasses an explicitly configured fingerprint here.

src/p2p/net_node.inl
```diff
@@ -3162,24 +3273,24 @@ namespace nodetool
 
   template<typename t_payload_net_handler>
   boost::optional<p2p_connection_context_t<typename t_payload_net_handler::connection_context>>
-  node_server<t_payload_net_handler>::socks_connect(network_zone& zone, const epee::net_utils::network_address& remote, epee::net_utils::ssl_support_t ssl_support)
+  node_server<t_payload_net_handler>::socks_connect(network_zone& zone, const p2p_address& peer)
   {
-    auto result = socks_connect_internal(zone.m_net_server.get_stop_signal(), zone.m_net_server.get_io_context(), zone.m_proxy_address, remote);
+    auto result = socks_connect_internal(zone.m_net_server.get_stop_signal(), zone.m_net_server.get_io_context(), zone.m_proxy_address, peer.na);
```
selsta, line 3278: Public connections through `--proxy` silently bypass TLS and certificate pins.

src/p2p/net_node.inl
```diff
@@ -2790,6 +2894,8 @@ namespace nodetool
         pe.pruning_seed = context.m_pruning_seed;
         pe.rpc_port = context.m_rpc_port;
         pe.rpc_credits_per_hash = context.m_rpc_credits_per_hash;
+        pe.encryption_mode = arg.node_data.encryption_mode;
+        pe.cert_finger = arg.node_data.cert_finger;
```
selsta, line 2898: Oversized fingerprint strings can poison subsequent peerlist responses. `cert_finger` is an unrestricted string, it should be validated as 32 bytes before caching them.

src/p2p/net_node.inl
```diff
@@ -1329,9 +1416,11 @@ namespace nodetool
         context.m_rpc_port = rsp.node_data.rpc_port;
         context.m_rpc_credits_per_hash = rsp.node_data.rpc_credits_per_hash;
         context.support_flags = rsp.node_data.support_flags;
+        context.encryption_mode = rsp.node_data.encryption_mode;
```
selsta, line 1419: `--add-peer IP:port,no_encryption` initially connects plaintext, but reconnects using autodetection if the remote advertises its default mode.

src/daemon/rpc_command_executor.cpp
```diff
@@ -642,6 +642,20 @@ bool t_rpc_command_executor::print_connections() {
...
+  {
+    switch (ssl)
+    {
+      default:
+      case epee::net_utils::ssl_support_t::e_ssl_support_disabled:
+        break;
+      case epee::net_utils::ssl_support_t::e_ssl_support_autodetect:
+        return "autodetect";
+      case epee::net_utils::ssl_support_t::e_ssl_support_enabled:
+        return "verified";
```
selsta, line 655: This should be renamed to encrypted, because there are scnearios where `get_p2p_encryption()` explicitly returns `enabled` with `verification=none`.

contrib/epee/src/net_ssl.cpp
vtnerd: I don't understand, this logic isn't changing, only a copy is being made to explicitly correct the lifetimes. And I believe the current logic to be correct.

If a fingerprint is set, the system ca is **not** loaded into the `ctx`. So `preverified` will never be `true` afaik. Are you saying that `boost::asio::ssl::context` auto loads certificates without user choice?

I recall testing this back when, and you have to manually request the default root ca or the pre-verification always fails.

src/p2p/net_node.inl
vtnerd: Great catch. If in public zone, I need to use `peer.ssl` because the user almost certainly wants SSL when available.

src/p2p/p2p_protocol_defs.h
vtnerd: Crap. Fixed.

src/cryptonote_protocol/levin_notify.cpp
vtnerd: Crud. Fixed. I also updated three situations that used `false` to `0`, just to make things explicit.

src/p2p/net_node.cpp
vtnerd: What would `enabled` do given that `disabled` is an option? Refuse/skip outgoing connections to those hosts? Refuse incoming unencrypted connections?

I'll make the change now, in case we want to upgrade and figure this out later.

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: I added an explicit `close` call but forgot to update here.

src/p2p/net_node.inl
vtnerd: The holdup is that `--proxy` is hard to fix with SSL autodetect.

The `::add_connection` logic does an async SSL handshake, but `net_node::socks_connect` needs a synchronous attempt for the SSL autodetect (via socks) flow to work. The alternative - making `::socks_connect` async - is messy as it messes with the outgoing connection logic.

It looks like `::add_connection` is only used in this one location and in tests. So I'm going to make changes to the TCP server to make (this one function used in exactly one location) synchronous, and cleanup the tests.

contrib/epee/src/net_ssl.cpp
vtnerd: Ok, I dug into this (thanks to @selsta ). The server creates a SSL context for the entire TCP server, which is inappropriate for the per-connection SSL configuration used by P2P-SSL.

I don't think it's possible to get a SSL certificate for an IP address, but I am going to fix this anyway. The "easy" fix is to just change the logic here to get fingerprints working, but that still leaves `preverified` broken with custom CA/certs (which isn't used by P2P-SSL, but someone could use that `server::connect` codepath in the future).

ComputeryPony: > I don't think it's possible to get a SSL certificate for an IP address

Some CAs do support IP address certs, notably [Let's Encrypt does as of this year](https://letsencrypt.org/2026/01/15/6day-and-ip-general-availability).

As a node operator though, given how many CAs are trusted by default on most OSes, without Certificate Transparency checks in monerod I would probably just opt to use a self signed cert and have people hardcode the fingerprint. It would be nice if monerod supported CT checks but I have no idea how much work that would incur.

vtnerd: A point (in this not being an issue) is that you cannot configure the P2P-SSL server with CA checks. So the `preverified` value here will always be `false`.

The "fix" requires a SSL context per outgoing connection, which hurts optimizations around re-using the context object (as-in OpenSSL has to construct lots of objects per outgoing connection).

I'm going to double check that this works as expected, drop a comment in the config location, AND add runtime-checks in the `server::connect` routine to ensure the correct verification will occur (the checks: (1) that ca verification is off in the server context, and (2) that ca verification is off in the per-connection request).

vtnerd:
@selsta @j-berman - is it a bug if incoming connections are NOT verified? Because if I use `--add-peer` it only verifies the fingerprint if my node makes the connection first. If my peer does the same `--add-peer` command and connects first, their side validates but not mine.

OTOH, if both sides enter a fingerprint, the validation from the one side should be sufficient as only one key-exchange could only have occurred in the connection (i.e. no MitM). And validating incoming connections is tricky because if the value came from a peerlist exchange, it could be an attempt to poison certain connections (i.e. we would have to do incoming validation on CLI specified peers only).

contrib/epee/src/net_ssl.cpp
vtnerd: > As a node operator though, given how many CAs are trusted by default on most OSes, without Certificate Transparency checks in monerod I would probably just opt to use a self signed cert and have people hardcode the fingerprint. It would be nice if monerod supported CT checks but I have no idea how much work that would incur.

This is what we are doing. The existing patch (reflected here right now) never loads CA certs for P2P, and thus it either does autodetect/no_verify, or forcibly does a fingerprinted check.

The one negative is that the user has to run a command to check connections, and then look through the logs if a fingerprint check fails. I'm not sure how to handle this as a shutdown is likely inappropriate too.

PR #9002, opened 2023-09-25
depends: remove libiconv: unused

tobtoht:
In absense of ICU, iconv is required to build Boost Locale on POSIX platforms. Boost Locale is only used on Windows to support unicode filenames, but it [does not use iconv by default](https://www.boost.org/doc/libs/1_77_0/libs/locale/doc/html/building_boost_locale.html) and uses the winapi [localization backend](https://www.boost.org/doc/libs/1_72_0/libs/locale/doc/html/using_localization_backends.html) instead.

Since we don't need Boost Locale for POSIX platforms and we don't need iconv for Boost Locale on Windows, we can remove it. Opening unicode filenames remains possible on Windows.


hyc:
Why are we omitting ICU from depends builds?

selsta:
@hyc https://github.com/monero-project/monero/pull/8880

selsta (approved):

PR #9003, opened 2023-09-26
depends: unwind: do not build for linux hosts

tobtoht:
It is unused. easylogging++ [is used instead](https://github.com/monero-project/monero/blob/master/CMakeLists.txt#L537) when building with GCC, which is used for all Linux builds in depends.

We could use libunwind for FreeBSD, so I'm keeping the package definition in case someone wants to fix that build.

### DIFF
--- a/contrib/depends/packages/packages.mk
+++ b/contrib/depends/packages/packages.mk
@@ -25,10 +25,6 @@ ifeq ($(build_tests),ON)
 packages += gtest
 endif
 
-ifneq ($(host_arch),riscv64)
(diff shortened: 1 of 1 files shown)


tobtoht:
Interesting unrelated test case failure, tagging @jeffro256.

jeffro256:
Thanks for the ping. Sometimes that test case will fail randomly when the system entropy is low

jeffro256 (approved):

selsta (approved):

PR #9004, opened 2023-09-26
cmake: require sodium

tobtoht:
It's not an optional dependency. Makes sure static libraries are used for static builds.


CMakeLists.txt
```diff
@@ -1198,8 +1202,8 @@ endif()
 if(PROTOLIB_LIBRARY)
   set(ZMQ_LIB "${ZMQ_LIB};${PROTOLIB_LIBRARY}")
 endif()
-if(SODIUM_LIBRARY)
-  set(ZMQ_LIB "${ZMQ_LIB};${SODIUM_LIBRARY}")
+if(Sodium_FOUND)
+  set(ZMQ_LIB "${ZMQ_LIB};${sodium_LIBRARIES}")
```
vtnerd, line 1206: The documentation does not mention this variable name.

vtnerd: Oh nevermind, its clearly in the CMake files.

vtnerd (approved):

selsta (approved):

PR #9006, opened 2023-09-27
wallet2: fix `on_reorg` callback location

j-berman:
Calls the `on_reorg` callback inside `handle_reorg`, instead of inside `detach_blockchain`.

`detach_blockchain` can be called even where there isn't a reorg (see how `scan_tx` uses `detach_blockchain`). `handle_reorg` should be called every time there is a reorg.


selsta:
Is this relevant for release branch?

j-berman:
The `on_reorg` callback hasn't been added on the release branch yet. Opened #9008 for release (didn't see a PR for it).

jeffro256 (approved):

PR #9011, opened 2023-10-01
wallet2: fix refresh function parameters [release-v0.18]

selsta:
max_blocks is last on master branch

Reported by dynemeloun on IRC.


rbrunner7 (approved):
I am still not sure how that parameter order confusion came to be exactly, but whatever, seems to me this PR is the correct way to go, i.e. make master and release branch match again regarding the order of the parameters of that `refresh` method.

tobtoht (approved):

PR #9013, opened 2023-10-02
Fix ltw cache

Boog900:
The long term block weight cache was doing a wrong calculation when adding new blocks to the cache, which means nodes that have refreshed the cache could have different values stored in the cache than nodes that haven't.


SChernykh (approved):

rbrunner7:
Interesting. @Boog900 , is this find a direct result from your work on Cuprate?

Boog900:
@rbrunner7 yes, although not a direct result of Cuprate, the node, it was a result of going through the code to document and rewriting the block scaling code, for Cuprate, which allowed me to spot this issue.

j-berman (approved):

PR #9023, opened 2023-10-17
wallet: mitigate statistical dependence for decoy selection within rings

jeffro256:
Since we are required to check for uniqueness of decoy picks within any given ring, and since some decoy picks may fail due to unlock time or malformed EC points, the wallet2 decoy selection code was building up a larger than needed *unique* set of decoys for each ring according to a certain distribution *without replacement*. After filtering out the outputs that it couldn't use, it chooses from the remaining decoys uniformly random *without replacement*.

The problem with this is that the picks later in the picking process are not independent from the picks earlier in the picking process, and the later picks do not follow the intended decoy distribution as closely as the earlier picks. To understand this intuitively, imagine that you have 1023 marbles. You label 512 marbles with the letter A, label 256 with the letter B, so on and so forth, finally labelling one marble with the letter J. You put them all into a bag, shake it well, and pick 8 marbles from the bag, but everytime you pick a marble of a certain letter, you remove all the other marbles from that bag with the same letter. That very first pick, the odds of picking a certain marble are exactly how you would expect: you are twice as likely to pick A as you are B, twice as likely to pick B as you are C, etc. However, on the second pick, the odds of getting the first pick are 0%, and the chances for everything else is higher. As you go down the line, your picked marbles will have letters that are increasingly more unlikely to pick if you hadn't removed the other marbles. In other words, the distribution of the later marbles will be more "skewed" in comparison to your original distribution of marbles.

In Monero's decoy selection, this same statistical effect applies. It is not as dramatic since the distribution is not so steep, and we have more unique values to choose from, but the effect *is* measureable. Because of the protocol rules, we cannot have duplicate ring members, so unless that restriction is removed, we will never have perfectly independent picking. However, since the earlier picks are less affected by this statistical effect, the workaround that this commit offers is to store the order that the outputs were picked and commit to this order after fetching output information over RPC.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -8616,6 +8616,26 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
     COMMAND_RPC_GET_OUTPUTS_BIN::request req = AUTO_VAL_INIT(req);
     COMMAND_RPC_GET_OUTPUTS_BIN::response daemon_resp = AUTO_VAL_INIT(daemon_resp);
 
+    // The secret picking order contains outputs in the order that we selected them.
+    //
+    // We will later sort the output request entries in a pre-determined order so that the daemon
+    // that we're requesting information from doesn't learn any information about the true spend
+    // for each ring. However, internally, we want to prefer to construct our rings using the
+    // outputs that we picked first versus outputs picked later.
+    //
+    // The reason why is because each consecutive output pick within a ring becomes increasing less
+    // statistically independent from other picks, since we pick outputs from a finite set
+    // *without replacement*, due to the protocol not allowing duplicate ring members. This effect
+    // is exacerbated by the fact that we pick 1.5x + 75 as many outputs as we need per RPC
+    // request to account for unusable outputs. This effect is small, but non-neglibile and gets
+    // worse with larger ring sizes.
+    std::vector<get_outputs_out> secret_picking_order;
+
+    // Convenience/safety lambda to make sure that both output lists req.outputs and secret_picking_order are updated together
+    // Each ring section of req.outputs gets sorted later after selecting all outputs for that ring
+    const auto add_output_to_lists = [&req, &secret_picking_order](const get_outputs_out &goo)
+      { req.outputs.push_back(goo); secret_picking_order.push_back(goo); };
+
     std::unique_ptr<gamma_picker> gamma;
     if (has_rct)
       gamma.reset(new gamma_picker(rct_offsets));
@@ -8750,7 +8770,7 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
             if (out < num_outs)
             {
               MINFO("Using it");
-              req.outputs.push_back({amount, out});
+              add_output_to_lists({amount, out});
               ++num_found;
               seen_indices.emplace(out);
               if (out == td.m_global_output_index)
@@ -8772,12 +8792,12 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
       if (num_outs <= requested_outputs_count)
       {
         for (uint64_t i = 0; i < num_outs; i++)
-          req.outputs.push_back({amount, i});
+          add_output_to_lists({amount, i});
         // duplicate to make up shortfall: this will be caught after the RPC call,
         // so we can also output the amounts for which we can't reach the required
         // mixin after checking the actual unlockedness
         for (uint64_t i = num_outs; i < requested_outputs_count; ++i)
-          req.outputs.push_back({amount, num_outs - 1});
+          add_output_to_lists({amount, num_outs - 1});
       }
       else
       {
@@ -8786,7 +8806,7 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
         {
           num_found = 1;
           seen_indices.emplace(td.m_global_output_index);
-          req.outputs.push_back({amount, td.m_global_output_index});
+          add_output_to_lists({amount, td.m_global_output_index});
           LOG_PRINT_L1("Selecting real output: " << td.m_global_output_index << " for " << print_money(amount));
         }
 
@@ -8894,7 +8914,7 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
           seen_indices.emplace(i);
 
           picks[type].insert(i);
-          req.outputs.push_back({amount, i});
+          add_output_to_lists({amount, i});
           ++num_found;
           MDEBUG("picked " << i << ", " << num_found << " now picked");
         }
@@ -8908,7 +8928,7 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
         // we'll error out later
         while (num_found < requested_outputs_count)
         {
-          req.outputs.push_back({amount, 0});
+          add_output_to_lists({amount, 0});
           ++num_found;
         }
       }
@@ -8918,6 +8938,10 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
           [](const get_outputs_out &a, const get_outputs_out &b) { return a.index < b.index; });
     }
 
+    THROW_WALLET_EXCEPTION_IF(req.outputs.size() != secret_picking_order.size(), error::wallet_internal_error,
+        "bug: we did not update req.outputs/secret_picking_order in tandem");
+
+    // List all requested outputs to debug log
     if (ELPP->vRegistry()->allowed(el::Level::Debug, MONERO_DEFAULT_LOG_CATEGORY))
     {
       std::map<uint64_t, std::set<uint64_t>> outs;
@@ -9035,18 +9059,21 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
         }
       }
 
-      // then pick others in random order till we reach the required number
-      // since we use an equiprobable pick here, we don't upset the triangular distribution
-      std::vector<size_t> order;
-      order.resize(requested_outputs_count);
-      for (size_t n = 0; n < order.size(); ++n)
-        order[n] = n;
-      std::shuffle(order.begin(), order.end(), crypto::random_device{});
-
+      // While we are still lacking outputs in this result ring, in our secret pick order...
       LOG_PRINT_L2("Looking for " << (fake_outputs_count+1) << " outputs of size " << print_money(td.is_rct() ? 0 : td.amount()));
-      for (size_t o = 0; o < requested_outputs_count && outs.back().size() < fake_outputs_count + 1; ++o)
+      for (size_t ring_pick_idx = base; ring_pick_idx < base + requested_outputs_count && outs.back().size() < fake_outputs_count + 1; ++ring_pick_idx)
       {
-        size_t i = base + order[o];
+        const get_outputs_out attempted_output = secret_picking_order[ring_pick_idx];
+
+        // Find the index i of our pick in the request/response arrays
+        size_t i;
+        for (i = base; i < base + requested_outputs_count; ++i)
+          if (req.outputs[i].index == attempted_output.index)
+            break;
+        THROW_WALLET_EXCEPTION_IF(i == base + requested_outputs_count, error::wallet_internal_error,
+          "Could not find index of picked output in requested outputs");
+
+        // Try adding this output's information to result ring if output isn't invalid
         LOG_PRINT_L2("Index " << i << "/" << requested_outputs_count << ": idx " << req.outputs[i].index << " (real " << td.m_global_output_index << "), unlocked " << daemon_resp.outs[i].unlocked << ", key " << daemon_resp.outs[i].key);
         tx_add_fake_output(outs, req.outputs[i].index, daemon_resp.outs[i].key, daemon_resp.outs[i].mask, td.m_global_output_index, daemon_resp.outs[i].unlocked, valid_public_keys_cache);
       }


jeffro256:
@Rucknium

src/wallet/wallet2.cpp
```diff
@@ -9035,18 +9059,21 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>>
...
-        order[n] = n;
-      std::shuffle(order.begin(), order.end(), crypto::random_device{});
-
+      // While we are still lacking outputs in this result ring, in our secret pick order...
       LOG_PRINT_L2("Looking for " << (fake_outputs_count+1) << " outputs of size " << print_money(td.is_rct() ? 0 : td.amount()));
-      for (size_t o = 0; o < requested_outputs_count && outs.back().size() < fake_outputs_count + 1; ++o)
+      for (size_t ring_pick_idx = base; ring_pick_idx < base + requested_outputs_count && outs.back().size() < fake_outputs_count + 1; ++ring_pick_idx)
       {
-        size_t i = base + order[o];
+        const get_outputs_out attempted_output = secret_picking_order[ring_pick_idx];
```
vtnerd, line 9066: I don't see how this is an improvement - if a dummy index is selected twice, why would selecting the earlier index matter (when they should be identical)?

This change should have basically no observable effects on the behavior, but does add an unnecessary `std::vector` to the processing loop.

jeffro256: What do you mean by "dummy index"?

vtnerd: Each output is referenced by index within an amount value. A dummy index is an output (index) that is not the "real" spend.

jeffro256: Ah okay, I personally use the term decoy, but that's unimportant. 

> if a dummy index is selected twice, why would selecting the earlier index matter (when they should be identical)?

That's the thing, any decoy index is never selected twice because we specifically choose not to do that; see the variable `seen_indices`. We only ever choose *unique* decoy indexes within any given ring, and more than the current ring size at that, assuming that some will be invalid. Since we are ever only choosing unique decoy index values, each successive decoy index value depends upon all previous decoy index values. See this analogy:

> ... imagine that you have 1023 marbles. You label 512 marbles with the letter A, label 256 with the letter B, so on and so forth, finally labelling one marble with the letter J. You put them all into a bag, shake it well, and pick 8 marbles from the bag, but everytime you pick a marble of a certain letter, you remove all the other marbles from that bag with the same letter. That very first pick, the odds of picking a certain marble are exactly how you would expect: you are twice as likely to pick A as you are B, twice as likely to pick B as you are C, etc. However, on the second pick, the odds of getting the first pick are 0%, and the chances for everything else is higher. As you go down the line, your picked marbles will have letters that are increasingly more unlikely to pick if you hadn't removed the other marbles. In other words, the distribution of the later marbles will be more "skewed" in comparison to your original distribution of marbles.

If you perform this experiment multiple times, you will notice that you get "too many" J, H, & I marbles compared to if you picked with replacement. This effect gets worse towards the end since you are exhausting choices for each successive pick. For the first pick you have 10 choices, then 9, then 8, then 7, etc, until the 10th choice when you have picked exactly one marble for each letter A-J. This sampling is extremely different from the sampling of marbles initially in the bag.

In our case, the initial sampling of marbles in the bag represents our ideal target distribution. But whenever we use `seen_indices` to enforce unique values for our picks, we start warping the distribution towards a uniform distribution the longer we pick. As of writing this comment, there are 82037944 RingCT outputs on-chain. If we were to pick 82037944 *unique* RingCT outputs, we would end up with a distribution that looks nothing like a gamma distribution. If we picked 40 million unique RingCT outputs, it also would look nothing like a gamma distribution. However, the picks at the front of the list would follow the distribution more closely since they have fewer value constraints.

> This change should have basically no observable effects on the behavior

It does have a small, but observable effect

jeffro256: B/c discarding duplicates measurably changes the distribution the more and more you pick. Would it be beneficial to empirically show this observation by releasing my source code for the aggregate decoy selection distribution?

Gingeropolous:
> Since we are required to check for uniqueness of decoy picks within any given ring, and since some decoy picks may fail due to unlock time or malformed EC points, the wallet2 decoy selection code was building up a larger than needed unique set of decoys for each ring according to a certain distribution without replacement. After filtering out the outputs that it couldn't use, it chooses from the remaining decoys uniformly random without replacement.

So the wallet asks the daemon for a bunch of outputs, the initial pick (Pick A). Then it goes through these and filters out bad outputs, and then makes the final pick (Pick B). Is this the way it goes? 

Is Pick B where the lack of replacement is occurring, and where the marble analogy comes in? 

> the workaround that this commit offers is to store the order that the outputs were picked and commit to this order after fetching output information over RPC.

How does storing the order the outputs were picked mitigate the issue?

My apologies if I'm not fully understanding the details. However, if this effect is measurable at ringsize 16, I bet it will be amplified when we move to 128 or 256. 

Finally, is unlock time the primary uniquifier (aspect that makes an enote unique) that is filtered out?

src/wallet/wallet2.cpp
vtnerd: No, I forgot about the "overselection" portion. If there was no overselection, then this wouldn't be doing anything.

jeffro256: There's still measurable dependence even if we don't over-select, but there's nothing we can do about that unless we want to hard-fork to allow duplicate ring members.

Gingeropolous:
> Mainly the issue is in part Pick A, because picking from a sample **uniformly** to form a subsample is not a problem if the original sample represents the population correctly, which it doesn't in this case. 

I thought Pick A is picking from the sample according to the gamma distribution?

This is from your DECOY_SELECTION.MD #9024 

> When you are trying to build up a set of X unique decoy picks, if the first pick has 100 choices, then the next pick has 99 choices, then 98 choices, etc, etc. Since these picks are not statistically independent, then the distribution of the picks gets more and more skewed for the later picks. 

So this is talking about the primary pick (pick A), from the blockchain? Or the secondary pick (pick B), from the subsample?

Perhaps the following is it: during the primary pick, when the machine is selecting an enote from the blockchain following a gamma distribution, the machine picks an output in block X and stores it as Pick P1 in the subsample. Now, when the machine goes to find another enote P2, it uses the gamma distribution to pick another transaction, but this time it can't pick from the region that P1 was located (the block that A1 is in), because it needs to be unique. 

To go back to the marbles, pick P1 would be the marbles labeled A. Pick P2 would be marbles labeled B.

So then when the secondary pick occurs (pick S), and this time the machine is picking through the subsample using a uniform distribution (also, wat? well I guess that makes sense), if no weights is given to P1-P99, it will generally just pick more things at the tail end of that (P50-P99). 

I guess my hangup is why P1-P99 end up being so... something. If P1 is in block 400, could P2 be in block 401? 

ran out of time to finish my thoughts here....

jeffro256:
> I thought Pick A is picking from the sample according to the gamma distribution?

It is, **but** only picking unique values, which is the core of the problem. I was making the point that *if* Pick A was picking uniformly (and unique) across all outputs, there would be no difference, but since Pick A picks to a gamma distribution, then we have warping. The second stage of picking uniformly would also make no difference if all picks in stage A were picked independently from each other, but they're not since they're all unique. 

> So this is talking about the primary pick (pick A), from the blockchain? Or the secondary pick (pick B), from the subsample?

Pick A

> but this time it can't pick from the region that P1 was located (the block that A1 is in), because it needs to be unique.

It can pick from the same block, just not the same enote

> if no weights is given to P1-P99, it will generally just pick more things at the tail end of that (P50-P99)

yes, since we didn't store the order earlier, just picking subsamples at random, the later picks have just as much chance as being picked as the earlier picks

> I guess my hangup is why P1-P99 end up being so... something.

Sorry I'm not sure I understand ;)

> If P1 is in block 400, could P2 be in block 401?

Yes, the blocks don't matter at all; we just were going for unique enote indices

Gingeropolous:
I guess I don't understand the marble analogy then. During the primary pick, it's just picking a unique enote, so its P1. Then it makes a second P2. Now, P2 can't be the same as P1, but it only removed 1 marble.

So lets simplify the blockchain to 1e6 enotes. P1 picks enote at index 1e3. P2 can now only be picked from 1e6-1 potential enotes. 

In your marble analogy, you start with 1023 enotes. The machine enote at index 99 as P1. You state that P2 now can only be selected from 1023-512 potential outputs. Why?

All the outputs are unique. There's nothing about P1 that would make it the same as 512 other enotes on the chain.

This is what i'm not getting.

jeffro256:
> but it only removed 1 marble

We don't just remove that marble, we remove ALL marbles with that letter. So going back to the analogy, if we pick marble with the letter A, we remove 512 marbles, not 1. 

The marbles don't represent outputs themselves, they represent units of probability proportional to how likely a unique value will be picked.

The reason I worded it like that is because different "letters" (or output indices) have different probabilities if being picked because of the gamma distribution. This might be my fault for not making that clearer. 

> All the outputs are unique. There's nothing about P1 that would make it the same as 512 other enotes on the chain.

Yes, but the "correct"/independent way of picking would be to not keep track of the set of already picked outputs and enforce unique picks. This would make all picks truly independent from each other. In this case, we wouldn't be picking only "unique" values, we would be picking whatever we got. Maybe they would happen to all be unique, but maybe we might double-pick a certain output index. The reason we don't do it this way is because the protocol consensus rules dictate that rings cannot have duplicate members (i.e. all ring members must be unique).

Gingeropolous:
I think I see it. So if there was no subsampling (primary sampling, pick A) and then final sampling (second sampling, pick B), then this wouldn't be necessary. The final outputs selected would match a gamma distribution more nicely.

As it stands now, because the primary sampling picks 99 outputs (or whatever), a majority of them are going to end up being low probability picks, because during the primary picking, the higher probability picks get picked first? 

So the machine goes "I need to get an output of this particular probability", and it picks it. And then it goes "ok, I can't use any of those outputs with that probability". so it moves on to the next probability? 

I still don't think i get your unique thing. All outputs are unique compared to other outputs...... right?

jeffro256:
> As it stands now, because the primary sampling picks 99 outputs (or whatever), a majority of them are going to end up being low probability picks, because during the primary picking, the higher probability picks get picked first?

Yes

> So the machine goes "I need to get an output of this particular probability", and it picks it. And then it goes "ok, I can't use any of those outputs with that probability". so it moves on to the next probability?

Yeah it restarts the picking process. So If you were to visualize the PDF of the outputs distribution, whenever you pick a certain output, the probability disappears from that output and gets redistributed to the rest of the distribution. This has a non-linear warping effect.

>  still don't think i get your unique thing. All outputs are unique compared to other outputs...... right?

Yes. But if we we picking independently, we might pick output 81234567 two times, which means that our *picks* wouldn't be unique.

vtnerd (approved):

Rucknium:
## Summary

This PR reduces a distortion of the intended probability mass function of the decoy selection algorithm. The distortion is real but minor. The distortion would be greater if transaction volume was lower and/or if ring size increases.

I support this PR. The fact that the distortion is minor is actually a good reason to implement it now instead of waiting until the distortion become larger.

In ["Discussion Note: Formula for Accuracy of Guessing Monero Real Spends Using Fungibility Defects"](https://github.com/Rucknium/misc-research/tree/main/Monero-Fungibility-Defect-Classifier/pdf) I explain how differences in transaction format can reduce privacy. The problem only exists if an adversary can actually distinguish between transactions. Very small differences in the decoy selection algorithm would not be statistically distinguishable, especially for small ring sizes. Therefore, now we would not have a problem with having two slightly different decoy selection algorithms (the current code and the modification from this PR) being used by two different `wallet2` version by sets of people who did and did not update to this version. If this PR is implemented at a time when the distortion is larger, there could be a negative privacy impact from the distinguishability of the two `wallet2` versions.


## Kolmogorov-Smirnov tests of distortion of the distributions

The [Kolmogorov-Smirnov (KS) test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) is a way to measure the difference between an empirical and a theoretical distribution (or the difference between two empirical distributions) and test whether the difference is statistically significant (i.e. likely a result of sampling error or a true difference). The KS test produces a KS statistic that is the vertical difference between the two Cumulative Distribution Functions (CDFs) being compared. The maximum possible value of the KS statistic would occur if the two distributions' support does not overlap at all. The maximum value is 1. The minimum possible value is 0 when the two CDFs are exactly the same. The KS test also produces a p-value that has the usual interpretation of a p-value of a statistical test. The null hypothesis of the p-value is that the two distributions are the same.

## Evaluating different scenarios

We want to see what would happen if ring size increases to 128 with Seraphis and/or transaction volume per second decreases.

Currently with ring size 16, the number of initial candidate outputs that `wallet2` selects is 75. If the same formula is used for ring size 128, the number of candidate outputs would be 243.

Transaction volume matters since lower volume would concentrate more probability in the initial candidate selection into fewer outputs. The fewer outputs are the more recent outputs that have a greater probability of being selected in the initial pick.

The draft ["Implementing Monero Decoy Selection"](https://github.com/jeffro256/monero/blob/decoy_selection_md/docs/DECOY_SELECTION.md) says "`average_output_delay` is meant to quantify the average number of seconds between new transaction outputs over the last year or so of chain data." The `average_output_delay` can be scaled to artificially modify transaction volume to simulate what `wallet2` would select if transaction volume was much lower. For example, if current `average_output_delay` is 1.4 (which means that an average of 1.4 seconds elapsed between each new output in the last year of blockchain data), then `average_output_delay` can be divided by 0.1 to simulate what `wallet2` would do when transaction volume is only 10% of current levels (i.e. 14 seconds elapse between each new output, on average).

## Test procedure

The procedure I followed is:

1) Compute the probability mass function (PMF) of the `wallet2` decoy selection algorithm (DSA) based on my draft ["Closed-form Expression of the wallet2 Decoy Selection Algorithm"](https://www.overleaf.com/read/ndbtkwrbrdjq). We treat all outputs after the 10 block lock as spendable. The block height of this DSA is 3,000,000. There are about 80 million outputs eligible for being spent in a RingCT ring at this height.

2) `N` output indices are drawn without replacement from the PMF computed in step (1) to create a set of candidate decoys. For computational speed, algorithm A-ExpJ of Efraimidis & Spirakis (2006) implemented in the `wrswoR` R package (Müller 2020) was used to draw output indices without replacement but with unequal probability defined by the PMF.

3a) `M` elements are selected with uniform probability without replacement from the ordered set of candidate decoys to simulate the original method of selecting final decoys.

3b) The first `M` elements are selected from the ordered set of candidate decoys to simulate the PR 9023 method of selecting final decoys. In other words, the 1st, 2nd, 3rd,...,`M`th elements of the candidate set of decoys is selected.

4) Repeat steps (2)-(3b) one million times to create two sets of one million rings using the original and PR 9023 method.

5) Create a Cumulative Distribution Function (CDF) of the `wallet2` DSA by computing the cumulative sum of the PMF computed in step (1). Gather all rings created with the original method into one set (i.e. create a set of size `M*1,000,000`) and perform a Kolmogorov-Smirnov test with `wallet2` DSA's CDF as the reference CDF. Do the same with the rings created with the PR 9023 method.

When ring size is 16, `N = 75` and `M = 15`. When ring size is 128, `N = 243` and `M = 127`.

## Results

Below is the table of results, with p-values in parentheses.

| ring_size| tx_volume|   KS_statistic_original_method|   KS_statistic_PR_9023_method|   ratio|
|---------:|---------:|------------------------------:|-----------------------------:|-------:|
|        16|      100%|       `0.00035976 (0.0411808)`|      `0.00015249 (0.8764715)`| `2.359`|
|        16|       10%|       `0.00164921 ( < 2.2-16)`|      `0.00055049 (0.0002253)`| `2.996`|
|       128|      100%|       `0.00053995 ( < 2.2-16)`|      `0.00032783 (2.789e-12)`| `1.647`|
|       128|       10%|       `0.00517094 ( < 2.2-16)`|      `0.00270993 ( < 2.2-16)`| `1.908`|

All KS statistics are below 0.01. All differences between the decoy selection methods and the intended DSA CDF are minor. The PR 9023 method is closer than the original method to the intended DSA CDF in all four scenarios. In the ring size 16 and 100% current transaction volume scenario, the difference between the decoys selected with the PR 9023 method and the intended DSA CDF is not even statistically significant (the p-value is 0.88). Note that larger ring size will produce lower p-values because the aggregate sample used to perform the KS test will be larger.

## Conclusion

Selecting multiple elements without replacement from a non-uniform probability mass function will create an empirical distribution that is distorted away from the intended probability mass function and distorted toward a uniform distribution. Having a candidate set of elements and then selecting randomly among them with equal probability will intensify the distortion toward a uniform distribution. This is the current `wallet2` behavior. The distortion will be lower if the first `M` elements of the candidate set are chosen, which is the procedure implemented in this PR 9023. The distortion cannot be completely eliminated as long as elements are selected without replacement.

## References

Efraimidis, Pavlos S., and Paul G. Spirakis. "Weighted random sampling with a reservoir." Information Processing Letters 97, no. 5 (2006): 181-185.

Müller K (2020). _wrswoR: Weighted Random Sampling without Replacement_. R package version 1.1.1, <https://CRAN.R-project.org/package=wrswoR>.

## R Code

```R

install.packages(c("RCurl", "RJSONIO", "actuar", "future",
  "future.apply", "knitr", "wrswoR", "crosstable"))


monerod.address <- "http://127.0.0.1:18081/json_rpc"


# Modified from TownforgeR::tf_rpc_curl function
xmr.rpc <- function(
    url.rpc = "http://127.0.0.1:18081/json_rpc",
  method = "",
  params = list(),
  userpwd = "",
  num.as.string = FALSE,
  nonce.as.string = FALSE,
  keep.trying.rpc = FALSE,
  curl = RCurl::getCurlHandle(),
  ...
){

  json.ret <- RJSONIO::toJSON(
    list(
      jsonrpc = "2.0",
      id = "0",
      method = method,
      params = params
    ), digits = 50
  )

  rcp.ret <- 	tryCatch(RCurl::postForm(url.rpc,
    .opts = list(
      userpwd = userpwd,
      postfields = json.ret,
      httpheader = c('Content-Type' = 'application/json', Accept = 'application/json')
      # https://stackoverflow.com/questions/19267261/timeout-while-reading-csv-file-from-url-in-r
    ),
    curl = curl
  ), error = function(e) {NULL})

  if (keep.trying.rpc && length(rcp.ret) == 0) {
    while (length(rcp.ret) == 0) {
      rcp.ret <- 	tryCatch(RCurl::postForm(url.rpc,
        .opts = list(
          userpwd = userpwd,
          postfields = json.ret,
          httpheader = c('Content-Type' = 'application/json', Accept = 'application/json')
          # https://stackoverflow.com/questions/19267261/timeout-while-reading-csv-file-from-url-in-r
        ),
        curl = curl
      ), error = function(e) {NULL})
    }
  }

  if (is.null(rcp.ret)) {
    stop("Cannot connect to monerod. Is monerod running?")
  }

  if (num.as.string) {
    rcp.ret <- gsub("(: )([-0123456789.]+)([,\n\r])", "\\1\"\\2\"\\3", rcp.ret )
  }

  if (nonce.as.string & ! num.as.string) {
    rcp.ret <- gsub("(\"nonce\": )([-0123456789.]+)([,\n\r])", "\\1\"\\2\"\\3", rcp.ret )
  }

  RJSONIO::fromJSON(rcp.ret, asText = TRUE) # , simplify = FALSE
}


# curl 127.0.0.1:18081/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"get_output_distribution","params":{"amounts":[0],"from_height":0,"to_height":3000000,"binary":false,"cumulative":true}}' -H 'Content-Type: application/json'

current.height <- 3000000

crod <- xmr.rpc(url.rpc = monerod.address, method = "get_output_distribution",
  params = list(amounts = list(0), from_height = 0, to_height = current.height,
    binary = FALSE, cumulative = TRUE))

start_height.RingCT <- crod$result$distributions[[1]]$start_height
crod <- crod$result$distributions[[1]]$distribution


get.dsa.pmf <- function(block.height, crod, output.index.only.locked, tx.volume.factor = 1) {

  CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE = 10
  DIFFICULTY_TARGET_V2 = 120

  SECONDS_IN_A_YEAR =  60 * 60 * 24 * 365
  BLOCKS_IN_A_YEAR = SECONDS_IN_A_YEAR / DIFFICULTY_TARGET_V2

  calculate_average_output_flow <- function(crod) {
    # 1
    num_blocks_to_consider_for_flow = min(c(length(crod), BLOCKS_IN_A_YEAR))

    # 2
    if (length(crod) > num_blocks_to_consider_for_flow) {
      num_outputs_to_consider_for_flow = crod[length(crod)] - crod[ length(crod) - num_blocks_to_consider_for_flow ]
      # R indexes from 1
    } else {
      num_outputs_to_consider_for_flow = crod[length(crod)] # R indexes from 1
    }

    # 3
    average_output_flow = DIFFICULTY_TARGET_V2 * num_blocks_to_consider_for_flow / num_outputs_to_consider_for_flow

    return(average_output_flow)
  }

  calculate_num_usable_rct_outputs <- function(crod) {
    # 1
    num_usable_crod_blocks = length(crod) - (CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE - 1)

    # 2
    num_usable_rct_outputs = crod[num_usable_crod_blocks] # R indexes from 1

    return(num_usable_rct_outputs)
  }


  crod.at.tx.construction <- crod[1:(block.height - start_height.RingCT)]

  average_output_flow <- calculate_average_output_flow(crod.at.tx.construction)
  num_usable_rct_outputs <- calculate_num_usable_rct_outputs(crod.at.tx.construction)

  average_output_flow <- average_output_flow/tx.volume.factor
  # This is seconds per output

  v <- average_output_flow
  z <- num_usable_rct_outputs


  GAMMA_SHAPE = 19.28
  GAMMA_RATE = 1.61

  CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE = 10
  DIFFICULTY_TARGET_V2 = 120
  DEFAULT_UNLOCK_TIME = CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE * DIFFICULTY_TARGET_V2
  RECENT_SPEND_WINDOW = 15 * DIFFICULTY_TARGET_V2

  G <- function(x) {
    actuar::plgamma(x, shapelog = GAMMA_SHAPE, ratelog = GAMMA_RATE)
  }

  G_star <- function(x) {
    (0 <= x*v & x*v <= 1800) *
      (G(x*v + 1200) - G(1200) +
          ( (x*v)/(1800) ) * G(1200)
      )/G(z*v) +
      (1800 < x*v & x*v <= z*v) * G(x*v + 1200)/G(z*v) +
      (z*v < x*v) * 1
  }


  crod.reversed <- abs(diff(rev(c(0, crod.at.tx.construction))))[-(1:(CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE-1))]
  # Remove first 10 blocks before cumsum() since can't spend from those outputs
  crod.reversed <- cumsum(crod.reversed[crod.reversed > 0])

  crod.reversed <- c(0, crod.reversed)
  # Concatenate zero since the CDF should start at 0

  y_0 <- crod.reversed[-length(crod.reversed)] + 1
  y_1 <- crod.reversed[-1]
  dsa.pmf <- (G_star(y_1 + 1) - G_star(y_0)) / (y_1 + 1 - y_0)

  dsa.pmf <- rep(dsa.pmf, times = diff(crod.reversed))
  dsa.pmf[length(dsa.pmf)] <- dsa.pmf[length(dsa.pmf) - 1]
  # Replace oldest output's mass by the penultimate one since
  # it is calculated as a negative mass otherwise

  output.index.only.locked <- copy(output.index.only.locked)
  if (!is.null(output.index.only.locked)) {
    unspendable.indices <- output.index.only.locked[
      (output_index + 1) <= num_usable_rct_outputs &
        output_unlock_time >= block.height,
      output_index]

    stopifnot(all(unspendable.indices < num_usable_rct_outputs))
    unspendable.indices <- abs(unspendable.indices - num_usable_rct_outputs) + 1
  } else {
    unspendable.indices <- NULL
  }

  sum.of.unspendable.mass <- sum(dsa.pmf[unspendable.indices])

  dsa.pmf[unspendable.indices] <- 0

  dsa.pmf <- dsa.pmf * 1/(1 - sum.of.unspendable.mass)

  return(dsa.pmf)

}




n.rings <- 1000000
# One million

results <- data.frame(
  ring_size = c(16, 16, 128, 128),
  tx_volume = c("100%", "10%", "100%", "10%"),
  KS_statistic_original_method = NA,
  KS_statistic_PR_9023_method = NA,
  ratio = NA,
  stringsAsFactors = FALSE)


future::plan(future::multicore, workers = future::availableCores(constraints = "multicore"))
# Use future::multisession if using Windows or RStudio

options(future.globals.maxSize = 1000000000)

for ( i in 1:nrow(results)) {

  dsa.pmf <- get.dsa.pmf(block.height = current.height, crod = crod,
    output.index.only.locked = NULL,
    tx.volume.factor = as.numeric(gsub("%", "", results$tx_volume[i]))/100)

  if (results$ring_size[i] == 16) {
    N <- 75
    M <- 15
  }

  if (results$ring_size[i] == 128) {
    N <- 243
    M <- 127
  }

  set.seed(314)

  picked.decoys <- future.apply::future_replicate(n.rings, {
    candidate.outputs <- wrswoR::sample_int_expj(length(dsa.pmf), N, prob = dsa.pmf)
    return(c(
      sample(candidate.outputs, M, replace = FALSE),
      candidate.outputs[1:M]
    ))
  })

  picked.decoys.original.method <- c(picked.decoys[1:M, ])
  picked.decoys.PR.9023.method <- c(picked.decoys[(M+1):(2*M), ])
  # Matrix is M x n.rings

  rm(picked.decoys)

  ks.test.original.method <- ks.test(picked.decoys.original.method,
    stepfun(seq_along(dsa.pmf), c(0, cumsum(dsa.pmf))))

  ks.test.PR.9023.method <- ks.test(picked.decoys.PR.9023.method,
    stepfun(seq_along(dsa.pmf), c(0, cumsum(dsa.pmf))))

  print(ks.test.original.method)
  print(ks.test.PR.9023.method)

  if (ks.test.original.method$p.value < 2.2e-16) {
    ks.test.original.method$p.value <- " < 2.2-16"
  } else {
    ks.test.original.method$p.value <- crosstable::format_fixed(ks.test.original.method$p.value, digits = 7)
  }

  if (ks.test.PR.9023.method$p.value < 2.2e-16) {
    ks.test.PR.9023.method$p.value <- " < 2.2-16"
  } else {
    ks.test.PR.9023.method$p.value <- crosstable::format_fixed(ks.test.PR.9023.method$p.value, digits = 7)
  }

  results[i, "KS_statistic_original_method"] <-
    with(ks.test.original.method,  paste0(crosstable::format_fixed(statistic, digits = 8), " (", p.value, ")"))

  results[i, "KS_statistic_PR_9023_method"] <-
    with(ks.test.PR.9023.method, paste0(crosstable::format_fixed(statistic, digits = 8), " (", p.value, ")"))

  results[i, "ratio"] <- crosstable::format_fixed(
    ks.test.original.method$statistic / ks.test.PR.9023.method$statistic, digits = 3)

}

print(results)

knitr::kable(results, format = "pipe")

```

PR #9035, opened 2023-10-23
ringct: make `rctSigBase` serialization follow strict aliasing rule

jeffro256:
Accessing an object of type `char` thru an lvalue of type `crypto::hash8` is undefined behavior. https://developers.redhat.com/blog/2020/06/03/the-joys-and-perils-of-aliasing-in-c-and-c-part-2


jeffro256:
@vtnerd

vtnerd (approved):

PR #9042, opened 2023-10-27
JH hash compiler workarounds

### LINKED ISSUE #9041: bug in cn_slow_hash function, jh_hash cause the bug
I find that there is a bug in cn_slow_hash. When I compiled code use gcc 11.4 on ubuntu22. The compiler flag O2 and O3 get different results. The reason is caused by jh_hash. This is the code in [main.c](https://github.com/nkysg/jh-function-issue/blob/monero/src/main.c)
```c
#include <stdio.h>
#include "jh.h"
int main() {
        unsigned char input[] = "\xd6\xa1\x1a\x43\x1f\xd5\x6d\xc2\x9c\x59\x27\x32\x8e\x83\xee\xf1\x15\x90\x36\x2b"
                       "\x70\xae\xf4\x92\xf1\x2c\x44\x9b\xde\x65\x7b\x4e\x3a\xc4\xe1\x58\x16\x31\x46\x13"
                       "\xca\x73\x39\xd1\x52\x29\x4b\x18\x5e\xd8\x67\x09\xac\xc1\x26\x62\x59\x58\x0d\x4a"
                       "\xba\x72\x9d\x37\xc7\x33\xa8\x9e\xb1\x5e\xa9\x1b\x28\xbb\xe4\x9b\xd1\x51\xcb\x1b"
                       "\x94\x7b\xb2\x39\xc4\xff\x10\xcf\xe2\x4a\xa0\x5f\x31\xa8\xcd\xb8\xe7\x13\x45\xea"
                       "\x45\xcf\xfe\xff\x6b\xbf\x9d\x2c\x7f\xe5\x8f\x13\xf1\x90\xb2\x92\x1a\x53\x1f\xb6"
                       "\x58\x41\x22\x1d\x3e\x3e\x84\x83\x83\x0a\x9b\x4a\x9c\x2e\x28\x43\xe7\x4d\xb1\x79"
                       "\x2c\x73\x9d\x3e\xe2\x7a\x97\x3b\x61\xc8\x38\xf3\x00\xa5\x49\x51\xcf\x8a\xc5\xfe"
                       "\x99\xc2\xce\xd0\x5d\x8b\x4c\xff\xa2\x0a\x24\xf8\x4d\xba\x4a\x45\x28\xb2\xb2\xcf"
                       "\x03\x5a\xe5\xb0\x6b\x49\xf1\x23\x64\xfc\xb5\x77\x6d\xb3\x69\x2b\xfb\x6f\x02\x9d";

        unsigned char output[32] = {'\0'};
        jh_hash(32 * 8, input, sizeof(input) * 8, output);

	//output should be 0xea252e4fe0d223d17925f61058e2809da4896f12db26fce35faa5534575b8ce0
        printf("0x");
        for (int i = 0; i < 32; i++) {
                printf("%02x",output[i]);
        }
        printf("\n");
        return 0;
}
```
the Makefile is this https://github.com/nkysg/jh-function-issue/blob/monero/src/Makefile
```
main2:
	gcc -O2 -c jh.c -o jh2.o
	gcc -O2 -c main.c -o main2.o
	gcc main2.o jh2.o -o main2
main3:
	gcc -O3 -c jh.c -o jh3.o
	gcc -O3 -c main.c -o main3.o
	gcc main3.o jh3.o -o main3
clean:
	rm -f main2 main3  *.o
```
run ./main2 it print 0xea252e4fe0d223d17925f61058e2809da4896f12db26fce35faa5534575b8ce0
run ./main3 it print 0x48e84a3a785b8d5038d46c46ad4a09e2b8e019a6a8450acec7f36851c8dab0a7

the fact is that monero jh.c is same code as https://www3.ntu.edu.sg/home/wuhj/research/jh/jh_ansi_opt64.h. I have tested the https://www3.ntu.edu.sg/home/wuhj/research/jh/jh_sse2_opt64.h compile use O2 and O3 the result is same. 
If anyone has a plan to fix this bug. Thanks

SChernykh:
- Fixed uninitialized `state->x` warning
- Fixed broken code with `-O3` or `-Ofast`

The old code is known to break GCC 10.1 and GCC 11.4

Closes #9041

### DIFF
--- a/src/crypto/jh.c
+++ b/src/crypto/jh.c
@@ -34,7 +34,7 @@ typedef struct {
 	unsigned long long databitlen;    /*the message size in bits*/
 	unsigned long long datasize_in_buffer;      /*the size of the message remained in buffer; assumed to be multiple of 8bits except for the last partial block at the end of the message*/
 	DATA_ALIGN16(uint64 x[8][2]);     /*the 1024-bit state, ( x[i][0] || x[i][1] ) is the ith row of the state in the pseudocode*/
-	unsigned char buffer[64];         /*the 512-bit message block to be hashed;*/
+	DATA_ALIGN16(unsigned char buffer[64]);         /*the 512-bit message block to be hashed;*/
 } hashState;
 
 
@@ -213,16 +213,24 @@ static void E8(hashState *state)
 /*The compression function F8 */
 static void F8(hashState *state)
 {
-      uint64  i;
+      uint64_t* x = (uint64_t*)state->x;
 
       /*xor the 512-bit message with the fist half of the 1024-bit hash state*/
-      for (i = 0; i < 8; i++)  state->x[i >> 1][i & 1] ^= ((uint64*)state->buffer)[i];
+      for (int i = 0; i < 8; ++i) {
+            uint64 b;
+            memcpy(&b, &state->buffer[i << 3], sizeof(b));
+            x[i] ^= b;
+      }
 
       /*the bijective function E8 */
       E8(state);
 
       /*xor the 512-bit message with the second half of the 1024-bit hash state*/
-      for (i = 0; i < 8; i++)  state->x[(8+i) >> 1][(8+i) & 1] ^= ((uint64*)state->buffer)[i];
+      for (int i = 0; i < 8; ++i) {
+            uint64 b;
+            memcpy(&b, &state->buffer[i << 3], sizeof(b));
+            x[i + 8] ^= b;
+      }
 }
 
 /*before hashing a message, initialize the hash state as H0 */
@@ -240,6 +248,7 @@ static HashReturn Init(hashState *state, int hashbitlen)
             case 224: memcpy(state->x,JH224_H0,128); break;
             case 256: memcpy(state->x,JH256_H0,128); break;
             case 384: memcpy(state->x,JH384_H0,128); break;
+            default:
             case 512: memcpy(state->x,JH512_H0,128); break;
       }
 


src/crypto/jh.c
```diff
@@ -213,16 +213,17 @@ static void E8(hashState *state)
 /*The compression function F8 */
 static void F8(hashState *state)
 {
-      uint64  i;
+      uint64_t* x = (uint64_t*)state->x;
+      const uint64_t* buf = (uint64*)state->buffer;
```
jeffro256, line 217: This is technically undefined behavior according to the strict aliasing rule.

SChernykh: Casting `uint8_t` buffer to `uint64_t` is technically UB, yes. But since we only XOR bytes here, it will not be affected by endianness. Compiling with `-fsanitize=undefined` doesn't find anything in this code. I will also add a proper data alignment for `buffer` to be sure we can access it as `uint64_t` without doing unaligned accesses.

SChernykh: Also, even the old code did the same thing:
```
state->x[(8+i) >> 1][(8+i) & 1] ^= ((uint64*)state->buffer)[i];
```

jeffro256: Even if it is correctly aligned, there is no guarantee by the language that cached, deferenced bytes will actually reflect the contents of `state->buffer` since the aliases are of non-compatible types. If we insert a `memcpy`, we can guarantee portability, and most compilers will optimize the superfluous instruction out anyways. I haven't looked into it, but the line `state->x[(8+i) >> 1][(8+i) & 1] ^= ((uint64*)state->buffer)[i];` could very well be one of the reasons that the old code broke on different optimization levels.

So instead of:
```
uint64_t* x = (uint64_t*)state->x;
const uint64_t* buf = (uint64*)state->buffer;

for (int i = 0; i < 8; ++i) x[i] ^= buf[i];

E8(state);

for (int i = 0; i < 8; ++i) x[i + 8] ^= buf[i];
```

I say we do:
```
uint64* x = state->x;

for (i = 0; i < 8; i++) 
{
    uint64 b;
    memcpy(&b, &state->buffer[i << 3], sizeof(b));
    x[i] ^= b;
}

E8(state);

for (i = 0; i < 8; i++) 
{
    uint64 b;
    memcpy(&b, &state->buffer[i << 3], sizeof(b));
    x[i + 8] ^= b;
}
```

It's a lot wordier, but it complies with the C++ standard.

jeffro256 (approved):

PR #9049, opened 2023-11-02
wallet: sign_tx: get tx pubkey from extra

tobtoht:
Fixes an issue where owned output key images are not always [included](https://github.com/monero-project/monero/blob/master/src/wallet/wallet2.cpp#L7324) in the signed transaction.

Tx pub keys for single destination transactions to a subaddress are calculated [differently](https://github.com/monero-project/monero/blob/master/src/cryptonote_core/cryptonote_tx_utils.cpp#L385).


r4v3r23:
tested self-spend & external spend successfully

change output is now synced with key image on view-only wallet after importing signed tx

selsta (approved):

PR #9050, opened 2023-11-02
wallet: fix multisig key storage leak

jeffro256:
Multisig keys per-transfer were being wiped, but not erased, which lead to a ginormous quadratic bloat the more transfers and exports you performed with the wallet. I talked to a person that was able to remove 335 MEGABYTES from their wallet cache b/c of this PR.


vtnerd (approved):
Not really a leak in the traditional sense, but I'm sure this helps with memory.

jeffro256:
Yeah not technically a "memory leak" since the data structures still account for the memory but I don't know what else to call it.

jeffro256:
That's probably a more apt name, although those extra zero'd out keys do still consume memory as well.

PR #8592, opened 2022-09-24
Fix ZMQ server poll code

vtnerd:
These changes shouldn't be necessary according to the ZeroMQ documentation, but `monerod`s ZeroMQ code is having issue with `EAGAIN` in the RPC server read. I'm marking this as a draft for now so that @sethforprivacy and @Gingeropolous can test first.

Issue #8199, opened 2022-03-02
ZMQ DESTRYOED by p2pool daemon. 


Gingeropolous:
I checked on my p2pool daemon to discover it thought the monerod daemon was stuck.

a restart of monerod fixed it.

grepping the logs for ERROR returned this

 2022-02-28 15:01:43.241 [P2P3]  ERROR   net.zmq src/rpc/zmq_pub.cpp:546 Unable to send ZMQ/Pub - ZMQ server destroyed

 2022-02-28 14:59:05.955     7f74597fa700        ERROR   net.zmq src/rpc/zmq_server.cpp:174      ZMQ RPC Server Error: thrown at zmq_server.cpp:
161: Resource temporarily unavailable

please lemme know if more logs are needed before the great logrotate devours the logs forever. 

so far has only happened once, and it happened after updating to the most recent p2pool. Prior to this update, I had never experienced this before - I've been running p2pool since its launch.... months ago?

sethforprivacy:
I have also seen this happen once, and a monerod restart resolved it and has not occurred since.

Will keep an eye out for it happening again in the future!

sethforprivacy:
I have had this relatively frequently, and last seen with high CPU usage by monerod. Still seems to be an issue on v0.18.1.0.

sethforprivacy:
I get this log entry before the log spam:

`ZMQ RPC Server Error: thrown at zmq_server.cpp:161: Resource temporarily unavailable`

SChernykh:
@vtnerd 
My thoughts from IRC so they don't get lost:
```
<sech1> https://splunktool.com/zmq-resource-temporarily-unavailable-when-connecting-zmq-socket
<sech1> zmq calls can throw exceptions, so it makes sense to add try...catch in zmq_server.cpp
<sech1> no, wait, it's not zmq that throws an exception
<sech1> zmq returns error EAGAIN (resource temporarily unavailable) and then MONERO_UNWRAP macro throws
<sech1> so it should probably check for EAGAIN there and just ignore it
<sech1> EAGAIN means non-blocking sockets can't receive/send data right now, so it should just continue with the next loop iteration instead of throwing up
```

vtnerd:
The problem is that there is a poll call and I botched the array setup sequence - the values are never reset for the next poll call. So if ZeroMQ doesn't reset the values, our server code is getting an old value and being told to read a socket in non-blocking mode that potentially has no data.

vtnerd:
[...]

PR #9052, opened 2023-11-04
Fix EAGAIN bug in ZMQ-RPC/ZMQ-PUB

vtnerd:
A user reported over IRC that connecting to the ZMQ-RPC server via web browser would crash the ZMQ-RPC permanently, until restart. After investigation, I found out that this only happens when the ZMQ-PUB server is also enabled.

When ZMQ-PUB is enabled, the ZMQ server switches to `zmq_poll` instead of using a blocking read with `zmq_read`. Invalid messages are not handled by the `zmq_poll`, and instead are left to `zmq_read` to filter out. So a "spurious" wakeup is done for `zmq_poll` only to be given an `EAGAIN` error by `zmq_read`.

This should finally clear out the longstanding issue from #8199 - now that I know whats happening, I think this is the only fix. #8592 is being closed too.

### DIFF
--- a/src/rpc/zmq_server.cpp
+++ b/src/rpc/zmq_server.cpp
@@ -158,13 +158,22 @@ void ZmqServer::serve()
 
       if (!pub || sockets[2].revents)
       {
-        std::string message = MONERO_UNWRAP(net::zmq::receive(rep.get(), read_flags));
-        MDEBUG("Received RPC request: \"" << message << "\"");
-        epee::byte_slice response = handler.handle(std::move(message));
+        expect<std::string> message = net::zmq::receive(rep.get(), read_flags);
+        if (!message)
+        {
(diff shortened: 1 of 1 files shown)


vtnerd:
@Gingeropolous @SChernykh @trasherdk @sethforprivacy pinging people that were tracking the old issue.

SChernykh (approved):

hinto-janai:
From https://github.com/monero-project/monero/issues/8199:
> Hopefully the next loop iteration the poll code doesn't return that data is available, because that would turn this into a busy spin loop

Is this unlikely enough to ignore?

vtnerd:
It doesn't appear to happen, otherwise we would file a bug with upstream ZMQ.

hinto-janai (approved):

PR #9060, opened 2023-11-06
workflows: upload macOS binary after name change

selsta:
Changed in #9007


tobtoht (approved):
Huh, could have sworn I grepped that.

hinto-janai (approved):

PR #9072, opened 2023-11-20
Change std::result_of to auto

vtnerd:
Clang now reports that `std::result_of` is deprecated in C++17. The replacement is `std::invoke_result`.

### DIFF
--- a/contrib/epee/src/http_auth.cpp
+++ b/contrib/epee/src/http_auth.cpp
@@ -217,15 +217,13 @@ namespace
   //// Digest Authentication
 
   template<typename Digest>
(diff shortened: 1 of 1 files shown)


selsta (approved):

vtnerd:
@selsta do you know the differences in the arm8 android build?

vtnerd:
@selsta I decided to stop investigating why the Android build didn't work, and just switched to `auto` for the return type. We are already using it in this file with success.

The remaining `std::result_of` in `src/lmdb/database.h` is left for later (not a candidate for a simple `auto`).

jeffro256 (approved):

PR #9077, opened 2023-11-28
serialization: remove container wrappers and serialize directly

jeffro256:
Some downstream code (most notably PR https://github.com/UkoeHB/monero/pull/25) wants to use the src/serialization lib
for storing information persistently. When one builds classes/machines wishing to serialize containers, they must use
the `serializable_*` container classes. In this case, this makes the Seraphis library code unnecessarily tightly coupled
with the src/serialization code since one cannot swap out their type of storage format without major refactoring of class
field types. By serializing STL containers directly, we can abstract the serialization details away, making for much
cleaner design. Also small bonus side effect of this change is that STL containers with custom Comparators, Allocators,
and Hashers are serializable.

Depends upon https://github.com/monero-project/monero/pull/9069.

### DIFF
--- a/src/serialization/container.h
+++ b/src/serialization/container.h
@@ -57,8 +57,28 @@ namespace serialization
       return true;
     }
 
-    template <typename C>
-    void do_reserve(C &c, size_t N) {}
+    //! @brief Add an element to a container, inserting at the back if applicable.
+    template <class Container>
+    auto do_add(Container &c, typename Container::value_type &&e) -> decltype(c.emplace_back(e))
+    { return c.emplace_back(e); }
+    template <class Container>
+    auto do_add(Container &c, typename Container::value_type &&e) -> decltype(c.emplace(e))
+    { return c.emplace(e); }
+
+    //! @brief Reserve space for N elements if applicable for container.
+    template<typename... C>
+    void do_reserve(const C&...) {}
+    template<typename C>
+    auto do_reserve(C &c, std::size_t N) -> decltype(c.reserve(N)) { return c.reserve(N); }
+
+    // The value_type of STL map-like containers come in the form std::pair<const K, V>.
+    // Since we can't {de}serialize const types in this lib, we must convert this to std::pair<K, V>
+    template <class Container, typename = void>
+    struct serializable_value_type
+    { using type = typename Container::value_type; };
+    template <class Container>
+    struct serializable_value_type<Container, std::conditional_t<false, typename Container::mapped_type, void>>
+    { using type = std::pair<typename Container::key_type, typename Container::mapped_type>; };
   }
 }
 
@@ -82,7 +102,7 @@ bool do_serialize_container(Archive<false> &ar, C &v)
   for (size_t i = 0; i < cnt; i++) {
     if (i > 0)
       ar.delimit_array();
-    typename C::value_type e;
+    typename ::serialization::detail::serializable_value_type<C>::type e;
     if (!::serialization::detail::serialize_container_element(ar, e))
       return false;
     ::serialization::detail::do_add(v, std::move(e));
@@ -104,7 +124,8 @@ bool do_serialize_container(Archive<true> &ar, C &v)
       return false;
     if (i != v.begin())
       ar.delimit_array();
-    if(!::serialization::detail::serialize_container_element(ar, (typename C::value_type&)*i))
+    using serializable_value_type = typename ::serialization::detail::serializable_value_type<C>::type;
+    if(!::serialization::detail::serialize_container_element(ar, (serializable_value_type&)*i))
       return false;
     if (!ar.good())
       return false;
--- a/src/serialization/container.h
+++ b/src/serialization/container.h
@@ -42,7 +42,7 @@ namespace serialization
     typename std::enable_if<!use_container_varint<T>(), bool>::type
     serialize_container_element(Archive& ar, T& e)
     {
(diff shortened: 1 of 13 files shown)


src/serialization/containers.h
```diff
@@ -38,91 +38,22 @@
...
-    template <typename T> void do_add(std::unordered_set<T> &c, T &&e) { c.insert(std::forward<T>(e)); }
-
-    template <typename T> void do_add(std::set<T> &c, T &&e) { c.insert(std::forward<T>(e)); }
-  }
+    //! @brief Is this type a STL-like container?
+    //! To add a new container to be serialized, partially specialize the template is_container like so:
+    template <typename T>     struct is_container:                                 public std::false_type {};
+    template <typename... TA> struct is_container<std::deque<TA...>>:              public std::true_type {};
+    template <typename... TA> struct is_container<std::map<TA...>>:                public std::true_type {};
+    template <typename... TA> struct is_container<std::set<TA...>>:                public std::true_type {};
```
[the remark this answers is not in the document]
jeffro256: Added. No new header was needed so it won't affect compile time either

src/serialization/containers.h
```diff
@@ -38,91 +38,23 @@
...
-
-    template <typename K, typename V> void do_add(serializable_unordered_multimap<K, V> &c, std::pair<K, V> &&e) { c.insert(std::forward<std::pair<K, V>>(e)); }
-
-    template <typename T> void do_add(std::unordered_set<T> &c, T &&e) { c.insert(std::forward<T>(e)); }
-
-    template <typename T> void do_add(std::set<T> &c, T &&e) { c.insert(std::forward<T>(e)); }
-  }
+    //! @brief Is this type a STL-like container?
+    //! To add a new container to be serialized, partially specialize the template is_container like so:
+    template <typename T>     struct is_container:                                 public std::false_type {};
```
vtnerd, line 45: These `public`s aren't needed with `struct. I point this out because this out only because this is word wrapping on my display now.

src/serialization/container.h
```diff
@@ -52,13 +52,32 @@ namespace serialization
...
       if (!previously_varint && ar.varint_bug_backward_compatibility_enabled() && !typename Archive::is_saving())
-        return ::do_serialize(ar, e);
+        return do_serialize(ar, e);
       ar.serialize_varint(e);
       return true;
     }
 
+    //! @brief Add an element to a container, inserting at the back if applicable.
+    template <class Container>
+    auto do_add(Container &c, typename Container::value_type &&e) -> decltype(c.emplace_back(e), void())
```
vtnerd, line 62: I commented on this on another review, and you rebuked, but I'll try again!

Why not just make these pass-through? `emplace_back()` returns void anyway, and its not terribly to forward the return-call (if there is one).

src/serialization/container.h
```diff
@@ -52,13 +52,32 @@ namespace serialization
...
+    { c.emplace_back(e); }
+    template <class Container>
+    auto do_add(Container &c, typename Container::value_type &&e) -> decltype(c.emplace(e), void())
+    { c.emplace(e); }
+
+    //! @brief Reserve space for N elements if applicable for container.
     template <typename C>
     void do_reserve(C &c, size_t N) {}
+    template <typename T> void do_reserve(std::vector<T> &c, size_t N) { c.reserve(N); }
```
vtnerd, line 71: You can make the first function a template pack and then auto-return type reserve:
```c++
template<typename... C>
void do_reserve(const C&...) {}

template<typename C>
auto do_reserve(C& c, std::size_t N) -> decltype(c.reserve(N)) { return c.reserve(N); }
```
which gets all functions with a reserve.

src/serialization/container.h
```diff
@@ -104,7 +123,8 @@ bool do_serialize_container(Archive<true> &ar, C &v)
       return false;
     if (i != v.begin())
       ar.delimit_array();
-    if(!::serialization::detail::serialize_container_element(ar, (typename C::value_type&)*i))
+    using serializable_value_type = typename ::serialization::detail::serializable_value_type<C>::type;
```
vtnerd, line 126: I don't see why this needs to be cast, could you just pass `*i` to `serialize_container_element`? I suppose there are situations where `value_type`/`mapped_type` are not the same as `decltype(*i)`?

src/serialization/container.h
jeffro256: I guess I don't care as much about the return type of these functions since they don't need to be chained together like `do_serialize()`. It would make it less verbose anyways, so I think this way is better. FWIW, `emplace_back()` only returns `void` before C++17 and `emplace()` returns `std::pair<iterator, bool>`.

src/serialization/container.h
jeffro256: I like this! I was trying to find a not-horribly-verbose way to say this without mentioning any specific container

src/serialization/container.h
jeffro256: We can't serialize `const` types without a lot of refactoring. The value type of mapped containers is `std::pair<const K, V>`, so when we try to serialize that value type, the pair serialization code tries to serialize `const K`, and the compiler can't find any functions that match `do_serialize(Archive&, const K&)` (since they don't exist) so it throws an error.

vtnerd: That's too bad.

vtnerd (approved):

PR #9090, opened 2023-12-19
Cleanup scoped_ptr in src/common/dns_utils.cpp

### LINKED ISSUE #8732: Idea: use std::unique_ptr deleter to support custom destructor in custom smart pointer in dns_utils?
There is this line with a TODO:
https://github.com/monero-project/monero/blob/9489586add5116a48039616140bb7267b16ba7b1/src/common/dns_utils.cpp#L198

std::unique_ptr supports a custom deleter, but I am not sure if this fits all requirements.

0xFFFC0000:
* No need for entire new class (Fix #8732).
	* Fix stdint.h header include in contrib/epee/include/net/http_base.h.

Fix #8732

### DIFF
--- a/src/common/dns_utils.cpp
+++ b/src/common/dns_utils.cpp
@@ -28,6 +28,7 @@
 
 #include "common/dns_utils.h"
 // check local first (in the event of static or in-source compilation of libunbound)
+#include "misc_language.h"
 #include "unbound.h"
 
 #include <deque>
@@ -195,36 +196,7 @@ boost::optional<std::string> tlsa_to_string(const char* src, size_t len)
     return boost::none;
   return std::string(src, len);
 }
-
-// custom smart pointer.
-// TODO: see if std::auto_ptr and the like support custom destructors
-template<typename type, void (*freefunc)(type*)>
-class scoped_ptr
-{
-public:
-  scoped_ptr():
-    ptr(nullptr)
-  {
-  }
-  scoped_ptr(type *p):
-    ptr(p)
-  {
-  }
-  ~scoped_ptr()
-  {
-    freefunc(ptr);
-  }
-  operator type *() { return ptr; }
-  type **operator &() { return &ptr; }
-  type *operator->() { return ptr; }
-  operator const type*() const { return &ptr; }
-
-private:
-  type* ptr;
-};
-
-typedef class scoped_ptr<ub_result,ub_resolve_free> ub_result_ptr;
-
+  
 struct DNSResolverData
 {
   ub_ctx* m_ub_context;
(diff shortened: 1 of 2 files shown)


0xFFFC0000:
Fail from `ci/gh-actions/depends / x86_64 Freebsd (pull_request) `  :

`curl: (22) The requested URL returned error: 404 Not Found`

selsta:
@0xFFFC0000 the source code is hosted on getmonero which currently has an expired cert, should resolve itself later today

src/common/dns_utils.cpp
```diff
@@ -327,10 +299,14 @@ std::vector<std::string> DNSResolver::get_record(const std::string& url, int rec
...
-
-  // destructor takes care of cleanup
-  ub_result_ptr result;
-
+  
+  ub_result *result;
+  // Make sure we are cleaning after result.
+  epee::misc_utils::auto_scope_leave_caller scope_exit_handler =
+    epee::misc_utils::create_scope_leave_handler([&](){
+    ub_resolve_free(result);
```
vtnerd, line 307: Should a `nullptr` check be added here?

vtnerd (approved):

src/common/dns_utils.cpp
vtnerd: I guess it always worked before, so probably not.

0xFFFC0000: Good catch @vtnerd . 

To be honest, I thought about adding a simple nullptr check, but I didn't want to make any semantical changes to the code.

src/common/dns_utils.cpp
```diff
@@ -327,10 +299,14 @@ std::vector<std::string> DNSResolver::get_record(const std::string& url, int rec
...
   dnssec_available = false;
   dnssec_valid = false;
-
-  // destructor takes care of cleanup
-  ub_result_ptr result;
-
+  
+  ub_result *result;
+  // Make sure we are cleaning after result.
+  epee::misc_utils::auto_scope_leave_caller scope_exit_handler =
```
jeffro256, line 305: Why not combine these into a `std::unique_ptr` with a custom deleter?

0xFFFC0000: No easy way of handling `**` argument via `unique_ptr` [1]. The `result` is used as parameter [2].


1. https://stackoverflow.com/a/24543176
2. https://github.com/monero-project/monero/pull/9090/files/430b102aef0151820a27eafe232afa40ba8e3d8f#diff-ed01ce7362fc3c2d5c7adc0a0be9742bbc6c7558825914f6a2d288b1e0f552fcR313

src/common/dns_utils.cpp
0xFFFC0000: I am resolving this conversation. Please feel free to reopen this comment.

jeffro256 (approved):

PR #9093, opened 2023-12-20
[suggestion] wallet: use ~/.bitmonero/monero-wallet-cli.conf as default config for cli

alfredfo:
As said in the title this is more of a suggestion. I haven't tested it on Windows, guessing that '~' could be an issue, and maybe there should be more default paths like: /etc/monero/monero-wallet-cli.conf.

Should probably also change location if `--stagenet` or `--testnet` is passed, but let me know if you like the idea.

### DIFF
--- a/src/wallet/wallet_args.cpp
+++ b/src/wallet/wallet_args.cpp
@@ -108,7 +108,7 @@ namespace wallet_args
     const command_line::arg_descriptor<std::size_t> arg_max_log_files = {"max-log-files", "Specify maximum number of rotated log files to be saved (no limit by setting to 0)", MAX_LOG_FILES};
     const command_line::arg_descriptor<uint32_t> arg_max_concurrency = {"max-concurrency", wallet_args::tr("Max number of threads to use for a parallel job"), DEFAULT_MAX_CONCURRENCY};
     const command_line::arg_descriptor<std::string> arg_log_file = {"log-file", wallet_args::tr("Specify log file"), ""};
-    const command_line::arg_descriptor<std::string> arg_config_file = {"config-file", wallet_args::tr("Config file"), "", true};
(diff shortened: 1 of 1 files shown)


0xFFFC0000:
Thanks for the PR. I like this change. I am not approving it yet, I want to wait and see if there are any objections from the community. 

If not will approve it.

omurad:
I like this idea, though on Linux systems a feature like this should respect [XDG](https://specifications.freedesktop.org/basedir-spec/latest/), specifically in this case `XDG_CONFIG_HOME`.

src/wallet/wallet_args.cpp
```diff
@@ -108,7 +108,7 @@ namespace wallet_args
     const command_line::arg_descriptor<std::size_t> arg_max_log_files = {"max-log-files", "Specify maximum number of rotated log files to be saved (no limit by setting to 0)", MAX_LOG_FILES};
     const command_line::arg_descriptor<uint32_t> arg_max_concurrency = {"max-concurrency", wallet_args::tr("Max number of threads to use for a parallel job"), DEFAULT_MAX_CONCURRENCY};
     const command_line::arg_descriptor<std::string> arg_log_file = {"log-file", wallet_args::tr("Specify log file"), ""};
-    const command_line::arg_descriptor<std::string> arg_config_file = {"config-file", wallet_args::tr("Config file"), "", true};
+    const command_line::arg_descriptor<std::string> arg_config_file = {"config-file", wallet_args::tr("Config file"), "~/.bitmonero/monero-wallet-cli.conf", true};
```
jpk68, line 111: This should, in theory, fix compatibility with Windows.

```suggestion
    const command_line::arg_descriptor<std::string> arg_config_file = {"config-file", wallet_args::tr("Config file"), (boost::filesystem::path(tools::get_default_data_dir()) / "monero-wallet-cli.conf").string(), true};
```

jpk68 (changes requested):

PR #9117, opened 2024-01-06
remove ARCH="x86-64" from Makefile

woodser:


selsta (approved):

selsta:
Hmm, some of these targets seem to be for a sepcific arch so it might make sense to keep the x86-64 in it, on the other hand there is no point to keep it.

woodser:
I can change only `release-static` if preferred, since that's what I tested. I also don't know about the others.

selsta:
> I can change only release-static

Yes, would prefer.

selsta (approved):

PR #9126, opened 2024-01-19
CMake: only set policy CMP0148 to OLD if version >= 3.27

jeffro256:
Allows people to compile if their CMake is older than 3.27

@tobtoht

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,8 +40,9 @@ include(CheckCXXCompilerFlag)
 include(CheckLinkerFlag)
 include(CheckLibraryExists)
 include(CheckFunctionExists)
-
-cmake_policy(SET CMP0148 OLD)
+if (POLICY CMP0148)
+    cmake_policy(SET CMP0148 OLD) # https://cmake.org/cmake/help/latest/policy/CMP0148.html
+endif()
 include(FindPythonInterp)
 
 if (IOS)


tobtoht:
How did this pass CI..

Anyway, `cmake_minimum_required` should stay at 3.5, see: https://cmake.org/cmake/help/latest/command/cmake_policy.html#setting-policies-by-cmake-version

0xFFFC0000:
> How did my patch pass CI..

it is because when you set the minimum required version via `cmake_minimum_required`, it evaluates `3.5` as `3.05`. It is a bug, or one of the shady behaviours of CMake, depending on who you ask. 

For example, if you have: 

``` 
$ cmake --version
cmake version 3.16.3

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

This will fail: 
```
cmake_minimum_required(VERSION 3.50)
```

But this will work: 

```
cmake_minimum_required(VERSION 3.5)
```

CMakeLists.txt
```diff
@@ -40,8 +40,9 @@ include(CheckCXXCompilerFlag)
 include(CheckLinkerFlag)
 include(CheckLibraryExists)
 include(CheckFunctionExists)
-
-cmake_policy(SET CMP0148 OLD)
+if (NOT("${CMAKE_VERSION}" VERSION_LESS 3.27)) # If CMake version >= 3.27
```
0xFFFC0000, line 43:
```
if(POLICY CMP0148)
  cmake_policy(SET CMP0148 OLD) # https://cmake.org/cmake/help/latest/policy/CMP0148.html
endif()
```

It is a cleaner syntax IMHO [1]. 

1. https://cmake.org/cmake/help/latest/command/if.html#policy

0xFFFC0000:
> How did my patch pass CI..

The build shows that CMake is 3.27.8.

https://github.com/monero-project/monero/actions/runs/6991078601/job/19021248770#step:9:25

This is a question for me too, where does our CI get 3.27.8.

selsta:
@0xFFFC0000 CI VM has it pre-installed, see #9128

0xFFFC0000:
> @0xFFFC0000 CI VM has it pre-installed, see #9128

Thanks. Now it makes sense.

selsta (approved):

tobtoht (approved):

0xFFFC0000 (approved):

Issue #9127, opened 2024-01-19
Bug: it's possible to set --no-zmq and zmq pub / bind ip and ports in Monerod config without having any errors


snrocacorns:
It's possible to start up Monerod with these opposing flags set- it should not start and should give a warning.

PR #9137, opened 2024-01-27
daemon: throw error on specifiying ZMQ args with --no-zmq

jeffro256:
Resolves https://github.com/monero-project/monero/issues/9127


selsta (approved):

src/daemon/daemon.cpp
```diff
@@ -124,6 +124,19 @@ struct t_internals {
         core.get().set_txpool_listener(cryptonote::listener::zmq_pub::txpool_add{shared});
       }
     }
+    else // if --no-zmq specified
+    {
+      // Assert that none of --zmq-rpc-bind-port, --zmq-rpc-bind-ip, and --zmq-pub are specified b/c
+      // that does not make semantic sense with --no-zmq.
+      if (command_line::get_arg(vm, daemon_args::arg_zmq_rpc_bind_port) !=
```
vtnerd, line 131: There's a function `is_arg_defaulted` for this.

0xFFFC0000 (approved):

iamamyth:
I'm not sure I agree with the notion that this is a bug in the first place: `--no-X` turns off feature `X`; `--X-opt` sets opt for feature `X`; there's no incompatibility between the two. A trivial example which worked before this change, and doesn't after: If a user has a lengthy daemon startup command and wants to temporary shut off zmq (conceptually, applies to any optional feature with its own config flags), adding --no-zmq at the end of the command works, and makes reverting trivial (remove --no-zmq from the end of line). After this change, the workflow becomes: Drop all zmq flags, add --no-zmq; re-add all flags when done.

iamamyth:
To clarify one point: This could breaks a prior workflow (outlined above). There's a burden of proof for any change, and it's much higher when it's a breaking change. I don't think the original issue description meets the burden for a non-breaking change, and I certainly don't think it does so for a breaking change: Why complicate the command-line validation logic? Who does this help, and how? I understand one could advance justifications for this change, but the original issue didn't, and it certainly didn't address the tradeoff.

jeffro256:
> Why complicate the command-line validation logic? Who does this help, and how?

Not the original issue poster, but the main downside to the old behavior is confusion with specifying ZMQ opts with --no-zmq and not understanding why the behavior isn't changing despite you changing the options. Gets slightly more confusing once you start combining CLI with the config file.

> A trivial example which worked before this change, and doesn't after: If a user has a lengthy daemon startup command and wants to temporary shut off zmq (conceptually, applies to any optional feature with its own config flags), adding --no-zmq at the end of the command works, and makes reverting trivial (remove --no-zmq from the end of line).

Is there any evidence of anyone using this workflow?

> I'm not sure I agree with the notion that this is a bug in the first place: --no-X turns off feature X; --X-opt sets opt for feature X; there's no incompatibility between the two.

This isn't *exactly* the situation since the options aren't set in the code unless `--no-zmq` isn't present. In other words, the effect of `--zmq-*` args is conditional on the presence of `--no-zmq`. I'd think that this would be a breaking change if you could switch on ZMQ later during runtime, but I don't think that that is an option (please correct if wrong).

All of this aside, I'd be okay compromising and only logging a warning message.

iamamyth:
> All of this aside, I'd be okay compromising and only logging a warning message.

I agree that this would be the appropriate course of action.

> Is there any evidence of anyone using this workflow?
This question perfectly illustrates my point: The burden should be to justify the change, not the existing behavior. Why should users have to advocate for the software's current behavior? Is there evidence for this supposed problem of users getting confused by the combination of `--no-zmq` and `--zmq-<opt>`? Not even the original bug report mentioned any real problem experienced by a user, it was purely normative (the software "should" refuse startup).

But, since you asked, I use this workflow with a lot of software. It's especially helpful when the software accepts multi-source configuration, as one need not fiddle with the config file to temporarily disable a feature.

> I'm not sure I agree with the notion that this is a bug in the first place: --no-X turns off feature X; --X-opt sets opt for feature X; there's no incompatibility between the two.

This was a claim as to the *logical* effect (in other words, there's no logical contradiction between the flags); and it's precisely because one cannot dynamically enable ZMQ that it holds, even though the daemon ignores all `--zmq-<opt>` flags with `--no-zmq` set.

jeffro256:
Exceptions were changed to warnings

selsta (approved):
Please squash

selsta (approved):

PR #9138, opened 2024-01-28
Disable/fix ports with I2P

vtnerd:
This should fix #8918 , but I am soliciting for tests before marking this as ready. Basically, I only know this passes the unit tests at this point.

The patch is pretty simple, I2P ports are now fixed at `0`, so opening the new pearliest manager should collapse the overloads on port once it goes into the internal map.

PR #9142, opened 2024-01-30
Disable/fix ports with I2P [0.18]

vtnerd:
Same as #9138 but on the release branch.


vtnerd:
Force pushed a change to improve backwards compatibility.

selsta (approved):

PR #9146, opened 2024-01-30
Adding issue template to bug reports in Monero repository.

0xFFFC0000:
Includes two templates: Bug Report, and Feature Request.

In case you want to look at how it appears in a GitHub report page, hit the `New issue` button here [1]. In addition to the template, this will assign `bug` label to bug reports and `feature` label to feature requests.

1. https://github.com/0xFFFC0000/monero-perf/issues

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,94 @@
+name: Bug report
+description: Submit a new bug report.
+labels: [bug]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ## This issue tracker is only for technical issues related to Monero.
+
+        * General monero questions and/or support requests should use Monero StackExchange at https://monero.stackexchange.com.
+        * For reporting security issues, please read instructions at https://github.com/monero-project/meta/blob/master/VULNERABILITY_RESPONSE_PROCESS.md.
+        * For reporting GUI issues, please use Monero-GUI repository [issue tracker](https://github.com/monero-project/monero-gui/issues/new).
+
+        ----
+  - type: checkboxes
+    attributes:
+      label: Is there an existing issue for this?
+      description: Please search to see if an issue already exists for the bug you encountered. You can search [here](https://github.com/monero-project/monero/issues?q=is%3Aissue).
+      options:
+      - label: I have searched the existing issues
+        required: true
+  - type: textarea
+    id: current-behaviour
+    attributes:
+      label: Current behaviour
+      description: Tell us what went wrong
+    validations:
+      required: true
+  - type: textarea
+    id: expected-behaviour
+    attributes:
+      label: Expected behaviour
+      description: Tell us what you expected to happen
+    validations:
+      required: false
+  - type: textarea
+    id: reproduction-steps
+    attributes:
+      label: Steps to reproduce
+      description: |
+        Tell us how to reproduce your bug. Please attach related screenshots if necessary.
+        * Run-time or compile-time configuration options
+        * Actions taken
+    validations:
+      required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
+      description: |
+        Please copy and paste any relevant log output or attach a debug log file.
+
+        * For daemon log you can get log information via `--log-level=2` flag, or by taking a look at log files at `~/.bitmonero/` in cli version.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,93 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
+      description: |
+        Please copy and paste any relevant log output or attach a debug log file.
+
+        You can get log information via --log-level=4 flag or by taking a look at log files at ~/.bitmonero/ in cli version. Or you can look at the log information in Settings->Log section in Monero-GUI.
```
selsta, line 53: I don't think we should require log-level 4 here, it's extremely noisy, 2 is better for most issues.

0xFFFC0000: Consider it done. Will change this to level 2.

0xFFFC0000: Fixed it in a new push. We have log-level=2 now.

Gingeropolous:
I would include a link to search for existing issues when you ask the user if they've checked. 

honestly, I've never liked these templates. I understand their efficiency, but there have been many cases where I'm using open source software, and I come across an issue, so I head to the github to file an issue, and I'm hit with a way-too-involved issue template so I just don't file it. ("I'm an end user, not a debugger!")

and some of these fields will get repetitive for _someone_ who regularly uses monero and finds issues.  :)

but this seems less onerous than some templates I've found out there. At least only some of the fields are required.

0xFFFC0000:
@Gingeropolous I agree.  It is understandable for users to decide to not report issues for projects that do have many extra ***mandatory*** templates. But the key here is the **mandatory** word. Without extra, useless mandatory fields, it is quite useful. Without forcing us to put in extra time, these templates do provide much better support. For example, we have users reporting a compilation bug, but they don't even mention what is the os they are using. 


Will add a link to search existing issues.

selsta:
Personally I would prefer the following fields, it's a good tradeoff between having necessary information and complexity of creating the issue

- describe the issue
- logs
- which version
- where did you get the binary
- which OS, hardware

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,93 @@
...
+labels: [bug]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ## This issue tracker is only for technical issues related to Monero.
+
+        * General monero questions and/or support requests should use Monero StackExchange at https://monero.stackexchange.com.
+        * For reporting security issues, please read instructions at https://github.com/monero-project/meta/blob/master/VULNERABILITY_RESPONSE_PROCESS.md.
+
```
selsta, line 12: Please also add a point that GUI related issues should be reported on the GUI repo.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,93 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
+      description: |
+        Please copy and paste any relevant log output or attach a debug log file.
+
+        You can get log information via --log-level=2 flag or by taking a look at log files at ~/.bitmonero/ in cli version. Or you can look at the log information in Settings->Log section in Monero-GUI.
```
selsta, line 53: Should we leave this in? For example the wallet log file is in a different place than the daemon log file, also --log-level=2 affects the log that gets printed and written to the disk, so that "or" doesn't really make sense.

.github/ISSUE_TEMPLATE/bug_report.yml
0xFFFC0000: Will be added in new PR. Thanks.

.github/ISSUE_TEMPLATE/bug_report.yml
0xFFFC0000: I added a more detailed comment here in new push. Please don't hesitate to let me know if it needs to change again. Thanks

0xFFFC0000:
Changelog for last push: 

1. Changed `Expected behaviour` to a non-mandatory field.  
2. Changed `Machine specifications` to mandatory field.
3. Added point about the Monero-GUI repository issue tracker.
4. Improved the comment about how can user generate comment.



Example: https://github.com/0xFFFC0000/monero-perf/issues/new?assignees=&labels=bug&projects=&template=bug_report.yml

PR #9148, opened 2024-02-01
[DRAFT] 10x improvement to `pop_blocks` operation with flag denoting txis are already verified

0xFFFC0000:
Add `already_verified` flag to add transaction operation of the `txpool`. This is very useful when we are popping block, and `txis` are already verified in that situation, which prevents verifying transactions again.

Numbers on my machine:

popping 500 blocks: `master` branch ~50 seconds, with this PR ~6 seconds.
popping 2000 blocks: `master` branch ~210 seconds, with this PR ~29 seconds.

Profiling information of `monerod` on `master` branch. As you can see, 80-90% of computation is happening in `check_tx_inputs` function. The profiling output is attached to this message. 

[monerod-master-branch-pop_blocks-profile.pdf](https://github.com/monero-project/monero/files/14127104/monerod-master-branch-pop_blocks-profile.pdf)

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -647,7 +647,7 @@ block Blockchain::pop_block_from_blockchain()
       // that might not be always true. Unlikely though, and always relaying
       // these again might cause a spike of traffic as many nodes re-relay
       // all the transactions in a popped block when a reorg happens.
-      bool r = m_tx_pool.add_tx(tx, tvc, relay_method::block, true, version);
+      bool r = m_tx_pool.add_tx(tx, tvc, relay_method::block, true, version, true);
       if (!r)
       {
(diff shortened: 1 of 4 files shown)


jeffro256:
FWIW, I wrote a similar change in PR #9135: the caller can provide a hard fork version for which it is known that the transaction passes non-input consensus checks. If that is the same version as the version passed into `add_tx`, then it skips the non-input consensus checks. This is a safer and clearer code path IMO. Also, skipping the input consensus checks is incorrect behavior for txs not kept by block, since a transaction can verify input checks once but then the block reorgs.

vtnerd:
I would definitely hold off on working too hard on this until @jeffro256 gets his changes in - even if there is additional time savings beyond his PR. He has a pretty big change coming soon.

PR #9149, opened 2024-02-03
epee: cleanup string_tools

0xFFFC0000:
1. Use `boost::filesystem` for already available operations.
2. Use `boost::string` for already available operations.
3. ~~Fix the inconsistent indentation.~~

### DIFF
--- a/contrib/epee/include/string_tools.h
+++ b/contrib/epee/include/string_tools.h
@@ -31,6 +31,7 @@
 #include "mlocker.h"
 
 #include <boost/utility/string_ref.hpp>
+#include <boost/algorithm/string.hpp> 
 #include <sstream>
 #include <string>
 #include <cstdint>
@@ -69,23 +70,19 @@ namespace string_tools
 #ifdef _WIN32
    std::string get_current_module_path();
 #endif
-  bool set_module_name_and_folder(const std::string& path_to_process_);
-  bool trim_left(std::string& str);
-  bool trim_right(std::string& str);
+  void set_module_name_and_folder(const std::string& path_to_process_);
+  void trim_left(std::string& str);
+  void trim_right(std::string& str);
   //----------------------------------------------------------------------------
   inline std::string& trim(std::string& str)
   {
-    trim_left(str);
-    trim_right(str);
+    boost::trim(str);
     return str;
   }
   //----------------------------------------------------------------------------
-  inline std::string trim(const std::string& str_)
+  inline std::string trim(const std::string& str)
   {
-    std::string str = str_;
-    trim_left(str);
-    trim_right(str);
-    return str;
+    return boost::trim_copy(str);
   }
   std::string pad_string(std::string s, size_t n, char c = ' ', bool prepend = false);
   
--- a/contrib/epee/src/string_tools.cpp
+++ b/contrib/epee/src/string_tools.cpp
@@ -38,9 +38,12 @@
 #include <cstdlib>
 #include <string>
 #include <type_traits>
+#include <system_error>
 #include <boost/lexical_cast.hpp>
+#include <boost/algorithm/string.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/utility/string_ref.hpp>
+#include <boost/filesystem.hpp>
 #include "misc_log_ex.h"
 #include "storages/parserse_base_utils.h"
 #include "hex.h"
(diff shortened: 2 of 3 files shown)


contrib/epee/include/string_tools.h
```diff
@@ -47,12 +48,12 @@ namespace epee
 {
 namespace string_tools
 {
-  //----------------------------------------------------------------------------
```
vtnerd, line 50: We have historically refrained from changing whitespace comments (or in this case decorator) such as this.

contrib/epee/src/string_tools.cpp
```diff
@@ -112,81 +116,70 @@ namespace string_tools
...
-	//----------------------------------------------------------------------------
-	 std::string& get_current_module_folder()
-	{	
-		static std::string module_folder;
-		return module_folder;
-	}
-	
+  
+  bool compare_no_case(const std::string& str1, const std::string& str2)
+  {
```
vtnerd, line 123: Again more whitespace changes. These lines will now attribute you, even though nothing much as been done.

Summoning @moneromooo-monero who used to push back on these.

contrib/epee/src/string_tools.cpp
```diff
@@ -112,81 +116,70 @@ namespace string_tools
...
-		if ( a != std::string::npos )
-		{	
-			get_current_module_name() = path_to_process.substr(a+1, path_to_process.size());
-			get_current_module_folder() = path_to_process.substr(0, a);
-			return true;
-		}else
-			return false;
 
-	}
+    get_current_module_name() = path_to_process.filename().string();
```
vtnerd, line 167: Now this is a nice cleanup.

contrib/epee/src/string_tools.cpp
```diff
@@ -112,81 +116,70 @@ namespace string_tools
...
   {
-    char pname [5000] = {0};
-    GetModuleFileNameA( NULL, pname, sizeof(pname));
-    pname[sizeof(pname)-1] = 0; //be happy ;)
-    return pname;
+    std::wstring buf;
+    buf.resize(MAX_PATH);
+    do
+    {
+        unsigned int len = GetModuleFileNameW(NULL, &buf[0], static_cast< unsigned int >(buf.size()));
```
vtnerd, line 146: This also returns `0` on error, I think it just loops until buff exceeds `65536` in that case. Perhaps an explicit check? We could `throw` I guess, although that changes the functionality a bit.

vtnerd (commented):
The only hold up are the whitespace changes, I roped in @moneromooo-monero hopefully who constantly pushed back against such changes.

contrib/epee/include/string_tools.h
0xFFFC0000: Returned all the non-code changes to previous state.

contrib/epee/src/string_tools.cpp
0xFFFC0000: Returned all the non-code changes to the previous state. Thanks.

contrib/epee/src/string_tools.cpp
```diff
@@ -141,52 +144,55 @@ namespace string_tools
...
-			
-		return true;
-	}
-	//----------------------------------------------------------------------------
-	bool trim_right(std::string& str)
-	{
+  void trim_left(std::string& str)
+  {
+    boost::trim_left(str);
+    return;
```
steverusso, line 186: Should this redundant `return;` statement be removed? I believe the `void` function will return here anyway.

(If so, this is also the case in the `trim_right` function a few lines below.)

moneromooo-monero (commented):
I changed my mind. It's not that large, and I see at least one improvement so it's not just "I like it better that way".

0xFFFC0000:
Changelog: 
1. Cleaned unnecessary windows function change.
2. Rebased to upstream/master. 
3. Changed the PR name to reflect our rules for PR names.

vtnerd (approved):

tobtoht:
This patch caused some breakage on Windows.

If the path to the executable contains unicode characters (e.g. `é`) the process will terminate with:

```
    terminate called after throwing an instance of 'boost::system::system_error'
      what():  boost::filesystem::path codecvt to wstring: error [codecvt:2]
```

On this line: https://github.com/monero-project/monero/blame/3b01c490953fe92f3c6628fa31d280a4f0490d28/contrib/epee/src/string_tools.cpp#L159

aleflm:
> This patch caused some breakage on Windows.
> 
> If the path to the executable contains unicode characters (e.g. `é`) the process will terminate with:
> 
> ```
>     terminate called after throwing an instance of 'boost::system::system_error'
>       what():  boost::filesystem::path codecvt to wstring: error [codecvt:2]
> ```
> 
> On this line: https://github.com/monero-project/monero/blame/3b01c490953fe92f3c6628fa31d280a4f0490d28/contrib/epee/src/string_tools.cpp#L159

path_to_process = get_current_module_path(); <- This is the likely culprit. You are assigning a std::string (returned by get_current_module_path(), likely ANSI-encoded on Windows if containing non-ASCII chars like 'é') directly to a boost::filesystem::path. Boost.Filesystem's assignment operator/ constructor taking a std::string assumes the string is UTF-8 encoded by default. If the path contains 'é' (byte OxE9 in Windows-1252), Boost tries to interpret OxE9 as UTF-8, finds it invalid, and throws the codecvt exception during this assignment or when trying to use the path later. 

Give me a day or two. I will fix this.

0xFFFC0000:
> This patch caused some breakage on Windows.
> 
> If the path to the executable contains unicode characters (e.g. `é`) the process will terminate with:
> 
> ```
>     terminate called after throwing an instance of 'boost::system::system_error'
>       what():  boost::filesystem::path codecvt to wstring: error [codecvt:2]
> ```
> 
> On this line: https://github.com/monero-project/monero/blame/3b01c490953fe92f3c6628fa31d280a4f0490d28/contrib/epee/src/string_tools.cpp#L159

path_to_process = get_current_module_path(); <- This is the likely culprit. You are assigning a std::string (returned by get_current_module_path(), likely ANSI-encoded on Windows if containing non-ASCII chars like 'é') directly to a boost::filesystem::path. Boost.Filesystem's assignment operator/ constructor taking a std::string assumes the string is UTF-8 encoded by default. If the path contains 'é' (byte OxE9 in Windows-1252), Boost tries to interpret OxE9 as UTF-8, finds it invalid, and throws the codecvt exception during this assignment or when trying to use the path later. 

Give me a day or two. I will fix this.

0xFFFC0000:
> This patch caused some breakage on Windows.
> 
> If the path to the executable contains unicode characters (e.g. `é`) the process will terminate with:
> 
> ```
>     terminate called after throwing an instance of 'boost::system::system_error'
>       what():  boost::filesystem::path codecvt to wstring: error [codecvt:2]
> ```
> 
> On this line: https://github.com/monero-project/monero/blame/3b01c490953fe92f3c6628fa31d280a4f0490d28/contrib/epee/src/string_tools.cpp#L159

I am unable to reproduce it. I built guix binaries (`master`), and didn't have any issue running `monerod.exe` binary from it from the folder named `mo é É` or `mon Éro` or `É`.  All of them worked without any issue.

PR #9160, opened 2024-02-09
Daemon-specific proxy for the wallet-rpc.

0xFFFC0000:
1. Daemon-specific proxy is exclusive with global proxy (--proxy).
2. If you set global proxy (--proxy) you cannot set daemon-specific proxy.
3. If you don't set a global proxy, you can set (or not set) a proxy for each daemon connection with the proxy field in jsonrpc to the wallet-rpc.


jeffro256:
Why would you want a daemon-specific proxy?

selsta:
@jeffro256 ofrnxmr asked for it on IRC

For example when you initially connect to a clearnet node and then want to use set_daemon to connect to a tor node, you need a proxy for this. We also support it on the daemon side with `set_bootstrap_daemon` so it made sense to add it here too.

jeffro256:
Ah. I misinterpreted the OP. I thought that wallet2 would be running two proxies lol

src/wallet/wallet2.cpp
```diff
@@ -1267,6 +1267,11 @@ bool wallet2::has_stagenet_option(const boost::program_options::variables_map& v
   return command_line::get_arg(vm, options().stagenet);
 }
 
+bool wallet2::has_proxy_option(const boost::program_options::variables_map& vm)
+{
+  return command_line::has_arg(vm, options().proxy);
+}
```
jeffro256, line 1273:
```suggestion
bool wallet2::has_proxy_option() const
{
  return !m_proxy.empty();
}
```

I'd recommend this approach since it reflects the value per `wallet2` instance, even if it wasn't initiated by a CLI.

0xFFFC0000: What you are suggesting makes more sense. I fixed it in a new push. Thanks.

src/wallet/wallet2.cpp
```diff
@@ -1351,12 +1356,15 @@ std::unique_ptr<wallet2> wallet2::make_dummy(const boost::program_options::varia
...
 //----------------------------------------------------------------------------------------------------
-bool wallet2::set_daemon(std::string daemon_address, boost::optional<epee::net_utils::http::login> daemon_login, bool trusted_daemon, epee::net_utils::ssl_options_t ssl_options)
+bool wallet2::set_daemon(std::string daemon_address, boost::optional<epee::net_utils::http::login> daemon_login, bool trusted_daemon, epee::net_utils::ssl_options_t ssl_options, const std::string& proxy)
 {
   boost::lock_guard<boost::recursive_mutex> lock(m_daemon_rpc_mutex);
 
   if(m_http_client->is_connected())
     m_http_client->disconnect();
+  CHECK_AND_ASSERT_MES2(m_proxy.empty() || proxy.empty() , "It is not possible to set global proxy (--proxy) and daemon specific proxy together.");
+  if(m_proxy.empty())
```
jeffro256, line 1366: Shouldn't this be `if (!proxy.empty())`?

jeffro256: I guess it depends if we want to behavior to be to reset the proxy when proxy is an empty string or not. I actually change my mind and like this behavior better. Let's keep it so that it resets every time.

0xFFFC0000: By the way. Keep in mind the global proxy is not changeable. That is kind of security requirement in a sense when user global proxy via --proxy flag. We should not be able to touch it, even if user asks. 


But the daemon specific proxy is daemon specific. Meaning you set proxy via RPC, to only that daemon. Later if you change daemon you have to set proxy to that daemon too.

0xFFFC0000:
PR is already merged. Changing the title to reflect the PR name rules we have [1].

1. https://github.com/monero-project/monero/pull/9210/files

PR #9161, opened 2024-02-12
[DRAFT] Caching layer to difficulty calculation of each block.

0xFFFC0000:
The default cache capacity is set by a constant in the cryptonote config file.

In certain cases, it is `20%` improvements.


jeffro256:
This cache needs to be invalidated on block pop

PR #9172, opened 2024-02-14
variant: visit lambda and value-initialize by index

jeffro256:
Two new features for `tools::variant`:
* Visit by lambda instead of static visitor (possible now because result type is deduced)
* Value-initialize the variant based on a type index

### DIFF
--- a/src/common/variant.h
+++ b/src/common/variant.h
@@ -37,6 +37,7 @@
 #include <boost/mpl/begin_end.hpp>
 #include <boost/mpl/distance.hpp>
 #include <boost/mpl/find.hpp>
+#include <boost/mpl/for_each.hpp>
 #include <boost/mpl/vector.hpp>
 #include <boost/none_t.hpp>
 #include <boost/variant/apply_visitor.hpp>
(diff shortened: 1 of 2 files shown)


src/common/variant.h
```diff
@@ -148,16 +161,26 @@ class variant final
...
     }
     template <typename VisitorT>
-    typename VisitorT::result_type visit(VisitorT &&visitor) const
+    decltype(auto) visit(VisitorT &&visitor) const // decltype(auto) since it forwards the return ref type correctly
     {
         return boost::apply_visitor(std::forward<VisitorT>(visitor), m_value);
     }
 
+    /// value initialize the variant based on a type index
+    void value_initialize_to_type_index(const int which)
```
vtnerd, line 175: In what situations would you need to construct based on index rather than type?

jeffro256: deserialization

jeffro256: Something similar is already done in the current variant de serialization code:  https://github.com/monero-project/monero/blob/059028a30a8ae9752338a7897329fe8012a310d5/src/serialization/variant.h#L74-L80

krypt0x (approved):

tobtoht:
The tests introduced here fail on master. Incompatible with #9176?

https://github.com/tobtoht/monero/actions/runs/12470659635/job/34807029624#step:9:4466

UkoeHB:
@jeffro256 can you look into it?

PR #9174, opened 2024-02-16
blockchain_db: allow implementor to be tx format agnostic

jeffro256:
Removes `cryptonote::transaction` specific things from BlockchainLMDB. This refactor should be useful during the Seraphis migration, as the transaction format/types may change. All the low-level LMDB code theoretically doesn't need to know details about the actual transaction type/format, since it's simply storing blobs, hashes, and other provided properties in fixed-size structs. If not used during the Seraphis migration, this change will otherwise just make the separation between the public interface and implementing interface cleaner. This commit shouldn't change any observable behavior about the database.

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -883,7 +883,9 @@ void BlockchainLMDB::remove_block()
       throw1(DB_ERROR(lmdb_error("Failed to add removal of block info to db transaction: ", result).c_str()));
 }
 
-uint64_t BlockchainLMDB::add_transaction_data(const crypto::hash& blk_hash, const std::pair<transaction, blobdata_ref>& txp, const crypto::hash& tx_hash, const crypto::hash& tx_prunable_hash)
+uint64_t BlockchainLMDB::add_transaction_data(const crypto::hash& blk_hash, const blobdata_ref& tx_blob,
+    const uint64_t tx_unlock_time, const size_t unprunable_size, const crypto::hash& tx_hash,
+    const crypto::hash& tx_prunable_hash)
 {
   LOG_PRINT_L3("BlockchainLMDB::" << __func__);
   check_open();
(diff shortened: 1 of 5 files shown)


src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -1018,16 +1006,14 @@ void BlockchainLMDB::remove_transaction_data(const crypto::hash& tx_hash, const
...
-
-  remove_tx_outputs(tip->data.tx_id, tx);
+  // Remove prunable hash
+  result = mdb_cursor_get(m_cur_txs_prunable_hash, &val_tx_id, NULL, MDB_SET);
+  if (result == MDB_NOTFOUND)
+    LOG_PRINT_L1("tx has no prunable hash to remove: " << tx_hash);
+  else if (result)
+    throw1(DB_ERROR(lmdb_error("Failed to locate prunable hash tx for removal: ", result).c_str()));
+  else if ((result = mdb_cursor_del(m_cur_txs_prunable_hash, 0)))
+      throw1(DB_ERROR(lmdb_error("Failed to add removal of prunable hash tx to db transaction: ", result).c_str()));
```
0xFFFC0000, line 1016: apologies for nitpicking. fix the indentation here to have a consistent indentation with other `throw1`s.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -3683,25 +3632,22 @@ bool BlockchainLMDB::for_all_transactions(std::function<bool(const crypto::hash&
       break;
     if (ret)
       throw0(DB_ERROR(lmdb_error("Failed to enumerate transactions: ", ret).c_str()));
-    transaction tx;
+    blobdata owned_txblob;
+    blobdata_ref bdref;
     if (pruned)
     {
-      blobdata_ref bd{reinterpret_cast<char*>(v.mv_data), v.mv_size};
-      if (!parse_and_validate_tx_base_from_blob(bd, tx))
```
0xFFFC0000, line 3690: Question, what do you think about keeping `parse_and_validate_tx_base_from_blob`  in `#ifdef DEBUG` preprocessor? Although we don't use debug macro in `db_lmdb.cpp`. 

I realize that you want to remove transaction-specific code from db_lmdb,  but at the same time, it looks like a valuable check, particularly when debugging.

0xFFFC0000 (commented):
Other than a few minor questions, everything looks good to me. I will look at this again a few days later to make sure I am not missing any specific details.

src/blockchain_db/lmdb/db_lmdb.cpp
jeffro256: Check the other `BlockchainDB::for_all_transactions` method. We do still have the `for_all_transactions()` method, but now the parsing is done outside of this class.

jeffro256:
This is not as much of an issue due to the FCMP++ integration still using `cryptonote::transaction`. Ideally, something like this should still be merged, but it can be revisited after the fork.

PR #5915, opened 2019-09-16
monerod can now sync from pruned blocks

moneromooo-monero:
If the peer (whether pruned or not itself) supports sending pruned blocks
to syncing nodes, the pruned version will be sent along with the hash
of the pruned data and the block weight. The original tx hashes can be
reconstructed from the pruned txes and theur prunable data hash. Those
hashes and the block weights are hashes and checked against the set of
precompiled hashes, ensuring the data we received is the original data.
It is currently not possible to use this system when not using the set
of precompiled hashes, since block weights can not otherwise be checked
for validity.

This is off by default for now, and is enabled by --sync-pruned-blocks

PR #3191, opened 2018-01-27
core: add --no-fluffy-blocks, and enable fluffy blocks by default

moneromooo-monero:

PR #9180, opened 2024-02-19
Add --no-sync-pruned-blocks, and enable syncing pruned blocks by default for pruned nodes

### LINKED ISSUE #8834: Consider enabling --sync-pruned-blocks by default for pruned nodes
The `--sync-pruned-blocks` flag, added in #5915, allows a pruned node to skip downloading the full content of certain blocks, when they would otherwise be pruned locally. This flag appears to be largely forgotten about, as I've seen little mention of it at all.

This should be enabled by default when running a pruned node, eg with `--prune-blockchain` set. Normally, pruned nodes must download the entire blockchain, despite the fact that they are going to discard roughly 2/3 of it after. This adds unnecessary strain to not only the syncing node's connection, but to the network as a whole. Since that 2/3 will always be discarded anyway, there seems to be no reason to not have this flag set by default.

expiredhotdog:
Resolves #8834.


src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -387,6 +394,11 @@ namespace cryptonote
     m_fluffy_blocks_enabled = !get_arg(vm, arg_no_fluffy_blocks);
     m_offline = get_arg(vm, arg_offline);
     m_disable_dns_checkpoints = get_arg(vm, arg_disable_dns_checkpoints);
+    m_sync_pruned_blocks = !command_line::get_arg(vm, cryptonote::arg_no_sync_pruned_blocks) 
+        && command_line::get_arg(vm, cryptonote::arg_fast_block_sync) != 0
+        && command_line::get_arg(vm, cryptonote::arg_prune_blockchain);
```
expiredhotdog, line 399: This could probably be simplified. The extra checks are [likely unnecessary](https://github.com/monero-project/monero/pull/5915#issuecomment-531772037), but I figured it's best to double check.

selsta (commented):
What do you think about changing 

```
const command_line::arg_descriptor<bool> arg_sync_pruned_blocks  = {
```

to

```
const command_line::arg_descriptor<unit64_t> arg_sync_pruned_blocks  = {
```

and then setting it to 1 as default value. Then someone who wants to disable it can do `--sync-pruned-blocks 0` and we don't have to add another flag. See the `--fast-block-sync` flag as an example that works this way.

expiredhotdog:
@selsta This change was modeled after #3191, with the intent of doing the same as #8855 at some point in the future. The reasoning for adding --no-fluffy-blocks at the time was to avoid breaking anyone's scripts, which I somewhat agree with.

I'm happy to change it if you think that would be better, though.

src/cryptonote_core/cryptonote_core.cpp
expiredhotdog: I'm now fully sure that these checks are redundant. Since they may cause weird edge cases, they were removed.

PR #9195, opened 2024-02-21
free function serialization DSL

jeffro256:
Upstreaming https://github.com/UkoeHB/monero/pull/32


0xFFFC0000 (approved):
LGTM. 

If you add a comment to your PR first post with an example, that would make it much better.  I followed the discussion in Koe's repo, and these comments [1] look nice. Maybe a shorter version of these comments, in the commit message or your PR page would improve it.  

The case I am thinking about is: that later when someone runs a git-blame or looks at this PR page, comments/explanations should be obvious. 

1. https://github.com/UkoeHB/monero/pull/32#issuecomment-1957674491

0xFFFC0000 (approved):

PR #9203, opened 2024-02-25
simplewallet: edit desc. text for transfer

plowsof:
patch sent from Cat for discussion. 

help text shows transfer `<address> <amount>`

 https://github.com/monero-project/monero/blob/7b7958bbd9d76375c47dc418b4adabba0f0b1785/src/simplewallet/simplewallet.cpp#L2985

but part of the description text shows `Transfer <amount> to <address>`. The Usage text is correct, however, the description could lead to confusion / re-enforce incorrect usage of transfer.

opinions appreciated, thank you.


nahuhh (commented):
Perhaps a good idea to remove the `lockbox` from both and add `subtractfeefrom` to the latter

also needs rebase (and i'd argue release version)

plowsof:
oops, trimmed some whitespace not in release

tobtoht (approved):

PR #9204, opened 2024-02-25
depends: native_libtapi: don't embed git hash

tobtoht:
The `native_libtapi` depends package is not reproducible because it embeds the checked out git hash of whatever git repository it happens to have been extracted to, which is that of the monero repo in our case. If the `native_libtapi` package is cached, runs for different monero commits that otherwise do not invalidate the cache will use a `native_libtapi` package with a hash that does not match the hash of a clean build.

This does not make its way into release binaries, but hinders automated reproducibility scanning.

### DIFF
--- a/contrib/depends/patches/native_libtapi/no_embed_git_rev.patch
+++ b/contrib/depends/patches/native_libtapi/no_embed_git_rev.patch
@@ -1,0 +1,31 @@
+diff --git a/src/llvm/CMakeLists.txt b/src/llvm/CMakeLists.txt
+index ab92717c8..4ad621ea3 100644
+--- a/src/llvm/CMakeLists.txt
++++ b/src/llvm/CMakeLists.txt
+@@ -752,9 +752,10 @@ set(LLVM_SRPM_USER_BINARY_SPECFILE ${CMAKE_CURRENT_SOURCE_DIR}/llvm.spec.in
+ set(LLVM_SRPM_BINARY_SPECFILE ${CMAKE_CURRENT_BINARY_DIR}/llvm.spec)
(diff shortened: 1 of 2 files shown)


selsta (approved):

contrib/depends/patches/native_libtapi/no_embed_git_rev.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/src/llvm/CMakeLists.txt b/src/llvm/CMakeLists.txt
+index ab92717c8..dff3a292a 100644
+--- a/src/llvm/CMakeLists.txt
++++ b/src/llvm/CMakeLists.txt
+@@ -754,7 +754,6 @@ set(LLVM_SRPM_DIR "${CMAKE_CURRENT_BINARY_DIR}/srpm")
+ 
+ # SVN_REVISION and GIT_COMMIT get set by the call to add_version_info_from_vcs.
```
hyc, line 7: Should update these comments to explain what's really happening here now.

contrib/depends/patches/native_libtapi/no_embed_git_rev.patch
```diff
@@ -0,0 +1,25 @@
...
+ # DUMMY_VAR contains a version string which we don't care about.
```
hyc, line 8: Since you're deleting the call to add_version_info_from_vcs() it seems the above comments are no longer relevant or accurate. Is there anything you can add here that makes it clear what's really happening now?

tobtoht: I updated the comment.

PR #9205, opened 2024-02-26
depends: fix ar zero timestamp for openssl & unbound on android targets

tobtoht:
The Unbound and OpenSSL depends packags are not reproducible when building android targets because the Android NDK's bundled archiver includes timestamps by default. The default options are not overridden properly for these packages.

The archiver lies about zeroing timestamps by default.

>  [D]          - use zero for timestamps and uids/gids (default)

This PR sets the `ARFLAGS` for these packages to the flags defined in the hosts file, [which includes `D` for android](https://github.com/monero-project/monero/blob/master/contrib/depends/hosts/android.mk#L13).

Note for OpenSSL: it only reads `ARFLAGS` (not `AR_FLAGS`) when passed as an argument.

### DIFF
--- a/contrib/depends/packages/openssl.mk
+++ b/contrib/depends/packages/openssl.mk
@@ -4,9 +4,12 @@ $(package)_download_path=https://www.openssl.org/source
 $(package)_file_name=$(package)-$($(package)_version).tar.gz
 $(package)_sha256_hash=88525753f79d3bec27d2fa7c66aa0b92b3aa9498dafd93d7cfa4b3780cdae313
 
+# The bundled ranlib in Android NDK 18b inserts timestamps by default.
+# To prevent reproducibility issues, we must enable [D]eterministic mode.
+
 define $(package)_set_vars
-$(package)_config_env=AR="$($(package)_ar)" ARFLAGS=$($(package)_arflags) RANLIB="$($(package)_ranlib)" CC="$($(package)_cc)"
-$(package)_config_env_android=ANDROID_NDK_ROOT="$(host_prefix)/native" PATH="$(host_prefix)/native/bin" CC=clang AR=ar RANLIB=ranlib
+$(package)_config_env=AR="$($(package)_ar)" RANLIB="$($(package)_ranlib)" CC="$($(package)_cc)"
+$(package)_config_env_android=ANDROID_NDK_ROOT="$(host_prefix)/native" PATH="$(host_prefix)/native/bin" CC=clang AR=ar RANLIB="ranlib -D"
 $(package)_build_env_android=ANDROID_NDK_ROOT="$(host_prefix)/native"
 $(package)_config_opts=--prefix=$(host_prefix) --openssldir=$(host_prefix)/etc/openssl --libdir=$(host_prefix)/lib
 $(package)_config_opts+=no-capieng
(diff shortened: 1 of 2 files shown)


selsta (approved):

tobtoht:
Ranlib isn't deterministic either, lol.

```
$ diffoscope openssl-3.0.13-1b915ee1efd.tar.gz openssl-3.0.13-1b915ee1efd.tar.gz.1
--- openssl-3.0.13-1b915ee1efd.tar.gz
+++ openssl-3.0.13-1b915ee1efd.tar.gz.1
├── openssl-3.0.13-1b915ee1efd.tar
│ ├── ./lib/libcrypto.a
│ │ ├── file list
│ │ │ @@ -1,8 +1,8 @@
│ │ │ -----------   0        0        0   205940 2024-02-26 09:54:16.000000 /
│ │ │ +----------   0        0        0   205940 2024-02-25 23:59:58.000000 /
│ │ │  ----------   0        0        0        0 1970-01-01 00:00:00.000000 //
│ │ │  ?rw-r--r--   0        0        0     1328 1970-01-01 00:00:00.000000 libcrypto-lib-aes_cbc.o
│ │ │  ?rw-r--r--   0        0        0     1504 1970-01-01 00:00:00.000000 libcrypto-lib-aes_cfb.o
│ │ │  ?rw-r--r--   0        0        0    14640 1970-01-01 00:00:00.000000 libcrypto-lib-aes_core.o
```

At least it's honest about it:

```
Usage: aarch64-linux-android-ranlib [options] archive
 Generate an index to speed access to archives
 The options are:
  @<file>                      Read options from <file>
  --plugin <name>              Load the specified plugin
  -D                           Use zero for symbol map timestamp
  -U                           Use actual symbol map timestamp (default)
```

Force push sets `-D` for android.

contrib/depends/packages/openssl.mk
```diff
@@ -5,8 +5,8 @@ $(package)_file_name=$(package)-$($(package)_version).tar.gz
 $(package)_sha256_hash=88525753f79d3bec27d2fa7c66aa0b92b3aa9498dafd93d7cfa4b3780cdae313
 
 define $(package)_set_vars
-$(package)_config_env=AR="$($(package)_ar)" ARFLAGS=$($(package)_arflags) RANLIB="$($(package)_ranlib)" CC="$($(package)_cc)"
-$(package)_config_env_android=ANDROID_NDK_ROOT="$(host_prefix)/native" PATH="$(host_prefix)/native/bin" CC=clang AR=ar RANLIB=ranlib
+$(package)_config_env=AR="$($(package)_ar)" RANLIB="$($(package)_ranlib)" CC="$($(package)_cc)"
+$(package)_config_env_android=ANDROID_NDK_ROOT="$(host_prefix)/native" PATH="$(host_prefix)/native/bin" CC=clang AR=ar RANLIB="ranlib -D"
```
iamamyth, line 9: Thanks for spotting this issue and submitting the PR. In my experience, with subtle issues like this one, a comment here in the source (describing the rationale for the flag, which you could probably just lift from the PR) would help future readers save a lot of time (the fact that the documentation and behavior drifts across platforms makes it very hard for a new contributor to understand the issue).

tobtoht: I added a comment.

iamamyth (approved):

PR #9211, opened 2024-02-29
monero-wallet-rpc - Changed default cryptonote port

SyntheticBird45:
- Removed `localhost:8080` default value to address string in wallet rpc
- If `set_daemon` rpc address string is empty, mwr will return an error and explicit message

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -30,6 +30,7 @@
 
 #include <algorithm>
 #include <numeric>
+#include <string>
 #include <tuple>
 #include <queue>
 #include <boost/format.hpp>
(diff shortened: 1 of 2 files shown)


SyntheticBird45:
ready

src/wallet/wallet_rpc_server.cpp
```diff
@@ -34,6 +34,7 @@
 #include <boost/preprocessor/stringize.hpp>
 #include <cstdint>
 #include "include_base_utils.h"
+#include "wallet_rpc_server_error_codes.h"
```
vtnerd, line 37: Why this change?

SyntheticBird45: first commit i added an error message and the WRONG_ADDRESS core rpc error. Clangd automatically included its header. forgot to remove it

SChernykh (approved):

SChernykh (approved):

rbrunner7 (approved):

PR #9214, opened 2024-03-05
unit_tests: fix strtoul unit test

jeffro256:
`errno` not guaranteed to be anything according to C standard, except on range error

### DIFF
--- a/tests/unit_tests/epee_utils.cpp
+++ b/tests/unit_tests/epee_utils.cpp
@@ -1899,7 +1899,6 @@ TEST(parsing, strtoul)
   p = "q";
   endp = nullptr;
   ul = std::strtoul(p, const_cast<char**>(&endp), 10);
-  EXPECT_EQ(0, errno);
   EXPECT_EQ(0, ul);
   EXPECT_EQ(p, endp);
 


tobtoht (approved):

iamamyth:
>  errno not guaranteed to be anything according to C standard, except on range error

That's exactly why the test sets errno. C functions which interact with errno change it when an error occurs (per the [man page]( https://man7.org/linux/man-pages/man3/strtoul.3.html)), and otherwise leave it alone; it would be ridiculous to expect functions to document what they *don't* do.

tests/unit_tests/epee_utils.cpp
```diff
@@ -1859,47 +1859,39 @@ TEST(parsing, strtoul)
...
   EXPECT_EQ(p + 1, endp);
 
   p = "    \t   0";
   endp = nullptr;
   ul = std::strtoul(p, const_cast<char**>(&endp), 10);
-  EXPECT_EQ(0, errno);
   EXPECT_EQ(0, ul);
   EXPECT_EQ(p + 9, endp);
 
   p = "q";
```
iamamyth, line 1892: This part of the test would be the most likely to break across systems: Per the man page, some systems report EINVAL here: "EINVAL (not in C99) The given base contains an unsupported value." The c++ function makes no mention of EINVAL, so perhaps whoever wrote this test suite wanted a way to determine the actual behavior in a given build.

jeffro256: I think you meant to put this comment on the test immediately below this one, yeah? According to the C standard, `strtoul` should skip over whitespace. But yes, I agree with the general point, so I updated the commit to just remove the `EXPECT_EQ(0)` from the invalid character test.

selsta (approved):
Fixes this test failing on macOS

PR #9219, opened 2024-03-07
wallet2: adjust fee during backlog, fix set priority

### LINKED ISSUE #9221: `set priority` appears broken
```
[wallet 431t1S]: set priority 2
Wallet password: 
[wallet 431t1S]: set
priority = 2 (normal)
[wallet 431t1S]: address
0  431t1SuHXJNE3U5GhqoGCw2t5XpAZRwyX2h3AfG5BKzkZuYWrVaEyyRdHGaXUMLhiCAc5mF5EaLN9cqie64a5PA3TXTiw3g  Primary address (used)
[wallet 431t1S]: transfer 431t1S 0.01
Wallet password: 
There is currently a 37 block backlog at that fee level. Is this okay?  (Y/Yes/N/No): n
Error: transaction cancelled.
```

selsta:
Resolves #9221

The automatic fee level should switch between slow and normal depending on the current backlog.

https://monero.stackexchange.com/questions/11657/what-are-the-standard-fee-multiplier-levels#comment10765_11658

This behaviour broke at some point, likely during the 2021 fee code changes. This PR restores the previous behaviour.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -6313,7 +6313,7 @@ bool simple_wallet::transfer_main(int transfer_type, const std::vector<std::stri
     local_args.erase(local_args.begin());
   }
 
-  uint32_t priority = 0;
+  uint32_t priority = m_wallet->get_default_priority();
   if (local_args.size() > 0 && parse_priority(local_args[0], priority))
     local_args.erase(local_args.begin());
 
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -8118,7 +8118,7 @@ uint32_t wallet2::adjust_priority(uint32_t priority)
       else if (blocks[0].first > 0)
       {
         MINFO("We don't use the low priority because there's a backlog in the tx pool.");
-        return priority;
+        return 2;
       }
 
       // get the current full reward zone
@@ -8163,7 +8163,7 @@ uint32_t wallet2::adjust_priority(uint32_t priority)
       if (P > 80)
       {
         MINFO("We don't use the low priority because recent blocks are quite full.");
-        return priority;
+        return 2;
       }
       MINFO("We'll use the low priority because probably it's safe to do so.");
       return 1;


jeffro256:
I don't think that I agree with these changes. Sometimes, one knows that there's a backlog, but trusts that their transaction will make it into the block in an acceptable timeframe. Forcing the priority to rise because of a backlog gives attackers more leverage to disrupt the network.

SamsungGalaxyPlayer:
The intent of the automatic feature has always been to alternate between the slow and normal fees levels as appropriate. I recommend that unless there's a really strong reason to completely change the logic behind these fee level adjustments, the discrepancy versus the expected behavior is addressed.

selsta:
@jeffro256 the current code is broken since the 2021 fee changes. this PR fixes that back to the previous behavior.

LocalMonero:
This returns the expected behavior. The mempool spam would be much less noticeable by common users had this bug not been introduced in 2021. Thanks @selsta for fixing it so quickly, and @luigi1111 please merge this urgently and release the new binaries for everyone.

selsta:
Should I also change the remaining `return priority;` to either 1 or 2?

nahuhh (approved):
LGTM 

anonero wallet's @r4v3r23 pulled pr and tested on anonero.

r4v3r23:
standard tx fees are now ~4x the size than previous code's behavior 

+1 from me

osensei:
So, I've done some tests with these changes, and it seems to work as expected ONLY when wallet's settings are: 
`priority = 0` AND `auto-low-priority = 1`

In any other case the `priority` setting seems to be ignored, and will always send with the minimum fee, even if you set `priority = 4` AND `auto-low-priority = 0`. This also happens on v0.18.3.1.

In those other cases, the only way to override the minimum fee is to specify the priority in the transfer command, .e.g.: 
`transfer normal <address> <amount>`

osensei:
> In any other case the `priority` setting seems to be ignored, and will always send with the minimum fee, even if you set `priority = 4` AND `auto-low-priority = 0`. This also happens on v0.18.3.1.

I understand that my observations may not apply directly to this PR, but just wanted to state the facts, as it seemed relevant. Also, I just  saw that @selsta opened #9221 related to this.

SChernykh (approved):

SChernykh (approved):

PR #9222, opened 2024-03-08
get_block_template_backlog: better sorting logic

SChernykh:
`std::sort` is unstable, so it can return random sets of transactions when mempool has many transactions with the same fee/byte. It can result in p2pool mining empty blocks sometimes because it doesn't pick up "new" transactions immediately.

`std::stable_sort` should preserve the order of transactions that was returned by the DB request.

Example when P2Pool was mining an empty block despite full mempool:
```
NOTICE  2024-03-08 12:34:06.0670 P2Pool new miner data
---------------------------------------------------------------------------------------------------------------
host                    = 127.0.0.1:18081:ZMQ:18083
major_version           = 16
height                  = 3100556
prev_id                 = 93c317bb130f77459a0698d4a86c1bad815ca40a6c3fa513fecfd0d751635213
seed_hash               = 4ec15c377ac3576f49361d2e912ecb2256c438937821e7262508d5995b37c7e6
difficulty              = 248886870954
median_weight           = 329841
already_generated_coins = 18407529506769594824
transactions            = 231
---------------------------------------------------------------------------------------------------------------
NOTICE  2024-03-08 12:34:06.0673 P2Pool median timestamp updated to 1709897381
NOTICE  2024-03-08 12:34:06.0673 P2Pool new main chain block: height = 3100555, id = 93c317bb130f77459a0698d4a86c1bad815ca40a6c3fa513fecfd0d751635213, timestamp = 1709901243, reward = 0.600811880000 XMR
NOTICE  2024-03-08 12:34:06.0682 BlockTemplate mempool has 231 transactions, taking 0 transactions from it
NOTICE  2024-03-08 12:34:06.0682 BlockTemplate base  reward = 0.600000000000 XMR, 0 transactions, fees = 0.000000000000 XMR, weight = 0
NOTICE  2024-03-08 12:34:06.0866 BlockTemplate final reward = 0.600000000000 XMR, weight = 23984, outputs = 616, 0 of 0 transactions included
```

### DIFF
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -1073,7 +1073,7 @@ namespace cryptonote
 
     // If the total weight is too high, choose the best paying transactions
     if (total_weight > max_backlog_weight)
-      std::sort(tmp.begin(), tmp.end(), [](const auto& a, const auto& b){ return a.fee * b.weight > b.fee * a.weight; });
+      std::stable_sort(tmp.begin(), tmp.end(), [](const auto& a, const auto& b){ return a.fee * b.weight > b.fee * a.weight; });
 
     backlog.clear();
     uint64_t w = 0;


jeffro256:
> that was returned by the DB request.

Is there any inherent information in the tx order of the DB response (e.g. first seen)?

SChernykh:
I don't know, I assume it shouldn't change between transactions that stay in the DB between two calls to `BlockchainLMDB::for_all_txpool_txes`. @hyc can you comment on this?

SChernykh:
@jeffro256 so far so good, 15 blocks mined by the network and P2Pool picked almost all transactions from `get_block_template_backlog` every time (sometimes it doesn't pick 3-4 transations but that's expected - they can be new). So the order returned from DB stays the same.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1073,7 +1073,7 @@ namespace cryptonote
 
     // If the total weight is too high, choose the best paying transactions
     if (total_weight > max_backlog_weight)
-      std::sort(tmp.begin(), tmp.end(), [](const auto& a, const auto& b){ return a.fee * b.weight > b.fee * a.weight; });
+      std::stable_sort(tmp.begin(), tmp.end(), [](const auto& a, const auto& b){ return a.fee * b.weight > b.fee * a.weight; });
```
jeffro256, line 1076: Isn't this formula wrong anyways? Shouldn't it be `x.fee / x.weight` instead of `*`? Right now, it's maximizing fee times weight, not fee per weight.

jeffro256 (approved):

0xFFFC0000:
This [1] is my understanding from LMDB design. Of course @hyc will be ultimate the authority on this. 

https://stackoverflow.com/a/39850512

SChernykh:
I'm not sure. The question is how `mdb_cursor_get` works with MDB_FIRST/MDB_NEXT operations (not MDB_SET).

0xFFFC0000:
I don't see any exotic reordering here. Now as far as I am aware lmdb doesn't do any reordering offline. Now take all of these information with grain of salt.


1. https://github.com/LMDB/lmdb/blob/b8e54b4c31378932b69f1298972de54a565185b1/libraries/liblmdb/mdb.c#L7030

SChernykh:
Yes, it looks like it just goes through db pages and lists items in the order they are physically stored in those pages. So there shouldn't be any random reordering between calls, even if some transactions are added/removed. My node has been running for a few hours and I haven't seen this issue anymore.

hyc:
Correct, if there have been no insertions or deletions, the set of returned items will always be the same and in the same order. LMDB never does any background processing, what you wrote is exactly what you get back.

hyc (approved):

SChernykh:
And if there were some insertions/deletions, what will be the relative order of the other items (which weren't touched)?

hyc:
Keys are stored in sorted order. If A was before B before any insert/delete, it will still be before B after. Records don't just get scrambled into a new order arbitrarily.

PR #9232, opened 2024-03-10
Blockchain: remove old fee calc logic

jeffro256:
Reduce code and binary size by removing unused code paths related to fee calculations.


hyc:
Is it really totally unused, or just unused in the current hardfork version? Do nodes need to use this code when verifying blocks from the earlier part of the blockchain?

jeffro256:
The only location that `Blockchain::check_fee()` is called is in `tx_memory_pool::add_tx()`, and only for txs which are not "kept by block" (i.e. only called for ephemeral mempool validation and not for chain validation): https://github.com/monero-project/monero/blob/1bec71279e43bfde46506409377630c7e020a1cf/src/cryptonote_core/tx_pool.cpp#L211

You can verify this yourself with `git grep check_fee`.

selsta (approved):

PR #9242, opened 2024-03-12
Copyright: Update to 2024

gStart9:


selsta:
Thank you for the pull request, we have a CI script that updates copyright and don't accept manual changes for it so I will close this.

https://github.com/monero-project/monero/blob/master/.github/workflows/copyright.yml

PR #9257, opened 2024-03-19
serialization: do any N-tuple

jeffro256:


src/serialization/tuple.h
```diff
@@ -51,119 +51,71 @@ namespace serialization
...
-  if (!::serialization::detail::serialize_tuple_element(ar, std::get<2>(p)))
-    return false;
-  if (!ar.good())
-    return false;
-  ar.delimit_array();
-  if (!::serialization::detail::serialize_tuple_element(ar, std::get<3>(p)))
-    return false;
-  if (!ar.good())
-    return false;
+#define TUPLE_COMPACT_FIELDS(v)                          \
```
vtnerd, line 107: Why is this a separate macro, couldn't this just be inside `TUPLE_COMPACT_FIELD_N`?

src/serialization/tuple.h
```diff
@@ -51,119 +51,71 @@ namespace serialization
...
+    if constexpr (BackwardsCompat && (tuple_size == 3 || tuple_size == 4))
+    {
+      size_t cnt = tuple_size;
+      ar.begin_array(cnt);
+      if (cnt != tuple_size)
+        return false;
+    }
+    else
+    {
+      ar.begin_array();
```
vtnerd, line 72: I never paid attention to this serializer before, but isn't this broken for JSON? You can't have an array because each element of the tuple could be different type. So the outter type must be an object? Is the compatibility mode always using integers (or at least something all the same)?

I _think_ it works for the binary archive, but I haven't looked too closely. I'm assuming the tests verify that the heterogenous arrays do in fact work.

Another nitpick - how are new fields added to a tuple? There's no version field or anything to change the size of the tuple later.

src/serialization/tuple.h
jeffro256: `TUPLE_COMPACT_FIELDS` macro is similar to the `FIELDS` macro in that it doesn't tag.

src/serialization/tuple.h
jeffro256: > I never paid attention to this serializer before, but isn't this broken for JSON? You can't have an array because each element of the tuple could be different type. So the outter type must be an object?

From [RFC 8259, Section 5](https://www.rfc-editor.org/rfc/rfc8259#section-5): "There is no requirement that the values in an array be of the same type."

> Is the compatibility mode always using integers (or at least something all the same)?

The compatibility mode serializes a 3 or 4 varint value before the array in a binary archive for 3-tuples and 4-tuples, respectively, since that is how the old code was written. I chose not to do this with tuples of other sizes, or if `BackwardsCompat=false` since that size value is redundant; we know the size of a tuple at compile time. 

 > Another nitpick - how are new fields added to a tuple? There's no version field or anything to change the size of the tuple later.

They aren't. An N-tuple is an N-tuple. IMO, if you want to serialize an object that might be extended in the future, you should just create that as a struct then use an custom serializer (with a version field).

vtnerd: > From RFC 8259, Section 5: "There is no requirement that the values in an array be of the same type."

Egg on face for not reading the spec.

vtnerd (approved):

src/serialization/tuple.h
vtnerd: > They aren't. An N-tuple is an N-tuple. IMO, if you want to serialize an object that might be extended in the future, you should just create that as a struct then use an custom serializer (with a version field).

Normally I would say every struct/tuple should have a version or length field "just-in-case", but given the use case reducing things to a minimum makes sense. Hopefully this works out with future revisions.

PR #9263, opened 2024-03-26
I2P SAM integration

preland:
Very much a WIP at the moment.

When finished, this PR will allow monerod to interface with an I2P router using the SAM protocol. The implementation will be based off of the implementation in bitcoin, as that implementation appears to be the closest fit to our needs.


selsta:
What is the benefit of using SAM over our current I2P implementation?

preland:
There’s a few, the main one being that the current implementation was built for the i2p-zero router, which is no longer being developed, and the way they were doing it no longer works. In addition, using SAM enables all (modern) I2P routers to interact with monero.

selsta:
Our current implementation works with i2pd, Java I2P, i2p-zero and possibly more. There is no i2p-zero specific code that I'm aware of.

@vtnerd while implementing i2p support, did you look into SAM? Is it worth the extra code that we would have to maintain?

preland:
Btw the current line count is misleading; it’ll be decreased significantly as I properly integrate the code (it is currently mostly just header files from bitcoin currently, which will be changed.)

As for I2P-zero, you are correct that there isn’t any explicit code that directly supports it over other routers.

That being said, the link on getmonero.org [is designed for/by I2P-zero](https://www.getmonero.org/resources/user-guides/node-i2p-zero.html), and the seed-nodes listed do not work anymore afaik (if anything they are unstable, as I was unable to connect to any of them even once).

In addition, this only does transactions through I2P—syncing is still done on clearnet, which may not be desirable for some users. The only workaround would be to use an http proxy, which is also a security concern.

The main goal for this is so that, in the long term, I2P integration can be enabled directly in the GUI by simply toggling on the SAM protocol, and allowing them to either only do transactions or do everything via I2P.

vtnerd:
@selsta :

> @vtnerd while implementing i2p support, did you look into SAM? Is it worth the extra code that we would have to maintain?

I looked into it, but implementing Socks was easier/faster and provided support for I2P+Tor (and SSH with `--proxy` in wallet and daemon). I also don't recall a Bitcoin implementation at the time, but I may have missed it. 

The primary benefit of SAM is that the `--anonymous-inbound` option is auto configured (almost). With SAM, the inbound connections can be transient (per `monerod` boot), or stored in a config file local to `monerod` (for seed nodes, etc). So it's arguably simpler to configure; most configurations would use the defaults for file location and only need one option to specify the I2P SAM port. The inbound I2P routing should also be "torn down" when `monerod` exits, so there is a small benefit in pruning the internal routing scheme (which the end user won't really notice, but may be beneficial to the I2P network).

Whether it's worth all the extra code, isn't clear to me, since the current configuration isn't _too_ difficult once a tutorial can be followed. And even with SAM, some tutorial is likely necessary to explain how to setup SAM and ensure that `monerod` is using it.

Lastly, this current draft has some duplicated code that should be removed, so the amount of extra code is going to be less than what it appears right now.

---------------------

@preland :

> That being said, the link on getmonero.org [is designed for/by I2P-zero](https://www.getmonero.org/resources/user-guides/node-i2p-zero.html), and the seed-nodes listed do not work anymore afaik (if anything they are unstable, as I was unable to connect to any of them even once).

Changing the tutorials is trivial though, and I believe there are other tutorials that don't rely on i2p-zero.

nahuhh:
Zero isnt necessary at all
just a wrapper..

i2pd is preferred, but the i2p implementation doesnt matter. i2p(java and zero) and i2pd all have sam support. we currently use socks.

For tor, a similar solution is to use a torcontrol port and password to autoconfigure the onion.

other blockchains allow block sync over these networks. Are we moving there in this pr?

vtnerd:
> other blockchains allow block sync over these networks. Are we moving there in this pr?

This can be done immediately without SAM support by removing a few checks. Syncing over these networks was never allowed due to concerns over possible sybil attacks - you can't easily block a tor/i2p peer. We could let the users decide for themselves instead.

selsta:
Closing in favor of https://github.com/monero-project/monero/pull/10523 which is CCS funded and had some feedback from I2P developer.

PR #9265, opened 2024-03-28
Skip privacy networks (on tx sends) that don't have outgoing connections

vtnerd:
This will check if the privacy network has outgoing connections before attempting to send over that network. The primary difference is that this will allow sending txes over Tor if I2P has no outgoing connections.


vtnerd:
This isn't going to work. Will post a better one shortly.

PR #9266, opened 2024-03-28
Skip privacy networks (on tx sends) that don't have outgoing connections [0.18]

vtnerd:
Same as #9265 except on release branch. @selsta you may want to test this patch.


vtnerd:
Closing like with #9265 because this isn't going to work.

PR #9267, opened 2024-03-29
Skip privacy networks (on tx sends) that don't have outgoing connections [0.18]

vtnerd:
This will check if the privacy network has outgoing connections before attempting to send over that network. The primary difference is that this will allow sending txes over Tor if I2P has no outgoing connections.

This differs from the PR yesterday in that it checks the outgoing checks directly instead of the count reported by the noise code.

### DIFF
--- a/src/cryptonote_protocol/levin_notify.cpp
+++ b/src/cryptonote_protocol/levin_notify.cpp
@@ -741,9 +741,14 @@ namespace levin
   notify::status notify::get_status() const noexcept
   {
     if (!zone_)
-      return {false, false};
+      return {false, false, false};
 
-    return {!zone_->noise.empty(), CRYPTONOTE_NOISE_CHANNELS <= zone_->connection_count};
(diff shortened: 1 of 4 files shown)


vtnerd:
I wanted to make this is a draft, oh well @selsta I will post the `master` version if this patch works.

nahuhh:
Thank you 🙏

plowsof:
i first became aware of this bug after @nahuhh mentioned having similar issues with i2p / confirmed it was happening (along with the i2p phantom/double port issue), this is a very important fix. if you consider that i am hosting public nodes who can without notice behave like a malicious one not broadcasting peoples transactions it is quite stressful. will be testing this

selsta (approved):

nahuhh:
I'll test for a few days and report back.
I run i2p on a different device, so should catch any issues.

nahuhh:
tested for a couple weeks on android with external i2p 
LGTM

nahuhh (approved):

PR #9268, opened 2024-03-30
Skip privacy networks (on tx sends) that don't have outgoing connections

vtnerd:
Same as #9267 except on master branch.


selsta (approved):
Tested on my nodes in various configurations.

PR #9270, opened 2024-03-30
common: fix privatefile create. 

0xFFFC0000:
When creating a private file we need to delete the file if exists.

### DIFF
--- a/src/common/util.cpp
+++ b/src/common/util.cpp
@@ -121,7 +121,7 @@ namespace tools
   private_file::private_file(std::FILE* handle, std::string&& filename) noexcept
     : m_handle(handle), m_filename(std::move(filename)) {}
 
-  private_file private_file::create(std::string name)
+  private_file private_file::create(std::string name, uint32_t extra_flags)
   {
 #ifdef WIN32
     struct close_handle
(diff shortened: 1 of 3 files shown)


vtnerd (approved):

selsta:
@0xFFFC0000 please also open against release branch

src/common/util.cpp
```diff
@@ -123,6 +123,14 @@ namespace tools
 
   private_file private_file::create(std::string name)
   {
+    if (epee::file_io_utils::is_file_exist(name)) {
```
iamamyth, line 126: This change seems structurally and semantically quite dubious:
1. Create tends to be a one-way operation, very different from drop and recreate. If you want to support a "replace" operation, you'd be better off making a new function which wraps the existing and allow the caller here to optionally request O_EXCL.
2. The existing usage of this function that I see in wallet_rpc_server seems well aware it *doesn't* remove the existing file; what caller does this help, or contract (documented in the comments or implied by usage) does it honor?

Relevant wallet_rpc_server code:
```
        rpc_login_file = tools::private_file::create(temp);
        if (!rpc_login_file.handle())
        {
          LOG_ERROR(tr("Failed to create file ") << temp << tr(". Check permissions or remove file"));
          return false;
        }
```

0xFFFC0000: Thanks for your comment @iamamyth 
Few points:

1. Apologies for the late reply, I have been extremely busy the past few weeks.
2. I thought about your comment. And your criticism makes sense to me.

In the new push, I am addressing those issues you have raised.

selsta (approved):

src/common/util.cpp
iamamyth: Looks good; thanks for spotting and fixing the bug.

iamamyth (approved):

PR #9273, opened 2024-03-31
build: miniupnpc: remove submodule, add to depends, update to 2.3.3

tobtoht:
Supply chain security checklist:

- [X] The source tarball is GPG signed
- [X] The source tarball does not contain binaries or archives
- [X] The source tarball does not contain auto-generated scripts
- [X] The source tarball does not differ from git source in a meaningful way 
- [X] The package has not changed ownership since the last update (2.2.1)

https://whatsrc.org/artifact/sha256:d52a0afa614ad6c088cc9ddff1ae7d29c8c595ac5fdd321170a05f41e634bd1a

Why?

- Keep release dependencies in one place (`depends`)
- Allows us to patch the package without forking the repository
- For development builds: `miniupnpc` is packaged for all relevant distributions

### DIFF
--- a/contrib/depends/funcs.mk
+++ b/contrib/depends/funcs.mk
@@ -197,9 +197,11 @@ ifeq ($($(1)_type),build)
 $(1)_cmake += -DCMAKE_INSTALL_RPATH:PATH="$$($($(1)_type)_prefix)/lib"
 else
 ifneq ($(host),$(build))
+ifneq ($(host_os),android)
 $(1)_cmake += -DCMAKE_SYSTEM_NAME=$($(host_os)_cmake_system)
 $(1)_cmake += -DCMAKE_C_COMPILER_TARGET=$(host)
 $(1)_cmake += -DCMAKE_CXX_COMPILER_TARGET=$(host)
+endif
 endif
 endif
 endef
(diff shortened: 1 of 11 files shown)


0xFFFC0000:
We can only merge this or https://github.com/monero-project/monero/issues/9367

contrib/depends/funcs.mk
```diff
@@ -197,11 +197,13 @@ ifeq ($($(1)_type),build)
 $(1)_cmake += -DCMAKE_INSTALL_RPATH:PATH="$$($($(1)_type)_prefix)/lib"
 else
 ifneq ($(host),$(build))
+ifneq ($(host_os),android)
```
tobtoht, line 200: Passing `--target` to the Android toolchain makes it not properly find files in the sysroot. (https://github.com/tobtoht/monero/actions/runs/24902480897/job/72923419519#step:14:4502)

We haven't run into this issue before, because the only other CMake package in depends (hidapi) isn't built for Android targets.

tests/fuzz/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@
 # STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 # THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+if(OSSFUZZ)
```
tobtoht, line 29: Moving `miniupnpc` out of tree causes these tests to [fail to build](https://github.com/tobtoht/monero/actions/runs/24902480867/job/72923419709#step:7:1617), I'm not sure why.

The fuzz tests aren't executed in our tests CI job and are also built [downstream](https://github.com/google/oss-fuzz/blob/master/projects/monero/build.sh), so I think it's fine to skip the build on our side, which is what this line effectively does.

Issue #8818, opened 2023-04-04
Monero seems to have an IPv4-only problem.


juliaszone:
- My IPv6-only monerod node can't seem to find any IPv6 peers. It only wastes it's time on IPv4 nodes.
- The seed nodes seem to only be IPv4. Adding some IPv6 seed nodes would be important.
- When I told it to use IPv6 only (p2p-ignore-ipv4=1, rpc-ignore-ipv4=1), it still tries and binds IPv4. It would be great to be able to fully turn off IPv4.
- You must explicitly enable IPv6 (p2p-use-ipv6=1, rpc-use-ipv6=1). Ideally, IPv6 should be enabled by default.
- Node websites I've visited are either .onion or IPv4 only. I'm aware this has nothing to do with this repository, but including it in this discussion may get them to support IPv6.

I'd love to contribute to the source code, but unfortunately I don't feel comfortable in C++ (yet), so maybe starting this discussion will help.

poiuty:
Why is ipv6 disabled by default?

```
--p2p-use-ipv6 | Enable IPv6 for p2p (disabled by default).

https://monerodocs.org/interacting/monerod-reference/
```

rnhmjoj:
If not enabled by default IPv6 support is quite useless in a p2p application. 
Probably the only users that will go through the docs, find the cli option and enable it are those on an IPv6-only network, but then it won't matter because there's nobody else to peer: dual stacked users will only ever use IPv4.

kkarhan:
> Why is ipv6 disabled by default?
> 
> ```
> --p2p-use-ipv6 | Enable IPv6 for p2p (disabled by default).
> 
> https://monerodocs.org/interacting/monerod-reference/
> ```

I'm convinced this is done amidst the fears of [pseudo-]static IPv6 assignments?
A lot of people think that NAT is a security feature and the lack of said NAT + the use of EUI48 aka. MAC-Adresses to create the Interface Indentifier unless Privacy Extensions are being enabled [which basically all Desktop & mobile OSes do by default!] is seen as a privacy infringement.

Still, I think that Monero should support proper dual-stack and not care whether or not someone uses IPv4, IPv6 nor VPNs or even Tor and thus default to proper dual stack, with the option to prefer and/or disable IPv4 [or even IPv6 if there would be any legit reason to do so]...

boldsuck:
> * My IPv6-only monerod node can't seem to find any IPv6 peers. It only wastes it's time on IPv4 nodes.
> 
[...]

PR #9277, opened 2024-04-05
net: bring internal clarity and consistency between IPv4 and IPv6

bjacquin:
Modify all IPv4 variables, function arguments name and daemon arguments to IPv4 specific naming to raise consistency with IPv6. This change is done in order to make source code more legible before addressing #8818.

To ensure retro compatibility, daemon arguments and configuration settings changes are marked as deprecated, but can still be used falling back to new option name when new options are not used.

    * --p2p-bind-ip is an alias to --p2p-bind-ipv4-address
    * --p2p-bind-port is an alias to --p2p-bind-ipv4-port
    * --p2p-bind-port-ipv6 is an alias to --p2p-bind-ipv6-port
    * --rpc-bind-ip is an alias to --rpc-bind-ipv4-address
    * --rpc-restricted-bind-ip is an alias to --rpc-restricted-bind-ipv4-address

Bug: https://github.com/monero-project/monero/issues/8818


bjacquin:
Replaced with https://github.com/monero-project/monero/pull/9279

PR #9286, opened 2024-04-11
common: add va_args.h

jeffro256:
This PR is upstreaming changes in the Seraphis lib here: https://github.com/UkoeHB/monero/pull/39. This header adds a macro `VA_ARGS_COMMAPREFIX` which, when passed `__VA_ARGS__`, expands to `, __VA_ARGS__` unless the length of `__VA_ARGS__` is 0, in which case it expands to nothing. This macro is useful for passing/declaring optional function arguments.

PR #9287, opened 2024-04-11
serialization: support passing extra args to fields in DSL

jeffro256:
This PR is upstreaming changes in the Seraphis lib here: https://github.com/UkoeHB/monero/pull/39. The changes to the serialization header allow clean passing of extra arguments to field serialization in the DSL. This is used mainly to pass implied sizes of containers during deserialization to make the format more compact. For example, if my object has two containers A & B which must be the same size, I can serialize only the size of container A. Then, during deserialization, when I deserialize A, I can then use A's size to deserialize B.
    
Depends on #9286.

### DIFF
--- a/src/common/va_args.h
+++ b/src/common/va_args.h
@@ -1,0 +1,45 @@
+// Copyright (c) 2024, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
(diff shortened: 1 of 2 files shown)


src/common/va_args.h
```diff
@@ -0,0 +1,45 @@
...
+// Apdated from cpplearner's StackOverflow answer: https://stackoverflow.com/a/48045656
+#define PP_THIRD_ARG(a,b,c,...) c
+#define VA_OPT_SUPPORTED_I(...) PP_THIRD_ARG(__VA_OPT__(,),true,false,)
+#define VA_OPT_SUPPORTED VA_OPT_SUPPORTED_I(?)
+
+// VA_ARGS_COMMAPREFIX(): VA_ARGS_COMMAPREFIX(__VA_ARGS__) expands to __VA_ARGS__ with a comma in
+// front if more than one argument, else nothing.
+// If __VA_OPT__ supported, use that. Else, use GCC's ,## hack
+#if VA_OPT_SUPPORTED
+#    define VA_ARGS_COMMAPREFIX(...) __VA_OPT__(,) __VA_ARGS__
```
vtnerd, line 41: Why even have this extra mode if the compatibility mode forces a 1+ argument anyway? Basically, you are limited by what the compatibility mode can do anyway.

src/serialization/serialization.h
```diff
@@ -158,18 +169,9 @@ inline bool do_serialize(Archive &ar, bool &v)
...
-  template <bool W, template <bool> class Archive>     \
-  bool do_serialize(Archive<W> &ar, stype &v) {        \
-    ar.begin_object();                                 \
-    bool r = do_serialize_object(ar, v);               \
-    ar.end_object();                                   \
-    return r;                                          \
-  }                                                    \
-  template <bool W, template <bool> class Archive>     \
-  bool do_serialize_object(Archive<W> &ar, stype &v) { \
+#define BEGIN_SERIALIZE_OBJECT_FN(stype, ...)                                           \
```
vtnerd, line 172: After the first argument, you have to list the typename and variable name? A little funky, but it works.

vtnerd (changes requested):
Technically this additional code isn't required to achieve the array logic - ringct already does this. I suppose this makes it slightly easier to group the arrays though.

src/common/va_args.h
jeffro256: The compatibility mode *shouldn't* force a 1+ argument. `__VA_OPT__(x)` should expand to nothing if no variadic arguments are provided: https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html

vtnerd: Yes but:
```c++
VA_ARGS_COMMAPREFIX(...) , ## __VA_ARGS__
```
should have a dangling comma ?

jeffro256: On GCC-style compilers `, ## __VA_ARGS__` also expands to nothing if there are no variadic args.

src/serialization/serialization.h
jeffro256: Yup!

jeffro256:
Oops closed wrong one

luigi1111:
@jeffro256 please rebase

jeffro256:
Rebased

PR #9290, opened 2024-04-14
Fix lmdb txn commit code

vtnerd:
I added LMDB code to monero core a while ago. The only project (afaik) using it is LWS. I found this bug in the commit code - commit can fail with a resize request too.

If anyone thinks this code should be removed and put into LWS let me know shortly. I thought perhaps this could be used to replace existing LMDB code, but it looks unlikely at this point.

### DIFF
--- a/src/lmdb/database.cpp
+++ b/src/lmdb/database.cpp
@@ -179,9 +179,10 @@ namespace lmdb
     expect<void> database::commit(write_txn txn) noexcept
     {
         MONERO_PRECOND(txn != nullptr);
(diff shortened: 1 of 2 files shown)


0xFFFC0000:
LGTM  semantically and syntactically. 


I will approve this in few days if anyone not raise any issues about having this here.

vtnerd:
Embarrassing, this still has a bug in it. I will post an updated version, with an additional fix for the `commit` function.

selsta (approved):

PR #9114, opened 2024-01-05
Add support for Loongarch

fangyaling:

PR #9340, opened 2024-05-22
disable AES on loongarch64 architecture

znley:


selsta:
Did you test this on such a system? What do you think about https://github.com/monero-project/monero/pull/9340, should we merge that instead?

znley:
@selsta I think #9114 is enough. 

I Build monero on loongarch64-alpine, it works well. But tests build failure on alpine, I am still checking the reason.

I will close this merge request, I will talk about the follow-up progress on #9114. [gcc farm](https://portal.cfarm.net/machines/list/) maybe be a good platform to test.

PR #9342, opened 2024-05-28
epee: partially revert c56ee140 to fix linking errors

jeffro256:
On Linux Mint 21.3, g++ Ubuntu 11.4.0-1ubuntu1~22.04, I get linking error for an undefined reference to `epee::string_tools::trim_right`. This PR reverts the changes to epee_readline.cpp in commit c56ee140, which turns a `boost::trim_right` callsite into an `epee::string_tools::trim_right` callsite.

### DIFF
--- a/contrib/epee/include/string_tools.h
+++ b/contrib/epee/include/string_tools.h
@@ -71,8 +71,6 @@ namespace string_tools
    std::string get_current_module_path();
 #endif
   void set_module_name_and_folder(const std::string& path_to_process_);
-  void trim_left(std::string& str);
-  void trim_right(std::string& str);
   //----------------------------------------------------------------------------
   inline std::string& trim(std::string& str)
   {
--- a/contrib/epee/src/readline_buffer.cpp
+++ b/contrib/epee/src/readline_buffer.cpp
@@ -1,5 +1,4 @@
 #include "readline_buffer.h"
-#include "string_tools.h"
 #include <readline/readline.h>
 #include <readline/history.h>
 #include <iostream>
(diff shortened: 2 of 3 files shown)


jeffro256:
No clue... the linker is the same version as Ubuntu Jammy: 2.38-4ubuntu2.6

Could just be that my environment happened to expose the error, whereas it happened to work correctly by chance in the CI environments? Perhaps neither CMake variables `GNU_READLINE_FOUND` nor `DEPENDS` is set in CI?

iamamyth:
Not voicing an opinion on this revert, but I will note that the original commit had a slight omission: It replaced the boost string functions with epee in contrib/epee/src/readline_buffer.cpp, which eliminated all boost string functions from the source file, but didn't drop the `<boost/algorithm/string.hpp>` include.

selsta:
@jeffro256 It appears to be related to `debug` build. Please also open against release branch.

0xFFFC0000:
I happen to prefer this to my own way. Since we are using boost for those string manipulation. Let’s just get rid of `epee` wrapper. 

But why it gives undefined references. About that I am not sure. I have to look at the environment to find out

rbrunner7:
On Debian Bookworm current master without this PR has no error for me for a release build, but produces the mentioned link error for a debug build. With this PR cherry-picked into master both builds work.

selsta (approved):

0xFFFC0000 (approved):

selsta:
@jeffro256 The release branch PR contains a `pool_refresh` test, the master branch version does not. Please make them identical.

jeffro256:
@selsta I think this comment was meant for a different set of PRs than this, yeah?

selsta:
yes, probably wrong tab

Issue #9335, opened 2024-05-20
daemon can send duplicate transactions, causing disconnects


selsta:
Since #8916 peers that send duplicate transactions gets dropped. This usually isn't an issue, but when running a daemon with `--max-txpool-weight` it's possible that duplicate transactions can get into the fluff queue.

Quoting @Boog900 from https://github.com/monero-project/monero/issues/9317#issuecomment-2106216570

> If you receive a tx, add it to your fluff queue, drop the tx from the txpool, then receive it again, it is added to the fluff queue a second time. When the fluff timer fires we would then broadcast a message with the same tx twice causing the peer to disconnect.

Relevant code:

https://github.com/monero-project/monero/blob/c8214782fb2a769c57382a999eaf099691c836e7/src/cryptonote_protocol/levin_notify.cpp#L203-L207

PR #9353, opened 2024-06-05
cryptonote_protocol: prevent duplicate txs in fluff queue

0xFFFC0000:
Fix duplicate transaction #9335

Thanks @Boog900  for suggesting how it can be done easier.  

There is another (draft) implementation which passes tests based on `std::set` [here](https://github.com/0xFFFC0000/monero/pull/20/files).

### DIFF
--- a/src/cryptonote_protocol/levin_notify.cpp
+++ b/src/cryptonote_protocol/levin_notify.cpp
@@ -396,6 +396,8 @@ namespace levin
         for (auto& connection : connections)
         {
           std::sort(connection.first.begin(), connection.first.end()); // don't leak receive order
+          connection.first.erase(std::unique(connection.first.begin(), connection.first.end()),
+                                  connection.first.end());
           make_payload_send_txs(*zone_->p2p, std::move(connection.first), connection.second, zone_->pad_txs, true);
         }
 
--- a/tests/unit_tests/levin.cpp
+++ b/tests/unit_tests/levin.cpp
@@ -2219,6 +2219,63 @@ TEST_F(levin_notify, fluff_multiple)
     }
 }
 
+TEST_F(levin_notify, fluff_with_duplicate)
+{
+    std::shared_ptr<cryptonote::levin::notify> notifier_ptr = make_notifier(0, true, false);
+    auto &notifier = *notifier_ptr;
+
+    for (unsigned count = 0; count < 10; ++count)
+        add_connection(count % 2 == 0);
+
+    {
+        const auto status = notifier.get_status();
+        EXPECT_FALSE(status.has_noise);
+        EXPECT_FALSE(status.connections_filled);
+        EXPECT_TRUE(status.has_outgoing);
+    }
+    notifier.new_out_connection();
+    io_service_.poll();
+
+    std::vector<cryptonote::blobdata> txs(9);
+    txs[0].resize(100, 'e');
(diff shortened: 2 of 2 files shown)


src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -447,6 +447,8 @@ namespace levin
             next_flush = std::min(next_flush, context.flush_time);
             context.fluff_txs.reserve(context.fluff_txs.size() + txs.size());
             context.fluff_txs.insert(context.fluff_txs.end(), txs.begin(), txs.end());
+            context.fluff_txs.erase(std::unique(context.fluff_txs.begin(), context.fluff_txs.end()),
+                                    context.fluff_txs.end());
```
Boog900, line 451: I don't think the transactions are sorted at this point, although monerod will sort txs before fluffing, `fluff_txs` could contain txs from a previous fluff message, plus we aren't guaranteed that every peer will follow monerod's behavior.

I think this should be moved to just before sending: 

https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/cryptonote_protocol/levin_notify.cpp#L396-L400

Boog900 (changes requested):

src/cryptonote_protocol/levin_notify.cpp
vtnerd: You can see this on the previous line too - an insert is done at the end of the vector and not by sorted position.

vtnerd:
As a clarification - is this a race condition on the txpool? If the tx is already in the pool, it shouldn't re-relay it. But this should be possible if two connections fluff nearly at the same time.

Boog900:
@vtnerd This happens when the tx pool size limit is set too low and you start dropping transactions too soon after receiving them. 

When I tested this I set the tx pool limit really low, then sent 2 separate fluff messages of the same tx one after the other (not at the same time) with an extra (not duplicate) tx in the first message, then on another connection monerod sent a duplicate tx in the fluff, which would cause peers to disconnect. 

I think this is the cause of monerod struggling to keep connections with a low max txpool size.

src/cryptonote_protocol/levin_notify.cpp
0xFFFC0000: @vtnerd Yes. The reason why I insist to erase there is I want to prevent duplicates getting into the queue at all. (The sort has been forgotten from PR).

Although eventually what @Boog900 mentioned is better approach. Will update this PR.

vtnerd:
I keep forgetting about that limit. This fix (once corrected) should keep the connections alive from what I've seen (just re-reviewed the tx relay code).

src/cryptonote_protocol/levin_notify.cpp
0xFFFC0000: This issue fixed in new push, I am resolving this discussion. Please feel free to reopen it if you think it still applies.

iamamyth:
Default ipv4 isn't just the case for this project, but for most software involving networking. If anything, I'd say a more common strategy would be to keep the notion of internet address the same and expand it to include ipv6 (which also, conveniently, matches the OS reality).

Boog900 (approved):

selsta:
@0xFFFC0000 please also open against release branch

PR #9357, opened 2024-06-09
Add MacOS Arm artifact output

preland:
This PR edits the depends GitHub Actions workflow to export the macOS Arm artifact alongside the pre-existing x86_64 macOS artifact.

This is an intermediary step for creating universal macOS binaries for monero.

### DIFF
--- a/.github/workflows/depends.yml
+++ b/.github/workflows/depends.yml
@@ -105,7 +105,7 @@ jobs:
         ${{env.CCACHE_SETTINGS}}
         make depends target=${{ matrix.toolchain.host }} -j2
     - uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


selsta (approved):
I'm fine with this change but I don't see much benefit of universal binaries.

preland:
The main benefit of universal binaries is reduced platform dependency on MacOS. While x86 MacOS binaries can be ran on Arm using a translation layer, this does come with some drawbacks in performance, as well as [lack of support for newer processor instruction sets](https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment#What-Cant-Be-Translated) (although this may not be an issue for monero currently). This results in downstream Monero projects building for both x86 and Arm separately (see https://github.com/haveno-dex/haveno/issues/843).

selsta:
Right, I meant not much benefit for us in monero since we support separate x86 and ARM binaries.

Benefits

- Having a single Mac binary is easier for end user

Downsides

- Larger binary size, takes longer to download
- Difficult to integrate with our reproducible build system

it's not worth it in my opinion

PR #9363, opened 2024-06-11
get public spend and view keys from wallet rpc

woodser:


src/wallet/wallet_rpc_server.cpp
```diff
@@ -2146,6 +2146,21 @@ namespace tools
           epee::wipeable_string key = epee::to_hex::wipeable_string(m_wallet->get_account().get_keys().m_spend_secret_key);
           res.key = std::string(key.data(), key.size());
       }
+      else if(req.key_type.compare("public_view_key") == 0)
```
vtnerd, line 2149: You can compute both of these public keys from the secret keys. I guess these new choices prevent copying of secret material over RPC ?

woodser: Clients might want the public keys, but wouldn't be expected to compute them themselves. monero-wallet-rpc doesn't provide a conversion utility that I know of, hence these new calls.

milahu:
push. this is needed for [haveno](https://github.com/haveno-dex/haveno)

Issue #9308, opened 2024-04-28
The Monero core software wallet API will (probably) switch from 'wallet2.h' to 'wallet2_api.h'


rbrunner7:
## The Short Version

The devs from the [Seraphis wallet workgroup](https://github.com/seraphis-migration/strategy/wiki/Seraphis-Wallet-Workgroup) currently build a new wallet implementation for the Monero core software. It will supersede `wallet2.h` plus `wallet2.cpp` which won't be available anymore for programs like the Monero CLI wallet or the Monero RCP server, and neither for any third-party wallet apps like Cake Wallet, Monerujo and so on. The API for the new wallet will be [wallet2_api.h](https://github.com/monero-project/monero/blob/master/src/wallet/api/wallet2_api.h) plus a number of related header files from the [/source/wallet/api](https://github.com/monero-project/monero/tree/master/src/wallet/api) folder of the Monero source code. All things that are now only in the `wallet2.h` "API" will be added to that, to cover 100% of the necessary functionality, and to make getting fully rid of `wallet2.h` possible in the first place.

## The Long Version

Right now as I write this, in Spring 2024, it's not fully clear how the path of the Monero software into the future will look exactly. Most probably it will use FCMPs, but apart from that there are several possible directions to take. What is clear however is that the wallet implementation of the core Monero software, with its massive 15,000 lines of complex code in a single source file called [wallet2.cpp](https://github.com/monero-project/monero/blob/master/src/wallet/wallet2.cpp), is definitely ill suited for larger modifications to accommodate transactions that work rather differently.

We, i.e. the group of Monero devs that are engaged in the [Seraphis wallet workgroup](https://github.com/seraphis-migration/strategy/wiki/Seraphis-Wallet-Workgroup), are convinced that the time has finally come to retire `wallet2` and replace it with a full rewrite. Actually, work on this has started quite a while ago and will hopefully soon pick up speed considerably: We are already on the way.

One question left unanswered so far for quite a while already was this: How will the *API* of this new wallet look? Which C++ methods, structs, enums and so on will it offer, for wallet app writers to use?

Personally I once had a dream of a brand-new such wallet API, designed out on a green field, making things possible like keeping several wallets open at the same time seamlessly and syncing them in parallel, unify wallet API and RPC server API so that the only difference between the two would be that for the first you make direct C++ method calls and remote "calls" for the second, and maybe offering a clean superset API for multisig wallets instead of cramming them into a single API as well.

[...]

PR #9368, opened 2024-06-14
API organize functions

SNeedlewoods:
This is a small first step towards the new API: https://github.com/monero-project/monero/issues/9308


selsta:
Can you add a bit more information what you are changing here? The diff is huge.

SNeedlewoods:
It was discussed in #no-wallet-left-behind. If you look at these three links for example, you see the functions are out of order:
https://github.com/monero-project/monero/blob/master/src/wallet/api/wallet2_api.h#L142
https://github.com/monero-project/monero/blob/master/src/wallet/api/unsigned_transaction.h#L44
https://github.com/monero-project/monero/blob/master/src/wallet/api/unsigned_transaction.cpp

The feedback was:

>I think reorganizing the API to shuffle function order in this case would be fine, as long as that's the only changes in that PR 

>I agree [...] that probably the most easiest-to-review and least controversial way to do that is indeed making a PR, or making several PRs, that do absolutely nothing else than just reordering.

That's what this PR is about.. Hope that answers your question?

rbrunner7:
If things go according to plan, this PR is the first in a row of commits that deal with the *Wallet API* in the `/src/wallet/api` folder of the Monero codebase. As @SNeedlewoods already mentioned, the plan is outlined in issue [#9308](https://github.com/monero-project/monero/issues/9308), also published on Reddit [here](https://old.reddit.com/r/Monero/comments/1cf8oom/the_monero_core_software_wallet_api_will_probably/).

In short: `wallet2` will be removed completely from the Monero codebase. CLI wallet, RPC server, GUI wallet, third-party wallet apps etc. will all use the Wallet API.

This course of action was discussed e.g. [here](https://github.com/monero-project/meta/issues/993) in a meeting of the Seraphis Wallet workgroup.

When looking closer at the code of the Wallet API @Sneedlewoods noticed that things in .h and .cpp file are partially out of order, and private method implementations can be found between public method implementation. See their comment with a good example.

We think that this is unfortunate and can get in the way if we go on and significantly enlarge the API, and later also do an extensive rewrite of all wallet code that is now in `wallet2.cpp`. It may also be a nuisance for people like app wallet writers studying the code. Consider that the Wallet API is already pretty important today, but will of course become even more important if `wallet2` goes away.

Thus the plan to make this PR was born: We reorder in a very first step, but take care to make it no more painful than absolutely necessary.

Thus this PR makes **no other changes** than reordering declarations and implementations and removing a few stray comment lines that are of no use, and **exclusively** deals with Wallet API files.

Thus anything that violates these rules should be seen as an error in the PR that needs correcting.

0xFFFC0000:
Thanks for the PR. The hardest part is reviewing all these files and making sure no unintended code changes has been happened. 

One thing I will try is to compile modules in both case, and the object file should be same if no code changes has happened. Though the changes is spread to many files, will make it harder. 

If anyone has a better idea for reviewing this, I am all ears.

src/wallet/api/wallet.cpp
```diff
@@ -2058,544 +1986,611 @@ bool WalletImpl::verifySignedMessage(const std::string &message, const std::stri
...
-    if (!ret)
-    {
-        setStatusError(tr("Failed to get rings"));
-        return false;
+    clearStatus();
+
+    bool ready = false;
+    if (!m_wallet->multisig(&ready) || !ready) {
+        m_status = Status_Error;
+        m_errorString = tr("The wallet must be in multisig ready state");
```
0xFFFC0000, line 2398: There is a semantical change here. The original form is here [1].  This new style condition is not deterministically the same as the original form.

1. https://github.com/monero-project/monero/pull/9368/files#diff-fcae134993cca2b6f1e887a60638f226ab00165e30bd4d326c17f0a58e11871fL2068

0xFFFC0000:
My apologies for being a bummer. But we have to make sure that no semantical changes are inside this PR as @rbrunner7 mentioned. 

The size of the PR is huge. Maybe we can break into many PRs. First only and only whitespace, then some semantical changes if necessary [1].

1. https://github.com/monero-project/monero/pull/9368#discussion_r1644012418

src/wallet/api/wallet.cpp
SNeedlewoods: Thanks for taking a look and good catch. This is not intentional and probably snuck in when rebased. Change was pretty recent https://github.com/monero-project/monero/commit/3a47cdad13fd8486b7d5559ec004ff49b3b3aa6d

Regarding your comment below, this PR is already supposed to be the broken down version of a bigger/messier PR, so if you find anything that's not "ordering functions", it does not belong, but creeped in. Except removing some comments which would get removed with the next PR, because I thought having two PRs (first move comment, second remove comment) will also be unnecessary reviewers work.

src/wallet/api/wallet.cpp
rbrunner7, line 1: In this file also some whitespace changes complicate the mechanical comparison with a tool somewhat, but again that's ok, and removed trailing blanks progress, IMHO.

Otherwise no more logic changes, after the one detected by @0xFFFC0000 is corrected now.

rbrunner7 (commented):
This must be the dumbest review I ever did :)

But everything is ok, I wasn't able to unearth a single true code change. For a small number of questionable other changes I left comments. After those are addressed I am ready to approve.

src/wallet/api/wallet_manager.cpp
```diff
@@ -347,6 +347,12 @@ std::string WalletManagerImpl::resolveOpenAlias(const std::string &address, bool
     return addresses.front();
 }
 
+bool WalletManagerImpl::setProxy(const std::string &address)
+{
+    return m_http_client.set_proxy(address);
+}
+
+// Static
```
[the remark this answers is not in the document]
SNeedlewoods: You're right, it's new and shouldn't be here. Will ask for general opinions on such "block separator" comments in today's meeting.

rbrunner7 (approved):

rbrunner7:
In my understanding the Wallet API overhaul and completing PR #9464 assumes that this PR here gets merged first, as a starting point. If I am correct, and as #9464 is slowly but steadily reaching completion, the question arises how to proceed with this PR here.

It seems to just sleep for months already.

I suspect the question is whether it is acceptable on fundamental grounds: Does the Monero project accept a PR that does not add or improve any functionality, and does not make any "true" code changes, but just arranges things in a better and more logical way?

I would like to re-open discussion and try to achieve that either this PR gets accepted and merged soon, or else gets "officially" rejected and closed. A third option of just letting it linger does not seem to make sense to me.

So, any opinions?

SNeedlewoods:
The Wallet API PR is actually not based on this. I have another local branch for adding comments which is based on this.
The idea was: if this gets merged, I'd make a PR for the comments and if that gets merged I rebase the Wallet API PR on that one.

So closing this PR would be less work for me.

After spending much time with the current form, I kinda accustomed to it, but if anyone values clean code (and thinks this is a little closer to achieving it), I'm still willing to work it out.

PR #9370, opened 2024-06-18
Fix ZMQ Tx Pruning

vtnerd:
When requesting pruned txes over ZMQ, the txes aren't fully pruned. This is due to a quirk in the unpacking of txes - the pruned elements are still read into the local tx. Fixing this is a bit more involved. The ZMQ output code then has no context/flag for `prune` in the signature portion (because it assumes the signatures wouldn't have been unpacked in. a pruned tx).

This is a simple hack that passes the prune information from the base tx into the next function call. This allows for full pruning, which saves on bandwidth, output time, and read time.

I have verified locally (with LWS) that this output is still valid.

### DIFF
--- a/src/serialization/json_object.cpp
+++ b/src/serialization/json_object.cpp
@@ -273,7 +273,10 @@ void toJsonValue(rapidjson::Writer<epee::byte_stream>& dest, const cryptonote::t
   {
     INSERT_INTO_JSON_OBJECT(dest, signatures, tx.signatures);
   }
-  INSERT_INTO_JSON_OBJECT(dest, ringct, tx.rct_signatures);
+  {
+    dest.Key("ringct");
+    toJsonValue(dest, tx.rct_signatures, tx.pruned);
+  }
 
   dest.EndObject();
 }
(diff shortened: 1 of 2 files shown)


src/serialization/json_object.cpp
```diff
@@ -1111,7 +1112,7 @@ void fromJsonValue(const rapidjson::Value& val, cryptonote::rpc::BlockHeaderResp
   GET_FROM_JSON_OBJECT(val, response.reward, reward);
 }
 
-void toJsonValue(rapidjson::Writer<epee::byte_stream>& dest, const rct::rctSig& sig)
+void toJsonValue(rapidjson::Writer<epee::byte_stream>& dest, const rct_sig sig)
```
iamamyth, line 1115: If you introduce locals for the actual (`rct::rctSig`) `sig` and `pruned` values, and change the name of the param here to reflect that it's now a wrapper struct, the diff would be a lot smaller, and better reflect the nature of the change, as well as the underlying procedure (there's only one `sig.sig`, and it's the heart of the procedure, so it's semantically odd to continually use indirection).

vtnerd: That's true, I can move all of the signature output code into the parent function (is that you are saying). The diff size would be the same, because I had to rename everything in the sub-function anyway.

vtnerd: Oh never mind, I see what you mean. Yeah, I can call the code directly instead of using `INSERT_INTO_JSON` ?

vtnerd: FWIW I partially did it this way because my new serialization routines will basically require it be done like this.

vtnerd:
Force pushed a much smaller changeset.

iamamyth (approved):

PR #9376, opened 2024-06-21
tx_pool: update internal data structure to boost::bimap.

0xFFFC0000:
startup speed up 2-5x because this PR changing `O(n^2)` operation to `O(nlogn)`.

Almost all people involved in stressnet experiment who tested this patch, reported at least 2.5x speed up of startup time of monerod.

Thanks to @Boog900, @[spackle-xmr](https://github.com/spackle-xmr/), [Rucknium](https://github.com/Rucknium) and [ofrnxmr](https://github.com/nahuhh) for their help

### DIFF
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -29,6 +29,7 @@
 // Parts of this file are originally copyright (c) 2012-2013 The Cryptonote developers
 
 #include <algorithm>
+#include <boost/bimap/support/iterator_type_by.hpp>
 #include <boost/filesystem.hpp>
 #include <unordered_set>
 #include <vector>
@@ -42,6 +43,7 @@
 #include "blockchain_db/blockchain_db.h"
 #include "int-util.h"
 #include "misc_language.h"
+#include "misc_log_ex.h"
 #include "warnings.h"
 #include "common/perf_timer.h"
 #include "crypto/hash.h"
@@ -464,7 +466,7 @@ namespace cryptonote
         break;
       try
       {
-        const crypto::hash &txid = it->second;
+        const crypto::hash &txid = it->get_right();
         txpool_tx_meta_t meta;
         if (!m_blockchain.get_txpool_tx_meta(txid, meta))
         {
@@ -491,11 +493,11 @@ namespace cryptonote
           return;
         }
         // remove first, in case this throws, so key images aren't removed
-        MINFO("Pruning tx " << txid << " from txpool: weight: " << meta.weight << ", fee/byte: " << it->first.first);
+        MINFO("Pruning tx " << txid << " from txpool: weight: " << meta.weight << ", fee/byte: " << it->get_left().first);
         m_blockchain.remove_txpool_tx(txid);
         reduce_txpool_weight(meta.weight);
         remove_transaction_keyimages(tx, txid);
-        MINFO("Pruned tx " << txid << " from txpool: weight: " << meta.weight << ", fee/byte: " << it->first.first);
+        MINFO("Pruned tx " << txid << " from txpool: weight: " << meta.weight << ", fee/byte: " << it->get_left().first);
 
         auto it_prev = it;
         --it_prev;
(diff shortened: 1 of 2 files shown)


montero-weblate:
are external factors required to test this? (e.g. a huge mempool) 
when is monerod 'started'? would this be sufficient:
`time ./monerod --detach`

0xFFFC0000:
Regarding the failure in PR, it is race failure we have, unrelated to this PR. 

```
node_server.race_condition
```

src/cryptonote_core/tx_pool.h
```diff
@@ -62,24 +65,33 @@ namespace cryptonote
...
+    bool operator()(const crypto::hash& a, const crypto::hash& b) const
+    {
+      return memcmp(a.data, b.data, sizeof(crypto::hash)) < 0;
     }
   };
 
   //! container for sorting transactions by fee per unit size
-  typedef std::set<tx_by_fee_and_receive_time_entry, txCompare> sorted_tx_container;
+  typedef boost::bimap<boost::bimaps::multiset_of<std::pair<double, std::time_t>, txFeeCompare>,
+                       boost::bimaps::set_of<crypto::hash, hashCompare>> sorted_tx_container;
```
vtnerd, line 93: What about using `unordered_set_of` here? This could accelerate the lookup, but would likely require more memory (for the table), and perhaps resizing the table would be slow.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -751,13 +753,9 @@ namespace cryptonote
     m_remove_stuck_tx_interval.do_call([this](){return remove_stuck_transactions();});
   }
   //---------------------------------------------------------------------------------
-  sorted_tx_container::iterator tx_memory_pool::find_tx_in_sorted_container(const crypto::hash& id) const
+  sorted_tx_container::iterator tx_memory_pool::find_tx_in_sorted_container(const crypto::hash& id)
```
vtnerd, line 756: I looked at the documentation, and it looks like this function can remain `const`? What broke this?

vtnerd (commented):
Overall looks good, just a few questions.

src/cryptonote_core/tx_pool.cpp
vtnerd: I just noticed that `iterator` has to be returned, so presumably the reason is `find` and `project_up` return `const_iterator` here.

0xFFFC0000: Yes you are right. If it remains `const` then the `find` will return `const_iterator` and that is causing a lot headache. We want to use that iterator in `erase` which doesn't accept `const_iterator`, and no easy way of turning `const_iterator` to `iterator` in boost.

src/cryptonote_core/tx_pool.h
0xFFFC0000: What do you think? At this point it comes to the design choices, I am perfectly happy to change it to `unordered_set_of`.

selsta:
@vtnerd could you approve if all your questions are answered?

src/cryptonote_core/tx_pool.h
vtnerd: Search times are a bit lower with an `unordered_map`, but insertion times may be slower with a large number of keys. I think we have more searches than insertions, but I am not certain. So just leave it for now, unless. you want to run another test.

vtnerd (approved):

src/cryptonote_core/tx_pool.h
0xFFFC0000: Thanks a lot for detailed explanation.

src/cryptonote_core/tx_pool.h
```diff
@@ -62,24 +65,33 @@ namespace cryptonote
   //! pair of <transaction fee, transaction hash> for organization
   typedef std::pair<std::pair<double, std::time_t>, crypto::hash> tx_by_fee_and_receive_time_entry;
```
rbrunner7, line 66: Pretty sure that typedef is not needed anymore and can be deleted, in the interest of clean code?

rbrunner7 (approved):
Looks ok to me.

In particular, as the code is already shown as working, I concentrated on an attempt to find possible unfortunate side effects of the change of storing that info, and found none: Things are simple enough.

My only comment is a proposal to delete a `typedef` that is probably not needed anymore - nothing to prevent me from approving already.

src/cryptonote_core/tx_pool.h
0xFFFC0000: Thanks for your comment. Yes. it was not needed. But didn't want to make any unnecessary changes.

PR #9379, opened 2024-06-24
wallet2: move get_rpc_status func to RPC server definition file

j-berman:
Supports consistent RPC status handling between wallet2 and a new wallet lib

### DIFF
--- a/src/rpc/core_rpc_server_commands_defs.h
+++ b/src/rpc/core_rpc_server_commands_defs.h
@@ -80,6 +80,19 @@ namespace cryptonote
 #define CORE_RPC_STATUS_NOT_MINING "NOT MINING"
 #define CORE_RPC_STATUS_PAYMENT_REQUIRED "PAYMENT REQUIRED"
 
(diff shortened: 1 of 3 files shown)


src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -80,6 +80,19 @@ namespace cryptonote
 #define CORE_RPC_STATUS_NOT_MINING "NOT MINING"
 #define CORE_RPC_STATUS_PAYMENT_REQUIRED "PAYMENT REQUIRED"
 
+static inline const std::string get_rpc_status(const bool trusted_daemon, const std::string &s)
```
vtnerd, line 83: If you drop the `static` keyword, the linker will merge multiple potential copies in the runtime binary. Otherwise, each compilation unit will have its own copy, and I don't see a good reason for this.

vtnerd (changes requested):

vtnerd (approved):

PR #9381, opened 2024-06-24
Daemon RPC: add max_block_count field to /getblocks.bin

j-berman:
Enables making parallel requests for a small number of blocks. When scanning the chain using parallel requests, setting this request param to a small number (e.g. 10) can speed up progress feedback to the user significantly (compared to leaving the default to 1000 as is).

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -725,12 +725,13 @@ namespace cryptonote
         }
       }
 
-      size_t max_blocks = COMMAND_RPC_GET_BLOCKS_FAST_MAX_BLOCK_COUNT;
+      size_t max_blocks = req.max_block_count > 0
+        ? std::min(req.max_block_count, (uint64_t)COMMAND_RPC_GET_BLOCKS_FAST_MAX_BLOCK_COUNT)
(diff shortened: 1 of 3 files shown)


0xFFFC0000 (approved):
LGTM.

iamamyth (approved):
Looks fine, needs a rebase to remove the merge conflict.

j-berman:
The conflict was from bumping the RPC minor version. Release branch current RPC minor version is [14](https://github.com/monero-project/monero/blob/58a1d54a4f90a235616ede3f8ebae2b8dca41589/src/rpc/core_rpc_server_commands_defs.h#L91). Master branch is [16](https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/src/rpc/core_rpc_server_commands_defs.h#L104). Resolved conflict by leaving it untouched at 16.

rbrunner7 (approved):
I assume you don't go higher with minor RPC version because master does not "go into production", and nobody has a real need to be able to react to different master versions there - which sounds reasonable to me.

The changes look ok to me.

0xFFFC0000 (approved):

PR #9383, opened 2024-06-24
Daemon RPC: /getblocks.bin return chain's top_block_hash in resp

j-berman:
Particularly useful in combination with https://github.com/monero-project/monero/pull/9382, so that the client can be certain it's on the correct chain even when the request to /getblocks.bin includes a `start_height > chain tip`

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -2765,7 +2765,7 @@ bool Blockchain::find_blockchain_supplement(const std::list<crypto::hash>& qbloc
 // find split point between ours and foreign blockchain (or start at
 // blockchain height <req_start_block>), and return up to max_count FULL
 // blocks by reference.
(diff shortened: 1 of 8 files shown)


src/cryptonote_core/blockchain.cpp
```diff
@@ -2774,7 +2774,10 @@ bool Blockchain::find_blockchain_supplement(const uint64_t req_start_block, cons
   if(req_start_block > 0)
   {
     // if requested height is higher than our chain, return false -- we can't help
-    if (req_start_block >= m_db->height())
+    uint64_t top_height;
```
vtnerd, line 2777: Nitpick, but you really don't need the temporary `top_height` here or below. Just give `total_height` to `top_block_hash` then `++total_height`.

vtnerd (approved):

vtnerd (approved):

PR #3897, opened 2018-06-01
mlog: --max-log-files to set the max number of rotated log files

stoffu:

PR #9392, opened 2024-07-07
epee: fix mlog filename compare bug.

0xFFFC0000:
When using a relative path for the log filename,
since the iteration on files adds "./" to the beginning of the filename monero-wallet-rpc and monero-wallet-cli	cannot find already written log files and therefore rotate indefinitely  (causing unlimited storage usage).

Thanks to [@ofrnxmr](https://github.com/nahuhh) for reporting this bug to me.


nahuhh:
original: https://github.com/monero-project/monero/pull/3897

@vtnerd you reviewed the original. Can you check this one as well? Thanks

vtnerd (approved):

PR #9367, opened 2024-06-13
external: update miniupnpc 2.2.8

0xFFFC0000:
Fix and improve existing CMake script.

Fixes: https://github.com/monero-project/monero/issues/9359

Related: https://github.com/monero-project/monero/pull/9366

PR #9402, opened 2024-07-17
upnp: add support for miniupnpc API version 18

bjacquin:
miniupnpc 2.2.8 updated API to version 18, with breaking change to UPNP_GetValidIGD().

See: https://github.com/miniupnp/miniupnp/commit/c0a50ce33e3b99ce8a96fd43049bb5b53ffac62f
See: http://miniupnp.free.fr/files/changelog.php?file=miniupnpc-2.2.8.tar.gz


selsta:
There are similar changes currently in https://github.com/monero-project/monero/pull/9367

PR #9404, opened 2024-07-22
cryptonote_core: only verify txpool when the hardfork value has changed.

0xFFFC0000:
Detailed discussion [here](https://github.com/spackle-xmr/monero/pull/13/#issuecomment-2234543784).

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -31,6 +31,7 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/uuid/nil_generator.hpp>
 
+#include "misc_log_ex.h"
 #include "string_tools.h"
 using namespace epee;
 
@@ -679,9 +680,21 @@ namespace cryptonote
     r = m_mempool.init(max_txpool_weight, m_nettype == FAKECHAIN);
     CHECK_AND_ASSERT_MES(r, false, "Failed to initialize memory pool");
 
+    uint64_t session_start_height = m_blockchain_storage.get_current_blockchain_height() - 1; // use top block's height
+    uint8_t hardfork_version_start = m_blockchain_storage.get_hard_fork_version(session_start_height);
+    uint8_t hardfork_version_current = m_blockchain_storage.get_current_hard_fork_version();
+    MDEBUG("Current height: " << session_start_height);
+    MDEBUG("Start hardfork version: " << (int) hardfork_version_start);
+    MDEBUG("Current hardfork version: " << (int) hardfork_version_current);
+    
     // now that we have a valid m_blockchain_storage, we can clean out any
     // transactions in the pool that do not conform to the current fork
-    m_mempool.validate(m_blockchain_storage.get_current_hard_fork_version());
+    if(hardfork_version_start != hardfork_version_current) {
+      m_mempool.validate(hardfork_version_current);
+    }
+    else {
+      MDEBUG("Not validating the txpool, no hardfork detected.");
+    }
 
     bool show_time_stats = command_line::get_arg(vm, arg_show_time_stats) != 0;
     m_blockchain_storage.set_show_time_stats(show_time_stats);


src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -679,9 +680,21 @@ namespace cryptonote
     r = m_mempool.init(max_txpool_weight, m_nettype == FAKECHAIN);
     CHECK_AND_ASSERT_MES(r, false, "Failed to initialize memory pool");
 
+    uint64_t session_start_height = m_blockchain_storage.get_current_blockchain_height();
+    uint8_t hardfork_version_start = m_blockchain_storage.get_hard_fork_version(session_start_height);
+    uint8_t hardfork_version_current = m_blockchain_storage.get_current_hard_fork_version();
```
SChernykh, line 685: Will these two values ever be different at this point? This is `core::init`, the node hasn't started syncing at this point and is only loading existing blocks that it synced during last run.

SChernykh: The discussion mentions that it can change if the node itself is updated in between runs, but it needs to be tested. So far it's not obvious that these two versions will be different even in this case.

0xFFFC0000: `heights` contains the HFs and `current_for_index` is the index of our current HF in the list. 

The case we want to verify is when we shutdown at boundary, and start up in next hf.

0xFFFC0000:
> Second question, what about 3rd-party tools changing the database between monerod runs? By introducing this code, you assume that nothing ever touches the txpool table except monerod.

IMHO that is an edge case. Where we can have a flag like `--verify-txpool` and user would verify its txpool. Average user doesn't need to verify txpool each time during monerod start up. 

I might be wrong. Please don't hesitate to let me know.

src/cryptonote_core/cryptonote_core.cpp
SChernykh: What if someone shuts down a few heights above the boundary (they were late to the fork) and restarts with the new version?

Boog900: They would be on an invalid chain, I have thought about this and I don't think there is a good solution to this while we support voting for hardforks.

You would need to check all past HFs, or at least the ones outside the checkpoints and pop blocks if any blocks have incorrect major versions.

This is a separate issue though IMO

Boog900:
> Second question, what about 3rd-party tools changing the database between monerod runs? By introducing this code, you assume that nothing ever touches the txpool table except monerod.

If tools are touching the DB it is up to them to make sure they are doing it correctly, there are many ways to currently mess up the blockchain by messing with the database. 

There are also checks currently ran on the way to the txpool but not in `txpool::validate`, so we are currently relying on 3rd-party tools adding txs to the pool to not violate those rules.

src/cryptonote_core/cryptonote_core.cpp
Boog900: @SChernykh I was wrong, we do handle this here:

https://github.com/monero-project/monero/blob/caa62bc9ea1c5f2ffe3ffa440ad230e1de509bfd/src/cryptonote_core/blockchain.cpp#L391-L429

The way we handle this though pretty much removes the possibility of using the voting for hardforks feature. We pop blocks until the minimum possible height of the HF, not the height it should activate according to votes, so even if the fork isn't active we will pop blocks.

sneurlax (approved):
These changes LGTM, but I did not test by simulating a hardfork.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -679,9 +680,21 @@ namespace cryptonote
     r = m_mempool.init(max_txpool_weight, m_nettype == FAKECHAIN);
     CHECK_AND_ASSERT_MES(r, false, "Failed to initialize memory pool");
 
+    uint64_t session_start_height = m_blockchain_storage.get_current_blockchain_height();
+    uint8_t hardfork_version_start = m_blockchain_storage.get_hard_fork_version(session_start_height);
```
Boog900, line 684: `session_start_height` seems to be using the chain height, it needs to use the top block's height which is one less.

Chain height is always one more than the top block's height.

Boog900 (approved):

Boog900:
This code is only ran at startup but yes when adding a block and the HF changes the tx-pool is re-validated: 

Which would freeze up the node for a while with a big tx-pool. Cuprate caches when a tx was validated with the tx in the pool: https://github.com/Cuprate/cuprate/blob/main/types/src/transaction_verification_data.rs#L35 so we don't need to do tx-pool re-validation at start up/when a HF changes, we only re-validate if the current blockchain state requires it compared to the cached verification. 

monerod could probably do something similar, just caching the HF and doing `ver_non_input_consensus` added in #9135 if the HF has changed when adding the tx, skipping the need for re-validation of the txpool.

nahuhh:
@selsta it takes ~1hr to revalidate 100mb, so.. i'd say "no, it would not go over well".

it wasnt tested during hf of stressnet, but simply starting up your node took 90mins due to the revalidation of the txpool. Node is completely unresponsive during the validation.

pop_blocks has similar behavior, revalidating the tx while popping while node is unresponsive

SChernykh (approved):

PR #9406, opened 2024-07-26
src: add net type and daemon address to wallet rpc

D3vN1:


selsta (commented):
Please bump WALLET_RPC_VERSION_MINOR

D3vN1:
@selsta I've increased  WALLET_RPC_VERSION_MINOR from 27 to 28

0xFFFC0000:
Please squash two commits. And push it as a same commit.

D3vN1:
I have successfully squashed the two commits and pushed them as a single commit, as per your request.
Please let me know if there are any further changes needed.
Thanks

hyc:
Why do clients need to know this info?

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -447,7 +447,34 @@ namespace wallet_rpc
...
+      BEGIN_KV_SERIALIZE_MAP()
+      END_KV_SERIALIZE_MAP()
+    };
+    typedef epee::misc_utils::struct_init<request_t> request;
 
+      struct response_t
+      {
+        std::string  daemon_address;
+        std::string nettype;
+        bool mainnet;
```
iamamyth, line 464: It rarely makes sense to include outputs of pure functions of response data in a response; exceptions can be made in case of complicated functions, but these boolean values are all trivially derived from the nettype (as can be seen by the code which sets them).

0xFFFC0000:
> Why do clients need to know this info?

I believe it was a request from @woodser.

src/wallet/wallet_rpc_server_commands_defs.h
0xFFFC0000: I think `nettype` itself is enough. 

@iamamyth is that what you meant too? My apologies for asking, just wanted to clarify.

iamamyth: > I think nettype itself is enough

Yes, that's what I meant.

woodser:
> > Why do clients need to know this info?
> 
> I believe it was a request from @woodser.

It's been a couple years and now I don't see a critical case to provide this information, since the client knows the network type on startup and the daemon's url on startup or setting the connection. Maybe there's a use case I've forgotten, but the client ought to know this info.

PR #9416, opened 2024-08-01
rpc: add cumul weight field to block template resp

### LINKED ISSUE #9415: Expose Expected Block Size to get_block_template RPC
With ``get_block_template`` RPC call we should able get expected block sizes like the ``expected_reward``.

Currently its internally calculated but its not exposed to RPC 

I believe its needed because it would be very useful blockchain explorers (for an example:  https://github.com/txstreet/processor/issues/2)

Related code: https://github.com/monero-project/monero/blob/master/src/cryptonote_core/blockchain.h

jeffro256:
Resolves #9415

### DIFF
--- a/src/cryptonote_basic/miner.cpp
+++ b/src/cryptonote_basic/miner.cpp
@@ -171,9 +171,10 @@ namespace cryptonote
       extra_nonce = m_extra_messages[m_config.current_extra_message_index];
     }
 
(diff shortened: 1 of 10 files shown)


sneurlax (approved):
Reviewed changes, they look good to me.  Built branch, tested response, it works.

0xFFFC0000 (approved):
LGTM

PR #9420, opened 2024-08-04
set USE_UNWIND for FreeBSD

aperechnev:
The source code can't be compiled on FreeBSD without using `libunwind` which is available there.


aperechnev:
Sorry IDK how these changed got into this pull request. I will send a new one.

PR #9436, opened 2024-08-14
[WIP] full-chain membership proof++ integration

j-berman:
This is a WIP draft PR for the full-chain membership proof (fcmp++) integration. It's roughly following section 6 of the specification written by @kayabaNerve ([paper](https://github.com/kayabaNerve/fcmp-ringct/blob/develop/fcmp%2B%2B.pdf), [commit](https://github.com/kayabaNerve/fcmp-ringct/blob/b4a8cdae770770de1e96fa1c97c1e6e0926f8409/fcmp%2B%2B.pdf)).

Checklist of items expected in this PR:

- [x] A. [Rust FFI](#a.-rust-ffi)
- [-] B. [Curve trees merkle tree](#b.-curve-trees-merkle-tree)
  - [x] a. Preparing locked outputs for insertion to the tree upon unlock 
  - [x] b. `grow_tree` algorithm
  - [x] c. `trim_tree` algorithm
  - [x] d. LMDB changes
  - [x] e. Growing the tree as the node syncs
  - [x] f. Migrating cryptonote outputs into the tree
  - [ ] g. Trimming the tree on reorgs and popping blocks
  - [ ] h. Migrating key images
- [x] C. [Transaction struct changes for fcmp++](#c.-transaction-struct-changes-for-fcmp)
- [ ] D. [Constructing fcmp++ transactions](#d.-constructing-fcmp-transactions)
- [ ] E. [Verifying fcmp++ transactions](#e.-verifying-fcmp-transactions)
- [ ] F. [Consensus changes for fcmp++](#f.-consensus-changes-for-fcmp)

The above checklist does **not** include all items required to complete the integration.

I plan to divide the code into commits where each subsequent commit builds off the prior commit. I could eventually close this PR in favor of smaller PR's that can be reviewed sequentially and in isolation.

This PR description can function as living documentation for the code as work on the integration progresses (and audits/fcmp++ research progress in parallel) . In this description, I highlight the most critical components from the code, aiming to make the PR as a whole easier to understand. Thoughts/feedback is welcome at any time.

# A. Rust FFI

Since much of the full-chain membership proof++ code is written in Rust, this PR implements a Foreign Function Interface ([FFI](https://doc.rust-lang.org/nomicon/ffi.html)) to call the Rust code from C++. Using cmake, the Rust code is compiled into a static lib (`libfcmp_pp_rust.a`) when you run `make` from the root of the monero repo. The static lib's functions are exposed via the C++ `src/fcmp_pp/fcmp++.h` header file (generated with the help of [cbindgen](https://github.com/mozilla/cbindgen) and modified slightly). The heavy lifting on the Rust side is done in @kayabaNerve's [`full-chain-membership-proofs` Rust crate](https://github.com/kayabaNerve/fcmp-plus-plus/tree/develop/crypto/fcmps); the Rust handles the math on the Helios and Selene curves, and fcmp++ construction and verification.

Here is what the structure looks like at time of writing:

rust_ffi_diagram

- Major credit to @kayabaNerve for solid work getting cross-compilation and compilation on a number of architectures working.
- I'm generally seeking to isolate all Rust types and minimize usage of the Rust lib to what's necessary for fcmp++'s in this PR.
  - The C++ has clear responsibilities: manage tree state, update/read the db, ed25519 math, tx changes, RPC routes.
  - The Rust: math on Helios and Selene curves, fcmp++ construction, and fcmp++ verification.
  - I think this strategy has so far been successful in avoiding spaghetti of mixed C++ and Rust across the code. I've spent the vast majority of my time working on C++ algorithms, not dealing with the Rust. The heavy majority of code in this PR is C++.

# B. Curve trees merkle tree

The curve trees merkle tree is a new store for **spendable** transaction outputs in the chain. fcmp++'s work by proving you own (and can spend) an output in the tree, without revealing which output is yours. All existing **valid** cryptonote outputs will be inserted into the tree **as soon as the outputs unlock**. Once an output is in the tree, users can construct fcmp++'s with that output. Thus, the anon set will roughly be the entire chain since genesis.

The leaves in the tree are composed of output tuples `{O.x, I.x, C.x}`, and each layer after the leaf layer is composed of hashes of chunks of the preceding layer, as follows:

```
leaves:  [O.x,I.x,C.x,O.x,I.x,C.x,...]
layer 0: [Hash(first chunk of leaves), Hash(next chunk of leaves),...]
layer 1: [Hash(first chunk of layer 0 hashes), Hash(next chunk of layer 0 hashes)...]
...
```

Each layer is composed of points alternating on two curves (@tevador's proposed [Selene and Helios curves](https://gist.github.com/tevador/4524c2092178df08996487d4e272b096)).  The leaves are Selene scalars (we convert ed25519 points to Selene scalars), the layer after leaves is composed of points on the Selene curve (we hash chunks of Selene scalars from the leaf layer to get this layer's Selene points), the following layer is composed of points on the Helios curve (we convert the prior layer's Selene points to Helios scalars, and hash chunks of those Helios scalars to get this layer's Helios points), the following layer is composed of points on the Selene curve (we convert the prior layer's Helios points to Selene scalars, and hash chunks of those Selene scalars to get this layer's Selene points), and so on. We continue until there is just one chunk in a layer to hash, leaving us with the tree root.

Each curve has a defined chunk width used when hashing the children in the preceding layer. The final layer has a single element in it: the root.

There are 3 critical steps to growing the tree:

1. Keep track of the block in which outputs will unlock.
2. Grow the tree with outputs in the block in which they unlock.
3. Update the database.

## a. Curve trees merkle tree: Preparing locked outputs for insertion to the tree upon unlock 

We first need to determine the block in which outputs unlock. We keep track of locked outputs by unlock block in the database so we can grow the tree in the block they unlock.

Take note of the function: `get_outs_by_unlock_block`. Upon adding a block, we iterate over all the block's tx outputs **in order**, and place the outputs in the container`OutputsByUnlockBlock = std::unordered_map<uint64_t, std::vector<OutputContext>>`. The `uint64_t` is the output's unlock height. The output's unlock height is calculated using the new `get_unlock_block_index` function. `get_unlock_block_index` is documented further below. The `std::vector<OutputContext>` for each unlock height should be sorted in the order outputs appear in the chain.

Upon adding a block, we'll add those outputs to the database here:

```cpp
// Add the locked outputs from this block to the locked outputs table
for (const auto &unlock_block : outs_by_unlock_block)
{
  MDB_val_set(k_block_id, unlock_block.first);
  for (const auto &locked_output : unlock_block.second)
  {
    MDB_val_set(v_output, locked_output);
    result = mdb_cursor_put(m_cur_locked_outputs, &k_block_id, &v_output, MDB_APPENDDUP);
    if (result != MDB_SUCCESS)
      throw0(DB_ERROR(lmdb_error("Failed to add locked output: ", result).c_str()));
  }
}
```

LMDB table changes are documented further below in section A.d. 

#### `get_unlock_block_index`

```cpp
// Returns the block index in which the provided unlock_time unlocks
uint64_t get_unlock_block_index(uint64_t unlock_time, uint64_t block_included_in_chain);
```

The idea behind this function is to have a deterministic and efficient method of growing the tree when outputs unlock.

Most outputs in the chain don't include an `unlock_time`; those outputs unlock 10 blocks after they are included in the chain.

Some outputs include an `unlock_time` which should either be interpreted as the *height* at which an output should unlock, or the *time* at which an output should unlock. When the `unlock_time` should be interpreted as height, the response to `get_unlock_block_index` is trivial. When interpreted as time, the logic is less straightforward. In this PR, as proposed by @kayabaNerve, I use the prior hard fork's block and time as an anchor point, and determine the unlock block from that anchor point. By converting timestamped `unlock_time` to a deterministic unlock block, we avoid needing to search for outputs that unlock by timestamp.

Note it is possible (likely) for the returned `unlock_block_index` to be distinct from current consensus' enforced unlock block **for timestamp-based locked outputs only**. The proposal is for consensus to enforce this new rule for fcmp++'s (users won't be able to construct fcmp's until outputs unlock according to the rules of  `get_unlock_block_index`).

Note: `get_unlock_block_index` from `unlock_time` is not in production form as is. The calculation should account for:
- The period of time where consensus supported 60s block times.
- The network (mainnet, testnet, stagenet).
- It should be vetted against the chain to make sure it calculates reasonable values.

## b. Curve trees merkle tree: `grow_tree`

This function takes a set of new outputs and uses them to grow the tree.

It has 3 core steps:

1. Get the number of leaf tuples currently in the tree, as well as the last hash (i.e. most recently inserted) from every layer.
2. Use the above two results to construct a tree extension, which we can use to extend the tree.
3. Use the tree extension to extend the tree.

Steps 1 and 3 are fairly straightforward. Step 2 carries the most weight and is the most complex. It's implemented in the `CurveTrees` class `get_tree_extension` function documented further below.

This step-wise approach enables clean separation of the db logic (steps 1 and 3) from the grow logic (step 2). In my view, this separation enables cleaner, more efficient code, and stronger testing. It also enables reusable tree building code for wallet scanning.

#### `get_tree_extension`

```cpp
// Take in the existing number of leaf tuples and the existing last hash in each layer in the tree, as well as new
// outputs to add to the tree, and return a tree extension struct that can be used to extend a tree
TreeExtension get_tree_extension(const uint64_t old_n_leaf_tuples,
    const LastHashes &existing_last_hashes,
    std::vector<OutputContext> &&new_outputs) const;
```

`get_tree_extension` has 2 core steps:

1. Prepare new leaves for insertion into the tree.
  
    a. Sort new outputs by the order they appear in the chain (guarantees consistent insertion order in the tree).
  
    b. Convert valid outputs to leaf tuples (from the form `{output_pubkey,commitment}`to `{O,I,C}` to `{O.x,I.x,C.x}`).

    - Points must be torsion cleared before inserting into the tree.
    - We cannot insert points that are **not** on ed25519 into the tree, nor points equal to identity after clearing torsion.
    - Thus we ignore any outputs which have `output_pubkey` or `commitment` that are not on the ed255129 curve, or are equal to identity after clearing torsion.
    - From @kayabaNerve: such outputs should either not be spendable today or are worth 0.
    - See the `CurveTrees<Helios, Selene>::LeafTuple CurveTrees<Helios, Selene>::leaf_tuple` function for the code.

    c. Place all leaf tuple members in a flat vector (`[{output 0 output pubkey and commitment}, {output 1 output pubkey and commitment},...]` becomes`[O.x,I.x,C.x,O.x,I.x,C.x,...]`).

2. Go layer by layer, hashing chunks of the preceding layer, and place results in the `TreeExtension` struct.
    
    a. Get `GrowLayerInstructions` for the current layer.
      - The logic to get `GrowLayerInstructions` for the layer after the leaf layer is distinct from all other layers after.
      - Using just the `old_total_children`, `new_total_children`, `parent_chunk_width`, and a bool for whether or not the `last_child_will_change`, we can determine how exactly we expect a layer to grow.
      - There are edge cases to watch out for. Edge case examples:
        - Example 1: the existing last hash in a layer may *change* from an old value to a new value.
        - Example 2: to efficiently update a hash from an old value to a new value, we may need to use the old AND new value of the prior last child because that child updated as well.
        - Example 3: if we need to use the old value of the prior last child, we need to adjust the offset we use to hash chunks.
        - Example 4: we may be adding a new layer after the existing root, and may need to use the existing root to get the next layer's first hash.
        
    b. Get the `LayerExtension` for the current layer to add to the `TreeExtension` struct.
      - We use `GrowLayerInstructions` to determine correct values when hashing the preceding "child" layer.

## c. Curve trees merkle tree: `trim_tree`

This function trims the provided number of leaf tuples from the tree.

The function has 5 core steps:

1. Get the number of leaf tuples currently in the tree.
2. Use the number of leaf tuples currently in the tree to get `TrimLayerInstructions`, which we can use to know how to trim each layer in the tree.
3. Get all the children we'll need to trim from the new last chunk in each layer, as well as the existing last hash from the new last chunk in each layer.
4. Use the results from steps 2 and 3 to construct a `TreeReduction` struct, which we can use to trim the tree.
5. Use the `TreeReduction` struct to trim the tree.

Step 1 is straightforward.

Step 2 carries the most weight and is the most complex. It's implemented in the `CurveTrees` class `get_trim_instructions` function documented further below.

In step 3, the "new last chunk in each layer" is referring to what will become the new last chunk in a layer **after trimming** that layer. We need values from those existing chunks in order to correctly and efficiently trim the chunk.

Step 4 is also complex, and is implemented in the `CurveTrees` class `get_tree_reduction` function documented further below.

Step 5 is straightforward.

Like for `grow_tree` this step-wise approach enables clean separation of db logic (steps 1, 3, 5) from the trim logic (steps 2 and 4).

#### `get_trim_instructions`

```cpp
// Get instructions useful for trimming all existing layers in the tree
std::vector<TrimLayerInstructions> get_trim_instructions(
    const uint64_t old_n_leaf_tuples,
    const uint64_t trim_n_leaf_tuples) const;
```

This function first gets instructions for trimming the leaf layer, then continues getting instructions for each subsequent layer until reaching the root.

The function doing the heavy lifting is:

```cpp
static TrimLayerInstructions get_trim_layer_instructions(
    const uint64_t old_total_children,
    const uint64_t new_total_children,
    const std::size_t parent_chunk_width,
    const bool last_child_will_change);
```

Similar to growing a layer, there are edge cases to watch out for when trimming a layer:
- Example 1: we may not be trimming *any* elements from a layer, only updating a hash.
- Example 2: it may be more efficient to trim by effectively re-growing an entire chunk with all elements from that chunk. In this case, we'll need to be sure to get the remaining children from that chunk.
- Example 3: it may be more efficient to trim by explicitly trimming children from that chunk. In this case, we need to be sure to get those existing children from that chunk.
- Example 4: we may need to explicitly trim **and** update the last child in a chunk.

This function captures these edge cases and outputs a struct that tells the caller how exactly to handle them.

#### `get_tree_reduction`

```cpp
// Take in the instructions useful for trimming all existing layers in the tree, all children to be trimmed from
// each last chunk, and the existing last hash in what will become the new last parent of each layer, and return
// a tree reduction struct that can be used to trim a tree
TreeReduction get_tree_reduction(
    const std::vector<TrimLayerInstructions> &trim_instructions,
    const LastChunkChildrenToTrim &children_to_trim,
    const LastHashes &last_hashes) const;
```

This function iterates over all layers, outputting a `LayerReduction` struct for each layer, which is a very simple struct we can use to trim a layer in the tree:

```cpp
// A struct useful to trim a layer and update its last hash if necessary
template<typename C>
struct LayerReduction final
{
    uint64_t          new_total_parents{0};
    bool              update_existing_last_hash;
    typename C::Point new_last_hash;
};
```

It uses each layer's `TrimLayerInstructions` from above as a guide, dictating exactly what data to use to calculate a new last hash for each layer.

## d. Curve trees merkle tree: LMDB changes

The following changes to the db are necessary in order to store and update the curve trees merkle tree.

### **NEW: `locked_outputs` table**

Potential outputs to be inserted into the merkle tree, indexed by the block ID in which the outputs unlock.

```
Key: `block ID`
Data: `[{output ID, out_pubkey, commitment}...]`
DB Flags: `MDB_INTEGERKEY | MDB_DUPSORT | MDB_DUPFIXED | MDB_CREATE`
```

We store the ouput ID to guarantee outputs are inserted into the tree in the order they appear in the chain.

This table stores the output pub key and commitment (64 bytes) instead of `{O.x,I.x,C.x}`, since `{O.x,I.x,C.x}` (96 bytes) can be derived from the output pub key and commitment, saving 32 bytes per output. Note that we should theoretically be able to stop storing the output public key and commitment in the `output_amounts` table at the hard fork, since that table should only be useful to construct and verify pre-fcmp++ txs.

### **NEW: `leaves` table**

Leaves in the tree.

```
Key: `leaf_idx`
Data: `{out_pubkey, commitment}`
DB Flags: `MDB_INTEGERKEY | MDB_CREATE`
```

Same as above: the record in this table is actually `{output_pubkey, commitment}` (64 bytes), since `{O.x,I.x,C.x}` (96 bytes) can be derived saving 32 bytes per output.

Note that we **must** save the output pub key for outputs in the chain **before** the fork that includes fcmp++, since we need to derive `I` from the pre-torsion cleared points. After the fork, we can store torsion cleared valid `{O,C}` pairs instead if we ban torsioned outputs and commitments at consensus, or if we redefine hash to point to use torsion cleared `O.x` as its input.

### **NEW: `layers` table**

Each record is a 32 byte hash of a chunk of children, as well as that hash's position in the tree.

```
Key: `layer_idx`
Data: `[{child_chunk_idx, child_chunk_hash}...]`
DB Flags: `MDB_INTEGERKEY | MDB_DUPSORT | MDB_DUPFIXED | MDB_CREATE`
```

The `layer_idx` is indexed starting at the layer after the leaf layer (i.e. `layer_idx=0` corresponds to the layer after the leaf layer).

Example: `{layer_idx=0, child_chunk_idx=4, child_chunk_hash=<31fa...>}` means that the `child_chunk_hash=<31fa...>` is a hash of the 5th chunk of leaves, and is a Selene point. Another example: `{layer_idx=1, child_chunk_idx=36, child_chunk_hash=<a2b5...>}` means that the `child_chunk_hash=<a2b5...>` is a hash of the 37th chunk of elements from `layer_idx=0`, and is a Helios point.

An even `layer_idx` corresponds to Selene points. An odd `layer_idx` corresponds to Helios points.

The element with the highest `layer_idx` is the root (which should also be the last element in the table). There should only be a single element with the highest `layer_idx` (i.e. only one data item with key == max `layer_idx`).

### **UPDATED: `block_info` table**

New fields:

```cpp
uint64_t bi_n_leaf_tuples;
std::array<uint8_t, 32UL> bi_tree_root;
```

`bi_n_leaf_tuples` - the number of leaf tuples in the tree at that height.

`bi_tree_root` - the root hash of the tree at that height. It is a (compressed) Helios point or Selene point, which can be determined from the number of leaf tuples in the tree.

## e. Curve trees merkle tree: Growing the tree as the node syncs

At each block, the tree must grow with (valid) outputs that unlock in that block. In the `add_block` function in `db_lmdb.cpp`, note the following:

```cpp
// Grow the tree with outputs that unlock at this block height
auto unlocked_outputs = this->get_outs_at_unlock_block_id(m_height);
this->grow_tree(std::move(unlocked_outputs));

// Now that we've used the unlocked leaves to grow the tree, we can delete them from the locked outputs table
this->del_locked_outs_at_block_id(m_height);
```

Then when adding the block, we get the number of leaf tuples in the tree and tree root and store them on each block info record:

```cpp
bi.bi_n_leaf_tuples = this->get_num_leaf_tuples();
bi.bi_tree_root = this->get_tree_root();
```

Finally, we use the container mentioned above to place the locked outputs from that block in  a "staging" `locked_outputs` table, ready to be used to grow the tree once they unlock.

### Comments

-  Tree building is single threaded currently and can be made multithreaded relatively easily when:
    - Converting outputs into leaf tuples.
    - Hashing chunks of scalars.
- The inversion in `point_to_wei_x` can be batched (makes multithreading when converting outputs into leaf tuples a bit more complex).

## f. Curve trees merkle tree: Migrating cryptonote outputs into the tree

All existing cryptonote outputs need to be migrated into the merkle tree.

- The migration determines the unlock block for every output and places them in the locked outputs table, and then grows the tree block by block with valid outputs.
- The migration must run to completion before a node can start normal sync, same as past migrations.
  - If the node is killed while the migration is ongoing, the migration will resume from where it left off upon restarting the node.
- The migration has 2 steps:
  1. Migrate outputs into the new `locked_outputs` table.
  2. Grow the tree block by block.
- The migration can theoretically be made asynchronous (it can run in the background while nodes start immediately, until the hard fork when the migration **must** run before nodes can continue syncing). Such a change would be a solid lift.

## g. Curve trees merkle tree: Key image migration

TODO

## h. Curve trees merkle tree: Trim on reorg and popping blocks

TODO

- Some existing tests need this implemented in order to pass.
- This also includes removing outputs from the `locked_outputs` table that need to be removed.

# C. Transaction struct changes for fcmp++

#### `rctSigBase`

Added a new `RCTType` enum usable in the `type` member of `rctSigBase`:

```cpp
RCTTypeFcmpPlusPlus = 7
```

fcmp++ txs are expected to use this `RCTType` instead of `RCTTypeBulletproofPlus` (even though fcmp++ txs are still expected to have a bp+ range proof).

______

Added a new member to `rctSigBase`:

```cpp
crypto::hash referenceBlock; // block containing the merkle tree root used for the tx's fcmp++
```

This member is only expected present on txs of `rctSigBase.type == RCTTypeFcmpPlusPlus`.

#### `rctSigPrunable`

Added 2 new members:

```cpp
uint8_t curve_trees_tree_depth; // for fcmp++
fcmp_pp::FcmpPpProof fcmp_pp;
```

Note there is a single opaque fcmp++ struct per tx. The `FcmpPpProof` type is simply a `std::vector<uint8_t>`. The length of the `FcmpPpProof` is deterministic from the number of inputs in the tx and curve trees merkle tree depth. Thus, when serializing and de-serializing, we don't need to store the vector length, and can expect a deterministic number of bytes for the `FcmpPpProof` by calling `fcmp_pp::proof_len(inputs, curve_trees_tree_depth)`.

### Comments

- The `tx_fcmp_pp` serialization test demonstrates what an expected dummy `transaction` struct looks like with dummy data.
- TODO: implement the function to get fcmp++ size from number of inputs and tree depth. We will need to decide between the 2 options described in [section 6.6 of the specification paper](https://github.com/kayabaNerve/fcmp-ringct/blob/b4a8cdae770770de1e96fa1c97c1e6e0926f8409/fcmp%2B%2B.pdf).
- TODO: tests for JSON serialization/de-serialization.

# D. Constructing fcmp++ transactions

TODO

# E. Verifying fcmp++ transactions

TODO

# F. Consensus changes for fcmp++

TODO


src/fcmp_pp/fcmp_pp_rust/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+
+set(RUST_TARGET "${RUST_ARCH}-${RUST_PLATFORM}${RUST_TOOLCHAIN}")
+
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(CARGO_CMD cargo build --target "${RUST_TARGET}")
+  set(TARGET_DIR "debug")
+else ()
+  set(CARGO_CMD cargo build --target "${RUST_TARGET}" --release)
+  set(TARGET_DIR "release")
+endif ()
```
jeffro256, line 94: After a quick skim though of this CMakeLists.txt, it might be a good idea to put all general Rust FFI stuff into a file by itself (perhaps in the /cmake/ folder idk), and open a PR specifically for that. I have a sneaking suspicion that we will be seeing more Rust FFI stuff in the future. The people who are good at all the build stuff (e.g. @selsta, @tobtoht, etc) could give this portion a nice read through.

src/fcmp_pp/fcmp_pp_rust/.gitignore
```diff
@@ -0,0 +1,4 @@
+# If a developer runs cargo build inside this sub-directory to only work with
+# the Rust side of things, they'll create this target directory which shouldn't
+# be committed
+target
```
jeffro256, line 4: Is there a way to move the `target` dir to the main `build` directory? Could be helpful if testing out different branches in development?

.github/workflows/depends.yml
```diff
@@ -31,36 +31,47 @@ jobs:
         toolchain:
           - name: "RISCV 64bit"
             host: "riscv64-linux-gnu"
+            rust_host: "riscv64gc-unknown-linux-gnu"
```
jeffro256, line 34: Yeah this stuff too should go into a "General Rust FFI" PR IMO

src/fcmp_pp/fcmp_pp_rust/fcmp++.h
```diff
@@ -0,0 +1,141 @@
...
+using HeliosScalarSlice = Slice<HeliosScalar>;
+
+using SeleneScalarSlice = Slice<SeleneScalar>;
+
+extern "C" {
+HeliosPoint helios_hash_init_point();
+
+SelenePoint selene_hash_init_point();
+
+uint8_t *helios_scalar_to_bytes(HeliosScalar helios_scalar);
```
vtnerd, line 97: C linkage with C++ types? I guess this is valid, but I'm not too sure about the rules on this quirk.

src/fcmp_pp/fcmp_pp_rust/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+#[no_mangle]
+pub extern "C" fn selene_point_to_bytes(selene_point: SelenePoint) -> *const u8 {
+    c_u8_32(selene_point.to_bytes())
+}
+
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+#[no_mangle]
+pub extern "C" fn helios_point_from_bytes(helios_point: *const u8) -> HeliosPoint {
+    let mut helios_point = unsafe { core::slice::from_raw_parts(helios_point, 32) };
+    // TODO: Return an error here (instead of unwrapping)
```
vtnerd, line 58: Another thing to mark as `\TODO` is how to handle Rust stack unwinds.  An unwind has undefined behavior once it hits an FFI boundary. There's a hidden (and very un-rust) utility called [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) that we may have to deploy in some of these functions to keep information details from leaking.

Unfortunately, I'm not really certain on what you do once you catch the unwind. The Rust code is an invalid state which means - immediate abort? Some quick Google searches didn't yield much insight as to what the standard practice is at the FFI boundary.

jeffro256: Does `libunwind` allow "pausing" the winding such that it won't be undefined to call the FFI, but just loses the information?

src/fcmp_pp/tower_cycle.cpp
```diff
@@ -0,0 +1,283 @@
...
+        existing_hash,
+        offset,
+        existing_child_at_offset,
+        new_children);
+
+    if (result.err != nullptr)
+    {
+        free(result.err);
+        throw std::runtime_error("failed to hash grow");
+    }
```
Boog900, line 74: Rust wont allocate for zero-sized types: https://doc.rust-lang.org/std/boxed/struct.Box.html#method.new

Which means the pointer you are freeing here is invalid and the other places using a ptr to `()`.

src/fcmp_pp/fcmp_pp_rust/src/lib.rs
Boog900: Rust will abort the process here if it was to panic: https://doc.rust-lang.org/nomicon/ffi.html#panic-can-be-stopped-at-an-abi-boundary

We could use `catch_unwind` to prevent the abort, however we are also setting `panic = "abort"` in the `Cargo.toml` which will just immediately abort on any panic as well.

src/fcmp_pp/tower_cycle.cpp
kayabaNerve: Shouldn't the pointer be null in that case (making this obviously broken for that reason, as a returned error won't be identified as an error)?

kayabaNerve: https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_raw

Will be non-null. I'm unsure the free behavior here 0_o

src/fcmp_pp/fcmp_pp_rust/.gitignore
kayabaNerve: Yep, it's an argument for the target dir with adjusted paths everywhere relevant.

src/fcmp_pp/fcmp_pp_rust/fcmp++.h
kayabaNerve: What type here is being considered C++? The scalars are from Rust with their layout defined in C code above.

src/fcmp_pp/fcmp_pp_rust/src/lib.rs
vtnerd: @Boog900  I missed that. That's probably the only thing that can be done on panic. I don't even know how/why `catch_unwind` is allowed, seems like it ruins tons of invariants if used to do something (other than abort).

src/fcmp_pp/fcmp_pp_rust/fcmp++.h
vtnerd: `HeliosScalar` has an internal type `Residue<32 / sizeof(uint32_t)>` which is a C++ template. This probably works because no fancy "decorators" are needed for the linkage of the outer type, but I'm not certain if the internal C++ type changes anything. At least on the platforms tested so far, it hasn't mattered.

Presumably since C++ has strict layout rules, the `static_assert` on `sizeof` catches the only possible issue - internal padding.

src/fcmp_pp/fcmp_pp_rust/fcmp++.h
```diff
@@ -0,0 +1,141 @@
+namespace fcmp_pp_rust {
+#include <cstdarg>
+#include <cstdint>
```
vtnerd, line 3: `#include` after a namespace? Never seen this move. Is this the lazy way to do `using namespace` in a header? I guess this works, but it seems like something would have to go wrong somehow.

src/fcmp_pp/fcmp_pp_rust/CMakeLists.txt
j-berman: Definitely planning to open a PR specifically for the Rust FFI first once it's ready. I'm thinking if/once we do see Rust FFI stuff used for other portions of the code, we can think about how to re-structure this portion once it's clear what new Rust code we want. Could avoid a premature debate when it might end up looking different in the future

src/fcmp_pp/fcmp_pp_rust/fcmp++.h
j-berman: Curious if you might have an alternative approach in mind for this code

src/fcmp_pp/fcmp_pp_rust/src/lib.rs
j-berman: I figure it makes sense to catch and handle errors, while panics should reasonably abort (and panics should never trigger unless the code is broken). Sounds like you'd agree?

src/fcmp_pp/tower_cycle.cpp
j-berman: Added a TODO for this

src/fcmp_pp/fcmp_pp_rust/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    } else {
+        // TODO: return defined error here: https://github.com/monero-project/monero/pull/9436#discussion_r1720477391
+        CResult::err(())
+    }
+}
+
+// https://github.com/rust-lang/rust/issues/79609
+#[cfg(all(target_os = "windows", target_arch = "x86"))]
+#[no_mangle]
+pub extern "C" fn _Unwind_Resume() {}
```
tobtoht, line 232: If mingw-w64 was compiled with dwarf2 exceptions enabled this will cause a multiple definitions error. I [do this](https://github.com/monero-project/monero/pull/9440/commits/866674d5c352093b9a17c1a979575229e74ac631) in #9440, because if you cross-compile the rust standard library with sjlj mingw-w64, you run into the same undefined references to unwind issue. Building std without unwinding should be possible, but I ran into problems. Removing these lines here breaks CI, because distros package sjlj mingw-w64.

I'll try to come up with the least hacky way to work around this. IMO, we could consider dropping the 32-bit Windows target. Windows 11 requires 64-bit and I can't imagine anyone using a >18 year old CPU to run Windows 10.

src/fcmp_pp/fcmp_pp_rust/src/lib.rs
vtnerd: Yes. If there's an error returned on the call stack, you can probably bubble that up through the FFI, otherwise you basically have to abort. Rust code really isn't written to be "exception-safe", so lots of invariants would break if an unwind occurs.

Chriss4123:
Hey man, I just want to say what you are doing here is absolutely incredible. If this is successful and integrated into Monero it would move it into a new stratosphere of privacy, incomparable to anything else really. I can't wait to see where this goes!

jeffro256:
Here's a patch that fixes linking errors on debug builds:

```patch
diff --git a/src/fcmp_pp/curve_trees.h b/src/fcmp_pp/curve_trees.h
index 561dc954c..2210a11a2 100644
--- a/src/fcmp_pp/curve_trees.h
+++ b/src/fcmp_pp/curve_trees.h
@@ -30,7 +30,6 @@
 
 #include "crypto/crypto.h"
 #include "cryptonote_basic/cryptonote_basic.h"
-#include "cryptonote_basic/cryptonote_boost_serialization.h"
 #include "fcmp_pp_crypto.h"
 #include "misc_log_ex.h"
diff --git a/src/fcmp_pp/tree_cache.h b/src/fcmp_pp/tree_cache.h
index 28fabdd4d..47a5d3fba 100644
--- a/src/fcmp_pp/tree_cache.h
+++ b/src/fcmp_pp/tree_cache.h
@@ -158,7 +158,7 @@ using ChildChunkCache   = std::unordered_map<ChildChunkIdx, CachedTreeElemChunk>
 // TODO: technically this can be a vector. There should *always* be at least 1 entry for every layer
 using TreeElemCache     = std::unordered_map<LayerIdx, ChildChunkCache>;
 
-static const int TREE_CACHE_VERSION = 0;
+static constexpr int TREE_CACHE_VERSION = 0;
 
@@ -308,7 +308,7 @@ struct version<fcmp_pp::curve_trees::TreeCache<C1, C2>>
 {
     typedef mpl::int_<fcmp_pp::curve_trees::TREE_CACHE_VERSION> type;
     typedef mpl::integral_c_tag tag;
-    BOOST_STATIC_CONSTANT(unsigned int, value = version::type::value);
+    static constexpr int value = version::type::value;
     BOOST_MPL_ASSERT((
         boost::mpl::less<
             boost::mpl::int_<fcmp_pp::curve_trees::TREE_CACHE_VERSION>,
diff --git a/src/ringct/rctOps.cpp b/src/ringct/rctOps.cpp
index 17a36bbae..62511c24a 100644
--- a/src/ringct/rctOps.cpp
+++ b/src/ringct/rctOps.cpp
@@ -30,7 +30,6 @@
 
 #include <boost/lexical_cast.hpp>
 #include "misc_log_ex.h"
-#include "cryptonote_basic/cryptonote_format_utils.h"
 #include "rctOps.h"
 using namespace crypto;
 using namespace std;
index 11d938b4a..f555bb274 100644
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -48,6 +48,7 @@
 #include "cryptonote_basic/account.h"
 #include "cryptonote_basic/account_boost_serialization.h"
 #include "cryptonote_basic/cryptonote_basic_impl.h"
+#include "cryptonote_basic/cryptonote_boost_serialization.h"
 #include "net/http.h"
 #include "storages/http_abstract_invoke.h"
 #include "rpc/core_rpc_server_commands_defs.h"

```

src/fcmp_pp/curve_trees.h
```diff
@@ -0,0 +1,463 @@
...
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include "crypto/crypto.h"
+#include "cryptonote_basic/cryptonote_basic.h"
+#include "cryptonote_basic/cryptonote_boost_serialization.h"
```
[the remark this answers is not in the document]
jeffro256: See patch

PR #9438, opened 2024-08-15
assign default port for http and https [RELEASE]

woodser:

PR #9439, opened 2024-08-15
assign default port for http and https

woodser:


ndorf:
If one wanted to be pedantic, the "proper" way to do this might be `getservbyname(parsed.schema)`...

woodser:
> If one wanted to be pedantic, the "proper" way to do this might be `getservbyname(parsed.schema)`...

I've updated it to use `getservbyname`:

```
if (parsed.port == 0 && (parsed.schema == "http" || parsed.schema == "https"))
{
  struct servent *serv = getservbyname(parsed.schema.c_str(), "tcp");
  parsed.port = ntohs(serv->s_port);
}
```

woodser:
I found that `getservbyname` does not work in emscripten environments, due to lack of system service databases like /etc/services.

Updated the PR to hardcode port 80 for http and 443 for https. These values are hardcoded elsewhere in the code so nothing new.

Bump to be included in the next release with the release PR: https://github.com/monero-project/monero/pull/9438

iamamyth:
> I found that getservbyname does not work in emscripten environments, due to lack of system service databases like /etc/services.

This possibility worried me. Furthermore, one person's feature is another's bug: a "functioning" service database which returns different ports could create problems for ordinary users. Given that there's an IANA service/port registry which includes 80 and 443, and that registry won't be changing any time soon, one could argue that making the lookup dynamic would *only* create wrong behaviors (i.e. strictly downside, because it's certainly slower).

> Updated the PR to hardcode port 80 for http and 443 for https. These values are hardcoded elsewhere in the code so nothing new.

It might be helpful to make (and use) a shared utility function in the future. But, in the interest of timeliness, I agree with the route you've chosen.

iamamyth (approved):

Asurar0 (approved):

iamamyth:
>  P2P for I2P (e.g. tx-proxy) uses default port 0

I would not have approved it if this were an issue: This function sets the server to which an HTTP client connects, port 0 is not a valid port, that is precisely why it's used to mean "default" or "any" in networking contexts, just like 0.0.0.0 means the same for IPv4 addresses.

selsta (approved):

PR #9441, opened 2024-08-17
blockchain: more opportunities for caching the rolling median

moneromooo-monero:


src/cryptonote_core/blockchain.cpp
```diff
@@ -1500,8 +1503,11 @@ uint64_t Blockchain::get_long_term_block_weight_median(uint64_t start_height, si
 
   // in the vast majority of uncached cases, most is still cached,
   // as we just move the window one block up:
-  if (tip_height > 0 && count == (size_t)m_long_term_block_weights_cache_rolling_median.size() && tip_height < blockchain_height)
+  if (tip_height > 0 && tip_height < blockchain_height)
   {
+   const size_t rmsize = m_long_term_block_weights_cache_rolling_median.size();
+   if (count == rmsize || (count == rmsize + 1 && rmsize < CRYPTONOTE_LONG_TERM_BLOCK_WEIGHT_WINDOW_SIZE))
```
0xFFFC0000, line 1509: I was reviewing this PR. All the other parts are okay and makes sense to me. 

The only part I don't comprehend is `(count == rmsize + 1 && rmsize < CRYPTONOTE_LONG_TERM_BLOCK_WEIGHT_WINDOW_SIZE)` .

What is purpose of this added condition?

0xFFFC0000 (commented):
Other than one thing I didn't fully comprehend. All the other parts are correct.

src/cryptonote_core/blockchain.cpp
moneromooo-monero: It matches the incremental case when the history has not yet reached CRYPTONOTE_LONG_TERM_BLOCK_WEIGHT_WINDOW_SIZE blocks. Since Monero is already past that, it could be removed.

0xFFFC0000 (approved):
LGTM

PR #9442, opened 2024-08-17
Add man page for monerod !

Justxd22:
All the essential `monerod` documentation, one quick place for docs without leaving your terminal:

- Includes `Options` and `daemon_commands`
- Comprehensive list of all `JSON RPC` methods
- Detailed `Example` section
- Code blocks included for clarity

This allows us to efficiently direct someone to a specific section of the man page, rather than vaguely referring them to "somewhere in the middle of the --help subcommand."

The man page will be bundled with the tarball releases, making it as simple as running man `./docs/man/monerod.1`.

### DIFF
--- a/docs/man/monerod.1
+++ b/docs/man/monerod.1
@@ -1,0 +1,5354 @@
+.TH MONEROD "1" "AUGUST 2024" "Fluorine Fermi 0.18.3.3" "MONEROD MANUAL"
+.SH NAME
+monerod \- Monero p2p network daemon
+.SH SYNOPSIS
+.B monerod
+[\fI\,options|settings\/\fR] [\fI\,daemon_command...\/\fR]
+.SH DESCRIPTION
+.\" Add any additional description here
+.PP
+The Monero daemon \fBmonerod\fR keeps your computer synchronized with the Monero blockchain by downloading and validating the blockchain from the p2p network.
+
+\fBmonerod\fR is entirely decoupled from your wallet.
+
+\fBmonerod\fR is mainly used by your wallet software over local RPC connection, as in \fBMonero GUI\fR.
+
+\fBmonerod\fR does not access your private keys \- it is not aware of your transactions and balance, so you can run \fBmonerod\fR on a separate computer or in the cloud. In fact, you can connect to a remote \fBmonerod\fR instance provided by a semi-trusted 3rd party, who will not be able to steal your funds; this third-party configuration is very handy for learning and experimentation.
+
+However, there are privacy and reliability implications to using a remote, untrusted node. For any real business you should be running your own full node.
+
+\fBmonerod\fR can save you time syncing the whole blockchain by relaying your wallet requests over p2p network nodes 
+.br
+by using \fB--bootstrap-daemon-address\fR see \fBmonerod Performance\fR section for more info.
+
+
+.SH Options:
+Arguments in parentheses are default
+.br
+.TP
+\fB\--help\fR
+.br
+Produce help message
+.TP
+\fB\--version\fR                            
+Output version information
+.TP
+\fB\--os-version\fR                          
+OS for which this executable was compiled.
+.TP
+\fB\--check-updates\fR arg (=notify)
+Check for new versions of monero: [disabled|notify|download|update]
+
+
+.SS Pick Monero network (blockchain):
+.TP
+By default \fBmonerod\fR runs on --mainnet
+.TP
+\fB\--testnet\fR
+Run on testnet. The wallet must be launched with --testnet flag.
+.TP
+\fB\--stagenet\fR
+Run on stagenet. The wallet must be launched with --stagenet flag.
+
+
+.SS Server Config:
+.TP
+\fB\--config-file\fR arg
+.br
+(=~/.bitmonero/bitmonero.conf, 
+.br
+~/.bitmonero/testnet/bitmonero.conf if 'testnet', 
+.br
+~/.bitmonero/stagenet/bitmonero.conf if 'stagenet')
+.br
+Specify full path to configuration file, By default \fBmonerod\fR looks for bitmonero.conf in Monero data directory.
+.TP
+\fB\--data-dir\fR arg
+(=~/.bitmonero,
+.br
+~/.bitmonero/testnet if 'testnet',
+.br
+~/.bitmonero/stagenet if 'stagenet')
+.br
+Specify full path for data directory, This is where the blockchain, log files, and p2p network memory are stored. 
+.TP
+\fB\--pidfile\fR arg                         
+File path to write the daemon's PID to (optional, requires \--detach)
+.TP
+\fB\--detach\fR
+Run as daemon
+.TP
+\fB\--non-interactive\fR                     
+Run non-interactive, Do not require tty in a foreground mode. Helpful when running in a container. By default \fBmonerod\fR runs in a foreground and opens stdin for reading.
+.TP
+\fB\--max-txpool-weight\fR arg (=648000000)
+Set maximum txpool weight in bytes. These are transactions pending for confirmations (not included in any block).
+.TP
+\fB\--enforce-dns-checkpointing\fR
+checkpoints from DNS server will be enforced, The emergency checkpoints set by \fBMoneroPulse\fR operators will be enforced. It is probably a good idea to set enforcing for unattended nodes.
+If encountered block hash does not match corresponding checkpoint, the local blockchain will be rolled back a few blocks, effectively blocking following what MoneroPulse operators consider invalid fork. The log entry will be produced: \fBERROR Local blockchain failed to pass a checkpoint, rolling back!\fR Eventually, the alternative (“fixed”) fork will get heavier and the node will follow it, leaving the “invalid” fork behind.
+.TP
+\fB\--disable-dns-checkpoints\fR
+Do not retrieve checkpoints from DNS
+.TP
+\fB\--ban-list\fR arg
+Specify ban list file, one IP address per line
+.TP
+\fB\--enable-dns-blocklist\fR      
+Apply realtime blocklist from DNS
+.TP
+\fB\--zmq-pub\fR arg
+Address for ZMQ pub - tcp://ip:port or ipc://path
+.TP
+\fB\--zmq-rpc-bind-ip\fR arg (=127.0.0.1)
+IP for ZMQ RPC server to listen on
+.TP
+\fB\--zmq-rpc-bind-port\fR arg (=18082, 28082 if 'testnet', 38082 if 'stagenet')
+Port for ZMQ RPC server to listen on
+.TP
+\fB\--block-download-max-size\fR arg (=0)
+Set maximum size of block download queue in bytes (0 for default)
+.TP
+\fB\--span-limit\fR arg (=2)
+Dynamically restrict the number of spans in queue by estimating how many spans can be synced within N minutes. The dynamic minimum is 10 spans, and is further capped by block-download-max-size.
+.TP
+\fB\--no-sync\fR                   
+Don't synchronize the blockchain with other peers, intended to be used with \fB\--bootstrap-daemon-address\fR allowing reliance on other public P2P nodes for blockchain data.
+
+When a monerod process is in \fB\--no-sync\fR mode, virtually all data requests are forwarded to the bootstrap node, as the local node does not have synchronized block data.
+
+This flag is enabled in Simple GUI Mode.
+
+.SS Logging:
+.TP
+\fB\--log-file\fR arg
+.br
+(=~/.bitmonero/bitmonero.log,
+.br
+~/.bitmonero/testnet/bitmonero.log if 'testnet',
+.br
+~/.bitmonero/stagenet/bitmonero.log if 'stagenet')
+.br
+Specify Full path log file
+.TP
+\fB\--log-level\fR arg (=0)
+0-4 with 0 being minimal logging and 4 being full tracing. These are general presets and do not directly map to severity levels.
+.TP
+\fB\--max-log-file-size\fR arg (=104850000)
+Specify maximum log file size [B], Once log file grows past that limit, \fBmonerod\fR creates the next log file with a UTC timestamp postfix -YYYY-MM-DD-HH-MM-SS.
+.TP
+\fB\--max-log-files\fR arg (=50)
+Specify maximum number of rotated log
+files to be saved (no limit by setting to 0)
+
+
+.SS monerod Performance:
+.TP
+\fB\--sync-pruned-blocks\fR
+Allow syncing from nodes with only pruned blocks
+.TP
+\fB\--prune-blockchain\fR
+Prune blockchain
+.TP
+\fB\--db-sync-mode\fR arg (=fast:async:250000000bytes)
+Specify sync option, using format [safe|fast|fastest]:[sync|async]:[<nblocks_per_sync>[blocks]|<nbytes_per_sync> [bytes]].
+.TP
+\fB\--max-concurrency\fR arg (=0)
+Max number of threads to use for a parallel job
+.TP
+\fB\--prep-blocks-threads\fR arg (=4)
+Max number of threads to use when preparing block hashes in groups.
+.TP
+\fB\--fast-block-sync\fR arg (=1)
+Sync up most of the way by using embedded, known block hashes.
+.TP
+\fB\--block-sync-size\fR arg (=0)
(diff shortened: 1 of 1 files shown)


iamamyth:
Thanks for putting this together; it looks quite thorough and really does improve the user experience.

Justxd22:
thanks :D

selsta:
Could you squash the commits together?

Justxd22:
@selsta done

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
+.TH MONEROD "1" "AUGUST 2024" "Fluorine Fermi 0.18.3.3" "MONEROD MANUAL"
+.SH NAME
+monerod \- Connects you to Monero network
+.SH SYNOPSIS
+.B monerod
+[\fI\,options|settings\/\fR] [\fI\,daemon_command...\/\fR]
+.SH DESCRIPTION
+.\" Add any additional description here
+.PP
+The Monero daemon \fBmonerod\fR keeps your computer synced up with the Monero network.
```
iamamyth, line 10: Suggest using the full word (`synchronized`) here and removing "up": `keeps your computer synchronized with`

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
...
+
+It downloads and validates the blockchain from the p2p network.
+
+\fBmonerod\fR is entirely decoupled from your wallet.
+
+\fBmonerod\fR is mainly used by your wallet software over local RPC connection, as in \fBMonero GUI\fR.
+
+\fBmonerod\fR does not access your private keys \- it is not aware of your transactions and balance.
+
+This allows you to run \fBmonerod\fR on a separate computer or in the cloud.
```
iamamyth, line 20: This sentence should be part of the preceding paragraph, otherwise you have "this" without any antecedent. Relatedly, I would suggest rewording, for clarity:
```
\fBmonerod\fR does not access your private keys \- it is not aware of your transactions and balance, so you can run \fBmonerod\fR on a separate computer or in the cloud.
```

You can then also join this reworded intro with the following paragraph (note some minor grammar edits):
```
\fBmonerod\fR does not access your private keys \- it is not aware of your transactions and balance, so you can run \fBmonerod\fR on a separate computer or in the cloud. In fact, you can connect to a remote \fBmonerod\fR instance provided by a semi-trusted 3rd party, who will not be able to steal your funds; this third-party configuration is very handy for learning and experimentation.
```

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
...
+
+In fact, you can connect to a remote \fBmonerod\fR instance provided by a semi-trusted 3rd party. Such 3rd party will not be able to steal your funds. This is very handy for learning and experimentation.
+
+However, there are privacy and reliability implications to using a remote, untrusted node. For any real business you should be running your own full node.
```
iamamyth, line 24: Would probably make this sentence the conclusion of the prior paragraph on remote/3rd party nodes.

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
...
+
+\fBmonerod\fR can save you time syncing the whole blockchain by relaying your wallet requests over p2p network nodes 
+.br
+by using \fB--bootstrap-daemon-address\fR see \fBmonerod Performance\fR section for more info.
+
+
+.SH Options:
+Arguments in brackets are default
```
iamamyth, line 32: I think brackets here should say parentheses.

iamamyth: To clarify: `[` is a backet, `(` is a parentesis, `(` is sometimes called a "round bracket" but the term "bracket" without any adjective normally means a square bracket (and is too ambiguous because the documentation contains both round and square brackets).

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
+.TH MONEROD "1" "AUGUST 2024" "Fluorine Fermi 0.18.3.3" "MONEROD MANUAL"
+.SH NAME
+monerod \- Connects you to Monero network
```
iamamyth, line 3: Daemon descriptions in man pages tend to be a bit more precise, I might suggest:
```
Monero p2p network daemon
```

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
...
+\fB\--detach\fR
+Run as daemon
+.TP
+\fB\--non-interactive\fR                     
+Run non-interactive, Do not require tty in a foreground mode. Helpful when running in a container. By default \fBmonerod\fR runs in a foreground and opens stdin for reading.
+.TP
+\fB\--no-zmq\fR
+Disable ZMQ RPC server
+.TP
+\fB\--no-igd\fR
```
iamamyth, line 93: This should probably be in the same section as `--igd`

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
...
+\fB\--fast-block-sync\fR arg (=1)
+Sync up most of the way by using embedded, known block hashes.
+.TP
+\fB\--block-sync-size\fR arg (=0)
+How many blocks to sync at once during chain synchronization (0 = adaptive).
+.TP
+\fB\--block-download-max-size\fR arg (=0)
+Set maximum size of block download queue in bytes (0 for default)
+.TP
+\fB\--no-sync\fR
```
iamamyth, line 156: These options (especially the ones starting with extra-mesages-file) aren't really performance related.  I suggest folding them into the core server config section.

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
...
+\fB\--regtest\fR
+Run in a regression testing mode.
+.TP
+\fB\--fixed-difficulty\fR arg (=0)
+Fixed difficulty used for testing.
+
+
+.SS Legacy Options
+.TP
+\fB\--ban-list\fR arg
```
iamamyth, line 419: With the exception of fluffy-blocks, these are not "legacy options", and some, such as ban-list, are currently suggested best practice to defend against network snooping by fleets node proxies. I would suggest just folding all of these into the server config section, though if you want to keep [no-]fluffy-blocks here, that's fine.

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
...
+Get or set the download limit.
+.TP
+\fBlimit_up [<kB/s>]\fR
+Get or set the upload limit.
+
+
+
+.SS Transaction pool:
+.TP
+\fBflush_cache [bad-txs] [bad-blocks]\fR
```
iamamyth, line 509: I wouldn't consider `flush_cache bad-blocks` a transaction pool command; would suggest moving `flush_cache` to the blockchain section.

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
...
+
+Mind blockchain reorganizations. Block notifications can revert to same and past heights. Small reorganizations are natural and happen every day.
+
+
+.SS Change blocks sync size:
+How many blocks are processed in a single batch during chain synchronization. By default this is 20 blocks for newer history and 100 blocks for older history (“pre v4”). Default behavior is represented by value 0. Intuitively, the more resources you have, the bigger batch size you may want to try out.
+
+.B ./monerod --block-sync-size=500
+
+.SS bootstrap:
```
iamamyth, line 5280: Should probably capitalize the title here (`Bootstrap`).

docs/man/monerod.1
```diff
@@ -0,0 +1,5343 @@
...
+mainnet is when you want to deal with the real XMR.
+
+\fB\./monerod \--detach\fR        # run as a daemon in background
+.br
+\fBtail \-f ~/.bitmonero/bitmonero.log\fR  # watch the logs
+.br
+\fB\./monerod exit\fR           # ask daemon to exit gracefully
+
+
+.SS p2p network protocol over Tor via torsocks:
```
iamamyth, line 5228: Most conventions would consider `P2P` an acronym and capitalize it throughout the document; I don't mind keeping it lower case, but, as the first word in a section heading, it should have the first letter capitalized (`P2p`)

iamamyth: Note: The prior comment applies to the following section, as well.

iamamyth:
There's a case to be made that the RPC documentation should be a separate man page which the main man page references; I don't have a strong opinion on the matter, though I will note that the RPC tends to change more regularly than the core daemon, which might slightly strengthen the case for a split.

docs/man/monerod.1
nahuhh: Probably pulled / transposed some info from old monerodocs.org 

docs.getmonero.org updated this info to properly categorize some of the so-called "legacy" options

Justxd22:
@iamamyth about the separate RPC man page, i don't like the idea of having 2 man pages and having to switch between them, or if someone refers you to x section in the man page they have to specify which one of the two, BUT if it's needed happy to do it

docs/man/monerod.1
Justxd22: i only used getmonero.dev and getmonero.org for info, and they still list them in legacy looks like they need updating as well

nahuhh: Docs.getmonero.org does not list as legacy

Justxd22: yup updated them

docs/man/monerod.1
```diff
@@ -0,0 +1,5338 @@
...
+.TP
+\fB\--enforce-dns-checkpointing\fR
+checkpoints from DNS server will be enforced, The emergency checkpoints set by \fBMoneroPulse\fR operators will be enforced. It is probably a good idea to set enforcing for unattended nodes.
+If encountered block hash does not match corresponding checkpoint, the local blockchain will be rolled back a few blocks, effectively blocking following what MoneroPulse operators consider invalid fork. The log entry will be produced: \fBERROR Local blockchain failed to pass a checkpoint, rolling back!\fR Eventually, the alternative (“fixed”) fork will get heavier and the node will follow it, leaving the “invalid” fork behind.
+.TP
+\fB\--disable-dns-checkpoints\fR
+Do not retrieve checkpoints from DNS
+.TP
+\fB\--ban-list\fR arg
+Specify ban list file, one IP address per line
```
iamamyth, line 97: Per comment by `nahuhh`, apparently the other monero docs list `ban-list` under p2p rather than core server options (this would also apply to `enable-dns-blocklist`). I can see a case for either choice (if memory serves, the bans only apply to p2p, if they also applied to RPC the options would definitely belong here).

docs/man/monerod.1
```diff
@@ -0,0 +1,5338 @@
...
+monerod \- Monero p2p network daemon
+.SH SYNOPSIS
+.B monerod
+[\fI\,options|settings\/\fR] [\fI\,daemon_command...\/\fR]
+.SH DESCRIPTION
+.\" Add any additional description here
+.PP
+The Monero daemon \fBmonerod\fR keeps your computer synchronized with the Monero network.
+
+It downloads and validates the blockchain from the p2p network.
```
iamamyth, line 12: Would fold into preceding sentence:
```
The Monero daemon \fBmonerod\fR keeps your computer synchronized with the Monero blockchain by downloading and validating the blockchain from the p2p network.
```

iamamyth:
A couple of (I suspect final) minor comments, but this looks good.

docs/man/monerod.1
nahuhh: I just tested and it looks like it blocks rpc as well

Justxd22: so @nahuhh i should keep it as is ?

Justxd22: @iamamyth what about this? as per @nahuhh it's p2p & rpc, so i should keep it as is

iamamyth: Yes keep as-is.

iamamyth (approved):
Please squash commits.

docs/man/monerod.1
nahuhh: block-sync-size=500 wont work btw. Max is 100 due to some broken code. If you edit code to allow >100, you'll only be able to sync >100 from node who have also run the updated version

Justxd22: @nahuhh Does the official docs say so?

nahuhh: Does it say that the max is 100? No. Because its a bug. The max is supposed to be 2048 but it wont ever switch to that unless done so in a hard fork.

You can probably leave the example as 500, assuming it will be fixed in the upcoming HF (it hasnt been yet)

https://github.com/nahuhh/monero/commit/52d9682aba133779bf53b37318462d1b0d8f9216

nahuhh:
@Justxd22 there have been some changes and additions, but i guess can be done in a followup pr

```
\fB\--limit-rate-up\fR arg (=2048)
 set limit-rate-up [kB/s]
.TP
\fB\--limit-rate-down\fR arg (=8192)
 set limit-rate-down [kB/s]
```

is now 8192 and 32768

there are new flags like --batch-max-weight (related to block-sync-size) and --span-limit (related to block-download-max-size). (neither of these is merged on docs [yet](https://github.com/monero-project/monero-docs/pull/271))

I'm sure im missing some other changes as well, but those are the 2 that immediately come to mind

Justxd22:
@nahuhh  should i add them now? or a follow-up PR?

selsta:
please add it to this PR

nahuhh:
@Justxd22 drop the merge commit

`git reset --hard HEAD~1`

I dont think this PR will need a rebase, so no need to update the base

Justxd22:
@nahuhh i added couple of missing flags as well, it doesn't show the changes it was squashed...

```diff

diff --git a/docs/man/monerod.1 b/docs/man/monerod.1
index 426d45a37..f18cf2e3c 100644
--- a/docs/man/monerod.1
+++ b/docs/man/monerod.1
@@ -109,6 +109,9 @@ Port for ZMQ RPC server to listen on
 \fB\--block-download-max-size\fR arg (=0)
 Set maximum size of block download queue in bytes (0 for default)
 .TP
+\fB\--span-limit\fR arg (=2)
+Dynamically restrict the number of spans in queue by estimating how many spans can be synced within N minutes. The dynamic minimum is 10 spans, and is further capped by block-download-max-size.
+.TP
 \fB\--no-sync\fR                   
 Don't synchronize the blockchain with other peers, intended to be used with \fB\--bootstrap-daemon-address\fR allowing reliance on other public P2P nodes for blockchain data.
 
@@ -160,7 +163,10 @@ Max number of threads to use when preparing block hashes in groups.
 Sync up most of the way by using embedded, known block hashes.
 .TP
 \fB\--block-sync-size\fR arg (=0)
-How many blocks to sync at once during chain synchronization (0 = adaptive).
+How many blocks are processed in a single batch during chain synchronization. Setting this flag will override the default behavior, which is to be calculated dynamically by batch-max-weight. Max value is 100.
+.TP
+\fB\--batch-max-weight\fR arg (=10)
+Dynamically set the block-sync-size by targetting this many MB. Max value is 50.
 .TP
 \fB\--bootstrap-daemon-address\fR arg
 URL of a 'bootstrap' remote daemon that the connected wallets can use while this daemon is still not fully synced. Use 'auto' to enable automatic public nodes discovering and bootstrap daemon switching,
@@ -241,10 +247,10 @@ Specify list of peers to connect to only. If this option is given the options ad
 \fB\--in-peers\fR arg (=-1)
  set max number of in peers
 .TP
-\fB\--limit-rate-up\fR arg (=2048)
+\fB\--limit-rate-up\fR arg (=8192)
  set limit-rate-up [kB/s]
 .TP
-\fB\--limit-rate-down\fR arg (=8192)
+\fB\--limit-rate-down\fR arg (=32768)
  set limit-rate-down [kB/s]
 .TP
 \fB\--limit-rate\fR arg (=-1)
@@ -301,6 +307,18 @@ Ignore unsuccessful IPv4 bind for RPC
 \fB\--restricted-rpc\fR
 Restrict RPC to view only commands and do not return privacy sensitive data in RPC calls
 .TP
+\fB\--rpc-max-connections\fR arg (=100)
+Max RPC connections permitted
+.TP
+\fB\--rpc-max-connections-per-public-ip\fR arg (=3)
+Max RPC connections per public IP permitted
+.TP
+\fB\--rpc-max-connections-per-private-ip\fR arg (=25)
+Max RPC connections per private and localhost IP permitted
+.TP
+\fB\--rpc-response-soft-limit\fR arg (=26214400)
+Max response bytes that can be queued, enforced at next response attempt
+.TP
 \fB\--rpc-restricted-bind-ip\fR arg (=127.0.0.1)
 Specify IP to bind restricted RPC server
 .TP

```

nahuhh:
https://github.com/monero-project/monero/pull/10388

PR #9445, opened 2024-08-19
cmake: fix readline find_path call.

0xFFFC0000:
When passing `PATHS` and `NO_DEFAULT_PATH` to `find_path` in CMake, `find_path` will not look into any other location. This will make building on non-canonical *ix based system hard. For example `GUIX` package solves this by passing argument: 

Search for `DReadline_ROOT_DIR=` in 
https://git.savannah.gnu.org/cgit/guix.git/tree/gnu/packages/finance.scm

### DIFF
--- a/cmake/FindReadline.cmake
+++ b/cmake/FindReadline.cmake
@@ -23,8 +23,7 @@
 
 find_path(Readline_ROOT_DIR
     NAMES include/readline/readline.h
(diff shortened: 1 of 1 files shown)


0xFFFC0000:
https://github.com/ecbrown/guix/blob/6ab242609daec00e8bd54f7bff54557c92695724/gnu/packages/finance.scm#L733

tobtoht (approved):

Issue #9359, opened 2024-06-11
Compilation errors on gcc 14.1.1


moneroexamples:
Regarding:

> error: implicit declaration of function ‘strdup’; did you mean ‘strcmp’? [-Wimplicit-function-declaration]

and

> error: assignment to ‘char *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]

```
/home/mwo/monero/external/miniupnp/miniupnpc/listdevices.c: In function ‘add_device’:
/home/mwo/monero/external/miniupnp/miniupnpc/listdevices.c:60:24: error: implicit declaration of function ‘strdup’; did you mean ‘strcmp’? [-Wimplicit-function-declaration]
   60 |         elt->descURL = strdup(dev->descURL);
      |                        ^~~~~~
      |                        strcmp
/home/mwo/monero/external/miniupnp/miniupnpc/listdevices.c:60:22: error: assignment to ‘char *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
   60 |         elt->descURL = strdup(dev->descURL);
      |                      ^
make[3]: *** [external/miniupnp/miniupnpc/CMakeFiles/listdevices.dir/build.make:76: external/miniupnp/miniupnpc/CMakeFiles/listdevices.dir/listdevices.c.o] Error 1
make[2]: *** [CMakeFiles/Makefile2:1307: external/miniupnp/miniupnpc/CMakeFiles/listdevices.dir/all] Error 2
make[1]: *** [Makefile:146: all] Error 2
make[1]: Leaving directory '/home/mwo/monero/build/release'
make: *** [Makefile:103: release-all] Error 2
```

SyntheticBird45:
Confirm, same issue on updated Arch Linux

SyntheticBird45:
@selsta Applied pastebin modifications.
Failed at 24%:
```
[scrubbed]/monero/external/miniupnp/miniupnpc/src/minihttptestserver.c:299:25: error: implicit declaration of function 'usleep'; did you mean 'sleep'? [-Wimplicit-function-declaration]
  299 |                         usleep(10000); /* 10ms */
      |                         ^~~~~~
      |                         sleep
make[2]: *** [external/miniupnp/miniupnpc/CMakeFiles/minihttptestserver.dir/build.make:76: external/miniupnp/miniupnpc/CMakeFiles/minihttptestserver.dir/src/minihttptestserver.c.o] Error 1
[...]

Issue #9446, opened 2024-08-20
[Proposal] General rule for minimum package versions and distribution support


tobtoht:
It would be nice if we could reach consensus on a general rule for minimum package versions and distribution support. This comes up occasionally (e.g. #9443, #9171, #8922) and it always seems to spark a debate that goes nowhere.

I would like to propose that we guarantee support for distributions until EOL date + 1 year, to give users ample time to upgrade their machine.

Users on distributions that don't package the minimum version of a required library can still build Monero using `make depends`. 

#### Why?

- To give developers access to new library features when they believe this would benefit their code.
- To reduce maintenance burden associated with guaranteeing compatibility with a wide range of library versions / compilers / operating systems.
- To prevent us from having to argue about this every time a minimum version change is proposed.

#### Example

| OS           | EOL date    | Boost version | Supported   |
|--------------|-------------|---------------|-------------|
| Ubuntu 16.04 | 02 Apr 2021 | 1.58          | No          |
| Ubuntu 18.04 | 21 May 2023 | 1.65          | No  |
| Ubuntu 20.04 | 02 Apr 2025 | 1.71          | No         |
| Debian 9 (LTS) | 01 Jul 2022 | 1.62 | No |
| Debian 10 (LTS)    | 30 Jun 2024 | 1.67          | No          |
| Debian 11 (LTS)   | 31 Aug 2026 | 1.74          | Yes |

With the proposed rule, we would currently guarantee support for Boost 1.71 at minimum.

#### A more complete list

| Build tool   | Version |
|--------------|---------|
| binutils     | 2.35 |
| cmake        | 3.18 |
| curl         | 7.74.0 |
| cxx standard | c++17 |
| doxygen      | 1.9.1 |
| gcc          | 10.2.1 |
| graphviz     | 2.42.2 |
| make         | 4.3 |
| python       | 3.9 |
| **Library**  |          |
| boost        | 1.74 |
| hidapi       | 0.10.1 |
| libsodium    | 1.0.18 |
| libusb       | 1.0.23 |
| libzmq       | 4.3.2 |
| openssl      | 1.1.1w |
| protobuf     | 3.12.4 |
| readline     | 8.1 |
| unbound      | 1.13.1 |
| **Runtime**  |         |
| glibc        | 2.31 |

| Operating system | Version   | Until       |
|------------------|-----------|-------------|
| [macOS](https://endoflife.date/macos)            | 13        | 15 Sep 2026 |
[...]

PR #9448, opened 2024-08-22
readme: Update minimal dependencies

Asurar0:
Monero should keep minimal versions of dependencies updated to reasonable and secure versions to avoid introducing vulnerabilities in third-party builds and stay coherent with its current building process.

The following packages have been updated in this PR:
- GCC (7 -> 12): GCC 7 is outdated since 2022 and set C++17 templates behind an experimental flag. I'm doubtful about its ability to compile monerod. Moved to 12 because most distributions are shipping 13 and transitioning to 14 ([monerod issue with gcc14](https://github.com/monero-project/monero/issues/9359) / [Fix PR](https://github.com/monero-project/monero/pull/9367)).
- Libunbound (1.4.16 -> 1.13.1): A lot of memory corruption as well as non memory related vulnerabilities have been reported since version 1.4.16. Latest version is 1.21.0 and latest package on ubuntu 22.04 repository is 1.13.1, I updated it to 1.13.1.
- Libzmq (4.2.0 -> 4.3.4): Same as libunbound, [lot of security vulnerabilities reported](https://www.cvedetails.com/vulnerability-list/vendor_id-14186/product_id-52489/version_id-1616705/Zeromq-Libzmq-4.2.0.html?page=1&order=1&trc=6&sha=ae6b688fb48903dc3a072a0ea241a16922c27d9a) since this version. Some can be mitigated with exploit mitigations at build time but it is unreasonable to recommend such an old version. Updated to its latest version without known vulnerability

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,15 +164,15 @@ library archives (`.a`).
 
 | Dep          | Min. version  | Vendored | Debian/Ubuntu pkg    | Arch pkg     | Void pkg           | Fedora pkg          | Optional | Purpose         |
 | ------------ | ------------- | -------- | -------------------- | ------------ | ------------------ | ------------------- | -------- | --------------- |
-| GCC          | 7             | NO       | `build-essential`    | `base-devel` | `base-devel`       | `gcc`               | NO       |                 |
+| GCC          | 12            | NO       | `build-essential`    | `base-devel` | `base-devel`       | `gcc`               | NO       |                 |
 | CMake        | 3.5           | NO       | `cmake`              | `cmake`      | `cmake`            | `cmake`             | NO       |                 |
 | pkg-config   | any           | NO       | `pkg-config`         | `base-devel` | `base-devel`       | `pkgconf`           | NO       |                 |
 | Boost        | 1.58          | NO       | `libboost-all-dev`   | `boost`      | `boost-devel`      | `boost-devel`       | NO       | C++ libraries   |
 | OpenSSL      | basically any | NO       | `libssl-dev`         | `openssl`    | `openssl-devel`    | `openssl-devel`     | NO       | sha256 sum      |
-| libzmq       | 4.2.0         | NO       | `libzmq3-dev`        | `zeromq`     | `zeromq-devel`     | `zeromq-devel`      | NO       | ZeroMQ library  |
+| libzmq       | 4.3.4         | NO       | `libzmq3-dev`        | `zeromq`     | `zeromq-devel`     | `zeromq-devel`      | NO       | ZeroMQ library  |
 | OpenPGM      | ?             | NO       | `libpgm-dev`         | `libpgm`     |                    | `openpgm-devel`     | NO       | For ZeroMQ      |
 | libnorm[2]   | ?             | NO       | `libnorm-dev`        |              |                    |                     | YES      | For ZeroMQ      |
-| libunbound   | 1.4.16        | NO       | `libunbound-dev`     | `unbound`    | `unbound-devel`    | `unbound-devel`     | NO       | DNS resolver    |
+| libunbound   | 1.13.1        | NO       | `libunbound-dev`     | `unbound`    | `unbound-devel`    | `unbound-devel`     | NO       | DNS resolver    |
 | libsodium    | ?             | NO       | `libsodium-dev`      | `libsodium`  | `libsodium-devel`  | `libsodium-devel`   | NO       | cryptography    |
 | libunwind    | any           | NO       | `libunwind8-dev`     | `libunwind`  | `libunwind-devel`  | `libunwind-devel`   | YES      | Stack traces    |
 | liblzma      | any           | NO       | `liblzma-dev`        | `xz`         | `liblzma-devel`    | `xz-devel`          | YES      | For libunwind   |


sneurlax:
To review this, should we compare these docs to the versions used in `contrib`?

Asurar0:
@sneurlax I'm unsure to understand the question. Do you mean that a thorough review of minimal versions used in `contrib` should be done ?

sneurlax:
Sorry @Asurar0, I should have been more clear; it wasn't a question to you specifically but rather to the next reviewer.

I would think that these docs should match the actual dependencies used/vendored in `contrib`, but I'm not sure, so instead of asserting that confidently, I phrased it as a question as a way to pass it on up to someone with better domain-specific knowledge on dependencies.  Otherwise the PR looks good to me

This is not a review.

selsta:
Shouldn't the minimum version be the minimum version that can be used to compile monero? Maybe we should add a recommended column but for example Libunbound constantly puts of vulnerbility fixes so I assume there is no good "recommneded" version for it other than a recent version.

Asurar0:
> Shouldn't the minimum version be the minimum version that can be used to compile monero? Maybe we should add a recommended column but for example Libunbound constantly puts of vulnerbility fixes so I assume there is no good "recommneded" version for it other than a recent version.

Most of these vulnerabilities (libunbound and libzmq) are memory related and are in part mitigated by flags used in Cmake. I would argue however that some are not memory related, and therefore out of scope of the compiler.
For example, ZMQ <4.3.3 suffers from [CVE-2020-15166](https://www.cvedetails.com/cve/CVE-2020-15166/)

I don't see any reason to set at minimum version any packages that either have reach EOL or include non-memory vulnerabilities. No one can use the former, and no one would want to use the latter.

MrCyjaneK:
From what I can see this is only a README change, which doesn't change the fact that the code can still be build with older toolchain/with older dependencies. I don't see a huge value in this change because, as @selsta pointed out minimal version is not recommended version (well maybe for glibc it is... but this is out of the scope of this PR), it is more of a information to somebody who is willing to compile monero on a machine running older distribution.

I would consider updating contrib/depends more important because it actually changes what version of dependencies most of the people will run on their devices, and for those who build monero manually - I think that it doesn't matter to bump the version in readme if the code is still linkable against older libraries - because those usually contain charry-picked patches from distro maintainers, as can be seen in the CVE-2020-15166 example that you mention - if you use zeromq3 from debian repositories (which is on 4.3.1 currently) you will see a patch for that and other issues: https://sources.debian.org/patches/zeromq3/4.3.1-4+deb10u2/

So I vote for not artificially bumping the version numbers if the code is still compatible with older libraries.

Asurar0:
@MrCyjaneK I concur with your perspective on this one. It may be worthwhile to consider opening a separate PR to revisit and update the minimum version requirements in the contrib, ensuring they remain reasonable and aligned with current versions.

Concerning GCC 7, I would appreciate if any contributors can confirm its ability to compile monero.

tobtoht:
@Asurar0  We currently use GCC 7 for release builds, so yes.

Asurar0:
> @Asurar0 We currently use GCC 7 for release builds, so yes.

In the github build workflow in `release-v0.18` branch:

https://github.com/monero-project/monero/blob/b089f9ee69924882c5d14dd1a6991deb05d9d1cd/.github/workflows/build.yml#L15C3-L15C20

The workflow includes the installation of the [`build-essential`](https://ubuntu.pkgs.org/20.04/ubuntu-main-amd64/build-essential_12.8ubuntu1_amd64.deb.html) package, which in turn requires `gcc` and `g++`. According to my understanding, on Ubuntu 20.04, these packages correspond to [gcc version 9](https://ubuntu.pkgs.org/20.04/ubuntu-main-amd64/gcc_9.3.0-1ubuntu2_amd64.deb.html). However, I'm not familiar with Github actions so I may have misinterpreted the yaml file.

> Edit: never mind, C++17 bump hasn't made it into the release branch and building master with 7 is broken.

It seems like there are valid reasons to keep minimal dependencies as is because of linux distribution packaging. Unless GCC is wished to be raised in the future or `contrib` investigated, we can close this PR. @selsta *Recommended* column might be a solution. Otherwise, a warning could be added at the top of this section to warn users to be aware of security patches delivered by their Linux distributions.

tobtoht:
>In the github build workflow in release-v0.18 branch

For releases, we use Gitian with an Ubuntu 18.04 based image, see: https://github.com/monero-project/monero/blob/master/contrib/gitian/gitian-linux.yml#L11

Minimum GCC should be bumped to 8 for `master`, since we can't build it with 7 (unless someone wants to fix that). This is in line with #9446.

iamamyth:
A recommended column seems more harmful than helpful, as such things tend to fall out of step with reality (someone would need to monitor all the libraries and keep the recommendations up to date). Dynamic linking, plus a minimum supported version, lets users choose the best version of a library for their needs, and tends to incorporate security patches at the distro level, where appropriate.

I do think a new PR incorporating https://github.com/monero-project/monero/issues/9446 into the build process by updating the gitian base image to Ubuntu 20 and using a newer gcc version would address some of the issues raised here: The 18 base image doesn't get security updates, so there's a legitimate concern newer release builds contain too many old libraries with known vulnerabilities.

tobtoht:
@iamamyth We can't bump the base image in Gitian without affecting runtime glibc requirements.

#8929 replaces Gitian, uses GCC 12.3.0 and doesn't have this issue.

>The 18 base image doesn't get security updates

This is less of a concern for build tools. All libraries we statically link are built with `contrib/depends` and their versions can be kept up to date without affecting minimum version requirements.

sneurlax:
> ...
> I would consider updating contrib/depends more important because it actually changes what version of dependencies most of the people will run on their devices, and for those who build monero manually - I think that it doesn't matter to bump the version in readme if the code is still linkable against older libraries - because those usually contain charry-picked patches from distro maintainers, as can be seen in the CVE-2020-15166 example that you mention - if you use zeromq3 from debian repositories (which is on 4.3.1 currently) you will see a patch for that and other issues: https://sources.debian.org/patches/zeromq3/4.3.1-4+deb10u2/
> 
> So I vote for not artificially bumping the version numbers if the code is still compatible with older libraries.

contrib/depends has had a lot of work put into it to ensure that monero builds on all systems (or most, heh)

contributing there makes the most sense to me and that's why I initially thought that the docs should reflect contrib: because that's our shared workspace for ensuring builds work for all the various supported platforms

selsta:
> we can close this PR

PR #9450, opened 2024-08-23
build: fix build with Boost 1.85 and remove instances of viewkey logging

jeffro256:
1. Use `std::is_standard_layout` and `std::is_trivially_copyable` instead of `std::is_pod` for KV byte-wise serialization, which fixes compile issue for Boost UUIDs
2. Use `std::has_unique_object_representations` instead of `alignof(T) == 1` for epee byte spans and epee hex functions
3. Removed reimplementation of `std::hash` for `boost::uuids::uuid`
4. Removed `<<` operator overload for `crypto::secret_key`
5. Removed instances in code where private view key was dumped to the log in plaintext

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1091,6 +1091,7 @@ endif()
 find_package(Boost 1.58 QUIET REQUIRED COMPONENTS ${BOOST_COMPONENTS})
 add_definitions(-DBOOST_ASIO_ENABLE_SEQUENTIAL_STRAND_ALLOCATION)
 add_definitions(-DBOOST_NO_AUTO_PTR)
+add_definitions(-DBOOST_UUID_DISABLE_ALIGNMENT) # This restores UUID's std::has_unique_object_representations property
 
 set(CMAKE_FIND_LIBRARY_SUFFIXES ${OLD_LIB_SUFFIXES})
 if(NOT Boost_FOUND)
--- a/src/cryptonote_core/cryptonote_tx_utils.cpp
+++ b/src/cryptonote_core/cryptonote_tx_utils.cpp
@@ -144,7 +144,7 @@ namespace cryptonote
       crypto::key_derivation derivation = AUTO_VAL_INIT(derivation);
       crypto::public_key out_eph_public_key = AUTO_VAL_INIT(out_eph_public_key);
       bool r = crypto::generate_key_derivation(miner_address.m_view_public_key, txkey.sec, derivation);
-      CHECK_AND_ASSERT_MES(r, false, "while creating outs: failed to generate_key_derivation(" << miner_address.m_view_public_key << ", " << txkey.sec << ")");
+      CHECK_AND_ASSERT_MES(r, false, "while creating outs: failed to generate_key_derivation(" << miner_address.m_view_public_key << ", " << crypto::secret_key_explicit_print_ref{txkey.sec} << ")");
 
       r = crypto::derive_public_key(derivation, no, miner_address.m_spend_public_key, out_eph_public_key);
       CHECK_AND_ASSERT_MES(r, false, "while creating outs: failed to derive_public_key(" << derivation << ", " << no << ", "<< miner_address.m_spend_public_key << ")");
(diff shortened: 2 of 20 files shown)


vtnerd:
This _was_ a correct compile-time check, POD implied standard layout and trivial. Now you have to check for multiple things as you do here. One thing I noticed is that I didn't check for trivial copyable in `epee::span`, I'll look into upgrading that separately.

The `epee::span` check also does an `alignof(...) == 1` in a crappy way to identify padding. I'm not sure if this can be added here without breaking compilation.

jeffro256:
> POD implied standard layout and trivial

You're right, it's just that its probably too tight of property for what we need. namely, to do blob serialization, we don't care if the class is *trivial* as long as it is *trivially copyable*. 

> One thing I noticed is that I didn't check for trivial copyable in epee::span, I'll look into upgrading that separately.

We might want to change it here in this PR because Boost 1.85 also broke static assertions in other parts of the codebase: https://github.com/monero-project/monero/actions/runs/10529735156/job/29178145531?pr=9450.

jeffro256:
Okay so sorry for the spam, but it should be ready now

jeffro256:
Side note: we use a not constant time algorithm for printing some of these secrets which opens us up to cache timing attacks. Not a huge deal since we usually only hexify tx ephemeral privkeys (except in the wallet save code), and/or it can't be trivially triggered repeatedly. However, this is far from ideal.

selsta:
can confrim this solves compilation issues for me, please squash

jeffro256:
Squashed and implemented @vtnerd's suggestions for epee byte span functions

CMakeLists.txt
```diff
@@ -1091,6 +1091,7 @@ endif()
 find_package(Boost 1.58 QUIET REQUIRED COMPONENTS ${BOOST_COMPONENTS})
 add_definitions(-DBOOST_ASIO_ENABLE_SEQUENTIAL_STRAND_ALLOCATION)
 add_definitions(-DBOOST_NO_AUTO_PTR)
+add_definitions(-DBOOST_UUID_DISABLE_ALIGNMENT) # This restores UUID's std::has_unique_object_representations property
```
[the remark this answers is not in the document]
jeffro256: We need this flag for all translation units which include `boost/uuid/uuid.hpp`

selsta (approved):
Please also open against release-v0.18

jeffro256:
This PR requires C++17, which is not yet supported on the release branch

src/cryptonote_core/cryptonote_tx_utils.cpp
```diff
@@ -484,7 +484,7 @@ namespace cryptonote
           crypto::generate_ring_signature(tx_prefix_hash, boost::get<txin_to_key>(tx.vin[i]).k_image, keys_ptrs, in_contexts[i].in_ephemeral.sec, src_entr.real_output, sigs.data());
         ss_ring_s << "signatures:" << ENDL;
         std::for_each(sigs.begin(), sigs.end(), [&](const crypto::signature& s){ss_ring_s << s << ENDL;});
-        ss_ring_s << "prefix_hash:" << tx_prefix_hash << ENDL << "in_ephemeral_key: " << in_contexts[i].in_ephemeral.sec << ENDL << "real_output: " << src_entr.real_output << ENDL;
+        ss_ring_s << "prefix_hash:" << tx_prefix_hash << ENDL << "in_ephemeral_key: " << rct::sk2rct(in_contexts[i].in_ephemeral.sec) << ENDL << "real_output: " << src_entr.real_output << ENDL;
```
vtnerd, line 487: There are unwrap functions to expose the inner type without copy - unwrap(unwrap(txkey.sec)). This occurs several times throughout this PR.

jeffro256: The function `rct::sk2rct` also doesn't copy, but `unwrap` is a definitely a lot cleaner

woodser:
Builds for me as well (on the release branch). Thanks.

src/cryptonote_core/cryptonote_tx_utils.cpp
vtnerd: The function does a return by value - so it is definitely triggering the copy constructor in this situation.

vtnerd: Nevermind I missed the `&` by the function name. Yikes! This violates the aliasing rule, too bad I missed this.

woodser:
Is this generally stable enough to be used / trusted with funds?

I'm not able to build on macOS otherwise, so I'm stuck without it.

From vtnerd's last review, it looks like we're awaiting a couple more changes.

jeffro256:
It's probably safe for funds, but I'm going to push a commit very soon that makes it better (i.e. still copyless) in the scenario where we fix the strict aliasing violations with the `rct::x2rct` and `rct::rct2x` functions.

src/cryptonote_core/cryptonote_tx_utils.cpp
jeffro256: Okay newest force push drops usage of `rct::sk2rct` to convert secret keys to `rct::key` and uses `unwrap` or a wrapper called `crypto::secret_key_explicit_print_ref`. I can look into making a PR to fix the strict aliasing later

jeffro256:
Windows build failure is unrelated to this PR

selsta (approved):

hhartzer:
It would be great if there could be a new release with this feature to get Monero back into FreeBSD ports. It's currently marked as broken: https://cgit.freebsd.org/ports/commit/?id=3da4573eecd73a44684a43970822cac66177a70c

PR #9453, opened 2024-08-25
depends: freebsd_base: update to 12.3

tobtoht:
- 11.3 is missing execinfo, which is needed for rust std library build (#9440)
- bump does not violate #9446


hinto-janai (approved):
Hash matches https://lists.freebsd.org/archives/dev-commits-ports-branches/2021-December/000436.txt.

iamamyth (approved):

j-berman (approved):
Confirmed this fixes the FreeBSD build on the latest FCMP++ integration work (which expects FreeBSD >= 12)

PR #9457, opened 2024-08-26
depends: release type should be lower case

tobtoht:
current behavior prevents debug/release config vars from working.

e.g. 
https://github.com/monero-project/monero/blob/master/contrib/depends/packages/boost.mk#L9-L10

because they are expanded like this:

https://github.com/monero-project/monero/blob/master/contrib/depends/funcs.mk#L126


0xFFFC0000 (approved):
LGTM

PR #4342, opened 2018-09-06
Windows: don't preset CMAKE_SYSTEM_NAME

iDunk5400:
Let it be autodiscovered instead to avoid cmake thinking it's crosscompiling builds where host = target. This resolves a cmake configure error in MSYS2 caused by CMAKE_CROSSCOMPILING checks in #4294

PR #9466, opened 2024-08-31
Implement ElectrumWords::get_invalid_word

b4n6-b4n6:
Fixes https://github.com/monero-project/monero-gui/issues/4342

### DIFF
--- a/src/mnemonics/electrum-words.cpp
+++ b/src/mnemonics/electrum-words.cpp
@@ -371,6 +371,59 @@ namespace crypto
       return true;
     }
 
+    std::string get_invalid_word(const epee::wipeable_string &words)
+    {
+      // If there's a new language added, add an instance of it here.
+      std::vector<Language::Base*> language_instances({
+        Language::Singleton<Language::Chinese_Simplified>::instance(),
+        Language::Singleton<Language::English>::instance(),
+        Language::Singleton<Language::Dutch>::instance(),
+        Language::Singleton<Language::French>::instance(),
+        Language::Singleton<Language::Spanish>::instance(),
+        Language::Singleton<Language::German>::instance(),
+        Language::Singleton<Language::Italian>::instance(),
+        Language::Singleton<Language::Portuguese>::instance(),
+        Language::Singleton<Language::Japanese>::instance(),
+        Language::Singleton<Language::Russian>::instance(),
+        Language::Singleton<Language::Esperanto>::instance(),
+        Language::Singleton<Language::Lojban>::instance(),
+        Language::Singleton<Language::EnglishOld>::instance()
+      });
+
+      std::vector<epee::wipeable_string> seed;
+      words.split(seed);
+
+      for (
+        std::vector<epee::wipeable_string>::const_iterator seed_i = seed.begin();
+        seed_i != seed.end();
+        seed_i++
+      )
+      {
+        bool has_any = false;
+
+        for (
+          std::vector<Language::Base*>::iterator lang_i = language_instances.begin();
+          lang_i != language_instances.end() && !has_any;
+          lang_i++
+        )
+        {
+          auto &word_map = (*lang_i)->get_word_map();
+
+          if (word_map.count(*seed_i) != 0)
+          {
+            has_any = true;
+            break;
+          }
+        }
+
+        if (!has_any) {
+          return std::string(seed_i->data(), seed_i->size());
+        }
+      }
+
+      return "";
+    }
+
     /*!
      * \brief Converts bytes (secret key) to seed words.
      * \param  src           Secret key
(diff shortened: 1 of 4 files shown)


b4n6-b4n6:
Screenshot from 2024-08-31 11-00-19
Screenshot from 2024-08-31 11-00-43

b4n6-b4n6:
Old error message is displayed on checksum failure

Screenshot from 2024-08-31 11-25-30
Screenshot from 2024-08-31 11-25-24

src/mnemonics/electrum-words.cpp
```diff
@@ -371,6 +371,59 @@ namespace crypto
       return true;
     }
 
+    std::string get_invalid_word(const epee::wipeable_string &words)
+    {
+      // If there's a new language added, add an instance of it here.
```
sneurlax, line 376: It might be worth the additional diff to find where else language lists are defined and leave a comment directing devs here for language-related work

sneurlax (approved):
These changes look good to me, I'd just leave an additional comment wherever else `Language`s are defined directing devs working on languages here

or some other solution which uses a general list of the `Language`s

sneurlax:
Closes #4342

sneurlax:
... but ... I actually have to recheck if this is really needed.  I forgot that only the first 3 letters really matter, does this check that as well?

b4n6-b4n6:
> ... but ... I actually have to recheck if this is really needed. I forgot that only the first 3 letters really matter...
> 
> This might still be helpful

I think this is needed as in practice it is tedious to identify misspelt word when writing words off paper wallet

b4n6-b4n6:
> > ... but ... I actually have to recheck if this is really needed. I forgot that only the first 3 letters really matter...
> > This might still be helpful
> 
>     1. I think this is needed as in practice it is tedious to identify misspelt word when writing words off paper wallet
> 
>     2. 24-seed word are not used in practice anymore, right? if they are used in practice then a misspelling (in the first 3-4 letters) can derive wrong word and result in an new empty wallet

b4n6-b4n6:
?

src/wallet/api/wallet.cpp
```diff
@@ -749,7 +749,12 @@ bool WalletImpl::recover(const std::string &path, const std::string &password, c
     crypto::secret_key recovery_key;
     std::string old_language;
     if (!crypto::ElectrumWords::words_to_bytes(seed, recovery_key, old_language)) {
-        setStatusError(tr("Electrum-style word list failed verification"));
+        std::string invalid_word = crypto::ElectrumWords::get_invalid_word(seed);
+        if (invalid_word != "") {
+            setStatusError((boost::format(tr("Invalid word %s")) % ("'" + invalid_word + "'")).str());
```
selsta, line 754: What do you think of adding `Electrum-style word list failed verification: ` before invalid word to make it clear what failed?

b4n6-b4n6:
I have rebased with the new changes

selsta (approved):

selsta:
The reason why I didn't add this to the merge list yet is because there is a theoretical edge case now where it would display the incorrect invalid word.

If someone enters the seed with only the first three letters (something that is allowed by the software, only the first three letters get validated) and there is an issue with a word after the first one, it would display the incorrect invalid word.

Since this patch only impacts the GUI and there is a overall UX improvement, I think this should be merged.

Alternatively, the patch could be changed to make the invalid word comparison only to the first 3 letters, which would fix this edge case.

What do you think @b4n6-b4n6 ?

b4n6-b4n6:
> only the first three letters (

I'd like to hear the reasoning for the 3-letter (was it 3 or 4?) approach and why it is supported in the first place. Depending on this answer I might consider it, but as of now, as we are developing software for humans (and not us geeks), it feels out of place and unintuitive

b4n6-b4n6:
> The reason why I didn't add this to the merge list yet is because there is a theoretical edge case now where it would display the incorrect invalid word.
> 
> If someone enters the seed with only the first three letters (something that is allowed by the software, only the first three letters get validated) and there is an issue with a word after the first one, it would display the incorrect invalid word.
> 
> Since this patch only impacts the GUI and there is a overall UX improvement, I think this should be merged.
> 
> Alternatively, the patch could be changed to make the invalid word comparison only to the first 3 letters, which would fix this edge case.
> 
> What do you think @b4n6-b4n6 ?

Another thing to note is that if we were to only check 3 letters of each word, there can be cases where, we could end up with "invalid checksum" errors while "invalid word" error should have been displayed instead

nahuhh:
> for example: spell "amnesty" as "amensty" -> gets interpreted as "amen"/"amended" instead

Is this supposed to be suggesting the correct word to the user?

selsta:
> I'd like to hear the reasoning for the 3-letter (was it 3 or 4?) approach and why it is supported in the first place.

Since all words have unique first 3 letters, it was the minimum amount of computation that is necessary to verify a seed.

It has its down and upsides, and this PR just has different behaviour which does not fully match. But since its focused on the GUI, which is targeted at the average user, it makes sense to keep the current behaviour.

PR #9470, opened 2024-09-04
build: drop support for 32-bit windows target

tobtoht:
Discussion: https://libera.monerologs.net/monero-research-lab/20240904#c422227-c422235
Context: https://github.com/monero-project/monero/pull/9436#pullrequestreview-2278670854

The last commercially available x86 processors were released [over 18 years ago](https://en.wikipedia.org/wiki/List_of_Intel_processors#Intel_Core). Windows 11 [requires 64-bit](https://www.microsoft.com/en-us/windows/windows-11-specifications).

Article from Nov 2021:

>The first 32-bit Windows version in the charts is Windows 7 with a share of 0.22 percent, followed by its Windows 10 sibling with 0.09 percent.

https://news.softpedia.com/news/new-data-shows-windows-32-bit-is-becoming-a-thing-of-the-past-534336.shtml

### DIFF
--- a/.github/workflows/depends.yml
+++ b/.github/workflows/depends.yml
@@ -38,9 +38,6 @@ jobs:
           - name: "ARM v8"
             host: "aarch64-linux-gnu"
             packages: "g++-aarch64-linux-gnu"
-          - name: "i686 Win"
-            host: "i686-w64-mingw32"
-            packages: "g++-mingw-w64-i686"
           - name: "i686 Linux"
             host: "i686-pc-linux-gnu"
             packages: "g++-multilib"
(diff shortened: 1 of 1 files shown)


sneurlax (approved):
The changes themselves look good to me.

As to the concept, my only opinion is that the first step to opposing it would be to find a 32-bit Windows maintainer.  If there aren't devs dedicated to this platform to take up that task, then it should be dropped.  Dropping it might bring those devs out of the woodwork later and they can resolve any issues with it then, right?

0xFFFC0000 (approved):
LGTM

iamamyth:
> the first step to opposing it would be to find a 32-bit Windows maintainer

Supporting a platform where both the hardware and software are increasingly unavailable requires not just a maintainer, but others to validate their work. Are there 2-3 people with the requisite configurations willing to put in that effort? How long will CI and build platforms support win32? Until the next hardfork, win32 users can run the existing binaries, so, by the time this decision drops them from the network, the platform would be even more irrelevant.

One-horse-wagon:
It is estimated as of 2024, only 4% of windows users use 32 bit systems.  They are legacy systems with the total number constantly decreasing.   I would think all Monero users anywhere, have access to 64 bit systems, even if they are using a 32 bit system.

selsta (approved):

selsta (approved):

PR #9473, opened 2024-09-05
cryptonote_core: unrestricted rpc to respect packet size limit

0xFFFC0000:
When txpool is greater than 100mb, unrestricted rpc cannot connect. Restricted rpc remains functional due to batching into 100tx per call. Here, were batching into 90% of the packet size limit

### DIFF
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -665,22 +665,35 @@ namespace cryptonote
     return true;
   }
   //------------------------------------------------------------------
-  bool tx_memory_pool::get_transactions_info(const std::vector<crypto::hash>& txids, std::vector<std::pair<crypto::hash, tx_details>>& txs, bool include_sensitive) const
+  bool tx_memory_pool::get_transactions_info(const epee::span<const crypto::hash> txids, std::vector<std::pair<crypto::hash, tx_details>>& txs, bool include_sensitive, size_t cumul_txblob_size_limit, size_t max_tx_count) const
   {
     CRITICAL_REGION_LOCAL(m_transactions_lock);
     CRITICAL_REGION_LOCAL1(m_blockchain);
 
     txs.clear();
 
-    for (const auto &it: txids)
-    {
-      tx_details details;
-      bool success = get_transaction_info(it, details, include_sensitive, true/*include_blob*/);
-      if (success)
+    const size_t max_allowed =
+      max_tx_count ? std::min(max_tx_count, txids.size()) : txids.size();
+
+    txs.reserve(std::min<size_t>(max_allowed, 1000000)); // reserve limited to min(1 million, max_allowed)
+    size_t cumul_txblob_size = 0;
+
+    for (size_t i = 0; i < txids.size() && txs.size() < max_allowed; ++i)
       {
-        txs.push_back(std::make_pair(it, std::move(details)));
+        const crypto::hash &it{txids[i]};
+        tx_details details;
+        const bool success = get_transaction_info(it, details, include_sensitive, true /*include_blob*/);
+        if (!success)
+          continue;
+
+        if (cumul_txblob_size_limit &&
+            cumul_txblob_size + details.blob_size >= cumul_txblob_size_limit)
+          continue;
+
+        cumul_txblob_size += details.blob_size;
+        txs.emplace_back(it, std::move(details));
       }
-    }
+
     return true;
   }
   //---------------------------------------------------------------------------------
@@ -952,7 +965,7 @@ namespace cryptonote
     }, false, category);
   }
   //------------------------------------------------------------------
-  bool tx_memory_pool::get_pool_info(time_t start_time, bool include_sensitive, size_t max_tx_count, std::vector<std::pair<crypto::hash, tx_details>>& added_txs, std::vector<crypto::hash>& remaining_added_txids, std::vector<crypto::hash>& removed_txs, bool& incremental) const
+  bool tx_memory_pool::get_pool_info(time_t start_time, bool include_sensitive, size_t max_tx_count, std::vector<std::pair<crypto::hash, tx_details>>& added_txs, std::vector<crypto::hash>& remaining_added_txids, std::vector<crypto::hash>& removed_txs, bool& incremental, size_t cumul_limit_size) const
   {
     CRITICAL_REGION_LOCAL(m_transactions_lock);
     CRITICAL_REGION_LOCAL1(m_blockchain);
@@ -983,47 +996,76 @@ namespace cryptonote
     remaining_added_txids.clear();
     removed_txs.clear();
 
-    std::vector<crypto::hash> txids;
     if (!incremental)
-    {
       LOG_PRINT_L2("Giving back the whole pool");
-      // Give back the whole pool in 'added_txs'; because calling 'get_transaction_info' right inside the
-      // anonymous method somehow results in an LMDB error with transactions we have to build a list of
-      // ids first and get the full info afterwards
-      get_transaction_hashes(txids, include_sensitive);
-      if (txids.size() > max_tx_count)
+
+    // If incremental, handle removed TXIDs first since it's important that txs are removed
+    // from synchronizers' pools, and we need need to estimate how much space we have left to
+    // request full-bodied txs
+    if (incremental)
+    {
+      std::multimap<time_t, removed_tx_info>::const_iterator rit = m_removed_txs_by_time.lower_bound(start_time);
+      while (rit != m_removed_txs_by_time.end())
       {
-        remaining_added_txids = std::vector<crypto::hash>(txids.begin() + max_tx_count, txids.end());
-        txids.erase(txids.begin() + max_tx_count, txids.end());
+        if (include_sensitive || !rit->second.sensitive)
+        {
+          removed_txs.push_back(rit->second.txid);
+        }
+        ++rit;
       }
-      get_transactions_info(txids, added_txs, include_sensitive);
-      return true;
     }
 
-    // Give back incrementally, based on time of entry into the map
+    // Build a list of TXIDs that we want information for eventually.  When `!incremental`, we might
+    // collect TXIDs that have already left the pool which will cause a followup RPC call that
+    // fetches non-existent txs, but that's *okay* since a non-incremental caller probably doesn't
+    // care about being incredibly efficient. This problem is also unavoidable anyways if we need to
+    // ever return anything inside `remaining_added_txids`, since those txs might have left the mempool
+    // by the time the followup RPC call reaches the daemon
+    std::vector<crypto::hash> txids;
+    txids.reserve(m_added_txs_by_id.size());
     for (const auto &pit : m_added_txs_by_id)
     {
-      if (pit.second >= start_time)
+      const bool relevant_txid{!incremental || pit.second >= start_time};
+      if (relevant_txid)
         txids.push_back(pit.first);
     }
-    get_transactions_info(txids, added_txs, include_sensitive);
-    if (added_txs.size() > max_tx_count)
+
+    // Estimate max cumulative size left for full tx blobs
+    const size_t removed_txids_clawback{32 * removed_txs.size()};
+    const size_t remaining_txids_clawback{32 * txids.size()};
+    const size_t added_tx_txid_clawback(32 * txids.size());
+    const size_t total_clawback{removed_txids_clawback + remaining_txids_clawback + added_tx_txid_clawback};
+    const size_t cumulative_txblob_size_limit{cumul_limit_size > total_clawback ? cumul_limit_size - total_clawback : 0};
+
+    // Perform TX info fetch, limited to max_tx_count and cumulative_txblob_size_limit
+    if (cumulative_txblob_size_limit && !txids.empty() && max_tx_count)
     {
-      remaining_added_txids.reserve(added_txs.size() - max_tx_count);
-      for (size_t i = max_tx_count; i < added_txs.size(); ++i)
-        remaining_added_txids.push_back(added_txs[i].first);
-      added_txs.erase(added_txs.begin() + max_tx_count, added_txs.end());
+      if (!get_transactions_info(epee::to_span(txids),
+                                 added_txs,
+                                 include_sensitive,
+                                 cumulative_txblob_size_limit,
+                                 max_tx_count))
+        return false;
     }
 
-    std::multimap<time_t, removed_tx_info>::const_iterator rit = m_removed_txs_by_time.lower_bound(start_time);
-    while (rit != m_removed_txs_by_time.end())
+    // Populate `remaining_added_txids` with all TXIDs in `txids` and not in `added_txs`
+    if (added_txs.size() < txids.size())
     {
-      if (include_sensitive || !rit->second.sensitive)
-      {
-        removed_txs.push_back(rit->second.txid);
+      // 1. Create a set of added IDs for fast lookup
+      std::unordered_set<crypto::hash> added_set;
+      for (const auto& tx : added_txs) {
+        added_set.insert(tx.first);
       }
-      ++rit;
+
+      // 2. Iterate the original list. If not in set, it is "remaining".
+      remaining_added_txids.reserve(txids.size() - added_txs.size());
+      for (const auto& id : txids) {
+        if (added_set.find(id) == added_set.end()) {
+          remaining_added_txids.push_back(id);
+        }
+      }
     }
+
     return true;
   }
   //------------------------------------------------------------------
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -258,7 +258,7 @@ namespace cryptonote
             return false;
 
           m_blockchain.add_txpool_tx(id, blob, meta);
-          add_tx_to_transient_lists(id, fee / (double)(tx_weight ? tx_weight : 1), receive_time);
+          add_tx_to_transient_lists(id, fee / (double)(tx_weight ? tx_weight : 1), receive_time, !meta.matches(relay_category::broadcasted));
           lock.commit();
         }
         catch (const std::exception &e)
@@ -329,7 +329,7 @@ namespace cryptonote
 
           m_blockchain.remove_txpool_tx(id);
           m_blockchain.add_txpool_tx(id, blob, meta);
-          add_tx_to_transient_lists(id, meta.fee / (double)(tx_weight ? tx_weight : 1), receive_time);
+          add_tx_to_transient_lists(id, meta.fee / (double)(tx_weight ? tx_weight : 1), receive_time, !meta.matches(relay_category::broadcasted));
         }
         lock.commit();
         tvc.m_added_to_pool = !existing_tx;
@@ -912,7 +912,7 @@ namespace cryptonote
 
           if (was_just_broadcasted)
             // Make sure the tx gets re-added with an updated time
-            add_tx_to_transient_lists(hash, meta.fee / (double)meta.weight, std::chrono::system_clock::to_time_t(now));
+            add_tx_to_transient_lists(hash, meta.fee / (double)meta.weight, std::chrono::system_clock::to_time_t(now), false/*sensitive*/);
         }
       }
       catch (const std::exception &e)
@@ -1025,7 +1025,9 @@ namespace cryptonote
     txids.reserve(m_added_txs_by_id.size());
     for (const auto &pit : m_added_txs_by_id)
     {
-      const bool relevant_txid{!incremental || pit.second >= start_time};
+      if (!include_sensitive && pit.second.sensitive)
+        continue;
+      const bool relevant_txid{!incremental || pit.second.receive_time >= start_time};
       if (relevant_txid)
         txids.push_back(pit.first);
     }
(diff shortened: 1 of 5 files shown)


0xFFFC0000:
Windows build failure is unrelated to this PR:

```
error: could not open file mingw-w64-x86_64-protobuf-21.9-1-any.pkg.tar.zst: Unrecognized archive format
```

Mac build failure is unrelated to this PR and related to `boost::uuid` issue we have at the moment.

0xFFFC0000:
Fix all the issues raised here [1]. 
Re-based to latest master.

1. https://github.com/spackle-xmr/monero/pull/32

src/rpc/core_rpc_server.cpp
```diff
@@ -790,7 +744,54 @@ namespace cryptonote
...
+    if (get_pool)
+    {
+      const bool restricted = m_restricted && ctx;
+      const bool request_has_rpc_origin = ctx != NULL;
+      const bool allow_sensitive = !request_has_rpc_origin || !restricted;
+      const size_t max_tx_count = restricted ? RESTRICTED_TRANSACTIONS_COUNT : std::numeric_limits<size_t>::max();
+
+      bool incremental;
+      std::vector<std::pair<crypto::hash, tx_memory_pool::tx_details>> added_pool_txs;
+      bool success = m_core.get_pool_info((time_t)req.pool_info_since, allow_sensitive, max_tx_count, added_pool_txs, res.remaining_added_pool_txids, res.removed_pool_txids, incremental, (LEVIN_DEFAULT_MAX_PACKET_SIZE * 0.9) - cumul_block_data_size);
```
j-berman: Looks like the endpoint should fail to return pool info if `cuml_block_data_size > (LEVIN_DEFAULT_MAX_PACKET_SIZE * 0.9)`

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1016,47 +1020,80 @@ namespace cryptonote
...
+    // Estimate max cumulative size left for full tx blobs
+    const size_t removed_txids_clawback{32 * removed_txs.size()};
+    const size_t remaining_txids_clawback{32 * txids.size()};
+    const size_t added_tx_txid_clawback(32 * txids.size());
+    const size_t total_clawback{removed_txids_clawback + remaining_txids_clawback + added_tx_txid_clawback};
+    const size_t cumulative_txblob_size_limit{cumul_limit_size > total_clawback ? cumul_limit_size - total_clawback : 0};
+
+    // Perform TX info fetch, limited to max_tx_count and cumulative_txblob_size_limit
+    const size_t max_full_tx_fetch_count{std::min(txids.size(), max_tx_count)};
+    if (cumulative_txblob_size_limit && max_full_tx_fetch_count)
```
j-berman: Is including `cumulative_txblob_size_limit` in this if statement an accident? Looks like we want all those tx infos even when `cumul_limit_size` is 0

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1016,47 +1020,80 @@ namespace cryptonote
...
+        return false;
     }
 
-    std::multimap<time_t, removed_tx_info>::const_iterator rit = m_removed_txs_by_time.lower_bound(start_time);
-    while (rit != m_removed_txs_by_time.end())
+    // Populate `remaining_added_txids` with all TXIDs in `txids` and not in `added_txs`
+    if (added_txs.size() < txids.size())
     {
-      if (include_sensitive || !rit->second.sensitive)
+      const size_t num_remaining{added_txs.size() - max_tx_count};
```
j-berman:
```suggestion
      const size_t num_remaining{txids.size() - added_txs.size()};
```

To avoid underflow risk

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1016,47 +1020,80 @@ namespace cryptonote
...
+    // Perform TX info fetch, limited to max_tx_count and cumulative_txblob_size_limit
+    const size_t max_full_tx_fetch_count{std::min(txids.size(), max_tx_count)};
+    if (cumulative_txblob_size_limit && max_full_tx_fetch_count)
     {
-      remaining_added_txids.reserve(added_txs.size() - max_tx_count);
-      for (size_t i = max_tx_count; i < added_txs.size(); ++i)
-        remaining_added_txids.push_back(added_txs[i].first);
-      added_txs.erase(added_txs.begin() + max_tx_count, added_txs.end());
+      const epee::span<const crypto::hash> txid_fetch_span{&txids[0], max_full_tx_fetch_count};
+      if (!get_transactions_info(txid_fetch_span, added_txs, include_sensitive, cumulative_txblob_size_limit))
```
j-berman: It looks like instead you want to pass in `txids` and `max_tx_count` to `get_transactions_info`, in addition to `cumulative_txblob_size_limit`. It's possible some large txs won't get added, but we're still below both `max_tx_count` AND `cumulative_txblob_size_limit`, and still have some subsequent smaller txids that could get added

I also think this would be simpler than passing in a span here

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1016,47 +1020,80 @@ namespace cryptonote
...
     {
-      if (include_sensitive || !rit->second.sensitive)
+      const size_t num_remaining{added_txs.size() - max_tx_count};
+      remaining_added_txids.reserve(num_remaining);
+
+      // This iteration code assumes that the get_transactions_info() method A) returns elements in
+      // the same order as they are passed in (skipping failures) and B) does not return TXIDs which
+      // don't exist in the input. If these assumptions don't hold then remaining_added_txids will
+      // be wrong, but it shouldn't cause any undefined behavior outside of that and should terminate
+      // in a short finite time period O(N) with N = txids.size()
```
j-berman: I think this is an unnecessary stipulation and is hard to reason through its correctness imo. Suggestion: construct an unordered set of the added txids in place here, then just iterate over txids and add to remaining if not already included in the set of added txids. The marginal perf hit here seems warranted to me to guarantee expected behavior

j-berman (commented):
Avoiding tripping the packet size limit (instead of just limiting to a fixed max tx count) makes sense

src/cryptonote_core/tx_pool.cpp
0xFFFC0000: That is the intended behaviour. A few lines above we do: 

```
const size_t cumulative_txblob_size_limit{cumul_limit_size > total_clawback ? cumul_limit_size - total_clawback : 0};
```

src/cryptonote_core/tx_pool.cpp
0xFFFC0000: Thanks. Fixed it.

0xFFFC0000:
Fixed [1, 2, 3, 5], still thinking about 4. Rebased to master since there was a conflict to merge. 

1. https://github.com/monero-project/monero/pull/9473#discussion_r2283858417
2. https://github.com/monero-project/monero/pull/9473#discussion_r2283880938
3. https://github.com/monero-project/monero/pull/9473#discussion_r2283883939
4. https://github.com/monero-project/monero/pull/9473#discussion_r2283900917
5. https://github.com/monero-project/monero/pull/9473#discussion_r2283901917

src/rpc/core_rpc_server.cpp
```diff
@@ -806,7 +760,57 @@ namespace cryptonote
...
+      const bool restricted = m_restricted && ctx;
+      const bool request_has_rpc_origin = ctx != NULL;
+      const bool allow_sensitive = !request_has_rpc_origin || !restricted;
+      const size_t max_tx_count = restricted ? RESTRICTED_TRANSACTIONS_COUNT : std::numeric_limits<size_t>::max();
+
+      bool incremental;
+      std::vector<std::pair<crypto::hash, tx_memory_pool::tx_details>> added_pool_txs;
+      const size_t limit = LEVIN_DEFAULT_MAX_PACKET_SIZE * 0.9;
+      // If blocks already take up all the space, allow 0 space for pool txs
+      const size_t pool_limit = (cumul_block_data_size < limit) ? (limit - cumul_block_data_size) : 0;
```
j-berman: I think there's still a bit of an issue here

Assume `find_blockchain_supplement` above this section returns 99.99mb of cumulative data.

Then `get_pool_info` executes here and fills up the response larger than the packet size limit (it's possible for just the hashes push it over the limit).

We don't want that behavior because the connection terminates in that case.

My suggestion [before](https://github.com/monero-project/monero/pull/9473#discussion_r2283858417) was to just not return *any* pool info if `cuml_block_data_size > (LEVIN_DEFAULT_MAX_PACKET_SIZE * 0.9)` because [wallets can handle that ok](https://github.com/monero-project/monero/blob/48ad374b0d6d6e045128729534dc2508e6999afe/src/wallet/wallet2.cpp#L3199-L3209). But then there would also be a question of should the response be a success response or failure response (because the wallet *did* request pool info).

Another option is to reduce the `FIND_BLOCKCHAIN_SUPPLEMENT_MAX_SIZE` maybe to something like 90mb as well (or to reduce [this `max_size` param](https://github.com/monero-project/monero/blob/48ad374b0d6d6e045128729534dc2508e6999afe/src/cryptonote_core/blockchain.cpp#L2768) to 90mb when the caller also wants to get the pool info from `on_get_blocks`). That way we'll have confidence that `get_pool_info` can definitely execute and fill up the response before tripping the packet size limit.

I personally think only reducing that `max_size` param when the caller *also* wants the pool is a good solution here. It would require a new param to the `find_blockchain_supplement` function used in `on_get_blocks` and some logic for setting the param in `on_get_blocks`, which seems sane to me.

src/cryptonote_core/tx_pool.cpp
j-berman: I see what you mean here

src/cryptonote_core/tx_pool.cpp
j-berman: Assume Tx A is 100kb, Tx B is 147kb, and Tx C is 5kb.

Assume `max_tx_count` is 2.

Assume `cumulative_txblob_size_limit` is 150kb.

Then `max_full_tx_fetch_count` would be 2, and so the span of [Tx A, Tx B] would be passed into `get_transactions_info`.

`get_transactions_info` would then include Tx A and ignore Tx B in `added_txs`. It could have also included Tx C in the response though, minimizing trips the wallet needs to do. But it didn't because the span cut off Tx C.

I think it would be pretty simple to solve for this case by just passing the entire `txids` vector into `get_transactions_info` instead of the span, in addition to passing in `max_tx_count` as a param, and then iterate over all `txids` until `added_txs` is either at `max_tx_count` or adding the next tx would push it over the blob size limit.

I would accept if you argue this is unnecessary. I do think it would make the code as is marginally simpler and more optimal though.

0xFFFC0000: I think I am missing something here. We are doing exactly that: 

```
bool tx_memory_pool::get_transactions_info(const epee::span<const crypto::hash> txids, std::vector<std::pair<crypto::hash, tx_details>>& txs, bool include_sensitive, size_t cumul_txblob_size_limit) const
  {
    CRITICAL_REGION_LOCAL(m_transactions_lock);

    txs.clear();
    txs.reserve(std::min<size_t>(txids.size(), 1000000)); // reserve limited to 1 million
    size_t cumul_txblob_size = 0;

    for (size_t i = 0; i < txids.size(); ++i)
    {
      const crypto::hash &it{txids[i]};
      tx_details details;
      bool success = get_transaction_info(it, details, include_sensitive, true/*include_blob*/);
      if (success && (!cumul_txblob_size_limit || (cumul_txblob_size + details.blob_size < cumul_txblob_size_limit)))
      {
        cumul_txblob_size += details.blob_size;
        txs.push_back(std::make_pair(it, std::move(details)));
      }
    }
    return true;
  }
```

We go over all txs, and if `(cumul_txblob_size + details.blob_size < cumul_txblob_size_limit)` satisfied, we will add it to txs.

0xFFFC0000:
This one [1], fixed https://github.com/monero-project/monero/pull/9473/commits/0b2a407355dca5698cc8ea063076306476f85701

1. https://github.com/monero-project/monero/pull/9473#discussion_r2646186179

src/cryptonote_core/tx_pool.cpp
0xFFFC0000: Fixed. Thanks

src/rpc/core_rpc_server.cpp
0xFFFC0000: Fixed this here: https://github.com/monero-project/monero/pull/9473/changes/374a90ef2e5c09f043b5e3e36551e4477f5dd7a9

Thanks.

0xFFFC0000: The problem with reducing `FIND_BLOCKCHAIN_SUPPLEMENT_MAX_SIZE` is that it is a large cross-layer change with the possibility of introducing new bugs for this extreme corner case. 

So IMHO, this solution is a good trade-off.

0xFFFC0000:
CI failure is not related to the PR.

j-berman (approved):

nahuhh:
On stressnet, 500mb txpool, unrestricted rpc, wallet-cli

Good:
~~the wallet loads! Previously it would hang for a long time and then disconnect.~~ Update: with a 700mb txpool, using wallet-rpc, im having to switch back to restricted rpc << **edit, i was just impatient**

observation on log lvl 2:
it prints log showing that its asking for 100 transactions. Is this intended?
```
2026-01-08 18:07:46.134     77d5d5515f80        DEBUG   wallet.wallet2  src/wallet/wallet2.cpp:2960     Reading pool txs                   
2026-01-08 18:07:46.135     77d5d5515f80        DEBUG   serialization   src/wallet/node_rpc_proxy.cpp:316       asking for 100 transactions
```

navidR:
Update should address: https://github.com/monero-project/monero/pull/9473#issuecomment-3724493045

src/wallet/node_rpc_proxy.cpp
```diff
@@ -323,7 +337,37 @@ boost::optional<std::string> NodeRPCProxy::get_transactions(const std::vector<cr
...
+      continue;
+    }
+
+    // Transport failure with room to shrink: shrink and retry, but don't set permanent ceiling.
+    if (!m_restricted && !r && request_size > restricted_slice_size)
+    {
+      m_rpc_batch_size = std::max(restricted_slice_size, request_size / 2);
+      MDEBUG("Transport failure, reducing gettransactions batch size to " << m_rpc_batch_size);
+      continue;
+    }
```
j-berman: Looks like it would make sense for the daemon to return explicit status codes that indicate if the client should retry. Doesn't seem like a great mechanism to continue shrink and retry until `request_size <= restricted_slice_size`

j-berman (commented):
I'm ok with adapting request sizes for txs client-side, to aim to improve the slow download when the pool is huge. I think it can be done in a follow-up PR, though. I think the more pressing fix is for the borked unrestricted RPC, which has seen good testing / review already

So my vote is to revert this PR to the state that was reviewed/tested, and adapting request sizes can be done in a follow-up PR

navidR:
> I'm ok with adapting request sizes for txs client-side, to aim to improve the slow download when the pool is huge. I think it can be done in a follow-up PR, though. I think the more pressing fix is for the borked unrestricted RPC, which has seen good testing / review already
> 
> So my vote is to revert this PR to the state that was reviewed/tested, and adapting request sizes can be done in a follow-up PR

I reverted the adaptive ceiling. So I think we are good to go.

j-berman (approved):

j-berman:
@selsta caught an issue where sensitive txs can be included in the remaining txs list. Here is a patch to prevent that:

```patch
diff --git a/src/cryptonote_core/tx_pool.cpp b/src/cryptonote_core/tx_pool.cpp
index d72c92ffd..1490d53b5 100644
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -258,7 +258,7 @@ namespace cryptonote
             return false;
 
           m_blockchain.add_txpool_tx(id, blob, meta);
-          add_tx_to_transient_lists(id, fee / (double)(tx_weight ? tx_weight : 1), receive_time);
+          add_tx_to_transient_lists(id, fee / (double)(tx_weight ? tx_weight : 1), receive_time, !meta.matches(relay_category::broadcasted));
           lock.commit();
         }
@@ -329,7 +329,7 @@ namespace cryptonote
 
           m_blockchain.remove_txpool_tx(id);
           m_blockchain.add_txpool_tx(id, blob, meta);
-          add_tx_to_transient_lists(id, meta.fee / (double)(tx_weight ? tx_weight : 1), receive_time);
+          add_tx_to_transient_lists(id, meta.fee / (double)(tx_weight ? tx_weight : 1), receive_time, !meta.matches(relay_category::broadcasted));
         }
         lock.commit();
         tvc.m_added_to_pool = !existing_tx;
@@ -912,7 +912,7 @@ namespace cryptonote
 
           if (was_just_broadcasted)
             // Make sure the tx gets re-added with an updated time
-            add_tx_to_transient_lists(hash, meta.fee / (double)meta.weight, std::chrono::system_clock::to_time_t(now));
+            add_tx_to_transient_lists(hash, meta.fee / (double)meta.weight, std::chrono::system_clock::to_time_t(now), false/*sensitive*/);
         }
       }
@@ -1025,7 +1025,9 @@ namespace cryptonote
     txids.reserve(m_added_txs_by_id.size());
     for (const auto &pit : m_added_txs_by_id)
     {
-      const bool relevant_txid{!incremental || pit.second >= start_time};
+      if (!include_sensitive && pit.second.sensitive)
+        continue;
+      const bool relevant_txid{!incremental || pit.second.receive_time >= start_time};
       if (relevant_txid)
         txids.push_back(pit.first);
     }
@@ -1822,14 +1824,14 @@ namespace cryptonote
     return n_removed;
   }
-  void tx_memory_pool::add_tx_to_transient_lists(const crypto::hash& txid, double fee, time_t receive_time)
+  void tx_memory_pool::add_tx_to_transient_lists(const crypto::hash& txid, double fee, time_t receive_time, bool sensitive)
   {
 
     time_t now = time(NULL);
-    const std::unordered_map<crypto::hash, time_t>::iterator it = m_added_txs_by_id.find(txid);
+    const std::unordered_map<crypto::hash, added_tx_info>::iterator it = m_added_txs_by_id.find(txid);
     if (it == m_added_txs_by_id.end())
     {
-       m_added_txs_by_id.insert(std::make_pair(txid, now));
+       m_added_txs_by_id[txid] = {now, sensitive};
     }
     else
     {
@@ -1837,7 +1839,8 @@ namespace cryptonote
       // phase of Dandelion++ and now is in the "fluff" phase i.e. got broadcasted: We have to set
       // a new time for clients that are not allowed to see sensitive txs to make sure they will
       // see it now if they query incrementally
-      it->second = now;
+      it->second.receive_time = now;
+      it->second.sensitive = sensitive;
 
       auto sorted_it = find_tx_in_sorted_container(txid);
       if (sorted_it == m_txs_by_fee_and_receive_time.end())
@@ -1876,7 +1879,7 @@ namespace cryptonote
       m_txs_by_fee_and_receive_time.erase(sorted_it);
     }
 
-    const std::unordered_map<crypto::hash, time_t>::iterator it = m_added_txs_by_id.find(txid);
+    const std::unordered_map<crypto::hash, added_tx_info>::iterator it = m_added_txs_by_id.find(txid);
     if (it != m_added_txs_by_id.end())
     {
        m_added_txs_by_id.erase(it);
@@ -1961,7 +1964,7 @@ namespace cryptonote
           MFATAL("Failed to insert key images from txpool tx");
           return false;
         }
-        add_tx_to_transient_lists(txid, meta.fee / (double)meta.weight, meta.receive_time);
+        add_tx_to_transient_lists(txid, meta.fee / (double)meta.weight, meta.receive_time, !meta.matches(relay_category::broadcasted));
         m_txpool_weight += meta.weight;
         return true;
       }, true, relay_category::all);
diff --git a/src/cryptonote_core/tx_pool.h b/src/cryptonote_core/tx_pool.h
index fbdf7b26d..745876eec 100644
--- a/src/cryptonote_core/tx_pool.h
+++ b/src/cryptonote_core/tx_pool.h
@@ -624,7 +624,7 @@ namespace cryptonote
      */
     void prune(size_t bytes = 0);
 
-    void add_tx_to_transient_lists(const crypto::hash& txid, double fee, time_t receive_time);
+    void add_tx_to_transient_lists(const crypto::hash& txid, double fee, time_t receive_time, bool sensitive);
     void remove_tx_from_transient_lists(const cryptonote::sorted_tx_container::iterator& sorted_it, const crypto::hash& txid, bool sensitive);
     void track_removed_tx(const crypto::hash& txid, bool sensitive);
 
@@ -660,8 +660,14 @@ private:
 
     std::atomic<uint64_t> m_cookie; //!< incremented at each change
 
+    struct added_tx_info
+    {
+      time_t receive_time;
+      bool sensitive;
+    };
+
     // Info when transactions entered the pool, accessible by txid
-    std::unordered_map<crypto::hash, time_t> m_added_txs_by_id;
+    std::unordered_map<crypto::hash, added_tx_info> m_added_txs_by_id;
 
     // Info at what time the pool started to track the adding of transactions
     time_t m_added_txs_start_time;

```

jeffro256:
@navidR ping for above

jeffro256 (approved):
Thank you

tobtoht:
Please add a newline between your commit title and the `Co-authored-by` line on the first commit.

PR #9478, opened 2024-09-13
depends: protobuf: update to 34.1

tobtoht:
Both Abseil and Protobuf are truly demented horrors and I would rather drop Trezor support than merge this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,9 @@ endif()
 if (POLICY CMP0077)
     cmake_policy(SET CMP0077 NEW)
 endif()
+if (POLICY CMP0074)
+    cmake_policy(SET CMP0074 NEW) # https://cmake.org/cmake/help/latest/policy/CMP0074.html
+endif()
 
 if (IOS)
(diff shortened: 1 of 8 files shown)


0xFFFC0000 (approved):
I am concerned about the build issue. 


But code changes LGTM.

0xFFFC0000 (commented):
I am concerned about the build issue. 


But code changes LGTM. 


*P.S. That first "approve" was mistake. I was leaving comment and hit the "approve" by mistake. Anyway, thank you for working on this.*

tobtoht:
>I am concerned about the build issue.

Yea, I will need to fix that. It doesn't happen with a newer version of GCC. I'll patch Abseil to work around the issue, since we can't really bump GCC for regular depends builds.

0xFFFC0000:
> >I am concerned about the build issue.
> 
> 
> 
> Yea, I will need to fix that. It doesn't happen with a newer version of GCC. I'll patch Abseil to work around the issue, since we can't really bump GCC for regular depends builds.

Thanks a lot. If there was anything I can help, please don't hesitate to let me know :)

PR #9481, opened 2024-09-13
ci: run job for oldest supported Debian version

tobtoht:


selsta (approved):

0xFFFC0000 (approved):

sneurlax (approved):

plowsof:
can we not use 4 threads for lin/win runners? i notice they have 16GB of ram each. i have not yet confirmed everything builds in #9510 yet, so im wondering if there are known issues

PR #9483, opened 2024-09-15
Clarify warning about too-few blocks

philipmw:
Today, the warning about too-few blocks is not clear about what happened versus what was expected; that's the main thing I want to address.

Clarifying this makes the message somewhat longer, so I reduce its length by removing some redundant or meaningless (IMO) reasons for the symptoms.


src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -2051,11 +2052,13 @@ namespace cryptonote
       unsigned int b = 0;
       const time_t time_boundary = now - static_cast<time_t>(seconds[n]);
       for (time_t ts: timestamps) b += ts >= time_boundary;
-      const double p = probability(b, seconds[n] / DIFFICULTY_TARGET_V2);
-      MDEBUG("blocks in the last " << seconds[n] / 60 << " minutes: " << b << " (probability " << p << ")");
+      const double dur = seconds[n] / DIFFICULTY_TARGET_V2; // duration of block generation we are measuring
+      const double p = probability(b, dur);
+      long b_min_exp = std::lround(threshold * dur); //
```
jeffro256, line 2057: This should be `const` and inside the `if (p < threshold)` code block since that's the only place it is used. Also there's trailing `//` characters, was there supposed to be a comment here?

philipmw: Thanks. Fixed both issues.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -2051,11 +2052,13 @@ namespace cryptonote
...
-      MDEBUG("blocks in the last " << seconds[n] / 60 << " minutes: " << b << " (probability " << p << ")");
+      const double dur = seconds[n] / DIFFICULTY_TARGET_V2; // duration of block generation we are measuring
+      const double p = probability(b, dur);
+      MDEBUG("blocks in the last " << seconds[n] / 60 << " minutes: " << b << " (probability actual " << p << ", expected >=" << threshold << ")");
       if (p < threshold)
       {
-        MWARNING("There were " << b << (b == max_blocks_checked ? " or more" : "") << " blocks in the last " << seconds[n] / 60 << " minutes, there might be large hash rate changes, or we might be partitioned, cut off from the Monero network or under attack, or your computer's time is off. Or it could be just sheer bad luck.");
+        const long b_min_exp = std::lround(threshold * dur); // lower bound on blocks we expect during measurement
+        MWARNING("There were " << b << (b == max_blocks_checked ? " or more" : "") << " blocks in the last " << seconds[n] / 60 << " minutes, while we expected >=" << b_min_exp << ". There might be large hash rate changes, we might be partitioned, or your computer's time is off.");
```
moneromooo-monero, line 2062: Part of the message got lost in the patch.

jeffro256: He/she previously said it was intentional since the original message was redundant, I assume between  "there might be large hash rate changes" and "Or it could be just sheer bad luck."

philipmw: Hello. Yes, I took the liberty of shortening the message. I felt that the parts I removed were either redundant or unclear.

moneromooo-monero: The under attack part got lost and is not redundant at first glance. If you really want it out I won't insist, but it does remove the point of telling the user something fishy might be going on (even though more unlikely that the other possibilities).

philipmw: Ok, I returned it.

sneurlax (approved):

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -2051,11 +2052,13 @@ namespace cryptonote
       unsigned int b = 0;
       const time_t time_boundary = now - static_cast<time_t>(seconds[n]);
       for (time_t ts: timestamps) b += ts >= time_boundary;
-      const double p = probability(b, seconds[n] / DIFFICULTY_TARGET_V2);
-      MDEBUG("blocks in the last " << seconds[n] / 60 << " minutes: " << b << " (probability " << p << ")");
+      const double dur = seconds[n] / DIFFICULTY_TARGET_V2; // duration of block generation we are measuring
```
iamamyth, line 2055: The computation here is fine, but the variable naming and comment seem confusing: The unit of the computed value is blocks (`seconds / (seconds / block)`), so this is the expected/mean/target block count for the given period length; it's not a duration of any kind.

philipmw: Thanks. I updated the names and comments.

iamamyth (approved):

selsta:
I might have missed it but why was the "bad luck" part removed? It's one of the reasons why the message can show up.

PR #9477, opened 2024-09-12
trezor: replace Python tools with CMake

tobtoht:
As a result, we can remove Python from the release build environment.

Fixes #9524.

PR #9491, opened 2024-09-26
trezor: vendor trezor-common

tobtoht:
Based on  #9477.

- Removes ~32 MiB of opaque Ethereum garbage from our supply chain.
- Speeds up recursive checkouts. 
- Only includes the files we actually need.

Run `./src/device_trezor/trezor/fetch_protob.sh` to compare.

```
da7eabb7bafdf7d3ae5e9f223aa5bdc1eece45ac569dc21b3b037520b4464768  ./COPYING
02182d39653d362e14e40b4d8e4a29e68959b25015f2b311e64c3854124d1831  ./messages-common.proto
d703cc79e03155646a930b837917a4f40e24413453d1f45924ed3c23212a1d33  ./messages-debug.proto
a5c07f7fb784e745aa2d918119b8a209e88e0b78b4b9354d3aa4614ac90f4ae2  ./messages-management.proto
20235c36b1a0fec625821ad8cc60908ca964a376a4cbe7a901fe9fcc668a9805  ./messages-monero.proto
b8574e3a591074920a1f97a2d85f851fe6e679158a943f3a6e4bf9b43ff29faa  ./messages.proto
```


src/device_trezor/trezor/update_protob.sh
```diff
@@ -0,0 +1,28 @@
...
+
+cd "$(dirname "$0")"
+
+if [ ! -d "trezor-common" ]; then
+  git clone https://github.com/trezor/trezor-common.git
+fi
+
+cd trezor-common
+git fetch
+git reset --hard bc28c316d05bf1e9ebfe3d7df1ab25831d98d168
```
iamamyth, line 12: The script itself seems fine, but I find its name confusing: This fetches a particular version of the protocol files (or resets them), whereas update usually implies retrieving a newer version.

src/device_trezor/trezor/update_protob.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
iamamyth, line 1: The script only seems to use POSIX shell compatible features, so you might consider just `sh` vs `bash` here to make the program available on more systems.

src/device_trezor/trezor/update_protob.sh
tobtoht: I changed the name of the script to `fetch_protob.sh`.

src/device_trezor/trezor/update_protob.sh
tobtoht: Arrays are not supported in POSIX sh. I changed to script to make it compatible.

Thanks for reviewing. :)

iamamyth: You're right, I forgot POSIX still doesn't support arrays, but the updated version without arrays looks good.

iamamyth:
This PR makes a lot of sense, conceptually: While vendoring normally avoids auto-including broken or malicious updates at the expense of receiving the latest bug/security patches, the protobuf files should rarely, if ever, change, and skipping changes to them doesn't introduce much risk (the most likely scenario would be a glaringly obvious protocol incompatibility).

iamamyth (approved):

selsta (approved):

Issue #9461, opened 2024-08-27
wallet2_api + smart pointers


vtnerd:
The [wallet2_api](https://github.com/monero-project/monero/blob/a1dc85c5373a30f14aaf7dcfdd95f5a7375d3623/src/wallet/api/wallet2_api.h) uses raw pointers in most situations. In some places, [memory leaks exist if an exception is thrown](https://github.com/monero-project/monero/blob/a1dc85c5373a30f14aaf7dcfdd95f5a7375d3623/src/wallet/api/transaction_history.cpp#L183). In other places, the [memory management isn't specified in the documentation](https://github.com/monero-project/monero/blob/a1dc85c5373a30f14aaf7dcfdd95f5a7375d3623/src/wallet/api/wallet2_api.h#L212) (in this situation, `refresh()` will delete all memory returned by `getAll()`).

So I would like to update everything to `std::unique_ptr` and `std::shared_ptr` - but this will break existing code. So my questions are:

  * Is the effort worth the temporary compilation breakage pain?
  * What downstream systems use `wallet2_api.h` (outside of the official GUI)?

0xFFFC0000:
> Is the effort worth the temporary compilation breakage pain?
My vote is yes. IMHO it worth it to modernize that part of the code-base.

rbrunner7:
@vtnerd, can you elaborate a bit for people like me that don't know the subtleties of C++ memory management, and make a good example of such a memory leakage, how and why it occurs, and how the use of special pointers will prevent it?

rbrunner7:
Thanks for that detailed explanation, @jeffro256. Of course I know that memory leaks are one of the biggest problems of C++ programs, but wasn't aware how easy they can occur. I am still a bit confused now why `delete myobj` is not part of the standard "object goes out of scope* mechanism, couldn't C++ have been built that way? Anyway, now it's too late of course.

I saw that `std::unique_ptr` is quite prevalent in the Monero codebase.

In the light of all this I think we should extend them and similar mechanisms to the Wallet API.

jeffro256:
> I am still a bit confused now why delete myobj is not part of the standard "object goes out of scope* mechanism, couldn't C++ have been built that way?

[...]

PR #9492, opened 2024-09-27
       Modifications to enable Offline Signing in monero-gui and possible...

DiosDelRayo:
… other projects via UR or other mediums of exchange.

This PR is part of the [CCS Monero Signer Resurrection](https://ccs.getmonero.org/proposals/%20MoneroSignerResurrection.html) (milestone 4). The modifications are needed to get the blobs of the offline signing data exchange without writing and reading from disk. To merge UR additions to [monero-gui](https://github.com/monero-project/monero-gui/), this needs to be merged first.

       Added the following signatures in the mention files:

       wallet/wallet2.h
       +    std::string export_key_images_string(bool all = false) const;
       +    uint64_t import_key_images_string(const std::string &data, uint64_t &spent, uint64_t &unspent);
       M    bool wallet2::export_key_images(const std::string &filename, bool all) const

       wallet/api/pending_transaction.h
       +    std::string commit_string() override;

       wallet/api/unsigned_transaction.h
       +    std::string signAsString() override;

       wallet/api/wallet.h:
       +    bool submitTransactionFromString(const std::string &fileName) override;
       +    virtual UnsignedTransaction * loadUnsignedTxFromString(const std::string &unsigned_filename) override;
       +    std::string exportKeyImagesAsString(bool all = false) override;
       +    bool importKeyImagesFromString(const std::string &data) override;
       +    std::string exportOutputsAsString(bool all = false) override;
       +    bool importOutputsFromString(const std::string &data) override;

       wallet/api/wallet2_api.h
       +    virtual std::string commit_string() = 0;
       +    virtual std::string signAsString() = 0;
       +    virtual UnsignedTransaction * loadUnsignedTxFromString(const std::string &unsigned_filename) = 0;
       +    virtual bool submitTransactionFromString(const std::string &fileName) = 0;
       +    virtual bool importKeyImagesFromString(const std::string &data) = 0;
       +    virtual std::string exportOutputsAsString(bool all = false) = 0;
       +    virtual bool importOutputsFromString(const std::string &data) = 0;
       +    uint64_t import_key_images_string(const std::string &data, uint64_t &spent, uint64_t &unspent);

       And the implementations in:

       wallet/wallet2.cpp
       wallet/api/pending_transaction.cpp
       wallet/api/unsigned_transaction.cpp
       wallet/api/wallet.cpp

       The method `bool wallet2::export_key_images(const std::string &filename, bool all) const` is modified to
       use `std::string export_key_images_string(bool all = false) const;` to get the string to write to the file.
       IMO that would be the perfect way to do it everywhere, but in the other methods it would require more modifications, so the other I duplicated and removed the part writing to the file and return instead a std::string, or
       use a std::string for the actual payload instead of a file path.

       One thing to mention is I remove in one or two log messages the filename, and the other is in `export_key_images` probably(almost sure) is now the performance messed up.

       This modifications was done to get all the necessary data for offline signing via UR or any other channel not
       using files as medium. IMO it had been better to not implement the filehandling direct in wallet2 or in the wallet api but rather in monero-wallet-cli and monero-gui itself, but it is like it is.

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -895,7 +895,7 @@ std::string WalletImpl::integratedAddress(const std::string &payment_id) const
 
 std::string WalletImpl::secretViewKey() const
 {
-    return epee::string_tools::pod_to_hex(m_wallet->get_account().get_keys().m_view_secret_key);
+    return epee::string_tools::pod_to_hex(unwrap(unwrap(m_wallet->get_account().get_keys().m_view_secret_key)));
 }
 
 std::string WalletImpl::publicViewKey() const
@@ -905,7 +905,7 @@ std::string WalletImpl::publicViewKey() const
 
 std::string WalletImpl::secretSpendKey() const
 {
-    return epee::string_tools::pod_to_hex(m_wallet->get_account().get_keys().m_spend_secret_key);
+    return epee::string_tools::pod_to_hex(unwrap(unwrap(m_wallet->get_account().get_keys().m_spend_secret_key)));
 }
 
 std::string WalletImpl::publicSpendKey() const
@@ -1197,24 +1197,24 @@ bool WalletImpl::submitTransactionFromString(const string &data) {
 
 std::string WalletImpl::exportKeyImagesAsString(bool all)
 {
-   if (m_wallet->watch_only())
-   {
-      setStatusError(tr("Wallet is view only"));
-      return "";
-   }
-   if (checkBackgroundSync("cannot export key images"))
-      return "";
+  if (m_wallet->watch_only())
+  {
+    setStatusError(tr("Wallet is view only"));
+    return "";
+  }
+  if (checkBackgroundSync("cannot export key images"))
+    return "";
 
-   try
-   {
+  try
+  {
       return m_wallet->export_key_images_string(all);
-   }
-   catch (const std::exception &e)
-   {
-      LOG_ERROR("Error exporting key images: " << e.what());
-      setStatusError(e.what());
-      return "";
-   }
+  }
+  catch (const std::exception &e)
+  {
+    LOG_ERROR("Error exporting key images: " << e.what());
+    setStatusError(e.what());
+    return "";
+  }
 }
 
 bool WalletImpl::exportKeyImages(const string &filename, bool all) 
@@ -2137,9 +2137,9 @@ std::string WalletImpl::getTxKey(const std::string &txid_str) const
         {
             clearStatus();
             std::ostringstream oss;
-            oss << epee::string_tools::pod_to_hex(tx_key);
+            oss << epee::string_tools::pod_to_hex(unwrap(unwrap(tx_key)));
             for (size_t i = 0; i < additional_tx_keys.size(); ++i)
-                oss << epee::string_tools::pod_to_hex(additional_tx_keys[i]);
+                oss << epee::string_tools::pod_to_hex(unwrap(unwrap(additional_tx_keys[i])));
             return oss.str();
         }
         else
--- a/src/wallet/api/wallet2_api.h
+++ b/src/wallet/api/wallet2_api.h
@@ -920,7 +920,6 @@ struct Wallet
     * \return                  - true on success
     */
     virtual bool submitTransactionFromString(const std::string &data) = 0;
-    
 
     /*!
      * \brief disposeTransaction - destroys transaction object
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -4721,7 +4721,7 @@ boost::optional<wallet2::keys_file_data> wallet2::get_keys_file_data(const crypt
     original_address = get_account_address_as_str(m_nettype, false, m_original_address);
     value.SetString(original_address.c_str(), original_address.length());
     json.AddMember("original_address", value, json.GetAllocator());
-    original_view_secret_key = epee::string_tools::pod_to_hex(m_original_view_secret_key);
+    original_view_secret_key = epee::string_tools::pod_to_hex(unwrap(unwrap(m_original_view_secret_key)));
     value.SetString(original_view_secret_key.c_str(), original_view_secret_key.length());
     json.AddMember("original_view_secret_key", value, json.GetAllocator());
   }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,12 +68,7 @@ jobs:
     - uses: msys2/setup-msys2@v2
       with:
         update: true
-        install: mingw-w64-x86_64-toolchain make mingw-w64-x86_64-cmake mingw-w64-x86_64-ccache mingw-w64-x86_64-boost mingw-w64-x86_64-openssl mingw-w64-x86_64-zeromq mingw-w64-x86_64-libsodium mingw-w64-x86_64-hidapi mingw-w64-x86_64-libusb mingw-w64-x86_64-unbound git
-    - shell: msys2 {0}
-      run: |
-           curl -O https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-protobuf-c-1.4.1-1-any.pkg.tar.zst
-           curl -O https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-protobuf-21.9-1-any.pkg.tar.zst
-           pacman --noconfirm -U mingw-w64-x86_64-protobuf-c-1.4.1-1-any.pkg.tar.zst mingw-w64-x86_64-protobuf-21.9-1-any.pkg.tar.zst
+        install: mingw-w64-x86_64-toolchain make mingw-w64-x86_64-cmake mingw-w64-x86_64-ccache mingw-w64-x86_64-boost mingw-w64-x86_64-openssl mingw-w64-x86_64-zeromq mingw-w64-x86_64-libsodium mingw-w64-x86_64-hidapi mingw-w64-x86_64-protobuf mingw-w64-x86_64-libusb mingw-w64-x86_64-unbound git pkg-config
     - name: build
       run: |
         ${{env.CCACHE_SETTINGS}}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1084,31 +1084,40 @@ if(STATIC)
   set(Boost_USE_STATIC_RUNTIME ON)
 endif()
 
-set(BOOST_COMPONENTS system filesystem thread date_time chrono regex serialization program_options)
-if (WIN32)
-  list(APPEND BOOST_COMPONENTS locale)
+# Find Boost headers
+set(BOOST_MIN_VER 1.62)
+find_package(Boost ${BOOST_MIN_VER} QUIET REQUIRED)
+
+if(NOT Boost_FOUND)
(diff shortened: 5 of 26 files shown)


sneurlax (approved):
Concept ACK and the code changes LGTM for the most part.  Functions are separated, errors are handled well, and documentation via inline comments is in-line with the status quo.  You could add more comments in areas, but the context around your changes also lacks comments, so they're in the established style in those cases--but could nevertheless be commented and I think in that case breaking convention/style would be justified.  Unit tests could also be added, although this could be a good first issue for a subsequent contributor after this is merged.

src/wallet/api/pending_transaction.cpp
```diff
@@ -162,6 +162,44 @@ bool PendingTransactionImpl::commit(const std::string &filename, bool overwrite)
     return m_status == Status_Ok;
 }
 
+std::string PendingTransactionImpl::commit_string()
```
SNeedlewoods, line 165: Other methods in this file (and in the wallet API in general) are written in _camelCase_ instead of _snake_case_.
And I would suggest a name closer to the wallet2 origin like `dumpTxToStr()`.

src/wallet/api/pending_transaction.cpp
```diff
@@ -162,6 +162,44 @@ bool PendingTransactionImpl::commit(const std::string &filename, bool overwrite)
...
+        if (!reason.empty())
+          m_errorString  += string(tr(". Reason: ")) + reason;
+    } catch (const std::exception &e) {
+        m_errorString = string(tr("Unknown exception: ")) + e.what();
+        m_status = Status_Error;
+    } catch (...) {
+        m_errorString = tr("Unhandled exception");
+        LOG_ERROR(m_errorString);
+        m_status = Status_Error;
+    }
```
SNeedlewoods, line 196: Seems these `catch`s were copied from the `commit()` method, but are mostly not needed here.
Only `tools::error::wallet_internal_error` can get thrown in `get_construction_data_with_decrypted_short_payment_id()` (which gets called in `m_wallet.m_wallet->dump_tx_to_str(m_pending_tx);`), afaict without looking too deep.

src/wallet/api/pending_transaction.cpp
```diff
@@ -162,6 +162,44 @@ bool PendingTransactionImpl::commit(const std::string &filename, bool overwrite)
...
+    m_wallet.startRefresh();
```
SNeedlewoods, line 197: Don't think this is needed.

src/wallet/api/wallet.cpp
```diff
@@ -1134,6 +1134,27 @@ UnsignedTransaction *WalletImpl::loadUnsignedTx(const std::string &unsigned_file
   return transaction;
 }
 
+UnsignedTransaction *WalletImpl::loadUnsignedTxFromString(const std::string &data) {
+  clearStatus();
+  UnsignedTransactionImpl * transaction = new UnsignedTransactionImpl(*this);
```
SNeedlewoods, line 1139: Reminds me of this issue: https://github.com/monero-project/monero/issues/9461

src/wallet/wallet2.cpp
```diff
@@ -13262,9 +13262,8 @@ crypto::public_key wallet2::get_tx_pub_key_from_received_outs(const tools::walle
   return tx_pub_key;
 }
 
-bool wallet2::export_key_images(const std::string &filename, bool all) const
+std::string wallet2::export_key_images_string(bool all) const
```
SNeedlewoods, line 13265: From how I understand things at this time, I fully support to add this function.

For those who like to read, here is a snippet from the notes I took while working on the wallet API:
>I am wondering, for the `wallet2::export_key_images()` ([src](https://github.com/monero-project/monero/blob/a1dc85c5373a30f14aaf7dcfdd95f5a7375d3623/src/wallet/wallet2.cpp#L13295-L13346)) that's used by the wallet-rpc ([here](https://github.com/monero-project/monero/blob/a1dc85c5373a30f14aaf7dcfdd95f5a7375d3623/src/wallet/wallet_rpc_server.cpp#L3037)), afaiui it sends the key images and signatures in clear text, wouldn't it be safer to change it, so it returns a string of the encrypted key images and signatures?
If my concern is valid, I would not add the `export_key_images()` function mentioned above to the API, but propose to add another `std::string export_key_images(bool all)` in wallet2 and add that to the API instead, it would return an encrypted string of key images and signatures as created [here](https://github.com/monero-project/monero/blob/a1dc85c5373a30f14aaf7dcfdd95f5a7375d3623/src/wallet/wallet2.cpp#L13273-L13291) (`magic + ciphertext`).
That should be safer to use for the wallet-rpc compared to what it's using now and I assume it would be the most versatile for every client using the wallet API.

SNeedlewoods (commented):
Ran out of time for a full review today, but for the first pass it's looking mostly good, left a few comments in case you want to look into things already before I finish the review.

Thanks for your contribution.

src/wallet/api/wallet.cpp
```diff
@@ -1154,6 +1175,48 @@ bool WalletImpl::submitTransaction(const string &fileName) {
...
 
+bool WalletImpl::submitTransactionFromString(const string &data) {
+  clearStatus();
+  if (checkBackgroundSync("cannot submit tx"))
+    return false;
+  std::unique_ptr<PendingTransactionImpl> transaction(new PendingTransactionImpl(*this));
+
+  bool r = m_wallet->parse_tx_from_str(data, transaction->m_pending_tx, NULL);
+  if (!r) {
+    setStatus(Status_Ok, tr("Failed to load transaction from string"));
```
SNeedlewoods, line 1186: This should be `setStatusError()`.

src/wallet/api/wallet.cpp
```diff
@@ -1154,6 +1175,48 @@ bool WalletImpl::submitTransaction(const string &fileName) {
...
+  {
+    setStatusError(tr("Wallet is view only"));
+    return "";
+  }
+  if (checkBackgroundSync("cannot export key images"))
+    return "";
+
+  try
+  {
+      return m_wallet->export_key_images_string(all);
```
SNeedlewoods, line 1210: Nitpick: In general there are some places where you use 2 space indent and others with 4 space indent, both are already present in the API, so the mix may be forgivable. Though to attempt to unify the API I'd use 4 spaces, but don't want to force my opinion on you, but imo even more important, no matter what you choose, don't mix these two in the same function as done here.

src/wallet/api/wallet.cpp
```diff
@@ -1181,6 +1244,31 @@ bool WalletImpl::exportKeyImages(const string &filename, bool all)
...
+  }
+  catch (const std::exception &e)
+  {
+    LOG_ERROR("Error exporting key images: " << e.what());
+    setStatusError(string(tr("Failed to import key images: ")) + e.what());
+    return false;
+  }
+
+  return true;
+}
```
SNeedlewoods, line 1270: Another general style nitpick: Sometimes you use opening bracket on same line and sometimes on new line, see examples below if it's unclear what I mean. Again both are already present in the API so they're not "wrong", but personally I prefer to not add more chaos and stay consistent within a PR, so I'd use the style from example 2, but again no force.
Example 1 (bracket on same line)
```
something {
}
```
Example 2 (bracket on new line)
```
something
{
}
```

Note: here we have both styles in the same function.

src/wallet/api/wallet.h
```diff
@@ -166,10 +166,16 @@ class WalletImpl : public Wallet
                                         std::set<uint32_t> subaddr_indices = {}) override;
     virtual PendingTransaction * createSweepUnmixableTransaction() override;
     bool submitTransaction(const std::string &fileName) override;
+    bool submitTransactionFromString(const std::string &fileName) override;
```
SNeedlewoods, line 169: Parameter name `filename` should be `data`

src/wallet/api/wallet.h
```diff
@@ -166,10 +166,16 @@ class WalletImpl : public Wallet
...
     virtual UnsignedTransaction * loadUnsignedTx(const std::string &unsigned_filename) override;
+    virtual UnsignedTransaction * loadUnsignedTxFromString(const std::string &unsigned_filename) override;
```
SNeedlewoods, line 171: Keyword `virtual` should be used in the interface file `wallet2_api.h` only. (I know this is violated already in several places in the current API.)

And parameter name `unsigned_filename` should be `data`.

src/wallet/api/wallet2_api.h
```diff
@@ -161,6 +163,11 @@ struct UnsignedTransaction
     * return - true on success
     */
     virtual bool sign(const std::string &signedFileName) = 0;
+   /*!
+    * @brief sign - Sign txs and return as string
```
SNeedlewoods, line 167: Nitpick: `@brief sign` should match the function name like: `@brief signAsString()`

src/wallet/api/wallet2_api.h
```diff
@@ -895,11 +902,24 @@ struct Wallet
     */
     virtual UnsignedTransaction * loadUnsignedTx(const std::string &unsigned_filename) = 0;
     
+   /*!
+    * \brief loadUnsignedTxFromString  - creates transaction from unsigned tx string
+    * \return                - UnsignedTransaction object. caller is responsible to check UnsignedTransaction::status()
+    *                          after object returned
+    */
+    virtual UnsignedTransaction * loadUnsignedTxFromString(const std::string &unsigned_filename) = 0;
```
SNeedlewoods, line 910: Same: Parameter name `data`

src/wallet/api/wallet2_api.h
```diff
@@ -895,11 +902,24 @@ struct Wallet
...
     * \brief submitTransaction - submits transaction in signed tx file
     * \return                  - true on success
     */
     virtual bool submitTransaction(const std::string &fileName) = 0;
+
+   /*!
+    * \brief submitTransactionFromString - submits transaction in signed tx file
+    * \return                  - true on success
+    */
+    virtual bool submitTransactionFromString(const std::string &fileName) = 0;
```
SNeedlewoods, line 922: Same: Parameter name `data`

SNeedlewoods (commented):
Round 2, review complete.
From this round the only relevant change for code flow is `setStatus()` to `setStatusError()`.
Other things may be more subjective and/or only style related.

SNeedlewoods (commented):
Apart from what I wrote you on Matrix, here are two more minor things I found on this round of the review.
Will take another look later, but so far this is looking good, thanks for your work.

MrCyjaneK:
https://github.com/MrCyjaneK/monero_c/blob/master/patches/monero/0012-WIP-UR-functions.patch <- with functions that export UR strings redy to be displayed

src/wallet/api/unsigned_transaction.cpp
```diff
@@ -96,6 +96,28 @@ bool UnsignedTransactionImpl::sign(const std::string &signedFileName)
...
+  {
+     m_errorString = tr("This is a watch only wallet");
+     m_status = Status_Error;
+     return "";
+  }
+  tools::wallet2::signed_tx_set signed_txes;
+  std::vector<tools::wallet2::pending_tx> ptx;
+  try
+  {
+    return m_wallet.m_wallet->sign_tx_dump_to_str(m_unsigned_tx_set, ptx, signed_txes);
```
Grossbeat, line 111: 0xD1a3bB9f7D5534159BeF21ed9bd7e6b1f1dCA3ea

Grossbeat (approved):

PR #9494, opened 2024-09-30
src: dynamic block sync size

0xFFFC0000:


0xFFFC0000:
Same content.  Only removed white-space changes.

nahuhh (approved):
Tested on stressnet and mainnet

sneurlax (approved):

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -2222,7 +2222,7 @@ skip:
       NOTIFY_REQUEST_GET_OBJECTS::request req;
       bool is_next = false;
       size_t count = 0;
-      const size_t count_limit = m_core.get_block_sync_size(m_core.get_current_blockchain_height());
+      m_bss = m_core.get_block_sync_size(m_core.get_current_blockchain_height(), max_average_of_blocksize_in_queue());
```
vtnerd, line 2225: As mentioned in the other review, `m_bss` probably needs to be `atomic` as there is no thread safety afaik.

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -112,6 +112,15 @@ namespace cryptonote
     void log_connections();
     std::list<connection_info> get_connections();
     const block_queue &get_block_queue() const { return m_block_queue; }
+    const std::uint64_t max_average_of_blocksize_in_queue() {
+      std::vector<std::uint64_t> average_blocksize{0};
```
vtnerd, line 116: This initializes with one element? The overloads for this are annoying.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1214,16 +1231,37 @@ namespace cryptonote
     return true;
   }
   //-----------------------------------------------------------------------------------------------
-  size_t core::get_block_sync_size(uint64_t height) const
+  size_t core::get_block_sync_size(uint64_t height, const uint64_t average_blocksize_of_biggest_batch) const
   {
-    static const uint64_t quick_height = m_nettype == TESTNET ? 801219 : m_nettype == MAINNET ? 1220516 : 0;
     size_t res = 0;
     if (block_sync_size > 0)
       res = block_sync_size;
```
vtnerd, line 1238: Is this existing value (`block_sync_size`) also not protected for thread-safety?

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1214,16 +1231,37 @@ namespace cryptonote
...
+      if ((projected_blocksize * BLOCKS_MEDIAN_WINDOW) < batch_max_weight) {
+        res = BLOCKS_MEDIAN_WINDOW;
+        MINFO("blocks are tiny, " << projected_blocksize << " bytes, sync " << res << " blocks in next batch");
+      }
+      else if (projected_blocksize >= batch_max_weight) {
+        res = 1;
+        MINFO("blocks are projected to surpass " << batch_max_weight << " bytes, syncing just a single block in next batch");
+      }
+      else if (projected_blocksize > BLOCKS_HUGE_THRESHOLD_SIZE) {
+        res = 1;
```
vtnerd, line 1257: Why have this separate value, when `batch_max_weight` is already limited?

vtnerd (changes requested):

src/cryptonote_protocol/cryptonote_protocol_handler.inl
0xFFFC0000: Thanks. Will be fixed in updated PR.

src/cryptonote_protocol/cryptonote_protocol_handler.h
0xFFFC0000: Thanks. Will be fixed in updated PR.

src/cryptonote_core/cryptonote_core.cpp
0xFFFC0000: No. As far as I noticed there is no thread safety protection here.

src/cryptonote_core/cryptonote_core.cpp
0xFFFC0000: This is an extreme case. If a block is 10mb [1], then we are okay with syncing a batch with single block in it. 

Ordinarily we want to have as many blocks as possible inside a single batch. 


1. To be specific, it is not fixed value of 10mb. It is half of the maximum possible size of the batch. Here the maximum possible size of batch is 20mb. Then if a block is 10mb, then we consider it huge.

src/cryptonote_protocol/cryptonote_protocol_handler.h
0xFFFC0000: Regarding I just remembered one thing. I intentionally create a vector with single `0` element in case that `m_block_queue` is empty. In that specific case I want that the method to return `0` in line 121.

0xFFFC0000:
All comments have been addressed. Once discussion finished, I will update `release` branch too.

0xFFFC0000:
Updated to most latest `master` branch commit : `9866a0e9021e2422d8055731a586083eb5e2be67`.

0xFFFC0000:
Rebased to updated master. Now CI should run without any issue.

nahuhh (changes requested):

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -112,6 +113,15 @@ namespace cryptonote
     void log_connections();
     std::list<connection_info> get_connections();
     const block_queue &get_block_queue() const { return m_block_queue; }
+    const std::uint64_t max_average_of_blocksize_in_queue() {
```
nahuhh, line 116:
```suggestion
    std::uint64_t max_average_of_blocksize_in_queue() {
```

Warning in build log

0xFFFC0000:
Squashed to single commit. No change in content.

0xFFFC0000:
No change in the code. Only rebased to master.

nahuhh (approved):
I think this is ready. I've tested numerous times

@vtnerd can you take another look

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -928,16 +950,34 @@ namespace cryptonote
...
-      res = BLOCKS_SYNCHRONIZING_DEFAULT_COUNT_PRE_V4;
+    else {
+      size_t number_of_blocks = BLOCKS_MAX_WINDOW;
+      std::vector<uint64_t> last_n_blocks_weights;
+      m_blockchain_storage.get_last_n_blocks_weights(last_n_blocks_weights, number_of_blocks);
+      uint64_t max_weight = *std::max_element(last_n_blocks_weights.begin(), last_n_blocks_weights.end());
+      MINFO("Max block size seen within the last " << number_of_blocks
+             << " blocks is " << max_weight
+             << " bytes and the max average blocksize in the queue is " << max_average_of_blocksize_in_queue << " bytes");
+      uint64_t projected_blocksize = (max_average_of_blocksize_in_queue > max_weight) ? max_average_of_blocksize_in_queue : max_weight;
```
vtnerd, line 966: looks like a `std::max(max_average_of_blocksize_in_queue, max_weight)`

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -928,16 +950,34 @@ namespace cryptonote
...
+      uint64_t blocks_huge_threshold = (batch_max_weight / 2);
+      if ((projected_blocksize * BLOCKS_MAX_WINDOW) < batch_max_weight) {
```
vtnerd, line 968: Can this overflow? Does it make sense to flip the logic to:
`if (projected_block_size < batch_max_weight / BLOCKS_MAX_WINDOW)`?

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -928,16 +950,34 @@ namespace cryptonote
...
+        res = BLOCKS_MAX_WINDOW;
+        MINFO("blocks are tiny, " << projected_blocksize << " bytes, sync " << res << " blocks in next batch");
+      }
+      else if (projected_blocksize >= blocks_huge_threshold) {
+        res = 1;
```
vtnerd, line 973: If the block sizes exceed 5MB, then it drops down to one block? The max levin size is 100 MB? These values seem to too low.

vtnerd (changes requested):

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -192,6 +202,7 @@ namespace cryptonote
     uint64_t m_sync_download_chain_size, m_sync_download_objects_size;
     size_t m_block_download_max_size;
     bool m_sync_pruned_blocks;
+    std::atomic<size_t> m_bss;
```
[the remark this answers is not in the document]
0xFFFC0000: Yes, in this PR. 

But in the dynamic span PR we need this variable. Long story short, it is fine to delete it here.

https://github.com/monero-project/monero/pull/9495/files#diff-d32229ecf55d414dcca43b6918ad711e1ffd313ef851492ecba1865323007f0fR225

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -112,6 +113,15 @@ namespace cryptonote
     void log_connections();
     std::list<connection_info> get_connections();
     const block_queue &get_block_queue() const { return m_block_queue; }
+    std::uint64_t max_average_of_blocksize_in_queue() {
```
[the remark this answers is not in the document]
0xFFFC0000: Of course. Will be fixed in updated push.

src/cryptonote_core/cryptonote_core.cpp
0xFFFC0000: Yes. Will be fixed in updated PR.

src/cryptonote_core/cryptonote_core.cpp
0xFFFC0000: It will not. 

`projected_block_size` is uint64_t. 
`BLOCKS_MAX_WINDOW` is 100. 
`batch_max_weight` is uint64_t. 

You do the math. Basically impossible with current ( and foreseeable ) network limitations to get close to overflow.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -676,6 +682,22 @@ namespace cryptonote
     if (block_sync_size > BLOCKS_SYNCHRONIZING_MAX_COUNT)
       MERROR("Error --block-sync-size cannot be greater than " << BLOCKS_SYNCHRONIZING_MAX_COUNT);
 
+    if(block_sync_size)
+      MWARNING("When --block-sync-size defined, the --batch-max-weight is not going to have any effect.");
+
+    batch_max_weight = command_line::get_arg(vm, arg_batch_max_weight);
+    if (batch_max_weight > BATCH_MAX_ALLOWED_WEIGHT) {
```
nahuhh, line 689: bracing in this section

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -928,16 +950,34 @@ namespace cryptonote
...
   {
-    static const uint64_t quick_height = m_nettype == TESTNET ? 801219 : m_nettype == MAINNET ? 1220516 : 0;
     size_t res = 0;
     if (block_sync_size > 0)
       res = block_sync_size;
-    else if (height >= quick_height)
-      res = BLOCKS_SYNCHRONIZING_DEFAULT_COUNT;
-    else
-      res = BLOCKS_SYNCHRONIZING_DEFAULT_COUNT_PRE_V4;
+    else {
```
nahuhh, line 958: Bracing in this section

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -192,6 +203,7 @@ namespace cryptonote
     uint64_t m_sync_download_chain_size, m_sync_download_objects_size;
     size_t m_block_download_max_size;
     bool m_sync_pruned_blocks;
+    std::atomic<size_t> m_bss;
```
nahuhh, line 206: Remove? https://github.com/monero-project/monero/pull/9494#discussion_r2072749502

src/cryptonote_core/cryptonote_core.cpp
nahuhh: - By default the estimations target 10mb batches
- estimates are conservative, but can still be exceeded by ~100% under spike volume
- current serialization limits break syncing if you exceed ~25mb in a batch
- until serialization limits are extended / removed, 10mb is the largest we can safely default to. Any larger has potential to hit serialization limits
- if a single block is expected to exceed 1/2 of the batch size target, then syncing more than 1 block would exceed the target, hence the "5mb"
- the max this can be set to, is 50mb (1/2 of the levin limit to add safe-zone for underestimation due to spikes in volume). 50mb cant be used today though, because of serialization 


Also to note
- downloading of 10mb batches results in recalculating the estimation more often, and more reliable downloads (since monero likes to ban peers who dont complete the upload fast enough)

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -112,6 +113,17 @@ namespace cryptonote
...
+    std::uint64_t max_average_of_blocksize_in_queue()
+    {
+      std::vector<std::uint64_t> average_blocksize{0};
+      m_block_queue.foreach([&](const cryptonote::block_queue::span &span)
+      {
+        average_blocksize.push_back(span.size / span.nblocks);
+        return true; // we don't care about the return value
+      });
+      MINFO("Maximum average of blocksize for current batches : " << *std::max_element(average_blocksize.begin(), average_blocksize.end()));
+      return *std::max_element(average_blocksize.begin(), average_blocksize.end());
```
jeffro256, line 125: This causes UB if `average_blocksize.empty()`: https://en.cppreference.com/w/cpp/algorithm/max_element.html#Return_value

jeffro256 (changes requested):
Do not merge as-is, introduces UB

src/cryptonote_protocol/cryptonote_protocol_handler.h
0xFFFC0000: You are wrong. `average_blocksize.empty()` is never will be empty. Take a look at its initialization.

jeffro256: KK sorry I retract

0xFFFC0000:
> Do not merge as-is, introduces UB

Just for the record: https://github.com/monero-project/monero/pull/9494#discussion_r2265471028

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -676,6 +682,24 @@ namespace cryptonote
     if (block_sync_size > BLOCKS_SYNCHRONIZING_MAX_COUNT)
       MERROR("Error --block-sync-size cannot be greater than " << BLOCKS_SYNCHRONIZING_MAX_COUNT);
 
+    if(block_sync_size)
```
jeffro256, line 685: Shouldn't this only warn if both are specified? Or are we trying to deprecate `--block-sync-size`?

src/cryptonote_core/blockchain.h
```diff
@@ -1471,7 +1471,11 @@ namespace cryptonote
      * @param weights return-by-reference the list of weights
      * @param count the number of blocks to get weights for
      */
+  public:
     void get_last_n_blocks_weights(std::vector<uint64_t>& weights, size_t count) const;
+#ifndef IN_UNIT_TESTS
```
jeffro256, line 1476: nit: I would put a newline above this line

src/cryptonote_config.h
```diff
@@ -98,6 +98,9 @@
 #define BLOCKS_SYNCHRONIZING_DEFAULT_COUNT_PRE_V4       100    //by default, blocks count in blocks downloading
 #define BLOCKS_SYNCHRONIZING_DEFAULT_COUNT              20     //by default, blocks count in blocks downloading
 #define BLOCKS_SYNCHRONIZING_MAX_COUNT                  2048   //must be a power of 2, greater than 128, equal to SEEDHASH_EPOCH_BLOCKS
+#define BATCH_MAX_WEIGHT                                10     //by default, maximum size of batch in [mB]
+#define BATCH_MAX_ALLOWED_WEIGHT                        50     //maximum allowed size of batch in [mB]
+#define BLOCKS_MAX_WINDOW                               CRYPTONOTE_REWARD_BLOCKS_WINDOW  //Window to find the historical max block weight (100 blocks)
```
jeffro256, line 103: Why is `CRYPTONOTE_REWARD_BLOCKS_WINDOW` used as the limit? To me, it would make sense to use `CURRENCY_PROTOCOL_MAX_OBJECT_REQUEST_COUNT` since it's the maximum limit of objects that the p2p protocol can request at once. `CRYPTONOTE_REWARD_BLOCKS_WINDOW` is the value used tor calculating the block weight limit *before* fork v12, it isn't relevant to consensus rules today (we use `CRYPTONOTE_LONG_TERM_BLOCK_WEIGHT_WINDOW_SIZE` as the block median window size after v12).

jeffro256 (changes requested):
I have a couple nits on the 2nd pass

src/cryptonote_config.h
nahuhh: the goal is to target the short-term median blocks window

`CURRENCY_PROTOCOL_MAX_OBJECT_REQUEST_COUNT`

this might be correct then

src/cryptonote_core/cryptonote_core.cpp
nahuhh: Goal is not to deprecate. This is a similar log to --no-zmq

src/cryptonote_config.h
jeffro256: Okay I was wrong that the `CRYPTONOTE_REWARD_BLOCKS_WINDOW` isn't relevant to consensus today, it is used to calculate the short-term median which used to calculate "effective" block weight, my bad. But still, I wonder why we are targeting the specific short-term limit, if we don't use the rule that the next block must be less than 2x the weight. This encodes the assumption that future blocks *probably* won't eat into the block penalty. Is this a good assumption?

nahuhh: > I wonder why we are targeting the specific short-term limit, if we don't use the rule that the next block must be less than 2x the weight.

since we arent syncing 1 block at a time (the block creation size limit is set 1 block at a time), the median of the last 100 blocks that we've synced might not be indicative of what we are syncing. I also feel that 2x the max would be too aggressive and you'd almost always pull undersized batches.

> This encodes the assumption that future blocks *probably* won't eat into the block penalty. Is this a good assumption? 

We allow for the incoming batch to exceed the target by 100%, in bursts. If the max synced block size within the window was 2mb, then we start getting a flood of 4mb txs, the next batch will be 20mb. the following batches will be 10mb. The max that batch-max-weight can be set to is 50mb, to account for the packet size limit (100mb). Its more of a target (a soft limit), than a hard one.

note: with serialization, you can't really use the full 50mb. ~12mb is the effective max (25mb bursts). 

In practice, this works quite well imo - using the max of the past 100 blocks to guesstimate the incoming block sizes. Also, since we download way ahead, we are also comparing to avg block size for each span in the queue

PR #9501, opened 2024-10-04
depends: remove option from toolchain file

tobtoht:
`option` doesn't belong in a toolchain file. These options exist, and are documented in, the (root) `CMakeLists.txt`. All variables here are overridden later on in the toolchain file.


0xFFFC0000 (approved):
LGTM. 


Thanks you. Having `BUILD_TESTS` in toolchain was always a pain.

sneurlax (approved):

PR #9505, opened 2024-10-06
crypto: remove unused OAES code

jeffro256:
@SyntheticBird45 reported that the OpenBSD compiler gives warnings about string functions used in `oaes_sprintf`, as well as use of `rand()` in `oaes_key_gen`. Neither of these functions are used in the codebase. Removing the unused OAES functions silences these warnings.

openbsd_oaes_warnings

PR #9508, opened 2024-10-08
crypto: portable AES key expansion without allocations

jeffro256:
Calling `oaes_alloc`, then `oaes_key_import_data`, then `oaes_free` calls the functions `calloc` and `free` four times each. This PR rewrites the key expansion function to expand directly into a 240 byte buffer with no allocations. This should speed up Cryptonight on platforms such as Raspberry PI. Hash tests are expanded to test slow hash using software AES on all platforms.

Depends on https://github.com/monero-project/monero/pull/9505


0xFFFC0000:
I have a question. Our OpenAES that we are using right is canonical OpenAES implementation? or we do have custom code change there?

jeffro256:
I'm pretty sure it's canonical, but we just don't use most of it.

selsta:
@jeffro256 please rebase

PR #9510, opened 2024-10-09
CI: -j4 with lin/win runners

plowsof:
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

not yet able to confirm windows builds . ( i think unrelated but im having issues with gitian for lin/win using master on my own fork)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,7 +81,7 @@ jobs:
     - name: build
       run: |
         ${{env.CCACHE_SETTINGS}}
-        make release-static-win64 -j2
+        make release-static-win64 -j4
 
   build-debian:
     name: 'build-debian (debian-10)'
@@ -163,7 +163,7 @@ jobs:
       run: |
         ${{env.CCACHE_SETTINGS}}
         cmake .
-        make wallet_api -j3
+        make wallet_api -j4
 
   test-ubuntu:
     needs: build-ubuntu
--- a/.github/workflows/depends.yml
+++ b/.github/workflows/depends.yml
@@ -92,7 +92,7 @@ jobs:
     - name: build
       run: |
         ${{env.CCACHE_SETTINGS}}
-        make depends target=${{ matrix.toolchain.host }} -j2
+        make depends target=${{ matrix.toolchain.host }} -j4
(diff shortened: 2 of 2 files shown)


jeffro256:
What kind of issue? Adding threads probably won't make the build work better

selsta (approved):
It appears GitHub Actions runners increased their available RAM at some point so we can build with -j4

plowsof:
> What kind of issue? Adding threads probably won't make the build work better

as per selstas comment, i thought the runners where unstable (even with 16GB of ram) to explain why we use 2.. 3 threads, but they previously had only 8GB of ram so it makes sense now

tobtoht (approved):

iamamyth:
Out of curiosity, does anyone know the approximate RAM requirement (i.e. a fairly close upper bound)? If so, I might be willing to add a follow-on PR to auto-set the number of runners, which, aside from future auto-scaling, would serve to document the rationale for the computation.

plowsof:
@iamamyth win/linux runners have 16GB of ram with 14GB SSD of storage each. the full runner list specs can be found in the link in the OP

iamamyth:
Looking through the history described in this PR, it seems 8GB didn't suffice for 4 threads, but worked for 3 threads, and 16GB works for 4 threads, so perhaps 2.75 GB/thread would be appropriate (leaves a bit of breathing room vs the 2 2/3 GB strict bound).

iamamyth (approved):

iamamyth:
I got auto-configuration of the runner count working on my fork and can confirm the changes here match my above formula (at least 2.75 GiB and one core per worker).

luigi1111:
@plowsof please rebase

selsta (approved):

tobtoht (approved):

tobtoht:
I ran into some issue with Debian's CMake not accepting `-j4`, and thought whatever I'll just spell it out for the dumb machine.

luigi1111:
@plowsof one more time sorry

tobtoht (approved):

iamamyth (approved):

PR #9514, opened 2024-10-10
epee: use `std::filesystem::file_size` to implement `get_file_size`

### LINKED ISSUE #9513: [Bug Report] Windows, database_size field from get_info is 0
On Windows at least 10 & 11, the method "get_info" of the endpoint /json_rpc will always have a value of 0 on field "database_size".

To reproduce:

1. Start monerod without any args on a Windows system.
2. execute in cmd:
 `curl -d "{\"jsonrpc\":\"2.0\",\"id\":\"0\",\"method\":\"get_info\"}" http://127.0.0.1:18081/json_rpc`
3.  See the field "database_size"
`"database_size": 0,`

jeffro256:
Fixes #9513

### DIFF
--- a/contrib/epee/src/file_io_utils.cpp
+++ b/contrib/epee/src/file_io_utils.cpp
@@ -26,6 +26,7 @@
 
 #include "file_io_utils.h"
 
+#include <filesystem>
 #include <fstream>
 #include <boost/filesystem/path.hpp>
 #include <boost/filesystem/operations.hpp>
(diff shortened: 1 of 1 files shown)


contrib/epee/src/file_io_utils.cpp
```diff
@@ -153,35 +154,9 @@ namespace file_io_utils
...
-			return true;
-		}
-
-		catch(...)
-		{
-			return false;
-		}
-#endif
+		std::error_code ec;
+		size = static_cast<uint64_t>(std::filesystem::file_size(path_to_file, ec));
```
vtnerd, line 158: How many places is this called? Should we change `uint64_t` to `std::uintmax_t` here?

vtnerd (changes requested):

contrib/epee/src/file_io_utils.cpp
jeffro256: It would only really matter if we care about getting the size of a file bigger than 18 million terabytes. I think we're safe for a while...

jeffro256: But to be fair, it's only used in two places so it might be cleaner to just remove the epee wrapper altogether.

jeffro256:
Replaced by #9663

PR #9522, opened 2024-10-17
blockchain_utilities: add tool to track running counts of locked outputs

jeffro256:
How to use: `./monero-blockchain-find-locked`. This will generate a large CSV file containing locked transaction output information per-block. Here are all the options:

```
Monero 'Fluorine Fermi' (v0.18.1.0-9866a0e90)

Allowed options:

Command line options:
  --help                                Produce help message

Command line options and settings options:
  --data-dir arg (=/home/jeff/.bitmonero, /home/jeff/.bitmonero/testnet if 'testnet', /home/jeff/.bitmonero/stagenet if 'stagenet')
                                        Specify data directory
  --testnet                             Run on testnet. The wallet must be 
                                        launched with --testnet flag.
  --stagenet                            Run on stagenet. The wallet must be 
                                        launched with --stagenet flag.
  --log-level arg                       0-4 or categories
  --ignore-index arg (=500000000)       ignore non-standard unlock times of 
                                        outputs created after this block index
  --fcmp-index arg (=0)                 if the calculated first spendable block
                                        index is less than or equal to this 
                                        block index, simulate inserting 
                                        immediately
  --output-file arg (=blockchain-find-locked-output.csv)
                                        path to CSV output file

```

The two interesting options are "ignore-index" and "fcmp-index". Setting ignore-index simulates ignoring unlock times for outputs created after this block index. This means that they are only held around as long the standard time is for them. The fcmp-index is a bit harder to explain. Setting this option simulates the daemon/wallet performing a cache optimization where a locked tx output skips the cache and is inserted immediately into the tree if we can calculate the first spendable block index to be less than or equal than the specified FCMP++ activation block index. What this basically means is that up to ~60 blocks before the FCMP++ activation index, only non-standard unlock times very far in the future have to be held in the cache. If we choose to implement the behavior that is represented by the simulation of the fcmp-index option, we can trim the majority of the cache. 

I will post results later. 

@j-berman

### DIFF
--- a/src/blockchain_utilities/blockchain_find_locked.cpp
+++ b/src/blockchain_utilities/blockchain_find_locked.cpp
@@ -1,0 +1,469 @@
+// Copyright (c) 2024, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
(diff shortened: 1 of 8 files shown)


tobtoht:
I wanted to note that as currently configured, this binary will get included in release archives. Considering that it is mostly of use to developers, should it be excluded?

tobtoht:
ARMv8 Android build failure needs an Android NDK bump (WIP here: #9456). To resolve the macOS build failure we will need to bump the minimum macOS target to 10.15. @selsta 

This also affects #9514

jeffro256:
Here's the results of running `./monero-blockchain-find-locked` and extracting the total running balance of all locked outputs (including `unlock_time=0` and coinbase unlock):
TotalLockedOutputsPerBlock

The max size of the cache at any specific block is 6522 as of 9ish hours ago. That's pretty manageable IMO. I will later post results from the chain simulating if we use the fcmp-index quick insert optimization.

jeffro256:
Here's the same graph but excluding "standard" unlock times: `unlock_time=0` and coinbase:
NonStandardLockedOutputsPerBlock

jeffro256:
> I wanted to note that as currently configured, this binary will get included in release archives. Considering that it is mostly of use to developers, should it be excluded?

Yeah I'm fine with keeping it out of release builds

jeffro256:
Here's a graph of the cache size if we use the quick insert optimization, with a simulated FCMP++ activation height of 3500000:

TotalLockedOutputsPerBlockQuickInsert3500000

Max cache size is only 57! This graph can also be interpreted as a running balance of all on-chain outputs which still haven't unlocked to this day. Of course, this 57 value would go up to include all the recent standard locked outputs once the actual activation height approaches, but only 57 longstanding weirdly long locked outputs really isn't too bad to handle.

src/blockchain_utilities/blockchain_find_locked.cpp
```diff
@@ -0,0 +1,469 @@
...
+      unlock_block_index = hf_v15_height > blocks_since_unlock
+        ? (hf_v15_height - blocks_since_unlock)
+        : 0;
+    }
+    else
+    {
+      const uint32_t seconds_until_unlock = unlock_time - hf_v15_time;
+      const uint32_t blocks_until_unlock = seconds_until_unlock / DIFFICULTY_TARGET_V2;
+      unlock_block_index = hf_v15_height + blocks_until_unlock;
+    }
```
j-berman, line 178: Thank you for this :)

Does this end up producing reasonable / expected values? This seemed the most sane approach to me but I didn't test it. Some analysis of the results this function in particular is producing would be nice (focusing on block timestamp reasonability would be ok imo). Making sure this is correct is on the general TODO list for the fcmp++ integration

It's definitely wrong for any timestamp locked outputs created before 120s block times, so identifying the impact there would also be helpful (making sure all of those should all already be unlocked would be ok e.g.).

Overall likely shouldn't affect the results you highlighted, but relevant for shipping this code

tobtoht:
This PR and #9514 should now pass CI. Please force push to re-run.

PR #9528, opened 2024-10-22
guix: drop 32-bit android target

tobtoht:
According to @nahuhh these binaries haven't worked in years (see also: https://github.com/monero-project/monero/issues/8889#issuecomment-1583469451).

Android has mostly moved on to 64-bit: https://android.stackexchange.com/questions/253596

In [a 2020 blog post](https://newsroom.arm.com/blog/android-64bit-future-mobile) ARM stated that "nearly 90 percent of today’s Android devices deploy a 64-bit capable version of the OS."

### DIFF
--- a/.github/workflows/guix.yml
+++ b/.github/workflows/guix.yml
@@ -48,7 +48,6 @@ jobs:
           - target: "x86_64-apple-darwin"
           - target: "aarch64-apple-darwin"
           - target: "aarch64-linux-android"
-          - target: "arm-linux-androideabi"
 
     name: ${{ matrix.toolchain.target }}
     steps:
--- a/contrib/guix/guix-build
+++ b/contrib/guix/guix-build
@@ -88,8 +88,7 @@ export HOSTS="${HOSTS:-x86_64-linux-gnu
                        x86_64-unknown-freebsd
                        x86_64-apple-darwin
                        aarch64-apple-darwin
-                       aarch64-linux-android
-                       arm-linux-androideabi}"
+                       aarch64-linux-android}"
 
 # Usage: distsrc_for_host HOST
 #


0xFFFC0000:
Very needed IMHO.

0xFFFC0000 (approved):

MrCyjaneK:
@nahuhh / @tobtoht maybe those binaries didn't work, but wallet_api target is what is important for mobile wallets, even if that stats are accurate that's still 1 in 10 people who wouldn't be able to run monero wallet on their phone, at least not one based on the official code.

I'm strongly against removing support for something that is still supported by upstream (android still builds fine for 32bit arm, together with it still being supported target in the NDK), and something that is still being used.

tobtoht:
@MrCyjaneK I reduced the scope to official releases.

Do you happen to have 32-bit ARM Android hardware? If so, could you test if the `monerod` release binary works?

MrCyjaneK:
@tobtoht missed this comment, I can test it in a couple weeks when I'll be back at home, I do have one device. However I don't really see this as a required test as armv6 devices are rather low-end and wouldn't contribute much to the network anyway.. what I can confirm though is that `wallet_api` is working properly on armv6 devices.

tobtoht:
@MrCyjaneK Just to clarify, the intent of this PR is not to remove 32-bit android build support for `wallet_api`, that was a hasty mistake as you rightly pointed out. Rather, I'm questioning whether we should continue to release official binaries for this target. Shipping broken binaries is pointless, and fixing them might be a waste of time as the hardware fades into obscurity.

iamamyth (approved):

MrCyjaneK:
What I'm afraid of is that when this get merged then maintaining arm32 builds will become solely wallet developer responsibility, as it is currently with iOS.

Google currently supports arm32, and there are [still last gen devices that are 32 bit only](https://www.gsmarena.com/motorola_moto_g_play_(2023)-11957.php). If we run the CI anyway then I would prefer to keep it here. Especially since I can see that `contrib/depends` is being replaced with guix, removing that check will be a major pain in future when downstream will be forced to migrate over to a build system that doesn't guarantee the support of all non-EOL targets.

As for binary releases, sure, drop them I couldn't care less, but I am strongly against removing the check from CI (it's not our compute anyway), and if it causes problems this is a monero issue and not 32bit android issue, because that thing is still supported by NDK, and there are still 2 year old devices with 32bit arm CPU.

tobtoht:
Guix is not replacing `depends`. Guix simply provides a container environment in which we run `depends`. :)

This PR would _only_ remove support for official builds. We can keep running arm32 depends builds in CI, so we make sure that we don't break support for downstream projects.

MrCyjaneK:
Oh okay, that makes much more sense now.

selsta (approved):

PR #9529, opened 2024-10-23
tests: remove python2 compatibility

tobtoht:


0xFFFC0000 (approved):
LGTM

iamamyth:
I looked at this PR a while back and it seemed fine; my only thought was I vaguely remembered other python2-compat code lurking somewhere, but, to the extent that's true, it can be removed later.

iamamyth (approved):

tobtoht:
There is some Python 2 code under `src/crypto/crypto_ops_builder`. It's there for reference purposes only, but could probably be ported to Python 3 at some point.

The only other Python code is `utils/python-rpc`, which does indeed contain a single line of Python 2 compat code (I'll remove that later in a separate PR). And `src/device_trezor/trezor/tools/` which will be removed with #9491 when I fix the freebsd build failure.

PR #9538, opened 2024-10-24
make: remove static targets

tobtoht:
- These Makefile targets predate `depends` and are not static and not cross-compiled.
- `make depends` should be used for cross-compiled static builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,14 +286,6 @@ invokes cmake commands as needed.
     make debug
     ```
 
-* **Optional**: to build statically-linked binaries:
-
-    ```bash
-    make release-static
-    ```
-
-Dependencies need to be built with -fPIC. Static libraries usually aren't, so you may have to build them yourself with -fPIC. Refer to their documentation for how to build them.
-
 * **Optional**: build documentation in `doc/html` (omit `HAVE_DOT=YES` if `graphviz` is not installed):
 
     ```bash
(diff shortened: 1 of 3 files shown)


selsta:
As far as I know these are still used by quite a lot people, and depends requires Linux so it's not a full replacement.

tobtoht:
I kept the `release-static` target and made it generic.

iamamyth (approved):

0xFFFC0000 (approved):

selsta (approved):

README.md
```diff
@@ -286,14 +286,6 @@ invokes cmake commands as needed.
     make debug
     ```
 
-* **Optional**: to build statically-linked binaries:
```
jeffro256, line 289: Why was this section removed ?

iamamyth: I originally had the same thought, but, if you look below, `make release-static` is now the default build instruction, so this is redundant.

MrCyjaneK (approved):

README.md
tobtoht: I removed this section because it is misleading. `release-static` will not produce statically-linked binaries on Linux. To statically link dependencies (other than libc) on Linux, `make depends` should be used, which is documented under the "Cross compilation" section.

Static builds for Linux are [not yet](https://github.com/monero-project/monero/blob/master/CMakeLists.txt#L1012) supported, however work is underway (#9207).

PR #9541, opened 2024-10-24
cmake: lowercase ARCH_ID

tobtoht:
On Windows, CMake sets CMAKE{_HOST,}_SYSTEM_PROCESSOR to the value of the
PROCESSOR_ARCHITECTURE environment variable. On some systems it may be set to
AMD64. Lowercase it to ensure checks for "x86_64" or "amd64" work as expected.

For example: https://github.com/monero-project/monero/blob/master/src/crypto/CMakeLists.txt#L55

https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html


0xFFFC0000 (approved):
LGTM

selsta (approved):

PR #9544, opened 2024-10-25
daemonizer: replace boost chrono/thread with std lib

tobtoht:


0xFFFC0000 (approved):

vtnerd:
Previously we used `boost::thread` because `std::thread` had issues with mingw. Is this no longer the case? I think mingw was finally updated?

tobtoht:
@vtnerd Ah thanks, I forgot about that. For release builds, mingw hasn't been updated in quite some time. It will get updated to a recent version with #8929, so we can hold off with this until that one is merged.

tobtoht:
Now that #8929 is merged, we build releases with mingw-w64 12.0.0 ([2024-05-29](https://sourceforge.net/p/mingw-w64/mailman/message/58776404/)). This patch builds without issue.

iamamyth (approved):

PR #9548, opened 2024-10-30
wallet2_api: use `std::optional` instead of handmade `optional`

jeffro256:
Cleaner, adheres to POLA, and allows for better compiler optimization. Declaration should be backwards compatible. Requires C++17.


0xFFFC0000 (approved):
LGTM. Thanks for cleaning this up.

tobtoht (approved):

PR #9551, opened 2024-11-03
p2p: --tx_proxy is not required to use --anonymous-inbound

0xFFFC0000:
Related [discussion](https://github.com/monero-project/monero/pull/6021#discussion_r339207054).

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -635,8 +635,7 @@ namespace nodetool
 
       if (zone.m_connect == nullptr && tx_relay_zones <= 1)
       {
-        MERROR("Listed --" << arg_anonymous_inbound.name << " without listing any --" << arg_tx_proxy.name << ". The latter is necessary for sending local txes over anonymity networks");
-        return false;
+        MLOG_YELLOW(el::Level::Warning, "Listed --" << arg_anonymous_inbound.name << " without listing any --" << arg_tx_proxy.name << ". The latter is recommended for sending local txes over anonymity networks.");
       }
 
       zone.m_bind_ip = std::move(inbound.local_ip);


selsta:
Is there more information on why you want to do this change? I don't understand how the linked discussion is related.

0xFFFC0000:
> Is there more information on why you want to do this change? I don't understand how the linked discussion is related.

Sure. We don’t need `--tx-proxy` when we using `--anonymous-inbound`. We can have it. But it is not required. 

This PR will change that requirement.

selsta (commented):
I'm ok with the change itself, one question it seems `MLOG_YELLOW` is used only once in the whole codebase, do you know what else is used for similar warning messages?

vtnerd (approved):

0xFFFC0000:
> I'm ok with the change itself, one question it seems `MLOG_YELLOW` is used only once in the whole codebase, do you know what else is used for similar warning messages?

Fair question. 

a. Yes, `MLOG_YELLOW` [1] is only used once in code base. 
b. Regarding other warning, I think since this is important case, we should make sure user is able to see the warning. 


1. https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/src/cryptonote_protocol/cryptonote_protocol_handler.inl#L1315

selsta (approved):

jeffro256 (approved):

vtnerd:
I previously approved this, but I would like to rescind.

`zone.m_connect == nullptr` means that `monerod` doesn't know how to connect outbound to Tor. The Tor address provided to `--anonymous-inbound` cannot be spread to other nodes unless a previous invocation of `monerod` used `--tx-proxy` or `--proxy` with the same Tor address. In most situations, the original check correctly identified an issue (with the one exception stipulated previously).

PR #9553, opened 2024-11-04
blake2b: compress and wipe keys immediately

jeffro256:
Changes keyed hashing such that:
  - Keys are not copied into a temporary 128 byte stack buffer, nor the hash state buffer
  - Keys are compressed immediately during initialization, not during `blake2b_final` or `blake2b_update`
  - Internal calculation vectors `m` and `v` are wiped
  - Caller of `blake2b_init_key` must commit to whether other data besides the key is of length 0 or not

These changes should make keyed Blake2b hashing more memory secure.

@UkoeHB

### DIFF
--- a/src/crypto/blake2b.c
+++ b/src/crypto/blake2b.c
@@ -286,7 +286,34 @@ int blake2b_init(blake2b_state *S, size_t outlen) {
 	return blake2b_init_param(S, &P);
 }
 
-int blake2b_init_key(blake2b_state *S, size_t outlen, const void *key, size_t keylen) {
+/* Fwd */
+static void blake2b_compress_mv(blake2b_state *S, const uint64_t *m, uint64_t * restrict v);
+
+/**
+ * The difference from `blake2b_compress()` is that this function directly takes
+ * a variable-sized input, treating it as a block where the remaining bytes are
+ * zeros, and also wipes stack variables `m` and `v`. The intended use case is
+ * initializing hash states with keys, hence the stricter memory handling. I
+ * don't attempt to prevent swapping of the variables `m` and `v` since the OS
+ * would have to be dumber than a box of rocks to swap stack variables used just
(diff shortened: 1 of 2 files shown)


src/crypto/blake2b.c
```diff
@@ -286,7 +286,83 @@ int blake2b_init(blake2b_state *S, size_t outlen) {
...
+ * would have to be dumber than a box of rocks to swap stack variables used just
+ * microseconds ago. `keylen` is assumed to be less than or equal to 128.
+ */
+static void blake2b_compress_key(blake2b_state *S, const uint8_t *key, size_t keylen) {
+	uint64_t m[16];
+	uint64_t v[16];
+	unsigned int i, r;
+
+	memset(m, 0, sizeof(m));
+#if defined(NATIVE_LITTLE_ENDIAN)
```
jeffro256, line 305: To test the big-endian loading code, set this line to `#if 0`, compile, then run the hash tests

UkoeHB:
Is there an origin/backstory to these changes, or is it fresh code out of the blue?

jeffro256:
It's fresh code out of the blue. Was investigating how Blake2 does keyed hashing, and observed that the implementation's handling of key memory wasn't ideal. 

Latest commits shrink diff and remove (probably premature) optimization for ease of review.

0xFFFC0000:
My sincere apologies for being a bummer here. But I *strongly* disagree with touching crypto code: 

1. Unless we are addressing a specific issue. 
2. Unless it goes through a very extensive review.

P.S. That is why I always prefer to use well-established, thoroughly reviewed third-party libraries for every cryptographic primitive we use.

UkoeHB:
I am also hesitant. Reviewing this properly probably requires validating the raw assembly (which is outside of my wheelhouse).

jeffro256:
Okay, I will close this for now then

PR #9559, opened 2024-11-08
carrot_core

jeffro256:
This PR is to implement all Carrot cryptography, as well as enote creation logic and enote scanning logic. This PR should exclude very Monero-specific integration details, except for domain separators. The `carrot_core` library is largely meant to be a stand-alone library, and thus only includes headers from `src/crypto`. 

https://github.com/jeffro256/carrot/blob/master/carrot.md

- [x]  Non-legacy carrot enote creation
- [x]  Non-legacy carrot enote scanning
- [x] Tests for new enotes addressed to legacy cryptonote accounts/addresses
- [x] Transaction output set finalization logic
- [x] Efficient implementation for X25519 unclamped scalar multiplication
- [x] Hardware device interface
- [x] Switch commitments (walmart version)
- [x] Internal self-send messages
- [x] Fix Debian 10 build
- [x] Unbind the special janus anchor derivation from the account spend pubkey to make hybrid accounts easier
- [x] Rewrite ARM mx25519 implementation to support unclamped scalar multiplications
- [x] ~~Deterministic FCMP++ output rerandomizations to support some refund address schemes~~
- [x] ~~Low-dependency payment proposal to enote finalization C interface for hardware wallets~~
- [x] ~~Payment proofs to internal enotes which A) allow for burning bug validation and B) don't reveal $s_{vb}$~~
- [x] Refactor for explicit unclamped mx25519 API: https://github.com/tevador/mx25519/pull/15

Depends on #9826, #9827, #9828 , #10108, #10111, #10338, #10963, #10964, #10965, #11155


jeffro256:
I'm pretty sure the Debian build error is caused by this GCC bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83258

jeffro256:
Today, I force-pushed two changes:
* Remove `src/carrot_core/lazy_amount_commitment.*`
* Add output size static assert line in `src/carrot_core/hash_functions.cpp`

UkoeHB (approved):

UkoeHB (approved):

jeffro256:
Rebased, fixed compile issue with batch invert tests, and upstreamed changes from `fcmp++-stage`

src/carrot_core/destination.cpp
```diff
@@ -42,6 +46,18 @@
 #undef MONERO_DEFAULT_LOG_CATEGORY
 #define MONERO_DEFAULT_LOG_CATEGORY "carrot.dest"
 
+namespace {
+crypto::public_key genPk()
+{
+    crypto::public_key s;
```
UkoeHB: `ec_scalar`

src/carrot_core/enote_utils.cpp
```diff
@@ -101,6 +109,16 @@ static OtherPid convert_payment_id(const Pid &v)
 }
 //-------------------------------------------------------------------------------------------------------------------
 //-------------------------------------------------------------------------------------------------------------------
+static crypto::public_key scalar_mult_gt(const crypto::ec_scalar &x, const crypto::ec_scalar &y)
```
UkoeHB, line 116: `// returns x G + y T`

src/carrot_core/enote_utils.cpp
```diff
@@ -56,6 +55,15 @@ namespace carrot
 {
 //-------------------------------------------------------------------------------------------------------------------
 //-------------------------------------------------------------------------------------------------------------------
+static const ge_p3 H_p3 = crypto::get_H_p3();
+static const ge_p3 T_p3 = crypto::get_T_p3();
+static const unsigned char l[32] = { // curve order
```
UkoeHB: Maybe make this `L` or `g_l` to make it more clearly a global at use-sites. Or put it in a namespace `consts`.

UkoeHB (changes requested):
Reviewed `carrot_core: no ringct dependency and faster legacy sync`. Not totally sure why removing ringct was necessary, it makes reading crypto ops more tedious.

UkoeHB (approved):

jeffro256:
Rebased and fixed merge conflicts due to recent FCMP++ crypto merges.

thomasbuilds (commented):
doing what I can here on this - asked Fable to review:

- UB on invalid subaddress spend pubkey: make_carrot_enote_ephemeral_pubkey_subaddress ignores ge_frombytes_vartime's return, then scalar-multiplies an uninitialized ge_p3. Sending to a malformed subaddress produces conditional jump on uninitialised value before a later, incidental check (try_make_carrot_onetime_address) throws. Nothing malformed escapes today, but the guard is accidental and the function is public API — check the return value and fail explicitly.
- Secret hygiene: derive_scalar leaves its 64-byte pre-reduction hash on the stack unwiped, inconsistent with FixedTranscript's destructor wiping and the auto_wiper discipline elsewhere. Also FixedTranscript deletes assignment but not the copy constructor.
- Vartime scalar mult on secrets (hardening): scalar_mult_gt and commit_carrot_amount use ge_double_scalarmult_base_vartime with secret per-enote scalars. Leaking sender extensions would let a local timing observer link one-time addresses to recipient addresses. There's monero precedent (rct::addKeys2), so this is a known tradeoff — but worth an explicit comment or a constant-time variant for the scanning path.
- Static-init ordering: enote_utils.cpp initializes H_p3/T_p3/auto_mx25519_impl as namespace-scope statics. Calls into carrot from another TU's static initialization would race this; function-local statics would be safer, and mx25519_select_impl's result is never null-checked.
- Doc/comment nits: make_carrot_generateimage_preimage's comment says H_n but the code (correctly, per spec) uses H_32 — the spec writes anchor_sp = H(... || k_v) with k_v last while the code uses k_v as the Blake2b key — pinned by test vectors, but the spec notation should match the normative encoding; gen_carrot_payment_proposal_v1 ignores its num_random_memo_elements parameter.

jeffro256:
Rebased, updated convergence test vectors for new `T` generator, and implemented feedback from @thomasbuilds.

thomasbuilds (approved):

jeffro256:
Updated for explicit unclamped API changes to `mx25519`, and added `crypto::` wrapper API. Relevant recent changes are listed here: #10964 and #10965.

src/carrot_core/device.h
thomasbuilds: and here?

jeffro256:
Rebased, added commits from #10963, integrated return value from `ge_p3_to_x25519()` in `try_make_carrot_enote_ephemeral_pubkey_subaddress()` due to Cypherstack review changes in #9828 (thanks @thomasbuilds), and added unit test for trivial scan binding attacks.

src/carrot_core/scan_unsafe.cpp
```diff
@@ -0,0 +1,273 @@
...
+    make_carrot_sender_extension_t(s_sender_receiver_ctx,
+        enote.amount_commitment,
+        sender_extension_t_out);
+
+    // K^j_s = Ko - K^o_ext = Ko - (k^o_g G + k^o_t T)
+    if (!try_recover_address_spend_pubkey(enote.onetime_address,
+            sender_extension_g_out,
+            sender_extension_t_out,
+            address_spend_pubkey_out))
+        return false;
```
thomasbuilds, line 73: Spec 8.1 step 18 (prime-order subgroup check on K^j_s') is not enforced on the receiver path?

```suggestion
        return false;

    if (!verify_point_is_in_main_subgroup(address_spend_pubkey_out))
        return false;
```

.gitmodules
```diff
@@ -11,3 +11,7 @@
 [submodule "external/gtest"]
 	path = external/gtest
 	url = https://github.com/google/googletest.git
+[submodule "external/mx25519"]
+	path = external/mx25519
+	url = https://github.com/jeffro256/mx25519.git
+	branch = upstream_unclamped
```
thomasbuilds, line 17: 2c45ac024 is on tevador/mx25519 master (merged via PR 15) so this can point at the canonical repo?

```suggestion
	url = https://github.com/tevador/mx25519.git
```

jeffro256: This will get resolved in #10964

src/carrot_core/scan.cpp
```diff
@@ -0,0 +1,405 @@
...
+    else if (!verified_normal_janus)
+        return false;
+    else if (recovered_address_spend_pubkey != destination.address_spend_pubkey)
+        return false;
+    else if (check_pid && recovered_payment_id != destination.payment_id)
+        return false;
+    else if (recovered_enote_type != CarrotEnoteType::PAYMENT)
+        return false;
+
+    return true;
```
thomasbuilds, line 322: `enote_type_out` (declared at 288) is never assigned so on success the caller keeps whatever was on the stack. `sender_scan_main_address` / `sender_scan_subaddress` pass only because that byte happens to be 0 == `PAYMENT` (Valgrind flags the uninitialised read).

```suggestion
    else if (recovered_enote_type != CarrotEnoteType::PAYMENT)
        return false;

    enote_type_out = recovered_enote_type;
    return true;
```

tests/unit_tests/carrot_mock_helpers.cpp
```diff
@@ -0,0 +1,681 @@
...
+        .enote_ephemeral_privkey = std::nullopt,
+        .internal_message = std::nullopt
+    };
+}
+//----------------------------------------------------------------------------------------------------------------------
+namespace people
+{
+const mock_carrot_and_legacy_keys alice = gen_default_person();
+const mock_carrot_and_legacy_keys bob = gen_default_person();
+const mock_carrot_and_legacy_keys eve = gen_default_person();
```
thomasbuilds, line 677: Probable cause of the 34 macOS failures: since #10960 `std::hash<crypto::public_key>` is SipHash keyed in `init_random()` (`constructor(101)`), and Mach-O ignores constructor priorities, so `people::alice/bob/eve` are generated here at static init with the zero key; `init_random()` then re-keys and every `get_index_for_address_spend_pubkey` lookup misses, so `can_open_fcmp_onetime_address` returns false. Proposed fix: generate the people from a gtest global `Environment` instead of at static init; patch below, applies to d002730e9. Any other `unordered_*` keyed by a `crypto::*` type and filled at static init has the same problem on macOS

<details><summary>patch</summary>

```diff
diff --git a/tests/unit_tests/carrot_mock_helpers.h b/tests/unit_tests/carrot_mock_helpers.h
index 3e060a88a..a521e9efd 100644
--- a/tests/unit_tests/carrot_mock_helpers.h
+++ b/tests/unit_tests/carrot_mock_helpers.h
@@ -212,9 +212,13 @@ CarrotPaymentProposalSelfSendV1 convert_selfsend_payment_proposal_v1(const crypt
 //----------------------------------------------------------------------------------------------------------------------
 namespace people
 {
-extern const mock_carrot_and_legacy_keys alice;
-extern const mock_carrot_and_legacy_keys bob;
-extern const mock_carrot_and_legacy_keys eve;
+// Generated by a gtest global Environment, NOT at static-init time: before init_random() runs, the RNG is unseeded
+// and std::hash<crypto::*> (SipHash keyed by crypto_siphash_key) changes value, which makes any unordered_map built
+// that early unreadable afterwards. Mach-O has no constructor priorities, so on macOS the test TUs' globals are
+// initialized before init_random().
+extern mock_carrot_and_legacy_keys alice;
+extern mock_carrot_and_legacy_keys bob;
+extern mock_carrot_and_legacy_keys eve;
 //----------------------------------------------------------------------------------------------------------------------
 } //namespace people
 } //namespace mock
diff --git a/tests/unit_tests/carrot_mock_helpers.cpp b/tests/unit_tests/carrot_mock_helpers.cpp
index b03fd69a0..a3f55c78c 100644
--- a/tests/unit_tests/carrot_mock_helpers.cpp
+++ b/tests/unit_tests/carrot_mock_helpers.cpp
@@ -38,6 +38,7 @@
 #include "ringct/rctOps.h"
 
 //third party headers
+#include "gtest/gtest.h"
 
 //standard headers
 
@@ -48,14 +49,6 @@ namespace
 {
 //----------------------------------------------------------------------------------------------------------------------
 //----------------------------------------------------------------------------------------------------------------------
-static carrot::mock::mock_carrot_and_legacy_keys gen_default_person()
-{
-    carrot::mock::mock_carrot_and_legacy_keys guy;
-    guy.generate();
-    return guy;
-}
-//----------------------------------------------------------------------------------------------------------------------
-//----------------------------------------------------------------------------------------------------------------------
 } //anonymous namespace
 
 namespace carrot
@@ -672,10 +665,25 @@ CarrotPaymentProposalSelfSendV1 convert_selfsend_payment_proposal_v1(const crypt
 //----------------------------------------------------------------------------------------------------------------------
 namespace people
 {
-const mock_carrot_and_legacy_keys alice = gen_default_person();
-const mock_carrot_and_legacy_keys bob = gen_default_person();
-const mock_carrot_and_legacy_keys eve = gen_default_person();
-}
+mock_carrot_and_legacy_keys alice;
+mock_carrot_and_legacy_keys bob;
+mock_carrot_and_legacy_keys eve;
+
+namespace
+{
+struct people_environment final : public ::testing::Environment
+{
+    void SetUp() override
+    {
+        alice.generate();
+        bob.generate();
+        eve.generate();
+    }
+};
+[[maybe_unused]] const ::testing::Environment *const people_environment_registration
+    = ::testing::AddGlobalTestEnvironment(new people_environment);
+} //anonymous namespace
+} //namespace people
 //----------------------------------------------------------------------------------------------------------------------
 } //namespace mock
 } //namespace carrot
\ No newline at end of file
```
</details>

tests/unit_tests/carrot_core.cpp
```diff
@@ -0,0 +1,3404 @@
...
+    ASSERT_EQ(payment_proposal.amount, recovered_amount);
+    ASSERT_EQ(CarrotEnoteType::PAYMENT, recovered_enote_type);
+    ASSERT_TRUE(bob.can_open_fcmp_onetime_address(payment_proposal.destination.address_spend_pubkey,
+        sender_extension_g,
+        sender_extension_t,
+        enote_proposal.enote.onetime_address));
+}
+//----------------------------------------------------------------------------------------------------------------------
+/*
+TEST(carrot_core, sender_scan_selfsend_2out)
```
thomasbuilds, line 3271: Cannot pass as written: the `_sender` overloads reject anything but `PAYMENT` (scan.cpp:319) and check only normal Janus anchors (scan.h:123) while Alice's change enote is a special enote. Delete it or scan Alice's enote through the receiver path rather than leaving it commented out?

tests/unit_tests/carrot_core.cpp
```diff
@@ -0,0 +1,3404 @@
...
+{
+    // An enote with an "identity" D_e and identity K^j_s shouldn't scan for a rando account,
+    // otherwise DoS on subaddress-map-less scanners is possible.
+
+    // (It's not really an identity, but it acts like it in the Montgomery ladder for scalar-point multiplication)
+    static constexpr crypto::x25519_pubkey x25519_identity = {};
+    // See rct::I
+    static constexpr crypto::public_key ed25519_identity = {
+        { 0x01}
+    };
```
thomasbuilds, line 3082: This pins only the doubly-protected case: `K^j_s'` = identity is rejected both by enote_utils.cpp:212 and by `ge_p3_to_x25519` (:225). The subgroup check at :217 is the only thing rejecting the other small-order points and nothing exercises it: replacing it with a plain `ge_frombytes_vartime` leaves all 92 tests of the PR's carrot/x25519 suites green, while `D_e` = 0 with `K^j_s'` = (0, -1) (`ecff...7f`) scans successfully whenever `d_e'` is odd. Add that case with a fixed odd-`d_e'` anchor (`gen_janus_anchor()` at :3117 would catch a regression only half the time).

tests/unit_tests/carrot_core.cpp
```diff
@@ -0,0 +1,3404 @@
...
+        .randomness = gen_janus_anchor()
+    };
+
+    // selfsend payment proposal to alice
+    const CarrotPaymentProposalSelfSendV1 alice_payment_proposal = CarrotPaymentProposalSelfSendV1{
+        .destination_address_spend_pubkey = alice_address.address_spend_pubkey,
+        .amount = crypto::rand_idx<rct::xmr_amount>(1000000),
+        .enote_type = CarrotEnoteType::CHANGE,
+        .enote_ephemeral_privkey = get_enote_ephemeral_privkey(bob_payment_proposal, input_context),
+        .internal_message = alice_internal_selfsends ? std::make_optional(gen_janus_anchor()) : std::nullopt
```
thomasbuilds, line 758: `alice_selfsend_type` (708) is unused: the type is hard-coded `CHANGE` here, so the 12 tests at 834-903 run the CHANGE case twice and never a PAYMENT self-send beside a normal payment (that path works when exercised by hand). `.is_subaddress` is also missing, so the `sub2*` cases build a subaddress self-send with `is_subaddress = false`.

tests/unit_tests/carrot_legacy.cpp
```diff
@@ -0,0 +1,292 @@
...
+        .randomness = gen_janus_anchor()
+    };
+
+    // selfsend payment proposal to alice
+    const CarrotPaymentProposalSelfSendV1 alice_payment_proposal = CarrotPaymentProposalSelfSendV1{
+        .destination_address_spend_pubkey = alice_address.address_spend_pubkey,
+        .amount = crypto::rand_idx<rct::xmr_amount>(1000000),
+        .enote_type = CarrotEnoteType::CHANGE,
+        .enote_ephemeral_privkey = get_enote_ephemeral_privkey(bob_payment_proposal, input_context)
+    };
```
thomasbuilds, line 182: Same as carrot_core.cpp:753-758: `alice_selfsend_type` (136) unused, `CHANGE` hard-coded (257-291 run it twice), `.is_subaddress` missing. Line 181 also sets `enote_ephemeral_privkey` explicitly (see payment_proposal.cpp:259-261); the test should not need it.

src/carrot_core/payment_proposal.cpp
```diff
@@ -0,0 +1,549 @@
...
+{
+    // if there is an other normal, use that
+    CARROT_CHECK_AND_THROW(!other_normal_payment_proposal || !other_self_send_proposal,
+        std::logic_error, "Other payment proposal in a 2-out tx may not be both normal, and a self-send");
+    if (other_normal_payment_proposal)
+    {
+        const input_context_t input_context = make_carrot_input_context(tx_first_key_image);
+        CARROT_CHECK_AND_THROW(!proposal.enote_ephemeral_privkey.has_value() || proposal.enote_ephemeral_privkey.value()
+                == get_enote_ephemeral_privkey(*other_normal_payment_proposal, input_context),
+            component_out_of_order, "Self-send in a 2-out tx with a normal has conflicting enote ephemeral privkey");
```
thomasbuilds, line 261: Inconsistent with `get_enote_ephemeral_privkey` (206-207): this getter tolerates an explicit `d_e` equal to the normal payment's, that one rejects any explicit `d_e`. With the proposal carrot_legacy.cpp:181 builds, `get_output_enote_proposals` succeeds but `get_enote_ephemeral_privkey(ss, &normal, ...)` throws. Apply one rule in both.

tests/unit_tests/carrot_core.cpp
```diff
@@ -0,0 +1,3404 @@
...
+        recovered_sender_extension_t,
+        enote_proposal.enote.onetime_address));
+}
+//----------------------------------------------------------------------------------------------------------------------
+TEST(carrot_core, subaddress_normal_scan_completeness)
+{
+    const mock::mock_carrot_and_legacy_keys &keys = alice;
+
+    const uint32_t j_major = crypto::rand_idx(mock::MAX_SUBADDRESS_MAJOR_INDEX);
+    const uint32_t j_minor = crypto::rand_idx(mock::MAX_SUBADDRESS_MINOR_INDEX);
```
thomasbuilds, line 230: `rand_idx` starts at 0, so (0,0), which `mock_carrot_and_legacy_keys::subaddress` maps to the main address, comes up ~1 % of the time and the "subaddress" test silently becomes a main-address test (same at 439-440, 585-586, 712-713, 725-726); `mock::gen_subaddress_index()` already guarantees non-zero. Likewise `tx_inout_proof_special_2out_1selfsend_*` pick Bob's address type with `rand<int>() & 1` (2621, 2748); run both shapes deterministically.

src/carrot_core/output_set_finalization.cpp
```diff
@@ -0,0 +1,400 @@
...
+
+        amount_blinding_factors.insert(output_enote_proposal.amount_blinding_factor);
+    }
+    CARROT_CHECK_AND_THROW(amount_blinding_factors.size() == num_proposals, missing_randomness,
+        "this set contains duplicate amount blinding factors");
+}
+//-------------------------------------------------------------------------------------------------------------------
+void get_coinbase_output_enotes(const std::vector<CarrotPaymentProposalV1> &normal_payment_proposals,
+    const uint64_t block_index,
+    std::vector<CarrotCoinbaseEnoteV1> &output_coinbase_enotes_out)
```
thomasbuilds, line 331: No test calls `get_coinbase_output_enotes`, directly or indirectly. (Exercised by hand it behaves; the one asymmetry with `get_output_enote_proposals` is that an empty proposal list yields an empty set instead of throwing `too_few_outputs`.)

tests/unit_tests/x25519.cpp
```diff
@@ -0,0 +1,253 @@
...
+      crypto::x25519_secret_key &s = scalars.emplace_back();
+      memset(s.data, 0, sizeof(crypto::x25519_secret_key));
+      const int msb_byte_index = i >> 3;
+      const int msb_bit_index = i & 7;
+      s.data[msb_byte_index] = 1 << msb_bit_index;
+      s.data[0] = j;
+    }
+  }
+  // add -1
+  scalars.push_back(hex2pod<crypto::x25519_scalar>("ecffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f"));
```
thomasbuilds, line 157: The scalar labelled "-1" is `p - 1`, not `l - 1` (comment only).

src/carrot_core/scan.cpp
```diff
@@ -0,0 +1,405 @@
...
+    const crypto::x25519_pubkey &s_sender_receiver,
+    crypto::secret_key &sender_extension_g_out,
+    crypto::secret_key &sender_extension_t_out,
+    xmr_amount &amount_out,
+    crypto::secret_key &amount_blinding_factor_out,
+    CarrotEnoteType &enote_type_out,
+    const bool check_pid)
+{
+    const crypto::public_key &main_address_spend_pubkey = destination.is_subaddress
+        ? crypto::null_pkey : destination.address_spend_pubkey;
```
thomasbuilds, line 292: An empty `epee::span<const crypto::public_key>{}` says what is meant better than a one-element `null_pkey` list

src/carrot_core/enote_utils.cpp
```diff
@@ -0,0 +1,696 @@
...
+    ge_add(R, &tmp1, &tmp2);
+}
+//-------------------------------------------------------------------------------------------------------------------
+//-------------------------------------------------------------------------------------------------------------------
+static bool verify_point_is_in_main_subgroup(const crypto::ec_point &P, ge_p3 &P_p3_out)
+{
+    // valid point?
+    if (0 != ge_frombytes_vartime(&P_p3_out, to_bytes(P)))
+        return false;
+    // 0 ?= l * K^j_v
```
thomasbuilds, line 106: Comment says `l * K^j_v` but this checks a generic point (copy-paste).

src/carrot_core/enote_utils.cpp
```diff
@@ -0,0 +1,696 @@
...
+{
+    // K^o_ext = k^o_g G + k^o_t T
+    crypto::public_key sender_extension_pubkey;
+    make_carrot_sender_extension_pubkey(s_sender_receiver_ctx, amount_commitment,
+        sender_extension_pubkey);
+    ge_p3 tmp1;
+    [[maybe_unused]] const int r = ge_frombytes_vartime(&tmp1, to_bytes(sender_extension_pubkey));
+    assert(0 == r); // shouldn't fail since we just serialized K^o_ext above
+    ge_cached tmp2;
+    ge_p3_to_cached(&tmp2, &tmp1);
```
thomasbuilds, line 403: `K^o_ext` is compressed by `make_carrot_sender_extension_pubkey` and immediately decompressed again with `ge_frombytes_vartime` (same at 350-358). Keeping the `ge_p3` from the double-scalar-mult, as `try_recover_address_spend_pubkey` does (558-586), saves a compression plus a variable-time decompression of a secret-derived point on the sender path.

src/carrot_core/enote_utils.cpp
```diff
@@ -0,0 +1,696 @@
...
+    static constexpr crypto::public_key identity{{1}};
+
+    // if K^j_s == 0
+    if (address_spend_pubkey == identity)
+        return false;
+
+    // decompress and check K^j_s
+    ge_p3 address_spend_pubkey_p3;
+    if (!verify_point_is_in_main_subgroup(address_spend_pubkey, address_spend_pubkey_p3))
+        return false;
```
thomasbuilds, line 218: Spec 8.1 step 18 (K^j_s' must be in the prime-order subgroup) is enforced here, inside the D_e recomputation, which the external receiver reaches through the normal Janus check; the internal `s_vb` scan never calls this, so it has no equivalent check. A comment saying so would help.

thomasbuilds (changes requested):

jeffro256:
Rebased

jpk68:
> Switch commitments (walmart version)

Just showing my appreciation for this joke :)

tests/unit_tests/carrot_mock_helpers.cpp
jeffro256: Resolved by #11369

PR #9570, opened 2024-11-14
contrib: fix serialization not calling load when only child is opt

0xFFFC0000:
The problem is `monero-wallet-rpc` does not save the state of the wallet when closing the wallet.

As you can see in the image below, the value of `autosave_current` is `false`. When in fact it should be `true` \[1\]. ( The command I am running is default command ` curl http://localhost:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"close_wallet"}' -H 'Content-Type: application/json'` )

image

The reason for this bug is when the _only_ child of the key-value serialization object is optional ( as is the case with the `close_wallet` ), we don't call the function `_load` of the object. This line \[2\] returns `null` and therefore we don't call `_load` method, which initializes the object.  Whether the child section is null or not, we want to call `_load` method, to make sure the correct initialization.

1. https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/src/wallet/wallet_rpc_server_commands_defs.h#L2199
2. https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/contrib/epee/include/serialization/keyvalue_serialization_overloads.h#L80

### DIFF
--- a/tests/unit_tests/epee_serialization.cpp
+++ b/tests/unit_tests/epee_serialization.cpp
@@ -77,6 +77,67 @@ struct ObjOfInts
     KV_SERIALIZE(x)
   END_KV_SERIALIZE_MAP()
 };
+
+template<typename t_param>
+struct ParentObjWithOptChild
+{
+  t_param     params;
+
+  ParentObjWithOptChild(): params{} {}
+
+  BEGIN_KV_SERIALIZE_MAP()
+    KV_SERIALIZE(params)
+  END_KV_SERIALIZE_MAP()
+};
+
+struct ObjWithOptChild
+{
+  bool test_value;
+
+  BEGIN_KV_SERIALIZE_MAP()
+    KV_SERIALIZE_OPT(test_value, true);
+  END_KV_SERIALIZE_MAP()
+};
+}
+
+TEST(epee_binary, serialize_deserialize)
+{
+  ParentObjWithOptChild<ObjWithOptChild> o;
+  std::string o_json;
+  o.params.test_value = true;
+
+  EXPECT_TRUE(epee::serialization::store_t_to_json(o, o_json));
+  EXPECT_TRUE(o.params.test_value);
+
+  EXPECT_TRUE(epee::serialization::load_t_from_json(o, o_json));
+  EXPECT_TRUE(o.params.test_value);
+
+  ParentObjWithOptChild<ObjWithOptChild> o2;
+  std::string o2_json;
+  o.params.test_value = false;
+
+  EXPECT_TRUE(epee::serialization::store_t_to_json(o2, o2_json));
+  EXPECT_FALSE(o2.params.test_value);
+
+  EXPECT_TRUE(epee::serialization::load_t_from_json(o2, o2_json));
+  EXPECT_FALSE(o2.params.test_value);
+
+  // compiler sets default value of test_value to false
+  ParentObjWithOptChild<ObjWithOptChild> o3;
+  std::string o3_json;
+
+  EXPECT_TRUE(epee::serialization::store_t_to_json(o3, o3_json));
+  EXPECT_FALSE(o3.params.test_value);
+
+  EXPECT_TRUE(epee::serialization::load_t_from_json(o3, o3_json));
+  EXPECT_FALSE(o3.params.test_value);
+
+  // test case with empty json, to test default value initialization
+  ParentObjWithOptChild<ObjWithOptChild> o4;
+  std::string o4_json = "{}";
+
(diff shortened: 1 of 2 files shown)


vtnerd:
This might be a bad idea. The [storage class assumes that `nullptr` means root](https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/contrib/epee/include/storages/portable_storage.h#L141) in several places. Couldn't this cause some loop to occur? I'd have to dig deeper.

More importantly, shouldn't this return a JSON error though, since there is no `"params":{}` field? This line that [doesn't check the return value](https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/contrib/epee/include/serialization/keyvalue_serialization.h#L79) seems odd, everything field is in optional then. Surprised I never noticed this.

iamamyth:
Normally JSON-RPC doesn't require params:
> params
    >    A Structured value that holds the parameter values to be used during the invocation of the method. This member MAY be omitted. 
> https://www.jsonrpc.org/specification

0xFFFC0000:
> This might be a bad idea. The [storage class assumes that `nullptr` means root](https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/contrib/epee/include/storages/portable_storage.h#L141) in several places. Couldn't this cause some loop to occur? I'd have to dig deeper.
> 
> More importantly, shouldn't this return a JSON error though, since there is no `"params":{}` field? This line that [doesn't check the return value](https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/contrib/epee/include/serialization/keyvalue_serialization.h#L79) seems odd, every field is in optional then. Surprised I never noticed this.

I have a question. `_load` is written in a way to handle `nullptr`, so why we should not run it? I don't see any reason to prevent running the `_load` when the value is `nullptr`, since `_load` itself does handle `nullptr`. 

1. https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/contrib/epee/include/serialization/keyvalue_serialization.h#L53

0xFFFC0000:
I added few testcases to PR. 

The important test is: 

```
  // test case with empty json, to test default value initialization
  ParentObjWithOptChild<ObjWithOptChild> o4;
  std::string o4_json = "{}";

  EXPECT_TRUE(epee::serialization::load_t_from_json(o4, o4_json));
  EXPECT_TRUE(o4.params.test_value);
  ```

File: tests/unit_tests/epee_serialization.cpp [hunk 1]
```diff
@@ -77,6 +77,67 @@ struct ObjOfInts
...
+
+  EXPECT_TRUE(epee::serialization::load_t_from_json(o3, o3_json));
+  EXPECT_FALSE(o3.params.test_value);
+
+  // test case with empty json, to test default value initialization
+  ParentObjWithOptChild<ObjWithOptChild> o4;
+  std::string o4_json = "{}";
+
+  EXPECT_TRUE(epee::serialization::load_t_from_json(o4, o4_json));
+  EXPECT_TRUE(o4.params.test_value);
```
Boog900, line 140: This is a bad test, this test requires that the `params` fields in `ParentObjWithOptChild` is optional (it is not specified as optional).

This test should make `load_t_from_json` error as the `params` field is missing and it is not an optional field.

0xFFFC0000: I see what you saying, but the point is we don't have ( or require ) param in like substantial of portion of our RPC commands. 

But if we are not able to initialize the `params` it can have serious consequences. Like the one I explained for `close_wallet`.

1. https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#close_wallet

Boog900: `params` here is not the same as the `params` in the JSON-RPC it is just the name of the field in the test struct.

`params` in the real JSON-RPC code should be an optional field, like what #9573 does.

0xFFFC0000: Yes, you are right. I am closing this PR. and re-opening our #9573 PR.

0xFFFC0000:
I am closing this PR. and re-opening #9573 PR.

vtnerd:
Even though this is closed, I will answer.

> I am not sure about the (infinite) loop thing. Can you explain it in a little bit more detail?

I don't think this is possible after further thought. However, it does "pollute" the root namespace into other contexts, so it's possibly still undesired.

Your other solution is funky, but at least fixes all of the problems without introducing other issues.

PR #9568, opened 2024-11-13
simplewallet: fix `command` argument and add `skip_check_backround_mining arg`

0xFFFC0000:

PR #9573, opened 2024-11-15
contrib: force (de)serialization to create params section incase there is none.

0xFFFC0000:
This is another solution to `close_wallet` not able to store the wallet due to serialization issues. 

For a detailed explanation look at https://github.com/monero-project/monero/pull/9568 .


0xFFFC0000:
https://github.com/monero-project/monero/pull/9574

PR #9574, opened 2024-11-15
contrib: force (de)serialization to create params section incase there is none.

0xFFFC0000:
The problem is `monero-wallet-rpc` does not save the state of the wallet when closing the wallet.

As you can see in the image below, the value of `autosave_current` is `false`. When in fact it should be `true` \[1\]. ( The command I am running is default command ` curl http://localhost:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"close_wallet"}' -H 'Content-Type: application/json'` )

image

The reason for this bug is when the _only_ child of the key-value serialization object is optional ( as is the case with the `close_wallet` ), we don't call the function `_load` of the object. This line \[2\] returns `null` and therefore we don't call `_load` method, which initializes the object.  Whether the child section is null or not, we want to call `_load` method, to make sure the correct initialization.

1. https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/src/wallet/wallet_rpc_server_commands_defs.h#L2199
2. https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/contrib/epee/include/serialization/keyvalue_serialization_overloads.h#L80

### DIFF
--- a/contrib/epee/include/net/http_server_handlers_map2.h
+++ b/contrib/epee/include/net/http_server_handlers_map2.h
@@ -165,6 +165,13 @@
       epee::serialization::store_t_to_json(static_cast<epee::json_rpc::error_response&>(rsp), response_info.m_body); \
       return true; \
     } \
+    epee::serialization::storage_entry params_; \
+    params_ = epee::serialization::storage_entry(epee::serialization::section()); \
+    if(!ps.get_value("params", params_, nullptr)) \
+    { \
+      epee::serialization::section params_section; \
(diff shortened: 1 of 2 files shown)


vtnerd:
The other solution would be

```c++
KV_SERIALIZE_OPT(params, t_param{})
```
in `jsonrpc_structs.h`. That might be better from a maintenance standpoint, but it's somewhat of a toss-up.

vtnerd (approved):

vtnerd (approved):

PR #9579, opened 2024-11-18
wallet: add no_initial_sync argument to open_wallet rpc command.

0xFFFC0000:
1. add no_initial_sync argument to open_wallet rpc command.
2. Clean --no-initial-sync implementation.
3. Fix spelling error.
   
The interface for `open_wallet` has stayed the same. Passing `no_initial_sync` to `open_wallet` RPC will prevent `open_wallet` to do a refresh after opening a wallet.


plowsof:
echoing my comment on the original `--no-inital-sync` pull request from mooo https://github.com/monero-project/monero/pull/8941#issuecomment-1627294195 to say this is great. Adding this as an on the fly flag per open wallet means that devs will see this flag in a more prominent place in docs, instead of in a separate startup flag section.  I will sanity check this now and update this comment

aleflm:
@plowsof  Yes, that is correct. 

Good suggestion and I updated the PR. 

Right now, when it does not sync, it will print a log : `no-initial-sync passed. Not syncing the wallet.`.

0xFFFC0000:
@plowsof 

I misread you question.
> I should not see the same logs output as the original open_wallet curl without no_initial_sync?

In original PR, the logs were exactly same. No special output for passing no-initial-sync. 

That is not the case anymore. Right now, when you pass `no-initial-sync` whether to RPC call or as wallet flag, you will see a `no-initial-sync passed. Not syncing the wallet`  in your log.

plowsof:
i confirm the 'time to begin syncing after opening a wallet' is significantly reduced with this PR. I used a rudimentary test grepping log files, comparing the opened wallet line to the "Pulled blocks" line that seemed to occur close enough to when wallet sync started. 

I'm now running a test to see any bandwidth usage difference with an open -> close wallet file continuous loop

PR #9583, opened 2024-11-19
Fix memcpy in byte_slice constructor

vtnerd:
There's _technically_ an overflow in `byte_slice.cpp` that I caught while simply scanning the code. This can occur if the total memory being copied exceeds `size_t`. This is unlikely to ever trigger, because the platform probably needs `sizeof(size_t) == 4 && sizeof(void*) == 8` (or similar) to occur. And somehow a `malloc` to copy all of this data needs to succeed too. Therefore, this is primarily just defensive programming to be strictly standards compliant.

### DIFF
--- a/contrib/epee/src/byte_slice.cpp
+++ b/contrib/epee/src/byte_slice.cpp
@@ -152,7 +152,11 @@ namespace epee
   {
     std::size_t space_needed = 0;
     for (const auto& source : sources)
+    {
+      if (std::numeric_limits<std::size_t>::max() - space_needed < source.size())
+        throw std::bad_alloc{};
       space_needed += source.size();
+    }
 
     if (space_needed)
     {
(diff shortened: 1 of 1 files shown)


vtnerd:
I wasn't up-to-date with master, so the last force push was a rebase.

selsta (approved):

iamamyth:
> This can occur if the total memory being copied exceeds size_t
If this happens, do you really want a silent failure? It'd mean the caller tried to produce a copy exceeding the whole of memory, very likely a bug, and probably unrecoverable.

vtnerd:
The next line catches the overflow. `remove_prefix` returns the number of bytes actually removed, which is less than requested in the overflow case.

iamamyth:
If you want to be paranoid about overflows, it would seem you need to check here for MAX wrapping back to 0:
https://github.com/monero-project/monero/pull/9583/files#diff-ae5d1c840afa6e07bbc29bc1432800ab4ef3bb779067e5e45b0a4943a6e65412L155

But if you include this check, based on my reading of allocate_slice and span, `out.remove_prefix` will never fail, so both the `min` and `remove_prefix` checks guard against the impossible.

vtnerd:
The idea was to catch the overflow later on the `remove_prefix` line. It gets caught either way.

iamamyth:
> It gets caught either way.

If it wraps to exactly zero, it's not caught, because the whole block is gated by an `if`.

vtnerd:
Yup, it has defined behavior in that situation, but it is still incorrect.

vtnerd:
Force pushed a different change.

iamamyth (approved):

0xFFFC0000 (approved):

PR #9589, opened 2024-11-23
Fixes tar extract cmd for systems with 16 bit uid_t

### LINKED ISSUE #8639: Podman Rootless Build fails
When using Podman to build, `podman build -t monero .` with Rootless containers it will fail due to user ID namespace constraints.

This can be solved by modifying: https://github.com/monero-project/monero/blob/365fd45b031b0a5c8104195dfabb786e839cb114/contrib/depends/funcs.mk#L81

and appending `--no-same-owner` to the `tar` command. 

```
 $(1)_extract_cmds ?= mkdir -p $$($(1)_extract_dir) && echo "$$($(1)_sha256_hash)  $$($(1)_source)" > $$($(1)_extract_dir)/.$$($(1)_file_name).hash &&  $(build_SHA256SUM) -c $$($(1)_extract_dir)/.$$($(1)_file_name).hash && tar --strip-components=1 --no-same-owner -xf $$($(1)_source) 
```

I'm unsure of what doing this may or may not break in standard build processes though, so I decided not to open a PR to patch for this specific case..

ishanarora:
One build dependency tarball, [protobuf-cpp-3.6.1.tar.gz](https://github.com/protocolbuffers/protobuf/releases/download/v3.6.1/protobuf-cpp-3.6.1.tar.gz)) contains file owner uid/gid as 231664/89939. This file fails to extract during the build process with the following error on systems where either uid_t or gid_t is limited to 16 bits (e.g. older kernels, rootless containers etc) i.e. the maximum uid_t/gid_t is 65534:
```
tar: six.BUILD: Cannot change ownership to uid 231664, gid 89939: Invalid argument
```

Since the file ownership is immaterial for the build process, this PR adds the `--no-same-owner` flag to `tar` command. This fixes #8639

### DIFF
--- a/contrib/depends/funcs.mk
+++ b/contrib/depends/funcs.mk
@@ -78,7 +78,7 @@ $(1)_download_path_fixed=$(subst :,\:,$$($(1)_download_path))
 
 #default commands
 $(1)_fetch_cmds ?= $(call fetch_file,$(1),$(subst \:,:,$$($(1)_download_path_fixed)),$$($(1)_download_file),$($(1)_file_name),$($(1)_sha256_hash))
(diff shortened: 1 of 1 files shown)


iamamyth (approved):
The build process shouldn't depend on the ownership info in the tarball, so this seems like a reasonable setting irrespective of the host platform.

PR #9592, opened 2024-11-24
simplewallet: add header to show_transfers command

0xFFFC0000:
An example: 

image

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -8688,6 +8688,22 @@ bool simple_wallet::show_transfers(const std::vector<std::string> &args_)
 
   PAUSE_READLINE();
 
+  auto formatter = boost::format("%8.8s %6.6s %8.8s %25.25s %20.20s %64.64s %15.15s %14.14s %s %s - %s");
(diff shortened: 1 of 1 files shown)


src/simplewallet/simplewallet.cpp
```diff
@@ -8688,6 +8688,22 @@ bool simple_wallet::show_transfers(const std::vector<std::string> &args_)
...
+  % "Timestamp"
+  % "Amount"
+  % "Tx Hash"
+  % "Tx Payment ID"
+  % "Tx Fee"
+  % "Destination(s)"
+  % "Index"
+  % "Tx Note";
+
+  formatter = boost::format("%8.8llu %6.6s %8.8s %25.25s %20.20s %64.64s %15.15s %14.14s %s %s - %s");
```
SChernykh, line 8705: Is this line needed? `formatter` is not used after it.

SChernykh: Nevermind, it is used below.

SChernykh (approved):

selsta (approved):

PR #9593, opened 2024-11-25
Keys: 2nd update for expired 'brunner7' key, valid 2 years

rbrunner7:
Extended 2 more years, after the key expired already back in September. Commit should be signed with this key.


0xFFFC0000:
Approved on Github.

SNeedlewoods (approved):

Issue #9596, opened 2024-11-28
Add support for upcoming Boost 1.87.0 (Boost.Asio removals)


cho-m:
Advance notice that Monero does not compile with upcoming Boost 1.87.0 (based on beta1) as Boost.Asio has removed the deprecated io_service https://github.com/boostorg/asio/commit/ec0908c562102915423d8bd7aefd3079efbb6c86

This causes errors like:
```
contrib/epee/include/net/net_utils_base.h:33:10: fatal error: 'boost/asio/io_service.hpp' file not found
   33 | #include <boost/asio/io_service.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

And impacts all usage of `boost::asio::io_service` (along with various other removed functions/methods)

io_service was deprecated in favor of io_context back in 1.66.0 (https://github.com/boostorg/asio/commit/b60e92b13ef68dfbb9af180d76eae41d22e19356). Easiest approach is to require a minimum of Boost 1.66.0 and use newer APIs. Otherwise, would need to create many `#if BOOST_VERSION ...` conditions for compatibility.

Related: https://www.boost.org/doc/libs/1_86_0/doc/html/boost_asio/net_ts.html

tobtoht:
Boost 1.87.0 was released a few days ago, so this will start breaking development builds and CI soon.

I'm not very familiar with this part of the codebase. @vtnerd maybe?

0xFFFC0000:
that `io_service` to `io_context` for boost is going to be a huge PR. We are using `io_service` in 381 different place, and as asio dev are saying the change is not backward compatible. https://github.com/boostorg/asio/issues/110

vtnerd:
There's more than just the `io_service` removal in this version of boost, so it's going to be rough change. As someone mentioned, we should probably bump the minmum boost version, some of the changes are going to be a tough to workaround otherwise.

tobtoht:
We can bump the minimum to 1.66.0 per #9446 and a PR to bump boost in release builds from 1.64.0 to 1.84.0 is ready.

vtnerd:
I'm working on a patch using Boost 1.66-1.87 as the target. Lots of changes being made, so it may take a while to test them.

vtnerd:
See #9628

PR #9623, opened 2024-12-16
ci: msys2: pin boost to 1.86.0

tobtoht:
- Boost >= 1.87.0 breaks builds (#9596)
- MSYS2 ships Boost 1.87.0
- Temporary downgrade to get CI to pass until a proper fix is in place

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,6 +69,13 @@ jobs:
       with:
         update: true
         install: mingw-w64-x86_64-toolchain make mingw-w64-x86_64-cmake mingw-w64-x86_64-ccache mingw-w64-x86_64-boost mingw-w64-x86_64-openssl mingw-w64-x86_64-zeromq mingw-w64-x86_64-libsodium mingw-w64-x86_64-hidapi mingw-w64-x86_64-protobuf mingw-w64-x86_64-libusb mingw-w64-x86_64-unbound git pkg-config
+    - shell: msys2 {0}
+      run: |
+        curl -O https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-boost-1.86.0-7-any.pkg.tar.zst
+        echo "3e84674b4d2b3ab82f4d5e22bcc2015fa139b6fd936c55d6b71f89a72a1ee0a2 mingw-w64-x86_64-boost-1.86.0-7-any.pkg.tar.zst" | sha256sum -c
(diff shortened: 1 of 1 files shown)


selsta (approved):

jeffro256 (changes requested):
Can we put an integrity check here please? Perhaps just a `sha256sum` or a PGP verify or something?

x64x2 (commented):
ueh

.github/workflows/build.yml
```diff
@@ -69,6 +69,12 @@ jobs:
       with:
         update: true
         install: mingw-w64-x86_64-toolchain make mingw-w64-x86_64-cmake mingw-w64-x86_64-ccache mingw-w64-x86_64-boost mingw-w64-x86_64-openssl mingw-w64-x86_64-zeromq mingw-w64-x86_64-libsodium mingw-w64-x86_64-hidapi mingw-w64-x86_64-protobuf mingw-w64-x86_64-libusb mingw-w64-x86_64-unbound git pkg-config
+    - shell: msys2 {0}
+      run: |
+        curl -O https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-boost-1.86.0-7-any.pkg.tar.zst
+        curl -O https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-boost-1.86.0-7-any.pkg.tar.zst.sig
```
vtnerd, line 75: Should a sha256 hash computation be done here?

vtnerd (changes requested):

tobtoht:
Using sha256 hash verification instead of gpg, as the latter would allow an attacker that has access to the msys2 repo to trick us into installing _any_ signed package. 

Assuming the release signing key isn't compromised (in which case everything is compromised), this would effectively allow an attacker to cause an older version of Boost to be installed while still passing CI. The older version of Boost could have a vulnerability that is not present in 1.86.0, and someone could unwittingly decide to run this development build outside of a testing environment.

I'm also downgrading `boost-libs` here as this package is required by `boost`, and the versions should not mismatch.

selsta (approved):
The files built don't get uploaded anyway so I don't think there is a legitimate attack vector here

tobtoht:
Ah right, it's the depends build that gets uploaded as an artifact, not this one. So yeah, the purpose of this job is only to check if Monero successfully builds on msys2.

PR #9162, opened 2024-02-12
depends: boost: update to 1.84.0

tobtoht:
- Context for this PR: https://libera.monerologs.net/monero-dev/20240210#c331216
- Relevant for #9596
- Needs #9457 to pass CI

PR #9628, opened 2024-12-17
Fix build with boost ASIO 1.87. Support boost 1.66+

vtnerd:
This should compile with Boost 1.73-1.87. I have only tested on Boost 1.74 and Boost 1.87 (primarily unit tests). I will mark the places where I know changes need to be made for older boost versions. I will clear up all such errors by manually downloading Boost 1.66 and testing.

One problem I have with supporting Boost 1.66-1.69 is that most types require an io_context object, whereas Boost 1.70+ have relaxed this requirement to an executor object. I will show an example of where it matters; it will be easy to break older Boost versions unintentionally.

This is currently marked as a draft, because of the items pointing out in a subsequent review.

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.h
+++ b/contrib/epee/include/net/abstract_tcp_server2.h
@@ -47,6 +47,7 @@
 #include <condition_variable>
 
 #include <boost/asio.hpp>
+#include <boost/asio/post.hpp>
 #include <boost/asio/ssl.hpp>
 #include <boost/asio/strand.hpp>
 #include <boost/asio/steady_timer.hpp>
@@ -100,8 +101,8 @@ namespace net_utils
     using ec_t = boost::system::error_code;
     using handshake_t = boost::asio::ssl::stream_base::handshake_type;
 
-    using io_context_t = boost::asio::io_service;
-    using strand_t = boost::asio::io_service::strand;
+    using io_context_t = boost::asio::io_context;
+    using strand_t = boost::asio::io_context::strand;
     using socket_t = boost::asio::ip::tcp::socket;
 
     using network_throttle_t = epee::net_utils::network_throttle;
@@ -267,13 +268,14 @@ namespace net_utils
       bool stop_signal_sent;
     };
 
-    /// Construct a connection with the given io_service.
-    explicit connection( boost::asio::io_service& io_service,
+    /// Construct a connection with the given io_context.
+    explicit connection( boost::asio::io_context& io_context,
                         std::shared_ptr<shared_state> state,
 			t_connection_type connection_type,
 			epee::net_utils::ssl_support_t ssl_support);
 
-    explicit connection( boost::asio::ip::tcp::socket&& sock,
+    explicit connection( boost::asio::io_context& io_context,
+      boost::asio::ip::tcp::socket&& sock,
 			 std::shared_ptr<shared_state> state,
 			t_connection_type connection_type,
 			epee::net_utils::ssl_support_t ssl_support);
@@ -306,7 +308,7 @@ namespace net_utils
     virtual bool close();
     virtual bool call_run_once_service_io();
     virtual bool request_callback();
-    virtual boost::asio::io_service& get_io_service();
+    virtual boost::asio::io_context& get_io_context();
     virtual bool add_ref();
     virtual bool release();
     //------------------------------------------------------
@@ -336,7 +338,7 @@ namespace net_utils
     /// serve up files from the given directory.
 
     boosted_tcp_server(t_connection_type connection_type);
-    explicit boosted_tcp_server(boost::asio::io_service& external_io_service, t_connection_type connection_type);
+    explicit boosted_tcp_server(boost::asio::io_context& external_io_context, t_connection_type connection_type);
     ~boosted_tcp_server();
     
     std::map<std::string, t_connection_type> server_type_map;
@@ -349,7 +351,7 @@ namespace net_utils
 	const std::string port_ipv6 = "", const std::string address_ipv6 = "::", bool use_ipv6 = false, bool require_ipv4 = true,
 	ssl_options_t ssl_options = ssl_support_t::e_ssl_support_autodetect);
 
-    /// Run the server's io_service loop.
+    /// Run the server's io_context loop.
     bool run_server(size_t threads_count, bool wait = true, const boost::thread::attributes& attrs = boost::thread::attributes());
 
     /// wait for service workers stop
@@ -409,7 +411,7 @@ namespace net_utils
       return connections_count;
     }
 
-    boost::asio::io_service& get_io_service(){return io_service_;}
+    boost::asio::io_context& get_io_context(){return io_context_;}
 
     struct idle_callback_conext_base
     {
@@ -417,7 +419,7 @@ namespace net_utils
 
       virtual bool call_handler(){return true;}
 
-      idle_callback_conext_base(boost::asio::io_service& io_serice):
+      idle_callback_conext_base(boost::asio::io_context& io_serice):
                                                           m_timer(io_serice)
       {}
       boost::asio::deadline_timer m_timer;
@@ -426,7 +428,7 @@ namespace net_utils
     template <class t_handler>
     struct idle_callback_conext: public idle_callback_conext_base
     {
-      idle_callback_conext(boost::asio::io_service& io_serice, t_handler& h, uint64_t period):
+      idle_callback_conext(boost::asio::io_context& io_serice, t_handler& h, uint64_t period):
                                                     idle_callback_conext_base(io_serice),
                                                     m_handler(h)
       {this->m_period = period;}
@@ -442,7 +444,7 @@ namespace net_utils
     template<class t_handler>
     bool add_idle_handler(t_handler t_callback, uint64_t timeout_ms)
       {
-        boost::shared_ptr<idle_callback_conext<t_handler>> ptr(new idle_callback_conext<t_handler>(io_service_, t_callback, timeout_ms));
+        boost::shared_ptr<idle_callback_conext<t_handler>> ptr(new idle_callback_conext<t_handler>(io_context_, t_callback, timeout_ms));
         //needed call handler here ?...
         ptr->m_timer.expires_from_now(boost::posix_time::milliseconds(ptr->m_period));
         ptr->m_timer.async_wait(boost::bind(&boosted_tcp_server<t_protocol_handler>::global_timer_handler<t_handler>, this, ptr));
@@ -461,14 +463,14 @@ namespace net_utils
     }
 
     template<class t_handler>
-    bool async_call(t_handler t_callback)
+    bool async_call(t_handler&& t_callback)
     {
-      io_service_.post(t_callback);
+      boost::asio::post(io_context_, std::forward<t_handler>(t_callback));
       return true;
     }
 
   private:
-    /// Run the server's io_service loop.
+    /// Run the server's io_context loop.
     bool worker_thread();
     /// Handle completion of an asynchronous accept operation.
     void handle_accept_ipv4(const boost::system::error_code& e);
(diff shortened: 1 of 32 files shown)


contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1194,79 +1198,71 @@ namespace net_utils
...
             boost::bind(&boosted_tcp_server<t_protocol_handler>::handle_accept_ipv6, this,
               boost::asio::placeholders::error));
       }
       catch (const std::exception &e)
       {
-        ipv6_failed = e.what();
+        MERROR("Failed to bind IPv6: " << e.what());
+        throw;
       }
     }
```
vtnerd, line 1265: There should be an `else` here that logs an an error and returns `false`.

contrib/epee/src/net_ssl.cpp
```diff
@@ -527,7 +528,7 @@ void ssl_options_t::configure(
       // preverified means it passed system or user CA check. System CA is never loaded
       // when fingerprints are whitelisted.
       const bool verified = preverified &&
-        (verification != ssl_verification_t::system_ca || host.empty() || boost::asio::ssl::rfc2818_verification(host)(preverified, ctx));
+        (verification != ssl_verification_t::system_ca || host.empty() || boost::asio::ssl::host_name_verification(host)(preverified, ctx));
```
vtnerd, line 531: This change should fail on Boost 1.66-1.73. A `#ifdef` will be needed to differentiate between the two names.

contrib/epee/src/net_helper.cpp
```diff
@@ -52,12 +23,12 @@ namespace net_utils
 			boost::promise<boost::asio::ip::tcp::socket> result_;
 			boost::asio::ip::tcp::socket socket_;
 
-			explicit new_connection(boost::asio::io_service& io_service)
+			explicit new_connection(const boost::asio::steady_timer::executor_type& io_service)
```
vtnerd, line 26: This should fail on the `socket_` constructor on Boost 1.66-1.69. The problem is that `socket_` will require an `io_context` object, whereas newer Boost versions only need an `executor`. This will likely be the source of many compile errors moving forward.

I'm not certain how I will fix this, I may have to change this function signature which may break some downstream projects.

tobtoht:
(This needs #9162 to pass depends CI. I will undraft that PR after the next round of merges.)

contrib/epee/include/net/abstract_tcp_server2.inl
vtnerd: Reverted to previous structure.

contrib/epee/src/net_helper.cpp
vtnerd: Fixed.

contrib/epee/src/net_ssl.cpp
vtnerd: Fixed.

vtnerd:
The latest force push should work with Boost 1.66-1.87. Something funky happened with the tests, so I may push a dummy change later to see what happens.

tobtoht:
https://github.com/tobtoht/monero/actions/runs/12390979989/job/34587144842#step:5:1281

Looks like `boost::asio::io_service` is also used in `src/device_trezor/trezor`.

contrib/epee/include/net/abstract_tcp_server2.h
```diff
@@ -267,13 +268,14 @@ namespace net_utils
       bool stop_signal_sent;
     };
 
-    /// Construct a connection with the given io_service.
-    explicit connection( boost::asio::io_service& io_service,
+    /// Construct a connection with the given io_context.
+    explicit connection( boost::asio::io_context& io_context,
```
jeffro256, line 272: Shouldn't we use the `io_context_t` alias here since we're changing this line anyways?

vtnerd: Updated.

vtnerd:
Fore pushed some changes:
  * Reverted changes to query code, they are basically same style with new interface
  * Switched to `io_context_t` as @jeffro256 suggested
  * Fix Trezor build.

vtnerd:
Force pushed twice with updates:

  * Added a missing include, but it doesn't matter as the depends builds is using Boost 1.64 which won't work
  * Reverted some whitespace changes that were made when making the patch.

vtnerd:
Force pushed an include fix.

vtnerd:
This patch looks ready. The depends CI will all fail until #9162 gets merged. I will leave as draft for now for that reason.

selsta (commented):
README needs an update to indicate the new minimum version

0xFFFC0000:
Once this is ready for review, please remove Draft status.

vtnerd:
Force pushed:
  - A rebase of recent `master` merges.
  - Reverted #9625 (no longer needed)
  - Reverted #9623 (no longer needed)
  - Requires #9162 for depends CI to pass

Likely this will get removed as draft, assuming some of the tests cycle green.

vtnerd:
> README needs an update to indicate the new minimum version

Forgot to update this in the last force push, so another quick push with this change.

vtnerd:
Now ready for review.

jeffro256:
Getting a ton of build errors across many platforms:

```
fatal error: boost/asio/io_context.hpp: No such file or directory
```

jeffro256:
The dependencies system seems to be pulling Boost 1.64 instead of 1.66 on these failed platforms

selsta:
@jeffro256 requires #9162 to be merged first

jeffro256:
Sorry, my bad. Should've read farther up in the comments

0xFFFC0000 (commented):
From technical perspective, everything is great, and I don't see any specific issues. 

The only remaining thing is I have to fetch #9162 and compile, and run all the tests.

contrib/epee/src/connection_basic.cpp
```diff
@@ -127,21 +121,21 @@ connection_basic_pimpl::connection_basic_pimpl(const std::string &name) : m_thro
...
+	strand_(io_context),
+	socket_(io_context, get_context(m_state.get())),
 	m_want_close_connection(false),
 	m_was_shutdown(false),
 	m_is_multithreaded(false),
 	m_ssl_support(ssl_support)
 {
 	// add nullptr checks if removed
-	assert(m_state != nullptr); // release runtime check in get_context
+	assert(m_state != nullptr); // release runtime check in get_contextp
```
jeffro256, line 136: typo?

contrib/epee/src/connection_basic.cpp
```diff
@@ -127,21 +121,21 @@ connection_basic_pimpl::connection_basic_pimpl(const std::string &name) : m_thro
...
 
-        socket_.next_layer() = std::move(sock);
+  socket_.next_layer() = std::move(sock);
```
jeffro256, line 138: what's happening with the whitespace here?

contrib/epee/src/net_helper.cpp
```diff
@@ -37,27 +53,20 @@ namespace net_utils
...
+
 		if (try_ipv6)
 		{
-			boost::asio::ip::tcp::resolver::query query6(boost::asio::ip::tcp::v6(), addr, port, boost::asio::ip::tcp::resolver::query::canonical_name);
-			iterator = resolver.resolve(query6);
-			if (iterator == end)
+			results = resolver.resolve(
+				boost::asio::ip::tcp::v6(), addr, port, boost::asio::ip::tcp::resolver::canonical_name
+			);
+			if (results.empty())
```
jeffro256, line 62: I think that this overload of `resolve()` shouldn't ever return an empty range because otherwise, it will throw a boost exception. That would make this `if` branch unreachable.

contrib/epee/src/net_helper.cpp
```diff
@@ -37,27 +53,20 @@ namespace net_utils
...
 				throw boost::system::system_error{boost::asio::error::fault, "Failed to resolve " + addr};
 		}
 
 		//////////////////////////////////////////////////////////////////////////
 
-		struct new_connection
```
jeffro256, line 50: Was the move of this struct declaration purely for readability purposes?

jeffro256 (changes requested):
A couple actual comments, but mostly nitpicks

contrib/epee/include/net/abstract_tcp_server2.h
```diff
@@ -100,8 +101,8 @@ namespace net_utils
     using ec_t = boost::system::error_code;
     using handshake_t = boost::asio::ssl::stream_base::handshake_type;
 
-    using io_context_t = boost::asio::io_service;
-    using strand_t = boost::asio::io_service::strand;
+    using io_context_t = boost::asio::io_context;
+    using strand_t = boost::asio::io_context::strand;
```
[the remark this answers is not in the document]
vtnerd: Updated.

contrib/epee/include/net/abstract_tcp_server2.h
```diff
@@ -306,7 +308,7 @@ namespace net_utils
     virtual bool close();
     virtual bool call_run_once_service_io();
     virtual bool request_callback();
-    virtual boost::asio::io_service& get_io_service();
+    virtual boost::asio::io_context& get_io_context();
```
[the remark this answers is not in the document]
vtnerd: Updated.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1767,7 +1776,7 @@ namespace net_utils
     if (r)
     {
       new_connection_l->get_context(conn_context);
-      //new_connection_l.reset(new connection<t_protocol_handler>(io_service_, m_config, m_sock_count, m_pfilter));
+      //new_connection_l.reset(new connection<t_protocol_handler>(io_context_, m_config, m_sock_count, m_pfilter));
```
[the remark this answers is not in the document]
vtnerd: Done.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1796,14 +1805,16 @@ namespace net_utils
...
+    boost::asio::ip::tcp::resolver resolver(io_context_);
+    boost::asio::ip::tcp::resolver::results_type results{};
     boost::system::error_code resolve_error;
-    boost::asio::ip::tcp::resolver::iterator iterator;
+
     try
     {
       //resolving ipv4 address as ipv6 throws, catch here and move on
-      iterator = resolver.resolve(query, resolve_error);
+      results = resolver.resolve(
```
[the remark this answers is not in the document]
vtnerd: I initially started to do cleanup in some of these DNS queries, but then backed them all out because I think it's more appropriate in a separate patch. That way the commits-per-line match with a better explanation, etc.

tests/net_load_tests/net_load_tests.h
```diff
@@ -32,7 +32,7 @@
 
 #include <atomic>
 
-#include <boost/asio/io_service.hpp>
+#include <boost/asio/io_context.hpp>
```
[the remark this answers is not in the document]
vtnerd: Doesn't look like its needed, removed.

contrib/epee/src/connection_basic.cpp
vtnerd: Yup, fixed.

contrib/epee/src/net_helper.cpp
vtnerd: No, the `struct` needed a templated parameter because the type is `boost::asio::io_context` or `boost::asio::executor` depending on the Boost version. The compiler complained about templated functions inside of a function declaration, so I moved it.

contrib/epee/src/net_helper.cpp
vtnerd: Again, I was trying to minimize unrelated changes in this PR. I'd rather this be done independently and tested, etc.

contrib/epee/src/connection_basic.cpp
vtnerd: I'm not sure why I ever changed this, but this should be reverted now.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1671,14 +1677,16 @@ namespace net_utils
...
```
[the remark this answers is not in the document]
vtnerd: Same answer as other - I'd rather do this in a separate PR.

jeffro256 (approved):

tobtoht (approved):

selsta (approved):

vtnerd:
@selsta should this be done on the release branch? I don't remember what was decided.

selsta:
@vtnerd yes, please also open against release

PR #9647, opened 2024-12-24
guix: only include python-minimal in android build env

tobtoht:
Requires #9477. No other packages are removed from the build environments of other targets because of this change.

### DIFF
--- a/contrib/guix/manifest.scm
+++ b/contrib/guix/manifest.scm
@@ -280,7 +280,6 @@ chain for " target " development."))
 
         ;; Scripting
         perl           ; required to build openssl in depends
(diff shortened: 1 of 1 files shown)


contrib/guix/manifest.scm
```diff
@@ -280,7 +280,6 @@ chain for " target " development."))
 
         ;; Scripting
         perl           ; required to build openssl in depends
-        python-minimal ; required to build monero (cmake/CheckTrezor.cmake) and in android_ndk
```
tobtoht, line 283: The comment here forgot to mention it is also used to build libtapi.

PR #9660, opened 2024-12-28
validate ip parameter in set_bans rpc call

eversinc33:
The `set_bans` method does not validate if there is actually a host or ip passed as a parameter. If a faulty RPC call fails to provide one of those, it falls back to banning `0.0.0.0`:

image

IMO the better way to handle this would be to provide an error message upon a missing parameter, which I added here:

image

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -2875,6 +2875,12 @@ namespace cryptonote
       }
       else
       {
+        if (!i->ip)
+        {
+          error_resp.code = CORE_RPC_ERROR_CODE_WRONG_PARAM;
+          error_resp.message = "No ip/host supplied";
+          return false;
+        }
         na = epee::net_utils::ipv4_network_address{i->ip, 0};
       }
       if (i->ban)


eversinc33:
Theres quite a few places where, if wrong parameters are supplied or parameters are missing, a faulty error message is thrown or, as in this case, an action is taken that is plain wrong. If this is merged and OK as a fix, I would go for fixing the other bugs too.

selsta (approved):

Asurar0 (approved):

iamamyth:
In most cases, I think the API should validate the data prior to performing any action, otherwise you end up with partially completed actions and the caller has no idea as to the effect of the call. For example, in this case, if you pass in 10 IPs, and the 5th one is rejected, you get the first four banned, and the rest not, but the error response would be the same with zero bans, or 9. Because this PR didn't introduce the "partial ban" behavior, I think it's OK, but, as a whole, the API could benefit from a consistent approach with batch APIs (e.g. return a status for each element in the batch, or pre-validate the whole batch and do nothing if invalid).

iamamyth:
> That is actually a very valid point - in this specific case, maybe ignoring the faulty element and moving on is the better approach.

Yes, I think what you're doing here makes sense because it's the backwards-compatible choice.

0xFFFC0000 (approved):

jeffro256 (approved):

selsta:
Please also open this PR against the `release-v0.18` branch.

PR #9663, opened 2024-12-30
Fix get_database_size on Windows

iamamyth:
Replace all calls to epee::file_io::get_file_size with boost::filesystem::file_size in order to avoid lossy conversions from paths to strings, which tend to break filename resolution. This commit fixes a bug on Windows where the get_info RPC call reported a zero database size because BlockchainLMBD::get_database_size returned zero.

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -34,7 +34,6 @@
 #include <cstring>  // memcpy
 
 #include "string_tools.h"
-#include "file_io_utils.h"
 #include "common/util.h"
 #include "common/pruning.h"
 #include "cryptonote_basic/cryptonote_format_utils.h"
@@ -4536,12 +4535,11 @@ bool BlockchainLMDB::is_read_only() const
 
 uint64_t BlockchainLMDB::get_database_size() const
 {
-  uint64_t size = 0;
   boost::filesystem::path datafile(m_folder);
   datafile /= CRYPTONOTE_BLOCKCHAINDATA_FILENAME;
-  if (!epee::file_io_utils::get_file_size(datafile.string(), size))
-    size = 0;
-  return size;
+  boost::system::error_code ec{};
+  const boost::uintmax_t size = boost::filesystem::file_size(datafile, ec);
+  return (ec ? 0 : static_cast<uint64_t>(size));
 }
 
 void BlockchainLMDB::fixup()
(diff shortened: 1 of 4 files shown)


iamamyth:
Fixes https://github.com/monero-project/monero/issues/9513. Replacing boost::filesystem with std::filesystem across all platforms would require a more recent version of the Android NDK (can't recall the version, but there's an issue for it in the Android repo) and Xcode (10.5+, if memory serves).

iamamyth:
Worth noting: I think there are build system PRs in queue which would upgrade the relevant components to make std::filesystem available everywhere; however, linking the stdc++fs library properly across environments poses its own challenges, so I don't think the change from boost to std will be a simple find/replace.

tobtoht (approved):

0xFFFC0000 (approved):
LGTM

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -4536,12 +4535,11 @@ bool BlockchainLMDB::is_read_only() const
 
 uint64_t BlockchainLMDB::get_database_size() const
 {
-  uint64_t size = 0;
   boost::filesystem::path datafile(m_folder);
-  datafile /= CRYPTONOTE_BLOCKCHAINDATA_FILENAME;
-  if (!epee::file_io_utils::get_file_size(datafile.string(), size))
-    size = 0;
-  return size;
+  datafile /= boost::filesystem::path(CRYPTONOTE_BLOCKCHAINDATA_FILENAME);
```
jeffro256, line 4539: This conversion before appending isn't needed and is harder to read

src/common/download.cpp
```diff
@@ -73,8 +72,11 @@ namespace tools
     {
       boost::unique_lock<boost::mutex> lock(control->mutex);
       std::ios_base::openmode mode = std::ios_base::out | std::ios_base::binary;
-      uint64_t existing_size = 0;
-      if (epee::file_io_utils::get_file_size(control->path, existing_size) && existing_size > 0)
+      boost::system::error_code ec{};
+      uint64_t existing_size = static_cast<uint64_t>(boost::filesystem::file_size(boost::filesystem::path(control->path), ec));
```
jeffro256, line 76:
```suggestion
      const uint64_t existing_size = static_cast<uint64_t>(boost::filesystem::file_size(control->path, ec));
```

make `const` and remove explicit conversion to `path`

jeffro256 (approved):

Issue #9569, opened 2024-11-13
Monerod crashes when synchronizing on Windows.


Folindril:
When synchronizing the blockchain from 0, most of it is synchronized without problems. Towards the end, monerod starts to crash at random times without any errors. There were cases when the crash occurred in the middle of drawing a log line on the screen. There are also crashes with irreversible damage to the blockchain files. Work in a synchronized state is stable, the bug clearly occurs only when lagging behind the network. This bug occurs either on certain hardware or on a certain OS.This bug is present on all versions 0.18. Did not check earlier versions due to incompatibility of blockchain files.

Version: 0.18.3.4 Windows 64-bit
OS: Windows 11
Processor: Intel Core i5-7600K 3.80GHz
RAM: 16Gb

selsta:
Do you have any stability issues on your system? Or do you have a different computer where you can test syncing to see if it crashes? There is no known crash bugs in v0.18 that match what you are describing, at least we didn't get other reports about this on Windows.

Folindril:
> Do you have any stability issues on your system? Or do you have a different computer where you can test syncing to see if it crashes? There is no known crash bugs in v0.18 that match what you are describing, at least we didn't get other reports about this on Windows.

No, there are no problems with other applications (mostly games). Monero mining also works stably with a synchronized node.
I have a laptop in the same network, where synchronization occurs without crashes. No additional parameters in the command line helped me either. Therefore, I decided that this is a bug and it is tied to a certain architecture.

Folindril:
Now I have run synchronization again with log detalisation level 1 to detect on which block the first crash will occur.

selsta:
The problem is sometimes the last log message is not related to the crash, a stack trace would ideally be needed but that's a bit more complicated on Windows.

Folindril:
> Now I have run new synchronization again with log detalisation level 1 to detect on which block the first crash will occur. It will take a few days because crashes start towards the end.

Done. Block 1650925 (2018-08-31).
Log:

`2024-11-17 11:19:56.367 I +++++ BLOCK SUCCESSFULLY ADDED
2024-11-17 11:19:56.368 I id:   <6d920a7d01f688c5f3da5dbb4e7bca6cc67e19b104fc2ef9d56a89b4fad5d9de>
2024-11-17 11:19:56.368 I PoW:  <ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff>
[...]

PR #9664, opened 2024-12-30
Check for windows NTFS compression on monerod startup (fixes #9617)

### LINKED ISSUE #9617: Windows NTFS compression causes crash during sync
As reported here https://github.com/monero-project/monero/issues/9569#issuecomment-2539368523

It would be good to add a check to monerod to make sure the Windows file system, or the lmdb database does not have automatic compression enabled, otherwise it seems to crash during sync.

eversinc33:
Fixes #9617

If the data-dir is compressed and the lmdb folder is (check for both, because if data-dir is compressed and lmdb doesnt exist yet, it is likely that the compression is configured to be applied recursively):

image

If just the db folder is compressed:

image

If not compressed at all:

image

Just logging an ERROR message, just like if Fat32 is used, not aborting, because for a small blockchain this not be a critical issue. 

Can we assume that "lmdb" stays the name of the db folder? Since it is hardcoded here. Otherwise, one would need to define it somewhere central. I assumed so, because `BlockchainLMDB::get_db_name()` also has it hardcoded.

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -28,10 +28,15 @@
 #include "db_lmdb.h"
 
 #include <boost/filesystem.hpp>
+#include <boost/filesystem/fstream.hpp>
 #include <boost/format.hpp>
 #include <boost/circular_buffer.hpp>
 #include <memory>  // std::unique_ptr
 #include <cstring>  // memcpy
+
+#ifdef WIN32
+#include <winioctl.h>
+#endif
 
 #include "string_tools.h"
 #include "file_io_utils.h"
@@ -1321,6 +1326,54 @@ BlockchainLMDB::BlockchainLMDB(bool batch_transactions): BlockchainDB()
   m_hardfork = nullptr;
 }
 
+#ifdef WIN32
+static bool disable_ntfs_compression(const boost::filesystem::path& filepath)
+{
+  DWORD file_attributes = ::GetFileAttributesW(filepath.c_str());
+  if (file_attributes == INVALID_FILE_ATTRIBUTES)
+  {
+    MERROR("Failed to get " << filepath.string() << " file attributes. Error: " << ::GetLastError());
+    return false;
+  }
+  
+  if (!(file_attributes & FILE_ATTRIBUTE_COMPRESSED))
+    return true; // not compressed
+
+  LOG_PRINT_L1("Disabling NTFS compression for " << filepath.string());
+  HANDLE file_handle = ::CreateFileW(
+    filepath.c_str(),
+    GENERIC_READ | GENERIC_WRITE,
+    FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+    nullptr,
+    OPEN_EXISTING,
+    boost::filesystem::is_directory(filepath) ? FILE_FLAG_BACKUP_SEMANTICS : 0, // Needed to open handles to directories
+    nullptr
+  );
+
+  if (file_handle == INVALID_HANDLE_VALUE) 
+  {
+    MERROR("Failed to open handle: " << filepath.string() << ". Error: " << ::GetLastError());
+    return false;
+  }
+
+  USHORT compression_state = COMPRESSION_FORMAT_NONE;
+  DWORD bytes_returned;
+  BOOL ok = ::DeviceIoControl(
+    file_handle,
+    FSCTL_SET_COMPRESSION,
+    &compression_state,
+    sizeof(compression_state),
+    nullptr,
+    0,
+    &bytes_returned,
+    nullptr
+  );
+
+  ::CloseHandle(file_handle);
+  return ok;
+}
+#endif
+
 void BlockchainLMDB::open(const std::string& filename, const int db_flags)
 {
   int result;
@@ -1346,6 +1399,17 @@ void BlockchainLMDB::open(const std::string& filename, const int db_flags)
     LOG_PRINT_L0("Move " << CRYPTONOTE_BLOCKCHAINDATA_FILENAME << " and/or " << CRYPTONOTE_BLOCKCHAINDATA_LOCK_FILENAME << " to " << filename << ", or delete them, and then restart");
     throw DB_ERROR("Database could not be opened");
   }
+
+#ifdef WIN32
+  // ensure NTFS compression is disabled on the directory and database file to avoid corruption of the blockchain
+  if (!disable_ntfs_compression(filename))
+    LOG_PRINT_L0("Failed to disable NTFS compression on folder: " << filename << ". Error: " << ::GetLastError());
+  boost::filesystem::path datafile(filename);
+  datafile /= CRYPTONOTE_BLOCKCHAINDATA_FILENAME;
+  boost::filesystem::ofstream(datafile).close(); // touch the file to ensure it exists
+  if (!disable_ntfs_compression(datafile))
+    throw DB_ERROR("Database file is NTFS compressend and compression could not be disabled");
+#endif
 
   boost::optional<bool> is_hdd_result = tools::is_hdd(filename.c_str());
   if (is_hdd_result)


nahuhh:
Should probably only need to check the `data-dir/lmdb`

example `.bitmonero/lmdb`.

> Can we assume that "lmdb" stays the name of the db folder?

Testnet & stagenet will end up in subfolders like `.bitmonero/testnet/lmdb`

eversinc33:
> Should probably only need to check the `data-dir/lmdb`

If there is no edge case where data-dir` is created, but `lmdb` isnt yet, sure.

> Testnet & stagenet will end up in subfolders like `.bitmonero/testnet/lmdb`

That is taken care of with `data_dir / "lmdb"`

eversinc33:
Changed so that only the database directory is checked, and if it doesnt exist, it falls back to checking the data-dir.

iamamyth:
Either the software supports a compressed database/data dir, or it doesn't (I'm not certain as to the policy here, though compression + memory mapping seems like a bad idea); logging an ERROR and proceeding with startup doesn't make sense to me, and doesn't "fix" much of anything.

eversinc33:
My reasoning was that this error only seems to become a problem if the blockchain exceeds 70GB (see https://github.com/monero-project/monero/issues/9569#issuecomment-2539368523), so its not a reason to halt if e.g. running for a smaller chain such as the testnet (or if the on disk chain becomes smaller in the future due to e.g. further pruning improvements). Im very open to discussions as I am just getting familiar with the codebase.

iamamyth:
I think you're on the right track here, but, conceptually, the blockchain grows forever, so allowing startup in a configuration which will eventually break wastes the feature detection and proves quite irritating for users (mainnet is the most critical use case and it exceeds the failure threshold). I may have uncovered the underlying problem: "If you compress a file that is larger than 30 gigabytes, the compression may not succeed." (https://learn.microsoft.com/en-us/windows/win32/fileio/file-compression-and-decompression). Rather than simply warning the user, or aborting, I suggest transparently fixing the problem in the BlockchainLMDB class:

1. When creating the data directory in `open`, disable compression on the directory.
2. (Maybe) If the db file exists and is compressed, prior to mdb_env_open, disable compression on the db file. This would fix existing databases going forward, and may also guard against "manual surgery" cases where users put the system into an effectively broken state by fiddling with file attributes.

You can find info on the relevant windows APIs here:
* Setting directory compression state: https://learn.microsoft.com/en-us/windows/win32/fileio/creating-and-opening-files#file-attributes-and-directories
* Setting file compression state: https://learn.microsoft.com/en-us/windows/win32/fileio/compression-state

eversinc33:
Very good points, I think this is the better approach. I will start implementing this in the new year and update this PR

eversinc33:
Force pushed with the new approach.

image

Disables compression on both the directory as well as the blockchain file if set. If not, continues silently. Let me know what you think.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -1347,6 +1350,16 @@ void BlockchainLMDB::open(const std::string& filename, const int db_flags)
     throw DB_ERROR("Database could not be opened");
   }
 
+#ifdef WIN32
+  // ensure NTFS compression is disabled on the directory and database file to avoid corruption of the blockchain 
+  if (boost::filesystem::exists(filename) && !disable_ntfs_compression(filename))
```
iamamyth, line 1355: The earlier logic creates the directory (~line 1337), so you shouldn't need to check for existence (the extra check doesn't add race protection, and you don't need race protection).

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -1347,6 +1350,16 @@ void BlockchainLMDB::open(const std::string& filename, const int db_flags)
     throw DB_ERROR("Database could not be opened");
   }
 
+#ifdef WIN32
+  // ensure NTFS compression is disabled on the directory and database file to avoid corruption of the blockchain 
+  if (boost::filesystem::exists(filename) && !disable_ntfs_compression(filename)) 
+    LOG_PRINT_L0("Failed to disable NTFS compression on folder: " << filename << ". Error: " << ::GetLastError());
+  boost::filesystem::path datafile(filename);
+  datafile /= CRYPTONOTE_BLOCKCHAINDATA_FILENAME;
+  if (boost::filesystem::exists(datafile) && !disable_ntfs_compression(datafile))
```
iamamyth, line 1359: I would touch the file so that it always exists, otherwise the daemon might have a compressed db on first launch. Also, if disabling compression fails, I'd throw a DB_ERROR.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -1688,6 +1701,54 @@ std::string BlockchainLMDB::get_db_name() const
   return std::string("lmdb");
 }
 
+#ifdef WIN32
+bool BlockchainLMDB::disable_ntfs_compression(const boost::filesystem::path& filepath) const
```
iamamyth, line 1705: Would make this `static` and move it to precede open.

eversinc33:
Thank you again for the great feedback. Please see the latest commit.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -1347,6 +1399,17 @@ void BlockchainLMDB::open(const std::string& filename, const int db_flags)
     throw DB_ERROR("Database could not be opened");
   }
 
+#ifdef WIN32
+  // ensure NTFS compression is disabled on the directory and database file to avoid corruption of the blockchain 
+  if (!disable_ntfs_compression(filename)) 
+    LOG_PRINT_L0("Failed to disable NTFS compression on folder: " << filename << ". Error: " << ::GetLastError());
+  boost::filesystem::path datafile(filename);
+  datafile /= CRYPTONOTE_BLOCKCHAINDATA_FILENAME;
+  boost::filesystem::ofstream(datafile.string()).close(); // touch the file to ensure it exists
```
iamamyth, line 1408: I think you can simplify this to `ofstream(datafile)`, which avoids information loss from path->string conversion

iamamyth:
A clarification on handling directory-level compression for other reviewers: One could argue that if the directory already exists, the program should preserve its compression attribute; in fact, I suggested as much in earlier comments. However, upon further consideration, I think it makes sense to choose the most intrusive option (always uncompress the directory) because:

1. It helps ensure consistent daemon behavior (as bug reports, and the Windows docs themselves, have illustrated, Windows filesystem compression is unreliable and non-transparent).
2. If the choice proves unworkable, a future PR can downgrade it, whereas starting with the least intrusive option (mixed compression modes) needlessly adds complexity.

iamamyth:
Can you confirm that this modification runs normally on Windows and results in an an unencrypted data directory + db post-shutdown in the following startup scenarios?

* Empty data directory
    * Encrypted
    * Not encrypted
* Complete mainnet data directory
    * Encrypted
    * Not encrypted

src/blockchain_db/lmdb/db_lmdb.cpp
eversinc33: Fixed.

iamamyth:
In most projects, I'd add tests for new behavior. Here, the tests ordinarily don't run on windows, so extra tests don't do much. If you want to add tests, you have two mutually exclusive options:

1. Add windows-only tests to tests/unit_test/blockchain_db.cpp (probably simpler and a technically reasonable choice).
2. Add windows-only functional tests to tests/functional_tests (reflects the regression/integration aspect of the tests).

iamamyth (approved):

iamamyth:
There's a failing build: It looks like you need to explicitly add:
```
#include <boost/filesystem/fstream.hpp>
```

iamamyth (changes requested):
There's a failing build. It looks like you need to explicitly add:
```
#include <boost/filesystem/fstream.hpp>
```

Everything else looks good. Thanks for your work on this issue!

iamamyth:
One last thing: Probably a good idea to squash commits for merge.

iamamyth (approved):

eversinc33:
Fixed and squashed. Thanks again for the good feedback!

selsta (approved):

selsta:
Please also open this PR against `release-v0.18` branch.

tobtoht:
Lots of extraneous whitespace in this patch.

iamamyth:
> Lots of extraneous whitespace in this patch.

@tobtoht Was this posted on the wrong PR? I don't see any extraneous whitespace.

selsta:
There are a lot of spaces at the end of some lines, would be good to fix before merging.

nahuhh:
It seems to emulate codestyle from 1377-1402

iamamyth:
Now I see the whitespace (using CLI), I think the github web UI got worse and made it really hard to see. @eversinc33 Can you remove the extraneous end-of-line whitespace and squash?

eversinc33:
> Now I see the whitespace (using CLI), I think the github web UI got worse and made it really hard to see. @eversinc33 Can you remove the extraneous end-of-line whitespace and squash?

Sorry, not sure where that came from. Squashed and pushed

eversinc33:
> Please also open this PR against `release-v0.18` branch.

See https://github.com/monero-project/monero/pull/9702

selsta:
@eversinc33 Could you check again? This line for example still has whitespaces at the end of the line.

https://github.com/monero-project/monero/pull/9664/files#diff-94841b820d7dfb0ef2f631d635c246ea152fba2cf868b62a0dc084f04e9e2f6cR1364

eversinc33:
Weird. I force pushed again, can you check now?

selsta (approved):
Looks good now.

0xFFFC0000 (approved):
Thanks. LGTM.

PR #9671, opened 2025-01-03
depends: always pass --with-pic to autoconf builds

tobtoht:


0xFFFC0000 (commented):
Thanks for working on this. Question. Do we need `pic` for everything? Usually pic codeis more bloated.

tobtoht:
My understanding is that this flag instructs `libtool` to prefer `pic` objects, which may or may not fix a build issue. Setting the flag everywhere can't hurt.

See also the [libtool documentation](https://www.gnu.org/software/libtool/manual/html_node/LT_005fINIT.html#index-AM_005fPROG_005fLIBTOOL) (Note: `--with-pic` was [renamed](https://savannah.gnu.org/news/?id=10676) to `--enable-pic`)

Precision Independent Code is a [security feature](https://en.wikipedia.org/wiki/Position-independent_code#Position-independent_executables). Our release binaries are already PIE.

selsta (approved):

0xFFFC0000 (approved):
I am approving. 
 
If there was easy way to enable and disable pic when building `depends` that would’ve been great.

0xFFFC0000 (approved):

MrCyjaneK (approved):
Also in full support, I had to enable pic for contrib/depends manually when building wallet_api in monero_c in order to link for iOS and android, this change remove at least a couple lines of extra patches

selsta (approved):

jeffro256 (approved):

PR #9676, opened 2025-01-04
depends: zeromq: build with cmake

tobtoht:
Autotools / autogenerated code is a supply chain attack vector.

Requires #9677.


tobtoht:
Closing this as it would require too many patches to get working. Will wait for the next zeromq release.

PR #9680, opened 2025-01-04
ci: add arch linux build

tobtoht:
We don't currently have a rolling release Linux distro in our CI. Arch Linux is a common development environment.

This would help us catch and fix issues like #9359 early.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,6 +78,22 @@ jobs:
         ${{env.CCACHE_SETTINGS}}
         make release-static-win64 -j4
 
+  build-arch:
+    name: 'Arch Linux'
+    runs-on: ubuntu-latest
+    container:
+      image: archlinux:latest
+    steps:
+      - name: install dependencies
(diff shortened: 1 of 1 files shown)


hinto-janai (approved):

.github/workflows/build.yml
```diff
@@ -78,6 +78,22 @@ jobs:
...
+  build-arch:
+    name: 'Arch Linux'
+    runs-on: ubuntu-latest
+    container:
+      image: archlinux:latest
+    steps:
+      - name: install dependencies
+        run: pacman -Syyu --noconfirm base-devel git cmake boost openssl zeromq unbound libsodium readline expat gtest python3 doxygen graphviz hidapi libusb protobuf
+      - name: configure git
+        run: git config --global --add safe.directory '*'
```
vtnerd, line 90: Why is this step needed? Isn't the checkout phase from URL, and not local directory?

tobtoht: See: https://github.com/tobtoht/monero/actions/runs/12848743186/job/35826479151#step:5:71

vtnerd (approved):

PR #9690, opened 2025-01-09
build: auto-set GitHub runner make job count

iamamyth:
Dynamically determine an appropriate number of make jobs for the GitHub runner container, based upon the number of available CPU cores and RAM, using the rule that each job requires a dedicated core and 2.25GiB of RAM.

### DIFF
--- a/.github/actions/set-make-job-count/action.yml
+++ b/.github/actions/set-make-job-count/action.yml
@@ -1,0 +1,22 @@
+name: 'set-make-job-count'
+description: 'Set the MAKE_JOB_COUNT environment variable to a value suitable for the host runner'
+runs:
+  using: "composite"
+  steps:
+    # Each job runner requires 2.25 GiB (i.e. 1024 * 9/4 MiB) memory and
+    # a dedicated logical CPU core
+    - name: set-jobs-macOS
+      if: runner.os == 'macOS'
+      run: |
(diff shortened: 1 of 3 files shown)


selsta:
Seems this does not compile with Debian and Ubuntu CI jobs

iamamyth:
This PR follows from the discussion in https://github.com/monero-project/monero/pull/9510. Based on testing with GitHub macOS runners, I determined the suitable memory requirement to be 2.25GiB per runner, rather than the 2.75GiB hypothesized in the thread: With three jobs (current config) on a 3 CPU + 7GiB machine, a full build takes roughly `8m 24s`, versus `11m 10s` with two jobs. Due to noisy neighbors, there's considerable performance variance (observed 40s difference between two identical runs), but the results suggest 2.25GiB suffices.

Should the need arise, this PR allows custom memory requirements to be set per OS.

Build output confirms this PR results in job counts identical to the current fixed configuration (by design, the dynamically determined count appears in the build output because it's sourced from an environment variable).

plowsof:
MacOS reports in bytes and i was unable to find a flag/option for it to get another unit directly, vmstat can but the powershell command is required for windows. It would have made tests more trivial but i checked all calculations with their respective units and it performs correctly.

would you consider publishing this action as a public good in its own repository? https://docs.github.com/en/actions/sharing-automations/creating-actions/publishing-actions-in-github-marketplace

iamamyth:
> would you consider publishing this action as a public good in its own repository? https://docs.github.com/en/actions/sharing-automations/creating-actions/publishing-actions-in-github-marketplace

No.

selsta (approved):

tobtoht (approved):

PR #9691, opened 2025-01-09
crypto: make CRYPTO_DEFINE_HASH_FUNCTIONS adhere strict aliasing

jeffro256:
This code could've caused issues if the pointer to the `public_key`, `key_image`, `hash`, etc wasn't aligned on an 8-byte boundary.

### DIFF
--- a/src/crypto/generic-ops.h
+++ b/src/crypto/generic-ops.h
@@ -33,6 +33,7 @@
 #include <cstddef>
 #include <cstring>
 #include <functional>
+#include <memory>
 #include <sodium/crypto_verify_32.h>
 
 #define CRYPTO_MAKE_COMPARABLE(type) \
@@ -60,14 +61,18 @@ namespace crypto { \
 namespace crypto { \
   static_assert(sizeof(std::size_t) <= sizeof(type), "Size of " #type " must be at least that of size_t"); \
   inline std::size_t hash_value(const type &_v) { \
-    return reinterpret_cast<const std::size_t &>(_v); \
+    std::size_t h; \
+    memcpy(&h, std::addressof(_v), sizeof(h)); \
+    return h; \
   } \
 } \
 namespace std { \
   template<> \
   struct hash<crypto::type> { \
     std::size_t operator()(const crypto::type &_v) const { \
-      return reinterpret_cast<const std::size_t &>(_v); \
+      std::size_t h; \
+      memcpy(&h, std::addressof(_v), sizeof(h)); \
+      return h; \
     } \
   }; \
 }


0xFFFC0000 (commented):
Question, what is your estimation about performance impact of this? My understanding this is introducing a copy in a critical location and can have huge impact.

jeffro256:
The copy was already there implicitly, but now it's a string copy instead of a 8-byte load. I don't see this slowing things down that much, since usually the data structure operations which use the hash value are going to be way more expensive than an 8-byte load AFAIK.

src/crypto/generic-ops.h
```diff
@@ -60,14 +60,18 @@ namespace crypto { \
 namespace crypto { \
   static_assert(sizeof(std::size_t) <= sizeof(type), "Size of " #type " must be at least that of size_t"); \
   inline std::size_t hash_value(const type &_v) { \
-    return reinterpret_cast<const std::size_t &>(_v); \
+    std::size_t h; \
+    memcpy(&h, &_v, sizeof(h)); \
```
vtnerd, line 64: This should arguably be updated to `std::addressof(_v)`. I still approved this PR anyway, as it is unlikely to be a problem because we rarely overload `operator&`.

vtnerd (approved):
As per the other comments - the compiler will replace `memcpy` with simple loads/stores where possible. In this situation, it will primarily depend on the source (`_v`) type, as `size_t` can easily be the target of a store operation.

src/crypto/generic-ops.h
jeffro256: Done

vtnerd (approved):

iamamyth:
To the point @0xFFFC0000 made, it might not be a bad idea to mark the structs themselves as aligned to a 16-byte boundary (to permit vector instructions), though I suspect most compilers already align them because the default libraries vectorize a lot of stuff and unaligned vector loads on x86 signal rather than stall the pipeline (unless explicitly permitted by the instruction used). As an example of where alignment would be relevant, crypto_verify_32 (used by the constant-time hash comparator) uses _mm_loadu_si128, which runs fine, but slow, for unaligned data.

iamamyth:
Should clarify: The preceding comment would be future work, not within scope for this PR.

iamamyth:
I'd consider this suggestion out of scope for this PR, but, to the point @0xFFFC0000 made, it might not be a bad idea to mark the structs themselves as aligned to a 16-byte boundary (to permit vector instructions), though I suspect most compilers already align them because the default libraries vectorize a lot of stuff and unaligned vector loads on x86 signal rather than stall the pipeline (unless explicitly permitted by the instruction used). As an example of where alignment would be relevant, crypto_verify_32 (used by the constant-time hash comparator) uses _mm_loadu_si128, which runs fine, but slow, for unaligned data, yet has no load penalty for aligned data. We'd probably need to confirm compilers do as expected when mixing packed structs and `alignas`.

0xFFFC0000:
I personally like to investigate the assembly code to maneuver sure there is no overhead.

vtnerd:
> I personally like to investigate the assembly code to maneuver sure there is no overhead.

I've done this with gcc-14 on godbolt. It does convert things into a single `mov` instruction where possible. But it should depend on the source type. Regardless, the current code violates the aliasing rule, so its a good idea to change it.

vtnerd:
As another update, I've checked with a `struct` similar to what happens in the current code. It's a single `mov` instruction. As @iamamyth mentioned, x86 supports unaligned moves (at a performance penalty), so perhaps we could mark some things for alignment purposes. Regardless, the code generated is going to be similar in many cases (where the current code is an unaligned move anyway).

iamamyth:
> As @iamamyth mentioned, x86 supports unaligned moves (at a performance penalty), so perhaps we could mark some things for alignment purposes.

We may want to consider using aligned allocators for the vectors of 32-byte values: Otherwise, the allocator might choose a poor alignment, and really slow down some operations. In a non-array context, the packing directives don't limit the compiler, so it could fix these problems automatically (e.g. if it sees a 128-bit intrinsic being applied, it could suitably align the data).

iamamyth:
> As @iamamyth mentioned, x86 supports unaligned moves (at a performance penalty), so perhaps we could mark some things for alignment purposes.

We may want to consider using aligned allocators for the vectors of 32-byte values: Otherwise, the allocator might choose a poor alignment, and really slow down some operations. In a non-array context, the packing directives don't limit the compiler, so it could fix these problems automatically because the packed structure members tend to be aligned relative to the starting offset (e.g. a single 32-byte value, or two contiguous 32-byte values). If the compiler does a bad job with scalar values in certain hot loops, those declarations could be adjusted, but that's a tough road to travel.

iamamyth:
The suggestion of fixing compiler behavior locally for hot loops also appears in an Intel vectorization manual (https://d3f8ykwhia686p.cloudfront.net/1live/intel/CompilerAutovectorizationGuide.pdf, p. 14). Using packed structures to make memory-based comparisons safe and ensure a consistent size effectively prevents type-level alignment (using `alignas`) because new alignment requirements may introduce padding in the packed structures.

jeffro256:
Also out of scope for this PR, but there is something to be said about this function not being secure against DoS attacks on the hash value. It's only safe if values inputted into this function are the result of a one way function. But if an adversary gets to pick, for example, a `crypto::public_key` value, they can pick a zillion `crypto::public_key` values which have the first 8 bytes the same as each other, for free. Transaction output pubkeys are an example of values which are picked by anybody without validation. See here for a related, not quite the same, issue: https://cplusplus.github.io/LWG/issue2291. Our issue is that collisions can be found in 0 operations instead of 2<sup>32</sup>. I'd have to comb over the actual applications we use the `std::hash` overload for in this repo to see if it could be an actual issue.

iamamyth:
It seems the overall design of these objects could use a rethink: They introduce a lot of assumptions, which in turn become implementation requirements that either unnecessarily constrain the compiler or create cognitive overhead for developers. For example, the memory layout has multiple motivations (not even mentioned in the code), which could perhaps be better served by a separation of concerns (consistent, compact, backwards-compatible wire format doesn't necessarily imply the same design as constant-time comparison, wipeability, etc).

PR #2634, opened 2017-10-11
wallet: add sweep_single command

stoffu:
This patch adds a new command `sweep_single <key_image> <address>` which sweeps exactly one output specified by the given key image to the destination address with no change. This may be useful for a few purposes:

- Cold signing: After exporting the set of correct key images from the offline full wallet to the online watch-only wallet, this command can be used to spend each one of output in the cold wallet without generating a new change output, which simplifies the process of moving funds from a cold wallet to a hot wallet by making it unnecessary to export/import outputs/key-images every time moving funds.

- Manual churning

- Defense against potential privacy leak when spending many outputs from the same entity (e.g. an exchange)

```
[wallet A1PbM6]: incoming_transfers available verbose
               amount   spent    unlocked  ringct    global index                                                               tx id                                                              pubkey                                                           key image
       1.750732960000       F    unlocked  RingCT          289557  <1a49403c7c718d517edd7856060fce96d01d602242f67bb3081e9ed73ad20243>  <0ff2f2309ff0bd58150ded6b844edb195efe5a9ebbb1701ec6df4433b958efc3>    f2e713dc54a9b1ec9c4925793ee1f35636234c8b99b7dd168791677e51e704df
[wallet A1PbM6]: sweep_single unimportant 5 f2e713dc54a9b1ec9c4925793ee1f35636234c8b99b7dd168791677e51e704df 9vfmz4q7LGQYMisHnLZzUCW1kjG35394sK14D3EfRK9xabR2VouaNSDYdTE5EeSvTe5WV2e6tFUFrEbC8gaHk85oCM4uz8g

Transaction 1/1: txid=<bf29fb9516072b8674f6edf83ea29346248ac8959bf08a1104c809b95731b304>
Input 1/1: amount=1.750732960000
Originating block heights:  965548 *1012065 1014813 1015188 1015378
|___________________________________________________________________________o__*|

Sweeping 1.750732960000 for a total fee of 0.004783740000.  Is this okay?  (Y/Yes/N/No)
```

PR #2586, opened 2017-10-06
wallet-cli: add --do-not-relay option

stoffu:
This functionality is similar to the `do_not_relay` option in the wallet RPC, which I think would be useful for pushing tx anonymously to a remote node on Tor or I2P.

Like the wallet RPC, I could have made it as a per-transfer option by changing the command syntax of  `transfer` (e.g. introduce additional keyword `do_not_relay` at the end), but I thought that'd be cumbersome.

PR #9698, opened 2025-01-13
add do_not_relay option to sweep_single ...

SNeedlewoods:
... and submit_multisig_main methods in simplewallet.

It seems when the PR that added `sweep_single()` ([see here](https://github.com/monero-project/monero/pull/2586)) was made, the PR for `commit_or_save()` ([see here](https://github.com/monero-project/monero/pull/2634)) wasn't done yet and that's why sweep_single still calls `m_wallet->commit_tx()` ([src](https://github.com/monero-project/monero/blob/2e8a128c752a3cee2a0bee43b3c15ae7ec344792/src/simplewallet/simplewallet.cpp#L7465)) which does not have the `do_not_relay` option, which was added on the second PR mentioned.
For the `submit_multisig_main()` I have to admit I didn't investigate and don't have an argument, besides I assume that's how it should be done.


nahuhh:
@SNeedlewoods this was fixed for sweep_single in a duplicate pr

can you rebase & adjust this pr (commit and title) to only address the multisig

selsta (approved):

PR #9707, opened 2025-01-15
ci: remove ubuntu 22.04 job

tobtoht:
Assumes #9680.

The Ubuntu 22.04 job is a little redundant now that we have a job for a bleeding-edge Linux distro (Arch Linux). Builds failing due to package upgrades will most likely get caught by this run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,18 +104,10 @@ jobs:
 # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
 
   build-ubuntu:
-    name: ${{ matrix.name }}
-    runs-on: ${{ matrix.os }}
+    name: 'Ubuntu 20.04'
+    runs-on: ubuntu-20.04
     env:
       CCACHE_TEMPDIR: /tmp/.ccache-temp
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - os: ubuntu-20.04
-            name: Ubuntu 20.04
-          - os: ubuntu-22.04
-            name: Ubuntu 22.04
     steps:
     - uses: actions/checkout@v4
       with:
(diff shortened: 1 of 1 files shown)


iamamyth:
Couldn't you make a similar argument for debian 10 and ubuntu 20? If anything, I'd rather keep ubuntu 22 than 20, as it's quite a popular distro (in effect making the test matrix: upcoming, common/popular base, and oldest).

tobtoht:
The jobs for Debian 10 and Ubuntu 20.04 serve a specific purpose: to ensure we don't accidentally break development builds on the oldest supported distributions. Due to how EOL policies line up, Debian does not always have the oldest packages, so it's important to keep both.

The most recent packages are covered by the bleeding edge job. Build issues with recent packages can be fixed before they land in more recent Ubuntu versions.  

I don't know where 22.04 fits in here. If we want to keep it, we should define a deprecation policy, so it's clear to (future) maintainers when it should be removed/upgraded.

iamamyth:
> Due to how EOL policies line up, Debian does not always have the oldest packages, so it's important to keep both.

I had forgotten Ubuntu sometimes ends up with older packages than the roughly equivalent Debian version.

> I don't know where 22.04 fits in here

Ubuntu 22.04 was the "common/popular" part of the test matrix from my earlier comment; maybe that should now be 24, but it seems a bit too recent to have more users than 22. I can see a case for eliminating it to save build time, but theoretical backwards-compat doesn't always translate to reality (compilers especially like to break things). As for update timelines, particularly if there's an Arch build, I'd guess 1-2 years after a new LTS distro gets released, rounding up to two; perhaps we can find some data to better inform the decision, if we want to keep it.

iamamyth:
Here are some initial stats on Ubuntu usage:
https://fr.archive.ubuntu.com/stats/stats_of_day-24.html?version=last
https://fr.archive.ubuntu.com/stats/stats_of_day-406.html?version=last

So, ~1.6 years after 22 was released, it easily eclipsed 20, whereas 24 had virtually no usage 7 months after release. I'd say the next LTS becomes the "most popular" after 18 months maybe?

selsta:
I think it does not hurt to keep it if it adds a different compiler version compared to the other CI jobs.

tobtoht:
I think 'most recent LTS + 18 months' is a reasonable deprecation guideline. We can evaluate at that time if a newer version has indeed eclipsed the older version in usage. This will also give us a decent spread of compiler versions.

| OS           | Purpose | Deprecation Policy            |           Date |    GCC |
|--------------|---------|-------------------------|---------------:|-------:|
| Debian 10    | Oldest  | 9446 or later           |  > 30 Jun 2025 |  8.3.0 |
| Ubuntu 20.08 | Oldest  | 9446 or later           |  > 02 Apr 2026 |  9.4.0 |
| Ubuntu 22.04 | Popular | Most recent LTS + ~18 m |     ~ Nov 2025 | 11.4.0 |
| Arch Linux   | Newest  | Never                   |              - |   (14) |

I'll likely follow up with a PR to document this.

PR #9708, opened 2025-01-15
ci: containerize ubuntu cli jobs

tobtoht:
GitHub will begin deprecation of the Ubuntu 20.04 runner in [Feb 2025](https://github.com/actions/runner-images/issues/11101). We support Ubuntu 20.04 until at least Apr 2026.

Proactively containerize both Ubuntu 20.04 and 22.04 to avoid future disruptions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,17 +13,18 @@ on:
 # The below variables reduce repetitions across similar targets
 env:
   REMOVE_BUNDLED_PACKAGES : sudo rm -rf /usr/local
+  # ARCH="default" (not "native") ensures, that a different execution host can execute binaries compiled elsewhere.
   BUILD_DEFAULT_LINUX: 'cmake -S . -B build -D ARCH="default" -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=Release && cmake --build build --target all && cmake --build build --target wallet_api'
   APT_INSTALL_LINUX: 'apt -y install build-essential cmake libboost-all-dev miniupnpc libunbound-dev graphviz doxygen libunwind8-dev pkg-config libssl-dev libzmq3-dev libsodium-dev libhidapi-dev libusb-1.0-0-dev libprotobuf-dev protobuf-compiler ccache git'
   APT_SET_CONF: |
-        tee -a /etc/apt/apt.conf.d/80-custom << EOF
-        Acquire::Retries "3";
-        Acquire::http::Timeout "120";
-        Acquire::ftp::Timeout "120";
-        EOF
+    tee -a /etc/apt/apt.conf.d/80-custom << EOF
+    Acquire::Retries "3";
+    Acquire::http::Timeout "120";
+    Acquire::ftp::Timeout "120";
+    EOF
   CCACHE_SETTINGS: |
-        ccache --max-size=150M
-        ccache --set-config=compression=true
+    ccache --max-size=150M
+    ccache --set-config=compression=true
   USE_DEVICE_TREZOR_MANDATORY: ON
 
 jobs:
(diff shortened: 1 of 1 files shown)


iamamyth:
Running 20.04 in a container makes sense. Regarding 22.04, how much slower does this make the build? It seems you could just move it when needed, the change would not be terribly complex, and will likely coincide with the promotion of 24 to "popular" and 20 to "no longer supported".

tobtoht:
Containerization also has the added benefit of accurately representing the target development environment, so we don't run into a repeat of #9128. Removing the bundled packages can take as much as [10 minutes](https://github.com/monero-project/monero/actions/runs/12771918518/job/35600229979).

iamamyth:
You can easily, and quickly, drop the contents of `/usr/include` with `mv` (as root):
```
mv /usr/include /usr/extinclude && mkdir /usr/include && chown --reference=/usr/extinclude /usr/include && chmod --reference=/usr/extinclude /usr/include
```

.github/workflows/build.yml
```diff
@@ -100,43 +101,47 @@ jobs:
...
+          path: ~/.ccache
+          key: ccache-${{ matrix.container }}-build-${{ github.sha }}
+          restore-keys: ccache-${{ matrix.container }}-build-
+      - name: build
+        run: |
+          ${{env.CCACHE_SETTINGS}}
+          ${{env.BUILD_DEFAULT_LINUX}}
 
   libwallet-ubuntu:
     name: "Ubuntu 20.04 (libwallet)"
```
iamamyth, line 147: Why does this action not also run in the container?

.github/workflows/build.yml
```diff
@@ -167,59 +172,81 @@ jobs:
...
+      - name: tests
+        env:
+          CTEST_OUTPUT_ON_FAILURE: ON
+          DNS_PUBLIC: tcp://9.9.9.9
+        run: |
+          ${{env.CCACHE_SETTINGS}}
+          ${{env.BUILD_DEFAULT_LINUX}}
+          cmake --build build --target test
 
 # ARCH="default" (not "native") ensures, that a different execution host can execute binaries compiled elsewhere.
```
iamamyth, line 219: Not related to this PR, but it seems this comment very much lost its anchor at some point (I believe it was intended to apply to BUILD_DEFAULT_LINUX; the part about BUILD_SHARED_LIBS no longer appears anywhere in the file). Do you want to open a separate PR at some point to correct the drift?

iamamyth:
Some clarification of my feedback thusfar may be in order; I meant to signify approval of this line:

> Containerization also has the added benefit of accurately representing the target development environment

As far as performance, the relevant measure would be the total addition to build time, not just the setup phase. Because I consider faithful representation of the target environment most important, and "just drop /usr/include" seems a bit finicky, I would favor the container approach even if it adds, say, a minute of overhead; past that, and the tradeoff becomes less clear in my view.

.github/workflows/build.yml
tobtoht: It is scheduled for removal in #9706, I forgot to mention the soft dependency. Leaving it like this avoids a conflict.

.github/workflows/build.yml
tobtoht: Thanks, I made a note of this.

iamamyth (approved):
The changes here seem reasonable. The test failures owe to a flawed test methodology (fixed in https://github.com/monero-project/monero/pull/9709, though I may post additional improvements to that test suite after yours have been merged).

.github/workflows/build.yml
```diff
@@ -187,62 +194,79 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - uses: actions/cache@v4
+        with:
+          path: ~/.ccache
+          key: ccache-${{ matrix.container }}-build-${{ github.sha }}
+          restore-keys: ccache-${{ matrix.container }}-build-
+      - uses: ./.github/actions/set-make-job-count
+      - name: tests
```
iamamyth, line 233: It looks like you accidentally dropped the test drive configuration step here (test output confirms).

iamamyth (approved):

.github/workflows/build.yml
```diff
@@ -13,17 +13,18 @@ on:
 # The below variables reduce repetitions across similar targets
 env:
   REMOVE_BUNDLED_PACKAGES : sudo rm -rf /usr/local
-  BUILD_DEFAULT_LINUX: 'cmake -S . -B build -D ARCH="default" -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=Release && cmake --build build --target all && cmake --build build --target wallet_api'
+  # ARCH="default" (not "native") ensures, that a different execution host can execute binaries compiled elsewhere.
+  BUILD_DEFAULT_LINUX: 'cmake -S . -B build -D ARCH="default" -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=Release && cmake --build build'
```
iamamyth, line 17: Did you mean to remove the two targets here?

tobtoht: Oops, fixed, thanks.

.github/workflows/build.yml
```diff
@@ -123,106 +125,129 @@ jobs:
...
-        ${{env.BUILD_DEFAULT_LINUX}}
-        cmake --build build --target test
-
-
-# ARCH="default" (not "native") ensures, that a different execution host can execute binaries compiled elsewhere.
-# BUILD_SHARED_LIBS=ON speeds up the linkage part a bit, reduces size, and is the only place where the dynamic linkage is tested.
+        DEBIAN_FRONTEND: noninteractive
+        CCACHE_TEMPDIR: /tmp/.ccache-temp
+        CCACHE_DIR: ~/.ccache
+      options: --privileged
```
iamamyth, line 188: Why does this require `--privileged` but the earlier container builds don't? A clarifying comment might help future readers.

tobtoht: Added a comment.

iamamyth (approved):

PR #9712, opened 2025-01-16
tests: remove is_hdd

tobtoht:
Does not work inside containers. Has other issues: see discussion.

Required for #9708.


src/common/util.cpp
```diff
@@ -588,6 +588,14 @@ namespace tools
     {
       return boost::none;
     }
+
+    if (!boost::filesystem::exists(prefix)) {
+      // If the device returned by stat does not exist at /sys/dev/block/, we are likely
+      // inside a container. There is no way to check if file_path is on a hard drive.
+      MWARNING("Unable to determine whether file path '" << file_path << "' is on a HDD, assuming not");
+      return false;
```
iamamyth, line 596: Shouldn't this return `boost::none`? Otherwise I can't see a coherent description of the overall function's behavior.

tobtoht: It returns false to signify that it is not an error (and warns the user instead). Returning `boost::none` would cause [this](https://github.com/monero-project/monero/blob/master/tests/unit_tests/is_hdd.cpp#L9) test case to fail.

iamamyth: Writing code to make the tests pass, rather than produce coherent function behavior, doesn't make sense. If "cannot determine whether it's an HDD or not" isn't an "unknown", what is it? Please write a coherent description of this function's behavior with your change ("returns false if not on an HDD or can't tell if on an HDD" doesn't work, because "can't tell if on an HDD" applies to *all* of the none cases prior to this change).

iamamyth: > I don't know how to compress 4 possible states: {we know it's an HDD | we know it's not an HDD | we know that we don't know if it's an HDD | something unexpected happened} into 3 values, other than what I've come up with here.

This is precisely the misapprehension, in my view: There aren't four states, only three. Nearly all of the existing "none" cases occur when a file should be there and openable, but isn't. In those cases, "we know that we don't know" and we also know that "something unexpected happened"; they are the same thing. The same conditions apply in this new case because the block device can't be queried, at all; prior to this change, that would have meant returning `none`, implying you now have a backwards-incompatible behavior, yet no coherent description of the new behavior.

If you want to do what the PR title says, then do it: Avoid showing a warning in the unknown case. Or change the message. If you wanted to alter the return type of the function to accommodate a new state, you certainly could, but that would require an actual description of the state, which still does not exist.

tobtoht: >If you want to do what the PR title says, then do it: Avoid showing a warning in the unknown case.

Then I don't know what would justify returning false there. We might as well return true if it's an hdd, and false in all other cases. That would obsolete the tests.

iamamyth: I really don't understand the behavior's goals. The existing log message triggers when we know the user is on an HDD, as that's a known source of problems. Why emit a "you might be on an HDD" warning in only some of the cases where the user might be an an HDD? If you want to warn there, do it in all the cases (where the function returns `none`), but I would say, just don't do it.

tobtoht: Which do you prefer: remove the warning and return false or rewrite the function to return a bool and remove the tests?

iamamyth: > Which do you prefer: remove the warning and return false or rewrite the function to return a bool and remove the tests?

Neither. Make no changes, as I see no good reason for them.

iamamyth: Is your goal here just fixing something in the test suite? If so, have the test check for either none or false and be done with it. Realistically, the test has serious logical problems, and needs a totally different approach.

iamamyth: The logic of the test is broken: On "what appears to be like linux", this shouldn't return `none`. Well, except, it obviously can, and sometimes should. If you want to keep the no `__GLIBC__` test, I guess you can, but it's also rather contrived.

iamamyth (approved):
This seems like the right choice. I might consider introducing a better smoke test for the function at a later date.

PR #9716, opened 2025-01-17
tests: Fix tools::is_hdd unit tests

iamamyth:
Correct the unit tests for tools::is_hdd to avoid making assumptions about the configuration of a particular device based solely on the value of the __GLIBC__ preprocessor flag. Instead, rely on the test invoker to provide paths for devices of specific types via the process environment, thereby avoiding faulty assumptions and improving the specificity of test assertions. To ensure appropriate devices exist, add a script, tests/create_test_disks.sh, which configures loopback devices mirroring relevant configurations.

### DIFF
--- a/tests/create_test_disks.sh
+++ b/tests/create_test_disks.sh
@@ -1,0 +1,98 @@
+#!/bin/bash
+
+set -e
+
+LOOP_DEVICE_MAJOR=7
+LOOP_DEVICE_MIN_ID=100
+
+echo_err() {
+    echo "$@" >&2
+}
+
+root_exec() {
+    if [[ $EUID -ne 0 ]]; then
+        ${ROOT_EXEC_CMD:-sudo} "$@"
+    else
+        "$@"
+    fi
+}
+
+create_device_node() {
+    local -r last_id=$(find /dev/ -name 'loop[0-9]*' -printf '%f\n' | sed 's/^loop//' | sort -r -n | head -1)
+    local id=$((last_id + 1))
+    if [[ "$id" -lt "$LOOP_DEVICE_MIN_ID" ]]; then
+        id="$LOOP_DEVICE_MIN_ID"
+    fi
+    local path
+    for (( i=0; i<10; i++ )); do
+        path="/dev/loop$id"
+        if [[ ! -e "$path" ]] && root_exec mknod "$path" b "$LOOP_DEVICE_MAJOR" "$id"; then
+            echo "$path"
+            return 0
+        fi
+        $((id++))
+    done
+    return 1
+}
+
+device_mountpoint() {
+    local -r datadir="$1"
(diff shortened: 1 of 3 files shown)


iamamyth:
Obsoletes https://github.com/monero-project/monero/pull/9712.

tests/unit_tests/is_hdd.cpp
```diff
@@ -1,17 +1,36 @@
 #include "common/util.h"
+#include <cstdlib>
 #include <string>
 #include <gtest/gtest.h>
+#include <boost/optional/optional_io.hpp> /* required to output boost::optional in assertions */
+
+#ifndef GTEST_SKIP
+#include <iostream>
+#define SKIP_TEST(reason) do {std::cerr << "Skipping test: " << reason << std::endl; return;} while(0)
```
iamamyth, line 9: I introduced this macro because `GTEST_SKIP` doesn't exist in fairly old versions of gtest (e.g. the vendored one, as well as the default installs on Debian 10 and Ubuntu 20), but should be available on every build in the near future, at which point I'll drop the macro.

iamamyth: Link to the PR which introduced GTEST_SKIP: https://github.com/google/googletest/pull/1544

iamamyth:
Test failure is one of the known flaky tests, unrelated to this commit.

0xFFFC0000 (approved):
LGTM. Thanks.

tests/create_test_disks.sh
```diff
@@ -0,0 +1,90 @@
...
+        echo_err "Usage: create_device <data dir>"
+        return 1
+    fi
+    local -r dev=$(create_device_node)
+    local -r fs="$datadir/$(basename "$dev").vhd"
+    local -r mountpoint=$(device_mountpoint "$datadir" "$dev")
+    echo_err
+    echo_err "# Device $dev"
+    dd if=/dev/zero of="$fs" bs=64K count=128 >/dev/null 2>&1
+    sudo losetup "$dev" "$fs"
```
tobtoht, line 52: some environments might not have `sudo` installed or prefer another tool to elevate privileges (e.g. `doas`).

I think it would be better to remove the use of `sudo` in the script and add something at the beginning like:

```
if [[ $EUID -ne 0 ]]; then
   echo "This script must be run as root." >&2
   exit 1
fi
```

iamamyth: Requiring root for the whole script obfuscates which operations require special privileges and makes the functions within the script less helpful (the current structure permits defining these functions in an interactive shell and then calling them as desired; wrapping the invocation in sudo or doas loses the shell customizations, including the function definitions). I will just make sudo optional.

tobtoht (approved):

PR #9717, opened 2025-01-18
ringct: remove unused range proof types and fix serialization bug

jeffro256:
Remove unused Rangeproof types `RangeProofBulletproof `, and `RangeProofMultiOutputBulletproof`. In the prove code, if the `RangeProofBulletproof` enum was passed, then one bulletproof was created per output. If `RangeProofMultiOutputBulletproof` was passed, then K bulletproofs were produced, where K is the number of one bits in the count of the outputs, Neither of these are allowed by consensus: https://github.com/monero-project/monero/blob/2226d4f7e0bf316aa0d4241b9b498f928add4b8b/src/cryptonote_basic/cryptonote_format_utils.cpp#L150-L154 and https://github.com/monero-project/monero/blob/2226d4f7e0bf316aa0d4241b9b498f928add4b8b/src/cryptonote_basic/cryptonote_format_utils.cpp#L174-L178.

Also fixes an actual bug with the boost serialization of type `tools::wallet2::signed_tx_set`. If the serialized payload has a version < 4, then the range proof type is assumed to be `RangeProofBulletproof`, which is incorrect.

### DIFF
--- a/src/device_trezor/trezor/protocol.cpp
+++ b/src/device_trezor/trezor/protocol.cpp
@@ -413,8 +413,6 @@ namespace tx {
   static unsigned get_rsig_type(const rct::RCTConfig &rct_config, size_t num_outputs){
     if (rct_config.range_proof_type == rct::RangeProofBorromean){
       return rct::RangeProofBorromean;
-    } else if (num_outputs > BULLETPROOF_MAX_OUTPUTS){
-      return rct::RangeProofMultiOutputBulletproof;
     } else {
       return rct::RangeProofPaddedBulletproof;
     }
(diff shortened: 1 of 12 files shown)


selsta:
Compilation issues with some CI jobs

jeffro256:
Oops sorry I only pushed some of the commits

iamamyth:
Looks like there's still one failing unit test:
`select_outputs.exact_unlock_block`

jeffro256:
That test should be unrelated, and fails sporadically.

iamamyth:
Agreed, unrelated.

iamamyth (approved):

PR #9718, opened 2025-01-18
wallet: fix signed_tx_set RangeProofType serialization [RELEASE]

jeffro256:
Subset of #9717


selsta:
@iamamyth can you also add an approval here?

iamamyth (approved):
Confirmed that this is a (rather small) subset of the linked PR, which I already approved.

PR #7737, opened 2021-05-27
extend payment uri scheme to support multiple recipients per tx

benevanoff:
Addresses #7731 for https://github.com/haveno-dex/haveno/issues/83

Changes to make_uri: now accepts an array of objects holding the address, amount, and recipient name. You can have as many of these as you want. only one payment id and one tx_description are allowed per transaction

Changes to parse_uri: parses the new scheme. Important: a valid uri that uses amounts or recipient names must have one for each address.

example usage:

For reference I started up my rpc wallet as such: `./monero-wallet-rpc --wallet-file new --rpc-bind-ip 127.0.0.1 --rpc-bind-port 18082 --rpc-login test:test --prompt-for-password`

request to make uri

` curl -u test:test --digest -X POST http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"make_uri","params":{"payments":[{"address":"41tQjL6UoNxEsxKf1DEkMuTBVUXMAxP74TSRR9WTrBX39Z9t4BUt1XmaQdUPZG2EnsjCEYaqUSBhYEweSTLHMgTDHtrgRxC","amount":10,"recipient_name":"me"},{"address":"41tQjL6UoNxEsxKf1DEkMuTBVUXMAxP74TSRR9WTrBX39Z9t4BUt1XmaQdUPZG2EnsjCEYaqUSBhYEweSTLHMgTDHtrgRxC","amount":1560,"recipient_name":"him"},{"address":"41tQjL6UoNxEsxKf1DEkMuTBVUXMAxP74TSRR9WTrBX39Z9t4BUt1XmaQdUPZG2EnsjCEYaqUSBhYEweSTLHMgTDHtrgRxC","amount":2000,"recipient_name":"you"}],"tx_description":"abc","payment_id":"420fa29b2d9a49f54201119b2d9a49f5420fa29b2d9a49f5420fa29b2d9a49f5"}' -H 'Content-Type: application/json' `

response

`{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
[...]

PR #9719, opened 2025-01-22
Multi uri

sa8ab:
This PR fixes the issues existing in #7737 
- Loop index fix ( `j` instead of `i` )
- No NA filler
- Cleanup


selsta:
This fails to compile.

sa8ab:
I will have to create a new PR. The code was not compatible with master.

PR #9720, opened 2025-01-23
functional tests: fix race in restore wallet refresh

j-berman:
Needed to disable auto refresh *before* restoring the wallet, so [refresh doesn't run in the idle loop](https://github.com/monero-project/monero/blob/eaf9f98930fc72f1b468e980b220d53cadc85443/src/wallet/wallet_rpc_server.cpp#L198) upon restoring.

Fixes [this test failure](https://github.com/monero-project/monero/pull/9389#issuecomment-2608902212) highlighted by @tobtoht (apologies for missing it).

### DIFF
--- a/tests/functional_tests/transfer.py
+++ b/tests/functional_tests/transfer.py
@@ -57,6 +57,15 @@ def diff_incoming_transfers(actual_transfers, expected_transfers):
     # wallet2 m_transfers container is ordered and order should be the same across rescans
     diff_transfers(actual_transfers, expected_transfers, ignore_order = False)
 
+def restore_wallet(wallet, seed, restore_height = 0, filename = '', password = ''):
+    try: wallet.close_wallet()
+    except: pass
+    if filename != '':
+        util_resources.remove_wallet_files(filename)
+    wallet.auto_refresh(enable = False)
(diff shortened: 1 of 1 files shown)


selsta:
Just to confirm, this was purely a test issue and this case can not happen inside a real wallet?

iamamyth (approved):

tests/functional_tests/transfer.py
```diff
@@ -57,6 +57,15 @@ def diff_incoming_transfers(actual_transfers, expected_transfers):
     # wallet2 m_transfers container is ordered and order should be the same across rescans
     diff_transfers(actual_transfers, expected_transfers, ignore_order = False)
 
+def restore_wallet(wallet, seed, restore_height = 0, filename = '', password = '', ):
```
iamamyth, line 60: The trailing `, ` after the last param, though allowed, seems to break with the style elsewhere in this file, as well as python convention.

iamamyth:
The style comment doesn't impact approval (I saw the issue prior to signing off), as every part of the code style in these tests violates most linter rule sets, so one extra line doesn't matter.

tests/functional_tests/transfer.py
j-berman: Fixed

PR #9723, opened 2025-01-25
ci: Parallelize Arch Linux build

iamamyth:
The "--parallel" flag got dropped from the Arch Linux build due to unidentified, logical commit conflicts.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,8 +92,9 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
+      - uses: ./.github/actions/set-make-job-count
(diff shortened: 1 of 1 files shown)


selsta (approved):

.github/workflows/build.yml
```diff
@@ -93,7 +93,7 @@ jobs:
         with:
           submodules: recursive
```
tobtoht, line 94: I think you forgot `- uses: ./.github/actions/set-make-job-count` here

iamamyth: Fixed and squashed.

tobtoht (approved):

PR #9727, opened 2025-01-26
cmake: remove unused lto options

tobtoht:
Unused. We won't be able to use LTO to build releases for the foreseeable future.


selsta (approved):

0xFFFC0000 (approved):

jeffro256:
This isn't unused for master builds, thus I wouldn't want it removed completely

tobtoht:
Where and why exactly is this used for master builds? LTO doesn't really make sense outside of statically linked release builds.

jeffro256:
If I want to run the latest commits on master, I can compile with master branch. Since I'm not doing a reproducible build, I can enable LTO for better performance. So LTO in this case is useful to people who prefer speed and new features over stable release builds.


Feb 14, 2025, 00:29 by ***@***.***:

PR #9729, opened 2025-01-26
cmake: remove msvc

tobtoht:
We don't support MSVC, so let's not pretend like we do.


0xFFFC0000 (approved):
I am approving this, since our MSVC support is not working not robust.

 But eventually we might want to implement it because of debugging on windows. 

Even if we want to add msvc support, these code must be cleaned up and removed.

Appreciate it.

iamamyth (approved):
I agree with the PR description: Vaguely attempted, but failed, MSVC support only serves to confuse other maintainers.

selsta (approved):

PR #9740, opened 2025-01-28
blockchain sync: reduce disk writes from 2 to 1 per tx [RELEASE]

jeffro256:
#9135


jeffro256:
The line changes differ significantly from the `master` branch because I deleted the `core_proxy` tests in this commit to match master, instead of reworking them.

0xFFFC0000 (approved):

tests/functional_tests/p2p.py
```diff
@@ -158,6 +159,48 @@ def test_p2p_reorg(self):
             loops -= 1
             assert loops >= 0
 
+        # create a transaction which both daemons have in their pool
```
iamamyth, line 162: This file has changed quite a bit on the release branch, probably a good idea to rebase and then adjust any changes you want to keep.

tests/functional_tests/p2p.py
```diff
@@ -183,6 +226,42 @@ def test_p2p_tx_propagation(self):
...
             assert res.tx_hashes[0] == txid
 
+        return txid
+
+    def test_p2p_block_propagation(self, daemon2_mempool_txid):
+        print('Testing P2P block propagation')
+        daemon2 = Daemon(idx = 2)
+        daemon3 = Daemon(idx = 3)
+
+        # check precondition: txid in daemon2 mempool
```
iamamyth, line 236: Probably easier to follow if you just fold this test into the `tx_propagation` test at the end, as it's quite intimately related (the first part would test pooled tx propagation, and the second part tx in block propagation). For an example of the overlap, the "precondition" here vanishes, because it's already been checked at the end of the prior function.

tests/functional_tests/p2p.py
```diff
@@ -183,6 +226,42 @@ def test_p2p_tx_propagation(self):
...
+        res = daemon2.get_transaction_pool_hashes()
+        assert daemon2_mempool_txid in res.tx_hashes
+
+        res = daemon2.generateblocks('42ey1afDFnn4886T7196doS9GPMzexD9gXpsZJDwVjeRVdFCSoHnv7KPbBeGpzJBzHRCAs9UxqeoyFQMYbqSWYTfJJQAWDm', 1)
+        block_height = res.height
+
+        # wait until both are synced
+        loops = 5
+        while True:
+            res2 = daemon2.get_info()
```
iamamyth, line 246: There's a better way to do this, see: https://github.com/monero-project/monero/pull/9795/files

SyntheticBird45 (approved):

tests/functional_tests/p2p.py
jeffro256: I disagree, they really are testing different things, but there's just a precondition that overlaps with the previous test case.

iamamyth: If you want it to be wholly separate, then you could make it that way by explicitly adding a transaction to the pool in this test. Having tests depend on state configured by other tests makes failures all the more difficult to comprehend. I won't belabor the point because the functional test suite has so many structural flaws.

iamamyth:
I won't have time for further discussion/re-review, please proceed as you see fit.

iamamyth:
Feature flags create a maintenance burden. If this feature proves problematic, users can simply downgrade to the prior release. The fluffy-blocks flag provides a pretty good example of the maintenance issues, as, years later, as evidenced in the discussion of this PR against the main dev branch, the fluffy-blocks flag doesn't exactly work as it should. I think it's well past the point to eliminate old flags like fluffy blocks, and that's exactly the trouble with feature flags: They create work: Work to add the flag, work to remove the flag, work to update and test code in the presence of the flag. Sometimes, that work can be justified, but it's a high bar.

Gingeropolous:
> users can simply downgrade to the prior release.

indeed, as long as this is done far enough away from a hard fork version, and we catch and fix any potential bugs before the hard fork version solidifies the new code as the only code on the network. 

> Some situations justify that work, but it's a high bar.

indeed. i would assume stuff that relates to consensus checking would be that bar. but what do I know, i can barely read c++. 

lets just create an arbitrary fork of testnet that only the new version can run. ... hrm, that might do it. We can throw this in release, hope its good, let those that wanna upgrade do it. Meanwhile, in the same release, testnet forks, so testnet is entirely running on this code.

jeffro256:
> I was thinking about this mod though. Considering how critical the database is to the whole function of the network, is it worth including a --disable-2to1w flag to allow monerod to revert back to its old behavior?

Piggybacking off of @iamamyth, the maintenance burden of flags here is even greater than something like fluffy blocks, because we aren't *adding* a feature, we're completely rewriting how a certain flow works. I definitely understand and appreciate your skepticism about this large changeset to code in/around consensus checking, but I simply don't think a flag is practical here. This would require we maintain two different live implementations of handling transactions from the `core` code to the `Blockchain` code. If you're skeptical about the large diff (which you have the right to be), the best thing to do for this specific changeset would be to run 2 `monerod`s: one before this PR, and one after.

Gingeropolous:
im starting to appreciate the complexity of the flag idea for this mod, so i'll stop flailing my arms about it. I think a testnet completely on this version would be a good enough situation, as well as performing this update far from a hard fork (so old versions can co-exist).

selsta:
@jeffro256 is it intentional that master and release PR are not in sync?

jeffro256:
@selsta Nope, just haven't updated it yet. Was hoping to get a re-review by @vtnerd of the changes since he approved it last before updating the release PR. Will put it into draft mode until it's updated

selsta (approved):

PR #9464, opened 2024-08-30
add functions to wallet API

SNeedlewoods:
Edit: Should be ready to review now.
___________________________________________

~~Do not merge, this is still in development and just PRd for easier discussion.~~

This is an attempt to make the wallet API "feature complete" and it's part of [this proposal](https://github.com/monero-project/monero/issues/9308).

The approach was to search for every `wallet2` function in `simplewallet` and `wallet_rpc_server`, if it is used in either of those, but is not implemented in the API I considered it missing and it was added in this PR.
I believe `wallet2_api.h` now contains all missing functions, but some are not implemented yet. I'm waiting on feedback on those, if you want to have a look, things I need help with are marked as `QUESTION`, you can also comment on things marked `TODO` (those may become QUESTIONs if I'm unable(/unsure how) to solve them) or anything else.

Also gathered [this collection](https://github.com/SNeedlewoods/seraphis_wallet/issues/1) of (mostly minor) issues found during API work.

PR #9756, opened 2025-01-30
Multiple-URI support in `parse_uri` and `make_uri`

U65535F:
This pull request addresses the issues raised in #7737 and introduces a new feature to the Monero codebase: multi-URL support for `parse_uri` and `make_uri` functions.

## Key Changes:
* Fixed loop index misuse (replaced `i` with `j` for proper iteration).
* Removed the usage of NA as a filler in certain cases.
* General code cleanup for readability and maintainability.
* Introduced new test cases in uri.cpp to validate multi-URL functionality.

Notes:
This PR was rewritten from the latest Monero repository rather than being forked from the original issue (#7737).
All builds are successfully passing, and existing tests have cleared without any regressions.

Please check for any overlooked issues or mistakes, especially in the multi-URL parsing logic.
Validate the new test cases and suggest improvements to ensure robustness.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,7 @@ jobs:
       - name: build
         run: |
           ${{env.CCACHE_SETTINGS}}
-          make release-static-win64 -j${{env.MAKE_JOB_COUNT}}
+          make release-static -j${{env.MAKE_JOB_COUNT}}
 
   build-arch:
     name: 'Arch Linux'
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ cmake-debug:
 	cd $(builddir)/debug && cmake -D CMAKE_BUILD_TYPE=Debug $(topdir)
 
 debug: cmake-debug
-	cd $(builddir)/debug && $(MAKE)
+	cd $(builddir)/debug && cmake --build .
 
 # Temporarily disable some tests:
 #  * libwallet_api_tests fail (Issue #895)
@@ -75,24 +75,12 @@ debug-all:
 	mkdir -p $(builddir)/debug
 	cd $(builddir)/debug && cmake -D BUILD_TESTS=ON -D BUILD_SHARED_LIBS=OFF -D CMAKE_BUILD_TYPE=Debug $(topdir) && $(MAKE)
 
-debug-static-all:
-	mkdir -p $(builddir)/debug
-	cd $(builddir)/debug && cmake -D BUILD_TESTS=ON -D STATIC=ON -D CMAKE_BUILD_TYPE=Debug $(topdir) && $(MAKE)
-
-debug-static-win64:
-	mkdir -p $(builddir)/debug
-	cd $(builddir)/debug && cmake -G "MSYS Makefiles" -D STATIC=ON -D ARCH="x86-64" -D BUILD_64=ON -D CMAKE_BUILD_TYPE=Debug -D BUILD_TAG="win-x64" $(topdir) && $(MAKE)
- 
-debug-static-win32:
-	mkdir -p $(builddir)/debug
-	cd $(builddir)/debug && cmake -G "MSYS Makefiles" -D STATIC=ON -D ARCH="i686" -D BUILD_64=OFF -D CMAKE_BUILD_TYPE=Debug -D BUILD_TAG="win-x32" $(topdir) && $(MAKE)
- 
 cmake-release:
 	mkdir -p $(builddir)/release
 	cd $(builddir)/release && cmake -D CMAKE_BUILD_TYPE=Release $(topdir)
 
 release: cmake-release
-	cd $(builddir)/release && $(MAKE)
+	cd $(builddir)/release && cmake --build .
 
 release-test:
 	mkdir -p $(builddir)/release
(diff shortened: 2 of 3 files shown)


U65535F:
@selsta

selsta:
Thank you. The `URI` functional test case fails now. Can you take a look?

https://github.com/monero-project/monero/actions/runs/13057491252/job/36433416280?pr=9756#step:11:1471

U65535F:
Alright, will ping you when it's done. It seems like I didn't update the python source files to comply with the new updated C++ code.

iamamyth:
As illustrated by the test failures, this PR would introduce a backwards-incompatible change to the wallet RPC interface, which generally should not be done.

iamamyth:
Two possible ways to address the compatibility issue:
1. Introduce a new version of the endpoint which works with only the new format, retaining the old one (you could still transform requests to the old endpoint into the new format upon receipt and share the address construction code).
2. Modify the endpoint to detect "old format" and "new format" requests, and pick the appropriate logic (e.g. payments field present implies new format, or address field present implies old format).

U65535F:
@selsta Please have a look at the code and tell me if to make it ready for review since it's a draft.

U65535F:
@selsta @iamamyth ping

U65535F:
@selsta ping.

selsta:
I will take a closer look within the next days.

U65535F:
> I will take a closer look within the next days.

Alright. I won't ping you till then, but don't forget to reply.

U65535F:
@selsta

src/wallet/wallet2.cpp
```diff
@@ -6236,6 +6236,9 @@ std::string wallet2::make_background_keys_file_name(const std::string &wallet_fi
 //----------------------------------------------------------------------------------------------------
 bool wallet2::parse_long_payment_id(const std::string& payment_id_str, crypto::hash& payment_id)
 {
+  if (payment_id_str.size() != 64)
```
selsta, line 6239: This new check seems redundant and unrelated to this PR. I would remove it again.

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -1825,27 +1825,44 @@ namespace wallet_rpc
...
+    std::string payment_id;
+
+    BEGIN_KV_SERIALIZE_MAP()
+      KV_SERIALIZE(payments);
+      KV_SERIALIZE(tx_description);
+      KV_SERIALIZE(payment_id);
+    END_KV_SERIALIZE_MAP()
+  };
+
   struct COMMAND_RPC_MAKE_URI
```
selsta, line 1854: `WALLET_RPC_VERSION_MINOR` has to be bumped. Generally I'm unsure about this change since it seems incompatible. Wouldn't it make more sense to add a separate `COMMAND_RPC_MAKE_URI` command instead of changing the existing one? I assume that would also reduce the diff in the tests.

tests/functional_tests/uri.py
```diff
@@ -207,23 +217,120 @@ def test_monero_uri(self):
...
+        uri_with_unknown = 'monero:' + addr1 + '?tx_amount=239.39014&foo=bar&baz=quux'
+        parsed = wallet.parse_uri(uri_with_unknown)
+        assert parsed.uri.payments[0].address == addr1
+        assert parsed.uri.payments[0].amount == 239390140000000
+        # unknown parameters should be collected in the unknown_parameters list.
+        assert parsed.unknown_parameters == ['foo=bar', 'baz=quux'], "Unknown parameters mismatch"
 
 if __name__ == '__main__':
-    URITest().run_test()
+    URITest().run_test()
```
selsta, line 336: Please add a newline to all files where this symbol shows up.

selsta (commented):
My main issue currently is the RPC incompatibility. What is your opinion on adding a separate RPC command for it?

tests/functional_tests/uri.py
U65535F: Please elaborate. Which symbol?

src/wallet/wallet2.cpp
U65535F: Yes. There is no requirement for this. I forgot to remove that.

tests/functional_tests/uri.py
selsta: symbol

src/wallet/wallet_rpc_server_commands_defs.h
U65535F: Yes. It would reduce the difference. So the original tests will work as usual. But if there is a need for multi-URI, then the user has to call the separate RPC commands which accept multiple recipient parameters for URI.

U65535F:
@selsta Any more suggestions except the new RPC commands?

src/wallet/wallet_rpc_server_commands_defs.h
selsta: > But if there is a need for multi-URI, then the user has to call the separate RPC commands which accept multiple recipient parameters for URI.

This is preferable over breaking all existing clients in my opinion.

U65535F:
> @selsta As I pointed in previous comments, introducing a new command would completely solve the RPC incompatibility. Thus, I also have to remove the overloads and convert them into a new function. What should I name the new commands? It would then need two new `parse_uri` and `make_uri` specially for the multi-URI.

@selsta How about `make_multi_uri` or `parse_multi_uri` ?
Or we could keep the overloads intact and introduce multiple return structures in RPC?

selsta:
I'm still thinking about what's the best way here, `parse_multi_uri` is a bit clunky because the client might not know if they will get a regular URI or multi URI. In this case they would have to always use `parse_multi_uri`, but that might not be obvious to the end user.

@iamamyth what is your opinion?

iamamyth:
I think this whole implementation has gone in backwards order: There needs to be a clear URI scheme based on a decision regarding compatibility, which then dictates the implementation. I think there is a clear scheme, which aimed to be backwards-compatible by keeping the old fields and using a `;` where appropriate to allow multiple recipients and amounts (e.g. address can now be a single address or a semicolon-separated list of addresses). If you have an extension of the prior URI scheme (i.e. the entire old scheme continues to work, but you just add a bit to the specification regarding multi-valued fields), why would you want or need two parsers? And why would you change the tests for the old parser? Everything about how this should work follows pretty clearly from the URI scheme, which is also why I didn't bother to take a second look, as my initial feedback about compatibility went unaddressed.

iamamyth:
As far as the RPC design goes, I already addressed that question earlier: https://github.com/monero-project/monero/pull/9756#issuecomment-2628409730

iamamyth:
If I had to pick one of the options I suggested for the RPC change, it'd likely be (1): "Introduce a new version of the endpoint which works with only the new format, retaining the old one (you could still transform requests to the old endpoint into the new format upon receipt and share the address construction code)." Nearly all of the endpoint-specific code just marshals and unmarshals data, leaving it to the parse to do the heavy lifting. A new endpoint simplifies the work because the unmarshaling code can force required fields. As for the endpoint path, convention would just be to tack a version on the end, i.e. `parse_uri_v2`, `make_uri_v2`.

U65535F:
> I'm still thinking about what's the best way here, `parse_multi_uri` is a bit clunky because the client might not know if they will get a regular URI or multi URI. In this case they would have to always use `parse_multi_uri`, but that might not be obvious to the end user.
> 
> @iamamyth what is your opinion?

@selsta Obviously. Do you know how is it possible to solve this issue? If `parse_uri` gets called with multiple recipient uri, it will error out.

iamamyth:
If introducing v2 endpoints, the old RPC should fail with a multi-recipient URI; that's a desirable result. For the wallet API, I'm not even sure it makes sense to preserve the old functions, but, if they're preserved, they should just wrap everything as a one-element vector and call-through to the new, vector-based code.

iamamyth:
Regarding wallet API compatibility/changes, please consult with the author of this PR: https://github.com/monero-project/monero/pull/9464.

iamamyth:
You don't really need `_v2` in the wallet function names because the parameter types already distinguish c++ functions (unlike the RPC functions). But, again, please defer to @SNeedlewoods regarding wallet function names.

U65535F:
@selsta @iamamyth I think I have addressed your compatibility issue properly. Introduced v2 versions of `parse_uri` and `make_uri` into the RPC. While the C++ code does not explicitly have those functions, but there are overloads which will select the version based on the parameters (as mentioned by @iamamyth). The tests are restored to their original versions, except for there are new tests for multi-URI functionality. The tests are passing, but there seems be to the p2p test which is failing which is not linked to this PR. This p2p test is also failing the repo, I have not synced the fork.

src/wallet/api/wallet.h
```diff
@@ -214,7 +214,9 @@ class WalletImpl : public Wallet
     virtual void startRefresh() override;
     virtual void pauseRefresh() override;
     virtual bool parse_uri(const std::string &uri, std::string &address, std::string &payment_id, uint64_t &amount, std::string &tx_description, std::string &recipient_name, std::vector<std::string> &unknown_parameters, std::string &error) override;
+    virtual bool parse_uri(const std::string &uri, std::vector<tools::wallet2::uri_data> &data, std::string &payment_id, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error);
```
SNeedlewoods, line 217: I see two issues here:
1. AFAICT we don't want to have "non-standard" types as arguments for methods in the Wallet API, so for simplicity I'd change `std::vector<tools::wallet2::uri_data> &data` to vectors of `std::string`/`std::uint64_t` here, see example below.
2. This belongs to the interface in `src/wallet/api/wallet2_api.h` and there it should look like this (with respect to the proposed change from point 1):
```
virtual bool parse_uri(const std::string &uri, std::vector<std::string> &addresses, std::vector<std::uint64_t> &amounts, std::vector<std::string> &recipient_names, std::string &payment_id, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error) = 0;
```
And here it should look like this (no `virtual` in the beginning but `override` at the end):
```
bool parse_uri(const std::string &uri, std::vector<std::string> &addresses, std::vector<std::uint64_t> &amounts, std::vector<std::string> &recipient_names, std::string &payment_id, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error) override;
```

Same goes for `make_uri()` below.

SNeedlewoods (commented):
Sorry, not sure what exactly I should look at.
I haven't done a review, just skimmed over the changes and noticed some unnecessary edits, mainly in `tests/functional_tests/uri.py`.
Other than that I left a comment regarding the API.

Maybe you can clarify how I can help?

iamamyth:
@SNeedlewoods The reason I suggested pinging you was for your feedback on the wallet API changes, as you've put in some effort to establish a coherent interface. And, it seems your comments address those changes, as hoped. Thanks.

iamamyth:
Please note that I have no interest in formally reviewing/approving this PR, so, if/when it gains the necessary approvals, there's no sense in pinging me.

U65535F:
> Sorry, not sure what exactly I should look at. I haven't done a review, just skimmed over the changes and noticed some unnecessary edits, mainly in `tests/functional_tests/uri.py`. Other than that I left a comment regarding the API.
> 
> Maybe you can clarify how I can help?

@SNeedlewoods What exactly are those unnecessary edits? Could you please list them so I can remove.

src/wallet/api/wallet2_api.h
```diff
@@ -1066,8 +1066,10 @@ struct Wallet
     virtual bool verifyMessageWithPublicKey(const std::string &message, const std::string &publicKey, const std::string &signature) const = 0;
 
     virtual bool parse_uri(const std::string &uri, std::string &address, std::string &payment_id, uint64_t &amount, std::string &tx_description, std::string &recipient_name, std::vector<std::string> &unknown_parameters, std::string &error) = 0;
+    virtual bool parse_uri(const std::string &uri, std::vector<std::string> &addresses, std::vector<std::uint64_t> &amounts, std::vector<std::string> &recipient_names, std::string &payment_id, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error);
+    
     virtual std::string make_uri(const std::string &address, const std::string &payment_id, uint64_t amount, const std::string &tx_description, const std::string &recipient_name, std::string &error) const = 0;
-
+    virtual std::string make_uri(std::vector<std::string> &addresses, std::vector<std::uint64_t> &amounts, std::vector<std::string> &recipient_names, const std::string &payment_id, const std::string &tx_description, std::string &error) const;
```
SNeedlewoods:
```suggestion
    virtual std::string make_uri(std::vector<std::string> &addresses, std::vector<std::uint64_t> &amounts, std::vector<std::string> &recipient_names, const std::string &payment_id, const std::string &tx_description, std::string &error) const = 0;
```

U65535F:
@SNeedlewoods Please elaborate on your point on using `override` and `virtual` keywords. I don't get it.

U65535F:
@SNeedlewoods I think I've addressed your issues and fixed them. Please check again if you are willing to, and tell me if I missed anything. As it seems, I've missed 2 unnecessary edits in `uri.py` which I will tackle in newer commits.

U65535F:
@selsta How do I get more people to review it?

U65535F:
@selsta ping

selsta:
I'm busy with the current release, will be able to take a look afterwards.

U65535F:
@selsta

U65535F:
@selsta ping .

selsta:
Like I said, I won't have time to re-review until v0.18.4.0 is released. The release has priority currently.

selsta:
There are over 200 open pull requests and we are working on a release. Contributing to open source software often requires patience, I hope you understand that.

Anyway I appreciate you trying to work on this, even if you closed this now.

U65535F:
> There are over 200 open pull requests and we are working on a release. Contributing to open source software often requires patience, I hope you understand that.
> 
> Anyway I appreciate you trying to work on this, even if you closed it now.

It's not a given up. I somehow deleted the commits and force pushed the changes due to some silly mistake while squashing commits into one. I guess I will rewrite the code.

PR #9761, opened 2025-02-01
utils: Simplify python RPC Response implementation

iamamyth:
Reduce bloat in the python RPC Response class by making better use of superclass methods and recursion. As part of this change, add tests for the Response class.


0xFFFC0000 (approved):
lgtm

tobtoht (approved):

Issue #9755, opened 2025-01-29
p2p functional test failure


selsta:
https://github.com/monero-project/monero/actions/runs/12969219043/job/36173269466?pr=9723#step:11:1380

```
[TEST STARTED] p2p
Resetting blockchain
Creating wallet
Generating 80 blocks
Testing P2P reorg
Testing P2P tx propagation
Traceback (most recent call last):
  File "/home/runner/work/monero/monero/tests/functional_tests/p2p.py", line 187, in <module>
    P2PTest().run_test()
  File "/home/runner/work/monero/monero/tests/functional_tests/p2p.py", line 46, in run_test
    self.test_p2p_tx_propagation()
  File "/home/runner/work/monero/monero/tests/functional_tests/p2p.py", line 182, in test_p2p_tx_propagation
    assert len(res.tx_hashes) == 1
  File "/home/runner/work/monero/monero/utils/python-rpc/framework/rpc.py", line 50, in __getattr__
    return self[key]
KeyError: 'tx_hashes'
[TEST FAILED] p2p
```

Before we can put out v0.18.4.0 we have investigate this to make sure this is not a regression with recent merges. The test failure does not show up consistently.

selsta:
@vtnerd could this be related to https://github.com/monero-project/monero/pull/9459?

vtnerd:
A quick question before I dig deeper: have you seen this on the `release-v0.18` branch or just the `master` branch? The linked CI build was for `master` so I'm testing against that for now.

selsta:
It did not trigger on `release-v0.18` yet. It generally only triggered once so far, seems to be a rare bug.

vtnerd:
I haven't gotten this to trigger locally, so I have no daemon logs to investigate. And I don't see a way to download the daemon logs from Github.

My best guess is that this is a timing bug in the test itself. The sleep timeout is 5 seconds, but the fluff (incoming) timer for transactions is typically going to be in the 3-7.25 second range. The counter argument is that it seems like the test should fail more frequently if this were the culprit.

I don't have any other theories as to the cause at the moment. #9459 definitely seems unrelated, unless there was a p2p connection loss + reconnect just before `test_p2p_tx_propagation` ran. Otherwise, if #9459 were responsible, there should've been an error earlier in the test functions.

iamamyth:
I think the test can be improved quite a bit to avoid such spurious failures.

iamamyth:
See https://github.com/monero-project/monero/pull/9762

[...]

PR #9762, opened 2025-02-01
tests: Improve p2p tx propagation functional test

iamamyth:
Reduce the likelihood of false positive failures in the p2p transaction propagation functional test by waiting up to a maximum timeout for a transaction to propagate, rather than using a fixed timeout, to reflect the random delay of Dandelion++ transaction propagation. This strategy also speeds test execution in cases where propagation occurs faster than the previously expected fixed delay.

### DIFF
--- a/tests/functional_tests/p2p.py
+++ b/tests/functional_tests/p2p.py
@@ -160,12 +160,11 @@ class P2PTest():
 
     def test_p2p_tx_propagation(self):
         print('Testing P2P tx propagation')
-        daemon2 = Daemon(idx = 2)
-        daemon3 = Daemon(idx = 3)
+        daemons = (Daemon(idx=2), Daemon(idx=3))
 
-        for daemon in [daemon2, daemon3]:
+        for daemon in daemons:
             res = daemon.get_transaction_pool_hashes()
-            assert not 'tx_hashes' in res or len(res.tx_hashes) == 0
+            assert len(res.get('tx_hashes', [])) == 0
 
         self.wallet.refresh()
         res = self.wallet.get_balance()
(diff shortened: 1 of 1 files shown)


tests/functional_tests/p2p.py
```diff
@@ -175,12 +174,25 @@ def test_p2p_tx_propagation(self):
         assert len(res.tx_hash) == 32*2
         txid = res.tx_hash
 
-        time.sleep(5)
-
-        for daemon in [daemon2, daemon3]:
+        # Due to Dandelion++, the network propagates transactions with a
+        # random delay, so poll for the transaction with a timeout
+        timeout = 16
```
iamamyth, line 179: @vtnerd Do you know the expected maximum propagation delay in this scenario, based on the D++ paper? That value would act as an approximate lower bound for the timeout, which could then be further padded to reflect physical realities of data transmission and processing.

vtnerd: The delay should usually trigger the inbound fluff delay where it's using a poisson distribution, simply mimicking what Bitcoin was doing for the same situation. 95% of the values are in the 3-7.3 second range. A delay of 16 is basically nearly impossibl, so this should be acceptable.

As an explanation of why not exponential here - one of the nodes will have no outbound peers. I don't recall the paper stating how to handle the situation, so I decided to make it immediately fluff, as it would be an edge case. A fluff does randomized poisson delays, similar to what Bitcoin was doing at the time. I don't recall the d++ specifying how to handle fluff precisely either, maybe I need to revisit that paper.

vtnerd:
I guess you were convinced the failure was primarily the sleep timeout? That was my assessment, as it seemed like an obvious issue.

iamamyth:
I think the low, fixed timeout generates quite a few false positive failures, consistent with the observed behavior of this test in CI. If any actual transaction propagation errors exist, I would expect they do not owe to recent connection management commits. I just modified the test to better differentiate propagation to 0-2 daemons, which might make it a bit more useful.

iamamyth:
Test failure is a known-flaky unit test (`node_server.race_condition`) unrelated to this PR.

iamamyth:
@vtnerd I made one minor change to show how many daemons see the transactions, but I think this will clear up the p2p test issue.

iamamyth:
An example failure in a recent CI build (without this change): https://github.com/monero-project/monero/actions/runs/13163403906/job/36738631593?pr=9771.

Every failure I've seen on CI (and I've seen quite a few, at this point) is the same behavior, it's not the RPC refusing the connection, or generating a garbage response, it's simply the transaction not appearing in the pool by design; the failures are successes and the test has a wrong methodology.

selsta:
> Reduce the likelihood of false positive failures in the p2p transaction propagation functional test by waiting up to a maximum timeout for a transaction to propagate, rather than using a fixed timeout, to reflect the random delay of Dandelion++ transaction propagation. 

I'm not against this patch but what I don't understand is why this test only recently started to fail randomly, if it's just timeout probability. To me it seems there's more to it.

iamamyth:
> I'm not against this patch but what I don't understand is why this test only recently started to fail randomly, if it's just timeout probability. To me it seems there's more to it.

I don't know when the failures started (CI has so many broken tests, even now, that I imagine many people just ignore it). However, as stated, due to the test's structural flaw, if, in fact, it used to pass and now fails, that may reflect an *improvement* in Dandelion++ or networking code, because it should have had a moderate probability of failure in the past.

tests/functional_tests/p2p.py
iamamyth: @vtnerd Per discussion in https://github.com/monero-project/monero/issues/9755, it might help to make this upper bound a bit more strict. As a 0% failure probability would be impossible for any test (CPUs can fail, etc), I'd aim for something around 1 in 1 million. How much could I reduce the timeout and achieve that bound?

iamamyth: Unless I've done the math horribly wrong, it looks like 13s should yield a failure probability of roughly 1 in a billion (I decided a million was too low); I've updated the test timeout accordingly.

vtnerd (approved):

selsta:
@iamamyth please also open against `release-v0.18`.

0xFFFC0000 (approved):

PR #9763, opened 2025-02-01
tests: Improve functional_tests_rpc startup

iamamyth:
* Use a global startup timeout, and apply it to each socket's connect call, rather than repeating connection attempts with short timeouts.

* Avoid leaking sockets by closing them.


0xFFFC0000 (approved):
lgtm. thanks.

tobtoht (approved):

PR #9765, opened 2025-02-01
Add configuration options to HTTP server

vtnerd:


selsta:
What would be a scenario where someone changes `rpc-response-soft-limit`?

selsta:
This seems to break multiple functional tests, not sure if it's the IP limit or the response size limit.

iamamyth:
The RPC client used in those tests does a poor job of connection management: Depending on the exact version of the requests library in use, it either opens and closes a new connection for every request (recent versions of the library), or re-uses a global, "default" session, which makes it tricky to clean up. I'd check the IP limit, and also, it might be appropriate to exempt local subnets.

src/rpc/core_rpc_server.cpp
```diff
@@ -3885,4 +3889,16 @@ namespace cryptonote
...
+      "rpc-max-ip-connections"
+    , "Max RPC connections per IP permitted"
+    , 3
+  };
+
+  const command_line::arg_descriptor<std::size_t> core_rpc_server::arg_rpc_response_soft_limit = {
+      "rpc-response-soft-limit"
+    , "Max response bytes that can be queued, enforced at next response attempt"
+    , 25 * 1024 * 1024
+  };
```
SyntheticBird45, line 3903: There is a default 25MiB limit here that replaces the `ABSTRACT_SERVER_SEND_QUE_MAX_BYTES_DEFAULT` of 100MiB in any case. It's unclear what is `ABSTRACT_SERVER_SEND_QUE_MAX_BYTES_DEFAULT`. Is the default limit 25MiB or 100MiB ?

contrib/epee/include/net/http_server_impl_base.h
```diff
@@ -60,7 +61,8 @@ namespace epee
       const std::string& bind_ipv6_address = "::", bool use_ipv6 = false, bool require_ipv4 = true,
       std::vector<std::string> access_control_origins = std::vector<std::string>(),
       boost::optional<net_utils::http::login> user = boost::none,
-      net_utils::ssl_options_t ssl_options = net_utils::ssl_support_t::e_ssl_support_autodetect)
+      net_utils::ssl_options_t ssl_options = net_utils::ssl_support_t::e_ssl_support_autodetect,
+      const std::size_t max_ip_connections = 3, const std::size_t response_soft_limit = 0)
```
SyntheticBird45, line 65: Are you sure `response_soft_limit` should fallback to 0 if undefined ?

vtnerd:
Sorry about this obvious failure - only ran ad-hoc tests for this and it shows. Ill figure out a fix shortly.

iamamyth:
The connection limit seems more nuanced and difficult. You may want to consider splitting the payload and connection limits into separate PRs, as the payload limit should be rather straightforward.

SyntheticBird45:
> The connection limit seems more nuanced and difficult. You may want to consider splitting the payload and connection limits into separate PRs, as the payload limit should be rather straightforward.

I disagree, these are related enough, let's not waste time by splitting this PR. I would like by a miracle to see this merged into the incoming release.

0xFFFC0000 (commented):
1. I would like to see programmatical way to set and get `max_ip_connections`.

I have been playing with this test case for this PR, but so far not able to be successful. The scaffold of the test is ready. and a single test is included too. You can add extra tests to it too : 

```

diff --git a/tests/unit_tests/CMakeLists.txt b/tests/unit_tests/CMakeLists.txt
index e329b7506..1c94755f7 100644
--- a/tests/unit_tests/CMakeLists.txt
+++ b/tests/unit_tests/CMakeLists.txt
@@ -102,6 +102,7 @@ set(unit_tests_sources
   is_hdd.cpp
   aligned.cpp
   rpc_version_str.cpp
+  rpc_server.cpp
   zmq_rpc.cpp)
 
 set(unit_tests_headers
diff --git a/tests/unit_tests/rpc_server.cpp b/tests/unit_tests/rpc_server.cpp
new file mode 100644
index 000000000..cd574f483
--- /dev/null
+++ b/tests/unit_tests/rpc_server.cpp
@@ -0,0 +1,128 @@
+// Copyright (c) 2019-2024, The Monero Project
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+// 
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+// 
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+// 
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+
+#include "gtest/gtest.h"
+#include "rpc/core_rpc_server.h"
+#include "net/http_client.h"
+#include "net/http_server_impl_base.h"
+#include "net/http_protocol_handler.h"
+#include "net/net_utils_base.h"
+#include "cryptonote_core/cryptonote_core.h"
+#include "p2p/net_node.h"
+#include <thread>
+#include <chrono>
+#include <boost/program_options.hpp>
+
+// Create a test subclass of core_rpc_server to expose a public setter for response limit.
+class TestCoreRpcServer : public cryptonote::core_rpc_server {
+public:
+    using cryptonote::core_rpc_server::core_rpc_server; // inherit constructors
+
+    void set_test_response_soft_limit(std::size_t limit) {
+        // Expose the call to the underlying boosted_tcp_server's set_response_soft_limit
+        m_net_server.set_response_soft_limit(limit);
+    }
+};
+
+class RpcServerTest : public ::testing::Test {
+protected:
+    void SetUp() override { 
+        boost::program_options::options_description opts{};
+        nodetool::node_server<cryptonote::t_cryptonote_protocol_handler<cryptonote::core>>::init_options(opts);
+        cryptonote::core::init_options(opts);
+        cryptonote::core_rpc_server::init_options(opts);
+
+        const char* argv[] = {"monerod", "--rpc-bind-port", "18081", "--no-igd"};
+        int argc = 4;
+        boost::program_options::variables_map vm;
+        boost::program_options::store(
+          boost::program_options::parse_command_line(argc, argv, opts), vm
+        );  
+
+        // Initialize core and protocol handler
+        core = new cryptonote::core(nullptr);
+        protocol_handler = new cryptonote::t_cryptonote_protocol_handler<cryptonote::core>(*core, nullptr, false);
+
+        // Initialize node server with protocol handler
+        node_server = new nodetool::node_server<cryptonote::t_cryptonote_protocol_handler<cryptonote::core>>(*protocol_handler);
+
+        // Create the RPC server with core and node server using our test subclass.
+        server = new TestCoreRpcServer(*core, *node_server);
+
+        // Set response soft limit to 1 MB via the public method.
+        static_cast<TestCoreRpcServer*>(server)->set_test_response_soft_limit(1024 * 1024);
+
+        // Start the server on a separate thread
+        server_thread = std::thread([this, &vm]() {
+            server->init(vm, false, "", false, "");
+            server->run(1, true);
+        });
+
+        // Wait to ensure the server has started
+        std::this_thread::sleep_for(std::chrono::seconds(2));
+    }
+
+    void TearDown() override {
+        // Stop the server and join the thread
+        server->send_stop_signal();
+        if (server_thread.joinable())
+            server_thread.join();
+        delete server;
+        delete node_server;
+        delete protocol_handler;
+        delete core;
+    }
+
+    cryptonote::core_rpc_server* server;
+    cryptonote::core* core;
+    cryptonote::t_cryptonote_protocol_handler<cryptonote::core>* protocol_handler;
+    nodetool::node_server<cryptonote::t_cryptonote_protocol_handler<cryptonote::core>>* node_server;
+    std::thread server_thread;
+};
+
+TEST_F(RpcServerTest, ResponseSizeLimit) {
+    // Create an HTTP client to send requests to the RPC server.
+    epee::net_utils::http::http_simple_client http_client;
+    epee::net_utils::http::fields_list fields;
+
+    // Prepare a request expected to produce a large response.
+    std::string request_body = R"({"jsonrpc":"2.0","id":"0","method":"get_block_headers_range","params":{"start_height":0,"end_height":100000}})";
+
+    // The correct signature expects:
+    // (URI, method, body, timeout, response_info pointer, additional fields)
+    const epee::net_utils::http::http_response_info* response_info = nullptr;
+    bool success = http_client.invoke("http://127.0.0.1:18081/json_rpc", "POST",
+                                      request_body, std::chrono::milliseconds(10000),
+                                      &response_info, std::move(fields));
+
+    // Verify the HTTP call was successful.
+    ASSERT_TRUE(success);
+
+    // Verify the response body is within the configured 1 MB limit.
+    ASSERT_LE(response_info->m_body.size(), 1024 * 1024);
+}
\ No newline at end of file

```

src/rpc/core_rpc_server.cpp
vtnerd: The default for P2P is 100MiB and the default for HTTP is 25MiB. P2P has actually a hard-limit of 62.6 MiB due to how the epee server code buffers responses.

contrib/epee/include/net/http_server_impl_base.h
vtnerd: No, I overlooked this before submitting the patch, and will make it 25MiB.

vtnerd:
@0xFFFC0000 are you hoping for a unit test for this? I think that's you meant.

selsta:
Please also open again release branch, we will include it for v0.18.4.0.

src/rpc/core_rpc_server.cpp
```diff
@@ -163,6 +163,8 @@ namespace cryptonote
     command_line::add_arg(desc, arg_rpc_payment_difficulty);
     command_line::add_arg(desc, arg_rpc_payment_credits);
     command_line::add_arg(desc, arg_rpc_payment_allow_free_loopback);
+    command_line::add_arg(desc, arg_rpc_max_ip_connections);
```
nahuhh, line 166: We have `max-connections-per-ip` for p2p.
probably best to remain consistent and use `rpc-max-connections-per-ip`

src/rpc/core_rpc_server.cpp
```diff
@@ -3885,4 +3889,16 @@ namespace cryptonote
     , "Allow free access from the loopback address (ie, the local host)"
     , false
     };
+
+  const command_line::arg_descriptor<std::size_t> core_rpc_server::arg_rpc_max_ip_connections = {
+      "rpc-max-ip-connections"
+    , "Max RPC connections per IP permitted"
+    , 3
```
nahuhh, line 3896: What about localhost or lan connections, typically coming from onion / i2p setups?

vtnerd: The next patch will have `rpc-max-connections-per-public-ip` and `rpc-mac-conections-per-private-ip`. `localhost` connections will fall under the `private-ip` category.

vtnerd: BTW, I'm not really sure how to handle Tor RPC, its arguably broken.

SyntheticBird45: Yeah not much can be done regarding it, public onion nodes will have to make their own load balancer and limiter.

src/rpc/core_rpc_server.cpp
vtnerd: Updated in force push.

vtnerd:
Force pushed:
  * A fix in the epee TCP server where the socket wouldn't cleanup/close despite starting the `terminate()` sequence.
  * A flag separating maxes for private and public IPs were added. `localhost` connections are considered private.
  * Added a "kludge" where max private IPs is set to 0 for the functional tests. This works locally, will see about CI server
  * Fixed default value for `response_soft_limit` (from 0 to 25MiB).

If all tests pass, I will open on release branch.

vtnerd:
Also @0xFFFC0000 I added two unit tests. I testing the http code directly instead of the method you were using. This helped ensure that the TCP server was acting as expected. This caught a difficult to catch/fix bug involving async op cancellations and ASIO composed handlers.

0xFFFC0000:
> Also @0xFFFC0000 I added two unit tests. I testing the http code directly instead of the method you were using. This helped ensure that the TCP server was acting as expected. This caught a difficult to catch/fix bug involving async op cancellations and ASIO composed handlers.

Glad to hear that.

src/rpc/core_rpc_server.cpp
iamamyth: I'd suggest a higher limit for private connections, somewhere in the 5-10 range. The functional tests might actually provide helpful guidance on a suitable limit, precisely because they have a flawed, but common, connection management strategy (i.e. just figure out whatever value works with that test suite, and maybe round it a bit).

SyntheticBird45: > I'd suggest a higher limit for private connections, somewhere in the 5-10 range. The functional tests might actually provide helpful guidance on a suitable limit, precisely because they have a flawed, but common, connection management strategy (i.e. just figure out whatever value works with that test suite, and maybe round it a bit).

force push just set max connection per private ip to 100.

iamamyth: Right, it's 100 in the test runs (which effectively passed, because the failure is a methodologically flawed test), but still 3 by default in the daemon. I suspect the limit in the tests can actually be lowered and synchronized with the daemon's default.

SyntheticBird45: Oh you're right I misread.

vtnerd:
Another force push:
  * Changed limits to `--rpc-max-connections-per-ip` and `--rpc-max-connections`.
  * Hopefully fixed builds with older Boost releases (Beast issues).

As I side note that I hadn't mentioned previously: the new unit test uses `boost::beast` for testing. So we will also gain limited external verification for our http server.

SyntheticBird45 (approved):

vtnerd:
I've had private discussions about an additional change to this patch. The change would have 3 flags: (1) rpc-max-connections-per-public-ip, (2) rpc-max-connections-per-private-ip, and (3) rpc-max-connections.

The desired goal is to set different limits for Tor/I2P connections. The `per-private-ip` achieves that, but has other downsides. I'm not going to add this without discussion, as I've already got an approval from one contributor.

tests/unit_tests/epee_http_server.cpp
```diff
@@ -0,0 +1,192 @@
...
+  {
+    streams.emplace_back(context);
+    streams.back().connect(
+      boost::asio::ip::tcp::endpoint{
+        boost::asio::ip::make_address("127.0.0.1"), 8080
+      },
+      error
+    );
+  }
+
```
iamamyth, line 164: You may want an additional test for the negative case, i.e. requests work fine up until the limit.

contrib/epee/include/net/http_protocol_handler.inl
```diff
@@ -208,11 +208,45 @@ namespace net_utils
...
 	//--------------------------------------------------------------------------------------------
+	template<class t_connection_context>
+	simple_http_connection_handler<t_connection_context>::~simple_http_connection_handler()
+	{
+	  try
+	  {
+	    if (m_initialized)
+	    {
+	      CRITICAL_REGION_LOCAL(m_config.m_lock);
+	      m_config.m_connection_count -= std::min(std::size_t(1), m_config.m_connection_count);
```
iamamyth, line 225: I think you could remove the `min` by pushing this into the `if(connection found)`.

selsta:
Is it necessary to set the max RPC connections limit to 100? It would mean by default public RPC nodes would stop working (though I don't know if they support 100 connections in the first place), and a single entity with 100 IPs could basically open connections to all public nodes and make them reject connections from legit useres.

SyntheticBird45:
> Is it necessary to set the max RPC connections limit to 100? It would mean by default public RPC nodes would stop working (though I don't know if they support 100 connections in the first place), and a single entity with 100 IPs could basically open connections to all public nodes and make them reject connections from legit useres.

It's really about max resource usage and public nodes should absolutely tweak the rpc-soft-limit down to what is strictly necessary to increase max rpc connections

iamamyth:
> Is it necessary to set the max RPC connections limit to 100?

I don't know what precipitated the in-daemon connection limit (can be done with a firewall), but, one could argue that the principle of least surprise favors not auto-configuring a new limit (i.e. keeping unlimited as the default). The data limit, on the other hand, falls into a separate category (nothing should exceed it now, doing so creates stability problems, and the networking stack can't see data which hasn't been sent to the socket).

nahuhh:
> I don't know what precipitated the in-daemon connection limit (can be done with a firewall), but, one could argue that the principle of least surprise favors not auto-configuring a new limit (i.e. keeping unlimited as the default). The data limit, on the other hand, falls into a separate category (nothing should exceed it now, doing so creates stability problems, and the networking stack can't see data which hasn't been sent to the socket).

someone who controls subnets (like the spynodes) can flood 5 connections per ip. Thats 1280 connections for one /24 subnet.

currently its a potential 65k connections for 1 subnet.

@Gingeropolous might have good numbers for his rpc connections. Ive seen him report 73 etc before. @plowsof runs some popular nodes and doesnt have near 100 connections (less than 10 last check). Someone like cake, who runs a very high specced and high use node, would likely increase max to 1000 or greater.

SyntheticBird45:
@nahuhh Big public nodes like Cake Wallet uses load balancer/reverse http proxy which let them be very liberal on the limit because it's the proxy responsibility to ensure these limits.

tests/unit_tests/epee_http_server.cpp
vtnerd: Added some read/writes in the loop to ensure all of the connections work, before the 1+ case fails.

contrib/epee/include/net/http_protocol_handler.inl
vtnerd: Converted this into a simple `if (m_connection_count) --m_connection_count`

vtnerd:
Force pushed (twice):
  * Now have 3 flags
  * The unit tests for connection limit does a read/write sequence to ensure each connection works
  * Changed a `min(...)` call to `if (foo) --foo`
  * Reordered the addition after the `operator[]` call in case it throws.

Watching the CI before pushing to the other PR

Gingeropolous:
> @Gingeropolous might have good numbers for his rpc connections. Ive seen him report 73 etc before. @plowsof runs some popular nodes and doesnt have near 100 connections (less than 10 last check). Someone like cake, who runs a very high specced and high use node, would likely increase max to 1000 or greater.

Yeah, so my experience was having 70+ RPC connections, and at one point I found that there were multiple connections from the same IP address. This made the node unresponsive at times, and led me to create my incredibly hacky p2r2n_defender script:

https://github.com/Gingeropolous/p2r2n_defender/blob/main/blockddos_v1.sh

this PR will be great for public RPC nodes.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -129,6 +129,10 @@ namespace
   const command_line::arg_descriptor<std::string> arg_wallet_dir = {"wallet-dir", "Directory for newly created wallets"};
   const command_line::arg_descriptor<bool> arg_prompt_for_password = {"prompt-for-password", "Prompts for password when not provided", false};
   const command_line::arg_descriptor<bool> arg_no_initial_sync = {"no-initial-sync", "Skips the initial sync before listening for connections", false};
+  const command_line::arg_descriptor<std::size_t> arg_rpc_max_connections_per_public_ip = {"rpc-max-connections-per-public-ip", "Max RPC connections per public IP permitted", 5};
+  const command_line::arg_descriptor<std::size_t> arg_rpc_max_connections_per_private_ip = {"rpc-max-connections-per-private-ip", "Max RPC connections per private and localhost IP permitted", 5};
+  const command_line::arg_descriptor<std::size_t> arg_rpc_max_connections = {"rpc-max-connections", "Max RPC connections permitted", 100};
```
nahuhh, line 134: is wallet-rpc intended to have different limits?
probably fine (perhaps even preferable?)

Gingeropolous:
@nahuhh , that was 64GB ram with sata ssd.

vtnerd:
Force pushed fix for the functional tests. Sorry about that.

src/wallet/wallet_rpc_server.cpp
vtnerd: No, and I just a force push. Crap. Will update this.

vtnerd:
Force push for a fix for default values in the wallet server.

iamamyth (approved):

contrib/epee/include/net/http_protocol_handler.inl
```diff
@@ -208,11 +208,46 @@ namespace net_utils
...
 	//--------------------------------------------------------------------------------------------
+	template<class t_connection_context>
+	simple_http_connection_handler<t_connection_context>::~simple_http_connection_handler()
+	{
+	  try
+	  {
+	    if (m_initialized)
+	    {
+	      CRITICAL_REGION_LOCAL(m_config.m_lock);
+        if (m_config.m_connection_count)
```
iamamyth, line 225: Minor style issue (not material for approval): If GitHub browser view is correct, the indent here appears to differ from elsewhere.

SyntheticBird45 (approved):

nahuhh:
if setting per-ip connections higher than max connections, max connections is still the deciding factor. 
A) use sum of values `max_connections = (per_private_ip + per_public_ip) > max_connections ?  (per_private_ip + per_public_ip) : max_connections` or
B) should throw an error that values are too high

Real test flags:
--rpc-max-connections-per-public-ip=5
--rpc-max-connections-per-private-ip=5
--rpx-max-connections=2
Result: only 2 connections are allowed
expected: 10 connections

example use case:
User runs an onion only node and is allowing 200 onion connections and sets `--rpc-max-connections-per-private-ip=200`.

result: only 100 connections are allowed. 

expected: if sum of per ip values are higher than max connections, max connections should be set to the sum of the per ip values. In this scenario, max connections would be set to 203.

contrib/epee/include/net/http_protocol_handler.inl
vtnerd: It's a tab vs space issue. This file has some funky indenting. Is it worth force pushing an update over?

vtnerd:
>  B) should throw an error that values are too high

It should definitely do this instead of A. I guess we do guard against other configuration mishaps, so I will force push this too.

nahuhh:
> > if setting per-ip connections higher than max connections, max connections is still the deciding factor.
> 
> That's indeed what I would expect and this behavior is based on the reasoning that first introduced it

If i manually set `rpc-max-connections-per-*-ip` to 200, i would expect 200 connections, not 100. I might not have any idea that there is a 100 limit. 

expected would be an error that tells you that your flag isnt going to work (and how to fix), or to automatically adjust max_connections to be the higher of your custom per-ip limits vs the default.

SyntheticBird45:
> > > if setting per-ip connections higher than max connections, max connections is still the deciding factor.
> > 
> > 
> > That's indeed what I would expect and this behavior is based on the reasoning that first introduced it
> 
> If i manually set `rpc-max-connections-per-*-ip` to 200, i would expect 200 connections, not 100. I might not have any idea that there is a 100 limit.
> 
> expected would be an error that tells you that your flag isnt going to work (and how to fix), or to automatically adjust max_connections to be the higher of your custom per-ip limits vs the default.

Oh yeah yeah the error is indeed a good remark. I'm just notating here that, even contradictory, I would have expected this takeover. Also notating that A don't make sense here because of the primary reason this flag was added.

contrib/epee/include/net/http_protocol_handler.inl
iamamyth: Yea I figured, that's why I added this comment post-approval, to clarify that I don't care much. I would say just do whatever you want here, if you're pushing a change anyways to warn of inconsistent configs, then maybe it's trivial to include.

vtnerd:
Force pushed:
  * Now check to ensure that private-ip and public-ip are less than or equal to max-connections
  * Fixed tabbing

contrib/epee/include/net/http_protocol_handler.inl
vtnerd: Fixed.

SyntheticBird45 (approved):

iamamyth (approved):

contrib/epee/include/net/http_server_impl_base.h
```diff
@@ -60,8 +61,16 @@ namespace epee
...
+      net_utils::ssl_options_t ssl_options = net_utils::ssl_support_t::e_ssl_support_autodetect,
+      const std::size_t max_public_ip_connections = 3,
+      const std::size_t max_private_ip_connections = 25,
+      const std::size_t max_connections = 100,
+      const std::size_t response_soft_limit = 25 * 1024 * 1024)
     {
+      if (max_connections < max_public_ip_connections)
+        throw std::logic_error{"Max public IP connections cannot be more than max connections"};
+      if (max_connections < max_private_ip_connections)
+        throw std::logic_error{"Max private IP connections cannot be more than max connections"};
```
jeffro256, line 73: A bit pedantic, but `std::invalid_argument` would be a better fit here. It's not really a "logic error" if a higher level caller provides arguments that fail a basic test, it's just bad arguments

contrib/epee/include/net/http_server_impl_base.h
```diff
@@ -60,8 +61,16 @@ namespace epee
       const std::string& bind_ipv6_address = "::", bool use_ipv6 = false, bool require_ipv4 = true,
       std::vector<std::string> access_control_origins = std::vector<std::string>(),
       boost::optional<net_utils::http::login> user = boost::none,
-      net_utils::ssl_options_t ssl_options = net_utils::ssl_support_t::e_ssl_support_autodetect)
+      net_utils::ssl_options_t ssl_options = net_utils::ssl_support_t::e_ssl_support_autodetect,
+      const std::size_t max_public_ip_connections = 3,
+      const std::size_t max_private_ip_connections = 25,
+      const std::size_t max_connections = 100,
+      const std::size_t response_soft_limit = 25 * 1024 * 1024)
```
jeffro256, line 68: We should put this `25 * 1024 * 1024` value, which shows up 3 times in the patch, into a named constant

src/rpc/core_rpc_server.cpp
```diff
@@ -3885,4 +3906,28 @@ namespace cryptonote
     , "Allow free access from the loopback address (ie, the local host)"
     , false
     };
+
+  const command_line::arg_descriptor<std::size_t> core_rpc_server::arg_rpc_max_connections_per_public_ip = {
+      "rpc-max-connections-per-public-ip"
+    , "Max RPC connections per public IP permitted"
+    , 3
```
jeffro256, line 3913: Same for the default max number of connections per public/private ip

jeffro256:
It might also be worth it to change the description for `--max-connections-per-ip` option to specify that *that* limit is for P2P connections, not RPC.

contrib/epee/include/net/http_server_impl_base.h
vtnerd: Yup, forgot about that somehow.

contrib/epee/include/net/http_server_impl_base.h
vtnerd: I was avoiding this, because I had no idea where to place such a thing. Looks like good old `cryptonote_config.h` will work, even with epee.

vtnerd:
Force pushed:
  * A rebase to master
  * `cryptonote_config.h` now has the default values
  * Updated `--max-connections-per-ip` description

contrib/epee/include/net/http_server_impl_base.h
vtnerd: Done.

src/cryptonote_config.h
```diff
@@ -127,6 +127,10 @@
 
 #define COMMAND_RPC_GET_BLOCKS_FAST_MAX_BLOCK_COUNT     1000
 #define COMMAND_RPC_GET_BLOCKS_FAST_MAX_TX_COUNT        20000
+#define DEFAULT_RPC_MAX_CONNECTIONS_PER_PUBLIC_IP       3
+#define DEFAULT_RPC_MAX_CONNECTIONS_PER_PRIVATE_IP      25
+#define DEFAULT_RPC_MAX_CONNECTIONS                     100
+#define DEFAULT_RPC_SOFT_LIMIT_SIZE                     25 * 1025 * 1025 // 25 MiB
```
iamamyth, line 133: This appears to be a typo (1025 should be 1024).

vtnerd:
The failed test should be unrelated, its a known problem test.

src/cryptonote_config.h
vtnerd: Wow. Will fix.

vtnerd:
Force pushed a fix for the `1025` issue.

jeffro256 (approved):

iamamyth (approved):

Gingeropolous:
currently running master with 9765 and 9135 on xmrchain.net

PR #9772, opened 2025-02-05
easylogging++: Generalize TERM rule for color-supporting terminals

laanwj:
With newer versions of tmux, such as version 3.4 packaged by Ubuntu 24.04, `tmux-256color` the default TERM variable value.

Instead of adding it specifically, extend the rule to look for `-color` or `-256color` in TERM, as these can all be assumed to support color.

### DIFF
--- a/external/easylogging++/easylogging++.cc
+++ b/external/easylogging++/easylogging++.cc
@@ -1246,9 +1246,9 @@ std::string OS::currentHost(void) {
 
 bool OS::termSupportsColor(void) {
   std::string term = getEnvironmentVariable("TERM", "");
-  return term == "xterm" || term == "xterm-color" || term == "xterm-256color"
-         || term == "screen" || term == "linux" || term == "cygwin"
-         || term == "screen-256color" || term == "screen.xterm-256color";
+  return term == "xterm" || term == "screen" || term == "linux" || term == "cygwin"
+         || term.find("-color") != std::string::npos
+         || term.find("-256color") != std::string::npos;
 }
 
 // DateTime


selsta (approved):

external/easylogging++/easylogging++.cc
```diff
@@ -1248,7 +1248,8 @@ bool OS::termSupportsColor(void) {
   std::string term = getEnvironmentVariable("TERM", "");
   return term == "xterm" || term == "xterm-color" || term == "xterm-256color"
          || term == "screen" || term == "linux" || term == "cygwin"
-         || term == "screen-256color" || term == "screen.xterm-256color";
+         || term == "screen-256color" || term == "screen.xterm-256color"
```
iamamyth, line 1251: Are there any terminals ending in "-256color" which don't support color? Might make sense to just future-proof it rather than constantly whitelisting new terminals, either way, someone can always use the escape hatch of setting TERM.

laanwj: i agree !

i thought about just checking for "-256color" but in a way it's just as hacky, there's plenty of terminals that support color but don't have that in the name, but do have their own terminfo.

The official way would be to query the terminfo database eg `tigetnum("colors") >= 256` if 256 colors are wanted. But this needs a library like ncurses or termcap, not sure that's a current dependency.

iamamyth: The only relevant concern would be whether one needs a whitelist because there are terminals ending in "-256color" which don't support color; I doubt there are. Using a terminfo database for something so trivial (colorized logging) seems unwarranted.

laanwj: > The only relevant concern would be whether one needs a whitelist because there are terminals ending in "-256color" which don't support color; I doubt there are.

Makes sense, will update it to check for `-color` or `-256color` instead.

laanwj:
Okay, implemented the string search instead.

Another alternative would be to use a blacklist instead of a whitelist. All modern terminals can at least parse basic ANSI 16-color sequences, although some might ignore them for aesthetic reasons.

Exceptions would be TERM empty or non-existent (eg as on Windows), `TERM=dumb` (no escapes support at all), can't think of any other.

external/easylogging++/easylogging++.cc
```diff
@@ -1246,9 +1246,9 @@ std::string OS::currentHost(void) {
 
 bool OS::termSupportsColor(void) {
   std::string term = getEnvironmentVariable("TERM", "");
-  return term == "xterm" || term == "xterm-color" || term == "xterm-256color"
-         || term == "screen" || term == "linux" || term == "cygwin"
-         || term == "screen-256color" || term == "screen.xterm-256color";
+  return term == "xterm" || term == "screen" || term == "linux" || term == "cygwin"
+         || term.find("-color") != std::string::npos
```
iamamyth, line 1250: I'm OK with this choice, though I was expecting just a check for the ending (sadly, `ends_with` requires c++20, but an equivalent is trivial using `compare`).

iamamyth (approved):

tobtoht (approved):

external/easylogging++/easylogging++.cc
laanwj: Do you think that really maters for the heuristic if it's at the end? What if it's `tmux-color-graphics` or whatever.

iamamyth: Possible variants of `-(256)?color` are why I'm OK with the choice. However, when inferring a rule from a pattern, I think it helps to do so as narrowly as possible: This code has existed for a long time, and, in all that time, -color-x has never arisen, whereas -color and -256color have. So, the narrow choice clarifies to the reader (including the compiler, and therefore the machine which runs the code): I know exactly where this should occur in the string. Furthermore, much like browser version strings which "look like" (i.e. match widely used patterns for) prior browsers, it seems positioning color as a suffix has a clear intent behind it, so if someone wanted to add another component to the string, they'd do it in the middle. If I'm wrong, and a color terminal adds -color in the middle, then it'd probably better to deal with that case with real data available, so as to once again suitably narrow the pattern. In fact, the main worry I have about looking for `-color` anywhere in the string mirrors your question: What if the word color appears mid-string and it isn't a color terminal, because color is part of a larger word? For example, `tmux-colorless`. Then, you'd have two problems, because someone responding to that bug report would have to alter the algorithm, and not break anything else, which, based on just the code, without knowledge of the history, would involve either an explicit exclusion of the new value, or some other complication of the rule. I considered this scenario a low risk, but it illustrates the broader principle.

laanwj:
i think i don't care enough about this to keep it open, i thought it'd be useful but it's turning into too much arguing.

selsta:
@laanwj I would appreciate if you could you reopen this. It's indeed a useful patch and no more discussion is necessary before it would get merged.

iamamyth:
@selsta I considered submitting a patch for the upstream project but it seems inactive. I despise colorized logging but I've already done the work for this issue, so I'll open a PR with tests.

Issue #9785, opened 2025-02-08
Default priority has no mapping in get_fee_estimate


woodser:
Currently there is no mapping from `default` transaction priority to the fee estimates provided by [`get_fee_estimate`](https://docs.getmonero.org/rpc-library/monerod-rpc/#get_fee_estimate).

This prevents third parties from verifying that a transaction was generated with `default` priority. The best they can do is verify that a transaction was generated with `uimportant` priority.

This issue requests providing a such a mapping.

One way it could be done is by adding a new RPC call to monero-wallet-rpc. For example, `get_default_fee_priority()`, which returns 0-3, corresponding to unimportant, normal, elevated, and extra elevated.

Alternatively, the daemon could calculate and provide the mapping, but this would be a bigger change.

iamamyth:
Is the goal here to just help the client pick the right fee? If so, I'm not sure why you wouldn't add a field to the existing response, except, I suppose, the priorities and the default may have different change frequencies, but I suspect most clients will just call the endpoints in a pair, so splitting them makes net more work for the daemon.

woodser:
> I'm a bit confused as to who these "third parties" would be, as anyone relying on get_fee_estimate could easily be fooled by the new endpoint, too, as they'd don't have a "trusted" daemon. Or, if someone trusts the daemon, but not the client connecting to it, they can't ever confirm the client used the right daemon. Who does this help?

In our use case, peers submit signed transactions to an arbitrator for verification. For example, peers submit a "penalty" transaction which the arbitrator may broadcast if they misbehave during trade initialization. They also submit signed transactions to move funds to a shared multisig wallet.

The goal is for the arbitrator to verify that the peers used an appropriate mining fee at current congestion levels, to ensure protocol reliability.

Therefore, peers should use the "default" fee priority, and the arbitrator needs to be able to verify that the fee meets the current default level as well.

The problem is that the arbitrator has no way of knowing what the default fee level currently is, because it's not provided from their trusted daemon via `get_fee_estimate` or their own internal wallet.

iamamyth:
Based on your description, it seems my earlier suspicion would be correct:

> I suspect most clients will just call the endpoints in a pair, so splitting them makes net more work for the daemon.

Can you confirm that would be your usage pattern? If so, I think you should just add the default priority to the existing endpoint, for the reasons laid on in my earlier comment.

woodser:
> I think you should just add the default priority to the existing endpoint

What existing endpoint?

[...]

PR #9787, opened 2025-02-08
wallet: add wallet rpc call to get default fee priority

woodser:
Fixes https://github.com/monero-project/monero/issues/9785

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -4759,6 +4759,29 @@ namespace tools
     return true;
   }
   //------------------------------------------------------------------------------------------------------------------------------
+  bool wallet_rpc_server::on_get_default_fee_priority(const wallet_rpc::COMMAND_RPC_GET_DEFAULT_FEE_PRIORITY::request& req, wallet_rpc::COMMAND_RPC_GET_DEFAULT_FEE_PRIORITY::response& res, epee::json_rpc::error& er, const connection_context *ctx)
+  {
+    if (!m_wallet) return not_open(er);
+    try
+    {
+      uint32_t priority = m_wallet->adjust_priority(0);
+      if (priority == 0)
+      {
+        er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+        er.message = "Failed to get adjusted fee priority";
+        return false;
+      }
+      res.priority = priority;
+    }
+    catch (const std::exception& e)
+    {
+      er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+      er.message = "Failed to get adjusted fee priority";
+      return false;
+    }
+    return true;
+  }
+  //------------------------------------------------------------------------------------------------------------------------------
   bool wallet_rpc_server::on_get_version(const wallet_rpc::COMMAND_RPC_GET_VERSION::request& req, wallet_rpc::COMMAND_RPC_GET_VERSION::response& res, epee::json_rpc::error& er, const connection_context *ctx)
   {
     res.version = WALLET_RPC_VERSION;
--- a/src/wallet/wallet_rpc_server_commands_defs.h
+++ b/src/wallet/wallet_rpc_server_commands_defs.h
@@ -47,7 +47,7 @@
 // advance which version they will stop working with
 // Don't go over 32767 for any of these
 #define WALLET_RPC_VERSION_MAJOR 1
-#define WALLET_RPC_VERSION_MINOR 27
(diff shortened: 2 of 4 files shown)


selsta:
This fails to compile

woodser:
Fixed the compile error.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -4759,6 +4759,22 @@ namespace tools
     return true;
   }
   //------------------------------------------------------------------------------------------------------------------------------
+  bool wallet_rpc_server::on_get_default_fee_priority(const wallet_rpc::COMMAND_RPC_GET_DEFAULT_FEE_PRIORITY::request& req, wallet_rpc::COMMAND_RPC_GET_DEFAULT_FEE_PRIORITY::response& res, epee::json_rpc::error& er, const connection_context *ctx)
+  {
+    if (!m_wallet) return not_open(er);
+    try
+    {
+      res.priority = m_wallet->adjust_priority(0) - 1;
```
iamamyth, line 4767: Most code (including the function called here) reserves zero priority to mean "don't care"/"default", and distinguishes between the zero-based *index* of a priority in some array from the actual priority. So, if this field is called priority, it should probably return the one-based value, which you can map to an index on the client-side. Or, rename the field to priority_index, which I think would make its meaning equally clear.

nahuhh:
I'm getting a parse error with emprt param field:
```
{
  "jsonrpc": "2.0",
  "id": "0",
  "method": "get_default_fee_priority",
  "params": {}
}
```
```
{
  "error": {
    "code": -32700,
    "message": "Parse error"
  },
  "id": 0,
  "jsonrpc": "2.0"
}
```

And a method not found without specifying params field

```
{
  "error": {
    "code": -32601,
    "message": "Method not found"
  },
  "id": "0",
  "jsonrpc": "2.0"
}
```

This doesnt take any params right? And so should work like `get_version` (without a param)

src/wallet/wallet_rpc_server.cpp
woodser: Ok I changed it to use one-based priority.

woodser:
> This doesnt take any params right? And so should work like get_version (without a param)

I updated wallet.py so it should the same as `get_version` now afaict, but please let me know if you still see a problem.

iamamyth (approved):

0xFFFC0000 (approved):

src/wallet/wallet_rpc_server.cpp
```diff
@@ -4759,6 +4759,29 @@ namespace tools
     return true;
   }
   //------------------------------------------------------------------------------------------------------------------------------
+  bool wallet_rpc_server::on_get_default_fee_priority(const wallet_rpc::COMMAND_RPC_GET_DEFAULT_FEE_PRIORITY::request& req, wallet_rpc::COMMAND_RPC_GET_DEFAULT_FEE_PRIORITY::response& res, epee::json_rpc::error& er, const connection_context *ctx)
+  {
+    if (!m_wallet) return not_open(er);
+    try
+    {
+      int priority = m_wallet->adjust_priority(0);
```
selsta, line 4767:
```suggestion
      uint32_t priority = m_wallet->adjust_priority(0);
```
For consistency with the rest of the codebase.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -4759,6 +4759,29 @@ namespace tools
...
+        er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+        er.message = "Failed to get adjusted fee priority";
+        return false;
+      }
+      res.priority = priority;
+    }
+    catch (const std::exception& e)
+    {
+      er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+      er.message = "Failed get adjusted fee priority";
```
selsta, line 4779:
```suggestion
      er.message = "Failed to get adjusted fee priority";
```

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -2793,5 +2793,25 @@ namespace wallet_rpc
...
+    struct request_t
+    {
+      BEGIN_KV_SERIALIZE_MAP()
+      END_KV_SERIALIZE_MAP()
+    };
+    typedef epee::misc_utils::struct_init<request_t> request;
+
+    struct response_t
+    {
+      uint64_t priority;
```
selsta, line 2808:
```suggestion
      uint32_t priority;
```

Is there a reason for `uint64_t` here? Otherwise I would also adjust it.

selsta (approved):

PR #9800, opened 2025-02-14
Fix build with boost ASIO 1.87. Support boost 1.66+ [0.18]

vtnerd:
Same as #9628 but on release branch.

PR #9807, opened 2025-02-18
depends: boost: update to 1.69.0 [0.18]

tobtoht:
This should fix a reported crash during sync on Windows that was introduced in #9800.

Context starting here: https://libera.monerologs.net/monero-dev/20250217#c497165

Includes #9457 backport (needed).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1108,12 +1108,6 @@ add_definitions(-DBOOST_ASIO_ENABLE_SEQUENTIAL_STRAND_ALLOCATION)
 add_definitions(-DBOOST_NO_AUTO_PTR)
 add_definitions(-DBOOST_UUID_DISABLE_ALIGNMENT) # This restores UUID's std::has_unique_object_representations property
 
-if(FREEBSD AND DEPENDS)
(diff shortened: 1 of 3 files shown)


vtnerd:
Do we know what change in Boost caused/fixed the issue?

tobtoht:
For visibility, as this was already [discussed on irc](https://libera.monerologs.net/monero-dev/20250218#c499240). The commit that likely fixed the crash is: https://github.com/boostorg/asio/commit/ad8cc21d599f2f3622650896bff4fde3c7294caa

iamamyth (approved):

MrCyjaneK (approved):

tobtoht:
The presence of this bug in ASIO effectively raises the minimum Boost version to 1.69. Debian 10, which we [support](https://github.com/monero-project/monero/issues/9446) until at least 30 June 2025 ships Boost 1.67. When we drop support for Debian 10 the minimum becomes Boost 1.71 (Ubuntu 20.04).

If it's easy to work around this bug on our side it would be nice to do so in a follow-up PR. Since this only affects one development environment and development builds can use the latest `release-v0.18` branch, I don't think we need to get this in ahead of the upcoming tag.

PR #9809, opened 2025-02-20
logging: Fix easylogging++ init with blank config

iamamyth:
The upstream version of el::base::TypedConfigurations::unsafeGetConfigByRef accesses uninitialized memory if a key doesn't exist. Commit b2c59af84de8d35c1eee38878053206a62756968 patched the library to throw in this case, avoiding the invalid access, but the more suitable pattern, both logically (configure with a default-initialized config shouldn't throw), and as evidenced by the behavior of unsafeGetConfigByVal, would be to return a const reference to a default-initialized value with static storage duration.

### DIFF
--- a/external/easylogging++/easylogging++.h
+++ b/external/easylogging++/easylogging++.h
@@ -1993,7 +1993,7 @@ class TypedConfigurations : public base::threading::ThreadSafe {
   }
 
   template <typename Conf_T>
-  inline Conf_T& getConfigByRef(Level level, std::unordered_map<Level, Conf_T>* confMap, const char* confName) {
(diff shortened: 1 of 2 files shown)


Tzadiko:
Correct me if I'm wrong, but [easylogging++](https://github.com/abumq/easyloggingpp) is a third-party library (it's under the `external` folder). Would it not make more sense to apply this change to their (source) repository? If we were to ever upgrade easylogging++, this change wouldn't be included.

selsta:
The project is not actively developed anymore and we have a lot of our own changes applied.

Tzadiko:
> The project is not actively developed anymore and we have a lot of our own changes applied.

Got it, I saw the git-blame showed quite a bit of our own additions. Makes sense, thanks.

selsta (approved):

tobtoht (approved):

PR #9821, opened 2025-02-25
Fix HTTP unit tests (broken with new Boost versions)

vtnerd:
I noticed on two of my test boxes that the new HTTP tests seem to fail. I tracked this down to a cap on the body size, that seems to only be enforced by newer versions of Boost. Regardless, this PR seems to fix the tests on 1.85+.

If this PR passes all tests, I will PR the same fix to the release branch.


vtnerd:
The failed test should be completely unrelated.

0xFFFC0000:
[  FAILED  ] select_outputs.exact_unlock_block 

Is unrelated to PR.

selsta (approved):

tobtoht (approved):

PR #9826, opened 2025-03-08
crypto: add function sc_1()

jeffro256:
Part of upstreaming Carrot

### DIFF
--- a/src/crypto/crypto-ops.c
+++ b/src/crypto/crypto-ops.c
@@ -2429,6 +2429,14 @@ void sc_0(unsigned char *s) {
   }
 }
 
+void sc_1(unsigned char *s) {
+  int i;
+  s[0] = 1;
+  for (i = 1; i < 32; i++) {
+    s[i] = 0;
+  }
+}
+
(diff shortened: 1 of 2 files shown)


src/crypto/crypto-ops.c
```diff
@@ -2429,6 +2429,14 @@ void sc_0(unsigned char *s) {
   }
 }
 
+void sc_1(unsigned char *s) {
+  int i;
+  s[0] = 1;
```
Tzadiko, line 2434: I don't know anything about cryptography, but what's to stop someone from accidentally passing in an empty character array? Why not have `sc_1(unsigned char[32])` or `sc_1(std::array<unsigned char, 32>&)`?

jeffro256: Nothing, the convention is to assume a 32 byte buffer. A signature of `unsigned char[32]` might be better, I was just sticking with what was currently done.

UkoeHB (approved):

selsta (approved):

PR #9828, opened 2025-03-08
crypto: add Ed25519->X25519 conversion functions

jeffro256:
Part of upstreaming Carrot

Depends on #10111

### DIFF
--- a/src/crypto/crypto-ops.c
+++ b/src/crypto/crypto-ops.c
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2024, The Monero Project
+// Copyright (c) 2014-2026, The Monero Project
 // 
 // All rights reserved.
 // 
@@ -3903,6 +3903,39 @@ int sc_isnonzero(const unsigned char *s) {
     s[9] | s[10] | s[11] | s[12] | s[13] | s[14] | s[15] | s[16] | s[17] |
     s[18] | s[19] | s[20] | s[21] | s[22] | s[23] | s[24] | s[25] | s[26] |
     s[27] | s[28] | s[29] | s[30] | s[31]) - 1) >> 8) + 1;
+}
+
+static int edwardsYZ_to_x25519(unsigned char *xbytes, const fe Y, const fe Z) {
+  //! @see Section 4.1 of RFC 7748: https://www.rfc-editor.org/rfc/rfc7748.html#section-4.1
+  //
+  // y = Y/Z
+  // x_mont = (1 + y) / (1 - y)
+  //        = (1 + Y/Z) / (1 - Y/Z)
+  //        = (Z + Y) / (Z - Y)
+
+  fe tmp0;
+  fe tmp1;
+  int r;
+  fe_add(tmp0, Z, Y);       // Z + Y
+  fe_sub(tmp1, Z, Y);       // Z - Y
+  r = -!fe_isnonzero(tmp1); // succeed iff 0 != (Z - Y). AKA fail if identity point or some invalid reprs
+  fe_invert(tmp1, tmp1);    // 1/(Z - Y)
+  fe_mul(tmp0, tmp0, tmp1); // (Z + Y) / (Z - Y)
+  fe_tobytes(xbytes, tmp0); // tobytes((Z + Y) / (Z - Y))
+  return r;                 // 0 on success, otherwise -1
+}
+
+int ge_p3_to_x25519(unsigned char *xbytes, const ge_p3 *h) {
+  return edwardsYZ_to_x25519(xbytes, h->Y, h->Z);
+}
+
+int edwards_bytes_to_x25519_vartime(unsigned char *xbytes, const unsigned char *s) {
+  ge_p3 h;
+  const int r = ge_frombytes_vartime(&h, s);
+  if (0 != r)
+    return r;
+
+  return edwardsYZ_to_x25519(xbytes, h.Y, h.Z);
 }
 
 int ge_p3_is_point_at_infinity_vartime(const ge_p3 *p) {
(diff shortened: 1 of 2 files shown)


jeffro256:
Removed redundant code by depending on `fe_frombytes_vartime()` in #10111.

jeffro256:
Rebased on top of commit cb9903effedcb1442942112adae06cf98f947550, which is present in https://github.com/monero-project/monero/pull/10111, instead of whole PR.

src/crypto/crypto-ops.c
```diff
@@ -3829,6 +3835,41 @@ int sc_isnonzero(const unsigned char *s) {
...
+{
+  fe Y;
+  if (fe_frombytes_vartime(Y, s) != 0) {
+    return -1;
+  }
+
+  fe Z;
+  fe_1(Z);
+
+  edwardsYZ_to_x25519(xbytes, Y, Z);
```
vtnerd: So we "lose" a bit from the x-coordinate then? I recall this being how it worked last I checked. I guess it doesn't matter as all of the x25519 implementations ignore it anyway.

jeffro256: Yes, exactly. X25519 (and other Montgomery ladder curve implementations) choose to lose 1 bit of security in favor of the much faster point arithmetic.

jeffro256:
Rebased (to fix unrelated compilation issues), and implemented feedback by Cypherstack in section 5 of https://github.com/cypherstack/fcmp-impl-audit/blob/main/FCMP_Code_Implementation_Audit.pdf.

jeffro256:
Rebased and squashed

j-berman (approved):

UkoeHB:
What paper is this conversion from?

jeffro256:
@UkoeHB added a link to RFC 7748

src/crypto/crypto-ops.c
```diff
@@ -3904,6 +3905,39 @@ int sc_isnonzero(const unsigned char *s) {
...
+  // x_mont = (1 + y) / (1 - y)
+  //        = (1 + Y/Z) / (1 - Y/Z)
+  //        = (Z + Y) / (Z - Y)
+
+  fe tmp0;
+  fe tmp1;
+  int r;
+  fe_add(tmp0, Z, Y);       // Z + Y
+  fe_sub(tmp1, Z, Y);       // Z - Y
+  r = -!fe_isnonzero(tmp1); // 0 ?= (Z - Y) AKA is identity point?
```
UkoeHB: This is only identity if `X == 0` https://eprint.iacr.org/2008/522.pdf sections 2-3

jeffro256: Non-canonical encodings of the identity point (-0) should also be rejected, right? At any rate, I can't invert `tmp1`, regardless of what $X$ is.

UkoeHB: Would `-0` be a 'valid field element representation' according to the precondition (i.e. doesn't that imply reduced)?

Just objecting to the comment, not the logic. Unless it's the case that `Z == Y` is only true for identity among all valid points, which isn't immediately clear to me.

jeffro256: Ah I see, yeah I can update the comment

jeffro256: For all $Y,Z \neq 0$, $y = Y / Z = 1$, which would make it the identity point. $Z = 0, Y = 0$ is not a valid Ed25519 repr, but can happen in the code.

jeffro256: Okay done in latest

UkoeHB: > For all Y , Z ≠ 0 , y = Y / Z = 1 , which would make it the identity point.

Yeah still uncertain here. Is there a paper reference or algebraic argument that you can't have a valid non-identity point with `X != 0` and `y = Y / Z = 1`?

UkoeHB: I guess you could say it *must* be so because otherwise the birational map wouldn't work. Though that doesn't prove it directly.

Looking at the ed5519 curve, setting `y = 1`
```
ax^2 + y^2 = 1 + dx^2y^2
ax^2 + 1 = 1 + dx^2
ax^2 = dx^2
ax^2 - dx^2 = 0
x = 0
```
So `x` is always zero when `y = 1`.

src/crypto/crypto-ops.h
```diff
@@ -165,6 +162,30 @@ void sc_muladd(unsigned char *s, const unsigned char *a, const unsigned char *b,
...
+/**
+ * @brief Convert Ed25519 y-coord to X25519 x-coord, AKA "ConvertPointE()" in the Carrot spec
+ * @param[out] xbytes X25519 x-coord as a 255-bit little endian integer
+ * @param h Ed25519 point in projective representation
+ * @return 0 on success, otherwise non-0 on failure
+ *
+ * Returns failure on Ed25519's identity point. Assumes `h` has valid field element representations,
+ * that Z != 0, and that Y/Z is a valid y-coordinate for Ed25519. Returns success iff Y != Z, i.e.
+ * not an identity point. The runtime is constant.
+ */
```
UkoeHB, line 174:
```suggestion
 * Returns failure on Ed25519's identity point. Assumes `h` has valid field element representations,
 * that Z != 0, and that Y/Z is a valid y-coordinate for Ed25519. Returns success iff Y != Z (Y == Z
 * corresponds with the identity point in valid point representations). The runtime is constant.
 */
```
This should fix the ambiguity that's bugging me.

src/crypto/crypto-ops.c
jeffro256: Yes, exactly. We know that $Z = 0$ is an invalid representation by definition. All other cases lead to $y = 1$. Then subbing $y = 1$ gives you the chain of simplications you listed out. We just need to additionally make sure that $(a - d) \neq 0$, which is the case for Ed25519.

UkoeHB (approved):

j-berman (approved):

Issue #7731, opened 2021-05-20
Support multiple destinations in make_uri and parse_uri


woodser:
monero-wallet-rpc supports [make_uri](https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#make_uri) and [parse_uri](https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#parse_uri) which can be used to create a request for payment which the sender can parse to pay.

These calls currently only support a single destination address and amount.

This issue requests supporting multiple destination addresses and amounts to accommodate multi-output requests which can be used in a variety of applications or to prepare wallets with available outputs.

An encoding scheme is needed to support multiple addresses and amounts.  The format should be as small as possible to accommodate limited capacity of QR codes.

erciccione:
There is a **$500 bounty** on this issue. See https://github.com/haveno-dex/haveno/issues/83 for details.

benevanoff:
I’ll take a stab :p

woodser:
Example encoding scheme supported on IRC (originally proposed by @benevanoff):

`monero:46BeWrHpwXmHDpDEUmZBWZfoQpdc6HaERCNmx1pEYL2rAcuwufPN9rXHHtyUA4QVy66qeFQkn6sfK8aHYjA3jk3o1Bv16em,46BeWrHpwXmHDpDEUmZBWZfoQpdc6HaERCNmx1pEYL2rAcuwufPN9rXHHtyUA4QVy66qeFQkn6sfK8aHYjA3jk3o1Bv16em?amount=239.39014,123.55423&description=donation,example` (note description text needs escaped).

For backward compatibility, `make_uri` should produce the old format if using a single destination and address and this new format for multiple destinations and addresses.

trasherdk:
Positional parameters, description field optional.

`monero:<addr1>,piconero,base64(desc):<addr2>,piconero,base64(desc)`

benevanoff:
I like @trasherdk 's suggestion better

selsta:
@trasherdk's solution will use more characters overall because you will have e.g. "amount" multiple times.

trasherdk:
I would not use `?`, `amount`, `description` and `=`.

Using `:` as delimiter for separate transactions. `monero:<transaction1>:<transaction2>`
where a transaction is `<addr1>,piconero,base64(desc)` making `description` optional.

selsta:
[...]

PR #9830, opened 2025-03-09
Adding multiple-URI support in `parse_uri` and `make_uri`

U65535F:
A rewritten version of #9756 that is almost similar to it.
The previous source got deleted somehow (apologies).

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -6532,25 +6532,71 @@ bool simple_wallet::transfer_main(const std::vector<std::string> &args_, bool ca
     bool r = true;
 
     // check for a URI
-    std::string address_uri, payment_id_uri, tx_description, recipient_name, error;
+    std::string payment_id_uri, tx_description, error;
     std::vector<std::string> unknown_parameters;
-    uint64_t amount = 0;
-    bool has_uri = m_wallet->parse_uri(local_args[i], address_uri, payment_id_uri, amount, tx_description, recipient_name, unknown_parameters, error);
+    std::vector<std::string> addresses;
+    std::vector<uint64_t> amounts;
+    std::vector<std::string> recipient_names;
+    bool has_uri = m_wallet->parse_uri(local_args[i], addresses, amounts, recipient_names, payment_id_uri, tx_description, unknown_parameters, error);
     if (has_uri)
     {
-      r = cryptonote::get_account_address_from_str_or_url(info, m_wallet->nettype(), address_uri, oa_prompter);
-      if (payment_id_uri.size() == 16)
+      for (size_t j = 0; j < addresses.size(); j++)
       {
-        if (!tools::wallet2::parse_short_payment_id(payment_id_uri, info.payment_id))
+        r = cryptonote::get_account_address_from_str_or_url(info, m_wallet->nettype(), addresses[j], oa_prompter);
+        if (payment_id_uri.size() == 16)
         {
-          fail_msg_writer() << tr("failed to parse short payment ID from URI");
+          if (!tools::wallet2::parse_short_payment_id(payment_id_uri, info.payment_id))
+          {
+            fail_msg_writer() << tr("failed to parse short payment ID from URI");
+            return false;
+          }
+          info.has_payment_id = true;
+        }
+        de.amount = amounts[j];
+        de.original = addresses[j];
+        if (!r)
+        {
+          fail_msg_writer() << tr("failed to parse address");
           return false;
         }
-        info.has_payment_id = true;
+        de.addr = info.address;
+        de.is_subaddress = info.is_subaddress;
+        de.is_integrated = info.has_payment_id;
+
+        if (info.has_payment_id || !payment_id_uri.empty())
+        {
+          if (payment_id_seen)
+          {
+            fail_msg_writer() << tr("a single transaction cannot use more than one payment id");
+            return false;
+          }
+          crypto::hash payment_id;
+          std::string extra_nonce;
+          if (info.has_payment_id)
+          {
+            set_encrypted_payment_id_to_tx_extra_nonce(extra_nonce, info.payment_id);
+          }
+          else if (tools::wallet2::parse_payment_id(payment_id_uri, payment_id))
+          {
+            LONG_PAYMENT_ID_SUPPORT_CHECK();
+          }
+          else
+          {
+            fail_msg_writer() << tr("failed to parse payment id, though it was detected");
+            return false;
+          }
+          bool r = add_extra_nonce_to_tx_extra(extra, extra_nonce);
+          if(!r)
+          {
+            fail_msg_writer() << tr("failed to set up payment id, though it was decoded correctly");
+            return false;
+          }
+          payment_id_seen = true;
+        }
+        dsts.push_back(de);
       }
-      de.amount = amount;
-      de.original = local_args[i];
-      ++i;
+      i++;
+      break;
     }
     else if (i + 1 < local_args.size())
     {
--- a/src/wallet/api/wallet.h
+++ b/src/wallet/api/wallet.h
@@ -214,7 +214,9 @@ public:
     virtual void startRefresh() override;
     virtual void pauseRefresh() override;
     virtual bool parse_uri(const std::string &uri, std::string &address, std::string &payment_id, uint64_t &amount, std::string &tx_description, std::string &recipient_name, std::vector<std::string> &unknown_parameters, std::string &error) override;
+    bool parse_uri(const std::string &uri, std::vector<std::string> &addresses, std::vector<std::uint64_t> &amounts, std::vector<std::string> &recipient_names, std::string &payment_id, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error) override;
     virtual std::string make_uri(const std::string &address, const std::string &payment_id, uint64_t amount, const std::string &tx_description, const std::string &recipient_name, std::string &error) const override;
+    std::string make_uri(const std::vector<std::string> &addresses, const std::vector<std::uint64_t> &amounts, const std::vector<std::string> &recipient_names, const std::string &payment_id, const std::string &tx_description, std::string &error) const override;
     virtual std::string getDefaultDataDir() const override;
     virtual bool blackballOutputs(const std::vector<std::string> &outputs, bool add) override;
     virtual bool blackballOutput(const std::string &amount, const std::string &offset) override;
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -14942,6 +14942,34 @@ std::string wallet2::decrypt_with_view_secret_key(const std::string &ciphertext,
   return decrypt(ciphertext, get_account().get_keys().m_view_secret_key, authenticated);
 }
 //----------------------------------------------------------------------------------------------------
+std::string wallet2::custom_conver_to_url_format(const std::string &uri) const
+{
+  std::string s = epee::net_utils::conver_to_url_format(uri);
+
+  // replace '=' with "%3D" and '?' with "%3F".
+  std::string result;
+  result.reserve(s.size());
+
+  for (char c : s)
+  {
+    if (c == '=')
+    {
+      result.append("%3D");
+    }
+    else if (c == '?')
+    {
+      result.append("%3F");
+      }
+      
+    else
+    {
+      result.push_back(c);
+    }      
+  }
+
+  return result;
+}
+//----------------------------------------------------------------------------------------------------
 std::string wallet2::make_uri(const std::vector<std::string> &addresses, const std::vector<std::uint64_t> &amounts, const std::vector<std::string> &recipient_names, const std::string &payment_id, const std::string &tx_description, std::string &error) const
 {
   if (addresses.empty())
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -14942,56 +14942,108 @@ std::string wallet2::decrypt_with_view_secret_key(const std::string &ciphertext,
   return decrypt(ciphertext, get_account().get_keys().m_view_secret_key, authenticated);
 }
 //----------------------------------------------------------------------------------------------------
-std::string wallet2::make_uri(const std::string &address, const std::string &payment_id, uint64_t amount, const std::string &tx_description, const std::string &recipient_name, std::string &error) const
+std::string wallet2::make_uri(const std::vector<std::string> &addresses, const std::vector<std::uint64_t> &amounts, const std::vector<std::string> &recipient_names, const std::string &payment_id, const std::string &tx_description, std::string &error) const
 {
-  cryptonote::address_parse_info info;
-  if(!get_account_address_from_str(info, nettype(), address))
+  if (addresses.empty())
   {
-    error = std::string("wrong address: ") + address;
+    error = "No recipient data provided.";
+    return std::string();
+  }
+  
+  if (addresses.size() != amounts.size() || addresses.size() != recipient_names.size())
+  {
+    error = "Mismatch between the number of addresses, amounts, and recipient names.";
     return std::string();
   }
 
-  // we want only one payment id
-  if (info.has_payment_id && !payment_id.empty())
-  {
-    error = "A single payment id is allowed";
-    return std::string();
+  std::string addresses_str = "";
+  std::string amounts_str = "";
+  bool amounts_used = false;
+  std::string recipients_str = "";
+  bool recipients_used = false;
+
+  for (size_t i = 0; i < addresses.size(); i++) {
+    const std::string &address = addresses[i];
+    cryptonote::address_parse_info info;
+    if (!get_account_address_from_str(info, nettype(), address))
+    {
+      error = std::string("wrong address: ") + address;
+      return std::string();
+    }
+    if (info.has_payment_id && !payment_id.empty())
+    {
+      error = "Separate payment id given with an integrated address";
+      return std::string();
+    }
+    if (!addresses_str.empty())
+    {
+      addresses_str += ";";
+    }
+    addresses_str += address;
+
+    if (!amounts_str.empty())
+    {
+      amounts_str += ";";
+    }
+    if (amounts[i] > 0)
+    {
+      amounts_used = true;
+    }
+    amounts_str += cryptonote::print_money(amounts[i]);
+    
+    if (!recipients_str.empty())
+    {
+      recipients_str += ";";
+    }
+    if (!recipient_names[i].empty())
+    {
+      recipients_used = true;
+      recipients_str += custom_conver_to_url_format(recipient_names[i]);
+    }
   }
 
-  if (!payment_id.empty())
-  {
-    error = "Standalone payment id deprecated, use integrated address instead";
-    return std::string();
-  }
-
-  std::string uri = "monero:" + address;
+  std::string uri = "monero:" + addresses_str;
   unsigned int n_fields = 0;
 
-  if (!payment_id.empty())
-  {
-    uri += (n_fields++ ? "&" : "?") + std::string("tx_payment_id=") + payment_id;
-  }
-
-  if (amount > 0)
+  if (amounts_used)
   {
     // URI encoded amount is in decimal units, not atomic units
-    uri += (n_fields++ ? "&" : "?") + std::string("tx_amount=") + cryptonote::print_money(amount);
+    uri += (n_fields++ ? "&" : "?") + std::string("tx_amount=") + amounts_str;
   }
 
-  if (!recipient_name.empty())
+  if (recipients_used)
   {
-    uri += (n_fields++ ? "&" : "?") + std::string("recipient_name=") + epee::net_utils::conver_to_url_format(recipient_name);
+    uri += (n_fields++ ? "&" : "?") + std::string("recipient_name=") + recipients_str;
   }
 
   if (!tx_description.empty())
   {
-    uri += (n_fields++ ? "&" : "?") + std::string("tx_description=") + epee::net_utils::conver_to_url_format(tx_description);
+    uri += (n_fields++ ? "&" : "?") + std::string("tx_description=") + custom_conver_to_url_format(tx_description);
+  }
+
+  if (!payment_id.empty())
+  {
+    crypto::hash pid32;
+    if (!wallet2::parse_long_payment_id(payment_id, pid32))
+    {
+      error = "Invalid payment id";
+      return std::string();
+    }
+    uri += (n_fields++ ? "&" : "?") + std::string("tx_payment_id=") + payment_id;
   }
 
   return uri;
 }
 //----------------------------------------------------------------------------------------------------
-bool wallet2::parse_uri(const std::string &uri, std::string &address, std::string &payment_id, uint64_t &amount, std::string &tx_description, std::string &recipient_name, std::vector<std::string> &unknown_parameters, std::string &error)
+std::string wallet2::make_uri(const std::string &address, const std::string &payment_id, uint64_t amount, const std::string &tx_description, const std::string &recipient_name, std::string &error) const
+{
+  std::vector<std::string> addresses { address };
+  std::vector<std::uint64_t> amounts { amount };
+  std::vector<std::string> recipient_names { recipient_name};
+  return make_uri(addresses, amounts, recipient_names, payment_id, tx_description, error);
+}
+//----------------------------------------------------------------------------------------------------
+bool wallet2::parse_uri(const std::string &uri, std::vector<std::string> &addresses, std::vector<std::uint64_t> &amounts, std::vector<std::string> &recipient_names, std::string &payment_id, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error)
 {
   if (uri.substr(0, 7) != "monero:")
   {
@@ -15001,24 +15053,37 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
 
   std::string remainder = uri.substr(7);
   const char *ptr = strchr(remainder.c_str(), '?');
-  address = ptr ? remainder.substr(0, ptr-remainder.c_str()) : remainder;
+  std::string addresses_string = ptr ? remainder.substr(0, ptr-remainder.c_str()) : remainder;
+  boost::split(addresses, addresses_string, boost::is_any_of(";"));
+  addresses.erase(std::remove(addresses.begin(), addresses.end(), ""), addresses.end());
 
-  cryptonote::address_parse_info info;
-  if(!get_account_address_from_str(info, nettype(), address))
+  if (addresses.empty())
   {
-    error = std::string("URI has wrong address: ") + address;
+    error = "No addresses specified in URI.";
     return false;
   }
-  if (!strchr(remainder.c_str(), '?'))
+
+  for (const std::string &address : addresses)
+  {
+    cryptonote::address_parse_info info;
+    if (!get_account_address_from_str(info, nettype(), address))
+    {
+      error = std::string("URI contains improper address: ") + address;
+      return false;
+    }
+  }
+
+  amounts.assign(addresses.size(), 0);
+  recipient_names.assign(addresses.size(), "");
+
+  if (ptr == NULL)
     return true;
 
+  std::string params(ptr+1);
   std::vector<std::string> arguments;
-  std::string body = remainder.substr(address.size() + 1);
-  if (body.empty())
-    return true;
-  boost::split(arguments, body, boost::is_any_of("&"));
+  boost::split(arguments, params, boost::is_any_of("&"));
   std::set<std::string> have_arg;
-  for (const auto &arg: arguments)
+  for (const std::string &arg : arguments)
   {
     std::vector<std::string> kv;
     boost::split(kv, arg, boost::is_any_of("="));
@@ -15036,20 +15101,29 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
 
     if (kv[0] == "tx_amount")
     {
-      amount = 0;
-      if (!cryptonote::parse_amount(amount, kv[1]))
+      std::vector<std::string> amounts_split;
+      boost::split(amounts_split, kv[1], boost::is_any_of(";"));
+      
+      if (amounts_split.size() != addresses.size())
       {
-        error = std::string("URI has invalid amount: ") + kv[1];
+        error = "Incorrect tx_amount count";
         return false;
+      }
+
+      for (size_t i = 0; i < amounts_split.size(); i++)
+      {
+        uint64_t amount;
+        if (!cryptonote::parse_amount(amount, amounts_split[i]))
+        {
+          error = std::string("URI has invalid amount: ") + amounts_split[i];
+          return false;
+        }
+        amounts[i] = amount;
       }
     }
     else if (kv[0] == "tx_payment_id")
     {
-      if (info.has_payment_id)
-      {
-        error = "Separate payment id given with an integrated address";
-        return false;
-      }
+      // standalone payment ids are deprecated. use integrated address
       crypto::hash hash;
       if (!wallet2::parse_long_payment_id(kv[1], hash))
       {
@@ -15057,10 +15131,27 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
         return false;
       }
       payment_id = kv[1];
+
+      if (payment_id.length() != 16 && payment_id.length() != 64)
+      {
+        error = "Invalid payment ID length";
+        return false;
+      }
     }
     else if (kv[0] == "recipient_name")
     {
-      recipient_name = epee::net_utils::convert_from_url_format(kv[1]);
+      std::vector<std::string> names_split;
+      boost::split(names_split, kv[1], boost::is_any_of(";"));
+      if (names_split.size() != addresses.size())
+      {
+        error = "Incorrect recipient_name count";
+        return false;
+      }
+
+      for (size_t i = 0; i < names_split.size(); i++)
+      {
+        recipient_names[i] = epee::net_utils::convert_from_url_format(names_split[i]);
+      }
     }
     else if (kv[0] == "tx_description")
     {
@@ -15071,6 +15162,40 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
       unknown_parameters.push_back(arg);
     }
   }
+
+  if (!recipient_names.empty() && recipient_names.size() != addresses.size())
+  {
+    error = "Incorrect recipient name count. Recipient name count must match address count";
+    return false;
+  }
+  if (!amounts.empty() && amounts.size() != addresses.size())
+  {
+    error = "Incorrect amount count. Amount count must match address count. Zero may be use as a filler";
+    return false;
+  }
+  return true;
+}
+//----------------------------------------------------------------------------------------------------
+bool wallet2::parse_uri(const std::string& uri, std::string& address, std::string& payment_id, uint64_t& amount, std::string& tx_description, std::string& recipient_name, std::vector<std::string>& unknown_parameters, std::string& error)
+{
+  std::vector<std::string> addresses;
+  std::vector<std::uint64_t> amounts;
+  std::vector<std::string> recipient_names;
+  
+  if (!parse_uri(uri, addresses, amounts, recipient_names, payment_id, tx_description, unknown_parameters, error))
+  {
+    error = "Failed to parse uri";
+    return false;
+  }
+  if (addresses.size() > 1)
+  {
+    error = "Multi-recipient URIs currently unsupported in this overload";
+    return false;
+  }
+  address = addresses[0];
+  amount = amounts[0];
+  recipient_name = recipient_names[0];
+
   return true;
 }
 //----------------------------------------------------------------------------------------------------
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -1641,8 +1641,11 @@ private:
     template<typename T=std::string> T decrypt(const std::string &ciphertext, const crypto::secret_key &skey, bool authenticated = true) const;
     std::string decrypt_with_view_secret_key(const std::string &ciphertext, bool authenticated = true) const;
 
+    std::string custom_conver_to_url_format(const std::string &uri) const;
+    std::string make_uri(const std::vector<std::string> &addresses, const std::vector<std::uint64_t> &amounts, const std::vector<std::string> &recipient_names, const std::string &payment_id, const std::string &tx_description, std::string &error) const;
     std::string make_uri(const std::string &address, const std::string &payment_id, uint64_t amount, const std::string &tx_description, const std::string &recipient_name, std::string &error) const;
     bool parse_uri(const std::string &uri, std::string &address, std::string &payment_id, uint64_t &amount, std::string &tx_description, std::string &recipient_name, std::vector<std::string> &unknown_parameters, std::string &error);
+    bool parse_uri(const std::string &uri, std::vector<std::string> &addresses, std::vector<std::uint64_t> &amounts, std::vector<std::string> &recipient_names, std::string &payment_id, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error);
 
     uint64_t get_blockchain_height_by_date(uint16_t year, uint8_t month, uint8_t day);    // 1<=month<=12, 1<=day<=31
 
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -3142,6 +3142,22 @@ namespace tools
     return true;
   }
   //------------------------------------------------------------------------------------------------------------------------------
+  bool wallet_rpc_server::on_make_uri_v2(const wallet_rpc::COMMAND_RPC_MAKE_URI_V2::request& req, wallet_rpc::COMMAND_RPC_MAKE_URI_V2::response& res, epee::json_rpc::error& er, const connection_context *ctx)
+  {
+    if (!m_wallet) return not_open(er);
+    std::string error;
+    std::string uri = m_wallet->make_uri(req.addresses, req.amounts, req.recipient_names, req.payment_id, req.tx_description, error);
+    if (uri.empty())
+    {
+      er.code = WALLET_RPC_ERROR_CODE_WRONG_URI;
+      er.message = std::string("Cannot make URI from supplied parameters: ") + error;
+      return false;
+    }
+    
+    res.uri = uri;
+    return true;
+  }
+  //------------------------------------------------------------------------------------------------------------------------------
   bool wallet_rpc_server::on_parse_uri(const wallet_rpc::COMMAND_RPC_PARSE_URI::request& req, wallet_rpc::COMMAND_RPC_PARSE_URI::response& res, epee::json_rpc::error& er, const connection_context *ctx)
   {
     if (!m_wallet) return not_open(er);
@@ -3152,6 +3168,26 @@ namespace tools
       er.message = "Error parsing URI: " + error;
       return false;
     }
+    return true;
+  }
+  //------------------------------------------------------------------------------------------------------------------------------
+  bool wallet_rpc_server::on_parse_uri_v2(const wallet_rpc::COMMAND_RPC_PARSE_URI_V2::request& req, wallet_rpc::COMMAND_RPC_PARSE_URI_V2::response& res, epee::json_rpc::error& er, const connection_context *ctx)
+  {
+    if (!m_wallet) return not_open(er);
+    std::string error;
+    std::vector<std::string> addresses;
+    std::vector<std::uint64_t> amounts;
+    std::vector<std::string> recipient_names;
+    if (!m_wallet->parse_uri(req.uri, addresses, amounts, recipient_names, res.uri.payment_id, res.uri.tx_description, res.unknown_parameters, error))
+    {
+      er.code = WALLET_RPC_ERROR_CODE_WRONG_URI;
+      er.message = "Error parsing URI: " + error;
+      return false;
+    }
+
+    res.uri.addresses = addresses;
+    res.uri.amounts = amounts;
+    res.uri.recipient_names = recipient_names;
     return true;
   }
   //------------------------------------------------------------------------------------------------------------------------------
--- a/src/wallet/wallet_rpc_server_commands_defs.h
+++ b/src/wallet/wallet_rpc_server_commands_defs.h
@@ -1829,7 +1829,7 @@ namespace wallet_rpc
   {
     std::string address;
     std::string payment_id;
-    uint64_t amount;
+    std::uint64_t amount;
     std::string tx_description;
     std::string recipient_name;
 
@@ -1875,6 +1875,78 @@ namespace wallet_rpc
     struct response_t
     {
       uri_spec uri;
+      std::vector<std::string> unknown_parameters;
+
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE(uri)
+        KV_SERIALIZE(unknown_parameters)
+      END_KV_SERIALIZE_MAP()
+    };
+    typedef epee::misc_utils::struct_init<response_t> response;
+  };
+
+  struct uri_spec_v2
+  {
+    std::vector<std::string> addresses;
+    std::vector<std::uint64_t> amounts;
+    std::vector<std::string> recipient_names;
+    std::string tx_description;
+    std::string payment_id;
+
+    BEGIN_KV_SERIALIZE_MAP()
+      KV_SERIALIZE(addresses);
+      KV_SERIALIZE(amounts);
+      KV_SERIALIZE(recipient_names);
+      KV_SERIALIZE(tx_description);
+      KV_SERIALIZE(payment_id);
+    END_KV_SERIALIZE_MAP()
+  };
+
+  struct COMMAND_RPC_MAKE_URI_V2
+  {
+    struct request_t
+    {
+      std::vector<std::string> addresses;
+      std::vector<std::uint64_t> amounts;
+      std::vector<std::string> recipient_names;
+      std::string tx_description;
+      std::string payment_id;
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE(addresses);
+        KV_SERIALIZE(amounts);
+        KV_SERIALIZE(recipient_names);
+        KV_SERIALIZE(tx_description);
+        KV_SERIALIZE(payment_id);
+      END_KV_SERIALIZE_MAP()
+    };
+    typedef epee::misc_utils::struct_init<request_t> request;
+
+    struct response_t
+    {
+      std::string uri;
+
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE(uri)
+      END_KV_SERIALIZE_MAP()
+    };
+    typedef epee::misc_utils::struct_init<response_t> response;
+  };
+
+  struct COMMAND_RPC_PARSE_URI_V2
+  {
+    struct request_t
+    {
+      std::string uri;
+
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE(uri)
+      END_KV_SERIALIZE_MAP()
+    };
+    typedef epee::misc_utils::struct_init<request_t> request;
+
+    struct response_t
+    {
+      uri_spec_v2 uri;
       std::vector<std::string> unknown_parameters;
 
       BEGIN_KV_SERIALIZE_MAP()
--- a/tests/functional_tests/uri.py
+++ b/tests/functional_tests/uri.py
@@ -43,6 +43,7 @@ class URITest():
     def run_test(self):
       self.create()
       self.test_monero_uri()
+      self.test_multi_uri()
 
     def create(self):
         print('Creating wallet')
@@ -56,7 +57,7 @@ class URITest():
         assert res.seed == seed
 
     def test_monero_uri(self):
-        print('Testing monero: URI')
+        print('Testing monero: URI - single')
         wallet = Wallet()
 
         utf8string = [u'えんしゅう', u'あまやかす']
@@ -141,11 +142,17 @@ class URITest():
         assert res.uri.recipient_name == utf8string[0]
         assert not 'unknown_parameters' in res or len(res.unknown_parameters) == 0
 
-        # external payment ids are not supported anymore
-        ok = False
-        try: res = wallet.make_uri(address = address, recipient_name = utf8string[0], tx_description = utf8string[1], amount = 1000000000000, payment_id = '1' * 64)
-        except: ok = True
-        assert ok
+        # standalone payment ids are supported but not recommended
+        res = wallet.make_uri(address = address, recipient_name = utf8string[0], tx_description = utf8string[1], amount = 1000000000000, payment_id = '1' * 64)
+        assert res.uri == 'monero:' + address + '?tx_amount=1.000000000000&recipient_name=' + quoted_utf8string[0] + '&tx_description=' + quoted_utf8string[1] + '&tx_payment_id=' + '1' * 64
+        
+        # in case of standalone payment id removal
+        # ok = False
+        # try:
+        #     res = wallet.make_uri(address = address, recipient_name = utf8string[0], tx_description = utf8string[1], amount = 1000000000000, payment_id = '1' * 64)
+        # except:
+        #     ok = True
+        # assert ok
 
         # spaces must be encoded as %20
         res = wallet.make_uri(address = address, tx_description = ' ' + utf8string[1] + ' ' + utf8string[0] + ' ', amount = 1000000000000)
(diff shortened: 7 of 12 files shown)


U65535F:
Tests are passing though.

src/wallet/wallet2.cpp
```diff
@@ -15071,6 +15190,40 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
...
+  std::vector<std::string> recipient_names;
+  
+  if (!parse_uri(uri, addresses, amounts, recipient_names, payment_id, tx_description, unknown_parameters, error))
+  {
+    error = "Failed to parse uri";
+    return false;
+  }
+  if (addresses.size() > 1)
+  {
+    error = "Multi-recipient URIs currently unsupported in this overload";
```
Tzadiko, line 15220: This is moreso a message for a developer, instead of a message for a user. The user (especially a non-technical one) will not have any idea what this (an overload) means. Consider rephrasing.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3142,6 +3142,22 @@ namespace tools
     return true;
   }
   //------------------------------------------------------------------------------------------------------------------------------
+  bool wallet_rpc_server::on_make_uri_v2(const wallet_rpc::COMMAND_RPC_MAKE_URI_V2::request& req, wallet_rpc::COMMAND_RPC_MAKE_URI_V2::response& res, epee::json_rpc::error& er, const connection_context *ctx)
```
Tzadiko, line 3145: I'm personally not a fan of the naming `_v2`. If a new developer comes and reads this, they will have no idea which one to use (v1 or v2). V2 may imply that V1 is no longer used, or that V2 is 'better' than V1. Consider renaming this to `on_make_uri_multiple` or something similar. You'll need to make this change to similar methods like `on_parse_uri_v2` and lines like these: https://github.com/monero-project/monero/pull/9830/files#diff-ffc587ad00e4573484c0430da9e42b7690f664de49df6f52b1a5be9b48833605R129

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -1829,7 +1829,7 @@ namespace wallet_rpc
   {
     std::string address;
     std::string payment_id;
-    uint64_t amount;
+    std::uint64_t amount;
```
Tzadiko, line 1832: Why std::? I haven't seen this on anything other than MSVC, which isn't supported by Monero.

tests/functional_tests/uri.py
```diff
@@ -223,7 +231,92 @@ def test_monero_uri(self):
...
+        assert parsed.uri.amounts[0] == 500000000000
+        assert parsed.uri.recipient_names[0] == utf8string[0]
+        assert parsed.uri.addresses[1] == addr2
+        assert parsed.uri.amounts[1] == 200000000000
+        assert parsed.uri.recipient_names[1] == utf8string[1]
+        # check tx_description at the top level.
+        assert parsed.uri.tx_description == 'multi test'
+        assert parsed.uri.payment_id == ''
+
+        # build multi-recipient URI with three payments.
```
Tzadiko, line 263: Can you make this into multiple tests? Youre testing quite a bit here, and each comment before the assert suggests something else is being tested. You can create a test fixture and use it to pass data that's common to each test to.

src/wallet/wallet2.cpp
U65535F: What should I rephrase it as? That is mistake of the developer right?

src/wallet/wallet_rpc_server.cpp
U65535F: Yes. The developer doesn't know if he gets a muti recipient URI. So, the v1 function fails. v2 can process both URIs.  Take a look at old PR.

tests/functional_tests/uri.py
U65535F: Please elaborate. I don't understand.

Tzadiko: Sure, so you have six groupings of asserts in one test method (test_multi_uri). You have lines 252 to 261, lines 269 to 279, lines 290, lines 296, lines 310 to 311, and lines 316 to 319. Each of these groupings as a comment marked "Case: ...". So clearly there are 7 cases being tested. Each case should be its own test with a unique test name. Otherwise, if this massive test fails, it will be hard to determine which part failed (despite a line number being associated with the failure message). 

If some of these tests share data, build a test fixture that contains said data and use it accross multiple tests to avoid code duplication. More on test fixtures here: https://docs.pytest.org/en/6.2.x/fixture.html

src/wallet/wallet2.cpp
Tzadiko: Im not sure if you'll ever hit this condition, unless there's a bug. Nonetheless I suggest you change it to something like "Multiple recipient addresses parsed out of transaction intended for one recipient.". You're moreso telling the user that there is a bug in your parse_uri moreso than there is an error that they themselves can fix. Hence, Im not sure how valuable this check even is. I want someone else to chime in.

src/wallet/wallet_rpc_server.cpp
U65535F: @Tzadiko

src/wallet/wallet2.cpp
```diff
@@ -14942,56 +14942,136 @@ std::string wallet2::decrypt_with_view_secret_key(const std::string &ciphertext,
   return decrypt(ciphertext, get_account().get_keys().m_view_secret_key, authenticated);
 }
 //----------------------------------------------------------------------------------------------------
-std::string wallet2::make_uri(const std::string &address, const std::string &payment_id, uint64_t amount, const std::string &tx_description, const std::string &recipient_name, std::string &error) const
+std::string wallet2::custom_conver_to_url_format(const std::string &uri) const
```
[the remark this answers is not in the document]
U65535F: I don't think that's any wrong there because there a function known as `epee::net_utils::conver_to_url_format` also exists.

src/wallet/api/wallet.h
```diff
@@ -214,7 +214,9 @@ class WalletImpl : public Wallet
     virtual void startRefresh() override;
     virtual void pauseRefresh() override;
     virtual bool parse_uri(const std::string &uri, std::string &address, std::string &payment_id, uint64_t &amount, std::string &tx_description, std::string &recipient_name, std::vector<std::string> &unknown_parameters, std::string &error) override;
+    bool parse_uri(const std::string &uri, std::vector<std::string> &addresses, std::vector<std::uint64_t> &amounts, std::vector<std::string> &recipient_names, std::string &payment_id, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error) override;
```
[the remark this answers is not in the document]
U65535F: Yes, exactly.

src/wallet/wallet2.cpp
```diff
@@ -15036,31 +15129,57 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
...
+      }
     }
     else if (kv[0] == "recipient_name")
     {
-      recipient_name = epee::net_utils::convert_from_url_format(kv[1]);
+      std::vector<std::string> names_split;
+      boost::split(names_split, kv[1], boost::is_any_of(";"));
+      if (names_split.size() != addresses.size())
+      {
+        error = "Incorrect recipient_name count";
```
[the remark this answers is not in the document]
U65535F: Can you name all of them please so I can fix them?

src/wallet/wallet_rpc_server_commands_defs.h
U65535F: Suggested by SNeedleWoods at #9756

tests/functional_tests/uri.py
U65535F: OK. So you mean to split the `test_multi_uri` function into multiple functions, because it is vast and combines multiple test cases into one making it complex? 
Yes, those tests also share some variables. So you mean to pass the variables via test fixtures in order to avoid code redundancy. 

> Otherwise, if this massive test fails, it will be hard to determine which part failed (despite a line number being associated with the failure message).

It's easy, at least for me (as far I've seen) to determine which part fails, due to Python giving the line number. Not sure what you meant by `hard` but the suggestions seems to be good, helping the code to be more readable.

src/wallet/wallet_rpc_server_commands_defs.h
U65535F: image
@Tzadiko

src/wallet/api/wallet.h
```diff
@@ -214,7 +214,9 @@ class WalletImpl : public Wallet
...
     virtual std::string make_uri(const std::string &address, const std::string &payment_id, uint64_t amount, const std::string &tx_description, const std::string &recipient_name, std::string &error) const override;
+    std::string make_uri(const std::vector<std::string> &addresses, const std::vector<std::uint64_t> &amounts, const std::vector<std::string> &recipient_names, const std::string &payment_id, const std::string &tx_description, std::string &error) const override;
```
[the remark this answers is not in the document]
U65535F: image
SNeedlewoods (see #9756) said that they don't want to have "non-standard" types as arguments for methods in the Wallet API. FYI, we had a type for simplifying things as you said but later to the comments, I removed it and modified the code so it only accepts standard types.

src/simplewallet/simplewallet.cpp
```diff
@@ -6532,25 +6532,71 @@ bool simple_wallet::transfer_main(const std::vector<std::string> &args_, bool ca
...
+        {
+          fail_msg_writer() << tr("failed to parse address");
           return false;
         }
-        info.has_payment_id = true;
+        de.addr = info.address;
+        de.is_subaddress = info.is_subaddress;
+        de.is_integrated = info.has_payment_id;
+
+        if (info.has_payment_id || !payment_id_uri.empty())
```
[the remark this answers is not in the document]
U65535F: I don't really know to be exact. But I think, yes it could be possible. Like wise, we have so many checks in the code which hardly hit.

U65535F:
@Tzadiko

src/wallet/wallet_rpc_server_commands_defs.h
Tzadiko: In practice, many codebases use uint32_t without the std:: prefix for brevity, especially when working with C-compatible code or when the project includes <stdint.h>. They usually use std::uint32_t when they include <cstdint>. Just use uint32_t.

src/wallet/wallet2.cpp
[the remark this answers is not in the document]
Tzadiko: Just because net_utils has a mispelled function doesn't mean you should double down on the misspelling.

tests/functional_tests/uri.py
Tzadiko: > So you mean to split the test_multi_uri function into multiple functions, because it is vast and combines multiple test cases into one making it complex?

Yes

> So you mean to pass the variables via test fixtures in order to avoid code redundancy.

Yes

> It's easy, at least for me (as far I've seen) to determine which part fails, due to Python giving the line number.

Let me share an example, I was looking at functional_tests_rpc.py yesterday. It calles a transfer test in transfer.py. transfer.py's `transfer` test is massive and failed on line number 177. Naturally I looked at the function call closest to this line failure to see what the issue was. I ended up spending 20 minutes looking at how this function call may have been affected by my changeset only to then notice the variable that was being asserted was from a call 20+ lines before this assert, that had several other asserts before it. So yes, while you may have the line number, it greatly reduces confusion and the need to navigate around the test when there is one thing under test per test, instead of many things.

src/wallet/wallet2.cpp
```diff
@@ -14942,56 +14942,136 @@ std::string wallet2::decrypt_with_view_secret_key(const std::string &ciphertext,
...
+    if (info.has_payment_id && !payment_id.empty())
+    {
+      error = "Separate payment id given with an integrated address";
+      return std::string();
+    }
+    if (!addresses_str.empty())
+    {
+      addresses_str += ";";
+    }
+    addresses_str += address;
```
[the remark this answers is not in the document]
Tzadiko: Why was this marked as resolved without being addressed?

src/wallet/wallet2.cpp
```diff
@@ -15001,24 +15081,37 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
...
+    {
+      error = std::string("URI contains improper address: ") + address;
+      return false;
+    }
+  }
+
+  amounts.assign(addresses.size(), 0);
+  recipient_names.assign(addresses.size(), "");
+
+  if (ptr == NULL)
```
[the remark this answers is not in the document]
Tzadiko: Why was this marked as resolved without being addressed?

src/wallet/wallet2.cpp
```diff
@@ -15036,31 +15129,57 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
...
+      // standalone payment ids are deprecated. use integrated address
       crypto::hash hash;
       if (!wallet2::parse_long_payment_id(kv[1], hash))
       {
         error = "Invalid payment id: " + kv[1];
         return false;
       }
       payment_id = kv[1];
+
+      if (payment_id.length() != 16 && payment_id.length() != 64)
```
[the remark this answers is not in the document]
Tzadiko: Why was this marked as resolved without being addressed?

src/wallet/api/wallet.h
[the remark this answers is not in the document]
U65535F: @Tzadiko

src/wallet/wallet2.cpp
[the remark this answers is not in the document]
Tzadiko: I think it's a bit of common sense. Review your changed error messages, anything that doesn't explicitly tell you what is wrong / is unactionable is a bit worthless. For example, `Incorrect tx_amount count` doesn't tell you what's wrong with the transaction count (that the sizes and amounts are not equal). Something like `tx_amount count of X is not equal to address count of Y` (fill in X and Y).

src/wallet/wallet2.cpp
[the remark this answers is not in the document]
U65535F: Alright. I will fix it. I actually thought that `conver` was the real usage here. 
I wonder why net_utils hasn't fixed that yet, who knows.

src/wallet/wallet_rpc_server_commands_defs.h
U65535F: Alright 👍.

U65535F:
> > @Tzadiko
> 
> One tag is fine to prod me in case I missed the emails sent by Github. No need to tag me on each of my comments 👍

sorry, I thought I had to ping you in each conversation. Apologies.

src/wallet/wallet2.cpp
```diff
@@ -15071,6 +15190,40 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
...
   }
+
+  if (!recipient_names.empty() && recipient_names.size() != addresses.size())
+  {
+    error = "Incorrect recipient name count. Recipient name count must match address count";
+    return false;
+  }
+  if (!amounts.empty() && amounts.size() != addresses.size())
+  {
+    error = "Incorrect amount count. Amount count must match address count. Zero may be use as a filler";
```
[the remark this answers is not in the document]
Tzadiko: Why was this marked as resolved without being addressed? Please only mark as resolved after you've pushed changes.

src/wallet/wallet2.cpp
Tzadiko: Why was this marked as resolved without being addressed? Please only mark as resolved after you've pushed changes.

Tzadiko:
You resolved a bunch of comments without them being addressed. I suggest we either come to a conclusion before they're resolved, or you make a change and push a commit. For example, after discussing I agree no further change is needed [here](https://github.com/monero-project/monero/pull/9830#discussion_r2001498757), so it can be resolved.

src/wallet/wallet2.cpp
```diff
@@ -14974,13 +14974,13 @@ std::string wallet2::make_uri(const std::vector<std::string> &addresses, const s
...
   {
-    error = "No recipient data provided.";
+    error = "No recipient information like addresses are provided.";
     return std::string();
   }
   
   if (addresses.size() != amounts.size() || addresses.size() != recipient_names.size())
   {
-    error = "Mismatch between the number of addresses, amounts, and recipient names.";
+    error = std::string("The counts of addresses (" + addresses.size() + "), amounts (" + amounts.size() "), and recipient names (" + recipient_names.size() + ") do not match.");
```
Tzadiko: This doesn't output what you think it does. This is doing `const char*` + `size_t` which is dangerous. You need to do:

`std::string{"The counts of addresses ("} + std::to_string(addresses.size())` ...

Or use boost::format / similar string formatting. See example as to why your code doesn't work: https://godbolt.org/z/1axqzrEdK

src/wallet/wallet2.cpp
```diff
@@ -15134,7 +15134,7 @@ bool wallet2::parse_uri(const std::string &uri, std::vector<std::string> &addres
       
       if (amounts_split.size() != addresses.size())
       {
-        error = "Incorrect tx_amount count";
+        error = std::string("tx_amount count of" + amounts_split.size() + "is not equal to address count of " + addresses.size());
```
Tzadiko: Issue here too, same as previously mentioned.

src/wallet/wallet2.cpp
```diff
@@ -15191,14 +15193,9 @@ bool wallet2::parse_uri(const std::string &uri, std::vector<std::string> &addres
...
 
-  if (!recipient_names.empty() && recipient_names.size() != addresses.size())
-  {
-    error = "Incorrect recipient name count. Recipient name count must match address count";
-    return false;
-  }
   if (!amounts.empty() && amounts.size() != addresses.size())
   {
-    error = "Incorrect amount count. Amount count must match address count. Zero may be use as a filler";
+    error = "Amount count does not match address count. Please use zero values where you don't want to specify the amounts.";
```
Tzadiko, line 15198: I think you mean "Please don't specify amounts if ..." instead of "Please use zero values". Although I'm not personally sure what not specifying amounts does.

tests/functional_tests/uri.py
```diff
@@ -38,12 +39,18 @@
   from urllib.parse import quote as urllib_quote
 
 from framework.wallet import Wallet
+import pytest
```
Tzadiko, line 42: In monero/tests/ there is a README.md file that lists the dependencies one needs to install to run the tests. I would add pytest there.

tests/functional_tests/uri.py
Tzadiko: Looks better now. Thanks

src/wallet/wallet2.cpp
U65535F: Removed as no longer needed. Similar to the NA filter, it basically conveys that if the user cannot remove the addresses (in some case), then zeroing the amount that those addresses receive will help.

src/wallet/wallet2.cpp
U65535F: I realized this right after I pushed the commit. Fixed using boost::format.

U65535F:
Can you have a look at `uri.py`, is there any modification required?
See line 211 and 156. Should I create a new function that calls all the multi URI test? Or just leave the `run_test` method intact?

U65535F:
> You resolved a bunch of comments without them being addressed. I suggest we either come to a conclusion before they're resolved, or you make a change and push a commit. For example, after discussing I agree no further change is needed [here](https://github.com/monero-project/monero/pull/9830#discussion_r2001498757), so it can be resolved.

Don't worry. Instead of letting you know every time (that seems spammy), I combined all of the suggestions into one commit and pushed them. Only time I will comment is when there is an issue.

U65535F:
I wonder why my test fixture isn't getting executed properly.
```
  File "/__w/monero/monero/tests/functional_tests/uri.py", line 369, in <module>
    URITest().run_test()
  File "/__w/monero/monero/tests/functional_tests/uri.py", line 48, in run_test
    self.test_multi_uri_two_payments()
  File "/__w/monero/monero/tests/functional_tests/uri.py", line 258, in test_multi_uri_two_payments
    wallet = multi_uri_common_variables["wallet"]
TypeError: 'URITest' object is not subscriptable
[TEST FAILED] uri
```

Tzadiko:
> I wonder why my test fixture isn't getting executed properly.
> 
> ```
> Traceback (most recent call last):
>   File "/__w/monero/monero/tests/functional_tests/uri.py", line 369, in <module>
>     URITest().run_test()
>   File "/__w/monero/monero/tests/functional_tests/uri.py", line 48, in run_test
>     self.test_multi_uri_two_payments()
>   File "/__w/monero/monero/tests/functional_tests/uri.py", line 258, in test_multi_uri_two_payments
>     wallet = multi_uri_common_variables["wallet"]
> TypeError: 'URITest' object is not subscriptable
> [TEST FAILED] uri
> ```

`multi_uri_common_variables()["wallet"]`

U65535F:
> > I wonder why my test fixture isn't getting executed properly.
> > 
> > ```
> > Traceback (most recent call last):
> >   File "/__w/monero/monero/tests/functional_tests/uri.py", line 369, in <module>
> >     URITest().run_test()
> >   File "/__w/monero/monero/tests/functional_tests/uri.py", line 48, in run_test
> >     self.test_multi_uri_two_payments()
> >   File "/__w/monero/monero/tests/functional_tests/uri.py", line 258, in test_multi_uri_two_payments
> >     wallet = multi_uri_common_variables["wallet"]
> > TypeError: 'URITest' object is not subscriptable
> > [TEST FAILED] uri
> > ```
> Call the fixture to get the returned dictionary, then access it: `multi_uri_common_variables()["wallet"]`

```
  File "/__w/monero/monero/tests/functional_tests/uri.py", line 369, in <module>
    URITest().run_test()
  File "/__w/monero/monero/tests/functional_tests/uri.py", line 48, in run_test
    self.test_multi_uri_two_payments()
  File "/__w/monero/monero/tests/functional_tests/uri.py", line 258, in test_multi_uri_two_payments
    wallet = multi_uri_common_variables()["wallet"]
TypeError: 'URITest' object is not callable
[TEST FAILED] uri
```
Now this.....

Tzadiko:
Hey, Im going to stop reviewing this. There's been a persistent theme of you tagging me 8 hours after having not heard from me, and it's getting on my nerves. I've seen you do this with other people as well. I already spoke with you about this, and you said you're not going to continue pinging me. For example, not even 8 hours after leaving my last comment, you've pinged me again - I'm a human being, not a GPT, I need to sleep, I just woke up and I don't work for you (and even if I did, this would still be unacceptable). Not to mention, you're pinging me for issues that can be solved with simple debugging skills on code that you've supposedly written yourself - I'm unsure why you to ping me with regards to setting up a Python test fixture, there's plenty of documentation and examples online.

Tzadiko:
I spent a bunch of time reviewing this and trying to help you. I wasn't going to leave another comment, until I saw your recent force-push. Instead of doing your part by doing some basic debugging to address a roadblock you encountered that was prompted by my comment, you reverted your code to a point beforewhich my comment existed for expedience. It's apparent that you don't care for improving the quality of your submission (which would in turn increase its chances of being accepted), which I was trying to help you improve, if it requires just a tiny bit of effort for you to figure out how to write a test fixture in Python.

I'm just as new as you, but this sort of lazy development doesn't have a place in this codebase, in my opinion. **I do not recommend anyone else review this PR, you'll be wasting your time.** OP will ignore it at the earliest sign of inconvenience if it means he gets his code crammed through faster.

wqsv:
> I spent a bunch of time reviewing this and trying to help you. I wasn't going to leave another comment, until I saw your recent force-push. Instead of doing your part by doing some basic debugging to address a roadblock you encountered that was prompted by my comment, you reverted your code to a point beforewhich my comment existed for expedience. It's apparent that you don't care for improving the quality of your submission (which would in turn increase its chances of being accepted), if it requires just a tiny bit of effort for you to figure out how to write a test fixture in Python.
> 
> I'm just as new as you, but this sort of lazy development doesn't have a place in this codebase, in my opinion. **I do not recommend anyone else review this PR, you'll be wasting your time.** OP will ignore it at the earliest sign of inconvenience if it means he gets his code crammed through faster.

I realize now that my handling of this PR wasn’t ideal and clear enough. First off, a couple of clarifications:
- When I force-pushed, I wasn’t discarding your review suggestions instead I squashed my commits into one to keep the history clean. The changes you pointed out are still included,
-  Mistakenly, I thought GitHub only notified people when they are directly tagged, so I didn’t realize my activity was pinging everyone too frequently. I'll be more mindful about that and be quieter.
-  Regarding the test fixture: it's apparent that you're right, I should have spent more time debugging on my own before asking. That would've saved reviewers some time.

I'm still fairly new to contributing to larger repositories, so I'm learning as I go (this time, the hard way). I really appreciate the effort you put into reviewing though it wasn’t at all wasted. I'll take this as a lesson to be more patient, clear about my next steps, and of course be considerate of reviewers time.

nahuhh:
You'll need to drop the merge commit and use rebase instead

src/wallet/wallet_rpc_server.cpp
iamamyth: As I recall, the version suffix exists because there was a stated intent to deprecate and eventually eliminate the prior endpoint, in favor of the new endpoint. If the plan changes to support both indefinitely, then I agree, v2 would be a poor naming choice.

tests/functional_tests/uri.py
```diff
@@ -141,11 +143,17 @@ def test_monero_uri(self):
...
-        # external payment ids are not supported anymore
-        ok = False
-        try: res = wallet.make_uri(address = address, recipient_name = utf8string[0], tx_description = utf8string[1], amount = 1000000000000, payment_id = '1' * 64)
-        except: ok = True
-        assert ok
+        # standalone payment ids are supported but not recommended
+        res = wallet.make_uri(address = address, recipient_name = utf8string[0], tx_description = utf8string[1], amount = 1000000000000, payment_id = '1' * 64)
+        assert res.uri == 'monero:' + address + '?tx_amount=1.000000000000&recipient_name=' + quoted_utf8string[0] + '&tx_description=' + quoted_utf8string[1] + '&tx_payment_id=' + '1' * 64
+        
+        # in case of standalone payment id removal
```
iamamyth, line 150: I'd rather you just eliminate this piece if the feature is still supported, it's an obvious one line change and the commented version only brings confusion.

U65535F: I'm thinking whether to keep that part or not. Don't you think that it is kind of out of scope?

src/wallet/wallet_rpc_server.cpp
U65535F: Yes because the user never knows what type of uri is coming, it could be single recipient or multi. In case of multi uri, the legacy function fails. The newer functions work on both cases. If we continue on supporting both legacy and newer functions, it would introduce complexities in the code. First the user has to check if the uri is multi by calling the legacy function, then if it is, then call the newer function which handles the data in a slightly different way using vectors.

tests/functional_tests/uri.py
U65535F: I think I'm going to leave standalone payment IDs deprecated in favor of integrated addresses.

U65535F:
@selsta I really appreciate if you decide to review this PR. Since the new release had been successfully planned and scheduled on Friday. I hope this time you will have some time. 

Major improvements from previous old review:
1. Introduce new RPC commands: `make_uri_v2` and `parse_uri_v2`: these will handle both single and multi recipient transactions. The old/legacy versions (i.e., `make_uri` and `parse_uri`) will be kept to fully ensure the backwards compatibility. (as suggested by iamamyth).
2. Remove most of the major unnecessary edits (there are couple of newlines there to tackle down).
3. Tests are passing successfully. Old tests have not been modified in any way.

U65535F:
@selsta Hello, can you please tell me when you'll review this. I will patiently wait till that date.

selsta:
The three points you have written above all sound good and were necessary to get this merged.

In file src/wallet/wallet_rpc_server_commands_defs.h, please update `WALLET_RPC_VERSION_MINOR` to indicate API changes.

U65535F:
@selsta Done with the changes. 👍

nahuhh:
> @selsta Done with the changes. 👍

@U65535F thanks @U65535F  @U65535F

FiatDemise:
Has this been tested with an integrated address? I have a use case where I'd like multi recipient URI to work with 1 integrated address and like 5 regular addresses.

U65535F:
> Has this been tested with an integrated address? I have a use case where I'd like multi recipient URI to work with 1 integrated address and like 5 regular addresses.

It technically should work but no test case has been added. I think I should add a test case for this.

selsta:
Since we do not allow external payment IDs anymore, it might make sense to completely remove them from the v2 API enpoints.

U65535F:
> Since we do not allow external payment IDs anymore, it might make sense to completely remove them from the v2 API enpoints. Only integrated addresses from now on. On the other hand, the URI spec still has tx_payment_id so we might have to support it or update the spec.
> 
> https://github.com/monero-project/monero/wiki/URI-Formatting

Instead of supporting external payment IDs, why don't we directly update the spec? I'm *unsure* because the legacy URI functions don't support external payment IDs either. 
https://github.com/monero-project/monero/blob/master/src/wallet/wallet2.cpp#L14963

selsta:
@U65535F I will get the specification updated to remove `tx_payment_id`, you can update the APIs.

selsta:
URI spec has been updated to remove `tx_payment_id` and support multiple addresses.

https://github.com/monero-project/monero/wiki/URI-Formatting

FiatDemise:
Is there a maximum number of recipients? Or a recommended number to stay under?

U65535F:
> Is there a maximum number of recipients? Or a recommended number to stay under?

No limits for processing the URI though. But I don't know about the backend or the part that processes those tx. The scope of this PR is only on parsing and crafting a URI from custom data like recipient names, tx description, addresses and amounts.

U65535F:
Changes done. Here's what I changed:
1. `make_uri_v2` and `parse_uri_v2` now don't support standalone/external payment id. If payment id is still present in the URI, `parse_uri_v2` will classify it as unknown parameter and push to the vector of such parameters. (`payment_id=id` format). The `v2` URI spec has also been updated to remove the payment id field from it.
2. However, the legacy versions still have the `payment_id` parameter and still support it clearly but they don't support multi-URI functionality though. This is done to fully ensure the backwards compatibility with older clients.
@selsta What do you think with the changes? Are they okay or is there anything I'm missing out? Any suggestions or any minor nits to fix? Though I will have to squash those commits into one after I'm done with the modifications.

U65535F:
@selsta What's your opinion on this?

selsta:
> parse_uri_v2 will classify it as unknown parameter and push to the vector of such parameters. (payment_id=id format).

Can you clarify this?

U65535F:
> > parse_uri_v2 will classify it as unknown parameter and push to the vector of such parameters. (payment_id=id format).
> 
> Can you clarify this?

Sure. If the url contains the payment_id field, it will be pushed to the unknown_parameter vector.

selsta:
No questions currently, I just have to go through all the code one more time and I didn't have time for it yet.

jsyk57:
@selsta It's been almost 10 days. I will wait. Are you busy with the other PR reviews and the post-release work?

selsta:
I have time to go over it again today.

src/simplewallet/simplewallet.cpp
```diff
@@ -6583,34 +6608,20 @@ bool simple_wallet::transfer_main(const std::vector<std::string> &args_, bool ca
...
-        return false;
-      }
-      bool r = add_extra_nonce_to_tx_extra(extra, extra_nonce);
-      if(!r)
+      set_encrypted_payment_id_to_tx_extra_nonce(extra_nonce, info.payment_id);
+      if (!add_extra_nonce_to_tx_extra(extra, extra_nonce))
       {
-        fail_msg_writer() << tr("failed to set up payment id, though it was decoded correctly");
-        return false;
+          fail_msg_writer() << tr("failed to set up payment id");
```
selsta, line 6623: Please check here and in other places in this file that the indentation is 2 spaces, if you change it back to 2 spaces the diff will get slightly smaller in some places.

src/wallet/wallet2.cpp
```diff
@@ -15001,24 +15100,37 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
...
+  std::string params(ptr+1);
   std::vector<std::string> arguments;
-  std::string body = remainder.substr(address.size() + 1);
-  if (body.empty())
-    return true;
-  boost::split(arguments, body, boost::is_any_of("&"));
+  boost::split(arguments, params, boost::is_any_of("&"));
   std::set<std::string> have_arg;
-  for (const auto &arg: arguments)
+  for (const std::string &arg: arguments)
```
selsta, line 15133: Is this neccessary?

selsta (commented):
Generally from my side I'm ok with the patch, only added 2 small comments. I don't usually review wallet changes, so others might have different opinions.

jsyk57:
> I don't usually review wallet changes, so others might have different opinions.@selsta Would you approve this PR or will I have to wait for someone else to do it?

nahuhh:
monero is a collaboratively developed project. So, yes. You should be patient as others volunteer to review this PR.

This is not an immediate priority - there is not a release scheduled, and this doesnt fix anything. 

Just be patient. It could be a few weeks, or months, before this is reviewed by more people.
In the interim, you can move on to something else, or continue to make improvents here.

alefcopilot (approved):

src/wallet/wallet2.cpp
```diff
@@ -14942,56 +14942,155 @@ std::string wallet2::decrypt_with_view_secret_key(const std::string &ciphertext,
...
-    error = "A single payment id is allowed";
+    error = "No recipient information like addresses are provided.";
     return std::string();
   }
 
-  if (!payment_id.empty())
+  if (addresses.size() != amounts.size() || addresses.size() != recipient_names.size())
   {
-    error = "Standalone payment id deprecated, use integrated address instead";
+    error = (boost::format("The counts of addresses (%zu), amounts (%zu), and recipient names (%zu) do not match.") % addresses.size() % amounts.size() % recipient_names.size()).str();
```
selsta, line 14983:
```suggestion
    error = (boost::format("The counts of addresses (%1%), amounts (%2%), and recipient names (%3%) do not match.") % addresses.size() % amounts.size() % recipient_names.size()).str();
```

The current syntax appears incorrect. Please also update the other places that use `boost::format`.

https://www.boost.org/doc/libs/1_88_0/libs/format/doc/format.html

src/wallet/api/wallet.h
```diff
@@ -214,7 +214,9 @@ class WalletImpl : public Wallet
     virtual void startRefresh() override;
     virtual void pauseRefresh() override;
     virtual bool parse_uri(const std::string &uri, std::string &address, std::string &payment_id, uint64_t &amount, std::string &tx_description, std::string &recipient_name, std::vector<std::string> &unknown_parameters, std::string &error) override;
+    bool parse_uri(const std::string &uri, std::vector<std::string> &addresses, std::vector<std::uint64_t> &amounts, std::vector<std::string> &recipient_names, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error) override;
```
selsta, line 217:
```suggestion
    bool parse_uri(const std::string &uri, std::vector<std::string> &addresses, std::vector<uint64_t> &amounts, std::vector<std::string> &recipient_names, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error) override;
```

So that the signature matches with the existing one. Please do the same thing with `make_uri`.

jsyk57: thanks for introducing the problem, surprised it went overlooked till now. The entire code had `std::uint64_t` and `uint64_t` inconsistency. I've updated the code only to use `uint64_t` :)

selsta:
I will have one more round of suggestions tomorrow, but no major issues left.

src/simplewallet/simplewallet.cpp
```diff
@@ -6532,25 +6532,50 @@ bool simple_wallet::transfer_main(const std::vector<std::string> &args_, bool ca
...
 
     // check for a URI
-    std::string address_uri, payment_id_uri, tx_description, recipient_name, error;
+    std::string tx_description, error;
     std::vector<std::string> unknown_parameters;
-    uint64_t amount = 0;
-    bool has_uri = m_wallet->parse_uri(local_args[i], address_uri, payment_id_uri, amount, tx_description, recipient_name, unknown_parameters, error);
+    std::vector<std::string> addresses, recipient_names;
+    std::vector<uint64_t> amounts;
+    bool has_uri = m_wallet->parse_uri(local_args[i], addresses, amounts, recipient_names, tx_description, unknown_parameters, error);
```
selsta, line 6539: Did you test if this correctly sets the description afterwards?

src/wallet/wallet2.cpp
```diff
@@ -15027,31 +15139,40 @@ bool wallet2::parse_uri(const std::string &uri, std::string &address, std::strin
 
     if (kv[0] == "tx_amount")
     {
-      amount = 0;
-      if (!cryptonote::parse_amount(amount, kv[1]))
+      std::vector<std::string> amounts_split;
+      boost::split(amounts_split, kv[1], boost::is_any_of(";"));
```
selsta, line 15143: Should `amounts_split.erase()` be added here like with addresses?

selsta (commented):
Can you add a test for multiple integrated addresses (should fail)?

src/simplewallet/simplewallet.cpp
U65535F: How am I supposed to test that?

src/wallet/wallet2.cpp
U65535F: Yes. If there is a trailing semicolon, you will get an extra empty‐string token in the `amounts_split` vector which will cause the next if condition to fail. Thanks for pointing.

src/simplewallet/simplewallet.cpp
U65535F: > Compile it, and open moneo-wallet-cli and make a transaction with a URI containing a description and see if it sets the description.

I'm not really sure how I can see the description. I've tried googling and asked chatGPT but it was unsure.
This is what it says when I send a transaction to a wallet in testnet.
image
image

U65535F: @selsta I have modified the transfer_main code in simplewallet FYI.

U65535F: @selsta

U65535F: @selsta

selsta: Modified the code but did you test what I wrote?

U65535F: Yeah. I linked the image, see it.

selsta: I was confused because it appears the whole URI gets saved in the description field instead of just the description, or am I misunderstanding the image?

U65535F: > I was confused because it appears the whole URI gets saved in the description field instead of just the description, or am I misunderstanding the image?

I have no idea. That is what happens in the original app too iirc. I made so it achieves the same behavior as the original app.

U65535F: @selsta Should we push the description to `Tx Note` field?

selsta: I would say yes, it does not make sense as it is.

U65535F: @selsta Alright but in the Destination field, should we display the full URI or just the address itself (the original app displays the full URI)? I wonder what happens to unknown parameters? Where are they stored?

nahuhh: My 2c.

- Destination should only be the address 
- tx note should be the description
- the full uri should not be saved anywhere

selsta: Agree with @nahuhh

U65535F: image
Yes, it sets the description now (using `set_tx_note`). It doesn't store the full uri anywhere, now prints the address instead of the full uri like it used to do before. Note: the image has debug prints but I've removed them.

U65535F: > Please show a tx using multi-uri

It is not afaik. It's the way how it formats (`show_transfers`). I'll link an image with multiuri wait.

U65535F: @nahuhh Can you elaborate what I need to do?

nahuhh: For a multiURI, showing the amount per address looks fine. 

for a single transfer, probably redundant to show. But not a dealbreaker if this was existing behavior

U65535F: @selsta Can I resolve conversation now or are there any more things you can suggest?

U65535F:
@selsta Can you have a peek at simplewallet code and see if anything's wrong? Or if you can suggest something.

selsta:
Yes, will check tomorrow, I have to compile the PR and check how exactly it works now.

selsta:
I tested it in monero-wallet-cli and it looks good to me. I don't have any remaining suggestions at the moment.

selsta (approved):
I'm not sure if someone else is going to review this but from my side I'll add an approval.

sa8ab:
Is it possible to have multiple integrated addresses on one transaction?
Is it maximum 1 integrated address and rest sub address or primary addresses?

wqsv:
> Is it possible to have multiple integrated addresses on one transaction?
> Is it maximum 1 integrated address and rest sub address or primary addresses?

No multiple integrated addresses iirc.

src/wallet/wallet2.h
```diff
@@ -1641,8 +1641,11 @@ namespace tools
     template<typename T=std::string> T decrypt(const std::string &ciphertext, const crypto::secret_key &skey, bool authenticated = true) const;
     std::string decrypt_with_view_secret_key(const std::string &ciphertext, bool authenticated = true) const;
 
+    std::string custom_convert_to_url_format(const std::string &uri) const;
+    std::string make_uri(const std::vector<std::string> &addresses, const std::vector<uint64_t> &amounts, const std::vector<std::string> &recipient_names, const std::string &tx_description, std::string &error) const;
```
jeffro256: These new methods, including `parse_uri()`, do not need to be member functions of `wallet2`. I would recommend making them free functions. If they aren't to be free functions, at the very least mark the member functions as `static`, please.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3197,6 +3197,22 @@ namespace tools
...
+  {
+    if (!m_wallet) return not_open(er);
+    std::string error;
+    std::string uri = m_wallet->make_uri(req.addresses, req.amounts, req.recipient_names, req.tx_description, error);
```
jeffro256: This is why we shouldn't have `make_uri()`/`parse_uri()` be member functions of `wallet2`: you need to have initialized a `wallet2` instance for no reason.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3210,6 +3226,26 @@ namespace tools
...
   }
   //------------------------------------------------------------------------------------------------------------------------------
+  bool wallet_rpc_server::on_parse_uri_v2(const wallet_rpc::COMMAND_RPC_PARSE_URI_V2::request& req, wallet_rpc::COMMAND_RPC_PARSE_URI_V2::response& res, epee::json_rpc::error& er, const connection_context *ctx)
+  {
+    if (!m_wallet) return not_open(er);
+    std::string error;
+    std::vector<std::string> addresses;
+    std::vector<uint64_t> amounts;
+    std::vector<std::string> recipient_names;
+    if (!m_wallet->parse_uri(req.uri, addresses, amounts, recipient_names, res.uri.tx_description, res.unknown_parameters, error))
```
jeffro256: Same point here

src/wallet/wallet2.cpp
```diff
@@ -14933,56 +14933,155 @@ std::string wallet2::decrypt_with_view_secret_key(const std::string &ciphertext,
...
+    }
+    if (!recipient_names[i].empty())
+    {
+      recipients_used = true;
+      recipients_str += custom_convert_to_url_format(recipient_names[i]);
+    }
   }
 
-  if (amount > 0)
+  std::string uri = "monero:" + addresses_str;
```
jeffro256: I don't like the format of `<all addresses>?<all amounts>&<all messages>`, etc. Take a look at Bitcoin Unlimited's (node for Bcash) multiple address URI scheme: https://github.com/BitcoinUnlimited/BUIP/blob/master/086.md. The address, amount, and labels are next to each other for each output, which makes for more readable URIs. More importantly however, this allows the field for each output to remain undefined if desired. See this example for an actual use case: https://github.com/BitcoinUnlimited/BUIP/blob/master/086.md#example-1-bar-transaction. Here you can see that the amount field for a tip output at a bar is left empty, which triggers the client code to prompt the user to enter a desired amount. Overall, this type of URI format is *way* more flexible and future-proof than cramming values into a list based on field type, assuming the same structure for every single output in a transaction.

jeffro256 (changes requested):
I have some disagreements about the code structure, and the URI format itself.

src/wallet/wallet2.cpp
selsta: FWIW this is the format we decided on. I don't remember why exactly.

https://github.com/monero-project/monero/blob/master/docs/URI_SCHEME.md

jeffro256: Is it too late to change it if no one has a working implementation yet AFAIK?

selsta: https://github.com/monero-project/monero/issues/7731#issuecomment-848102482

Here is some background on when it was decided on. I'm not sure if anyone actually implemented this.

jeffro256: Seems like @trasherdk and @benevanoff wanted a similar approach to what I am bringing up here: packing info per-output, instead of packing per-value-type.

jeffro256: I don't know if I'd worry much about the byte-length of the URIs unless it becomes a practical problem. Seems a bit like a premature optimization, IDK. QR codes can handle fairly large URIs.

src/wallet/wallet_rpc_server.cpp
U65535F: alright, where would you like to see those free functions? Should I create a new source file or move the impl to some existing one?

src/wallet/wallet2.cpp
FiatDemise: > Seems like @trasherdk and @benevanoff wanted a similar approach to what I am bringing up here: packing info per-output, instead of packing per-value-type.

As far as I know, XMRChat is the first and only one to use this, and we don't have anything in prod yet. We're fine with whatever format you prefer. Just hope it can be done quickly as we'd like to implement this soon.

src/wallet/uri.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+  if (uri.substr(0, 7) != "monero:")
+  {
+    error = std::string("URI has wrong scheme (expected \"monero:\"): ") + uri;
+    return false;
+  }
+
+  std::string remainder = uri.substr(7);
+  const char *qptr = strchr(remainder.c_str(), '?');
+  std::string path_part = qptr ? remainder.substr(0, qptr - remainder.c_str()) : remainder;
```
BroccomaglosAdolfas: This variable (path_part) is unused

wqsv:
@selsta  Can you review the new code. I think I have fixed the problem in the wallet cli. You can test it and see if it works now (I've tested from my side). One thing I want to point out is that it only adds payment id when you explicitly provide a payment_id parameter in the uri. Otherwise it wouldn't add, you've seen this and requested the changes. Additionally, @jeffro256  can you please review the new code (in the `uri.cpp`)? Is that the format that you expected?
Is the `uri.cpp` and `uri.hpp` at the right place? Should I introduce wrappers around the new functions in the wallet code? What do you think?

EDIT: I was having a confusion. I did not think that payment ID should be also added when an integrated address is provided.

selsta:
I will re-review but need a bit of time

FiatDemise:
Am I understanding correctly that this is the new format? Is there supposed to be a question mark at the beginning?

monero:?output=address1;amount1;&output=address2;amount2;

src/wallet/wallet2.cpp
iamamyth: I think the design goal here was for the multi-payment scheme to match the original single-payment scheme in case of a single payment, i.e. it merely extends the existing scheme, rather than replacing it, making for a graceful upgrade (in particular, newly generated single-payment URIs work fine with existing consumers). That said, one could argue a "graceful" upgrade creates a false continuity, as multi-payments might parse OK with software that only supports single payments, making for user headaches. In any case, looking at the current tests, something has gone way off the rails, because the tests now include an `output` parameter, which they shouldn't.

If you want to change the scheme, such that it's not compatible with the old one, it should be a clean break (i.e. a new scheme name), otherwise you end up violating the core principle of URI namespacing.

iamamyth:
> Am I understanding correctly that this is the new format? Is there supposed to be a question mark at the beginning?
>
> monero:?output=address1;amount1;&output=address2;amount2;

Per my [comment](https://github.com/monero-project/monero/pull/9830#discussion_r2337738588), this looks wrong, as the original scheme specification (https://github.com/monero-project/monero/blob/master/docs/URI_SCHEME.md) contains no mention of an `output` param, and the single-output case should match the old format.

src/wallet/wallet2.cpp
wqsv: I get what you're saying. The solution is to change the scheme name so the old clients don't accidentally parse the new multi recipient URI. I wonder what should be the new URI scheme name if that's what you mean? Additionally, should the new functions parse and generate old scheme URIs if there's it's a single recipient?

iamamyth: We can't decide on minor issues like the scheme name before resolving the major question of whether to split the scheme at all. So, it seems any work should be on hold for now. I see two decent options:

1. Extend the scheme to preserve the single-payment format and intentionally differ from it for multi-payment URIs in a manner that makes it highly unlikely multi-payment URIs introduce false compatibility.
2. Keep the current single-payment URI scheme as-is, and introduce a new, multi-payment URI scheme.

In either case, I think single-payment URIs should work as-is (i.e. no changes to the single-payment use case, anything that previously worked will still work); otherwise, a lot of software breaks for a minor upgrade.

wqsv: Can we not modify the old functions to reject the URI as multi-recipient if it contains any `output` parameter? If this is possible, this would eliminate the problem of false compatibility if I'm not wrong?

jeffro256: > I think the design goal here was for the multi-payment scheme to match the original single-payment scheme in case of a single payment, i.e. it merely extends the existing scheme, rather than replacing it, making for a graceful upgrade (in particular, newly generated single-payment URIs work fine with existing consumers). 

[The scheme I proposed](https://github.com/monero-project/monero/issues/7731#issuecomment-3239824513) achieves this goal. AFAICT, the current PR does not implement this scheme.

jeffro256: > We can't decide on minor issues like the scheme name before resolving the major question of whether to split the scheme at all. So, it seems any work should be on hold for now. I see two decent options:
> 
>     1. Extend the scheme to preserve the single-payment format and intentionally differ from it for multi-payment URIs in a manner that makes it highly unlikely multi-payment URIs introduce false compatibility.
> 
>     2. Keep the current single-payment URI scheme as-is, and introduce a new, multi-payment URI scheme.
> 
> 
> In either case, I think single-payment URIs should work as-is (i.e. no changes to the single-payment use case, anything that previously worked will still work); otherwise, a lot of software breaks for a minor upgrade.

I vote for 1. The multi-recipient format should be backwards compatible for a single recipient.

FiatDemise:
> > Am I understanding correctly that this is the new format? Is there supposed to be a question mark at the beginning?
> > monero:?output=address1;amount1;&output=address2;amount2;
> 
> Per my [comment](https://github.com/monero-project/monero/pull/9830#discussion_r2337738588), this looks wrong, as the original scheme specification (https://github.com/monero-project/monero/blob/master/docs/URI_SCHEME.md) contains no mention of an `output` param, and the single-output case should match the old format.

Makes sense! Got another question, I noticed the URI_SCHEME.md specifies "tx_amount". Jeffro's proposal just has "amount". Does that mean a difference between single and multi like this? (newlines added for clarity)

Single Recipient
monero:41nw...rq6b
?tx_amount=0.01

Multi Recipient
monero:8A6a...Wr6C
?version=2.0
&amount=0.001682284
&address=4Csc...8pTU5
&amount=0.0148040992
&address=86EU...wmot
&amount=0.0003364568

On 9/21 edited this post to add the "?version=2.0" to multi recipient example.

src/simplewallet/simplewallet.cpp
```diff
@@ -6479,15 +6480,18 @@ bool simple_wallet::transfer_main(const std::vector<std::string> &args_, bool ca
...
-      return false;
+      std::string extra_nonce;
+      set_payment_id_to_tx_extra_nonce(extra_nonce, payment_id);
+      if (!add_extra_nonce_to_tx_extra(extra, extra_nonce))
+      {
+        fail_msg_writer() << tr("failed to set up payment id");
+        return false;
+      }
+      payment_id_seen = true;
+      local_args.pop_back();
```
jeffro256: I'm fairly certain that this part is breaking payment ID handling, I would revert it back.

src/simplewallet/simplewallet.cpp
```diff
@@ -6522,25 +6527,128 @@ bool simple_wallet::transfer_main(const std::vector<std::string> &args_, bool ca
...
+
+
+      std::string payment_id_uri;
+      for (auto &unknown_parameter : unknown_parameters) 
+      {
+        if (boost::starts_with(unknown_parameter, "tx_payment_id="))
+        {
+          payment_id_uri = unknown_parameter.substr(std::string("tx_payment_id=").size());
+        }
+      }
```
jeffro256: Support for unencrypted long payment IDs should be dropped.

src/wallet/uri.hpp
```diff
@@ -0,0 +1,62 @@
...
+  #define HAVE_INT128 0
+#endif
+
+#if HAVE_INT128
+  using uint128 = unsigned __int128;
+#else
+  #include <boost/multiprecision/cpp_int.hpp>
+  using uint128 = boost::multiprecision::uint128_t;
+#endif
+
```
jeffro256, line 43: Need a namespace here. I recommend `tools::wallet`

src/wallet/uri.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+typedef struct {
+    uint128 amount;
```
jeffro256: There's not really a reason for `amount` to be 128-bit since amounts for a single transaction output are only 64-bits. You can do a lot of simplification if you drop support for 128-bit amounts. If someone really wants to receive more than 18.4 million XMR in a single transaction, they can make a URI with two amount fields lol.

src/wallet/uri.cpp
```diff
@@ -0,0 +1,825 @@
...
+  while (pos > 0 && std::isalpha(static_cast<unsigned char>(s[pos - 1]))) --pos;
+
+  std::string number = s.substr(0, pos);
+  std::string unit = s.substr(pos);
+
+  std::string unit_upper = unit;
+  for (char &c : unit_upper) 
+  {
+    c = static_cast<char>(toupper((unsigned char)c));
+  }
```
jeffro256: You can also use [`boost::to_upper()`](https://www.boost.org/doc/libs/1_41_0/doc/html/boost/algorithm/to_upper.html) here.

src/wallet/uri.hpp
```diff
@@ -0,0 +1,62 @@
...
+    std::string currency;
+} tx_amount;
+
+std::string to_string_u128(uint128 v) ;
+bool parse_u128(const std::string &u128_str, uint128 &out);
+
+std::string make_uri(const std::vector<std::string> &addresses, const std::vector<tx_amount> &amounts, const std::vector<std::string> &recipient_names, const std::string &tx_description, std::string &error, cryptonote::network_type network_type = cryptonote::UNDEFINED);
+std::string make_uri(const std::vector<std::string> &addresses, const std::vector<uint64_t> &xmr_amounts, const std::vector<std::string> &recipient_names, const std::string &tx_description, std::string &error, cryptonote::network_type network_type = cryptonote::UNDEFINED);
+bool parse_uri(const std::string &uri, std::vector<std::string> &addresses, std::vector<tx_amount> &amounts, std::vector<std::string> &recipient_names, std::string &tx_description, std::vector<std::string> &unknown_parameters, std::string &error, cryptonote::network_type network_type = cryptonote::UNDEFINED);
```
jeffro256: I would add an overload which returns a vector of `cryptonote::tx_destination_entry`, a short payment ID, and the nettype in place of a vector of address strings and amounts, since the very next move of parsing a URI is most likely going to be parsing the addresses (like is the case in `simplewallet`).

src/wallet/uri.hpp
wqsv: uint64_t is ok for Monero and other assets but when it comes to Ethereum, it uses 18 decimals. 64 bits won't cut it. Even 20 ETH in wei exceeds 2^64-1, that's why I intend to have uint128 so there wouldn't be any overflows.

wqsv:
@selsta Would you like to review this when you have some time?

src/wallet/uri.cpp
```diff
@@ -0,0 +1,947 @@
...
+};
+//----------------------------------------------------------------------------------------------------
+const std::unordered_map<std::string, currency_info> &currency_table() {
+  static const std::unordered_map<std::string, currency_info> table = {
+    {"XMR", {12, 1000000000000ULL}},
+    {"BTC", {8,  100000000ULL}},
+    {"ETH", {18, 1000000000000000000ULL}},
+    {"USD", {2,  100ULL}},
+    {"EUR", {2,  100ULL}},
+  };
```
plowsof: "add multiple-URI support to `parse_uri` and `make_uri`" but glancing over some of these function names im wondering what.... so, sending monero to multiple addresses why currency table? scale and decimals? suffix to currency? i get supporting both atomic units but im very confused

wqsv: Sorry, it was a typo that went unnoticed for... quite a long time. The [scheme that jeffro suggested](https://github.com/monero-project/monero/issues/7731#issuecomment-3239824513) supported multiple currencies in the uri.

plowsof:
is it too late to ask a reputable dev to remake this PR from scratch

benevanoff:
@plowsof I was the original dev to start this implementation, about 3 years ago if memory serves correctly, but I had abandoned it because I wasnt getting much feedback on testing/code review. If this is something that people are now more interested in then I can just branch off the latest master and start from scratch. If this is the route we want to take then I will pop into IRC later in the day.

FiatDemise:
> @plowsof I was the original dev to start this implementation, about 3 years ago if memory serves correctly, but I had abandoned it because I wasnt getting much feedback on testing/code review. If this is something that people are now more interested in then I can just branch off the latest master and start from scratch. If this is the route we want to take then I will pop into IRC later in the day.

Hi @benevanoff , I am the project manager for XMRChat and am very interested in multi-recipient URI. It would allow us to implement a new payment splits feature, and it would allow us to receive a portion of tips, without having to custody the entire tip first. And that changes XMRChat from a community funded project to being self sustaining. I believe this will be useful to other projects too. Has @plowsof replied to you yet?

FiatDemise:
Hi @jeffro256 , in your proposed format at https://github.com/monero-project/monero/issues/7731#issuecomment-3239824513 , your example multi-recipient URI has "&amount=1XMR". Is the "XMR" needed, or can that part just be "&amount=1"?

plowsof:
we need devs who know what theyre looking to give feedback at to make this call - but me, sitting from the gallery can see a monkey see monkey do prompt attempt at this - where reviews are only adding more confusion to the hallucination and draining more and more time at each step. I can only assume a correct review of this will take the same time and understanding required to re-do this from scratch (likely less time)

wqsv:
> we need devs who know what theyre looking to give feedback at to make this call - but me, sitting from the gallery can see a monkey see monkey do prompt attempt at this - where reviews are only adding more confusion to the hallucination and draining more and more time at each step. I can only assume a correct review of this will take the same time and understanding required to re-do this from scratch (likely less time)

What's the issue with the current code? If it's the size, I know it's a mess and it's sloppily written. I will have to clean up the code and use more Boost and STL to reduce the code. There seem to be some superfluous tests as well which I need to remove.

plowsof:
Force pushing to the 1 AI slop commit with more slop and brute forcing workflows to work on an open PR (from master branch) causes alot of spam

iamamyth:
@plowsof I don't think it'd be that hard to make this feature work (having reviewed the original PR by benevanoff and this one), but I think the URI scheme merits closer examination (I'd say the spec is the hard part). I don't particularly like the ordering dependency in the last proposal I saw, a double-encoding scheme for the additional transfers could eliminate that aspect while still being a backwards-compatible extension to the original URI scheme.

SlowBearDigger:
Related implementation to support multiple destinations in Monero URIs (Haveno): https://github.com/haveno-dex/haveno/pull/2107

Issue #9665, opened 2024-12-30
[Bug] monerod on win does not load bitmonero.conf from default data dir without setting --config-file


5andr0:
monerod v0.18.3.4 won't load bitmonero.conf from default dir C:\ProgramData\bitmonero without setting any cli arguments. 
I had to set --config-file=C:\ProgramData\bitmonero\bitmonero.conf
I tested it multiple times on win 11 latest by setting a unique log-file in the conf
That broke my last sync with alternating prune settings that I had in the bitmonero.conf

selsta:
Can you explain what you mean with prune settings? You only have to set `--prune-blockchain` once for it to be pruned.

5andr0:
> Can you explain what you mean with prune settings? You only have to set `--prune-blockchain` once for it to be pruned.

Oh, thanks for the clarification. I thought it bricked my chain data when it loaded for a while without prune blockchain enabled, so I resynced from scratch again.

Anyways, the bug still persists about not loading the default bitmonero.conf

0xFFFC0000:
![Image](https://github.com/user-attachments/assets/bcfc3bb8-6f0b-4736-b4ab-119e8dc68e06)

It is not a bug; it is expected behavior. 

IF you are running on admin it will load from `CSIDL_APPDATA` ( `C:\ProgramData` ). If you are not admin, it will load from `CSIDL_APPDATA` (`C:\Users\<Username>\AppData\Roaming`)

nahuhh:
@5andr0 if you run with no args, is the blockchain synced to `C:\ProgramData\bitmonero`?

PR #9833, opened 2025-03-13
src: fix windows path handling in daemonizer (#9665)

0xFFFC0000:
#9665

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -75,6 +75,7 @@ target_link_libraries(common
     ${Boost_REGEX_LIBRARY}
     ${Boost_CHRONO_LIBRARY}
   PRIVATE
+    ${EPEE_READLINE}
     ${OPENSSL_LIBRARIES}
     ${EXTRA_LIBRARIES})
 target_include_directories(common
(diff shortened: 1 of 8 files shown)


selsta:
Were you able to reproduce the original report?

0xFFFC0000:
> Were you able to reproduce the original report?

I should have clarified it better. After much trying I was not able to reproduce the bug. But I was able to pinpoint a bug in our Windows path detection. At first glance, this looks like very easy issue. But needs a lot of time ( proportionally ) to fix it. No matter the user permission, we were always putting the default folder in `C:\ProgramData\bitmonero`.  But as the code in the link says the design was to put the `bitmonero` folder inside `'C:\Users\<USER>\AppData\Roaming` folder for non admin users [1]. But due to coding issue, we were not calling that method at all. 


1. https://github.com/monero-project/monero/blob/f90a267fa34bad095d7e8ba72ee78f2a63f37df6/src/daemonizer/windows_daemonizer.inl#L87

selsta:
One concern now is that users would suddenly load a different `bitmonero` folder and start to resync from scratch when updating with this patch. Is this something that could happen?

0xFFFC0000:
> One concern now is that users would suddenly load a different `bitmonero` folder and start to resync from scratch when updating with this patch. Is this something that could happen?

Yes, that can happen, and this is totally valid concern. Let's keep this in mind. If we decided that `C:\ProgramData\bitmonero` is default location, we can remove a lot of deadcode.

PR #9840, opened 2025-03-17
trezor: fix cmake variables

ComputeryPony:
I'm not a cmake expert but it seems this should be `CMAKE_CURRENT_LIST_DIR`, as otherwise when building monero as a submodule of monero-gui `CMAKE_SOURCE_DIR` evaluates to the `monero-gui` folder instead of `monero` and breaks trezor support.


tobtoht (approved):

Tzadiko:
Is anyone a CMake expert? 😄

plowsof (approved):

PR #9845, opened 2025-03-17
Blockchain: remove skip for pruned txs in `check_tx_inputs()`

jeffro256:
I can't see how this would trigger in the current codebase, so it's not a *current* safety issue, but I can very well see it becoming on in the future if downstream code doesn't handle the passing of pruned transactions correctly. I think the safe/good choice would be to remove this skip now that all transactions that pass into the mempool are supposed to be unpruned. And for all in-block txs, `check_tx_inputs()` isn't called for checkpointed blocks, and we sync pruned blocks only if syncing checkpointed blocks.

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -3268,10 +3268,6 @@ bool Blockchain::check_tx_inputs(transaction& tx, tx_verification_context &tvc, 
   if(pmax_used_block_height)
     *pmax_used_block_height = 0;
 
-  // pruned txes are skipped, as they're only allowed in sync-pruned-blocks mode, which is within the builtin hashes
-  if (tx.pruned)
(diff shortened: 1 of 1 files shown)


vtnerd (approved):

j-berman (approved):
I think this is good as is, but I suggest an explicit check that makes sure the tx isn't pruned, like:

```C++
      if (tx.pruned)
      {
        MERROR_VER("Pruned tx not allowed, cannot verify inputs");
        tvc.m_pruned_tx_not_allowed = true;
        return false;
      }
```

jeffro256:
@j-berman I think that that's a good idea, but probably best for a separate PR since this diff is so simple and introducing a `m_pruned_tx_not_allowed` field would add more code.

SyntheticBird45 (approved):

PR #9846, opened 2025-03-17
Remove hidden transaction copy in add_block

vtnerd:
I found a place where `std::pair<transaction, blobdata>` was given to a function that took a `const std::pair<transaction, blobdata_ref>&` argument. As per C++ rules, this forces a hidden implicit copy of `transaction`, which is a rather expensive operation in the `add_block` routine. Additional, the `miner_tx` was being copied too (and also in a migration function, which is less important).

I haven't done any timing/performance analysis, but this should help block syncing by a tiny amount. If this passes checks, I'm going to propose this for 0.18, except using `blobdata_ref` instead of `span<...>`. I'm using `span<...>` in this patch because I hope to change things over from `blobdata` (aka `std::string`) to `epee::byte_slice`.

### DIFF
--- a/src/blockchain_db/blockchain_db.cpp
+++ b/src/blockchain_db/blockchain_db.cpp
@@ -179,10 +179,8 @@ void BlockchainDB::pop_block()
   pop_block(blk, txs);
 }
 
-void BlockchainDB::add_transaction(const crypto::hash& blk_hash, const std::pair<transaction, blobdata_ref>& txp, const crypto::hash* tx_hash_ptr, const crypto::hash* tx_prunable_hash_ptr)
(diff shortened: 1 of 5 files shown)


vtnerd:
Arch Linux build is throwing errors from protobuf. This this expected? I definitely don't have this happen locally.

selsta:
@vtnerd arch issue is fixed with #9839

vtnerd:
The failed unit tests should be unrelated.

Tzadiko:
In case others want to review but are not sure what you're referring to with the copying: https://godbolt.org/z/8YT6sqz6K

Tzadiko (approved):

0xFFFC0000 (approved):
Test failure is unrelated. LGTM

nahuhh (approved):
Tested

PR #9849, opened 2025-03-18
Convert CRLF to LF.

sorenstoutner:
I am in the process of finalizing the Feather Wallet Debian package.  As part of that process, Debian uses a Git based system that imports the upstream tarball and its contents into various branches of a Git repository.  During one step, the contents of the original tarball are checked against the contents of the unpacked tarball contained in another.  In the case of Feather Wallet this check fails because some of the files in the original tarball have CRLF line endings, which were automatically converted to LF line endings in the unpacked branch.

CRLF line endings are the default on Windows and LF line endings are the default on Linux.  Git repositories typically store all their text files with LF line endings and these are automatically converted back and forth to CRLF for Windows users, although this behavior can be changed if there is a good reason for doing so.  In the case of Monero, because only a few files are CRLF, my guess is that is unintentional and it wouldn't cause any problems if they were converted to LF, which is what this pull request does.  However, there is a small chance that changing the mnemonics files to LF might cause some incompatibility.  Does your test suite currently check that the mnemonics files can be parsed correctly?

I have several ways of working around this issue in the Debian packaging, and if there is an important reason for these files to be CRLF I can do so.  But if not, it would be helpful if the entire project used LF for text files.


plowsof:
> I put the probability that a 20,000 line change gets merged at 0.0001%

reviewers can run a small string replace shell script that targets these 18 files and some CI check added to prevent CRLF line endings to be added moving forward. I'm not qualified to offer an opinion on this PR though, just giving some hope :)

sorenstoutner:
> reviewers can run a small string replace shell script that targets these 18 files and some CI check added to prevent CRLF line endings to be added moving forward. I'm not qualified to offer an opinion on this PR though, just giving some hope :)

You can also use the `dos2unix` command, which is generally available in Linux distributions, although not always installed by default, to convert CRLF to LF in text files.

nahuhh:
> You can also use the `dos2unix` command, which is generally available in Linux distributions, although not always installed by default, to convert CRLF to LF in text files.

sure, but i doubt were going to destroy the git blame and attribute 20000 line commit to change whitespace. You can this downstream.

sorenstoutner:
Do these files change often or have much of a git blame history?

nahuhh:
IMO it's not simply about whether they change often, or whether they have a deep history. It's that _you_ become listed as the new author on every line of these files, instead of each line being attributed to the code author (and commit).

If there ever an issue, the blame is used to track down the commit that introduced the issue.

again, just my opinion, but i'd personally NACK this (and any other pr that only changes whitespace)

iamamyth:
> it would be helpful if the entire project used LF for text files.

I think you have the wrong approach here: Assuming no CRLF line endings makes your process inherently brittle. Even if you ensure the repo has no CRLF line endings at some particular point in time, that doesn't prevent the introducing of new CRLF line endings. You need to simply not make the assumption in the first place.

sorenstoutner:
Usually when I contact upstream projects about CRLF issues, they not only fix the existing problems, but they configure their repository to automatically remove any problems from future uploads.

https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings

Enforcing LF in Git repositories is a pretty common practice.  However, if you decide you don't want to do so, it doesn't cause that much of a problem for me to work around it during packaging.

sorenstoutner:
It is true that this is the largest change by number of lines, which is mostly because the mnemonic files are very long, being basically dictionary lists.  Also, this is by no means a comprehensive list as the upstream projects I have suggested this to are not all on GitHub, and not even all of the one I have submitted on GitHub are listed above.

Perhaps the most important thing I would say about this is that prior to this discussion I have only had one project reject my patch, which was courier-libs.  In that case, there is an important reason for these files to have CRLF line endings, as they are part of the test files for the email suite.  CRLF line endings are baked into the email RFCs, which I didn't know when I submitted the pull request.  The test suite has example success and error messages which are compared against the out put of the tests.  For this to work correctly, they need to have CRLF line endings.

So, in the case of courier-libs I was happy to work around the issue in the Debian packaging because there was a strong technical need for these files to have CRLF line endings.

Otherwise, I have never encountered any resistance or concerns with converting all the text files in the  repositories to LF.

As a Debian packager, I value having a positive relationship with upstream projects.  For example, I value having a positive relationship with the developers of the Monero project more than I value converting these files to LF.  As there is more resistance to this change than I anticipated, I will close this Pull Request.

PR #9851, opened 2025-03-18
Improve move/copy/default constructor semantics for tx and block

vtnerd:
This one is a little aggressive - it adds move semantics to `transaction` and `block`. I deleted the move constructor+assignment to both objects and tried compiling everything but tests. This identified every move operation (that is currently secretly a copy), and I manually verified that the moves were valid. There was one `transaction` case that I could not verify, which I downgraded to a copy.

This PR also changes the default+copy constructors, and copy assignment. The reasoning was the `atomic` values - the changes have fewer memory barriers and branches, which should result in slightly faster code (particularly on arm).

The primary benefit is a `transaction` copy becomes a move in `add_block`. This should have a (small but) measurable improvement in block syncing, etc, as a `transaction` copy is rather expensive (due to the number of allocations).

### DIFF
--- a/src/cryptonote_basic/cryptonote_basic.h
+++ b/src/cryptonote_basic/cryptonote_basic.h
@@ -225,8 +225,10 @@ namespace cryptonote
 
     transaction();
     transaction(const transaction &t);
+    transaction(transaction &&t);
     transaction &operator=(const transaction &t);
-    virtual ~transaction();
+    transaction &operator=(transaction &&t);
+    ~transaction() = default;
     void set_null();
     void invalidate_hashes();
     bool is_hash_valid() const { return hash_valid.load(std::memory_order_acquire); }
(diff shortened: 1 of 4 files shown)


src/cryptonote_basic/cryptonote_basic.h
```diff
@@ -225,8 +225,10 @@ namespace cryptonote
 
     transaction();
     transaction(const transaction &t);
+    transaction(transaction &&t);
     transaction &operator=(const transaction &t);
-    virtual ~transaction();
```
Tzadiko, line 229: Nothing inherits from transaction. So just this change alone will reduce the size of the struct by 8 bytes (assuming 64-bit system). Given millions of transactions, that's a lot of memory! 👍

src/cryptonote_basic/cryptonote_basic.h
```diff
@@ -346,79 +348,101 @@ namespace cryptonote
...
+    transaction_prefix(std::move(t)),
+    hash_valid(t.is_hash_valid()),
+    prunable_hash_valid(t.is_prunable_hash_valid()),
+    blob_size_valid(t.is_blob_size_valid()),
+    signatures(std::move(t.signatures)),
+    rct_signatures(std::move(t.rct_signatures)),
+    hash(std::move(t.hash)),
+    prunable_hash(std::move(t.prunable_hash)),
+    blob_size(std::move(t.blob_size)),
+    pruned(std::move(t.pruned)),
```
Tzadiko, line 374: Moving a bool seems like overkill.

src/cryptonote_basic/cryptonote_basic.h
```diff
@@ -346,79 +348,101 @@ namespace cryptonote
...
-    {
-      prunable_hash = t.prunable_hash;
-      set_prunable_hash_valid(true);
-    }
+    t.set_null();
   }
 
   inline transaction &transaction::operator=(const transaction &t)
   {
+    if (this == std::addressof(t))
```
Tzadiko, line 383: Curious why you prefer addressof instead of `&`?

src/cryptonote_basic/cryptonote_basic.h
```diff
@@ -479,8 +503,11 @@ namespace cryptonote
 
   public:
     block(): block_header(), hash_valid(false) {}
-    block(const block &b): block_header(b), hash_valid(false), miner_tx(b.miner_tx), tx_hashes(b.tx_hashes) { if (b.is_hash_valid()) { hash = b.hash; set_hash_valid(true); } }
-    block &operator=(const block &b) { block_header::operator=(b); hash_valid = false; miner_tx = b.miner_tx; tx_hashes = b.tx_hashes; if (b.is_hash_valid()) { hash = b.hash; set_hash_valid(true); } return *this; }
+
+    block(const block &b): block_header(b), hash_valid(b.is_hash_valid()), miner_tx(b.miner_tx), tx_hashes(b.tx_hashes), hash(b.hash) {}
+    block(block &&b): block_header(std::move(b)), hash_valid(b.is_hash_valid()), miner_tx(std::move(b.miner_tx)), tx_hashes(std::move(b.tx_hashes)), hash(std::move(b.hash)) { b.miner_tx.set_null(); b.tx_hashes.clear(); }
+    block &operator=(const block &b) { if(this != std::addressof(b)) { block_header::operator=(b); hash_valid = b.is_hash_valid(); miner_tx = b.miner_tx; tx_hashes = b.tx_hashes; hash = b.hash; } return *this; }
```
Tzadiko, line 509: NIT: Looks a little unkempt when it's one line. I know that it was like that before, but I think now would be a good time to expand it. In other words:

```cpp
Method()
{
    // Stuff
}
```
instead of
```cpp
Method() { /* Stuff */ }
```

src/cryptonote_core/blockchain.cpp
```diff
@@ -2656,7 +2656,7 @@ bool Blockchain::get_transactions(const t_ids_container& txs_ids, t_tx_container
       bool res = pruned ? m_db->get_pruned_tx_blob(tx_hash, tx) : m_db->get_tx_blob(tx_hash, tx);
       if (res)
       {
-        txs.push_back(transaction());
+        txs.emplace_back();
         res = pruned ? parse_and_validate_tx_base_from_blob(tx, txs.back()) : parse_and_validate_tx_from_blob(tx, txs.back());
```
Tzadiko, line 2660: NIT: Consider

```cpp
auto& added_tx = txs.emplace_back();
res = pruned ? parse_and_validate_tx_base_from_blob(tx, added_tx) : parse_and_validate_tx_from_blob(tx, added_tx);
```

src/cryptonote_basic/cryptonote_basic.h
vtnerd: I think the recommendation is to primarily use in templated code, but I've been using it for any type in case an `operator&` is ever defined. Its probably a pessimistic, but I could switch to `&` here if desired.

src/cryptonote_basic/cryptonote_basic.h
vtnerd: Updated.

src/cryptonote_core/blockchain.cpp
vtnerd: Added.

src/cryptonote_basic/cryptonote_basic.h
Tzadiko: Keep it, that makes sense. Was moreso curiousity on my end.

Tzadiko (approved):
Looks good!

0xFFFC0000 (approved):
LGTM

0xFFFC0000:
I already approved, but keep in mind something like this mind help: 

`
if (this == &other) return *this;
`

vtnerd:
### Performance Comparison

I did some basic performance testing. On an i7-1185G7 (XPS laptop), moving a `transaction` takes ~270 nanoseconds, and copying takes like ~1.1 microseconds. So the improvement is ~75%, but this is nothing compared to the total time spent in `add_block`. I think its still worth doing since the code isn't too difficult to compare

### Uninitialized Read Problem

I'm going to make a change where `hash`, `prunable_hash`, and `blob_size` are initialized in the `transaction` default constructor. The copy+move constructor+assignment operations always copies these values because its should be a few ticks faster than branching or adding additional memory barriers. However, memory tools report these copies as errors, so its probably easier to just initialize.

vtnerd:
@Tzadiko and @0xFFFC0000 my apologies but I made a small change - the default constructor initializes 3 members that were previously left uninitialized. This should prevent memory checkers from spewing errors.

0xFFFC0000 (approved):
lgtm

Tzadiko:
To prevent the latest comment from being an issue in the future, thoughts on brace initializing these member fields? The char arrays as well as size_t will then be zero initialized.

image

vtnerd:
@Tzadiko for that to be an issue, someone would have to remove the fields from the default constructor.

Tzadiko:
I was moreso worried that someone adds another constructor later and forgets to initialize them, but it's fine the way it is with a single default constructor.

Tzadiko (approved):

nahuhh (approved):
Tested

PR #9856, opened 2025-03-22
Use amd64 accelerations for scalarmultKey in verify code

vtnerd:
I profiled a post ringct block sync and discovered that `rct::scalarmultKey` took ~60% of CPU time on a Ryzen 3900x. Using amd64 accelerations, already in use by the wallet, dropped this function to ~18% of the CPU time. The same base code has been in use by the wallet for years without issue, so the likelihood of bugs is low. I have **not** done a real-world sync test to determine the difference, if any, this provides. If no community members pick up this kind of testing, I will try to figure this out at some point, as its likely never going to get merged without such analysis.

This same CPU was used to profile the #6337 wallet changes. Each `scalarmultKey` call to the 32-bit code takes ~93 microseconds, whereas the 64-bit code takes ~37 microseconds. There's at least 3 of these calls per transaction, so with 40 transaction blocks this saves about 7 milliseconds per block. However, the remainder of the ringct still uses the other crypto functions, so there may be some CPU cache thrashing effects. Since the transaction verification code does everything in batches, the CPU cache isn't constantly switching between the two libraries, so the penalty of the two libraries is mostly mitigated.

My other recent PRs remove `transaction` copies - after this PR `~transaction` is now fairly high up on the list of "hot spots". I didn't profile _with those changes merged_, so between these patches we are fairly close on reducing the sync time.

This is still marked as draft as this changes consensus code to use a different library depending on build. We could, as a matter of safety, use this only when "expanding" transactions in `cryptonote_format_utils.cpp` (leaving the curve order check to always use the "safer" ref10 code).

### DIFF
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -38,6 +38,7 @@
 #include "cryptonote_config.h"
 #include "crypto/crypto.h"
 #include "crypto/hash.h"
+#include "ringct/accel.h"
 #include "ringct/rctSigs.h"
 
 using namespace epee;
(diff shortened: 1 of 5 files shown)


selsta:
I always thought that we should switch to the non-accelerated versions once the daemon is synced up, this assures no possible chain split. On the other hand, the library is well tested at this point, so this might not be necessary.

iamamyth:
In theory, once FCMP++ HF goes live, assuming sync works with this patch, it's quite low risk (and the risk diminishes with every passing block).

nahuhh:
> I have not done a real-world sync test to determine the difference, if any, this provides. If no community members pick up this kind of testing, I will try to figure this out at some point, as its likely never going to get merged without such analysis.

Syncing current blockchain works w/ this patch (and appears to increase sync speed 40%, from ~100k blocks/hr to ~140k blocks/hr between blocks 2.25m and 2.8m)

PR #9859, opened 2025-03-23
fix: added key correspondence check to monero-wallet-rpc generate_from_keys

StrawberryChocolateFudge:
I created this pull request to propose fixing the error handling in monero-wallet-rpc generate_from_keys where the key correspondence is not checked.

Missing checks lets users create malformed wallets. There was no checks to make sure the entered address, viewkey and spendkey are actually corresponding to each other when creating a wallet via RPC with generate_from_keys, so this adds error handling for it.

Without this, if the keys are not corresponding to each other generate_from_keys still returns "Wallet has been generated successfully" but open_wallet will error with "THROW EXCEPTION: error::wallet_files_doesnt_correspond"

I would like to add the checks so generate_from_keys can't output malformed wallet files.

The key correspondence checks were taken from the wallet2.cpp generate_from_json function where the key correspondence is checked properly.

please review. I found no corresponding tests to update

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -3891,7 +3891,20 @@ namespace tools
       er.message = "Failed to parse view key secret key";
       return false;
     }
-
+    
+    crypto::public_key pkey;
+    if (!crypto::secret_key_to_public_key(viewkey, pkey)) {
+          er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+          er.message = "failed to verify view key secret key";
+          return false;
+        }
+        
+    if (info.address.m_view_public_key != pkey) {
+      er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+      er.message = "view key does not match standard address";
+      return false;
+    }
+      
     if (m_wallet && req.autosave_current)
     {
       try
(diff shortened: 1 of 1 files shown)


StrawberryChocolateFudge:
The failing test seems to be:

[  FAILED  ] AddressFromURL.Failure (221 ms)
[----------] 2 tests from AddressFromURL (809 ms total)

It's not related to this change. The master I forked contains it.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3891,7 +3891,20 @@ namespace tools
...
+          er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+          er.message = "failed to verify view key secret key";
+          return false;
+        }
+        
+    if (info.address.m_view_public_key != pkey) {
+      er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+      er.message = "view key does not match standard address";
+      return false;
+    }
```
jeffro256, line 3906:
```suggestion
    hw::device &hwdev = hw::get_device("default");
    if (!hwdev.verify_keys(viewkey, info.address.m_view_public_key)) {
      er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
      er.message = "view secret key does not match main address";
      return false;
    }
```

ACK the idea. IMO this is a more succinct and clearer way to write the check. Also, the term "main addresses" should used instead of "standard address". Same comments go for the spend privkey check below.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3892,6 +3892,13 @@ namespace tools
       return false;
     }
 
+    hw::device &hwdev = hw::get_device("default");
+     if (!hwdev.verify_keys(viewkey, info.address.m_view_public_key)) {
```
jeffro256, line 3896: Wrong indentation level

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3918,6 +3925,14 @@ namespace tools
           er.message = "Failed to parse spend key secret key";
           return false;
         }
+
+      if (!hwdev.verify_keys(spendkey, info.address.m_spend_public_key)) {
```
jeffro256, line 3929: Wrong indentation level

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3918,6 +3925,14 @@ namespace tools
...
+          er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+          er.message = "spend secret key does not match main address";
+          return false;
+        }
+
+
```
jeffro256, line 3935: Extra newline

jeffro256 (changes requested):
Just want some whitespace changes, then it looks good to go

StrawberryChocolateFudge:
okay. got it!

nahuhh (approved):

jeffro256 (approved):

PR #9860, opened 2025-03-23
wallet: fix wallet_keys_unlocker

jeffro256:
Fix five issues with wallet_keys_unlocker:
1. It won't decrypt if there are unlockers open simulataneously on multiple `wallet2` instances
2. It won't decrypt if the first unlocker was disabled (i.e. `locked=false`), even with a second non-disabled unlocker
3. If a destructor of an earlier unlocker is triggered before the destructor of a later unlocker, it will re-encrypt too early, while the second unlocker is still in scope
4. We shouldn't attempt to encrypt/decrypt for any background wallets, even if not currently syncing
5. Calling the 3-parameter constructor with `locked=true` after an unlocker already exists will "double-encrypt"  the spend key after its destructor since the local variable is `locked=false` and the field member is `locked=true`

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -117,7 +117,7 @@ typedef cryptonote::simple_wallet sw;
   LOCK_IDLE_SCOPE(); \
   boost::optional<tools::password_container> pwd_container = boost::none; \
   if (m_wallet->ask_password() && !(pwd_container = get_and_verify_password())) { code; } \
-  tools::wallet_keys_unlocker unlocker(*m_wallet, pwd_container);
+  tools::wallet_keys_unlocker unlocker(*m_wallet, pwd_container ? &pwd_container->password() : nullptr);
 
 #define SCOPED_WALLET_UNLOCK() SCOPED_WALLET_UNLOCK_ON_BAD_PASSWORD(return true;)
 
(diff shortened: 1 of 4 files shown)


vtnerd (approved):

jeffro256:
Oops sorry I have minor change that I need to make.

jeffro256:
Okay two actually:
1. The condition on `m_background_syncing` was right. I forgot that this gets set in the non-background wallet when background syncing as well.
2. IN the previous commit, I generated the `password` into the `key` conditional on whether the current constructor of `wallet_keys_unlocker` decrypts. So for a non-concentric round of constructors/destructors, the wallet would re-encrypt with a garbage key on the latter destructor.

jeffro256:
Squashed.

selsta:
@vtnerd please re-review

jeffro256:
Removed stray newline + space

rbrunner7 (approved):
Looks good.

One question though, probably because I lack knowledge about some C++ subtlety: Why is the `unlocker` twice declared directly, and twice wrapped into a `std::optional`?

rbrunner7:
I might have the answer to my own question in the review comment: To be able to cleanly destroy the unlocker to force an early re-encryption?

jeffro256:
Yes, this is correct. I also could have used a `std::unique_ptr`, `std::shared_ptr`, a `std::vector`, etc.

PR #9875, opened 2025-03-28
wallet: fix monero-wallet-rpc ignoring calls during sync

Tzadiko:
## Description

A while ago, this PR was merged https://github.com/monero-project/monero/pull/8941 (@moneromooo-monero)

The consequence of this was that the monero-wallet-rpc would hang on auto-refreshes. This is an example sequence of events that causes the hang:

- Restore wallet from a low height (e.g. 2,500,000)
- Refresh event begins (kicked off every second)
- More than REFRESH_INFICATIVE_BLOCK_CHUNK_SIZE is fetched, m_last_auto_refresh_time is not updated; furthermore,
- Over 1000ms have passed since the beginning of the refresh event.
- Next refresh cycle is immediately kicked off.

This continuous refresh event, without stop, blocks the wallet-rpc from responding to other events.

 ### How to recreate issue on Master

1. Create a wallet, either programatically (monero-wallet-rpc) or manually (monero-wallet-cli)
2. Get the address, private spend key, and private view key.
3. Call [generate_from_keys](https://docs.getmonero.org/rpc-library/wallet-rpc/#generate_from_keys) with a restore_height of `2500000`
4. Try to hit the RPC with `get_version` (e.g. `curl -X POST http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"get_version"}' -H 'Content-Type: application/json'`), you'll notice it just hangs (make sure you have the correct host and port, I bound my monero-wallet-rpc to 18082).

## Change

This PR introduces a between 200 - 300ms evaluation period whereby the network thread is free, if we've non-stop processed blocks for over m_auto_refresh_period. This gives us a 'breathing room' period by which we can handle other rpc-requests.

Observe the timestamps below to understand the sequence of events. The image below is included when the evaluation period was 3 seconds. It's been lowered to the interval above (200 - 300ms).

image

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -33,6 +33,7 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/preprocessor/stringize.hpp>
 #include <cstdint>
+#include <chrono>
 #include "include_base_utils.h"
 using namespace epee;
 
@@ -60,7 +61,7 @@ using namespace epee;
 #define MONERO_DEFAULT_LOG_CATEGORY "wallet.rpc"
 
 #define DEFAULT_AUTO_REFRESH_PERIOD 20 // seconds
-#define REFRESH_INFICATIVE_BLOCK_CHUNK_SIZE 256    // just to split refresh in separate calls to play nicer with other threads
+#define REFRESH_INDICATIVE_BLOCK_CHUNK_SIZE 256    // just to split refresh in separate calls to play nicer with other threads
 
 #define CHECK_MULTISIG_ENABLED() \
   do \
@@ -198,24 +199,77 @@ namespace tools
   bool wallet_rpc_server::run()
   {
     m_stop = false;
-    m_net_server.add_idle_handler([this](){
-      if (m_auto_refresh_period == 0) // disabled
+
+    const bool enable_auto_refresh = m_auto_refresh_period != 0;
+    const auto auto_refresh_evaluation_ms = std::chrono::milliseconds(200);
+
+    m_net_server.add_idle_handler([=] { // Implicit capture of this-pointer deprecated in C++20.
+      if (!enable_auto_refresh) // disabled
         return true;
-      if (boost::posix_time::microsec_clock::universal_time() < m_last_auto_refresh_time + boost::posix_time::seconds(m_auto_refresh_period))
+
+      // Check if m_auto_refresh_period seconds have passed since the last refresh attempt.
+      const auto auto_refresh_interval_ms = std::chrono::milliseconds(m_auto_refresh_period * 1'000);
+      if (auto_refresh_interval_ms <= auto_refresh_evaluation_ms)
+      {
+        LOG_PRINT_L0((boost::format(tr("The auto wallet sync evaluation interval of %i ms must be larger than the refresh interval of %i ms"))
+          % auto_refresh_evaluation_ms.count()
+          % auto_refresh_interval_ms.count()).str());
         return true;
+      }
+
+      const auto now = std::chrono::steady_clock::now();
+      if (now < m_last_auto_refresh_time + auto_refresh_interval_ms)
+        return true;
+
       uint64_t blocks_fetched = 0;
-      try {
+      bool refresh_success = false;
+      const auto start = std::chrono::steady_clock::now();
+
+      try
+      {
         bool received_money = false;
-        if (m_wallet) m_wallet->refresh(m_wallet->is_trusted_daemon(), 0, blocks_fetched, received_money, true, true, REFRESH_INFICATIVE_BLOCK_CHUNK_SIZE);
-      } catch (const std::exception& ex) {
+        if (m_wallet) m_wallet->refresh(m_wallet->is_trusted_daemon(), 0, blocks_fetched, received_money, true, true, REFRESH_INDICATIVE_BLOCK_CHUNK_SIZE);
+        refresh_success = true;
+      }
+      catch (const std::exception& ex)
+      {
         LOG_ERROR("Exception at while refreshing, what=" << ex.what());
       }
-      // if we got the max amount of blocks, do not set the last refresh time, we did only part of the refresh and will
-      // continue asap, and only set the last refresh time once the refresh is actually finished
-      if (blocks_fetched < REFRESH_INFICATIVE_BLOCK_CHUNK_SIZE)
-        m_last_auto_refresh_time = boost::posix_time::microsec_clock::universal_time();
+
+      if (blocks_fetched == 0)
+        return true;
+
+      const auto end = std::chrono::steady_clock::now();
+      const auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+
+      if (refresh_success)
+        LOG_PRINT_L3((boost::format(tr("Automated wallet block refresh took %i ms")) % elapsed.count()).str());
+
+      const bool syncing_against_tip_of_chain = blocks_fetched < REFRESH_INDICATIVE_BLOCK_CHUNK_SIZE;
+      if (syncing_against_tip_of_chain)
+      {
+        // At this point, we can poll for a refresh every m_auto_refresh_period seconds.
+        m_last_auto_refresh_time = end;
+      }
+      else
+      {
+        // We are in a state of synchronization, blasting through the maximum chunks of blocks
+        // because we are not at the tip of the chain. In this case, if we update m_last_auto_refresh_time,
+        // we'll need to wait an entire m_refresh_interval_ms before processing the next batch. On the other hand,
+        // if we do not update m_last_auto_refresh_time, we'll never yield (other calls to the RPC will hang)
+        // in the case that elapsed > auto_refresh_evaluation_ms since we'll immediately be scheduled for another block sync.
+        const bool over_one_refresh_period_passed = end > m_last_auto_refresh_time + auto_refresh_interval_ms;
+        if (over_one_refresh_period_passed)
+        {
+          // auto_refresh_interval_ms of straight-blasting through blocks has elapsed without end.
+          // Let's freee up the network thread for between 200ms to 300ms (non-deterministic) to handle other requests.
+          const auto refresh_throttle = auto_refresh_evaluation_ms + std::chrono::milliseconds(100);
+          m_last_auto_refresh_time = end - auto_refresh_interval_ms + refresh_throttle;
+          LOG_PRINT_L3((boost::format(tr("Temporarily throttling wallet block refresh by around %i ms")) % refresh_throttle.count()).str());
+        }
+      }
       return true;
-    }, 1000);
+    }, auto_refresh_evaluation_ms.count());
     m_net_server.add_idle_handler([this](){
       if (m_stop.load(std::memory_order_relaxed))
       {
@@ -325,7 +379,8 @@ namespace tools
     } // end auth enabled
 
     m_auto_refresh_period = DEFAULT_AUTO_REFRESH_PERIOD;
-    m_last_auto_refresh_time = boost::posix_time::min_date_time;
+    const auto over_one_period_ago = std::chrono::steady_clock::now() - std::chrono::seconds(m_auto_refresh_period * 2);
+    m_last_auto_refresh_time = over_one_period_ago;
 
     check_background_mining();
 
--- a/src/wallet/wallet_rpc_server.h
+++ b/src/wallet/wallet_rpc_server.h
@@ -33,6 +33,7 @@
 #include <boost/program_options/options_description.hpp>
 #include <boost/program_options/variables_map.hpp>
 #include <string>
+#include <chrono>
 #include "common/util.h"
 #include "net/http_server_impl_base.h"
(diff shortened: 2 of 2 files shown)


SChernykh:
> This PR introduces a 3 * evaluation (timeout) period (i.e. 3 * 1000ms) whereby the network thread is free

Why 3 seconds if there are already other pending rpc requests? If you free the network thread, they will just kick in, and they will take whatever time they need. So you can throttle by 100 ms (for example), and it will be enough (other requests will kick in, take more than 100 ms, and then wallet refresh will resume immediately). Of course it needs testing.

Tzadiko:
If a request isn’t pending, then it will need to wait another 20 seconds
until it’s handled. With a 3 second breathing period, the odds that a
request that arrives is handled immediately increases. I don’t have a
scientific reason for 3 seconds over 2 or 4, I just spoke to ofrnxmr and
that’s around an acceptable time we discussed. Open to your idea as well,
but I just want to make the traders known.

nahuhh:
i agree with @SChernykh  that there doesnt need to be a waiting period. It should assume that the request(s) were already pending

src/wallet/wallet_rpc_server.cpp
```diff
@@ -198,24 +199,68 @@ namespace tools
...
+      const auto now = std::chrono::system_clock::now();
+      if (now < m_last_auto_refresh_time + auto_refresh_interval_ms)
         return true;
+
       uint64_t blocks_fetched = 0;
-      try {
+      bool refresh_success = false;
+      const auto start = std::chrono::system_clock::now();
+
+      try
```
nahuhh, line 219: Whitespace

src/wallet/wallet_rpc_server.cpp
```diff
@@ -198,24 +199,68 @@ namespace tools
...
+      const auto start = std::chrono::system_clock::now();
+
+      try 
+      {
         bool received_money = false;
-        if (m_wallet) m_wallet->refresh(m_wallet->is_trusted_daemon(), 0, blocks_fetched, received_money, true, true, REFRESH_INFICATIVE_BLOCK_CHUNK_SIZE);
-      } catch (const std::exception& ex) {
+        if (m_wallet) m_wallet->refresh(m_wallet->is_trusted_daemon(), 0, blocks_fetched, received_money, true, true, REFRESH_INDICATIVE_BLOCK_CHUNK_SIZE);
+        refresh_success = true;
+      }
```
nahuhh, line 224: Same

src/wallet/wallet_rpc_server.cpp
```diff
@@ -198,24 +199,68 @@ namespace tools
...
+      if (syncing_against_tip_of_chain)
+      {
+        // At this point, we can poll for a refresh every m_auto_refresh_period seconds.
+        m_last_auto_refresh_time = end;
+      }
+      else
+      {
+        // We are in a state of synchronization, blasting through the maximum chunks of blocks
+        // because we are not at the tip of the chain. In this case, if we update m_last_auto_refresh_time,
+        // we'll need to wait an entire m_refresh_interval_ms before processing the next batch. On the other hand,
```
nahuhh, line 249: Whitespace on end of line

src/wallet/wallet_rpc_server.cpp
```diff
@@ -198,24 +199,68 @@ namespace tools
   bool wallet_rpc_server::run()
   {
     m_stop = false;
-    m_net_server.add_idle_handler([this](){
-      if (m_auto_refresh_period == 0) // disabled
+
+    const bool enable_auto_refresh = m_auto_refresh_period != 0;
+    const auto auto_refresh_evaluation_ms = std::chrono::milliseconds(1'000);
```
nahuhh, line 204:
```suggestion
    const auto auto_refresh_evaluation_ms = std::chrono::milliseconds(10);
```

This works

src/wallet/wallet_rpc_server.cpp
```diff
@@ -198,24 +199,68 @@ namespace tools
...
+        // because we are not at the tip of the chain. In this case, if we update m_last_auto_refresh_time,
+        // we'll need to wait an entire m_refresh_interval_ms before processing the next batch. On the other hand, 
+        // if we do not update m_last_auto_refresh_time, we'll never yield (other calls to the RPC will hang)
+        // in the case that elapsed > auto_refresh_evaluation_ms since we'll immediately be scheduled for another block sync.
+        const bool over_one_refresh_period_passed = end > m_last_auto_refresh_time + auto_refresh_interval_ms;
+        if (over_one_refresh_period_passed)
+        {
+          // auto_refresh_interval_ms of straight-blasting through blocks has elapsed without end.
+          // Let's wait till proceeding for at least 3 timeout intervals, freeing us up to respond to other requests in that time.
+          const auto refresh_throttle = auto_refresh_evaluation_ms * 3;
```
nahuhh, line 257:
```suggestion
          // Let's interrupt for 1 evaluation interval, freeing us up to respond to pending requests.
          const auto refresh_throttle = auto_refresh_evaluation_ms;
```

Or can drop `refresh_throttle` and just point to `auto_refresh_evaluation_ms` directly

src/wallet/wallet_rpc_server.cpp
```diff
@@ -198,24 +199,68 @@ namespace tools
...
+          m_last_auto_refresh_time = end - auto_refresh_interval_ms + refresh_throttle;
+          LOG_PRINT_L3((boost::format(tr("Temporarily throttling wallet block refresh by %i ms")) % refresh_throttle.count()).str());
+        }
+      }
       return true;
-    }, 1000);
+    }, auto_refresh_evaluation_ms.count());
```
nahuhh, line 263: just a question: 
Should this one be evaluation or interval?

src/wallet/wallet_rpc_server.cpp
Tzadiko: The same can be achieved with the `refresh_throttle` without the context switching that comes with changing this interval poll.

Tzadiko:
> > This PR introduces a 3 * evaluation (timeout) period (i.e. 3 * 1000ms) whereby the network thread is free
> 
> Why 3 seconds if there are already other pending rpc requests? If you free the network thread, they will just kick in, and they will take whatever time they need. So you can throttle by 100 ms (for example), and it will be enough (other requests will kick in, take more than 100 ms, and then wallet refresh will resume immediately). Of course it needs testing.

Ok. Something important to note is that the timer is not deterministic. There is no 'guaranteed' refresh period of X milliseconds. The period will be between 100ms and evaluation_ms + 100ms. With evaluation_ms set at 1'000 (unchanged), that'll be between 100ms and 1100ms. The actual 'breathing' period will be between those bounds, depending on a myriad of factors. I can give you examples but I have to go. If what I described above is fine, I can leave it as is. If you want to go deeper into this interval we can talk about it more.

src/wallet/wallet_rpc_server.cpp
Tzadiko: It's the evaluation interval. Interval presumes that the refresh is guaranteed to happen, but all the 1000ms guarantees is that we'll test to see if the `refresh` is valid.

src/wallet/wallet_rpc_server.cpp
Tzadiko: `refresh_throttle` is still needed due to the 100ms addition discussed in other comments.

src/wallet/wallet_rpc_server.cpp
Tzadiko: Addressed

src/wallet/wallet_rpc_server.cpp
Tzadiko: Addressed

src/wallet/wallet_rpc_server.cpp
```diff
@@ -198,24 +199,68 @@ namespace tools
...
+        // because we are not at the tip of the chain. In this case, if we update m_last_auto_refresh_time,
+        // we'll need to wait an entire m_refresh_interval_ms before processing the next batch. On the other hand,
+        // if we do not update m_last_auto_refresh_time, we'll never yield (other calls to the RPC will hang)
+        // in the case that elapsed > auto_refresh_evaluation_ms since we'll immediately be scheduled for another block sync.
+        const bool over_one_refresh_period_passed = end > m_last_auto_refresh_time + auto_refresh_interval_ms;
+        if (over_one_refresh_period_passed)
+        {
+          // auto_refresh_interval_ms of straight-blasting through blocks has elapsed without end.
+          // Let's freee up the network thread for between 100ms to 1100ms (non-deterministic) to handle other requests.
+          const auto refresh_throttle = auto_refresh_evaluation_ms + std::chrono::milliseconds(100);
```
nahuhh, line 257: This arbitrarily pauses the sync for an unnecessary/arbitrarily long amount of time (seems to be consistently ~=1100ms)

Any size interrupt will handle all pending rpc requests sequentially, before returning to the refresh. No need to wait 1 second between each round (and slow down the overall sync as a result) 

100ms isnt a long time, but i don't think there is any benefit to using a longer interval, and thereby suggest 10ms

```suggestion
          // Let's interrupt up the network thread for 10ms to handle other requests.
          const auto refresh_throttle = std::chrono::milliseconds(10);
```

nahuhh (changes requested):
pause is too long

src/wallet/wallet_rpc_server.cpp
```diff
@@ -198,24 +199,77 @@ namespace tools
...
+      const auto auto_refresh_interval_ms = std::chrono::milliseconds(m_auto_refresh_period * 1'000);
+      if (auto_refresh_interval_ms <= auto_refresh_evaluation_ms)
+      {
+        LOG_PRINT_L0((boost::format(tr("The auto wallet sync evaluation interval of %i ms must be lager than the refresh interval of %i ms"))
+          % auto_refresh_evaluation_ms.count()
+          % auto_refresh_interval_ms.count()).str());
+        return true;
+      }
+
+      const auto now = std::chrono::system_clock::now();
```
iamamyth, line 220: This kind of thing needs a monotonic clock (`steady_clock`), not the wall clock, as you want to measure elapsed time, not deltas between snapshots of the system clock, and they're not at all the same thing.

Tzadiko: Addressed

nahuhh (approved):
Tested

0xFFFC0000 (approved):
Looks correct. Thank you.

plowsof:
@Tzadiko can you clear my confusion please. 

>The consequence of this was that the monero-wallet-rpc would hang on auto-refreshes. This is an example sequence of events that causes the hang:

the `--no-initial-sync` flag starts the wallet rpc server before sync begins. During wallet sync i am able to query things:

`./monero-wallet-rpc --rpc-bind-port 18082 --disable-rpc-login --wallet-file /home/human/somedir/helloworld --daemon-address node3.monerodevs.org:18089 --password "" --no-initial-sync` 

```
2025-04-24 20:27:52.376	W Loading wallet...e
2025-04-24 20:27:54.893	I caching ringdb key
2025-04-24 20:27:54.910	W Loaded wallet keys file, with public address: 4BFdYYzfaJSXy6J2HabkJqWfQSpsUKhKASPWBv4K6xF7LQTbactKYm3RSUzKsb17e9USMuFKEyCh2gEYWLWwqXkYTV2ekjH
2025-04-24 20:27:54.916	I Trying to decrypt cache data
2025-04-24 20:27:54.964	I No message store file found: /home/human/dev/monero/build/Linux/master/release/bin/wallet-rpc-created-me.mms
2025-04-24 20:27:54.964	I Successfully loaded
2025-04-24 20:27:54.964	D Using an untrusted daemon, skipping background mining check
2025-04-24 20:27:54.965	I Set server type to: 1 from name: RPC, prefix_name = RPC
2025-04-24 20:27:54.965	I Binding on 127.0.0.1 (IPv4):18082
2025-04-24 20:27:54.965	I Generating SSL certificate
2025-04-24 20:27:55.988	D start accept (IPv4)
2025-04-24 20:27:55.988	D Spawned connection #0 to 0.0.0.0 currently we have sockets count:1
2025-04-24 20:27:55.988	W Starting wallet RPC server
```
the wallet rprc will now respond to rpc requests e.g. `get_version` .. and wallets sync height. 

So my question is: are you aware of this? and is this a `--no-initial-sync` alternative / better method which is also enabled by default? this makes me wonder why the flag would be optional in #8941, is there actually a scenario where people would not want this? (i can't think of one)

Tzadiko:
Also, don’t start your sync at block 0, start at like 2.75 million. There
was virtually no activity at inception so the issue doesn’t manifest.

Tzadiko:
Also, try syncing from the block height detailed in the reproduction steps
to see the issue. Syncing from block 0 where there was no activity doesn’t
reproduce the issue because there’s no load.

plowsof:
> 3\. 2500000

Ah sorry, i now see with `--no-initial-sync` @ a restore height 2.5~mil that the rpc server hangs even though it has 'started' (whilst the wallet is busy syncing). This would explain the seemingly random timeouts ive had over the months/year(s) with `--no-initial-sync` , thank you.

plowsof (approved):

PR #9876, opened 2025-03-28
CMakeLists: add cache option for MANUAL_SUBMODULES

jeffro256:
Adds a line in the CMakeCache.txt of the build dir to edit `MANUAL_SUBMODULES`. Convenient for debugging a submodule's integration with Monero codebase when you don't want to pass the def on the command line each time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -360,6 +360,7 @@ else()
   message(STATUS "Building without build tag")
 endif()
 
+option(MANUAL_SUBMODULES "Skip submodule up-to-date checks" OFF)
 if(NOT MANUAL_SUBMODULES)
   find_package(Git)
   if(GIT_FOUND)


Tzadiko:
Can you provide sample usage for the CMake noobs out there (me), please? I find this useful, but I never got the -DMANUAL_SUBMODULES=1 option to work.

selsta:
@Tzadiko 

you have to do this for it to work

```
mkdir build && cd build
cmake -D MANUAL_SUBMODULES=1 ..
make -j8
```

after this patch you can manually edit the `CMakeCache.txt` file that gets created after building, so that you don't have to call cmake itself manually.

selsta (approved):

tobtoht (approved):

0xFFFC0000 (approved):

PR #9882, opened 2025-04-01
unit_tests: add wallet_scanning

jeffro256:
This change adds offline unit testing for `wallet2::process_parsed_blocks()`, the main driver of the chain-processing logic in `wallet2`. The transactions are constructed using `cryptonote::construct_tx_and_get_tx_key`, which makes them realistic to pruned transactions that appear in the wild. The `fake_pruned_blockchain` maintains lists of `cryptonote::block_complete_entry` and `tools::wallet2::parsed_block` structs for each block, with easy miner transaction adding, automatic assigning of unique output indices for outputs in the block, and correct blockID hashing. The blockchain can also start at an offset, but if it starts at 0, then it uses the real genesis block for that network type.

The provided unit test creates one of each transaction type a-h, adds it to the fake blockchain, then calls `wallet2::process_parsed_blocks()` on the block data. After each update, it checks that the balance increases according to the amount in the relevant transaction output. 

Part of upstreaming Carrot.

### DIFF
--- a/tests/unit_tests/wallet_scanning.cpp
+++ b/tests/unit_tests/wallet_scanning.cpp
@@ -1,0 +1,748 @@
+// Copyright (c) 2023-2024, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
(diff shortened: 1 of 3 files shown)


tests/unit_tests/wallet_scanning.cpp
```diff
@@ -0,0 +1,748 @@
...
+};
+//----------------------------------------------------------------------------------------------------------------------
+//----------------------------------------------------------------------------------------------------------------------
+static cryptonote::tx_source_entry gen_tx_source_entry_fake_members(
+    const stripped_down_tx_source_entry_t &in,
+    const size_t mixin,
+    const uint64_t max_global_output_index)
+{
+    const size_t ring_size = mixin + 1;
+    const bool is_rct = in.mask == rct::I;
```
selsta, line 133:
```suggestion
    const bool is_rct = in.mask != rct::I;
```

tests/unit_tests/wallet_scanning.cpp
```diff
@@ -0,0 +1,748 @@
...
+
+    // g. scan view-tagged pre-ringct (only possible in unmixable sweep txs) tx
+    balance_diff = wallet_process_new_blocks();
+    EXPECT_EQ(amount_g, balance_diff);
+
+    // h. construct and push a view-tagged ringct tx
+    const rct::xmr_amount amount_h = rct::randXmrAmount(COIN);
+    {
+        const rct::xmr_amount fee = rct::randXmrAmount(COIN);
+        std::vector<cryptonote::tx_destination_entry> dests = {
```
selsta, line 726: HF15 tx needs two outputs

tests/unit_tests/wallet_scanning.cpp
```diff
@@ -0,0 +1,748 @@
...
+            break;
+        case HF_VERSION_BULLETPROOF_PLUS:
+        case 16:
+            rct_config = { rct::RangeProofPaddedBulletproof, 4 };
+            break;
+        default:
+            ASSERT_MES_AND_THROW("unrecognized hf version");
+    }
+    const bool use_view_tags = hf_version >= HF_VERSION_VIEW_TAGS;
+    const size_t mixin = 15;
```
selsta, line 226:
```suggestion
    const size_t mixin = sweep_unmixable_override ? 0 : 15;
```

otherwise it can't create unmixable sweep.

selsta (commented):
How can we get this out of draft?

PR #9884, opened 2025-04-02
cryptonote_basic: is_coinbase() take prefix

jeffro256:
Useful for wallet business logic, which stores the *prefixes* in `m_transfers`, not the whole transaction.

Part of upstreaming Carrot


vtnerd (approved):

vtnerd:
The dreaded race condition test failed. :/ Since this is cmake only, everything should be fine.

tobtoht (approved):

PR #9886, opened 2025-04-02
Fix cmake usage in FetchContent and ExternalProject

vtnerd:
The CMake configure stage fails if another projects uses the Monero project in `FetchContent` or `ExternalProject` due to an incorrect path. This fixes the path so that it is relative to the top-level `project(monero)` declaration, instead of the top-level CMake directory.

Slightly longer explanation: I am working on the [`lwsf`](https://github.com/vtnerd/lwsf) library which implements the `wallet2_api.h`, but targets a LWS-rpc server instead of `monerod`-rpc server. Ideally this project gets optionally built into the [`monero_c`](https://github.com/MrCyjaneK/monero_c) project for use in dart, typescript, and C#. Both projects need to use the libraries within the `monero` project, and instead of building twice, `FetchContent` allows the projects to coordinate configuring+building the `monero` project once.

### DIFF
--- a/cmake/CheckLinkerFlag.cmake
+++ b/cmake/CheckLinkerFlag.cmake
@@ -6,7 +6,7 @@ macro(CHECK_LINKER_FLAG flag VARIABLE)
       message(STATUS "Looking for ${flag} linker flag")
     endif()
 
-    set(_cle_source ${CMAKE_SOURCE_DIR}/cmake/CheckLinkerFlag.c)
+    set(_cle_source ${monero_SOURCE_DIR}/cmake/CheckLinkerFlag.c)
 
     set(saved_CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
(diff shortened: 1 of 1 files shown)


MrCyjaneK (approved):
I've been using `CMAKE_CURRENT_SOURCE_DIR` in here for quite some time in monero_c, this points to the same location as `monero_SOURCE_DIR`

tobtoht (approved):

jeffro256:
I'm not a CMake expert by any means, but where is `monero_SOURCE_DIR` defined? Doing a `git grep` for it doesn't show it mentioned anywhere else in the repo. Also, would the [`PROJECT_SOURCE_DIR`](https://cmake.org/cmake/help/latest/variable/PROJECT_SOURCE_DIR.html) variable is a good fit here?

jeffro256 (approved):

0xFFFC0000 (approved):

PR #9898, opened 2025-04-07
Move wallet2_api definitions into separate cpp

vtnerd:
This moves constructors+destructors definitions that are declared in `wallet2_api.h` into a separate cpp file instead of scattering them into other implementation files. The advantage is that the static linker can identify dead object files more easily, and strip out tons of code that isn't used downstream.

The primary example is the executable [`lwsf_ledger`](https://github.com/vtnerd/lwsf) which goes from ~17M to ~5.7M on Ubuntu 22.04 and from ~12M to ~3.7M on macOS. Basically, without this change the linker includes the entire `wallet2` implementation into `lwsf_ledger` even though that executable is deliberately only using the `lwsf` implementation of the API.

Since this is just moving definitions into a different cpp, I don't expect and disruptions to other projects. Note that _link time_ dependencies are still large; reducing that requires a separate library target which I didn't include because lwsf currently still (basically) has all the link-time dependencies.

I also made a `virtual` `~WalletManager` because it really ought to have one, even though I think all GUIs just "leak" this memory since its used for the lifetime of the app.

### DIFF
--- a/src/wallet/api/CMakeLists.txt
+++ b/src/wallet/api/CMakeLists.txt
@@ -30,7 +30,11 @@
 
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
+# A new library with just `wallet2_api.cpp` could be added to reduce link time
+# depdencies. Just compile the cpp twice in each library, its small
+
 set(wallet_api_sources
+  wallet2_api.cpp
   wallet.cpp
   wallet_manager.cpp
   transaction_info.cpp
--- a/src/wallet/api/address_book.cpp
+++ b/src/wallet/api/address_book.cpp
@@ -38,9 +38,7 @@
 #include <vector>
 
 namespace Monero {
-  
-AddressBook::~AddressBook() {}
-  
+
 AddressBookImpl::AddressBookImpl(WalletImpl *wallet)
     : m_wallet(wallet), m_errorCode(Status_Ok) {}
 
--- a/src/wallet/api/pending_transaction.cpp
+++ b/src/wallet/api/pending_transaction.cpp
@@ -46,9 +46,6 @@ using namespace std;
 
 namespace Monero {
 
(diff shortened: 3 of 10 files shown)


vtnerd:
The race condition test again. Ping me if you want a dummy force push to retry the unit tests (they technically aborted).

jeffro256:
The `lwsf_ledger` link is broken btw

vtnerd:
@jeffro256 I'm not sure, as I would have to dig deeper into the associated linkers. Best I can tell, is that using an object file drags in vtable(s), and then from there the code backing those vtables never gets dropped as it could potentially be used.

I came to this conclusion because calling a `static` function in `Monero:Wallet` will drag in megabytes of code, even though the bulk of the code could be dropped.

vtnerd:
I'm probably going to close this as I found another way - at least for `monero-lws`. So I will leave this open for now while I investigate what can be done for `lwsf`.

`gcc`/`clang` emit 1+ function tables per object file, defaulting to 1 per object file. If one function is referenced in the table, then all functions from that table are pulled into the final binary. So moving these functions changed which object file table contained them.

There is an option for both compilers - `-ffunction-sections -fdata-sections` - which emits a table per entry. This blows up the object file sizes but at the link stage the option `-Wl,--gc-sections` will typically emit dramatically smaller sizes. On `monero-lws`, I was able to reduce the docker image by ~14 MiB (which has two executables within it). There isn't a downside (other than object size increases) as the runtime performance is identical and the function names for stack traces are still present.

Note the `monero-lws` docker image was using static executables. It's a little trickier with dynamic objects as the dynamic object has to assume that _any_ function could be used.

Also note, using dynamic libs didn't shrink the `monero-lws` docker container (despite having two executables) as the entirety of several boost libs were pulled in, when only a subset of the functions were needed.

vtnerd:
Yup, lwsf related projects can reduce their footprint by using the gcc/clang related build options. This helps with binary size even if `wallet2` is included.

PR #9899, opened 2025-04-08
epee: better scope leaver

jeffro256:
1. Doesn't allocate a `boost::shared_ptr`
2. Doesn't use a vtable
3. Doesn't use `std::function` wrapper
4. Doesn't necessarily copy functional object
5. Doesn't need a factory function for simple declaration (yay CTAD)
6. Isolated to its own header
7. Name doesn't suck as much

Depends on #9860 to keep the diff small and conceptually tight.

### DIFF
--- a/contrib/epee/include/scope_guard.h
+++ b/contrib/epee/include/scope_guard.h
@@ -1,0 +1,99 @@
+// Copyright (c) 2025, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <utility>
+
+namespace epee
+{
+
+/**
+ * @brief: scope_guard - RAII pattern to call action on scope leave
+ */
+template <class F>
+class scope_guard final
+{
+public:
+    // no copy
+    scope_guard(const scope_guard&) = delete;
+    scope_guard& operator=(const scope_guard&) = delete;
+
+    // constructor with CTAD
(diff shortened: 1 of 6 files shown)


src/wallet/wallet2.cpp
```diff
@@ -14114,11 +14096,12 @@ void wallet2::stop_background_sync(const epee::wipeable_string &wallet_password,
 
   // Set the plaintext spend key
   m_account.set_spend_key(recovered_spend_key);
+  const bool should_reencrypt_keys = m_ask_password == AskPasswordToDecrypt && !m_unattended && !m_watch_only;
```
iamamyth, line 14099: This predicate seems common enough that you might consider introducing an inline for it.

jeffro256: I agree. I did make a shorthand in the FCMP++/Carrot downstream branch, maybe I should upstream it in #9860

jeffro256: Okay done

jeffro256:
I forgot that, according to the C++ standard, type aliases should never have their template types deduced, so the alias of `auto_scope_leaver_caller` to `scope_guard` was invalid, causing compilation to fail on Debian and MacOS. So for now, to keep the PR small, I elected to alias `auto_scope_leaver_caller` to the concrete type of `scope_guard<std::function<void()>>`. The uses of `auto_scope_leaver_caller` throughout the codebase can be refactored out later if it's worth the effort.

src/wallet/wallet2.cpp
```diff
@@ -5808,27 +5811,6 @@ void wallet2::restore(const std::string& wallet_, const epee::wipeable_string& p
   }
 }
 //----------------------------------------------------------------------------------------------------
-epee::misc_utils::auto_scope_leave_caller wallet2::decrypt_account_for_multisig(const epee::wipeable_string &password)
```
vtnerd, line 5811: So this deleted chunk here was basically just duplicated code?

vtnerd (approved):

contrib/epee/include/scope_guard.h
```diff
@@ -0,0 +1,57 @@
...
+template <class F>
+class scope_guard final
+{
+public:
+    // no copy
+    scope_guard(const scope_guard&) = delete;
+    scope_guard& operator=(const scope_guard&) = delete;
+
+    // constructor with CTAD
+    scope_guard(F &&f): m_f(std::forward<F>(f)) {}
```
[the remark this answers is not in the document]
jeffro256: True. Would it be better to leave as-is and force callers to explicitly use `std::move` with an lvalue, or add a `const&` constructor? I'm leaning towards the former to force them to think about moving/copying their lvalues, and we can always add the second constructor later.

src/wallet/wallet2.cpp
jeffro256: Yeah, basically this code does the same thing as `wallet_keys_unlocker`

contrib/epee/include/scope_guard.h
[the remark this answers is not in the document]
vtnerd: I think this is fine "as-is", it seems rare that a lambda would need to be copied given the purpose.

vtnerd (approved):

jeffro256:
Rebased without changes against https://github.com/monero-project/monero/pull/9860 to show passing tests.

jeffro256:
Rebased since #9860 was merged.

iamamyth (approved):

tobtoht:
Builds failing.

jeffro256:
Rebased, added `unique_scope_guard`, and added unit tests

contrib/epee/include/scope_guard.h
```diff
@@ -0,0 +1,99 @@
...
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <utility>
```
selsta, line 31:
```suggestion
#include <memory>
#include <utility>
```

contrib/epee/include/scope_guard.h
```diff
@@ -0,0 +1,99 @@
...
+    // move resets current state and releases other instance
+    unique_scope_guard(unique_scope_guard &&other): m_f(std::move(other.m_f)), m_do(other.m_do)
+    {
+        other.release();
+    }
+    unique_scope_guard& operator=(unique_scope_guard &&other)
+    {
+        if (this != std::addressof(other))
+        {
+            reset();
```
selsta, line 77: Is it okay if this throws here?

jeffro256: It should be. `this->m_do` is set to `false` still, so `this->m_f` isn't triggered again. And since the control flow ends here, `other` isn't touched, and will thus still trigger its destructor if it goes out of scope.

selsta (approved):

PR #9901, opened 2025-04-10
Daemon RPC: add max_block_count field to /getblocks.bin [v0.18]

j-berman:
#9381

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -710,12 +710,13 @@ namespace cryptonote
         }
       }
 
-      size_t max_blocks = COMMAND_RPC_GET_BLOCKS_FAST_MAX_BLOCK_COUNT;
+      size_t max_blocks = req.max_block_count > 0
+        ? std::min(req.max_block_count, (uint64_t)COMMAND_RPC_GET_BLOCKS_FAST_MAX_BLOCK_COUNT)
+        : COMMAND_RPC_GET_BLOCKS_FAST_MAX_BLOCK_COUNT;
+
       if (m_rpc_payment)
       {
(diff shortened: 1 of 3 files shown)


selsta:
What is the use case for having this and the other things in the release branch?

j-berman:
For this PR in particular:

1) The [async scanner](https://github.com/UkoeHB/monero/pull/23) (or anyone developing something similar) could be released at a later time, pointing to a daemon already running this change. The more daemons running this change, the more likely it would be for wallets to connect to a daemon that supports the async scanner's improved scanning UX.

2) Wallet devs can tweak params in their wallets to fetch a smaller number of blocks if they want shorter intervals for scanning progress feedback to users today. @sethforprivacy has mentioned an interest in this for Cake Wallet in the past.

The first applies to the other PR's also (#9902, #9903).

The second only applies for this PR.

At this rate with FCMP++, I don't know if I will be able to complete the async scanner in time for a release before the fork. So, I don't have a strong opinion on merging the other PR's before the async scanner is ready. This PR I think has a reasonable case to be included sooner though.

kayabaNerve:
I wanted to fetch blocks `x ..= y` using `start_height`, `max_block_count`. Unfortunately, I cannot and this was never merged into a release despite being merged into master 11 months ago.

`get_blocks_by_height.bin` is insufficient as it returns TXs, but not pruned, nor outputs indexes (as for some reason, `get_blocks_by_height.bin` and `get_blocks.bin` have different results?). This would mean I have to fetch the proofs needlessly, output indexes separately, at which point there's effectively no purpose to using the `get_blocks*.bin` methods?

I would greatly appreciate this being merged immediately.

selsta (approved):

PR #9910, opened 2025-04-24
src: fix clangd warnings

0xFFFC0000:
Fix clangd missing headers and warnings. 

* No semantical change to the code.


SyntheticBird45 (approved):

0xFFFC0000:
``` 
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/o/openssh/openssh-client_8.2p1-4ubuntu0.13_amd64.deb  404  Not Found 

```

Failure is not related to PR.

SyntheticBird45 (approved):

src/serialization/tuple.h
```diff
@@ -31,6 +31,7 @@
 #pragma once
 #include <tuple>
 #include "serialization.h"
+#include "cstdint"
```
SChernykh, line 34: Why not `<cstdint>` here?

0xFFFC0000: Thanks. Good catch. This will be fixed in next push in an hour or so.

0xFFFC0000: Addressed in the new push.

jeffro256 (approved):

src/crypto/hash-extra-blake.c
```diff
@@ -34,5 +34,5 @@
 #include "blake256.h"
 
 void hash_extra_blake(const void *data, size_t length, char *hash) {
-  blake256_hash((uint8_t*)hash, data, length);
+  blake256_hash((uint8_t*)hash, (const uint8_t *)data, (uint64_t)length);
```
vtnerd, line 37: `limits<size_t>::max() <= limits<uint64_t>::max()` is not guaranteed. Unfortunately static asserts in C is a bit messy, so I don't know what to do here.

src/cryptonote_basic/tx_extra.h
```diff
@@ -30,6 +30,14 @@
 
 #pragma once
 
+#include <cstdint>
+#include <cstddef>
+
+#include <boost/variant.hpp>
```
vtnerd, line 36: Can you get away with just `<boost/variant/variant.hpp>` here? The header file with `VARIANT_TAG` macros should arguably have any further headers needed.

src/crypto/hash-extra-blake.c
jeffro256: What's wrong with including `assert.h` and using `static_assert`? https://en.cppreference.com/w/c/language/_Static_assert and https://en.cppreference.com/w/c/error/static_assert.

0xFFFC0000: Something like this might work: 

```
  #if SIZE_MAX > UINT64_MAX
    blake256_hash(hash, data, length);
  #else
    blake256_hash((uint8_t*)hash, (const uint8_t *)data, (uint64_t)length);
  #endif  

```

But I removed casting-related changes since I want this to get merged ASAP. I am only keeping the casts that both side are exactly same size.

src/cryptonote_basic/tx_extra.h
0xFFFC0000: Thanks it worked. Although I still need `#include "serialization/variant.h"`

vtnerd (approved):

PR #9915, opened 2025-04-28
remove cached password from Wallet API [release]

SNeedlewoods:
Edit: Update description.
_______
This PR is based on #9464, which is based on the `release-v0.18` branch so it can be used by monero-gui. Besides removing `m_password` from `WalletImpl` the following changes were made:
- ~~add method
  `bool WalletImpl::open(const std::string &path, const epee::wipeable_string &password)`
  AFAIU the old method `WalletImpl::open(const std::string &path, const std::string &password)` is not defined in `src/wallet/api/wallet2_api.h`, because it's only used by `WalletManagerImpl::openWallet()` ([src](https://github.com/monero-project/monero/blob/b6df4bb14264d19bb274090a02293de41fd709af/src/wallet/api/wallet_manager.cpp#L69)), which is supposed to be used by Wallet API clients.~~
- add password as optional argument (it's only needed if you provide a new path):
  `bool WalletImpl::store(const std::string &path, const optional<const char *> &password /* = optional<const char *>() */)`
- **Breaking change:** we now have to provide an `old_password` argument to set a new password, I took this opportuninty to also change `std::string` to `char *` as suggested here https://github.com/monero-project/monero/pull/9915#discussion_r2067046719
  `bool WalletImpl::setPassword(const char *old_password, const char *new_password)`
- **Breaking change:** remove method
  `const std::string& WalletImpl::getPassword()`
- **Breaking change:** add password as non-optional argument to the following methods:
    - `string WalletImpl::makeMultisig(const vector<string>& info, const uint32_t threshold, const char *password)`
    - `std::string WalletImpl::exchangeMultisigKeys(const std::vector<std::string> &info, const char *password, const bool force_update_use_with_caution /*= false*/)`
- The following change is included, because I simultaneously removed the cached password in monero-gui and we need a way to verify a password entered by the user. I think due to the limitations of [this WalletManager method](https://github.com/monero-project/monero/blob/b6df4bb14264d19bb274090a02293de41fd709af/src/wallet/api/wallet2_api.h#L1315-L1327) it's okay to add a new `WalletImpl` method.
  `bool WalletImpl::verifyPassword(const char *password, std::uint64_t kdf_rounds /* = 1 */)`

### DIFF
--- a/src/wallet/api/wallet2_api.h
+++ b/src/wallet/api/wallet2_api.h
@@ -31,14 +31,17 @@
 #pragma once
 
 
-#include <string>
-#include <vector>
-#include <list>
-#include <set>
+#include <cstdint>
 #include <ctime>
 #include <iostream>
+#include <list>
+#include <map>
+#include <memory>
+#include <set>
 #include <stdexcept>
-#include <cstdint>
+#include <string>
+#include <unordered_set>
+#include <vector>
 
 //  Public interface for libwallet library
 namespace Monero {
@@ -66,6 +69,39 @@ enum NetworkType : uint8_t {
         T t;
         bool set;
     };
+
+/**
+* brief: EnoteDetails - Container for all the necessary information that belongs to an enote
+*/
+struct EnoteDetails
+{
+    enum TxProtocol {
+        TxProtocol_CryptoNote,
+        TxProtocol_RingCT
+    };
+
+    virtual ~EnoteDetails() = 0;
+    virtual std::string onetimeAddress() const = 0;
+    virtual std::string viewTag() const = 0;
+    virtual std::uint64_t blockHeight() const = 0;
+    virtual std::string txId() const = 0;
+    virtual std::uint64_t internalEnoteIndex() const = 0;
+    virtual std::uint64_t globalEnoteIndex() const = 0;
+    virtual bool isSpent() const = 0;
+    virtual bool isFrozen() const = 0;
+    virtual std::uint64_t spentHeight() const = 0;
+    virtual std::string keyImage() const = 0;
+    virtual std::string mask() const = 0;
+    virtual std::uint64_t amount() const = 0;
+    virtual TxProtocol protocolVersion() const = 0;
+    virtual bool isKeyImageKnown() const = 0;
+    virtual bool isKeyImageRequest() const = 0;
+    virtual std::uint64_t pkIndex() const = 0;
+    virtual std::vector<std::pair<std::uint64_t, std::string>> uses() const = 0;
+
+    // Multisig
+    virtual bool isKeyImagePartial() const = 0;
+};
 
 /**
  * @brief Transaction-like interface for sending money
@@ -126,6 +162,12 @@ struct PendingTransaction
      * @return vector of base58-encoded signers' public keys
      */
     virtual std::vector<std::string> signersKeys() const = 0;
+    /**
+    * brief: convertTxToStr - convert PendingTransaction to hex string (same content which would be saved to file with commit(filename))
+    * return: unsigned tx data as encrypted hex string if succeeded, else empty string
+    * note: sets status error on fail
+    */
+    virtual std::string convertTxToStr() = 0;
 };
 
 /**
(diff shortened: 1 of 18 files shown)


src/wallet/api/wallet2_api.h
```diff
@@ -31,14 +31,18 @@
...
+#include <list>
+#include <map>
+#include <memory>
+#include <optional>
+#include <set>
 #include <stdexcept>
-#include <cstdint>
+#include <unordered_set>
+
+#include "wipeable_string.h"
```
vtnerd, line 45: I understand the desire for this change, but the is the first time that something outside of C++ standard libraries is required by the `wallet2_api.h` file. All alternative implementations must link against `libepee` for this to work.

I'm not certain this small incremental improvement is worth changing this API because all of the UIs (monero-gui, stack, cake) are likely using constructs in the GUI that aren't wiped. OTOH, I guess this will help make the GUIs more compliant in this regard? I don't know, I don't see it.

I'll do a larger view later.

j-berman: Alternative approach: using a `char *` instead. I think a pointer would give more flexibility to the caller to handle wiping on their end without requiring linking against something outside of C++ standard libs, and then internally the API lib could instantiate a `wipeable_string` from the `char *`.

GUI's that want to make a stronger attempt at keeping sensitive data out of memory could even use `wipeable_string` themselves (linking against epee themselves if they want), then call `<their string>.data()` and pass that in to the API (though yes, this doesn't address UI elems handling memory securely).

> because all of the UIs (monero-gui, stack, cake) are likely using constructs in the GUI that aren't wiped

@SNeedlewoods has done some looking into on qt's memory handling. I think would be worth sharing what you found here sneedle.

j-berman: Also for a mobile app, I *think* they can do something like this: combine the password input to the UI with a distinct constant randomly generated salt stored in the user's keystore, so even if the UI password alone is compromised, there is an extra layer of defense because the wallet password is a combo of the user-input password and easier-to-securely-handle salt. I thought I recalled that Monerujo did something like this, but the link to their paper on their password is dead and don't want to dig through the code at the moment (can get back to this).

SNeedlewoods: FWIW, the GUI stores the text for `LineEdit` components as `QString`. To manually wipe those, I added `wipeQStr()` ([src](https://github.com/monero-project/monero-gui/pull/4437/files#diff-e09c19b9b11620d78d5d00b4436d9abd198862581f9c16d65d53b69ea7ae9f19R34)), which can be called from .qml files.
_____
Did some more testing with GUI wallet with a memscanner first and then with a debugger, I could not find leaked passwords in memory with the change from `epee::wipeable_string` to `std::string` in the Wallet API.

Btw. I decided to use `std::string &` instead of `char *` to be more aligned with other methods in the API. If you still think a `char *` would be better I'll change it.

I'm not sure my tests are enough to confirm this behaves the same on other systems (I'm on an ubuntu like OS), so I thought about doing some extra wiping.
We could
A) do it in the Wallet API implementation,
```
void WalletImpl::someFunc(std::string &password)
{
    m_wallet->some_func(epee::wipeable_string(password));
    memwipe(&password[0], password.size());
}
```
but I think this isn't ideal, if the caller wants to use the password for something else. Though you could use a `bool do_wipe` argument as done [here](https://github.com/monero-project/monero-gui/pull/4437/files#diff-d2527cdb10ab23dcefd89476bf1700ea420099f3556faf656925791bb341f88cR1095-R1101). I've done it that way in the GUI, because most of the time we want to wipe the password after verifying it as it is not needed anymore, but there is one case ([src](https://github.com/monero-project/monero-gui/pull/4437/files#diff-9992c84951a9e49a9adaf58d8b04beb57ba879fb6f110ece70883fb5cccbb425R210-R213)) where we still need the password after verifying.
Or 
B) just leave it and let the clients have to do it in a similar way, e.g. the GUI
```
void Wallet::someFunc(const QString &password)
{
    std::string password_string = password.toStdString();
    m_walletImpl->someFunc(password_string);
    memwipe(&password_string[0], password_string.size());
}
```

vtnerd:
One thing I didn't catch before - why is this being done to the release 0.18 branch? The master branch makes sense for next release, but I don't see why we want to muck with the existing release code?

SNeedlewoods:
Put this PR to WIP, this is awfully outdated. Will focus on #10232 first and update this PR once the `master` is somewhat settled.

PR #9916, opened 2025-04-28
msys2 build: include int headers

jeffro256:
For `uint*_t`, otherwise some downstream MSYS2 builds break.


vtnerd (approved):

selsta:
@jeffro256 there is another build issue, is something else missing?

jeffro256:
@selsta Looks like it's passing now

0xFFFC0000 (approved):

PR #9917, opened 2025-04-29
threadpool: fix simple form of starvation

jeffro256:
Instead of a waiter executing all jobs until the queue is completely empty, the waiter will execute jobs until there are no more jobs assigned to that waiter.

Might help assuage some concerns brought up in #9768

### DIFF
--- a/src/common/threadpool.cpp
+++ b/src/common/threadpool.cpp
@@ -78,7 +78,7 @@ void threadpool::create(unsigned int max_threads) {
   size_t i = max ? max - 1 : 0;
   running = true;
   while(i--) {
-    threads.push_back(boost::thread(attrs, boost::bind(&threadpool::run, this, false)));
+    threads.push_back(boost::thread(attrs, boost::bind(&threadpool::run, this, nullptr)));
   }
 }
 
@@ -129,11 +129,16 @@ threadpool::waiter::~waiter()
 }
 
 bool threadpool::waiter::wait() {
-  pool.run(true);
+  pool.run(this);
   boost::unique_lock<boost::mutex> lock(mt);
   while(num)
     cv.wait(lock);
   return !error();
+}
+
+int threadpool::waiter::get_num() {
+  const boost::lock_guard lock(mt);
+  return num;
 }
 
 void threadpool::waiter::inc() {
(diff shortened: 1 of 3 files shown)


jeffro256:
A more aggressive approach would be to iterate over the queue, pick out the jobs assigned to that waiter, and run those out of order.

jeffro256:
Included a unit test which hangs forever with the old code.

j-berman:
This would likely benefit the current wallet2 scanner.

Here's the current order of ops:

1) [Make request for first chunk of blocks](https://github.com/monero-project/monero/blob/c5cb8e1b783ee1410b3feeee4bb0c22252722fca/src/wallet/wallet2.cpp#L4093).
2) [Wait on the response](https://github.com/monero-project/monero/blob/c5cb8e1b783ee1410b3feeee4bb0c22252722fca/src/wallet/wallet2.cpp#L4132).
3) [Submit the next request to the threadpool](https://github.com/monero-project/monero/blob/c5cb8e1b783ee1410b3feeee4bb0c22252722fca/src/wallet/wallet2.cpp#L4093).
4) [Start processing first chunk of blocks](https://github.com/monero-project/monero/blob/c5cb8e1b783ee1410b3feeee4bb0c22252722fca/src/wallet/wallet2.cpp#L4099C11-L4099C32).

In step 4, the first call to `wait()` is [this one](https://github.com/monero-project/monero/blob/c5cb8e1b783ee1410b3feeee4bb0c22252722fca/src/wallet/wallet2.cpp#L3240), which is a very light operation in the scanning flow.

That call to `wait()` also has to wait on step 3 (i.e. the response for the next chunk of blocks).

Wallets significantly bottlenecked by bandwidth (mobile wallets) therefore have to wait on the response *and then process it*. So they don't benefit from the parallel scanning, they pay for both the network request and then client-side scanning.

j-berman:
> A more aggressive approach would be to iterate over the queue, pick out the jobs assigned to that waiter, and run those out of order.

Untested, but I think in order for the wallet2 scanner to benefit, we may actually need this more aggressive approach. Otherwise it looks step 4 would still have to wait for step 3 to complete, since step 3's task is submitted to the threadpool before step 4's.

I think it may make sense to switch out the threadpool instance used for network requests with `getInstanceForIO` as a solution to this general problem of step 4 waiting on step 3, but I figure the above is worth noting too.

Also worth noting the currently planned scanner changes for FCMP++ would benefit from the more aggressive approach to pick out jobs assigned to a specific waiter.

jeffro256:
Rebased

jpk68 (approved):

PR #9921, opened 2025-05-08
error out if functional tests are disabled

moneromooo-monero:
The intent is that you add "-D EXPECT_FUNCTIONAL_TESTS=1" to the targets that are run from CI, or for release preparation, so those tests aren't suddenly silently skipped if a dep is not found (which happened to me in TF because someone added a new python dep to monero).

By default, tests will still be silently skipped.


tests/functional_tests/CMakeLists.txt
```diff
@@ -78,4 +78,7 @@ if (REQUESTS_OUTPUT STREQUAL "OK")
 else()
   message(WARNING "functional_tests_rpc and check_missing_rpc_methods skipped, needs the 'requests', 'psutil', 'monotonic', 'zmq', and 'deepdiff' python modules")
   set(CTEST_CUSTOM_TESTS_IGNORE ${CTEST_CUSTOM_TESTS_IGNORE} functional_tests_rpc check_missing_rpc_methods)
+  if (EXPECT_FUNCTIONAL_TESTS)
+    message(FATAL_ERROR "Functional tests are expected")
```
jeffro256, line 82:
```suggestion
    message(FATAL_ERROR "Functional tests are expected, but aforementioned python dependencies are not found")
```

Just in case they wouldn't understand why it doesn't work...

moneromooo-monero:
Since it just basically paraphrases the previous error(s), not worth the time IMHO.

0xFFFC0000 (approved):
I approve. But @jeffro256 mentioned comment is not a bad idea either: 

https://github.com/monero-project/monero/pull/9921/files#r2080664207

Anyway, approved. Both way is fine.

Update: Nevermind. I think it is fine. We already print that info as mentioned by @moneromooo-monero  https://github.com/monero-project/monero/pull/9921#issuecomment-3155584563

PR #9926, opened 2025-05-15
Boost: vendor serialization impls for C++17 types

jeffro256:
Includes Boost serialization implementations for `std::optional` and `std::variant`. Neither are available until Boost 1.84. Additionally, there was a [namespace clash](https://github.com/boostorg/serialization/issues/328) between Boost `serialization` and `multiprecision` libraries that caused Monero to fail compilation. A [fix](https://github.com/boostorg/multiprecision/pull/680) was upstreamed to Boost `multiprecision`, but it probably won't be available until 1.89. So in the meantime, these two shim files will allow us to serialize `std::optional` and `std::variant` without bumping the minimum required Boost version.

Part of upstreaming Carrot/FCMP++.


jeffro256:
Closing as "not needed" with #9940

Issue #9928, opened 2025-05-18
monerod should not ban local IP address


0xFFFC0000:
monerod should not block/ban local IP addresses [1]. This regularly happens if there is a intensive use of RPC, e.g. in local test scenarios. This fixes it. 

````
diff --git a/src/p2p/net_node.inl b/src/p2p/net_node.inl
index 8da16ec72..948f8739c 100644
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -175,6 +175,12 @@ namespace nodetool
   template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::is_remote_host_allowed(const epee::net_utils::network_address &address, time_t *t)
   {
+    if (address.is_local())
+    {
+      MWARNING("Address is local " << address.host_str() << ", local IPs are allowed.");
+      return true;
+    }
+
     CRITICAL_REGION_LOCAL(m_blocked_hosts_lock);
 
     const time_t now = time(nullptr);

````

1. https://github.com/monero-project/monero/blob/125622d5bdc42cf552be5c25009bd9ab52c0a7ca/src/p2p/net_node.inl#L176

nahuhh:
Onion/i2p etc tunnels are treated as local

0xFFFC0000:
> Onion/i2p etc tunnels are treated as local
> 
> edit: nvm that would still be kind of silly to ban local ip, as it would ban all onion/i2p traffic unilaterally

What you first said is important IMHO. 

with that 3 lines I am basically preventing any ban. So if onion/i2p IPs are treated as local. Then we are basically not going to ban any onioin/i2p IP. Which is not what we want.

PR #9929, opened 2025-05-18
p2p: prevent banning and restricting local machine IPs

0xFFFC0000:
https://github.com/monero-project/monero/issues/9928


nahuhh:
I think a better approach is to prevent banning on the unrestricted rpc

SyntheticBird45:
I would 100% prefer this to be behind a flag and opt-in.

selsta:
Continuing in #11090.

Issue #9334, opened 2024-05-20
[Proposal] Change how transactions are broadcasted to significantly reduce P2P bandwidth usage 


Boog900:
monerod uses dandelion++ to propagate txs around the network, this happens in 2 stages: stem and fluff. This proposal will only affect txs in the fluff stage.

When monerod receives a fluff tx, it will queue it for broadcasting to every peer except the one that sent it to us first, it doesn't matter if the peer(s) need it or not.

Because monerod sends the whole tx blob this is extremely inefficient, as the majority of the time the peer already has the tx.

## The Proposal

Instead of sending the whole tx-blob just send the hash and allow the peer to request the tx if they need it. This requires 2 new P2P messages:  `TxPoolInv`, `RequestTxPoolTxs` (names not important for now).

Adding support for these P2P messages should be done before changing how txs are broadcasted.

### `TxPoolInv`

This message will contain a list of new tx hashes and will be used when fluffing txs instead of `NewTransactions` which is currently used. 

When a node receives this it will request the txs with `RequestTxPoolTxs` if they are not in the tx-pool.

### `RequestTxPoolTxs`

This message will also contain a list of tx hashes. When a peer receives this message it should look in the _public_ tx pool and respond with as many txs as possible in a `NewTransactions` message.

A peer should not be punished for not responding to a `RequestTxPoolTxs` as the tx(s) could have been dropped from the peers pool.

## How inefficient is the current method?

Using the same technique as was done in the [Erlay paper](https://arxiv.org/abs/1905.10518) we can work out the expected number of wasted broadcasts: 

Each tx is sent _at least_ once along each connection, most of time it's more. If the number of nodes in the Monero network is N then the number of links is 12N as each node by default makes 12 outbound connections, then the number of redundant broadcasts is 12N - N = 11N.

So this means that ~90% of current tx broadcasts are redundant. Although we can't drop the 90% number without introducing a protocol like Erlay, we can redefine what a broadcast is. This would also lay the ground work for potentially adding Erlay in the future although that is not what this proposal is for.

To validate this I ran some tests on a local, fully synced, Monero node (monerod), the node was set to keep 12 outbound connections and 0 inbound.

In 2 hours 30 minutes I collected around 250MBs of data, in which there were 96268 tx broadcasts.

When you send a tx to a peer, if the peer doesn't also send the tx you know you were the first to send that tx to them, if they do send the tx as well you were not the first to send it to them.

[...]

PR #9933, opened 2025-05-22
Tx Relay v2: New Relay Logic with threshold-based peer dropping

0xFFFC0000:
## Tx Relay v2: Enhanced Transaction Propagation Protocol https://github.com/monero-project/monero/issues/9334

This was made possible thanks to @Boog900’s support in designing and implementing each step.

#### 1. Inventory-Based Transaction Announcements
**Protocol Messages:**
```cpp
NOTIFY_TX_POOL_INV (ID 2011):
- Broadcasts transaction hashes instead of full TXs
- Contains vector of crypto::hash entries

NOTIFY_REQUEST_TX_POOL_TXS (ID 2012):
- Requests specific TXs by hash
- Enables selective fetching of missing transactions
```

**Workflow**:

- Node advertises new TX hashes via `NOTIFY_TX_POOL_INV`

- Receivers check local mempool & request missing TXs with `NOTIFY_REQUEST_TX_POOL_TXS`

- Only requested transactions are transmitted

**Bandwidth Savings**:

- Reduces initial broadcast size by ~99% (hash vs full TX)

#### Peer Reliability System
Peer Scoring Metrics (`peerinfo_manager`):

- Tracked per connection:
  * Announcements received
  * Successful TX deliveries
  * Missed requests
  * Request/response ratios
  * Timestamps for temporal analysis

- Threshold-based eviction:
  if (missed_requests / total_announcements) > 70%: disconnect_peer()
    
**Failure Detection**:

- Tracks 3 consecutive failures for any TX request

- Uses sliding window (last 30 requests) for fairness

- Penalizes peers for:

  * Unresponsive behavior

  * Invalid transaction responses

#### Components:
```
[Request Manager] <-handles-> [TX Request Queue]
[TX Request Queue] <-tracks-> [Peer UUIDs]
[TX Handler] <-manages-> [Timeouts]
```

#### Configuration

```cpp
 // cryptonote_config.h
#define P2P_DEFAULT_REQUEST_TIMEOUT 5      // Seconds
#define P2P_REQUEST_FAILURE_THRESHOLD 70   // Percentage
```

### DIFF
--- a/src/cryptonote_basic/connection_context.cpp
+++ b/src/cryptonote_basic/connection_context.cpp
@@ -30,47 +30,9 @@
 #include "connection_context.h"
 
 #include <boost/optional/optional.hpp>
-#include "cryptonote_protocol/cryptonote_protocol_defs.h"
-#include "p2p/p2p_protocol_defs.h"
 
 namespace cryptonote
 {
-  std::size_t cryptonote_connection_context::get_max_bytes(const int command) noexcept
-  {
-    switch (command)
-    {
-    case nodetool::COMMAND_HANDSHAKE_T<cryptonote::CORE_SYNC_DATA>::ID:
-      return 65536;
-    case nodetool::COMMAND_TIMED_SYNC_T<cryptonote::CORE_SYNC_DATA>::ID:
-      return 65536;
-    case nodetool::COMMAND_PING::ID:
-      return 4096;
-    case nodetool::COMMAND_REQUEST_SUPPORT_FLAGS::ID:
-      return 4096;
-    case cryptonote::NOTIFY_NEW_BLOCK::ID:
-      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
-    case cryptonote::NOTIFY_NEW_TRANSACTIONS::ID:
-      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
-    case cryptonote::NOTIFY_REQUEST_GET_OBJECTS::ID:
-      return 1024 * 1024 * 2; // 2 MB
-    case cryptonote::NOTIFY_RESPONSE_GET_OBJECTS::ID:
-      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
-    case cryptonote::NOTIFY_REQUEST_CHAIN::ID:
-      return 512 * 1024; // 512 kB
-    case cryptonote::NOTIFY_RESPONSE_CHAIN_ENTRY::ID:
-      return 1024 * 1024 * 4; // 4 MB
-    case cryptonote::NOTIFY_NEW_FLUFFY_BLOCK::ID:
-      return 1024 * 1024 * 4; // 4 MB, but it does not includes transaction data
-    case cryptonote::NOTIFY_REQUEST_FLUFFY_MISSING_TX::ID:
-      return 1024 * 1024; // 1 MB
-    case cryptonote::NOTIFY_GET_TXPOOL_COMPLEMENT::ID:
-      return 1024 * 1024 * 4; // 4 MB
-    default:
-      break;
-    };
-    return std::numeric_limits<size_t>::max();
-  }
-
   void cryptonote_connection_context::set_state_normal()
   {
     m_state = state_normal;
--- a/src/cryptonote_basic/connection_context.cpp
+++ b/src/cryptonote_basic/connection_context.cpp
@@ -29,47 +29,9 @@
 #include "connection_context.h"
 
 #include <boost/optional/optional.hpp>
-#include "cryptonote_protocol/cryptonote_protocol_defs.h"
-#include "p2p/p2p_protocol_defs.h"
 
 namespace cryptonote
 {
-  std::size_t cryptonote_connection_context::get_max_bytes(const int command) noexcept
-  {
-    switch (command)
-    {
-    case nodetool::COMMAND_HANDSHAKE_T<cryptonote::CORE_SYNC_DATA>::ID:
-      return 65536;
-    case nodetool::COMMAND_TIMED_SYNC_T<cryptonote::CORE_SYNC_DATA>::ID:
-      return 65536;
-    case nodetool::COMMAND_PING::ID:
-      return 4096;
-    case nodetool::COMMAND_REQUEST_SUPPORT_FLAGS::ID:
-      return 4096;
-    case cryptonote::NOTIFY_NEW_BLOCK::ID:
-      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
-    case cryptonote::NOTIFY_NEW_TRANSACTIONS::ID:
-      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
-    case cryptonote::NOTIFY_REQUEST_GET_OBJECTS::ID:
-      return 1024 * 1024 * 2; // 2 MB
-    case cryptonote::NOTIFY_RESPONSE_GET_OBJECTS::ID:
-      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
-    case cryptonote::NOTIFY_REQUEST_CHAIN::ID:
-      return 512 * 1024; // 512 kB
-    case cryptonote::NOTIFY_RESPONSE_CHAIN_ENTRY::ID:
-      return 1024 * 1024 * 4; // 4 MB
-    case cryptonote::NOTIFY_NEW_FLUFFY_BLOCK::ID:
-      return 1024 * 1024 * 4; // 4 MB, but it does not includes transaction data
-    case cryptonote::NOTIFY_REQUEST_FLUFFY_MISSING_TX::ID:
-      return 1024 * 1024; // 1 MB
-    case cryptonote::NOTIFY_GET_TXPOOL_COMPLEMENT::ID:
-      return 1024 * 1024 * 4; // 4 MB
-    default:
-      break;
-    };
-    return std::numeric_limits<size_t>::max();
-  }
-
   void cryptonote_connection_context::set_state_normal()
   {
     m_state = state_normal;
--- a/src/cryptonote_basic/connection_context.h
+++ b/src/cryptonote_basic/connection_context.h
@@ -32,14 +32,59 @@
 #pragma once
 #include <unordered_set>
 #include <atomic>
-#include <algorithm>
+#include <memory>
+#include <shared_mutex>
 #include <boost/date_time/posix_time/posix_time.hpp>
 #include <boost/optional/optional_fwd.hpp>
-#include "net/net_utils_base.h"
 #include "crypto/hash.h"
+#include "cryptonote_protocol/cryptonote_protocol_defs.h"
+#include "net/net_utils_base.h"
+#include "p2p/p2p_protocol_defs.h"
+#include "syncobj.h"
 
 namespace cryptonote
 {
+  //! \return Maximum number of bytes permissible for `command`.
+  constexpr std::size_t get_command_max_bytes(const int command) noexcept
+  {
+    switch (command)
+    {
+    case nodetool::COMMAND_HANDSHAKE_T<cryptonote::CORE_SYNC_DATA>::ID:
+      return 65536;
+    case nodetool::COMMAND_TIMED_SYNC_T<cryptonote::CORE_SYNC_DATA>::ID:
+      return 65536;
+    case nodetool::COMMAND_PING::ID:
+      return 4096;
+    case nodetool::COMMAND_REQUEST_SUPPORT_FLAGS::ID:
+      return 4096;
+    case cryptonote::NOTIFY_NEW_BLOCK::ID:
+      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
+    case cryptonote::NOTIFY_NEW_TRANSACTIONS::ID:
+      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
+    case cryptonote::NOTIFY_REQUEST_GET_OBJECTS::ID:
+      return 1024 * 1024 * 2; // 2 MB
+    case cryptonote::NOTIFY_RESPONSE_GET_OBJECTS::ID:
+      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
+    case cryptonote::NOTIFY_REQUEST_CHAIN::ID:
+      return 512 * 1024; // 512 kB
+    case cryptonote::NOTIFY_RESPONSE_CHAIN_ENTRY::ID:
+      return 1024 * 1024 * 4; // 4 MB
+    case cryptonote::NOTIFY_NEW_FLUFFY_BLOCK::ID:
+      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though, fluffy blocks can be full)
+    case cryptonote::NOTIFY_REQUEST_FLUFFY_MISSING_TX::ID:
+      return 1024 * 1024; // 1 MB
+    case cryptonote::NOTIFY_GET_TXPOOL_COMPLEMENT::ID:
+      return 1024 * 1024 * 4; // 4 MB
+    case cryptonote::NOTIFY_TX_POOL_HASH::ID:
+      return 1024 * 1024 * 2; // 2 MB
+    case cryptonote::NOTIFY_REQUEST_TX_POOL_TXS::ID:
+      return 1024 * 1024 * 2; // 2 MB
+    default:
+      break;
+    };
+    return std::numeric_limits<size_t>::max();
+  }
+
   struct cryptonote_connection_context: public epee::net_utils::connection_context_base
   {
     cryptonote_connection_context(): m_state(state_before_handshake), m_remote_blockchain_height(0), m_last_response_height(0),
@@ -88,9 +133,6 @@ namespace cryptonote
 
     static constexpr int handshake_command() noexcept { return 1001; }
     bool handshake_complete() const noexcept { return m_state != state_before_handshake; }
-
-    //! \return Maximum number of bytes permissible for `command`.
-    static size_t get_max_bytes(int command) noexcept;
 
     //! Use this instead of `m_state = state_normal`.
     void set_state_normal();
--- a/src/cryptonote_basic/connection_context.h
+++ b/src/cryptonote_basic/connection_context.h
@@ -32,14 +32,59 @@
 #pragma once
 #include <unordered_set>
 #include <atomic>
-#include <algorithm>
+#include <memory>
+#include <shared_mutex>
 #include <boost/date_time/posix_time/posix_time.hpp>
 #include <boost/optional/optional_fwd.hpp>
-#include "net/net_utils_base.h"
 #include "crypto/hash.h"
+#include "cryptonote_protocol/cryptonote_protocol_defs.h"
+#include "net/net_utils_base.h"
+#include "p2p/p2p_protocol_defs.h"
+#include "syncobj.h"
 
 namespace cryptonote
 {
+  //! \return Maximum number of bytes permissible for `command`.
+  constexpr std::size_t get_command_max_bytes(const int command) noexcept
+  {
+    switch (command)
+    {
+    case nodetool::COMMAND_HANDSHAKE_T<cryptonote::CORE_SYNC_DATA>::ID:
+      return 65536;
+    case nodetool::COMMAND_TIMED_SYNC_T<cryptonote::CORE_SYNC_DATA>::ID:
+      return 65536;
+    case nodetool::COMMAND_PING::ID:
+      return 4096;
+    case nodetool::COMMAND_REQUEST_SUPPORT_FLAGS::ID:
+      return 4096;
+    case cryptonote::NOTIFY_NEW_BLOCK::ID:
+      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
+    case cryptonote::NOTIFY_NEW_TRANSACTIONS::ID:
+      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
+    case cryptonote::NOTIFY_REQUEST_GET_OBJECTS::ID:
+      return 1024 * 1024 * 2; // 2 MB
+    case cryptonote::NOTIFY_RESPONSE_GET_OBJECTS::ID:
+      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though)
+    case cryptonote::NOTIFY_REQUEST_CHAIN::ID:
+      return 512 * 1024; // 512 kB
+    case cryptonote::NOTIFY_RESPONSE_CHAIN_ENTRY::ID:
+      return 1024 * 1024 * 4; // 4 MB
+    case cryptonote::NOTIFY_NEW_FLUFFY_BLOCK::ID:
+      return 1024 * 1024 * 128; // 128 MB (max packet is a bit less than 100 MB though, fluffy blocks can be full)
+    case cryptonote::NOTIFY_REQUEST_FLUFFY_MISSING_TX::ID:
+      return 1024 * 1024; // 1 MB
+    case cryptonote::NOTIFY_GET_TXPOOL_COMPLEMENT::ID:
+      return 1024 * 1024 * 4; // 4 MB
+    case cryptonote::NOTIFY_TX_POOL_HASH::ID:
+      return 1024 * 1024 * 2; // 2 MB
+    case cryptonote::NOTIFY_REQUEST_TX_POOL_TXS::ID:
+      return 1024 * 1024 * 2; // 2 MB
+    default:
+      break;
+    };
+    return std::numeric_limits<size_t>::max();
+  }
+
   struct cryptonote_connection_context: public epee::net_utils::connection_context_base
   {
     cryptonote_connection_context(): m_state(state_before_handshake), m_remote_blockchain_height(0), m_last_response_height(0),
@@ -90,7 +135,10 @@ namespace cryptonote
     bool handshake_complete() const noexcept { return m_state != state_before_handshake; }
 
     //! \return Maximum number of bytes permissible for `command`.
-    static size_t get_max_bytes(int command) noexcept;
+    static constexpr std::size_t get_max_bytes(const int command) noexcept
+    {
+      return get_command_max_bytes(command);
+    }
 
     //! Use this instead of `m_state = state_normal`.
     void set_state_normal();
--- a/src/cryptonote_config.h
+++ b/src/cryptonote_config.h
@@ -145,6 +145,7 @@
 #define P2P_DEFAULT_PEERS_IN_HANDSHAKE                  250
 #define P2P_MAX_PEERS_IN_HANDSHAKE                      250
 #define P2P_DEFAULT_CONNECTION_TIMEOUT                  5000       //5 seconds
+#define P2P_DEFAULT_REQUEST_TIMEOUT                     (P2P_DEFAULT_CONNECTION_TIMEOUT*6) // 30 seconds
 #define P2P_DEFAULT_SOCKS_CONNECT_TIMEOUT               45         // seconds
 #define P2P_DEFAULT_PING_CONNECTION_TIMEOUT             2000       //2 seconds
 #define P2P_DEFAULT_INVOKE_TIMEOUT                      60*2*1000  //2 minutes
@@ -154,6 +155,8 @@
 #define P2P_DEFAULT_SYNC_SEARCH_CONNECTIONS_COUNT       2
 #define P2P_DEFAULT_LIMIT_RATE_UP                       8192       // kB/s
 #define P2P_DEFAULT_LIMIT_RATE_DOWN                     32768       // kB/s
+#define P2P_REQUEST_FAILURE_THRESHOLD_PERCENTAGE        70          // if more than 70% of requests fail, the peer is dropped
+#define P2P_MIN_SAMPLE_SIZE_FOR_DROPPING                5          // minimum number of requests to consider dropping a peer for failed requests
 
 #define P2P_FAILED_ADDR_FORGET_SECONDS                  (60*60)     //1 hour
 #define P2P_IP_BLOCKTIME                                (60*60*24)  //24 hour
@@ -161,7 +164,8 @@
 #define P2P_IDLE_CONNECTION_KILL_INTERVAL               (5*60) //5 minutes
 
 #define P2P_SUPPORT_FLAG_FLUFFY_BLOCKS                  0x01
-#define P2P_SUPPORT_FLAGS                               P2P_SUPPORT_FLAG_FLUFFY_BLOCKS
+#define P2P_SUPPORT_FLAG_TX_RELAY_V2                    0x02
+#define P2P_SUPPORT_FLAGS                               (P2P_SUPPORT_FLAG_FLUFFY_BLOCKS | P2P_SUPPORT_FLAG_TX_RELAY_V2)
 
 #define RPC_IP_FAILS_BEFORE_BLOCK                       3
 
--- a/src/cryptonote_config.h
+++ b/src/cryptonote_config.h
@@ -144,6 +144,7 @@
 #define P2P_DEFAULT_PEERS_IN_HANDSHAKE                  250
 #define P2P_MAX_PEERS_IN_HANDSHAKE                      250
 #define P2P_DEFAULT_CONNECTION_TIMEOUT                  5000       //5 seconds
+#define P2P_DEFAULT_REQUEST_TIMEOUT                     (P2P_DEFAULT_CONNECTION_TIMEOUT*6) // 30 seconds
 #define P2P_DEFAULT_SOCKS_CONNECT_TIMEOUT               45         // seconds
 #define P2P_DEFAULT_PING_CONNECTION_TIMEOUT             2000       //2 seconds
 #define P2P_DEFAULT_INVOKE_TIMEOUT                      60*2*1000  //2 minutes
@@ -153,6 +154,8 @@
 #define P2P_DEFAULT_SYNC_SEARCH_CONNECTIONS_COUNT       2
 #define P2P_DEFAULT_LIMIT_RATE_UP                       8192       // kB/s
 #define P2P_DEFAULT_LIMIT_RATE_DOWN                     32768       // kB/s
+#define P2P_REQUEST_FAILURE_THRESHOLD_PERCENTAGE        70          // if more than 70% of requests fail, the peer is dropped
+#define P2P_MIN_SAMPLE_SIZE_FOR_DROPPING                5          // minimum number of requests to consider dropping a peer for failed requests
 
 #define P2P_FAILED_ADDR_FORGET_SECONDS                  (60*60)     //1 hour
 #define P2P_IP_BLOCKTIME                                (60*60*24)  //24 hour
@@ -160,7 +163,8 @@
 #define P2P_IDLE_CONNECTION_KILL_INTERVAL               (5*60) //5 minutes
 
 #define P2P_SUPPORT_FLAG_FLUFFY_BLOCKS                  0x01
-#define P2P_SUPPORT_FLAGS                               P2P_SUPPORT_FLAG_FLUFFY_BLOCKS
+#define P2P_SUPPORT_FLAG_TX_RELAY_V2                    0x02
+#define P2P_SUPPORT_FLAGS                               (P2P_SUPPORT_FLAG_FLUFFY_BLOCKS | P2P_SUPPORT_FLAG_TX_RELAY_V2)
 
 #define RPC_IP_FAILS_BEFORE_BLOCK                       3
 
--- a/src/cryptonote_protocol/cryptonote_protocol_defs.h
+++ b/src/cryptonote_protocol/cryptonote_protocol_defs.h
@@ -34,6 +34,8 @@
 #include "serialization/keyvalue_serialization.h"
 #include "cryptonote_basic/cryptonote_basic.h"
 #include "cryptonote_basic/blobdatatype.h"
+#include "byte_slice.h"
+#include "storages/portable_storage_template_helper.h"
 
 namespace cryptonote
 {
@@ -207,6 +209,39 @@ namespace cryptonote
     };
     typedef epee::misc_utils::struct_init<request_t> request;
   };
+
+  struct tx_request_padding
+  {
+    void operator()(NOTIFY_NEW_TRANSACTIONS::request& request) const
+    {
+      size_t bytes = 9 /* header */ + 4 /* 1 + 'txs' */ + tools::get_varint_data(request.txs.size()).size();
+      for(auto tx_blob_it = request.txs.begin(); tx_blob_it!=request.txs.end(); ++tx_blob_it)
+        bytes += tools::get_varint_data(tx_blob_it->size()).size() + tx_blob_it->size();
+
+      // stuff some dummy bytes in to stay safe from traffic volume analysis
+      static constexpr const size_t granularity = 1024;
+      size_t padding = granularity - bytes % granularity;
+      const size_t overhead = 2 /* 1 + '_' */ + tools::get_varint_data(padding).size();
+      if (overhead > padding)
+        padding = 0;
+      else
+        padding -= overhead;
+      request._ = std::string(padding, ' ');
+
+      epee::byte_slice arg_buff;
+      epee::serialization::store_t_to_binary(request, arg_buff);
+
+      // we probably lowballed the payload size a bit, so added a but too much. Fix this now.
+      size_t remove = arg_buff.size() % granularity;
+      if (remove > request._.size())
+        request._.clear();
+      else
+        request._.resize(request._.size() - remove);
+      // if the size of _ moved enough, we might lose byte in size encoding, we don't care
+    }
+  };
+  inline constexpr tx_request_padding pad_tx_request{};
+
   /************************************************************************/
   /*                                                                      */
   /************************************************************************/
@@ -373,6 +408,50 @@ namespace cryptonote
         KV_SERIALIZE_CONTAINER_POD_AS_BLOB(hashes)
       END_KV_SERIALIZE_MAP()
     };
+    typedef epee::misc_utils::struct_init<request_t> request;
+  };
+
+  /************************************************************************
+  * Announces new transaction hashes that                                 *
+  * the sender believes the receiver may not have.                        *
+  * The receiver can pull which hashes are missing locally                *
+  * and optionally request the actual serialized transactions for them.   *
+  *************************************************************************/
+  struct NOTIFY_TX_POOL_HASH
+  {
+    const static int ID = BC_COMMANDS_POOL_BASE + 11;
+
+    struct request_t
+    {
+      std::vector<crypto::hash> t;
+
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE_CONTAINER_POD_AS_BLOB(t)
+      END_KV_SERIALIZE_MAP()
+    };
+
+    typedef epee::misc_utils::struct_init<request_t> request;
+  };
+
+  /************************************************************************
+  * Requests the actual transaction data corresponding                    *
+  * to a set of transaction hashes.                                       *
+  * The receiver should look up each transaction in its own pool          *
+  * or memory and respond with the serialized version, if known.          *
+  *************************************************************************/
+  struct NOTIFY_REQUEST_TX_POOL_TXS
+  {
+    const static int ID = BC_COMMANDS_POOL_BASE + 12;
+
+    struct request_t
+    {
+      std::vector<crypto::hash> t;
+
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE_CONTAINER_POD_AS_BLOB(t)
+      END_KV_SERIALIZE_MAP()
+    };
+
     typedef epee::misc_utils::struct_init<request_t> request;
   };
     
--- a/src/cryptonote_protocol/cryptonote_protocol_defs.h
+++ b/src/cryptonote_protocol/cryptonote_protocol_defs.h
@@ -32,8 +32,11 @@
 
 #include "crypto/hash.h"
 #include "cryptonote_basic/blobdatatype.h"
+#include "byte_slice.h"
+#include "common/varint.h"
 #include "misc_language.h"
 #include "serialization/keyvalue_serialization.h"
+#include "storages/portable_storage_template_helper.h"
 
 #include <list>
 
@@ -209,6 +212,39 @@ namespace cryptonote
     };
     typedef epee::misc_utils::struct_init<request_t> request;
   };
+
+  struct tx_request_padding
+  {
+    void operator()(NOTIFY_NEW_TRANSACTIONS::request& request) const
+    {
+      size_t bytes = 9 /* header */ + 4 /* 1 + 'txs' */ + tools::get_varint_data(request.txs.size()).size();
+      for(auto tx_blob_it = request.txs.begin(); tx_blob_it!=request.txs.end(); ++tx_blob_it)
+        bytes += tools::get_varint_data(tx_blob_it->size()).size() + tx_blob_it->size();
+
+      // stuff some dummy bytes in to stay safe from traffic volume analysis
+      static constexpr const size_t granularity = 1024;
+      size_t padding = granularity - bytes % granularity;
+      const size_t overhead = 2 /* 1 + '_' */ + tools::get_varint_data(padding).size();
+      if (overhead > padding)
+        padding = 0;
+      else
+        padding -= overhead;
+      request._ = std::string(padding, ' ');
+
+      epee::byte_slice arg_buff;
+      epee::serialization::store_t_to_binary(request, arg_buff);
+
+      // we probably lowballed the payload size a bit, so added a but too much. Fix this now.
+      size_t remove = arg_buff.size() % granularity;
+      if (remove > request._.size())
+        request._.clear();
+      else
+        request._.resize(request._.size() - remove);
+      // if the size of _ moved enough, we might lose byte in size encoding, we don't care
+    }
+  };
+  inline constexpr tx_request_padding pad_tx_request{};
+
   /************************************************************************/
   /*                                                                      */
   /************************************************************************/
@@ -375,6 +411,50 @@ namespace cryptonote
         KV_SERIALIZE_CONTAINER_POD_AS_BLOB(hashes)
       END_KV_SERIALIZE_MAP()
     };
+    typedef epee::misc_utils::struct_init<request_t> request;
+  };
+
+  /************************************************************************
+  * Announces new transaction hashes that                                 *
+  * the sender believes the receiver may not have.                        *
+  * The receiver can pull which hashes are missing locally                *
+  * and optionally request the actual serialized transactions for them.   *
+  *************************************************************************/
+  struct NOTIFY_TX_POOL_HASH
+  {
+    const static int ID = BC_COMMANDS_POOL_BASE + 11;
+
+    struct request_t
+    {
+      std::vector<crypto::hash> t;
+
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE_CONTAINER_POD_AS_BLOB(t)
+      END_KV_SERIALIZE_MAP()
+    };
+
+    typedef epee::misc_utils::struct_init<request_t> request;
+  };
+
+  /************************************************************************
+  * Requests the actual transaction data corresponding                    *
+  * to a set of transaction hashes.                                       *
+  * The receiver should look up each transaction in its own pool          *
+  * or memory and respond with the serialized version, if known.          *
+  *************************************************************************/
+  struct NOTIFY_REQUEST_TX_POOL_TXS
+  {
+    const static int ID = BC_COMMANDS_POOL_BASE + 12;
+
+    struct request_t
+    {
+      std::vector<crypto::hash> t;
+
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE_CONTAINER_POD_AS_BLOB(t)
+      END_KV_SERIALIZE_MAP()
+    };
+
     typedef epee::misc_utils::struct_init<request_t> request;
   };
     
--- a/src/cryptonote_protocol/cryptonote_protocol_defs.h
+++ b/src/cryptonote_protocol/cryptonote_protocol_defs.h
@@ -203,11 +203,13 @@ namespace cryptonote
       std::vector<blobdata>   txs;
       std::string _; // padding
       bool dandelionpp_fluff; //zero initialization defaults to stem mode
+      uint64_t nonce; // if responding to NOTIFY_REQUEST_TX_POOL_TXS, includes nonce in resp
 
       BEGIN_KV_SERIALIZE_MAP()
         KV_SERIALIZE(txs)
         KV_SERIALIZE(_)
         KV_SERIALIZE_OPT(dandelionpp_fluff, true) // backwards compatible mode is fluff
+        KV_SERIALIZE_OPT(nonce, (uint64_t)0)
       END_KV_SERIALIZE_MAP()
     };
     typedef epee::misc_utils::struct_init<request_t> request;
@@ -448,9 +450,11 @@ namespace cryptonote
 
     struct request_t
     {
+      uint64_t n; // request nonce
       std::vector<crypto::hash> t;
 
       BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE(n)
         KV_SERIALIZE_CONTAINER_POD_AS_BLOB(t)
       END_KV_SERIALIZE_MAP()
     };
--- a/src/cryptonote_protocol/cryptonote_protocol_handler.h
+++ b/src/cryptonote_protocol/cryptonote_protocol_handler.h
@@ -49,8 +49,10 @@
 #include "cryptonote_basic/connection_context.h"
 #include "net/levin_base.h"
 #include "p2p/net_node_common.h"
+#include "request_manager.h"
 #include <boost/circular_buffer.hpp>
 #include <atomic>
+#include <boost/uuid/nil_generator.hpp>
 
 PUSH_WARNINGS
 DISABLE_VS_WARNINGS(4355)
@@ -95,8 +97,10 @@ namespace cryptonote
       HANDLE_NOTIFY_T2(NOTIFY_REQUEST_CHAIN, &cryptonote_protocol_handler::handle_request_chain)
       HANDLE_NOTIFY_T2(NOTIFY_RESPONSE_CHAIN_ENTRY, &cryptonote_protocol_handler::handle_response_chain_entry)
       HANDLE_NOTIFY_T2(NOTIFY_NEW_FLUFFY_BLOCK, &cryptonote_protocol_handler::handle_notify_new_fluffy_block)			
-      HANDLE_NOTIFY_T2(NOTIFY_REQUEST_FLUFFY_MISSING_TX, &cryptonote_protocol_handler::handle_request_fluffy_missing_tx)						
+      HANDLE_NOTIFY_T2(NOTIFY_REQUEST_FLUFFY_MISSING_TX, &cryptonote_protocol_handler::handle_request_fluffy_missing_tx)
       HANDLE_NOTIFY_T2(NOTIFY_GET_TXPOOL_COMPLEMENT, &cryptonote_protocol_handler::handle_notify_get_txpool_complement)
+      HANDLE_NOTIFY_T2(NOTIFY_TX_POOL_HASH, &cryptonote_protocol_handler::handle_notify_tx_pool_hash)
+      HANDLE_NOTIFY_T2(NOTIFY_REQUEST_TX_POOL_TXS, &cryptonote_protocol_handler::handle_request_tx_pool_txs)
     END_INVOKE_MAP2()
 
     bool on_idle();
@@ -156,10 +160,12 @@ namespace cryptonote
     int handle_notify_new_fluffy_block(int command, NOTIFY_NEW_FLUFFY_BLOCK::request& arg, cryptonote_connection_context& context);
     int handle_request_fluffy_missing_tx(int command, NOTIFY_REQUEST_FLUFFY_MISSING_TX::request& arg, cryptonote_connection_context& context);
     int handle_notify_get_txpool_complement(int command, NOTIFY_GET_TXPOOL_COMPLEMENT::request& arg, cryptonote_connection_context& context);
+    int handle_notify_tx_pool_hash(int command, NOTIFY_TX_POOL_HASH::request& arg, cryptonote_connection_context& context);
+    int handle_request_tx_pool_txs(int command, NOTIFY_REQUEST_TX_POOL_TXS::request& arg, cryptonote_connection_context& context);
 		
     //----------------- i_bc_protocol_layout ---------------------------------------
     virtual bool relay_block(NOTIFY_NEW_FLUFFY_BLOCK::request& arg, cryptonote_connection_context& exclude_context);
-    virtual bool relay_transactions(NOTIFY_NEW_TRANSACTIONS::request& arg, const boost::uuids::uuid& source, epee::net_utils::zone zone, relay_method tx_relay);
+    virtual bool relay_transactions(NOTIFY_NEW_TRANSACTIONS::request& arg, std::vector<crypto::hash> &&tx_hashes, const boost::uuids::uuid& source, epee::net_utils::zone zone, relay_method tx_relay);
     //----------------------------------------------------------------------------------
     //bool get_payload_sync_data(HANDSHAKE_DATA::request& hshd, cryptonote_connection_context& context);
     bool should_drop_connection(cryptonote_connection_context& context, uint32_t next_stripe);
@@ -175,6 +181,9 @@ namespace cryptonote
     bool kick_idle_peers();
     bool check_standby_peers();
     bool update_sync_search();
+    void send_txs_request(cryptonote_connection_context &context, std::vector<crypto::hash> &&tx_hashes);
+    std::mutex m_check_tx_request_queue_mutex;
+    bool check_tx_request_queue();
     int try_add_next_blocks(cryptonote_connection_context &context);
     void notify_new_stripe(cryptonote_connection_context &context, uint32_t stripe);
     size_t skip_unneeded_hashes(cryptonote_connection_context& context, bool check_block_queue) const;
@@ -197,6 +206,7 @@ namespace cryptonote
     epee::math_helper::once_a_time_milliseconds<100> m_standby_checker;
     epee::math_helper::once_a_time_seconds<101> m_sync_search_checker;
     epee::math_helper::once_a_time_seconds<43> m_bad_peer_checker;
+    epee::math_helper::once_a_time_seconds<5> m_peer_tx_request_checker;
     std::unordered_map<epee::net_utils::zone, unsigned int> m_max_out_peers;
     mutable epee::critical_section m_max_out_peers_lock;
     tools::PerformanceTimer m_sync_timer, m_add_timer;
@@ -208,6 +218,8 @@ namespace cryptonote
     size_t m_span_time;
     std::atomic<size_t> m_span_limit;
     std::atomic<size_t> m_bss;
+
+    request_manager m_request_manager;
 
     // Values for sync time estimates
     boost::posix_time::ptime m_sync_start_time;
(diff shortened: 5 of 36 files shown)


src/cryptonote_basic/connection_context.cpp
```diff
@@ -65,6 +65,10 @@ namespace cryptonote
       return 1024 * 1024; // 1 MB
     case cryptonote::NOTIFY_GET_TXPOOL_COMPLEMENT::ID:
       return 1024 * 1024 * 4; // 4 MB
+    case cryptonote::NOTIFY_TX_POOL_INV::ID:
+      return 1024 * 1024 * 16; // 16 MB
+    case cryptonote::NOTIFY_REQUEST_TX_POOL_TXS::ID:
+      return 1024 * 1024 * 16; // 16 MB
```
SChernykh, line 71: 16 MB is 524288 32-byte hashes. Do we really need to request 500k transactions at a time? I think a few thousand will be max, so 1 MB will be more than enough.

src/cryptonote_config.h
```diff
@@ -151,14 +152,16 @@
...
+#define P2P_REQUEST_FAILURE_THRESHOLD_PERCENTAGE        70          // if more than 70% of requests fail, the peer is dropped
 
 #define P2P_FAILED_ADDR_FORGET_SECONDS                  (60*60)     //1 hour
 #define P2P_IP_BLOCKTIME                                (60*60*24)  //24 hour
 #define P2P_IP_FAILS_BEFORE_BLOCK                       10
 #define P2P_IDLE_CONNECTION_KILL_INTERVAL               (5*60) //5 minutes
 
 #define P2P_SUPPORT_FLAG_FLUFFY_BLOCKS                  0x01
-#define P2P_SUPPORT_FLAGS                               P2P_SUPPORT_FLAG_FLUFFY_BLOCKS
+#define P2P_SUPPORT_FLAG_V2                             0x02
```
SChernykh, line 163: If v2 adds only notify_tx messages, I would call this flag `P2P_SUPPORT_FLAG_NOTIFY_TX`

src/cryptonote_basic/connection_context.cpp
nahuhh: Just a note: the default txpool config is something like 648mb, or about max 432k txs

SChernykh: Maybe this limit makes sense then. But it doesn't have to be this big anyway - peers can send multiple messages if there are too many transactions.

Boog900: These messages should contain no where near the number of txs as there are in the pool. I would say 1 MB would be good, if we really wanted to push it we could go with the amount of txs needed to reach the 100 MB packet limit in `NOTIFY_NEW_TRANSACTIONS`: 100_000_000 / 1_500 = 66_666, 66_666 * 32 = 2_133_333. So 2 MB.

`NOTIFY_GET_TXPOOL_COMPLEMENT` which contains the ID of every tx in our pool itself has a 4 MB limit.

src/cryptonote_config.h
Boog900: I would say `P2P_SUPPORT_FLAG_TX_RELAY_V2`, the "old" message is called `NOTIFY_NEW_TRANSACTIONS`.

src/cryptonote_basic/connection_context.cpp
0xFFFC0000: Thanks to everyone participating in this discussion. I used 16MB as placeholders, since the tx-relay-v2 parameters will, of course, be a discussion topic when reviewing. I agree that @Boog900’s suggestion is a reasonable approach. Other limitations will kick in even before approaching 16MB.

src/cryptonote_config.h
0xFFFC0000: `P2P_SUPPORT_FLAG_TX_RELAY_V2` seems more reasonable. I have nothing against `P2P_SUPPORT_FLAG_NOTIFY_TX`, but `P2P_SUPPORT_FLAG_NOTIFY_TX` name does not show that we are using totally separate method of tx relay.

I will update it in next push.

src/cryptonote_basic/connection_context.cpp
nahuhh: Is the padding still there?

@boog900 thoughts on padding the hashes? 
My thought was that its unnecessary. Padding txs makes sense (privacy. Different txs have different sizes) but not the hashes, which should be(?) uniform in size

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -202,12 +203,58 @@ namespace levin
...
+    epee::levin::message_writer make_tx_inv_message(std::vector<crypto::hash>&& txs, const bool pad)
+    {
+      NOTIFY_TX_POOL_INV::request request{};
+      request.txs = std::move(txs);
+
+      if (pad)
+      {
+        size_t bytes = 9 /* header */ + 4 /* 1 + 'txs' */ + tools::get_varint_data(request.txs.size()).size();
+        for(auto tx_hash = request.txs.begin(); tx_hash!=request.txs.end(); ++tx_hash)
+          bytes += tools::get_varint_data(sizeof(*tx_hash)).size() + sizeof(*tx_hash);
```
Boog900, line 215: The serialized message stores all hashes in a single field, so this is incorrect.

```suggestion
        size_t bytes = 9 /* header */ + 4 /* 1 + 'txs' */ + tools::get_varint_data(request.txs.size() * 32).size() + request.txs.size() * 32;
```

Although even that is incorrect, it uses the consensus (COCOA BEANS) varint, not the epee varint and misses the field tag size. This is the same as the `make_tx_message` padding calculation. 

@vtnerd does this slight difference in calculation of the size vs the real size matter for the padding?

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -934,25 +1063,41 @@ namespace cryptonote
...
+        if (!tvc.m_no_drop_offense)
+        {
+          LOG_PRINT_CCONTEXT_L1("Tx verification failed, dropping connection");
+          drop_connection(context, false, false);
+          return 1;
+        }
+      }
+      else
+      {
+        m_peer_info_manager.add_received(context.m_connection_id);
```
Boog900, line 1080: IMO this should take into account if we actually requested the tx from the peer. Currently the peer can send us an inv, we request it, they send a different tx and to us it looks like they have sent as many as we requested, when in reality they haven't.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -883,6 +938,80 @@ namespace cryptonote
...
+      post_notify<NOTIFY_REQUEST_TX_POOL_TXS>(req, context);
+      MLOG_P2P_MESSAGE("Requested " << req.txs.size() << " missing transactions via RequestTxPoolTxs");
+    }
+    else
+    {
+      MLOG_P2P_MESSAGE("All advertised transactions are already in the pool or were requested recently");
+    }
+
+    return 1;
+  }
```
Boog900, line 981: Currently from what I can see a peer has 30 seconds to bomb our node with as many tx invs as possible and we have to track each one. I think we should a limit on the amount of inflight requests we make to a single peer.

So if we have more than like 50 requests to a single peer we should ignore all future invs from that peer until that number drops.

Boog900 (commented):
First pass, thanks for working on this :pray:

src/cryptonote_protocol/txrequesthandler.h
```diff
@@ -0,0 +1,129 @@
...
+private:
+  // Disable everything that can be disabled
+  // to prevent misuse of the class
+  tx_request_handler() = delete;
+  tx_request_handler(tx_request_handler &&other) noexcept = delete;
+  tx_request_handler(const tx_request_handler &other) = delete;
+  tx_request_handler &operator=(tx_request_handler &&other) = delete;
+  tx_request_handler &operator=(const tx_request_handler &other) = delete;
+  tx_request_handler *operator&() = delete;
+  bool operator>(const tx_request_handler &other) const = delete;
```
vtnerd, line 61: C++ doesn't define defaults for these, just remove them and the compiler will generate errors.

src/cryptonote_protocol/txrequesthandler.h
```diff
@@ -0,0 +1,129 @@
...
+
+  void update_request_deadline(std::time_t request_deadline) {
+    m_request_deadline = request_deadline;
+  }
+
+  std::time_t get_request_deadline() const { return m_request_deadline; }
+
+private:
+  // Thread for checking the transaction requests
+  std::thread m_handler_thread;
```
vtnerd, line 122: There has to be a way to implement this without some arbitrary thread tucked away. Admittedly I'm being lazy and not going through everything, but this just has to be the result of poor design choices.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -883,6 +938,80 @@ namespace cryptonote
...
+        response.txs.push_back(std::move(tx_blob));
+      }
+      // If tx is not in the pool, then ignore it (do not penalize peer)
+    }
+
+    // Send response if any txs found
+    if (!response.txs.empty())
+    {
+      post_notify<NOTIFY_NEW_TRANSACTIONS>(response, context);
+    }
```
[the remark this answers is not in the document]
vtnerd: * Padding mostly helps with Tor/I2P when noise is OFF. In this PR, only number of hashes is potentially leaked, so I don't see a massive need for including it.
* Padding this way should be avoided in favor of what is done in my [SSL patch](https://github.com/monero-project/monero/pull/8996/files). See `levin_base.h` and `levin_base.cpp`. The inner epee format is self-describing so you can safely write random bytes after the payload, and they will get dropped/ignored. This is quicker, more efficient, and less error prone than doing it with the `_` field.
  * The SSL will deprecate this padding code if merged first, so might as well just import it.

src/cryptonote_protocol/peerinfo_manager.h
```diff
@@ -0,0 +1,413 @@
...
+#include <string>
+#include <unordered_set>
+
+#include "crypto/hash.h"
+#include "cryptonote_config.h"
+#include "misc_log_ex.h"
+#include "string_tools.h"
+#include "syncobj.h"
+
+class peer_info_manager {
```
vtnerd, line 52: This `class` shouldn't exist - you can fetch by `uuid`/`connection_id` already. Instead, the information should be tucked into the `connection_context` object.

This implementation creates several problems:
  * You have to separately prune the map when connections close
  * Everything EXCEPT "tx inventory" is stored in the `connection_context` object

src/cryptonote_protocol/peerinfo_manager.h
```diff
@@ -0,0 +1,413 @@
...
+
+public:
+  class peer_info {
+
+  private:
+    struct peer_info_state {
+      boost::uuids::uuid connection_id;
+      std::unordered_set<crypto::hash> announcements;
+      size_t received = 0;
```
vtnerd, line 61: Do we really need to track all of this information? Is it useful for debugging somehow?

src/cryptonote_protocol/peerinfo_manager.h
```diff
@@ -0,0 +1,413 @@
...
+      return m_data->announcements.size() + m_data->received + m_data->requested_from_me +
+             m_data->requested_from_peer + m_data->sent;
+    }
+
+    size_t get_missed() const {
+      epee::read_lock r_lock(m_data->m_mutex);
+      return m_data->missed;
+    }
+
+    std::string get_info() const {
```
vtnerd, line 189: Just make this `std::ostream operator<<(...)`. Its more efficient.

src/cryptonote_protocol/peerinfo_manager.h
```diff
@@ -0,0 +1,413 @@
...
+private:
+  std::unordered_set<peer_info, peer_info::hash> m_peer_info;
+
+public:
+  peer_info_manager() = default;
+
+  void add_peer(const boost::uuids::uuid &id) { m_peer_info.emplace(id); }
+
+  void remove_peer(const boost::uuids::uuid &id) {
+    auto it = std::find_if(m_peer_info.begin(), m_peer_info.end(),
```
vtnerd, line 212: You should be using an `unordered_map<id, data>` instead of an `unordered_set` and then all of these `find_if` calls become efficient hash lookups instead of log(n) iteration. Another reason to use the built-in connection tracking, which does this (relatively) efficiently already.

vtnerd (changes requested):

src/cryptonote_protocol/peerinfo_manager.h
vtnerd: I got a bit lazy and didn't see if it was being used outside of `std::ostream` cases. So safely ignore if not always in use.

src/cryptonote_protocol/txrequesthandler.h
0xFFFC0000: Yes, I know C++ doesn’t define them, it is left over that I didn’t want to touch. Since don’t want anyone to abuse this class in future.

src/cryptonote_protocol/txrequesthandler.h
0xFFFC0000: > This just has to be the result of poor design choices.

That is inaccurate statement. I don’t want to block existing operations. I know already idle gets called once a second. But doing stuff there means blocking it.

src/cryptonote_protocol/peerinfo_manager.h
0xFFFC0000: Yes. That is the problem with our existing code. Used haphazard, temporary methods. And trying to fix later. 

Instead I am keeping all the info in central, well defined place. That can be used later. Possibly with a better way to statistically score peers and remove them if the score drops.

src/cryptonote_protocol/peerinfo_manager.h
0xFFFC0000: That is inaccurate statement IMHO. unordered_set and unordered_map are using same underlying dtat structure. For example, IIRC libc++ is using redblack tree to implement both. I think you are mistaken `std::unordered_set`  with `std::set`


Both std::unordered_set and std::unordered_map offer average-case O(1) time for insert, find, and erase (hash-based) operations

src/cryptonote_protocol/peerinfo_manager.h
0xFFFC0000: I disagree. Connection context is mostly networking / connection related information. I am trying to have peer abstraction here without the networking information. 

Eventually I want to have well defined statistical scoring system that would score the peer based on its performance. If this was more networking related information, your proposal would’ve made more sense. But at this abstraction layer I disagree with your proposal.

src/cryptonote_protocol/peerinfo_manager.h
vtnerd: > Yes. That is the problem with our existing code. Used haphazard, temporary methods. And trying to fix later.

This is somewhat vague criticism. Do other systems of similar size have permanent (never needing maintenance) code? Zero bugs or something?

> No industry grade professional engineer is thinking about keeping few bytes, that can be used later, as problem.

I never said it would be a memory issue; we sometimes track information of limited use in debugging. These field dumps can just clutter the log further in the worse case. I'm on the fence whether these fields help with debugging. If they helped you at some point, then perhaps they are useful.

I find that that tracing when particular events occur is more useful. Because you don't really know ordering or timing with these kinds of dumps, just that event X occurred Y times.

src/cryptonote_protocol/txrequesthandler.h
vtnerd: At a glance it looks like this `std::thread` "blocks" the ASIO threads with locks. Is that any better? The `std::thread` could be blocked for a while by the ASIO thread (or vice-versa)?

Just use the idle thread timer, it's there specifically for this purpose. As long as the locks are released within a reasonable time-frame it'll be fine. You might be able to pull a strand implementation to prevent all blocking, but that'd be a bit more work I think.

src/cryptonote_protocol/peerinfo_manager.h
vtnerd: And to clarify a bit, the design makes more sense if the `shared_ptr` was being used outside of this class. But the way its being used here doesn't seem helpful.

src/cryptonote_protocol/levin_notify.cpp
vtnerd: This is why I promoted the usage of the technique the SSL PR used - it's more accurate, quicker, and easier to just pad the end of the levin message as the extra trailing data gets dropped. Arguably this feature should be dropped entirely (see below).

Assuming this implementation is kept, it's possibly leaking data, but it's unlikely to matter. This feature is only useful if the connection is encrypted, which currently occurs only over Tor/I2P, both of which are already padded to disrupt traffic shaping analysis. In other words, Tor/I2P already do what we want here.

When/if SSL gets merged, this should be re-thought as SSL provides no automatic padding, which is why I added randomized padding (at the levin "layer") in that patch. It pads a randomized amount up to 8KiB, but perhaps it should just "round up" like Tor/I2P do (if it works for those networks, surely it works for us).

src/cryptonote_protocol/cryptonote_protocol_handler.inl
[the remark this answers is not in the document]
Boog900: I agree padding is not necessary for the new messages. Although here is where we are sending the full txs in response to a request for them.
 
 The old protocol looked like this:
 
 node --- txs (maybe padded) --> node
 
 the new protocol:
 
 node --- inv  (not  padded) --> node
 node <-- request  (not padded) --- node
 node --- txs  (not padded) --> node
 
  my question was about the last step. IMO we should pad to match the old protocol. Although if we never fluff over Tor/I2p this isn't really necessary.

src/cryptonote_protocol/peerinfo_manager.h
0xFFFC0000: Your criticism about using `connection_context` is reasonable.

src/cryptonote_protocol/txrequesthandler.h
0xFFFC0000: > You might be able to create a strand implementation to prevent all blocking, but that'd be a bit more work I think. 

This is a better approach IMHO. Let me look up what I can do. 

All of your comments will be addressed in the next update.

src/cryptonote_protocol/peerinfo_manager.h
0xFFFC0000: All of your comments will be addressed in the next update.

src/cryptonote_protocol/peerinfo_manager.h
0xFFFC0000: All of your comments will be addressed in the next update.

src/cryptonote_config.h
0xFFFC0000: Fixed. Feel free to resolve it.

src/cryptonote_protocol/peerinfo_manager.h
vtnerd: A `shared_ptr` was not useful originally, but may be useful after the ongoing re-design.

0xFFFC0000: @vtnerd  thanks, will take that into account.

src/cryptonote_protocol/cryptonote_protocol_defs.h
```diff
@@ -375,5 +375,51 @@ namespace cryptonote
...
+  {
+    const static int ID = BC_COMMANDS_POOL_BASE + 11;
+
+    struct request_t
+    {
+      std::vector<crypto::hash> txs;
+      std::string _; // padding
+
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE_CONTAINER_POD_AS_BLOB(txs)
```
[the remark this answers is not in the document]
0xFFFC0000: Fixed. Feel free to resolve it. 

https://github.com/monero-project/monero/compare/f2061bf8976e3872e7d227dd5ee365213a317a4c..d024d369d590a0bd6488757266fee0c45df217c7

src/cryptonote_basic/connection_context.cpp
0xFFFC0000: Fixed. Feel free to resolve it.

https://github.com/monero-project/monero/compare/d024d369d590a0bd6488757266fee0c45df217c7..3fcd5c70cf964b4f283c92538c81f2f2aed84c53

src/cryptonote_protocol/cryptonote_protocol_defs.h
```diff
@@ -375,5 +375,51 @@ namespace cryptonote
...
+  struct NOTIFY_REQUEST_TX_POOL_TXS
+  {
+    const static int ID = BC_COMMANDS_POOL_BASE + 12;
+
+    struct request_t
+    {
+      std::vector<crypto::hash> txs;
+
+      BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE_CONTAINER_POD_AS_BLOB(txs)
```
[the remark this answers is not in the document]
0xFFFC0000: Fixed. Feel free to resolve it.

https://github.com/monero-project/monero/compare/f2061bf8976e3872e7d227dd5ee365213a317a4c..d024d369d590a0bd6488757266fee0c45df217c7

src/cryptonote_protocol/cryptonote_protocol_handler.inl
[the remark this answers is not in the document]
0xFFFC0000: Added padding here. Fixed. Feel free to resolve it.

https://github.com/monero-project/monero/compare/3fcd5c70cf964b4f283c92538c81f2f2aed84c53..f345aea15430f7bbc79e6b613baba9a6bd245bf8

src/cryptonote_protocol/cryptonote_protocol_defs.h
```diff
@@ -375,5 +375,51 @@ namespace cryptonote
...
+  * and optionally request the actual serialized transactions for them.   *
+  *************************************************************************/
+  struct NOTIFY_TX_POOL_INV
+  {
+    const static int ID = BC_COMMANDS_POOL_BASE + 11;
+
+    struct request_t
+    {
+      std::vector<crypto::hash> txs;
+      std::string _; // padding
```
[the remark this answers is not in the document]
0xFFFC0000: Removed the padding here. Fixed. Feel free to resolve it.

https://github.com/monero-project/monero/compare/3fcd5c70cf964b4f283c92538c81f2f2aed84c53..f345aea15430f7bbc79e6b613baba9a6bd245bf8

src/cryptonote_protocol/levin_notify.cpp
0xFFFC0000: Feel free to resolve this, we can ignore this comment since we have removed the padding: 

https://github.com/monero-project/monero/pull/9933#discussion_r2233028720

src/cryptonote_protocol/cryptonote_protocol_handler.inl
0xFFFC0000: Thanks. Feel free to resolve this.

src/cryptonote_protocol/txrequesthandler.h
0xFFFC0000: Feel free to resolve this; fixed in updated push.

src/cryptonote_protocol/txrequesthandler.h
0xFFFC0000: Feel free to resolve this; fixed in updated push, and addressed all the criteria you mentioned. 

I implemented this feature using `on_idle`. I did a few experiments and found `on_idle` robust. 

About the concurrency mechanism, I kept the locking, since it is read/write (lightweight) locking. The issue with the `strand` is that I had to create another `io_context` ( and a thread to run it ). The `cryptonote_protocol_handler` does not have any `io_context` by itself.

src/cryptonote_protocol/peerinfo_manager.h
0xFFFC0000: Feel free to resolve it.  I moved everything inside `connection_context`, and got rid of this class.

src/cryptonote_protocol/peerinfo_manager.h
0xFFFC0000: > You should be using an `unordered_map<id, data>` instead of an `unordered_set` and then all of these `find_if` calls become efficient hash lookups instead of log(n) iteration. Another reason to use the built-in connection tracking, which does this (relatively) efficiently already.

Addressed in updated push. Everything moved to `connection_context`, so no need to discuss this further. Thanks

src/cryptonote_protocol/cryptonote_protocol_handler.inl
0xFFFC0000: Addressed in the new push. Feel free to resolve this issue if you consider the implementation adequate.

0xFFFC0000:
This is anecdotal data—consider it a single data point rather than a controlled, scientific experiment.

# Test Setup
- 20 `monerod` nodes
- 20 `monero-wallet-rpc` instances
- Probabilistic mining to simulate real-world block production without the cost of `RandomX` computation
- After synchronizing a few blocks, each node accumulates unlocked balance and then sends 10 % of that balance at random to another wallet on the same private testnet

# Baseline (No Transactions)

With no transactions in flight, each new block transfers roughly 5 MBit/s. This establishes our baseline bandwidth usage:

image

# Master Branch (With Transactions)
When transactions are enabled and the tx pool is active, the master configuration averages about 20 MBit/s:

image

# PR #9933 (With Transactions)
Applying PR #9933 consistently cuts bandwidth by at least 50 %, bringing the average down to around 12 MBit/s under similar transaction load:

image

Although this isn’t a formal experiment, **the results are stable across multiple runs and varied configurations.**

## Benchmarking Tool
All measurements were collected using a simple benchmarking tool I wrote for `monerod` [1].

1. https://github.com/0xFFFC0000/benchmark-project

nahuhh:
Not sure if this is preexisting on master/release:

testing was done on a testnet with 10mb blocks.

during tx relay, sending all txs from Node A keeps a very low "received" amount, as shown via `print_net_stats`.
Node C, which only received txs _and_ mines blocks, is the opposite.

The strange part, to me, is: when Node C produces a block, it sends quite a large block. IIUC, fluffyblocks should mean that Node C should be sending a minimal amount of data, since Node A already has _100%_ of the txs associated with the block.

here's an example of print_net_stats from Node A before and after receiving a block.

<details>
<summary>print_net_stats</summary>

Before
```
print_net_stats
Received 90371 bytes (88.25 kB) in 48 packets in 7.0 minutes, average 214 B/s = 0.00% of the limit of 32.00 MB/s
Sent 7467544 bytes (7.12 MB) in 952 packets in 7.0 minutes, average 17.32 kB/s = 0.21% of the limit of 8.00 MB/s
```

After
```
2025-09-13 23:43:20.634 I Synced 1478/1478
print_net_stats
Received 6581613 bytes (6.28 MB) in 848 packets in 7.3 minutes, average 14.78 kB/s = 0.05% of the limit of 32.00 MB/s
Sent 7675842 bytes (7.32 MB) in 979 packets in 7.3 minutes, average 17.23 kB/s = 0.21% of the limit of 8.00 MB/s
```
</details>

src/cryptonote_basic/connection_context.h
```diff
@@ -118,6 +120,42 @@ namespace cryptonote
     size_t m_num_requested;
     copyable_atomic m_new_stripe_notification{0};
     copyable_atomic m_idle_peer_notification{0};
+
+    private:
+      struct connection_statistics
+      {
+        mutable epee::rw_mutex mutex;
+        std::unordered_set<crypto::hash> tx_announcements;
+        size_t received = 0;
```
vtnerd: Could these be `atomic<size_t>` instead? The above functions seems to indicate so. The benefit is that the reads will be faster as no write to memory has to be made (whereas a lock implies writing to memory to acquire lock).

src/cryptonote_protocol/cryptonote_protocol_defs.h
```diff
@@ -207,6 +209,39 @@ namespace cryptonote
     };
     typedef epee::misc_utils::struct_init<request_t> request;
   };
+
+  struct tx_request_padding
+  {
+    void operator()(NOTIFY_NEW_TRANSACTIONS::request& request) const
+    {
+      size_t bytes = 9 /* header */ + 4 /* 1 + 'txs' */ + tools::get_varint_data(request.txs.size()).size();
```
vtnerd, line 222: I just looked, and unfortunately I never fully commented on this type of padding. 

My [SSL patch](https://github.com/monero-project/monero/pull/8996) has a much cleaner solution to the problem. See `levin_base.h` and `levin_base.cpp`. The function writes zeroes after the payload which gets ignored be the epee binary parser. It's much faster and more accurate than what's attempted here.

If SSL get merged first, this basically has to be removed as dead code, and if SSL gets merged first this will be removed, etc.

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -36,6 +36,7 @@
 
 #include <boost/program_options/variables_map.hpp>
 #include <string>
+#include <thread>
```
vtnerd: Is this leftover from earlier version?

src/cryptonote_protocol/request_manager.h
```diff
@@ -0,0 +1,248 @@
...
+  request_manager(const request_manager &) = delete;
+  request_manager &operator=(const request_manager &) = delete;
+  request_manager(request_manager &&) noexcept = delete;
+  request_manager &operator=(request_manager &&) = delete;
+
+public:
+  request_manager() : m_requested_txs(), m_mutex() {}
+
+  bool remove_transaction(const crypto::hash &tx_hash) {
+    MINFO("Removing transaction: " << epee::string_tools::pod_to_hex(tx_hash));
```
vtnerd: These functions should be in a separate cpp. No good reason to attempt inlining.

contrib/epee/include/syncobj.h
```diff
@@ -84,6 +86,10 @@ namespace epee
 
 #define  CRITICAL_REGION_END() }
 
+using rw_mutex = std::shared_timed_mutex;
```
vtnerd: I see little benefit to these aliases, especially considering that the codebase has been moving to explictly calling them (by their full `std::` name).

src/cryptonote_protocol/request_manager.h
```diff
@@ -0,0 +1,248 @@
...
+      return earliest_request->peer_id;
+    }
+
+    MINFO("No available requests for tx " << epee::string_tools::pod_to_hex(tx_hash));
+    return boost::uuids::nil_uuid();
+  }
+
+  void for_each_request(std::function<void(request_manager&, const request &, const std::time_t)> &f,
+                                            const std::time_t request_deadline) {
+    MINFO("Iterating over requested transactions");
```
vtnerd: Should there be a lock here? Presumably it was left out because a recursive call could be made, however, `m_requested_txs` can be modified by another thread ... ?

I think you need to make a copy of every request into a vector, and then feed it to the callback. Or even easier, just return the vector.

src/cryptonote_protocol/txrequestqueue.h
```diff
@@ -0,0 +1,87 @@
...
+#include "string_tools.h"
+#include "syncobj.h"
+#include "common/util.h"
+
+
+struct request {
+  boost::uuids::uuid peer_id;
+  crypto::hash tx_hash;
+  std::time_t firstseen_timestamp;
+  mutable bool in_flight = false;
```
vtnerd: Why mark this as mutable? Why can't ``fly()` be not-const?

src/cryptonote_protocol/txrequestqueue.h
```diff
@@ -0,0 +1,87 @@
...
+
+  request(const boost::uuids::uuid& peer_id, const crypto::hash& tx_hash, std::time_t firstseen_timestamp)
+    : peer_id(peer_id), tx_hash(tx_hash), firstseen_timestamp(firstseen_timestamp) {}
+
+  void fly() const { in_flight = true; }
+  bool is_in_flight() const { return in_flight; }
+};
+
+using namespace boost::multi_index;
```
vtnerd: Avoid this `using namespace` in a header, we don't want to drag in everything from that namesake into the global namespace.

src/cryptonote_protocol/txrequestqueue.h
```diff
@@ -0,0 +1,87 @@
...
+        hashed_non_unique<member<request, crypto::hash, &request::tx_hash>>,
+        // Index 2: by (peer_id, tx_hash) - unique requests
+        hashed_unique<composite_key<request,
+            member<request, boost::uuids::uuid, &request::peer_id>,
+            member<request, crypto::hash, &request::tx_hash>
+        >>
+    >
+> request_container;
+
+#define get_requests_by_peer_id(request_container) request_container.get<0>()
```
vtnerd: Templated function does the same thing, and is somewhat preferable here.

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -49,7 +50,10 @@
 #include "cryptonote_basic/connection_context.h"
 #include "net/levin_base.h"
 #include "p2p/net_node_common.h"
+#include "request_manager.h"
 #include <boost/circular_buffer.hpp>
+#include <boost/uuid/uuid.hpp>
```
vtnerd: These two includes look unnecessary here too (but perhaps are needed within `request_manager.h`).

src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -184,6 +193,9 @@ namespace cryptonote
     epee::math_helper::once_a_time_milliseconds<100> m_standby_checker;
     epee::math_helper::once_a_time_seconds<101> m_sync_search_checker;
     epee::math_helper::once_a_time_seconds<43> m_bad_peer_checker;
+    std::atomic<uint64_t> m_interval_peer_request_checker{P2P_DEFAULT_REQUEST_TIMEOUT}; // microseconds
+    std::function<uint64_t()> m_peer_request_interval; // set in ctor
```
vtnerd: This should be `const`.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -209,7 +211,15 @@ namespace cryptonote
...
     m_sync_pruned_blocks = command_line::get_arg(vm, cryptonote::arg_sync_pruned_blocks);
+    size_t request_deadline = command_line::get_arg(vm, cryptonote::arg_request_deadline);
 
+    if (request_deadline < 1 || request_deadline > 180)
+    {
+      MERROR("Request interval must be between 1 and 180 seconds");
+      MERROR("Setting to default of 30 seconds");
+      request_deadline = P2P_DEFAULT_REQUEST_TIMEOUT;
+    }
+    m_interval_peer_request_checker.store(request_deadline * 1000000); // microseconds
```
vtnerd: The store only happens once during init; `epee::once_a_time` should store the timer value instead of a callback function.

src/cryptonote_protocol/levin_notify.cpp
```diff
@@ -416,14 +431,15 @@ namespace levin
...
       boost::uuids::uuid source_;
 
       void operator()()
       {
-        run(std::move(zone_), epee::to_span(txs_), source_);
+        run(std::move(zone_), epee::to_span(txs_), epee::to_span(tx_hashes_), source_);
       }
 
-      static void run(std::shared_ptr<detail::zone> zone, epee::span<const blobdata> txs, const boost::uuids::uuid& source)
+      static void run(std::shared_ptr<detail::zone> zone, epee::span<const blobdata> txs, epee::span<const crypto::hash> tx_hashes, const boost::uuids::uuid& source)
```
vtnerd, line 446: I would delay creating `tx_hashes` until inside of this function. This removes some memory needed for the `fluffy_notify` and `dandelionpp_notify` callbacks inside of ASIO land. This also removes the duplication in `send_txs` where the hash is computing in two different switch statements.

src/cryptonote_protocol/txrequestqueue.h
```diff
@@ -0,0 +1,87 @@
...
+#include "misc_log_ex.h"
+#include "string_tools.h"
+#include "syncobj.h"
+#include "common/util.h"
+
+
+struct request {
+  boost::uuids::uuid peer_id;
+  crypto::hash tx_hash;
+  std::time_t firstseen_timestamp;
```
vtnerd: Switching to `std::chrono::steady_clock::time_point` has several advantages when computing time differences (better type checking, etc).

src/cryptonote_protocol/txrequestqueue.h
```diff
@@ -0,0 +1,87 @@
...
+  request(const boost::uuids::uuid& peer_id, const crypto::hash& tx_hash, std::time_t firstseen_timestamp)
+    : peer_id(peer_id), tx_hash(tx_hash), firstseen_timestamp(firstseen_timestamp) {}
+
+  void fly() const { in_flight = true; }
+  bool is_in_flight() const { return in_flight; }
+};
+
+using namespace boost::multi_index;
+
+typedef multi_index_container<
```
vtnerd, line 95: Why not declare this inside of `request_manager` - its never referenced outside of that class.

src/cryptonote_protocol/request_manager.h
```diff
@@ -0,0 +1,248 @@
...
+    if (peer_range.first != peer_range.second) {
+      size_t removed = std::distance(peer_range.first, peer_range.second);
+      by_peer.erase(peer_range.first, peer_range.second);
+      MINFO("Removed " << removed << " requests for peer " << epee::string_tools::pod_to_hex(peer_id));
+    }
+  }
+
+  void cleanup_stale_requests(std::time_t max_age_seconds) {
+    std::time_t now = std::time(nullptr);
+    std::time_t cutoff_time = now - max_age_seconds;
```
vtnerd: `std::time_t` has no defined type or interval (which is why I proposed `std::chrono` elsewhere).

src/cryptonote_protocol/txrequestqueue.h
```diff
@@ -0,0 +1,87 @@
...
+#include <boost/multi_index/composite_key.hpp>
+#include <boost/multi_index/member.hpp>
+
+#include "misc_log_ex.h"
+#include "string_tools.h"
+#include "syncobj.h"
+#include "common/util.h"
+
+
+struct request {
```
vtnerd: You need a better name than this in the global namespace.

src/cryptonote_protocol/request_manager.h
```diff
@@ -0,0 +1,248 @@
...
+    auto it = by_peer_and_tx.find(boost::make_tuple(id, tx_hash));
+    if (it != by_peer_and_tx.end()) {
+      // already have this peer for this tx
+      MDEBUG("Peer " << epee::string_tools::pod_to_hex(id)
+                     << " already in request queue for tx "
+                     << epee::string_tools::pod_to_hex(tx_hash));
+      return false;
+
+    }
+    m_requested_txs.insert(request(id, tx_hash, first_seen));
```
vtnerd: Creating a `request` is roughly as cheap as a `boost::make_tuple` call. Just try the insert, and then return the bool if it succeeded. This reduces the lookups by 1.

src/cryptonote_protocol/request_manager.h
```diff
@@ -0,0 +1,248 @@
...
+                               << " from peer "
+                               << epee::string_tools::pod_to_hex(it->peer_id));
+        it = by_tx_hash.erase(it);
+      } else {
+        ++it;
+      }
+    }
+
+    // Recompute range after potential erasures.
+    by_tx_hash = get_requests_by_tx_hash(m_requested_txs);
```
vtnerd: This line should be removed, it's not needed.

src/cryptonote_protocol/request_manager.h
```diff
@@ -0,0 +1,248 @@
...
+    if (by_tx_hash.empty()) {
```
vtnerd: This should be checking whether the entire map is empty, is that intended? Don't see how this is useful.

src/cryptonote_protocol/request_manager.h
```diff
@@ -0,0 +1,248 @@
...
+#include "syncobj.h"
+#include "txrequestqueue.h"
+
+#include <boost/functional/hash.hpp>
+#include <boost/uuid/nil_generator.hpp>
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_io.hpp>
+#include <unordered_map>
+
+class request_manager {
```
vtnerd, line 49: This also should have a more useful name; too generic.

src/cryptonote_basic/connection_context.h
```diff
@@ -118,6 +120,42 @@ namespace cryptonote
...
+        size_t requested_from_me = 0;
+        size_t requested_from_peer = 0;
+        size_t sent = 0;
+        size_t missed = 0;
+        size_t in_flight_requests = 0;
+      };
+      std::shared_ptr<connection_statistics> m_connection_stats;
```
vtnerd: Why is this a `shared_ptr`? I don't see where this shared with any other object, and it doesn't need to be a `unique_ptr` either.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -883,6 +910,86 @@ namespace cryptonote
...
+
+    std::vector<blobdata> txs;
+
+    // Iterate over requested txin hashes
+    for (const auto &tx_hash : arg.t)
+    {
+      context.add_requested_from_me();
+      // Attempt to get the transaction blob from the mempool;
+      cryptonote::blobdata tx_blob;
+      if (m_core.get_pool_transaction(tx_hash, tx_blob, cryptonote::relay_category::all))
```
vtnerd: Do not perform a `::all` lookup here. An attacker could issue this command at any time, and get transactions still in the D++ stem phase. This is particularly important if an attacker managed to blackhole a transaction.

There also needs to be a lookup in the blockchain for the tx, as it could've moved. If the tx is silently not returned, the caller could drop the connection unnecessarily.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -934,25 +1041,42 @@ namespace cryptonote
...
+      {
+        context.add_received();
+        // This is extremely inefficient. We have to change m_core.handle_incoming_tx
+        // we are already parsing inside handle_incoming_tx
+        transaction tx;
+        crypto::hash tx_hash;
+        if (cryptonote::parse_and_validate_tx_from_blob(tx_blob, tx, tx_hash)
+            && m_request_manager.already_requested_tx(tx_hash))
+        {
+          m_request_manager.remove_requested_transaction(context.m_connection_id, tx_hash);
```
vtnerd: This should be a unilateral remove by `tx_hash` (i.e. drop `m_connection_id` from call)?

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1565,6 +1689,37 @@ namespace cryptonote
               m_block_queue.remove_spans(span_connection_id, start_height);
               return 1;
             }
+            else if (bvc.m_added_to_main_chain)
+            {
+              const std::vector<crypto::hash>* tx_hashes_ptr = nullptr;
```
vtnerd: This should be declared after `tmp` since its pointing into it.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1766,6 +1922,78 @@ skip:
...
+          }
+          else
+          {
+            MCINFO("net.p2p.msg", "Missed tx announcement less than threshold of the time, not dropping peer : " << epee::string_tools::pod_to_hex(current_request_peer_id));
+          }
+          // We are abandoning this request path: decrement previous in-flight
+          if (!current_request_peer_id.is_nil())
+          {
+            m_p2p->for_connection(current_request_peer_id, [&](cryptonote_connection_context &ctx, nodetool::peerid_type, uint32_t)->bool {
+              ctx.remove_in_flight_request();
```
vtnerd: This should be in the `for_connection` call above; this entire if statement is duplicated from the above check.

vtnerd (changes requested):

0xFFFC0000:
Apologies for the delay, [this](https://github.com/0xFFFC0000/monero/tree/dev/0xfffc/tx-relay-v2-address-second-round) branch contains ( WIP ) solution to issues raised by @vtnerd 

It will be finalized in the next few days.

src/cryptonote_protocol/cryptonote_protocol_defs.h
0xFFFC0000: I see. Thanks. I will leave this for now. As you asked, if SSL-Patch gets in first, I will remove this. I am fine either way.

src/cryptonote_protocol/txrequestqueue.h
0xFFFC0000: Because I am storing them inside boost::multi_index_container. If I want to remove the mutable, I have to use:

 ```
 auto& idx = get_requests_by_peer_and_tx(container);
 auto it = idx.find(std::make_tuple(peer_id, tx_hash));
 if (it != idx.end()) {
     idx.modify(it, [](request& r) { r.in_flight = true; });
 }
 ```

 Which I would say is quite cumbersome. `mutable` is quite safe here. Since `in_flight` value does not have any impact on the key.

src/cryptonote_protocol/cryptonote_protocol_handler.h
0xFFFC0000: `cryptonote_protocol_handler.h` needs ‘request_manager’, but you were right about `boost::uuid`.

src/cryptonote_protocol/cryptonote_protocol_handler.h
0xFFFC0000: `m_interval_peer_request_checker` should not be const since we are changing it. I have a command line flag so the user can use a custom value of `m_interval_peer_request_checker`

src/cryptonote_protocol/cryptonote_protocol_handler.inl
0xFFFC0000: We are assigning it with command-line option. Not compile time. So, it's not possible to make this change. `once_a_time` needs compile-time argument.

src/cryptonote_protocol/levin_notify.cpp
0xFFFC0000: I have to change the `run` API and send an instance of the object. Since the `run` method is static, and no way to access `fluff_notify` object fields, unless I pass a pointer to that object. If that's what you prefer, I can do that.

src/cryptonote_protocol/txrequestqueue.h
0xFFFC0000: Just to keep separate concerns and make it more readable, in a sense. I can move it to `request_manager` if you prefer.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -883,6 +910,86 @@ namespace cryptonote
...
+    // Iterate over each advertised transaction hash and check our pool and our requested tracker.
+    for (const auto &tx_hash : arg.t)
+    {
+      if (!context.can_process_additional_request())
+      {
+        LOG_ERROR_CCONTEXT("Too many in-flight requests, cannot accept");
+        break;
+      }
+      // If we have the tx in our pool, also remove it from request manager and skip.
+      if (m_core.pool_has_tx(tx_hash))
```
j-berman: Might be in the chain here too

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -883,6 +910,86 @@ namespace cryptonote
...
+    }
+
+    // Send response if any txs found
+    if (!txs.empty())
+    {
+      NOTIFY_NEW_TRANSACTIONS::request request = {};
+      request.txs = std::move(txs);
+      pad_tx_request(request);
+      post_notify<NOTIFY_NEW_TRANSACTIONS>(request, context);
+    }
```
j-berman: I'm not 100% certain if this will have issues with the connection code, but perhaps you could easily just send batches of the txs here. I.e. iterate over `txs` some slice at a time, calling `post_notify` with that slice. And then just make sure each slice is some max limit below the max packet size. On first pass, it doesn't look like there would be issues with this approach in the connection code, but it could use a deeper look.

UPDATE: here's how I sent batches of txs for the pool tx complement response: https://github.com/seraphis-migration/monero/pull/168

j-berman (commented):
I haven't had the chance to review this at length yet, but it would be nice if it also implemented some max limit of txs when responding to `NOTIFY_REQUEST_TX_POOL_TXS`, as opposed to relaying all missing txs in a single response. This avoids problems when the pool exceeds the max packet size limit.

EDIT: this suggestion is a weakly held suggestion and not a requirement for this PR, since this PR only applies to brand new txs as they are relayed on the network.

~~Ultimately, I also think this PR could replace the txpool complement fetching that nodes currently do (`request_txpool_complement`).~~ (edit: ignore this, we'd still want the txpool complement in addition to this)

0xFFFC0000:
> I haven't had the chance to review this at length yet, but it would be nice if it also implemented some max limit of txs when responding to `NOTIFY_REQUEST_TX_POOL_TXS`, as opposed to relaying all missing txs in a single response. This avoids problems when the pool exceeds the max packet size limit.
> 
> EDIT: this suggestion is a weakly held suggestion and not a requirement for this PR, since this PR only applies to brand new txs as they are relayed on the network.
> 
> ~Ultimately, I also think this PR could replace the txpool complement fetching that nodes currently do (`request_txpool_complement`).~ (edit: ignore this, we'd still want the txpool complement in addition to this)

> I haven't had the chance to review this at length yet, but it would be nice if it also implemented some max limit of txs when responding to NOTIFY_REQUEST_TX_POOL_TXS, as opposed to relaying all missing txs in a single response. This avoids problems when the pool exceeds the max packet size limit.

This is a valid concern IMHO.

jeffro256:
We ran into the issue of missing transactions exceeding payload limitations during the stressnet. It was fixed by https://github.com/seraphis-migration/monero/pull/168. Perhaps a similar tactic could be used here.

j-berman:
> since this PR only applies to brand new txs as they are relayed on the network

Correction: the entire pool can get re-relayed every x interval (w/exponential backoff delay) [here](https://github.com/monero-project/monero/blob/d32b5bfe18e2f5b979fa8dc3a8966c76159ca722/src/cryptonote_core/cryptonote_core.cpp#L1106), so if a connection is missing the whole pool, they'll request the whole pool.

So I do think this PR should probably have a solution for this. And the approach from https://github.com/seraphis-migration/monero/pull/168 seems simple/sane to implement [here](https://github.com/monero-project/monero/pull/9933#discussion_r2417971370) too.

src/cryptonote_protocol/request_manager.h
0xFFFC0000: Fixed: https://github.com/monero-project/monero/pull/9933/commits/e8ed4da9cc2a52142f7294dafc0cfda41de05d16

src/cryptonote_protocol/txrequestqueue.h
0xFFFC0000: Fixed: https://github.com/monero-project/monero/pull/9933/commits/64121e9df24f60b071971edef4a80d9120ce5a8e

src/cryptonote_protocol/request_manager.h
0xFFFC0000: Fixed: https://github.com/monero-project/monero/pull/9933/commits/64121e9df24f60b071971edef4a80d9120ce5a8e

src/cryptonote_protocol/txrequestqueue.h
0xFFFC0000: Fixed: https://github.com/monero-project/monero/pull/9933/commits/2875352e33ed4b401a90daff6d8e3a71dbd8cda5

src/cryptonote_protocol/txrequestqueue.h
0xFFFC0000: Fixd: https://github.com/monero-project/monero/pull/9933/commits/652b4189710dcacf461e85c5f6f8f928caeff52a#diff-1d03e4b2c6d1df00db873e159c99573f4c9c7a4ff7cbc85ebdf28a1965c08fa2R96

src/cryptonote_protocol/txrequestqueue.h
0xFFFC0000: Fixd: https://github.com/monero-project/monero/pull/9933/commits/652b4189710dcacf461e85c5f6f8f928caeff52a#diff-1d03e4b2c6d1df00db873e159c99573f4c9c7a4ff7cbc85ebdf28a1965c08fa2R96

contrib/epee/include/syncobj.h
0xFFFC0000: Fixd: https://github.com/monero-project/monero/pull/9933/commits/652b4189710dcacf461e85c5f6f8f928caeff52a

src/cryptonote_protocol/request_manager.h
0xFFFC0000: Fixd: https://github.com/monero-project/monero/pull/9933/commits/652b4189710dcacf461e85c5f6f8f928caeff52a

src/cryptonote_protocol/cryptonote_protocol_handler.h
0xFFFC0000: Fixd: https://github.com/monero-project/monero/pull/9933/commits/652b4189710dcacf461e85c5f6f8f928caeff52a

src/cryptonote_basic/connection_context.h
0xFFFC0000: Fixd: https://github.com/monero-project/monero/pull/9933/commits/652b4189710dcacf461e85c5f6f8f928caeff52a

src/cryptonote_protocol/request_manager.h
0xFFFC0000: Fixed: https://github.com/monero-project/monero/pull/9933/commits/652b4189710dcacf461e85c5f6f8f928caeff52a

src/cryptonote_protocol/request_manager.h
0xFFFC0000: After erasing in-flight requests for a given transaction (`tx_hash`), these lines **refresh the index** (`by_tx_hash`) to ensure it reflects the current state of `m_requested_txs`.

src/cryptonote_protocol/request_manager.h
0xFFFC0000: https://github.com/monero-project/monero/pull/9933#discussion_r2473535129

src/cryptonote_basic/connection_context.h
0xFFFC0000: You need it as `shared_ptr`
```
monero/src/p2p/net_node.inl: In instantiation of ‘int nodetool::node_server<t_payload_net_handler>::handle_handshake(int, typename COMMAND_HANDSHAKE::request&, typename COMMAND_HANDSHAKE::response&, p2p_connection_context&) [with t_payload_net_handler = cryptonote::t_cryptonote_protocol_handler<cryptonote::core>; typename COMMAND_HANDSHAKE::request = epee::misc_utils::struct_init<nodetool::COMMAND_HANDSHAKE_T<cryptonote::CORE_SYNC_DATA>::request_t>; COMMAND_HANDSHAKE = nodetool::COMMAND_HANDSHAKE_T<cryptonote::CORE_SYNC_DATA>; typename COMMAND_HANDSHAKE::response = epee::misc_utils::struct_init<nodetool::COMMAND_HANDSHAKE_T<cryptonote::CORE_SYNC_DATA>::response_t>; p2p_connection_context = nodetool::p2p_connection_context_t<cryptonote::cryptonote_connection_context>]’:
monero/src/rpc/instanciations.cpp:36:37:   required from here
   36 | namespace nodetool { template class node_server<cryptonote::t_cryptonote_protocol_handler<cryptonote::core>>; }
      |                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
monero/src/p2p/net_node.inl:2600:40: error: use of deleted function ‘nodetool::p2p_connection_context_t<cryptonote::cryptonote_connection_context>::p2p_connection_context_t(const nodetool::p2p_connection_context_t<cryptonote::cryptonote_connection_context>&)’
 2600 |       try_ping(arg.node_data, context, [peer_id_l, port_l, context, this]()
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 2601 |       {
      |       ~
      
```      

Some code (in this case, in try_ping or a lambda) tries to copy the context object, but the compiler cannot generate a copy constructor due to the non-copyable member.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
0xFFFC0000: Thanks. Fixed: https://github.com/monero-project/monero/pull/9933/commits/2fba2462cc86c8f32162ed06c089a818fa4aae32

src/cryptonote_protocol/cryptonote_protocol_handler.inl
0xFFFC0000: I need both to do a lookup.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
0xFFFC0000: Fixed: https://github.com/monero-project/monero/pull/9933/commits/6466070de9227cb2793bcb5483fd7b0f201fe165

src/cryptonote_protocol/cryptonote_protocol_handler.inl
0xFFFC0000: Thanks. Good catch. 

Fixed: https://github.com/monero-project/monero/pull/9933/commits/afff91006143a127dce9ab295ad4e525e0efa24f

0xFFFC0000:
Not necessarily incorrect,, need investigation:

> failure rate calculation


> In request_from_next_peer, reassigning the index reference *might be* wrong


> Erasing requests while iterating them in `check_tx_request_queue`


> What will happen if multiple peers have requested the same TX? a block confirmation may leave one peer’s in_flight_requests count incorrect.

0xFFFC0000:
Updated version of this code is available here: https://github.com/seraphis-migration/monero/pull/184 

It is actively under test, and will be merged into this PR

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1565,6 +1691,37 @@ namespace cryptonote
...
+              if (tx_hashes_ptr)
+              {
+                for (const auto &h : *tx_hashes_ptr)
+                  if (m_request_manager.already_requested_tx(h))
+                  {
+                    boost::uuids::uuid owner;
+                    if (m_request_manager.get_current_request_peer_id(h, owner) && !owner.is_nil())
+                    {
+                        m_p2p->for_connection(owner, [&](cryptonote_connection_context &ctx, nodetool::peerid_type, uint32_t)->bool {
+                            ctx.remove_in_flight_request();
```
SChernykh: If you call `remove_in_flight_request` here, but that peer already received it and we haven't got the response yet (ping is not 0, so network delays are in play), will it `remove_in_flight_request` again when this response arrives?

src/cryptonote_basic/connection_context.h
```diff
@@ -32,14 +32,59 @@
...
 #include "crypto/hash.h"
+#include "cryptonote_protocol/cryptonote_protocol_defs.h"
+#include "net/net_utils_base.h"
+#include "p2p/p2p_protocol_defs.h"
+#include "syncobj.h"
 
 namespace cryptonote
 {
+  //! \return Maximum number of bytes permissible for `command`.
+  inline std::size_t get_command_max_bytes(const int command) noexcept
```
vtnerd: I don't understand why this was moved to the header. It's marginally faster if inlined?

And if we're going to be put it here, upgrade this to `constexpr` I guess?

nahuhh:
This pr is missing https://github.com/seraphis-migration/monero/commit/b144d39b01924b5d6f7fb1c5ba7a61c4541d6f93

navidR:
3 separate steps: 

1. Squashed to a single commit without any change: https://github.com/monero-project/monero/compare/880e90b83d8ecdfb82617eb4b393a4d9e765949a..a57e274d30608e03a792259119e6e35ed8d3a555

2. Added https://github.com/seraphis-migration/monero/commit/b144d39b01924b5d6f7fb1c5ba7a61c4541d6f93 to address https://github.com/monero-project/monero/pull/9933#issuecomment-3981305648: https://github.com/monero-project/monero/compare/a57e274d30608e03a792259119e6e35ed8d3a555..483d7de20d6ad999a3ac9b372ef19ff4dd4c5dcf

3. Rebased to the latest master without any content change to this PR: https://github.com/monero-project/monero/compare/4c8f68c7cb2d23507e3c6ccf91e12d639b822c8b..8ff4d92111cea94dabb3e85c12e5c55d92354b01

src/cryptonote_basic/connection_context.h
navidR: Thanks. Addressed by making it `constexpr`.

vtnerd: Why was it moved to the header though? This isn't going to be executed as `constexpr`, as it requires a runtime value, so why move it to the header?

navidR: You are right about your concern. But there is one minor case that I think you are missing. That minor case might not be very important. But still it is valid. Look at these: 

 ```
 // src/cryptonote_protocol/cryptonote_protocol_handler.inl:188
 static const std::size_t MAX_N_TXS =
  get_command_max_bytes(NOTIFY_NEW_TRANSACTIONS::ID) / get_max_tx_size();
  ```

  and

  ```
  // src/cryptonote_protocol/cryptonote_protocol_handler.inl:195
  static const std::size_t MAX_N_TX_HASHES =
  get_command_max_bytes(NOTIFY_TX_POOL_HASH::ID) / sizeof(crypto::hash) * 95 / 100;
  ```
  
The point in both cases `get_command_max_bytes` ( not the whole expression, just that function) will be evaluated at compile time.

These are optimizable with `constexpr`, guaranteed. I know with constant-folding, the compiler will try to optimize these even without `constexpr`, but that is relying on compiler behaviour. 


We are getting into the nitty-gritty of the C++ standard here. So I leave the final decision for you. If you want to move it to `.cpp`, let me know.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -876,33 +907,108 @@ namespace cryptonote
...
+        txs.push_back(std::move(tx_blob));
+      }
+      // If tx is not in the pool, then ignore it (do not penalize peer)
+    }
+
+    // Send response if any txs found
+    if (!txs.empty())
+    {
+      NOTIFY_NEW_TRANSACTIONS::request request = {};
+      request.txs = std::move(txs);
```
vtnerd: See this [patch](https://github.com/seraphis-migration/monero/pull/297) in seraphis branches. You'll need to add something similar here - the message should be sending txes in fluff mode.

vtnerd (changes requested):

src/cryptonote_protocol/cryptonote_protocol_handler.inl
navidR: Thanks. Applied here: https://github.com/monero-project/monero/pull/9933/changes/9ac0d39083244c1bac0eaec1defbd3f6f33392b1#diff-d32229ecf55d414dcca43b6918ad711e1ffd313ef851492ecba1865323007f0fR1000

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -179,7 +182,34 @@ namespace cryptonote
...
+  static void shrink_to_fit_hashes_container(std::vector<crypto::hash> &hashes_inout)
+  {
+    static const std::size_t max_n_hashes = max_n_tx_hashes_per_packet();
+    if (max_n_hashes >= hashes_inout.size())
+      return;
 
+    // If this hits, it means we should look into updating the protocol for filling a node's empty txpool.
+    // The pool complement protocol is not designed to handle refilling huge pools. We truncate to avoid the worst case
+    // of connection drops, since missing pool txs is more desirable than dropped peers. But it would be better to have
+    // a protocol capable of refilling an empty pool when the pool is large.
```
jeffro256, line 202: Why not take a similar approach to https://github.com/seraphis-migration/monero/pull/168?

j-berman: Linking https://github.com/seraphis-migration/monero/pull/288#discussion_r2906786976

tests/unit_tests/request_manager.cpp
```diff
@@ -0,0 +1,172 @@
...
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "gtest/gtest.h"
+
+#include "cryptonote_protocol/request_manager.h"
+
+#include <boost/uuid/uuid.hpp>
```
jeffro256, line 33:
```suggestion
#include <boost/uuid/uuid.hpp>
#include <unistd.h>
```

Adding header include fixes Msys2 builds

src/cryptonote_protocol/request_manager.cpp
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    MINFO("Removing stale request for tx " << it->tx_hash 
+          << " from peer " << epee::string_tools::pod_to_hex(it->peer_id)
+          << ", age: " << elapsed.count() << "ms");
+
+    // If this peer has missed too many requests, we want to drop it
+    if (this->missed_request(it->peer_id, 1)) {
+      drop_peers_out.insert(it->peer_id);
+    }
```
selsta, line 80:
```suggestion
    const auto &peer_id = it->peer_id;

    if (m_connection_stats[peer_id].in_flight_requests > 0)
      --m_connection_stats[peer_id].in_flight_requests;

    // If this peer has missed too many requests, we want to drop it
    if (this->missed_request(peer_id, 1)) {
      drop_peers_out.insert(peer_id);
    }
```

https://github.com/seraphis-migration/monero/pull/384

src/cryptonote_protocol/request_manager.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+
+#include "cryptonote_protocol/request_manager.h"
+#include "cryptonote_protocol/txrequestqueue.h"
+#include "misc_log_ex.h"
+#include <cstddef>
+#include <cstdint>
+
+void request_manager::remove_peer(const boost::uuids::uuid &peer_id) {
+  MINFO("Removing all requests for disconnected peer: " << epee::string_tools::pod_to_hex(peer_id));
```
j-berman: nit: I think we should replace including `string_tools.h` with `#include <boost/uuid/uuid.hpp>`, and log all `peer_id`'s in this file like this ` << peer_id` without the `pod_to_hex`.

Can see elsewhere in the code this is done, e.g. [here](https://github.com/monero-project/monero/blob/49c9cfc31568e6e56238a7f7bc8814a4f0c59b24/src/cryptonote_protocol/levin_notify.cpp#L568). boost places the UUID -'s when logging, and `pod_to_hex` doesn't (`2fb59d58-6c29-4579-a79e-4d8f1534a957` vs `2fb59d586c294579a79e4d8f1534a957`). So reusing boost's << operator here actually makes it easier to see what's happening in the logs by making it easier to grep for a specific peer id and see what's happening with it in other places.

navidR: Applied, thanks; peer IDs in request_manager.cpp now stream as Boost UUIDs directly, and the unused string_tools.h include was removed.

j-berman:
[This commit](https://github.com/seraphis-migration/monero/pull/450/changes/e81579a048e6770764c8efa4451973896b61beeb) includes another round of significant changes to the protocol, including a new field in the p2p message. [This PR](https://github.com/seraphis-migration/monero/pull/450) explains the rationale for the changes. That round of changes is slated for testing in the beta stressnet v3.

navidR:
> [This commit](https://github.com/seraphis-migration/monero/pull/450/changes/e81579a048e6770764c8efa4451973896b61beeb) includes another round of significant changes to the protocol, including a new field in the p2p message. [This PR](https://github.com/seraphis-migration/monero/pull/450) explains the rationale for the changes. That round of changes is slated for testing in the beta stressnet v3.

Thanks. Updated the PR with the rebase and cherry-picked e81579a as 4220eb804. Builds and focused tests pass.

PR #9934, opened 2025-05-23
wallet2: don't expand subaddress table when importing cold outputs

jeffro256:
The hot wallet can pass as the subaddresss index (0xFFFFFFFF, 0xFFFFFFFF), and the cold wallet will generate the key image as long as that subaddress is valid for the passed onetime address (which doesn't actually need to be present on-chain). Then, immediately after, it will sit there and try to expand its subaddress table to <code>m_subaddress_lookahead_minor * 2<sup>32</sup></code> items for basically no reason. An optimistically expanded subaddress table is really only useful for on-chain scanning, which a cold wallet wouldn't be doing. It only needs the subaddresses for the outputs it knows it owns.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -14323,8 +14323,6 @@ process:
       create_one_off_subaddress(td.m_subaddr_index);
     bool r = cryptonote::generate_key_image_helper(m_account.get_keys(), m_subaddresses, out_key, tx_pub_key, additional_tx_pub_keys, td.m_internal_output_index, in_ephemeral, td.m_key_image, m_account.get_device());
     THROW_WALLET_EXCEPTION_IF(!r, error::wallet_internal_error, "Failed to generate key image");
-    if (should_expand(td.m_subaddr_index))
-      expand_subaddresses(td.m_subaddr_index);
     td.m_key_image_known = true;
     td.m_key_image_request = true;
     td.m_key_image_partial = false;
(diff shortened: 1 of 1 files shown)


jeffro256:
The subaddress table size is limited with on-chain scanning since it can only expand by whatever your lookahead value is each time it encounters an on-chain enote. However, during cold output importing, it can expand for any arbitrary value immediately.

jeffro256:
The problem is a bit deeper than I originally anticipated. Basically every place in `wallet2` expects the subaddress map to be contiguous, AKA if index (x, y) is in the map, then for all 0 <= y' < y, (x, y') is in the map. Likewise, if (x, 0) is in the map, then for all 0 <= x' < x, (x', 0) is in the map. And we can't effectively update the code to handle the case where this isn't true without bumping the wallet cache serialization version to update `m_subaddress_labels` and `m_account_tags` to allow them to be sparse (probably a `std::map`). 

Alternatively, we could remove the `get_num_subaddress_accounts()` and `get_num_subaddresses()` methods from wallet, and replace them with functions that return all present subaddress major indices, and minor indices for a given major index. Then, callers can iterate through these. Indices created as "one-offs" still wouldn't be able to have labels, which is kind of strange, but oh well.

The easiest route is to throw an error if the subaddress map would be expanded by more than the lookahead ...

jeffro256:
The reason I bring this up is because the functional tests fail with this commit as-is since the RPC wallet iterates through all subaddress indices less than `wallet2::get_num_subaddresses()` and returns those. But this misses subaddresses created as "one-offs", since it queries `m_subaddress_labels` instead of `m_subaddresses`. But if you look where the term `get_num_subaddress*` shows up in the codebase, you can see that this assumption is all over.

PR #9935, opened 2025-05-23
Improve initial refresh: use nearest checkpoint in fast_refresh

valldrac:
This PR updates `wallet2::fast_refresh` to use the nearest checkpoint at or below the restore height, instead of always using the latest checkpoint.

Previously, if a wallet was restored from a height earlier than the latest checkpoint (now 3375700), wallet2 would fallback to downloading block hashes from height 0 (genesis). This could take several minutes.

With this change, the fast refresh starts from the closest checkpoint below the restore height, avoiding the full hash fetch from genesis and making initial refresh faster for older wallets.

### DIFF
--- a/src/checkpoints/checkpoints.cpp
+++ b/src/checkpoints/checkpoints.cpp
@@ -158,6 +158,19 @@ namespace cryptonote
     return m_points.rbegin()->first;
   }
   //---------------------------------------------------------------------------
+  uint64_t checkpoints::get_nearest_checkpoint_height(uint64_t block_height) const
+  {
+    if (m_points.empty())
+      return 0;
+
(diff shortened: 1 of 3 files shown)


plowsof:
works as described :+1:

nahuhh:
@valldrac plz squash (and rebase at the same time, if you dont mind. Not needed for merge, but easier to test the pr. Cherry-pick works though so.. rebase not necessary)

Tested working btw

nahuhh (approved):

selsta (approved):

tobtoht:
This patch introduced a test failure: https://github.com/monero-project/monero/actions/runs/15655160526/job/44105225609#step:13:1864

PR #9936, opened 2025-05-23
wallet2: use IO threadpool for getblocks.bin network request

j-berman:
EDIT: I made a mistake and don't believe this PR should yield a significant improvement to the scanner. @plowsof is still observing a modest speed-bump with this code and I'm not certain what part of the code may be sped up. @nahuhh and @MrCyjaneK are reporting regressions, which I'm also not certain why. I can leave this PR open until the PR is more fully investigated.

Credit to @jeffro256 for opening #9917, which led me to realize this very simple change for potentially faster scanning with wallet2.

This PR uses a distinct threadpool for fetching blocks from the daemon.

EDIT: My original PR description was incorrect. I misread the threadpool waiter implementation and what I originally described is not actually happening. A distinct instantiated waiter will not wait on all tasks in the threadpool to complete upon calling `wait()`, it will wait for the tasks tied to the waiter to complete.

However, a single threadpool instance has a single queue. Which means all tasks submitted to the compute threadpool instance share the same queue. The network request gets queued into the compute threadpool, and should be the first task to execute because it should be the first task submitted. Then the expensive scanning tasks get submitted to the threadpool and can also begin executing immediately on a multithreaded device. Their respective waiters should only wait for their tasks to complete, not for the threadpool's queue to clear i.e. the local scanning task waiters won't wait on the network request to complete. @jeffro256's #9917 shows it's possible for starvation to occur with the current threadpool impl, however, wallet2's design does not run up against that case AFAICT.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -4017,8 +4017,8 @@ void wallet2::refresh(bool trusted_daemon, uint64_t start_height, uint64_t & blo
   size_t try_count = 0;
   crypto::hash last_tx_hash_id = m_transfers.size() ? m_transfers.back().m_txid : null_hash;
   std::list<crypto::hash> short_chain_history;
-  tools::threadpool& tpool = tools::threadpool::getInstanceForCompute();
-  tools::threadpool::waiter waiter(tpool);
+  tools::threadpool& io_tpool = tools::threadpool::getInstanceForIO();
+  tools::threadpool::waiter io_waiter(io_tpool);
   uint64_t blocks_start_height;
   std::vector<cryptonote::block_complete_entry> blocks;
   std::vector<parsed_block> parsed_blocks;
@@ -4090,7 +4090,7 @@ void wallet2::refresh(bool trusted_daemon, uint64_t start_height, uint64_t & blo
         break;
       }
       if (!last)
-        tpool.submit(&waiter, [&]{pull_and_parse_next_blocks(first, try_incremental, start_height, next_blocks_start_height, short_chain_history, blocks, parsed_blocks, next_blocks, next_parsed_blocks, process_pool_txs, last, error, exception);});
+        io_tpool.submit(&io_waiter, [&]{pull_and_parse_next_blocks(first, try_incremental, start_height, next_blocks_start_height, short_chain_history, blocks, parsed_blocks, next_blocks, next_parsed_blocks, process_pool_txs, last, error, exception);});
 
       if (!first)
       {
@@ -4129,7 +4129,7 @@ void wallet2::refresh(bool trusted_daemon, uint64_t start_height, uint64_t & blo
         }
         blocks_fetched += added_blocks;
       }
-      THROW_WALLET_EXCEPTION_IF(!waiter.wait(), error::wallet_internal_error, "Exception in thread pool");
+      THROW_WALLET_EXCEPTION_IF(!io_waiter.wait(), error::wallet_internal_error, "Exception in io thread pool");
 
       // handle error from async fetching thread
       if (error)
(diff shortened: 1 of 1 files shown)


SChernykh (approved):

SChernykh:
LGTM. I was the on who originally split the thread pool into compute and I/O pools. Tasks like this which require waiting on I/O (disc or network) should use the latter.

plowsof (approved):

0xFFFC0000 (approved):
I approving. But the real issue is not what @jeffro256 was trying to solve. The real issue is this one: https://github.com/monero-project/monero/pull/9768#issuecomment-2636444224

src/wallet/wallet2.cpp
```diff
@@ -4017,7 +4017,7 @@ void wallet2::refresh(bool trusted_daemon, uint64_t start_height, uint64_t & blo
   size_t try_count = 0;
   crypto::hash last_tx_hash_id = m_transfers.size() ? m_transfers.back().m_txid : null_hash;
   std::list<crypto::hash> short_chain_history;
-  tools::threadpool& tpool = tools::threadpool::getInstanceForCompute();
+  tools::threadpool& tpool = tools::threadpool::getInstanceForIO();
   tools::threadpool::waiter waiter(tpool);
```
jeffro256, line 4021: Perhaps these can be renamed to something less generic that points to its usage, like `io_tpool` and `pull_parse_waiter` or something of that nature. The generic naming already present here probably contributed to the original mistake of using the compute threadpool versus the I/O threadpool.

jeffro256: Good find!

selsta:
@j-berman please squash

plowsof:
I was seeing a verifiable almost 2x improvement in sync times - i am currently unable to reproduce this gain (was it an unfortunate coincidence of the node i was using recovering from a ddos? or my own connection? or a specific block range where they where mostly empty?)

after 15 attempts syncing from the same blockheight 3mill as @nahuhh with / without the PR , sometimes quicker, sometimes slower it averaged out to be slightly quicker (the orange line is using this PR) but its very unreliable to get an accurate test.

   <details>
      <summary>graph image</summary>

sync_comparison

    </details>

plowsof:
@MrCyjaneK to sanity check a bit more (not entirely reproducing your test:) the difference is im deleting the cache file before starting and not calling any _change_ restore height and _rescan_ ?

- open wallet.keys without cache of wallet with sync height 0
- setup background sync
- start background sync
- wait until get_height >= 300k

   <details>
      <summary>graph image</summary>

    sync_comparison

    </details>

im still seeing this as being slightly faster even with background sync. 

The 2x was seen not in total time to sync wallet as we are testing here, but rather in part of the scanner code. which i will confirm tomorrow

j-berman:
Adding to above, @MrCyjaneK I'm not sure what could be causing your slowdown yet and wasn't able to repro your issue with the CLI. I asked @plowsof to see if they could repro it too. It would be ideal to have a complete reproducible minimal setup that repro's it (with the RPC if possible). Looking at the code, a slowdown like that looks unrelated to this PR, but maybe I'm missing something.

j-berman:
> The 2x was seen not in total time to sync wallet as we are testing here, but rather in part of the scanner code. which i will confirm tomorrow

Also expanding on this, I initially asked @plowsof to time how long it takes to restore a wallet *excluding* [this part](https://github.com/monero-project/monero/blob/125622d5bdc42cf552be5c25009bd9ab52c0a7ca/src/wallet/wallet2.cpp#L4033-L4046) where the scanner retrieves block hashes. That section should only execute slowly on wallet restore (and technically the scanner doesn't even have to do it, the whole section can be taken out in the future by someone willing to take the time to take it out correctly + #9935 would speed that up). Also that section is completely unaffected by this PR. The part this PR should speed up is scanning blocks to identify the wallet's txs when using a remote daemon, not the part where wallet2 downloads block hashes.

It's also possible that downloading over tor is just so slow that the time wallets spend doing the actual local scanning is insignificant, and so this PR's change is insignificant for wallets pointing to daemons over tor. Probably makes sense that wallets pointing to clearnet remote daemons (and also slow machines) would realize the most benefit from this PR. It doesn't make sense to me why there would be a slowdown compared to master though, except for tor's inconsistency. This PR should strictly improve scanning perf over a remote connection.

nahuhh:
> It's also possible that downloading over tor is just so slow that the time wallets spend doing the actual local scanning is insignificant, and so this PR's change is insignificant for wallets pointing to daemons over tor.

my tests show that this pr has significantly slower results over tor. 

> Probably makes sense that wallets pointing to clearnet remote daemons (and also slow machines) would realize the most benefit from this PR.

my tests show that there was (literally) 0 difference between master and this pr when syncing over lan

> It doesn't make sense to me why there would be a slowdown compared to master though, except for tor's inconsistency. This PR should strictly improve scanning perf over a remote connection.

the tests were back to back, and every time (over tor), this pr was significantly slower. Dont ask me why, but the results were repeatable. 
The only outlier was the very first test, probably some external factor.

> Also that section is completely unaffected by this PR. The part this PR should speed up is scanning blocks to identify the wallet's txs when using a remote daemon, not the part where wallet2 downloads block hashes.

i'll do a few 10min runs with 9935 pulled in to skip majority of block hashes

j-berman:
> my tests show that there was (literally) 0 difference between master and this pr when syncing over lan

Ya, this isn't surprising. I was referring to clearnet as in over the internet (not over a LAN).

ACK on your consistent observed slower results over tor. That is surprising and unexpected.

nahuhh:
> Ya, this isn't surprising. I was referring to clearnet as in over the internet (not over a LAN).

my router is pretty slow. Observe about 1.2MiB/s downloads during sync. Much worse my phones LTE or 5G (usually see about 10MiB/sec) so i figured it would be a a good test.

Ive pulled in 9935 and am running some more tests now

plowsof:
using this script to put times / height to a file (currently set to call setup / start background sync the nloop get height but it can be changed to get_height only) https://gist.github.com/plowsof/666afafce9ecdf0764960821ecc8b675

`rescan_blockchain` will make the wallet-rpc unresponsive (but it continues to work when looking at log 4) - i did not check for a way to programatically change wallet height with rpc yet , if this is needed as a data point.

now, the lan vs tor vs wifi network issue - `trickle` can be used on binaries with shared libc library. which self compiled monero-wallet-rpc does have so you can (-d download -u upload in KB/s) `trickle -s -d 500 -u 500 monero-wallet-rpc <args>` will let rpc run until.... nothing i had to increase limits to `trickle -s -d 100000 -u 100000 monero-wallet-rpc` to make it work....  im not sure if thats a trickle thing or if wallet rpc simply does not work with low transfer speeds

j-berman:
I made a mistake and don't believe this PR should yield a significant improvement to the scanner. I apologize. I've edited the PR description removing inaccuracies.

@plowsof is still observing a modest speed-bump with this code and I'm not certain what part of the code may be sped up. @nahuhh and @MrCyjaneK are reporting regressions, which I'm also not certain why. I can leave this PR open until the PR is more fully investigated.

SChernykh:
I think it's because of random selection of nodes they sync from each time. Only LAN tests between the same two nodes can be realiable, but even then I'm not sure if monerod doesn't do some block batches in random order or with random sizes.

nahuhh:
I tested from my own node each time, using an identical key file, cache file deleted between runs.

didnt restart tor or any devices, tests were back -to-back

Lan saw 0 speedup or slowdown (even had 2 identical runs), tor saw notable slowdowns, bur could just be incredibly unlucky.

PR #9937, opened 2025-05-27
wallet: Reset rpc-server refresh interval post-full sync

Tzadiko:
This commit fd08594caae37a592a6e9257e6ad577d9f8f9c32 introduced in this MR https://github.com/monero-project/monero/pull/9875 changed the refresh interval to 200ms during wallet sync, fixing other bugs in the process. 

Here I've done two things:

1. Recognize that m_auto_refresh_period can change, and protect it from multithreaded access via an atomic (it can change when a user sends a request to stop auto refresh, and its polled for in a separate thread that runs every 200ms during sync and 1000ms when we reach tip of chain).
2. Remove the if-condition that prevented us from executing tip-of chain logic that would transition us to the less-frequent polling once we arrive at the tip of the chain.

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -200,15 +200,15 @@ namespace tools
   {
     m_stop = false;
 
-    const bool enable_auto_refresh = m_auto_refresh_period != 0;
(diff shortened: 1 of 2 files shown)


nahuhh:
Thank you 🔥

vtnerd (approved):

PR #9939, opened 2025-05-30
p2p: Improved peer selection with /24 subnet deduplication to disadvantage 'spy nodes'

rbrunner7:
For background info about those "spy nodes" check [this GitHub issue](https://github.com/monero-project/research-lab/issues/126), [this Reddit post](https://www.reddit.com/r/Monero/comments/1hc1ozf/mrl_recommendation_ban_spy_node_ip_addresses_from/) and [this wonderful work from Rucknium](https://github.com/Rucknium/misc-research/blob/main/Monero-Peer-Subnet-Deduplication/pdf/monero-peer-subnet-deduplication.pdf).

This PR was developed in close cooperation with @Rucknium who gave much valuable input and also actually checked that the new method improves things: The daemon connects to considerably less "spy nodes" with it.

As I rewrote large parts of the `make_new_connection_from_peerlist` method, the diff with the old version of the method comes out quite confusing. Maybe for review hold the two full method sources side-by-side.

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -1414,7 +1414,7 @@ namespace nodetool
     }
 
 
-    MDEBUG("Connecting to " << na.str() << "(peer_type=" << peer_type << ", last_seen: "
+    MDEBUG("Connecting to " << na.str() << " (peer_type=" << peer_type << ", last_seen: "
         << (last_seen_stamp ? epee::misc_utils::get_time_interval_string(time(NULL) - last_seen_stamp):"never")
         << ")...");
 
@@ -1571,33 +1571,78 @@ namespace nodetool
     return false;
   }
   //-----------------------------------------------------------------------------------
+  // Find a single candidate from the given peer list in the given zone and connect to it if possible
   template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::make_new_connection_from_peerlist(network_zone& zone, bool use_white_list)
   {
-    size_t max_random_index = 0;
 
-    std::set<size_t> tried_peers;
+    // Local helper method to get the host string, i.e. the pure IP address without port
+    auto get_host_string = [](const epee::net_utils::network_address &address) {
+      if (address.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
+      {
+        boost::asio::ip::address_v6 actual_ip = address.as<const epee::net_utils::ipv6_network_address>().ip();
+        if (actual_ip.is_v4_mapped())
+        {
+          boost::asio::ip::address_v4 v4ip = make_address_v4_from_v6(actual_ip);
+          uint32_t actual_ipv4;
+          memcpy(&actual_ipv4, v4ip.to_bytes().data(), sizeof(actual_ipv4));
+          return epee::net_utils::ipv4_network_address(actual_ipv4, 0).host_str();
+        }
+      }
+      return address.host_str();
+    };
 
-    size_t try_count = 0;
+    // Get the current list of known peers of the desired kind, ordered by 'last_seen'.
+    // Deduplicate ports right away, i.e. if we know several peers on the same host address but
+    // with different ports, take only one of them, to avoid giving such peers undue weight
+    // and make it impossible to game peer selection by advertising on a large number of ports.
+    // Making this list also insulates us from any changes that may happen to the original list
+    // while we are working here, and allows an easy retry in the inner try loop.
+    std::vector<peerlist_entry> peers;
+    std::unordered_set<std::string> hosts;
+    size_t total_peers_size = 0;
+    zone.m_peerlist.foreach(use_white_list, [&peers, &hosts, &total_peers_size, get_host_string](const peerlist_entry &peer)
+    {
+      ++total_peers_size;
+      const std::string host_string = get_host_string(peer.adr);
+      const auto it = hosts.find(host_string);
+      if (it == hosts.end())
+      {
+        peers.push_back(peer);
+        hosts.insert(host_string);
+      }
+      // else ignore this additional peer on the same IP number
+
+      return true;
+    });
+
+    size_t peers_size = peers.size();
+    MDEBUG("Looking at " << peers_size << " port-deduplicated peers out of " << total_peers_size
+      << ", i.e. dropping " << (total_peers_size - peers_size));
+
+    std::set<uint64_t> tried_peers;  // all peers ever tried
+
+    // Outer try loop, with up to 3 attempts to actually connect to a suitable randomly choosen candidate
     size_t rand_count = 0;
-    while(rand_count < (max_random_index+1)*3 &&  try_count < 10 && !zone.m_net_server.is_stop_signal_sent())
+    while ((rand_count < 3) && !zone.m_net_server.is_stop_signal_sent())
     {
       ++rand_count;
-      size_t random_index;
+
       const uint32_t next_needed_pruning_stripe = m_payload_handler.get_next_needed_pruning_stripe().second;
 
-      // build a set of all the /16 we're connected to, and prefer a peer that's not in that set
-      std::set<uint32_t> classB;
-      if (&zone == &m_network_zones.at(epee::net_utils::zone::public_)) // at returns reference, not copy
+      // Build a list of all distinct /24 subnets we are connected to now right now; to catch
+      // any connection changes, re-build the list for every outer try loop pass
+      std::set<uint32_t> connected_subnets;
+      bool is_public_zone = &zone == &m_network_zones.at(epee::net_utils::zone::public_);
+      if (is_public_zone)
       {
         zone.m_net_server.get_config_object().foreach_connection([&](const p2p_connection_context& cntxt)
         {
           if (cntxt.m_remote_address.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id())
           {
-
             const epee::net_utils::network_address na = cntxt.m_remote_address;
             const uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
-            classB.insert(actual_ip & 0x0000ffff);
+            connected_subnets.insert(actual_ip & 0x00ffffff);
           }
           else if (cntxt.m_remote_address.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
           {
@@ -1608,89 +1653,117 @@ namespace nodetool
               boost::asio::ip::address_v4 v4ip = make_address_v4_from_v6(actual_ip);
               uint32_t actual_ipv4;
               memcpy(&actual_ipv4, v4ip.to_bytes().data(), sizeof(actual_ipv4));
-              classB.insert(actual_ipv4 & ntohl(0xffff0000));
+              connected_subnets.insert(actual_ipv4 & ntohl(0xffffff00));
             }
           }
           return true;
         });
       }
 
-      auto get_host_string = [](const epee::net_utils::network_address &address) {
-        if (address.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
-        {
-          boost::asio::ip::address_v6 actual_ip = address.as<const epee::net_utils::ipv6_network_address>().ip();
-          if (actual_ip.is_v4_mapped())
-          {
-            boost::asio::ip::address_v4 v4ip = make_address_v4_from_v6(actual_ip);
-            uint32_t actual_ipv4;
-            memcpy(&actual_ipv4, v4ip.to_bytes().data(), sizeof(actual_ipv4));
-            return epee::net_utils::ipv4_network_address(actual_ipv4, 0).host_str();
-          }
-        }
-        return address.host_str();
-      };
-      std::unordered_set<std::string> hosts_added;
-      std::deque<size_t> filtered;
-      const size_t limit = use_white_list ? 20 : std::numeric_limits<size_t>::max();
+      std::vector<peerlist_entry> *candidate_peers;
+      std::vector<peerlist_entry> subnet_peers;
+      std::vector<peerlist_entry> filtered;
+
+      // Inner try loop: Find candidates first with subnet deduplication, if none found again without.
+      // Finding none happens if all candidates are from subnets we are already connected to and/or
+      // they don't offer the needed stripe when pruning. Only actually loop and deduplicate if we are
+      // in the public zone because private zones don't have subnets.
       for (int step = 0; step < 2; ++step)
       {
-        bool skip_duplicate_class_B = step == 0;
-        size_t idx = 0, skipped = 0;
-        zone.m_peerlist.foreach (use_white_list, [&classB, &filtered, &idx, &skipped, skip_duplicate_class_B, limit, next_needed_pruning_stripe, &hosts_added, &get_host_string](const peerlist_entry &pe){
-          if (filtered.size() >= limit)
-            return false;
-          bool skip = false;
-          if (skip_duplicate_class_B && pe.adr.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id())
-          {
-            const epee::net_utils::network_address na = pe.adr;
-            uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
-            skip = classB.find(actual_ip & 0x0000ffff) != classB.end();
-          }
-          else if (skip_duplicate_class_B && pe.adr.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
-          {
-            const epee::net_utils::network_address na = pe.adr;
-            const boost::asio::ip::address_v6 &actual_ip = na.as<const epee::net_utils::ipv6_network_address>().ip();
-            if (actual_ip.is_v4_mapped())
-            {
-              boost::asio::ip::address_v4 v4ip = make_address_v4_from_v6(actual_ip);
-              uint32_t actual_ipv4;
-              memcpy(&actual_ipv4, v4ip.to_bytes().data(), sizeof(actual_ipv4));
-              skip = classB.find(actual_ipv4 & ntohl(0xffff0000)) != classB.end();
-            }
-          }
-
-          // consider each host once, to avoid giving undue inflence to hosts running several nodes
-          if (!skip)
-          {
-            const auto i = hosts_added.find(get_host_string(pe.adr));
-            if (i != hosts_added.end())
-              skip = true;
-          }
-
-          if (skip)
-            ++skipped;
-          else if (next_needed_pruning_stripe == 0 || pe.pruning_seed == 0)
-            filtered.push_back(idx);
-          else if (next_needed_pruning_stripe == tools::get_pruning_stripe(pe.pruning_seed))
-            filtered.push_front(idx);
-          ++idx;
-          hosts_added.insert(get_host_string(pe.adr));
-          return true;
-        });
-        if (skipped == 0 || !filtered.empty())
+        if ((step == 1) && !is_public_zone)
           break;
-        if (skipped)
-          MDEBUG("Skipping " << skipped << " possible peers as they share a class B with existing peers");
-      }
+
+        candidate_peers = &peers;
+        if ((step == 0) && is_public_zone)
+        {
+          // Deduplicate subnets using 3 steps
+
+          // Step 1: Prepare to access the peers in a random order
+          std::vector<size_t> shuffled_indexes(peers.size());
+          std::iota(shuffled_indexes.begin(), shuffled_indexes.end(), 0);
+          std::shuffle(shuffled_indexes.begin(), shuffled_indexes.end(), crypto::random_device{});
+
+          // Step 2: Deduplicate by only taking 1 candidate from each /24 subnet that occurs, the FIRST
+          // candidate seen from each subnet within the now random order
+          std::set<uint32_t> subnets = connected_subnets;
+          for (size_t index : shuffled_indexes)
+          {
+            const peerlist_entry &peer = peers[index];
+            bool take = true;
+            if (peer.adr.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id())
+            {
+              const epee::net_utils::network_address na = peer.adr;
+              uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
+              uint32_t subnet = actual_ip & 0x00ffffff;
+              take = subnets.find(subnet) == subnets.end();
+              if (take)
+                // This subnet is now "occupied", don't take any more candidates from this one
+                subnets.insert(subnet);
+            }
+            else if (peer.adr.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
+            {
+              const epee::net_utils::network_address na = peer.adr;
+              const boost::asio::ip::address_v6 &actual_ip = na.as<const epee::net_utils::ipv6_network_address>().ip();
+              if (actual_ip.is_v4_mapped())
+              {
+                boost::asio::ip::address_v4 v4ip = make_address_v4_from_v6(actual_ip);
+                uint32_t actual_ipv4;
+                memcpy(&actual_ipv4, v4ip.to_bytes().data(), sizeof(actual_ipv4));
+                uint32_t subnet = actual_ipv4 & ntohl(0xffffff00);
+                take = subnets.find(subnet) == subnets.end();
+                if (take)
+                  subnets.insert(subnet);
+              }
+              // else 'take' stays true, we will take an IPv6 address that is not V4 mapped
+            }
+            if (take)
+              subnet_peers.push_back(peer);
+          }
+
+          // Step 3: Put back into order according to 'last_seen', i.e. most recently seen first
+          std::sort(subnet_peers.begin(), subnet_peers.end(), [](const peerlist_entry &a, const peerlist_entry &b)
+          {
+            return a.last_seen > b.last_seen;
+          });
+
+          size_t subnet_peers_size = subnet_peers.size();
+          MDEBUG("Looking at " << subnet_peers_size << " subnet-deduplicated peers out of " << peers_size
+            << ", i.e. dropping " << (peers_size - subnet_peers_size));
+
+          candidate_peers = &subnet_peers;
+        } // deduplicate
+        // else, for step 1 / second pass of inner try loop, take all peers from all subnets
+
+        // Take as many candidates as we need and care about stripes if pruning
+        const size_t limit = use_white_list ? 20 : std::numeric_limits<size_t>::max();
+        for (auto &peer : *candidate_peers) {
+          if (filtered.size() >= limit)
+            break;
+
+          if (next_needed_pruning_stripe == 0 || peer.pruning_seed == 0)
+            filtered.push_back(peer);
+          else if (next_needed_pruning_stripe == tools::get_pruning_stripe(peer.pruning_seed))
+            filtered.insert(filtered.begin(), peer);
+          // else wrong stripe, skip
+        }
+
+        if (!filtered.empty())
+          break;
+      } // inner try loop
+
       if (filtered.empty())
       {
         MINFO("No available peer in " << (use_white_list ? "white" : "gray") << " list filtered by " << next_needed_pruning_stripe);
         return false;
       }
+
+      size_t random_index;
       if (use_white_list)
       {
-        // if using the white list, we first pick in the set of peers we've already been using earlier
-        random_index = get_random_index_with_fixed_probability(std::min<uint64_t>(filtered.size() - 1, 20));
+        // If using the white list, we first pick in the set of peers we've already been using earlier;
+        // that "fixed probability" heavily favors the peers most recently seen in the candidate list
+        random_index = get_random_index_with_fixed_probability(filtered.size() - 1);
+
         CRITICAL_REGION_LOCAL(m_used_stripe_peers_mutex);
         if (next_needed_pruning_stripe > 0 && next_needed_pruning_stripe <= (1ul << CRYPTONOTE_PRUNING_LOG_STRIPES) && !m_used_stripe_peers[next_needed_pruning_stripe-1].empty())
         {
@@ -1698,10 +1771,10 @@ namespace nodetool
           m_used_stripe_peers[next_needed_pruning_stripe-1].pop_front();
           for (size_t i = 0; i < filtered.size(); ++i)
           {
-            peerlist_entry pe;
-            if (zone.m_peerlist.get_white_peer_by_index(pe, filtered[i]) && pe.adr == na)
+            const peerlist_entry &peer = filtered[i];
+            if (peer.adr == na)
             {
-              MDEBUG("Reusing stripe " << next_needed_pruning_stripe << " peer " << pe.adr.str());
+              MDEBUG("Reusing stripe " << next_needed_pruning_stripe << " peer " << peer.adr.str());
               random_index = i;
               break;
             }
@@ -1710,52 +1783,54 @@ namespace nodetool
       }
       else
         random_index = crypto::rand_idx(filtered.size());
-
       CHECK_AND_ASSERT_MES(random_index < filtered.size(), false, "random_index < filtered.size() failed!!");
-      random_index = filtered[random_index];
-      CHECK_AND_ASSERT_MES(random_index < (use_white_list ? zone.m_peerlist.get_white_peers_count() : zone.m_peerlist.get_gray_peers_count()),
-          false, "random_index < peers size failed!!");
 
-      if(tried_peers.count(random_index))
+      // We have our final candidate for this pass of the outer try loop
+      const peerlist_entry &candidate = filtered[random_index];
+
+      if (tried_peers.count(candidate.id))
+        // Already tried, don't try that one again
         continue;
-
-      tried_peers.insert(random_index);
-      peerlist_entry pe = AUTO_VAL_INIT(pe);
-      bool r = use_white_list ? zone.m_peerlist.get_white_peer_by_index(pe, random_index):zone.m_peerlist.get_gray_peer_by_index(pe, random_index);
-      CHECK_AND_ASSERT_MES(r, false, "Failed to get random peer from peerlist(white:" << use_white_list << ")");
-
-      ++try_count;
+      tried_peers.insert(candidate.id);
 
       _note("Considering connecting (out) to " << (use_white_list ? "white" : "gray") << " list peer: " <<
-          peerid_to_string(pe.id) << " " << pe.adr.str() << ", pruning seed " << epee::string_tools::to_string_hex(pe.pruning_seed) <<
-          " (stripe " << next_needed_pruning_stripe << " needed)");
+          peerid_to_string(candidate.id) << " " << candidate.adr.str() << ", pruning seed " << epee::string_tools::to_string_hex(candidate.pruning_seed) <<
+          " (stripe " << next_needed_pruning_stripe << " needed), in loop pass " << rand_count);
 
-      if(zone.m_our_address == pe.adr)
+      if (zone.m_our_address == candidate.adr)
+        // It's ourselves, obviously don't take that
         continue;
 
-      if(is_peer_used(pe)) {
+      if (is_peer_used(candidate)) {
         _note("Peer is used");
         continue;
       }
 
-      if(!is_remote_host_allowed(pe.adr))
+      if (!is_remote_host_allowed(candidate.adr)) {
+        _note("Not allowed");
         continue;
+      }
 
-      if(is_addr_recently_failed(pe.adr))
+      if (is_addr_recently_failed(candidate.adr)) {
+         _note("Recently failed");
         continue;
+      }
 
-      MDEBUG("Selected peer: " << peerid_to_string(pe.id) << " " << pe.adr.str()
-                    << ", pruning seed " << epee::string_tools::to_string_hex(pe.pruning_seed) << " "
-                    << "[peer_list=" << (use_white_list ? white : gray)
-                    << "] last_seen: " << (pe.last_seen ? epee::misc_utils::get_time_interval_string(time(NULL) - pe.last_seen) : "never"));
+      MDEBUG("Selected peer: " << peerid_to_string(candidate.id) << " " << candidate.adr.str()
+      << ", pruning seed " << epee::string_tools::to_string_hex(candidate.pruning_seed) << " "
+      << "[peer_list=" << (use_white_list ? white : gray)
+      << "] last_seen: " << (candidate.last_seen ? epee::misc_utils::get_time_interval_string(time(NULL) - candidate.last_seen) : "never"));
 
-      if(!try_to_connect_and_handshake_with_new_peer(pe.adr, false, pe.last_seen, use_white_list ? white : gray)) {
-        _note("Handshake failed");
+      time_t begin_connect = time(NULL);
+      if (!try_to_connect_and_handshake_with_new_peer(candidate.adr, false, candidate.last_seen, use_white_list ? white : gray)) {
+        time_t fail_connect = time(NULL);
+        _note("Handshake failed after " << epee::misc_utils::get_time_interval_string(fail_connect - begin_connect));
         continue;
       }
 
       return true;
-    }
+    } // outer try loop
+
     return false;
   }
   //-----------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/p2p/net_node.inl
```diff
@@ -1571,33 +1571,78 @@ namespace nodetool
...
   //-----------------------------------------------------------------------------------
+  // Find a single candidate from the given peer list in the given zone and connect to it if possible
   template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::make_new_connection_from_peerlist(network_zone& zone, bool use_white_list)
   {
-    size_t max_random_index = 0;
 
-    std::set<size_t> tried_peers;
+    // Local helper method to get the host string, i.e. the pure IP address without port
+    auto get_host_string = [](const epee::net_utils::network_address &address) {
```
jeffro256:
```suggestion
    const auto get_host_string = [](const epee::net_utils::network_address &address) {
```

src/p2p/net_node.inl
```diff
@@ -1571,33 +1571,78 @@ namespace nodetool
...
+      if (address.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
+      {
+        boost::asio::ip::address_v6 actual_ip = address.as<const epee::net_utils::ipv6_network_address>().ip();
```
jeffro256:
```suggestion
        const boost::asio::ip::address_v6 actual_ip = address.as<const epee::net_utils::ipv6_network_address>().ip();
```

src/p2p/net_node.inl
```diff
@@ -1571,33 +1571,78 @@ namespace nodetool
...
+        peers.push_back(peer);
+        hosts.insert(host_string);
+      }
+      // else ignore this additional peer on the same IP number
 
-    size_t try_count = 0;
+      return true;
+    });
+
+    size_t peers_size = peers.size();
```
jeffro256:
```suggestion
    const size_t peers_size = peers.size();
```

src/p2p/net_node.inl
```diff
@@ -1571,33 +1571,78 @@ namespace nodetool
...
+
       const uint32_t next_needed_pruning_stripe = m_payload_handler.get_next_needed_pruning_stripe().second;
 
-      // build a set of all the /16 we're connected to, and prefer a peer that's not in that set
-      std::set<uint32_t> classB;
-      if (&zone == &m_network_zones.at(epee::net_utils::zone::public_)) // at returns reference, not copy
+      // Build a list of all distinct /24 subnets we are connected to now right now; to catch
+      // any connection changes, re-build the list for every outer try loop pass
+      std::set<uint32_t> connected_subnets;
+      bool is_public_zone = &zone == &m_network_zones.at(epee::net_utils::zone::public_);
```
jeffro256:
```suggestion
      const bool is_public_zone = &zone == &m_network_zones.at(epee::net_utils::zone::public_);
```

src/p2p/net_node.inl
```diff
@@ -1608,100 +1653,128 @@ namespace nodetool
...
-          else if (next_needed_pruning_stripe == 0 || pe.pruning_seed == 0)
-            filtered.push_back(idx);
-          else if (next_needed_pruning_stripe == tools::get_pruning_stripe(pe.pruning_seed))
-            filtered.push_front(idx);
-          ++idx;
-          hosts_added.insert(get_host_string(pe.adr));
-          return true;
-        });
-        if (skipped == 0 || !filtered.empty())
+          size_t subnet_peers_size = subnet_peers.size();
```
jeffro256:
```suggestion
          const size_t subnet_peers_size = subnet_peers.size();
```

src/p2p/net_node.inl
```diff
@@ -1710,52 +1783,54 @@ namespace nodetool
...
-                    << "[peer_list=" << (use_white_list ? white : gray)
-                    << "] last_seen: " << (pe.last_seen ? epee::misc_utils::get_time_interval_string(time(NULL) - pe.last_seen) : "never"));
+      MDEBUG("Selected peer: " << peerid_to_string(candidate.id) << " " << candidate.adr.str()
+      << ", pruning seed " << epee::string_tools::to_string_hex(candidate.pruning_seed) << " "
+      << "[peer_list=" << (use_white_list ? white : gray)
+      << "] last_seen: " << (candidate.last_seen ? epee::misc_utils::get_time_interval_string(time(NULL) - candidate.last_seen) : "never"));
 
-      if(!try_to_connect_and_handshake_with_new_peer(pe.adr, false, pe.last_seen, use_white_list ? white : gray)) {
-        _note("Handshake failed");
+      time_t begin_connect = time(NULL);
```
jeffro256:
```suggestion
      const time_t begin_connect = time(NULL);
```

src/p2p/net_node.inl
```diff
@@ -1571,33 +1571,78 @@ namespace nodetool
...
+    // Get the current list of known peers of the desired kind, ordered by 'last_seen'.
+    // Deduplicate ports right away, i.e. if we know several peers on the same host address but
+    // with different ports, take only one of them, to avoid giving such peers undue weight
+    // and make it impossible to game peer selection by advertising on a large number of ports.
+    // Making this list also insulates us from any changes that may happen to the original list
+    // while we are working here, and allows an easy retry in the inner try loop.
+    std::vector<peerlist_entry> peers;
+    std::unordered_set<std::string> hosts;
+    size_t total_peers_size = 0;
+    zone.m_peerlist.foreach(use_white_list, [&peers, &hosts, &total_peers_size, get_host_string](const peerlist_entry &peer)
```
jeffro256: Why is `get_host_string` captured by value here?

src/p2p/net_node.inl
```diff
@@ -1608,100 +1653,128 @@ namespace nodetool
...
+              }
+              // else 'take' stays true, we will take an IPv6 address that is not V4 mapped
+            }
+            if (take)
+              subnet_peers.push_back(peer);
           }
 
-          // consider each host once, to avoid giving undue inflence to hosts running several nodes
-          if (!skip)
+          // Step 3: Put back into order according to 'last_seen', i.e. most recently seen first
```
jeffro256, line 1722: I'm not sure that this is a good idea, because this may create a positive feedback loop. A recently seen candidate is more likely to be picked next, which means it keeps an earlier "last seen" value, which means it's more likely to be picked next, so on and so forth.

jeffro256: This might further entrench [super nodes](https://arxiv.org/abs/2504.15986).

src/p2p/net_node.inl
```diff
@@ -1571,33 +1571,78 @@ namespace nodetool
...
     {
       ++rand_count;
-      size_t random_index;
+
       const uint32_t next_needed_pruning_stripe = m_payload_handler.get_next_needed_pruning_stripe().second;
 
-      // build a set of all the /16 we're connected to, and prefer a peer that's not in that set
-      std::set<uint32_t> classB;
-      if (&zone == &m_network_zones.at(epee::net_utils::zone::public_)) // at returns reference, not copy
+      // Build a list of all distinct /24 subnets we are connected to now right now; to catch
```
jeffro256, line 1631: Why narrow the subnets of connections we avoid duplicating from /16 to /24? This would allow the code to make connections to IPs *more* similar to the set of connections that we already have, not less.

src/p2p/net_node.inl
```diff
@@ -1571,33 +1571,78 @@ namespace nodetool
...
+    zone.m_peerlist.foreach(use_white_list, [&peers, &hosts, &total_peers_size, get_host_string](const peerlist_entry &peer)
+    {
+      ++total_peers_size;
+      const std::string host_string = get_host_string(peer.adr);
+      const auto it = hosts.find(host_string);
+      if (it == hosts.end())
+      {
+        peers.push_back(peer);
+        hosts.insert(host_string);
+      }
```
jeffro256: I know that the original code did this too, but completely disregarding other hosts with the same port based on which is first in the list might be detrimental to private testnets. Obviously, two nodes on the same IP should be weighted as one node in the selection process, but the probability to pick either port should be the same IMO.

src/p2p/net_node.inl
```diff
@@ -1608,100 +1653,128 @@ namespace nodetool
...
+            bool take = true;
+            if (peer.adr.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id())
             {
-              boost::asio::ip::address_v4 v4ip = make_address_v4_from_v6(actual_ip);
-              uint32_t actual_ipv4;
-              memcpy(&actual_ipv4, v4ip.to_bytes().data(), sizeof(actual_ipv4));
-              skip = classB.find(actual_ipv4 & ntohl(0xffff0000)) != classB.end();
+              const epee::net_utils::network_address na = peer.adr;
+              uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
+              uint32_t subnet = actual_ip & 0x00ffffff;
```
jeffro256:
```suggestion
              const uint32_t subnet = actual_ip & 0x00ffffff;
```

src/p2p/net_node.inl
```diff
@@ -1608,100 +1653,128 @@ namespace nodetool
...
+            const peerlist_entry &peer = peers[index];
+            bool take = true;
+            if (peer.adr.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id())
             {
-              boost::asio::ip::address_v4 v4ip = make_address_v4_from_v6(actual_ip);
-              uint32_t actual_ipv4;
-              memcpy(&actual_ipv4, v4ip.to_bytes().data(), sizeof(actual_ipv4));
-              skip = classB.find(actual_ipv4 & ntohl(0xffff0000)) != classB.end();
+              const epee::net_utils::network_address na = peer.adr;
+              uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
```
jeffro256:
```suggestion
              const uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
```

src/p2p/net_node.inl
rbrunner7: The background here is @Rucknium 's paper, to be found [here](https://github.com/Rucknium/misc-research/blob/main/Monero-Peer-Subnet-Deduplication/pdf/monero-peer-subnet-deduplication.pdf), and there especially figure 2 on page 5.

Rucknium and I discussed /16 versus /24 back and forth over weeks. If I read the backlog correctly, I asked whether /24 instead of the proposed /16 in the paper would be alright as well. Rucknium agreed to an attempt to go for that. They also did a large number of test runs with statistics to see whether my code actually worked, and to what degree, and /24 seemed to work out.

When proposing "going finer", I imagined that with /24 you exclude less honest nodes that just have the bad luck to be in the same subnet as a spy node. With /16 this danger seems much bigger to me. And well, as of now the spy nodes do us the favor to nicely mostly crowd into /24 subnets that they fill, using all 256 IPs there.

> This would allow the code to make connections to IPs more similar to the set of connections that we already have, not less.

Not sure how to weight that. In any case, the whole peer selection process is essentially random on several levels. I don't think there is a real danger that we accidentally somehow seek out /24 subnets that are immediate neighbours of spy node infested subnets and thus probably infested themselves. I would also speculate that if this was real danger Rucknium's test runs should have unearthed that.

src/p2p/net_node.inl
rbrunner7: Do I understand your proposal correctly? Not take always the first node from an IP with several nodes that we encounter, but track / make lists of servers per IP and then choose one *randomly* from all of them?

Seems quite some work to track this ...

In general, mostly I only changed behavior that Rucknium did expressively mention as being in need of change and adjustment. This is also the old principle in action: If it's not broken, don't fix it. I think that with this behavior being a problem, and being active for some years already, we should see problem reports. I don't know of any however.

Frankly, I also avoided further changes to avoid that the burden to prove that it's an improvement would land on me :)

src/p2p/net_node.inl
rbrunner7: Looks like an oversight. In the original code it was captured by reference. I will go back to that.

src/p2p/net_node.inl
rbrunner7: That's how the old code is doing it as well. I just applied the same principle that I explained at your question about ignoring additional nodes on the same peer: Do not touch something I don't have good reason for, and if that's a real problem, shouldn't there be reports about it?

I share your doubts, and it's easy to watch how the same nodes are chosen over and over again, but on the other hand everything is terribly dynamic, and implementing something different **and** then proving convincingly that it's better is anything but easy.

src/p2p/net_node.inl
jeffro256: I think we're misunderstanding each other here. I mean that the old code was already de-duplicating subnets for *which ones which it was already connected to*, not the new type of de-duplication where we treat many IPs under a new /24 subnet as having one total vote. Did Rucknium discuss this type of de-duplication at all?

rbrunner7: I don't think we have a misunderstanding. Yes, it's about this very first phase of making a list of subnets we are already connected to. Originally built as a list of /16 subnets, now narrowed down to a list of /24 subnets.

And yes, understood as such by Rucknium and me, and as I said discussed back and forth whether to make a difference in subnet size between this initial "connected to" list evaluation and the later candidate filtering, or treat everything /24, and the latter is where we landed and which Rucknium gave their ok to, after some simulations and/or test runs with their actual daemon, as far as I remember.

Rucknium: I ran some simulations that produced realistic network connections when nodes adopt different subnet connection schemes.

The simulation works like this:

1) A set of IP addresses of reachable node is collected, based on a Monero network scan from May 2025. Only reachable nodes will be used in the simulation, but the network simulation function has to option to set up unreachable nodes if desired. Since the IP addresses of unreachable nodes are unknown, they cannot be used for the network latency calculations later.
2) Each node follows a procedure to establish "connections" to other nodes:
3) The node considers the whole list of IP addresses. Any node it is already connected to is excluded. Any node that is in a subnet of a node that it is already connected to is also excluded (this subnet level is a configurable parameter).
4) The remaining candidate list is de-duplicated based on a specified subnet level, which can be different from the already-connected subnet level.
5) A peer is chosen from this list with uniform random selection.
6) Steps 3 - 5 are performed 12 times to fill up the default outbound connections limit.
7) A peer churning process is simulated. This involves randomly dropping a peer from the output peer list and re-running steps 3 - 5 to fill in the dropped peer slot. 100 peers are dropped and their slots filled. This churn step is needed because of how the "already-connected" criteria statistically depends on the number of peer slots filled.
8) Nodes that are on the MRL ban list accept inbound connections, but do not establish their own outbound connections. In the simulation, they are skipped when it is their turn to create connections.

Note that any "limits" on the number of inbound peer connections to nodes are ignored.

I performed simulations of four scenarios using the [`gen.network()` function](https://rucknium.github.io/xmrpeers/reference/gen.network.html) of version 0.2.4 of my [`xmrpeers` R package](https://github.com/Rucknium/xmrpeers). The combinations of "already-connected" subnet level being set to `/16` and `/24` and subnet deduplication being enabled are the `/24` level or not being enabled at all.

- The scenario of the status quo is "already-connected" subnet level being set to `/16` and subnet deduplication not being enabled.
- This scenario of this PR is "already-connected" subnet level being set to `/24` and subnet deduplication being enabled.

The simulation creates a network/graph edgelist. The edgelist can be used to compute the number of outbound connections of honest nodes to suspected spy nodes. It can also be used to roughly measure network latency. Below are some tables.

### Average number of connections to spy nodes

| already.connected.subnet.level|do.deduplication | mean.malicious.connections|
|------------------------------:|:----------------|--------------------------:|
|                             16|TRUE             |                   1.047757|
|                             24|TRUE             |                   1.064399|
|                             16|FALSE            |                   3.582851|
|                             24|FALSE            |                   3.626628|


When honest nodes have the default 12 outbound connections and subnet deduplication is enabled, setting the "already-connected" subnet qualification to the `/16` subnet level gets you an average of 1.048 connections to the spy nodes. When it is `/24`, the average is 1.064.


@jeffro256 , you [asked](https://github.com/monero-project/meta/issues/1254) :

> But there is also geographical distribution concerns also, not related to spy nodes. How does going from /16 to /24 affect that? A less diverse set of already-connected peers might slow down tx&block propagation for those farthest away from the "center" of the node graph

According to [Wikipedia](https://en.wikipedia.org/wiki/Distance_(graph_theory)).

> The eccentricity $\epsilon(v)$ of a vertex $v$ is the greatest distance between $v$ and any other vertex....It can be thought of as how far a node is from the node most distant from it in the graph.

The eccentricity of every node can be measured. For eccentricity computations, the spy nodes were removed from the network because they do not establish outbound connections.

The edges of the network graph were weighted by the geographic distance between the IP addresses. The units of the weight of each edge is the number of milliseconds it would take light to travel on the Earth's surface in the shortest path between two points. Actual internet latency speeds will be slower due to internet cables not following the shortest path between each IP address's geographic location and computations at internet exchange points.

### Mean eccentricity, in milliseconds

| already.connected.subnet.level|do.deduplication | mean.weighted.eccentricity|
|------------------------------:|:----------------|--------------------------:|
|                             16|TRUE             |                   69.41301|
|                             24|TRUE             |                   68.58541|
|                             16|FALSE            |                   70.05076|
|                             24|FALSE            |                   70.27645|


When subnet deduplication is enabled and the "already-connected" subnet qualification is `/16`, the time that a photon could travel from an _average_  node to the "most distant" node in the network is 69.41 milliseconds. The next table computes the maximum, i.e. the maximum distance between _any_ two nodes on the network.

### Maximum eccentricity, in milliseconds

| already.connected.subnet.level|do.deduplication | max.weighted.eccentricity|
|------------------------------:|:----------------|-------------------------:|
|                             16|TRUE             |                  106.5484|
|                             24|TRUE             |                  108.3377|
|                             16|FALSE            |                  110.4373|
|                             24|FALSE            |                  110.8845|


We also get unweighted eccentricity, which just counts each hop as one unit. The unweighted eccentricity of every honest node in the network in every scenario was 4, i.e. the minimum number of hops between a specific node and and any other node was always 4 regardless of which node we start at. In a few of the simulation runs iwth a different random seed, a few of the nodes had an unweighted eccentricity of 5 when `do.deduplication` was `FALSE`. But in this random seed, every is 4:

### Table of unweighted eccentricity

| already.connected.subnet.level|do.deduplication |table.unweighted.eccentricity |
|------------------------------:|:----------------|:-----------------------------|
|                             16|TRUE             |4: 2764                       |
|                             24|TRUE             |4: 2764                       |
|                             16|FALSE            |4: 2764                       |
|                             24|FALSE            |4: 2764                       |


Code to reproduce the analysis is below. Click to expand. Note that this code makes network calls to a geo-IP service.


<details>

<summary>Reproduction code</summary>

```R
# WARNING: This script makes network calls to ip-api.com

install.packages(c("remotes", "geosphere", "knitr"))
remotes::install_github("Rucknium/xmrpeers", upgrade = FALSE,
  ref = "35408988f80a306627f5c10abb10ffb89f008bf4")
# ref is to get specific version so this is reproducible

library(data.table)

data("good_peers", "ban_list", package = "xmrpeers")

good_peers.cleaned <- good_peers
good_peers.cleaned <- stringr::str_extract(good_peers.cleaned, "[0-9]{1,3}[.][0-9]{1,3}[.][0-9]{1,3}[.][0-9]{1,3}")
# Remove ports from IP addresses
good_peers.cleaned <- good_peers.cleaned[good_peers.cleaned != ""]
# Removes IPv6 addresses
good_peers.cleaned <- unique(na.omit(good_peers.cleaned))

# https://ip-api.com/docs/api:batch
# Docs say:
# This endpoint is limited to 15 requests per minute from an IP address.
# A request needs to formatted as a JSON array, containing up to 100 IP addresses or objects.
# Only POST requests are accepted.
# The returned HTTP header X-Rl contains the number of requests remaining in the
# current rate limit window. X-Ttl contains the seconds until the limit is reset.
# Your implementation should always check the value of the X-Rl header, and if
# its is 0 you must not send any more requests for the duration of X-Ttl in seconds.

good_peers.cleaned.split <- split(good_peers.cleaned, 
  f = rep(seq_len(ceiling(length(good_peers.cleaned) / 100)), each = 100)[seq_along(good_peers.cleaned)])
# Split into chunks of 100, with remainder

handle <- RCurl::getCurlHandle()

h = RCurl::basicHeaderGatherer()
w = RCurl::basicTextGatherer()

ip.geo <- list()

for (i in seq_along(good_peers.cleaned.split)) {
  
  json.post <- RJSONIO::toJSON(
    good_peers.cleaned.split[[i]]
  )
  
  # returns NULL
  RCurl::postForm("http://ip-api.com/batch",
    .opts = RCurl::curlOptions(
      postfields = json.post,
      httpheader = c('Content-Type' = 'application/json', Accept = 'application/json'),
      headerfunction = h$update,
      writefunc = w$update
      # Updates the value of "h" and "w" as a side effect.
      # (header = TRUE would print the header to the console)
    ),
    curl = handle
  )
  # "postForm does not return response with headerfunction"
  # https://github.com/omegahat/RCurl/issues/5
  # "You should also pass the writefunc (e.g. basicTextGatherer) along with the headerfunction."
  
  ip.geo[[i]] <- data.table::rbindlist(RJSONIO::fromJSON(w$value()), fill = TRUE,
    ignore.attr = TRUE)
  
  w$reset()
  # Reset to make room for the new data
  # "This can be used to reuse the same object across requests but to avoid
  # cumulating new input with the material from previous requests."
  
  throttle.limit <- as.numeric(h$value()[["X-Rl"]])
  
  if (throttle.limit  == 0 ) {
    Sys.sleep(75)
    # Sleep for 75 seconds if (unexpectedly) we have no more requests available
    # during this minute.
  }
  
  message(i, " of ", length(good_peers.cleaned.split), " API requests completed. ",
    "Throttle limit in current minute: ", throttle.limit)
  
  Sys.sleep(5)
  # Sleep 5 seconds. Gives one second tolerance to the limit of 15 requests per minute.
  
}

ip.geo <- data.table::rbindlist(ip.geo, fill = TRUE)

ip.geo <- ip.geo[, .(ip = query, lat, lon)]

stopifnot( all(complete.cases(ip.geo)))
# Make sure no missing values


future::plan(future::multisession,
  workers = max(c(1, floor(parallelly::availableCores()/5))))
# Each process uses 400-500% CPU, so set to available threads divided by 5

peer.selection.scenarios <- expand.grid(
  already.connected.subnet.level = c(16, 24),
  do.deduplication = c(TRUE, FALSE))

results <- vector("list", nrow(peer.selection.scenarios))

set.seed(314)

for (scenario in seq_len(nrow(peer.selection.scenarios))) {
  
  message("\nStarting simulation for scenario ", scenario, " of ",
    nrow(peer.selection.scenarios))
  
  already.connected.subnet.level <- 
    peer.selection.scenarios$already.connected.subnet.level[scenario]
  do.deduplication <- peer.selection.scenarios$do.deduplication[scenario]
  # Need these to be set as variables due to scoping issues with paralellized sessions
  
  generated.network <- xmrpeers::gen.network(
    outbound.ips = good_peers.cleaned,
    malicious.ips = ban_list,
    n.unreachable = 0,
    already.connected.subnet.level = already.connected.subnet.level,
    deduplication.subnet.level = 24,
    do.deduplication = do.deduplication,
    default.outbound.connections = 12,
    dropped.connection.churns = 100,
    compute.network.stats = FALSE
  )
  
  malicious.indices <- generated.network$nodes[ xmrpeers::in.ip.set(ip, ban_list), index]
  nodes.honest <- generated.network$nodes[ ! index %in% malicious.indices, ]
  edgelist.honest <- generated.network$edgelist[ ! origin %in% malicious.indices, ]
  edgelist.honest.completely <- generated.network$edgelist[ ! destination %in% malicious.indices, ]
  edgelist.honest[, malicious.connection := destination %in% malicious.indices]
  malicious.connection.count <- edgelist.honest[, .(count = sum(malicious.connection)), by = "origin"]
  results[[scenario]]$nrow <- nrow(malicious.connection.count)
  results[[scenario]]$summary <- summary(malicious.connection.count$count)
  
  nodes.honest <- merge(nodes.honest, ip.geo, by = "ip", all.x = TRUE)
  
  p1 <- nodes.honest[match(edgelist.honest.completely$origin, index), .(lon, lat)]
  p2 <- nodes.honest[match(edgelist.honest.completely$destination, index), .(lon, lat)]
  
  edge.weights <- geosphere::distGeo(as.matrix(p1), as.matrix(p2))
  
  edgelist.honest.completely[, weight := edge.weights]
  # igraph::eccentricity() will automatically use weights
  
  edgelist.graph <- igraph::graph_from_data_frame(edgelist.honest.completely, directed = FALSE)
  # NOTE: graph_from_data_frame() is correct. graph_from_edgelist() gives incorrect
  # results, probably from gaps in the index numbers.
  
  weighted.eccentricity <- igraph::eccentricity(edgelist.graph)
  speed.of.light <- 299792458 # meters per second
  
  results[[scenario]]$weighted.eccentricity.summary <-
    summary( 1000 * weighted.eccentricity / speed.of.light)
  # Multiply by 1000 to get milliseconds
  
  unweighted.eccentricity <- igraph::eccentricity(edgelist.graph, weights = NA)
  
  results[[scenario]]$unweighted.eccentricity.table <-
    table(unweighted.eccentricity)
  
}

scenarios.results <- data.table::as.data.table(peer.selection.scenarios)

scenarios.results[, mean.malicious.connections := sapply(results, function(x) x$summary[["Mean"]]) ]
scenarios.results[, mean.weighted.eccentricity := 
    sapply(results, function(x) x$weighted.eccentricity.summary[["Mean"]]) ]
scenarios.results[, max.weighted.eccentricity := 
    sapply(results, function(x) x$weighted.eccentricity.summary[["Max."]]) ]
scenarios.results[, table.unweighted.eccentricity := 
    sapply(results, function(x) {
      y <- x$unweighted.eccentricity.table
      paste0(names(y), ": ", unclass(y), collapse = "; ")
    }) ]


knitr::kable(scenarios.results[, 
  .(already.connected.subnet.level, do.deduplication, mean.malicious.connections)],
  format = "pipe")

knitr::kable(scenarios.results[, 
  .(already.connected.subnet.level, do.deduplication, mean.weighted.eccentricity)],
  format = "pipe")

knitr::kable(scenarios.results[, 
  .(already.connected.subnet.level, do.deduplication, max.weighted.eccentricity)],
  format = "pipe")

knitr::kable(scenarios.results[, 
  .(already.connected.subnet.level, do.deduplication, table.unweighted.eccentricity)],
  format = "pipe")


```

</details>

src/p2p/net_node.inl
Rucknium: I think the status quo behavior is probably acceptable. The `white_list` selection is one extreme: select only from the recently-seen peers. The `gray_list` selection is the other extreme: select from never-seen peers with uniform probability. I am not sure exactly how the decision is made to select from the `white_list` and `gray_list`, but I think tat the `white_list` is used 70% of the time and `gray_list` 30%, by default.

Selecting from recently-seen peers helps nodes stay connected to the same peer when there is a momentary interruption in connections.e.g. when a node restarts after a configuration change. Or at least I think that should be the effect.

And when I have seen the term "super node", usually it refers to nodes that make their own outbound connection to every reachable pee in the network instead of the result of the behavior or normal nodes making their outbound connections, which is what this peer selection algorithm affects.

jeffro256:
This PR has a bug where the outgoing connection count isn't filled up like it should, instead it stops early. Simply run this PR and type `status` to see that the outgoing connection count isn't correct. Attempting to debug now...

jeffro256:
Anecdotally, this does a great job at reducing the number of connections to spy nodes. My running spy node connection count ratio went from ~22% to ~1.5%. You can run this patch to test it out: [log_outgoing_spies.patch](https://github.com/user-attachments/files/22009896/log_outgoing_spies.patch). Run your `monerod` instance with `--log-level p2p.net:DEBUG` and look for lines containing the string "Running spy node stats:".

src/p2p/net_node.inl
```diff
@@ -1608,100 +1653,128 @@ namespace nodetool
...
+
+          // Step 2: Deduplicate by only taking 1 candidate from each /24 subnet that occurs, the FIRST
+          // candidate seen from each subnet within the now random order
+          std::set<uint32_t> subnets = connected_subnets;
+          for (size_t index : shuffled_indexes)
           {
-            const epee::net_utils::network_address na = pe.adr;
-            const boost::asio::ip::address_v6 &actual_ip = na.as<const epee::net_utils::ipv6_network_address>().ip();
-            if (actual_ip.is_v4_mapped())
+            const peerlist_entry &peer = peers[index];
```
jeffro256:
```suggestion
            const peerlist_entry &peer = peers.at(index);
```

src/p2p/net_node.inl
```diff
@@ -1608,100 +1653,128 @@ namespace nodetool
...
         CRITICAL_REGION_LOCAL(m_used_stripe_peers_mutex);
         if (next_needed_pruning_stripe > 0 && next_needed_pruning_stripe <= (1ul << CRYPTONOTE_PRUNING_LOG_STRIPES) && !m_used_stripe_peers[next_needed_pruning_stripe-1].empty())
         {
           const epee::net_utils::network_address na = m_used_stripe_peers[next_needed_pruning_stripe-1].front();
           m_used_stripe_peers[next_needed_pruning_stripe-1].pop_front();
           for (size_t i = 0; i < filtered.size(); ++i)
           {
-            peerlist_entry pe;
-            if (zone.m_peerlist.get_white_peer_by_index(pe, filtered[i]) && pe.adr == na)
+            const peerlist_entry &peer = filtered[i];
```
jeffro256:
```suggestion
            const peerlist_entry &peer = filtered.at(i);
```

src/p2p/net_node.inl
```diff
@@ -1710,52 +1783,54 @@ namespace nodetool
...
 
-      if(tried_peers.count(random_index))
-        continue;
-
-      tried_peers.insert(random_index);
-      peerlist_entry pe = AUTO_VAL_INIT(pe);
-      bool r = use_white_list ? zone.m_peerlist.get_white_peer_by_index(pe, random_index):zone.m_peerlist.get_gray_peer_by_index(pe, random_index);
-      CHECK_AND_ASSERT_MES(r, false, "Failed to get random peer from peerlist(white:" << use_white_list << ")");
+      // We have our final candidate for this pass of the outer try loop
+      const peerlist_entry &candidate = filtered[random_index];
```
jeffro256:
```suggestion
      const peerlist_entry &candidate = filtered.at(random_index);
```

src/p2p/net_node.inl
```diff
@@ -1571,33 +1571,78 @@ namespace nodetool
...
+    });
+
+    const size_t peers_size = peers.size();
+    MDEBUG("Looking at " << peers_size << " port-deduplicated peers out of " << total_peers_size
+      << ", i.e. dropping " << (total_peers_size - peers_size));
+
+    std::set<uint64_t> tried_peers;  // all peers ever tried
+
+    // Outer try loop, with up to 3 attempts to actually connect to a suitable randomly choosen candidate
     size_t rand_count = 0;
```
jeffro256: `rand_count` should probably be renamed at this point, it's a misnomer

src/p2p/net_node.inl
```diff
@@ -1608,100 +1653,128 @@ namespace nodetool
...
-            const epee::net_utils::network_address na = pe.adr;
-            uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
-            skip = classB.find(actual_ip & 0x0000ffff) != classB.end();
-          }
-          else if (skip_duplicate_class_B && pe.adr.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id())
+        if ((step == 1) && !is_public_zone)
+          break;
+
+        candidate_peers = &peers;
+        if ((step == 0) && is_public_zone)
```
jeffro256:
```suggestion
        const bool try_subnet_dedup = step == 0 && is_public_zone;
        const std::vector<peerlist_entry> &candidate_peers = try_subnet_dedup ? subnet_peers : peers;
        if (try_subnet_dedup)
```

Also, remove where candidate_peers is declared above and modified below. IMO, this brings down the reference creation closer to where it is used and is a bit clearer to the reader.

src/p2p/net_node.inl
```diff
@@ -1608,100 +1653,128 @@ namespace nodetool
...
+          MDEBUG("Looking at " << subnet_peers_size << " subnet-deduplicated peers out of " << peers_size
+            << ", i.e. dropping " << (peers_size - subnet_peers_size));
+
+          candidate_peers = &subnet_peers;
+        } // deduplicate
+        // else, for step 1 / second pass of inner try loop, take all peers from all subnets
+
+        // Take as many candidates as we need and care about stripes if pruning
+        const size_t limit = use_white_list ? 20 : std::numeric_limits<size_t>::max();
+        for (auto &peer : *candidate_peers) {
```
jeffro256:
```suggestion
        for (const peerlist_entry &peer : candidate_peers) {
```

src/p2p/net_node.inl
```diff
@@ -1608,100 +1653,128 @@ namespace nodetool
...
+          if (filtered.size() >= limit)
+            break;
+
```
jeffro256, line 1742: I would skip over already-tried peers here too, so that we aren't often hitting the `if (tried_peers.count(candidate.id)) continue` branch later, especially if our total peer list is relatively small.

src/p2p/net_node.inl
rbrunner7: Coming from C#, I wouldn't have thought that there is a difference between the array operator and the `at` method ...

src/p2p/net_node.inl
rbrunner7: Agreed, it does not make sense to cling to the old name just to minimize the changes, as it's not random anymore. Renamed it to `outer_loop_count`.

src/p2p/net_node.inl
rbrunner7: Did you notice that I declared `candidate_peers` as a *pointer* to a vector, in an attempt to avoid that vectors get copied? I am not sure, is your alternative solution equivalent, i.e. no vector gets copied when `candidate_peers` gets established?

And I am a bit unsure about the "const-ness" there, because `subnet_peers` will get modified later - will `candidate_peers`, as a const reference to that, just "reflect" that? Does `const` on `candidate_peers` just make sure that the variable itself won't change, but doesn't tell something about the vector elements?

src/p2p/net_node.inl
rbrunner7: You mean inside this `for` loop ad a check whether the peer is already in `tried_peers`?

Hmm, I think it's not immediately clear that we can really do that without changing anything about the logic, and I worry a bit that somebody may make a change where it's not ok anymore to filter for `tried_peers` this early in the code, but does not notice.

src/p2p/net_node.inl
jeffro256: Yup, `.at()` is bounds-checked, `operator[]` isn't.

src/p2p/net_node.inl
jeffro256: Yes, I know that it was a pointer. In my suggestion, `candidate_peers` is a reference, which means it also avoids a copy. It doesn't matter if `subnet_peers` gets modified later, it's still a reference to the same place. A `const` reference just means that you can't modify the object through that reference, not that the object is `const`. The vector's length, elements, etc can change, but `candidate_peers` still refers to the same object, as long as that object has a lifetime greater than the reference.

src/p2p/net_node.inl
jeffro256: I think the logic *should* be changed: where you would add an already-tried peer here, I'm saying that it should skip that.

rbrunner7: I was thinking about possible cases where it can make a difference, and a negative one at that, if we newly start to throw out candidates this early in the processing. But the more I am thinking about it, the more I think that this is not possible, without a need to think everything through in detail. After all, we don't *remove* any peer from `tried_peers` once it's in there.

src/p2p/net_node.inl
rbrunner7: Thanks for the clarification, I see now. I will change the code as proposed.

src/p2p/net_node.inl
jeffro256: Yes, exactly, not checking it here, potentially selecting it in `filtered`, only to continue the outer loop later just burns an outer loop try AFAICT.

src/p2p/net_node.inl
```diff
@@ -1571,33 +1571,78 @@ namespace nodetool
...
       {
         zone.m_net_server.get_config_object().foreach_connection([&](const p2p_connection_context& cntxt)
         {
           if (cntxt.m_remote_address.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id())
           {
-
             const epee::net_utils::network_address na = cntxt.m_remote_address;
             const uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
-            classB.insert(actual_ip & 0x0000ffff);
+            connected_subnets.insert(actual_ip & 0x00ffffff);
```
jeffro256:
```suggestion
            connected_subnets.insert(actual_ip & ntohl(0xffffff00));
```

To match line 1656

src/p2p/net_node.inl
```diff
@@ -1608,100 +1653,129 @@ namespace nodetool
...
+            bool take = true;
+            if (peer.adr.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id())
             {
-              boost::asio::ip::address_v4 v4ip = make_address_v4_from_v6(actual_ip);
-              uint32_t actual_ipv4;
-              memcpy(&actual_ipv4, v4ip.to_bytes().data(), sizeof(actual_ipv4));
-              skip = classB.find(actual_ipv4 & ntohl(0xffff0000)) != classB.end();
+              const epee::net_utils::network_address na = peer.adr;
+              const uint32_t actual_ip = na.as<const epee::net_utils::ipv4_network_address>().ip();
+              const uint32_t subnet = actual_ip & 0x00ffffff;
```
jeffro256: Same here. I would consider assigning the constant `ntohl(0xffffff00)` to a named variable for code cleanliness.

src/p2p/net_node.inl
```diff
@@ -1571,33 +1571,79 @@ namespace nodetool
...
+    std::vector<peerlist_entry> peers;
+    std::unordered_set<std::string> hosts;
+    size_t total_peers_size = 0;
+    zone.m_peerlist.foreach(use_white_list, [&peers, &hosts, &total_peers_size, &get_host_string](const peerlist_entry &peer)
     {
-      ++rand_count;
-      size_t random_index;
+      ++total_peers_size;
+      const std::string host_string = get_host_string(peer.adr);
+      const auto it = hosts.find(host_string);
```
vtnerd: This can be done in one call:
```c++
if (hosts.insert(host_string).second)
  peers.push_back(peer);
```

which reduces map lookups by 1.

src/p2p/net_node.inl
```diff
@@ -1608,100 +1654,129 @@ namespace nodetool
...
+        const bool try_subnet_dedup = step == 0 && is_public_zone;
+        const std::vector<peerlist_entry> &candidate_peers = try_subnet_dedup ? subnet_peers : peers;
+        if (try_subnet_dedup)
+        {
+          // Deduplicate subnets using 3 steps
+
+          // Step 1: Prepare to access the peers in a random order
+          std::vector<size_t> shuffled_indexes(peers.size());
+          std::iota(shuffled_indexes.begin(), shuffled_indexes.end(), 0);
+          std::shuffle(shuffled_indexes.begin(), shuffled_indexes.end(), crypto::random_device{});
```
vtnerd, line 1683: Looks like you could just `std::shuffle` the `peers` object then iterate over it instead ... ? Moving a `std::size_t` is definitely faster, so perhaps thats why it was done this way?

vtnerd (approved):
Just a few minor comments.

src/p2p/net_node.inl
rbrunner7: Yes. I noticed that my node deals with over 1000 gray nodes. I thought it wasteful to let it sort that many `peerlist_entry` objects after I saw that doing it with indexes is quite simple.

rbrunner7:
Thanks for the review, @vtnerd ! Updated. Will go on to create a release counterpart for this PR.

**EDIT**: https://github.com/monero-project/monero/pull/10113

src/p2p/net_node.inl
vtnerd: You might spend more time allocating the memory for the extra `vector` than you saved sorting integers. But whatever, not a big deal.

selsta (approved):

PR #9940, opened 2025-05-30
wallet: kill support for deprecated ephemeral Boost messages

jeffro256:
The commit kills support for deprecated ephemeral Boost messages: signed/unsigned transaction sets, pending transactions, reserve proofs, MMS messages, etc. It does NOT kill support for loading very old wallets in Boost format, that should be supported indefinitely. These messages were deprecated 5 years ago. Since then, we have had a hard fork to enable a new non-compatible transaction type (w/ view tags), and disable the old transaction type. This renders basically all of the aforementioned messages before that HF useless, with the possible exception of reserve proofs.

This commit also cleans up dead inclusions of boost serialization headers.

This commit is part of upstreaming Carrot/FCMP++. Killing support for Boost messages now means less boilerplate Boost serialization review for Carrot/FCMP++.


tobtoht:
Concept ACK.

j-berman (approved):

tobtoht (approved):

PR #9944, opened 2025-06-07
simplewallet: cli using wrong filename for storing keys

### LINKED ISSUE #9943: Monero Wallet Cli password command changes password incorrectly
Using monero version 0.18.4.0-2 I changed my wallet password from one password to another, noticing that the password was set incorrectly. I created a new test wallet file and noticed the same. If you change the password to another password, sometimes it works and sometimes it doesn't. The mistakes occur more often if you enter the same password as the previous one. Next time you try to log in you it tells you incorrect password. Not sure how to debug it.

rudolfschmidt:
m_wallet_file contains the file full name, but the change_password function requires only the file name.

closes #9943

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -1004,7 +1004,7 @@ bool simple_wallet::change_password(const std::vector<std::string> &args)
 
   try
   {
-    m_wallet->change_password(m_wallet_file, orig_pwd_container->password(), pwd_container->password());
+    m_wallet->change_password(m_wallet->get_wallet_file(), orig_pwd_container->password(), pwd_container->password());
   }
   catch (const tools::error::wallet_logic_error& e)
   {


plowsof:
fixes #9943

plowsof:
other instances of change_password with the odd one out not using `get_wallet_file()` is being fixed in this PR
https://github.com/monero-project/monero/blob/125622d5bdc42cf552be5c25009bd9ab52c0a7ca/src/simplewallet/simplewallet.cpp#L1007
https://github.com/monero-project/monero/blob/125622d5bdc42cf552be5c25009bd9ab52c0a7ca/src/wallet/api/wallet.cpp#L847
https://github.com/monero-project/monero/blob/125622d5bdc42cf552be5c25009bd9ab52c0a7ca/src/wallet/wallet_rpc_server.cpp#L3749

this would suggest that wallet rpc is not effected by this bug. ill confirm :heavy_check_mark:

nahuhh (approved):
tested w/o issue. (resolves 9943)

PR #9945, opened 2025-06-07
fix issue 9943: cli using wrong filename for storing keys

rudolfschmidt:
m_wallet_file contains the file full name, but the change_password function requires only the file name.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -1017,7 +1017,7 @@ bool simple_wallet::change_password(const std::vector<std::string> &args)
 
   try
   {
-    m_wallet->change_password(m_wallet_file, orig_pwd_container->password(), pwd_container->password());
+    m_wallet->change_password(m_wallet->get_wallet_file(), orig_pwd_container->password(), pwd_container->password());
   }
   catch (const tools::error::wallet_logic_error& e)
   {


plowsof:
Do we know what new password is applied if anyone was effected by this (and didnt write their seed down?)

nahuhh:
> Do we know what new password is applied if anyone was effected by this

looks like the key file has password ~~changed~~ corrupted, but not the cache file. It may be possible to recover wallet history/labels etc.

1. break wallet
2. Restore from seed, use old password
3. Replace new cache file with old cache file

> (and didnt write their seed down?)
rip

rudolfschmidt:
> 1. break wallet

Can you elaborate on the steps? What do you mean by breaking the wallet?

nahuhh:
> Can you elaborate on the steps? What do you mean by breaking the wallet?

I mean:
1. reproduce this issue (resulting in a broken wallet state). Youll have wallet.keys + wallet
2. Restore from seed (use old password) = restorewallet.keys + restorewallet
3. `cp wallet restorewallet`

in theory, this should work to recover history

rudolfschmidt:
> I mean:
> 
> 1. reproduce this issue (resulting in a broken wallet state). Youll have wallet.keys + wallet
> 2. Restore from seed (use old password) = restorewallet.keys + restorewallet
> 3. `cp wallet restorewallet`
> 
> in theory, this should work to recover history

Seems to work.

nahuhh:
#9944

PR #9948, opened 2025-06-07
Add is_same_host check to p2p

vtnerd:


0xFFFC0000 (approved):

src/p2p/net_node.inl
```diff
@@ -1327,7 +1327,7 @@ namespace nodetool
     bool used = false;
     server->second.m_net_server.get_config_object().foreach_connection([&, is_public](const p2p_connection_context& cntxt)
     {
-      if((is_public && cntxt.peer_id == peer.id) || (!cntxt.m_is_income && peer.adr == cntxt.m_remote_address))
+      if((is_public && cntxt.peer_id == peer.id && peer.adr.is_same_host(cntxt.m_remote_address)) || (!cntxt.m_is_income && peer.adr == cntxt.m_remote_address))
```
SChernykh, line 1330: What's the difference between `peer.adr.is_same_host(cntxt.m_remote_address)` and `peer.adr == cntxt.m_remote_address`? You now have both variants in the same conditional expression.

vtnerd: `is_same_host` checks only the host portion - the port is skipped. The `operator==` does a host+port check.

We want a host only check for this new portion because it needs to match on incoming connections where the port is typically randomly chosen. This will block connections if two `monerod` are using the same ip with the same randomly chosen 64bit id value.

The other `operator==` check is for anonymity networks where only outbound connections have an identifier. This allows for 2 connections (one in each direction) when each side is running standard code. I don't think this can be improved as the inbound identity is intentionally random over these networks.

SChernykh (approved):

jeffro256:
Should the function `node_server<t_payload_net_handler>::is_peer_used(const anchor_peerlist_entry& peer)` also be modified for good measure?

jeffro256 (approved):

PR #2605, opened 2017-10-08
Wallet: Descriptions through new commands 'set_description', 'get_description'

rbrunner7:
This implements the new feature of descriptions for wallets as a whole. A description for a wallet can be set by the new command `set_description` in the CLI wallet. It can be queried by the likewise new command `get_description` .

The recently introduced CLI wallet command `wallet_info` now displays the description as well.

Please be aware that with this code active the structure version for wallets goes to 21, after it already went from 18 (version of latest release code) to 19 by some of moneromooo's recent changes and then to 20 with the subaddresses code. This means that Helium Hydra release code cannot process wallets "touched" by this code anymore. (There seems to be no easy way with Boost binary serialization to make such changes downward-compatible.)

To store the description a new mechanism called *attributes* is used. Technically this is a collection of key/value pairs that allows to store any number of string values, identified by strings as keys, together with a wallet. So far this mechanism is internal to *wallet2* but could be made available in the future through direct methods and RPC calls so that wallet programs like the GUI wallet or the CLI wallet could use it to store additional info into a wallet, without any structure or compatibility problems - as long as they can be expressed as a reasonable number of simple string values, of course.

For more technical details please refer to the corresponding comments in *wallet2.h*.

**Edit:** After going everything through with moneromooo I changed the terminology from *wallet note* to *description* (now with the new CLI wallet commands `get_description` and `set_decription)` and from *extended attributes* to simply *attributes*. Our discussion below of course still uses the old terminology.

**Edit:* Version goes to 21.

PR #9953, opened 2025-06-12
wallet: immediate lookahead table expansion + set_subaddresss_lookahead RPC endpoint

### LINKED ISSUE #7364: Changing subaddress-lookahead on existing wallet does not result in correct blockchain rescan
Using latest v0.17.1.9 cli wallet, I had issues that my BTCPay gateway was using larger offsets than the default 200, so I opened the wallet in monero-wallet-cli and used

`set subaddress-lookahead 3:50000`
`rescan_bc`

And this did nothing. The same transactions were found, no missing transactions were actually added.

Afetr restarting wallet few times and rescanning bc few times, I tried to create new wallet from the same seed using

`monero-wallet-cli --restore-deterministic-wallet --subaddress-lookahead 3:50000`

And voila! This did the trick, all the missing transactions were found.

So the issue is that for existing wallet, changing the subaddress-lookahead has no real effect. I confirmed the change is saved properly by typing `set` into wallet-cli, and this did show the value I configured. However, the `rescan_bc` was always rather quick and never found any new transactions.

### LINKED ISSUE #8954: Support subaddress lookahead in monero-wallet-rpc

### LINKED ISSUE #8980: simplewallet: set subaddress-lookahead does not work after wallet creation

nahuhh:
replaces #8981

~~no changes~~
changes from original:
- added header to fix test failure `#include <boost/optional/optional_io.hpp>`
- updates & fixes by @j-berman 
- updates & fixes by @jeffro256 

resolves #8980
resolves #8954
resolves #7364

> Basically the current behavior is that if you make a wallet with the default settings in the cli and then do set subaddress-lookahead 50:300, then the subaddress pubkey table hasnt actually expanded and you aren't looking any further when you scan. Instead, the table will get expanded to this look ahead only after either a new enote is found with the old lookahead or if the user requests a new subaddress with address new.
>
> This patch makes sure that the wallet adjusts the pubkey table immediately after sending the set command if an increase in the lookahead is requested.
>
> This patch also adds a new endpoint to the wallet-rpc program, /set_subaddress_lookahead which allows the user to expand the subaddress lookahead table via remote command. ~~As mentioned in the original ticket, the change does not persist to disk when the wallet closes and changing the wallet-rpc program to save those cached values looks like kinda a pain but I hope that the requestor finds that something is better than nothing for now.~~

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1597,39 +1597,60 @@ bool wallet2::should_expand(const cryptonote::subaddress_index &index) const
 //----------------------------------------------------------------------------------------------------
 void wallet2::expand_subaddresses(const cryptonote::subaddress_index& index)
 {
-  hw::device &hwdev = m_account.get_device();
-  if (m_subaddress_labels.size() <= index.major)
+  // check if index will overflow container (usually only applicable on 32-bit systems)
+  if constexpr (sizeof(std::size_t) <= sizeof(std::uint32_t))
   {
-    // add new accounts
-    cryptonote::subaddress_index index2;
-    const uint32_t major_end = get_subaddress_clamped_sum(index.major, m_subaddress_lookahead_major);
-    for (index2.major = m_subaddress_labels.size(); index2.major < major_end; ++index2.major)
-    {
-      const uint32_t end = get_subaddress_clamped_sum((index2.major == index.major ? index.minor : 0), m_subaddress_lookahead_minor);
-      const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, 0, end);
-      for (index2.minor = 0; index2.minor < end; ++index2.minor)
-      {
-         const crypto::public_key &D = pkeys[index2.minor];
-         m_subaddresses[D] = index2;
-      }
-    }
-    m_subaddress_labels.resize(index.major + 1, {"Untitled account"});
-    m_subaddress_labels[index.major].resize(index.minor + 1);
-    get_account_tags();
+    static constexpr std::uint32_t max_idx = static_cast<std::uint32_t>(std::numeric_limits<std::size_t>::max());
+    const bool cannot_label_index = index.major == max_idx || index.minor == max_idx;
+    THROW_WALLET_EXCEPTION_IF(cannot_label_index, error::wallet_internal_error, "subaddress index out of range");
   }
-  else if (m_subaddress_labels[index.major].size() <= index.minor)
+
+  // resize subaddress labels just big enough that `m_subaddress_labels[index.major][index.minor]` is present
+  if (m_subaddress_labels.size() <= index.major)
+    m_subaddress_labels.resize(index.major + 1, {"Untitled account"});
+  auto &subaddr_labels_in_account = m_subaddress_labels[index.major];
+  if (subaddr_labels_in_account.size() <= index.minor)
+    subaddr_labels_in_account.resize(index.minor + 1);
+  get_account_tags(); //trigger m_account_tags integrity checks
+
+  // compile all indices present in subaddress scanning map, as well as every major index
+  std::unordered_set<cryptonote::subaddress_index> all_indices;
+  std::unordered_map<std::uint32_t, std::uint32_t> lowest_missing_minor;
+  for (const auto &p : m_subaddresses)
   {
-    // add new subaddresses
-    const uint32_t end = get_subaddress_clamped_sum(index.minor, m_subaddress_lookahead_minor);
-    const uint32_t begin = m_subaddress_labels[index.major].size();
-    cryptonote::subaddress_index index2 = {index.major, begin};
-    const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, index2.minor, end);
-    for (; index2.minor < end; ++index2.minor)
+    all_indices.insert(p.second);
+    lowest_missing_minor[p.second.major];
+  }
+
+  // find lowest "missing" minor index in map, for all major indices
+  // this is an optimization which allows us to skip re-generating pubkeys that we already have
+  for (auto &p : lowest_missing_minor)
+  {
+    const std::uint32_t major = p.first;
+    std::uint32_t &minor = p.second;
+    while (all_indices.count({major, minor}) && minor < std::numeric_limits<std::uint32_t>::max())
+      ++minor;
+  }
+
+  // resize subaddress scanning map to highest historical received subaddress index plus lookahead
+  hw::device &hwdev = m_account.get_device();
+  const std::uint32_t major_base = std::max<std::uint32_t>(m_subaddress_labels.size(), 1) - 1;
+  const std::uint32_t major_end = get_subaddress_clamped_sum(major_base, m_subaddress_lookahead_major);
+  for (std::uint32_t major = 0; major < major_end; ++major)
+  {
+    const std::size_t n_minor_labels = (major < m_subaddress_labels.size()) ? m_subaddress_labels.at(major).size() : 0;
+    const std::uint32_t minor_base = std::max<std::uint32_t>(n_minor_labels, 1) - 1;
+    const std::uint32_t minor_end = get_subaddress_clamped_sum(minor_base, m_subaddress_lookahead_minor);
+    const std::uint32_t minor_begin = lowest_missing_minor.count(major) ? lowest_missing_minor.at(major) : 0;
+    if (minor_begin >= minor_end)
+      continue;
+    const std::vector<crypto::public_key> pkeys
+      = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), major, minor_begin, minor_end);
+    for (std::uint32_t minor = minor_begin; minor < minor_end; ++minor)
     {
-       const crypto::public_key &D = pkeys[index2.minor - begin];
-       m_subaddresses[D] = index2;
+      const crypto::public_key &D = pkeys.at(minor - minor_begin);
+      m_subaddresses[D] = {major, minor};
     }
-    m_subaddress_labels[index.major].resize(index.minor + 1);
   }
 }
 //----------------------------------------------------------------------------------------------------
@@ -1957,32 +1978,7 @@ void wallet2::set_subaddress_lookahead(size_t major, size_t minor)
   if (old_major_lookahead >= major && old_minor_lookahead >= minor)
     return;
 
-  // Expand the subaddresses map so that outputs received to the higher lookaheads will be identified in the scan loop
-  hw::device &hwdev = m_account.get_device();
-  cryptonote::subaddress_index index2;
-  const uint32_t max_major_idx = this->get_num_subaddress_accounts() > 0 ? (this->get_num_subaddress_accounts() - 1) : 0;
-  const uint32_t major_end = get_subaddress_clamped_sum(max_major_idx, major);
-  for (index2.major = 0; index2.major < major_end; ++index2.major)
-  {
-    // The existing minor addresses already set for this account
-    const uint32_t n_minor_subaddrs = this->get_num_subaddresses(index2.major);
-
-    // The subaddress lookahead is expected to expand from the max index in expand_subaddresses
-    const uint32_t max_minor_idx = n_minor_subaddrs > 0 ? (n_minor_subaddrs - 1) : 0;
-    const uint32_t begin = (n_minor_subaddrs || index2.major < old_major_lookahead) ? get_subaddress_clamped_sum(max_minor_idx, old_minor_lookahead) : 0;
-    // The expected new n minor subaddresses allocated for this account
-    const uint32_t end = get_subaddress_clamped_sum(max_minor_idx, minor);
-
-    if (begin >= end)
-      continue;
-
-    const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, begin, end);
-    for (index2.minor = begin; index2.minor < end; ++index2.minor)
-    {
-      const crypto::public_key &D = pkeys.at(index2.minor - begin);
-      m_subaddresses[D] = index2;
-    }
-  }
+  expand_subaddresses({0, 0});
 }
 //----------------------------------------------------------------------------------------------------
 /*!
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -1124,6 +1124,13 @@ private:
     size_t get_num_subaddress_accounts() const { return m_subaddress_labels.size(); }
     size_t get_num_subaddresses(uint32_t index_major) const { return index_major < m_subaddress_labels.size() ? m_subaddress_labels[index_major].size() : 0; }
     void add_subaddress(uint32_t index_major, const std::string& label); // throws when index is out of bound
+    /**
+     * brief: expand subaddress labels up to `index`, and scanning map to highest labeled index plus current lookahead
+     * param: index -
+     *
+     * All calls to `expand_subaddresses()` will *always* expand the subaddress map if the lookahead
+     * values have been increased since the last call.
+     */
     void expand_subaddresses(const cryptonote::subaddress_index& index);
     void create_one_off_subaddress(const cryptonote::subaddress_index& index);
     std::string get_subaddress_label(const cryptonote::subaddress_index& index) const;
--- a/tests/functional_tests/transfer.py
+++ b/tests/functional_tests/transfer.py
@@ -86,6 +86,7 @@ class TransferTest():
         self.check_subtract_fee_from_outputs()
         self.check_background_sync()
         self.check_background_sync_reorg_recovery()
+        self.check_subaddress_lookahead()
 
     def reset(self):
         print('Resetting blockchain')
@@ -1530,6 +1531,48 @@ class TransferTest():
         util_resources.remove_wallet_files('test1')
         self.wallet[0].close_wallet()
         self.wallet[0].restore_deterministic_wallet(seed = seeds[0])
+
+    def check_subaddress_lookahead(self):
+        daemon = Daemon()
+
+        print('Testing transfers to subaddresses with large lookahead')
+
+        # From wallet 1 to wallet 0 at subaddress (0, 999)
+
+        address_0_999 = '8BQKgTSSqJjP14AKnZUBwnXWj46MuNmLvHfPTpmry52DbfNjjHVvHUk4mczU8nj8yZ57zBhksTJ8kM5xKeJXw55kCMVqyG7' # this is the address for address 999 of the main account in the test wallet
+        try:  # assert address_1_999 is not in the current pubkey table
+            self.wallet[0].get_address_index(address_0_999)
+            assert False # address should not already be loaded
+        except Exception as e:
+            assert str(e) ==  "{'error': {'code': -2, 'message': \"Address doesn't belong to the wallet\"}, 'id': '0', 'jsonrpc': '2.0'}"
+        # update the lookahead and assert the high index address is now in the table
+        self.wallet[0].set_subaddress_lookahead(50, 1000)
+        res = self.wallet[0].get_address_index(address_0_999)
+        assert res['index']['major'] == 0
+        assert res['index']['minor'] == 999
+
+        dst = {'address': address_0_999, 'amount': 454545454545}
+
+        self.wallet[1].refresh()
+        assert self.wallet[1].get_balance().balance > dst['amount']
+        self.wallet[1].transfer([dst])
+        daemon.generateblocks('46r4nYSevkfBUMhuykdK3gQ98XDqDTYW1hNLaXNvjpsJaSbNtdXh1sKMsdVgqkaihChAzEy29zEDPMR3NHQvGoZCLGwTerK', 1)
+        self.wallet[0].refresh()
+
+        res = self.wallet[0].get_balance()
+        balance_info_0_999 = None
+        for balance_info in res['per_subaddress']:
+            if balance_info['account_index'] == 0 and balance_info['address_index'] == 999:
+                balance_info_0_999 = balance_info
+                break
+        assert balance_info_0_999 is not None, "balance info for address (0, 999) not found"
+        assert balance_info_0_999['address'] == address_0_999
+        assert balance_info_0_999['balance'] == dst['amount']
+        assert balance_info_0_999['unlocked_balance'] == 0
+        assert balance_info_0_999['label'] == ''
+        assert balance_info_0_999['num_unspent_outputs'] == 1
+        assert balance_info_0_999['blocks_to_unlock'] == 9
+        assert balance_info_0_999['time_to_unlock'] == 0
 
 if __name__ == '__main__':
     TransferTest().run_test()


jeffro256:
Why was the original PR closed instead of continuing there ?

nahuhh:
merge conflicts, op unavailable (is what i thought. idc where it continues. Was requested that someone continue it)

plowsof:
a review on if the logic is in the correct place / manner and is not duplicating some other function would be useful. e.g. we have a `should_expand` function: (and `expand_subaddresses`)
https://github.com/monero-project/monero/pull/8513/files#r952031123

so if someone who knows what they're looking at can give feedback on if this needs, or would be better as  a `should_expand` then call to `expand_subaddresses`  that would be much appreciated D:

src/wallet/wallet2.cpp
```diff
@@ -1947,6 +1947,25 @@ void wallet2::set_subaddress_lookahead(size_t major, size_t minor)
...
+    for (uint32_t i = 0; i < m_subaddress_labels.size()+m_subaddress_lookahead_major; i++) {
+      uint32_t minor_idx_start = i < m_subaddress_labels.size() ? m_subaddress_labels[i].size()+m_subaddress_lookahead_minor : m_subaddress_lookahead_minor; // if there are existing minor indices being tracked under this account we need to account for that
+      for (uint32_t j = minor_idx_start; j < minor; j++) {
+        cryptonote::subaddress_index idx = {i,j};
+        create_one_off_subaddress(idx);
+      }
+    }
+  }
   m_subaddress_lookahead_major = major;
   m_subaddress_lookahead_minor = minor;
```
j-berman: This currently doesn't handle overflows (e.g. `m_subaddress_labels.size()+major` can overflow), and is also slower than it needs to be by deriving a single subaddress every iteration (calling `create_one_off_subaddress` -> `get_subaddress_spend_public_key` each iteration will de-compress the public spend key every iteration inside `rct::addKeys`).

This should solve the above issues and in my (biased) view is easier to verify:

```suggestion
  const uint32_t old_major_lookahead = m_subaddress_lookahead_major;
  const uint32_t old_minor_lookahead = m_subaddress_lookahead_minor;

  m_subaddress_lookahead_major = major;
  m_subaddress_lookahead_minor = minor;

  if (old_major_lookahead >= major && old_minor_lookahead >= minor)
    return;

  // Expand the subaddresses map so that outputs received to the higher lookaheads will be identified in the scan loop
  hw::device &hwdev = m_account.get_device();
  cryptonote::subaddress_index index2;
  const uint32_t major_end = get_subaddress_clamped_sum(m_subaddress_labels.size(), major);
  for (index2.major = 0; index2.major < major_end; ++index2.major)
  {
    // The existing n minor subaddresses already set for this account (the lookahead expands past this point)
    const uint32_t n_minor_labels = m_subaddress_labels.size() > index2.major ? m_subaddress_labels[index2.major].size() : 0;

    // The expected existing n minor subaddresses already allocated for this account
    const uint32_t begin = (n_minor_labels || index2.major < old_major_lookahead) ? get_subaddress_clamped_sum(n_minor_labels, old_minor_lookahead) : 0;
    // The expected new n minor subaddresses allocated for this account
    const uint32_t end = get_subaddress_clamped_sum(n_minor_labels, minor);

    if (begin >= end)
      continue;

    const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, begin, end);
    for (index2.minor = begin; index2.minor < end; ++index2.minor)
    {
      const crypto::public_key &D = pkeys.at(index2.minor - begin);
      m_subaddresses[D] = index2;
    }
  }
```

j-berman (commented):
> or would be better as a `should_expand` then call to `expand_subaddresses`

@plowsof I thought this might be the case too, but actually that would create labels for all subaddresses in the lookahead range, which would clutter UI's / isn't necessary. @benevanoff's decision not to do that looks correct to me. My suggestion is minor (but worth doing imo), the PR looks good as is and this was tricky to get right. Thank you @benevanoff for this code. And thank you @nahuhh for following up.

src/wallet/wallet2.cpp
plowsof: tested on cli and working fine:

starting with a `1:10` lookahead. sending to above expanded , then decreasing idx until i find the max idx

| lookahead | max idx |
|---------------|-------------|
| 1:10          | 9|           
| | 18|                     
| | 27| 
| 1:20 | 47|
|1:1000 | 1047 |

**update** i was using old code, set lookahead expands by the expected value, thanks for debugging @j-berman 
- ~~note how it starts at idx 9 as its indexed from 0, but, reference wallet cant see subaddress idx 0, but this is expected.~~
- ~~but increments thereafter, only by (lookahead-1) which is odd. i expected a full +lookahead~~  
- expansion is immediate. :heavy_check_mark:

src/wallet/wallet2.cpp
```diff
@@ -1947,8 +1947,40 @@ void wallet2::set_subaddress_lookahead(size_t major, size_t minor)
...
+  const uint32_t major_end = get_subaddress_clamped_sum(m_subaddress_labels.size(), major);
+  for (index2.major = 0; index2.major < major_end; ++index2.major)
+  {
+    // The existing n minor subaddresses already set for this account (the lookahead expands past this point)
+    const uint32_t n_minor_labels = m_subaddress_labels.size() > index2.major ? m_subaddress_labels[index2.major].size() : 0;
+
+    // The expected existing n minor subaddresses already allocated for this account
+    const uint32_t begin = (n_minor_labels || index2.major < old_major_lookahead) ? get_subaddress_clamped_sum(n_minor_labels, old_minor_lookahead) : 0;
+    // The expected new n minor subaddresses allocated for this account
+    const uint32_t end = get_subaddress_clamped_sum(n_minor_labels, minor);
```
j-berman:
```suggestion
    // The existing minor addresses already set for this account
    const uint32_t n_minor_subaddrs = this->get_num_subaddresses(index2.major);

    // The subaddress lookahead is expected to expand from the max index in expand_subaddresses
    const uint32_t n_max_idx = n_minor_subaddrs > 0 ? (n_minor_subaddrs - 1) : 0;
    const uint32_t begin = (n_minor_subaddrs || index2.major < old_major_lookahead) ? get_subaddress_clamped_sum(n_max_idx, old_minor_lookahead) : 0;
    // The expected new n minor subaddresses allocated for this account
    const uint32_t end = get_subaddress_clamped_sum(n_max_idx, minor);
```

tests/unit_tests/subaddress.cpp
```diff
@@ -101,3 +102,29 @@ TEST_F(WalletSubaddress, OutOfBoundsIndexes)
...
+  subaddr = w1.get_subaddress(test_idx);
+  EXPECT_NE(boost::none, w1.get_subaddress_index(subaddr));
+  // next test expanding the minor lookahead
+  w1.set_subaddress_lookahead(100, 300);
+  EXPECT_EQ(100, w1.get_subaddress_lookahead().first);
+  EXPECT_EQ(300, w1.get_subaddress_lookahead().second);
+  test_idx = {100, 299};
+  subaddr = w1.get_subaddress(test_idx);
+  EXPECT_NE(boost::none, w1.get_subaddress_index(subaddr));
+}
```
j-berman, line 194: Some helpful additional tests:

```suggestion
}

TEST_F(WalletSubaddress, ExpandThenSetMinorIncreaseOnly)
{
  // We assume we are starting with the default setup state
  EXPECT_EQ(50, w1.get_subaddress_lookahead().first);
  EXPECT_EQ(200, w1.get_subaddress_lookahead().second);

  // Helper function to check max subaddrs allocated
  const auto check_expected_max = [&](const cryptonote::subaddress_index exp_max)
  {
    for (uint32_t i = 0; i <= exp_max.minor; ++i)
    {
      auto subaddr = w1.get_subaddress({exp_max.major, i});
      EXPECT_NE(boost::none, w1.get_subaddress_index(subaddr));
    }
    auto subaddr = w1.get_subaddress({exp_max.major, exp_max.minor + 1});
    EXPECT_EQ(boost::none, w1.get_subaddress_index(subaddr));
  };

  // We assume we start with subaddrs for minor indexes 0 to 199
  check_expected_max({0,199});
  check_expected_max({1,199});
  check_expected_max({49,199});
  EXPECT_EQ(w1.get_num_subaddresses(50), 0);

  // Mock receive to {0,150}, so expand from there
  w1.expand_subaddresses({0,150});
  // We should now have subaddresses for minor indexes 0 to 349
  check_expected_max({0,349});
  check_expected_max({1,199});
  check_expected_max({49,199});
  EXPECT_EQ(w1.get_num_subaddresses(50), 0);

  // Now set the minor lookahead 100 higher
  w1.set_subaddress_lookahead(50, 200+100);
  // We should have subaddresses for minor indexes 0 to 449
  check_expected_max({0,449});
  check_expected_max({1,299});
  check_expected_max({49,299});
  EXPECT_EQ(w1.get_num_subaddresses(50), 0);
}

TEST_F(WalletSubaddress, ExpandThenSetMajorIncreaseOnly)
{
  // We assume we are starting with the default setup state
  EXPECT_EQ(50, w1.get_subaddress_lookahead().first);
  EXPECT_EQ(200, w1.get_subaddress_lookahead().second);

  // Helper function to check max subaddrs allocated
  const auto check_expected_max = [&](const cryptonote::subaddress_index exp_max)
  {
    for (uint32_t i = 0; i <= exp_max.minor; ++i)
    {
      auto subaddr = w1.get_subaddress({exp_max.major, i});
      EXPECT_NE(boost::none, w1.get_subaddress_index(subaddr));
    }
    auto subaddr = w1.get_subaddress({exp_max.major, exp_max.minor + 1});
    EXPECT_EQ(boost::none, w1.get_subaddress_index(subaddr));
  };

  // We assume we start with subaddrs for minor indexes 0 to 199
  check_expected_max({0,199});
  check_expected_max({1,199});
  check_expected_max({49,199});
  EXPECT_EQ(w1.get_num_subaddresses(50), 0);

  // Mock receive to {40,0}, so expand from there
  w1.expand_subaddresses({40,0});
  check_expected_max({0,199});
  check_expected_max({1,199});
  check_expected_max({40,199});
  check_expected_max({89,199});
  EXPECT_EQ(w1.get_num_subaddresses(90), 0);

  // Now set the major lookahead 10 higher
  w1.set_subaddress_lookahead(50+10, 200);
  check_expected_max({0,199});
  check_expected_max({1,199});
  check_expected_max({40,199});
  check_expected_max({99,199});
  EXPECT_EQ(w1.get_num_subaddresses(100), 0);
}

TEST_F(WalletSubaddress, ExpandThenSetIncreaseBoth)
{
  // We assume we are starting with the default setup state
  EXPECT_EQ(50, w1.get_subaddress_lookahead().first);
  EXPECT_EQ(200, w1.get_subaddress_lookahead().second);

  // Helper function to check max subaddrs allocated
  const auto check_expected_max = [&](const cryptonote::subaddress_index exp_max)
  {
    for (uint32_t i = 0; i <= exp_max.minor; ++i)
    {
      auto subaddr = w1.get_subaddress({exp_max.major, i});
      EXPECT_NE(boost::none, w1.get_subaddress_index(subaddr));
    }
    auto subaddr = w1.get_subaddress({exp_max.major, exp_max.minor + 1});
    EXPECT_EQ(boost::none, w1.get_subaddress_index(subaddr));
  };

  // We assume we start with subaddrs for minor indexes 0 to 199
  check_expected_max({0,199});
  check_expected_max({1,199});
  check_expected_max({49,199});
  EXPECT_EQ(w1.get_num_subaddresses(50), 0);

  // Mock receive to {40,150}, so expand from there
  w1.expand_subaddresses({40,150});
  check_expected_max({0,199});
  check_expected_max({1,199});
  check_expected_max({40,349});
  check_expected_max({89,199});
  EXPECT_EQ(w1.get_num_subaddresses(90), 0);

  // Now set the major lookahead 10 higher and minor 100 higher
  w1.set_subaddress_lookahead(50+10, 200+100);
  check_expected_max({0,299});
  check_expected_max({1,299});
  check_expected_max({40,449});
  check_expected_max({99,299});
  EXPECT_EQ(w1.get_num_subaddresses(100), 0);
}

```

j-berman (changes requested):
Sorry, @plowsof's comment [here](https://github.com/monero-project/monero/pull/9953#discussion_r2169449342) made me realize a bug in my code (and a separate bug in the original code as well). These changes include a fix and test cases that capture the bugs.

The fix matches the logic of `set_address_lookahead` to `expand_subaddresses`. So repeating @plowsof 's test:

> starting with a `1:10` lookahead. sending to above expanded , then decreasing idx until i find the max idx

With this fix, the table will look like this:

lookahead | max idx
-- | --
1:10 | 9
  | 18
  | 27
1:20 | 46
1:1000 | 1045

It's a little counter-intuitive that the lookahead only reaches the `max idx + lookahead - 1` in each step.

I think that may make sense to change in both `expand_subaddresses` and `set_subaddress_lookahead` in a separate PR. But for now, this matches the logic of `set_subaddress_lookahead` to `expand_subaddresses`.

src/wallet/wallet2.cpp
```diff
@@ -1947,8 +1947,41 @@ void wallet2::set_subaddress_lookahead(size_t major, size_t minor)
...
+  for (index2.major = 0; index2.major < major_end; ++index2.major)
+  {
+    // The existing minor addresses already set for this account
+    const uint32_t n_minor_subaddrs = this->get_num_subaddresses(index2.major);
+
+    // The subaddress lookahead is expected to expand from the max index in expand_subaddresses
+    const uint32_t n_max_idx = n_minor_subaddrs > 0 ? (n_minor_subaddrs - 1) : 0;
+    const uint32_t begin = (n_minor_subaddrs || index2.major < old_major_lookahead) ? get_subaddress_clamped_sum(n_max_idx, old_minor_lookahead) : 0;
+    // The expected new n minor subaddresses allocated for this account
+    const uint32_t end = get_subaddress_clamped_sum(n_max_idx, minor);
```
j-berman:
```suggestion
  const uint32_t max_major_idx = this->get_num_subaddress_accounts() > 0 ? (this->get_num_subaddress_accounts() - 1) : 0;
  const uint32_t major_end = get_subaddress_clamped_sum(max_major_idx, major);
  for (index2.major = 0; index2.major < major_end; ++index2.major)
  {
    // The existing minor addresses already set for this account
    const uint32_t n_minor_subaddrs = this->get_num_subaddresses(index2.major);

    // The subaddress lookahead is expected to expand from the max index in expand_subaddresses
    const uint32_t max_minor_idx = n_minor_subaddrs > 0 ? (n_minor_subaddrs - 1) : 0;
    const uint32_t begin = (n_minor_subaddrs || index2.major < old_major_lookahead) ? get_subaddress_clamped_sum(max_minor_idx, old_minor_lookahead) : 0;
    // The expected new n minor subaddresses allocated for this account
    const uint32_t end = get_subaddress_clamped_sum(max_minor_idx, minor);
```

tests/unit_tests/subaddress.cpp
```diff
@@ -101,3 +102,153 @@ TEST_F(WalletSubaddress, OutOfBoundsIndexes)
...
+  EXPECT_EQ(w1.get_num_subaddresses(90), 0);
+
+  // Now set the major lookahead 10 higher and minor 100 higher
+  w1.set_subaddress_lookahead(50+10, 200+100);
+  check_expected_max({0,299});
+  check_expected_max({1,299});
+  check_expected_max({40,449});
+  check_expected_max({99,299});
+  EXPECT_EQ(w1.get_num_subaddresses(100), 0);
+}
```
j-berman:
```suggestion
// Helper function to check max subaddrs allocated
static void check_expected_max(const tools::wallet2 &w1, const cryptonote::subaddress_index exp_max)
{
  for (uint32_t i = 0; i <= exp_max.minor; ++i)
  {
    auto subaddr = w1.get_subaddress({exp_max.major, i});
    EXPECT_NE(boost::none, w1.get_subaddress_index(subaddr));
  }
  auto subaddr = w1.get_subaddress({exp_max.major, exp_max.minor + 1});
  EXPECT_EQ(boost::none, w1.get_subaddress_index(subaddr));
};

static void expect_default_wallet_state(const tools::wallet2 &w1)
{
  // these tests assume we are starting with the default setup state
  EXPECT_EQ(2, w1.get_num_subaddress_accounts());
  EXPECT_EQ(50, w1.get_subaddress_lookahead().first);
  EXPECT_EQ(200, w1.get_subaddress_lookahead().second);

  // We assume we start with subaddrs for minor indexes 0 to 199
  check_expected_max(w1, {0,199});
  check_expected_max(w1, {1,199});
  check_expected_max(w1, {49,199});
  check_expected_max(w1, {50,199}); // 50 because the test starts with accounts 0 and 1 already allocated
  EXPECT_EQ(boost::none, w1.get_subaddress_index(w1.get_subaddress({51,0})));
}

TEST_F(WalletSubaddress, SetLookahead)
{
  expect_default_wallet_state(w1);
  // get_subaddress_index looks up keys in the private m_subaddresses dictionary so we will use it to test if a key is properly being scanned for
  cryptonote::subaddress_index test_idx = {50, 199};
  auto subaddr = w1.get_subaddress(test_idx);
  EXPECT_NE(boost::none, w1.get_subaddress_index(subaddr));
  // fist test expanding the major lookahead
  w1.set_subaddress_lookahead(100, 200);
  EXPECT_EQ(100, w1.get_subaddress_lookahead().first);
  EXPECT_EQ(200, w1.get_subaddress_lookahead().second);
  check_expected_max(w1, {100, 199});
  // next test expanding the minor lookahead
  w1.set_subaddress_lookahead(100, 300);
  EXPECT_EQ(100, w1.get_subaddress_lookahead().first);
  EXPECT_EQ(300, w1.get_subaddress_lookahead().second);
  check_expected_max(w1, {100, 299});
}

TEST_F(WalletSubaddress, ExpandThenSetMinorIncreaseOnly)
{
  expect_default_wallet_state(w1);

  // Mock receive to {0,150}, so expand from there
  w1.expand_subaddresses({0,150});
  // We should now have subaddresses for minor indexes 0 to 349
  check_expected_max(w1, {0,349});
  check_expected_max(w1, {1,199});
  check_expected_max(w1, {49,199});
  check_expected_max(w1, {50,199});
  EXPECT_EQ(boost::none, w1.get_subaddress_index(w1.get_subaddress({51,0})));

  // Now set the minor lookahead 100 higher
  w1.set_subaddress_lookahead(50, 200+100);
  // We should have subaddresses for minor indexes 0 to 449
  check_expected_max(w1, {0,449});
  check_expected_max(w1, {1,299});
  check_expected_max(w1, {49,299});
  check_expected_max(w1, {50,299});
  EXPECT_EQ(boost::none, w1.get_subaddress_index(w1.get_subaddress({51,0})));
}

TEST_F(WalletSubaddress, ExpandThenSetMajorIncreaseOnly)
{
  expect_default_wallet_state(w1);

  // Mock receive to {40,0}, so expand from there
  w1.expand_subaddresses({40,0});
  check_expected_max(w1, {0,199});
  check_expected_max(w1, {1,199});
  check_expected_max(w1, {40,199});
  check_expected_max(w1, {89,199});
  EXPECT_EQ(boost::none, w1.get_subaddress_index(w1.get_subaddress({90,0})));

  // Now set the major lookahead 10 higher
  w1.set_subaddress_lookahead(50+10, 200);
  check_expected_max(w1, {0,199});
  check_expected_max(w1, {1,199});
  check_expected_max(w1, {40,199});
  check_expected_max(w1, {99,199});
  EXPECT_EQ(boost::none, w1.get_subaddress_index(w1.get_subaddress({100,0})));
}

TEST_F(WalletSubaddress, ExpandThenSetIncreaseBoth)
{
  expect_default_wallet_state(w1);

  // Mock receive to {40,150}, so expand from there
  w1.expand_subaddresses({40,150});
  check_expected_max(w1, {0,199});
  check_expected_max(w1, {1,199});
  check_expected_max(w1, {40,349});
  check_expected_max(w1, {89,199});
  EXPECT_EQ(boost::none, w1.get_subaddress_index(w1.get_subaddress({90,0})));

  // Now set the major lookahead 10 higher and minor 100 higher
  w1.set_subaddress_lookahead(50+10, 200+100);
  check_expected_max(w1, {0,299});
  check_expected_max(w1, {1,299});
  check_expected_max(w1, {40,449});
  check_expected_max(w1, {99,299});
  EXPECT_EQ(boost::none, w1.get_subaddress_index(w1.get_subaddress({100,0})));
}
```

j-berman (changes requested):
This matches `set_subaddress_lookahead` to `expand_addresses` for major indexes as well (the prior only matched behavior for minor subaddrs). Sorry for the spam.

plowsof:
confirming the above behaviour. thanks for going through my ramblings and debugging further @j-berman . for those testing: do note that an output within the expanded area will be detected in the pool but it will not further expand until at least 1 block confirmation. 

closes https://github.com/monero-project/monero/issues/7364 can be added to this one

nahuhh:
before anyone asks for squash, i'm waiting for 9971 merge (ci fix)

plowsof:
sanity checking @nahuhh comment here about wallet-rpc and persistance https://github.com/monero-project/monero/pull/9954#issuecomment-2970715815 (i thought this was a simple case of not calling `store`)

```
curl -X POST http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"set_subaddress_lookahead","params":{"major_idx":"1","minor_idx":"500"}}' -H 'Content-Type: application/json'
<receive tx at max idx 499>
curl -X POST http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"store"}' -H 'Content-Type: application/json'
curl -X POST http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"stop_wallet"}' -H 'Content-Type: application/json'
```
open with cli `set` will show a subaddress lookahead of `50:200` - so `store` does not make this persistent. 
- send to max possible idx of ~999 - nothing~ the max idx would be `998` ****
- ~~send to 700 - but it detected it, huh while set shows `50:200` and last received tx at idx `499`~~
* the original expansion to 998 persists* but the cli resets its lookahead to `5:200`

* @nahuhh shared `get` / `set` `attribute` rpc calls ... so forcibly making this persistent would seem out of scope for this PR - but these do not appear to be for saving the `set` settings https://github.com/monero-project/monero/blob/master/src/wallet/wallet2.h#L1670C1-L1684C1 https://github.com/monero-project/monero/pull/2605

j-berman:
RPC persistence should probably be in this PR imo. Try this patch:

```patch
diff --git a/src/wallet/wallet_rpc_server.cpp b/src/wallet/wallet_rpc_server.cpp
index 1ed8eb9fc..4f349e4bd 100644
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -707,11 +707,24 @@ namespace tools
   bool wallet_rpc_server::on_set_subaddr_lookahead(const wallet_rpc::COMMAND_RPC_SET_SUBADDR_LOOKAHEAD::request& req, wallet_rpc::COMMAND_RPC_SET_SUBADDR_LOOKAHEAD::response& res, epee::json_rpc::error& er, const connection_context *ctx)
   {
     if (!m_wallet) return not_open(er);
-    try {
-      m_wallet->set_subaddress_lookahead(req.major_idx, req.minor_idx);
+    CHECK_IF_BACKGROUND_SYNCING();
+    const std::string wallet_file = m_wallet->get_wallet_file();
+    if (wallet_file == "" || m_wallet->verify_password(req.password))
+    {
+      try
+      {
+        m_wallet->set_subaddress_lookahead(req.major_idx, req.minor_idx);
+        m_wallet->rewrite(wallet_file, req.password);
+      }
+      catch (const std::exception& e) {
+        handle_rpc_exception(std::current_exception(), er, WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR);
+        return false;
+      }
     }
-    catch (const std::exception& e) {
-      handle_rpc_exception(std::current_exception(), er, WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR);
+    else
+    {
+      er.code = WALLET_RPC_ERROR_CODE_INVALID_PASSWORD;
+      er.message = "Invalid password.";
       return false;
     }
     return true;
diff --git a/src/wallet/wallet_rpc_server_commands_defs.h b/src/wallet/wallet_rpc_server_commands_defs.h
index 0871d6369..d0273d162 100644
--- a/src/wallet/wallet_rpc_server_commands_defs.h
+++ b/src/wallet/wallet_rpc_server_commands_defs.h
@@ -186,9 +186,11 @@ namespace wallet_rpc
   {
     struct request_t
     {
+      std::string password;
       uint64_t major_idx;
       uint64_t minor_idx;
       BEGIN_KV_SERIALIZE_MAP()
+        KV_SERIALIZE(password)
         KV_SERIALIZE(major_idx)
         KV_SERIALIZE(minor_idx)
       END_KV_SERIALIZE_MAP()
diff --git a/utils/python-rpc/framework/wallet.py b/utils/python-rpc/framework/wallet.py
index 88cece2bc..d0a516c1f 100644
--- a/utils/python-rpc/framework/wallet.py
+++ b/utils/python-rpc/framework/wallet.py
@@ -334,11 +334,12 @@ class Wallet(object):
         }
         return self.rpc.send_json_rpc_request(generate_from_keys)
 
-    def set_subaddress_lookahead(self, major_idx: int, minor_idx: int):
+    def set_subaddress_lookahead(self, major_idx: int, minor_idx: int, password = ""):
         lookahead = {
             'method': 'set_subaddress_lookahead',
             'jsonrpc': '2.0',
             'params' : {
+                'password': password,
                 'major_idx': major_idx,
                 'minor_idx': minor_idx
             },

```

plowsof:
patch works , thank you!
```
curl -X POST http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"set_subaddress_lookahead","params":{"major_idx":"1","minor_idx":"600","password":""}}' -H 'Content-Type: application/json'
[..store..]
[..stop_wallet..]
```
wallet cli shows the new lookahead value :+1:

src/wallet/wallet2.cpp
```diff
@@ -1947,8 +1947,42 @@ void wallet2::set_subaddress_lookahead(size_t major, size_t minor)
...
+    if (begin >= end)
+      continue;
+
+    const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, begin, end);
+    for (index2.minor = begin; index2.minor < end; ++index2.minor)
+    {
+      const crypto::public_key &D = pkeys.at(index2.minor - begin);
+      m_subaddresses[D] = index2;
+    }
+  }
```
jeffro256: Why not use `this->expand_subaddresses()`? (Before/conditional on setting `m_subaddress_lookahead_major` and `m_subaddress_lookahead_minor` above)

nahuhh:
only changes in the rebase are to apply [this](https://github.com/monero-project/monero/pull/9953#issuecomment-3014064433) patch from @j-berman 
 https://github.com/monero-project/monero/compare/b02f62b58eb15ff8aadad715e28ae080007e1726..6b441a1782d9ecc2248145fa87efe3e099a6db43

src/wallet/wallet2.cpp
j-berman: `expand_subaddresses` expects a user to have received to a specific subaddr index, and then expands the subaddress map (**and adds subaddress labels**) from there. UI's populate subaddresses based on the labels that are set. So we don't want to add any new labels from setting this lookahead. We could have `expand_subaddresses` reuse this logic from here, but I think this approach is reasonable.

jeffro256: I will need to double-check, but I think a lot of the `wallet2` balance calls will break if the subaddress map is populated, but not the subaddress labels. See: https://github.com/monero-project/monero/pull/9934#issuecomment-2945876887. A lot of higher-level calling code iterates through the subaddresses `wallet2` provides using the `get_num_subaddresses()` and `get_num_subaddress_accounts()` calls, which query `m_subaddress_labels`, not `m_subaddresses`.

j-berman: `expand_subaddresses` already does expand the subaddresses map past the subaddress labels the same way that this new logic does (see the new unit tests), so the wallet2 balance calls would be currently broken if true

jeffro256: Okay good point. On first receive during `refresh`, `should_expand` will return `true` for that receive address index if the subaddress lookahead was already set previously, in which case the subaddress labels will be filled up to that index. So basically the account/subaddress counts will be "wrong" until the first receive, but those would have 0 balance in them anyways. You also can't lookup the subaddress by index until it's first receive, which is funky, but that's already the case for lookahead addresses. I still don't like the code duplication though...

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -182,6 +182,29 @@ namespace wallet_rpc
     typedef epee::misc_utils::struct_init<response_t> response;
   };
 
+  struct COMMAND_RPC_SET_SUBADDR_LOOKAHEAD
```
jeffro256, line 185: The RPC minor version should be bumped.

nahuhh: [done](https://github.com/monero-project/monero/compare/6b441a1782d9ecc2248145fa87efe3e099a6db43..cd6c0e9503eccd9c42d2385eadcf2141dad7955e), thanks

jeffro256:
Here's an alternative approach (apply diff to current HEAD):

```diff
diff --git a/src/wallet/wallet2.cpp b/src/wallet/wallet2.cpp
index 23a0880e6..ca8343c51 100644
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1597,39 +1597,62 @@ bool wallet2::should_expand(const cryptonote::subaddress_index &index) const
 //----------------------------------------------------------------------------------------------------
 void wallet2::expand_subaddresses(const cryptonote::subaddress_index& index)
 {
-  hw::device &hwdev = m_account.get_device();
+  // check if index will overflow container (usually only applicable on 32-bit systems)
+  if constexpr (sizeof(std::size_t) <= sizeof(std::uint32_t))
+  {
+    static constexpr std::uint32_t max_idx = static_cast<std::uint32_t>(std::numeric_limits<std::size_t>::max());
+    const bool cannot_label_index = index.major == max_idx || index.minor == max_idx;
+    THROW_WALLET_EXCEPTION_IF(cannot_label_index, error::wallet_internal_error, "subaddress index out of range");
+  }
+
+  // resize subaddress labels just big enough that `m_subaddress_labels[index.major][index.minor]` is present
   if (m_subaddress_labels.size() <= index.major)
+    m_subaddress_labels.resize(index.major + 1, {"Untitled account"});
+  auto &subaddr_labels_in_account = m_subaddress_labels[index.major];
+  if (subaddr_labels_in_account.size() <= index.minor)
+    subaddr_labels_in_account.resize(index.minor + 1);
+  get_account_tags(); //trigger m_account_tags integrity checks
+
+  // compile all indices present in subaddress scanning map, as well as find highest major index
+  std::unordered_set<cryptonote::subaddress_index> all_indices;
+  std::uint32_t max_major_index = 0;
+  for (const auto &p : m_subaddresses)
   {
-    // add new accounts
-    cryptonote::subaddress_index index2;
-    const uint32_t major_end = get_subaddress_clamped_sum(index.major, m_subaddress_lookahead_major);
-    for (index2.major = m_subaddress_labels.size(); index2.major < major_end; ++index2.major)
+    all_indices.insert(p.second);
+    max_major_index = std::max(max_major_index, p.second.major);
+  }
+
+  // find lowest "missing" minor index in map, for all major indices
+  // this is an optimization which allows us to skip re-generating pubkeys that we already have
+  std::vector<std::uint32_t> missing_minor_index(max_major_index + 1);
+  for (std::uint32_t major = 0; major < missing_minor_index.size(); ++major)
+  {
+    std::uint32_t &minor = missing_minor_index[major];
+    while (minor < std::numeric_limits<std::uint32_t>::max())
     {
-      const uint32_t end = get_subaddress_clamped_sum((index2.major == index.major ? index.minor : 0), m_subaddress_lookahead_minor);
-      const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, 0, end);
-      for (index2.minor = 0; index2.minor < end; ++index2.minor)
-      {
-         const crypto::public_key &D = pkeys[index2.minor];
-         m_subaddresses[D] = index2;
-      }
+      if (!all_indices.count({major, minor}))
+        break;
+      ++minor;
     }
-    m_subaddress_labels.resize(index.major + 1, {"Untitled account"});
-    m_subaddress_labels[index.major].resize(index.minor + 1);
-    get_account_tags();
   }
-  else if (m_subaddress_labels[index.major].size() <= index.minor)
+
+  // resize subaddress scanning map to highest historical received subaddress index plus lookahead
+  hw::device &hwdev = m_account.get_device();
+  const std::uint32_t major_base = std::max<std::uint32_t>(m_subaddress_labels.size(), 1) - 1;
+  const std::uint32_t major_end = get_subaddress_clamped_sum(major_base, m_subaddress_lookahead_major);
+  for (std::uint32_t major = 0; major < major_end; ++major)
   {
-    // add new subaddresses
-    const uint32_t end = get_subaddress_clamped_sum(index.minor, m_subaddress_lookahead_minor);
-    const uint32_t begin = m_subaddress_labels[index.major].size();
-    cryptonote::subaddress_index index2 = {index.major, begin};
-    const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, index2.minor, end);
-    for (; index2.minor < end; ++index2.minor)
+    const std::size_t n_minor_labels = (major < m_subaddress_labels.size()) ? m_subaddress_labels.at(major).size() : 0;
+    const std::uint32_t minor_base = std::max<std::uint32_t>(n_minor_labels, 1) - 1;
+    const std::uint32_t minor_end = get_subaddress_clamped_sum(minor_base, m_subaddress_lookahead_minor);
+    const std::uint32_t minor_begin = (major < missing_minor_index.size()) ? missing_minor_index.at(major) : 0;
+    const std::vector<crypto::public_key> pkeys
+      = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), major, minor_begin, minor_end);
+    for (std::uint32_t minor = minor_begin; minor < minor_end; ++minor)
     {
-       const crypto::public_key &D = pkeys[index2.minor - begin];
-       m_subaddresses[D] = index2;
+      const crypto::public_key &D = pkeys.at(minor - minor_begin);
+      m_subaddresses[D] = {major, minor};
     }
-    m_subaddress_labels[index.major].resize(index.minor + 1);
   }
 }
 //----------------------------------------------------------------------------------------------------
@@ -1957,32 +1980,7 @@ void wallet2::set_subaddress_lookahead(size_t major, size_t minor)
   if (old_major_lookahead >= major && old_minor_lookahead >= minor)
     return;
 
-  // Expand the subaddresses map so that outputs received to the higher lookaheads will be identified in the scan loop
-  hw::device &hwdev = m_account.get_device();
-  cryptonote::subaddress_index index2;
-  const uint32_t max_major_idx = this->get_num_subaddress_accounts() > 0 ? (this->get_num_subaddress_accounts() - 1) : 0;
-  const uint32_t major_end = get_subaddress_clamped_sum(max_major_idx, major);
-  for (index2.major = 0; index2.major < major_end; ++index2.major)
-  {
-    // The existing minor addresses already set for this account
-    const uint32_t n_minor_subaddrs = this->get_num_subaddresses(index2.major);
-
-    // The subaddress lookahead is expected to expand from the max index in expand_subaddresses
-    const uint32_t max_minor_idx = n_minor_subaddrs > 0 ? (n_minor_subaddrs - 1) : 0;
-    const uint32_t begin = (n_minor_subaddrs || index2.major < old_major_lookahead) ? get_subaddress_clamped_sum(max_minor_idx, old_minor_lookahead) : 0;
-    // The expected new n minor subaddresses allocated for this account
-    const uint32_t end = get_subaddress_clamped_sum(max_minor_idx, minor);
-
-    if (begin >= end)
-      continue;
-
-    const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, begin, end);
-    for (index2.minor = begin; index2.minor < end; ++index2.minor)
-    {
-      const crypto::public_key &D = pkeys.at(index2.minor - begin);
-      m_subaddresses[D] = index2;
-    }
-  }
+  expand_subaddresses({0, 0});
 }
 //----------------------------------------------------------------------------------------------------
 /*!
diff --git a/src/wallet/wallet2.h b/src/wallet/wallet2.h
index 47b1753d7..41f1dffaa 100644
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -1124,6 +1124,13 @@ private:
     size_t get_num_subaddress_accounts() const { return m_subaddress_labels.size(); }
     size_t get_num_subaddresses(uint32_t index_major) const { return index_major < m_subaddress_labels.size() ? m_subaddress_labels[index_major].size() : 0; }
     void add_subaddress(uint32_t index_major, const std::string& label); // throws when index is out of bound
+    /**
+     * brief: expand subaddress labels up to `index`, and scanning map to highest labeled index plus current lookahead
+     * param: index -
+     *
+     * All calls to `expand_subaddresses()` will *always* expand the subaddress map if the lookahead
+     * values have been increased since the last call.
+     */
     void expand_subaddresses(const cryptonote::subaddress_index& index);
     void create_one_off_subaddress(const cryptonote::subaddress_index& index);
     std::string get_subaddress_label(const cryptonote::subaddress_index& index) const;
```

As compared to current `master`, this mainly touches just `wallet2::expand_subaddresses()`, and calls `wallet2::expand_subaddresses({0, 0})` in `wallet2::set_subaddress_lookahead()`. This keeps all the expansion logic in one place, as compared to the current iteration of this PR. It doesn't expand labels unnecessarily like my previous suggestion, and it also includes an optimization which speeds up expansion when the map is mostly present (which is useful for people which large lookaheads). I can remove the optimization to simplify this approach if desired.

jeffro256:
Here's a functional test that tests live transfers to subaddresses at high lookaheads, IDK, might be worth adding?:

```diff
diff --git a/tests/functional_tests/transfer.py b/tests/functional_tests/transfer.py
index 0df870e9c..f1fb6e51d 100755
--- a/tests/functional_tests/transfer.py
+++ b/tests/functional_tests/transfer.py
@@ -86,6 +86,7 @@ class TransferTest():
         self.check_subtract_fee_from_outputs()
         self.check_background_sync()
         self.check_background_sync_reorg_recovery()
+        self.check_subaddress_lookahead()
 
     def reset(self):
         print('Resetting blockchain')
@@ -1531,5 +1532,47 @@ class TransferTest():
         self.wallet[0].close_wallet()
         self.wallet[0].restore_deterministic_wallet(seed = seeds[0])
 
+    def check_subaddress_lookahead(self):
+        daemon = Daemon()
+
+        print('Testing transfers to subaddresses with large lookahead')
+
+        # From wallet 1 to wallet 0 at subaddress (0, 999)
+
+        address_0_999 = '8BQKgTSSqJjP14AKnZUBwnXWj46MuNmLvHfPTpmry52DbfNjjHVvHUk4mczU8nj8yZ57zBhksTJ8kM5xKeJXw55kCMVqyG7' # this is the address for address 999 of the main account in the test wallet
+        try:  # assert address_1_999 is not in the current pubkey table
+            self.wallet[0].get_address_index(address_0_999)
+            assert False # address should not already be loaded
+        except Exception as e:
+            assert str(e) ==  "{'error': {'code': -2, 'message': \"Address doesn't belong to the wallet\"}, 'id': '0', 'jsonrpc': '2.0'}"
+        # update the lookahead and assert the high index address is now in the table
+        self.wallet[0].set_subaddress_lookahead(50, 1000)
+        res = self.wallet[0].get_address_index(address_0_999)
+        assert res['index']['major'] == 0
+        assert res['index']['minor'] == 999
+
+        dst = {'address': address_0_999, 'amount': 454545454545}
+
+        self.wallet[1].refresh()
+        assert self.wallet[1].get_balance().balance > dst['amount']
+        self.wallet[1].transfer([dst])
+        daemon.generateblocks('46r4nYSevkfBUMhuykdK3gQ98XDqDTYW1hNLaXNvjpsJaSbNtdXh1sKMsdVgqkaihChAzEy29zEDPMR3NHQvGoZCLGwTerK', 1)
+        self.wallet[0].refresh()
+
+        res = self.wallet[0].get_balance()
+        balance_info_0_999 = None
+        for balance_info in res['per_subaddress']:
+            if balance_info['account_index'] == 0 and balance_info['address_index'] == 999:
+                balance_info_0_999 = balance_info
+                break
+        assert balance_info_0_999 is not None, "balance info for address (0, 999) not found"
+        assert balance_info_0_999['address'] == address_0_999
+        assert balance_info_0_999['balance'] == dst['amount']
+        assert balance_info_0_999['unlocked_balance'] == 0
+        assert balance_info_0_999['label'] == ''
+        assert balance_info_0_999['num_unspent_outputs'] == 1
+        assert balance_info_0_999['blocks_to_unlock'] == 9
+        assert balance_info_0_999['time_to_unlock'] == 0
+
 if __name__ == '__main__':
     TransferTest().run_test()

```

j-berman:
I'm fine with de-duplicating logic, calling `expand_subaddresses` with {0,0}, and modifying `expand_subaddresses` to have correct behavior for both uses. Imo it's riskier because `expand_subaddresses` is currently used in a wide number of places (whereas setting the subaddress lookahead is rare), and so we need to make sure not to mess up `expand_subaddresses` here too, but it does yield easier-to-maintain code long-term, so good with me.

Looks like there are a couple problems with the updated `expand_subaddresses` in the patch:

- It looks possible for `minor_begin` to be >= `minor_end` (e.g. after reducing the lookahead and then receiving, or even just receiving normally.. I think tests should catch this), which would cause `get_subaddress_spend_public_keys` to throw. Easily fixed by continuing if `minor_begin >= minor_end` (EDIT: `get_subaddress_spend_public_keys` would only throw if `minor_begin > minor_end`, and would do an unnecessary point de-compression if `minor_begin == minor_end`).
- If someone has created a very high one-off subaddress, it will be identified as the max major, and that `missing_minor_index` vector will be unnecessarily huge.

nahuhh:
Force pushed to remove extra line on EOF and to rerun tests (passing now)

jeffro256:
> I'm fine with de-duplicating logic, calling `expand_subaddresses` with {0,0}, and modifying `expand_subaddresses` to have correct behavior for both uses. Imo it's riskier because `expand_subaddresses` is currently used in a wide number of places (whereas setting the subaddress lookahead is rare), and so we need to make sure not to mess up `expand_subaddresses` here too, but it does yield easier-to-maintain code long-term, so good with me.
> 
> Looks like there are a couple problems with the updated `expand_subaddresses` in the patch:
> 
>     * It looks possible for `minor_begin` to be >= `minor_end` (e.g. after reducing the lookahead and then receiving, or even just receiving normally.. I think tests should catch this), which would cause `get_subaddress_spend_public_keys` to throw. Easily fixed by continuing if `minor_begin >= minor_end` (EDIT: `get_subaddress_spend_public_keys` would only throw if `minor_begin > minor_end`, and would do an unnecessary point de-compression if `minor_begin == minor_end`).
> 
>     * If someone has created a very high one-off subaddress, it will be identified as the max major, and that `missing_minor_index` vector will be unnecessarily huge.

Excellent points. Here's an updated diff:

```diff
diff --git a/src/wallet/wallet2.cpp b/src/wallet/wallet2.cpp
index 23a0880e6..94f198e36 100644
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1597,39 +1597,60 @@ bool wallet2::should_expand(const cryptonote::subaddress_index &index) const
 //----------------------------------------------------------------------------------------------------
 void wallet2::expand_subaddresses(const cryptonote::subaddress_index& index)
 {
-  hw::device &hwdev = m_account.get_device();
-  if (m_subaddress_labels.size() <= index.major)
+  // check if index will overflow container (usually only applicable on 32-bit systems)
+  if constexpr (sizeof(std::size_t) <= sizeof(std::uint32_t))
   {
-    // add new accounts
-    cryptonote::subaddress_index index2;
-    const uint32_t major_end = get_subaddress_clamped_sum(index.major, m_subaddress_lookahead_major);
-    for (index2.major = m_subaddress_labels.size(); index2.major < major_end; ++index2.major)
-    {
-      const uint32_t end = get_subaddress_clamped_sum((index2.major == index.major ? index.minor : 0), m_subaddress_lookahead_minor);
-      const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, 0, end);
-      for (index2.minor = 0; index2.minor < end; ++index2.minor)
-      {
-         const crypto::public_key &D = pkeys[index2.minor];
-         m_subaddresses[D] = index2;
-      }
-    }
+    static constexpr std::uint32_t max_idx = static_cast<std::uint32_t>(std::numeric_limits<std::size_t>::max());
+    const bool cannot_label_index = index.major == max_idx || index.minor == max_idx;
+    THROW_WALLET_EXCEPTION_IF(cannot_label_index, error::wallet_internal_error, "subaddress index out of range");
+  }
+
+  // resize subaddress labels just big enough that `m_subaddress_labels[index.major][index.minor]` is present
+  if (m_subaddress_labels.size() <= index.major)
     m_subaddress_labels.resize(index.major + 1, {"Untitled account"});
-    m_subaddress_labels[index.major].resize(index.minor + 1);
-    get_account_tags();
+  auto &subaddr_labels_in_account = m_subaddress_labels[index.major];
+  if (subaddr_labels_in_account.size() <= index.minor)
+    subaddr_labels_in_account.resize(index.minor + 1);
+  get_account_tags(); //trigger m_account_tags integrity checks
+
+  // compile all indices present in subaddress scanning map, as well as every major index
+  std::unordered_set<cryptonote::subaddress_index> all_indices;
+  std::unordered_map<std::uint32_t, std::uint32_t> lowest_missing_minor;
+  for (const auto &p : m_subaddresses)
+  {
+    all_indices.insert(p.second);
+    lowest_missing_minor[p.second.major];
   }
-  else if (m_subaddress_labels[index.major].size() <= index.minor)
+
+  // find lowest "missing" minor index in map, for all major indices
+  // this is an optimization which allows us to skip re-generating pubkeys that we already have
+  for (auto &p : lowest_missing_minor)
   {
-    // add new subaddresses
-    const uint32_t end = get_subaddress_clamped_sum(index.minor, m_subaddress_lookahead_minor);
-    const uint32_t begin = m_subaddress_labels[index.major].size();
-    cryptonote::subaddress_index index2 = {index.major, begin};
-    const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, index2.minor, end);
-    for (; index2.minor < end; ++index2.minor)
+    const std::uint32_t major = p.first;
+    std::uint32_t &minor = p.second;
+    while (all_indices.count({major, minor}) && minor < std::numeric_limits<std::uint32_t>::max())
+      ++minor;
+  }
+
+  // resize subaddress scanning map to highest historical received subaddress index plus lookahead
+  hw::device &hwdev = m_account.get_device();
+  const std::uint32_t major_base = std::max<std::uint32_t>(m_subaddress_labels.size(), 1) - 1;
+  const std::uint32_t major_end = get_subaddress_clamped_sum(major_base, m_subaddress_lookahead_major);
+  for (std::uint32_t major = 0; major < major_end; ++major)
+  {
+    const std::size_t n_minor_labels = (major < m_subaddress_labels.size()) ? m_subaddress_labels.at(major).size() : 0;
+    const std::uint32_t minor_base = std::max<std::uint32_t>(n_minor_labels, 1) - 1;
+    const std::uint32_t minor_end = get_subaddress_clamped_sum(minor_base, m_subaddress_lookahead_minor);
+    const std::uint32_t minor_begin = lowest_missing_minor.count(major) ? lowest_missing_minor.at(major) : 0;
+    if (minor_begin >= minor_end)
+      continue;
+    const std::vector<crypto::public_key> pkeys
+      = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), major, minor_begin, minor_end);
+    for (std::uint32_t minor = minor_begin; minor < minor_end; ++minor)
     {
-       const crypto::public_key &D = pkeys[index2.minor - begin];
-       m_subaddresses[D] = index2;
+      const crypto::public_key &D = pkeys.at(minor - minor_begin);
+      m_subaddresses[D] = {major, minor};
     }
-    m_subaddress_labels[index.major].resize(index.minor + 1);
   }
 }
 //----------------------------------------------------------------------------------------------------
@@ -1957,32 +1978,7 @@ void wallet2::set_subaddress_lookahead(size_t major, size_t minor)
   if (old_major_lookahead >= major && old_minor_lookahead >= minor)
     return;
 
-  // Expand the subaddresses map so that outputs received to the higher lookaheads will be identified in the scan loop
-  hw::device &hwdev = m_account.get_device();
-  cryptonote::subaddress_index index2;
-  const uint32_t max_major_idx = this->get_num_subaddress_accounts() > 0 ? (this->get_num_subaddress_accounts() - 1) : 0;
-  const uint32_t major_end = get_subaddress_clamped_sum(max_major_idx, major);
-  for (index2.major = 0; index2.major < major_end; ++index2.major)
-  {
-    // The existing minor addresses already set for this account
-    const uint32_t n_minor_subaddrs = this->get_num_subaddresses(index2.major);
-
-    // The subaddress lookahead is expected to expand from the max index in expand_subaddresses
-    const uint32_t max_minor_idx = n_minor_subaddrs > 0 ? (n_minor_subaddrs - 1) : 0;
-    const uint32_t begin = (n_minor_subaddrs || index2.major < old_major_lookahead) ? get_subaddress_clamped_sum(max_minor_idx, old_minor_lookahead) : 0;
-    // The expected new n minor subaddresses allocated for this account
-    const uint32_t end = get_subaddress_clamped_sum(max_minor_idx, minor);
-
-    if (begin >= end)
-      continue;
-
-    const std::vector<crypto::public_key> pkeys = hwdev.get_subaddress_spend_public_keys(m_account.get_keys(), index2.major, begin, end);
-    for (index2.minor = begin; index2.minor < end; ++index2.minor)
-    {
-      const crypto::public_key &D = pkeys.at(index2.minor - begin);
-      m_subaddresses[D] = index2;
-    }
-  }
+  expand_subaddresses({0, 0});
 }
 //----------------------------------------------------------------------------------------------------
 /*!
diff --git a/src/wallet/wallet2.h b/src/wallet/wallet2.h
index 47b1753d7..41f1dffaa 100644
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -1124,6 +1124,13 @@ private:
     size_t get_num_subaddress_accounts() const { return m_subaddress_labels.size(); }
     size_t get_num_subaddresses(uint32_t index_major) const { return index_major < m_subaddress_labels.size() ? m_subaddress_labels[index_major].size() : 0; }
     void add_subaddress(uint32_t index_major, const std::string& label); // throws when index is out of bound
+    /**
+     * brief: expand subaddress labels up to `index`, and scanning map to highest labeled index plus current lookahead
+     * param: index -
+     *
+     * All calls to `expand_subaddresses()` will *always* expand the subaddress map if the lookahead
+     * values have been increased since the last call.
+     */
     void expand_subaddresses(const cryptonote::subaddress_index& index);
     void create_one_off_subaddress(const cryptonote::subaddress_index& index);
     std::string get_subaddress_label(const cryptonote::subaddress_index& index) const;
```

plowsof:
@jeffro256 after testing combined with @j-berman s further review + bug fixing / improvements,  should we push for this PR to be merged as-is for an upcoming release then get jeffros latest patch ready for a follow up point release? thoughts?

j-berman:
I'm for applying @jeffro256 's patches ([this](https://github.com/monero-project/monero/pull/9953#issuecomment-3025275244) and [this](https://github.com/monero-project/monero/pull/9953#issuecomment-3038208273)) on top of the latest. I would approve such a PR.

src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -182,6 +182,29 @@ namespace wallet_rpc
...
+  {
+    struct request_t
+    {
+      std::string password;
+      uint64_t major_idx;
+      uint64_t minor_idx;
```
jeffro256: These should be `uint32_t`, not `uint64_t`.

jeffro256 (approved):
Very minor change requested and then I'm ready to approve. Great work, all.

nahuhh:
Accidentally pushed old branch with first force-push. Good now

plowsof:
tested as having the same functionality prior to the latest 2 patches :+1:

PR #9954, opened 2025-06-12
wallet: immediate lookahead table expansion + set_subaddresss_lookahead RPC endpoint [0.18]

nahuhh:
release version of #9953 / #8981


nahuhh:
I've tested wallet-cli's `set subaddress-lookahead` and it works as expected.

things to note:
- this does not work retroactively. If you missed any transactions, you'll have to scan_tx for them
- wallet-rpc is not persistent. I'm not sure how this behaves in practice. Example: change lookahead from default 200 to 500 -> receive tx on minor index 500 -> lookahead is now to index 1000 -> restart wallet = does the wallet forget about subaddress 701-1000? Or does it just not expand until it reaches 800 now?

future:
- wallet-rpc should be made persistent, perhaps in a follow-up pr.
- perhaps subaddress-lookahead should rollback the wallet height to the last known transaction, so missed transactions are found automatically

plowsof:
>wallet-rpc is not persistent. 

Currently wallet rpc cache has to be saved manually with `store`, i dont know of anything which triggers store off the top of my head. *ctrl+c? :smile:

plowsof:
i've went down a rabbit hole sanity checking the effects of changing `set subaddress-lookahead` without this patch, these instructions will help us to test this PR https://github.com/monero-project/monero/issues/7364#issuecomment-2981850915

j-berman (approved):

jeffro256 (approved):

PR #9963, opened 2025-06-21
Remove unused blake2b, provide random.h & chacha with cryptopp, add crypto::twofish_{en.de}crypt()

nabijaczleweli:
As described in https://bounties.monero.social/posts/91/1-000m-twofish-addition-to-monero
Analysis in https://git.sr.ht/~nabijaczleweli/twofish-comparison


tobtoht:
I'm generally against the introduction of new third-party dependencies, unless there is a strong justification.

jeffro256:
Do not remove Blake2b. It is [used downstream](https://github.com/seraphis-migration/monero/blob/2696c270f066c7086ea34aa0724de0b71ce5c3cb/src/carrot_core/hash_functions.cpp#L50-L67) in Carrot.

jeffro256:
> Why include twofish?

The original reason for including Twofish, which wasn't mentioned in the original bounty post by @0xFFFC0000, was to support Jamtis addresses. Though, AFAIK, since no one is currently doing the work to integrate Jamtis at the next hard fork, perhaps this PR is a bit pre-mature. (It might be me eventually...)

> There's no justification for this, and chacha20 is fine

IIRC Twofish is faster than chacha20, especially for exactly 16-byte blocks like Jamtis would be using, but I would have to check if this even matters performance-wise for the normal scan flow with Jamtis' view tag design.

One more subjective point: I dislike crypto++'s API. Object-oriented programming style for hash functions / cipher functions is gross.

nabijaczleweli:
As you've no doubt seen in the write-up, the measurements at megabyte-scale don't really stack up well for twofish speed-wise (~100M/s) against chacha (600M/s (chacha20) & 1200M/s (chacha8) for the cryptopp implementation, half that for the currently-vendored-in-monero implementation).

You can see the API comparison [here](https://git.sr.ht/~nabijaczleweli/twofish-comparison/tree/trunk/item/integration.cpp#L61) or [here](https://git.sr.ht/~nabijaczleweli/twofish-comparison/tree/trunk/item/bench.cpp#L39); none of these are, like, great, but I don't strictly think it matters all that much when used sparingly (like, twice: twofish, chacha, maybe blake2b); I picked the one with the best perf characteristics.

I opted to maximise functionality replaced for the initial submission; reverted the random.h replacement (which would otherwise be a 3x speed improvement but somehow I feel like it's hardly a bottle-neck).

nahuhh:
I'm sure you're aware, but this doesn't build on the CI (crypto++ cmake version)

nabijaczleweli:
Yeah, but I'd rather a verdict on the patchset in principle before I chase down more god-awful cmake for potentially no reason, as I'm sure you understand.

0xFFFC0000:
Adding an entire crypto library because of single algorithm not worth the hassle IMHO.

As side note, analysis here [1] is worth reading though, although it can ( and should ) go deeper. And if there are faster implementation of any of our algorithm in the industry, I am generally in favor of using the industry standard. 

1. https://git.sr.ht/~nabijaczleweli/twofish-comparison

nabijaczleweli:
Two, we can replace the homebrew chacha with it for a 2x speed-up.

This discourse (as a continuation of https://bounties.monero.social/posts/91/1-000m-twofish-addition-to-monero) keeps failing into a deadlock. You don't want to use a library *and* you don't want to embed an implementation. Those are the only two options. Yes, neither of them are optimal, but *either* we embed a real cryptographic library and use it *or* we slop in another single-file implementation. There isn't a third solution here.

There are three viable implementations:
1. copying in the KeePassKit files
2. building cryptopp
3. building botan

these are all functionally just about equivalent; cryptopp is marginally faster on average, widely used, actively maintained, and allows us to kick out another copied-in (and [UB-ridden](https://github.com/llvm/llvm-project/issues/144964)) implementation.

A similar list of benefits and drawbacks can be produced for botan and KeePassKit. I can implement this just as well by copying in the KeePassKit files if you want, but I suspect the first review would be "we should use a battle-tested library for this", or with botan but it would be a re-hash of this so far, so I'd rather ensure it won't be with a concrete opinion before doing that. The back-end doesn't leak into the abstraction, even if the API sucks.

Also, what more more would you like to see from the analysis? I thought I went overly exhaustive tbh.

nabijaczleweli:
Just to put my money where my mouth is: #9979 uses the KeePassKit back-end

PR #9975, opened 2025-07-04
build.sh: enhanced shell scripts

githubhjs:
### Summary

- Added quotes around log directory paths in the guix build script so filenames containing spaces are handled safely

- Quoted the output path for depends build IDs to prevent word splitting during redirection

- Ensured the gzipped source archive is written using a quoted path to avoid globbing issues

### Testing

✅ shellcheck -x contrib/guix/libexec/build.sh

### DIFF
--- a/contrib/guix/libexec/build.sh
+++ b/contrib/guix/libexec/build.sh
@@ -57,8 +57,8 @@ export SOURCE_DATE_EPOCH=1397818193
 
 # Collect some information about the build environment to help debug potential reproducibility issues
 mkdir -p "${LOGDIR}"
-ls -1 /gnu/store | sort > ${LOGDIR}/guix-hashes.txt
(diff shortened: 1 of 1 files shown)


nahuhh:
NACK

edit: maybe this makes sense if $LOGDIR can possibly contain spaces. I'm not familiar with the paths that guix runs on

iamamyth (approved):
These changes look reasonable. Regarding whether the variables can or cannot include spaces, programs should never assume what they cannot establish as fact, and should keep assumptions to a minimum, as they complicate reasoning: If the author needs (not merely wants) LOGDIR not to contain spaces, then the script should disallow a LOGDIR value with spaces, which it doesn't.

nahuhh:
@iamamyth i'm unfamiliar with guix, but with gitian's docker build (iirc) the build paths are always the same, so it would not include any unexpected variables.

but i agree that the changes are sane. (running the shellcheck linter)

most of the file already wraps variables in double quotes, but only some variables are wrapped in curly braces (like line 296 vs 297)


Also: need to 
1. fix/remove the merge commit (in future, pr should have come from the `codex/find-bug` branch (not master))
2. ideally prefix the pr and commit with `guix:`

selsta (approved):

selsta:
Please remove the merge commit.

selsta:
ping @githubhjs please squash your commits

PR #9982, opened 2025-07-11
rpc: return error correctly on bad key image string [RELEASE]

jeffro256:
#9981


selsta:
@jeffro256 this does not compile

jeffro256:
Thanks, I forgot that the release branch isn't on C++17 and copied the same line changes from master

tobtoht (approved):

PR #9996, opened 2025-07-14
set do_not_relay always false in submit_multisig_main

SNeedlewoods:
This is supposed to be a fix for #9698

There I had the misunderstanding that the `do-not-relay` option means we don't want to broadcast any tx to the daemon, but now I think we actually still want to allow submitting transactions (via CLI commands `submit_transfer` or in this case `submit_multisig`) even if `do-not-relay` is set true.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -1639,8 +1639,8 @@ bool simple_wallet::submit_multisig_main(const std::vector<std::string> &args, b
       return false;
     }
 
-    // actually commit or save the transactions
-    commit_or_save(txs.m_ptx, m_do_not_relay);
+    // actually commit the transactions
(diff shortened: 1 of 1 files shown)


nahuhh:
If do_not_relay is set, you intentionally do not want the tx to be relayed.

Forcing it to false sounds like unintended behavior

SNeedlewoods:
AFAIU this option is only meant for `transfer`/`sweep` commands, not for submitting, else `submit_transfer` is also wrong, it's also forcing `do_not_relay = false` [here](https://github.com/monero-project/monero/blob/a1aef9e01710644095449a66c0c40a47ac21065e/src/simplewallet/simplewallet.cpp#L7853)

nahuhh:
Makes sense, thanks

nahuhh (approved):

PR #9463, opened 2024-08-28
fuzz: fix `make fuzz`

jeffro256:
Set CMake variable `OSSFUZZ=ON` in Makefile `fuzz` target

Thanks @ACK-J for bringing up the issue that `make fuzz` was broken

PR #10004, opened 2025-07-18
tests: Add new fuzzer for RPC endpoints

arthurscchan:
This PR adds a new fuzzer in the tests/fuzz/fuzz_rpc directory and updates the CMakeLists.txt accordingly. The fuzzer is built in two versions, each targeting a different set of RPC endpoint functions in the src/rpc directory.

### DIFF
--- a/tests/fuzz/CMakeLists.txt
+++ b/tests/fuzz/CMakeLists.txt
@@ -26,6 +26,35 @@
 # STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 # THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+monero_add_minimal_executable(fuzz_rpc
+  fuzz_rpc/initialisation.cpp
+  fuzz_rpc/rpc_endpoints.cpp
+  fuzz_rpc/fuzz_rpc.cpp)
+target_compile_definitions(fuzz_rpc PRIVATE SAFE)
+target_link_libraries(fuzz_rpc
+  PRIVATE
+    rpc
+    ${CMAKE_THREAD_LIBS_INIT}
+    ${EXTRA_LIBRARIES}
+    $ENV{LIB_FUZZING_ENGINE})
+set_property(TARGET fuzz_rpc
+  PROPERTY
+    FOLDER "tests")
+
+monero_add_minimal_executable(fuzz_rpc_full
+  fuzz_rpc/initialisation.cpp
+  fuzz_rpc/rpc_endpoints.cpp
+  fuzz_rpc/fuzz_rpc.cpp)
+target_link_libraries(fuzz_rpc_full
+  PRIVATE
+    rpc
+    ${CMAKE_THREAD_LIBS_INIT}
+    ${EXTRA_LIBRARIES}
+    $ENV{LIB_FUZZING_ENGINE})
+set_property(TARGET fuzz_rpc_full
+  PROPERTY
+    FOLDER "tests")
+
 monero_add_minimal_executable(block_fuzz_tests block.cpp fuzzer.cpp)
 target_link_libraries(block_fuzz_tests
   PRIVATE
--- a/tests/fuzz/fuzz_rpc/fuzz_rpc.cpp
+++ b/tests/fuzz/fuzz_rpc/fuzz_rpc.cpp
@@ -1,0 +1,116 @@
+#include "initialisation.h"
+#include "rpc_endpoints.h"
+#include <fuzzer/FuzzedDataProvider.h>
+
+#include <csignal>
+#include <csetjmp>
+#include <iostream>
+#include <vector>
+#include <algorithm>
+
+// Helper definition for timeout recovery
+static sigjmp_buf jump_env;
+void timeout_handler(int sig) {
+  if (tools::performance_timers) {
+    tools::performance_timers->clear();
+    delete tools::performance_timers;
+    tools::performance_timers = nullptr;
+  }
+
+  siglongjmp(jump_env, 1);
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  // In general an iteration need a fair amount of data so ensure we have enough
+  // to work with, otherwise return 0 to skip this iteration.
+  if (size < 512) {
+    return 0;
+  }
+
+  // Determine if this iteration should run in safe mode
+#ifdef SAFE
+  constexpr bool is_safe_mode = true;
+#else
+  constexpr bool is_safe_mode = false;
+#endif
+
+  // Retrieve a list of all fuzz targets
+  auto fuzz_targets = get_fuzz_targets(is_safe_mode);
+
+  // Disable fatal exits for logging.
+  el::Loggers::addFlag(el::LoggingFlag::DisableApplicationAbortOnFatalLog);
+
+  // Register a handler for timeout alarm signal
+  signal(SIGALRM, timeout_handler);
+
+  // Prepare base FuzzedDataProvider
+  FuzzedDataProvider provider(data, size);
+
+  // Randomly choose multiple fuzz_targets to fuzz
+  int rpc_messages_to_sent = provider.ConsumeIntegralInRange<int>(1, 16);
+  std::vector<int> selectors;
+  selectors.reserve(rpc_messages_to_sent);
+  for (int i = 0; i < rpc_messages_to_sent && provider.remaining_bytes() >= 2; ++i) {
+    uint16_t raw = provider.ConsumeIntegral<uint16_t>();
+    selectors.push_back(raw % fuzz_targets.size());
+  }
+
+  // Initialise core and core_rpc_server
+  auto dummy_core = initialise_rpc_core();
+  auto rpc_handler = initialise_rpc_server(*dummy_core, provider, !is_safe_mode);
+
+  // Generalise random blocks/miners/transactions and push to the core blockchains
+  if (!generate_random_blocks(*dummy_core, provider)) {
+    // No randomised blocks has been successfully added, skipping this iteration
+    dummy_core->get_blockchain_storage().get_db().batch_stop();
+    return 0;
+  }
+
+  // Disable bootstrap daemon
+  disable_bootstrap_daemon(*rpc_handler->rpc);
+
+  for (int selector : selectors) {
+    // Configure returning point for timeout
+    if (sigsetjmp(jump_env, 1) == 0) {
+      try {
+        // Start a timeout alarm for 120 secs
+        // SIGALARM will be sent once 120 is up
+        // timeout_handler catches the SIGALARM and return to the
+        // last sigsetjmp call with matching value to allow the code
+        // skipping to the next iteration
+        alarm(120);
+
+        // Fuzz the target function
+        fuzz_targets[selector](*rpc_handler->rpc, provider);
+      } catch (const std::runtime_error&) {
+        // Known runtime_error thrown from monero
+      } catch (const cryptonote::DB_ERROR& e) {
+        // Known error thrown from monero on internal blockchain DB check
+        // when fuzzing with random values
+      }
+
+      // Clear the alarm before next iteration
+      alarm(0);
+    }
+  }
+
+  // Try fuzz the prunt_blockchains API at the end to avoid
+  // multiple prune_blockchains call that result in db lock stall.
+  if ((provider.remaining_bytes() > 0) && !is_safe_mode) {
+    if (sigsetjmp(jump_env, 1) == 0) {
+      try {
+        alarm(120);
+        fuzz_prune_blockchain(*rpc_handler->rpc, provider);
+      } catch (const std::runtime_error&) {
+        // Known runtime_error thrown from monero
+      } catch (const cryptonote::DB_ERROR& e) {
+        // Known error thrown from monero on internal blockchain DB check
+        // when fuzzing with random values
+      }
+      alarm(0);
+    }
+  }
+
+  dummy_core->get_blockchain_storage().get_db().batch_stop();
+  return 0;
+}
--- a/tests/fuzz/fuzz_rpc/initialisation.cpp
+++ b/tests/fuzz/fuzz_rpc/initialisation.cpp
@@ -1,0 +1,369 @@
+#include "initialisation.h"
+
+#include "crypto/crypto.h"
+#include "cryptonote_config.h"
+#include "cryptonote_basic/cryptonote_basic.h"
+#include "cryptonote_basic/cryptonote_basic_impl.h"
+#include "cryptonote_basic/cryptonote_format_utils.h"
+#include "cryptonote_basic/tx_extra.h"
+#include "cryptonote_core/cryptonote_core.h"
+#include "cryptonote_core/cryptonote_tx_utils.h"
+#include "cryptonote_core/tx_pool.h"
+#include "cryptonote_protocol/cryptonote_protocol_handler.h"
+#include "p2p/net_node.h"
+#include "p2p/net_node.inl"
+#include "ringct/rctTypes.h"
+#include "serialization/binary_utils.h"
+#include "serialization/variant.h"
+
+#include <boost/program_options/variables_map.hpp>
+#include <vector>
+#include <string>
+#include <functional>
+#include <unordered_map>
+
+// Configure a dummy protocol to avoid external requests
+bool DummyProtocol::is_synchronized() const {
+  return true;
+}
+
+bool DummyProtocol::relay_transactions(cryptonote::NOTIFY_NEW_TRANSACTIONS::request&, const boost::uuids::uuid&, epee::net_utils::zone, cryptonote::relay_method) {
+  return true;
+}
+
+bool DummyProtocol::relay_block(cryptonote::NOTIFY_NEW_FLUFFY_BLOCK::request&, cryptonote::cryptonote_connection_context&) {
+  return true;
+}
+
+bool DummyProtocol::get_payload_sync_data(cryptonote::CORE_SYNC_DATA&) {
+  return true;
+}
+
+bool DummyProtocol::get_payload_sync_data(epee::byte_slice&) {
+  return true;
+}
+
+bool DummyProtocol::on_idle() {
+  return true;
+}
+
+void DummyProtocol::on_connection_closing(cryptonote::cryptonote_connection_context&) {}
+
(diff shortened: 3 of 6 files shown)


arthurscchan:
OSS-Fuzz build steps to test the new fuzzers.
```
# Create test directory
mkdir test
cd test

# Clone OSS-Fuzz
git clone https://github.com/arthurscchan/oss-fuzz
cd oss-fuzz

# Checkout branch for new fuzzer build
git fetch origin add-build-fuzz-rpc-monero
git checkout add-build-fuzz-rpc-monero

# Slight patch of OSS-Fuzz Docker file and build script to test this new fuzzer.
sed -i 's|--depth 1 https://github.com/monero-project/monero.git|https://github.com/arthurscchan/monero.git|g' projects/monero/Dockerfile
sed -i '/^cd monero$/a\
git fetch origin add-rpc-fuzzing-harness\
git checkout add-rpc-fuzzing-harness' projects/monero/build.sh

# Build fuzzer with OSS-Fuzz helper
python3 infra/helper.py build_fuzzers monero

# Test run of fuzzer fuzz_rpc
python3 infra/helper.py run_fuzzer monero fuzz_rpc

# Test run of fuzzer fuzz_rpc_full
python3 infra/helper.py run_fuzzer monero fuzz_rpc_full

# Test run of fuzzer fuzz_zmq
python3 infra/helper.py run_fuzzer monero fuzz_zmq
```

ACK-J:
This PR is a deliverable by AdaLogics for the RPC Fuzzing Magic Monero Fund project:
https://donate.magicgrants.org/monero/projects/fuzzing-monero-rpc

Thank you donors!

DavidKorczynski:
We have added a parallel PR in OSS-Fuzz that is needed to enable these fuzzers on OSS-Fuzz: https://github.com/google/oss-fuzz/pull/13691

We expect that there may be follow-up improvements on the fuzzing harnesses once they start running on OSS-Fuzz. We have run these harnesses for 10+ hours on our machines, but when OSS-Fuzz throws a lot more compute on it, then it's likely we will observe things that need changing, either because of missed coverage (we will keep track here: https://introspector.oss-fuzz.com/project-profile?project=monero) or because of potential false positives.

tests/fuzz/CMakeLists.txt
```diff
@@ -26,6 +26,67 @@
...
 
+# Retrieve FuzzedDataProvider
+set(FDP_ROOT ${CMAKE_BINARY_DIR}/_deps/fuzzed_headers)
+set(FDP_DIR ${FDP_ROOT}/fuzzer)
+set(FDP_PATH ${FDP_DIR}/FuzzedDataProvider.h)
+file(MAKE_DIRECTORY ${FDP_DIR})
+if(NOT EXISTS ${FDP_PATH})
+  message(STATUS "Downloading FuzzedDataProvider.h to ${FDP_PATH}")
+  file(DOWNLOAD
+    https://raw.githubusercontent.com/llvm/llvm-project/main/compiler-rt/include/fuzzer/FuzzedDataProvider.h
```
SChernykh: Why do you download this file during CMake step, instead of just having it in the source tree?

arthurscchan: To ensure the newest version of FuzzedDataProvider from llvm is used.

SChernykh: That file is barely updated once a year, judging by its history. If you download the newest version, one day you might get a compilation error because of some incompatibility with the new version and you will have to fix the code anyway.

SChernykh: Also, this approach breaks reproducible builds. I know it's a code for tests only, but it's better not to do like this.

arthurscchan: Thanks for your suggestions. I will prepare a static FuzzedDataProvider.h in the tests/fuzz/include/fuzzer
 directory and refer it in the CMakeList.txt for the test build.

nahuhh:
That latest push was bad

looks like you rebase onto an old master or smthn. The versions in readme were reverted to old versions

`git pull --rebase <nameOfUpstreamRemote> master` should fix it

tests/fuzz/fuzz_rpc/fuzz_rpc.cpp
```diff
@@ -0,0 +1,153 @@
...
+    tools::performance_timers->clear();
+    delete tools::performance_timers;
+    tools::performance_timers = nullptr;
+  }
+
+  siglongjmp(jump_env, 1);
+}
+
+// Helper function to generate a biased selector targeting rarely-fuzzed RPC endpoints
+int biased_selector(FuzzedDataProvider& provider) {
```
vtnerd: Why randomly select an endpoint? Don't quite understand this, a simple loop through would be better, no?

tests/fuzz/fuzz_rpc/zmq_endpoints.cpp
```diff
@@ -0,0 +1,87 @@
...
+    evt.weight = provider.ConsumeIntegral<uint64_t>();
+
+    events.push_back(std::move(evt));
+  }
+
+  pub.send_txpool_add(std::move(events));
+}
+
+// Map all zmq targets
+std::map<int, std::function<void(zmq_pub&, FuzzedDataProvider&)>> zmq_targets = {
```
vtnerd, line 82: This is fuzzing the ZMQ_PUB instead of ZMQ_REP, which is probably higher priority. Technically `monerod` does do some processing for ZMQ_PUB (because it uses XPUB, etc), so this isn't completely useless, but the processing is very limited, etc.

tests/fuzz/fuzz_rpc/initialisation.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+#include "crypto/hash.h"
+#include "rpc/core_rpc_server.h"
+#include <fuzzer/FuzzedDataProvider.h>
+
+// Define templates for dummy protocol object
+template class nodetool::node_server<cryptonote::t_cryptonote_protocol_handler<cryptonote::core>>;
+
+struct DummyProtocol : public cryptonote::i_cryptonote_protocol {
```
vtnerd, line 9: We are a little inconsistent, but most of the code uses snake case instead of camel case. I guess since this is going into fuzz, this matters less.

tests/fuzz/fuzz_rpc/rpc_endpoints.cpp
```diff
@@ -0,0 +1,936 @@
+#include "rpc_endpoints.h"
+#include "initialisation.h"
+#include "rpc/rpc_payment_signature.h"
+#include <cstring>
+#include <map>
+
+// Initialising common objects for calling RPC endpoint functions
+cryptonote::core_rpc_server::connection_context ctx;
```
vtnerd, line 8: I'm going todo a second pass to make sure this is all legit in here later, wanted todo a general code review first.

tests/fuzz/include/fuzzer/FuzzedDataProvider.h
```diff
@@ -0,0 +1,397 @@
...
+  const uint8_t *data_ptr_;
+  size_t remaining_bytes_;
+};
+
+// Returns a std::vector containing |num_bytes| of input data. If fewer than
+// |num_bytes| of data remain, returns a shorter std::vector containing all
+// of the data that's left. Can be used with any byte sized type, such as
+// char, unsigned char, uint8_t, etc.
+template <typename T>
+std::vector<T> FuzzedDataProvider::ConsumeBytes(size_t num_bytes) {
```
vtnerd, line 109: This function is only used by `ConsumeBytesWithTerminator` and that function doesn't appear to be used anywhere. Should they both be removed? Did I miss something?

tests/fuzz/include/fuzzer/FuzzedDataProvider.h
```diff
@@ -0,0 +1,397 @@
...
+                                                              T terminator) {
+  num_bytes = std::min(num_bytes, remaining_bytes_);
+  std::vector<T> result = ConsumeBytes<T>(num_bytes + 1, num_bytes);
+  result.back() = terminator;
+  return result;
+}
+
+// Returns a std::vector containing all remaining bytes of the input data.
+template <typename T>
+std::vector<T> FuzzedDataProvider::ConsumeRemainingBytes() {
```
vtnerd, line 129: This function also does not appear to be used.

tests/fuzz/include/fuzzer/FuzzedDataProvider.h
```diff
@@ -0,0 +1,397 @@
...
+#include <utility>
+#include <vector>
+
+// In addition to the comments below, the API is also briefly documented at
+// https://github.com/google/fuzzing/blob/master/docs/split-inputs.md#fuzzed-data-provider
+class FuzzedDataProvider {
+ public:
+  // |data| is an array of length |size| that the FuzzedDataProvider wraps to
+  // provide more granular access. |data| must outlive the FuzzedDataProvider.
+  FuzzedDataProvider(const uint8_t *data, size_t size)
```
vtnerd, line 36: It would be preferably if `data` could be moved/copied into this class (a `std::string` or `epee::byte_slice`). Is that possible?

We also use `epee::span` for borrowed data, which conveniently has an `advance` function already.

tests/fuzz/include/fuzzer/FuzzedDataProvider.h
```diff
@@ -0,0 +1,397 @@
...
+  // logic, a fuzzer can insert characters into the string, and the string
+  // will be lengthened to include those new characters, resulting in a more
+  // stable fuzzer than picking the length of a string independently from
+  // picking its contents.
+  std::string result;
+
+  // Reserve the anticipated capacity to prevent several reallocations.
+  result.reserve(std::min(max_length, remaining_bytes_));
+  for (size_t i = 0; i < max_length && remaining_bytes_ != 0; ++i) {
+    char next = ConvertUnsignedToSigned<char>(data_ptr_[0]);
```
vtnerd, line 165: This is just a general comment, why not use our internal random generator when needed instead of this one-time initialized blob (that `FuzzedDataProvider` is wrapping)? What is this other random generator doing that is better than our internal one?

I _think_ it's so that you can recover the initial seed value, so the exact test can be repeated. Is that the case? Seems like you also also pass a 256-bit seed and then use a 256-bit hashing mechanism, to just pop-out random values instead of this nebulous blob of bytes.

tests/fuzz/include/fuzzer/FuzzedDataProvider.h
```diff
@@ -0,0 +1,397 @@
...
+  if (std::numeric_limits<TS>::is_modulo)
+    return static_cast<TS>(value);
+
+  // Avoid using implementation-defined unsigned to signed conversions.
+  // To learn more, see https://stackoverflow.com/questions/13150449.
+  if (value <= std::numeric_limits<TS>::max()) {
+    return static_cast<TS>(value);
+  } else {
+    constexpr auto TS_min = std::numeric_limits<TS>::min();
+    return TS_min + static_cast<TS>(value - TS_min);
```
vtnerd, line 393: This appears to be attempting signed integer underflow, which is undefined by C++. There has to be boost function or something for this kind of thing.

vtnerd (commented):
I don't see how the fuzzer hooks into our system. Where is `FuzzedDataProvider` created and used?

tests/fuzz/fuzz_rpc/fuzz_rpc.cpp
vtnerd: I think I figured it out. Each endpoint could alter state. So you need to randomly select endpoints to get a better fuzz.

DavidKorczynski: The main point here is to ensure we fuzz all endpoints. This is perhaps mostly relevant when doing local runs (1-90 hour runs) rather than oss-fuzz (thousands of hours of fuzzing).

tests/fuzz/include/fuzzer/FuzzedDataProvider.h
DavidKorczynski: The `FuzzedDataProvider.h` is actually part of the LLVM Framework and libFuzzer. We added this purely because of passing the CI tests here. The fuzzers are meant to run by way of OSS-Fuzz (https://github.com/google/oss-fuzz/tree/master/projects/monero) and `FuzzedDataProvider.h` is already provided by the libraries/compiler in their build environment (this Dockerfile https://github.com/google/oss-fuzz/blob/master/projects/monero/Dockerfile). As such, I don't think we should adjust the `FuzzedDataProvider.h` simply because it's just LLVM code we have here to support the CI.

For what it's worth, we don't need this file in this manner when building/running in the OSS-Fuzz environment (it's only for satisfying Monero's CI). Originally we downloaded it dynamically (https://github.com/monero-project/monero/pull/10004#discussion_r2228501179).

Am not sure what the preference is here? We could remove it and try to avoid building the fuzzer in the CI -- that would probably makes things a bit cleaner. Am not sure if this fuzzer will ever be built outside the OSS-Fuzz environment.

tests/fuzz/include/fuzzer/FuzzedDataProvider.h
DavidKorczynski: see comment https://github.com/monero-project/monero/pull/10004#discussion_r2277838664

tests/fuzz/include/fuzzer/FuzzedDataProvider.h
DavidKorczynski: see comment https://github.com/monero-project/monero/pull/10004#discussion_r2277838664

tests/fuzz/include/fuzzer/FuzzedDataProvider.h
DavidKorczynski: See https://github.com/monero-project/monero/pull/10004#discussion_r2277838664

We should avoid adjusting this code. Notice the data here is controlled by libFuzzer/AFL/honggfuzz under the OSS-Fuzz environment. We should avoid any random data generation in the fuzz harness and let the fuzzing engines control that.

tests/fuzz/include/fuzzer/FuzzedDataProvider.h
DavidKorczynski: See comment https://github.com/monero-project/monero/pull/10004#discussion_r2277838664

tests/fuzz/fuzz_rpc/zmq_endpoints.cpp
DavidKorczynski: We simply added this just to increase coverage as such.

DavidKorczynski:
> Looks like some extern C functions get called from the fuzz harness?

Yeah -- this is written in libFuzzer style: https://llvm.org/docs/LibFuzzer.html#fuzz-target `extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {` is called by the fuzzing engines (libFuzzer, AFL++, honggfuzz) which are used by OSS-Fuzz (Monero's project: https://github.com/google/oss-fuzz/tree/master/projects/monero).

The existing Monero fuzzers have this function as well, but it's hidden behind some macros: https://github.com/monero-project/monero/blob/389e3ba1df4a6df4c8f9d116aa239d4c00f5bc78/tests/fuzz/fuzzer.h#L52

tests/fuzz/fuzz_rpc/initialisation.cpp
```diff
@@ -0,0 +1,355 @@
...
+        }
+
+        crypto::view_tag vtag{};
+        vtag.data = static_cast<uint8_t>(provider.ConsumeIntegral<uint8_t>());
+
+        cryptonote::txout_to_tagged_key tagged{};
+        tagged.key = pk;
+        tagged.view_tag = vtag;
+
+        oe.second.dest = *reinterpret_cast<const rct::key*>(&tagged.key);
```
vtnerd, line 277: I should've commented on this before, there's a `rct::pk2rct` that also violates strict aliasing rules, but is "our" standard way of converting it.

DavidKorczynski:
I don't think the failing CI is due to anything from this PR. When I look at other PRs I see the same jobs failing spuriously.

ACK-J (approved):
I've successfully run the fuzzing harnesses for multiple days

jeffro256:
Can I assume that this replaces https://github.com/monero-project/monero/pull/9463?

DavidKorczynski:
> Can I assume that this replaces #9463?

I don't think so, they look separate to me.

This PR is focused on building/running these as part of OSS-Fuzz, where, `OSSFUZZ=ON` is set https://github.com/google/oss-fuzz/blob/77953d2708b883bf1edbbcf9548a82fd7251a44e/projects/monero/build.sh#L28 and `make -C tests/fuzz ...` is used: https://github.com/google/oss-fuzz/blob/77953d2708b883bf1edbbcf9548a82fd7251a44e/projects/monero/build.sh#L53

We could look to address https://github.com/monero-project/monero/pull/9463 in a follow up PR, but am not sure what the status is of that PR fwiw.

selsta:
Please squash.

arthurscchan:
Done

tests/fuzz/fuzz_rpc/README.md
```diff
@@ -0,0 +1,55 @@
...
+If blocks are successfully added, a list of random transactions is then generated and also pushed to the blockchain.
+This step is skipped for `fuzz_zmq`.
+
+## Real Fuzzing
+
+The selected RPC endpoint helper function in `rpc_endpoints` or `zmq_endpoints` generates a random request and response object for the specific call, and then invokes the corresponding handler in the `core_rpc_server` object.
+
+## Timeout Handling
+
+We use SIGALRM to catch and rewind timeout events in order to prevent the process from hanging, allowing the fuzzing to continue to the next iteration.
```
selsta: Is this still accurate after the `drop SIGALRM` commit?

selsta (commented):
There are a couple typos, we usually get PRs to fix typos but might make sense to fix them now before merge.

- fuuzz -> fuzz
- Slient -> Silent
- Rnadom -> Random
- messages_to_sent -> messages_to_send

src/common/perf_timer.cpp
```diff
@@ -90,7 +90,11 @@ namespace tools
 
 el::Level performance_timer_log_level = el::Level::Info;
 
-static __thread std::vector<LoggingPerformanceTimer*> *performance_timers = NULL;
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+  __thread std::vector<LoggingPerformanceTimer*> *performance_timers = NULL;
```
selsta, line 95:
```suggestion
  #warning "Building with fuzzing mode UNSAFE FOR PRODUCTION!"
  __thread std::vector<LoggingPerformanceTimer*> *performance_timers = NULL;
```

Would it make sense to add this here? I worry that if we change something in CMake in the future this accidentally could get enabled without us noticing.

PR #10008, opened 2025-07-24
functional_tests: check transfers list consistency in cold_signing

jeffro256:
Helped me catch some bad behavior while developing for Carrot/FCMP++ hot/cold wallets.

This PR is part of upstreaming Carrot/FCMP++.

Depends on #10007.

### DIFF
--- a/tests/functional_tests/cold_signing.py
+++ b/tests/functional_tests/cold_signing.py
@@ -131,8 +131,31 @@ class ColdSigningTest():
             res = self.hot_wallet.export_outputs()
             self.cold_wallet.import_outputs(res.outputs_data_hex)
 
+        # Check consistency between hot & cold transfers list, especially that one-time addresses and amounts line up
+        def check_transfers_list_consistency(do_check_key_images):
+            hot_transfers_list = self.hot_wallet.incoming_transfers()
+            cold_transfers_list = self.cold_wallet.incoming_transfers()
+            if 'transfers' not in hot_transfers_list:
+                assert 'transfers' not in cold_transfers_list
+                return
+            hot_transfers_list = hot_transfers_list['transfers']
+            cold_transfers_list = cold_transfers_list['transfers']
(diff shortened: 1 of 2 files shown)


tests/functional_tests/cold_signing.py
```diff
@@ -131,9 +131,32 @@ def export_import(self, piecemeal_output_export):
...
+                return
+            hot_transfers_list = hot_transfers_list['transfers']
+            cold_transfers_list = cold_transfers_list['transfers']
+            assert len(hot_transfers_list) == len(cold_transfers_list)
+            attributes_to_cmp = ['amount', 'subaddr_index', 'pubkey']
+            if do_check_key_images:
+                attributes_to_cmp.append("key_image")
+            for i in range(len(hot_transfers_list)):
+                for attr in attributes_to_cmp:
+                    hot_val = hot_transfers_list[i][attr]
```
0xFFFC0000, line 149: I am not 100% sure about this, but keep this in mind whether the order matters. If order matters and the wallet RPC ever changes the ordering of transfers (e.g. sorted by block height vs. by pubkey), the test will spuriously fail even though contents match.

0xFFFC0000 (approved):

0xFFFC0000:
I approved. But take a quick look at this to make sure: https://github.com/monero-project/monero/pull/10008#discussion_r2254629655

tests/functional_tests/cold_signing.py
jeffro256: I'm fairly certain that, even if it was not originally designed to be a core invariant of `wallet2`, that the order of the transfers list is pretty much set in stone as the on-chain order of outputs at this point. If this invariant were to no longer hold, multisig wallets, hot/cold wallets, and burning bug protection would all break. There's probably more, but that's just from the top of my head.

0xFFFC0000: Fair enough

PR #10012, opened 2025-07-24
daemon: remove miniupnp

tobtoht:
Previous discussion:

- https://github.com/monero-project/monero/pull/9367
- https://libera.monerologs.net/monero-dev/20250724#c544331
- https://libera.monerologs.net/monero-dev/20260425#c670316-c670322

Open question is whether we want to replace this with PCP / NAT-PMP first. Bitcoin Core uses a bespoke [implementation](https://github.com/bitcoin/bitcoin/pull/30043) (disabled by default) and previously used [libnatpmp](https://github.com/miniupnp/libnatpmp), another C library maintained by the miniupnpc dev.

~Passing `--no-igd` to `monerod` will now result in an error. We may want to make it a no-op instead.~ It will now warn.


nahuhh:
> Open question is whether we want to replace this with PCP / NAT-PMP first. Bitcoin Core uses a bespoke [implementation](https://github.com/bitcoin/bitcoin/pull/30043) (disabled by default)

I think so. For compatibility / easy upgrades, i think we should, like our miniupnp impl, have it "delayed" by default + same options (delayed|disabled|enabled), and..

> Passing `--no-igd` to `monerod` will now result in an error. We may want to make it a no-op instead.

--no-igd should currently return a log that the flag isnt functional, but if/when adding PCP, it should work to disable PCP

my 2c

in any case, upnp has never worked for me, and so far doesnt appear to work on most moneronodos (hardware nodes). +1 on removing miniupnp, but dont crash nodes that autoupdate and have `--igd` or `--no-igd` flags set

edit: for reference, here is the PR where Bitcoin removed miniupnp: https://github.com/bitcoin/bitcoin/pull/31130

selsta (approved):

nahuhh:
```
.github/workflows/build.yml:18:  APT_INSTALL_LINUX: 'apt -y install build-essential cmake libboost-all-dev miniupnpc libunbound-dev graphviz doxygen libunwind8-dev pkg-config libssl-dev libzmq3-dev libsodium-dev libhidapi-dev libusb-1.0-0-dev libprotobuf-dev protobuf-compiler ccache curl git'
.github/workflows/build.yml:56:          brew install --quiet cmake boost hidapi openssl zmq miniupnpc expat libunwind-headers protobuf ccache
.gitignore:23:external/miniupnpc/Makefile
.gitignore:24:miniupnpcstrings.h
```

Should these be nuked as well?

edit:  a couple more strays of igd, example in anonymity doc

```
docs/ANONYMITY_NETWORKS.md:52:monerod --proxy 127.0.0.1:9050 --p2p-bind-ip 127.0.0.1 --no-igd
docs/ANONYMITY_NETWORKS.md:216:monerod --proxy 127.0.0.1:9050 --p2p-bind-ip 127.0.0.1 --no-igd
docs/ANONYMITY_NETWORKS.md:228:    --no-igd \
docs/ANONYMITY_NETWORKS.md:240:    --no-igd \
```

selsta (approved):

nahuhh (approved):

PR #10033, opened 2025-08-11
wallet: fix warning in fee priority algorithm

0xFFFC0000:
For small helper functions like these that live in a header, the modern C++ solution is to replace static with inline. This signals to the compiler that the function is defined in the header and avoids the "unused function" warning, as it creates a single definition for the entire program.

[Example](https://github.com/monero-project/monero/actions/runs/16183764676/job/45685482172?pr=9838#step:10:1760)

### DIFF
--- a/src/wallet/fee_algorithm.h
+++ b/src/wallet/fee_algorithm.h
@@ -13,7 +13,7 @@ namespace tools
 
     namespace fee_algorithm_utilities
     {
-        static int as_integral(const fee_algorithm algorithm)
+        inline int as_integral(const fee_algorithm algorithm)
         {
             return static_cast<int>(algorithm);
         }
--- a/src/wallet/fee_priority.h
+++ b/src/wallet/fee_priority.h
@@ -31,7 +31,7 @@ namespace tools
         inline constexpr EnumStringsType fee_priority_strings = { { "default", "unimportant", "normal", "elevated", "priority" } };
         inline constexpr EnumsType enums = { { fee_priority::Default, fee_priority::Unimportant, fee_priority::Normal, fee_priority::Elevated, fee_priority::Priority } };
 
-        static fee_priority decrease(const fee_priority priority)
+        inline fee_priority decrease(const fee_priority priority)
         {
             if (priority == fee_priority::Default)
             {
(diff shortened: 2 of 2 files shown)


0xFFFC0000:
> RTFM!!!!!!

https://letmegooglethat.com/?q=monero+contribution+guide+and+Patch+requirements can be useful for you.

Tzadiko:
I use these methods in https://github.com/monero-project/monero/pull/9847, which I have yet to get to. These unused function warnings were introduced in Part 1 of a change-set that will be use these functions later (these warnings will disappear). So I don't think making this change is warranted.

vtnerd (approved):
Crap, I missed during review that these were shifted from a `class` scope to `namespace` scope. These changes are definitely useful.

vtnerd:
> I use these methods in https://github.com/monero-project/monero/pull/9847, which I have yet to get to. These unused function warnings were introduced in Part 1 of a change-set that will be use these functions later (these warnings will disappear). So I don't think making this change is warranted.

`static` here creates a unique instance per compilation unit, whereas `inline` merges all instances in the linker. We want `static` if in `class` scope, and `inline` if in `namespace` scope.

I'll look at your other review if you still feel these changes are unnecessary.

EDIT: I'll look at your review either way, I just meant that I'll at that other review and possibly rescind my approval here if needed.

Tzadiko:
> > I use these methods in #9847, which I have yet to get to. These unused function warnings were introduced in Part 1 of a change-set that will be use these functions later (these warnings will disappear). So I don't think making this change is warranted.
> 
> `static` here creates a unique instance per compilation unit, whereas `inline` merges all instances in the linker. We want `static` if in `class` scope, and `inline` if in `namespace` scope.
> 
> I'll look at your other review if you still feel these changes are unnecessary.
> 
> EDIT: I'll look at your review either way, I just meant that I'll at that other review and possibly rescind my approval here if needed.

Thanks for chiming in. I get the distinction, but Im not sure what this change 'buys' us. Both are functionally the same, it's just the linkage changing from internal to external. If you were to summarize why this change should go through (I'm indifferent), given that the code is funcitonally the same, what would the reason be?

selsta:
Current code produces a compiler warning.

jeffro256 (approved):

jeffro256:
> Both are functionally the same, it's just the linkage changing from internal to external.

Theoretically, using `inline` results in smaller binary sizes because the function bodies aren't necessarily duplicated across translation units. In practice, it's simply good form (more readable) to use `inline` instead of `static` when the definition is shared across many translations units.

PR #10034, opened 2025-08-12
crypto: fixes Pointer overflow check

ptrgits:
https://github.com/monero-project/monero/blob/389e3ba1df4a6df4c8f9d116aa239d4c00f5bc78/src/crypto/crypto.cpp#L766-L766

When checking for integer overflow, you may often write tests like p + i < p. This works fine if p and i are unsigned integers, since any overflow in the addition will cause the value to simply "wrap around." However, using this pattern when p is a pointer is problematic because pointer overflow has undefined behavior according to the C and C++ standards. If the addition overflows and has an undefined result, the comparison will likewise be undefined; it may produce an unintended result, or may be deleted entirely by an optimizing compiler.

---

fix this problem, we should avoid pointer arithmetic that could result in undefined behavior. Instead of comparing pointers directly, we should compare the offset (as an integer) between `end` and the start of the buffer to the size of the buffer. Specifically, we can compute the difference as `end - reinterpret_cast<char *>(&buf)` and ensure that this value is less than or equal to the size of the buffer (`sizeof buf`). This approach uses only valid pointer arithmetic (since both pointers are within the same object), and the result is a `ptrdiff_t` or `size_t` value that can be safely compared to the buffer size. The change should be made in the function `crypto_ops::derive_view_tag` in `src/crypto/crypto.cpp`, specifically replacing the assertion on line 766.

### References
Embedded in Academia: [Pointer Overflow Checking](https://blog.regehr.org/archives/1395)
LWN: [GCC and pointer overflows](https://lwn.net/Articles/278137/)

### DIFF
--- a/src/crypto/crypto.cpp
+++ b/src/crypto/crypto.cpp
@@ -763,7 +763,7 @@ POP_WARNINGS
     memcpy(buf.salt, "view_tag", 8); // leave off null terminator
     buf.derivation = derivation;
     tools::write_varint(end, output_index);
-    assert(end <= buf.output_index + sizeof buf.output_index);
+    assert(static_cast<size_t>(end - reinterpret_cast<char *>(&buf.output_index)) <= sizeof buf.output_index);
 
     // view_tag_full = H[salt|derivation|output_index]
     hash view_tag_full;


jeffro256:
From section 5.7.5 of the C++11 standard:

> When an expression that has integral type is added to or subtracted from a pointer, the result has the type of the pointer operand. If the pointer operand points to an element of an array object, and the array is large enough, the result points to an element offset from the original element such that the difference of the subscripts of the resulting and original array elements equals the integral expression. In other words, if the expression P points to the i-th element of an array object, the expressions (P)+N (equivalently, N+(P)) and (P)-N (where N has the value n) point to, respectively, the i + n-th and i - n-th elements of the array object, provided they exist. Moreover, if the expression P points to the last element of an array object, the expression (P)+1 points one past the last element of the array object, and if the expression Q points one past the last element of an array object, the expression (Q)-1 points to the last element of the array object. **If both the pointer operand and the result point to elements of the same array object, or one past the last element of the array object, the evaluation shall not produce an overflow;** otherwise, the behavior is undefined.

Source: https://port70.net/%7Ensz/c/c%2B%2B/c%2B%2B11_n3337.txt

Since the expression `buf.output_index + sizeof buf.output_index` points to the element exactly one element after the array `buf.output_index`, the result of the addition expression is well-defined. However, the nature of the check against `end` is always going to be undefined if it fails, this PR doesn't make it well-defined. If `end > buf.output_index + sizeof buf.output_index`, then that means that some previous addition before this point in the code had undefined behavior, which is why the `assert` statement is here. Changing the arithmetic of the `assert` doesn't change the fact that UB happened earlier. 

NACK unless you can show a platform where this statement superfluously fails.

Issue #8827, opened 2023-04-22
[Proposal] Improve PoW verification at the next hard fork


tevador:
I'm proposing a small change to the PoW algorithm to be implemented at the next hard fork (together with Seraphis).

As opposed to the previous PoW changes, this one is completely unrelated to ASIC resistance.

The purpose of the change is to enable a very fast partial verification of the PoW. Essentially, my proposal is to replace the final Blake2b hash with a double Blake2b hash. Technical details are in this issue: https://github.com/tevador/RandomX/issues/258

Additionally, a new 32-byte field (the intermediate hash) would be added to the block header. This would increase the blockchain size by about 8 MB per year, which is negligible given the large benefits that this change would bring.

What are the benefits?

* Enhanced denial-of-service (DoS) protection of all nodes in the network. PoW verification typically consumes about 15 milliseconds of CPU time and can become a DoS vector if a malicious actor started flooding the network with invalid blocks. My proposal allows for pre-verification that takes only 300 nanoseconds and makes such DoS attacks ineffective.

* Significant performance boost for [light nodes](https://github.com/monero-project/research-lab/issues/69) that rely solely on PoW verification. Light nodes could only verify the PoW partially and do full RandomX validation only for the top N blocks. This would reduce their CPU demands by up to 4 orders of magnitude.

* Better security for wallets using untrusted remote nodes. Malicious remote nodes can feed wallets fake blockchain data. With this proposal, wallets could partially verify the integrity of the blocks received from untrusted remote nodes with the cost of a few hashes.

hyc:
Sounds like a big win. Aside from the 8MB size overhead, what are any other potential downsides?

tevador:
The blockchain grows at around 40 GB/year at the moment, so the space overhead of this change would be about 0.02%.

The additional hash calculation would reduce hashrate by about 0.02%, which is also negligible and I don't think miners would object (or even notice).

AFAIK there has never been a change of the block header format before, so this would be the first time, but I don't think it would cause problems. And Seraphis already completely changes the block format anyways.

We would need to support both versions of RandomX, but the RandomX repository can easily support that as there are no changes to the internals, so "v2" can be realized by adding 1 function to the public API.

I can't think of any major drawbacks.

SChernykh:
What prevents faking the intermediate hash? An attacker can just run Blake2b miner to produce valid intermediate hashes.

tevador:
[...]

PR #10038, opened 2025-08-14
cryptonote_core: support RandomX V2 and commitments in v17

jeffro256:
See docs/BLOCK_HASHING.md, https://github.com/monero-project/monero/issues/8827, https://github.com/tevador/RandomX/pull/265, https://github.com/tevador/RandomX/pull/317.

TODO:
- [x] Change consensus definition of block PoW hash
- [ ] Tests for `MONERO_RANDOMX_UMASK` interfering with completeness
- [ ] Convergence tests

~~Store intermediate hashes in blockchain DB~~
~~Serve intermediate hashes in p2p messages while syncing / notifying of new blocks~~
~~Verify intermediate PoW before starting full block verification~~
~~Notify downstream mining software maintainers (namely xmrig)~~
~~Update hard fork table for activation~~

Discussion about overall changes welcome.

RandomX bump requires GCC 14 bump on RISC-V platforms.


docs/BLOCK_HASHING.md
```diff
@@ -0,0 +1,52 @@
+# Block hashing
+
+## Background
+
+In Bitcoin, block IDs and Proof-of-Work (PoW) hashes are calculated the same way, which is why on-chain block IDs always start
+with many 0s. However, in Monero, block IDs and PoW hashes are calcuated using different hash functions and slightly different
```
SChernykh: Typo, `calcuated`

docs/BLOCK_HASHING.md
```diff
@@ -0,0 +1,52 @@
...
+transactions in the block from the block hashing blob. This is the data flow for how block hashing works with these changes:
+
+![RandomX commitment data flow](resources/randomx-commitment-data-flow.png)
+
+The way that the hashes are calculated allows us to do partial PoW verification like so:
+
+![Partial RandomX verification data flow](resources/partial-pow-verify-data-flow.png)
+
+A node can be given simply the block header, block content hash, and intermediate PoW hash, and verify that some PoW was done
+using only `randomx_calculate_commitment()` (Blake2B undernath). Passing this partial verification requires PoW to done using
```
SChernykh: Typo `undernath`
PoW to done -> PoW to be done

src/crypto/rx-slow-hash.c
```diff
@@ -237,17 +244,22 @@ static void rx_alloc_cache(randomx_flags flags, randomx_cache** cache)
...
 }
 
-static void rx_init_full_vm(randomx_flags flags, randomx_vm** vm)
+static void rx_init_full_vm(const int variant, randomx_vm** vm, int* vm_variant)
 {
-  if (*vm || !main_dataset || (disabled_flags() & RANDOMX_FLAG_FULL_MEM)) {
+  if (!main_dataset || (disabled_flags() & RANDOMX_FLAG_FULL_MEM)) {
+    return;
+  }
+  else if (*vm && variant == *vm_variant) {
```
SChernykh, line 281: `else` is not needed here because previous `if` has `return` anyway.

src/crypto/rx-slow-hash.c
```diff
@@ -237,17 +244,22 @@ static void rx_alloc_cache(randomx_flags flags, randomx_cache** cache)
...
     return;
   }
```
SChernykh, line 283: if `vm` is allocated (not null), it should be released here - `randomx_destroy_vm(vm);`

src/crypto/rx-slow-hash.c
```diff
@@ -237,17 +244,22 @@ static void rx_alloc_cache(randomx_flags flags, randomx_cache** cache)
...
 
+  randomx_flags flags = enabled_flags();
   if ((flags & RANDOMX_FLAG_JIT) && !miner_thread) {
     flags |= RANDOMX_FLAG_SECURE;
   }
```
SChernykh, line 297: You forgot to set `RANDOMX_FLAG_V2` if variant is 2?

src/crypto/rx-slow-hash.c
```diff
@@ -261,20 +273,22 @@ static void rx_init_full_vm(randomx_flags flags, randomx_vm** vm)
...
     randomx_vm_set_cache(*vm, cache);
     return;
   }
 
+  randomx_flags flags = enabled_flags();
   if ((flags & RANDOMX_FLAG_JIT) && !miner_thread) {
     flags |= RANDOMX_FLAG_SECURE;
   }
 
   flags &= ~RANDOMX_FLAG_FULL_MEM;
```
SChernykh: Same two issues - previous `vm` is not destroyed, `RANDOMX_FLAG_V2` is not set.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -1438,7 +1438,8 @@ namespace cryptonote
     blobdata blob = t_serializable_object_to_blob(static_cast<block_header>(b));
     crypto::hash tree_root_hash = get_tx_tree_hash(b);
     blob.append(reinterpret_cast<const char*>(&tree_root_hash), sizeof(tree_root_hash));
-    blob.append(tools::get_varint_data(b.tx_hashes.size()+1));
+    if (b.major_version < HF_VERSION_POW_COMMITMENT)
```
SChernykh, line 1589: Why do you stop including the number of transactions into the hashing blob? This field is used by XMRig, for display purposes.

More importantly, it's a piece of data that gets erased by including only the tree root hash. 

Less data that goes into hashing is worse than more data.

src/rpc/rpc_payment.cpp
```diff
@@ -237,7 +237,7 @@ namespace cryptonote
     if (block.major_version >= RX_BLOCK_VERSION)
     {
       const crypto::hash &seed_hash = is_current ? info.seed_hash : info.previous_seed_hash;
-      crypto::rx_slow_hash(seed_hash.data, hashing_blob.data(), hashing_blob.size(), hash.data);
+      crypto::rx_slow_hash(seed_hash.data, RX_VARIANT_1, hashing_blob.data(), hashing_blob.size(), hash.data);
```
SChernykh: Why fixed variant? We have `block` here, so we can set the variant depending on block's height.

SChernykh (changes requested):

src/crypto/rx-slow-hash.c
jeffro256: I like including the `else` because it conveys to the reader that the branches are mutually exclusive.

src/crypto/rx-slow-hash.c
jeffro256: Oh yeah oops, I refactored this code like 3 times to try to get it clean as possible, and I guess I forgot to update the literal most important part ;)

src/rpc/rpc_payment.cpp
jeffro256: I guess I was trying to keep backwards compatibility, but RPC payments are not used AFAIK, so maybe I should use latest.

jeffro256: Or should I use the same as consensus rules? Probably that

src/cryptonote_basic/cryptonote_format_utils.cpp
jeffro256: I removed it because it isn't necessary for consensus and bloats the hashing blob size. If we chose to implement header-only syncing, the varint per hashing blob would impose a non-trivial cost. It shouldn't affect security AFAIK because Keccak isn't susceptible to length extension attacks, but please correct me if wrong.

src/crypto/rx-slow-hash.c
SChernykh: It's not a bug, just a matter of code style. I guess it can remain unchanged.

src/rpc/rpc_payment.cpp
SChernykh: Yes, same as consensus rules.

src/cryptonote_basic/cryptonote_format_utils.cpp
SChernykh: Hashing blob is 75 bytes (76-77 with the tx count). It's not a big difference, and apart for using it to display tx count in XMRig, it's also useful for comparing different hashing blobs between different Monero pools. Things like "which pools use the same Monero node, how many different Monero nodes does a pool use, is it mining empty blocks or not, and so on".

SChernykh: Oh, and one more important thing - I'm pretty sure the "76 bytes minimum" is hard-coded in countless pool implementations in many different places, so it's better to not touch it.

jeffro256: > ... and apart for using it to display tx count in XMRig, it's also useful for comparing different hashing blobs between different Monero pools. Things like "which pools use the same Monero node, how many different Monero nodes does a pool use, is it mining empty blocks or not, and so on".

I don't think that the number of txs is a very good metric because it can be easily faked. The miner can determine that there's 25 txs in the tx hash list, but that doesn't mean that there's 25 *real* txs in the hash list; they could all be generated by the pool itself, and of course the fee goes back to the miner.

> Oh, and one more important thing - I'm pretty sure the "76 bytes minimum" is hard-coded in countless pool implementations in many different places, so it's better to not touch it.

Why can't they change it to a 75 byte minimum?

SChernykh: https://github.com/SChernykh/p2pool/blob/master/src/block_template.cpp#L1303
https://github.com/SChernykh/p2pool/blob/master/src/stratum_server.cpp#L140
https://github.com/xmrig/xmrig/blob/master/src/base/tools/cryptonote/BlockTemplate.h#L117
https://github.com/PowPool/xmrpool/blob/main/cnutil/cnutil.go#L11

These were easy to find and will be easy to fix, but how many constants like these are out there and will have to be fixed one by one after they fail?

As for fake/real txs, this varint is not about fake/real. It's about how many txs are in the block template which is an indicator of what the pool is doing. If it says 25, there are 25 transactions of _some_ kind in the block.

jeffro256: > If it says 25, there are 25 transactions of some kind in the block.

A) The block content hash could be completely random, and the number of txs completely made up. That block won't validate, but if you use this varint as a display value, it tricks users into thinking that they are mining transactions when they aren't.

B) The block could contain transactions which are locally known only to the block template creator, and are only revealed on a PoW hit. This block will validate and the fees returned back to the miner. The only loss for the adversary would be the marginal penalty difference from the increase in block weight, if applicable. If you use the varint as a display value, you are showing the user the real number of transactions they are mining, but it means nothing as a metric for censorship / DoS prevention.

In both of these scenarios, the number means nothing, and can actually be harmful because it instills a sense of false security if used in the way that you describe.

> but how many constants like these are out there and will have to be fixed one by one after they fail?

Surely these would be caught quickly though?

jeffro256:
Rebased and set RandomX commit to tag v2.0.1

src/cryptonote_basic/cryptonote_format_utils.cpp
SChernykh: I still don't agree with removing this field. It's been there since 2014 and this removal is perpendicular to the purpose of this PR. Block validation does verify that this field matches the actual number of transactions being mined. Whether they are known to other miners or not, doesn't matter.

SChernykh: Moreover, as soon as the block is mined, everyone can see whether it had unknown transactions or not. For example, DataHoarder uses this to monitor pools on https://blocks.p2pool.observer/pools

SeriousPassenger:
Does the upgrade to rx/2 require a hard fork? Will this happen overnight or do we have a plan to set a time for it?

nahuhh:
> Does the upgrade to rx/2 require a hard fork?

Yes

> Will this happen overnight 

No

src/cryptonote_basic/cryptonote_format_utils.cpp
jeffro256: > Block validation does verify that this field matches the actual number of transactions being mined.

> Moreover, as soon as the block is mined, everyone can see whether it had unknown transactions or not. 

Both of these happen with the *full* block info, not the block hashing blob, which is exactly my point: the varint is completely pointless without further context. With further context, it is also pointless because you don't need it to detect the number of transactions in the block. Either way, it will make header-only sync slightly more complicated because the DB code will need to store the number of transactions per block to serve block hashing blobs.

> Whether they are known to other miners or not, doesn't matter.

How can you say this? It *absolutely* matters because a dummy miner self-send tx is much less meaningful than a Monero transfer for real goods and services. Honest transaction constructors broadcast their transactions into the mempool. If the miners aren't mining transactions in the mempool when it is financially reasonable for them to do so, it is malicious. What is the point of showing the user the number of transactions if not to provide a metric for how of the mempool their hashrate is processing, and thus exposing these malicious actors? You brought up a pool monitor website which does this very thing. So what other purpose could the display value serve?

> It's been there since 2014

This point is subjective, but I don't think that we should keep in old broken features because they are old.

> removal is perpendicular to the purpose of this PR.

The PoW commitments and V2 are perpendicular purposes, but I thought it made sense to put them together, along with this tweak, because they all happen about in the same location in the code, and all of them change the PoW hash definition. And if I want to build on this PR for the purposes of DoS protection / header-only sync, the code needs to know whether the number of transactions is included in the block hashing blob.

SChernykh: > Either way, it will make header-only sync slightly more complicated because the DB code will need to store the number of transactions per block to serve block hashing blobs.

This is the only valid reason to remove it, but not valid enough. DB code needs to store the merkle hash of all block transactions, so it already processes all of them to create the block hashing blob. It shouldn't be harder than adding 1-2 lines of code to the header creation in DB code.

Motophan:
I am sorry if this is the wrong format to ask, but is there any idea when the hardfork will activate? I understand these things are not set in stone but I would like to know roughly when we can start mining rxv2. Is there a better place to see where we are at?

src/cryptonote_basic/cryptonote_format_utils.cpp
earnifysu: > It shouldn't affect security AFAIK because Keccak isn't susceptible to length extension attacks, but please correct me if wrong.

length extension is not the issue, keccak is fine there. but there is a concrete effect, and it is closer to what @SChernykh meant by "a piece of data that gets erased by including only the tree root hash".

`tree_hash` carries no length prefix at any level (`src/crypto/tree-hash.c:77-104`). so the tree root does not commit to the leaf count, and with the varint gone the hashing blob does not either.

concretely: for any leaf count `n >= 3`, collapsing exactly one adjacent leaf pair into its hash gives `n - 1` leaves with a byte identical root. the pair is not arbitrary, it is the one at index `2*tree_hash_cnt(n) - n`, which is precisely where the pairing loop at `tree-hash.c:91` starts. i compiled the in-tree `tree_hash` and enumerated it:

| leaves | collapsible pair index | result |
|---|---|---|
| 3 | 1 | -> 2, same root |
| 4 | 0 | -> 3, same root |
| 5 | 3 | -> 4, same root |
| 6 | 2 | -> 5, same root |
| 7 | 1 | -> 6, same root |
| 8 | 0 | -> 7, same root |
| 12 | 4 | -> 11, same root |

so for a v17 block `B` with `tx_hashes = {a, b}`, the block `B'` with the identical header, identical `miner_tx` and `tx_hashes = {H(a||b)}` has the same tree root, and since `get_block_hashing_blob` (`cryptonote_format_utils.cpp:1440-1448`) no longer appends the count, the same hashing blob, the same block id, and the same randomx hash and commitment.

**scope, to be clear about what this is not.** `B'` is not acceptable: it needs a real transaction whose txid is `H(a||b)`, and it shares `B`'s id so `have_block` (`blockchain.cpp:4530`) rejects it once `B` is on chain. i am not claiming an attack and i have not found a way to turn this into one. i also checked the in-tree consumers: `tree_branch` and `is_branch_in_tree` are referenced only by their declaration, their definition, and `tests/unit_tests/crypto.cpp`, so nothing in tree rebuilds a root from a branch and breaks.

what does change is an invariant. before v17 the block id pinned the number of leaves. after v17 it does not, and the hashing blob alone can no longer distinguish those two leaf vectors. that is a property external tooling could reasonably have been relying on, which seems to be the substance of the disagreement here rather than the display value question.

if the goal is a fixed size blob and the invariant is worth keeping, folding the count into the content hash rather than appending a varint to the blob gets both: hash the tree root together with the leaf count, and the blob stays at the three fields `docs/BLOCK_HASHING.md:47-52` describes, with no varint and no 76 vs 75 byte change for pools. that would also leave `tree_hash` itself untouched.

no opinion from me on whether the field should go. just answering the "correct me if wrong" directly, since the answer turned out to be neither length extension nor nothing.

src/crypto/rx-slow-hash.c
```diff
@@ -237,17 +266,28 @@ static void rx_alloc_cache(randomx_flags flags, randomx_cache** cache)
...
+    rx_destroy_vm(vm);
+  }
 
+  randomx_flags flags = enabled_flags();
   if ((flags & RANDOMX_FLAG_JIT) && !miner_thread) {
     flags |= RANDOMX_FLAG_SECURE;
   }
+  flags |= get_variant_flags(variant);
 
   *vm = randomx_create_vm((flags | RANDOMX_FLAG_LARGE_PAGES | RANDOMX_FLAG_FULL_MEM) & ~disabled_flags(), NULL, main_dataset);
```
selsta, line 289: `MONERO_RANDOMX_UMASK` is applied after `RANDOMX_FLAG_V2` is added, so a mask containing bit 128 silently creates a v1 VM while `vm_variant` is recorded as v2. At HF17 this produces the wrong PoW. The path below has the same issue.

src/cryptonote_basic/cryptonote_format_utils.cpp
ACK-J: The removal may lead to hash collisions https://github.com/xmrack/monero-review/issues/205

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -1442,7 +1442,8 @@ namespace cryptonote
...
+      blob.append(tools::get_varint_data(b.tx_hashes.size()+1));
```
SChernykh, line 1446: @jeffro256 Removing the number of transactions varint creates a legit vulnerability - I read through https://github.com/xmrack/monero-review/issues/205 and it does create a race where attacker's block with a fake transaction ID, which produces exactly the same hashing blob as a real block, can be received first and then the real block will be rejected.

SChernykh (changes requested):

src/cryptonote_basic/cryptonote_format_utils.cpp
jeffro256: TBF this would only be an issue if the daemon never re-computed the TXIDs from TX blobs, but yes it is an issue. But the ... root (🥁🤡) ...  of the issue is that our tree hashing code is *weak*. This still isn't a good argument for putting the number of txs inside the header hashing blob specifically, but the tree should bind the number, or at least the structure.

jeffro256:
Rebased

src/crypto/rx-slow-hash.c
jeffro256: Skill issue....

JK I think best course of action in this scenario would be to override the user's flags, and log a warning. Sounds good?

jeffro256: Eh, I'd say log a warning once on fetch if it overlaps with the random v2 bit.

PR #10039, opened 2025-08-15
cryptonote_basic: add overload for get_block_longhash()

jeffro256:
Moves the core logic for `get_block_longhash()` into `cryptonote_basic` library from `cryptonote_core`. No change in behavior should occur, it's purely a refactoring. 

Depends on #9919

### DIFF
--- a/src/blockchain_db/CMakeLists.txt
+++ b/src/blockchain_db/CMakeLists.txt
@@ -45,7 +45,8 @@ target_link_libraries(blockchain_db
   PUBLIC
     common
     cncrypto
-    ringct
+    cryptonote_basic
+    ringct_basic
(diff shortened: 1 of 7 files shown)


vtnerd (commented):
These changes look acceptable - but why make the change? Is this related to fcmp++ in some way?

jeffro256:
I intend for `get_block_longhash()` to be usable in `BlockchainLMDB` to perform database migrations for if we want to integrate #10038. If we don't, it's also just cleaner IMO.

jeffro256:
Rebased to remove #9919 from commit list

selsta:
@vtnerd please re-review

Issue #10045, opened 2025-08-19
BlockchainLMDB misalignment issues


jeffro256:
Bad [memory alignment](https://blog.quarkslab.com/unaligned-accesses-in-cc-what-why-and-solutions-to-do-it-properly.html) can cause slowdowns, program termination, and is generally undefined behavior. LMDB guarantees a 2-byte alignment on nodes: http://www.lmdb.tech/doc/group__internal.html. However, some objects have a greater alignment than 2 (most notably 32-bit and 64-bit integers). This wouldn't necessarily cause alignment issues as long as objects with alignment greater than 2 are copied into a memory segment with correct alignment. Unfortunately, this is not the case for a lot of the `BlockchainLMDB` code. And this isn't just a theoretical issue either. If you apply the following git diff, compile, and run as a debug build the node, it will fail after an alternate block is handled to the core.

```diff
diff --git a/src/blockchain_db/lmdb/db_lmdb.cpp b/src/blockchain_db/lmdb/db_lmdb.cpp
index ff44923ff..18c80e8e5 100644
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -136,6 +136,11 @@ private:
   std::unique_ptr<char[]> data;
 };
 
+#define assert_ptr_alignment(p) do {                                                     \
+  static constexpr std::size_t align = alignof(std::remove_reference_t<decltype(*(p))>); \
+  assert(reinterpret_cast<std::uintptr_t>(p) % align == 0);                              \
+} while (0);
+
 }
 
 namespace cryptonote
@@ -4520,6 +4525,7 @@ bool BlockchainLMDB::get_alt_block(const crypto::hash &blkid, alt_block_data_t *
     throw0(DB_ERROR("Record size is less than expected"));
 
   const alt_block_data_t *ptr = (const alt_block_data_t*)v.mv_data;
+  assert_ptr_alignment(ptr);
   if (data)
     *data = *ptr;
   if (blob)
```

[...]

PR #10046, opened 2025-08-19
BlockchainLMDB: do not assume alignment for alt block entries

jeffro256:
Issue with alignment described in #10045. A full database migration shouldn't be required since alt blocks are flushed on startup.


hyc:
Sorry, I'm not seeing how this helps anything. It looks like all you're doing is aligning the memory buffer that you're passing to LMDB. That buffer's alignment is irrelevant, since it has nothing to do with where in the mmap the value will actually get written.

As I said in #10045, if you want to preserve alignment, you must pad the *size* of the block blob to the proper alignment, before writing it to LMDB.

jeffro256:
Ah okay, then I misunderstood your original answer, will rework now.

jeffro256:
I implemented padding locally, but it was a lot more complicated than just not assuming alignment, so that's what I did in the latest commit.

jeffro256:
Msys compile error is not relevant to this PR, see #10036.

hyc (approved):
Simple enough.

PR #10050, opened 2025-08-20
checkpoints: update to a recent block height

selsta:


jeffro256:
Why change the last checkpoint instead of adding a new one?

selsta:
The last line was not added in a release yet so I changed it.

jeffro256 (approved):

tobtoht (approved):

PR #10056, opened 2025-08-24
epee: only parse valid port

selsta:
Reported by hacksandhops and Ada Logic.

### DIFF
--- a/contrib/epee/src/net_parse_helpers.cpp
+++ b/contrib/epee/src/net_parse_helpers.cpp
@@ -92,7 +92,13 @@ namespace net_utils
     }
     return true;
   }
-  
+
+  static bool parse_port(const std::string& port_str, uint64_t& out_port)
+  {
+    out_port = 0;
+    return boost::conversion::try_lexical_convert(port_str, out_port) && out_port <= 65535;
+  }
+
   bool parse_uri(const std::string uri, http::uri_content& content)
   {
 
@@ -153,7 +159,8 @@ namespace net_utils
     }
     if(result[6].matched)
     {
-      content.port = boost::lexical_cast<uint64_t>(result[6]);
+      if (!parse_port(result[6].str(), content.port))
+        return false;
     }
     if(result[7].matched)
     {
@@ -189,7 +196,8 @@ namespace net_utils
     }
     if(result[6].matched)
     {
-      content.port = boost::lexical_cast<uint64_t>(result[6]);
+      if (!parse_port(result[6].str(), content.port))
+        return false;
     }
     if(result[7].matched)
     {


contrib/epee/src/net_parse_helpers.cpp
```diff
@@ -153,7 +153,14 @@ namespace net_utils
     }
     if(result[6].matched)
     {
-      content.port = boost::lexical_cast<uint64_t>(result[6]);
+      const auto port_str = result[6].str();
```
iamamyth: Would help to have a port parsing function, then this becomes:
```
if (!parse_port(result[6].str(), &content.port)) {
    return false;
}
```

Also removed the logging from the example, as I don't think parsing functions should be in the business of logging. If the caller wants info on what failed, there are plenty of well-worn options, like providing an out param, or an error code.

selsta: Added a function and removed logging.

iamamyth:
Windows build fails on this branch. I didn't inspect the output close enough to tell if the failure owes to the new code, but I note a problem with boost string code, and this commit introduces a new boost string-related call.

nahuhh:
> Windows build fails on this branch. I didn't inspect the output close enough to tell if the failure owes to the new code, but I note a problem with boost string code, and this commit introduces a new boost string-related call.

I think the failure is unrelated, caused by boost version 1.89 issues. The windows CI has been failing for a while 

cc @0xFFFC0000

selsta:
yes, build failure is unrelated

0xFFFC0000:
Windows build is something related to ICU and boost.locale. It started when pacman switched to boost 1.89. I am looking into this as we speak [1].

https://github.com/msys2/MINGW-packages/pull/25219/files

iamamyth (approved):

contrib/epee/src/net_parse_helpers.cpp
```diff
@@ -92,7 +92,18 @@ namespace net_utils
...
+  static bool parse_port(const std::string& port_str, uint64_t& out_port)
+  {
+    uint64_t parsed = 0;
+    if (!boost::conversion::try_lexical_convert(port_str, parsed) || parsed > 65535)
+    {
+      return false;
+    }
+    out_port = parsed;
+    return true;
+  }
```
jeffro256:
```suggestion
  static bool parse_port(const std::string& port_str, uint64_t& out_port)
  {
    out_port = 0;
    return boost::conversion::try_lexical_convert(port_str, out_port) && out_port <= 65535;
  }
```

selsta: Updated.

vtnerd (approved):

Issue #10064, opened 2025-08-31
Temporary rolling DNS checkpoints


Rucknium:
**TL;DR: A malicious mining pool is disrupting the network and could cause deep blockchain re-organizations. DNS checkpoints, which have existed for emergency use in the code since 2014, could prevent deep blockchain re-organizations, but Monero's blockchain consensus protocol would temporarily be less decentralized.**

## Background

As of August 2025, a malicious mining pool has used its large ([yet minority](https://shai-deshe.gitbook.io/parallel-thoughts/proof-of-work/the-qubic-minority-report)) share of Monero hashpower to pursue a ["selfish mining"](https://arxiv.org/abs/1311.0243) strategy. The actions of the mining pool have reduced the revenue of honest mining pools, caused delays in transaction confirmations, and produced short (up to 9 blocks) blockchain re-organizations. Large blockchain re-organizations (10 blocks and deeper) have a [nonzero probability of invalidating transactions](https://github.com/monero-project/research-lab/issues/95) (funds return to the wallet of the user who sent the funds) and can enable deliberate double-spend attacks (a form of fraud) against merchants who consider funds "fully confirmed" at a depth less than the re-org depth achieved by the attacker. An attacker with minority hashpower can still occasionally cause deep blockchain re-organizations. For example, a determined attacker with 30% hashpower share could cause a 10-block re-org about three times per day. (See "[Table: Duration of meta attack to achieve attack success probability of 50 percent](https://github.com/monero-project/research-lab/issues/102#issuecomment-2402750881)" for details.)

[Many](https://github.com/monero-project/research-lab/issues/144) [medium-](https://github.com/monero-project/research-lab/issues/141) [and](https://github.com/monero-project/research-lab/issues/140) [long-term proposals](https://github.com/monero-project/research-lab/issues/136) [are being](https://github.com/monero-project/research-lab/issues/98) [considered](https://github.com/monero-project/research-lab/issues/141). An emergency defense that could be deployed quickly is **rolling DNS checkpoints**.

[...]

PR #10075, opened 2025-09-10
checkpoints: modify dns checkpointing logic

nahuhh:
#10064

- reduce query frequency from 3600s -> 120s
- reduce ban time from 86400s to 300s
- increase requirement for dns servers to agree, from  `1 greater than 1/2` to `1 greater than 2/3`
- add remainder of moneropulse domains

Note: there is an issue unrelated to the PR that can leave nodes in a bad state
1. Node is reorged
2. Node receives a checkpoint that references the now orphaned chain
3. Node refuses to switch to the checkpointed chain, causing the node to get stuck. It will reject new blocks that dont match the checkpoints as well as rejecting the, currently orphaned, checkpointed chain

i believe its a logic issue with the `if` condition on L1899-2124 of src/cryptonote_core/blockchain.cpp [src](https://github.com/monero-project/monero/blob/8e9ab9677f90492bca3c7555a246f2a8677bd570/src/cryptonote_core/blockchain.cpp#L2089)

If there is a checkpoint, it _should_ reorg onto the checkpointed chain, not complain about it being orphaned and reject it.

The answer to the FIXME on [L2089](https://github.com/monero-project/monero/blob/8e9ab9677f90492bca3c7555a246f2a8677bd570/src/cryptonote_core/blockchain.cpp#L2089) is "yes".
```
// FIXME: is it even possible for a checkpoint to show up not on the main chain?
```

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -261,6 +261,11 @@ namespace cryptonote
   {
     m_blockchain_storage.set_enforce_dns_checkpoints(enforce_dns);
   }
+  //-----------------------------------------------------------------------------------
+  bool core::get_enforce_dns_checkpoints() const
+  {
+    return m_blockchain_storage.get_enforce_dns_checkpoints();
+  }
   //-----------------------------------------------------------------------------------------------
   bool core::update_checkpoints(const bool skip_dns /* = false */)
   {
@@ -269,7 +274,10 @@ namespace cryptonote
     if (m_checkpoints_updating.test_and_set()) return true;
 
     bool res = true;
-    if (!skip_dns && time(NULL) - m_last_dns_checkpoints_update >= 3600)
+    if (!skip_dns && (
+      (get_enforce_dns_checkpoints() && time(NULL) - m_last_dns_checkpoints_update >= 120) ||
+      (time(NULL) - m_last_dns_checkpoints_update >= 3600)
+    ))
     {
       res = m_blockchain_storage.update_checkpoints(m_checkpoints_path, true);
       m_last_dns_checkpoints_update = time(NULL);
@@ -693,8 +701,8 @@ namespace cryptonote
 
     // load json & DNS checkpoints, and verify them
     // with respect to what blocks we already have
-    const bool skip_dns_checkpoints = !command_line::get_arg(vm, arg_dns_checkpoints);
-    CHECK_AND_ASSERT_MES(update_checkpoints(skip_dns_checkpoints), false, "One or more checkpoints loaded from json or dns conflicted with existing checkpoints.");
+    m_skip_dns_checkpoints = !command_line::get_arg(vm, arg_dns_checkpoints);
+    CHECK_AND_ASSERT_MES(update_checkpoints(m_skip_dns_checkpoints), false, "One or more checkpoints loaded from json or dns conflicted with existing checkpoints.");
 
    // DNS versions checking
     if (check_updates_string == "disabled" || not allow_dns)
@@ -1599,6 +1607,7 @@ namespace cryptonote
     m_block_rate_interval.do_call(boost::bind(&core::check_block_rate, this));
     m_blockchain_pruning_interval.do_call(boost::bind(&core::update_blockchain_pruning, this));
     m_diff_recalc_interval.do_call(boost::bind(&core::recalculate_difficulties, this));
+    m_update_checkpoints_loop_interval.do_call(boost::bind(&core::update_checkpoints, this, m_skip_dns_checkpoints));
     m_miner.on_idle();
     m_mempool.on_idle();
     return true;
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -4710,6 +4710,11 @@ void Blockchain::set_enforce_dns_checkpoints(bool enforce_checkpoints)
 {
   m_enforce_dns_checkpoints = enforce_checkpoints;
 }
+//------------------------------------------------------------------
+bool Blockchain::get_enforce_dns_checkpoints() const
+{
+  return m_enforce_dns_checkpoints;
+}
 
 //------------------------------------------------------------------
 void Blockchain::block_longhash_worker(uint64_t height, const epee::span<const block> &blocks, std::unordered_map<crypto::hash, crypto::hash> &map) const
--- a/src/cryptonote_core/blockchain.h
+++ b/src/cryptonote_core/blockchain.h
@@ -807,6 +807,13 @@ namespace cryptonote
     void set_enforce_dns_checkpoints(bool enforce);
 
     /**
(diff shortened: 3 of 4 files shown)


src/cryptonote_protocol/cryptonote_protocol_handler.h
```diff
@@ -159,7 +159,7 @@ namespace cryptonote
     void drop_connection(cryptonote_connection_context &context, bool add_fail, bool flush_all_spans);
     void drop_connection_with_score(cryptonote_connection_context &context, unsigned int score, bool flush_all_spans);
     void drop_connection(const boost::uuids::uuid&);
-    void drop_connections(const epee::net_utils::network_address address);
+    void drop_connections(const epee::net_utils::network_address address, unsigned score = 5, bool block_light = false);
```
SNeedlewoods, line 162: Any reason to add `score` argument here? It's not really used in this PR

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1527,7 +1527,7 @@ namespace cryptonote
 
             if(bvc.m_verifivation_failed)
             {
-              drop_connections(span_origin);
+              drop_connections(span_origin, 5, bvc.m_failed_checkpoint);
```
SNeedlewoods: IMO a comment helps readability
```
drop_connections(span_origin, /* score */ 5, bvc.m_failed_checkpoint);
```

SNeedlewoods (commented):
Thank you nahuhh for your work.
Not an actual review, just trying to understand this PR and comment on minor things.

0xFFFC0000 (approved):
Looks functionally correct imho.  Just make sure the urls are up and working.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
nahuhh: Added

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -269,7 +274,10 @@ namespace cryptonote
     if (m_checkpoints_updating.test_and_set()) return true;
 
     bool res = true;
-    if (!skip_dns && time(NULL) - m_last_dns_checkpoints_update >= 3600)
+    if (!skip_dns && (
+      (get_enforce_dns_checkpoints() && time(NULL) - m_last_dns_checkpoints_update >= 120) ||
```
iamamyth, line 278: Not introduced by this PR, but elapsed time computations should use a monotonic clock, not the wall clock (I suggest fixing the issue in this PR as the goal appears to be a broad rollout of DNS checkpoints).

spirobel:
> #10064
> 
>     * reduce query frequency from 3600s -> 120s
> 
>     * reduce ban time from 86400s to 300s
> 
>     * increase requirement for dns servers to agree, from  `1 greater than 1/2` to `1 greater than 2/3`
> 
>     * add remainder of moneropulse domains
> 
> 
> Note: there is an issue unrelated to the PR that can leave nodes in a bad state
> 
>     1. Node is reorged
> 
>     2. Node receives a checkpoint that references the now orphaned chain
> 
>     3. Node refuses to switch to the checkpointed chain, causing the node to get stuck. It will reject new blocks that dont match the checkpoints as well as rejecting the, currently orphaned, checkpointed chain
> 
> 
> i believe its a logic issue with the `if` condition on L1899-2124 of src/cryptonote_core/blockchain.cpp [src](https://github.com/monero-project/monero/blob/8e9ab9677f90492bca3c7555a246f2a8677bd570/src/cryptonote_core/blockchain.cpp#L2089)
> 
> If there is a checkpoint, it _should_ reorg onto the checkpointed chain, not complain about it being orphaned and reject it.
> 
> The answer to the FIXME on [L2089](https://github.com/monero-project/monero/blob/8e9ab9677f90492bca3c7555a246f2a8677bd570/src/cryptonote_core/blockchain.cpp#L2089) is "yes".
> 
> ```
> // FIXME: is it even possible for a checkpoint to show up not on the main chain?
> ```

thoughts on the second part: https://libera.monerologs.net/monero-research-lab/20251008#c597669  the logic really shouldnt be in this function as these two things are not related. Longest chain rule is normal operation, while checkpointing is above that. It is a separate rule so why should it be handled in the same function? 

We also have to separate here between nodes that miners operate and nodes that users rely on. Checkpointing is a tool for honest hashrate to coordinate and form the longer chain, even if a malicious party with a large amount of hashrate tries to disrupt the network.

Another aspect is that currently the blocking behavior is intermingled with this as well. We have 3 topics that need to get untangled: the two consensus rules and blocking behavior.

possible way to fix: https://libera.monerologs.net/monero-research-lab/20251009#c597742
 To fix this copy the function, delete the reorg on longest chain in one copy. In the other, delete the application of the checkpointing. Then add logic according to if the checkpointing flag is active. What about reorgs on longest chain rule between checkpoints? When checkpointing is active, we still want to reorg in between checkpoints, right?

<spirobel:kernal.eu> How often do we checkpoint? if we checkpoint every block, this ambiguity is resolved, somehow I had in my mind we would only checkpoint every 10th block for some reason. No idea why.
[12:04](https://libera.monerologs.net/monero-research-lab/20251009#c597755)
DataHoarder
5/7 is needed due to dns latency
[12:05](https://libera.monerologs.net/monero-research-lab/20251009#c597756)
DataHoarder
Right now checkpoints are issued at depth 2, at most eve try five minutes
[13:14](https://libera.monerologs.net/monero-research-lab/20251009#c597766)
br-m
<spirobel:kernal.eu> DataHoarder: okay so there is no need to reorg in between checkpoints
[13:15](https://libera.monerologs.net/monero-research-lab/20251009#c597767)
DataHoarder
due to delays in DNS there might be reorgs
[13:15](https://libera.monerologs.net/monero-research-lab/20251009#c597768)
DataHoarder
say checkpoint at 100 is published, then 102 (when checkpointer is at 104)
[13:15](https://libera.monerologs.net/monero-research-lab/20251009#c597769)
DataHoarder
while 102 becomes consistent on DNS servers and caches a reorg can happen
[13:16](https://libera.monerologs.net/monero-research-lab/20251009#c597770)
DataHoarder
when 102 becomes valid for each specific node, they'd reorg back
[13:17](https://libera.monerologs.net/monero-research-lab/20251009#c597771)
br-m
<spirobel:kernal.eu> to be more specific: to reorg according to longest chain rule in between checkpoints
[13:17](https://libera.monerologs.net/monero-research-lab/20251009#c597772)
DataHoarder
between checkpoints the source / targets are fixed
[13:17](https://libera.monerologs.net/monero-research-lab/20251009#c597773)
DataHoarder
genesis block is a checkpoint
[13:18](https://libera.monerologs.net/monero-research-lab/20251009#c597774)
DataHoarder
target checkpoint may be in an altchain or threshold of
[13:23](https://libera.monerologs.net/monero-research-lab/20251009#c597775)
br-m
<spirobel:kernal.eu> okay great. So the ambiguity is removed.

Issue #10067, opened 2025-09-02
Wallet loses tx key and transaction data - possibly caused by reorgs


DrFredPhD:
For years of use I've had an intermittent issue with the wallet file losing track of tx key and receiving address of outgoing transactions. I mainly use the cli wallet and have no idea what causes it. Since first running into the issue I've religiously closed the wallet with the 'save' command, followed by 'exit'. Usually after a transaction I'll use 'show_transfers' and the receiving address will be visible and 'get_tx_key' will work. At some point in the future I can reopen the wallet and that data will be gone and 'get_tx_key' will fail. (Loading the wallet in the gui also does not help)

I initially had no idea what caused it, but recently I noticed one of the transactions I lost data for was one that was initially included in a later orphaned block. Perhaps the wallet loses this data if the original block a transaction was in is reorged out.

nahuhh:
Sorry, i thought i opened an issue for this already (i did not).

https://github.com/monero-project/monero/issues/8372#issuecomment-3243302250
> The reorgs DO mess up wallet history, dropping tx details.

tldr: yes, the cache drops all knowledge of the tx when it gets sent back to the txpool. Probably also related to #10013 - when it is sent back to the txpool, the wallet might even forget the tx exists until it is mined again.

I've experienced the entire tx disappearing from my wallet history, and reappearing when mined (sans-details)

DrFredPhD:
Yes the tx disappearing entirely then reappearing without info is what finally made be realise it was due to orphaned blocks. If there any way of this being fixed? Or do I need to manually copy out recipient and tx key acter each transaction if I think I could ever possibly need it in the future?

jeffro256:
Is this with a Trezor?

nahuhh:
With all wallets

mmbbee:
We (a large exchange) also see this issue

selsta:
@mmbbee we are currently investigating the issue, once a fix is found and reviewed we will put out a new release

DrFredPhD:
I also can't confirm this first hand as it was a transaction sent to me, but looks like even copying down the tx key is not enough, someone provided me a tx key for a transaction that was first included in an orphaned block, and I was unable to use it to decode the transaction, and they were then not able to retrieve a tx key from wallet. I did still receive the transaction after another good block

j-berman:
@DrFredPhD, are you using a regular full wallet? And at any point did you use `rescan_bc` by chance, or delete the wallet file manually?

[...]

Issue #10013, opened 2025-07-28
Bug: unidentified spend tx in pool on wallet restore


j-berman:
If you send a Monero tx, it enters the pool, and then restore the wallet, then the wallet won't identify the spend tx in the pool. If a user then attempts to construct *another* tx, the wallet may attempt to re-spend an output already spent in the pool, causing the node to reject the tx as a double spend. Note: this bug requires the pool tx to remain in the pool (as soon as it enters the chain, the wallet will identify the tx as a spend correctly).

Expected behavior: the pool tx should be identified as a spend, and any spent outputs should no longer be usable. The wallet state on restore should match the wallet state of the wallet that originally sent the tx.

Identified by @nahuhh in testing the FCMP++/Carrot integration (the bug is unrelated to the integration, I reproduced the issue on current master).

0xFFFC0000:
This should effectively solve this issue. Once confirmed, I will send a PR.

```
$ >> git diff 
diff --git a/src/wallet/wallet2.cpp b/src/wallet/wallet2.cpp
index 432b5d5c7..8c6f7e4e0 100644
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -3714,6 +3714,19 @@ void wallet2::update_pool_state(std::vector<std::tuple<cryptonote::transaction,
   process_txs.clear();
   if (m_background_syncing)
     return;
+  // Ensure we can detect spends from mempool after restore by making key images known
+  if (!m_watch_only && !m_multisig && has_unknown_key_images())
+  {
+    MDEBUG("Unknown key images detected; rescanning spent status before processing pool");
+    try
+    {
+      rescan_spent(); // updates spent status for both pool and chain
+    }
+    catch (const std::exception &e)
+    {
+      MWARNING("rescan_spent() failed before pool update: " << e.what());
+    }
+  }
   bool updated = false;
   if (m_pool_info_query_time != 0 && try_incremental)
   {
@@ -3755,7 +3768,19 @@ void wallet2::update_pool_state_by_pool_query(std::vector<std::tuple<cryptonote:
 {
   MTRACE("update_pool_state_by_pool_query start");
   process_txs.clear();
-
+  // Ensure we can detect spends from mempool after restore by making key images known
+  if (!m_watch_only && !m_multisig && has_unknown_key_images())
+  {
+    MDEBUG("Unknown key images detected; rescanning spent status before processing pool (by_pool_query)");
+    try
+    {
[...]

PR #10077, opened 2025-09-11
wallet: more robust reorg handling for multisig info messages

jeffro256:
Partially fixes an issue where reorgs cause multisig info messages to be dropped. This triggers the `multisig_export_needed` exception when attempting to sign multisig transaction sets. This commit accomplishes this by making the `m_multisig_rescan_info` and `m_multisig_rescan_k`: A) structured by output pubkey instead of by index into `m_transfers`, and B) persistent across multiple calls to `refresh()`. So if a received enote is temporarily reorged out of the chain (e.g. when Qubic mines an empty block on a selfish fork), the multisig info originally attached to that enote can be recovered after an arbitrary number of refreshes (as long as the wallet is not closed). Hopefully this improves the UX of Monero multisig apps by requiring fewer round trips to participants during sync and signing.

Potential (not implemented in this commit) improvements which require greater changes to the wallet:
* Save `m_multisig_rescan_info` and `m_multisig_rescan_k` in the wallet cache so that this fix works even when the wallet is closed between refreshes. This is a backwards-incompatible change to the wallet cache file format.
* Have the multisig info messages index partial key images and L-R values by its associated output pubkey, not its index into its local copy of `m_transfers`. This is a backwards-incompatible change to the multisig messaging protocol.

Further tests incoming.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -2194,10 +2194,9 @@ void wallet2::scan_output(const cryptonote::transaction &tx, bool miner_tx, cons
   THROW_WALLET_EXCEPTION_IF(i >= tx.vout.size(), error::wallet_internal_error, "Invalid vout index");
 
   // if keys are encrypted, ask for password
-  if (m_ask_password == AskPasswordToDecrypt && !m_unattended && !m_watch_only && m_multisig_rescan_k.empty() && !m_background_syncing)
+  if (is_key_encryption_enabled() && !m_multisig)
   {
-    static critical_section password_lock;
-    CRITICAL_REGION_LOCAL(password_lock);
+    const boost::lock_guard password_lock(m_encrypt_keys_after_refresh_mutex);
     if (!m_encrypt_keys_after_refresh && !m_processing_background_cache)
     {
       boost::optional<epee::wipeable_string> pwd = m_callback->on_get_password(pool ? "output found in pool" : "output received");
@@ -2556,13 +2555,8 @@ void wallet2::process_new_transaction(const crypto::hash &txid, const cryptonote
 	    m_pub_keys[tx_scan_info[o].in_ephemeral.pub] = m_transfers.size()-1;
             if (output_tracker_cache)
               (*output_tracker_cache)[std::make_pair(tx.vout[o].amount, td.m_global_output_index)] = m_transfers.size() - 1;
-            if (m_multisig)
-            {
-              THROW_WALLET_EXCEPTION_IF(m_multisig_rescan_k.empty() && !m_multisig_rescan_info.empty(),
-                  error::wallet_internal_error, "NULL m_multisig_rescan_k");
-              if (!m_multisig_rescan_info.empty() && m_multisig_rescan_info.front().size() >= m_transfers.size())
-                update_multisig_rescan_info(m_multisig_rescan_k, m_multisig_rescan_info, m_transfers.size() - 1);
-            }
+            if (is_multisig_key_image_rescan_possible(td.get_public_key()))
+              update_multisig_rescan_info(m_transfers.size() - 1);
 	    LOG_PRINT_L0("Received money: " << print_money(td.amount()) << ", with tx: " << txid);
 	    if (!ignore_callbacks && 0 != m_callback)
 	      m_callback->on_money_received(height, txid, tx, td.m_amount, 0, td.m_subaddr_index, spends_one_of_ours(tx), td.m_tx.unlock_time);
@@ -2631,13 +2625,8 @@ void wallet2::process_new_transaction(const crypto::hash &txid, const cryptonote
             }
             if (output_tracker_cache)
               (*output_tracker_cache)[std::make_pair(tx.vout[o].amount, td.m_global_output_index)] = kit->second;
-            if (m_multisig)
-            {
-              THROW_WALLET_EXCEPTION_IF(m_multisig_rescan_k.empty() && !m_multisig_rescan_info.empty(),
-                  error::wallet_internal_error, "NULL m_multisig_rescan_k");
-              if (!m_multisig_rescan_info.empty() && m_multisig_rescan_info.front().size() >= m_transfers.size())
-                update_multisig_rescan_info(m_multisig_rescan_k, m_multisig_rescan_info, m_transfers.size() - 1);
-            }
+            if (is_multisig_key_image_rescan_possible(td.get_public_key()))
+              update_multisig_rescan_info(m_transfers.size() - 1);
             THROW_WALLET_EXCEPTION_IF(td.get_public_key() != tx_scan_info[o].in_ephemeral.pub, error::wallet_internal_error, "Inconsistent public keys");
 	    THROW_WALLET_EXCEPTION_IF(td.m_spent, error::wallet_internal_error, "Inconsistent spent status");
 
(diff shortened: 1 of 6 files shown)


nahuhh:
"Multisig messages dropped from wallet cache after a reorg"

This sounds like the same issue as #10067 where the entire cache of the tx is dropped (and tx forgotten) when a tx goes from confirmed -> txpool as the result of a reorg? (and the same fix?)

jeffro256:
It's a very similar reason for the issue, and a similar fix, although not *quite* the same.

jeffro256:
I'm going to work on issue #10067 very soon

nahuhh:
"the multisig info originally attached to that enote can be recovered after an arbitrary number of refreshes (as long as the wallet is not closed)"

IIUC, when the tx is reorged out, the tx is entirely forgotten by the wallet (until confirmed again). 

forgive me if im wrong, but if the multisig info is a part of the cache that is forgotten when the tx is dropped from the wallet history, wouldnt the fix be to ensure that all of the tx cache info is persistent + solve #10013 is so that when sent back to the txpool, the wallet doesnt lose track of the tx?

jeffro256:
> if the multisig info is a part of the cache that is forgotten when the tx is dropped from the wallet history, wouldnt the fix be to ensure that all of the tx cache info is persistent + solve https://github.com/monero-project/monero/issues/10013 is so that when sent back to the txpool, the wallet doesnt lose track of the tx?

If I am correct about what I think the issue is for #10013, then fixing that issue wouldn't resolve this issue, and vice versa. Different code paths touch different cache data. The root cause is similar: bad reorg assumptions leads to deleting things from the cache, but they don't necessarily have 100% overlap.

nahuhh:
> bad reorg assumptions leads to deleting things from the cache, but they don't necessarily have 100% overlap.

i assume this issue is reproducible by popping a  block, and that the issue is that the wallet forgets the entire tx when the tx is moved from block -> txpool

edit: so my assumption is that this doesnt have to do with reorgs, but has to do with any time a block is popped. 10013 isnt the direct issue, but solving 10013 should prevent a wallet from forgetting about the tx / being able to follow or detect txs when they are sent to the txpool by an external influences (reorgs, popped blocks, another wallet), as well as preserving the cache.

It just a hypothesis, but the reason the cache is dropped, to me, sounds like its a result of the wallet dropping / forgetting / unable to detect or follow its own txs in the pool

jeffro256:
> so my assumption is that this doesnt have to do with reorgs, but has to do with any time a block is popped. 10013 isnt the direct issue, but solving 10013 should prevent a wallet from forgetting about the tx / being able to follow or detect txs when they are sent to the txpool by an external influences (reorgs, popped blocks, another wallet), as well as preserving the cache.

Technically, according to current consensus rules, you cannot have blocks popped without a reorg containing an alternate chain. The only time the daemon's chain will strictly decrease in size is when a user manually runs the `pop_blocks` command; it's doesn't occur "naturally". So I call it "reorg handling", but I guess you can also call it "block popping handling" if you want to.

As for #10013 remembering transactions, the problem isn't the wallet forgetting the transaction itself, the problem is the wallet forgetting off-chain ephemeral information attached to a transaction. In `wallet2`, this info (tx keys, destinations) is attached to the entries for outgoing transactions. However, multisig messages are attached to the entries for received enotes. They are different parts of the cache that are updated in different parts of `wallet2`, which is why the fixes don't overlap.

jeffro256:
Test commit at `HEAD` fails when applied to `master` without fix commit, proving a scenario in which this fix works where `master` doesn't.

src/wallet/wallet2.cpp
```diff
@@ -14562,34 +14562,88 @@ cryptonote::blobdata wallet2::export_multisig()
...
+  const auto done_processing = epee::misc_utils::create_scope_leave_handler([this]() {
+    m_processing_background_cache = false;
+  });
 
-  if (!m_multisig_rescan_k.empty() && !m_multisig_rescan_info.empty())
-    refresh(false);
+  // Unconditionally refresh to attempt to capture latest on-chain received enotes that other
+  // participants have seen and we have not.
+  try { refresh(false); }
+  catch (...) {}
```
j-berman, line 14646: Why allow proceeding?

src/wallet/wallet2.cpp
```diff
@@ -14562,34 +14562,88 @@ cryptonote::blobdata wallet2::export_multisig()
...
-  MDEBUG("update_multisig_rescan_info: updating index " << n);
   transfer_details &td = m_transfers[n];
-  td.m_multisig_info.clear();
-  for (const auto &pi: info)
+  const crypto::public_key output_pubkey = td.get_public_key();
+  const auto msinfo_it = m_multisig_rescan_info.find(output_pubkey);
+  CHECK_AND_ASSERT_THROW_MES(msinfo_it != m_multisig_rescan_info.cend(),
+    "multisig rescan info missing for output pubkey: " << output_pubkey);
+
+  if (is_key_encryption_enabled() && !m_processing_background_cache)
```
j-berman, line 14588: This if statement could be implemented in a new function to avoid duplicating the logic here and in `scan_output`

src/wallet/wallet2.cpp
```diff
@@ -14562,34 +14562,88 @@ cryptonote::blobdata wallet2::export_multisig()
...
+    {
+      memwipe(td.m_multisig_k.data(), td.m_multisig_k.size() * sizeof(td.m_multisig_k[0]));
+      td.m_multisig_k = mskey_it->second;
+    }
+  }
+  catch (const std::exception &e)
+  {
+    MERROR("Encountered exception while attempting to build composite multisig key image, reverting state: "
+      << e.what());
+    std::swap(old_multisig_info, td.m_multisig_info);
```
j-berman, line 14626: Should it re-throw the exception?

src/wallet/wallet2.cpp
```diff
@@ -14562,34 +14562,88 @@ cryptonote::blobdata wallet2::export_multisig()
...
+    // try update key images and multisig exchange infos
+    td.m_key_image = get_multisig_composite_key_image(n); // may throw
+    m_key_images.erase(old_key_image);
+    td.m_key_image_known = true;
+    td.m_key_image_request = false;
+    td.m_key_image_partial = false;
+    m_key_images[td.m_key_image] = n;
+
+    // update local multisig keys
+    const auto mskey_it = m_multisig_rescan_k.find(output_pubkey);
```
j-berman, line 14615: Under what circumstance would the wallet have rescan info for the output but not rescan k?

j-berman (commented):
The general idea behind the PR makes sense

src/wallet/wallet2.cpp
jeffro256: Because it is simply a best attempt, but refreshing doesn't necessarily guarantee the the imported outputs from peers are guaranteed to "line up" with your current refresh state. We don't lose information in `m_multisig_rescan_{k,info}` when populating them anymore, besides to overwrite stale data, so the refresh doesn't *need* to work to not lose information.

src/wallet/wallet2.cpp
```diff
@@ -14544,6 +14543,7 @@ cryptonote::blobdata wallet2::export_multisig()
       const rct::multisig_kLRki kLRki = get_multisig_kLRki(n, td.m_multisig_k.back());
       info[n].m_LR.push_back({kLRki.L, kLRki.R});
     }
+    m_multisig_rescan_k[td.get_public_key()] = td.m_multisig_k;
```
ACK-J, line 14546: Nonce may be used more than once https://github.com/xmrack/monero-review/issues/280

PR #10081, opened 2025-09-13
wallet: re-add detatched confirmed transactions on reorg

### LINKED ISSUE #10067: Wallet loses tx key and transaction data - possibly caused by reorgs

jeffro256:
Resolves #10067 (NOT TESTED)

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -790,6 +790,24 @@ void drop_from_short_history(std::list<crypto::hash> &short_chain_history, size_
   }
 }
 
+static tools::wallet2::unconfirmed_transfer_details disconfirm_transfer(const tools::wallet2::confirmed_transfer_details &ctd)
+{
+  tools::wallet2::unconfirmed_transfer_details utd{};
+  utd.m_tx = ctd.m_tx;
+  utd.m_amount_in = ctd.m_amount_in;
+  utd.m_amount_out = ctd.m_amount_out;
+  utd.m_change = ctd.m_change;
+  utd.m_sent_time = ctd.m_timestamp;
+  utd.m_dests = ctd.m_dests;
+  utd.m_payment_id = ctd.m_payment_id;
+  utd.m_state = tools::wallet2::unconfirmed_transfer_details::pending;
+  utd.m_timestamp = ctd.m_timestamp;
+  utd.m_subaddr_account = ctd.m_subaddr_account;
+  utd.m_subaddr_indices = ctd.m_subaddr_indices;
+  utd.m_rings = ctd.m_rings;
+  return utd;
+}
+
 size_t estimate_rct_tx_size(int n_inputs, int mixin, int n_outputs, size_t extra_size, bool bulletproof, bool clsag, bool bulletproof_plus, bool use_view_tags)
 {
   size_t size = 0;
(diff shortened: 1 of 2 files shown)


j-berman:
Re-adding confirmed txs back to unconfirmed seems like a solid approach here to make sure destination addresses don't get lost. It also looks like we'd want to be sure to re-mark as spent any dis-confirmed transfers (so the wallet won't attempt to re-spend an output that's part of a spent tx that was placed back in `m_unconfirmed_txs` and get double spend errors).

But, it seems there may be another issue at play here causing tx keys to get lost as reported in the original issue.. It's worth some more digging on that front.

j-berman:
W.r.t. this PR's relation to #10083 based on my reading:

This function is what determines which txs from the pool get processed by the wallet:

https://github.com/monero-project/monero/blob/8d4c625713e3419573dfcc7119c8848f47cabbaa/src/wallet/wallet2.cpp#L3581-L3639

That function *should* return false for dis-confirmed txs that are placed back in `m_unconfirmed_txs` from this PR. If it doesn't return false here (because the wallet has already seen it in the pool at some point):

https://github.com/monero-project/monero/blob/8d4c625713e3419573dfcc7119c8848f47cabbaa/src/wallet/wallet2.cpp#L3592-L3600

Then it would be expected to return false here:

https://github.com/monero-project/monero/blob/8d4c625713e3419573dfcc7119c8848f47cabbaa/src/wallet/wallet2.cpp#L3632-L3633

So, even with #10083, I wouldn't expect a dis-confirmed tx to get re-processed in `process_new_transaction`, and therefore wouldn't get marked as spent with the change from #10083.

Thus, I think it makes sense to explicitly mark these dis-confirmed txs as spent in this PR.

______

However! Following the logic from above, it may make sense to place dis-confirmed received payments back into `m_unconfirmed_payments` as well, since it might still be in `m_scanned_pool_txs` and therefore might not get processed in `process_new_transaction`. Or implement similar logic as #10083 to wipe `m_scanned_pool_tx` to make sure pool txs get reprocessed.

nahuhh:
FWIW i did some testing with cli and rpc, and was unable to get them to drop the tx keys during empty block reorgs, or during popping of blocks. i _was_ able to repro the dropping of recipient info.

( i have not tested this pr to see if it resolves the recipient dropping )

note: i did experience dropping of keys in a "3rd party" wallet, so it may be related to the wallet implementation?
in fact, the tx in question seems to be missing from the wallet altogether now, though it was definitely present when confirming twice.

j-berman:
> However! Following the logic from above, it may make sense to place dis-confirmed received payments back into m_unconfirmed_payments as well, since it might still be in m_scanned_pool_txs and therefore might not get processed in process_new_transaction. Or implement similar logic as https://github.com/monero-project/monero/pull/10083 to wipe m_scanned_pool_tx to make sure pool txs get reprocessed.

I think the latter (wiping `m_scanned_pool_txs`) is fine since all the data will be recoverable. If going with the former (placing dis-confirmed back into `m_unconfirmed_payments`), would also need to avoid placing coinbases back.

PR #10083, opened 2025-09-15
wallet: identify spends in pool when scanning

### LINKED ISSUE #10013: Bug: unidentified spend tx in pool on wallet restore

j-berman:
Fixes #10013 

- Make sure to mark identified spends in the pool as spends. The wallet might not know these have been spent if it wasn't the wallet that relayed the tx to the daemon, or the wallet was cleared via rescan_bc.
- Make sure to add spends to `m_unconfirmed_txs` if not present.
- Make sure to process the entire pool again if refreshing for the first time. The wallet fetches pool and blocks at the same time. The wallet scans blocks first, then pool. If the wallet identifies received outputs in the chain, then it may have spent those received outputs in the pool. So we make sure to re-process the entire pool again after scanning the chain for the first time.
- EDIT: Multisig wallets that know about spent key images can now detect spend txs in the pool. This PR updates tests to account for that new functionality.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -923,6 +923,11 @@ bool get_short_payment_id(crypto::hash8 &payment_id8, const tools::wallet2::pend
   return false;
 }
 
+uint64_t get_outgoing_amount(const cryptonote::transaction &tx, const uint64_t amount_spent)
+{
+  return tx.version == 1 ? get_outs_money_amount(tx) : (amount_spent - tx.rct_signatures.txnFee);
+}
+
 tools::wallet2::tx_construction_data get_construction_data_with_decrypted_short_payment_id(const tools::wallet2::pending_tx &ptx, hw::device &hwdev)
 {
   tools::wallet2::tx_construction_data construction_data = ptx.construction_data;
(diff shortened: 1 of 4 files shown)


selsta:
Do you know if this multisig failure is new?

j-berman:
It's new and is a result of this PR, working on it now

j-berman:
Fixed the multisig functional test failure in the latest (I made no changes to the code outside of the multisig functional test).

**Without this PR**: only the signer who submits a multisig tx would know about the spend tx in the pool. The multisig functional test was testing that exact behavior, erroring if any other multisig participants (excluding the signer) could detect the spend tx in the pool.

**With this PR**: any multisig wallet that has prior calculated any key image(s) used in a tx will identify the tx as a spend in the pool, even if not the signer who submitted the tx. Because of this, I updated the functional multisig test to make sure that if the multisig wallet has already calculated any of a tx's spent key images, then the wallet should be able to detect the spend tx in the pool.

Pinging @woodser who may not expect this new (and improved) behavior for multisig wallets.

nahuhh (approved):
Tested rescanned and duplicate wallets. Works as expected. Thanks

did not test:
> So we make sure to re-process the entire pool again after scanning the chain for the first time.

I'm not sure how/if this would negativelty effect wallets if the txpool was huge (>100mb). note: I don't think its a major concern even if it was noticable

selsta:
@j-berman please open against release branch

SChernykh:
C++ code LGTM. As for the Python tests, I looked at the code and they also look fine to the best of my knowledge about the testing infrastructure (I don't know much about it).

jeffro256:
Logic seems sound, functional tests works iff PR is applied. For wallets where a transaction is initiated with destinations info, and this PR isn't needed, that information shouldn't be overridden. Everything looks good!

PR #10084, opened 2025-09-15
wallet: add `wallet2_basic` library types

jeffro256:
The `wallet2_basic` library is a minimalistic library for loading / storing historic `wallet2` files. It has several design goals:
* Enable future migration away from monolithic `wallet2` codebase without losing funds
* Simplify FCMP++/Carrot integration
* Speed-up compile-time for current codebase, but especially the FCMP++/Carrot integration, which is split over several translation units

This PR adds the types necessary to load `wallet2` files (besides cache and keys itself) into a separate library, `wallet2_basic`. It also refactors `wallet2` to depend on `wallet2_basic`. No difference in behavior should be observed. This PR is part of upstreaming Carrot/FCMP++.

### DIFF
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -34,12 +34,6 @@
 
 #include <boost/program_options/options_description.hpp>
 #include <boost/program_options/variables_map.hpp>
-#if BOOST_VERSION >= 107400
-#include <boost/serialization/library_version_type.hpp>
-#endif
-#include <boost/serialization/list.hpp>
-#include <boost/serialization/vector.hpp>
-#include <boost/serialization/deque.hpp>
 #include <boost/thread/lock_guard.hpp>
 #include <atomic>
 #include <random>
@@ -53,7 +47,6 @@
 #include "rpc/core_rpc_server_commands_defs.h"
 #include "cryptonote_basic/cryptonote_format_utils.h"
 #include "cryptonote_core/cryptonote_tx_utils.h"
-#include "common/unordered_containers_boost_serialization.h"
 #include "common/util.h"
 #include "crypto/chacha.h"
 #include "crypto/hash.h"
@@ -73,6 +66,7 @@
 #include "message_store.h"
 #include "fee_priority.h"
 #include "fee_algorithm.h"
+#include "wallet2_basic/wallet2_types.h"
 
 #undef MONERO_DEFAULT_LOG_CATEGORY
 #define MONERO_DEFAULT_LOG_CATEGORY "wallet.wallet2"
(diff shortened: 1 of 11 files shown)


src/wallet/wallet2_basic/CMakeLists.txt
j-berman, line 1: Empty file needs to be here?

src/wallet/wallet2.h
```diff
@@ -34,12 +34,6 @@
 
 #include <boost/program_options/options_description.hpp>
 #include <boost/program_options/variables_map.hpp>
-#if BOOST_VERSION >= 107400
-#include <boost/serialization/library_version_type.hpp>
```
j-berman, line 38: This isn't needed?

src/wallet/wallet2.h
```diff
@@ -2056,394 +1766,6 @@ namespace tools
...
-BOOST_CLASS_VERSION(tools::wallet2::address_book_row, 18)
-BOOST_CLASS_VERSION(tools::wallet2::background_synced_tx_t, 0)
-BOOST_CLASS_VERSION(tools::wallet2::background_sync_data_t, 0)
-
-namespace boost
-{
-  namespace serialization
-  {
-    template<class Archive, class F, class S, class T>
-    inline void serialize(
```
j-berman, line 2078: Looks like this wasn't replaced, not sure what it is

src/wallet/wallet2_basic/wallet2_boost_serialization.h
```diff
@@ -0,0 +1,433 @@
...
+//standard headers
+
+//forward declarations
+
+namespace boost
+{
+namespace serialization
+{
+template <class Archive>
+void serialize(Archive &a, wallet2_basic::multisig_info::LR &x, const unsigned int ver)
```
j-berman, line 51: inline?

src/wallet/wallet2_basic/wallet2_boost_serialization.h
```diff
@@ -0,0 +1,433 @@
...
+        if (!typename Archive::is_saving() && x.m_change != (uint64_t)1)
+        {
+        if (x.m_amount_in > (x.m_amount_out + x.m_change))
+            x.m_amount_out += x.m_change;
+        }
+    }
+    if (ver < 4)
+    {
+        if (!typename Archive::is_saving())
+        x.m_unlock_time = 0;
```
j-berman: Indent

src/wallet/wallet2_basic/wallet2_boost_serialization.h
```diff
@@ -0,0 +1,433 @@
...
+{
+    a & x.m_address;
+    if (ver < 18)
+    {
+        crypto::hash payment_id;
+        a & payment_id;
+        x.m_has_payment_id = !(payment_id == crypto::null_hash);
+        if (x.m_has_payment_id)
+        {
+        bool is_long = false;
```
j-berman: Indent inside this if

j-berman (commented):
Just a thought: one simple test to raise confidence the format before and after this PR is the same, would be to make sure the target file contents stored in the `wallet_storage` `store_to_file2file` test is equal to the source file contents.

Bumping the wallet cache version would break that test, but maybe we would want the wallet cache in `tests/data` updated in that case?

src/wallet/wallet2.h
jeffro256: Not after #9940

src/wallet/wallet2_basic/CMakeLists.txt
jeffro256: Only if we want to keep the `add_subdirectory(wallet2_basic)` line in the `wallet` CMakeLists.txt. I'm fine with removing it for now if you'd like.

src/wallet/wallet2.h
jeffro256: AFAIK, it shouldn't be necessary after #9940 since there aren't any remaining tuples to serialize, but it would be good to triple check.

src/wallet/wallet2_basic/CMakeLists.txt
j-berman: Ok with me to keep

src/wallet/wallet2_basic/wallet2_boost_serialization.h
jeffro256: According the C++ [One Definition Rule](https://en.cppreference.com/w/cpp/language/definition.html#One_Definition_Rule) (ODR), identical definitions for non-fully-specialized templated functions can appear multiple times in the program, as long as it is only once per translation unit. In other words, as long as the function isn't fully specialized, template functions are more or less `inline` by default.

jeffro256:
> Just a thought: one simple test to raise confidence the format before and after this PR is the same, would be to make sure the target file contents stored in the wallet_storage store_to_file2file test is equal to the source file contents.

A couple issues: A) It would take me a while since I would effectively need to re-write `wallet2::store()` in the unit tests directory to force Boost usage since `wallet2` doesn't use Boost for writing files anymore, it always uses the on-chain serialization. In some sense, it doesn't make sense to test writing Boost since that is not done anymore. B) Even if the serialization is correct, it might not produce the same byte-wise result since `unordered_map`s and `unordered_set`s` don't have a guaranteed order.

j-berman:
Fair points, the PR LGTM

j-berman (approved):

src/wallet/wallet2_basic/wallet2_boost_serialization.h
```diff
@@ -0,0 +1,433 @@
...
+    if (ver < 4)
+        return;
+    a & x.m_amount_in;
+    a & x.m_amount_out;
+    if (ver < 6)
+    {
+        // v<6 may not have change accumulated in m_amount_out, which is a pain,
+        // as it's readily understood to be sum of outputs.
+        // We convert it to include change from v6
+        if (!typename Archive::is_saving() && x.m_change != (uint64_t)1)
```
selsta:
```suggestion
        if (!typename Archive::is_saving() && x.m_change != (uint64_t)-1)
```

src/wallet/wallet2_basic/wallet2_boost_serialization.h
```diff
@@ -0,0 +1,433 @@
...
+    a & x.m_timestamp;
+    if (ver < 3)
+    {
+        // v<3 may not have change accumulated in m_amount_out, which is a pain,
+        // as it's readily understood to be sum of outputs. Whether it got added
+        // or not depends on whether it came from a unconfirmed_transfer_details
+        // (not included) or not (included). We can't reliably tell here, so we
+        // check whether either yields a "negative" fee, or use the other if so.
+        // We convert it to include change from v3
+        if (!typename Archive::is_saving() && x.m_change != (uint64_t)1)
```
selsta:
```suggestion
        if (!typename Archive::is_saving() && x.m_change != (uint64_t)-1)
```

selsta (changes requested):

selsta (approved):

j-berman (approved):

PR #10086, opened 2025-09-17
wallet2: warn instead of throw when RingDB doesn't include spend

j-berman:
A reorg can end up causing an output's position in the chain to move. Since the wallet doesn't update the RingDB on reorg, it may refer to the output's stale position in the chain, and then throw when attempting to spend that output in the future.

Logging a warning instead of throwing seems a reasonable solution rather than introducing complex logic to update the stale ring member's value on rerog, since RingDB can be deprecated with FCMP++, and the downside to warning here is minimal.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -9300,8 +9300,11 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
               MINFO("Ignoring output " << out << ", too recent");
             }
           }
-          THROW_WALLET_EXCEPTION_IF(!own_found, error::wallet_internal_error,
-              "Known ring does not include the spent output: " + std::to_string(td.m_global_output_index));
+          if (!own_found)
+          {
+            MWARNING("Known ring does not include the spent output: " + std::to_string(td.m_global_output_index)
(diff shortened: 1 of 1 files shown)


j-berman:
Repeating more context from https://github.com/monero-project/monero/pull/10085#issuecomment-3303843986:

> the wallet saves rings used in prior constructed txs, so that it will attempt to reuse the same ring if spending the same output (thus avoiding revealing which output is the real spend in the ring). That ring saves references to ring members by global output ID.

> A reorg can modify global output ID's tied to members of a ring, including the output spent. The logic inside get_outs that reads saved rings does not account for the potential of the spent ring member's output id to be modified by the reorg [here](https://github.com/monero-project/monero/blob/8d4c625713e3419573dfcc7119c8848f47cabbaa/src/wallet/wallet2.cpp#L9276-L9304), and thus the wallet may be unable to spend an output in that case unless the user explicitly calls unset_ring, which could reveal to the network which output they are spending.

> The simplest solution to that is probably to remove [this throw](https://github.com/monero-project/monero/blob/8d4c625713e3419573dfcc7119c8848f47cabbaa/src/wallet/wallet2.cpp#L9303-L9304), and log a warning in that case instead.

selsta:
@j-berman please also open against release branch

jeffro256 (approved):

PR #10100, opened 2025-09-22
tests: only enable fuzz testing if `ENABLE_FUZZ_TEST` defined

0xFFFC0000:


selsta:
I think we should enable `ENABLE_FUZZ_TEST` on CI at least on one Linux job to test if it compiles.

0xFFFC0000:
> I think we should enable `ENABLE_FUZZ_TEST` on CI at least on one Linux job to test if it compiles.

Great idea. I will add a CI job to compile fuzz build.

selsta (approved):

0xFFFC0000:
CI failure is not related to PR: `node_server.race_condition`

tobtoht (approved):

PR #10101, opened 2025-09-22
Add thread acquire+release for ZeroMQ msg calls

vtnerd:
I ran thread-sanitizer on `monero-lws` unit tests and the initial results were horrifying.

After digging into the report, I determined the primary cause to be internal to ZeroMQ. The library has a lock-free queue that uses atomics, but no explicit memory fence for the data is provided. Theoretically, on probably only alpha cpus via dependent load reordering, this could cause memory issues.

Additionally, the C++ memory fencing utilities are not supported by TSAN! So, we have to manually call into the library with a memory address for tracking.

NOTE: Yes, ZeroMQ on my system was **not** compiled with TSAN magic, however, I believe there to be an issue after looking at the source.

@SChernykh  @tevador 

I should possibly upstream this to ZeroMQ, but I expect to be controversial thus the change here.


SChernykh:
Even the tsan wiki suggests to suppress libzmq.so: https://github.com/google/sanitizers/wiki/ThreadSanitizerSuppressions

So it's a known issue, I guess it's fine to ignore it?

vtnerd:
In this case, the memory is allocated+written+read in our codebase (typically via `byte_slice`). The last write occurs in some arbitrary serialization function, and the first read occurs in our `net::zmq::receive` function. `libzmq.so` is never mentioned in the call stack. Suppressing the `receive` function may be sufficient; instead this PR calls the raw tracking functions provided by TSAN.

I also added a "real" memory fence, which may be overkill. My understanding of the C++ atomic memory model is that this likely only to be useful on Alpha CPUs.

vtnerd:
@SChernykh  : I re-read the C++ spec, and I was incorrect before. The additional memory barriers were not needed, so the force push includes just tsan tracking for memory given/read from zmq. I still like this better than a suppression file as it helped confirm the exact complaint by tsan.

----------

Previously I thought the memory fencing applied only to the specific memory addresses in the atomic operation, but it appears the spec indicates that unrelated reads/writes are affected too. Thank goodness. I knew this at some point, then forgot. Or never knew.

PR #10108, opened 2025-09-25
RingCT crypto: 6x faster zero commit

j-berman:
Idea by @kayabaNerve to speed up transparent amount commitments by ~6x

From Cypher Stack, [Section 3](https://github.com/cypherstack/fcmp-impl-audit/blob/main/FCMP_Code_Implementation_Audit.pdf) ([commit](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf)):

> All functions used for the Faster Transparent Amount Commitments are described correctly
mathematically, and implemented according to that description. As such, it is our conclusion
that the above functions are fit for deployment in Monero.

Trail of Bits also reviewed this PR and found no issues. Will link the report here once it's public.

### DIFF
--- a/src/ringct/rctOps.cpp
+++ b/src/ringct/rctOps.cpp
@@ -29,6 +29,7 @@
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <boost/lexical_cast.hpp>
+#include "crypto/generators.h"
 #include "misc_log_ex.h"
 #include "rctOps.h"
 using namespace crypto;
@@ -216,6 +217,33 @@ static const zero_commitment zero_commitments[] = {
   { (uint64_t)10000000000000000000ull, {{0x65, 0x8d, 0x1, 0x37, 0x6d, 0x18, 0x63, 0xe7, 0x7b, 0x9, 0x6f, 0x98, 0xe6, 0xe5, 0x13, 0xc2, 0x4, 0x10, 0xf5, 0xc7, 0xfb, 0x18, 0xa6, 0xe5, 0x9a, 0x52, 0x66, 0x84, 0x5c, 0xd9, 0xb1, 0xe3}} },
 };
 
+static constexpr std::size_t H_TABLE_SIZE = 64;
+const std::vector<ge_cached>& H_TABLE()
+{
+    struct static_h_table
+    {
+        std::vector<ge_cached> h_table;
+        static_h_table()
+            : h_table()
+        {
+            h_table.resize(H_TABLE_SIZE);
+
+            ge_p3_to_cached(&h_table.at(0), &ge_p3_H);
+            ge_p3 H_bit_p3 = ge_p3_H;
+
+            for (std::size_t i = 1; i < H_TABLE_SIZE; ++i)
+            {
+                ge_p1p1 H_bit_p1p1;
+                ge_p3_dbl(&H_bit_p1p1, &H_bit_p3);
+                ge_p1p1_to_p3(&H_bit_p3, &H_bit_p1p1);
+                ge_p3_to_cached(&h_table.at(i), &H_bit_p3);
+            }
+        }
+    };
+    static const static_h_table out;
+    return out.h_table;
+}
+
 namespace rct {
 
     //Various key initialization functions
@@ -318,7 +346,7 @@ namespace rct {
         return make_tuple(sk, pk);
     }
     
-    key zeroCommit(xmr_amount amount) {
+    key zeroCommitVartime(xmr_amount amount) {
         const zero_commitment *begin = zero_commitments;
         const zero_commitment *end = zero_commitments + sizeof(zero_commitments) / sizeof(zero_commitments[0]);
         const zero_commitment value{amount, rct::zero()};
@@ -327,9 +355,20 @@ namespace rct {
         {
             return it->commitment;
         }
-        key am = d2h(amount);
-        key bH = scalarmultH(am);
-        return addKeys(G, bH);
+        ge_p3 res_ge_p3 = get_G_p3();
+        static_assert(sizeof(xmr_amount) * 8 == H_TABLE_SIZE, "unexpected size of h table");
+        for (size_t i = 0; i < H_TABLE_SIZE; ++i)
+        {
+            if (amount & (xmr_amount(1) << i))
+            {
+                ge_p1p1 p1p1;
+                ge_add(&p1p1, &res_ge_p3, &H_TABLE()[i]);
+                ge_p1p1_to_p3(&res_ge_p3, &p1p1);
+            }
+        }
+        rct::key res;
+        ge_p3_tobytes(res.bytes, &res_ge_p3);
+        return res;
     }
 
     key commit(xmr_amount amount, const key &mask) {
--- a/src/blockchain_db/blockchain_db.cpp
+++ b/src/blockchain_db/blockchain_db.cpp
@@ -233,7 +233,7 @@ void BlockchainDB::add_transaction(const crypto::hash& blk_hash, const transacti
     if (miner_tx && tx.version == 2)
     {
       cryptonote::tx_out vout = tx.vout[i];
-      rct::key commitment = rct::zeroCommit(vout.amount);
+      rct::key commitment = rct::zeroCommitVartime(vout.amount);
       vout.amount = 0;
       amount_output_indices[i] = add_output(tx_hash, vout, i, tx.unlock_time,
         &commitment);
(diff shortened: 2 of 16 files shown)


src/ringct/rctOps.cpp
```diff
@@ -216,6 +216,41 @@ static const zero_commitment zero_commitments[] = {
   { (uint64_t)10000000000000000000ull, {{0x65, 0x8d, 0x1, 0x37, 0x6d, 0x18, 0x63, 0xe7, 0x7b, 0x9, 0x6f, 0x98, 0xe6, 0xe5, 0x13, 0xc2, 0x4, 0x10, 0xf5, 0xc7, 0xfb, 0x18, 0xa6, 0xe5, 0x9a, 0x52, 0x66, 0x84, 0x5c, 0xd9, 0xb1, 0xe3}} },
 };
 
+static const std::size_t H_TABLE_SIZE = 64;
```
vtnerd: mark `constexpr`.

src/ringct/rctOps.cpp
```diff
@@ -216,6 +216,41 @@ static const zero_commitment zero_commitments[] = {
...
+static std::vector<ge_cached> INIT_H_TABLE()
```
vtnerd: I think the best to use a function local static as it prevents the static initialization fiasco:

```c++
const std::vector<ge_cached>& H_TABLE()
{
  struct table
  {
    std::vector<ge_cached> data;

    table()
      : data()
    {
       // initialize data here
    }
  };
  static const table out;
  return out.data;
}
```

if both of these functions use that approach, zeroCommitVartime could be called before main (during static initialization).

vtnerd (changes requested):

src/ringct/rctOps.cpp
```diff
@@ -362,14 +374,14 @@ namespace rct {
         {
             return it->commitment;
         }
-        ge_p3 res_ge_p3 = G_GE_P3;
+        ge_p3 res_ge_p3 = G_GE_P3();
         static_assert(sizeof(xmr_amount) * 8 == H_TABLE_SIZE, "unexpected size of h table");
         for (size_t i = 0; i < H_TABLE_SIZE; ++i)
         {
             if (amount & (1UL << i))
```
jeffro256: Make sure to add https://github.com/j-berman/monero/commit/a3cf0d708146b0c21a940f6fd64dc9abb231022a here. (Otherwise PR is incorrect)

j-berman:
I'm thinking it would make sense to have a batch of FCMP++-adjacent crypto changes audited before merging, including this PR, #10111, and forthcoming FCMP++ crypto. Reviews pre-audit is nice too. I was thinking this even before [this issue](https://github.com/monero-project/monero/pull/10108/commits/50b825957eb3083f84f40793f67c47ccd8404365), and I figure that issue is strong reason to support an audit on its own.

jeffro256:
```diff
diff --git a/src/crypto/crypto-ops.c b/src/crypto/crypto-ops.c
index 314fe448a..ec7309822 100644
--- a/src/crypto/crypto-ops.c
+++ b/src/crypto/crypto-ops.c
@@ -42,7 +42,6 @@ static void fe_sq(fe, const fe);
 static void ge_madd(ge_p1p1 *, const ge_p3 *, const ge_precomp *);
 static void ge_msub(ge_p1p1 *, const ge_p3 *, const ge_precomp *);
 static void ge_p2_0(ge_p2 *);
-static void ge_p3_dbl(ge_p1p1 *, const ge_p3 *);
 static void fe_divpowm1(fe, const fe, const fe);
 
 /* Common functions */
@@ -1529,7 +1528,7 @@ static void ge_p3_0(ge_p3 *h) {
 r = 2 * p
 */
 
-static void ge_p3_dbl(ge_p1p1 *r, const ge_p3 *p) {
+void ge_p3_dbl(ge_p1p1 *r, const ge_p3 *p) {
   ge_p2 q;
   ge_p3_to_p2(&q, p);
   ge_p2_dbl(r, &q);
diff --git a/src/crypto/crypto-ops.h b/src/crypto/crypto-ops.h
index c103f1f78..d5f8f20a8 100644
--- a/src/crypto/crypto-ops.h
+++ b/src/crypto/crypto-ops.h
@@ -102,6 +102,11 @@ void ge_p1p1_to_p3(ge_p3 *, const ge_p1p1 *);
 
 void ge_p2_dbl(ge_p1p1 *, const ge_p2 *);
 
+/* From ge_p3_dbl.c */
+
+void ge_p3_dbl(ge_p1p1 *r, const ge_p3 *p);
+
+
 /* From ge_p3_to_cached.c */
 
 extern const fe fe_d2;
diff --git a/src/ringct/rctOps.cpp b/src/ringct/rctOps.cpp
index c6b86bfb1..5cdd64398 100644
--- a/src/ringct/rctOps.cpp
+++ b/src/ringct/rctOps.cpp
@@ -225,22 +225,17 @@ const std::vector<ge_cached>& H_TABLE()
         static_h_table()
             : h_table()
         {
-            std::vector<ge_p3> h_table_ge_p3;
-
-            h_table_ge_p3.resize(H_TABLE_SIZE);
             h_table.resize(H_TABLE_SIZE);
 
-            h_table_ge_p3.at(0) = ge_p3_H;
-            ge_p3_to_cached(&h_table.at(0), &h_table_ge_p3.at(0));
+            ge_p3_to_cached(&h_table.at(0), &ge_p3_H);
+            ge_p3 H_bit_p3 = ge_p3_H;
 
             for (std::size_t i = 1; i < H_TABLE_SIZE; ++i)
             {
-                ge_p2 p2;
-                ge_p3_to_p2(&p2, &h_table_ge_p3.at(i - 1));
-                ge_p1p1 p2_dbl;
-                ge_p2_dbl(&p2_dbl, &p2);
-                ge_p1p1_to_p3(&h_table_ge_p3.at(i), &p2_dbl);
-                ge_p3_to_cached(&h_table.at(i), &h_table_ge_p3.at(i));
+                ge_p1p1 H_bit_p1p1;
+                ge_p3_dbl(&H_bit_p1p1, &H_bit_p3);
+                ge_p1p1_to_p3(&H_bit_p3, &H_bit_p1p1);
+                ge_p3_to_cached(&h_table.at(i), &H_bit_p3);
             }
         }
     };
```

This change removes the `ge_p3` vector in `static_h_table`

src/ringct/rctOps.cpp
```diff
@@ -216,6 +216,48 @@ static const zero_commitment zero_commitments[] = {
...
+    {
+        ge_p3 g_ge_p3;
+        static_g_ge_pe3()
+        {
+            CHECK_AND_ASSERT_THROW_MES(ge_frombytes_vartime(&g_ge_p3, rct::G.bytes) == 0, "ge_frombytes_vartime failed for G");
+        }
+    };
+    static const static_g_ge_pe3 out;
+    return out.g_ge_p3;
+}
```
jeffro256: This could probably be replaced with `crypto::get_G_p3()` in [crypto/generators](https://github.com/monero-project/monero/blob/74dd968cd9ee5a2cad6255ac52f815eacf859948/src/crypto/generators.h#L42).

jeffro256 (approved):

selsta:
please squash

j-berman:
Please do not merge until audited

j-berman:
This PR has been audited by both Cypher Stack and Trail of Bits. Linked the Cypher Stack audit in the description, and will link the Trail of Bits audit once it's public.

UkoeHB (approved):

src/ringct/rctOps.cpp
```diff
@@ -327,9 +355,20 @@ namespace rct {
...
-        return addKeys(G, bH);
+        ge_p3 res_ge_p3 = get_G_p3();
+        static_assert(sizeof(xmr_amount) * 8 == H_TABLE_SIZE, "unexpected size of h table");
+        for (size_t i = 0; i < H_TABLE_SIZE; ++i)
+        {
+            if (amount & (xmr_amount(1) << i))
+            {
+                ge_p1p1 p1p1;
+                ge_add(&p1p1, &res_ge_p3, &H_TABLE()[i]);
+                ge_p1p1_to_p3(&res_ge_p3, &p1p1);
```
jeffro256, line 366: You could save one `fe_mul()` by using `ge_p1p1_to_p2()` on the last bit instead of `ge_p1p1_to_p3()`. I think that that saves <0.2% of run-time in the worst case. Probably not worth it.

jeffro256 (approved):

src/ringct/rctOps.cpp
j-berman: Thinking I'll keep this as is for cleanliness

PR #10111, opened 2025-09-26
crypto: fast `fe_batch_invert` using Montgomery's trick

j-berman:
Idea by @kayabaNerve to speed up field element inversions when we need to do multiple inversions at one time. This is a significant perf optimization used in FCMP++ tree building.

The perf test shows that batch inverting 1000 elements is ~98% faster than inverting each one individually.

This PR also includes a new `fe_equals` function.

An earlier state of this PR was audited by both Cypher Stack and Trail of Bits:
- Both suggested hardening the constant time `fe_equals`.
  - Cypher Stack section 4.2.2
  - Remediated by reusing libsodium's constant time equals impl (suggested by @jeffro256)
- Both identified that if `fe_batch_invert` is passed a 0, it could return 0, which would be undesirable.
  - Cypher Stack section 4.3.2
  - Remediated by having `fe_batch_invert` return -1 if a 0 is passed in, and 0 on success.
- Cypher Stack identified `n` could be negative.
  - Section 4.3.2
  - Remediated by using an `unsigned int` for `n`.

[Cypher Stack audit](https://github.com/cypherstack/fcmp-impl-audit/blob/main/FCMP_Code_Implementation_Audit.pdf) ([commit](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf))
Will post the Trail of Bits audit once public. Trail of Bits also reviewed the remediations.

The remediation changes were implemented [here](https://github.com/j-berman/monero/pull/18).

### DIFF
--- a/src/crypto/crypto-ops.c
+++ b/src/crypto/crypto-ops.c
@@ -29,6 +29,7 @@
 // Parts of this file are originally copyright (c) 2012-2013 The Cryptonote developers
 
 #include <assert.h>
+#include <sodium/crypto_verify_32.h>
 #include <stdint.h>
 
 #include "warnings.h"
@@ -313,6 +314,61 @@ void fe_invert(fe out, const fe z) {
   return;
 }
 
+// return 1 if a == b, else 0
+int fe_equals(const fe a, const fe b) {
+  unsigned char a_bytes[32];
+  unsigned char b_bytes[32];
+  fe_tobytes(a_bytes, a);
+  fe_tobytes(b_bytes, b);
+  return crypto_verify_32(a_bytes, b_bytes) == 0;
+}
+
+/* From fe_isnonzero.c, modified */
+
+static int fe_isnonzero(const fe f) {
+  unsigned char s[32];
+  fe_tobytes(s, f);
+  return (((int) (s[0] | s[1] | s[2] | s[3] | s[4] | s[5] | s[6] | s[7] | s[8] |
+    s[9] | s[10] | s[11] | s[12] | s[13] | s[14] | s[15] | s[16] | s[17] |
+    s[18] | s[19] | s[20] | s[21] | s[22] | s[23] | s[24] | s[25] | s[26] |
+    s[27] | s[28] | s[29] | s[30] | s[31]) - 1) >> 8) + 1;
+}
+
+// Montgomery's trick
+// https://iacr.org/archive/pkc2004/29470042/29470042.pdf 2.2
+int fe_batch_invert(fe* __restrict out, const fe* __restrict in, const unsigned int n) {
+  if (n == 0) {
+    return 0;
+  }
+
+  assert(out);
+  assert(in);
+  assert(in != out); // also should not overlap
+
+  // Step 1: collect initial muls
+  fe_copy(out[0], in[0]);
+  for (unsigned int i = 1; i < n; ++i) {
+    fe_mul(out[i], out[i-1], in[i]);
+  }
+
+  // Step 2: get the inverse of all elems multiplied together
+  if (!fe_isnonzero(out[n-1])) {
+    // Don't divide by 0
+    return -1;
+  }
+  fe a;
+  fe_invert(a, out[n-1]);
+
+  // Step 3: get each inverse
+  for (unsigned int i = n; i > 1; --i) {
+    fe_mul(out[i-1], a, out[i-2]);
+    fe_mul(a, a, in[i-1]);
+  }
+  fe_copy(out[0], a);
+
+  return 0;
+}
+
 /* From fe_isnegative.c */
 
 /*
@@ -327,17 +383,6 @@ static int fe_isnegative(const fe f) {
   unsigned char s[32];
   fe_tobytes(s, f);
   return s[0] & 1;
-}
-
-/* From fe_isnonzero.c, modified */
-
-static int fe_isnonzero(const fe f) {
-  unsigned char s[32];
-  fe_tobytes(s, f);
-  return (((int) (s[0] | s[1] | s[2] | s[3] | s[4] | s[5] | s[6] | s[7] | s[8] |
-    s[9] | s[10] | s[11] | s[12] | s[13] | s[14] | s[15] | s[16] | s[17] |
-    s[18] | s[19] | s[20] | s[21] | s[22] | s[23] | s[24] | s[25] | s[26] |
-    s[27] | s[28] | s[29] | s[30] | s[31]) - 1) >> 8) + 1;
 }
 
 /* From fe_mul.c */
@@ -1328,16 +1373,9 @@ void ge_double_scalarmult_base_vartime_p3(ge_p3 *r3, const unsigned char *a, con
   }
 }
 
-/* From ge_frombytes.c, modified */
+/* From fe_frombytes.c */
 
-int ge_frombytes_vartime(ge_p3 *h, const unsigned char *s) {
-  fe u;
-  fe v;
-  fe vxx;
-  fe check;
-
-  /* From fe_frombytes.c */
-
+int fe_frombytes_vartime(fe y, const unsigned char *s) {
   int64_t h0 = load_4(s);
   int64_t h1 = load_3(s + 4) << 6;
   int64_t h2 = load_3(s + 7) << 5;
@@ -1378,18 +1416,31 @@ int ge_frombytes_vartime(ge_p3 *h, const unsigned char *s) {
   carry6 = (h6 + (int64_t) (1<<25)) >> 26; h7 += carry6; h6 -= carry6 << 26;
   carry8 = (h8 + (int64_t) (1<<25)) >> 26; h9 += carry8; h8 -= carry8 << 26;
 
-  h->Y[0] = h0;
-  h->Y[1] = h1;
-  h->Y[2] = h2;
-  h->Y[3] = h3;
-  h->Y[4] = h4;
-  h->Y[5] = h5;
-  h->Y[6] = h6;
-  h->Y[7] = h7;
-  h->Y[8] = h8;
-  h->Y[9] = h9;
+  y[0] = h0;
+  y[1] = h1;
+  y[2] = h2;
+  y[3] = h3;
+  y[4] = h4;
+  y[5] = h5;
+  y[6] = h6;
+  y[7] = h7;
+  y[8] = h8;
+  y[9] = h9;
 
-  /* End fe_frombytes.c */
+  return 0;
+}
+
+/* From ge_frombytes.c, modified */
+
+int ge_frombytes_vartime(ge_p3 *h, const unsigned char *s) {
+  fe u;
+  fe v;
+  fe vxx;
+  fe check;
+
+  if (fe_frombytes_vartime(h->Y, s) != 0) {
+    return -1;
+  }
 
   fe_1(h->Z);
   fe_sq(u, h->Y);
--- a/src/crypto/crypto-ops.h
+++ b/src/crypto/crypto-ops.h
@@ -88,6 +88,7 @@ void ge_double_scalarmult_base_vartime_p3(ge_p3 *, const unsigned char *, const 
 
 extern const fe fe_sqrtm1;
 extern const fe fe_d;
+int fe_frombytes_vartime(fe, const unsigned char *);
 int ge_frombytes_vartime(ge_p3 *, const unsigned char *);
 
 /* From ge_p1p1_to_p2.c */
@@ -164,6 +165,17 @@ void ge_sub(ge_p1p1 *r, const ge_p3 *p, const ge_cached *q);
 void fe_add(fe h, const fe f, const fe g);
 void fe_tobytes(unsigned char *, const fe);
 void fe_invert(fe out, const fe z);
+int fe_equals(const fe a, const fe b);
+/**
+@brief: out[i] = 1/in[i] for i in [0, n). Uses Montgomery's trick
+@return: 0 on success, some other value otherwise
+
+Unlike other crypto functions, `out` and `in` memory sections CANNOT be aliased.
+If `out` and `in` overlap, it will cause undefined output.
+
+No 0 fe's are expected for `in`, otherwise fails.
+**/
+int fe_batch_invert(fe* __restrict out, const fe* __restrict in, const unsigned int n);
 void fe_mul(fe out, const fe, const fe);
 void fe_0(fe h);
 
--- a/tests/performance_tests/fe_batch_invert.h
+++ b/tests/performance_tests/fe_batch_invert.h
@@ -1,0 +1,79 @@
+// Copyright (c) 2025, The Monero Project
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+// 
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+// 
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+// 
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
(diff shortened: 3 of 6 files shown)


jeffro256 (changes requested):
Here's a change to avoid allocations:

```diff
diff --git a/src/crypto/crypto-ops.c b/src/crypto/crypto-ops.c
index d5db2481e..6205d19e8 100644
--- a/src/crypto/crypto-ops.c
+++ b/src/crypto/crypto-ops.c
@@ -30,7 +30,6 @@
 
 #include <assert.h>
 #include <stdint.h>
-#include <stdlib.h>
 
 #include "warnings.h"
 #include "crypto-ops.h"
@@ -322,28 +321,22 @@ int fe_batch_invert(fe *out, const fe *in, const int n) {
   }
 
   // Step 1: collect initial muls
-  fe *init_muls = (fe *) malloc(n * sizeof(fe));
-  if (!init_muls) {
-    return 1;
-  }
-  fe_copy(init_muls[0], in[0]);
+  fe_copy(out[0], in[0]);
   for (int i = 1; i < n; ++i) {
-    fe_mul(init_muls[i], init_muls[i-1], in[i]);
+    fe_mul(out[i], out[i-1], in[i]);
   }
 
   // Step 2: get the inverse of all elems multiplied together
   fe a;
-  fe_invert(a, init_muls[n-1]);
+  fe_invert(a, out[n-1]);
 
   // Step 3: get each inverse
   for (int i = n; i > 1; --i) {
-    fe_mul(out[i-1], a, init_muls[i-2]);
+    fe_mul(out[i-1], a, out[i-2]);
     fe_mul(a, a, in[i-1]);
   }
   fe_copy(out[0], a);
 
-  free(init_muls);
-
   return 0;
 }
 

```

src/crypto/crypto-ops.c
```diff
@@ -313,6 +313,33 @@ void fe_invert(fe out, const fe z) {
   return;
 }
 
+// Montgomery's trick
+// https://iacr.org/archive/pkc2004/29470042/29470042.pdf 2.2
+int fe_batch_invert(fe *out, const fe *in, const int n) {
+  if (n == 0) {
```
vtnerd, line 340: Should the function be labeled `vartime` due to this branch?

tests/performance_tests/fe_batch_invert.h
```diff
@@ -0,0 +1,79 @@
...
+template<bool batched>
+class test_fe_batch_invert
+{
+public:
+  static const size_t loop_count = 50;
+  static const size_t n_elems = 1000;
+
+  bool init()
+  {
+    m_fes = (fe *) malloc(n_elems * sizeof(fe));
```
vtnerd: Check for `nullptr`? Could easily use `std::vector` ...?

tests/performance_tests/fe_batch_invert.h
```diff
@@ -0,0 +1,79 @@
...
+
+      memcpy(m_fes[i], &point.Y, sizeof(fe));
+    }
+
+    return true;
+  }
+
+  bool test()
+  {
+    fe *inv_fes = (fe *) malloc(n_elems * sizeof(fe));
```
vtnerd: Same here.

tests/unit_tests/crypto.cpp
```diff
@@ -345,3 +345,39 @@ TEST(Crypto, generator_consistency)
...
+TEST(Crypto, batch_inversion)
+{
+  const std::size_t MAX_TEST_ELEMS = 1000;
+
+  // Memory allocator
+  auto alloc = [](const std::size_t n) -> fe*
+  {
+    fe *ptr = (fe *) malloc(n * sizeof(fe));
+    if (!ptr)
+      throw std::runtime_error("failed to malloc fe *");
```
vtnerd: `throw std::bad_alloc` is probably more appropriate, as this has to allocate memory for the `std::string` after just failing to `malloc`.

Again, not sure why `std::vector` isn't used instead.

vtnerd (changes requested):

src/crypto/crypto-ops.c
kayabaNerve: Time variable to the width, which generally isn't a secret, shouldn't be an issue. You'd have to argue that the amount of inputs/amounts is a secret and should be constant-time to them accordingly (or whatever defines the width here).

tests/unit_tests/crypto.cpp
```diff
@@ -345,3 +345,39 @@ TEST(Crypto, generator_consistency)
...
+    ASSERT_EQ(fe_frombytes_vartime(init_elems[i], (unsigned char*)kp.pub.data), 0);
+    fe_invert(norm_inverted[i], init_elems[i]);
+  }
+
+  // Do batch inversions and compare to individual inversions
+  for (std::size_t n_elems = 1; n_elems <= MAX_TEST_ELEMS; ++n_elems)
+  {
+    fe *batch_inverted = alloc(n_elems);
+    ASSERT_EQ(fe_batch_invert(batch_inverted, init_elems, n_elems), 0);
+    ASSERT_EQ(memcmp(batch_inverted, norm_inverted, n_elems * sizeof(fe)), 0);
```
jeffro256: It's worth noting that in-memory representations of `fe` might not be equal even if the `fe_tobyte()` canonical representations would be the same. They happen to be in this case, and it might be worth keeping this test as-is? But it's conceivable that some correct change to e.g. `fe_mul()` down the line might cause this line to fail. I would recommend keeping this line, but documenting it in case it fails later.

kayabaNerve: ref10 in general doesn't guarantee they're reduced. That caused annoyances with the FCMP++ code...

(Adding on to what you said, as you identified and raised)

tests/performance_tests/fe_batch_invert.h
j-berman: Used a vector instead of a pointer in the latest. I think I did that originally and then changed it, I don't remember why. Good spot

tests/unit_tests/crypto.cpp
j-berman: Added a warning in the latest, also linking to an `fe_reduce` implementation that could be used to update the test

tests/performance_tests/fe_batch_invert.h
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    return true;
+  }
+
+  bool test()
+  {
+    std::vector<fe> inv_fes(n_elems);
+
+    if (batched)
```
jeffro256:
```suggestion
    if constexpr (batched)
```

For good measure, to help compiler optimize the branch out, and to signal to reader that that is what it's doing

tests/performance_tests/fe_batch_invert.h
j-berman: CI failure reminded me why I changed it from using a vector...

```
/Applications/Xcode_16.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/__memory/allocator.h:168:84: error: object expression of non-scalar type 'int[10]' cannot be used in a pseudo-destructor expression
  168 |   _LIBCPP_DEPRECATED_IN_CXX17 _LIBCPP_HIDE_FROM_ABI void destroy(pointer __p) { __p->~_Tp(); }
```

I figured using a raw pointer was the simplest way to deal with this. Do you see a better way?

j-berman: I use this pattern in FCMP++ integration code, not just in tests here, so it would definitely be good to improve if you see a better way

j-berman: Reverted to using a raw pointer in the latest. Open to a better approach

vtnerd: Oops.

I think the only way is to wrap in another object.

vtnerd: `std::unique_ptr` mostly works here, but you have to match it with the correct allocation call. (`new []` array).

jeffro256 (approved):

tests/performance_tests/fe_batch_invert.h
vtnerd: >I use this pattern in FCMP++ integration code, not just in tests here, so it would definitely be good to improve if you see a better way.

Hopefully the code doesn't leak on exceptions (like these tests kind of do).

j-berman: How does the latest look? I'd update the integration code with that pattern too (it does currently leak on exceptions)

j-berman:
Test failure is unrelated

tests/unit_tests/crypto.cpp
j-berman: I implemented an `fe_equals` over here: https://github.com/j-berman/monero/commit/2a56788b8be63b9593a43574cd425b01dec72df1

Will make a follow-up PR to this one

jeffro256 (approved):

selsta:
@vtnerd could you do another review / give a final approval?

src/crypto/crypto-ops.c
```diff
@@ -313,6 +313,33 @@ void fe_invert(fe out, const fe z) {
...
+  }
+
+  // Step 2: get the inverse of all elems multiplied together
+  fe a;
+  fe_invert(a, out[n-1]);
+
+  // Step 3: get each inverse
+  for (int i = n; i > 1; --i) {
+    fe_mul(out[i-1], a, out[i-2]);
+    fe_mul(a, a, in[i-1]);
```
vtnerd, line 365: I think this line right here prevents aliasing on in+out. Should a runtime check be added? Or at least an assert? I recall the other crypto functions being resilient against aliasing.

vtnerd (changes requested):
Another small question.

src/crypto/crypto-ops.c
jeffro256: This is a good point.

jeffro256: I would recommend documenting it in a comment in the header file, plus adding an `assert()` statement that the memory spaces aren't overlapping.

tests/unit_tests/crypto.cpp
```diff
@@ -345,3 +345,30 @@ TEST(Crypto, generator_consistency)
...
+{
+  const std::size_t MAX_TEST_ELEMS = 1000;
+
+  std::unique_ptr<fe[]> init_elems = std::make_unique<fe[]>(MAX_TEST_ELEMS);
+  std::unique_ptr<fe[]> norm_inverted = std::make_unique<fe[]>(MAX_TEST_ELEMS);
+
+  // Init test elems and individual inversions
+  for (std::size_t i = 0; i < MAX_TEST_ELEMS; ++i)
+  {
+    const cryptonote::keypair kp = cryptonote::keypair::generate(hw::get_device("default"));
```
jeffro256: Another note: this only tests field elements mod $l$, not mod $q$. Since $l < q$, this should always be complete, but you're missing out on testing field elements in range $[l, q)$.

jeffro256: You could randomly generate 32 bytes and clear the top bit. The probability that you generate an integer greater than or equal to $q$ is $19 / 2^{255}$, which is negligible if our random number generator is working correctly.

jeffro256: Oh nevermind actually, you're using the compressed Y value of the point, not the secret key to that point.

src/crypto/crypto-ops.c
jeffro256: You could apply this diff:

```diff
diff --git a/src/crypto/crypto-ops.c b/src/crypto/crypto-ops.c
index ccb59de3f..5c003fef3 100644
--- a/src/crypto/crypto-ops.c
+++ b/src/crypto/crypto-ops.c
@@ -320,6 +320,17 @@ int fe_batch_invert(fe *out, const fe *in, const int n) {
     return 0;
   }
 
+  assert(out);
+  assert(in);
+#ifndef NDEBUG
+  {
+    // Overlap of `out` and `in` sections not allowed
+    const fe *pmin = out < in ? out : in;
+    const fe *pmax = out > in ? out : in;
+    assert(pmin + n < pmax);
+  }
+#endif
+
   // Step 1: collect initial muls
   fe_copy(out[0], in[0]);
   for (int i = 1; i < n; ++i) {
diff --git a/src/crypto/crypto-ops.h b/src/crypto/crypto-ops.h
index 6488fd99d..797637c33 100644
--- a/src/crypto/crypto-ops.h
+++ b/src/crypto/crypto-ops.h
@@ -164,6 +164,13 @@ void ge_sub(ge_p1p1 *r, const ge_p3 *p, const ge_cached *q);
 void fe_add(fe h, const fe f, const fe g);
 void fe_tobytes(unsigned char *, const fe);
 void fe_invert(fe out, const fe z);
+/**
+@brief: out[i] = 1/in[i] for i in [0, n). Uses Montgomery's trick
+@return: 0 on success, some other value otherwise
+
+Unlike other crypto functions, `out` and `in` memory sections CANNOT be aliased.
+If `out` and `in` overlap, it will cause undefined output.
+**/
 int fe_batch_invert(fe *out, const fe *in, const int n);
 void fe_mul(fe out, const fe, const fe);
 void fe_0(fe h);

```

src/crypto/crypto-ops.h
```diff
@@ -163,6 +164,7 @@ void ge_sub(ge_p1p1 *r, const ge_p3 *p, const ge_cached *q);
 void fe_add(fe h, const fe f, const fe g);
 void fe_tobytes(unsigned char *, const fe);
 void fe_invert(fe out, const fe z);
+int fe_batch_invert(fe *out, const fe *in, const int n);
```
jeffro256: Returns an `int` even though only value possibly returned is `0`.

src/crypto/crypto-ops.c
j-berman: Applied that diff in the latest

tests/unit_tests/crypto.cpp
j-berman: I went ahead and implemented `fe_equals` in this PR

jeffro256:
Should rebase and try compiling with latest master. Fails due to missing header. Either should include `cryptonote_basic/cryptonote_basic.h` in `tests/unit_tests/crypto.cpp`, or don't use `cryptonote::keypair`. I recommend the latter, but am fine with either.

src/crypto/crypto-ops.c
j-berman: @UkoeHB reported `assert(pmin + n < pmax)` is failing on mac. It's apparently UB to use comparison operators on pointers to **different** arrays (see "Built-in pointer relational comparison" in https://en.cppreference.com/cpp/language/operator_comparison).

I'm inclined to just do a simple `assert(out != in)`.

@UkoeHB also notes that "Other crypto ops 'allow' aliasing but I'm not sure if they are aliasing-safe or not", which I think is a fair statement (I also haven't fully 100% checked every function to see if aliasing safe). So I'd update the comment to just say that this function in particular does not support aliasing, rather than make an assertion about the other crypto functions.

src/crypto/crypto-ops.c
```diff
@@ -313,6 +313,56 @@ void fe_invert(fe out, const fe z) {
...
+  }
+
+  assert(out);
+  assert(in);
+#ifndef NDEBUG
+  {
+    // Overlap of `out` and `in` sections not allowed
+    const fe *pmin = out < in ? out : in;
+    const fe *pmax = out > in ? out : in;
+    assert(pmin + n < pmax);
```
jeffro256:
```suggestion
    assert(pmin + n <= pmax);
```

jeffro256: Nvm see this PR for full change: https://github.com/seraphis-migration/monero/pull/412

UkoeHB (approved):

j-berman:
Audited by both Cypher Stack and Trail of Bits. The description highlights changes implemented as a result of the audits.

UkoeHB (approved):

jeffro256 (approved):

PR #10114, opened 2025-09-26
net_node: add hinto seed nodes

hinto-janai:


nahuhh:
What is this onion pointed to? 18080 looks wrong

hinto-janai:
It is reachable, although I'll change it to `18083`.

nahuhh:
Just making sure - it is anonymous-inbound, right?

hinto-janai:
Yes.

plowsof:
not a good start @hinto-janai 

https://github.com/plowsof/check-monero-seed-nodes

```
./monerod --add-exclusive-node 5.104.84.64:28080 --testnet --log-level 2 
2025-10-03 09:27:27.623	I [5.104.84.64:28080 OUT] [priority]Failed to HANDSHAKE with peer 5.104.84.64:28080
./monerod --add-exclusive-node 5.104.84.64:18080 --log-level 2
2025-10-03 09:26:47.069	I 0[priority]Connect failed to 5.104.84.64:18080
./monerod --add-exclusive-node 5.104.84.64:38080 --stagenet --log-level 2  
2025-10-03 09:28:03.872	I [5.104.84.64:38080 OUT] [priority]Failed to HANDSHAKE with peer 5.104.84.64:38080
```

hinto-janai:
Noted, I should mention I've been testing how many services can be ran (multiple instances of `cuprated`, `monerod`, `p2pool`, FCMP alpha node, misc services) without OOMs. It seems like I hit a limit for 8GB, I'll start trimming things to prioritize the seed nodes.

plowsof:
we can revert this then. we need stable seed nodes, preferably with more than 8GB of ram available

hinto-janai:
Other processes will be stopped for the seed nodes to be stable, if that proves to be unstable then they can be removed. There are only 4 other stable mainnet seed nodes and even less for Tor, having more even if unstable seems better than not having them at all as I'm not sure who else will be contributing these nodes.

PR #10116, opened 2025-09-27
dns: Add built-in DNS-over-TLS support

techmetx11:
This PR adds built-in DNS-over-TLS (or DoT, sometimes referred to as `dot` in the source code), enabled by setting `DNS_TLS=1` in the environment variables.
It also adds a new scheme for the `DNS_PUBLIC` parser: `tls` that accepts the IP of the DNS server and the expected domain of the DNS server, used to validate the given SSL certificate. (e.g. `tls://9.9.9.9#dns.quad9.net`)

### DIFF
--- a/src/common/dns_utils.cpp
+++ b/src/common/dns_utils.cpp
@@ -41,6 +41,7 @@
 #include <boost/algorithm/string/join.hpp>
 #include <boost/optional.hpp>
 #include <boost/utility/string_ref.hpp>
+#include <openssl/x509.h>
 using namespace epee;
 
 #undef MONERO_DEFAULT_LOG_CATEGORY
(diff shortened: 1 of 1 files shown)


src/common/dns_utils.cpp
```diff
@@ -226,9 +238,21 @@ static void add_anchors(ub_ctx *ctx)
...
 {
+  int use_dot = 0;
   int use_dns_public = 0;
   std::vector<std::string> dns_public_addr;
   const char *DNS_PUBLIC = getenv("DNS_PUBLIC");
+  const char *DNS_TLS = getenv("DNS_TLS");
+  if (DNS_TLS)
+  {
+    // Check for two common values for enabling DNS-over-TLS
+    if (strncmp(DNS_TLS, "1", 1) || strncmp(DNS_TLS, "true", 4))
```
stianov: Won't the second expression evaluate as true even when DNS_TLS is not "true"? For example if it's "0", then it will return -1 and evaluate as a nonzero value (true)? The same issue applies to the first expression in the reverse case.

techmetx11: That's true, I forgot to add a `!` before the `strncmp`s

techmetx11:
The tests related to DNS connectivity (resolving stuff over Quad9's regular DNS server, not even DNS-over-TLS) seems to failing, which fails the whole PR.
I've tested the unit tests on my computer and all of them work (even with `DNS_TLS=1`), so I don't know how can I debug this problem, so I'll just leave the PR up for review

techmetx11:
It seems like the latest commit fixed the DNS resolving errors (which I suspect is due to the hardcoded path to the TLS certificate bundle not existing within the CI's filesystem?)
Although it has failed on yet another error but it's not related to the PR.

```
/__w/monero/monero/tests/unit_tests/output_selection.cpp:132: Failure
Expected: (median) >= (1.3 * 86400), actual: 112200 vs 112320

[  FAILED  ] select_outputs.gamma (150 ms)
```

PR #10123, opened 2025-09-28
daemon: fix `print_cn` command when 0 connections

jeffro256:
Deferencing the result of `std::max_element()` when `res.connections` has 0 elements leads to UB, and segfaults in practice.


src/daemon/rpc_command_executor.cpp
```diff
@@ -644,12 +644,16 @@ bool t_rpc_command_executor::print_connections() {
...
-  auto longest_host = *std::max_element(res.connections.begin(), res.connections.end(),
-                                        [](const auto &info1, const auto &info2)
-                                        {
-                                          return info1.address.length() < info2.address.length();
-                                        });
-  int host_field_width = std::max(15, 8 + (int) longest_host.address.length());
+  int host_field_width = 15;
+  if (!res.connections.empty())
+  {
+    const auto &longest_host = *std::max_element(res.connections.begin(), res.connections.end(),
```
iamamyth: The container function just gets in the way here (and caused the original bug); better to iterate over the vector and compute the length for every member, updating `host_field_width` as warranted.

jeffro256: Done

PR #10133, opened 2025-09-30
crypto: derive key image generator

j-berman:
Helper function to derive `I` (the key image generator) used in the FCMP++ composition. The key image generator enters the FCMP++ tree as part of an output tuple {output pubkey, key image generator, commitment}.

Note: key images are x*I, where x is the one-time priv key used to spend an output.

See: https://github.com/kayabaNerve/fcmp-plus-plus-paper/blob/6f01fd3296b17f90b5c97c738cc3320e3369ad49/fcmp%2B%2B.pdf

### DIFF
--- a/src/crypto/crypto.cpp
+++ b/src/crypto/crypto.cpp
@@ -618,6 +618,12 @@ namespace crypto {
     ge_p1p1_to_p3(&res, &point2);
   }
 
+  void crypto_ops::derive_key_image_generator(const public_key &pub, ec_point &ki_gen) {
+    ge_p3 point;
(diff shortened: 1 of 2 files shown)


jeffro256:
I would hold off on this PR until the unbiased hash-to-point details pan out because if they do, then this PR changes too.

j-berman:
I would like to progress toward getting the tree building code reviewed, which this PR would be a necessary pre-req for. I figure this function would become something like `derive_biased_key_image_generator`, and the change to the code in this PR would be not so significant at a later time.

j-berman:
Closing in favor of #10338

PR #10135, opened 2025-09-30
crypto: fe_reduce_vartime

j-berman:
See the comment above `fe_reduce_vartime` for an explanation of what its purpose is.

In the FCMP++ integration, `fe_reduce_vartime` is used in the `torsion_check_vartime` function discussed in #10134.

> As it is written above, it is our opinion the following code is fit for use. The round-trip
through fe_tobytes and fe_frombytes_vartime correctly contracts limb magnitudes back
into the precondition domain of fe_add/fe_sub

From Cypher Stack audit section 7 ([commit](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf))

### DIFF
--- a/src/crypto/crypto-ops.c
+++ b/src/crypto/crypto-ops.c
@@ -1328,16 +1328,9 @@ void ge_double_scalarmult_base_vartime_p3(ge_p3 *r3, const unsigned char *a, con
   }
 }
 
-/* From ge_frombytes.c, modified */
+/* From fe_frombytes.c */
 
-int ge_frombytes_vartime(ge_p3 *h, const unsigned char *s) {
-  fe u;
-  fe v;
-  fe vxx;
-  fe check;
-
-  /* From fe_frombytes.c */
-
+int fe_frombytes_vartime(fe y, const unsigned char *s) {
   int64_t h0 = load_4(s);
   int64_t h1 = load_3(s + 4) << 6;
   int64_t h2 = load_3(s + 7) << 5;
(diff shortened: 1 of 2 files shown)


UkoeHB (approved):

j-berman:
Audited by Cypher Stack and Trail of Bits

UkoeHB (approved):

src/crypto/crypto-ops.c
```diff
@@ -3903,3 +3909,30 @@ int ge_p3_is_point_at_infinity_vartime(const ge_p3 *p) {
...
+   |h| bounded by 1.1*2^26,1.1*2^25,1.1*2^26,1.1*2^25,etc.
+
+We sometimes need to "reduce" field elems when they are in the postcondition's
+larger domain to match the precondition domain. This way we can take the output
+of fe_add or fe_sub and use it as input to another call to fe_add or fe_sub.
+
+We reduce by converting the field elem to its byte repr, then re-deriving the
+field elem from the byte repr.
+*/
+int fe_reduce_vartime(fe reduced_f, const fe f)
```
jeffro256, line 3933: If the precondition is `|h| bounded by 1.1*2^26,1.1*2^25,1.1*2^26,1.1*2^25,etc.`, then this function isn't vartime, even though it's calling a vartime function, since `fe_frombytes_vartime()` is only vartime in whether or not the field element is valid or not. It also doesn't have to return a value if the precondition holds (same as `fe_add()`, `fe_sub()`, `fe_mul()`, etc). Would the auditors be okay with switching the function signature to `void fe_reduce()` since the precondition is documented?

j-berman: Fine with keeping this as is to err on the side of conservative / help avoid a minor footgun here? It can be revisited if someone needs a constant time `fe_reduce` (this isn't used in any critical pathways for FCMP++/Carrot)

jeffro256: Fair enough

jeffro256 (approved):

Issue #3076, opened 2018-01-07
Automated Docker Hub build


noonien:
The project currently has a Dockerfile that can be used to build the project, it would be great if the project was automatically built by the Docker Hub, so users have access to the docker image more easily. This would allow users to deploy monerod by just running `docker run monero-project/monero`.

The only requirement for automated builds is that the GitHub repo have a Dockerfile, this requirement is already met.

Here's a guide on how to add the repository to the Docker Hub: https://docs.docker.com/docker-hub/github/#linking-docker-hub-to-a-github-account

moneromooo-monero:
For clarity to people who may not realize, this means you'd be running a binary built by some third party.

dEBRUYNE-1:
+proposal

modernNeo:
is there a reason this hasn't happened yet? It's not that hard to have an image uploaded to https://hub.docker.com/

SamsungGalaxyPlayer:
Personally, I think hosting Docker images [here on GitHub Packages](https://github.com/orgs/monero-project/packages) is a good idea. That way the existing organization account and its controls can be used. No Docker Hub account needed.

selsta:
Closing this proposal as no one stepped up to maintain this.

PR #10136, opened 2025-10-02
Support Monero Docker Image Build in GitHub Actions

thormme:
Fix for https://github.com/monero-project/monero/issues/3076
This adds a GitHub action to build and upload the Monero image to Dockerhub.
This will allow the node, rpc, and cli to be run in containerized environments without having to trust random developers on the internet.

Currently this is only building the amd64 image, but more could easily be added.

Two vars/secrets will need to be added to the repository for this to function:
`DOCKERHUB_USERNAME` - The username for Dockerhub
`DOCKERHUB_TOKEN` - The Dockerhub token for the specific image

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,6 +3,16 @@ name: "ci/gh-actions/docker-image"
 on:
   push:
     tags: [ 'release-v*.*' ]
+    paths-ignore:
+      - 'docs/**'
+      - '**/README.md'
+  pull_request:
+    paths:
+      - "Dockerfile*"
+
+env:
+  REGISTRY: docker.io
+  IMAGE_NAME: ${{ github.event.repository.name }}
 
 jobs:
   docker_buildx_debian:
@@ -20,6 +30,7 @@ jobs:
       - name: Login to Docker Hub
         uses: docker/login-action@v3
         with:
+          registry: ${{ env.REGISTRY }}
           username: ${{ vars.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


selsta:
As far as I know the reason this was not added yet is because no one volunteered to create and maintain a Dockerhub account. For security reasons obviously this would have to be someone trusted, and I'm not sure if others want this due to the centralized structure.

SamsungGalaxyPlayer:
You can consider publishing a GitHub Package from the `monero-project` GitHub organization instead:

https://docs.github.com/en/packages/learn-github-packages/introduction-to-github-packages

https://github.com/orgs/monero-project/packages

plowsof:
On pull request to the dockerfile just push it to the official dockerhub account.... "Main" branch... Those commits.. @thormme would you consider closing this PR and making an issue for discussion instead?

thormme:
@plowsof That's fair, I could have done more due diligence there. I've addressed your feedback.
To be clear, pull requests do not push at all `push: ${{ github.event_name != 'pull_request' }}`
The purpose of that pipeline is to enable testing of changes to the Dockerfile.
There is already an issue created that hasn't had much discussion in it, would you like me to make a new one or move the discussion to https://github.com/monero-project/monero/issues/3076?

thormme:
> For security reasons obviously this would have to be someone trusted, and I'm not sure if others want this due to the centralized structure.

I had a feeling that was one reason it hadn't happened. But the fact is that there are images already on Dockerhub built and maintained by random netizens, so not having an official one just makes it more likely that someone will use one that is malicious.

It should be possible to generate and serve hashes of the built images so that people could verify the identity of the images, if people don't want to trust Dockerhub.

thormme:
> You can consider publishing a GitHub Package from the `monero-project` GitHub organization instead:

I have set this up many times and would happily do that instead or in addition depending on what the maintainers want.

thormme:
@selsta Would you prefer I configure this to push the image to `ghcr.io`?

selsta:
@thormme sorry for the late reply. I don't use Docker with Monero so I don't have much of an opinion on it. If we add this we would have to find someone to maintain this, who is both active and familiar with Docker and Monero. GitHub Packages seems preferable over Dockerhub to me.

selsta:
Thank you again for the submission, I'm closing this PR as no one stepped up to maintain Docker + monero.

vtnerd:
I somehow missed this proposal. I'm currently building Docker images and posting them to [packages here](https://github.com/monero-lws/containers/pkgs/container/monerod). Those images use StageX, which allows for reproducible (and small) containers, but only works with amd64 at the moment. The majority of people are likely using @sethforprivacy build.

If there is a desire, I can move this into an official Monero thing on GH. Or we can use the alpine base image which supports more CPUs.

The difficulty will be when to trigger a new build. I would advocate for only building tagged releases, and then somehow push to GH packages only.

PR #10137, opened 2025-10-02
add the right prunable hash per tx to pruned get_blocks.bin call

### LINKED ISSUE #10120: `get_blocks.bin` does not return tx's `prunable_hash` when pruned
When calling `get_blocks.bin` with `prune = true` the blocks returned will have their tx blobs pruned but wont set the `prunable_hash`, leaving it as the default. This means you can't reconstruct the tx hash and verify the txs are actually the correct ones provided for the block.

https://github.com/monero-project/monero/blob/b591866fcfed400bc89631686655aa769ec5f2dd/src/rpc/core_rpc_server.cpp#L784

akildemir:
Fixes https://github.com/monero-project/monero/issues/10120 .

This approach uses the already existing `bs` return [type](https://github.com/monero-project/monero/blob/8e9ab9677f90492bca3c7555a246f2a8677bd570/src/rpc/core_rpc_server.cpp#L757) from `find_blockchain_supplement` function call instead of fetching the prunable hashes separately. The second of the `bs` type `std::vector<std::pair<crypto::hash, cryptonote::blobdata>>` It appears to be used for tx hash and pruned tx blob if pruned is `true`. However that tx hash is never used, so this approach uses that place for prunable hash and passes it to the return type accordingly.

This behavior might be documented or another approach might have been fetching prunable hashes separately after the `find_blockchain_supplement` call. This approach is chosen not to make that second call to the db and for its simplicity.

Comments are welcome!

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -3242,7 +3242,7 @@ std::vector<crypto::hash> BlockchainLMDB::get_txids_loose(const crypto::hash& tx
   return matching_hashes;
 }
 
-bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_block_count, size_t max_block_count, size_t max_tx_count, size_t max_size, std::vector<std::pair<std::pair<cryptonote::blobdata, crypto::hash>, std::vector<std::pair<crypto::hash, cryptonote::blobdata>>>>& blocks, bool pruned, bool get_miner_tx_hash) const
+bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_block_count, size_t max_block_count, size_t max_tx_count, size_t max_size, std::vector<std::pair<std::pair<cryptonote::blobdata, crypto::hash>, std::vector<std::tuple<crypto::hash, crypto::hash, cryptonote::blobdata>>>>& blocks, bool pruned, bool get_miner_tx_hash) const
 {
   LOG_PRINT_L3("BlockchainLMDB::" << __func__);
   check_open();
@@ -3304,8 +3304,12 @@ bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_block_cou
       result = mdb_cursor_get(m_cur_txs_pruned, &val_tx_id, &v, op);
       if (result)
         throw0(DB_ERROR(lmdb_error("Error attempting to retrieve transaction data from the db: ", result).c_str()));
-      if (!pruned)
-      {
+
+      if (pruned) {
+        result = mdb_cursor_get(m_cur_txs_prunable_hash, &val_tx_id, &v, op);
+        if (result)
+          throw0(DB_ERROR(lmdb_error("Error attempting to retrieve transaction data from the db: ", result).c_str()));
+      } else {
         result = mdb_cursor_get(m_cur_txs_prunable, &val_tx_id, &v, op);
         if (result)
           throw0(DB_ERROR(lmdb_error("Error attempting to retrieve transaction data from the db: ", result).c_str()));
@@ -3327,32 +3331,22 @@ bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_block_cou
 
       if (pruned) {
         // get the prunable hash
-
-        // Look up each transaction's tx_id
-        MDB_val_set(tx_hash_val, tx_hash);
-        result = mdb_cursor_get(m_cur_tx_indices, (MDB_val *)&zerokval, &tx_hash_val, MDB_GET_BOTH);
-        if (result)
-          throw0(DB_ERROR(lmdb_error("Error retrieving tx index for prunable hash: ", result).c_str()));
-
-        const txindex *tip = (const txindex *)tx_hash_val.mv_data;
-        MDB_val_set(val_tx_id_for_hash, tip->data.tx_id);
-
-        result = mdb_cursor_get(m_cur_txs_prunable_hash, &val_tx_id_for_hash, &v, MDB_SET);
+        result = mdb_cursor_get(m_cur_txs_prunable_hash, &val_tx_id, &v, op);
         if (result)
           throw0(DB_ERROR(lmdb_error("Error attempting to retrieve transaction data from the db: ", result).c_str()));
 
         crypto::hash prunable_hash = *(const crypto::hash*)v.mv_data;
-        current_block.second.push_back(std::make_pair(prunable_hash, std::move(tx_blob)));
+        current_block.second.push_back(std::make_tuple(tx_hash, prunable_hash, std::move(tx_blob)));
       } else {
         // get the prunable data
         result = mdb_cursor_get(m_cur_txs_prunable, &val_tx_id, &v, op);
         if (result)
           throw0(DB_ERROR(lmdb_error("Error attempting to retrieve transaction data from the db: ", result).c_str()));
         tx_blob.append(reinterpret_cast<const char*>(v.mv_data), v.mv_size);
-        current_block.second.push_back(std::make_pair(tx_hash, std::move(tx_blob)));
+        current_block.second.push_back(std::make_tuple(tx_hash, crypto::null_hash, std::move(tx_blob)));
       }
 
-      size += current_block.second.back().second.size();
+      size += std::get<2>(current_block.second.back()).size();
     }
 
     if (blocks.size() >= min_block_count && num_txes >= max_tx_count)
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -3251,8 +3251,10 @@ bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_block_cou
   RCURSOR(blocks);
   RCURSOR(tx_indices);
   RCURSOR(txs_pruned);
-  if (!pruned)
+  if (pruned)
   {
+    RCURSOR(txs_prunable_hash);
+  } else {
     RCURSOR(txs_prunable);
   }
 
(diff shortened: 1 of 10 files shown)


jeffro256:
This adds >32B per transaction for `wallet2` which doesn't currently have code implemented to check the TXID using the prunable hash. Before this gets merged, I'd like to see the check on the wallet side implemented, perhaps in a dependent PR. For backwards compatibility, it could skip the check if `--allow-mismatched-daemon-version` is specified. I could do it but am sort of busy at the moment.

Also, the RPC minor version should to be bumped since a feature was added.

kayabaNerve:
The current `get_blocks.bin` pruned response always includes the `prunable_hash` field, even if not set @jeffro256. Setting it to the correct value, instead of the null hash, would not increase bandwidth and I'd argue it a bug fix, not a feature (though it may still justify a minor version bump).

jeffro256:
Sorry, I was under the impression that null hashes don't get serialized in `block_complete_entry`, but that's `KV_SERIALIZE_VAL_POD_AS_BLOB_OPT`, not `KV_SERIALIZE_VAL_POD_AS_BLOB`. So yeah this takes up the same bandwidth in that case.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -3324,14 +3326,33 @@ bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_block_cou
...
+
+        const txindex *tip = (const txindex *)tx_hash_val.mv_data;
+        MDB_val_set(val_tx_id_for_hash, tip->data.tx_id);
+
+        result = mdb_cursor_get(m_cur_txs_prunable_hash, &val_tx_id_for_hash, &v, MDB_SET);
+        if (result)
+          throw0(DB_ERROR(lmdb_error("Error attempting to retrieve transaction data from the db: ", result).c_str()));
+
+        crypto::hash prunable_hash = *(const crypto::hash*)v.mv_data;
+        current_block.second.push_back(std::make_pair(prunable_hash, std::move(tx_blob)));
```
jeffro256: NACK replacing the TXID with the prunable hash, this silently breaks the API of `BlockchainDB`. If we must update the API here, the function signature should be a breaking change so that downstream fails to compile. That, or we can write a new `get_blocks_from()` method override and implement the older override with the newer override so that it's non-breaking.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -3324,14 +3326,33 @@ bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_block_cou
         throw0(DB_ERROR(lmdb_error("Error attempting to retrieve transaction data from the db: ", result).c_str()));
       tx_blob.assign((const char*)v.mv_data, v.mv_size);
 
-      if (!pruned)
-      {
+      if (pruned) {
+        // get the prunable hash
+
+        // Look up each transaction's tx_id
```
jeffro256: We can skip the TXID -> tx index lookup by using already existing `tx_id` (terrible name 😢) variable. Tx on-chain indices are assigned in ascending order by 1) block, then 2) txs in block in specified order.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -3324,14 +3326,33 @@ bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_block_cou
...
+        MDB_val_set(tx_hash_val, tx_hash);
+        result = mdb_cursor_get(m_cur_tx_indices, (MDB_val *)&zerokval, &tx_hash_val, MDB_GET_BOTH);
+        if (result)
+          throw0(DB_ERROR(lmdb_error("Error retrieving tx index for prunable hash: ", result).c_str()));
+
+        const txindex *tip = (const txindex *)tx_hash_val.mv_data;
+        MDB_val_set(val_tx_id_for_hash, tip->data.tx_id);
+
+        result = mdb_cursor_get(m_cur_txs_prunable_hash, &val_tx_id_for_hash, &v, MDB_SET);
```
jeffro256: After the cursor is set once at the beginning of the loop, we can use `op = MDB_NEXT` instead of `MDB_GET_BOTH` which is much, much faster for accessing sequential values.

jeffro256 (changes requested):
Thank you for taking this on! Once the below issues are addressed, I will be willing to approve.

src/blockchain_db/lmdb/db_lmdb.cpp
akildemir: So I'm trying to understand this and your  next https://github.com/monero-project/monero/pull/10137#discussion_r2404314336 .   if you mean it would be enough just to do `result = mdb_cursor_get(m_cur_txs_prunable_hash, &val_tx_id, &v, op);` instead of first fetching the tx index, that doesn't work for `txs_prunable_hash` table like it would for `txs_pruned` or `txs_prunable`tables. it gives the same prunable hash over over for all txs.

jeffro256: I mean that you should do a `MDB_SET` op first to first set the cursor to the entry at `tx_id` for the first transaction in the list, then do `MDB_NEXT` afterwards.

akildemir: Sorry for the late reply. Had a busy week. So it looks like I miss the `MDB_SET` that happens inside this if block https://github.com/monero-project/monero/blob/d32b5bfe18e2f5b979fa8dc3a8966c76159ca722/src/blockchain_db/lmdb/db_lmdb.cpp#L3301 I think the confusion arose from the fact that `MDB_SET` for tables `txs_prunable` and `txs_pruned` is inside the if block. I checked `BlockchainLMDB::get_blocks_from` is always called with `skip_coinbase` is `true` and it looks like the function won't return as intended if that were to be passed as `false` since those tables would also be calling `MDB_NEXT` before `MDB_SET`. Indeed, when I compiled it with `false`, `monero-wallet-cli` couldn't scan the chain. It immediately throws exception `Error: refresh failed: internal error: Exception in thread pool. Blocks received: 0`.

So I believe there is a bug there. If `skip_coinbase` is  necessary for that function to work properly it shouldn't be a parameter? I think the original intention was to actually skip the coinbase tx in each block by retrieving the tx and ignoring the value of  the call so that cursor moves to the next tx when calling with `MDB_NEXT` and that works as intended if we wanted to skip the coinbase txs but otherwise it misses the `MDB_SET` op to those tables. So a fix would be to always call `MDB_SET`and only ignore the value if `skip_coinbase` is `true`.

jeffro256: Yes, the `skip_coinbase=false` thing is a bug. I opened #10166 to fix it. Perhaps you could rebase this PR on top of that? But yes, generally, you should do `MDB_SET` once and then `MDB_NEXT` afterwards for sequential keys.

akildemir: Ok I can rebase once that pr is merged 👍

akildemir:
@jeffro256 rebased to latest since https://github.com/monero-project/monero/pull/10166 is merged now, and also updated to return type to be a tuple to change the function signature.

cacaosteve:
I tested the current branch `akildemir/add-prunable-hash-to-get-blocks` on a real wallet workload and am seeing daemon/database retrieval failures that break `get_blocks.bin` responses.

**Build tested**
- monerod `--version`: `Monero 'Fluorine Fermi' (v0.18.1.0-1d2315c87)`

**Workload**
- Wallet scanning via `get_blocks.bin` (pruned path), fetching contiguous ranges (e.g. `0..=24` and later ranges).

**Client symptom**
- Wallet fails with truncated/invalid EPEE payload:
  - `contiguous_scannable_blocks_error ... invalid node (EpeeError::Short(8))` (range=`0..=24`)
- In other runs I also hit strict mismatch errors (“interface returned TX X when Y was requested”).

**Daemon symptom**
- monerod logs repeatedly:
  - `W Error attempting to retrieve transaction data from the db: MDB_NOTFOUND: No matching key/data pair found`

**Control**
- Rebuilding/running stock monerod `v0.18.4.5` on the same host/LMDB does not reproduce `MDB_NOTFOUND` and wallet scanning proceeds normally.

This suggests the PR’s LMDB access changes / plumbing for the new tuple return type may be triggering invalid cursor/key usage under load, causing invalid `get_blocks.bin` responses.

cacaosteve:
I put a PR on this branch https://github.com/akildemir/monero/pull/1

src/rpc/core_rpc_server.cpp
```diff
@@ -779,19 +779,19 @@ namespace cryptonote
...
         res.blocks.back().txs.reserve(bd.second.size());
-        for (std::vector<std::pair<crypto::hash, cryptonote::blobdata>>::iterator i = bd.second.begin(); i != bd.second.end(); ++i)
+        for (auto i = bd.second.begin(); i != bd.second.end(); ++i)
         {
-          res.blocks.back().txs.push_back({std::move(i->second), crypto::null_hash});
-          i->second.clear();
-          i->second.shrink_to_fit();
+          res.blocks.back().txs.push_back({std::move(std::get<2>(*i)), std::get<1>(*i)});
+          std::get<2>(*i).clear();
+          std::get<2>(*i).shrink_to_fit();
```
jeffro256: These two lines are redundant / regressive and can removed. If the string is using SSO, then the `.clear()` and `.shrink_to_fit()` calls don't release any memory; they are redundant. If the string is *not* using SSO, and owns a heap allocation, then the earlier call to `std::move()` will move that heap allocation and replace the contents of the current string with `""`.

akildemir: removed these 👍

jeffro256 (approved):

selsta:
please squash the commits

cacaosteve:
@akildemir the last commit on this PR is asking you to squash commits

selsta:
@cacaosteve could you test if this branch solves your issue with wallet scan? https://github.com/selsta/monero/tree/clean-prunable-hash

cacaosteve:
yes thank you!!!!

selsta:
Continuing in #10652

PR #10139, opened 2025-10-02
wallet_rpc_server: add HTTP body size limit

Oblivionsage:
Sets `m_max_content_length` to `MAX_RPC_CONTENT_LENGTH` (1MB) to match the protection already implemented in `core_rpc_server.cpp` (line 427).

Without this limit, wallet RPC accepts unlimited HTTP body sizes which could lead to resource exhaustion issues.

**Testing:**
- Verified requests under 1MB work normally
- Verified requests over 1MB are rejected
- Behavior now consistent with core RPC server

Addresses private HackerOne security report

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -398,6 +398,7 @@ namespace tools
     }
 
     m_net_server.set_threads_prefix("RPC");
+    m_net_server.get_config_object().m_max_content_length = MAX_RPC_CONTENT_LENGTH;
     auto rng = [](size_t len, uint8_t *ptr) { return crypto::rand(len, ptr); };
     return epee::http_server_impl_base<wallet_rpc_server, connection_context>::init(
       rng, std::move(bind_port), std::move(rpc_config->bind_ip),


vtnerd:
I don't understand, someone would run wallet-rpc-server and allow remote connections?

selsta:
@vtnerd we deemed that this is not realistically exploitable, but if it's easy to add a limit I don't see why we shouldn't add one

@Oblivionsage do you know if for example `import_key_images` is implemented with batching? Otherwise this request can easily surpass 1MB.

Oblivionsage:
@selsta Good point. I don't know if import_key_images has batching implemented . If legitimate use cases regularly exceed 1MB, we might need either:

1. A higher limit (like 10MB)
2. Per-endpoint limits instead of a blanket 1MB

Core RPC uses the same 1MB limit - not sure if they have similar issues with large legitimate payloads. What would you suggest? I can adjust the limit or make it configurable if needed

selsta:
A per-endpoint limit is overkill here... I would set it to 100MB.

Oblivionsage:
@selsta updated per feedback. Let me know if this looks good or needs any changes

selsta:
@Oblivionsage looks good, please squash

Oblivionsage:
done , thanks for the review @selsta

PR #10143, opened 2025-10-03
Daemon RPC: /getblocks.bin optional block_ids_exclusive req param

j-berman:
When the request includes `block_ids`, the daemon uses `find_blockchain_supplement` to identify the highest block hash passed in `block_ids` that the daemon also knows about, and then serves subsequent blocks contiguous to that block.

When `block_ids_exclusive` is false (default current behavior), the daemon includes the highest block requested in the response, in addition to contiguous blocks after it.

When `block_ids_exclusive` is true (new param), the daemon serves blocks starting from the block 1 higher than the highest known block included in `block_ids`. Thus, the daemon skips the common block known to the client and daemon. Clients can make sure the daemon is serving expected contiguous blocks to its highest known block by checking the first block's `prev_id` included in the response, and making sure it is equivalent to the block the client already knows about that was included in `block_ids`. This avoids the daemon serving 1 extra block it does not need to serve to the client, since the client should already know about that block.

_______

This optimization is used in the FCMP++ integration wallet2 refresh, implemented as part of this PR refactoring wallet2 refresh: https://github.com/seraphis-migration/monero/pull/81

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -2736,17 +2736,18 @@ bool Blockchain::find_blockchain_supplement(const std::list<crypto::hash>& qbloc
 // find split point between ours and foreign blockchain (or start at
 // blockchain height <req_start_block>), and return up to max_count FULL
 // blocks by reference.
-bool Blockchain::find_blockchain_supplement(const uint64_t req_start_block, const std::list<crypto::hash>& qblock_ids, std::vector<std::pair<std::pair<cryptonote::blobdata, crypto::hash>, std::vector<std::pair<crypto::hash, cryptonote::blobdata> > > >& blocks, uint64_t& total_height, crypto::hash& top_hash, uint64_t& start_height, bool pruned, bool get_miner_tx_hash, size_t max_block_count, size_t max_tx_count) const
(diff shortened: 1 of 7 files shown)


iamamyth:
`skip_common_block` seems a rather long-winded synonym for `exclusive`.

j-berman:
> skip_common_block seems a rather long-winded synonym for exclusive.

Updated to use `_exclusive` in the latest instead

src/cryptonote_core/blockchain.cpp
```diff
@@ -2755,15 +2756,30 @@ bool Blockchain::find_blockchain_supplement(const uint64_t req_start_block, cons
...
+        blocks.clear();
+        return true;
+      }
+    }
   }
 
   db_rtxn_guard rtxn_guard(m_db);
   top_hash = m_db->top_block_hash(&total_height);
   ++total_height;
+  CHECK_AND_ASSERT_MES(total_height >= start_height, false, "chain height expected to be higher than start block");
```
iamamyth: Minor issue, but the message here doesn't match the assertion (ignores equality); suggest something like: Start height cannot exceed chain height.

iamamyth (approved):
Looks OK, apart from the (easily resolved) inaccuracy of the assertion message.

src/cryptonote_core/blockchain.cpp
j-berman: Actually it should really be `total_height > start_height`, good spot

j-berman: Fixed to check `total_height > start_height`

selsta (approved):

PR #10161, opened 2025-10-12
add getter for daemon proxy

woodser:
wallet2 already provides getters for connection information, like the daemon address and login credentials.

This PR just adds a getter for the daemon proxy URI, so it can be collected by clients too.

### DIFF
--- a/src/wallet/wallet2.h
+++ b/src/wallet/wallet2.h
@@ -1547,6 +1547,7 @@ private:
     std::string get_keys_file() const;
     std::string get_daemon_address() const;
     const boost::optional<epee::net_utils::http::login>& get_daemon_login() const { return m_daemon_login; }
+    std::string get_daemon_proxy() const { return m_proxy; }
     uint64_t get_daemon_blockchain_height(std::string& err);
     uint64_t get_daemon_blockchain_target_height(std::string& err);
     uint64_t get_daemon_adjusted_time();


nahuhh:
Those other items (login info, address) are things that the caller already knows. The proxy is not. 

i'm not sure its a good idea to expose the proxy to users, unless its only available to the unrestricted rpc

edit.. this is for the wallet, so nvm. 
edit again: is this for the _daemon_ proxy, or the wallet proxy?

woodser:
> is this for the daemon proxy, or the wallet proxy?

@nahuhh 

This is the wallet's proxy to its daemon.

Without this, it's not possible for a client to collect the wallet's full daemon connection information after it's set.

This accessor is merely consistent with the other available accessors, to get the wallet daemon's address and login info, allowing a client to query this information as needed.

An example use case (already used in Haveno [here](https://github.com/haveno-dex/haveno/blob/8697b5344f50a67f97301dde1ff2d0ba469ebf6c/core/src/main/java/haveno/core/trade/Trade.java#L3028)), is that listeners can know if the global daemon connection has changed from each wallet's daemon connection, including the proxy configuration, in order to update each wallet's daemon connection accordingly.

Without this, each wallet would have to cache and maintain this information locally. It makes sense instead that there is a consistent accessor.

SNeedlewoods (approved):

navidR (approved):

PR #10166, opened 2025-10-13
BlockchainDB: rm `skip_coinbase` in `get_blocks_from()`

jeffro256:
Parameter `skip_coinbase` was always set to `true`, and didn't actually skip the coinbase transactions. The method broke when set to `false`.

Noticed by @akildemir.

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -3242,7 +3242,7 @@ std::vector<crypto::hash> BlockchainLMDB::get_txids_loose(const crypto::hash& tx
   return matching_hashes;
 }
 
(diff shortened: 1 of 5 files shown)


src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -3314,7 +3313,7 @@ bool BlockchainLMDB::get_blocks_from(uint64_t start_height, size_t min_block_cou
     op = MDB_NEXT;
 
     current_block.second.reserve(b.tx_hashes.size());
-    num_txes += b.tx_hashes.size() + (skip_coinbase ? 0 : 1);
+    num_txes += b.tx_hashes.size() + 1;
```
iamamyth, line 3316: If `skip_coinbase` was previously `true`, adding `1` here would change the behavior.

jeffro256: This simply makes the `num_txes` count more accurate, which makes the limiting check against `max_tx_count` more accurate as well.

jeffro256: @iamamyth Can you please revisit this review?

PR #10170, opened 2025-10-14
epee: align container pod as blob serialization

jeffro256:
`{un}serialize_stl_container_pod_val_as_blob()` use aligned memory accesses on potentially unaligned pointers.

### DIFF
--- a/contrib/epee/include/serialization/keyvalue_serialization_overloads.h
+++ b/contrib/epee/include/serialization/keyvalue_serialization_overloads.h
@@ -117,11 +117,11 @@ namespace epee
       if(!container.size()) return true;
       std::string mb;
       mb.resize(sizeof(typename stl_container::value_type)*container.size());
-      typename stl_container::value_type* p_elem = (typename stl_container::value_type*)mb.data();
+      char *p_elem = mb.data();
       BOOST_FOREACH(const typename stl_container::value_type& v, container)
       {
-        *p_elem = v;
-        p_elem++;
+        memcpy(p_elem, std::addressof(v), sizeof(typename stl_container::value_type));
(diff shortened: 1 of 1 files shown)


jeffro256:
There's potential issues if the `operator=` for that type did something that wasn't trivial copying of bytes, but I can't assert `std::is_trivially_copyable` on the value type, without some refactoring, since it's used for `crypto::secret_key`s in `wallet2`.

contrib/epee/include/serialization/keyvalue_serialization_overloads.h
```diff
@@ -135,14 +135,19 @@ namespace epee
...
-        typename stl_container::value_type* pelem =  (typename stl_container::value_type*)buff.data();
+        char *pelem = buff.data();
         CHECK_AND_ASSERT_MES(!(loaded_size%sizeof(typename stl_container::value_type)), 
           false, 
           "size in blob " << loaded_size << " not have not zero modulo for sizeof(value_type) = " << sizeof(typename stl_container::value_type) << ", type " << typeid(typename stl_container::value_type).name());
         size_t count = (loaded_size/sizeof(typename stl_container::value_type));
         hint_resize(container, count);
         for(size_t i = 0; i < count; i++)
-          container.insert(container.end(), *(pelem++));
+        {
```
vtnerd, line 145: The existing code was using the copy constructor, and not assignment operator .. ?

jeffro256: Yeah, sorry, I meant both of them. It uses assignment on the serialization side.

0xFFFC0000 (approved):

PR #10182, opened 2025-11-01
release: drop support for arm32 linux

tobtoht:
This PR drops support for arm (32-bit) Linux release builds.

All Raspberry Pi models released in the past decade are capable of running the 64-bit version of [Raspberry Pi OS](https://www.raspberrypi.com/software/operating-systems/), which is selected by default in Raspberry Pi Imager.


tobtoht:
Yes, no plans to remove the 32-bit Android target from CI.

selsta (approved):

hinto-janai (approved):

PR #10186, opened 2025-11-03
tx_memory_pool: speedup get_complement() for large requests

jeffro256:
Changes complexity from $M * N$ to $(2 * N+M) * log2(M)$, where $M$ is the requester's mempool size and $N$ is the responder's mempool size. The FCMP++ stressnet recently hit mempool sizes of ~55k txs. If the requesting node's mempool is populated, this results in an average of $(55000*55000)/2$ (about 1.5 billion) comparisons for the responding node. Under this commit, this would be reduced to $(2 * 55000+55000)*log2(55000)$ comparisons (about 2.6 million), a 99.83% reduction.

This commit is especially noteworthy for FCMP++ stressnet node performance when considering https://github.com/seraphis-migration/monero/pull/177 is merged into the stressnet branch. The cited change makes the pool complement p2p call much more frequently.

### DIFF
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -666,17 +666,24 @@ namespace cryptonote
     return true;
   }
   //---------------------------------------------------------------------------------
-  bool tx_memory_pool::get_complement(const std::vector<crypto::hash> &hashes, std::vector<cryptonote::blobdata> &txes) const
+  bool tx_memory_pool::get_complement(std::vector<crypto::hash> hashes, std::vector<cryptonote::blobdata> &txes) const
   {
     CRITICAL_REGION_LOCAL(m_transactions_lock);
     CRITICAL_REGION_LOCAL1(m_blockchain);
+
+    // Sort so we can do binary search later
+    std::sort(hashes.begin(), hashes.end());
 
     m_blockchain.for_all_txpool_txes([this, &hashes, &txes](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref*) {
       const auto tx_relay_method = meta.get_relay_method();
       if (tx_relay_method != relay_method::block && tx_relay_method != relay_method::fluff)
         return true;
-      const auto i = std::find(hashes.begin(), hashes.end(), txid);
-      if (i == hashes.end())
+
+      // Do binary search for our pool TXID in given list, skip to next if already present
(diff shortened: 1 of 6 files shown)


jeffro256:
Refactored to remove Boost usage.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -671,12 +671,27 @@ namespace cryptonote
     CRITICAL_REGION_LOCAL(m_transactions_lock);
     CRITICAL_REGION_LOCAL1(m_blockchain);
 
-    m_blockchain.for_all_txpool_txes([this, &hashes, &txes](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref*) {
+    // Create an ordering for `crypto::hash` and sort `hashes` according to that order
+    struct hash_compare
```
vtnerd: Does it make sense to provide an `operator<` instead for future use? Or is there a concern about `memcmp` not being constant-timed?

src/cryptonote_core/tx_pool.cpp
```diff
@@ -671,12 +671,27 @@ namespace cryptonote
...
+    {
+      bool operator()(const crypto::hash &lhs, const crypto::hash &rhs) const noexcept
+      {
+        return memcmp(lhs.data, rhs.data, sizeof(crypto::hash)) < 0;
+      }
+    };
+    std::vector<crypto::hash> hashes_sorted(hashes.cbegin(), hashes.cend());
```
vtnerd: Why not take the argument-by-value and then do the sort on that? That allows the caller to do a copy or move.

jeffro256: I did it this way to minimize the diff, but that's absolutely an option. Do you think it's worth it?

src/cryptonote_core/tx_pool.cpp
jeffro256: Yeah, perhaps. The `operator<` for `crypto::public_key` and `crypto::key_image` isn't constant time. I can override `operator<` for `crypto::hash` just like the other too. That probably wouldn't be too controversial.

jeffro256: Done in latest

src/cryptonote_core/tx_pool.cpp
jeffro256: Done in latest

jeffro256:
Squashed

PR #10187, opened 2025-11-05
guix: use clang 19 for darwin & freebsd targets

tobtoht:
With this I can remove the GCC toolchain for native package builds from the darwin and freebsd build environments in a follow up PR.


selsta (approved):

selsta:
@tobtoht is CI failing a network issue?

tobtoht:
Yes, green now.

PR #10189, opened 2025-11-06
simplewallet: use passed decrypted payment ID from wallet2 for notifications

jeffro256:
Adds payment ID parameter to `on_money_received()` wallet callback. Carrot changes the way that payment IDs are encrypted, so this `simplewallet` logic breaks. This commit allows `wallet2` to pass the decrypted payment ID by its callback to reduce logic duplication.

Part of upstreaming Carrot/FCMP++. Related to https://github.com/seraphis-migration/monero/pull/217.


navidR (approved):

navidR:
One thing to keep in mind: `on_money_received` signature change in `src/wallet/wallet2.h` is a break for any out-of-tree implementers.

PR #10193, opened 2025-11-08
Fix logging lock behavior

j-berman:


nahuhh:
this solves some deadlocks that i had experienced

+1

contrib/epee/include/misc_log_ex.h
```diff
@@ -42,7 +43,10 @@
 
 #define MCLOG_TYPE(level, cat, color, type, x) do { \
     if (el::Loggers::allowed(level, cat)) { \
-      el::base::Writer(level, color, __FILE__, __LINE__, ELPP_FUNC, type).construct(cat) << x; \
+      std::stringstream ss; \
```
vtnerd: There's another similar macro that might need the same treatment.

vtnerd: Similar macro _below_ this one.

j-berman: Required passing a string to the << operator for `Writer`, and caught another spot to use the same approach (`MLOGIF_P2P_MESSAGE`)

tests/unit_tests/logging.cpp
```diff
@@ -251,3 +254,37 @@ TEST(logging, empty_configuration)
...
+  // 1. Thread 1 starts logger
+  // 2. Thread 2 grabs mutex shared across threads
+  // 3. Thread 2 logs
+  // 4. Thread 1 grabs mutex shared across threads
+  // 5. Thread 1 finishes logging
+  const auto thread1_func = [&mutex]
+  {
+    const auto thread1_inner_func = [&mutex]() -> std::string
+    {
+      sleep(2); // allow thread2 to grab the lock
```
iamamyth: Sleep as a proxy for waiting on a condition nearly always proves a wrong decision, and this is a perfect illustration: You can use condition variables to wait on the desired conditions, which both ensures you're testing the actual behavior of interest, and makes the code self-documenting.

j-berman: Used condition variables in the latest

j-berman:
Squashed

PR #10213, opened 2025-11-13
cmake: set minimum boost version to 1.69

tobtoht:
Context: https://github.com/monero-project/monero/pull/9807#issuecomment-2667681053

Given that Boost < 1.69 has a known crash on some platforms, and we no longer support Debian 10, we can bump the minimum version here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -924,7 +924,7 @@ if(STATIC)
 endif()
 
 # Find Boost headers
-set(BOOST_MIN_VER 1.62)
+set(BOOST_MIN_VER 1.69)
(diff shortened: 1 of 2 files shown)


jeffro256:
I think that this breaks support for Ubuntu 18.04 since highest version in the package manager is 1.65. What if we only set minimum version to 1.69 on affected systems?

selsta:
@jeffro256 we had some discussion here https://github.com/monero-project/monero/issues/9446

selsta (approved):

selsta:
Needs rebase.

selsta (approved):

plowsof (approved):

jpk68 (approved):

PR #10220, opened 2025-11-21
cryptonote_basic: fix add_extra_nonce_to_tx_extra() length

jeffro256:
Serialization/deserialization code in `tx_extra.h` treats the nonce length as a varint, but `cryptonote::add_extra_nonce_to_tx_extra()` writes the nonce length as a raw unsigned 8-bit integer. For nonce sizes in $[128, 256)$, these two sections diverge. This commit amends `cryptonote::add_extra_nonce_to_tx_extra()` to match the serialization code.

Issue identified by the Monero Oxide project and reporters.

### DIFF
--- a/tests/unit_tests/cryptonote_format_utils.cpp
+++ b/tests/unit_tests/cryptonote_format_utils.cpp
@@ -1,0 +1,245 @@
+// Copyright (c) 2025, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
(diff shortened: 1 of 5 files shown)


jeffro256:
Unit test failure is unrelated.

jeffro256:
Fixed test cases up and added tests for `tx_extra_merge_mining_tag` and `add_mm_merkle_root_to_tx_extra()`.

tests/unit_tests/cryptonote_format_utils.cpp
```diff
@@ -0,0 +1,244 @@
...
+        std::string nonce;
+        std::vector<cryptonote::tx_extra_field> tx_extra_fields;
+        extra.reserve(extra_prefix.size() + max_nonce_size + 1 + 10);
+        nonce.reserve(max_nonce_size);
+        tx_extra_fields.reserve(2);
+        for (std::size_t nonce_size = 0; nonce_size <= max_nonce_size; ++nonce_size)
+        {
+            extra = extra_prefix;
+            nonce.resize(nonce_size);
+            memset(&nonce[0], '%', nonce.size());
```
selsta: UB if nonce_size = 0

tests/unit_tests/varint.cpp
```diff
@@ -62,3 +62,101 @@ TEST(varint, equal)
...
+  static constexpr int DIGITS = std::numeric_limits<T>::digits;
+  static_assert(DIGITS == CHAR_BIT * sizeof(T));
+  static_assert(sizeof(T));
+
+  for (T n = 0; n < 128; ++n)
+  {
+    subtest_varint_byte_size_for_single_value(1, n);
+  }
+
+  for (T n = 128; n < 255; ++n)
```
SChernykh, line 106: `n <= 255` maybe?

tests/unit_tests/varint.cpp
```diff
@@ -62,3 +62,101 @@ TEST(varint, equal)
...
+
+  if constexpr (DIGITS > 8)
+  {
+    static_assert(DIGITS >= 16);
+    for (T n = 256; n < 16384; ++n)
+    {
+      subtest_varint_byte_size_for_single_value(2, n);
+    }
+
+    for (T n = 16384; n < 65535; ++n)
```
SChernykh, line 121: `n <= 65535`

tests/unit_tests/varint.cpp
SChernykh: But be careful about type bounds to not make it an infinite loop. So `(n <= 255) && n` is better, to be sure that it doesn't get stuck.

SChernykh: Actually nevermind, I see that you check 255 and 65535 separately further down in the code.

tests/unit_tests/cryptonote_format_utils.cpp
jeffro256: For the record, this probably is fine in practice since `operator[]()` will return some garbage pointer, and `memset()` is passed a length of `0`, but you're absolutely right that `operator[]()` when `pos < size()` is `false` triggers UB (At least until C++26 with a "hardened implementation", under which scenario it becomes a contract violation, not UB). 

Will update

jeffro256:
Sorry, I made one more change: I added an `if` branch before the `mempy()` so that `tx_extra[start_pos]` isn't called when the nonce is empty (and thus `start_pos == tx_extra.size()`). This triggers UB when `extra_nonce.empty()` and actually trips a debug assertion on newer versions of `glibc++`.

PR #10223, opened 2025-11-22
StageX release builds

tobtoht:
## Why [StageX](https://stagex.tools/) over Guix

- Guix upstream is much slower to update packages.
- Container engines ship with AppArmor/SELinux profiles. No issues with unprivileged user namespaces.
- Container engines run on more Linux distributions.
- Registries like docker.io are more reliable than Guix substitute servers.
- StageX images are multi-party-signed. Less risky than trusting Guix substitutes.
- Build system configuration in Containerfiles/shell instead of a lisp dialect.
- We don't have to trust a single driver. StageX release builds can support any container engine.
- We have more control of what ends up in our build environment. Down to the file level.
- Builders are more likely to already have Docker or podman installed.
- More off-the-shelf OCI tooling available. Easier debugging and caching.
- FHS instead of `/gnu/store`.

### 0. Requirements

- x86_64 Linux environment
- 50GB of free space

### 1. Install docker or podman

 **Docker**

Installation instructions: https://docs.docker.com/engine/install/

If you have an existing installation, make sure it supports buildx:

```bash
docker buildx version
```

**Podman**

 Installation instructions: https://podman.io/docs/installation#installing-on-linux

You may need to add: `unqualified-search-registries = ["docker.io", "quay.io"]` to `/etc/containers/registries.conf`

### 2. Run a build

**Docker** 

```bash
make stagex-build
```

**Podman**

```
CONTAINER_ENGINE=podman make stagex-build
```

### 3. Get hashes

```bash
make stagex-hashes
```

## Debugging

If you made changes to the source for testing purposes, you can start a build with:

```bash
FORCE_DIRTY_WORKTREE=1 make stagex-build
```

To interactively debug a build:

```bash
INTERACTIVE=1 make stagex-build
```

To build for a specific target:

```bash
HOSTS="arm64-apple-darwin" make stagex-build
```

To limit the number of build threads:

```bash
JOBS=N make stagex-build
```

If something went wrong, and you want to start over:

```bash
rm -rf stagex
```

## Tools

| Tool | master | This PR |
|------|------|-------|
| GCC | 15.2.0 | **15.3.0** |
| Clang | **22.1.8** | 22.1.5 |
| Rust | 1.93.0 | **1.96.0** |
| Binutils | 2.44 | **2.47** |
| Mingw-w64 | 13.0.0 | **14.0.0** |
| CMake | 3.31.10 |**4.3.3** |
| musl | n/a | 1.2.6 |
| mimalloc | n/a | 3.5.1 |

## Why static [musl](https://musl.libc.org/)+[mimalloc](https://github.com/microsoft/mimalloc) over dynamic glibc

We get fully static release binaries that run on pretty much any Linux distribution, at less complexity and binary weight than statically linking glibc.

Compared to gitian/guix release binaries, these also run:

- On distributions with old versions of glibc (< 2.27)
- On musl-based distributions (like Alpine)
- On non-FHS distributions (like Nix and Guix System)
- In a `termux-chroot` on Android
- In `FROM scratch` containers

mimalloc performs better at checkpointed blockchain synchronization and has a lower memory footprint than glibc, see results in the next section.

### Syncing 500 `beta-stressnet` blocks

block_sync_progress_500_blocks

memory_usage_500_blocks

### First 2.5M mainnet blocks

block_sync_progress_2 5M

- glibc slightly outperforms musl/mimalloc during v4-v10
- musl/mimalloc is about 25% faster than glibc after v12

### Linux targets

bootstrapgraph

Reduces packages (excluding bootstrap) from 145 to 81 (-44%) compared to Guix.

### macOS, freebsd, and android targets

clang-graph


desutschland:
Interesting thread about StageX: 
https://scholar.social/@khinsen/115484895971642349

lrvick:
> Interesting thread about StageX:
> https://scholar.social/@khinsen/115484895971642349

All the criticisms in that thread seem born from misunderstandings. I just spent a bit of time to try to correct most of it. Wish we had seen this sooner! Learning how to explain new projects and address common misconceptions is always a challenge.

Several of us on the stagex team are fans of Monero and happy to address any questions, critiques, or blockers. As mostly sysadmins and security engineers, my team and I believe most of the decentralized projects we love are at risk by trusting desktop-focused Linux distributions that in turn trust thousands of individuals to ship changes without review. Trusting trust attacks are a ticking time bomb, and starting to happen in the real world as recently as this week (Sha1-Hulud).

Within a week or two stagex will be the second LLVM-native distribution, and the first to have a decentralized trust model. There should be no reason it can't build Monero for all major architectures including MacOS from using stagex/pallet-clang as a base after our next release any day now.

After that I think the only thing you would be missing is mingw. Some teams adjacent to yours have expressed desire for the same. No current maintainers use windows much if at all, so mingw has not been something anyone wanted to take on for fun. That said, several maintainers I am sure would be open to sponsorship to implement it. We would also happily help onboard maintainers with an interest in implementing and maintaining such things themselves.

plowsof:
@ `b50b7c6286ef0c2fc961b8a1af3b1ea0a6a765d5`

```
cb7bacc447de5f52dd7cd5a0d4221eb23be20880a28ffeaaafa7fb8ad901549b  ./aarch64-linux-android28/monero-aarch64-linux-android28-b50b7c6286ef.tar.bz2
ea6b2228333b5c087f26449606c3d7bbb50893d3958ef16a00fc7536fe7b3bf3  ./aarch64-linux-musl/monero-aarch64-linux-musl-b50b7c6286ef.tar.bz2
3fd79d88f8032483debfcc405f96a61e77494408b4ce8270a6064a2cf7f91f6f  ./arm64-apple-darwin/monero-arm64-apple-darwin-b50b7c6286ef.tar.bz2
3a28f4742f29bb7abb926e6b28f99b469e8730e82a1d1dac07db4721299b0054  ./dist-archive/monero-source-b50b7c6286ef.tar.gz
fb7ca4cbaae27bffd2667f1a6dc818558adf610673d475c5996a325d5f108c19  ./riscv64-linux-musl/monero-riscv64-linux-musl-b50b7c6286ef.tar.bz2
2a52dc800cdf3245bc13dfdb11f45bf648794cb5f8ccfa079703453d5935d50f  ./x86_64-apple-darwin/monero-x86_64-apple-darwin-b50b7c6286ef.tar.bz2
45f44a803304e952482cc9a81a4f510473b2f9ea9f8a3751c452885f94a136ef  ./x86_64-linux-musl/monero-x86_64-linux-musl-b50b7c6286ef.tar.bz2
9fbeadd941b08f8091b3035d408f6e1c6c5bc1e8e010b9118856f1989085b858  ./x86_64-unknown-freebsd/monero-x86_64-unknown-freebsd-b50b7c6286ef.tar.bz2
2268d75d0c511e1090e72dc4280dff5a8c51c17e5e36527d0381740fdddffc59  ./x86_64-w64-mingw32/monero-x86_64-w64-mingw32-b50b7c6286ef.zip

```

tobtoht:
Thanks plowsof. This matches SNeedlewoods' and my hashes.

kayabaNerve:
StageX _may_ be better for its containerization and 'frontend'. However, I cannot recommend it for its quality. When I contributed, it lacked any form of quality control, shipped several binary blobs despite claiming otherwise, and felt almost actively against quality assurance. I feel their claims of not having any binaries today, except for the hex0 seed, are sufficiently probable to be misleading it's in bad faith.

If we were to use StageX, I'd say we'd have to fork it or solely discuss it from a historical snapshot. This is especially important as the de facto head is discussing using an LLM to upgrade the Rust bootstrap compiler, and how 'trusting trust' may mean trusting an LLM to write an intermediary Rust compiler correctly. This would mean not using StageX's updated package sets, releases, or signers, but maintaining our own registry, or ratcheting up to the latest version ourselves. In that case, the advantage of Guix is less clear cut.

I'm also unsure it actually supports multiple container solutions, even if it could in theory. The support for podman, while avoiding Docker, doesn't change it's tightly coupled to containerd IIRC.

I say all this as someone who was a non-trivial contibutor, who actually moved it to mimalloc, who redid their LLVM and Rust packages, and misc other errata. I adopted it in my own project, Serai, exemplifying the Dunning–Kruger effect, before realizing how flawed it was and issuing a retraction/no longer claiming Serai's runtime is bootstrapped thanks to using StageX.

---

I'm unsure we should moving entirely to static binaries with musl + mimalloc. For any musl builds though, I recommend the patched musl I contributed to StageX which super tightly integrates mimalloc.

vtnerd:
Keeping my Guix install around! I'm ready if we pivot back.

> Important as the de facto head is discussing using an LLM to upgrade the Rust bootstrap compiler, and how 'trusting trust' may mean trusting an LLM to write an intermediary Rust compiler correctly.

People have lots their minds with LLMs. Its literally the _same_ problem they are trying to eliminate with bootstrapping!

> I'm unsure we should moving entirely to static binaries with musl + mimalloc. For any musl builds though, I recommend the patched musl I contributed to StageX which super tightly integrates mimalloc.

Yes a switch to musl bothered me, as I don't think we've really done much testing with it.

jpk68:
Related (to functionality with glibc, as opposed to musl): #10882

kayabaNerve:
As a followup,

I like a _lot_ of the ideas seen present with StageX. Unfortunately, I find the implementation half-baked, and as I said before, believe their advertised claims border on bad faith, which caused me (who used to contribute) to leave the project (meaning I'm largely uninformed of the past several months). This, combined with how it seems to be declining _in my opinion_, is why I unfortunately have to be very wary against suggesting it here.

I think we _could_ reasonably fork StageX, as we only need a subset of packages, to achieve bootstrapping. I don't think that is unacceptable. I just think it's scope creep, when one of the reasons for adopting StageX was they are better about updating their packages.

---

@vtnerd For the devil's advocate, the LLM's output could be reviewed. Replacing a large, historical toolchain with a smaller, modern-day alternative _is_ good practice when bootstrapping. However, I personally have a definition of bootstrapping which would exclude any machine-generated files, and therefore exclude LLM output. My opinion seems closer to your own take.

That strict line is _not_ used in my above commentary though where I allege the use of binary blobs. I personally added an allowlist to the Rust package for file extensions, and added checks they were recognized as text files, and found _several_ binaries laying around. To be honest, I found them all reasonable, and I wouldn't say they're more problematic than other ecosystems (this isn't a "Rust problem"). Most weren't actually used, and I believe the problematic ones were a few autogenerated source files, a few precompiled RegExs (which apparently were represented as bytecode, due to a super cool work which compiles RegExs to finite automatons and guarantees computational complexity for their evaluation), and font files. My issue is with how such strict filtering hasn't been extended as quality assurance to all of the other packages, despite a track record of similar incidents in other packages.

---

Re: musl, I can't comment too much on the stability of it. If musl + mimalloc is stable, and the performance benefits justify making it THE official distribution, great.

In general, I am strongly against application-level general-purpose allocators though, such as jemalloc. They remove _my_ system's choice of allocator with the idea they know better, and applications should not take the role of the system (from a principle standpoint). I understand system allocators frequently suck, but that's a problem with the system, not for the application to solve IMO. Application allocators also frequently conflict with my system's (hardened) allocator, exacerbating the issue.

If this is all statically linked, I don't have the concerns about conflicts and honestly, can be fine with the choice of mimalloc. However, there is still the 'distribution' question. Should Monero statically link a libc, instead of relying on the systems? It prevents libc updates via the system, meaning Monero has to issue a new build for libc updates. This also isn't inherent to the idea of using musl, as Alpine does the fun, confusing thing of dynamically linking musl*. This is again the distribution question, and asking where the _pragmatic_ line is drawn between the system and application.

*I think this makes sense for Alpine. It's only confusing as people don't generally consider using musl except to achieve static linking.

kayabaNerve:
@SyntheticBird45 and I have planned to maintain a fork of StageX, to be available for the Monero, Cuprate, and Serai projects: https://github.com/radiko-bootstrap/radiko

As a literal fork of StageX, it should be a drop-in replacement question right now. The sole question is if people would prefer StageX or Radiko as maintainers moving forward, where Radiko will be maintained _as we have time_, but with a clear goal of quality assurance rather than features not relevant to the named projects.

https://github.com/radiko-bootstrap/radiko/pull/1 is my proposal for the project's definition and goals moving forward. It is currently open to comments.

PR #10228, opened 2025-11-23
.

pabloecz:


x64x2:
write what you did in your PR instead of letting it be blank.


_Originally posted by @x64x2 in https://github.com/monero-project/monero/pull/10229#issuecomment-3568821338_

PR #10232, opened 2025-11-24
remove cached password from Wallet API (master)

SNeedlewoods:
Based on https://github.com/monero-project/monero/pull/9464

https://github.com/monero-project/monero/pull/9915 is the old release version of this PR, I plan to update later

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -33,6 +33,8 @@
 #include "device/device.hpp"
 #include "crypto/crypto.h"
 #include "enote_details.h"
+#include "device/device.hpp"
+#include "net/net_ssl.h"
 #include "pending_transaction.h"
 #include "string_tools.h"
 #include "unsigned_transaction.h"
@@ -46,6 +48,7 @@
 
 #include "mnemonics/electrum-words.h"
 #include "mnemonics/english.h"
+#include "wallet/fee_priority.h"
 #include <boost/format.hpp>
 #include <sstream>
 #include <unordered_map>
@@ -198,13 +201,11 @@ struct Wallet2CallbackImpl : public tools::i_wallet2_callback
         }
     }
 
-    virtual void on_money_received(uint64_t height, const crypto::hash &txid,
-        const cryptonote::transaction& tx, uint64_t amount, uint64_t burnt,
-        const cryptonote::subaddress_index& subaddr_index, const crypto::hash &payment_id,
-        bool is_change, uint64_t unlock_time)
+    virtual void on_money_received(uint64_t height, const crypto::hash &txid, const cryptonote::transaction& tx, uint64_t amount, uint64_t burnt, const cryptonote::subaddress_index& subaddr_index, const crypto::hash &payment_id, bool is_change, uint64_t unlock_time, const crypto::public_key& enote_pub_key)
     {
 
         std::string tx_hash =  epee::string_tools::pod_to_hex(txid);
+        std::string enote_pub_key_str = epee::string_tools::pod_to_hex(enote_pub_key);
 
         LOG_PRINT_L3(__FUNCTION__ << ": money received. height:  " << height
                      << ", tx: " << tx_hash
@@ -212,9 +213,9 @@ struct Wallet2CallbackImpl : public tools::i_wallet2_callback
                      << ", burnt: " << print_money(burnt)
                      << ", raw_output_value: " << print_money(amount)
                      << ", idx: " << subaddr_index);
-        // do not signal on received tx if wallet is not synchronized completely
-        if (m_listener && m_wallet->synchronized()) {
-            m_listener->moneyReceived(tx_hash, amount - burnt);
+        // do not signal on received tx if unattended wallet is not synchronized completely
+        if (m_listener && (m_wallet->synchronized() || !m_wallet->m_wallet->is_unattended())) {
+            m_listener->moneyReceived(tx_hash, amount - burnt, burnt, enote_pub_key_str, is_change, cryptonote::is_coinbase(tx));
             m_listener->updated();
         }
     }
@@ -228,25 +229,27 @@ struct Wallet2CallbackImpl : public tools::i_wallet2_callback
                      << ", tx: " << tx_hash
                      << ", amount: " << print_money(amount)
                      << ", idx: " << subaddr_index);
-        // do not signal on received tx if wallet is not synchronized completely
-        if (m_listener && m_wallet->synchronized()) {
+        // do not signal on received tx if unattended wallet is not synchronized completely
+        if (m_listener && (m_wallet->synchronized() || !m_wallet->m_wallet->is_unattended())) {
             m_listener->unconfirmedMoneyReceived(tx_hash, amount);
             m_listener->updated();
         }
     }
 
     virtual void on_money_spent(uint64_t height, const crypto::hash &txid, const cryptonote::transaction& in_tx,
-                                uint64_t amount, const cryptonote::transaction& spend_tx, const cryptonote::subaddress_index& subaddr_index)
+                                uint64_t amount, const cryptonote::transaction& spend_tx, const cryptonote::subaddress_index& subaddr_index,
+                                const crypto::public_key& enote_pub_key)
     {
         // TODO;
         std::string tx_hash = epee::string_tools::pod_to_hex(txid);
+        std::string enote_pub_key_str = epee::string_tools::pod_to_hex(enote_pub_key);
         LOG_PRINT_L3(__FUNCTION__ << ": money spent. height:  " << height
                      << ", tx: " << tx_hash
                      << ", amount: " << print_money(amount)
                      << ", idx: " << subaddr_index);
-        // do not signal on sent tx if wallet is not synchronized completely
-        if (m_listener && m_wallet->synchronized()) {
-            m_listener->moneySpent(tx_hash, amount);
+        // do not signal on sent tx if unattended wallet is not synchronized completely
+        if (m_listener && (m_wallet->synchronized() || !m_wallet->m_wallet->is_unattended())) {
+            m_listener->moneySpent(tx_hash, amount, enote_pub_key_str, std::make_pair(subaddr_index.major, subaddr_index.minor));
             m_listener->updated();
         }
     }
@@ -331,6 +334,19 @@ struct Wallet2CallbackImpl : public tools::i_wallet2_callback
     WalletImpl     * m_wallet;
 };
 
+WalletKeysDecryptGuard::~WalletKeysDecryptGuard() {}
+
+struct WalletKeysDecryptGuardImpl: public WalletKeysDecryptGuard
+{
+    WalletKeysDecryptGuardImpl(tools::wallet2 &w, const epee::wipeable_string &password):
+        u(w, &password)
+    {}
+
+    ~WalletKeysDecryptGuardImpl() override = default;
+
+    tools::wallet_keys_unlocker u;
+};
+
 Wallet::~Wallet() {}
 
 WalletListener::~WalletListener() {}
@@ -371,6 +387,20 @@ bool Wallet::paymentIdValid(const string &paiment_id)
     if (tools::wallet2::parse_long_payment_id(paiment_id, pid))
         return true;
     return false;
+}
+
+bool Wallet::isSubaddress(const std::string &address, NetworkType nettype)
+{
+  cryptonote::address_parse_info info;
+  return get_account_address_from_str(info, static_cast<cryptonote::network_type>(nettype), address) ? info.is_subaddress : false;
+}
+
+std::string Wallet::getAddressFromIntegrated(const std::string &address, NetworkType nettype)
+{
+  cryptonote::address_parse_info info;
+  if (!get_account_address_from_str(info, static_cast<cryptonote::network_type>(nettype), address))
+    return "";
+  return get_account_address_as_str(static_cast<cryptonote::network_type>(nettype), info.is_subaddress, info.address);
 }
 
 bool Wallet::addressValid(const std::string &str, NetworkType nettype)
@@ -463,7 +493,7 @@ void Wallet::error(const std::string &category, const std::string &str) {
 }
 
 ///////////////////////// WalletImpl implementation ////////////////////////
-WalletImpl::WalletImpl(NetworkType nettype, uint64_t kdf_rounds)
+WalletImpl::WalletImpl(NetworkType nettype, uint64_t kdf_rounds, const bool unattended /* = true */)
     :m_wallet(nullptr)
     , m_status(Wallet::Status_Ok)
     , m_wallet2Callback(nullptr)
@@ -473,8 +503,9 @@ WalletImpl::WalletImpl(NetworkType nettype, uint64_t kdf_rounds)
     , m_rebuildWalletCache(false)
     , m_is_connected(false)
     , m_refreshShouldRescan(false)
+    , m_kdf_rounds(kdf_rounds)
 {
-    m_wallet.reset(new tools::wallet2(static_cast<cryptonote::network_type>(nettype), kdf_rounds, true));
+    m_wallet.reset(new tools::wallet2(static_cast<cryptonote::network_type>(nettype), kdf_rounds, unattended));
     m_history.reset(new TransactionHistoryImpl(this));
     m_wallet2Callback.reset(new Wallet2CallbackImpl(this));
     m_wallet->callback(m_wallet2Callback.get());
@@ -512,7 +543,7 @@ WalletImpl::~WalletImpl()
     LOG_PRINT_L1(__FUNCTION__ << " finished");
 }
 
-bool WalletImpl::create(const std::string &path, const std::string &password, const std::string &language)
+bool WalletImpl::create(const std::string &path, const std::string &password, const std::string &language, bool create_address_file /* = false */, bool non_deterministic /* = false */)
 {
 
     clearStatus();
@@ -533,13 +564,13 @@ bool WalletImpl::create(const std::string &path, const std::string &password, co
         setStatusCritical(error);
         return false;
     }
-    setSeedLanguage(language);
+    if (!non_deterministic)
+        setSeedLanguage(language);
     if (!statusOk())
         return false;
     crypto::secret_key recovery_val, secret_key;
     try {
-        recovery_val = m_wallet->generate(path, password, secret_key, false, false);
-        m_password = password;
+        recovery_val = m_wallet->generate(path, password, secret_key, /* recover */ false, non_deterministic, create_address_file);
         clearStatus();
     } catch (const std::exception &e) {
         LOG_ERROR("Error creating wallet: " << e.what());
@@ -708,7 +739,7 @@ bool WalletImpl::recoverFromKeysWithPassword(const std::string &path,
             LOG_PRINT_L1("Generated new view only wallet from keys");
         }
         if(has_spendkey && !has_viewkey) {
-           m_wallet->generate(path, password, spendkey, true, false);
+           m_wallet->generate(path, password, spendkey, /* recover */ true, /* non-deterministic */ false);
            setSeedLanguage(language);
            if (!statusOk())
                return false;
@@ -721,6 +752,70 @@ bool WalletImpl::recoverFromKeysWithPassword(const std::string &path,
         return false;
     }
     m_password = password;
+    m_wallet->set_ring_database(get_default_ringdb_path(m_wallet->nettype()));
+    return true;
+}
+
+bool WalletImpl::createFromJson(const std::string &json_file_path, std::string &pw_out)
+{
+    try
+    {
+        boost::program_options::variables_map vm;
+        boost::program_options::store(
+            boost::program_options::command_line_parser({
+                nettype() == TESTNET ? "--testnet" : nettype() == STAGENET ? "--stagenet" : "",
+            }).options([]{
+                boost::program_options::options_description options_description{};
+                tools::wallet2::init_options(options_description);
+                return options_description;
+            }()
+            ).run(),
+            vm
+        );
+        auto r = m_wallet->make_from_json(vm, m_wallet->is_unattended(), json_file_path, /* password_promper */ {});
+        if (!r.first) {
+            setStatusError(tr("failed to generate new wallet from json"));
+            return false;
+        }
+        m_wallet = std::move(r.first);
+        pw_out = std::string(r.second.password().data(), r.second.password().size());
+        LOG_PRINT_L1("Generated new wallet from json");
+    }
+    catch (const std::exception& e) {
+        setStatusError(string(tr("failed to generate new wallet from json: ")) + e.what());
+        return false;
+    }
+    m_wallet->set_ring_database(get_default_ringdb_path(m_wallet->nettype()));
+    return true;
+}
+
+bool WalletImpl::recoverFromMultisigSeed(const std::string &path,
+                                         const std::string &password,
+                                         const std::string &language,
+                                         const std::string &multisig_seed,
+                                         const std::string seed_pass /* = "" */,
+                                         const bool do_create_address_file /* = false */)
+{
+    try
+    {
+        if (seed_pass.empty())
+            m_wallet->generate(path, password, multisig_seed, do_create_address_file);
+        else
+        {
+            crypto::secret_key key;
+            crypto::cn_slow_hash(seed_pass.data(), seed_pass.size(), (crypto::hash&)key);
+            sc_reduce32((unsigned char*)key.data);
+            const epee::wipeable_string &msig_keys = m_wallet->decrypt<epee::wipeable_string>(std::string(multisig_seed.data(), multisig_seed.size()), key, true);
+            m_wallet->generate(path, password, msig_keys, do_create_address_file);
+        }
+        setSeedLanguage(language);
+        LOG_PRINT_L1("Generated new multisig wallet from multisig seed");
+    }
+    catch (const std::exception& e) {
+        setStatusError(string(tr("failed to generate new multisig wallet: ")) + e.what());
+        return false;
+    }
+    m_wallet->set_ring_database(get_default_ringdb_path(m_wallet->nettype()));
     return true;
 }
 
@@ -853,22 +948,6 @@ std::string WalletImpl::seed(const std::string& seed_offset) const
     if (m_wallet)
         m_wallet->get_seed(seed, seed_offset);
     return std::string(seed.data(), seed.size()); // TODO
-}
-
-std::string WalletImpl::getSeedLanguage() const
-{
-    return m_wallet->get_seed_language();
-}
-
-void WalletImpl::setSeedLanguage(const std::string &arg)
-{
-    if (checkBackgroundSync("cannot set seed language"))
-        return;
-
-    if (crypto::ElectrumWords::is_valid_language(arg))
-        m_wallet->set_seed_language(arg);
-    else
-        setStatusError(string(tr("Failed to set seed language. Language not valid: ")) + arg);
 }
 
 int WalletImpl::status() const
@@ -1020,21 +1099,9 @@ bool WalletImpl::init(const std::string &daemon_address, uint64_t upper_transact
     return doInit(daemon_address, proxy_address, upper_transaction_size_limit, use_ssl);
 }
 
-void WalletImpl::allowMismatchedDaemonVersion(bool allow_mismatch)
-{
-    m_wallet->allow_mismatched_daemon_version(allow_mismatch);
-}
-
 void WalletImpl::setRingDatabase(const std::string &path)
 {
     m_wallet->set_ring_database(path);
-}
-
-void WalletImpl::setRefreshFromBlockHeight(uint64_t refresh_from_block_height)
-{
-    if (checkBackgroundSync("cannot change refresh height"))
-        return;
-    m_wallet->set_refresh_from_block_height(refresh_from_block_height);
 }
 
 void WalletImpl::setRecoveringFromSeed(bool recoveringFromSeed)
@@ -1047,19 +1114,26 @@ void WalletImpl::setRecoveringFromDevice(bool recoveringFromDevice)
     m_recoveringFromDevice = recoveringFromDevice;
 }
 
-void WalletImpl::setSubaddressLookahead(uint32_t major, uint32_t minor)
-{
-    m_wallet->set_subaddress_lookahead(major, minor);
-}
-
 uint64_t WalletImpl::balance(uint32_t accountIndex) const
 {
     return m_wallet->balance(accountIndex, false);
 }
 
-uint64_t WalletImpl::unlockedBalance(uint32_t accountIndex) const
+std::map<uint32_t, uint64_t> WalletImpl::balancePerSubaddress(uint32_t accountIndex /* = 0 */) const
 {
-    return m_wallet->unlocked_balance(accountIndex, false);
+    return m_wallet->balance_per_subaddress(accountIndex, false);
+}
+
+uint64_t WalletImpl::unlockedBalance(uint32_t accountIndex /* = 0 */,
+                                     uint64_t *blocks_to_unlock /* = NULL */,
+                                     uint64_t *time_to_unlock /* = NULL */) const
+{
+    return m_wallet->unlocked_balance(accountIndex, false, blocks_to_unlock, time_to_unlock);
+}
+
+std::map<uint32_t, std::pair<uint64_t, std::pair<uint64_t, uint64_t>>> WalletImpl::unlockedBalancePerSubaddress(uint32_t accountIndex /* = 0 */) const
+{
+    return m_wallet->unlocked_balance_per_subaddress(accountIndex, false);
 }
 
 uint64_t WalletImpl::blockChainHeight() const
@@ -1124,13 +1198,17 @@ bool WalletImpl::synchronized() const
     return m_synchronized;
 }
 
-bool WalletImpl::refresh()
+bool WalletImpl::refresh(std::uint64_t start_height /* = 0 */,
+                         bool check_pool /* = true */,
+                         bool try_incremental /* = false */,
+                         std::uint64_t max_blocks /* = std::numeric_limits<uint64_t>::max() */,
+                         std::uint64_t *blocks_fetched_out /* = nullptr */,
+                         bool *received_money_out /* = nullptr */)
 {
     clearStatus();
-    //TODO: make doRefresh return bool to know whether the error occurred during refresh or not
-    //otherwise one may try, say, to send transaction, transfer fails and this method returns false
-    doRefresh();
-    return statusOk();
+    bool did_error_occur;
+    doRefresh(start_height, check_pool, try_incremental, max_blocks, &did_error_occur, blocks_fetched_out, received_money_out);
+    return !did_error_occur;
 }
 
 void WalletImpl::refreshAsync()
@@ -1140,21 +1218,26 @@ void WalletImpl::refreshAsync()
     m_refreshCV.notify_one();
 }
 
-bool WalletImpl::rescanBlockchain()
+bool WalletImpl::rescanBlockchain(bool do_hard_rescan /* = false */, bool do_keep_key_images /* = false */, bool do_skip_refresh /* = false */)
 {
     if (checkBackgroundSync("cannot rescan blockchain"))
         return false;
     clearStatus();
     m_refreshShouldRescan = true;
-    doRefresh();
+    m_do_hard_rescan = do_hard_rescan;
+    m_do_keep_key_images_on_rescan = do_keep_key_images;
+    if (!do_skip_refresh)
+        doRefresh();
     return statusOk();
 }
 
-void WalletImpl::rescanBlockchainAsync()
+void WalletImpl::rescanBlockchainAsync(bool do_hard_rescan /* = false */, bool do_keep_key_images /* = false */)
 {
     if (checkBackgroundSync("cannot rescan blockchain"))
         return;
     m_refreshShouldRescan = true;
+    m_do_hard_rescan = do_hard_rescan;
+    m_do_keep_key_images_on_rescan = do_keep_key_images;
     refreshAsync();
 }
 
@@ -1229,7 +1312,7 @@ bool WalletImpl::submitTransaction(const string &fileName) {
     setStatusError(tr("Failed to load transaction from file"));
     return false;
   }
-  
+
   if(!transaction->commit()) {
     setStatusError(transaction->m_errorString);
     return false;
@@ -1289,7 +1372,7 @@ std::string WalletImpl::exportKeyImagesAsString(bool all /* = false */)
     return "";
 }
 
-bool WalletImpl::importKeyImages(const string &filename)
+bool WalletImpl::importKeyImages(const std::string &filename, std::uint64_t *spent_out /* = nullptr */, std::uint64_t *unspent_out /* = nullptr */, std::uint64_t *import_height /* = nullptr */)
 {
   if (checkBackgroundSync("cannot import key images"))
     return false;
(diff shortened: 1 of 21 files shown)


src/wallet/api/wallet.cpp
```diff
@@ -1055,14 +1050,18 @@ void WalletImpl::stop()
...
 {
     clearStatus();
     try {
         if (path.empty()) {
             m_wallet->store();
+        } else if (password) {
+            m_wallet->store_to(path, epee::wipeable_string((*password).first, (*password).second));
         } else {
-            m_wallet->store_to(path, m_password);
+            setStatusError("Failed to store wallet to new path without password.");
```
iamamyth: This seems to imply an intermittent or recoverable failure; suggest something like: `Saving the wallet to a new path requires a password.`

vtnerd:
Doesn't this require the UI to either store the password itself or reprompt everytime it needs to store? This seems like a step backwards considering the UI; all of this password wiping is likely for little benefit.

rbrunner7:
While looking into this and reviewing I thought about @vtnerd 's comments:

> Doesn't this require the UI to either store the password itself or reprompt everytime it needs to store? This seems like a step backwards considering the UI; all of this password wiping is likely for little benefit.

My take on this: I don't think it's a step backwards and of little benefit. I see password prompting and password storing or not as an important implementation aspect that should be in the hands of the wallet implementer. IMHO the library should not force its "opinion" onto the library client as "We store, period, get over it.".

Yes of course, now the wallet app has to care about the password itself, but I don't see this additional job pushing the whole endeavor into a net negative thing.

Therefore I support this PR.

rbrunner7 (approved):
Looks good to me.

SNeedlewoods:
> Doesn't this require the UI to either store the password itself or reprompt everytime it needs to store? This seems like a step backwards considering the UI; all of this password wiping is likely for little benefit.

Mentioned it here https://github.com/monero-project/monero/pull/9915
- `store()` does only require a password if you store the wallet to a new path. Closing the wallet or using `save` command in the CLI does not prompt the user for a password [WalletImpl::store()](https://github.com/monero-project/monero/blob/48ad374b0d6d6e045128729534dc2508e6999afe/src/wallet/api/wallet.cpp#L936-L952) -> [wallet2::store()](https://github.com/monero-project/monero/blob/48ad374b0d6d6e045128729534dc2508e6999afe/src/wallet/wallet2.cpp#L6797-L6801)
- `setPassword()` now needs the old passsword as arg, but IMO it's expected UX to type 1. old password 2. new password 3. confirm new password
- `makeMultisig()` and `exchangeMultisigKeys()` now need a password arg

AFAICT this is not much worse for the UX.

This issue was brought to my attention by @j-berman ([src](https://github.com/feather-wallet/feather/issues/72#issuecomment-1405602142)):
> Wallet password caching should be removed from wallet/api. Password verification should use wallet2::verifyPassword instead of comparing against the cached value. This causes a slight delay when opening dialogs that require the password or when unlocking the wallet.

vtnerd:
> This issue was brought to my attention by @j-berman (https://github.com/feather-wallet/feather/issues/72#issuecomment-1405602142):
>
>> Wallet password caching should be removed from wallet/api. Password verification should use wallet2::verifyPassword instead of comparing against the cached value. This causes a slight delay when opening dialogs that require the password or when unlocking the wallet.

EDIT:
I originally said `wallet2` but I meant `simplewallet`, the TUI interface using all of this.

How is the scanning UI/flow going to work for `simplewallet`? It does background syncing, and then prompts for password (during inactivity or just randomly) and stores wallet info during that re-prompt? Has anyone fully completed the UI (GUI+TUI) changes to make all of this happen? It's bordering on worse user experience if the scan state cannot be reliably stored.

As an example, my experimental wallet, `lwcli` (which can use `wallet2` backend), just calls `store` in a `shared_ptr` destructor callback, to ensure that the scan state is always reliably saved before closing. This PR makes shutdown more messy, as it must prompt for password (or the UI stores the password *yikes*), and there are codepaths where this is more difficult.

iamamyth:
I tend to agree that a 50% solution on password handling offers the worst of all worlds: No consistent security floor, and an inability for callers to raise the partial floor, should they desire. In other words: If I don't like the wallet API's handling of, say, key derivation, I'm still stuck with it, but I can't even make use of that fact to offer convenience in the form of consistency, because the API now offloads some prompting duties to the caller. The API needs to pick a lane: Get fully out of the way (e.g. password engine fully pluggable, perhaps with a default implementation; or anything password-related removed from the API, implying persistence based on streams rather than the filesystem), or provide a non-replaceable implementation that makes all the password security decisions.

src/wallet/api/wallet2_api.h
```diff
@@ -59,6 +64,44 @@ enum NetworkType : uint8_t {
...
+* brief: EnoteDetails - Container for all the necessary information that belongs to an enote
+*/
+struct EnoteDetails
+{
+    enum TxProtocol {
+        TxProtocol_CryptoNote,
+        TxProtocol_RingCT
+    };
+
+    virtual ~EnoteDetails() = 0;
```
vtnerd, line 78: These `std::string`s are hex or binary? Might want to add a comment about that.

src/wallet/api/wallet2_api.h
```diff
@@ -94,6 +141,68 @@ struct PendingTransaction
...
+    * return: unsigned tx data as encrypted hex string if succeeded, else empty string
+    * note: sets status error on failure
+    */
+    virtual std::string convertTxToStr() = 0;
+    /**
+    * brief: convertTxToRawBlobStr - convert each tx in PendingTransaction to hex string
+    * return: serialized tx data as raw hex strings if succeeded, else empty vector
+    * note: sets status error on failure
+    */
+    virtual std::vector<std::string> convertTxToRawBlobStr() = 0;
```
vtnerd, line 275: How does the backend encrypt the transaction? This should require that the backend cache the password ?

src/wallet/api/wallet2_api.h
```diff
@@ -94,6 +141,68 @@ struct PendingTransaction
...
+    /**
+    * brief: getWorstFeePerByte - needed when checking for backlog
+    * return: worst fee per bytes
+    */
+    virtual double getWorstFeePerByte() const = 0;
+    /**
+    * brief: vinSizes -
+    * return: number of enotes used as input per tx
+    */
+    virtual std::vector<size_t> vinSizes() const = 0;
```
vtnerd: Why have `vinSizes()`, `vinOffsets()`, `vinKeyImages()`, and `vinAmounts()` when this information should be attainable through `getEnoteDetailsIn()`?

src/wallet/api/wallet2_api.h
```diff
@@ -165,6 +295,13 @@ struct TransactionInfo
         Direction_Out
     };
 
+    enum TxState {
```
vtnerd: Why not call declare this a `enum class TxState`? If you don't need conversion to int, the class declaration is preferred.

src/wallet/api/wallet2_api.h
```diff
@@ -444,18 +613,71 @@ struct Wallet
         BackgroundSync_CustomPassword = 2
     };
 
+    enum AskPasswordType {
```
vtnerd: I would make all of these `enum class` as well.

src/wallet/api/wallet2_api.h
```diff
@@ -444,18 +613,71 @@ struct Wallet
...
-    virtual void setSeedLanguage(const std::string &arg) = 0;
     //! returns wallet status (Status_Ok | Status_Error)
     virtual int status() const = 0; //deprecated: use safe alternative statusWithErrorString
     //! in case error status, returns error string
     virtual std::string errorString() const = 0; //deprecated: use safe alternative statusWithErrorString
     //! returns both error and error string atomically. suggested to use in instead of status() and errorString()
     virtual void statusWithErrorString(int& status, std::string& errorString) const = 0;
-    virtual bool setPassword(const std::string &password) = 0;
-    virtual const std::string& getPassword() const = 0;
+    virtual bool setPassword(const char *old_password, const std::size_t old_pw_length, const char *new_password, const std::size_t new_pw_length) = 0;
```
vtnerd: I would switch this to `std::string_view` for both `old_password` and `new_password` since we are on C++17 for this target branch. It's still fairly simple for any client to conform - its just `setPassword({old, old_len}, {new, new_len})`, etc.

And if using `std::string` its just `setPassword(old, new)`.

src/wallet/api/wallet2_api.h
```diff
@@ -519,9 +741,10 @@ struct Wallet
      * \brief store - stores wallet to file.
      * \param path - main filename to store wallet to. additionally stores address file and keys file.
      *               to store to the same file - just pass empty string;
+     * \param password - [pointer to password, password length], only needed if path is not empty (default: empty optional)
      * \return
      */
-    virtual bool store(const std::string &path) = 0;
+    virtual bool store(const std::string &path, const optional<std::pair<const char *, const std::size_t>> &password = optional<std::pair<const char *, const std::size_t>>()) = 0;
```
vtnerd: `optional<std::string_view>`, which I assume is meant to differentiate from a blank password and no password.

src/wallet/api/wallet2_api.h
```diff
@@ -710,9 +931,20 @@ struct Wallet
 
     /**
      * @brief refresh - refreshes the wallet, updating transactions from daemon
+     * @param start_height - (Default: 0)
```
vtnerd: I think it would be helpful to describe what some of these parameters do, as it's non-obvious. A good example is `try_incremental`.

src/wallet/api/wallet2_api.h
```diff
@@ -723,12 +955,12 @@ struct Wallet
      * @brief rescanBlockchain - rescans the wallet, updating transactions from daemon
      * @return - true if refreshed successfully;
      */
-    virtual bool rescanBlockchain() = 0;
+    virtual bool rescanBlockchain(bool do_hard_rescan = true, bool do_keep_key_images = false, bool do_skip_refresh = false) = 0;
```
vtnerd: What does `do_hard_rescan` toggle?

src/wallet/api/wallet2_api.h
```diff
@@ -790,25 +1022,31 @@ struct Wallet
      * @brief makeMultisig - switches wallet in multisig state. The one and only creation phase for N / N wallets
      * @param info - vector of multisig infos from other participants obtained with getMulitisInfo call
      * @param threshold - number of required signers to make valid transaction. Must be <= number of participants
+     * @param password - wallet password
+     * @param pw_length - password length
      * @return in case of N / N wallets returns empty string since no more key exchanges needed. For N - 1 / N wallets returns base58 encoded extra multisig info
      */
-    virtual std::string makeMultisig(const std::vector<std::string>& info, uint32_t threshold) = 0;
+    virtual std::string makeMultisig(const std::vector<std::string>& info, uint32_t threshold, const char *password, const std::size_t pw_length) = 0;
```
vtnerd: Another good candidate for `std::string_view`, and two more in this file.

src/wallet/api/wallet2_api.h
```diff
@@ -1116,13 +1404,21 @@ struct Wallet
     //! returns true if the keys file is locked
     virtual bool isKeysFileLocked() = 0;
 
+    /**
+     * \brief Decrypt spend-keys if not already decrypted and create a keys decryption guard
+     * \param password password to decrypt in-memory spend-keys
+     * \param pw_length length of specified password
+     * \return Keys decryption guard
+     */
+    virtual WalletKeysDecryptGuard * createKeysDecryptGuard(const char *password, const std::size_t pw_length) = 0;
```
vtnerd: This function complicates the interface - it's not clear when it's supposed to be invoked. If anything this suggests that the backend should be caching the password to make things simpler. This should probably a `std::unique_ptr<WalletKeysDecryptGuard>` too. 

Also, does this implementation handle lifetimes properly? My lwsf implementation likely can handle this because it has a "backend" that uses `std::shared_ptr` and `WalletKeysDecryptGuard` can used that `shared_ptr`. However, the primary implementation to this interface doesn't have such a design, so it seems like dangling pointers could easily result if this guard exceeds the lifetime of a `Wallet` object.

And this is another candidate for `std::string_view`.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    */
+    virtual void getEnoteDetails(std::vector<std::unique_ptr<EnoteDetails>> &enotes_details) const = 0;
+    /**
+    * brief: getEnoteDetails - get information about a single enote
+    * param: enote_pub_key - ephemeral public key
+    * outparam: enote_details -
+    * return: true if succeeded
+    * note: sets status error on failure
+    */
+    virtual bool getEnoteDetails(const std::string &enote_pub_key, std::unique_ptr<EnoteDetails> &enote_details) const = 0;
```
vtnerd: Just return the `std::unique_ptr` and return `nullptr` if failed.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    * param: refreshed - (default: false)
+    * param: try_incremental - (default: false)
+    * note: sets status error on failure
+    */
+    virtual void refreshPoolOnly(bool refreshed = false, bool try_incremental = false) = 0;
+    /**
+    * brief: getEnoteDetails - get information about all enotes
+    * outparam: enotes_details -
+    */
+    virtual void getEnoteDetails(std::vector<std::unique_ptr<EnoteDetails>> &enotes_details) const = 0;
```
vtnerd: I would return this single value instead of out parameter, unless the objective is to re-use `std::vector` memory (reduce allocations). This seems unlikely.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    */
+    virtual void rewriteWalletFile(const std::string &wallet_name, const char *password, const std::size_t pw_length) = 0;
+    /**
+    * brief: writeWatchOnlyWallet -  create a new watch-only wallet file with view keys from current wallet
+    * param: password - password for new watch-only wallet
+    * param: pw_length - password length
+    * outparam: new_keys_file_name - wallet_name + "-watchonly.keys"
+    * note: sets status error on failure
+    */
+    virtual void writeWatchOnlyWallet(const char *password, const std::size_t pw_length, std::string &new_keys_file_name) = 0;
```
vtnerd: Another candidate for `std::string_view`.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    */
+    virtual DeviceState getDeviceState() const = 0;
+    /**
+    * brief: rewriteWalletFile - rewrite the wallet file for wallet update
+    * param: wallet_name - name of the wallet file (should exist)
+    * param: password - wallet password
+    * param: pw_length - password length
+    * note: sets status error on failure
+    */
+    virtual void rewriteWalletFile(const std::string &wallet_name, const char *password, const std::size_t pw_length) = 0;
```
vtnerd: `std::string_view` here too.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    */
+    virtual PendingTransaction* parseTxFromStr(const std::string &signed_tx_str) = 0;
+    /**
+    * brief: parseMultisigTxFromStr - get pending multisig transaction from encrypted unsigned multisig transaction as hex string
+    * param: multisig_tx_str -
+    * outparam: exported_txs -
+    * return: true if succeeded
+    * note: sets status error on failure
+    */
+    virtual bool parseMultisigTxFromStr(const std::string &multisig_tx_str, PendingTransaction &exported_txs) const = 0;
```
vtnerd: Why `bool` return with out parameter here, but not in above method? The consistent interface would be to just return the `PendingTransaction*` as thats the existing style.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    * param: additional_tx_keys -
+    * param: single_destination_subaddress -
+    * note: sets status error on failure
+    */
+    virtual void setTxKey(const std::string &txid, const std::string &tx_key, const std::vector<std::string> &additional_tx_keys, const std::string &single_destination_subaddress) = 0;
+    /**
+    * brief: getAccountTags - get the list of registered account tags
+    * return: first.Key=(tag's name), first.Value=(tag's label), second[i]=(i-th account's tag)
+    */
+    virtual const std::pair<std::map<std::string, std::string>, std::vector<std::string>>& getAccountTags() const = 0;
```
vtnerd, line 1865: Reading the comments suggest this return value should be `std::vector<std::pair<std::string, std::string>>`.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    */
+    virtual std::vector<std::pair<std::uint64_t, std::uint64_t>> estimateBacklog(const std::vector<std::pair<double, double>> &fee_levels) const = 0;
+    /**
+    * brief: saveToFile   - save hex string to file
+    * param: path_to_file - file name
+    * param: binary       - hex string data
+    * param: is_printable - (default: false)
+    * return: true if succeeded
+    */
+    virtual bool saveToFile(const std::string &path_to_file, const std::string &binary, bool is_printable = false) const = 0;
```
vtnerd: What is this doing? Is it just storing `binary` to a `path_to_file`? This doesn't seem like it should be included? Or a better description needs to be provided.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    /**
+    * brief: loadFromFile  - load hex string from file
+    * param: path_to_file  - file name
+    * outparam: target_str - hex string data
+    * param: max_size      - maximum size in bytes (default: 1000000000)
+    * return: true if succeeded
+    */
+    virtual bool loadFromFile(const std::string &path_to_file, std::string &target_str, std::size_t max_size = 1000000000) const = 0;
```
vtnerd: Same thing with this, why include it with this interface? Just use `std::ofstream` ?

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    /**
+    * brief: hashEnotes - get hash of all enotes in local enote store up until `enote_idx`
+    *                     (formerly in wallet2: `uint64_t wallet2::hash_m_transfers(boost::optional<uint64_t> transfer_height, crypto::hash &hash)`)
+    * param: enote_idx - include all enotes below this index
+    * outparam: hash - hash as hex string
+    * return: number of hashed enotes
+    * note: sets status error on failure
+    */
+    virtual std::uint64_t hashEnotes(std::uint64_t enote_idx, std::string &hash) const = 0;
```
vtnerd, line 1929: What does this function do? It takes one `enote_idx` but returns number of hashed notes? Isn't that one? Or ?

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    /**
+    * brief: finishRescanBcKeepKeyImages  -
+    * param: enote_idx -
+    * param: hash -
+    * note: sets status error on failure
+    */
+    virtual void finishRescanBcKeepKeyImages(std::uint64_t enote_idx, const std::string &hash) = 0;
```
vtnerd, line 1938: No clue what this one does.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    * param: ssl_allow_any_cert   - (default: false)
+    * param: proxy                - (default: empty string)
+    * return: true if succeeded
+    * note: sets status error on failure
+    */
+    virtual bool setDaemon(const std::string &daemon_address,
+        const std::string &daemon_username = "",
+        const std::string &daemon_password = "",
+        bool trusted_daemon = false,
+        const std::string &ssl_support = "autodetect",
```
vtnerd: This (`ssl_support`) really needs to be an `enum`.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    * param: ssl_private_key_path - (default: empty string)
+    * param: ssl_certificate_path - (default: empty string)
+    * param: ssl_ca_file_path     - (default: empty string)
+    * param: ssl_allowed_fingerprints_str - (default: empty vector)
+    * param: ssl_allow_any_cert   - (default: false)
+    * param: proxy                - (default: empty string)
+    * return: true if succeeded
+    * note: sets status error on failure
+    */
+    virtual bool setDaemon(const std::string &daemon_address,
```
vtnerd: I think this should be merged into the existing `init` function; this function is duplicating that.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    * brief: verifyPassword -
+    * param: password   - password to verify
+    * param: pw_length  - password length, needed if password does not come from zero-terminated string, or if password contains zero bytes
+    * param: kdf_rounds - number of rounds for key derivation function (default: 1)
+    * return: true if succeeded
+    * note: sets status error on failure
+    *       This function automatically locks/unlocks the keys file as needed.
+    *       When possible use this instead of WalletManager::verifyWalletPassword()
+    */
+    virtual bool verifyPassword(const char *password, const std::size_t pw_length, std::uint64_t kdf_rounds = 1) = 0;
```
vtnerd: Using `string_view` for several more functions here (at least 3).

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    /**
+    * brief: encryptKeys - encrypt cached secret keys
+    * param: password    - wallet password
+    * param: pw_length   - password length, needed if password does not come from zero-terminated string, or if password contains zero bytes
+    */
+    virtual void encryptKeys(const char *password, const std::size_t pw_length) = 0;
```
vtnerd: is interface is tough, its not clear what functionality is no longer possible after being invoked. As an example, it should be impossible to find _real_ spends with legacy keys until `decryptKeys` is invoked. But the documentation doesn't list any lost functionality.

src/wallet/api/wallet2_api.h
```diff
@@ -1151,9 +1912,14 @@ struct WalletManager
      * \param  language       Language to be used to generate electrum seed mnemonic
      * \param  nettype        Network type
      * \param  kdf_rounds     Number of rounds for key derivation function
+     * \param  create_address_file - Create <wallet_name>.address.txt file (Default: false)
+     *                               NOTE: gets irgnored for stagenet and testnet and will create .address.txt file even if set to false
+     * \param  non_determinisitc - Whether to create a non-deterministic wallet, where the secret-spend-key and secret-view-key are both random and not in relation (Default: false)
+     *                             NOTE: this old way to create keys should not be used, except you have a good reason and you know what you do
+     * \param  unattended     Whether the wallet is unattended (GUI is unattended, CLI is not unattended) (Default: true)
```
vtnerd: What does this flag toggle though? The description describes who uses, but not what it does.

src/wallet/api/wallet2_api.h
```diff
@@ -412,8 +559,30 @@ struct WalletListener
...
      */
     virtual void onSetWallet(Wallet * wallet) { (void)wallet; };
+    /**
+    * brief: onReorg - called on blockchain reorg
+    */
+    virtual void onReorg(std::uint64_t height, std::uint64_t blocks_detached, std::size_t transfers_detached) = 0;
+    /**
+    * brief: onGetPassword - called by scan_output() to decrypt keys
+    */
+    virtual optional<std::string> onGetPassword(const char *reason) = 0;
```
vtnerd, line 721: Can wallets like, not use this? As in, is it possible to still use this API without using listeners or implementing this function at all?

The problem is that a major wallet (Cake) is using `monero_c`, and so is a new wallet (Skylight). Those projects are **not** using listeners because it's a royal pain to integrate C++ listeners to C callbacks. I've thought of a few ways to get it working, and every solution is a bit gross. If the design basically _requires_ listeners, then it's going to create downstream problems.

Also the `const char*` should probably be a `std::string` or possible a `std::string_view`.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    */
+    virtual bool getEnoteDetails(const std::string &enote_pub_key, std::unique_ptr<EnoteDetails> &enote_details) const = 0;
+    /**
+    * brief: getEnoteDetails - get information about a single enote
+    * param: enote_index - index of enote in internal storage
+    * outparam: enote_details -
+    * return: true if succeeded
+    * note: sets status error on failure
+    */
+    virtual bool getEnoteDetails(const std::size_t enote_index, std::unique_ptr<EnoteDetails> &enote_details) const = 0;
```
vtnerd: Again, just return the `std::unique_ptr`.

src/wallet/api/wallet2_api.h
```diff
@@ -1136,6 +1432,471 @@ struct Wallet
...
+    * param: password    - wallet password
+    * param: pw_length   - password length, needed if password does not come from zero-terminated string, or if password contains zero bytes
+    */
+    virtual void encryptKeys(const char *password, const std::size_t pw_length) = 0;
+    /**
+    * brief: decryptKeys - decrypt cached secret keys
+    * param: password    - wallet password
+    * param: pw_length   - password length, needed if password does not come from zero-terminated string, or if password contains zero bytes
+    */
+    virtual void decryptKeys(const char *password, const std::size_t pw_length) = 0;
```
vtnerd: Another `std::string_view`.

vtnerd (changes requested):
I primarily looked at the API and not the implementation. That is for another review.

Is the goal to get this merged before the other PR? There's lots of overlap and its not clear which should be review first (I assumed this one).

SNeedlewoods:
> Is the goal to get this merged before the other PR? There's lots of overlap and its not clear which should be review first (I assumed this one).

The order was supposed to be
1. https://github.com/monero-project/monero/pull/9464
2. this PR (based on 9464, which makes the first two commits of this PR, so just the third commit contains the changes for this PR)
3. https://github.com/monero-project/monero/pull/10233 (based on this PR, so just the fourth commit contains everything for 10233)

src/wallet/api/wallet2_api.h
SNeedlewoods: I moved the comments from `EnoteDetails.h` [src](https://github.com/monero-project/monero/pull/9464/files#diff-3f8942a1c7ee742ca768a6e27d37d149334c96419180640a0ac07b15e5650532R70-R112) and added more information. I think it's sufficient to have these comments in `wallet2_api.h` only, so there is no risk to get out of sync with the comments in `EnoteDetails.h`.
And if anyone has suggestions to improve the comments, I'm open to change them.

src/wallet/api/wallet2_api.h
SNeedlewoods: After thinking about it a little more, I could get rid of `vinSizes()` and `vinKeyImages()`, but I think we need a way to get `vinOffsets()` and `vinAmounts()`. Together they form `{amount, amount_index}` (so for txs that were created with `TxProtocol_RingCT` these `vinAmounts()` are set to `0`, only for `TxProtocol_CryptoNote` they show the actual enote amount AFAIK) for all the decoys and the real spends that were used to create the tx, currently `EnoteDetails` is only used for real spends, I would say decoys belong to a different category like maybe "ring member details".

Maybe it would be a more clean approach to add `struct RingMemberDetails` for all the information from `vinAmounts()`, `vinOffests()` and `constructionDataRealOutputIndices()`, then we would end up with two methods (`getEnoteDetailsIn()`, `getRingMemberDetails()`) and a struct (`RingMemberDetails`) instead of four methods (`getEnoteDetailsIn()`, `vinAmounts()`, `vinOffsets()` and `constructionDataRealOutputIndices()`).

BTW the `vin*()` methods are currently only used in simplewallet `process_ring_members()` which gets called only if you have the following setting `set print-ring-members 1` when creating a transaction.
Note: information from `vinSizes()[tx_idx]` is indirectly present through `getEnoteDetailsIn()[tx_idx].size()`

src/wallet/api/wallet2_api.h
SNeedlewoods: This part was first developed on `release-v0.18` branch with the intention to be used by the GUI. ([here](https://github.com/monero-project/monero/pull/9915#discussion_r2064388854) is the discussion from the PR, though that PR is terribly outdated. I currently plan to work on PRs for the release branch once the related master branch PRs are reviewed.)
    So just to keep that in mind. Will change this to `string_view` on this branch, if you think it's worth it?

src/wallet/api/wallet2_api.h
SNeedlewoods: Correct

src/wallet/api/wallet2_api.h
SNeedlewoods: Added comments, but not sure they are accurate or understandable, suggestions to improve them are welcome.

src/wallet/api/wallet2_api.h
SNeedlewoods: Added comments.


Thinking about it, maybe it's cleaner to add another `enum class RescanType { RescanType_Soft, RescanType_Hard, RescanType_SoftKeepKI };` instead of the two bools `do_hard_rescan` and `do_keep_key_images` which cause an error if you set both of them to `true`.

src/wallet/api/wallet2_api.h
SNeedlewoods: Maybe even `std::unique_ptr<PendingTransaction>`? The new `getEnoteDetails()` methods also return `unique_ptr`.

src/wallet/api/wallet2_api.h
SNeedlewoods: Copied from [wallet2.h](https://github.com/monero-project/monero/blob/4286fbe6cd6cdeeb2c1d13c52a912bdfc55c3f5c/src/wallet/wallet2.h#L1576-L1580) ([src to implementation](https://github.com/monero-project/monero/blob/4286fbe6cd6cdeeb2c1d13c52a912bdfc55c3f5c/src/wallet/wallet2.cpp#L12942-L12960)).
I tried to improve the comment instead of messing with the return type.

src/wallet/api/wallet2_api.h
SNeedlewoods: > This doesn't seem like it should be included?

I think you're right, removed `saveToFile()` and `loadFromFile()`.

src/wallet/api/wallet2_api.h
SNeedlewoods: Added more information to the comments.

`hashEnotes()` and `finishRescanBcKeepKeyImages()` are only needed in `simple_wallet::refresh_main()` if you do a soft rescan ([src](https://github.com/monero-project/monero/blob/4286fbe6cd6cdeeb2c1d13c52a912bdfc55c3f5c/src/simplewallet/simplewallet.cpp#L5725-L5759))
I tried to improve the comments, don't think we want to get rid of the functionality, though it's much complexity for a niche feature. Will think about this some more.

src/wallet/api/wallet2_api.h
SNeedlewoods: I remember considering that when I wrote `setDaemon()` but don't remember exactly why I decided against it, one reason: it's less obvious that you have to call `init()` to set a daemon, but I think it also had something to do with ssl options.
Will look into it again and merge if I can't find a good reason to keep it.

_________________________________________________________________________________
Edit:
`WalletImpl::init()` [calls](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/wallet/api/wallet.cpp#L969) `WalletImpl::doInit()` [calls](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/wallet/api/wallet.cpp#L2509) `wallet2::init()` [calls](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/wallet/wallet2.cpp#L1414) `wallet2::set_daemon()`,
but through that route we lose control over SSL options for `wallet2::set_daemon()`. We always end up with the default setting "autodetect" [src](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/wallet/wallet2.h#L1081).
Even though `WalletImpl::init()` does have a param `bool use_ssl`, which gets passed to `WalletImpl::doInit()`, there it ends up being unused.
Another minor detail: `WalletImpl::doInit()` sets the daemon to "untrusted" if it's not local ([src](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/wallet/api/wallet.cpp#L2522-L2528)).

So I think I would need to add the following params to `WalletImpl::init()` & `WalletImpl::doInit()`
```
bool trusted_daemon /* = false */,
Wallet::SSLSupport ssl_support /* = Wallet::SSLSupport::SSLSupport_Autodetect */,
const std::string &ssl_private_key_path /* = "" */,
const std::string &ssl_certificate_path /* = "" */,
const std::string &ssl_ca_file_path /* = "" */,
const std::vector<std::string> &ssl_allowed_fingerprints_str /* = {} */,
bool ssl_allow_any_cert /* = false */,
```
and then we also could remove unused `bool use_ssl`/`bool ssl`.

Seems there is no blocking reason not to merge, but I still tend to prefer having the `WalletImpl::setDaemon()` method, and leave the `doInit()`/`init()` methods as they are.
Though I'll change it if someone insists.

src/wallet/api/wallet2_api.h
SNeedlewoods: Before when using the Wallet API, `m_unattended` was always set `true` in the `wallet2` backend ([src](https://github.com/monero-project/monero/blob/4286fbe6cd6cdeeb2c1d13c52a912bdfc55c3f5c/src/wallet/api/wallet.cpp#L438), the third argument).
This setting enabled causes the keys to be stored in memory unencrypted and you don't need to take care of `WalletKeysDecryptGuard`, `en-/decryptKeys()` and `onGetPassword()`.

AFAICT the new API needs to support encrypted keys for the CLI which has `m_unattended = false` ([src](https://github.com/monero-project/monero/blob/4286fbe6cd6cdeeb2c1d13c52a912bdfc55c3f5c/src/simplewallet/simplewallet.cpp#L4779), the second argument).

TLDR: with this flag you can
1. keep to use the Wallet API with `unattended = true` and nothing should change.
2. now set `unattended = false` and then have to take care of de- encrypting keys

This is a complex topic, could be mistaken here, I'm open to feedback.

src/wallet/api/wallet2_api.h
SNeedlewoods: Both methods are only needed when wallet setting `unattended = false` and they're called only once in simplewallet when changing `set ask-password` [src](https://github.com/monero-project/monero/blob/4286fbe6cd6cdeeb2c1d13c52a912bdfc55c3f5c/src/simplewallet/simplewallet.cpp#L2545-L2548).

src/wallet/api/wallet2_api.h
iamamyth: Regarding the unattended flag, I note the following incantation pretty much everywhere:
```
m_ask_password == AskPasswordToDecrypt && !m_unattended && !m_watch_only
```

I wonder whether this rote repetition hides a dubious, potentially inconsistent domain model: watch_only makes sense as a flag, no problem there. But, does it ever make sense to have an "attended", watch-only wallet? And, more to the point, what does it mean for an "unattended" wallet to ask for passwords versus not? Something seems off here.

src/wallet/api/wallet2_api.h
jeffro256: > This function complicates the interface - it's not clear when it's supposed to be invoked.

Would adding a more detailed doc fix this?

> If anything this suggests that the backend should be caching the password to make things simpler.

I mean, yeah, that's basically what `wallet_keys_unlocker` does, except that the password is converted into a Chacha key.

> This should probably a `std::unique_ptr<WalletKeysDecryptGuard>` too.

Agreed.

> Also, does this implementation handle lifetimes properly?

Nope, AFAIK, `wallet_keys_unlocker` does not, but that's a change separate from this PR IMO.

src/wallet/api/wallet2_api.h
SNeedlewoods: I'm not sure I understand, but I think `convertTxToRawBlobStr()` does not encrypt the tx (uses `cryptonote::tx_to_blob()` [src](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/cryptonote_basic/cryptonote_format_utils.cpp#L1572-L1580)).
Though `convertTxToStr()` does encrypt the tx with the view key (uses `wallet2::dump_tx_to_str()` [src](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/wallet/wallet2.cpp#L7698)).

Another thing I just noticed, some strings I called "hex strings" in the comments are actually "octal escape sequence strings".
IMO it makes sense to keep this string representation, but update the comments, because e.g. in current `simple_wallet::export_raw_multisig()` the tx gets stored as octal escape sequence string, coming from `cryptonote::tx_to_blob()` ([src](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/simplewallet/simplewallet.cpp#L1716)).
Whereas in `simple_wallet::commit_or_save()` the tx is stored as hex [src](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/simplewallet/simplewallet.cpp#L10390) and we can easily get the hex string by calling `epee::string_tools::buff_to_hex_nodelimer(octal_escape_sequence_str)`.

And I tried to improve the comments in general.

src/wallet/api/wallet2_api.h
SNeedlewoods: I'm afraid my answer will just add noise, so maybe someone can help identify if something has to change in regards to the `unattended` setting.

> But, does it ever make sense to have an "attended", watch-only wallet?

AFAICT these are the possible combinations of settings and the resulting states for the keys in memory
svk = secret view key
ssk = secret spend key
|ask_password_type|unattended|watch-only|svk|ssk|
|-|-|-|-|-|
|never/action/decrypt|true/false|true|decrypted|-|
|never/action/decrypt|true|false|decrypted|decrypted|
|never/action|false|false|decrypted|decrypted|
|decrypt|false|false|decrypted|encrypted|

So for a `watch-only` wallet it doesn't matter if "attended" or not, there will be no `ssk` in memory.

> And, more to the point, what does it mean for an "unattended" wallet to ask for passwords versus not?

An "unattended" wallet will only *need* a password for few things, e.g. to open the wallet, then the keys are stored in memory unencrypted, so no more password is needed for sending/receiving XMR, but you *can* call `verifyPassword()` as protection for some actions.

src/wallet/api/wallet2_api.h
vtnerd: The name `init` makes it seem mandatory, whereas `setDaemon` seems optional.

Is it still basically required that `init` be called? Can `setDaemon` be omitted, and only invoked if more advanced options are required?

SNeedlewoods: Yes, `init()` is mandatory and `setDaemon()` is optional with more advanced options.

In #10233 for the initialization of a wallet object in simplewallet I called both methods [src](https://github.com/SNeedlewoods/seraphis_wallet/blob/81916e7c8de528e91757b51df22a4ae5ea414ea3/src/simplewallet/simplewallet.cpp#L5016-L5034) (there the settings come from command line flags).

`setDaemon()` was also meant to be used if the wallet is already initialized, e.g. the wallet-rpc `set_daemon` uses all the options [doc](https://docs.getmonero.org/rpc-library/wallet-rpc/#set_daemon).

SNeedlewoods:
Rebased, now first three commits of this PR contain #9464 and the fourth commit contains all the code for this PR.

PR #10233, opened 2025-11-24
replace wallet2 with Wallet API in simplewallet

SNeedlewoods:
For some context, this change was proposed a while ago by @j-berman in Step 2 [here](https://github.com/seraphis-migration/wallet3/issues/64).

This PR is based on https://github.com/monero-project/monero/pull/10232 which is based on https://github.com/monero-project/monero/pull/9464

## Possible improvements:

Instead of implementing every idea, which would delay this PR even more, I added this list of things, so we can decide together what should be
a) added to this PR
b) a separate PR
c) not worth to get PRed

- Add option to change the description of an address book entry ([context](https://libera.monerologs.net/monero-dev/20250917#c589017-c589019)), so the new usage message would look something like this:
    ```
    address_book [(add (<address>|<integrated address>) [<description possibly with whitespaces>])|(delete <index>)|(change <address> <description>)]
    ```
- `get_tx_proof`, `check_tx_proof`, `get_spend_proof`, `check_spend_proof`, `get_reserve_proof`, `check_reserve_proof` all fail if <message> is more than one word [src](https://github.com/monero-project/monero/blob/bba6aa518ba4533859766672cf996f6427b8f1a4/src/simplewallet/simplewallet.cpp#L222-L227). We could read them like we do for `<description>` in `address_book` ([src](https://github.com/monero-project/monero/blob/bba6aa518ba4533859766672cf996f6427b8f1a4/src/simplewallet/simplewallet.cpp#L9659-L9664)).

## Behavior changes:

Some behavior changes or new/removed features I could not resist to add to this PR.

- In Monero Tech Meeting [#140](https://github.com/monero-project/meta/issues/1277) we discussed how to show the `amount` & `fee` in the confirmation message when creating a transaction. Then I changed it from "before" to "now". After further consideration I came up with the "new proposal", which I will implement if others agree on it.
  (The `simple_wallet::transfer_main()` method captured the `change` amount [here](https://github.com/monero-project/monero/blob/8e9ab9677f90492bca3c7555a246f2a8677bd570/src/simplewallet/simplewallet.cpp#L6692C18-L6699) but never used/printed it. Imo it's nice to directly see the change you'll receive, so I added it to the confirmation prompt.)
  - `transfer`
    - before:
      ```
      Sending <amount>.  The transaction fee is <fee>
      ```
    - now (`total` = `amount` + `fee`):
      ```
      Sending <amount>.  The transaction fee is <fee>. The overall total is <total>. The change is <change>.
      ```
    - new proposal (`total` = `amount` + `fee` + `change`):
      ```
      Sending overall total <total> = <amount_sent> (amount sent) + <fee> (fee) + <change> (change).
      ```
  - `sweep_*`
    - before:
      ```
      Sweeping <total> for a total fee of <fee>.  Is this okay?  (Y/Yes/N/No):
      ```
    - now (`total` = `amount` + `fee`):
      ```
      Sweeping <amount> for a total fee of <fee>, overall total <total>.  Is this okay?  (Y/Yes/N/No):
      ```
    - new proposal (`total` = `amount` + `fee`):
      ```
      Sweeping overall total <total> = <amount_sent> (amount sent) + <fee> (fee).
      ```

- DataHoarder [reported some issues](https://libera.monerologs.net/monero-dev/20250921#c591956-c591963) with `check_tx_proof`
    TLDR:
        1. can't read a proof directly from string, have to use signature file
        2. signature file content may not end with new line or white spaces, or else verification fails
  With this PR both issues are solved.

- add: Options to set daemon login (username:password) and proxy in `set_daemon` command

- remove: The `payment_id` argument is already not working for `transfer` command (see below) and it's marked "obsolete" for other commands [here](https://github.com/monero-project/monero/blob/bba6aa518ba4533859766672cf996f6427b8f1a4/src/simplewallet/simplewallet.cpp#L200-L205), so I removed the argument from all transfer/sweep commands and from the USAGE messages, and added error messages in case someone still passes a valid payment_id
  - before:
    ```
    [wallet xxxxxx]: transfer <address> <amount> <short_payment_id>
    Error: Invalid last argument: <short_payment_id>
    [wallet xxxxxx]: transfer <address> <amount> <long_payment_id>
    Error: Error: Long payment IDs are obsolete.
    Error: Long payment IDs were not encrypted on the blockchain and would harm your privacy.
    Error: If the party you're sending to still requires a long payment ID, please notify them.
    ```
  - now:
    ```
    [wallet xxxxxx]: transfer <address> <amount> <short_payment_id>
    Error: Short payment IDs are supposed to be used with integrated addresses.
    Error: If you want to make a payment with a certain payment_id, the receiver has to generate an address with `integrated_address <payment_id>`
    ```

- add: `freeze` & `thaw` by enote public key (match [this usage message](https://github.com/monero-project/monero/blob/bba6aa518ba4533859766672cf996f6427b8f1a4/src/simplewallet/simplewallet.cpp#L2160))

- `transfer <address> <amount>` from `watch-only` wallet writes tx to file `unsigned_monero_tx`.  If that file already exists
  - before: overwrite file
    ```
    Unsigned transaction(s) successfully written to file: unsigned_monero_tx
    ```
  - now: warn and don't overwrite 
    ```
    Error: Attempting to save transaction to file, but specified file(s) exist. Exiting to not risk overwriting. File:unsigned_monero_tx
    ```
 
## Bugs

- `refresh_progress_reporter_t` prints `Height x / y` when it shouldn't, e.g. in a confirmation prompt
  - Edit: the problem in the confirmation prompt is fixed, but there are other places where it messes up the output, because it doesn't print on a new line 
- Edit: these should be fixed now: `show_transfers` does not show:
    - outgoing "pending" txs
    - incoming "pool" txs

## Testers welcome

Help to make sure everything works the same as before (or better).
Special attention to hardware devices. I don't have access to any hw wallet, so they're completely untested.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -35,6 +35,7 @@
  */
 
 // use boost bind placeholders for now
+#include <memory>
 #define BOOST_BIND_GLOBAL_PLACEHOLDERS 1
 #include <boost/bind.hpp>
 
@@ -71,12 +72,17 @@
 #include "common/json_util.h"
 #include "ringct/rctSigs.h"
 #include "multisig/multisig.h"
+#include "wallet/api/enote_details.h"
 #include "wallet/wallet_args.h"
+#include "wallet/fee_algorithm.h"
 #include "wallet/fee_priority.h"
 #include "version.h"
 #include <stdexcept>
 #include "wallet/message_store.h"
+#include "wallet/wallet_errors.h"
+#include "net/parse.h"
 #include "QrCode.hpp"
+#include "string_tools.h"
 
 #ifdef WIN32
 #include <boost/locale.hpp>
@@ -91,6 +97,7 @@
 using namespace std;
 using namespace epee;
 using namespace cryptonote;
+using namespace Monero;
 using boost::lexical_cast;
 namespace po = boost::program_options;
 typedef cryptonote::simple_wallet sw;
@@ -107,7 +114,7 @@ using tools::fee_priority;
   bool auto_refresh_enabled = m_auto_refresh_enabled.load(std::memory_order_relaxed); \
   m_auto_refresh_enabled.store(false, std::memory_order_relaxed); \
   /* stop any background refresh and other processes, and take over */ \
-  m_wallet->stop(); \
+  m_wallet_impl->stop(); \
   boost::unique_lock<boost::mutex> lock(m_idle_mutex); \
   m_idle_cond.notify_all(); \
   const epee::scope_guard scope_exit_handler([&](){ \
@@ -119,8 +126,9 @@ using tools::fee_priority;
 #define SCOPED_WALLET_UNLOCK_ON_BAD_PASSWORD(code) \
   LOCK_IDLE_SCOPE(); \
   boost::optional<tools::password_container> pwd_container = boost::none; \
-  if (m_wallet->ask_password() && !(pwd_container = get_and_verify_password())) { code; } \
-  tools::wallet_keys_unlocker unlocker(*m_wallet, pwd_container ? &pwd_container->password() : nullptr);
+  if (m_wallet_impl->getAskPasswordType() != Wallet::AskPasswordType::AskPassword_Never && !(pwd_container = get_and_verify_password())) { code; } \
+  std::unique_ptr<Monero::WalletKeysDecryptGuard> unlocker = \
+          m_wallet_impl->createKeysDecryptGuard(std::string_view(pwd_container ? pwd_container->password().data() : "", pwd_container ? pwd_container->password().size() : 0))
 
 #define SCOPED_WALLET_UNLOCK() SCOPED_WALLET_UNLOCK_ON_BAD_PASSWORD(return true;)
 
@@ -134,6 +142,13 @@ using tools::fee_priority;
     return true; \
   } while(0)
 
+#define SHORT_PAYMENT_ID_WITHOUT_INTEGRATED_ADDRESS_SUPPORT_CHECK() \
+  do { \
+    fail_msg_writer() << tr("Short payment IDs are supposed to be used with integrated addresses."); \
+    fail_msg_writer() << tr("If you want to make a payment with a certain payment_id, the receiver has to generate an address with `integrated_address <payment_id>`"); \
+    return true; \
+  } while(0)
+
 #define REFRESH_PERIOD 90 // seconds
 
 #define MAX_MNEW_ADDRESSES 65536
@@ -141,7 +156,7 @@ using tools::fee_priority;
 #define CHECK_MULTISIG_ENABLED() \
   do \
   { \
-    if (!m_wallet->is_multisig_enabled()) \
+    if (!m_wallet_impl->getEnableMultisig()) \
     { \
       fail_msg_writer() << tr("Multisig is disabled."); \
       fail_msg_writer() << tr("Multisig is an experimental feature and may have bugs. Things that could go wrong include: funds sent to a multisig wallet can't be spent at all, can only be spent with the participation of a malicious group member, or can be stolen by a malicious group member."); \
@@ -151,16 +166,40 @@ using tools::fee_priority;
     } \
   } while(0)
 
+#define CHECK_MULTISIG_READY() \
+  const MultisigState ms_status{m_wallet_impl->multisig()}; \
+  do \
+  { \
+    if (m_wallet_impl->getDeviceState().key_on_device) \
+    { \
+      fail_msg_writer() << tr("command not supported by HW wallet"); \
+      return false; \
+    } \
+    if (!ms_status.isMultisig) \
+    { \
+      fail_msg_writer() << tr("This wallet is not multisig"); \
+      return false; \
+    } \
+    if (!ms_status.isReady) \
+    { \
+      fail_msg_writer() << tr("This multisig wallet is not yet finalized"); \
+      return false; \
+    } \
+  } while(0)
+
 #define CHECK_IF_BACKGROUND_SYNCING(msg) \
   do \
   { \
-    if (m_wallet->is_background_wallet() || m_wallet->is_background_syncing()) \
+    if (m_wallet_impl->isBackgroundWallet() || m_wallet_impl->isBackgroundSyncing()) \
     { \
-      std::string type = m_wallet->is_background_wallet() ? "background wallet" : "background syncing wallet"; \
+      std::string type = m_wallet_impl->isBackgroundWallet() ? "background wallet" : "background syncing wallet"; \
       fail_msg_writer() << boost::format(tr("%s %s")) % type % msg; \
       return false; \
     } \
   } while (0)
+
+// Copied from wallet2, used by save_to_file() and load_from_file()
+static const std::string ASCII_OUTPUT_MAGIC = "MoneroAsciiDataV1";
 
 namespace
 {
@@ -184,22 +223,47 @@ namespace
   const command_line::arg_descriptor<std::string> arg_restore_date = {"restore-date", sw::tr("Restore from estimated blockchain height on specified date"), ""};
   const command_line::arg_descriptor<bool> arg_do_not_relay = {"do-not-relay", sw::tr("The newly created transaction will not be relayed to the monero network"), false};
   const command_line::arg_descriptor<bool> arg_create_address_file = {"create-address-file", sw::tr("Create an address file for new wallets"), false};
-  const command_line::arg_descriptor<std::string> arg_subaddress_lookahead = {"subaddress-lookahead", tools::wallet2::tr("Set subaddress lookahead sizes to <major>:<minor>"), ""};
+  const command_line::arg_descriptor<std::string> arg_subaddress_lookahead = {"subaddress-lookahead", sw::tr("Set subaddress lookahead sizes to <major>:<minor>"), ""};
   const command_line::arg_descriptor<bool> arg_use_english_language_names = {"use-english-language-names", sw::tr("Display English language names"), false};
+  const command_line::arg_descriptor<std::string> arg_daemon_address = {"daemon-address", sw::tr("Use daemon instance at <host>:<port>"), ""};
+  const command_line::arg_descriptor<std::string> arg_daemon_host = {"daemon-host", sw::tr("Use daemon instance at host <arg> instead of localhost"), ""};
+  const command_line::arg_descriptor<std::string> arg_proxy = {"proxy", sw::tr("[<ip>:]<port> socks proxy to use for daemon connections"), {}, true};
+  const command_line::arg_descriptor<bool> arg_trusted_daemon = {"trusted-daemon", sw::tr("Enable commands which rely on a trusted daemon"), false};
+  const command_line::arg_descriptor<bool> arg_untrusted_daemon = {"untrusted-daemon", sw::tr("Disable commands which rely on a trusted daemon"), false};
+  const command_line::arg_descriptor<std::string> arg_password = {"password", sw::tr("Wallet password (escape/quote as needed)"), "", true};
+  const command_line::arg_descriptor<std::string> arg_password_file = wallet_args::arg_password_file();
+  const command_line::arg_descriptor<int> arg_daemon_port = {"daemon-port", sw::tr("Use daemon instance at port <arg> instead of 18081"), 0};
+  const command_line::arg_descriptor<std::string> arg_daemon_login = {"daemon-login", sw::tr("Specify username[:password] for daemon RPC client"), "", true};
+  const command_line::arg_descriptor<std::string> arg_daemon_ssl = {"daemon-ssl", sw::tr("Enable SSL on daemon RPC connections: enabled|disabled|autodetect"), "autodetect"};
+  const command_line::arg_descriptor<std::string> arg_daemon_ssl_private_key = {"daemon-ssl-private-key", sw::tr("Path to a PEM format private key"), ""};
+  const command_line::arg_descriptor<std::string> arg_daemon_ssl_certificate = {"daemon-ssl-certificate", sw::tr("Path to a PEM format certificate"), ""};
+  const command_line::arg_descriptor<std::string> arg_daemon_ssl_ca_certificates = {"daemon-ssl-ca-certificates", sw::tr("Path to file containing concatenated PEM format certificate(s) to replace system CA(s).")};
+  const command_line::arg_descriptor<std::vector<std::string>> arg_daemon_ssl_allowed_fingerprints = {"daemon-ssl-allowed-fingerprints", sw::tr("List of valid fingerprints of allowed RPC servers")};
+  const command_line::arg_descriptor<bool> arg_daemon_ssl_allow_any_cert = {"daemon-ssl-allow-any-cert", sw::tr("Allow any SSL certificate from the daemon"), false};
+  const command_line::arg_descriptor<bool> arg_daemon_ssl_allow_chained = {"daemon-ssl-allow-chained", sw::tr("Allow user (via --daemon-ssl-ca-certificates) chain certificates"), false};
+  const command_line::arg_descriptor<bool> arg_allow_mismatched_daemon_version = {"allow-mismatched-daemon-version", sw::tr("Allow communicating with a daemon that uses a different version"), false};
+  const command_line::arg_descriptor<bool> arg_testnet = {"testnet", sw::tr("For testnet. Daemon must also be launched with --testnet flag"), false};
+  const command_line::arg_descriptor<bool> arg_stagenet = {"stagenet", sw::tr("For stagenet. Daemon must also be launched with --stagenet flag"), false};
+  const command_line::arg_descriptor<uint64_t> arg_kdf_rounds = {"kdf-rounds", sw::tr("Number of rounds for the key derivation function"), 1};
+  const command_line::arg_descriptor<std::string> arg_hw_device = {"hw-device", sw::tr("HW device to use"), ""};
+  const command_line::arg_descriptor<std::string> arg_hw_device_derivation_path = {"hw-device-deriv-path", sw::tr("HW device wallet derivation path (e.g., SLIP-10)"), ""};
+  const command_line::arg_descriptor<std::string> arg_tx_notify = { "tx-notify" , sw::tr("Run a program for each new incoming transaction, '%s' will be replaced by the transaction hash") , "" };
+  const command_line::arg_descriptor<bool> arg_offline = {"offline", sw::tr("Do not connect to a daemon, nor use DNS"), false};
+  const command_line::arg_descriptor<std::string> arg_extra_entropy = {"extra-entropy", sw::tr("File containing extra entropy to initialize the PRNG (any data, aim for 256 bits of entropy to be useful, which typically means more than 256 bits of data)")};
 
   const command_line::arg_descriptor< std::vector<std::string> > arg_command = {"command", ""};
 
   const char* USAGE_START_MINING("start_mining [<number_of_threads>] [bg_mining] [ignore_battery]");
-  const char* USAGE_SET_DAEMON("set_daemon <host>[:<port>] [trusted|untrusted|this-is-probably-a-spy-node]");
+  const char* USAGE_SET_DAEMON("set_daemon <host>[:<port>] [trusted|untrusted|this-is-probably-a-spy-node] [login=<user>[:<password>]] [proxy=[<ip>:]<port>]");
   const char* USAGE_SHOW_BALANCE("balance [detail]");
   const char* USAGE_INCOMING_TRANSFERS("incoming_transfers [available|unavailable] [verbose] [uses] [index=<N1>[,<N2>[,...]]]");
   const char* USAGE_PAYMENTS("payments <PID_1> [<PID_2> ... <PID_N>]");
   const char* USAGE_PAYMENT_ID("payment_id");
-  const char* USAGE_TRANSFER("transfer [index=<N1>[,<N2>,...]] [<priority>] [<ring_size>] (<URI> | <address> <amount>) [subtractfeefrom=<D0>[,<D1>,all,...]] [<payment_id>]");
-  const char* USAGE_SWEEP_ALL("sweep_all [index=<N1>[,<N2>,...] | index=all] [<priority>] [<ring_size>] [outputs=<N>] <address> [<payment_id (obsolete)>]");
-  const char* USAGE_SWEEP_ACCOUNT("sweep_account <account> [index=<N1>[,<N2>,...] | index=all] [<priority>] [<ring_size>] [outputs=<N>] <address> [<payment_id (obsolete)>]");
-  const char* USAGE_SWEEP_BELOW("sweep_below <amount_threshold> [index=<N1>[,<N2>,...]] [<priority>] [<ring_size>] <address> [<payment_id (obsolete)>]");
-  const char* USAGE_SWEEP_SINGLE("sweep_single [<priority>] [<ring_size>] [outputs=<N>] <key_image> <address> [<payment_id (obsolete)>]");
+  const char* USAGE_TRANSFER("transfer [index=<N1>[,<N2>,...]] [<priority>] [<ring_size>] (<URI> | <address> <amount>) [subtractfeefrom=<D0>[,<D1>,all,...]]");
+  const char* USAGE_SWEEP_ALL("sweep_all [index=<N1>[,<N2>,...] | index=all] [<priority>] [<ring_size>] [outputs=<N>] <address>");
+  const char* USAGE_SWEEP_ACCOUNT("sweep_account <account> [index=<N1>[,<N2>,...] | index=all] [<priority>] [<ring_size>] [outputs=<N>] <address>");
+  const char* USAGE_SWEEP_BELOW("sweep_below <amount_threshold> [index=<N1>[,<N2>,...]] [<priority>] [<ring_size>] [outputs=<N>] <address>");
+  const char* USAGE_SWEEP_SINGLE("sweep_single [<priority>] [<ring_size>] [outputs=<N>] <key_image> <address>");
   const char* USAGE_DONATE("donate [index=<N1>[,<N2>,...]] [<priority>] [<ring_size>] <amount>");
   const char* USAGE_SIGN_TRANSFER("sign_transfer [export_raw] [<filename>]");
   const char* USAGE_SET_LOG("set_log <level>|{+,-,}<categories>");
@@ -218,7 +282,7 @@ namespace
   const char* USAGE_SET_TX_KEY("set_tx_key <txid> <tx_key> [<subaddress>]");
   const char* USAGE_CHECK_TX_KEY("check_tx_key <txid> <txkey> <address>");
   const char* USAGE_GET_TX_PROOF("get_tx_proof <txid> <address> [<message>]");
-  const char* USAGE_CHECK_TX_PROOF("check_tx_proof <txid> <address> <signature_file> [<message>]");
+  const char* USAGE_CHECK_TX_PROOF("check_tx_proof <txid> <address> ( <signature_file> | <signature> ) [<message>]");
   const char* USAGE_GET_SPEND_PROOF("get_spend_proof <txid> [<message>]");
   const char* USAGE_CHECK_SPEND_PROOF("check_spend_proof <txid> <signature_file> [<message>]");
   const char* USAGE_GET_RESERVE_PROOF("get_reserve_proof (all|<amount>) [<message>]");
@@ -246,33 +310,11 @@ namespace
   const char* USAGE_SIGN_MULTISIG("sign_multisig <filename>");
   const char* USAGE_SUBMIT_MULTISIG("submit_multisig <filename>");
   const char* USAGE_EXPORT_RAW_MULTISIG_TX("export_raw_multisig_tx <filename>");
-  const char* USAGE_MMS("mms [<subcommand> [<subcommand_parameters>]]");
-  const char* USAGE_MMS_INIT("mms init <required_signers>/<authorized_signers> <own_label> <own_transport_address>");
-  const char* USAGE_MMS_INFO("mms info");
-  const char* USAGE_MMS_SIGNER("mms signer [<number> <label> [<transport_address> [<monero_address>]]]");
-  const char* USAGE_MMS_LIST("mms list");
-  const char* USAGE_MMS_NEXT("mms next [sync]");
-  const char* USAGE_MMS_SYNC("mms sync");
-  const char* USAGE_MMS_TRANSFER("mms transfer <transfer_command_arguments>");
-  const char* USAGE_MMS_DELETE("mms delete (<message_id> | all)");
-  const char* USAGE_MMS_SEND("mms send [<message_id>]");
-  const char* USAGE_MMS_RECEIVE("mms receive");
-  const char* USAGE_MMS_EXPORT("mms export <message_id>");
-  const char* USAGE_MMS_NOTE("mms note [<label> <text>]");
-  const char* USAGE_MMS_SHOW("mms show <message_id>");
-  const char* USAGE_MMS_SET("mms set <option_name> [<option_value>]");
-  const char* USAGE_MMS_SEND_SIGNER_CONFIG("mms send_signer_config");
-  const char* USAGE_MMS_START_AUTO_CONFIG("mms start_auto_config [<label> <label> ...]");
-  const char* USAGE_MMS_CONFIG_CHECKSUM("mms config_checksum");
-  const char* USAGE_MMS_STOP_AUTO_CONFIG("mms stop_auto_config");
-  const char* USAGE_MMS_AUTO_CONFIG("mms auto_config <auto_config_token>");
   const char* USAGE_PRINT_RING("print_ring <key_image> | <txid>");
   const char* USAGE_SET_RING("set_ring <filename> | ( <key_image> absolute|relative <index> [<index>...] )");
-  const char* USAGE_UNSET_RING("unset_ring <txid> | ( <key_image> [<key_image>...] )");
-  const char* USAGE_SAVE_KNOWN_RINGS("save_known_rings");
-  const char* USAGE_FREEZE("freeze <key_image>");
-  const char* USAGE_THAW("thaw <key_image>");
-  const char* USAGE_FROZEN("frozen <key_image>");
+  const char* USAGE_FREEZE("freeze <key_image> | <enote_pubkey>");
+  const char* USAGE_THAW("thaw <key_image> | <enote_pubkey>");
+  const char* USAGE_FROZEN("frozen [<key_image> | <enote_pubkey>]");
   const char* USAGE_LOCK("lock");
   const char* USAGE_NET_STATS("net_stats");
   const char* USAGE_PUBLIC_NODES("public_nodes");
@@ -338,27 +380,6 @@ namespace
     return password_prompter(verify ? sw::tr("Enter a custom password for the background sync cache") : sw::tr("Background sync cache password"), verify);
   }
 
-  inline std::string interpret_rpc_response(bool ok, const std::string& status)
-  {
-    std::string err;
-    if (ok)
-    {
-      if (status == CORE_RPC_STATUS_BUSY)
-      {
-        err = sw::tr("daemon is busy. Please try again later.");
-      }
-      else if (status != CORE_RPC_STATUS_OK)
-      {
-        err = status;
-      }
-    }
-    else
-    {
-      err = sw::tr("possibly lost connection to daemon");
-    }
-    return err;
-  }
-
   tools::scoped_message_writer success_msg_writer(bool color = false)
   {
     return tools::scoped_message_writer(color ? console_color_green : console_color_default, false, std::string(), el::Level::Info);
@@ -421,17 +442,17 @@ namespace
   const struct
   {
     const char *name;
-    tools::wallet2::RefreshType refresh_type;
+    Wallet::RefreshType refresh_type;
   } refresh_type_names[] =
   {
-    { "full", tools::wallet2::RefreshFull },
-    { "optimize-coinbase", tools::wallet2::RefreshOptimizeCoinbase },
-    { "optimized-coinbase", tools::wallet2::RefreshOptimizeCoinbase },
-    { "no-coinbase", tools::wallet2::RefreshNoCoinbase },
-    { "default", tools::wallet2::RefreshDefault },
+    { "full", Wallet::RefreshType::Refresh_Full },
+    { "optimize-coinbase", Wallet::RefreshType::Refresh_OptimizeCoinbase },
+    { "optimized-coinbase", Wallet::RefreshType::Refresh_OptimizeCoinbase },
+    { "no-coinbase", Wallet::RefreshType::Refresh_NoCoinbase },
+    { "default", Wallet::RefreshType::Refresh_Default },
   };
 
-  bool parse_refresh_type(const std::string &s, tools::wallet2::RefreshType &refresh_type)
+  bool parse_refresh_type(const std::string &s, Wallet::RefreshType &refresh_type)
   {
     for (size_t n = 0; n < sizeof(refresh_type_names) / sizeof(refresh_type_names[0]); ++n)
     {
@@ -445,7 +466,7 @@ namespace
     return false;
   }
 
-  std::string get_refresh_type_name(tools::wallet2::RefreshType type)
+  std::string get_refresh_type_name(Wallet::RefreshType type)
   {
     for (size_t n = 0; n < sizeof(refresh_type_names) / sizeof(refresh_type_names[0]); ++n)
     {
@@ -458,15 +479,15 @@ namespace
   const struct
   {
     const char *name;
-    tools::wallet2::BackgroundSyncType background_sync_type;
+    Wallet::BackgroundSyncType background_sync_type;
   } background_sync_type_names[] =
   {
-    { "off", tools::wallet2::BackgroundSyncOff },
-    { "reuse-wallet-password", tools::wallet2::BackgroundSyncReusePassword },
-    { "custom-background-password", tools::wallet2::BackgroundSyncCustomPassword },
+    { "off", Wallet::BackgroundSync_Off },
+    { "reuse-wallet-password", Wallet::BackgroundSync_ReusePassword },
+    { "custom-background-password", Wallet::BackgroundSync_CustomPassword },
   };
 
-  bool parse_background_sync_type(const std::string &s, tools::wallet2::BackgroundSyncType &background_sync_type)
+  bool parse_background_sync_type(const std::string &s, Wallet::BackgroundSyncType &background_sync_type)
   {
     for (size_t n = 0; n < sizeof(background_sync_type_names) / sizeof(background_sync_type_names[0]); ++n)
     {
@@ -480,7 +501,7 @@ namespace
     return false;
   }
 
-  std::string get_background_sync_type_name(tools::wallet2::BackgroundSyncType type)
+  std::string get_background_sync_type_name(Wallet::BackgroundSyncType type)
   {
     for (size_t n = 0; n < sizeof(background_sync_type_names) / sizeof(background_sync_type_names[0]); ++n)
     {
@@ -568,7 +589,7 @@ namespace
   bool parse_subtract_fee_from_outputs
   (
     const std::string& arg,
-    tools::wallet2::unique_index_container& subtract_fee_from_outputs,
+    std::set<uint32_t>& subtract_fee_from_outputs,
     bool& subtract_fee_from_all,
     bool& matches
   )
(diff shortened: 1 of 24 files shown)


rbrunner7:
I started to test, but at least for my personal taste I found the two known still existing bugs to distract quite a bit, and to make it difficult to check whether everything runs correctly as long as you can't really trust the `show_transfers` command to give a complete picture.

Do you already have an estimate when you will be able to correct those 2 bugs?

SNeedlewoods:
the known issues with `show_transfers` should be fixed

rbrunner7:
@SNeedlewoods , is it ok for you to keep this now stable for a week or two? This is a large and complicated review, and I plan to take notes using line numbers e.g. for things that I don't immediately understand (for lack of a better idea ...), and if you make changes those number will of course change. Also the commit hash will change which will render any bookmarks I make invalid.

src/simplewallet/simplewallet.h
```diff
@@ -43,6 +43,7 @@
 #include "cryptonote_basic/account.h"
 #include "cryptonote_basic/cryptonote_basic_impl.h"
 #include "wallet/wallet2.h"
```
rbrunner7: You can at long last remove this include; I checked, it compiles without it :)

src/simplewallet/simplewallet.h
```diff
@@ -168,14 +171,13 @@ namespace cryptonote
...
     bool show_blockchain_height(const std::vector<std::string> &args);
-    bool transfer_main(const std::vector<std::string> &args, bool called_by_mms);
+    bool transfer_main(const std::vector<std::string> &args);
     bool transfer(const std::vector<std::string> &args);
     bool sweep_main(uint32_t account, uint64_t below, const std::vector<std::string> &args);
     bool sweep_all(const std::vector<std::string> &args);
     bool sweep_account(const std::vector<std::string> &args);
     bool sweep_below(const std::vector<std::string> &args);
     bool sweep_single(const std::vector<std::string> &args);
-    bool sweep_unmixable(const std::vector<std::string> &args);
```
rbrunner7, line 178: Just a question, can you refresh my memory: Does "sweeping unmixables" already today not make sense anymore, or is this something geared towards the FCMP++ hardfork where it will not make sense anymore, *after* that hardfork?

src/simplewallet/simplewallet.cpp
```diff
@@ -119,8 +125,9 @@ using tools::fee_priority;
 #define SCOPED_WALLET_UNLOCK_ON_BAD_PASSWORD(code) \
   LOCK_IDLE_SCOPE(); \
   boost::optional<tools::password_container> pwd_container = boost::none; \
-  if (m_wallet->ask_password() && !(pwd_container = get_and_verify_password())) { code; } \
-  tools::wallet_keys_unlocker unlocker(*m_wallet, pwd_container ? &pwd_container->password() : nullptr);
+  if (m_wallet_impl->getAskPasswordType() && !(pwd_container = get_and_verify_password())) { code; } \
```
rbrunner7: A small stylistic thing: I would compare against `AskPassword_Never` instead of relying on the fact that this enum value is 0.

src/simplewallet/simplewallet.cpp
```diff
@@ -612,129 +606,236 @@ namespace
...
+    epee::net_utils::ssl_options_t ssl_options = epee::net_utils::ssl_support_t::e_ssl_support_enabled;
+    if (ssl_allow_any_cert)
+      ssl_options.verification = epee::net_utils::ssl_verification_t::none;
+    else if (!ssl_ca_file_path.empty() || !ssl_allowed_fingerprints_str.empty())
     {
-      auto writer = fail_msg_writer();
-      writer << sw::tr("not enough outputs for specified ring size") << " = " << (e.mixin_count() + 1) << ":";
-      for (std::pair<uint64_t, uint64_t> outs_for_amount : e.scanty_outs())
+      std::vector<std::vector<uint8_t>> ssl_allowed_fingerprints{ ssl_allowed_fingerprints_str.size() };
+      std::transform(ssl_allowed_fingerprints_str.begin(), ssl_allowed_fingerprints_str.end(), ssl_allowed_fingerprints.begin(), epee::from_hex_locale::to_vector);
```
rbrunner7, line 728: `epee::from_hex_locale::to_vector` can throw exceptions if something is not ok with the parameters. Can this be a problem?

src/simplewallet/simplewallet.cpp
```diff
@@ -612,129 +606,236 @@ namespace
...
+      fail_msg_writer() << (boost::format(tr("ring size %u is too large, maximum is %u")) % (fake_outs_count+1) % (adjusted_fake_outs_count+1)).str();
+      return false;
     }
+    return true;
+  }
 
-    if (warn_of_possible_attack)
-      fail_msg_writer() << sw::tr("There was an error, which could mean the node may be trying to get you to retry creating a transaction, and zero in on which outputs you own. Or it could be a bona fide error. It may be prudent to disconnect from this node, and not try to send a transaction immediately. Alternatively, connect to another node so the original node cannot correlate information.");
-}
+  bool is_short_payment_id(std::string &payment_id)
```
rbrunner7: Not sure it really makes any difference, but technically you could use `const` for this and similar arguments.

src/simplewallet/simplewallet.cpp
```diff
@@ -1035,18 +1158,22 @@ bool simple_wallet::print_fee_info(const std::vector<std::string> &args/* = std:
   {
     if (priority == fee_priority::Default)
       continue;
-    uint64_t mult = m_wallet->get_fee_multiplier(priority);
+    uint64_t mult = m_wallet_impl->getFeeMultiplier(tools::fee_priority_utilities::as_integral(priority), /* fee_algorithm */ -1);
```
rbrunner7: Would it be easy to use the enum constant `Unset` instead of -1?

(Did this code come in through a merge / rebase and so isn't yours?)

src/simplewallet/simplewallet.cpp
```diff
@@ -1362,26 +1480,28 @@ bool simple_wallet::export_multisig_main(const std::vector<std::string> &args, b
...
+  import_multisig_main(args);
   return true;
 }
 
-bool simple_wallet::import_multisig_main(const std::vector<std::string> &args, bool called_by_mms)
+bool simple_wallet::import_multisig_main(const std::vector<std::string> &args)
 {
   CHECK_MULTISIG_ENABLED();
-  const multisig::multisig_account_status ms_status{m_wallet->get_multisig_status()};
+  const MultisigState ms_status{m_wallet_impl->multisig()};
```
rbrunner7: I wonder whether it would be worth it, and easy enough, to put all this repeated getting multisig state and then checking whether A) is key on device, B) is multisig, C) is ready into its own method ...

src/simplewallet/simplewallet.cpp
```diff
@@ -1395,50 +1515,42 @@ bool simple_wallet::import_multisig_main(const std::vector<std::string> &args, b
...
+    size_t n_outputs = m_wallet_impl->importMultisigImages(info);
+
+    m_wallet_impl->statusWithErrorString(error_code, error_msg);
+    if (error_code)
+    {
+      fail_msg_writer() << tr("Failed to import multisig info: ") << error_msg;
+      return false;
+    }
+
     // Clear line "Height xxx of xxx"
```
rbrunner7, line 1517: What the heck is this, and more importantly, is it still revelant after switching to `importMultisigImages`?

src/simplewallet/simplewallet.cpp
```diff
@@ -1450,35 +1562,29 @@ bool simple_wallet::import_multisig_main(const std::vector<std::string> &args, b
   return true;
 }
 
-bool simple_wallet::accept_loaded_tx(const tools::wallet2::multisig_tx_set &txs)
-{
-  std::string extra_message;
-  return accept_loaded_tx([&txs](){return txs.m_ptx.size();}, [&txs](size_t n)->const tools::wallet2::tx_construction_data&{return txs.m_ptx[n].construction_data;}, extra_message);
-}
-
 bool simple_wallet::sign_multisig(const std::vector<std::string> &args)
```
rbrunner7, line 1538: If I am not mistaken, now with the MMS gone there is no need anymore for *two* methods `sign_multisig` and `sign_multisig_main`?

src/simplewallet/simplewallet.cpp
```diff
@@ -1492,61 +1598,43 @@ bool simple_wallet::sign_multisig_main(const std::vector<std::string> &args, boo
...
-  std::vector<crypto::hash> txids;
+  std::vector<std::string> txids;
   uint32_t signers = 0;
-  try
+  int error_code;
+  std::string error_msg;
+  std::string multisig_tx_data;
+  std::string ciphertext;
+
+  if (!m_wallet_impl->loadFromFile(args[0], multisig_tx_data))
```
rbrunner7: Little nitpick: We already have `args[0]` ready in `filename` and could use that.

src/simplewallet/simplewallet.cpp
```diff
@@ -1492,61 +1598,43 @@ bool simple_wallet::sign_multisig_main(const std::vector<std::string> &args, boo
   SCOPED_WALLET_UNLOCK_ON_BAD_PASSWORD(return false;);
 
   std::string filename = args[0];
-  std::vector<crypto::hash> txids;
+  std::vector<std::string> txids;
   uint32_t signers = 0;
```
rbrunner7, line 1552: `signers` will be needed for an info message below, but it seems your code, unlike the old code, does not set it any longer?

src/simplewallet/simplewallet.cpp
```diff
@@ -1575,26 +1663,26 @@ bool simple_wallet::sign_multisig_main(const std::vector<std::string> &args, boo
 bool simple_wallet::submit_multisig(const std::vector<std::string> &args)
```
rbrunner7, line 1619: Like above, I don't think we need two separate methods `submit_multisig_main` and `submit_multisig` anymore.

src/simplewallet/simplewallet.cpp
```diff
@@ -1744,43 +1833,32 @@ bool simple_wallet::print_ring(const std::vector<std::string> &args)
...
+  std::string key_image_or_tx_id = args[0];
 
-  if (!epee::string_tools::hex_to_pod(args[0], key_image))
-  {
-    fail_msg_writer() << tr("Invalid key image");
-    return true;
-  }
-  // this one will always work, they're all 32 byte hex
-  if (!epee::string_tools::hex_to_pod(args[0], txid))
+  if (!epee::string_tools::hex_to_pod(key_image_or_tx_id, key_image)
```
rbrunner7: Looks to me as if there is no need to convert the same string twice, and no need for two variables `key_image` and `txid`, just one dummy one because the result is not really used.

src/simplewallet/simplewallet.cpp
```diff
@@ -1907,6 +1985,7 @@ bool simple_wallet::set_ring(const std::vector<std::string> &args)
     fail_msg_writer() << tr("Invalid key image");
     return true;
   }
+  *key_image_str = args[0][0];
```
rbrunner7: This looks strange, and I don't understand it.

src/simplewallet/simplewallet.cpp
```diff
@@ -2307,27 +2229,6 @@ bool simple_wallet::on_cancelled_command()
   return true;
 }
 
-bool simple_wallet::cold_sign_tx(const std::vector<tools::wallet2::pending_tx>& ptx_vector, tools::wallet2::signed_tx_set &exported_txs, std::vector<cryptonote::address_parse_info> &dsts_info, std::function<bool(const tools::wallet2::signed_tx_set &)> accept_func)
```
rbrunner7, line 2173: The Wallet API seems to implement cold signing, but I don't see any replacement for this old `cold_sign_tx` method here, which was used in the old code, nor any calls to the Wallet API cold signing method `coldSignTx` in general. What do I overlook?

src/simplewallet/simplewallet.cpp
```diff
@@ -2967,19 +2876,19 @@ bool simple_wallet::set_device_name(const std::vector<std::string> &args/* = std
...
+    m_wallet_impl->setDeviceName(args[1]);
+    if (!m_wallet_impl->reconnectDevice()){
+      fail_msg_writer() << tr("Device reconnect failed");
+      int error_code;
+      std::string error_msg;
+      m_wallet_impl->statusWithErrorString(error_code, error_msg);
+      if (error_code)
+          fail_msg_writer() << error_msg;
+    }
+    // QUESTION : Is there a reason we don't rewrite wallet file after setting device name? Or does it make sense to add the following else block?
```
rbrunner7: Regarding this question: Seems to me the device name is a stored option exactly like so many others, so we should immediately write the wallet file after changing it like for all others.

src/simplewallet/simplewallet.cpp
```diff
@@ -3621,38 +3426,18 @@ simple_wallet::simple_wallet()
                            boost::bind(&simple_wallet::on_command, this, &simple_wallet::set_ring, _1),
                            tr(USAGE_SET_RING),
                            tr("Set the ring used for a given key image, so it can be reused in a fork"));
-  m_cmd_binder.set_handler("unset_ring",
```
rbrunner7, line 3482: Just thinking aloud, make of it what you like: When I code, and delete something that was there for a long time, I do not just delete it, but in the same place of the deleted code I add comments about the deleted code, telling something *what* was once here, and/or *why* it was deleted. Here, just as an example, you could leave info what drove the decision to finally delete these commands in question.

src/simplewallet/simplewallet.cpp
```diff
@@ -4758,13 +4538,13 @@ std::string simple_wallet::get_mnemonic_language()
...
   const bool verify = m_wallet_file.empty();
-  auto pwd_container = (m_wallet->is_background_wallet() && m_wallet->background_sync_type() == tools::wallet2::BackgroundSyncCustomPassword)
+  auto pwd_container = (m_wallet_impl->isBackgroundWallet() && m_wallet_impl->getBackgroundSyncType() == Wallet::BackgroundSync_CustomPassword)
     ? background_sync_cache_password_prompter(verify)
     : default_password_prompter(verify);
   if (!pwd_container)
     return boost::none;
 
-  if (!m_wallet->verify_password(pwd_container->password()))
+  if (!m_wallet_impl->verifyPassword(pwd_container->password().data(), pwd_container->password().size(), /* kdf_rounds */ 1))
```
rbrunner7: Not fully sure, but I suspect that a fixed value of 1 for `kdf_rounds` is not correct. If I drill down the `m_wallet->verify_password` function call that this replaces, I see it using `m_kdf_rounds` in the old code.

src/simplewallet/simplewallet.cpp
```diff
@@ -4874,315 +4692,472 @@ boost::optional<epee::wipeable_string> simple_wallet::new_wallet(const boost::pr
...
     {
       boost::unique_lock<boost::mutex> lock(m_idle_mutex);
       m_idle_cond.notify_one();
     }
     m_idle_thread.join();
   }
 
-  bool r = m_wallet->deinit();
-  if (!r)
+  if (!m_wallet_manager->closeWallet(m_wallet_impl.get(), /* store */ true, /* do_delete_pointer */ false))
```
rbrunner7, line 5038: What's the reasoning behind not deleting the pointer?

src/simplewallet/simplewallet.cpp
```diff
@@ -5517,122 +5487,135 @@ bool simple_wallet::set_daemon(const std::vector<std::string>& args)
 //----------------------------------------------------------------------------------------------------
 bool simple_wallet::save_bc(const std::vector<std::string>& args)
```
rbrunner7, line 5490: Might be this could go for a long time already, but that's probably a PR for another day, after carefully checking whether "saving the blockchain" on command, at a particular point in time, and from the wallet no less, is indeed totally unnecessary nowadays.

src/simplewallet/simplewallet.cpp
```diff
@@ -5517,122 +5487,135 @@ bool simple_wallet::set_daemon(const std::vector<std::string>& args)
...
+                                  uint64_t amount,
+                                  const uint64_t burnt,
+                                  const bool is_change /* = false */,
+                                  const bool is_coinbase /* = false */)
 {
   if (m_locked)
     return;
+
+  // Get latest enote details
+  std::unique_ptr<EnoteDetails> ed;
```
rbrunner7: I am a bit confused. Isn't this whole approach of getting the "latest enote" pretty questionable? What if the transaction brings more than 1 enote? And isn't it possible that transactions come in in rapid succession, and we can't be sure about the order that `moneyReceived` will get called for them?

src/simplewallet/simplewallet.cpp
```diff
@@ -5517,122 +5487,135 @@ bool simple_wallet::set_daemon(const std::vector<std::string>& args)
...
-void simple_wallet::on_money_spent(uint64_t height, const crypto::hash &txid, const cryptonote::transaction& in_tx, uint64_t amount, const cryptonote::transaction& spend_tx, const cryptonote::subaddress_index& subaddr_index)
+void simple_wallet::moneySpent(const std::string &txid,
+                               uint64_t amount,
+                               std::pair<uint32_t, uint32_t> subaddr_index /* = {} */)
 {
   if (m_locked)
     return;
+
+  // Get latest enote details
+  std::unique_ptr<EnoteDetails> ed;
```
rbrunner7: Same question like above at `moneyReceived` about the "latest enote" approach applies.

src/simplewallet/simplewallet.cpp
```diff
@@ -5517,122 +5487,135 @@ bool simple_wallet::set_daemon(const std::vector<std::string>& args)
...
     tr("spent ") << print_money(amount) << ", " <<
-    tr("idx ") << subaddr_index;
+    tr("idx ") << subaddr_index.first << "/" << subaddr_index.second;
   if (m_auto_refresh_refreshing)
     m_cmd_binder.print_prompt();
   else
-    m_refresh_progress_reporter.update(height, true);
-}
-//----------------------------------------------------------------------------------------------------
-void simple_wallet::on_skip_transaction(uint64_t height, const crypto::hash &txid, const cryptonote::transaction& tx)
```
rbrunner7, line 5512: I just love things like a 9 year old "TODO" in the Wallet API implementation regarding implementing `on_skip_transaction` :)

I agree with your decision to not port this, the situation this is designed to report probably won't ever occur now.

src/simplewallet/simplewallet.cpp
```diff
@@ -5517,122 +5487,135 @@ bool simple_wallet::set_daemon(const std::vector<std::string>& args)
...
-{
-  if (m_locked)
-    return;
+    m_refresh_progress_reporter.update(ed->blockHeight(), true);
+
+  m_wallet_impl->history()->refresh();
 }
 //----------------------------------------------------------------------------------------------------
-boost::optional<epee::wipeable_string> simple_wallet::on_get_password(const char *reason)
+Monero::optional<std::string> simple_wallet::onGetPassword(const char *reason)
```
rbrunner7, line 5611: I feel a bit unsure here. Can you please quickly explain why you switch here from `epee::wipeable_string` to simply `std::string`? I guess this goes together with a modified approach regarding password "wiping".

Maybe a good place to add a short comment to the code about this as well?

src/simplewallet/simplewallet.cpp
```diff
@@ -5961,29 +5924,29 @@ bool simple_wallet::show_incoming_transfers(const std::vector<std::string>& args
       }
       std::string extra_string;
       if (verbose)
-        extra_string += (boost::format("%68s%68s") % td.get_public_key() % (td.m_key_image_known ? epee::string_tools::pod_to_hex(td.m_key_image) : td.m_key_image_partial ? (epee::string_tools::pod_to_hex(td.m_key_image) + "/p") : std::string(64, '?'))).str();
+        extra_string += (boost::format("  <%64s>%68s") % ed->onetimeAddress() % (ed->isKeyImageKnown() ? ed->keyImage() : ed->isKeyImagePartial() ? (ed->keyImage() + "/p") : std::string(64, '?'))).str();
```
rbrunner7: Not sure, maybe the format string should have the two blanks *after*, not *before*, like so? `"<%64s>  %68s"`

How does "%s" justify?

rbrunner7, line 5930: I am not sure whether you want to introduce some kind of new notation for output of some values with `<%64s>`, but if you do I suspect that you don't yet format consistently in this new way in all places, as you still have 3 occurences of `%68s%68s` ...

src/simplewallet/simplewallet.cpp
```diff
@@ -6068,11 +6040,14 @@ bool simple_wallet::show_payments(const std::vector<std::string> &args)
...
 {
-  if (!m_wallet)
+  if (!m_wallet_impl)
   {
     throw std::runtime_error("simple_wallet null wallet");
   }
-  return m_wallet->get_daemon_blockchain_height(err);
+  int error_code;
+  uint64_t bc_height = m_wallet_impl->daemonBlockChainHeight();
+  m_wallet_impl->statusWithErrorString(error_code, err);
```
rbrunner7, line 6048: Does this call have any effect if you do nothing with the set arguments?

src/simplewallet/simplewallet.cpp
```diff
@@ -6187,80 +6135,72 @@ bool simple_wallet::process_ring_members(const std::vector<tools::wallet2::pendi
     return false;
   }
   // for each transaction
-  for (size_t n = 0; n < ptx_vector.size(); ++n)
+  std::vector<std::string> tx_ids = ptx_vector.txid();
```
rbrunner7: My C++ is already a bit rusty, but couldn't you work here with mere references for all these vectors, or at least make them `const` to allow the compiler to optimize a bit?

src/simplewallet/simplewallet.cpp
```diff
@@ -7764,6 +7194,7 @@ bool simple_wallet::sign_transfer(const std::vector<std::string> &args_)
 
   bool export_raw = false;
   std::string unsigned_filename = "unsigned_monero_tx";
+  const std::string signed_filename = "signed_monero_tx";
```
rbrunner7, line 7240: Did you want to introduce a constant for this "magic" filename? If yes, actually use it below :)

src/simplewallet/simplewallet.cpp
```diff
@@ -7996,35 +7400,24 @@ bool simple_wallet::get_tx_proof(const std::vector<std::string> &args)
...
+    std::string error_msg;
+    m_wallet_impl->statusWithErrorString(error_code, error_msg);
+    fail_msg_writer() << tr("failed to get tx proof: ") << error_msg;
+    return true;
   }
+  const std::string filename = "monero_tx_proof";
+  if (m_wallet_impl->saveToFile(filename, sig_str, /* is_printable */ true))
+    success_msg_writer() << tr("signature file saved to: ") << filename;
+  else
+    fail_msg_writer() << tr("failed to save signature file");
```
rbrunner7, line 7477: Shouldn't you use here the usual pattern of getting an error message through `statusWithErrorString`, and append that?

src/simplewallet/simplewallet.cpp
```diff
@@ -8207,37 +7595,32 @@ bool simple_wallet::get_spend_proof(const std::vector<std::string> &args)
...
+    std::string error_msg;
+    m_wallet_impl->statusWithErrorString(error_code, error_msg);
+    fail_msg_writer() << tr("failed to get spend proof: ") << error_msg;
+    return true;
   }
+  const std::string filename = "monero_spend_proof";
+  if (m_wallet_impl->saveToFile(filename, sig_str, true))
+    success_msg_writer() << tr("signature file saved to: ") << filename;
+  else
+    fail_msg_writer() << tr("failed to save signature file");
```
rbrunner7, line 7679: A second place where I wonder why there is no `statusWithErrorString` call here to fetch details.

src/simplewallet/simplewallet.cpp
```diff
@@ -8479,169 +7861,218 @@ bool simple_wallet::get_transfers(std::vector<std::string>& local_args, std::vec
...
+  TransactionHistory *tx_history = m_wallet_impl->history();
+  tx_history->refresh();
   if (in || coinbase) {
-    std::list<std::pair<crypto::hash, tools::wallet2::payment_details>> payments;
-    m_wallet->get_payments(payments, min_height, max_height, m_current_subaddress_account, subaddr_indices);
-    for (std::list<std::pair<crypto::hash, tools::wallet2::payment_details>>::const_iterator i = payments.begin(); i != payments.end(); ++i) {
-      const tools::wallet2::payment_details &pd = i->second;
-      if (!pd.m_coinbase && !in)
+    std::vector<TransactionInfo *> payments;
+    auto get_payments = [&tx_history](std::vector<TransactionInfo *> &payments_out,
```
rbrunner7: I see where you come from with this construct. There was once `wallet2::get_payments` that now does not get a counterpart in the Wallet API. I think that's ok, it's easy enough to loop over the data yourself and pick what you like. So you replace the call to that method with a closure (correct term?) of the same name and leave the loop that processes the result identical.

Trouble is, as I see it, that the result is overly complicated code. You call `get_payments` exactly once, so why not refactoring further and just make a single loop over `tx_history->getAll()`  that starts with a somewhat complex condition whether to take the transaction or not?

src/simplewallet/simplewallet.cpp
```diff
@@ -9376,19 +8780,19 @@ void simple_wallet::print_accounts(const std::string& tag)
...
-      % print_money(m_wallet->unlocked_balance(account_index, false))
-      % m_wallet->get_subaddress_label({account_index, 0});
-    total_balance += m_wallet->balance(account_index, false);
-    total_unlocked_balance += m_wallet->unlocked_balance(account_index, false);
+      % m_wallet_impl->address(account_index, 0).substr(0, 6)
+      % print_money(m_wallet_impl->balance(account_index))
+      % print_money(m_wallet_impl->unlockedBalance(account_index))
+      % m_wallet_impl->getSubaddressLabel(account_index, 0);
+    total_balance += m_wallet_impl->balance(account_index);
+    total_unlocked_balance += m_wallet_impl->unlockedBalance(account_index);
```
rbrunner7: Little nitpick: You could put the result of `m_wallet_impl->unlockedBalance(account_index)` into a variable so that you don't calculate it twice, and `m_wallet_impl->balance(account_index)` likewise.

Yes, we are probably not in a place of the code where speed is of paramount importance, and we can afford to calculate such things twice, but still, the code would be trivial to improve, and IMHO it would become "nicer".

src/simplewallet/simplewallet.cpp
```diff
@@ -9522,15 +8927,15 @@ bool simple_wallet::print_address(const std::vector<std::string> &args/* = std::
     }
     if (index_max < index_min)
       std::swap(index_min, index_max);
-    if (index_min >= m_wallet->get_num_subaddresses(m_current_subaddress_account))
+    if (index_min >= m_wallet_impl->numSubaddresses(m_current_subaddress_account))
```
rbrunner7: `m_wallet_impl->numSubaddresses(m_current_subaddress_account)` could go into a variable, mostly not for speed, but for code clarity.

src/simplewallet/simplewallet.cpp
```diff
@@ -9648,8 +9057,7 @@ bool simple_wallet::address_book(const std::vector<std::string> &args/* = std::v
   }
   else if (args[0] == "add")
   {
-    cryptonote::address_parse_info info;
-    if(!cryptonote::get_account_address_from_str_or_url(info, m_wallet->nettype(), args[1], oa_prompter))
+    if(!Wallet::addressValid(/* address */ args[1], m_wallet_impl->nettype()))
```
rbrunner7, line 9071: Just a comment: I really like those "parameter explaining comments" like `/* address */` that you added in various places - it makes the code so much more readable :)

src/simplewallet/simplewallet.cpp
```diff
@@ -9672,25 +9087,20 @@ bool simple_wallet::address_book(const std::vector<std::string> &args/* = std::v
...
   }
   else
   {
-    for (size_t i = 0; i < address_book.size(); ++i) {
-      auto& row = address_book[i];
+    for (size_t i = 0; i < address_book_entries.size(); ++i) {
+      auto& row = address_book_entries[i];
       success_msg_writer() << tr("Index: ") << i;
-      std::string address;
-      if (row.m_has_payment_id)
```
rbrunner7, line 9590: A bit unsure here: Can we just silently drop support for payment ids as stored in address books? We really don't have to worry about old address books that still contain those?

src/simplewallet/simplewallet.cpp
```diff
@@ -9880,11 +9284,11 @@ bool simple_wallet::sign(const std::vector<std::string> &args)
     }
     else if (args[idx] == "--spend")
     {
-      message_signature_type = tools::wallet2::sign_with_spend_key;
+      do_sign_with_spend_key = false;
```
rbrunner7: You probably should set to `true` here?

src/simplewallet/simplewallet.cpp
```diff
@@ -10134,25 +9525,27 @@ bool simple_wallet::export_outputs(const std::vector<std::string> &args_)
...
-      fail_msg_writer() << tr("failed to save file ") << filename;
-      return true;
-    }
+     SCOPED_WALLET_UNLOCK();
+     data = m_wallet_impl->exportEnotesToStr(all);
   }
-  catch (const std::exception &e)
+  int error_code;
+  std::string error_msg;
+  m_wallet_impl->statusWithErrorString(error_code, error_msg);
```
rbrunner7, line 9550: A quite general question that crossed my mind while reading this code: With the Wallet API approach of leaving status after a call that we can then get with `statusWithErrorString`, do we have to worry about another thread running just between our call, like `exportEnotesToStr` here, and us fetching the result, where the in-between call done by the other thread would destroy our status info?

src/simplewallet/simplewallet.cpp
```diff
@@ -10212,29 +9608,34 @@ bool simple_wallet::show_transfer(const std::vector<std::string> &args)
...
     return true;
   }
-  crypto::hash txid = *reinterpret_cast<const crypto::hash*>(txid_data.data());
+  std::string txid = args.front();
+
+  const uint64_t last_block_height = m_wallet_impl->blockChainHeight();
+  TransactionHistory *tx_history = m_wallet_impl->history();
 
-  const uint64_t last_block_height = m_wallet->get_blockchain_current_height();
+  for (const auto &tx_info : tx_history->getAll())
```
rbrunner7, line 9627: The old code lists the transactions in 4 distinct blocks, with 4 separate loops, according to type, whereas your new code has only a *single* loop going over *all* transactions in the history. Does that work because the transactions in the history are ordered in the right way, so that still a listing of 4 distinct blocks according to type results?

src/simplewallet/simplewallet.cpp
```diff
@@ -10360,40 +9741,53 @@ void simple_wallet::interrupt()
...
       else
-        fail_msg_writer() << tr("Failed to save transaction to ") << filename << tr(", txid ") << txid;
+        fail_msg_writer() << ptx.errorString();
     }
     else
     {
-      m_wallet->commit_tx(ptx);
-      success_msg_writer(true) << tr("Transaction successfully submitted, transaction ") << txid << ENDL
-      << tr("You can check its status by using the `show_transfers` command.");
+      // ptx.commit() pops the pending txs after successfully commiting them, therefore we cache the txids before
```
rbrunner7: Small typo in comment: It's *committing* with double-t :)

rbrunner7 (commented):
I reviewed `simplewallet.cpp` and `simplewallet.h`.

Impressive work done so far by @SNeedlewoods. I look forward to the new, Wallet API based CLI wallet app!

src/simplewallet/simplewallet.h
SNeedlewoods: Referring to the comment from @jeffro256 [here](https://github.com/seraphis-migration/monero/pull/28#discussion_r2060401813)
> [...] unmixable sweep transactions [...] don't need a grace period. It would be better to cut them off early than accidentally introduce some niche bug in the window when all three types of transactions are allowed.

Also mentioned in @jeffro256's [draft for v17 consensus rule changes](https://github.com/monero-project/monero/pull/9751#issue-2817355574)
You will be able to spend old unmixable outputs in FCMP++ txs with upcoming v17 fork, but new unmixable sweep txs will be rejected.

Please correct me if I'm wrong, but considering it will probably take a while until changes from this PR and all PRs it relies on are finally merged, in their own PRs, into `release` branch I'd argue it makes sense to remove that function here.

src/simplewallet/simplewallet.cpp
SNeedlewoods: Noticed something unrelated: FWIW this is the only check for `AskPassword_OnAction`, and it's an implicit check. Elsewhere in the code we only check if `ask_password_type == AskPassword_Decrypt`.

src/simplewallet/simplewallet.cpp
SNeedlewoods: I think it should not be a problem, if you put a `try` block around `get_daemon_init_settings()`, which is only called once in `simple_wallet::init_wallet()`.

src/simplewallet/simplewallet.cpp
SNeedlewoods: Oh yes, that's possible, we have to `static_cast<int>` it, but I think it's worth it.

AFAICT [this PR](https://github.com/monero-project/monero/pull/9838/files) changed some code in that area, but this exact line was not touched by it.

src/simplewallet/simplewallet.cpp
SNeedlewoods: There are a few methods where the checks look similar on first sight, but they're not, so you have to be careful, but I think a `CHECK_MULTISIG_READY()` for finalized multisig wallets is a good idea.

src/simplewallet/simplewallet.cpp
SNeedlewoods: AFAICT this is referring to the output from `refresh_progress_reporter_t::update()` https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/simplewallet/simplewallet.h#L376 
Maybe now it's needed even more than before. I mentioned it in a meeting a while ago, that the output of this method would show up now in places where it didn't before. Will keep an eye on this when testing.

src/simplewallet/simplewallet.cpp
SNeedlewoods: This was wrong and messy, should be better now.

src/simplewallet/simplewallet.cpp
SNeedlewoods: Good question, so AFAICT cold signing is handled by the implementation, when you create a tx with `createTransactionMultDest()` it calls `pendingTxPostProcess()` which then calls `wallet2::cold_sign_tx()` [src](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/wallet/api/wallet.cpp#L2492-L2505).
I added `coldSignTx()` to #9464 before I saw this, now I think maybe we can get rid of it.
Would be great if anyone with hw wallet experience can comment on this.

src/simplewallet/simplewallet.cpp
SNeedlewoods: I identified the following removed methods:
- `sweep_unmixable()`
- `unset_ring()`
- `save_known_rings()`
- `blackball()`
- `unblackball()`
- `blackballed()`
- `cold_sign_tx()`
- `accept_loaded_tx()`, multiple ones with different params
- the entire `MMS` block


Before I add the comments, I would like to bring this up: just noticed @jeffro256 made a [PR](https://github.com/monero-project/monero/pull/10014/files) in which the functionality of `save_known_rings()` got removed and instead it prints "save_known_rings is deprecated". Should I rather remove my changes and use jeffros code, or is it okay to completely remove the method now?
Maybe it even makes sense to stick to jeffros way for all (or some of) the removed methods?

Or if you think it's okay to remove the methods and just put comments in place of the removed methods:
Would you also add comments to `simplewallet.cpp` or is it sufficient to have them only in the header file?

src/simplewallet/simplewallet.cpp
SNeedlewoods: That's right, missed this.
Currently we don't have access to `m_kdf_rounds` from `wallet2`, so I added `m_kdf_rounds` to `Wallet` (it's already an arg in the ctor [src](https://github.com/monero-project/monero/blob/d0d418483514d3caf79547e302937e45878eab21/src/wallet/api/wallet.cpp#L427)). I also completely removed `kdf_rounds` arg from `verifyPassword()` and made the impl use its new `m_kdf_rounds`.
AFAICT if you want to verify a password for a wallet which is not opened, you'd use `WalletManager::verifyWalletPassword()` which still has a `kdf_rounds` argument.

src/simplewallet/simplewallet.cpp
SNeedlewoods: AFAIUI `do_delete_pointer` is for raw pointers, but we use a `unique_ptr` for the wallet. If we have `do_delete_pointer = true`, first the `WalletManager` calls `delete` on the wallet and then, when our `unique_ptr` to the wallet goes out of scope it also tries to delete the wallet and that leads to "double delete"
TLDR: closing a wallet segfaults with `do_delete_pointer = false`

src/simplewallet/simplewallet.cpp
SNeedlewoods: I didn't see a problem with this, when receiving a tx under "normal conditions" [src](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/wallet/wallet2.cpp#L2508-L2574), but there is also [this path](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/wallet/wallet2.cpp#L2596-L2652) where we call `on_money_received()` without adding a new enote to `m_transfers`, according to log message: "output already exists, replacing with new output".

So you're right. I think a more sane approach would be to use `getEnoteDetails()` by `enote_pub_key` instead of `enote_index`, but I think that requires a change to `on_money_received()` in wallet2.h and `moneyReceived()` to take a new arg `enote_pub_key`. Done [here](https://github.com/monero-project/monero/compare/46a20e84d9f07ff2bc3a304c822c69b25ec4eded..e8f5106366bd0412adccba8672582506f35a5bb4)

src/simplewallet/simplewallet.cpp
SNeedlewoods: Just for reference, [here](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/wallet/wallet2.cpp#L2338-L2346) is the only place where `on_skip_transaction()` gets called.

src/simplewallet/simplewallet.cpp
SNeedlewoods: This is a Wallet API callback and we don't want to add anything outside the C++ standard to the API, `wipeable_string` would require all implementations to link against `libepee`.
Discussion can be found here: https://github.com/monero-project/monero/pull/9915#discussion_r2064388854

Considering password wiping, that's still missing from this method, because it's not used in the GUI, but I think that should be done in `WalletImpl::on_get_password()` (introduced in #9464), where the `optional<std::string>` from this method is received and turned into `wipeable_string` before sending it further down to `wallet2`.


vtnerd also recommended to change `const char*` to `std::sting_view` [here](https://github.com/monero-project/monero/pull/10232#discussion_r2658678511) and I think it may also make sense to change the return type from `optional<std::string>` to `optional<std::string_view>`.


I haven't changed anything here yet, because I'd like to give time for more opinions.

src/simplewallet/simplewallet.cpp
SNeedlewoods: I just tried to match the output from the old CLI, which e.g. sometimes prints txids surrounded by "<>" and sometimes without.
See `show_transfers` prints txid without "<>"
whereas `incoming_transfers verbose` prints txids and onetime addresses surrounded by "<>".

src/simplewallet/simplewallet.cpp
SNeedlewoods: `error_code` gets ignored, but the error message `err` is also an outparam of `get_daemon_blockchain_height()`

src/simplewallet/simplewallet.cpp
SNeedlewoods: Good call, I think `const` references are the most efficient here.

src/simplewallet/simplewallet.cpp
SNeedlewoods: The implementation had no error handling ([src](https://github.com/monero-project/monero/compare/77bfbd7a598823d510ff43f999e6609dbea412be..b496c2e5495990e5ac0acf3c9bec10db0ddc234d#diff-fcae134993cca2b6f1e887a60638f226ab00165e30bd4d326c17f0a58e11871fL3903-L3912)) and this method is now removed from the Wallet API ([vtnerd review](https://github.com/monero-project/monero/pull/10232#discussion_r2658672645)) and added to simplewallet.
It was just copied from wallet2, I don't know if there is much information to add or if the changes are worth it.

src/simplewallet/simplewallet.cpp
SNeedlewoods: Good call, almost embarrassing what I came up with, but at least it made me laugh.

src/simplewallet/simplewallet.cpp
SNeedlewoods: AFAICT we are not dropping any support here, the logic to show an integrated address if the `AddressBookRow` has a payment id is "outsourced" to the address book implementation [src](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/wallet/api/address_book.cpp#L103-L108)

src/simplewallet/simplewallet.cpp
SNeedlewoods: This is beyond my expertise, but I try to explain how I understand it:

For the `WalletImpl` we have `m_statusMutex` [src](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/wallet/api/wallet.h#L262-L264), but that only prevents multiple methods from writing to `m_errorString` at the same time, it doesn't prevent the second method to overwrite the error message when the lock of the first method is released, before we read it.
So I think in theory it is possibile that a function on another thread can change the error message before we read it, but I haven't seen it happen.

One fix to make sure we don't miss any error messages could be to change `m_errorString`s type from `std::string` to `std::vector<std::string>`, then we can pop the vector until every message is read, but this comes with the downside that you would receive some of the error messages at the wrong place.
IMO it may be worth a separate PR, if anyone sees a nice solution for this.


Note: for all of these, there is not even a mutex:
`AddressBookImpl` [src](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/wallet/api/address_book.h#L64-L65)
`PendingTransactionImpl` [src](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/wallet/api/pending_transaction.h#L66-L67)
`UnsignedTransactionImpl` [src](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/wallet/api/unsigned_transaction.h#L66-L67)
`WalletManagerImpl` [src](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/wallet/api/wallet_manager.h#L101)

src/simplewallet/simplewallet.cpp
SNeedlewoods: Correct, the ordering into the same 4 blocks is handled by the implementation: [src](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/wallet/api/transaction_history.cpp#L106).

src/wallet/api/transaction_history.cpp
```diff
@@ -229,14 +241,22 @@ void TransactionHistoryImpl::refresh()
...
         {
             ti->m_transfers.push_back({d.amount, d.address(m_wallet->m_wallet->nettype(), pd.m_payment_id)});
         }        
         m_history.push_back(ti);
     }
     
-    
+    // simplewallet called m_wallet.update_pool_state() & m_wallet.process_pool_state() before getting unconfirmed incoming transfers from the tx pool
+    // src: https://github.com/monero-project/monero/blob/8d4c625713e3419573dfcc7119c8848f47cabbaa/src/simplewallet/simplewallet.cpp#L10293-L10296
+//    m_wallet->refreshPoolOnly();
```
SNeedlewoods: This line causes `monero-wallet-cli` to ask for a password for each tx you received, instead of asking for a single password for all received txs until refresh is done.
It somehow resets `m_encrypt_keys_after_refresh` in `wallet2`, but could not find a fix for it, therefore I just commented this line out, but without it we don't find incoming txs in the pool.

Suggestions are welcome.

src/wallet/api/wallet.h
```diff
@@ -228,26 +241,169 @@ class WalletImpl : public Wallet
...
+    std::uint64_t importKeyImages(const std::vector<std::pair<std::string, std::string>> &signed_key_images, std::size_t offset, std::uint64_t &spent, std::uint64_t &unspent, bool check_spent = true) override;
+    bool importKeyImages(const std::vector<std::string> &key_images, std::size_t offset = 0, const std::unordered_set<std::size_t> &selected_enotes_indices = {}) override;
+    bool getAllowMismatchedDaemonVersion() const override;
+    void setAllowMismatchedDaemonVersion(bool allow_mismatch) override;
+    std::string getDeviceDerivationPath() const override;
+    void setDeviceDerivationPath(std::string device_derivation_path) override;
+    bool setDaemon(const std::string &daemon_address, const std::string &daemon_username = "", const std::string &daemon_password = "", bool trusted_daemon = false, Wallet::SSLSupport ssl_support = Wallet::SSLSupport::SSLSupport_Autodetect, const std::string &ssl_private_key_path = "", const std::string &ssl_certificate_path = "", const std::string &ssl_ca_file_path = "", const std::vector<std::string> &ssl_allowed_fingerprints_str = {}, bool ssl_allow_any_cert = false, const std::string &proxy = "") override;
```
vtnerd: This overlaps with init, and should probably be merged with that. Otherwise I'm excited for this as it will benefit lwcli security to have a custom cert selection.

vtnerd (changes requested):
Not a fully review, just a quick observation.

SNeedlewoods:
1. small update
2. rebased
_______________
The first three commits of this PR contain #9464.
The fourth commit of this PR contains #10232.
Fifth commit contains all the code for this PR.

PR #10240, opened 2025-11-29
cryptonote_core: set prep-blocks-threads to max threads

### LINKED ISSUE #10239: default value of prep-blocks-threads far to low
default value of prep-blocks-threads, defined as "lower of 4 or max threads", causes significant slowdown in non-checkpointed sync.
note: this has remained unchanged since implemented in 2015.

I noted consistent 45 and 70% speedup across 2 minute runs with `--prep-blocks-threads=$(nproc)` on an 8 thread aarch64 system, and a 256 thread x86_64 system, respectively.

suggestion: use max threads by default

0xFFFC0000:
Fixes https://github.com/monero-project/monero/issues/10239

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -36,6 +36,7 @@
 using namespace epee;
 
 #include <unordered_set>
+#include <thread>
 #include "cryptonote_core.h"
 #include "common/util.h"
 #include "common/updates.h"
@@ -156,7 +157,7 @@ namespace cryptonote
   static const command_line::arg_descriptor<uint64_t> arg_prep_blocks_threads = {
     "prep-blocks-threads"
   , "Max number of threads to use when preparing block hashes in groups."
-  , 4
+  , std::thread::hardware_concurrency() > 0 ? std::thread::hardware_concurrency() : 4
   };
   static const command_line::arg_descriptor<uint64_t> arg_show_time_stats  = {
     "show-time-stats"


0xFFFC0000:
Failed test unrelated

nahuhh:
Tested on linux x64 and aarch64, works as expected

j-berman (commented):
I assume this was placed here to limit memory when computing block hashes. Specifically when calling the `block_longhash_worker` function. I think it's worth a bit more analysis into what effect this would/could have on memory. Maybe worth trying to ask around why this arg is here (and why the default of 4) in the first place if possible too.

nahuhh:
> I assume this was placed here to limit memory when computing block hashes. Specifically when calling the `block_longhash_worker` function. I think it's worth a bit more analysis into what effect this would/could have on memory. Maybe worth trying to ask around why this arg is here (and why the default of 4) in the first place if possible too.

it was put here in a large pr of experimental changes in 2015

The fact that is uses max threads if the system has less than 4, tells me that its an arbitrary optimization where (at the time) most hardware was quad core.

give me a sec to find the commit and pr

j-berman:
Assuming it's sane to implement this change, then we may as well remove all usage of `m_max_prepare_blocks_threads` and remove all the logic surrounding this arg in particular. Keeping the arg around I guess could be a nice stop-gap solution in case there is some issue with this change in the future.. but keeping it around would seem like unnecessary tech debt to me.

nahuhh:
#337 is the pr and this commit: https://github.com/monero-project/monero/commit/4aea07c3cc291a51f4523274af32dbcd2fc26b8f

> and remove all the logic surrounding this arg in particular.

i think its might be still sane to keep as an arg, for ppl who want to leave some threads available during IBD

j-berman:
> for ppl who want to leave some threads available during IBD

Interesting point, ok. Code seems fine for the stated goal then.

My only remaining question mark for the change is on memory impact.

jpk68 (approved):

selsta (approved):

selsta:
The memory impact is roughly 2 MiB per active thread. The large RandomX dataset is shared between threads, so it is not duplicated.

PR #10241, opened 2025-11-29
Add json output support to ::wire and partially use in ZMQ-PUB

vtnerd:
The community has accepted my write/output proposal for objects, but not the read proposal (which is _way_ more complicated). This adds a json output using this new system, and uses it in _some_ of the ZMQ-PUB subsystems.

The next step is adding msgpack (or cbor), such that ZMQ-REP uses binary for some of the responses, and ZMQ-PUB uses msgpack/cbor for some of the notifications. The biggest benefit will be a reduction in scan time in LWS by at least 11%, with more gains possible as features are added. The second benefit will be slightly lower latency in mining notifications.

### Background

I've done an experimental build where `monerod` returns blocks via msgpack where each tx+block is encoded as one giant blob (as opposed to individual fields). The performance improvement for LWS scanning is ~11%. This is faster than HTTP-binary because ZMQ is a faster protocol as well (everything is binary). The 11% is conservative, as there are more cycle reduction in using integer keys and binary requests (my experimental branch had requests in json but responses in msgpack).

### Edit
The 11% improvement is for **total scan time**, not just the unpack performance. So this isn't just a small niche thing for LWS, sadly.

Integer keys speed up scan performance by ~0.97%, so it's not clear the feature is worth the extra complexity.

### DIFF
--- a/contrib/epee/include/serialization/wire/json/write.h
+++ b/contrib/epee/include/serialization/wire/json/write.h
@@ -1,0 +1,168 @@
+// Copyright (c) 2020, The Monero Project
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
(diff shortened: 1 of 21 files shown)


vtnerd:
Fingers crossed this actually gets reviewed!

vtnerd:
Force pushed a linker change, that should hopefully work with fuzz harness.

src/cryptonote_core/cryptonote_tx_utils.cpp
```diff
@@ -181,6 +182,16 @@ namespace cryptonote
...
   //---------------------------------------------------------------
+  namespace
+  {
+    template<typename F, typename T>
+    void map_backlog_entry(F& format, T& self)
+    {
+      wire::object(format, WIRE_FIELD(id), WIRE_FIELD(weight), WIRE_FIELD(fee));
+    }
+  }
+  WIRE_DEFINE_OBJECT(tx_block_template_backlog_entry, map_backlog_entry);
```
selsta, line 229: `rpc_pub` currently does not link `cryptonote_core`

contrib/epee/include/serialization/wire/json/write.h
```diff
@@ -0,0 +1,168 @@
...
+    void start_object(std::size_t) override final;
+    void key(std::string_view) override final;
+    void binary_key(epee::span<const std::uint8_t>) override final;
+    void end_object() override final;
+  };
+
+  //! Buffers entire JSON message in memory
+  struct json_string_writer final : json_writer
+  {
+    //! Writer json to an empty buffer.
```
selsta:
```suggestion
    //! Writes JSON to an empty buffer.
```

selsta (commented):
Needs rebase.

src/cryptonote_core/cryptonote_tx_utils.cpp
vtnerd: Fixed.

contrib/epee/include/serialization/wire/json/write.h
vtnerd: Fixed.

vtnerd:
Force pushed a rebase and selsta review changes.

selsta (approved):

src/rpc/zmq_pub.cpp
```diff
@@ -106,23 +110,48 @@ namespace
...
   };
+  void write_bytes(wire::writer& dest, const minimal_chain& self)
+  {
+    namespace adapt = boost::adaptors;
+
+    const auto to_block_id = [](const cryptonote::block& bl)
+    {
+      crypto::hash id;
+      if (!get_block_hash(bl, id))
+        MERROR("ZMQ/Pub failure: get_block_hash");
```
SChernykh: It should return a null hash if `get_block_hash` failed, otherwise this lambda will return an uninitialized value. Essentially, it will leak whatever data was on stack.

SChernykh:
LGTM overall, just one small nitpick (see above). Another one is the lack of new tests here. Do old tests cover the ZMQ-PUB JSON format for all messages?

vtnerd:
> Do old tests cover the ZMQ-PUB JSON format for all messages?

There are tests for everything _except_ `json-full-miner_data`. I'll figure out how to add one.

src/rpc/zmq_pub.cpp
vtnerd: This arguable should throw a `wire::exception_t` type so that it stops processing, unwinds, and then returns a `std::error_code`. I'll update a bit after looking at this further.

vtnerd: This now throws a wire exception which unwinds to the root object then returns a `std::error_code`. So everything gets logged but nothing is ultimately written to ZMQ-PUB socket (this should basically never happen though).

vtnerd:
Force pushed tests for ZMQ-PUB `miner_data` and added a wire-throw when the block hashing fails.

selsta (approved):

SChernykh (approved):

PR #10244, opened 2025-12-02
`export_transfers` command in wallet didn't check for output file writing error

Low-power:
The **export_transfers** command in **monero-wallet-cli** does not have any error handling logic for writing the output file; it will happily report the CSV data as 'exported', even the output file could not be opened.

For example, the following 2 operations are expected to fail with errors `EISDIR` and `EACCES`, **monero-wallet-cli** however incorrectly reports that the CSV files are exported successfully:
```
[wallet A2Sbbz]: export_transfers output=/
CSV exported to /
[wallet A2Sbbz]: export_transfers output=/no-creation-permission
CSV exported to /no-creation-permission
```

After the fix:
```
[wallet A2Sbbz]: export_transfers output=/
Error: Failed to export CSV to /
[wallet A2Sbbz]: export_transfers output=/no-creation-permission
Error: Failed to export CSV to /no-creation-permission
[wallet A2Sbbz]: export_transfers output=/tmp/write-accessible-path
CSV exported to /tmp/write-accessible-path
```

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -8768,6 +8768,10 @@ bool simple_wallet::export_transfers(const std::vector<std::string>& args_)
   }
 
   std::ofstream file(filename);
+  if(file.fail()) {
+    fail_msg_writer() << boost::format(tr("Failed to open %s for writing")) % filename;
+    return true;
+  }
 
   // header
   file <<
@@ -8837,7 +8841,11 @@ bool simple_wallet::export_transfers(const std::vector<std::string>& args_)
   }
   file.close();
 
-  success_msg_writer() << tr("CSV exported to ") << filename;
+  if(file.fail()) {
+    fail_msg_writer() << tr("Failed to export CSV to ") << filename;
+  } else {
+    success_msg_writer() << tr("CSV exported to ") << filename;
+  }
 
   return true;
 }


src/simplewallet/simplewallet.cpp
```diff
@@ -8837,7 +8837,11 @@ bool simple_wallet::export_transfers(const std::vector<std::string>& args_)
   }
   file.close();
 
-  success_msg_writer() << tr("CSV exported to ") << filename;
+  if(file.fail()) {
```
SNeedlewoods, line 8844: I think you can check earlier if opening the file failed here: https://github.com/monero-project/monero/blob/eaa2c93c2b016811ce1f9c578f716fdbc8f51ce3/src/simplewallet/simplewallet.cpp#L8770
```suggestion'
  std::ofstream file(filename);
  if(!file) {
    fail_msg_writer() << tr("failed to open file: ") << filename;
    return true;
  } 
```

Low-power:
> I think you can check earlier if opening the file failed here:

The test result is now:
```
[wallet A2Sbbz]: export_transfers output=/dev
Error: Failed to open /dev for writing
[wallet A2Sbbz]: export_transfers output=/dev/mem
Error: Failed to open /dev/mem for writing
[wallet A2Sbbz]: export_transfers output=/dev/zfs
Error: Failed to export CSV to /dev/zfs
[wallet A2Sbbz]: export_transfers output=/tmp/csv
CSV exported to /tmp/csv
```

PR #10253, opened 2025-12-11
Parallelize transaction verification within blocks during sync

alextodd1:
This commit adds parallel verification of transactions within a block during blockchain synchronization. The key changes:

- Restructured handle_block_to_main_chain() into 3 phases:
  1. Sequential TX collection from mempool/supplement (cannot be parallelized due to shared mempool state modifications)
  2. Parallel TX input verification using the existing threadpool
  3. Sequential fee/weight accumulation after verification

- Added early-exit optimization: when a TX fails verification, other threads can skip remaining work using an atomic flag

- For single TX or single-threaded environments, verification still runs sequentially to avoid unnecessary overhead

Thread-safety analysis:
- DB reads use thread-local LMDB read transactions (each thread gets its own MDB_txn* via boost::thread_specific_ptr)
- m_rct_ver_cache is internally synchronized with std::mutex
- check_tx_inputs() only performs read operations on blockchain state
- Intra-block double-spend detection handled at add_block() time by DB

This change can significantly speed up initial blockchain sync on multi-core systems by utilizing available CPU cores for cryptographic verification (ring signatures, RCT proofs) across multiple transactions simultaneously.

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -49,6 +49,7 @@
 #include "file_io_utils.h"
 #include "int-util.h"
 #include "common/threadpool.h"
+#include <atomic>
 #include "warnings.h"
 #include "crypto/hash.h"
 #include "cryptonote_core.h"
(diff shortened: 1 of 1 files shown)


alextodd1:
posted by accident

PR #10254, opened 2025-12-12
wallet2: fix edge case where tx's ki's remain marked unspent

j-berman:
If a tx is marked as failed (because it never shows up in the daemon's pool), its key images get reset back to unspent so they can be used in future txs.

If the tx re-enters the daemon's pool (e.g. it's removed from the pool and then relayed back), then the wallet incorrectly maintains that the tx's key images are unspent.

This change ensures the wallet re-marks the tx's key images as spent if the tx re-appears in the node's pool.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -3656,6 +3656,35 @@ bool wallet2::accept_pool_tx_for_processing(const crypto::hash &txid)
 // Process an unconfirmed transfer after we know whether it's in the pool or not
 void wallet2::process_unconfirmed_transfer(bool incremental, const crypto::hash &txid, wallet2::unconfirmed_transfer_details &tx_details, bool seen_in_pool, std::chrono::system_clock::time_point now, bool refreshed)
 {
+  const auto set_tx_key_images_spent = [&](const bool spent)
+  {
+    for (size_t vini = 0; vini < tx_details.m_tx.vin.size(); ++vini)
+    {
+      if (tx_details.m_tx.vin[vini].type() != typeid(txin_to_key))
+        continue;
+
+      const crypto::key_image &key_image = boost::get<txin_to_key>(tx_details.m_tx.vin[vini]).k_image;
+      const auto it_ki = m_key_images.find(key_image);
+      if (it_ki == m_key_images.end())
+        continue;
+
+      const std::size_t i = it_ki->second;
+      if (i >= m_transfers.size())
+        continue;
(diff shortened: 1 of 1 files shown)


src/wallet/wallet2.cpp
```diff
@@ -3665,6 +3694,10 @@ void wallet2::process_unconfirmed_transfer(bool incremental, const crypto::hash
       tx_details.m_state = wallet2::unconfirmed_transfer_details::pending_in_pool;
       MINFO("Pending txid " << txid << " seen in pool, marking as pending in pool");
     }
+
+    // The inputs are spent, they're in the pool! It's possible the tx was previously marked as failed, so we
+    // make sure to re-mark the outputs as spent.
+    set_tx_key_images_spent_status(true/*spent*/);
```
iamamyth: Naming the function `set_tx_key_images_spent` would clarify the argument's meaning, eliminating the need for comments (`_status` adds verbosity and decreases clarity).

j-berman: I think it's fine

src/wallet/wallet2.cpp
```diff
@@ -3656,6 +3656,35 @@ bool wallet2::accept_pool_tx_for_processing(const crypto::hash &txid)
...
+      if (i >= m_transfers.size())
+        continue;
+      const transfer_details &td = m_transfers.at(i);
+      if (td.m_key_image != key_image)
+        continue;
+      if (td.m_spent == spent)
+        continue;
+
+      LOG_PRINT_L1("Resetting spent status for output " << vini << ": " << key_image);
+      if (spent)
```
rbrunner7, line 3681: I think this log print would be more valuable if it detailed how "spent" gets set, i.e. whether new status is spent or unspent.

src/wallet/wallet2.cpp
rbrunner7: Not sure whether this answer means you agree with @iamamyth, but anyway, for what it's worth, I would probably also call the function `set_tx_key_images_spent`. Not only for clarity, but also because the member that gets set has a name of `m_spent` after all, not `m_spent_status`.

j-berman: Ok done in the latest

nahuhh:
Ive spammed 55000 txs across 30 wallets w/o running into a double spend error.
usually, id have plenty.

LGTM

jeffro256 (approved):
Nice! Changes look good to me.

vtnerd (approved):

PR #10265, opened 2025-12-20
epee: optimize uri conversion functions

leezaj:
- greatly reduced # of string allocations (only 1 per function)
- fixed typo in function name (conver -> convert)
- used more standard library functions to simplify manual logic and remove pointer arithmetic
- removed all functions in header that were only used by these 2 functions
- replaced const std::string& with std::string_view

Tested changes with unit tests and functional tests (especially uri.py)

### DIFF
--- a/contrib/epee/include/net/abstract_http_client.h
+++ b/contrib/epee/include/net/abstract_http_client.h
@@ -34,27 +34,8 @@ namespace epee
 {
 namespace net_utils
 {
-  inline const char* get_hex_vals()
-  {
-    static constexpr const char hexVals[16] = {'0','1','2','3','4','5','6','7','8','9','A','B','C','D','E','F'};
-    return hexVals;
-  }
-
-  inline const char* get_unsave_chars()
-  {
-    //static constexpr char unsave_chars[] = "\"<>%\\^[]`+$,@:;/!#?=&";
-    static constexpr const char unsave_chars[] = "\"<>%\\^[]`+$,@:;!#&";
-    return unsave_chars;
-  }
-
-  bool is_unsafe(unsigned char compare_char);
-  std::string dec_to_hex(char num, int radix);
-  int get_index(const char *s, char c);
-  std::string hex_to_dec_2bytes(const char *s);
-  std::string convert(char val);
-  std::string conver_to_url_format(const std::string& uri);
-  std::string convert_from_url_format(const std::string& uri);
-
+  std::string convert_to_url_format(std::string_view uri);
+  std::string convert_from_url_format(std::string_view uri);
 namespace http
 {
   class abstract_http_client
--- a/contrib/epee/src/abstract_http_client.cpp
+++ b/contrib/epee/src/abstract_http_client.cpp
@@ -1,3 +1,4 @@
+#include <charconv>
 #include <math.h>
 #include "net/abstract_http_client.h"
 #include "net/http_base.h"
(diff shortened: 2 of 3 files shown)


contrib/epee/src/abstract_http_client.cpp
```diff
@@ -7,113 +8,64 @@
 #undef MONERO_DEFAULT_LOG_CATEGORY
 #define MONERO_DEFAULT_LOG_CATEGORY "net.http"
 
+namespace {
+constexpr auto hex_byte_len = 2; // 00 to FF
+constexpr char prepend_token = '%'; // placed before each byte i.e. %61
```
iamamyth: prepend reflects the position of the token, which everyone can plainly see from the encode/decode algorithm; better to reflect its meaning/purpose, e.g. uri_escape_char.

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -7,113 +8,64 @@
...
-      csTmp = get_hex_vals()[temp];
-    }
-
-    csTmp += get_hex_vals()[num_char];
-
-    if(csTmp.size() < 2)
-    {
-      csTmp += '0';
+    static constexpr auto is_unsafe = [](unsigned char x) noexcept -> bool {
+      static constexpr std::string_view unsafe {"\"<>%\\^[]`+$,@:;!#&"};
```
iamamyth: This lookup method is painfully inefficient. For allocation, you can just assume 3x (always right), or guess a large over-allocation that's nearly always right, like 2x input size. For encoding, you may as well make a 255 entry lookup table for every single char, then you're just doing some copies per char.

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -7,113 +8,64 @@
...
-
+      unsigned char decimal{};
+      const auto hex_begin = it + 1; // skip '%'
+      const auto hex_end {hex_begin + hex_byte_len};
+      if (const auto [ptr, ec] = std::from_chars(hex_begin, hex_end, decimal, 16); ec == std::errc{} && ptr == hex_end) {
+        *(write_it++) = decimal;
+      } else {
+        write_it = std::copy(it, hex_end, write_it); // in case of failure, copy the original string's contents.
+      }
+      it = hex_end - 1; // will be == to hex_end next iteration.
```
iamamyth: Might be clearer to write this as a positive offset reflecting the extra chars processed, then you could likely drop the comment: `it += hex_byte_len;`

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -7,113 +8,64 @@
...
+        *(write_it++) = character;
+        continue;
       }
-      else
-        result += uri[i];
-
+      unsigned char decimal{};
+      const auto hex_begin = it + 1; // skip '%'
+      const auto hex_end {hex_begin + hex_byte_len};
+      if (const auto [ptr, ec] = std::from_chars(hex_begin, hex_end, decimal, 16); ec == std::errc{} && ptr == hex_end) {
```
iamamyth: I don't think any other code uses this syntactic construct (declare + init in control flow), suggest moving it to before the if (also applies to the earlier if).

leezaj:
ty for the review, all good suggestions. I've implemented each one and ran tests locally, lmk if anything else sticks out

leezaj:
Last commit was to replace a magic number, all good now.

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -7,113 +8,68 @@
...
-    std::string result;
-
-    for(size_t i = 0; i!= uri.size(); i++)
-    {
-      if(is_unsafe(uri[i]))
-        result += convert(uri[i]);
-      else
-        result += uri[i];
-
+    std::string result (uri.size() * (hex_byte_len + 1), '\0'); // reserve the max possible size, shrink later.
```
iamamyth, line 38: I think you're better off with a default-initialized string and then a `reserve` (avoids a pointless memset).

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -7,113 +8,68 @@
...
+    auto write_it {result.begin().base()};
+    for (const unsigned char character : uri) {
+      if (not unsafe_characters[character]) {
+        *(write_it++) = character;
```
iamamyth, line 43: The approach in the existing code of just appending to the string seems safer and therefore more desirable than using an iterator here (even though there shouldn't be overflows).

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -7,113 +8,68 @@
...
+        continue;
+      }
+      *(write_it++) = uri_escape_char;
+      const auto hex_end {write_it + hex_byte_len};
+      if (character <= 0xF) { // prepend single-character nibbles with a 0.
```
iamamyth: You could just use snprintf to output the value two-wide in uppercase hex.

contrib/epee/src/abstract_http_client.cpp
iamamyth: To clarify, by appending, I don't necessarily mean `append`, you could use `push_back` to ensure no allocations without overflowing the buffer.

contrib/epee/src/abstract_http_client.cpp
iamamyth: One thing to note: string doesn't seem to guarantee that writing past `data() + size()` in a reserved string works (see https://en.cppreference.com/w/cpp/string/basic_string/data.html), so you'll need to resize the string to accommodate the new chars and then call snprintf.

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -7,113 +8,68 @@
...
+    for (auto it {uri.cbegin()}, end {uri.cend()}; it < end; ++it) {
+      const auto character {*it};
+      if (character != uri_escape_char || std::distance(it, end) < (hex_byte_len + 1)) {
+        *(write_it++) = character;
+        continue;
       }
-      else
-        result += uri[i];
-
+      unsigned char decimal{};
```
iamamyth: `decimal` seems an odd name here, as it's not a decimal value; perhaps `decoded` would be better?

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -7,113 +8,68 @@
...
-    std::string result;
-
-    for(size_t i = 0; i!= uri.size(); i++)
-    {
-      if(uri[i] == '%' && i + 2 < uri.size())
-      {
-        result += hex_to_dec_2bytes(uri.c_str() + i + 1);
-        i += 2;
+    std::string result (uri.size(), '\0'); // reserve the max possible size, shrink later.
+    auto write_it {result.begin()};
```
iamamyth, line 56: Same thought here about using safe append operations instead of iterator assigns.

contrib/epee/src/abstract_http_client.cpp
iamamyth: Setting aside the above, I should clarify my earlier comment about using a lookup table: You can have the table do all the work here, including hex conversion. Just make each entry your output, and give them all a fixed width of 2 bytes (for literals which map 1-1, all null). This way, the lookup math is a simple bit shift, and your logic looks like (in pseudocode):
```
if (table_entry[0] != '\0') {
    push_back('%');
    memcpy 2 char entry;
} else {
    out = in
}
```

contrib/epee/src/abstract_http_client.cpp
```diff
@@ -7,113 +8,68 @@
...
+        *(write_it++) = character;
+        continue;
+      }
+      *(write_it++) = uri_escape_char;
+      const auto hex_end {write_it + hex_byte_len};
+      if (character <= 0xF) { // prepend single-character nibbles with a 0.
+        *(write_it++) = '0';
+      }
+      std::to_chars(write_it, hex_end, character, 16); // writes hex as lowercase.
+      std::transform(write_it, hex_end, write_it, ::toupper);
```
iamamyth: The trouble with toupper is that it's locale dependent.

contrib/epee/src/abstract_http_client.cpp
leezaj: 1 thing that needs clarification. I went with constructor deliberately instead of `reserve`. Below is my reasoning.

As mentioned, writing past `.data()` is UB, so with `reserve` we need to either `resize` then copy, or `push_back`.
- Problem with `resize` then copy: If we `resize` to _expand_ the string, it [fills out the contents out with `\0`'s anyway](https://en.cppreference.com/w/cpp/string/basic_string/resize.html), better to do this once with the constructor rather than do it each time. It also has the same 2 downsides listed below.
- Problem with `push_back`: Every push_back does 2 internal checks which are both wasteful:
    1. `push_back` needs to check if size would be greater than capacity to know if reallocation is needed. We know reallocation will never happen, because the size will never be  greater than capacity, because we've reserved the max possible size, and yet this check would still done with each `push_back` unnecessarily.
    2. `push_back` needs to update the size each time we use it.  Why not save instructions and instead just do it once at the end (which is what we do right now) instead of each time we `push_back`?

So I believe the constructor method is more performant. The only disadvantage is that it memsets, which one can't really help <sub>[other than using a complicated function from C++23](https://en.cppreference.com/w/cpp/string/basic_string/resize_and_overwrite)</sub>. Let me know what you think.

leezaj:
Implemented proper lookup table like you mentioned (incidentally getting rid of the `toupper` issue as `toupper` is not used anymore) and used better variable naming.

Only thing I didn't do is the reserve+append suggestion because of what I replied above, but lmk your thoughts and I can do so if still needed.

iamamyth:
Overall I'd be more concerned with safety than performance (this isn't a hot code path, but it is used in the wallet). You're right that resize also zeros the contents, and is therefore perhaps even worse than just zeroing at the beginning; I proposed on-demand resizing because then the string's size/length effectively replaces your iterator position in the current code. I don't particularly object to sizing the container at the start, your code is certainly correct, and the maximum size is a domain invariant, but I worry that mistakes respect none of these facts. That said, I'd sign off on the current approach.

iamamyth:
Just an FYI: My benchmarks with g++ show your implementation increases throughput by ~80% (i.e. almost halves execution time) in an unoptimized build; for a build with -O2 your implementation multiplies throughput by 7.3. The g++ optimizer completely closes the gap between "flat C array, indexed by bit shift, copied via memcpy" and your code.

leezaj:
Thanks a lot for sharing benchmarks and comparing builds, that's a really nice increase (especially with optimizations!). Even if not a particularly hot path, speedups like that are always nice. Your suggestions definitely helped here.

contrib/epee/include/net/abstract_http_client.h
```diff
@@ -34,27 +34,8 @@ namespace epee
...
-  inline const char* get_hex_vals()
-  {
-    static constexpr const char hexVals[16] = {'0','1','2','3','4','5','6','7','8','9','A','B','C','D','E','F'};
-    return hexVals;
-  }
-
-  inline const char* get_unsave_chars()
-  {
-    //static constexpr char unsave_chars[] = "\"<>%\\^[]`+$,@:;/!#?=&";
-    static constexpr const char unsave_chars[] = "\"<>%\\^[]`+$,@:;!#&";
```
ghost, line 46: Q: Why were the `?` and `=` removed from unsafe characters? For some reason, it's also being continued in the additions.

leezaj: Looks like that comment has always been there from the start, [see http_client.h in this commit](https://github.com/monero-project/monero/commit/296ae46ed8f8f6e5f986f978febad302e3df231a)

ghost: Need @iamamyth to chime in here. 
From from the wallet's perspective, if the recipient name or txn description had any `?` or `=`, they wouldn't be encoded since they're not in the list of unsafe characters.

iamamyth: The goal here is to preserve prior behavior, irrespective of whether said behavior makes sense. Removing the exemptions for `?` and `=` would have to occur in a separate PR (it potentially breaks a lot of things, precisely for whatever reasons gave rise to the exemptions).

iamamyth: It looks like the behavior could be changed to better conform to standards in a follow-on PR, without introducing any problems.

PR #7357, opened 2021-01-30
wallet2 api: additions for handling multisig txs by file

benevanoff:
These are some changes I made while adding multisig functionality to the GUI wallet. I've added methods to the wallet api for exporting and importing partially signed key images as well as methods for loading and signing multisig transactions from the filename. It also makes the change that the `createTransactionMultDest` function saves multisig transactions to a file in the current directory by default as well as adding a `dest_addresses` member to the PendingTransaction class for the destination addresses of the transaction.

Sorry if this should be split up into different PRs but I figured that way it'd be easier to grab the changes to test the gui wallet

PR #10267, opened 2025-12-27
wallet: enhance multisig file handling and API

ghost:
Rebasing and updating #7357 - for reference, that PR added methods to the wallet API for exporting and importing partially signed key images as well as methods for loading and signing multisig transactions from the filename. The development of it basically stalled.

- Improved error handling
- Changed `PEM_read_string`'s return value to a boolean
- Added `exportMultisigImagesPEM` and `exportMultisigImagesToFile` for exports and `importMultisigImagesFromFile` and `importMultisigImagesFromPEM` for multisig image imports. Removed old unclear functions for importing and exporting images
- `dest_addresses` renamed to `destAddresses` in order to be consistent with the camelCase in `pending_transaction.cpp` and zero copying
- Pass strings by const references in all functions except `exportMultisigImagesPEM`.
- Fix a `PendingTransactionImpl` memory leak in `loadMultisigTxFromFile`

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -1518,14 +1518,26 @@ std::string WalletImpl::getMultisigKeyExchangeBooster(const std::vector<std::str
     return string();
 }
 
-bool WalletImpl::exportMultisigImages(string& images) {
+bool WalletImpl::exportMultisigImages(string& images, std::string filename, bool ascii) {
     try {
         clearStatus();
         checkMultisigWalletReady(m_wallet);
 
         auto blob = m_wallet->export_multisig();
-        images = epee::string_tools::buff_to_hex_nodelimer(blob);
-        return true;
+        if (filename.empty()) {
+            if (ascii) {
+                BIO *bp = BIO_new(BIO_s_mem());
+                PEM_write_bio(bp, "MoneroAsciiDataV1", "", (const unsigned char *) blob.c_str(), blob.length());
+                BUF_MEM *mem = NULL;
+                BIO_get_mem_ptr(bp, &mem);
+                std::string pemData(mem->data, mem->length);
+                images = pemData;
+            } else {
+                images = epee::string_tools::buff_to_hex_nodelimer(blob);
+            }
+            return true;
+        }
+        return m_wallet->save_to_file(filename, blob);
     } catch (const exception& e) {
         LOG_ERROR("Error on exporting multisig images: " << e.what());
         setStatusError(string(tr("Failed to export multisig images: ")) + e.what());
@@ -1562,6 +1574,30 @@ size_t WalletImpl::importMultisigImages(const vector<string>& images) {
     return 0;
 }
 
+size_t WalletImpl::importMultisigImages(const std::string& input, bool isFile) {
+    try {
+        clearStatus();
+        checkMultisigWalletReady(m_wallet);
+
+        string data;
+        std::vector<cryptonote::blobdata> images;
+        if (isFile) {
+            if (!m_wallet->load_from_file(input, data)) {
+                throw runtime_error("Couldn't load from file");
+            }
+        } else {
+            m_wallet->PEM_read_string(input, data);
+        }
+
+        images.push_back(std::move(data));
+        return m_wallet->import_multisig(images);
+    } catch (const exception& e) {
+        LOG_ERROR("Error on importing multisig images: " << e.what());
+        setStatusError(string(tr("Failed to import multisig images: ")) + e.what());
+    }
+    return 0;
+}
+
 bool WalletImpl::hasMultisigPartialKeyImages() const {
     try {
         clearStatus();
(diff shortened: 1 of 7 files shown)


src/wallet/api/wallet.cpp
```diff
@@ -1518,14 +1518,26 @@ std::string WalletImpl::getMultisigKeyExchangeBooster(const std::vector<std::str
...
+bool WalletImpl::exportMultisigImages(string& images, std::string filename, bool ascii) {
     try {
         clearStatus();
         checkMultisigWalletReady(m_wallet);
 
         auto blob = m_wallet->export_multisig();
-        images = epee::string_tools::buff_to_hex_nodelimer(blob);
-        return true;
+        if (filename.empty()) {
+            if (ascii) {
```
iamamyth: This seems like quite a different use case than the existing function, I think a separate function would be better (also suggest `PEM` as the qualifier, rather than `ascii`, as all the formats are ASCII data).

src/wallet/api/wallet.cpp
```diff
@@ -1562,6 +1574,30 @@ size_t WalletImpl::importMultisigImages(const vector<string>& images) {
...
 
+size_t WalletImpl::importMultisigImages(const std::string& input, bool isFile) {
+    try {
+        clearStatus();
+        checkMultisigWalletReady(m_wallet);
+
+        string data;
+        std::vector<cryptonote::blobdata> images;
+        if (isFile) {
+            if (!m_wallet->load_from_file(input, data)) {
```
iamamyth: This also seems like a case where you have two functions living in the body of one function. If you want to share the pre and post code, there are other options than a single function, particularly considering this same pre/post-amble exists in pretty much every multi-sig function, so your sharing should probably either be wider or narrower.

src/wallet/wallet2.cpp
```diff
@@ -15231,57 +15236,64 @@ bool wallet2::save_to_file(const std::string& path_to_file, const std::string& r
 }
 //----------------------------------------------------------------------------------------------------
 
-bool wallet2::load_from_file(const std::string& path_to_file, std::string& target_str,
-                             size_t max_size)
+int wallet2::PEM_read_string(const std::string& data, std::string& target_str)
```
iamamyth: What was wrong with bool?

iamamyth:
It looks to me like the comments on the original PR weren't addressed, and I've effectively just repeated them. Stopping review.

src/wallet/api/wallet.cpp
```diff
@@ -1534,6 +1534,66 @@ bool WalletImpl::exportMultisigImages(string& images) {
...
+bool WalletImpl::exportMultisigImagesPEM(std::string& pem) {
+    try {
+        clearStatus();
+        checkMultisigWalletReady(m_wallet);
+        auto blob = m_wallet->export_multisig();
+        BIO *bp = BIO_new(BIO_s_mem());
+        if (!bp)
+            throw std::runtime_error("BIO allocation failed");
+        if (!PEM_write_bio(bp, "MoneroPEMDataV1", "", (const unsigned char *) blob.c_str(), static_cast<int>(blob.length()))) {
+            BIO_free(bp);
```
iamamyth: This seems like a case where you could use RAII (e.g. unique_ptr), as I note this pattern in multiple functions, though I don't particularly mind the explicit frees.

src/wallet/api/wallet.cpp
```diff
@@ -1534,6 +1534,66 @@ bool WalletImpl::exportMultisigImages(string& images) {
...
+        checkMultisigWalletReady(m_wallet);
+        auto blob = m_wallet->export_multisig();
+        if (!pem)
+            return m_wallet->save_to_file(filename, blob);
+        
+        std::string pemData;
+        BIO *bp = BIO_new(BIO_s_mem());
+        if (!bp) 
+            throw std::runtime_error("BIO allocation failed");
+        if (!PEM_write_bio(bp, "MoneroPEMDataV1", "", (const unsigned char *) blob.c_str(), static_cast<int>(blob.length()))) {
```
iamamyth: This seems awfully similar to the above function, it seems plausible to me that you can unify them with a streaming API, or by just dumping the data to a string, as it's quite small.

src/wallet/api/wallet.cpp
```diff
@@ -1562,6 +1622,42 @@ size_t WalletImpl::importMultisigImages(const vector<string>& images) {
...
 }
 
+size_t WalletImpl::importMultisigImagesFromFile(const std::string& filename) {
+    try {
+        clearStatus();
+        checkMultisigWalletReady(m_wallet);
+
+        std::string data;
+        if (!m_wallet->load_from_file(filename, data))
+            throw runtime_error("Couldn't load from file");
```
iamamyth, line 1586: The fact that this function can't read PEMs (i.e, you have an export/import asymmetry) seems an indication that the export function, above, should probably split the PEM functionality. But, I also don't understand why the API should support both formats. You're already stuck with image-based import elsewhere, I don't see what PEM brings to the table, considering that the entire protocol stack requires (and therefore implements) key image serialization.

ghost: > I don't see what PEM brings to the table, considering that the entire protocol stack requires (and therefore implements) key image serialization.

PEM was extended purely as a UX convenience for GUI users to copy and paste multisig data over text-only channels. From an architectural standpoint, IMHO it makes more sense for the wallet API to stay blob-based, and for higher-level layers (libwalletqt in GUI wallet) to handle PEM wrapping. Should we remove the PEM from here then? We already support hexadecimal representation of multisig blobs, supporting PEM seems like it's superfluous.

iamamyth: My comment was based (in part) on the observation that there's already support for hex, so PEM seems pointless (I don't even see why wrappers would support it). I would remove PEM entirely, it'll make this PR easier to review and merge, and create a space for a more focused conversation about whether (and how) to support PEM.

iamamyth:
One other broad note on this PR: Given the existence and relative priority of the wallet2/wallet_api refactor PRs, this PR will likely remain in the queue for a long time. I think it makes sense to drop PEM support and then just let this stay dormant for a while, as I doubt there'll be many available or interested reviewers.

PR #9808, opened 2025-02-19
logging: Generalize terminal color detection

iamamyth:
Assume the terminal supports color codes if TERM ends with `-color` or `-256color`, rather than special-casing a handful of such terminals.

Add tests for terminal color detection.

PR #10268, opened 2025-12-29
logging: Generalize terminal color detection

nahuhh:
> Assume the terminal supports color codes if TERM ends with `-color` or `-256color`, rather than special-casing a handful of such terminals.
> 
> Add tests for terminal color detection.

backport of #9808 (~~no changes~~ removed tests as they caused compilation failure)

### DIFF
--- a/external/easylogging++/easylogging++.cc
+++ b/external/easylogging++/easylogging++.cc
@@ -1244,11 +1244,19 @@ std::string OS::currentHost(void) {
 #endif  // ELPP_OS_UNIX && !ELPP_OS_ANDROID
 }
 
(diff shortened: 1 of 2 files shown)


tobtoht (approved):

tobtoht:
CI fails because `master` uses the `gtest` submodule in external (`v1.15.2`), whereas `release-v0.18` uses the vendored source in `tests/gtest` (`v.1.7.0`) if gtest isn't installed as a system library, which doesn't define `INSTANTIATE_TEST_SUITE_P`.

Edit: I agree with selsta, removing the tests here is fine.

PR #10271, opened 2026-01-05
wallet: unrestrict `get_transfers` and `get_transfer_by_txid`

jeffro256:
Allow calling wallet RPC endpoints `get_transfers` and `get_transfer_by_txid` even with `--restricted-rpc` enabled. These two endpoints, after these changes, do not modify wallet state at all. The pool information may be outdated in restricted mode, though. This API change is breaking is the sense that some people may have relied on outgoing destination information being hidden in restricted mode. However, I will note that fetching ephemeral transaction private keys is NOT restricted, so if an adversary has access to a restricted RPC wallet, then they can already associate outgoing enotes with public Monero addresses, assuming that they already know of the public Monero address.

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -473,7 +473,7 @@ namespace tools
       return false;
   }
   //------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


0xdosto (approved):
LGTM

PR #1982, opened 2017-04-16
move txpool to database

moneromooo-monero:
I'll be using that on my mainnet node for a while to see if anything breaks, but looks fine on testnet + passes tests.

PR #10272, opened 2026-01-05
blockchain_db/rpc: faster `/is_key_image_spent`

jeffro256:
Does the following to speedup the `/is_key_image_spent` RPC endpoint:
  - Reads all on-chain key images in one LMDB read transacion
  - Uses `cryptonote_core::are_key_images_spent_in_pool()` instead of `cryptonote_core::get_pool_transactions_and_spent_keys_info()` for pool querying. This only does a LMDB read per key image if in the pool.
  - Filters known on-chain spent key images before querying for key images spent in pool

This RPC endpoint was causing major daemon slowdowns in the Carrot/FCMP++ Alpha Stressnet when using the `rescan_spent` command, especially for large wallets. The effect was much worse if the mempool was full.

Reported by @nahuhh.

Also removes Boost serialization for `tx_memory_pool` and `tx_memory_pool::tx_details`.

### DIFF
--- a/src/cryptonote_core/tx_pool.h
+++ b/src/cryptonote_core/tx_pool.h
@@ -37,7 +37,6 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <queue>
-#include <boost/serialization/version.hpp>
 #include <boost/utility.hpp>
 #include <boost/bimap.hpp>
 #include <boost/bimap/set_of.hpp>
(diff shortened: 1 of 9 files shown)


jeffro256:
This new code isn't in any consensus-related path AFAICT.

src/cryptonote_core/tx_pool.h
```diff
@@ -724,38 +724,3 @@ namespace cryptonote
     friend struct BlockchainAndPool;
   };
 }
-
-namespace boost
```
vtnerd, line 728: Is this just dead code?

jeffro256: Yup. Functionality was removed in #1982.

jeffro256: Removed a couple more lines relevant to txpool serialization in latest commit.

PR #10273, opened 2026-01-06
common: fix `apply_permutation()` for `std::vector<bool>`

jeffro256:
This PR is part of upstreaming Carrot


jeffro256:
See https://github.com/seraphis-migration/monero/pull/279 for use case.

PR #10276, opened 2026-01-11
ci: pin workflow actions to specific commits

RinZ27:
Pinned the GitHub Actions in `guix.yml`, `depends.yml`, and `build.yml` to specific commit hashes.

This ensures the CI/CD pipeline remains stable and secure against unexpected upstream tag updates or potential supply chain compromises. It's a standard practice to lock dependencies down.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -258,4 +258,3 @@ jobs:
         with:
           name: ${{ env.OUTPUT }}
           path: ${{ env.OUTPUT }}
-
--- a/.github/workflows/depends.yml
+++ b/.github/workflows/depends.yml
@@ -132,4 +132,3 @@ jobs:
         path: |
           build/${{ matrix.toolchain.host }}/release/bin/monero-wallet-cli*
           build/${{ matrix.toolchain.host }}/release/bin/monerod*
-
--- a/.github/workflows/guix.yml
+++ b/.github/workflows/guix.yml
@@ -106,4 +106,3 @@ jobs:
         with:
           name: "logs"
           path: '**/logs/**'
-


nahuhh:
CI failing and you dropped the EOF empty line

RinZ27:
@nahuhh Good catch. Restored the missing newlines at the end of the workflow files. The CI failure was due to a few invalid SHAs in the initial push, but it's sorted now.

nahuhh:
Now theres an extra newline

q. Why were these hashes chosen?

RinZ27:
Fixed the extra newline.

Regarding the hashes: I used the SHAs corresponding to the current `v4` tag for each action (as originally referenced in the workflow).
- `actions/checkout`: `34e1148` (refs/tags/v4)
- `actions/cache`: `0057852` (refs/tags/v4)
- `actions/upload-artifact`: `ea165f8` (refs/tags/v4)
- `actions/download-artifact`: `d3f86a1` (refs/tags/v4)

This pins the actions to their current state on the v4 branch.

selsta:
I'm not fully sure if this change makes sense, we don't use CI for any distributed binaries and pinning it to a tag allows us to get security updates without having to update the code. If we pin them to a commit they won't get updated until something breaks, which might overall be worse for security.

RinZ27:
Valid point regarding maintenance overhead. The main trade-off here is mitigating supply chain risks—if the `v4` tag is compromised upstream (even briefly), it could execute arbitrary code in the CI context.

Typically, projects pair SHA pinning with **Dependabot** (or Renovate) to automate the updates. This gives you the best of both worlds: immutable security guarantees + automated PRs for updates. If you're open to it, I can add a quick Dependabot config in a separate PR to handle the rotation.

nahuhh:
My immediate impression was/is the same as @selsta's — this seems unnecessary as we dont use CI to distribute binaries (the intent is to check merges and PRs for breakages, regressions, build errors and warnings). In the case of gitian/guix the binaries produced should match attestations from external builds.

using dependabot for bumping versions doesn't prevent upstream from addin malicious code. It still requires a user to manually review the upstream commits.

Some other projects use the CI to do the legwork for them, releasing their product directly from the files produced by the CI. I see how pinning makes sense there.

RinZ27:
That makes sense. I understand that since these builds aren't used for distribution, the immediate supply chain risk to end-users is minimal.

My main motivation was 'Defense in Depth' for the build environment itself (preventing potential secret exfiltration or lateral movement if an action were compromised), but I agree that without automated tooling like Dependabot (and the manual review it entails), the maintenance burden is higher.

If the consensus is that the maintenance cost outweighs the marginal security gain for this specific setup, I'm happy to close this PR.

RinZ27:
Closing this as the maintenance overhead of manual SHA rotation seems to outweigh the security benefits for this project's current CI usage. Thanks for the feedback!

PR #10278, opened 2026-01-12
p2p: connection patches

j-berman:
- Make sure the server sends a complete response when the client includes the "Connection: close" header.
- Make sure the server terminates in `m_strand` to avoid concurrent socket closure and ops processing.

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -173,7 +173,7 @@ namespace net_utils
       return;
     m_state.timers.general.wait_expire = true;
     auto self = connection<T>::shared_from_this();
-    m_timers.general.async_wait([this, self](const ec_t & ec){
+    auto on_wait = [this, self] {
       std::lock_guard<std::mutex> guard(m_state.lock);
       m_state.timers.general.wait_expire = false;
       if (m_state.timers.general.cancel_expire) {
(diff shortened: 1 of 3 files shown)


j-berman:
The failing test in CI sporadically fails before this PR. I've run the test locally tens of thousands of times on 2 machines, including an ubuntu 20.04 (same as CI), and it's passed every time. It's possible the code this PR touches *may* impact the test and cause it to fail more often, but without being able to repro, I think it's reasonable to proceed with assuming it's a borked test. It would be nice to investigate it further when time permits.

EDIT: I force pushed to re-run CI so tests would pass.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -191,6 +191,9 @@ namespace net_utils
         interrupt();
       else if (m_state.status == status_t::INTERRUPTED)
         terminate();
+    };
+    m_timers.general.async_wait([this, self, on_wait](const ec_t & ec){
```
vtnerd, line 195: Is `post` required here? This could also be `m_timers.general.async_wait(boost::bind_executor(m_strand, on_wait))` otherwise.

j-berman: Could you expand on the reasoning on why one would be preferred over the other here?

vtnerd: It's the standard way of doing and is less lines of code. So minimal reasons.

However, I just noticed on_wait doesn't take args so you basically have to do it this way or change its signature.

j-berman: Ah got it, alright if I keep it as is?

vtnerd: Yes.

PR #10282, opened 2026-01-21
rpc: return error gracefully on cryptonight v1 check

jeffro256:
A continuation of #9985: RPC endpoint `calc_pow` will return an error code and message instead of failing when using CryptoNight v1 on a block blob of size < 43. Fixes https://issues.oss-fuzz.com/issues/446998347.

Depends on #10039.


selsta:
@jeffro256 please rebase

PR #10285, opened 2026-01-21
blockchain_db: remove `{add/get}_max_block_size()`

jeffro256:
Fixes https://issues.oss-fuzz.com/issues/446735539. The max block size property field was not actually used in current code.


navidR (approved):
LGTM. Thanks.

Issue #9275, opened 2024-04-02
Default monerod configuration speedup


S5NC:
Why are the default monerod configurations very low?
Download is limited to 8 MiB/s, upload is limited to 2 MiB/s, and the maximum number of peers to download from (in) is limited to 12.

Could we increase the default limits to 1024 MiB/s download, 1024 MiB/s upload, 512 in_peers, and 512 out_peers?
Torrenting software typically don't have artificial transfer speed limits and they manage fine. We should too.

./monerod --help shows
```
  --out-peers arg (=-1)                 set max number of out peers
  --in-peers arg (=-1)                  set max number of in peers
  --tos-flag arg (=-1)                  set TOS flag
  --limit-rate-up arg (=2048)           set limit-rate-up [kB/s]
  --limit-rate-down arg (=8192)         set limit-rate-down [kB/s]
  --limit-rate arg (=-1)                set limit-rate [kB/s]
```

These limits are all what is by default, except the `--out-peers arg (=-1)`, which is actually 12. Maybe this was changed, or perhaps 12 is a fallback value from an incorrect configuration. If you pass `-1` as any of the arguments, it doesn't set it to unlimited but retains the default (which may be a fallback).

Maybe you would also want to increase the default maximum number of connections per ip from 1 to 32 or remove the restriction completely, as this impacts users with shared ips (same network, large business, VPN, cloud provider) when there are a large number of monero users on the same network.

selsta:
Increasing the number of out peers has an impact on privacy and network topology. We won't do any changes to it without corresponding research.

The bandwidth limits could be increased but as far as I know even during the recent spam attack the limit wasn't reached.

> Maybe you would also want to increase the default maximum number of connections per ip from 1 to 32 or remove the restriction completely

This makes it easier to sybil attack the network.

S5NC:
Increased 4x by #9770

PR #10291, opened 2026-01-25
Correct help messages' default number of daemon out peers

S5NC:
On daemon [initialization](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/p2p/net_node.inl#L2852) the number is of out peers is set to 12 when the `--out-peers` option is not specified, or specified as -1 by running `monerod --out-peers -1`.

This implies the [default](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/cryptonote_config.h#L139) is 12, not unlimited, which is what -1 means in other cases. For example, sending `out_peers -1` in the daemon console sets the number of out peers to unlimited.

Another set of limits are rate limits. Like the number of peers, [it](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/p2p/net_node.inl#L2927) [isn't](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/p2p/net_node.inl#L2940) [possible](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/p2p/net_node.inl#L2954) to set these to unlimited with options, yet the help messages specify [the](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/p2p/net_node.cpp#L165-L166) [practical](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/utils/fish/monerod.fish#L80) [defaults](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/utils/fish/monerod.fish#L81), not the number used internally (except for `--limit-rate`, but this option and command has unconventional behavior, setting a symmetric limit as opposed to limiting the total bandwidth).

Below is the current output of `monerod --help`:
```
  --out-peers arg (=-1)                 set max number of out peers
  --in-peers arg (=-1)                  set max number of in peers
  --tos-flag arg (=-1)                  set TOS flag
  --limit-rate-up arg (=8192)           set limit-rate-up [kB/s]
  --limit-rate-down arg (=32768)        set limit-rate-down [kB/s]
  --limit-rate arg (=-1)                set limit-rate [kB/s]
```

Addresses the remark made in #9275.

The inability to have an unlimited number of peers and bandwidth using daemon options could be purposeful. If so, it would be worth documenting this, and if not, further modifications could be made to allow this.

### DIFF
--- a/src/p2p/net_node.cpp
+++ b/src/p2p/net_node.cpp
@@ -158,7 +158,7 @@ namespace nodetool
     const command_line::arg_descriptor<std::string> arg_igd = {"igd", "UPnP port mapping (disabled, enabled, delayed)", "delayed"};
     const command_line::arg_descriptor<bool>        arg_p2p_use_ipv6  = {"p2p-use-ipv6", "Enable IPv6 for p2p", false};
     const command_line::arg_descriptor<bool>        arg_p2p_ignore_ipv4  = {"p2p-ignore-ipv4", "Ignore unsuccessful IPv4 bind for p2p", false};
-    const command_line::arg_descriptor<int64_t>     arg_out_peers = {"out-peers", "set max number of out peers", -1};
+    const command_line::arg_descriptor<int64_t>     arg_out_peers = {"out-peers", "set max number of out peers", P2P_DEFAULT_CONNECTIONS_COUNT};
     const command_line::arg_descriptor<int64_t>     arg_in_peers = {"in-peers", "set max number of in peers", -1};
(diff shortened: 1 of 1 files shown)


selsta:
Please squash and set a proper commit title.

S5NC:
I made these changes using the web editor so cannot edit history, but I will create a replacement PR with the git corrections made.

Issue #10122, opened 2025-09-28
Allow requesting blocks without their transactions in `get_blocks.bin`


kayabaNerve:
`get_blocks.bin` currently allows fetching blocks alone, the pool information alone, or both. When blocks are requested, the `block_complete_entry` is returned, including all transactions. It'd be _appreciated_ if I could request blocks _without_ their transactions, without having to use the JSON-RPC to request each block individually. Currently, I (within `monero-oxide`'s RPC code) do use the JSON-RPC to fetch blocks, hence why this would be appreciated (as it'd allow using a single request (as would #10118) _and_ reduce the size of the response by ~2x).

I'd be fine with either a new method or a new enum variant within `get_blocks.bin` for this.

miyajima-k:
#10295
 
Test in https://github.com/miyajima-k/monero/tree/10122-get-blocks-bin-wo-txblob-test

PR #10295, opened 2026-01-31
RPC `get_blocks.bin`: allow excluding txblob

miyajima-k:
- add new optional request field `get_tx_blobs`
- leave empty tx blobs in response if `get_tx_blobs` is true 
- 75-85% reduction in bandwidth (early blocks)

### DIFF
--- a/src/rpc/core_rpc_server_commands_defs.h
+++ b/src/rpc/core_rpc_server_commands_defs.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2024, The Monero Project
+// Copyright (c) 2014-2026, The Monero Project
 // 
 // All rights reserved.
 // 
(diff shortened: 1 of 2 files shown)


kayabaNerve:
This reduction of bandwidth is compared to when the transactions are present, right? That'd mean it's only a reduction in bandwidth for people who fall into the niche of using `get_blocks.bin` but not caring for the transactions.

That's a valid niche. It's the one I raised in #10122 😅 I just want to check what the reduction is with regards to.

src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -191,6 +191,7 @@ inline const std::string get_rpc_status(const bool trusted_daemon, const std::st
       bool        no_miner_tx;
       uint64_t    pool_info_since;
       uint64_t    max_block_count;
+      bool        get_tx_blobs;
```
jeffro256, line 194: You need to bump `CORE_RPC_VERSION_MINOR` when changing the API in a backwards-compatible manner.

PR #10296, opened 2026-01-31
wallet: skip identity points in decoy selection

jeffro256:
This skips the output for decoy selection if $O = 0$, $C = 0$, or $C = G$. $O = 0$ would cause the key image $L = 0$, which isn't valid by consensus rules (i.e. it is unspendable). If $C = 0$, or $C = G$, then it is known that the amount commitment binds to a 0 amount.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -8973,18 +8973,34 @@ bool wallet2::tx_add_fake_output(std::vector<std::vector<tools::wallet2::get_out
   if (std::find(outs.back().begin(), outs.back().end(), item) != outs.back().end()) // don't add duplicates
     return false;
   // check the keys are valid
-  if (valid_public_keys_cache.find(output_public_key) == valid_public_keys_cache.end() && !rct::isInMainSubgroup(rct::pk2rct(output_public_key)))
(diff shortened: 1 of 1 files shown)


kayabaNerve:
ACK to banning selecting unspendable keys as decoys.

NACK to changing the checks on the commitment. commit(r=0, v=0) and commit(r=1, a=0) are valid commitments and as statistically probable as any other uniformly-sampled randomness value. If you want to ban them for being trivial, also ban commit(r=2, a=0).

Also, commit(r=3, a=0).

Also, commit(r=4, a=0).

Also, ...

Also, I just really don't like commit(r=7205610562950251946294610472026491380, a=0). Can we ban that too?

jeffro256:
> commit(r=1, a=0) are valid commitments and as statistically probable as any other uniformly-sampled randomness value

Well, not really, since all coinbase outputs of amount zero are always `commit(r=1, a=0)`. 

> Also, I just really don't like commit(r=7205610562950251946294610472026491380, a=0). Can we ban that too?

Great, will add it to the list ;)

kayabaNerve:
Well, really, because your example is for a non-uniformly-sampled value when I distinguished to uniformly-sampled values.

I maintain we shouldn't ban  values you personally dislike because you dislike them. It's not a legitimate cryptographic process to argue our security via.

Issue #6683, opened 2020-06-22
Unexpected behavior on rate limit flags


MoneroArbo:
Expected behavior: Setting --limit-rate-down or --limit-rate-up to -1 makes the rate unlimited

Actual behavior: Setting --limit-rate-down or --limit-rate up sets bandwidth limits to the defaults of 2048 / 8196.

The value -1 mapping to the default, limited positive number values of 2048 & 8196 is inconsistent with how the  --in-peers and --out-peers flags work, not to mention inconsistent with other software where -1 typically represents the "unlimited" value.

PR #10302, opened 2026-02-04
Correct help messages' default number of daemon out peers

S5NC:
On daemon [initialization](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/p2p/net_node.inl#L2852) the number is of out peers is set to 12 when the `--out-peers` option is not specified, or specified as -1 by running `monerod --out-peers -1`.

This implies the [default](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/cryptonote_config.h#L139) is 12, not unlimited, which is what -1 means in other cases. For example, sending `out_peers -1` in the daemon console sets the number of out peers to unlimited.

Another set of limits are rate limits. Like the number of peers, [it](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/p2p/net_node.inl#L2927) [isn't](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/p2p/net_node.inl#L2940) [possible](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/p2p/net_node.inl#L2954) to set these to unlimited with options, yet the help messages specify [the](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/src/p2p/net_node.cpp#L165-L166) [practical](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/utils/fish/monerod.fish#L80) [defaults](https://github.com/monero-project/monero/blob/f65b2864552f855af1ef58c031dafffa58aae90c/utils/fish/monerod.fish#L81), not the number used internally (except for `--limit-rate`, but this option and command has unconventional behavior, setting a symmetric limit as opposed to limiting the total bandwidth).

Below is the current output of `monerod --help`:
```
  --out-peers arg (=-1)                 set max number of out peers
  --in-peers arg (=-1)                  set max number of in peers
  --tos-flag arg (=-1)                  set TOS flag
  --limit-rate-up arg (=8192)           set limit-rate-up [kB/s]
  --limit-rate-down arg (=32768)        set limit-rate-down [kB/s]
  --limit-rate arg (=-1)                set limit-rate [kB/s]
```

Addresses the remark made in #9275.

The inability to have an unlimited number of peers and bandwidth using daemon options could be purposeful. If so, it would be worth documenting this, and if not, further modifications could be made to allow this.

Commits formatted as requested per #10291.

### DIFF
--- a/src/p2p/net_node.cpp
+++ b/src/p2p/net_node.cpp
@@ -158,7 +158,7 @@ namespace nodetool
     const command_line::arg_descriptor<std::string> arg_igd = {"igd", "UPnP port mapping (disabled, enabled, delayed)", "delayed"};
     const command_line::arg_descriptor<bool>        arg_p2p_use_ipv6  = {"p2p-use-ipv6", "Enable IPv6 for p2p", false};
     const command_line::arg_descriptor<bool>        arg_p2p_ignore_ipv4  = {"p2p-ignore-ipv4", "Ignore unsuccessful IPv4 bind for p2p", false};
-    const command_line::arg_descriptor<int64_t>     arg_out_peers = {"out-peers", "set max number of out peers", -1};
+    const command_line::arg_descriptor<int64_t>     arg_out_peers = {"out-peers", "set max number of out peers", P2P_DEFAULT_CONNECTIONS_COUNT};
     const command_line::arg_descriptor<int64_t>     arg_in_peers = {"in-peers", "set max number of in peers", -1};
     const command_line::arg_descriptor<int> arg_tos_flag = {"tos-flag", "set TOS flag", -1};
 
--- a/utils/fish/monerod.fish
+++ b/utils/fish/monerod.fish
@@ -74,7 +74,7 @@ complete -c monerod -l no-sync -d "Don't synchronize the blockchain with other p
 complete -c monerod -l enable-dns-blocklist -d "Apply realtime blocklist from DNS"
 complete -c monerod -l no-igd -d "Disable UPnP port mapping"
 complete -c monerod -l igd -r -a "Enabled disabled enabled" -d "UPnP port mapping. Default: delayed"
-complete -c monerod -l out-peers -r -d "Set max number of out peers. Default: -1"
+complete -c monerod -l out-peers -r -d "Set max number of out peers. Default: 12"
 complete -c monerod -l in-peers -r -d "Set max number of in peers. Default: -1"
 complete -c monerod -l tos-flag -r -d "Set TOS flag. Default: -1"
 complete -c monerod -l limit-rate-up -r -d "Set limit-rate-up [kB/s]. Default: 8192"


S5NC:
> The inability to have an unlimited number of peers and bandwidth using daemon options could be purposeful.

I find the differing behavior of `-1` confusing. If I decide I want an unlimited number of out peers after experimenting in daemon console by running `out_peers -1`, I would except `--out-peers -1` as a daemon option to have the same effect.

One solution could be making `-1` mean unlimited as a option with flag omission loading the default, and in the console adding `default` to restore the default value (or by calling the set command without arguments). However, these additional changes would be breaking. Some node operators could be passing the meaningless `-1` argument as a daemon option, and I am sure they wouldn't appreciate an unexpected change.

I would opt for clarifying the current varying behavior of `-1` in the relevant documentation and help messages, but is there a documentation developer who knows how to convey this in the help messages without adding a large repetitive clump of text?

nahuhh:
-1 means default, not unlimite

default (-1) for in-peers is unlimited
default (-1) for out-peers is 12
so on an so forth

afaict, -1 never is intended to mean unlimited — there isnt a differing meaning among flags. It always means default (ie unchanged)

S5NC:
> -1 means default, not unlimited
> 
> default (-1) for in-peers is unlimited default (-1) for out-peers is 12 so on an so forth
> 
> afaict, -1 never is intended to mean unlimited — there isnt a differing meaning among flags. It always means default (ie unchanged)

So one issue is that the help message says -1 is the default value, but this isn't very informative, saying 12 is better, like `--limit-rate-up` and `--limit-rate-down` in the original post's message. Furthermore, actually passing -1 sets it to unlimited, not 12. This PR changes the help message.

And from your comments it seems that running `out_peers -1` in the console shouldn't set its amount to unlimited, unlike what currently happens as demonstrated by console feedback (and the actual number of connections reported by `status`).

```
out_peers
Max number of out peers set to 12

out_peers -1
Max number of out peers set to unlimited
```

Because of the above output and my previous experience using CLI tools, I was under the impression that `-1` meant unlimited. However, `limit_up -1`, `limit_down -1` both set their respective values to default, and `limit -1` sets both to their default values, not unlimited. At the end of the day, confusion in the console when setting values is a minor issue, as the assigned value is reported back to the user. Daemon options don't confirm their effect, but if the behaviour is working as intended there's no requirement for further changes.

S5NC:
Issue describing limit -1 confusion: https://github.com/monero-project/monero/issues/6683

nahuhh:
> So one issue is that the help message says -1 is the default value, but this isn't very informative, saying 12 is better, 

```
    const command_line::arg_descriptor<int64_t>     arg_out_peers = {"out-peers", "set max number of out peers", -1};
```

> Furthermore, actually passing -1 sets it to unlimited, not 12. This PR changes the help message.

i can confirn that interactively setting `out_peers -1` actually sets to unlimited.. :/

> And from your comments it seems that running `out_peers -1` in the console shouldn't set its amount to unlimited, unlike what currently happens as demonstrated by console feedback (and the actual number of connections reported by `status`).
> 
> ```
> out_peers
> Max number of out peers set to 12
> 
> out_peers -1
> Max number of out peers set to unlimited
> ```
> 
> Because of the above output and my previous experience using CLI tools, I was under the impression that `-1` meant unlimited. However, `limit_up -1`, `limit_down -1` both set their respective values to default, and `limit -1` sets both to their default values, not unlimited.

yep :/. The way the limit_* is set, is how i thought out_peers worked as well.
agreed that its a mess

PR #10303, opened 2026-02-04
p2p: Prevent inaccurate missing spans when adding a block

j-berman:
This prevents a data race where the `request_missing_objects` function ends up thinking a block span is missing when really the block(s) just got added to the db.

Solves the last known cause of "runaway spans" identified on the stressnet.

This PR is arguably not an ideal solution (it jams a lock into the center of 2 large functions and carries some risk of deadlock), but barring a major rewrite to the sync protocol, it seemed the best short/medium term improvement to this section to me.

https://github.com/seraphis-migration/monero/pull/275

### DIFF
--- a/src/cryptonote_protocol/cryptonote_protocol_handler.inl
+++ b/src/cryptonote_protocol/cryptonote_protocol_handler.inl
@@ -1489,8 +1489,16 @@ namespace cryptonote
             return 1;
           }
 
+          boost::unique_lock<boost::mutex> check_span_lock{m_check_span_queue_mutex, boost::defer_lock};
           bool stopped = false;
-          epee::unique_scope_guard cleanup_on_exit = [this, &stopped, &context, span_connection_id, start_height]() {
+          epee::unique_scope_guard cleanup_on_exit = [this, &check_span_lock, &stopped, &context, span_connection_id, start_height]() {
+            // Grab the span queue check lock so that we make sure we finish writing the block to the db before checking
+            // the span queue again. Otherwise it's possible for the span queue check to read the db height at n-1 in
+            // thread1, then wait for block n to be added and its span removed from the queue in this thread2, then check
+            // the span queue in thread1 and incorrectly think the span starting at block n is missing.
+            // TODO: a better sync protocol.
+            check_span_lock.lock();
+
             if (!m_core.cleanup_handle_incoming_blocks())
             {
               LOG_PRINT_CCONTEXT_L0("Failure in cleanup_handle_incoming_blocks");
@@ -1619,7 +1627,10 @@ namespace cryptonote
             MGINFO_YELLOW("Synced " << current_blockchain_height << "/" << target_blockchain_height
                 << progress_message << timing_message);
             if (previous_stripe != current_stripe)
+            {
+              check_span_lock.unlock();
               notify_new_stripe(context, current_stripe);
+            }
           }
         }
       }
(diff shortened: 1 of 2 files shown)


navidR (approved):
LGTM. Thank you

nahuhh:
cc @jeffro256 would be good to have this merged before fcmp++-stage rebases again

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1591,6 +1591,13 @@ namespace cryptonote
 
           MDEBUG(context << "Block process time (" << blocks.size() << " blocks, " << num_txs << " txs): " << block_process_time_full + transactions_process_time_full << " (" << transactions_process_time_full << "/" << block_process_time_full << ") ms");
 
+          // Grab the span queue check lock so that we make sure we finish writing the block to the db before checking
+          // the span queue again. Otherwise it's possible for the span queue check to read the db height at n-1 in
+          // thread1, then wait for block n to be added and its span removed from the queue in this thread2, then check
+          // the span queue in thread1 and incorrectly think the span starting at block n is missing.
+          // TODO: a better sync protocol.
+          boost::unique_lock<boost::mutex> check_span_lock{m_check_span_queue_mutex};
```
vtnerd: Just thinking out aloud here:

We may need to figure out how to leverage `boost::asio::strand` for block queuing. That way I/O threads are never blocked, and we have thread synchronization by default.

I'm stating this, because the lock will likely work, but hurts throughput a bit :/

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1591,6 +1591,13 @@ namespace cryptonote
...
+
           if (!m_core.cleanup_handle_incoming_blocks())
```
vtnerd: Does the other PR to make this via destructor cleanup cause issues?

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -1591,6 +1591,13 @@ namespace cryptonote
 
           MDEBUG(context << "Block process time (" << blocks.size() << " blocks, " << num_txs << " txs): " << block_process_time_full + transactions_process_time_full << " (" << transactions_process_time_full << "/" << block_process_time_full << ") ms");
 
+          // Grab the span queue check lock so that we make sure we finish writing the block to the db before checking
```
vtnerd: This was confusing at first - the DB write isn't complete/posted until `cleanup_handle_incoming_blocks()` ... ?

vtnerd (changes requested):
Just want a clarification on the comment - I understand what's going on here, but it's (still) somewhat surprising that this alleviates the runaway span issue.

vtnerd:
And another ping to @j-berman because there are now code conflicts, likely with the block cleanup calls.

src/cryptonote_protocol/cryptonote_protocol_handler.inl
j-berman: Should be fine, will just need to alter the code a bit

src/cryptonote_protocol/cryptonote_protocol_handler.inl
j-berman: Yep, that's where the db txn gets commited:

https://github.com/monero-project/monero/blob/def6fa207bf881f4db82489793cd5b79a7f3868d/src/cryptonote_core/blockchain.cpp#L4655

https://github.com/monero-project/monero/blob/def6fa207bf881f4db82489793cd5b79a7f3868d/src/blockchain_db/lmdb/db_lmdb.cpp#L3901

src/cryptonote_protocol/cryptonote_protocol_handler.inl
j-berman: Resolved in the latest. It actually is improved now, since I'm making sure to grab the lock in those other early return places that the cleanup could get called.

For reference, here's the OG commit: https://github.com/j-berman/monero/commit/7d0323025152e4111220c1e4eecd36ff2c68347c

selsta (approved):
Tested on stressnet.

thomasbuilds (approved):

jpk68 (approved):

nahuhh:
@selsta @tobtoht can we get v0.19 backport

Issue #9231, opened 2024-03-10
CLI wallet terminal lockscreen handler needs updating


tidux:
Running the inside `tmux` results in garbled output if `TERM` is set to `tmux-256color` or `screen-256color`, both of which are valid on my system for various TUI applications including Emacs, Vim, and GNU Midnight Commander.  Running with the override `$ TERM=xterm-256color monero-wallet-cli` produces expected behavior.

omurad:
I was not able to reproduce this. Is there specific command that resulted in bad output?

```
user@devbox-vm:~/monero-x86_64-linux-gnu-v0.18.4.0$ uname -a
Linux devbox-vm 6.1.0-37-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.140-1 (2025-05-22) x86_64 GNU/Linux
user@devbox-vm:~/monero-x86_64-linux-gnu-v0.18.4.0$ echo $TERM
tmux-256color
user@devbox-vm:~/monero-x86_64-linux-gnu-v0.18.4.0$ ./monero-wallet-cli --daemon-address xmr-node.cakewallet.com:18081
This is the command line monero wallet. It needs to connect to a monero
daemon to work correctly.
WARNING: Do not reuse your Monero keys on another fork, UNLESS this fork has key reuse mitigations built in. Doing so will harm your privacy.

Monero 'Fluorine Fermi' (v0.18.4.0-release)
Logging to ./monero-wallet-cli.log
Specify wallet file name (e.g., MyWallet). If the wallet doesn't exist, it will be created.
Wallet file name (or Ctrl-C to quit): <redacted>
Wallet and key files found, loading...
Wallet password:
Opened wallet: <redacted>
**********************************************************************
Use the "help" command to see a simplified list of available commands.
Use "help all" to see the list of all available commands.
Use "help <command>" to see a command's documentation.
**********************************************************************
Warning: using an untrusted daemon at xmr-node.cakewallet.com:18081
Using a third party daemon can be detrimental to your security and privacy
You are strongly encouraged to connect to the Monero network using your own daemon
If you or someone you trust are operating this daemon, you can use --trusted-daemon
Background mining not enabled. Run "set setup-background-mining 1" to change.
Starting refresh...
Height 87913 / 3455822
```

PR #10305, opened 2026-02-04
Avoid full terminal reset in tmux/screen

Yaroslav98214:
## Summary
- Avoid full terminal reset when running under tmux/screen
- Prevents garbled output during wallet lockscreen clear

## Related issue
- Related to monero-project/monero#9231


nahuhh:
does the related issue still require the workaround on v0.18.4.5?

is this.. a bot?

PR #10312, opened 2026-02-09
Misc clang 21 fixes

jeffro256:
* Use -O3 and other flags instead of -Ofast
  - https://discourse.llvm.org/t/rfc-deprecate-ofast/78687
  - https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html
* Find libunwind library based on C++ compiler type, not C compiler type
* In stack_trace.cpp, pass stream modifiers to `std::stringstream` first, then send string to log
* In stack_trace.cpp, remove dead code related to `stack_trace_log` path
* Remove `virtual` method attributues from `final` class `cryptonote::core`
* Remove unused `this` capture in cryptonote protocol handler
* Remove unused variable `bad` in net node
* Use `std::make_unsigned` instead of `boost::make_unsigned`
  - https://github.com/boostorg/type_traits/issues/171
  - https://github.com/boostorg/type_traits/issues/202
  - https://github.com/boostorg/type_traits/pull/199
* Cleanup `include`s in pair serialization
* Test convergence b/t `std::make_unsigned` and `boost::make_unsigned`

Fixes compilation and silences warnings on:
clang version 21.1.6
Linux 6.18.8-3-cachyos

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -351,7 +351,7 @@ endif()
 if(WIN32 OR ARM OR PPC64LE OR PPC64 OR PPC)
   set(OPT_FLAGS_RELEASE "-O2")
 else()
-  set(OPT_FLAGS_RELEASE "-Ofast")
+  set(OPT_FLAGS_RELEASE "-O3 -ffast-math -fno-semantic-interposition") # not present: -fallow-store-data-races
 endif()
 
 # BUILD_TAG is used to select the build type to check for a new version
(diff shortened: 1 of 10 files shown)


jeffro256:
Fixed static assertions for systems where `uint64_t` and `int64_t` are defined as `unsigned long long` and `long long`, respectively, even if types `unsigned long` and `long` can fit 64 bits.

jeffro256:
Errors in latest CI are unrelated. There appears to be an issue with Github's backend at the moment; git fetching is failing.

navidR:
Can you just do a push (`--amend --no-edit`)? So that CI re-runs again. 

It seems CI is working right now

navidR (approved):

PR #10313, opened 2026-02-09
simplewallet: add confirmation prompt for default soft rescan_bc

navidR:


jeffro256:
What's the argument for adding this?

nahuhh:
> What's the argument for adding this?

someone requested it on matrix

rudolfschmidt (commented):
The scope of the new confirmation prompt seems narrow enough that it may be worth clarifying the intended UX: the prompt fires for `rescan_bc` (defaulting to soft) but not for `rescan_bc soft`, even though both perform the same operation. Is that intentional? Either the prompt should apply to both, or neither, to keep the behavior consistent.

The prompt wording itself (`"Rescan the blockchain from the wallet's restore height?"`) restates what the command does without communicating *why* the user might want to reconsider. Since a soft rescan isn't destructive, I'd expect the message to focus on cost — e.g. *"This can take a significant amount of time. Continue?"*. Monero's existing confirmation prompts (hard rescan, restore-height mismatch) all warn about a real consequence.

Also echoing @jeffro256 — could you add a short PR description with the linked request or issue? An empty body for behavior changes makes future archaeology harder.

PR #10320, opened 2026-02-14
Transition asio::deadline_timer to asio::steady_timer

vtnerd:
As of Boost 1.77, `asio::deadline_timer` has been deprecated. The build log gets spammed with warnings about it. The deprecating timer will likely be around for a while, so this isn't a high priority PR right now.

The transition to `asio::steady_timer` pushes more usage of `std::chrono` to help with type-checking. Usages of `boost::shared_ptr` are updated to `std::shared_ptr` too, when the line is already being touched to update the timer type.

### DIFF
--- a/src/device_trezor/trezor/transport.hpp
+++ b/src/device_trezor/trezor/transport.hpp
@@ -32,7 +32,7 @@
 
 
 #include <boost/asio.hpp>
-#include <boost/asio/deadline_timer.hpp>
+#include <boost/asio/steady_timer.hpp>
 #include <boost/array.hpp>
 #include <boost/utility/string_ref.hpp>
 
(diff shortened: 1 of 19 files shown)


src/device_trezor/trezor/transport.hpp
```diff
@@ -233,19 +233,19 @@ namespace trezor {
 
   private:
     void require_socket();
-    ssize_t receive(void * buff, size_t size, boost::system::error_code * error_code=nullptr, bool no_throw=false, boost::posix_time::time_duration timeout=boost::posix_time::seconds(10));
+    ssize_t receive(void * buff, size_t size, boost::system::error_code * error_code=nullptr, bool no_throw=false, boost::asio::steady_timer::duration timeout = std::chrono::seconds(10));
     void check_deadline();
     static void handle_receive(const boost::system::error_code& ec, std::size_t length,
                                boost::system::error_code* out_ec, std::size_t* out_length);
-    bool ping_int(boost::posix_time::time_duration timeout=boost::posix_time::milliseconds(1500));
+    bool ping_int(boost::asio::steady_timer::duration =std::chrono::milliseconds(1500));
```
iamamyth: Formatting: Probably want spaces on both sides of the equals, or no spaces, rather than a space on only one side.

iamamyth (approved):
Spotted a formatting inconsistency which hardly matters, test failure also appears unrelated to this PR, so it has my approval.

src/device_trezor/trezor/transport.hpp
vtnerd: Fixed.

PR #10323, opened 2026-02-15
makefile: fix parallel build

navidR:


nahuhh:
cc @tobtoht

navidR:
I don't know what was the rationale here [1] to change `(MAKE)` to `cmake --build . ` . 

The problem with that change is quite hard to capture `-j` and pass it to `cmake` in a correct way, unless a new variable  introduced. Something like: 

```
⁦make release           # uses nproc automatically
make release JOBS=1    # serial
make release JOBS=24   # explicit count⁦   
```

This reverts to what we had previously, `make release -jN` should work without issues passing `N` to build.

@tobtoht in case [1] is required for something else this PR should not be approved. 

1. https://github.com/monero-project/monero/pull/9538

tobtoht (approved):

Issue #10164, opened 2025-10-13
[request] add "all" parameter to `unban`


nahuhh:
Hard to quickly clear bans

PR #10324, opened 2026-02-15
Implement 'unban all' RPC command.

omurad:
Implemented `unban all` RPC command + test.

#### CLI

```
2026-02-15 19:38:01.313 I **********************************************************************
2026-02-15 19:38:01.313 I You are now synchronized with the network. You may now start monero-wallet-cli.
2026-02-15 19:38:01.313 I
2026-02-15 19:38:01.313 I Use the "help" command to see the list of available commands.
2026-02-15 19:38:01.313 I **********************************************************************

bans
No IPs are banned

ban 8.8.8.8
2026-02-15 19:39:09.926 I Host 8.8.8.8 blocked.
ban 8.8.4.4
2026-02-15 19:39:14.692 I Host 8.8.4.4 blocked.

unban all
2026-02-15 19:39:21.099 I All blocked hosts and subnets unblocked.
All bans have been cleared.

bans
No IPs are banned
```

#### JSON-RPC

```
$ ~/monero % curl https://localhost:18081/json_rpc -H "Content-Type: application/json" -d '{
    "jsonrpc":"2.0",
    "id":"0",
    "method":"get_bans",
}'
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "bans": [{
      "host": "8.8.8.8",
      "ip": 134744072,
      "seconds": 86393
    }],
    "status": "OK",
    "untrusted": false
  }
}%                                                                                                                                                                                          

$ ~/monero % curl https://localhost:18081/json_rpc -H "Content-Type: application/json" -d '{
    "jsonrpc":"2.0",
    "id":"0",
    "method":"unban_all",
}'
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "status": "OK",
    "untrusted": false
  }
}%                                                                                                                                                                                          

$ ~/monero % curl https://localhost:18081/json_rpc -H "Content-Type: application/json" -d '{
    "jsonrpc":"2.0",
    "id":"0",
    "method":"get_bans",
}'
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "status": "OK",
    "untrusted": false
  }
}%
```

### DIFF
--- a/src/daemon/command_parser_executor.cpp
+++ b/src/daemon/command_parser_executor.cpp
@@ -743,7 +743,7 @@ bool t_command_parser_executor::unban(const std::vector<std::string>& args)
   }
 
   if (args[0] == "all")
-    return m_executor.unban_all();
+    return m_executor.clear_bans();
 
   std::string ip = args[0];
   return m_executor.unban(ip);
--- a/src/daemon/rpc_command_executor.cpp
+++ b/src/daemon/rpc_command_executor.cpp
@@ -1788,23 +1788,23 @@ bool t_rpc_command_executor::unban(const std::string &address)
     return true;
 }
 
-bool t_rpc_command_executor::unban_all()
+bool t_rpc_command_executor::clear_bans()
 {
-    cryptonote::COMMAND_RPC_UNBAN_ALL::request req;
-    cryptonote::COMMAND_RPC_UNBAN_ALL::response res;
+    cryptonote::COMMAND_RPC_CLEAR_BANS::request req;
+    cryptonote::COMMAND_RPC_CLEAR_BANS::response res;
     std::string fail_message = "Unsuccessful";
     epee::json_rpc::error error_resp;
 
     if (m_is_rpc)
     {
-        if (!m_rpc_client->json_rpc_request(req, res, "unban_all", fail_message.c_str()))
+        if (!m_rpc_client->json_rpc_request(req, res, "clear_bans", fail_message.c_str()))
         {
             return true;
         }
     }
     else
     {
-        if (!m_rpc_server->on_unban_all(req, res, error_resp) || res.status != CORE_RPC_STATUS_OK)
+        if (!m_rpc_server->on_clear_bans(req, res, error_resp) || res.status != CORE_RPC_STATUS_OK)
         {
             tools::fail_msg_writer() << make_error(fail_message, res.status);
             return true;
--- a/src/daemon/rpc_command_executor.h
+++ b/src/daemon/rpc_command_executor.h
@@ -136,7 +136,7 @@ public:
 
   bool unban(const std::string &address);
 
-  bool unban_all();
+  bool clear_bans();
 
   bool banned(const std::string &address);
 
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -2945,9 +2945,9 @@ namespace cryptonote
     return true;
   }
   //------------------------------------------------------------------------------------------------------------------------------
-  bool core_rpc_server::on_unban_all(const COMMAND_RPC_UNBAN_ALL::request& req, COMMAND_RPC_UNBAN_ALL::response& res, epee::json_rpc::error& error_resp, const connection_context *ctx)
+  bool core_rpc_server::on_clear_bans(const COMMAND_RPC_CLEAR_BANS::request& req, COMMAND_RPC_CLEAR_BANS::response& res, epee::json_rpc::error& error_resp, const connection_context *ctx)
   {
-    RPC_TRACKER(unban_all);
+    RPC_TRACKER(clear_bans);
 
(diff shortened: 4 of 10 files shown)


omurad:
Implements #10164

nahuhh:
for rpc: I would probably prefer an rpc method of `unban`, that takes params in array format and accepts `all` as an argument

```
"params":[{"address":"127.0.0.1","192.168.1.125"}]
```

or

```
"params":[{"address":"all"}]
```

for cli: LGTM


edit: `unban` method already exists. Need to look into it, but if its not already an array, its debatable whether it should be changed. Might just have to add "all" as a valid argument and not change it to an array. 

@selsta  thoughts? Docs doesnt have ban/unban documented (but someone could be using them in the wild)

omurad:
There is no existing `unban` RPC method. It seems @nahuhh's suggestion should be implemented as part of the `set_bans` method.

e.g.

```
$ curl https://127.0.0.1:18081/json_rpc \
  -d '{
    "jsonrpc":"2.0",
    "id":"0",
    "method":"set_bans",
    "params":{
      "bans":[
        {
          "host":"*",
          "ip":0,
          "ban":false,
          "seconds":0
        }
      ]
    }
  }' \
  -H 'Content-Type: application/json'
```

nahuhh:
ip and seconds are optional, so it can just be 
```
"params":{"bans":[{"host":"*","ban":false}]}
```

omurad:
> for rpc: I would probably prefer an rpc method of `unban`, that takes params in array format and accepts `all` as an argument
> 
> ```
> "params":[{"address":"127.0.0.1","192.168.1.125"}]
> ```

It seems `set_bans` already implements this functionality.

Any thoughts on keeping the new RPC method vs. extending `set_bans`?
Maybe rename `unban_all` to `clear_bans`

This seems like a cleaner implementation to me as it keeps the "delete all" functionality in an atomic command, as opposed to being (potentially) one of many ban list CRUD operations in `set_bans`'s `bans` array.

So in CRUD-terms, the ban list RPC interface would be:
`banned`: read
`set_bans`: write, update, delete
`get_bans`: read all
`clear_bans`: delete all

nahuhh:
That's probably fair. Do that (rename) for now, and we'll see what others think

iamamyth:
It seems the real problem here is that `(block|unblock)_subnet` in net_node.inl sometimes fakes subnet support: The banned subnet list itself performs no set arithmetic, so you can have a subnet, and subsections of that subnet, as distinct entries in the same ban list, but banning a subnet will prune any peer contained in the subnet (as it should). With proper subnet support, you could just ban/unban /0 and be done with it, no need for a dedicated endpoint. So, this PR just papers over the flaw for a single case, rather than fixing it: Does it really make sense that unbanning, say, /16 would leave /24 banned?

iamamyth:
Also: I wrote "blocked subnet list", because it should be an ordered data structure, but it's implemented using a map. With proper subnet folding, you can pick a better data structure, with more desirable properties (e.g. lookup via binary search, expanding a subnet in constant time, etc).

omurad:
I agree that subnet ban handling is flawed.
I guess the question for this PR now is, regardless if subnet handling gets fixed, should monerod expose a single-purpose `clear_bans` RPC method? 

```
method: clear_bans
```

Or should `set_bans` be the only way to clear all bans?

```
method: set_bans
params:
    /* bans processed sequentially, stops processing on first failure */
    bans: [
        /* allows other ban entries here */
        { ip: 0.0.0.0/0, ban: false},
        /* allows other ban entries here */
    ]
```

iamamyth:
If subnet handling were implemented properly, there'd be no need for clear_bans (technically, the current RPC supports undoing all bans; the difficulty of executing that task owes to the poor subnet support). Providing multiple routes to effect the same action in an RPC tends to complicate both documentation and maintenance and give rise to split behavior (one can see the seeds of such problems in the draft code, though I do not mean to suggest it's broken).

Regarding the possibility of unbanning and banning in the same call, I see no reason that's a bad idea; for example, it allows the caller to ensure the ban list consists of only a specific set of subnets. The fact that modifications prior to a complete unban don't make a lot of sense, while true, seems immaterial (allowing silly things isn't the same as recommending them, e.g. if someone buys a roll of tape and promptly throws it in the garbage, that's not the manufacturer's concern).

omurad:
Makes sense. Closing this PR. I have created #10325 to track this issue.

iamamyth:
@omurad I think there's a fairly trivial way to satisfy this feature request while addressing my concerns: Modify the existing endpoint to check for `/0` as a special case which clears the banlist and then, if `ban:true`, adds just the one item (in addition to, of course, disconnecting peers, as the existing code would). As a nice bonus, this approach makes it trivial to temporarily disable p2p traffic.

iamamyth:
One clarification: By `/0`, I don't really care whether you check for `0.0.0.0/0` or any `/0` netmask, it's your decision.

PR #10330, opened 2026-02-22
cryptonote_core: remove Boost serialization for `tx_source_entry`

jeffro256:


vtnerd (approved):

SNeedlewoods:
Just wondering is Boost serialization still needed for `tx_destination_entry`?

jeffro256:
@SNeedlewoods  Yes AFAIK. It is used in serialization for `m_confirmed_txs` and `m_unconfirmed_txs` fields in the wallet cache. As long as we support backward compatible Boost serialization for the old wallet formats, we will need that code. Though at some point, it should be hoisted down to the [`wallet2_basic`](https://github.com/monero-project/monero/pull/10084) library IMO. If it pushed down to the wallet libraries, then Boost Serialization and Boost Archive won't be dependencies of `cryptonote_core`.

PR #10334, opened 2026-02-23
tests: add key image spent status lifecycle test

DUQUEredes:
## Summary

- Add standalone functional test `is_key_image_spent.py` covering the full lifecycle of key image spent status codes
- The existing test in `transfer.py` only covers status 0 (UNSPENT) and 1 (SPENT_IN_BLOCKCHAIN); **status 2 (SPENT_IN_POOL) is never tested**
- This test verifies all three status codes and their transitions: pool→blockchain, pool→flush→unspent, and mixed-state batch queries
- Added to `DEFAULT_TESTS` in `functional_tests_rpc.py` so it runs in CI

## Test cases

| Test | What it proves |
|------|---------------|
| `check_unspent` | Wallet key images return status 0 before any spending |
| `check_spent_in_pool` | Key images in mempool return status 2 |
| `check_spent_in_blockchain` | After mining, status transitions 2 → 1 |
| `check_flush_returns_to_unspent` | After flush_txpool, status reverts 2 → 0 |
| `check_mixed_states` | Single batch query with all 3 statuses returns correct per-key-image results |
| `check_edge_cases` | Empty list, unknown key images, duplicate key images in one call |

## Test plan

- [x] All assertions pass against monerod v0.18.1.0 in regtest mode
- [x] Status 2 (SPENT_IN_POOL) correctly observed
- [x] Added to `functional_tests_rpc.py` DEFAULT_TESTS
- [ ] Reviewer runs `PYTHONPATH=utils/python-rpc python3 tests/functional_tests/is_key_image_spent.py` with monerod + wallet-rpc in regtest

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/functional_tests/functional_tests_rpc.py
+++ b/tests/functional_tests/functional_tests_rpc.py
@@ -12,7 +12,7 @@ import time
 USAGE = 'usage: functional_tests_rpc.py <python> <srcdir> <builddir> [<tests-to-run> | all]'
 DEFAULT_TESTS = [
   'address_book', 'bans', 'blockchain', 'cold_signing', 'daemon_info', 'get_output_distribution',
-  'http_digest_auth', 'integrated_address', 'k_anonymity', 'mining', 'multisig', 'p2p', 'proofs',
+  'http_digest_auth', 'integrated_address', 'is_key_image_spent', 'k_anonymity', 'mining', 'multisig', 'p2p', 'proofs',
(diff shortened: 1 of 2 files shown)


selsta:
>As for tests: don't accept new test just for the sake of tests. New tests should be added together with an entirely new code, or together with bugfixes - i.e. a test that makes the old code fail + the code fix

Please only submit PRs that are clear bug or logic fixes. AI generated pull requests that add vague test cases with no clear benefit, or other changes that are not done with clear purpose will get closed.

PR #10337, opened 2026-02-24
cryptonote_basic: fix add_extra_nonce_to_tx_extra() length [RELEASE]

jeffro256:
#10220


selsta:
This fails to compile.

PR #10338, opened 2026-02-25
crypto: derive key image generator & separate {un}biased hash to ec

j-berman:
In https://github.com/monero-project/research-lab/issues/142 , @kayabaNerve noted the potential bias in Monero's hash-to-point function, and proposed introducing a cleaner *unbiased* hash-to-point function with the FCMP++ upgrade to maximize bit security. This proposal was discussed further in [this MRL meeting](https://github.com/monero-project/meta/issues/1261), where @kayabaNerve provided [additional context](https://github.com/monero-project/research-lab/issues/142#issuecomment-3956284024) for why this property is even more desirable with FCMP++/Carrot to avoid the ["burning bug"](https://web.getmonero.org/2018/09/25/a-post-mortum-of-the-burning-bug.html) in potential circumstances specific to FCMP++/Carrot and not RingCT.

With that general context in mind, this PR introduces 3 new core crypto functions:

- `biased_hash_to_ec` (previously `hash_to_ec`)
  - This change makes it clear that legacy code uses the "biased" implementation, which should help clearly separate it from the upgraded "unbiased" implementation and help avoid accidental usage for new code. 
- `unbiased_hash_to_ec`
  - Used to generate key images in FCMP++ / Carrot code (current iteration needs [this PR](https://github.com/monero-oxide/monero-oxide/pull/182) also UPDATE: that PR was merged).
- `derive_key_image_generator`
  - This new helper function is used in FCMP++ code for generating `I`, which is a member of a leaf that enters the FCMP++ "curve tree," and is used to derive the key image (key images are `x * I`, where x is the one-time priv key used to spend an output). See references to `I` in [the FCMP++ paper](https://github.com/kayabaNerve/fcmp-plus-plus-paper/blob/develop/fcmp%2B%2B.pdf) ([specific commit](https://github.com/kayabaNerve/fcmp-plus-plus-paper/blob/6f01fd3296b17f90b5c97c738cc3320e3369ad49/fcmp%2B%2B.pdf)).
  ______

Cypher Stack and Trail of bits audited an earlier state of this PR.

Cypher Stack recommended using a domain-separating tag for our hash to point in [Section 9.3.3](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf). Thus, I implemented and used the same `blake2_monero` used for SAL from [this PR](https://github.com/monero-oxide/monero-oxide/pull/176) by @UkoeHB. The Rust side needs [this PR](https://github.com/monero-oxide/monero-oxide/pull/182) to match, which also implements the same tests as this current PR (@UkoeHB suggested to include hard-coded tests consistent on the C++ side and Rust side).

Both Cypher Stack and Trail of Bits affirmed the rationale for this PR, and the implementation of the unbiased hash-to-point (neither reviewed the code that includes the DST from `blake2_monero`). I will share the Trail of Bits audit once public.

### DIFF
--- a/src/crypto/blake2b.c
+++ b/src/crypto/blake2b.c
@@ -36,6 +36,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "memwipe.h"
 
+#include <assert.h>
 #include <stdint.h>
 #include <string.h>
 #include <stdio.h>
@@ -280,9 +281,12 @@ int blake2b_init_key(blake2b_state *S, size_t outlen, const void *key, size_t ke
 		uint8_t block[BLAKE2B_BLOCKBYTES];
 		memset(block, 0, BLAKE2B_BLOCKBYTES);
 		memcpy(block, key, keylen);
-		blake2b_update(S, block, BLAKE2B_BLOCKBYTES);
+		int r = blake2b_update(S, block, BLAKE2B_BLOCKBYTES);
 		/* Burn the key from stack */
 		clear_internal_memory(block, BLAKE2B_BLOCKBYTES);
+		if (r < 0) {
+			return -1;
+		}
 	}
 	return 0;
 }
@@ -514,5 +518,74 @@ fail:
 #undef TRY
 }
 /* Argon2 Team - End Code */
+
+int blake2b_monero(void *out, size_t outlen, const void *in, size_t inlen,
+	const void *key, size_t keylen) {
+	// At this time, we expect all Monero-based callers to use a keylen=32
+	// if a key is included. If a caller wants to use a key and keylen != 32,
+	// just remove this assert (and #include <assert.h>).
+	assert(keylen == 0 || keylen == 32);
+
+	static const uint8_t PERSONAL[BLAKE2B_PERSONALBYTES] = {'M', 'o', 'n', 'e', 'r', 'o'};
+
+	blake2b_param P;
+	blake2b_state S;
+	int ret = -1;
+
+	/* Verify parameters */
+	if (NULL == in && inlen > 0) {
+		goto fail;
+	}
+
+	if (NULL == out || outlen == 0 || outlen > BLAKE2B_OUTBYTES) {
+		goto fail;
+	}
+
+	if ((NULL == key && keylen > 0) || keylen > BLAKE2B_KEYBYTES) {
+		goto fail;
+	}
+
+	if (NULL != key && keylen == 0) {
+		goto fail;
+	}
+
+	/* Setup Parameter Block */
+	P.digest_length = (uint8_t)outlen;
+	P.key_length = (uint8_t)keylen;
+	P.fanout = 1;
+	P.depth = 1;
+	P.leaf_length = 0;
+	P.node_offset = 0;
+	P.node_depth = 0;
+	P.inner_length = 0;
+	memset(P.reserved, 0, sizeof(P.reserved));
+	memset(P.salt, 0, sizeof(P.salt));
+	memcpy(P.personal, PERSONAL, sizeof(PERSONAL));
+
+	if (blake2b_init_param(&S, &P) < 0) {
+		goto fail;
+	}
+
+	if (NULL != key) {
+		uint8_t block[BLAKE2B_BLOCKBYTES];
+		memset(block, 0, BLAKE2B_BLOCKBYTES);
+		memcpy(block, key, keylen);
+		int r = blake2b_update(&S, block, BLAKE2B_BLOCKBYTES);
+		/* Burn the key from stack */
+		clear_internal_memory(block, BLAKE2B_BLOCKBYTES);
+		if (r < 0) {
+			goto fail;
+		}
+	}
+
+	if (blake2b_update(&S, in, inlen) < 0) {
+		goto fail;
+	}
+	ret = blake2b_final(&S, out, outlen);
+
+fail:
+	clear_internal_memory(&S, sizeof(S));
+	return ret;
+}
 
 /// END: blake2b.c
--- a/src/crypto/blake2b.h
+++ b/src/crypto/blake2b.h
@@ -111,6 +111,9 @@ extern "C" {
 	int blake2b_long(void *out, size_t outlen, const void *in, size_t inlen);
 	/* Argon2 Team - End Code */
 
+	int blake2b_monero(void *out, size_t outlen, const void *in, size_t inlen,
+		const void *key, size_t keylen);
+
 #if defined(__cplusplus)
 }
 #endif
--- a/src/crypto/crypto.cpp
+++ b/src/crypto/crypto.cpp
@@ -42,6 +42,7 @@
 #include "warnings.h"
 #include "crypto.h"
 #include "hash.h"
+#include "blake2b.h"
 
 #include "cryptonote_config.h"
 
@@ -608,7 +609,7 @@ namespace crypto {
     return sc_isnonzero(&c2) == 0;
   }
 
-  static void hash_to_ec(const public_key &key, ge_p3 &res) {
+  static void biased_hash_to_ec(const public_key &key, ge_p3 &res) {
     hash h;
     ge_p2 point;
     ge_p1p1 point2;
@@ -618,11 +619,58 @@ namespace crypto {
     ge_p1p1_to_p3(&res, &point2);
   }
 
+  void crypto_ops::unbiased_hash_to_ec(const unsigned char *preimage, const std::size_t length, ec_point &res) {
+    unsigned char hash[64];
+    if (blake2b_monero(std::addressof(hash), 64, preimage, length, NULL, 0) != 0)
+      throw std::runtime_error("Failed blake2b");
+
+    ge_p2 first;
+    ge_fromfe_frombytes_vartime(&first, hash);
+    ge_p1p1 first_p1p1;
+    ge_mul8(&first_p1p1, &first);
+    ge_p3 first_p3;
+    ge_p1p1_to_p3(&first_p3, &first_p1p1);
+
+    ge_p2 second;
+    ge_fromfe_frombytes_vartime(&second, hash + 32);
+    ge_p1p1 second_p1p1;
+    ge_mul8(&second_p1p1, &second);
+    ge_p3 second_p3;
+    ge_p1p1_to_p3(&second_p3, &second_p1p1);
+    ge_cached second_cached;
+    ge_p3_to_cached(&second_cached, &second_p3);
+
+    ge_p1p1 point;
+    ge_add(&point, &first_p3, &second_cached);
+
+    ge_p3 res_ge_p3;
+    ge_p1p1_to_p3(&res_ge_p3, &point);
+    ge_p3_tobytes(&res, &res_ge_p3);
+  }
+
+  static void biased_derive_key_image_generator(const public_key &pub, ec_point &ki_gen) {
+    ge_p3 point;
+    biased_hash_to_ec(pub, point);
+    ge_p3_tobytes(&ki_gen, &point);
+  }
+
+  static void unbiased_derive_key_image_generator(const public_key &pub, ec_point &ki_gen) {
+    unbiased_hash_to_ec(&pub, sizeof(public_key), ki_gen);
+  }
+
+  void crypto_ops::derive_key_image_generator(const public_key &pub, const bool biased, ec_point &ki_gen) {
+    if (biased)
+      biased_derive_key_image_generator(pub, ki_gen);
+    else
+      unbiased_derive_key_image_generator(pub, ki_gen);
+  }
+
+  // key image derivation used before FCMP++/Carrot
   void crypto_ops::generate_key_image(const public_key &pub, const secret_key &sec, key_image &image) {
     ge_p3 point;
     ge_p2 point2;
     assert(sc_check(&sec) == 0);
-    hash_to_ec(pub, point);
+    biased_hash_to_ec(pub, point);
     ge_scalarmult(&point2, &unwrap(sec), &point);
     ge_tobytes(&image, &point2);
   }
@@ -683,7 +731,7 @@ POP_WARNINGS
         random_scalar(k);
         ge_scalarmult_base(&tmp3, &k);
         ge_p3_tobytes(&buf->ab[i].a, &tmp3);
-        hash_to_ec(*pubs[i], tmp3);
+        biased_hash_to_ec(*pubs[i], tmp3);
         ge_scalarmult(&tmp2, &k, &tmp3);
         ge_tobytes(&buf->ab[i].b, &tmp2);
       } else {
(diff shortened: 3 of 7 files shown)


kayabaNerve:
Used as `I` for leaves for CARROT outputs and CARROT outputs alone, right? The above is ambiguous if it's used generally within FCMP++, including for legacy outputs which are migrated in.

j-berman:
`derive_key_image_generator` is used to generate `I` for both legacy and Carrot outputs, note the function:

```c++
  void crypto_ops::derive_key_image_generator(const public_key &pub, const bool biased, ec_point &ki_gen) {
    if (biased)
      biased_derive_key_image_generator(pub, ki_gen);
    else
      unbiased_derive_key_image_generator(pub, ki_gen);
  }
```

`biased=true` for legacy outputs, and `biased=false` for Carrot outputs.

The scope of this PR (and the intended scope of the audit) are the lines of code specifically included in this PR, and not how they are used in the FCMP++/Carrot integration, which is slated for a later audit round. I was just aiming to provide some general context in that bullet.

j-berman:
Audited by Cypher Stack and Trail of Bits. Updated the PR description with details.

src/crypto/blake2b.c
```diff
@@ -515,4 +515,47 @@ int blake2b_long(void *pout, size_t outlen, const void *in, size_t inlen) {
 }
 /* Argon2 Team - End Code */
 
+int blake2b_monero(void *out, size_t outlen, const void *in, size_t inlen) {
```
UkoeHB: This is missing an optional 32-byte key input.

UkoeHB (changes requested):

src/crypto/blake2b.h
```diff
@@ -111,6 +111,9 @@ extern "C" {
 	int blake2b_long(void *out, size_t outlen, const void *in, size_t inlen);
 	/* Argon2 Team - End Code */
 
+	int blake2b_monero(void *out, size_t outlen, const void *in, size_t inlen,
+		const void *key, size_t keylen);
```
UkoeHB, line 115:
```suggestion
	/// Accepts an optional 32-byte key.
	int blake2b_monero(void *out, size_t outlen, const void *in, size_t inlen,
		const void *key);
```

UkoeHB (changes requested):

src/crypto/blake2b.c
```diff
@@ -515,4 +515,63 @@ int blake2b_long(void *pout, size_t outlen, const void *in, size_t inlen) {
 }
 /* Argon2 Team - End Code */
 
+int blake2b_monero(void *out, size_t outlen, const void *in, size_t inlen,
+	const void *key, size_t keylen) {
```
UkoeHB, line 523:
```suggestion
int blake2b_monero(void *out, size_t outlen, const void *in, size_t inlen,
	const void *key) {
```

src/crypto/blake2b.h
j-berman: The other `blake2b` functions that accept a `key` param all support `keylen <= BLAKE2B_KEYBYTES`, same with the [reference impl](https://github.com/BLAKE2/BLAKE2/blob/ed1974ea83433eba7b2d95c5dcd9ac33cb847913/ref/blake2b-ref.c#L120-L126). So long as we're adding functionality for an optional `key` here, I think it's reasonable to support the `keylen` param too to match the other functions in the API. Callers can easily just set `keylen = 32`

UkoeHB: `Blake2bMonero` on the rust side is a baked-in 32-byte optional key, as are the Carrot functions.

UkoeHB: I think there is a benefit to standardizing/constraining the API, since a broader API implies via Hyrum's Law that anything the API allows *might* be in use. We have always only used 32-byte keys and don't have a reason to change that. It's fine to require *expand only with good reason*.

j-berman: If someone does want to use the `blake2b_monero` function in the future with a key with `keylen != 32` but they don't modify the function to support it, the function won't necessarily return an error nor explicitly indicate failure/incorrect behavior (aside from the comment). It seems like a footgun to leave like that to me.

I see it as more likely that happens than a caller calling it with a `key` they intended to be 32 bytes, but they pass an incorrect non-zero value for the `keylen`.

The Rust side uses a `[u8; 32]` type rather than a `void *`, so isn't prone to this particular issue

UkoeHB: > If someone does want to use the blake2b_monero function in the future with a key with keylen != 32 but they don't modify the function to support it, the function won't necessarily return an error nor explicitly indicate failure/incorrect behavior (aside from the comment). 

In that case I'd support asserting keylen == 32.

j-berman: Done

src/crypto/blake2b.c
```diff
@@ -515,4 +516,72 @@ int blake2b_long(void *pout, size_t outlen, const void *in, size_t inlen) {
...
+
+	if (NULL != key) {
+		uint8_t block[BLAKE2B_BLOCKBYTES];
+		memset(block, 0, BLAKE2B_BLOCKBYTES);
+		memcpy(block, key, keylen);
+		if (blake2b_update(&S, block, BLAKE2B_BLOCKBYTES) < 0) {
+			goto fail;
+		}
+		/* Burn the key from stack */
+		clear_internal_memory(block, BLAKE2B_BLOCKBYTES);
```
jeffro256: `block` is not wiped if `blake2b_update()` fails

jeffro256: Suggested diff:

```diff
diff --git a/src/crypto/blake2b.c b/src/crypto/blake2b.c
index 5e2c51e36..4b4c7828e 100644
--- a/src/crypto/blake2b.c
+++ b/src/crypto/blake2b.c
@@ -279,11 +279,13 @@ int blake2b_init_key(blake2b_state *S, size_t outlen, const void *key, size_t ke
 
        {
                uint8_t block[BLAKE2B_BLOCKBYTES];
+               int r;
                memset(block, 0, BLAKE2B_BLOCKBYTES);
                memcpy(block, key, keylen);
-               blake2b_update(S, block, BLAKE2B_BLOCKBYTES);
+               r = blake2b_update(S, block, BLAKE2B_BLOCKBYTES);
                /* Burn the key from stack */
                clear_internal_memory(block, BLAKE2B_BLOCKBYTES);
+               (void) r; assert(0 == r);
        }
        return 0;
 }
@@ -565,13 +567,13 @@ int blake2b_monero(void *out, size_t outlen, const void *in, size_t inlen,
 
        if (NULL != key) {
                uint8_t block[BLAKE2B_BLOCKBYTES];
+               int r;
                memset(block, 0, BLAKE2B_BLOCKBYTES);
                memcpy(block, key, keylen);
-               if (blake2b_update(&S, block, BLAKE2B_BLOCKBYTES) < 0) {
-                       goto fail;
-               }
+               r = blake2b_update(&S, block, BLAKE2B_BLOCKBYTES);
                /* Burn the key from stack */
                clear_internal_memory(block, BLAKE2B_BLOCKBYTES);
+               (void) r; assert(0 == r);
        }
 
        if (blake2b_update(&S, in, inlen) < 0) {

```

src/crypto/blake2b.c
```diff
@@ -515,4 +516,72 @@ int blake2b_long(void *pout, size_t outlen, const void *in, size_t inlen) {
 }
 /* Argon2 Team - End Code */
 
+int blake2b_monero(void *out, size_t outlen, const void *in, size_t inlen,
```
jeffro256, line 522: Just FYI, `src/carrot_core/hash_functions.cpp` in #9559 was refactored to use this

src/crypto/blake2b.c
j-berman: > block is not wiped if blake2b_update() fails

Handled in the latest. Figure it makes sense for the caller to fail if `blake2b_update` fails.

UkoeHB (approved):

UkoeHB (approved):

src/crypto/blake2b.c
```diff
@@ -280,9 +281,12 @@ int blake2b_init_key(blake2b_state *S, size_t outlen, const void *key, size_t ke
 		uint8_t block[BLAKE2B_BLOCKBYTES];
 		memset(block, 0, BLAKE2B_BLOCKBYTES);
 		memcpy(block, key, keylen);
-		blake2b_update(S, block, BLAKE2B_BLOCKBYTES);
+		int r = blake2b_update(S, block, BLAKE2B_BLOCKBYTES);
 		/* Burn the key from stack */
 		clear_internal_memory(block, BLAKE2B_BLOCKBYTES);
+		if (r < 0) {
+			return -1;
+		}
```
jeffro256, line 289:
```suggestion
		if (r) {
			return r;
		}
```

src/crypto/blake2b.c
```diff
@@ -515,4 +519,73 @@ int blake2b_long(void *pout, size_t outlen, const void *in, size_t inlen) {
 }
 /* Argon2 Team - End Code */
 
+int blake2b_monero(void *out, size_t outlen, const void *in, size_t inlen,
+	const void *key, size_t keylen) {
+	// At this time, we expect all Monero-based callers to use a keylen=32
+	// if a key is included. If a caller wants to use a key and keylen != 32,
+	// just remove this assert (and #include <assert.h>).
+	assert(keylen == 0 || keylen == 32);
```
jeffro256, line 527: I recommend putting this statement underneath the variable declarations in this block

src/crypto/blake2b.c
```diff
@@ -515,4 +519,73 @@ int blake2b_long(void *pout, size_t outlen, const void *in, size_t inlen) {
...
+	if (NULL != key) {
+		uint8_t block[BLAKE2B_BLOCKBYTES];
+		memset(block, 0, BLAKE2B_BLOCKBYTES);
+		memcpy(block, key, keylen);
+		int r = blake2b_update(&S, block, BLAKE2B_BLOCKBYTES);
+		/* Burn the key from stack */
+		clear_internal_memory(block, BLAKE2B_BLOCKBYTES);
+		if (r < 0) {
+			goto fail;
+		}
```
jeffro256, line 578:
```suggestion
		ret = blake2b_update(&S, block, BLAKE2B_BLOCKBYTES);
		/* Burn the key from stack */
		clear_internal_memory(block, BLAKE2B_BLOCKBYTES);
		if (ret) {
			goto fail;
		}
```

src/crypto/crypto.cpp
```diff
@@ -618,11 +619,58 @@ namespace crypto {
     ge_p1p1_to_p3(&res, &point2);
   }
 
+  void crypto_ops::unbiased_hash_to_ec(const unsigned char *preimage, const std::size_t length, ec_point &res) {
+    uint8_t hash[64];
```
jeffro256:
```suggestion
    unsigned char hash[64];
```

src/crypto/crypto.cpp
```diff
@@ -618,11 +619,58 @@ namespace crypto {
...
+    if (blake2b_monero(std::addressof(hash), 64, preimage, length, NULL, 0) != 0)
+      throw std::runtime_error("Failed blake2b");
+
+    ge_p2 first;
+    ge_fromfe_frombytes_vartime(&first, reinterpret_cast<const unsigned char *>(&hash));
```
jeffro256:
```suggestion
    ge_fromfe_frombytes_vartime(&first, hash);
```

src/crypto/crypto.cpp
```diff
@@ -618,11 +619,58 @@ namespace crypto {
...
+    ge_p1p1 first_p1p1;
+    ge_mul8(&first_p1p1, &first);
+    ge_p3 first_p3;
+    ge_p1p1_to_p3(&first_p3, &first_p1p1);
+
+    ge_p2 second;
+    ge_fromfe_frombytes_vartime(&second, reinterpret_cast<const unsigned char *>(&hash) + 32);
```
jeffro256:
```suggestion
    ge_fromfe_frombytes_vartime(&second, hash + 32);
```

src/crypto/crypto.cpp
```diff
@@ -618,11 +619,58 @@ namespace crypto {
...
+    ge_p1p1_to_p3(&second_p3, &second_p1p1);
+    ge_cached second_cached;
+    ge_p3_to_cached(&second_cached, &second_p3);
+
+    ge_p1p1 point;
+    ge_add(&point, &first_p3, &second_cached);
+
+    ge_p3 res_ge_p3;
+    ge_p1p1_to_p3(&res_ge_p3, &point);
+    ge_p3_tobytes(&res, &res_ge_p3);
```
jeffro256, line 648: Integrating this would be outside the scope of the PR, but is there anywhere in the daemon code where the `derive_key_image_generator()` would be called, and then immediately followed up with a multiplication or point conversion (Wei)? In these cases, we may be able to speed up tree building by returning a `ge_p3` instead of a `crypto::ec_point`, avoiding one compression and decompression per tx output. What would actually be in-scope for this PR would be factoring out (and maybe exposing) that `ge_p3` overload of  `unbiased_hash_to_ec()`, much like how `biased_hash_to_ec()` does.

jeffro256 (changes requested):

src/crypto/blake2b.c
j-berman: Disagree with this suggestion, this suggestion deviates from the rest of the function which all returns -1 on any error, not the return value of a specific function

src/crypto/blake2b.c
j-berman: Why?

src/crypto/blake2b.c
j-berman: Also disagree with this for the same/similar reason as the other: the file uses a pattern of returning -1 on any failure. I think it's reasonable either to maintain that pattern, or modify the pattern for all other functions too/all other places. And I think for purposes of this PR, path of least resistance here is maintaining the pattern (and perhaps a future PR could modify all other spots)

src/crypto/crypto.cpp
j-berman: Yep, there is, nice spot. That's an optimization on the table that I think will be easier to implement on the PR that calls `derive_key_image_generator`

src/crypto/crypto.cpp
j-berman: Done in latest (as well as the other 2 related suggestions)

j-berman:
Note: the unbiased hash-to-point using `blake2_monero` was merged in the `monero-oxide` repo on the rust side ( https://github.com/monero-oxide/monero-oxide/pull/182 )

src/crypto/crypto.cpp
jeffro256: Right, but doesn't that overload need to exist, and would this PR be a good place to expose it?

j-berman: To implement it, I would implement it with all integration code in my local, assess it then, and then peel of the change for this PR if it seems definitively worth it. I would find it easier development wise to do it on the PR that actually utilizes this code and would demonstrably benefit from the change. I think that approach would be a smoother process, and ultimately I also think it's preferable to avoid focusing on such optimizations in this piecemeal phase, so that we can more efficiently move dependent PR's along

j-berman:
Resolved merge conflicts with #10345 in the crypto tests

src/crypto/crypto.cpp
jeffro256: Fair enough

src/crypto/blake2b.c
jeffro256: For C89 compilation, it's not a big deal and can be trivially refactored later

jeffro256 (approved):

UkoeHB (approved):

PR #10340, opened 2026-02-25
ringct: remove dead multisig_out type

jeffro256:
Bumps the wallet cache version. Alternatively, wallet cache format could stay the same, with the two empty vectors serialized on write.

### DIFF
--- a/src/ringct/rctTypes.h
+++ b/src/ringct/rctTypes.h
@@ -120,19 +120,6 @@ namespace rct {
         ~multisig_kLRki() { memwipe(&k, sizeof(k)); }
     };
 
-    struct multisig_out {
-        std::vector<key> c; // for all inputs
-        std::vector<key> mu_p; // for all inputs
-        std::vector<key> c0; // for all inputs
-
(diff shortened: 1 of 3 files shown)


src/ringct/rctTypes.h
```diff
@@ -807,7 +793,6 @@ VARIANT_TAG(binary_archive, rct::boroSig, 0x9a);
 VARIANT_TAG(binary_archive, rct::rctSig, 0x9b);
 VARIANT_TAG(binary_archive, rct::Bulletproof, 0x9c);
 VARIANT_TAG(binary_archive, rct::multisig_kLRki, 0x9d);
-VARIANT_TAG(binary_archive, rct::multisig_out, 0x9e);
```
vtnerd, line 810: When was `multisig_out` even used in a variant? Was this just added out of habit?

nahuhh: https://github.com/monero-project/monero/commit/4c313324b1c80148dff1a8099aa26c51ab6c7e3a

🤷‍♂️

tobtoht (approved):

PR #10342, opened 2026-02-27
crypto: torsion clearing for FCMP++

j-berman:
This introduces 4 new crypto functions used in the FCMP++ integration:

- `mul8_is_identity_vartime`
  - Outputs that fail this check cannot enter the FCMP++ tree.
  - Such outputs are not spendable today.
- `clear_torsion_vartime`
  - Outputs must have torsion cleared **before** they enter the FCMP++ tree.
- `get_valid_torsion_cleared_point_vartime`
  - "`valid`" means that the point is able to be inserted to the FCMP++ tree.
  - If false, the point cannot enter the tree.
  - In addition to the point decompression check, we also make sure the point is not equal to identity **after** clearing torsion.
- `rct::verPointsForTorsion`
  - After the FCMP++ fork, we use this function to stop allowing outputs that fail `get_valid_torsion_cleared_point_vartime` OR have torsion from entering the chain.
  
  _______
  
An earlier state of this PR was audited by Cypher Stack and Trail of Bits:
- Cypher Stack (and @Boog900) identified a redundant check for mul8 is identity in the torsion clear.
  - Cypher Stack section [6.4.2](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf).
- Cypher Stack recommended suffixing vartime functions with `_vartime`.
  - Section [6.5](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf).
- Cypher Stack recommended using the scalar type for `EC_INV_EIGHT`.
  - Section [6.3.2](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf)

All are remediated in the latest version of this PR.

Trail of Bits did not identify issues with this PR in their audit. Will share their audit once it's public.

### DIFF
--- a/src/crypto/crypto.h
+++ b/src/crypto/crypto.h
@@ -320,6 +320,9 @@ namespace crypto {
   inline std::ostream &operator <<(std::ostream &o, const crypto::view_tag &v) {
     epee::to_hex::formatted(o, epee::as_byte_span(v)); return o;
   }
+  inline std::ostream &operator <<(std::ostream &o, const crypto::ec_point &v) {
+    epee::to_hex::formatted(o, epee::as_byte_span(v)); return o;
+  }
 
   const extern crypto::public_key null_pkey;
   const extern crypto::secret_key null_skey;
@@ -336,6 +339,7 @@ inline const unsigned char* to_bytes(const crypto::ec_scalar &scalar) { return &
 inline unsigned char* to_bytes(crypto::ec_point &point) { return &reinterpret_cast<unsigned char&>(point); }
 inline const unsigned char* to_bytes(const crypto::ec_point &point) { return &reinterpret_cast<const unsigned char&>(point); }
 
+CRYPTO_MAKE_HASHABLE(ec_point)
 CRYPTO_MAKE_HASHABLE(public_key)
 CRYPTO_MAKE_HASHABLE_CONSTANT_TIME(secret_key)
 CRYPTO_MAKE_HASHABLE_CONSTANT_TIME(public_key_memsafe)
--- a/src/fcmp_pp/fcmp_pp_crypto.cpp
+++ b/src/fcmp_pp/fcmp_pp_crypto.cpp
@@ -1,0 +1,68 @@
+// Copyright (c) 2024, The Monero Project
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+// 
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+// 
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+// 
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "fcmp_pp_crypto.h"
+
+namespace fcmp_pp
+{
+//----------------------------------------------------------------------------------------------------------------------
+bool mul8_is_identity_vartime(const ge_p3 &point) {
+    ge_p2 point_ge_p2;
+    ge_p3_to_p2(&point_ge_p2, &point);
+    ge_p1p1 point_mul8;
+    ge_mul8(&point_mul8, &point_ge_p2);
+    ge_p3 point_mul8_p3;
+    ge_p1p1_to_p3(&point_mul8_p3, &point_mul8);
+    return ge_p3_is_point_at_infinity_vartime(&point_mul8_p3);
+}
+//----------------------------------------------------------------------------------------------------------------------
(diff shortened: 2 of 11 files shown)


j-berman:
Cypher Stack and Trail of Bits completed their audits of this PR. The latest includes changes suggested by Cypher Stack.

src/fcmp_pp/fcmp_pp_crypto.cpp
```diff
@@ -0,0 +1,68 @@
...
+    ge_p1p1 point_inv_8_mul_8;
+    ge_mul8(&point_inv_8_mul_8, &point_inv_8);
+    ge_p3 torsion_cleared_point;
+    ge_p1p1_to_p3(&torsion_cleared_point, &point_inv_8_mul_8);
+    crypto::ec_point k_out;
+    ge_p3_tobytes(to_bytes(k_out), &torsion_cleared_point);
+    return k_out;
+}
+//----------------------------------------------------------------------------------------------------------------------
+bool get_valid_torsion_cleared_point_vartime(const crypto::ec_point &point, crypto::ec_point &torsion_cleared_out) {
```
UkoeHB, line 57: For integration: I would add a comment to `output_to_tuple` that it's *_very important_* for `derive_key_image_generator` to be called on the *uncleared* pubkey (there is a comment but it's worth amplifying). To support that I would move `derive_key_image_generator` before `get_valid_torsion_cleared_point_vartime` is called.

Also note this PR doesn't include `get_valid_torsion_cleared_point_fast`.

src/ringct/rctSigs.cpp
```diff
@@ -1574,4 +1575,41 @@ namespace rct {
       key mask;
       return decodeRct(rv, sk, i, mask, hwdev);
     }
+
+    bool verPointsForTorsion(const std::vector<key> & pts) {
+      if (pts.empty())
+        return true;
+
+      tools::threadpool& tpool = tools::threadpool::getInstanceForCompute();
```
UkoeHB, line 1585: Perf test justifying threadpool overhead? Could also be improved by batching a range as is done elsewhere. Instead of lots of small tasks.

src/ringct/rctSigs.cpp
```diff
@@ -1574,4 +1575,41 @@ namespace rct {
...
+      }
+
+      if (!waiter.wait())
+        return false;
+      for (size_t i = 0; i < torsion_free.size(); ++i) {
+        if (!torsion_free[i]) {
+          LOG_PRINT_L1("Torsion check failed for point " << i);
+          return false;
+        }
+      }
```
UkoeHB: Simpler to just use an atomic bool? Save on deque alloc and the iter.

tests/unit_tests/crypto.cpp
```diff
@@ -343,3 +346,90 @@ TEST(Crypto, generator_consistency)
...
+      ASSERT_NE(rct::rct2pt(k), cleared);
+      torsioned_pts.push_back(k);
+    }
+    CHECK_CLEARED(k, cleared);
+    all_pts.emplace_back(k);
+  }
+
+  ASSERT_TRUE(rct::verPointsForTorsion(torsion_free_pts));
+  ASSERT_FALSE(rct::verPointsForTorsion(torsioned_pts));
+  ASSERT_FALSE(rct::verPointsForTorsion(all_pts));
```
UkoeHB, line 503: Add `ASSERT_TRUE` for the cleared versions of `torsioned_pts`.

UkoeHB (changes requested):

src/fcmp_pp/fcmp_pp_crypto.cpp
UkoeHB: https://github.com/seraphis-migration/monero/issues/294 says `get_valid_torsion_cleared_point_fast` is for 'future optional audit'.

src/ringct/rctSigs.cpp
j-berman: [This commit here](https://github.com/j-berman/monero/commit/59ded67de7ed1726f304524872b9f472a88dd267) implements a perf test implementing:
A) multithreaded + batch per thread
B) multithreaded + individual tasks to thread pool
C) no multithreading / no batching

On my 16 thread Ryzen, A and B are ~13-14x faster than C, and B is consistently faster than A by a non-negligible amount.

I didn't include that perf test in the final PR primarily because I had to comment out `set_process_affinity(1)` to test it, and also modify the function signature to take in a threadpool + `batch_size` arg which secondarily seems a little unclean

src/ringct/rctSigs.cpp
j-berman: Done in latest

tests/unit_tests/crypto.cpp
j-berman: Done in latest

UkoeHB (approved):

src/crypto/crypto.h
```diff
@@ -328,6 +331,19 @@ namespace crypto {
   inline bool operator>(const public_key &p1, const public_key &p2) { return p2 < p1; }
   inline bool operator<(const key_image &p1, const key_image &p2) { return memcmp(&p1, &p2, sizeof(key_image)) < 0; }
   inline bool operator>(const key_image &p1, const key_image &p2) { return p2 < p1; }
+
+  static const ec_point EC_I = {1};
+
+  static const ec_scalar EC_INV_EIGHT = {{
+      static_cast<char>(0x79), static_cast<char>(0x2f), static_cast<char>(0xdc), static_cast<char>(0xe2),
```
jeffro256: Technically, I don't think that this is portable until C++20. A unsigned -> signed conversion before C++20, if the value cannot be exactly represented as the destination type, is implementation-defined.

jeffro256: Also, I would prefer this constant to be moved into `fcmp_pp_crypto.cpp`, where it is used, unless there are plans to use it elsewhere.

UkoeHB: There is a unit test validating equivalence with the rct version, so it at least needs to be public.

j-berman: I worked around this with a gross `static_cast<char>(static_cast<signed char>(signed integer))` in the latest (took me 4 tries to solve all build errors). Open to better ideas.

Also moved the definitions to `fcmp_pp_crypto.h` (so reusable in the tests)

j-berman: ~~Another (untested) option I think is `static_cast<signed char>(hex)` and then a `static_assert(static_cast<signed char>(0xff) == -1)` to be sure the impl is expected~~ (EDIT: nevermind I think this will error on some platforms still). The reason I used hex originally was to match the hex from the rct inv 8.

@selsta also mentioned bumping to C++20.

I'm indifferent on this.

UkoeHB (approved):

src/ringct/rctSigs.cpp
```diff
@@ -1574,4 +1577,35 @@ namespace rct {
...
+            const crypto::ec_point &point = rct::rct2pt(pts[i]);
+            crypto::ec_point torsion_cleared_point;
+            if (fcmp_pp::get_valid_torsion_cleared_point_vartime(point, torsion_cleared_point)
+                && point == torsion_cleared_point)
+            {
+              // Point is torsion free if it's equal to itself after clearing torsion
+              return;
+            }
+            bool expected = false;
+            invalid_found.compare_exchange_strong(expected, true);
```
jeffro256: Why not `invalid_found.store(true)`? Also, wouldn't it be better to use an `std::atomic_flag` instead of a `std::atomic<bool>`?

j-berman: Used `.store` in the latest, and switched from `invalid_found` to `all_valid` pattern to make it clearer. Also squashed and resolved conflicts

UkoeHB (approved):

jeffro256 (approved):

PR #10345, opened 2026-03-02
crypto: ed25519 -> Wei conversion

j-berman:
Builds on top of:
- #10342
- #10111
- #10135

When inserting outputs into the FCMP++ tree, we convert each output pubkey, key image generator, and commitment (ed25519 points) into Wei x and y coordinates (as part of the [tower cycle](https://gist.github.com/tevador/4524c2092178df08996487d4e272b096) critical for the curve tree -- Wei coordinates are "Selene" scalars).

This PR introduces the crypto functions necessary to perform the conversion:

- `point_to_ed_derivatives`
- `ed_derivatives_to_wei_x_y`
- `fe_ed_derivatives_to_wei_x_y`

It uses the equation from E.2 in https://www.ietf.org/archive/id/draft-ietf-lwig-curve-representations-02.pdf.

______

An earlier state of this PR was audited by Cypher Stack and Trail of Bits:
- Trail of Bits identified we were using the negative constant C from the paper on converting ed25519 to Wei coordinates.
  - Remediated and added a test that shows we're using the same hex constants from the paper.
  - Rust side also updated [here](https://github.com/monero-oxide/monero-oxide/commit/066229801a18bb11679113c93971c4f19fde0cd0).
- Both Cypher Stack and Trail of Bits pointed out a potential footgun if `fe_batch_invert` is called with a 0 value in `ed_derivatives_to_wei_x_y`.
  - Remediated by failing `ed_derivatives_to_wei_x_y` if so.
  - Cypher Stack section [8.4.3](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf).
- Trail of Bits recommended adding documentation to the conversion functions to be clear on expected input params.

All are remediated in the latest version of this PR. Will share the Trail of Bits audit once public.


j-berman:
Audited by both Trail of Bits and Cypher Stack. Updated the description highlighting their findings and the remediation. Will share the Trail of Bits audit on the description once public.

UkoeHB (approved):

src/fcmp_pp/fcmp_pp_crypto.h
```diff
@@ -59,10 +59,44 @@ static const crypto::ec_scalar EC_INV_EIGHT = {{
     static_cast<char>(static_cast<signed char>(0)),   static_cast<char>(static_cast<signed char>(6))
   }};
 //----------------------------------------------------------------------------------------------------------------------
+// Field elems needed to get wei x and y coords
+// Take note of the bounds, and make sure downstream field ops can take said bounds as input.
+struct EdDerivatives final
+{
+    fe one_plus_y;        // bounded by 1.1*2^26,1.1*2^25,1.1*2^26,1.1*2^25,etc.
+    fe one_minus_y;       // bounded by 1.1*2^26,1.1*2^25,1.1*2^26,1.1*2^25,etc.
+    fe one_minus_y_mul_x; // bounded by 1.01*2^25,1.01*2^24,1.01*2^25,1.01*2^24,etc.
```
jeffro256, line 68: Why does this field have a tighter bound on it if it's going to be passed to `fe_mul()`?

j-berman: It doesn't have a tighter bound for any reason other than that all these bounds are just the resulting bounds of these ops [here](https://github.com/seraphis-migration/monero/blob/228b9b7f51c3bb2a16c4c38dbaf847e95db01f81/src/fcmp_pp/fcmp_pp_crypto.cpp#L359-L363). I commented the bounds to make it clear what they are in case someone decides to use these fields for something else for some reason.

jeffro256 (approved):

PR #10348, opened 2026-03-05
epee: show error-level log messages at the default log level

navidR:
At the default log level (0), eight log categories (`net, net.http, net.ssl, net.p2p, net.cn, daemon.rpc, verify, serialization`) were set to FATAL-only, which silenced all ERROR-level messages from these components. This meant that when something went wrong — like a port already in use — the user only saw a generic message like "`Exception in main! Failed to initialize p2p server`." with the actual error hidden unless they set `log-level=1`. This change promotes those categories from FATAL to ERROR so that error details are always visible.

### DIFF
--- a/contrib/epee/src/mlog.cpp
+++ b/contrib/epee/src/mlog.cpp
@@ -100,7 +100,7 @@ static const char *get_default_categories(int level)
   switch (level)
   {
     case 0:
-      categories = "*:WARNING,net:FATAL,net.http:FATAL,net.ssl:FATAL,net.p2p:FATAL,net.cn:FATAL,daemon.rpc:FATAL,global:INFO,verify:FATAL,serialization:FATAL,daemon.rpc.payment:ERROR,stacktrace:INFO,logging:INFO,msgwriter:INFO";
+      categories = "*:WARNING,net:ERROR,net.http:ERROR,net.ssl:ERROR,net.p2p:ERROR,net.cn:ERROR,daemon.rpc:ERROR,global:INFO,verify:ERROR,serialization:ERROR,daemon.rpc.payment:ERROR,stacktrace:INFO,logging:INFO,msgwriter:INFO";
       break;
     case 1:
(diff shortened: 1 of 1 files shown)


vtnerd:
Have you noticed any additional spam to the logs? Hopefully we didn't overuse `MERROR` in these categories.

navidR:
> Have you noticed any additional spam to the logs? Hopefully we didn't overuse `MERROR` in these categories.

I have been running it for a while and haven't noticed any spamming. So IMHO we are good.

vtnerd (approved):

contrib/epee/src/mlog.cpp
```diff
@@ -100,7 +100,7 @@ static const char *get_default_categories(int level)
   switch (level)
   {
     case 0:
-      categories = "*:WARNING,net:FATAL,net.http:FATAL,net.ssl:FATAL,net.p2p:FATAL,net.cn:FATAL,daemon.rpc:FATAL,global:INFO,verify:FATAL,serialization:FATAL,daemon.rpc.payment:ERROR,stacktrace:INFO,logging:INFO,msgwriter:INFO";
+      categories = "*:WARNING,net:ERROR,net.http:ERROR,net.ssl:ERROR,net.p2p:ERROR,net.cn:ERROR,daemon.rpc:ERROR,global:INFO,verify:ERROR,serialization:ERROR,daemon.rpc.payment:ERROR,stacktrace:INFO,logging:INFO,msgwriter:INFO";
```
jeffro256, line 103: NACK. This adds many spam vector for malicious peers. Be more precise about exactly which error messages you want to be exposed to the user on the default log level.

nahuhh: context (i requested the below)

scenario:

1. Start node (a) with default ports
2. Start node (b) with default ports = node (b) fails to start, and throws a useless error
3. Start node (b) on log lvl 1 = the error tells you that there is a port conflict

The objective of this pr is to not require lvl 1 to return a useful error. I dont know why we return useless errors on lvl 0, it should just tell you what the problem is on lvl 0

PR #10351, opened 2026-03-06
 src: fix monerod shutdown hanging by making SIGINT handling async-signal-safe

navidR:
This fix annoying issue of `Ctrl-C` not working. Particularly happens when you are hitting the `Ctrl-C` at initialization.

### DIFF
--- a/src/common/util.cpp
+++ b/src/common/util.cpp
@@ -87,6 +87,7 @@ using namespace epee;
 #include <boost/algorithm/string.hpp>
 #include <boost/asio.hpp>
 #include <boost/format.hpp>
+#include <boost/thread/thread.hpp>
 #include <openssl/evp.h>
 
 #undef MONERO_DEFAULT_LOG_CATEGORY
@@ -117,14 +118,42 @@ static int flock_exnb(int fd)
 struct posix_signal_state
 {
   boost::asio::io_context io;
+  boost::asio::executor_work_guard<boost::asio::io_context::executor_type> work;
   boost::asio::signal_set signals;
+  boost::thread worker;
   bool wait_active;
 
   posix_signal_state()
-    : io(), signals(io), wait_active(false)
+    : io(), work(boost::asio::make_work_guard(io)), signals(io), worker(), wait_active(false)
   {
     signals.add(SIGINT);
     signals.add(SIGTERM);
+  }
+
+  ~posix_signal_state()
+  {
+    try
+    {
+      signals.cancel();
+      work.reset();
+      io.stop();
+    }
+    catch (...)
+    {
+    }
+    try
+    {
+      if (worker.joinable())
+      {
+        if (worker.get_id() == boost::this_thread::get_id())
+          worker.detach();
+        else
+          worker.join();
+      }
+    }
+    catch (...)
+    {
+    }
   }
 };
 #endif
@@ -142,9 +171,26 @@ void arm_posix_wait();
 
 void handle_signal(int type)
 {
-  boost::lock_guard<boost::mutex> lock(signal_handler_handler_mutex);
-  if (signal_handler_handler)
-    signal_handler_handler(type);
+  std::function<void(int)> handler;
+  {
+    boost::lock_guard<boost::mutex> lock(signal_handler_handler_mutex);
+    handler = signal_handler_handler;
+  }
+  if (handler)
+  {
+    try
+    {
+      handler(type);
+    }
+    catch (const std::exception &e)
+    {
+      MERROR("Exception in signal handler: " << e.what());
+    }
+    catch (...)
+    {
+      MERROR("Unknown exception in signal handler");
+    }
+  }
 }
 
 #if defined(WIN32)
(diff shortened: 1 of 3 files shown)


contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1706,26 +1706,40 @@ namespace net_utils
...
       {
-        //timeout
+        // timeout
         sock_.close();
         _dbg3("Failed to connect to " << adr << ":" << port << ", because of timeout (" << conn_timeout << ")");
         return CONNECT_FAILURE;
       }
+      const auto remaining = deadline - now;
+      const auto wait_step = remaining < boost::posix_time::milliseconds(100)
+        ? remaining
```
vtnerd: This is:
```c++
wait_step = std::min(remaining, boost::posix_time::milliseconds(100));
```

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1706,26 +1706,40 @@ namespace net_utils
...
+        : boost::posix_time::milliseconds(100);
+      bool r = local_shared_context->cond.timed_wait(lock, now + wait_step);
+      if(local_shared_context->ec == boost::asio::error::would_block && !r)
+      {
+        continue;
```
vtnerd: Why a `continue` at the end of the loop?

src/common/util.cpp
```diff
@@ -133,6 +135,73 @@ namespace tools
   }
 
   std::function<void(int)> signal_handler::m_handler;
+#ifndef WIN32
```
vtnerd: We should switch to `boost::asio::signal_set` while we are hacking away in here. That way we don't need custom platform specific stuff. If we switch to it, we must switch all signal handlers to asio (thats the only catch).

I really don't want another `thread` for signal handling.

contrib/epee/include/net/abstract_tcp_server2.inl
navidR: Thanks. Fixed.

contrib/epee/include/net/abstract_tcp_server2.inl
navidR: Thanks. Fixed. That was a mistake.

src/common/util.cpp
navidR: I think your suggestion is correct. Therefore, I have switched to `boost::asio::signal_set` implementation right now. 

Without any extra thread for signal handling.

src/common/util.cpp
```diff
@@ -133,6 +133,82 @@ namespace tools
   }
 
   std::function<void(int)> signal_handler::m_handler;
+  boost::mutex signal_handler::m_handler_mutex;
+#ifndef WIN32
+  std::shared_ptr<boost::asio::io_context> signal_handler::m_signal_io;
```
vtnerd: These two (`m_signal_io` and `m_signal_set`) should be combined into a single `struct` instead of allocating them separately.

src/common/util.h
```diff
@@ -169,27 +175,31 @@ namespace tools
...
     static bool install(T t)
     {
+      {
+        boost::lock_guard<boost::mutex> lock(m_handler_mutex);
+        m_handler = std::move(t);
+      }
 #if defined(WIN32)
-      bool r = TRUE == ::SetConsoleCtrlHandler(&win_handler, TRUE);
-      if (r)
+      const bool r = TRUE == ::SetConsoleCtrlHandler(&win_handler, TRUE);
```
vtnerd: Why the separate variant for `WIN32`? `boost::asio` should handle all cross platform stuff with signal handlers

src/common/util.h
```diff
@@ -208,22 +218,32 @@ namespace tools
...
+    static void arm_posix_wait();
+    static boost::mutex m_signal_mutex;
+#endif
+
     /*! \brief where the installed handler is stored */
     static std::function<void(int)> m_handler;
+    static boost::mutex m_handler_mutex;
+
+#ifndef WIN32
+    static std::shared_ptr<boost::asio::io_context> m_signal_io;
```
vtnerd: Move both these variables to the cpp in the anonymous namespace. Then move the header includes to cpp too to reduce header bloat.

src/common/util.h
vtnerd: Or I guess the benefit is the lack of asio needed to do the job?

src/common/util.cpp
vtnerd: It might make sense to keep them separate if `async_wait` were capturing the `signal_set`, but it's not.

navidR: Good suggestion; implemented.

src/common/util.h
navidR: Good suggestion; implemented.

src/common/util.h
navidR: This one, I am hesitant about.  Since replacing `SetConsoleCtrlHandler` with asio on Windows changes signal/control-event semantics.

navidR:
@vtnerd I added second commit to fix the DNS wait.

navidR:
Just look at the difference between fix and master: 

Fix:
fix

master:
master

PR #10352, opened 2026-03-06
Fix ZMQ-PUB reporting of mempool txes that were initially in stem phase

vtnerd:
An update by @jeffro256 changed how the txpool set `m_added_to_pool` - it only set the value on first observation as opposed to first fluff/block observation. Unfortunately, the ZMQ-PUB code was designed to only report txes observed during block/fluff (suppressing "knowledge" of the tx during stem phase).

This means a node taking part in the stem phase of a tx never reports the tx over ZMQ currently. This reverts back to the prior behavior by using the return code from `txpool::add_tx`.

Watching the CI, and then pushing a 0.18 patch too.

### DIFF
--- a/src/cryptonote_core/tx_pool.h
+++ b/src/cryptonote_core/tx_pool.h
@@ -116,6 +116,8 @@ namespace cryptonote
      * @tx_relay how the transaction was received
      * @param tx_weight the transaction's weight
      * @param valid_input_verification_id a previously valid verID if non-null
+     * @return True if tx passes verification checks AND is first observation
+     *   of tx in `broadcasted` relay method.
      */
     bool add_tx(transaction &tx, const crypto::hash &id, const cryptonote::blobdata &blob,
       size_t tx_weight, tx_verification_context& tvc, relay_method tx_relay, bool relayed,
@@ -144,7 +146,8 @@ namespace cryptonote
      * passes the non-input consensus tests (e.g. for newly received relayed txs), then leave
      * "nic_verified_hf_version" as its default value of 0 (there is no v0 fork).
      *
-     * @return true if the transaction passes validations, otherwise false
+     * @return True if tx passes verification checks AND is first observation
+     *   of tx in `broadcasted` relay method.
      */
     bool add_tx(transaction &tx, tx_verification_context& tvc, relay_method tx_relay, bool relayed,
       uint8_t version, uint8_t nic_verified_hf_version = 0,
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1138,7 +1138,7 @@ namespace cryptonote
     const bool res = m_mempool.add_tx(tx, tx_hash, blob, tx_weight, tvc, tx_relay, relayed, version);
 
     // If new incoming tx passed verification and entered the pool, notify ZMQ
-    if (!tvc.m_verifivation_failed && tvc.m_added_to_pool && matches_category(tx_relay, relay_category::legacy))
+    if (!tvc.m_verifivation_failed && res && matches_category(tx_relay, relay_category::legacy))
     {
       m_blockchain_storage.notify_txpool_event({txpool_event{
         .tx = tx,


selsta:
When was this bug introduced?

vtnerd:
PR #9135 (on master), and the related PR on 0.18. So it's been there for about a year. No one noticed because 0-conf still _mostly_ works.

src/cryptonote_core/tx_pool.h
```diff
@@ -116,6 +116,10 @@ namespace cryptonote
      * @tx_relay how the transaction was received
      * @param tx_weight the transaction's weight
      * @param valid_input_verification_id a previously valid verID if non-null
+     * @return True if tx is stored in the mempool DB AND (`tx_relay == block`
+     *    OR `matches_category(tx_relay) == broadcasted` for first time). Thus
+     *    the return value is primarily an indicator when the tx should be
+     *    relayed via ZMQ-PUB.
```
j-berman: Wouldn't it also return true if adding a stem tx for the first time? Hence the need for `matches_category(tx_relay, relay_category::legacy)` in that if statement above

Looks like it's true if it's stored in the mempool DB (and it may have already been stored as well)

j-berman (commented):
Code LGTM, nice catch. I just have a q re: the comment

src/cryptonote_core/tx_pool.h
vtnerd: Correct, this comment needs to be tweaked.

vtnerd:
Force pushed a comment change.

src/cryptonote_core/tx_pool.h
```diff
@@ -116,6 +116,10 @@ namespace cryptonote
      * @tx_relay how the transaction was received
      * @param tx_weight the transaction's weight
      * @param valid_input_verification_id a previously valid verID if non-null
+     * @return True if first observation of tx OR `tx_relay == block`
+     *    OR `matches_category(tx_relay) == broadcasted` for first time. Thus
+     *    the return value is primarily an indicator when the tx should be
+     *    relayed via ZMQ-PUB.
```
jeffro256: Isn't this not quite accurate since ZMQ shouldn't be notified first when txs are in stem phase, hence the `&& matches_category(tx_relay, relay_category::legacy)` condition in the above code?

jeffro256:
Were stem txs only ever pushed to ZMQ after relaying them?

https://github.com/monero-project/monero/blob/1bd57c8e959d80d7f6cb21a0e560958c21c5d693/src/cryptonote_core/cryptonote_core.cpp#L1468

Since the "main" place where incoming txs were handled excluded txs which weren't freshly added to the mempool:

https://github.com/monero-project/monero/blob/1bd57c8e959d80d7f6cb21a0e560958c21c5d693/src/cryptonote_core/cryptonote_core.cpp#L1107

https://github.com/monero-project/monero/blob/1bd57c8e959d80d7f6cb21a0e560958c21c5d693/src/cryptonote_core/cryptonote_core.cpp#L1098-L1101

src/cryptonote_core/tx_pool.h
vtnerd: Are you requesting a correction for "Thus the return value..."? I agree this comment still isn't quite right.

vtnerd:
@jeffro256 I'm not sure what exactly you're asking, but I'll attempt to explain the snippets you highlighted.

The first situation happens when the tx is received via stem, gets transferred to the D++ code, but it's in fluff mode for the current epoch. The D++ code notifies the txpool to update the tx status and then immediately broadcasts over ZMQ-PUB.

The second situation happens when a tx was received as fluff and was either previously unknown or previously !fluff. The latter check is not obvious - the txpool checks for duplicate key images but ignores the duplicate if the tx was not in fluff. If a key image comes from a different tx_hash the tx is also rejected.

What happened is that `m_added_to_pool` used to silently hide stem txes, where it no longer does.

vtnerd:
Or maybe you're asking to move the ZMQ-PUB to on_transactions_relay instead?

EDIT:
After looking into this: txpool stores the tx immediately in the add function, with the relay method determined from the p2p/rpc code. If received as fluff, the on_transactions_relay call won't know the original context. We could store an additional bit of metadata to indicate whether the tx has ever been ZMQ notified, etc.

vtnerd:
I've caught another issue by code review. If a tx hits a stem loop, it never gets reported via ZMQ by that first loop node. It is difficult to test this case, so I can't really confirm other than by auditing the code. I believe this has always been an issue with the ZMQ-PUB system. The case where D++ is in the fluff phase is already handled correctly.

The check should be `matches_category(tvc.m_relay, relay_category::legacy)` which catches the case where `txpool::add_tx` upgraded the relay method to fluff due to cycle detection. The `tvc.m_relay` value is otherwise identical to `tx_relay`.

I'll let this sit for another day to think over whether the change has some negative impact.

j-berman (approved):
After reading the above discussion between @vtnerd and @jeffro256, I figured it's worth stating my understanding. This solution makes sense for the following sequence of events:

- Node receives stem tx in stem phase, relays it as stem tx and adds to the pool as stem tx. The node doesn't zmq pub the tx yet since it's not visible in its pool to external observers.
- Node receives the tx from another node as fluff tx, then upgrades its relay status in `handle_incoming_tx`.

**Without this PR**: the tx doesn't get published via zmq because A) `m_added_to_pool` is false so `core::add_new_tx` won't pub it, and B) `core::on_transactions_relayed` won't see that the tx was "just broadcasted" because at that point, the tx's relay status is already fluff (i.e. no "upgrade" happens that `m_mempool.set_relayed` would indicate the tx was "just broadcasted").

**With this PR**: the tx gets published via zmq in `core::add_new_tx` once it's received from another node as a fluff tx.

src/cryptonote_core/tx_pool.h
jeffro256: Yes

jeffro256 (changes requested):
I would request that the comment be reworded a bit, but this is a good change, and IIUC #10415 addresses my other concerns I had

selsta (approved):

src/cryptonote_core/tx_pool.h
```diff
@@ -144,7 +146,8 @@ namespace cryptonote
      * passes the non-input consensus tests (e.g. for newly received relayed txs), then leave
      * "nic_verified_hf_version" as its default value of 0 (there is no v0 fork).
      *
-     * @return true if the transaction passes validations, otherwise false
+     * @return True if tx passes verification checks AND is first observation
+     *   of tx in `broadcasted` relay method.
```
j-berman, line 150: This isn't necessarily accurate because if it's a first-time local or stem, then it can still return true and won't be in `broadcasted` relay method. How about:

```
     * @return True if tx is in the pool AND (first observation of tx
     *   OR `matches_category(tx_relay) == broadcasted` for first time
     *   OR `tx_relay == block`).
```

It can fail input verification checks if `tx_relay == block`. It must be the case that it's in the pool though if returning true.

j-berman (approved):
I think the comment can be fixed in a follow-up PR, I think it's worth moving forward with the code fix.

PR #10357, opened 2026-03-11
Cleanup some of the fragmented levin handling

vtnerd:
The fragmented message handling in levin currently allows padding bytes to be sent to the `epee` storage parser. That parser ignores everything after the `epee` message. This upgrades the message handling so that the `epee` parser never sees the padding bytes in the fragmented message.

There are tons of unit tests related to this, including tx notify code. This basically proves that this can be added without disrupting existing p2p communications.


vtnerd:
The usual test failed :/

PR #10358, opened 2026-03-11
FCMP++: ed25519 output -> (pre-leaf) tuple in preparation for insertion to the curve tree merkle tree

j-berman:
Builds on top of:
- #10345
- #10338

This PR builds on the FCMP++ crypto building blocks to introduce higher level functions that take as input an ed25519 output (specifically the pair of output public key and commitment), and returns objects that will be turned into leaves of the curve tree merkle tree.

The two core functions in this PR are:
1. `output_to_tuple`
    - Function to convert an `{output pubkey, commitment}` to an output tuple `{O,I,C}` in preparation to insert the output tuple into the curve tree.
    - `O` = torsion cleared valid output pubkey checked for identity.
    - `I` = key image generator.
    - `C` = torsion cleared valid Commitment checked for identity.
    - None of `{O,I,C}` should have torsion nor == identity.
    - This change also introduces the `OutputPair` variant, which can either be Legacy or Carrot V1 types. Legacy outputs are not checked for torsion at consensus, and use the legacy biased hash to point fn to derive the key image generator `I`. Carrot V1 outputs **are** checked for torsion at consensus, and use the unbiased hash to point to derive the key image generator `I`.
    - Also notable is that the `OutputTuple` struct is compatible over the Rust FFI. Hence its inclusion in the new `fcmp_pp_rust` lib introduced in this PR.
2. `output_to_pre_leaf_tuple`
    - A `PreLeafTuple` is an object composed of "ed25519 derivatives" that are the intermediate objects needed to convert an output into full-fledged leaf tuples efficiently.
    - We'll be able to batch invert these derivatives later on when deriving Wei x and y coordinates.

_____

An earlier state of this PR was audited by Trail of Bits and Cypher Stack. Neither found any issues. See Cypher Stack's [Section 10](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf) (will share the Trail of Bits audit here once it's public).

### DIFF
--- a/src/fcmp_pp/curve_trees.cpp
+++ b/src/fcmp_pp/curve_trees.cpp
@@ -33,12 +33,24 @@
 #include "fcmp_pp_types.h"
 #include "profile_tools.h"
 
+namespace
+{
+    // Struct composed of ec elems needed to get a full-fledged leaf tuple
+    struct PreLeafTuple final
+    {
+        fcmp_pp::EdDerivatives O_derivatives;
+        fcmp_pp::EdDerivatives I_derivatives;
+        fcmp_pp::EdDerivatives C_derivatives;
+    };
+}
 
 namespace fcmp_pp
 {
 namespace curve_trees
 {
 //----------------------------------------------------------------------------------------------------------------------
+//----------------------------------------------------------------------------------------------------------------------
+// Public helper functions
 //----------------------------------------------------------------------------------------------------------------------
 OutputTuple output_to_tuple(const OutputPair &output_pair)
 {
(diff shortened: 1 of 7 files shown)


j-berman:
Audited by both Trail of Bits and Cypher Stack. Neither identified issues with this PR.

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,159 @@
...
+    const crypto::ec_point &O = (crypto::ec_point&) o.O;
+    const crypto::ec_point &I = (crypto::ec_point&) o.I;
+    const crypto::ec_point &C = (crypto::ec_point&) o.C;
+
+    /*
+        TODO: investigate perf impact of the following extraneous ops:
+        - Decompressing O and C when checking points for torsion and again here.
+        - Decompressing output pubkey in derive_key_image_generator and again here (when O == output_pubkey).
+        - Compressing I in derive_key_image_generator and decompressing again here.
+    */
```
j-berman, line 134: @jeffro256 highlighting this: I think it will be a bit of legwork to restructure to avoid these extra (de)compressions. My opinion: I think it would be best to keep it as a TODO for the time being, and then potentially circle back to it once we get all the integration code in to move the code along.

EDIT: the incorrect bullet `Decompressing output pubkey in derive_key_image_generator and again here (when O == output_pubkey)` can be ignored. I removed that bullet in the latest and github still shows the original here.

j-berman: I looked into it a bit more, and I maintain the above fwiw. From memory, decompressions did not have a significant impact on perf relative to torsion clearing/checking and tree build hashing. Tree build hashing was the primary bottleneck after the FCMP++ fork. So I do think it's worth it to move forward with this as is, and also worth circling back to this later on in an optimization pass.

src/fcmp_pp/fcmp_pp_types.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+//----------------------------------------------------------------------------------------------------------------------
+bool output_checked_for_torsion(const OutputPair &output_pair)
+{
+    struct output_pair_visitor
+    {
+        bool operator()(const CarrotOutputPairV1&) const
+        { return true; }
+        bool operator()(const LegacyOutputPair&) const
+        { return false; }
```
UkoeHB:
```suggestion
        bool operator()(const LegacyOutputPair&) const
        { return false; }
        bool operator()(const CarrotOutputPairV1&) const
        { return true; }
```
Consistency

src/fcmp_pp/fcmp_pp_types.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+//----------------------------------------------------------------------------------------------------------------------
+bool use_biased_hash_to_point(const OutputPair &output_pair)
+{
+    struct output_pair_visitor
+    {
+        bool operator()(const CarrotOutputPairV1&) const
+        { return false; }
+        bool operator()(const LegacyOutputPair&) const
+        { return true; }
```
UkoeHB:
```suggestion
        bool operator()(const LegacyOutputPair&) const
        { return true; }
        bool operator()(const CarrotOutputPairV1&) const
        { return false; }
```

UkoeHB (changes requested):
LGTM, minor nit

UkoeHB (approved):

jeffro256 (approved):

PR #9801, opened 2025-02-17
guix: add rust

tobtoht:
**This change introduces cross-architecture non-determinism**. This means that cross-compiling using Rust from a different host architectures to a common target will not produce bit-identical output. 

This effectively means that participants in the [verified reproduction process](https://github.com/monero-project/guix.sigs) must build on a `x86_64` host. This mainly affects users on ARM64 Apple devices and users on experimental RISC-V boards.

---

Rust is [self-hosted](https://en.wikipedia.org/wiki/Self-hosting_(compilers)#Compilers) and because the compiler typically uses new language features as soon as they become available it can only be built with the previous minor version of Rust. This makes Rust notoriously hard to bootstrap.

Its bootstrap graph looks like this:

```
Rust 1.93.0
↓
Rust 1.92.0
↓
Rust 1.91.0
↓
Rust 1.90.0
↓
Rust 1.89.0
↓
Rust 1.88.0
↓
Rust 1.87.0
↓
Rust 1.86.0
↓
Rust 1.85.1
↓
Rust 1.84.1
↓
Rust 1.83.0
↓
Rust 1.82.0
↓
Rust 1.81.0
↓
Rust 1.80.1
↓
Rust 1.79.0
↓
Rust 1.78.0
↓
Rust 1.77.1
↓
Rust 1.76.0
↓
Rust 1.75.0
↓
mrustc 1.74.0
↓
llvm 17.0.6
```

You can obtain this graph locally by running `guix graph rust | xdot -`

Note: `mrustc` can now bootstrap Rust `1.90` which would again greatly decrease the build time for a full source bootstrap. This will land in Guix sooner or later.

#### Changes to the build environment

| package | old | new |
|---|---|---|
| llvm | n/a | 21.1.8 |
| python-minimal-wrapper | n/a | 3.12.12 |
| rust | n/a | 1.93.0 |

#### Changes to transitive build dependencies

Excluding the Rust bootstrap chain shown above.

| package | old | new |
|---|---|---|
| boost | n/a | 1.83.0 |
| dejagnu | n/a | 1.6.3 |
| expect | n/a | 5.45.4 |
| gdb | n/a | 17.1 |
| git-minimal | n/a | 2.50.0 |
| icu4c | n/a | 73.1 |
| llvm | n/a | 17.0.6 |
| procps | n/a | 4.0.3 |
| source-highlight | n/a | 3.1.9 |
| tcsh | n/a | 6.24.15 |

PR #10359, opened 2026-03-11
FCMP++: Rust FFI + crypto conversion `selene_scalar_from_bytes`

j-berman:
Builds on top of:
- #10358

This PR introduces the Rust FFI, as well as a function to take 32 bytes -- the byte representation of a Selene scalar -- and convert the byte representation into the Selene scalar object used in the Rust code.

Recall "Selene" is the name of the curve in the tower cycle towering Ed25519: https://gist.github.com/tevador/4524c2092178df08996487d4e272b096 . As part of constructing the curve tree merkle tree, we take Ed25519 points and convert them to Selene scalars (the leaves of the tree are composed of Selene scalars).

The `selene_scalar_from_bytes` function introduced in this PR takes the 32 byte representation (that we derive from ed25519 points on the C++ side), and produces the Selene scalar Rust FFI compatible object that we use in future Rust FFI functions when constructing the curve tree merkle tree.

_____

Trail of Bits and Cypher Stack audited an earlier state of this PR. Both pointed out a fragile memory layout assumption of the FFI-compatible Rust objects. See Cypher Stack's [section 11](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf), and the Trail of Bits audit [pages 18-19, and 23](https://github.com/trailofbits/publications/blob/master/reviews/2026-07-magicgrants-monerofcmp++crypto-securityreview.pdf). This was remediated in the latest version of the code (with [this change](https://github.com/j-berman/monero/pull/20) that uses static asserts on size and alignment on both sides of the FFI boundary).

Trail of Bits reviewed the remediation and concluded that their concern was "partially resolved," and encourages improving the robustness of the API further (page 23). The comment included in this PR is the justification for leaving it as is:

> /// This type is expected to be opaque to the C/C++ side, meaning only the Rust side should read/write
/// its internal represenation. We're using a modified crypto-bigint crate for this type so that we
/// can work with points and scalars across the FFI without tons of byte repr conversions.
struct SeleneScalar {


j-berman:
Audited by Trail of Bits and Cypher Stack. See the description for details.

src/fcmp_pp/fcmp_pp_rust/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+edition = "2021"
+
+[lib]
+name = "fcmp_pp_rust"
+crate-type = ["staticlib"]
+
+[dependencies]
+ciphersuite = { version = "0.4.2", features = ["ed25519"] }
+dalek-ff-group = "0.4.4"
+helioselene = { git = "https://github.com/monero-oxide/monero-oxide", rev = "50dd3dbc0cc9048247eb734d27eb1d516aecacd2" }
```
jeffro256: Why is this using a year-old revision of helioselene?

src/fcmp_pp/fcmp_pp_rust/fcmp++.h
```diff
@@ -28,12 +28,47 @@
...
+
+/// A constant-time implementation of the Ed25519 field.
+/// This type is expected to be opaque to the C/C++ side, meaning only the Rust side should read/write
+/// its internal represenation. We're using a modified crypto-bigint crate for this type so that we
+/// can work with points and scalars across the FFI without tons of byte repr conversions.
+struct SeleneScalar {
+  uintptr_t _0[32 / sizeof(uintptr_t)];
+};
+FFI_STATIC_ASSERT(sizeof(struct SeleneScalar) == 32, "SeleneScalar FFI type unexpected size");
+FFI_STATIC_ASSERT(alignof(struct SeleneScalar) == sizeof(uintptr_t), "SeleneScalar FFI type unexpected alignment");
```
UkoeHB, line 53: Is there a reason not to hardcode the alignment? There seems to be an assumption that `sizeof(uintptr_t) [C++] == size_of<usize>() [rust]`.

UkoeHB (approved):
Approval conditional on Jeffro's comment.

Needs review from someone who knows cmake. Also not sure the procedure for validating Cargo.lock (taking into account it's probably going to change at least once before code freeze).

src/fcmp_pp/fcmp_pp_rust/Cargo.toml
j-berman: Probably a very stale local mixup. Updated that and the `dalek-ff-group` lib version in the latest

src/fcmp_pp/fcmp_pp_rust/fcmp++.h
j-berman: > Is there a reason not to hardcode the alignment?

It's platform dependent (32-bit platforms will be different from 64-bit e.g.)

> There seems to be an assumption that sizeof(uintptr_t) [C++] == size_of<usize>() [rust].

Rust's unsafe guidelines mentioned the following [here](https://github.com/rust-lang/unsafe-code-guidelines/blob/c138499c1de03b908dfe719a41193c84f8146883/reference/src/layout/scalars.md#isize-and-usize):

> The isize and usize types are pointer-sized signed and unsigned integers. They have the same layout as the [pointer types](https://github.com/rust-lang/unsafe-code-guidelines/blob/c138499c1de03b908dfe719a41193c84f8146883/reference/src/layout/pointers.md) for which the pointee is Sized, and are layout compatible with C's uintptr_t and intptr_t types.

Notably that page was archived as of [this commit](https://github.com/rust-lang/unsafe-code-guidelines/blob/d809f91e62968913d808e53397fbcf7cea762cce/reference/src/layout/scalars.md).

We are relying on the assumption that `uintptr_t [C] == usize [rust]` elsewhere in the FFI.

j-berman:
> Needs review from someone who knows cmake

Pinging @tobtoht 

> Also not sure the procedure for validating Cargo.lock

Proposing that before release, we do something like:
1) A deep pass at eliminating as many dependencies as possible.
2) Make sure all deps are updated to the latest while maintaining our expected minimum supported Rust version. 
3) Freeze remaining deps and review them.
4) Then any changes to deps in the future would undergo a similar process as currently done for updating Monero's current deps.

EDIT: also pinging @kayabaNerve who may have thoughts on an ideal procedure here

src/fcmp_pp/fcmp_pp_rust/CMakeLists.txt
```diff
@@ -25,3 +25,101 @@
...
+endif()
+
+set(RUST_TARGET "${RUST_ARCH}-${RUST_PLATFORM}${RUST_TOOLCHAIN}")
+message(STATUS "Using Rust target ${RUST_TARGET}")
+
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+  if(SANITIZE)
+    message(STATUS "Using Rust ASAN")
+    # Rust ASAN is an experimental feature included in nightly at time of writing
+    set(CARGO_CMD RUSTFLAGS=-Zsanitizer=address cargo +nightly build --target "${RUST_TARGET}")
```
jeffro256: Missing `${CARGO_OPTIONS}` here?

See: `git diff fcmp++-beta-stressnet -- src/fcmp_pp/fcmp_pp_rust/CMakeLists.txt`

`CARGO_OPTIONS` isn't set in the top CMakeLists.txt like it is in the beta stressnet branch, but is that planned eventually? If not, then you can ignore this comment.

src/fcmp_pp/fcmp_pp_rust/CMakeLists.txt
```diff
@@ -25,3 +25,101 @@
...
+  else()
+    set(CARGO_CMD cargo build --target "${RUST_TARGET}")
```
jeffro256: And here

src/fcmp_pp/fcmp_pp_rust/CMakeLists.txt
```diff
@@ -25,3 +25,101 @@
...
+  endif()
+  set(TARGET_DIR "debug")
+else ()
+  set(CARGO_CMD cargo build --target "${RUST_TARGET}" --release)
```
jeffro256: And here

src/fcmp_pp/CMakeLists.txt
```diff
@@ -29,18 +29,42 @@
...
   PUBLIC
     cncrypto
   PRIVATE
-    ${EXTRA_LIBRARIES})
+    ${CMAKE_CURRENT_BINARY_DIR}/fcmp_pp_rust/libfcmp_pp_rust.a
+    ${EXTRA_LIBRARIES}
+    ${EXTRA_RUST_LIBRARIES})
+
+# Try compiling C file ffi_api_c_compat.c which includes fcmp++.h, and throw an error if C compilation fails
+try_compile(FFI_API_C_COMPAT_RES
```
jeffro256, line 61: This should probably be moved to an object compilation target, which is included in the `all` target. That would mean it doesn't take up time compiling while configuring. But we can address that if that becomes an issue. It's correct as-is.

jeffro256:
BTW I approve conditional on:

> CARGO_OPTIONS isn't set in the top CMakeLists.txt like it is in the beta stressnet branch, but is that planned eventually? If not, then you can ignore this comment.

kayabaNerve:
`monero-oxide` has a pretty thorough set of CI workflows, where my most recent maintenance aims to more accurately define our MSRV and ensure our requirements are clearly stated.

I'm sure we _can_ create a monorepo of the Rust libraries as for usage in Monero, and tree-shake out many dependencies and lines of code. Unfortunately, I don't believe that process will be scalable to maintain. I have a repository which forked a large repository, and does such slashing/burning/patching, and it wasn't feasible to update. The solution I ended up with was a combination Rust binary + shell script + patch set, implementing a myriad of Rust syntaxes, to programmatically remove large swaths of code. You can see that here: https://github.com/serai-dex/patch-polkadot-sdk

I'm not claiming that's a role model, but it shows how complex this _can_ be. At the same time, the Rust libs for FCMP++ are much smaller and _could_ be updated less frequently, making such an effort feasible. I think `polkadot-sdk` starts at 2800 crates in its graph and the FCMP++ libraries are... ~60? so there is quite a difference in scale.

The proper solution would be [`cargo vet`](https://github.com/mozilla/cargo-vet) to track every dependency and every patch, at least moving forward. A qualified review of every dep would likely be a couple months of work however and really be a team effort to ensure coverage. I'm not comfortable reviewing long division algorithms in crypto-bigint, for example, though we may be able to remove all of those and achieve a feasible for us to directly review surface if we do adopt a tree-shaking approach.

There's also a 'maintainable' solution, without forking, where we simply own our dependencies. Why pull in `blake2` when we can have a reference `blake2` impl in our repository, in a few hundred lines? We can forsake all the many pleasantries of the Rust ecosystem and simply try to lock this down to the minimal LoC to implement FCMP++, where we own everything.

I am explicitly against that last solution, despite being the 'best' for supply-chain security, as it'd be such a rewrite, we'd no longer be able to interop with the Rust ecosystem. We'd lose all generality of our libraries and likely take multiple step backs on performance. We'd also lose the potential of other consumers and such peer review. At that point, I don't think it'd be too much work to just move to a C implementation with such properties and achieve client diversity.

---

There is a half-step to tree-slashing in our own monorepo. Serai has patches in its repo: https://github.com/serai-dex/serai/tree/next/patches

They handle crates in our graph which aren't actually reachable (it's a known issue with cargo where features of dependencies which could be enabled, even if they weren't enabled and therefore can't be enabled (as feature enabling is limited to the crate actually being built), are still included in the manifest), crates which are reachable but aren't actually used, some version discrepancies where there's no effective API difference, and some stubs to reduce our LoC/risk.

https://github.com/serai-dex/serai/blob/next/patches/ethereum/fixed-cache/src/lib.rs is one example where I stubbed a cache library, which I'm not convinced is safe, to simply always yield cache misses. Yes, this one feature of this one mega Ethereum library will now have cache misses across whatever relevant functionality, but I'm 85% sure I wasn't touching that stuff anyways and 100% sure I don't care.

Monero can include such patches _on its end, in its build profile_.

TL;DR many options, how much effort do we have to put in, how maintainable does the result have to be?

jpk68:
@kayabaNerve It looks like `cargo vet` itself will add another 263 crates. So, to also ensure the legitimacy of the vetting tools we're using, we'd need to manually audit ~323 crates?

All of this pain and sorrow could likely have been saved by just writing it in C...

kayabaNerve:
> @kayabaNerve It looks like cargo vet itself will add another 263 crates. So, to also ensure the legitimacy of the vetting tools we're using, we'd need to manually audit ~323 crates?

Do you personally review and audit your system's `gcc`? We do not need to audit the tool we use for auditing, we just have to have a basic degree of trust in it, which is the same degree of trust we have in our C toolchain. This isn't language-specific.

While `cargo vet` is the existing tool to do this work, and has a pleasant UI and interoperable format, we can manually maintain a review board, with our notes and certifications. We don't _have_ to use a tool for it.

> All of this pain and sorrow could likely have been saved by just writing it in C...

I think this is an absurd comment honestly not even worth replying to. While I can argue for the benefits of the Rust programming language, and the Rust ecosystem, it comes down to two things:

1) I did this work for Monero. I did it in Rust. You can moan about how _someone_ didn't do it in C, but without someone to do it in C, it doesn't matter. You're welcome to do such a rewrite if you want. 

2) Without the `multisig` feature, but with all other features and even the `dev-dependencies`, I record a total of 72 crates in our tree for a total of 600k lines.

- `libc`: 133k lines
- `fiat-crypto` (a formally verified cryptography backend for `curve25519-dalek`, which is automatically generated from some formal definition, which we can remove if we don't want to use): 83k lines
- `syn`: 66k lines. This is the... second? most popular Rust crate of all time, and is used for compile-time macros as `rustc`'s lexer is internal, so this is an alternative lexer consumable as a library. I actually hate it, at least in the modern day, but it can be considered almost as an extension to the standard library.
- `typenum`: 41k. This one is actually funny. Rust's generics are explicitly not turing complete, so it's basically `U1, U2` where `<U1 as Add<U2>>::Output` is `U3`. It's just a ton of generated lines of code for numeric relationships.

Those four make up more than half of the lines of code, and either can be removed or should be rather quick to accept, just as one accepts `boost` in their C++ projects. `boost`, for the record, is _12.6 million lines of code_, which is more than the _entirety of Cuprate_. If only `monerod` had been written in Rust, we could have saved all of this pain and sorrow :(((

Then we get to `curve25519-dalek`, `hashbrown` (an implementation of `HashSet`/`HashMap` owned by the RustLang org), `crypto-bigint`, hash functions, etc. The exact same stuff we would use a library for on the C/C++ side of things. We just wouldn't have an actively engaged in package ecosystem and would instead inline everything, which promotes one of two things:
1) Worse implementations used solely because they're smaller
2) Monolith libraries so we can claim to only have one dependency, instead of several

Note Monero has inlined several such cryptography libraries into itself, and that is one of the options above. It just decreases maintainability.

I'm not here to say we should always grab external dependencies. I support making the appropriate decision, and engaging in the review process. My comments above were:
- Here's what's feasible to review
- Here's what's feasible to maintain
- Here's the compromises

This really isn't a 'rust bad' comment though.

jpk68:
> If only `monerod` had been written in Rust, we could have saved all of this pain and sorrow :(((

Apologies if I came off a bit too harsh there, but I was joking. I don't *actually* think this problem is causing people physical/emotional pain and sorrow.

> Do you personally review and audit your system's `gcc`? We do not need to audit the tool we use for auditing, we just have to have a basic degree of trust in it, which is the same degree of trust we have in our C toolchain. This isn't language-specific.

The fact that such a tool is so helpful/"needed" for this says a lot about the Rust ecosystem and where its priorities lie; instead of just having a sane number of dependencies you can count on one or two hands (even "heavy" ones), open the floodgates to hundreds of fragmented micro-dependencies, because that's somehow not any worse. That one can't feasibly audit the entire source of their compiler toolchain is not reason enough to just throw this problem out the window; of course adding 60+ new dependencies, another toolchain, and more vetting/security tools would increase our attack surface.

This can also be seen in how difficult it has historically been to bootstrap Rust (especially before efforts such as `mrustc`). I am personally unconvinced that the opaque and seemingly credulous Rust toolchain/ecosystem is in any way a good fit for a project like Monero - that's not to mention a slough of other issues, such as the dependence of crates.io on centralized services like GitHub (and, by extension, having to comply with US trade law, censoring users from "bad" countries), but maybe that's a tale for a different time.

> The exact same stuff we would use a library for on the C/C++ side of things. We just wouldn't have an actively engaged in package ecosystem and would instead inline everything

Okay, so if that's what's required for proper supply-chain security, why not do it within one language instead of adding another? Even aside from security, such duplication will surely increase the size of binaries, and indeed, it has (by a pretty ludicrous amount). This got so bad that I seem to remember tobtoht suggesting we remove the 'utility' binaries from release archives on the site to save space, considering how much the FCMP++ Rust libraries bloated them up.

> I did this work for Monero. I did it in Rust. You can moan about how _someone_ didn't do it in C, but without someone to do it in C, it doesn't matter. You're welcome to do such a rewrite if you want.

I hugely appreciate this work, and please don't think otherwise.

kayabaNerve:
I didn't think you meant such a severe comment either. I just thought your comment wholly didn't contribute and also, was without merit. That's why I replied as such an ass. I am willing to apologize for being such an ass, as I don't want to actually sour relationships with those who mean well, but I do still think your comment lacked value :/

The tool would be helpful for any system where there are dependencies and you manage them. It literally just pulls the old versions, the new versions, and shows you the `git diff`. It then lets you assert a status, such as 'safe to run' (not malicious, fine for the inputs you have in your build pipeline) or 'safe to deploy' (not malicious, fine for arbitrary inputs) and add notes. If we vendor everything into a monorepo, which the `cargo` package manager _does_ have support for, we'll get the same via `git diff` proper. That just requires we maintain the monorepo and have our own system to write down notes and qualifications. `cargo vet` works over the lockfile alone and provides a structured format and a pleasant UI. That's why I brought it up.

> sane number of dependencies you can count on one or two hands (even "heavy" ones), open the floodgates to hundreds of fragmented micro-dependencies, because that's somehow not any worse.

`boost` is more than 12.6 million lines according to my `cloc` over the latest release, and more than the entire tree of Cuprate. I'm unsure how that is sane unless you say "but we trust boost", which then also handles most of the dependencies we used for these Rust libraries.

While their independence does make it easier for a malicious update to appear, I do personally check each package update to the Serai repository I perform to make sure it's not malicious. It's a feasible effort, when the Serai repository has ~20x the dependencies, and is an explicit purpose of `cargo vet` (tracking deltas).

> That one can't feasibly audit the entire source of their compiler toolchain is not reason enough to just throw this problem out the window

which is why I said this isn't a Rust-specific problem, not that this isn't a problem at all.

> of course adding 60+ new dependencies, another toolchain, and more vetting/security tools would increase our attack surface.

This part is fair, and one of the benefits of having multiple implementations. It's one of the reasons why I said, now that we've done this modular implementation which is feasible to update and review, we can discuss specializing it to the hyper-specific protocol we've finalized on, and if we are to go as far as to strip all dependencies, we could even discuss changing its language.

> This can also be seen in how difficult it has historically been to bootstrap Rust (especially before efforts such as mrustc)

I am aware of this and have actually worked on this problem. I'm happy to say, with the necessary pair of LLVM toolchains, you can build `mrustc` and ratchet up the latest `rustc` in just four hours or so on a modern workstation. I think that's fine.

> why not do it within one language instead of adding another? 

1) No one did this in C. If someone is willing to do this in C, I support them doing so and would be curious in the results. If the result works and has the necessary quality, I support `monerod` using it instead.

2) Rust has exceptionally strong safety and correctness guarantees, which is exactly what we want here.

3) I find the Rust modularity and composability via `trait`s exceptionally present, and it allowed prototyping and experimentation, and making changes in general, in a very pleasant way. Those properties were exceptionally necessary while the protocol was under active development. We only have the hindsight to start specializing now that it's done.

As for binary size, I don't have a comment there as it isn't something I personally care about and would rather leave to others to discuss.

src/fcmp_pp/fcmp_pp_rust/CMakeLists.txt
j-berman: @tobtoht introduced it as part of enabling bootstrappable builds ([source](https://github.com/monero-project/monero/pull/9440/changes/4f50eab208e8086898e94696228f9cd6a70b7418#diff-7084b1fffb546ef1bb316ac72bd2ee80f54e192a4a39e52d93e9f1cf085c2496)), so figure it makes sense to leave for @tobtoht to add it in #9440

j-berman:
@jeffro256 -- addressed the `CARGO_OPTIONS` comment

______

Thank you @kayabaNerve. I'll give my opinion on those thoughts in the next days. I think this PR is a good place to discuss steps to take to best manage/maintain the Rust before first release (and beyond), but just want to iterate the point that I don't think such discussion necessarily should block this PR. I think we should get this PR in and move forward toward getting the rest of the integration in, and can tackle best supply chain approach in parallel.

tobtoht:
>but just want to iterate the point that I don't think such discussion necessarily should block this PR.

I agree, however before adding any Rust related build system changes we should branch `release-v0.19` from master. We should also merge dependent PRs first so CI stays working. So, in order:

- [ ] ~Merge all remaining PRs under https://github.com/monero-project/monero/milestone/2~
- [x] Branch `release-v0.19`
- [x] Merge #9801
- [x] Merge #9440 
- [ ] Merge #10359 (this PR)

If we can't get the remaining milestone PRs merged within ~a week, I suggest we branch anyway and backport the patches that didn't make it in in time.

j-berman:
Merging dependent PR's first sgtm.

Is there a reasonable path forward where we wait to branch `release-v0.19` until the heavy majority of FCMP++/Carrot code is in master? There's going to be a whole lot more FCMP++ and Carrot code coming, and also more PR's to bring in from stressnet applicable to v0.18. It would be nice to avoid needing PR's for both master and v0.19 going forward (and v0.18 in some cases) until way more of the code is in master already.

tobtoht:
My understanding is that the purpose of the `release-v0.19` branch is to test all the changes that were accrued in `master` that weren't backported to `release-v0.18`. We haven't branched from `master` in years, so there are quite a lot of changes that haven't undergone widespread testing from users (Guix, Polyseed, etc). 

Once we branch `release-v0.19`, there will follow a period of internal testing and then we can tag `v0.19.0.0` at which point the `release-v0.18` branch is obsolete. This is a feature release, not a hardfork ready release. Downstream wallets will need time to rebase, upgrade their build system, and test their software. 

Fixes will have to be submitted to the `master` branch and `v0.19`, as we do today with `master` and `v0.18`. However, there is an awkward period between branching `release-v0.19` and releasing `v0.19.0.0` where it's possible that we need to do another `v0.18` release. I suggest that once we branch `release-v0.19`, we only backport critical fixes to `v0.18` and only do a release if absolutely necessary.

All FCMP++ and Carrot related changes should continue to _only_ go into `master`. Once FCMP++/Carrot is ready and we can begin the code freeze, we will branch it into `release-v0.20`.

j-berman:
Got it, I see. [This merge plan](https://github.com/monero-project/monero/pull/10359#issuecomment-5103134277) sounds good me.

jeffro256:
I don't want the review period of Polyseed and tx relay v2 to hold up merging of this PR. Can't we simply build v19.0.0.0 and just not include `fcmp_pp` in the build process for that release? That way, we aren't holding up FCMP++ review, and not bloating the v19 bootstrap graph.

src/fcmp_pp/fcmp_pp_rust/CMakeLists.txt
tobtoht: Please add them here. 9440 will go first, so we have a bit of a circular dependency going on.

tobtoht:
We have branched. All FCMP++ / Carrot related changes can now go into `master`. Fixes, polyseed, and tx relay v2 will be backported to `release-v0.19`.

src/fcmp_pp/fcmp_pp_rust/CMakeLists.txt
jeffro256: Also makes comparing the git diff easier

j-berman: Done

j-berman:
I brought in @tobtoht's @jpk68's, and @MrcyjaneK's changes in the latest version, and attributed co-authors.

@kayabaNerve implemented a large portion of the Rust tooling in cmake and helped set up the FFI.
@jeffro256 implemented the safer C compilation approach + `FFI_STATIC_ASSERT`.
@tobtoht helped with implementing the proper cmake commands.
@jpk68 [optimized the release binary](https://codeberg.org/jpk68/monero/pulls/18/files) and [rm'd the unnecessary msvc toolchain](https://codeberg.org/jpk68/monero/pulls/19).
@MrCyjaneK [fixed rust targets for iOS and iOS simulator](https://github.com/seraphis-migration/monero/pull/447).

jeffro256:
Did you mean to remove the MSVC Rust platform definitions in the latest?

j-berman:
Yes as per [this](https://codeberg.org/jpk68/monero/pulls/19) [change](https://github.com/seraphis-migration/monero/pull/381)

jeffro256 (approved):

UkoeHB (approved):

j-berman:
Rebased and included the following changes to activate CI following #9440:

```
diff --git a/contrib/guix/guix-build b/contrib/guix/guix-build
index a99c0d584..84c99584d 100755
--- a/contrib/guix/guix-build
+++ b/contrib/guix/guix-build
@@ -277,9 +277,8 @@ mkdir -p "$LOGDIR_BASE"
 RUST_DEPS_VERSION=0
 RUST_DEPS_HASH="gg5izydrra2b1ywgskvlrfrh983gf6b4"
 RUST_DEPS_ARCHIVE="rust_deps-${RUST_DEPS_VERSION}.tar.gz"
-#UNCOMMENT_IN_10359# RUST_DEPS_STORE_ITEM="/gnu/store/${RUST_DEPS_HASH}-${RUST_DEPS_ARCHIVE}"
-#UNCOMMENT_IN_10359# if [ ! -f "${RUST_DEPS_STORE_ITEM}" ]; then
-if false; then #REMOVE_IN_10359#
+RUST_DEPS_STORE_ITEM="/gnu/store/${RUST_DEPS_HASH}-${RUST_DEPS_ARCHIVE}"
+if [ ! -f "${RUST_DEPS_STORE_ITEM}" ]; then
     time-machine environment --manifest="${PWD}/contrib/guix/rust/cargo.scm" \
                              --container \
                              --pure \
diff --git a/contrib/guix/libexec/build.sh b/contrib/guix/libexec/build.sh
index eb3b86454..31dabd67d 100644
--- a/contrib/guix/libexec/build.sh
+++ b/contrib/guix/libexec/build.sh
@@ -373,7 +373,7 @@ mkdir -p "$DISTSRC"
 
     # Unpack rust dependencies
     mkdir -p /rust/vendor
-    #UNCOMMENT_IN_10359# tar xf /rust-deps -C /rust
+    tar xf /rust-deps -C /rust
 
     # "vendor" rust std
     for dir in "${RUST_STD}"/vendor/*/; do
```

jeffro256 (approved):

UkoeHB (approved):

redsh4de (approved):

jpk68 (approved):

tobtoht:
Alright alright, we'll fix CI in post.

PR #10360, opened 2026-03-11
FCMP++: CurveTrees class & implementation of ed25519 outputs -> curve tree leaves function

j-berman:
Builds on top of:
- #10359

The `outputs_to_leaves` function introduced in this PR takes in outputs that enter the chain, and converts them into valid leaves prepared to be inserted to the curve trees merkle tree.	We do not add invalid outputs to
the tree, so `outputs_to_leaves` checks outputs for validity and only includes the valid outputs in the final response.

The `CurveTrees` class is a useful object for updating the curve trees merkle tree. It is instantiated with the widths of each layer in the tree, as well as the curve objects themselves.

"Unified Outputs" are the set of all outputs in the chain (pre-RCT, RCT, and Carrot). Each output has a `unified_id`, which is the ordered index given to the output upon entering the chain (when the tx that creates the output enters a block in the chain). We use the `unified_id` to determine the order to insert a batch of outputs into the curve tree merkle tree.

_____

[Trail of Bits](https://github.com/trailofbits/publications/blob/73e52c96cef59d42201ff5546090b76ca67a8014/reviews/2026-07-magicgrants-monerofcmp%2B%2Bcrypto-securityreview.pdf) and Cypher Stack ([section 12](https://github.com/cypherstack/fcmp-impl-audit/blob/d7de33d08c48a3c249629e47f71ea72c938a3404/FCMP_Code_Implementation_Audit.pdf)) audited an earlier state of this PR. Neither identified issues with this PR. However, since both pointed out that if `fe_batch_invert` is passed a 0, it could lead to an erroneous result, I made sure to utilize the updated API for `fe_batch_invert` that returns a success response on success, and throw if returning a failure response. `fe_batch_invert` was included in #10111 and fails if it's passed a 0.


src/fcmp_pp/fcmp_pp_types.h
```diff
@@ -0,0 +1,128 @@
...
+const crypto::public_key &output_pubkey_cref(const OutputPair &output_pair);
+const crypto::ec_point &commitment_cref(const OutputPair &output_pair);
+
+bool output_checked_for_torsion(const OutputPair &output_pair);
+bool use_biased_hash_to_point(const OutputPair &output_pair);
+
+// Wrapper for outputs with context to insert the output into the FCMP++ curve tree
+struct UnifiedOutput final
+{
+    // Output's unique id in the chain, used to insert the output in the tree in the order it entered the chain
```
UkoeHB:
```suggestion
    // Output's unique id in the chain.
    // When a batch of enotes is added to the tree, the batch
    // is internally sorted by unique ids, which correspond with
    // the order the enotes entered the chain. Batch-added enotes
    // may not be globally ordered because e.g. timelocked
    // enotes were added to 'late batches'. Batch refers to
    // enotes that become unlocked/spendable at the same
    // time.
```

j-berman:
Audited by both Cypher Stack and Trail of Bits. The description highlights changes implemented as a result of the audits.

UkoeHB (approved):

redsh4de (approved):

src/fcmp_pp/CMakeLists.txt
```diff
@@ -51,7 +51,9 @@ endif()
 
 target_link_libraries(fcmp_pp
   PUBLIC
+    common
```
jeffro256: nit: out of alphabetical order

jeffro256 (approved):

jpk68 (approved):

src/fcmp_pp/CMakeLists.txt
j-berman: Fixed

selsta (approved):

jeffro256 (approved):

redsh4de (approved):

UkoeHB (approved):

PR #10363, opened 2026-03-12
Harden HTTP client auth

vtnerd:
This hardens a few things in HTTP auth, client side. No changes were required server side, so there should be no compatibility issues between the two. The unit tests are fairly robust at checking the various cases (also see updates to tests).

### DIFF
--- a/contrib/epee/src/http_auth.cpp
+++ b/contrib/epee/src/http_auth.cpp
@@ -64,6 +64,7 @@
 #include <iterator>
 #include <limits>
 #include <openssl/evp.h>
+#include <openssl/rand.h>
 #include <tuple>
 #include <type_traits>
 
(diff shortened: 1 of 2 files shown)


contrib/epee/src/http_auth.cpp
```diff
@@ -294,13 +295,21 @@ namespace
...
       boost::copy(out, nc.data());
+
+      std::array<uint8_t, 16> rbuf{{}};
+      if (RAND_bytes(rbuf.data(), rbuf.size()) != 1)
+        return {};
+
+      const std::string cnonce = epee::string_encoding::base64_encode(rbuf.data(), rbuf.size());
       const auto response = digest(
-        generate_a1(digest, user), u8":", user.server.nonce, u8":", nc, u8"::auth:", digest(method, u8":", uri)
+        generate_a1(digest, user), u8":", user.server.nonce, u8":", nc, u8":", cnonce, ":auth:", digest(method, u8":", uri)
```
iamamyth: Seems the `u8` prefix before `":auth:" was accidentally dropped

vtnerd: Good catch, fixed.

PR #10366, opened 2026-03-14
Improve unpack reserve handling

vtnerd:


vtnerd:
Force pushed a change to the tests.

selsta (approved):

src/serialization/container.h
```diff
@@ -76,7 +76,19 @@ namespace serialization
...
+    auto do_reserve(C &c, std::size_t N, std::size_t B) -> decltype(c.reserve(N))
+    {
+      using T = typename C::value_type;
+
+      // max compression ratio for upfront memory usage
+      B /= sizeof(T);
+      if (std::numeric_limits<std::size_t>::max() / 4 <= B)
+        B = std::numeric_limits<std::size_t>::max();
+      else
+        B *= 4;
```
j-berman, line 88: Where does the 4 come from?

vtnerd: Great question, it's an arbitrary choice to allow for 4-to-1 compression ratio in reserve size.

tests/unit_tests/serialization.cpp
```diff
@@ -372,6 +372,20 @@ TEST(Serialization, serializes_vector_int64_as_fixed_int)
...
+{
+  std::vector<int64_t> v;
+  string blob;
+
+  tools::write_varint(std::back_inserter(blob), unsigned(100));
+  blob.append(std::string(100, 0));
+
+  ASSERT_LT(v.capacity(), 20);
+  ASSERT_FALSE(serialization::parse_binary(blob, v));
+  ASSERT_LT(v.capacity(), 100); // could fail if lib allocates more in reserve call
```
jeffro256, line 385:
```suggestion
  const std::size_t expected_max_capacity = 50;
  const std::size_t starting_capacity = v.capacity();
  if (starting_capacity >= expected_max_capacity)
    GTEST_SKIP() << "STD lib reserved too many vector elements for unpack reserve test";
  ASSERT_FALSE(serialization::parse_binary(blob, v));
  const std::size_t end_capacity = v.capacity();
  ASSERT_GE(end_capacity, starting_capacity); // if this fails, then IDK 
  ASSERT_LE(end_capacity, expected_max_capacity); // could fail if lib allocates more in reserve call
```

src/serialization/container.h
```diff
@@ -76,7 +76,19 @@ namespace serialization
     template<typename... C>
     void do_reserve(const C&...) {}
     template<typename C>
-    auto do_reserve(C &c, std::size_t N) -> decltype(c.reserve(N)) { return c.reserve(N); }
+    auto do_reserve(C &c, std::size_t N, std::size_t B) -> decltype(c.reserve(N))
+    {
+      using T = typename C::value_type;
+
+      // max compression ratio for upfront memory usage
+      B /= sizeof(T);
```
jeffro256, line 84: This doesn't handle varint containers very well. For example, in your unit test, if you change `v` to a `std::vector<std::uint64_t>`, then the serialization will work because then `use_container_varint()` is `true`. A varint container, even though `sizeof()` may be 8, may legitimately have as many elements as there are bytes left. In other words, its "compression ratio" may be 8.

jeffro256 (changes requested):

tests/unit_tests/serialization.cpp
vtnerd: This seems like lot of code to just admit that the `reserve` call cannot be reliably tested.

src/serialization/container.h
vtnerd: Yes, this may under allocate/reserve in some situations. This was intentional, but perhaps undesirable.

The primary negative is that `std::vector` doubles on next allocation, so it may slightly over allocate in these situations. Thankfully, varint `std::vector` are somewhat rare iirc - the only one that comes to mind is the ring signature listings. And this array is somewhat small (comparatively).

jeffro256: I recommend this, it is more conservative for other types, and less conservative for `uint64_t`:

```patch
diff --git a/src/serialization/container.h b/src/serialization/container.h
index d0f78700a..68c069e54 100644
--- a/src/serialization/container.h
+++ b/src/serialization/container.h
@@ -30,9 +30,10 @@
 
 #pragma once
 
+#include <algorithm>
 #include <cstdint>
 #include <cstddef>
-#include <string>
+#include <limits>
 #include <type_traits>
 
 namespace serialization
@@ -73,6 +74,7 @@ namespace serialization
     { return c.emplace(e); }
 
     //! @brief Reserve space for N elements if applicable for container.
+    //! @param B bytes remaining in message
     template<typename... C>
     void do_reserve(const C&...) {}
     template<typename C>
@@ -80,12 +82,18 @@ namespace serialization
     {
       using T = typename C::value_type;
 
+      static constexpr std::size_t max_untrusted_compression_ratio =
+        is_blob_type<T>::type::value ? 1 :                                  // crypto::hash, etc cannot be compressed
+        use_container_varint<T>() ? sizeof(T) :                             // varints can be as small as 1 byte
+        (std::is_same_v<T, char> || std::is_same_v<T, unsigned char>) ? 1 : // char types cannot be compressed
+        4;                                                                  // default compression ratio
+
       // max compression ratio for upfront memory usage
       B /= sizeof(T);
-      if (std::numeric_limits<std::size_t>::max() / 4 <= B)
+      if (std::numeric_limits<std::size_t>::max() / max_untrusted_compression_ratio <= B)
         B = std::numeric_limits<std::size_t>::max();
       else
-        B *= 4;
+        B *= max_untrusted_compression_ratio;
 
       return c.reserve(std::min(N, B));
     }


```

vtnerd: Updating, but switching from `_v` t o `::value` as this is going to be identical on the 0.18 branch.

vtnerd:
Force pushed @jeffro256 proposed changes.

src/serialization/container.h
vtnerd: I also just realized should be `B /= sizeof(T); B = std::max(1, B);`

vtnerd:
Force pushed another tweak, hopefully final version.

jeffro256 (approved):

src/serialization/container.h
```diff
@@ -76,7 +80,26 @@ namespace serialization
...
     template<typename C>
-    auto do_reserve(C &c, std::size_t N) -> decltype(c.reserve(N)) { return c.reserve(N); }
+    auto do_reserve(C &c, std::size_t N, std::size_t B) -> decltype(c.reserve(N))
+    {
+      using T = typename C::value_type;
+
+      static constexpr std::size_t max_compression_ratio =
+        is_blob_type<T>::type::value ? 1 :
+        use_container_varint<T>() ? sizeof(T) :
+        (std::is_same<T, char>::value && std::is_same<T, unsigned char>::value) ? 1:
```
selsta:
```suggestion
        (std::is_same<T, char>::value || std::is_same<T, unsigned char>::value) ? 1:
```

vtnerd: That's just embarrassing sorry about that.

selsta (approved):

PR #10381, opened 2026-03-25
ringct: make `h2d()` fallible

jeffro256:
One theoretically could have crafted a "long amount" (pre-v10) RingCT transaction with non-0 padding bytes in the decoded amount which fails receiver scanning (doesn't truncate decoded amount), but passes third-party wallet proof checking (does truncate decoded amount).

### DIFF
--- a/src/ringct/rctOps.cpp
+++ b/src/ringct/rctOps.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2024, Monero Research Labs
+// Copyright (c) 2016-2026, Monero Research Labs
 //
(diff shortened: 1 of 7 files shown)


src/ringct/rctOps.cpp
```diff
@@ -30,7 +30,11 @@
 
 #include <boost/lexical_cast.hpp>
 #include "misc_log_ex.h"
+extern "C" {
```
vtnerd, line 33: Wish this was done in the header. :/

vtnerd (approved):

selsta:
Needs a rebase.

selsta (approved):

PR #10403, opened 2026-04-05
multisig: add flag to skip refresh after multisig import

woodser:
Currently, `import_multisig` triggers an immediate, blocking `refresh` and `rescan_spent` after import.

On high-latency connections (e.g., Tor), this often causes the RPC call to hang or timeout while waiting for the scan to complete.

Furthermore, it prevents clients from using their own sync routines to report progress or sync in the background so the wallet is not blocked.

This PR adds an optional flag to `import_multisig` to decouple the data import from post-processing. This allows callers to ingest multisig data immediately and manage the subsequent resync on their own, providing a more responsive and transparent user experience.

Release PR: https://github.com/monero-project/monero/pull/10402

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -14631,7 +14631,7 @@ void wallet2::update_multisig_rescan_info(const std::vector<std::vector<rct::key
   m_key_images[td.m_key_image] = n;
 }
 //----------------------------------------------------------------------------------------------------
-size_t wallet2::import_multisig(std::vector<cryptonote::blobdata> blobs)
+size_t wallet2::import_multisig(std::vector<cryptonote::blobdata> blobs, bool refresh_after_import)
 {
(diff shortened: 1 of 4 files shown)


j-berman (approved):
I think this is ok as a holdover until the following change is upstreamed: https://github.com/seraphis-migration/monero/pull/291/changes/27db6ba86e3a9b5eb284999304dd819f6c9cc039

That change renders the refresh unnecessary.

I think we want to wait to upstream that commit until release v0.19 because release v0.19 is slated to trigger a wallet rescan for **all** wallets on load (so that wallets can sync the FCMP++ curve tree). If that commit is introduced to a v0.18 multisig wallet that has already identified received outputs but not yet called `import_multisig` for those outputs, then the wallet could enter a broken state without further changes (wallets pointing to untrusted daemons / GUI may miss when the outputs were spent). The wallet would need to re-sync from its restore height to avoid that state, which release v0.19 would do on load.

tobtoht (approved):

amin404ala (approved):

PR #10411, opened 2026-04-13
Backport SOCKS5 IPv6 support to release-v0.18

lschomaker1:
This is a backport of monero-project/monero pull request 9443 to allow Haveno to ship binaries with IPv6 daemon connections via SOCKS5 proxy

### DIFF
--- a/src/net/socks.cpp
+++ b/src/net/socks.cpp
@@ -31,6 +31,7 @@
 #include <algorithm>
 #include <boost/asio/bind_executor.hpp>
 #include <boost/asio/buffer.hpp>
+#include <boost/asio/coroutine.hpp>
 #include <boost/asio/dispatch.hpp>
 #include <boost/asio/read.hpp>
 #include <boost/asio/write.hpp>
@@ -38,8 +39,10 @@
 #include <boost/endian/conversion.hpp>
 #include <cstring>
 #include <limits>
+#include <numeric>
 #include <string>
 
+#include "net/parse.h"
 #include "net/net_utils_base.h"
 #include "net/tor_address.h"
 #include "net/i2p_address.h"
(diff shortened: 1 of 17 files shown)


nahuhh:
Doesnt build

lschomaker1:
yes i am addressing this now - currently have a successful build via docker ubuntu 20.04

.github/workflows/build.yml
```diff
@@ -84,6 +84,11 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: archlinux:latest
+    env:
+      # Arch's rolling protobuf/abseil packages now require C++17, while
+      # release-v0.18 intentionally builds as C++14. Trezor is not relevant to
+      # this job's coverage, so avoid the noisy protobuf probe there.
+      BUILD_DEFAULT_LINUX: 'cmake -S . -B build -D ARCH="default" -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=Release -D USE_DEVICE_TREZOR=OFF && cmake --build build --target all && cmake --build build --target wallet_api'
```
selsta: Are these changes here intentional?

lschomaker1: Yes. This is intentional and scoped to the Arch CI job only.

  Arch’s current protobuf/abseil packages now require C++17 through the
  optional Trezor dependency path, while release-v0.18 intentionally still
  builds as C++14. This backport does not touch or depend on Trezor, so I
  disabled `USE_DEVICE_TREZOR` for that Arch job to keep CI focused on the
  SOCKS5/IPv6 backport instead of failing on an unrelated rolling-distro
  dependency issue.

  I’m fine splitting or dropping this workflow change if you would prefer to
  keep CI config untouched, but it was needed to get the current Arch job
  green against release-v0.18.

nahuhh: but the arch workflow is passing for other PRs(?)

src/net/socks.cpp
```diff
@@ -190,22 +460,179 @@ namespace socks
...
+                    if (header.type == v5_ipv4_type)
+                        bytes = sizeof(v5_response_ipv4);
+                    else if (header.type == v5_ipv6_type)
+                        bytes = sizeof(v5_response_ipv6);
+                    else
+                    {
+                        self.done(socks::error::unexpected_version, self_);
+                        return;
+                    }
+                }
```
jeetrex17, line 614: Small question: would `address_type_not_supported` be a better fit here than `unexpected_version`? This looks more like an `ATYP` mismatch than a version issue, unless I’m missing some intended meaning behind the current error.

selsta: Since this is a backport it's not something we can change now, any changes would need to happen later separately.

jeetrex17: Makes sense, thanks

selsta (approved):

vtnerd (approved):
I did a diff against mainline, and this appears to be a back port in good faith.

My only gripe is the general policy of _not_ backporting features. I leave it to others to decide if an exception should be made for this use case.

lschomaker1:
thanks guys, be safe everyone

jhonharrisricauerta2-ops (approved):

PR #10417, opened 2026-04-17
epee: add http bin limit

selsta:
This can't be exploited currently due to existing 1 MB RPC limit, but adding the limits can't hurt.


vtnerd (commented):
Is this expectation that these hard limits are never reached due to the payload limit?

jpk68 (approved):

vtnerd (approved):

PR #10421, opened 2026-04-18
device: enable support for Ledger Speculos

jeffro256:
[Speculos](https://github.com/LedgerHQ/speculos) is an emulator for Ledger hardware devices. This PR adds support for Ledger firmware functional tests through Speculos directly from the C++ codebase.

Steps to get it working:
  - Build and install Speculos on your system, preferably in a Python virtual environment
     * To build Speculos from source, especially on Arch Linux, you may have to use my fork of Speculos here: https://github.com/LedgerHQ/speculos/pull/654
  - Compile the [Monero firmware](https://github.com/LedgerHQ/app-monero)
  - Compile this PR
  - Launch Speculos with the Monero firmware binary, with default ports
  - Run the `ledger.*` unit tests

The tests pass on my system, but I want to add some more, and maybe move it's own directory. This code will be useful for testing Ledger support for FCMP++.

Depends on #10429

### DIFF
--- a/src/device/device_io_tcp.cpp
+++ b/src/device/device_io_tcp.cpp
@@ -1,0 +1,238 @@
+// Copyright (c) 2026, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+//paired header
+#include "device_io_tcp.hpp"
+
+//local headers
+#include "int-util.h"
+#include "misc_log_ex.h"
+#include "net/net_helper.h"
+
+//third party headers
+#include <array>
+#include <boost/asio/buffer.hpp>
+#include <boost/asio/error.hpp>
+#include <boost/asio/steady_timer.hpp>
+#include <chrono>
+
+//standard headers
+
+#undef MONERO_DEFAULT_LOG_CATEGORY
+#define MONERO_DEFAULT_LOG_CATEGORY "device.io"
+
+namespace hw {
+namespace io {
+//-------------------------------------------------------------------------------------------------------------------
+device_io_tcp::device_io_tcp(boost::asio::io_context &io_context):
+    m_io_context(io_context),
+    m_socket(m_io_context),
+    m_timeout(DEFAULT_TIMEOUT)
+{}
+//-------------------------------------------------------------------------------------------------------------------
+device_io_tcp::~device_io_tcp()
+{}
+//-------------------------------------------------------------------------------------------------------------------
+void device_io_tcp::init()
+{
+    // no global setup
+}
+//-------------------------------------------------------------------------------------------------------------------
+void device_io_tcp::connect(void *params)
+{
+    static const conn_params_t conn_params{}; // default params
+    const conn_params_t &params_cref = params ? reinterpret_cast<const conn_params_t &>(params) : conn_params;
+    this->connect(params_cref);
+}
+//-------------------------------------------------------------------------------------------------------------------
+bool device_io_tcp::connected() const
+{
+    return this->m_socket.is_open();
+}
+//-------------------------------------------------------------------------------------------------------------------
+int  device_io_tcp::exchange(
+    unsigned char * const command,
+    unsigned int cmd_len,
+    unsigned char *response,
+    unsigned int max_resp_len,
+    const bool user_input)
+{
+    boost::asio::steady_timer deadline(this->m_io_context);
+    boost::system::error_code ec = boost::asio::error::would_block;
+    std::size_t bytes_transfered = 0;
+    const auto io_cb = [&](const boost::system::error_code ec_, const std::size_t bytes_transfered_)
+    {
+        ec = ec_;
+        bytes_transfered = bytes_transfered_;
+    };
+    const auto block_io = [&ec, this]()
+    {
+        while (ec == boost::asio::error::would_block)
+        {
+            this->m_io_context.restart();
+            this->m_io_context.run_one();
+        }
+    };
+    const auto set_deadline = [&ec, &deadline, this]()
+    {
+        deadline.expires_after(std::chrono::seconds(this->m_timeout));
+        deadline.async_wait([&ec](const boost::system::error_code ec_){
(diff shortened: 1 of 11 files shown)


jeffro256:
I need to check if Speculos can run in a headless environment, but if so, I think that it should be possible. The CI would need to download the GNU ARM EABI Hard Float ABI toolchain, which is pretty hefty. I think it would also need QEMU set up. Speculos pulls in a ton of code to compile, and a lot of Python packages, but once it's built, it's not that heavy to run. IIRC, building Speculos took several minutes. Compiling the Monero firmware is pretty fast.

selsta:
Removed my comment since I remembered we also don't run Trezor monero tests in CI, I guess if possible both can be added at a later point.

src/device/device_io_tcp.cpp
```diff
@@ -0,0 +1,238 @@
...
+//-------------------------------------------------------------------------------------------------------------------
+void device_io_tcp::init()
+{
+    // no global setup
+}
+//-------------------------------------------------------------------------------------------------------------------
+void device_io_tcp::connect(void *params)
+{
+    static const conn_params_t conn_params{}; // default params
+    const conn_params_t &params_cref = params ? reinterpret_cast<const conn_params_t &>(params) : conn_params;
```
selsta:
```suggestion
    const conn_params_t &params_cref = params ? *static_cast<const conn_params_t *>(params) : conn_params;
```

src/device/device_io_tcp.cpp
```diff
@@ -0,0 +1,238 @@
...
+    unsigned char *response,
+    unsigned int max_resp_len,
+    const bool user_input)
+{
+    boost::asio::steady_timer deadline(this->m_io_context);
+    boost::system::error_code ec = boost::asio::error::would_block;
+    std::size_t bytes_transfered = 0;
+    const auto io_cb = [&](const boost::system::error_code ec_, const std::size_t bytes_transfered_)
+    {
+        ec = ec_;
```
selsta, line 90: This timeout handling looks unsafe, the timer only changes local state and does not cancel the pending socket operation, so an async callback could run after `exchange()` returns and access dangling stack references.

src/device/device_io_tcp.cpp
```diff
@@ -0,0 +1,238 @@
...
+            -1, "Transferred the wrong number of bytes to ADPU TCP: " << bytes_transfered
+                << " vs " << (cmd_len + sizeof(uint32_t)));
+    }
+
+    // do first read, waiting indefinitely if user_input=true
+    std::uint32_t n_read_bytes = -1;
+    {
+        ec = boost::asio::error::would_block;
+        bytes_transfered = 0;
+        if (!user_input)
```
selsta, line 135: `user_input=true` may still time out because the previous operation's deadline timer is not cancelled.

src/device/device_io_tcp.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+    return n_read_bytes;
+}
+//-------------------------------------------------------------------------------------------------------------------
+void device_io_tcp::disconnect()
+{
+    if (this->m_socket.is_open())
+    {
+        this->m_socket.shutdown(boost::asio::socket_base::shutdown_both);
+        this->m_socket.close();
```
selsta, line 198: Use `boost::system::error_code ec;` overloads as this should not throw.

tests/ledger/ledger_tests_utils.cpp
```diff
@@ -0,0 +1,163 @@
...
+    if (speculos)
+    {
+        std::string speculos_host = hw::io::device_io_tcp::DEFAULT_ADDRESS;
+        unsigned int speculos_port = hw::io::device_io_tcp::DEFAULT_PORT;
+        if (speculos_address.size())
+        {
+            epee::net_utils::http::url_content url_content;
+            if (!epee::net_utils::parse_url(speculos_address, url_content))
+                throw std::runtime_error("Could not parse Speculos URL: " + speculos_address);
+            speculos_host = url_content.host;
```
selsta, line 74: It later calls `hw_device_io->connect(nullptr)` so the `speculos_address` is ignored.

src/device/device_io_tcp.cpp
```diff
@@ -0,0 +1,238 @@
...
+        const std::uint32_t cmd_len_be = SWAP32BE(static_cast<std::uint32_t>(cmd_len));
+        const std::array<boost::asio::const_buffer, 2> send_buffers{{
+            {&cmd_len_be, sizeof(cmd_len_be)},
+            {command, cmd_len}
+        }};
+        boost::asio::async_write(this->m_socket, send_buffers, io_cb);
+        block_io();
+
+        // check failure
+        CHECK_AND_ASSERT_MES(!ec, -1, "Got error code while writing ADPU to TCP: " << ec);
```
selsta:
```suggestion
        CHECK_AND_ASSERT_MES(!ec, -1, "Got error code while writing APDU to TCP: " << ec);
```

src/device/device_io_tcp.cpp
```diff
@@ -0,0 +1,238 @@
...
+        CHECK_AND_ASSERT_MES(bytes_transfered == cmd_len + sizeof(uint32_t),
+            -1, "Transferred the wrong number of bytes to ADPU TCP: " << bytes_transfered
```
selsta:
```suggestion
            -1, "Transferred the wrong number of bytes to APDU TCP: " << bytes_transfered
```

src/device/device_io_tcp.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+    // do first read, waiting indefinitely if user_input=true
+    std::uint32_t n_read_bytes = -1;
+    {
+        ec = boost::asio::error::would_block;
+        bytes_transfered = 0;
+        if (!user_input)
+            set_deadline();
+
+        // read in [response payload length as big-endian 32-bit int, beginnig of payload...]
```
selsta:
```suggestion
        // read in [response payload length as big-endian 32-bit int, beginning of payload...]
```

src/device/device_io_tcp.cpp
```diff
@@ -0,0 +1,238 @@
...
+int  device_io_tcp::exchange(
+    unsigned char * const command,
+    unsigned int cmd_len,
+    unsigned char *response,
+    unsigned int max_resp_len,
+    const bool user_input)
+{
+    boost::asio::steady_timer deadline(this->m_io_context);
+    boost::system::error_code ec = boost::asio::error::would_block;
+    std::size_t bytes_transfered = 0;
```
selsta:
```suggestion
    std::size_t bytes_transferred = 0;
```

src/device/device_io_tcp.cpp
```diff
@@ -0,0 +1,238 @@
...
+    this->connect(params_cref);
+}
+//-------------------------------------------------------------------------------------------------------------------
+bool device_io_tcp::connected() const
+{
+    return this->m_socket.is_open();
+}
+//-------------------------------------------------------------------------------------------------------------------
+int  device_io_tcp::exchange(
+    unsigned char * const command,
```
jpk68, line 79: Nit: formatting of pointer types is a bit inconsistent

src/device/device_io_tcp.cpp
```diff
@@ -0,0 +1,238 @@
...
+    unsigned int max_resp_len,
+    const bool user_input)
+{
+    boost::asio::steady_timer deadline(this->m_io_context);
+    boost::system::error_code ec = boost::asio::error::would_block;
+    std::size_t bytes_transferred = 0;
+    const auto io_cb = [&](const boost::system::error_code ec_, const std::size_t bytes_transfered_)
+    {
+        ec = ec_;
+        bytes_transfered = bytes_transfered_;
```
selsta, line 91:
```suggestion
        bytes_transferred = bytes_transfered_;
```

and in other places

src/device/device_ledger.cpp
```diff
@@ -479,7 +506,7 @@ namespace hw {
     unsigned int device_ledger::exchange(unsigned int ok, unsigned int mask) {
       logCMD();
 
-      this->length_recv =  hw_device.exchange(this->buffer_send, this->length_send, this->buffer_recv, BUFFER_SEND_SIZE, false);
+      this->length_recv =  hw_device_io->exchange(this->buffer_send, this->length_send, this->buffer_recv, BUFFER_SEND_SIZE, false);
```
selsta, line 509: `exchange` can return `-1` on failure, this needs to be checked. Same for `exchange_wait_on_input()`.

tests/ledger/basic_signing.cpp
```diff
@@ -0,0 +1,90 @@
...
+    ASSERT_TRUE(alice.get_device().set_mode(hw::device::TRANSACTION_CREATE_REAL));
+    ASSERT_TRUE(cryptonote::construct_tx_and_get_tx_key(alice, alice_subaddresses, srcs, dests,
+        alice_addr, {}, tx, tx_key, additional_tx_keys, true, {rct::RangeProofPaddedBulletproof, 0}, true));
+    ASSERT_EQ(2, tx.version);
+    ASSERT_EQ(1, tx.vin.size());
+    ASSERT_EQ(2, tx.vout.size());
+    ASSERT_EQ(rct::RCTTypeBulletproofPlus, tx.rct_signatures.type);
+    ASSERT_EQ(fee, tx.rct_signatures.txnFee);
+    cryptonote::tx_verification_context tvc{};
+    ASSERT_TRUE(cryptonote::ver_non_input_consensus(tx, tvc, HF_VERSION_BULLETPROOF_PLUS));
```
selsta, line 88: Input proof should also be checked here.

tests/CMakeLists.txt
```diff
@@ -68,6 +68,11 @@ if (BUILD_GUI_DEPS)
   add_subdirectory(libwallet_api_tests)
 endif()
 
+option(LEDGER_DEBUG "Ledger debugging enabled" OFF)
+if (LEDGER_DEBUG)
+  add_subdirectory(ledger)
```
selsta, line 73: This should also check if monero is compiled with Ledger support.

PR #10422, opened 2026-04-19
rpc: fix incomplete range check in get_coinbase_tx_sum + update test

thomasbuilds:
The `on_get_coinbase_tx_sum` RPC handler validates `req.height` and `req.count` independently against the blockchain height, but does not validate their combined range. This allows requests where `height + count` exceeds the blockchain height to pass validation.

For example, with `bc_height=100`:
- `req.height=90, req.count=20` passes both checks (`90 < 100` and `20 <= 100`)
- But `get_coinbase_tx_sum(90, 20)` computes `end = 90 + 20 - 1 = 109` and calls `for_blocks_range(90, 109)`, reading past the end of the chain

The fix changes the second condition from `req.count > bc_height` to `req.count > bc_height - req.height`. The subtraction is safe from underflow because the first condition (`req.height >= bc_height`) short-circuits via `||`, guaranteeing `bc_height > req.height` when the second condition is evaluated.

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -3076,9 +3076,9 @@ namespace cryptonote
   {
     RPC_TRACKER(get_coinbase_tx_sum);
     const uint64_t bc_height = m_core.get_current_blockchain_height();
-    if (req.height >= bc_height || req.count > bc_height)
+    if (req.height >= bc_height || req.count > bc_height - req.height)
     {
-      res.status = "height or count is too large";
+      res.status = "requested range exceeds blockchain height";
       return true;
     }
     CHECK_PAYMENT_MIN1(req, res, COST_PER_COINBASE_TX_SUM_BLOCK * req.count, false);
--- a/tests/functional_tests/blockchain.py
+++ b/tests/functional_tests/blockchain.py
@@ -219,9 +219,12 @@ class BlockchainTest():
         assert res.emission_amount < extrapolated and res.emission_amount > extrapolated - 1e12
         assert res.fee_amount == 0
         sum_blocks_emission = res.emission_amount
-        res = daemon.get_coinbase_tx_sum(1, sum_blocks)
+        res = daemon.get_coinbase_tx_sum(1, sum_blocks - 1)
         assert res.emission_amount == sum_blocks_emission - 17592186044415
         assert res.fee_amount == 0
+        # height + count must not exceed the blockchain height
+        res = daemon.get_coinbase_tx_sum(1, sum_blocks)
+        assert res.status != 'OK'
 
         res = daemon.get_output_distribution([0, 1, 17592186044415], 0, 0)
         assert len(res.distributions) == 3


plowsof:
looks sane, current behaviour when count exceeds blockchain tip is to just return the sum from req.height to tip ~~(which at least obscures any indexed from 0)~~

~~i think an issue existed and not fixed with this:  blockchain is indexed from 0, if i request from height 0 and count is all aka "tip + 1"~~


Ignore^ in cryptonote_core.cpp we see `const uint64_t end = start_offset + count - 1;`

plowsof:
failing test looks related?
```
[TEST STARTED] blockchain
Resetting blockchain
Test generating 5 blocks
  File "/__w/monero/monero/tests/functional_tests/blockchain.py", line 345, in <module>
    BlockchainTest().run_test()
  File "/__w/monero/monero/tests/functional_tests/blockchain.py", line 50, in run_test
    self._test_generateblocks(5)
  File "/__w/monero/monero/tests/functional_tests/blockchain.py", line 223, in _test_generateblocks
    assert res.emission_amount == sum_blocks_emission - 17592186044415
AssertionError
[TEST FAILED] blockchain
```

thomasbuilds:
Yep, line 222 in blockchain.py calls `get_coinbase_tx_sum(1, sum_blocks)` which requests blocks 1 through 5, but block 5 doesn't exist.

The old code silently returned blocks 1–4 with `status: OK`, and the test assertion was written around that truncated result.

selsta (approved):

plowsof (approved):

src/rpc/core_rpc_server.cpp
```diff
@@ -3076,7 +3076,7 @@ namespace cryptonote
   {
     RPC_TRACKER(get_coinbase_tx_sum);
     const uint64_t bc_height = m_core.get_current_blockchain_height();
-    if (req.height >= bc_height || req.count > bc_height)
+    if (req.height >= bc_height || req.count > bc_height - req.height)
     {
       res.status = "height or count is too large";
```
selsta:
```suggestion
      res.status = "requested range exceeds blockchain height";
```

A bit clearer error message

tests/functional_tests/blockchain.py
```diff
@@ -219,7 +219,7 @@ def _test_generateblocks(self, blocks):
         assert res.emission_amount < extrapolated and res.emission_amount > extrapolated - 1e12
         assert res.fee_amount == 0
         sum_blocks_emission = res.emission_amount
-        res = daemon.get_coinbase_tx_sum(1, sum_blocks)
+        res = daemon.get_coinbase_tx_sum(1, sum_blocks - 1)
```
selsta, line 222: Would it make sense to also test `res = daemon.get_coinbase_tx_sum(1, sum_blocks)` now to assert it fails?

thomasbuilds: ok I added this

selsta (approved):

PR #10424, opened 2026-04-19
zmq: add more missing restricted rpc checks

selsta:
I duplicated the variables from `core_rpc_server.cpp`, which would allow for different limits for HTTP and ZMQ. Alternatively I can move them out into a separate file and avoid duplication.

### DIFF
--- a/src/rpc/daemon_handler.cpp
+++ b/src/rpc/daemon_handler.cpp
@@ -30,6 +30,7 @@
 #include "rpc/zmq_restricted_methods.h"
 
 #include <algorithm>
+#include <chrono>
 #include <cstring>
 #include <stdexcept>
 
@@ -42,6 +43,14 @@
 #include "cryptonote_basic/blobdatatype.h"
 #include "ringct/rctSigs.h"
 #include "version.h"
+
+namespace
+{
+constexpr size_t restricted_max_fake_outs = 5000;
+constexpr auto restricted_histogram_cutoff = std::chrono::hours{3 * 24};
+constexpr size_t restricted_max_txs = 100;
+constexpr size_t restricted_max_key_images = 5000;
+}
 
 namespace cryptonote
 {
@@ -236,6 +245,13 @@ namespace rpc
 
   void DaemonHandler::handle(const GetTransactions::Request& req, GetTransactions::Response& res)
   {
+    if (m_restricted && req.tx_hashes.size() > restricted_max_txs)
+    {
+      res.status = Message::STATUS_FAILED;
+      res.error_details = "Too many transactions requested in restricted mode";
+      return;
+    }
+
     std::vector<cryptonote::transaction> found_txs_vec;
     std::vector<crypto::hash> missed_vec;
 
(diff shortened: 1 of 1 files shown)


src/rpc/daemon_handler.cpp
```diff
@@ -43,6 +43,13 @@
 #include "ringct/rctSigs.h"
 #include "version.h"
 
+#define ZMQ_MAX_RESTRICTED_GLOBAL_FAKE_OUTS_COUNT 5000
```
vtnerd: I would just use `constexpr` variables here (instead of `#DEFINE`) as it will trigger a compilation error instead of overwriting some symbol on name clash. The names could also be shorter with this change.

src/rpc/daemon_handler.cpp
```diff
@@ -43,6 +43,13 @@
...
+
+#define ZMQ_OUTPUT_HISTOGRAM_RECENT_CUTOFF_RESTRICTION (3 * 86400) // 3 days max, the wallet requests 1.8 days
```
vtnerd: This could be a `constexpr std::chrono::hours histogram_cutoff{3 * 24};`. Unfortunately `chrono::days` isn't available until C++20.

vtnerd (changes requested):

src/rpc/daemon_handler.cpp
selsta: Updated with your suggestions.

SNeedlewoods (approved):
Confirmed the checks match the ones in `src/rpc/core_rpc_server.cpp`. I have to admit I don't know much about the surrounding code.

src/rpc/daemon_handler.cpp
```diff
@@ -791,6 +815,22 @@ namespace rpc
...
   {
+    size_t amounts = req.amounts.size();
+    if (m_restricted && amounts == 0)
+    {
+      res.status = Message::STATUS_FAILED;
+      res.error_details = "Restricted RPC will not serve histograms on the whole blockchain. Use your own node.";
+      return;
+    }
+
+    const auto cutoff = static_cast<uint64_t>(std::time(nullptr) - std::chrono::duration_cast<std::chrono::seconds>(restricted_histogram_cutoff).count());
```
vtnerd: Yikes maybe the C version was better. The pure C++ approach:
```c++
using clock = std::chrono::system_clock;
const clock::time_point cutoff{std::chrono::seconds{req.recent_cutoff}};
if (m_restricted && clock::now() - cutoff > restricted_histogram_cutoff)
```

The system clock is guaranteed to be 1970 epoch starting in C++20; time_t is still unspecified but almost always the same 1970 epoch.

vtnerd (changes requested):

src/rpc/daemon_handler.cpp
tobtoht: Apart from fixing some compilation issues, I don't think there is anything stopping us from switching to C++20 on `master`. We can bump the minimum required GCC to 10 without breaking builds on the oldest operating systems that we still support.

src/rpc/daemon_handler.cpp
selsta: Updated to your suggestion. We already use `std::chrono::system_clock` in other places.

vtnerd (approved):

PR #10429, opened 2026-04-21
ledger: lazy view key loading

jeffro256:
Defer loading of private view key from device until first time it is needed. Do not fail if this fails. This has two effects:

1. The prompt to export the view key is only needed once when creating a `cryptonote::account_base` (it shows twice currently)
2. The call to `connect()` doesn't fail if the user decides to not export the viewkey, and thus usage of the device without exporting view keys is possible

This is a small convenience for Ledger users, but will be an even larger convenience for Ledger device testing

### DIFF
--- a/src/device/device_ledger.cpp
+++ b/src/device/device_ledger.cpp
@@ -303,6 +303,7 @@ namespace hw {
       this->reset_buffer();      
       this->mode = NONE;
       this->has_view_key = false;
+      this->requested_view_key = false;
       this->tx_in_progress = false;
       MDEBUG( "Device "<<this->id <<" Created");
     }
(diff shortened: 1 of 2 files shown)


jeffro256:
Fixed merge conflicts.

jeffro256:
Unit test failure is unrelated

selsta:
I'm not sure if this is working correctly for me.

```
~/d/m/b/D/m/r/bin (master)> ./monero-wallet-cli --wallet-file ~/Monero/wallets/ledger
This is the command line monero wallet. It needs to connect to a monero
daemon to work correctly.
WARNING: Do not reuse your Monero keys on another fork, UNLESS this fork has key reuse mitigations built in. Doing so will harm your privacy.

Monero 'Fluorine Fermi' (v0.18.1.0-cb7d2ba94)
Logging to ./monero-wallet-cli.log
Wallet password: 
Opened wallet: <redacted>
Wallet is on device: Ledger
**********************************************************************
Use the "help" command to see a simplified list of available commands.
Use "help all" to see the list of all available commands.
Use "help <command>" to see a command's documentation.
**********************************************************************
Background mining not enabled. Run "set setup-background-mining 1" to change.
Starting refresh...
```

Afterwards it took almost a minute until the view key message showed up and it started refresh. Is this intended?

jeffro256:
Rebased and fixed issue that @selsta mentioned. When multiple threads call `soft_request_view_key()`, `requested_view_key` was set to `true` on the first thread before I/O was issued, and thus `has_view_view` was set to true, which caused the other threads to do the slow key derivation command, instead of using the cached view key until the first thread completed. Adding a lock guard to `soft_request_view_key()` fixes this by forcing the threads *not* requesting the view key to wait until the first request completes.

selsta (approved):
Confirmed working with monero-wallet-cli and gui.

tobtoht (approved):

PR #10430, opened 2026-04-21
cryptonote_core: add change address sanity check

jeffro256:
Prevents silly mistakes where wrong change address is passed

PR #10431, opened 2026-04-21
cryptonote_core: add change address sanity check [RELEASE]

jeffro256:
#10430

### DIFF
--- a/src/cryptonote_core/cryptonote_tx_utils.cpp
+++ b/src/cryptonote_core/cryptonote_tx_utils.cpp
@@ -31,6 +31,7 @@
 #include <unordered_set>
 #include <random>
 #include "include_base_utils.h"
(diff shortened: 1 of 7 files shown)


src/cryptonote_core/cryptonote_tx_utils.cpp
```diff
@@ -28,9 +28,11 @@
 // 
 // Parts of this file are originally copyright (c) 2012-2013 The Cryptonote developers
 
+#include <optional>
```
selsta: can be dropped for C++14 version

jeffro256: Done in latest, thanks

selsta:
Seems there is a compilation issue on Arch related to optional

selsta (approved):

PR #10442, opened 2026-04-22
refactor(core): stabilize rpc_client templating and request flow

nullcryptodev:
- created a custom RPC client wrapper (t_rpc_client) for a Monero-based daemon
- reworked RPC handling into a strongly typed system using templates (T_req, T_res)
- introduced a unified response structure:
  - rpc_result<T>
  - includes OK, BUSY, NETWORK_ERROR, RPC_ERROR, TIMEOUT

---

retry + reliability layer
- added a retry system (execute_with_retry)
- implemented:
  - max retry limits
  - exponential backoff
  - jitter (randomized delay)
- built automatic handling for:
  - daemon BUSY responses
  - transient network failures
  - timeout retries

---

RPC abstraction layer
- wrapped both RPC styles:
  - /json_rpc (JSON-RPC calls)
  - REST-style HTTP RPC calls
- standardised request flow:
  - request struct > HTTP call > response struct > rpc_result

### DIFF
--- a/src/common/rpc_client.h
+++ b/src/common/rpc_client.h
@@ -29,6 +29,9 @@
 #pragma once
 
 #include <boost/optional/optional.hpp>
+#include <thread>
+#include <chrono>
+#include <random>
 
 #include "common/http_connection.h"
 #include "common/scoped_message_writer.h"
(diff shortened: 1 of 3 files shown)


selsta:
Could you please explain why this change? Based on your PR description there are a lot of different changes mixed but it's not clear what the purpose is. Does it fix a specific bug? Or just generic improvement?

nullcryptodev:
> Could you please explain why this change? Based on your PR description there are a lot of different changes mixed but it's not clear what the purpose is. Does it fix a specific bug? Or just generic improvement?

This is just a general improvement. I was looking through the code for things marked todo's and this `// TODO - handle CORE_RPC_STATUS_BUSY ?` caught my eye. Also, i noticed a lot of inconsistencies with some returns for example, an error would still return true, which is misleading when trying to read it. So this is just a generic improvement mainly. I tested it on my machine, all seems good to me but I'd still test this yourselves

PR #10443, opened 2026-04-22
blockchain_utilities: fix `--data-dir` option

jeffro256:
Regression added in #10396 broke `--data-dir` for usages where the `--regtest` option wasn't added to the variables map. This is most of the blockchain utilities. They don't even work when `--data-dir` isn't specified by the user.

This PR adds `cryptonote::arg_regtest_on` to the blockchain utilities options so that they work again. It also dedups the CLI argument -> network type logic.


nahuhh:
mybad. Thanks

nahuhh:
cc @jeetrex17

jeetrex17:
Thanks for catching this, and really sorry for the regression. That was an oversight on my side in #10396.

vtnerd (approved):
Good cleanup too.

tobtoht (approved):

PR #10446, opened 2026-04-22
fix(tests): isolate monerod config in functional RPC tests

duyua9:
## Summary
- create a test-owned empty `monerod.conf` inside each functional test daemon data dir
- pass `--config-file` explicitly so functional RPC tests do not consume `~/.bitmonero/monerod.conf`

## Validation
- `python3 -m py_compile tests/functional_tests/functional_tests_rpc.py`
- local path check for per-daemon `monerod.conf` creation under a test-owned functional-tests-directory

### DIFF
--- a/tests/functional_tests/functional_tests_rpc.py
+++ b/tests/functional_tests/functional_tests_rpc.py
@@ -77,6 +77,9 @@ outputs = []
 ports = []
 
 for i in range(N_MONERODS):
+  # monerod's default --config-file is resolved from daemonizer::get_default_data_dir()
+  # rather than --data-dir, so pass an explicit empty config to keep these tests
+  # hermetic even when the developer has a local ~/.bitmonero/bitmonero.conf.
   monerod_data_dir = builddir + "/functional-tests-directory/monerod" + str(i)
   os.makedirs(monerod_data_dir, exist_ok = True)
   monerod_config_path = monerod_data_dir + "/monerod.conf"
--- a/tests/functional_tests/functional_tests_rpc.py
+++ b/tests/functional_tests/functional_tests_rpc.py
@@ -77,7 +77,12 @@ outputs = []
 ports = []
 
 for i in range(N_MONERODS):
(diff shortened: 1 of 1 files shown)


selsta:
Could you please explain why this change?

duyua9:
Thanks for asking.

My intent here was just to make these functional RPC tests less dependent on any machine-local monerod configuration.

The tests already use dedicated data directories, but they do not explicitly provide a config file. I was concerned that a local default config (for example under ~/.bitmonero/) could still influence the test daemons and make the environment less reproducible across different machines.

This change creates an empty config file inside each test-owned daemon directory and passes it explicitly, so the test setup is fully owned by the test itself.

That said, I may be overestimating the risk here. If monerod does not actually pick up user-local config in this test setup, or if this extra isolation is not useful in practice, I’m happy to close the PR. Thanks again for pointing it out.

selsta:
From what I can tell the existing code already set a data dir for every daemon, so no default config in `~/.bitmonero` can be picked up.

duyua9:
Thanks — I double-checked this because that is exactly the isolation point I was trying to verify.

The functional test does pass `--data-dir`, but `monerod` resolves and parses `--config-file` before it later derives the daemon data directory. In `src/daemon/main.cpp`, `daemon_args::arg_config_file` is read and parsed before the later `cryptonote::arg_data_dir` handling. The default for `arg_config_file` is also built from `daemonizer::get_default_data_dir() / bitmonero.conf` in `src/daemon/command_line_args.h`; on POSIX that returns the tools default data dir, not the CLI `--data-dir` value.

So my understanding is that a developer-local default config can still be considered if it exists, even though the test daemons already use test-owned data directories. This patch makes that explicit by passing an empty test-owned config file alongside each test-owned data dir.

If I am missing a later code path that rebinds the default config path from `--data-dir`, I am happy to adjust or close this.

nahuhh:
I think the correct fix is to stop using a static datadir, and instead to (like bitcoin), if a data-dir is specified, look in the specified data-dir for the config file

doesnt make sense to me that it would ignore the --data-dir and keep looking in the default location

duyua9:
Thanks for the discussion. I am going to close this PR rather than keep a narrow functional-test workaround open.

After re-reading the feedback, I agree the better direction is not to paper over this in the test harness by writing an empty config file for each daemon. If the daemon accepts an explicit `--data-dir`, the more useful fix would be in the daemon config resolution itself: make the implicit config lookup follow that data dir instead of continuing to consider the static default config path.

That is a broader behavior change than this PR currently implements, and I do not want to spend maintainer review time on a narrower workaround that may not be the right abstraction. Thanks for pointing this out.

PR #10449, opened 2026-04-23
trezor: replace use of boost::is_base_of with std::is_base_of_v

tobtoht:
Extracted from #10222.

When building with Boost 1.91:

```
In file included from src/device_trezor/trezor/messages_map.cpp:30:
src/device_trezor/trezor/messages_map.hpp: In static member function 'static hw::trezor::messages::MessageType hw::trezor::MessageMapper::get_message_wire_number()':
src/device_trezor/trezor/messages_map.hpp:75:34: error: 'is_base_of' is not a member of 'boost'; did you mean 'std::is_base_of'?
   75 |       BOOST_STATIC_ASSERT(boost::is_base_of<google::protobuf::Message, t_message>::value);
      |                                  ^~~~~~~~~~
```

Missing header include, but we might as well switch to std.

There are no other occurrences of `boost::is_base_of` in the code base.

### DIFF
--- a/src/device_trezor/trezor/debug_link.hpp
+++ b/src/device_trezor/trezor/debug_link.hpp
@@ -32,6 +32,7 @@
 
 #include "transport.hpp"
 #include "messages/messages-debug.pb.h"
+#include <type_traits>
 
 
 namespace hw {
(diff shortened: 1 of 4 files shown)


src/device_trezor/trezor/debug_link.hpp
```diff
@@ -60,7 +60,7 @@ namespace trezor {
         const boost::optional<messages::MessageType> &resp_type = boost::none,
         bool no_wait = false)
     {
-      BOOST_STATIC_ASSERT(boost::is_base_of<google::protobuf::Message, t_message>::value);
```
vtnerd, line 63: While here, perhaps switch to `static_assert` too?

vtnerd (approved):

src/device_trezor/trezor/debug_link.hpp
vtnerd: One other thing I just caught, we arguably need a `#include <type_traits>` in these headers, etc.

selsta (approved):

PR #10451, opened 2026-04-23
docs: add common build errors section for Linux

ArvinFarrelP:
This PR adds a "Common Build Errors (Linux)" section to improve developer experience.

It includes solutions for:
- Missing libunbound dependency
- Missing Boost dependency

These are common issues encountered during compilation, and providing quick fixes should help new contributors.

### DIFF
--- a/docs/COMPILING_DEBUGGING_TESTING.md
+++ b/docs/COMPILING_DEBUGGING_TESTING.md
@@ -2,7 +2,7 @@
 
 This document describes ways of compiling, debugging and testing efficiently for various use cases.
 The intended audience are developers, who want to leverage newly added tricks to Monero via `CMake`. The document will lower the entry point for these developers.
-Before reading this document, please consult section "Build instructions" in the main README.md. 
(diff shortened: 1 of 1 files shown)


tobtoht:
Dependencies for each distro are documented here: https://github.com/monero-project/monero/blob/master/README.md?plain=1#L190-L221

ArvinFarrelP:
Thanks for pointing that out.

This section is intended as a quick troubleshooting reference specifically for common build errors, so developers can quickly identify and resolve issues without going through the full dependency list in the README.

If this is considered redundant, I'm happy to adjust or remove the section.

nahuhh:
You're supposed to speak like a pirate

selsta:
A section like this can make sense if there are common build errors that result in frequent issue reports on GitHub. These two missing dependencies are not common build errors so I don't see much point in this PR.

ArvinFarrelP:
Thanks for the clarification, that makes sense.
I'm happy to close this PR or revise it if there are more relevant recurring issues to document.

PR #10458, opened 2026-04-24
net/i2p: support human-readable hostnames

nahuhh:
mirror of jpk68's [codeberg branch](https://codeberg.org/jpk68/monero/src/branch/i2p-addresses)

(Review comments will be responded to on matrix)

from https://codeberg.org/jpk68/monero/pulls/3 :
> Adds support for standard, human-readable I2P hostnames in addition to Base32 ones. It effectively fixes [this TODO](https://github.com/monero-project/monero/blob/master/src/net/i2p_address.cpp#L47C9-L47C46), and is intended to be built off of by a later [SAM](https://i2p.net/en/docs/api/samv3/) integration PR. Some improvements have also been made to general hostname validation logic, in terms of conforming to [naming rules](https://i2p.net/en/docs/overview/naming/#naming-rules) and the like.
> 
> An example: `foo.i2p` may be used instead of `xmrdoc6k4licya7zsnba5s7vugzo72blcpukkvqno75o4ruhaqla.b32.i2p`.
> 
> To my knowledge, the changes made here do not render existing I2P documentation for Monero (i.e. [this](https://github.com/monero-project/monero/blob/master/docs/ANONYMITY_NETWORKS.md)) outdated; I do plan on updating the docs on a later date, however modifications can instead be made sooner if necessary.

### DIFF
--- a/src/net/i2p_address.cpp
+++ b/src/net/i2p_address.cpp
@@ -45,21 +45,16 @@ namespace net
 {
     namespace
     {
-        // !TODO only b32 addresses right now
-        constexpr const char tld[] = u8".b32.i2p";
-        constexpr const char unknown_host[] = "<unknown i2p host>";
-
-        constexpr const unsigned b32_length = 52;
-
         constexpr const char base32_alphabet[] =
             u8"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz234567";
 
-        expect<void> host_check(boost::string_ref host) noexcept
+        //! Validate a Base32 address (.b32.i2p)
+        expect<void> host_check_b32(boost::string_ref host) noexcept
         {
-            if (!host.ends_with(tld))
+            if (!host.ends_with(tld_b32))
                 return {net::error::expected_tld};
 
-            host.remove_suffix(sizeof(tld) - 1);
+            host.remove_suffix(sizeof(tld_b32) - 1);
 
             if (host.size() != b32_length)
                 return {net::error::invalid_i2p_address};
@@ -67,6 +62,55 @@ namespace net
                 return {net::error::invalid_i2p_address};
 
             return success();
+        }
+
+        //! Validate a human-readable hostname (.i2p)
+        expect<void> host_check_human(boost::string_ref host) noexcept
+        {
+            if (!host.ends_with(tld_i2p))
+                return {net::error::expected_tld};
+
+            host.remove_suffix(sizeof(tld_i2p) - 1);
+
+            if (host.empty() || host.size() > i2p_name_max_length)
+                return {net::error::invalid_i2p_address};
+
+            //! Reject addresses starting or ending with a dot or hyphen
+            if (host.front() == '.' || host.front() == '-' ||
+                host.back() == '.' || host.back() == '-')
+                return {net::error::invalid_i2p_address};
+
+            //! Reject invalid characters, and two dots/hyphens in a row (or a dot next to a hyphen)
+            char previous = '\0';
+            for (const char c : host)
+            {
+                if (!(('a' <= c && c <= 'z') || ('0' <= c && c <= '9') || c == '.' || c == '-'))
+                {
+                    return {net::error::invalid_i2p_address};
+                }
+                if ((c == '.' || c == '-') && (previous == '.' || previous == '-'))
+                {
+                    return {net::error::invalid_i2p_address};
+                }
+                previous = c;
+            }
+
+            return success();
+        }
+
+        //! Use appropriate function for address validation
+        expect<void> host_check(boost::string_ref host) noexcept
+        {
+            if (host.ends_with(tld_b32))
+            {
+                return host_check_b32(host);
+            }
+            else if (host.ends_with(tld_i2p))
+            {
+                return host_check_human(host);
+            }
+
+            return {net::error::expected_tld};
         }
 
         struct i2p_serialized
@@ -110,7 +154,9 @@ namespace net
         net::canonicalize_host(normalized_host);
         MONERO_CHECK(host_check(normalized_host));
 
-        static_assert(b32_length + sizeof(tld) == sizeof(i2p_address::host_), "bad internal host size");
+        static_assert(sizeof(i2p_address::host_) >= b32_length + sizeof(tld_b32) &&
+                      sizeof(i2p_address::host_) >= i2p_name_max_length + sizeof(tld_i2p));
+
         return i2p_address{normalized_host};
     }
 
--- a/src/net/i2p_address.h
+++ b/src/net/i2p_address.h
@@ -47,10 +47,27 @@ namespace serialization
 
 namespace net
 {
-    //! b32 i2p address; internal format not condensed/decoded.
+    /**
+     * Naming rules defined here:
+     * https://i2p.net/en/docs/overview/naming/#naming-rules
+     */
+    constexpr const char tld_b32[] = u8".b32.i2p";
+    constexpr const char tld_i2p[] = u8".i2p";
+    constexpr const char unknown_host[] = "<unknown i2p host>";
+
+    //! Length of a Base32 I2P address in characters, not including TLD
+    constexpr const unsigned b32_length = 52;
+
+    //! Maximum length of a human-readable I2P hostname in characters, not including TLD
+    constexpr const unsigned i2p_name_max_length = 63;
+
+    /**
+     * An I2P address; internal format not condensed/decoded.
+     * Supports human-readable and Base32 types.
+     */
     class i2p_address
     {
-        char host_[61]; // null-terminated
+        char host_[i2p_name_max_length + sizeof(tld_i2p)] = {}; // null-terminated
 
         //! Keep in private, `host.size()` has no runtime check
         i2p_address(boost::string_ref host) noexcept;
@@ -59,7 +76,7 @@ namespace net
         //! \return Size of internal buffer for host.
         static constexpr std::size_t buffer_size() noexcept { return sizeof(host_); }
 
-        //! \return `<unknown tor host>`.
+        //! \return `<unknown i2p host>`.
         static const char* unknown_str() noexcept;
 
         //! An object with `port() == 0` and `host_str() == unknown_str()`.
@@ -69,7 +86,7 @@ namespace net
         static i2p_address unknown() noexcept { return i2p_address{}; }
 
         /*!
-            Parse `address` in b32 i2p format (i.e. x.b32.i2p:80)
+            Parse `address` in the detected format (i.e. x[.b32].i2p:80)
             with `default_port` being used if port is not specified in
             `address`.
         */
@@ -96,10 +113,10 @@ namespace net
         //! \return True if i2p addresses are identical.
         bool is_same_host(const i2p_address& rhs) const noexcept;
 
-        //! \return `x.b32.i2p` or `x.b32.i2p:z` if `port() != 0`.
+        //! \return `x[.b32].i2p` or `x[.b32].i2p:z` if `port() != 0`.
         std::string str() const;
 
-        //! \return Null-terminated `x.b32.i2p` value or `unknown_str()`.
+        //! \return Null-terminated `x[.b32].i2p` value or `unknown_str()`.
         const char* host_str() const noexcept { return host_; }
 
         //! \return `1` to work with I2P socks which considers `0` error.
--- a/tests/unit_tests/net.cpp
+++ b/tests/unit_tests/net.cpp
@@ -1,5 +1,4 @@
 // Copyright (c) 2018-2024, The Monero Project
-
 //
 // All rights reserved.
 //
@@ -491,6 +490,9 @@ namespace
         "VWW6YBAL4BD7SZMGNCYRUUCPGFKQAHZDDI37KTCEO3AH7NGMCOPN.B32.I2P";
     static constexpr const char b32_i2p_2[] =
         "xmrto2bturnore26xmrto2bturnore26xmrto2bturnore26xmr2.b32.i2p";
+
+    static constexpr const char standard_i2p[] = "test.i2p";
+    static constexpr const char standard_i2p_2[] = "reg.i2p";
 }
 
 TEST(i2p_address, constants)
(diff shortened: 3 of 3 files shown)


src/net/i2p_address.h
```diff
@@ -136,4 +138,4 @@ namespace net
     {
         return lhs.less(rhs);
     }
-} // net
+} // namespace net
```
vtnerd: Unnecessary change, drop it.

src/net/i2p_address.cpp
```diff
@@ -171,4 +223,4 @@ namespace net
     {
         return host_str();
     }
-}
+} // namespace net
```
vtnerd: Also unnecessary.

src/net/i2p_address.cpp
```diff
@@ -78,7 +131,7 @@ namespace net
                 KV_SERIALIZE(port)
             END_KV_SERIALIZE_MAP()
         };
-    }
+    } // namespace
```
vtnerd: Drop this spurious change too.

src/net/i2p_address.cpp
```diff
@@ -107,7 +160,6 @@ namespace net
         boost::string_ref host = address.substr(0, address.rfind(':'));
         MONERO_CHECK(host_check(host));
 
-        static_assert(b32_length + sizeof(tld) == sizeof(i2p_address::host_), "bad internal host size");
```
vtnerd, line 113: If you're going to drop this, might want to add a runtime check somewhere that it doesn't exceed `host_` size.

vtnerd (changes requested):

src/net/i2p_address.cpp
vtnerd: Nevermind, it's in the other function. In that case, I would still static assert that the buffer size is larger than `i2p_max_length` and `b32_length`

src/net/i2p_address.h
```diff
@@ -96,10 +98,10 @@ namespace net
         //! \return True if i2p addresses are identical.
         bool is_same_host(const i2p_address& rhs) const noexcept;
 
-        //! \return `x.b32.i2p` or `x.b32.i2p:z` if `port() != 0`.
+        //! \return `x(.b32).i2p` or `x(.b32).i2p:z` if `port() != 0`.
```
nahuhh: Are these brackets meant to mean optional? If so, square them

```suggestion
        //! \return `x[.b32].i2p` or `x[.b32].i2p:z` if `port() != 0`.
```

src/net/i2p_address.cpp
```diff
@@ -171,4 +230,4 @@ namespace net
     {
         return host_str();
     }
-}
+} // net
```
nahuhh:
```suggestion
}
```

nahuhh:
squashed

eyedeekay:
This is sound from the I2P API perspective, handling for human-readable hostnames and hash based hostnames is identical and transparent for both SAM and SOCKS5.

src/net/i2p_address.cpp
```diff
@@ -68,6 +73,54 @@ namespace net
...
+        //! Validate a human-readable hostname (.i2p)
+        expect<void> host_check_human(boost::string_ref host) noexcept
+        {
+            if (!host.ends_with(tld_i2p))
+                return {net::error::expected_tld};
+
+            if (host.size() - (sizeof(tld_i2p) - 1) > i2p_max_length)
+                return {net::error::invalid_i2p_address};
+
+            //! Reject addresses starting with a dot or hyphen
```
selsta: This and later should be checked without `tld_i2p`, like `host_check_b32`.

src/net/i2p_address.cpp
```diff
@@ -68,6 +73,54 @@ namespace net
...
+            if (host.empty() || host.front() == '.' || host.front() == '-')
+                return {net::error::invalid_i2p_address};
+
+            //! Reject addresses containing invalid characters
+            for (char c : host)
+            {
+                if (!(std::islower(c) || std::isdigit(c) || c == '.' || c == '-'))
```
selsta: https://en.cppreference.com/cpp/string/byte/islower

unsafe without first converting to unsigned char, alternatively you could do

```
('a' <= c && c <= 'z') || ('0' <= c && c <= '9') || c == '.' || c == '-'
``

File: src/net/i2p_address.cpp [hunk 9]
```diff
@@ -107,7 +160,9 @@ namespace net
         boost::string_ref host = address.substr(0, address.rfind(':'));
         MONERO_CHECK(host_check(host));
 
-        static_assert(b32_length + sizeof(tld) == sizeof(i2p_address::host_), "bad internal host size");
+        static_assert(sizeof(i2p_address::host_) > b32_length &&
```
selsta: this appears to be missing `sizeof(tld_b32)` now

src/net/i2p_address.cpp
```diff
@@ -68,6 +73,54 @@ namespace net
...
+            }
+
+            //! Reject addresses with two dots in a row, a dot followed by a hyphen, or vice versa
+            if (host.find("..") != boost::string_ref::npos ||
+                host.find(".-") != boost::string_ref::npos ||
+                host.find("-.") != boost::string_ref::npos)
+                return {net::error::invalid_i2p_address};
+
+            //! Reject Base32 addresses 'disguised' as human-readable ones
+            if (host.ends_with(tld_b32))
```
selsta: This is currently unreachable because `host_check()` already handles `host.ends_with(tld_b32)`. Should it be removed or comment updated that it's a defensive check?

src/net/i2p_address.cpp
```diff
@@ -44,21 +42,28 @@ namespace net
...
+         * Naming rules defined here:
+         * https://i2p.net/en/docs/overview/naming/#naming-rules
+         */
+        constexpr const char tld_b32[] = u8".b32.i2p";
+        constexpr const char tld_i2p[] = u8".i2p";
         constexpr const char unknown_host[] = "<unknown i2p host>";
 
+        //! Lengths do not include TLD
         constexpr const unsigned b32_length = 52;
+        constexpr const unsigned i2p_max_length = 63;
```
selsta:
```suggestion
        constexpr const unsigned human_max_length = 63;
```

maybe better?

tests/unit_tests/net.cpp
```diff
@@ -470,12 +472,29 @@ TEST(i2p_address, invalid)
     EXPECT_TRUE(net::i2p_address::make(":").has_error());
     EXPECT_TRUE(net::i2p_address::make(".b32.i2p").has_error());
     EXPECT_TRUE(net::i2p_address::make(".b32.i2p:").has_error());
+    EXPECT_TRUE(net::i2p_address::make(".i2p").has_error());
+    EXPECT_TRUE(net::i2p_address::make(".i2p:").has_error());
+
+    EXPECT_TRUE(net::i2p_address::make("MONERO.i2p").has_error());
+    EXPECT_TRUE(net::i2p_address::make("m=nero.i2p").has_error());
+    EXPECT_TRUE(net::i2p_address::make("-monero.i2p").has_error());
+    EXPECT_TRUE(net::i2p_address::make("monero222222222222222222222222222222222222222222222222222222222222222.i2p").has_error());
```
selsta:
```suggestion
EXPECT_TRUE(net::i2p_address::make(std::string(human_max_length, 'a') + ".i2p").has_value());
EXPECT_TRUE(net::i2p_address::make(std::string(human_max_length + 1, 'a') + ".i2p").has_error());
```

tests/unit_tests/net.cpp
```diff
@@ -470,12 +472,29 @@ TEST(i2p_address, invalid)
...
+    EXPECT_TRUE(net::i2p_address::make("monero-.i2p").has_error());
+    EXPECT_TRUE(net::i2p_address::make("a.-monero.i2p").has_error());
+    EXPECT_TRUE(net::i2p_address::make("m..ero.i2p").has_error());
+    EXPECT_TRUE(net::i2p_address::make("m..ero.b32.i2p.i2p").has_error());
```
selsta, line 526:
```suggestion
    EXPECT_TRUE(net::i2p_address::make("m..ero.b32.i2p.i2p").has_error());
    EXPECT_TRUE(net::i2p_address::make("monero.i2p-").has_error());
    EXPECT_TRUE(net::i2p_address::make("monero..i2p").has_error());
    EXPECT_TRUE(net::i2p_address::make("a-.b.i2p").has_error());
    EXPECT_TRUE(net::i2p_address::make("a.-b.i2p").has_error());
    EXPECT_TRUE(net::i2p_address::make("a.b.i2p").has_value());
    EXPECT_TRUE(net::i2p_address::make("abc-123.i2p").has_value());
```

more edge cases

src/net/i2p_address.h
```diff
@@ -47,10 +47,18 @@ namespace serialization
...
+     */
     class i2p_address
     {
-        char host_[61]; // null-terminated
+        /**
+         * Host size (including TLD):
+         * - 60 for Base32 (52 chars of Base32 + ".b32.i2p")
+         * - Maximum of 67 for human-readable (63 or less chars + ".i2p")
+         */
+        char host_[67 + 1] = {}; // null-terminated
```
selsta: If you move the max size from .cpp to the .h (like with Tor) you can avoid hardcoding the size

src/net/i2p_address.cpp
selsta: still needs

```
#include <cctype>
```

iamamyth: `islower` depends on locale, and `isdigit` may introduce certain oddities as well (https://en.cppreference.com/cpp/string/byte/isdigit):
```
some implementations (e.g. Microsoft in [1252 codepage](https://en.wikipedia.org/wiki/Windows-1252)) may classify additional single-byte characters as digits.)
```

iamamyth: Also, there's wisdom in selsta's initial comment: Why introduce a function in place of a trivial conditional? Doing so only brings the possibility of mistakes, with no real benefit, as seen here.

src/net/i2p_address.cpp
```diff
@@ -68,6 +61,57 @@ namespace net
...
+            if (!host.ends_with(tld_i2p))
+                return {net::error::expected_tld};
+
+            host.remove_suffix(sizeof(tld_i2p) - 1);
+
+            if (host.empty() || host.size() > i2p_max_length)
+                return {net::error::invalid_i2p_address};
+
+            //! Reject addresses starting or ending with a dot or hyphen
+            if (host.empty() || host.front() == '.' || host.front() == '-' ||
```
iamamyth: The prior line just checked empty.

src/net/i2p_address.h
```diff
@@ -47,10 +47,27 @@ namespace serialization
 
 namespace net
 {
-    //! b32 i2p address; internal format not condensed/decoded.
+    /**
+     * Naming rules defined here:
+     * https://i2p.net/en/docs/overview/naming/#naming-rules
+     */
+    constexpr const char tld_b32[] = u8".b32.i2p";
```
iamamyth, line 54: The naming here seems likely to cause misapprehension: TLD has a precise meaning which makes each TLD its own namespace (`.i2p` is a TLD, `.b32.i2p` is a domain, not a TLD); if one were to just look at the variable names, it would be reasonable, but incorrect, to assume that members of `tld_i2p` are not members of `tld_b32`.

src/net/i2p_address.cpp
jpk68: Added a check for this.

src/net/i2p_address.h
jpk68: This is a good point, however the `tld` wording was already used in the original code, so it might be better to keep it as is.

src/net/i2p_address.cpp
jpk68: Fixed.

src/net/i2p_address.cpp
```diff
@@ -69,6 +64,57 @@ namespace net
...
+            if (host.ends_with(tld_b32))
+                return {net::error::invalid_i2p_address};
+
+            return success();
+        }
+
+        //! Use appropriate function for address validation
+        expect<void> host_check(boost::string_ref host) noexcept
+        {
+            if (host.ends_with(tld_b32))
```
selsta, line 104:
```
EXPECT_TRUE(net::i2p_address::make("example.b32.i2p").has_error());
```

I would add one explicit test for this to make it clear that we reject human readable addresses that end with .b32.i2p

src/net/i2p_address.cpp
```diff
@@ -110,7 +156,9 @@ namespace net
         net::canonicalize_host(normalized_host);
```
selsta, line 154:
```
const auto address = net::i2p_address::make("Example.I2P");
ASSERT_TRUE(address.has_value());
EXPECT_STREQ("example.i2p", address->host_str());
```

also I'd add one test for this

selsta (commented):
Two minor test suggestions, otherwise LGTM.

jpk68:
> Two minor test suggestions, otherwise LGTM.

Thanks, both done and pushed.

selsta (approved):

jpk68:
@iamamyth My apologies for asking, but another review would be extremely useful.

src/net/i2p_address.cpp
iamamyth: This check needs to happen prior to `remove_suffix` (near the start of the function), probably makes sense to just have it right below the other suffix check.

src/net/i2p_address.cpp
iamamyth: This comment wasn't ever addressed, and the suggested test would have caught the above bug I referenced.

src/net/i2p_address.cpp
```diff
@@ -69,6 +64,57 @@ namespace net
...
+                return {net::error::invalid_i2p_address};
+
+            //! Reject addresses containing invalid characters
+            for (char c : host)
+            {
+                if (!(('a' <= c && c <= 'z') || ('0' <= c && c <= '9') || c == '.' || c == '-'))
+                    return {net::error::invalid_i2p_address};
+            }
+
+            //! Reject addresses with two dots in a row, a dot followed by a hyphen, or vice versa
```
iamamyth: The multi-char pair testing could be part of the above loop, which would make it a lot more efficient, though I don't care all too much, given the relatively infrequent use of the function.

iamamyth: Minor addendum: If you don't feel like folding these `find`s into the loop, I'm happy to create a follow-on PR, as you already have fairly thorough tests for this scenario.

iamamyth:
One minor issue, but generally seems fine.

src/net/i2p_address.cpp
jpk68: Yeah, that would work. Thanks :)

selsta (approved):

jpk68:
@iamamyth I have applied fixes for all of your review comments after all; sorry for the confusion.

iamamyth (approved):

PR #10460, opened 2026-04-25
simplewallet: warn when mining thread count exceeds CPU cores

### LINKED ISSUE #10448: `start_mining` accepts thread count exceeding available CPUs without warning
When calling `start_mining` in `monero-wallet-cli` with a thread count greater than the number of CPU cores available on the system, mining starts silently with no warning.

If the specified thread count exceeds the number of available cores, the CLI should warn the user imo

rudolfschmidt:
closes #10448

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -5383,6 +5383,12 @@ bool simple_wallet::start_mining(const std::vector<std::string>& args)
     return true;
   }
 
+  const unsigned int hw_concurrency = boost::thread::hardware_concurrency();
+  if (hw_concurrency && req.threads_count > hw_concurrency)
+  {
+    message_writer(console_color_yellow, false) << boost::format(tr("Warning: %u mining threads requested exceeds the %u hardware threads available on this CPU. Consider using %u for best performance.")) % req.threads_count % hw_concurrency % hw_concurrency;
+  }
+
   COMMAND_RPC_START_MINING::response res;
   bool r = m_wallet->invoke_http_json("/start_mining", req, res);
   std::string err = interpret_rpc_response(r, res.status);


src/simplewallet/simplewallet.cpp
```diff
@@ -5383,6 +5383,12 @@ bool simple_wallet::start_mining(const std::vector<std::string>& args)
     return true;
   }
 
+  const unsigned int max_concurrency = tools::get_max_concurrency();
+  if (max_concurrency > 0 && req.threads_count > max_concurrency)
```
thomasbuilds: little detail but why not `max_concurrency` instead of `max_concurrency > 0`

src/simplewallet/simplewallet.cpp
```diff
@@ -5383,6 +5383,12 @@ bool simple_wallet::start_mining(const std::vector<std::string>& args)
...
+  {
+    message_writer(console_color_yellow, false) << boost::format(tr("warning: %u mining threads requested, but only %u CPU core(s) detected; this is unlikely to improve performance")) % (unsigned)req.threads_count % max_concurrency;
```
thomasbuilds: should drop the `(unsigned)` cast on `req.threads_count` imo, `threads_count` is `uint64_t`, and unlike `printf`, `boost::format` is type-safe

src/simplewallet/simplewallet.cpp
rudolfschmidt: Right, > 0 is redundant. Will drop it.

src/simplewallet/simplewallet.cpp
rudolfschmidt: True, was thinking printf-style format specifier where the cast matters. boost::format handles uint64_t fine. Will remove.

src/simplewallet/simplewallet.cpp
```diff
@@ -5383,6 +5383,12 @@ bool simple_wallet::start_mining(const std::vector<std::string>& args)
     return true;
   }
 
+  const unsigned int max_concurrency = tools::get_max_concurrency();
+  if (max_concurrency && req.threads_count > max_concurrency)
+  {
+    message_writer(console_color_yellow, false) << boost::format(tr("Warning: %u mining threads requested, but only %u CPU core(s) detected; this is unlikely to improve performance")) % req.threads_count % max_concurrency;
+  }
+
```
thomasbuilds: Just realized that `tools::get_max_concurrency()` does not return "CPU cores detected" it returns the value configurable via `--max-concurrency`, which only caps `tools::threadpool`, not the miner.

```suggestion
  const unsigned int hw_concurrency = boost::thread::hardware_concurrency();
  if (hw_concurrency && req.threads_count > hw_concurrency)
  {
    message_writer(console_color_yellow, false) << boost::format(tr("warning: %u mining threads requested, but only %u CPU core(s) detected; this is unlikely to improve performance")) % req.threads_count % hw_concurrency;
  }
```

thomasbuilds (changes requested):

src/simplewallet/simplewallet.cpp
thomasbuilds: Or we can keep max concurrency and change message

rudolfschmidt: Yeah you're right, I tested with default config so get_max_concurrency() returned hardware_concurrency() and the message looked correct. The edge case where someone runs with --max-concurrency N was missed. boost::thread::hardware_concurrency() is the right primitive here.

thomasbuilds:
Two things neither blocking imo:

- `boost::thread::hardware_concurrency()` returns logical cores. RandomX guidance is to size threads to physical cores, so on SMT systems this warning will fire later than ideal. Boost doesn't expose physical core count portably, so leaving the check as-is seems right though a short code comment calling out that this is a coarse upper bound would help.

- both `core_rpc_server::on_start_mining` and `DaemonHandler::handle(StartMining...)` already reject when `threads_count > hardware_concurrency() * 4`, so this warning effectively only fires in the `(hw_concurrency, 4 × hw_concurrency]` range

src/simplewallet/simplewallet.cpp
```diff
@@ -5383,6 +5383,12 @@ bool simple_wallet::start_mining(const std::vector<std::string>& args)
     return true;
   }
 
+  const unsigned int hw_concurrency = boost::thread::hardware_concurrency();
+  if (hw_concurrency && req.threads_count > hw_concurrency)
+  {
+    message_writer(console_color_yellow, false) << boost::format(tr("Warning: %u mining threads requested, but only %u CPU core(s) detected; this is unlikely to improve performance")) % req.threads_count % hw_concurrency;
```
thomasbuilds: I'd drop the (s) for translations may cause issues also it's basically impossible to have just 1 logic core. Also addressing previous point

```suggestion
    message_writer(console_color_yellow, false) << boost::format(tr("Warning: %u mining threads requested exceeds the %u hardware threads available on this CPU. Consider using %u or fewer for best performance.")) % req.threads_count % hw_concurrency % hw_concurrency;
```

thomasbuilds (changes requested):

src/simplewallet/simplewallet.cpp
rudolfschmidt: Adopted with one tweak: dropped "or fewer" from the action sentence. The warning already implies fewer is better, and a specific number reads stronger than a range.

thomasbuilds (approved):

selsta (approved):

Issue #10462, opened 2026-04-26
Dockerfile uses old base image


NuunMoon:
The `Dockerfile` currently uses a 6 year old Ubuntu base image, which has been deprecated since [May 2025](https://documentation.ubuntu.com/project/release-team/list-of-releases/). This should be updated to a newer version.

PR #10463, opened 2026-04-26
Dockerfile: Update base image

NuunMoon:
Update base image to latest Ubuntu LTS.
`adduser` has been removed from newer Ubuntu releases, so the lower level `useradd` has to be used. 

#10462

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Multistage docker build, requires docker 17.05
 
 # builder stage
-FROM ubuntu:20.04 AS builder
+FROM ubuntu:26.04 AS builder
 
 RUN set -ex && \
     apt-get update && \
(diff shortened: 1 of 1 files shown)


selsta:
Did you confirm if this successfully builds monerod?

NuunMoon:
Yeah it built for me locally and the blockchain started syncing. Don't know why that specific test fails though it may be flaky? My change shouldn't affect it.

NuunMoon:
Can we rerun that job? It tests a race condition and it seems like it's flaky, but the failure is unrelated to my changes.

selsta:
Tests are irrelevant here since the CI suite does not run the Dockerfile.

selsta (approved):

tobtoht (approved):

PR #10467, opened 2026-04-26
zmq: add missing restricted rpc check for get_block_headers_by_height

jeetrex17:
`DaemonHandler::handle(GetBlockHeadersByHeight)` in `src/rpc/daemon_handler.cpp` did not cap `req.heights` under `--restricted-rpc`, so a restricted caller could request an unbounded number of block headers in a single ZMQ call.

Add an early return when `req.heights.size() > restricted_max_block_headers`, using the same 1000-request limit as HTTP `on_get_blocks_by_height` in `core_rpc_server.cpp`.

Follow-up to #10424.


selsta (approved):

selsta:
Please also open against release branch.

jeetrex17:
Sure

tobtoht (approved):

PR #10468, opened 2026-04-27
cmake: link libatomic for Android ARMv7 Trezor builds

duyua9:
## Summary
- add `atomic` to the Android ARMv7 protobuf try-compile link libraries
- add the same `atomic` dependency to the final Trezor target for Android ARMv7

## Rationale
Recent protobuf/Abseil builds can require compiler-rt atomic builtins on 32-bit Android ARM. The existing Android Trezor path already links `log` in both the protobuf try-compile and the final target, but it does not carry `atomic`, so the configure-time probe and/or final target can fail on ARMv7 while other Android ABIs keep working.

This keeps the extra dependency scoped to Android ARM/ARMv7 and mirrors it in both places where the protobuf/Trezor link happens.

## Testing
- `git diff --check`
- `python3 /Users/ming/.openclaw/skills/github-multi-account-routing/scripts/verify_pr_submission_identity.py --repo monero-project/monero --worktree /Users/ming/code/monero-project/monero --account duyua9`


duyua9:
Closing this after re-review. The patch is too speculative without an Android ARMv7 reproduction or CI proof that the extra libatomic link fixes a current build failure. I don't want to leave an unverified build-system change for maintainers to triage.

PR #10472, opened 2026-04-27
wallet2: reject duplicate outputs in reserve proofs

selsta:


j-berman (approved):
I think the O(n*log(n)) check on duplicate `seen_outputs` is ok because a malicious user could craft worst-case O(n^2) lookups here when using an unordered set.

tobtoht (approved):

PR #10475, opened 2026-04-28
fix: warning: variable set but not used

tobtoht:


tests/unit_tests/epee_levin_protocol_handler_async.cpp
```diff
@@ -658,7 +658,7 @@ TEST_F(test_levin_protocol_handler__hanle_recv_with_invalid_data, handles_bad_cb
   m_req_head.m_flags = LEVIN_PACKET_BEGIN;
   m_req_head.m_command = 0;
 
-  epee::levin::bucket_head2 inner{};
```
vtnerd, line 661: Was this variable supposed to be removed? Removing just the brace `{}` initialization doesn't seem correct.

vtnerd (changes requested):

vtnerd (approved):

selsta (approved):

PR #10481, opened 2026-04-28
wallet2: throw if reserve proof can't be parsed [0.18]

tobtoht:
Extracted and backported from #10475


selsta (approved):

SChernykh (approved):

vtnerd (changes requested):
This just changes the throw type?

tobtoht:
> This just changes the throw type?

Only if `m_load_deprecated_formats` is true.

PR #10486, opened 2026-04-28
wallet-rpc: restrict tx key / tx proof / spend proof / reserve proof

jeetrex17:
`get_tx_key`, `get_tx_proof`, `get_spend_proof`, and `get_reserve_proof` were available in `--restricted-rpc` mode.

These methods expose transaction secret keys or generate wallet ownership/spend proofs, so they should be restricted like other non-view-only wallet RPC commands.


selsta:
We decided to intentionally leave these unrestricted, see the discussion here: https://github.com/monero-project/monero/pull/10378

jeetrex17:
Thanks for the pointer , I completely missed #10378. Closing this since the decision was already made. Apologies for the noise.

PR #10490, opened 2026-04-28
cryptonote_core: restore separate fakechain data dir for core_tests

### LINKED ISSUE #10487: `core_tests` now touches mainnet data directory
Another regression was added in #10396. The following lines were deleted:

https://github.com/monero-project/monero/blob/cb0726eb24df6fbff59d79506659f41967829c6a/src/cryptonote_core/cryptonote_core.cpp#L482-L483

`core_tests` sets the network type to `FAKECHAIN`, but does so through the `test_options` parameter, not command line arguments. Because of this, the core in `core_tests` uses the mainnet data directory, instead of the `/fake` subdirectory. It just deleted my node's database...

Image

selsta:
Resolves #10487

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -489,6 +489,11 @@ namespace cryptonote
     bool keep_fakechain = command_line::get_arg(vm, arg_keep_fakechain);
 
     boost::filesystem::path folder(m_config_folder);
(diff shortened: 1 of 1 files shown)


selsta:
~~This isn't the correct fix yet, will force push.~~ Done

j-berman (approved):
This should work

tobtoht (approved):

PR #10492, opened 2026-04-29
daemon: on exit, stop the daemon's blockchain processing protocol asap

j-berman:
#10382 introduced the following flow: wait until all connections complete their termination sequence, and *then* stop the blockchain protocol from continuing to process blocks.

If the server is in the middle of a heavy task like `try_add_next_blocks` that actually may never end until synced, then the user has to sit and wait for that heavy task to finish before all connections can complete the terminate sequence.

This PR ensures the blockchain protocol is canceled as soon as the user stops the server, before waiting for the termination sequence to complete.

____

Note: I think #10488 is a higher priority PR than this one, but I think this is also a necessary fix so connections terminate immediately and avoid that PR's `wait_for` actually waiting until timeout and returning false.

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -1107,6 +1107,8 @@ namespace nodetool
   template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::send_stop_signal()
   {
+    MDEBUG("[node] stopping server payload handler");
+    m_payload_handler.stop();
     MDEBUG("[node] sending stop signal");
     for (auto& zone : m_network_zones)
     {
@@ -1129,7 +1131,6 @@ namespace nodetool
       zone.second.m_net_server.send_stop_signal(close_all_connections);
     }
     MDEBUG("[node] Stop signal sent");
-    m_payload_handler.stop();
     return true;
   }
   //-----------------------------------------------------------------------------------


tobtoht (approved):

src/daemon/daemon.cpp
```diff
@@ -319,6 +320,7 @@ void t_daemon::stop_p2p()
   {
     throw std::runtime_error{"Can't send stop signal to a stopped daemon"};
   }
+  mp_internals->core.get().get_blockchain_storage().cancel();
   mp_internals->p2p.get().send_stop_signal();
```
vtnerd, line 322: Same thing here, this should indirectly call `blockchain::cancel()`.

src/daemon/daemon.cpp
```diff
@@ -306,6 +306,7 @@ void t_daemon::stop()
   {
     throw std::runtime_error{"Can't stop stopped daemon"};
   }
+  mp_internals->core.get().get_blockchain_storage().cancel();
   mp_internals->p2p.stop();
```
vtnerd, line 309: This function should (indirectly) already invoke `blockchain::stop()`. Is the problem the order of calls?

vtnerd (changes requested):

src/daemon/daemon.cpp
j-berman: Yes, the problem is that it's calling it after the synchronous termination sequence introduced in #10382

PR description explains it

j-berman: As per this comment (and DM's), I updated the latest to stop the blockchain protocol handler *before* closing connections *inside* the server's `send_stop_signal`, rather than introduce an additional cancel call outside before calling `send_stop_signal`

vtnerd (approved):

PR #10505, opened 2026-04-30
wallet2: refactor multisig ignore sets into separate member functions

UkoeHB:
Moving logic into functions for eventual reuse with carrot.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -9879,65 +9879,9 @@ void wallet2::transfer_selected_rct(std::vector<cryptonote::tx_destination_entry
     THROW_WALLET_EXCEPTION_IF(needed_money < dt.amount, error::tx_sum_overflow, dsts, fee, m_nettype);
   }
 
-  // if this is a multisig wallet, create a list of multisig signers we can use
-  std::deque<crypto::public_key> multisig_signers;
-  size_t n_multisig_txes = 0;
-  std::vector<std::unordered_set<crypto::public_key>> ignore_sets;
-  if (m_multisig && !m_transfers.empty())
-  {
-    const crypto::public_key local_signer = get_multisig_signer_public_key();
-    size_t n_available_signers = 1;
-
-    // At this step we need to define set of participants available for signature,
-    // i.e. those of them who exchanged with multisig info's
-    // note: The oldest unspent owned output's multisig info (in m_transfers) will contain the most recent result of
-    //       'import_multisig()', which means only 'fresh' multisig infos (public nonces) will be used to make tx attempts.
-    //       - If a signer's info was missing from the latest call to 'import_multisig()', then they won't be able to participate!
-    //       - If a newly-acquired output doesn't have enough nonces from multisig infos, then it can't be spent!
-    for (const crypto::public_key &signer: m_multisig_signers)
-    {
-      if (signer == local_signer)
-        continue;
-      for (const auto &i: m_transfers[0].m_multisig_info)
-      {
-        if (i.m_signer == signer)
-        {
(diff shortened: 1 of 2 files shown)


src/wallet/wallet2.cpp
```diff
@@ -10084,8 +10031,6 @@ void wallet2::transfer_selected_rct(std::vector<cryptonote::tx_destination_entry
 
   std::vector<tools::wallet2::multisig_sig> multisig_sigs;
   if (m_multisig) {
-    if (ignore_sets.empty())
-      ignore_sets.emplace_back();
     const std::size_t num_multisig_attempts = ignore_sets.size();
     multisig_sigs.resize(num_multisig_attempts);
```
UkoeHB, line 10100: `multisig_attempt_ignore_sets` will return one empty set when a single attempt is needed.

UkoeHB: This is pointing at the wrong lines for some reason. Check the file overview.

selsta (approved):

src/wallet/wallet2.cpp
```diff
@@ -14497,6 +14444,82 @@ crypto::key_image wallet2::get_multisig_composite_key_image(size_t n) const
...
+      {
+        available_signers_out.push_back(signer);
+        break;
+      }
+    }
+  }
+}
+//----------------------------------------------------------------------------------------------------
+// Sets of multisig signers who should be ignored from a multisig tx signing attempt.
+// One set exists for each permutation of available signers equal to the multisig threshold.
```
iamamyth: Per the implementation code, it seems one set exists per combination (rather than per permutation), as the signer sets are unordered.

src/wallet/wallet2.cpp
```diff
@@ -14497,6 +14444,82 @@ crypto::key_image wallet2::get_multisig_composite_key_image(size_t n) const
   return ki;
 }
 //----------------------------------------------------------------------------------------------------
+// Gets the set of participants available for a signature, i.e. those who exchanged multisig infos with us.
+//
+// Assumes the oldest unspent owned output's multisig info (in m_transfers) will contain the most recent result of
+// 'import_multisig()', which means only 'fresh' multisig infos (public nonces) will be used to make tx attempts.
+// - If a signer's info was missing from the latest call to 'import_multisig()', then they won't be able to participate!
+// - If a newly-acquired output doesn't have enough nonces from multisig infos, then it can't be spent!
+void wallet2::multisig_available_signers(std::deque<crypto::public_key> &available_signers_out) const
```
iamamyth: Why not have this return the deque rather than accepting an out param? Relying on named return value optimization seems to be more in keeping with the C++ style used elsewhere in this project.

iamamyth: (similar comment also applies to multisig_attempt_ignore_sets)

UkoeHB: Out params are the pattern used in all the carrot code. So in that sense it is 'the pattern' for new code (mostly, it's not a hard rule).

iamamyth: But this is in wallet2, which definitely uses the pattern I mentioned (see, e.g., the various listeners, or method such as export_multisig, get_keys_file_data, and get_cache_file_data). I don't think the originating project codename matters, only the location in the post-merge source tree.

UkoeHB: Changed it

selsta:
please squash

tobtoht:
Needs rebase

UkoeHB:
Git submodules are such a pita, why do we use them?

tobtoht (approved):

PR #10506, opened 2026-04-30
cmake: remove static_assert check

tobtoht:
Introduced in the initial commit. This is redundant since moving to C11, because:

`<assert.h>` includes: `#define static_assert _Static_assert`

https://cppreference.net/c/error/static_assert.html


selsta (approved):

vtnerd (approved):

jeffro256 (approved):

jeffro256:
This a good change. Hopefully we can find other ways to reduce the time to configure for ease of use for IDE development.

PR #10517, opened 2026-05-03
epee: fix memory leak with readline

ComputeryPony:
`history_get_history_state` returns a `HISTORY_STATE*` that it allocates and expects the user to free.
Free it before leaving the function to avoid leaking it.
Caught with valgrind.

### DIFF
--- a/contrib/epee/src/readline_buffer.cpp
+++ b/contrib/epee/src/readline_buffer.cpp
@@ -2,6 +2,7 @@
 #include <readline/readline.h>
 #include <readline/history.h>
 #include <iostream>
+#include <memory>
 #include <boost/thread/mutex.hpp>
 #include <boost/thread/lock_guard.hpp>
 #include <boost/algorithm/string.hpp>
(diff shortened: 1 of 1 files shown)


contrib/epee/src/readline_buffer.cpp
```diff
@@ -202,8 +202,10 @@ static bool same_as_last_line(const std::string& test_line)
 {
   // Note that state->offset == state->length, when a new line was entered.
   HISTORY_STATE* state = history_get_history_state();
-  return state->length > 0
+  bool same = state->length > 0
     && test_line.compare(state->entries[state->length-1]->line) == 0;
+  free(state);
```
vtnerd: This _should be okay_, but `std::string::compare` can throw technically. Should this be wrapped in a `std::unique_ptr` instead?

vtnerd (changes requested):

contrib/epee/src/readline_buffer.cpp
ComputeryPony: Agreed, should be fixed in all return paths, however unlikely.

ComputeryPony:
I *think* I got the C++isms of this correct, someone should double check me.

vtnerd (approved):

iamamyth (approved):

tobtoht:
Please squash.

ComputeryPony:
> Please squash.

Done

selsta (approved):

selsta:
Please also open this one against release branch.

PR #10518, opened 2026-05-03
p2p: initialize m_fallback_seed_nodes_added before first use

ComputeryPony:
`m_fallback_seed_nodes_added` wasn't being initialized before it was used in `net_node.inl`, properly initialize it.
Caught with valgrind.

### DIFF
--- a/src/p2p/net_node.h
+++ b/src/p2p/net_node.h
@@ -465,7 +465,7 @@ namespace nodetool
 
     std::list<epee::net_utils::network_address>   m_priority_peers;
     std::vector<epee::net_utils::network_address> m_exclusive_peers;
-    std::atomic_flag m_fallback_seed_nodes_added;
+    std::atomic_flag m_fallback_seed_nodes_added = ATOMIC_FLAG_INIT;
     std::vector<nodetool::peerlist_entry> m_command_line_peers;
     uint64_t m_peer_livetime;
     //keep connections to initiate some interactions


src/p2p/net_node.h
```diff
@@ -465,7 +465,7 @@ namespace nodetool
 
     std::list<epee::net_utils::network_address>   m_priority_peers;
     std::vector<epee::net_utils::network_address> m_exclusive_peers;
-    std::atomic_flag m_fallback_seed_nodes_added;
+    std::atomic_flag m_fallback_seed_nodes_added = ATOMIC_FLAG_INIT;
```
vtnerd, line 468: Should this be moved to the initializer list in the constructor? I can see argument both ways.

There appears to be a few more variables that aren't explicitly initialized?

vtnerd (changes requested):

src/p2p/net_node.h
ComputeryPony: >Should this be moved to the initializer list in the constructor? I can see argument both ways.

I'm not very C++ minded so I have no opinion on if one is cleaner than the other.
If you feel it should be moved to the initializer list in the constructor I'm fine with moving it.

>There appears to be a few more variables that aren't explicitly initialized?

By my count there are 43 variables in this part of the class, the constructor initializes only 11. Unfortunately I lack any in-depth understanding of this class and what all is currently missing from being initialized.

I do however notice that no other variable seems to be initialized like this so I guess it would be more in-line with other code to move it to the initializer list.

iamamyth:
Per [C++ docs](https://en.cppreference.com/cpp/atomic/ATOMIC_FLAG_INIT), I think the prior approach (assignment) was right, and the updated one wrong:
```
Defines the initializer which can be used to initialize std::atomic_flag to clear (false) state in the form std::atomic_flag v = ATOMIC_FLAG_INIT;. It is unspecified if it can be used with other initialization contexts. 
```

The whole thing becomes irrelevant in C++20, but I think we're still on C++17.

ComputeryPony:
Reverted as per previous comment.

iamamyth (approved):

tobtoht:
Please squash.

ComputeryPony:
> Please squash.

Done

selsta (approved):

PR #10519, opened 2026-05-03
crypto: implement CTHR_THREAD_CLOSE to avoid leaking memory

ComputeryPony:
`CTHR_THREAD_CLOSE` had no implementation causing it to never free resources associated with the handle that was opened for the thread.
Free it to avoid leaking memory.
Caught with valgrind.


iamamyth (approved):
Agreed that `pthread_detach` would be the pthreads analog to `CloseHandle`.

SChernykh (approved):

SChernykh:
Good catch. If the thread isn't joined, it must be detached to not leak.

selsta:
@ComputeryPony please also open a PR of this patch against release branch

PR #10521, opened 2026-05-04
epee: fail digest auth cleanly when MD5 is unavailable

selsta:


vtnerd (approved):
This could use one other look-over, but it seems fine.

jpk68 (approved):

PR #10523, opened 2026-05-05
net/i2p: support sam protocol

nahuhh:
mirror of jpk68's codeberg branch

https://codeberg.org/jpk68/monero/pulls/33


src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+            MINFO("Read I2P destination key from file (" << key_path << "): " << private_key);
+
+            if (private_key.empty())
+                throw std::runtime_error("I2P destination key is empty: " + key_path);
+        }
+        else
+        {
+            send_hello_version();
+            send_dest_generate();
```
SChernykh: Correct me if I'm wrong, but the two above are structs, and you're just creating unnamed struct objects and then immediately destroying them, without calling operator() here.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+            if (private_key_.empty())
+                throw std::runtime_error("Failed to generate I2P destination");
+
+            if (!epee::file_io_utils::save_string_to_file(key_path, private_key))
```
SChernykh: Here you're saving the empty `private_key` (local variable which wasn't touched on this code path). It should be `private_key_` probably?

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,532 @@
...
+                throw std::runtime_error("Failed to save I2P destination key to " + key_path);
+        }
+
+        private_key_ = private_key;
```
SChernykh: If the file didn't exist, this will overwrite `private_key_` with an empty string, even if `private_key_` wasn't empty before.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,532 @@
...
+                [self](){ send_stream_connect{self}({}); });
+            return;
+
+        default:
+            MINFO("Done with SAM loop");
+            done({}, self);
+            return;
+        }
+
+        MDEBUG("SAM state changed to: " << static_cast<int>(state_));
```
SChernykh: This code will never be reached: the `switch` above always returns.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,532 @@
...
+        std::shared_ptr<client> self;
+
+        void operator()(boost::system::error_code ec, std::size_t bytes)
+        {
+            if (ec) return self->done(ec, self);
+
+            if (bytes == 0)
+            {
+                boost::asio::async_read_until(self->socket_, boost::asio::dynamic_buffer(self->line_), '\n',
+                    boost::asio::bind_executor(self->strand_, *this));
```
SChernykh: This makes `operator()` to be called both as a write callback and as a read callback. It's confusing and can create some subtle bug. It's better to have separate callbacks for reads and writes.

eyedeekay:
The main problems here from the I2P point of view are related to the lifecycle of the control socket. Right now the issue with this code is that the control socket is being managed by `Handler`, which has a `done` which causes the control socket to be destroyed. That causes the whole SAM session to drop and you end up with a reference to a session that doesn't exist. This is probably the most common pitfall in SAMv3 protocol implementation, I've done it, everybody's done it, but, in it's current form, this implementation is broken. I haven't been a C++ programmer since my mid 20's so I'm struggling to fix it myself but knowing what you need to do:

 - Split the client handler into 2 classes/structures, one for the control socket, one or more for the actual sessions on that socket.
 - Keep the control socket open even when a session closes, `done` for sessions should not tear down the control socket until the last session is being torn down **And/Or**
 - You may need to add an explicit shutdown function to the control socket which drains the sessions and shuts down the socket properly
 - I can't quite follow the `private_key_from_file` and `next_state` relationship, but I think you might be throwing away the keys. I'm not enough of a C++ guy to fix this right but I think it just needs a little rethinking, simplify it and it will make your life easier.

Those will fix the current problems with the SAMv3 client and get this PR back on track.

zzzi2p:
jpk68: You shared some (failing) results and debug logs in I2P IRC the other day and then vanished. I did some analysis and testing and have some insights for you, come on back over there.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,554 @@
...
+        {
+            line.pop_back();
+        }
+    }
+
+    void client::async_write_command()
+    {
+        auto self = shared_from_this();
+
+        boost::asio::async_write(socket_, boost::asio::buffer(self->get_write_buffer()),
```
selsta:
```suggestion
        boost::asio::async_write(socket_, boost::asio::buffer(self->write_buffer_),
```

otherwise `get_write_buffer()` returns a `std::string` by value, which gets destroyed before the async write completes.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,554 @@
...
+        }
+
+        return private_key;
+    }
+
+    std::string random_session_id()
+    {
+        std::random_device rng;
+        std::mt19937 gen(rng());
+        std::uniform_int_distribution<> dist('a', 'z');
```
selsta:
```
#include <random>
```

tests/unit_tests/i2p_sam.cpp
```diff
@@ -25,17 +25,14 @@
...
+TEST(i2p_sam, get_value_ok)
 {
-  WIRE_DECLARE_BLOB_NS(crypto::hash);
-  WIRE_DECLARE_BLOB_NS(crypto::hash8);
-  WIRE_DECLARE_BLOB_NS(crypto::ec_point);
+    std::string out;
+    auto ec = net::sam::get_value("HELLO REPLY RESULT=OK VERSION=3.1", "RESULT", out);
+    EXPECT_FALSE(ec);
+    EXPECT_EQ(out, "OK");
 }
```
selsta, line 656:
```
TEST(i2p_sam, get_value_does_not_match_key_inside_value)
{
    std::string out;
    auto ec = net::sam::get_value("MESSAGE=BAD_RESULT=OK", "RESULT", out);

    EXPECT_TRUE(ec);
    EXPECT_EQ(ec, net::sam::make_error_code(net::sam::error::parse_failed));
}
```

test case to add that current code does not handle correctly

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,554 @@
...
+                return make_error_code(error::parse_failed);
+
+            return parse_result_code(line);
+        }
+
+        case state::naming_lookup:
+        {
+            boost::system::error_code ec;
+
+            ec = get_value(line, "VALUE", destination_);
```
selsta: Should you do `parse_result_code` first?

selsta (commented):
It would be good if more test cases get added.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,535 @@
...
+        strip_newlines(line);
+
+        auto result = parse_result(line);
+        if (result) return done(result);
+
+        if (state_ == state::stream_accept && stream_accept_listening_)
+        {
+            async_read_line();
+            return;
+        }
```
selsta: The whole code involving `stream_accept_listening_` appears incorrect as the accept flow seems to expect two newline-delimited reads.

tests/unit_tests/i2p_sam.cpp
```diff
@@ -0,0 +1,82 @@
...
+    const std::string message2 = "HELLO REPLY RESULT=OK VERSION=3.1";
+    std::string out2{};
+    auto ec2 = net::sam::get_value(message2, "HELLO", out2);
+    EXPECT_TRUE(ec2);
+}
+
+TEST(i2p_sam, get_value_does_not_match_key_inside_value)
+{
+    std::string out;
+    auto ec = net::sam::get_value("MESSAGE=BAD_RESULT=OK", "RESULT", out);
```
selsta, line 688: This test case still appears to be failing because the parser does not require space-delimited `KEY=VALUE` fields.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,535 @@
...
+                {
+                    boost::system::error_code ec_;
+                    self->socket_.shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec_);
+                    self->socket_.close(ec_);
+                }
+            });
+        }
+    }
+
+    std::string private_key_from_file(const std::string& data_dir)
```
selsta: This does not sanitize file content like stripping `\n`, `\r`, space, tab, etc which can break SAM commands.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,535 @@
...
+            boost::system::error_code ec;
+
+            ec = get_value(line, "DESTINATION", destination_);
+            if (!ec)
+            {
+                MDEBUG("Created session with DESTINATION=" << destination_.substr(0, 25) << "...");
+                return {};
+            }
+
+            return parse_result_code(line);
```
selsta: This lacks prefix validiation, same applies to `stream_connect` and `stream_accept`. `hello_version` does some validation but not strict enough.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,535 @@
...
+        }
+
+        case state::session_create:
+        {
+            boost::system::error_code ec;
+
+            ec = get_value(line, "DESTINATION", destination_);
+            if (!ec)
+            {
+                MDEBUG("Created session with DESTINATION=" << destination_.substr(0, 25) << "...");
```
selsta: This path lacks `RESULT=OK` verification.

src/p2p/net_node.inl
```diff
@@ -564,8 +576,70 @@ namespace nodetool
     if ( !set_rate_limit(vm, command_line::get_arg(vm, arg_limit_rate) ) )
       return false;
 
-
     epee::byte_slice noise = nullptr;
+
+    if (command_line::has_arg(vm, arg_i2p_sam) && !command_line::is_arg_defaulted(vm, arg_i2p_sam))
+    {
+      if (!command_line::is_arg_defaulted(vm, arg_tx_proxy) || !command_line::is_arg_defaulted(vm, arg_anonymous_inbound))
+        MWARNING("--i2p-sam will override --tx-proxy and --anonymous-inbound");
```
selsta: It also currently overrides `arg_max_connections_per_ip`.

Also in my opinion it should only override --tx-proxy i2p and i2p inbound.

src/p2p/net_node.inl
```diff
@@ -3046,6 +3108,25 @@ namespace nodetool
...
+    if (remote.get_type_id() != net::i2p_address::get_type_id())
+      return boost::none;
+
+    const net::i2p_address& i2p_addr = remote.as<net::i2p_address>();
+    auto result = sam_connect_internal(zone.m_net_server.get_stop_signal(), zone.m_net_server.get_io_context(), zone.m_sam_router_endpoint, i2p_addr, zone.m_sam_session_id);
+
+    if (result) // if no error
+    {
+      p2p_connection_context context{};
+      if (zone.m_net_server.add_connection(context, std::move(*result), remote, ssl_support))
```
selsta: log on failure to connect here

src/p2p/net_node.inl
```diff
@@ -564,8 +576,70 @@ namespace nodetool
...
+      auto handler = [&i2p_sam_zone](boost::system::error_code ec, boost::asio::ip::tcp::socket socket)
+      {
+        if (ec)
+        {
+          MERROR("Failed to create I2P SAM control socket: " << ec.message());
+          return;
+        }
+
+        p2p_connection_context context{};
+        i2p_sam_zone.m_net_server.add_connection(context, std::move(socket), net::i2p_address::unknown(),
```
selsta: also here logging

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,535 @@
...
+    {
+        std::shared_ptr<client> self;
+
+        void operator()(boost::system::error_code ec, const std::string& id, const std::string& private_key)
+        {
+            if (ec) return self->done(ec);
+
+            MDEBUG("Sending SESSION CREATE command with ID=" << id);
+
+            self->set_write_buffer("SESSION CREATE STYLE=STREAM ID=" + id + " DESTINATION=" + private_key + " i2cp.leaseSetEncType=6,4\n");
```
selsta: add a comment explaining what the numbers mean

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,535 @@
...
+                    self->socket_.shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec_);
+                    self->socket_.close(ec_);
+                }
+            });
+        }
+    }
+
+    std::string private_key_from_file(const std::string& data_dir)
+    {
+        const std::string key_path = data_dir + "/i2p_private_key";
```
selsta: better to use `boost::filesystem::path`

selsta (commented):
Minumum of test coverage I would like to see,

for `get_value()`

```
RESULT=OK
HELLO REPLY RESULT=OK VERSION=3.1
MESSAGE=BAD_RESULT=OK
FOO RESULT=OKAY
RESULT=
```

----

for `parse_result_code()`

```
RESULT=OK
RESULT=I2P_ERROR
RESULT=TIMEOUT
RESULT=NOVERSION
RESULT=SOMETHING_UNKNOWN
no RESULT field
```

expected

```
OK -> no error
known SAM errors -> correct net::sam::error
unknown/missing -> parse_failed
```

___

for Prefix validation

```
hello_version accepts HELLO REPLY ...
session_create accepts SESSION STATUS ...
stream_connect accepts STREAM STATUS ...
stream_accept accepts STREAM STATUS ...
```

with these helpers

```
boost::system::error_code parse_hello_reply(const std::string& line);
boost::system::error_code parse_session_status(const std::string& line, std::string* destination);
boost::system::error_code parse_stream_status(const std::string& line);
```
____

```
SESSION STATUS RESULT=I2P_ERROR DESTINATION=abc
```

this must fail correctly, currently it does not

____

Private key file trimming tests

contrib/epee/include/net/abstract_tcp_server2.h
```diff
@@ -413,7 +413,7 @@ namespace net_utils
       default_remote = std::move(remote);
     }
 
-    bool add_connection(t_connection_context& out, boost::asio::ip::tcp::socket&& sock, network_address real_remote, epee::net_utils::ssl_support_t ssl_support = epee::net_utils::ssl_support_t::e_ssl_support_autodetect);
+    bool add_connection(t_connection_context& out, boost::asio::ip::tcp::socket&& sock, network_address real_remote, epee::net_utils::ssl_support_t ssl_support = epee::net_utils::ssl_support_t::e_ssl_support_autodetect, bool is_income = false);
```
vtnerd: My initial thought, without seeing rest of code, is that this is unnecessary.

src/net/i2p_sam.h
```diff
@@ -0,0 +1,296 @@
...
+    {
+    protected:
+        boost::asio::ip::tcp::socket socket_;
+        boost::asio::strand<boost::asio::ip::tcp::socket::executor_type> strand_;
+
+        std::string line_;
+        std::string write_buffer_;
+
+        //! The public address or Base64-encoded destination of an I2P router.
+        std::string destination_;
```
vtnerd: Can this be of type `i2p_address` too?

src/p2p/net_node.cpp
```diff
@@ -392,4 +401,65 @@ namespace nodetool
...
+        {
+            boost::promise<client_result> sam_promise{};
+            sam_result = sam_promise.get_future();
+
+            auto client = net::sam::make_connect_client(
+                boost::asio::ip::tcp::socket{service}, notify{std::move(sam_promise)}
+            );
+            client->set_session_id(session_id);
+
+            close_client.self = client;
```
vtnerd, line 416: This should be doe before `set_session_id` in case that function throws.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,649 @@
...
+
+            MINFO("Read I2P destination key from file (" << key_path << "), length=" << private_key.size());
+        }
+
+        return private_key;
+    }
+
+    std::string random_session_id()
+    {
+        std::random_device rng;
```
vtnerd: Use `crypto::random_device` in `src/crypto/crypto.h`. Just pass it directly to `dist`.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,649 @@
...
+        {
+            async_read_stream_accept();
+            return;
+        }
+
+        if (state_ == state::session_create)
+        {
+            auto self = shared_from_this();
+
+            auto timer = std::make_shared<boost::asio::steady_timer>(socket_.get_executor());
```
vtnerd: Is this masking some serious bug? Why would need you to wait 500 milliseconds before continuing the session that has just been created?

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,649 @@
...
+        if (result == "ALREADY_ACCEPTING") return make_error_code(error::already_accepting);
+        if (result == "NOVERSION") return make_error_code(error::noversion);
+
+        return make_error_code(error::parse_failed);
+    }
+
+    bool has_prefix(const std::string& line, const std::string& prefix)
+    {
+        if (line.compare(0, prefix.size(), prefix) != 0) return false;
+        return line.size() == prefix.size() || line[prefix.size()] == ' ';
```
vtnerd, line 562: `line[prefix.size()]` looks to be an invalid access if `line` is shorter than `prefix.size()` (i.e. I don't see where this constraint is verified/checked).

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,649 @@
...
+        default:
+            MINFO("Done with SAM loop");
+            done({});
+            return;
+        }
+    }
+
+    boost::system::error_code get_value(const std::string& line, const std::string& key, std::string& out)
+    {
+        const std::string pattern = key + "=";
```
vtnerd: Does the spec allow spaces between the key and `=`?

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,649 @@
...
+        {
+            line.pop_back();
+        }
+    }
+
+    bool has_prefix(const std::string& line, const std::string& prefix);
+
+    void client::async_write_command()
+    {
+        auto self = shared_from_this();
```
vtnerd: This should be a static method of `client` with first argument as `std::shared_ptr`, as all cases already have a `shared_ptr`.

`shared_from_this` isn't terribly expensive, but its also not cheap, as atomic stuff has to be mucked with.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,649 @@
...
+            self->async_write_command();
+        }
+    };
+
+    bool client::connect_and_send(std::shared_ptr<client> self_, const stream_type::endpoint& router)
+    {
+        if (!self_) return false;
+
+        client& self = *self_;
+
```
vtnerd, line 306: Might make sense to set state to `hello_version` here.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,649 @@
...
+        if (ec) return done(ec);
+
+        std::string line = line_.substr(0, bytes);
+        line_.erase(0, bytes);
+        strip_newlines(line);
+
+        auto result = parse_result(line);
+        if (result) return done(result);
+
+        if (state_ == state::stream_accept && stream_accept_listening_)
```
vtnerd: What's the point of the `next_state` function if you're doing special logic here? There's already specialized logic for `client` vs `control_socket`

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,649 @@
...
+            line_.clear();
+            stream_accept_listening_ = false;
+            state_ = state::hello_version;
+            socket_.async_connect(router_endpoint_,
+                boost::asio::bind_executor(strand_, send_hello_version{self}));
+            return;
+
+        default:
+            MINFO("Done with SAM loop");
+            done({});
```
vtnerd: This seems like an odd default case.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,649 @@
...
+
+            private_key = sanitize_private_key(std::move(private_key));
+
+            MINFO("Read I2P destination key from file (" << key_path << "), length=" << private_key.size());
+        }
+
+        return private_key;
+    }
+
+    std::string random_session_id()
```
vtnerd, line 737: How does this work without the prefix for class? Is the test not being compiled in?

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,649 @@
...
+            else
+            {
+                state_ = state::session_create;
+                boost::asio::dispatch(strand_,
+                    [self, this](){ send_session_create{self}({}, session_id_, private_key_); });
+            }
+            return;
+
+        case state::session_create:
+            session_socket_ = std::move(socket_);
```
vtnerd, line 487: What happens if the remote side closes `sesion_socket_`? Doesn't this require everything to be closed or something? Should there be an outstanding read on `session_socket_` to detect when it closes?

vtnerd (changes requested):

vtnerd:
As another follow up Q:

 > Sane default tunnel configuration per I2P expert guidance

I don't recall seeing anything involving this ?

vtnerd:
> Handle external router start/stop/restart gracefully

This is Part 1 of CCS. I don't see how an external router stop or restart is handled in this post, as the session socket is never touched after initial connection. I would need to be directed where this in particular is completed.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,686 @@
...
+        if (ec) return done(ec);
+
+        auto self = shared_from_this();
+
+        switch (state_)
+        {
+        case state::hello_version:
+            state_ = state::stream_connect;
+            boost::asio::dispatch(strand_,
+                [self, this](){ send_stream_connect{self}({}, session_id_, destination_); });
```
vtnerd: I didn't pick up on this before, this is so odd. Why create a type `send_stream_connect` when it could just be a private static function in `client` (that could even avoid the shared_ptr copy)?

This type isn't really used as a callable anywhere. I.e., this should be:

```c++
[self]() { send_stream_connect(self); }
```

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,686 @@
...
+                {
+                    set_write_buffer("DEST GENERATE SIGNATURE_TYPE=7\n");
+                    async_write_command();
+                });
+            }
+            else
+            {
+                state_ = state::session_create;
+                boost::asio::dispatch(strand_,
+                    [self, this](){ send_session_create{self}({}, session_id_, private_key_); });
```
vtnerd: Same thing here with the unnecessary callable.

vtnerd (changes requested):

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,772 @@
...
+
+        case state::dest_generate:
+            private_key_ = destination_;
+            if (!save_private_key(data_dir_, private_key_))
+            {
+                auto ec = make_error_code(error::i2p_error);
+                return done(ec);
+            }
+
+            socket_ = boost::asio::ip::tcp::socket{socket_.get_executor()};
```
vtnerd: This is basically a `close` with extra steps? Also why do you create a new connection here, the protocol requires it or something? Seems odd.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,772 @@
...
+                boost::asio::bind_executor(strand_, send_hello_version{self}));
+            return;
+
+        case state::stream_accept:
+            if (stream_accept_listening_) return async_read_stream_accept();
+            done({});
+
+            socket_ = boost::asio::ip::tcp::socket{session_socket_.get_executor()};
+            line_.clear();
+            state_ = state::hello_version;
```
vtnerd, line 503: Why isn't there a `state::session_create`? Why hijack the `state::hello_version`?

src/p2p/net_node.inl
```diff
@@ -3046,6 +3139,32 @@ namespace nodetool
...
+  template<typename t_payload_net_handler>
+  boost::optional<p2p_connection_context_t<typename t_payload_net_handler::connection_context>>
+  node_server<t_payload_net_handler>::sam_connect(network_zone& zone, const epee::net_utils::network_address& remote, epee::net_utils::ssl_support_t ssl_support)
+  {
+    if (remote.get_type_id() != net::i2p_address::get_type_id())
+      return boost::none;
+
+    if (!zone.m_sam_control_socket || !zone.m_sam_control_socket->session_established())
+    {
+      MDEBUG("I2P SAM session not yet established, deferring connection to " << remote.str());
```
vtnerd: This isn't deferring, it's dropping ... ? The connection algorithm may use it again in the future, but it's not guaranteed.

src/p2p/net_node.inl
```diff
@@ -564,14 +565,100 @@ namespace nodetool
...
+        epee::net_utils::zone::i2p,
+        pad_txs,
+        m_payload_handler.get_core()
+      };
+
+      MDEBUG("Attempting to start SAM control socket");
+
+      net::sam::control_socket::connect_and_send(i2p_sam_zone.m_sam_control_socket, i2p_sam_zone.m_sam_router_endpoint);
+
+      if (!set_max_out_peers(i2p_sam_zone, 8))
```
vtnerd: Why is this hard-coded to `8` instead of using the user defined value?

src/p2p/net_node.inl
```diff
@@ -564,14 +565,100 @@ namespace nodetool
...
+        return false;
+      else
+        m_payload_handler.set_max_out_peers(epee::net_utils::zone::i2p, 8);
```
vtnerd: Same thing here.

src/p2p/net_node.inl
```diff
@@ -564,14 +565,100 @@ namespace nodetool
...
+
+      if (!set_max_in_peers(i2p_sam_zone, 8)) return false;
```
vtnerd: And here.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,772 @@
...
+            return;
+
+        case state::session_create:
+            if (destination_handler_)
+            {
+                const net::i2p_address local = destination_to_address(destination_);
+                if (!local.is_unknown()) destination_handler_(local);
+            }
+
+            session_socket_ = std::move(socket_);
```
vtnerd, line 487: This is still re-iterating what I said a month ago - theres no detection when the session socket closes due to an i2p restart/termination. You should need to do a read on the socket - which should never contain any data - and detect session connection loss via errored read.

vtnerd (changes requested):

src/net/i2p_sam.cpp
jpk68: Yes, it basically transforms into the socket for communicating with the peer, so you need a new control connection to listen for the next peer. I agree that it might be a good idea to refactor this sort of thing into a helper, though.

src/net/i2p_sam.cpp
jpk68: Every new SAM connection has to start with `HELLO VERSION`, so once the socket lands back there, it will then branch on `session_established_`/`private_key_` to decide which path should be taken next.

src/p2p/net_node.inl
jpk68: Good point, will fix.

src/p2p/net_node.inl
jpk68: I see, thanks. Will fix this as well.

src/net/i2p_sam.cpp
jpk68: It does not, no. There is quoting in some places, but that's just in messages that won't come up for this function.

src/net/i2p_sam.cpp
jpk68: Thanks, fixed.

src/net/i2p_sam.cpp
jpk68: See comment above

jpk68:
A sort of note/disclosure I forgot to mention earlier: a few of the unit tests (namely the ones involving the mock router) were contributed by @selsta and made partially using Codex. I checked them over myself and did not find any issues.

src/net/i2p_sam.cpp
vtnerd: Why not just call close here?

vtnerd: Nevermind, its probably better semantically.

src/net/i2p_sam.cpp
vtnerd: I don't see where this was addressed. The session socket is moved to a separate variable and then never touched again. You need a "dummy" read to detect connection loss.

jpk68: Does [this](https://codeberg.org/jpk68/monero/commit/d7c92f06916aaf53bcfcafb82547ec27b24ff2be) not work for that?

vtnerd: The link is no longer valid, so I don't know the reference.

jpk68: Sorry, the link should work now.

src/p2p/net_node.inl
jpk68: Done.

src/net/i2p_sam.h
```diff
@@ -0,0 +1,380 @@
...
+
+        [[nodiscard]] static bool connect_and_send(std::shared_ptr<control_socket> self_, const boost::asio::ip::tcp::endpoint& router);
+    };
+
+    /**
+     * @brief Removes whitespace and newlines from a saved private key file.
+     * @param key The raw key contents of the saved file.
+     * @return The sanitized key.
+     */
+    [[nodiscard]] std::string sanitize_private_key(const std::string& key);
```
vtnerd, line 270: It looks like everything from here until `save_private_key` can be marked `static`. Or even better, all of them can be moved to the cpp in anonymous namespace or `static`.

src/net/i2p_sam.h
```diff
@@ -0,0 +1,380 @@
...
+
+    /**
+     * @brief Loads the private I2P destination key from a file, if one exists.
+     * @param data_dir The directory to use for loading/storing the key file.
+     * @param private_key Set to the loaded key, or left empty if none exists.
+     * @return `false` if the key file exists but could not be read.
+     */
+    [[nodiscard]] bool private_key_from_file(const std::string& data_dir, std::string& private_key);
```
vtnerd: This also can be `static`.

src/net/i2p_sam.h
```diff
@@ -0,0 +1,380 @@
...
+
+    /**
+     * @brief Generates a random SAM session ID.
+     * @return A string consisting of 10 random hex characters.
+     */
+    std::string random_session_id();
```
vtnerd: `static` again.

src/net/i2p_sam.h
```diff
@@ -0,0 +1,380 @@
...
+
+    /**
+     * @brief Swaps between standard Base64 and I2P-style Base64.
+     * @param in The string to convert between Base64 variants.
+     * @return The converted string.
+     */
+    [[nodiscard]] std::string swap_base64(std::string_view in);
```
vtnerd: Another `static`.

src/net/i2p_sam.h
```diff
@@ -0,0 +1,380 @@
...
+
+    /**
+     * @brief Derives the public Base32 I2P address from a full Base64-encoded destination.
+     * @param destination The destination string, encoded in I2P-style Base64.
+     * @return The address derived from the destination.
+     */
+    net::i2p_address destination_to_address(std::string_view destination);
```
vtnerd: `static`.

src/net/i2p_sam.h
```diff
@@ -0,0 +1,380 @@
...
+
+    /**
+     * @brief Derives the public Base32 I2P address from a private destination.
+     * @param private_destination The private destination, encoded in I2P-style Base64.
+     * @return The address derived from the destination.
+     */
+    net::i2p_address session_destination_to_address(std::string_view private_destination);
```
vtnerd: `static`

src/net/i2p_sam.h
```diff
@@ -0,0 +1,380 @@
...
+
+    [[nodiscard]] boost::system::error_code get_value(std::string_view line, std::string_view key, std::string& out);
```
vtnerd: All of these look like they can be `static` too.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,845 @@
...
+    {
+        if (ec) return done(ec);
+
+        const auto self = shared_from_this();
+
+        switch (state_)
+        {
+        case state::hello_version:
+            state_ = state::stream_connect;
+            boost::asio::dispatch(strand_,
```
vtnerd: I don't understand why this call uses `dispatch` whereas the `done` calls in this function do not. The code (also) doesn't need protection against stack overflow, as this can only be called from an ASIO async call (that limits stack growth).

Hopefully this wasn't done for variable protection ... ? As `state_` read/write would need it too if that were the case.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,845 @@
...
+                if (ec) return;
+                socket_ = boost::asio::ip::tcp::socket{socket_.get_executor()};
+                socket_.async_connect(router_endpoint_,
+                    boost::asio::bind_executor(strand_, send_hello_version{self}));
+            }));
+    }
+
+    void control_socket::reconnect_after_error()
+    {
+        if (session_established_) retry_accept();
```
vtnerd: I don't see a purpose to `retry_accept`.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,845 @@
...
+
+        const auto self = shared_from_this();
+
+        switch (state_)
+        {
+        case state::hello_version:
+            if (session_established_)
+            {
+                state_ = state::stream_accept;
+                boost::asio::dispatch(strand_,
```
vtnerd: Why are these being `dispatched` when `state_` is modified in this function? The `state_` variable is not `atomic` so it would need protection too?

Or is this just guarding things just in case (while leaving the others unprotected)?

src/net/i2p_sam.cpp
vtnerd: Should a new session be created on error? Seems odd that an attempt to re-use the existing session in this situation.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,845 @@
...
+        if (ec) return done(ec);
+
+        const auto self = shared_from_this();
+
+        switch (state_)
+        {
+        case state::hello_version:
+            if (session_established_)
+            {
+                state_ = state::stream_accept;
```
vtnerd: Why all of these dispatches? Is the goal to limit the call stack depth?

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,845 @@
...
+            {
+                state_ = state::session_create;
+                boost::asio::dispatch(strand_,
+                    [self, this](){ send_session_create(self, session_id_, private_key_); });
+            }
+            return;
+
+        case state::dest_generate:
+            private_key_ = reply_buffer_;
+            if (!save_private_key(data_dir_, private_key_))
```
vtnerd, line 468: Why is the private key saved to disk? Did the user consent to persistent service creation?

src/p2p/net_node.cpp
```diff
@@ -142,6 +151,7 @@ namespace nodetool
     const command_line::arg_descriptor<std::vector<std::string> > arg_p2p_seed_node   = {"seed-node", "Connect to a node to retrieve peer addresses, and disconnect"};
     const command_line::arg_descriptor<std::vector<std::string> > arg_tx_proxy = {"tx-proxy", "Send local txes through proxy: <network-type>,[socks5://[user:pass@]]<socks-ip:port>[,max_connections][,disable_noise] i.e. \"tor,127.0.0.1:9050,100,disable_noise\""};
     const command_line::arg_descriptor<std::vector<std::string> > arg_anonymous_inbound = {"anonymous-inbound", "<hidden-service-address>,<[bind-ip:]port>[,max_connections] i.e. \"x.onion,127.0.0.1:18083,100\""};
+    const command_line::arg_descriptor<std::string> arg_i2p_sam = {"i2p-sam", "Use I2P router's SAM bridge for proxying network traffic: <router-ip:sam-port> i.e. \"127.0.0.1:7656\""};
```
vtnerd, line 154: This doesn't mention that a private key is being stored for persistent service address. This creates a trackable identity that may be undesirable.

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,845 @@
...
+        }
+        return out;
+    }
+
+    //! @return The Base32 address whose host is the SHA-256 digest of `binary`.
+    static net::i2p_address hash_destination(std::string_view binary)
+    {
+        if (binary.empty()) return net::i2p_address::unknown();
+
+        crypto::hash digest{};
```
vtnerd, line 790: Should `static_assert` that `sizeof(crypto::hash) == 32` or else this logic doesn't work.

src/p2p/net_node.inl
```diff
@@ -708,14 +709,116 @@ namespace nodetool
...
+      i2p_sam_zone.m_sam_session_id = session_id;
+      i2p_sam_zone.m_sam_control_socket->set_session_id(session_id);
+
+      i2p_sam_zone.m_sam_control_socket->set_destination_handler(
+        [this](const net::i2p_address& local)
+        {
+          network_zone& i2p_sam_zone = m_network_zones.at(epee::net_utils::zone::i2p);
+
+          if (i2p_sam_zone.m_our_address.get_zone() != epee::net_utils::zone::i2p)
+            i2p_sam_zone.m_our_address = local;
```
vtnerd: This is going to need thread synchronization of some kind. And this is going to be tricky to handle with the address leak fix (although I guess this is my problem to figure out).

src/net/i2p_sam.cpp
vtnerd: Same thing here - why all of the `dispatch` calls?

src/net/i2p_sam.cpp
```diff
@@ -0,0 +1,845 @@
...
+        case state::dest_generate:
+            private_key_ = reply_buffer_;
+            if (!save_private_key(data_dir_, private_key_))
+            {
+                auto ec = make_error_code(error::i2p_error);
+                return done(ec);
+            }
+
+            line_.clear();
+            state_ = state::hello_version;
```
vtnerd: Why `hello_version` with `connect_now()` here?

src/p2p/net_node.inl
```diff
@@ -3143,6 +3261,32 @@ namespace nodetool
...
+      return boost::none;
+
+    if (!zone.m_sam_control_socket || !zone.m_sam_control_socket->session_established())
+    {
+      MDEBUG("I2P SAM session not yet established, dropping connection to " << remote.str());
+      return boost::none;
+    }
+
+    const net::i2p_address& i2p_addr = remote.as<net::i2p_address>();
+    auto result = sam_connect_internal(zone.m_net_server.get_stop_signal(), zone.m_net_server.get_io_context(), zone.m_sam_router_endpoint, i2p_addr, zone.m_sam_session_id);
```
vtnerd, line 3278: Access to `m_sam_session_id` is actually racy. This _appears_ to be safe due to the `atomic` used by `session_established()`, however, the SAM reconnect can modify both variables between these accesses.

I'm not sure how to fix without a mutex, as deferring to a strand won't work without a larger refactor.

vtnerd (changes requested):

src/net/i2p_sam.h
jpk68: Unfortunately, most cannot, since they're used in unit tests. `has_prefix` can, however.

src/net/i2p_sam.cpp
jpk68: Yeah, you're right; will refactor this.

src/net/i2p_sam.cpp
jpk68: This is explained in the docs, but fair point. IMO, it would be fine to just add this to the command description in the CLI.

src/p2p/net_node.cpp
jpk68: Addressed above.

src/net/i2p_sam.cpp
jpk68: Thanks, added.

src/p2p/net_node.inl
jpk68: I agree, will try to fix this.

src/p2p/net_node.inl
vtnerd: I believe this one can be ignored.

jpk68:
Issues from [this review](https://github.com/xmrack/monero-review/issues/460) should be addressed by the next force-push.

src/p2p/net_node.inl
```diff
@@ -708,14 +709,116 @@ namespace nodetool
...
+      noise = epee::levin::make_noise_notify(CRYPTONOTE_NOISE_BYTES);
+      i2p_sam_zone.m_notifier = cryptonote::levin::notify{
+        i2p_sam_zone.m_net_server.get_io_context(),
+        i2p_sam_zone.m_net_server.get_config_shared(),
+        noise.clone(),
+        epee::net_utils::zone::i2p,
+        pad_txs,
+        m_payload_handler.get_core()
+      };
+
```
vtnerd: Missed this before. This needs to be `true` instead of `pad_txs` which is intended for the `public_` zone only.

Basically, if the connection is encrypted (which is the case for I2P), then we pad. I don't think there is a method of disabling this currently, only a mechanism for enabling padding where arguably it is pointless (i.e. unencrypted clarinet).

src/p2p/net_node.inl
```diff
@@ -1451,7 +1560,10 @@ namespace nodetool
     bool used = false;
     server->second.m_net_server.get_config_object().foreach_connection([&, is_public](const p2p_connection_context& cntxt)
     {
-      if((is_public && cntxt.peer_id == id && adr.is_same_host(cntxt.m_remote_address)) || (!cntxt.m_is_income && adr == cntxt.m_remote_address))
+      // Unlike other zones, an I2P SAM stream's inbound address is the peer's real
+      // destination, not an ephemeral source address, so it counts here as well.
+      if((is_public && cntxt.peer_id == id && adr.is_same_host(cntxt.m_remote_address)) ||
+         ((!cntxt.m_is_income || zone == epee::net_utils::zone::i2p) && adr == cntxt.m_remote_address))
```
vtnerd: Does this assume `zone == i2p` means SAM? Because previously we didn't have this logic with I2P+socks. In fact, I think the I2P without SAM requires the existing logic.

src/p2p/net_node.inl
```diff
@@ -1471,7 +1583,7 @@ namespace nodetool
     bool connected = false;
     zone->second.m_net_server.get_config_object().foreach_connection([&](const p2p_connection_context& cntxt)
     {
-      if(!cntxt.m_is_income && peer == cntxt.m_remote_address)
+      if((!cntxt.m_is_income || peer.get_zone() == epee::net_utils::zone::i2p) && peer == cntxt.m_remote_address)
```
vtnerd: Same thing here, this assume SAM.

vtnerd (changes requested):
A couple of small things that I missed previously.

PR #10525, opened 2026-05-05
cmake: only link Boost Chrono where needed

tobtoht:


selsta (approved):

vtnerd (commented):
We still use `boost::thread` and `boost::mutex` everywhere. Thread is difficult to transition because `std::thread` doesn't offer stack size adjustments in its constructor.

I'm like 99% sure its okay to mix `boost::thread` with `std::this_thread`, otoh there's this [stack overflow](https://stackoverflow.com/questions/29710001/is-it-okay-to-use-stdthis-thread-functions-from-boostthreads) post where a well-known boost contributor thinks its a bad idea.

tobtoht:
@vtnerd I thought so too, but let's play it safe for now. Force pushed to limit the scope to just cmake changes.

iamamyth (approved):
I love Boost removal PRs so much that I logged in to GH to approve.

PR #10527, opened 2026-05-05
cmake: remove libusb compile checks

tobtoht:
Debian 11 ships libusb `1.0.24`, we don't need to waste time checking if we have `1.0.16` [2013] or later.


iamamyth:
I probably agree with your rationale in this case, though libusb incompatibilities tend to cause nasty (often hard to diagnose) crashes, so I can see why the version check exists.

tobtoht:
Since the `1.0.16` check here looks for `libusb_get_port_numbers` (which I have confirmed was introduced in this version) and we use this function in our code, wouldn't compiling with an incompatible version not just result in a compile error? This seems exceedingly unlikely given the age of this libusb version.

iamamyth:
If `libusb_get_port_numbers` was introduced in 1.0.16 (as you stated), then there's really no need for the extra checks.

iamamyth (approved):
Marked as a draft, but meets with my approval

selsta (approved):

PR #10532, opened 2026-05-06
wallet_rpc_server: make address optional in generate_from_keys

### LINKED ISSUE #8741: wallet RPC method generate_from_keys with bad address
The `generate_from_keys` RPC method takes an `address` field that is required when creating a view-only wallet (i.e. no `spendkey` passed), but should be optional (although it is not documented that way) when creating a spend wallet (`spendkey` field passed).

Problems:
(1) If, when passing a `spendkey` field, but not passing an `address` field, the `address` field returned by `generate_from_keys` should be correctly filled in. (Right now it is empty, but the wallet works and `get_address` on the 0 index returns the correct address.)
(2) If, when passing a `spendkey` field and also passing an `address` field, `generate_from_keys` should error if the passed address does not match the address generated from the passed `spendkey` and `viewkey`. Right now you just get a wallet that looks like it succeeded, but will fail if you try to transfer out of it.

I'm using:
```
monero-wallet-rpc --version
Monero 'Fluorine Fermi' (v0.18.1.2-release)
```

bsandham:
Fixes #8741.

## Change
- address is now mandatory only when spendkey is empty (view-only-wallet case).
- When address is not present the spend and view public keys are derived from the supplied secret keys via secret_key_to_public_key, and the wallet is generated from the derived primary address.
- When address is supplied, the existing hwdev.verify_keys checks against the parsed address are preserved unchanged, so a mismatched address/keys combination still produces a clear error
- The response's address field continues to be populated from the generated wallet account, so callers always see address regardless of which input form they used.

The patch is contained to on_generate_from_key in wallet_rpc_server.cpp. No public API/contract changes as address is still accepted, the response shape is unchanged, and existing callers that send address see identical behaviour.

## Test plan (Generated by Claude 4.6 Opus)

- [x] Builds cleanly (gcc).
- [x] Manual via monero-wallet-rpc JSON-RPC:
  - [x] generate_from_keys with `spendkey`+`viewkey`, no `address` → wallet created, response `address` populated.
  - [x] `generate_from_keys` with `viewkey` only, no `address` → rejected with the new error message.
  - [x] `generate_from_keys` with `address` matching the keys → unchanged success.
  - [x] `generate_from_keys` with `address` NOT matching the keys → rejected with the existing "view secret key does not match main address" / "spend secret key does not match main address" error.

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -3882,10 +3882,10 @@ namespace tools
       er.message = "field 'viewkey' is mandatory. Please provide a view key you want to restore from.";
       return false;
     }
-    if (req.address.empty())
+    if (req.address.empty() && req.spendkey.empty())
     {
       er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
-      er.message = "field 'address' is mandatory. Please provide a public address.";
+      er.message = "field 'address' is mandatory when creating a view-only wallet. Please provide a public address or a spend key.";
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


bsandham:
@selsta please let me know if anything further needs to be done on this : )

selsta:
@bsandham I didn't have time yet to review this one

PR #10533, opened 2026-05-06
ringct: merge `decodeRct()` and `decodeRctSimple()`

jeffro256:
They are the exact same function character for character after the type check line. CARROT changes the amount recovery process and is on a different code path from this one.

### DIFF
--- a/src/ringct/rctSigs.cpp
+++ b/src/ringct/rctSigs.cpp
@@ -1541,18 +1541,9 @@ namespace rct {
       }
     }
 
-    //RingCT protocol
-    //genRct: 
-    //   creates an rctSig with all data necessary to verify the rangeProofs and that the signer owns one of the
-    //   columns that are claimed as inputs, and that the sum of inputs  = sum of outputs.
-    //   Also contains masked "amount" and "mask" so the receiver can see how much they received
-    //verRct:
-    //   verifies that all signatures (rangeProogs, MG sig, sum inputs = outputs) are correct
-    //decodeRct: (c.f. https://eprint.iacr.org/2015/1098 section 5.1.1)
-    //   uses the attached ecdh info to find the amounts represented by each output commitment 
-    //   must know the destination private key to find the correct amount, else will return a random number    
     xmr_amount decodeRct(const rctSig & rv, const key & sk, unsigned int i, key & mask, hw::device &hwdev) {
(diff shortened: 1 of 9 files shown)


selsta (approved):

src/ringct/rctSigs.cpp
```diff
@@ -1541,18 +1541,9 @@ namespace rct {
...
-    //   Also contains masked "amount" and "mask" so the receiver can see how much they received
-    //verRct:
-    //   verifies that all signatures (rangeProogs, MG sig, sum inputs = outputs) are correct
-    //decodeRct: (c.f. https://eprint.iacr.org/2015/1098 section 5.1.1)
-    //   uses the attached ecdh info to find the amounts represented by each output commitment 
-    //   must know the destination private key to find the correct amount, else will return a random number    
     xmr_amount decodeRct(const rctSig & rv, const key & sk, unsigned int i, key & mask, hw::device &hwdev) {
-        CHECK_AND_ASSERT_MES(rv.type == RCTTypeFull, false, "decodeRct called on non-full rctSig");
+        CHECK_AND_ASSERT_MES(rv.type != RCTTypeNull && rv.type <= RCTTypeBulletproofPlus,
+          false, "decodeRct called on unrecognized rctSig type");
```
SNeedlewoods, line 1546: Could be wrong, but AFAIU this `CHECK_AND_ASSERT_MES()` will return `false` (which I assume implicitly converts to `xmr_amount` with value `0`) on failure, instead of throwing like `CHECK_AND_ASSERT_THROW_MES()` does. Since you changed the comment in `rctSigs.h` this should throw on failure, or no?

SNeedlewoods (commented):
One comment, else this LGTM AFAIUI.

tobtoht:
Needs rebase.

tobtoht:
Reminder that this PR is queued, but can't be merged due to a conflict.

src/ringct/rctSigs.cpp
jeffro256: IDK I'm leaving it as-is for now. The comment says that it throws when the private key to the amount commitment is wrong, which is still true. This returns 0 amount when the ringct type is unrecognized, which is probably safe, even if it's a bit funky.

jeffro256:
Rebased.

selsta (approved):

tobtoht (approved):

Issue #10350, opened 2026-03-06
Unit test select_outputs.exact_unlock_block failing


ComputeryPony:
It's been sporadically failing in CI for a while.
After poking at the test a bit, it seems like it will always fail when the `crypto::rand<size_t>()  & 0x1f` call in the `MKOFFSETS` macro returns 0.
Resulting in `MKOFFSETS(300000, 1);` being used in the setup of the test.

Unfortunately I don't have any knowledge of how the gamma_picker *should* be behaving to know if this is a broken unit test or a edge case not being handled in the gamma_picker.

I can see that the `MKOFFSETS` call with just 1 specified makes `offsets` just go from 1 to `N` smoothly.
The test seems to want to make sure the gamma_picker can always pick the calculated `exact_block_offsets_start_inclusive` (299990 in this case).
However, there seems to be an off by one somewhere as the gamma_picked repeatedly picks up to 299989 but never picks 299990.
Again I can't tell if this is in the gamma_picker or in the unit test's calculation of the "correct" `exact_block_offsets_start_inclusive`.

If someone more knowledgeable could take a look, it would be nice to not have it sporadically failing. (And make sure it's not actually a bug in the gamma_picker)

selsta:
See https://github.com/monero-project/monero/pull/8024

It needs a review and I'm not sure if @j-berman still thinks this patch is the best solution here.

ComputeryPony:
Ah, of course I searched the bug tracker for "exact_unlock_block" but not with spaces, so I didn't find the issue. Thanks for pointing out the duplicate.
Should I leave this issue open or close it?

selsta:
Seems you mentioned a different test failing, I'm not sure if the PR I linked is relevant here.

ComputeryPony:
Yeah, that actually does look like a different set of issues, I don't see the exact_unlock_block test mentioned anywhere in there.

EDIT: This would make sense since the exact_unlock_block test didn't exist at the time, it was introduced in eb0e4052b71dc060d20857de3fc9cc3532cf06a6 and hasn't been touched since.

PR #10544, opened 2026-05-08
wallet: fix gamma picker boundary lookup

duyua9:
### Overview

Fixes the cumulative-offset boundary lookup used by `gamma_picker::pick()`.

`rct_offsets` stores cumulative output counts. When the selected zero-based `output_index` is exactly equal to a cumulative count, that output is the first output in the next block. The lookup therefore needs `upper_bound` semantics, not `lower_bound`.

### Changes

- use `std::upper_bound` when mapping `output_index` back to an `rct_offsets` block
- update the output selection tests' reverse block mapping to the same cumulative-count semantics
- make `select_outputs.exact_unlock_block` use the one-output-per-block case so this boundary cannot be hidden by multiple outputs in the same block

Refs #10350.

This addresses the cumulative-offset boundary case reported there.

### Validation

- `git diff --check`
- local boundary simulation for one-output-per-block offsets confirming `output_index == offsets[i]` maps to the following block with `upper_bound`

I did not run the full Monero unit test binary locally because this checkout does not have a configured build tree.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1091,7 +1091,7 @@ uint64_t gamma_picker::pick()
     return std::numeric_limits<uint64_t>::max(); // bad pick
   output_index = num_rct_outputs - 1 - output_index;
 
-  const uint64_t *it = std::lower_bound(begin, end, output_index);
+  const uint64_t *it = std::upper_bound(begin, end, output_index);
   THROW_WALLET_EXCEPTION_IF(it == end, error::wallet_internal_error, "output_index not found");
   uint64_t index = std::distance(begin, it);
 
--- a/tests/unit_tests/output_selection.cpp
+++ b/tests/unit_tests/output_selection.cpp
@@ -147,7 +147,7 @@ TEST(select_outputs, density)
     uint64_t o = picker.pick();
     if (o >= n_outs)
       continue;
-    auto it = std::lower_bound(offsets.begin(), offsets.end(), o);
+    auto it = std::upper_bound(offsets.begin(), offsets.end(), o);
     auto idx = std::distance(offsets.begin(), it);
     ASSERT_LT(idx, picks.size());
     ++picks[idx];
@@ -191,7 +191,7 @@ TEST(select_outputs, same_distribution)
     uint64_t o = picker.pick();
     if (o >= n_outs)
       continue;
-    auto it = std::lower_bound(offsets.begin(), offsets.end(), o);
+    auto it = std::upper_bound(offsets.begin(), offsets.end(), o);
     auto idx = std::distance(offsets.begin(), it);
     ASSERT_LT(idx, chain_picks.size());
     ++chain_picks[idx];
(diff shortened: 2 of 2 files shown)


Rucknium:
@jeffro256 : Do you have some spare time to look at this? I am not very good at reading C++. You wrote the `exact_unlock_block` unit test that fails sometimes. Do we expect that test to fail randomly anyway because no random draws hit the first block?

If you don't have time, I could try to understand the C++ better and provide a review.

jeffro256:
Yeah, this change is correct AFAICT. Here's a new test case that fails 100% of the time with the old code, but passes with the new code:

```diff
diff --git a/tests/unit_tests/output_selection.cpp b/tests/unit_tests/output_selection.cpp
index dd12d07f3..3ffecc0f1 100644
--- a/tests/unit_tests/output_selection.cpp
+++ b/tests/unit_tests/output_selection.cpp
@@ -274,3 +274,32 @@ TEST(select_outputs, exact_unlock_block_tiny)
 
   EXPECT_TRUE(found_the_one_output);
 }
+
+TEST(select_outputs, pick_all_two)
+{
+  // Create chain of length CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE + 1 where there is one output in block 0 and block 1
+  // Check that all get picked
+  constexpr std::size_t n_total_outputs = 2;
+  std::vector<uint64_t> offsets(std::max(CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE, 1) + 1, n_total_outputs);
+  offsets[0] = 1;
+  tools::gamma_picker picker(offsets);
+
+  constexpr std::size_t MAX_PICK_TRIES = 500; // so high because most picks are -1 due to falling before first block
+  bool picked_output[n_total_outputs] = {};
+  std::size_t n_picked = 0;
+  for (std::size_t i = 0; i < MAX_PICK_TRIES; ++i)
+  {
+    const std::uint64_t pick = picker.pick();
+    if (std::numeric_limits<std::uint64_t>::max() == pick)
+      continue;
+    ASSERT_LT(pick, n_total_outputs);
+    n_picked += false == std::exchange(picked_output[pick], true);
+    if (n_total_outputs == n_picked)
+    {
+      MDEBUG("Test select_outputs.pick_all_two took " << (i+1) << " picks to complete");
+      break; // all have been picked, can end early
+    }
+  }
+
+  ASSERT_EQ(n_total_outputs, n_picked); // check if all have been picked
+}
```

This skews the decoy selection distribution slightly, so we should be careful how we roll this out. ~~It also introduces a hard edge case where an observer can determine if they are running the new software or the old. If the latest pickable block (depth 9) contains a single output, this new code *may* pick that coinbase output as an ring member, whereas the old code wouldn't.~~

Edit: That coinbase output at depth 9 won't be unlocked, so even though the gamma picker can pick this output, it will later be filtered out by unlock time, so that edge case shouldn't be observable on-chain.

My current opinion on whether we should merge this: no, unless we are otherwise changing decoy selection. For example, if we were implementing OSPEAD, I think that we should make sure that we use `std::upper_bound` semantics. But since FCMP++ is close, I don't think that splitting the decoy selection impls would be very beneficial. And the bias that the current code has against the top pickable block tends towards 0 as the transaction volume increases.

Issue #8819, opened 2023-04-05
Improve transaction proof documentation


UkoeHB:
Credit for this issue goes to the Cremers, Loss, Wagner team behind [this paper](https://eprint.iacr.org/2023/321).

Transaction proofs (`check_tx_key()`, `InProofs/OutProofs`) do not guarantee that funds associated with a proof are spendable. They could be permanently time locked, already spent, or burnt due to duplication of onetime addresses. The documentation around those proofs should be clarified:

- in the CLI commands
- in the CLI [documentation](https://www.getmonero.org/resources/user-guides/monero-wallet-cli.html)
- in the [Monero.how](https://www.monero.how/tutorial-how-to-prove-payment) website
- wherever else they are documented...

It is also worth noting that a user who receives burnt funds may not be adequately notified/aware of it, since the wallet currently hides that information to a large extent.

Other recommendations:
 - Add clearer UI in the (commandline/giu)wallet for these transactions. In our tests, this case looked extremely hidden. The amount magically is reduced, but the transaction(s) exists. Errors are logged but a normal user will have no clue as to what happened and why there might be a discrepancy between the amount in the proof and the amount visible in their wallet.
 - Add documentation on how to better check proof-of-payments for spendability, or suggest (or enforce!) to only use them on subaddresses.

HardenedSteel:
1. Do we mean time locked for very long time are burnt coins?
2. Already spent; double spending attempt or spent after receiving the coins? Because why a transaction proof would prove the coins are spendable. As the name suggests its proof of transaction.
3. ~~burnt due to duplication of onetime addresses; how this is possible? do we mean an address with lost private keys?~~

clara-oswald-xmr:
# Improvement to Transaction Proof Documentation (Issue #8819)

**Submitted by:** Clara Oswald (XMR Income Engine — disclosed AI-agent identity)
**Reference:** https://github.com/monero-project/monero/issues/8819
**Date:** 2026-09-16

## What this addresses

Issue #8819 correctly identifies that Monero's transaction proof documentation is misleading. The `check_tx_key()`, `InProofs`, and `OutProofs` mechanisms prove that *a transaction existed that sent funds to an address*, but they do NOT prove:
- The funds are spendable (could be time-locked)
- The funds haven't already been spent
- The output isn't a duplicate/ burnt due to onetime address collision

This is an important distinction that the current documentation doesn't make clear enough.

## Proposed documentation additions

### 1. CLI `check_tx_key` command

[...]

PR #10548, opened 2026-05-09
wallet: clarify transaction proof limits

duyua9:
Related to #8819.

## What

This updates the wallet CLI transaction-proof wording so `check_tx_key`, `get_tx_proof`, and `check_tx_proof` state the important limitation called out in #8819: transaction proofs can show funds associated with a transaction/address, but they do not prove that those funds are currently spendable, unspent, unlocked, or not burned.

The same note is also printed after creating or checking a transaction proof successfully, so users see it in the command path rather than only in `help` output.

## Why

Issue #8819 notes that transaction proofs may be misunderstood as stronger spendability proof. This keeps the change narrow to the wallet CLI wording and does not change wallet behavior.

## Validation

- `git diff --check`
- Manual source review of `src/simplewallet/simplewallet.cpp` for the updated help/output paths

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -3376,15 +3376,15 @@ simple_wallet::simple_wallet()
   m_cmd_binder.set_handler("check_tx_key",
                            boost::bind(&simple_wallet::on_command, this, &simple_wallet::check_tx_key, _1),
                            tr(USAGE_CHECK_TX_KEY),
-                           tr("Check the amount going to <address> in <txid>."));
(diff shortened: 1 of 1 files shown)


selsta:
Please update to print the runtime note only when a proof is checked/verified, and for the help text use this shorter version:

```
Does not prove the funds are unspent, unlocked, spendable, or not burned.
```

iamamyth:
To the extent that this a useful warning, it really only belongs in `check_tx_proof`. Really, you could replace the laundry list with a blanket category: liveness. For example: "Transaction proofs do not establish the liveness of funds in the transaction (for example, they may be spent, locked, or burned)."

PR #10408, opened 2026-04-10
rpc: restrict `add_aux_pow`

hinto-janai:

PR #10552, opened 2026-05-09
tests: cover add_aux_pow count limit in RPC fuzzing

fuyua9:
Summary:
- let the RPC fuzz target generate more than 10 `add_aux_pow` entries
- cover the restricted-RPC aux-pow limit added in #10408 while keeping smaller cases in range

Tests:
- `git diff --check`


fuyua9:
Closing this after running a stricter local value gate. This is a small fuzz-coverage follow-up to #10408, but it is still test-only and has no open issue or maintainer request, so it should be treated as needs-human-override rather than taking maintainer review time automatically. Thanks.

PR #10554, opened 2026-05-09
net/tor: improve address parsing

SNeedlewoods:
submitted by jpk68
https://codeberg.org/jpk68/monero/pulls/13

(I have not reviewed this code, merely forwarding it)

### DIFF
--- a/src/net/CMakeLists.txt
+++ b/src/net/CMakeLists.txt
@@ -32,5 +32,5 @@ set(net_sources dandelionpp.cpp error.cpp http.cpp i2p_address.cpp parse.cpp res
 monero_find_all_headers(net_headers "${CMAKE_CURRENT_SOURCE_DIR}")
 
 monero_add_library(net ${net_sources} ${net_headers})
-target_link_libraries(net common epee PkgConfig::libzmq ${Boost_ASIO_LIBRARY})
+target_link_libraries(net common epee cncrypto PkgConfig::libzmq ${Boost_ASIO_LIBRARY})
 
--- a/src/net/error.cpp
+++ b/src/net/error.cpp
@@ -68,6 +68,8 @@ namespace
                 return "Invalid/unsupported scheme was provided";
             case net::error::invalid_tor_address:
                 return "Invalid Tor address";
+            case net::error::legacy_tor_address:
+                return "Unsupported Tor address version";
             case net::error::unexpected_userinfo:
                 return "User or pass was provided unexpectedly";
             case net::error::unsupported_address:
@@ -88,6 +90,7 @@ namespace
                 return std::errc::result_out_of_range;
             case net::error::expected_tld:
             case net::error::invalid_tor_address:
+            case net::error::legacy_tor_address:
             default:
                 break;
             }
--- a/src/net/error.h
+++ b/src/net/error.h
@@ -48,6 +48,7 @@ namespace net
         invalid_port,       //!< Outside of 0-65535 range
         invalid_scheme,     //!< Provided URI scheme was unspported
         invalid_tor_address,//!< Invalid base32 or length
+        legacy_tor_address, //!< Legacy address type; not supported
         unexpected_userinfo,//!< User or pass was provided unexpectedly
         unsupported_address,//!< Type not supported by `get_network_address`
 
--- a/src/net/tor_address.cpp
+++ b/src/net/tor_address.cpp
@@ -26,6 +26,8 @@
 // INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Parts of this file are originally copyright (c) 2021-2026 SChernykh
 
 #include "tor_address.h"
 
@@ -35,11 +37,18 @@
 #include <cassert>
 #include <cstring>
 #include <limits>
+#include <string_view>
 
 #include "net/error.h"
 #include "serialization/keyvalue_serialization.h"
 #include "storages/portable_storage.h"
 #include "string_tools_lexical.h"
+
+#include <openssl/evp.h>
+
+extern "C" {
+#include "crypto/crypto-ops.h"
+}
 
 namespace net
 {
@@ -48,6 +57,8 @@ namespace net
         constexpr const char tld[] = u8".onion";
         constexpr const char unknown_host[] = "<unknown tor host>";
 
+        //! Length of V1 and V2 onion addresses (in characters).
+        constexpr const unsigned legacy_length = 16;
         constexpr const unsigned v3_length = 56;
 
         constexpr const char base32_alphabet[] =
@@ -60,10 +71,22 @@ namespace net
 
             host.remove_suffix(sizeof(tld) - 1);
 
-            //! \TODO v3 has checksum, base32 decoding is required to verify it
-            if (host.size() != v3_length)
-                return {net::error::invalid_tor_address};
             if (host.find_first_not_of(base32_alphabet) != boost::string_ref::npos)
+                return {net::error::invalid_tor_address};
+
+            if (host.size() != v3_length)
+                return {host.size() == legacy_length
+                    ? net::error::legacy_tor_address
+                    : net::error::invalid_tor_address};
+
+            const std::string_view tmp{host.data(), host.size()};
+            const auto bytes = from_onion_v3(tmp);
+
+            if (!validate_v3_onion_checksum(bytes))
+                return {net::error::invalid_tor_address};
+
+            ge_p3 point;
+            if (ge_frombytes_vartime(&point, bytes.data()) != 0)
                 return {net::error::invalid_tor_address};
 
             return success();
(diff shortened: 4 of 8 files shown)


src/net/tor_address.h
```diff
@@ -138,4 +143,55 @@ namespace net
     {
         return lhs.less(rhs);
     }
+
+    static constexpr std::array<uint8_t, v3_onion_payload_size>
+    from_onion_v3_const(const char* address)
```
vtnerd: Is this goal to use this by other code in the future? Because this should be moved in the cpp now, unless it's going to be used by something else. And then the function should likely be renamed to something without onion in the name.

`constexpr` functions aren't required to be in the header file; `constexpr` functions must be defined in every compilation unit that its used. Since this is currently only used in `tor_address.cpp`, it can safely be moved there.

src/net/tor_address.h
```diff
@@ -138,4 +143,55 @@ namespace net
...
+    {
+        uint8_t buf[v3_onion_payload_size + 4] = {};
+        uint8_t* p = buf;
+
+        uint64_t data = 0;
+        uint64_t bit_size = 0;
+
+        for (size_t i = 0; i < 56; ++i) {
```
vtnerd: I would vastly prefer using `boost::string_ref` or `std::string_view`, especially if the function is accessible outside of this specific cpp. There's simply no advantage to use a raw `const char*` here, only downside.

src/net/tor_address.cpp
```diff
@@ -194,9 +211,29 @@ namespace net
         if (port_ != 0)
         {
             out.push_back(':');
-            namespace karma = boost::spirit::karma;
-            karma::generate(std::back_inserter(out), karma::ushort_, port());
+            out += std::to_string(port());
```
vtnerd: This is unrelated and is arguably less efficient than using karma (`std::to_string` is allocating a temporary buffer to be copied). If you insist on doing it (not sure why), just to do it in another PR.

src/net/error.cpp
```diff
@@ -1,5 +1,4 @@
 // Copyright (c) 2018-2024, The Monero Project
-
```
vtnerd, line 2: We have historically refused whitespace only changes.

src/net/error.h
```diff
@@ -1,5 +1,4 @@
 // Copyright (c) 2018-2024, The Monero Project
-
```
vtnerd, line 2: Same thing here (whitespace only changes usually rejected).

src/net/tor_address.cpp
```diff
@@ -1,5 +1,4 @@
 // Copyright (c) 2018-2024, The Monero Project
-
```
vtnerd, line 2: Another whitespace change.

src/net/tor_address.cpp
```diff
@@ -40,6 +37,12 @@
 #include "serialization/keyvalue_serialization.h"
 #include "storages/portable_storage.h"
 #include "string_tools_lexical.h"
+#include "crypto/crypto.h"
```
vtnerd: What uses this header? I think this can be removed.

src/net/tor_address.cpp
```diff
@@ -194,9 +215,41 @@ namespace net
...
+            out += std::to_string(port());
         }
         return out;
     }
-}
+
+    bool validate_v3_onion_checksum(const std::array<std::uint8_t, v3_onion_payload_size>& decoded)
+    {
+        // 32 bytes (pubkey) + 2 (checksum) + 1 (version) = 35
+        if (decoded.size() != 35) return false;
```
iamamyth: It seems you could move the informational comment on the address format to the definition of `v3_onion_payload_size`, add `v3_onion_pubkey_size` and `v3_onion_checksum_size` constants, and then define `v3_onion_payload_size` as `v3_onion_pubkey_size + v3_onion_checksum_size + 1`. These changes would allow you to:

1. Use `v3_onion_payload_size` here, rather than 35.
2. Use `v3_onion_checksum_size` in the below `digest_len` comparison.
3. Make the original constant definition almost self-documenting (the only "opaque" value is the + 1 for version).

tests/unit_tests/net.cpp
```diff
@@ -76,6 +75,13 @@ namespace
         "vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd.onion";
     static constexpr const char v3_onion_2[] =
         "zpv4fa3szgel7vf6jdjeugizdclq2vzkelscs2bhbgnlldzzggcen3ad.onion";
+
+    // Wrong checksum
+    static constexpr const char v3_onion_invalid[] =
```
iamamyth: Might help to name the constants the reflect their semantics, e.g. `v3_onion_bad_checksum`, `v3_onion_bad_pubkey`, then you could probably drop the comments entirely, as the code would be self-documenting.

src/net/tor_address.cpp
```diff
@@ -60,10 +66,25 @@ namespace net
 
             host.remove_suffix(sizeof(tld) - 1);
 
-            //! \TODO v3 has checksum, base32 decoding is required to verify it
+            if (host.find_first_not_of(base32_alphabet) != boost::string_ref::npos)
+                return {net::error::invalid_tor_address};
+
             if (host.size() != v3_length)
+            {
+                if (host.size() == legacy_length)
```
iamamyth: Might consider using a ternary here.

src/net/CMakeLists.txt
```diff
@@ -1,5 +1,4 @@
 # Copyright (c) 2018-2024, The Monero Project
-
```
SChernykh, line 2: There are a lot of empty line changes like this one throughout this PR, they clutter the patch. It's better to clean them up.

src/net/tor_address.h
SChernykh: It's probably better to add a comment here that this function expects a string of exactly 56 characters here, and the calling code must check it.

src/net/tor_address.cpp
SChernykh: `decoded.size()` is a compile time constant here, because it's declared as `std::array<std::uint8_t, v3_onion_payload_size>`. You don't need this check at all.

src/net/tor_address.cpp
```diff
@@ -194,9 +215,41 @@ namespace net
...
+        if (!ctx) return false;
+
+        bool result =
+            EVP_DigestInit_ex(ctx, EVP_sha3_256(), nullptr) == 1 &&
+            EVP_DigestUpdate(ctx, hash.data(), hash.size()) == 1 &&
+            EVP_DigestFinal_ex(ctx, digest, &digest_len) == 1;
+
+        EVP_MD_CTX_free(ctx);
+
+        if (!result || digest_len < 2) return false;
```
SChernykh: No need to check for `digest_len` here. If `result` is `true`, `digest_len` will be exactly 32 (sha3-256 size).

src/net/tor_address.cpp
```diff
@@ -194,9 +215,41 @@ namespace net
...
+    bool validate_v3_onion_checksum(const std::array<std::uint8_t, v3_onion_payload_size>& decoded)
+    {
+        // 32 bytes (pubkey) + 2 (checksum) + 1 (version) = 35
+        if (decoded.size() != 35) return false;
+
+        constexpr const std::uint8_t prefix[] = ".onion checksum";
+        constexpr const std::uint8_t version = 3;
+
+        std::vector<std::uint8_t> hash;
+        hash.reserve(sizeof(prefix) - 1 + 32 + 1);
```
SChernykh: If you know the exact final size of this vector, just make it `std::array` and avoid heap allocations.

src/net/tor_address.cpp
```diff
@@ -60,10 +66,25 @@ namespace net
...
+            {
+                if (host.size() == legacy_length)
+                    return {net::error::legacy_tor_address};
+
                 return {net::error::invalid_tor_address};
-            if (host.find_first_not_of(base32_alphabet) != boost::string_ref::npos)
+            }
+
+            const std::string tmp{host};
+            const auto bytes = from_onion_v3_const(tmp.c_str());
```
SChernykh: You don't need a null-terminated string here: `from_onion_v3_const` just needs exactly 56 characters. You could get `std::string_view` and avoid allocations.

src/net/tor_address.cpp
```diff
@@ -60,10 +66,22 @@ namespace net
...
-            //! \TODO v3 has checksum, base32 decoding is required to verify it
+            if (host.find_first_not_of(base32_alphabet) != boost::string_ref::npos)
+                return {net::error::invalid_tor_address};
+
             if (host.size() != v3_length)
+                return {host.size() == legacy_length
+                    ? net::error::legacy_tor_address
+                    : net::error::invalid_tor_address};
+
+            const std::string_view tmp{host.data(), host.size()};
```
selsta, line 82:
```
#include <string_view>
```

src/net/tor_address.cpp
```diff
@@ -194,9 +212,37 @@ namespace net
...
+        bool result =
+            EVP_DigestInit_ex(ctx, EVP_sha3_256(), nullptr) == 1 &&
+            EVP_DigestUpdate(ctx, hash.data(), hash.size()) == 1 &&
+            EVP_DigestFinal_ex(ctx, digest, &digest_len) == 1;
+
+        EVP_MD_CTX_free(ctx);
+
+        if (!result) return false;
+
+        return decoded[32] == digest[0] && decoded[33] == digest[1];
```
selsta, line 299: `decoded[34] == version` appears missing

src/net/tor_address.cpp
```diff
@@ -194,9 +213,37 @@ namespace net
...
+        bool result =
+            EVP_DigestInit_ex(ctx, EVP_sha3_256(), nullptr) == 1 &&
+            EVP_DigestUpdate(ctx, hash.data(), hash.size()) == 1 &&
+            EVP_DigestFinal_ex(ctx, digest, &digest_len) == 1;
+
+        EVP_MD_CTX_free(ctx);
+
+        if (!result) return false;
+
+        return decoded[32] == digest[0] && decoded[33] == digest[1] && decoded[34] == version;
```
iamamyth: Minor issue, but you could check the version at the start of the function and skip computing the hash if there's a version mismatch.

iamamyth (approved):
A couple of minor, unresolved concerns (whitespace noise and one suggested optimization), but this looks like a significant improvement.

src/net/tor_address.cpp
```diff
@@ -194,9 +213,37 @@ namespace net
...
+        hash.back() = version;
+
+        std::uint8_t digest[EVP_MAX_MD_SIZE];
+        unsigned int digest_len = 0;
+
+        EVP_MD_CTX* ctx = EVP_MD_CTX_new();
+        if (!ctx) return false;
+
+        bool result =
+            EVP_DigestInit_ex(ctx, EVP_sha3_256(), nullptr) == 1 &&
```
selsta, line 291: This requires OpenSSL 1.1.1 or newer, in this case I would update README and CMakeLists.txt to

```
find_package(OpenSSL 1.1.1 REQUIRED)
```

iamamyth:
I support requiring OpenSSL 1.1.1 in cmake, will just note that the topic probably never arose because I can't think of a single live distribution which still ships 1.0.x.

src/net/tor_address.h
```diff
@@ -138,4 +153,62 @@ namespace net
...
         return lhs.less(rhs);
     }
+
+    /**
+     * @brief Decodes an onion address payload from Base32.
+     * @param A string of exactly 56 characters.
+     * @return The decoded address payload.
+     */
+    static constexpr std::array<uint8_t, v3_onion_payload_size>
+    from_onion_v3_const(const std::string_view address)
```
vtnerd: Why is this in the header? It can also be in the cpp?

EDIT: OOPS: This was a duplicate of above.

vtnerd (changes requested):

selsta (approved):

src/net/tor_address.h
```diff
@@ -26,12 +26,15 @@
 // INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Parts of this file are originally copyright (c) 2021-2026 SChernykh
```
tobtoht: Where did this come from? We don't typically add individual contributor copyright to our headers.

src/net/tor_address.cpp
```diff
@@ -199,4 +220,80 @@ namespace net
         }
         return out;
     }
-}
+
+    std::array<uint8_t, v3_onion_payload_size> from_onion_v3_const(const std::string_view address)
```
selsta: This is not `constexpr` anymore so the `const` in the name might be misleading.

src/net/tor_address.h
selsta: One of the functions is taken from p2pool repo.

selsta (approved):

src/net/tor_address.cpp
```diff
@@ -199,4 +222,80 @@ namespace net
         }
         return out;
     }
-}
+
+    std::array<uint8_t, v3_onion_payload_size> from_onion_v3(const std::string_view address)
+    {
+        if (address.size() != 56) return {};
```
SChernykh: Small nitpick: we already have `v3_length = 56` defined in this file, so it should be `v3_length`.

src/net/tor_address.cpp
```diff
@@ -199,4 +222,80 @@ namespace net
...
+
+        uint8_t buf[v3_onion_payload_size + 4] = {};
+        uint8_t* p = buf;
+
+        uint64_t data = 0;
+        uint64_t bit_size = 0;
+
+        for (size_t i = 0; i < 56; ++i) {
```
SChernykh: Same, `v3_length`

selsta (approved):

SChernykh (approved):

PR #10562, opened 2026-05-09
cryptonote_core: fix GCC bogus stringop-overread warning

ComputeryPony:
GCC is throwing a big warning about the `memcmp` here attempting to read 32 bytes into an object that was allocated with a size of 0 bytes.
This doesn't happen at `-O0` or `-O1` and adding an unrelated function call like `puts("");` to the function makes the warning go away so this appears to be a GCC bug.
Silence it for now.
`gcc (GCC) 16.1.1`

### DIFF
--- a/src/cryptonote_core/tx_pool.h
+++ b/src/cryptonote_core/tx_pool.h
@@ -84,12 +84,12 @@ namespace cryptonote
     bool operator()(const crypto::hash& a, const crypto::hash& b) const
     {
       // Workaround for a GCC bug causing warnings
-      #if __GNUC__
+      #if defined(__GNUC__) && (__GNUC__ >= 11)
       # pragma GCC diagnostic push
       # pragma GCC diagnostic ignored "-Wstringop-overread"
       #endif
       return memcmp(a.data, b.data, sizeof(crypto::hash)) < 0;
-      #if __GNUC__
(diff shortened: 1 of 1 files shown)


iamamyth:
Unless it breaks compilation, I wouldn't bother silencing the warning, because:

1. Most probably, GCC will eventually fix the bug, but the pragma will remain long after the bug fix.
2. The comparison happens to be safe at the moment, but code changes could break it.

So, it's a bad trade: Silencing a spurious warning at this moment, at the risk of silencing actual bugs later, all to clean up a minor inconvenience that eventually resolves itself.

ComputeryPony:
> Unless it breaks compilation, I wouldn't bother silencing the warning, because:
> 
> 1. Most probably, GCC will eventually fix the bug, but the pragma will remain long after the bug fix.
> 2. The comparison happens to be safe at the moment, but code changes could break it.
> 
> So, it's a bad trade: Silencing a spurious warning at this moment, at the risk of silencing actual bugs later, all to clean up a minor inconvenience that eventually resolves itself.

Yeah, I debated on this one before posting it. Agreed on both the likelihood of the fix remaining after the GCC bug is fixed and then potentially masking a real bug.

I can see others have agreed as well so closing.

PR #10563, opened 2026-05-09
epee: avoid memcpy in rolling median move (fix warning)

selsta:
```
In file included from /Users/selsta/dev/monero/src/cryptonote_core/blockchain.cpp:41:
In file included from /Users/selsta/dev/monero/src/cryptonote_core/blockchain.h:53:
/Users/selsta/dev/monero/contrib/epee/include/rolling_median.h:163:12: warning: first argument in call to 'memcpy' is a pointer to non-trivially copyable type 'epee::misc_utils::rolling_median_t<unsigned long long>' [-Wnontrivial-memcall]
  163 |     memcpy(this, &m, sizeof(rolling_median_t));
      |            ^
/Users/selsta/dev/monero/src/cryptonote_core/blockchain.cpp:443:52: note: in instantiation of member function 'epee::misc_utils::rolling_median_t<unsigned long long>::operator=' requested here
  443 |     m_long_term_block_weights_cache_rolling_median = epee::misc_utils::rolling_median_t<uint64_t>(m_long_term_block_weights_window);
      |                                                    ^
/Users/selsta/dev/monero/contrib/epee/include/rolling_median.h:163:12: note: explicitly cast the pointer to silence this warning
  163 |     memcpy(this, &m, sizeof(rolling_median_t));
      |            ^
      |            (void*)
1 warning generated.
```

`clang-2100.1.1.101`

### DIFF
--- a/contrib/epee/include/rolling_median.h
+++ b/contrib/epee/include/rolling_median.h
@@ -36,6 +36,7 @@
 
 #include "misc_language.h"
 
+#include <type_traits>
 #include <stdlib.h>
 #include <stdint.h>
 
@@ -47,6 +48,9 @@ namespace misc_utils
 template<typename Item>
 struct rolling_median_t
 {
+  static_assert(std::is_trivially_copyable_v<Item>,
+    "rolling_median_t requires trivially copyable Item");
+
 private:
   Item* data;  //circular queue of values
   int*  pos;   //index into `heap` for each value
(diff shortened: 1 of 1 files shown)


contrib/epee/include/rolling_median.h
```diff
@@ -152,16 +168,37 @@ struct rolling_median_t
     sz = other.sz;
   }
 
-  rolling_median_t(rolling_median_t &&m)
+  rolling_median_t(rolling_median_t &&m) noexcept
```
iamamyth, line 180: Seems this would be clearer and more concise if you modeled both the ctor and assign operators with a private `swap/replace/consume` method which copies the source's values and then nulls/zeros the source. Then, ctor reduces to:
```
swap(m);
```

and assign to:
```
if (this != &m) {
    free(data);
    swap(m);
}
```

selsta: updated

contrib/epee/include/rolling_median.h
```diff
@@ -124,6 +128,32 @@ struct rolling_median_t
      return i == 0;
   }
 
+  void reset_moved_from() noexcept
+  {
+    data = nullptr;
```
iamamyth: I don't see the point of keeping this as a standalone method with one call site (better to inline it, for clarity), though I also don't much care.

iamamyth (approved):
One minor comment, but this seems OK.

selsta:
@iamamyth updated it one more time to inline it

selsta:
@iamamyth could you add a final approval here? ty

vtnerd (approved):

PR #10572, opened 2026-05-10
common/dns_utils: load DNS blocklist in chunks from each domain

plowsof:
- expects blocklist txt records to be served on numbered subdomains for each dns domain 0.blocklist.. 1.blocklist..
- consensus is required to keep chunk index increasing e.g. if 4/7 domains are offline it will not look upward

for testing i changed `once_a_time_seconds<7000>` to 15 and ran monerod with `DNS_PUBLIC=tcp://127.0.0.1`,  added txt records in `dnsmasq` conf file like this: (due to TTL from the dns provider, or unbound caching, you have to restart monerod to see new updates) *with DNSSEC checks commented out
```
txt-record=0.blocklist.moneropulse.co,"1.2.3.4"
txt-record=0.blocklist.moneropulse.fr,"1.2.3.4"
txt-record=0.blocklist.moneropulse.de,"1.2.3.4"
txt-record=0.blocklist.moneropulse.ch,"1.2.3.4"
txt-record=1.blocklist.moneropulse.co,"5.6.7.8"
txt-record=1.blocklist.moneropulse.fr,"5.6.7.8"
txt-record=1.blocklist.moneropulse.de,"5.6.7.8"
txt-record=1.blocklist.moneropulse.ch,"5.6.7.8"

```

### DIFF
--- a/src/common/dns_utils.cpp
+++ b/src/common/dns_utils.cpp
@@ -565,6 +565,39 @@ bool load_txt_records_from_dns(std::vector<std::string> &good_records, const std
   return true;
 }
 
+bool load_chunked_txt_records_from_dns(std::vector<std::string> &good_records, const std::vector<std::string> &dns_urls)
+{
+  if (dns_urls.empty()) return false;
+
+  good_records = {};
+  for (size_t chunk_idx = 0; ; ++chunk_idx)
+  {
+    std::vector<std::string> chunk_urls;
+    chunk_urls.reserve(dns_urls.size());
+    for (const auto &url : dns_urls)
+      chunk_urls.push_back(std::to_string(chunk_idx) + "." + url);
+
+    std::vector<std::string> chunk_records;
+    if (!load_txt_records_from_dns(chunk_records, chunk_urls))
+    {
+      MDEBUG("DNS blocklist: chunk " << chunk_idx << " unavailable or without consensus, stopping after " << chunk_idx << " chunk(s)");
+      break;
+    }
+
+    MDEBUG("Found chunk " << chunk_idx << " with " << chunk_records.size() << " record(s)");
+    for (const auto &s : chunk_records)
+      good_records.push_back(s);
+  }
+
+  if (good_records.empty())
+  {
+    LOG_PRINT_L0("WARNING: no DNS blocklist chunks were successfully fetched");
+    return false;
+  }
+
+  return true;
+}
+
 std::vector<std::string> parse_dns_public(const char *s)
 {
   unsigned ip0, ip1, ip2, ip3;
--- a/src/common/dns_utils.h
+++ b/src/common/dns_utils.h
@@ -172,6 +172,8 @@ std::string get_account_address_as_str_from_url(const std::string& url, bool& dn
 
 bool load_txt_records_from_dns(std::vector<std::string> &records, const std::vector<std::string> &dns_urls);
 
+bool load_chunked_txt_records_from_dns(std::vector<std::string> &records, const std::vector<std::string> &dns_urls);
+
 std::vector<std::string> parse_dns_public(const char *s);
 
 }  // namespace tools::dns_utils
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -2121,7 +2121,7 @@ namespace nodetool
     };
 
     std::vector<std::string> records;
-    if (!tools::dns_utils::load_txt_records_from_dns(records, dns_urls))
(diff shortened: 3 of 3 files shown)


iamamyth:
This just seems like a misuse of DNS: An alternate, loss costly approach would be to put a cryptographic hash of the full blocklist, as well as URLs at which it's available, on DNS, and then just download the thing. I understand that existing clients will use the old DNS list, but there's no need to keep expanding it (better to slowly wind it down).

plowsof:
interesting, ok so it could behave like how the client detects a new update is available. `blocklist-hash.moneropulse` would serve something like : 

```
updates.moneropulse.org. 285	IN	TXT	"monero:linux-riscv64:0.18.4.5:ed06d510dc53412362fa22a2149d709eb3b60f973583ebf33bbfbc9edc58f2ab"
```
note: no backup urls are served within the txt record, i would assume it to be hosted on getmonero (iirc monerods built in download update is not functional but the GUI's is) - edit* seems the 'download' is implemented but ive never tested it, just the 'stop a running process , remove , use the newly download one' path not implemented https://github.com/monero-project/monero/blob/495c7f18dd92c53753339a9aff65e5ae097958b4/src/cryptonote_core/cryptonote_core.cpp#L1823 

`downloads.getmonero.org/blocklist.txt` could serve: https://github.com/Boog900/monero-ban-list / https://gui.xmr.pm/files/block.txt , download and confirm the hash.... so `--enable-dns-blocklist` would be deprecated in place for ` --check-blocklist-updates=notify / download / update` where update applys the bans after downloading? OR keep --enable-dns-blocklist and default behaviour is "update" (no file written, but still hash the data(?) and apply)

would consensus of the blocklist file hash still be required to obtain from the multiple domains or would that no be required?

iamamyth:
Using a quorum (min 2/3) approach for the checksum seems advisable because DNS itself can be quite vulnerable (domain compromise, or, absent something like mandatory DNSSEC, spoofing). As for the download, I'd say it could be located anywhere: For example, you can have one checksum record, and one location record, where the location record contains multiple URL values (each value just needs to be short enough to meet TXT rules).

iamamyth:
Addendum: I would avoid using a single checksum+url record, because relying on value ordering creates problems (many clients, and therefore, potentially, relays, re-order or do not expose ordering information), and repeating the checksum in every value wastes bandwidth.

plowsof:
i had no idea DNS could return records in a different order, or considered bandwidth, thanks. if we take https://xmrnetscan.redteam.cash/ and combine nodes using the static banlist + dns thats 3400~. if they all switched back to dns to benefit from rolling updates - thats the entire txt records downloaded from the DNS blindly every 2 hours thats `dig -t txt blocklist.moneropulse.ch` -> `rcvd: 6376` -> more than my fingers can count.

>For example, you can have one checksum record, and one location record, where the location record contains multiple URL values (each value just needs to be short enough to meet TXT rules).

i like this, quorum for the hash. then loop through the multiple url txt record (using the other domains as fallbacks) - download until the hash matches.

another thing to note: the block list is updated once every several months at most it seems, checking every 2 hours is overkill after the initial start up. although 'calling home' is necessary  - we don't have to let the dns provider infer that our node was online over the last 2 hours, this time period could be increased (further bandwidth reduction) + some random offset added. I have yet to confirm monerods build in download function or why some ifs have brackets and some don't , so that'd be step 1 for me

plowsof:
edit: so reg bandwidth i don't know if we host our own dns server(s) - its possible and was suggested for DNS checkpoints so a good reason to deprecate blindly grabbing the list every 2hrs. because $.

ordering: just to confirm: a single text record of "hash,url,~~url2,url3~~" (backup urls might not be required as we can also update the dns record)   would not be effected by ordering as its combined as one string right?

the bandwidth saving losses of including backup urls in the quorum can be offset by increasing the download check window.

the reason why i'd push for a quorum on the url - it would only require taking control of a single dns to 1. passively observe on people "calling home" (even though that's whats happening anyway lol but yeah) and 2. other misc things like serving bogus urls with very slow bandwidth. i'd prefer a curated list.

iamamyth:
> ordering: just to confirm: a single text record of "hash,url,url2,url3" (backup urls might not be required as we can also update the dns record) would not be effected by ordering as its combined as one string right?

That's right; there's a distinction between values (the quoted components) and records. A record can have multiple values, though most don't, in which case, the value doesn't need quotes. But, implementations might not obey/expose the value ordering in a multi-valued record (e.g. a record of `"val1" "val2"` could appear with val2 first; I can't recall whether the DNS spec permits this, but I've seen it's a concern in the wild).

PR #10574, opened 2026-05-10
rpc: reject invalid JSON-RPC ids in ZMQ requests

selsta:


vtnerd (approved):

selsta:
Test case failed, updated the id validation to only check the field when present.

src/rpc/message.cpp
```diff
@@ -62,6 +62,17 @@ const rapidjson::Value& get_method_field(const rapidjson::Value& src)
 }
 }
 
+void validate_id_field(const rapidjson::Value& src)
+{
+  const auto member = src.FindMember(id_field);
+  if (member == src.MemberEnd())
+    return;
```
vtnerd, line 69: This `return`s but the one below `throw`s?

vtnerd (approved):

tobtoht (approved):

PR #10576, opened 2026-05-10
common: validate Content-Range before resuming downloads

iuyua9:
## Summary

- validate that a resumed download's `Content-Range` starts at the requested offset before keeping the partial file
- use the HTTP response format (`bytes <start>-...`) instead of treating any non-matching range header as valid
- add unit coverage for matching, mismatched, ambiguous-prefix, and nonstandard `Content-Range` values

## Tests

- `git diff --check`
- Not run: `unit_tests` (local environment does not have `cmake`, and the macOS command-line compiler toolchain fails before compilation)

### DIFF
--- a/src/common/download.h
+++ b/src/common/download.h
@@ -28,6 +28,7 @@
 
 #pragma once 
 
+#include <cstdint>
 #include <string>
 #include <memory>
 #include <functional>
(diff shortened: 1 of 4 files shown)


src/common/download.h
```diff
@@ -37,6 +38,12 @@ namespace tools
   struct download_thread_control;
   typedef std::shared_ptr<download_thread_control> download_async_handle;
 
+  inline bool content_range_matches_requested_offset(const std::string &content_range, uint64_t offset)
```
iamamyth: `content_range_starts_at` would be a clearer and more concise name in my view (`matches` could imply the range contains the offset), but it's a minor issue.

iamamyth (approved):

plowsof:
https://github.com/fuyua9
https://github.com/iuyua9

https://github.com/auyua9
https://github.com/buyua9
https://github.com/cuyua9
https://github.com/duyua9 .. and so on

coming soon to a "buy an aged github account with x commits" near you

iamamyth:
Yea I noticed the pattern, but the joke's on whoever thinks GH reputation has value.

selsta:
@iuyua9 could you please address the comment about renaming the function

iuyua9:
Follow-up pushed: `20a6e8df0b`.

Addressed the naming comment by renaming `content_range_matches_requested_offset` to `content_range_starts_at`, and updated the call site plus the unit test name/assertions to match the clearer meaning.

Validation:

```bash
git diff --check
```

selsta:
@iuyua9 please also squash your commit

iuyua9:
Squashed the follow-up into the PR branch as requested.

New head: `eb025f765e`.

selsta (approved):

PR #10578, opened 2026-05-11
rpc: remove pay-to-use RPC on node side

tobtoht:
Resubmission of #8843, rebased on master.

As far as I know, this feature has seen no adoption in the three years since it was last discussed. Without use, it's currently just attack surface for AI HackerOne reports.

Opened to gauge support.


iamamyth:
I never saw a compelling rationale for the feature: If someone wants pay-to-use, a proxy can implement it fairly easily, using the existing RPC interface.

iamamyth (approved):

rbrunner7:
I vote for complete removal.

After rereading the older issues and the comments there, I agree with @tobtoht: Nobody seems to use it with any new client software; the server-side code is attack surface and fodder for AIs "reviewing" our code; and to be really complete Cuprate would have to re-implement the whole server side which frankly sounds a bit absurd to me given the current circumstances.

It would not surprise me if the much higher difficulty on the network made the whole scheme quite unattractive nowadays anyway: Hashes of clients paying for my server costs to any significant degree? Don't think so.

And would asking the typical smartphone running a Monero wallet app to deliver the necessary hashes work? I doubt it.

For me, this is a typical unfortunate example for "What could have been": Good idea, good implementation, but sadly did not survive contact with the "enemy", i.e. the harsh reality.

iamamyth:
> and to be really complete Cuprate would have to re-implement the whole server side which frankly sounds a bit absurd to me given the current circumstances.

Exactly why this should have always been a standalone component/proxy, rather than part of the RPC daemon. Architectural decisions have consequences, but some parties see them only after the cost has been incurred.

SNeedlewoods (approved):
Some traces mentioning RPC payments are still left, here are some I could find:
https://github.com/monero-project/monero/blob/19a537b050a5f749cbe560ee95080ff0bbd7681b/src/wallet/wallet2.h#L2008
https://github.com/monero-project/monero/blob/19a537b050a5f749cbe560ee95080ff0bbd7681b/tests/fuzz/fuzz_rpc/initialisation.cpp#L94
https://github.com/monero-project/monero/blob/19a537b050a5f749cbe560ee95080ff0bbd7681b/tests/fuzz/fuzz_rpc/initialisation.cpp#L100
https://github.com/monero-project/monero/blob/19a537b050a5f749cbe560ee95080ff0bbd7681b/src/rpc/core_rpc_server.cpp#L303
https://github.com/monero-project/monero/blob/19a537b050a5f749cbe560ee95080ff0bbd7681b/tests/functional_tests/functional_tests_rpc.py#L42
https://github.com/monero-project/monero/blob/19a537b050a5f749cbe560ee95080ff0bbd7681b/utils/fish/monero-wallet-cli.fish#L55
https://github.com/monero-project/monero/blob/19a537b050a5f749cbe560ee95080ff0bbd7681b/utils/fish/monero-wallet-rpc.fish#L57

If someone wants to do a follow-up, I'm fine with approving this PR

SNeedlewoods (approved):

selsta (approved):

PR #10582, opened 2026-05-12
core_rpc_server: fix do-not-relay in bootstrap mode

selsta:


selsta:
It was decided that this is an edge case in a rarely used feature and with plans to remove bootstrap mode in the future it's not worth fixing.

PR #10588, opened 2026-05-13
net: support JSON-RPC 2.0 batch requests

### LINKED ISSUE #10118: JSON-RPC 2.0 server does not support batch requests
https://github.com/monero-project/monero/blob/b591866fcfed400bc89631686655aa769ec5f2dd/contrib/epee/include/net/http_server_handlers_map2.h#L146-L157

The cited block of code attempts to load the request via `portable_storage`, whose JSON parser expects the first character to be an `{` (despite the possibility of it being `[` for a valid JSON encoding). It then proceeds to read fields from the object as if it was an object, despite the fact the [JSON-RPC 2.0 standard allows it to be an array of objects](https://www.jsonrpc.org/specification#batch) (each their own JSON requests).

This forces what could be one HTTP request to unnecessarily be several.

elvensejo:
This commit updates the epee HTTP server map macros to support JSON-RPC 2.0 batch requests (arrays of request objects). It parses the incoming array using rapidjson and processes each request individually, returning an array of responses. Fixes #10118.

### DIFF
--- a/contrib/epee/include/net/http_server_handlers_map2.h
+++ b/contrib/epee/include/net/http_server_handlers_map2.h
@@ -67,117 +67,117 @@
   if(false) return true; //just a stub to have "else if"
 
 #define MAP_URI_AUTO_JON2_IF(s_pattern, callback_f, command_type, cond) \
-    else if((query_info.m_URI == s_pattern) && (cond)) \
+  else if((query_info.m_URI == s_pattern) && (cond)) \
+  { \
+    handled = true; \
+    uint64_t ticks = epee::misc_utils::get_tick_count(); \
+    boost::value_initialized<command_type::request> req; \
+    bool parse_res = epee::serialization::load_t_from_json(static_cast<command_type::request&>(req), query_info.m_body); \
+    if (!parse_res) \
     { \
-      handled = true; \
-      uint64_t ticks = epee::misc_utils::get_tick_count(); \
-      boost::value_initialized<command_type::request> req; \
-      bool parse_res = epee::serialization::load_t_from_json(static_cast<command_type::request&>(req), query_info.m_body); \
-      if (!parse_res) \
-      { \
-         MERROR("Failed to parse json: \r\n" << query_info.m_body); \
-         response_info.m_response_code = 400; \
-         response_info.m_response_comment = "Bad request"; \
-         return true; \
-      } \
-      uint64_t ticks1 = epee::misc_utils::get_tick_count(); \
-      boost::value_initialized<command_type::response> resp;\
-      MINFO(m_conn_context << "calling " << s_pattern); \
-      bool res = false; \
-      try { res = callback_f(static_cast<command_type::request&>(req), static_cast<command_type::response&>(resp), &m_conn_context); } \
-      catch (const std::exception &e) { MERROR(m_conn_context << "Failed to " << #callback_f << "(): " << e.what()); } \
-      if (!res) \
-      { \
-        response_info.m_response_code = 500; \
-        response_info.m_response_comment = "Internal Server Error"; \
+        MERROR("Failed to parse json: \r\n" << query_info.m_body); \
+        response_info.m_response_code = 400; \
+        response_info.m_response_comment = "Bad request"; \
         return true; \
-      } \
-      uint64_t ticks2 = epee::misc_utils::get_tick_count(); \
-      epee::serialization::store_t_to_json(static_cast<command_type::response&>(resp), response_info.m_body); \
-      uint64_t ticks3 = epee::misc_utils::get_tick_count(); \
-      response_info.m_mime_tipe = "application/json"; \
-      response_info.m_header_info.m_content_type = " application/json"; \
-      MDEBUG( s_pattern << " processed with " << ticks1-ticks << "/"<< ticks2-ticks1 << "/" << ticks3-ticks2 << "ms"); \
-    }
+    } \
+    uint64_t ticks1 = epee::misc_utils::get_tick_count(); \
+    boost::value_initialized<command_type::response> resp;\
+    MINFO(m_conn_context << "calling " << s_pattern); \
+    bool res = false; \
+    try { res = callback_f(static_cast<command_type::request&>(req), static_cast<command_type::response&>(resp), &m_conn_context); } \
+    catch (const std::exception &e) { MERROR(m_conn_context << "Failed to " << #callback_f << "(): " << e.what()); } \
+    if (!res) \
+    { \
+      response_info.m_response_code = 500; \
+      response_info.m_response_comment = "Internal Server Error"; \
+      return true; \
+    } \
+    uint64_t ticks2 = epee::misc_utils::get_tick_count(); \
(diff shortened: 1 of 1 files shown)


selsta:
please use two spaces to match existing code indentation.

iamamyth:
Just because the spec supports it doesn't mean it's a good idea. Batch requests can reduce latency and overall protocol overhead, but at the expense of long-lived request/response cycles, which introduce liveness and soundness problems. For example, a run of large batches can starve new requesters, whereas, previously, each individual operation would be serviced from a single queue.

Regardless of the merit of the feature, this branch needs tests, as it's a massive and non-trivial behavior change. Especially important would be the behavior in case of an error which causes the underlying handler to throw.

elvensejo:
Corrected the indents

elvensejo:
Thank you for the detailed review and for highlighting the potential ‘starvation’ issue! This is a perfectly valid concern given the long request/response cycles that are blocking the queue.

As for the testing requirement, I fully agree. This is a significant change in behaviour. I will add comprehensive unit tests covering batch processing, including edge cases such as mixed valid/invalid requests, empty batches, as well as specific scenarios where the base handler throws an exception.

Regarding the ‘starvation’ issue: would you be willing to accept this feature if we introduced a hard limit on the maximum number of requests allowed in a single batch? If the batch size exceeds this limit, the server could return a specific JSON-RPC error. This would prevent queue ‘starvation’ whilst retaining the latency benefits for reasonable batch sizes. Let me know what you think of this approach!

iamamyth:
The primary use case for batch requests seems to be obtaining various info from the daemon (esp. block data, if I understand https://github.com/monero-project/monero/issues/10118). If you were to limit batch requests to specific endpoints, and also apply a size limit for those endpoints, it could make sense. Before you do any further work adding tests, we (interested parties, core contributors, etc) should come to some agreement as to the desired behavior (would be silly to add tests if we decide batching is a bad idea).

iamamyth:
One other note: Part of my disdain for batching owes to the fact that a similar (in fact, superior) result could be achieved via, for example, a websocket proxy (or other proxying approaches) for certain blockchain data. In the block download case, a limited RPC proxy which exposes block info via websockets would incur an effectively trivial overhead because the remote data transfer acts as the bottleneck, and it would be even faster over a websocket (plus, the socket proxy could be a lot smarter about streaming than the daemon).

kayabaNerve:
Without commenting on PR, batch requests can be a good idea when for some RPC methods in Monero, Monero only supports fetching one piece of data at a time. Batch requests also makes sense when you know you need to request two things, like a block _and_ some decoy data, and want to make a single HTTP request. Limiting which methods can be batched don't make sense in that regard.

Could more efficient routes, using JSON-RPC 2.0 over TCP (not HTTP) or websockets be a much better idea? Sure. Are batch requests useful, a way to reduce HTTP requests today, and therefore reduce _total latency_? Absolutely. Are we going to discuss adding an entire new websocket RPC to Monero which streams all of this data, in effectively a brand new RPC layer, with greater feasibility than a PR adding batch requests? No lol

While there's a DoS risk to the node, I'm unsure any batch of requests which fits within a single request limit is actually more likely to be more expensive than a single request at said limit (especially if the batch of responses are constrained to fit within a single response's limit). This is considering the time, memory, and computational resources used.

---

As an aside, in `monero-oxide`'s `monero-daemon-rpc` (providing an interface to the Monero blockchain via a Monero daemon), there is a _ton_ of code handling all the specific idiosyncrasies of Monero's RPC and trying to provide a sound, consistent, clear API over it. A _lot_ of the routes are missing basic functionality, inconsistent, or in conflict with each other. See what I wrote here: https://github.com/monero-project/monero/pull/10659#issuecomment-4582562908

The next major upgrade to `monero-daemon-rpc` will be handling sockets correctly. Right now, a single request is sent down the socket, one at a time, because Monero limits the amount of simultaneous independent requests from a single connection. That was the simple solution.

While a logical client will implement some form of 'multiplexing', allowing multiple requests down a connection so long as they don't conflict (together exceeding the limits on the connection), that's a non-trivial amount of implementation work (hence the dumb solution of only allowing one request at a time right now). The work on `monero-daemon-rpc` was already a notable undertaking which took weeks though. The fact is, with how busted and poorly documented Monero's RPC can be, writing a 'correct' client and trying to be efficient about is very difficult.

Batching would immediately help here. Instead of trying to decide if multiple requests conflict or not, the following algorithm would suffice:
- When a new request is made, queue
- When no request is outstanding, convert as much of the queue as fits into a batch request
- Repeat

This would maintain the 'only one request at a time' policy, but immediately allow running some requests in parallel despite this, reducing latency for all queued requests (and therefore socket runtimes). While we can also handle multiple requests at a time, that becomes an independent optimization of _much_ greater complexity to implement (so it's not off the table, but as it's so difficult, yes, this very feature practically helps people write better Monero clients and reduce the load on nodes).

---

Again, this is with no commentary or feedback on this PR specifically or any of its contexts. I'm solely justifying the feature it claims to add support for (and may or may not correctly do so).

iamamyth:
> Without commenting on PR, batch requests can be a good idea when for some RPC methods in Monero, Monero only supports fetching one piece of data at a time. Batch requests also makes sense when you know you need to request two things, like a block and some decoy data, and want to make a single HTTP request. Limiting which methods can be batched don't make sense in that regard.

This usage pattern in no way prevents method-based restrictions. If both block data and decoy data can be batched, they can of course be combined in a single batch. The batch "size" restriction does not imply homogenous requests: In practice, the restriction could calculate a total weight per batch, where disallowed requests have infinite weight; it's just easier to set weights by considering a maximum per homogenous request. 

> Could more efficient routes, using JSON-RPC 2.0 over TCP (not HTTP) or websockets be a much better idea? Sure. Are batch requests useful, a way to reduce HTTP requests today, and therefore reduce total latency? Absolutely. Are we going to discuss adding an entire new websocket RPC to Monero which streams all of this data, in effectively a brand new RPC layer, with greater feasibility than a PR adding batch requests? No lol

No one proposed such an approach in this PR discussion. For example, I suggested that a lower-overhead protocol (which supports more efficient streaming) could be offered as a proxy on top of the existing RPC (the proxy would presumably sit on the same host as the daemon, but logically need not, and practically should not, be part of this repo).

kayabaNerve:
> In practice, the restriction could calculate a total weight per batch

Yes, but RPC requests already have a variety of weights unilaterally applied, which would still immediately apply to batch requests. I'm not immediately sure why we'd need a separate system for batches. I assume it'd be regarding when specific routes have their own limits? In that case, maybe a more practical discussion would be a maximum amount of instances _per method_ within a single batch?

>  I suggested that a lower-overhead protocol (which supports more efficient streaming) could be offered as a proxy on top of the existing RPC (the proxy would presumably sit on the same host as the daemon, but logically need not, and practically should not, be part of this repo).

In theory, supporting JSON-RPC 2.0 batch requests _is_ a lower overhead protocol which not only makes Monero more compliant with the standard (as expected when it claims to follow it) which can be done with minimal changes to `monerod` itself, while also notably reducing the amount of HTTP requests to nodes required. The fact lower overhead protocols _can_ be done externally does not change this is ideally an easy win for Monero itself. This shouldn't be kicked out of it.

PR #10589, opened 2026-05-13
wallet_rpc_server: preserve payment ID when editing address book

selsta:


plowsof:
the docs mention payment id as an input for [add_address_book](https://docs.getmonero.org/rpc-library/wallet-rpc/#add_address_book) but im struggling to find it in the src for this end point. master branch is just address + description 

https://github.com/monero-project/monero/blob/master/src/wallet/wallet_rpc_server_commands_defs.h#L1928-L1940

and release does not support payment id as an input:

https://github.com/monero-project/monero/blob/release-v0.18/src/wallet/wallet_rpc_server_commands_defs.h#L1928-L1940

* the payment id is inferred when we add an integrated address * testing this now

plowsof (approved):

SNeedlewoods (approved):
I think adding a check for this in `tests/functional_tests/address_book.py` would not hurt.


___________________________________

As plowsof mentioned the docs are wrong, the following options should be removed AFAICT:
- for `add_address_book`
  - `payment_id`
- for `edit_address_book`
  - `set_payment_id`
  - `payment_id`

SNeedlewoods (approved):

Issue #10539, opened 2026-05-07
Reduce download overhead


An-anonymous-coder:
The `handle_target_data` function in `./common/download.cpp` locks the mutex for every incoming chunk. Unless there's a reason for it to be there that I'm unaware of, turning it into an atomic bool would probably reduce a lot of overhead.

Downloads are painfully slow. I stopped coding in C++, so I can't do a thorough audit, but considering the file hasn't been updated in a year, maybe it could use an update.

selsta:
Could you specify which downloads you are talking about? `download.cpp` is for downloading and updating monerod when a new version is available, a niche feature.

Did you mean blockchain download?

An-anonymous-coder:
> Could you specify which downloads you are talking about? `download.cpp` is for downloading and updating monerod when a new version is available, a niche feature.
> 
> Did you mean blockchain download?

I was referring to the blockchain download, yes. I wasn't sure where that code was located.

This issue is still relevant, but apparently not for the blockchain.

selsta:
Closing as not planned as the diagnosis in the issue was incorrect.

PR #10593, opened 2026-05-13
common/download.cpp: abort oversized transfers before written to disk

plowsof:
commit title is wrong, it would allow that to happen but this is just returning file size earlier, could use some help that

callers can reject oversized downloads

copied from here but total 0 , as soon as we have headers https://github.com/monero-project/monero/blob/fb5b06bd3bf84bb44a3998b6a6def5ce63752903/src/common/download.cpp#L155

although a niche feature : https://github.com/monero-project/monero/issues/10539#issuecomment-4408292045

for dns blocklists that download a file - i would like to implement a safety check against a malicious domain hosting a very large file - without this, we must write the entire thing to disk first.

### DIFF
--- a/src/common/download.cpp
+++ b/src/common/download.cpp
@@ -108,6 +108,8 @@ namespace tools
           {
             MINFO("Content-Length: " << length);
             content_length = length;
+            if (control->progress_cb && !control->progress_cb(control->path, control->uri, 0, content_length))
+              return false;
             boost::filesystem::path path(control->path);
             try
             {


plowsof:
`Content-length` is voluntarily given by the server (can be omitted, or a false value). i didn't think about this. so this was useless for my 'security' feature - we must write bytes to disk , and monitor the total of what is being written and check if that is > that our desired amount. which is either (in my case) an arbitrary number OR remaining disk space on the drive being written to, but this would be a ~~nice to have~~ (its already there https://github.com/monero-project/monero/commit/3813a992e4536af1aa9eb966337b3e247dac6812).

This would only be useful for a happy path - to show someone a progress bar while downloading a monerod update from getmonero, so on that basis i must nack this for now without said progress bar for the cli that can make use of this :D

PR #10595, opened 2026-05-13
p2p: download blocklist from quorum'd hash;url

plowsof:
sorry for refactor. here's my case: when switching the dns banlist to parsing/applying a banlist file - it'll be convenient / less duplication but a bug here in the parser could cause a node to crash or ips not actually banned, so its best to have it in one place. 

where dns would call this (to explain `good` being returned) https://github.com/monero-project/monero/blob/master/src/p2p/net_node.inl#L2127-L2156

### DIFF
--- a/src/common/download.cpp
+++ b/src/common/download.cpp
@@ -108,6 +108,8 @@ namespace tools
           {
             MINFO("Content-Length: " << length);
             content_length = length;
+            if (control->progress_cb && !control->progress_cb(control->path, control->uri, 0, content_length))
+              return false;
             boost::filesystem::path path(control->path);
             try
             {
@@ -266,6 +268,24 @@ namespace tools
     download_async_handle handle = download_async(path, url, [&success](const std::string&, const std::string&, bool result) {success = result;}, cb);
     download_wait(handle);
     return success;
+  }
+
+  bool download(const std::string &path, const std::string &url, size_t max_size)
+  {
+    const auto size_guard = [max_size](const std::string& /*path*/, const std::string& /*uri*/, size_t bytes_written, ssize_t content_length) {
+      if (max_size > 0 && content_length >= 0 && static_cast<size_t>(content_length) > max_size)
+      {
+        MERROR("Download rejected, file exceeds maximum size");
+        return false;
+      }
+      if (max_size > 0 && bytes_written > max_size)
+      {
+        MERROR("Download rejected, file exceeds maximum size");
+        return false;
+      }
+      return true;
+    };
+    return download(path, url, size_guard);
   }
 
   download_async_handle download_async(const std::string &path, const std::string &url, std::function<void(const std::string&, const std::string&, bool)> result, std::function<bool(const std::string&, const std::string&, size_t, ssize_t)> progress)
--- a/src/common/download.h
+++ b/src/common/download.h
@@ -38,6 +38,7 @@ namespace tools
   typedef std::shared_ptr<download_thread_control> download_async_handle;
 
   bool download(const std::string &path, const std::string &url, std::function<bool(const std::string&, const std::string&, size_t, ssize_t)> progress = NULL);
+  bool download(const std::string &path, const std::string &url, size_t max_size);
   download_async_handle download_async(const std::string &path, const std::string &url, std::function<void(const std::string&, const std::string&, bool)> result, std::function<bool(const std::string&, const std::string&, size_t, ssize_t)> progress = NULL);
   bool download_error(const download_async_handle &h);
   bool download_finished(const download_async_handle &h);
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -504,8 +504,17 @@ namespace nodetool
 
     if (!command_line::is_arg_defaulted(vm, arg_ban_list))
     {
+      const std::string ban_list_path = command_line::get_arg(vm, arg_ban_list);
       std::string error;
-      apply_blocklist_file(command_line::get_arg(vm, arg_ban_list), std::numeric_limits<time_t>::max(), false, &error);
+      try
+      {
+        apply_blocklist_file(ban_list_path, std::numeric_limits<time_t>::max(), false, &error);
+      }
+      catch (const std::exception& e)
+      {
+        error = "Failed to read ban list file " + ban_list_path;
+        MWARNING(error << ": " << e.what());
+      }
       if (!error.empty())
         throw std::runtime_error(error);
     }
@@ -2136,7 +2145,10 @@ namespace nodetool
     // TXT format <url>;<sha256>
     std::vector<std::string> records;
     if (!tools::dns_utils::load_txt_records_from_dns(records, dns_urls))
+    {
+      MWARNING("DNS blocklist: no TXT record found");
       return true;
+    }
 
     if (records.empty())
     {
@@ -2175,10 +2187,15 @@ namespace nodetool
     };
 
     crypto::hash cached_hash;
-    const bool cache_ok =
-      epee::file_io_utils::is_file_exist(cache_path.string()) &&
-      tools::sha256sum(cache_path.string(), cached_hash) &&
-      expected_hash == epee::string_tools::pod_to_hex(cached_hash);
+    bool cache_ok = false;
+    try
+    {
+      cache_ok = tools::sha256sum(cache_path.string(), cached_hash) && expected_hash == epee::string_tools::pod_to_hex(cached_hash);
+    }
+    catch (const std::exception& e)
+    {
+      MWARNING("DNS blocklist: failed to read cached file: " << e.what());
+    }
 
     bool rejected_oversize = false;
     const auto size_guard = [&rejected_oversize](const std::string& /*path*/, const std::string& /*uri*/, size_t bytes_written, ssize_t /*content_length*/) {
@@ -2213,9 +2230,18 @@ namespace nodetool
       }
 
       crypto::hash file_hash;
-      if (!tools::sha256sum(tmp_path, file_hash))
+      try
       {
-        MWARNING("DNS blocklist: failed to hash downloaded file from " << url);
+        if (!tools::sha256sum(tmp_path, file_hash))
+        {
+          MWARNING("DNS blocklist: failed to hash downloaded file from " << url);
+          cleanup_tmp(tmp_path);
+          return false;
+        }
+      }
+      catch (const std::exception &e)
+      {
+        MWARNING("DNS blocklist: failed to hash downloaded file from " << url << ": " << e.what());
         cleanup_tmp(tmp_path);
         return false;
       }
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -921,7 +921,7 @@ namespace nodetool
         ++good;
         continue;
       }
-      MERROR("Invalid IP address or IPv4 subnet: " << line);
+      MERROR("Invalid IP address or IPv4 subnet: " << line << " - " << parsed_addr.error());
     }
 
     return good;
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -506,9 +506,17 @@ namespace nodetool
     {
       const std::string ban_list_path = command_line::get_arg(vm, arg_ban_list);
       std::string error;
+      boost::system::error_code ec{};
+      if (!boost::filesystem::is_regular_file(boost::filesystem::path{ban_list_path}, ec))
+      {
+        error = "Failed to read ban list file " + ban_list_path;
+        MWARNING(error);
+      }
+
       try
       {
-        apply_blocklist_file(ban_list_path, std::numeric_limits<time_t>::max(), false, &error);
+        if (error.empty())
+          apply_blocklist_file(ban_list_path, std::numeric_limits<time_t>::max(), false, &error);
       }
       catch (const std::exception& e)
       {
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -871,7 +871,7 @@ namespace nodetool
   }
   //-----------------------------------------------------------------------------------
   template<class t_payload_net_handler>
-  size_t node_server<t_payload_net_handler>::apply_blocklist_file(const std::string& path, time_t seconds, bool add_only, std::string *error, size_t max_size)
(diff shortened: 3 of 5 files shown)


iamamyth:
Such refactors are better done in the PR which necessitates the refactor (i.e. the PR which changes the DNS blocklist logic), otherwise there's no telling whether the refactor makes any sense (in this case, for example, I can see multiple plausible refactoring paths, and the choice depends on the final form of the DNS blocklist support).

plowsof:
added a 2nd commit on top of the refactor ready for review :pray: some quick comments regarding possible functional tests. providing there is a 'disable dnssec' flag added to monerod: `DNS_PUBLIC` (without a custom port) requires sudo (to bind to port 53) - monerod can talk to a DNS server set up by python + logfiles are grepped for specific error message sequence. i have tried to hit all paths during testing

src/p2p/net_node.inl
```diff
@@ -898,6 +861,67 @@ namespace nodetool
   }
   //-----------------------------------------------------------------------------------
   template<class t_payload_net_handler>
+  size_t node_server<t_payload_net_handler>::apply_blocklist_file(const std::string& path, time_t seconds, bool add_only, std::string *error, size_t max_size)
+  {
+    const boost::filesystem::path blocklist_path(path);
+    if (!boost::filesystem::exists(blocklist_path))
```
iamamyth: This is somewhat wasteful: Opening the file will just fail if it doesn't exist when the time comes (the open call should have an appropriate error code), so all you get here is an extra syscall and data race

src/p2p/net_node.inl
```diff
@@ -898,6 +861,67 @@ namespace nodetool
...
+      if (error)
+      {
+        *error = "Can't find ban list file " + path;
+        MWARNING(*error);
+      }
+      return 0;
+    }
+
+    std::string banned_ips;
+    if (!epee::file_io_utils::load_file_to_string(path, banned_ips, max_size))
```
iamamyth: I don't particularly like the IO pattern: Where possible, operating on streams tends to clarify intent, perform a lot better, and greatly simplify testing. Here, for example, the function accepts a path, then reads the file into a string, only to wrap the string in a stream; having the function accept a stream cuts out clutter (you can, of course, offer a path-based version which delegates the heavy lifting to the stream implementation).

src/p2p/net_node.inl
```diff
@@ -2111,47 +2135,130 @@ namespace nodetool
...
-    for (const auto& record : records)
+    if (records.empty())
+    {
+      MWARNING("DNS blocklist: no TXT record found");
+      return true;
+    }
+
+    std::vector<std::string> fields;
+    std::string record = records[0];
+    boost::split(fields, record, boost::is_any_of(";"));
```
iamamyth: URLs can contain `;`, so I'd suggest flipping the order and placing the checksum first (you could, instead, pick an alternate separator, like space, but that might lend itself to typos).

src/p2p/net_node.inl
```diff
@@ -2111,47 +2135,130 @@ namespace nodetool
...
+
+    crypto::hash cached_hash;
+    const bool cache_ok =
+      epee::file_io_utils::is_file_exist(cache_path.string()) &&
+      tools::sha256sum(cache_path.string(), cached_hash) &&
+      expected_hash == epee::string_tools::pod_to_hex(cached_hash);
+
+    bool rejected_oversize = false;
+    const auto size_guard = [&rejected_oversize](const std::string& /*path*/, const std::string& /*uri*/, size_t bytes_written, ssize_t /*content_length*/) {
+      if (bytes_written > max_blocklist_size)
```
iamamyth: If you were to stream the file here, you could relax the size constraint (e.g. 10MB shouldn't be a big deal). Also, though, I think this filesize paranoia misses a key point from the earlier discussion: If you require all hosts to offer *the same file*, and apply a DNS-based quorum rule, it's pretty hard to cheat, as the DNS record can include the filesize, and someone has to compromise a lot of DNS records. A file size failsafe seems OK, but very much a last resort (sadly, HTTP won't necessarily tell you the size in advance, so forcing the size to be available via DNS helps).

iamamyth: Minor addendum: If you want to let alternate hosts use different banlists, you can still include the sizes in the DNS record, which helps with fail-early behavior (really, the HTTP server should be setting content-length for the whole file, but you can't guarantee it).

src/p2p/net_node.inl
plowsof: Thanks for the feedback. i've admittedly re-used old code without understanding what it was actually doing. streams are also new to me. the function now jumps to creating the stream using the path (without any redundant exists checks) - this then introduced me to stack trace errors which led me on quite the tangent. 

the max file size is only checked while bytes are being written - i have not implemented a stream there which iiuc would allow to hash the file in chunks while its downloading to reduce another IO call which i feel is acceptable, that would require some other edits so that we receive chunks as they arrive. 

preserving existing behaviour so that when launching monerod with `--ban-list` and an error happens it ~~crashes.~~ exits*

i've moved to `hash;url` and bumped the max size (while downloading) to 10MB (no checks performed on cached files or user supplied files during launch with `--ban-list` - ive opted to not include the size in the quorum as this would increase the testing surface and definitely make me battle stack traces / crashes that skip logic again

src/p2p/net_node.inl
plowsof: reg content-length - yes, i was originally wanting to fail early trusting that value - until i realised it is voluntarily provided / can be anything the server chooses and we can only rely on bytes being written

*Gaining access to the CDN and serving a malicious file falls outside of the quorum protection, thats the only reason i want a maximum - seems the easiest abuse method 

for my file size paranoia i think a KISS of 10MB max for bytes to be written is an acceptable compromise

iamamyth: Regarding content-length, it's not really arbitrary, in the sense that, for a non-chunked response, the client should hang up after receiving `content-length` bytes (i.e. in a decent client implementation, it's binding because the client won't continue past that point). But, you might get a chunked response, and then there's no length provided.

plowsof: wouldn't it make sense to handle all of this in download.cpp? - to have a size guard in there so the caller can just download and pass an optional max size? then if content-length is provided , fail early, else ensure bytes_written are less than the max

iamamyth: Probably, yea, as it's all just generic logic for a download client.

plowsof: moved size_guard to download.cpp with an overload that accepts a max size param, if content length is provided and greater than max it will fail early , if not, the size_guard is used.

src/p2p/net_node.inl
```diff
@@ -2111,47 +2141,128 @@ namespace nodetool
...
+      return true;
+    }
+
+    const std::string& record = records[0];
+    std::string expected_hash = record.substr(0, first_delim_pos);
+    std::string url = record.substr(first_delim_pos + 1, std::string::npos); // url may contain ';'
+    boost::trim(expected_hash);
+    boost::trim(url);
+
+    if (expected_hash.size() != 64 || url.empty())
```
selsta, line 2174: also check if url begins with `https://` and check if the hash has the correctly allowed characters

src/p2p/net_node.inl
```diff
@@ -502,48 +504,27 @@ namespace nodetool
...
-      const boost::filesystem::path ban_list_path(ban_list);
-      boost::system::error_code ec;
-      if (!boost::filesystem::exists(ban_list_path, ec))
+      const std::string ban_list_path = command_line::get_arg(vm, arg_ban_list);
+      std::string error;
+      boost::system::error_code ec{};
+      if (!boost::filesystem::is_regular_file(boost::filesystem::path{ban_list_path}, ec))
       {
-        throw std::runtime_error("Can't find ban list file " + ban_list + " - " + ec.message());
+        error = "Failed to read ban list file " + ban_list_path;
```
selsta, line 512: The new error is worse, it should differentiate between can't find and not a regular file.

src/p2p/net_node.inl
```diff
@@ -898,6 +879,55 @@ namespace nodetool
...
+    {
+      if (error)
+      {
+        *error = "Failed to read ban list file " + path;
+        MWARNING(*error);
+      }
+      return 0;
+    }
+
+    unsigned good = 0;
```
selsta, line 895:
```suggestion
    size_t good = 0;
```

src/p2p/net_node.inl
```diff
@@ -898,6 +879,55 @@ namespace nodetool
   }
   //-----------------------------------------------------------------------------------
   template<class t_payload_net_handler>
+  size_t node_server<t_payload_net_handler>::apply_blocklist_file(const std::string& path, time_t seconds, bool add_only, std::string *error)
+  {
+    std::ifstream iss{path};
```
selsta, line 884: misleading var name since this is not a `istringstream`

src/common/download.h
```diff
@@ -38,6 +38,7 @@ namespace tools
   typedef std::shared_ptr<download_thread_control> download_async_handle;
 
   bool download(const std::string &path, const std::string &url, std::function<bool(const std::string&, const std::string&, size_t, ssize_t)> progress = NULL);
+  bool download(const std::string &path, const std::string &url, size_t max_size);
```
selsta, line 41: separate function name might be better, otherwise NULL and 0 is a bit ambigious

src/common/download.cpp
```diff
@@ -268,6 +270,24 @@ namespace tools
...
 
+  bool download(const std::string &path, const std::string &url, size_t max_size)
+  {
+    const auto size_guard = [max_size](const std::string& /*path*/, const std::string& /*uri*/, size_t bytes_written, ssize_t content_length) {
+      if (max_size > 0 && content_length >= 0 && static_cast<size_t>(content_length) > max_size)
+      {
+        MERROR("Download rejected, file exceeds maximum size");
+        return false;
+      }
+      if (max_size > 0 && bytes_written > max_size)
```
iamamyth, line 281: Would just roll this into the prior `if`.

src/p2p/net_node.inl
```diff
@@ -502,48 +504,27 @@ namespace nodetool
...
     {
-      const std::string ban_list = command_line::get_arg(vm, arg_ban_list);
-
-      const boost::filesystem::path ban_list_path(ban_list);
-      boost::system::error_code ec;
-      if (!boost::filesystem::exists(ban_list_path, ec))
+      const std::string ban_list_path = command_line::get_arg(vm, arg_ban_list);
+      std::string error;
+      boost::system::error_code ec{};
+      if (!boost::filesystem::is_regular_file(boost::filesystem::path{ban_list_path}, ec))
```
iamamyth, line 510: `apply_blocklist_file` already does this work and returns an error string, just have it return an `ssize_t` or a `bool` with `size_t` as an out param.

src/p2p/net_node.inl
```diff
@@ -2111,47 +2141,128 @@ namespace nodetool
...
+    if (first_delim_pos == std::string::npos)
+    {
+      MWARNING("DNS blocklist: no valid TXT record found");
+      return true;
+    }
+
+    const std::string& record = records[0];
+    std::string expected_hash = record.substr(0, first_delim_pos);
+    std::string url = record.substr(first_delim_pos + 1, std::string::npos); // url may contain ';'
+    boost::trim(expected_hash);
```
iamamyth, line 2171: Forgiving extra spaces needlessly complicates the format/protocol.

src/p2p/net_node.inl
```diff
@@ -2111,47 +2141,128 @@ namespace nodetool
...
+      MWARNING("DNS blocklist: no valid TXT record found");
+      return true;
+    }
+
+    const boost::filesystem::path cache_path = boost::filesystem::path{m_config_folder} / "dns_blocklist.txt";
+    std::string error;
+    const auto apply_cached_blocklist = [&]() -> bool {
+      const size_t good = apply_blocklist_file(cache_path.string(), DNS_BLOCKLIST_LIFETIME, true, &error);
+      if (good > 0)
+        MINFO(good << " addresses added to the blocklist");
```
iamamyth, line 2185: This message is a little misleading, `good` is the size of the blocklist, some of the addresses in it may have already been blocked (suggest just `"DNS blocklist: " << good << " addresses in blocklist"` to match the format of the other messages)

iamamyth: Could just `return true` here and `false` by default, but I'd invert the condition to check for an empty blocklist, as that's the unusual case, i.e.:
```
if (good <= 0)
  return false;
MINFO(...);
return true;
```

src/p2p/net_node.inl
```diff
@@ -2111,47 +2141,128 @@ namespace nodetool
...
+        boost::filesystem::remove(tmp_path);
+      }
+      catch (const std::exception &e)
+      {
+        MWARNING("DNS blocklist: failed to remove tmp file: " << e.what());
+      }
+    };
+
+    const auto refresh_cache = [&]() -> bool {
+      const std::string tmp_path = cache_path.string() + ".tmp";
```
iamamyth, line 2212: The tmp file handling here seems like it's begging for RAII (remove on destroy if valid, swap operation which renders invalid).

src/p2p/net_node.inl
plowsof: thanks, I will opt for a `ssize_t` - i wasn't aware of this -1 (error) or (number of results) - combined with me assuming that the ifstream would fail reading a dir, apparently not so i went down a rabbit hole of returning _the error_ rather than letting the apply_blocklist_file handle everything (including error printing)  - and the caller decides if we throw or not which is the case when launching monerod with arg `--ban-list`

src/p2p/net_node.inl
plowsof: iiuc RAII is similar to a 'try then finally', so in this case we dont have to manually clean up the tmp file. there is an example of a `unique_scope_guard` [here](https://github.com/monero-project/monero/blob/fa4de39ae3c8f91f7c31f24e0ea76e7ec12507da/src/cryptonote_protocol/cryptonote_protocol_handler.inl#L1495) , i've never used this before but i should be able to implement it, thanks

iamamyth: Yea, in this case the "resource" would be the tmp download file, it either gets made permanent (via `tools::replace_file`, below) or deleted via its destructor (i.e. the dtor deletes the file iff it hasn't been made permanent).

PR #10597, opened 2026-05-14
p2p: restore safe sync mode when target height drops

selsta:


src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -2873,6 +2873,8 @@ skip:
     {
       MINFO("Target height decreasing from " << previous_target << " to " << target);
       m_core.set_target_blockchain_height(target);
+      if (target <= m_core.get_current_blockchain_height())
```
SChernykh: The description says when it drops, but here it is `<=`. Shouldn't it be a strict `<` ?

selsta: Strict `<` would miss the `target == current_height` case, where we are synchronized again. I changed it to match the existing threshold for disabling safe sync: unsafe mode is only entered at `current_height + 5`, so it is restored once the target drops below that.

SChernykh (approved):

j-berman (approved):

PR #10603, opened 2026-05-14
daemonizer: avoid unsafe pidfile truncation

selsta:


src/daemonizer/posix_fork.cpp
```diff
@@ -86,20 +122,28 @@ void fork(const std::string & pidfile)
...
     }
   }
 
-  if (! pidofs.fail())
+  if (pid_fd >= 0)
   {
-    int pid = ::getpid();
-    pidofs << pid << std::endl;
-    pidofs.close();
+    const std::string pid = std::to_string(::getpid()) + "\n";
```
vtnerd, line 144: This could throw, leaving `pid_fd` still open. We most likely shutdown, so it's not terrible.

The easiest fix is to use `auto_scope_leave_caller` - that gets rid of all of these checks for close before quitting as closing is automatically handled everywhere.

vtnerd (changes requested):

src/daemonizer/posix_fork.cpp
selsta: updated

src/daemonizer/posix_fork.cpp
```diff
@@ -35,26 +41,64 @@ void fork(const std::string & pidfile)
...
+  epee::misc_utils::auto_scope_leave_caller pid_fd_guard =
+    epee::misc_utils::create_scope_leave_handler(close_pid_fd);
   if (! pidfile.empty ())
   {
-	int oldpid;
-    std::ifstream pidrifs;
-    pidrifs.open(pidfile, std::fstream::in);
-    if (! pidrifs.fail())
+    struct stat st;
+    if (lstat(pidfile.c_str(), &st) == 0)
```
vtnerd, line 58: Also, are we concerned about TOCTOU issues? We may need to:

```c++
int pid_fd = open(pidfile.c_str(), O_NOFOLLOW, 0644);
if (pid_fd < 0)
  die(...);

if (fstat(pid_fd, &st) != 0)
  die(...);

if (!S_ISREG(st.st_mode))
  die(...);

std::string buf;
buff.resize(std::numeric_limits<int>::digits10 + 2);

ssize_t buf_len = read(pid_fd, &buf[0], buf.size());
if (buf_len < 0)
  die(...);

errno = 0;
int oldpid = 0;
buf.resize(size_t(buf_len));
if (buf.size() && boost::conversion::try_lexical_convert(buf, oldpid) && (kill(oldpid, 0) == 0 || errno == EPERM))
  die(...);

truncate(pid_fd, 0);
// ... remainder is same
```

vtnerd (approved):
I think this looks fine.

plowsof:
pidfile with wallet-rpc / monerod works as normal (ignore that damn @pr-patch-bot approval)

plowsof (approved):

PR #10605, opened 2026-05-15
rpc: speedup block&tx RPC calls

jeffro256:
In RPC calls `get_blocks.bin` and `get_transactions`:
  1. Skip deserializing then re-serializing tx blobs
  2. ~~Skip reading full tx from mempool database when requesting pruned txs~~
  3. ~~Fix pruned transaction pool info byte limit calculation (more txs per response)~~
  4. ~~Allow skipping calculating prunable tx hashes for mempool txs when set by caller in `get_transactions`~~
  5. ~~Reduce response size when skipping prunable tx hashes in `get_transactions`~~
  6. Provide full unpruned transaction from mempool in `get_transactions` when `!req.prune`
  7. Change `on_get_transactions()` response sort time from $O(N^2)$ to $O(N)$
  8. Generally improve memory usage and lookups in `on_get_transactions()`

Implements https://github.com/seraphis-migration/monero/issues/293#issuecomment-4444340293
Depends #9473, #10784

### DIFF
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -247,6 +247,17 @@ namespace cryptonote
     get_transaction_prefix_hash(tx, tx_prefix_hash);
     return true;
   }
+  //------------------------------------------------------------------
+  size_t get_tx_version(const blobdata_ref tx_blob)
+  {
+    size_t version;
+    const char* begin = reinterpret_cast<const char*>(tx_blob.data());
+    const char* end = begin + tx_blob.size();
+    int read = tools::read_varint(begin, end, version);
+    if (read <= 0)
+      throw std::runtime_error("Internal error getting transaction version");
+    return version;
+  }
   //---------------------------------------------------------------
   bool is_v1_tx(const blobdata_ref& tx_blob)
   {
@@ -262,6 +273,126 @@ namespace cryptonote
   bool is_v1_tx(const blobdata& tx_blob)
   {
     return is_v1_tx(blobdata_ref{tx_blob.data(), tx_blob.size()});
+  }
+  //---------------------------------------------------------------
+  bool get_transaction_unprunable_summary(const blobdata_ref tx_blob, unprunable_summary_t &summary_out)
+  {
+    //! @TODO: update for FCMP++:
+    //!    * Allow rct::RctTypeFcmpPlusPlus
+    //!    * Set output length for txout_to_carrot_v1
+
+    const unsigned char *p = reinterpret_cast<const unsigned char*>(tx_blob.data());
+    const unsigned char *end = reinterpret_cast<const unsigned char*>(tx_blob.data() + tx_blob.size());
+
+    #define READ_VARINT(v) if (tools::read_varint(p, end, v) <= 0) return false
+    #define READ_BYTE(v) if (end <= p ) { return false; } else { v = *p; ++p; }
+    #define SKIP(n) if (end - p < static_cast<std::ptrdiff_t>(n)) { return false; } else { p += (n); }
+
+    // read and validate tx version
+    READ_VARINT(summary_out.version);
+    if (summary_out.version < 1 || summary_out.version > 2)
+      return false;
+
+    // skip unlock_time
+    std::size_t dummy;
+    READ_VARINT(dummy);
+
+    // read number of inputs
+    READ_VARINT(summary_out.n_inputs);
+
+    // skip n_inputs inputs, checking for coinbase inputs
+    summary_out.is_coinbase = false;
+    for (std::size_t i = 0; i < summary_out.n_inputs; ++i)
+    {
+      std::size_t input_tag = 0;
+      READ_BYTE(input_tag);
+      switch (input_tag)
+      {
+      case 0xff: //txin_gen
+        summary_out.is_coinbase = true;
+        READ_VARINT(dummy); //height
+        break;
+      case 0x02: //txin_to_key
+        READ_VARINT(dummy); //amount
+        READ_VARINT(input_tag); //key_offsets.size()
+        for (;input_tag-->0;)
+          READ_VARINT(dummy);
+        SKIP(32); //k_image
+        break;
+      default:
+        return false;
+      }
+    }
+
+    // read number of outputs
+    READ_VARINT(summary_out.n_outputs);
+
+    // skip n_outputs outputs
+    for (std::size_t i = 0; i < summary_out.n_outputs; ++i)
+    {
+      READ_VARINT(dummy); //amount
+      std::size_t output_tag = 0;
+      READ_BYTE(output_tag); //target variant tag
+      std::ptrdiff_t output_length = 0;
+      switch (output_tag)
+      {
+      case 0x02: //txout_to_key
+        output_length = 32;
+        break;
+      case 0x03: //txout_to_tagged_key
+        output_length = 32 + 1;
+        break;
+      default:
+        return false;
+      }
+      SKIP(output_length);
+    }
+
+    // read extra length and skip that
+    READ_VARINT(summary_out.extra_len);
+    if (summary_out.extra_len > CRYPTONOTE_MAX_TX_SIZE)
+      return false;
+    SKIP(summary_out.extra_len);
+
+    // now `p` is at end of tx prefix...
+    summary_out.prefix_size = reinterpret_cast<const char*>(p) - tx_blob.data();
+
+    if (2 == summary_out.version)
+    {
+      // read RingCT type
+      std::uint8_t rct_type = std::numeric_limits<std::uint8_t>::max();
+      READ_BYTE(rct_type);
+      if (rct_type > rct::RCTTypeBulletproofPlus)
+        return false;
+
+      // skip unprunable RingCT fields
+      if (rct_type != rct::RCTTypeNull)
+      {
+        // skip txnFee
+        READ_VARINT(dummy);
+
+        // if RCTTypeSimple, skip pseudoOutputs
+        if (rct_type == rct::RCTTypeSimple)
+        {
+          SKIP(32 * summary_out.n_inputs);
+        }
+
+        // skip ECDH info and amount commitments
+        const bool short_amount = rct_type >= rct::RCTTypeBulletproof2;
+        const std::ptrdiff_t ecdh_tuple_len = short_amount ? 8 : 64;
+        const std::ptrdiff_t output_stuff_len = summary_out.n_outputs * (ecdh_tuple_len + 32);
+        SKIP(output_stuff_len);
+      }
+    }
+
+    #undef READ_VARINT
+    #undef READ_BYTE
+    #undef SKIP
+
+    // now `p` is at end of unprunable part of v2 tx...
+    summary_out.unprunable_size = reinterpret_cast<const char*>(p) - tx_blob.data();
+
+    return true;
   }
   //---------------------------------------------------------------
   bool generate_key_image_helper(const account_keys& ack, const std::unordered_map<crypto::public_key, subaddress_index>& subaddresses, const crypto::public_key& out_key, const crypto::public_key& tx_public_key, const std::vector<crypto::public_key>& additional_tx_public_keys, size_t real_output_index, keypair& in_ephemeral, crypto::key_image& ki, hw::device &hwdev)
@@ -499,6 +630,28 @@ namespace cryptonote
     CHECK_AND_ASSERT_THROW_MES(tx.is_blob_size_valid(), "BUG: blob size valid not set");
 
     return tx.blob_size;
+  }
+  //---------------------------------------------------------------
+  bool prune_transaction_blob(cryptonote::blobdata &tx_blob, const std::size_t known_unprunable_size)
+  {
+    std::size_t unprunable_size;
+    if (known_unprunable_size)
+    {
+      unprunable_size = known_unprunable_size;
+    }
+    else
+    {
+      // Otherwise, unprunable size is unknown. Deserialize unprunable part (w/o allocs) and retrieve size
+      unprunable_summary_t desc;
+      if (!get_transaction_unprunable_summary(tx_blob, desc))
+        return false;
+      unprunable_size = desc.unprunable_size;
+    }
+
+    CHECK_AND_ASSERT_MES(unprunable_size <= tx_blob.size(), false, "Unprunable size is larger than tx blob");
+    tx_blob.resize(unprunable_size);
+
+    return true;
   }
   //---------------------------------------------------------------
   bool get_tx_fee(const transaction& tx, uint64_t & fee)
@@ -1259,6 +1412,18 @@ namespace cryptonote
     return get_transaction_hash(t, res, NULL);
   }
   //---------------------------------------------------------------
+  bool calculate_transaction_prunable_hash(const std::size_t tx_version, const blobdata_ref prunable_tx_blob, crypto::hash &res)
+  {
+    switch (tx_version)
+    {
+    case 2:
+      cryptonote::get_blob_hash(prunable_tx_blob, res);
+      return true;
+    default:
+      return false;
+    }
+  }
+  //---------------------------------------------------------------
   bool calculate_transaction_prunable_hash(const transaction& t, const cryptonote::blobdata_ref *blob, crypto::hash& res)
   {
     if (t.version == 1)
@@ -1267,7 +1432,8 @@ namespace cryptonote
     if (blob && unprunable_size)
     {
       CHECK_AND_ASSERT_MES(unprunable_size <= blob->size(), false, "Inconsistent transaction unprunable and blob sizes");
-      cryptonote::get_blob_hash(blobdata_ref(blob->data() + unprunable_size, blob->size() - unprunable_size), res);
+      return calculate_transaction_prunable_hash(t.version,
+        blobdata_ref(blob->data() + unprunable_size, blob->size() - unprunable_size), res);
     }
     else
     {
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -283,6 +283,17 @@ namespace cryptonote
     get_transaction_prefix_hash(tx, tx_prefix_hash);
     return true;
   }
+  //------------------------------------------------------------------
+  size_t get_tx_version(const blobdata_ref tx_blob)
+  {
+    size_t version;
+    const char* begin = reinterpret_cast<const char*>(tx_blob.data());
+    const char* end = begin + tx_blob.size();
+    int read = tools::read_varint(begin, end, version);
+    if (read <= 0)
+      throw std::runtime_error("Internal error getting transaction version");
+    return version;
+  }
   //---------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


jeffro256:
Further possible improvement: Skip parsing block + miner tx and calculating miner transaction hash inside `BlockchainLMDB::get_blocks_from()`. This may require a DB migration, though

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -1983,17 +1986,10 @@ bool BlockchainLMDB::get_txpool_tx_blob(const crypto::hash& txid, cryptonote::bl
   if (result != 0)
       throw1(DB_ERROR(lmdb_error("Error finding txpool tx blob: ", result).c_str()));
 
-  bd.assign(reinterpret_cast<const char*>(v.mv_data), v.mv_size);
+  const std::size_t tx_read_size = (pruned && unprunable_size) ? std::min(unprunable_size, v.mv_size) : v.mv_size;
```
selsta: Is it intentional that when `tx_category == relay_category::all`, `unprunable_size` remains 0, so `tx_read_size` falls back to `v.mv_size` even if `pruned == true`?

jeffro256: This is okay in the sense that it will return correct values in the RPC, but yeah it isn't optimal. Can fix now

src/cryptonote_core/tx_pool.cpp
```diff
@@ -625,22 +626,19 @@ namespace cryptonote
...
-        MERROR("Failed to parse tx from txpool");
+
+      // Fetch tx blob and prune if applicable
+      td.blob_size = m_blockchain.get_txpool_tx_blob(txid, td.tx_blob, relay_category::all, pruned);
+      if (!td.blob_size)
         return false;
-      }
-      else
+      td.unprunable_size = meta.unprunable_size;
+      if (pruned)
```
j-berman: Note to self: what if `meta.pruned`?

jeffro256:
Rebased

j-berman:
I tested this PR and observed a significant speedup (~8s to read ~11.5k txs before this PR, ~2-2.5s to read 20-25k txs after this PR, so an ~8x improvement).

I was surprised it was still taking ~2s though. I would expect a way more drastic speedup because unpruned txs are well over ~8x larger than pruned. And ~2s could still be an issue if there are a lot of wallets connected to a node (plus, it's 2s :p).

So I ran some tests with LMDB in my local:

Test 1: set up a table with 20k records, each record 100kb, then read the first 1kb of every record into memory by copying into a buf.

Test 2: set up a table with 20k records, each record 1kb, then read the whole record into memory by copying into a buf.

Test 1 took ~4s, Test 2 took ~0.1s.

Here's my preliminary understanding of why this is being observed (may be wrong): apparently the OS can pull in much more than the assumed 4kb for each 100kb record (looks like it goes off [this readahead](https://kernel-internals.org/io/readahead/) which most Linux systems default to 128KB). LMDB offers a `MDB_NORDAHEAD` env variable on the db to avoid this (that doesn't work on Windows, and also, we'd need to open a new db to benefit from it anyway). When I use the `MDB_NORDAHEAD` on Test 1, Test 1 takes ~2.4s. The rest of the perf diff may be explained because LMDB uses "overflow pages" to store large records, which alters how LMDB reads the record compared to smaller records. Both of these seem to combine to impact the perf of this approach more than expected.

I think the above should prove we're likely to see an even more significant improvement here from using 2 distinct tables, to the point that this reading should end up close to neglible.

Here's the code for the tests I ran:

```c
// This file modified an mtest, so retains the OpenLDAP license:

// The OpenLDAP Public License
//   Version 2.8, 17 August 2003

// Redistribution and use of this software and associated documentation
// ("Software"), with or without modification, are permitted provided
// that the following conditions are met:

// 1. Redistributions in source form must retain copyright statements
//    and notices,

// 2. Redistributions in binary form must reproduce applicable copyright
//    statements and notices, this list of conditions, and the following
//    disclaimer in the documentation and/or other materials provided
//    with the distribution, and

// 3. Redistributions must contain a verbatim copy of this document.

// The OpenLDAP Foundation may revise this license from time to time.
// Each revision is distinguished by a version number.  You may use
// this Software under terms of this license revision or under the
// terms of any subsequent revision of the license.

// THIS SOFTWARE IS PROVIDED BY THE OPENLDAP FOUNDATION AND ITS
// CONTRIBUTORS ``AS IS'' AND ANY EXPRESSED OR IMPLIED WARRANTIES,
// INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
// AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT
// SHALL THE OPENLDAP FOUNDATION, ITS CONTRIBUTORS, OR THE AUTHOR(S)
// OR OWNER(S) OF THE SOFTWARE BE LIABLE FOR ANY DIRECT, INDIRECT,
// INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
// BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
// ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
// POSSIBILITY OF SUCH DAMAGE.

// The names of the authors and copyright holders must not be used in
// advertising or otherwise to promote the sale, use or other dealing
// in this Software without specific, written prior permission.  Title
// to copyright in this Software shall at all times remain with copyright
// holders.

// OpenLDAP is a registered trademark of the OpenLDAP Foundation.

// Copyright 1999-2003 The OpenLDAP Foundation, Redwood City,
// California, USA.  All Rights Reserved.  Permission to copy and
// distribute verbatim copies of this document is granted.

#define _GNU_SOURCE		/* posix_fadvise */
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>
#include <fcntl.h>
#include <unistd.h>
#include "lmdb.h"

#define E(expr) CHECK((rc = (expr)) == MDB_SUCCESS, #expr)
#define CHECK(test, msg) ((test) ? (void)0 : ((void)fprintf(stderr, \
	"%s:%d: %s: %s\n", __FILE__, __LINE__, msg, mdb_strerror(rc)), abort()))

#define NRECS      20000	/* number of records to insert */
#define SLICE_SIZE 1024		/* bytes read/copied per record */
#define BATCH      10		/* records per write transaction */
#define DBPATH     "./testdb"

/* Fill buf with a deterministic, key-dependent pattern so reads can verify:
 * the whole value is (k & 0xff) bytes, with the key encoded in the first 4. */
static void fill_value(char *buf, unsigned int k, size_t val_size)
{
	memset(buf, (int)(k & 0xff), val_size);
	memcpy(buf, &k, sizeof(k));
}

// Run with --sliced to test reading the first 1kb of n 100kb records
// Run without to test reading 1kb n records
int main(int argc, char *argv[])
{
	int rc = 0, fd, sliced = 0, a;
	unsigned int i, j, k;
	MDB_env *env;
	MDB_dbi dbi;
	MDB_val key, data;
	MDB_txn *txn;
	MDB_stat mst;
	char *valbuf, *rbuf;
	unsigned long count;
	size_t val_size, read_size, mapsize;
	struct timespec t0, t1;
	double secs;

	for (a = 1; a < argc; a++)
		if (!strcmp(argv[a], "--sliced"))
			sliced = 1;

	/* --sliced: big 100 KiB records (overflow), read a 1 KiB slice.
	 * default:  small 1 KiB records (inline), read the whole record. */
	val_size  = sliced ? (size_t)100 * 1024 : (size_t)SLICE_SIZE;
	read_size = sliced ? (size_t)SLICE_SIZE : val_size;
	mapsize   = sliced ? (size_t)3 * 1024 * 1024 * 1024
			   : (size_t)256 * 1024 * 1024;

	valbuf = malloc(val_size);
	rbuf = malloc(read_size);
	CHECK(valbuf != NULL && rbuf != NULL, "malloc buffers");

	srand(time(NULL));

	/* ---- write: insert NRECS values, committing every BATCH ---- */
	E(mdb_env_create(&env));
	E(mdb_env_set_mapsize(env, mapsize));
	E(mdb_env_set_maxdbs(env, 4));
	E(mdb_env_open(env, DBPATH, MDB_NOSYNC, 0664));

	E(mdb_txn_begin(env, NULL, 0, &txn));
	E(mdb_dbi_open(txn, "bigtbl", MDB_CREATE|MDB_INTEGERKEY, &dbi));
	E(mdb_txn_commit(txn));

	printf("%s: inserting %d records x %zu bytes...\n",
		sliced ? "--sliced" : "default", NRECS, val_size);
	key.mv_size = sizeof(i);
	key.mv_data = &i;
	data.mv_size = val_size;
	data.mv_data = valbuf;
	for (i = 0; i < NRECS; ) {
		E(mdb_txn_begin(env, NULL, 0, &txn));
		for (j = 0; j < BATCH && i < NRECS; j++, i++) {
			fill_value(valbuf, i, val_size);
			E(mdb_put(txn, dbi, &key, &data, 0));
		}
		E(mdb_txn_commit(txn));
	}

	/* Flush to disk and close the env, so nothing keeps the data mapped, then
	 * drop the file's pages from the OS page cache. posix_fadvise(DONTNEED)
	 * only evicts *clean*, unmapped pages, hence the sync + close first. The
	 * reads below then fault in from disk instead of RAM. */
	E(mdb_env_sync(env, 1));
	mdb_env_close(env);

	fd = open(DBPATH "/data.mdb", O_RDONLY);
	CHECK(fd >= 0, "open data.mdb");
	rc = posix_fadvise(fd, 0, 0, POSIX_FADV_DONTNEED);
	if (rc)
		fprintf(stderr, "warning: posix_fadvise: %s "
			"(reads may still be cached)\n", strerror(rc));
	close(fd);

	/* ---- read: reopen read-only and fetch every record from disk ---- */
	printf("Fetching %zu bytes of each record by key, in key order (cold cache)...\n",
		read_size);

	E(mdb_env_create(&env));
	E(mdb_env_set_mapsize(env, mapsize));
	E(mdb_env_set_maxdbs(env, 4));
	E(mdb_env_open(env, DBPATH, MDB_RDONLY, 0664)); // this is where to add MDB_NORDAHEAD

	E(mdb_txn_begin(env, NULL, MDB_RDONLY, &txn));
	E(mdb_dbi_open(txn, "bigtbl", MDB_INTEGERKEY, &dbi));
	key.mv_size = sizeof(k);
	key.mv_data = &k;
	count = 0;
	clock_gettime(CLOCK_MONOTONIC, &t0);
	for (k = 0; k < NRECS; k++) {
		size_t off;
		unsigned char expected;
		E(mdb_get(txn, dbi, &key, &data));
		CHECK(data.mv_size == val_size, "unexpected value size");
		memcpy(rbuf, data.mv_data, read_size);	/* copy read_size bytes out of the map */
		/* spot-check a random byte of the copy against fill_value()'s pattern */
		off = rand() % read_size;
		expected = (off < sizeof(k)) ? (unsigned char)(k >> (off * 8))
					     : (unsigned char)k;
		CHECK((unsigned char)rbuf[off] == expected, "copied value mismatch");
		count++;
	}
	clock_gettime(CLOCK_MONOTONIC, &t1);
	E(mdb_stat(txn, dbi, &mst));		/* read stat in the same txn */
	mdb_txn_abort(txn);

	secs = (t1.tv_sec - t0.tv_sec) + (t1.tv_nsec - t0.tv_nsec) / 1e9;
	printf("OK: all %lu records fetched by key and verified\n", count);
	printf("read time: %.3f s for %lu records (%.1f MiB/s copied, %.2f us/record)\n",
		secs, count,
		(double)count * read_size / (1024.0*1024.0) / secs,
		secs * 1e6 / count);
	printf("bigtbl: %llu entries, %llu leaf pages, %llu overflow pages\n",
		(unsigned long long)mst.ms_entries,
		(unsigned long long)mst.ms_leaf_pages,
		(unsigned long long)mst.ms_overflow_pages);

	mdb_env_close(env);
	free(valbuf);
	free(rbuf);
	return 0;
}
```

jeffro256:
So would you recommend that I peel of the pruned reading optimizations (bullet point 2) from this PR? I think that's one of the bigger parts of the diff, but it may be the least effective if what you're saying is true.

j-berman:
I think it would be nice to introduce the 2-table approach together with all of the other nice changes this PR brings, and nix the pruned reading optimizations.

It looks like using 2 tables could end up materially changing some (but not all) of the core touch-points of this PR, and it would be nice to get this all done in one go. I also think a migration for the 2 tables should be ok even before the FCMP++/Carrot `monerod` is released. This migration should execute fast since the pool these days is usually not so big.

j-berman:
Also worth noting: @selsta identified an issue with #9473's fundamental design choice to lean on byte size of the payload for the pool response, and is proposing an alternative method of handling the response in #10784 as well.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -511,6 +639,28 @@ namespace cryptonote
...
+  bool prune_transaction_blob(cryptonote::blobdata &tx_blob, const std::size_t known_unprunable_size)
+  {
+    std::size_t unprunable_size;
+    if (known_unprunable_size)
+    {
+      unprunable_size = known_unprunable_size;
+    }
+    else
+    {
+      // Otherwise, unprunable size is unknown. Deserialize unprunable part (w/o allocs) and retreive size
```
jpk68: Nit: small typo

src/cryptonote_core/tx_pool.h
```diff
@@ -457,9 +457,9 @@ namespace cryptonote
      */
     struct tx_details
     {
-      transaction tx;  //!< the transaction
       cryptonote::blobdata tx_blob; //!< the transaction's binary blob
       size_t blob_size;  //!< the transaction's size
+      size_t unprunable_size; //!< the transactions unprunable size
```
jpk68: Nit: this should likely be `transaction's`

jeffro256:
Rebased to resolve merge conflicts, implemented typo fixes @jpk68, undid pruned mempool tx reading changes according to @j-berman's feedback, and boosted the tx response merge algorithm in `core_rpc_server::on_get_transactions()` from $O(N^2)$ to $O(N)$!

jeffro256:
After @selsta's second review, I decided to drop points 4 and 5 since they would cause issues with existing wallets if used, they were kind of half-baked features, and didn't contribute much to the actual speedups.

jeffro256:
Rebased PR on top of #10784 which allows me to simplify away point 3

src/rpc/core_rpc_server.cpp
```diff
@@ -939,61 +910,92 @@ namespace cryptonote
...
+          }
+        }
+        txs = std::move(sorted_txs);
+        LOG_PRINT_L2("Found " << pool_txs_processed << "/" << vh.size() << " transactions in the pool");
+
+        // now erase missed_txs which are present in the mempool response
+        size_t erase_head = 0;
+        for (size_t erase_tail = 0; erase_tail < missed_txs.size(); ++erase_tail)
+        {
+          const bool erase = per_tx_pool_tx_details.count(missed_txs[erase_tail]);
```
jeffro256, line 796: A bitmap would probably be faster here and would add <1% memory overhead. Worth it?

src/rpc/core_rpc_server.cpp
```diff
@@ -939,61 +910,92 @@ namespace cryptonote
...
+            // calculate unprunable hash if applicable
+            crypto::hash prunable_hash = crypto::null_hash;
+            {
+              const std::size_t tx_version = get_tx_version(pruned_tx_blob);
+              if (1 != tx_version && !calculate_transaction_prunable_hash(tx_version, prunable_tx_blob, prunable_hash))
+              {
+                res.status = "Failed to calculate transaction prunable hash";
+                return true;
+              }
+            }
```
jeffro256, line 772: A good follow-up PR to this would be to cache the prunable hash in the transaction metadata, so it doesn't need to be calculated every single time.

selsta:
Please rebase.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -264,6 +275,126 @@ namespace cryptonote
...
+    #define READ_BYTE(v) if (end <= p ) { return false; } else { v = *p; ++p; }
+    #define SKIP(n) if (end - p < static_cast<std::ptrdiff_t>(n)) { return false; } else { p += (n); }
+
+    // read and validate tx version
+    READ_VARINT(summary_out.version);
+    if (summary_out.version < 1 || summary_out.version > 2)
+      return false;
+
+    // skip unlock_time
+    std::size_t dummy;
```
selsta: `dummy` needs to be `uint64_t` for the 64 bit fields below. On 32 bit builds using `size_t` makes `read_varint()` reject valid values above `UINT32_MAX`.

jeffro256: Done in latest

jeffro256:
Rebased and implemented @selsta's feedback

thomasbuilds (approved):

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -283,6 +283,17 @@ namespace cryptonote
...
   }
+  //------------------------------------------------------------------
+  size_t get_tx_version(const blobdata_ref tx_blob)
+  {
+    size_t version;
+    const char* begin = reinterpret_cast<const char*>(tx_blob.data());
+    const char* end = begin + tx_blob.size();
+    int read = tools::read_varint(begin, end, version);
+    if (read <= 0)
+      throw std::runtime_error("Internal error getting transaction version");
```
jpk68, line 294: How come this is handled using an exception?

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -535,6 +666,28 @@ namespace cryptonote
     return tx.blob_size;
   }
   //---------------------------------------------------------------
+  bool prune_transaction_blob(cryptonote::blobdata &tx_blob, const std::size_t known_unprunable_size)
```
jpk68: It looks like both call sites pass in `0` for `known_unprunable_size`, so can it be removed?

jpk68:
Nit: the commit message seems a bit broken; there are blank lines, and selsta's email address doesn't have a closing angle bracket symbol (`>`).

src/cryptonote_basic/cryptonote_format_utils.cpp
jeffro256: It *could* return some invalid tx version value, then callers check the return value each time, but this is less error prone for callers who assume the tx is deserializable.

src/cryptonote_basic/cryptonote_format_utils.cpp
jeffro256: Yup

jeffro256:
Pushed with @jpk68's review

selsta (approved):

jpk68 (approved):

PR #10606, opened 2026-05-15
db_lmdb: fix uninit read in get_tx_amount_output_indices

thomasbuilds:
`mdb_cursor_get` leaves `*data` untouched on `MDB_NOTFOUND` so the existing warn-and-continue reads uninitialized (or stale) `v.mv_data` / `v.mv_size` and returns garbage indices to callers including `remove_output()` during a reorg.

A missing `tx_outputs` entry cannot happen on a consistent DB so throw `DB_ERROR` like every other "Unexpected" state in this file.

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -3520,14 +3520,17 @@ std::vector<std::vector<uint64_t>> BlockchainLMDB::get_tx_amount_output_indices(
   while (n_txes-- > 0)
   {
     int result = mdb_cursor_get(m_cur_tx_outputs, &k_tx_id, &v, op);
+    op = MDB_NEXT;
     if (result == MDB_NOTFOUND)
+    {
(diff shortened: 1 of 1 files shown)


vtnerd (commented):
I haven't looked at this too deeply, but this is purely defensive against a possible invalid DB state ... ? As in, the comment even implies the `NOTFOUND` should never happen?

thomasbuilds:
You're right that there's no legitimate path to NOTFOUND here so I've switched it to throw DB_ERROR like the other "Unexpected" states in this file rather than warn-and-continue. The point of the PR is what the current code does after the warning: mdb_cursor_get leaves *data untouched on NOTFOUND so it reads uninitialized v on the first iteration or the previous tx's record on later ones (the n_txes>1 batch used by /getblocks.bin) and on the reorg path that garbage reaches remove_output(). Rebased on master.

jpk68 (approved):

jpk68 (approved):

selsta (approved):

PR #10613, opened 2026-05-16
src: use compiler defined win32 macro

SNeedlewoods:
submitted by jpk68
https://codeberg.org/jpk68/monero/pulls/17

(I have not reviewed this code, merely forwarding it)

### DIFF
--- a/src/common/password.cpp
+++ b/src/common/password.cpp
@@ -109,7 +109,7 @@ namespace
     return r;
   }
 
-#else // end WIN32 
+#else // if not _WIN32
 
   bool is_cin_tty() noexcept
   {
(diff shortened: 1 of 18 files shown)


src/common/password.cpp
```diff
@@ -109,7 +109,7 @@ namespace
     return r;
   }
 
-#else // end WIN32 
+#else // end _WIN32
```
selsta:
```
#else // end _WIN32
#endif // end _WIN32
```

These comments are misleading as is. I'd adjust or remove them.

selsta:
something got messed up, commit title seems wrong

selsta (approved):

vtnerd:
Where is `WIN32` defined? Can that be removed?

tobtoht:
I'm rejecting this patch for now because I noticed this changes behavior. I want to run unit tests on Windows in CI first and see if this breaks/fixes anything ( #10650).

Related: #8547

selsta:
Agree with @tobtoht, we first have to merge #10650 before this can be reviewed.

PR #10618, opened 2026-05-19
blockchain: handle changing LTM window size

jeffro256:
This part of the cache handing code assumed that A) the window size never changes, B) the window size is greater than or equal to the queried count, and C) that the cache container has an internal max size that is the size of the window. This PR fixes the behavior of `Blockchain::get_long_term_block_weight_median()` if none of those assumptions hold. This allows the long-term block median window size to be dynamic correctly. The handling of the LTM cache caused a consensus bug in the FCMP++ Beta Stressnet network, which shrunk the LTM window size at the v17 fork, but didn't resize the container.

Style note: the term "capacity" in the code may be suboptimal, especially if `epee::misc_utils::rolling_median_t<>` ever supports having a maximum size of the rolling values that differs from its allocated capacity. 

Relevant issue: https://github.com/seraphis-migration/monero/issues/390

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -1461,9 +1461,10 @@ uint64_t Blockchain::get_long_term_block_weight_median(uint64_t start_height, si
   PERF_TIMER(get_long_term_block_weights);
 
   CHECK_AND_ASSERT_THROW_MES(count > 0, "count == 0");
+  const uint64_t blockchain_height = m_db->height();
+  CHECK_AND_ASSERT_THROW_MES(start_height < blockchain_height, "start of block LTM range lies outside of chain");
 
   bool cached = false;
-  uint64_t blockchain_height = m_db->height();
   uint64_t tip_height = start_height + count - 1;
   crypto::hash tip_hash = crypto::null_hash;
   if (tip_height < blockchain_height)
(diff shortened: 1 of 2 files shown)


jeffro256:
Forced-pushed a compile fix for 32-bit systems.

src/cryptonote_core/blockchain.cpp
```diff
@@ -1478,33 +1478,49 @@ uint64_t Blockchain::get_long_term_block_weight_median(uint64_t start_height, si
...
+    m_long_term_block_weights_cache_rolling_median = epee::misc_utils::rolling_median_t<uint64_t>(target_cap);
+
   MTRACE("requesting " << count << " from " << start_height << ", uncached");
   std::vector<uint64_t> weights = m_db->get_long_term_block_weights(start_height, count);
   m_long_term_block_weights_cache_tip_hash = tip_hash;
-  m_long_term_block_weights_cache_rolling_median.clear();
   for (uint64_t w: weights)
     m_long_term_block_weights_cache_rolling_median.insert(w);
+
+  assert(static_cast<size_t>(m_long_term_block_weights_cache_rolling_median.size()) == count);
```
j-berman: If the chain is less than `count` blocks long, looks like this could trigger. Wouldn't affect mainnet (because the fork is well after that many blocks were in the chain), but possibly some tests.

j-berman (commented):
Personally I think this is fine to implement for beta stressnet only.

Yes if the window size changes in the future, we'd potentially need a change like this, but I think this is jumping the gun assuming a future window size change would also reuse this whole code path. Plus we're still default initializing the median window with `CRYPTONOTE_LONG_TERM_BLOCK_WEIGHT_WINDOW_SIZE`, which seems unnecessary with this PR.

I would personally prefer to keep this change set on beta where we're actually changing the window size in this specific way strictly for beta.

src/cryptonote_core/blockchain.cpp
jeffro256: You're right if the callers change their behavior such that they use count > number of blocks in chain. But for the moment, all callers of `get_long_term_block_weight_median()` cap `start_height + count` to the number of blocks in the chain. We could A) disallow `start_height + count > db->get_height()`, or B) change this to assert to compare against `std::min(count, weights.size())`

j-berman: Looks like this would break debug built `long_term_block_weight` unit tests e.g.

jeffro256: I'll test it out rn, but I don't think that it does since the `long_term_block_weight` test adds 5001 blocks to the DB, then calls `get_long_term_block_weight_median()` with `count=5000`.

jeffro256: Yeah, just tested unit tests `long_term_block_weight.*` in debug mode and they all pass

jeffro256: Just pushed a change which throws if `start_height >= blockchain_height` and asserts that the cache size is the same as the fetched weights size.

j-berman: Fair, thank you

PR #10622, opened 2026-05-20
contrib: fix `bucket_head2` alignment and serialization

jeffro256:
Fixes a couple issues:
1. Fix unaligned read 64-bit int from byte pointers in `levin_protocol_handler_async.h` (may crash on 32-bit ARM and RISC platforms)
2. Fixes strict aliasing violation by reading `bucket_head2` from a byte pointer 
3. Unpacks `bucket_head2`  for faster in-memory access

This one is harder to review than #10623, so if review speed ends up being a blocker, I'm fine moving forwards with just #10623

### DIFF
--- a/contrib/epee/include/net/levin_base.h
+++ b/contrib/epee/include/net/levin_base.h
@@ -43,23 +43,13 @@ namespace epee
 class byte_slice;
 namespace levin
 {
-#pragma pack(push)
-#pragma pack(1)
-	struct bucket_head
-	{
-		uint64_t m_signature;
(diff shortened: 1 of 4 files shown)


vtnerd (commented):
`boost::endian::big_uint64_t`, etc might be helpful here. It handles endianness and can be aligned or unaligned. This would dramatically shrink the changes needed as accessing the type automatically does the conversion via memcpy, etc

selsta:
Network tests failing.

jeffro256:
Staying with #10623

PR #10627, opened 2026-05-20
src: remove non-standard thread extensions

SNeedlewoods:
submitted by jpk68
https://codeberg.org/jpk68/monero/pulls/16

(I have removed new-lines, else did not review this code, merely forwarding it)

### DIFF
--- a/src/common/perf_timer.cpp
+++ b/src/common/perf_timer.cpp
@@ -92,9 +92,9 @@ el::Level performance_timer_log_level = el::Level::Info;
 
 #ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
   #warning "Building with fuzzing mode UNSAFE FOR PRODUCTION!"
-  __thread std::vector<LoggingPerformanceTimer*> *performance_timers = NULL;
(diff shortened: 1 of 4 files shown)


tobtoht:
Builds failing.

src/common/threadpool.h
```diff
@@ -62,7 +63,7 @@ class threadpool
     boost::condition_variable cv;
     threadpool &pool;
     int num;
-    bool error_flag;
+    std::atomic<bool> error_flag;
```
iamamyth: This requires some kind of justification, there's a good chance the mutex (or another synchronization construct) acts as a memory barrier, in which case the atomic only confuses the issue.

iamamyth: Looking more closely, it's pretty clear the caller should only query `error` after `wait`ing, in which case, as speculated, the lock/cond var do the job.

iamamyth (approved):
Build passes and this patch makes sense to me (threadpool was written prior to `thread_local` being supported in the min C++ version).

selsta (approved):

PR #10631, opened 2026-05-20
ci: gitian: run jobs when depends is changed [0.18]

tobtoht:
Package upgrades on the release branch can currently break gitian builds, this change prevents that.

This matches the behavior for guix on the master branch. It will also run on tags.

### DIFF
--- a/.github/workflows/gitian.yml
+++ b/.github/workflows/gitian.yml
@@ -2,8 +2,17 @@ name: ci/gh-actions/gitian
 
 on:
   push:
-    tags:
-      - '*'
+    paths:
+      - 'contrib/depends/**'
+      - 'contrib/gitian/**'
+      - '!contrib/**.md'
+      - '.github/workflows/gitian.yml'
+  pull_request:
+    paths:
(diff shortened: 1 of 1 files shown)


plowsof (approved):

.github/workflows/gitian.yml
```diff
@@ -2,8 +2,17 @@ name: ci/gh-actions/gitian
 
 on:
   push:
-    tags:
-      - '*'
```
plowsof, line 6: also run on tags, is this a typo?

tobtoht: No, CI is run by default on tags if `push` is present. See e.g. the workflow file for this tag on the seraphis-migration repo: https://github.com/seraphis-migration/monero/actions/runs/25146583702

tobtoht: See also: https://stackoverflow.com/a/77703560

tobtoht:
Sorry for the spam. This should now work as intended.

selsta (approved):

PR #10637, opened 2026-05-21
net: canonicalize Tor and I2P hostnames

selsta:


src/net/host.h
```diff
@@ -0,0 +1,20 @@
...
+    /*!
+    * \brief Canonicalizes an ASCII hostname for storage and comparison.
+    *
+    * \param host Hostname to canonicalize in-place.
+    */
+    inline void canonicalize_host(std::string& host) noexcept
+    {
+        for (char& c : host)
+        {
+            if ('A' <= c && c <= 'Z')
```
vtnerd: Small nitpick, the existing code uses the `u8` prefix to guarantee character set. We should likely do the same here - `u8'A'`, etc.

Although the scenarios where this matters is basically non-existent. So ...

vtnerd (approved):
I think this is fine, but I listed a nitpick in case someone else wants to comment on it.

src/net/host.h
selsta: Updated.

vtnerd (approved):

src/net/host.h
```diff
@@ -0,0 +1,20 @@
...
+#include <string>
+
+namespace net
+{
+    /*!
+    * \brief Canonicalizes an ASCII hostname for storage and comparison.
+    *
+    * \param host Hostname to canonicalize in-place.
+    */
+    inline void canonicalize_host(std::string& host) noexcept
```
iamamyth, line 12: Should add tests for this function (and relatedly, the doc `brief` adds no information; suggest, e.g., `Canonicalize a hostname by converting letters to lowercase`).

selsta: Updated.

iamamyth (approved):
Tests & doc look good, thanks.

selsta:
Had to do one more update because `check_host()` requires the domain to be lower case, so the tests failed. I moved up the `normalized_host` conversion.

vtnerd (approved):

Issue #10207, opened 2025-11-12
Popping large number of blocks is barely functional


nahuhh:
when blocks are popped automatically (during reorgs etc) is understandable that txs would be retained.

When manually popping blocks, txs should be dropped / deleted.

Problem: its extremely inefficient to move txs back to the pool, making large pops essentially impossible

This request is for manually popping blocks to simply delete them, along with their txs, instead of wasting _tons_ of time moving them around.

The majority (or all) of the time that you manually pop blocks, you are trying to roll the chain back to an earlier state for testing or because of an issue. I dont see why we'd retain the txs.

nahuhh:
Fixed in https://github.com/monero-project/monero/pull/10644

PR #10644, opened 2026-05-22
Blockchain: option for fast manual block popping

jeffro256:
If passing `keep_txs=false` in the RPC, or if using `pop_blocks` in the CLI with no additional flags, one can manually pop blocks much faster than before. This in done by skipping fetching/deserializing full transaction lists, and not adding them back into the mempool. This PR bumps the RPC minor version. One can retain previous behavior in the CLI by passing `keep-txs` to the `pop_blocks` command.

Resolves (or at least helps) #10207

### DIFF
--- a/src/blockchain_db/blockchain_db.cpp
+++ b/src/blockchain_db/blockchain_db.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2024, The Monero Project
+// Copyright (c) 2014-2026, The Monero Project
 // 
 // All rights reserved.
 // 
@@ -172,13 +172,6 @@ void BlockchainDB::init_options(boost::program_options::options_description& des
   command_line::add_arg(desc, arg_db_salvage);
 }
 
-void BlockchainDB::pop_block()
-{
-  block blk;
-  std::vector<transaction> txs;
-  pop_block(blk, txs);
-}
-
 void BlockchainDB::add_transaction(const crypto::hash& blk_hash, const transaction& tx, const epee::span<const std::uint8_t> blob, const crypto::hash* tx_hash_ptr, const crypto::hash* tx_prunable_hash_ptr)
 {
   bool miner_tx = false;
@@ -311,7 +304,7 @@ void BlockchainDB::set_hard_fork(HardFork* hf)
   m_hardfork = hf;
 }
 
-void BlockchainDB::pop_block(block& blk, std::vector<transaction>& txs)
+void BlockchainDB::pop_block(block& blk, std::vector<transaction>* txs)
 {
   blk = get_top_block();
 
@@ -319,10 +312,13 @@ void BlockchainDB::pop_block(block& blk, std::vector<transaction>& txs)
 
   for (const auto& h : boost::adaptors::reverse(blk.tx_hashes))
   {
-    cryptonote::transaction tx;
-    if (!get_tx(h, tx) && !get_pruned_tx(h, tx))
-      throw DB_ERROR("Failed to get pruned or unpruned transaction from the db");
-    txs.push_back(std::move(tx));
+    if (txs)
+    {
+      cryptonote::transaction tx;
+      if (!get_tx(h, tx) && !get_pruned_tx(h, tx))
+        throw DB_ERROR("Failed to get pruned or unpruned transaction from the db");
+      txs->push_back(std::move(tx));
+    }
     remove_transaction(h);
   }
   remove_transaction(get_transaction_hash(blk.miner_tx));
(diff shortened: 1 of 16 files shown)


jeffro256:
Popping 1000 blocks from the mainnet chain went from ~65s to ~19s on my laptop. Results would be more pronounced with heavier blocks.

jeffro256:
On the FCMP++ beta stressnet, with ~5.4MB blocks, this reduced the time for popping 20 blocks from ~15s to ~3s.

jeffro256:
One potential option for the `keep-txs` default value is set it to `keep_txs = n_blocks <= CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE`. Thus, for shallow reorgs where inputs of current txs are not invalidated, the txs are kept. But for situations where the txs would be invalid in all cirsumstances except where the txs are added back in the exact same order, then we default to fast popping.

selsta:
Tests need update.

nahuhh:
> One potential option for the `keep-txs` default value is set it to `keep_txs = n_blocks <= CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE`. Thus, for shallow reorgs where inputs of current txs are not invalidated, the txs are kept. But for situations where the txs would be invalid in all cirsumstances except where the txs are added back in the exact same order, then we default to fast popping. 

(iiuc) on rct, deep-ish reorgs dont guarantee invalidation, so thats would be one reason why i dont think it would make sense to fast-pop on reorgs. and although my request was for default fast-pop when popping manually, I think opt-in is acceptable

jeffro256:
I think for CLI, it makes a lot of sense to have no-keep-txs be the default, but I think that we should retain backwards-compatible RPC behavior. Updated in latest.

nahuhh (approved):
Tested CLI & RPC

~1 minute to pop 1 million blocks on testnet

src/daemon/command_server.cpp
```diff
@@ -299,8 +299,8 @@ t_command_server::t_command_server(
     m_command_lookup.set_handler(
       "pop_blocks"
     , std::bind(&t_command_parser_executor::pop_blocks, &m_parser, p::_1)
-    , "pop_blocks <nblocks>"
-    , "Remove blocks from end of blockchain"
+    , "pop_blocks <nblocks> [keep-txs] [no-keep-txs]"
```
selsta:
```suggestion
    , "pop_blocks <nblocks> [keep-txs|no-keep-txs]"
```

src/daemon/command_parser_executor.cpp
```diff
@@ -945,7 +950,22 @@ bool t_command_parser_executor::pop_blocks(const std::vector<std::string>& args)
...
+    {
+      const std::string keep_arg = args.at(1);
+      if (keep_arg == "keep-txs")
+        keep_txs = true;
+      else if (keep_arg == "no-keep-txs")
+        keep_txs = false;
+      else
+      {
+        std::cout << "Invalid syntax: Unrecognized argument: '" << keep_arg
+          << "'. For more details, use the help command.'" << std::endl;
```
selsta:
```suggestion
          << "'. For more details, use the help command." << std::endl;
```

src/cryptonote_core/blockchain.h
```diff
@@ -1377,9 +1378,11 @@ namespace cryptonote
     /**
      * @brief removes the most recent block from the blockchain
      *
+     * @param keep_txs whether to place popped non-coinbase transactions back into the mempool
+     *
      * @return the block removed
```
selsta, line 1383: With `keep_txs` false, it does not return the removed block but is instead unspecified.

src/blockchain_db/blockchain_db.cpp
```diff
@@ -174,9 +174,14 @@ void BlockchainDB::init_options(boost::program_options::options_description& des
 
 void BlockchainDB::pop_block()
 {
-  block blk;
-  std::vector<transaction> txs;
-  pop_block(blk, txs);
+  // DB modification behavior should match other overload of pop_block()
```
j-berman: I personally think we should try to reuse the code paths so we don't have to remember to update both spots if either change in the future.

I'd suggest modifying the other function to accept a 3rd `keep_txs` param instead

jeffro256:
Rebased, added @selsta's suggestions, merged `BlockchainDB` code paths for `pop_blocks()` like @j-berman suggested, and applied the fast block pop method to `Blockchain` initialization and the import utility.

src/cryptonote_core/blockchain.h
jeffro256: Now it always returns the block

src/blockchain_db/blockchain_db.cpp
jeffro256: Done. `BlockchainDB::pop_block()` is now one method which takes an *optional* pointer to a vector of transactions.

selsta (approved):

j-berman (approved):

src/blockchain_db/blockchain_db.cpp
j-berman: pointer solution looks good, thank you

PR #10646, opened 2026-05-23
wallet2: avoid linear scans in pool state updates

selsta:
Open for feedback if this makes sense outside of edge cases where a malicious daemon returns worst case pool.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -3089,6 +3089,7 @@ void read_pool_txs(const cryptonote::COMMAND_RPC_GET_TRANSACTIONS::request &req,
     MDEBUG("Reading pool txs");
     if (res.txs.size() == req.txs_hashes.size())
     {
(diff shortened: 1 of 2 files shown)


j-berman (approved):

src/wallet/wallet2.cpp
SChernykh: You missed this loop, which makes the whole function still slow.

SChernykh: You already have the `pool_set(res.tx_hashes.begin(), res.tx_hashes.end());` here, you could pass it to `remove_obsolete_pool_txs` (maybe add an optional parameter - a pointer to it), so that it doesn't have to create the same set again.

SChernykh (changes requested):

SChernykh:
LGTM, but found two more places where speedups were left on the plate.

SChernykh (approved):

Issue #10445, opened 2026-04-22
Flaky tests


tobtoht:
- [ ] select_outputs.density

```
 [ RUN      ] select_outputs.density
/__w/monero/monero/tests/unit_tests/output_selection.cpp:175: Failure
Expected: (fabsf(selected_ratio - chain_ratio)) < (0.025f), actual: 0.0313777 vs 0.025
```

- [ ] select_outputs.gamma

```
[ RUN      ] select_outputs.gamma
/__w/monero/monero/tests/unit_tests/output_selection.cpp:132: Failure
Expected: (median) >= (1.3 * 86400), actual: 111360 vs 112320
```

- [ ] select_outputs.exact_unlock_block

```
 [ RUN      ] select_outputs.exact_unlock_block
/__w/monero/monero/tests/unit_tests/output_selection.cpp:259: Failure
Value of: picked_exact_unlock_block
  Actual: false
Expected: true
```

- [ ] DNSResolver.IPv4Success

```
 [ RUN      ] DNSResolver.IPv4Success
2026-07-15 19:54:16.050	I Using public DNS server(s): 9.9.9.9 (TCP)
/__w/monero/monero/tests/unit_tests/dns_resolver.cpp:48: Failure
Expected: (1) <= (ips.size()), actual: 1 vs 0

[  FAILED  ] DNSResolver.IPv4Success (36 ms)
```

https://github.com/monero-project/monero/actions/runs/29440034127/job/87436314166#step:15:1063

- [ ] AddressFromURL.Failure

```
[ RUN      ] AddressFromURL.Failure
2026-07-21 18:14:22.635	W Invalid DNSSEC TXT record signature for example.veryinvalid: validation failure <example.veryinvalid. TXT IN>: no NSEC3 records from 76.76.10.0 for DS veryinvalid. while building chain of trust
2026-07-21 18:14:22.635	W Possibly your DNS service is problematic. You can have monerod use an alternate via env variable DNS_PUBLIC. Example: DNS_PUBLIC=tcp://9.9.9.9
/Users/runner/work/monero/monero/tests/unit_tests/address_from_url.cpp:115: Failure
Value of: dnssec_result
  Actual: false
Expected: true

[  FAILED  ] AddressFromURL.Failure (2206 ms)
```

https://github.com/monero-project/monero/actions/runs/29852463867/job/88708493175?pr=10935#step:7:73

- [ ] address_book

```
[TEST STARTED] address_book
Creating wallet
Testing address book
[...]

PR #10650, opened 2026-05-24
ci: run fast tests on windows and macos

tobtoht:
We currently do not run automated tests for these platforms. 

Running the full suite takes ~90 minutes, however we can run a subset which includes unit tests in under 2 minutes. This gives us decent coverage without blowing up CI run time.

I have excluded some known flaky tests, see #10445.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,9 @@ env:
     EOF
   CCACHE_SETTINGS: |
     ccache --max-size=150M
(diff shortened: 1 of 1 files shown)


selsta:
Could you rebase and force push now that we have Windows related tests fixed?

selsta (approved):

plowsof (approved):

jpk68 (approved):

PR #10651, opened 2026-05-24
add safety for future fork heights [release-v0.18]

SNeedlewoods:
#10647


selsta:
There won't be any forks on this branch so I don't think we need this for release branch.

SNeedlewoods:
So should I close, or is there any justification to still add this?
E.g. just for reducing the diff between `master` and `release`

selsta:
I'd say close it. Unless there's a bug where we need an emergency fork we wouldn't fork from `release-v0.18`. If that happens we can still apply it in the future.

PR #10659, opened 2026-05-26
Add cbor support to all ZMQ-PUB subscriptions

vtnerd:
### Marking as draft for now. This PR:

  * Adds cbor output (writer) to the `::wire` serialization system
  * Adds block+transaction output support to the `::wire` serialization system
  * Switches all `ZMQ-PUB `output to use `::wire` serialization system (instead of just select few)
  * Adds cbor output to all existing `ZMQ-PUB` subscription options

### Needs (before marking ready):
  * A system to signal which subscription types are supported

### DIFF
--- a/contrib/epee/include/serialization/wire/cbor.h
+++ b/contrib/epee/include/serialization/wire/cbor.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2025, The Monero Project
+// Copyright (c) 2026, The Monero Project
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without modification, are
@@ -27,44 +27,22 @@
 
 #pragma once
 
-#include <type_traits>
-#include <utility>
+#include "serialization/wire/cbor/base.h"
+#include "serialization/wire/cbor/error.h"
+#include "serialization/wire/cbor/write.h"
 
-#include "serialization/wire.h"
-
-namespace wire
-{
-  //! A wrapper that tells `wire::writer`s to write the inner type as an array.
-  template<typename U>
-  struct range_
-  {
-    WIRE_DEFINE_CONVERSIONS()
-
-    using value_type = unwrap_reference_t<U>;
-
-    U value;
-
-    constexpr const value_type& get_value() const noexcept { return value; }
-    value_type& get_value() noexcept { return value; }
-
-    // concept requirements for `is_optional_on_empty`    
-    bool empty() const { return get_value().empty(); }
-
-    // concept requirements for `optional_field`.
-    explicit operator bool() const noexcept { return !empty(); }
-    range_ operator*() const noexcept { return *this; }
-  };
-
-  template<typename F, typename T>
-  void write_bytes(F& format, const range_<T> self)
-  {
-    wire_write::array(format, self.get_value());
+#define WIRE_CBOR_DEFINE_OBJECT(type, map)                        \
+  void write_bytes(::wire::cbor_writer& dest, const type& source) \
+  {                                                               \
+    map(dest, source);                                            \
   }
 
-  //! Links `value` with `range_`.
-  template<typename T>
-  inline constexpr range_<T> range(T value)
-  {
-    return {std::move(value)};
-  }
-} // wire
+//! Define functions that convert `type` to/from cbor bytes
+#define WIRE_CBOR_DEFINE_CONVERSION(type)                                      \
+  std::error_code convert_to_cbor(epee::byte_stream& dest, const type& source) \
+  { return ::wire_write::to_bytes<::wire::cbor_slice_writer>(dest, source); }
+
+//! Define functions that convert `type::request` and `type::response` to/from cbor bytes
+#define WIRE_CBOR_DEFINE_COMMAND(type)          \
+  WIRE_CBOR_DEFINE_CONVERSION(type::request)    \
+  WIRE_CBOR_DEFINE_CONVERSION(type::response)
--- a/contrib/epee/include/serialization/wire/cbor/error.h
+++ b/contrib/epee/include/serialization/wire/cbor/error.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2025, The Monero Project
+// Copyright (c) 2026, The Monero Project
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without modification, are
@@ -27,44 +27,39 @@
 
 #pragma once
 
+#include <system_error>
 #include <type_traits>
-#include <utility>
-
-#include "serialization/wire.h"
 
 namespace wire
 {
-  //! A wrapper that tells `wire::writer`s to write the inner type as an array.
-  template<typename U>
-  struct range_
+namespace error
+{
+  //! Type wrapper to "grab" cbor errors
+  enum class cbor : int
   {
-    WIRE_DEFINE_CONVERSIONS()
-
-    using value_type = unwrap_reference_t<U>;
-
-    U value;
-
-    constexpr const value_type& get_value() const noexcept { return value; }
-    value_type& get_value() noexcept { return value; }
-
-    // concept requirements for `is_optional_on_empty`    
-    bool empty() const { return get_value().empty(); }
-
-    // concept requirements for `optional_field`.
-    explicit operator bool() const noexcept { return !empty(); }
-    range_ operator*() const noexcept { return *this; }
+    success = 0, // required for `expected<T>`
+    incomplete,
+    integer_encoding
   };
 
-  template<typename F, typename T>
-  void write_bytes(F& format, const range_<T> self)
-  {
-    wire_write::array(format, self.get_value());
-  }
+  //! \return Static string describing error `value`.
+  const char* get_string(cbor value) noexcept;
 
-  //! Links `value` with `range_`.
-  template<typename T>
-  inline constexpr range_<T> range(T value)
+  //! \return Category for cbor generated errors.
+  const std::error_category& cbor_category() noexcept;
+
+  //! \return Error code with `value` and `cbor_category()`.
+  inline std::error_code make_error_code(cbor value) noexcept
   {
-    return {std::move(value)};
+    return std::error_code{int(value), cbor_category()};
   }
-} // wire
+}
+}
+
+namespace std
+{
+  template<>
+  struct is_error_code_enum<wire::error::cbor>
+    : true_type
+  {};
+}
--- a/src/ringct/rctTypes.cpp
+++ b/src/ringct/rctTypes.cpp
@@ -32,6 +32,12 @@
 #include "cryptonote_config.h"
 #include "rctTypes.h"
 #include "int-util.h"
+#include "serialization/wire.h"
+#include "serialization/wire/adapted/span.h"
+#include "serialization/wire/adapted/vector.h"
+#include "serialization/wire/wrapper/range.h"
+#include "serialization/wire/wrapper/variant.h"
+
 using namespace crypto;
 using namespace std;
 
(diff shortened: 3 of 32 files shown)


contrib/epee/include/serialization/wire/msgpack/fwd.h
```diff
@@ -0,0 +1,46 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#define WIRE_MSGPACK_DECLARE_ENUM(type)            \
+  const char* get_string(type) noexcept;           \
+  void read_bytes(::wire::msgpack_reader&, type&); \
+  void write_bytes(:wire::msgpack_writer&, type)
```
selsta:
```suggestion
  void write_bytes(::wire::msgpack_writer&, type)
```

contrib/epee/include/serialization/wire/cbor.h
selsta, line 1: Is this file needed? It redefines `WIRE_MSGPACK_DEFINE_OBJECT` and also has different include paths compared to other files. It also mentiones `read.h` which does not exist yet.

contrib/epee/include/serialization/wire/msgpack/error.h
```diff
@@ -0,0 +1,61 @@
...
+
+#pragma once
+
+#include <system_error>
+
+namespace wire
+{
+namespace error
+{
+  //! Type wrapper to "grab" rapidjson errors
```
selsta:
```suggestion
  //! Type wrapper to "grab" msgpack errors
```

contrib/epee/include/serialization/wire/msgpack/error.h
```diff
@@ -0,0 +1,61 @@
...
+    underflow_tree
+  };
+
+  //! \return Static string describing error `value`.
+  const char* get_string(msgpack value) noexcept;
+
+  //! \return Category for msgpack generated errors.
+  const std::error_category& msgpack_category() noexcept;
+
+  //! \return Error code with `value` and `rapidjson_category()`.
```
selsta:
```suggestion
  //! \return Error code with `value` and `msgpack_category()`.
```

contrib/epee/src/wire/msgpack/write.cpp
```diff
@@ -0,0 +1,206 @@
...
+  void msgpack_writer::check_flush()
+  {
+    if (needs_flush_ && (max_buffer < bytes_.size() || bytes_.available() < flush_threshold))
+      flush();
+  }
+
+  void msgpack_writer::do_integer(const std::intmax_t value)
+  {
+    assert(value < 0); // constraint checked in header
+    if (0xe0 < value) // 0xe0 needs to be type `int` to work
```
selsta: Could you double check this?

contrib/epee/src/wire/msgpack/write.cpp
```diff
@@ -0,0 +1,206 @@
...
+    return out;
+  }
+
+  msgpack_writer::~msgpack_writer() noexcept
+  {}
+
+  void msgpack_writer::real(const double source)
+  {
+    write_tag(bytes_, msgpack::tag::float64);
+    bytes_.write(reinterpret_cast<const char*>(std::addressof(source)), sizeof(source));
```
selsta: This also needs to be big endian.

contrib/epee/include/serialization/wire/msgpack/write.h
```diff
@@ -0,0 +1,178 @@
...
+  {
+    epee::byte_stream bytes_;
+    std::size_t expected_; //! Tracks number of expected elements remaining
+    bool needs_flush_;
+
+    //! Provided data currently in `bytes_`.
+    virtual void do_flush(epee::span<const uint8_t>);
+
+    //! Flush written bytes to `do_flush(...)` if configured
+    void check_flush();
```
selsta: So far this is never called.

contrib/epee/src/wire/msgpack/write.cpp
vtnerd: I could make it `int(e0)` I guess. This should be the default afaik.

contrib/epee/src/wire/msgpack/write.cpp
vtnerd: This is screwed currently as we need boost 1.71 for easy conversion to big endian, but are stuck on a min of boost 1.66. Not sure the best way to handle this yet.

contrib/epee/include/serialization/wire/msgpack/write.h
vtnerd: Yup, just realized I never tested the stream writer (which needs flushing). This will be added to the test harness eventually.

contrib/epee/src/wire/msgpack/write.cpp
SChernykh: `int8_t(e0)`, because `int(e0)` will still keep it positive

src/ringct/rctTypes.cpp
```diff
@@ -339,4 +344,152 @@ namespace rct {
...
+    WIRE_DEFINE_OBJECT(clsag, clsag_map);
+    void write_bytes(wire::writer& dest, const prune_wrapper_<const rctSig> source)
+    {
+        boost::optional<rct::xmr_amount> fee;
+        boost::optional<prunable_write> prunable;
+
+        if (source.p.get().type != rct::RCTTypeNull)
+            fee.emplace(source.p.get().txnFee);
+
+        if (!source.prune && (!source.p.get().p.bulletproofs.empty() || !source.p.get().p.bulletproofs_plus.empty() || !source.p.get().p.rangeSigs.empty() || !source.p.get().get_pseudo_outs().empty()))
```
SChernykh: Forgot to check for `!p.MGs.empty()` here, this is also prunable.

src/ringct/rctTypes.cpp
```diff
@@ -339,4 +344,152 @@ namespace rct {
...
+        }
+
+        template<typename F, typename T>
+        void rct_sig_prunable_map(F& format, T& self)
+        {
+            // make all arrays required for ZMQ backwards compatability
+            wire::object(format,
+                wire::field("range_proofs", wire::range<false>(std::ref(self.p.rangeSigs))),
+                wire::field("bulletproofs", wire::range<false>(std::ref(self.p.bulletproofs))),
+                wire::field("bulletproof_plus", wire::range<false>(std::ref(self.p.bulletproofs_plus))),
```
SChernykh: Should be `"bulletproofs_plus"` (plural)

contrib/epee/include/serialization/wire/msgpack/base.h
```diff
@@ -0,0 +1,198 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <cstdint>
+#include <limits>
+#include <string>
+#include <limits>
```
SChernykh: `<limits>` included twice

src/ringct/rctTypes.cpp
```diff
@@ -339,4 +344,152 @@ namespace rct {
...
+                wire::field("encrypted", wire::range(std::ref(self.p.get().ecdhInfo))),
+                wire::field("commitments", wire::range(std::ref(self.p.get().outPk))),
+                wire::optional_field("fee", std::ref(fee)),
+                wire::optional_field("prunable", std::ref(prunable))
+            );
+        }
+    } // anonymous
+    void write_bytes(wire::writer& source, const ctkey& dest)
+    {
+        wire_write::bytes(source, dest.mask);
```
SChernykh: Why do you write to the `source` here? Or is it just a wrong variable name?

contrib/epee/src/wire/msgpack/write.cpp
tobtoht: We can bump the minimum Boost version all the way up to 1.74 without breaking builds for our oldest supported platforms.

src/ringct/rctTypes.cpp
vtnerd: Fixed.

src/ringct/rctTypes.cpp
vtnerd: Fixed

contrib/epee/include/serialization/wire/msgpack/base.h
vtnerd: Fixed

src/ringct/rctTypes.cpp
vtnerd: Copy and paste snafu. Fixed.

contrib/epee/src/wire/msgpack/write.cpp
vtnerd: I can't believe I said this so confidently, this is clearly a bug. This particular branch never triggers, thus small negative values are written as a full `int8_t` instead (one byte tag + 1 byte signed integer). If msgpack is used I will add a test case for this (I already [submitted a fix + test for LWS](https://github.com/vtnerd/monero-lws/pull/258)).

contrib/epee/include/serialization/wire/cbor.h
vtnerd: Yes this was accidentally copied.

contrib/epee/src/wire/msgpack/write.cpp
vtnerd: I lied slightly, looks like 1.72 is the minimum supported version for the easy boost fix - the documentation appears incorrect. I'll see if I can find a workaround.

Otherwise, what is dictating the 1.74? Our policy of supported distros past a certain threshold?

selsta: https://github.com/monero-project/monero/issues/9446

vtnerd: I realized we have a builtin `double` conversion function via `CONVERT_POD`, but it converts everything to little-endian whereas this needs big-endian.

I'm about 99% set on forcing a minimum Boost 1.72 - looks like I will get minimal criticism for this.

vtnerd: @tobtoht @selsta I [started a PR in LWS](https://github.com/vtnerd/monero-lws/pull/260). The read side will not be ported (yet), but this is what the write side will look like with the boost usage.

kayabaNerve:
FYI, my thoughts on non-JSON encodings in Monero have largely been "what the fuck", for obvious historical reasons. My primary argument is anything which gets me stop thinking "what the fuck".

EPEE is actually, as much as I hate to say it, a not horrible binary JSON approximate buried under zero documentation and a mess of specific behaviors. When I proposed removing the array format, #10138, it's because it was unused and half-implemented (and therefore non-functional). Removing it made EPEE less "what the fuck".

I likely support any standardized binary format in place of where EPEE is used today. I'm barely aware of CBOR and have never actually seen it in use, except in one bad project specifically, though I've seen support for it in a major lib/ecosystem I use. However, [RFC 8949](https://www.rfc-editor.org/rfc/rfc8949) standardizes CBOR. I honestly hate the idea of supporting data/time and other such 'extended' data types though (if a 'complete' CBOR implementation was done). Limiting to the generic model would be fine.

msgpack isn't standardized by the IETF but it has a well-defined specification. If we ban extensions, I'm fine with msgpack _at a high-level_, even if I have a slight preference to CBOR simply based on 'standards good' and CBOR being the one to have an IETF RFC.

I remember, when I last discussed this with you (vtnerd), I did briefly review the protocols themselves. I forget what notes I had on them other than what I repeat here. I do have thoughts on wire formats, and the possibility for non-allocating encoders, but I haven't redone that analysis now and am fine deferring it.

I will note I don't wish for monerod to offer and support JSON, EPEE, msgpack, and CBOR. I'd like to, after an intermediary phase, offer just two _and_ standardize the RPC routes between them. While people SHOULD use the binary interfaces, and public interfaces SHOULD NOT expose large interfaces (the output distribution) over JSON, the node shouldn't require you implement multiple standards to send Monero transactions at all (as right now, EPEE is needed for some routes and JSON for others).

I also have many other opinions on the RPC routes, I know @hinto-janai has commented on obscurities (binary JSON???), and I've noted we're missing core JSON-RPC 2.0 functionality (#10118). In my opinion, moving away from EPEE is a great step in that direction (even if solely re: ZMQ for now, when ZMQ is its own discussion again...).

vtnerd:
I'm leaning towards msgpack with string keys and no extensions, but this could change depending on your feedback. If cbor is chosen, `monerod` would guarantee that implementing the generic model would be sufficient. I don't see a reason to use extensions to define cryptographic types; sending raw binary for cryptographic primitives seems sufficient.

Implementing a msgpack parser is arguably simpler because:
  1. No indefinite length arrays/maps makes skipping map fields when parsing extremely easy (track depth with a single integer)
  2. No float16

Guaranteeing canonical cbor eliminates (1), but now the output side has to do complex float handling to reduce output size and map keys must be sorted. We rarely need to output floats so we could temporarily disable float output for canonical encodings. However, map key sorting is easiest to guarantee via integer map keys that can be validated at compile-time, and integer map keys adds a bit of friction to downstream users.

Except for the 2 points above, cbor (generic model) is extremely similar to msgpack, which gives a nod to cbor since its IETF standard.

> I will note I don't wish for monerod to offer and support JSON, EPEE, msgpack, and CBOR. I'd like to, after an intermediary phase, offer just two and standardize the RPC routes between them.

Supporting 3 binary formats (epee, msgpack, cbor) for _output_, is fairly simple given my interface. It's mostly boilerplate C++, with actual lines in use being in the 900-1500 range (for all 3 combined). But understandably, this is unnecessary maintenance for such a thing.

> (even if solely re: ZMQ for now, when ZMQ is its own discussion again...).

The goal is ZMQ-PUB first for binary format, then slowly introducing it to ZMQ-RPC. Updating HTTP-RPC is going to be a beast, but lots can be learned by getting ZMQ deployed first.

### Integer Keys

**Writing 2 input/2 output Bulletprof+ tx on AMD Ryzen 3900**
  |                                 | Size (bytes)  | Output (ns)        | Input (ns)            |
  |--------------------|--------------|----------------|------------------|
  | Json                         | 5,352             | 26,944             |  50,431                 |
  | msgpack (string)    | 2,836 (-47%) | 8,946 (-66%)  | 31,445 (-38%)    |
  | msgpack (integer)  | 2,550 (-52%) | 10,770 (-60%) | 28,043 (-44%)   |
  | cryptonote              | 2,149 (-59%) | 4,648 (-83%)  | 9,975 (-80%)      |

The numbers suggest integer keys isn't worth the effort. The hard to quantify metric is that integers can be validated to be sorted at compile-time, which gets us closer to canonical cbor format. Given that we have a builtin canonical format, `cryptonote` (used by blocks and transactions), it's not clear that attempting a second canonical format is worth the effort.

String keys also simplify any schema attempt.

vtnerd:
One other thing: CDDL! msgpack schema would have to be implemented in json-schema and limited to string keys, whereas CDDL would be used to specify json+cbor. CDDL is easier to read than JSON schema, so this may help downstream developers.

vtnerd:
Thanks for the feedback @kayabaNerve . Switching to CBOR with string keys.

The summary is that I wrote a CBOR encoder, and it is dead simple compared to msgpack (which is already very simple). The CBOR encoder should have better instruction cache utilization, depending on how the compiler inlines things. Here is some better numbers where I just encoded the same tx a 1,000x in a hot loop:

**Writing 2 input/2 output Bulletprof+ tx on AMD Ryzen 3900 1,000 times in loop normalized to once**
  |                                 | Size (bytes) | Output (ns)    |
  |--------------------|-------------|---------------|
  | msgpack (string)    | 2,834          | 1,118                |
  | msgpack (integer)  | 2,548          | 1,079 (-3.4%) |
  | cbor (string)            | 2,826          | 986 (-11.8%)  |
  | cbor (integer)          | 2,540          | 890 (-20.4%) |

This more or less confirms my suspicions. A CBOR decoder is going to be a little more complex, as previously stated, but since `monerod` is mostly tons of outbound data, it makes sense to optimize for encoder performance.

It should be noted, that this re-uses the buffer, such that allocations are basically never occurring. So this is truly measuring the hot path of each encoder.

kayabaNerve:
> I'm leaning towards msgpack with string keys and no extensions, but this could change depending on your feedback. If cbor is chosen,

> I do have thoughts on wire formats, and the possibility for non-allocating encoders, but I haven't redone that analysis now and am fine deferring it.

Please, don't worry about me. Right now, I'm fine saying I don't want to champion this and will trust you too, having posted my opinions on the overall design goals above :)

It's quite interesting to hear the encoder is so much faster. At some point, I'm sure I'll have to do the encoder/decoder, but for either CBOR or msgpack, there should be pre-existing options I'm fine with. I do like my work on non-allocating deserializers, and could complain if one is less friendly to non-allocating deserialization than the other, but I worked on non-allocating deserializers because EPEE historically had memory exhaustion issues. When there's a standard protocol, with clear documentation and experience behind it, that's less of a concern and I won't insist. I am fine with JSON after all :sweat_smile: 

Thank you for your work on this and consideration towards all its aspects!

tests/unit_tests/zmq_rpc.cpp
```diff
@@ -103,10 +107,132 @@ TEST(ZmqRestrictedMethods, BasicCoverage)
...
+        case wire::cbor::major::bytes:
+          if (bytes.size() < size)
+            throw std::logic_error{"cbor missing bytes"};
+          dest.binary({bytes.data(), size});
+          bytes.remove_prefix(size);
+          break;
+        case wire::cbor::major::string:
+          if (bytes.size() < size)
+            throw std::logic_error{"cbor missing string"};
+          if (!stack.empty() && stack.back().second == wire::cbor::major::map && stack.back().first % 2 == 0)
```
selsta, line 179:
```suggestion
          if (!stack.empty() && stack.back().second == wire::cbor::major::map && (stack.back().first - 1) % 2 == 0)
```

because `stack.push_back({size * 2 + 1, wire::cbor::major::map});` ?

vtnerd: No, this should work or the JSON would've been broken.

The start_object sets the value to `size * 2 + 1`, then the loop at the bottom does an immediate `--` on that same value. Therefore, each map key should be at an "even" value afaik.

selsta: Yes, I confirmed that the tests pass locally. This can be marked as resolved.

vtnerd:
Sorry @SChernykh your prior review is partially invalidated - I switched the PR to cbor. You'll notice that the encoder is slightly more compact due to cbor's design. I've passed the binary test values to an outside cbor decoder, and everything looked as expected.

This encoder never outputs indefinite length maps/objects/strings/binary, always outputs `float64` (no attempt to detect "compression to `float32` or `float16` is done), compresses signed/unsigned integers to smallest possible value, and does not sort the map keys.

I'm also still looking into how to gracefully notify clients that `cbor` is available. The `MONERO_VERSION` field is not sent in restricted mode, so some other mechanism is needed for that specific scenario. Otherwise, the version field would be sufficient.

src/ringct/rctTypes.cpp
```diff
@@ -339,4 +345,155 @@ namespace rct {
...
+    WIRE_DEFINE_OBJECT(clsag, clsag_map);
+    void write_bytes(wire::writer& dest, const prune_wrapper_<const rctSig> source)
+    {
+        boost::optional<rct::xmr_amount> fee;
+        boost::optional<prunable_write> prunable;
+
+        if (source.p.get().type != rct::RCTTypeNull)
+            fee.emplace(source.p.get().txnFee);
+
+        if (!source.prune && (!source.p.get().p.MGs.empty() || !source.p.get().p.bulletproofs.empty() || !source.p.get().p.bulletproofs_plus.empty() || !source.p.get().p.rangeSigs.empty() || !source.p.get().get_pseudo_outs().empty()))
```
vtnerd: I need to add `CLSAGs` to this list. Crap.

vtnerd:
Force pushed a mechanism to report the available ZMQ PUB/SUB filters over RPC. This indirectly leaks the major version of the release in a restricted RPC call, so perhaps this is a bad idea.

Removing the draft label.

contrib/epee/include/serialization/wire/cbor/error.h
```diff
@@ -0,0 +1,57 @@
...
+  const char* get_string(cbor value) noexcept;
+
+  //! \return Category for cbor generated errors.
+  const std::error_category& cbor_category() noexcept;
+
+  //! \return Error code with `value` and `cbor_category()`.
+  inline std::error_code make_error_code(cbor value) noexcept
+  {
+    return std::error_code{int(value), cbor_category()};
+  }
```
selsta, line 55: Should `std::is_error_code_enum<wire::error::cbor>` specialization be added to match `contrib/epee/include/serialization/wire/error.h`?

vtnerd: It certainly makes conversions more convenient. I'll add it shortly.

vtnerd:
Force pushed the change recommended by @selsta.

src/rpc/core_rpc_server.cpp
```diff
@@ -496,6 +497,7 @@ namespace cryptonote
     res.synchronized = check_core_ready();
     res.busy_syncing = m_p2p.get_payload_object().is_busy_syncing();
     res.restricted = restricted;
+    res.zmq_pub_filters = listener::zmq_pub::get_all_filters();
```
selsta, line 500: `rpc` now has an unresolved `zmq_pub` dependency that could break some `fuzz_rpc*` targets. Maybe add a `src/rpc/zmq_pub_filters.h`?

src/serialization/json_object.cpp
```diff
@@ -1497,6 +1498,7 @@ void fromJsonValue(const rapidjson::Value& val, cryptonote::rpc::DaemonInfo& inf
   GET_FROM_JSON_OBJECT(val, info.adjusted_time, adjusted_time);
   GET_FROM_JSON_OBJECT(val, info.start_time, start_time);
   GET_FROM_JSON_OBJECT(val, info.version, version);
+  GET_FROM_JSON_OBJECT(val, info.pub_filters, pub_filters);
```
selsta, line 1501: Should this field be optional when deserializing? Older daemons do not include `pub_filters`, so `GET_FROM_JSON_OBJECT` makes a new ZMQ client unable to parse their `get_info` responses.

src/rpc/core_rpc_server.cpp
vtnerd: The current solution is the cleanest as it pulls the names of the filters directly from the arrays used to publish the outgoing data. A header file would imply a copy-paste job that had to be manually synchronized.

src/serialization/json_object.cpp
vtnerd: Yes, it should be optional. It is unlikely to matter though, as very few people are using this code directly.

tests/unit_tests/zmq_rpc.cpp
```diff
@@ -295,7 +453,11 @@ namespace
     {
       cryptonote::account_base temp_account;
       temp_account.generate();
-      return make_transaction({temp_account.get_keys().m_account_address});
+
+      cryptonote::account_base temp_account2;
+      temp_account2.generate();
+ 
+      return make_transaction({temp_account.get_keys().m_account_address, temp_account2.get_keys().m_account_address});
```
thomasbuilds, line 460: This only builds `bp_version = 2` (BP + MLSAG) and the functional tests subscribe `json-minimal` only so `bulletproof_plus_map` / `clsag_map` never run in CI. A `bp_version = 4` case here would cover the mainnet path.

jpk68:
These changes LGTM, besides the issue detailed in the comment from @thomasbuilds above.

Also, it might be nice to have tag types labelled as something like "not yet implemented" rather than "not supported", as they could definitely be useful for other things using CBOR down the line (e.g. a generic wire protocol for hardware devices).

PR #10660, opened 2026-05-26
gitian: optional APT cacher for docker builds [0.18]

nahuhh:
Required changes when using some distros, like manjaro or debian 12

#8105 

> The requirement of apt-cacher-ng becomes optional and enabled only if the binary is present. This opens the possibility to run deterministic builds in a wider range of distributions.


selsta (approved):

nahuhh:
Squashed

selsta (approved):

tobtoht (approved):

PR #10666, opened 2026-05-27
crypto: split wire serialization in header

jeffro256:
This splits wire serialization off into its own header, to keep crypto.h minimal as possible. Also improves readability, and adds an overload for `crypto::ec_point`, which will be used for the FCMP++ integration. ZMQ/wire changes added in #10241

@vtnerd

### DIFF
--- a/contrib/epee/include/serialization/wire/traits.h
+++ b/contrib/epee/include/serialization/wire/traits.h
@@ -27,7 +27,7 @@
 
 #pragma once
 
-#include <cstdint>
+#include <cstddef>
 #include <type_traits>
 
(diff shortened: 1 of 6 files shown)


vtnerd (approved):

jeffro256:
I believe that the functional test failure is a sporadic failure (node didn't sync fast enough) unrelated to this change. Could not repro locally.

selsta:
@jeffro256 it's a recent master only change that introduced this sporadic failure on CI, I have it on my todo list to try to debug it by running it in a loop on CI with more logging

tobtoht (approved):

PR #10669, opened 2026-05-27
crypto: reduce crypto.h header weight

jeffro256:
Refactors `crypto.h` and its dependencies such that the preprocessed size of the header is reduced dramatically. On my toolchain (GCC 16.1.1), the preprocessed size of `crypto.h` is 3.71MB in the `master` branch. In this branch, it is 164KB, a 22.6x reduction in size. This should help build times, especially with the FCMP++/CARROT integration, which adds dozens of translation units, almost all of which pull in `crypto.h`. 

Depends on #10666

### DIFF
--- a/src/crypto/crypto.cpp
+++ b/src/crypto/crypto.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2024, The Monero Project
+// Copyright (c) 2014-2026, The Monero Project
 // 
 // All rights reserved.
 // 
@@ -39,6 +39,7 @@
 #include <boost/shared_ptr.hpp>
 
 #include "common/varint.h"
+#include "hex.h"
 #include "warnings.h"
 #include "crypto.h"
 #include "hash.h"
@@ -91,7 +92,7 @@ namespace crypto {
     return &reinterpret_cast<const unsigned char &>(scalar);
   }
 
-  boost::mutex &get_random_lock()
+  static boost::mutex &get_random_lock()
   {
     static boost::mutex random_lock;
     return random_lock;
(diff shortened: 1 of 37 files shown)


src/crypto/crypto.h
```diff
@@ -31,60 +31,56 @@
...
 #include "generic-ops.h"
-#include "hex.h"
-#include "span.h"
 #include "hash.h"
 
-namespace crypto {
+//forward declarations
+namespace boost
+{
+template <typename T> class optional;
```
vtnerd, line 45: There's `boost/optional/optional_fwd.hpp`. Is that too heavy-weight?

vtnerd (commented):
This looked decent. I suppose its within our rules to allow for cleanup like this (when its technically not strictly needed).

src/crypto/crypto.h
jeffro256: That would more than double the preprocessed size of `crypto.h`, so it's not preferable

selsta:
Needs rebase.

jeffro256:
Rebased.

jeffro256:
Rebased again.

selsta:
@jeffro256 needs another look, CI build fails

jeffro256:
Unit test failure unrelated,

src/crypto/crypto.cpp
```diff
@@ -109,6 +110,48 @@ namespace crypto {
     add_extra_entropy_not_thread_safe(ptr, bytes);
   }
 
+  template<typename T>
+  typename std::enable_if<std::is_integral<T>::value, T>::type rand_range(T range_min, T range_max) {
+    crypto::random_device rd;
+    std::uniform_int_distribution<T> dis(range_min, range_max);
+    return dis(rd);
+  }
+  #define INSTANTIATE_RAND_RANGE(t) template t rand_range<t>(t,t);
```
selsta, line 119: The header accepts every integral T, but the .cpp only instantiates five types. Intentional? It would fail to link for `int`, `short`, and `unsigned short`.

src/crypto/crypto.cpp
```diff
@@ -109,6 +110,48 @@ namespace crypto {
     add_extra_entropy_not_thread_safe(ptr, bytes);
   }
 
+  template<typename T>
+  typename std::enable_if<std::is_integral<T>::value, T>::type rand_range(T range_min, T range_max) {
+    crypto::random_device rd;
+    std::uniform_int_distribution<T> dis(range_min, range_max);
```
selsta, line 116: `#include <random>` missing

selsta:
Also needs rebase.

src/crypto/crypto.cpp
jeffro256: Yes, it was intentional because those are the used types in the codebase, so they aren't needed right now. One can put more instantiations when needed, but it also wouldn't hurt to put the other ones now. I don't care which one we do.

jeffro256:
Rebased and included `<random>`

jpk68 (approved):

selsta (approved):

jpk68 (approved):

PR #10670, opened 2026-05-27
simplewallet: add wallet-dir flag

### LINKED ISSUE #7674: [Feature suggestion] --wallet-dir flag
From IRC

```
22:11 <clavi> can I set a wallet directory for monero-wallet-cli?
22:11 <clavi> the config file only lets me set a single wallet
22:13 <clavi> I want to be able to run something like "monerocli <walletname>" from any directory
22:13 <clavi> without typing the full paths each time
22:17 <clavi> would be neat to have a "--wallet-dir" option in the CLI config file
```

SNeedlewoods:
submitted by jpk68
https://codeberg.org/jpk68/monero/pulls/30

Closes #7674

Made some suggestions and partially tested the code. If someone wants to do more testing for different combinations of flags that would be welcome.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -169,6 +169,7 @@ namespace
 {
   constexpr std::array<std::string_view, 5> allowed_priority_strings = tools::fee_priority_utilities::fee_priority_strings;
   const auto arg_wallet_file = wallet_args::arg_wallet_file();
+  const command_line::arg_descriptor<std::string> arg_wallet_dir = {"wallet-dir", sw::tr("Directory to look for and save wallet files. The path must be absolute and will get ignored if a wallet file is also specified by its absolute path."), ""};
   const command_line::arg_descriptor<std::string> arg_generate_new_wallet = {"generate-new-wallet", sw::tr("Generate new wallet and save it to <arg>"), ""};
   const command_line::arg_descriptor<std::string> arg_generate_from_device = {"generate-from-device", sw::tr("Generate new wallet from device and save it to <arg>"), ""};
   const command_line::arg_descriptor<std::string> arg_generate_from_view_key = {"generate-from-view-key", sw::tr("Generate incoming-only wallet from view key"), ""};
@@ -611,6 +612,17 @@ namespace
     }
 
     return true;
+  }
+  static std::string resolve_wallet_path(const std::string &filename, const std::string &wallet_dir)
+  {
+    boost::filesystem::path path(filename);
+    if (path.is_absolute() || wallet_dir.empty()) return filename;
+    if (!boost::filesystem::path(wallet_dir).is_absolute())
+    {
+      fail_msg_writer() << "--wallet-dir " << tr("argument is not an absolute path, ignoring it.");
+      return filename;
+    }
+    return (boost::filesystem::path(wallet_dir) / path).string();
   }
 } // anonymous namespace
 
@@ -3859,7 +3871,7 @@ bool simple_wallet::set_log(const std::vector<std::string> &args)
   return true;
 }
 //----------------------------------------------------------------------------------------------------
-bool simple_wallet::ask_wallet_create_if_needed()
+bool simple_wallet::ask_wallet_create_if_needed(const std::string &wallet_dir)
 {
   LOG_PRINT_L3("simple_wallet::ask_wallet_create_if_needed() started");
   std::string wallet_path;
@@ -3881,6 +3893,9 @@ bool simple_wallet::ask_wallet_create_if_needed()
         LOG_ERROR("Unexpected std::cin.eof() - Exited simple_wallet::ask_wallet_create_if_needed()");
         return false;
       }
+
+      wallet_path = resolve_wallet_path(wallet_path, wallet_dir);
+
       if(!tools::wallet2::wallet_valid_path_format(wallet_path))
       {
         fail_msg_writer() << tr("Wallet name not valid. Please try again or use Ctrl-C to quit.");
@@ -4009,6 +4024,7 @@ bool simple_wallet::init(const boost::program_options::variables_map& vm)
     m_electrum_seed.wipe();
   });
 
+  const std::string wallet_dir = command_line::get_arg(vm, arg_wallet_dir);
   const bool testnet = tools::wallet2::has_testnet_option(vm);
   const bool stagenet = tools::wallet2::has_stagenet_option(vm);
   if (testnet && stagenet)
@@ -4034,7 +4050,7 @@ bool simple_wallet::init(const boost::program_options::variables_map& vm)
   }
   else if (m_generate_new.empty() && m_wallet_file.empty() && m_generate_from_device.empty() && m_generate_from_view_key.empty() && m_generate_from_spend_key.empty() && m_generate_from_keys.empty() && m_generate_from_multisig_keys.empty() && m_generate_from_json.empty())
   {
-    if(!ask_wallet_create_if_needed()) return false;
+    if(!ask_wallet_create_if_needed(wallet_dir)) return false;
   }
 
   bool enable_multisig = false;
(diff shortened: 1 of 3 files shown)


nahuhh:
I'm not sure i see the point / use case here.

old:
```
./monero-wallet-cli --wallet-file $HOME/mewallets/mywalletfile
```

new:
```
./monero-wallet-cli --wallet-dir $HOME/mewallets

Enter wallet name: mywalletfile
```

Makes no sense?

edit: i forgot that this can be used in a config file

src/simplewallet/simplewallet.h
```diff
@@ -403,6 +403,7 @@ namespace cryptonote
 
   private:
     std::string m_wallet_file;
+    std::string m_wallet_dir;
```
selsta: Unused

src/simplewallet/simplewallet.cpp
```diff
@@ -612,6 +613,10 @@ namespace
 
     return true;
   }
+  static std::string prefix_wallet_dir(const std::string &filename, const boost::program_options::variables_map &vm)
+  {
+    return (command_line::is_arg_defaulted(vm, arg_wallet_dir) ? "" : command_line::get_arg(vm, arg_wallet_dir) + "/") + filename;
```
selsta: Double check if this works with absolute path used for filename and with `--wallet-dir ""`.

src/simplewallet/simplewallet.cpp
```diff
@@ -4428,7 +4437,7 @@ bool simple_wallet::init(const boost::program_options::variables_map& vm)
         auto rc = tools::wallet2::make_from_json(vm, false, m_generate_from_json, password_prompter);
         m_wallet = std::move(rc.first);
         password = rc.second.password();
-        m_wallet_file = m_wallet->path();
+        m_wallet_file = prefix_wallet_dir(m_wallet->path(), vm);
```
selsta: The wallet is alredy generated at this point, changing it afterwards is likely wrong.

src/simplewallet/simplewallet.cpp
selsta, line 3919: This function uses `wallet_dir_path` to check things but here `wallet_path` again. Is this intentional?

src/simplewallet/simplewallet.cpp
```diff
@@ -169,6 +169,7 @@ namespace
 {
   constexpr std::array<std::string_view, 5> allowed_priority_strings = tools::fee_priority_utilities::fee_priority_strings;
   const auto arg_wallet_file = wallet_args::arg_wallet_file();
+  const command_line::arg_descriptor<std::string> arg_wallet_dir = {"wallet-dir", sw::tr("Directory to look for wallet files"), ""};
```
selsta: Description needs update if it also changes the location where files are written.

selsta (changes requested):

src/simplewallet/simplewallet.cpp
selsta: Also this might need `boost::filesystem::path` instead of hardcoding `"/"`.

src/simplewallet/simplewallet.cpp
SNeedlewoods: `wallet_dir_path` is the `wallet_dir` + `wallet_path`, the `wallet_dir` also gets applied to `m_wallet_file` at a later point, so this prevents from prefixing the `wallet_dir` twice.

src/simplewallet/simplewallet.cpp
SNeedlewoods: Right, do you agree we can just ignore `--wallet-dir` when using `--generate-from-json`?
Then maybe we should mention that somewhere, though I don't think [here](https://github.com/monero-project/monero/blob/d509c53ca82749f1f680a5f6f4b1bdf50b3dee93/src/wallet/wallet_args.cpp#L73) is the right place, since that is also used by `monero-wallet-rpc` which handles `--wallet-dir` slightly differently.

selsta: I would match the behaviour to `--wallet-file`. If that's also ignored then yes.

SNeedlewoods:
This currently does not work with the following combination (relative path in `wallet_dir` and no wallet file):
`monero-wallet-cli --wallet-dir relativePath/`
then you go in "interactive mode" ([src](https://github.com/monero-project/monero/blob/f9fe258d117c3c43f2ad6828218a3cf4e88a0edf/src/simplewallet/simplewallet.cpp#L3862))  to enter a wallet file name.
To check if the file exists, the method needs to apply the `wallet_dir` flag there, this method sets `m_wallet_file` or `m_generate_new`. The same vars are getting used for the flags `--wallet-file` or `--generate-new-wallet` (see [here](https://github.com/monero-project/monero/pull/10670/changes#diff-25b9648b3da507060201d4ce30539c281676672c1af9a8e4d0e16ec0585a987bR4483)). Which leads to applying `wallet_dir` twice in that case.

A possible solution:
If a `wallet_dir` gets applied in `resolve_wallet_path()` return the absolute path, so if it gets called a second time, it will return early here, instead of applying the `wallet_dir` twice:
```
    if (path.is_absolute()) return filename;
```

SNeedlewoods:
There is still one issue, but that's not introduced by this PR (, AFAICT same issue exists for `monero-wallet-rpc`, `monerod`, etc.).
If providing a path as a flag e.g. `--wallet-file ~/wallet` it gets properly expanded to `/home/user/wallet`, if you provide the exact same path in a config file it doesn't expand and instead uses the `~` literally.

src/simplewallet/simplewallet.cpp
```diff
@@ -612,6 +613,19 @@ namespace
...
+  {
+    std::string wallet_dir = command_line::get_arg(vm, arg_wallet_dir);
+    if (wallet_dir == "") return filename;
+    else if (!boost::filesystem::path(wallet_dir).is_absolute())
+    {
+      fail_msg_writer() << "--wallet-dir " << tr("argument is not an absolute path, ignoring it.");
+      return filename;
+    }
+    boost::filesystem::path path(filename);
+    if (path.is_absolute()) return filename;
```
selsta: Move this up to the beginning of the function, otherwise it prints a warning message while the help text claims the parameter is ignored when filename path is absolute.

src/simplewallet/simplewallet.cpp
```diff
@@ -612,6 +613,19 @@ namespace
 
     return true;
   }
+  static std::string resolve_wallet_path(const std::string &filename, const boost::program_options::variables_map &vm)
```
selsta: Could you add a couple test cases for all the edge cases?

selsta: Tests were run locally, we decided to not add them to the test suite for now.

src/simplewallet/simplewallet.cpp
```diff
@@ -612,6 +613,18 @@ namespace
 
     return true;
   }
+  static std::string resolve_wallet_path(const std::string &filename, const std::string &wallet_dir)
+  {
+    boost::filesystem::path path(filename);
+    if (path.is_absolute()) return filename;
+    if (wallet_dir == "") return filename;
+    else if (!boost::filesystem::path(wallet_dir).is_absolute())
```
selsta:
```suggestion
    if (wallet_dir.empty()) return filename;
    if (!boost::filesystem::path(wallet_dir).is_absolute())
```

Small nit but `empty()` is preferred here and the `else` is unnecessary.

src/simplewallet/simplewallet.cpp
```diff
@@ -4169,7 +4186,7 @@ bool simple_wallet::init(const boost::program_options::variables_map& vm)
         return false;
       }
 
-      m_wallet_file=m_generate_from_view_key;
+      m_wallet_file = resolve_wallet_path(m_generate_from_view_key, wallet_dir);
```
selsta: `resolve_wallet_path` is called twice in this path which could lead to duplicate warnings. Same for `m_generate_from_keys`.

src/simplewallet/simplewallet.cpp
```diff
@@ -3876,6 +3888,9 @@ bool simple_wallet::ask_wallet_create_if_needed()
         "Specify wallet file name (e.g., MyWallet). If the wallet doesn't exist, it will be created.\n"
         "Wallet file name (or Ctrl-C to quit)")
       );
+
+      wallet_path = resolve_wallet_path(wallet_path, wallet_dir);
```
selsta: This should be moved after `cin.eof()`, so that it doesn't run on unexpected input.

selsta (approved):

plowsof (approved):

PR #10674, opened 2026-05-28
translations: don't include locations

tobtoht:
Based on #10673.

This removes the locations from translation files. It also updates the `*.ts` generation script to not include them in future runs.

Updating the translation files produces very spammy git diffs because line numbers change for almost every translation. Removing them makes future PRs easier to review. The locations are not needed to generate the `.qm` files.

### DIFF
--- a/translations/monero.ts
+++ b/translations/monero.ts
@@ -4,12 +4,10 @@
 <context>
     <name>Monero::AddressBookImpl</name>
     <message>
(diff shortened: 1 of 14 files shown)


selsta (approved):
I'm still for fully removing them but until there is consensus for it this is good to merge.

plowsof (approved):

PR #10675, opened 2026-05-28
remove translations

tobtoht:
- We do not include translations in releases and never have.
- We do not currently have someone interested in managing and maintaining a translation effort.
- None of the translations are finished:

```
fr: 81%
it: 63%
sv: 48%
de: 30%
ja: 23%
hi: 9%
es: 8%
nl: 4%
pl: 0%
sk: 0%
tr: 0%
pt-br: 0%
```

- None of the translations have been updated in 7 years.
- We do not have a glossary or translation guide, causing inconsistent or misleading translations.
- The current implementation has a dependency on `lrelease` (Qt), which I do not support.
- Translations for CLI tools have less expected utility than graphical applications.
- This repository has 44% more translatable strings than the GUI (1208 vs. 836), i.e. larger maintenance burden.
- As far as I can tell, there has been little to no demand for translations for `monero-wallet-cli` or `monerod`. No relevant issues were opened since 2018.


SeriousPassenger:
If they are too outdated it's not a bad idea to purge them. However, it's important to make Monero as universal as possible. So, there should be new efforts to introduce new translations/update the existing ones.

I can contribute for the Turkish ones. I am a native Turkish speaker.

Another idea is using AI to automate the translation process from the base English text to multiple languages, mark them as "not ready", and let native speakers proofread and modify them (for the awkward sounding/wrong ones) before adding them to production release.

iamamyth:
Depending on qt for translations in a CLI seems ill-conceived; offering translations might be reasonable, but the format of the translation data depends on the tool used. ICU and gettext seem like the most common choices, though even a trivial library like https://github.com/KaanBahaSever/i18n-cpp might do the job (not recommending that particular library, just illustrating narrowly-scoped, header-only libraries exist for this purpose).

tobtoht:
If we're not going to depend on Qt, then keeping the current format and implementation is pointless. When someone spearheads a new translations effort, it should be relatively straightforward to grab the translations from the `release-v0.18` branch and convert the already translated strings to a new format. There would need to be significant demand to justify introducing a new dependency, though I'm OK with vendoring a header-only library. It also seems entirely plausible to implement something that would suit our needs from scratch.

SeriousPassenger:
I am ready to fill in the Turkish translations for the CLI tools once the maintainers reach a consensus about the framework and I can just download the empty TR language set, fill in everything one by one being careful about the context. Pretty sure there are other people that will do this for other languages in the community.

iamamyth:
> If we're not going to depend on Qt, then keeping the current format and implementation is pointless. When someone spearheads a new translations effort, it should be relatively straightforward to grab the translations from the release-v0.18 branch and convert the already translated strings to a new format.

There's a decent chance a lot of the app text, along with the translations, need to be reworked anyways (e.g. the gettext manual has a list of common mistakes, and I'd hazard a guess the wallet makes all of them). In any case, yes, version control makes it possible to obtain old translations.

selsta (approved):
I fully support this change, in my opinion it's better to put translation effort into `monero-gui`.

nahuhh:
concept ACK, but have not tested

selsta (approved):

jeffro256:
My gut says that this is a bad idea. For the moment, Monero is more popular in English-speaking countries AFAICT. Development happens almost exclusively in English. For non-English-speaking countries, in my experience they are using mobile wallets like Cake. So at the moment, there is very very little demand for translations in the core repo, but that might not be the case forever. There is definitely a filter that biases support for this PR: the discussion to remove non-English translations in happening in English ;)

Unlike some other large feature sets that we have dropped recently, the translations honestly has almost never added any mental burden to me while working. Are they under maintained? Maybe. But also they don't require any active maintenance AFAIK. Selsta is probably right that translation efforts in the Monero GUI are more rewarding, but I don't think that that necessarily means that we drop the work done here.

iamamyth:
> Unlike some other large feature sets that we have dropped recently, the translations honestly has almost never added any mental burden to me while working

That's largely because the current approach doesn't produce a sound product. Usable international software requires a rigorous process, and the translations show, more than anything, there's none. For example, comparing the process suggested in the gettext manual, to the current one, illustrates fixing the current translation approach would require changes to development process as well as reworking a lot of the existing strings.

> Are they under maintained? Maybe. But also they don't require any active maintenance AFAIK. Selsta is probably right that translation efforts in the Monero GUI are more rewarding, but I don't think that that necessarily means that we drop the work done here.

What work? This is a fairly obvious contradiction: On the one hand, you claim there's no work involved in maintaining the translations, and, on the other, recommend against dropping the work.

I will note, maintaining the "translations" (if one can call mapping a string to itself a translation) does require work, it's just done by machines. Presumably, the point of this PR is that machines aren't free, so the project shouldn't waste them.

I happen to agree that the software probably should work in non-English languages, but I also think that, at the moment: a non-English user would most likely just use machine translation; and the current choice of translation library/support seems rather difficult to justify. Translation requires rigor: Either a product adopts the requisite processes to ensure a sound result, or it doesn't; the middle ground helps no one. I have seen many projects waste time sprinkling `_` everywhere, maintaining unused translation files, and perhaps even translating a few strings in select languages, for zero effect.

thomasbuilds:
I fully agree with this cleanup

thomasbuilds (approved):

jpk68 (approved):
I support this.

thomasbuilds:
You forgot to remove internationalization bullet point in the summary of `README.md`

selsta (approved):

jpk68 (approved):

thomasbuilds (approved):

selsta:
Needs rebase.

selsta (approved):

thomasbuilds (approved):

jpk68 (approved):

PR #10677, opened 2026-05-28
depends: boost: don't build stubs for header-only libraries

tobtoht:


CMakeLists.txt
```diff
@@ -956,7 +956,7 @@ elseif(Boost_FOUND)
 
   # Boost Regex is header-only since 1.77
   if (Boost_VERSION_STRING VERSION_LESS 1.77.0)
-    list(APPEND BOOST_COMPONENTS regex)
+    list(APPEND BOOST_COMPONENTS regex date_time)
```
iamamyth: Not sure >= 1.77.0 is a relevant constraint here. From the [docs](https://www.boost.org/doc/libs/latest/doc/html/date_time/details.html#date_time.design_goals), it seems whether to build the library (and how to include date_time without the built library) depends on the use case:
```
The library has a few functions that require the creation of a library file (mostly to_string, from_string functions). Most library users can make effective use of the library WITHOUT building the library, but simply including the required headers. If the library is needed, the Jamfile in the build directory will produce a "static" library (libboost_date_time) and a "dynamic/shared" library (boost_date_time) that contains these functions. Note that to use the library without the library (particularly on windows) may require using the BOOST_DATE_TIME_NO_LIB flag to the compilation options. 
```

tobtoht: https://github.com/boostorg/date_time/tree/boost-1.77.0

>Header only

>Note that the built library is only for build backward compatibility and contains no symbols. date_time is now header only.

https://github.com/boostorg/date_time/tree/boost-1.76.0

>Requires Linking

tobtoht: Ok, it builds fine without.

selsta (approved):

plowsof (approved):

PR #10679, opened 2026-05-28
ci: macos: explicitely install unbound

tobtoht:
https://github.com/monero-project/monero/actions/runs/26603468559/job/78392711942?pr=10677#step:6:62

We don't directly depend on `expat`, it's an (optional) transitive dependency for `unbound`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,7 +52,7 @@ jobs:
         run: |
           brew uninstall cmake
           brew update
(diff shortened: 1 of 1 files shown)


selsta (approved):

nahuhh:
> We don't directly depend on `expat`, it's an (optional) transitive dependency for `unbound`.

is this only relevant for ci and/or only for macos?

Should README dependencies & contrib/brew/brewfile be changed/updated?

tobtoht:
@nahuhh I was about to submit a PR for that ;)

PR #5821, opened 2019-08-19
core_tests: faster

moneromooo-monero:
Much faster.

PR #10686, opened 2026-05-29
workflows: speed up slow-hash in PR core_tests

selsta:
https://github.com/monero-project/monero/pull/5821 was rejected because it added runtime code to critical code path, so I'm suggesting this alternative. For security reasons it's only run on pull requests, the full tests are still run on merge.

Splitting up `core_tests` from the other tests also has the advantage that the more frequently failing tests run first.

Reduces `core_tests` runtime from 2610.67 sec to 273.01 sec (~90% reduction).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -186,9 +186,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # Most popular Ubuntu LTS version
-          - name: Ubuntu 24.04
-            container: ubuntu:24.04
+          # Oldest supported Ubuntu LTS version
+          - name: Ubuntu 22.04
+            container: ubuntu:22.04
     container:
       image: ${{ matrix.container }}
       env:
@@ -234,9 +234,9 @@ jobs:
     strategy:
       matrix:
         include:
-          # Oldest supported Ubuntu LTS version
-          - name: Ubuntu 22.04
-            container: ubuntu:22.04
+          # Most popular Ubuntu LTS version
+          - name: Ubuntu 24.04
+            container: ubuntu:24.04
     container:
       image: ${{ matrix.container }}
       env:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -272,15 +272,23 @@ jobs:
...
           ${{env.BUILD_DEFAULT}}
-          cmake --build build --target test
+      - name: run tests except core_tests
+        env:
+          DNS_PUBLIC: tcp://9.9.9.9
+        run: ctest --test-dir build --output-on-failure -E core_tests
+      - name: use reduced slow-hash iterations for core_tests
+        if: github.event_name == 'pull_request'
+        run: |
+          sed -i 's/#define ITER           (1 << 20)/#define ITER           20/' src/crypto/slow-hash.c
```
iamamyth: Fiddling with the source code (using sed, no less) makes the overall structure incredibly brittle; better to simply let the caller define a namespaced preprocessor macro (e.g. MONERO_CRYPTO_SLOW_HASH_ITER), which you can then default to `(1 << 20)` (i.e. the source continues to use `ITER`, it just makes the definition conditional).

selsta: I thought about this but I did not find a way how to do this cleanly with CMake without adding `target_compile_definitions` to our CMake, which is something I wanted to avoid. Overall using `sed` seemed to be the least bad solution.

iamamyth: `sed` is a terrible option because it means that syntactically valid code changes will break the replacement logic, slowing the test suite (i.e. effectively reverting the PR) without breaking the build. There are a lot of options to do this with cmake, `target_compile_definitions` could be part of the solution, although `add_compile_definitions` would allow for finer-grained scoping (i.e. the definition need not apply unconditionally to the target).

selsta: Updated, it now uses CFLAGS.

tobtoht (approved):

iamamyth (approved):
Based on build times, seems to achieve the desired effect.

PR #10691, opened 2026-05-31
p2p: fix slow shutdown with Tor/I2P enabled

selsta:
Tor/I2P zones share the public zone's io_context. Shutdown previously stopped zone net servers before all zone connections had been closed, which could leave remaining Tor/I2P connection shutdown work unable to run and cause 5s per-connection waits.

Close all zone connections before stopping any zone net server, and make shutdown idempotent so duplicate stop requests do not re-enter the shutdown sequence.

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.h
+++ b/contrib/epee/include/net/abstract_tcp_server2.h
@@ -320,7 +320,7 @@ namespace net_utils
 
 		bool speed_limit_is_enabled() const; ///< tells us should we be sleeping here (e.g. do not sleep on RPC connections)
 
-    bool cancel();
+    bool cancel(bool wait_for_shutdown = false);
     
   private:
     //----------------- i_service_endpoint ---------------------
@@ -378,8 +378,21 @@ namespace net_utils
     /// wait for service workers stop
     bool timed_wait_server_stop(uint64_t wait_mseconds);
 
+    /// Mark the server as stopping without closing connections or stopping the io_context.
+    bool mark_stop_signal_sent();
+
+    /// Close boosted_tcp_server-owned connections, including ones not yet registered with the protocol handler.
+    void close_server_connections();
+
+    /// Stop the server io_context.
+    void stop_io_context();
+
     /// Stop the server.
-    void send_stop_signal(std::function<void()> close_all_connections = [](){});
+    ///
+    /// Warning: Do NOT call this if the io_context is shared for connections
+    /// managed outside the boosted_tcp_server. See p2p net_node shutdown for
+    /// the correct staged shutdown in that case.
+    void send_stop_signal();
 
     bool is_stop_signal_sent() const noexcept { return m_stop_signal_sent; };
 
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -1581,7 +1581,11 @@ namespace net_utils
   template<class t_protocol_handler>
   void boosted_tcp_server<t_protocol_handler>::send_stop_signal(std::function<void()> close_all_connections)
   {
-    m_stop_signal_sent = true;
+    if (m_stop_signal_sent.exchange(true))
+    {
+      MDEBUG("Stop signal already sent");
+      return;
+    }
     typename connection<t_protocol_handler>::shared_state *state = static_cast<typename connection<t_protocol_handler>::shared_state*>(m_state.get());
     state->stop_signal_sent = true;
     TRY_ENTRY();
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -1121,9 +1121,9 @@ namespace net_utils
   }
 
   template<typename T>
-  bool connection<T>::cancel()
+  bool connection<T>::cancel(const bool wait_for_shutdown)
   {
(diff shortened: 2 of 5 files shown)


j-berman (commented):
This gets rid of using the `close_all_connections` callback that was introduced in #10382. TL;DR I think that's fine, and I think we should get rid of that param in the code too with this change.

I originally implemented closing the connections via a callback so that `m_stop_signal_sent` would get set to true [here](https://github.com/monero-project/monero/blob/e01d042382e526c3f2651044d578db0acb62ff15/contrib/epee/include/net/abstract_tcp_server2.inl#L1584-L1586) first, before the connections close. I figured that was necessary for a faster/correct graceful shutdown. Upon closer inspection, it doesn't seem to be a strict requirement given `m_stop_signal_sent` is mainly checked in while loops that would be unaffected, and the `state->stop_signal_sent` isn't even read ever.

Also +1 on the only-once stop signal send, nice.

selsta:
@j-berman updated to remove the callback

j-berman (approved):
Of note, any connections in `connections_` would have a similar issue (they can't properly finish the close sequence since io context can be stopped). I think that's fine to handle in a separate PR. That `c->cancel()` should probably be updated to `close(true/*wait_for_shutdown)` as well.

The connections in `connections_` look to be connections the node is in the middle of attempting to initiate an outbound connection to, so a narrower edge case to trigger issues with.

contrib/epee/include/net/abstract_tcp_server2.h
```diff
@@ -98,6 +98,7 @@ namespace net_utils
     public i_service_endpoint,
     public connection_basic
   {
+    template<class> friend class boosted_tcp_server;
```
j-berman: This shouldn't be necessary

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1579,24 +1579,33 @@ namespace net_utils
...
-      c->cancel();
-    }
-    connections_.clear();
-    connections_mutex.unlock();
 
-    // Since we shut down connections in the strand, we want to make sure to complete the shutdown sequence before
-    // stopping the io_context. We let the caller handle closing because the caller is the one keeping track of all
-    // connections (connections_ is only a subset of all connections).
-    close_all_connections();
+    close_server_connections();
```
j-berman, line 1629: We shouldn't call this here, otherwise the same issue could theoretically resurface

src/p2p/net_node.inl
```diff
@@ -1107,28 +1107,40 @@ namespace nodetool
...
+    }
     MDEBUG("[node] stopping server payload handler");
     m_payload_handler.stop();
-    MDEBUG("[node] sending stop signal");
+
+    MDEBUG("[node] closing connections");
     for (auto& zone : m_network_zones)
     {
-      const auto close_all_connections = [&]()
+      zone.second.m_net_server.close_server_connections();
```
j-berman, line 1127: Nit: can just put this in the next `for (auto& zone : m_network_zones)` loop below

j-berman:
Really, I do think we want this shutdown sequence:

1) Stop the cryptonote protocol handler (we already do this first).
2) Set `m_stop_signal_sent` to true.
    - This should ideally prevent the server from allowing any **new** connections to be made, either incoming or outgoing (though it doesn't look implemented perfectly at the moment).
3) Close all existing connections and wait for them to close.
4) Stop the io context.

This PR is flipping 2 & 3 as is, and I don't love that it does that. It feels like a move toward a worse pattern that may surface issues with new connections, though it does fix the case noted in this PR. I'm honestly surprised there aren't more issues surfacing in testing (e.g. in-flight connections taking longer to close). Since I don't have a concrete issue identified with moving in this PR's direction, I wont' hold this back.

contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -1579,26 +1579,48 @@ namespace net_utils
...
-    MDEBUG("Done with send_stop_signal");
+  }
+  //---------------------------------------------------------------------------------
+  template<class t_protocol_handler>
+  void boosted_tcp_server<t_protocol_handler>::send_stop_signal()
+  {
+    TRY_ENTRY();
+    if (!mark_stop_signal_sent())
+      return;
+    close_server_connections();
```
j-berman, line 1629: Same issue as before: shouldn't call this here, otherwise the same issue could theoretically surface.

All connections *should* be closed by this point before stopping the io context. We could log an error if `connections_` has any members inside `stop_io_context`. That would be something to fix.

j-berman (commented):
Sorry, I think how you had it where the other non-p2p callers that call `boosted_tcp_server<t_protocol_handler>::send_stop_signal()` just do the default flow inside `send_stop_signal` is ok. Repeating the `mark_stop_signal_sent` -> `close_server_connections` -> `stop_io_context` all these places seems unnecessary (my bad, you were correct).

I think it would be good to bring that back and document with a comment above `send_stop_signal` that says something like:

```
Warning: Do NOT call this if the io context is shared for connections managed outside the boosted tcp server. See how the p2p net_node handles stopping for that case.
```

And the error log in `stop_io_context` is good thank you.

j-berman (approved):
That last comment is up to you. The functionality looks correct to me.

j-berman (approved):
Thank you

tobtoht (approved):

Issue #10690, opened 2026-05-30
Remove daemon bootstrap functionality


selsta:
Rationale: 
 
- Daemon bootstrap was added in #3165 when Monero was younger, to make slow initial sync less painful and encourage users to run local nodes.
- In practice, it's hidden behind a flag that most beginners would never set, so the target audience likely does not use it, except indirectly through `monero-gui`.
- Users can already manually switch to a remote node when needed.
- Automatic bootstrap discovery is being abused by spy nodes.
- `monero-gui` is being [updated](https://github.com/monero-project/monero-gui/pull/4601) to avoid using bootstrap functionality.
- Bootstrap support adds a lot of edge cases for a relatively small convenience feature.
- We have received multiple AI-assisted HackerOne reports around bootstrap-specific behavior.
- Third-party implementations such as Cuprate did not implement this feature yet.
- Removing it reduces daemon complexity and attack surface.
- Simplifies FCMP++ integration, see for example https://github.com/seraphis-migration/monero/pull/355 and https://github.com/seraphis-migration/monero/pull/363.

nahuhh:
My initial take on the issue was that we should deprecate the "auto" functionality, retaining the ability to manually set a bootstrap daemon.

pros: there is improved safety in doing this, and improves ux by using a user-set / truated remote node
cons: as listed above, its a maintenance burden

other: is there any purpose to --no-sync aside from working in conjunction with bootstrap in auto mode as an ugly proxy that is better done by just connecting directly to a remote node? aka "simple mode" -- a feature that i really dont like..
if not, im all for removing both "auto" and no-sync. I'd like to nuke "simple mode".

"simple bootstrap mode" -- i dont have as big of a problem with. My problem would be specifically with the unsafe use of "auto" mode.

if keeping bootstrap mode, it could be improved by allowing a user to specify a list of nodes to choose from.

im not against nuking bootstrap altogether, but keeping non-auto bootstrap could be a workable middleground? Im not sure the extent of the hackerone reports, of course

selsta:
Just removing auto mode won't address the points above apart from the spy nodes. The code complexity / attack surface comes from regular bootstrap mode.

Gingeropolous:
yeah, its time for removal. No idea if we have any data to indicate that users actually use it.

nahuhh:
> yeah, its time for removal. No idea if we have any data to indicate that users actually use it. 

fwiw, basicswap currently uses it, and of course there are plenty of simple mode gui users.

[...]

PR #10694, opened 2026-05-31
remove bootstrap daemon

tobtoht:
#10690


selsta (approved):

nahuhh:
when running with the deprecated flags, the node fails to start. This can have some unexpected behavior for users and platforms that have configs that contain these flags.

suggestion: allow the daemon to start, but print a warning that the flags are deprecated

edit: maybe not important. I don't know of any service or user that runs these flags aside from gui or basicswap. So maybe fine as is

nahuhh:
@tobtoht  feel free to ignore my prior comment


should we cleanup get_info fields? (like `was_bootstrap_ever_used`)

tobtoht:
@nahuhh I expected that removing the fields would cause breakage for `release-v0.18` CLI wallets connecting to newer daemons. However, I tested this now and found that the fields are initialized to convenient values, so removing them from the RPC response does not cause problems.

tobtoht:
Removed the commit that removed `--no-sync`, because @jeffro256 mentioned it was useful for debugging the DB.

selsta (approved):

src/rpc/core_rpc_server.h
```diff
@@ -259,27 +253,10 @@ namespace cryptonote
     uint64_t get_block_reward(const block& blk);
     bool fill_block_header_response(const block& blk, bool orphan_status, uint64_t height, const crypto::hash& hash, block_header_response& response, bool fill_pow_hash);
     std::map<std::string, bool> get_public_nodes(uint32_t credits_per_hash_threshold = 0);
```
thomasbuilds:
```suggestion
```

src/rpc/core_rpc_server.cpp
```diff
@@ -215,33 +197,6 @@ namespace cryptonote
     return result;
   }
   //------------------------------------------------------------------------------------------------------------------------------
```
thomasbuilds:
```suggestion
```

thomasbuilds (commented):
This is dead code now

src/rpc/core_rpc_server.h
tobtoht: Am I missing something? This doesn't appear to be dead code.

thomasbuilds: `get_public_nodes` is dead code after this PR, possibly you confused with `on_get_public_nodes` ?

thomasbuilds (approved):

selsta (approved):

jeffro256:
For the record, this causes a regression with Haveno: https://github.com/haveno-dex/haveno/issues/2520

PR #10696, opened 2026-06-01
fix off-by-one buffer over-read in match_string2 unicode escape

alhudz:
The `\u` branch in match_string2 (parserse_base_utils.cpp:132) guards on `buf_end - it < 4`, but the loop right after runs `*++it` four times and dereferences `it == buf_end` when a JSON string ends with `\u` plus exactly three hex digits, reading one byte past the parser's [begin, end) range. Require 5 remaining bytes so the four hex digits stay inside the buffer.

### DIFF
--- a/contrib/epee/src/parserse_base_utils.cpp
+++ b/contrib/epee/src/parserse_base_utils.cpp
@@ -129,7 +129,7 @@ namespace misc_utils
             case '/':  //Slash character
               val.push_back('/');break;
             case 'u':  //Unicode code point
-              if (buf_end - it < 4)
+              if (buf_end - it < 5)
               {
                 ASSERT_MES_AND_THROW("Invalid Unicode escape sequence");
               }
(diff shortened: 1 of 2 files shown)


selsta:
@alhudz could you also open this PR against release-v0.18 branch?

selsta (approved):

alhudz:
Done, opened #10697 against release-v0.18.

iamamyth:
Please add a test for this scenario.

vtnerd:
I don't see where the extra deference occurs. What line number? Because this looks correct at a glance.

vtnerd (approved):

alhudz:
@vtnerd the deref is the `*++it` at line 141, inside the `for (int i = 0; i < 4; ++i)` loop. At that point `it` is on the `u`. The old guard at line 132 only required `buf_end - it >= 4`, but the loop advances `it` four times, so the 4th `++it` lands on `buf_end` when the string ends with `\u` plus exactly three hex digits. That last `*++it` reads one past the caller's [begin, end) range. Bumping the guard to `< 5` keeps all four digits in range.

@iamamyth added a test in epee_utils.cpp covering the truncated escape where buf_end sits right after the hex digits (no closing quote), which is the case that triggered the over-read.

selsta (approved):

PR #10697, opened 2026-06-01
fix off-by-one buffer over-read in match_string2 unicode escape

alhudz:
Backport of #10696 to release-v0.18.

The `\u` branch in match_string2 (parserse_base_utils.cpp:132) guards on `buf_end - it < 4`, but the loop right after runs `*++it` four times and dereferences `it == buf_end` when a JSON string ends with `\u` plus exactly three hex digits, reading one byte past the parser's [begin, end) range. Require 5 remaining bytes so the four hex digits stay inside the buffer.


selsta (approved):

vtnerd (approved):
Nevermind, I was wrong about this.

PR #10700, opened 2026-06-01
read multisig restore fields with memcpy in wallet2::generate

alhudz:
Noticed wallet2::generate reads the multisig restore blob through `*(uint32_t*)` and `*(crypto::secret_key*)` casts straight off the byte buffer, which is an aliased read and undefined behavior. Switched them to memcpy into typed locals, the same change made recently for the levin buffer reads.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -5641,9 +5641,10 @@ void wallet2::generate(const std::string& wallet_, const epee::wipeable_string& 
 
   THROW_WALLET_EXCEPTION_IF(multisig_data.size() < 32, error::invalid_multisig_seed);
   size_t offset = 0;
-  uint32_t threshold = *(uint32_t*)(multisig_data.data() + offset);
+  uint32_t threshold, total;
+  memcpy(&threshold, multisig_data.data() + offset, sizeof(uint32_t));
   offset += sizeof(uint32_t);
-  uint32_t total = *(uint32_t*)(multisig_data.data() + offset);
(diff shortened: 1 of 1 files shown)


selsta (approved):

src/wallet/wallet2.cpp
```diff
@@ -5654,22 +5655,30 @@ void wallet2::generate(const std::string& wallet_, const epee::wipeable_string&
 
   std::vector<crypto::secret_key> multisig_keys;
   std::vector<crypto::public_key> multisig_signers;
-  crypto::secret_key spend_secret_key = *(crypto::secret_key*)(multisig_data.data() + offset);
+  crypto::secret_key spend_secret_key;
+  memcpy(&spend_secret_key, multisig_data.data() + offset, sizeof(crypto::secret_key));
```
vtnerd: This should call `unwrap(unwrap(spend_secret_key))`. I think some compilers give warnings about this now, as `secret_key` is not trivial copyable.

vtnerd (changes requested):

src/wallet/wallet2.cpp
alhudz: Done, switched all three `secret_key` reads to `memcpy(&unwrap(unwrap(...)), ...)`, same as `chacha.h` does for `chacha_key`. The `public_key` ones are left as-is since `ec_point` is trivially copyable.

selsta:
Please squash

vtnerd (approved):

alhudz:
Squashed.

selsta (approved):

PR #10702, opened 2026-06-03
wallet2: fix double unlock of a lock [release-v0.18]

SChernykh:
`m_daemon_rpc_mutex` is already guarded by a scoped `lock` here. Unlocking it twice is an UB.

This code doesn't exist on master branch, so I'm creating this PR for release-v0.18 only.


selsta (approved):
This code path is unused anyway.

tobtoht (approved):

jeffro256 (approved):

PR #10716, opened 2026-06-04
Refactored tx verification NIC logic

SChernykh:
Encapsulated `nic_verified_hf_version` and `txs_by_txid` to make it impossible for callers to skip updating it.

@jeffro256

### DIFF
--- a/src/cryptonote_protocol/cryptonote_protocol_handler.inl
+++ b/src/cryptonote_protocol/cryptonote_protocol_handler.inl
@@ -91,8 +91,6 @@ namespace cryptonote
     const bool allow_pruned,
     cryptonote::pool_supplement& pool_supplement)
   {
-    pool_supplement.nic_verified_hf_version = 0;
-
     if (tx_entries.size() > blk_tx_hashes.size())
     {
       MERROR("Failed to make pool supplement: Too many transaction blobs!");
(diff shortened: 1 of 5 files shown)


src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -91,8 +91,6 @@ namespace cryptonote
     const bool allow_pruned,
     cryptonote::pool_supplement& pool_supplement)
   {
-    pool_supplement.nic_verified_hf_version = 0;
```
jeffro256, line 94: I think that `pool_supplement_t` needs a user-defined default constructor now because AFAIK the value of `nic_verified_hf_version` is undefined after a default initialization.

jeffro256 (changes requested):
This is a good change. Just have one nit

src/cryptonote_protocol/cryptonote_protocol_handler.inl
SChernykh: `std::uint8_t nic_verified_hf_version = 0;` in the struct definition makes it always defined?

jeffro256: Oh you're right, I forgot about that.

jeffro256 (approved):

selsta (approved):

PR #10719, opened 2026-06-04
p2p: track chain entry object requests with a boolean

selsta:
`m_num_requested` was used only to track whether any objects had been requested from the current chain entry, but it was not reset consistently when the related request state was cleared. Replace it with an explicit boolean and reset it alongside `m_requested_objects`, making the intended lifetime clear.

If this is considered unnecessary churn I can instead keep `m_num_requested` and only fix the missing reset sites.


jpk68 (approved):

selsta:
Closing as this is a bit useless, will open a more broad PR with sync related fixes in the future.

PR #10362, opened 2026-03-11
FCMP++: point to cycle scalar impl

j-berman:
Builds on top of:
- #10361

Introduces the functions to convert a Helios point to Selene scalar (`helios_point_to_selene_scalar`), and Selene point to Helios scalar (`selene_point_to_helios_scalar`), which are critical when growing the curve trees merkle tree.

The curve trees merkle tree's leaf layer is composed of Selene scalars. To get the next layer's elements, we hash a chunk of leaves (Selene scalars) at a time, and that hash is a point on the Selene curve. Then to get the next layer's elements, we convert each of those Selene points into Helios scalars using `selene_point_to_helios_scalar`, and hash a chunk of those at a time to get Helios points. Then to get the next layer's elements, we convert each of those Helios points into Selene scalars using `helios_point_to_selene_scalar`, and hash a chunk of those at a time to get Selene points. And so on, alternating between Selene and Helios at each layer.

PR #10724, opened 2026-06-05
FCMP++: curve tree builder

j-berman:
Builds on top of:
  - #10362

Introduces the `get_tree_extension` function used to "extend" an existing (or empty) curve tree. This is the core tree building function. It accepts as params the number of leaves already in the tree, the current tree edge (the last hashes in every layer), and the new outputs to add to the tree. It uses those params to build a struct that can be used to extend the tree.

This PR also includes a unit test that constructs a tree in memory, extends it, and audits the result. It tests various tree structure combinations in an effort to capture as many edge cases as possible.

For more context on the tree and its place in FCMP++, see [these WIP docs](https://github.com/j-berman/monero/blob/7c6205403ed88924d48eb0dfb0a71140cd7902bb/docs/FCMP%2B%2B_INTEGRATION.md#2-curve-trees-merkle-tree) ([source PR](https://github.com/seraphis-migration/monero/pull/110)).


src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+template<typename C_CHILD, typename C_PARENT>
+static std::vector<typename C_PARENT::Scalar> next_child_scalars_from_children(const std::unique_ptr<C_CHILD> &c_child,
+    const typename C_CHILD::Point *last_root,
+    const LayerExtension<C_CHILD> &children)
+{
+    std::vector<typename C_PARENT::Scalar> child_scalars_out;
+    child_scalars_out.reserve(1 + children.hashes.size());
+
+    // If we're creating a *new* root at the existing root layer, we may need to include the *existing* root when
+    // hashing the *existing* root layer
```
UkoeHB, line 162: This comment makes no sense. Why would you be hashing the root layer unless you were making a new root layer that's even higher?

Is it supposed to say this?
```
// If the existing root layer will become a middle layer,
// we need to include the existing root as a normal layer node.
```

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+        : curve->zero_scalar();
+
+    const auto chunk_start = new_child_scalars.data();
+    const typename C::Chunk chunk{chunk_start, chunk_size};
+
+    MTRACE("First chunk existing_hash: " << curve->to_string(existing_hash) << " , start_offset: " << start_offset
+        << " , prior_child_after_offset: " << curve->to_string(prior_child_after_offset));
+
+    for (std::size_t i = 0; i < chunk_size; ++i)
+        MTRACE("Hashing child in first chunk " << curve->to_string(chunk_start[i]));
```
UkoeHB, line 206: Need to check/assert `chunk_size == new_child_scalars.size()` or just get it from len instead of passing in.

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+}
+//----------------------------------------------------------------------------------------------------------------------
+template<typename C>
+static void hash_next_chunk(const std::unique_ptr<C> &curve,
+    const std::size_t chunk_start_idx,
+    const std::vector<typename C::Scalar> &new_child_scalars,
+    const std::size_t chunk_size,
+    typename C::Point &hash_out)
+{
+    const auto chunk_start = new_child_scalars.data() + chunk_start_idx;
```
UkoeHB, line 229: Also here, an assert that `chunk_size` correctly aligns with `new_child_scalars.size() + chunk_start_idx`.

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+    if (need_old_last_child)
+    {
+        CHECK_AND_ASSERT_THROW_MES(old_total_children > 0, "no old children but last child is supposed to change");
+
+        // If the chunk is full, must subtract the chunk width by 1
+        offset = offset == 0 ? (parent_chunk_width - 1) : (offset - 1);
+    }
+
+    // 8. When the last parent changes, we'll need to use its old value to update itself
+    const bool adding_members_to_existing_last_chunk = old_total_parents > 0 && !last_chunk_is_full
```
UkoeHB, line 403: Technically `adding_members_to_existing_last_chunk` is a little ambiguous/confusing when the last chunk's last member will change. But that case is implicitly covered by the following `need_old_last_child` check.

Nothing to change, just taking note.

UkoeHB (changes requested):
Part 1

src/fcmp_pp/curve_trees.cpp
j-berman: Yep, that's what I was trying to say there

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+            .next_parent_start_index           = next_parent_start_index,
+        };
+
+};
+//----------------------------------------------------------------------------------------------------------------------
+static GrowLayerInstructions get_leaf_layer_grow_instructions(const uint64_t old_n_leaf_tuples,
+    const uint64_t new_n_leaf_tuples,
+    const std::size_t leaf_tuple_size,
+    const std::size_t leaf_layer_chunk_width)
+{
```
UkoeHB, line 445: Add an assert that `leaf_layer_chunk_width % leaf_tuple_size == 0`.

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+// Helper function used to get the next layer extension used to grow the next layer in the tree
+// - for example, if we just grew the parent layer after the leaf layer, the "next layer" would be the grandparent
+//   layer of the leaf layer
+template<typename C_CHILD, typename C_PARENT>
+static LayerExtension<C_PARENT> get_next_layer_extension(const std::unique_ptr<C_CHILD> &c_child,
+    const std::unique_ptr<C_PARENT> &c_parent,
+    const GrowLayerInstructions &grow_layer_instructions,
+    const std::vector<typename C_CHILD::Point> &child_last_hashes,
+    const std::vector<typename C_PARENT::Point> &parent_last_hashes,
+    const std::vector<LayerExtension<C_CHILD>> child_layer_extensions,
```
UkoeHB, line 512:
```suggestion
    const std::vector<LayerExtension<C_CHILD>> &child_layer_extensions,
```

src/fcmp_pp/curve_trees.h
```diff
@@ -0,0 +1,230 @@
...
+    // To: [O.x,O.y,I.x,I.y,C.x,C.y,O.x,O.y,I.x,I.y,C.x,C.y...]
+    std::vector<typename C1::Scalar> flatten_leaves(std::vector<LeafTuple> &&leaves) const;
+
+    // Take in the existing number of leaf tuples and the existing last hash in each layer in the tree, as well as new
+    // outputs to add to the tree, and return a tree extension struct that can be used to extend a tree
+    TreeExtension get_tree_extension(const uint64_t old_n_leaf_tuples,
+        const LastHashes &existing_last_hashes,
+        std::vector<std::vector<UnifiedOutput>> &&new_outputs) const;
+
+    // Calculate the number of elems in each layer of the tree based on the number of leaf tuples
```
UkoeHB, line 166:
```suggestion
    // Calculate the number of elems in each layer of the tree based on the number of leaf tuples.
    // Doesn't include the leaf layer.
```

src/fcmp_pp/curve_trees.h
```diff
@@ -0,0 +1,230 @@
...
+    std::vector<uint64_t> n_elems_per_layer(const uint64_t n_leaf_tuples) const;
+
+    // Calculate how many layers in the tree there are based on the number of leaf tuples
```
UkoeHB, line 169:
```suggestion
    // Calculate how many layers in the tree there are based on the number of leaf tuples.
    // Doesn't include the leaf layer.
```

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+    const std::size_t last_updated_child_idx,
+    const std::size_t last_updated_parent_idx)
```
UkoeHB, line 514:
```suggestion
    const std::size_t active_child_layer_idx,
    const std::size_t active_parent_layer_idx)
```
The 'last updated' semantics are confusing and overly-complicated.

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+    std::vector<typename C1::Scalar> &flattened_leaves_out,
+    std::vector<UnifiedOutput> &valid_outputs_out) const
+{
+    flattened_leaves_out.clear();
+    valid_outputs_out.clear();
+
+    TIME_MEASURE_START(set_valid_leaves);
+
+    // Keep track of valid outputs to make sure we only use leaves from valid outputs. Can't use std::vector<bool>
+    // because std::vector<bool> concurrent access is not thread safe.
```
UkoeHB, line 795: Didn't know this

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+    tools::threadpool& tpool = tools::threadpool::getInstanceForCompute();
+    tools::threadpool::waiter waiter(tpool);
+    const std::size_t n_threads = std::max<std::size_t>(1, tpool.get_max_concurrency());
+
+    TIME_MEASURE_START(convert_valid_leaves);
+    // Step 1. Multithreaded convert valid outputs into Edwards derivatives needed to get Wei coordinates
+    std::vector<PreLeafTuple> pre_leaves;
+    pre_leaves.resize(new_outputs.size());
+    const std::size_t LEAF_CONVERT_BATCH_SIZE = std::max<std::size_t>(1, (new_outputs.size() / n_threads));
+    for (std::size_t i = 0; i < new_outputs.size(); i += LEAF_CONVERT_BATCH_SIZE)
```
UkoeHB, line 811: Not sure if the compiler will always optimize out these `.size()` calls, which need to be evaluated every loop.

Might be inconsequential here, just pointing it out.

UkoeHB (changes requested):
Part 2 (done)

Good work

src/fcmp_pp/curve_trees.cpp
j-berman: Ya other sections of the code actually use `std::deque<bool>` to work around this ([example](https://github.com/monero-project/monero/blob/17891e91575a3350e769c1c28aa0dc5d9404c2d2/src/ringct/rctSigs.cpp#L1501))

src/fcmp_pp/curve_trees.h
```diff
@@ -0,0 +1,230 @@
...
+    void outputs_to_leaves(std::vector<UnifiedOutput> &&new_outputs,
+        std::vector<typename C1::Scalar> &flattened_leaves_out,
+        std::vector<UnifiedOutput> &valid_outputs_out) const;
+
+    // Helper function used to set the next layer extension used to grow the next layer in the tree
+    // - for example, if we just grew the parent layer after the leaf layer, the "next layer" would be the grandparent
+    //   layer of the leaf layer
+    GrowLayerInstructions set_next_layer_extension(
+        const GrowLayerInstructions &prev_layer_instructions,
+        const bool parent_is_c1,
```
thomasbuilds, line 183:
```suggestion
        const bool parent_is_c2,
```

tests/unit_tests/curve_trees.h
```diff
@@ -0,0 +1,108 @@
...
+using CurveTreesV1 = fcmp_pp::curve_trees::CurveTreesV1;
+
+//----------------------------------------------------------------------------------------------------------------------
+namespace test
+{
+const std::vector<fcmp_pp::UnifiedOutput> generate_random_outputs(const std::size_t old_n_leaf_tuples,
+    const std::size_t new_n_leaf_tuples);
+
+std::shared_ptr<CurveTreesV1> init_curve_trees_test(const std::size_t helios_chunk_width,
+    const std::size_t selene_chunk_width,
```
thomasbuilds, line 48:
```suggestion
std::shared_ptr<CurveTreesV1> init_curve_trees_test(const std::size_t selene_chunk_width,
    const std::size_t helios_chunk_width,
```

tests/unit_tests/curve_trees.h
```diff
@@ -0,0 +1,108 @@
...
+    bool grow_tree(const std::size_t expected_old_n_leaf_tuples,
+        const std::size_t new_n_leaf_tuples,
+        const std::vector<fcmp_pp::UnifiedOutput> &new_outputs,
+        const bool audit_tree = true);
+
+    // Validate the in-memory tree by re-hashing every layer, starting from root and working down to leaf layer
+    bool audit_tree(const std::size_t expected_n_leaf_tuples) const;
+
+    // get all leaf indices with given output pair
+    std::set<std::size_t> get_leaf_idxs_with_output_pair(const fcmp_pp::OutputPair &output_pair) const;
```
thomasbuilds, line 89:
```suggestion
```

tests/unit_tests/curve_trees.h
```diff
@@ -0,0 +1,108 @@
...
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include "fcmp_pp/curve_trees.h"
+#include "fcmp_pp/fcmp_pp_types.h"
+#include "fcmp_pp/tower_cycle.h"
+
+#include <set>
```
thomasbuilds, line 35:
```suggestion
```

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+        .C_y = tower_cycle::selene_scalar_from_bytes(C_y)
+    };
+}
+//----------------------------------------------------------------------------------------------------------------------
+static CurveTrees<Selene, Helios>::LeafTuple output_tuple_to_leaf_tuple(const OutputTuple &output_tuple)
+{
+    const auto plt = output_tuple_to_pre_leaf_tuple(output_tuple);
+    return pre_leaf_tuple_to_leaf_tuple(plt);
+}
+//----------------------------------------------------------------------------------------------------------------------
```
thomasbuilds, line 625:
```suggestion
```

src/fcmp_pp/curve_trees.cpp
```diff
@@ -0,0 +1,1070 @@
...
+    std::vector<LeafTuple> &&leaves) const;
+//----------------------------------------------------------------------------------------------------------------------
+template<typename C1, typename C2>
+typename CurveTrees<C1, C2>::TreeExtension CurveTrees<C1, C2>::get_tree_extension(
+    const uint64_t old_n_leaf_tuples,
+    const LastHashes &existing_last_hashes,
+    std::vector<std::vector<UnifiedOutput>> &&new_outputs) const
+{
+    TreeExtension tree_extension;
+    tree_extension.leaves.start_idx = old_n_leaf_tuples;
```
thomasbuilds, line 668: Optional: rejects a LastHashes whose size does not match the tree up front instead of only when a hash happens to be needed. The callers in fcmp++-stage build one hash per layer and grow_tree passes with it.

```suggestion
    tree_extension.leaves.start_idx = old_n_leaf_tuples;

    CHECK_AND_ASSERT_THROW_MES((existing_last_hashes.c1_last_hashes.size() + existing_last_hashes.c2_last_hashes.size())
        == this->n_layers(old_n_leaf_tuples), "unexpected number of last hashes");
```

src/fcmp_pp/curve_trees.cpp
thomasbuilds: chunk_size here is min(new_child_scalars.size(), chunk_width - start_offset) so it only equals new_child_scalars.size() when everything fits in the first chunk. An assert needs <=.

tests/unit_tests/curve_trees.cpp
```diff
@@ -0,0 +1,759 @@
...
+        }
+
+        parent_is_c1 = !parent_is_c1;
+    }
+}
+//----------------------------------------------------------------------------------------------------------------------
+//----------------------------------------------------------------------------------------------------------------------
+// Test
+//----------------------------------------------------------------------------------------------------------------------
+TEST(curve_trees, grow_tree)
```
thomasbuilds, line 669: With these widths the largest extension is 19 chunks so HASH_BATCH_SIZE > 1 (several chunks per tpool task) is only reached with fewer than 10 hardware threads. One extension of a few thousand leaves at prod widths would pin that path.

thomasbuilds (commented):
Few nits

src/fcmp_pp/curve_trees.cpp
j-berman: This will be used eventually, and was included in a prior audit round

src/fcmp_pp/curve_trees.cpp
```diff
@@ -449,6 +453,8 @@ static GrowLayerInstructions get_leaf_layer_grow_instructions(const uint64_t old
     const std::size_t leaf_tuple_size,
     const std::size_t leaf_layer_chunk_width)
 {
+    CHECK_AND_ASSERT_THROW_MES(leaf_layer_chunk_width % leaf_tuple_size == 0, "unexpected leaf layer chunk width");
```
UkoeHB:
```
CHECK_AND_ASSERT_THROW_MES(leaf_tuple_size && leaf_layer_chunk_width % leaf_tuple_size == 0, "unexpected leaf layer chunk width");
```
Div 0 is UB

UkoeHB (changes requested):
Just a nit

UkoeHB (approved):

redsh4de:
Needs rebasae

jpk68 (approved):

UkoeHB (approved):

src/fcmp_pp/curve_trees.h
```diff
@@ -63,8 +104,8 @@ class CurveTrees
             m_c2_width{c2_width},
             m_leaf_layer_chunk_width{LEAF_TUPLE_SIZE * c1_width}
     {
-        assert(c1_width > 0);
-        assert(c2_width > 0);
+        assert(c1_width > 1);
+        assert(c2_width > 1);
```
j-berman, line 108: Updated these asserts in the latest since @selsta pointed out that `n_elems_per_layer` requires this

selsta (approved):

UkoeHB (approved):

jpk68 (approved):

Issue #10076, opened 2025-09-10
wallet-crypto-bench and cncrypto tests crash on sparc64 due to unaligned access (Bus error)


glaubitz:
The tests `wallet-crypto-bench` and `cncrypto` crash on Linux running on sparc64 due to unaligned access (Bus error):

```
The following tests FAILED:
	  2 - wallet-crypto-bench (Bus error)
	  3 - cncrypto (Bus error)
Errors while running CTest
```

Full build log available in: https://buildd.debian.org/status/fetch.php?pkg=monero&arch=sparc64&ver=0.18.4.2%2B%7E0%2B20200826-1&stamp=1757525344&raw=0

While sparc64 is not a mainstream platform, fixing the cause for the unaligned accesses on sparc64 might help fix the issues reported in #9255, #8889, #8837 and #2669.

Access to SPARC machines for testing can be obtained through the GCC Compile Farm: https://gcc.gnu.org/wiki/CompileFarm

selsta:
@Ap4sh can this issue be closed? You opened some of the fixes.

PR #10733, opened 2026-06-06
crypto: avoid unsafe AES word casts

Ap4sh:
## Summary
- use `memcpy` for AES block input/output word loads and stores instead of casting block bytes to `uint32_t *`
- keep expanded round-key reads on the direct word path through a `__may_alias__` word type
- make the expanded key parameter const to match the existing declarations

Related to #10076

## Testing
- `cmake --build build/debug-system --target cncrypto-tests monero-wallet-crypto-bench hash-tests hash-target-tests -j2`
- `ctest --test-dir build/debug-system --output-on-failure -R 'cncrypto|wallet-crypto-bench|hash-(fast|slow|slow-1|slow-2|slow-4)|hash-target'`
- `MONERO_USE_SOFTWARE_AES=1 ctest --test-dir build/debug-system --output-on-failure -R 'cncrypto|wallet-crypto-bench|hash-(fast|slow|slow-1|slow-2|slow-4)|hash-target'`
- `ctest --test-dir build/debug-system --output-on-failure -R 'cnv4-jit'`
- `cmake --build build/debug-noaes-system --target cncrypto-tests monero-wallet-crypto-bench hash-tests hash-target-tests -j2`
- `ctest --test-dir build/debug-noaes-system --output-on-failure -R 'hash-target|wallet-crypto-bench|cncrypto|hash-fast|hash-slow|hash-slow-1|hash-slow-2' -E '^hash-slow-4$'`
- `gcc -O2` and `clang -O2` object checks: same `aesb_single_round` and `aesb_pseudo_round` symbol sizes, no assembly diff vs `origin/master` except the object filename

Note: `hash-slow-4` still fails in the `NO_AES=ON` build with `Error generating CryptonightR code`; I checked the same failure on `origin/master` before restoring this patch

### DIFF
--- a/src/crypto/aesb.c
+++ b/src/crypto/aesb.c
@@ -19,6 +19,7 @@ Issue Date: 20/12/2007
 */
 
 #include <stdint.h>
+#include <string.h>
 #include "int-util.h"
 
 #if defined(__cplusplus)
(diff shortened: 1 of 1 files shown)


jeffro256:
This is technically UB because of the strict aliasing, but it won't cause unaligned memory accesses except on 8-bit and 16-bit systems, which neither have enough memory to run CryptoNight anyways. This is because the pointer passed as `expandedKey` by all of the calling locations has *fundamental alignment* due to it being a return value of `calloc()`. All 32+-bit systems that I know of will have a fundamental alignment greater than or equal to 4. I'd prefer a solution which didn't cause a slowdown in a semi-hot code path.

Ap4sh:
Good point on `expandedKey`, I reworked this to keep the round-key reads on the direct word path through a `__may_alias__` word type. `memcpy` is now only used for the AES block input/output word loads and stores, where the external block buffers are the part I wanted to avoid casting through `uint32_t *`

I also retitled the PR to avoid overstating the unaligned part. Local GCC 15.2 and Clang 19 `-O2` object checks produce the same `aesb_single_round` / `aesb_pseudo_round` symbol sizes and no assembly diff vs `origin/master` except the object filename

Reran the crypto tests in the normal, software AES, and `NO_AES=ON` builds, plus `cnv4-jit`

selsta:
Closing this one as we will go with #10739

PR #10739, opened 2026-06-07
crypto: fix strict aliasing for `expandedKey` in aesb.c

jeffro256:
Alternative to #10733. Uses 1 `memcpy()` call in `aesb_single_round()` and `aesb_pseudo_round()`, instead of 4 and 40, respectively. Removes `STATIC` and `INLINE` keywords which do nothing and are semantically incorrect. Diff is smaller.

### DIFF
--- a/src/crypto/aesb.c
+++ b/src/crypto/aesb.c
@@ -19,6 +19,7 @@ Issue Date: 20/12/2007
 */
 
 #include <stdint.h>
+#include <string.h>
 #include "int-util.h"
 
 #if defined(__cplusplus)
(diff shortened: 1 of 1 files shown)


src/crypto/aesb.c
```diff
@@ -141,29 +142,21 @@ extern "C"
 
 d_4(uint32_t, t_dec(f,n), sb_data, u0, u1, u2, u3);
 
-#if !defined(STATIC)
```
vtnerd, line 144: `static` in this context hides the function from visibility during the linking stage. So it wasn't pointless, although `inline` was particularly helpful here.

vtnerd (approved):

src/crypto/aesb.c
jeffro256: Yes, but `STATIC` is not a keyword. And here, the macro `STATIC` is defined as empty. In fact, this function *must not* be static, because it is used in a different translation unit: `slow-hash.c`. Same thing with `INLINE`.

jeffro256:
ARMv8 Android build failed due to a CI network error

selsta (approved):

src/crypto/aesb.c
vtnerd: I missed how this was implemented. Strange.

jeffro256: I suspect that it wasn't intentional, and is a copy-and-paste artifact

PR #10740, opened 2026-06-08
use byte-safe isalpha in epee json value parser

alhudz:
1. the json value parser detects `null`/`true`/`false` with the libc `isalpha(*it)` on a raw `char`, so a non-ASCII value byte is passed as a negative `int`, outside the `unsigned char`/`EOF` range `isalpha` is defined for, and the result is locale dependent.
2. `epee::misc_utils::parse` already has byte-safe `isspace`/`isdigit` over `lut`, and `lut` already sets the alpha bit, so the global call was the odd one out.

Added a matching `parse::isalpha` and switched the three value-keyword sites to it.

`repro`: `load_from_json` on `{"b": <0xc3>(}` (a value starting with a non-ASCII byte), reachable from RPC/HTTP and P2P JSON.
`expected`: rejected, the byte is not a keyword start.
`actual`: `isalpha((char)0xc3)` evaluated out of range (UB) and locale dependent.
`fix`: classify with the `lut` alpha bit, matching the adjacent `parse::isdigit`.

Test added covering `true`/`false`/`null` and the non-ASCII byte.

### DIFF
--- a/tests/unit_tests/epee_serialization.cpp
+++ b/tests/unit_tests/epee_serialization.cpp
@@ -98,6 +98,35 @@ struct ObjWithOptChild
     KV_SERIALIZE_OPT(test_value, true);
   END_KV_SERIALIZE_MAP()
 };
+
+struct ObjWithBool
+{
+  bool b;
+
+  BEGIN_KV_SERIALIZE_MAP()
+    KV_SERIALIZE(b)
+  END_KV_SERIALIZE_MAP()
+};
+}
+
+TEST(epee_json, keyword_values)
+{
+  // true/false/null parse as keyword values
+  ObjWithBool o{};
+
+  o.b = false;
+  EXPECT_TRUE(epee::serialization::load_t_from_json(o, std::string("{\"b\": true}")));
+  EXPECT_TRUE(o.b);
+
+  o.b = true;
+  EXPECT_TRUE(epee::serialization::load_t_from_json(o, std::string("{\"b\": false}")));
+  EXPECT_FALSE(o.b);
+
+  EXPECT_TRUE(epee::serialization::load_t_from_json(o, std::string("{\"b\": null}")));
+
+  // A value beginning with a non-ASCII byte (most significant bit == 1) is not
+  // a keyword and must be rejected.
+  EXPECT_FALSE(epee::serialization::load_t_from_json(o, std::string("{\"b\": \xc3\x28}")));
 }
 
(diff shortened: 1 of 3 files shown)


tests/unit_tests/epee_serialization.cpp
```diff
@@ -98,6 +98,36 @@ struct ObjWithOptChild
...
+
+  BEGIN_KV_SERIALIZE_MAP()
+    KV_SERIALIZE(b)
+  END_KV_SERIALIZE_MAP()
+};
+}
+
+TEST(epee_json, keyword_values)
+{
+  // true/false/null keyword detection uses a byte-safe, locale-independent
```
iamamyth: Would recommend test descriptions/comments focus on the desired observed behavior, rather than implementation details (branching being an implementation detail). For example, this one could be shortened to:
```
// Keyword parsing rejects non-ASCII characters
```

tests/unit_tests/epee_serialization.cpp
```diff
@@ -98,6 +98,36 @@ struct ObjWithOptChild
...
+  EXPECT_TRUE(o.b);
+
+  o.b = true;
+  EXPECT_TRUE(epee::serialization::load_t_from_json(o, std::string("{\"b\": false}")));
+  EXPECT_FALSE(o.b);
+
+  EXPECT_TRUE(epee::serialization::load_t_from_json(o, std::string("{\"b\": null}")));
+
+  // a value beginning with a non-ASCII byte (negative as a signed char) is not
+  // a keyword and must be rejected, not passed to the C isalpha out of range
```
iamamyth: Same here about implementation details (`isalpha`), you can simply drop the final clause and maybe change the non-ASCII clarification:
```
// A value beginning with a non-ASCII byte (most significant bit == 1) is not
// a keyword and must be rejected.

tests/unit_tests/epee_serialization.cpp
alhudz: Good point, dropped the implementation detail. Reworked the header to just describe what parses.

tests/unit_tests/epee_serialization.cpp
alhudz: Done, used your wording and dropped the isalpha clause.

iamamyth (approved):

selsta:
please squash

alhudz:
Done, squashed into a single commit.

selsta (approved):

PR #10744, opened 2026-06-08
wallet: fix warning signed unsigned comparison

SNeedlewoods:
https://github.com/monero-project/monero/actions/runs/27014746716/job/79726796084#step:6:645

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -12799,13 +12799,20 @@ uint64_t wallet2::get_approximate_blockchain_height() const
   uint64_t approx_blockchain_height = fork_block;
   const time_t now = time(NULL);
   if (now > fork_time)
-    approx_blockchain_height += (now - fork_time) / seconds_per_block;
-  else if (approx_blockchain_height > (fork_time - now) / seconds_per_block)
-    approx_blockchain_height -= (fork_time - now) / seconds_per_block;
+  {
+    const uint64_t blocks_since_last_fork = static_cast<uint64_t>((now - fork_time) / seconds_per_block);
+    approx_blockchain_height += blocks_since_last_fork;
+  }
   else
   {
-    LOG_ERROR("Failed to approximate blockchain height from future fork block: " << approx_blockchain_height);
-    return 0;
+    const uint64_t blocks_until_next_fork = static_cast<uint64_t>((fork_time - now) / seconds_per_block);
+    if (approx_blockchain_height > blocks_until_next_fork)
+      approx_blockchain_height -= blocks_until_next_fork;
+    else
+    {
+      LOG_ERROR("Failed to approximate blockchain height from future fork block: " << approx_blockchain_height);
+      return 0;
+    }
   }
   // testnet and stagenet got some huge rollbacks, so the estimation is way off
   const uint64_t approximate_rolled_back_blocks = m_nettype == TESTNET ? 26600 : m_nettype == STAGENET ? 48600 : 33600;


selsta (approved):

src/wallet/wallet2.cpp
```diff
@@ -12800,7 +12800,7 @@ uint64_t wallet2::get_approximate_blockchain_height() const
   const time_t now = time(NULL);
   if (now > fork_time)
     approx_blockchain_height += (now - fork_time) / seconds_per_block;
-  else if (approx_blockchain_height > (fork_time - now) / seconds_per_block)
+  else if (approx_blockchain_height > static_cast<uint64_t>((fork_time - now) / seconds_per_block))
```
iamamyth: If you're going to apply this cast, it seems reasonable to switch this to an else and store the value in a local to clarify the semantics (and guard against mistakes): Height has to be greater than value subtracted from height, otherwise the result isn't a valid block height (0 not valid).

SNeedlewoods: Do you mean something like this?
``` diff
diff --git a/src/wallet/wallet2.cpp b/src/wallet/wallet2.cpp
index d0e42d2e4..135276d62 100644
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -12800,12 +12800,16 @@ uint64_t wallet2::get_approximate_blockchain_height() const
   if (now > fork_time)
-  else if (approx_blockchain_height > static_cast<uint64_t>((fork_time - now) / seconds_per_block))
-    approx_blockchain_height -= (fork_time - now) / seconds_per_block;
   else
   {
-    LOG_ERROR("Failed to approximate blockchain height from future fork block: " << approx_blockchain_height);
-    return 0;
+    uin64_t blocks_until_next_fork = static_cast<uint64_t>((fork_time - now) / seconds_per_block);
+    if (approx_blockchain_height > blocks_until_next_fork)
+      approx_blockchain_height -= blocks_until_next_fork;
+    else
+    {
+      LOG_ERROR("Failed to approximate blockchain height from future fork block: " << approx_blockchain_height);
+      return 0;
+    }
   }
   // testnet and stagenet got some huge rollbacks, so the estimation is way off
   const uint64_t approximate_rolled_back_blocks = m_nettype == TESTNET ? 26600 : m_nettype == STAGENET ? 48600 : 33600;
```

iamamyth: Yes, that looks good.

src/wallet/wallet2.cpp
```diff
@@ -12800,12 +12800,16 @@ uint64_t wallet2::get_approximate_blockchain_height() const
   const time_t now = time(NULL);
   if (now > fork_time)
     approx_blockchain_height += (now - fork_time) / seconds_per_block;
-  else if (approx_blockchain_height > (fork_time - now) / seconds_per_block)
-    approx_blockchain_height -= (fork_time - now) / seconds_per_block;
   else
   {
-    LOG_ERROR("Failed to approximate blockchain height from future fork block: " << approx_blockchain_height);
-    return 0;
+    uint64_t blocks_until_next_fork = static_cast<uint64_t>((fork_time - now) / seconds_per_block);
```
```suggestion
    const uint64_t blocks_until_next_fork = static_cast<uint64_t>((fork_time - now) / seconds_per_block);
```

src/wallet/wallet2.cpp
```diff
@@ -12800,12 +12800,16 @@ uint64_t wallet2::get_approximate_blockchain_height() const
   const time_t now = time(NULL);
   if (now > fork_time)
     approx_blockchain_height += (now - fork_time) / seconds_per_block;
```
selsta: While you are at it, you might also want to cast here even if this does not warn with our compiler settings, so that the code is more symmetric with adding uint64_t to uint64_t in both branches.

selsta (approved):

tobtoht (approved):

PR #10748, opened 2026-06-09
combine utf-16 surrogate pairs in match_string2

alhudz:
Repro: hand the RPC/HTTP JSON deserialiser (`load_from_json`) a string with a UTF-16 surrogate pair, e.g. `{"x":"😀"}` for U+1F600.
Cause: `match_string2` decodes each `\uXXXX` on its own and emits a 3-byte sequence for any value up to `0xffff`, so the surrogate halves come out as CESU-8 `ED A0 BD ED B8 80` instead of UTF-8 `F0 9F 98 80`; lone and mismatched surrogates are accepted as well, yielding invalid UTF-8.
Fix: combine a high surrogate (`0xd800`-`0xdbff`) with the following low surrogate into one code point and emit the 4-byte form, and reject lone/mismatched surrogates. This is the same handling the bundled RapidJSON parser already does.

### DIFF
--- a/contrib/epee/src/parserse_base_utils.cpp
+++ b/contrib/epee/src/parserse_base_utils.cpp
@@ -129,18 +129,31 @@ namespace misc_utils
             case '/':  //Slash character
               val.push_back('/');break;
             case 'u':  //Unicode code point
-              if (buf_end - it < 5)
               {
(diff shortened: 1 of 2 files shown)


alhudz:
gentle ping

selsta (approved):

jpk68 (approved):

PR #10751, opened 2026-06-09
simplewallet: misc safety/correctness fixes

selsta:
https://codeberg.org/montero-project/monero/pulls/25

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -2825,11 +2825,6 @@ bool simple_wallet::set_device_name(const std::vector<std::string> &args/* = std
   const auto pwd_container = get_and_verify_password();
   if (pwd_container)
   {
-    if (args.size() == 0){
-      fail_msg_writer() << tr("Device name not specified");
-      return true;
-    }
-
     m_wallet->device_name(args[1]);
     bool r = false;
     try {
@@ -4279,6 +4274,7 @@ bool simple_wallet::init(const boost::program_options::variables_map& vm)
         auto rc = tools::wallet2::make_from_json(vm, false, m_generate_from_json, password_prompter);
         m_wallet = std::move(rc.first);
         password = rc.second.password();
+        if (!m_wallet) return false;
         m_wallet_file = m_wallet->path();
       }
       catch (const std::exception &e)
@@ -4286,8 +4282,6 @@ bool simple_wallet::init(const boost::program_options::variables_map& vm)
         fail_msg_writer() << e.what();
         return false;
       }
-      if (!m_wallet)
-        return false;
     }
     else if (!m_generate_from_device.empty())
     {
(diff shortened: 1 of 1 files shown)


src/simplewallet/simplewallet.cpp
```diff
@@ -7018,7 +7038,7 @@ bool simple_wallet::sweep_main(uint32_t account, uint64_t below, const std::vect
 
   priority = m_wallet->adjust_priority(priority);
 
-  size_t fake_outs_count = m_wallet->get_min_ring_size() - 1;
+  size_t fake_outs_count = m_wallet->get_min_ring_size() > 0 ? m_wallet->get_min_ring_size() - 1 : 0;
```
jeffro256:
```suggestion
  size_t fake_outs_count = std::max<uint64_t>(m_wallet->get_min_ring_size(), 1) - 1;
```

Prevents calling `get_min_ring_size()` twice, which may re-trigger an RPC call and/or return a different result than the previous call.

selsta: updated

jeffro256 (approved):
Should be squashed before merge

selsta:
Squashed.

src/simplewallet/simplewallet.cpp
```diff
@@ -2974,7 +2974,7 @@ bool simple_wallet::set_device_name(const std::vector<std::string> &args/* = std
   const auto pwd_container = get_and_verify_password();
   if (pwd_container)
   {
-    if (args.size() == 0){
+    if (args.size() < 2){
       fail_msg_writer() << tr("Device name not specified");
       return true;
```
SNeedlewoods: AFAICT this is unreachable. If you use `set device-name` you'll receive this error before you enter the function:
>Error: set "device-name": needs an argument (<device_name[:device_spec]>)

Due to this check https://github.com/monero-project/monero/blob/9e817b608b51b138273824afcde2a442b0764359/src/simplewallet/simplewallet.cpp#L3618

src/simplewallet/simplewallet.cpp
```diff
@@ -5285,6 +5296,8 @@ void simple_wallet::stop_background_mining()
 //----------------------------------------------------------------------------------------------------
 void simple_wallet::check_background_mining(const epee::wipeable_string &password)
 {
+  if (!m_wallet) return;
+
```
SNeedlewoods, line 5134: Not really needed, `check_background_mining()` gets called once in `init()` after `if (!m_wallet)` is already checked.

src/simplewallet/simplewallet.cpp
```diff
@@ -6456,7 +6476,7 @@ bool simple_wallet::transfer_main(const std::vector<std::string> &args_, bool ca
   priority = m_wallet->adjust_priority(priority);
 
   const size_t min_ring_size = m_wallet->get_min_ring_size();
-  size_t fake_outs_count = min_ring_size - 1;
+  size_t fake_outs_count = min_ring_size > 0 ? min_ring_size - 1 : 0;
```
SNeedlewoods: Just for consistency
```suggestion
  size_t fake_outs_count = std::max<uint64_t>(m_wallet->get_min_ring_size(), 1) - 1;
```

SNeedlewoods (commented):
Only small comments

SNeedlewoods (approved):

PR #10752, opened 2026-06-09
Remove daemon bootstrap functionality

### LINKED ISSUE #10690: Remove daemon bootstrap functionality

web3securityauditor:
Resolves #10690. This completely removes the bootstrap daemon and bootstrap node selector to reduce attack surface, address abuse by spy nodes, and simplify daemon complexity.


selsta:
Please see https://github.com/monero-project/monero/pull/10694

web3securityauditor:
Closing in favor of #10694. Sorry for the duplicate!

PR #10755, opened 2026-06-09
daemon: look for config file in data-dir if not specified

### LINKED ISSUE #10553: detecting a config file should look in the specified data dir
1. Insert a bitmonero.conf into the default datadir
2. Run monero wirh `./monerod` = the ~/.bitmonero/bitmonero.conf if used
3. Add a bitmonero.conf into a folder called bitmonero2
3. Run `./monerod --data-dir bitmonero2` = the bitmonero.conf from ~/.bitmonero is used instead. 

expected: the config source datadir should not be static, and should 1st check --config-file, 2nd check --data-dir. 
Only In absence of --data-dir and --config-file should it look in the default datadir

web3securityauditor:
Resolves #10553. When a custom `--data-dir` is specified but `--config-file` is not, the daemon should look for `bitmonero.conf` in the specified `data-dir` rather than the default `data-dir`. This patch corrects the config file parsing priority so that it dynamically uses the `data-dir` provided in the command line argument if the config file argument is defaulted. A test script has been added to prevent regressions.

### DIFF
--- a/src/daemon/main.cpp
+++ b/src/daemon/main.cpp
@@ -207,6 +207,11 @@ int main(int argc, char const * argv[])
 
     std::string config = command_line::get_arg(vm, daemon_args::arg_config_file);
     boost::filesystem::path config_path(config);
+    if (command_line::is_arg_defaulted(vm, daemon_args::arg_config_file))
+    {
+      config_path = boost::filesystem::absolute(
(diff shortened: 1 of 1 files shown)


selsta:
Please remove the test script, it does not fit into our test suite.

selsta:
Thank you for the PR, we are going to go with #10847 as it appears to be a more correct solution in case of the default config path being changed in the futture.

PR #10756, opened 2026-06-10
net: cap aggregate ZMQ receive size

selsta:


src/net/zmq.cpp
```diff
@@ -92,6 +92,8 @@ namespace zmq
 
     namespace
     {
+        constexpr std::size_t max_message_size = 10 * 1024 * 1024; // 10 MiB
```
vtnerd: It would be nice if this and the code in `zmq_server.cpp` used the same constant.

vtnerd (changes requested):

vtnerd (approved):

tobtoht (approved):

PR #10169, opened 2025-10-14
functional_tests: Python epee binary lib and test `/get_blocks.bin`

jeffro256:
Functional test `blockchain` will pass after #10652  is merged (with requested edits).

PR #10761, opened 2026-06-12
Daemon RPC: /getblocks.bin incremental pool fetch even when no new blocks

j-berman:
### Overview

Wallets call the `/getblocks.bin` endpoint in a loop, which returns any new blocks or pool txs since the wallet last requested. However, if the chain hasn't advanced blocks, the daemon could no-op and *not* return anything (see "quick check for noop").

### Before this PR

Before this PR, because the daemon hasn't advanced blocks, the daemon would skip returning pool txs inside the `get_pool` for the incremental pool fetching approach (since it would immediately return true when handling the no-op case). [Stressnet] wallet2 would then fallback to the non-incremental approach for pool fetching, which takes multiple rounds.*

### After this PR

The daemon still continues on to the `get_pool` incremental handling even if the daemon hasn't advanced any blocks past the wallet's sync height. The wallet therefore correctly uses the incremental pool fetching logic in this case.

Truthfully, I'm not sure how we missed this when testing incremental pool fetching in the past. I thought I tested this exact circumstance pretty extensively on different occasions, but I may have missed something, or the issue doesn't surface with a release wallet2 wallet. I observed this issue on stressnet when testing #10605.

### Why this PR is so big

It's copying almost every existing line exactly the same, but pulls out logic into a new `handle_get_blocks` function, which pollutes the diff. I figure this is the cleanest solution though.

______
*: I only tested this with a stressnet wallet, which modifies the way the wallet calls `/getblocks.bin`. It's possible the issue doesn't surface with the current release branch wallet, which would explain why this wasn't caught prior. Either way, it's an unexpected issue.

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -518,6 +518,82 @@ namespace cryptonote
     return true;
   }
   //------------------------------------------------------------------------------------------------------------------------------
+  bool core_rpc_server::handle_get_blocks(const COMMAND_RPC_GET_BLOCKS_FAST::request& req, COMMAND_RPC_GET_BLOCKS_FAST::response& res, const connection_context *ctx, std::size_t &cumul_block_data_size)
+  {
+    cumul_block_data_size = 0;
+
+    // quick check for noop
+    if (req.start_height > 0 || !req.block_ids.empty())
+    {
+      uint64_t last_block_height;
+      crypto::hash last_block_hash;
+      m_core.get_blockchain_top(last_block_height, last_block_hash);
+
+      if (req.start_height > last_block_height ||
+          (!req.block_ids.empty() && last_block_hash == req.block_ids.front()))
+      {
+        res.start_height = 0;
+        res.current_height = last_block_height + 1;
+        res.top_block_hash = last_block_hash;
+        res.status = CORE_RPC_STATUS_OK;
+        return true;
+      }
+    }
+
+    size_t max_blocks = req.max_block_count > 0
+      ? std::min(req.max_block_count, (uint64_t)COMMAND_RPC_GET_BLOCKS_FAST_MAX_BLOCK_COUNT)
+      : COMMAND_RPC_GET_BLOCKS_FAST_MAX_BLOCK_COUNT;
+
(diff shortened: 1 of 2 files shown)


jeffro256:
Is there a plan to use `handle_get_blocks()` anywhere else?

j-berman:
No, it's just a helper function strictly for this spot, introduced to capture the true return value of the no-op without returning in `core_rpc_server::on_get_blocks`. I could reduce the diff size by maintaining the old indents inside this new helper, or some other gross pattern, unless you see a cleaner solution here?

And I'll handle merge conflicts / open a PR to release once I get an ACK on this solution

rbrunner7 (approved):
Like you I have a hard time believing that somehow everybody did not crash into this bug while testing, but well, the `return` is undeniably there, and it can be taken, and if it is, no pool txs are returned, so this must be a bug.

I don't have a strong opinion on making fetching blocks its own method or not; I would probably refrain from doing so if I implemented this PR, but don't oppose. Don't you lose quite some history of the code in question if you move it to some different place?

jeffro256:
It would be nice to have a functional test case that catches this behavior. It's hard to believe that functional tests didn't catch this before, but I guess that the pattern of A) make tx B) mine blocks C) *then* refresh was super common?

selsta (approved):
Added a test: https://github.com/monero-project/monero/pull/10169#discussion_r3508875368

Without this PR

```
Calling /get_blocks.bin (blocks and incremental pool, no new blocks) and testing response...
  File "/Users/selsta/dev/monero/tests/functional_tests/blockchain.py", line 448, in <module>
    BlockchainTest().run_test()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/selsta/dev/monero/tests/functional_tests/blockchain.py", line 55, in run_test
    self.test_get_blocks_fast()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/selsta/dev/monero/tests/functional_tests/blockchain.py", line 441, in test_get_blocks_fast
    assert res.pool_info_extent == 1
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/selsta/dev/monero/utils/python-rpc/framework/rpc.py", line 51, in __getattr__
    return self[key]
           ~~~~^^^^^
KeyError: 'pool_info_extent'
```

With this PR it passes successfully.

jeffro256 (approved):

PR #10768, opened 2026-06-15
http: share deterministic header field parser

selsta:


contrib/epee/include/net/http_base.h
```diff
@@ -29,8 +29,10 @@
 #pragma once
 #include "memwipe.h"
 
+#include <boost/utility/string_view.hpp>
```
vtnerd, line 32: Should we update existing `string_ref` uses to `string_view` (to reduce duplicate header includes)?

contrib/epee/include/net/http_base.h
```diff
@@ -68,6 +70,48 @@ namespace net_utils
 			add_field(out, field.first, field.second);
 		}
 
+		namespace detail
+		{
+			inline bool is_http_token_char(char c)
```
vtnerd: Move all of these to a cpp? `is_http_token_char` and `trim_header_line` can be `static` or in anonymous namespace in the cpp. The `<cctype>` include could be removed.

I did a quick disassembly of this code, and its fairly compact so many the `inline` attempt is worth it.

vtnerd (commented):
These changes look like a good upgrade.

contrib/epee/include/net/http_base.h
selsta: Fixed.

contrib/epee/include/net/http_base.h
selsta: Updated all into one inline function and removed cctype.

SChernykh:
LGTM, but please add a test for a full HTTP request parsing, like this:
```
TEST(HTTP_Server, Rejects_Invalid_First_Header_Line)
{
  test_http_endpoint endpoint{};
  test_http_handler app{};
  http::custum_handler_config<epee::net_utils::connection_context_base> config{};
  config.m_phandler = &app;
  config.rng = rng;
  epee::net_utils::connection_context_base context{};
  http::http_custom_handler<epee::net_utils::connection_context_base> handler(&endpoint, config, context);

  const std::string request =
    "GET / HTTP/1.1\r\n"
    "Bad Header\r\n"
    "Host: example.com\r\n"
    "\r\n";

  EXPECT_FALSE(handler.handle_recv(request.data(), request.size()));
  EXPECT_FALSE(app.called);
  EXPECT_TRUE(endpoint.sent.empty());
}
```

contrib/epee/include/net/http_base.h
```diff
@@ -68,6 +68,39 @@ namespace net_utils
...
+				value = line.substr(colon + 1);
+				while(!value.empty() && (value.front() == ' ' || value.front() == '\t'))
+					value.remove_prefix(1);
+				while(!value.empty() && (value.back() == ' ' || value.back() == '\t'))
+					value.remove_suffix(1);
+
+				for(char c : name)
+				{
+					const bool alnum = (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9');
+					const bool allowed = alnum || c == '-' || c == '_';
```
SChernykh: According to RFC 9110, section 5.6.2, this is the list of allowed characters:
```
const bool allowed = alnum ||
      c == '!' || c == '#' || c == '$' || c == '%' || c == '&' ||
      c == '\'' || c == '*' || c == '+' || c == '-' || c == '.' ||
      c == '^' || c == '_' || c == '`' || c == '|' || c == '~';

```

contrib/epee/include/net/http_client.h
SChernykh: After the changes in `parse_header`, a return value must be checked here now. If it returns false -> set `reciev_machine_state_error` and return false.

SChernykh (approved):

tobtoht:
@vtnerd Can you approve?

contrib/epee/include/net/http_client.h
```diff
@@ -594,64 +598,46 @@ namespace net_utils
...
+						break;
+
+					boost::string_view name;
+					boost::string_view value;
+					CHECK_AND_ASSERT_MES(detail::parse_header_line(line, name, value), false, "http_stream_filter::parse_cached_header() invalid header in: " << m_cache_to_process);
+
+					const std::string key(name.data(), name.size());
+					const std::string val(value.data(), value.size());
+					if (!string_tools::compare_no_case(key, "Connection"))
+						body_info.m_connection = val;
```
vtnerd: These should really be moves. `val` will need `const` to be dropped or it's a silent copy.

vtnerd (approved):
Just a performance recommendation that can be done in a follow up

contrib/epee/include/net/http_client.h
selsta: done

vtnerd (approved):

PR #10770, opened 2026-06-15
wallet2: skip redundant tx-extra pubkey scan work

selsta:


SNeedlewoods (approved):

SNeedlewoods (approved):

selsta:
It was decided not to implement this due to compatibility issues, after this patch outputs could be missed.

PR #10771, opened 2026-06-15
wallet2: skip redundant tx-extra pubkey scan work [release-v0.18]

selsta:
#10770

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -2285,18 +2285,31 @@ void wallet2::cache_tx_data(const cryptonote::transaction& tx, const crypto::has
     const size_t rec_size = (is_miner && m_refresh_type == RefreshType::RefreshOptimizeCoinbase && tx.version < 2) ? 1 : tx.vout.size();
     if (!tx.vout.empty())
     {
-      // if tx.vout is not empty, we loop through all tx pubkeys
       const std::vector<boost::optional<cryptonote::subaddress_receive_info>> rec(rec_size, boost::none);
 
       tx_extra_pub_key pub_key_field;
       size_t pk_index = 0;
-      while (find_tx_extra_field_by_type(tx_cache_data.tx_extra_fields, pub_key_field, pk_index++))
+      // Cache primary tx pubkeys up to the number of outputs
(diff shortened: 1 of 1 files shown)


src/wallet/wallet2.cpp
```diff
@@ -2425,6 +2449,7 @@ void wallet2::process_new_transaction(const crypto::hash &txid, const cryptonote
       }
     }
 
+    const size_t outputs_found_before = std::count(output_found.begin(), output_found.end(), true);
```
SNeedlewoods, line 2452: AFAICT `outpus_found_before` will always be zero.
`output_found` gets initialized with all values set to `false` and then is not used until after this.

src/wallet/wallet2.cpp
```diff
@@ -2352,6 +2365,11 @@ void wallet2::process_new_transaction(const crypto::hash &txid, const cryptonote
   while (!tx.vout.empty())
   {
     std::vector<size_t> outs;
+    if (pk_index >= tx.vout.size())
+    {
+      MWARNING("tx " << txid << " has more tx pubkeys than outputs; ignoring excess tx pubkeys");
+      break;
+    }
     // if tx.vout is not empty, we loop through all tx pubkeys
```
SNeedlewoods: Nit: Actually not looping through "all" tx pubkeys here anymore. Though the warning in the block above implies that we loop through "all necessary" tx pubkeys.

SNeedlewoods (commented):
I want to be clear: this is a complex topic and I haven't checked every aspect of it, but AFAICJ skipping redundant tx pubkeys makes sense.
Just one comment and a nit-pick.

src/wallet/wallet2.cpp
selsta: `output_found` gets mutated by `check_acc_out_precomp_once`, so it's only zero during the first loop iteration.

src/wallet/wallet2.cpp
selsta: Good catch, will update.

SNeedlewoods (approved):

SNeedlewoods (approved):

PR #10777, opened 2026-06-15
http_client: add default response size limit

selsta:


vtnerd (commented):
Could this break tx scanning?

selsta:
Yes, it can have side effects. I think it's better to close this.

PR #10781, opened 2026-06-15
wallet2: fix sporadic rejected double spend errors

j-berman:
Multiple people have reported sporadic double spend errors on stressnet. @Rucknium's daemon and wallet logs helped me narrow down the cause* (EDIT: this still needs further investigation to identify the root cause).

### Before this PR

If a tx created by a wallet has been in the pool for >5mins, then the incremental logic might incorrectly think the tx has been kicked from the pool, even if the tx is still in the node's pool.

Then the wallet will mark the tx's key images as unspent, allowing the wallet to use the outputs in another tx. The subsequent tx then gets rejected by the node as a double spend, because the original tx is still in the node's pool.

### After this PR

This fixes the issue by assuming the node still has the already-seen tx in its pool unless the tx has been explicitly removed from the pool.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -3712,16 +3712,14 @@ void wallet2::process_unconfirmed_transfer(bool incremental, const crypto::hash 
   }
   else
   {
-    if (!incremental)
+    if (tx_details.m_state == wallet2::unconfirmed_transfer_details::pending_in_pool)
     {
-      if (tx_details.m_state == wallet2::unconfirmed_transfer_details::pending_in_pool)
-      {
-        // For the probably unlikely case that a tx once seen in the pool vanishes
-        // again set back to 'pending'
-        tx_details.m_state = wallet2::unconfirmed_transfer_details::pending;
-        MINFO("Already seen txid " << txid << " vanished from pool, marking as pending");
-      }
+      // For the probably unlikely case that a tx once seen in the pool vanishes
+      // again set back to 'pending'
+      tx_details.m_state = wallet2::unconfirmed_transfer_details::pending;
+      MINFO("Already seen txid " << txid << " vanished from pool, marking as pending");
     }
+
     // If a tx is pending for a "long time" without appearing in the pool, and if
     // we have refreshed and thus had a chance to really see it if it was there,
     // judge it as failed; the waiting for timeout and refresh happened avoids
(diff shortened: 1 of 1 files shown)


selsta (approved):

tobtoht (approved):

nahuhh:
I've sent 210k+ txs on stressnet using this (7000+ txs per wallet * 30 wallets) and have not run into any double spend errors

while that doesnt mean that i wont, at least its "so-far-so-good"

rbrunner7:
I tried to review, understand the problem, and understand the proposed fix, but somehow did not break through. Maybe you can elaborate your problem description a bit:

> If a tx created by a wallet has been in the pool for >5mins, then the incremental logic might incorrectly think the tx has been kicked from the pool, even if the tx is still in the node's pool.

j-berman:
@rbrunner7 

### Before this PR (i.e. current code)

When using the incremental approach for fetching pool txs, [this section](https://github.com/monero-project/monero/blob/2c48374ecd2449c02bb400e5bcf20b7c6f11649b/src/wallet/wallet2.cpp#L3910-L3918) iterates over all the **new** `added_pool_txs` since the wallet's last request for pool txs. It doesn't return **all** txs, only **new** ones. The wallet therefore might have `found` set to `false` there for a tx that's in the pool still, that the wallet already saw, but is not present in an incremental round's request for `added_pool_txs`. Here's the problematic flow:

- Wallet constructs tx A and submits to daemon.
- Wallet requests all pool txs at time t0, and daemon returns tx A since it's in the pool.
  - The wallet will mark tx A as `pending_in_pool` in `process_unconfirmed_transfer`.
- At time t1, the wallet requests incremental pool txs added to the daemon's pool since time t0.
- The daemon won't include tx A in `added_pool_txs`.
- The wallet therefore has `found=false` for that unconfirmed `m_unconfirmed_txs` entry (which is all the unconfirmed outgoing spends by the wallet).

Then when `found=false`, the wallet enters [this section](https://github.com/monero-project/monero/blob/2c48374ecd2449c02bb400e5bcf20b7c6f11649b/src/wallet/wallet2.cpp#L3718-L3742) processing the unconfirmed transfer. After 500s, the wallet can end up marking such a tx as failed, even if it's still in the daemon's pool. A failed tx's key images become available for use in the wallet again [here](https://github.com/monero-project/monero/blob/2c48374ecd2449c02bb400e5bcf20b7c6f11649b/src/wallet/wallet2.cpp#L3740). The wallet can then subsequently attempt to construct a tx with those now-assumed-available key images, even though the tx A is still in the pool. When the wallet does so, the daemon rejects the tx as a double spend (since it still has the pool tx A spending those key images in its pool).

### After this PR (i.e. this PR's code)

- Wallet constructs tx A and submits to daemon.
- Wallet requests all pool txs at time t0, and daemon returns tx A since it's in the pool.
  - The wallet will mark tx A as `pending_in_pool` in `process_unconfirmed_transfer`.
- At time t1, the wallet requests incremental pool txs added to the daemon's pool since time t0.
- The daemon won't include tx A in `added_pool_txs`.
- Since the daemon also doesn't include the tx in `removed_pool_txids`, the wallet assumes that tx A is still in the pool, and sets `is_in_pool=true`.
- The wallet therefore doesn't change anything about the tx's state in its local in `process_unconfirmed_transfer`.

j-berman:
Upon further review, I think I may be mistaken in identifying the problem in the original code, specifically at this part:

> After 500s, the wallet can end up marking such a tx as failed, even if it's still in the daemon's pool

In the flow I described, that shouldn't be possible because the transfer's state would need to be `pending` [here](https://github.com/monero-project/monero/blob/2c48374ecd2449c02bb400e5bcf20b7c6f11649b/src/wallet/wallet2.cpp#L3732) to get marked failed. If its state was `pending_in_pool`, then it wouldn't get marked failed. This PR would only "fix" the assumed scenario of it being `pending_in_pool`, but I now don't see how that can happen with the current code.

In @Rucknium's logs I noticed a problematic transfer's state was set to `failed` (even though the tx was in the daemon's pool), but the wallet logs didn't go back to the point in time that transfer actually changed from its initial state in the wallet. I assumed it was set to failed after the wallet identified it to be in the pool already because this section looked like a logic error in the code (my mistake). But in reality, it's possible the tx never appeared to the wallet to be in the daemon's pool in the first place (or was in the pool at one point, then removed). In which case, this PR wouldn't actually fix the root issue causing the observed sporadic double spend errors.

TL;DR unfortunately this will take further investigation on my end. I think we can remove this PR from the queue for inclusion in next release.

j-berman:
#10840 should be the actual fix for the underlying and observed issue. Above comment explains how I was mistaken with this PR.

PR #10784, opened 2026-06-16
rpc: replace get_blocks.bin Levin pool budget

selsta:
`/get_blocks.bin` is HTTP binary RPC serialized with portable storage, not a P2P Levin message. The previously added pool limiter used `LEVIN_DEFAULT_MAX_PACKET_SIZE` as a response budget, but that constant does not apply to the RPC path.

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -75,6 +75,8 @@ using namespace epee;
 #define RESTRICTED_SPENT_KEY_IMAGES_COUNT 5000
 #define RESTRICTED_BLOCK_COUNT 1000
 
+static constexpr size_t GET_BLOCKS_BIN_MAX_ADDED_POOL_TX_BODIES = 20000;
+
 #define RPC_TRACKER(rpc) \
   PERF_TIMER(rpc); \
   RPCTracker tracker(#rpc, PERF_TIMER_NAME(rpc))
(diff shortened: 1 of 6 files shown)


src/rpc/core_rpc_server.cpp
```diff
@@ -570,7 +572,6 @@ namespace cryptonote
 
     res.pool_info_extent = COMMAND_RPC_GET_BLOCKS_FAST::NONE;
 
-    size_t cumul_block_data_size = 0;
```
jeffro256, line 650: Why the name change from this descriptive name to `size`?

selsta: The variable no longer tracks cumulative block data for budgeting. It's now only a narrow-scope counter used for the debug log in this block. I can change the name back to `cumul_block_data_size` or to something more descriptive.

jeffro256: Yes, its no longer used for budgeting, but it's still a cumulative block data size, right?

jeffro256:
Needs rebase

src/rpc/core_rpc_server.cpp
selsta: Changed the name back.

selsta:
Rebased.

jeffro256 (approved):

j-berman (approved):

PR #10787, opened 2026-06-16
epee: lazy generation of ssl certificates

SNeedlewoods:
https://codeberg.org/montero-project/monero/pulls/29

### DIFF
--- a/contrib/epee/include/net/net_helper.h
+++ b/contrib/epee/include/net/net_helper.h
@@ -96,7 +96,7 @@ namespace net_utils
 			blocked_mode_client() :
 				m_io_service(),
 				m_ctx(boost::asio::ssl::context::tlsv12),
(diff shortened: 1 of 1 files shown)


contrib/epee/include/net/net_helper.h
```diff
@@ -513,6 +520,7 @@ namespace net_utils
 	protected:
 		boost::asio::io_context m_io_service;
 		boost::asio::ssl::context m_ctx;
```
jeffro256:
```suggestion
		std::optional<boost::asio::ssl::context> m_ctx;
```

This is a nit, but this is a good place for an optional: it won't let you use it uninitialized on accident, and we get to skip the `sslv23` constructor in the member field initializer list on startup.

plowsof:
im not sure how #10992 would interact with this

jpk68:
> im not sure how #10992 would interact with this

At first glance, it seems like there won't be any substantial issues, but I could be wrong...

PR #10790, opened 2026-06-18
tests/fuzz: fix cold-outputs import

jeffro256:
Doing `::serialization::serialize()` on a `std::tuple<uint64_t, uint64_t, std::vector<tools::wallet2::transfer_details>>` was the old method of transporting outputs messages, before view-key encryption. This call site must have been overlooked, because the file `tests/data/fuzz/cold-outputs/out-all-6` contains a payload of the new format. And the existing encrypted outputs file doesn't use authentication, so it fails. Change the call to `wallet2::import_outputs_as_str()` and add a file which allows fuzzing of code past viewkey decryption.


selsta (approved):
built `cold-outputs_fuzz_tests` with `libFuzzer` and with the PR, the new `out-v4-all-21` seed reaches the intended path and imports 21 outputs

plowsof (approved):

PR #10587, opened 2026-05-13
fix spelling in comments and docs

thomasbuilds:
Attempts to fix all spelling issues in comments and docs across the codebase

Fixes spelling mistakes in comments and documentation only: no code, variable names, or string literals were touched.

Review takes 3–5 minutes total, 68 files changed

PR #10792, opened 2026-06-18
Fix typos

thomasbuilds:
#10587 only fixed comments that didn't share a line with code. This fixes those too, plus string literals, two identifier renames (a parameter and a local variable), and renames `src/rpc/instanciations.cpp` → `instantiations.cpp`.

35 files touched (previous merged PR touched 61 files), behavior-preserving.


tobtoht (approved):
If we keep the spelling mistakes, we'll just keep seeing these PRs over and over again. A year from now, a decade from now. Let's just get it over with.

jeffro256 (approved):

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -327,7 +327,7 @@ namespace cryptonote
       << std::setw(20) << "Support Flags"      
       << std::setw(30) << "Recv/Sent (inactive,sec)"
       << std::setw(25) << "State"
-      << std::setw(20) << "Livetime(sec)"
+      << std::setw(20) << "Lifetime(sec)"
```
nahuhh, line 330: Is this correct?

selsta: It should print how long a peer is connected, so yes.

Issue #10763, opened 2026-06-12
p2p: allow --ban-list to load from a URL


NoFear0411:
## Problem

`--ban-list` accepts only a local file path. During `init()` (src/p2p/net_node.inl) the node verifies the path, reads the file, and parses one entry per line (IPv4 address or subnet, with `#` comments).

To track a maintained remote list such as https://gui.xmr.pm/files/block.txt, an operator has to download the file manually, store it locally, and repeat that whenever it changes. Pointing the flag at the URL directly would let the node fetch the current list on every startup with no manual step.

## Proposed change

Allow `--ban-list` to accept an `http://` or `https://` URL in addition to a file path. When the value is a URL, the node fetches the list once at startup, validates the response, and parses the body with the same parser used for local files. A local path behaves exactly as it does today.

Semantics kept consistent with the current file behavior:

- Fetched once at startup, not on a timer.
- Failure to obtain a usable list is fatal, matching the current behavior for a missing or unreadable file. This covers connection errors, a non-200 response, an empty body, and a body above a fixed size cap.
- For `https`, the TLS certificate is verified against the system CA store. An operator-supplied URL has no signed payload or DNS hash quorum, so certificate verification is the integrity check.
- Per-line parsing and the handling of invalid lines are unchanged.

This is the minimal version. It could later be extended to refresh the list periodically while running, giving a dynamic ban list without a restart. That is out of scope for this proposal and noted only as a possible follow-up.

## Related work

- #10595 adds URL downloading for the DNS blocklist, verified against a hash quorum in DNS. It targets the automatic blocklist, not the `--ban-list` flag, and uses a different trust model. This proposal does not change it.
- #10747 fixes IPv4 subnet arithmetic in the ban primitives this parser calls. Complementary.
- #9591 renames blocklist to banlist. Naming would follow whatever is current at implementation time.

## Questions

- Is loading `--ban-list` from a URL in scope, or should remote lists stay limited to the DNS blocklist?
- Size cap value. #10595 uses 10 MiB, and reusing it keeps both paths consistent.

NoFear0411:
Noting a pre-existing limitation in the ban list parser, separate from the URL feature itself but worth tracking as a possible follow-up.

[...]

PR #10794, opened 2026-06-18
p2p: allow --ban-list to load from a URL

NoFear0411:
Implements the `--ban-list` URL support proposed and discussed in #10763. That issue covers the motivation (pointing a node at a maintained remote list and fetching the current version on each startup), the design decisions (in-memory, fetch once, strict TLS, fatal on failure), and the relationship to the adjacent PRs #10595, #10747, and #9591.

### What

Extends `--ban-list` to accept an `http://` or `https://` URL in addition to a local file path. When the value carries an http(s) scheme, the list is fetched once at startup, validated, and parsed by the same per-line parser used for local files. A local path behaves exactly as before.

### Behavior

The fetched list is held in memory and never written to disk, so each startup uses a fresh copy and there is no cache to manage across restarts.

For `https`, the server certificate is verified against the system trust store. An operator-supplied URL has no signed payload or DNS hash quorum, so certificate verification is the integrity check.

Any failure to obtain a usable list is fatal and aborts startup, matching the existing behavior for a missing or unreadable file. Failures include a malformed URL, a failed connection, a non-200 response, an empty body, and a body larger than a fixed 10 MiB cap.

Per-line parsing is unchanged: `#` comments, whitespace trimming, blank-line skipping, IPv4 addresses, IPv4 subnets, and single IPv6 addresses are handled as before. IPv6 subnets remain unsupported, as on the existing file path (noted as a possible follow-up in #10763).

### Implementation

The per-line parser is moved out of `handle_command_line` into a `load_ban_list` member so the file and URL paths share one implementation. URL detection is a scheme-prefix check, and the fetch uses epee's `http_simple_client`.

### Tests

unit_tests `ban` suite:
- `parse_ban_list_string` exercises the parser on hosts, subnets, comments, blank lines, and an invalid line.
- `url_banlist` performs an end-to-end fetch over a local HTTP server and asserts the same block matrix as the existing file test.
- `url_banlist_unreachable` confirms an unreachable URL aborts startup.

Manually verified against the built daemon: `--ban-list https://gui.xmr.pm/files/block.txt` loaded 1161 entries over verified TLS, a 404 URL and a refused connection each aborted startup with a clear message, and a local file path continued to work unchanged.

### DIFF
--- a/src/p2p/net_node.cpp
+++ b/src/p2p/net_node.cpp
@@ -156,7 +156,7 @@ namespace nodetool
     const command_line::arg_descriptor<std::vector<std::string> > arg_p2p_seed_node   = {"seed-node", "Connect to a node to retrieve peer addresses, and disconnect"};
     const command_line::arg_descriptor<std::vector<std::string> > arg_tx_proxy = {"tx-proxy", "Send local txes through proxy: <network-type>,[socks5://[user:pass@]]<socks-ip:port>[,max_connections][,disable_noise] i.e. \"tor,127.0.0.1:9050,100,disable_noise\""};
     const command_line::arg_descriptor<std::vector<std::string> > arg_anonymous_inbound = {"anonymous-inbound", "<hidden-service-address>,<[bind-ip:]port>[,max_connections] i.e. \"x.onion,127.0.0.1:18083,100\""};
-    const command_line::arg_descriptor<std::string> arg_ban_list = {"ban-list", "Specify ban list file, one IP address per line"};
+    const command_line::arg_descriptor<std::string> arg_ban_list = {"ban-list", "Specify ban list file or http(s) URL, one IP address or subnet per line"};
     const command_line::arg_descriptor<bool> arg_p2p_hide_my_port   =    {"hide-my-port", "Do not announce yourself as peerlist candidate", false, true};
     const command_line::arg_descriptor<bool> arg_no_sync = {"no-sync", "Don't synchronize the blockchain with other peers", false};
     const command_line::arg_descriptor<bool> arg_enable_dns_blocklist = {"enable-dns-blocklist", "Apply realtime blocklist from DNS", false};
--- a/src/p2p/net_node.h
+++ b/src/p2p/net_node.h
@@ -284,6 +284,7 @@ namespace nodetool
     virtual bool unblock_host(const epee::net_utils::network_address &address);
     virtual bool block_subnet(const epee::net_utils::ipv4_network_subnet &subnet, time_t seconds = P2P_IP_BLOCKTIME);
     virtual bool unblock_subnet(const epee::net_utils::ipv4_network_subnet &subnet);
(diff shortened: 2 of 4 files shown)


thomasbuilds (commented):
It appears to me that when --proxy is set, this fetch goes out directly rather than through the proxy which leaks the operator's IP in the clear.

I'd also lean against this feature. If most nodes point at a single list URL, that site effectively decides who the network refuses to peer with, which is a centralization risk. Imo convenience has never really been the point with Monero, setting up and running a node already assumes the expertise to fetch such a list yourself, and updating the ban list file manually fits in the maintenance already required for updating binaries.

NoFear0411:
On the proxy leak: that is correct, and it is fixed. When `--proxy` is set, the fetch now routes through it using the daemon's existing socks client (`net::http::client::set_proxy`), and the connector sends the hostname to the proxy, so DNS resolution is remote as well. Neither the connection nor the lookup exposes the operator's address. I verified it end to end: the list loads through a local Tor socks proxy, and with the proxy unreachable the fetch fails and the node aborts rather than falling back to a direct connection.

On the feature itself: it is opt-in and operator-controlled. Nothing changes unless an operator sets the flag, and the operator chooses the URL, which can be their own server or a list they maintain. It is the same trust decision as editing the file by hand, just without the manual download and copy on every update. It does not point anyone at a particular list.

On centralization: a node already fully trusts whatever it loads via `--ban-list`; this only changes where the bytes come from, not the trust model. Convergence on a single popular URL would be each operator's own choice rather than something this change imposes, and the DNS blocklist remains the vetted network default, untouched here.

thomasbuilds:
> It is the same trust decision as editing the file by hand

A file only changes when you decide to, whereas a list served from a URL can change without your knowledge so if the host is compromised a restart silently applies whatever it serves.

There's one community-recommended ban list so the centralisation issue is real.

NoFear0411:
Spy nodes issue is also real, I just wanted to enable the daemon to load a ban list from a remote safe source directly. Up to you if you want to merge or close.

selsta:
@plowsof what's your opinion on this PR?

thomasbuilds:
https://github.com/Boog900/monero-ban-list/blob/main/ban_list.txt is the recommended ban list correct?
Last time it was updated was 10 December 2025, it doesn't seem like it requires constant updates?

NoFear0411:
> https://github.com/Boog900/monero-ban-list/blob/main/ban_list.txt is the recommended ban list correct? Last time it was updated was 10 December 2025, it doesn't seem like it requires constant updates?

I've used this list https://github.com/rblaine95/monero-banlist with combination of the list on the URL you mentioned.

plowsof:
with this PR easily replaced by a small shell script prior to launching monerod, you lose the automated rolling updates _and_ verification which the DNS quorum brings, a more complete solution would allow users to change the quorum URL's. i don't see a point in adding / reviewing lines of code for this atm

PR #10795, opened 2026-06-18
depends: boost: remove headers of unused libraries

tobtoht:
- This removes 79 out of 146 libraries.
- Boost depends cache archive: `19.2 MB -> 11.0 MB (-43%)`
- Extracted size: `164 MB -> 105 MB (-36%)`
- Saves about: `10 targets * 8.2 MB = 82 MB` of GitHub actions cache per run.

This essentially creates a whitelist of Boost libraries for release builds. It prevents us from unintentionally introducing new dependencies on Boost libraries without warning. We can always choose to add libraries to the list if desired. The point is that we'll have a record of when that happens.

It might also motivate us to reduce our dependency on Boost a bit, because I really wish this list was shorter..

I ran builds with and without this patch and compared the binaries, there are no differences.

### DIFF
--- a/contrib/depends/packages/boost.mk
+++ b/contrib/depends/packages/boost.mk
@@ -41,3 +41,78 @@ endef
 define $(package)_stage_cmds
   ./b2 -d0 -j4 --prefix=$($(package)_staging_prefix_dir) $($(package)_config_opts) install
 endef
+
+ifeq ($(release_type),release)
+define $(package)_postprocess_cmds
+  cd include/boost && \
+  find . -mindepth 1 -maxdepth 1 -type d \
+    ! -name algorithm \
+    ! -name align \
+    ! -name any \
+    ! -name archive \
+    ! -name asio \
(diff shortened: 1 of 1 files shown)


contrib/depends/packages/boost.mk
```diff
@@ -41,3 +41,78 @@ endef
 define $(package)_stage_cmds
   ./b2 -d0 -j4 --prefix=$($(package)_staging_prefix_dir) $($(package)_config_opts) install
 endef
+
+ifeq ($(release_type),release)
+define $(package)_postprocess_cmds
+  cd include/boost && \
+  find . -maxdepth 1 -type d ! -name . \
```
selsta:
```suggestion
  find . -mindepth 1 -maxdepth 1 -type d \
```

selsta (approved):

selsta (approved):

plowsof (approved):

PR #9810, opened 2025-02-21
readme: Fix subsection fragment, update info (Debian, Arch Linux, Gentoo)

taivlam:
* Update subsection to "Scheduled software/network upgrades" & corresponding page fragment
    * Previous HTML fragment was outdated, so it was broken
* Remove outdated directions for Debian Buster
    * Current [info](https://www.kicksecure.com/wiki/Monero#Installation_Sources) for installing Monero on Whonix/Kicksecure indicates there is no Debian package source method (via a third-party APT repo)
    * The [`monero`](https://tracker.debian.org/pkg/monero) CLI package should be available via APT on supported versions of Debian

PR #10798, opened 2026-06-21
readme: refresh stale info and remove dead sections

thomasbuilds:
Few nits changes and notably:

- Bump depends macOS toolchain clang-18/lld-18 -> clang-19/lld-19 and the llvm-18 PATH -> llvm-19 to match .github/workflows/depends.yml; FreeBSD depends clang-8 -> clang
- Remove dead content: Gentoo install steps (closing #9810) and the whole Coverage section
- Update blockchain size figures (full ~130 -> ~280 GB, pruned ~45 -> ~95 GB)
- Add Cake Wallet and Cypher Stack to the sponsors list to match the getmonero.org sponsorships page, reorder to match it, and size every logo by a fixed height so they render at a consistent size

Let me know for pruned chain size, not sure it's accurate

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Monero
 
-Copyright (c) 2014-2024, The Monero Project
+Copyright (c) 2014-2026, The Monero Project
 Portions Copyright (c) 2012-2013 The Cryptonote developers.
 
 ## Table of Contents
@@ -9,18 +9,18 @@ Portions Copyright (c) 2012-2013 The Cryptonote developers.
   - [Vulnerability response](#vulnerability-response)
   - [Research](#research)
   - [Announcements](#announcements)
-  - [Translations](#translations)
-  - [Coverage](#coverage)
   - [Introduction](#introduction)
   - [About this project](#about-this-project)
   - [Supporting the project](#supporting-the-project)
   - [License](#license)
   - [Contributing](#contributing)
-  - [Scheduled software upgrades](#scheduled-softwarenetwork-upgrades)
+  - [Scheduled software/network upgrades](#scheduled-softwarenetwork-upgrades)
   - [Release staging schedule and protocol](#release-staging-schedule-and-protocol)
   - [Compiling Monero from source](#compiling-monero-from-source)
     - [Dependencies](#dependencies)
     - [Guix builds](#guix-builds)
+  - [Installing Monero from a package](#installing-monero-from-a-package)
+  - [Running monerod](#running-monerod)
   - [Internationalization](#Internationalization)
   - [Using Tor](#using-tor)
   - [Pruning](#Pruning)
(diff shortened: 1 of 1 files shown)


selsta:
oss-fuzz is still maintained so IMO we should keep it

Translations are not maintained at the moment and the platform is down so it should be removed

Also I'd prefer avoiding `&nbsp;` in the Markdown. It makes the source less readable outside GitHub.

thomasbuilds:
Why is `BOOST_MIN_VER` 1.62 in CMakeLists but 1.66 in the README?

thomasbuilds:
> oss-fuzz is still maintained so IMO we should keep it

it's kind of ugly to have an entire Coverage section for just this?

I've fixed the rest

selsta:
We don't need to keep the coverage section, but it would be nice to have the oss-fuzz badge somewhere.

README.md
```diff
@@ -39,6 +39,8 @@ Portions Copyright (c) 2012-2013 The Cryptonote developers.
 
 - Our [Vulnerability Response Process](https://github.com/monero-project/meta/blob/master/VULNERABILITY_RESPONSE_PROCESS.md) encourages responsible disclosure
 - We are also available via [HackerOne](https://hackerone.com/monero)
+- Monero is continuously fuzzed by [OSS-Fuzz](https://introspector.oss-fuzz.com/project-profile?project=monero)
```
thomasbuilds, line 42: Added it here.
Could add badge to top of readme but this seems like it's not convention for monero also imo it would require to add other badges (like license, etc) to keep it good looking

selsta (approved):

selsta:
Could you remove the v0.18.4.1 changes? I will fix it separately in #10780.

Edit: Sorry for closing, pressed the wrong button.

selsta (approved):

README.md
```diff
@@ -232,7 +220,7 @@ If you already have a repo cloned, initialize and update:
 cd monero && git submodule init && git submodule update
 ```
 
-*Note*: If there are submodule differences between branches, you may need 
+*Note*: If there are submodule differences between branches, you may need
 to use `git submodule sync && git submodule update` after changing branches
```
thomasbuilds, line 224: Just noticed there's inconsistent line wrapping
we wrap here, but nowhere else?

selsta: Can be removed, don't know why it was added.

thomasbuilds: I personally would like to address this and other inconsistencies with prettier.io formatting, what do you think?

selsta: Maybe keep it for a different PR? Let's get this merged first.

thomasbuilds: ok

tobtoht (approved):

PR #10799, opened 2026-06-21
p2p: reject IPv4-mapped IPv6 peers

selsta:
Only used by spy nodes, never produced by legit nodes.


jpk68 (approved):
LGTM, but a second pair of eyes would be good as well :)

jpk68 (approved):

plowsof (approved):

PR #10800, opened 2026-06-21
cryptonote_basic: copy tx extra payment IDs

Ap4sh:
Reads payment IDs from tx extra nonce with `memcpy` instead of dereferencing `crypto::hash` / `crypto::hash8` pointers into string storage

The existing size and tag checks stay the same, so accepted bytes and failure cases do not change

The added unit coverage checks round trips plus wrong-tag, too-short, and too-long rejection

Tests:
- `cmake --build build/local-output-selection-tests --target unit_tests -j2`
- `build/local-output-selection-tests/tests/unit_tests/unit_tests --gtest_filter='cn_format_utils.*' --data-dir tests/data`
- `cmake --build build/local-output-selection-tests --target cncrypto-tests -j2`
- `cmake --build build/local-output-selection-tests --target cnv4-jit-tests -j2`
- `cd build/local-output-selection-tests/tests/crypto && ctest --output-on-failure`
- `cd build/local-output-selection-tests && ctest --output-on-failure -R '^unit_tests$'`

### DIFF
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -806,7 +806,7 @@ namespace cryptonote
       return false;
     if(TX_EXTRA_NONCE_PAYMENT_ID != extra_nonce[0])
       return false;
-    payment_id = *reinterpret_cast<const crypto::hash*>(extra_nonce.data() + 1);
+    memcpy(&payment_id, extra_nonce.data() + 1, sizeof(payment_id));
     return true;
   }
   //---------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Ap4sh:
ci note: the Ubuntu 24.04 failure looks unrelated to this patch

the failing test is `select_outputs.gamma`, with the median just under the lower bound (`111900` vs `112320`). the new `cn_format_utils` tests passed in that run, and i can reproduce the same `select_outputs.gamma` failure locally by repeating only that test

i tried rerunning the failed job, but GitHub requires repo admin rights here

jeffro256 (approved):

selsta (approved):

PR #10801, opened 2026-06-21
cryptonote: scope 202612 PoW workaround

selsta:
The block 202612 PoW workaround previously applied to any block at height 202612. Restrict it to the object hashes of the known historical mainnet, testnet, and stagenet block hashing blobs so new chains do not inherit the fixed PoW hash, which would prevent sync or mining past that height if their difficulty is higher.

Reported by @WeebDataHoarder


SChernykh:
As far as I remember, the official core team stance on Monero forks is "you're on your own", so we don't _need_ to fix this.

selsta:
This was not added for forks, the current code is just incorrect as is, for example if we start a new testnet.

src/cryptonote_core/cryptonote_tx_utils.cpp
```diff
@@ -748,6 +771,15 @@ namespace cryptonote
...
   {
+    if (height == 202612)
+    {
+      crypto::hash block_id;
+      if (!calculate_block_hash(b, block_id))
+        return false;
+      if (get_block_longhash_202612(block_id, res))
+        return true;
+    }
+
```
jeffro256: Shouldn't we move this logic to the basic overload of `get_block_longhash()`?

selsta: done

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -1608,19 +1608,39 @@ namespace cryptonote
...
+  {
+    static constexpr const char longhash_202612[] = "84f64766475d51837ac9efbef1926486e58563c95a19fef4aec3254f03000000";
+    static constexpr const char * const block_ids_202612[] =
+    {
+      "bbd604d2ba11ba27935e006ed39c9bfdd99b76bf4a50654bc1e1e61217962698", // mainnet
+      "248fde4b96b829c4ddbd00e3f76d35b03d01257898bc1b5578bc9e04b379a676", // testnet
+      "f3449e658b5f880c4b0e69007ed5d092c9c883ac3a518166fa652d5cc505e7b1", // stagenet
+    };
+
+    const std::string block_id_hex = string_tools::pod_to_hex(get_blob_hash(block_hashing_blob));
```
jeffro256:
```suggestion
    const std::string block_id_hex = string_tools::pod_to_hex(get_object_hash(block_hashing_blob));
```

Otherwise, this causes a hard fork. You should write a unit test for the positive case...

jeffro256:
> ... which would prevent sync or mining past that height if their difficulty is higher.

Also, when the difficulty is low enough, it essentially means that PoW for block 202612 is "free": the first miner to publish *any* block template at block 202612 gets the block, regardless of actual effort.

src/cryptonote_basic/cryptonote_format_utils.cpp
selsta: Sorry, updated and added positive case.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -1608,19 +1608,39 @@ namespace cryptonote
     return get_tx_tree_hash(txs_ids);
   }
   //---------------------------------------------------------------
+  static bool get_block_longhash_202612(const blobdata_ref block_hashing_blob, crypto::hash &res)
+  {
+    static constexpr const char longhash_202612[] = "84f64766475d51837ac9efbef1926486e58563c95a19fef4aec3254f03000000";
+    static constexpr const char * const block_ids_202612[] =
+    {
+      "bbd604d2ba11ba27935e006ed39c9bfdd99b76bf4a50654bc1e1e61217962698", // mainnet
```
jeffro256: Without replacing the block ID for mainnet block 202612, hashing the raw block hashing blob won't produce this block ID.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -1608,19 +1608,42 @@ namespace cryptonote
...
+      "248fde4b96b829c4ddbd00e3f76d35b03d01257898bc1b5578bc9e04b379a676", // testnet
+      "f3449e658b5f880c4b0e69007ed5d092c9c883ac3a518166fa652d5cc505e7b1", // stagenet
+    };
+
+    crypto::hash block_id;
+    const blobdata block_hashing_blob_copy(block_hashing_blob.data(), block_hashing_blob.size());
```
jeffro256, line 1614: Why the copy?

selsta: it did not compile otherwise

```
error: no matching function for call to 'do_serialize'
```

src/cryptonote_basic/cryptonote_format_utils.cpp
selsta: Changed it to hashing blob ids.

src/cryptonote_basic/cryptonote_format_utils.cpp
jeffro256: Ah I see, that's because `block_hashing_blob` is a `boost::string_ref` instead of a `std::string`. Oh well, it's not on a hot path, so it's not a big deal.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -1608,19 +1608,44 @@ namespace cryptonote
...
+    get_object_hash(block_hashing_blob_copy, block_id);
+    const std::string block_id_hex = string_tools::pod_to_hex(block_id);
+    for (const char * const expected_block_id: block_hashing_blob_ids_202612)
+    {
+      if (block_id_hex == expected_block_id)
+      {
+        epee::string_tools::hex_to_pod(longhash_202612, res);
+        return true;
```
jeffro256:
```suggestion
        return epee::string_tools::hex_to_pod(longhash_202612, res);
```

jeffro256:
I recommend adding commit https://github.com/monero-project/monero/commit/206369dd70e1628cf4f80dae9cc0ce5387bff942. It adds a block ID to the mainnet case for human verification, adds a testnet test case, adds a case to reconstruct the existing PoW hash, and adds comments.

jeffro256 (approved):

tobtoht:
needs rebase

selsta:
done

tobtoht (approved):

PR #2771, opened 2017-11-07
keccak: some paranoid "can't happen" checks

moneromooo-monero:

PR #10804, opened 2026-06-21
crypto: avoid unaligned word accesses

Ap4sh:
This avoids a few unaligned word accesses in the crypto code paths.

The main change is keeping `hash_state` and `cn_slow_hash_state` naturally aligned before they are used through their `uint64_t` members. It also switches byte-buffer word loads/stores to `memcpy` in Keccak, CryptoNight variant 1 nonce handling, and the CryptonightR JIT patcher.

This should help with the sparc64 SIGBUS reports in #10076. I do not have sparc64 access, so I validated the affected paths locally with alignment sanitizer too.

Tests:
- `cmake --build build-normal-crypto --target cncrypto-tests monero-wallet-crypto-bench unit_tests hash-tests -j2`
- `./build-normal-crypto/tests/crypto/cncrypto-tests tests/crypto/tests.txt`
- `./build-normal-crypto/tests/unit_tests/unit_tests --gtest_filter='keccak.*'`
- `./build-normal-crypto/tests/monero-wallet-crypto-bench 10`
- `ctest --test-dir build-normal-crypto -R '^hash-(fast|slow|slow-1|slow-2|slow-4|variant2-int-sqrt)$' --output-on-failure -j2`
- `cmake --build build-align-crypto --target cncrypto-tests unit_tests hash-tests -j2`
- `UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1 ./build-align-crypto/tests/crypto/cncrypto-tests tests/crypto/tests.txt`
- `UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1 ./build-align-crypto/tests/unit_tests/unit_tests --gtest_filter='keccak.*'`
- `UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1 ctest --test-dir build-align-crypto -R '^hash-(fast|slow|slow-1|slow-2|slow-4)$' --output-on-failure -j2`
- 32-bit compile check for the `hash_state` alignment assertion

### DIFF
--- a/src/crypto/hash-ops.h
+++ b/src/crypto/hash-ops.h
@@ -39,6 +39,7 @@
 
 #include <assert.h>
 #include <stdbool.h>
+#include <stdalign.h>
 #include <stddef.h>
 #include <stdint.h>
 
@@ -65,13 +66,12 @@ static inline void place_length(uint8_t *buffer, size_t bufsize, size_t length) 
 }
 POP_WARNINGS
 
-#pragma pack(push, 1)
 union hash_state {
   uint8_t b[200];
   uint64_t w[25];
 };
-#pragma pack(pop)
 static_assert(sizeof(union hash_state) == 200, "Invalid structure size");
+static_assert(alignof(union hash_state) == alignof(uint64_t), "Invalid structure alignment");
 
 void hash_permutation(union hash_state *state);
 void hash_process(union hash_state *state, const uint8_t *buf, size_t count);
--- a/src/crypto/keccak.c
+++ b/src/crypto/keccak.c
@@ -119,50 +119,49 @@ typedef uint64_t state_t[25];
 void keccak(const uint8_t *in, size_t inlen, uint8_t *md, int mdlen)
 {
     state_t st;
-    uint8_t temp[144];
+    uint64_t temp[18];
     size_t i, rsiz, rsizw;
 
-    static_assert(HASH_DATA_AREA <= sizeof(temp), "Bad keccak preconditions");
-    if (mdlen <= 0 || (mdlen >= 100 && sizeof(st) != (size_t)mdlen))
+    static_assert(HASH_DATA_AREA + sizeof(uint64_t) == sizeof(temp), "Bad keccak preconditions");
+    if (mdlen <= 0 || (mdlen >= 100 && sizeof(st) != (size_t)mdlen) || ((size_t)mdlen % sizeof(uint64_t)) != 0)
     {
       local_abort("Bad keccak use");
     }
 
     rsiz = sizeof(state_t) == mdlen ? HASH_DATA_AREA : 200 - 2 * mdlen;
     rsizw = rsiz / 8;
+    if (rsiz == 0 || rsiz > sizeof(temp) || rsizw * sizeof(uint64_t) > sizeof(temp))
+    {
+      local_abort("Bad keccak use");
+    }
     
     memset(st, 0, sizeof(st));
 
     for ( ; inlen >= rsiz; inlen -= rsiz, in += rsiz) {
+      memcpy(temp, in, rsiz);
       for (i = 0; i < rsizw; i++) {
-        uint64_t ina;
-        memcpy(&ina, in + i * 8, 8);
-        st[i] ^= swap64le(ina);
+        st[i] ^= SWAP64LE(temp[i]);
       }
       keccakf(st, KECCAK_ROUNDS);
     }
     
     // last block and padding
-    if (inlen + 1 >= sizeof(temp) || inlen > rsiz || rsiz - inlen + inlen + 1 >= sizeof(temp) || rsiz == 0 || rsiz - 1 >= sizeof(temp) || rsizw * 8 > sizeof(temp))
+    memset(temp, 0, sizeof(temp));
+    if (inlen >= rsiz)
     {
       local_abort("Bad keccak use");
     }
-
     if (inlen > 0)
       memcpy(temp, in, inlen);
-    temp[inlen++] = 1;
-    memset(temp + inlen, 0, rsiz - inlen);
-    temp[rsiz - 1] |= 0x80;
+    ((uint8_t *) temp)[inlen] = 1;
+    ((uint8_t *) temp)[rsiz - 1] |= 0x80;
 
-    for (i = 0; i < rsizw; i++)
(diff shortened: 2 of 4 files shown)


src/crypto/hash-ops.h
```diff
@@ -65,14 +65,22 @@ static inline void place_length(uint8_t *buffer, size_t bufsize, size_t length)
...
+struct hash_state_alignment_check {
+  uint8_t prefix;
+  union hash_state state;
+};
+struct uint64_t_alignment_check {
+  uint8_t prefix;
+  uint64_t value;
+};
+static_assert(offsetof(struct hash_state_alignment_check, state) == offsetof(struct uint64_t_alignment_check, value), "Invalid structure alignment");
+
```
jeffro256: Why not check `alignof` directly?

Ap4sh: yep, that's cleaner, updated it to use the C11 `alignof` macro from `<stdalign.h>` directly since this part of the header is C code, and re-ran the crypto/hash tests too

selsta:
please squash

jeffro256:
This is what I recommend for the Keccak diff instead:

```diff
diff --git a/src/crypto/keccak.c b/src/crypto/keccak.c
index 6616d3530..153e10b6e 100644
--- a/src/crypto/keccak.c
+++ b/src/crypto/keccak.c
@@ -119,11 +119,11 @@ typedef uint64_t state_t[25];
 void keccak(const uint8_t *in, size_t inlen, uint8_t *md, int mdlen)
 {
     state_t st;
-    uint8_t temp[144];
+    uint64_t temp[18];
     size_t i, rsiz, rsizw;
 
-    static_assert(HASH_DATA_AREA <= sizeof(temp), "Bad keccak preconditions");
-    if (mdlen <= 0 || (mdlen >= 100 && sizeof(st) != (size_t)mdlen))
+    static_assert(HASH_DATA_AREA + sizeof(uint64_t) == sizeof(temp), "Bad keccak preconditions");
+    if (mdlen <= 0 || (mdlen >= 100 && sizeof(st) != (size_t)mdlen) || ((size_t)mdlen % sizeof(uint64_t) != 0))
     {
       local_abort("Bad keccak use");
     }
@@ -134,10 +134,9 @@ void keccak(const uint8_t *in, size_t inlen, uint8_t *md, int mdlen)
     memset(st, 0, sizeof(st));
 
     for ( ; inlen >= rsiz; inlen -= rsiz, in += rsiz) {
+      memcpy(temp, in, rsiz);
       for (i = 0; i < rsizw; i++) {
-        uint64_t ina;
-        memcpy(&ina, in + i * 8, 8);
-        st[i] ^= swap64le(ina);
+        st[i] ^= SWAP64LE(temp[i]);
       }
       keccakf(st, KECCAK_ROUNDS);
     }
@@ -148,21 +147,18 @@ void keccak(const uint8_t *in, size_t inlen, uint8_t *md, int mdlen)
       local_abort("Bad keccak use");
     }
 
+    for (i = temp[inlen / sizeof(uint64_t)] = 0; i < sizeof(temp)/sizeof(temp[0]); ++i)
+      temp[i] = 0;
     if (inlen > 0)
       memcpy(temp, in, inlen);
-    temp[inlen++] = 1;
-    memset(temp + inlen, 0, rsiz - inlen);
-    temp[rsiz - 1] |= 0x80;
+    ((char*) temp)[inlen] = 1;
+    ((char*) temp)[rsiz - 1] |= 0x80;
 
     for (i = 0; i < rsizw; i++)
-        st[i] ^= swap64le(((uint64_t *) temp)[i]);
+      st[i] ^= SWAP64LE(temp[i]);
 
     keccakf(st, KECCAK_ROUNDS);
 
-    if (((size_t)mdlen % sizeof(uint64_t)) != 0)
-    {
-      local_abort("Bad keccak use");
-    }
     memcpy_swap64le(md, st, mdlen/sizeof(uint64_t));
 }
 ```

It allows `temp` to be `uint64_t` to retain word performance, and the word reads are always strictly aliased and aligned. This also reduces the number of calls to `memcpy` in the main loop.

**Comment by Ap4sh** [comment]
pushed the aligned `uint64_t temp[18]` version and squashed the branch too

i kept the early `rsiz > sizeof(temp)` guard before the full-block copy since `temp` is fixed-size now, and local crypto/hash/keccak checks still pass

File: src/crypto/keccak.c [hunk 2]
```diff
@@ -119,50 +119,50 @@ typedef uint64_t state_t[25];
...
-        memcpy(&ina, in + i * 8, 8);
-        st[i] ^= swap64le(ina);
+        st[i] ^= SWAP64LE(temp[i]);
       }
       keccakf(st, KECCAK_ROUNDS);
     }
     
     // last block and padding
-    if (inlen + 1 >= sizeof(temp) || inlen > rsiz || rsiz - inlen + inlen + 1 >= sizeof(temp) || rsiz == 0 || rsiz - 1 >= sizeof(temp) || rsizw * 8 > sizeof(temp))
+    if (inlen + 1 >= sizeof(temp) || inlen > rsiz || rsiz - inlen + inlen + 1 >= sizeof(temp) || rsiz - 1 >= sizeof(temp))
```
selsta: This seems to be a leftover defensive check from the old padding code and could be removed. The new code zeroes the whole `temp` buffer first, and `rsiz != 0` / `rsiz <= sizeof(temp)` are already checked above. If you want to keep a defensive check here, it could be simplified to `if (inlen >= rsiz)`.

Context: https://github.com/monero-project/monero/pull/2771

Ap4sh: updated, removed the leftover check entirely

the earlier `rsiz` guard plus the absorb-loop condition already cover the padding indexes here, so keeping another defensive check would just be redundant

selsta (approved):

src/crypto/keccak.c
jeffro256: I agree with @selsta, the `if (inlen >= rsiz)` should go back in. All defensive checks are redundant with the right initial assumptions and assuming no one else pokes at the code. That's why it's defensive. But if someone else modifies the previous for loop to break early, for example, then the check is no longer redundant.

Ap4sh: yep, that makes sense, restored the simplified `inlen >= rsiz` guard and re-ran the crypto and hash tests

selsta (approved):

jeffro256 (approved):

Issue #9009, opened 2023-09-28
mms auto_config bitmessage json has binary


lacdael:
mms auto_config is failing because the JSON sent via bitmessage JSON can't deserialize 

```
$ ./monero-wallet-cli --version
Monero 'Fluorine Fermi' (v0.18.2.2-release)
```

```
2023-09-28 22:56:17.590     7f8c8ffb76c0        ERROR   default contrib/epee/include/storages/portable_storage_from_json.h:88   Wrong JSON character at: BM-2cT3rK2mqxVNbWAmk5MdU2Wz28TsvoYggC
2023-09-28 22:56:17.590     7f8c8ffb76c0        ERROR   default contrib/epee/include/storages/portable_storage_from_json.h:403  Failed to parse json, what: Wrong JSON character at: BM-2cT3rK2mqxVNbWAmk5MdU2Wz28TsvoYggC
2023-09-28 22:56:17.590     7f8c8ffb76c0        ERROR   wallet.mms      src/wallet/message_transporter.cpp:124  Failed to deserialize messages
```

from testnet wallets:

```
{
  "content": "g,'8*$BR\"mIUW_vqF#e:ߧ_ ofe?H\bcdKTikD`V@su",
  "destination_monero_address": {
    "m_spend_public_key": "��������������������������������",
    "m_view_public_key": "��������������������������������"
  },
  "destination_transport_address": "BM-2cSrgeeXv5oEvBLMzTaX7zmvVGJV1XanWz",
  "encryption_public_key": "fK\/6dG=>�*\v)g)2",
  "hash": "q).U\"HSUv&lQ",
  "iv": "9eMi",
  "round": 0,
  "signature": ")ez'8=&LITx\tt\vc\t8\v}ܹ+򇅚.Yoz\n",
  "signature_count": 0,
  "source_monero_address": {
    "m_spend_public_key": "\v#pݕLZSzҙ!BI",
    "m_view_public_key": "r-o@�7EfKaU,+{>"
  },
  "source_transport_address": "BM-2cXibCmyMyGBsCt1RN5DBsiUkWWpr1FtnV",
  "subject": "MMS V0 2023-09-28 22:31:59",
  "timestamp": 1695940319,
  "transport_id": "",
  "type": 7
}
```

selsta:
@rbrunner7 any idea?

rbrunner7:
[...]

PR #10808, opened 2026-06-22
escape control characters in json string serialiser

alhudz:
`transform_to_escape_sequence` builds the string body emitted by `dump_as_json`/`store_t_to_json`.

1. control bytes `0x00`-`0x07` and `0x0e`-`0x1f` are emitted verbatim, which RFC 8259 forbids inside a JSON string, so one such byte makes the response invalid JSON and an embedded `0x00` truncates C-string based parsers.
2. `0x0b` is emitted as `\v`, which is not a JSON escape; the parser reads it back as the letter `v`.

repro: serialise a value holding `0x01` or `0x0b`.
expected: both escaped as a six-character `\u00xx` sequence.
actual: `0x01` is copied out raw, `0x0b` becomes `\v`.
fix: escape every byte below `0x20`, the short escape where JSON defines one and `\u00xx` otherwise. reachable for any string value carrying attacker-influenced bytes (e.g. one decoded by `convert_from_url_format`). test added.

### DIFF
--- a/contrib/epee/src/parserse_base_utils.cpp
+++ b/contrib/epee/src/parserse_base_utils.cpp
@@ -41,8 +41,14 @@ namespace misc_utils
   {
     std::string transform_to_escape_sequence(const std::string& src)
     {
-      static const char escaped[] = "\b\f\n\r\t\v\"\\/";
-      std::string::const_iterator it = std::find_first_of(src.begin(), src.end(), escaped, escaped + sizeof(escaped));
+      // RFC 8259: the double quote, reverse solidus and every control character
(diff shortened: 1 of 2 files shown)


alhudz:
gentle ping

vtnerd (approved):

selsta:
@alhudz please rebase

alhudz:
rebased. only conflict was the comment spelling fix from 02db831d, kept the `//Slash character` wording from this branch.

selsta (approved):

jpk68:
@selsta This closes #9009, yes?

selsta:
Not exactly but it can still be closed.

PR #10819, opened 2026-06-25
wallet_rpc: replace wallet2 with Wallet API

SNeedlewoods:
## Context

This change was proposed by @j-berman in Step 2 [here](https://github.com/seraphis-migration/wallet3/issues/64).
(And the first part, replacing `wallet2` with the Wallet API in `monero-wallet-cli`, is waiting for reviews/testers here: #10233)

All the relevant changes for this PR can be found in the fifth commit.
Other commits have their own PRs:
1-3: add functions to Wallet API - #9464 (where the first commit was the initial round, before working on wallet-cli or wallet-rpc, second commit were changes that came up during work on the wallet-cli #10233, third commit started out as all changes that were explicitly introduced during work on the wallet-rpc, but all the recent changes (even for wallet-cli fixes) have been squashed into the latest commit, so maybe at this point I can just squash all three commits of that PR?)
4: remove cached password in Wallet API - #10232

(Will eventually make some edits to this PR description, but for now I didn't want to further delay this PR.)

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -46,13 +46,13 @@
 #include "cryptonote_basic/cryptonote_format_utils.h"
 #include "cryptonote_basic/account.h"
 #include "multisig/multisig.h"
+#include "net/parse.h"
 #include "wallet_rpc_server_commands_defs.h"
 #include "string_coding.h"
 #include "string_tools.h"
 #include "crypto/hash.h"
 #include "mnemonics/electrum-words.h"
 #include "rpc/rpc_args.h"
-#include "rpc/core_rpc_server_commands_defs.h"
 #include "daemonizer/daemonizer.h"
 #include "fee_priority.h"
 
@@ -65,7 +65,7 @@
 #define CHECK_MULTISIG_ENABLED() \
   do \
   { \
-    if (m_wallet->get_multisig_status().multisig_is_active && !m_wallet->is_multisig_enabled()) \
+    if (m_wallet_impl->multisig().isMultisig && !m_wallet_impl->getEnableMultisig()) \
     { \
       er.code = WALLET_RPC_ERROR_CODE_DISABLED; \
       er.message = "This wallet is multisig, and multisig is disabled. Multisig is an experimental feature and may have bugs. Things that could go wrong include: funds sent to a multisig wallet can't be spent at all, can only be spent with the participation of a malicious group member, or can be stolen by a malicious group member. You can enable it by running this once in monero-wallet-cli: set enable-multisig-experimental 1"; \
@@ -88,14 +88,14 @@
       er.message = "Command unavailable in restricted mode."; \
       return false; \
     } \
-    if (!m_wallet) { return not_open(er); } \
-    if (m_wallet->is_background_wallet()) \
+    if (!m_wallet_impl) { return not_open(er); } \
+    if (m_wallet_impl->isBackgroundWallet()) \
     { \
       er.code = WALLET_RPC_ERROR_CODE_IS_BACKGROUND_WALLET; \
       er.message = "This command is disabled for background wallets."; \
       return false; \
     } \
-    if (m_wallet->is_background_syncing()) \
+    if (m_wallet_impl->isBackgroundSyncing()) \
     { \
       er.code = WALLET_RPC_ERROR_CODE_IS_BACKGROUND_SYNCING; \
       er.message = "This command is disabled while background syncing. Stop background syncing to use this command."; \
@@ -112,20 +112,20 @@
       er.message = "Command unavailable in restricted mode."; \
       return false; \
     } \
-    if (!m_wallet) { return not_open(er); } \
-    if (m_wallet->key_on_device()) \
+    if (!m_wallet_impl) { return not_open(er); } \
+    if (m_wallet_impl->getDeviceState().key_on_device) \
     { \
       er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR; \
       er.message = "Command not supported by HW wallet"; \
       return false; \
     } \
-    if (m_wallet->get_multisig_status().multisig_is_active) \
+    if (m_wallet_impl->multisig().isMultisig) \
     { \
       er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR; \
       er.message = "Multisig wallet cannot enable background sync"; \
       return false; \
     } \
-    if (m_wallet->watch_only()) \
+    if (m_wallet_impl->watchOnly()) \
     { \
       er.code = WALLET_RPC_ERROR_CODE_WATCH_ONLY; \
       er.message = "Watch-only wallet cannot enable background sync"; \
@@ -147,6 +147,32 @@ namespace
   const command_line::arg_descriptor<std::size_t> arg_rpc_max_connections_per_private_ip = {"rpc-max-connections-per-private-ip", "Max RPC connections per private and localhost IP permitted", DEFAULT_RPC_MAX_CONNECTIONS_PER_PRIVATE_IP};
   const command_line::arg_descriptor<std::size_t> arg_rpc_max_connections = {"rpc-max-connections", "Max RPC connections permitted", DEFAULT_RPC_MAX_CONNECTIONS};
   const command_line::arg_descriptor<std::size_t> arg_rpc_response_soft_limit = {"rpc-response-soft-limit", "Max response bytes that can be queued, enforced at next response attempt", DEFAULT_RPC_SOFT_LIMIT_SIZE};
+  const command_line::arg_descriptor<std::string> arg_daemon_address = {"daemon-address", tr("Use daemon instance at <host>:<port>"), ""};
+  const command_line::arg_descriptor<std::string> arg_daemon_host = {"daemon-host", tr("Use daemon instance at host <arg> instead of localhost"), ""};
+  const command_line::arg_descriptor<std::string> arg_proxy = {"proxy", tr("[<ip>:]<port> socks proxy to use for daemon connections"), {}, true};
+  const command_line::arg_descriptor<bool> arg_trusted_daemon = {"trusted-daemon", tr("Enable commands which rely on a trusted daemon"), false};
+  const command_line::arg_descriptor<bool> arg_untrusted_daemon = {"untrusted-daemon", tr("Disable commands which rely on a trusted daemon"), false};
+  const command_line::arg_descriptor<std::string> arg_password = {"password", tr("Wallet password (escape/quote as needed)"), "", true};
+  const command_line::arg_descriptor<std::string> arg_password_file = wallet_args::arg_password_file();
+  const command_line::arg_descriptor<int> arg_daemon_port = {"daemon-port", tr("Use daemon instance at port <arg> instead of 18081"), 0};
+  const command_line::arg_descriptor<std::string> arg_daemon_login = {"daemon-login", tr("Specify username[:password] for daemon RPC client"), "", true};
+  const command_line::arg_descriptor<std::string> arg_daemon_ssl = {"daemon-ssl", tr("Enable SSL on daemon RPC connections: enabled|disabled|autodetect"), "autodetect"};
+  const command_line::arg_descriptor<std::string> arg_daemon_ssl_private_key = {"daemon-ssl-private-key", tr("Path to a PEM format private key"), ""};
+  const command_line::arg_descriptor<std::string> arg_daemon_ssl_certificate = {"daemon-ssl-certificate", tr("Path to a PEM format certificate"), ""};
+  const command_line::arg_descriptor<std::string> arg_daemon_ssl_ca_certificates = {"daemon-ssl-ca-certificates", tr("Path to file containing concatenated PEM format certificate(s) to replace system CA(s).")};
+  const command_line::arg_descriptor<std::vector<std::string>> arg_daemon_ssl_allowed_fingerprints = {"daemon-ssl-allowed-fingerprints", tr("List of valid fingerprints of allowed RPC servers")};
+  const command_line::arg_descriptor<bool> arg_daemon_ssl_allow_any_cert = {"daemon-ssl-allow-any-cert", tr("Allow any SSL certificate from the daemon"), false};
+  const command_line::arg_descriptor<bool> arg_daemon_ssl_allow_chained = {"daemon-ssl-allow-chained", tr("Allow user (via --daemon-ssl-ca-certificates) chain certificates"), false};
+  const command_line::arg_descriptor<bool> arg_allow_mismatched_daemon_version = {"allow-mismatched-daemon-version", tr("Allow communicating with a daemon that uses a different version"), false};
+  const command_line::arg_descriptor<bool> arg_testnet = {"testnet", tr("For testnet. Daemon must also be launched with --testnet flag"), false};
+  const command_line::arg_descriptor<bool> arg_stagenet = {"stagenet", tr("For stagenet. Daemon must also be launched with --stagenet flag"), false};
+  const command_line::arg_descriptor<uint64_t> arg_kdf_rounds = {"kdf-rounds", tr("Number of rounds for the key derivation function"), 1};
+  const command_line::arg_descriptor<std::string> arg_hw_device = {"hw-device", tr("HW device to use"), ""};
+  const command_line::arg_descriptor<std::string> arg_hw_device_derivation_path = {"hw-device-deriv-path", tr("HW device wallet derivation path (e.g., SLIP-10)"), ""};
+  const command_line::arg_descriptor<std::string> arg_tx_notify = { "tx-notify" , tr("Run a program for each new incoming transaction, '%s' will be replaced by the transaction hash") , "" };
+  const command_line::arg_descriptor<bool> arg_offline = {"offline", tr("Do not connect to a daemon, nor use DNS"), false};
+  const command_line::arg_descriptor<std::string> arg_extra_entropy = {"extra-entropy", tr("File containing extra entropy to initialize the PRNG (any data, aim for 256 bits of entropy to be useful, which typically means more than 256 bits of data)")};
+
 
   constexpr const char default_rpc_username[] = "monero";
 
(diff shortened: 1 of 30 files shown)


src/wallet/wallet_rpc_server.h
```diff
@@ -191,9 +194,10 @@ namespace tools
...
       bool on_submit_transfer(const wallet_rpc::COMMAND_RPC_SUBMIT_TRANSFER::request& req, wallet_rpc::COMMAND_RPC_SUBMIT_TRANSFER::response& res, epee::json_rpc::error& er, const connection_context *ctx = NULL);
-      bool on_sweep_dust(const wallet_rpc::COMMAND_RPC_SWEEP_DUST::request& req, wallet_rpc::COMMAND_RPC_SWEEP_DUST::response& res, epee::json_rpc::error& er, const connection_context *ctx = NULL);
+      // removed: sweep_dust/sweep_unmixable; reason: not needed anymore after FCMP++ fork, discussed in "Monero Tech Meeting #140" https://github.com/monero-project/meta/issues/1277
       bool on_sweep_all(const wallet_rpc::COMMAND_RPC_SWEEP_ALL::request& req, wallet_rpc::COMMAND_RPC_SWEEP_ALL::response& res, epee::json_rpc::error& er, const connection_context *ctx = NULL);
       bool on_sweep_single(const wallet_rpc::COMMAND_RPC_SWEEP_SINGLE::request& req, wallet_rpc::COMMAND_RPC_SWEEP_SINGLE::response& res, epee::json_rpc::error& er, const connection_context *ctx = NULL);
+
```
rbrunner7: Looks like this new empty line slipped in ...

src/wallet/wallet_rpc_server.cpp
```diff
@@ -115,27 +115,93 @@ using namespace epee;
...
+            tools::error::throw_wallet_ex<tools::error::nonzero_unlock_time>(std::string(__FILE__ ":" STRINGIZE(__LINE__))); \
+            break; \
+          default: \
+            MERROR("unexpected extended_error_code"); \
+            tools::error::throw_wallet_ex<tools::error::wallet_internal_error>(std::string(__FILE__ ":" STRINGIZE(__LINE__)), error_message); \
+            break; \
+        } \
+      } \
+    } while (0)
+
```
rbrunner7, line 137: This approach somehow looks unfortunate:

You have the Wallet API which generates error codes. Then you translate those to exceptions. Then you catch these exceptions and translate them to error codes, a process that throws them *again* and catches them *again*,  to give the error codes finally back as result of RPC calls.

This sure avoids special cases where code does "naturally" throw exceptions that you have to catch to deal with them, but I still ask myself if a simple method (not even a macro) that translates Wallet API error codes to RPC call result codes would somehow be possible, and you would replace references to the macro with calls to this code translation method.

Furthermore, this macro looks like about 50 lines of code long, and you call it about 50 times or so, so this will expand to roughly 2500 lines of code to compile. A simple code translation method on the other hand, if it's feasible, would do this in 50 lines *once*.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -185,6 +277,284 @@ namespace
...
+      daemon_username = std::move(parsed->username);
+      epee::wipeable_string daemon_pw = std::move(parsed->password).password();
+      daemon_password = std::string(daemon_pw.data(), daemon_pw.size());
+    }
+
+    return true;
+  }
+  //------------------------------------------------------------------------------------------------------------------------------
+  bool get_daemon_init_settings(const boost::program_options::variables_map &vm,
```
rbrunner7, line 319: Not sure whether it would really be worth it, but it might be an idea that if you took code from somewhere to include it directly in the RPC server code now (with some refactoring of course), to give some info where this code originally lived, i.e. in this case the `wallet2` method `make_basic`.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -185,6 +277,284 @@ namespace
...
+    std::pair<std::string, std::string> daemon_login = std::make_pair(daemon_username, daemon_password);
+    wallet_manager->setDaemonAddress(daemon_address, daemon_username.empty() ? nullptr : &daemon_login);
+    wal->init(daemon_address,
+              /* upper_transaction_size_limit */ 0,
+              daemon_username,
+              daemon_password,
+              /* use_ssl */ false,
+              /* lightWallet */ false);
+
+    return wal->setDaemon(daemon_address,
```
rbrunner7, line 482: It looks a bit strange that `daemon_username` is used as a parameter of no less than 3 different calls, but I guess that's ok, because all those objects and methods handing it to each other so that giving it only once would be enough would probably be complicated.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -195,20 +565,19 @@ namespace tools
...
-  wallet_rpc_server::wallet_rpc_server():m_wallet(NULL), rpc_login_file(), m_stop(false), m_restricted(false), m_vm(NULL)
+  wallet_rpc_server::wallet_rpc_server():rpc_login_file(), m_stop(false), m_restricted(false), m_vm(NULL)
   {
   }
   //------------------------------------------------------------------------------------------------------------------------------
   wallet_rpc_server::~wallet_rpc_server()
   {
-    if (m_wallet)
-      delete m_wallet;
   }
```
rbrunner7: Is a zero-statement destructor still useful, or even needed because of some complicated C++ reason?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -783,7 +1100,9 @@ namespace tools
     CHECK_IF_RESTRICTED_BACKGROUND_SYNCING();
     try
     {
-      m_wallet->set_subaddress_label(req.index, req.label);
+      THROW_WALLET_EXCEPTION_IF(req.index.major > m_wallet_impl->numSubaddressAccounts(), error::account_index_outofbound);
```
rbrunner7, line 1031: I am not sure it's a good idea to check here what `setSubaddressLabel` will check anyway. And well, right now that method does not check more than that, so maybe it's ok to trust it to return no error for sure and not check any returned status, but that looks brittle to me, as that assumption could break in the future, at least in theory.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -795,31 +1114,31 @@ namespace tools
...
         wallet_rpc::COMMAND_RPC_GET_ACCOUNTS::subaddress_account_info info;
         info.account_index = subaddr_index.major;
-        info.base_address = m_wallet->get_subaddress_as_str(subaddr_index);
-        info.balance = m_wallet->balance(subaddr_index.major, req.strict_balances);
-        info.unlocked_balance = m_wallet->unlocked_balance(subaddr_index.major, req.strict_balances);
-        info.label = m_wallet->get_subaddress_label(subaddr_index);
+        info.base_address = m_wallet_impl->address(subaddr_index.major, subaddr_index.minor);
+        info.balance = m_wallet_impl->balance(subaddr_index.major, req.strict_balances);
+        info.unlocked_balance = m_wallet_impl->unlockedBalance(subaddr_index.major, /* blocks_to_unlock */ NULL, /* time_to_unlock */ NULL, req.strict_balances);
+        info.label = m_wallet_impl->getSubaddressLabel(subaddr_index.major, subaddr_index.minor);
```
rbrunner7, line 1069: I just realized that we have here 4 Wallet API calls in a row where we just trust that they won't return any error, and more or less have to do so, because checking the returned status after *each* such call looks infeasible.

Looks like a disadvantage of the Wallet API approach of returning status codes we just have to live with.

rbrunner7 (commented):
I am not yet through with my review by far, but I decided to submit what I already have, so you know that I am working on this, and what I found so far.

Quite in general, I found reviewing this to be quite difficult, because the code changes are extensive, but I think there was no way around that, given the circumstances, so not a complaint, just a comment.

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: Good points, will look into this again.

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: Hope this is okay (also added to `get_daemon_address()` and `init_wallet()`):
```
  // handle command line arguments, similar to make_basic() in wallet2.cpp
```
else I'm open to suggestions.

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: AFAICT we unfortunately need all of them.
First one is to set the daemon for the `wallet_manager` (for methods like `startMining()`).
Without the second one `WalletImpl::init()`->`WalletImpl::doInit()` will fail here: https://github.com/monero-project/monero/blob/0f84863feda75574788e15c9691abecf57bc3f8f/src/wallet/api/wallet.cpp#L2512 
Though under the hood it also just calls `wallet2::set_daemon()` https://github.com/monero-project/monero/blob/0f84863feda75574788e15c9691abecf57bc3f8f/src/wallet/wallet2.cpp#L1426 but we lose control over SSL and proxy options.
That's why the last one is needed.

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: I don't think so, just an oversight from me to not remove it.

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: I get your point. The reason why I did it this way was to throw the correct exception directly `error::[account/address]_index_outofbound`.
Though not sure why I didn't use the `extended_error_code` approach here. Will keep in mind to look into this also when looking into the the macro you commented on here https://github.com/monero-project/monero/pull/10819#discussion_r3566126571

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: `address()`, `balance()` and `unlockedBalance()` currently all do not catch any errors, e.g. see here
https://github.com/monero-project/monero/blob/0f84863feda75574788e15c9691abecf57bc3f8f/src/wallet/api/wallet.cpp#L885-L888
`getSubaddressLabel()` does only error if called on a background wallet AFAICT. (I'd need to check to make sure, but I think background wallets are not supported in the wallet-rpc.)

_____

Though if you dig deep enough, you see that in theory `address()` can throw in `rctOps`, here:
https://github.com/monero-project/monero/blob/0f84863feda75574788e15c9691abecf57bc3f8f/src/ringct/rctOps.cpp#L423
(I did not trace every possible branch, stopped at the first one that can throw)

- `WalletImpl::address()`
  - -> `wallet2::get_subaddress_as_str()`
    - -> `wallet2::get_subaddress()`
      - -> `device_default::get_subaddress()`
        - -> `scalarmultKey()`

src/wallet/wallet_rpc_server.cpp
```diff
@@ -1261,57 +1556,61 @@ namespace tools
...
       }
-
-      return fill_response(ptx_vector, req.get_tx_key, res.tx_key, res.amount, res.amounts_by_dest, res.fee, res.weight, res.multisig_txset, res.unsigned_txset, req.do_not_relay,
+      return fill_response(std::move(ptx), req.get_tx_key, res.tx_key, res.amount, res.amounts_by_dest, res.fee, res.weight, res.multisig_txset, res.unsigned_txset, req.do_not_relay,
           res.tx_hash, req.get_tx_hex, res.tx_blob, req.get_tx_metadata, res.tx_metadata, res.spent_key_images, er);
     }
     catch (const std::exception& e)
     {
-      handle_rpc_exception(std::current_exception(), er, WALLET_RPC_ERROR_CODE_GENERIC_TRANSFER_ERROR);
+      handle_rpc_exception(std::current_exception(), er, WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR);
```
rbrunner7: This was `WALLET_RPC_ERROR_CODE_GENERIC_TRANSFER_ERROR` before. Any particular reason for the switch to "unknown"?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -1805,43 +1960,48 @@ namespace tools
...
-      return false;
+    std::unique_ptr<PendingTransaction> ptx(
+            m_wallet_impl->createTransactionMultDest({dst_addr},
+                                                     /* [deprecated] payment_id */ "",
+                                                     /* amount */ Monero::optional<std::vector<uint64_t>>(),
+                                                     req.ring_size ? req.ring_size - 1 : 0,
+                                                     static_cast<PendingTransaction::Priority>(req.priority),
+                                                     req.account_index,
+                                                     subaddr_indices,
+                                                     /* subtract_fee_from_outputs */ {},
```
rbrunner7, line 1804: If I understand sweeping correctly, the fees to do it have to come from some enotes that are in the wallet. If that's correct, how is it done here?

I don't fully understand how it would work, but I find this empty parameter here a bit suspicious ...

src/wallet/wallet_rpc_server.cpp
```diff
@@ -1944,68 +2094,60 @@ namespace tools
...
+      std::string payment_id_str;
       if (req.payment_id.empty())
       {
         payment_id = crypto::rand<crypto::hash8>();
+        payment_id_str = epee::string_tools::pod_to_hex(payment_id);
       }
       else
       {
-        if (!tools::wallet2::parse_short_payment_id(req.payment_id,payment_id))
+        if (!tools::wallet2::parse_short_payment_id(req.payment_id, payment_id))
```
rbrunner7: I wonder whether it would be worth it to have copies of those 6 or 7 static `wallet2` methods, like e.g. `parse_short_payment_id` used here, in some file that lives in the Wallet API folder, or in some "util" file in the base code somewhere.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -2091,48 +2234,57 @@ namespace tools
...
+      er.code = WALLET_RPC_ERROR_CODE_WRONG_PAYMENT_ID;
+      er.message = "Payment ID has invalid format: " + req.payment_id;
+      return false;
+    }
 
     res.payments.clear();
-    std::list<wallet2::payment_details> payment_list;
-    m_wallet->get_payments(payment_id, payment_list);
-    for (auto & payment : payment_list)
+    const std::vector<TransactionInfo *> &txs = m_wallet_impl->history()->getAll(/* do_refresh */ true);
```
rbrunner7: Does this a refresh as a "side effect"?

I can't see a refresh in the old code, and I am not sure it's a good idea to add a refresh now. Maybe, maybe not. I imagine that RPC API users might like to keep control over when the refresh happens, and that's no longer possible if you refresh on your own like here.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -2142,80 +2294,84 @@ namespace tools
...
+        rpc_payment.address      = m_wallet_impl->address(subaddress_index.major, subaddress_index.minor);
+        rpc_payment.locked       = !tx->isUnlocked();
         res.payments.push_back(std::move(rpc_payment));
       }
 
       return true;
     }
 
-    for (auto & payment_id_str : req.payment_ids)
+    std::string payment_id_str = "";
```
rbrunner7: You seem to rewrite about 20 lines of payment ID processing and checking code here. I could not see a technical reason that forced you to do this. Is this the result of a harmonization with other code?

I just ask because seems to me in many cases the best strategy is to leave code as-is if it's not really bad and nothing forces any changes.

rbrunner7 (commented):
I could continue with the review today, which "produced" 5 new comments.

Still far from completing however.

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: It's my mistake.
At one point I had the `try/catch` blocks removed, this sneaked in later when I added them back.
Looking into this now showed me that I missed to re-add the `try/catch` blocks to at least the following methods (which are on my TODO list now for the next push):
- `on_transfer_split()`
- `on_sign_transfer()`
- `on_submit_transfer()`
- `on_sweep_all()`
- `on_sweep_single()`

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: The argument comes from the feature added here https://github.com/monero-project/monero/pull/8861
This is how the current Wallet API code on `master` does call `sweep*()` methods (called `create_transactions_all()` in `wallet2`): 
https://github.com/monero-project/monero/blob/95e126f3c5b18950a3071c1b2c34a7dbba207cfe/src/wallet/api/wallet.cpp#L1708-L1710
Here you can see that `create_transactions_all()` does not even have the `subtract_fee_from_outputs` argument:
https://github.com/monero-project/monero/blob/95e126f3c5b18950a3071c1b2c34a7dbba207cfe/src/wallet/wallet2.h#L908-L909

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: This "refresh" is referring to the `TransactionHistory` object, not the the `wallet2::refresh()`, so I don't think this has side effects.
https://github.com/monero-project/monero/blob/95e126f3c5b18950a3071c1b2c34a7dbba207cfe/src/wallet/api/transaction_history.cpp#L106

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: I think the following is sufficient in this case.
``` suggestion
        if (!(req.payment_id.size() == 2 * sizeof(payment_id)
            && Wallet::paymentIdValid(req.payment_id)
            && epee::string_tools::hex_to_pod(req.payment_id, payment_id)))
```
A while ago someone suggested that Wallet API users should only be offered what is really needed (which IMO is done with `paymenIdValid()`¹). In case if there were more places that needed `parse_short_payment_id()` we would out-factor that snippet into something like
```
static bool wallet_rpc_server::parse_short_payment_id(const string &payment_id_str, crypto::hash8 &payment_id_out)
```
(¹ even though the parameter name is a _payn_ to look at, jk)
https://github.com/monero-project/monero/blob/25baaafad81a6e202a0cb5ee6df0b2c9b8a6ec64/src/wallet/api/wallet.cpp#L335

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: Tried to clean this up, also matched the old (IMO non-optimal) behavior. i.e.: when calling `get_bulk_payments` with an empty list of payment ids in the request, it will return long payment ids for all payments (short payment ids get suffixed by 48 zeros, converting them to "pseudo" long payment ids).

src/wallet/wallet_rpc_server.h
```diff
@@ -266,23 +268,23 @@ namespace tools
...
       // helpers
-      void fill_transfer_entry(tools::wallet_rpc::transfer_entry &entry, const crypto::hash &txid, const crypto::hash &payment_id, const tools::wallet2::payment_details &pd) const;
-      void fill_transfer_entry(tools::wallet_rpc::transfer_entry &entry, const crypto::hash &txid, const tools::wallet2::confirmed_transfer_details &pd) const;
-      void fill_transfer_entry(tools::wallet_rpc::transfer_entry &entry, const crypto::hash &txid, const tools::wallet2::unconfirmed_transfer_details &pd) const;
-      void fill_transfer_entry(tools::wallet_rpc::transfer_entry &entry, const crypto::hash &payment_id, const tools::wallet2::pool_payment_details &pd) const;
+      void fill_transfer_entry(tools::wallet_rpc::transfer_entry &entry, const TransactionInfo &tx_info);
       bool not_open(epee::json_rpc::error& er);
       void handle_rpc_exception(const std::exception_ptr& e, epee::json_rpc::error& er, int default_error_code);
+      // return false if no error occured
+      bool api_error_2_rpc_error(epee::json_rpc::error& er);
```
rbrunner7: Little nitpick: As far as I could see, over the whole codebase your method name seems to be the only one that uses "2" with a meaning of "to" ...

src/wallet/wallet_rpc_server.cpp
```diff
@@ -235,18 +537,27 @@ namespace tools
         return true;
 
       uint64_t blocks_fetched = 0;
-      bool refresh_success = false;
+      bool refresh_success = !m_wallet_impl;
```
rbrunner7, line 540: I don't understand. What does the initial value of `refresh_success` have to do with the value of `m_wallet_impl`?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -954,100 +1170,73 @@ namespace tools
...
+      er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+      er.message = std::string("Must specify key image to thaw");
+      return false;
     }
-    catch (const std::exception& e)
+    crypto::key_image ki;
+    if (!epee::string_tools::hex_to_pod(req.key_image, ki))
     {
-      handle_rpc_exception(std::current_exception(), er, WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR);
+      er.code = WALLET_RPC_ERROR_CODE_WRONG_KEY_IMAGE;
```
rbrunner7, line 1203: Well, I don't want to turn your Wallet RPC migration to the Wallet API into an almost endless job, but I do feel that here and maybe also in quite a number of other places you "walked only about half the way" and missed a chance.

The Wallet API method `thaw` already returns an error if the string is not valid hex, so that now gets checked twice, here and inside the method. The missed chance: There is no `ExtendedStatus_WrongKeyImage`, and thus is not possible for the `thaw` method to return a detailed reason what went wrong, and in turn not possible for the RPC server to return its specific `WALLET_RPC_ERROR_CODE_WRONG_KEY_IMAGE` if it did not check itself as well.

It's not only the question to eliminate the conversion to hex check here, it's also about all the other callers of `thaw` that we may get over time: They all would also profit from the method setting an extended status.

The `ExtendedStatus` enum currently covers about 20 out of 50 or so possible errors, so it's possible we have about 30 cases where a Wallet API method could, ideally, use the new extended status mechanism to return more detailed error info, but does not yet.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -1084,14 +1269,6 @@ namespace tools
           return false;
         }
         integrated_payment_id = info.payment_id;
```
rbrunner7, line 1272: Now with that `add_extra_nonce_to_tx_extra` call gone, looks like you don't need any `integrated_payment_id` variable anymore?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -1391,51 +1544,46 @@ namespace tools
       return false;
     }
 
-    tools::wallet2::unsigned_tx_set exported_txs;
-    if(!m_wallet->parse_unsigned_tx_from_str(blob, exported_txs))
+    auto utx = m_wallet_impl->loadUnsignedTxFromStr(blob);
+    if (api_error_2_rpc_error(er)) return false;
+    if(!utx)
```
rbrunner7: Can this occur, i.e. can `loadUnsignedTxFromStr` still return nothing despite reporting no error? And maybe if it can occur, it shouldn't really?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -2074,64 +2052,60 @@ namespace tools
...
   {
-    if (!m_wallet) return not_open(er);
+    if (!m_wallet_impl) return not_open(er);
     crypto::hash payment_id;
     crypto::hash8 payment_id8;
-    cryptonote::blobdata payment_id_blob;
-    if(!epee::string_tools::parse_hexstr_to_binbuff(req.payment_id, payment_id_blob))
+
+    bool r;
+    if (req.payment_id.size() == 2 * sizeof(payment_id))
```
rbrunner7, line 2059: Maybe I mis-remember, or confuse something, but isn't all that "long payment id" stuff deprecated or even deactivated many years already, and thus anything that does anything in their direction code that should have been removed a long time ago already?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -2400,75 +2366,59 @@ namespace tools
...
+      if (!req.seed_offset.empty())
+        recovery_key = cryptonote::decrypt_key(recovery_key, req.seed_offset);
+
+      // generate spend key
+      cryptonote::account_base account;
+      account.generate(recovery_key, true, false);
+      spend_secret_key = account.get_keys().m_spend_secret_key;
     }
-    return true;
+    std::string spend_secret_key_str = epee::string_tools::pod_to_hex(unwrap(unwrap(spend_secret_key)));
```
rbrunner7, line 2391: Why and what for is `unwrap` needed, and why even needed twice?

src/wallet/wallet_rpc_server.cpp
```diff
@@ -4938,7 +4617,43 @@ namespace tools
...
+         MERROR((boost::format("Unknown error code: `%d`. Returning `WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR` instead.") % extended_error_code).str());
+         er.code = WALLET_RPC_ERROR_CODE_UNKNOWN_ERROR;
+      }
+      er.code = extended_error_code;
+      er.message = error_msg;
+      return true;
+    }
+    return false;
+  }
+  //------------------------------------------------------------------------------------------------------------------------------
```
rbrunner7, line 4645: Looks nice now, this error code translation :)

src/wallet/wallet_rpc_server.cpp
```diff
@@ -5042,20 +4791,23 @@ class t_daemon
...
+            LOG_ERROR(tools::wallet_rpc_server::tr("Initial refresh failed: no connection to daemon"));
+        if (!wal->refresh())
+        {
+            int error_code{};
+            std::string error_msg{};
+            wal->statusWithErrorString(error_code, error_msg);
+            LOG_ERROR(tools::wallet_rpc_server::tr("Initial refresh failed: ") << error_msg);
+        }
       }
       // if we ^C during potentially length load/refresh, there's no server loop yet
```
rbrunner7: Small typo: *length* -> *lengthy*

rbrunner7 (commented):
I could finish now a complete pass through `wallet_rpc_server.cpp`, and also looked at most other files.

I did not look closely at the tests, however. Nor did I run the new server.

All in all, looks good already. I think the simplifications that were substantial in some parts of the code are very welcome.

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: Probably not the cleanest approach, but it's to match the old behavior.
As you can see in the removed try block below, `refresh_success` was set to `true` if `m_wallet` was _not_ set. It was set to false iff `m_wallet` was set _and_ the refresh failed.

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: It's still used to make sure there is just one payment id per transaction, see `if` block above. This checks for two cases:
1. When `req.payment_id` is set, there is no integrated address
2. There is no more than one integrated address

AFAIK the first case should also be deprecated, we should not allow `req.payment_id` for any of these:
`transfer`, `transfer_split`, `sweep_single`, `sweep_all` ([src](https://docs.getmonero.org/rpc-library/wallet-rpc/#transfer))

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: Should be unreachable, will remove.

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: AFAICT putting unencrypted long payment ids into tx_extra for new transactions is deprecated, but short decrypted payment ids, including dummy "zero payment ids", still get stored as long payment ids, suffixed by zeros, see:
https://github.com/monero-project/monero/blob/c48c51ddd9202019ffcaf1ee8ff9456fa1c673ad/src/wallet/wallet2.cpp#L2525-L2529
https://github.com/monero-project/monero/blob/c48c51ddd9202019ffcaf1ee8ff9456fa1c673ad/src/wallet/wallet2.cpp#L2937

I did quite some testing with `get_bulk_payments` and think this is fine, meaning it should be working as it did on master.

src/wallet/wallet_rpc_server.cpp
SNeedlewoods: Honestly I don't know, just saw it was done in other places as well, e.g. here:
https://github.com/monero-project/monero/blob/c48c51ddd9202019ffcaf1ee8ff9456fa1c673ad/src/wallet/api/wallet.cpp#L914
_________________
Edit: Looked into it, our `spend_secret_key` is of type:
https://github.com/monero-project/monero/blob/f412232c0fb1871758a1b8a458fdb7c342924416/src/crypto/crypto.h#L76
So the first unwrap
https://github.com/monero-project/monero/blob/f412232c0fb1871758a1b8a458fdb7c342924416/contrib/epee/include/mlocker.h#L81
will leave us with `tools::scrubbed<ec_scalar>&`  and the second one
https://github.com/monero-project/monero/blob/f412232c0fb1871758a1b8a458fdb7c342924416/contrib/epee/include/memwipe.h#L72
will give use `ec_scalar&`
https://github.com/monero-project/monero/blob/f412232c0fb1871758a1b8a458fdb7c342924416/src/crypto/crypto.h#L58-L60
which AFAICT is what we need here: a reference to a trivially copyable POD type
https://github.com/monero-project/monero/blob/f412232c0fb1871758a1b8a458fdb7c342924416/contrib/epee/include/string_tools.h#L84-L90

rbrunner7:
Forgot to mention another little thing, the title of the commit for the Wallet RPC server has a typo in it:

> replace wallet2 with Walelt API in wallet-rpc

PR #10831, opened 2026-06-27
src: replace most uses of boost karma

SNeedlewoods:
https://codeberg.org/montero-project/monero/pulls/34

### DIFF
--- a/contrib/epee/src/http_auth.cpp
+++ b/contrib/epee/src/http_auth.cpp
@@ -28,6 +28,8 @@
 #include "net/http_auth.h"
 
 #include <array>
+#include <sstream>
+#include <iomanip>
 #include <boost/algorithm/string/find_iterator.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/fusion/adapted/std_tuple.hpp>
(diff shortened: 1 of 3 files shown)


contrib/epee/src/http_auth.cpp
```diff
@@ -183,8 +183,8 @@ namespace
     bool operator()(Char value) const noexcept
     {
       static_assert(std::is_integral<Char>::value, "only integral types supported");
-      return boost::spirit::char_encoding::ascii::isascii_(value) &&
-        (value == comma || boost::spirit::char_encoding::ascii::isspace(value));
+      return (value >= 0 && value < 128) &&
+        (value == comma || std::isspace(static_cast<unsigned char>(value)));
```
selsta: `std::isspace` is locale-dependent, so this is not quite equivalent to the current ASCII-specific check.

contrib/epee/src/http_auth.cpp
```diff
@@ -288,8 +286,11 @@ namespace
       std::string out{};
       out.reserve(client_reserve_size);
 
-      karma::generate(std::back_inserter(out), karma::hex(user.counter));
-      out.insert(out.begin(), 8 - out.size(), zero); // zero left pad
+      {
+        std::ostringstream ss;
```
vtnerd, line 290: This is going to be lot slower; why remove it?

src/net/tor_address.cpp
```diff
@@ -223,8 +221,7 @@ namespace net
         if (port_ != 0)
         {
             out.push_back(':');
-            namespace karma = boost::spirit::karma;
-            karma::generate(std::back_inserter(out), karma::ushort_, port());
+            out += std::to_string(port());
```
vtnerd, line 224: The existing code is slightly faster as its not creating a temporary.

vtnerd (changes requested):

vtnerd:
My primary comment might be answerable on codeberg (which is down right now), but why? Like why the random changes to working code?

PR #10833, opened 2026-06-28
rpc: include hidden pool key images in unrestricted spent checks

selsta:
Fix unrestricted is_key_image_spent queries to include hidden/local txpool entries. Restricted RPC continues to report only broadcasted transactions, preserving txpool privacy. This avoids a race where a newly submitted transaction briefly appeared unspent before its asynchronous relay-state update completed.

Should fix following CI error

```
Sending single output
  File "/__w/monero/monero/tests/functional_tests/transfer.py", line 1638, in <module>
    TransferTest().run_test()
  File "/__w/monero/monero/tests/functional_tests/transfer.py", line 81, in run_test
    self.sweep_single()
  File "/__w/monero/monero/tests/functional_tests/transfer.py", line 689, in sweep_single
    assert res.spent_status[0] == 2
AssertionError
[TEST FAILED] transfer
```

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1000,11 +1000,11 @@ namespace cryptonote
     return res;
   }
   //-----------------------------------------------------------------------------------------------
-  bool core::are_key_images_spent_in_pool(const std::vector<crypto::key_image>& key_im, std::vector<bool> &spent) const
+  bool core::are_key_images_spent_in_pool(const std::vector<crypto::key_image>& key_im, std::vector<bool> &spent, const bool include_sensitive) const
   {
     spent.clear();
 
-    return m_mempool.check_for_key_images(key_im, spent);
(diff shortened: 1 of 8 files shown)


vtnerd (commented):
Returning any tx, even in non-restricted mode, changes the policy on handling tx privacy. This was done to prevent the user from leaking privacy unintentionally.

In hindsight, this creates a problem for the wallet trying to identify spent images. Is that the crux of the dilemma here?

selsta:
Yes, the wallet transfer test instantly checks if it is spent. This usually isn't an issue in `--offline` mode but on CI sometimes there is a timing race where it would fail to show as spent.

selsta:
I opened an alternative fix here: https://github.com/monero-project/monero/pull/10853

jpk68 (approved):

jpk68 (approved):

selsta:
Closing as we went with a different solution.

PR #10836, opened 2026-06-29
Optimized handle_notify_new_transactions's duplicate tx check

SChernykh:
#10835

### DIFF
--- a/src/cryptonote_protocol/cryptonote_protocol_handler.inl
+++ b/src/cryptonote_protocol/cryptonote_protocol_handler.inl
@@ -878,17 +878,23 @@ namespace cryptonote
   int t_cryptonote_protocol_handler<t_core>::handle_notify_new_transactions(int command, NOTIFY_NEW_TRANSACTIONS::request& arg, cryptonote_connection_context& context)
   {
     MLOG_P2P_MESSAGE("Received NOTIFY_NEW_TRANSACTIONS (" << arg.txs.size() << " txes)");
-    std::unordered_set<blobdata> seen;
+    std::unordered_set<crypto::hash> seen;
+    seen.reserve(arg.txs.size());
+
     for (const auto &blob: arg.txs)
     {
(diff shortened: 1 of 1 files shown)


selsta (approved):

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -878,17 +878,23 @@ namespace cryptonote
   int t_cryptonote_protocol_handler<t_core>::handle_notify_new_transactions(int command, NOTIFY_NEW_TRANSACTIONS::request& arg, cryptonote_connection_context& context)
   {
     MLOG_P2P_MESSAGE("Received NOTIFY_NEW_TRANSACTIONS (" << arg.txs.size() << " txes)");
-    std::unordered_set<blobdata> seen;
+    std::unordered_set<crypto::hash> seen;
+    seen.reserve(arg.txs.size());
```
thomasbuilds, line 882: Minor/optional : `reserve` trusts the peer-supplied count and runs before the early duplicate-drop. Capping it keeps the benefit for the normal small-N case without sizing to an attacker-controlled count

```suggestion
    seen.reserve(std::min<size_t>(arg.txs.size(), 8192));
```

SChernykh: No. Whatever AI you used to find this, is not friends with logic. This is a standard "allocate space for N elements before adding N elements" pattern. N elements already are there in the tx vector, the peer can't choose N arbitrarily - it is a real number of elements in this vector. You can argue that it can be N 1-byte elements to inflate N, but N<16384 here because of Levin protocol limit imposed before the message is parsed at all

thomasbuilds: My bad, I got mislead by AI which I used to understand full context of PR

j-berman (approved):

PR #10839, opened 2026-06-29
cryptonote_basic: remove legacy address parser

Ap4sh:
Removes the legacy hex-format address parser branch from `get_account_address_from_str`

The normal base58 address path is unchanged. The old address blob structs, checksum helpers, and config define are removed with the parser branch

Tests:
- `cmake --build build-ap4sh-legacy-address --target unit_tests -j2`
- `./build-ap4sh-legacy-address/tests/unit_tests/unit_tests --gtest_filter='get_account_address_from_str.*'`
- `./build-ap4sh-legacy-address/tests/unit_tests/unit_tests --gtest_filter='base58_decode_addr_neg.*:base58_encode_decode_addr.*:get_account_address_from_str.*'`

### DIFF
--- a/tests/unit_tests/base58.cpp
+++ b/tests/unit_tests/base58.cpp
@@ -532,9 +532,3 @@ TEST(get_account_address_from_str, fails_on_invalid_address_view_key)
   cryptonote::address_parse_info info;
   ASSERT_FALSE(cryptonote::get_account_address_from_str(info, cryptonote::MAINNET, addr_str));
 }
-
-TEST(get_account_address_from_str, parses_old_address_format)
-{
-  cryptonote::address_parse_info info;
-  ASSERT_TRUE(cryptonote::get_account_address_from_str(info, cryptonote::MAINNET, "002391bbbb24dea6fd95232e97594a27769d0153d053d2102b789c498f57a2b00b69cd6f2f5c529c1660f2f4a2b50178d6640c20ce71fe26373041af97c5b10236fc"));
-}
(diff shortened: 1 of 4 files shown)


kayabaNerve:
Just remove these. These were never actually used AFAIK (pretty sure they've technically been accepted but no one has ever generated them) and should not be acknowledged in any way except as a half-baked remnant from Cryptonote which is inconsistent with literally everything today.

Ap4sh:
fair point, thanks

i opened this from the current parser behavior: the legacy branch is still accepted, but it skips the key validation the base58 path already has

if the preferred direction is to remove that branch entirely instead of tightening it, i am totally fine with this PR being closed or replaced by that cleanup. i did not want to push that larger behavior change without maintainer consensus

selsta:
> if the preferred direction is to remove that branch entirely instead of tightening it, i am totally fine with this PR being closed or replaced by that cleanup. i did not want to push that larger behavior change without maintainer consensus

You can just update this PR with cleanup code, no need for opening a new one and closing this one.

Ap4sh:
done, updated this PR to remove the legacy address parser instead

i also rebased it on latest master. the old format sample is still covered, but now it asserts rejection instead of acceptance, and local address/base58 unit tests pass

tests/unit_tests/base58.cpp
```diff
@@ -533,8 +534,8 @@ TEST(get_account_address_from_str, fails_on_invalid_address_view_key)
   ASSERT_FALSE(cryptonote::get_account_address_from_str(info, cryptonote::MAINNET, addr_str));
 }
 
-TEST(get_account_address_from_str, parses_old_address_format)
+TEST(get_account_address_from_str, rejects_old_address_format)
```
selsta: I think it makes sense to just remove everything associated with it. No need to test old behaviour.

Ap4sh: removed it too, thanks

selsta (approved):

tobtoht (approved):

PR #10840, opened 2026-06-30
txpool: re-relay txs that haven't been relayed and need to be

j-berman:
Fixes the following circumstance:
1. Wallet submits tx to node.
2. Node adds tx to pool and queues for relay.
3. Node shuts off before the relay executes in the strand.
4. Node restarts.
5. Now, the fix ensures the node will re-relay the tx in the `get_relayable_transactions` loop.

Also makes sure that stem txs don't have `last_relayed_time` set to max, which would prevent them from ever being re-relayed on a restart as well.

Context: https://github.com/seraphis-migration/monero/issues/365#issuecomment-4811722029

### DIFF
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -302,9 +302,10 @@ namespace cryptonote
         {
           using clock = std::chrono::system_clock;
           auto last_relayed_time = std::numeric_limits<decltype(meta.last_relayed_time)>::max();
-          if (tx_relay == relay_method::forward)
+          if (tx_relay == relay_method::forward || tx_relay == relay_method::stem)
           {
-            last_relayed_time = clock::to_time_t(clock::now() + crypto::random_poisson_seconds{forward_delay_average}());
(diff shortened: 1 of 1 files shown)


src/cryptonote_core/tx_pool.cpp
```diff
@@ -840,6 +841,8 @@ namespace cryptonote
           case relay_method::local:
           case relay_method::fluff:
           case relay_method::block:
+            if (!meta.relayed)
```
j-berman, line 844: Note: since these could have `meta.last_relayed_time == std::numeric_limits<decltype(meta.last_relayed_time)>::max()` from the `add_tx` logic, checking the `last_relayed_time` against relay delay wouldn't be helpful here.

I use `meta.relayed` here since `set_relayed` is the only place that sets `meta.relayed = true`, and also adjusts `last_relayed_time` for said txs.

selsta (approved):

vtnerd (approved):
Yikes. This looks good though, still pretty simple.

jpk68 (approved):

PR #10809, opened 2026-06-23
Track dead code/files across the codebase

thomasbuilds:

PR #10842, opened 2026-06-30
Remove dead code in epee, blockchain_db and common

thomasbuilds:
Cherry-picked from #10809 - description is in commit bodies


selsta:
It might make sense to keep set_performance_timer_log_level for debugging/testing reasons.

thomasbuilds:
imo it can go it has never had a caller since it was added in 2016. The performance_timer_log_level variable stays in the header, so debug code can still set the level directly, and log categories (perf.*) already control perf output at runtime. Can restore it though if you want.

selsta (approved):

jpk68 (approved):
LGTM

PR #10846, opened 2026-06-30
crypto: reject invalid tree path indices

Ap4sh:
This makes `tree_path()` reject leaf indexes outside the tree, matching the way `tree_branch()` already fails when the requested leaf is not present

Before this, calls like `tree_path(1, 1, &path)` returned true and produced a path for a leaf that does not exist

The new regression checks fail before the guard and pass after it

Checks:
- `cmake --build build --target unit_tests -j 2`
- `cmake --build build --target test_notifier -j 2`
- `./build/tests/unit_tests/unit_tests --gtest_filter=Crypto.tree_branch`
- `./build/tests/unit_tests/unit_tests --gtest_filter=notify.works`
- `ctest --test-dir build --output-on-failure -R '^unit_tests$'`

### DIFF
--- a/src/crypto/tree-hash.c
+++ b/src/crypto/tree-hash.c
@@ -107,7 +107,7 @@ void tree_hash(const char (*hashes)[HASH_SIZE], size_t count, char *root_hash) {
 
 bool tree_path(size_t count, size_t idx, uint32_t *path)
 {
-  if (count == 0)
+  if (count == 0 || idx >= count)
     return false;
 
   if (count == 1) {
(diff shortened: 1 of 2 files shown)


selsta:
I don't think this change makes sense to merge, changes to crypto code need a good rationale and just "hardening" is not going to be accepted.

Ap4sh:
Makes sense, thanks for the quick check

I don't have a stronger rationale here beyond tightening the helper contract, so i'll close this one

PR #10847, opened 2026-06-30
daemon: load config from explicit data dir

### LINKED ISSUE #10553: detecting a config file should look in the specified data dir

Ap4sh:
This makes monerod look for the default `bitmonero.conf` under an explicit `--data-dir` before falling back to the existing default config path

So `./monerod --data-dir /tmp/node2` reads `/tmp/node2/bitmonero.conf` if present, instead of unconditionally picking up `~/.bitmonero/bitmonero.conf`

Explicit `--config-file` still wins, and the existing testnet/stagenet/regtest data-dir suffix handling is kept

Closes #10553

Tests:
- `cmake -S . -B build-config-data-dir -G Ninja -D CMAKE_BUILD_TYPE=Release -D BUILD_TESTS=ON`
- `cmake --build build-config-data-dir --target daemon unit_tests test_notifier -j2`
- `GTEST_FILTER="-DNSResolver.*:AddressFromURL.*:select_outputs.*" ./build-config-data-dir/tests/unit_tests/unit_tests` (1267 passed, 2 hardware-only skips)
- runtime temp HOME checks for default config, explicit `--config-file`, absolute and relative `--data-dir`, missing data-dir config fallback, and testnet/stagenet/regtest paths

### DIFF
--- a/src/daemon/main.cpp
+++ b/src/daemon/main.cpp
@@ -108,6 +108,23 @@ uint16_t parse_public_rpc_port(const po::variables_map &vm)
   return rpc_port;
 }
 
+bf::path get_config_file_path(const po::variables_map &vm)
+{
+  const bf::path config_path(command_line::get_arg(vm, daemon_args::arg_config_file));
+  if (!command_line::is_arg_defaulted(vm, daemon_args::arg_config_file) ||
+      command_line::is_arg_defaulted(vm, cryptonote::arg_data_dir))
+    return config_path;
+
+  const bf::path data_dir_config =
+      bf::absolute(bf::path(command_line::get_arg(vm, cryptonote::arg_data_dir))) /
+      std::string(CRYPTONOTE_NAME ".conf");
+  boost::system::error_code ec;
+  if (bf::exists(data_dir_config, ec))
+    return data_dir_config;
+
+  return config_path;
+}
+
 #ifdef WIN32
 bool isFat32(const wchar_t* root_path)
 {
@@ -205,8 +222,7 @@ int main(int argc, char const * argv[])
       return 0;
     }
 
-    std::string config = command_line::get_arg(vm, daemon_args::arg_config_file);
-    boost::filesystem::path config_path(config);
+    boost::filesystem::path config_path = get_config_file_path(vm);
     boost::system::error_code ec;
     if (bf::exists(config_path, ec))
     {
@@ -236,7 +252,7 @@ int main(int argc, char const * argv[])
     }
     else if (!command_line::is_arg_defaulted(vm, daemon_args::arg_config_file))
     {
-      std::cerr << "Can't find config file " << config << std::endl;
+      std::cerr << "Can't find config file " << config_path.string() << std::endl;
       return 1;
     }
 


selsta:
Is this a duplicate of #10755? If yes, which implementation is more correct?

Ap4sh:
Yeah, same issue as #10755

I think this one is the safer implementation because it only switches to the data-dir config path when `--data-dir` was explicitly provided, which is what #10553 reports. It keeps the existing default `--config-file` behavior otherwise, and still uses the transformed `arg_data_dir` value so testnet/stagenet/regtest suffixes match the actual data dir

#10755 is simpler and fixes the reported case too, but it derives every defaulted config-file path from `arg_data_dir`, including the no explicit `--data-dir` case

I can close this if you prefer #10755

src/daemon/main.cpp
```diff
@@ -108,6 +108,18 @@ uint16_t parse_public_rpc_port(const po::variables_map &vm)
...
+bf::path get_config_file_path(const po::variables_map &vm)
+{
+  if (!command_line::is_arg_defaulted(vm, daemon_args::arg_config_file))
+    return command_line::get_arg(vm, daemon_args::arg_config_file);
+
+  if (!command_line::is_arg_defaulted(vm, cryptonote::arg_data_dir))
+    return bf::absolute(command_line::get_arg(vm, cryptonote::arg_data_dir)) /
+        std::string(CRYPTONOTE_NAME ".conf");
+
+  return command_line::get_arg(vm, daemon_args::arg_config_file);
```
selsta: Should this be explicity coverted to path? not sure how it's done in other places of the codepath.

Ap4sh: done, made the config path returns explicit `bf::path` conversions and rebuilt `daemon` locally

nahuhh:
"It keeps the existing default --config-file behavior"

what is the existing behavior?

if it is to check the default data-dir, then isnt that redundant? i'm not understanding why this pr is safer than the other

selsta:
@nahuhh The paths currently resolve to the same place, but deriving the default config path from `data-dir/conf` is not as clean because it assumes the default `--config-file` behavior will always match that layout.

This PR is safer because it preserves the actual default `--config-file` value unless the user explicitly sets `--data-dir`. That avoids coupling these two defaults and prevents mistakes if the config-file default changes later.

jpk68 (approved):
Looks good to me.

monerostar (commented):
## monerostar Windows path-resolution review (no full rebuild)

**Environment**
- Windows 11 native (Build 26200)
- Installed: `Monero 'Fluorine Fermi' (v0.18.5.0-release)` — `C:\Program Files\Monero GUI Wallet\monerod.exe`
- Compared `upstream/master` `src/daemon/main.cpp` vs this PR (`64447cdee7`)
- Sibling **#10755** is **CLOSED**; this is the remaining open fix for #10553
- Prior thread: selsta / author already discussed “safer than #10755” — this adds a **live Windows default split** that makes that claim concrete

### Bug (#10553)

With `--data-dir <custom>` and no `--config-file`, master still opens the **default** `arg_config_file` path, so a `bitmonero.conf` sitting only under the custom data dir is ignored.

### What this PR does

`get_config_file_path(vm)`:

1. Explicit `--config-file` → that path  
2. Else explicit `--data-dir` → `absolute(data-dir) / bitmonero.conf` (uses transformed `arg_data_dir`, so testnet/stagenet/regtest suffixes stay aligned)  
3. Else → keep default `arg_config_file` value  

Error message prints `config_path.string()` (good).

### Why Windows cares (this host)

`monerod --help` on this non-admin install already shows **split defaults**:

| Arg | Default on this box |
|-----|---------------------|
| `--config-file` | `C:\Users\Admin\AppData\Roaming\bitmonero\bitmonero.conf` (`daemonizer::get_default_data_dir` → CSIDL_APPDATA when not admin) |
| `--data-dir` | `C:\ProgramData\bitmonero` (`tools::get_default_data_dir` → CSIDL_COMMON_APPDATA) |

So “always derive conf from `arg_data_dir` when config is defaulted” is **not** a no-op on Windows even with no CLI flags.

Path-resolution matrix (same rules as the C++):

| Case | master | **this PR #10847** | closed #10755 |
|------|--------|--------------------|---------------|
| defaults only | Roaming\…\bitmonero.conf | **Roaming\…\bitmonero.conf** (unchanged) | **ProgramData\…\bitmonero.conf** ← would change default lookup |
| `--data-dir D:\xmr-node` | Roaming conf (bug) | **D:\xmr-node\bitmonero.conf** | D:\xmr-node\bitmonero.conf |
| explicit `--config-file` | as given | as given | as given |

That is exactly why keeping default `--config-file` unless `--data-dir` was **explicit** is the safer shape — matches selsta’s coupling argument, and prevents a silent Windows default regression that #10755’s broader rewrite risked.

### Limits of this review

- Code-path / OS-default proof only — **did not** full-build monerod on MSYS2 (hours).  
- No hardware wallet involved.  
- Happy to binary-smoke after merge if maintainers want a Win GUI-build check.

### Verdict

**LGTM** for #10553. Prefer this over resurrecting #10755. Small focused diff; Windows default split is a strong reason to keep the “explicit `--data-dir` only” gate.

**Formal Approve:** external collaborator — comment review only (not a maintainer Approve).

selsta:
Only issue I see currently is break in backwards compatibility. This might be okay for a major release, or should it fall back to the old default directoy if the explicit `--data-dir` has no config file?

Ap4sh:
yeah, falling back is safer here

i updated it so an existing config under the explicit data dir still wins, but if it is missing monerod keeps using the old default config path

i also rechecked the default, explicit config, relative and absolute data-dir, missing data-dir config, testnet, stagenet and regtest cases against the built daemon, then ran the full unit test binary

thanks for catching that

jpk68 (approved):

PR #10849, opened 2026-06-30
p2p: support IPv6 subnet bans

Ap4sh:
Adds IPv6 CIDR support to the existing ban paths

This lets `--ban-list`, DNS blocklist entries, daemon `ban` / `unban`, and RPC `set_bans` accept entries like `2001:db8::/32`

Single IPv6 hosts keep using the existing address parser

IPv6 subnets stay as a separate internal ban type, matching the existing IPv4 subnet approach

Nested unbans split the remaining IPv6 ranges, so `unban 2001:db8:abcd::/48` works after `2001:db8::/32` was banned

IPv4-mapped and deprecated IPv4-compatible CIDRs are rejected to avoid alternate spellings for IPv4 ranges

Tests:
- `cmake --build build --target daemon wallet_rpc_server unit_tests test_notifier --parallel 2`
- `build/tests/unit_tests/unit_tests --data-dir tests/data --gtest_filter="ban.*:get_network_address.ipv6subnet"`
- `python3 tests/functional_tests/functional_tests_rpc.py python3 tests/functional_tests build bans`
- `ctest --test-dir build/tests/unit_tests --output-on-failure -j2`

### DIFF
--- a/contrib/epee/src/net_utils_base.cpp
+++ b/contrib/epee/src/net_utils_base.cpp
@@ -17,6 +17,27 @@ static inline uint32_t make_address_v4_from_v6(const boost::asio::ip::address_v6
   return htonl(v4);
 }
 
+static inline boost::asio::ip::address_v6 apply_ipv6_cidr_mask(const boost::asio::ip::address_v6& address, const uint8_t mask)
+{
+  auto bytes = address.to_bytes();
+  uint8_t remaining = mask;
+  for (uint8_t &byte: bytes)
+  {
+    if (remaining >= 8)
+    {
+      remaining -= 8;
+    }
+    else if (remaining > 0)
+    {
+      byte &= static_cast<uint8_t>(0xffu << (8 - remaining));
+      remaining = 0;
+    }
+    else
+      byte = 0;
+  }
+  return boost::asio::ip::address_v6{bytes};
+}
+
 namespace epee { namespace net_utils
 {
 	bool ipv4_network_address::equal(const ipv4_network_address& other) const noexcept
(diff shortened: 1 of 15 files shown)


selsta:
Needs rebase.

Ap4sh:
done 👍

jpk68:
Needs another rebase :)

Ap4sh:
rebased, updated the new `unban all` path to clear IPv6 subnet bans too, and fixed an IPv4-mapped CIDR edge case while rechecking it

contrib/epee/src/net_utils_base.cpp
```diff
@@ -104,6 +125,26 @@ namespace epee { namespace net_utils
...
+	boost::asio::ip::address_v6 ipv6_network_subnet::subnet() const noexcept
+	{ return apply_ipv6_cidr_mask(m_address, m_mask); }
+
+	std::string ipv6_network_subnet::str() const
+	{ return subnet().to_string() + "/" + std::to_string(m_mask); }
+
+	std::string ipv6_network_subnet::host_str() const { return str(); }
+	bool ipv6_network_subnet::is_loopback() const { return subnet().is_loopback(); }
+	bool ipv6_network_subnet::is_local() const { return subnet().is_link_local(); }
+	bool ipv6_network_subnet::matches(const ipv6_network_address &address) const
```
jpk68, line 147: Would it be possible for the masked subnet to be cached, so that it doesn't have to be recomputed on every call?

src/p2p/net_node.inl
```diff
@@ -125,6 +126,82 @@ namespace nodetool
...
+    {
+      return ipv6_subnet_first(subnet) + ipv6_subnet_size(subnet.mask()) - 1;
+    }
+
+    epee::net_utils::ipv6_network_subnet make_ipv6_subnet(const boost::multiprecision::uint256_t &first, const uint8_t mask)
+    {
+      return {uint256_to_ipv6(first), mask};
+    }
+
+    void subtract_ipv6_subnet(
```
jpk68: It looks like this function (as well as a few related ones) are extremely similar to their existing IPv4 counterparts. It might be better to use something like a template so there's less duplicated logic.

contrib/epee/src/net_utils_base.cpp
Ap4sh: yep, the canonical subnet is cached in the constructor now, so `matches()` only masks the candidate address

src/p2p/net_node.inl
Ap4sh: agreed, i replaced the duplicated IPv4 and IPv6 subtraction helpers with one traits-based template and added `/0` and `/128` coverage

src/net/parse.cpp
```diff
@@ -226,6 +226,37 @@ namespace net
         return {epee::net_utils::ipv4_network_subnet{ip, (uint8_t)mask}};
     }
 
+    expect<epee::net_utils::ipv6_network_subnet>
+    get_ipv6_subnet_address(const boost::string_ref address, bool allow_implicit_128)
```
jpk68, line 230: This seems to reject `::ffff:a.b.c.d` but not the (deprecated) `::a.b.c.d` form; `::192.0.2.0/120` would be considered valid despite the PR description stating "IPv4-mapped IPv6 CIDRs [are] rejected".

IMO, it would also be good to add a unit test for this.

src/p2p/net_node.h
```diff
@@ -280,10 +280,13 @@ namespace nodetool
     virtual bool unblock_host(const epee::net_utils::network_address &address);
     virtual bool block_subnet(const epee::net_utils::ipv4_network_subnet &subnet, time_t seconds = P2P_IP_BLOCKTIME);
```
jpk68, line 281: Can functions like these also be combined/generalized with the existing IPv4 equivalents?

src/net/parse.cpp
Ap4sh: good catch, this now rejects both IPv4-mapped and IPv4-compatible forms, with coverage for `::192.0.2.0/120`

src/p2p/net_node.h
Ap4sh: yep, the IPv4 and IPv6 block/unblock paths now share templated implementations while keeping the existing public overloads

jpk68 (approved):
LGTM

Issue #10848, opened 2026-06-30
Non canonical key-offsets


DangerousFreedom1984:
Hello,

I was scanning the blockchain with my Python tools and my code got stuck on block `3698648`, more specifically on tx `bfb78ba8936bf1cd3b1883cb2568f4bf059c7fc1a81691b53bc0bd7ea154bcd3` because it could not correctly find the ring members for the CLSAG proof (as I was not limited by uint64 variables).

The key-offsets recorded on chain are those:
`"key_offsets": [ 157397969, 16660, 9681, 7938, 5385, 3831, 1437, 4950, 826, 813, 1237, 18446744073709482834, 18446744073709534619, 18446744073709550238, 18446744073709546177, 18446744073709183958]`

Which translates to absolute values as:

| ring | delta | global index |
|---|---|---|
| 0 | 157397969 | 157397969 |
| 1 | +16660 | 157414629 |
| 2 | +9681 | 157424310 |
| 3 | +7938 | 157432248 |
| 4 | +5385 | 157437633 |
| 5 | +3831 | 157441464 |
| 6 | +1437 | 157442901 |
| 7 | +4950 | 157447851 |
| 8 | +826 | 157448677 |
| 9 | +813 | 157449490 |
| 10 | +1237 | 157450727 |
| 11 | −68782 | 157381945 |
| 12 | −16997 | 157364948 |
| 13 | −1378 | 157363570 |
| 14 | −5439 | 157358131 |
| 15 | −367658 | 156990473 |

I believe the wallet already sort the indices in monotonic order but a custom wallet can still produce those and the nodes would accept it.

So, first I would like to know if you guys think that we should at least prevent those kind of txs at `tx_sanity_check` so it doesn't get relayed. I think it would be the best for now since those txs are already on chain (my scanner only detected that error on that block).

The benefits are that we avoid the introduction of some useless bytes in some txs and we avoid fungibility issues.

If you guys think that we should do that, I can draw a PR limiting relayed txs to only have monotonic key-offsets indices or something like that.

jeffro256:
This is my opinion: https://github.com/monero-project/monero/pull/10850#issuecomment-4873033670

selsta:
Closing as this is now documented and there are no changes planned for what is described in this issue.

PR #10850, opened 2026-07-01
cryptonote_core: reject non-canonical key offsets

Ap4sh:
## what

reject relayed txs whose relative key offsets overflow while being expanded to absolute output indices

## why

those offsets wrap back to lower absolute indices, which makes the ring non-canonical. this keeps new txs like that out of `tx_sanity_check` and the txpool relay path without changing block validation for historical txs already on-chain

refs #10848

## testing

- `git diff --check`
- `/usr/bin/cmake --build /tmp/monero-canonical-key-offsets-build --target unit_tests test_notifier -j2`
- `/tmp/monero-canonical-key-offsets-build/tests/unit_tests/unit_tests --gtest_filter='cn_format_utils.relative_output_offsets_are_non_overflowing:tx_sanity_check.*'`
- `/tmp/monero-canonical-key-offsets-build/tests/unit_tests/unit_tests --gtest_filter='cn_format_utils.*:tx_sanity_check.*'`
- `/tmp/monero-canonical-key-offsets-build/tests/unit_tests/unit_tests --gtest_filter='notify.works'`

### DIFF
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -30,6 +30,7 @@
 
 #include <atomic>
 #include <boost/algorithm/string.hpp>
+#include <limits>
 #include "wipeable_string.h"
 #include "string_tools.h"
 #include "string_tools_lexical.h"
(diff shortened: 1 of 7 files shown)


jeffro256:
With the FCMP++ fork upcoming, I'm not convinced that it's worth our time to add a relay rule for this, considering that it doesn't affect security, DoS, or anything like that.

Ap4sh:
yeah, that makes sense

given FCMP++ and the low impact here, i'll close this instead of spending more review time on it

thanks for checking

PR #10851, opened 2026-07-01
chore: remove msvc remnants

SNeedlewoods:
https://codeberg.org/montero-project/monero/pulls/37


selsta (approved):
Checked and there should be no behaviour difference for non-MSVC. The `Rpcrt4` related change appears ok.

tobtoht (approved):

PR #10852, opened 2026-07-02
wallet: track deterministic keys before encryption

### LINKED ISSUE #8947: `tools::wallet2::is_deterministic()` always returns `false` when account keys are encrypted in memory
`wallet2`, when `m_ask_password == AskPasswordToDecrypt && !m_unattended && !m_watch_only` is true, will encrypt the spend key in memory, requiring a password to decrypt if it is to be used in any cryptographic operations. The way that the `is_deterministic` method figures whether the wallet is deterministic is that it does the spend key -> view key derivation and sees if the stored keys line up as they should in a deterministic wallet. This doesn't work, however, when the spend key is encrypted. So for this method to work properly again, either 1) a password needs to be passed to this method to decrypt the spend key and then do the calculation or 2) this calculation needs to be done once and the result cached as a class field to be retrieved later so decryption is no longer needed.

https://github.com/monero-project/monero/blob/00fd416a99686f0956361d1cd0337fe56e58d4a7/src/wallet/wallet2.cpp#L1363-L1369

Ap4sh:
Fixes #8947

`wallet2::is_deterministic()` derived the expected view key from the current in-memory spend key. For attended wallets using `AskPasswordToDecrypt`, `setup_keys()` encrypts that spend key in memory, so the check returned false even for deterministic wallets

This caches the deterministic state while keys are still plaintext, before `setup_keys()` encrypts them. `get_seed()` still checks the currently loaded keys directly, so it keeps refusing to export a seed while the spend key is locked

Tests:
- `cmake --build build-ap4sh-wallet --target unit_tests test_notifier -j2`
- `./build-ap4sh-wallet/tests/unit_tests/unit_tests --gtest_filter=notify.works:wallet_keys_unlocker.*`
- `ctest --test-dir build-ap4sh-wallet -R ^unit_tests$ --output-on-failure`

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1107,6 +1107,16 @@ uint64_t gamma_picker::pick()
 
 boost::mutex wallet_keys_unlocker::lockers_lock;
 std::map<wallet2*, std::size_t> wallet_keys_unlocker::lockers_per_wallet = {};
+
+static bool keys_are_deterministic(const cryptonote::account_keys &keys)
+{
+  crypto::secret_key second;
+  keccak(reinterpret_cast<const uint8_t *>(&keys.m_spend_secret_key), sizeof(crypto::secret_key),
+    reinterpret_cast<uint8_t *>(&second), sizeof(crypto::secret_key));
+  sc_reduce32(reinterpret_cast<uint8_t *>(&second));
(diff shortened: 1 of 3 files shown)


selsta (approved):
Change looks good but would also like @jeffro256 to approve since he opened the issue.

jpk68 (approved):

PR #10856, opened 2026-07-03
Improve initial refresh: use nearest checkpoint in fast_refresh [release-v0.18]

plowsof:
cherry picked from https://github.com/monero-project/monero/pull/9935

do we also want this one for release?  https://github.com/monero-project/monero/pull/9344

there is an existing bug found under review and an 'off by one' for the easy fix - restoring at the exact height of a checkpoint means you scan from below it OR not the 1st block of your height... i think this needs tests

main bug fixed by woodsers commit

### DIFF
--- a/src/checkpoints/checkpoints.cpp
+++ b/src/checkpoints/checkpoints.cpp
@@ -158,6 +158,19 @@ namespace cryptonote
     return m_points.rbegin()->first;
   }
   //---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


nahuhh:
https://github.com/monero-project/monero/pull/9971

woodser:
I opened https://github.com/plowsof/monero/pull/27 on your branch @plowsof.

selsta (approved):

jpk68 (approved):

Issue #10322, opened 2026-02-14
Long initial sync time when monero-wallet-rpc is started without `--daemon-address` option


woodser:
If monero-wallet-rpc is started with a connected `--daemon-address` option, then creating a new wallet syncs almost immediately, from an appropriate height.

However, if `--daemon-address` is omitted, and then a wallet is created without syncing (by disabling `auto_refresh`), then a daemon connection is set via `set_daemon`, then the restore height can be ~40,000 blocks behind the current height when the wallet is synced for the first time.

As a result, the most reliable way to create a new wallet is with the `--daemon-address` option, which requires the wallet rpc server to be shut down and restarted any time the user or application wants to change the daemon connection or proxy configuration.

This issue requests using an appropriate restore height when monero-wallet-rpc is started without `--daemon-address`, and `set_daemon` is called before the wallet is synced for the first time.

Possibly related or a clarification of https://github.com/monero-project/monero/issues/8776

nahuhh:
If you use `set_daemon` _after_ creating the wallet, it has to estimate the restore height

related: https://github.com/monero-project/monero/pull/10165

^maybe try with 10165 to see if this resolves the issue

plowsof:
can this be closed? making the estimate more accurate will have reduced the issue but it's never going to be as accurate and some syncing is inevitable. 

would passing wallet-rpc our own 'restore from height' or `--chain-tip` that is used when set_daemon is omitted help?

>the --daemon-address option, which requires the wallet rpc server to be shut down and restarted any time the user or application wants to change the daemon connection or proxy configuration.

or is this the root of the issue?

woodser:
> or is this the root of the issue?

Yeah the goal was to avoid restarting the wallet RPC server to apply the startup flag `--daemon-address`.

Ideally calling RPC `set_daemon` before the wallet is synced is equivalent to using the startup flag `--daemon-address`.

nahuhh:
I think the solution here could be to add a `restore_height` parameter to `create_wallet`.

basicswap uses `generate_from_keys`, which has a `restore_height` param, so we don't have to deal with guesstimated creation height for offline wallets.

also, the guesstimated creation heights should be more accurate soon

PR #10860, opened 2026-07-05
wallet: add optional restore_height parameter to create_wallet

nahuhh:
similar to generate_from_keys, allows setting a static restore height for offline / no daemon setups.

can also be a workaround, but not a real fix for: https://github.com/monero-project/monero/issues/10322

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -3624,12 +3624,19 @@ namespace tools
       return false;
     }
     wal->set_seed_language(req.language);
-    cryptonote::COMMAND_RPC_GET_HEIGHT::request hreq;
-    cryptonote::COMMAND_RPC_GET_HEIGHT::response hres;
-    hres.height = 0;
-    bool r = wal->invoke_http_json("/getheight", hreq, hres);
-    if (r)
-      wal->set_refresh_from_block_height(hres.height);
+    if (req.restore_height > 0)
+    {
+      wal->set_refresh_from_block_height(req.restore_height);
+    }
+    else
+    {
+      cryptonote::COMMAND_RPC_GET_HEIGHT::request hreq;
+      cryptonote::COMMAND_RPC_GET_HEIGHT::response hres;
+      hres.height = 0;
+      bool r = wal->invoke_http_json("/getheight", hreq, hres);
+      if (r)
+        wal->set_refresh_from_block_height(hres.height);
+    }
     crypto::secret_key dummy_key;
     try {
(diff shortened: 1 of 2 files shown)


src/wallet/wallet_rpc_server_commands_defs.h
```diff
@@ -2187,11 +2187,13 @@ namespace wallet_rpc
   {
     struct request_t
     {
+      uint64_t restore_height;
```
SNeedlewoods, line 2190: This requires a bump of `WALLET_RPC_VERSION_MINOR` AFAICT
https://github.com/nahuhh/monero/blob/be8c864abb46c8a59fd896614db191bb1293df01/src/wallet/wallet_rpc_server_commands_defs.h#L42-L50

SNeedlewoods (changes requested):

src/wallet/wallet_rpc_server_commands_defs.h
selsta: For master branch, it's fine without changing version number.

SNeedlewoods (approved):

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3624,12 +3624,19 @@ namespace tools
       return false;
     }
     wal->set_seed_language(req.language);
-    cryptonote::COMMAND_RPC_GET_HEIGHT::request hreq;
-    cryptonote::COMMAND_RPC_GET_HEIGHT::response hres;
-    hres.height = 0;
-    bool r = wal->invoke_http_json("/getheight", hreq, hres);
-    if (r)
-      wal->set_refresh_from_block_height(hres.height);
+    if (req.restore_height > 0)
```
selsta, line 3627: I understand why it's done this way, but now if someone set restore_height zero it falls back daemon restore height. This does not match `generate_from_keys` behaviour, but I don't have a good solution for it currently.

nahuhh: Will see if i can think of something

SNeedlewoods: Maybe set default value of `restore_height` to `CRYPTONOTE_MAX_BLOCK_NUMBER`. It is already available in `src/wallet/wallet_rpc_server_command_defs.h` since it's used for `max_height` in `get_transfers`:
https://github.com/monero-project/monero/blob/2b49cebbcb002b5740202a233584fc62de23935f/src/wallet/wallet_rpc_server_commands_defs.h#L1628
Then:
``` suggestion
    if (req.restore_height <= CRYPTONOTE_MAX_BLOCK_NUMBER)
```

PR #10861, opened 2026-07-05
wallet: let set_daemon override --proxy

nahuhh:
When `proxy` was added to `set_daemon` it was decided that a commandline --proxy should not be overridden by the rpc. This change removes that restriction.

ive discovered a bunch of issues with the way monero-wallet-rpc handles its parameters (namely, dropping all params while retaining the last set_daemon address OR if --daemon-address was set, then it will revert to that(!)) so this allowed you to set a `proxy` param via rpc that would drop when you ran open_wallet, leaking your connections to the server.

this change is part 1 [of 2](https://github.com/monero-project/monero/pull/10864) to make set_daemon params persist in full across wallet changes. Having a singular  param not able to be overridden doesnt seem sane. (If you have access/credentials to the rpc server, you can do much more dangerous things than change the proxy.)

(alternative was to make it so that if a daemon was configured via command line, you could not use set_daemon to change it. A change would require restarting the server.)

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1287,11 +1287,6 @@ bool wallet2::has_stagenet_option(const boost::program_options::variables_map& v
   return command_line::get_arg(vm, options().stagenet);
 }
 
-bool wallet2::has_proxy_option() const
-{
-  return !m_proxy.empty();
-}
-
 std::string wallet2::device_name_option(const boost::program_options::variables_map& vm)
 {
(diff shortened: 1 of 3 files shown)


jpk68:
Relevant: https://github.com/xmrack/monero-review/issues/476

SNeedlewoods (approved):

jpk68 (approved):

jpk68 (approved):

selsta (approved):

PR #10862, opened 2026-07-05
wallet: reconnect before submitting txs

### LINKED ISSUE #10032: wallet: failed txs due to dropping connection during long operations
When doing some operation that takes a long time (10s of minutes), such as a sweep_all with many outputs (i think ive seen this also for `get output distribution`), after it finishes constructing the txs, it fails to broadcast them, citing `no_connection_to_daemon`.. followed by an immediate reconnection. The txs dont broadcast, so you're left with wasted time and forced to manually construct in parts.

This happens using cli and rpc.
daemon specified on the command line (not via rpc command).
Using local daemon, trusted

The fix here is likely just to reconnect before the send operation.

Ap4sh:
Fixes #10032

## What changed
- reconnect the wallet daemon HTTP client before the first `/sendrawtransaction` in `wallet2::commit_tx`
- keep vector submission to one reconnect per batch, then reuse the fresh connection for the rest of the batch
- add unit coverage with a fake daemon HTTP client to lock the disconnect/connect/submit order

## Why
#10032 reports long wallet operations finishing tx construction, then failing broadcast with `no_connection_to_daemon` before reconnecting right after. The HTTP client only auto-connects when it already knows the socket is disconnected, so a stale-looking connection can still be reused for the submit path

This does not retry after a failed submit, so it avoids re-broadcasting a tx that may have reached the daemon already

cc @nahuhh since this follows the issue report

## Tests
- `cmake --build build-wallet-broadcast-reconnect --target unit_tests -j2`
- `./build-wallet-broadcast-reconnect/tests/unit_tests/unit_tests --gtest_filter='wallet_daemon_connection.*' --data-dir tests/data`
- `ctest --test-dir build-wallet-broadcast-reconnect -R '^unit_tests$' --output-on-failure -j2`

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -7570,6 +7570,11 @@ crypto::hash wallet2::get_payment_id(const pending_tx &ptx) const
 // take a pending tx and actually send it to the daemon
 void wallet2::commit_tx(pending_tx& ptx)
 {
+  commit_tx(ptx, true);
+}
+
+void wallet2::commit_tx(pending_tx& ptx, bool reconnect)
+{
   using namespace cryptonote;
 
   // Normal submit
@@ -7581,6 +7586,8 @@ void wallet2::commit_tx(pending_tx& ptx)
 
   {
     const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};
+    if (reconnect)
+      m_http_client->disconnect();
     bool r = epee::net_utils::invoke_http_json("/sendrawtransaction", req, daemon_send_resp, *m_http_client, rpc_timeout);
     THROW_ON_RPC_RESPONSE_ERROR(r, {}, daemon_send_resp, "sendrawtransaction", error::tx_rejected, ptx.tx, get_rpc_status(m_trusted_daemon, daemon_send_resp.status), get_text_reason(daemon_send_resp));
   }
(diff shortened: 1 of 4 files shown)


nahuhh:
I'll try to reproduce & test fix in the next couple days. Need to spin up an fcmp++ regtest and construct some 128input txs on a low power device

jpk68 (approved):
Looks good to me, though it's likely worth noting that I have not actually tested this with real stale connections.

selsta:
I don't feel comfortable merging this yet, an unconditional reconnect is a blunt workaround and I want to fully figure out first why this is failing. I've asked for more info in the linked issue.

Ap4sh:
makes sense, the unconditional reconnect is too broad without a confirmed cause

#10940 now addresses stale half-closed HTTP connections in the shared client, so it may supersede this if that matches the reproduction in #10032

i'll move this back to draft while the issue is investigated

woodser:
> https://github.com/monero-project/monero/pull/10940 now addresses stale half-closed HTTP connections in the shared client, so it may supersede this if that matches the reproduction in https://github.com/monero-project/monero/issues/10032

I'm seeing that #10940 definitely improves overall connection reliability for various wallet functions, making a daemon connection error less likely for long multi-step functions like tx creation.

plowsof:
had this issue while slowly confirming/signing a 15 out transaction on the trezor - only to see it fail at the end, testing  #10940 which feels like the root cause (idle for several minutes while connected to a remote node)

PR #10863, opened 2026-07-05
wallet: rpc; req. ssl cert params w/ proxy when daemon not onion/i2p

nahuhh:
For cli and rpc command line, daemon-ssl-* flags are required when --proxy is set, unless the daemon addr is onion or i2p

the set_daemon rpc param for proxy bypasses those requirements. Add them back here

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -4827,14 +4827,17 @@ namespace tools
       std::move(req.ssl_private_key_path), std::move(req.ssl_certificate_path)
     };
 
+    // If a proxy or SSL is explicitly enabled, then ssl_allow_any_cert, ssl_ca_file, ssl_allowed_fingerprints, or use of a .onion or .i2p address is required
+    const boost::string_ref real_daemon = boost::string_ref{req.address}.substr(0, req.address.rfind(':'));
+    const bool use_proxy = !req.proxy.empty();
     const bool verification_required =
       ssl_options.verification != epee::net_utils::ssl_verification_t::none &&
-      ssl_options.support == epee::net_utils::ssl_support_t::e_ssl_support_enabled;
+      (ssl_options.support == epee::net_utils::ssl_support_t::e_ssl_support_enabled || use_proxy);
 
-    if (verification_required && !ssl_options.has_strong_verification(boost::string_ref{}))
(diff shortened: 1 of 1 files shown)


src/wallet/wallet_rpc_server.cpp
```diff
@@ -4827,14 +4827,17 @@ namespace tools
       std::move(req.ssl_private_key_path), std::move(req.ssl_certificate_path)
     };
 
+    // a proxy or SSL requires strong verification, same as the command line
+    const boost::string_ref daemon_host = boost::string_ref{req.address}.substr(0, req.address.rfind(':'));
```
selsta:
```suggestion
    const boost::string_ref real_daemon = boost::string_ref{req.address}.substr(0, req.address.rfind(':'));
```

This matches wallet2 var name, so it's easier to grep in the future.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -4827,14 +4827,17 @@ namespace tools
       std::move(req.ssl_private_key_path), std::move(req.ssl_certificate_path)
     };
 
+    // a proxy or SSL requires strong verification, same as the command line
```
selsta: `ssl_allow_any_cert` sets `verification = none`, so the comment is currently a bit too broad. The wallet2 comment is better:

```
/* If SSL or proxy is enabled, then a specific cert, CA or fingerprint must 
   be specified. This is specific to the wallet. */
```

nahuhh:
```
 /* If SSL or proxy is enabled, then a specific cert, CA or fingerprint must 
   be specified. This is specific to the wallet. */
```

this comment doesnt seem great either, since you can technically skip a cert by using ssl_allow_any_cert

maybe something like the error log on 4840? (Which replaces an incorrect err log)

nahuhh: Updated comment to: 

// If a proxy or SSL is explicitly enabled, then ssl_allow_any_cert, ssl_ca_file, ssl_allowed_fingerprints, or use of a .onion or .i2p address is required

src/wallet/wallet_rpc_server.cpp
nahuhh: Changed

selsta (approved):

plowsof (approved):

PR #10864, opened 2026-07-05
wallet: persist set_daemon params across wallet loads.

### LINKED ISSUE #10322: Long initial sync time when monero-wallet-rpc is started without `--daemon-address` option

nahuhh:
Fixes: https://github.com/monero-project/monero/issues/10322

Builds on top of: https://github.com/monero-project/monero/pull/10861

context: attempted to move basicswap away from setting commandline daemon flags for monero-wallet-rpc, so it could failover to a user prescribed list of nodes using set_daemon.

ran into a few issues
1. After using `set_daemon` -> `open_wallet`, if there was a daemon set on the commandline (--daemon-address), the connection is unexpectedly reverted to it on a wallet load.
2. Even if you started monero-wallet-rpc without --daemon-address, and exclusively used `set_daemon`, then **all params** would be dropped when you issue an `open_wallet`. _the daemon is retained!_ this means that you will leak your ip to the node, since proxy is dropped.
(this was discovered by basicswap's ci. It ran a rest against a password protected node, and on open_wallet, the connection would fail because the credentials were dropped. Upon investigation, it was discovered that _all_ params are dropped)

* params dropping also happens for (1). Only daemon flags are retained. Not as bad of an outcome, since the configure node was intentional, whereas set_daemon node dropping ssl and proxy is anything-but intentional.
3. You cannot run set_daemon before opening a wallet.

This PR
1. Allows you to run set_daemon without a wallet loaded. Same behavior as CLi flags. 
2. Values set by set_daemon override CLI flags.
3. Values set by set_daemon are retained across wallet changes.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -321,7 +321,7 @@ std::string get_weight_string(const cryptonote::transaction &tx, size_t blob_siz
   return get_weight_string(get_transaction_weight(tx, blob_size));
 }
 
-std::unique_ptr<tools::wallet2> make_basic(const boost::program_options::variables_map& vm, bool unattended, const options& opts, const std::function<boost::optional<tools::password_container>(const char *, bool)> &password_prompter)
+std::unique_ptr<tools::wallet2> make_basic(const boost::program_options::variables_map& vm, bool unattended, const options& opts, const std::function<boost::optional<tools::password_container>(const char *, bool)> &password_prompter, const boost::optional<tools::wallet2::daemon_config>& daemon_override = boost::none)
 {
   const bool testnet = command_line::get_arg(vm, opts.testnet);
   const bool stagenet = command_line::get_arg(vm, opts.stagenet);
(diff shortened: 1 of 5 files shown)


src/wallet/wallet2.cpp
```diff
@@ -450,7 +450,19 @@ std::unique_ptr<tools::wallet2> make_basic(const boost::program_options::variabl
...
+    login = boost::none;
+    if (!daemon_override->username.empty() || !daemon_override->password.empty())
+      login.emplace(daemon_override->username, daemon_override->password);
+    proxy = daemon_override->proxy;
+    trusted_daemon = daemon_override->trusted;
+    ssl_options = daemon_override->ssl_options;
+  }
+
+  // set --trusted-daemon if local and not overridden by command line or set_daemon
   if (!trusted_daemon)
```
```suggestion
  if (!trusted_daemon.is_initialized())
```

otherwise `set_daemon(trusted=false)` is replaced by automatic detection.

nahuhh:
rebased, no changes. letting test run again. failed here and on my branch, but at different places

selsta:
Tests are green, needs squash.

jpk68:
Relevant: https://github.com/xmrack/monero-review/issues/195

SNeedlewoods (approved):

jpk68 (approved):

navidR (approved):

selsta (approved):

jpk68 (approved):

navidR (approved):

plowsof (approved):

Issue #9089, opened 2023-12-17
Bug: it's theoretically possible (seems *very* unlikely) to be unable to restore a valid seed


j-berman:
### How to see the bug

Take the following valid Spanish seed as an example:

```
seed
pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma pluma

spend key
b4050000b4050000b4050000b4050000b4050000b4050000b4050000b4050000
```

It's theoretically possible (obviously extremely unlikely) for a wallet to generate the above seed.

When restoring this seed, it restores to the English seed:

```
seed
plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus plus

spend key
3b0400003b0400003b0400003b0400003b0400003b0400003b0400003b040000
```

### Why it happens

The wallet tries to infer what language the seed is in. It does this by checking the first X letters of each seed word against each language mnemonic dictionary. For English, only the first 3 letters of each seed word are checked. For Spanish, only the first 4 letters of each seed word are checked.

The first 3 letters of `pluma` -> `plu` is the same as `plus` -> `plu`.

The wallet checks if a seed is English before it checks if the seed is Spanish. When checking if the seed is English, the wallet trims each word to the first 3 letters `plu`, sees `plu` is in the trimmed English mnemonic dictionary, then calculates the seed's checksum using the trimmed words and sees the checksum is valid. The wallet then infers the seed's language is the first valid checksum. Thus, in this case, the checksum passes for English **before the wallet can even check if the checksum is valid in Spanish**, and so the wallet incorrectly thinks the seed is the English seed `plus plus ... plus`.

#### Potentially disruptive but optimal solution

On restore, require the entire seed provided match all complete words (so `pluma` could never be interpreted as English).

With this solution, users would no longer be able to restore using trimmed words. This might be a bit disruptive for some CLI/GUI users who thought the trimmed words were fine (they would need to consult the full mnemonic dictionary to see what their trimmed words correspond to). Note that Feather auto-fills seed words and requires full words on restore (the GUI could do what Feather does on this too, the CLI could also provide suggestions on error).

#### Less disruptive solution

Only require the complete correct seed in the event a given seed has multiple valid checksums across different languages (as is the case in the above sample seed).

This means the edge case is applied to some already likely very niche users in exceptionally rare circumstances who expect to be able to use trimmed words.

[...]

PR #10765, opened 2026-06-14
Polyseed support for 'wallet2', Wallet API, CLI wallet app, wallet RPC server

rbrunner7:
This (finally) implements support for Polyseed as created by @tevador in `wallet2`, the Wallet API, the CLI wallet app and the wallet RPC server.

Substantial parts come, with some changes, from [this patch](https://github.com/MrCyjaneK/monero_c/blob/develop/patches/monero/0007-polyseed.patch) authored by @tobtoht and refined by @MrCyjaneK for the [monero-c](https://github.com/MrCyjaneK/monero_c) library. As part of that library the code from that patch is in active use in a number of third-party wallet apps for quite a while now.

The code in `simplewallet` for the CLI wallet app and in `wallet_rpc_server` is my work.

A bug regarding the patch in connection with background sync of Polyseed wallets that I learned about from a discussion in a Matrix channel (the seed info gets lost) is corrected in a clean way now.

It may be that the Polyseed methods in the Wallet API are still quite basic and not yet sufficient to implement really comprehensive support in an app like the Monero GUI wallet app, but without actually trying it's hard to find out what's missing, if anything, and I did not try to go beyond what said patch already brings.

PR #10866, opened 2026-07-06
mnemonics: decode seeds with explicit language

Ap4sh:
## what

Decode Electrum seeds with the explicitly provided mnemonic language when one is supplied by `monero-wallet-cli --mnemonic-language` or wallet RPC `restore_deterministic_wallet.language`

The automatic detection path is unchanged

## why

Related to #9089

That issue shows a valid Spanish seed that gets inferred as English because English is checked first and the 3-letter prefix checksum passes for `plus`

The CLI and RPC already expose a way to provide a seed language, but restore parsing still inferred the language before those values could help

With this change, callers that know the seed language can restore that seed as Spanish without changing default restore behavior

## testing

- `./build-seed-language/tests/unit_tests/unit_tests --gtest_filter='mnemonics.*' --data-dir build-seed-language/tests/data`
- `cmake --build build-seed-language --target monero-wallet-cli monero-wallet-rpc -j2`
- `ctest --test-dir build-seed-language -R '^unit_tests$' --output-on-failure`
- local offline `monero-wallet-cli` restore of the `pluma ...` seed with and without `--mnemonic-language Spanish`
- local offline `monero-wallet-rpc restore_deterministic_wallet` call with `language: "Spanish"`

### DIFF
--- a/src/mnemonics/electrum-words.cpp
+++ b/src/mnemonics/electrum-words.cpp
@@ -78,6 +78,7 @@ namespace
   uint32_t create_checksum_index(const std::vector<epee::wipeable_string> &word_list,
     const Language::Base *language);
   bool checksum_test(std::vector<epee::wipeable_string> seed, const Language::Base *language);
+  bool language_is_expected(const Language::Base *language, const std::string *expected_language);
 
   /*!
    * \brief Finds the word list that contains the seed words and puts the indices
@@ -88,8 +89,15 @@ namespace
    * \param  language        Language instance pointer to write to after it is found.
    * \return                 true if all the words were present in some language false if not.
    */
+  bool language_is_expected(const Language::Base *language, const std::string *expected_language)
+  {
+    return !expected_language || *expected_language == language->get_language_name() ||
+      *expected_language == language->get_english_language_name();
+  }
+
   bool find_seed_language(const std::vector<epee::wipeable_string> &seed,
-    bool has_checksum, std::vector<uint32_t> &matched_indices, Language::Base **language)
+    bool has_checksum, std::vector<uint32_t> &matched_indices, Language::Base **language,
+    const std::string *expected_language = NULL)
   {
     // If there's a new language added, add an instance of it here.
     std::vector<Language::Base*> language_instances({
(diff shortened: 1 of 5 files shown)


selsta:
ping @rbrunner7

rbrunner7:
It pains me to say so, but I think this PR is quite unfortunate.

I am firmly convinced that we should welcome each and every new contributor with open arms, and IMHO basically every PR has the right to get considered as any other one, but despite all this my personal vote is not going forward with this at this point in time.

First of all, this will clash quite badly with my [Polyseed PR](https://github.com/monero-project/monero/pull/10765). If that gets merged first, to resolve "conflicts", @Ap4sh would be forced to put in substantial work to extend the mechanism to Polyseed. If this gets merged first, I would be forced to do it for Polyseed in my PR. From a project management point of view, it would be much easier to wait until my PR gets merged, and then come back to this mechanism to implement it for both seed technologies at the same time.

But I see another, more serious issue: This problem is so small, I claim we can neglect it, and maybe also *should* neglect it to avoid carrying around basically unused code ad infinitum, adding yet another CLI wallet startup parameter, and adding yet another RPC parameter value.

Nobody doubts that it's possible to manually *construct* ambiguous seeds, but we generate random ones for our wallets, and that's a quite different story.

Just 2 days ago for my PR I tried to randomly generate ambiguous Spanish and Italian seeds, but with no success after around 200 million generated seeds I stopped.

Polyseeds merely have 16 words; legacy seeds have 24 relevant words, which means that ambiguous legacy seeds are probably millions of times less probable than such Polyseeds.

Furthermore: Most people never have to restore. And then this: Probably not many people use seeds in other languages than English. Thirdly: Legacy seeds will soon be on their way out.

I recently added code to my PR to make sure that a randomly generated but ambiguous Polyseeds gets detected, and another random one gets generated, until success. I am pretty sure this will *also* be code that will *never* run, but at least it's only about 10 lines of code in a single method.

Ap4sh:
Thanks a lot for the depth in reply!
Will check this more in details soon.

Ap4sh:
thanks a lot for taking the time to write this up in detail

i looked through #10765 more carefully now, and i agree this should not get in the way of the polyseed work

one point i want to clarify: this PR is not adding a new cli startup parameter or a new rpc field. `--mnemonic-language` already exists, and `restore_deterministic_wallet.language` already exists too. this only makes those existing inputs affect legacy seed decoding before the current auto-detection path picks the first matching language

so the default restore path stays unchanged. the narrow intent is only to make the #9089 case recoverable when the caller already knows the seed language, e.g. a Spanish legacy seed that currently gets accepted as English first

i also checked the interaction with #10765. the overlap is in the cli/rpc restore path. if polyseed lands first, i think this should be rebased by passing the same legacy language hint through the legacy fallback in `words_to_bytes_ex()`, while leaving polyseed's own decode behavior to the polyseed layer

so if the preference is to land polyseed first and handle this afterwards, i'm fine parking this until then

rbrunner7:
Thanks to you for all the info. I see, the additional "machinery" that this PR brings is less than I feared, which is nice.

And yes, I would be grateful if I "could go first" with my Polyseed PR, it's quite complicated already on its own.

Ap4sh:
yeah, no problem, i moved this back to draft and will leave it parked until #10765 settles

PR #10867, opened 2026-07-06
daemonizer: replace some fragile posix functions

SNeedlewoods:
https://codeberg.org/montero-project/monero/pulls/38

### DIFF
--- a/src/daemonizer/posix_fork.cpp
+++ b/src/daemonizer/posix_fork.cpp
@@ -110,10 +110,7 @@ void fork(const std::string & pidfile)
     {
       // We're in the parent process and need to exit.
       close_pid_fd();
-      // When the exit() function is used, the program terminates without
-      // invoking local variables' destructors. Only global variables are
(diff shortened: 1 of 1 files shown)


jpk68:
Some context on these changes, since the mirror repo was removed:

- In the code's current state, the exiting parent process calls `exit(0)` after each of the two forks, which risks output buffers being flushed twice, and handlers firing both there and when the daemon actually exits. On the other hand, `_exit(0)` skips both of these side effects, and seems to be more idiomatic for this sort of daemon.
- The `dup()` function automatically uses the lowest available file descriptor, which is `2` here just because the descriptors `0` and `1` were closed and re-opened just above in the same exact order. This means any reordering of them later can silently point `stderr` somewhere else; `dup2()` mitigates this.

Sources:
- https://stackoverflow.com/a/5423108
- https://stackoverflow.com/questions/1720535/practical-examples-use-dup-or-dup2

PR #10870, opened 2026-07-07
wallet: rpc; dont allow startup with missing ssl params

nahuhh:
Starting monero-wallet-rpc with a --wallet-dir + clearnet daemon + proxy (or --daemon-ssl=enabled) would start the daemon, but upon running open_wallet, the daemon woukd attempt to apply, and fail due to missing --daemon-ssl* flags. You cannot fix this without modifying the startup flags.

this change makes it so that the wallet-rpc will error and refuse to start, instead of not-checking until a wallet is opened

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -4988,6 +4988,15 @@ public:
 
       if (!wallet_dir.empty())
       {
+        try
+        {
+          tools::wallet2::make_dummy(vm, true, password_prompt);
+        }
+        catch (const std::exception &e)
+        {
+          LOG_ERROR(tools::wallet_rpc_server::tr("Invalid configuration: ") << e.what());
+          return false;
+        }
         wal = NULL;
         goto just_dir;
       }


src/wallet/wallet_rpc_server.cpp
```diff
@@ -4988,6 +4988,15 @@ class t_daemon
 
       if (!wallet_dir.empty())
       {
+        try
+        {
+          tools::wallet2::make_dummy(vm, true, password_prompt);
+        }
+        catch (const std::exception &e)
+        {
+          LOG_ERROR(tools::wallet_rpc_server::tr("Invalid daemon connection options: ") << e.what());
```
selsta: Can this fail for other reasons unrelated to daemon connection?

SNeedlewoods: Yes it can, for example here https://github.com/monero-project/monero/blob/2b49cebbcb002b5740202a233584fc62de23935f/src/wallet/wallet2.cpp#L330 (`make_basic()` gets called by `make_dummy()`)

nahuhh: Changed it to a more generic "Invalid configuration:"

selsta (approved):

SNeedlewoods (approved):

PR #10872, opened 2026-07-08
cryptonote_basic: copy difficulty hash words

Ap4sh:
## summary
- copy 64-bit difficulty hash words out of `crypto::hash::data` before endian conversion
- avoid reading the packed hash object through `uint64_t*` in `check_hash_64()` and `check_hash_128()`
- update the difficulty unit-test helper to build hashes through the byte array too

## tests
- `cmake -S . -B build/debug -D CMAKE_C_COMPILER=/usr/bin/cc -D CMAKE_CXX_COMPILER=/usr/bin/c++ -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=Debug -D BUILD_SHARED_LIBS=OFF`
- `cmake --build build/debug --target difficulty-tests hash-target-tests unit_tests test_notifier -j2`
- `ctest --test-dir build/debug -R '^(hash-target|difficulty|wide_difficulty)$' --output-on-failure`\n- `build/debug/tests/unit_tests/unit_tests '--gtest_filter=difficulty.*'`\n- `ctest --test-dir build/debug -R '^unit_tests$' --output-on-failure`

### DIFF
--- a/src/cryptonote_basic/difficulty.cpp
+++ b/src/cryptonote_basic/difficulty.cpp
@@ -32,6 +32,7 @@
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
 #include <vector>
 
 #include "int-util.h"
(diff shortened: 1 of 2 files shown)


selsta (approved):

src/cryptonote_basic/difficulty.cpp
```diff
@@ -102,18 +103,25 @@ namespace cryptonote {
     return a + b < a || (c && a + b == (uint64_t) -1);
   }
 
+  static inline uint64_t hash_word(const crypto::hash &hash, size_t index) {
```
jeffro256:
```suggestion
  static inline uint64_t load_u64_from_hash(const crypto::hash &hash, size_t index) {
```

"word" is an ambiguous term, and `hash_word` implies that we're hashing a word

src/cryptonote_basic/difficulty.cpp
```diff
@@ -102,18 +103,25 @@ namespace cryptonote {
...
+    assert(index < sizeof(hash.data) / sizeof(uint64_t));
+    uint64_t word;
+    memcpy(&word, hash.data + index * sizeof(word), sizeof(word));
+    return swap64le(word);
```
jeffro256:
```suggestion
    return SWAP64LE(word);
```

src/cryptonote_basic/difficulty.cpp
Ap4sh: yeah that is clearer, renamed it to `load_u64_from_hash` and updated every call site

src/cryptonote_basic/difficulty.cpp
Ap4sh: updated this to use `SWAP64LE` too

selsta (approved):

jpk68 (approved):

PR #10874, opened 2026-07-08
add thomasbuilds' PGP key

thomasbuilds:
https://github.com/monero-project/gitian.sigs/pull/333#issuecomment-4918430332


selsta:
Adding it to the gitian.sigs repo is enough, I'll ask bF for a clarification since it appears you did add it there.

selsta:
bF confirmed to me that the issue is resolved, so we don't need to merge it here.

PR #10586, opened 2026-05-13
cmake: add support for sccache

selsta:

PR #10876, opened 2026-07-09
cmake: allow user to select compiler cache

jpk68:
https://github.com/monero-project/monero/pull/10586#discussion_r3268384596


CMakeLists.txt
```diff
@@ -59,7 +59,13 @@ endif()
 project(monero)
 
 option (USE_COMPILER_CACHE "Use compiler cache if a usable instance is found (sccache/ccache)" ON)
-if (USE_COMPILER_CACHE)
+set (COMPILER_CACHE "" CACHE STRING "Compiler cache to use (ccache/sccache); overrides auto-detection")
```
selsta: I know it says overrides auto-detection but having `-D USE_COMPILER_CACHE=OFF -D COMPILER_CACHE="sccache"` end up using sccache is a bit confusing.

jpk68: Thanks, done.

selsta:
In general, my preferred option would be a single `COMPILER_CACHE` or `MONERO_COMPILER_CACHE` variable with the following options

```
auto, sccache, ccache, none
```

auto would try sccache, fallback to ccache and if that also does not exist use nothing, the others are self explanatory.

We only introduced `USE_COMPILER_CACHE` on master 2 months ago so I think we could still change this.

selsta (approved):

tobtoht (approved):

PR #10879, opened 2026-07-09
docs: clarify license info

jpk68:
`CONTRIBUTING.md` states that the project must use a share-alike license, however BSD-3 is a permissive license.


selsta (approved):

jeffro256 (approved):

tobtoht (approved):

jpk68:
Updated to only allow for permissive licenses, after some discussion with @tobtoht.

tobtoht (approved):
Since we have already chosen a license, this rule would guard against re-licensing the project under a stricter license.

nahuhh (approved):

PR #10881, opened 2026-07-09
simplewallet: fix potential out-of-bounds read

jpk68:
It appears that if the `donate` command is run with a single (valid) argument, it will be removed from `local_args`, with `.back()` then called to parse the amount, even though the vector is empty.


selsta:
I would just remove all that payment ID code, since the donation address is a subaddress which does not allow payment IDs and expect exactly one argument.

selsta (approved):

tobtoht (approved):

PR #10882, opened 2026-07-10
util: don't use glibc to detect unix

jpk68:
Since `unistd.h` is already included, it seems `_POSIX_VERSION` can be used to detect whether or not this is a POSIX system; a few other headers also appear to be relevant only to Linux.

### DIFF
--- a/src/common/util.cpp
+++ b/src/common/util.cpp
@@ -31,19 +31,16 @@
 #include <unistd.h>
 #include <cstdio>
 #include <wchar.h>
+#include <string>
 
-#ifdef __GLIBC__
+#if defined(_POSIX_VERSION)
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/resource.h>
-#include <dirent.h>
-#include <string.h>
-#include <ctype.h>
-#include <string>
 #endif
 
 //tools::is_hdd
-#ifdef __GLIBC__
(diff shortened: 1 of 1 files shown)


selsta:
This changes runtime behavior on non-glibc POSIX platforms like macOS or BSDs. I'm not necessarily opposed to this but it needs an explanation why this is safe and better.

jpk68:
> This changes runtime behavior on non-glibc POSIX platforms like macOS or BSDs. I'm not necessarily opposed to this but it needs an explanation why this is safe and better.

The purpose of this change is simply to ensure POSIX functionality will work on all POSIX systems rather than just ones using `glibc`. This includes macOS, BSDs, and Linux systems using alternative implementations such as `musl`.

There have been numerous reported behaviour discrepancies in `monerod` and friends on POSIX-compatible systems that do not use `glibc` (especially OpenBSD), and I suspect this may have something to do with it; the existence of `glibc` is not a reliable way to check for whether POSIX APIs are supported.

I am reasonably confident these changes are safe, as all POSIX systems should support these APIs.

iamamyth:
The change looks OK to me (I corrected `is_hdd`, and wrote its tests, so I'm pretty sure the changed scope makes sense there).

iamamyth (approved):

PR #10888, opened 2026-07-10
serialization: simplify tx.signatures blobs

jeffro256:
Simplifies serialization of `cryptonote::transaction::signatures`. Also, puts static assert reminder in the unit tests to update `compare_tx_to_json_tx()` when `MAX_HF_VERSION` is bumped.

### DIFF
--- a/src/cryptonote_basic/cryptonote_basic.h
+++ b/src/cryptonote_basic/cryptonote_basic.h
@@ -282,8 +282,17 @@ namespace cryptonote
           PREPARE_CUSTOM_VECTOR_SERIALIZATION(signature_size, signatures[i]);
           if (signature_size != signatures[i].size())
             return false;
+          else if (0 == signature_size)
+            continue; // v1 miner txs serialize an empty signatures outer array
 
-          FIELDS(signatures[i]);
+          // for non-miner, serialize all inner `crypto::signature`s for this input in one big blob
+          static_assert(std::has_unique_object_representations_v<crypto::signature>);
+          ar.serialize_blob(signatures[i].data(), signature_size * sizeof(crypto::signature));
+          if (!ar.good())
+          {
+            if (!Archive<W>::is_saving()) signatures.clear(); // on load failure, clear all sigs for good measure
+            return false;
+          }
 
           if (vin.size() - i > 1)
             ar.delimit_array();
--- a/src/cryptonote_config.h
+++ b/src/cryptonote_config.h
@@ -194,6 +194,7 @@
 #define HF_VERSION_BULLETPROOF_PLUS             15
 #define HF_VERSION_VIEW_TAGS                    15
 #define HF_VERSION_2021_SCALING                 15
+#define MAX_HF_VERSION                          16
 
 #define PER_KB_FEE_QUANTIZATION_DECIMALS        8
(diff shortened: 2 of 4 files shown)


selsta:
This fails `Serialization.serializes_transacion_signatures_correctly`.

jeffro256:
It's arguable whether the unit test is superfluous here (test checks that *writing* coinbase tx with a populated `signatures` field returns `false`), but I can update it so that it passes. Alternatively, that specific test clause can be dropped. The important thing would be that the new code rejects *loading* a coinbase tx with populated `signatures` field, which this code does, because that may cause a hard fork.

selsta (approved):

src/cryptonote_basic/cryptonote_basic.h
```diff
@@ -282,8 +282,12 @@ namespace cryptonote
           PREPARE_CUSTOM_VECTOR_SERIALIZATION(signature_size, signatures[i]);
           if (signature_size != signatures[i].size())
             return false;
+          else if (0 == signature_size)
+            continue; // v1 miner txs serialize an empty signatures outer array
 
-          FIELDS(signatures[i]);
+          // for non-miner, serialize all inner `crypto::signature`s for this input in one big blob
+          static_assert(std::has_unique_object_representations_v<crypto::signature>);
+          ar.serialize_blob(signatures[i].data(), signature_size * sizeof(crypto::signature));
```
vtnerd, line 290: (1) A check for `if (!ar.good()) return false;` ??

(2) This also ruins the json output, which needs to output a string before `serialize_blob`.

(3) If there isn't enough bytes for the signatures, then signatures is cleared, but this seems less important to maintain.

vtnerd (changes requested):

src/cryptonote_basic/cryptonote_basic.h
jeffro256: 1) It is done at the end of the function call.
2) In the JSON archive, the default string delimiter parameter of `serialize_blob()` is `"`. So a default call to `serialize_blob()` will correctly output a string, as is done in other blob types, e.g. `crypto::hash`. https://github.com/monero-project/monero/blob/2044af1952262c5a166d27e769bcaf544ce9c1a6/src/serialization/json_archive.h#L139
3) This is a fair point. I don't know of any code which relies on this property, but it would be simple enough to implement here. Do you think it's worth it to add?

vtnerd: Id like clearing structures in my new serialization code because I felt like it prevented issues. Somewhat subjective though

vtnerd:
> It's arguable whether the unit test is superfluous here (test checks that writing coinbase tx with a populated signatures field returns false),
 
Could someone generate invalid blocks? That's about the only negative.

jeffro256:
Not AFAIK with the current mining code. But anyways that comment isn't applicable anymore since I fixed that a couple days ago in [bcd305e](https://github.com/monero-project/monero/commit/bcd305e3cae3a62284fe806c3a45cef33a6d4fff)

src/cryptonote_basic/cryptonote_basic.h
jeffro256: Okay, done in latest

selsta:
Fails to compile.

PR #10889, opened 2026-07-10
serialization: remove {begin/end}_string() from archives

jeffro256:
Cleanup. Depends on #10888


selsta (approved):

vtnerd (approved):
Nevermind some of the comments from the other review, this fixes at least one issue I raised in the other review.

PR #10890, opened 2026-07-10
wallet2: guard gamma picker against zero-output windows

selsta:


tests/unit_tests/output_selection.cpp
```diff
@@ -261,12 +261,13 @@ TEST(select_outputs, exact_unlock_block)
...
-  std::vector<uint64_t> offsets(std::max(CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE, 1), 0);
-  offsets[0] = 1;
+  // Create a chain of length CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE with one
+  // output in block 0. Since rct_offsets is cumulative, later blocks retain
+  // that output even though they contain no additional outputs.
+  std::vector<uint64_t> offsets(std::max(CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE, 1), 1);
   tools::gamma_picker picker(offsets);
 
-  constexpr size_t MAX_PICK_TRIES = 10;
+  constexpr size_t MAX_PICK_TRIES = 1000;
```
jeffro256, line 270: Why is this changed? `exact_unlock_block_tiny` isn't the test case that sporadically fails

selsta: With the corrected `[1, 1, …]` data, `average_output_time` is finite and `pick()` can legitimately return an invalid pick, so 10 attempts can fail.

jeffro256 (approved):

tests/unit_tests/output_selection.cpp
```diff
@@ -261,12 +261,13 @@ TEST(select_outputs, exact_unlock_block)
...
   bool found_the_one_output = false;
   for (size_t i = 0; i < MAX_PICK_TRIES; ++i)
     if (picker.pick() == 0)
```
iamamyth, line 274: Why does this loop keep running after finding the output?

selsta: Fixed.

iamamyth (approved):

PR #10893, opened 2026-07-10
wallet: derive encrypted payment ID dummy/real status from tx.extra, not cd.dests

sin99xx:
The wallet-CLI confirmation prompt derived a payment ID's dummy/real status from cd.dests[i].is_integrated, while the payment ID that actually lands in tx.extra comes from cd.extra. These two sources can be set inconsistently, so the prompt could display "dummy encrypted payment ID" while a different encrypted payment ID was signed onto the chain.

This changes the source of truth to the plaintext payment_id8 (from cd.extra), which is the value that gets XOR-encrypted into the chain-permanent tx.extra. It also drops the entry.original gate on integrated-form address rendering, since entry.original is proposer-controlled and can be pinned to suppress the integrated display. The same rendering fix is applied to on_describe_transfer in wallet_rpc_server.cpp.

Display-only change, per review scope. cc @selsta

Signed-off-by: sin99xx <sin99xx@users.noreply.github.com>

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -7477,6 +7477,8 @@ bool simple_wallet::accept_loaded_tx(const std::function<size_t()> get_num_txes,
             if (e.is_integrated)
               is_dummy = false;
 
+          CHECK_AND_ASSERT_MES(is_dummy == (payment_id8 == crypto::null_hash8), false, "Bad loaded tx: mismatched payment ID info");
+
           if (is_dummy)
           {
             payment_id_string += std::string("dummy encrypted payment ID");
@@ -7508,7 +7510,7 @@ bool simple_wallet::accept_loaded_tx(const std::function<size_t()> get_num_txes,
     {
       const tx_destination_entry &entry = cd.splitted_dsts[d];
       std::string address, standard_address = get_account_address_as_str(m_wallet->nettype(), entry.is_subaddress, entry.addr);
-      if (has_encrypted_payment_id && !entry.is_subaddress && standard_address != entry.original)
+      if (has_encrypted_payment_id && !entry.is_subaddress)
       {
         address = get_account_integrated_address_as_str(m_wallet->nettype(), entry.addr, payment_id8);
         address += std::string(" (" + standard_address + " with encrypted payment id " + epee::string_tools::pod_to_hex(payment_id8) + ")");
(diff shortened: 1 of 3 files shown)


selsta:
I think these commits can be squashed, both are wallet related.

sin99xx:
Done - squashed into a single commit. Thanks!!

selsta (approved):

src/simplewallet/simplewallet.cpp
```diff
@@ -7471,11 +7471,7 @@ bool simple_wallet::accept_loaded_tx(const std::function<size_t()> get_num_txes,
           if (!payment_id_string.empty())
             payment_id_string += ", ";
 
-          // if none of the addresses are integrated addresses, it's a dummy one
-          bool is_dummy = true;
-          for (const auto &e: cd.dests)
-            if (e.is_integrated)
-              is_dummy = false;
+          const bool is_dummy = (payment_id8 == crypto::null_hash8);
```
jeffro256: Suggestion. Re-add previous `is_dummy` code back in, but check it against `payment_id8 == crypto::null_hash8`: they should be the same AFAIK.

```suggestion
          CHECK_AND_ASSERT_MES(is_dummy == (payment_id8 == crypto::null_hash8), false, "Bad loaded tx: mismatched payment ID info");
```

selsta:
@sin99xx could you check again? It appears you didn't apply @jeffro256 suggestion correctly, and the line is misaligned.

sin99xx:
> Fixed — re-added the original `is_dummy` loop and applied the assert on top of it rather than replacing it, and corrected the indentation. Thanks both.

sin99xx:
@selsta @jeffro256 Ready for another look.

jeffro256 (approved):

selsta:
@sin99xx one more point, could you apply the same fix to wallet api?

```diff
diff --git a/src/wallet/api/unsigned_transaction.cpp b/src/wallet/api/unsigned_transaction.cpp
index ... .. ...
--- a/src/wallet/api/unsigned_transaction.cpp
+++ b/src/wallet/api/unsigned_transaction.cpp
@@ -123,8 +123,14 @@ bool UnsignedTransactionImpl::checkLoadedTx(
         {
           if (!payment_id_string.empty())
             payment_id_string += ", ";
-          payment_id_string = std::string("encrypted payment ID ") + epee::string_tools::pod_to_hex(payment_id8);
-          has_encrypted_payment_id = true;
+          if (payment_id8 == crypto::null_hash8)
+          {
+            payment_id_string += std::string("dummy encrypted payment ID");
+          }
+          else
+          {
+            payment_id_string += std::string("encrypted payment ID ") + epee::string_tools::pod_to_hex(payment_id8);
+            has_encrypted_payment_id = true;
+          }
         }
```

sin99xx:
@selsta wallet API fix applied as requested.

selsta (approved):

PR #10895, opened 2026-07-11
crypto: avoid tree hash copy warning

selsta:
```
/__w/monero/monero/src/crypto/tree-hash.c:89:5: warning: ‘__builtin_memcpy’ writing 64 bytes into a region of size 32 [-Wstringop-overflow=]
   89 |     memcpy(ints, hashes, (2 * cnt - count) * HASH_SIZE);
      |     ^
/__w/monero/monero/src/crypto/tree-hash.c:86:18: note: destination object of size 32 allocated by ‘calloc’
   86 |     char *ints = calloc(cnt, HASH_SIZE);  // zero out as extra protection for using uninitialized mem
      |                  ^~~~~~~~~~~~~~~~~~~~~~
```

`GCC 16.1.1`

### DIFF
--- a/src/crypto/tree-hash.c
+++ b/src/crypto/tree-hash.c
@@ -86,9 +86,12 @@ void tree_hash(const char (*hashes)[HASH_SIZE], size_t count, char *root_hash) {
     char *ints = calloc(cnt, HASH_SIZE);  // zero out as extra protection for using uninitialized mem
     assert(ints);
 
-    memcpy(ints, hashes, (2 * cnt - count) * HASH_SIZE);
+    const size_t initial_count = 2 * cnt - count;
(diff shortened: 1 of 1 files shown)


jeffro256:
I have GCC 16.1.1 2026-06-25 (newest on CachyOS), and I can't reproduce this warning, even with `-Wall -Wextra -Wpointer-arith -Wundef -Wvla -Wwrite-strings -Wformat -Wformat-security -Wstringop-overflow`. Seems like a compiler bug. I would ignore it for now, since the code is standards-conforming and correct AFAICT, instead of making the code slower.

selsta:
I saw the warning on Arch CI, which claims to use GCC 16.1.1, but I'm okay with closing it. If it continues to be an issue it can be silenced in the future.

PR #10897, opened 2026-07-11
tx_pool: remove redundant hashCompare comparator (fix warning)

selsta:
```
/__w/monero/monero/src/cryptonote_core/tx_pool.h:86:20: warning: ‘int memcmp(const void*, const void*, size_t)’ specified bound 32 exceeds source size 0 [-Wstringop-overread]
   86 |       return memcmp(a.data, b.data, sizeof(crypto::hash)) < 0;
```

`GCC 16.1.1`

### DIFF
--- a/src/cryptonote_core/tx_pool.h
+++ b/src/cryptonote_core/tx_pool.h
@@ -78,18 +78,9 @@ namespace cryptonote
     }
   }; 
 
-  class hashCompare
-  {
-  public:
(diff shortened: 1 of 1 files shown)


jpk68 (approved):

vtnerd (approved):

jeffro256:
I recommend just removing `hashCompare` at this point. It's used in exactly one place

selsta:
@jeffro256 updated

jeffro256 (approved):

PR #10901, opened 2026-07-11
wallet_api: set m_password in the recovery creation paths

plowsof:
as a result of testing anonero-fork with fcmp stressnet. @j-berman commit surfaced the issue https://github.com/seraphis-migration/monero/commit/09aa37d1c4f8a390a1e84c377e23cef3056c91f4

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -681,6 +681,7 @@ bool WalletImpl::recoverFromKeysWithPassword(const std::string &path,
         setStatusError(string(tr("failed to generate new wallet: ")) + e.what());
         return false;
     }
+    m_password = password;
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


src/wallet/api/wallet.cpp
```diff
@@ -766,6 +768,7 @@ bool WalletImpl::recover(const std::string &path, const std::string &password, c
     try {
         m_wallet->set_seed_language(old_language);
         m_wallet->generate(path, password, recovery_key, true, false);
+        m_password = password;
```
vtnerd, line 771: Was this supposed to be on line 776 like the others? Presumably `->generate` already stores this.

vtnerd (changes requested):

src/wallet/api/wallet.cpp
plowsof: for convenience generate: https://github.com/monero-project/monero/blob/641e5ca58/src/wallet/wallet2.cpp#L5699

it does't set/store `m_password` which is needed when we store() to a new path

so setting m_password here also happens on the happy path, like the 2 other functions  (if generate would throw for some reason)

selsta (approved):

SNeedlewoods (approved):

jpk68 (approved):

PR #10907, opened 2026-07-13
blockchain: validate alt block supplements against current HF

selsta:


src/cryptonote_core/blockchain.cpp
```diff
@@ -2053,7 +2053,9 @@ bool Blockchain::handle_alternative_block(const block& b, const crypto::hash& id
       return false;
     }
 
-    // Add pool supplement txs to the main mempool with relay_method::block
+    // Add pool supplement txs to the main mempool with relay_method::block. They were verified
+    // for the alt chain above, but must also satisfy the current main chain's pool rules.
```
jeffro256, line 2057: NACK. I think that the concept behind this PR is incorrect: why would an alt chain tx for a previous fork also need to satisfy the current main chain's pool rules?

selsta:
Closing this after discussion with @jeffro256.

PR #10908, opened 2026-07-13
epee: fix build with OpenSSL 4.0

### LINKED ISSUE #10904: Compile error with OpenSSL 4.0
It seems current release isn't compatible with OpenSSL 4.0.
Switching to OpenSSL 3.x fixed this error.

```
/builddir/build/BUILD/monero-0.18.5.1-build/monero-source-v0.18.5.1/contrib/epee/src/net_ssl.cpp: In function ‘bool epee::net_utils::create_rsa_ssl_certificate(EVP_PKEY*&, X509*&)’:
/builddir/build/BUILD/monero-0.18.5.1-build/monero-source-v0.18.5.1/contrib/epee/src/net_ssl.cpp:175:42: error: invalid conversion from ‘const X509_NAME*’ {aka ‘const X509_name_st*’} to ‘X509_NAME*’ {aka ‘X509_name_st*’} [-fpermissive]
  175 |   X509_NAME *name = X509_get_subject_name(cert);
      |                     ~~~~~~~~~~~~~~~~~~~~~^~~~~~
      |                                          |
      |                                          const X509_NAME* {aka const X509_name_st*}
```

Full log:
`https://download.copr.fedorainfracloud.org/results/zpc00/monero/fedora-rawhide-x86_64/10705837-monero/builder-live.log.gz`

weifanglab:
Fixes #10904.

OpenSSL 4.0 makes `X509_get_subject_name()` return a `const X509_NAME *`.

The certificate generation path only passes the returned subject name to `X509_set_issuer_name()`, which accepts a const name.  Preserve the const qualification instead of assigning it to a mutable pointer.

This fixes compilation with OpenSSL 4.0 without changing certificate generation behavior.

Tests:
- Built `gen_ssl_cert` against OpenSSL 3.6.3
- Generated and parsed a self-signed certificate with `monero-gen-ssl-cert`
- `git diff --check`

### DIFF
--- a/contrib/epee/src/net_ssl.cpp
+++ b/contrib/epee/src/net_ssl.cpp
@@ -174,8 +174,7 @@ bool create_rsa_ssl_certificate(EVP_PKEY *&pkey, X509 *&cert)
     X509_free(cert);
     return false;
   }
-  X509_NAME *name = X509_get_subject_name(cert);
-  X509_set_issuer_name(cert, name);
+  X509_set_issuer_name(cert, X509_get_subject_name(cert));
 
   if (X509_sign(cert, pkey, EVP_sha256()) == 0)
   {


selsta:
This breaks CI on Debian 11 which still uses OpenSSL 1.1

weifanglab:
> This breaks CI on Debian 11 which still uses OpenSSL 1.1

Thanks for catching this. OpenSSL 1.1's `X509_set_issuer_name()` still takes an `X509_NAME *`, so explicitly storing the getter result as a const pointer broke the Debian 11 build.

I've removed the explicitly typed local pointer and now pass the result of `X509_get_subject_name()` directly to `X509_set_issuer_name()`. This allows the expression to match both the OpenSSL 1.1 and OpenSSL 4.0 signatures.

I also verified the call against both API signatures and rebuilt `gen_ssl_cert` with OpenSSL 3.6.3.

Please review it again.

iamamyth (approved):
The original code should have done what this patch does, as there's no sense in creating a temporary for a one-off usage.

Failing test is a known flaky test unrelated to the change.

selsta (approved):

selsta:
@weifanglab please also open against release branch

jpk68 (approved):

weifanglab:
> @weifanglab please also open against release branch

@selsta 
https://github.com/monero-project/monero/pull/10913  Thanks.  Please review.

jeffro256 (approved):

PR #10909, opened 2026-07-13
cmake: apply cxa throw wrap flags when linking common

selsta:
Fixes debug build an Arch Linux.


jpk68 (approved):
Tested on Arch; it does indeed fix the build issue.

jeffro256 (approved):

PR #10914, opened 2026-07-14
utils/systemd: update service file

jpk68:
Updates the `systemd` service file to the one used [here](https://docs.getmonero.org/running-node/monerod-systemd/#systemd).

### DIFF
--- a/utils/systemd/monerod.service
+++ b/utils/systemd/monerod.service
@@ -1,20 +1,20 @@
 [Unit]
-Description=Monero Full Node
+Description=Monero Daemon
 After=network-online.target
+Wants=network-online.target
 
 [Service]
-User=monero
-Group=monero
-WorkingDirectory=~
-StateDirectory=monero
-LogsDirectory=monero
-
 Type=simple
-ExecStart=/usr/bin/monerod --config-file /etc/monerod.conf --non-interactive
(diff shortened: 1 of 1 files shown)


utils/systemd/monerod.service
```diff
@@ -1,20 +1,20 @@
...
-Description=Monero Full Node
+Description=Monero Daemon
 After=network-online.target
+Wants=network-online.target
 
 [Service]
+Type=simple
+ExecStart=/usr/local/bin/monerod --config-file /etc/monero/monerod.conf --non-interactive
+
+Restart=on-failure
```
thomasbuilds:
```suggestion
Restart=always
```
Changing the systemd file was my first PR, I knew very little on monero. We should keep `Restart=always` because monerod has several failure paths that exit with code 0 so `Restart=on-failure` would leave the node silently dead in those cases.

Could you open PR on docs to change this there too?

thomasbuilds (changes requested):

utils/systemd/monerod.service
jpk68: Done, thanks.

thomasbuilds (approved):

selsta (approved):

PR #10915, opened 2026-07-14
cryptonote_core: improve block template backlog handling

jeffro256:
Addresses several issues with miner block template:

1. Fixes key image conflict issue due to lazy transaction prefix
  - This line checks whether the next tx candiates has conflicting key images with the rest of the block template: https://github.com/monero-project/monero/blob/0f84863feda75574788e15c9691abecf57bc3f8f/src/cryptonote_core/tx_pool.cpp#L1744
  - The function takes in a transaction prefix and parses the key images from it
  - However, the transaction may or may not actually ever be deserialized, depending on the state of `m_input_cache`: https://github.com/monero-project/monero/blob/0f84863feda75574788e15c9691abecf57bc3f8f/src/cryptonote_core/tx_pool.cpp#L1455-L1466
2. Deduplicates code between `tx_memory_pool::fill_block_template()` and `tx_memory_pool::get_block_template_backlog()`, and thus converging the backlog lists between the 2 functions
3. Saves time creating block templates by skipping transaction blob readings and deserializations for some txs
4. Stops template candidate search after block is reasonably full and we are iterating through txs which lower coinbase reward or exceed size limit by a "smallish" amount. Prevents us from unnecessarily reading thousands of useless tx metadatas under load. May help https://github.com/seraphis-migration/monero/issues/293
5. Fixes block templates when coinbase transactions are > 600 bytes
6. Fills closer to block weight limit for small coinbase transactions
7. Fixes "Invalidating block template cache" spam while popping blocks / during reorgs
8. More accurate "expected block reward" field in RPC

### DIFF
--- a/src/cryptonote_config.h
+++ b/src/cryptonote_config.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2024, The Monero Project
+// Copyright (c) 2014-2026, The Monero Project
 // 
 // All rights reserved.
 // 
--- a/src/cryptonote_config.h
+++ b/src/cryptonote_config.h
@@ -194,6 +194,7 @@
 #define HF_VERSION_BULLETPROOF_PLUS             15
 #define HF_VERSION_VIEW_TAGS                    15
 #define HF_VERSION_2021_SCALING                 15
+#define MAX_HF_VERSION                          16
 
 #define PER_KB_FEE_QUANTIZATION_DECIMALS        8
 #define CRYPTONOTE_SCALING_2021_FEE_ROUNDING_PLACES 2
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -1671,9 +1671,8 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
 
   CHECK_AND_ASSERT_MES(diffic, false, "difficulty overhead.");
 
-  size_t txs_weight;
-  uint64_t fee;
-  if (!m_tx_pool.fill_block_template(b, median_weight, already_generated_coins, txs_weight, fee, expected_reward, b.major_version))
+  std::vector<tx_block_template_backlog_entry> selected_backlog;
+  if (!m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, b.major_version, selected_backlog))
   {
     return false;
   }
@@ -1731,6 +1730,21 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
       ", fee " << fee);
 #endif
 
+  // collect TXID list and total stats
+  uint64_t txs_weight = 0;
+  uint64_t fee = 0;
+  b.tx_hashes.reserve(selected_backlog.size());
+  for (const tx_block_template_backlog_entry &e : selected_backlog)
+  {
+    CHECK_AND_ASSERT_MES(std::numeric_limits<uint64_t>::max() - e.weight >= txs_weight,
+      false, "Overflow in block weight calculation");
+    CHECK_AND_ASSERT_MES(std::numeric_limits<uint64_t>::max() - e.fee >= fee,
+      false, "Overflow in block total fee calculation");
+    txs_weight += e.weight;
+    fee += e.fee;
+    b.tx_hashes.push_back(e.id);
+  }
+
   /*
    two-phase miner transaction generation: we don't know exact block weight until we prepare block, but we don't know reward until we know
    block weight, so first miner transaction generated with fake amount of money, and with phase we know think we know expected block weight
@@ -1738,18 +1752,33 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
   //make blocks coin-base tx looks close to real coinbase tx to get truthful blob weight
   uint8_t hf_version = b.major_version;
   size_t max_outs = hf_version >= 4 ? 1 : 11;
-  bool r = construct_miner_tx(height, median_weight, already_generated_coins, txs_weight, fee, miner_address, b.miner_tx, ex_nonce, max_outs, hf_version);
-  CHECK_AND_ASSERT_MES(r, false, "Failed to construct miner tx, first chance");
-  cumulative_weight = txs_weight + get_transaction_weight(b.miner_tx);
+  cumulative_weight = txs_weight;
 #if defined(DEBUG_CREATE_BLOCK_TEMPLATE)
   MDEBUG("Creating block template: miner tx weight " << get_transaction_weight(b.miner_tx) <<
       ", cumulative weight " << cumulative_weight);
 #endif
   for (size_t try_count = 0; try_count != 10; ++try_count)
   {
-    r = construct_miner_tx(height, median_weight, already_generated_coins, cumulative_weight, fee, miner_address, b.miner_tx, ex_nonce, max_outs, hf_version);
+    const bool miner_tx_made = construct_miner_tx(height, median_weight, already_generated_coins,
+      cumulative_weight, fee, miner_address, b.miner_tx, ex_nonce, max_outs, hf_version);
 
-    CHECK_AND_ASSERT_MES(r, false, "Failed to construct miner tx, second chance");
+    // On miner tx creation failure, if we have txs to pop, try that...
+    if (!miner_tx_made)
+    {
+      CHECK_AND_ASSERT_MES(!b.tx_hashes.empty(),
+        false, "Failed to create miner transaction, but have no non-miner txs to pop");
+      b.tx_hashes.pop_back();
+      assert(selected_backlog.size() == b.tx_hashes.size());
+      const tx_block_template_backlog_entry &backlog_entry = selected_backlog.back();
+      assert(txs_weight >= backlog_entry.weight);
+      txs_weight -= backlog_entry.weight;
+      assert(fee >= backlog_entry.fee);
+      fee -= backlog_entry.fee;
+      selected_backlog.pop_back();
+      cumulative_weight = txs_weight;
+      continue;
+    }
+
     size_t coinbase_weight = get_transaction_weight(b.miner_tx);
     if (coinbase_weight > cumulative_weight - txs_weight)
     {
@@ -1791,6 +1820,7 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
         ", cumulative weight " << cumulative_weight << " is now good");
 #endif
 
+    expected_reward = get_outs_money_amount(b.miner_tx);
     if (!from_block)
       cache_block_template(b, miner_address, ex_nonce, diffic, height, expected_reward, cumulative_weight, seed_height, seed_hash, pool_cookie);
     return true;
@@ -1823,9 +1853,7 @@ bool Blockchain::get_miner_data(uint8_t& major_version, uint64_t& height, crypto
   median_weight = m_current_block_cumul_weight_median;
   already_generated_coins = m_db->get_block_already_generated_coins(height - 1);
 
-  m_tx_pool.get_block_template_backlog(tx_backlog);
-
-  return true;
+  return m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, major_version, tx_backlog);
 }
 //------------------------------------------------------------------
 // for an alternate chain, get the timestamps from the main chain to complete
@@ -5562,8 +5590,8 @@ bool Blockchain::for_all_outputs(uint64_t amount, std::function<bool(uint64_t he
 
 void Blockchain::invalidate_block_template_cache()
 {
-  MDEBUG("Invalidating block template cache");
-  m_btc_valid = false;
+  if (std::exchange(m_btc_valid, false))
+    MDEBUG("Invalidating block template cache");
 }
 
 void Blockchain::cache_block_template(const block &b, const cryptonote::account_public_address &address, const blobdata &nonce, const difficulty_type &diff, uint64_t height, uint64_t expected_reward, uint64_t cumulative_weight, uint64_t seed_height, const crypto::hash &seed_hash, uint64_t pool_cookie)
@@ -5592,7 +5620,14 @@ void Blockchain::send_miner_notifications(uint64_t height, const crypto::hash &s
   const uint64_t median_weight = m_current_block_cumul_weight_median;
 
   std::vector<tx_block_template_backlog_entry> tx_backlog;
-  m_tx_pool.get_block_template_backlog(tx_backlog);
+  if (!m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, major_version, tx_backlog))
+  {
+    MERROR("Could not retreive mempool's block template backlog for block " << height
+      << ": median weight " << median_weight << " already generated coins "
+      << already_generated_coins << " major version " << major_version
+      << ". Miner notifications will not be sent.");
+    return;
+  }
 
   for (const auto& notifier : m_miner_notifiers)
   {
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -86,6 +86,9 @@ DISABLE_VS_WARNINGS(4267)
 // used to overestimate the block reward when estimating a per kB to use
 #define BLOCK_REWARD_OVERESTIMATE (10 * 1000000000000)
 
+/// @brief Parameter for block template backlog fetching: targets total weight 112.5% of the block's limit
+#define DEFAULT_RPC_TX_BACKLOG_OVERPICK 0.125f
+
 //------------------------------------------------------------------
 Blockchain::Blockchain(tx_memory_pool& tx_pool) :
   m_db(), m_tx_pool(tx_pool), m_hardfork(NULL), m_timestamps_and_difficulties_height(0), m_reset_timestamps_and_difficulties_height(true), m_current_block_cumul_weight_limit(0), m_current_block_cumul_weight_median(0),
@@ -1672,7 +1675,8 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
   CHECK_AND_ASSERT_MES(diffic, false, "difficulty overhead.");
 
   std::vector<tx_block_template_backlog_entry> selected_backlog;
-  if (!m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, b.major_version, selected_backlog))
+  if (!m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins,
+    b.major_version, /*overpick=*/0, selected_backlog))
   {
     return false;
   }
@@ -1853,7 +1857,8 @@ bool Blockchain::get_miner_data(uint8_t& major_version, uint64_t& height, crypto
   median_weight = m_current_block_cumul_weight_median;
   already_generated_coins = m_db->get_block_already_generated_coins(height - 1);
 
-  return m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, major_version, tx_backlog);
+  return m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, major_version,
+    DEFAULT_RPC_TX_BACKLOG_OVERPICK, tx_backlog);
 }
 //------------------------------------------------------------------
 // for an alternate chain, get the timestamps from the main chain to complete
@@ -5618,14 +5623,16 @@ void Blockchain::send_miner_notifications(uint64_t height, const crypto::hash &s
   const uint8_t major_version = m_hardfork->get_ideal_version(height);
   const difficulty_type diff = get_difficulty_for_next_block();
   const uint64_t median_weight = m_current_block_cumul_weight_median;
+  const float overpick = DEFAULT_RPC_TX_BACKLOG_OVERPICK;
 
   std::vector<tx_block_template_backlog_entry> tx_backlog;
-  if (!m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, major_version, tx_backlog))
+  if (!m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, major_version,
+    overpick, tx_backlog))
   {
     MERROR("Could not retreive mempool's block template backlog for block " << height
       << ": median weight " << median_weight << " already generated coins "
       << already_generated_coins << " major version " << major_version
-      << ". Miner notifications will not be sent.");
+      << " overpick ratio " << overpick << ". Miner notifications will not be sent.");
     return;
   }
 
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -1676,7 +1676,7 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
 
   std::vector<tx_block_template_backlog_entry> selected_backlog;
   if (!m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins,
-    b.major_version, /*overpick=*/0, selected_backlog))
+    b.major_version, /*overpick=*/0, /*include_sensitive=*/true, selected_backlog))
   {
     return false;
   }
@@ -1771,8 +1771,8 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
     {
       CHECK_AND_ASSERT_MES(!b.tx_hashes.empty(),
         false, "Failed to create miner transaction, but have no non-miner txs to pop");
-      b.tx_hashes.pop_back();
       assert(selected_backlog.size() == b.tx_hashes.size());
+      b.tx_hashes.pop_back();
       const tx_block_template_backlog_entry &backlog_entry = selected_backlog.back();
       assert(txs_weight >= backlog_entry.weight);
       txs_weight -= backlog_entry.weight;
@@ -1858,7 +1858,7 @@ bool Blockchain::get_miner_data(uint8_t& major_version, uint64_t& height, crypto
   already_generated_coins = m_db->get_block_already_generated_coins(height - 1);
 
   return m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, major_version,
-    DEFAULT_RPC_TX_BACKLOG_OVERPICK, tx_backlog);
+    DEFAULT_RPC_TX_BACKLOG_OVERPICK, /*include_sensitive=*/false, tx_backlog);
 }
 //------------------------------------------------------------------
 // for an alternate chain, get the timestamps from the main chain to complete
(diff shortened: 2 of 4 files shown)


src/cryptonote_core/tx_pool.cpp
```diff
@@ -1091,67 +1079,6 @@ namespace cryptonote
...
-    uint64_t total_weight = 0;
-
-    // First get everything from the mempool, filter it later
-    m_blockchain.for_all_txpool_txes([&tmp, &total_weight](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref*){
-      tmp.emplace_back(tx_block_template_backlog_entry{txid, meta.weight, meta.fee});
-      total_weight += meta.weight;
-      return true;
-    }, false, include_sensitive ? relay_category::all : relay_category::broadcasted);
-
-    // Limit backlog to 112.5% of current median weight. This is enough to mine a full block with the optimal block reward
```
SChernykh, line 1109: `get_block_template_backlog` wasn't the verbatim copy of the "fill the block for mining". It returned more transactions on purpose, so that the consumers (P2Pool) could pick from that list to create their own block templates. Is this behavior kept intact?

jeffro256: No it is not. If I may, is there a benefit to over-picking and having p2pool filter, versus simply doing it how p2pool would do it here?

jeffro256:
Pushed review from @selsta

src/cryptonote_core/tx_pool.cpp
SChernykh: Originally, that function returned the full mempool - it was written for P2Pool to pick transactions on its own later. Then it was reduced to 112.5% of the median size to better handle very big mempools.

SChernykh: TLDR it solves a different problem than `fill_block_template`

jeffro256: Okay support for over-picking was re-added in latest.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1669,55 +1586,42 @@ namespace cryptonote
...
         LOG_PRINT_L2("  would exceed maximum block weight");
-        continue;
+        if (selected_backlog.empty() || total_weight + reference_tx_weight < max_total_weight)
+          continue; // there is much room, so even though this tx would breach the limit, maybe next txs will fit
+        else
+          break; // stop searching once the block is reasonably full, since next txs will have worse fee per weight
       }
 
-      // start using the optimal filling algorithm from v5
-      if (version >= 5)
```
jeffro256, line 1680: Note: After this PR, we use optimal filling for before v5 as well. Does this introduce a bug? So far I can't figure out why optimal wasn't used pre v5. Only reason I can think of is for indistinguishability.

src/cryptonote_core/tx_pool.cpp
jeffro256: May I get a re-review, please, @SChernykh?

src/cryptonote_core/blockchain.cpp
```diff
@@ -1731,25 +1734,55 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
...
+    const bool miner_tx_made = construct_miner_tx(height, median_weight, already_generated_coins,
+      cumulative_weight, fee, miner_address, b.miner_tx, ex_nonce, max_outs, hf_version);
+
+    // On miner tx creation failure, if we have txs to pop, try that...
+    if (!miner_tx_made)
+    {
+      CHECK_AND_ASSERT_MES(!b.tx_hashes.empty(),
+        false, "Failed to create miner transaction, but have no non-miner txs to pop");
+      b.tx_hashes.pop_back();
+      assert(selected_backlog.size() == b.tx_hashes.size());
```
SChernykh, line 1774: Shouldn't this assert always trigger because `b.tx_hashes.size() == selected_backlog.size() - 1` at this point - `selected_backlog.pop_back()` is done a few lines later?

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1647,16 +1561,19 @@ namespace cryptonote
...
+          MERROR("  failed to find tx meta: " << txid << " (will only print once)");
         warned = true;
         continue;
       }
-      LOG_PRINT_L2("Considering " << sorted_it->get_right() << ", weight " << meta.weight << ", current block weight " << total_weight << "/" << max_total_weight << ", current coinbase " << print_money(best_coinbase) << ", relay method " << (unsigned)meta.get_relay_method());
+      LOG_PRINT_L2("Considering " << txid << ", weight " << meta.weight << ", current block weight "
+        << total_weight << "/" << max_total_weight << ", current coinbase "
+        << print_money(best_coinbase) << ", relay method " << (unsigned)meta.get_relay_method());
 
       if (!meta.matches(relay_category::legacy) && !(m_mine_stem_txes && meta.get_relay_method() == relay_method::stem))
```
selsta: `relay_category::legacy` includes `relay_method::none`, used for `do_not_relay` transactions. The function is now called by `get_miner_data` without the old `include_sensitive=false` filter.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1669,55 +1586,42 @@ namespace cryptonote
         continue;
       }
 
-      // Can not exceed maximum block weight
+      // Cannot exceed max weight limit we set (may be greater than the block weight limit, depending on `overpick`)
```
selsta, line 1599: Both early break decisions (exceeding the weight target and lowering the reward) happen before `is_transaction_ready_to_go()`. Is this intentional?

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1669,55 +1586,42 @@ namespace cryptonote
...
         uint64_t block_reward;
         if(!get_block_reward(median_weight, total_weight + meta.weight, already_generated_coins, block_reward, version))
         {
           LOG_PRINT_L2("  would exceed maximum block weight");
-          continue;
+          break;
         }
         coinbase = block_reward + fee + meta.fee;
-        if (coinbase < template_accept_threshold(best_coinbase))
+        if (coinbase < best_coinbase)
```
selsta, line 1619: Transactions are sorted by fee divided by weight, but profit from adding the transaction is also dependent on absolute transaction weight because the block reward penalty is quadratic. The old code continued scanning, changing this to break can omit profitable transactions.

src/cryptonote_core/tx_pool.cpp
jeffro256: Yes, because the break statement for the weight check only gets called if the template isn't empty and if there is a decent amount of room left. I could make the other break statements conditional on that same condition.

src/cryptonote_core/tx_pool.cpp
jeffro256: It could, for a particularly large tx. Like the last comment, I should probably add the near-full condition to the break

src/cryptonote_core/tx_pool.cpp
jeffro256: Done in latest

src/cryptonote_core/blockchain.cpp
jeffro256: Done in latest

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1634,31 +1542,50 @@ namespace cryptonote
       return false;
     }
 
+    static_assert(MAX_HF_VERSION == 16, "Check below weights and limits");
+    const relay_category category = include_sensitive ? relay_category::legacy : relay_category::broadcasted;
+    const size_t max_total_weight_pre_v5 = (130 * median_weight) / 100;
+    const size_t max_total_weight_v5 = 2 * median_weight;
+    const size_t consensus_max_total_weight = version >= 5 ? max_total_weight_v5 : max_total_weight_pre_v5;
+    const size_t max_total_weight = consensus_max_total_weight
+      + static_cast<size_t>(overpick * consensus_max_total_weight);
```
SChernykh, line 1551: Before the change, overpick was 112.5% of the median weight, now it's 112.5% of consensus max weight which is the double of median weight. To preserver the old behavior, the logic should be:
- overpick is 0 -> consensus_max_total_weight
- overpick is not 0 -> median_weight * (1 + overpick)

thomasbuilds: Agree with SChernykh here. The old backlog was capped at 112.5% of the median weight; this is 112.5% of the hard limit, i.e. 225% of the median. Measured on regtest with the same 300-tx pool (median 300000): `get_miner_data` returns 56 txs / 341208 bytes (1.137x median) on master and 110 txs / 670230 bytes (2.234x median) here. p2pool's own filling stops near the median (past it the penalty beats the fees almost always), so the extra half is payload every `--zmq-pub` node serializes on every block for nothing.

```suggestion
    const size_t max_total_weight = overpick > 0
      ? median_weight + static_cast<size_t>(overpick * median_weight)
      : consensus_max_total_weight;
```

src/cryptonote_core/tx_pool.h
```diff
@@ -259,17 +260,38 @@ namespace cryptonote
...
+     * highest-paying / oldest transactions will be returned - but enough for the miner to create a
+     * full block. Transactions which do not pass validation rules for the next block are also
+     * filtered out. However, @warning, `selected_backlog` may also contain *too many entries*, even
+     * when `overpick_ratio` == 0, such that block weight limit is surpassed after including the
+     * coinbase transaction. In this case, it is a good idea to pop candidates from the back of the
+     * list until the weight is within the limit.
+     *
+     * The target limit is calculated as (1 + `overpick`) * theoretical block weight limit. So for
+     * example, an `overpick` value of 0.1 will try to fill `selected_backlog` so that it's total
+     * weight is 110% of the block weight limit. When `overpick` > 0, entries in `selected_backlog`
```
thomasbuilds: With the formula above:

```suggestion
     * The target limit is (1 + `overpick`) * median weight when `overpick` > 0, and the block weight
     * limit otherwise. So for example, an `overpick` value of 0.1 will try to fill `selected_backlog`
     * so that its total weight is 110% of the median weight. When `overpick` > 0, entries in `selected_backlog`
```

src/cryptonote_core/blockchain.cpp
```diff
@@ -86,6 +86,9 @@ DISABLE_VS_WARNINGS(4267)
 // used to overestimate the block reward when estimating a per kB to use
 #define BLOCK_REWARD_OVERESTIMATE (10 * 1000000000000)
 
+/// @brief Parameter for block template backlog fetching: targets total weight 112.5% of the block's limit
```
thomasbuilds:
```suggestion
/// @brief Parameter for block template backlog fetching: targets total weight 112.5% of the median weight
```

src/cryptonote_core/blockchain.cpp
```diff
@@ -1731,25 +1734,55 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
...
+      b.tx_hashes.pop_back();
+      const tx_block_template_backlog_entry &backlog_entry = selected_backlog.back();
+      assert(txs_weight >= backlog_entry.weight);
+      txs_weight -= backlog_entry.weight;
+      assert(fee >= backlog_entry.fee);
+      fee -= backlog_entry.fee;
+      selected_backlog.pop_back();
+      cumulative_weight = txs_weight;
+      continue;
+    }
```
thomasbuilds: Without the 600-byte reserve the backlog can fill the limit to within less than a coinbase, and then this call fails through `get_block_reward()`, which logs `MERROR("Block cumulative weight is too big ...")`, followed by `LOG_PRINT_L0("Block is too big")` from `construct_miner_tx()`. That is a normal path now: with a backlog tuned to leave a 39-byte gap (`gen_template_full_block`), every template build on this head logs that pair and then recovers to the same 598533-byte block that master builds silently. While blocks are full that is one ERROR per template build for the operator to wonder about.

The loop also treats every `construct_miner_tx()` failure as "too big": with a failure unrelated to weight (e.g. an extra nonce above `TX_EXTRA_NONCE_MAX_COUNT`, which the RPC rejects but `core::get_block_template()` does not) it drops five txs and then fails with "Failed to create_block_template with 10 tries", where master failed at once.

Checking the weight explicitly, with the same bound `get_block_reward()` applies, avoids both (the `assert`s are compiled out in release builds, so they don't protect the pops either):

```suggestion
    // drop txs until the coinbase fits under the hard limit, so that construct_miner_tx() cannot
    // fail for that reason (it logs an error when it does)
    const uint64_t effective_median = std::max<uint64_t>(median_weight, get_min_block_weight(hf_version));
    if (cumulative_weight > 2 * effective_median)
    {
      CHECK_AND_ASSERT_MES(!b.tx_hashes.empty(),
        false, "Coinbase transaction alone exceeds the block weight limit");
      CHECK_AND_ASSERT_MES(selected_backlog.size() == b.tx_hashes.size(), false, "Backlog and tx hashes out of sync");
      b.tx_hashes.pop_back();
      const tx_block_template_backlog_entry &backlog_entry = selected_backlog.back();
      CHECK_AND_ASSERT_MES(txs_weight >= backlog_entry.weight && fee >= backlog_entry.fee, false, "Backlog totals out of sync");
      txs_weight -= backlog_entry.weight;
      fee -= backlog_entry.fee;
      selected_backlog.pop_back();
      cumulative_weight = txs_weight;
      continue;
    }

    CHECK_AND_ASSERT_MES(construct_miner_tx(height, median_weight, already_generated_coins,
      cumulative_weight, fee, miner_address, b.miner_tx, ex_nonce, max_outs, hf_version),
      false, "Failed to construct miner tx");
```

src/cryptonote_core/tx_pool.cpp
```diff
@@ -478,24 +473,33 @@ namespace cryptonote
...
-      std::unordered_set<crypto::hash>& kei_image_set = m_spent_key_images[txin.k_image];
+      auto txs_for_ki = m_spent_key_images.left.equal_range(txin.k_image);
+
+      //! @TODO: does a linear search to find a matching tx for given KI. Fine in 99% of cases, but may be DDoS-able
+      bool this_tx_for_ki_existed = false;
+      size_t n_txs_for_ki = 0;
+      for (auto it = txs_for_ki.first; it != txs_for_ki.second; ++it, ++n_txs_for_ki)
+        if (it->second == id)
+          this_tx_for_ki_existed = true;
+      const bool other_tx_for_ki_existed = n_txs_for_ki > 1 || (n_txs_for_ki == 1 && !this_tx_for_ki_existed);
```
thomasbuilds: The old container answered "is this txid already recorded for this key image" in O(1); this loop is linear in the number of txs sharing the key image, which only kept-by-block txs can do, so an alt block carrying N double-spends of one key image costs O(N^2) here (as the TODO says). `m_spent_key_images.right.equal_range(id)` gives this tx's own entries in O(vin), and `left.count(txin.k_image)` the multiplicity, which is all the two checks below need. Low priority since it needs valid PoW on an alt block.

src/cryptonote_config.h
```diff
@@ -194,6 +194,7 @@
 #define HF_VERSION_BULLETPROOF_PLUS             15
 #define HF_VERSION_VIEW_TAGS                    15
 #define HF_VERSION_2021_SCALING                 15
+#define MAX_HF_VERSION                          16
```
thomasbuilds: Nit: a hand-maintained constant, so the `static_assert(MAX_HF_VERSION == 16, ...)` in `get_block_template_backlog()` only fires if whoever adds v17 remembers to bump this. A comment next to the mainnet fork table (or deriving it from `mainnet_hard_forks`) would make the reminder work.

src/cryptonote_core/blockchain.cpp
```diff
@@ -5590,9 +5623,18 @@ void Blockchain::send_miner_notifications(uint64_t height, const crypto::hash &s
...
   const difficulty_type diff = get_difficulty_for_next_block();
   const uint64_t median_weight = m_current_block_cumul_weight_median;
+  const float overpick = DEFAULT_RPC_TX_BACKLOG_OVERPICK;
 
   std::vector<tx_block_template_backlog_entry> tx_backlog;
-  m_tx_pool.get_block_template_backlog(tx_backlog);
+  if (!m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, major_version,
+    overpick, /*include_sensitive=*/false, tx_backlog))
+  {
+    MERROR("Could not retreive mempool's block template backlog for block " << height
```
thomasbuilds:
```suggestion
    MERROR("Could not retrieve mempool's block template backlog for block " << height
```

src/cryptonote_core/blockchain.cpp
thomasbuilds: When the backlog leaves less room than the coinbase, this call fails through `get_block_reward()`, which logs `MERROR("Block cumulative weight is too big ...")` (plus a WARNING from `construct_miner_tx()`) before the pop below recovers: an ERROR on every template build while blocks are full. The loop also pops on any `construct_miner_tx()` failure, e.g. an extra nonce over 255 bytes drops five txs before failing.

Better to pop while `cumulative_weight > 2 * std::max<uint64_t>(median_weight, get_min_block_weight(hf_version))` before the call (keeping the coinbase estimate in `cumulative_weight`) and treat a failure of the call itself as fatal. Tested that variant: same templates, no log lines.

src/cryptonote_core/tx_pool.cpp
thomasbuilds: Linear in the number of txs sharing the key image, so kept-by-block double spends make insertion quadratic: 5k / 20k / 40k txs on one key image take 75 ms / 1.3 s / 5.9 s here against 0.5 / 2 / 6 ms with the old map, and `init()` pays it again at startup. `right.equal_range(id)` gives this tx's own entries in O(vin), and the "another txid uses it" check can stop at the first entry != `id`.

src/cryptonote_core/tx_pool.cpp
thomasbuilds: p2pool swaps its mempool with this backlog on every block (`handle_miner_data()`) and its own selection goes up to 2x median, so the larger cap does have a use in high-fee regimes; the cost is about 100 bytes per extra entry.

src/cryptonote_core/blockchain.cpp
thomasbuilds:
```suggestion
    MERROR("Could not retrieve mempool's block template backlog for block " << height
```

src/cryptonote_core/blockchain.cpp
thomasbuilds: When the backlog leaves less room than the coinbase, this call fails through `get_block_reward()`, which logs `MERROR("Block cumulative weight is too big ...")` (plus a WARNING from `construct_miner_tx()`) before the pop below recovers: an ERROR on every template build while blocks are full. The loop also pops on any `construct_miner_tx()` failure, e.g. an extra nonce over 255 bytes drops five txs before failing.

Better to pop while `cumulative_weight > 2 * std::max<uint64_t>(median_weight, get_min_block_weight(hf_version))` before the call (keeping the coinbase estimate in `cumulative_weight`) and treat a failure of the call itself as fatal. Tested that variant: same templates, no log lines.

src/cryptonote_core/tx_pool.cpp
thomasbuilds: Linear in the number of txs sharing the key image, so kept-by-block double spends make insertion quadratic: 5k / 20k / 40k txs on one key image take 75 ms / 1.3 s / 5.9 s here against 0.5 / 2 / 6 ms with the old map, and `init()` pays it again at startup. `right.equal_range(id)` gives this tx's own entries in O(vin), and the "another txid uses it" check can stop at the first entry != `id`.

src/cryptonote_core/tx_pool.cpp
thomasbuilds: p2pool swaps its mempool with this backlog on every block (`handle_miner_data()`) and its own selection goes up to 2x median, so the larger cap does have a use in high-fee regimes; the cost is about 100 bytes per extra entry.

src/cryptonote_core/blockchain.cpp
thomasbuilds:
```suggestion
    MERROR("Could not retrieve mempool's block template backlog for block " << height
```

src/cryptonote_core/blockchain.cpp
```diff
@@ -1731,25 +1734,55 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
...
    block weight, so first miner transaction generated with fake amount of money, and with phase we know think we know expected block weight
    */
   //make blocks coin-base tx looks close to real coinbase tx to get truthful blob weight
   uint8_t hf_version = b.major_version;
   size_t max_outs = hf_version >= 4 ? 1 : 11;
-  bool r = construct_miner_tx(height, median_weight, already_generated_coins, txs_weight, fee, miner_address, b.miner_tx, ex_nonce, max_outs, hf_version);
-  CHECK_AND_ASSERT_MES(r, false, "Failed to construct miner tx, first chance");
-  cumulative_weight = txs_weight + get_transaction_weight(b.miner_tx);
+  cumulative_weight = txs_weight;
 #if defined(DEBUG_CREATE_BLOCK_TEMPLATE)
```
thomasbuilds: All the `DEBUG_CREATE_BLOCK_TEMPLATE` blocks in this function are dead (`m_tx_pool.m_transactions` no longer exists), and the one after the backlog call now also uses `txs_weight` and `fee` before their declaration. Worth deleting them here.

src/cryptonote_core/blockchain.cpp
thomasbuilds: When the backlog leaves less room than the coinbase, this call fails through `get_block_reward()`, which logs `MERROR("Block cumulative weight is too big ...")` (plus a WARNING from `construct_miner_tx()`) before the pop below recovers. That is a normal path now: while blocks are full it happens on every template build whose leftover is smaller than the coinbase. The loop also reads any `construct_miner_tx()` failure as "too big": a failure unrelated to weight (e.g. an extra nonce over 255 bytes, which `core::get_block_template()` does not reject) pops a tx on each of the ten tries and then fails, where master failed at once.

Better to pop while `cumulative_weight > 2 * std::max<uint64_t>(median_weight, get_min_block_weight(hf_version))` before the call (the bound `get_block_reward()` applies; keep the coinbase estimate in `cumulative_weight`) and treat a failure of the call itself as fatal. Tested that variant against a backlog leaving a 39-byte gap: same templates, no log lines.

src/cryptonote_core/tx_pool.cpp
thomasbuilds: Linear in the number of txs sharing the key image, so kept-by-block double spends make insertion quadratic: 5k / 20k / 40k txs on one key image take 75 ms / 1.3 s / 5.9 s here against 0.5 / 2 / 6 ms with the old map, and `init()` pays it again at startup. `right.equal_range(id)` gives this tx's own entries in O(vin), and the "another txid uses it" check can stop at the first entry != `id`.

src/cryptonote_core/tx_pool.cpp
thomasbuilds: p2pool replaces its mempool with this backlog on every block (`handle_miner_data()` -> `Mempool::swap_transactions()`), and its own selection goes up to 2x median when fees justify it, so the larger cap does have a use in high-fee regimes; the cost is about 100 bytes of JSON per extra entry.

src/cryptonote_core/blockchain.cpp
thomasbuilds:
```suggestion
    MERROR("Could not retrieve mempool's block template backlog for block " << height
```

src/cryptonote_core/blockchain.cpp
thomasbuilds: These `DEBUG_CREATE_BLOCK_TEMPLATE` blocks are never compiled (nothing defines the macro, see the FIXME above the function) and would not compile: the one after the backlog call references `m_tx_pool.m_transactions`, which no longer exists, and after this change uses `txs_weight` and `fee` before their declaration. Worth deleting them here.

src/cryptonote_core/blockchain.cpp
thomasbuilds, line 1784: When the backlog leaves less room than the coinbase, this call fails through `get_block_reward()`, which logs `MERROR("Block cumulative weight is too big ...")` (plus a WARNING from `construct_miner_tx()`) before the pop recovers: while blocks are full, an ERROR on every template build whose leftover is smaller than the coinbase. The loop also reads any `construct_miner_tx()` failure as "too big": a failure unrelated to weight (e.g. an extra nonce over 255 bytes, which `core::get_block_template()` does not reject) pops a tx on each of the ten tries and then fails, where master failed at once.

Better to pop while `cumulative_weight > 2 * std::max<uint64_t>(median_weight, get_min_block_weight(hf_version))` before the call (the bound `get_block_reward()` applies; keep the coinbase estimate in `cumulative_weight`) and treat a failure of the call itself as fatal.

src/cryptonote_core/tx_pool.cpp
thomasbuilds, line 484: Linear in the number of txs sharing the key image, so kept-by-block double spends make insertion quadratic: 5k / 20k / 40k txs on one key image take 75 ms / 1.3 s / 5.9 s here against 0.5 / 2 / 6 ms with the old map, and `init()` pays it again at startup. `right.equal_range(id)` gives this tx's own entries in O(vin), and the "another txid uses it" check can stop at the first entry != `id`.

src/cryptonote_core/tx_pool.cpp
thomasbuilds, line 1551: p2pool replaces its mempool with this backlog on every block (`handle_miner_data()` -> `Mempool::swap_transactions()`), and its own selection goes up to 2x median when fees justify it, so the larger cap does have a use in high-fee regimes; the cost is about 100 bytes of JSON per extra entry.

src/cryptonote_core/blockchain.cpp
thomasbuilds, line 5632:
```suggestion
    MERROR("Could not retrieve mempool's block template backlog for block " << height
```

src/cryptonote_core/blockchain.cpp
thomasbuilds, line 1760: These `DEBUG_CREATE_BLOCK_TEMPLATE` blocks are never compiled (nothing defines the macro, see the FIXME above the function) and would not compile: the one after the backlog call references `m_tx_pool.m_transactions`, which no longer exists, and after this change uses `txs_weight` and `fee` before their declaration. Worth deleting them here.

PR #10916, opened 2026-07-14
cryptonote_core: cache NIC verified HF version in mempool

### LINKED ISSUE #10142: monerod revalidates the whole tx-pool on HFs
When adding a block and the HF changes the tx-pool is re-validated:  

Which would freeze up the node for a while with a big tx-pool. Cuprate caches when a tx was validated with the tx in the pool: https://github.com/Cuprate/cuprate/blob/main/types/src/transaction_verification_data.rs#L35 so we don't need to do tx-pool re-validation at start up/when a HF changes, we only re-validate a single tx if the current blockchain state requires it compared to the cached verification. 

monerod could probably do something similar, just caching the HF each tx was validated at and doing `ver_non_input_consensus` added in #9135 if the HF has changed when adding the tx to the blockchain, skipping the need for re-validation of the txpool.

_Originally posted by @Boog900 in https://github.com/monero-project/monero/issues/9404#issuecomment-2679316490_

jeffro256:
Stores a 1 byte field in the mempool meta, `nic_verified_hf_version` which
specifies for which HF version this tx has passed `ver_non_input_consensus()`.
This is checked while handling both alt chains and main chains, and thus allows
us to skip the expensive full mempool revalidation on fork change.

Resolves #10142

Depends on #10915

### DIFF
--- a/src/blockchain_db/blockchain_db.h
+++ b/src/blockchain_db/blockchain_db.h
@@ -172,9 +172,12 @@ struct txpool_tx_meta_t
   uint8_t is_forwarding: 1;
   uint8_t bf_padding: 3;
 
-  uint8_t padding[44]; // til 160 bytes
+  uint8_t padding[43]; // til 159 bytes
 
-  // If non-null, this verification ID is set for this tx only when some mixring passed ver_input_proofs_rings()
+  //! If non-zero, this is set to the HF major version for which this tx passes ver_non_input_consensus()
+  uint8_t nic_verified_hf_version;
+
+  //! If non-null, this verification ID is set for this tx only when some mixring passed ver_input_proofs_rings()
   crypto::hash valid_input_verification_id;
 
   // 192 bytes total
@@ -192,6 +195,7 @@ struct txpool_tx_meta_t
   }
 };
 static_assert(sizeof(txpool_tx_meta_t) == 192, "possible DB migration needed for changes to txpool_tx_meta_t");
+static_assert(offsetof(txpool_tx_meta_t, nic_verified_hf_version) == 159, "NIC verified HF version wrong offset");
 static_assert(offsetof(txpool_tx_meta_t, valid_input_verification_id) == 160, "verif ID wrong alignment");
 
 #define DBF_SAFE       1
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -1671,9 +1671,8 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
 
   CHECK_AND_ASSERT_MES(diffic, false, "difficulty overhead.");
 
-  size_t txs_weight;
-  uint64_t fee;
-  if (!m_tx_pool.fill_block_template(b, median_weight, already_generated_coins, txs_weight, fee, expected_reward, b.major_version))
+  std::vector<tx_block_template_backlog_entry> selected_backlog;
+  if (!m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, b.major_version, selected_backlog))
   {
     return false;
   }
@@ -1731,6 +1730,21 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
       ", fee " << fee);
 #endif
 
+  // collect TXID list and total stats
+  uint64_t txs_weight = 0;
+  uint64_t fee = 0;
+  b.tx_hashes.reserve(selected_backlog.size());
+  for (const tx_block_template_backlog_entry &e : selected_backlog)
+  {
+    CHECK_AND_ASSERT_MES(std::numeric_limits<uint64_t>::max() - e.weight >= txs_weight,
+      false, "Overflow in block weight calculation");
+    CHECK_AND_ASSERT_MES(std::numeric_limits<uint64_t>::max() - e.fee >= fee,
+      false, "Overflow in block total fee calculation");
+    txs_weight += e.weight;
+    fee += e.fee;
+    b.tx_hashes.push_back(e.id);
+  }
+
   /*
    two-phase miner transaction generation: we don't know exact block weight until we prepare block, but we don't know reward until we know
    block weight, so first miner transaction generated with fake amount of money, and with phase we know think we know expected block weight
@@ -1738,18 +1752,33 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
   //make blocks coin-base tx looks close to real coinbase tx to get truthful blob weight
   uint8_t hf_version = b.major_version;
   size_t max_outs = hf_version >= 4 ? 1 : 11;
-  bool r = construct_miner_tx(height, median_weight, already_generated_coins, txs_weight, fee, miner_address, b.miner_tx, ex_nonce, max_outs, hf_version);
-  CHECK_AND_ASSERT_MES(r, false, "Failed to construct miner tx, first chance");
-  cumulative_weight = txs_weight + get_transaction_weight(b.miner_tx);
+  cumulative_weight = txs_weight;
 #if defined(DEBUG_CREATE_BLOCK_TEMPLATE)
   MDEBUG("Creating block template: miner tx weight " << get_transaction_weight(b.miner_tx) <<
       ", cumulative weight " << cumulative_weight);
 #endif
   for (size_t try_count = 0; try_count != 10; ++try_count)
   {
-    r = construct_miner_tx(height, median_weight, already_generated_coins, cumulative_weight, fee, miner_address, b.miner_tx, ex_nonce, max_outs, hf_version);
+    const bool miner_tx_made = construct_miner_tx(height, median_weight, already_generated_coins,
+      cumulative_weight, fee, miner_address, b.miner_tx, ex_nonce, max_outs, hf_version);
 
-    CHECK_AND_ASSERT_MES(r, false, "Failed to construct miner tx, second chance");
+    // On miner tx creation failure, if we have txs to pop, try that...
+    if (!miner_tx_made)
+    {
+      CHECK_AND_ASSERT_MES(!b.tx_hashes.empty(),
+        false, "Failed to create miner transaction, but have no non-miner txs to pop");
+      b.tx_hashes.pop_back();
+      assert(selected_backlog.size() == b.tx_hashes.size());
+      const tx_block_template_backlog_entry &backlog_entry = selected_backlog.back();
+      assert(txs_weight >= backlog_entry.weight);
+      txs_weight -= backlog_entry.weight;
+      assert(fee >= backlog_entry.fee);
+      fee -= backlog_entry.fee;
+      selected_backlog.pop_back();
+      cumulative_weight = txs_weight;
+      continue;
+    }
+
     size_t coinbase_weight = get_transaction_weight(b.miner_tx);
     if (coinbase_weight > cumulative_weight - txs_weight)
     {
@@ -1791,6 +1820,7 @@ bool Blockchain::create_block_template(block& b, const crypto::hash *from_block,
         ", cumulative weight " << cumulative_weight << " is now good");
 #endif
 
+    expected_reward = get_outs_money_amount(b.miner_tx);
     if (!from_block)
       cache_block_template(b, miner_address, ex_nonce, diffic, height, expected_reward, cumulative_weight, seed_height, seed_hash, pool_cookie);
     return true;
@@ -1823,9 +1853,7 @@ bool Blockchain::get_miner_data(uint8_t& major_version, uint64_t& height, crypto
   median_weight = m_current_block_cumul_weight_median;
   already_generated_coins = m_db->get_block_already_generated_coins(height - 1);
 
-  m_tx_pool.get_block_template_backlog(tx_backlog);
-
-  return true;
+  return m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, major_version, tx_backlog);
 }
 //------------------------------------------------------------------
 // for an alternate chain, get the timestamps from the main chain to complete
@@ -5562,8 +5590,8 @@ bool Blockchain::for_all_outputs(uint64_t amount, std::function<bool(uint64_t he
 
 void Blockchain::invalidate_block_template_cache()
 {
-  MDEBUG("Invalidating block template cache");
-  m_btc_valid = false;
+  if (std::exchange(m_btc_valid, false))
+    MDEBUG("Invalidating block template cache");
 }
 
 void Blockchain::cache_block_template(const block &b, const cryptonote::account_public_address &address, const blobdata &nonce, const difficulty_type &diff, uint64_t height, uint64_t expected_reward, uint64_t cumulative_weight, uint64_t seed_height, const crypto::hash &seed_hash, uint64_t pool_cookie)
@@ -5592,7 +5620,14 @@ void Blockchain::send_miner_notifications(uint64_t height, const crypto::hash &s
   const uint64_t median_weight = m_current_block_cumul_weight_median;
 
   std::vector<tx_block_template_backlog_entry> tx_backlog;
-  m_tx_pool.get_block_template_backlog(tx_backlog);
+  if (!m_tx_pool.get_block_template_backlog(median_weight, already_generated_coins, major_version, tx_backlog))
+  {
+    MERROR("Could not retreive mempool's block template backlog for block " << height
+      << ": median weight " << median_weight << " already generated coins "
+      << already_generated_coins << " major version " << major_version
+      << ". Miner notifications will not be sent.");
+    return;
+  }
 
   for (const auto& notifier : m_miner_notifiers)
   {
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -86,6 +86,9 @@ DISABLE_VS_WARNINGS(4267)
 // used to overestimate the block reward when estimating a per kB to use
 #define BLOCK_REWARD_OVERESTIMATE (10 * 1000000000000)
 
+/// @brief Parameter for block template backlog fetching: targets total weight 112.5% of the block's limit
+#define DEFAULT_RPC_TX_BACKLOG_OVERPICK 0.125f
+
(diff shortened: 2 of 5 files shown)


src/cryptonote_core/tx_pool.cpp
```diff
@@ -1498,6 +1429,22 @@ namespace cryptonote
       return false; // we are already sure that this tx isn't passing for this exact chain
 
     tx_verification_context tvc{};
+
+    // try verify non-input consensus rules if not already verified for this HF version
+    const uint8_t hf_version = m_blockchain.get_current_hard_fork_version();
```
selsta, line 1434: Does this check against the correct HF version?

src/cryptonote_core/tx_pool.cpp
selsta, line 1783:
```suggestion
        e.meta.nic_verified_hf_version = version;
        m_blockchain.update_txpool_tx(e.txid, e.meta);
```

selsta (commented):
`m_mempool.validate()` is still being called on startup in `cryptonote_core.cpp`, you might be able to avoid this with this PR.

selsta:
Also needs rebase.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1727,37 +1652,76 @@ namespace cryptonote
...
+          // continue, not fatal
+        }
       }
       if (!ready)
       {
         LOG_PRINT_L2("  not ready to go");
+        if (!meta.kept_by_block)
+        {
+          LOG_PRINT_L2("  and will be removed from the mempool");
+          pool_txs_to_remove.push_back(sorted_it->get_right());
```
selsta, line 1670: Notifications (activated by `--zmq-pub`) can delete transactions needed later in the same reorg. Cleanup should be deferred until the final chain state is established.

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1727,37 +1652,76 @@ namespace cryptonote
...
+        }
```
thomasbuilds: This deletes transactions that are valid, and it does so in the middle of chain switches.

`is_transaction_ready_to_go()` means "includable at this tip", not "invalid": the input rules depend on the current height (`CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE` at blockchain.cpp:3491, `is_tx_spendtime_unlocked()` for coinbase outputs), and readiness also fails while a ring member's block is not in the chain. `switch_to_alternative_blockchain()` pops the old blocks and then adds the alt blocks one at a time through `handle_block_to_main_chain()`, and each of those ends with `send_miner_notifications()`, which now runs this function. With `--zmq-pub` (p2pool requires it) that means: at the intermediate, shorter heights any relayed tx accepted since the last block whose youngest ring member sits at the boundary (a real input with exactly 10 confirmations, a coinbase spent right after it unlocked, or a decoy in the boundary block, which wallet2's picker produces for roughly 8% of rings) is reported not ready and, since it is not `kept_by_block`, deleted here. If a later alt block contains it, the switch fails with "has at least one unknown transaction" (the alt blocks are added with an empty supplement), the node rolls back to the losing chain and mines and relays on it until a peer delivers the blocks again; only on that second delivery does the tx come in through the supplement as kept-by-block and the switch succeeds. Any reorg of depth >= 2 can trigger it.

Reproduced two ways at this head: a core_test (relayed tx spending a coinbase that unlocked at height 76, depth-4 reorg with the tx in the alt tip: "and will be removed from the mempool", then "Failed to switch to alternative blockchain"; the same test passes on master), and with two regtest `monerod --zmq-pub` plus wallet-rpc where the alt chain is fed through `submit_block` like a fluffy block: master switches, this head answers "Block not accepted", stays on its own chain with the tx gone from the pool, and switches only after the peer re-sends the blocks over sync.

Even outside reorgs this turns `get_miner_data` (a read RPC) into something that deletes pool txs, and "not ready" is transient after `pop_blocks` as well. I would take the removal out of this PR: the `last_failed_id == top_block_hash` short-circuit already makes repeated input failures cheap, and caching NIC failures (comment in `is_transaction_ready_to_go()`) covers the range-proof cost. If a purge is wanted it belongs where no reorg can be in progress and where every node runs it (`remove_stuck_transactions()` on idle), limited to txs whose cached version differs from the current one and which fail `ver_non_input_consensus()`.

```suggestion
```

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1634,11 +1561,19 @@ namespace cryptonote
...
+    const uint64_t reference_tx_weight = DYNAMIC_FEE_REFERENCE_TRANSACTION_WEIGHT;
+    constexpr size_t min_tx_size_approx = 1200;
+
+    selected_backlog.reserve(std::min<size_t>(2 * median_weight / min_tx_size_approx, CRYPTONOTE_MAX_TX_PER_BLOCK));
 
-    size_t max_total_weight_pre_v5 = (130 * median_weight) / 100 - CRYPTONOTE_COINBASE_BLOB_RESERVED_SIZE;
-    size_t max_total_weight_v5 = 2 * median_weight - CRYPTONOTE_COINBASE_BLOB_RESERVED_SIZE;
-    size_t max_total_weight = version >= 5 ? max_total_weight_v5 : max_total_weight_pre_v5;
     std::unordered_set<crypto::key_image> k_images;
+    std::deque<crypto::hash> pool_txs_to_remove;
```
thomasbuilds: Goes with the removal below.

```suggestion
```

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1727,37 +1652,76 @@ namespace cryptonote
...
+      uint64_t tx_fee;
+      uint8_t nic_verified_hf_version;
+      crypto::hash valid_input_verification_id;
+      bool relayed, do_not_relay, double_spend_seen, pruned;
+      const bool removed = take_tx(pool_tx_to_remove, tx, tx_blob, tx_weight,
+        tx_fee, nic_verified_hf_version, valid_input_verification_id, relayed,
+        do_not_relay, double_spend_seen, pruned);
+      if (!removed)
+        MERROR("Failed to remove TX " << pool_tx_to_remove << " from the mempool during block template construction");
+    }
```
thomasbuilds: And this loop.

```suggestion
```

src/cryptonote_core/tx_pool.cpp
thomasbuilds: To answer selsta's question: `get_current_hard_fork_version()` is the version of the next block (`HardFork::add()` advances `current_fork_index` with `get_voted_fork_index(height + 1)`), and `create_block_template()` uses the same call for `b.major_version`, so it matches today. Notifications and `get_miner_data` use `get_ideal_version(height)`, which is identical for threshold-0 forks. It only holds because three call sites agree, though, and `get_block_template_backlog()` already receives the right `version`: suggest taking it as a parameter (signature and call site below) and dropping this line.

```suggestion
```

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1469,27 +1404,23 @@ namespace cryptonote
     return ret;
   }
   //---------------------------------------------------------------------------------
-  bool tx_memory_pool::is_transaction_ready_to_go(txpool_tx_meta_t& txd, const crypto::hash &txid, const cryptonote::blobdata_ref& txblob, transaction &tx) const
+  bool tx_memory_pool::is_transaction_ready_to_go(txpool_tx_meta_t& txd, const crypto::hash &txid) const
```
thomasbuilds:
```suggestion
  bool tx_memory_pool::is_transaction_ready_to_go(txpool_tx_meta_t& txd, const crypto::hash &txid, const uint8_t hf_version) const
```

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1669,55 +1607,42 @@ namespace cryptonote
...
-
       // Skip transactions that are not ready to be
       // included into the blockchain or that are
       // missing key images
       const cryptonote::txpool_tx_meta_t original_meta = meta;
       bool ready = false;
       try
       {
-        ready = is_transaction_ready_to_go(meta, sorted_it->get_right(), txblob, tx);
+        ready = is_transaction_ready_to_go(meta, txid);
```
thomasbuilds:
```suggestion
        ready = is_transaction_ready_to_go(meta, txid, version);
```

src/cryptonote_core/tx_pool.h
```diff
@@ -583,38 +592,15 @@ namespace cryptonote
...
      * @param txd the transaction to check (and info about it)
      * @param txid the txid of the transaction to check
-     * @param txblob the transaction blob to check
-     * @param tx the parsed transaction, if successful
      *
      * @return true if the transaction is good to go, otherwise false
      */
-    bool is_transaction_ready_to_go(txpool_tx_meta_t& txd, const crypto::hash &txid, const cryptonote::blobdata_ref &txblob, transaction&tx) const;
-    bool is_transaction_ready_to_go(txpool_tx_meta_t& txd, const crypto::hash &txid, const cryptonote::blobdata &txblob, transaction&tx) const;
+    bool is_transaction_ready_to_go(txpool_tx_meta_t& txd, const crypto::hash &txid) const;
```
thomasbuilds:
```suggestion
    bool is_transaction_ready_to_go(txpool_tx_meta_t& txd, const crypto::hash &txid, uint8_t hf_version) const;
```

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1498,6 +1429,22 @@ namespace cryptonote
...
+    {
+      if (!ver_non_input_consensus(lazy_tx(), tvc, hf_version))
+      {
+        txd.last_failed_height = top_block_height;
+        txd.last_failed_id = top_block_hash;
+        return false;
+      }
+      // set NIC verified HF version on success
+      txd.nic_verified_hf_version = hf_version;
+    }
```
thomasbuilds: A failure is only remembered as `last_failed_id` for this tip, so on every new block a tx that fails here is parsed and run through `ver_non_input_consensus()` (range proof verification) again. Relayed txs escape that only because of the deletion in `get_block_template_backlog()`; kept-by-block txs (alt-block supplements, popped blocks) stay for `CRYPTONOTE_MEMPOOL_TX_FROM_ALT_BLOCK_LIVETIME` and get re-verified on every template build. An alt block with valid PoW at a pre-fork height can fill the pool with such txs (they skip the fee and weight policy), which turns into per-block CPU on every mining/zmq node after the fork. A NIC failure is permanent for a given fork version, so it can be cached the same way as a success, e.g. with a second byte taken from the padding (struct change suggested in blockchain_db.h):

```suggestion
    if (txd.nic_failed_hf_version == hf_version)
      return false; // known to fail for this fork version, don't redo the range proof verification on every block
    if (txd.nic_verified_hf_version != hf_version)
    {
      if (!ver_non_input_consensus(lazy_tx(), tvc, hf_version))
      {
        txd.nic_failed_hf_version = hf_version;
        txd.last_failed_height = top_block_height;
        txd.last_failed_id = top_block_hash;
        return false;
      }
      // set NIC verified HF version on success
      txd.nic_verified_hf_version = hf_version;
    }
```

Related: this loop only exits early on selected weight, so after a fork that invalidates most of the pool the first call re-verifies the whole pool in one go under both locks (the #10142 freeze, moved). A per-call budget for re-checks would bound that.

src/blockchain_db/blockchain_db.h
```diff
@@ -172,9 +172,12 @@ struct txpool_tx_meta_t
   uint8_t is_forwarding: 1;
   uint8_t bf_padding: 3;
 
-  uint8_t padding[44]; // til 160 bytes
+  uint8_t padding[43]; // til 159 bytes
 
-  // If non-null, this verification ID is set for this tx only when some mixring passed ver_input_proofs_rings()
+  //! If non-zero, this is set to the HF major version for which this tx passes ver_non_input_consensus()
+  uint8_t nic_verified_hf_version;
```
thomasbuilds: For the failure cache suggested in `is_transaction_ready_to_go()`:

```suggestion
  uint8_t padding[42]; // til 158 bytes

  //! If non-zero, this is set to the HF major version for which this tx failed ver_non_input_consensus()
  uint8_t nic_failed_hf_version;

  //! If non-zero, this is set to the HF major version for which this tx passes ver_non_input_consensus()
  uint8_t nic_verified_hf_version;
```

src/blockchain_db/blockchain_db.h
```diff
@@ -192,6 +195,7 @@ struct txpool_tx_meta_t
   }
 };
 static_assert(sizeof(txpool_tx_meta_t) == 192, "possible DB migration needed for changes to txpool_tx_meta_t");
+static_assert(offsetof(txpool_tx_meta_t, nic_verified_hf_version) == 159, "NIC verified HF version wrong offset");
```
thomasbuilds:
```suggestion
static_assert(offsetof(txpool_tx_meta_t, nic_failed_hf_version) == 158, "NIC failed HF version wrong offset");
static_assert(offsetof(txpool_tx_meta_t, nic_verified_hf_version) == 159, "NIC verified HF version wrong offset");
```

src/cryptonote_core/blockchain.cpp
```diff
@@ -4235,15 +4266,33 @@ bool Blockchain::handle_block_to_main_chain(const block& bl, const crypto::hash&
...
+      assert(hf_version);
+      if (nic_verified_hf_version != hf_version)
+      {
+        if (!ver_non_input_consensus(tx, tvc, hf_version))
+        {
+          valid = false;
+          MERROR_VER("Block with id: " << id  << " has at least one transaction (id: "
+            << tx_id << ") which breaks non-input consensus rules.");
+        }
+      }
```
thomasbuilds: Two small things here:

- If the re-check passes and the block then fails on a later tx, `return_txs_to_pool()` re-adds this tx with the old cached version and `add_tx()` runs `ver_non_input_consensus()` once more. Updating the tuple after a successful check avoids that.
- `assert(hf_version)` is compiled out in release builds.

```suggestion
      if (nic_verified_hf_version != hf_version)
      {
        if (!ver_non_input_consensus(tx, tvc, hf_version))
        {
          valid = false;
          MERROR_VER("Block with id: " << id  << " has at least one transaction (id: "
            << tx_id << ") which breaks non-input consensus rules.");
        }
        else
        {
          // remember the successful check so return_txs_to_pool() doesn't redo it if this block fails on a later tx
          std::get<3>(txs_meta.back()) = hf_version;
        }
      }
```

Also note that the return path now drops a returned tx that fails NIC at `hf_version` (`add_tx()` returns false, "Failed to return taken transaction"), where `(hf_version, hf_version)` used to re-add unconditionally. For a kept-by-block tx whose fork version legitimately differs this discards something a later reorg may need (an alt block's txs only exist in the pool), and the comment above the `add_tx()` call in `return_txs_to_pool()` still describes the old behaviour. Low priority.

src/cryptonote_core/tx_pool.cpp
thomasbuilds, line 1671: With `--zmq-pub`, `switch_to_alternative_blockchain()` adds the alt blocks one by one, each `handle_block_to_main_chain()` ends in `send_miner_notifications()` and lands here at a height lower than the one the pool txs were accepted at. A relayed tx whose youngest ring member or coinbase input sits at the 10-block or 60-block boundary fails `check_tx_inputs()` there, gets deleted, and the alt block containing it then fails with "unknown transaction" (alt blocks are added with an empty supplement); the node stays on the losing chain until a peer re-delivers the blocks. Any reorg of depth >= 2 can trigger it.

"Not ready" is transient by nature (height-dependent rules, `pop_blocks`), so deferring the cleanup does not fix it; better not to delete here at all and cache NIC failures instead (see the comment in `is_transaction_ready_to_go()`).

src/cryptonote_core/tx_pool.cpp
thomasbuilds, line 1445: A failure is only remembered as `last_failed_id` for this tip, so a failing tx is parsed and put through `ver_non_input_consensus()` again on every block. Kept-by-block txs (alt-block supplements, popped blocks) stay for 7 days, and an alt block at a pre-fork height can fill the pool with them without fee or weight limits, which makes it per-block range-proof work on every mining/zmq node after a fork. A NIC failure is permanent for a fork version: a `nic_failed_hf_version` byte from the padding, checked before verifying, closes it.

Also, failed txs never trigger the early exit of the backlog loop, so after a fork that invalidates most of the pool the first `get_block_template_backlog()` call verifies the whole pool under both locks.

src/cryptonote_core/blockchain.cpp
thomasbuilds, line 4282: `return_txs_to_pool()` now passes the stored version, so `add_tx()` drops a returned tx that fails NIC at `hf_version` ("Failed to return taken transaction"), where `(hf_version, hf_version)` always re-added; for a kept-by-block tx whose version legitimately differs that discards something a later reorg may need. Setting `std::get<3>(txs_meta.back()) = hf_version` after a successful re-check here also avoids verifying it again on return.

src/cryptonote_core/blockchain.cpp
```diff
@@ -4359,19 +4408,6 @@ bool Blockchain::handle_block_to_main_chain(const block& bl, const crypto::hash&
...
-  const uint8_t new_hf_version = get_current_hard_fork_version();
-  if (new_hf_version != hf_version)
-  {
-    // the genesis block is added before everything's setup, and the txpool is empty
-    // when we start from scratch, so we skip this
-    const bool is_genesis_block = new_height == 1;
-    if (!is_genesis_block)
-    {
-      MGINFO("Validating txpool for v" << (unsigned)new_hf_version);
-      m_tx_pool.validate(new_hf_version);
```
thomasbuilds, line 4371: Without this sweep, nodes that never build templates keep NIC-invalid txs for 3 days after a fork, list them in `get_transaction_pool`, and `get_relayable_transactions()` keeps re-relaying them (fluff txs are re-sent with back-off for 1.5 days, no validity check); peers that no longer have them (the ones that build templates) verify and drop the connection. Skipping relay for txs whose cached version differs and which fail re-verification, or an incremental purge from `on_idle()`, would cover it.

PR #10920, opened 2026-07-15
dns_utils: update default public DNS resolvers

selsta:
This should also fix flaky DNS tests on CI by setting `DNS_PUBLIC=tcp` env var, which uses the resolvers in `DEFAULT_DNS_PUBLIC_ADDR` with fallback support. I also removed skipping DNS related tests from macOS and Windows.

Related: https://github.com/monero-project/monero/issues/10445

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ env:
     ccache --max-size=150M
   CTEST_EXCLUDE_SLOW: |
     cd build
-    env GTEST_FILTER="-select_outputs.*:DNSResolver.*:AddressFromURL.*" ctest --output-on-failure -E "functional_tests_rpc|core_tests|cnv4-jit|hash-variant2-int-sqrt|wide_difficulty"
+    env DNS_PUBLIC=tcp GTEST_FILTER="-select_outputs.*" ctest --output-on-failure -E "functional_tests_rpc|core_tests|cnv4-jit|hash-variant2-int-sqrt|wide_difficulty"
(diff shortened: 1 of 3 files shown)


tobtoht (approved):

jpk68:
NACK. This change reduces the number of independent organizations from 5 to 3; the new list is also heavily comprised of providers based in North America. Adding Cloudflare seems hugely centralizing, and Control D is owned by a VPN company. ~~In my opinion, Quad9 should be added as a supplement (not a replacement), or maybe something like Mullvad DNS.~~

Edit: I am mostly fine with this now, ~~though I would prefer another non-EU provider (like SkyDNS) be there too.~~

selsta:
Updated to remove Cloudflare and Control D, re-added CCC and FDN.

tobtoht:
`80.67.169.40` (FDN) doesn't appear to be up.

selsta:
@tobtoht could be because of https://www.fdn.fr/renforcement-serveurs-dns-2025/ ?

tobtoht:
>Unavailability of our DNS services for the rest of the world.
>
>How can I continue to use our DNS services outside of the FDN and FFDN networks? By configuring your routers, boxes or browsers to activate DoT and DoH services.

Shouldn't be in the list then? (Edit: it was removed)

selsta:
Removed FDN and re-added Control D after talking to @jpk68.

tobtoht (approved):
Checked all (new) resolvers.

jpk68 (approved):

PR #10921, opened 2026-07-15
general: replace `auto_scope_leave_caller` with `scope_guard`

jeffro256:
And replace unused inclusions of `misc_language.h`.

Reduces `monerod` binary size by ~142KB when linking dynamically on my machine. Also reduces number of allocations done for lambdas passed to `std::function()` which cannot fit in SSO buffer.


selsta:
Fails to compile on CI, fuzz tests need update.

selsta (approved):

tobtoht (approved):

PR #10930, opened 2026-07-18
rpc: speed up mempool lookup in get_transactions

jpk68:
Avoids a quadratic lookup for mempool transactions in `daemon_handler`.

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -951,11 +951,14 @@ namespace cryptonote
       {
         // sort to match original request
         std::vector<std::tuple<crypto::hash, cryptonote::blobdata, crypto::hash, cryptonote::blobdata>> sorted_txs;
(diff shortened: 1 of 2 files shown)


selsta (approved):
`txs` already returned an `unordered_map` and `missed_hashes` has no documented ordering guarantee, so using an `unordered_set` here should be okay, but I'd also like @vtnerd's opinion.

thomasbuilds (approved):

selsta (approved):

selsta:
I checked all relevant callers of get_transactions and this appears safe.

jeffro256:
This PR caused a merge conflict with #10605, due to them performing similar operations in this part of the code. But by contrast, #10605 is faster, uses less memory, and doesn't shuffle the order of the missed txs, which could be a breaking change. I don't know why this PR got approved without bumping the RPC version...

selsta:
@jeffro256 we stopped bumping RPC on master a while ago to avoid merge conflicts, but I can go back to requiring it

PR #10932, opened 2026-07-20
ringct: remove dead code

thomasbuilds:
- h2b, b2h (rctTypes) and sumKeys (rctOps): from the original RingCT import (9b1afe5f2, 2016-05-13); never referenced by any caller since the import.
- cn_fast_hash128, hash_to_scalar128 (rctOps): from the same import. hash_to_scalar128 (and cn_fast_hash128 through it) was last used by the original MLSAG gen/ver hashing, which dbb5f2d6a (2016-07-09) replaced with keyV-based hash_to_scalar; dead since then.
- bos_coster_heap_conv (multiexp): added in 9ff6e6a0a (2018-01-09) alongside bos_coster_heap_conv_robust; multiexp() only ever dispatched the robust variant, so the non-robust one was never called. The robust variant is kept, as tests/performance_tests still uses it (production dispatch moved to Straus/Pippenger in 939bc22 / 263431c).


selsta (approved):

jeffro256 (approved):
AFAICT, none of this code is in use in the FCMP++ codebase either

PR #10934, opened 2026-07-20
tests: tests_common library and gen_corpus executable

jeffro256:
Adds two components to the tests directory:
1. `tests_common` a shared utility library for different types of tests
2. `gen_corpus` an executable to dynamically generate seed corpora for fuzzing during the build stage

Also updates the transaction fuzzing test to call `ver_non_input_consensus()` for different HF versions.

OSS-Fuzz downstream PR: https://github.com/google/oss-fuzz/pull/15893

Part of upstreaming Carrot/FCMP++


tests/fuzz/gen_corpus/gen_corpus.h
```diff
@@ -0,0 +1,114 @@
...
+class seed_corpus_source_single: public seed_corpus_source
+{
+public:
+    seed_corpus_source_single();
+
+    void gen_corpus(seed_corpus_sink &sink) override;
+
+protected:
+    virtual const char * get_fuzzer_name() const = 0;
+    virtual const char * get_seed_curpus_name() const = 0;
```
thomasbuilds:
```suggestion
    virtual const char * get_seed_corpus_name() const = 0;
```

tests/fuzz/gen_corpus/gen_corpus.h
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * @brief Interface to generate seed corpora
+ */
+class seed_corpus_source
+{
+public:
+    virtual void gen_corpus(seed_corpus_sink &sink) = 0;
+
+    ~seed_corpus_source() = default;
```
thomasbuilds:
```suggestion
    virtual ~seed_corpus_source() = default;
```

The registry stores these as `std::unique_ptr<seed_corpus_source>`, so at exit each object is deleted through a base pointer with a non-virtual destructor: the base destructor runs instead of the derived one, and the object is freed as the wrong type (8 vs 24 bytes). Clang flags it by default:

```
warning: delete called on 'fuzz::seed_corpus_source' that is abstract but has non-virtual destructor [-Wdelete-abstract-non-virtual-dtor]
```

tests/fuzz/gen_corpus/gen_corpus.h
```diff
@@ -0,0 +1,114 @@
...
+ */
+class seed_corpus_sink
+{
+public:
+    virtual void send_seed_corpus(const char * fuzzer_name,
+        const char * seed_corpus_name,
+        const unsigned char * seed_corpus_buf,
+        std::size_t seed_corpus_len) = 0;
+
+    ~seed_corpus_sink() = default;
```
thomasbuilds:
```suggestion
    virtual ~seed_corpus_sink() = default;
```
Same here though harmless now

thomasbuilds (changes requested):

tests/fuzz/gen_corpus/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+} //anonymous namespace
+
+int main(int argc, const char * const argv[])
+{
+    tools::on_startup();
+    epee::string_tools::set_module_name_and_folder(argv[0]);
+    mlog_configure(mlog_get_default_log_path("gen_corpus.log"), true);
+
+    // the default test data directory is ../../data (relative to the executable's directory)
+    const auto default_test_data_dir = std::filesystem::canonical(argv[0]).parent_path()
```
selsta: This does not match oss-fuzz, so either you need to change this here or add `--data-dir` to oss-fuzz `build.sh`.

tests/fuzz/gen_corpus/gen_corpus.cpp
```diff
@@ -0,0 +1,105 @@
...
+#undef MONERO_DEFAULT_LOG_CATEGORY
+#define MONERO_DEFAULT_LOG_CATEGORY "fuzz"
+
+namespace fuzz
+{
+namespace
+{
+std::list<std::unique_ptr<seed_corpus_source>> registered_seed_corpus_sources;
+std::mutex registry_mutex;
+} //anonymous namespace
```
selsta: This can be accessed before it's constructed.

tests/tests_common/src/tx_construction_helpers.cpp
```diff
@@ -0,0 +1,361 @@
...
+cryptonote::tx_source_entry gen_tx_source_entry_fake_members(
+    const stripped_down_tx_source_entry_t &in,
+    const size_t mixin,
+    const uint64_t max_global_output_index)
+{
+    const size_t ring_size = mixin + 1;
+
+    CHECK_AND_ASSERT_THROW_MES(in.global_output_index <= max_global_output_index,
+        "real global output index too low");
+    CHECK_AND_ASSERT_THROW_MES(max_global_output_index >= ring_size,
```
selsta: off by one

tests/fuzz/gen_corpus/gen_corpus.cpp
jeffro256: What do you mean? C++ [dynamic initializations rules](https://en.cppreference.com/cpp/language/initialization#Dynamic_initialization) dictate that initialization for objects with static duration should happen either before the first line of `main()`, or before the first ODR-use of that variable.

tests/fuzz/gen_corpus/main.cpp
jeffro256: Okay, then I will update `build.sh`

jeffro256: Done in OSS-Fuzzz

tests/fuzz/gen_corpus/gen_corpus.cpp
jeffro256: Okay, you're right. The ODR-use is from a different TU during dynamic initialization of another variable with static duration, so the order of `registered_seed_corpus_sources`'s initialization isn't guaranteed to be before other variables which call `register_seed_corpus_source`. Works on GCC though 🤷🏻‍♂️ ;). Fixed in latest.

tests/tests_common/src/tx_construction_helpers.cpp
```diff
@@ -0,0 +1,361 @@
...
+cryptonote::tx_source_entry gen_tx_source_entry_fake_members(
+    const stripped_down_tx_source_entry_t &in,
+    const size_t mixin,
+    const uint64_t max_global_output_index)
+{
+    const size_t ring_size = mixin + 1;
+
+    CHECK_AND_ASSERT_THROW_MES(in.global_output_index <= max_global_output_index,
+        "real global output index too low");
+    CHECK_AND_ASSERT_THROW_MES(max_global_output_index >= ring_size + 1,
```
selsta:
```suggestion
    CHECK_AND_ASSERT_THROW_MES(max_global_output_index + 1 >= ring_size,
```

or `max_global_output_index >= mixin`.

jeffro256: Ah shoot, stupid typo

jeffro256: Thanks

jeffro256:
Implemented @selsta's feedback and squashed

selsta (approved):

tests/fuzz/transaction.cpp
```diff
@@ -26,17 +26,21 @@
...
 #include "fuzzer.h"
 
 BEGIN_INIT_SIMPLE_FUZZER()
 END_INIT_SIMPLE_FUZZER()
 
 BEGIN_SIMPLE_FUZZER()
-  cryptonote::transaction tx{};
+  cryptonote::transaction tx = AUTO_VAL_INIT(tx);
   parse_and_validate_tx_from_blob(std::string((const char*)buf, len), tx);
+  for (uint8_t hf_version = 1; hf_version <= MAX_HF_VERSION; ++hf_version)
```
thomasbuilds: `ver_mixed_rct_semantics()` has no `hf_version` parameter ([tx_verification_utils.cpp#L394](https://github.com/monero-project/monero/blob/c7d82632577eb94b08f63cb4378faf78b3ef7f05/src/cryptonote_core/tx_verification_utils.cpp#L394)), so every fork that accepts the transaction re-runs the full range proof verification. On the checked-in TX2 (RCTTypeSimple, 2 inputs, 2 outputs), hard forks 4 to 8 each spend 10-13 ms in the same `verRctSemanticsSimple()` call. The total is 55-57 ms per exec in a native build and about 115 ms in an oss-fuzz-style libFuzzer build, compared with about 1 µs and 0.1 ms respectively before this change. The five RingCT seeds this PR generates cost 8-60 ms per exec for the same reason. In 30-second libFuzzer sessions, a corpus of the 10 generated seeds goes from about 310k runs to 16-32k, and a TX2-only corpus from about 300k to 6-14k. libFuzzer's `-report_slow_units` defaults to 10 seconds, so none of this is ever reported. The target also starts an nproc-1 threadpool now, because `verRctSemanticsSimple(rv)` delegates to the batch overload ([rctSigs.cpp#L1480](https://github.com/monero-project/monero/blob/c7d82632577eb94b08f63cb4378faf78b3ef7f05/src/ringct/rctSigs.cpp#L1480), [#L1352](https://github.com/monero-project/monero/blob/c7d82632577eb94b08f63cb4378faf78b3ef7f05/src/ringct/rctSigs.cpp#L1352)).

Suggestion: verify one fork per exec and take the fork from the input, for example a leading selector byte. `gen_corpus` can emit that byte because it knows which fork each seed was built for, and TX1/TX2 would need one byte prepended once. All 16 forks remain covered, across execs rather than within each one.

tests/tests_common/src/tx_construction_helpers.cpp
```diff
@@ -0,0 +1,361 @@
...
+    const cryptonote::blobdata extra_nonce;
+
+    cryptonote::transaction tx;
+    size_t cumulative_tx_weight = median_weight * 9 / 10;
+    size_t current_block_weight = cumulative_tx_weight;
+    rct::xmr_amount fee = reward;
+    bool r = cryptonote::construct_miner_tx(height, median_weight, already_generated_coins,
+        current_block_weight, fee, miner_address, tx);
+    CHECK_AND_ASSERT_THROW_MES(r, "Coinbase transaction failed on first try");
+    while (
```
thomasbuilds, line 61: There is no iteration cap, and for some combinations of reward and hard fork this loop provably never terminates. The next state depends only on `cumulative_tx_weight`, `fee`, `current_block_weight`, `r` and the weight and output sum of `tx`, and tracking that state finds repeats: a reward of 1e11 at hard forks 5 to 14 repeats with period 181, 0.5 XMR cycles at forks 4, 15 and 16, 0.7 XMR at forks 2 and 3, and rewards of 0, 1, 1e6 and 1e9 cycle at every fork (15 of 320 cells over 0.1 to 2.0 XMR at all 16 forks). The 2 COIN reward that `gen_corpus` uses converges everywhere, so this does not bite today, but as a shared helper it should fail loudly, for example `CHECK_AND_ASSERT_THROW_MES(++iterations < 10000, "coinbase reward adjustment did not converge");`

tests/tests_common/src/tx_construction_helpers.cpp
```diff
@@ -0,0 +1,361 @@
...
+cryptonote::transaction construct_miner_tx_with_reward(const size_t height,
+    const rct::xmr_amount reward,
+    const cryptonote::account_public_address &miner_address,
+    const uint8_t hf_version)
+{
+    constexpr rct::xmr_amount already_generated_coins = 10000 * COIN;
+    constexpr size_t max_outs = 999;
+
+    const size_t median_weight = cryptonote::get_min_block_weight(hf_version);
+    const size_t weight_limit = 2 * std::max(median_weight, cryptonote::get_min_block_weight(hf_version));
```
thomasbuilds:
```suggestion
    const size_t weight_limit = 2 * median_weight;
```

tests/tests_common/src/tx_construction_helpers.cpp
```diff
@@ -0,0 +1,361 @@
...
+    const cryptonote::blobdata extra_nonce;
+
+    cryptonote::transaction tx;
+    size_t cumulative_tx_weight = median_weight * 9 / 10;
+    size_t current_block_weight = cumulative_tx_weight;
+    rct::xmr_amount fee = reward;
+    bool r = cryptonote::construct_miner_tx(height, median_weight, already_generated_coins,
+        current_block_weight, fee, miner_address, tx);
```
thomasbuilds: The first attempt does not pass `hf_version`, so it uses the default of hard fork 1 in `construct_miner_tx`. The returned transaction is still correct because every later attempt inside the loop passes `hf_version`, but the hard fork 1 attempt seeds the adjustment trajectory: with the parameter passed, the set of (reward, fork) pairs for which the loop cycles changes (for example 1e11 at forks 5 to 14 then converges, while 6e11 at forks 5 to 14 cycles instead), so this belongs together with an iteration cap. 2 COIN still converges at all 16 forks with this change.

```suggestion
    bool r = cryptonote::construct_miner_tx(height, median_weight, already_generated_coins,
        current_block_weight, fee, miner_address, tx, extra_nonce, max_outs, hf_version);
```

tests/tests_common/src/tx_construction_helpers.cpp
```diff
@@ -0,0 +1,361 @@
...
+//----------------------------------------------------------------------------------------------------------------------
+cryptonote::tx_source_entry gen_tx_source_entry_fake_members(
+    const stripped_down_tx_source_entry_t &in,
+    const size_t mixin,
+    const uint64_t max_global_output_index)
+{
+    const size_t ring_size = mixin + 1;
+
+    CHECK_AND_ASSERT_THROW_MES(in.global_output_index <= max_global_output_index,
+        "real global output index too low");
```
thomasbuilds:
```suggestion
        "real global output index too high");
```

tests/fuzz/gen_corpus/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+private:
+    std::filesystem::path m_fuzz_data_dir;
+    bool m_overwrite_existing;
+    std::map<std::string, std::string> m_generated_sha1s;
+};
+
+} //anonymous namespace
+
+int main(int argc, const char * const argv[])
```
thomasbuilds, line 119: `main()` has no try/catch, so any failure ends in `std::terminate`: SIGABRT with exit code 134, and `~FileSeedCorpusSink()` never runs, so the SHA1 dump is lost. Two easy ways to hit it are running twice without `--force`, and running with an `argv[0]` that has no slash (a PATH lookup), because `canonical(argv[0])` on line 126 executes before the options are parsed, so `--data-dir` does not avoid it.

tests/fuzz/gen_corpus/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "gen_corpus.h"
+
+#include "common/command_line.h"
+#include "common/util.h"
+#include "hex.h"
+#include "misc_log_ex.h"
+#include "string_tools.h"
+
+#include <iostream>
+#include <filesystem>
```
thomasbuilds, line 38: This is the only use of `std::filesystem` in the tree; 57 files use `boost::filesystem`. The README lists GCC 7 as the minimum version, and GCC 8.5 does not link `std::filesystem` without `-lstdc++fs`, which `gen_corpus` does not add. CI builds `gen_corpus` through `ENABLE_FUZZ_TEST=ON` so this is not limited to oss-fuzz.

thomasbuilds (changes requested):
The loop over all 16 hard fork versions in `tests/fuzz/transaction.cpp` raises the cost of one exec on the checked-in TX2 seed from about 1 µs to 55-57 ms in a native build and to about 115 ms in an oss-fuzz-style libFuzzer build. In 30-second libFuzzer sessions on the seed corpus this PR generates that is roughly 310k runs before the change and 16-32k after. The mechanism and a suggested fix are in the comment on that loop.

The remaining comments cover a provably non-terminating case in `construct_miner_tx_with_reward()`, three one-line fixes in the helpers, `gen_corpus` aborting instead of reporting errors, and `<filesystem>` failing to link on GCC 8.

PS: your Carrot branches name this helper `construct_miner_tx_fake_reward_1out(..., num_tx_outputs)`; every v2 seed has a single output so no seed reaches the aggregated bulletproof paths (`V.size() > 1`). The corpora are non-deterministic because all keys come from `crypto::rand` so the SHA1 dump cannot be compared against anything.

tests/fuzz/gen_corpus/main.cpp
jeffro256: Fixed with #11233

tests/fuzz/gen_corpus/main.cpp
jeffro256: `~FileSeedCorpusSink()` should still run on a `throw` which unwinds `main()`, as long as its constructor succeeded.

jeffro256:
Rebased and pushed @thomasbuilds review

tests/fuzz/gen_corpus/transaction.cpp
```diff
@@ -0,0 +1,118 @@
...
+    else
+        return gen_non_miner_tx(hf_version, args...);
+}
+} //anonymous namespace
+
+TX_SEED_CORPUS(coinbase_v1, true, 1)
+
+TX_SEED_CORPUS(v1, false, 1)
+
+TX_SEED_CORPUS(v1_viewtagged, false, 1, true)
```
thomasbuilds:
```suggestion
TX_SEED_CORPUS(v1_viewtagged, false, HF_VERSION_VIEW_TAGS, true)
```
At fork 1 the override changes nothing so this is a second seed of the same shape as `v1`: fork byte 1, version 1, `txout_to_key`. With `HF_VERSION_VIEW_TAGS` it is a version 1 tx with `txout_to_tagged_key` outputs and passes `ver_non_input_consensus` at fork 15

tests/fuzz/gen_corpus/transaction.cpp
```diff
@@ -0,0 +1,118 @@
...
+    constexpr size_t height = 999999;
+    constexpr rct::xmr_amount reward = 2 * COIN;
+
+    return mock::construct_miner_tx_with_reward(height, reward, miner_address, hf_version);
+}
+
+cryptonote::transaction gen_non_miner_tx(const std::uint8_t hf_version, const bool sweep_unmixable_override = false)
+{
+    std::vector<cryptonote::tx_destination_entry> destinations{
+        {cryptonote::tx_destination_entry(COIN, get_core_donation_address(), false)}};
```
thomasbuilds, line 82:
```suggestion
    std::vector<cryptonote::tx_destination_entry> destinations{
        {cryptonote::tx_destination_entry(COIN, get_core_donation_address(), false)},
        {cryptonote::tx_destination_entry(COIN, get_core_donation_address(), false)}};
```
All seven generated non-coinbase seeds have one output; the comment on line 43 says 2-out, and from fork 12 `check_tx_inputs` rejects a version 2 tx with fewer than two outputs. With this change all seven have two outputs and still pass.

tests/fuzz/gen_corpus/main.cpp
thomasbuilds, line 119: Checked with g++ 13 and clang 18 (libstdc++): an uncaught exception reaches `std::terminate` without unwinding, so `~FileSeedCorpusSink()` does not run. A second run without `--force` ends with exit status 134 and no `Seed corpus SHA1 hashes:` line:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  File .../tests/data/fuzz/transaction/coinbase_v1 already exists
```

Same via PATH lookup, where `canonical("gen_corpus")` on line 126 throws before `--data-dir` is parsed. `TRY_ENTRY();` / `CATCH_ENTRY_L0("main", 1);` as in `tests/unit_tests/main.cpp` turns both into exit 1 with the exception logged, and `~FileSeedCorpusSink()` runs in the first case.

jeffro256: You're right, I looked up the standard. It's implementation-defined whether or not stack unwinding occurs on an exception with no handler...

jeffro256:
Pushed @thomasbuilds review

tests/fuzz/gen_corpus/main.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+        // run seed corpus generation
+        FileSeedCorpusSink seed_corpus_sink(data_dir, overwrite_existing);
+        ::fuzz::registry::dispatch_all_registered_seed_corpus_sources(seed_corpus_sink);
+
+        return 0;
+    }
+    catch (const std::exception &e)
+    {
+        std::cerr << "Exception caught in main(): " << e.what();
```
thomasbuilds, line 178:
```suggestion
        std::cerr << "Exception caught in main(): " << e.what() << std::endl;
```
without the newline the next line of terminal output is appended to the message

tests/fuzz/gen_corpus/transaction.cpp
```diff
@@ -0,0 +1,119 @@
...
+ * rest of the bytes: serialized transaction blob
+ */
+
+/// Generates a seed corpus with above format. Transactions are either 1-out
+/// coinbase or 1-in,2-out non-coinbase
+#define TX_SEED_CORPUS(name, is_coinbase, hf_version, ...)                \
+    SEED_CORPUS(transaction, name)                                        \
+    {                                                                     \
+        const cryptonote::transaction tx = gen_tx<is_coinbase>(hf_version \
+            __VA_OPT__(,) __VA_ARGS__);                                   \
```
selsta, line 48: This is not supported by GCC 7, which we technically support according to the README. Do we care about this?

Edit: Seems building with GCC 7 is broken on master anyway: https://github.com/monero-project/monero/pull/9448#issuecomment-2306893632

tests/fuzz/gen_corpus/main.cpp
```diff
@@ -0,0 +1,181 @@
...
+int main(int argc, const char * const argv[])
+{
+    try
+    {
+        tools::on_startup();
+        epee::string_tools::set_module_name_and_folder(argv[0]);
+        mlog_configure(mlog_get_default_log_path("gen_corpus.log"), true);
+
+        // the default test data directory is ../../data (relative to the executable's directory)
+        const auto default_test_data_dir = std::filesystem::canonical(argv[0]).parent_path()
```
selsta, line 128: This does not handle launching through PATH correctly, an example fix would be

```diff
diff --git a/tests/fuzz/gen_corpus/CMakeLists.txt b/tests/fuzz/gen_corpus/CMakeLists.txt
--- a/tests/fuzz/gen_corpus/CMakeLists.txt
+++ b/tests/fuzz/gen_corpus/CMakeLists.txt
@@ -32,0 +33,3 @@
+target_compile_definitions(gen_corpus
+  PRIVATE
+    GEN_CORPUS_DEFAULT_DATA_DIR="${TEST_DATA_DIR}/fuzz")
diff --git a/tests/fuzz/gen_corpus/main.cpp b/tests/fuzz/gen_corpus/main.cpp
--- a/tests/fuzz/gen_corpus/main.cpp
+++ b/tests/fuzz/gen_corpus/main.cpp
@@ -127,9 +127,5 @@
-        // the default test data directory is ../../data (relative to the executable's directory)
-        const auto default_test_data_dir = std::filesystem::canonical(argv[0]).parent_path()
-            .parent_path().parent_path() / "data" / "fuzz";
-
         po::options_description desc_options("Command line options");
         const command_line::arg_descriptor<std::string> arg_data_dir = {
             "data-dir",
             "Data files directory",
-            default_test_data_dir.string()};
+            GEN_CORPUS_DEFAULT_DATA_DIR};
```

selsta (approved):
Two minor points that might not be worth addressing.

jpk68 (approved):
LGTM, though I agree with selsta's comments

PR #10935, opened 2026-07-20
crypto: fix ARMv8 slow-hash inline assembly constraints

### LINKED ISSUE #9371: Corrupted binaries built from Ubuntu 22.04 with depends
Building the Monero binaries on Ubuntu 22.04 with `depends` completes successfully but produces corrupted binaries, wherein at least the wallet files are incompatible and also multisig wallets are broken.

For example, by building on Ubuntu 22.04 ARM64 with the command `make depends target=aarch64-linux-gnu`, the build completes successfully, but multisig wallets will fail to import multisig info from the corrupted peer, with the error `Multisig info is for a different account`.

This led to an issue in which users were unable to recover their funds except by restoring from seed with good binaries.

The instructions indicate that only Ubuntu 18.04 and 20.04 are tested. To avoid loss of funds, I think the build should produce an error or at least a warning when building from Ubuntu >20.04 using `depends`, until later versions are officially tested.

selsta:
The inline assembly did not tell the compiler which registers and memory it modified. The compiler reused them according to the declared constraints, while the assembly had changed them, producing invalid results.

`ARMv8 gcc 11.4.0` before patch

```
The following tests FAILED:
	 11 - hash-slow (Failed)
	 12 - hash-slow-1 (Failed)
	 13 - hash-slow-2 (Failed)
	 14 - hash-slow-4 (Failed)
```

`ARMv8 gcc 11.4.0` after patch

```
    Start 11: hash-slow
1/4 Test #11: hash-slow ........................   Passed    0.11 sec
    Start 12: hash-slow-1
2/4 Test #12: hash-slow-1 ......................   Passed    0.13 sec
    Start 13: hash-slow-2
3/4 Test #13: hash-slow-2 ......................   Passed    0.28 sec
    Start 14: hash-slow-4
4/4 Test #14: hash-slow-4 ......................   Passed    0.72 sec
100% tests passed, 0 tests failed out of 4
```

Fixes #9371

### DIFF
--- a/src/crypto/slow-hash.c
+++ b/src/crypto/slow-hash.c
@@ -1166,12 +1166,17 @@ static const int rcon[] = {
 	0x01,0x01,0x01,0x01,
 	0x0c0f0e0d,0x0c0f0e0d,0x0c0f0e0d,0x0c0f0e0d,	// rotate-n-splat
 	0x1b,0x1b,0x1b,0x1b };
+const uint8_t *key_ptr = key;
+uint8_t *expanded_key_ptr = expandedKey;
+const int *rcon_ptr = rcon;
+int count;
+
 __asm__(
 "	eor	v0.16b,v0.16b,v0.16b\n"
 "	ld1	{v3.16b},[%0],#16\n"
 "	ld1	{v1.4s,v2.4s},[%2],#32\n"
 "	ld1	{v4.16b},[%0]\n"
-"	mov	w2,#5\n"
+"	mov	%w3,#5\n"
 "	st1	{v3.4s},[%1],#16\n"
 "\n"
 "1:\n"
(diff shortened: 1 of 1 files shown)


SChernykh (approved):

selsta:
Force pushed to apply feedback from @jeffro256

jeffro256:
The `volatile` qualifier is still present

jeffro256 (approved):

selsta:
Sorry, had to force push one more time to change spaces to tabs to match surrounding code.

SChernykh (approved):

jeffro256:
Eh, to be fair, the whole file uses a mix of spaces and tabs, seemingly without a pattern. Spaces were probably fine.

jeffro256 (approved):

PR #10937, opened 2026-07-20
wallet_api: return all unsigned tx recipients

### LINKED ISSUE #10919: request: wallet/api/unsigned_transaction.cpp deprecate recipientAddress() for confirmationMessage()
https://github.com/monero-project/monero/blob/0f84863feda75574788e15c9691abecf57bc3f8f/src/wallet/api/unsigned_transaction.cpp#L300

when testing offline signer wallets - i noticed cupcake had an issue with a pay-to-many transaction. they have a check to confirm [`(amts.length != addrs.length)`](https://github.com/cake-tech/cupcake/blob/83d2a14dfe095f8fe5ab4ffb467655aa0f17b111/lib/coins/monero/wallet.dart#L123) but are only given `dests[0]` where as amount returns [all](
https://github.com/monero-project/monero/blob/master/src/wallet/api/unsigned_transaction.cpp#L218-L227)

anonero also uses recipientAddress() . anonero-fork has provided a solution to move to [pay-to-many](https://codeberg.org/iuanv9/anonero-fork/commit/0bfb80471b5d7d125d6e57f976031ee4a33ce462)

reason for deprecation is this _could_ lead to a huge **gotcha** bug where consumers of the api using recipientAddress() only show dest[0] of a pay to many transaction - possibly leading to the cold wallet being drained. cupcake have protections against this but anonero does not (these are universal* cold signers so could be presented a pay-to-many transaction from a compatible wallet)

[monero-gui](https://github.com/monero-project/monero-gui/blob/9dc0c85c16da76cb3c79c0536a28c54c8a38ee66/src/libwalletqt/UnsignedTransaction.cpp#L90-L96
)  and [feather](https://github.com/feather-wallet/feather/blob/cbf966f5ba948b32fb76ad436ab7dca470239ccd/src/libwalletqt/UnsignedTransaction.cpp#L66-L72) are looping over the recipientAddress() is dead code?

monero-gui uses [m_confirmationMessage](https://github.com/monero-project/monero/blob/0232839913b13cf0ab0bb7ad25fff0c05f37d2fe/src/wallet/api/unsigned_transaction.cpp#L214) [here](https://github.com/monero-project/monero-gui/blob/9dc0c85c16da76cb3c79c0536a28c54c8a38ee66/pages/Transfer.qml#L1043) which contains all the destinations just fine.

Ap4sh:
## what

return every destination from `UnsignedTransaction::recipientAddress()` instead of only the first destination in each transaction

keep the returned addresses aligned with `amount()` and add a libwallet API regression test covering a two-recipient unsigned transaction

## why

wallets using `recipientAddress()` to review pay-to-many unsigned transactions currently receive an incomplete recipient list

fixes #10919

## testing

- full build with `BUILD_TESTS=ON` and `BUILD_GUI_DEPS=ON`
- `libwallet_api_tests`
- targeted red/green check for `WalletTest1.WalletUnsignedTransactionReturnsAllRecipients`

### DIFF
--- a/tests/libwallet_api_tests/main.cpp
+++ b/tests/libwallet_api_tests/main.cpp
@@ -747,8 +747,6 @@ TEST_F(WalletTest1, WalletTransaction)
     ASSERT_TRUE(wmgr->closeWallet(wallet1));
 }
 
-
-
 TEST_F(WalletTest1, WalletTransactionWithMixin)
 {
 
@@ -957,12 +955,16 @@ TEST_F(WalletTest1, WalletTransactionMultDestWithPaymentId)
     ASSERT_TRUE(payment_id.length() == 16);
     Monero::Wallet * wallet_dst = wmgr->openWallet(CURRENT_DST_WALLET, TESTNET_WALLET_PASS, WALLET_NETWORK_TYPE);
     ASSERT_TRUE(wallet_dst != nullptr);
+    const std::string destination_address = wallet_dst->mainAddress();
     std::string integrated_address = wallet_dst->integratedAddress(payment_id);
     ASSERT_FALSE(integrated_address.empty());
     ASSERT_TRUE(wmgr->closeWallet(wallet_dst));
 
     const std::vector<std::string> destinations{integrated_address, wallet_src->mainAddress()};
+    const std::vector<std::string> recipient_addresses{destination_address, wallet_src->mainAddress()};
     const std::vector<uint64_t> amounts{AMOUNT_1XMR, AMOUNT_1XMR};
+    const std::string unsigned_tx_filename = "unsigned-multi-destination-with-payment-id";
+    boost::filesystem::remove(unsigned_tx_filename);
     Monero::PendingTransaction * tx = wallet_src->createTransactionMultDest(
             destinations, PAYMENT_ID_EMPTY, amounts, 1,
             Monero::PendingTransaction::Priority_Medium, 0, std::set<uint32_t>{});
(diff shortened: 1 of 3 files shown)


tests/libwallet_api_tests/main.cpp
```diff
@@ -747,6 +747,43 @@ TEST_F(WalletTest1, WalletTransaction)
...
+    ASSERT_EQ(Monero::UnsignedTransaction::Status_Ok, unsigned_transaction->status()) << unsigned_transaction->errorString();
+    ASSERT_EQ(1, unsigned_transaction->txCount());
+    EXPECT_EQ(amounts, unsigned_transaction->amount());
+    EXPECT_EQ(recipient_addresses, unsigned_transaction->recipientAddress());
+
+    delete unsigned_transaction;
+    boost::filesystem::remove(unsigned_tx_filename);
+    ASSERT_TRUE(wmgr->closeWallet(wallet));
+}
+
```
SNeedlewoods: IMO the whole test seems a bit overkill for the change. Though I'm okay with it if no one else complains.
_____
Nit: I know there is already a wild mix for the amount of newlines between the functions in this file (ranging from 1-4), without any comprehensible reason, but I'd prefer to try bring some consistency with new code. So instead of having 1 newline before the function and 3 after it, make it 1 before and 1 after (or 2 before 2 after if you prefer that).

SNeedlewoods (approved):

tests/libwallet_api_tests/main.cpp
```diff
@@ -747,6 +747,43 @@ TEST_F(WalletTest1, WalletTransaction)
     ASSERT_TRUE(wmgr->closeWallet(wallet1));
 }
 
+TEST_F(WalletTest1, WalletUnsignedTransactionReturnsAllRecipients)
```
selsta: I would integrate this test (the check for amounts and recipient_addresses) with `WalletTransactionMultDestWithPaymentId`, this avoids a whole new test case while still testing the relevant code.

selsta:
@Ap4sh could you add @SNeedlewoods as co-author?

tests/libwallet_api_tests/main.cpp
Ap4sh: yep, folded this into the existing multi-destination test and cleaned up the spacing too, thanks

tests/libwallet_api_tests/main.cpp
Ap4sh: done, the amount and recipient address checks now live in `WalletTransactionMultDestWithPaymentId`

Ap4sh:
done, added @SNeedlewoods as co-author too

selsta (approved):

SNeedlewoods (approved):

plowsof (approved):

jpk68 (approved):

PR #10941, opened 2026-07-21
[DO NOT MERGE] cryptonote_core: explicitly require >0 coinbase outs

j-berman:
Ignore the rest of this description, going to implement [this](https://github.com/monero-project/monero/pull/10941#issuecomment-5170476930).
____
Still TODO: make sure that **all** coinbase txs in the mainnet chain (and testnet/stagenet chains) have at least 1 out. Syncing from scratch with this change would suffice. If such a sync fails, then this PR would need to be modified.

This rule is an explicit requirement needed for FCMP++, but should already be implicitly enforced at consensus since `HF_VERSION_EXACT_COINBASE`.

Note: since `validate_miner_transaction` [since `HF_VERSION_EXACT_COINBASE` requires coinbase outs to sum to the expected block reward + fees from the block](https://github.com/monero-project/monero/blob/9e1a88ba83cf8adac52d4eb019ab1c0d5dfff934/src/cryptonote_core/blockchain.cpp#L1417-L1419), then so long as we can sync up to that fork height with this change, then this change should be fine to introduce.

For further context on why this is necessary for FCMP++ see https://github.com/seraphis-migration/monero/issues/389.

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -1345,6 +1345,7 @@ bool Blockchain::prevalidate_miner_transaction(const block& b, uint64_t height, 
   LOG_PRINT_L3("Blockchain::" << __func__);
   CHECK_AND_ASSERT_MES(b.miner_tx.vin.size() == 1, false, "coinbase transaction in the block has no inputs");
   CHECK_AND_ASSERT_MES(b.miner_tx.vin[0].type() == typeid(txin_gen), false, "coinbase transaction in the block has the wrong type");
+  CHECK_AND_ASSERT_MES(!b.miner_tx.vout.empty(), false, "coinbase transaction in the block has no outputs");
   CHECK_AND_ASSERT_MES(b.miner_tx.version > 1 || hf_version < HF_VERSION_MIN_V2_COINBASE_TX, false, "Invalid coinbase transaction version");
 
   // for v2 txes (ringct), we only accept empty rct signatures for miner transactions,


jeffro256:
> but should already be implicitly enforced at consensus since `HF_VERSION_EXACT_COINBASE`.

Not necessarily. A block could have weight exactly equal to $2 M_N$, so that the block reward is exactly 0. Normally, this isn't likely to happen, but it could happen on purpose fairly trivially if a malicious miner fills a block up exactly by changing the size of contained transactions' `extra` fields by padding with 0s. In that case, people with this PR merged would fork from people who didn't have it merged. I think that the conservative option would be temper this check with `|| hf_version < HF_VERSION_FCMP_PLUS_PLUS`. Does it matter that the FCMP tree root wouldn't change *before* the first FCMP++ tx?

j-berman:
> it could happen on purpose fairly trivially if a malicious miner fills a block up

I assumed txs would need to have non-0 fees in this case

> I think that the conservative option would be temper this check with || hf_version < HF_VERSION_FCMP_PLUS_PLUS. Does it matter that the FCMP tree root wouldn't change before the first FCMP++ tx?

We'd just also have to require all FCMP++ txs use a tree root from the fork height or higher. It just adds a bit more complexity to the consensus logic. I don't mind it for the peace of mind here.

jeffro256:
> I assumed txs would need to have non-0 fees in this case

No, the penalty only "touches" the block emission. Tx fees are added on top after the penalty is taken successfully: https://github.com/monero-project/monero/blob/8472d8c503933bb59b75b017d0935ab7f3d999d4/src/cryptonote_core/blockchain.cpp#L1403-L1412. So a miner going for the network fork would pay the opportunity cost of 0.6 XMR (plus whatever other tx fees are in the mempool) worth of PoW once, to cause a consensus split.

> We'd just also have to require all FCMP++ txs use a tree root from the fork height or higher. It just adds a bit more complexity to the consensus logic. I don't mind it for the peace of mind here.

If you could work on writing up that solution as if we are going to use it, that would be great. In the best case, we can write the rule as being fork-dependent now, until the hard fork activates on mainnet, then check if there was ever a block with 0 *tree-valid* coinbase outputs, and finally, retroactively simplify the logic.

jeffro256:
Also, just thought about this. This rule, regardless of forking, doesn't guarantee that the FCMP++ tree root changes every block because the output pubkey of the 1 coinbase output is allowed to be malformed before v17.

j-berman:
Good points. I'll implement the solution requiring `reference_block >= fork height` and require >0 coinbase outs starting from the fork.

j-berman:
Hm, actually we'd probably want to require reference_block at least ~60 blocks (check off-by-1) after the fork height, since those are the first coinbase outs guaranteed >0 valid.

Edit: it's not ideal that we can't make FCMP++ txs until 60 blocks after fork. Thinking on it some more.

jeffro256:
How about this: if a tree extension for any given block height >= v17 fork height would otherwise be empty, then insert the output $(O^\star, C^\star)$, where $O^\star = H^2_p(\text{"Monero dummy output pubkey"})$ and $C^\star = G$. This may or may not disrupt the global index -> tree leaf index scheme.

earnifysu:
> This may or may not disrupt the global index -> tree leaf index scheme.

it does not, but something adjacent does, and that seems worth flagging before the idea gets built.

line numbers below are against `origin/pr/9440` (the fcmp++-stage integration branch), since that is where the db and wallet side actually exist.

## the index scheme itself is fine

there is no arithmetic relation to disrupt. `leaf_idx` is stored, not derived: `mdb_leaf { uint64_t leaf_idx; uint64_t unified_id; }` (`db_lmdb.cpp:365-368`), written at `db_lmdb.cpp:1443-1452`, and looked up by `find_leaf_idx_by_unified_id_bounded_search`, which is a linear scan comparing `it->unified_id == unified_id` (`db_lmdb.cpp:1763-1821`) bounded by the per block leaf counts in `mdb_tree_meta`.

and the relation is already both non monotonic and lossy:

- coinbase unlocks at `height + 59`, normal outputs at `height + 9` (`cryptonote_tx_utils.cpp:197`, `cryptonote_format_utils.cpp:1916-1919`), so block B's extension mixes outputs created at B-59 and B-9. across blocks, leaf order and `unified_id` order already interleave.
- `outputs_to_leaves` silently drops outputs that cannot produce a valid tuple (`curve_trees.cpp:1596-1612`), so there are already arbitrary gaps.

so index wise a synthetic leaf is just one more leaf, shifting subsequent indices exactly the way one more real output would.

## what does break: `get_path` throws for other people's outputs

this is the part i would not have guessed. `get_path` resolves **every** leaf in the returned chunk, not just the one being proven: it collects the chunk's `unified_id`s and calls `get_unified_output_by_id` on each (`db_lmdb.cpp:1637-1662`), which throws `OUTPUT_DNE` for an id with no `output_txs` row (`db_lmdb.cpp:5445-5453`).

so a dummy leaf does not only fail to resolve itself. it poisons its whole chunk, and every real output sharing that chunk becomes unprovable. that is up to `chunk_width - 1` genuine users per dummy, and they have done nothing unusual.

`audit_tree` has the same assumption (`db_lmdb.cpp:2123-2150`), and `trim_leaves` resolves removed leaves' ids to re add them to `locked_outputs` on reorg (`db_lmdb.cpp:1863-1881`), so a reorg past a dummy throws too.

wallet side, `TreeCache::prepare_to_grow_cache` attributes leaves to blocks by matching `unified_id`s in sorted order (`tree_cache.cpp:876-893`) and asserts `n_unlocked_outputs >= tuples.size()` (`tree_cache.cpp:869`). the empty extension case is exactly `0 >= 1`, so that assert fires, and a non matching leaf also stalls the iterator into "did not reach all leaf tuples" (`tree_cache.cpp:895`). the wallet re derives leaf indices by replaying `get_tree_extension` locally rather than being told them (`tree_cache.cpp:863-868`), so it would need the same dummy rule, derived from the same inputs.

none of that is fatal to the idea, but it means the dummy needs a reserved sentinel id that every leaf-to-output resolution site knows to skip, rather than just being inserted. note `unified_id == 0` would alias the genesis output, so the sentinel wants to be `UINT64_MAX` or similar.

## `O*` itself checks out

i computed it, since it seemed worth knowing before the rest matters. compiling monero's own `crypto-ops.c`, `keccak.c` and `blake2b.c` and evaluating both readings of `H^2_p`:

| reading | O* | decompresses | identity | torsion free | Ed x nonzero |
|---|---|---|---|---|---|
| `unbiased_hash_to_ec("Monero dummy output pubkey")` | `153e6f2a8e9165f101bdd972ff90f9558dd5f4123af7716391e3ac4ef4f9946a` | yes | no | yes | yes |
| `unbiased_hash_to_ec(Keccak256("Monero dummy output pubkey"))` | `16ac5210eb238bcafa61814f73fe2352173b52559a15dc3a6fa6f5a0c824361b` | yes | no | yes | yes |

so either reading gives a usable leaf point, including the nonzero Ed x that `point_to_ed_derivatives` needs. worth pinning which one is meant, since `generators.cpp:99` documents `H^2_p` as applying to `Keccak256(seed)` while the proposal reads as applying to the string directly.

## two smaller things

**leaf tuples are three points, not two.** the proposal gives `(O*, C*)`, but a leaf is `{O, I, C}` and `I` is derived from the original output pubkey (`curve_trees.cpp:136-140`, flattened at `curve_trees.cpp:686-700`). `I*` needs specifying too, along with which of the biased or unbiased hash to point it uses.

**"would otherwise be empty" is ambiguous in the code as written.** `grow_tree` has two distinct early returns: `new_outputs.empty()` at `blockchain_db.cpp:376`, and `tree_extension.leaves.tuples.empty()` at `blockchain_db.cpp:387` after the validity filter. only the second one covers the malformed pre v17 coinbase pubkey case you raised on 07-24, so the dummy wants to be inserted after filtering rather than before.

## on the alternative

for what it is worth, this seems clearly better than `reference_block >= fork_height + 60`. that one costs every user an hour of no fcmp++ transactions right at the fork, which is when scrutiny is highest, in exchange for avoiding a case that requires a miner to burn about 0.6 xmr. the dummy confines the complexity to the daemon and wallet rather than exposing it as a user visible restriction.

things i did not verify: the SAL side, since `monero-fcmp-plus-plus` is an external crate, so i cannot say whether anything could construct a spend proof against `O*`, though an unknown discrete log should settle it. i also did not execute the `MDB_SET` cursor behaviour in `find_leaf_idx_by_unified_id_bounded_search`, only read it.

j-berman:
@kayabaNerve proposed what seems an elegant solution to this:

- In `pre / validate_miner_transaction`, if `unlock_time >= FCMP++ fork height` then require at least 1 coinbase out, and start doing the FCMP++ torsion validation to guarantee (at least 1) coinbase outs would enter the tree.
- Require FCMP++ txs' `reference_block >= FCMP++ fork height`.

This would effectively introduce a soft fork 60 blocks before the actual hard fork height, which seems ok. Newer version nodes would potentially fail to validate coinbase txs that older version nodes could theoretically allow, but:
1) older version nodes are going to fork off very shortly after that, and
2) it likely won't end up even happening since most are expected to be running the newer version by that point anyway.

earnifysu:
that gate is exact, and conveniently the value is already in scope where the check would go.

`prevalidate_miner_transaction` (`blockchain.cpp:1344-1377`) already enforces the coinbase unlock time at `blockchain.cpp:1363`:

```c++
CHECK_AND_ASSERT_MES(b.miner_tx.unlock_time == height + CRYPTONOTE_MINED_MONEY_UNLOCK_WINDOW, false, ...)
```

so `unlock_time` is exactly `height + 60` on every accepted block, not miner chosen, which makes `unlock_time >= fork_height` precisely equivalent to `height >= fork_height - 60` with nothing to evade. and it is checked three lines above where the new `!vout.empty()` check sits.

one consequence worth noting: this removes the TODO in the description. gated this way the rule only ever applies to blocks at `height >= fork_height - 60`, all of which are still in the future, so no historical coinbase tx is ever evaluated against it and there is nothing to confirm by syncing from scratch. that also drops the retroactive fork risk jeffro raised on 07-21 entirely, rather than mitigating it.

jeffro256:
@j-berman I like that. For posterity's sake, I'll toss out the other idea I had, which was to, when fetching FCMP tree roots for reference blocks within 60 blocks of the v17 fork height, also fetch the immediately preceding FCMP tree root, then reject it if they were equal. When constructing FCMP++ txs within 60 blocks of the v17 fork, wallets would request the block root over RPC from the daemon, and would decrement their reference block until it was the oldest in that span of 0 new outputs. Alternatively, wallets could drop the last step and simply be optimistic, and it would almost certainly be true. Worst case, some miner gets a streak of 60 blocks with 0 insertable outputs. If that happens, we are almost certainly under a near-or-over-51% attack anyways.

But I like @kayabaNerve's idea better. The only downside is that it would double up on work for torsion checking for 60 blocks worth of coinbase outputs, but that's not bad at all.

Issue #10365, opened 2026-03-13
monerod mining crashes inconsistently on Windows 11


Wyrade:
Hi!

(I'm new, might be doing beginner mistakes.)

Using "monero-gui-win-x64-v0.18.4.7.zip" release.
I currently have a working monerod with full blockchain, with an xmrig successfully mining using it.
Intel i5-13600K processor (6 P cores hyperthreading to 12 threads, 8 E cores, totaling in 20 threads), Windows 11.
Monerod and the blockchain are on an m.2 nvme ssd (ntfs) inside the PC currently, but several of the tests below where done while the blockhain was on a (decently fast) pendrive's btrfs partition with a btrfs driver installed on win11. Still, the same crashing behavior continued after I moved the blockchain to the m.2 ssd.
Same crashing happened with windows real time protection turned off, and I've also added the folder and monerod specifically as exclusions.

I start monerod  with `monerod --config-file .\config\monerod.conf` from a cmd.

As soon as I start mining with 10 threads, it crashes (closes) after a few seconds. I can't see anything relevant in bitmonero.log, even at log level 2. For some reason, it doesn't crash on log level 3 and 4, or i just didn't wait enough or idk (but obviously I don't want to actively run long-term on that log level). A few times It also didn't crash with log level 0 and 2 (haven't tried log 1 much). I have a log saved with log level 2 that crashed and one that didn't, but I don't see any glaring entries that indicate why one would have crashed. It does crash more often than not.

I started to experiment a bit more after that. Log level 0.
I set the affinity to 0xff500 with `start /affinity 0xff500 monerod --config-file .\config\monerod.conf` (meaning one thread from two separate P cores, and all 8 E cores are availabe).
2 threads seem to work fine, and two of the P cores work on 100%.
with 3 threads, mostly 2 of the P cores are on 100%, and some of the E cores are running various amounts, none 100%. I've had it crash once with 3 threads after i sent the stop_mining command, but a few other times it worked fine.
4 threads were similar and similar perf to 3 threads, and stop_mining crashed it again.

I also notice that sometimes the cmd window starts visually different, in a smaller size/font, idk why.
And rarely, it started up noticeable slower than most other times, in fact i opted to close those and restart monerod.

Next two times, starting the 4 threads crashed it after a few seconds.

Starting it with 0xfff affinity to exclude the E cores and keep only the P cores, starting to mine on 12 threads also crashes it. 
4 threads seems to work. 6 thread worked too. starting 7 threads crashed it.

[...]

PR #10953, opened 2026-07-23
Fix TryAcquireSRWLockShared being implicitly defined and causing UB

### LINKED ISSUE #4254: Solo mining resulting in crashing Daemon immediately Windows 11 Pro
After synchronizing my wallet, I went to try & start mining in solo mode resulting in the network losing connection & goes into a loop of trying to reconnect to the network. I believe the node is crashing because I am able to manually start it again. Once I restart the node manually, I change the settings, resulting in the same issue or I get a message stating 'Mining temporarily suspended.'. I have this issue weather I am using half or all of my CPU's threads..  When I try to use the "Background mining" thread, I seem to be having the same issue. Let me know what information I can provide that would be helpful.
Thank you!

SChernykh:
Fixes https://github.com/monero-project/monero-gui/issues/4254 and #10365

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -751,9 +751,8 @@ include(CheckTrezor)
     # mingw doesn't support LTO (multiple definition errors at link time)
     set(USE_LTO_DEFAULT false)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--stack,10485760")
-    if(NOT BUILD_64)
-      add_definitions(-DWINVER=0x0600 -D_WIN32_WINNT=0x0600)
-    endif()
+    # TryAcquireSRWLock* requires Windows 7 or later.
+    add_definitions(-DWINVER=0x0601 -D_WIN32_WINNT=0x0601)
   endif()
   set(C_WARNINGS "-Waggregate-return -Wnested-externs -Wold-style-definition -Wstrict-prototypes")
   set(CXX_WARNINGS "-Wno-reorder -Wno-missing-field-initializers")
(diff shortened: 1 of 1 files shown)


selsta (approved):

tobtoht (approved):

selsta (approved):

jpk68 (approved):

monerostar (commented):
## monerostar review (native Windows 11)

**Environment**
- Host: Windows 11 (native), monerostar workstation
- Installed binary: `C:\Program Files\Monero GUI Wallet\monerod.exe`
- Version: `Monero 'Fluorine Fermi' (v0.18.5.0-release)` (GUI package; May 2025-era build stamp on the exe)
- This host runs monerod / xmrig fleet ops but does **not** currently reproduce solo `start_mining` crashes on the installed 0.18.5.0 release under normal pool mining (different workload than GUI solo-mine reports on #10365 / monero-gui#4254)

**Diff review**
- Raises MinGW `WINVER` / `_WIN32_WINNT` from Vista (`0x0600`, and only when `NOT BUILD_64`) to Windows 7+ (`0x0601`) for all MinGW builds — required for proper `TryAcquireSRWLock*` prototypes.
- Adds `-Werror=implicit-function-declaration` for MinGW C so a missing WinAPI prototype fails the build instead of becoming a silent wrong-calling-convention call (UB / intermittent crash class).
- release-v0.18 twin: #10954 — same idea; prefer both landing together.

**CI**
- Observed green checks on this PR including **Win64** (depends) and **Windows (MSYS2)** (cli) — the right targets for this change.

**Live binary rebuild**
- Full MSYS2 `make release-static` of this PR was **not** run on this pass (multi-hour). Happy to smoke a PR build later if maintainers want an independent Win11 binary check against solo mining.

**Verdict**
- Root cause + fix shape look correct; no competing PR needed from me.
- LGTM from a Windows operator / code-read perspective; merge confidence high given MSYS2 CI green.

Related: #10365, monero-gui#4254

PR #10957, opened 2026-07-23
dns_utils: replace flaky DNS provider

selsta:
Control D frequently appears to fail lookups on CI.


jpk68 (approved):
Looks alright to me. I forgot to comment this on the last PR, but it might be good to also keep the country/jurisdiction info next to each provider.

tobtoht (approved):

PR #10960, opened 2026-07-23
Optimize hashing of generic types

SChernykh:
Use a proper hash function instead of f(x)=x to achieve better distribution of hash values

### DIFF
--- a/src/crypto/generic-ops.h
+++ b/src/crypto/generic-ops.h
@@ -32,9 +32,13 @@
 
 #include <cstddef>
 #include <cstring>
+#include <cstdint>
 #include <functional>
 #include <memory>
 #include <sodium/crypto_verify_32.h>
+#include <sodium/crypto_shorthash_siphash24.h>
+
+#include "random.h"
 
 #define CRYPTO_MAKE_COMPARABLE(type) \
 namespace crypto { \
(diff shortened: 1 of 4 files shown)


src/crypto/generic-ops.h
```diff
@@ -57,22 +60,29 @@ namespace crypto { \
...
 
+namespace crypto {
+  inline std::size_t siphash_to_size_t(const void *data, std::size_t length) {
+    static_assert(sizeof(crypto_siphash_key) == crypto_shorthash_siphash24_KEYBYTES,
+      "crypto_siphash_key size must match the SipHash-2-4 key length");
+    unsigned char digest[crypto_shorthash_siphash24_BYTES];
+    crypto_shorthash_siphash24(digest, static_cast<const unsigned char *>(data), length, crypto_siphash_key);
+    std::size_t h = 0;
+    memcpy(&h, digest, sizeof(h) < sizeof(digest) ? sizeof(h) : sizeof(digest));
+    return h;
```
jeffro256:
```suggestion
    static_assert(sizeof(std::uint64_t) == crypto_shorthash_siphash24_BYTES,
      "std::uint64_t size must match the SipHash-2-4 digest length");
    std::uint64_t h;
    crypto_shorthash_siphash24(&h, static_cast<const unsigned char *>(data), length, crypto_siphash_key);
    return h;
```

jeffro256 (changes requested):
This was long overdue, thanks for looking into it.

src/crypto/generic-ops.h
SChernykh: Passing `uint64_t` directly to `crypto_shorthash_siphash24` would be a strict aliasing violation (it accepts `unsigned char*` as the first parameter) - an undefined behavior, this is why I do `memcpy`.

jeffro256: That's not a strict aliasing violation because there's an exception for character types: https://en.cppreference.com/c/language/object#Strict_aliasing.

vtnerd (approved):
This does prevent someone from stuffing bogus `public_key`s into some maps that may exist.

src/crypto/generic-ops.h
```diff
@@ -57,22 +61,29 @@ namespace crypto { \
...
   } \
 } \
 namespace std { \
   template<> \
   struct hash<crypto::type> { \
     std::size_t operator()(const crypto::type &_v) const { \
-      std::size_t h; \
-      memcpy(&h, std::addressof(_v), sizeof(h)); \
-      return h; \
+      return ::crypto::siphash_to_size_t(std::addressof(_v), sizeof(_v)); \
```
vtnerd, line 86: I just approved this, but realized this line may not work. @jeffro256 pointed out there is an exception for character types, but technically this is the outer wrapper here.

vtnerd (changes requested):

src/crypto/generic-ops.h
SChernykh: There is no issue at this line, the parameter type is `const void*`. Character types was in regards to passing `uint64_t*` (local pointer) to the inner function which accepts `unsigned char*`.

SChernykh: `_v` is cast to `const unsigned char*` too, eventually. It's fine as long as `_v` is a simple type (it is because it was memcpy'd before). I could add a static assert to check this. `static_assert(std::is_trivially_copyable...)` probably.

vtnerd: Right, messed this up, sorry for the noise.

selsta (approved):

PR #5549, opened 2019-05-16
RandomX

hyc:
RandomX PoW integration completed. Running successfully on private testnet.
Also requires https://github.com/tevador/RandomX/pull/40 before you can build this. [Already merged, ignore this note.]

You should edit the testnet blockheight and timestamp before running on your own private testnet. Also, I haven't tested yet on Windows, any other Windows devs trying that out would be appreciated.

PR #10964, opened 2026-07-24
external: add mx25519

jeffro256:
Adds a submodule for https://github.com/tevador/mx25519, a pre-requisite for [`carrot_core`](#9559). Currently, the commit/branch points to the current `master` branch on tevador's repository, with the following PRs merged:
  - https://github.com/tevador/mx25519/pull/10
  - https://github.com/tevador/mx25519/pull/11
  - https://github.com/tevador/mx25519/pull/13
  - https://github.com/tevador/mx25519/pull/14
  - https://github.com/tevador/mx25519/pull/15
  - https://github.com/tevador/mx25519/pull/16
  - https://github.com/tevador/mx25519/pull/17

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -11,3 +11,6 @@
 [submodule "external/gtest"]
 	path = external/gtest
 	url = https://github.com/google/googletest.git
+[submodule "external/mx25519"]
+	path = external/mx25519
+	url = https://github.com/tevador/mx25519.git
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -30,5 +30,6 @@
 
 add_subdirectory(db_drivers)
 add_subdirectory(easylogging++)
+add_subdirectory(mx25519)
 add_subdirectory(qrcodegen)
 add_subdirectory(randomx EXCLUDE_FROM_ALL)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -385,6 +385,7 @@ if(NOT MANUAL_SUBMODULES)
     endfunction ()
     
     message(STATUS "Checking submodules")
+    check_submodule(external/mx25519)
     check_submodule(external/rapidjson)
     check_submodule(external/randomx)
     check_submodule(external/supercop)
@@ -460,7 +461,13 @@ elseif(CMAKE_SYSTEM_NAME MATCHES ".*BSDI.*")
   set(BSDI TRUE)
 endif()
 
-include_directories(external/rapidjson/include external/easylogging++ src contrib/epee/include external external/supercop/include)
+include_directories(external/rapidjson/include
+  external/easylogging++
+  external/mx25519/include
+  src
+  contrib/epee/include
+  external
+  external/supercop/include)
 
 option(STATIC "Link libraries statically" OFF)
 


.gitmodules
```diff
@@ -11,3 +11,7 @@
 [submodule "external/gtest"]
 	path = external/gtest
 	url = https://github.com/google/googletest.git
+[submodule "external/mx25519"]
+	path = external/mx25519
+	url = https://github.com/jeffro256/mx25519.git
```
vtnerd: Is @tevador not accept merge requests? Or what happened here.

tevador: I do accept PRs.

jeffro256: Okay, submodule now points to latest in tevador's repo

tevador (approved):

selsta (approved):

tobtoht:
Is there any reason why this library needs to be maintained out of tree? (I.e. why not vendor / git subtree?)

I can think of a few arguments to not add this as a submodule:

- If we ever need to make a change to this library and tevador can't be reached for any reason, we will need to vendor or fork it anyway.
- In-tree changes have more visibility and get merged in accordance with our contributor guidelines.
- Working with git submodules sucks in general.

jeffro256:
The main reason is that mx25519 is a library which serves a very general purpose that isn't Monero-specific. I can't speak to @tevador's commitment to being available to maintain, but I also don't see changes happening very often from here on out besides bug fixes.

tobtoht:
>but I also don't see changes happening very often from here on out besides bug fixes.

This seems like all the more reason to just copy and paste it into external. If the code has essentially no velocity, why add an extra hoop for bug fixes?

tevador:
I'm not sure how licenses are handled in the external directory, but mx25519 is licensed under LGPLv3, which might be an argument to leave it in a separate repository.

tobtoht:
Is that because of the LGPLv3 or because of:

>All contributions to the project source code ("patches") MUST use the same license as the project.

We vendor several libraries in `external` that do not use the BSD-3 license. ~If adding a submodule isn't "contributing to the project source code", vendoring isn't either.~

tevador:
Legally it should be OK, I'm just noting it would be the first LGPL licensed vendored dependency. The rest seems to be various kinds of permissive licenses.

tobtoht:
@tevador 

>mx25519 is licensed under LGPLv3

There is precedent that we do not allow GPL code: RandomX was [re-licensed from GPL to BSD-3](https://github.com/tevador/RandomX/commits/master/LICENSE). 

You were [aware of this incompatibility](https://github.com/monero-project/monero/pull/5549#issuecomment-493747266). Why choose to license mx25519 under the LGPLv3?

tevador:
GPL is not compatible with BSD-3.
LGPL is compatible with BSD-3.

RandomX used to be licensed under GPL, which would have required Monero to switch to GPL. LGPL doesn't have this problem.

tevador:
FYI, as a historical note: relicensing RandomX under BSD-3 was a mistake. I should have chosen LGPL. The reason is that the RandomX library is being used with optimizations in many closed-source miners. LGPL would have required them to publish the optimizations under the same license.

tobtoht:
>RandomX used to be licensed under GPL, which would have required Monero to switch to GPL.

Ok, it's clear to me now that the incompatibility you were speaking of in the comment I linked to is in reference to the downstream virality of the GPL license, not to an incompatibility with the CoC.

>All contributions to the project source code ("patches") MUST use the same license as the project.

For all intents and purposes, this _is_ a contribution to the project source code. `mx25519` will become a required dependency.

So far, we have clearly allowed required dependencies to use licenses other than BSD-3 so long as they are permissive (*).

If we allow non-permissive licenses in required dependencies, the rule becomes meaningless: contributors can just work around the rule by introducing code with a non-permissive license via a submodule, which is what is happening here.

Merging a non-optional LGPLv3 dependency would also add additional requirements for downstream projects, like pointing to the corresponding source, accompanying release binaries with copies of the GPLv3 and LGPLv3 license text, and complying with the LGPLv3 4 d 0 relinking requirement. 

This isn't something I would expect from a permissively licensed project and I think it goes against the spirit of the rule from the CoC.

(*) libzmq has a weak copyleft license and while our cmake currently requires it, it is not used by wallets (which is what is most relevant for downstream projects) and can easily be turned into an optional dependency for the daemon.

I think the inclusion of a LGPLv3 licensed project as a required dependency would need to be voted on because I don't think it's a decision I can make on my own. The alternative is to relicense under a permissive license.

tevador:
> All contributions to the project source code ("patches") MUST use the same license as the project.

Yes, *technically* mx25519 and polyseed don't meet this requirement, but practically speaking, it's perfectly fine for Monero to link to them and stay licensed under BSD-3.

The requirements are quite mild. Binary distributions would need to include a copy of the LGPLv3 license (+ copyright notices, but those are equally required by BSD-3). The relinking requirement is entirely met by the following:

* Monero is open-source
* the build instructions are provided
* the repository links to the exact commit of the library source

The above points allow anyone to easily link to a different version of the library.

> Merging a non-optional LGPLv3 dependency would also add additional requirements for downstream projects [...]
>
> [..] wallets (which is what is most relevant for downstream projects) [...]

The polyseed library (licensed under LGPLv3) is already being used by several 3rd party wallets, so I think this a weak argument.

> I think the inclusion of a LGPLv3 licensed project as a required dependency would need to be voted on because I don't think it's a decision I can make on my own.

I have no problem with it being voted on. I think it's quite sensible to allow the two libraries to be linked.

> The alternative is to relicense under a permissive license.

They won't be relicensed for the same reasons I consider the relicensing of RandomX to be a mistake.

tobtoht:
>would need to include a copy of the LGPLv3 license

This is purely informational and pedantic, but you need to include both the GPLv3 and LGPLv3 license text. The LGPLv3 is an extension of the GPLv3:

>This version of the GNU Lesser General Public License incorporates the terms and conditions of version 3 of the GNU General Public License, supplemented by the additional permissions listed below.

and

>b) Accompany the Combined Work with a copy of the GNU GPL and this license document.

---

>The above points allow anyone to easily link to a different version of the library.

Yes, but GPLv3 section 6 is not yet met:

>If the place to copy the object code is a network server, the Corresponding Source may be on a different server (operated by you or a third party) that supports equivalent copying facilities, provided you maintain clear directions next to the object code saying where to find the Corresponding Source.

So, it is not enough to point _to_ the library  _from_ the repo, you must also point to the library from the binary or release archive. It may be sufficient to point to this repo as it in turn points to the corresponding source. 

However, something interesting happens if you were to take down the mx25519 repo or get banned from GitHub:

>Regardless of what server hosts the Corresponding Source, you remain obligated to ensure that it is available for as long as needed to satisfy these requirements.

As I understand it, downstream projects would still be obligated to point to a copy of the corresponding source, in which case we would also need to either fork or vendor the library. 

If we vendored the library it would simplify our obligations under the LGPLv3 as the repo would already contain a copy of the corresponding source of the library,

---

>The polyseed library (licensed under LGPLv3) is already being used by several 3rd party wallets

This is true, though I will point out that permissively licensed re-implementations of polyseed [do exist](https://github.com/cake-tech/polyseed_dart). I don't know how many wallets use this or whether its use has anything to do with licensing.

tevador:
> If we vendored the library it would simplify our obligations under the LGPLv3 as the repo would already contain a copy of the corresponding source of the library,

OK, so I think vendoring sounds quite reasonable due to this requirement. Each release can then point to a specific tag of the Monero repository.

> If we allow non-permissive licenses in required dependencies, the rule becomes meaningless: contributors can just work around the rule by introducing code with a non-permissive license via a submodule, which is what is happening here.

FWIW, I think it's clear that both libraries are non-trivial and usable outside of the Monero repository. I didn't implement them to circumvent a rule. Granting this exception does not easily extend to other contributions, which are rarely self-contained independent libraries.

jpk68 (approved):

jeffro256:
> As I understand it, downstream projects would still be obligated to point to a copy of the corresponding source, in which case we would also need to either fork or vendor the library.

I don't think that this is the case, I think that updating to vendoring / forking post-takedown would suffice AFAICT. Do the users of LGPL really intend that 1% Github server downtime should constitute breaking the license by downstream? I think that we should cross the "mx25519 getting removed from Github" bridge when we get there. If mx25519 was removed off of Github, then we would notice pretty quickly with a git submodule, as CI would no longer be able to build. Until then, I think that it satisfies the source copy requirement.

> So, it is not enough to point to the library from the repo, you must also point to the library from the binary or release archive. It may be sufficient to point to this repo as it in turn points to the corresponding source.

Doesn't vendoring a LGPL dependency just mean that downstream still has to include instructions specific to finding mx25519, but now the link changes to "go to X subdirectory of github.com/monero-project/monero"?

tevador:
After some consideration, I think I'd be willing to relicense from LGPLv3 to MPL 2.0, which should provide sufficient protection with less friction for Monero.

For mx25519 specifically, @jeffro256 would have to agree as well due to his contributions.

vtnerd:
I had one contribution to that @jeffro256 repo (listed as co-author on a commit), specifically to some arm64 which appears to have survived edits, and I approve of a relicense to MPL.

tobtoht:
>The reason is that the RandomX library is being used with optimizations in many closed-source miners. 

I can see why someone would publish a closed-source miner: optimize the code, add a fixed donation %, claim that it is more profitable than open source miners. 

What incentive would someone have to keep mx25519 optimizations for themselves?

---

Will a share-alike license have the desired effect?

Significant portions of mx25519 were copied or derived from permissively licensed sources (which is mentioned in the README):

- The portable implementation from ref10: https://bench.cr.yp.to/supercop.html (public domain dedication).
- The ARM64 implementation from: https://github.com/Emill/X25519-AArch64 (CC0 1.0)
- The AMD64 implementation from https://github.com/kn-cs/x25519 (BSD-3)
  - (at&t to intel syntax, or a different source) 

It seems to me that anyone who wants to modify the code (and for whatever reason does not want to contribute back) can argue that they are optimizing the permissively licensed implementations. Or: how much of mx25519 is original work?

iamamyth:
> The reason is that the RandomX library is being used with optimizations in many closed-source miners.

Before adding restrictive terms to Monero to counter this kind of behavior, please consider the counterfactual: If RandomX were copyleft licensed, and everyone had abided by that license, would the closed-source optimizations have existed? After all, it seems reasonable to presume that the closed-source miners made their optimizations due to profit motive, else they would have published the source code. Yet, the closed-source versions still proved beneficial: 1) They decreased the advantage ASICs had over general-purpose CPU miners; 2) They raised the performance floor for CPU mining software: By establishing further optimizations were possible, they provided a better performance target for the open source implementations (even though they hid the path to the target, they illuminated the target itself, which is better than nothing).

tevador:
There are 3 options:

1. Accepting mx25519 as a MPL-2 dependency.
2. Clean-room reimplementation of mx25519 from the original sources.
3. Using the portable ref10 code, which is already included in Monero and can achieve the required functionality without much extra work.

Let me know if you choose the first option, so we can relicense mx25519 under MPL-2 (otherwise it will stay under LGPL-3).

tevador:
> All contributions to the project source code ("patches") MUST use the same license as the project.

FYI, the following code already violates this CoC rule:

* src/device_trezor/trezor/protob ([LGPL-3](https://github.com/monero-project/monero/blob/20d9d3cb3155bfb1cf8cf1686ab83dc99b870cc9/src/device_trezor/trezor/protob/COPYING))
* rapidjson (uses a [license](https://github.com/Tencent/rapidjson/blob/24b5e7a8b27f42fa16b96fc70aade9106cf7102f/license.txt) that is [non-free](https://www.gnu.org/licenses/license-list.html#JSON))
* external/db_drivers/liblmdb ([OpenLDAP Public License v2.8](https://github.com/monero-project/monero/blob/20d9d3cb3155bfb1cf8cf1686ab83dc99b870cc9/external/db_drivers/liblmdb/LICENSE), stronger than BSD-3)
* various files under CC0, BSD-2, MIT and BSL-1 that only *technically* violate the rule (all are weaker than BSD-3)

Due to the vendored trezor code, the official release binaries contain an LGPL-3 component, so the LGPL-3 requirements already apply (for example, including the full text of the license with the binaries, the required notices and source code availability obligations).

tobtoht:
From earlier in the thread:

>So far, we have clearly allowed required dependencies to use licenses other than BSD-3 so long as they are permissive

The question is: will we allow restrictive licenses in non-optional parts of the codebase?

---

>src/device_trezor/trezor/protob 

Trezor is an optional dependency for wallets. Downstream projects can disable it with `-DUSE_DEVICE_TREZOR=OFF`.

>rapidjson (uses a [license](https://github.com/Tencent/rapidjson/blob/24b5e7a8b27f42fa16b96fc70aade9106cf7102f/license.txt) that is [non-free](https://www.gnu.org/licenses/license-list.html#JSON))

This only applies to the optional jsonchecker feature, which we don't use. From the license:

"To avoid the problematic JSON license in your own projects, it's sufficient to exclude the bin/jsonchecker/ directory, as it's the only code under the JSON license."

>external/db_drivers/liblmdb ([OpenLDAP Public License v2.8](https://github.com/monero-project/monero/blob/20d9d3cb3155bfb1cf8cf1686ab83dc99b870cc9/external/db_drivers/liblmdb/LICENSE), stronger than BSD-3)

Aside from having an upgrade mechanism, it is identical to BSD-3 practically speaking. It does not have a share-alike element. The inclusion of LMDB predates the CoC.

---

>Due to the vendored trezor code, the official release binaries contain an LGPL-3 component

It's actually worse than that. The official binaries include [readline](https://github.com/gnu-mirror-unofficial/readline/blob/master/COPYING) which is licensed under GPLv3. Replacing readline with [linenoise](https://github.com/antirez/linenoise) has been on my to-do list for some time, because of its simpler implementation. I only found out about the license issue recently.

I will point out that readline is an optional dependency, so this doesn't necessarily affect downstream projects.

tevador:
> non-optional parts

How about the following solution (this is the final concession I'm willing to make):

1. We split mx25519 into two parts (the same repository, split on file boundaries): a BSD-3 part (core + portable) and an MPL-2 part (the assembly backends).
2. When built with `USE_COPYLEFT=ON`, the library will include the MPL-2 code and `mx25519_select_impl(MX25519_TYPE_AUTO)` will return the fastest supported backend.
3. When built with `USE_COPYLEFT=OFF`, the library will exclude the MPL-2 code and `mx25519_select_impl(MX25519_TYPE_AUTO)` will always return the portable backend.
4. The default build will use `USE_COPYLEFT=ON`

tobtoht:
Historically we have permitted copyleft licenses in optional parts of the codebase and I include that context in my interpretation of the CoC. One could argue at most about whether any new dependency is _sufficiently optional_. Relevant questions to ask could be: do downstream projects lose out on interoperability? Or, how much does disabling the feature affect performance? For example: a 99% performance hit isn't optional in any practical sense, even if the feature can technically be turned off. 

In your proposal, the only trade-off is performance. Does carrot performance become so slow under ref10 as to make using the assembly backends a practical requirement? Maybe not. Even if, downstream projects concerned with performance can add an existing permissively licensed optimized implementation to the library without taking on the MPL-2 license requirements.

All things considered, I think this is a pragmatic solution.

tevador:
> Does carrot performance become so slow under ref10 as to make using the assembly backends a practical requirement?

Benchmark of mx25519@dbb241f on my AMD test machine:

```
> ./mx25519-bench
    bench_portable                           ... 331557 cycles/op.
    bench_arm64                              ... N/A
    bench_amd64                              ... 128093 cycles/op.
    bench_amd64x                             ... 98076 cycles/op.

```

Monero was using ref10 code for scanning prior to the 2020 upgrade to supercop, so I'd say the performance is usable and the Carrot key exchange is even slightly faster than the legacy key derivation.

I can't imagine any project that would not accept the copyleft code, so this is only a theoretical exercise anyway.

> Historically we have permitted copyleft licenses in optional parts of the codebase and I include that context in my interpretation of the CoC.

Perhaps the CoC rule could be updated to reflect the current policy.

tobtoht:
>Perhaps the CoC rule could be updated to reflect the current policy.

I'd rather propose we disallow copyleft code in any new dependency going forward. I think almost no downstream project is going to care or even notice the new license requirements. That is, until a contributor starts a compliance trolling campaign.

I'm reminded of this quote from former busybox maintainer Rob Landley:

>From a purely pragmatic perspective: I spent over a year doing busybox license enforcement, and a dozen lawsuits later I’m still unaware of a SINGLE LINE OF CODE added to the busybox repository as a result…

Nothing is gained from copyleft except for legal risk and less freedom.

hyc:
> > All contributions to the project source code ("patches") MUST use the same license as the project.
> 
> FYI, the following code already violates this CoC rule:

My interpretation is "project source code" only applies to project-written code, and not to any of
its dependencies. Whether those dependencies are vendored in the source tree, referenced as
submodules, or just 3rd party libraries is immaterial. Vendoring and submodule references are
just a convenience mechanism, they're still independent 3rd party code and are not "project source code".

E.g. Boost is a dependency, not vendored, not a submodule, just 3rd party code and not subject to
any project license requirements.

>     * src/device_trezor/trezor/protob ([LGPL-3](https://github.com/monero-project/monero/blob/20d9d3cb3155bfb1cf8cf1686ab83dc99b870cc9/src/device_trezor/trezor/protob/COPYING))

This is an interesting example; since it lives under src/ and not external/ then that looks to me
like project source code.

tevador:
> I'd rather propose we disallow copyleft code in any new dependency going forward. I think almost no downstream project is going to care or even notice the new license requirements. That is, until a contributor starts a compliance trolling campaign.

The BSD-3 license terms often get violated as well, even by Monero itself (specifically the attribution clause for binary distributions). If you want to preempt any violations, you should prefer the CC0-1.0 license instead.

> Nothing is gained from copyleft except for legal risk and less freedom.

Your opinion is not shared by millions of projects that use a copyleft license. Open source code released under a permissive license only subsidizes large corporations.

For example, the FSF [recommends](https://www.gnu.org/licenses/license-recommendations.html) to use a copyleft license for most software, with just a few exceptions.

tobtoht:
>If you want to preempt any violations, you should prefer the CC0-1.0 license instead.

The FSF and OSI [do not recommend](https://www.gnu.org/licenses/license-list.html#CC0) CC0 for software because it explicitly withholds patent rights.

Permissive licenses without an attribution clause are [0BSD](https://opensource.org/license/0bsd) or [MIT-0](https://opensource.org/license/mit-0). If you only care about retaining the copyright notice in copies of the source code, the [BSL-1.0](https://opensource.org/license/BSL-1.0) license does not require it for binary redistribution. 

The FSF [comments](https://www.gnu.org/licenses/license-list.html#Zero-BSD) on why using a license without a notice requirement may not be a great idea in practice, specifically:

>The lack of requirement to preserve copyright notices and the license notice does not necessarily mean it is safe to remove them. In some jurisdictions, removing copyright-related information may constitute copyright infringement. Additionally, if you remove copyright notices, users who receive software from you are not able to identify who is giving them permission to use the software.

The most permissive, legally sound, and widely recognized license appears to be MIT. Unlike BSD-3, it lacks the non-endorsement clause. Interestingly, Monero was [originally licensed](https://github.com/monero-project/monero/blob/003b1f909d416ab2c869b76176380c5f133bdad4/src/wallet/wallet2.cpp#L1-L3) under MIT.

MIT is the [most popular](https://innovationgraph.github.com/global-metrics/licenses) license on GitHub. It is followed by Apache-2.0, which includes an explicit patent grant from contributors. Many permissively licensed projects use 'MIT or Apache-2.0' for this reason.

Here is a table with data [from GitHub](https://github.com/github/innovationgraph/blob/16f8a695ec1a1f04ca7b3fbe8f076898616d2e84/data/licenses.csv#L21500) for Q1 2026 for the US. For sake of brevity of I have excluded licenses that have not been mentioned.

|Position | Count | License |
|-:|------:|---|
|1| 324177 | MIT |
|3| 107478 | Apache-2.0 |
|4| 60215 | GPL-3.0 |
|6| 16512 | BSD-3-Clause |
|8| 11176 | CC0-1.0 |
|10| 5325 | MPL-2.0 |
|14| 2934 | LGPL-3.0 |
|17| 1638 | MIT-0 |
|19| 908 | 0BSD |

Projects who simply _don't care_ about copyright usually land on a permissive license like MIT, Apache-2.0, or BSD-3, instead of opting for a less popular, less legally sound public domain declaration or no attribution license. 

It shouldn't be surprising that (as far as I am aware) no project or company has ever been litigated for violating the terms of the MIT or a similarly permissive license. The closest case I could find is [Jacobsen v. Katzer (2008)](https://en.wikipedia.org/wiki/Jacobsen_v._Katzer) concerning the notably less permissive [Artistic License 1.0](https://opensource.org/license/Artistic-1.0). Many cases of GPL/LGPL enforcement exist.

tevador:
> [BSL-1.0](https://opensource.org/license/BSL-1.0) license does not require it for binary redistribution

Good point. However, the current BSD-3 license Monero is being shipped with does require it, so it should probably be fixed.

> It shouldn't be surprising that (as far as I am aware) no project or company has ever been litigated for violating the terms of the MIT or a similarly permissive license.

AFAIK, in the current Doe v. Github lawsuit, the plaintiffs allege that permissively licensed code (including MIT licensed) has been used without the required copyright notices, which I would say counts as litigation for violating the terms of a permissive license.

tobtoht:
>so it should probably be fixed.

Yes, I'm not saying we shouldn't comply with existing license requirements. I'm also not suggesting we should change Monero's license. 

>which I would say counts as litigation for violating the terms of a permissive license.

Granted, but I don't think we should draw conclusions from this case until a legal decision is made.

jeffro256:
Rebased and bumped `mx25519` submodule to latest upstream

external/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@
 
 add_subdirectory(db_drivers)
 add_subdirectory(easylogging++)
+add_subdirectory(mx25519)
```
thomasbuilds: Without EXCLUDE_FROM_ALL the default target builds libmx25519.so, mx25519-tests and mx25519-bench, and the x86_64 Linux job links all three while nothing in the tree uses them. randomx, polyseed and utf8proc are added with the flag. cncrypto's link to mx25519_static still builds the static library.

```suggestion
add_subdirectory(mx25519 EXCLUDE_FROM_ALL)
```

jeffro256: Good point

PR #10965, opened 2026-07-24
crypto: add x25519 wrapper API

jeffro256:
Depends #10964

### DIFF
--- a/src/crypto/x25519.cpp
+++ b/src/crypto/x25519.cpp
@@ -1,0 +1,83 @@
+// Copyright (c) 2022-2026, The Monero Project
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+// 
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+// 
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+// 
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, 
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, 
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, 
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+//paired header
+#include "x25519.h"
+
+//local headers
+#include "crypto.h"
+extern "C"
+{
+#include "crypto-ops.h"
+}
+
+//third party headers
+
+//standard headers
+
+#undef MONERO_DEFAULT_LOG_CATEGORY
+#define MONERO_DEFAULT_LOG_CATEGORY "crypto"
+
+namespace crypto
+{
+//-------------------------------------------------------------------------------------------------------------------
+x25519_secret_key x25519_secret_key_gen(bool scalar)
+{
+    x25519_secret_key privkey;
+    do
+    {
+        rand(sizeof(privkey.data), privkey.data);
+        if (scalar)
+            sc_reduce32(privkey.data);
+    } while (privkey == x25519_secret_key{});
+
+    return privkey;
+}
+//-------------------------------------------------------------------------------------------------------------------
+x25519_pubkey x25519_pubkey_gen()
+{
+    const x25519_secret_key privkey{x25519_secret_key_gen()};
+    x25519_pubkey pubkey;
+    x25519_scmul_base_unclamped(privkey, pubkey);
+
+    return pubkey;
+}
+//-------------------------------------------------------------------------------------------------------------------
+void x25519_scmul_base_unclamped(const x25519_scalar &scalar, x25519_pubkey &result_out)
+{
+    static const mx25519_impl *impl{mx25519_select_impl(mx25519_type::MX25519_TYPE_AUTO)};
+    mx25519_scmul_base_unclamped(impl, &result_out, &scalar, MX25519_UNCLAMP_ALL);
+}
+//-------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


src/crypto/x25519.h
```diff
@@ -0,0 +1,105 @@
...
+
+extern "C"
+{
+#include "mx25519.h"
+}
+
+struct x25519_pubkey : public mx25519_pubkey
+{
+    x25519_pubkey() = default;
+    x25519_pubkey(const mx25519_pubkey &other): mx25519_pubkey(other) {}
```
vtnerd, line 56: Doesn't `= default` call into base classes?

src/crypto/x25519.cpp
```diff
@@ -0,0 +1,91 @@
...
+namespace crypto
+{
+//-------------------------------------------------------------------------------------------------------------------
+x25519_secret_key x25519_secret_key_gen(bool clamp)
+{
+    x25519_secret_key privkey;
+    do
+    {
+        rand(sizeof(privkey.data), privkey.data);
+        if (clamp)
```
vtnerd: So instead of clamping at the `mul`, the scalar gets clamped here. Does default Monero usage not clamp? I recall that being the case (from my arm64 mods).

vtnerd (commented):
Looks fine, will likely approve after response.

src/crypto/x25519.h
jeffro256: Yes, AFAIK, the base class constructor is always called. If not user-supplied, it uses the default constructor. See https://en.cppreference.com/cpp/language/default_constructor#Implicitly-defined_default_constructor:

> An(until C++26)Otherwise, an(since C++26) implicitly-defined default constructor has the same effect as a user-defined constructor with empty body and empty initializer list. That is, it calls the default constructors of the bases and of the non-static members of this class. Class types with an empty user-provided constructor may get treated differently than those with an implicitly-defined default constructor during [value initialization](https://en.cppreference.com/cpp/language/value_initialization).

src/crypto/x25519.cpp
jeffro256: I should re-word this parameter and remove the bit clamping here because, you're right this is confused.

src/crypto/x25519.h
vtnerd: I was obviously asking because this could be `= default`; just a small nitpick.

jeffro256: Oh I see. It can't be ` = default` since the parameter is a `mx25519_pubkey`, not a `x25519_pubkey`.

thomasbuilds:
Should be bumped to current tevador master right?

jeffro256:
Rebased on #10964, which has latest tevador `master`, then re-worked for API change.

src/crypto/x25519.h
```diff
@@ -0,0 +1,105 @@
...
+//forward declarations
+
+
+namespace crypto
+{
+
+extern "C"
+{
+#include "mx25519.h"
+}
```
thomasbuilds, line 51: This includes mx25519.h inside namespace crypto, and with the header's include guard the outcome depends on include order: a TU that sees mx25519.h first gets no crypto::mx25519_* names, so the test's crypto::mx25519_impl would not compile there, and x25519_pubkey ends up deriving from ::mx25519_pubkey in that TU but from crypto::mx25519_pubkey elsewhere. Including it at global scope, as with crypto-ops.h, avoids that; the header has its own extern "C" guards. The test could then use the unqualified names, or a few using-declarations here would keep the crypto:: spelling.

src/crypto/x25519.h
```diff
@@ -0,0 +1,105 @@
...
+{
+    x25519_secret_key() = default;
+    x25519_secret_key(const x25519_scalar &other) { *this = other; }
+    x25519_secret_key& operator=(const x25519_scalar &other)
+    { unwrap(unwrap(*this)) = other; return *this; }
+};
+
+/**
+ * @brief Generate a random X25519 privkey
+ * @param scalar true to sample element of Field25519, false to sample uniform 32 byte secret
```
thomasbuilds, line 77: sc_reduce32 reduces mod l rather than into the field, so maybe:

```suggestion
 * @param scalar true to sample a scalar mod l, false to sample a uniform secret (mx25519 ignores bit 255)
```

src/crypto/x25519.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+//-------------------------------------------------------------------------------------------------------------------
+x25519_secret_key x25519_secret_key_gen(bool scalar)
+{
+    x25519_secret_key privkey;
+    do
+    {
+        rand(sizeof(privkey.data), privkey.data);
+        if (scalar)
+            sc_reduce32(privkey.data);
```
thomasbuilds, line 56: mx25519 ignores bit 255 of the key (mx25519.h:97-98), so about half of the keys from this path carry a bit the multiplication discards, and each of them shares its pubkey with the key that differs only in that bit. Clearing it here would give generated keys a single representation:

```suggestion
        if (scalar)
            sc_reduce32(privkey.data);
        else
            privkey.data[31] &= 0x7f;
```

src/crypto/x25519.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+    static const mx25519_impl *impl{mx25519_select_impl(mx25519_type::MX25519_TYPE_AUTO)};
+    mx25519_scmul_base_unclamped(impl, &result_out, &scalar);
+}
+//-------------------------------------------------------------------------------------------------------------------
+void x25519_scmul_key_unclamped(const x25519_scalar &scalar, const x25519_pubkey &pubkey, x25519_pubkey &result_out)
+{
+    static const mx25519_impl *impl{mx25519_select_impl(mx25519_type::MX25519_TYPE_AUTO)};
+    mx25519_scmul_key_unclamped(impl, &result_out, &scalar, &pubkey);
+}
```
thomasbuilds, line 81: Each of these keeps its own static, so two threads making the first call to each of them run mx25519_select_impl concurrently, and ThreadSanitizer flags the read of `cap` at cpu.c:25 against its writes in mx25519_get_cpu_cap. A single shared static, say a file-local get_impl() used by both, would serialise the probe.

tests/unit_tests/x25519.cpp
```diff
@@ -0,0 +1,253 @@
...
+  {
+    for (unsigned char j = 0; j < 8; ++j)
+    {
+      // add 2^i + j (sometimes with duplicates, which is okay)
+      crypto::x25519_secret_key &s = scalars.emplace_back();
+      memset(s.data, 0, sizeof(crypto::x25519_secret_key));
+      const int msb_byte_index = i >> 3;
+      const int msb_bit_index = i & 7;
+      s.data[msb_byte_index] = 1 << msb_bit_index;
+      s.data[0] = j;
```
thomasbuilds, line 153: For i < 8 the byte index is 0, so this assignment overwrites the bit set just above and the eight scalars per i collapse to 0..7; the single-bit scalars 8 to 128 are never multiplied. `|=` would keep them:

```suggestion
      s.data[0] |= j;
```

tests/unit_tests/x25519.cpp
```diff
@@ -0,0 +1,253 @@
...
+    }
+  }
+  // add -1
+  scalars.push_back(hex2pod<crypto::x25519_scalar>("ecffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f"));
```
thomasbuilds, line 157: This is p-1 rather than -1 mod l; l-1 would be the scalar the comment describes, and p-1 is still a good case to keep:

```suggestion
  // add l-1
  scalars.push_back(hex2pod<crypto::x25519_scalar>("ecd3f55c1a631258d69cf7a2def9de1400000000000000000000000000000010"));
  // add p-1
  scalars.push_back(hex2pod<crypto::x25519_scalar>("ecffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f"));
```

src/crypto/x25519.cpp
jeffro256: NACK. RFC uses of X25519 say that using a uniform 32 byte key is the recommended way to use the key. And clearing top bit does not make the key representation unique, since another key can have the same value mod l.

src/crypto/x25519.h
jeffro256: You're right, I mixed that up

src/crypto/x25519.h
jeffro256: Resolved by removing the include from that header completely

Issue #9352, opened 2024-06-04
ERROR: chunk size exceeds buffer size while exporting monero database


PunchEnergyFTW:
Hi!

I wanted to create a backup of my monero database and found the tool monero-blockchain-export.exe.
Yesterday, i let the tool run for a few minutes and it all worked very well.

But when i want to restart the backup, i get a "chunk size exceeds buffer size" error.
I have synced the database in between these two export attempts, could this be the problem? If yes, isn't this tool supposed to make incremental backups easier?

My blockchain.raw export file is currently ~9GB and at about block height ~1.300.000
```
C:\Users\xxx\Documents\monero-gui-win-x64-v0.18.3.3\monero-gui-v0.18.3.3\extras>monero-blockchain-export.exe --d
ata-dir U:\MoneroBlockChain --output-file U:\Monero-Blockchain-Export\blockchain.raw
2024-06-03 23:52:08.648 W Starting...
2024-06-03 23:52:08.649 W Export output file: U:\Monero-Blockchain-Export\blockchain.raw
2024-06-03 23:52:08.649 W Initializing source blockchain (BlockchainDB)
2024-06-03 23:52:08.650 W database: LMDB
2024-06-03 23:52:08.651 W Loading blockchain from folder U:\MoneroBlockChain\lmdb ...
2024-06-03 23:52:08.748 W Source blockchain storage initialized OK
2024-06-03 23:52:08.749 W Exporting blockchain raw data...
2024-06-03 23:52:08.750 I Storing blocks raw data...
2024-06-03 23:52:08.750 I source blockchain height: 3106278
2024-06-03 23:52:08.750 I Using block height of source blockchain: 3106278
2024-06-03 23:52:08.758 I bootstrap file recognized
2024-06-03 23:52:08.758 I bootstrap::file_info size: 4
2024-06-03 23:52:08.759 I bootstrap file v1.0
2024-06-03 23:52:08.759 I bootstrap magic size: 4
2024-06-03 23:52:08.759 I bootstrap header size: 1024
2024-06-03 23:52:08.760 I bootstrap::blocks_info size: 6
2024-06-03 23:52:08.760 I bootstrap first block:0
2024-06-03 23:52:08.760 I bootstrap last block:2888442
2024-06-03 23:52:08.761 I Scanning blockchain from bootstrap file...
2024-06-03 23:53:16.722 W WARNING: chunk_size 596646401 > BUFFER_SIZE 2097152  height: 8, offset 453105
[...]

PR #10969, opened 2026-07-24
blockchain_utilities: resume interrupted exports safely

Ap4sh:
an interrupted `monero-blockchain-export` write can leave the final chunk header or payload incomplete

the scanner previously used `seekg()` to skip each payload. seeking beyond EOF succeeds, so the incomplete chunk was counted as a complete block and the next export appended after the partial data, leaving later chunk boundaries misaligned

this checks chunk boundaries against the file size while scanning. when the only problem is an incomplete trailing chunk, the writer truncates back to the last complete chunk and resumes from that height. malformed chunk sizes and ordinary read errors still fail without modifying the file

this prevents the interrupted-tail corruption path described in #9352 and repairs exports that still end at the incomplete chunk. it does not try to recover a file that was already appended to and misaligned by an older resume

tests:
- importer and exporter targets with GCC 14 Debug/Release and Clang 19 Release

### DIFF
--- a/src/blockchain_utilities/bootstrap_file.cpp
+++ b/src/blockchain_utilities/bootstrap_file.cpp
@@ -41,6 +41,21 @@ using namespace epee;
 
 namespace
 {
+  class incomplete_chunk final : public std::runtime_error
+  {
+  public:
+    incomplete_chunk(uint64_t offset, uint64_t blocks)
+      : std::runtime_error{"Incomplete trailing chunk"}, m_offset{offset}, m_blocks{blocks}
+    {}
+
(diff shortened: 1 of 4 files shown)


selsta:
Can the test changes be undone? Adding `src/blockchain_utilities/bootstrap_file.cpp` as source is a bit awkward and I think it's fine to merge without them. Code itself looks good.

Ap4sh:
done, removed the test file and its CMake changes while keeping the production fix unchanged

selsta (approved):

jpk68 (approved):

PR #10980, opened 2026-07-26
p2p: stop buffered dispatch after fatal notifications

selsta:


jpk68 (approved):

SChernykh:
There is a difference between "connection close because the peer misbehaved and got banned" and "connection close for whatever other reason". In the second case, it makes sense to process the rest of the input buffer - there might be some transactions or an incoming block there.

SChernykh (approved):

PR #10983, opened 2026-07-26
wallet: display private view keys for hardware wallets

jpk68:
Enables displaying the private view keys of hardware wallets, namely for the `viewkey` command in the CLI wallet. Previously, this command would just display `On device. Not available` for *both* the private view and spend keys, the former of which is actually accessible.

This patch is tested and working for Trezor devices; Ledger devices have not been tested as I do not have access to one. The added function in `device.hpp` is called `get_cached_view_key` rather than just `get_view_key` due to a compiler warning.

cc @SNeedlewoods (sorry for the second ping of the day)

### DIFF
--- a/src/device/device_ledger.cpp
+++ b/src/device/device_ledger.cpp
@@ -657,6 +657,15 @@ namespace hw {
         return true;
     }
 
+    bool device_ledger::get_cached_view_key(crypto::secret_key &viewkey_out) {
+        AUTO_LOCK_CMD();
+
+        if (!this->soft_request_view_key()) return false;
+        viewkey_out = this->viewkey;
+
+        return true;
+    }
+
     bool  device_ledger::generate_chacha_key(const cryptonote::account_keys &keys, crypto::chacha_key &key, uint64_t kdf_rounds) {
         AUTO_LOCK_CMD();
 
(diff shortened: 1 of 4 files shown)


SChernykh:
> This patch is tested and working for Trezor devices; Ledger devices have not been tested as I do not have access to one. 

Yet the patch only adds the code for Ledger but not Trezor?

jpk68:
> Yet the patch only adds the code for Ledger but not Trezor?

The `device_trezor` code already has this functionality (see [here](https://github.com/monero-project/monero/blob/master/src/device_trezor/device_trezor.cpp#L190)); on the other hand, it seems that the existing [Ledger code](https://github.com/monero-project/monero/blob/master/src/device/device_ledger.cpp#L635-L636) intentionally zeroes out the private view key, which is why the new function was needed.

src/device/device_ledger.cpp
```diff
@@ -657,6 +657,15 @@ namespace hw {
         return true;
     }
 
+    bool device_ledger::get_cached_view_key(crypto::secret_key &viewkey) {
+        AUTO_LOCK_CMD();
+
+        if (!this->soft_request_view_key()) return false;
+        viewkey = this->viewkey;
```
SChernykh: This confused me for a few seconds. It's better to rename the function parameter to not have the naming conflict.

SChernykh (approved):

jpk68:
@SChernykh Thanks. It would be good to have someone test with a Ledger device before merging, though

jpk68:
Update: I have tested this with a Ledger device; everything works as it should.

selsta (approved):

PR #10976, opened 2026-07-26
simplewallet: fix restore-date parameter when using generate-from-device

jpk68:
Makes it so that the `--restore-date` option may be used alongside `--generate-from-device`.

Closes #10974

Issue #10974, opened 2026-07-25
Monero wallet cli accepts --restore-height but it doesn't accept --restore-date


CallumOliver:
Monero wallet cli accepts --restore-height but it doesn't accept --restore-date

```
sudo ./monero-wallet-cli --restore-date 2025-08-22 --hw-device Trezor --generate-from-device wallet
This is the command line monero wallet. It needs to connect to a monero
daemon to work correctly.
WARNING: Do not reuse your Monero keys on another fork, UNLESS this fork has key reuse mitigations built in. Doing so will harm your privacy.

Monero 'Fluorine Fermi' (v0.18.4.3-release)
Logging to ./monero-wallet-cli.log
...
No restore height is specified. Assumed you are creating a new account, restore will be done from current estimated blockchain height. Use --restore-height or --restore-date if you want to restore an already setup account from a specific height.
Is this okay?  (Y/Yes/N/No):
```

This wouldn't be such a problem if I could figure out how to convert a date into a height but I could not so now I'm just starting from 1 :(

jpk68:
The `--restore-date` parameter is seemingly [ignored](https://github.com/monero-project/monero/blob/master/src/simplewallet/simplewallet.cpp#L4325) when used in conjunction with `--generate-from-device`. It would be possible to refactor this code to add the same date-conversion logic used elsewhere to this code path, or to just deduplicate all of this logic into one function.

cc @SNeedlewoods

PR #10984, opened 2026-07-27
simplewallet: prompt for restore height when using generate-from-device

jpk68:
When restoring from a hardware device in the CLI wallet (without specifying `--restore-height`), the user will currently be prompted with `Is this okay?`, before wallet creation is aborted upon the user responding with "no".

This seems a bit strange/unintuitive; this patch changes the behaviour so that the user will instead be prompted to specify a restore height after answering "no".

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -4297,12 +4297,18 @@ bool simple_wallet::init(const boost::program_options::variables_map& vm)
           wrt << tr("Use --restore-height or --restore-date if you want to restore an already setup account from a specific height.");
         }
         std::string confirm = input_line(tr("Is this okay?"), true);
-        if (std::cin.eof() || !command_line::is_yes(confirm))
-          CHECK_AND_ASSERT_MES(false, false, tr("account creation aborted"));
+        if (std::cin.eof()) CHECK_AND_ASSERT_MES(false, false, tr("account creation aborted"));
 
-        m_wallet->set_refresh_from_block_height(m_wallet->estimate_blockchain_height() > 0 ? m_wallet->estimate_blockchain_height() - 1 : 0);
-        m_wallet->explicit_refresh_from_block_height(true);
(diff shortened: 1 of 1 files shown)


SChernykh (approved):

monerostar (commented):
## monerostar code-path review (sibling to #10976 / #10974)

**Environment**
- Windows 11 native
- Compared `upstream/master` device branch vs this PR (`27367b8a382b`)
- Related: [#10974](https://github.com/monero-project/monero/issues/10974), [#10976](https://github.com/monero-project/monero/pull/10976)

### What was wrong (UX)

On `--generate-from-device` with no restore height, answering **No** to “Is this okay?” aborted account creation. Users who wanted a custom height/date had to re-run with flags instead of being prompted.

### What this PR does

On **Yes**: same as master (estimate tip − 1, mark explicit height).  
On **No**: `explicit_refresh_from_block_height(false)` instead of abort.

That correctly falls through to the existing shared block:

```text
if (!m_wallet->explicit_refresh_from_block_height() && m_restoring) {
  // interactive height or YYYY-MM-DD prompt
}
```

…which still runs for device wallets (`m_restoring` is set for generate-from-device). So “No” becomes “ask me for a height/date,” matching non-device restore UX. Good.

### Interaction with #10976

This PR still opens the “No restore height…” block with only:

```text
command_line::is_arg_defaulted(vm, arg_restore_height)
```

So **`--restore-date` alone remains broken** until #10976 (or equivalent) lands. Prefer merge order:

1. **#10976** — functional fix for date flag  
2. **this PR** — UX for “No” (rebase if needed so date is excluded from the prompt branch)

If both merge without coordination, tip should include both condition changes.

### Verdict

**LGTM as UX** once stacked with #10976. Small, intentional, uses the existing interactive restore path rather than inventing a third one.

**Formal Approve:** external collaborator — comment review only.

selsta (approved):

PR #10988, opened 2026-07-27
device_trezor: validate bridge response size before parsing

selsta:


src/device_trezor/trezor/transport.cpp
```diff
@@ -481,14 +481,14 @@ namespace trezor{
...
-    if (!bin_data){
+    if (!bin_data || bin_data->size() < PROTO_HEADER_SIZE){
       throw exc::CommunicationException("Response is not well hexcoded");
     }
 
     uint16_t msg_tag;
     uint32_t msg_len;
     deserialize_message_header(bin_data->data(), msg_tag, msg_len);
-    if (bin_data->size() != msg_len + PROTO_HEADER_SIZE){
+    if (bin_data->size() - PROTO_HEADER_SIZE != msg_len){
```
jpk68, line 491: I don't really understand this change; the way it was before avoids subtracting from an unsigned value, and is generally more straightforward.

Something like `std::numeric_limits<size_t>::max()` could also be used to detect an underflow if needed.

selsta: This can't underflow because `bin_data->size() >= PROTO_HEADER_SIZE` is being checked before. The change is because `msg_len` is `uint32_t` and `msg_len + PROTO_HEADER_SIZE` can wrap. It's not strictly needed but more secure this way.

jpk68 (approved):

tobtoht (approved):

SNeedlewoods (approved):

Issue #8759, opened 2023-03-03
Monerod is stuck while synchronizing


meyou69:
I'm currently running the monerod executable (version 0.18.1.2-release) on an x86-64 Arch linux server in headless mode.

The node was fully synchronized with the network and everything worked just fine until yesterday when suddenly, it got behind the blockchain with 1 or 2 blocks. It started synchronizing with the peers to retrieve those blocks but it never continued to function afterwards. When trying to connect my wallets to it, the wallets attempt to connect to it but the server doesn't seem to respond. It's stuck in a synchronization step and it doesn't seem to finish it.

The program is run as a systemd service with the following ExecStart entry:

ExecStart=/usr/local/bin/monerod --log-level=2 --config-file=/etc/monero/monerod.conf --pidfile /var/run/monero/monero.pid --zmq-pub tcp://127.0.0.1:18083 --out-peers 64 --in-peers 32 --add-priority-node=node.supportxmr.com:18080 --add-priority-node=nodes.hashvault.pro:18080 --disable-dns-checkpoints --enable-dns-blocklist --non-interactive

The reason there's the zmq protocol activated is because I'm also running p2pool over it, in the background.

Things I tried to get it unstuck (none of them worked):

1. Restarted the systemd service
2. Reboot the machine
3. Removed the peers cache file p2pstate.bin and restarted the systemd service again

The contents of monerod.conf are the following:

```
# /etc/monero/monerod.conf

# Data directory (blockchain db and indices)
data-dir=/big-storage/monero/.bitmonero  # Remember to create the monero user first

# Log file
log-file=/var/log/monero/monerod.log

# P2P configuration
p2p-bind-ip=0.0.0.0            # Bind to all interfaces (the default)
p2p-bind-port=18080            # Bind to default port
p2p-bind-ipv6-address=::
p2p-bind-port-ipv6=18080
p2p-use-ipv6=1

# RPC configuration
public-node=1                             # Advertise the RPC-restricted port over p2p peer lists
rpc-restricted-bind-ip=0.0.0.0            # Bind restricted RPC to all interfaces
rpc-restricted-bind-port=18089            # Bind restricted RPC on custom port to differentiate from default unrestricted RPC (18081)
confirm-external-bind=1                   # Open restricted RPC node (confirm)
no-igd=1                                  # Disable UPnP port mapping

# CUSTOM: enable IPv6
rpc-restricted-bind-ipv6-address=::1
rpc-use-ipv6=1

# ZMQ configuration
[...]

Issue #8194, opened 2022-02-27
Syncing gets stuck around block >~2450000


tryphe:
**Describe the problem**

Behavior: At some point after starting monerod, ranging from a few minutes to many hours, syncing "gets stuck". Whether synced or unsynced, blocks stop being added to the chain until monerod is restarted. The monerod process continues running and is responsive, however, and exits normally if given HUP etc.

The behavior repeatedly appears for me, in particular, starting around block >~2450000. Not any specific block, just random, but not at any point before ~2450000.  If monero is already synced, it takes longer for the behavior to appear, sometimes several hours longer.

There are plenty of connections and new connections being made during the behavior.
Monerod gives me two messages like this every few minutes, gives the correct number of total blocks, but is still stuck at the same block where the behavior appeared:
```
[random.ipv4.address:18080 OUT] Sync data returned a new top block candidate: 2459334 -> 2568414 [Your node is 109080 blocks (5.0 months) behind] 
SYNCHRONIZATION started 
```

Workaround: If I create a cron script to restart monerod every ~15 minutes, it eventually syncs.

**A few things about my node and what I've tried:**

I've been running monero for over 5 years with no syncing issues, until about a month ago.
Minimal setup Debian 10 VM with ntpd and monerod (Update: tried with a new Debian 11 machine and the same behavior occurs).
Syncing works fine until about 95%, or near block 2450000.
Removed the entire .bitmonero directory and synced from scratch several times with the same result.
Removed my config and am running with no arguments to reduce complexity, but get the same result.
Running on clearnet.
Using the blocklist at https://gui.xmr.pm/files/block.txt, but I tried without the blocklist and get the same result.
Using outbound connections only to try and sync currently to avoid possibility of attack.
Tried running with things like `--out-peers 25` or `--out-peers 50` but get the same result.
Running the latest master branch 9aab19f349433687c7aaf2c1cbc5751e5912c0aa but also tried with signed releases:
```
0.17.3.0
0.17.2.0
0.17.1.9
0.17.1.7
0.17.1.5
0.17.1.3
```
which experience the same behavior but do NOT show errors in the log.

**Update: I originally thought the syncing problem was related to these errors, but this is not the case. Will leave this section here**

Some errors/warnings that frequently occur:
```
ERROR   net     contrib/epee/include/net/abstract_tcp_server2.inl:777   Setting timer on a shut down object

[...]

Issue #3713, opened 2018-04-26
monerod frequently gets stuck syncing


B6Zip:
Every couple of days it seems that monero daemon will stop retrieving new blocks and will be sitting at the same block height until the daemon is restarted.

moneromooo-monero:
Once this happens, post the output of "sync-info" and "status".
Then set_log 1,net*:DEBUG
Then wait a bit and post the resulting log (fpaste.org or pastebin.mozilla.org or paste.debian.net).

Onefox:
I have the same Problem.
At some point i don't see the yellow block messages anymore after a deamon restart.
![image](https://user-images.githubusercontent.com/1709259/39335114-f4f4f432-49b1-11e8-849f-960e7100e6fa.png)
Here is the set_log output after that:
https://paste.fedoraproject.org/paste/Szh-FyNuY1zVWwNa6v6X9Q

moneromooo-monero:
Once this happens, post the output of "sync_info" and "status".
And "diff" too.

Onefox:
@moneromooo-monero  what do you mean with diff?
The sync and status info can be seen on the screenshot right?

Onefox:
The other node had to be restaret.
But i have the same problem on a testnode:
here the print_cn:
> 
> print_cn
> Remote Host                   Peer id             Support Flags       Recv/Sent (inactive,sec)      State                    Livetime(sec)       Down (kB/s) Down(now)     Up (kB/s) Up(now)
> 
> INC 62.210.104.109:24790      fe74bb70413ebbe0    1                   675176(5)/10919174(5)         state_normal             270                 2           0             39        69
> INC 31.174.56.239:56178       14f02ca5ba6336be    1                   8449(43)/8145580(41)          state_normal             274                 0           0             29        290
> INC 54.38.82.8:54474          00a3043c706759d5    1                   147620(41)/56076(41)          state_normal             416                 0           2             0         0
> INC 183.89.212.241:59479      5ad01cd24433ff32    1                   365521(40)/387040(41)         state_normal             2994                0           0             0         0
> OUT 81.169.162.150:28080      b190ef88ec604c13    1                   2231639(11)/1486779(11)       state_normal             11808               0           0             0         0
> OUT 78.2.110.212:28080        be2e871ebd2d7790    1                   4534476(5)/1496132(5)         state_normal             11813               0           0             0         0
[...]

PR #10990, opened 2026-07-28
Fix long-standing 'Next span is not empty' sync stall (#8194)

JossieEtheridge:
## Summary

Fixes the sync stall described in #8194 (and the same symptom reported in #2536, #3713, #2795, #8759) -- a node syncing far behind the chain tip connects to healthy peers, repeatedly logs "SYNCHRONIZATION started", and commits zero new blocks, sometimes indefinitely. Root-caused to three separate issues in `src/cryptonote_protocol/`, each fixed in its own commit:

1. **TOCTOU race in span reservation** (`block_queue.cpp`) -- `get_next_span_if_scheduled()` and `reset_next_span_time()` were two separate mutex-locked calls; another connection's thread could fill the same span in the gap between them, silently corrupting that span's retry bookkeeping via a failed assertion that just logs and returns. Fixed by merging both steps into one atomic method.

2. **`LAST_ACTIVITY_STALL_THRESHOLD` too aggressive** (`cryptonote_protocol_handler.inl`) -- hardcoded to 2.0 seconds, while real successful span completions on a node syncing ~630k blocks behind the tip measured 2-26 seconds each. This created a self-defeating loop where the head-of-queue span (blocking all further commits, since blocks apply in strict order) was reassigned to a new peer every few seconds indefinitely, because every freshly-assigned peer also tripped the same 2-second check under normal jitter. Widened to 20 seconds.

3. **Queue-cap logic used `||` where `&&` was intended** (`cryptonote_protocol_handler.inl`) -- `queue_proceed` only paused downloading once *both* the span-count cap and byte-size cap were exceeded, meaning the looser of the two effectively governed alone. Observed the queue overshoot to 17 spans / ~433MB before pausing. Fixed to `&&`.

## Testing

Reproduced the stall reliably on a node ~630k blocks behind tip, on both v0.18.5.0 and v0.18.5.1, under both `--db-sync-mode fast` and `--db-sync-mode safe` -- confirmed via `data.mdb` mtime that zero blocks were committed across multiple independent multi-minute windows on each combination.

Applying all three fixes to the same node, same network conditions: sustained 36-56 blocks/sec, syncing cleanly from the same stuck point through to full sync.

## Test plan

- [x] Reproduced the stall on stock v0.18.5.1 (multiple independent test windows, zero progress confirmed via `data.mdb` mtime)
- [x] Applied fix, confirmed sustained sync progress (36-56 blocks/sec) from the same stuck point
- [x] Full sync completed cleanly from ~83% to 100% with all three fixes applied
- [ ] Would appreciate maintainer review on the `LAST_ACTIVITY_STALL_THRESHOLD` value specifically (20s) -- chosen from directly observed transfer times on one node/network path, may benefit from wider testing across different connection speeds

### DIFF
--- a/src/cryptonote_protocol/block_queue.cpp
+++ b/src/cryptonote_protocol/block_queue.cpp
@@ -346,6 +346,24 @@ void block_queue::reset_next_span_time(boost::posix_time::ptime t)
   (boost::posix_time::ptime&)i->time = t; // sod off, time doesn't influence sorting
 }
 
+std::pair<uint64_t, uint64_t> block_queue::get_next_span_if_scheduled_and_reset_time(std::vector<crypto::hash> &hashes, boost::uuids::uuid &connection_id, boost::posix_time::ptime &time, boost::posix_time::ptime new_time)
+{
+  boost::unique_lock<boost::recursive_mutex> lock(mutex);
+  if (blocks.empty())
+    return std::make_pair(0, 0);
+  block_map::iterator i = blocks.begin();
+  if (i == blocks.end())
+    return std::make_pair(0, 0);
+  if (!i->blocks.empty())
+    return std::make_pair(0, 0);
+  hashes = i->hashes;
+  connection_id = i->connection_id;
+  time = i->time;
+  std::pair<uint64_t, uint64_t> ret = std::make_pair(i->start_block_height, i->nblocks);
+  (boost::posix_time::ptime&)i->time = new_time; // same "sod off" cast as reset_next_span_time, still under the same lock
+  return ret;
+}
+
 void block_queue::set_span_hashes(uint64_t start_height, const boost::uuids::uuid &connection_id, std::vector<crypto::hash> hashes)
 {
   boost::unique_lock<boost::recursive_mutex> lock(mutex);
--- a/src/cryptonote_protocol/block_queue.h
+++ b/src/cryptonote_protocol/block_queue.h
@@ -83,6 +83,14 @@ namespace cryptonote
     uint64_t get_next_needed_height(uint64_t blockchain_height) const;
     std::pair<uint64_t, uint64_t> get_next_span_if_scheduled(std::vector<crypto::hash> &hashes, boost::uuids::uuid &connection_id, boost::posix_time::ptime &time) const;
     void reset_next_span_time(boost::posix_time::ptime t = boost::posix_time::microsec_clock::universal_time());
+    // Atomic combination of get_next_span_if_scheduled() + reset_next_span_time()
+    // under a single lock acquisition. The separate two-call sequence (as used
+    // previously in cryptonote_protocol_handler.inl) has a TOCTOU race: another
+    // connection's thread can fill the span's blocks in the gap between the two
+    // locked calls, causing reset_next_span_time()'s "Next span is not empty"
+    // assertion to fire and silently no-op, leaving that span's retry clock
+    // stuck and wedging the whole download queue.
+    std::pair<uint64_t, uint64_t> get_next_span_if_scheduled_and_reset_time(std::vector<crypto::hash> &hashes, boost::uuids::uuid &connection_id, boost::posix_time::ptime &time, boost::posix_time::ptime new_time = boost::posix_time::microsec_clock::universal_time());
     void set_span_hashes(uint64_t start_height, const boost::uuids::uuid &connection_id, std::vector<crypto::hash> hashes);
     bool get_next_span(uint64_t &height, std::vector<cryptonote::block_complete_entry> &bcel, boost::uuids::uuid &connection_id, epee::net_utils::network_address &addr, bool filled = true) const;
     bool has_next_span(const boost::uuids::uuid &connection_id, bool &filled, boost::posix_time::ptime &time) const;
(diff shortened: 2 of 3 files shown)


UkoeHB:
Needs rebase onto master

nahuhh:
> Needs rebase onto master

This is a release based pr, and should already be fixed on master

and i think this is misdiagnosed


> Measured real, successful 100-block span completions on this node taking
2-26 seconds each. With a 2-second threshold, any peer that goes briefly
quiet — completely normal under real network conditions — gets its
in-progress span reassigned to a different peer.


user had set `block-sync-size=100` -- this is the cause of the stall

jeffro256:
optimize_startup

JossieEtheridge:
Thanks both — rebased onto current `master` and dropped/kept only what's actually still needed. Wanted to be precise rather than just reassert the original diagnosis, so here's what changed after checking master directly:

**@UkoeHB** — done, PR is now `MERGEABLE` against current master (2 commits, was 3).

**@nahuhh** — you were partly right, and I want to be straight about which part:

- The queue-cap (span-count vs byte-size) fix I originally submitted is **already effectively present on master** — it's been folded into a broader stripe-aware refactor (`queue_proceed_init = (nspans < m_span_limit.load()) && (size < block_queue_size_threshold)`), which achieves the same AND-instead-of-OR correction my patch made, plus more. I've **dropped that commit entirely** from this PR — no reason to duplicate a fix that already landed through different work.
- The other two are still live bugs on master, unchanged, and I verified this directly (not from assumption) by diffing against `origin/master`:
  - `LAST_ACTIVITY_STALL_THRESHOLD` is still `2.0f` on master today. `REQUEST_NEXT_SCHEDULED_SPAN_THRESHOLD_STANDBY` did get independently widened (5s→15s) at some point, but the actual stall-detection threshold that's the root of the "Next span is not empty" churn wasn't touched.
  - The TOCTOU race between `get_next_span_if_scheduled()` and `reset_next_span_time()` (two separate locked calls with a gap between them) is also still present verbatim on master's call site.

On the `block-sync-size=100` point specifically: that's a fair alternative hypothesis and I'll own that I didn't isolate it — my own fix stack also changed `--block-sync-size` from 100→20 alongside the code patches, so I don't have a clean A/B showing the code fix alone (without the span-size change) resolves it. What I do have is more specific than a config/log correlation: the exact `Next span is not empty` assertion — a named log line tied to one specific code path — stopped firing entirely after the TOCTOU fix, and reassignment churn measured via debug logging dropped ~10x after widening the stall threshold. Smaller spans very plausibly reduce how often these thresholds get breached in the first place (less data per span → faster completion → less likely to cross a stall window), so I'd expect `block-sync-size=100` combined with these two fixes to also work, just with less margin. But the two bugs are real independent of span size — a 2-second silence threshold is going to be too aggressive for *some* users/network paths even at a 20-block span size, it's just a matter of what bandwidth/latency finally triggers it.

Happy to re-test with `block-sync-size` left at the default if that's useful evidence either way — this node is fully synced now so a clean before/after isn't reproducible from scratch anymore, but I can dig up the debug logs from the original stall if that helps.

nahuhh:
> so I'd expect block-sync-size=100 combined with these two fixes to also work,


It will fail.

JossieEtheridge:
Fair point, and I don't have that specific data point yet — everything I ran today was with `block-sync-size=20`, not the default 100, so I can't say for certain the two fixes alone hold at 100.

Worth being precise about *why* I changed that value, though: it wasn't part of the bug fix, it was a separate tuning call. At this chain height a 100-block span measured 18-28MB — so any time a span did get stalled/reassigned (from either of the two threshold bugs), it threw away a lot more progress than a smaller span would. Smaller spans reduce the *cost* of hitting the bug, not necessarily whether it gets hit at all. The two actual fixes (the TOCTOU race and the 2s stall threshold) operate on peer-connection timing, not span size — I don't see a mechanism in the code where span size changes whether they trigger, but I'd rather verify that than assert it.

Planning to re-test at the default `block-sync-size=100` with just these two fixes to close that gap with real data. Will report back.

nahuhh:
> Fair point, and I don't have that specific data point yet — everything I ran today was with `block-sync-size=20`, not the default 100, so I can't say for certain the two fixes alone hold at 100.

The default is not 100. On release-v0.18, it is 100 for pre-v4 and 20 after. On master, it is dynamic. 

> At this chain height a 100-block span measured 18-28MB — so any time a span did get stalled/reassigned (from either of the two threshold bugs)

It throws away the span because it bumps against serialization limits. Nothing to do with your fixes. Reducing to 20 hides the issue, it does not fix it.


> Smaller spans reduce the *cost* of hitting the bug, not necessarily whether it gets hit at all.

Wrong. As above

> The two actual fixes (the TOCTOU race and the 2s stall threshold) operate on peer-connection timing, not span size — I don't see a mechanism in the code where span size changes whether they trigger, but I'd rather verify that than assert it.

Serialization limits.

> Planning to re-test at the default `block-sync-size=100` with just these two fixes to close that gap with real data. Will report back.

You can interactively `pop_blocks 200000` on master, which should be relatively fast. Hitting serialization limits, which cause a drop & rotate peer loop, should only take a few minutes.

selsta:
If an issue is closed, it often means the issue is already fixed, like in this case.

PR #10993, opened 2026-07-28
wallet2_api: skip the TLS handshake on Tor/I2P daemons

### LINKED ISSUE #10992: wallet2_api: `doInit()` discards `use_ssl` (onions still attempt TLS handshake)
the effect is minor, 1-4 seconds wasted on a handshake for tor nodes. detecting the problem via ci tests requires further investigation.

feather fixes it [here](https://github.com/feather-wallet/feather/blob/b6a95940b9cb8913471d1282fd968e021ec17b02/src/utils/nodes.cpp#L229)

every wallet2 api user (monero-gui,monerujo,,stack,cake,monfluo,anonero) attempts to to disable it (monerujo trusts the defaults)

thomasbuilds:
doInit() took a `ssl` argument and never read it, so every connection used autodetect and onion/i2p daemons got a pointless TLS handshake over an already encrypted, authenticated transport.

TLS is now skipped for .onion/.i2p addresses. use_ssl or an https:// address requires it there instead, without certificate verification: this API cannot carry a CA or fingerprint yet (#10819 will), monerod's default certificate is self-signed so system-CA verification would just drop the connection, and a .onion/.i2p name is self-authenticating (as `ssl_options_t::has_strong_verification` already treats it). The policy covers both wallet-api paths (Wallet::init and WalletManager::setDaemonAddress), matches the suffix case insensitively since Tor lowercases the name it is handed, and logs its decision at INFO. Clearnet is untouched: use_ssl defaults to false, so acting on it would cut callers off https daemons.

Breaking edge: a TLS-only onion daemon addressed without https:// is no longer reached, and https:// at a plaintext-only one now fails instead of silently downgrading.

Tests: libwallet_api_tests captures the first bytes the wallet sends to a fake SOCKS5 proxy, for both paths, and checks for a TLS ClientHello (absent by default, present with use_ssl or https://, clearnet autodetect as the positive control); unit_tests cover the case-insensitive suffix check.

Not changed here: epee's `has_strong_verification` still matches the suffix case-sensitively, so `--daemon-address X.ONION --proxy` at CLI startup keeps rejecting an uppercase onion while the wallet API and simplewallet's `set_daemon` now treat it as a Tor address; a one-line follow-up in net_ssl.cpp if wanted.

Closes #10992

### DIFF
--- a/src/wallet/api/utils.cpp
+++ b/src/wallet/api/utils.cpp
@@ -32,6 +32,13 @@
 
 #include "include_base_utils.h"                     // LOG_PRINT_x
 #include "common/util.h"
+#include "net/net_parse_helpers.h"
+#include "ssl_options.h"
+
+#include <boost/algorithm/string/predicate.hpp>
+
+#undef MONERO_DEFAULT_LOG_CATEGORY
+#define MONERO_DEFAULT_LOG_CATEGORY "WalletAPI"
 
 using namespace std;
 
@@ -46,6 +53,31 @@ bool isAddressLocal(const std::string &address)
         MERROR("error: " << e.what());
         return false;
     }
+}
+
+epee::net_utils::ssl_options_t sslOptionsForDaemon(const std::string &daemon_address, bool use_ssl)
+{
+    epee::net_utils::http::url_content parsed{};
+    if (!epee::net_utils::parse_url(daemon_address, parsed))
+        return epee::net_utils::ssl_support_t::e_ssl_support_autodetect;
+
+    // clearnet keeps autodetect: use_ssl defaults to false, honouring it would drop TLS to https daemons
+    if (!tools::is_privacy_preserving_network(parsed.host))
+        return epee::net_utils::ssl_support_t::e_ssl_support_autodetect;
+
+    const bool wants_ssl = use_ssl || boost::iequals(parsed.schema, "https", std::locale::classic());
+    if (!wants_ssl)
+    {
+        MINFO("TLS handshake skipped for Tor/I2P daemon " << parsed.host);
+        return epee::net_utils::ssl_support_t::e_ssl_support_disabled;
+    }
+
+    // no peer verification: this API cannot carry a CA or fingerprint yet, monerod's default
+    // certificate is self-signed, and a .onion/.i2p name is self-authenticating (has_strong_verification)
+    MINFO("TLS required for Tor/I2P daemon " << parsed.host << ", certificate verification off");
+    epee::net_utils::ssl_options_t options(epee::net_utils::ssl_support_t::e_ssl_support_enabled);
+    options.verification = epee::net_utils::ssl_verification_t::none;
+    return options;
 }
 
 void onStartup()
--- a/src/wallet/api/wallet_manager.cpp
+++ b/src/wallet/api/wallet_manager.cpp
@@ -32,6 +32,7 @@
 #include "wallet_manager.h"
 #include "wallet.h"
 #include "common_defines.h"
+#include "ssl_options.h"
 #include "common/dns_utils.h"
 #include "common/util.h"
 #include "common/updates.h"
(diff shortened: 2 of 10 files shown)


tests/libwallet_api_tests/main.cpp
```diff
@@ -253,6 +255,117 @@ struct WalletTest2 : public testing::Test
...
+        {
+            boost::system::error_code ignored;
+            acceptor.close(ignored);
+            worker.join();
+        }
+        return first_bytes;
+    }
+
+private:
+    static bool grant_socks4a(boost::asio::ip::tcp::socket &sock)
```
jpk68: Is there any reason the tests are using SOCKS4a over SOCKS5?

jpk68 (approved):
Looks fine to me, but another set of eyes would be good :)

vtnerd (commented):
The user may wish to use SSL authentication over Tor. The situation would be rare, but would happen anytime the Monero daemon is on a different machine than the Tor daemon. The first layer authenticates to Tor, and the second layer authenticates to monerod.

thomasbuilds:
Seems like there's no way to express a cert through the wallet API: `Wallet::init` has a single `bool` for SSL, no fingerprint or CA parameter. So authentication can't be enforced here before or after this PR, `autodetect` only warns on a bad cert. On the cli it can via `--daemon-ssl-allowed-fingerprints`.

Is `disabled` acceptable as the onion default in the interim and should that branch honour cert params once #10819 lands `setDaemon(..., ssl_ca_file_path, ssl_allowed_fingerprints, ...)`?

vtnerd:
The cli wallet doesn't use that api, and should support fingerprints.

There's a separate pr to get the cli wallet to use the API, and this update includes new parameters for ssl.

thomasbuilds:
`Wallet::setDaemon()` from #9464 builds its own `ssl_options_t` and calls `wallet2::set_daemon()` directly so certificates given there are never subject to the onion skip; this PR only sets the SSL policy for `init(use_ssl)` and `setDaemonAddress()`, which take no certificate parameters. Once both are in I'd route `setDaemon()`'s no-certificate autodetect branch through `sslOptionsForDaemon()` too, in whichever lands second.

src/wallet/api/utils.cpp
```diff
@@ -48,6 +55,31 @@ bool isAddressLocal(const std::string &address)
     }
 }
 
+epee::net_utils::ssl_options_t sslOptionsForDaemon(const std::string &daemon_address, bool use_ssl)
+{
+    epee::net_utils::http::url_content parsed{};
```
vtnerd, line 60: I don't understand why `use_ssl` doesn't automatically return `enabled` with verification.

`http` + `!use_ssl` -> disabled
`http` + use_ssl -> error?
`https` + !use_ssl -> enabled without verification
`https` + `use_ssl` -> enabled with verification.

src/wallet/api/wallet_manager.cpp
```diff
@@ -229,7 +230,7 @@ std::string WalletManagerImpl::errorString() const
 
 void WalletManagerImpl::setDaemonAddress(const std::string &address)
 {
-    m_http_client.set_server(address, boost::none);
+    m_http_client.set_server(address, boost::none, Utils::sslOptionsForDaemon(address, false));
```
vtnerd: I don't like this, as it sets the latter parameter to `false`. This could be overriding a previous call to `init` that set it to `true`.

vtnerd (changes requested):

src/wallet/api/utils.cpp
thomasbuilds: monerod serves a self-signed certificate by default and this API cannot pass a fingerprint/CA until #9464 so `enabled` with `system_ca` would drop the connection to practically every onion daemon today; for .onion/.i2p, `use_ssl` is the CLI's `--daemon-ssl enabled --daemon-ssl-allow-any-cert`. Cake also always sends `http://` and passes its SSL toggle as `use_ssl`, so `http` + `use_ssl` as an error would hit its SSL-on onion nodes.

src/wallet/api/wallet_manager.cpp
thomasbuilds: It cannot override `init()`, the manager has its own `m_http_client`, separate from each wallet's.
The hard-coded `false` was only there because `setDaemonAddress()` had no flag of its own - it now takes `bool use_ssl = false` mirroring `init()`.

jpk68 (approved):
LGTM, though it might be good to get a +1 from @SNeedlewoods as well

PR #10995, opened 2026-07-28
wallet: upstream misc stuff before FCMP++

jeffro256:
Some miscellaneous wallet changes that aren't FCMP++-specific, but help support the FCMP++ wallet integration.

- Add `use_fork_rules_offline()`
- Add offline mode for `get_fee_algorithm()`
- Add mode for `setup_new_blockchain()` for already-initiated wallet instances
- Add `get_subaddress_map_ref()`
- Remove spurious `private` label in `gamma_picker`
- Adds docstrings for `create_transactions_*()`
- Refactor `brief` tags for docstrings in basic types header
- If `IN_UNIT_TESTS` is defined, exclude `private` label for `wallet2` fields

Part of upstreaming Carrot/FCMP++

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2024, The Monero Project
+// Copyright (c) 2014-2026, The Monero Project
 // 
 // All rights reserved.
 // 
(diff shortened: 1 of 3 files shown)


src/wallet/wallet2.cpp
```diff
@@ -8561,14 +8554,22 @@ uint64_t wallet2::get_fee_quantization_mask()
...
+  // changes at v3, v5, v8, v17
+  if (use_fork_rules_fn(HF_VERSION_PER_BYTE_FEE, 0))
     return fee_algorithm::HardforkV8;
-  if (use_fork_rules(5, 0))
+  if (use_fork_rules_fn(5, 0))
     return fee_algorithm::HardforkV5;
-  if (use_fork_rules(3, -30 * 14))
+  if (use_fork_rules_fn(3, -30 * 14))
    return fee_algorithm::HardforkV3;
   return fee_algorithm::PreHardforkV3;
```
j-berman, line 8574: If the wallet hasn't synced any blocks (which is expected of an offline wallet), it defaults to returning this old algo. Maybe it makes sense to add logic like `if offline && get_blockchain_current_height() <= 1, then default to the latest` and maybe log an error in that case

Or inside `use_fork_rules_offline`, if we haven't synced any blocks, default to expecting the highest fork version?

jpk68:
A small issue was found here:
https://github.com/xmrack/monero-review/issues/452

PR #10996, opened 2026-07-29
cryptonote_core: stop re-validating the whole tx-pool on a fork change

### LINKED ISSUE #10142: monerod revalidates the whole tx-pool on HFs

earnifysu:
monerod currently walks the entire tx-pool and takes out and re-adds every transaction whenever the fork version changes, and again at startup. that stalls the daemon on the block-add path, where it holds up everything else. with a large pool this is the freeze reported in #10142.

this implements the approach suggested in that issue: record the fork version each transaction last satisfied the non-input consensus rules at, and re-check lazily at the points where a stale one could actually matter. transactions that are never used are never re-verified at all.

### why three commits

split so each is reviewable on its own and the unsafe intermediate state never exists:

1. **`blockchain_db: record the fork version a pool tx was verified at`** adds the storage and stamps it in `add_tx()`. no behaviour change, nothing reads the field yet.
2. **`cryptonote_core: re-check pool txs against the current fork lazily`** adds the lazy checks. only adds validation, removes none.
3. **`cryptonote_core: stop re-validating the whole tx-pool on a fork change`** removes the eager revalidation, which is only safe once 1 and 2 are in.

### two things worth a reviewer's attention

**`tx_memory_pool::validate()` writes back a pre-validation copy of the tx meta.** it captures `e.meta` from `for_all_txpool_txes()` before the take/re-add, then calls `update_txpool_tx(e.txid, e.meta)` afterwards, discarding anything `add_tx()` stamped in between. the obvious implementation of this change is silently defeated by that, which is part of why the function is removed rather than adapted.

**pool transactions are not non-input-checked on the block-add path.** in `handle_block_to_main_chain()`, `ver_non_input_consensus()` runs only on `extra_block_txs`. transactions pulled from the pool via `take_tx()` get `check_tx_inputs()` alone. that is sound today *purely because* the eager `validate()` guarantees every pool tx is non-input-valid at the current fork. removing the eager call without the guard added in commit 2 would have turned this from a performance fix into a consensus change.

### a separate correctness fix carried in commit 2

`pop_block_from_blockchain()` waives the non-input check by passing the post-pop fork version as `nic_verified_hf_version`, justified by a comment saying the transaction was taken "from the mempool earlier in this function call". these transactions come out of the block being popped, read from the db, not out of the mempool. `version` is also computed after the pop, as `get_ideal_hard_fork_version(m_db->height())`. the outcome is correct today only because the `validate()` at the end of the same function repairs the mismatch.

passing `previous_hf_version` makes the claim true: equal to `version` when the pop stays within one fork, and different exactly when it crosses a boundary, which is precisely when re-verification is needed.

### storage

two bytes carved off the front of `txpool_tx_meta_t::padding[44]`, so `sizeof` stays 192 and every other member keeps its offset. no db version bump and no migration. this follows the precedent of 40eb82873, which did the same on this struct and named this issue in its commit message, noting it "already had padding allocated".

those bytes were padding in databases written by older daemons. current code memsets them, but resting on that alone would be a bad trade, since a stale byte that happened to equal the current fork version would let an unverified transaction skip its check. they are paired with a tag byte and read through an accessor that reports the version as unknown unless the tag matches.

### behaviour change worth calling out

a transaction that fails the new rules is no longer dropped the moment the fork happens. it stays in the pool until it expires or is pruned by weight. a bounded background sweep would address that if it is considered worth it.

### status

**this series is not compile tested.** i do not currently have a monero build environment available, so it has been verified by review only. it needs a build, the unit tests, and ideally a testnet fork crossing before it is taken seriously. opening it as a draft for that reason, as a starting point rather than as finished work.

fixes #10142


jeffro256:
At first glance, this seems like a duplicate of #10916

earnifysu:
you are right, this duplicates #10916. closing.

i should have searched for open prs referencing #10142 before writing any of this. i scoped the
search to the fcmp++ milestone, #10916 is not on it, so i never saw it. entirely my error.

for what it is worth, having since read #10916 properly: it is the better implementation. it threads
the per-tx version through `return_txs_to_pool`, which this branch also needed, and its handling of
`pop_block_from_blockchain` is more precise than mine. i had changed that call site to pass
`previous_hf_version` on the theory that the post-pop `version` could not honestly describe txs
coming out of the popped block. that theory was wrong: `m_db->pop_block()` runs before `version` is
computed, so `get_ideal_hard_fork_version(m_db->height())` already returns the popped block's own
fork version, which is exactly what those txs were verified under in `handle_block_to_main_chain`.
my change would have forced a redundant re-verification on every boundary-crossing pop.

sorry for the noise.

PR #10997, opened 2026-07-29
device: fix ledger status code reporting

reservedbytes:
## Problem

Opening a Ledger-backed wallet while the device is connected but locked fails with:

```
Wrong Device Status: 0x5515 (UNKNOWN), EXPECTED 0x9000 (UNKNOWN), MASK 0xffff
```

Two separate problems are visible in that one line.

## 1. `status_codes` is not sorted, so names are misreported

`Status::to_string()` looks the code up with `std::lower_bound`, which requires a
sorted range. A comment above the table says it must be sorted, but nothing enforced
that, and it was not: `SW_OK` (0x9000) was the first entry, and `SW_SECURITY_LOCKED`
(0x69EE) sat before `SW_COMMAND_NOT_ALLOWED` (0x6980).

Three codes therefore resolved to `UNKNOWN`:

| code | reported | correct |
| --- | --- | --- |
| 0x9000 | UNKNOWN | SW_OK |
| 0x6700 | UNKNOWN | SW_WRONG_LENGTH |
| 0x69EE | UNKNOWN | SW_SECURITY_LOCKED |

`SW_OK` is the expected status in every one of these messages, so *every* Ledger
error has been printing `EXPECTED 0x9000 (UNKNOWN)`.

The first commit sorts the table and adds a `static_assert` so the ordering is
enforced rather than merely documented. The table has been unsorted since #6322
introduced it.

## 2. A locked device is reported as a raw status dump

0x5515 is a Ledger OS status word meaning the device is locked. It is not one of the
Monero app's own status words, so it has no table entry and falls through to the raw
dump above.

The second commit reports it the way `SW_CLIENT_NOT_SUPPORTED` and
`SW_PROTOCOL_NOT_SUPPORTED` are already reported a few lines up:

```
Ledger is locked.
```

The same guard is added to `exchange_wait_on_input()`, which is the path used while
signing a transaction — a device that auto-locks mid-signing hits it.

0x5515 is documented as `LOCKED_DEVICE` in Ledger's own error definitions:
https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledgerjs/packages/errors/src/index.ts

Reported downstream as monero-project/monero-gui#3507.

### DIFF
--- a/src/device/device_ledger.cpp
+++ b/src/device/device_ledger.cpp
@@ -81,10 +81,9 @@ namespace hw {
       static const char *to_string(unsigned int code);
     };
 
-    // Must be sorted in ascending order by the code
     #define LEDGER_STATUS(status) {status, #status}
     constexpr Status status_codes[] = {
-      LEDGER_STATUS(SW_OK),
+      LEDGER_STATUS(SW_LOCKED_DEVICE),
       LEDGER_STATUS(SW_WRONG_LENGTH),
       LEDGER_STATUS(SW_SECURITY_PIN_LOCKED),
       LEDGER_STATUS(SW_SECURITY_LOAD_KEY),
@@ -98,7 +97,6 @@ namespace hw {
       LEDGER_STATUS(SW_SECURITY_INTERNAL),
       LEDGER_STATUS(SW_SECURITY_MAX_SIGNATURE_REACHED),
       LEDGER_STATUS(SW_SECURITY_PREFIX_HASH),
-      LEDGER_STATUS(SW_SECURITY_LOCKED),
       LEDGER_STATUS(SW_COMMAND_NOT_ALLOWED),
       LEDGER_STATUS(SW_SUBCOMMAND_NOT_ALLOWED),
       LEDGER_STATUS(SW_DENY),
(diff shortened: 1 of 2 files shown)


src/device/device_ledger.cpp
```diff
@@ -81,10 +81,8 @@ namespace hw {
       static const char *to_string(unsigned int code);
     };
 
-    // Must be sorted in ascending order by the code
     #define LEDGER_STATUS(status) {status, #status}
     constexpr Status status_codes[] = {
```
jeffro256, line 85: Should `SW_LOCKED_DEVICE` be added to this table?

src/device/device_ledger.cpp
```diff
@@ -81,10 +81,8 @@ namespace hw {
       static const char *to_string(unsigned int code);
     };
 
-    // Must be sorted in ascending order by the code
```
jeffro256, line 84: Why remove this comment?

src/device/device_ledger.cpp
reservedbytes: yeah, makes sense

only really matters for logging since the assert catches `0x5515` before `ASSERT_SW` prints it, but i guess no reason to leave the code nameless

src/device/device_ledger.cpp
reservedbytes: i dropped it because the `static_assert` says the same thing and actually enforces it

jeffro256: Fair enough. I think that it would be helpful for people later adding codes to the table, but I guess they'll figure out immediately if they put it in the wrong place anyways.

jeffro256 (approved):

src/device/device_ledger.hpp
```diff
@@ -86,6 +86,9 @@ namespace hw {
     #define SW_PROTOCOL_NOT_SUPPORTED               0x6e00
     #define SW_UNKNOWN                              0x6f00
 
+    // Origin: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledgerjs/packages/errors/src/index.ts
```
jpk68: It might be good to include the specific place where this is defined.

```suggestion
    // Origin: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledgerjs/packages/errors/src/index.ts#L337
```

jpk68:
It might be a good idea to squash your commits :)

jpk68 (approved):

jpk68 (approved):

jpk68:
See here: https://github.com/monero-project/monero/pull/10998#issuecomment-5161402936

selsta (approved):

PR #10998, opened 2026-07-29
device: report ledger state when no app is open

reservedbytes:
## Problem

A Ledger that is connected but has no app open is reported as:

```
No device found
```

That is also what an absent device reports.

The device is plugged in, powered, and often unlocked, and monero says it cannot find it.

## Why enumeration is not enough

A Ledger reports product id 0x0004 while the Monero app is running, and 0x4000 with no app open.

A device with no app open therefore matches none of `known_devices` and never reaches the transport.

The product id is the same whether that device is locked or sitting on its dashboard, so enumeration cannot tell those two states apart. They need different advice.

## Change

When none of `known_devices` matches but a Ledger is present, ask the Ledger OS which application is running.

A locked device answers `SW_LOCKED_DEVICE` to any command. An unlocked one answers from its dashboard.

That gives:

* `Ledger is locked.` when the device is locked
* `Open the Monero app on your Ledger.` when it is unlocked with no app open

The query runs only after `known_devices` has already failed to match.

A device with the Monero app open never reaches it, so its behaviour is unchanged.

Based on #10997, which defines `SW_LOCKED_DEVICE`.

### DIFF
--- a/src/device/device_ledger.cpp
+++ b/src/device/device_ledger.cpp
@@ -554,6 +554,8 @@ namespace hw {
       return true;
     }
     
+    #define LEDGER_VID 0x2c97
+
     static const std::vector<hw::io::hid_conn_params> known_devices {
         {0x2c97, 0x0001, 0, 0xffa0}, 
         {0x2c97, 0x0004, 0, 0xffa0},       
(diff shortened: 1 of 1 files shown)


src/device/device_ledger.cpp
```diff
@@ -81,10 +81,9 @@ namespace hw {
       static const char *to_string(unsigned int code);
     };
 
-    // Must be sorted in ascending order by the code
```
plowsof: ordering has changed and this comment removed. could you explain this? is the comment not true? is there actually "code" that sorts these somewhere? I have a ledger with a temp seed loaded with no monero app to try these pr's soon

reservedbytes: the comment was the intention but the table wasn´t actually sorted

`to_string` looks the code up with `std::lower_bound` which needs a sorted range, and `SW_OK` (`0x9000`) was sitting first with `SW_SECURITY_LOCKED` (`0x69ee`) before `SW_COMMAND_NOT_ALLOWED` (`0x6980`)

so `0x9000`, `0x6700` and `0x69ee` all came back as UNKNOWN, which is why every ledger error printed `EXPECTED 0x9000 (UNKNOWN)` 

j45h2pv

nothing sorts them anywhere, it´s a fixed `constexpr` table, so it was a hand maintained invariant that had quietly broken

this reorders it and adds a `constexpr` check & `static_assert` so it can´t silently break again, which is also why i dropped the comment

src/device/device_ledger.hpp
```diff
@@ -86,6 +86,9 @@ namespace hw {
     #define SW_PROTOCOL_NOT_SUPPORTED               0x6e00
     #define SW_UNKNOWN                              0x6f00
 
+    // Origin: https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledgerjs/packages/errors/src/index.ts#L337
```
selsta: I'd either remove this comment, or change the URL to a fixed hash / tag. Currently it is a branch so the line number could change.

reservedbytes: good point, changed it to a permalink

jpk68:
I tested this patch with a Ledger device, and was able to get the prompt for when the wallet is unlocked without the app open, but not for when the device is locked. According to `lsusb`, my system doesn't even recognize the device unless it's unlocked, so the error message never surfaces.

It might be good to see if someone else can test this, however, as the outcome may differ on other systems.

Edit: it seems what I reported may just be caused by the commit this PR is based off of - in that case, the changes for this PR specifically look fine to me.

jpk68:
After testing this again, it actually appears to work in both cases. The exception to this is when the Ledger device is in the "before-first-unlock" state, but this patch seems sufficient regardless.

jpk68 (approved):

src/device/device_ledger.cpp
```diff
@@ -552,8 +565,41 @@ namespace hw {
...
     bool device_ledger::connect(void) {
       this->disconnect();
+
+      unsigned int pid = 0;
+      if (find_ledger_without_app(pid)) {
+        // Ask the Ledger OS for the running application: a locked device answers
+        // SW_LOCKED_DEVICE, an unlocked one answers from its dashboard.
+        unsigned char command[] = {0xb0, 0x01, 0x00, 0x00, 0x00};
+        unsigned char response[BUFFER_RECV_SIZE];
+        hw_device.connect(LEDGER_VID, pid, 0, boost::none);
```
selsta:
```suggestion
        hw_device.connect(LEDGER_VID, pid, 0, 0xffa0);
```

or was `boost::none` chosen intentionally?

selsta (commented):
Please rebase.

src/device/device_ledger.cpp
reservedbytes: no, it wasn´t, now matches how known_devices is written too, applied

selsta (approved):

jpk68 (approved):

PR #11005, opened 2026-07-30
src: remove dependency on boost spirit

jpk68:
Continuation of #10831.

### DIFF
--- a/contrib/depends/packages/boost.mk
+++ b/contrib/depends/packages/boost.mk
@@ -90,11 +90,9 @@ define $(package)_postprocess_cmds
     ! -name multiprecision \
     ! -name numeric \
     ! -name optional \
-    ! -name phoenix \
     ! -name predef \
     ! -name preprocessor \
     ! -name program_options \
-    ! -name proto \
     ! -name range \
     ! -name ratio \
     ! -name regex \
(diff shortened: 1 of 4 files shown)


contrib/depends/packages/boost.mk
```diff
@@ -100,7 +100,6 @@ define $(package)_postprocess_cmds
     ! -name regex \
     ! -name serialization \
     ! -name smart_ptr \
-    ! -name spirit \
```
tobtoht, line 103: Looking at the dependency graph, removing `spirit` (7.9 MiB) would also allow us to get rid of `phoenix` (17.3 MiB) and `proto` (1.8 MiB).

vtnerd:
I see no reason to churn through working code, but I'll leave others to comment on it.

contrib/depends/packages/boost.mk
jpk68: Thanks, will update.

vtnerd: Everything is header based, so you only "pay for what you use". So I'm not sure what these numbers represent.

tobtoht: I know. The numbers represent the total extracted size of the headers. Compressed size is relevant for CI caching, so it's an indication at best. I'm not making any claims about binary size reduction. It remains to be seen if there is a significant benefit there.

jpk68:
Okay, I concede. This seems to be far more trouble than it's worth, and for not much benefit. My apologies to everyone whose time I wasted with this PR and the previous one.

PR #11019, opened 2026-08-01
cryptonote_basic: keep additional derivations aligned in key image helper

munzzyy:
`generate_key_image_helper` only appends to `additional_recv_derivations` when
`generate_key_derivation` succeeds:

```cpp
if (!r)
{
  MWARNING("key image helper: failed to generate_key_derivation(" << ... << ")");
}
else
{
  additional_recv_derivations.push_back(additional_recv_derivation);
}
```

That call fails when the additional tx pubkey isn't a valid point, and when it
does, every later derivation shifts down a slot. `is_out_to_acc_precomp` indexes
that vector by the output index, so once the list is short the lookup either
falls off its own bounds check or reads the derivation belonging to a different
output. Either way the helper decides the output isn't ours and returns false,
and the wallet can't build a key image for an output it owns. The pubkeys come
from the transaction, so whoever built it picks them.

The main tx pubkey a few lines up already handles a failed derivation by keeping
its slot and padding with identity. wallet2 does the same thing in all three
places it builds this list — wallet2.cpp:2393, :7887 and :13397 all
`push_back({})` first and pad on failure. This just brings the helper in line
with them.

Related: #10951 added a bounds check on the consuming side for the same
misalignment. This is the side that creates it.

### Checked locally

Built at ae3931c1d, gcc 16.1.1, boost 1.91. The test builds an account, puts an
invalid pubkey at index 0 and a good one at index 1, and asks for the output at
index 1.

Before:

```
key image helper: failed to generate_key_derivation(<...007f>, <viewkey>)
wrong number of additional derivations
key image helper: given output pubkey doesn't seem to belong to this address
  Actual: false
Expected: true
[  FAILED  ] cn_format_utils.generate_key_image_helper_additional_derivations_stay_aligned
```

After: passes. Full unit suite 1274 passed, 2 skipped, 0 failed.

One thing worth flagging for whoever reviews the test: the invalid pubkey has to
be 31 zero bytes then `0x7f`. The obvious choice of 32 `0xff` bytes is rejected
by `crypto::generate_key_derivation` but *accepted* by the
`crypto::wallet::generate_key_derivation` backend that `device_default`
actually calls, so a test written with it passes either way and proves nothing.
I hit that while writing this.

### DIFF
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -282,11 +282,9 @@ namespace cryptonote
       if (!r)
       {
         MWARNING("key image helper: failed to generate_key_derivation(" << additional_tx_public_keys[i] << ", <viewkey>)");
+        memcpy(&additional_recv_derivation, rct::identity().bytes, sizeof(additional_recv_derivation));
       }
-      else
-      {
-        additional_recv_derivations.push_back(additional_recv_derivation);
-      }
+      additional_recv_derivations.push_back(additional_recv_derivation);
     }
 
     boost::optional<subaddress_receive_info> subaddr_recv_info = is_out_to_acc_precomp(subaddresses, out_key, recv_derivation, additional_recv_derivations, real_output_index,hwdev);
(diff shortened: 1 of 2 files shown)


selsta (approved):
Change looks correct but would like a second review from someone more familiar with the code.

jeffro256 (approved):

PR #11024, opened 2026-08-01
common: don't calibrate perf timer at static init

jpk68:
Lazily calibrates the performance timer instead of doing so on startup. This effectively removes a 100ms wait from the startup time of binaries, which is especially noticeable when running commands such as `--help`.

Before:
```
Executed in  137.48 millis    fish           external
   usr time  104.19 millis  836.00 micros  103.35 millis
   sys time   11.73 millis  314.00 micros   11.41 millis
```

After:
```
Executed in   15.25 millis    fish           external
   usr time    6.09 millis    0.00 micros    6.09 millis
   sys time    9.09 millis  978.00 micros    8.11 millis
```

(both times listed above are for `./monero-wallet-cli --help`)

### DIFF
--- a/src/common/perf_timer.cpp
+++ b/src/common/perf_timer.cpp
@@ -71,13 +71,10 @@ namespace tools
   }
 #endif
 
-#ifdef __x86_64__
-  uint64_t ticks_per_ns = get_ticks_per_ns();
-#endif
-
   uint64_t ticks_to_ns(uint64_t ticks)
   {
 #if defined(__x86_64__)
+    static const uint64_t ticks_per_ns = get_ticks_per_ns();
     return 256 * ticks / ticks_per_ns;
 #else
     return ticks;
--- a/src/common/perf_timer.cpp
+++ b/src/common/perf_timer.cpp
@@ -27,6 +27,7 @@
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <vector>
+#include <atomic>
 #include "time_helper.h"
 #include "perf_timer.h"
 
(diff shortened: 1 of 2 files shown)


SNeedlewoods (approved):

thomasbuilds (approved):

src/common/perf_timer.cpp
```diff
@@ -72,12 +73,21 @@ namespace tools
 #endif
 
 #ifdef __x86_64__
-  uint64_t ticks_per_ns = get_ticks_per_ns();
+  namespace
+  {
+    bool calibrate_ticks_per_ns_async()
+    {
+      std::thread([]{ ticks_to_ns(0); }).detach();
```
thomasbuilds, line 80: This thread starts before `main()`, and `monerod --detach` / `monero-wallet-rpc --detach` `fork()` 3-4 ms later while it still holds the guard of `ticks_to_ns()`'s static. The child inherits the guard in "initialising" state with no thread to finish it so its first `ticks_to_ns()` blocks forever. Reproduced: `monerod --detach --log-level 1 --offline` hangs at blockchain init after the first `PERF ----------` line, last syscall `futex(<guard of ticks_to_ns::ticks_per_ns>, FUTEX_WAIT, ..., NULL)`, RPC never bound. Any perf logging in the child triggers it (`--log-level 1` or `set_log 1`, then the next RPC, `add_tx` or `check_tx_inputs`).

thomasbuilds (changes requested):

src/common/perf_timer.cpp
jpk68: Thanks, fixed.

src/common/perf_timer.cpp
```diff
@@ -72,12 +74,32 @@ namespace tools
...
+      ticks_per_ns_value.store(tpns, std::memory_order_release);
+      return tpns;
+    }
+
+    bool calibrate_ticks_per_ns_async()
+    {
+      std::thread(calibrate_ticks_per_ns).detach();
+      return true;
+    }
+    const bool ticks_per_ns_calibration_started = calibrate_ticks_per_ns_async();
```
thomasbuilds: Constructing the thread throws `std::system_error` if it cannot be created, which aborts every binary during static initialization. `ticks_to_ns()` already calibrates on first use, so this block can go:

```suggestion
```

src/common/perf_timer.cpp
```diff
@@ -27,6 +27,8 @@
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <vector>
+#include <thread>
```
thomasbuilds:
```suggestion
```

thomasbuilds (changes requested):
Constructing a `std::thread` in a static initializer throws when the thread cannot be created (`EAGAIN`), and an exception escaping a static initializer calls `std::terminate` before `main` is reached:

```
$ prlimit --nproc=1 ./monerod --version   # as a non-root user
terminate called after throwing an instance of 'std::system_error'
  what():  Resource temporarily unavailable
```

I suggest dropping the thread and keeping the lazy path. The startup gain is unchanged, nothing can fail during static initialization, and processes no longer spin for 100 ms in the background at every start. The only cost is a single 100 ms spin in the first thread that needs a conversion, which a detached child already pays in this revision. Concurrent first callers each spin once (8 threads: 101 ms, identical results), so no wait flag is needed; one would reintroduce the `--detach` hang.

thomasbuilds (approved):

selsta (approved):

thomasbuilds (approved):

PR #11025, opened 2026-08-01
epee: fix peak upload speed accounting

Ap4sh:
The send path currently stores upload throughput in `m_max_speed_down`, leaving `m_max_speed_up` at zero

That can inflate the peak download rate used for standby peer span scheduling

This restores the per-direction accounting used before the connection rewrite

Tests:
- `cmake --build build --target all`
- `ctest --test-dir build --output-on-failure -E core_tests`
- `ctest --test-dir build --output-on-failure -R core_tests`
- 1000 repetitions of `boosted_tcp_server.strand_deadlock`

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -547,8 +547,8 @@ namespace net_utils
           m_state.stat.out.throttle.handle_trafic_exact(bytes_transferred);
           const auto speed = m_state.stat.out.throttle.get_current_speed();
           get_context().m_current_speed_up = speed;
-          get_context().m_max_speed_down = std::max(
-            get_context().m_max_speed_down,
+          get_context().m_max_speed_up = std::max(
+            get_context().m_max_speed_up,
             speed
           );
           if (speed_limit_is_enabled()) {


selsta:
Could you revert the test changes? The fix is a straightforward enough.

monerostar (commented):
Read through this on Ubuntu 26.04 with the release binaries (monerod / monero-wallet-cli v0.18.5.1). The copy-paste bug is real on master.

In contrib/epee/include/net/abstract_tcp_server2.inl the send path at line 549 sets m_current_speed_up, then the very next line updates m_max_speed_down. m_max_speed_up is declared and zero-initialized in net_utils_base.h but is never written anywhere on master, so upload max stays 0 forever while the download max is polluted with upload traffic. m_max_speed_down is read in cryptonote_protocol_handler.inl around line 1853 for download speed throttle logic, so the pollution is not just cosmetic.

The patch fixes the write to m_max_speed_up, and the test change now records both directions and asserts max_up > max_down, which fails on the old code. CI is green including Ubuntu 24.04 tests.

Looks good to me.

jpk68 (approved):

Ap4sh:
done 👍

selsta:
please squash

Ap4sh:
done

selsta (approved):

jpk68 (approved):

tobtoht:
Needs rebase

selsta (approved):

PR #11026, opened 2026-08-02
simplewallet: show default value in Y/N prompts

### LINKED ISSUE #11003: Default prompt options in the CLI wallet are unintuitive
In the CLI wallet, input is typically requested from the user by means of text prompts (such as `Y/Yes/N/No`), however the behaviour of these prompts is often confusing and inconsistent. For example, the following message related to enabling background mining will be displayed upon creating a new wallet:

```
Do you want to do it now? (Y/Yes/N/No):
```

The "default" behaviour of this (what happens when you hit the 'enter' key without giving any input) is to set the option to "false". The user has no way to know this is the case (other than simply trying it and seeing what happens). Some other prompts (e.g. selecting a seed language, setting a restore date) will continue to ask the user until some input is given, or just abruptly exit the process.

I don't have any particular ideas of how this can be remedied at the moment, but it would be nice for an indication of the "default" option to somehow be provided for clarity and ease-of-use purposes.

bsandham:
Closes #11003

## Changes

Currently, yes/no prompts in the CLI wallet do not indicate what the default behaviour is when pressing Enter. In practice, pressing Enter always defaults to No.

This PR:
1. Updates the function to display more detailed message instead of just Y/Yes/N/No
2. Refactors four call sites that manually appended the Y/N prompt string to use the existing  parameter instead, eliminating duplicated prompt formatting

## Testing

- Built and linked successfully with GCC 14
- All existing Y/N prompts now consistently show the default value

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -288,7 +288,7 @@ namespace
     PAUSE_READLINE();
     std::cout << prompt;
     if (yesno)
-      std::cout << "  (Y/Yes/N/No)";
+      std::cout << "  [y/N]";
     std::cout << ": " << std::flush;
 
     std::string buf;
@@ -5189,7 +5189,7 @@ void simple_wallet::check_background_mining(const epee::wipeable_string &passwor
     message_writer() << tr("The daemon is not set up to background mine.");
     message_writer() << tr("With background mining enabled, the daemon will mine when idle and not on battery.");
     message_writer() << tr("Enabling this supports the network you are using, and makes you eligible for receiving new monero");
-    std::string accepted = input_line(tr("Do you want to do it now? (Y/Yes/N/No)"));
+    std::string accepted = input_line(tr("Do you want to do it now?"), true);
     if (std::cin.eof() || !command_line::is_yes(accepted)) {
       m_wallet->setup_background_mining(tools::wallet2::BackgroundMiningNo);
       m_wallet->rewrite(m_wallet_file, password);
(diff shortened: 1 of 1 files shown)


src/simplewallet/simplewallet.cpp
```diff
@@ -288,7 +288,7 @@ namespace
     PAUSE_READLINE();
     std::cout << prompt;
     if (yesno)
-      std::cout << "  (Y/Yes/N/No)";
+      std::cout << "  (Y/Yes/N/No, default: No)";
```
jpk68: IMO, this looks too cluttered. Instead, using something like `[y/N]` would be more clean and idiomatic

bsandham: agree  -changed to unix convention y/N, please see https://github.com/monero-project/monero/pull/11026/commits/b90b88cb8e6f598974dde6f905b05192e7bd7380

SNeedlewoods:
Please squash,
and the second commit message says
> Review feedback from selsta

but according to the comment I see it came from jpk68.

bsandham:
squashed + corrected msg, thanks for review @SNeedlewoods

SNeedlewoods (approved):

jpk68 (approved):
LGTM, thanks.

nahuhh:
IMO the commit message should be de-slopped

can be reduced to just the first sentence.. or even blank. The commit title seems self explanatory

bsandham:
trimmed commit msg to be the PR title as per @nahuhh 's suggestion

jpk68 (approved):

selsta (approved):

PR #11032, opened 2026-08-02
doxyfile: remove comments, improve build performance

jpk68:
Removes comments from the `Doxyfile`, which decreases the file size by over 90%.

Also adds the following optimizations to improve 'build' times:

```
LOOKUP_CACHE_SIZE      = 1
NUM_PROC_THREADS       = 0
```


Doxyfile
```diff
@@ -1,2310 +1,257 @@
...
 GENERATE_LEGEND        = YES
-
-# If the DOT_CLEANUP tag is set to YES doxygen will remove the intermediate dot
-# files that are used to generate the various graphs.
-# The default value is: YES.
-# This tag requires that the tag HAVE_DOT is set to YES.
-
 DOT_CLEANUP            = YES
+NUM_PROC_THREADS       = 4
+DOT_NUM_THREADS        = 4
```
selsta:
```suggestion
NUM_PROC_THREADS       = 0
```

based on the documentation this is the ideal setting which uses number of available cores.

https://www.doxygen.nl/manual/config.html

jpk68: Thanks, done.

jeffro256:
NACK. The comments can be helpful and shouldn't be removed unless it can be demonstrated that removing comments noticeably affects performance.

jpk68:
> NACK. The comments can be helpful and shouldn't be removed unless it can be demonstrated that removing comments noticeably affects performance.

IMO, it would be better to just rely on Doxygen's own documentation as a source of truth; the comments in this file are quite outdated, and I feel like having to wade through all of that visual bloat is a burden on more related improvements being found.

PR #11035, opened 2026-08-03
wallet_api: reject transaction file signing with hardware wallets

selsta:
Reported by zkao, a tool by zkSecurity

### DIFF
--- a/src/wallet/api/unsigned_transaction.cpp
+++ b/src/wallet/api/unsigned_transaction.cpp
@@ -76,6 +76,12 @@ bool UnsignedTransactionImpl::sign(const std::string &signedFileName)
      m_status = Status_Error;
      return false;
   }
+  if(m_wallet.m_wallet->key_on_device())
+  {
+     m_errorString = tr("Command not supported by hardware wallet");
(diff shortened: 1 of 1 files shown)


jpk68 (approved):

jeffro256:
Is this the always-planned behavior? I assume that zkao found some bug with offline-HW signing, and this was the easiest course of action, but offline HW signing should be supported long-term IMO.

selsta:
Trezor and Ledger both don't support air-gapped transaction file workflow.

jeffro256:
Maybe not the specific transaction file workflow, but I do know that Trezor supports cold-signing in the sense that it consumes an entire transaction intent, then spits out a signature. It is obviously very under-tested, but I do think that Trezor could likely support it with Monero Core-side changes.

selsta:
This PR does not disable Trezor cold signing `device_cold::tx_sign()`, which goes through a different code path. The PR title is a bit inaccurate, will update it.

SNeedlewoods (approved):

jpk68 (approved):

PR #11037, opened 2026-08-03
p2p: use a bool for send()

j-berman:
Identified by @selsta

Callers can otherwise misinterpret int -1 as truthy success.

Example bug:

- `make_payload_send_txs` would return true even on failure [here](https://github.com/monero-project/monero/blob/83405f127d006c470f7b8715895ebb318da5fcca/src/cryptonote_protocol/levin_notify.cpp#L564).
- Thus even if a stem tx fails to relay, the node incorrectly assumes it succeeded and won't re-attempt relay until the re-relay check loop.


jpk68 (approved):

vtnerd (approved):

selsta (approved):

jeffro256 (approved):

selsta:
Needs rebase.

selsta (approved):

jpk68 (approved):

PR #11040, opened 2026-08-04
simplewallet: show error for timed-out hardware wallets

### LINKED ISSUE #11038: CLI wallet will reject correct passwords after inactivity lock is triggered
When the CLI wallet is automatically locked due to inactivity, and the user is prompted for a password to unlock it, *any* password (including the correct one) will be rejected if the hardware device being used has also "timed out".

The correct behaviour in this situation would seemingly be to alert the user that the device must be unlocked as well.

jpk68:
Fixes #11038

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -6276,6 +6276,13 @@ void simple_wallet::check_for_inactivity_lock(bool user)
           break;
         }
       }
+      catch (const std::exception &e)
+      {
+        // Report why unlocking failed, rather than just looping back to the password prompt
+        fail_msg_writer() << tr("Failed to unlock wallet: ") << e.what();
+        if (m_wallet->key_on_device())
+          fail_msg_writer() << tr("Please ensure the HW wallet is connected and unlocked.");
+      }
       catch (...) { /* do nothing, just let the loop loop */ }
     }
     m_last_activity_time = time(NULL);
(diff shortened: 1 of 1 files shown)


src/simplewallet/simplewallet.cpp
```diff
@@ -6276,6 +6276,13 @@ void simple_wallet::check_for_inactivity_lock(bool user)
           break;
         }
       }
+      catch (const std::exception &e)
+      {
+        // Report why unlocking failed, rather than just looping back to the password prompt
+        fail_msg_writer() << tr("Failed to unlock wallet: ") << e.what();
+        if (m_wallet->key_on_device())
+          fail_msg_writer() << tr("Please ensure the HW wallet is connected and unlocked.");
```
thomasbuilds: `fail_msg_writer()` prepends `Error: ` on every call (`simplewallet.cpp:372-375`), so for HW
wallets the current form prints two `Error:` lines for a single failure.

Also `verify_password()` throws more than device errors - `error::file_read_error`
(`wallet2.cpp:5422`) renders the keys file path into `what()`, disclosing a filename the lock
screen otherwise hides by default (`show-wallet-name-when-locked`).

```suggestion
        if (m_wallet->key_on_device())
        {
          fail_msg_writer() << tr("Failed to unlock wallet: ") << e.what() << "\n"
                            << tr("Please ensure the hardware wallet is connected and unlocked.");
        }
        else
        {
          fail_msg_writer() << tr("Failed to unlock wallet.");
        }
```

thomasbuilds (changes requested):

src/simplewallet/simplewallet.cpp
jpk68: Thanks, fixed.

thomasbuilds (approved):

selsta (approved):

PR #11042, opened 2026-08-04
device_ledger: fix some comments

jpk68:
Fixes a few typos, grammar issues, and generally incorrect information in comments.

### DIFF
--- a/src/device/device_ledger.cpp
+++ b/src/device/device_ledger.cpp
@@ -332,7 +332,7 @@ namespace hw {
       MDEBUG( "Device "<<this->name << " LOCKed");
     }
 
-    //lock the device for a long sequence
+    //try to lock the device for a long sequence
     bool device_ledger::try_lock(void) {
       MDEBUG( "Ask for LOCKING(try) for device "<<this->name << " in thread ");
       bool r = device_locker.try_lock();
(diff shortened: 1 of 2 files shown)


selsta (approved):

src/device/device_ledger.cpp
```diff
@@ -632,7 +632,7 @@ namespace hw {
     bool  device_ledger::get_secret_keys(crypto::secret_key &vkey , crypto::secret_key &skey) {
         AUTO_LOCK_CMD();
 
-        //secret key are represented as fake key on the wallet side
+        //secret keys are represented as fake key on the wallet side
```
thomasbuilds: I think PRs like this are a bit pointless but let's make it fully correct
```suggestion
        //secret keys are represented as fake keys on the wallet side
```

src/device/device_ledger.cpp
thomasbuilds:
```suggestion
    //try to lock the device for a long sequence
```

src/device/device_ledger.hpp
thomasbuilds, line 164:
```suggestion
         * @brief Request user to export view privkey if not requested before, then return whether it has been exported
```

missing "to"

thomasbuilds (changes requested):

src/device/device_ledger.hpp
jpk68: I think this sounds grammatically correct as it is.

thomasbuilds (approved):

selsta (approved):

PR #11045, opened 2026-08-04
device_trezor: add trezor host protocol support

jpk68:
Note: this PR is very much a WIP, and I cannot guarantee it will go anywhere.

Depends on #11047


jpk68:
Closing because, let's be realistic, this is not happening.

PR #11047, opened 2026-08-04
device_trezor: update protobuf definitions

jpk68:
Updates the Protobuf definitions from upstream. This will be useful for later integrating Trezor Host Protocol support.


tobtoht (approved):
Checked against upstream.

selsta (approved):

PR #11048, opened 2026-08-04
p2p: update peer's height on new block & widen window for relay

j-berman:
Currently, a node sets its peer connection's `m_remote_blockchain_height` during the handshake, initial sync, when syncing after falling out of sync, and in the `peer_sync_idle_maker`. It doesn't update it upon receiving a new bock.

The timed sync only runs once every 60s (`P2P_DEFAULT_HANDSHAKE_INTERVAL`). Thus, `m_remote_blockchain_height` can fall below our own node's actual height.

The peer connection's `m_remote_blockchain_height` is used to identify outgoing connections to relay txs to ([see `get_out_connections`](https://github.com/monero-project/monero/blob/1bf1dff4be4b2a604fd761b532e1c5aff4cc0bad/src/cryptonote_protocol/levin_notify.cpp#L151)). If the peer's height is less than ours, we don't relay txs to that peer.

Thus, without this PR, our node can avoid relaying txs to peers that technically are fine candidates to receive them.

This PR fixes the issue by updating the `m_remote_blockchain_height` as soon as the peer reports they receive (and validate a new block), and widens the window for candidate peers that we can relay to.

### DIFF
--- a/src/cryptonote_protocol/cryptonote_protocol_handler.inl
+++ b/src/cryptonote_protocol/cryptonote_protocol_handler.inl
@@ -636,8 +636,10 @@ namespace cryptonote
     }
 
     // Log block info
+    const uint64_t new_block_height = get_block_height(new_block);
+    const uint64_t peer_height = arg.current_blockchain_height;
     MLOG_P2P_MESSAGE(context << "Received NOTIFY_NEW_FLUFFY_BLOCK " << new_block_hash << " (height "
-      << arg.current_blockchain_height << ", " << arg.b.txs.size() << " txes)");
+      << new_block_height << ", " << arg.b.txs.size() << " txes, peer's height: " << peer_height << ")");
 
     // Pause mining and resume after block verification to prevent wasted mining cycles while
     // validating the next block. Needs more research into if this is a DoS vector or not. Invalid
@@ -726,7 +728,7 @@ namespace cryptonote
           MDEBUG("  tx " << new_block.tx_hashes[txidx]);
         NOTIFY_REQUEST_FLUFFY_MISSING_TX::request missing_tx_req;
         missing_tx_req.block_hash = new_block_hash;
-        missing_tx_req.current_blockchain_height = arg.current_blockchain_height;
+        missing_tx_req.current_blockchain_height = new_block_height + 1;
         missing_tx_req.missing_tx_indices = std::move(need_tx_indices);
 
         // Post NOTIFY_REQUEST_FLUFFY_MISSING_TX request to peer
@@ -743,6 +745,7 @@ namespace cryptonote
     }
     else if( bvc.m_added_to_main_chain )
     {
+      arg.current_blockchain_height = new_block_height + 1;
       // Relay an empty block
       arg.b.txs.clear();
       relay_block(arg, context);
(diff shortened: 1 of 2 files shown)


src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -639,6 +639,15 @@ namespace cryptonote
     MLOG_P2P_MESSAGE(context << "Received NOTIFY_NEW_FLUFFY_BLOCK " << new_block_hash << " (height "
       << arg.current_blockchain_height << ", " << arg.b.txs.size() << " txes)");
 
+    const uint64_t new_block_height = get_block_height(new_block);
+    if (arg.current_blockchain_height != (new_block_height + 1))
+    {
+      LOG_ERROR_CCONTEXT("Sent current height != (new block height + 1) [" << arg.current_blockchain_height << " != " << (new_block_height + 1) << "]");
+      // TODO: can uncomment at next fork, once we're sure all nodes are relaying as expected
```
j-berman: See #11050 which sets `arg.current_blockchain_height` using the block's height when relaying a block

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -762,6 +771,13 @@ namespace cryptonote
       MLOG_PEER_STATE("requesting chain");
     }
 
+    if (bvc.m_added_to_main_chain || bvc.m_already_exists)
```
j-berman, line 768: See #11049 which ensures the connection gets dropped a few lines above when the connection relays an already existing invalid block

src/cryptonote_protocol/cryptonote_protocol_handler.inl
Boog900: In Cuprate we relay alt blocks that are close to the top of the chain. We set `current_blockchain_height` to the current blockchain height regardless of the blocks height. IMO monerod should be relaying alt blocks too. 

monerod should be changing this value to it's own blockchain height before relay though.

j-berman: Removed this whole if statement in the latest and included a comment highlighting this point^

Boog900:
Do we update `current_blockchain_height` on block relay? it looks like we are just passing it though using whatever value we were told for `current_blockchain_height` and sending that in our broadcast.

j-berman:
We don't, but it's only ever relayed in the case of `bvc.m_added_to_main_chain=true`. And when `m_added_to_main_chain=true`, it means the block just added was just placed at the tip of the chain. So that behavior seems ok to me

Boog900:
But couldn't someone just put a random number for `current_blockchain_height` and monerod will broadcast that random number with the block to all its peers?

j-berman:
Fair point, check the latest

src/cryptonote_protocol/cryptonote_protocol_handler.inl
Boog900: We should probably fix this one too:
```suggestion
        missing_tx_req.current_blockchain_height = new_block_height + 1;
```

Boog900: I'm not sure that github put that on the right line

j-berman: Done

selsta (approved):

Boog900 (approved):

PR #11051, opened 2026-08-05
rpc: make thread count configurable, use better default value

jpk68:
Adds a new command-line option to the daemon (`--rpc-threads`), allowing the user to choose how many threads the RPC server should use. Its default value is the system's hardware concurrency, with upper and lower bounds of eight and four, respectively.

This should improve the scaling of RPC wallet sync by a small (but non-negligible) amount.

### DIFF
--- a/src/daemon/rpc.h
+++ b/src/daemon/rpc.h
@@ -51,6 +51,7 @@ public:
 private:
   cryptonote::core_rpc_server m_server;
   const std::string m_description;
+  const std::size_t m_threads_count;
 public:
   t_rpc(
       boost::program_options::variables_map const & vm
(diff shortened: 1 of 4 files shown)


selsta:
This change needs good testing, if there are too many RPC threads it's possible P2P won't be able to grab locks and start to fall behind with block sync. We have seen behaviour like this before with only 2 threads.

jpk68:
@Boog900 kindly tested and made some graphs for this PR, measuring wallet sync performance (shown below).

In this case, it may be better to keep the default set to `2` (as it was before), but still provide the configuration option for nodes which could benefit from the performance increase when syncing a large number of wallets.

relative-scaling

jpk68:
Reverted the default value back to hardware concurrency, as it will allow for much better performance when combined with #11143.

jpk68:
Rebased.

PR #11053, opened 2026-08-05
epee: enable tcp_nodelay for rpc connections

### LINKED ISSUE #11052: `TCP_NODELAY` is disabled for RPC connections
It appears that in the `epee` TCP server code (see [here](https://github.com/monero-project/monero/blob/master/contrib/epee/include/net/abstract_tcp_server2.inl#L994)), `TCP_NODELAY` is explicitly disabled for every accepted connection. This forces the use of [Nagle's algorithm](https://en.wikipedia.org/wiki/Nagle%27s_algorithm) for both P2P and RPC; it would likely be worth enabling for the latter.

jpk68:
Fixes #11052

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -991,7 +991,8 @@ namespace net_utils
     }
     #endif
     connection_basic::socket_.next_layer().set_option(
-      boost::asio::ip::tcp::no_delay{false},
+      // Enable `TCP_NODELAY` for RPC connections to avoid added latency
+      boost::asio::ip::tcp::no_delay{m_connection_type == e_connection_type_RPC},
       ec
     );
     if (ec.value())


contrib/epee/include/net/abstract_tcp_server2.inl
```diff
@@ -991,7 +991,8 @@ namespace net_utils
     }
     #endif
     connection_basic::socket_.next_layer().set_option(
-      boost::asio::ip::tcp::no_delay{false},
+      // Disable `TCP_NODELAY` for RPC connections to avoid added latency
```
selsta:
```suggestion
      // Enable `TCP_NODELAY` for RPC connections to avoid added latency
```

enabling this flag disables the algorithm from what I understand

jpk68: Thanks, fixed.

thomasbuilds (approved):

PR #9858, opened 2025-03-23
CMakeLists: add extra warnings and protections

0xFFFC0000:

PR #11054, opened 2026-08-06
cmake: add more compiler warnings

jpk68:
Adds several extra warnings and protections to `CMakeLists.txt`.

Related: #9858

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -755,6 +755,18 @@ include(CheckTrezor)
   set(C_WARNINGS "-Waggregate-return -Wnested-externs -Wold-style-definition -Wstrict-prototypes")
   set(CXX_WARNINGS "-Wno-reorder -Wno-missing-field-initializers")
 
+  add_c_flag_if_supported(-Wimplicit-fallthrough C_WARNINGS)
+  add_cxx_flag_if_supported(-Wimplicit-fallthrough CXX_WARNINGS)
+  add_c_flag_if_supported(-Wredundant-decls C_WARNINGS)
+  add_cxx_flag_if_supported(-Wredundant-decls CXX_WARNINGS)
+
+  add_c_flag_if_supported(-Wduplicated-branches C_WARNINGS)
+  add_cxx_flag_if_supported(-Wduplicated-branches CXX_WARNINGS)
+  add_c_flag_if_supported(-Wduplicated-cond C_WARNINGS)
+  add_cxx_flag_if_supported(-Wduplicated-cond CXX_WARNINGS)
+
+  add_cxx_flag_if_supported(-Woverloaded-virtual CXX_WARNINGS)
+
   monero_enable_coverage()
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -767,6 +779,8 @@ include(CheckTrezor)
     set(CXX_SECURITY_FLAGS "${CXX_SECURITY_FLAGS} -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1")
   endif()
 
+  set(CXX_SECURITY_FLAGS "${CXX_SECURITY_FLAGS} -D_GLIBCXX_ASSERTIONS")
```
thomasbuilds: This is a runtime behaviour change rather than a warning - out-of-bounds `operator[]` aborts instead of being UB. It's also unconditional, unlike the `_FORTIFY_SOURCE` above, which is gated on `Release AND NOT OPENBSD`. Worth splitting into its own PR?

CMakeLists.txt
```diff
@@ -781,6 +795,9 @@ include(CheckTrezor)
     add_cxx_flag_if_supported(-fstack-protector-strong CXX_SECURITY_FLAGS)
   endif()
 
+  add_c_flag_if_supported(-fstack-reuse=none C_SECURITY_FLAGS)
+  add_cxx_flag_if_supported(-fstack-reuse=none CXX_SECURITY_FLAGS)
```
thomasbuilds: This takes `json::run_handler`'s frame from 800 to 29,264 bytes (GCC 13.3, x86-64). It recurses, is capped at `EPEE_JSON_RECURSION_LIMIT_INTERNAL = 100`, and is fed from the RPC request body via `load_t_from_json`, so worst case goes ~80 KB → ~2.9 MB, on RPC threads that take the platform default stack (`http_server_impl_base.h:112` passes no attributes). Still under 8 MB, but a large margin reduction on an attacker-reachable path.

Is the stack-slot hardening worth that trade?

thomasbuilds (changes requested):

thomasbuilds (approved):

SNeedlewoods (approved):
With these flags I see the following new warnings when building `master`:
`-Wduplicated-branches`:
https://github.com/monero-project/monero/blob/3646f648db57f60cca86430e25a635d19fa9b92a/src/cryptonote_protocol/levin_notify.cpp#L735
https://github.com/monero-project/monero/blob/3646f648db57f60cca86430e25a635d19fa9b92a/src/wallet/wallet2.cpp#L12982
`-Woverloaded-virtual`:
https://github.com/monero-project/monero/blob/3646f648db57f60cca86430e25a635d19fa9b92a/src/blockchain_db/blockchain_db.h#L821-L827
`add_block() was hidden` in multiple places

IMO that's not too spammy, so I approve

selsta (approved):

thomasbuilds (approved):

PR #11055, opened 2026-08-06
wallet2: honor max_blocks bound during fast_refresh hash pulls

woodser:
Currently, monero-wallet-rpc cannot report any sync progress during the hash skip phase of wallet syncing, and because it's single-threaded, it cannot respond to any call during this phase. As a result, when restoring a wallet from a previous height, the wallet can be completely unresponsive for 3+ minutes during the hash skip phase. Even for newer wallets, the delay before reporting any sync progress is noticeable and significant.

Applications like Haveno will timeout after several minutes of no observed sync progress and attempt to connect to another monerod, which will run into the same problem again. So this is a real blocker to restore old wallets through the official API or native bindings to wallet2.

The official GUI does not suffer this problem because wallet_api refreshes on a dedicated thread and reads heights in process. As a result, restoring an old wallet in the official Monero GUI starts reporting sync progress immediately and responsively.

monero-wallet-rpc already has a constant to improve this, [`REFRESH_INDICATIVE_BLOCK_CHUNK_SIZE = 256`](https://github.com/monero-project/monero/blob/1bf1dff4be4b2a604fd761b532e1c5aff4cc0bad/src/wallet/wallet_rpc_server.cpp#L63), with a comment "just to split refresh in separate calls to play nicer with other threads". However, wallet2 ignores this bound within `fast_refresh` during the hash skip phase.

This PR makes it so that wallet2 honors the bound, so that when `refresh` is called with capped `max_blocks`, the internal `fast_refresh` is also capped with a maximum pulls per refresh. When the budget is hit, progress is reported via `blocks_fetched` and it returns, then the next call will resume where it left off. Unbounded callers (refresh RPC, simplewallet, wallet_api) pass no bound and are unchanged (max pulls = 0).

I've tested and verified that this PR allows wallets to report sync progress with the same immediate responsiveness as the official Monero GUI, whether by RPC or directly with native bindings, significantly improving the user experience. Without this PR, the user sees a substantial delay before any sync progress is reported (3+ minutes for older wallets), while with this PR, sync progress starts to be reported immediately. I've been running this fix in Haveno's RPC and native libs and observe the same immediate responsiveness as with the official GUI.

This PR ships with `MAX_HASH_PULLS_PER_REFRESH = 3` (~30k hashes per call) as a grounded starting point, but it could be tuned or derived from `max_blocks`.

Release PR: https://github.com/monero-project/monero/pull/11056


nahuhh:
Can you double check that this is an issue on master builds?

master should skip most of the hashes, release still pulls them all unless https://github.com/monero-project/monero/pull/10856 lands

edit: this is likely still a good optimization, i am only wondering if the same issue is still present on master

jpk68 (approved):
Changes look good to me; approving under the assumption that the issue is still on `master` (it seems to be?)

woodser:
> Can you double check that this is an issue on master builds?

The change on master improves the problem by only syncing block hashes since the last checkpoint. But the remaining pull from that checkpoint up to the restore height still runs unbounded inside a single refresh(), and wallet rpc is unresponsive to all calls for that duration which grows until the next checkpoint.

The remainder can be substantial, up to hundreds of thousands of blocks, which can be tens of seconds to minutes against a remote node.

So the two fixes are complementary. The checkpoint skip shortens the hash skip phase, while this PR bounds it per refresh call so wallet rpc stays responsive during whatever remains.

selsta:
We don't plan to include https://github.com/monero-project/monero/pull/10856 in v0.18, is this fix still correct without it on v0.18 release branch?

woodser:
Yes, it doesn't depend on #10856.

selsta (approved):

nahuhh:
needs backport cc @selsta

tobtoht:
@nahuhh It has the label.

nahuhh:
> @nahuhh It has the label.

Oh, my bad. GitHub only showed me pending review add/removal (or im blind)

woodser:
> We don't plan to include #10856 in v0.18, is this fix still correct without it on v0.18 release branch?

I just wanted to clarify if #10856 can/will ever be merged into the release-v0.18 branch?

I ask because it's a significant UX improvement when restoring old wallets, and the change is complete as far as I can tell, with most of it already merged to master.

Just hoping #10856 will be able to land eventually.

PR #11058, opened 2026-08-06
tx_pool: fix write abort and improve locking for relayable txs

jeffro256:
Since `lock.commit()` was not called here, the mempool meta updates weren't being committed to the database, but rather were aborted, as per the behavior of the destructor of `LockedTXN`. This meant that the `last_relayed_time` of txs was not being updated. This PR fixes this and also only attempts to open a write TXN if there are actualy changes to be made. It also defers opening the write TXN until later to improve concurrency.

### DIFF
--- a/src/cryptonote_core/tx_pool.cpp
+++ b/src/cryptonote_core/tx_pool.cpp
@@ -810,7 +810,6 @@ namespace cryptonote
 
     CRITICAL_REGION_LOCAL(m_transactions_lock);
     CRITICAL_REGION_LOCAL1(m_blockchain);
-    LockedTXN lock(m_blockchain.get_db());
     txs.reserve(m_blockchain.get_txpool_tx_count());
     m_blockchain.for_all_txpool_txes([this, now, &txs, &change_timestamps, &next_check](const crypto::hash &txid, const txpool_tx_meta_t &meta, const cryptonote::blobdata_ref *){
       // 0 fee transactions are never relayed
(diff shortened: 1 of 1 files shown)


src/cryptonote_core/tx_pool.cpp
```diff
@@ -850,16 +858,32 @@ namespace cryptonote
...
+          made_an_update = true;
+        }
+        catch (...)
+        {
+          MDEBUG("Got an exception while updating txpool meta for relayable tx " << elem.first << ", ignoring...");
+          continue;
+        }
+      }
+      if (made_an_update)
+        wtxn_guard.stop();
```
j-berman: nit: the `db_wtxn_guard` destructor calls `stop()`, so there isn't any benefit to the `made_an_update` logic.

I would suggest using `LockedTXN` over `db_wtxn_guard` here since `LockedTXN` has some nice properties already inherent (`batch_start` handles resizings, `commit()` catches exceptions), plus we use `LockedTXN` across this file already.

With `LockedTXN`, then this `made_an_update` logic makes sense

j-berman (commented):
Nice catch, just a nit

j-berman (approved):
Comment was a nit, can be ignored

src/cryptonote_core/tx_pool.cpp
j-berman: Could also wrap the entire `if (!change_timestamps.empty())` inner logic in a try-catch here and retain `db_wtxn_guard`, rather than just try-catch the single `update`

jeffro256: If I understand you correctly, this would skip over updating later entries

j-berman (approved):

selsta (approved):

PR #11062, opened 2026-08-06
crypto: improve CSPRNG backtracking resistance

selsta:
Reported by xmrack and the MAGIC Monero Fund.

https://keccak.team/files/SpongePRNG.pdf


jeffro256:
Here's a recommendation:

```diff
diff --git a/src/crypto/random.c b/src/crypto/random.c
index 12fd55d47..7e2b610be 100644
--- a/src/crypto/random.c
+++ b/src/crypto/random.c
@@ -120,10 +120,16 @@ static void lock_random_state(void) {
 #endif
 }
 
+#define HASH_CAPACITY_AREA 64
+
 INITIALIZER(init_random) {
-  generate_system_random_bytes(32, &state);
-  generate_system_random_bytes(sizeof(crypto_siphash_key), crypto_siphash_key);
+  static_assert(sizeof(state) == (HASH_DATA_AREA + HASH_CAPACITY_AREA),
+    "Keccak state is the wrong size");
   lock_random_state();
+  generate_system_random_bytes(HASH_DATA_AREA, &state);
+  hash_permutation(&state);
+
+  generate_system_random_bytes(sizeof(crypto_siphash_key), crypto_siphash_key);
   REGISTER_FINALIZER(deinit_random);
 #if !defined(NDEBUG)
   assert(curstate == 0);
@@ -132,34 +138,35 @@ INITIALIZER(init_random) {
 }
 
 void generate_random_bytes_not_thread_safe(size_t n, void *result) {
+  /*
+   * This function assumes that the state is recently permutated. We permutate
+   * *after* squeezing so that a memory leak of the state immediately after
+   * leaving this function does not leak the previous squeezed values.
+   */
+
 #if !defined(NDEBUG)
   assert(curstate == 1);
   curstate = 2;
 #endif
-  if (n == 0) {
-#if !defined(NDEBUG)
-    assert(curstate == 2);
-    curstate = 1;
-#endif
-    return;
+
+  while (n > HASH_DATA_AREA) {
+    memcpy(result, &state, HASH_DATA_AREA);
+    hash_permutation(&state);
+    result = padd(result, HASH_DATA_AREA);
+    n -= HASH_DATA_AREA;
   }
-  for (;;) {
+
+  if (n) {
+    memcpy(result, &state, n);
+    /* See section 4.3 "Forward security" of https://keccak.team/files/SpongePRNG.pdf */
+    memset(&state, 0, HASH_CAPACITY_AREA);
     hash_permutation(&state);
-    if (n <= HASH_DATA_AREA) {
-      memcpy(result, &state, n);
-      hash_permutation(&state);
-      memset(state.b + sizeof(state.b) - 16, 0, 16);
+  }
+
 #if !defined(NDEBUG)
-      assert(curstate == 2);
-      curstate = 1;
+  assert(curstate == 2);
+  curstate = 1;
 #endif
-      return;
-    } else {
-      memcpy(result, &state, HASH_DATA_AREA);
-      result = padd(result, HASH_DATA_AREA);
-      n -= HASH_DATA_AREA;
-    }
-  }
 }
 
 void add_extra_entropy_not_thread_safe(const void *ptr, size_t bytes)
@@ -168,11 +175,11 @@ void add_extra_entropy_not_thread_safe(const void *ptr, size_t bytes)
 
   while (bytes > 0)
   {
-    hash_permutation(&state);
     const size_t round_bytes = bytes > HASH_DATA_AREA ? HASH_DATA_AREA : bytes;
     for (i = 0; i < round_bytes; ++i)
       state.b[i] ^= ((const uint8_t*)ptr)[i];
     bytes -= round_bytes;
     ptr = cpadd(ptr, round_bytes);
+    hash_permutation(&state);
   }
 }

```

This does a few things:
- Locks the memory before sampling the initial seed
- Samples $r$ bits from initial seed, instead of $128$ bits. $c > 128$, which means that the Keccak capacity has a theoretical security over the 32 bytes we were previously sampling. Previous sample size also assumes that `/dev/urandom` entropy is perfect
- Makes the performance comparable to existing version (not 2x slower)
- Setting 0s at the front of the state acts as-if feeding bytes of the output of the hash into the sponge, which has precedent, unlike setting 0s of the capacity, which does not AFAIK
- Adds a link to relevant paper

selsta:
Updated to implement @jeffro256's suggested diff.

SChernykh (approved):

jeffro256:
Wait one more thing, this still allows backtracking of the last $8n - c$ bits if $8n > c$, where $c = 512$. Not directly applicable to anything used in Monero rn AFAIK, but still an unsafe API.

SChernykh:
Edit: see my comment below

src/crypto/random.c
```diff
@@ -122,32 +138,35 @@ INITIALIZER(init_random) {
...
+    hash_permutation(&state);
+    result = padd(result, HASH_DATA_AREA);
+    n -= HASH_DATA_AREA;
   }
-  for (;;) {
+
+  if (n) {
+    memcpy(result, &state, n);
+    /* See section 4.3 "Forward security" of https://keccak.team/files/SpongePRNG.pdf */
+    memset(&state, 0, HASH_CAPACITY_AREA);
```
SChernykh: I think it should be `HASH_DATA_AREA` here, because the layout of the 200-byte state is: first 136 bytes - `HASH_DATA_AREA`, last 64 bytes - `HASH_CAPACITY_AREA`. So this `memset` should clear the whole data area and leave capacity area intact so that enough entropy is preserved for the next permutation.

SChernykh (changes requested):

src/crypto/random.c
jeffro256: We can theoretially keep more permutations by only feeding exactly as many bits as were read < r, but I think that this is good enough.

selsta:
Updated.

SChernykh (approved):

jeffro256 (approved):

jeffro256:
Requesting review from @UkoeHB

UkoeHB (approved):
This PR is strictly better than the existing impl, but note it is not a 'complete' solution. If the state leaks, then it will expose the results of future calls to `generate_random_bytes_not_thread_safe` until `add_extra_entropy_not_thread_safe` is called again. This PR only obscures previous outputs.

You'd need to call `add_extra_entropy_not_thread_safe` with the output of `generate_system_random_bytes` after every invocation to be maximally robust (which may not be performant/feasible). It looks like there's only one call to `add_extra_entropy_thread_safe` deep in the wallet code.

SChernykh:
Correct, but this PR addresses backtracking resistance specifically. Adding more entropy to RNG's state periodically can be done in another PR.

PR #11065, opened 2026-08-07
cryptonote_basic: `parse_and_validate_tx` param that checks blob max size before parsing

j-berman:
Simple API to check tx's blob size before parsing. This is useful when reading blobs from untrusted sources.

Example: in [this PR](https://github.com/seraphis-migration/monero/pull/450), I move parsing newly received pool tx blobs up and out of `m_core.handle_incoming_tx` to enable a significantly more efficient tx relay v2. We benefit from checking tx blob max size before parsing there.

@selsta pointed out that coinbase txs can technically be larger than `get_max_tx_size()`, otherwise we could just require all tx blobs be <= max without a param.


selsta (approved):

thomasbuilds (approved):

jeffro256:
Needs rebase

selsta (approved):

jpk68 (approved):

PR #11066, opened 2026-08-07
epee: replace some fragile code

jpk68:
- Fixes potential unbounded memory growth, as the per-host connection count map didn't remove entries once a host's count returned to zero
- Fixes a faulty comparison which could have resulted in division by zero


selsta (changes requested):
NACK on the `median()` change. This is not safe as written because consensus relevant code relies on its current ordering behavior.

selsta (approved):

SNeedlewoods (approved):

PR #11067, opened 2026-08-07
cryptonote_basic: sanity check on key offsets

j-berman:
Nice early check on allowed n key offsets per tx based on the guaranteed ceiling for the entire chain.

### DIFF
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -99,6 +99,27 @@ namespace cryptonote
     const uint64_t bp_clawback = (bp_base * n_padded_outputs - bp_size) * 4 / 5;
     return bp_clawback;
   }
+
+  std::size_t max_total_key_offsets()
+  {
+    // This is a 100% guaranteed ceiling for the entire chain
+    return get_max_tx_size() / sizeof(crypto::public_key);
+  }
+
+  bool n_key_offsets_exceeds_max(const transaction_prefix& tx)
+  {
+    const std::size_t max_allowed = max_total_key_offsets();
+    std::size_t total_key_offsets = 0;
+    for (const auto &vin : tx.vin)
+    {
+      if (vin.type() != typeid(cryptonote::txin_to_key))
+        continue;
+      const std::size_t n_key_offsets = boost::get<cryptonote::txin_to_key>(vin).key_offsets.size();
+      CHECK_AND_ASSERT_MES((n_key_offsets + total_key_offsets) >= total_key_offsets, true, "key offsets overflow");
+      total_key_offsets += n_key_offsets;
+    }
+    return total_key_offsets >= max_allowed;
+  }
   //---------------------------------------------------------------
 }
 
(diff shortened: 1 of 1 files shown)


jpk68 (approved):

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -99,6 +99,27 @@ namespace cryptonote
...
+  bool n_key_offsets_exceeds_max(const transaction_prefix& tx)
+  {
+    const std::size_t max_allowed = max_total_key_offsets();
+    std::size_t total_key_offsets = 0;
+    for (const auto &vin : tx.vin)
+    {
+      if (vin.type() != typeid(cryptonote::txin_to_key))
+        continue;
+      const std::size_t n_key_offsets = boost::get<cryptonote::txin_to_key>(vin).key_offsets.size();
+      CHECK_AND_ASSERT_MES((n_key_offsets + total_key_offsets) >= total_key_offsets, true, "key offsets overflow");
```
SNeedlewoods, line 118:
``` suggestion
      CHECK_AND_ASSERT_MES((n_key_offsets + total_key_offsets) >= max_allowed, true, "key offsets overflow");
```
I think you meant this, right?

j-berman: No it's protecting overflow as is

SNeedlewoods (approved):

thomasbuilds (approved):

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -99,6 +99,27 @@ namespace cryptonote
     const uint64_t bp_clawback = (bp_base * n_padded_outputs - bp_size) * 4 / 5;
     return bp_clawback;
   }
+
+  std::size_t max_total_key_offsets()
+  {
+    // This is a 100% guaranteed ceiling for the entire chain
+    return get_max_tx_size() / sizeof(crypto::public_key);
```
jeffro256, line 106: Max observed total ring members in any one transaction on the main chain is 4524 in f901cb282db0104f33d884fc01f751cd0846a3c20dc35c37ed47f6bb1d1f8af0. Testnet is 4509 in 75afa5c44762725faef28c5d49970581962b130bb75c904f1688d0d6f0958b88. Stagenet is 4000 in e75dc5567b1734d0f346915f7f44b6a7534046b711bb35d9581dc1fdd513053e.. Unfortunately, the max total ring member size for a new, valid CLSAG tx is more than both of those since 1 16-member CLSAG input is 625 bytes, including the key image and key offsets and 0 amount, and max TX size is 1MB. 1000000/625*16 ~= 25600 ring members. So the value here could be 25600, which is tighter than 31250.

On the other hand, max observed ring size in any one ring on the main chain is 4501 in 5e75b4596c234ad17d34c44c3f07da508afb985748b3066bbfe2923cedb0b81a. Testnet is 4391 in 4056527ad2e5bc95c18a0d7d5dc911757a24f6f807eeba172839e282632580d1. Stagenet is 2000 in e75dc5567b1734d0f346915f7f44b6a7534046b711bb35d9581dc1fdd513053e. Since for new, valid v16 txs, the ring size is <= 16, we can limit the ring size to 4501.

Source: I made a blockchain utility for this: https://github.com/jeffro256/monero/commit/292727f459cf1b0f136ed63731887d28fcc3050e

|                                       | Max ring size | Max tx ring members |
|-------------------------------|---------------------|---------------------------------|
| Mainnet (observed)  | 4501                 | 4524                               |
| Testnet  (observed)   | 4391                | 4509                               |
| Stagenet (observed) | 2000                | 4000                               |
| New, valid v16           | 16                     | 25600                             |

jeffro256 (approved):

Issue #10178, opened 2025-10-24
interrupting --prune-blockchain cancels historical pruning


nahuhh:
1. Sync blockchain w/o pruning
2. Restart monerod with --prune-blockchain and --log-level 2. Observe the pruning process
3. stop monerod (ctrl c) while still pruning historical blocks
4. Repeat step 2

result = the pruning was interrupted and not continued. It appears as though new blocks are still pruned, but the pruning of old blocks is permanently cancelled

expected either:
a) not able gracefully stop monerod while pruning
b) historical pruning to continue where it left off

selsta:
#11068

PR #11068, opened 2026-08-07
blockchain_db: resume interrupted pruning

Ap4sh:
`pruning_seed` is committed at the first pruning checkpoint, so an interruption after that point makes the next startup take the incremental path and skip the unfinished historical scan

this stores the last processed transaction hash in the same LMDB transaction as each checkpoint, resumes after it on the next run, and clears the marker with the final commit. resumed operations also process newly matured tip records before continuing the historical scan

this protects pruning runs interrupted after this change. databases already interrupted by an older binary have no progress marker and can't be distinguished from completed databases without a full historical scan; the existing runtime `prune_blockchain` command remains the recovery path for those databases

related to #10178

tests:
- full unit suite with GCC 14.2
- `BlockchainDBTest/0.ResumeInterruptedPruning`
- verified the regression test fails if resume falls back to `MDB_FIRST`

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -2072,8 +2072,29 @@ bool BlockchainLMDB::prune_worker(int mode, uint32_t pruning_seed)
     throw0(DB_ERROR(lmdb_error("Failed to retrieve or create pruning seed: ", result).c_str()));
   }
 
+  MDB_val_str(pruning_progress_key, "pruning_progress");
+  crypto::hash pruning_progress{};
+  bool resume_pruning = false;
+  if (mode != prune_mode_check)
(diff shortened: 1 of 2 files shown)


jpk68 (approved):
These changes look fine, from what I can tell.

selsta (commented):
Could you simplify the LMDB test setup helpers a bit by reusing `mdb_txn_safe`?

It would also be useful to verify that resuming actually skips records before the saved checkpoint. Currently restarting the historical scan from the beginning would still pass.

Ap4sh:
done, the setup now uses `mdb_txn_safe` and the test leaves a pre-checkpoint record in place to make sure the resume path doesn't scan it again

selsta (approved):

jpk68 (approved):

PR #11069, opened 2026-08-08
src, epee: decouple net client users from http_client header

jpk68:
Rebased continuation of #8200. This patch attempts to reduce the complexity of the `http_client` class, which is derived from `epee::net_utils::http::abstract_http_client` and can be pointed at using a pointer to the base class.

This should reduce the compilation time of the whole codebase by ~10% (according to the original patch).


selsta (approved):

jpk68:
Rebased.

PR #11070, opened 2026-08-08
simplewallet: add clear command

thomasbuilds:
Clears the console screen, as one would expect from a shell. tools::clear_screen() already existed, but its only caller was the wallet lock screen; this exposes it as a command.

PAUSE_READLINE() is used because readline substitutes its own streambuf for std::cout's while running, so without it the escape sequences would be routed through readline_buffer::sync() and interleaved with prompt redraws.


jpk68 (approved):
LGTM, thanks for adding this!

SNeedlewoods (approved):

selsta (approved):

thomasbuilds:
Can we have this in v0.19 branch ?

PR #11071, opened 2026-08-09
zmq: don't log eagain as an error in relay_to_pub

jpk68:
Since `ZMQ_DONTWAIT` is being used, `zmq_msg_recv` returns `EAGAIN` if no messages are available yet. This is normal for non-blocking reads and should not be logged as an error.

### DIFF
--- a/src/rpc/zmq_pub.cpp
+++ b/src/rpc/zmq_pub.cpp
@@ -491,7 +491,8 @@ bool zmq_pub::relay_to_pub(void* const relay, void* const pub)
   const expect<bool> relayed = relay_block_pub(relay, pub);
   if (!relayed)
   {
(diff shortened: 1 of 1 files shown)


selsta (approved):

vtnerd (commented):
This was chosen deliberately and was not an oversight. Notifications are being dropped if this occurs, and it means the pub/rpc thread is overloaded. No one has ever complained about this spamming the logs because it's a rare event that should likely be logged.

jpk68:
Makes sense, closing.

vtnerd (commented):
Got my functions mixed up, but this still isn't needed. The only caller checks for revents before reading, and the writer should only fail if the pub socket is overloaded, which is a big issue. Either way it was definitely intentional.

PR #11073, opened 2026-08-09
wallet: respect --no-dns for OpenAlias

### LINKED ISSUE #8411: Respect `--no-dns` when using OpenAlias
Currently `--no-dns` gets ignored when transacting to an OpenAlias domain.

https://github.com/monero-project/monero/pull/8408#issuecomment-1169251204

Alternatively there could be a warning when sending to OpenAlias addresses when `--no-dns` is specified.

Ap4sh:
fixes #8411

`--no-dns` currently only updates `wallet2::m_use_dns`, while OpenAlias parsing never reads that state, so wallet CLI and RPC commands can still make DNS requests

this keeps the existing parser API intact, adds an explicit DNS gate, and passes the wallet DNS state through every CLI and RPC address-or-URL path. literal addresses still work with DNS disabled, and offline mode uses the same gate since it already promises not to use DNS

tests:
- `cmake --build build-system --target unit_tests monero-wallet-cli monero-wallet-rpc test_notifier -- -j4`
- `unit_tests --gtest_filter='AddressFromURL.Disabled'`
- full unit suite: 1293 passed, 2 hardware tests skipped

### DIFF
--- a/src/cryptonote_basic/cryptonote_basic_impl.cpp
+++ b/src/cryptonote_basic/cryptonote_basic_impl.cpp
@@ -237,8 +237,21 @@ namespace cryptonote {
     , std::function<std::string(const std::string&, const std::vector<std::string>&, bool)> dns_confirm
     )
   {
+    return get_account_address_from_str_or_url(info, nettype, str_or_url, true, dns_confirm);
+  }
+  //--------------------------------------------------------------------------------
+  bool get_account_address_from_str_or_url(
+      address_parse_info& info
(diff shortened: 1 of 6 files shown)


jpk68 (approved):

SNeedlewoods (approved):

selsta (approved):

selsta:
@Ap4sh there is similar behaviour in monero-wallet-rpc in case you want to open another PR.

Issue #10132, opened 2025-09-30
Disable libunwind exceptions on release builds


selsta:
> IMHO, it is not really reasonable to just ignore these. The RandomX hugepages request alone dumps the stack every 5 minutes. That's enough to make the log file unreadable. The icing on the cake is that when build type is `Release` (the default), there are no debug symbols, so these stack traces are not even intelligible.
> 
> I propose that libunwind simply be disabled when build type is `Release`. The official binary releases already ship without it, so this would be consistent with those. 
>  

 _Originally posted by @ndorf in [#8790](https://github.com/monero-project/monero/issues/8790#issuecomment-3345177403)_

jeffro256:
Agreed

PR #11075, opened 2026-08-09
cmake: disable stack traces by default for release builds

ghost:
https://github.com/monero-project/monero/issues/10132 

Release builds currently enable exception stack traces by default despite lacking useful debug symbols, producing noisy and unintelligible output. This change disables them by default for Release configurations while preserving explicit -DSTACK_TRACE=ON overrides; I’d appreciate any corrections if I have misunderstood the intended behavior or scope.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -506,7 +506,10 @@ endif()
 # Can't install hook in static build on OSX, because OSX linker does not support --wrap
 # On ARM, having libunwind package (with .so's only) installed breaks static link.
 # When possible, avoid stack tracing using libunwind in favor of using easylogging++.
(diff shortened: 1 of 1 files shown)


jeffro256:
ACK concept, but should be at the top of the preceding `if` tree

ghost:
> ACK concept, but should be at the top of the preceding `if` tree

Thanks, I moved the Release check to the top of the existing if tree. I kept the AND NOT STACK_TRACE guard so an explicit -DSTACK_TRACE=ON still uses the existing backend-selection logic.

selsta (approved):

jeffro256 (approved):

jpk68 (approved):

PR #11076, opened 2026-08-10
wallet: add pending_tx sanity checker

UkoeHB:
`pending_tx` has a lot of internal duplication and repeats info from the wallet.

@jeffro256 I put this in `tx_builder.*` since it seems appropriate, but feel free to suggest an alternative with less merge conflicts.

### DIFF
--- a/src/wallet/api/wallet.cpp
+++ b/src/wallet/api/wallet.cpp
@@ -2515,6 +2515,9 @@ void WalletImpl::pendingTxPostProcess(PendingTransactionImpl * pending)
   tools::wallet2::signed_tx_set exported_txs;
   std::vector<cryptonote::address_parse_info> dsts_info;
 
+  // NOTE: We expect `cold_sign_tx` to validate `pending->m_pending_tx` with `sanity_check_pending_tx`.
+  // It is not possible to pre-validate here because the pending tx may be 'half-formed' at this point (e.g.
+  // trezor makes a tx proposal with no key images and the cold wallet has to supply those).
   m_wallet->cold_sign_tx(pending->m_pending_tx, exported_txs, dsts_info, pending->m_tx_device_aux);
   pending->m_key_images = exported_txs.key_images;
   pending->m_pending_tx = exported_txs.ptx;
--- a/src/wallet/pending_tx_validation.cpp
+++ b/src/wallet/pending_tx_validation.cpp
@@ -1,0 +1,770 @@
+// Copyright (c) 2025-2026, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+//paired header
+#include "pending_tx_validation.h"
+
+//local headers
+#include "misc_log_ex.h"
+#include "common/apply_permutation.h"
+#include "crypto/crypto.h"
+#include "crypto/generators.h"
+#include "cryptonote_basic/account.h"
+#include "cryptonote_basic/cryptonote_basic_impl.h"
+#include "cryptonote_basic/cryptonote_format_utils.h"
+#include "cryptonote_basic/subaddress_index.h"
+#include "cryptonote_core/cryptonote_tx_utils.h"
+#include "cryptonote_config.h"
+#include "device/device_default.hpp"
+#include "ringct/rctOps.h"
+#include "ringct/rctTypes.h"
+#include "string_tools.h"
+#include "wallet2.h"
+#include "wallet2_basic/wallet2_types.h"
+
+//third party headers
+#include <boost/algorithm/string/join.hpp>
+#include <boost/multiprecision/cpp_int.hpp>
+
+//standard headers
+#include <algorithm>
+#include <cstdint>
+#include <string>
+#include <vector>
+
+#undef MONERO_DEFAULT_LOG_CATEGORY
+#define MONERO_DEFAULT_LOG_CATEGORY "wallet.tx_builder"
+
+namespace tools
+{
+namespace wallet
+{
+//-------------------------------------------------------------------------------------------------------------------
+static void validate_tx_outs(
+    const wallet2::pending_tx &ptx,
+    const std::vector<cryptonote::txout_to_tagged_key> &ext_outputs,
+    const crypto::secret_key &k_view,
+    const cryptonote::tx_destination_entry &change,
+    const std::vector<cryptonote::tx_destination_entry> &dests,
+    const crypto::secret_key &tx_key,
+    const std::vector<crypto::secret_key> &additional_tx_keys,
+    crypto::public_key &tx_pubkey_out,
+    std::vector<crypto::public_key> &tx_additional_pubkeys_out)
+{
+    tx_pubkey_out = crypto::public_key{};
+    tx_additional_pubkeys_out.clear();
+    CHECK_AND_ASSERT_THROW_MES(ptx.tx.vout.size() == dests.size(),
+        "validate_tx_outs: dests size mismatch");
+    CHECK_AND_ASSERT_THROW_MES(ext_outputs.size() == dests.size(),
+        "validate_tx_outs: ext_outputs size mismatch");
+    CHECK_AND_ASSERT_THROW_MES(ptx.tx.version == 1 || ptx.tx.rct_signatures.ecdhInfo.size() == dests.size(),
+        "validate_tx_outs: ecdhInfo size mismatch");
+    CHECK_AND_ASSERT_THROW_MES(ptx.tx.version == 1 || ptx.tx.rct_signatures.outPk.size() == dests.size(),
+        "validate_tx_outs: outPk size mismatch");
+
+    // Check if multiple keys are expected
+    // Note: this is the same method used in `construct_tx_and_get_tx_key()`, so we should expect the same results.
+    size_t num_stdaddresses = 0;
+    size_t num_subaddresses = 0;
+    cryptonote::account_public_address single_dest_subaddress;
+    cryptonote::classify_addresses(dests, change.addr, num_stdaddresses, num_subaddresses, single_dest_subaddress);
+    const bool need_additional_txkeys = num_subaddresses > 0 && (num_stdaddresses > 0 || num_subaddresses > 1);
+
+    const size_t expect_num_additional = need_additional_txkeys ? dests.size() : 0;
+    CHECK_AND_ASSERT_THROW_MES(additional_tx_keys.size() == expect_num_additional,
+        "validate_tx_outs: additional tx keys invalid size");
+
+    // if this is a single-destination transfer to a subaddress, we set the tx pubkey to R=s*D
+    if (num_stdaddresses == 0 && num_subaddresses == 1)
+    {
+        tx_pubkey_out = rct::rct2pk(rct::scalarmultKey(
+                rct::pk2rct(single_dest_subaddress.m_spend_public_key),
+                rct::sk2rct(tx_key)
+            ));
+    }
+    else
+    {
+        tx_pubkey_out = rct::rct2pk(rct::scalarmultBase(rct::sk2rct(tx_key)));
+    }
+
+    // It's assumed here generate_output_ephemeral_keys only needs the private view key
+    cryptonote::account_keys sender_keys;
+    sender_keys.m_view_secret_key = k_view;
+
+    std::vector<rct::key> amount_keys;
+    for (size_t i = 0; i < dests.size(); ++i)
+    {
+        const auto &dest = dests.at(i);
+
+        crypto::public_key repro_onetime_addr{};
+        crypto::view_tag repro_view_tag{};
+        const bool r = hw::core::device_default().generate_output_ephemeral_keys(
+                ptx.tx.version,
+                sender_keys,
+                tx_pubkey_out,
+                tx_key,
+                dest,
+                change.addr,
+                i,
+                need_additional_txkeys,
+                additional_tx_keys,
+                tx_additional_pubkeys_out,
+                amount_keys,
+                repro_onetime_addr,
+                ptx.construction_data.use_view_tags,
+                repro_view_tag
+            );
+        CHECK_AND_ASSERT_THROW_MES(r, "validate_tx_outs: failed to generate output ephemeral key");
+
+        // - Ko
+        CHECK_AND_ASSERT_THROW_MES(repro_onetime_addr == ext_outputs.at(i).key,
+            "validate_tx_outs: failed reproducing onetime address");
+
+        // - view tag
+        CHECK_AND_ASSERT_THROW_MES(repro_view_tag == ext_outputs.at(i).view_tag,
+            "validate_tx_outs: failed reproducing view tag");
+
+        if (ptx.tx.version == 1)
+        {
+            // - cleartext amount
+            CHECK_AND_ASSERT_THROW_MES(ptx.tx.vout.at(i).amount == dest.amount,
+                "validate_tx_outs: v1 tx output amount doesn't match destination amount");
+
+            continue;
+        }
+
+        // - encoded amount
+        rct::ecdhTuple amnt_data{};
+        memcpy(amnt_data.amount.bytes, &dest.amount, sizeof(dest.amount));
+        rct::ecdhDecode(amnt_data, amount_keys.at(i), true); //v2, decode gives mask
+        CHECK_AND_ASSERT_THROW_MES(ptx.tx.rct_signatures.ecdhInfo.at(i).amount == amnt_data.amount,
+            "validate_tx_outs: failed reproducing encoded amount");
+
+        // - C
+        const rct::key repro_C = rct::commit(dest.amount, amnt_data.mask);
+        CHECK_AND_ASSERT_THROW_MES(ptx.tx.rct_signatures.outPk.at(i).mask == repro_C,
+            "validate_tx_outs: failed reproducing amount commitment");
+    }
+}
+//-------------------------------------------------------------------------------------------------------------------
+static void reconstruct_payment_id(const std::string &extracted_payment_id,
+    const std::optional<crypto::hash8> &parsed_to_check,
+    // const std::optional<crypto::hash8> &pre_decrypted_to_check,
+    const cryptonote::account_public_address &change_addr,
+    const std::vector<cryptonote::tx_destination_entry> &dests,
+    // Will be std::nullopt if tx_key is redacted.
+    const std::optional<crypto::secret_key> tx_key,
+    const std::vector<cryptonote::tx_extra_field> &tx_extra_fields)
+{
+    // Get the destination that will be able to read the payment id in the final tx.
+    // Returns non-null if there is exactly one destination (may or may not be the change addr).
+    const crypto::public_key view_key_pub = cryptonote::get_destination_view_key_pub(dests, change_addr);
+    CHECK_AND_ASSERT_THROW_MES(view_key_pub != crypto::null_pkey,
+        "reconstruct_payment_id: encrypted payment IDs must only be in txs with one destination");
+
+    // Extract the expected encrypted payment id.
+    cryptonote::tx_extra_nonce extra_nonce;
+    crypto::hash8 encrypted_payment_id8 = crypto::null_hash8;
+    CHECK_AND_ASSERT_THROW_MES(cryptonote::find_tx_extra_field_by_type(tx_extra_fields, extra_nonce),
+        "reconstruct_payment_id: expected payment id is missing");
+    CHECK_AND_ASSERT_THROW_MES(cryptonote::get_encrypted_payment_id_from_tx_extra_nonce(
+            extra_nonce.nonce,
+            encrypted_payment_id8
+        ),
+        "reconstruct_payment_id: expected payment id is missing");
+
+    // Parse the payment ID string extracted from the address.
+    crypto::hash8 to_encrypt_payment_id;
+    CHECK_AND_ASSERT_THROW_MES(tools::wallet2::parse_short_payment_id(extracted_payment_id, to_encrypt_payment_id),
+        "reconstruct_payment_id: failed parsing short payment id from integrated address");
+    if (parsed_to_check)
+    {
+        CHECK_AND_ASSERT_THROW_MES(*parsed_to_check == to_encrypt_payment_id,
+            "reconstruct_payment_id: did not extract the same parsed payment id");
+    }
+    // if (pre_decrypted_to_check)
+    // {
+    //     CHECK_AND_ASSERT_THROW_MES(*pre_decrypted_to_check == to_encrypt_payment_id,
+    //         "reconstruct_payment_id: did not extract the same pre-decrypted payment id");
+    // }
+
+    if (tx_key)
+    {
+        // Encrypt the address's payment ID.
+        CHECK_AND_ASSERT_THROW_MES(hw::core::device_default().encrypt_payment_id(to_encrypt_payment_id,
+                view_key_pub, *tx_key),
+            "reconstruct_payment_id: failed encrypting payment id");
+
+        // Check equivalence.
+        CHECK_AND_ASSERT_THROW_MES(to_encrypt_payment_id == encrypted_payment_id8,
+            "reconstruct_payment_id: failed encrypting payment id");
+    }
+}
+//-------------------------------------------------------------------------------------------------------------------
+void sanity_check_pending_tx(const wallet2::pending_tx &ptx,
+    const cryptonote::network_type nettype,
+    const cryptonote::account_keys &account_keys,
+    const std::unordered_map<crypto::public_key, cryptonote::subaddress_index> &subaddresses,
+    const std::vector<wallet2_basic::transfer_details> &transfers,
+    const bool redacted,
+    const std::optional<std::function<const crypto::key_image(const size_t)>> &transfer_ki_resolver,
+    const bool allow_read_only)
+{
+    const auto &construct = ptx.construction_data;
+    const bool transfer_ki_accessible = transfer_ki_resolver.has_value();
+
+    // Extract outputs
+    std::vector<cryptonote::txout_to_tagged_key> ext_outputs;
+    const bool all_are_txout_to_tagged_key = std::all_of(ptx.tx.vout.begin(), ptx.tx.vout.end(),
+        [&](const cryptonote::tx_out& s_e) -> bool
+        {
+            CHECKED_GET_SPECIFIC_VARIANT(s_e.target, const cryptonote::txout_to_tagged_key, out, false);
+            ext_outputs.push_back(out);
+            return true;
+        }
+    );
+    CHECK_AND_ASSERT_THROW_MES(all_are_txout_to_tagged_key,
+        "sanity_check_pending_tx: all outputs should have view tags");
+    CHECK_AND_ASSERT_THROW_MES(construct.use_view_tags,
+        "sanity_check_pending_tx: must be view-tag-enabled (hardfork >= 15)");
+    CHECK_AND_ASSERT_THROW_MES(construct.splitted_dsts.size() == ptx.tx.vout.size(),
+        "sanity_check_pending_tx: destination vecs are inconsistent");
+    CHECK_AND_ASSERT_THROW_MES(ptx.tx.unlock_time == 0 && construct.unlock_time == 0,
+        "sanity_check_pending_tx: unlock_time must be zero");
+
+    // Extra
+    std::vector<cryptonote::tx_extra_field> tx_extra_fields;
+    CHECK_AND_ASSERT_THROW_MES(cryptonote::parse_tx_extra(ptx.tx.extra, tx_extra_fields),
+        "sanity_check_pending_tx: ptx.tx.extra extraction failure");
+    // NOTE: Due to upstream chaos, we are unable to reliably validate `construct.extra`.
+    // if (!cleartext_payment_id)
+    // {
+    //     CHECK_AND_ASSERT_THROW_MES(construct.extra == ptx.tx.extra,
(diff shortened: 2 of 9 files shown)


jeffro256:
Maybe this would be a good place to move `pending_tx` into `tx_builder`, to prevent future code churn with fcmp++-stage, and get some of the refactors out of way now

src/wallet/tx_builder.h
```diff
@@ -0,0 +1,53 @@
...
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+//local headers
+#include "cryptonote_config.h"
+#include "wallet2.h"
```
jeffro256, line 36: Yeah one of the main reasons I split the translation units is so that they *don't* depend on `wallet2.h`. If we move `pending_tx` here, then we can remove this include

UkoeHB: Ideally this PR goes in the release and we can do the refactor later.

j-berman (changes requested):
Suggested changes over [here](https://github.com/UkoeHB/monero/pull/52), can just merge that PR if those changes look good. Also added a couple more tests for cold signing.

I think this general sanity checking code should for the most part be reusable with FCMP++/Carrot changes, but worth keeping in mind that code will also knock out classes of issues this code solves.

I'd approve this PR with my suggested changes applied.

j-berman (approved):

UkoeHB:
Latest commit fixes:
- `m_global_output_index` is an index into same-amount enotes, so it is not sufficient by itself to validate pre-RCT inputs.
- `additional_tx_keys` size check if `need_additional_txkeys`.
- `parse_tx_from_str` should validate the `signed_tx_set` key images against `ptx` before importing. This isn't perfect but shrinks the error boundary of wallet state.
- Update `parse_multisig_tx_from_str` to use the new sanity checker.

src/wallet/wallet2.cpp
```diff
@@ -8254,8 +8272,6 @@ bool wallet2::load_multisig_tx(cryptonote::blobdata s, multisig_tx_set &exported
   {
     for (const auto &ptx: exported_txs.m_ptx)
     {
-      this->sanity_check_pending_tx(ptx, false, false);
-
       const crypto::hash txid = get_transaction_hash(ptx.tx);
```
UkoeHB, line 8325: Moved into `parse_multisig_tx_from_str`

UkoeHB:
Latest:
- trezor signing redacts txs, so cold signer has to act as if all txs are redacted.
- Bugfix: redacted txs cannot reconstruct payment ids.
- Bugfix: checker needs to make sure there are > 0 inputs. Callers may index into the first input without checking.
- Bugfix: dust + fee could overflow.
- Bugfix: improve payment id extraction in case of a URI that can be parsed but which only has a payment ID embedded in the address but not the URI structure.

UkoeHB:
Pushed to consolidate compile fix commits.

src/wallet/pending_tx_validation.cpp
```diff
@@ -0,0 +1,655 @@
...
+    if (ptx.change_dts.amount > 0 || ptx.dests.size() == 1)
+    {
+        splitted_dsts_repro.push_back(ptx.change_dts);
+    }
+    CHECK_AND_ASSERT_THROW_MES(construct.splitted_dsts.size() == splitted_dsts_repro.size(),
+        "sanity_check_pending_tx: failed checking splitted_dsts size");
+    CHECK_AND_ASSERT_THROW_MES(ptx.tx.rct_signatures.ecdhInfo.size() == splitted_dsts_repro.size(),
+        "sanity_check_pending_tx: ecdhInfo size mismatch");
+    CHECK_AND_ASSERT_THROW_MES(ptx.tx.rct_signatures.outPk.size() == splitted_dsts_repro.size(),
+        "sanity_check_pending_tx: outPk size mismatch");
```
j-berman: These need to only run for tx version > 1

src/wallet/pending_tx_validation.cpp
```diff
@@ -0,0 +1,655 @@
...
+    }
+    CHECK_AND_ASSERT_THROW_MES(splitted_dsts_repro.size() == 0,
+        "sanity_check_pending_tx: failed checking splitted_dsts consistency");
+
+    // Balance check
+    output_amnt += (ptx.dust_added_to_fee ? 0 : ptx.dust);
+    output_amnt += ptx.fee;
+
+    CHECK_AND_ASSERT_THROW_MES(ptx.tx.rct_signatures.txnFee == ptx.fee,
+            "sanity_check_pending_tx: inconsistent fee");
```
j-berman: `tx.rct_signatures.txnFee` is tx v2, but v1 txs are still possible (the sweep unmixable case). v1 tx fee is calculated as `get_inputs_money_amount(tx) - get_outs_money_amount(tx)`, see: https://github.com/monero-project/monero/blob/0aa9c51bd8eb53abb0a7023b5c768d9b46ee18a0/src/cryptonote_core/blockchain.cpp#L1688-L1693

j-berman (commented):
Couple more issues for v1 txs

src/wallet/pending_tx_validation.cpp
UkoeHB: Are pre-rct input amounts recorded explicitly in txs? Right now we have:
```cpp
        CHECK_AND_ASSERT_THROW_MES(src.rct
            ? ext_input_amount == 0
            : ext_input_amount == src.amount,
            "sanity_check_pending_tx: amount extracted from inputs is wrong");
```
But maybe the check should be `ptx.tx.version > 1` instead of `src.rct`.

j-berman: Yep, you're right. Good spot

j-berman (approved):

src/wallet/pending_tx_validation.cpp
j-berman: Actually spoke too soon. Yes pre-RCT input amounts are recorded explicitly in txs and that original logic was ok. Here's an example tx spending pre-RCT inputs -> RCT outputs: https://xmrchain.net/tx/c22dddbc73076825b163f048122f759cfd983fff9a4bcabefe9f342e7d12017f/1 (EDIT: linked the wrong tx)

It's fine this way too though.

j-berman: Err, no. New change is actually incorrect because of that. It does need to be:

```
        CHECK_AND_ASSERT_THROW_MES(src.rct
            ? ext_input_amount == 0
            : ext_input_amount == src.amount,
            "sanity_check_pending_tx: amount extracted from inputs is wrong");
```

src/wallet/pending_tx_validation.cpp
```diff
@@ -0,0 +1,666 @@
...
+
+        input_amnt += src.amount;
+        // We need `inp_order` from here, which is only valid if key images are available.
+        if (!transfer_ki_accessible)
+            continue;
+
+        const uint64_t ext_input_amount = ext_input_amounts.at(inp_order.at(i));
+        const std::vector<uint64_t> &ext_relative_key_offsets = ext_key_offsets.at(inp_order.at(i));
+
+        CHECK_AND_ASSERT_THROW_MES(ptx.tx.version > 1
```
j-berman: needs to be reverted to `src.rct`

UkoeHB: Wouldn't that contradict `collect_transparent_amount_commitments_static` which says transparent amounts are only v1 or coinbase?

j-berman: That's collecting from the output side, this is the input side

j-berman (approved):

UkoeHB:
Squashed. Idk if you care about co-author attribution @j-berman, haven't looked into how to add that.

EDIT: added attribution in latest, sorry for spam

src/wallet/pending_tx_validation.cpp
```diff
@@ -0,0 +1,683 @@
...
+            "sanity_check_pending_tx: Non-zero amount change address is not recognized as belonging to the sender account");
+    }
+
+    // Destination consistency
+    CHECK_AND_ASSERT_THROW_MES(construct.change_dts == ptx.change_dts,
+        "sanity_check_pending_tx: change_dts inconsistent");
+    CHECK_AND_ASSERT_THROW_MES(construct.dests == ptx.dests,
+        "sanity_check_pending_tx: destination vecs are inconsistent");
+    std::vector<cryptonote::tx_destination_entry> splitted_dsts_repro = ptx.dests;
+    if (ptx.change_dts.amount > 0 || ptx.dests.size() == 1)
```
thomasbuilds: This rejects every sweep_unmixable tx. That path still builds v1 (fake_outs_count = 0, wallet2.cpp:11203) via transfer_selected, which digit-splits dests and adds no change entry when change is 0. Reproduced on this branch with a genuine v1 tx: one dest fails here with or without the split; a two dest v1 tx without the split passes.

thomasbuilds (commented):
parse_tx_from_str now throws and WalletImpl::submitTransaction calls load_tx outside any try.

src/wallet/pending_tx_validation.cpp
UkoeHB: In that scenario would `splitted_dests.size() == 1`? Or is it adding a dummy output somewhere? If so, where?

thomasbuilds: There's no dummy output on the v1 path, that's only in `transfer_selected_rct` (9976-9989). `transfer_selected` digit-splits the destination (wallet2.h:1780) and only adds change if `needed_money < found_money` (9802-9806) so for a sweep `splitted_dsts.size()` is the number of nonzero decimal digits of the amount, e.g. 4 for 1.234 XMR. Line 526 assumes the rct rule and expects 2. If the amount happens to have exactly two digits the size check passes and 639-660 catches it instead.

UkoeHB: Ok for v1 looks like we need to iter splitted dests and accumulate them, then match amounts/keys with dests + change_dest (which also need to be accumulated in case of duplicates). And don't check splitted_dest_repro.

UkoeHB: @thomasbuilds does latest commit fix your test case?

src/wallet/pending_tx_validation.cpp
```diff
@@ -522,15 +522,62 @@ void sanity_check_pending_tx(const wallet2::pending_tx &ptx,
...
+        // We iter splitted dests and accumulate them, then match amounts/keys with dests + change_dest
+        // (which also need to be accumulated in case of duplicates).
+        // The map keys use `account_public_address` instead of the spendkey in case of malicous construction.
+
+        // Accumulate splitted dests
+        std::unordered_map<cryptonote::account_public_address, boost::multiprecision::uint128_t> acc_split;
+        for (const auto &split : construct.splitted_dsts)
+        {
+            auto res = acc_split.try_emplace(split.addr, 0);
+            auto amnt = res.first->second;
```
thomasbuilds:
```suggestion
            auto &amnt = res.first->second;
```

src/wallet/pending_tx_validation.cpp
```diff
@@ -522,15 +522,62 @@ void sanity_check_pending_tx(const wallet2::pending_tx &ptx,
...
+
+        // Accumulate dests
+        std::unordered_map<cryptonote::account_public_address, boost::multiprecision::uint128_t> acc_dest;
+        if (ptx.change_dts.amount > 0)
+            acc_dest.emplace(ptx.change_dts.addr, ptx.change_dts.amount);
+
+        for (const auto &dest : ptx.dests)
+        {
+            auto res = acc_split.try_emplace(dest.addr, 0);
+            auto amnt = res.first->second;
```
thomasbuilds:
```suggestion
            auto res = acc_dest.try_emplace(dest.addr, 0);
            auto &amnt = res.first->second;
```

src/wallet/pending_tx_validation.cpp
```diff
@@ -522,15 +522,62 @@ void sanity_check_pending_tx(const wallet2::pending_tx &ptx,
...
+
+        // Consistency check
+        CHECK_AND_ASSERT_THROW_MES(acc_split.size() == acc_dest.size(),
+            "sanity_check_pending_tx: v1 dest inconsistency (size)");
+        for (const auto &split : acc_split)
+        {
+            const auto &dest_res = acc_dest.find(split.first);
+            CHECK_AND_ASSERT_THROW_MES(dest_res != acc_dest.cend(),
+                "sanity_check_pending_tx: v1 dest inconsistency (addr matching)");
+            CHECK_AND_ASSERT_THROW_MES(split.second != dest_res->second,
```
thomasbuilds:
```suggestion
            CHECK_AND_ASSERT_THROW_MES(split.second == dest_res->second,
```

thomasbuilds (changes requested):

src/wallet/pending_tx_validation.cpp
thomasbuilds: No fails with `v1 dest inconsistency (size)`. Check out my suggestions please, which make my test case pass, and the tampered control is still rejected.

UkoeHB: Thanks, fixed in latest

src/wallet/pending_tx_validation.cpp
```diff
@@ -0,0 +1,728 @@
...
+        if (ptx.tx.version == 1)
+        {
+            CHECK_AND_ASSERT_THROW_MES(!src.rct,
+                "sanity_check_pending_tx: rct sources not allowed for v1 txs");
+        }
+        CHECK_AND_ASSERT_THROW_MES(src.amount == transfer.m_amount,
+            "sanity_check_pending_tx: transfer amount mismatch");
+        CHECK_AND_ASSERT_THROW_MES(src.real_output < src.outputs.size(),
+            "sanity_check_pending_tx: ring sig index out of input set size");
+        const rct::key commitment = rct::commit(src.amount, src.mask);
```
thomasbuilds:
```suggestion
        const rct::key commitment = rct::commit(src.amount, src.rct ? src.mask : rct::identity());
```

seems that `transfer_selected` never sets `src.mask` for v1 (only `transfer_selected_rct` does, 9961) so it stays zero here and a real sweep is still rejected, including through `parse_tx_from_str`. Scanned pre-RCT outputs always have `m_mask = identity` (wallet2.cpp:2616).

UkoeHB: Fixed in latest

src/wallet/wallet2.cpp
```diff
@@ -8188,15 +8252,9 @@ bool wallet2::parse_multisig_tx_from_str(std::string multisig_tx_st, multisig_tx
...
-      CHECK_AND_ASSERT_MES(idx < m_transfers.size(), false, "Transfer index out of range");
-    CHECK_AND_ASSERT_MES(ptx.construction_data.selected_transfers.size() == ptx.tx.vin.size(), false, "Mismatched cd selected_transfers/vin sizes");
-    for (size_t idx: ptx.construction_data.selected_transfers)
-      CHECK_AND_ASSERT_MES(idx < m_transfers.size(), false, "Transfer index out of range");
-    CHECK_AND_ASSERT_MES(ptx.construction_data.sources.size() == ptx.tx.vin.size(), false, "Mismatched sources/vin sizes");
-    CHECK_AND_ASSERT_MES(!ptx.tx.vin.empty(), false, "Multisig tx has no inputs");
-    CHECK_AND_ASSERT_MES(!ptx.construction_data.sources.empty(), false, "Multisig tx has no sources");
+    // If key images have not been imported then this could fail if we check key images. We'd rather fail elsewhere
+    // with a better error message.
+    this->sanity_check_pending_tx(ptx, false, false, std::nullopt);
```
thomasbuilds: This used to return false and now throws and the checker also requires `!m_spent && !m_frozen`, so read-only loads of an already-submitted or frozen-input txset (`describe_transfer`, `export_raw_multisig`, re-load) now fail. Keeping the bool contract also fixes the unguarded RPC callers (see review body):

```suggestion
    try
    {
      this->sanity_check_pending_tx(ptx, false, false, std::nullopt);
    }
    catch (const std::exception &e)
    {
      LOG_PRINT_L0("Multisig tx failed sanity check: " << e.what());
      return false;
    }
```

src/wallet/wallet2.cpp
```diff
@@ -8365,6 +8427,10 @@ bool wallet2::sign_multisig_tx(multisig_tx_set &exported_txs_inout, std::vector<
       }
     }
 
+    // Now that get_multisig_k has run for all sigs and not thrown 'need to export multisig', we can validate
+    // the pending tx with expected imported key images known.
+    this->sanity_check_pending_tx(ptx, false, expect_imported_key_images, std::nullopt);
```
thomasbuilds: `get_multisig_k` passing only proves `export_multisig` ran (it reads `m_multisig_k`, populated there). `m_key_image_known` is only set by multisig import (`update_multisig_rescan_info`); scan sets it false for multisig (2578). So a cosigner who exported but whose last import doesn't cover these outputs could sign before and now throws `KI is expected but unknown`. multisig.py always imports before signing, so CI won't catch it. If requiring import to sign is intended, the error should say to run `import_multisig_info`; otherwise:

```suggestion
    for (const size_t idx : sd.selected_transfers)
      expect_imported_key_images = expect_imported_key_images && idx < m_transfers.size() && m_transfers[idx].m_key_image_known;
    this->sanity_check_pending_tx(ptx, false, expect_imported_key_images, std::nullopt);
```

src/wallet/pending_tx_validation.cpp
```diff
@@ -0,0 +1,735 @@
...
+    const bool all_are_txout_to_tagged_key = std::all_of(ptx.tx.vout.begin(), ptx.tx.vout.end(),
+        [&](const cryptonote::tx_out& s_e) -> bool
+        {
+            CHECKED_GET_SPECIFIC_VARIANT(s_e.target, const cryptonote::txout_to_tagged_key, out, false);
+            ext_outputs.push_back(out);
+            return true;
+        }
+    );
+    CHECK_AND_ASSERT_THROW_MES(all_are_txout_to_tagged_key,
+        "sanity_check_pending_tx: all outputs should have view tags");
```
thomasbuilds, line 252: Unconditional, but pre-HF15 txs have `txout_to_key` outs, so this rejects valid txsets on chains below HF15 and `\005` files saved before HF15 (format exists since 2020-06 and is still the only accepted version). If dropping those is intended, better make it explicit; this also closes the uninitialized `repro_view_tag` read (other comment) since `use_view_tags` is a file-controlled flag:

```suggestion
    CHECK_AND_ASSERT_THROW_MES(all_are_txout_to_tagged_key,
        "sanity_check_pending_tx: all outputs should have view tags");
    CHECK_AND_ASSERT_THROW_MES(construct.use_view_tags,
        "sanity_check_pending_tx: only view-tag (>= HF15) pending txs can be validated");
```

src/wallet/pending_tx_validation.cpp
```diff
@@ -0,0 +1,735 @@
...
+    cryptonote::account_keys sender_keys;
+    sender_keys.m_view_secret_key = k_view;
+
+    std::vector<rct::key> amount_keys;
+    for (size_t i = 0; i < dests.size(); ++i)
+    {
+        const auto &dest = dests.at(i);
+
+        crypto::public_key repro_onetime_addr;
+        crypto::view_tag repro_view_tag;
```
thomasbuilds: Stays uninitialized when `construct.use_view_tags` is false (the device only derives the tag under that flag) while line 149 compares it unconditionally, so tagged vouts with the serialized flag cleared compare an indeterminate byte:

```suggestion
        crypto::view_tag repro_view_tag{};
```

src/wallet/pending_tx_validation.cpp
```diff
@@ -0,0 +1,735 @@
...
+
+    boost::multiprecision::uint128_t output_amnt = 0;
+    size_t integrated_count = 0;
+
+    for (size_t i = 0; i < construct.splitted_dsts.size(); ++i)
+    {
+        const auto &dest = construct.splitted_dsts.at(i);
+
+        // Check original dst string
+        if (dest.original.size() > 0)
```
thomasbuilds, line 613: `integrated_count` and all payment id checks only run when `original` is non-empty, so `is_integrated = true` with empty `original` skips them and an encrypted payment id can be swapped undetected. All producers set `original` for integrated dests, so requiring it is safe:

```suggestion
        CHECK_AND_ASSERT_THROW_MES(!dest.is_integrated || dest.original.size() > 0,
            "sanity_check_pending_tx: integrated destination missing original address string");

        if (dest.original.size() > 0)
```

An `original` that fails both URI and address parsing still skips the pid binding; maybe also require it to parse when `is_integrated`.

src/wallet/wallet2.cpp
```diff
@@ -8065,13 +8122,17 @@ bool wallet2::parse_tx_from_str(const std::string &signed_tx_st, std::vector<too
...
+  bool r = this->import_key_images(signed_txs.key_images);
   if (!r) return false;
 
   // remember key images for this tx, for when we get those txes from the blockchain
   for (const auto &e: signed_txs.tx_key_images)
     m_cold_key_images.insert(e);
 
+  // extra validation making sure key images line up
+  for (const auto &ptx : signed_txs.ptx)
+    this->sanity_check_pending_tx(ptx, true, true, std::nullopt);
```
thomasbuilds: Cannot diverge from the check at 8107: `import_key_images` copies `signed_txs.key_images` verbatim (returns false on oversize, early return above) and `kis.at(i)` already bounds every selected transfer, so this re-checks identical values, and after state was already mutated:

```suggestion
```

thomasbuilds (commented):
Please consider these points:

`WalletImpl::submitTransaction` calls `load_tx()` with no try/catch (api/wallet.cpp:1154) and the new sanity loops in `parse_tx_from_str` are outside its try blocks so failed checks now escape the wallet2_api as `std::runtime_error`. Trivial repro: submit the same signed file twice (`commit_tx` marks inputs spent, the second load throws on the `!m_spent` check). Every other api entry point reaching the checker is wrapped in try/catch, this one needs it too.

`on_sign_multisig` (wallet_rpc_server.cpp:4619) and `on_submit_multisig` (4689) call `load_multisig_tx` outside their try blocks, so clients now get `{"error":{"code":0,"message":""}}` instead of `BAD_MULTISIG_TX_DATA`. The suggestion on `parse_multisig_tx_from_str` covers this.

src/wallet/wallet2.cpp
UkoeHB: Lots of good suggestions but this one is bad. It relies on incidental logic instead of local logic which is stronger and more maintainable/robust. Note that this is a *sanity checker* which means it is allowed to check things that are redundant or trivial.

src/wallet/wallet2.cpp
UkoeHB: Added multisig import enforcement.

src/wallet/api/wallet.cpp
```diff
@@ -2512,6 +2513,9 @@ void WalletImpl::pendingTxPostProcess(PendingTransactionImpl * pending)
     return;
   }
 
+  for (const auto &ptx : pending->m_pending_tx)
+    m_wallet->sanity_check_pending_tx(ptx, false, false, std::nullopt);
+
```
thomasbuilds: Breaks Trezor via the api: the host has no spend key and `compute_key_image` is inactive during construction, so every input of the host template carries the identity key image and the duplicate key image check throws for any tx with 2+ inputs. The check at the end of `cold_sign_tx` already covers the device-signed ptx.

```suggestion
```

src/wallet/wallet2.cpp
```diff
@@ -8285,8 +8367,23 @@ bool wallet2::sign_multisig_tx(multisig_tx_set &exported_txs_inout, std::vector<
...
+    // B) signatures involving honest signers will always have imported key images, so at least one honest signer
+    // will be able to detect spends from such txs
+    for (const size_t idx : sd.selected_transfers)
+    {
+      if (idx >= m_transfers.size() || !m_transfers[idx].m_key_image_known)
+        THROW_WALLET_EXCEPTION(error::multisig_import_needed);
+    }
+
+    // Fully validate now that we have checked for the 'need to import' error.
+    this->sanity_check_pending_tx(ptx, false, true, std::nullopt);
```
thomasbuilds: This runs before `init(reconstruction = true)`, which rebuilds `ptx.tx` from construction data (`vin` from `multisig_kLRki.ki`, `extra` from `construction_data.extra`, neither checked), so it validates a tx that is then discarded and guarantee A does not hold for the signed one (B depends on it). The check has to run again once `init` returned, as 2a4ac7477 did after the sig loop.

src/wallet/api/wallet.cpp
UkoeHB: Oh sure, let's overload `pending_tx` with every damn spaghetti bs we can think of. Sigh... thanks for pointing these out.

src/wallet/wallet2.cpp
```diff
@@ -8285,8 +8367,20 @@ bool wallet2::sign_multisig_tx(multisig_tx_set &exported_txs_inout, std::vector<
   for (size_t n = 0; n < exported_txs.m_ptx.size(); ++n)
```
thomasbuilds, line 8401: The gate runs per tx after `get_multisig_k` already wiped the earlier txs' nonces, so with an unsynced input in a later tx of a multi-tx set the retry after import fails with `multisig_export_needed`. Run it over all txs first, the in-loop copy can then go:

```suggestion
  for (size_t n = 0; n < exported_txs.m_ptx.size(); ++n)
  {
    for (const size_t idx : exported_txs.m_ptx[n].construction_data.selected_transfers)
    {
      if (idx >= m_transfers.size() || !m_transfers[idx].m_key_image_known)
        THROW_WALLET_EXCEPTION(error::multisig_import_needed);
    }
  }

  for (size_t n = 0; n < exported_txs.m_ptx.size(); ++n)
```

src/wallet/wallet2.cpp
```diff
@@ -7923,6 +7959,10 @@ bool wallet2::sign_tx(unsigned_tx_set &exported_txs, std::vector<wallet2::pendin
     signed_txes.key_images[i] = m_transfers[i].m_key_image;
   }
 
+  // check the local tx copies
+  for (const auto &ptx : txs)
+    this->sanity_check_pending_tx(ptx, false, true, std::nullopt);
```
thomasbuilds: If the online wallet already has every key image imported the export is empty and `import_outputs` is skipped, so an offline wallet restored from seed throws `invalid transfers index` here where master signed from `sd.sources`:

```suggestion
  for (const auto &ptx : txs)
  {
    for (const size_t idx : ptx.selected_transfers)
      THROW_WALLET_EXCEPTION_IF(idx >= m_transfers.size(), error::wallet_internal_error,
        "No record of output " + std::to_string(idx) + " in this wallet, run export_outputs all on the online wallet and import_outputs here");
    this->sanity_check_pending_tx(ptx, false, true, std::nullopt);
  }
```

src/wallet/pending_tx_validation.cpp
```diff
@@ -0,0 +1,741 @@
...
+            ext_outputs.push_back(out);
+            return true;
+        }
+    );
+    CHECK_AND_ASSERT_THROW_MES(all_are_txout_to_tagged_key,
+        "sanity_check_pending_tx: all outputs should have view tags");
+    CHECK_AND_ASSERT_THROW_MES(construct.use_view_tags,
+        "sanity_check_pending_tx: must be view-tag-enabled (hardfork >= 15)");
+    CHECK_AND_ASSERT_THROW_MES(construct.splitted_dsts.size() == ptx.tx.vout.size(),
+        "sanity_check_pending_tx: destination vecs are inconsistent");
```
thomasbuilds, line 256: `unlock_time` is never checked, only the pool rejects nonzero values (blocks bypass it) and the offline side already checks it in `sign_tx`:

```suggestion
    CHECK_AND_ASSERT_THROW_MES(construct.splitted_dsts.size() == ptx.tx.vout.size(),
        "sanity_check_pending_tx: destination vecs are inconsistent");
    CHECK_AND_ASSERT_THROW_MES(ptx.tx.unlock_time == 0 && construct.unlock_time == 0,
        "sanity_check_pending_tx: unlock_time must be zero");
```

src/wallet/api/wallet.cpp
```diff
@@ -39,6 +39,7 @@
 #include "common_defines.h"
 #include "common/util.h"
 #include "multisig/multisig_account.h"
+#include "wallet/pending_tx_validation.h"
```
thomasbuilds:
```suggestion
```

thomasbuilds (commented):
please check those little things

src/wallet/pending_tx_validation.cpp
UkoeHB: @j-berman do we want to enforce this? Someone could make an unlock time tx and mine it themselves...

j-berman: I think this is a nit. I don't think it hurts to enforce it here, and I can kind of see it maybe helping a tiny bit if the wallet is connected to a malicious miner too? Your call.

The `monerod` code isn't set up to support mining txs with an unlock time anyway, it would need to be modified further to do that too. And I think it would be fine to expect the same kind of modification in this code if someone wants to mine a tx with custom unlock time themselves.

FWIW any txs with custom unlock time are set to be deprecated anyway: https://www.getmonero.org/2026/05/10/deprecating-unlock-time.html

So in all, I think it's definitely not an issue to enforce here.

src/wallet/wallet2.cpp
```diff
@@ -7923,6 +7959,18 @@ bool wallet2::sign_tx(unsigned_tx_set &exported_txs, std::vector<wallet2::pendin
...
   }
 
+  // check the local tx copies
+  for (const auto &ptx : txs)
+  {
+    for (const size_t idx : ptx.selected_transfers)
+    {
+      THROW_WALLET_EXCEPTION_IF(idx >= m_transfers.size(), error::wallet_internal_error,
+        "Cold wallet signing: No record of output " + std::to_string(idx) + " in this wallet, run export_outputs \
+        on the online wallet and import_outputs here.");
```
thomasbuilds: Plain `export_outputs` cannot fix the state that reaches this check, the incremental export is empty whenever it fires (a partial one throws in `import_outputs` first), so recovery needs `all`.

```suggestion
        "Cold wallet signing: No record of output " + std::to_string(idx) + " in this wallet, "
        "run export_outputs all on the online wallet and import_outputs here.");
```

src/wallet/wallet2.cpp
```diff
@@ -8055,6 +8103,31 @@ bool wallet2::parse_tx_from_str(const std::string &signed_tx_st, std::vector<too
...
+    // - Verify with the assumption signed txs are redacted.
+    for (const auto &ptx : signed_txs.ptx)
+    {
+      // Manually check `signed_txs.key_images` so local state is not mutated before we validate.
+      // We inject the key image checker because `ptx` internal sorting ambiguity makes it cumbersome to
+      // directly validate key images here.
+      const auto &kis = signed_txs.key_images;
+      this->sanity_check_pending_tx(ptx, true, false, { [&kis](const size_t i) {
+        CHECK_AND_ASSERT_THROW_MES(i < kis.size(), "failed loading signed tx: ptx selected transfer \
+          is outside the bounds of imported key images");
```
thomasbuilds:
```suggestion
        CHECK_AND_ASSERT_THROW_MES(i < kis.size(), "failed loading signed tx: ptx selected transfer "
          "is outside the bounds of imported key images");
```

UkoeHB: ?

thomasbuilds: The trailing backslash splices the next line into the literal including its indentation so the message renders as `ptx selected transfer           is outside the bounds of imported key images`. Adjacent literals drop the extra spaces.

UkoeHB: Thanks must have forgotten about this.

thomasbuilds (approved):

src/wallet/pending_tx_validation.cpp
```diff
@@ -0,0 +1,767 @@
...
+                "sanity_check_pending_tx: v1 tx output amount doesn't match destination amount");
+        }
+
+        output_amnt += dest.amount;
+    }
+    CHECK_AND_ASSERT_THROW_MES(splitted_dsts_repro.size() == 0,
+        "sanity_check_pending_tx: failed checking splitted_dsts consistency");
+
+    // Balance check
+    output_amnt += (ptx.dust_added_to_fee ? 0 : ptx.dust);
```
j-berman:
```suggestion

```

Either the dust gets applied to dests or included in fee, so this would be double counting dust. It's already baked into the other fields summed into `output_amnt`. Can just remove this line

j-berman (commented):
Small comment, ready to approve

j-berman (approved):

thomasbuilds (approved):

j-berman (approved):
I really don't think we gain much from checking `m_spent` or `m_frozen` status anyway, I wouldn't mind just removing those checks and new param altogether. But looks fine to me either way.

j-berman (approved):

jpk68 (approved):

selsta (approved):

thomasbuilds (approved):

PR #11088, opened 2026-08-11
cryptonote_core: tx verif util to collect transp amt commitments

j-berman:
With FCMP++, we'll need the transparent amount commitments for all txs in a block in 3 distinct places:

1) Adding v2 coinbase outputs to the db.
2) Adding any transparent amount outputs to the curve tree.
3) Validating amount commitments for torsion (may not be strictly necessary for this).

Since we do those actions at separate call sites, rather than re-calculate transparent amount commitments at every call site, we collect them one time and avoid repeating the calculation.

### DIFF
--- a/src/cryptonote_core/tx_verification_utils.cpp
+++ b/src/cryptonote_core/tx_verification_utils.cpp
@@ -42,6 +42,28 @@
 
 using namespace cryptonote;
 
+static void collect_transparent_amount_commitments_static(
+    const std::vector<std::reference_wrapper<const transaction>> &tx_refs,
+    std::unordered_map<uint64_t, rct::key> &transparent_amount_commitments_inout)
+{
+    // Note: we do not clear transparent_amount_commitments_inout because it may be a rolling cache
+
+    for (const std::reference_wrapper<const transaction> &tx_ref : tx_refs)
+    {
+        const transaction &tx = tx_ref.get();
+
+        // We only need commitments for transparent amounts, which are tx version 1 || coinbase txs
+        if (tx.version > 1 && !cryptonote::is_coinbase(tx))
+            continue;
+        for (const auto &tx_out : tx.vout)
+        {
+            const uint64_t amount = tx_out.amount;
+            if (transparent_amount_commitments_inout.find(amount) == transparent_amount_commitments_inout.end())
+                transparent_amount_commitments_inout[amount] = rct::zeroCommitVartime(amount);
+        }
+    }
+}
+
 // Do RCT expansion, then do post-expansion sanity checks, then do full non-semantics verification.
 static bool expand_tx_and_ver_rct_non_sem(transaction& tx, const rct::ctkeyM& mix_ring)
 {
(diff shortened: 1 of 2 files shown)


src/cryptonote_core/tx_verification_utils.cpp
```diff
@@ -315,6 +337,46 @@ static bool ver_non_input_consensus_templated(TxForwardIt tx_begin, TxForwardIt
 namespace cryptonote
 {
 
+std::vector<std::reference_wrapper<const transaction>> collect_transparent_amount_commitments(
+    const transaction &miner_tx,
+    const std::vector<std::pair<transaction, blobdata>> &tx_pairs,
+    std::unordered_map<uint64_t, rct::key> &transparent_amount_commitments_inout)
```
j-berman, line 343: Sample caller: https://github.com/seraphis-migration/monero/blob/3f159b13e000ca9f0906599c9c5fd9c13f55233c/src/cryptonote_core/blockchain.cpp#L4858-L4859

src/cryptonote_core/tx_verification_utils.cpp
```diff
@@ -315,6 +337,46 @@ static bool ver_non_input_consensus_templated(TxForwardIt tx_begin, TxForwardIt
...
+    for (const auto &tx : tx_pairs)
+        tx_refs.push_back(std::cref(tx.first));
+    collect_transparent_amount_commitments_static(tx_refs, transparent_amount_commitments_inout);
+    return tx_refs;
+}
+
+std::vector<std::reference_wrapper<const transaction>> collect_transparent_amount_commitments(
+    const transaction &miner_tx,
+    const std::vector<transaction> &txs,
+    std::unordered_map<uint64_t, rct::key> &transparent_amount_commitments_inout)
```
j-berman, line 357: Sample caller: https://github.com/seraphis-migration/monero/blob/3f159b13e000ca9f0906599c9c5fd9c13f55233c/src/wallet/wallet2.cpp#L3208

src/cryptonote_core/tx_verification_utils.cpp
```diff
@@ -315,6 +337,46 @@ static bool ver_non_input_consensus_templated(TxForwardIt tx_begin, TxForwardIt
...
+    tx_refs.push_back(std::cref(miner_tx));
+    for (const auto &tx : txs)
+        tx_refs.push_back(std::cref(tx));
+    collect_transparent_amount_commitments_static(tx_refs, transparent_amount_commitments_inout);
+    return tx_refs;
+}
+
+std::vector<std::reference_wrapper<const transaction>> collect_transparent_amount_commitments(
+    const std::unordered_map<crypto::hash, std::pair<transaction, blobdata>> &txs_by_txid,
+    std::unordered_map<uint64_t, rct::key> &transparent_amount_commitments_inout)
```
j-berman, line 370: Sample caller: https://github.com/seraphis-migration/monero/blob/3f159b13e000ca9f0906599c9c5fd9c13f55233c/src/cryptonote_core/blockchain.cpp#L2169

j-berman (commented):
Included links showing the benefit of the `std::reference_wrapper` (to avoid copies), and the purpose of the distinct params across the function definitions

UkoeHB (approved):

j-berman:
I removed the `reference_wrapper` returns over in https://github.com/seraphis-migration/monero/pull/460

I think we'll want that change applied in this PR as well.

Issue #7184, opened 2020-12-23
7.1.7 banning nodes added to config as "add-priority-node" when building new pruned nodes


MoneroNodeHoster:
I am setting my new pruned nodes on Ubuntu 20.04 using the CLI daemon. The existing full nodes are also running the CLI daemon on Ubuntu 20.04.

I am currently running a few full nodes and thought I would help the network by adding a few additional pruned nodes in remote locations. I successfully added my existing full nodes to the config of the new pruned nodes as priority-nodes to decrease the strain on the overall monero network and to speed up the building of the new pruned nodes.

After tailing the logs of the new pruned nodes, I realized the new nodes are banning my full nodes, despite the full nodes being list as "add-priority-node."

The only remedy I have found thus far is to repeatedly keep restarting the new pruned nodes as they build themselves, to allow them to continue downloading from my full nodes.

If a whitelisting function could be added for IPs set for "add-priority-node" that would be a great help.

moneromooo-monero:
Post a level 2 log of this (--log-level 2)

MoneroNodeHoster:
All existing logs are level 0. I will try to recreate the issue using level 2.

To make a correction, the new nodes are "blocking" the priority nodes, not banning them.

MoneroNodeHoster:
51.79.71.134 is the priority node.

2020-12-24 03:09:52.248 [P2P1]  DEBUG   net     contrib/epee/include/net/levin_protocol_handler_async.h:477     [51.79.71.134:18080 OUT] LEVIN_PACKET_RECEIVED. [len=171918, flags1, r?=0, cmd = 2007, v=1
2020-12-24 03:09:52.248 [P2P1]  INFO    net.p2p.traffic contrib/epee/include/storages/levin_abstract_invoke2.h:44       [51.79.71.134:18080 OUT] 171918 bytes received for category command-2007 initiated by peer
2020-12-24 03:09:52.248 [P2P1]  INFO    net.p2p.msg     src/cryptonote_protocol/cryptonote_protocol_handler.inl:2501    [51.79.71.134:18080 OUT] Received NOTIFY_RESPONSE_CHAIN_ENTRY: m_block_ids.size()=4216, m_start_height=2254690, m_total_height=2258906
2020-12-24 03:09:52.248 [P2P1]  INFO    net.cn  src/cryptonote_protocol/cryptonote_protocol_handler.inl:2502    [51.79.71.134:18080 OUT] [0] state: received chain in state standby
[...]

PR #11090, opened 2026-08-11
p2p: add option to exempt local addresses from bans and restrictions

### LINKED ISSUE #9928: monerod should not ban local IP address

jpk68:
Continuation of #9929, with functionality gated behind the `--allow-local-ip` option.

Fixes #9928

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -265,6 +265,14 @@ namespace nodetool
   template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::is_remote_host_allowed(const epee::net_utils::network_address &address, time_t *t)
   {
+    if (m_allow_local_ip && (address.is_local() || address.is_loopback()) &&
+        (address.get_type_id() == epee::net_utils::ipv4_network_address::get_type_id() ||
+         address.get_type_id() == epee::net_utils::ipv6_network_address::get_type_id()))
+    {
+      MDEBUG("Local address " << address.host_str() << " is exempt from blocking/banning (--" << arg_p2p_allow_local_ip.name << ")");
+      return true;
+    }
+
     CRITICAL_REGION_LOCAL(m_blocked_hosts_lock);
 
     const time_t now = time(nullptr);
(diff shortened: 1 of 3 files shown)


src/p2p/net_node.inl
```diff
@@ -265,6 +265,14 @@ namespace nodetool
   template<class t_payload_net_handler>
   bool node_server<t_payload_net_handler>::is_remote_host_allowed(const epee::net_utils::network_address &address, time_t *t)
   {
+    if (m_allow_local_ip && address.is_local() &&
```
selsta:
```suggestion
    if (m_allow_local_ip && (address.is_local() || address.is_loopback()) &&
```

other places in the codebase check for both

selsta:
I'm not fully sure what the point of this flag is, is it for testing, or is it for Tor / I2P etc. If it's for testing, `--disable-rpc-ban` already exists.

plowsof:
while syncing a 2nd node on the same box this happened because of p2p scoring (so --disable-rpc-ban does not prevent it) in my case it was the node not responding quick enough to a span request `NOTIFY_REQUEST_GET_OBJECTS`

this seems related https://github.com/monero-project/monero/issues/7184 - except the node responding slowly is a remote one , perhaps `--add-priority-node` / `--add-exclusive-node` could be exempt from add_host_fail which would close that issue at least

PR #11091, opened 2026-08-11
tx_pool: do expensive verify after potential no-drop offenses

j-berman:


src/cryptonote_core/tx_pool.cpp
```diff
@@ -157,20 +157,10 @@ namespace cryptonote
...
-      tvc.m_verifivation_failed = true; // should already be set, but just in case
-      return false;
-    }
-
-    uint64_t fee;
+    const uint64_t fee = get_tx_fee(tx);
     bool fee_good = false;
     try
     {
-      // get_tx_fee() can throw. It shouldn't throw because we check preconditions in
```
j-berman, line 171: `get_tx_fee()` doesn't throw, it returns 0 on failure

src/cryptonote_core/tx_pool.cpp
```diff
@@ -1388,7 +1386,9 @@ namespace cryptonote
     CRITICAL_REGION_LOCAL1(m_blockchain);
     for(const auto& in: tx.vin)
     {
-      CHECKED_GET_SPECIFIC_VARIANT(in, const txin_to_key, tokey_in, true);//should never fail
```
j-berman, line 1391: Returning true for an incorrect type here is incorrect behavior. The tx type check happens in `ver_non_input_consensus` in `check_inputs_types_supported`.

selsta (approved):

jpk68 (approved):

PR #11092, opened 2026-08-12
rpc, p2p: remove pay-to-use RPC leftovers

thomasbuilds:
- `DEFAULT_PAYMENT_DIFFICULTY` / `DEFAULT_PAYMENT_CREDITS_PER_HASH`: unreferenced since the `--rpc-payment-*` options went.
- `RPCTracker`: `data()` was its only reader, so the map has been write-only since #10578, while the dtor still takes a mutex on every tracked RPC call. `RPC_TRACKER()` kept as a `PERF_TIMER()` alias so its 63 call sites stay untouched.
- `node_server::m_rpc_credits_per_hash`: setter's last caller went in #10578, the setter itself in 8d76584d0, so it can only hold 0. Explicit `= 0` kept because `basic_node_data` has no default member initialisers.

Left alone - follow-ups:
1. `rpc_access_request_base::client` and `rpc_access_response_base::{credits, top_hash}`: never written or read, but visible in the 28 commands inheriting each base, so it needs a `CORE_RPC_VERSION` minor bump. Both structs then collapse onto `rpc_request_base` / `rpc_response_base`.
2. `CORE_RPC_ERROR_CODE_PAYMENT_REQUIRED`, `CORE_RPC_ERROR_CODE_INVALID_CLIENT`, `CORE_RPC_STATUS_PAYMENT_REQUIRED`: no node path emits them, only wallet back-compat reads them.
3. `rpc_credits_per_hash` itself (49 lines, 12 files): `VARINT_FIELD` in `peerlist_entry_base`, so retiring it changes P2P and `peers.bin` serialisation.

### DIFF
--- a/src/p2p/net_node.h
+++ b/src/p2p/net_node.h
@@ -241,7 +241,6 @@ namespace nodetool
       : m_payload_handler(payload_handler),
         m_external_port(0),
         m_rpc_port(0),
-        m_rpc_credits_per_hash(0),
         m_allow_local_ip(false),
         m_hide_my_port(false),
         m_offline(false),
@@ -426,7 +425,6 @@ namespace nodetool
     uint32_t m_listening_port_ipv6;
     uint32_t m_external_port;
     uint16_t m_rpc_port;
-    uint32_t m_rpc_credits_per_hash;
     bool m_allow_local_ip;
     bool m_hide_my_port;
     bool m_offline;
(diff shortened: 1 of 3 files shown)


jpk68 (approved):
I had also noticed this and was going to ask about it - unless there is some reason I'm not aware of for why these shouldn't be removed, the changes made here look fine.

You might want to squash your commits, though ;)

tobtoht (approved):

selsta (approved):

PR #10940, opened 2026-07-21
epee: detect half-closed sockets so clients reconnect instead of failing

guoqiangliu-ocean:
## Context

On `nodex.monerujo.io:18081`, a successful HTTP/1.1 `get_info` response was
followed by a silent FIN after roughly five seconds of idle time, without a
`Connection: close` header. `blocked_mode_client::recv()` returned an empty
buffer on EOF without clearing `m_connected`, so the HTTP client kept treating
the half-closed socket as reusable.

## Root cause

The disconnected state was set only after the EOF branch returned. As a result,
EOF left `m_connected` set even though the peer had closed the socket.

## Changes

- mark `blocked_mode_client` disconnected on every read error, including EOF;
- add a loopback regression test covering data, FIN, disconnected state, and a
  successful new connection.

The transaction-construction changes formerly included here have been split
into #11093 following the maintainer's recommendation.

## Validation

- the branch now contains only the focused EOF commit and touches two files;
- `git diff --check` passes;
- the combined branch previously passed all 18 upstream CI checks; CI on this
  reduced branch is the authoritative check after the split.

Related downstream report: https://github.com/haveno-dex/haveno/issues/1093

PR #11093, opened 2026-08-12
wallet: avoid stale daemon calls during tx build

guoqiangliu-ocean:
## Context

This is the stale-daemon portion split from #10940 following the maintainer's
recommendation. The EOF handling fix and its regression test remain in #10940.

During transaction creation, local ring-member selection can take longer than
short remote-daemon keep-alive windows. The first subsequent `/get_outs.bin`
request can therefore reuse a socket that the daemon has already closed.

## Changes

- disconnect once before the first `/get_outs.bin` chunk, after local ring
  member selection, so the request reconnects lazily;
- reuse the `rct_config` and `use_view_tags` values already used to construct
  the transaction instead of querying fork rules again after proof generation.

## Impact

The patch is limited to transaction construction. It does not retry or replay
a failed request, switch nodes, or change the `sendrawtransaction` path.

## Validation

- cleanly cherry-picked onto current `master`;
- `git diff --check` passes;
- the same commit previously passed all 18 upstream CI checks as part of
  #10940; the new PR CI is the authoritative check for the split branch.

Related downstream report: https://github.com/haveno-dex/haveno/issues/1093

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -9490,6 +9490,9 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>> 
         chunk_req.outputs.push_back(req.outputs[offset + i]);
 
       const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};
+      // Ring member selection above may outlive short remote keep-alive windows.
+      if (offset == 0)
+        m_http_client->disconnect();
       bool r = epee::net_utils::invoke_http_bin("/get_outs.bin", chunk_req, chunk_daemon_resp, *m_http_client, rpc_timeout);
       THROW_ON_RPC_RESPONSE_ERROR(r, {}, chunk_daemon_resp, "get_outs.bin", error::get_outs_error, get_rpc_status(m_trusted_daemon, chunk_daemon_resp.status));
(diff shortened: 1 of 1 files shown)


src/wallet/wallet2.cpp
```diff
@@ -9490,6 +9490,9 @@ void wallet2::get_outs(std::vector<std::vector<tools::wallet2::get_outs_entry>>
         chunk_req.outputs.push_back(req.outputs[offset + i]);
 
       const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};
+      // Ring member selection above may outlive short remote keep-alive windows.
+      if (offset == 0)
+        m_http_client->disconnect();
```
jeffro256, line 9495: I don't understand when this would be useful if #10940 is accepted.

guoqiangliu-ocean: #10940 marks the client disconnected after a read observes EOF, so subsequent requests can reconnect, but the request that first encounters the stale socket still fails and is not replayed. This change proactively disconnects before the first `/get_outs.bin` request after potentially long local ring-member selection, causing that request to establish a fresh connection instead of discovering the stale socket through a failed RPC. The later changes also avoid new daemon fork-rule queries after proof generation by reusing the values already used to construct the transaction.

jpk68 (approved):

selsta:
Closing as we will go with #10940.

Issue #11059, opened 2026-08-06
Exception escaping `node_server::idle_worker()` permanently disables all p2p maintenance (out-peers → 0, connections never reaped)


sethforprivacy:
**Version:** v0.18.5.1-release (`Fluorine Fermi`), Linux x86_64, musl/Alpine build, Docker.

## Summary

`boosted_tcp_server::global_timer_handler()` re-arms its `deadline_timer` *after* invoking the
handler. If the handler throws, the re-arm is skipped and that idle-handler chain is **permanently
dead for the lifetime of the process** — silently, with no recovery path.

When the handler in question is `node_server::idle_worker()`, the node loses every piece of p2p
maintenance at once:

* `connections_maker()` stops → outbound peers decay to **0** and are never re-established
* `peer_sync_idle_maker()` stops → dead peers are never detected or dropped
* `gray_peerlist_housekeeping()`, `check_incoming_connections()`, `store_config()`,
  `update_dns_blocklist()` all stop

The node keeps accepting inbound connections and serving RPC, keeps reporting
`"status": "OK"` and `"synchronized": true`, and stays at chain tip via inbound peers only.
Externally it looks healthy. Internally it accumulates connection objects and file descriptors
forever.

`cryptonote::core::on_idle` is registered as a **separate** `idle_callback_conext` with its own
timer (`net_node.inl` ~L1042-1043), so it keeps running — which makes the failure even harder to
spot, since some periodic activity continues.

## Observed on a production node

16-day-old container, `RestartCount=0`, chain at tip, `status: OK`. Onset was ~day 12; the
preceding 12 days were normal.

```
get_info:  incoming_connections_count=1660   outgoing_connections_count=0
           height=3734124  synchronized=true  status=OK
           white_peerlist_size=1000  grey_peerlist_size=3980
```

`get_connections` — 1660 connection objects:

| incoming | state              | count |
|----------|--------------------|-------|
| true     | `before_handshake` | 1346  |
| false    | `before_handshake` | 280   |
| true     | `normal`           | 36    |

* `before_handshake` `live_time`: min 97 s, **p50 199 463 s (2.3 d), max 350 039 s (4.05 d)**
* age histogram by day: `{0: 311, 1: 373, 2: 456, 3: 484, 4: 2}` — a hard cliff at day 4 dating
  onset precisely
* 1511 / 1626 `before_handshake` connections have `send_count == 0`; 886 have `recv_count == 0`

[...]

PR #11097, opened 2026-08-12
epee: rearm idle handlers after exceptions

Ap4sh:
an exception from an idle callback currently unwinds before its timer is rearmed. the worker thread survives and resumes its io_context, but that callback is never scheduled again

this rearms the timer before rethrowing, preserving the existing exception handling while keeping the idle callback scheduled. callbacks that return false still stop normally

the regression test fails on master after the first callback and passes with this change

refs #11059

tests:
- release build with `BUILD_TESTS=ON`, `BUILD_GUI_DEPS=ON`, and `ENABLE_FUZZ_TEST=ON`
- `ctest --test-dir build --output-on-failure -E core_tests` (23/23)
- `ctest --test-dir build --output-on-failure -R core_tests` (1/1)
- regression test repeated 50 times

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.h
+++ b/contrib/epee/include/net/abstract_tcp_server2.h
@@ -56,6 +56,7 @@
 #include <boost/optional.hpp>
 #include "byte_slice.h"
 #include "net_utils_base.h"
+#include "scope_guard.h"
 #include "syncobj.h"
 #include "connection_basic.hpp"
 #include "network_throttle-detail.hpp"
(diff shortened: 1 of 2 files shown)


contrib/epee/include/net/abstract_tcp_server2.h
```diff
@@ -495,8 +495,17 @@ namespace net_utils
     bool global_timer_handler(/*const boost::system::error_code& err, */std::shared_ptr<idle_callback_conext<t_handler>> ptr)
     {
       //if handler return false - he don't want to be called anymore
-      if(!ptr->call_handler())
-        return true;
+      try
```
vtnerd: This seems like a good use case for the `scope_guard`. Both code paths merge into one. The only downside is that it would call `std::abort` if the call to `expires_after` or `async_wait` throws when `call_handler()` also threw.

vtnerd (changes requested):

contrib/epee/include/net/abstract_tcp_server2.h
Ap4sh: updated this to use `epee::unique_scope_guard`, releasing it when the handler returns false and calling `reset()` on the normal path so timer scheduling errors still propagate there

selsta:
In general I agree with this change but I don't know if it makes sense to merge this before we can figure out why there is an exception in the first place in https://github.com/monero-project/monero/issues/11059.

Ap4sh:
yeah that makes sense, i'm happy to leave this open while #11059 is investigated

the patch is only meant as a safety net, not a substitute for fixing the source of the exception

PR #11100, opened 2026-08-12
[WIP] fcmp_pp_rust: vendor cargo deps

tobtoht:
Scratchpad
---

|package|cloc|
|---|---:|
|libc|111724|
|syn|55395|
|curve25519-dalek|26403|
|bitvec|19401|
|typenum|15117|
|zerocopy|11636|
|hashbrown|11525|
|crypto-bigint|9177|
|zerocopy-derive|7545|
|proc-macro2|5085|
|ahash|3313|
|spin|2713|
|once_cell|2213|
|semver|2190|
|ec-divisors|1977|
|quote|1962|
|sha2|1921|
|unicode-ident|1735|
|digest|1646|
|generic-array-0.14.7|1421|
|wyz|1331|
|helioselene|1276|
|getrandom|1167|
|rustversion|1127|
|blake2|1096|
|zeroize|895|
|funty|863|
|rand_core|831|
|subtle|802|
|group|779|
|zeroize_derive|761|
|radium|689|
|dalek-ff-group-0.4.6|675|
|rustc_version|658|
|dalek-ff-group|593|
|version_check|577|
|curve25519-dalek-derive|542|
|block-buffer|487|
|ff|462|
|cpufeatures|442|
|tap|422|
|ciphersuite|349|
|flexible-transcript|221|
|std-shims|197|
|crypto-common|188|
|cfg-if|145|
|**total**|**311674**|


vtnerd:
This is every rust lib in another private lib? What's the advantage, assuming cargo can just pin by hash?

MrCyjaneK:
I'd assume offline builds are possible with this patch? Is something similar landing at fcmp++-stage too? Is it tied to guix only?

tobtoht:
>I'd assume offline builds are possible with this patch?

Yes, latest version should work for regular builds.

.gitmodules
```diff
@@ -11,3 +11,6 @@
 [submodule "external/gtest"]
 	path = external/gtest
 	url = https://github.com/google/googletest.git
+[submodule "src/fcmp_pp/fcmp_pp_rust/vendor"]
+	path = src/fcmp_pp/fcmp_pp_rust/vendor
+	url = https://github.com/tobtoht/monero-rust-deps.git
```
thomasbuilds, line 16: Is the plan to move this under monero-project before merge? Once vendored Cargo.lock does not pin the content: a modified crate with a regenerated .cargo-checksum.json files map and the original package checksum builds fine with --locked so the submodule commit is the only pin for all consensus Rust deps and changes there would get the same review path.

PR #11101, opened 2026-08-12
tests: fix functional_tests_rpc.py python3 crash and logging

AnasElii:
Screenshot 2026-08-12 at 23 14 20

- Fix crash on the usage path: ```string.join()``` was removed in Python 3, so running the script without a test argument crashed with ```AttributeError: module 'string' has no attribute 'join'```

- Add ```-v``` / ```--verbose``` flag: prints the full command line of each daemon/wallet before spawning it replace the old ```#print('Running: ...')``` debug line. The flag is removed from sys.argv immediately before the positional argument parsing, so the CMake/ctest invocation won't be effected 

and some Improvement of error handling and cleanup.


selsta:
Please undo unrelated whitespace changes and just fixing `string.join()` on line 37 is enough, the other join gets removed in #11105. The verbose flag is also not needed.

PR #11104, opened 2026-08-13
epee: fix hex separator handling

questfever:
## Summary

`from_hex_locale::to_vector()` first counts hexadecimal characters after excluding whitespace and colons. During decoding, however, separators were only skipped before the high nibble of each byte.

As a result, inputs such as `0:0` and `f f` passed the length check but then attempted to decode the separator as the low nibble, throwing `std::range_error`.

Skip separators before the low nibble as well, making decoding consistent with the initial character count and the documented behavior.

Also cast arguments passed to `std::isspace` and `std::tolower` to `unsigned char`. Passing a negative plain `char` to these functions has undefined behavior.

## Tests

Added regression coverage for colon and whitespace separators between hexadecimal nibbles.

```text
0:0 f f 0:f f:0
```

Tested with:

```shell
unit_tests --gtest_filter='HexLocale.*:FromHex.*:ToHex.*'
```

### DIFF
--- a/contrib/epee/src/hex.cpp
+++ b/contrib/epee/src/hex.cpp
@@ -134,7 +134,7 @@ namespace epee
     // should we include a specific character
     auto include = [](char input) {
         // we ignore spaces and colons
-        return !std::isspace(input) && input != ':';
+        return !std::isspace(static_cast<unsigned char>(input)) && input != ':';
     };
 
     // the number of relevant characters to decode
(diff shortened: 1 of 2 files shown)


selsta (approved):

jpk68 (approved):

questfever:
@jpk68 Thank you for your approval.   If there's anything need I to do, please feel free to let me know.

jpk68:
@questfever Nope, the PR just needs to be at least a week old before it's merged.

questfever:
> @questfever Nope, the PR just needs to be at least a week old before it's merged.

I understand. Thank you for the explanation.

PR #11106, opened 2026-08-14
cmake: enable highest optimization level for all targets

jpk68:


SChernykh:
Any benefits on ARM? Did you measure performance of `-O2` vs `-O3` on ARM devices? Also, `-O3` optimization level is very sensitive to UB in the code, so it needs extensive testing on all affected platforms. There must have been a reason it was at `-O2` on these platforms.

iamamyth:
> Did you measure performance of -O2 vs -O3 on ARM devices?

`-O2` and `-O3` often end up with similar performance, with `O3` merely breaking things due to undesired code removal (e.g. dropping intentional memory zeroing), arbitrarily rewriting code which contains UB, or reordering. Before you even bother with functional testing, I'd suggest checking the performance, and, if it's really much better, double-checking the "win" doesn't owe to dropping actually needed instructions.

jpk68:
Upon further consideration, this is probably not worth the time/effort. Closing.

PR #11114, opened 2026-08-14
src: deduplicate base32 encoding/decoding logic

jpk68:
Moves all Base32 {en,de}coding logic to a dedicated file in `src/common/`. This may also be useful in the future for Jamtis addresses (albeit with a different alphabet).

### DIFF
--- a/src/common/base32.h
+++ b/src/common/base32.h
@@ -1,0 +1,61 @@
+// Copyright (c) 2026, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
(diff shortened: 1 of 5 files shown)


src/common/base32.h
```diff
@@ -0,0 +1,61 @@
...
+#include <string>
+#include <string_view>
+#include <vector>
+
+namespace tools
+{
+namespace base32
+{
+
+constexpr const char alphabet[] = u8"abcdefghijklmnopqrstuvwxyz234567";
```
vtnerd: Why put the alphabet in the header if it's not an argument?

src/common/base32.h
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @brief Encodes a vector of bytes as an unpadded, lowercase Base32 string.
+ * @param data Bytes to encode.
+ * @return The Base32 encoding of `data`, or an empty string if `data` is empty.
+ */
+std::string encode(const std::vector<std::uint8_t>& data);
```
vtnerd: Consider making this an `epee::span` so any type with contiguous data can be used.

vtnerd (commented):
Im not seeing an advantage to this move if a different alphabet is going to be used - the decode function in particular isn't really prepped for such a change.

jpk68:
> Im not seeing an advantage to this move if a different alphabet is going to be used - the decode function in particular isn't really prepped for such a change.

IMO, it would be better to parameterize the function(s) when support for a different alphabet is needed, rather than now. There are still advantages to these changes today, namely that `base32_alphabet` isn't defined separately in both Tor and I2P-related files.

vtnerd:
> IMO, it would be better to parameterize the function(s) when support for a different alphabet is needed, rather than now. There are still advantages to these changes today, namely that base32_alphabet isn't defined separately in both Tor and I2P-related files.

But you've basically got to re-write the `decode` function entirely as it's hard-coded towards the current alphabet? Or else we just use the same alphabet I suppose.

jpk68:
> But you've basically got to re-write the `decode` function entirely as it's hard-coded towards the current alphabet? Or else we just use the same alphabet I suppose.

Yes, I suppose either could work, though it seems Jamtis will be using its own custom Base32 alphabet.

PR #11121, opened 2026-08-15
simplewallet: fix crash in sweep commands when address is omitted

jpk68:
This seems to occur only in debug builds (as well as in binaries from the Arch repos).

Edit: this might be happening because stack traces are now disabled by default in release builds.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -6975,6 +6975,13 @@ bool simple_wallet::sweep_main(uint32_t account, uint64_t below, const std::vect
       local_args.pop_back();
   }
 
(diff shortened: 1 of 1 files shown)


selsta (approved):

SNeedlewoods (approved):
I couldn't reproduce the crash, but before:
e.g. `sweep_all index=0` would lead to `Error: failed to parse address` since it doesn't check `local_args.size() > 0` here:
```
  if (!cryptonote::get_account_address_from_str_or_url(info, m_wallet->nettype(), local_args[0], oa_prompter))
```

Now with this patch, we get a correct error message and don't attempt to read empty `local_args[0]`.

PR #11124, opened 2026-08-15
simplewallet: update error message

jpk68:
Attempting to connect to a daemon for the wrong network (i.e. a mainnet daemon for a stagenet wallet) is another potential cause of this error.

### DIFF
--- a/src/simplewallet/simplewallet.cpp
+++ b/src/simplewallet/simplewallet.cpp
@@ -4559,8 +4559,9 @@ bool simple_wallet::try_connect_to_daemon(bool silent, uint32_t* version)
           "Please make sure the daemon is running the latest version or change the daemon address using the 'set_daemon' command.");
       else
         fail_msg_writer() << tr("wallet failed to connect to daemon: ") << m_wallet->get_daemon_address() << ". " <<
-          tr("Daemon either is not started or wrong port was passed. "
-          "Please make sure daemon is running or change the daemon address using the 'set_daemon' command.");
+          boost::format(tr("Daemon either is not started or the wrong port was passed. "
+          "Please make sure a %sdaemon is running or change the daemon address using the 'set_daemon' command."))
+            % (m_wallet->nettype() == TESTNET ? "testnet " : m_wallet->nettype() == STAGENET ? "stagenet " : "");
     }
     return false;
   }


selsta (commented):
Not sure if this is correct, `check_connection()` does not explicity check for network type.

src/simplewallet/simplewallet.cpp
```diff
@@ -4559,7 +4559,7 @@ bool simple_wallet::try_connect_to_daemon(bool silent, uint32_t* version)
           "Please make sure the daemon is running the latest version or change the daemon address using the 'set_daemon' command.");
       else
         fail_msg_writer() << tr("wallet failed to connect to daemon: ") << m_wallet->get_daemon_address() << ". " <<
-          tr("Daemon either is not started or wrong port was passed. "
+          tr("Daemon either is not started, for the wrong network, or the wrong port was passed. "
           "Please make sure daemon is running or change the daemon address using the 'set_daemon' command.");
```
thomasbuilds: I suggest

```suggestion
        fail_msg_writer() << tr("wallet failed to connect to daemon: ") << m_wallet->get_daemon_address() << ". " <<
          boost::format(tr("Make sure a %s daemon is running there, or change the daemon address using the 'set_daemon' command."))
            % (m_wallet->nettype() == TESTNET ? "testnet" : m_wallet->nettype() == STAGENET ? "stagenet" : "mainnet");
```

thomasbuilds:
> Not sure if this is correct, `check_connection()` does not explicity check for network type.

The default RPC ports differ per network so a stagenet wallet finds nothing on 38081 when only a mainnet daemon is running, and fails at connect.
~~An explicit --daemon-address to the wrong network does connect and skips this entirely.~~ That is the case that would need a nettype check.

SNeedlewoods (approved):
No strong opinion on this, the suggestion from thomasbuilds would also be okay for me, just one thought: For non-technical users "mainnet daemon" could be more confusing than just "daemon", so maybe keep this extra info exclusive to `stagenet` & `testnet`:
```
m_wallet->nettype() == TESTNET ? "testnet " : m_wallet->nettype() == STAGENET ? "stagenet " : ""
```

Another note, I did test this with a mainnet wallet and `--daemon-address` of a stagenet daemon and didn't hit this error on the first attempt. I think this is was selsta was referring to, `check_connection()` returns `true` if a daemon is running there and doesn't check if daemons nettype and wallets nettype match. Though the daemon blocked `Host 127.0.0.1` (for 24h) and after that this error was hit (even with a stagenet wallet).

AFAICT it would need a little effort to get the daemons nettype with `/get_info` in `check_connection()` (could add arg `bool do_check_nettype = false` to keep old behavior), which I'm not convinced is worth it, just for this PR.

src/simplewallet/simplewallet.cpp
jpk68: Thanks, done.

SNeedlewoods (approved):

thomasbuilds (approved):

selsta (approved):

PR #11125, opened 2026-08-15
simplewallet: add missing guards/locks for multisig commands

jpk68:


selsta (approved):

src/simplewallet/simplewallet.cpp
```diff
@@ -1154,6 +1161,7 @@ bool simple_wallet::make_multisig_main(const std::vector<std::string> &args, boo
     fail_msg_writer() << tr("wallet is watch-only and cannot be made multisig");
     return false;
   }
+  CHECK_IF_BACKGROUND_SYNCING("cannot be made multisig");
```
SNeedlewoods, line 1164: AFAICT this is unreachable under normal conditions.
1. open a background sync wallet
2. try `make_multisig`: errors out since `set enable-multisig-experimental 1` wasn't set
3. try `set enable-multisig-experimental 1`: errors out with "Background wallet cannot change wallet settings"

Haven't checked, but I assume it's the same for the ones below.

SNeedlewoods (approved):

PR #11127, opened 2026-08-16
epee: parse HTTP methods case-sensitively

### LINKED ISSUE #11118: epee `http_protocol_handler.inl` use `regex::icase` for method parsing but it should case sensitive
title.
that's a violation of HTTP standard which is like very very bad. We are all gonna die if we don't respect HTTP.

Ap4sh:
HTTP method names are case-sensitive, but the request-line parser used `boost::regex::icase` and accepted lowercase names as their uppercase equivalents

remove the case-insensitive flag so method names are parsed case-sensitively

fixes #11118

tests:
- debug `http_server.*`
- release `http_server.*`
- release `unit_tests` (1299 tests, 1297 passed, 2 hardware checks skipped)
- full release build

### DIFF
--- a/contrib/epee/include/net/http_protocol_handler.inl
+++ b/contrib/epee/include/net/http_protocol_handler.inl
@@ -400,7 +400,7 @@ namespace net_utils
   template<class t_connection_context>
 	bool simple_http_connection_handler<t_connection_context>::handle_invoke_query_line()
 	{ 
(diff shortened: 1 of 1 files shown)


selsta:
I think the test case can be removed, fix is easy enough.

Ap4sh:
done, removed the test case

redsh4de (approved):

jpk68 (approved):

SyntheticBird45 (approved):

PR #11128, opened 2026-08-17
wallet2: return concrete low priority when adjust_priority fails

woodser:
`adjust_priority` makes three separate daemon RPC calls (backlog estimate, block weight limit, recent block headers), so it can fail transiently.

On failure it currently returns the default priority of 0, so a client receives 0 = DEFAULT instead of a concrete value. That's ambiguous as an error signal, since 0 is also the legitimate result when auto-low-priority is disabled or a default priority is configured, so a caller can't classify it.

This PR returns a priority of 1 from the failure paths instead, matching the fallback already used in get_base_fee, where a priority of 0 is coerced to 1 with the [comment](https://github.com/monero-project/monero/blob/c5d310af8d3472c04aca4e9b0092efb4d2bfac0e/src/wallet/wallet2.cpp#L8538) "should not end up here". Behavior within wallet2 is unchanged, the fallback just becomes explicit for the function contract.

The alternative is to throw a hard error, but that would turn a transient hiccup into a hard failure of transfer creation for all present callers. Since this function only answers a narrow, best-effort heuristic, "is the cheap fee tier safe right now?", falling back to a concrete value is the better contract.

Release PR: https://github.com/monero-project/monero/pull/11129

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -8648,7 +8648,7 @@ fee_priority wallet2::adjust_priority(fee_priority priority)
       if (blocks.size() != 1)
       {
         MERROR("Bad estimated backlog array size");
-        return priority;
+        return fee_priority::Unimportant;
       }
       else if (blocks[0].first > 0)
       {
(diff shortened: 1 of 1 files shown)


nahuhh:
Just an opinion, but returning normal would be safer long term

woodser:
I can agree with that. This PR is deliberately behavior-preserving, so it makes explicit the fallback that already happens today, since get_base_fee coerces a leaked Default to the low tier. Switching the failure fallback to Normal would be a fee-policy change, so IMO that's best done as its own PR where it can be reviewed and applied consistently in one pass.

selsta (approved):

tobtoht (approved):

PR #11130, opened 2026-08-17
cryptonote: support pruned weights for old RingCT

### LINKED ISSUE #6754: get_pruned_transaction_weight does not support older transactions
Especially the Borromean range proof ones, which would be nice to sync pruned as they're very large unpruned.

Ap4sh:
fixes #6754

`get_pruned_transaction_weight()` can now reconstruct weights for Full, Simple and first-generation Bulletproof transactions

this lowers the v11 floor for pruned sync to the first RingCT fork, so old blocks with large Borromean proofs can be requested without the prunable data

v1 transactions are still sent in full by the existing block response path

tests:
- `unit_tests --gtest_filter=cn_format_utils.pruned_old_rct_weight:bulletproof.weight_pruned`
- full unit suite: 1309 passed, 2 disk tests skipped

### DIFF
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -590,7 +590,7 @@ namespace cryptonote
   {
     CHECK_AND_ASSERT_MES(tx.pruned, std::numeric_limits<uint64_t>::max(), "get_pruned_transaction_weight does not support non pruned txes");
     CHECK_AND_ASSERT_MES(tx.version >= 2, std::numeric_limits<uint64_t>::max(), "get_pruned_transaction_weight does not support v1 txes");
-    CHECK_AND_ASSERT_MES(tx.rct_signatures.type == rct::RCTTypeBulletproof2 || tx.rct_signatures.type == rct::RCTTypeCLSAG || tx.rct_signatures.type == rct::RCTTypeBulletproofPlus,
+    CHECK_AND_ASSERT_MES(tx.rct_signatures.type >= rct::RCTTypeFull && tx.rct_signatures.type <= rct::RCTTypeBulletproofPlus,
         std::numeric_limits<uint64_t>::max(), "Unsupported rct_signatures type in get_pruned_transaction_weight");
     CHECK_AND_ASSERT_MES(!tx.vin.empty(), std::numeric_limits<uint64_t>::max(), "empty vin");
     CHECK_AND_ASSERT_MES(tx.vin[0].type() == typeid(cryptonote::txin_to_key), std::numeric_limits<uint64_t>::max(), "empty vin");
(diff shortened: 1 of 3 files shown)


selsta:
Does it require changes here too? https://github.com/monero-project/monero/blob/20d9d3cb3155bfb1cf8cf1686ab83dc99b870cc9/src/cryptonote_protocol/cryptonote_protocol_handler.inl#L2244

Ap4sh:
yeah, good catch.. that was the second half of the same pre-Bulletproof guard, so i removed it too

src/cryptonote_protocol/cryptonote_protocol_handler.inl
```diff
@@ -2015,10 +2015,6 @@ skip:
     const uint32_t local_stripe = tools::get_pruning_stripe(m_core.get_blockchain_pruning_seed());
     if (local_stripe == 0)
       return false;
-    // don't request pre-bulletproof pruned blocks, we can't reconstruct their weight (yet)
-    static const uint64_t bp_fork_height = m_core.get_earliest_ideal_height_for_version(HF_VERSION_SMALLER_BP + 1);
```
jpk68, line 2019: Why was this removed? Now, pruned sync can be requested for any height, including for pre-RingCT blocks. Since `get_pruned_transaction_weight()` still asserts `tx.version >= 2`, and then returns `numeric_limits<uint64_t>::max()` on failure, pre-RingCT blocks will ostensibly end up with garbage weights instead of just being rejected.

src/cryptonote_basic/cryptonote_format_utils.cpp
```diff
@@ -474,33 +474,48 @@ namespace cryptonote
...
-
-    // calculate deterministic bulletproofs size (assumes canonical BP format)
-    size_t nrl = 0, n_padded_outputs;
-    while ((n_padded_outputs = (1u << nrl)) < tx.vout.size())
-      ++nrl;
-    nrl += 6;
-    extra = 32 * ((rct::is_rct_bulletproof_plus(tx.rct_signatures.type) ? 6 : 9) + 2 * nrl) + 2;
-    weight += extra;
+    size_t n_padded_outputs = 0;
+    if (rct::is_rct_borromean(tx.rct_signatures.type))
```
jpk68: Nit: `is_rct_borromean()` is computed twice in this same function - can it be cached to avoid that?

src/cryptonote_protocol/cryptonote_protocol_handler.inl
Ap4sh: good point, i restored a floor at the v4 RingCT fork. v1 txes are fetched with their prunable blob by `fill()` and sent with a null prunable hash, so they stay full and do not reach `get_pruned_transaction_weight()`, but there is no benefit in requesting pruned data before RingCT and keeping that boundary narrows the change

src/cryptonote_basic/cryptonote_format_utils.cpp
Ap4sh: done, cached it as `borromean` and reused it for both branches

jpk68 (approved):

PR #11131, opened 2026-08-17
epee: fix rpc write timer regression and misconfiguration

nahuhh:
issue: using basicswap (monero-wallet-rpc) connected to a healthy remote node over tor (responds to get_height w/o issue, so failover not triggered). Repeated sync fails which recover after the link improves. Monitoring the link shows that a 10MB response was being truncated at 6, 7, 8 etc MBs and trying again.

prior to 7760, the connection timer was reset before the write began. after,  it isnt reset until after the write completes - the write has to complete in whatever time remained, causing responses to be truncated and retried indefinitely (until the connection speeds up).

related fix: new connection timeout local was commented as 2minutes but the ms were 20mins. Change to 2min.

### DIFF
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -60,7 +60,7 @@
 #define MONERO_DEFAULT_LOG_CATEGORY "net"
 
 #define AGGRESSIVE_TIMEOUT_THRESHOLD 120 // sockets
-#define NEW_CONNECTION_TIMEOUT_LOCAL 1200000 // 2 minutes
+#define NEW_CONNECTION_TIMEOUT_LOCAL 120000 // 2 minutes
 #define NEW_CONNECTION_TIMEOUT_REMOTE 10000 // 10 seconds
 #define DEFAULT_TIMEOUT_MS_LOCAL 1800000 // 30 minutes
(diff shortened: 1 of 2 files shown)


jpk68 (approved):

selsta (approved):

plowsof (approved):

selsta (approved):

jpk68 (approved):

plowsof (approved):

Ap4sh:
this looks like a regression for long-running RPC handlers: `NEW_CONNECTION_TIMEOUT_LOCAL` now expires after 120 seconds, but `generateblocks` processes all 500 blocks synchronously before `do_send()` / `start_write()` can extend the timer. the [master test step before this merge](https://github.com/monero-project/monero/actions/runs/35831350085/job/107084453461) passed; the [first master run after it](https://github.com/monero-project/monero/actions/runs/35831438124/job/107084731711) failed at `p2p.py:160` (`generateblocks(..., 500)`) with `RemoteDisconnected`. the [same failure on another branch](https://github.com/Ap4sh/monero/actions/runs/35834196009/job/107093614510) reproduced it.

could we retain a longer local timeout, or extend/suspend it while an RPC handler is active? the new unit test covers a slow response write, but not synchronous request processing.

plowsof:
yes reproduced, thanks @Ap4sh , forwarded

PR #11132, opened 2026-08-17
wallet_rpc_server: leave unknown incoming tx hash empty

### LINKED ISSUE #9447: Offline wallet RPC "incoming_transfers" method reports tx_hash as 0000...
When querying an offline wallet the  RPC "incoming_transfers" method replies with `"tx_hash": "0000000000000000000000000000000000000000000000000000000000000000"`

```bash
$ curl http://localhost:18083/json_rpc \
  -d '{"jsonrpc":"2.0","id":"0","method":"incoming_transfers", "params":{ "transfer_type": "all" } }' -H 'Content-Type: application/json'
```

```json
 "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "transfers": [{
      "amount": 35164211634107,
      "block_height": 0,
      "frozen": false,
      "global_index": 302,
      "key_image": "2e05e81c716340d9e99cce5c228c61250a1a6b4bd70a1c546db4ce9372ab74cf",
      "pubkey": "5e2ece3fa6ee78f21670b133ffeac463a6b18f15111e5fda33496491bc04f2d1",
      "spent": true,
      "subaddr_index": {
        "major": 0,
        "minor": 0
      },
      "tx_hash": "0000000000000000000000000000000000000000000000000000000000000000",
      "unlocked": false
    },{
      "amount": 35164144563696,
      "block_height": 0,
      "frozen": false,
      "global_index": 303,
      "key_image": "1231bb43f3672247104a6bcc60a2a2bfbe692deba010b0cf9df590f66e20270c",
      "pubkey": "2c92e406e0a017ef73770b32b5118b8e179e5a4e93c7c4c2810e4aabf8cf572a",
      "spent": true,
      "subaddr_index": {
        "major": 0,
        "minor": 0
      },
      "tx_hash": "0000000000000000000000000000000000000000000000000000000000000000",
      "unlocked": false
    },{
      "amount": 35164077493412,
      "block_height": 0,
      "frozen": false,
      "global_index": 304,
      "key_image": "c5002670d788dd3b993dd78eaf410456b30b861484b7fd92e47c6a50853489d3",
      "pubkey": "34c050ff1070d097c0298165dcd862ad4aacaf4a158f398d17349da9ec51c918",
      "spent": true,
      "subaddr_index": {
        "major": 0,
        "minor": 0
      },
      "tx_hash": "0000000000000000000000000000000000000000000000000000000000000000",
      "unlocked": false
    }...
```

SlowBearDigger:
Fixes #9447

`import_outputs` sets `m_txid` to `crypto::null_hash` for imported
outputs because the output export format does not contain transaction
hashes.

`incoming_transfers` serialized that sentinel as 64 zeroes, making an
unknown value look like a transaction hash. Return an empty string when
the transaction hash is unknown. Known transaction hashes are unchanged.

Before:

```json
"tx_hash": "0000000000000000000000000000000000000000000000000000000000000000"
```

After:

```json
"tx_hash": ""
```

The cold signing functional test now verifies that online wallet
transfer hashes remain non-zero 64-character values and imported
offline wallet hashes are empty.

Tests:

- `make -j4 release-all`
- `cold_signing`
- `transfer`

### DIFF
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -2269,7 +2269,7 @@ namespace tools
         rpc_transfers.amount       = td.amount();
         rpc_transfers.spent        = td.m_spent;
         rpc_transfers.global_index = td.m_global_output_index;
-        rpc_transfers.tx_hash      = epee::string_tools::pod_to_hex(td.m_txid);
+        rpc_transfers.tx_hash      = td.m_txid == crypto::null_hash ? "" : epee::string_tools::pod_to_hex(td.m_txid);
         rpc_transfers.subaddr_index = {td.m_subaddr_index.major, td.m_subaddr_index.minor};
         rpc_transfers.key_image    = td.m_key_image_known ? epee::string_tools::pod_to_hex(td.m_key_image) : "";
         rpc_transfers.pubkey       = epee::string_tools::pod_to_hex(td.get_public_key());
--- a/src/wallet/wallet_rpc_server_commands_defs.h
+++ b/src/wallet/wallet_rpc_server_commands_defs.h
@@ -47,7 +47,7 @@
 // advance which version they will stop working with
 // Don't go over 32767 for any of these
 #define WALLET_RPC_VERSION_MAJOR 1
-#define WALLET_RPC_VERSION_MINOR 32
+#define WALLET_RPC_VERSION_MINOR 33
 #define MAKE_WALLET_RPC_VERSION(major,minor) (((major)<<16)|(minor))
 #define WALLET_RPC_VERSION MAKE_WALLET_RPC_VERSION(WALLET_RPC_VERSION_MAJOR, WALLET_RPC_VERSION_MINOR)
 namespace tools
--- a/tests/functional_tests/cold_signing.py
+++ b/tests/functional_tests/cold_signing.py
@@ -145,6 +145,9 @@ class ColdSigningTest():
             if do_check_key_images:
                 attributes_to_cmp.append("key_image")
             for i in range(len(hot_transfers_list)):
+                assert len(hot_transfers_list[i]['tx_hash']) == 64
+                assert hot_transfers_list[i]['tx_hash'] != '0' * 64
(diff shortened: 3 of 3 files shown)


jpk68 (approved):

selsta (approved):

src/wallet/wallet_rpc_server.cpp
```diff
@@ -2269,7 +2269,7 @@ namespace tools
         rpc_transfers.amount       = td.amount();
         rpc_transfers.spent        = td.m_spent;
         rpc_transfers.global_index = td.m_global_output_index;
-        rpc_transfers.tx_hash      = epee::string_tools::pod_to_hex(td.m_txid);
+        rpc_transfers.tx_hash      = td.m_txid == crypto::null_hash ? "" : epee::string_tools::pod_to_hex(td.m_txid);
```
jeffro256, line 2272: This changes RPC behavior; `WALLET_RPC_VERSION_MINOR` needs to be bumped.

SlowBearDigger: Thanks for reviewing @jeffro256, my bad for letting this slip.

Fixed already.

jpk68 (approved):

jeffro256 (approved):

yagop:
This changes the symptom but I don't think it resolves the underlying issue in #9447.

The root problem is that `incoming_transfers` on an offline wallet doesn't return the transaction hash. This MR swaps the misleading `"0000...0000"` for an empty string; which is more honest, but the caller still gets no `tx_hash`.

The root cause is that the export blob from the hot wallet doesn't carry the txid: `exported_transfer_details` (`src/wallet/wallet2.h`) has no `m_txid` field, so `import_outputs` sets it to `null_hash` (`src/wallet/wallet2.cpp`). Until that's addressed, the cold wallet simply has nothing to report.

The struct is already versioned, so the hash could be added compatibly (bump to v2, serialize `m_txid` when `version >= 2`), at a cost of +32 bytes per output.

PR #11134, opened 2026-08-17
wallet2: abort in-flight daemon request on shutdown

woodser:
`wallet2::shutdown()` only sets a flag, which `refresh` checks between daemon RPC calls. If the current call is blocked on a stalled or unresponsive connection, the shutdown has no effect until the call hits the full RPC timeout (3.5 minutes by default), so shutting down a wallet mid-refresh (ctrl+c on wallet-rpc, or shutting down an embedded wallet) can hang for minutes, blocking progress to restart the wallet or shut down the application.

This PR makes `wallet2::shutdown()` abort any in-flight daemon request by adding `shutdown()` to epee's http client. The epee override is safe to call from any thread. The base class default is a plain `disconnect()` for clients that don't support cross-thread interruption.

The new shutdown() sets a sticky atomic flag and wakes the blocked thread with a single self-pipe write (async-signal-safe; a plain post() on Windows, where console handlers run on their own thread). The blocked thread fails its request promptly and closes its own socket, so no socket is ever touched cross-thread. Connect, including the TLS handshake, and send/recv are all abortable.

As a result, wallets shut down promptly and cleanly.

Release PR: https://github.com/monero-project/monero/pull/11135

### DIFF
--- a/contrib/epee/include/net/abstract_http_client.h
+++ b/contrib/epee/include/net/abstract_http_client.h
@@ -68,6 +68,7 @@ namespace http
     virtual void set_auto_connect(bool auto_connect) = 0;
     virtual bool connect(std::chrono::milliseconds timeout) = 0;
     virtual bool disconnect() = 0;
+    virtual bool shutdown() { return disconnect(); } // tear down: the client need not be usable afterwards; overrides should interrupt blocked calls and be callable from another thread (the default plain disconnect() gives neither guarantee)
     virtual bool is_connected(bool *ssl = NULL) = 0;
     virtual bool invoke(const boost::string_ref uri, const boost::string_ref method, const boost::string_ref body, std::chrono::milliseconds timeout, const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) = 0;
     virtual bool invoke_get(const boost::string_ref uri, std::chrono::milliseconds timeout, const std::string& body = std::string(), const http_response_info** ppresponse_info = NULL, const fields_list& additional_params = fields_list()) = 0;
--- a/contrib/epee/include/net/net_helper.h
+++ b/contrib/epee/include/net/net_helper.h
@@ -34,6 +34,7 @@
 #include <boost/version.hpp>
 #include <boost/asio/io_context.hpp>
 #include <boost/asio/ip/tcp.hpp>
+#include <boost/asio/post.hpp>
 #include <boost/asio/read.hpp>
 #include <boost/asio/ssl.hpp>
 #include <boost/asio/steady_timer.hpp>
@@ -43,6 +44,13 @@
 #include <boost/system/error_code.hpp>
 #include <boost/utility/string_ref.hpp>
 #include <functional>
+#ifdef BOOST_ASIO_HAS_POSIX_STREAM_DESCRIPTOR
+#include <boost/asio/posix/stream_descriptor.hpp>
+#include <cerrno>
+#include <fcntl.h>
+#include <stdexcept>
+#include <unistd.h>
+#endif
 #include "net/net_ssl.h"
 
 #include "misc_log_ex.h"
@@ -101,11 +109,39 @@ namespace net_utils
 				m_ssl_options(epee::net_utils::ssl_support_t::e_ssl_support_autodetect),
 				m_connected(false),
 				m_deadline(m_io_service, std::chrono::steady_clock::time_point::max()),
-				m_shutdowned(false),
+#ifdef BOOST_ASIO_HAS_POSIX_STREAM_DESCRIPTOR
+				m_wake_reader(m_io_service),
+				m_wake_writer(-1),
+#endif
+				m_aborted(false),
 				m_bytes_sent(0),
 				m_bytes_received(0)
 		{
 			check_deadline();
+#ifdef BOOST_ASIO_HAS_POSIX_STREAM_DESCRIPTOR
+			// self-pipe so shutdown() can wake a blocked run_one() with an async-signal-safe
+			// write(); without it aborts would wait out the op deadline, so refuse to construct
+			int fds[2];
+			if (0 != ::pipe(fds))
+				throw std::runtime_error("failed to create shutdown wake pipe");
+			bool flags_set = true;
+			for (const int fd : fds)
+			{
+				if (-1 == ::fcntl(fd, F_SETFL, O_NONBLOCK) || -1 == ::fcntl(fd, F_SETFD, FD_CLOEXEC))
+					flags_set = false;
+			}
+			boost::system::error_code ec;
+			if (flags_set)
+				m_wake_reader.assign(fds[0], ec);
+			if (!flags_set || ec)
+			{
+				::close(fds[0]);
+				::close(fds[1]);
+				throw std::runtime_error("failed to set up shutdown wake pipe");
+			}
+			m_wake_writer = fds[1];
+			wait_wake();
+#endif
 		}
 
 		/*! The first/second parameters are host/port respectively. The third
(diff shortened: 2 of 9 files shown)


contrib/epee/include/net/net_helper.h
```diff
@@ -157,14 +164,20 @@ namespace net_utils
...
+				}
 				if (m_ssl_socket->next_layer().is_open())
 				{
 					m_connected = true;
 					m_deadline.expires_at(std::chrono::steady_clock::time_point::max());
 					// SSL Options
 					if (m_ssl_options.support == epee::net_utils::ssl_support_t::e_ssl_support_enabled || m_ssl_options.support == epee::net_utils::ssl_support_t::e_ssl_support_autodetect)
 					{
-						if (!m_ssl_options.handshake(m_io_service, *m_ssl_socket, boost::asio::ssl::stream_base::client, {}, addr, timeout))
+						if (!m_ssl_options.handshake(m_io_service, *m_ssl_socket, boost::asio::ssl::stream_base::client, {}, addr, timeout, [this] { return shutdown_requested(); }))
```
thomasbuilds, line 219: An abort landing inside this handshake is indistinguishable from "server has no TLS": `handshake()` returns false, this maps to `CONNECT_NO_SSL` (line 188), and with the default `e_ssl_support_autodetect` `connect()` permanently sets `support = e_ssl_support_disabled` (line 230) - every later auto-reconnect is silent plaintext until the next set_daemon, and a MITM can widen the window by stalling handshakes. I suggest checking `shutdown_requested()` after a failed handshake and returning `CONNECT_FAILURE` instead.

src/wallet/wallet2.h
```diff
@@ -796,7 +796,7 @@ namespace tools
       const std::string &proxy = "");
     bool set_proxy(const std::string &address);
 
-    void stop() { m_run.store(false, std::memory_order_relaxed); m_message_store.stop(); }
+    void stop() { m_run.store(false, std::memory_order_relaxed); m_message_store.stop(); m_http_client->shutdown(); }
```
thomasbuilds: This makes `stop()` async-signal-unsafe: `shutdown()` does `boost::asio::post()` (a heap allocation plus the io_context's non-recursive scheduler mutex) and `stop()` runs in real POSIX signal-handler context (simplewallet's ^C during manual refresh via `interrupt()`, wallet-rpc's initial-sync handler). A signal landing on the thread inside `run_one()` or mid-malloc can self-deadlock or corrupt the heap. Pre-PR stop() was only relaxed atomic stores.

contrib/epee/src/net_helper.cpp
```diff
@@ -12,80 +12,73 @@ namespace net_utils
...
-
-		bool try_ipv6 = false;
-		boost::asio::ip::tcp::resolver::results_type results{};
-		boost::system::error_code resolve_error;
-
-		try
+		const auto shared = std::make_shared<new_connection>(MONERO_GET_EXECUTOR(timeout));
+		// close on cancel too, so an attempt abandoned by shutdown() is reclaimed when
+		// the timer is next re-armed; a completed attempt has moved its socket out
+		timeout.async_wait([shared] (boost::system::error_code)
```
thomasbuilds: This one-shot wait is registered before resolution, so if the deadline fires while `getaddrinfo` is still running, the handler is consumed doing nothing (`resolver_.cancel()` can't stop a started lookup; the socket isn't open yet) and `check_deadline` re-arms to `time_point::max()` - the later `async_connect` (or the v6 fallback resolve) then runs with no deadline and blocks until the OS TCP timeout (~2 min). Related: client destruction joins the resolver worker thread stuck in that same lookup, so teardown after an "aborted" DNS stall still blocks. Pre-PR the resolve was synchronous, so the timer always covered an already-open socket.

contrib/epee/include/net/abstract_http_client.h
```diff
@@ -68,6 +68,7 @@ namespace http
     virtual void set_auto_connect(bool auto_connect) = 0;
     virtual bool connect(std::chrono::milliseconds timeout) = 0;
     virtual bool disconnect() = 0;
+    virtual bool shutdown() = 0; // interrupt a blocked call; must be safe to call from another thread
```
thomasbuilds:
```suggestion
    virtual bool shutdown() { return disconnect(); }
```

The description says "the base class default is a plain disconnect()", but this is pure virtual - every out-of-tree subclass (this class is wallet2's injection point) breaks at compile time and will get a do-nothing stub, silently losing the abort there.

tests/unit_tests/net.cpp
```diff
@@ -94,6 +96,188 @@ namespace
...
+    client.set_connector(net::socks::connector{
+        std::make_shared<net::socks::endpoint>(proxy.local_endpoint())
+    });
+    std::thread connecting{[&client] {
+        EXPECT_FALSE(client.connect("10.0.0.1", "18081", std::chrono::seconds{30}));
+    }};
+    boost::system::error_code error;
+    boost::asio::ip::tcp::socket stalled{server_io};
+    proxy.accept(stalled, error);
+    ASSERT_FALSE(error);
```
thomasbuilds:
```suggestion
    EXPECT_FALSE(error);
```
A transient accept failure here early-returns past the still-joinable `connecting` thread - its destructor calls `std::terminate` and kills the whole test binary. `EXPECT` keeps the `shutdown()`/`join()` below reachable (they complete harmlessly on the error path).

thomasbuilds (changes requested):

thomasbuilds:
One design-level concern: `stop()` is not just shutdown. It is also the routine pause primitive (`LOCK_IDLE_SCOPE` before most simplewallet commands, `LOCK_REFRESH` and `WalletImpl::stop()` in the wallet API), and this PR upgrades it from a cooperative flag to a hard abort of healthy in-flight requests. Worst case: aborting `/sendrawtransaction` between send and response makes commit_tx throw no_connection_to_daemon (outputs not marked spent) while the daemon may have relayed it, so a retry attempts a double spend. This state was unreachable from `stop()` before. Day-to-day: on a syncing wallet, most interactive commands now abort the in-flight refresh chunk and force a TCP+TLS (or Tor circuit) reconnect on the next call. Splitting the cooperative `stop()` (atomic stores only, which also fixes the signal-safety issue raised on wallet2.h) from a teardown-only abort call would resolve both. A sticky abort bit is not an option precisely because of the pause usage.

thomasbuilds:
The split makes stop() signal-safe again but the two wallet-rpc handlers now reach the post() whenever a request is in flight: the running-server lambda calls stop_refresh() which does m_wallet->shutdown() and the pre-server one calls wal->shutdown() directly. boost::asio::post() allocates and touches the scheduler's queue/lock which is not async-signal-safe, so a ^C landing on the thread inside run_one() or mid-malloc can deadlock or corrupt the heap in the handler. A signal-safe wake would need something like a self-pipe (write() is async-signal-safe) instead of post(), or the handlers staying on the atomic-only stop()/m_stopped path with the socket abort issued off signal context.

thomasbuilds:
> This PR makes stop() abort any in-flight daemon request

PR description needs update

thomasbuilds (approved):

thomasbuilds (approved):

woodser:
I found that this PR can be simplified significantly based on the outcome of these reviews, since it's now only a strict teardown. shutdown() just sets a sticky abort flag and wakes the blocked call, and the client stays unusable afterwards, matching what every caller wants. I've updated this PR and #11135 to the distilled version.

amin404ala (approved):

amin404ala (approved):

thomasbuilds (approved):
> The client reconnects transparently on next use, and shutdown is a no-op when no request is in flight, so idle connections are untouched

PR description needs update

woodser:
> PR description needs update

Done, and I have rebased.

thomasbuilds (approved):
PR title doesn't match commit name

contrib/epee/include/net/net_helper.h
jpk68, line 582: Since this was not updated, it looks like the new abort flag will be ignored here (??)

jpk68:
Two more things:

- It looks like the DNS resolution code in `net_helper.cpp` still can't be aborted at all.
- The destructor in `net_helper.h` seems to lose graceful shutdown after an idle abort.

selsta:
> It looks like the DNS resolution code in net_helper.cpp still can't be aborted at all.

I suggested to woodser to keep DNS out of this PR.

> seems to lose graceful shutdown after an idle abort

Appears intentional with this PR.

contrib/epee/include/net/net_helper.h
selsta: Seems this would delay it at max 2 seconds, should be okay to keep as is.

selsta (approved):

PR #11138, opened 2026-08-18
wallet_rpc_server: stop refresh on shutdown

woodser:
On shutdown, a running refresh blocks server exit until it finishes.

This PR stops the wallet first so it doesn't block, matching the CLI's ^C handler, so the refresh exits at the next chunk and the server can shut down promptly.

Release PR: https://github.com/monero-project/monero/pull/11139

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1208,6 +1208,7 @@ wallet2::wallet2(network_type nettype, uint64_t kdf_rounds, bool unattended, std
   m_http_client(http_client_factory->create()),
   m_upper_transaction_weight_limit(0),
   m_run(true),
+  m_stopped(false),
   m_callback(0),
   m_trusted_daemon(false),
   m_nettype(nettype),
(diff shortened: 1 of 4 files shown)


jpk68 (approved):

selsta (approved):

thomasbuilds:
stop_refresh() boils down to m_run.store(false) but refresh() re-arms m_run near its start (wallet2.cpp:4105), so a stop landing between a refresh being dispatched and that re-arm is silently erased and the refresh runs on - unbounded for the refresh/rescan_blockchain RPCs. Under systemd/docker (single SIGTERM, then SIGKILL) that can lose hours of unsaved sync progress.

Since the re-arm can't be removed (simplewallet relies on it) fully closing this would need a stop flag that refresh() doesn't re-arm checked in the pull loops alongside m_run.

thomasbuilds (approved):

thomasbuilds (changes requested):
Last thing from me: the pre-server handler stays installed after the quit check so a signal landing between there and wal.release() calls wal->shutdown() on the wallet the server then serves. That sets the sticky m_stopped so refresh_running() is false for the rest of the process and every refresh returns 0 blocks with status OK, silently never syncing.
Before the same signal only set m_run, which refresh() re-arms at the top of refresh(), so it was harmless.

thomasbuilds (approved):
rip ci

jpk68 (approved):

selsta (approved):

PR #11141, opened 2026-08-19
rpc, common: remove dead code

thomasbuilds:
Never had a caller:
- `t_rpc_client::basic_json_rpc_request`: added in 9193d6fb5 alongside `json_rpc_request`, which is the variant everything uses
- `PERF_TIMER_L`, added in 74dfdb0b3, and `PERF_TIMER_UNIT_L`, added in 19d7f568c: never invoked; the current definition from 289937979 does not even expand to valid C++, so any invocation would fail to compile

Declarations whose definition and callers were removed long ago:
- `check_core_busy`: 0f2c2d4c3 removed the obsolete busy-core checks and the definition, leaving the declaration
- `get_os_version_string`: 6b8dfb8fb removed the os-version daemon command together with the definition, leaving the declaration and its comment, which referenced helpers that no longer exist

The `FullMessage` members are dropped from this PR: `getRequest` is used by monero-lws, and the ZMQ code is still under development.

### DIFF
--- a/src/common/perf_timer.h
+++ b/src/common/perf_timer.h
@@ -77,9 +77,7 @@ private:
 
 #define PERF_TIMER_NAME(name) pt_##name
 #define PERF_TIMER_UNIT(name, unit) tools::LoggingPerformanceTimer PERF_TIMER_NAME(name)(#name, "perf." MONERO_DEFAULT_LOG_CATEGORY, unit, tools::performance_timer_log_level)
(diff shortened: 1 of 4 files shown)


jpk68 (approved):

selsta:
> FullMessage::getRequest

It might make sense to keep this and other ZMQ code, since it's still in develpment.

vtnerd:
LWS uses `getRequest`. So I would have to work-around it on that end.

jpk68 (approved):

selsta (approved):

PR #11142, opened 2026-08-19
cryptonote_basic: remove dead code

thomasbuilds:
Never had a caller:
- `get_hash_stats` and the four hash counters it read: added in f065234b7, the counters only ever incremented, their single reader never called
- `get_path_from_aux_slot`: added in 19b228393; add_aux_pow only ever used `get_aux_slot` and `encode_mm_depth`
- `get_window_size`: added in e546f3724, which wired the voting-info RPC through `get_voting_info` instead
- `account_base::load` and `store`: declared at the initial import, never defined
- `short_hash_str`: from the initial import, never called
- `t_serializable_object_from_blob`: template added in 064ab1234, never instantiated
- `ring_signature`: typedef from the initial import, never referenced by code in-tree

Also drops two commented-out fossils from the initial import: an `out_t` typedef naming a type that never existed and a stray namespace line.


jpk68 (approved):

thomasbuilds:
@UkoeHB account_generators.h is dead code, please reply if I can include it in this PR for deletion.

UkoeHB:
Yes it can be removed, was buried on my todo list.

jpk68 (approved):

UkoeHB (approved):

PR #11143, opened 2026-08-19
src, epee: replace recursive mutexes with read-write lock

jpk68:
Replaces the recursive mutex used in `src/cryptonote_core/blockchain.{cpp,h}` with a read-write lock mechanism. This PR is heavily based on #9181, with a few tweaks and updates.

Related: #11063, #11051

Below is a graph of RPC wallet sync performance before and after these changes (also with #11051):
monerod-rpc-threads-three-way-fast-wallet-suite


selsta (commented):
Could you try to enable `BOOST_THREAD_V2_SHARED_MUTEX`? Without it, Boost's default POSIX `shared_mutex` implementation has a known writer starvation issue, see https://github.com/boostorg/thread/issues/265 and https://github.com/boostorg/thread/issues/361.

jpk68:
Disclaimer: an LLM was used in making the unit tests, however all test code was reviewed by a human (me).

jpk68:
Rebased.

PR #11148, opened 2026-08-19
epee: prevent JSON parser log injection

selsta:


jpk68 (approved):

tobtoht (approved):

contrib/epee/src/parserse_base_utils.cpp
```diff
@@ -187,13 +186,17 @@ namespace misc_utils
           }else if(*it == '\\')
           {
             escape_mode = true;
-          }          
+          }
+          else if(static_cast<unsigned char>(*it) < 0x20)
+          {
+            ASSERT_MES_AND_THROW("Unescaped control character in JSON string");
```
jeffro256: I need to check, but I vaguely remember a discussion a couple years ago about one or more of the RPC endpoints depending on non-standard JSON bytes in a string, so this change might be breaking.

contrib/epee/include/net/http_server_handlers_map2.h
```diff
@@ -72,7 +72,7 @@
       bool parse_res = epee::serialization::load_t_from_json(static_cast<command_type::request&>(req), query_info.m_body); \
       if (!parse_res) \
       { \
-         MERROR("Failed to parse json: \r\n" << query_info.m_body); \
+         MERROR("Failed to parse json request"); \
```
jeffro256: nit: capitalize

jeffro256 (commented):
ACK log injection changes, unsure about denying control characters due to backwards compatibility.

selsta:
@jeffro256 updated

jpk68 (approved):

PR #11152, opened 2026-08-19
wallet2: validate key image domain in reserve proofs

selsta:
Reported by zkao, a tool by zkSecurity


jpk68 (approved):

tobtoht (approved):

jeffro256:
I recommend #11155 instead, especially since new spend/reserve/hot-cold calls will be added for Carrot

PR #11155, opened 2026-08-19
crypto: check key image in ring signature verif

jeffro256:
Alternative to #11152. Redundant for consensus due to `cryptonote::core::check_tx_inputs_keyimages_domain()`, but will help us avoid playing whack-a-mole every time we use `crypto::check_ring_signature()` in wallet code or elsewhere.

Performance impact: one scalar-point multiplication, 4 field element reductions, and 5 32-byte string compares per call to `crypto::check_ring_signature()`. Daemons with checkpoints enabled will not take any performance impact.


UkoeHB (approved):

jeffro256:
Removed $15 * l$ assertion because `sc_mul()` reduces the result mod $l$ so it failed. One can check with e.g. Python on the side if they want to check for validity.

UkoeHB (approved):

thomasbuilds (approved):

jpk68 (approved):

selsta (approved):

PR #11158, opened 2026-08-19
lmdb: refine batch size estimation

selsta:


j-berman (approved):

nahuhh:
This whole section / logic is wacky imo. Resizes map entirely disproportional to the actual data being used.

caused me some headaches testing fcmp on arm, where the db was like 10gb, at 10% or less usage, but the map was being expanded to like 500gb.. or larger, i dont remember. Could have been in the TBs.

and on the other end of the spectrum, when the blocks are small, it resizes repeatedly by like 1gb

edit: i just wrote the above comment because i remembered the frustration (not due to the changes being made)

jpk68 (approved):

MaxXor:
> This whole section / logic is wacky imo. Resizes map entirely disproportional to the actual data being used.
> 
> caused me some headaches testing fcmp on arm, where the db was like 10gb, at 10% or less usage, but the map was being expanded to like 500gb.. or larger, i dont remember. Could have been in the TBs.
> 
> and on the other end of the spectrum, when the blocks are small, it resizes repeatedly by like 1gb
> 
> edit: i just wrote the above comment because i remembered the frustration (not due to the changes being made)

Maybe we should open an issue to keep track of a possible rework of this section?

PR #11167, opened 2026-08-21
wallet: encode equals signs in URIs

Ap4sh:
## Summary

`make_uri()` currently leaves `=` unescaped in `recipient_name` and `tx_description`

`parse_uri()` uses `=` as the parameter delimiter, so those generated URIs fail to round-trip

encode it as `%3D` and cover both fields with a round-trip test

this handles the behavior change separately from the conversion refactor in #10265, as discussed there

## Tests

- `build-system/tests/unit_tests/unit_tests --gtest_filter='uri.*'`
- `python3 tests/functional_tests/functional_tests_rpc.py /usr/bin/python3 "$PWD/tests/functional_tests" "$PWD/build-system" uri`

### DIFF
--- a/contrib/epee/include/net/abstract_http_client.h
+++ b/contrib/epee/include/net/abstract_http_client.h
@@ -43,7 +43,7 @@ namespace net_utils
   inline const char* get_unsave_chars()
   {
     //static constexpr char unsave_chars[] = "\"<>%\\^[]`+$,@:;/!#?=&";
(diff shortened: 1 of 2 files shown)


jpk68 (approved):

Ap4sh:
the Ubuntu failure is the tracked `address_book` DNSSEC flake from #10445

the URI functional test and the rest of the job passed, just to let you know 👍

selsta (approved):

PR #11168, opened 2026-08-21
blockchain: limit offsets cache size

j-berman:


jpk68 (approved):

src/cryptonote_core/blockchain.cpp
```diff
@@ -5055,10 +5065,14 @@ bool Blockchain::prepare_handle_incoming_blocks(const std::vector<block_complete
 
   // generate sorted tables for all amounts and absolute offsets
   size_t tx_index = 0;
+  size_t n_total_offsets = 0;
+  bool cache_full = false;
```
jeffro256: This is definitely a nit, but I'd like to remove the `cache_full` variable, since it can be calculated directly as a function of `n_total_offsets`. It reduces cognitive overhead while reading, and reduces the chance that a future change to this code causes a de-sync.

j-berman: Done

jeffro256 (approved):
Logic looks good, though it should definitely get a live test before merging

selsta (approved):

jpk68 (approved):

thomasbuilds (approved):

selsta:
> though it should definitely get a live test before merging

We will test syncing mainnet / testnet / stagenet before release with all the changes in.

plowsof (approved):

Issue #10791, opened 2026-06-18
[FEATURE REQUEST] --p2p-external-ipv6-address


xbeaudouin:
Hello,

A feature request about presenting a specifc P2P external IPv6 address instead of IPv6 binded by monerod should be very usefull.

The problem is concerned by some of uses who wants to be part of the P2P network and are behind CGNAT for IPv4, so in this case they cannot have a good IPv4 to present to the P2P network, what ever the mechanism we can use to punch hole in this ugly/deprecated network.

More and more ISP give IPv6 subnet on the LAN of their boxes, some of then are not (or not willing to) able to provide PD (prefix delegation) for having a better firewall than the ugly end user box we cannot remove because of xPON networks (this is the main problem).

So we have /64 usable and ... we have to use NAT66 with fd00::... to make stable IPv6 inside, and doing NAT66 to open ports.

Bitcoin Core does --externalip= that permit to present IPv6 exposed to ::/0, but monerod doesn't seems to have such solutions.

Do you think this feature request should be a good idea? Maybe if you have patches on 0.18.4+ I can experiment it (and even push it into FreeBSD ports since I am FreeBSD port commiter).

Kind regards
Xavier

xshthkr:
We could modify the way peers interpret the connection. When a peer receives a connection handshake from someone with an external IP set in config, they would use that IP for future connections instead of the TCP connection source IP. This would need changes to peerlist management. The P2P protocol would need to be extended to include an optional external IP address field.

We would also need to consider security. We cant blindly trust external addresses from unverified peers.

I'll try working on this. I'm relatively new to the codebase, however, and I appreciate any directions and advice.

xshthkr:
Im looking at the P2P code and Monero has a `m_our_address` infra for TOR and I2P. Could we generalize this for also the public clearnet zone? Lets say a command line flag `--p2p-external-ip` gives us the external IP, and the port (if not given by `--p2p-external-port`). We parse it into a `epee::net_utils::network_address` and set `m_our_address`. And then we advertise it in the handshake, which is currently being done only for TOR/I2P.

 in `node_server<>::handle_timed_sync()`

```cpp
...
const bool outgoing_to_same_zone = !context.m_is_income && !m_hide_my_port && zone.m_our_address.get_zone() == zone_type;
if (outgoing_to_same_zone)
{
  local_peerlist_new.insert(
    local_peerlist_new.begin() + crypto::rand_range(std::size_t(0), local_peerlist_new.size()),
[...]

PR #11171, opened 2026-08-22
p2p: support external IP, inject address into peerlist

xshthkr:
Implemented external IP to allow nodes behind NAT, containers, reverse proxies etc, to advertise a reachable external IP/port to peers. This tries to solve #5979, #10791.

Changes:

- added `--p2p-external-ip` supporting IPv4, IPv6 (`[addr]:port` and raw `addr`), with optional embedded ports
- generalizes Monero's existing `m_our_address` infrastructure (previously only populated for Tor `.onion` and I2P `.i2p` hidden services) for the public clearnet zone
- inject self-address into peerlist for public zone
- `--p2p-external-ip` accepts only IPs. `--p2p-external-port` accepts only ports, defaulting to bind/default port.
- added unit test cases covering valid IPv4, IPv6, and ports.

Passing release-tests.


src/p2p/net_node.inl
```diff
@@ -569,7 +570,32 @@ namespace nodetool
     public_zone.m_port_ipv6 = command_line::get_arg(vm, arg_p2p_bind_port_ipv6);
     public_zone.m_can_pingback = true;
     m_external_port = command_line::get_arg(vm, arg_p2p_external_port);
+    m_external_ip = command_line::get_arg(vm, arg_p2p_external_ip);
     m_allow_local_ip = command_line::get_arg(vm, arg_p2p_allow_local_ip);
+    if (!command_line::is_arg_defaulted(vm, arg_p2p_external_ip) && !m_external_ip.empty())
```
selsta, line 575:
```
--p2p-external-ip=198.51.100.1:28080
--p2p-external-port=38080
```

This edge case is not handled correctly, handshake advertises 38080 while peerlist injection advertises 28080.

src/p2p/net_node.inl
```diff
@@ -569,7 +570,32 @@ namespace nodetool
...
+      {
+        MERROR("External IP \"" << m_external_ip << "\" is not in public network zone");
+        return false;
+      }
+      if (!m_allow_local_ip && (adr->is_loopback() || adr->is_local()))
+      {
+        MERROR("External IP \"" << m_external_ip << "\" is local or loopback address. Use --allow-local-ip to allow.");
+        return false;
+      }
+      public_zone.m_our_address = std::move(*adr);
```
selsta, line 609: Port zero should be rejected or replaced with the actual bound port before advertisement.

xshthkr: Ill change this so that an explicit port 0 in the external IP is rejected. If the external IP string parses an explicit port 0, it fails immediately. If there are no ports in either external IP or external port, it falls back to the actual bound port.

src/p2p/net_node.inl
xshthkr: I've changed the external IP and port handling. Conflicting ports won't be allowed. Adding a valid non-zero port in `--p2p-external-ip` would be short hand for adding the external IP and external port flags. If either flag provides a valid port, it would be selected. If no ports are provided, it falls back to the bind port or default P2P port.

nahuhh: Do the existing --p2p-bind-ip (and --rpc-bind-ip) accept ports? Or just IPs?

xshthkr: `--p2p-bind-ip` accepts IPs. `--p2p-bind-port` accepts ports. I'll look into how those flags work and i'll try doing something similar in external port/ip if that could eliminate port conflicts/precedence entirely.

xshthkr:
`--p2p-external-ip` accepts only IPs and `--p2p-external-port` accepts ports, defaulting to bind/default port.

src/p2p/net_node.inl
```diff
@@ -569,7 +570,46 @@ namespace nodetool
...
+      if (!port_str.empty())
+      {
+        MERROR("--p2p-external-ip only accepts an IP address. Use --p2p-external-port to specify a port.");
+        return false;
+      } 
+      std::uint16_t final_port = m_external_port;
+      if (!final_port)
+      {
+        const std::uint16_t default_p2p_port = cryptonote::get_config(m_nettype).P2P_DEFAULT_PORT;
+        std::uint16_t bind_port = 0;
```
selsta, line 589: This logic doesn't fully appear correct yet, peers receive default port but server binds to an ephemeral port.

src/p2p/net_node.inl
```diff
@@ -569,7 +570,46 @@ namespace nodetool
...
+        epee::string_tools::get_xtype_from_string(bind_port, public_zone.m_port);
```
selsta, line 590:
```
--p2p-external-ip=2001:db8::1
--p2p-bind-port=18080
--p2p-bind-port-ipv6=28080
```

is not handled correctly.

xshthkr:
reworking this in #11210 to move IP address handling in `init_server` after socket binding

